PR #3983, opened 2024-06-28
Make `jj debug working-copy` work for non-default working copies

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/debug/mod.rs
+++ b/cli/src/commands/debug/mod.rs
@@ -22,6 +22,7 @@ pub mod snapshot;
 pub mod template;
 pub mod tree;
 pub mod watchman;
+pub mod working_copy;
 
 use std::any::Any;
 use std::fmt::Debug;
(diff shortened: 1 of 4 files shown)


cli/src/commands/debug/mod.rs
```diff
@@ -22,6 +22,7 @@ pub mod snapshot;
 pub mod template;
```
ilyagr, line 22: (Optional, these are just debug commands after all)

> We could instead have made the old `jj debug working-copy` command
work for all kinds of working copies (like the new command) and only
have extra information for the standard local-disk implementation. I
don't feel strongly either way and could do it other other way instead
if people prefer that.

I would prefer that.

If not, I think the description of local_working_copy should be more detailed. It can say something like "Print detailed information about a working copy, only works for ????". I'm not quite sure what should go into the ????.

ilyagr (approved):
We can always merge the two commands later.

yuja (approved):

cli/src/commands/debug/mod.rs
martinvonz: I kept it separate after all. My thinking is that it seems better to isolate code that's conditional on a certain backend, but I'm not sure why I feel that way. Maybe it's mostly that a similar command for a custom working copy will have to go into a separate command, so it's more consistent this way.

I update the help text to say that it only works for the default working copy type.

PR #3967, opened 2024-06-26
clean up remainder of old branch API, with minor behavior changes in CLI output

yuja:
# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #3987, opened 2024-06-28
cli: branch: print stats even if just one branch is updated

yuja:
Extracted from #3967, with no code changes.

cc @ilyagr 

# Checklist

If applicable:


jonathantanmy:
It makes sense to me to always print a message when a branch is deleted (and so on), but I'll let others give their opinion about the UX. The code change makes sense to me.

bnjmnt4n (approved):

PR #3989, opened 2024-06-29
lib id_prefix: look for divergent changes outside short prefix set 

ilyagr:
Fixes https://github.com/martinvonz/jj/issues/2476.

Previously, if there was a change id match within the short prefix
lookup set, `jj` would not look for commits with that same change id
outside the short prefix set. So, it wouldn't find the conflicted
commits for a commit with a divergent (AKA conflicted) change id.

# Checklist

If applicable:
- n/a I have updated the documentation (README.md, docs/, demos/)

### DIFF
--- a/lib/tests/test_id_prefix.rs
+++ b/lib/tests/test_id_prefix.rs
@@ -365,12 +365,19 @@ fn test_id_prefix_divergent() {
         c.shortest_change_prefix_len(repo.as_ref(), second_commit.change_id()),
         1
     );
-    // Short prefix does not find the first commit (as intended).
-    // TODO(#2476): BUG: Looking up the divergent commits by their change id prefix
-    // only finds the id within the lookup set.
+    // This tests two issues, both important:
+    // - We find both commits with the same change id, even though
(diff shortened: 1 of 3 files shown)


lib/tests/test_id_prefix.rs
```diff
@@ -253,3 +253,149 @@ fn test_id_prefix() {
...
+                .set_description(description)
+                .set_author(signature.clone())
+                .set_committer(signature);
+            if let Some(change_id) = change_id {
+                builder = builder.set_change_id(change_id);
+            }
+            builder.write().unwrap()
+        };
+
+    let first_commit = create_commit_with_change_id(root_commit_id, "first", None);
```
yuja, line 291: nit: maybe better to hardcode the change_id instead of relying on the seed?

yuja (approved):
Thanks.

lib/tests/test_id_prefix.rs
ilyagr: Good point, this simplifies a few things.

ilyagr:
Thank you, Yuya!

PR #3990, opened 2024-06-29
clippy fixes

ilyagr:


cli/src/commands/debug/watchman.rs
yuja, line 1: (no action needed) might be okay to disable "debug watchman" command (and module) if the feature is disabled.

yuja (approved):

thoughtpolice (approved):

cli/src/commands/debug/watchman.rs
ilyagr: Ideally, `jj debug watchman` would be a stub that accepts any arguments and prints an error message, I think.

Issue #2933, opened 2024-02-04
FR: search for text in commit history


jyn514:
**Is your feature request related to a problem? Please describe.**
I want to find the commit in which I removed a line of code. With git, I could use `git log -S foo`, or `git log -G foo` (the difference is subtle and not interesting for my purposes). I don't see any equivalent in jj.

**Describe the solution you'd like**
Add a way to filter commits by their diff. I think this could fit in well as a [revset](https://github.com/martinvonz/jj/blob/main/docs/revsets.md#functions) function; maybe `contents(foo)`, where `foo` is a string pattern? Then `jj log --revisions` would work without needing to add a new flag.

**Describe alternatives you've considered**
`jj log -p` and then searching in the pager, which *works* but is unnecessarily slow and shows more commits than the one i care about.

**Additional context**
```
; jj --version
jj 0.13.0-81b0e3bf3bf6f230740e93532b9a9767e8a4ef6e
```

yuja:
> I think this could fit in well as a [revset](https://github.com/martinvonz/jj/blob/main/docs/revsets.md#functions) function; maybe `contents(foo)`, where `foo` is a string pattern?

hg calls it `diffcontains(pattern)`. Implementing unoptimized version wouldn't be that hard.

martinvonz:
We'll have to decide whether paths passed on the CLI should feed into the revset. That is, should `jj log -r 'diffcontains(foo)' dir` search for changes to "foo" only inside `dir/` or also outside? We have a similar question with `jj log -p dir`, which currently filters both the log and the diff.

jyn514:
i would expect it to only search inside `dir/`, the same as `git log -S foo dir`. i think the alternative you're suggesting is "show commits that modify dir/ and also contain 'foo' anywhere in their diff"? i don't find that intuitive; in git the way i would do that is probably `git rev-list -- dir | xargs -I{} git rev-list {} -- foo`, but more likely i wouldn't try to do that at all.

jyn514:
oh, it turns out git has a `--full-diff` option to opt into that behavior. so that's another option available, limit both by default but only limit one if someone opts-in.

martinvonz:
[...]

PR #3991, opened 2024-06-29
str_util: support case‐insensitive string patterns

emilazy:
See the commit message for details. This resolves an old TODO and came up during the rework of https://github.com/martinvonz/jj/pull/3951. ~~I’m not totally sure about the defaults here, but the functionality seems good to have.~~

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj git clone some/nested/path` now creates the full directory tree for 
    nested destination paths if they don't exist.
 
+* String patterns now support case‐insensitive matching by suffixing any
+  pattern kind with `-i`. `mine()` uses case‐insensitive matching on your email
+  address unconditionally. Only ASCII case folding is currently implemented,
+  but this will likely change in the future.
+
 ### Fixed bugs
 
 ## [0.19.0] - 2024-07-03
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -125,11 +125,11 @@ pub trait RevsetFilterExtension: std::fmt::Debug + Any {
 pub enum RevsetFilterPredicate {
     /// Commits with number of parents in the range.
     ParentCount(Range<u32>),
-    /// Commits with description containing the needle.
+    /// Commits with description matching the pattern.
     Description(StringPattern),
-    /// Commits with author's name or email containing the needle.
+    /// Commits with author name or email matching the pattern.
     Author(StringPattern),
-    /// Commits with committer's name or email containing the needle.
+    /// Commits with committer name or email matching the pattern.
     Committer(StringPattern),
     /// Commits modifying the paths specified by the fileset.
     File(FilesetExpression),
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -687,8 +687,11 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
     });
     map.insert("mine", |function, context| {
         function.expect_no_arguments()?;
+        // Email address domains are inherently case‐insensitive, and the local‐parts
+        // are generally (although not universally) treated as case‐insensitive too, so
+        // we use a case‐insensitive match here.
         Ok(RevsetExpression::filter(RevsetFilterPredicate::Author(
-            StringPattern::Exact(context.user_email.to_owned()),
+            StringPattern::exact_i(&context.user_email),
         )))
     });
     map.insert("committer", |function, _context| {
@@ -2518,7 +2521,7 @@ mod tests {
         assert!(parse("mine(foo)").is_err());
         insta::assert_debug_snapshot!(
             parse("mine()").unwrap(),
-            @r###"Filter(Author(Exact("test.user@example.com")))"###);
+            @r###"Filter(Author(ExactI("test.user@example.com")))"###);
         insta::assert_debug_snapshot!(
             parse_with_workspace("empty()", &WorkspaceId::default()).unwrap(),
             @"NotIn(Filter(File(All)))");
--- a/lib/src/str_util.rs
+++ b/lib/src/str_util.rs
@@ -32,16 +32,26 @@ pub enum StringPatternParseError {
     GlobPattern(glob::PatternError),
 }
 
+fn parse_glob(src: &str) -> Result<glob::Pattern, StringPatternParseError> {
+    glob::Pattern::new(src).map_err(StringPatternParseError::GlobPattern)
+}
+
 /// Pattern to be tested against string property like commit description or
 /// branch name.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum StringPattern {
-    /// Matches strings exactly equal to `string`.
+    /// Matches strings exactly.
     Exact(String),
-    /// Unix-style shell wildcard pattern.
-    Glob(glob::Pattern),
-    /// Matches strings that contain `substring`.
+    /// Matches strings case‐insensitively.
+    ExactI(String),
+    /// Matches strings that contain a substring.
     Substring(String),
+    /// Matches strings that case‐insensitively contain a substring.
+    SubstringI(String),
+    /// Matches with a Unix‐style shell wildcard pattern.
+    Glob(glob::Pattern),
+    /// Matches with a case‐insensitive Unix‐style shell wildcard pattern.
+    GlobI(glob::Pattern),
 }
 
 impl StringPattern {
@@ -50,11 +60,12 @@ impl StringPattern {
         StringPattern::Substring(String::new())
     }
 
-    /// Parses the given string as a `StringPattern`. Everything before the
-    /// first ":" is considered the string's prefix. If the prefix is "exact:",
-    /// "glob:", or "substring:", a pattern of the specified kind is returned.
-    /// Returns an error if the string has an unrecognized prefix. Otherwise, a
-    /// `StringPattern::Exact` is returned.
+    /// Parses the given string as a [`StringPattern`]. Everything before the
+    /// first ":" is considered the string's prefix. If the prefix is
+    /// "exact[-i]:", "glob[-i]:", or "substring[-i]:", a pattern of the
+    /// specified kind is returned. Returns an error if the string has an
+    /// unrecognized prefix. Otherwise, a `StringPattern::Exact` is
+    /// returned.
     pub fn parse(src: &str) -> Result<StringPattern, StringPatternParseError> {
         if let Some((kind, pat)) = src.split_once(':') {
             StringPattern::from_str_kind(pat, kind)
@@ -63,26 +74,48 @@ impl StringPattern {
         }
     }
 
-    /// Creates pattern that matches exactly.
+    /// Constructs a pattern that matches exactly.
     pub fn exact(src: impl Into<String>) -> Self {
         StringPattern::Exact(src.into())
     }
 
-    /// Parses the given string as glob pattern.
+    /// Constructs a pattern that matches case‐insensitively.
+    pub fn exact_i(src: impl Into<String>) -> Self {
+        StringPattern::ExactI(src.into())
+    }
+
+    /// Constructs a pattern that matches a substring.
+    pub fn substring(src: impl Into<String>) -> Self {
+        StringPattern::Substring(src.into())
+    }
+
+    /// Constructs a pattern that case‐insensitively matches a substring.
+    pub fn substring_i(src: impl Into<String>) -> Self {
+        StringPattern::SubstringI(src.into())
+    }
+
+    /// Parses the given string as a glob pattern.
     pub fn glob(src: &str) -> Result<Self, StringPatternParseError> {
         // TODO: might be better to do parsing and compilation separately since
         // not all backends would use the compiled pattern object.
         // TODO: if no meta character found, it can be mapped to Exact.
-        let pattern = glob::Pattern::new(src).map_err(StringPatternParseError::GlobPattern)?;
-        Ok(StringPattern::Glob(pattern))
+        Ok(StringPattern::Glob(parse_glob(src)?))
     }
 
-    /// Parses the given string as pattern of the specified `kind`.
+    /// Parses the given string as a case‐insensitive glob pattern.
+    pub fn glob_i(src: &str) -> Result<Self, StringPatternParseError> {
+        Ok(StringPattern::GlobI(parse_glob(src)?))
+    }
+
+    /// Parses the given string as a pattern of the specified `kind`.
     pub fn from_str_kind(src: &str, kind: &str) -> Result<Self, StringPatternParseError> {
         match kind {
             "exact" => Ok(StringPattern::exact(src)),
+            "exact-i" => Ok(StringPattern::exact_i(src)),
+            "substring" => Ok(StringPattern::substring(src)),
+            "substring-i" => Ok(StringPattern::substring_i(src)),
             "glob" => StringPattern::glob(src),
-            "substring" => Ok(StringPattern::Substring(src.to_owned())),
+            "glob-i" => StringPattern::glob_i(src),
             _ => Err(StringPatternParseError::InvalidKind(kind.to_owned())),
         }
     }
@@ -96,9 +129,12 @@ impl StringPattern {
     ///
     /// This can be used to optimize map lookup by exact key.
     pub fn as_exact(&self) -> Option<&str> {
+        // TODO: Handle trivial case‐insensitive patterns here? It might make people
+        // expect they can use case‐insensitive patterns in contexts where they
+        // generally can’t.
         match self {
             StringPattern::Exact(literal) => Some(literal),
-            StringPattern::Glob(_) | StringPattern::Substring(_) => None,
+            _ => None,
         }
     }
 
@@ -106,8 +142,11 @@ impl StringPattern {
     pub fn as_str(&self) -> &str {
         match self {
             StringPattern::Exact(literal) => literal,
-            StringPattern::Glob(pattern) => pattern.as_str(),
+            StringPattern::ExactI(literal) => literal,
             StringPattern::Substring(needle) => needle,
+            StringPattern::SubstringI(needle) => needle,
+            StringPattern::Glob(pattern) => pattern.as_str(),
+            StringPattern::GlobI(pattern) => pattern.as_str(),
         }
     }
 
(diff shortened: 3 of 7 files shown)


yuja:
> This is a function argument rather than part of the pattern syntax as it doesn’t make sense for all uses of patterns to support case‐sensitivity

I think it should be a pattern kind so that the same syntax can be used in `jj branch list PATTERN` command for example. It's weird if these commands took a separate `--case-insensitive` flag.

> (e.g. Git branch settings, where the remote end has no support for case‐insensitive matching),

Perhaps, `.to_glob()` can be renamed to `.to_case_sensitive_glob()` and error out if the type was `Glob { .., case_insensitive: true }`?

> and it’s not clear how it would best be integrated into the pattern syntax anyway. The `glob` and `regex` crates also separate matching options like this from the pattern itself.

Something like `iglob:` or `icase-glob:` seems okay (though I'm not a big fan of these.)  For `regex:`, we won't need `iregex:`, but could be added for consistency.

@arxanas how do you deal with case-insensitive matches in git-branchless? Just using `regex:` pattern?
https://github.com/arxanas/git-branchless/wiki/Reference:-Revsets#patterns

> Case‐insensitivity is on by default; a casual search for commit names or descriptions is unlikely to care about specific case distinctions, the domain part of email addresses is de jure case‐insensitive, and the local‐part is case‐insensitive in practice too. This could be changed at a later date if it turns out to be undesirable default for any of these functions (`description("FIXME")`?).

Good point about email addresses, but I would make it case-sensitive by default mainly for consistency with command arguments and file patterns.

> This currently only handles ASCII case folding, due to the complexities of case‐insensitive Unicode comparison and the `glob` crate’s lack of support for it.

Sounds good to me.

Thanks!

emilazy:
Thank you for the review!

> I think it should be a pattern kind so that the same syntax can be used in `jj branch list PATTERN` command for example. It's weird if these commands took a separate `--case-insensitive` flag.

I was wavering on whether it should be part of the pattern or not, so I can switch back. Although I’m not sure how much use there is for searching branch names case‐insensitively, especially when Git remotes don’t support it. I didn’t bother adding it to the corresponding revset functions for that reason; I didn’t want to make people expect they could treat branch names case‐insensitively in all contexts. It would be nice for it to work in contexts where it does make sense without needing an explicit extension to the interface, though, I agree.

> Perhaps, `.to_glob()` can be renamed to `.to_case_sensitive_glob()` and error out if the type was `Glob { .., case_insensitive: true }`?

Sure, that works. It feels kind of awkward for some types of pattern to be rejected in some contexts, but I guess it’ll be inevitable for those code paths to error out if/when regex patterns are supported anyway.

> Something like `iglob:` or `icase-glob:` seems okay (though I'm not a big fan of these.) For `regex:`, we won't need `iregex:`, but could be added for consistency.

Taking inspiration from regex syntax (a dangerous phrase!), what about…

```
exact/i:foo
substring/i:foo
glob/i:foo

# Short for `substring/i:foo`
/i:foo
```

Too weird? I like that it clearly separates out the type from the common options, and I think that we probably want something shorter than `substring/i:` for quick commit searches if it’s not going to be the default, but I admit that `/i:foo` makes me think of DOS. Maybe there’s something in the region of this we can bikeshed. `substring(i):…`? I guess we could just make `i:` short for `isubstring:`.

> Good point about email addresses, but I would make it case-sensitive by default mainly for consistency with command arguments and file patterns.

Makes sense; it makes this less of a backwards‐incompatible change and can always be adjusted later. How about keeping `mine()` case‐insensitive, since it’s always dealing with email addresses? Not that it really matters, since it’s not likely common to have commits with email addresses differing only in case (though you can find it in Nixpkgs, like every other kind of metadata strangeness in the world).

yuja:
> Taking inspiration from regex syntax (a dangerous phrase!), what about…
> 
> ```
> exact/i:foo
> substring/i:foo
> glob/i:foo
> 
> # Short for `substring/i:foo`
> /i:foo
> ```
> 
> Too weird? I like that it clearly separates out the type from the common options, and I think that we probably want something shorter than `substring/i:` for quick commit searches if it’s not going to be the default, but I admit that `/i:foo` makes me think of DOS. Maybe there’s something in the region of this we can bikeshed. `substring(i):…`? I guess we could just make `i:` short for `isubstring:`.

It's unclear to me where we set a boundary (e.g. `cwd-glob:` vs `glob/cwd:` in file pattern), so I wouldn't introduce `/{option}` syntax. That said, I don't have strong feeling. If people like `/i`, I won't be against.

`i:` for `isubstring:` sounds good to me. If we add a short form, the canonical form could be `icase-substring:`.

> How about keeping `mine()` case‐insensitive, since it’s always dealing with email addresses? Not that it really matters, since it’s not likely common to have commits with email addresses differing only in case (though you can find it in Nixpkgs, like every other kind of metadata strangeness in the world).

That makes sense to me. Inline comment will help understand why it's case insensitive.

emilazy:
I’ve switched this over to being part of string patterns, and settled on `{kind}-i:…`/`i:…` syntax for now to keep the option clearly separated from the base kind without adding any new weird symbols. It was a little gnarlier than expected because of the default pattern kind being dependent on context (which I didn’t realize before and don’t entirely love, but it is what it is). I’d appreciate it if you could take another look!

lib/src/revset.rs
```diff
@@ -2293,28 +2295,58 @@ mod tests {
...
+            @r###"
+        Filter(
+            Author(
+                Substring(
+                    "foo@",
+                    Sensitive,
+                ),
+            ),
+        )
+        "###);
```
emilazy, line 2307: It might be nice if we can make the filter handle short two‐parameter structures like this; not sure the best way to do it, though.

yuja: Maybe it's simpler to add `ExactIcase`, `SubstringIcase`, etc. as separate variants?

Most of the users handle `Exact(_, Sensitive)`/`Exact(_, Insensitive)` separately, and it makes some sense that `exact()`/`is_exact()` implies case-sensitive match.

lib/src/str_util.rs
```diff
@@ -32,105 +32,189 @@ pub enum StringPatternParseError {
...
+        //
+        // For some discussion of this topic, see:
+        // <https://github.com/unicode-org/icu4x/issues/3151>
         match self {
-            StringPattern::Exact(literal) => haystack == literal,
-            StringPattern::Glob(pattern) => pattern.matches(haystack),
-            StringPattern::Substring(needle) => haystack.contains(needle),
+            StringPattern::Exact(literal, CaseSensitivity::Sensitive) => haystack == literal,
+            StringPattern::Exact(literal, CaseSensitivity::Insensitive) => {
+                haystack.to_ascii_lowercase() == literal.to_ascii_lowercase()
```
yuja, line 203: nit: `eq_ignore_ascii_case()` can be used

lib/src/str_util.rs
```diff
@@ -32,105 +32,189 @@ pub enum StringPatternParseError {
...
+                if needle.is_empty() {
+                    Some("*".into())
+                } else {
+                    Some(format!("*{}*", glob::Pattern::escape(needle)).into())
+                }
             }
+            StringPattern::Glob(pattern, CaseSensitivity::Sensitive) => {
+                Some(pattern.as_str().into())
+            }
+            _ => None,
```
yuja, line 179: nit: I'll list all variants exhaustively unless that looks too verbose.

lib/src/str_util.rs
```diff
@@ -32,105 +32,189 @@ pub enum StringPatternParseError {
...
     }
 
-    /// Parses the given string as pattern of the specified `kind`.
+    /// Parses the given string as a pattern of the specified `kind`.
     pub fn from_str_kind(src: &str, kind: &str) -> Result<Self, StringPatternParseError> {
-        match kind {
-            "exact" => Ok(StringPattern::exact(src)),
-            "glob" => StringPattern::glob(src),
-            "substring" => Ok(StringPattern::Substring(src.to_owned())),
+        let (base_kind, case_sensitivity) = match kind.strip_suffix("-i") {
```
yuja, line 105: nit: I'll list "exact", "exact-i", ... without resolving `case_sensitivity` separately. If we add "regex:", "regex:" and "regex-i:" will be processed differently.

lib/src/str_util.rs
```diff
@@ -32,105 +32,189 @@ pub enum StringPatternParseError {
...
+    /// Parses the given string as a pattern of the specified `maybe_kind` if
+    /// present, or `default_kind` otherwise.
+    pub fn from_str_maybe_kind(
+        src: &str,
+        maybe_kind: Option<&str>,
+        default_kind: &str,
+    ) -> Result<Self, StringPatternParseError> {
+        match maybe_kind {
+            None => Self::from_str_kind(src, default_kind),
+            Some("i") => Self::from_str_kind(src, &(default_kind.to_owned() + "-i")),
```
yuja, line 126: I think it's less confusing if "i:" is always mapped to "contains-i:".

It's unfortunate that we have two defaults, "substring:"/"exact:". It'll be nice if we can eventually pick "exact:" as the default, but I'm not sure if that level of breaking change is acceptable.

lib/src/revset.rs
emilazy: Yeah, that makes sense. The other strategies I had toyed around with were putting the case‐sensitivity in a wrapper struct, which didn’t seem better, and representing every pattern as a glob, which I kind of liked for implementation uniformity and maximum expressivity; all the existing operations can still be supported by e.g. checking that there aren’t any unescaped meta‐characters for conversion to an exact string. Of course, if we switched to supporting regexes that would get more complex. Separate constructors might be the best thing for now.

lib/src/str_util.rs
emilazy: Thanks, I didn’t know about that!

lib/src/str_util.rs
emilazy: It’s just the same three variants again but with `CaseSensitivity::Insensitive`. I can change it.

lib/src/str_util.rs
emilazy: I think it could be similar; we’d do e.g. `RegexBuilder::new(pat).case_insensitive(case_sensitivity == CaseSensitivity::Insensitive).build()`. But admittedly that’s not exactly saving much verbosity at that point and I probably got too clever here; I can change it.

lib/src/str_util.rs
emilazy: Hmm, it’d be pretty weird if putting `i:` in front of a branch pattern suddenly started operating on other branches because it turned into a substring match, wouldn’t it? On the other hand, for log queries, having to write `description(substring:todo)` (or even `contains:todo`) and not being able to do `author(martinvonz)` for log queries sounds really inconvenient (though I guess we could make aliases like `c:`, `ci:`, but it seems a bit obscure).

I do really dislike the inconsistency, but I’m not sure what the right thing would be here, even long‐term. Here’s my best idea: maybe glob should be the default, so that matching is exact by default, substring matching is as easy as `*foo*` for simple queries, and you only have to worry about the syntax if your query contains one of `?*[`? Then `i:` could be `glob-i:`, of course. Conveniently, globs are the most advanced syntax you can use for Git branches (because of protocol restrictions).

lib/src/revset.rs
emilazy: I sketched up the separate constructors version locally and it definitely feels nicer than what I have in this PR; I’ll hold off on pushing it until the other stuff is hashed out.

lib/src/str_util.rs
emilazy: I forgot to mention that moving to that default would of course have to include optimizing globs to exact patterns where possible, but that should be easy given I already implemented an equivalent check without issues (even just using `Exact` when there are no meta‐characters in the string should suffice for ≥90% of cases).

emilazy: I couldn’t resist and tried implementing the `glob` default locally. I like it! I think it’s probably the most viable universal default we could move to, and I’d be happy to finish off the work. It’s clearly too big and too breaking a change to do here, though, so I’d like to propose keeping the current PR behaviour for now. It’s the least surprising and most compatible thing given the existing inconsistency, and we can revisit the question afterwards, since any resolution to the inconsistency would break much more anyway.

martinvonz (commented):
> `mine()` matches case‐insensitively
unconditionally, since email addresses are conventionally
case‐insensitive and it doesn’t take a pattern anyway.

Is that already the case or do you mean that we should make it work that way? (It sounds like you're saying the former, but it doesn't look like that's what the code does.)

martinvonz:
> > `mine()` matches case‐insensitively
> > unconditionally, since email addresses are conventionally
> > case‐insensitive and it doesn’t take a pattern anyway.
> 
> Is that already the case or do you mean that we should make it work that way? (It sounds like you're saying the former, but it doesn't look like that's what the code does.)

Ah, you mean that it behaves that way after the commit. We don't really have a convention for what present tense means in commit messages. I use present tense to describe how it behaves before the patch (not saying you have to follow that, just explaining why I was confused).

emilazy:
I agree that it’s confusing; I’ll reword it. I try to stick to the Git quasi‐standard of imperative present tense, but it can get so awkward when going into detail, so I frequently slip…

(Thank you for the repository invite, by the way!)

lib/src/str_util.rs
yuja: I assumed that if pattern kind is specified in e.g. `jj branch delete i:foo`, the user would want to do some non-exact matching. And it would be easier to understand if `i:` is documented as short for `substring-i:`, not an context-dependent prefix. (If the syntax were `/i:`, it would look like short for `<default>/i:`, though.)

Maybe better to not add `i:` because at least two folks recognize it differently?

> Here’s my best idea: maybe glob should be the default, so that matching is exact by default,

It might be okay in revset, but I wouldn't expect CLI arguments are glob by default. It would be weird that `jj branch create NAME` isn't inherently glob, but `jj branch delete NAME` were.

lib/src/revset.rs
yuja: > The other strategies I had toyed around with were putting the case‐sensitivity in a wrapper struct, which didn’t seem better, and representing every pattern as a glob,

Beware that `StringPattern` is the type representing a pattern expression, and is passed to revset engine. It's similar to `RevsetExpression`/`FilesetExpression`, and the evaluation backend will do whatever transformation it needs there.

It might be convenient if the pattern had `.compile_glob()`/`.compile_regex()`, but I wouldn't apply heavy preprocessing to `StringPattern` itself.

lib/src/str_util.rs
emilazy: I’ve removed `i:`; the ergonomics aren’t fantastic now, but the foundation is in place for us to decide what we want the syntax/semantics to look like later, and I mostly just wanted this plumbing internally for the `.mailmap` work anyway.

I personally don’t find `create` and `delete` taking different syntaxes that weird; the ambiguity already exists for `jj branch {create,delete} exact:weird-name`, and glob syntax in branch names is pretty marginal (not sure how it would even works with refspecs?). Creation and queries usually use different syntax, after all. But hopefully we can bikeshed what the syntax and semantics should be separately to this PR :) The changes should be simpler now; I’d appreciate it if you could take another look!

yuja (approved):
Thanks!

FWIW, for mailmap, maybe we could also leverage `regex:` pattern. It's more expressive and supports case-insensitive matches. I'm thinking of adding it as it'll be useful in `diff_contains(pattern)` revset #2933.

CHANGELOG.md
```diff
@@ -125,6 +125,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj squash` now accepts a `--keep-emptied` option to keep the source commit.
 
+* String patterns now support case‐insensitive matching by suffixing any
+  pattern kind with `-i`. `mine()` uses case‐insensitive matching on your email
+  address unconditionally. Only ASCII case folding is currently implemented,
+  but this will likely change in the future.
+
```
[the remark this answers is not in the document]
emilazy: Whoops!

lib/src/str_util.rs
```diff
@@ -96,41 +129,86 @@ impl StringPattern {
...
+            StringPattern::SubstringI(needle) => haystack
+                .to_ascii_lowercase()
+                .contains(&needle.to_ascii_lowercase()),
+            StringPattern::Glob(pattern) => pattern.matches(haystack),
+            StringPattern::GlobI(pattern) => pattern.matches_with(
+                haystack,
+                glob::MatchOptions {
+                    case_sensitive: false,
+                    ..glob::MatchOptions::new()
+                },
```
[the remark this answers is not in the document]
emilazy: Added to `test_revset.rs`.

emilazy:
Hmm, for the `.mailmap` use‐case it’d be escaping a literal string to use as a regex just to get case‐insensitivity. I guess if we don’t want to support end‐user case‐insensitive querying without using a regex that’d be fine, but it doesn’t seem like the most natural way to implement it. I would be happy to see regex querying in general though!

yuja:
> for the `.mailmap` use‐case it’d be escaping a literal string to use as a regex just to get case‐insensitivity.

I thought we could leverage regex to combine multiple matches, but that might not be needed?

emilazy:
Oh, yes, that would work. I’m not sure if matching against `/(?i)^(a|b|c)$/` is better or worse for a database engine than `lower(x) = "a" OR lower(x) = "b" or lower(x) = "C"`; I guess it’d probably be faster for a local backend just because of the reduced indirection. Do you think that would be a better approach for all case‐insensitive matching, or should we merge this now and consider regex stuff as a potential optimization later? I’m also wondering if regex syntax compatibility issues between Rust and database engines might get in the way of making that happen soon.

yuja:
I don't have strong feeling. This PR looks good, but I'm not sure if we like `glob-i:` syntax. I don't know if we can come up with a better syntax, though.

Good point about database query. Regex version would be harder to optimize in principle.

jennings:
I know I'm late to the party, but as long as we've explicitly asked for bikeshedding...

- I think case-insensitive should be the default
- I think this should apply to file patterns as well.
- I think the hyphen should be omitted (`globi:` or `iglob:`)

# Case sensitivity

For string patterns, I think it's more user friendly to do a case-insensitive search by default, and let the user opt-in to case-sensitivity if they want it. That way the user can _usually_ be lazy and not worry about it, but has the option to be strict when necessary.

For file patterns, case-insensitive file systems are the default on Windows and macOS, and I think users are accustomed to not needing to be precise about casing.

# Hyphen

Especially on Windows, I feel like there's a lot of typing symbols when using jj. PowerShell treats parens without a `$` as a subshell, so most revsets need to be quoted. And then many string patterns require quotes around the value.

I don't think I can do anything about the quoting, but the hyphen feels unnecessary to me.

If I could choose anything unilaterally, I would probably have three kinds (using glob as the example):

- `iglob:` for case-insensitive
- `cglob:` for case-sensitive (the `c` prefix comes from PowerShell)
- `glob:` for either smart-case or user configuration

Many search interfaces support smart-casing: Case-insensitive if the search term is all lowercase, and case-sensitive if the search term contains any uppercase characters. I think that's a good default for unqualified `glob:`.

yuja:
Making the default case-insensitive has performance implications (especially in file matchers), so let's discuss that separately. It's rather a big change.

Smart-case seems useful, and can be introduced later. The underlying machinery wouldn't be different.

So the remaining issue is hyphen or not? Without hyphenation, `root-glob:` file pattern will be `root-iglob:`?
https://martinvonz.github.io/jj/prerelease/filesets/#file-patterns

emilazy:
I agree that a glob of some kind would be a good default for string patterns, maybe even a case‐insensitive one. It needs separate bikeshedding though, since currently some patterns default to `substring` and some to `exact`, which is a mess. I do have an implementation locally that seems pretty good.

For file patterns I’m not so sure. Case‐sensitivity is common filesystem semantics (albeit mostly on Linux rather than Windows or macOS) that I believe Jujutsu mostly inherits; I think we’d need to enforce case‐folding in trees for that to be the right choice, as nothing stops you having both `readme` and `README` in a repository.

Two motivations for the hyphen in `…-i`:

* Setting off the case‐insensitivity visually as a “modifier” to a base pattern kind, rather than its own entirely separate thing, hence my original weird proposal of `glob/i`. (Maybe in the future we’ll want more options? Although hopefully not.)

* Consistency with the file pattern kinds, which use hyphens to set off their own modifiers.

It just seemed a bit cleaner to me. The ergonomics aren’t great, but I already accepted we can’t get good ergonomics here without further discussion. I’d like `i:` for case‐insensitive matching with the same pattern kind as whatever the default is, but it requires making patterns more consistent in general so that users aren’t surprised.

That said, I mostly just implemented this to have the internals there so that the `.mailmap` implementation can use it, and I’m not particularly attached to the bikeshed colour, so if people prefer dropping the hyphen I’ll do that. Anything that makes the behaviour of patterns more consistent will be a fairly significant breaking change anyway, so we can adjust the decisions made here at the same time without further penalty.

jennings:
I’m so used to backwards compatibility being critical that I assumed it was important to get the naming right before the feature merges. But we see small breaking changes in most releases so clearly they’re acceptable.

I’m happy to have this merge now with any naming and discuss again later.

emilazy:
That’s the privilege of working on pre‐1.0 software :)

Not that we should gratuitously break things, but I think we’re all agreed that the current state of string patterns isn’t ideal and that it’s too big a job to fix in this PR. I’ll just hit the merge button now so I can finally finish off the `.mailmap` work and hopefully we can have a discussion (FWIW my preference is for globs as the default format, and I’m agnostic about case handling – case‐sensitive across the board might be best, as case‐insensitivity wouldn’t work properly for naming branches and filenames).

PR #3992, opened 2024-06-29
templater: use .map_or() to silence clippy without sacrificing readability

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commit_templater.rs
+++ b/cli/src/commit_templater.rs
@@ -1027,15 +1027,8 @@ impl RefNamesIndex {
         }
     }
 
-    #[allow(unknown_lints)] // XXX FIXME (aseipp): nightly bogons; re-test this occasionally
-    #[allow(clippy::manual_unwrap_or_default)]
-    #[allow(clippy::manual_unwrap_or)] // https://github.com/rust-lang/rust-clippy/issues/13018
     pub fn get(&self, id: &CommitId) -> &[Rc<RefName>] {
-        if let Some(names) = self.index.get(id) {
-            names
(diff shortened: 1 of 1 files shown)


PhilipMetzger (approved):

ilyagr (approved):
Thanks, this seems to work! Looks better too.

Of course, it's only a matter of time before clippy suggests replacing that with `unwrap_or` :)

cli/src/commit_templater.rs
```diff
@@ -1027,14 +1027,8 @@ impl RefNamesIndex {
...
 
-    #[allow(unknown_lints)] // XXX FIXME (aseipp): nightly bogons; re-test this occasionally
-    #[allow(clippy::manual_unwrap_or_default)]
     pub fn get(&self, id: &CommitId) -> &[Rc<RefName>] {
-        if let Some(names) = self.index.get(id) {
-            names
-        } else {
-            &[]
-        }
+        self.index.get(id).map_or(&[], |names| names)
```
ilyagr, line 1031: Optional: add a comment why `unwrap_or` does not work here? `|names| names` does look slightly awkward.

Also, I wouldn't mind understanding the issue better myself.

It's fine if you don't feel like it.

ilyagr:
Cc: https://github.com/rust-lang/rust-clippy/issues/13018

cli/src/commit_templater.rs
yuja: Ok, I added type annotation `names: &Vec<_>` to better explain `|x| x`.

PR #3994, opened 2024-06-30
docs: add customization example of color-words diff style

yuja:
# Checklist

If applicable:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -166,6 +166,18 @@ useful reminder to fill in things like BUG=, TESTED= etc.
 ui.default-description = "\n\nTESTED=TODO"
 ```
 
+### Diff colors and styles
+
+In color-words diffs, hunks are rendered with underline. You can override the
+default style with the following keys:
+
+```toml
+[colors]
+# Highlight hunks with background
+"diff removed token" = { bg = "#221111", underline = false }
+"diff added token" = { bg = "#002200", underline = false }
+```
+
 ### Diff format
 
 ```toml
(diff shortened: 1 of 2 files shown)


PhilipMetzger (approved):

docs/config.md
```diff
@@ -166,6 +166,18 @@ useful reminder to fill in things like BUG=, TESTED= etc.
...
 ```
 
+### Diff colors and styles
+
+In color-words diffs, hunks are rendered with underline. You can override the
+default style with the following keys:
+
+```toml
+[colors]
+# Highlight hunks with background
```
martinvonz, line 176: Worth saying that we use a stronger red because the human eye is less sensitive to red than to green? I'm worried users will otherwise waste time thinking it's a mistake and try using 220000 instead before the realize that's hard to see.

I use 221111 in my config. I think I tried 330000 and liked 221111 better, maybe because it's easier to see the foreground color on it.

ilyagr: (No action needed, maybe I should copy this elsewhere)

I was wrong about `delta` using 16 colors. For reference, here's what it uses. I like its colors, but I'm not sure it's objectively better than what you have.

Picture for reference:

image

the duller red is `\033[48;2;63;0;1,`, which I think corresponds to RGB of (63,0,1) (decimal), which I think is `#3F0001`.

The brighter red is (144,16,17), `#901011`.

The (duller) green is `(0,40,0)`, `#002800`.

There is also a brighter green (not shown), `(0,96,0)`, `#006000`.

Amusingly, GitHub shows the colors :)

Note that delta finds it necessary to remove syntax highlighting for deleted sections.

It also does some sort of terminal capability detection; it uses 256 colors on `xterm`.

ilyagr: (I updated Martin's comment to show the colors there too, though it doesn't help as much as I hoped)

yuja: Updated to `#221111` (without comment about color intensity.) It's less likely that user would think `#221111` is a typo.

PR #3995, opened 2024-06-30
fake-editor: replace expectpath with dump-path

mlcui-corp:
This allows us to assert the expected path in the test itself, rather than in the fake editor.

(This will be used for when we de-UNC-ify invocations of the editor, see #3986 for more details.)

# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #3996, opened 2024-06-30
windows: avoid UNC paths in `run_ui_editor`

mlcui-corp:
~~**This PR depends on #3995. Please only review the last commit until that PR has been merged.**~~

See https://github.com/martinvonz/jj/issues/3986 for more details. This is a no-op for non-Windows.

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_commit_command.rs
+++ b/cli/tests/test_commit_command.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use crate::common::TestEnvironment;
 
@@ -71,6 +71,24 @@ fn test_commit_with_editor() {
 
     JJ: Lines starting with "JJ: " (like this one) will be removed.
     "###);
+}
+
+#[test]
+fn test_commit_with_editor_avoids_unc() {
+    let mut test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let workspace_path = test_env.env_root().join("repo");
+    let edit_script = test_env.set_up_fake_editor();
+
+    std::fs::write(edit_script, "dump-path path").unwrap();
+    test_env.jj_cmd_ok(&workspace_path, &["commit"]);
+
+    let edited_path =
+        PathBuf::from(std::fs::read_to_string(test_env.env_root().join("path")).unwrap());
+    // While `assert!(!edited_path.starts_with("//?/"))` could work here in most
+    // cases, it fails when it is not safe to strip the prefix, such as paths
+    // over 260 chars.
+    assert_eq!(edited_path, dunce::simplified(&edited_path));
 }
 
 #[test]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1680,6 +1680,7 @@ dependencies = [
  "criterion",
  "crossterm",
  "dirs",
+ "dunce",
  "esl01-renderdag",
  "futures 0.3.30",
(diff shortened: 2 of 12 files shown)


cli/tests/test_commit_command.rs
```diff
@@ -73,6 +73,28 @@ fn test_commit_with_editor() {
...
+
+    std::fs::write(edit_script, ["dumppath path"].join("\0")).unwrap();
+    test_env.jj_cmd_ok(&workspace_path, &["commit"]);
+
+    let editor_path = std::fs::read_to_string(test_env.env_root().join("path")).unwrap();
+    // While `assert!(!editor_path.starts_with("//?/"))` could work here in most cases,
+    // it fails when it is not safe to strip the prefix, such as paths over 260 chars.
+    assert_eq!(
+        editor_path,
+        dunce::simplified(std::path::Path::new(&editor_path))
```
martinvonz, line 92: nit: make `std::path::Path` unqualified (it's already imported anyway). Same in the other tests, except for `test_describe_command`. Oh, maybe that's why you didn't import it here? I would still import it here (seems like Clippy might start complaining about it some day; i'm a bit surprised it doesn't already). I don't have much opinion whether to also import it in `test_describe_command` (guarded by `#[cfg(windows)]`) or to keep it local there.

cli/tests/test_commit_command.rs
```diff
@@ -73,6 +73,28 @@ fn test_commit_with_editor() {
...
+fn test_commit_with_editor_avoids_unc() {
+    let mut test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let workspace_path = test_env.env_root().join("repo");
+    let edit_script = test_env.set_up_fake_editor();
+
+    std::fs::write(edit_script, ["dumppath path"].join("\0")).unwrap();
+    test_env.jj_cmd_ok(&workspace_path, &["commit"]);
+
+    let editor_path = std::fs::read_to_string(test_env.env_root().join("path")).unwrap();
```
martinvonz, line 87: nit: is it simpler to wrap this in `Path::new()`? then you don't need to call `.to_str().unwrap()` below, i think. Same in the other tests

martinvonz (approved):
Thanks for working on this. I don't think we have many devs on Windows who would know what to do here (I certainly don't).

cli/tests/test_commit_command.rs
mlcui-corp: Done - also applied in #3995.

cli/tests/test_commit_command.rs
mlcui-corp: I've imported it inside these just in case we remove the other `std::path::Path` import. PTAL

mlcui-corp:
Resolved comments - PTAL.

cli/tests/test_commit_command.rs
```diff
@@ -73,6 +73,27 @@ fn test_commit_with_editor() {
     "###);
 }
 
+#[test]
+#[cfg(windows)]
```
yuja, line 77: nit: If the test passes on the other platforms, I would remove `cfg(windows)` (and move `use std::path::PathBuf` to top.) Windows-only code paths are easier to break.

yuja (approved):

cli/tests/test_commit_command.rs
mlcui-corp: Done.

cli/tests/test_commit_command.rs
mlcui-corp: Resolved as part of the parent PR.

PR #3997, opened 2024-06-30
cli: refactor `unified_diff_hunks()`, fix number of last context lines

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -771,7 +771,8 @@ fn unified_diff_hunks<'content>(
         lines: vec![],
     };
     let diff = Diff::for_tokenizer(&[left_content, right_content], diff::find_line_ranges);
-    for hunk in diff.hunks() {
+    let mut diff_hunks = diff.hunks().peekable();
+    while let Some(hunk) = diff_hunks.next() {
         match hunk {
             DiffHunk::Matching(content) => {
                 let mut lines = content.split_inclusive(|b| *b == b'\n').fuse();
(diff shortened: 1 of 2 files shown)


cli/tests/test_diff_command.rs
```diff
@@ -857,6 +857,97 @@ fn test_diff_skipped_context_nondefault() {
...
+    )
+    .unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["new"]);
+    std::fs::write(
+        repo_path.join("file1"),
+        "1\n2\n3\n4\n5\n6\nR\n7\n8\n9\n10\n11\n",
+    )
+    .unwrap();
+
+    // N <= 2 * num_context_lines + 1
```
martinvonz, line 879: How do I read the `N` here? It seems to refer to two different things. The first `N` seems to be the number of context lines available on each side in the input. The second `N` seems to be the total number of lines in the output.

cli/src/diff_util.rs
```diff
@@ -730,8 +730,8 @@ enum DiffLineType {
 }
 
 struct UnifiedDiffHunk<'content> {
-    left_line_range: Range<usize>,
-    right_line_range: Range<usize>,
+    removed_line_range: Range<usize>,
+    added_line_range: Range<usize>,
```
martinvonz, line 734: I think the old names were clearer because the left/right ranges are not all removed/added. In the case of context lines, you can have a left/right line that's not removed or added.

martinvonz (approved):
Thanks!

cli/src/diff_util.rs
yuja: Good point. Dropped the rename patch.

cli/tests/test_diff_command.rs
yuja: They are both `N=5`. Added short comment describing what could go wrong if `N <= 2 * context` (`+1` for color-words.)

PTAL.

yuja: Ah, okay. The test for color-words diff is wrong. I should use `--context=1` here.

yuja: I'll follow up if the test changes aren't still good.

PR #3998, opened 2024-06-30
commit: add `--reset-author` option

scott2000:
Since `jj describe --reset-author` works, I also expected `jj commit --reset-author` to work, so I think this would be a good option to add.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   commits with no description) if authored by the current user.
   [#2000](https://github.com/martinvonz/jj/issues/2000)
 
+* `jj commit` now accepts `--reset-author` option to match `jj describe`.
+
 ### Fixed bugs
 
 * `jj git push` now ignores immutable commits when checking whether a
--- a/cli/src/commands/commit.rs
+++ b/cli/src/commands/commit.rs
@@ -39,6 +39,16 @@ pub(crate) struct CommitArgs {
     /// Put these paths in the first commit
     #[arg(value_hint = clap::ValueHint::AnyPath)]
     paths: Vec<String>,
+    /// Reset the author to the configured user
+    ///
+    /// This resets the author name, email, and timestamp.
+    ///
+    /// You can use it in combination with the JJ_USER and JJ_EMAIL
+    /// environment variables to set a different author:
+    ///
+    /// $ JJ_USER='Foo Bar' JJ_EMAIL=foo@bar.com jj commit --reset-author
+    #[arg(long)]
+    reset_author: bool,
 }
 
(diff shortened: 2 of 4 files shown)


PhilipMetzger (approved):
LG, it also makes sense if you read `commit` as `desc + new`.

yuja (approved):
It might be okay for `jj commit` to always reset author (or timestamps at least.) Is that surprising?

scott2000:
> It might be okay for `jj commit` to always reset author (or timestamp at least.) Is that surprising?

I think this might be surprising, since as @PhilipMetzger said, I tend to think of `jj commit` as being the same as `jj describe` followed by `jj new`, and I don't think `describe` should update the author information. If we do change it, it might be good to have it be opt-in through a config option.

Are there any existing cases where the behavior is different for `commit` compared to `desc + new`?

yuja:
> Are there any existing cases where the behavior is different for `commit` compared to `desc + new`?

No, I don't think so (except for splitting.)

I just thought it might make sense for people coming from other VCSs if `commit` updated author timestamp. I don't have strong feeling about that.

scott2000:
> I just thought it might make sense to people coming from other VCSs if `commit` updated author timestamp. I don't have strong feeling about that.

Yeah, it does seem like it would better match the expectations of new users coming from Git. One concern I have is that new users might also use `jj split` to replicate `git commit -p` and then be confused why it doesn't behave the same as `jj commit`.

If we update `jj commit`, we could also update `jj split` to reset the author on the first commit, but I'm not sure whether that would be surprising when splitting commits other than `@`, since semantically the second commit might be the "new" one.

I'm curious how many Git users actually know that there's separate author and committer timestamps. It could be the case that most people don't even know/care how it works in Git, in which case it wouldn't matter that jj behaves differently.

PhilipMetzger:
> > I just thought it might make sense to people coming from other VCSs if `commit` updated author timestamp. I don't have strong feeling about that.
> 
> Yeah, it does seem like it would better match the expectations of new users coming from Git. One concern I have is that new users might also use `jj split` to replicate `git commit -p` and then be confused why it doesn't behave the same as `jj commit`.

Let's not change the meaning of `commit` to perfectly match Git, there's no reason to as Jujutsu is a separate system.

yuja:
> One concern I have is that new users might also use `jj split` to replicate `git commit -p` and then be confused why it doesn't behave the same as `jj commit`.

I wouldn't consider `jj split` as a command to finish up the first commit, so I wouldn't expect it to update author timestamp. However, I see your point.

emilazy:
> I'm curious how many Git users actually know that there's separate author and committer timestamps. It could be the case that most people don't even know/care how it works in Git, in which case it wouldn't matter that jj behaves differently.

My anecdotal experience is that most Git users barely think about timestamps at all.

PR #3999, opened 2024-07-01
github: publish binaries on push to main

### LINKED ISSUE #2544: FR: Produce nightly/wip builds for end-user testing
While testing #2503, along with the recent performance-related changes that @yuja has been writing, I realized that it would be nice, both for users and developers, to have binaries produced directly from `main` on a regular basis.

The reason I want this now is because I want to test the musl-linked Linux v0.11 build vs a musl-linked Linux version of HEAD, among other things, to see how ongoing changes to reduce memory use/improve performance are going. That means we'll have both a baseline and a "latest results" to compare against for all our ongoing changes.

Also, doing this also helps ensure changes like #2503 won't accidentally break the release process in some way that won't be discovered until Martin tries to do it, since it uses a whole different workflow entirely than what normal developers do.

Ideally this would use the same `release.yml` workflow it uses today.

---

Some examples of this I've seen are:

- Creating one on every commit, like [wasmtime](https://github.com/bytecodealliance/wasmtime) and buck2 do.
- Do it once a day, instead, at something like 3AM UTC.

In both cases, we should choose a tag like `dev` or `nightly` to attach to this build, and overwrite it on every run of this workflow. This has two consequences:

- The tag needs to be regularly force pushed (obviously), and
- We will not store copies of *every* binary, forever.
  - If we tried to do that, GitHub will eventually reach out to us and tell us to stop.
  - Plus, it clutters up the release page to have prereleases for *every* commit.

See also: #2483, though they only overlap a bit.

thoughtpolice:
This will cause every `push` to main to produce some binary artifacts that users can enjoy.

Fixes #2544.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -1,0 +1,77 @@
+name: binaries
+
+on:
+  push:
+    branches:
+      - main
+
+permissions: read-all
+
+jobs:
+  binaries:
+    strategy:
+      fail-fast: false
+
+      matrix:
+        build: [linux-x86_64-musl, linux-x86_64-gnu, linux-aarch64-musl, linux-aarch64-gnu, macos-aarch64, win-x86_64]
+        include:
+        - build: linux-x86_64-musl
+          os: ubuntu-24.04
+          target: x86_64-unknown-linux-musl
+        - build: linux-x86_64-gnu
+          os: ubuntu-24.04
+          target: x86_64-unknown-linux-gnu
+        - build: linux-aarch64-musl
+          os: ubuntu-24.04
+          target: aarch64-unknown-linux-musl
+        - build: linux-aarch64-gnu
+          os: ubuntu-24.04
+          target: aarch64-unknown-linux-gnu
+        - build: macos-aarch64
+          os: macos-14
+          target: aarch64-apple-darwin
+        - build: win-x86_64
+          os: windows-2022
+          target: x86_64-pc-windows-msvc
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 15 # NOTE (aseipp): keep in-sync with the build.yml timeout limit
+
+    name: Build binary artifacts
+    steps:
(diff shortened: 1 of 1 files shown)


thoughtpolice:
Note: after playing around, this currently only builds the aarch64 Darwin builds, because the x86_64 runners are severly lagging at this point.

thoughtpolice:
This now includes glibc builds (`-gnu`) in addition to musl builds for Linux/aarch64. I can follow up with a change to `release.yml` to add glibc builds there, too. So I think this is ready to go.

.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,77 @@
+name: binaries
```
PhilipMetzger, line 1: Nit: Call it something like "nightly-binaries"?

PhilipMetzger (approved):
LGTM with minor nit.

.github/workflows/build-binaries.yml
thoughtpolice: They aren't really nightlies, because they happen on every push, not every night. That said, I think I'm just going to file another follow up PR that standardizes our workflow file naming (consistent capitalization, etc.)

PhilipMetzger: Why though? Imo "nightly" does not need to imply the time of the release, e.g rustc nightlies or other ToT builds.

thoughtpolice: But the rustc nightly channel (what people collectively know as "nightlies") is only updated [once a day](https://rust-lang.github.io/rustup-components-history/). The CI system may produce binaries for every commit, but those aren't "nightlies".

PhilipMetzger: ok, scrap the "nigthly" and just call it ToT binaries in the followup.

emilazy: FWIW I don’t think we’d be the first project to use “nightly” for things that aren’t technically nightly and I think it conveys the right general idea as a term of art. But if you can think of a better term then I’m all for that.

PR #763, opened 2022-11-18
diff: consider uncommon words to match only if they have the same count

martinvonz:
Patience diff starts by lining up unique elements (e.g. lines) to find matching segments of the inputs. After that, it refines the non-matching segments by repeating the process. Histogram expands on that by not just considering unique elements but by continuing with elements of count 2, then 3, etc.

Before this commit, when diffing "a b a b b" against "a b a b a b", we would match the two "a"s in the first input against the first two "a"s in the second input. After this patch, we ignore the "a"s because their counts differ, so we try to align the "b"s instead.

I have had this commit lying around since I wrote the histogram diff implementation 18 months ago. I vaguely remember thinking that the way I had implemented it (without this commit) was a bit weird, but I wasn't sure if this commit would be an improvement or not. The bug report from @chooglen today of a case where we behave differently from Git is enough to make me think that we make this change after all.

Many unit tests of the diff algorithm are affected, mostly because we no longer match the leading space in " a " against the space in " b" and similar.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)

PR #4010, opened 2024-07-02
diff: stricter uncommon lcs, match up leading/trailing ranges instead

yuja:
The last patch is copied from #763 with a few modifications.

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_obslog_command.rs
+++ b/cli/tests/test_obslog_command.rs
@@ -59,13 +59,13 @@ fn test_obslog_with_or_without_diff() {
     @  rlvkpnrz test.user@example.com 2001-02-03 08:05:10 66b42ad3
     │  my description
     │  Resolved conflict in file1:
-    │     1    1: <<<<<<< Conflict 1 of 1resolved
+    │     1     : <<<<<<< Conflict 1 of 1
     │     2     : %%%%%%% Changes from base to side #1
     │     3     : -foo
     │     4     : +++++++ Contents of side #2
     │     5     : foo
     │     6     : bar
-    │     7     : >>>>>>> Conflict 1 of 1 ends
+    │     7    1: >>>>>>> Conflict 1 of 1 endsresolved
     ◉  rlvkpnrz hidden test.user@example.com 2001-02-03 08:05:09 cf73917d conflict
     │  my description
     ◉  rlvkpnrz hidden test.user@example.com 2001-02-03 08:05:09 068224a7
--- a/lib/src/diff.rs
+++ b/lib/src/diff.rs
@@ -230,12 +230,18 @@ fn unchanged_ranges_lcs(
     // the LCS.
     let Some(uncommon_shared_words) = left_histogram
         .count_to_words
-        .values()
-        .map(|left_words| -> Vec<&[u8]> {
+        .iter()
+        .map(|(left_count, left_words)| -> Vec<&[u8]> {
             left_words
                 .iter()
                 .copied()
-                .filter(|left_word| right_histogram.word_to_positions.contains_key(left_word))
+                .filter(|left_word| {
+                    let right_count = right_histogram
+                        .word_to_positions
+                        .get(left_word)
+                        .map_or(0, |right_positions| right_positions.len());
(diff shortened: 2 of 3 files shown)


jonathantanmy:
I haven't looked in-depth into #763 to know if this is a good change, but in any case, if we want this change, we should add a test case that tests the case where a diff result contains both the results of the leading/trailing code and the LCS code. As it is, all tests pass with this diff:

```
diff --git a/lib/src/diff.rs b/lib/src/diff.rs
index 57fb9cc2..af19cdb1 100644
--- a/lib/src/diff.rs
+++ b/lib/src/diff.rs
@@ -177,7 +177,8 @@ pub(crate) fn unchanged_ranges(
     if common_leading_len > 0 {
         let (left_leading_ranges, left_ranges) = left_ranges.split_at(common_leading_len);
         let (right_leading_ranges, right_ranges) = right_ranges.split_at(common_leading_len);
-        let mut result = unchanged_ranges_lcs(left, right, left_ranges, right_ranges);
+        // let mut result = unchanged_ranges_lcs(left, right, left_ranges, right_ranges);
+        let mut result = vec![];
         result.splice(
             0..0,
             iter::zip(
@@ -197,7 +198,8 @@ pub(crate) fn unchanged_ranges(
             left_ranges.split_at(left_ranges.len() - common_trailing_len);
         let (right_ranges, right_trailing_ranges) =
             right_ranges.split_at(right_ranges.len() - common_trailing_len);
-        let mut result = unchanged_ranges_lcs(left, right, left_ranges, right_ranges);
+        // let mut result = unchanged_ranges_lcs(left, right, left_ranges, right_ranges);
+        let mut result = vec![];
         result.extend(iter::zip(
             left_trailing_ranges.iter().cloned(),
             right_trailing_ranges.iter().cloned(),
```

The non-test code looks reasonable. Thanks for separating out the logical steps into their own commits.

yuja:
> add a test case that tests the case where a diff result contains both the results of the leading/trailing code and the LCS code. As it is, all tests pass with this diff:

Good point. Actually there was a bug that leading/trailing matches weren't trimmed from both ends, and that was the reason we had recursion there, I think. I've added a patch that removes recursion.

jonathantanmy:
OK, removing the recursion makes sense and addresses my concern above.

Let me summarize my current thoughts for other reviewers:

- This PR makes LCS more strict by requiring that if we use uncommon words, they must appear the same number of times in the LHS and RHS. But now, a diff between "a a a" and "a a" will have no matches, so this PR adds the matching prefix/suffix rule to compensate.
- If we believe that we need the same-number-of-times rule, then this PR is good.
- I'm not sure whether we really need the same-number-of-times rule.

So I would say that this PR is good from a code standpoint, but I'm not sure whether it's good from a feature standpoint. Maybe someone from #763 could review this PR.

yuja:
> * This PR makes LCS more strict by requiring that if we use uncommon words, they must appear the same number of times in the LHS and RHS. But now, a diff between "a a a" and "a a" will have no matches, so this PR adds the matching prefix/suffix rule to compensate.
> 
> * If we believe that we need the same-number-of-times rule, then this PR is good.
> 
> * I'm not sure whether we really need the same-number-of-times rule.

Thanks for summarizing it.

To add some context, I think the current implementation is a bit weird in that we choose the first `min(left, right)` occurrences. There's no clue that the first N matching pairs should line up, not the last N. Practically, this change tends to disable pairing unbalanced whitespace characters in color-words diffs. The resulting diff looks more readable to me.

martinvonz (commented):
>  I think the result is good so long as "uncommon LCS" matching is attempted first.

What happens if we don't?

yuja:
> > I think the result is good so long as "uncommon LCS" matching is attempted first.
> 
> What happens if we don't?

could produce inferior diffs. If I understand it correctly, Git roughly does LCS -> widen -> LCS ..., Bzr does LCS -> LCS -> ... -> widen.

image
instead of
image

martinvonz (approved):
Approved partly based on Jonathan's review. Thanks!

PR #3830, opened 2024-06-05
feat(squash): Add `--keep-emptied` flag

matts1:
If the user was thinking in terms of files, instead of commits, then we shouldn't abandon the commit, as that can have side effects such as moving branch pointers.

Fixes #3815

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [x] I have updated the documentation (README.md, docs/, demos/)
- [N/A] I have updated the config schema (cli/src/config-schema.json)
- [x] I have added tests to cover my changes

PR #4014, opened 2024-07-03
feat(rebase): Rename --skip-empty to --skip-emptied.

matts1:
This is based on @martinvonz's comment in #3830 about the inconsistency between squash --keep-emptied and rebase --skip-empty.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/rebase.rs
+++ b/cli/src/commands/rebase.rs
@@ -35,7 +35,7 @@ use crate::cli_util::{
     short_commit_hash, CommandHelper, RevisionArg, WorkspaceCommandHelper,
     WorkspaceCommandTransaction,
 };
-use crate::command_error::{user_error, CommandError};
+use crate::command_error::{cli_error, user_error, CommandError};
 use crate::ui::Ui;
 
 /// Move revisions to different parent(s)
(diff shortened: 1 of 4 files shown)


cli/src/commands/rebase.rs
```diff
@@ -188,12 +188,16 @@ pub(crate) struct RebaseArgs {
     )]
     insert_before: Vec<RevisionArg>,
 
+    /// Deprecated. Use --skip-emptied instead.
+    #[arg(long, conflicts_with = "revisions")]
+    skip_empty: bool,
```
yuja, line 193: nit: add `hide = true`

cli/src/commands/rebase.rs
```diff
@@ -213,8 +217,14 @@ Please use `jj rebase -d 'all:x|y'` instead of `jj rebase --allow-large-revsets
         ));
     }
 
+    if args.skip_empty {
+        return Err(user_error(
```
yuja, line 221: nit: maybe `cli_error()` is more appropriate (but other deprecation messages use `user_error()` or `ui.warning()`.)

yuja (approved):
cc @ilyagr

cli/src/commands/rebase.rs
matts1: I think `cli_error` is probably the right error to return, even if other places have done it differently.

Issue #1869, opened 2023-07-14
FR: Implement JJ `run` 


PhilipMetzger:
This is the tracking issue for `run`, mostly for myself. 

Related to #405 

- [ ] Add a barebones run command, which is hidden.
- [ ] Add the VFS integration point, for EdenFS and CitC.
- [ ] Implement the worker interface for the Git backend.
- [ ] Add a beatiful TUI with [superconsole](../../../facebookincubator/superconsole), based on @arxanas suggestion in #1486. 

this should establish the base features of `run`. 

And to add, here's a simple wishlist based on Discord conversations: 

- Add `run --bisect`, to support a `git bisect -x` like workflow based on [git-branchless](../../../arxanas/git-branchless)'s `git test`.[^1] 
- Supporting bayesian based bisection  [^2]

[^1]: My suggestion in [Discord](https://discord.com/channels/968932220549103686/969829516539228222/1120450012623671398)
[^2]: A suggestion from @thoughtpolice in [Discord](https://discord.com/channels/968932220549103686/969291218347524238/1129456549253873774)

arxanas:
[The `scm-bisect` crate](https://crates.io/crates/scm-bisect) contains bisection logic already. For `git run --bisect`, it would be nice if we could centralize logic into `scm-bisect` so that other version control systems could potentially use it.

thoughtpolice:
As an update to the entry for bayesian bisection support, this just came out of Google pretty recently: Flake Aware Culprit Finding. https://research.google/pubs/pub52048/

Algorithms 1 and 2 in Section 3 effectively summarize the whole algorithm operationally.

PhilipMetzger:
@matts1 Moving this conversation here, as this is the correct place for it. 

> Comments on jj run
> @PhilipMetzger, a few suggestions on your design doc for jj run (is there a good way to give feedback on design docs on > github - comments on a google doc work pretty well, but there doesn't seem to be anything like that for github):
> 
> If I understand correctly, exactly one of --dry-run, --rebase, --reparent, and --readonly must be chosen
> Probably worth calling out on there. It's very unclear, especially because --clean is stuck between them.

Just `--reparent` and `--rebase` are exclusive, as they operate on the repos history.  `--readonly` only exists for specific workflows, while `--clean` is there for manual garbage collection until there's a better way. And the section is only there to provide an overview of all requested features. 

> Based on the fact that you say that you can write by default, and that it's parallel by default, I assume that the default is -- >reparent?

[...]

Issue #3808, opened 2024-06-01
FR: support running in the working copy for `jj fix`


arxanas:
**Is your feature request related to a problem? Please describe.**
`jj fix`(/`hg fix`) are currently specialized towards formatters. However, these formatters have to operate in-memory on single files, which limits the utility somewhat. There are a few cases where you would prefer the full flexibility of running in the working copy:

- To set up formatters in a much more simple way:
  - Like `jj fix --exec 'cargo fmt'` without any complicated configuration.
  - This scales to even moderate-size repos without much thought.
- To support tools that can't be feasibly adjusted to work in-memory.
  - Like my work's linter setup 🙃
  - Like if you wanted to run your existing [`pre-commit`](https://pre-commit.com/) checks.
- To run whole-repo analyses, such as lints or builds:
  - Like `jj fix --exec 'cargo clippy --fix'`.
- To run tools which would operate on the commit history, such as to create new commits, modify existing commits, etc.
  - This is how [`git-branchless-submit`](https://github.com/arxanas/git-branchless/wiki/Command:-git-submit) works with Phabricator: it runs `arc diff` and actually amends the commit message with the effective change ID (the Phabricator URL).
  - That's an advanced feature with several design considerations which is probably not important to get in the initial implementation of operating on the working copy.

**Describe the solution you'd like**
`jj fix` can operate in the working copy, similar to existing designs for `jj run`. It would ideally be able to run in the current workspace, or in parallel across multiple workspaces.

**Describe alternatives you've considered**
- Add a `--fix` flag to `jj run` instead. That's kind of weird because "fixes" are more or less the same thing to the user, so there's no reason to arbitrarily split them up into two different subcommands.

**Additional context**
- `git-branchless-test fix` is implemented to operate on working copies: https://github.com/arxanas/git-branchless/wiki/Command:-git-test#fixing-formatting-and-linting-issues

hooper:
I'm partial to the run/fix distinction being equivalent to the per-working-copy/per-file distinction, but mostly due to familiarity with such an implementation for `hg`. I wonder if there are other granularities we should consider, and how that impacts our UI choices?

The `--exec` flag feels like an undesirable Git-ism, but I think we agree that a terse way of specifying a command is important in both the per-working-copy model and the per-file model. `hg fix` is lacking this, although command aliases can bridge some of the gap. It's more natural in the per-working-copy model because less configuration is needed (or maybe some of the info is instead communicated by the subprocess running `jj` commands).

[...]

Issue #3577, opened 2024-04-26
FR: Generalized hook support


matts1:
**Is your feature request related to a problem? Please describe.**
I was looking at #2845 and realized that it didn't meet my needs because it didn't support pre-upload hooks. I investigated further into hooks, and the closest I found was #405, which rather than generalized hooks, specifically is trying to solve pre-commit hooks.

**Describe the solution you'd like**
I was thinking of treating hooks like a pub-sub system. jj would publish events, which would consist of some metadata. Your hooks would simply be subscriptions to some subset of the events, matching specific metadata. For example:
```proto
message CommitDescription {
   string commit = 1;
   string description = 2;
}

message HookEvent {
    string workspace = 1;
    oneof hook_type {
        CommitDescriptionHook commit_description = 2;
        PreUploadHook pre_upload = 3;
    }
    ...
}
```

We would then, similar to my #3575, run all hook binaries that have subscribed to that event in the global config.toml. Similar to #3575, these hook binaries would be passed a file descriptor with a connection to the jj grpc server.  For example, my global `config.toml` might look like:
```toml
[hooks]
subscribe_to_every_event = { binary = "/path/to/foo" }
pre_upload = { binary = "/path/to/bar", type = "pre_upload" }
specific_workspace = { binary = "/path/to/baz", workspace = "/path/to/workspace" }
```

We can also implement per-repo hooks in the same way that git has them by allowing additional subscribers from a `hooks/jj.toml` or similar, though I'd hold off on this in the short term due to security concerns.:

For example, a pre-upload hook that forces your commit to have an associated bug might look like:
```py
conn = grpc.connect(file_descriptor(os.environ["JJ_SERVER"]))
main = importlib.import("/path/to/foo_ext.py", "main")
event = conn.get_event()

if "BUG=" not in event.commit_description.description:
    bug = input()
    description = f"{event.commit_description.description}\nBUG={bug}"
    conn.set_description(commit_id=event.commit_description.commit_id, description=description)
```

Though similar to my suggestion in #3575, we could provide preludes for common languages, and simply make the hook:
```py
def main(conn, event):
    if "BUG=" not in event.commit_description.description:
        bug = input()
        description = f"{event.commit_description.description}\nBUG={bug}"
        conn.set_description(commit_id=event.commit_description.commit_id, description=description)
```

We would thus provide a method in jj along the lines of:
```rust
fn publish_event(event: &HookEvent) -> anyhow::Result {
[...]

PR #4021, opened 2024-07-03
run: Flesh out a bare implementation of `jj run`.

PhilipMetzger:
This is basically a MVP based on `fix`, caching is not implemented yet. The core functionality is in `run_inner()` and `rewrite_commit()`.

This is a draft to get further input from @arxanas and @hooper and all interested (builds since the 17.9.2024) . Also progress on #1869.  

TODO: rewrite trees and expose all files

# Checklist

If applicable:


cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+
+/// The result of a single command invocation in `run_inner`.
+struct RunJobResult {
+    /// The old `CommitId` of the commit.
+    old_id: CommitId,
+    /// The new `CommitId` for the commit.
+    rewritten_id: CommitId,
+    /// The new tree generated from the commit.
+    new_tree: Tree,
+}
```
PhilipMetzger, line 208: Does this make sense?

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+    tx: WorkspaceCommandTransaction<'a>,
+    sender: Sender<RunJobResult>,
+    jobs: usize,
+    shell_command: &str,
+    commits: &[StoredCommit],
+) -> Result<(), RunError> {
+    let mut command_futures = JoinSet::new();
+    for commit in commits {
+        command_futures.spawn(rewrite_commit(tx, commit.deref(), shell_command));
+    }
```
PhilipMetzger, line 221: this should only run `jobs` amount of processes at once, not all.

cli/src/commands/run.rs
```diff
@@ -53,17 +303,84 @@ pub fn cmd_run(ui: &mut Ui, command: &CommandHelper, args: &RunArgs) -> Result<(
...
-    let _jobs = match args.jobs {
+    let jobs = match args.jobs {
         Some(0) => return Err(user_error("must pass at least one job")),
         Some(jobs) => Some(jobs),
         None => std::thread::available_parallelism().map(|t| t.into()).ok(),
     }
     // Fallback to a single user-visible job.
     .unwrap_or(1usize);
+
+    let mut done_commits = HashSet::new();
```
PhilipMetzger, line 322: Does this make sense to track the done commits for the UI?

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+        MergedTreeId::Legacy(tree_id) => tree_id.hex(),
+        MergedTreeId::Merge(tree_ids) => {
+            let ids = tree_ids
+                .map(|id| id.hex())
+                .iter_mut()
+                .enumerate()
+                .map(|(i, s)| {
+                    // Incredibly "smart" way to say, append "-" if the number is odd "+"
+                    // otherwise.
+                    if i & 1 != 0 {
```
arxanas, line 89:
```suggestion
                    if i % 2 != 0 {
```

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+                        s.push('-');
+                    } else {
+                        s.push('+');
+                    }
+                    s.to_owned()
+                })
+                .collect_vec();
```
arxanas, line 96: suggestion: IIRC you can `collect` directly into a `String` without having to first create an intermediate `Vec`.

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+                .map(|(i, s)| {
+                    // Incredibly "smart" way to say, append "-" if the number is odd "+"
+                    // otherwise.
+                    if i & 1 != 0 {
+                        s.push('-');
+                    } else {
+                        s.push('+');
+                    }
+                    s.to_owned()
+                })
```
arxanas, line 95: issue: I think this leaves a `+` or `-` at the end of the full string, which is probably not intended? I assumed the `+`/`-`s should only be between tree IDs.

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+                .collect_vec();
+            let mut obfuscated: String = ids.concat();
+            // `PATH_MAX` could be a problem for different operating systems, so truncate
+            // it.
+            if obfuscated.len() >= 255 {
+                obfuscated.truncate(200);
+            }
+            obfuscated
```
arxanas, line 103: issue: I would rather just have the filesystem operation fail, rather than silently truncate the full path. In particular, it would probably be possible to add more sides to a conflict and then have two unrelated trees get assigned the same path on disk.

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+        }
+    }
+}
+
+fn get_runtime(jobs: usize) -> tokio::runtime::Handle {
+    let mut builder = Builder::new_multi_thread();
+    if cfg!(watchman) {
+        // Watchman requires a multithreaded runtime, so just reuse it.
+        return runtime::Handle::current();
+    }
```
arxanas, line 113: issue: I would delete this, and not bother trying to coordinate the runtime with the Watchman subsystem. We don't want to have the code here silently depend on invariants of the code elsewhere. If it's a real problem, we'll have the caller pass in the runtime.
issue: I don't think this instantiates the runtime with the right number of jobs anyways.

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+    if jobs == 1 {
+        builder.max_blocking_threads(1);
+    } else {
+        builder.max_blocking_threads(jobs);
+    }
```
arxanas, line 118:
```suggestion
    builder.max_blocking_threads(jobs);
```

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+    let rt = builder.build().unwrap();
+    rt.handle().clone()
+}
+
+/// A commit stored under `.jj/run/`
+// TODO: Create a caching backend, which creates these on a dedicated thread or
+// threadpool.
+struct StoredCommit {
+    /// Obfuscated name for an easier lookup. If a tree/directory its not set
```
arxanas, line 127: question: Easier lookup by whom, for what purpose?

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+    name: Option<String>,
+    /// The respective commit unmodified.
+    commit: Commit,
+    /// The output directory of the commit, contains stdout and stderr for it
+    output_dir: PathBuf,
+    /// Self-explanatory
+    working_copy_dir: PathBuf,
+    state_dir: PathBuf,
```
arxanas, line 135: question: What's the difference between `state_dir` and `output_dir`?

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+fn create_output_files(id: CommitId, path: &PathBuf) -> Result<(File, File), io::Error> {
+    let stdout_path = path.join("output").join(format!("stdout.{id}"));
+    let stderr_path = path.join("output").join(format!("stderr.{id}"));
+    let mut file_options = OpenOptions::new();
+    let stdout = file_options.write(true).open(stdout_path)?;
+    let stderr = file_options.write(true).open(stderr_path)?;
+    Ok((stdout, stderr))
+}
+
+fn create_working_copies(
```
arxanas, line 173: suggestion: I would not preemptively create the working-copies; instead, I would invoke a function to create a working-copy on demand as part of running the job itself.

- We may as well benefit from the parallelization that `run` itself would use, rather than having to write it in two places.
- Extensions to `run` might make it so that we don't actually use all the working copies we allocated up-front.

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+            stdout,
+            stderr,
+        );
+        results.push(stored_commit);
+    }
+    Ok(results)
+}
+
+/// The result of a single command invocation in `run_inner`.
+struct RunJobResult {
```
arxanas, line 201: suggestion (non-blocking): Try to name this without using `Result`, since it it's not a `Result`.

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+            state_dir,
+            stdout,
+            stderr,
+        );
+        results.push(stored_commit);
+    }
+    Ok(results)
+}
+
+/// The result of a single command invocation in `run_inner`.
```
arxanas, line 200: suggestion: Don't refer to `run_inner` in this comment as it's primarily an implementation detail. You can say something like that it's the result of running the provided command on a single commit.

cli/src/commands/run.rs
arxanas: question: Under what circumstance is `new_tree` different from the tree for the `rewritten_id` commit?

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+
+    while let Some(res) = command_futures.join_next().await {
+        let done = res?;
+        sender.send(done?);
+    }
+    Ok(())
+}
+
+/// Rewrite a single `StoredCommit`.
+async fn rewrite_commit<'a>(
```
arxanas, line 231: issue: This doesn't actually rewrite any commits, so it shouldn't be called `rewrite_commit`.

cli/src/commands/run.rs
```diff
@@ -14,12 +14,262 @@
...
+        builder.max_blocking_threads(jobs);
+    }
+    let rt = builder.build().unwrap();
+    rt.handle().clone()
+}
+
+/// A commit stored under `.jj/run/`
+// TODO: Create a caching backend, which creates these on a dedicated thread or
+// threadpool.
+struct StoredCommit {
```
arxanas, line 126: question: In what sense is this a "stored" commit? All commits are stored. The name/comment don't explain the intention of this struct, just a detail about that it's persistent in some fashion.

cli/src/commands/run.rs
```diff
@@ -53,17 +303,84 @@ pub fn cmd_run(ui: &mut Ui, command: &CommandHelper, args: &RunArgs) -> Result<(
...
+    let resolved_commits: Vec<_> = workspace_command
+        .parse_revset(&args.revisions)?
+        .evaluate_to_commits()?
         .try_collect()?;
     // Jobs are resolved in this order:
     // 1. Commandline argument iff > 0.
     // 2. the amount of cores available.
     // 3. a single job, if all of the above fails.
-    let _jobs = match args.jobs {
+    let jobs = match args.jobs {
```
arxanas, line 314: issue: Isn't the convention that `0` means "use all available parallelism"? I think we should do that instead of producing an error here.

cli/src/commands/run.rs
arxanas: It seems fine; as opposed to what?

cli/src/commands/run.rs
PhilipMetzger: output is for other artifacts which aren't associated with the commit, e.g `stdout/stderr` and other files if necessary.

cli/src/commands/run.rs
PhilipMetzger: We need to replace the commits stored under `.jj/run/default/` if we have commits > jobs and that in-place. So the canonical commit is the initially created one, while those who overwrite it won't have a name. If that makes sense.

cli/src/commands/run.rs
PhilipMetzger: Ack, but let's first get it running somewhat.

cli/src/commands/run.rs
PhilipMetzger: it does rewrite the commits tree and I want to rewrite it in there if possible.

cli/src/commands/run.rs
PhilipMetzger: It never will be different, but to make my question a bit preciser should the caller or the callee do the actual "rewrite" of the commits tree?

arxanas: I think it makes sense for the callee/result of the individual job to construct the new tree for an individual commit (considering that working-copy snapshotting happens as part of most commands anyways, it's probably not surprising or too expensive). And then the caller can choose to rewrite the commit structure with the new trees, if so desired.

For similar organizational purposes, I think the callee should be responsible for procuring the working copy to operate on.

PhilipMetzger: Ok, I've removed the `rewritten_id` field, so the caller is responsible for creating it.

cli/src/commands/run.rs
PhilipMetzger: removed.

cli/src/commands/run.rs
PhilipMetzger: done

cli/src/commands/run.rs
PhilipMetzger: renamed to `RunJob`.

cli/src/commands/run.rs
PhilipMetzger: done

cli/src/commands/run.rs
PhilipMetzger: Renamed to `OnDiskCommit`, maybe that is better?

cli/src/commands/run.rs
PhilipMetzger: See #4149

PhilipMetzger:
This builds now, but fails when invoking the command with a "failed to write, no such fie or directory" somewhere.

cli/src/commands/run.rs
PhilipMetzger: > issue: I think this leaves a `+` or `-` at the end of the full string, which is probably not intended?

This is true after some local testing and should be something to avoid,

cli/src/commands/run.rs
PhilipMetzger: I'll leave it as that since that was its explained above.

cli/src/commands/run.rs
PhilipMetzger: This works for *Nix based systems but for Windows where [this option](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later) is not set, so it'll make the command unusable. So maybe there's a better solution?

cli/src/commands/run.rs
PhilipMetzger: I'd still defer this unitl some work from @hooper has happened.

cli/src/commands/run.rs
PhilipMetzger: imo, it does the core part now, so the name could stay.

cli/src/commands/run.rs
PhilipMetzger: any other variant, since we want to integrate this later with [superconsole](../../../../facebookincubator/superconsole)?

PhilipMetzger (commented):
I've added some replies since @hooper brought this just up in the Discord.

cli/src/commands/run.rs
hooper: A lot of this will shake out when we devise the abstractions for varying parallelism, execution order, caching, etc. This PR is probably too early to put much effort into that.

I think working copy creation will ultimately be triggered by the part that determines parallelism and execution order, so that we can hide the latency sometimes while also deferring the work until it is almost needed. Having it in the inner loop may make sense until then, although I wonder if the parallel access pattern would make it slower.

PhilipMetzger: > I think working copy creation will ultimately be triggered by the part that determines parallelism and execution order, so that we can hide the latency sometimes while also deferring the work until it is almost needed.

+1 from me for the general plan. 

> Having it in the inner loop may make sense until then, although I wonder if the parallel access pattern would make it slower.

I think a fan-out option makes sense if everything is accessed through the network, this also wouldn't hurt the planned on-disk implementation as much.

cli/src/commands/run.rs
PhilipMetzger: > One of the things I would like to see here is an extension point for custom output/status aggregation.

Ok, I'll add a TODO. I assume this is meant to stream the output potentially across the network, if I understand you correctly?

cli/src/commands/run.rs
arxanas: Why not write it the conventional way so that we can get rid of the comment?

PhilipMetzger: done

cli/src/commands/run.rs
PhilipMetzger: done

cli/src/commands/run.rs
PhilipMetzger: This path is now only hit if we have multiple sides, so for simple commits we should only get hex ids. I'm not sure if that's correct though.

cli/src/commands/run.rs
PhilipMetzger: I commented out the code for now. If that is something we want, I'll remove it.

PhilipMetzger:
This is the state from the [Discord message][dc], PTAL. Its just missing tests to land in the February release. 


[dc]: https://discord.com/channels/968932220549103686/969291218347524238/1309224061464743978

cli/src/commands/run.rs
```diff
@@ -14,15 +14,375 @@
...
+    // ...
+    // ```
+    tracing::debug!(
+        "trying to run {shell_command} on commit {id}",
+        id = stored_commit.commit.id().hex(),
+        shell_command = shell_command.as_str()
+    );
+    let mut command = tokio::process::Command::new(get_shell_executable())
+        // TODO: This is not portable as Windows expects a "/c" instead of a "-c".
+        .arg("-c")
```
PhilipMetzger, line 324: I consider this a problem.

cli/src/commands/run.rs
PhilipMetzger: I've added the TODO.

cli/src/commands/run.rs
```diff
@@ -14,15 +14,375 @@
...
+                ConflictMarkerStyle::Snapshot,
+                commit.clone(),
+                shell_command.clone(),
+            ),
+            handle,
+        );
+    }
+
+    while let Some(res) = command_futures.join_next().await {
+        let done = res.unwrap().expect("should not fail joining a job");
```
PhilipMetzger, line 286: I ran into a panic here when testing locally with `jj run 'sed/some/string/g <file>'`, as the last file was unaffected which returned a exitcode of 255 (2).

PhilipMetzger (commented):
Some minor discussion points.

cli/tests/test_run_command.rs
```diff
@@ -0,0 +1,87 @@
...
+// limitations under the License.
+//
+
+use std::path::PathBuf;
+
+use jj_lib::repo_path;
+
+use crate::common::TestEnvironment;
+
+/// Sets up a repo with three initial commits, containing some text.
```
martinvonz, line 22: nit: description seems stale (it only creates a single commit - for the working copy)

cli/tests/test_run_command.rs
```diff
@@ -0,0 +1,87 @@
...
+    let (stdout, stderr) = test_env.jj_cmd_ok(
+        &repo_path,
+        &[
+            "run",
+            "'fake-formatter --uppercase'",
+            "-r",
+            "'mutable()..@'",
+        ],
+    );
+    // all commits should be modified
```
martinvonz, line 47: this doesn't seem to test that

cli/tests/test_run_command.rs
PhilipMetzger: it's adjusted now.

cli/src/commands/run.rs
PhilipMetzger: also adjusted, but the tests still fail.

cli/src/commands/run.rs
PhilipMetzger: its deleted now.

cli/src/commands/run.rs
PhilipMetzger: it is now done in a cross-platform way.

ilyagr (commented):
I haven't thought about the tests or the code yet, but I tried using it and - lo and behold - it works! It rebases cleanly on top of main (except for the changelog)! I can now rename a file in a stack! 🎉 

I did make a bunch of very minor corrections that might be easiest to show as a patch, they are pretty self-evident (except for the first one, that simply reverts the change to the CHANGELOG). Some of them are clippy-related, which might be a bit early, but I thought I'd share them anyway. Others fix some simple thing I discovered while trying to use this version of `jj run`.

https://github.com/jj-vcs/jj/commit/ig-pr4021

You can pull the branch or use

https://github.com/jj-vcs/jj/commit/ig-pr4021.diff

cli/src/commands/run.rs
```diff
@@ -14,15 +14,376 @@
...
+    tracing::debug!(
+        "trying to run {shell_command} on commit {id}",
+        id = stored_commit.commit.id().hex(),
+        shell_command = shell_command.as_str()
+    );
+    let (prog, first_arg) = get_shell_executable_with_first_arg();
+    let mut command = tokio::process::Command::new(prog)
+        .arg(first_arg)
+        .arg(shell_command.as_str())
+        // set cwd to the working copy directory.
```
ilyagr, line 327: (Possibly for later)

For `jj run` command, it is surprising to me that commands are run in the working copy directory instead of the subdirectory I am in. For example, I think it's natural to be in the `cli/src` directory and to run `jj run "mv config.rs somethingelse.rs" -r ...`.

I think this should be governed by an option to `jj run`, and the default should be to run the command in the subdirectory. Until this is implemented, you could make `jj run` vocal about the fact that it's running the commands in the working copy dir.

PhilipMetzger: I don't think its for discussion here, as it out of scope for this MVP. Please add it either to #1869 or #3808

PhilipMetzger:
> I did make a bunch of very minor corrections that might be easiest to show as a patch, they are pretty self-evident (except for the first one, that simply reverts the change to the CHANGELOG). Some of them are clippy-related, which might be a bit early, but I thought I'd share them anyway. Others fix some simple thing I discovered while trying to use this version of `jj run`.

I already fixed those clippy lints locally, but thanks :-). 

> The same happens with `mkdir -p qq`, though I think either should work (if there is no `qq`).
> 
> **Update:** It might have little to do with the command, it might just have to do with something else. Maybe revisions with different commit ids but the same change ids as revisions the commmand was run before? I should probably take a look at the tests :)
> 
> In any case, ` cargo run -- run 'true' -r '@|@-'` results in the same error.
> 
> **Update 2:** A possible reason is that commit `@` was empty. Running `rm -r .jj/run/` seems to be a workaround.

^, is the error I face with the tests.

ilyagr:
> I already fixed those clippy lints locally, but thanks :-).

Oh, you can update the PR then. Did you also get these parts? They jumped at me when I tried to use `jj run`. (I also found one more I just added)

```diff
diff --git a/cli/src/commands/run.rs b/cli/src/commands/run.rs
index 9ccb313eba..c184cc2a4e 100644
--- a/cli/src/commands/run.rs
+++ b/cli/src/commands/run.rs
@@ -361,7 +361,7 @@
     let (dirty, _) = tree_state.snapshot(&options).unwrap();
     if !dirty {
         tracing::debug!(
-            "commit {} was not modified as the passed command did nothing",
+            "commit {} was not modified as the passed command did not modify any tracked files",
             commit.id()
         );
     }
@@ -404,7 +404,7 @@
     shell_command: String,
     /// The revisions to change.
     #[arg(long, short, default_value = "@", value_name = "REVSETS")]
-    revisions: RevisionArg,
+    revisions: Vec<RevisionArg>,
     /// A no-op option to match the interface of `git rebase -x`.
     #[arg(short = 'x', hide = true)]
     exec: bool,
@@ -417,7 +417,7 @@
     let mut workspace_command = command.workspace_helper(ui)?;
     // The commits are already returned in reverse topological order.
     let resolved_commits: Vec<_> = workspace_command
-        .parse_revset(ui, &args.revisions)?
+        .parse_union_revsets(ui, &args.revisions)?
         .evaluate_to_commits()?
         .try_collect()?;
     // Jobs are resolved in this order:
@@ -482,7 +482,7 @@
         // Yeet everything, caching is better implemented in a follow-up.
         fs::remove_dir_all(&run_path)?;
 
-        println!("No commits were rewritten as the command did nothing");
+        println!("No commits were rewritten as the command did not modify any tracked files");
         tx.finish(
             ui,
             format!("run: No-op on {visited} commits with {shell_command}"),
```

cli/src/commands/run.rs
PhilipMetzger: This TODO is related to this request from @matts1: 
> No mention was made about the working directory, so I'm curious how relative paths would work
IMO, we should set the $JJ_REPO environment variable so that a relative path to the root of the repo can be $JJ_REPO/foo
Maintaining relative paths has value
A unix mount namespace could solve this by literally just overlaying our new virtual file system on top of the real one
But unclear how this'd work on non-unix systems

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* A minimal version of `jj run` was implemented.
+  Follow [#1869](https://github.com/jj-vcs/issues/1869) for further
```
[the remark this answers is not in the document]
PhilipMetzger: done

nightscape:
I have been needing this quite frequently. @PhilipMetzger thank you for your efforts!
Could you give a high-level overview over what is missing for an initial version?

PhilipMetzger:
This is missing functional tests to land, after that there will be a bunch of "make it nice" work and then it'll move to the library with some abstracted traits so we can plug a Remote Execution (a la Bazel/Buck2) in.

jennings:
With Philip's blessing, I amended and pushed a new revision with these changes:

- Feature changes
  - Default revset is now `reachable(@, mutable())`, which aligns with `jj fix`. I think `@` was a surprising default because if that's what you wanted, you'd just run the command directly without `jj run`.
  - `run` subcommand is now shown in `jj -h` output
  - stdout and stderr from subprocesses is buffered and emitted all at once by the parent process. This allows scripts to output information and see them in the terminal or pipe them somewhere.
  - Execute commands from the same subdirectory of the working copy where `jj run` is run from.

- Implementation changes
  - Run workspaces are now keyed by commit ID instead of tree ID. An empty commit shares a tree ID with its parent, so they were colliding.
  - Close stdin on spawned subprocesses
  - `writeln!(ui.stderr(), ...)` instead of `println!(...)`
  - Tests are now passing, including the one that uses `cargo fmt`.
  - Reworded the commit message to say "minimal implementation", matching the CHANGELOG entry.

I haven't yet gone through open comments in this PR, I'll do that soon, hopefully.

cli/src/commands/run.rs
jennings: I added this. I think it's likely someone will try `jj run 'rustfmt file_in_this_dir.rs'` and expect it to work (nevermind that that particular thing should probably be done with `jj fix`).

jennings:
I _think_ all the comments people brought up have been answered or addressed. Kind of hard to tell because GitHub loses comment threads after a force push :(

jennings:
I think the first improvement I want to make after this is to have a pool of workspaces that are reused for multiple commits, and not deleted between invocations of `jj run`.

That doesn’t seem too hard to do if I only consider the one `jj run` instance, but if someone runs concurrently from two different terminals, there needs to be some kind of cross-process lock.

PhilipMetzger:
> Feature changes
> 
> * Default revset is now `reachable(@, mutable())`, which aligns with `jj fix`. I think `@` was a surprising default because if that's what you wanted, you'd just run the command directly without `jj run`.

If you add that please just copy the settings reading from `fix` in a separate commit. 

> * Reworded the commit message to say "minimal implementation", matching the CHANGELOG entry.

can you please leave my stuff as is when you're saying you're fixing them. I don't see much value in this. 

You can also just squash the commit which makes it visible.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
+    IoError(#[from] io::Error),
+    #[error("failed to create path {} with {}", .0.to_string_lossy(), .1)]
+    PathCreationFailure(PathBuf, io::Error),
+    #[error("failed to load a commits tree")]
+    TreeState(#[from] TreeStateError),
+    #[error(transparent)]
+    Backend(#[from] BackendError),
+    #[error(transparent)]
+    JobFailure(#[from] JoinError),
+}
```
PhilipMetzger, line 86: nit: missing documentation

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
 /// # Example
 ///
+/// ```shell
 /// # Run pre-commit on your local work
-/// $ jj run 'pre-commit run .github/pre-commit.yaml' -r (trunk()..@) -j 4
-///
-/// This allows pre-commit integration and other funny stuff.
+/// $ jj run 'pre-commit run .github/pre-commit.yaml' -j 4
+/// ```
+// TODO: align interface with `jj bisect run --`
```
PhilipMetzger: nit: address this because my local patch already does this.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
     /// The command to run across all selected revisions.
+    // TODO: align with bisect run --
     shell_command: String,
 
     /// The revisions to change.
     #[arg(
         long = "revision",
         short,
-        default_value = "@",
+        default_value = "reachable(@, mutable())",
```
PhilipMetzger: nit: drop and move to a separate commit copying the settings reading from `fix`.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
+    exec: bool,
 
     /// How many processes should run in parallel, uses by default all cores.
     #[arg(long, short)]
     jobs: Option<usize>,
+
+    /// Run the command from the working-copy root in each commit instead of
+    /// from the subdirectory `jj run` was invoked from.
+    #[arg(long)]
+    root: bool,
```
PhilipMetzger, line 396: nit: separate PR please

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
+        .into_iter()
+        .filter(|c| c.id() != &root_id)
+        .collect();
+    if resolved_commits.is_empty() {
+        return Ok(());
+    }
+
+    workspace_command
+        .check_rewritable(resolved_commits.iter().ids())
+        .await?;
```
PhilipMetzger: nit: this should just be the `chec_rewritable` check. Matching the interface of every other command

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
+    // root, which is canonical (per `CommandHelper::cwd` docs).
+    let subdir = if args.root {
+        PathBuf::new()
+    } else {
+        command
+            .cwd()
+            .strip_prefix(workspace_command.workspace_root())
+            .map(Path::to_path_buf)
+            .unwrap_or_default()
+    };
```
PhilipMetzger, line 449: nit: ditto with separate PR

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
+                && let Some(new_tree) = res.new_tree
+            {
+                done_commits.insert(res.old_id.clone());
+                rewritten_commits.insert(res.old_id.clone(), new_tree);
+            }
+            visited += 1;
+        }
+        if visited == stored_len {
+            break;
+        }
```
PhilipMetzger: nit: I don't think you've understood me here, they still should write to their respective stderr/stdin files.

cli/tests/test_run_command.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    insta::assert_snapshot!(
+        work_dir.run_jj(&["file", "list", "-r", "@-"]),
+        @r"
+    ran.txt
+    sub/file.txt
+    [EOF]
+    "
+    );
+}
```
PhilipMetzger, line 330: nit: separate PR

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
+        tracing::debug!(?base_path, "does not exist, so creating it");
+        fs::create_dir_all(&base_path)?;
+    }
+    tracing::debug!(path = ?base_path, "creating working copies in path: ");
+    for commit in commits {
+        // TODO: revisit once caching is implemented. Working-copy directories are
+        // keyed by commit id to keep paths unique; multiple commits in the input
+        // may share a tree (e.g. the working-copy commit immediately after `jj
+        // commit`), so a tree-keyed name would collide.
+        let commit_path = base_path.join(commit.id().hex());
```
PhilipMetzger: I personally don't like this change at all.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,588 @@
...
+    /// Was the tree even modified.
+    dirty: bool,
+    /// Bytes the subprocess wrote to its stdout, captured in full.
+    stdout: Vec<u8>,
+    /// Bytes the subprocess wrote to its stderr, captured in full.
+    stderr: Vec<u8>,
+    /// True if the command wasn't run because the per-commit working directory
+    /// (the subdirectory `jj run` was invoked from) didn't exist in this
+    /// commit's tree.
+    skipped: bool,
```
PhilipMetzger, line 188: nit: also something for later

cli/src/commands/mod.rs
PhilipMetzger, line 1: nit: squash this commmit

PhilipMetzger (commented):
As the original author I really don't like the changes.

cli/src/commands/run.rs
jennings: For what reason? I couldn't find the conversation about it to confirm, but I think this may have been the root cause for the `stdout`/`stderr` file contention that tests were running into.

jennings:
I ended up making @arxanas's suggestion to create working copies on-demand rather than all up front. This removed the need for the OnDiskCommit struct.

cli/src/commands/run.rs
jennings: It's already done and I think we want this behavior, don't we? EDIT: Oh are you just asking to split the working directory behavior to a new commit?

cli/src/commands/run.rs
jennings: I did, but I think it makes more sense to emit stdout and stderr to the parent process's stdout/stderr by default. My reasoning:

- It's what `git rebase` does, so people may expect it. I know "git does it" isn't a good reason to do anything, but "people expect it" is.
- I think it's more useful than writing to files buried in `.jj/run/default/*` that get deleted on a successful run.
- If the user wants to write them to files instead, they can add `> path/to/stdout 2> path/to/stderr` in their command.

cli/src/commands/run.rs
jennings: Yeah you're probably right. I kind of thought about it later but forgot to come back to it.

cli/src/commands/run.rs
PhilipMetzger: > I did, but I think it makes more sense to emit stdout and stderr to the parent process's stdout/stderr by default. My reasoning:
> 
> * It's what `git rebase` does, so people _may_ expect it. I know "git does it" isn't a good reason to do anything, but "people expect it" is.

Sorry I don't think "people expect it" in relation to a new command which loosely relates to a `git rebase -x` makes sense here. I'm quite sure that if people want that opt-in we should provide it, just not now. 

> * I think it's more useful than writing to files buried in `.jj/run/default/*` that get deleted on a successful run.

You know this was temporary for https://github.com/jj-vcs/jj/pull/4021#issuecomment-4571824408 and to cut down the size of this. Its a better experience if we can provide the failing `stdout/stderr` if the command fails on an individual commit. 

> * If the user wants to write them to files instead, they can add `> path/to/stdout 2> path/to/stderr` in their command.

I don't want to support this currently. If you open FR for it we can support that via a flag but its also out of scope for this PR.

cli/src/commands/run.rs
PhilipMetzger: > It's already done and I think we want this behavior, don't we? EDIT: Oh are you just asking to split the working directory behavior to a new commit?

No I'm just saying that is also a separate PR, for some reason you constantly choose to blow the scope of this instead of trying the minimal thing.

cli/src/commands/run.rs
PhilipMetzger: > For what reason?

It allows users to introspect the commits as soon as they stay permanently on-disk which is not something I want to support at all. 

> I couldn't find the conversation about it to confirm, but I think this may have been the root cause for the `stdout`/`stderr` file contention that tests were running into.

Again the issue I ran into is that two separate `run` processes started by `nextest` wanted to create a file which potentially already existed on disk. The naming scheme had no influence on this.

cli/src/commands/run.rs
martinvonz: > I think it makes more sense to emit stdout and stderr to the parent process's stdout/stderr by default.

Since `jj run` will allow the command to be run on multiple commits concurrently, that would result in interleaved output from many commands, right? I think we want to capture the output and present it to the user in a more structured way, though I don't know what that would be. @hooper: what does `hg run` do? Hmm, looking at the (internal) code, it actually looks like it interleaves the output. Is that correct? I know you made the `hg blaze` command print output in a nice way. Have you thought about how a generic `jj run` should present the output?

cli/src/commands/run.rs
jennings: I think we may as well come out of the gate with a useful revset as a default. @ isn't useful because you don't need `jj run` for that.

cli/src/commands/run.rs
jennings: > Since jj run will allow the command to be run on multiple commits concurrently, that would result in interleaved output from many commands, right?

This PR does ensure they're not interleaved, I did essentially what `jj fix` does with stderr: Buffer the stdout/stderr from each process, then as they complete they have to acquire a lock on `ui.stdout()`/`ui.stderr()` to emit it all at once.

My thought is that if the user wants structured data, they'll output one line per commit, then pipe it to something that can parse that. For example, to figure out which commits have passing tests:

```command
$ jj run -- bash -c 'cargo test >/dev/null;echo "$JJ_COMMIT_ID,$?"'
aaaaa,0
bbbbb,1
ccccc,0
```

Or, their command can write structured data to files elsewhere in the system, then they'll know where to read it from after `jj run` exits.

cli/src/commands/run.rs
PhilipMetzger: > I think we may as well come out of the gate with a useful revset as a default. @ isn't useful because you don't need `jj run` for that.

I will just add that commit here, since you want that. 

Edit: And requiring that people need to use `-r <revset>` to start out is not bad.

cli/src/commands/run.rs
jennings: The `--root` flag and `skipped` concept have been extracted to a separate commit on top of the minimal implementation commit.

cli/src/commands/run.rs
jennings: Done.

cli/src/commands/run.rs
PhilipMetzger: I've pushed the commit and deleted the respective change from you there.

cli/src/commands/run.rs
PhilipMetzger, line 1: > So each command no longer requires a the `--revisions` by default.

Drop the "a"

CHANGELOG.md
```diff
@@ -21,6 +16,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* A minimal version of `jj run` was implemented.
+  Follow [#1869](https://github.com/jj-vcs/jj/issues/1869) for further
+  improvements.
```
thoughtpolice: The tense here is weird, nobody necessarily knows a priori what `run` is and I wouldn't bother mentioning a discussion issue tbh. I'd suggest:

> `jj run` allows you to run a command over a set of changes, each with their own private working copy; the commands may update the working copy and changes/conflicts are propagated accordingly e.g. `jj run 'cargo check --all-features'` or `jj run 'cargo fix'` behaves as one might expect.

Also, move this into the release highlights IMO.

cli/Cargo.toml
```diff
@@ -103,6 +103,7 @@ shlex = { workspace = true }
 slab = { workspace = true }
 tempfile = { workspace = true }
 textwrap = { workspace = true }
+tokio = { workspace = true, features = ["fs", "process", "rt-multi-thread", "sync"] }
```
thoughtpolice: Put this in the main `Cargo.toml` workspace and then just inherit that feature set, please. We've let this slip a few times in `lib/` and should probably audit divergent feature flags like this, but AFAICS there's no need for it here.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,573 @@
...
+    spec: Arc<CommandSpec>,
+    base_path: Arc<PathBuf>,
+    commits: Arc<Vec<Commit>>,
+) -> Result<(), RunError> {
+    let base_ignores = tx.base_workspace_helper().base_ignores().unwrap().clone();
+    let mut command_futures = JoinSet::new();
+    for commit in commits.iter() {
+        command_futures.spawn_on(
+            rewrite_commit(
+                // TODO: handle/propagate error here
```
thoughtpolice, line 214: Old comment? This is handled by `join_next()`, is it not?

cli/src/commands/run.rs
```diff
@@ -14,71 +14,573 @@
...
+}
+
+/// A command, its arguments, and the workspace-relative directory it should
+/// run in.
+struct CommandSpec {
+    program: String,
+    args: Vec<String>,
+    /// Working directory for the subprocess, relative to the workspace root.
+    /// Empty means run from the workspace root.
+    subdir: PathBuf,
```
thoughtpolice: small nit: make this an `Option<PathBuf>` and then you don't even need the `is_dir()` check in the `--root` case below.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,573 @@
...
+    // start from a clean slate.
+    if commit_path.exists() {
+        tracing::debug!(
+            dir = ?commit_path,
+            commit = commit.id().hex(),
+            "removing leftover directory from a previous run"
+        );
+        fs::remove_dir_all(&commit_path)
+            .map_err(|e| RunError::PathCreationFailure(commit_path.clone(), e))?;
+    }
```
thoughtpolice, line 116: Is this not just a massive TOCTOU race? If you run `jj run -- slow_cmd` and `jj run -- fast_cmd` over the same set of commits, would the fast command not just obliterate the slow commands' working copies while they're running, causing spurious failures?

cli/src/commands/run.rs
```diff
@@ -14,71 +14,573 @@
...
+    // ```
+    tracing::debug!("trying to run command '{}' on commit {}", spec, commit.id());
+    // Pipe and buffer the subprocess's stdout/stderr so we can emit them
+    // atomically to the parent's stdout/stderr after the process exits. Writing
+    // concurrently from multiple jobs would interleave output.
+    let command = tokio::process::Command::new(&spec.program)
+        .args(&spec.args)
+        .current_dir(&exec_dir)
+        // TODO: relativize
+        // .env("JJ_PATH", working_copy_dir)
```
thoughtpolice: What does this comment mean? FWIW we could maybe get rid of it because you can just run a command that gives you `$CWD` in a shell or something?

thoughtpolice (commented):
I have more nits here but we should really just get this over the line and put it into users hands at this point, IMO.

My comment below about the working copy races is the most important one. Maybe I misread something and it's a non-issue, but otherwise we definitively need to fix that.

cli/src/commands/run.rs
PhilipMetzger: This was a TODO left over for this: 

> No mention was made about the working directory, so I'm curious how relative paths would work
IMO, we should set the $JJ_REPO environment variable so that a relative path to the root of the repo can be $JJ_REPO/foo
Maintaining relative paths has value
A unix mount namespace could solve this by literally just overlaying our new virtual file system on top of the real one
But unclear how this'd work on non-unix systems

> This is another great point, which never was brought up. I agree that the $JJ_REPO env var should be included from the beginning.

This from #1869 which originally was https://github.com/jj-vcs/jj/issues/3577#issuecomment-2113627939. This is a user request which we can punt to later.

cli/src/commands/run.rs
PhilipMetzger: it should be moved to line 216

cli/src/commands/run.rs
PhilipMetzger, line 1: nit: missing changelog entry update

PhilipMetzger, line 1: nit: also needs a changelog update and I missed the `config-schema.json` update here

PhilipMetzger (commented):
minor comments

cli/src/commands/run.rs
jennings: I don't think it's clear what "handle/propagate error" means we want to eventually do, so if _anything_ I'd rather have a comment explaining the problem with the current approach. But I don't know what that problem is, we seem to behave sensibly right now: If a command returns nonzero, we abort the whole transaction. I'll  remove the comment unless we can specify what we'd like to do instead?

PhilipMetzger: > I don't think it's clear what "handle/propagate error" means we want to eventually do, so if _anything_ I'd rather have a comment explaining the problem with the current approach.

This was initially written for cloning the `.gitignores` which may or may not exist when we created the snapshot options and the 64MB snapshot limit which should be read from the correct config. 

> But I don't know what that problem is, we seem to behave sensibly right now: If a command returns nonzero, we abort the whole transaction. I'll remove the comment unless we can specify what we'd like to do instead?

For a `-k --keep-going` flag we will need to ignore individual errors and not stop running processes but that's for later.

cli/src/commands/run.rs
jennings: It would. My follow up "pooled working copies" implementation that I'm working on adds a lockfile at `.jj/run/pool/N/lock` so multiple processes don't stomp on each other, we could probably use that here. What would you like to do?
- Implement something now before merging
- Implement it as a (hopefully) fast follow-up PR

cli/Cargo.toml
jennings: Fixed.

CHANGELOG.md
jennings: Fixed, thanks

cli/src/commands/run.rs
jennings: Changed, thanks.

cli/src/commands/run.rs
PhilipMetzger: > lockfile at `.jj/run/pool/N/lock`

can this be under `.jj/run/default/pool/` instead so its easier to pull out a trait for it?

cli/src/commands/run.rs
jennings: I added the config-schema update, the changelog has been updated.

cli/src/commands/run.rs
jennings: For the current implementation, I would put it in `.jj/run/default/<commit_id>/lock`

(Adding the following for context, not for discussion) My current implementation for pooled working copies creates directories `.jj/run/pool/{1-N}`  with the same layout as `.jj/run/default/<commit_id>` has. You choose which set of directories to use by passing either `--strategy isolated` or `--strategy pool`. So, my thought was that this "default" directory is where `--strategy isolated` does its work and the "pool" directory is where `--strategy pool` does its work.

PhilipMetzger: > For the current implementation, I would put it in `.jj/run/default/<commit_id>/lock`

If Austin wants that, SGTM. 

> (Adding the following for context, not for discussion) My current implementation for pooled working copies creates directories `.jj/run/pool/{1-N}` with the same layout as `.jj/run/default/<commit_id>` has. You choose which set of directories to use by passing either `--strategy isolated` or `--strategy pool`. So, my thought was that this "default" directory is where `--strategy isolated` does its work and the "pool" directory is where `--strategy pool` does its work.

These again are internal implementation details so putting them under `.jj/run/default/{strategy}/{commit-id}` should be fine.

jennings: What does "default" mean in this path? What other values besides "default" might there be?

PhilipMetzger: > What does "default" mean in this path?

It means that this is the default backend for run, since we want to give downstreams a way to extend it. 

> What other values besides "default" might there be?

I want to have a RE one built on https://bazel.build/remote/rbe and another one which is similar to Bazel's hybrid one, which races executions between a RE backend and the local one.

jennings: I think what I was tentatively calling "strategy" is the same concept as the backend you're describing.

- The "isolated" backend creates per-commit fresh working copies, using `.jj/run/default` as its workspace (or `.jj/run/isolated`).
- The "pool" backend reuses working copies between invocations, using `.jj/run/pool` as its workspace.
- A "remote" backend either wouldn't need any local directories at all, or it would use `.jj/run/remote` as its workspace (which would _not_ contain working copies, but instead probably bookkeeping info).
- A "race" backend wouldn't need its own workspace, but would use both `.jj/run/pool` and `.jj/run/remote`.

This makes the `.jj/run/default/{strategy}` path seem more nested than necessary to me.

PhilipMetzger: > I think what I was tentatively calling "strategy" is the same concept as the backend you're describing.
> 
> * The "isolated" backend creates per-commit fresh working copies, using `.jj/run/default` as its workspace (or `.jj/run/isolated`).
> * The "pool" backend reuses working copies between invocations, using `.jj/run/pool` as its workspace.

I mean you're first two strategies or "backends" use the same mechanism, that's why they should stay under `default` where all checked out working copies land. 

> * A "remote" backend either wouldn't need any local directories at all, or it would use `.jj/run/remote` as its workspace (which would _not_ contain working copies, but instead probably bookkeeping info).

A remote just needs a directory and a file with some internals in there. 

> * A "race" backend wouldn't need its own workspace, but would use both `.jj/run/pool` and `.jj/run/remote`.

A "hybrid" backend may choose to materialize when racing so having them having its own directories is a better design, since each backend is responsible for the management of the working copies. 

> This makes the `.jj/run/default/{strategy}` path seem more nested than necessary to me.

See my first point.

thoughtpolice: Worrying about the location of paths like this right now is not relevant. We can cross that bridge when/if any further work on abstracting or pulling out traits is done, not before, this is just splitting hairs over nothing and delaying this PR.

FWIW, I do think it's necessary to fix this. But I would be completely fine if this PR just had a single global lock around a run invocation, so that only one `jj run` can execute at a time per repository. To be clear, that single `run` invocation could be using concurrent threads executing many things in parallel over a large number of commits. But only one `jj run` _command_ can be executing at a time. This would be similar to how `cargo` locking works today and would be enough to stop a massive footgun from happening.

cli/src/commands/run.rs
thoughtpolice: That makes sense. We should probably just include `JJ_PATH` as it is and uncomment this line, then. We can handle any edge cases with relative paths coming up later.

cli/src/commands/run.rs
jennings: I added a single global lock at `.jj/run/default/lock` that is acquired before beginning execution.

cli/src/commands/run.rs
jennings: I added it as `JJ_WORKSPACE_ROOT` to match the env var that `jj util exec` provides. I can change it if that's not a good idea.

hooper:
I don't want to block this PR, so I will just mention a few things that will be important to Google based on our experience with `hg run`:
 - The ability to choose between "run sequentially in the working copy" and "run in parallel in remote/external working copies". We use a VFS to quickly materialize huge monorepo working copies, and they wouldn't be mounted under `.jj/run/`.
    - We also went with literally `--strategy foo` as a UI for this, but I think our users mostly go with the defaults we set for the command and its aliases (imagine something like `hg bazel foo` -> `hg run --strategy=magic_vfs_thing -j10 -- bazel foo` with some special handling for the progress output).
    - We generate these on demand, re-use them when reasonable, and have a cache of them. I'm happy to see consideration for these things already!
    - I guess we'll need a backend for registering custom/extension implementations of this so the user can choose between them. `create_working_copy` doesn't look like it supports that yet.
 - Passing output through to the terminal is a very useful default, and redirecting it to a file would ideally be possible via normal shell scripting techniques, even if there's another mechanism. Buffering/interleaving requires some care, and having a few reasonable behaviors available via flags/configs should be valuable. I just wouldn't want to over-index the implementation or UI on any one use case.
    - I also think -j1 is a good default, partly because it defers the output buffering/interleaving problem until the user asks for it by increasing parallelism. It can also mitigate a tendency for swapping or other kinds of resource exhaustion to be a foot gun.
    - When run on a tty, supporting interactive commands is useful.
 - The transform_descendants aspect is great, and much better than trying to have the user put something like `jj squash` into the command, so thank you for that. However, it should probably be possible to turn it off so the working copies can be treated as throwaway sandboxes. Some use cases might benefit a lot from avoiding the cost of checking for changes at the end, especially if caching allows them to avoid the setup cost too.
    - The option to use `--restore-descendants` would be useful for the same reasons `jj fix` always works that way.

Please take all of that as advisory, I apologize that I don't have much time to help with this feature right now.

jennings:
> I guess we'll need a backend for registering custom/extension implementations of this so the user can choose between them. `create_working_copy` doesn't look like it supports that yet.

It's not for this PR, but [I have an implementation](https://github.com/jj-vcs/jj/compare/push-mqxwqtnwnpts...jennings:jj:run-pool?expand=1) that lets you choose between `--strategy isolated` (this PR's behavior) and `--strategy pool` (reusable working copies). I'm hoping to follow up with that after this lands. 

> I also think -j1 is a good default, partly because it defers the output buffering/interleaving problem until the user asks for it by increasing parallelism. It can also mitigate a tendency for swapping or other kinds of resource exhaustion to be a foot gun.

Sounds good to me, I'll make this change, then in a follow-up maybe add a config to override the default.

> However, it should probably be possible to turn [amending] off so the working copies can be treated as throwaway sandboxes.

I think that makes sense. I think we can add `--no-amend` or something like that in a follow-up PR.
   
> The option to use `--restore-descendants` would be useful for the same reasons `jj fix` always works that way.

Agree.

cli/src/commands/run.rs
jennings: Since the current behavior matches hooper’s recommendation, I’m going to resolve this thread. We can revisit it in a new PR if necessary.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,594 @@
...
+
+impl From<RunError> for CommandError {
+    fn from(value: RunError) -> Self {
+        Self::new(CommandErrorKind::Cli, Box::new(value))
+    }
+}
+
+/// Creates the required directories for a StoredWorkingCopy.
+/// Returns a tuple of (`working_copy` and `state`).
+async fn create_working_copy_paths(path: &Path) -> Result<(PathBuf, PathBuf), RunError> {
```
thoughtpolice: i think we can just inline this function into `create_working_copy` tbqh

cli/src/commands/run.rs
```diff
@@ -14,71 +14,594 @@
...
+    tracing::debug!(?working_copy, ?state, "creating paths for a commit");
+
+    fs::create_dir(&working_copy)
+        .map_err(|e| RunError::PathCreationFailure(working_copy.clone(), e))?;
+    fs::create_dir(&state).map_err(|e| RunError::PathCreationFailure(state.clone(), e))?;
+
+    Ok((working_copy, state))
+}
+
+fn get_runtime() -> tokio::runtime::Runtime {
```
thoughtpolice: inline this too, it has only 1 call site.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,594 @@
...
+        .check_out(&commit.tree())
+        .map_err(|_| RunError::FailedCheckout(commit.id().clone()))?;
+
+    Ok((working_copy_dir, tree_state))
+}
+
+/// Compute and ensure the base directory under which each per-commit working
+/// copy lives, then acquire an exclusive lock on it. Returns the base path and
+/// the held lock (drop to release).
+fn ensure_base_path(repo_path: &Path) -> Result<(PathBuf, FileLock), RunError> {
```
thoughtpolice: yeah, let's just inline this too. i think doing so is good because keeping the FileLock closer to where it's actual lifetime is (cmd_run) makes it easier to understand imo. You can also use `drop(lock)` IMO to get rid of the unused variable suppression and make the lock scope a little more clear, IMO.

thoughtpolice (approved):
These are all mostly just readability nits, I leave them to your better judgement if you want to address them. Otherwise I'm happy with how this first version looks; tysm for helping get this over the line Stephen, and Philip for starting it.

thoughtpolice:
Oh, and one more thing: I'd probably just merge the first two commits together (the one that adds the `--root` flag into the OG commit.) But that's just an idle thought, don't put too much stock in it.

cli/src/commands/run.rs
```diff
@@ -14,71 +14,512 @@
...
+                rewrite_commit(base_ignores, base_path, commit, spec).await
+            },
+            handle,
+        );
+    }
+
+    while let Some(res) = command_futures.join_next().await {
+        let done = match res {
+            Ok(rj) => rj?,
+            Err(err) => return Err(RunError::JobFailure(err)),
```
mirkomartn, line 216: It seems this early return (and `should_quit` handling bellow) are fine w.r.t. cancel safety if `rewrite_commit` sets the `tokio::process::Command::kill_on_drop` which is not clear nor guaranteed it will be done. What I'm wondering is whether this invariant could be better enforced from `run_inner` directly, perhaps by constructing `Command` here (`run_inner` is already aware of using tokio so that shouldn't be a reservation)? The worry being that a later refactor might break the invariant.

PR #4026, opened 2024-07-04
workspace: abandon discardable working copy on forget

scott2000:
Sometimes I create a temporary workspace and then forget it when I'm done, but currently it leaves behind a working copy commit which I have to manually abandon. I think it would be more convenient if it was abandoned automatically like `jj` already does for discardable commits in other situations.

While working on this, I also noticed that if two workspaces are editing the same discardable commit, it will be abandoned whenever either workspace's working copy moves (and then it is immediately created again, but with a new change ID). I think that working copies should be treated similarly to branches and should also prevent discardable commits from being abandoned like branches do, so I included that change as well since it seems related. I don't feel strongly about this, so if this change isn't necessary I can remove it from the PR.

I also noticed a typo in the docs for `jj workspace`, so I fixed that as well.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   default branch of the remote as repository settings for
   `revset-aliases."trunk()"`.`
 
+* `jj workspace forget` now abandons the workspace's working-copy commit if it
+  was empty.
+
 ### Fixed bugs
 
 ## [0.19.0] - 2024-07-03
--- a/cli/tests/test_workspaces.rs
+++ b/cli/tests/test_workspaces.rs
@@ -628,14 +628,11 @@ fn test_workspaces_forget() {
     insta::assert_snapshot!(stderr, @"");
 
     // The old working copy doesn't get an "@" in the log output
-    // TODO: We should abandon the empty working copy commit
     // TODO: It seems useful to still have the "secondary@" marker here even though
     // there's only one workspace. We should show it when the command is not run
     // from that workspace.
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   individually instead of being passed a directory by setting
   `merge-tools.$TOOL.diff-invocation-mode="file-by-file"` in config.toml.
 
+* When creating a new working-copy commit, the old commit is no longer
+  abandoned if it is still the working-copy commit of a different workspace.
+
```
yuja, line 23: nit: I wouldn't add this changelog entry as the change isn't visible to user. (Or move it "fixed bugs" section if we think the current behavior is broken.)

yuja (approved):
Thanks!

cli/tests/test_workspaces.rs
```diff
@@ -712,6 +709,65 @@ fn test_workspaces_forget_multi_transaction() {
...
+    "###);
+    insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r###"
+    ◉  57d63245a308 fourth@ second@ third@
+    │ @  4e8f9d2be039 default@
+    ├─╯
+    ◉  000000000000
+    "###);
+
+    // delete the default workspace (should not abandon commit since not empty)
+    test_env.jj_cmd_ok(&main_path, &["workspace", "forget", "default"]);
```
yuja, line 744: nit: can you also add snapshot of `stderr`?

scott2000: Since the `stderr` is empty, I just switched it to use `jj_cmd_success`.

PR #4033, opened 2024-07-05
build: correctly set version hash for merge commits

scott2000:
I noticed a weird issue while trying to run tests with a working-copy merge commit. The build script concatenated the commit ID hashes of all of the merge commit's parents and put them all together in the version string, which caused `test_global_opts::test_version` to fail since it expects a 40-digit hash.

# Checklist

If applicable:

### DIFF
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -63,6 +63,7 @@ fn get_git_hash() -> Option<String> {
             "--color=never",
             "log",
             "--no-graph",
+            "--limit=1",
             "-r=@-",
             "-T=commit_id",
         ])


cli/build.rs
```diff
@@ -63,6 +63,7 @@ fn get_git_hash() -> Option<String> {
             "--color=never",
             "log",
             "--no-graph",
+            "--limit=1",
```
yuja, line 66: nit: I don't think picking arbitrary parent (in index order) is correct, but it's practically okay. We have a similar problem in `git` code path.

@ilyagr might have some opinion. iirc, he had a patch to improve version string.

scott2000: I had also considered using `heads(@ & merges() | @-)` to use `@` instead of `@-` if it's a merge, not sure if that would be any better.

PR #4041, opened 2024-07-06
backout: accept multiple revisions to back out

### LINKED ISSUE #3339: FR: `jj backout` should be able to revert an entire series of revisions
It's useful to occasionally not just backout one commit, but a series that landed and ultimately caused regressions. At the moment, `jj backout` takes exactly one revision, so doing this requires scripting around `jj log` in your shell to get the revisions.

Realistically speaking, this means that `jj backout -r` should take a revset, much like `jj abandon` does.

There are some edge cases to handle I think, because you need to effectively resolve the union of all provided revsets, then for any changes that are determined to have a parent/child relationship, you need to back them out in reverse order. But I don't think this would be too hard, so I'm tagging this with "good first issue" for anyone who wants to try it.

bnjmnt4n:
Closes #3339.

# Checklist

If applicable:
- [NA] I have updated the documentation (README.md, docs/, demos/)
- [NA] I have updated the config schema (cli/src/config-schema.json)
-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj rebase --skip-empty` has been renamed to `jj rebase --skip-emptied`
 
+* `jj backout --revision` has been renamed to `jj backout --revisions`.
+  The short alias `-r` is still supported.
+
 ### Deprecations
 
 ### New features
@@ -32,6 +35,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj backout` now includes the backed out commit's subject in the new commit
   message.
+
+* `jj backout` can now back out multiple commits at once.
 
 ### Fixed bugs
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj rebase --skip-empty` has been renamed to `jj rebase --skip-emptied`
 
+* `jj backout --revision` has been renamed to `jj backout --revisions`.
+  The short alias `-r` is still supported.
```
yuja, line 16: nit: might be better to leave flag name unchanged or aliased to singular form.

(This applies to various commands, and can be addressed later.)

yuja (approved):
Nice.

CHANGELOG.md
bnjmnt4n: Right, I changed this for rebase as well. I remember there were at least some other commands which used `--revisions`. It seemed kind of inconsistent for now, which is why I changed it.

bnjmnt4n: I will merge this first. Perhaps we can change the flag name later.

Issue #4031, opened 2024-07-04
Segfault in jj log with 4000+ revisions


tim-janik:
## Description

I'm using `jj log` with an fzf query string. Since jj log cannot grep for a query string in commits as of 0.18.0, I'm using `git log --format=%h -P --grep` for now. For wildcards in large repositories, this can yield 4000+ revisions, triggering a segfault in jj log.

## Steps to Reproduce the Problem

Enter a repository with 4000+ commits.

```
jj log -r $(GIT_DIR=$(jj root)/.jj/repo/store/git git log --all --format=%h --grep . | tr '\n' "|")'none()'
Segmentation fault
jj log -r $(GIT_DIR=$(jj root)/.jj/repo/store/git git log --all --format=%h --max-count 3500 --grep . | tr '\n' "|")'none()'
◉  uznpwzlt benjamin@dev.ofcr.se 2024-07-04 17:23:59 bnjmnt4n/push-uzynoulwtlvz@origin 5445df74
│  cli: add `jj operation show` command
◉  qztwlksp benjamin@dev.ofcr.se 2024-07-04 17:23:59 fe917b22
│  cli: add `jj operation diff` command
◉  qwwqyoxs benjamin@dev.ofcr.se 2024-07-04 17:23:42 541c031c
│  repo: add `RepoLoader::merge_operations`
```

## Expected Behavior

The` --max-count 3500` shouldn't be needed. Looking at the length of the revision string, I suspect there is a 32k limit in revision expression evaluation somewhere: 

```
GIT_DIR=$(jj root)/.jj/repo/store/git git log --all --format=%h --max-count 3585 --grep . | tr '\n' "|" | wc -c
32265
```

Ideally, revision strings of arbitrary length should be supported. And could be loaded from a file ;-)

## Actual Behavior

Unless --max-count  is used, wildcards yielding long revision lead to a segfault in e.g.:

```
fzf -q GREP-PATTERN --disabled --ansi  \
  --bind 'change:reload( jj --no-pager log --color=always -T builtin_log_oneline -r $(GIT_DIR=$(jj root)/.jj/repo/store/git git log --all --format=%h --grep {q} | tr "\n" "|")"none()" )' \
  --preview "[[ {} =~ ^[^abc-xyz()]*([k-xyz]{8,})\  ]] && jj --ignore-working-copy show --color=always  \${BASH_REMATCH[1]}" \
  </dev/null
```

## Specifications

[...]

PR #4045, opened 2024-07-07
fileset, revset: flatten union nodes in AST to save recursion stack

yuja:
#4031 

# Checklist

If applicable:

### DIFF
--- a/lib/src/fileset.rs
+++ b/lib/src/fileset.rs
@@ -17,6 +17,7 @@
 use std::collections::HashMap;
 use std::{iter, path, slice};
 
+use itertools::Itertools as _;
 use once_cell::sync::Lazy;
 use thiserror::Error;
 
@@ -227,18 +228,6 @@ impl FilesetExpression {
         FilesetExpression::Pattern(FilePattern::PrefixPath(path))
     }
 
-    /// Expression that matches either `self` or `other` (or both).
-    pub fn union(self, other: Self) -> Self {
-        match self {
-            // Micro optimization for "x | y | z"
-            FilesetExpression::UnionAll(mut expressions) => {
-                expressions.push(other);
-                FilesetExpression::UnionAll(expressions)
-            }
-            expr => FilesetExpression::UnionAll(vec![expr, other]),
-        }
-    }
-
     /// Expression that matches any of the given `expressions`.
     pub fn union_all(expressions: Vec<FilesetExpression>) -> Self {
         match expressions.len() {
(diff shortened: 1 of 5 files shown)


lib/src/fileset.rs
```diff
@@ -227,18 +228,6 @@ impl FilesetExpression {
...
-        match self {
-            // Micro optimization for "x | y | z"
-            FilesetExpression::UnionAll(mut expressions) => {
-                expressions.push(other);
-                FilesetExpression::UnionAll(expressions)
-            }
-            expr => FilesetExpression::UnionAll(vec![expr, other]),
-        }
-    }
-
```
emilazy, line 241: It might be nice to keep this helper function around so that you can still `files1.union(files2)`?

yuja: We can reintroduce a simpler version if needed?

I personally prefer `union_all([x, y])` than method chain `x.union(y)` as `x` can also be a multi-line chained expression.

emilazy: Seems reasonable; I was thinking that `.union()` is a closer match for the infix syntax used in the expression language, but the only uses in my `.mailmap` changes are actually prefix `RevsetExpression::union(a, b)`, so they can be migrated over without any stylistic change.

What do you think about changing `union_all` to take `impl Into<Vec<FilesetExpression>>` so you can save some syntactic noise by writing `RevsetExpression::union_all([a, b])` rather than `RevsetExpression::union_all(vec![a, b])`?

emilazy: Oh, I just noticed that this doesn’t remove or change `RevsetExpression::union` at all – is that intentional?

yuja: > Oh, I just noticed that this doesn’t remove or change `RevsetExpression::union` at all – is that intentional?

Yes. `RevsetExpression::Union` isn't migrated to `Vec<RevsetExpression>` yet, so the underlying representation is still binary form.

emilazy: Makes sense, sorry for the noise then!

jonathantanmy (approved):
It would be nice if instead of Binary(LHS, RHS) and UnionAll expressions, we only had Binary(Vec). But I tried doing that and it looks like that would change a lot of the existing code (the current code uses the fact that `FilesetExpression::union_all` exists, but not `FilesetExpression::intersection_all` etc.). So this looks good to me. We can change it to Binary(Vec) later if needed.

yuja:
`Binary(Vec)` doesn't always make sense. `(a::b)::c` could be transformed to `(:: a b c)`, but it doesn't look nicer than `(:: (:: a b) c)`. I think union (and maybe intersection) are special.

PR #4046, opened 2024-07-07
cli_util: short-prefixes for commit summary in transaction

scott2000:
I've run into change ID prefixes being 4-5 characters instead of the usual 1-2 characters in commands like `jj duplicate` and `jj split` fairly often, and it seems like this should resolve that. I tried to avoid creating a new `IdPrefixContext` unless changes have already been made in the transaction, since several commands call `tx.format_commit_summary()` before making any changes.

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1645,8 +1645,10 @@ impl WorkspaceCommandTransaction<'_> {
         formatter: &mut dyn Formatter,
         commit: &Commit,
     ) -> std::io::Result<()> {
-        // TODO: Use the disambiguation revset
-        let id_prefix_context = IdPrefixContext::new(self.helper.revset_extensions.clone());
+        let id_prefix_context = self
+            .helper
+            .new_id_prefix_context()
+            .expect("parse error should be confined by WorkspaceCommandHelper::new()");
         let language = CommitTemplateLanguage::new(
             self.tx.repo(),
             self.helper.workspace_id(),
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -999,25 +999,28 @@ impl WorkspaceCommandHelper {
         )
     }
 
+    fn new_id_prefix_context(&self) -> Result<IdPrefixContext, CommandError> {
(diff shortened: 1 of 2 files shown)


cli/src/cli_util.rs
```diff
@@ -1642,8 +1645,15 @@ impl WorkspaceCommandTransaction<'_> {
         formatter: &mut dyn Formatter,
         commit: &Commit,
     ) -> std::io::Result<()> {
-        // TODO: Use the disambiguation revset
-        let id_prefix_context = IdPrefixContext::new(self.helper.revset_extensions.clone());
+        // Use helper's cached IdPrefixContext if no changes made to repo
+        if !self.tx.repo().has_changes() {
+            return self.helper.write_commit_summary(formatter, commit);
```
yuja, line 1650: nit: this might be good, but appears not covered by any tests. Maybe we can go without optimized path until we find a performance problem?

yuja (approved):
Thanks. For future reference, can you copy PR description to commit message?

cli/src/cli_util.rs
scott2000: That's a good point, I hadn't noticed there weren't any existing tests covering it. The only place this seems to happen is when creating "instructions" messages for interactive editors. The main case that seems concerning to me from a performance perspective is that if somebody squashes N commits, it will unconditionally rebuild the `IdPrefixContext` N*2 times, which seems like it could add up. Maybe these should be directly calling the helper's `format_commit_summary()` instead of the transaction's?

yuja: Yes, they can bypass tx/mut_repo if the performance really matters.

BTW, they can also bypass instruction formatting at all if not interactive. If it's interactive session, the cost of short-prefixes evaluation would be negligible anyway.

scott2000: > BTW, they can also bypass instruction formatting at all if not interactive. If it's interactive session, the cost of short-prefixes evaluation would be negligible anyway.

That sounds like a better solution to me, so I'll remove this check for now and leave that change for a future PR.

PR #4050, opened 2024-07-08
readme: include link to wiki "Media" page and fix broken changelog link

bnjmnt4n:
# Checklist

If applicable:
- [NA] I have updated `CHANGELOG.md`
- [NA] I have updated the config schema (cli/src/config-schema.json)
- [NA] I have added tests to cover my changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ developers monitor both channels.
 
 ### News and Updates 📣
 
-- **Feb 2024**: Version 0.14 is released, which deprecates ["jj checkout" and "jj merge"](CHANGELOG.md#deprecations-1),
+- **Feb 2024**: Version 0.14 is released, which deprecates ["jj checkout" and "jj merge"](CHANGELOG.md#0140---2024-02-07),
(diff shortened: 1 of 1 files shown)


PhilipMetzger:
This not what I meant in Discord 😄 , I expected that you'd add the new articles to the media section.

ilyagr (approved):
Seems like a good idea to me. You might want to wait a bit before merging in case others have thoughts.

bnjmnt4n:
> This not what I meant in Discord 😄 , I expected that you'd add the new articles to the media section.

Was there any article in particular to add? I thought we shouldn't overload the readme with lots of articles, but instead Chris's intro and videos + Steve's tutorial would be sufficient. Instead, linking to other articles would allow users to explore different workflows only if they desired.

PhilipMetzger (approved):
There was no singular article to add, but all of them which were released in the time between march and july. It is fine as is, but I think we should re-discuss our approach to in-repo docs as the wiki isn't indexed by Google.

PR #4052, opened 2024-07-09
cli: skip formatting instructions if not required

scott2000:
Suggested in #4046.

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2248,12 +2248,12 @@ impl DiffSelector {
         left_tree: &MergedTree,
         right_tree: &MergedTree,
         matcher: &dyn Matcher,
-        instructions: Option<&str>,
+        format_instructions: impl FnOnce() -> String,
     ) -> Result<MergedTreeId, CommandError> {
         match self {
             DiffSelector::NonInteractive => Ok(restore_tree(right_tree, left_tree, matcher)?),
             DiffSelector::Interactive(editor) => {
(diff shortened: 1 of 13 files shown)


cli/src/cli_util.rs
```diff
@@ -2248,12 +2248,13 @@ impl DiffSelector {
...
         matcher: &dyn Matcher,
-        instructions: Option<&str>,
+        format_instructions: impl FnOnce() -> String,
     ) -> Result<MergedTreeId, CommandError> {
         match self {
             DiffSelector::NonInteractive => Ok(restore_tree(right_tree, left_tree, matcher)?),
             DiffSelector::Interactive(editor) => {
-                Ok(editor.edit(left_tree, right_tree, matcher, instructions)?)
+                let instructions = format_instructions();
+                Ok(editor.edit(left_tree, right_tree, matcher, Some(&instructions))?)
```
yuja, line 2257: nit: `format_instructions` could be passed in to `.edit()`. Not all editors use instructions.

BTW, instruction templates might have to be defined per editor, but I don't know how many customization points (and complexities) we'll want.

yuja (approved):
Thanks for adding tests.

cli/src/cli_util.rs
scott2000: That's a good idea. I updated it, could you take a look and see if it looks good when you get a chance?

yuja (approved):

PR #4067, opened 2024-07-10
--color=debug: combine segments with same labels

jonathantanmy:
I noticed this when doing the diff whitespace work.

(notice the repeated "diff removed line_number" labels) From
```
[38;5;1m<<diff removed line_number:: >><<diff removed line_number:: >><<diff removed line_number:: >><<diff removed line_number::1>>[39m<<diff:: >><<diff::    : >>[4m[38;5;1m<<diff removed token::foo>>[24m[39m
```
to
```
[38;5;1m<<diff removed line_number::   1>>[39m<<diff::     : >>[4m[38;5;1m<<diff removed token::foo>>[24m[39m
```

# Checklist

If applicable:

### DIFF
--- a/cli/src/formatter.rs
+++ b/cli/src/formatter.rs
@@ -264,7 +264,6 @@ impl Style {
 #[derive(Clone, Debug)]
 pub struct ColorFormatter<W: Write> {
     output: W,
-    debug: bool,
     rules: Arc<Rules>,
     /// The stack of currently applied labels. These determine the desired
     /// style.
@@ -272,6 +271,9 @@ pub struct ColorFormatter<W: Write> {
     cached_styles: HashMap<Vec<String>, Style>,
     /// The style we last wrote to the output.
     current_style: Style,
+    /// The debug string (space-separated labels) we last wrote to the output.
+    /// Initialize to None to turn debug strings off.
+    current_debug: Option<String>,
 }
 
 impl<W: Write> ColorFormatter<W> {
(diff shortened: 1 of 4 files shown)


ilyagr:
We discussed this idea with @yuja previously, and he was leaning against it.

However, I still like the idea, since it's so nice to have readable tests. Yuya, if you still would like to have the debug mode to report precisely what happened, would you perhaps be OK with having a separate debug mode for more readable output?

cli/src/formatter.rs
```diff
@@ -272,6 +272,8 @@ pub struct ColorFormatter<W: Write> {
     cached_styles: HashMap<Vec<String>, Style>,
     /// The style we last wrote to the output.
     current_style: Style,
+    /// The debug string (space-separated labels) we last wrote to the output.
+    current_debug: String,
```
yuja, line 276: nit: for extra compile-time guarantee, maybe we can combine `debug: bool` and `current_debug: String` as `Option<String>` or `current_debug_labels: Option<Vec<String>>`?

cli/src/formatter.rs
```diff
@@ -492,36 +514,19 @@ impl<W: Write> Write for ColorFormatter<W> {
...
-            if debug {
-                write!(output, "<<{}::", labels.join(" "))?;
-                write_sanitized(output, buf)?;
-                write!(output, ">>")?;
-            } else {
-                write_sanitized(output, buf)?;
-            }
-
+        fn write_line_exclusive<W: Write>(output: &mut W, buf: &[u8]) -> io::Result<()> {
+            write_sanitized(output, buf)?;
```
yuja, line 529: nit: inline this

yuja (approved):
I don't wanna introduce complexity in debug code, which was the main motivation to not deduplicate label outputs. I think this implementation is simple enough to get in. Thanks.

cli/src/formatter.rs
jonathantanmy: Done.

Issue #3529, opened 2024-04-18
FR: mark locally created changes as such and give `local()` revset (like `mine()`) 


dpc:
**Is your feature request related to a problem? Please describe.**

The longer description of my otivation is already in https://github.com/martinvonz/jj/discussions/3528

But in a nutshell: it's hard to comprehensively track own work in progress without branches.

**Describe the solution you'd like**

I wish `jj` allowed easier revset selection of changes that were created locally, as these are the changes users usually would care about. Seems easy to understand and reason about.

`mine()`, `author()`/`committer()` are not quite like it as they match tons of remote stuff.

`local()` seems trivial to implement. Just an additional set in the db, where any `jj new`-created change_id gets added.

To support cases where user would like to "import" stuff as local some kind of CLI command to add and remove changes to `local()` would be needed.

**Describe alternatives you've considered**
I wasted hours trying to come up with a revset formula that would show me only stuff I care about in a repository with lots of commits, release branches. Random branches where my commmits were cherry-picked, and bunch of random branches made by me in the past. A combination of picking `latest` of `commiter`/`author`, etc.

It always either undershoots or overshoots.

yuja:
> `local()` seems trivial to implement. Just an additional set in the db, where any `jj new`-created change_id gets added.

I don't think this particular model will work out (because growing set can be a problem in small repos), but some commit metadata (similar to #3402) might be useful to determine whether commits were originally created locally. Currently, the "locally-created" state could deduced from change id, but it's implementation detail of the Git backend.

dpc:
> I don't think this particular model will work out (because growing set can be a problem in small repos), but some commit metadata (similar to #3402) might be useful to determine whether commits were originally created locally.

I don't understand this part. Storing sets of things efficiently is all that databases do. By a "set" here I meant a list of keys. In an SQL db a table with primary key and no other columns.

dpc:
> I just mean the "local change ids" set will grow indefinitely

It definitely is not going to grow as fast as the whole history and `jj` deals with that and many other sets just fine. Eg. `mine()` or `author()` can potentially refer to even bigger set. I don't really have preferences how to implemented it BTW, just wanted to underline that logically it's just a plain set without much extra functionality/requirements.

[...]

PR #4068, opened 2024-07-10
revset: add tracked/untracked_remote_branches()

scott2000:
Mentioned in #3528 and #3529.

# Checklist

If applicable:

### DIFF
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -34,7 +34,7 @@ use crate::graph::GraphEdge;
 use crate::hex_util::to_forward_hex;
 use crate::id_prefix::IdPrefixContext;
 use crate::object_id::{HexPrefix, PrefixResolution};
-use crate::op_store::WorkspaceId;
+use crate::op_store::{RemoteRefState, WorkspaceId};
 use crate::repo::Repo;
 use crate::repo_path::RepoPathUiConverter;
 pub use crate::revset_parser::{
@@ -107,6 +107,7 @@ pub enum RevsetCommitRef {
     RemoteBranches {
         branch_pattern: StringPattern,
         remote_pattern: StringPattern,
+        remote_ref_state: Option<RemoteRefState>,
     },
     Tags,
     GitRefs,
@@ -239,11 +240,13 @@ impl RevsetExpression {
     pub fn remote_branches(
         branch_pattern: StringPattern,
         remote_pattern: StringPattern,
+        remote_ref_state: Option<RemoteRefState>,
     ) -> Rc<RevsetExpression> {
         Rc::new(RevsetExpression::CommitRef(
             RevsetCommitRef::RemoteBranches {
                 branch_pattern,
                 remote_pattern,
+                remote_ref_state,
             },
         ))
     }
@@ -626,22 +629,13 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
         Ok(RevsetExpression::branches(pattern))
     });
     map.insert("remote_branches", |function, _context| {
-        let ([], [branch_opt_arg, remote_opt_arg]) =
-            function.expect_named_arguments(&["", "remote"])?;
-        let branch_pattern = if let Some(branch_arg) = branch_opt_arg {
-            expect_string_pattern(branch_arg)?
-        } else {
-            StringPattern::everything()
-        };
-        let remote_pattern = if let Some(remote_arg) = remote_opt_arg {
-            expect_string_pattern(remote_arg)?
-        } else {
-            StringPattern::everything()
-        };
-        Ok(RevsetExpression::remote_branches(
-            branch_pattern,
-            remote_pattern,
-        ))
+        parse_remote_branches_arguments(function, None)
+    });
+    map.insert("tracked_remote_branches", |function, _context| {
+        parse_remote_branches_arguments(function, Some(RemoteRefState::Tracking))
+    });
+    map.insert("untracked_remote_branches", |function, _context| {
+        parse_remote_branches_arguments(function, Some(RemoteRefState::New))
     });
     map.insert("tags", |function, _context| {
         function.expect_no_arguments()?;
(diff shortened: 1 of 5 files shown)


lib/src/revset.rs
```diff
@@ -638,9 +641,19 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
...
             StringPattern::everything()
         };
+        let remote_ref_state = if let Some(tracked_arg) = tracked_opt_arg {
+            if revset_parser::expect_bool(tracked_arg)? {
+                Some(RemoteRefState::Tracking)
+            } else {
+                Some(RemoteRefState::New)
+            }
+        } else {
+            None
```
yuja, line 651: I think the default should also be explicitly specifiable by user, by using something like `tracked|untracked|any`? Separate `tracked_remote_branches()`/`untracked_remote_branches()` functions seem also good.

Thoughts?

lib/src/view.rs
```diff
@@ -168,17 +170,23 @@ impl View {
...
             .map(|(remote_name, remote_view)| {
-                branch_pattern.filter_btree_map(&remote_view.branches).map(
-                    |(branch_name, remote_ref)| {
+                branch_pattern
+                    .filter_btree_map(&remote_view.branches)
+                    .filter(move |(_, remote_ref)| {
+                        remote_ref_state
+                            .map(|state| remote_ref.state == state)
+                            .unwrap_or(true)
+                    })
```
yuja, line 185: nit: I'll move this to caller as there's no way to optimize the query.

lib/src/revset.rs
scott2000: I agree that having a way to explicitly specify the default value would be good. Maybe the argument name would be `state` then (e.g. `remote_branches(state=untracked)`)?

I think that having `tracked_remote_branches()` and `untracked_remote_branches()` would work as well, and it might be simpler due to not having to parse `tracked`, `untracked`, and `any`. It might also be more discoverable due to function name suggestions.

On the other hand, tracked and untracked remote branches don't seem like fundamentally different types of branches to me (at least not as different as local branches and remote branches are), so I'm not sure that they're different enough for it to make sense to have them be different revsets. I also think it would result in more duplication, since we'd probably want to support filtering by branch name and remote name for all 3 revsets.

I don't have a strong feeling either way. Which do you think is better?

yuja: I don't have strong preference either. `remote_branches(state=untracked)` looks good to me. If a typical use case is `remote_branches(state=untracked)` (without name pattern), `untracked_remote_branches()` might be better. The downside is that it's uneasy to extend if we have more search options.

@ilyagr any comments?

In any case, the underlying `RevsetExpression` enum can shared. (i.e. `{ branch, remote, state: Option<State> }`.)

scott2000: > If a typical use case is `remote_branches(state=untracked)` (without name pattern), `untracked_remote_branches()` might be better. 

My guess would be that it would usually be used without any names (e.g. marking all untracked branches as immutable), but sometimes it would be useful to filter by branch or remote.

> The downside is that it's uneasy to extend if we have more search options.

Yeah, in general it seems impractical to add new revsets for each boolean filter option, since with 2 filters we'd need 9 revsets to cover every combination. I can't think of any more filter options that would apply to remote branches though, so maybe that wouldn't be an issue.

I updated it to use the `state` argument version for now since it was a small change, but I'll wait to hear what @ilyagr thinks.

ilyagr: I think `untracked_remote_branches()` and `tracked_remote_branches()` is pretty safe for now.

Another option we could perhaps consider is `remote_branches(tracked=true/false/any)`.

I like `state=tracked/untracked` the least, for two reasons:

- I think generic-sounding keywords like `state=` are annoying and difficult to remember.
- We might want to add "branches(tracked_by_remotes=....)" or something to that effect. That would not work with "state".

Though it would also do for now.

-------

I wonder if we eventually need a general solution for this, a "branchset language" of some sort.
We could have a general `branchexpr()` function and have `remote_branches(x) = branchexpr( remote() & x )`.

scott2000: > Another option we could perhaps consider is `remote_branches(tracked=true/false/any)`.

I personally find it a bit strange to have `any` for a boolean option, since most programming languages don't use a three-valued logic, so I think I'd prefer `untracked_remote_branches()` and `tracked_remote_branches()`.

> I like `state=tracked/untracked` the least, for two reasons:
>
>   * I think generic-sounding keywords like `state=` are annoying and difficult to remember.
>
>   * We might want to add `branches(tracked_by_remotes="glob:work-*")` or something to that effect for local branches. That would not work with "state".

That makes sense to me.

> I wonder if we eventually need a general solution for this, a "branchset language" of some sort. We could have a general `branchexpr()` function and have `remote_branches(x) = branchexpr( remote() & x )`.

I had actually been thinking about that! If we were going to add more filters, that would probably be nicer than just adding them as arguments or separate revsets.

yuja: Let's settle on `remote_branches()`/`tracked_remote_branches()`/`untracked_remote_branches()` then. I don't like tri-state boolean, and I agree `state=` is a bit vague word.

Implementation-wise, inner arguments parsing function can be extracted.

lib/src/revset.rs
```diff
@@ -1609,11 +1626,17 @@ fn resolve_commit_ref(
...
+            remote_ref_state,
         } => {
             // TODO: should we allow to select @git branches explicitly?
             let commit_ids = repo
                 .view()
                 .remote_branches_matching(branch_pattern, remote_pattern)
+                .filter(|(_, remote_ref)| {
+                    remote_ref_state
+                        .map(|state| remote_ref.state == state)
+                        .unwrap_or(true)
```
yuja, line 1638: nit: `.map_or(true,` also works (if you like)

lib/tests/test_revset.rs
```diff
@@ -2061,7 +2061,7 @@ fn test_evaluate_expression_remote_branches() {
     let repo = &test_repo.repo;
     let remote_ref = |target| RemoteRef {
```
yuja, line 2062: nit: let's rename to `tracking_remote_ref()` as the state does matter now.

yuja (approved):
Thanks!

lib/src/revset.rs
scott2000: Oh I didn't know about that function, that's nice!

PR #4072, opened 2024-07-11
Fix stdout warnings

tim-janik:
This PR fixes warnings ending up on stdout.

As suggested by @crackcomm on discord, use eprintln!() to print warnings
to avoid messing up template output, e.g.:
    
    jj --no-pager --ignore-working-copy show --tool true -T change_id -r rv...
    rv...ignoring git submodule at "some/submodule"


yuja (approved):
Thanks.

PR #4079, opened 2024-07-12
Add multi-tool config for `jj fix`

hooper:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -19,10 +19,12 @@ use std::sync::mpsc::channel;
 
 use futures::StreamExt;
 use itertools::Itertools;
-use jj_lib::backend::{BackendError, BackendResult, CommitId, FileId, TreeValue};
+use jj_lib::backend::{BackendError, CommitId, FileId, TreeValue};
+use jj_lib::fileset::{self, FilesetExpression};
+use jj_lib::matchers::{EverythingMatcher, Matcher};
 use jj_lib::merged_tree::MergedTreeBuilder;
 use jj_lib::repo::Repo;
-use jj_lib::repo_path::RepoPathBuf;
+use jj_lib::repo_path::{RepoPathBuf, RepoPathUiConverter};
 use jj_lib::revset::{RevsetExpression, RevsetIteratorExt};
 use jj_lib::store::Store;
 use pollster::FutureExt;
@@ -31,8 +33,8 @@ use rayon::prelude::ParallelIterator;
 use tracing::instrument;
 
 use crate::cli_util::{CommandHelper, RevisionArg};
-use crate::command_error::{config_error_with_message, CommandError};
-use crate::config::CommandNameAndArgs;
+use crate::command_error::{config_error, CommandError};
+use crate::config::{to_toml_value, CommandNameAndArgs};
 use crate::ui::Ui;
 
 /// Update files with formatting fixes or other changes
@@ -42,26 +44,60 @@ use crate::ui::Ui;
 /// It can also be used to modify files with other tools like `sed` or `sort`.
 ///
 /// The changed files in the given revisions will be updated with any fixes
-/// determined by passing their file content through the external tool.
-/// Descendants will also be updated by passing their versions of the same files
-/// through the same external tool, which will never result in new conflicts.
-/// Files with existing conflicts will be updated on all sides of the conflict,
-/// which can potentially increase or decrease the number of conflict markers.
+/// determined by passing their file content through any external tools the user
+/// has configured for those files. Descendants will also be updated by passing
+/// their versions of the same files through the same tools, which will ensure
+/// that the fixes are not lost. This will never result in new conflicts. Files
+/// with existing conflicts will be updated on all sides of the conflict, which
+/// can potentially increase or decrease the number of conflict markers.
 ///
-/// The external tool must accept the current file content on standard input,
-/// and return the updated file content on standard output. The output will not
-/// be used unless the tool exits with a successful exit code. Output on
-/// standard error will be passed through to the terminal.
+/// The external tools must accept the current file content on standard input,
+/// and return the updated file content on standard output. A tool's output will
+/// not be used unless it exits with a successful exit code. Output on standard
+/// error will be passed through to the terminal.
 ///
-/// The configuration schema is expected to change in the future. For now, it
-/// defines a single command that will affect all changed files in the specified
-/// revisions. For example, to format some Rust code changed in the working copy
-/// revision, you could write this configuration:
+/// Tools are defined in a table where the keys are arbitrary identifiers and
+/// the values have the following properties:
+///  - `command`: The arguments used to run the tool. The first argument is the
+///    path to an executable file. Arguments can contain the substring `$path`,
+///    which will be replaced with the repo-relative path of the file being
+///    fixed. It is useful to provide the path to tools that include the path in
+///    error messages, or behave differently based on the directory or file
+///    name.
+///  - `patterns`: Determines which files the tool will affect. If this list is
+///    empty, no files will be affected by the tool. If there are multiple
+///    patterns, the tool is applied only once to each file in the union of the
+///    patterns.
+///
+/// For example, the following configuration defines how two code formatters
+/// (`clang-format` and `black`) will apply to three different file extensions
+/// (.cc, .h, and .py):
+///
+/// [fix.tools.clang-format]
+/// command = ["/usr/bin/clang-format", "--assume-filename=$path"]
+/// patterns = ["glob:'**/*.cc'",
+///             "glob:'**/*.h'"]
+///
+/// [fix.tools.black]
+/// command = ["/usr/bin/black", "-", "--stdin-filename=$path"]
+/// patterns = ["glob:'**/*.py'"]
+///
+/// Execution order of tools that affect the same file is deterministic, but
+/// currently unspecified, and may change between releases. If two tools affect
+/// the same file, the second tool to run will receive its input from the
+/// output of the first tool.
+///
+/// There is also a deprecated configuration schema that defines a single
+/// command that will affect all changed files in the specified revisions. For
+/// example, the following configuration would apply the Rust formatter to all
+/// changed files (whether they are Rust files or not):
 ///
 /// [fix]
 /// tool-command = ["rustfmt", "--emit", "stdout"]
 ///
-/// And then run the command `jj fix -s @`.
+/// The tool defined by `tool-command` acts as if it was the first entry in
+/// `fix.tools`, and uses `pattern = "all()"``. Support for `tool-command`
+/// will be removed in a future version.
 #[derive(clap::Args, Clone, Debug)]
 #[command(verbatim_doc_comment)]
 pub(crate) struct FixArgs {
@@ -82,6 +118,7 @@ pub(crate) fn cmd_fix(
     args: &FixArgs,
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
+    let tools_config = get_tools_config(ui, command.settings().config())?;
     let root_commits: Vec<CommitId> = if args.source.is_empty() {
         workspace_command.parse_revset(&RevisionArg::from(
             command.settings().config().get_string("revsets.fix")?,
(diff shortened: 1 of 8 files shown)


cli/src/commands/fix.rs
```diff
@@ -42,26 +44,61 @@ use crate::ui::Ui;
...
+///    empty, no files will be affected by the tool. If there are multiple
+///    patterns, the tool is applied only once to each file in the union of the
+///    patterns.
+///
+/// For example, the following configuration defines how two code formatters
+/// (`clang-format` and `black`) will apply to three different file extensions
+/// (.cc, .h, and .py):
+///
+/// [[fix.tools]]
+/// name = "clang-format"
```
yuja, line 81: Maybe better to use table instead of array? `[fix.tools.<name>]`

Array items can't be overridden individually iirc, so `[[fix.tools]]` in repo config would shadow tools defined in the user config at all. If a tool is keyed by name, user can disable one by overriding `<name>.patterns = "none()"`.

cli/src/commands/fix.rs
```diff
@@ -42,26 +44,61 @@ use crate::ui::Ui;
...
+/// command = ["clang-format", "--assume-filename=$path"]
+/// patterns = ["glob:\"**/*.cc\"",
+///             "glob:\"**/*.h\""]
```
yuja, line 84: nit: single quote also works. (`'glob:"**/*.cc"'` or `"glob:'**/*.cc'"`)

cli/src/commands/fix.rs
```diff
@@ -270,16 +302,46 @@ fn fix_file_ids<'a>(
...
+            let mut any_changes: bool = false;
+            for tool_config in tools_config.tools.iter() {
+                if tool_config.matcher.matches(&tool_input.repo_path) {
+                    if prev_content.is_none() {
+                        let mut content = vec![];
+                        let mut read =
+                            store.read_file(&tool_input.repo_path, &tool_input.file_id)?;
+                        read.read_to_end(&mut content).unwrap();
+                        prev_content = Some(content);
+                    }
```
yuja, line 319: nit: perhaps, this `Option<_>` can be removed by peeking matching tools. Something like
```rust
let mut matched_tools = tools.iter().filter(|tool| tool.matcher.matches(..)).peekable();
if matched_tools.peek().is_none() {
    return;
}
read_file(..);
for tool in matched_tools {
...
```

cli/src/commands/fix.rs
```diff
@@ -270,16 +302,46 @@ fn fix_file_ids<'a>(
...
+                    if let Ok(next_content) = run_tool(
+                        &tool_config.command,
+                        tool_input,
+                        prev_content.as_ref().unwrap(),
+                    ) {
+                        if next_content != *prev_content.as_ref().unwrap() {
+                            prev_content = Some(next_content);
+                            any_changes = true;
+                        }
+                    }
```
yuja, line 329: nit: should it terminate or continue on error?

cli/src/commands/fix.rs
```diff
@@ -328,3 +390,126 @@ fn run_tool(
...
+                    .into());
+                }
+                let patterns: Vec<String> = table
+                    .get("patterns")
+                    .ok_or(config::ConfigError::Message(format!(
+                        "`fix.tools` entry `{}` at index {} is missing required field `patterns`.",
+                        name, index
+                    )))?
+                    .clone()
+                    .into_array()?
```
yuja, line 461: nit: `.try_deserialize()` can be used to map `config::Value` to arbitrary type.

We can also define intermediate config struct (like `ExternalMergeTool`) and deserialize into it.

cli/src/commands/fix.rs
```diff
@@ -328,3 +390,126 @@ fn run_tool(
...
+                            )
+                            .map_err(|e| {
+                                config::ConfigError::Message(format!(
+                                    "`fix.tools` entry `{}` at index {}: {}",
+                                    name, index, e
+                                ))
+                            })?,
+                        ),
+                        matcher: workspace_command
+                            .parse_union_filesets(&patterns)?
```
yuja, line 496: nit: this will resolve patterns relative to cwd, not to workspace root. I'm not sure if that's correct or wrong.

cli/src/commands/fix.rs
```diff
@@ -13,26 +13,29 @@
 // limitations under the License.
 
 use std::collections::{HashMap, HashSet};
+use std::hash::Hash;
```
yuja, line 16: nit: `use Hash` is redundant

cli/src/commands/fix.rs
```diff
@@ -328,3 +382,111 @@ fn run_tool(
...
+/// Represents the `fix.tools` config table.
+struct ToolsConfig {
+    /// Some tools, stored in the order they will be executed if more than one
+    /// of them matches the same file.
+    tools: Vec<ToolConfig>,
+}
+
+/// Simplifies deserialization of the config values while building a ToolConfig.
+#[derive(Clone, Debug, Default, Eq, PartialEq, serde::Deserialize)]
+#[serde(default)]
```
yuja, line 405: nit: might be better to set `rename_all = "kebab-case"` though it doesn't matter right now.

cli/src/commands/fix.rs
```diff
@@ -328,3 +382,111 @@ fn run_tool(
...
+struct RawToolConfig {
+    command: Vec<String>,
+    patterns: Vec<String>,
+}
```
yuja, line 409: nit: `CommandNameAndArgs` can be embedded (if it's okay to support string and structured forms)
```rust
#[derive(Clone, Debug, Eq, PartialEq, serde::Deserialize)]
#[serde(rename_all = "kebab-case")]
struct RawToolConfig {
    command: CommandNameAndArgs,
    patterns: Vec<String>,
}
```

cli/src/commands/fix.rs
```diff
@@ -328,3 +382,111 @@ fn run_tool(
...
+                                .map(|arg| {
+                                    fileset::parse(
+                                        arg,
+                                        // TODO: This is probably a misuse of Fs, but there is
+                                        // currently no other option for non-cwd-relative fileset
+                                        // parsing.
+                                        &RepoPathUiConverter::Fs {
+                                            cwd: workspace_command.workspace_root().clone(),
+                                            base: workspace_command.workspace_root().clone(),
+                                        },
```
yuja, line 469: nit: `cwd: "".into(), base: "".into()` also works, and it's probably okay as a temporary hack.

cli/src/commands/fix.rs
```diff
@@ -328,3 +382,111 @@ fn run_tool(
...
+            .sorted_by(|a, b| a.0.cmp(&b.0))
+            .map(|(name, value)| -> Result<ToolConfig, CommandError> {
+                let tool: RawToolConfig = value.try_deserialize()?;
+                if tool.patterns.is_empty() {
+                    Err(config::ConfigError::Message(format!(
+                        "`fix.tools.{}.patterns` must not be empty. Use patterns=['all()'] if you \
+                         want to run this tool on every file.",
+                        name
+                    ))
+                    .into())
```
yuja, line 446: nit: maybe okay to accept explicit `patterns = []` to disable the tool? I agree that missing `patterns` should be an error, though.

Another option is to make it string and force users to write `"glob:'**/*.cc'|glob:'**/*.h'"`.

cli/src/commands/fix.rs
```diff
@@ -42,26 +45,64 @@ use crate::ui::Ui;
...
+/// and return the updated file content on standard output. A tool's output will
+/// not be used unless it exits with a successful exit code. Output on standard
+/// error will be passed through to the terminal.
 ///
-/// The configuration schema is expected to change in the future. For now, it
-/// defines a single command that will affect all changed files in the specified
-/// revisions. For example, to format some Rust code changed in the working copy
-/// revision, you could write this configuration:
+/// Tools are defined in a table, and are applied in the same order as they are
+/// defined. If two tools affect the same file, the second tool will receive its
```
yuja, line 61: nit: "in the same order as they are defined" is no longer correct?

We also need to update the example to use table syntax.

cli/src/config-schema.json
```diff
@@ -515,7 +515,35 @@
                     "items": {
                         "type": "string"
                     },
-                    "description": "Shell command that takes file content on stdin and returns fixed file content on stdout"
+                    "description": "Shell command that takes file content on stdin and returns fixed file content on stdout (deprecated)"
+                },
+                "tools": {
+                    "type": "array",
```
yuja, line 521: nit: has to be updated to table

cli/src/commands/fix.rs
```diff
@@ -328,3 +382,111 @@ fn run_tool(
...
+                }
+            })
+            .try_collect()?;
+        tools_config.tools.append(&mut tools);
+    }
+    if tools_config.tools.is_empty() {
+        // TODO: This is not a useful message when one or both fields are present but
+        // have the wrong type. After removing `fix.tool-command`, it will be simpler to
+        // propagate any errors from `config.get_array("fix.tools")`.
+        Err(config::ConfigError::Message(
```
yuja, line 485: nit: maybe we can use `config_error()` as the return type is `CommandError`.

cli/tests/test_fix_command.rs
```diff
@@ -16,6 +16,7 @@
 use std::os::unix::fs::PermissionsExt;
 use std::path::PathBuf;
 
+use gix::bstr::ByteVec;
```
yuja, line 19: nit: `use bstr::ByteVec` (I've added dependency on `bstr`)

yuja (approved):
I'm not familiar with `jj fix`, but the change looks good to me, thanks!

cli/src/commands/fix.rs
```diff
@@ -328,3 +382,111 @@ fn run_tool(
...
+        // Convert the map into a sorted vector early so errors are deterministic.
+        let mut tools: Vec<ToolConfig> = tools_table
+            .into_iter()
+            .sorted_by(|a, b| a.0.cmp(&b.0))
+            .map(|(name, value)| -> Result<ToolConfig, CommandError> {
+                let tool: RawToolConfig = value.try_deserialize()?;
+                if tool.patterns.is_empty() {
+                    Err(config::ConfigError::Message(format!(
+                        "`fix.tools.{}.patterns` must not be empty. Use patterns=['all()'] if you \
+                         want to run this tool on every file.",
```
PhilipMetzger, line 443: Nit: Probably simpler to just use a raw string here.

cli/src/commands/fix.rs
```diff
@@ -328,3 +382,111 @@ fn run_tool(
...
+                                })
+                                .try_collect()?,
+                        )
+                        .to_matcher(),
+                    })
+                }
+            })
+            .try_collect()?;
+        tools_config.tools.append(&mut tools);
+    }
```
PhilipMetzger, line 480: Nit: Add a deprecation warning here.

PhilipMetzger (approved):
LG

cli/src/commands/fix.rs
hooper: I don't think this works, because `CommandNameAndArgs` doesn't implement `Default`. Am I missing something?

yuja: Do we need `Default` at all? Both `command` and `patterns` are required.

hooper: I see what you mean now. Is it true that `Matcher` would still need some work to support `Deserialize` like this?

yuja: `Matcher` won't fit to `Deserialize` because it needs some environment parameters to build, I suppose. (Maybe could be build with `Fs { cwd: "", base: "" }`, but it seems bad idea to encode a specific environment into the type.)

cli/src/commands/fix.rs
PhilipMetzger: After we successfully parse the old syntax, there should be a warning like `"fix.tool-command is deprecated and will be removed in a future version, use fix.tool.<name> instead."`

hooper: Oh, good idea. Thanks. Please take another look.

PhilipMetzger: This is great.

PhilipMetzger (approved):
LG, there's one last Nit though.

cli/src/config-schema.json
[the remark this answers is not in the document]
hooper: Added a section there. Also tweaked the test to work on windows.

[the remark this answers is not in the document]
PhilipMetzger: Thanks, this also is great.

PR #3977, opened 2024-06-26
conflicts: insert trailing newline when materializing conflicts

ilyagr:
This happens for non-empty files only; the materialization code can
handle files and hunks with 0 or more lines, but each line should
terminate in a newline.

This is an imperfect fix to https://github.com/martinvonz/jj/issues/3968, a better fix is TODO; see https://github.com/martinvonz/jj/issues/3975.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- (n/a) I have updated the documentation (README.md, docs/, demos/)
- (n/a) I have updated the config schema (cli/src/config-schema.json)
- [x] I have added tests to cover my changes

PR #4082, opened 2024-07-13
diff: leverage BStr for debug printing

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/diff.rs
+++ b/lib/src/diff.rs
@@ -593,6 +593,18 @@ pub enum DiffHunk<'input> {
     Different(Vec<&'input [u8]>),
 }
 
+impl<'input> DiffHunk<'input> {
+    pub fn matching<T: AsRef<[u8]> + ?Sized>(content: &'input T) -> Self {
+        DiffHunk::Matching(content.as_ref())
+    }
+
+    pub fn different<T: AsRef<[u8]> + ?Sized + 'input>(
+        contents: impl IntoIterator<Item = &'input T>,
+    ) -> Self {
+        DiffHunk::Different(contents.into_iter().map(AsRef::as_ref).collect())
+    }
+}
+
 impl Debug for DiffHunk<'_> {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
         match self {
@@ -660,13 +672,13 @@ impl<'diff, 'input> Iterator for DiffHunkIterator<'diff, 'input> {
 /// the changed ranges at the word level.
 pub fn diff<'a>(left: &'a [u8], right: &'a [u8]) -> Vec<DiffHunk<'a>> {
     if left == right {
-        return vec![DiffHunk::Matching(left)];
+        return vec![DiffHunk::matching(left)];
     }
     if left.is_empty() {
-        return vec![DiffHunk::Different(vec![b"", right])];
+        return vec![DiffHunk::different([b"", right])];
     }
     if right.is_empty() {
-        return vec![DiffHunk::Different(vec![left, b""])];
+        return vec![DiffHunk::different([left, b""])];
     }
 
     Diff::default_refinement([left, right])
@@ -974,7 +986,7 @@ mod tests {
     #[test]
     fn test_diff_single_input() {
         let diff = Diff::default_refinement(["abc"]);
-        assert_eq!(diff.hunks().collect_vec(), vec![DiffHunk::Matching(b"abc")]);
+        assert_eq!(diff.hunks().collect_vec(), vec![DiffHunk::matching("abc")]);
     }
 
     #[test]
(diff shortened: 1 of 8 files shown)


ilyagr (approved):
Great, BStr seems like it should simplify diffing and conflicts, or at least debugging of them. Thank you!

lib/src/diff.rs
```diff
@@ -587,44 +587,24 @@ impl<'input> Diff<'input> {
...
 
-    pub fn different<T: AsRef<[u8]> + ?Sized + 'input>(
+    pub fn different<T: AsRef<BStr> + ?Sized + 'input>(
         contents: impl IntoIterator<Item = &'input T>,
     ) -> Self {
         DiffHunk::Different(contents.into_iter().map(AsRef::as_ref).collect())
     }
 }
 
-impl Debug for DiffHunk<'_> {
```
ilyagr, line 577: Are there any tests for this? I assumed there were, but I didn't immediately find them.

lib/src/diff.rs
```diff
@@ -471,14 +471,14 @@ impl<'input> Diff<'input> {
         diff
     }
 
-    pub fn unrefined<T: AsRef<[u8]> + ?Sized + 'input>(
+    pub fn unrefined<T: AsRef<BStr> + ?Sized + 'input>(
```
ilyagr, line 474: Naively, does `BStr` satisfy `AsRef<[u8]>`? Can it be built from `AsRef<u8]>`? If so, maybe it's better to keep the type as `AsRef<[u8]>`?

I do find `&*` confusing. My understanding of Rust is insufficient to understand exactly what happens there. I might, of course, look it up.

lib/src/diff.rs
ilyagr, line 1: Re commit message:

> b"" literals in tests are changed to &str to get around type incompatibility
between &[u8; N].

Naively, would https://docs.rs/bstr/latest/bstr/fn.B.html help here?

lib/src/diff.rs
ilyagr: I read up on the `bstr` crate, and it *seems* like it's supposed to work well with `[u8]` arguments everywhere. For example, in https://docs.rs/bstr/latest/bstr/index.html#bstr-in-public-apis it says that you *could* use BStr in public APIs, but there's usually no need for that.

I could easily be missing something important, though.

ilyagr:
Also, this is not related directly to this PR, but I should share the successor to #3977 soon. It seems to make a lot of sense to use the `bstr` in conflict processing too, e.g. https://docs.rs/bstr/latest/bstr/struct.LinesWithTerminator.html, but if we both work on that code, there will be merge conflicts.

lib/src/diff.rs
yuja: It probably works, but would look noisy.

lib/src/diff.rs
yuja: `[u8]` also works. I choose `BStr` just because it's shorter and seems to explain the API better.

lib/src/diff.rs
yuja: I don't think there are, but it might affect assertion failure.

ilyagr: (Optional) do you think it would make sense to add one?

lib/src/diff.rs
ilyagr: I still haven't fully figured out what's going on, but I think we should avoid `&*` if we can. If `[u8]` vs `BStr` makes a difference to that, I'd prefer the former. If not, either is fine.

ilyagr: Also, we're actually talking about `AsRef<[u8]>` vs `AsRef<BStr>`, and more types from other crates are likely to implement the former. So, `AsRef<[u8]>` seems better.

yuja: Maybe I should send an upstream PR that adds `AsRef<BStr> for Vec<u8>`.

lib/src/diff.rs
yuja: Maybe no? It's not so important, and the point of deriving `Debug` is that the generated code is less buggy.

lib/src/diff.rs
ilyagr: Just to be clear, I'm 100% behind using BStr internally in whichever types this makes debug-printing better for.

If it makes a difference, I'd be behind using it for other purposes as well. AFAICT based on what I read so far, the debug string is the only actual meaningful difference between `BStr` and `[u8]` because of the traits the `bstr` crate adds to `[u8]`, but I might be missing something.

>  I choose BStr just because it's shorter and seems to explain the API better.

I think Rust already conditions programmers to think "bytestring" when they see `&[u8]`, since that's the type of `b"bytestrings"`. So, I don't think there is much advantage to clarity in using BStr over it.

> Maybe I should send an upstream PR that adds AsRef<BStr> for Vec<u8>.

That wouldn't hurt, of course, but I'd still think that using `AsRef<[u8]>` would be slightly better. At that point, the difference would of course matter less.

lib/src/diff.rs
ilyagr: Sounds good.

lib/src/diff.rs
yuja: Ok, I don't have strong opinion, so switched to `[u8]`.

PR #4083, opened 2024-07-14
cli: include untracked remote branches in default immutable_heads()

yuja:
I used to use "remote_branches() & ~mine()" to exclude "their" branches from the default log, and I don't think that's uncommon requirement. Suppose untracked branches are usually read-only, it's probably okay to make them immutable by default.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj backout --revision` has been renamed to `jj backout --revisions`.
   The short alias `-r` is still supported.
 
+* [The default `immutable_heads()` set](docs/config.md#set-of-immutable-commits)
+  now includes `untracked_remote_branches()` with the assumption that untracked
+  branches aren't managed by you. Therefore, untracked branches are no longer
+  displayed in `jj log` by default.
+
 ### Deprecations
 
 ### New features
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -391,7 +391,7 @@
                 "immutable_heads()": {
                     "type": "string",
                     "description": "Revisions to consider immutable. Ancestors of these are also considered immutable. The root commit is always considered immutable.",
-                    "default": "trunk() | tags()"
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj backout --revision` has been renamed to `jj backout --revisions`.
   The short alias `-r` is still supported.
 
+* [The default `immutable_heads()` set](docs/config.md#set-of-immutable-commits)
```
ilyagr, line 18: Assuming this is true, you should mention that untracked branches no longer show up in the `jj log` default revset.

ilyagr (approved):
I think this is a good idea. Nice!

It seems like this will affect the default log revset automatically? See my other comment. Ultimately, I think the new behavior would be quite sensible.

If so, the worst I imagine happening is that people will be confused by the change to `jj log` default output, and maybe we should document in more places that you need to `jj track` branches for them to show up in the default `jj log`.

There is still a chance there is some other problem neither of us sees, but I think this would be nice to test out on anyone using the master branch.

Still, let's wait for a day before merging, in case others have comments (for example, if anyone wants us to wait longer).

ilyagr:
Also, I'll Cc: https://github.com/martinvonz/jj/issues/3529 . I haven't been following that conversation carefully, but on one hand it is connected, and on the other hand if we do come up with some `local()` revset it might be appropriate to add something based on `~local()` to the immutable heads.

scott2000:
I'm definitely in favor of this change. The main motivation for my PR for adding `untracked_remote_branches()` was so that I could add this to my `immutable_heads()` (mainly for the way it cleans up the log), and I imagine it would be a reasonable default for most people.

I also think it makes sense semantically for these commits to be immutable. Rewriting commits on untracked branches generally doesn't seem like a good idea since the remote branch will remain at the old commit even after it is rewritten, which probably isn't what a user would intend.

CHANGELOG.md
yuja: Update the changelog entry.

martinvonz (approved):
Good idea

PR #4085, opened 2024-07-14
diff: do not emit unified diff for binary files

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_diff_command.rs
+++ b/cli/tests/test_diff_command.rs
@@ -269,6 +269,152 @@ fn test_diff_empty() {
 }
 
 #[test]
+fn test_diff_file_mode() {
+    let test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    // Test content+mode/mode-only changes of empty/non-empty files:
+    // - file1: ("",  x) -> ("2", n)  empty, content+mode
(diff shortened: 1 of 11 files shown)


martinvonz (approved):
> ---/+++ lines are part of unified diff hunks, not Git diff header.

Ah, I suspect that's exactly what I had misunderstood. Thanks for fixing!

ilyagr (commented):
Thank you!

~~Re title: do you mean "for executable files"?~~

cli/tests/test_diff_command.rs
```diff
@@ -268,6 +268,146 @@ fn test_diff_empty() {
...
+    new mode 100644
+    index e69de29bb2..0cfbf08886
+    --- a/file1
+    +++ b/file1
+    @@ -1,0 +1,1 @@
+    +2
+    diff --git a/file2 b/file2
+    old mode 100755
+    new mode 100644
+    diff --git a/file3 b/file3
```
ilyagr, line 369: Do I understand correctly that this is a bug and it is not fixed in this PR?

If so, can you add a comment explaining what the bug is?

ilyagr (approved):

cli/tests/test_diff_command.rs
yuja: No, it's not a bug (if the quoted region is displayed correctly.) You can see header-only unified diffs in 5e2a9c274698f4b69b4b58cf226a811edd166e47.

The goal is to disable diff of binary files. I just found a couple of bugs while moving codes around, and I needed some tests covering these.

PR #4086, opened 2024-07-14
obslog: reverse order of predecessors in topo traversal

scott2000:
Currently, when there is a commit with two predecessors, the graph splits into two branches, and all of the predecessors on the first branch are printed before all of the predecessors on the second branch. This causes the graph to grow wider with each squashed commit, since the second branch must always get indented one level farther each time a commit is squashed. I have some commits where the graph is indented more than 10 levels due to squashing more than 10 times, making it very difficult to read.

Reversing the order and printing the second branch before the first branch prevents this unnecessary indentation and makes the graph easier to read. This does not change the order of the edges in the graph (i.e. the first predecessor is still the first edge and the second predecessor is still the second edge in the graph).

## Output before this PR

obslog before

## Output after this PR

obslog after

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/obslog.rs
+++ b/cli/src/commands/obslog.rs
@@ -113,7 +113,21 @@ pub(crate) fn cmd_obslog(
     let mut commits = topo_order_reverse_ok(
         vec![Ok(start_commit)],
         |commit: &Commit| commit.id().clone(),
-        |commit: &Commit| commit.predecessors().collect_vec(),
+        |commit: &Commit| {
+            let mut predecessors = commit.predecessors().collect_vec();
+            // Predecessors don't need to follow any defined order. However in
+            // practice, if there are multiple predecessors, then usually the
+            // first predecessor is the previous version of the same change, and
+            // the other predecessors are commits that were squashed into it. If
+            // multiple commits are squashed at once, then they are usually
+            // recorded in chronological order. We want to show squashed commits
+            // in reverse chronological order, and we also want to show squashed
+            // commits before the squash destination (since the destination's
+            // subgraph may contain earlier squashed commits as well), so we
+            // visit the predecessors in reverse order.
+            predecessors.reverse();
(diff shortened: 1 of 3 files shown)


cli/src/commands/obslog.rs
```diff
@@ -113,7 +113,11 @@ pub(crate) fn cmd_obslog(
     let mut commits = topo_order_reverse_ok(
         vec![Ok(start_commit)],
         |commit: &Commit| commit.id().clone(),
-        |commit: &Commit| commit.predecessors().collect_vec(),
+        |commit: &Commit| {
+            let mut predecessors = commit.predecessors().collect_vec();
```
yuja, line 117: nit: Can you add inline comment why walking from the second (or the last)? If I understand the idea, we assume that the first predecessor usually has longer history.

yuja (approved):

cli/src/commands/obslog.rs
scott2000: I struggled a bit with explaining my reasoning, so I added a graph example to try to clarify it. I'm not sure if it's too long of a comment now or if there's some easier way to explain it. Could you take a look and let me know what you think?

My main idea is that we want to show more recent changes towards the top of the log and less recent changes towards the bottom of the log, and the old iteration order was backwards for squashed commits (the most recent squashes appeared at the bottom of the log).

yuja: Okay, if the intent is to emit recent rewrites first, something like "visit from last because predecessors are usually recorded in chronological order on squash" will help.

My concern here was that `predecessors` has no defined order, so it's unclear what's our assumption.

scott2000: Ok I rewrote it to include that explanation so hopefully it's a bit clearer now. Let me know what you think!

yuja (approved):
Thanks!

Issue #3223, opened 2024-03-06
Unresolvable conflict when merging with empty file


matts1:
## Description

When one commit creates an empty file, and another commit creates a file of the same name, you get a conflict that can never be resolved.

I believe the core issue is that there are no conflict markers in such a file, and so jj says "file was unmodified" and hence thinks the resolution failed.

## Steps to Reproduce the Problem

1. `jj git init`
1. `touch foo`
1. `jj new @-`
2. `echo "abc" > foo`
3. `jj rebase -s @ <other commit>`
4. Attempt to resolve the merge that jj new created

## Expected Behavior

Either there is no conflict, or the conflict resolution works

## Actual Behavior

Conflict can never be resolved

## Specifications

- Platform: Linux
- Version: Internal unstable version

ilyagr:
Interesting. It should be possible to work around the bug by doing `jj restore --from some_commit file` (probably from one of the parent commits).

matts1:
Yeah, I worked around by renaming the file, running jj status, then renaming it back

Kintaro:
Following git's behavior, there is no conflict. So it seems `jj` is doing the same behavior here, but erroneously marks it as a conflict in the transaction.

Kintaro:
Should this be handled as a conflict or just merge the files as git does?

Kintaro:
Still new to the code base, but I think the issue stems from somewhere in `MergeTree::merge`. Looks like `as_resolved` doesn't try to resolve trivial conflicts.

martinvonz:
Copying my answer from Discord to here:

I think it ideally should be considered a conflict. The problem is that we use the empty string as a placeholder for a missing file when materializing/rendering the conflict markers. So the conflict becomes this:
```
<<<<<<<
+++++++
%%%%%%%
+abc
>>>>>>>
```

I.e. a conflict adding "abc\n" on one side, "" on the other side, and "" in the base. That gets automatically resolved to just "abc\n". (It doesn't actually work exactly this way.) I wonder how well it would work to replace the missing-file placeholder by e.g. "<missing file>". I'm not sure that's a good idea, and i don't want to use english-specific user-visible strings in the library code anyway

ilyagr:
This is a little confusing to me: it seems like we shouldn't *need& to change the conflict markers. As you said,

```
<<<<<<<
+++++++
%%%%%%%
+abc
>>>>>>>
```

[...]

PR #4088, opened 2024-07-15
Materialize files with conflict markers in a way that can be parsed

ilyagr:
Fixes https://github.com/martinvonz/jj/issues/3968

Fixes https://github.com/martinvonz/jj/issues/3975

# Checklist

If applicable:


lib/src/conflicts.rs
```diff
@@ -217,6 +224,12 @@ pub fn materialize_merge_result(
     single_hunk: Merge<ContentHunk>,
     output: &mut dyn Write,
 ) -> std::io::Result<()> {
+    let single_hunk: Merge<ContentHunk> =
+        single_hunk.into_map(|ContentHunk(side)| ContentHunk(encode_unmaterializeable_lines(side)));
+
+    // From now on, we assume some invariants guaranteed by the encoding function
+    // above. See its docstring for details.
+
```
yuja, line 232: As I said, I believe it's wrong to preprocess source texts to be merged. What we need here is to encode `merge_result`, right? Why do we have to do `source -> serialize1 -> merge -> serialize2` instead of `source -> merge -> serialize`?

lib/src/merge.rs
```diff
@@ -345,6 +345,12 @@ impl<T> Merge<T> {
         self.values.iter_mut()
     }
 
+    /// Creates a new merge by applying `f` to each remove and add.
+    pub fn into_map<U>(self, f: impl FnMut(T) -> U) -> Merge<U> {
```
yuja, line 349: nit: `into_map` sounds like returning a `map`. `map_owned()` or something might be better.

lib/src/conflicts.rs
```diff
@@ -331,7 +331,29 @@ fn diff_size(hunks: &[DiffHunk]) -> usize {
 /// invalid if they don't have the expected arity.
 // TODO: "parse" is not usually the opposite of "materialize", so maybe we
 // should rename them to "serialize" and "deserialize"?
-pub fn parse_conflict(input: &[u8], num_sides: usize) -> Option<Vec<Merge<ContentHunk>>> {
+pub fn parse_merge_result(input: &[u8], num_sides: usize) -> Option<Merge<ContentHunk>> {
+    let hunks = parse_conflict_into_list_of_hunks(input, num_sides)?;
+    let mut result: Merge<Vec<u8>> = Merge::from_vec(vec![vec![]; num_sides * 2 - 1]);
```
yuja, line 349: nit: you can build `Merge<ContentHunk>`, or `Vec<ContentHunk>` and apply `Merge::from_vec()` at the end.

lib/src/merge.rs
ilyagr: Thanks! `map_owned` sounds great!

lib/src/conflicts.rs
ilyagr: I thought the problem was that the transformation wasn't reversible.

The problem with serializing once is that there are many corner cases that I find it hard to reason about. By having an intermediate state, we don't have to worry about them.

Some corner cases:

- Conflict markers inside a conflict vs conflict markers that are the same on all sides of the conflict.
- "No newline at the end of file" outside a conflict (this one is pretty easy) vs on all sides of a conflict vs (on some sides of a conflict that is on the end of file for all sides) vs (on no sides of a conflict at the end of file) vs (one of the last two cases for a conflict that's only at the end of file for some of the sides.

This way, the implementation is much simpler and, more importantly, verifying that it works correctly is much simpler.

lib/src/conflicts.rs
ilyagr: I see you're trying to help out with this in #4106 . Thanks! I may clean it up in this PR, or we can do it separately; in any case we should figure out the bigger issues first.

lib/src/conflicts.rs
yuja: Hmm, but we have to instead design the `serialize1` part robust against line-by-line merging logic, right? Does it handle merge of two "no-eol" hunks for example? I suspect matching "no-eol" markers would be rendered out of the conflict region.

I didn't think about corner cases thoroughly, but I think nested conflict markers can be represented as a hunk (i.e. padded with " ") instead of a "base" content.

ilyagr: > Hmm, but we have to instead design the serialize1 part robust against line-by-line merging logic, right?

Yes, that's what I think I did. Or do you mean something else?

> Does it handle merge of two "no-eol" hunks for example? I suspect matching "no-eol" markers would be rendered out of the conflict region.

Yes, and then `jj` should import it correctly. I can add a test, it should look like:

```
<<<<<<<
aaa
...
>>>>>>>
\JJ: No newline at end of file
```

yuja: Ooh, so `\JJ:` lines are file global, not hunk local?

I would say it's super easy for users to lose track of them. I wouldn't expect that I had to fix up lines out of the loud conflict markers.

ilyagr: In principle, we could make it so that `\JJ: No newline at the end of file` is only inserted if there's a conflict at the end, and we could make sure it ends up inside the conflict markers.

This would cause a problem in the weird situation if the files actually have the text "\\JJ: No newline at the end of file\n" at the end (and there might be a conflict in that line). I see no way of solving this problem, but beyond bothering perfectionists, I don't think it's likely to come up.

If we think that is acceptable, we could make that change later.

This doesn't affect the treatment of literal conflict markers inside the file, for which I think this PR's solution is as good as it's going to get.

ilyagr: > I would say it's super easy for users to lose track of them. I wouldn't expect that I had to fix up lines out of the loud conflict markers.

If they lose track of them, that's perfectly fine. `jj` will import the side just as it was. The user would only need to pay attention if they want to change them inside the conflicted file view.

yuja: > if the sides actually have the text "\JJ: No newline at the end of file\n" at the end (and there might be a conflict in that line).

I *think* nested markers can be forced to be rendered as a hunk (i.e. escaped with " " prefix.)

Anyway, I prefer hunk-local syntax over perfection.

> If they lose track of them, that's perfectly fine. `jj` will import the side just as it was.

But the file-level conflict wouldn't be resolved until `\JJ` lines get removed by user? Or if it's automatically removed, things would look more weirder.

ilyagr: > But the file-level conflict wouldn't be resolved until \JJ lines get removed by user? Or if it's automatically removed, things would look more weirder.

It's automatically removed. This is crucial in the situation of

```
\JJ Verbatim Line:<<<<<<<
```

being automatically interpreted as the unconflicted text "<<<<<<<". In this case, the user *must not* remove the `\JJ` part themselves.

Similarly, the user is not expected to touch `\JJ No newline at the end of file`, though in this case, touching it has fewer bad consequences.

yuja: > It's automatically removed. 

You mean the working-copy content will be updated automatically?

1. user resolved conflict by editing file, leaving `\JJ No newline at ...` line (because it doesn't look like a conflict)
2. run `jj diff`

Does (2) remove `\JJ` line from the checked-out file?

ilyagr: Yes, once the file is not conflicted, the `\JJ No newline...` line (and the trailing newline) will disappear.

While there's at least one conflict within the file, that line will remain. Essentially, a "materialized conflicted file" becomes a special JJ-specific filetype with JJ-specific syntax, but the user shouldn't notice in the majority of cases. (I'm wondering whether that's a good or a bad thing, maybe making this more obvious to the user would be good. The idea from https://github.com/martinvonz/jj/pull/4088#discussion_r1682041261 would, OTOH, make it less obvious).

Again, for conflict markers inside files, this is the only practical solution I can think of, unless we decide it's not worth addressing. For #3223, I think this is a good solution. For the "\JJ No newline...", there might be other options, but I picked the solution that is analogous to the solution for the conflict markers and deleted sides.

ilyagr: Aside: we probably want a separate solution for `jj diff` showing the diff if the user removes a newline at the end of file, but I am not addressing that here. I think there the case-by-case solution should be fine: the `diff` code could just print " \JJ No newline at the end of file" in yellow, or "+\JJ: Added newline at the end of file" in green, or "-\JJ: Removed newline at the end of file" in red.

yuja: Okay, then I would say it's wrong that `jj diff` updates working-copy content. It's weird if `jj diff` suddenly fixed compile error, for example.

I think the conflict syntax needs a property that a content without conflicts is identical to the raw data. If the content had some meta data like `\JJ` line, it would still be considered a conflict.

ilyagr: I may have misunderstood your question. I assumed that before `jj diff`, something caused `jj` to materialize its view of the working copy.

You are right that changing a file this way could cause some problems. I was thinking that this might already be happening[^1] and that they are solvable, but it's worth a second look/thought.

[^1]: As you likely well know, `jj` already considers a conflicted file differently from an unconflicted file. If you take an unconflicted file and add conflict markers to it, it will not become conflicted. I thought this could change the file on materialization as well, but on second thought I'm not so sure.

> I think the conflict syntax needs a property that a content without conflicts is identical to the raw data. If the content had some meta data like \JJ line, it would still be considered a conflict.

How would you deal with the `\JJ Verbatim Line:<<<<<<<` situation? Taking your suggestion literally, this would be considered a conflict. If the user does want to keep `<<<<<<<` in the file, they'd have to remove the `\JJ Verbatim Line:` part, but then `jj` can no longer parse the file correctly.

My favorite example for the relevance of this is: what if our `tutorial.md` is conflicted? It contains:

https://github.com/martinvonz/jj/blob/5649ee4f45e02ded9cf6be589df80a590cf056c9/docs/tutorial.md?plain=1#L284-L291

(We need to update it to the current look of the conflict markers still. I also thought of various ways to put a space before those markers, but they are all a little cheap, though if we reject this PR entirely we should probably do something like that)

ilyagr: I suppose another thing to do would be to look at how and whether Mercurial/Git/Bitkeeper/RCS/Darcs deal with this. They have an easier time since conflicts are only very temporary, but they can still have conflict markers inside files or changes in whether there's a newline at the end of file. I might do some spelunking, but you probably know about Mercurial.

ilyagr: There's one more possibility that came to mind (not sure if it's good, but it's interesting):

`jj` could consider files with any `\JJ` conflicted, as you suggested. There could be a command like `jj mark-resolved` that would transfrom such files into unconflicted/normal state. This would normalize a situation like #3223 and provide a command to resolve such conflicts.

We could also use this to simplify the presentation of edit-deletion conflicts. Instead of showing the whole file as a messy conflict, it would be shown normally, but the commit would be considered conflicted until the user decides whether to keep the file and uses `jj mark-resolved`.

This plan would give up on the idea that *every* conflicted file can have its conflicts resolved via editing it in a text editor. That idea will remain true for 95% of conflicts, and `jj` could perhaps guide the user somehow in the other 5% of cases.

One downside is that this could be a lot of work for that 5% of cases, but maybe it's not too bad, I'm unsure. Also, improving the presentation of edit-delete conflicts might make it worth it.

joyously: I don't quite understand when the conflict markers appear in the files.
I just want to toss in an observation from Pijul. I would often get compile errors from conflict markers, and I wasn't even changing any code. I would pull changes and compile ...with errors. This could be due to the unordered nature of Pijul changes (which I don't agree with) or the pull putting markers in or maybe even someone pushed a file with a marker in it.
What I do know is that I should be able to clone the repo at any point and get an accurate set of files. It's the reproducibility of a state that is important for testing and releases.

ilyagr: > If the file is in conflict state, maybe we'll need to represent verbatim <<<<<<< as a fake 1-way conflict (to escape)? 

This is similar to a previous idea I had. I ultimately decided against it because I thought it would look super-messy and would be relatively complicated to implement.

The idea was to allow the opening conflict marker to have more than 7 `<`s. If it has, say, 9 `<`s, then all conflict markers in that conflict are required to be 9 characters long. So, a sequence of 7 `<`s would be considered as just part of the file.

This solves the problem of conflict markers inside conflicts. Then, we'd need some syntax for one-sided conflicts outside conflicts, e.g.

```
<<<<<<<<<
<<<<<<<
aaa
=======
bbb
>>>>>>>
>>>>>>>>>>
```

This has a two issues:

- The parser for conflict markers becomes more complicated. We couldn't parse it with regular expressions, and seemingly not with `pest` as it's not a simple enough grammar. We could do it manually or with another parser combinator library, probably (I forget what the one I'm thinking of is called).

- With the simplest implementation of materialization, the conflict example above would look much more complicated (though this is equivalent):

```
<<<<<<<<<
<<<<<<<
>>>>>>>>>>
aaa
<<<<<<<<<
=======
>>>>>>>>>>
bbb
<<<<<<<<<
>>>>>>>
>>>>>>>>>>
```

Or, we could do a simple greedy implementation that surrounds huge regions of the file with these markers, making the markers hard to find.

So, I ended up leaning away from the idea.

yuja: > The idea was to allow the opening conflict marker to have more than 7 `<`s. If it has, say, 9 `<`s, then all conflict markers in that conflict are required to be 9 characters long.

That seems not bad, and relatively simple to implement.

Let's say any of the base file contents have lines matching `^([<>]{7,})`, we'll use `len(max(matches)) + N` `<`s as conflict marker. When parsing, the (minimum?) number of the `<`s can be determined from the base files, I think.

ilyagr: > When parsing, the (minimum?) number of the <s can be determined from the base files, I think.

I initially forgot about this issue, and it's critical. Relying on the original conflict sides for this is a nice idea. I'm not entirely sure it will work, but it might.

To elaborate for others, for the "long conflict marker" approach to be any better than `\JJ Verbatim Line:`, `jj` must be able to tell the difference between:
  
  - a file that contains a 7-`<` conflict and
  - a file that used to have 9-`<` conflicts and text with embedded 7-`<` conflict markers (like one of the examples in https://github.com/martinvonz/jj/pull/4088#discussion_r1683242306), but now the user removed the conflict markers, and the remaining 7-`<` conflict markers should be part of the text.

The text of the file is identical in the two situations, but `jj` has to know there are no conflicts in the latter file.

ilyagr: I realized that if we get the required number of `<`s from the original conflict, there is no longer a technical need to surround unconflicted conflict markers with artificial "one-sided conflicts". OTOH, I'm worried that with or without those one-sided conflicts, they'll be confusing to the user.

Currently, my best idea is to add something like this to the beginning of the file:

```
<<<<<<<<< Conflict 1 of 5 (Informational, not a real conflict)
\\\\\\\\\ JJ: JJ conflict markers in this file are 9 characters long, longer than the normal 7.
\\\\\\\\\ JJ: This is because this file contains text that looks like conflict markers.
\\\\\\\\\ JJ: This fake conflict can be safely deleted, and must be deleted for the file to be considered resolved.
>>>>>>>>> End of informational conflict 1 of 5

Blah, blah. This wouldn't be a conflict:
<<<<<<<
aaa
=======
bbb
>>>>>>>
bleh bleh
```

yuja: I think, in any case, the user will need to resolve the last remaining conflict in a way that the file content is identical to the non-conflict materialization. Otherwise, the working-copy file would have diffs from the resolved version of the tree content.

I assumed the "minimum length marker" idea practically solves this problem as the user can inline verbatim `<<<<<<<` line if the base file contained verbatim `<<<<<<<` and therefore the marker line must be at least 8-char long.

PR #4092, opened 2024-07-15
diff: add unit tests for copy tracking in the git backend

fowles:
# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_copy_tracking.rs
+++ b/lib/tests/test_copy_tracking.rs
@@ -1,0 +1,120 @@
+// Copyright 2024 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashMap;
+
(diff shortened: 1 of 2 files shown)


PhilipMetzger (approved):

lib/tests/test_copy_tracking.rs
```diff
@@ -0,0 +1,120 @@
...
+) -> Commit {
+    let tree = create_tree(tx.base_repo(), content);
+    tx.mut_repo()
+        .new_commit(settings, parents, tree.id())
+        .write()
+        .unwrap()
+}
+
+#[test]
+fn test_git_detection() {
```
martinvonz, line 68: It would be good to have tests with more than one file in each commit to show that we detect the right copies. It would also be good to have tests where the contents don't match exactly.

fowles: I thought about this, but at some point we are just testing the implementation of gix.  Given that all logic for quality of comparison resides there, I would rather not.

lib/tests/test_copy_tracking.rs
```diff
@@ -0,0 +1,120 @@
...
+    backend: &GitBackend,
+    paths: &[RepoPathBuf],
+    a: &Commit,
+    b: &Commit,
+) -> HashMap<String, Vec<String>> {
+    let stream = backend
+        .get_copy_records(paths, &[a.id().clone()], &[b.id().clone()])
+        .unwrap();
+    let mut res: HashMap<String, Vec<String>> = HashMap::new();
+    for copy_record in block_on_stream(stream).filter_map(|r| r.ok()) {
```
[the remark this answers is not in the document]
fowles: https://github.com/martinvonz/jj/pull/4110

lib/tests/test_copy_tracking.rs
martinvonz: Makes sense.

PR #4093, opened 2024-07-15
cli `branch create/set`: add `--to` aliases to `-r`

ilyagr:
Now that `jj move` does not accept `-r` (since it has `--from` and `--to`), `jj set --to` seems more useful
than `jj set -r`. `create --to` is also added for ease of switching
between `branch create` and `branch move`.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/branch/create.rs
+++ b/cli/src/commands/branch/create.rs
@@ -25,7 +25,10 @@ use crate::ui::Ui;
 #[derive(clap::Args, Clone, Debug)]
 pub struct BranchCreateArgs {
     /// The branch's target revision
-    #[arg(long, short)]
+    //
+    // The `--to` alias exists for making it easier for the user to switch
+    // between `branch create`, `branch move`, and `branch set`.
+    #[arg(long, short, visible_alias = "to")]
     revision: Option<RevisionArg>,
 
     /// The branches to create
--- a/cli/src/commands/branch/set.rs
+++ b/cli/src/commands/branch/set.rs
@@ -25,7 +25,7 @@ use crate::ui::Ui;
 #[derive(clap::Args, Clone, Debug)]
 pub struct BranchSetArgs {
     /// The branch's target revision
-    #[arg(long, short)]
+    #[arg(long, short, visible_alias = "to")]
     revision: Option<RevisionArg>,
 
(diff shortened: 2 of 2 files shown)


yuja:
Adding `--to` alias seems good for compatibility with `move --to`. I don't think `create --at` helps in that regards. `create --to` sounds odd, but it's okay so long as `create -r` is the canonical form?

ilyagr:
I was wondering about that too. I think that, depending on the person, `create --at` might sound much more natural. How about we add both `--at` and `--to` as aliases to `-r` for `create`?

I'm also wondering whether `--to` should be the name and `-r` should be the alias. This makes more sense for `set` than for `create`.

yuja:
I agree "create at" sounds natural, but we don't have any `--at REV` in other commands iirc, so I wouldn't introduce the one. For the same reason, I wouldn't make `set --to` the canonical form. `--to` is usually paired with `--from`.

ilyagr:
OK, updated. `create --to` is definitely good to have. I still might miss `--at` a bit, but we'll see if I remember about it in a month or if anyone misses `--at` enough to ask for it.

cli/src/commands/branch/create.rs
```diff
@@ -25,7 +25,10 @@ use crate::ui::Ui;
 #[derive(clap::Args, Clone, Debug)]
 pub struct BranchCreateArgs {
     /// The branch's target revision
-    #[arg(long, short)]
+    ///
+    /// You can also use `--to` instead of `-r` for ease of switching between
+    /// `branch create`, `branch move`, and `branch set`.
```
yuja, line 30: nit: I wouldn't add this because it's visible alias and user can find it in `branch create --help`.

yuja (approved):
Thanks.

cli/src/commands/branch/create.rs
ilyagr: I felt like `create --to` deserves an explanation/excuse for why we'd go against the rules of English in this way.

yuja: make it non-doc comment then? I agree it's informative for developers.

ilyagr: OK, let's do that.

PR #4095, opened 2024-07-15
Enable the new graph nodes by default

algmyr:
It's been tested in various places now, so this is probably mature enough to be the default.

# Checklist

If applicable:


cli/tests/test_abandon_command.rs
yuja, line 1: Can you squash the second commit so the tests won't break at the first commit?

yuja (approved):
Thanks!

cli/tests/test_abandon_command.rs
algmyr: Will do, I kept it separate mainly to make reviewing a bit less miserable :)

PR #4096, opened 2024-07-15
cargo: build with crt-static on windows

### LINKED ISSUE #4005: Windows binaries require `vcruntime140.dll` to be installed
Windows `.exe` files produced by our CI setup have a dynamic requirement on `vcruntime140.dll`. This is the Visual C++ Runtime, and it comes with Visual Studio. This is used by most Rust-on-Windows developers (because it's a requirement for `rustc`) and I suspect 99.9% of all people compiling or using Jujutsu on Windows will have it installed (because culturally that's the most popular toolchain.)

But there's no real reason that a developer *using* Jujutsu needs that. For instance, if you use an interpreted language like Python or PHP, you can install them and program without ever installing Visual Studio; the things you install will include the Visual C++ Redistributables on your behalf so you don't need to install anything else.

Now we could go the route these tools have and produce installers with the redistributables included, because that's ~practically the only real way to do it while complying with vcruntime's (weird and not well understood) license, and in the future that might be a good idea if we distribute more complex apps (like Tauri ones), but that's a significant amount of work and isn't a blocking requirement for this one issue.

This particular problem can be alleviated with the rustc flag `-Ctarget-feature=+crt-static` which will cause a statically linked copy of the CRT to be included. This is strongly discouraged for `.dll/.lib` outputs for the obvious reasons, but we aren't providing those (and most of our integration design revolving around jj-as-a-server and jj-as-a-Rust-lib helps mitigate the need to produce dynamically linked objects anyway.)

(I discovered this while doing a bootstrapping build of `jj` using Buck2 on a Windows Sandbox and realized I need MSVC before I could even run our produced `jj.exe` binaries. I have also confirmed in the Buck2 branch that the above flag does produce a working binary without the `vcruntime140.dll` dependency.)

thoughtpolice:
Avoids a runtime dependency on vcruntime140.dll.

Closes #4005.

# Checklist

If applicable:

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+# NOTE: on Windows, build with the static CRT, so that produced .exe files don't
+# depend on vcruntime140.dll; otherwise the user requires visual studio if they
+# download a raw .exe
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-Ctarget-feature=+crt-static"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj diff --git` no longer shows the contents of binary files.
 
+* Windows binaries no longer require `vcruntime140.dll` to be installed
+  (normally through Visual Studio.)
+
 ## [0.19.0] - 2024-07-03
 
 ### Breaking changes


thoughtpolice:
Tested with binaries from `main` on my fork: https://github.com/thoughtpolice/jj/actions/runs/9948152804

PhilipMetzger (approved):

PR #3828, opened 2024-06-04
describe: allow updating the description of multiple commits

bnjmnt4n:
If multiple commits are provided, the description of each commit will be combined into a single file for editing.

Closes https://github.com/martinvonz/jj/issues/3120.

# Checklist

If applicable:
- [X] I have updated `CHANGELOG.md`
- [NA] I have updated the documentation (README.md, docs/, demos/)
- [NA] I have updated the config schema (cli/src/config-schema.json)
- [X] I have added tests to cover my changes

Issue #1027, opened 2023-01-14
Option to rebase/amend verbatim (like "ours" merge strategy)


ilyagr:
I think it would be nice for `jj rebase` and `jj amend` to have a `--verbatim-rebase` option that causes the rebase to preserve the contents of the commits being rebased (as opposed to the diff between them and their parents). For `jj rebase`, this flag could simply be `--verbatim`.

In hg and git, I think, this is done via the "ours" merge strategy.

Currently, I don't know a non-awkward way to do this. The best way, I think, is to `jj duplicate` the tip of the rebase, then rebase, and then `jj restore --from duplicate`, and finally `jj abandon duplicate`. Or you can `jj rebase`,  then `jj obslog` and finally `jj restore --from last_obslog_commit`. 

One example when this is useful is when reordering commits. If there's a conflict in the middle, you'd be able to `jj update middle`, resolve the conflict, and then `jj amend --verbatim-rebase`.

ilyagr:
Alternatively, there could be a better way to restore from the pre-rebase version of a rebased commit.

It shouldn't be hard to add a revset operator for "go back in the obslog", but I don't know if that will always do the right thing, especially if the idea from https://github.com/martinvonz/jj/issues/963#issuecomment-1368708868 is implemented.

arxanas:
On the topic of naming, git-branchless calls this [`git amend --reparent`](https://github.com/arxanas/git-branchless/wiki/Command:-git-amend#forcibly-amending-without-changing-children).

ilyagr:
I like the explanation you have in the docs. I'm not sure `--reparent` makes sense to me though. Why that word? The option changes the child, not the parent.

So far, the only other idea that came to my mind is `--freeze-children`.

ilyagr:
Also, I kind of like the word "verbatim". Is it confusing in some way?

martinvonz:
> Why that word?

I've used the word "reparent" before. I used it because the operation changes the (descendants') parent pointers without changing anything else in the commit (well, it does change the commit signature too).

> Also, I kind of like the word "verbatim". Is it confusing in some way?

I wouldn't say it's misleading, it's just not clear to me what it means. It could mean to apply the differences verbatim, although I don't know what that would actually mean.

ilyagr:
I'm not sure I agree, but this makes sense.

Another name this made me think of is `--verbatim-children`. I'm not sure whether that's clearer or more confusing.

Perhaps `jj amend --reparent-children` could also work.

[...]

PR #4097, opened 2024-07-16
cli `split`: add `--restore-from` (AKA `--from`) argument

ilyagr:
Tests are TODO for now, but I wanted to share the idea. I've been wondering for a while what's a sensible UI for this operation, and I think I figured one out.

------------------------

This is useful when you accidentally put some changes in the wrong commit.

In the future, we could add some shortcuts for common uses. For example, we could define `current(X)` to be the current revision with the same change id as `X` (which would usually be a hidden commit) and have a shortcut for `jj split --from X -r current(X)` (only valid if `current(X)` is one commit).

Or, we could have a similar operation for `obscurrent(X)`, defined as `obsheads(obsdescendants(X))`.

 However, let's see how useful the basic operation is first.

# Checklist

If applicable:


cli/src/commands/split.rs
```diff
@@ -51,6 +51,28 @@ pub(crate) struct SplitArgs {
...
+    /// the REVISION revision.
+    ///
+    /// This is especially useful if the FROM revision is a past version of
+    /// REVISION, with its commit id obtained via `jj obslog` or `jj log
+    /// --at-operation`.
+    //
+    // TODO(ilyagr): We could allow `--interactive --from`. It's unclear how
+    // useful that would be. It would mostly require writing tests and
+    // JJ-INSTRUCTIONS. More ambitiously, we could have a 3-pane interactive view
+    // with the FROM commit in the middle and the REVISION commit on the RHS.
```
yuja, line 68: Maybe this point suggests that `split --restore-from` should be a separate command? It's also unclear to me how path arguments will be handled.

If we don't have `unsquash`, I might call this `unsquash` and the first commit would be restored including commit metadata.

ilyagr: I'm currently mulling this over in my head and trying to get some experience of how I use this command in practice and maybe some fresh thoughts.

I'm not sure about much at this point, but here are my current thoughts.

> It's also unclear to me how path arguments will be handled.

To me, this seemed quite clear and is the main reason why I thought this fits into the `split` command: Doing the same thing as the `split` command already does seems to make a lot of sense. 

In the situation form the FAQ answer that I edited in this PR, if you want to restore only *some* of the files from the old version of the commit, you list them in the command line and the command should do the right thing.

For `jj split --restore-from --interactive`, I initially thought things are just as simple (it would let you edit the diff for the FROM commit, make the result of that the intermediate commit, and -- like always with `split` -- make the child commit identical to the -r commit), but there's a question of what to put as the base of the diff if `--from` commit has a different parent than `-r` commit.

> If we don't have unsquash, I might call this unsquash

This is an interesting idea, thanks! This is indeed the operation you'd use to undo a `squash`.

I'm not sure about it, though. In the situation from the FAQ, `unsquash` sounds less natural to me. Also, it depends a little on whether we allow path arguments and what we do with them.

> and the first commit would be restored including commit metadata.

I was wondering about this, but there are a couple of problems. Preserving the metadata would very often cause divergence in the situations where this command would be, I think, most useful. Also, this would be a bit weird if the `--from` commit had a different parent than the `-r` commit.

ilyagr: By the way, a way to fix the issue of "this would be a bit weird if the --from commit had a different parent than the -r commit" with your "preserve metadata" idea would be to rebase the `-r` commit onto the `--from` commit. At this point, the command becomes nearly identical to `rebase --verbatim` that @martinvonz has been mentioning, #1027.

My current opinion is that both `rebase --verbatim` and `split --from` are useful (perhaps under different names, these are just the ones I remember the ideas by), but if we could do everything with one of them, that's fine too.

yuja: Perhaps, I find `split --restore-from` unintuitive because I see `split` is the command to divide the specified commit into two, so the first and the second commits are equally important. If I consider it as a command to insert intermediate state, it makes sense that `--restore-from` "restores" tree into the first (= newly-created) commit.

ilyagr:
My current thoughts about this are:

- This is useful to me, and was just suggested in #4097. Also, #4097 suggested another use-case for this (use-case #2 in there) that I haven't considered before.
- There is one thing I'd like to change, but I am not sure how

So, I think I'll prepare this for merging as is, in the hopes that we might improve it later (unless somebody tells me to wait).

The one thing I'd like to change would be the description of the second commit. For example, for my use-cases, if the `--from` commit has the same description as the original commit, I'd usually want the description of the second commit to be empty. (This would be less useful fro the use-case of swapping commits from #4097). So, I'm wondering about putting the original description and/or the diff in a comment, but I'm unsure what's best.

Perhaps after #3828, we'll have an interface for `split` that edits both commit descriptions in one file, and then we'll have more options and will be able to come back to this.

cli/src/commands/split.rs
ilyagr: I'll call this "resolved", but of course we can continue the conversation.

yuja: To me, `split --restore-from` is unintuitive because it inserts new state (which doesn't exist in the commit to be split.) That said, I'm not strongly against it because the added functionality is useful.

ilyagr:
I think it makes sense to label this option as "Experimental", since I think the way it treats descriptions will change later, and perhaps we find a better approach for these usecases. I'll do that in a second.

cli/src/commands/split.rs
```diff
@@ -54,6 +54,31 @@ pub(crate) struct SplitArgs {
...
+    /// feedback are appreciated.
+    ///
+    /// With this option, the first part of the split will contain the changes
+    /// between the parent of the REVISION and the FROM revision. The second
+    /// part of the split will contain the changes between the FROM revision and
+    /// the REVISION revision.
+    ///
+    /// This is especially useful if the FROM revision is a past version of
+    /// REVISION, with its commit id obtained via `jj obslog` or `jj log
+    /// --at-operation`.
```
martinvonz, line 69: Do we want a note saying that it probably doesn't do what you want if you pass it past version that has since been rebased?

I can see both diff and interdiff behaviors being useful, btw, but the interdiff behavior seems more useful, mostly because the typical use case is what the one described above.

For the regular diff/restore behavior (i.e. what's implemented), you would probably more likely do it based on a commit that's still visible, but then it's unclear why you would want to split the current commit to match a still-visible one.

ilyagr: This is a good question.

In my mind, the main advantage of the regular restore behavior is that it's simpler to explain and understand, it doesn't create conflicts, and it works well in many cases. It is a good point thought that, when the difference does matter, the interdiff behavior might often be preferable (thought I'd guess it's not always).

If there was no cost to switching to interdiff, I'd do it (unless I can think of or remember some other reason not to). The main thing that is giving me pause is: what do we call it? How do we document it?

There is a simple approach to solve this that is not very elegant. It would be to keep this PR as is and add a `--interdiff --restore-from X` behavior. This would be easy to document, but it's a step away from "do what I mean" behavior.

martinvonz: Sorry, I realized later that I think I described the cases poorly as "diff" vs "interdiff". I think a better classification is "snapshot" vs "diff". So what I previously called the "interdiff" mode could be something like `jj split --copy-diff-from=X` or something like that. I think that's a pretty natural way to think of it. What do you think?

Btw, it's implied that the diff is copied to the first commit of the split. I think we can just explain that without having to call the option `jj split --copy-first-commit-diff-from=X` or similar.

yuja: Is it something that `jj rebase -r X --insert-before Y --reparent-descendants` would do?

`jj split --insert-diff-from`/`--insert-snapshot-from` might be better names. I find it's confusing that `jj split --restore-from` inserts a new arbitrary state in between.

ilyagr: > Is it something that jj rebase -r X --insert-before Y --reparent-descendants would do?

In theory, it's similar, though (unless we figure out a better but still consistent behavior) it would create divergence if X was hidden and Y had the same change id.

In practice, I tried this (without `--reparent-descendants`) in the case of `X` being hidden and having children and it's quite weird and buggy. It seems to try to rebase the children of `X`. Moreover, it leads to

```console
$ jj op diff
thread 'main' panicked at /Users/ilyagr/dev/jj/lib/src/dag_walk.rs:116:13:
graph has cycle
From operation 16d658cd9cdf: push all branches to git remote origin
  To operation 1eb8c899850d: rebase commit c9ee0eb96f90b40e88c9a83fb3268bc16ab748f4
```


> jj split `--insert-diff-from`/`--insert-snapshot-from` might be better names.

At a glance, I like these quite a bit, thanks!

yuja: > > Is it something that jj rebase -r X --insert-before Y --reparent-descendants would do?
> 
> In theory, it's similar, though (unless we figure out a better but still consistent behavior) it would create divergence if X was hidden and Y had the same change id.

Perhaps, I should use `jj duplicate` (in place of `rebase`.) 

fwiw, we should fix `op diff` bug separately. I guess it wouldn't be robust against diverged change ids.

ilyagr: > Perhaps, I should use jj duplicate (in place of rebase.)

Indeed, that sounds great to me. `duplicate` makes sense for hidden commits just as well as non-hidden ones.

I'm beginning to think that `rebase` doesn't actually make much sense for hidden commits; we might want to forbid that outright and suggest `duplicate` instead. If a person really, really needs it (and wants to create divergence) they could un-hide the revisions before rebasing.

If `jj duplicate -r X --insert-before Y --reparent-descendants` worked, it would do exactly the same thing as `jj split --insert-snapshot-from`, I think. `jj duplicate -r X --insert-before Y` would correspond to `jj split --insert-diff-from`. I am not sure whether the `split` options would be useful if the `duplicate` worked for this usecase; their only real use would be didactic. Depending on when we implement the `duplicate` functionality, we might want them until then.

I am a bit confused about whether we want `duplicate -s` or `duplicate -b`, but that could be a completely separate discussion. We can make it work, but I think the result is confusing (I should experiment with it a bit).

PR #4118, opened 2024-07-18
copy-tracking: plumb copy tracking through diff detection

fowles:
Re-approach copy tracking with a simpler signature.  Push through end-to-end copy for `diff -s`.

This is not ready to submit, but is complete enough that I would like feedback on the direction.  Remaining things to do:

- [x] update design doc
- [x] update `CHANGELOG.md`
- [x] plumb through all other diff commands
- [x] update the documentation?
- [x] update the config schema? (cli/src/config-schema.json)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* The following diff formats now include information about copies and moves:
+  `--color-words`, `--summary`
+
 ### Fixed bugs
 
 ## [0.20.0] - 2024-08-07
--- a/cli/examples/custom-backend/main.rs
+++ b/cli/examples/custom-backend/main.rs
@@ -178,10 +178,10 @@ impl Backend for JitBackend {
     fn get_copy_records(
         &self,
         paths: &[RepoPathBuf],
-        roots: &[CommitId],
-        heads: &[CommitId],
+        root: &CommitId,
+        head: &CommitId,
     ) -> BackendResult<BoxStream<BackendResult<CopyRecord>>> {
-        self.inner.get_copy_records(paths, roots, heads)
+        self.inner.get_copy_records(paths, root, head)
     }
 
     fn gc(&self, index: &dyn Index, keep_newer: SystemTime) -> BackendResult<()> {
--- a/cli/examples/custom-backend/main.rs
+++ b/cli/examples/custom-backend/main.rs
@@ -177,7 +177,7 @@ impl Backend for JitBackend {
 
     fn get_copy_records(
         &self,
-        paths: &[RepoPathBuf],
+        paths: Option<&[RepoPathBuf]>,
         root: &CommitId,
         head: &CommitId,
     ) -> BackendResult<BoxStream<BackendResult<CopyRecord>>> {
--- a/cli/src/commands/debug/copy_detection.rs
+++ b/cli/src/commands/debug/copy_detection.rs
@@ -43,7 +43,7 @@ pub fn cmd_debug_copy_detection(
     let commit = ws.resolve_single_rev(&args.revision)?;
     for parent_id in commit.parent_ids() {
         for CopyRecord { target, source, .. } in
-            block_on_stream(git.get_copy_records(&[], parent_id, commit.id())?)
+            block_on_stream(git.get_copy_records(None, parent_id, commit.id())?)
                 .filter_map(|r| r.ok())
         {
             writeln!(
--- a/cli/src/commands/debug/copy_detection.rs
+++ b/cli/src/commands/debug/copy_detection.rs
@@ -12,21 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![allow(unused, dead_code)]
-
 use std::fmt::Debug;
 use std::io::Write as _;
-use std::path::{Path, PathBuf};
 
 use futures::executor::block_on_stream;
-use futures::StreamExt;
-use jj_lib::backend::{Backend, CopyRecord, CopySource, CopySources};
-use jj_lib::default_index::{AsCompositeIndex as _, DefaultIndexStore};
-use jj_lib::op_walk;
-use jj_lib::repo_path::{RepoPath, RepoPathBuf};
+use jj_lib::backend::{Backend, CopyRecord};
 
 use crate::cli_util::{CommandHelper, RevisionArg};
-use crate::command_error::{internal_error, user_error, CommandError};
+use crate::command_error::CommandError;
 use crate::ui::Ui;
 
 /// Rebuild commit index
@@ -48,32 +41,18 @@ pub fn cmd_debug_copy_detection(
         return Ok(());
     };
     let commit = ws.resolve_single_rev(&args.revision)?;
-    let tree = commit.tree()?;
-
-    let paths: Vec<RepoPathBuf> = tree.entries().map(|(path, _)| path).collect();
-    let commits = [commit.id().clone()];
-    let parents = commit.parent_ids();
-    for copy_record in
-        block_on_stream(git.get_copy_records(&paths, parents, &commits)?).filter_map(|r| r.ok())
-    {
-        match copy_record.sources {
-            CopySources::Resolved(CopySource { path, .. }) => {
-                write!(ui.stdout(), "{}", path.as_internal_file_string());
-            }
-            CopySources::Conflict(conflicting) => {
-                let mut sorted: Vec<_> = conflicting
-                    .iter()
-                    .map(|s| s.path.as_internal_file_string())
-                    .collect();
-                sorted.sort();
-                write!(ui.stdout(), "{{ {} }}", sorted.join(", "));
-            }
+    for parent_id in commit.parent_ids() {
+        for CopyRecord { target, source, .. } in
+            block_on_stream(git.get_copy_records(&[], parent_id, commit.id())?)
+                .filter_map(|r| r.ok())
+        {
+            writeln!(
+                ui.stdout(),
+                "{} -> {}",
+                source.as_internal_file_string(),
+                target.as_internal_file_string()
+            )?;
         }
-        writeln!(
-            ui.stdout(),
-            " -> {}",
-            copy_record.target.as_internal_file_string()
-        );
     }
     Ok(())
 }
--- a/cli/src/commands/diff.rs
+++ b/cli/src/commands/diff.rs
@@ -14,14 +14,11 @@
 
 use itertools::Itertools;
 use jj_lib::backend::CopyRecords;
-use jj_lib::commit::Commit;
 use jj_lib::repo::Repo;
 use jj_lib::rewrite::merge_commit_trees;
 use tracing::instrument;
 
-use crate::cli_util::{
-    print_unmatched_explicit_paths, CommandHelper, RevisionArg, WorkspaceCommandHelper,
-};
+use crate::cli_util::{print_unmatched_explicit_paths, CommandHelper, RevisionArg};
 use crate::command_error::CommandError;
 use crate::diff_util::DiffFormatArgs;
 use crate::ui::Ui;
--- a/cli/src/commands/diff.rs
+++ b/cli/src/commands/diff.rs
@@ -12,9 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use itertools::Itertools;
+use jj_lib::backend::CopyRecords;
+use jj_lib::commit::Commit;
+use jj_lib::repo::Repo;
+use jj_lib::rewrite::merge_commit_trees;
 use tracing::instrument;
 
-use crate::cli_util::{print_unmatched_explicit_paths, CommandHelper, RevisionArg};
+use crate::cli_util::{
+    print_unmatched_explicit_paths, CommandHelper, RevisionArg, WorkspaceCommandHelper,
+};
 use crate::command_error::CommandError;
 use crate::diff_util::DiffFormatArgs;
 use crate::ui::Ui;
@@ -59,31 +66,50 @@ pub(crate) fn cmd_diff(
     args: &DiffArgs,
 ) -> Result<(), CommandError> {
     let workspace_command = command.workspace_helper(ui)?;
+    let resolve_revision = |r: &Option<RevisionArg>| {
+        workspace_command.resolve_single_rev(r.as_ref().unwrap_or(&RevisionArg::AT))
+    };
+
     let from_tree;
     let to_tree;
+    let mut copy_records = CopyRecords::default();
     if args.from.is_some() || args.to.is_some() {
-        let from =
-            workspace_command.resolve_single_rev(args.from.as_ref().unwrap_or(&RevisionArg::AT))?;
+        let from = resolve_revision(&args.from)?;
+        let to = resolve_revision(&args.to)?;
         from_tree = from.tree()?;
-        let to =
-            workspace_command.resolve_single_rev(args.to.as_ref().unwrap_or(&RevisionArg::AT))?;
         to_tree = to.tree()?;
+
+        copy_records.add_records(workspace_command.repo().store().get_copy_records(
+            None,
+            from.id(),
+            to.id(),
+        )?)?;
     } else {
-        let commit = workspace_command
-            .resolve_single_rev(args.revision.as_ref().unwrap_or(&RevisionArg::AT))?;
-        from_tree = commit.parent_tree(workspace_command.repo().as_ref())?;
-        to_tree = commit.tree()?
+        let to = resolve_revision(&args.revision)?;
+        let parents: Vec<_> = to.parents().try_collect()?;
+        from_tree = merge_commit_trees(workspace_command.repo().as_ref(), &parents)?;
+        to_tree = to.tree()?;
+
+        for p in &parents {
+            copy_records.add_records(workspace_command.repo().store().get_copy_records(
+                None,
+                p.id(),
+                to.id(),
+            )?)?;
+        }
     }
+
+    let diff_renderer = workspace_command.diff_renderer_for(&args.format)?;
     let fileset_expression = workspace_command.parse_file_patterns(&args.paths)?;
     let matcher = fileset_expression.to_matcher();
-    let diff_renderer = workspace_command.diff_renderer_for(&args.format)?;
     ui.request_pager();
     diff_renderer.show_diff(
         ui,
         ui.stdout_formatter().as_mut(),
         &from_tree,
         &to_tree,
-        matcher.as_ref(),
+        &matcher,
+        &copy_records,
         ui.term_width(),
     )?;
     print_unmatched_explicit_paths(
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -175,6 +175,7 @@ pub(crate) fn cmd_fix(
         let mut diff_stream = parent_tree.diff_stream(&tree, &matcher);
         async {
             while let Some(TreeDiffEntry {
+                source: _, // TODO handle copy tracking
                 target: repo_path,
                 value: diff,
             }) = diff_stream.next().await
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -22,7 +22,7 @@ use itertools::Itertools;
 use jj_lib::backend::{BackendError, CommitId, FileId, TreeValue};
 use jj_lib::fileset::{self, FilesetExpression};
 use jj_lib::matchers::{EverythingMatcher, Matcher};
-use jj_lib::merged_tree::MergedTreeBuilder;
+use jj_lib::merged_tree::{MergedTreeBuilder, TreeDiffEntry};
 use jj_lib::repo::Repo;
 use jj_lib::repo_path::{RepoPathBuf, RepoPathUiConverter};
 use jj_lib::revset::{RevsetExpression, RevsetIteratorExt};
@@ -174,7 +174,11 @@ pub(crate) fn cmd_fix(
         let parent_tree = commit.parent_tree(tx.repo())?;
         let mut diff_stream = parent_tree.diff_stream(&tree, &matcher);
         async {
-            while let Some((repo_path, diff)) = diff_stream.next().await {
+            while let Some(TreeDiffEntry {
+                target: repo_path,
+                value: diff,
+            }) = diff_stream.next().await
+            {
                 let (_before, after) = diff?;
                 // Deleted files have no file content to fix, and they have no terms in `after`,
                 // so we don't add any tool inputs for them. Conflicted files produce one tool
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -172,7 +172,8 @@ pub(crate) fn cmd_fix(
         // Also fix any new paths that were changed in this commit.
         let tree = commit.tree()?;
         let parent_tree = commit.parent_tree(tx.repo())?;
-        let mut diff_stream = parent_tree.diff_stream(&tree, &matcher);
+        let copy_records = Default::default();
+        let mut diff_stream = parent_tree.diff_stream(&tree, &matcher, &copy_records);
         async {
             while let Some(TreeDiffEntry {
                 source: _, // TODO handle copy tracking
--- a/cli/src/commit_templater.rs
+++ b/cli/src/commit_templater.rs
@@ -19,7 +19,7 @@ use std::io;
 use std::rc::Rc;
 
 use itertools::Itertools as _;
-use jj_lib::backend::{BackendResult, ChangeId, CommitId};
+use jj_lib::backend::{BackendResult, ChangeId, CommitId, CopyRecords};
 use jj_lib::commit::Commit;
 use jj_lib::extensions_map::ExtensionsMap;
 use jj_lib::fileset::{self, FilesetExpression};
@@ -1278,6 +1278,7 @@ pub struct TreeDiff {
     from_tree: MergedTree,
     to_tree: MergedTree,
     matcher: Rc<dyn Matcher>,
+    copy_records: CopyRecords,
 }
 
 impl TreeDiff {
@@ -1290,11 +1291,13 @@ impl TreeDiff {
             from_tree: commit.parent_tree(repo)?,
             to_tree: commit.tree()?,
             matcher,
+            copy_records: Default::default(),
         })
     }
 
     fn diff_stream(&self) -> TreeDiffStream<'_> {
-        self.from_tree.diff_stream(&self.to_tree, &*self.matcher)
+        self.from_tree
+            .diff_stream(&self.to_tree, &*self.matcher, &self.copy_records)
     }
 
     fn into_formatted<F, E>(self, show: F) -> TreeDiffFormatted<F>
--- a/cli/src/commit_templater.rs
+++ b/cli/src/commit_templater.rs
@@ -1394,8 +1394,15 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
             let path_converter = language.path_converter;
             let template = self_property
                 .map(move |diff| {
+                    let to_tree = diff.to_tree.clone();
                     diff.into_formatted(move |formatter, _store, tree_diff| {
-                        diff_util::show_diff_summary(formatter, tree_diff, path_converter)
+                        diff_util::show_diff_summary(
+                            formatter,
+                            tree_diff,
+                            path_converter,
+                            &Default::default(),
+                            &to_tree,
+                        )
                     })
                 })
                 .into_template();
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -20,7 +20,7 @@ use std::{io, mem};
 
 use futures::StreamExt;
 use itertools::Itertools;
-use jj_lib::backend::{BackendError, TreeValue};
+use jj_lib::backend::{BackendError, CopyRecords, TreeValue};
 use jj_lib::commit::Commit;
 use jj_lib::conflicts::{materialized_diff_stream, MaterializedTreeValue};
 use jj_lib::diff::{Diff, DiffHunk};
@@ -245,10 +245,19 @@ impl<'a> DiffRenderer<'a> {
         from_tree: &MergedTree,
         to_tree: &MergedTree,
         matcher: &dyn Matcher,
+        copy_records: &CopyRecords,
         width: usize,
     ) -> Result<(), DiffRenderError> {
         formatter.with_label("diff", |formatter| {
-            self.show_diff_inner(ui, formatter, from_tree, to_tree, matcher, width)
+            self.show_diff_inner(
+                ui,
+                formatter,
+                from_tree,
+                to_tree,
+                matcher,
+                copy_records,
+                width,
+            )
         })
     }
 
@@ -259,6 +268,7 @@ impl<'a> DiffRenderer<'a> {
         from_tree: &MergedTree,
         to_tree: &MergedTree,
         matcher: &dyn Matcher,
+        _copy_records: &CopyRecords,
         width: usize,
     ) -> Result<(), DiffRenderError> {
         let store = self.repo.store();
@@ -324,7 +334,15 @@ impl<'a> DiffRenderer<'a> {
     ) -> Result<(), DiffRenderError> {
         let from_tree = commit.parent_tree(self.repo)?;
         let to_tree = commit.tree()?;
-        self.show_diff(ui, formatter, &from_tree, &to_tree, matcher, width)
+        self.show_diff(
+            ui,
+            formatter,
+            &from_tree,
+            &to_tree,
+            matcher,
+            &Default::default(),
+            width,
+        )
     }
 }
 
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -22,7 +22,9 @@ use futures::StreamExt;
 use itertools::Itertools;
 use jj_lib::backend::{BackendError, CopyRecords, TreeValue};
 use jj_lib::commit::Commit;
-use jj_lib::conflicts::{materialized_diff_stream, MaterializedTreeValue};
+use jj_lib::conflicts::{
+    materialized_diff_stream, MaterializedTreeDiffEntry, MaterializedTreeValue,
+};
 use jj_lib::diff::{Diff, DiffHunk};
 use jj_lib::files::DiffLine;
 use jj_lib::matchers::Matcher;
@@ -567,7 +569,12 @@ pub fn show_color_words_diff(
 ) -> Result<(), DiffRenderError> {
     let mut diff_stream = materialized_diff_stream(store, tree_diff);
     async {
-        while let Some((path, diff)) = diff_stream.next().await {
+        while let Some(MaterializedTreeDiffEntry {
+            source: _, // TODO handle copy tracking
+            target: path,
+            value: diff,
+        }) = diff_stream.next().await
+        {
             let ui_path = path_converter.format_file_path(&path);
             let (left_value, right_value) = diff?;
 
@@ -714,7 +721,12 @@ pub fn show_file_by_file_diff(
     let right_wc_dir = temp_dir.path().join("right");
     let mut diff_stream = materialized_diff_stream(store, tree_diff);
     async {
-        while let Some((path, diff)) = diff_stream.next().await {
+        while let Some(MaterializedTreeDiffEntry {
+            source: _, // TODO handle copy tracking
+            target: path,
+            value: diff,
+        }) = diff_stream.next().await
+        {
             let ui_path = path_converter.format_file_path(&path);
             let (left_value, right_value) = diff?;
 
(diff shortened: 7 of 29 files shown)


cli/src/diff_util.rs
```diff
@@ -237,6 +272,46 @@ impl<'a> DiffRenderer<'a> {
...
+        )
+    }
+
+    /// Generates diff between two commits.
+    pub fn show_diff_commits(
+        &self,
+        ui: &Ui, // TODO: remove Ui dependency if possible
+        formatter: &mut dyn Formatter,
+        from: &Commit,
+        to: &Commit,
```
yuja, line 301: Perhaps, `from` has to be `&[Commit]`?

It might make sense to special case `from.len() == 1`, but we'll eventually need to handle copy info in a merge commit properly.

yuja (commented):
I can't comment on copy-tracing API changes, so just about diff_util part.

I think it's good to use `Commit` or `[Commit]` in place of `MergedTree`. I'm trying to implement commit description templates, and `show_diff()` callers in description_util.rs will be switched to be using a `Commit` object. It may be either a temporary `Commit` physically written to the store, or in-memory `Commit` with fake commit id and physically-stored tree and parents.

cli/src/diff_util.rs
fowles: I think you are correct, but I have found it very helpful for understanding how the parts fit together to aggressively simplify.  I believe that after I do a bit more down this path, I will be able to make the CopyRecordsMap thing handle multiple parents before long (at which point the concept of a conflict will reappear but in a more restricted context (I think)

fowles:
> I can't comment on copy-tracing API changes, so just about diff_util part.
> 
> I think it's good to use `Commit` or `[Commit]` in place of `MergedTree`. I'm trying to implement commit description templates, and `show_diff()` callers in description_util.rs will be switched to be using a `Commit` object. It may be either a temporary `Commit` physically written to the store, or in-memory `Commit` with fake commit id and physically-stored tree and parents.

This is great!  That will fit together much better.

cli/src/diff_util.rs
fowles: we now handle this better, but there is still a TODO for conflicts on the merge

martinvonz (commented):
Just one little comment for now. I'll have to review it in detail later

cli/examples/custom-backend/main.rs
```diff
@@ -175,13 +176,13 @@ impl Backend for JitBackend {
...
 
-    fn get_copy_records(
+    fn get_copy_records<'matcher>(
         &self,
-        paths: &[RepoPathBuf],
-        roots: &[CommitId],
-        heads: &[CommitId],
-    ) -> BackendResult<BoxStream<BackendResult<CopyRecord>>> {
-        self.inner.get_copy_records(paths, roots, heads)
+        matcher: &'matcher dyn Matcher,
```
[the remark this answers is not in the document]
fowles: How about simply removing the filter piece altogether until we have a concrete use case for it?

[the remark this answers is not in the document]
martinvonz: We do have a use case for it. When rebasing a commit that touches a few files across tens of thousands of upstream commits (e.g. at Google), it's wasteful to find copies affecting unrelated files.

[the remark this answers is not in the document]
fowles: alright, I can make that change

cli/examples/custom-backend/main.rs
PhilipMetzger, line 1: Nit: This should be the first commit but that probably generates a bunch of conflicts on your side.

PhilipMetzger (commented):
The things I mentioned in Discord.

cli/src/merge_tools/diff_working_copies.rs
```diff
@@ -131,7 +131,7 @@ pub(crate) fn check_out_trees(
     output_is: Option<DiffSide>,
 ) -> Result<DiffWorkingCopies, DiffCheckoutError> {
     let changed_files: Vec<_> = left_tree
-        .diff_stream(right_tree, matcher)
+        .diff_stream(right_tree, matcher, &Default::default())
```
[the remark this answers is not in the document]
fowles: I kinda lean against since it gives us a clear point where we need to add tracking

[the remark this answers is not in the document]
PhilipMetzger: That makes sense and its a small thing anyway.

cli/src/diff_util.rs
```diff
@@ -1156,27 +1181,71 @@ pub fn show_diff_summary(
...
+                writeln!(formatter.labeled("copied"), "C {before} => {after}")?
+            }
+            Type::Rename(before, after) => {
+                writeln!(formatter.labeled("renamed"), "R {before} => {after}")?
+            }
+            Type::Modification(path) => writeln!(formatter.labeled("modified"), "M {path}")?,
+            Type::Addition(path) => writeln!(formatter.labeled("added"), "A {path}")?,
+            Type::Deletion(path) => writeln!(formatter.labeled("removed"), "D {path}")?,
+        }
+    }
```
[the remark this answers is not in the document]
fowles: after all the surgery, this comment is hard to find in git cause of "Outdated" paths, so I am going to resolve it.

cli/src/diff_util.rs
```diff
@@ -572,20 +579,28 @@ pub fn show_color_words_diff(
...
-                        "Access denied to {ui_path}:"
+                        "Access denied to {left_ui_path}:"
+                    )?;
+                    writeln!(formatter, " {source}")?;
+                    continue;
+                }
+                (_, MaterializedTreeValue::AccessDenied(source)) => {
+                    write!(
+                        formatter.labeled("access-denied"),
+                        "Access denied to {right_ui_path}:"
```
martinvonz: Wasn't the point of pass an empty `CopyRecordMap` on line 308 so we don't have to worry about the `source` field here? I.e., can we undo the changes to this function?

lib/src/merged_tree.rs
```diff
@@ -659,56 +684,8 @@ impl<'matcher> TreeDiffIterator<'matcher> {
...
-        entries.reverse();
-        Self { entries }
-    }
-}
-
-impl Iterator for TreeDiffIterator<'_> {
-    type Item = TreeDiffEntry;
-
-    fn next(&mut self) -> Option<Self::Item> {
+    fn next_impl(&mut self) -> Option<TreeDiffEntry> {
```
martinvonz, line 688: nit: could you do the refactoring of defining `next_impl()` in a separate commit? thanks

lib/src/merged_tree.rs
```diff
@@ -791,10 +768,64 @@ impl Iterator for TreeDiffIterator<'_> {
...
+}
+
+impl Iterator for TreeDiffIterator<'_> {
+    type Item = TreeDiffEntry;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.next_impl()
+            .map(|e| e.adjust_for_copy_tracking(&self.source, self.copy_records))
+    }
+}
```
martinvonz: Could you add some tests in `test_merged_tree.rs`? Please also include tests for things that are not handled well yet and just leave a TODO in the test to say that we need to fix it later.

In case you're wondering what the advantages are compared to CLI tests:

* Can add comprehensive tests and then not need so comprehensive tests for each command that displays diffs (or not have to arbitrarily decide which of the commands should get comprehensive tests)
* Easier to produce corner cases (for example, you can pass in a `CopyRecordMap` you can't trigger via Git
* Faster

lib/src/merged_tree.rs
```diff
@@ -1059,6 +1090,16 @@ impl Stream for TreeDiffStreamImpl<'_> {
     }
 }
 
+impl Stream for TreeDiffStreamImpl<'_> {
+    type Item = TreeDiffEntry;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.as_mut()
+            .poll_next_impl(cx)
+            .map(|o| o.map(|e| e.adjust_for_copy_tracking(&self.source_tree, self.copy_records)))
```
martinvonz, line 1099: nit: could you use some clearer names than `o` and `e` here?

cli/src/diff_util.rs
```diff
@@ -1128,27 +1153,73 @@ pub fn show_diff_summary(
...
+    enum Type {
+        Copy(String, String),
+        Rename(String, String),
+        Modification(String),
+        Addition(String),
+        Deletion(String),
+    }
+
+    let mut file_modifications = Vec::<Type>::new();
+    let mut unresolved_renames = HashMap::<String, usize>::new();
```
martinvonz, line 1166: Before this commit, `jj diff -s -r <huge commit>` would start producing output as soon as possible, but now we need to iterate through the whole diff before we produce any output. The simplest solution is to simply not distinguish between `C` and `R`. I think that's what Mercurial does. If we/you care enough to distinguish them, perhaps we could look ahead in the target tree but that's more expensive and more work, so I would personally leave that for later.

FYI, this is especially unfortunate at Google where it can take a while to produce a large diff of uncached trees. I think it's especially useful to produce output incrementally as a hint that users are doing something they didn't intend when they accidentally produce a large diff (e.g. `jj diff --from cl/12345 --to cl/123456` when they meant `--from cl/123450` or something).

cli/src/diff_util.rs
```diff
@@ -1128,27 +1153,73 @@ pub fn show_diff_summary(
...
+                                file_modifications[idx] =
+                                    Type::Rename(before.clone(), after.clone());
+                            }
+                        } else {
+                            unresolved_renames
+                                .insert(before_ui_path.clone(), file_modifications.len());
+                            file_modifications.push(Type::Deletion(before_ui_path));
+                        }
+                    }
+                    _ => {}
```
martinvonz, line 1202: nit: make this explicitly `(false, false)` and make it call `unreachable!()` to clarify that it shouldn't happen (right?)

lib/src/merged_tree.rs
```diff
@@ -317,7 +317,8 @@ impl MergedTree {
 
 /// A single entry in a tree diff.
 pub struct TreeDiffEntry {
-    // pub source: RepoPathBuf,
+    /// The source path.
+    pub source: RepoPathBuf,
```
martinvonz, line 324: I wonder if it can be measurable slower to have this field here when it's typically the same as `target`. Maybe we should make this `Option<RepoPathBuf>` and not set it if it's the same as `target`. Just a thought. Let's leave it as is until we've noticed that it makes a difference.

lib/src/backend.rs
```diff
@@ -458,7 +444,8 @@ pub trait Backend: Send + Sync + Debug {
         sign_with: Option<&mut SigningFn>,
     ) -> BackendResult<(CommitId, Commit)>;
 
-    /// Get all copy records for `paths` in the dag range `roots..heads`.
+    /// Get copy records for the dag range `root..head`.  If `paths` is empty
+    /// include all paths, otherwise restrict to only `paths`.
```
yuja, line 485: nit: use `Option<&[RepoPathBuf]>`? It can be source of bugs if `[]` had a different meaning.

cli/src/commands/diff.rs
```diff
@@ -52,38 +59,60 @@ pub(crate) struct DiffArgs {
...
-        to_tree = commit.tree()?
+        let to = resolve_revision(&workspace_command, &args.revision)?;
+        let parents: Vec<_> = to.parents().try_collect()?;
+        for p in &parents {
+            copy_records.add_records(workspace_command.repo().store().get_copy_records(
+                &[],
+                p.id(),
+                to.id(),
+            )?)?;
+        }
```
yuja, line 103: Out of curiosity, how copy records are stored in a merge commit? This API suggests that the records are identified by `(parent_id, self_id)` even if `self.parents()[i] == parent`.

Suppose a file is renamed at one side of the merge,
```
D
|\
C |  rename "foo" -> "bar", modify "bar"
| |
| B  modify "foo"
: :
|/
A
```
when creating the merge commit `D` by `jj new C B`, auto-merge function would gather copy records of `A::C` and `A::B`, then merge file `D["bar"] = C["bar"] - A["foo"] + B["foo"]`. I'm not sure what would be recorded at `D`. If the auto-merged tree was built considering renames, the commit `D` would have no copy records (unless user made another rename in addition)?

If the backend doesn't have persistent copy records, I guess both `trace_copy(C to D)` and `trace_copy(B to D)` would report `"foo" -> "bar"`.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj diff` the following subcommands now include information about copies and
+  moves if supported by the backend: `jj diff --summary`
```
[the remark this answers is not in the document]
fowles: done

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+
```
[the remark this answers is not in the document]
fowles: done

cli/src/commands/diff.rs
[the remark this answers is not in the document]
fowles: done

cli/src/commands/diff.rs
```diff
@@ -52,38 +59,60 @@ pub(crate) struct DiffArgs {
...
+    let matcher = fileset_expression.to_matcher();
+
     let to_tree;
-    if args.from.is_some() || args.to.is_some() {
-        let from =
-            workspace_command.resolve_single_rev(args.from.as_ref().unwrap_or(&RevisionArg::AT))?;
-        from_tree = from.tree()?;
-        let to =
-            workspace_command.resolve_single_rev(args.to.as_ref().unwrap_or(&RevisionArg::AT))?;
+    let from_tree;
```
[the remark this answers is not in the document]
fowles: it isn't load bearing, but it does make the commit easier to read in my opinion, so I would prefer to leave it

cli/src/commands/diff.rs
fowles: this is currently left to `CopyRecordMap` to handle conflicts (which discards both copy traces).  Unfortunately, this is almost purely theoretical as git doesn't keep copy traces but infers them dynamically based on content.  I think that properly solving conflicting copies is going to have to be a follow on thing handled by Google folks, since their backend is the only one that can manifest them.

cli/src/commands/fix.rs
```diff
@@ -175,6 +175,7 @@ pub(crate) fn cmd_fix(
         let mut diff_stream = parent_tree.diff_stream(&tree, &matcher);
         async {
             while let Some(TreeDiffEntry {
+                source: _,
```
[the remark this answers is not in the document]
fowles: added TODOs

lib/src/backend.rs
```diff
@@ -180,6 +181,42 @@ pub struct CopyRecord {
     pub source_commit: CommitId,
 }
 
+/// A collection of CopyRecords.
+#[derive(Default, Debug)]
+pub struct CopyRecordMap {
+    records: Vec<CopyRecord>,
+    targets: HashMap<RepoPathBuf, usize>,
```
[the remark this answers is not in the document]
fowles: originally, this was because I also had a reverse map from sources.  Now I leave it because this allows me to keep the record while removing it from the conflict list.  I strongly suspect it will continue to need this indirection.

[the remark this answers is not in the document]
fowles: done

lib/src/backend.rs
```diff
@@ -180,6 +181,42 @@ pub struct CopyRecord {
     pub source_commit: CommitId,
 }
 
+/// A collection of CopyRecords.
+#[derive(Default, Debug)]
+pub struct CopyRecordMap {
```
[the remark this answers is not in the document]
fowles: done

lib/src/backend.rs
fowles: done

lib/src/merged_tree.rs
fowles: while that was the original design, it makes interacting with the thing more cumbersome, so I went with the simplest thing I could for this rework.

cli/src/commands/diff.rs
martinvonz: > I'm not sure what would be recorded at `D`.

Good question! We just don't know yet (or at least I don't know yet). I'm going to chat with @ilyagr and @torquestomp about this next week. We'll need to be able to reproduce the parent tree somehow. Maybe that means we'll need to store the copy information from `A::C` and `A::B`. We will also want to be able to represent conflicts in the copy information state itself (e.g. rename `foo -> bar` on one side and `foo -> baz` on the other side). And we'll want to be able to back out such a commit. Lots of unanswered questions :(

lib/src/merged_tree.rs
fowles: done

lib/src/merged_tree.rs
fowles: done

lib/tests/test_commit_builder.rs
```diff
@@ -27,10 +27,16 @@ use testutils::{assert_rebased_onto, create_tree, CommitGraphBuilder, TestRepo,
...
+        .map(
+            |TreeDiffEntry {
+                 target: path,
+                 value: diff,
+                 ..
+             }| {
+                let _ = diff.unwrap();
+                path
+            },
+        )
```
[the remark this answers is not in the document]
fowles: done

lib/tests/test_merged_tree.rs
```diff
@@ -36,12 +37,22 @@ fn file_value(file_id: &FileId) -> TreeValue {
 
 fn diff_stream_equals_iter(tree1: &MergedTree, tree2: &MergedTree, matcher: &dyn Matcher) {
     let iter_diff: Vec<_> = TreeDiffIterator::new(tree1.as_merge(), tree2.as_merge(), matcher)
-        .map(|(path, diff)| (path, diff.unwrap()))
+        .map(
+            |TreeDiffEntry {
+                 target: path,
+                 value: diff,
+             }| (path, diff.unwrap()),
```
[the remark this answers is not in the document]
fowles: yes, forcing revisiting the fields when I added `source` was really important to maintaining integrity during this set of changes, but I think the suggestion for a helper in this test is a good one

cli/src/diff_util.rs
fowles: `jj diff --stat` already has exactly the behavior you are talking about of walking everything before producing output, and it seems to do fine with it.  We can drop the thing that distinguishes 'C' from 'R' but if feels nice to match git's capabilities here.

cli/src/diff_util.rs
fowles: done

cli/src/commands/diff.rs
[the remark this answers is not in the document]
fowles: I did reorder things to be from,to consistently

lib/src/merged_tree.rs
fowles: I have added some tests in `test_merged_tree.rs`.  There aren't really many corner cases that can be triggered at this level.  The current api of `CopyRecords` doesn't allow conflicts and those are the cases that have corners and need to be designed.

cli/src/diff_util.rs
```diff
@@ -567,7 +569,12 @@ pub fn show_color_words_diff(
 ) -> Result<(), DiffRenderError> {
     let mut diff_stream = materialized_diff_stream(store, tree_diff);
     async {
-        while let Some((path, diff)) = diff_stream.next().await {
+        while let Some(MaterializedTreeDiffEntry {
+            source: _,
```
martinvonz: nit: add TODO about using this value (applies throughout this commit)

lib/src/merged_tree.rs
```diff
@@ -607,7 +607,6 @@ pub struct TreeDiffIterator<'matcher> {
     store: Arc<Store>,
     stack: Vec<TreeDiffItem>,
     matcher: &'matcher dyn Matcher,
-    copy_records: &'matcher CopyRecords,
```
martinvonz: This field was added in one commit, then removed in the next, and then added back again. So keep it in the middle commit, i guess.

cli/src/diff_util.rs
martinvonz: Yes, `jj diff --stat` does work at Google, it's just that it's silent for a long time if the diff is large. It gets extra slow because it needs to calculate the file diffs too. To be fair, it might be that finding copies between two commits far apart will also be slow, so it might still take a bit of time even if we don't consume the whole tree diff before we produce output.

If you don't mind, I would prefer to see it implemented like this:

1. Find copies
2. For each source path, check in `to_tree` if it's no longer there and then add it to a `deleted_sources` set
3. Walk tree diff and use `deleted_sources` to determine if a copy target is actually a rename

cli/src/diff_util.rs
```diff
@@ -1110,7 +1110,11 @@ pub fn show_diff_summary(
     path_converter: &RepoPathUiConverter,
 ) -> io::Result<()> {
     async {
-        while let Some((repo_path, diff)) = tree_diff.next().await {
+        while let Some(TreeDiffEntry {
+            target: repo_path,
+            value: diff,
```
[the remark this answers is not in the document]
fowles: I really don't like `diff` as the name for the view because it fundamentally is a pair of values (well `MaterializedTreeValue`).  I would rather rename it to `values` or `tree_values` and use that locally.  I mean, my strongest preference is to split it into `source_value` and `target_value`, but I was not able to accomplish that refactor despite three attempts

lib/src/merged_tree.rs
fowles: removed from this comment, so it isn't here until it is used

cli/src/diff_util.rs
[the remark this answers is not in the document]
martinvonz: I actually think `diff` is a good name for a pair of values :) But I won't insist, and maybe that refactoring you mention  will happen eventually anyway.

lib/src/merged_tree.rs
```diff
@@ -328,6 +328,29 @@ pub struct TreeDiffEntry {
...
+
+        Self {
+            source: source.clone(),
+            target: self.target,
+            value: match self.value {
+                Ok((_, target_value)) => source_tree
+                    .path_value(source)
+                    .map(|source_value| (source_value, target_value)),
+                Err(e) => Err(e),
+            },
```
martinvonz, line 349: nit: cam we calculate the new values before we construct `Self` and thereby avoid cloning `source`?

lib/src/merged_tree.rs
```diff
@@ -328,6 +328,29 @@ pub struct TreeDiffEntry {
...
+        };
+
+        Self {
+            source: source.clone(),
+            target: self.target,
+            value: match self.value {
+                Ok((_, target_value)) => source_tree
+                    .path_value(source)
+                    .map(|source_value| (source_value, target_value)),
+                Err(e) => Err(e),
```
martinvonz, line 348: nit: can use `.map()` and avoid the `match`

cli/src/merge_tools/diff_working_copies.rs
```diff
@@ -132,7 +132,7 @@ pub(crate) fn check_out_trees(
 ) -> Result<DiffWorkingCopies, DiffCheckoutError> {
     let changed_files: Vec<_> = left_tree
         .diff_stream(right_tree, matcher)
-        .map(|(path, _diff)| path)
+        .map(|TreeDiffEntry { target, .. }| target)
```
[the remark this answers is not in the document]
fowles: we don't need to since the `CopyRecords` are constructed directly here and guaranteed to be empty.  We would have to pass them in from an external source to make it work here at which point we are unlikely to forget anything since we are so close to this line.

cli/src/diff_util.rs
fowles: done

cli/src/commands/diff.rs
yuja: I wonder if this was the reason that `roots` in `get_copy_records()` was a list. For a linear history `A - B - C`, `get_copy_records(A, C)` would accumualte copy records of `B` and `C`. `A`'s records are not included because `A` is the base tree. Given that, for a merge `{D, E} - F`, both `get_copy_records(D, F)` and `get_copy_records(E, F)` would return the copy records of `F`, and should be a single query `get_copy_records([D, E], F)`?

lib/src/merged_tree.rs
fowles: no, we have to clone source because it is a borrow from the CopyRecords

lib/src/merged_tree.rs
fowles: oh nice.  I have to use `and_then`, but yes.

cli/src/diff_util.rs
fowles: Fixed!

cli/src/diff_util.rs
```diff
@@ -284,29 +284,36 @@ impl<'a> DiffRenderer<'a> {
                     show_diff_summary(formatter, tree_diff, path_converter)?;
                 }
                 DiffFormat::Stat => {
-                    let tree_diff = from_tree.diff_stream(to_tree, matcher, copy_records);
+                    let no_copy_tracking = Default::default();
+                    let tree_diff = from_tree.diff_stream(to_tree, matcher, &no_copy_tracking);
                     show_diff_stat(formatter, store, tree_diff, path_converter, width)?;
```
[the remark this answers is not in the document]
fowles: I am not 100% clear on what you are asking for.  I have moved the `no_copy_tracking` things back one change, cause I think that is what you wanted.

cli/src/diff_util.rs
fowles: oh, I forgot that I had implemented this one.  I fixed line 308 and updated the CHANGELOG.

cli/src/commands/diff.rs
fowles: I am going to resolve this conversation as out of scope for this PR.  Hopefully folks can figure out the details next week and build on this though!

yuja: Oh, I thought this would matter whether the first commit of this PR makes sense or not.

Maybe https://github.com/martinvonz/jj/pull/4235 is a better place to discuss this kind of things, but appears that #4235 is a blocker of this PR.

martinvonz: I also don't think the first commit really makes sense, but there's a lot of work after it that I don't want to block. I'm also not sure how the API should look. I.e., I'm not sure that the state before the first commit was right either. I agree with you that the reason it was a list was probably for merge commits, but I'm still not sure how to interpret it. I think @torquestomp and I have been out at different times so I haven't gotten a chance to ask him about it either.

And I would have preferred for #4235 to have been part of the first commit, btw.

cli/src/diff_util.rs
```diff
@@ -1135,28 +1151,57 @@ pub fn show_diff_summary(
...
             value: diff,
         }) = tree_diff.next().await
         {
             let (before, after) = diff.unwrap();
-            let ui_path = path_converter.format_file_path(&repo_path);
-            if before.is_present() && after.is_present() {
-                writeln!(formatter.labeled("modified"), "M {ui_path}")?;
-            } else if before.is_absent() {
-                writeln!(formatter.labeled("added"), "A {ui_path}")?;
+            let before_ui_path = path_converter.format_file_path(&before_path);
```
martinvonz, line 1170: nit: move into if-block below so we avoid calculating it when it won't be needed (as it won't be most of time)

cli/src/diff_util.rs
```diff
@@ -1135,28 +1151,57 @@ pub fn show_diff_summary(
...
+            let after_ui_path = path_converter.format_file_path(&after_path);
+            if before_path != after_path {
+                if to_tree
+                    .path_value(&before_path)
+                    .map_or(false, |v| v.is_absent())
```
martinvonz, line 1175: nit: I would prefer `.unwrap().is_absent()` to `map_or(false, ...)` so it's easier to notice later that we should fix this line (like line 1169 above)

martinvonz (approved):
Looks good to me, minus the few remaining nits.

But please give @yuja another chance to comment too.

@yuja: As I said elsewhere on this PR, I don't think the API changes in the first commit are quite right. However, I think the rest of the series looks correct, and it provides practical improvements for Git users. I therefore think it makes sense to get this in in its current form, even if we know we're going to have change the API later. I don't think this PR makes it significantly harder to make those changes later, whatever they are - the changes to the diff iterator/stream and the commands using them see like they probably won't change significantly. Let us know if you disagree, of course.

martinvonz:
And thanks a lot for your hard work and patience with this PR, @fowles!

cli/src/commands/debug/copy_detection.rs
```diff
@@ -20,8 +20,9 @@ use std::path::{Path, PathBuf};
 
 use futures::executor::block_on_stream;
 use futures::StreamExt;
-use jj_lib::backend::{Backend, CopyRecord, CopySource, CopySources};
+use jj_lib::backend::{Backend, CopyRecord};
 use jj_lib::default_index::{AsCompositeIndex as _, DefaultIndexStore};
+use jj_lib::matchers::EverythingMatcher;
```
yuja, line 25: nit: `EverythingMatcher` is unused (but unchecked because of `allow(unused, dead_code)`.)

cli/src/commands/diff.rs
```diff
@@ -52,38 +59,62 @@ pub(crate) struct DiffArgs {
     format: DiffFormatArgs,
 }
 
+fn resolve_revision(
+    workspace_command: &WorkspaceCommandHelper,
+    r: &Option<RevisionArg>,
+) -> Result<Commit, CommandError> {
+    workspace_command.resolve_single_rev(r.as_ref().unwrap_or(&RevisionArg::AT))
+}
```
yuja, line 67: nit: maybe better to inline this as a closure? It's even shorter, and will probably help follow the code.

lib/src/merged_tree.rs
```diff
@@ -705,23 +708,39 @@ impl Iterator for TreeDiffIterator<'_> {
...
+                            value: Ok((before, after)),
+                        });
                     } else {
                         unreachable!();
                     }
                 }
             };
 
+            let path = entry.target;
+            let (before, after) = entry.value.unwrap();
```
yuja, line 711: Perhaps, we wouldn't have to change the type of `TreeDiffDirItem`. We don't need source/dest directory paths. Then, this `.unwrap()` can be removed.

cli/src/diff_util.rs
```diff
@@ -1032,10 +1044,15 @@ pub fn show_git_diff(
...
+        while let Some(MaterializedTreeDiffEntry {
+            source: _, // TODO handle copy tracking
+            target: path,
+            value: diff,
+        }) = diff_stream.next().await
+        {
             let path_string = path.as_internal_file_string();
             let (left_value, right_value) = diff?;
-            let left_part = git_diff_part(&path, left_value)?;
+            let left_part = git_digf_part(&path, left_value)?;
```
yuja: nit: typo `diff` -> `digf`

cli/src/commit_templater.rs
```diff
@@ -1394,8 +1394,15 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
             let path_converter = language.path_converter;
             let template = self_property
                 .map(move |diff| {
+                    let to_tree = diff.to_tree.clone();
```
yuja, line 1397: nit: can you add a TODO comment?

The diff object knows `to_tree`, so it can be somehow obtained from `self` without cloning. Since `show_diff_summary()` needs both `tree_diff` and `to_tree`, it might be better to pass `from_tree, to_tree` pair in place of `tree_diff`.

lib/src/backend.rs
```diff
@@ -458,7 +497,7 @@ pub trait Backend: Send + Sync + Debug {
     /// unnecessary resources.
     fn get_copy_records(
         &self,
-        paths: &[RepoPathBuf],
+        paths: Option<&[RepoPathBuf]>,
```
yuja, line 505: nit: can you also update the doc string?
> If `paths` is empty include all paths

Ideally, this change could be moved to the first commit, but I don't mind.

cli/src/commands/diff.rs
yuja: Agreed. I'm not strongly against merging this PR. It can be easily reverted, and people here (including me) wouldn't be an expert of copy tracking.

> And I would have preferred for https://github.com/martinvonz/jj/pull/4235 to have been part of the first commit, btw.

Suppose the new API isn't right, I would rather leave the current design doc as is.

yuja (approved):

cli/src/commands/fix.rs
PhilipMetzger, line 1: Nit: This is a `lib:` cleanup commit.

cli/src/diff_util.rs
PhilipMetzger, line 1: Ditto

lib/src/merged_tree.rs
PhilipMetzger, line 1: Ditto

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* The following diff formats now include information about copies and moves if
+  supported by the backend (the Git backend does): `--color-words`, `--summary`
```
PhilipMetzger, line 18: Nit: Drop the `if supported by the backend` part, as it will confuse users. As many don't understand the whole native/Git backend differences and _when_ they're used.

lib/src/backend.rs
```diff
@@ -217,6 +217,11 @@ impl CopyRecords {
     pub fn for_target(&self, target: &RepoPath) -> Option<&CopyRecord> {
         self.targets.get(target).and_then(|&i| self.records.get(i))
     }
+
+    /// Gets all copy records.
+    pub fn records(&self) -> impl Iterator<Item = &CopyRecord> + '_ {
+        self.records.iter()
+    }
```
PhilipMetzger, line 224: Nit: This should just be `records_iter` as reading `copy_records.records()` isn't really pleasant.

PhilipMetzger (approved):
LGTM

cli/src/commands/diff.rs
PhilipMetzger: Sorry for being late on this thread. I asked for the extraction of the design doc changes from this PR as I thought we'd run into such a problem. So it was the easiest way to unblock @fowles work and to land it in a reasonable state. 

> > And I would have preferred for #4235 to have been part of the first commit, btw.
> 
> Suppose the new API isn't right, I would rather leave the current design doc as is.

This is a great point.

PhilipMetzger: And it isn't entirely clear, what the actual API should be the implementation which we have now based on `Gix` or the proposed one from @torquestomp which needs a implementation downstream.

cli/src/commands/diff.rs
```diff
@@ -52,38 +59,62 @@ pub(crate) struct DiffArgs {
...
 #[instrument(skip_all)]
 pub(crate) fn cmd_diff(
     ui: &mut Ui,
     command: &CommandHelper,
     args: &DiffArgs,
 ) -> Result<(), CommandError> {
     let workspace_command = command.workspace_helper(ui)?;
+    let diff_renderer = workspace_command.diff_renderer_for(&args.format)?;
+    let fileset_expression = workspace_command.parse_file_patterns(&args.paths)?;
+    let matcher = fileset_expression.to_matcher();
```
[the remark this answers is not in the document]
fowles: Done

cli/src/diff_util.rs
fowles: Done

cli/src/diff_util.rs
fowles: Done

cli/src/commands/debug/copy_detection.rs
fowles: there was a lot of unused stuff being supressed there.  Good catch.

cli/src/commands/diff.rs
fowles: Done

lib/src/merged_tree.rs
fowles: done

cli/src/diff_util.rs
fowles: done

cli/src/commit_templater.rs
fowles: I was thinking the same thing.  I added a TODO, but I will send the follow up for this next

lib/src/backend.rs
fowles: done

CHANGELOG.md
fowles: Done

lib/src/backend.rs
fowles: went with just `iter()`

cli/src/commands/fix.rs
fowles: discussed briefly in discord, leaving because it is part of the intent behind the refactor

PR #4121, opened 2024-07-19
Add private-commits setting to prevent commits from being pushed

### LINKED ISSUE #3376: FR: A revset to determine private commits that may not be pushed
In chat, people have asked for a way to define commits that Jujutsu refuses to push (examples: [one](https://discord.com/channels/968932220549103686/969291218347524238/1220413138814505031), [two](https://discord.com/channels/968932220549103686/968932220549103689/1222208640317001757)). Currently, the only restriction is that commits with an empty description may not be pushed.

Define a new revset alias named `private_roots()` that describes revisions (and their descendents) that may not be pushed to remotes. By default, this will be defined as:

```toml
revset-aliases.'private_roots()' = 'description(exact:"")'
```

The name is open for bikeshedding, some options:

- `private_roots()`
- `private_changes()`

When pushing changes, Jujutsu should determine if any changes being pushed match this revset. If so, then the push is aborted.

Suppose my convention is to write "wip:" at the beginning of my private changes. Then, I would define this revset alias to avoid pushing them by mistake:

```toml
revset-aliases.'private_roots()' = 'description(exact:"") | description(glob:"wip:*")'
```

**Alternative solutions**
Another option would be to explicitly mark changes as private. For example, running `jj private set -r REV` would mark the specified changes as private, and they could not be pushed until unmarked with `jj private clear -r REV`

jennings:
Closes #3376.

# Concerns

If commits in `git.private-commits` are already immutable, they can be pushed. When `cmd_git_push` is checking if commits being pushed have descriptions, author information, or conflicts, we don't check immutable commits. I kept this behavior because it seemed reasonable and I couldn't find a non-tedious way to check privacy and report it in the "Won't push commit {} since {}" message.

# Bikeshedding possibilities!

- Should `jj git push --allow-private` be named differently?

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -93,6 +93,9 @@ pub struct GitPushArgs {
     /// Allow pushing commits with empty descriptions
     #[arg(long)]
     allow_empty_description: bool,
+    /// Allow pushing commits that are private
+    #[arg(long)]
+    allow_private: bool,
     /// Push branches pointing to these commits (can be repeated)
     #[arg(long, short)]
     revisions: Vec<RevisionArg>,
@@ -400,7 +403,7 @@ fn validate_commits_ready_to_push(
         if commit.has_conflict()? {
             reasons.push("it has conflicts");
         }
-        if is_private(commit.id()) {
+        if !args.allow_private && is_private(commit.id()) {
             reasons.push("it is private");
         }
         if !reasons.is_empty() {
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -18,6 +18,7 @@ use std::{fmt, io};
 
 use clap::ArgGroup;
 use itertools::Itertools;
+use jj_lib::backend::CommitId;
 use jj_lib::git::{self, GitBranchPushTargets, GitPushError};
 use jj_lib::object_id::ObjectId;
 use jj_lib::op_store::RefTarget;
@@ -261,56 +262,7 @@ pub fn cmd_git_push(
         );
     }
 
-    // Check if there are conflicts in any commits we're about to push that haven't
-    // already been pushed.
-    let new_heads = branch_updates
-        .iter()
-        .filter_map(|(_, update)| update.new_target.clone())
-        .collect_vec();
-    let old_heads = repo
-        .view()
-        .remote_branches(&remote)
-        .flat_map(|(_, old_head)| old_head.target.added_ids())
-        .cloned()
-        .collect_vec();
-    // (old_heads | immutable_heads() | root())..new_heads
-    let commits_to_push = RevsetExpression::commits(old_heads)
-        .union(&revset_util::parse_immutable_heads_expression(
-            &tx.base_workspace_helper().revset_parse_context(),
-        )?)
-        .range(&RevsetExpression::commits(new_heads));
-    for commit in tx
-        .base_workspace_helper()
-        .attach_revset_evaluator(commits_to_push)?
-        .evaluate_to_commits()?
-    {
-        let commit = commit?;
-        let mut reasons = vec![];
-        if commit.description().is_empty() && !args.allow_empty_description {
-            reasons.push("it has no description");
-        }
-        if commit.author().name.is_empty()
-            || commit.author().name == UserSettings::USER_NAME_PLACEHOLDER
-            || commit.author().email.is_empty()
-            || commit.author().email == UserSettings::USER_EMAIL_PLACEHOLDER
-            || commit.committer().name.is_empty()
-            || commit.committer().name == UserSettings::USER_NAME_PLACEHOLDER
-            || commit.committer().email.is_empty()
-            || commit.committer().email == UserSettings::USER_EMAIL_PLACEHOLDER
-        {
-            reasons.push("it has no author and/or committer set");
-        }
-        if commit.has_conflict()? {
-            reasons.push("it has conflicts");
-        }
-        if !reasons.is_empty() {
-            return Err(user_error(format!(
-                "Won't push commit {} since {}",
-                short_commit_hash(commit.id()),
-                reasons.join(" and ")
-            )));
-        }
-    }
+    validate_commits_ready_to_push(&branch_updates, &remote, &tx, command, args)?;
 
     writeln!(ui.status(), "Branch changes to push to {}:", &remote)?;
     for (branch_name, update) in &branch_updates {
@@ -384,6 +336,81 @@ pub fn cmd_git_push(
     })?;
     writer.flush(ui)?;
     tx.finish(ui, tx_description)?;
+    Ok(())
+}
+
+/// Validates that the commits that will be pushed are ready (have authorship
+/// information, are not conflicted, etc.)
+fn validate_commits_ready_to_push(
+    branch_updates: &[(String, BranchPushUpdate)],
+    remote: &str,
+    tx: &WorkspaceCommandTransaction,
+    command: &CommandHelper,
+    args: &GitPushArgs,
+) -> Result<(), CommandError> {
+    let workspace_helper = tx.base_workspace_helper();
+    let repo = workspace_helper.repo();
+
+    let new_heads = branch_updates
+        .iter()
+        .filter_map(|(_, update)| update.new_target.clone())
+        .collect_vec();
+    let old_heads = repo
+        .view()
+        .remote_branches(remote)
+        .flat_map(|(_, old_head)| old_head.target.added_ids())
+        .cloned()
+        .collect_vec();
+    let commits_to_push = RevsetExpression::commits(old_heads)
+        .union(&revset_util::parse_immutable_heads_expression(
+            &tx.base_workspace_helper().revset_parse_context(),
+        )?)
+        .range(&RevsetExpression::commits(new_heads));
+
+    let config = command.settings().config();
+    let is_private = if let Ok(revset) = config.get_string("git.private-commits") {
+        workspace_helper
+            .parse_revset(&RevisionArg::from(revset))?
+            .evaluate()?
+            .containing_fn()
+    } else {
+        Box::new(|_: &CommitId| false)
+    };
+
+    for commit in workspace_helper
+        .attach_revset_evaluator(commits_to_push)?
+        .evaluate_to_commits()?
+    {
+        let commit = commit?;
+        let mut reasons = vec![];
+        if commit.description().is_empty() && !args.allow_empty_description {
+            reasons.push("it has no description");
+        }
+        if commit.author().name.is_empty()
+            || commit.author().name == UserSettings::USER_NAME_PLACEHOLDER
+            || commit.author().email.is_empty()
+            || commit.author().email == UserSettings::USER_EMAIL_PLACEHOLDER
+            || commit.committer().name.is_empty()
+            || commit.committer().name == UserSettings::USER_NAME_PLACEHOLDER
+            || commit.committer().email.is_empty()
+            || commit.committer().email == UserSettings::USER_EMAIL_PLACEHOLDER
+        {
+            reasons.push("it has no author and/or committer set");
+        }
+        if commit.has_conflict()? {
+            reasons.push("it has conflicts");
+        }
+        if is_private(commit.id()) {
+            reasons.push("it is private");
+        }
+        if !reasons.is_empty() {
+            return Err(user_error(format!(
+                "Won't push commit {} since {}",
+                short_commit_hash(commit.id()),
+                reasons.join(" and ")
+            )));
+        }
+    }
     Ok(())
 }
 
--- a/cli/tests/test_git_private_commits.rs
+++ b/cli/tests/test_git_private_commits.rs
@@ -97,6 +97,34 @@ fn test_git_private_commits_block_pushing() {
 }
 
 #[test]
+fn test_git_private_commits_can_be_overridden() {
+    let (test_env, workspace_root) = set_up();
+
+    test_env.jj_cmd_ok(&workspace_root, &["new", "main", "-m=private 1"]);
+    test_env.jj_cmd_ok(&workspace_root, &["branch", "set", "main"]);
+
+    // Will not push when a pushed commit is contained in git.private-commits
+    test_env.add_config(r#"git.private-commits = "description(glob:'private*')""#);
+    let stderr = test_env.jj_cmd_failure(&workspace_root, &["git", "push", "--all"]);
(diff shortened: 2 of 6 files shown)


yuja:
Suppose this is primarily for preventing accidental push, we wouldn't need a global concept "private"? User can set `git.never-push = <revset>` or something, and they can update log_node template accordingly if they want.

We'll also need an escape hatch to override it (like `--ignore-immutable`.)

cli/src/commands/git/push.rs
```diff
@@ -279,6 +279,22 @@ pub fn cmd_git_push(
             &tx.base_workspace_helper().revset_parse_context(),
         )?)
         .range(&RevsetExpression::commits(new_heads));
+
+    // Only the roots themselves need to be checked if they're being pushed. If
+    // a root is already on the remote, then it is "grandfathered in" and
+    // shouldn't block any descendents from being pushed (unless the descendents
+    // are in private_roots()).
+    let private_commits_revset = revset_util::parse_private_roots_expression(
```
yuja, line 287: What if the commits to be pushed are ancestors of the private roots?

cli/src/commands/git/push.rs
```diff
@@ -279,6 +279,22 @@ pub fn cmd_git_push(
...
+        &tx.base_workspace_helper().revset_parse_context(),
+    )?;
+
+    let private_commits = tx
+        .base_workspace_helper()
+        .attach_revset_evaluator(private_commits_revset)?
+        .evaluate_to_commits()?
+        .filter_map(|ch| ch.map(|c| c.id().to_owned()).ok())
+        .collect::<HashSet<_>>();
```
yuja, line 296: Perhaps, you can optimize it by using `containing_fn()`.

cli/src/commands/git/push.rs
jennings: In that scenario, the commit in `private_roots()` is not part of commits_to_push and so the push is allowed.

My thought is: Whelp, that commit is already public, so clearly it shouldn’t be treated as though it’s private anymore.

jennings: Then I believe the push will be allowed as long as the private root itself isn’t being pushed.

jennings: I should make sure there’s an explicit tests for that too:

   o private
   |
   o public main*
   |
   o public main@origin

Pushing main should be allowed because the private commit isn’t being pushed.

cli/src/commands/git/push.rs
jennings: Oh neat, I didn't know about that. Will change.

jennings:
> Suppose this is primarily for preventing accidental push, we wouldn't need a global concept "private"? User can set `git.never-push = <revset>` or something, and they can update log_node template accordingly if they want.

I was only thinking of this as a way to prevent accidentally pushing commits. For example, I have a commit at work that switches the configuration between dev and prod. To use it, I `jj new @ description("private:production")` and work from there. But I never want to accidentally push that commit.

I'm not sure what else "private" might control besides pushing, is this something relevant to Google's backend?

cli/src/commands/git/push.rs
jennings: Added test `test_git_private_commits_descending_from_commits_pushed_do_not_block_pushing` for this.

yuja: Oops, I misunderstood the definition. Sorry about the confusion.

BTW, if a typical use case is something like `private_roots = description("WIP")` (i.e. orthogonal to DAG), it's probably better to not define the roots, but just let user define which revision has to be private.

If the supposed use case is `private_roots = some_user_specified_branch_as_a_root_marker` (strictly defined based on DAG), any descendants of the roots shouldn't be allowed to push (no matter if the roots are already pushed.)

jennings: I believe that's how it actually works. If a commit is in `private_roots()` but is already on the remote when you push, then its descendants aren't automatically blocked from pushing. Because you might have this:

```
o new stuff main*
|
o main@origin
|
= (400 elided commits)
|
o revert "WIP: building my feature"
|
o WIP: building my feature
|
~
```

That "WIP" commit is already public, so there's no sense blocking the push of _main_.

I thought "roots" was good in the name because of this:

```
o merge main*
|\
| o WIP: private changes
| |
o | public changes
|/
o main@origin
```

Even though the commit named "merge" isn't in `private_roots()`, it is implicitly blocked from pushing because doing so would require pushing the "WIP" commit as well.

The revset name can be changed though. If we prefer it to be something like `private_commits()`, that's no problem.

jennings: I added a little more explanation to `revsets.md` and added a "Set of private commits" section to `config.md`.

jennings:
> We'll also need an escape hatch to override it (like `--ignore-immutable`.)

`--ignore-private` sounds to me like it would _prevent_ pushing, not _allow_ it, so my preferred names in order are:

1. `jj git push --allow-private`
2. `jj git push --ignore-privacy`

cli/src/commands/git/push.rs
yuja: If it's named "roots", I would expect the rule is enforced to any `roots::` revisions, just like we do for `immutable_heads()` and `immutable() = ::immutable_heads()`.

Anyway, I don't think the global `private_roots()` concept is needed at this point. Some `jj git push`-specific config is probably good, and it can be extended to per-remote settings if needed. I'm not sure if `private_roots()` apply to other forges.

jennings: _Ohhhh_, I think I understand your concerns now. Let me restate it to see if I'm understanding now.

First, the name `private_roots` implies too strongly that it would always make its descendants private, even if the root has already been pushed. I have no problem renaming it.

Second, you don't want private commits to be a "first-class feature" of jj, you'd rather it be a notion specific to the git backend. A backend like Google's doesn't really need this privacy concept since everything is published anyway.

If I made the following two changes, I think it addresses both concerns?

1. Remove the special "private" label and glyph I added to the built-in templates. The end-user can set this up if they want it.
2. Rename the config setting from `revset-aliases.private_roots()` to `git.private_commits()`.

yuja: > First, the name `private_roots` implies too strongly that it would always make its descendants private, even if the root has already been pushed. I have no problem renaming it.
> 
> Second, you don't want private commits to be a "first-class feature" of jj, you'd rather it be a notion specific to the git backend. A backend like Google's doesn't really need this privacy concept since everything is published anyway.

Yes. Thanks for summarizing my thoughts.

I have no idea about Google backend, but I'm not sure if `gerrit send` should respect the "private" labeling for example. And we'll eventually need per-remote configuration because the user might want to exchange private commits between their local/personal clones.

>  1. Remove the special "private" label and glyph I added to the built-in templates. The end-user can set this up if they want it.
> 
>  2. Rename the config setting from `revset-aliases.private_roots()` to `git.private_commits()`.

Yes, that's the idea. The config name should be `git.private_commits` (no parentheses), btw. There might be a better name, but I don't have any suggestion.

This feature is similar to empty description check, so the default could be `private_commits = description(exact:"")`. I'm not sure if merging these is a good idea, and this can be addressed separately.

jennings: I made these two changes:

- The setting is now named `git.private-commits` (the other options there used hyphens, not underscores, so I matched them)
- I dropped the commit that added glpyhs to the commit templater

cli/src/commands/git/push.rs
```diff
@@ -279,36 +279,59 @@ pub fn cmd_git_push(
             &tx.base_workspace_helper().revset_parse_context(),
         )?)
         .range(&RevsetExpression::commits(new_heads));
-    for commit in tx
-        .base_workspace_helper()
-        .attach_revset_evaluator(commits_to_push)?
-        .evaluate_to_commits()?
+
```
yuja, line 285: nit: you can also move `new_heads`/`old_heads`/`commits_to_push` inside the block. It might be better to extract a check function because `cmd_git_push()` is super long.

cli/src/commands/git/push.rs
```diff
@@ -279,36 +279,59 @@ pub fn cmd_git_push(
...
+                &tx.base_workspace_helper().revset_parse_context(),
+            )?;
+            Some(
+                workspace_helper
+                    .attach_revset_evaluator(private_commits_revset)?
+                    .evaluate()?
+                    .containing_fn(),
+            )
+        } else {
+            None
```
yuja, line 302: nit: could be `Box::new(|_: &CommitId| false)` instead of `Option`

cli/tests/test_private_commits.rs
yuja, line 1: nit: rename to `test_git_private_commits.rs`?

cli/tests/test_private_commits.rs
```diff
@@ -0,0 +1,227 @@
+// Copyright 2023 The Jujutsu Authors
```
yuja, line 1: nit: `s/2023/2024/`

lib/src/settings.rs
```diff
@@ -42,6 +42,7 @@ pub struct RepoSettings {
 pub struct GitSettings {
     pub auto_local_branch: bool,
     pub abandon_unreachable_commits: bool,
+    pub private_commits: Option<String>,
```
yuja, line 45: nit: it's unclear whether this config should be in library `GitSettings`, but I think it is closer to `git.push`/`git.fetch`, so belongs to the CLI.

yuja (commented):
Generally looks good to me, thanks!

@ilyagr, any comments about this feature?

cli/src/commands/git/push.rs
jennings: Ha, I actually wrote it that way at first but then changed my mind. I'll change it back.

lib/src/settings.rs
jennings: Oh, I didn't notice this was in the lib. I just saw where the other two `git.` settings were and put it alongside. I'll try to move this out to the cli.

cli/src/commands/git/push.rs
jennings: Created `validate_commits_ready_to_push`

cli/tests/test_git_private_commits.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+#[test]
+fn test_git_private_commits_are_not_checked_if_immutable() {
+    let (test_env, workspace_root) = set_up();
+
+    test_env.jj_cmd_ok(&workspace_root, &["new", "main", "-m=private 1"]);
+    test_env.jj_cmd_ok(&workspace_root, &["branch", "set", "main"]);
+
+    // Will not push when a pushed commit is contained in git.private-commits
```
yuja, line 134: nit: this comment looks wrong since the private commits in immutable set isn't checked.

yuja (approved):
Thanks.

jennings:
Thank you for your patience and guidance.

docs/config.md
```diff
@@ -809,6 +809,24 @@ example:
 
     git.push-branch-prefix = "martinvonz/push-"
 
+### Set of private commits
+
+You can configure the set of private commits by setting `git.private-commits` to
+a revset. The value is a revset of commits that Jujutsu will refuse to push. If
+unset, all commits are eligible to be pushed.
```
ilyagr, line 816: Currently, commits with empty descriptions cannot be pushed even if they are not in `private-commits`, right?

I'm wondering whether or not this is good. The good part of it is that we can have a more informative message.

In any case, should we document it here?

-------------

There is also a similar problem with "immutable_heads()": it is currently very inflexible to configure. There are revisions that are always immutable (`root()`), the `untracked_remotes()` that are a really good idea to keep there, there's `trunk()` which is there by default, and there are any additional heads the user wants to add in there. Currently, to do the latter, the user has to understand all of the former and include them in the config.

I wish we could solve this problem in such a way that there would be a natural way to include commits with no description or author into `private-commits`, without the user having to think about that whenever they want to add something more to their `private-commits`.

ilyagr (commented):
Sorry, I was a bit less attached to my computer than usual for a few days. This may or may not continue, I'm unsure.

Thank you for the feature!

As Yuya guessed in one of the discussions, I do wonder whether there should also be a per-remote setting.

docs/config.md
jennings: I see how what I wrote could be confusing. Where I wrote, "If unset, all commits are eligible to be pushed," I should have written, "If unset, no commits are private." The existing checks for authorship, empty description, conflicts, etc. should be unchanged.

Regarding `immutable_heads()`, I do this in my user config:

```toml
"immutable_heads()" = "trunk() | tags() | ... | additional_immutable_heads()"
"additional_immutable_heads()" = "none()"
```

Then in any repo, I can append immutable heads by setting `additional_immutable_heads()`. Maybe jj could do something like this would be baked into the binary? I'm not sure which would be better:

```toml
# jj only uses real_immutable_heads().
# Set immutable_heads() to append to the default revset.
# Set real_immutable_heads() to replace the default revset.
"immutable_heads()"      = "none()"
"real_immutable_heads()" = "trunk() | ... | immutable_heads()"

# jj only uses immutable_heads().
# Set additional_immutable_heads() to append to the default revset.
# Set immutable_heads() to replace the default revset.
"immutable_heads()"            = "trunk() | ... | additional_immutable_heads()"
"additional_immutable_heads()" = "none()"
```

If this would be a good idea for immutable heads, maybe private commits could work similarly? Something to the effect of:

```toml
private-commits = "none()"
real-private-commits = 'description(exact:"") | author(exact:"") | conflicts() | ...`
```

yuja: I wouldn't add `immutable_heads`/`additional_immutable_heads` abstraction to the default configuration. It can't subtract conditions from the base set, and if user override the base set to achieve that, the additional expression won't be used anymore.

yuja: Hmm, `builtin_*()` sounds good to me.

ilyagr: I also think `builtin_*` is a good idea. It's not perfect -- it's a bit difficult to discover for the user -- but I can't think of anything much better, and perhaps we can make it more discoverable later.

In principle, if needed (which probably it is not yet), we could subdivide the builtin into a union of several other revsets that the user could pick and choose from.

> I see how what I wrote could be confusing. Where I wrote, "If unset, all commits are eligible to be pushed," I should have written, "If unset, no commits are private." The existing checks for authorship, empty description, conflicts, etc. should be unchanged.

Yeah, I think we should update this a bit. Perhaps also mention that the commits without an author can never be pushed.

Also, no problem, I don't think this is problematic, just something that can be improved a bit.

PR #4127, opened 2024-07-20
RFC: commit_builder: add public interface that writes temporary commit to store

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/commit_builder.rs
+++ b/lib/src/commit_builder.rs
@@ -192,19 +192,8 @@ impl CommitBuilder<'_> {
         self
     }
 
-    pub fn write(mut self) -> BackendResult<Commit> {
-        let sign_settings = &self.sign_settings;
-        let store = &self.store;
-
-        let should_sign = store.signer().can_sign() && sign_settings.should_sign(&self.commit);
-        let sign_fn = |data: &[u8]| store.signer().sign(data, sign_settings.key.as_deref());
-
-        // Commit backend doesn't use secure_sig for writing and enforces it with an
-        // assert, but sign_settings.should_sign check above will want to know
-        // if we're rewriting a signed commit
-        self.commit.secure_sig = None;
-
-        let commit = store.write_commit(self.commit, should_sign.then_some(&mut &sign_fn))?;
+    pub fn write(self) -> BackendResult<Commit> {
+        let commit = write_to_store(&self.store, self.commit, &self.sign_settings)?;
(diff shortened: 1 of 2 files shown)


lib/src/commit_builder.rs
```diff
@@ -309,6 +325,14 @@ impl DetachedCommitBuilder {
...
         Ok(commit)
     }
+
+    /// Writes new commit without making it visible in the repo.
+    ///
+    /// This does not consume the builder, so you can reuse the current
+    /// configuration to create another commit later.
+    pub fn write_hidden(&mut self) -> BackendResult<Commit> {
+        write_to_store(&self.store, self.commit.clone(), &self.sign_settings)
+    }
```
yuja, line 320: > In order to render description template, we'll need a Commit object that
> represents the old state (with new tree and parents) before updating the
> commit description. The added functions will help generate an intermediate
> Commit object.
> 
> Alternatively, we can create an in-memory Commit object with some fake
> CommitId. It should be lightweight, but might cause weird issue because the
> fake id wouldn't be found in the store.
> 
> I think it's okay to write a temporary commit and rely on GC as we do for
> merge trees. However, I should note that temporary commits are more likely to
> be preserved as they are pinned by no-gc refs until "jj util gc".

@martinvonz do you think this is good idea?

The resulting `cmd_commit()` will look like:
```rust
    // set up temporary commit object
    let mut commit_builder = tx
        .mut_repo()
        .rewrite_commit(command.settings(), &commit)
        .detach();
    commit_builder.set_tree_id(tree_id);
    ...

    // render description ++ diff.summary() with the temporary commit
    // other commit keywords can be used, but branches won't work as expected
    let temp_commit = commit_builder.write_hidden()?;
    let template = description_template(&tx, None, &temp_commit, command.settings())?;
    let description = edit_description(tx.base_repo(), &template, command.settings())?

    // update description and write final commit object to store
    commit_builder.set_description(description);
    let new_commit = commit_builder.write(tx.mut_repo())?;
```

martinvonz: Makes sense. I'm not worried about the extra GC-able commits.

We don't have GC at Google, but we've naturally already decide to accept the amount of trees and commits we get from regular snapshotting, so creating an extra commit object in `jj commit` should be fine. And maybe we'll implement GC some day anyway.

martinvonz (approved):

PR #4130, opened 2024-07-20
Warn user about the working copy when configuring the author

### LINKED ISSUE #4042: Surprising author of the first working copy
## Description

I am writing a codelab-like tutorial for `jj` on the basis of Steve's tutorial. For that I tried to make up a `codelab@example.com` author, to avoid using personal data. However, it seems that the first working copy commit is created with global metadata, and that changing the author before doing anything does not fix that.

```
mkdir -p /tmp/test
cd /tmp/test
jj git init
jj config set --repo user.name 'Code Lab' && jj config set --repo user.email 'codelab@example.com'
jj log
```

## Expected Behavior

```
$ (/tmp/test) jj log
@  nlylsuyy codelab@example.com 2024-07-07 00:35:56 41444201
◉  zzzzzzzz root() 00000000
```

## Actual Behavior

```
$ (/tmp/test) jj log
@  nlylsuyy msmetko@msmetko.xyz 2024-07-07 00:35:56 41444201
◉  zzzzzzzz root() 00000000
```

(The username is an email address which is not supposed to be used for this repo. Subsequent commits, however, do use `codelab@example.com` mail address.)
## Specifications

- Platform: arch rolling
- Version: 0.18.0 &ndash; tested with 0.19.0, too

InCogNiTo124:
Closes #4042 

# Checklist

If applicable:

# TODO

- [x] ~~Add `jj git init` flags for user and email~~ Deprecated
- [x] ~~make `jj describe --no-edit` accept a revset~~ Deprecated, fixed by #3828

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * A tilde (`~`) at the start of the path will now be expanded to the user's home
   directory when configuring a `signing.key` for SSH commit signing.
+  
+* When reconfiguring the author, warn that the working copy won't be updated
 
 ### Fixed bugs
 
--- a/cli/src/commands/config/set.rs
+++ b/cli/src/commands/config/set.rs
@@ -12,10 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::io;
+
+use jj_lib::commit::Commit;
+use jj_lib::repo::Repo;
 use tracing::instrument;
 
 use super::ConfigLevelArgs;
-use crate::cli_util::{get_new_config_file_path, CommandHelper};
+use crate::cli_util::{get_new_config_file_path, CommandHelper, WorkspaceCommandHelper};
 use crate::command_error::{user_error, CommandError};
 use crate::config::{
     parse_toml_value_or_bare_string, write_config_value_to_file, ConfigNamePathBuf,
@@ -33,9 +37,15 @@ pub struct ConfigSetArgs {
     level: ConfigLevelArgs,
 }
 
+/// Denotes a type of author change
+enum AuthorChange {
+    Name,
+    Email,
+}
+
 #[instrument(skip_all)]
 pub fn cmd_config_set(
-    _ui: &mut Ui,
+    ui: &mut Ui,
     command: &CommandHelper,
     args: &ConfigSetArgs,
 ) -> Result<(), CommandError> {
@@ -46,7 +56,59 @@ pub fn cmd_config_set(
             path = config_path.display()
         )));
     }
+
     // TODO(#531): Infer types based on schema (w/ --type arg to override).
     let value = parse_toml_value_or_bare_string(&args.value);
+
+    // If the user is trying to change the author config, we should warn them that
+    // it won't affect the working copy author
+    if args.name == ConfigNamePathBuf::from_iter(vec!["user", "name"]) {
+        check_wc_author(ui, command, &value, AuthorChange::Name)?;
+    } else if args.name == ConfigNamePathBuf::from_iter(vec!["user", "email"]) {
+        check_wc_author(ui, command, &value, AuthorChange::Email)?;
+    };
+
     write_config_value_to_file(&args.name, value, &config_path)
+}
+
+/// Returns the commit of the working copy if it exists.
+fn maybe_wc_commit(helper: &WorkspaceCommandHelper) -> Option<Commit> {
+    let repo = helper.repo();
+    let id = helper.get_wc_commit_id()?;
+    repo.store().get_commit(id).ok()
+}
+
+/// Check if the working copy author name matches the user's config value
+/// If it doesn't, print a warning message
+fn check_wc_author(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    new_value: &toml_edit::Value,
+    author_change: AuthorChange,
+) -> io::Result<()> {
+    let helper = match command.workspace_helper(ui) {
+        Ok(helper) => helper,
+        Err(_) => return Ok(()), // config set should work even if cwd isn't a jj repo
+    };
+    if let Some(wc_commit) = maybe_wc_commit(&helper) {
+        let author = wc_commit.author();
+        let orig_value = match author_change {
+            AuthorChange::Name => &author.name,
+            AuthorChange::Email => &author.email,
+        };
+        if new_value.as_str() != Some(orig_value) {
+            warn_wc_author(ui, &author.name, &author.email)?
+        }
+    }
+    Ok(())
+}
+
+/// Prints a warning message about the working copy to the user
+fn warn_wc_author(ui: &Ui, user_name: &str, user_email: &str) -> io::Result<()> {
+    Ok(writeln!(
+        ui.warning_default(),
+        "This setting will only impact future commits.\nThe author of the working copy will stay \
+         \"{user_name} <{user_email}>\".\nTo change the working copy author, use \"jj describe \
+         --reset-author --no-edit\""
+    )?)
 }
--- a/cli/tests/test_config_command.rs
+++ b/cli/tests/test_config_command.rs
@@ -857,6 +857,43 @@ fn test_config_show_paths() {
     "###);
 }
 
+#[test]
+fn test_config_author_change_warning() {
+    let test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+    let (stdout, stderr) = test_env.jj_cmd_ok(
+        &repo_path,
+        &["config", "set", "--repo", "user.email", "'Foo'"],
+    );
+    insta::assert_snapshot!(stdout, @"");
+    insta::assert_snapshot!(stderr, @r###"
+    Warning: This setting will only impact future commits.
+    The author of the working copy will stay "Test User <test.user@example.com>".
+    To change the working copy author, use "jj describe --reset-author --no-edit"
+    "###);
+
+    // test_env.jj_cmd resets state for every invocation
+    // for this test, the state (user.email) is needed
+    let mut log_cmd = test_env.jj_cmd(&repo_path, &["describe", "--reset-author", "--no-edit"]);
+    log_cmd.env_remove("JJ_EMAIL");
+    log_cmd.assert().success();
+
+    let (stdout, _) = test_env.jj_cmd_ok(&repo_path, &["log"]);
+    assert!(stdout.contains("Foo"));
+}
+
+#[test]
+fn test_config_author_change_warning_root_env() {
+    let mut test_env = TestEnvironment::default();
+    let user_config_path = test_env.config_path().join("config.toml");
+    test_env.set_config_path(user_config_path);
+    test_env.jj_cmd_success(
+        test_env.env_root(),
+        &["config", "set", "--user", "user.email", "'Foo'"],
+    );
+}
+
 fn find_stdout_lines(keyname_pattern: &str, stdout: &str) -> String {
     let key_line_re = Regex::new(&format!(r"(?m)^{keyname_pattern} = .*$")).unwrap();
     key_line_re


cli/src/commands/config.rs
```diff
@@ -296,9 +297,56 @@ pub(crate) fn cmd_config_set(
...
+    let change_hash = match maybe_wc_commit {
+        Some(commit) => format!("{} ", &short_change_hash(commit.change_id())[..8]), /* extra space for formatting */
+        None => String::from(""),
+    };
+    Ok(writeln!(
+        ui.warning_default(),
+        "This setting will only impact future commits.\nThe author of the working copy \
+         {change_hash}will stay \"{user_name} <{user_email}>\". \nTo change it, use \"jj describe \
+         --reset-author --no-edit\""
+    )?)
```
yuja, line 347: nit: I don't think printing hash is needed (for warning/hint), but maybe you can use `write_commit_summary()` for this purpose.

cli/src/commands/config.rs
```diff
@@ -296,9 +297,56 @@ pub(crate) fn cmd_config_set(
...
     }
+
+    // If the user is trying to change the author config, we should warn them that
+    // it won't affect the working copy author
+    if args.name == ConfigNamePathBuf::from_iter(vec!["user", "name"]) {
+        let config = command.settings().config();
+        let user_name: String = config.get("user.name")?;
+
+        // only warn if the author is actually changing
+        if user_name != args.value {
```
yuja, line 308: `args.value` may be escaped in TOML syntax, whereas `user_name` is a parsed value.

Maybe we can instead reload the config file and compare name/email fields? That might be silly, but we can reuse the same function for `config edit`.

InCogNiTo124: > Maybe we can instead reload the config file and compare name/email fields?

That was my initial idea, but it it also felt silly. I would basically have to have 2 entire copies of the config, one before and the other after.

> `args.value` may be escaped in TOML syntax, whereas `user_name` is a parsed value.

WDYM by this? I thought both are just string values. I mean, I tested it locally and it works as intended, what am I missing here?

cli/src/commands/config.rs
InCogNiTo124: > maybe you can use write_commit_summary() for this purpose.

It prints a lot of different info directly to the formatter so it's of no use for me. I'll remove the reference to the change id.

cli/src/commands/config.rs
InCogNiTo124: Actually, scratch that, it does not work as intended

InCogNiTo124:
>  Add `jj git init` flags for user and email

It's not clear to me how I should approach this. I noticed there's a `config::ConfigBuilder` but I don't think I can easily override `UserSettings` or `RepoSettings`. All suggestions welcome (including takeover from someone with more know-how :smiley: )

yuja:
> > Add `jj git init` flags for user and email
> 
> It's not clear to me how I should approach this.

I personally don't think this is good idea, but if needed, it's probably easier to implement as global flags. (See how `--color` is translated to config value.)

cli/src/commands/config.rs
InCogNiTo124: The bug is now fixed, but I'm still using `args.value`

InCogNiTo124:
> if needed, it's probably easier to implement as global flags

I would argue that that's not a clean solution, either, since those flags are only needed for `jj git init`

yuja:
> > if needed, it's probably easier to implement as global flags
> 
> I would argue that that's not a clean solution, either, since those flags are only needed for `jj git init`

I'm not sure why it's needed for (and should be restricted to) `jj git init`. User name/email applies globally, so it might make sense to do `jj describe --user foo`. (To be clear, I don't think the global flag is needed because it can be achieved by `--config-toml ..`.)

cli/src/commands/config.rs
yuja: Try something like `jj config set user.name '"Foo"'`. That's the original issue I found.

InCogNiTo124:
Tested with following snippets that it works:

<details>
<summary>Testing snippet</summary>

```console
❯ JJ_USER=user1 JJ_EMAIL=email1 jj git init .
Initialized repo in "."
❯ jj log
@  smytynvq email1 2024-08-06 21:11:21 303fc106
◆  zzzzzzzz root() 00000000
❯ jj config set --repo user.email "email2"
Warning: This setting will only impact future commits.
The author of the working copy will stay "user1 <email1>". 
To change it, use "jj describe --reset-author --no-edit"
❯ jj log
@  smytynvq email1 2024-08-06 21:11:21 303fc106
◆  zzzzzzzz root() 00000000
❯ jj describe --reset-author --no-edit
Working copy now at: smytynvq 1f2c39c8 (empty) (no description set)
❯ jj log
@  smytynvq email2 2024-08-06 21:12:08 1f2c39c8
◆  zzzzzzzz root() 00000000
❯ echo 'a' > a; jj commit -m 'a'; jj log
Working copy now at: xywsopol 5973def1 (empty) (no description set)
Parent commit      : smytynvq 92a5caae a
@  xywsopol email2 2024-08-06 21:12:45 5973def1
○  smytynvq email2 2024-08-06 21:12:45 92a5caae
│  a
◆  zzzzzzzz root() 00000000
❯ jj config set --repo user.email "msmetko@msmetko.xyz"
Warning: This setting will only impact future commits.
The author of the working copy will stay "InCogNiTo124 <email2>". 
To change it, use "jj describe --reset-author --no-edit"
❯ jj describe --no-edit --reset-author 'author("email2")'; jj log
Updated 2 commits
Working copy now at: xywsopol cbae64d7 (empty) (no description set)
Parent commit      : smytynvq c8442055 a
@  xywsopol msmetko@msmetko.xyz 2024-08-06 21:13:53 cbae64d7
○  smytynvq msmetko@msmetko.xyz 2024-08-06 21:13:53 c8442055
│  a
◆  zzzzzzzz root() 00000000
```

</details>

cli/src/commands/config.rs
InCogNiTo124: `jj config set user.name '"Foo"'` works for me by setting the user to `Foo`. Isn't that expected?

yuja: > `jj config set user.name '"Foo"'` works for me by setting the user to `Foo`. Isn't that expected?

`args.value = '"Foo"'` is parsed as `Foo`, not `"Foo"` as you said, which means `user_name != args.value` doesn't work in that case. `user_name` is parsed `Foo` whereas `args.value` is raw `"Foo"`.

Perhaps, `write_config_value_to_file()` can be changed to take `value: toml_edit::Value` instead of `value_str: &str`.

InCogNiTo124: >  `user_name != args.value` doesn't work in that case

Ah now I understand. I thought not having quotes / parsing the string _was_ the intended behaviour.

> Perhaps, `write_config_value_to_file()` can be changed to take `value: toml_edit::Value` instead of `value_str: &str`.

I'll try to investigate in the upcoming days, but I'm not super confident. A naive change to `::Value` resulted in tests failing because everything ends up being a TOML string, because there's no toml parsing. But maybe I come up with a better way.

yuja: https://github.com/martinvonz/jj/pull/4244

Parsing is intended. We just need to use parsed value for comparison.

InCogNiTo124: Hmm I rebased past your PR and it's still the same:

```
❯ PATH="~/InCogNiTo124-contrib/jj/target/debug:${PATH}"; which jj
~/InCogNiTo124-contrib/jj/target/debug/jj
❯ jj git init
Initialized repo in "."
❯ jj config set --repo user.email "'Bar'"
Warning: This setting will only impact future commits.
The author of the working copy will stay "InCogNiTo124 <msmetko@msmetko.xyz>". 
To change the working copy author, use "jj describe --reset-author --no-edit"
❯ jj describe --reset-author --no-edit
Working copy now at: pontmtzu 64dc19f3 (empty) (no description set)
❯ jj log
@  pontmtzu Bar 2024-08-09 20:53:21 64dc19f3
◆  zzzzzzzz root() 00000000
```

yuja: Use parsed `value` instead of `args.value`.

InCogNiTo124: > Use parsed value instead of args.value.

still does not work as intended:
```console
❯ jj config set --repo user.email '"Foo"'
Warning: This setting will only impact future commits.
The author of the working copy will stay "InCogNiTo124 <msmetko@msmetko.xyz>". 
To change the working copy author, use "jj describe --reset-author --no-edit"
❯ jj describe --reset-author --no-edit
Working copy now at: vvtrswtz 84b3b3dc (empty) (no description set)
❯ jj log
@  vvtrswtz Foo 2024-08-10 13:50:54 84b3b3dc
◆  zzzzzzzz root() 00000000
❯ cat .jj/repo/config.toml
[user]
email = "Foo"
```

yuja: What do you mean by "as intended"? The test case would be something like
1. create working-copy commit as email: `Foo`
2. `jj config set --repo user.email '"Foo"'`
And 2 shouldn't suggest `--reset-author` because the author email is unchanged.

Oh, and you'll have to use `.as_str()` instead of `.to_string()`. The latter will encode `toml::Value` as a TOML expression.
https://docs.rs/toml_edit/latest/toml_edit/enum.Value.html#method.as_str

`.as_str()` is stricter than `config.get_string()`, but that's probably okay.

InCogNiTo124: The latest push should work:

<details>
<summary>Test case</summary>

```console
❯ # SETUP
❯ jj git init
Initialized repo in "."
❯ jj log
@  nwlvpyok msmetko@msmetko.xyz 2024-08-12 19:17:16 1d8a1de3
◆  zzzzzzzz root() 00000000

❯ # EXPECTED NO WARNING (same author)
❯ jj config set --repo user.email "msmetko@msmetko.xyz"

❯ # EXPECTED WARNING (new author)
❯ jj config set --repo user.email "Foo"
Warning: This setting will only impact future commits.
The author of the working copy will stay "InCogNiTo124 <msmetko@msmetko.xyz>". 
To change the working copy author, use "jj describe --reset-author --no-edit"

❯ # EXPECTED AUTHOR RESET
❯ jj describe --reset-author --no-edit
Working copy now at: nwlvpyok 8c648ab7 (empty) (no description set)
❯ jj log
@  nwlvpyok Foo 2024-08-12 19:18:14 8c648ab7
◆  zzzzzzzz root() 00000000

❯ # EXPECTED NO WARNING (same author)
❯ jj config set --repo user.email "'Foo'"
❯ jj config set --repo user.email '"Foo"'
```

</details>

cli/src/commands/config/set.rs
```diff
@@ -12,16 +12,17 @@
...
 
 use super::ConfigLevelArgs;
-use crate::cli_util::{get_new_config_file_path, CommandHelper};
+use crate::cli_util::{get_new_config_file_path, CommandHelper, WorkspaceCommandHelper};
 use crate::command_error::{user_error, CommandError};
 use crate::config::{
     parse_toml_value_or_bare_string, write_config_value_to_file, ConfigNamePathBuf,
 };
 use crate::ui::Ui;
-
```
yuja, line 24: nit: unrelated change. restore blank line.

cli/src/commands/config/set.rs
```diff
@@ -34,8 +35,8 @@ pub struct ConfigSetArgs {
 }
 
 #[instrument(skip_all)]
-pub fn cmd_config_set(
-    _ui: &mut Ui,
+pub(crate) fn cmd_config_set(
```
yuja, line 38: nit: unrelated. remove `(crate)`.

cli/src/commands/config/set.rs
```diff
@@ -46,7 +47,85 @@ pub fn cmd_config_set(
             path = config_path.display()
         )));
     }
-    // TODO(#531): Infer types based on schema (w/ --type arg to override).
```
yuja, line 49: nit: unrelated change. restore TODO comment

cli/src/commands/config/set.rs
```diff
@@ -46,7 +47,85 @@ pub fn cmd_config_set(
...
+}
+
+/// Check if the working copy author name matches the user's config value
+/// If it doesn't, print a warning message
+fn check_wc_user_name(
+    command: &CommandHelper,
+    ui: &mut Ui,
+    user_name: &toml_edit::Value,
+) -> Result<(), CommandError> {
+    let helper = command.workspace_helper(ui)?;
```
yuja, line 80: you'll need to suppress error by `let Ok(helper) = .. else { return .. };` or something.

`config set` should work even if cwd isn't a jj repo.

cli/src/commands/config/set.rs
```diff
@@ -46,7 +47,85 @@ pub fn cmd_config_set(
...
+    if let Some(wc_commit) = maybe_wc_commit(&helper) {
+        let author = wc_commit.author();
+        match user_name.as_str() {
+            None => {
+                return Ok(());
+            }
+            Some(user_name) => {
+                if author.name != user_name {
```
yuja, line 88: nit: perhaps, `if user_name.as_str() != Some(&author.name)` is good.

cli/src/commands/config/set.rs
```diff
@@ -46,7 +47,85 @@ pub fn cmd_config_set(
...
         )));
     }
-    // TODO(#531): Infer types based on schema (w/ --type arg to override).
     let value = parse_toml_value_or_bare_string(&args.value);
+    // If the user is trying to change the author config, we should warn them that
+    // it won't affect the working copy author
+    if args.name == ConfigNamePathBuf::from_iter(vec!["user", "name"]) {
+        check_wc_user_name(command, ui, &value)?;
+    } else if args.name == ConfigNamePathBuf::from_iter(vec!["user", "email"]) {
+        check_wc_user_email(command, ui, &value)?;
```
yuja, line 56: nit: perhaps, `check_wc_user_name`/`email` can be combined as a function that returns `Option<Signature>` or `bool`?

Something like
```rust
    if let Some(author) = get_wc_author(command) {
        if value.as_str() != Some(&author.name)) {
            warn_wc_author(ui, &author)?;
        }
    }
```

yuja (commented):
Thanks!

cli/src/commands/config/set.rs
InCogNiTo124: `cargo +nightly fmt --all` removes the line every time I add it.

cli/src/commands/config/set.rs
InCogNiTo124: It is! Thanks.

cli/src/commands/config/set.rs
InCogNiTo124: Done.

cli/src/commands/config/set.rs
InCogNiTo124: Done

cli/src/commands/config/set.rs
InCogNiTo124: I combined the functions in the best way I could.

cli/src/commands/config/set.rs
```diff
@@ -46,7 +54,63 @@ pub fn cmd_config_set(
...
+    maybe_wc_commit
+}
+
+/// Check if the working copy author name matches the user's config value
+/// If it doesn't, print a warning message
+fn check_wc_author(
+    command: &CommandHelper,
+    ui: &mut Ui,
+    new_value: &toml_edit::Value,
+    author_change: AuthorChange,
```
yuja, line 89: nit: let's move `ui` argument first for API consistency. `(ui, command, new_value, author_change)`

cli/src/commands/config/set.rs
```diff
@@ -46,7 +54,63 @@ pub fn cmd_config_set(
...
+        };
+        if new_value.as_str() != Some(orig_value) {
+            warn_wc_author(&author.name, &author.email, ui)?
+        }
+    }
+    Ok(())
+}
+
+/// Prints a warning message about the working copy to the user
+fn warn_wc_author(user_name: &str, user_email: &str, ui: &mut Ui) -> Result<(), CommandError> {
```
yuja, line 109: nit: ditto. `(ui, user_name, user_mail)` or `(ui, author)`

cli/src/commands/config/set.rs
```diff
@@ -46,7 +54,63 @@ pub fn cmd_config_set(
...
+}
+
+/// Check if the working copy author name matches the user's config value
+/// If it doesn't, print a warning message
+fn check_wc_author(
+    command: &CommandHelper,
+    ui: &mut Ui,
+    new_value: &toml_edit::Value,
+    author_change: AuthorChange,
+) -> Result<(), CommandError> {
```
yuja, line 90: nit: let's restrict the error type to `io::Result<()>` to ensure that `workspace_helper()` error is suppressed.

cli/src/commands/config/set.rs
```diff
@@ -46,7 +54,63 @@ pub fn cmd_config_set(
...
+    let helper = match command.workspace_helper(ui) {
+        Ok(helper) => helper,
+        Err(_) => return Ok(()), // config set should work even if cwd isn't a jj repo
+    };
+    if let Some(wc_commit) = maybe_wc_commit(&helper) {
+        let author = wc_commit.author();
+        let orig_value = match author_change {
+            AuthorChange::Name => &author.name,
+            AuthorChange::Email => &author.email,
+        };
```
yuja, line 100: nit: at your option, `author_change` can be a callback function

```rust
get_old_value: impl FnOnce(&Signature) -> &str,
```

I think the use of enum is also good.

cli/tests/test_config_command.rs
```diff
@@ -857,6 +857,32 @@ fn test_config_show_paths() {
...
+    // test_env.jj_cmd resets state for every invocation
+    // for this test, the state (user.email) is needed
+    let mut log_cmd = test_env.jj_cmd(&repo_path, &["describe", "--reset-author", "--no-edit"]);
+    log_cmd.env_remove("JJ_EMAIL");
+    log_cmd.assert().success();
+
+    let (stdout, _) = test_env.jj_cmd_ok(&repo_path, &["log"]);
+    assert!(stdout.contains("Foo"));
+}
+
```
yuja, line 885: Can you also add a test that runs `config set` at `&test_env.env_root()`?

yuja (approved):

CHANGELOG.md
```diff
@@ -124,6 +124,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj describe` can now update the description of multiple commits.
 
+* When reconfiguring the author, warn that the working copy won't be updated
```
bnjmnt4n, line 127: Nit: This should be shifted up to the unreleased section since the last release.

cli/tests/test_config_command.rs
InCogNiTo124: Hmm, adding the following test fails:

```rust
#[test]
fn test_config_author_change_warning_root_env() {
    test_env.jj_cmd_ok(
        test_env.env_root(),
        &["config", "set", "--user", "user.email", "'Foo'"],
    );
}
```

but everything works fine if I run it. Maybe I need to modify the environment somehow?

CHANGELOG.md
InCogNiTo124: Great catch, thanks!

cli/tests/test_config_command.rs
InCogNiTo124: Solved it by adding `config.toml` as some other examples in the file.

cli/src/commands/config/set.rs
```diff
@@ -46,7 +56,63 @@ pub fn cmd_config_set(
...
 }
+
+/// Returns the commit of the working copy if it exists.
+fn maybe_wc_commit(helper: &WorkspaceCommandHelper) -> Option<Commit> {
+    let repo = helper.repo();
+    let maybe_wc_commit = helper
+        .get_wc_commit_id()
+        .map(|id| repo.store().get_commit(id))
+        .transpose()
+        .unwrap();
```
yuja, line 81: Oh, there's a real error which I didn't notice, sorry. Suppose it's okay to suppress warning on storage error, it can be turned into `None`.
```rust
    let id = helper.get_wc_commit_id()?;
    repo.store().get_commit(id).ok()
```

yuja (approved):

InCogNiTo124:
@yuja I believe this can be merged

yuja:
> @yuja I believe this can be merged

Yes. Maybe you don't have a permission to merge the PR yet?

cc @martinvonz

martinvonz:
> > @yuja I believe this can be merged
> 
> Yes. Maybe you don't have a permission to merge the PR yet?

Yes, that's correct. Sorry that I didn't realize :(

PR #4135, opened 2024-07-22
Add hunk.nvim to list of community tools

julienvincent:
I just finished building a Neovim plugin which can be used as an alternative `diff-editor` for Jujutsu.

You can find it here: https://github.com/julienvincent/hunk.nvim

I wanted to make it a bit discoverable so figured it would be appropriate to add it here.

# Checklist

If applicable:

### DIFF
--- a/docs/community_tools.md
+++ b/docs/community_tools.md
@@ -18,6 +18,12 @@ workflows quite easy. Take a look at its README.md as it quite descriptive.
 
 Find it [here][gg].
 
+## Hunk.nvim
+
+Hunk.nvim is a Neovim based diff-editor for Jujutsu which can be used as an
+alternative to the default `:builtin` diff-editor.
+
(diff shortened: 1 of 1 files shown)


docs/community_tools.md
```diff
@@ -52,4 +59,5 @@ You can find other community contributed tools and integrations in our
 [jj_tui]: https://github.com/faldor20/jj_tui
 [lazyjj]: https://github.com/Cretezy/lazyjj
 [vjj]: https://github.com/noahmayr/vjj
+[hunk.nvim]: https://github.com/julienvincent/hunk.nvim
```
PhilipMetzger, line 62: These are sorted alphabetically, so this should be too.

PhilipMetzger (approved):
Thanks, a minor Nit though.

Issue #4108, opened 2024-07-17
jj git fetch accidentally deleted parent commits


allonsy:
## Description
Let's say I have a working copy with multiple parents as follows:
```
>>> jj log 
@    yyulnrot linuxbash8@gmail.com 2024-07-17 20:54:51 e7cd9783
├─╮  (empty) (no description set)
│ ◉  ulnnnzwl linuxbash8@gmail.com 2024-07-17 20:53:33 feat_2 e91d496b
│ │  implement feature two
◉ │  wonnuqqo linuxbash8@gmail.com 2024-07-17 20:52:41 feat_1 c955bbca
├─╯  implement feature one
◉  stnttlrq linuxbash8@gmail.com 2024-07-17 20:51:51 main 9a68623d
│  initial
~

>>> jj status
The working copy is clean
Working copy : yyulnrot e7cd9783 (empty) (no description set)
Parent commit: wonnuqqo c955bbca feat_1 | implement feature one
Parent commit: ulnnnzwl e91d496b feat_2 | implement feature two
```

This graph shows a working copy working on two different branches simultaneously. This has been set up by declaring the working copy with two parents, feat_1 and feat_2 (merge commit).

Now, let's say, in a side channel, like github, the PR for feature 2 is approved and squash merged into main and then the remote branch feat_2 is deleted on github. 

If I run `jj git fetch` and the `jj log` and `jj status` this is the new state of the repository:
```
>>> jj git fetch
branch: feat_2@origin [deleted] untracked
branch: main@origin   [updated] tracked
Abandoned 1 commits that are no longer reachable.
Rebased 1 descendant commits
Working copy now at: yyulnrot 864cb8db (empty) (no description set)
Parent commit      : wonnuqqo c955bbca feat_1 | implement feature one
Parent commit      : stnttlrq 9a68623d initial
Added 0 files, modified 0 files, removed 1 files

>>> jj log
@    yyulnrot linuxbash8@gmail.com 2024-07-17 21:11:52 864cb8db
├─╮  (empty) (no description set)
◉ │  wonnuqqo linuxbash8@gmail.com 2024-07-17 20:52:41 feat_1 c955bbca
├─╯  implement feature one
│ ◉  opuxpukk linuxbash8@gmail.com 2024-07-17 21:11:41 main 2ce2a2db
├─╯  implement feature two (#1)
◉  stnttlrq linuxbash8@gmail.com 2024-07-17 20:51:51 9a68623d
│  initial
~

[...]

PR #4143, opened 2024-07-24
git: on abandoning unreachable commits, don't count HEAD ref

yuja:
This basically reverts 20eb9ecec10f "git: don't abandon HEAD commit when it loses a branch." I think the new behavior is more consistent because the Git HEAD is equivalent to @- in jj, so it shouldn't be considered a named ref.

Note that we've made old HEAD branch not considered at 92cfffd84377 "git: on external HEAD move, do not abandon old branch."

#4108

# Checklist

If applicable:


martinvonz (approved):
Thanks

Issue #3809, opened 2024-06-01
FR: Use `-p`/`--path` everywhere for path arguments


arxanas:
**Is your feature request related to a problem? Please describe.**

I am constantly running `jj log <rev>` and then it silently doesn't work/hangs because `jj` is busy trying to find paths that match `<rev>` in the commit history. This is a common footgun and we've implemented a few checks to try to detect path arguments that look like revset arguments, but it continues to happen.

There's not a fundamental reason why paths are "more important" than revsets for most commands, so we should stop treating them as a "default" kind of argument for most (if not all) commands.

I believe a lot of the path arguments are positional primarily for historical reasons (namely, Git and Mercurial did it that way).

**Describe the solution you'd like**
- In all cases, where positional path arguments are currently accepted, arguments of the form `-p`/`--path` should also be accepted.
- In cases where positional path arguments are currently accepted, we should probably deprecate that and require the user to specify with them with `-p`/`--path`.

**Describe alternatives you've considered**
- Note that revsets also support querying commits that touch paths, which is a related feature, but not entirely the same. The details about which path was specifically selected carries useful semantic information.
  - For example, `git log --patch -- <path>` will limit the shown diff to only the parts touching the provided paths. I'm not sure if jj does anything similar at present.

**Additional context**
N/A

yuja:
For reference, there's a previous discussion in https://github.com/martinvonz/jj/issues/2554

BatmanAoD:
For `log` especially, this behavior was very confusing to me, coming from Git. The only way I'm aware of in Git to filter commits by which paths they've touched is the `-p` argument, which is actually a `git diff` flag that _just happens_ to have the desired effect on `git log`. (This behavior is confusing enough that it resulted in [my most popular StackOverflow answer](https://stackoverflow.com/a/22412252/1858225) by far.)

PR #4146, opened 2024-07-24
cli: rename `obslog` to `evolution`

martinvonz:
# Checklist

If applicable:


PhilipMetzger (approved):
🎉

thoughtpolice (approved):
Nice. BTW, I wonder if `jj evolution` should be able to take a revset argument without requiring `-r` like several other commands do... We can save that for a follow up though, and the `-r` handling of those other commands I mention need to be fixed up a little.

istudyatuni:
What do you thing about command like `jj changeset log`?

PhilipMetzger:
> I wonder if `jj evolution` should be able to take a revset argument without requiring `-r` like several other commands do... We can save that for a follow up though, and the `-r` handling of those other commands I mention need to be fixed up a little.

Which commands do you mean? I probably missed a conversation somewhere. 

> What do you thing about command like `jj changeset log`?

Let's not open the `revision`, `Change` bikeshed here.

martinvonz:
> What do you thing about command like `jj changeset log`?

That sounds too similar to what `jj log` is to me. Mercurial uses `changeset` as a synonym for `revision` and `commit`, so I would think of it as a log of commits in the usual sense (linking to parents, not to predecessors). I also don't want to create a `jj changeset` command family until we have other subcommands for it.

I'll leave this PR open for a while to give you and others more opportunity to comment.

ilyagr (approved):
This SGTM, it should also be easy to access as `jj evo<TAB>`, assuming we don't plan to have any other `evolution`-related commands. I can't really think of any at the moment.

If we could think of some, I'd go for `jj evolution log`, `jj evolog` for short.

yuja:
My only concern is that `jj evolution` sounds like a namespace like `jj config`. `log` isn't a verb, but it's common enough to recognize as a command to "show log".

martinvonz:
> My only concern is that `jj evolution` sounds like a namespace like `jj config`. `log` isn't a verb, but it's common enough to recognize as a command to "show log".

I agree with that but I couldn't think of a better name that people seemed to like. I'd be fine with `jj evolution-log` since I rely heavily on tab completion anyway, but maybe that's annoyingly long for people who can't use tab completion for some reason. OTOH, maybe `jj evolution` is already a bit long for them. Anyone knows?

seh:
If we're looking for verbs, there is also _chronicle_, which happens to also be a noun.

joyously:
Looking up the synonyms of evolution, I found `progression`.
For etymology, there is `derivation`, `development`, and `origin`.
For chronicle, there is `report`.

yuja:
> I'd be fine with `jj evolution-log` since I rely heavily on tab completion anyway, but maybe that's annoyingly long for people who can't use tab completion for some reason. OTOH, maybe `jj evolution` is already a bit long for them.

I slightly prefer `evolution-log` than `evolution` as the command output is quite similar to `jj log`. I would also do `jj evo[TAB]` anyway. `evolution` isn't easy to type.

avamsi:
Does `revlog` also sound too similar to `log`?

joyously:
> Does `revlog` also sound too similar to `log`?

Yes, but also, it would mean the same thing. This is for the command that shows how "changes" have evolved (through different revisions).
But as Philip said
> Let's not open the `revision`, `Change` bikeshed here.

avamsi:
> This is for the command that shows how "changes" have evolved (through different revisions).

I'm aware. Save for potential confusion with `log`, I do think `revlog` can reasonably be read as evolution of a change.
I'm also amused by how it sounds similar to `reflog`, even though they do different things.

seh:
Another couple of colors to consider for the shed: _change-history_ and _revision-history_. These are nouns, though.

PhilipMetzger:
For everyone here's the prior discussion: https://github.com/martinvonz/jj/discussions/3592

joyously:
What if it's just `log` with a flag `--evolution`?

ilyagr:
I wouldn't mind `evolution-log` or `evolutionlog`. If we go with the former, I wonder whether we should have "inter-diff" and "diff-edit", but that's a separate discussion (while I'd like eventual consistency on whether we use dashes or not, the possibility of tab completion and aliases makes it less important practically). Having `evolution log` would avoid this discussion, even if `log` is the only sub-command.

Either way, I like the "evo" prefix, and would like `evolog` as a visible and shorter alias for any of these three, unless people find it problematic. For one, if I'm talking on Discord, `evolog` is nicer to type. Also, "evo" would work for either https://github.com/martinvonz/jj/issues/4129#issuecomment-2241471661 (`evoheads`, `evoancestors`, etc) or the suggestion by Yuya in the following comment in that discussion (`evo(::x)` instead of `evoancestors(x)`).

Looking back at the discussion Philip linked, I still like "rewritelog" a bit, but less. The problem with that is that `rewrite(::x)` or `rewriteancestors` makes less sense to me than `evo(::x)`. The noun `rewrite` is ambiguous with the verb in many combinations. `evolvelog` is shorter than `evolutionlog`, but I think is even more problematic since `evolve` is clearly a verb.

Even though `jj evolution` sounds like a namespace, I'm OK with it too. I think it'd work with `evoheads`/`evo(x-)`, it's a nice and easy-to-remember word, and it avoids the dash-or-no-dash discussion. As I already said, it would be a mildly annoying transition if we later decided to make it into an actual namespace with several commands (it would negate all the advantages of choosing `jj evolution` now), but it is not an impossible transition.

ilyagr:
> What if it's just log with a flag --evolution

I think this is also a good idea. It would provide a good UX if we can make it work.

The downside is that we'd have to make `obslog` work with all the options to `log`, present and future. `log` already has a ton of options and the number will only grow. On one hand, that's annoying, and on the opposite hand we'd be making it harder to implement new `log` options by forcing anyone to think how it should work with `obslog`.

So, my first reaction is that I think I'd prefer `jj evolution log` to this, but I also think `--evolution` is worth considering.

ilyagr:
> BTW, I wonder if jj evolution should be able to take a revset argument without requiring -r like several other commands do...

I wonder whether it should take a path argument, like `jj log`. Even if that's not implemented, I think requiring `-r` saves a place for it (unless we go with #3809 or something like it).

joyously:
> The downside is that we'd have to make `obslog` work with all the options to `log`

I don't think all the options would have to apply, if you make them mutually exclusive. The templates would need to match maybe?

PhilipMetzger:
If we don't take the `evolution` namespace, then my preferred name would be `evolog` from all suggestions.

> What if it's just `log` with a flag `--evolution`?

While this works, I'd rather not mix concepts as the "Change's" evolution is separate from a repo's history. I also think this will hinder learning Jujutsu.

martinvonz:
I'm going to go with `evolution-log` with `evolog` as alias then. Let me know what you think (e.g. a simple thumbs up/down reaction) before I make the change in a few days or so.

martinvonz:
I've updated this to use `jj evolog` and `jj evolution-log` now

emilazy:
I personally rather liked just `evolution`, considerably more than `evolog`. (But I should have spoken up sooner; I don’t want to prolong the bikeshedding…)

cli/src/commands/evolog.rs
```diff
@@ -35,7 +35,7 @@ use crate::ui::Ui;
 ///
 /// Name is derived from Merciual's obsolescence markers.
```
bnjmnt4n, line 36: Should this be removed with the rename?

martinvonz: Good catch. Fixed.

cli/src/commands/mod.rs
```diff
@@ -91,6 +91,8 @@ enum Command {
     Diffedit(diffedit::DiffeditArgs),
     Duplicate(duplicate::DuplicateArgs),
     Edit(edit::EditArgs),
+    #[command(alias = "obslog")]
+    Evolog(evolog::EvologArgs),
```
ilyagr, line 95: Should there be an alias for `evolution-log` as well?

I'm OK with or without it, but if we don't have it, the commit description and the CHANGELOG need editing.

martinvonz: Oops, thanks for catching that.

martinvonz:
I'll merge this now

PR #4149, opened 2024-07-24
run: Treat `jj run -j0 <command>` as use all available cores

PhilipMetzger:
Waleed noticed this in #4021.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/run.rs
+++ b/cli/src/commands/run.rs
@@ -59,9 +59,8 @@ pub fn cmd_run(ui: &mut Ui, command: &CommandHelper, args: &RunArgs) -> Result<(
     // 2. the amount of cores available.
     // 3. a single job, if all of the above fails.
     let _jobs = match args.jobs {
-        Some(0) => return Err(user_error("must pass at least one job")),
+        Some(0) | None => std::thread::available_parallelism().map(|t| t.into()).ok(),
         Some(jobs) => Some(jobs),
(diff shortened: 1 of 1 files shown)


cli/src/commands/run.rs
```diff
@@ -59,7 +59,7 @@ pub fn cmd_run(ui: &mut Ui, command: &CommandHelper, args: &RunArgs) -> Result<(
     // 2. the amount of cores available.
     // 3. a single job, if all of the above fails.
     let _jobs = match args.jobs {
-        Some(0) => return Err(user_error("must pass at least one job")),
+        Some(0) => std::thread::available_parallelism().map(|t| t.into()).ok(),
         Some(jobs) => Some(jobs),
         None => std::thread::available_parallelism().map(|t| t.into()).ok(),
```
arxanas, line 64:
```suggestion
        Some(0) | None => std::thread::available_parallelism().map(|t| t.into()).ok(),
        Some(jobs) => Some(jobs),
```

suggestion: Can fold these cases together

arxanas (approved):

cli/src/commands/run.rs
PhilipMetzger: I initially did that but clippy treated it as unreachable. See https://github.com/martinvonz/jj/actions/runs/10081922624/job/27875038811

PhilipMetzger: Ok, it was a ordering problem.

Issue #4115, opened 2024-07-18
Can't find libgit2 on Mac/Homebrew


bklebe:
## Steps to Reproduce the Problem

1. Run a `brew upgrade` for libgit2

## Expected Behavior
`jj` works after the upgrade.

## Actual Behavior
While upgrading other packages, Homebrew reports:
```
==> Checking for dependents of upgraded formulae...
Disable this behaviour by setting HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK.
Hide these hints with HOMEBREW_NO_ENV_HINTS (see `man brew`).
==> Reinstalling 1 dependent with broken linkage from source:
jj
==> Fetching dependencies for jj: pkg-config, libgit2@1.7, z3, llvm and rust
==> Fetching pkg-config
==> Downloading https://ghcr.io/v2/homebrew/core/pkg-config/manifests/0.29.2_3
############################################################################################ 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/pkg-config/blobs/sha256:7b59abc0b5381065b1eab17421
############################################################################################ 100.0%
==> Fetching libgit2@1.7
==> Downloading https://ghcr.io/v2/homebrew/core/libgit2/1.7/manifests/1.7.2
############################################################################################ 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/libgit2/1.7/blobs/sha256:0cad95e1f543ee0a6af1472c5
############################################################################################ 100.0%
==> Fetching z3
==> Downloading https://ghcr.io/v2/homebrew/core/z3/manifests/4.13.0
Already downloaded: /Users/bklebe/Library/Caches/Homebrew/downloads/f76733dc8564492bde6da5ddd68bf485a259b09d276de374047a4600ed52edff--z3-4.13.0.bottle_manifest.json
==> Downloading https://ghcr.io/v2/homebrew/core/z3/blobs/sha256:d7a6ead5ea667caa7a45e1a623ecd386b0
Already downloaded: /Users/bklebe/Library/Caches/Homebrew/downloads/b83fa24637ba159214807443e0dada9de32c67d638c253cfb090a20ae9cfb096--z3--4.13.0.arm64_sonoma.bottle.tar.gz
==> Fetching llvm
==> Downloading https://ghcr.io/v2/homebrew/core/llvm/manifests/18.1.8
############################################################################################ 100.0%
[...]

Issue #2896, opened 2024-01-30
Starting with v0.13, the official Mac (x86_64) binary no longer works on Apple Silicon Macs via Rosetta.


ilyagr:
[v0.12 official binary](https://github.com/martinvonz/jj/releases/download/v0.12.0/jj-v0.12.0-x86_64-apple-darwin.tar.gz) works fine after Rosetta 2 is installed.

[v0.13 official binary](https://github.com/martinvonz/jj/releases/download/v0.13.0/jj-v0.13.0-x86_64-apple-darwin.tar.gz) results in:

```
$ cd jj-v0.13.0-x86_64-apple-darwin/
$ ./jj
dyld[82582]: Library not loaded: /usr/local/opt/libgit2/lib/libgit2.1.7.dylib
  Referenced from: <C7E5CA77-7147-3BB5-A67F-CAAC76589A3A> /Users/ilyagr/tmp-nobackup/jj-v0.13.0-x86_64-apple-darwin/jj
  Reason: tried: '/usr/local/opt/libgit2/lib/libgit2.1.7.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/opt/libgit2/lib/libgit2.1.7.dylib' (no such file), '/usr/local/opt/libgit2/lib/libgit2.1.7.dylib' (no such file), '/usr/local/lib/libgit2.1.7.dylib' (no such file), '/usr/lib/libgit2.1.7.dylib' (no such file, not in dyld cache)
fish: Job 1, './jj' terminated by signal SIGABRT (Abort)
```

Did we start linking libgit2 dynamically?

My machine is an M2 Mac, running a recent MacOS Sonoma.

```
$ uname -a
Darwin macaw.local 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:55:06 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6020 arm64 arm Darwin
```

-----

This will be less annoying after the next release if https://github.com/martinvonz/jj/pull/2895 gets merged, but it's still a little annoying.

ilyagr:
I'm not sure what's going on. I may have missed something, but I couldn't immediately find any relevant changes in `Cargo.toml` or `release.yml`. Perhaps something changed with the "git2" library's default features?

thoughtpolice:
[...]

PR #4163, opened 2024-07-26
cargo: enable vendored-libgit2, document how to properly dynamically link libgit2

ilyagr:
This PR succeeds #4120. 

This changes less than it seems. Our CI builds already mostly linked a vendored copy of libgit2. This is because before this commit, it turns out that `git2` could link `libgit2` *either* statically or dynamically based on whether it could find a version of libgit2 it liked to link dynamically. Our CI builds usually did not provide such a version AFAIK.

This made the kind of binary `cargo install` would produce unpredictable and may have contributed to #2896. I was once very surprised when I did `brew upgrade libgit2` and then
`cargo build --release` suddenly switched from building dynamically linked `jj` to the vendored version.

Instead, if a packager wants to link `libgit2` dynamically, they should set an environment variable, as described inside the diff of this commit. I also think we should recommend static linking as `git2` is quite picky about the versions of `libgit2` it supports. See also https://github.com/rust-lang/git2-rs/pull/1073

This might be related to #4115.

Also cc #4080, though I think this is orthogonal to upgrading out `git2` version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,16 @@ dunce = "1.0.4"
 either = "1.13.0"
 esl01-renderdag = "0.3.0"
 futures = "0.3.30"
-git2 = "0.18.3"
+git2 = { version = "0.18.3", features = [
+    # Do *not* disable this feature even if you'd like dynamic linking. Instead,
+    # set the environment variable `LIBGIT2_NO_VENDOR=1` if dynamic linking must
+    # be used (this will override the Cargo feature), and allow static linking
+    # in other cases. Rationale: If neither the feature nor the environment
+    # variable are set, `git2` may still decide to vendor `libgit2` if it
+    # doesn't find a version of `libgit2` to link to dynamically. See also
+    # https://github.com/rust-lang/git2-rs/commit/3cef4119f
+    "vendored-libgit2"
+] }
 gix = { version = "0.63.0", default-features = false, features = [
     "index",
     "max-performance-safe",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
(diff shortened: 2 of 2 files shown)


yuja (approved):
(I'm not a big fan of static linking, but) this should be better than `./configure`-like auto detection.

Cargo.toml
```diff
@@ -48,7 +48,16 @@ dunce = "1.0.4"
 either = "1.13.0"
 esl01-renderdag = "0.3.0"
 futures = "0.3.30"
-git2 = "0.18.3"
+git2 = { version = "0.18.3", features = [
+    # Do *not* disable this feature even if you'd like dynamic linking. Instead,
+    # set the environment variable `LIBGIT2_NO_VENDOR=1` if dynamic linking must
```
martinvonz, line 53: Should `LIBGIT2_NO_VENDOR=1` be set in `flake.nix` and in https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/version-management/jujutsu/default.nix after this PR? @thoughtpolice

ilyagr: Good point! I tried to inform Homebrew packagers in #4080, I wonder whether there are other people to inform and/or better ways to inform them.

I checked, and indeed Nixpkg's `jj` is successfully dynamically linked to `libgit2`. I guess/hope Nix has no problem supporting multiple versions of libgit2 at once; it will need a different version after #4080.

ilyagr: I could add a "Note to packagers" or "Breaking changes for packagers" to the changelog.

ilyagr: This is now done. I think that's good enough. If #4080 gets in, we should mention that libgit2 0.18 is needed for dynamic linking as well; this would also affect Nixpkgs.

martinvonz: @thoughtpolice or @necauqua may want to follow up with a change to `flake.nix`

martinvonz (approved):

PR #4165, opened 2024-07-26
cli: merge op heads and snapshot working copy by "op log" by default

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -327,6 +327,12 @@ impl CommandHelper {
             .map_err(|err| map_workspace_load_error(err, self.global_args.repository.as_deref()))
     }
 
+    /// Returns true if the working copy to be loaded is writable, and therefore
+    /// should usually be snapshotted.
+    pub fn is_working_copy_writable(&self) -> bool {
+        self.is_at_head_operation() && !self.global_args.ignore_working_copy
+    }
+
     /// Returns true if the current operation is considered to be the head.
     pub fn is_at_head_operation(&self) -> bool {
         // TODO: should we accept --at-op=<head_id> as the head op? or should we
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -300,8 +300,7 @@ impl CommandHelper {
         let workspace = self.load_workspace()?;
         let op_head = self.resolve_operation(ui, workspace.repo_loader())?;
         let repo = workspace.repo_loader().load_at(&op_head)?;
-        let loaded_at_head = self.global_args.at_operation == "@";
-        WorkspaceCommandHelper::new(ui, self, workspace, repo, loaded_at_head)
+        WorkspaceCommandHelper::new(ui, self, workspace, repo, self.is_at_head_operation())
     }
 
     pub fn get_working_copy_factory(&self) -> Result<&dyn WorkingCopyFactory, CommandError> {
@@ -328,13 +327,26 @@ impl CommandHelper {
             .map_err(|err| map_workspace_load_error(err, self.global_args.repository.as_deref()))
     }
 
+    /// Returns true if the current operation is considered to be the head.
+    pub fn is_at_head_operation(&self) -> bool {
+        // TODO: should we accept --at-op=<head_id> as the head op? or should we
+        // make --at-op=@ imply --ignore-workign-copy (i.e. not at the head.)
+        matches!(self.global_args.at_operation.as_deref(), None | Some("@"))
+    }
+
+    /// Resolves the current operation from the command-line argument.
+    ///
+    /// If no `--at-operation` is specified, the head operations will be
+    /// loaded. If there are multiple heads, they'll be merged.
     #[instrument(skip_all)]
     pub fn resolve_operation(
         &self,
         ui: &mut Ui,
         repo_loader: &RepoLoader,
     ) -> Result<Operation, CommandError> {
-        if self.global_args.at_operation == "@" {
+        if let Some(op_str) = &self.global_args.at_operation {
+            Ok(op_walk::resolve_op_for_load(repo_loader, op_str)?)
+        } else {
             op_heads_store::resolve_op_heads(
                 repo_loader.op_heads_store().as_ref(),
                 repo_loader.op_store(),
@@ -365,10 +377,6 @@ impl CommandHelper {
                         .clone())
                 },
             )
-        } else {
-            let operation =
-                op_walk::resolve_op_for_load(repo_loader, &self.global_args.at_operation)?;
-            Ok(operation)
         }
     }
 
@@ -2420,10 +2428,14 @@ pub struct GlobalArgs {
     /// Operation to load the repo at
     ///
     /// Operation to load the repo at. By default, Jujutsu loads the repo at the
-    /// most recent operation. You can use `--at-op=<operation ID>` to see what
-    /// the repo looked like at an earlier operation. For example `jj
-    /// --at-op=<operation ID> st` will show you what `jj st` would have
-    /// shown you when the given operation had just finished.
+    /// most recent operation, or at the merge of the concurrent operations if
+    /// any.
+    ///
+    /// You can use `--at-op=<operation ID>` to see what the repo looked like at
+    /// an earlier operation. For example `jj --at-op=<operation ID> st` will
+    /// show you what `jj st` would have shown you when the given operation had
+    /// just finished. `--at-op=@` is pretty much the same as the default except
+    /// that concurrent operations will never be merged.
     ///
     /// Use `jj op log` to find the operation ID you want. Any unambiguous
     /// prefix of the operation ID is enough.
@@ -2435,8 +2447,8 @@ pub struct GlobalArgs {
     /// earlier operation. Doing that is equivalent to having run concurrent
     /// commands starting at the earlier operation. There's rarely a reason to
     /// do that, but it is possible.
-    #[arg(long, visible_alias = "at-op", global = true, default_value = "@")]
-    pub at_operation: String,
+    #[arg(long, visible_alias = "at-op", global = true)]
+    pub at_operation: Option<String>,
     /// Enable debug logging
     #[arg(long, global = true)]
     pub debug: bool,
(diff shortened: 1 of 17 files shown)


cli/src/cli_util.rs
```diff
@@ -2420,10 +2433,12 @@ pub struct GlobalArgs {
     /// Operation to load the repo at
     ///
     /// Operation to load the repo at. By default, Jujutsu loads the repo at the
-    /// most recent operation. You can use `--at-op=<operation ID>` to see what
-    /// the repo looked like at an earlier operation. For example `jj
-    /// --at-op=<operation ID> st` will show you what `jj st` would have
-    /// shown you when the given operation had just finished.
+    /// most recent operation. If there are concurrent operations, they will be
```
ilyagr, line 2436: (Optional) There is a bit of mixed tenses here: "loads ... they will be merged". One way to phrase this:

I'd reword this a bit:

> By default, Jujutsu loads the repo at the most recent operation, or at the merge of the latest concurrent operations if there is more than one.

I'd also consider adding a paragraph break.

cli/src/cli_util.rs
```diff
@@ -2420,10 +2433,12 @@ pub struct GlobalArgs {
...
+    /// merged. You can use `--at-op=<operation ID>` to see what the repo looked
+    /// like at an earlier operation. For example `jj --at-op=<operation ID> st`
+    /// will show you what `jj st` would have shown you when the given operation
+    /// had just finished. `--at-op=@` is pretty much the same as the default
+    /// except that concurrent operations will never be merged.
```
ilyagr, line 2441: I find it a bit confusing that `--at-op=@` does not seem to imply `--ignore-working-copy`. Is this important, or can we make it more consistent by making it imply `--ignore-working-copy`?

If so, I'd delete this sentence and change one of the paragraphs below to:


> Using `--at-op` has two effects in addition to loading an earlier operation. First, the working copy is ignored, as if `--ignore-working-copy` was specified. Second, `jj` will *not* merge concurrent operations, if any, like it does by default.

cli/src/commands/operation/log.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::slice;
+
 use jj_lib::op_walk;
 
 use crate::cli_util::{format_template, CommandHelper, LogContentFormat};
```
ilyagr, line 19: Perhaps add some docs:

> Show the operation log
>
> Like other `jj` operations, by default `jj op log` may snapshot the working copy or merge most recent concurrent operations. If you'd like to make sure the operation is read-only, use `jj op log --at-op @ --ignore-working-copy`.

Also, if there is concurrency, it seems there is no way to see both concurrent operations in on `jj op log` without merging them, is there? It would be nice if `jj op log --at-op @` showed both (I guess it'd happen at the auto-merge of the concurrent operations without recording the merge), but I'm guessing now it'd be an error message (or did you figure out a better solution?).

If that's correct, I suppose we could add something like

> If there is no unique latest operation, `jj op log --at-op @` will produce an informative error message that gives ids for the most recent concurrent operations.

cli/src/cli_util.rs
ilyagr: One  confusing implication (even if possibly rarely in play) would be that if `abc` happens to refer to `@`, then `--at-op=abc` will not imply `--ignore-working-copy`, but once another operation happens, the same option will imply `--ignore-working-copy`.

cli/src/cli_util.rs
yuja: Thanks. Updated as
```
    /// Operation to load the repo at. By default, Jujutsu loads the repo at the
    /// most recent operation, or at the merge of the concurrent operations if
    /// any.
    ///
    /// You can use `--at-op=<operation ID>` to see what the repo looked like at
```

cli/src/cli_util.rs
yuja: I don't think `--at-op=@` should disable snapshot. And I personally think it's a bug that `--at-op=<head_id>` implies `--ignore-working-copy`.

The idea is that `--at-op=<no_head>` implies `--ignore-working-copy` because the working copy doesn't represent the specified operation.

cli/src/commands/operation/log.rs
yuja: I think the goal is to remove special case and docs about the special case (such as `op log` doesn't snapshot but `op diff` does.)

Maybe we can add more detailed hint if `op log --at-op=@` failed to load the concurrent heads, but I assume the situation would be pretty rare.

cli/src/cli_util.rs
ilyagr: >  I personally think it's a bug that --at-op=<head_id> implies --ignore-working-copy.

I can understand this from a purity standpoint. I find the idea of "fixing" this bug quite confusing though; I can't tell how bizarre the resulting behavior of `jj log --at-op @-` would be. Presumably, the result would be:

```
@ merge
| \
|  o new_snapshot
|  |
o  | former_@
| /
o  former_@-
```

whereas the user would likely expect (if they expect the operation to snapshot at all)

```
@ new_snapshot
o former_@
o former_@-
```

I do not understand merging well enough to understand whether the two resulting states `@` are the same or not. If not, I imagine the difference would be quite confusing.

------------------

Assuming you see a solution to the above problem, I'd be OK with "fixing the bug". If we don't fix the bug, I'd prefer consistently ignoring the working copy always, except perhaps if there is some use-case where snapshotting it is helpful, for example because of the confusing situation that you can't tell whether `jj log --at-op abc` will ignore working copy without knowing what `abc` is.

All that said, I'll ultimately leave this up to you and Martin, you've been discussing this for a while and my understanding of these issues feels a bit superficial at the moment.

cli/src/commands/operation/log.rs
ilyagr: > Maybe we can add more detailed hint if op log --at-op=@ failed to load the concurrent heads

I think that would be a fine alternative. If `--at-op=@` does not imply `--ignore-working-copy`, that might need a detailed hint too. Ultimately, I'll leave this up to you as well, but here's my motivation.

My main concern is that we might want to pay special attention to commands used to debug or recover from broken states. Knowing which commands are read-only can be critical in this case.

More philosophically, I think that beginner or intermediate `jj` users (especially early adopters) would often get into broken states. We should be motivated to help them, not just out of benevolence, but also because we disproportionately benefit from bug reports such users provide. So, it's in our interest to give them all the guidance we can on how to get meaningful debug info and how to recover. But I may be biased, since I tend to put myself into weird situations often.

cli/src/cli_util.rs
ilyagr: Or would `jj log --at-op @-` just fail (if it `--at-op` does not imply `--ignore-working-copy`) because after the snapshot and the merge, `@-` would no longer refer to `former_@-` as the user expects, but refers to two operations?

yuja: Maybe I don't follow your point, but `jj --at-op @-` fails if there are more than one  op heads. That's unchanged. This patch makes it even stricter, `jj --at-op @` now fails.

Oh, maybe you mean `jj --at-op @-` could do snapshot (if `@-` can be resolved to a single operation)? I don't think enabling snapshot makes sense in that situation (though it's technically possible if the working copy tree is identical to the `@-` one.)

What I meant is `jj --at-op @` and `jj --at-op <head-id>` (where `<head-id>` points to the current operation head) don't have to imply `--ignore-working-copy`.

ilyagr: I guess I misunderstood you. My understanding was that you suggested that `--at-op anything` shouldn't imply `--ignore-working-copy`, but I see that was wrong.

I also thought that with your implementation, neither `--at-op @` nor `--at-op <head_id>` implied `--ignore-working-copy`. It sounds like what you were actually saying is that while `--at-op @` *does not* imply `--ignore-working-copy`, `--at-op <head_id>` *does*, and *that*'s what you thought was a bug. 

To me, it would seem that there is no good choice for `--at-op <head_id>` if `--at-op <another_id>` and `--at-op @` have different behavior. If we **had** to choose I'd agree with you that making it behave the same as `--at-op @` is probably better, but it doesn't seem good.

This is why I thought it'd be simpler to make `--at-op <anything>` imply `--ignore-working-copy`. The only downside I see is that it's hard to have the behavior of "do snapshot but don't merge concurrent heads", but I don't know when that would be desired. If there is a use-case, we could perhaps make `jj debug snapshot --at-op <anything>` not `--ignore-working-copy`, which would cover that and other weird cases (for testing snapshotting maybe? I'm unsure).

But again, I could easily be missing something important still.

yuja: > while `--at-op @` _does not_ imply `--ignore-working-copy`, `--at-op <head_id>` _does_, and _that_'s what you thought was a bug.

Correct. It wouldn't practically matter, but seemed a bit odd.

> This is why I thought it'd be simpler to make `--at-op <anything>` imply `--ignore-working-copy`. The only downside I see is that it's hard to have the behavior of "do snapshot but don't merge concurrent heads",

I don't think this operation mode (= don't merge ops, do snapshot) is useful, but I don't think `--at-op=@` should disable snapshot either. These two flags are conceptually different. That said, I'm not strongly against your idea because the implementation will get simpler.

Updated the todo comment.

martinvonz (approved):
This all looks good to me, FWIW.

cli/src/commands/operation/log.rs
yuja: Update the help text. Can you take another look?
```
/// Like other commands, `jj op log` snapshots the current working-copy changes
/// and merges concurrent operations. Use `--at-op=@ --ignore-working-copy` to
/// inspect the current state without mutation.
```

ilyagr: This text looks good to me, thank you!

ilyagr: Although, it still bothers me that if the merge of concurrent operations is failing, there's no way to see both concurrent operations at once, since `--at-op @` will give an error. Hopefully that would be rare?

yuja: > Hopefully that would be rare?

I believe it is. Typical use case is `op log` or `op log --ignore-working-copy`, and concurrent ops get merged. `--at-op=@` is needed only when merge fails (i.e. repository is somewhat corrupted.)

PR #4168, opened 2024-07-26
docs: update tutorial, try to clarify conflict markers

ilyagr:


docs/tutorial.md
```diff
@@ -150,25 +149,36 @@ commit ID, which changes when you rewrite the commit. You can give either ID
...
+for specifying revisions. For example, `@` refers to the working-copy commit,
+`root()` refers to the root commit, `trunk()` refers to the repo's main branch
+(`master` in this case, as determined by `jj git clone`), and `trunk()..` refers
+to all commits that are not ancestors of the main branch. We can combine
+expressions with `|` for union, `&` for intersection and `~` for difference. For
+example:
 
 ```shell
-$ jj log -r '@ | root() | branches()'
+$ jj log -r 'root() | trunk() | trunk().. | @'
```
ilyagr, line 164: I'm wondering whether a larger change to the tutorial is needed now that `jj log` does not list remote branches by default, and now that the remote branches are immutable by default. It seems important to explain `jj log -r all()` and perhaps how to make the remote branch commits mutable. In fact, I'm not sure what is the minimum amount of info to explain to help people not get confused.

I'm considering removing the first commit from this PR until we get a better idea here.

docs/tutorial.md
```diff
@@ -282,14 +281,26 @@ Added 0 files, modified 0 files, removed 1 files
...
 b2
->>>>>>>
-$ echo resolved > file1
+>>>>>>> Conflict 1 of 1 ends
+```
+
+`file1` now contains a [representation of the
+conflict](conflicts.md#conflict-markers). We can resolve the conflict by editing
+the file[^resolve] to remove all the conflict markers and making it into what
+the resolved version should look like. Once we've resolved the conflict, we'll
```
amiryal, line 296: > making it into what the resolved version should look like

This may be too vague. I think, with just a few more words, we can explain here more explicitly what the resolved version is expected to look like. I am thinking something along these lines:

> We can resolve the conflict by manually making the changes (as indicated by sections marked by `%%%%%%%`) in the section marked by `+++++++`, and then removing all extraneous sections and markers.

docs/conflicts.md
```diff
@@ -85,22 +85,30 @@ you, making it easier to spot the differences to apply to the other side. Here's
...
 
 As in Git, the `<<<<<<<` and `>>>>>>>` lines mark the start and end of the
-conflict. The `%%%%%%%` line indicates the start of a diff. The `+++++++`
-line indicates the start of a snapshot (not a diff).
+conflict. The `%%%%%%%` line indicates the start of a diff. The `+++++++` line
+indicates the start of a snapshot (not a diff). The text after the conflict
+markers is merely a guide for the human reader.
+
+The base of the conflict corresponds to the base of Git's conflict, and the two
+sides correspond to the left/right sides of Git's conflict in some order.
```
amiryal, line 107: The drawing of similarities with Git should have ended before this point. Suffice to say here that the assignment of the sides to `#1` and `#2` convey no intended meaning.

docs/conflicts.md
```diff
@@ -85,22 +85,30 @@ you, making it easier to spot the differences to apply to the other side. Here's
...
+Currently, jj's conflict markers do not say which of the sides is the "left"
+side and which is the "right" side, nor do they say which commit each side comes
+from, see [#1176](https://github.com/martinvonz/jj/issues/1176) for progress on
+this issue. For now, `jj` merely calls the two sides "side #1" and "side #2".
```
amiryal, line 111: This discussion and reference to the issue is too much information. If the documentation has a stated goal of inviting discussion and contribution, then it’s fine to leave it here, but otherwise I think the documentation could do without this extra context.

amiryal (commented):
Every one of my comments should be mentally prefixed with “IMO” and suffixed with “WDYT?”, of course. ☺

PR #4172, opened 2024-07-28
cli: split commands/workspace.rs into sub modules

yuja:
# Checklist

If applicable:


cli/src/commands/workspace/update_stale.rs
```diff
@@ -0,0 +1,170 @@
...
+                    op_id.clone(),
+                    op_data,
+                ))?,
+                false,
+            ),
+            Err(e @ OpStoreError::ObjectNotFound { .. }) => {
+                writeln!(
+                    ui.status(),
+                    "Failed to read working copy's current operation; attempting recovery. Error \
+                     message from read attempt: {e}"
```
PhilipMetzger: Nit: Use a raw string here.

PhilipMetzger (approved):
LGTM

cli/src/commands/workspace/update_stale.rs
yuja: Hmm, I'm okay with either way, but if we don't like the default string wrapping, maybe we should disable it instead of abusing raw string?

ilyagr: I mildly prefer keeping this as-is or `indoc`-ed raw string to a raw string which breaks indentation.

I also agree with Yuya's point (if I understand the situation correctly): if we want to reconfigure `cargo fmt`, we should do it separately.

PhilipMetzger: > Hmm, I'm okay with either way, but if we don't like the default string wrapping, maybe we should disable it instead of abusing raw string?

I don't really care about string wrapping and just want to prevent some human errors with raw strings. You can leave it as is if you prefer that.

yuja: Ok, I'll leave it unchanged as this PR just moves codes around. 

FWIW, I don't think raw string is strictly better because it disables backslash escapes.

Issue #4062, opened 2024-07-10
FR: add a diff mode without materialized conflicts


martinvonz:
**Is your feature request related to a problem? Please describe.**

We currently display diffs by materializing conflicts as conflict markers. The resulting diffs can be hard to read.

**Describe the solution you'd like**

An alternative way of displaying diffs involving conflicts is to display multiple diffs. For example, a newly introduced conflict `A+(C-B)` on top of a base content of `D` can be displayed as diffs `A-D` and `C-B`. This can be especially useful since the former is often empty (i.e. `A=D`).

(This was planned from the beginning but the materialized conflicts are also useful.)

arxanas:
Does this help external diff/merge tools render the conflicts too? Or not really, since you still have multiple diffs that correspond to a single file? (I guess the advantage could be if you decided to synthetically create `foo.A-D`, `foo.C-B`, etc., for rendering purposes... but not sure how useful that would be in practice, either.)

ilyagr:
I was thinking on and off about this. I am not sure how often this would be a clearer presentation and whether it would sometimes be more confusing, but I certainly want to experiment with it. Additionally, the logic would also be useful for having blame and absorb work over conflicts (so that they know which side of a conflict comes from the parent, for example).

yuja:
The underlying machinery will also be useful to implement grepping https://github.com/martinvonz/jj/issues/2933
(Right now, I'm using the common materialized conflict diffs, but it makes more sense to diff each conflict pair.)

Issue #2988, opened 2024-02-07
FR: `jj blame`/`jj point-finger` a blame layer.


PhilipMetzger:
**Is your feature request related to a problem? Please describe.**
Jujutsu currently lacks a blame layer which determines who was responsible for which line of code. This is commonly called a blame layer.

**Describe the solution you'd like**
Add a `jj blame` command and a trait as integration point for large repos, as it may be very expensive to calculate each line for a file which was changed many times. 

**Additional context**

Maybe choosing a approach like [Sapling](../../../facebook/sapling) could be useful, as it has support to blame different types of things like commit metadata, the whole file or a incomplete file.

ilyagr:
`blame` might be related to absorb (#170, #64).

This thought might be pure distraction, but it'dbe nice to not focus on *lines* here, so that we could later use difftastic-like diff to attribute syntactic hunks.

PhilipMetzger:
> This thought might be pure distraction, but it'dbe nice to not focus on _lines_ here, so that we could later use difftastic-like diff to attribute syntactic hunks.

No, that's actually a great point. If we want to support difftastic-like diffs, having them available in `blame` would be very cool. This is actually very related to trait I proposed, as Code Review tools may want to use this functionality too.

joyously:
_FYI: Pijul chose to call it `credit` instead. There is also `annotate`._

ilyagr:
> _FYI: Pijul chose to call it `credit` instead. There is also `annotate`._

Reminds me of this song: https://youtu.be/UQHaGhC7C2E?si=n20KACC6SZmK480N&t=235

It's not the best Tom Lehrer song, but it might be worth a snicker depending on the mood.

PhilipMetzger:
I'm totally onboard with calling it `jj annotate` or `jj credit` as long as the `blame` alias exists, so everyone will be happy.

thoughtpolice:
> Ideally, `jj blame` should only show the lines with bugs.

It would be funny if we called it `jj credit`, and then make `jj blame` different and claim it "Only shows lines of code that has bugs" and then make it only shows your commits (aka `mine()`). I'm not saying we should do this, just that it would be pretty funny.

steveklabnik:
I have often said "`git blame` is an alias for `whoami`"...

epage:
So not familiar with `git absorb` enough to know how it is related or how the difftastic stuff would work but thought I'd chime in with some "blame" UI ideas I've been playing with.

[...]

PR #4176, opened 2024-07-29
Add command jj annotate/blame

allonsy:
This change adds a new annotate command (in git this is blame). 
In order to annotate a file, call: `jj annotate <FILE_PATH>`. 
In response, jj will traverse the commit graph, trying to find the source of the change for each line.
The output is something similar to this:
```
>>> jj annotate file.txt
tqqpzkzp bd94545f Alec Snyder 2024-07-29 18:39:13 1: this is an initial line
nqvnwumz b3a07c79 Alec Snyder 2024-07-29 18:39:35 2: this line is commit 1
wlpkktol 5f7de224 Alec Snyder 2024-07-29 18:40:01 3: this line is commit 2
```

I've also added a new module to jj_lib which exposes a function `get_annotation_for_file` which 
returns a line by line evaluation of the the given file in a more general interface.

A few caveats
* directories aren't supported
* conflicted files aren't supported as I'm not exactly sure what to print out in these cases.

# Checklist

If applicable:


PhilipMetzger (commented):
Just a initial comment, it should also mention that this starts work on #2988.

CHANGELOG.md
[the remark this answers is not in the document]
allonsy: oof, yay auto formatting in my editor!

cli/tests/cli-reference@.md.snap
```diff
@@ -111,6 +112,7 @@ To get started, see the tutorial at https://github.com/martinvonz/jj/blob/main/d
 ###### **Subcommands:**
 
 * `abandon` — Abandon a revision
+* `annotate` — Annotate a file
```
joyously, line 115: Can you describe it without using the command name?

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
+// Copyright 2020 The Jujutsu Authors
```
joyously, line 1: 2020?

allonsy: I copied from other files, wasn't sure if the year also gets updated

cli/tests/cli-reference@.md.snap
allonsy: I followed the structure of abandon but I can modify it

lib/src/annotate.rs
PhilipMetzger: This is a new file, so it should 2024

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+    /// file is required
+    #[error("File {0} must be a regular file, not a directory")]
+    UnsupportedFileType(String),
+    /// the file is in a conflicted state and can therefore not be annotated
+    /// properly
+    #[error("File {0} is conflicted at commit: {0}")]
+    Conflicted(String, String),
+    /// pass-through of uncaught backend errors
+    #[error("{0}")]
+    BackendError(BackendError),
```
PhilipMetzger, line 48: Nit: 
```suggestion
#[error(transparent)]
BackendError(#[from] BackendError)
```

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+    /// For each value in the array, the commit_id is the commit id of the
+    /// originator of the line and the string is the actual line itself (without
+    /// newline terminators). The vector is ordered by appearance in the
+    /// file
+    pub file_annotations: Vec<(CommitId, String)>,
+}
+
+struct PartialResults {
+    original_line_map: HashMap<usize, CommitId>,
+    local_line_map: HashMap<CommitId, HashMap<usize, usize>>, /* CommitId -> (local_line_index
```
PhilipMetzger, line 69: Nit: missing documentation.

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    Ok(partial_results.convert_to_results(&original_contents))
+}
+
+fn process_commits(
+    repo: &ReadonlyRepo,
+    starting_commit_id: &CommitId,
+    file_name: &RepoPath,
+    results: &mut PartialResults,
+) -> Result<(), AnnotateError> {
```
PhilipMetzger, line 187: Nit: Why is this not `fn process_commits(...) -> Result<PartialResults, AnnotateError)`?

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+    let mut current_commit = Some(repo.store().get_commit(starting_commit_id)?);
+    while let Some(next_commit) = current_commit {
```
PhilipMetzger, line 189: Nit: Remove redundant `Some(...)` wrapping.

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+    Ok(())
+}
+
+fn process_commit(
+    commit: &Commit,
+    repo: &ReadonlyRepo,
+    file_name: &RepoPath,
+    results: &mut PartialResults,
+) -> Result<(), AnnotateError> {
```
PhilipMetzger, line 209: Ditto

cli/tests/test_annotate_command.rs
```diff
@@ -0,0 +1,80 @@
+// Copyright 2022 The Jujutsu Authors
```
PhilipMetzger, line 1: Ditto with the copyright.

PhilipMetzger (changes requested):
Here's a second review.

ilyagr:
> conflicted files aren't supported as I'm not exactly sure what to print out in these cases.

I think that it should figure out which side of the conflict comes from which parent of the commit. Any sides that don't come from a parent (e.g. after a rebase) can be attributed as being added to the commit itself.

Infrastructure related to https://github.com/martinvonz/jj/issues/4062 would be helpful here.

Of course, none of this is necessary for a first version of annotate functionality.

cli/tests/cli-reference@.md.snap
ilyagr: For name ideas, `credit` is somewhat popular in #2988. I also like `attribute` a bit more than `annotate` if we want a neutral tone. Perhaps I'll mention that there.

If people are not super-excited about these, `annotate` is a fine default choice for a name. I hope it will change later, but it doesn't have to.

lib/src/annotate.rs
ilyagr: Taking notes: `original_line_map` seems to map the line number of a line in the file in the original commit to its annotation. That is what's used to make "AnnotateResults".

I'm guessing that `local_line_map` only tracks those lines in the given commit that correspond to some line in the original commit.

------

IIUC, the CommitId in `original_line_map` is just an annotation, whereas the CommitId in `local_line_map` tracks the progress of the algorithm, and for *those* CommitId-s it's important that you can extract the parent commit(s) from them. It might help if you do `type Annotation = CommitId` and replace the CommitId-s that are just annotations with them. For one, annotations will hopefully include a conflict side id at some point.

Similarly, `type LineNumber=usize` might help, though AFAICT pretty much all the usize-s are line numbers here.

-------

In general, a few tests that illustrate how these objects behave and/or a few comments describing the algorithm would help.

lib/src/annotate.rs
ilyagr: You might be able to use `break;` instead of `current_commit = None`.

lib/src/annotate.rs
ilyagr, line 1: At a glance, the underlying design seems reasonable to me so far, but the code seems like it will be a bit difficult to extend. (I'm mostly interested in extending it to support conflicts and some sort of an `absorb` command, if we decide to do that using blame). I'm wondering what we can do to make it more extensible/maintainable/modular (a good sign of it becoming more modular would be if we could split this into several commits in a reasonable way; this seems hard to do as-is), but I don't immediately have a clear suggestion overall.

I'm not sure how much of this burden has to lie on you and this PR, and how much could be cleaned up over time later; I wonder what others think.

In any case, thank you for moving the needle on this feature!

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+pub fn get_annotation_for_file(
+    file_name: &str,
+    repo: &ReadonlyRepo,
+    starting_commit_id: &CommitId,
+) -> Result<AnnotateResults, AnnotateError> {
+    let file_path = RepoPathBuf::from_relative_path(file_name);
+    if file_path.is_err() {
+        return Err(AnnotateError::FileNotFound(file_name.to_string()));
+    }
+    let file_path: RepoPathBuf = file_path.unwrap();
```
yuja, line 163: nit: for type safety, `file_name` can be converted to `RepoPath` by caller.

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    let current_commit = store.get_commit(starting_commit_id)?;
+    let current_tree = current_commit.tree()?;
+    let original_file_id = get_file_id(current_commit.id(), &current_tree, &file_path)?;
+    if original_file_id.is_none() {
+        return Err(AnnotateError::FileNotFound(file_name.to_string()));
+    }
+    let original_contents =
+        get_file_contents(repo.store(), &file_path, &original_file_id.unwrap())?;
+    let num_lines = original_contents.lines().count();
```
yuja, line 174: nit: use `split_inclusive(|b| *b == b'\n')` for consistency

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+        }
+        let remaining_lines = remaining_lines.unwrap();
+        for (_, original_line_number) in remaining_lines {
+            self.original_line_map
+                .insert(original_line_number, commit_id.clone());
+        }
+    }
+
+    fn get_next_commit(&self) -> Option<CommitId> {
+        return self.local_line_map.keys().next().cloned();
```
yuja, line 137: Instead of picking one randomly, I think we should visit parents in topological order, and terminate if there are no more lines to look for.

Mercurial appears to do DFS.

lib/src/annotate.rs
```diff
@@ -0,0 +1,335 @@
...
+    current_file_id: &FileId,
+    current_commit_id: &CommitId,
+    parent_file_id: &FileId,
+    parent_commit_id: &CommitId,
+) -> BackendResult<()> {
+    if current_file_id == parent_file_id {
+        results.swap_full_commit_id(current_commit_id, parent_commit_id);
+        return Ok(());
+    }
+    let current_contents = get_file_contents(store, file_name, current_file_id)?;
```
yuja, line 248: nit: might be better to cache original content in `local_line_map` instead of fetching again.

cli/tests/cli-reference@.md.snap
allonsy: We've got a few options
* annotate: I generally like it
* attribute: also fine with this
* credit: not my favorite, it doesn't really describe to me what it does
* blame: just silly (leftover from git). No reason to be so negative in the command.

lib/src/annotate.rs
allonsy: I can, but is there a specific reason for why you think it's better?

lib/src/annotate.rs
allonsy: For process_commits I can probably return the PartialResults. However, from process_commit, I'd rather pass in a mutable reference that mutates since I call it within a loop. To return it every time means that each call to process_commit needs to consume the old results and return new results which, if not boxed, just wastes unneeded copying on the stack. Therefore, I think process_commit is more ergonomic to pass a mutable reference.

lib/src/annotate.rs
allonsy: agreed, much cleaner with break. The some wrapping was leftover from an older code iteration that got changed

lib/src/annotate.rs
PhilipMetzger: Isn't linear control flow easier to read instead of having output parameters?

lib/src/annotate.rs
ilyagr: It's not that bad, but you could decrease the time it takes somebody to navigate the code, yes.

lib/src/annotate.rs
ilyagr: IMO, it's not too bad to have `& mut` parameters if the function receives a partially-processed object and processes it further, but then  it should be the first argument since modifying this object is the most important thing the function does.

Even in this case, I would often reach for a function that would take a `PartialResults` (not a reference) and return a `PartialResults`. But (perhaps unlike Philip), I don't see this as much better, I don't have a clear rule about how to choose.

Separately, re-designing the logic to decrease the need for such functions is a good thing, if possible. I'm not immediately sure if that could be done here.

A big no-no for me (which I think is not the case here) is if a function takes an empty object as &mut and uses &mut purely as an output.

lib/src/annotate.rs
PhilipMetzger: fair enough

martinvonz:
Thanks for working on this! I have not looked at the implementation yet, but I'll just quickly note that I think a TUI for the functionality would be really useful. I quite often end up trying to find where the previous version of a line came from, and quite often I end up repeating that process several times. GitHub has a "Blame prior to change abc123..." button for each line (or span of lines) for that. It's probably not a big deal to redo the calculation when the user asks for that option, but ideally we could reuse some of the information we already calculated.

allonsy:
> Thanks for working on this! I have not looked at the implementation yet, but I'll just quickly note that I think a TUI for the functionality would be really useful. I quite often end up trying to find where the previous version of a line came from, and quite often I end up repeating that process several times. GitHub has a "Blame prior to change abc123..." button for each line (or span of lines) for that. It's probably not a big deal to redo the calculation when the user asks for that option, but ideally we could reuse some of the information we already calculated.

That's a nice idea. I'm hoping to merge this in and maybe in a future PR I can implement a TUI? Unless you think implementing a TUI would radically change things enough to not make it worth it to merge this in now?

lib/src/annotate.rs
allonsy: do you mean cache the contents of each commit still in local_line_map?
Or do you actually mean the original content (content of the working copy)?

martinvonz:
> I'm hoping to merge this in and maybe in a future PR I can implement a TUI?

Sounds good. (And I didn't mean to ask you to create a TUI, but sounds great if you're interested in doing that.)

> Unless you think implementing a TUI would radically change things enough to not make it worth it to merge this in now?

Nope. I just mentioned it in case it would help any future work in this area.

allonsy:
> > conflicted files aren't supported as I'm not exactly sure what to print out in these cases.
> 
> I think that it should figure out which side of the conflict comes from which parent of the commit. Any sides that don't come from a parent (e.g. after a rebase) can be attributed as being added to the commit itself.
> 
> Infrastructure related to #4062 would be helpful here.
> 
> Of course, none of this is necessary for a first version of annotate functionality.

I agree with this for conflicts, I initially tried this but gave up for the first iteration as currently the conflict writers (lib/src/conflict.rs) doesn't have the commit information. So I'd have to forward that through the callers. Definitely doable but a little out of scope for this PR I think

lib/src/annotate.rs
allonsy: should be resolved now

lib/src/annotate.rs
allonsy: fixed

lib/src/annotate.rs
allonsy: done

lib/src/annotate.rs
allonsy: I've added documentation to the functions. Does it seem more manageable now?

cli/tests/cli-reference@.md.snap
joyously: I just thought of `map`, and then rereading `attribute` I thought of `byline`.

lib/src/annotate.rs
yuja: I don't think DFS vs BFS matters here. The thing we'll need to guarantee is topological ordering (all descendant lines should have been propagated to the current commit iiuic), and this property is guaranteed if we use revset API. It will also help terminate early.

BTW, `jj annotate lib/src/repo.rs` panics with the current implementation. It might be because of merge commits.

> In terms of graph pruning, I tried to make a few optimizations. Namely, if the parent and child share the same file id, I immediately move on to the parent of the parent

The default index backend wouldn't have any data to optimize it further. (We might want to add bloom filter or something to optimize `log PATH` query.) I assume the Google one will have optimized path for file-oriented query.

lib/src/annotate.rs
yuja: Cache the content of each commit. I assume it will be consumed soon (as long as the graph is linear.) Cache of the original content might also be good.

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,84 @@
...
+    args: &AnnotateArgs,
+) -> Result<(), CommandError> {
+    let workspace_command = command.workspace_helper(ui)?;
+    let starting_commit_id = workspace_command.get_wc_commit_id().unwrap();
+    let repo = workspace_command.repo();
+    let file_path = RepoPathBuf::from_relative_path(&args.path);
+    if file_path.is_err() {
+        eprintln!("Unable to locate file: {}", args.path);
+        return Ok(());
+    }
```
yuja, line 52: nit: you can use `.map_err(user_error)?`, `let Some(..) = .. else { return; };`, etc. instead of `.is_err()` and `.unwrap()`.
This applies globally.

lib/src/annotate.rs
allonsy: Nice catch on the bug. Just pushed a fix. 
It was when you have a file with a merge in its history where one side (of the merge) was ignored rather than integrated all changes. 
I've also added a test case to catch this going forward

yuja: Thanks. Now I understand why random choice works here. Since `forward_to_new_commit()` removes matching lines from the old map, the same matching lines will never propagate to the other parents. So the first parent is always prioritized (= the order is stabilized.)

Still I think topological order is better for optimization. With the current implementation, ancestors of fork point can be visited more than once, where contents are fetched and diffed again.

martinvonz: Could we use `RevsetExpression::ancestors()` filtered by path and then use `iter_graph()` to see how the commits are connected? Then it might also work at Google (where iterating over all ancestors is too slow).

Good point about copy tracing. I'll have to think more about that.

yuja: > Could we use `RevsetExpression::ancestors()` filtered by path and then use `iter_graph()` to see how the commits are connected?

I rejected the idea in my mind, but maybe I was wrong. If both graph walk and line propagation are based on `Vec<GraphEdge>` (instead of `commit.parents()`), it will probably work (and more efficient.)

allonsy: Generally I hesitated about using ancestors() since it most cases, we don't have to check every ancestor so we waste time computing the graph.
In the current implementation, it only goes as deep so as it needs and once it finds changes for every single line, it stops iterating. 
I think where I'm unsatisfied with my implementation is, as you mentioned, some elements of non-determinism due to the `get_next_commit` function. In this case, if two parent commits commit the exact same change to a file, it is not deterministic which commit is taken as the source. This isn't necessarily incorrect, as in that case it is indeed ambiguous which commit is the source commit. However, it would be nice to be able to say which branch is always taken. 
This non-determinism is solved by using variants to ancestors() and whatnot, if, perhaps, a little slower. 

Overall, here is what I'm going to do: based on the comments here, I'm going to try out a few implementations mentioned here and benchmark them across a few different files and cases: Some small files, some large files, some that only need to go shallow in the graph, and some that need to go deep into the commit graph and I'll see how each implementation performs.

I'll try the current the implementation, one with RevsetAncestors, and with RevsetAncestors() filtered to the specific files (of course with iter_graph since it should be lazy).
I'll report back with the results

joyously: > In this case, if two parent commits commit the exact same change to a file, it is not deterministic which commit is taken as the source. 

Can you use the commit time?

yuja: > Generally I hesitated about using ancestors() since it most cases, we don't have to check every ancestor so we waste time computing the graph.

Oh, you mean we can omit the second parent if all lines are propagated to the first parent? That's indeed correct. However, I think visiting ancestors (by using the index) is relatively cheap compared to diffing file contents multiple times. And it might avoid abstracting the whole annotation machinery by backends. As Martin said, the current implementation is unusable at Google scale.

`file(PATH)` query is slow with the current default index implementation, but I hope it will be improved later.

> This isn't necessarily incorrect, as in that case it is indeed ambiguous which commit is the source commit. However, it would be nice to be able to say which branch is always taken.

It's also important for test stability.

allonsy: Ok, So I've uploaded a new commit with a revset implementation that uses ancestors + a file filter. It's overall a little faster (maybe 10% faster) when we have to go deeper into the graph. For shallow graphs its a little more expensive but probably not noticeable and not really the cases we are trying to optimize for. 
This new approach is stable in terms of traversing the graph and may be more readable.

@martinvonz how do you think this implementation would perform on large google-like repos? Is there something I'm missing in terms of revsetExpression optimizations?

Overall, what do we think of this approach? I made the changes in a separate commit so we can just choose which one we like better

lib/src/annotate.rs
allonsy: Added a file cache

cli/tests/cli-reference@.md.snap
allonsy: map isn't bad actually. I'm starting to like it more and more. 
Especially if we eventually add a TUI to this, it will feel like navigating a file map line by line.

Anyone else want to weigh in? I don't really feel strongly here about the name, but we should decide something before merging

lib/src/annotate.rs
martinvonz: > @martinvonz how do you think this implementation would perform on large google-like repos?

Looks good, thanks!

cli/tests/cli-reference@.md.snap
martinvonz: I vote for `annotate` because it's familiar.

lib/src/annotate.rs
```diff
@@ -0,0 +1,421 @@
...
+    let mut reader = store.read_file(path, file_id)?;
+    let mut contents: Vec<u8> = Vec::new();
+    let err = reader.read_to_end(&mut contents);
+    if let Err(e) = err {
+        return Err(BackendError::ReadFile {
+            path: path.to_owned(),
+            id: file_id.to_owned(),
+            source: Box::new(e),
+        });
+    }
```
martinvonz, line 419: I think `map_err()` seems more idiomatic

cli/tests/cli-reference@.md.snap
spectral54: Also prefer `annotate`. It's available in:
- `svn` (also has `blame` and `praise` synonyms),
- `p4` (also has `blame` synonym, I think),
- `hg` (also has `blame` synonym)
- `git` (which uses a different output format from `git blame`, but fundamentally has the same behavior, I believe)

Probably others.

lib/src/annotate.rs
```diff
@@ -0,0 +1,421 @@
...
+        ));
+    }
+
+    let file_object = file_value.first().as_ref().unwrap();
+    match file_object {
+        TreeValue::File { id, .. } => Ok(Some(id.clone())),
+        _ => Err(AnnotateError::UnsupportedFileType(
+            file_name.to_internal_dir_string(),
+        )),
+    }
```
martinvonz, line 406: I think this can be simplified with `file_value.into_resolved().map_err()`

lib/src/annotate.rs
```diff
@@ -0,0 +1,421 @@
...
+            }
+        }
+    }
+    results.drain_remaining_for_commit_id(current_commit_id);
+
+    Ok(())
+}
+
+/// For two versions of the same file, for all the lines in common, overwrite
+/// the new mapping in the results for the new commit. Meaning, Let's say I have
```
martinvonz, line 308: nit: Drop "Meaning, "?

martinvonz (commented):
I'm not done reviewing yet, but here are my comments so far

lib/src/annotate.rs
```diff
@@ -0,0 +1,421 @@
...
+        .evaluate_programmatic(repo)
+        .unwrap();
+    let mut is_first = true;
+
+    for (cid, edge_list) in revset.iter_graph() {
+        if is_first {
+            results.swap_full_commit_id(starting_commit_id, &cid);
+            is_first = false;
+        }
+        process_commit(results, repo, file_name, &cid, &edge_list)?;
```
yuja, line 259: Perhaps, the starting commit has to be included in the graph (i.e. `ancestors(start) & (file(path) | start)`?

The starting-commit lines can be forwarded to multiple parents if there are a merge commit in between. `file(path)` won't include a merge commit if the commit is empty relative to auto-merge parents.

lib/src/annotate.rs
```diff
@@ -0,0 +1,421 @@
...
+    starting_commit_id: &CommitId,
+    file_name: &RepoPath,
+    num_lines: usize,
+) -> Result<(), AnnotateError> {
+    let predicate = RevsetFilterPredicate::File(FilesetExpression::file_path(file_name.to_owned()));
+    let revset = RevsetExpression::commit(starting_commit_id.clone())
+        .ancestors()
+        .filtered(predicate)
+        .evaluate_programmatic(repo)
+        .unwrap();
```
yuja, line 251: nit: propagate `BackendError`?

lib/src/annotate.rs
```diff
@@ -0,0 +1,421 @@
...
+    file_name: &RepoPath,
+    current_file_id: &FileId,
+    current_commit_id: &CommitId,
+    parent_file_id: &FileId,
+    parent_commit_id: &CommitId,
+) -> BackendResult<()> {
+    if current_file_id == parent_file_id {
+        results.swap_full_commit_id(current_commit_id, parent_commit_id);
+        return Ok(());
+    }
```
yuja, line 328: nit: might be okay to drop this optimization. It would be only applicable to the starting commit.

lib/src/annotate.rs
PhilipMetzger, line 1: nit: Squash into parent since this is the favored approach.

lib/src/annotate.rs
```diff
@@ -0,0 +1,421 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Methods that allow annotation (attribution and blame) for a file in a
+//! repository
+
```
PhilipMetzger, line 17: Nit: Add a `// TODO: Add support for different blame layers with a trait in the future. Like commit metadata and more. `

PhilipMetzger (approved):
LG to me after this round, but you should wait for a final approval from either Martin or Yuya.

lib/src/annotate.rs
PhilipMetzger: The documentation is great, although it is written in a different tense/style than all other documentation. Since this is a minor thing, it could be improved in a follow-up.

CHANGELOG.md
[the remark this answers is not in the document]
allonsy: Yeah, good reminder

lib/src/annotate.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// the requested file path was not found
+    #[error("Unable to locate file: {0}")]
+    FileNotFound(String),
+    /// the file type is incorrect. Usually a directory was given but a regular
+    /// file is required
+    #[error("File {0} must be a regular file, not a directory")]
+    UnsupportedFileType(String),
+    /// the file is in a conflicted state and can therefore not be annotated
+    /// properly
+    #[error("File {0} is conflicted at commit: {0}")]
```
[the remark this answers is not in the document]
allonsy: nice catch

lib/src/annotate.rs
```diff
@@ -0,0 +1,406 @@
...
+        old_commit_id: &CommitId,
+        old_local_line_number: usize,
+        new_commit_id: &CommitId,
+        new_local_line_number: usize,
+    ) {
+        let old_map = self.local_line_map.get_mut(old_commit_id);
+        if old_map.is_none() {
+            return;
+        }
+        let old_map = old_map.unwrap();
```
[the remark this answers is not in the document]
allonsy: That means that I have to wrap the whole rest of the function inside the if-let. Is that preferred?

lib/src/annotate.rs
```diff
@@ -0,0 +1,406 @@
...
+        self.file_cache.insert(commit_id.clone(), file_contents);
+
+        Ok(())
+    }
+
+    /// retrieves a file from the cache for a specific commit. If the file isn't
+    /// found, it panics. Make sure to call load_file_into_cache first.
+    fn get_file_from_cache(&self, commit_id: &CommitId) -> &[u8] {
+        self.file_cache.get(commit_id).unwrap()
+    }
```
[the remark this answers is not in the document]
allonsy: Yeah, I guess the function doesn't really offer anything here except unwrap

cli/src/commands/mod.rs
```diff
@@ -69,6 +70,7 @@ use crate::ui::Ui;
 #[derive(clap::Parser, Clone, Debug)]
 enum Command {
     Abandon(abandon::AbandonArgs),
+    Annotate(annotate::AnnotateArgs),
```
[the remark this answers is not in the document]
martinvonz: I don't personally mind the name `blame`, but I know that some people do, and that's the main reason we made `annotate` the canonical name. If we include a `blame` alias by default, I think it's likely that most people will use that instead. So, if we're trying to nudge people to use `annotate` instead, it seems counterproductive to include the alias. But I don't care strongly either way.

lib/src/annotate.rs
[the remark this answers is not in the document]
PhilipMetzger: I think its easier to read, but others should chime in.

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let res = get_annotation_for_file(&file_path.unwrap(), repo, starting_commit_id);
```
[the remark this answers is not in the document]
allonsy: done

lib/src/annotate.rs
allonsy: Good idea, implemented

lib/src/annotate.rs
allonsy: agreed

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[instrument(skip_all)]
+pub(crate) fn cmd_annotate(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &AnnotateArgs,
+) -> Result<(), CommandError> {
+    let workspace_command = command.workspace_helper(ui)?;
+    let starting_commit_id = workspace_command.get_wc_commit_id().unwrap();
```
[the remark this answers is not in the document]
allonsy: added a new revision arg. Stole the code from diff.rs

cli/src/commands/mod.rs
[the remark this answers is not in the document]
PhilipMetzger: Ok, then we should update the suggested aliases to in the docs to mention it but not add a core alias here, if that works for everyone.

lib/src/annotate.rs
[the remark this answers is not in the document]
martinvonz: I prefer it mostly because it avoids the `.unwrap()` calls

cli/tests/cli-reference@.md.snap
allonsy: I'm happy with annotate

lib/src/annotate.rs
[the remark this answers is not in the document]
yuja: I would use `let Some(old_map) = .. else { return; };` here.

cli/src/commands/mod.rs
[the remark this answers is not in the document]
yuja: Regarding the canonical name, `annotate` might be a subcommand of `file`.
https://github.com/martinvonz/jj/issues/2988#issuecomment-2257574873

(I would probably add `jj annotate` alias just like `jj cat`, though.)

CHANGELOG.md
yuja, line 1: nit: revert unintended formatting changes?

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,94 @@
...
+    let file_value = starting_commit.tree()?.path_value(&file_path)?;
+    if file_value.is_absent() {
+        let ui_path = workspace_command.format_file_path(&file_path);
+        return Err(user_error(format!("No such path: {ui_path}")));
+    }
+
+    let annotations = get_annotation_for_file(&file_path, repo, &starting_commit);
+    if let Err(e) = annotations {
+        eprintln!("{}", e);
+        return Ok(());
```
yuja, line 61: nit: `.map_err(..)?`

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,94 @@
...
+    repo: &ReadonlyRepo,
+    ui: &mut Ui,
+    settings: &UserSettings,
+    workspace_command: &WorkspaceCommandHelper,
+    results: &AnnotateResults,
+) -> Result<(), CommandError> {
+    let annotate_commit_summary_text = settings
+        .config()
+        .get_string("templates.annotate_commit_summary")?;
+    let template = workspace_command.parse_commit_template(&annotate_commit_summary_text)?;
```
yuja, line 84: nit: perhaps, template parsing can be moved to caller so the error will be reported earlier.

It will also help when we add `--template` argument.

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,94 @@
...
+    ui.request_pager();
+    let mut formatter = ui.stdout_formatter();
+    for (line_no, (commit_id, line)) in results.file_annotations.iter().enumerate() {
+        let commit = repo.store().get_commit(commit_id)?;
+        template.format(&commit, formatter.as_mut())?;
+        writeln!(formatter, " {}: {}", line_no + 1, line)?;
```
yuja, line 90: nit: `"{:>4}", line_no + 1` to align the output?

We can also label it a "line_number" as we do in diff_util.rs.

cli/src/commands/mod.rs
```diff
@@ -174,6 +176,7 @@ pub fn run_command(ui: &mut Ui, command_helper: &CommandHelper) -> Result<(), Co
         #[cfg(feature = "bench")]
         Command::Bench(args) => bench::cmd_bench(ui, command_helper, args),
         Command::Branch(args) => branch::cmd_branch(ui, command_helper, args),
+        Command::Annotate(args) => annotate::cmd_annotate(ui, command_helper, args),
```
yuja, line 179: nit: move next to `Abandon`

cli/src/config/templates.toml
```diff
@@ -14,6 +14,12 @@ if(remote,
 
 commit_summary = 'format_commit_summary_with_refs(self, branches)'
 
+annotate_commit_summary = '''
+self.change_id().shortest(8) ++ " " ++ 
+self.commit_id().shortest(8) ++ " " ++ 
+self.author().name() ++ " " ++ format_timestamp(self.author().timestamp())
+'''
```
yuja, line 21: nits: 
- use `format_short_id()` and `format_short_signature()`
- use `committer.timestamp()` for consistency with log templates
- signature field will need some padding, but we don't have padding function right now https://github.com/martinvonz/jj/issues/3183

```
separate(" ",
  format_short_id(change_id),
  format_short_id(commit_id),
  format_short_signature(author),
  format_timestamp(committer.timestamp()),
)
```

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+// limitations under the License.
+
+//! Methods that allow annotation (attribution and blame) for a file in a
+//! repository.
+//!
+//! TODO: Add support for different blame layers with a trait in the future.
+//! Like commit metadata and more.
+
+use std::collections::HashMap;
+use std::io::BufRead;
```
yuja, line 22: nit: `BufRead` can be removed if all `.lines()` are replaced with `.split_inclusive(..)`

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+}
+
+/// Annotation results for a specific file
+pub struct AnnotateResults {
+    /// An array of annotation results ordered by line.
+    /// For each value in the array, the commit_id is the commit id of the
+    /// originator of the line and the string is the actual line itself (without
+    /// newline terminators). The vector is ordered by appearance in the
+    /// file
+    pub file_annotations: Vec<(CommitId, String)>,
```
yuja, line 64: Line content should be bytes `Vec<u8>` instead of `String`.

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+
+        Ok(())
+    }
+}
+
+/// Get line by line annotations for a specific file path in the repo.
+pub fn get_annotation_for_file(
+    file_path: &RepoPath,
+    repo: &ReadonlyRepo,
+    starting_commit: &Commit,
```
yuja, line 194: nit: `repo` can be `&dyn Repo`. Annotation should work even if the repo is a MutableRepo.

And I would reorder arguments as `repo, starting_commit, file_path` (global context first.)

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+) -> Result<AnnotateResults, AnnotateError> {
+    let current_tree = starting_commit.tree()?;
+    let original_file_id = get_file_id(starting_commit.id(), &current_tree, file_path)?;
+    if original_file_id.is_none() {
+        return Err(AnnotateError::FileNotFound(
+            file_path.as_internal_file_string().to_string(),
+        ));
```
yuja, line 201: nit: maybe we can remove `FileNotFound` error and return empty result instead?

This `FileNotFound` error is correct, but the other user looks wrong. And, if annotation reached to non-file ancestor entry, it probably means the file is absent and should stop there.

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+        .union(
+            &RevsetExpression::commit(starting_commit_id.clone())
+                .ancestors()
+                .filtered(predicate),
+        )
+        .evaluate_programmatic(repo)
+        .map_err(|e| match e {
+            RevsetEvaluationError::StoreError(backend_error) => AnnotateError::from(backend_error),
+            RevsetEvaluationError::Other(_) => {
+                AnnotateError::FileNotFound(file_name.to_internal_dir_string())
```
yuja, line 240: nit: just `panic!()`? I don't think `Other(_)` means `FileNotFound`.

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+                    file_name,
+                    &current_file_id,
+                    current_commit_id,
+                    &parent_file_id,
+                    parent_commit.id(),
+                )?;
+            }
+        }
+    }
+    results.drain_remaining_for_commit_id(current_commit_id);
```
yuja, line 288: Perhaps, this, `process_file_ids()`, `forward_to_new_commit()`, and `drain_remaining_for_commit_id()` can be reorganized in a way that the current line_map entry is popped first.

```rust
let Some(current_map) = self.local_line_map.remove(current_commit_id);
for parent_edge in edges {
    ...
}
results.drain(current_map);
```

It will probably solve the issue of mutable borrows, and clarifies that the `current_map` is consumed.

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+    results: &mut PartialResults,
+    store: &Store,
+    file_name: &RepoPath,
+    current_file_id: &FileId,
+    current_commit_id: &CommitId,
+    parent_file_id: &FileId,
+    parent_commit_id: &CommitId,
+) -> BackendResult<()> {
+    results.load_file_into_cache(store, current_commit_id, file_name, current_file_id)?;
+    results.load_file_into_cache(store, parent_commit_id, file_name, parent_file_id)?;
```
yuja, line 312: nit: it might make sense to refactor (or rename) `file_id`-based API. We no longer have `file_id` fast path.

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+fn get_same_line_map(current_contents: &[u8], parent_contents: &[u8]) -> HashMap<usize, usize> {
+    let mut result_map = HashMap::new();
+    let inputs = vec![current_contents, parent_contents];
+    let diff = Diff::by_line(&inputs);
+    let mut current_line_counter: usize = 0;
+    let mut parent_line_counter: usize = 0;
+    for hunk in diff.hunks() {
+        match hunk {
+            DiffHunk::Matching(common_string) => {
+                for _ in common_string.lines() {
```
yuja, line 339: nit: `.split_inclusive(|b| *b == b'\n')` for consistency

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+                    for _ in current_output.lines() {
+                        current_line_counter += 1;
+                    }
+                }
+                if !parent_output.is_empty() {
+                    for _ in parent_output.lines() {
+                        parent_line_counter += 1;
+                    }
+                }
+            }
```
yuja, line 358: nit: 
```rust
                current_line_counter += current_output.split_inclusive(|b| *b == b'\n').count();
                parent_line_counter += parent_output.split_inclusive(|b| *b == b'\n').count();
```

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+        .map_err(|_| {
+            AnnotateError::Conflicted(file_name.to_internal_dir_string(), commit_id.hex())
+        })?
+        .unwrap();
+
+    match file_value {
+        TreeValue::File { id, .. } => Ok(Some(id.clone())),
+        _ => Err(AnnotateError::UnsupportedFileType(
+            file_name.to_internal_dir_string(),
+        )),
```
yuja, line 382: Perhaps, non-file types should be considered absent? If a directory is replaced with file, it means the file is created at that revision.

lib/src/annotate.rs
```diff
@@ -0,0 +1,397 @@
...
+    tree: &MergedTree,
+    file_name: &RepoPath,
+) -> Result<Option<FileId>, AnnotateError> {
+    let file_value = tree
+        .path_value(file_name)?
+        .into_resolved()
+        .map_err(|_| {
+            AnnotateError::Conflicted(file_name.to_internal_dir_string(), commit_id.hex())
+        })?
+        .unwrap();
```
yuja, line 376: Maybe better to use materialized conflict instead of erroring out?
See `conflicts::materialize_tree_value()` and its callers.

allonsy: I took a look at this API when implementing. I think this is interesting but not exactly what we want because it includes file dressing for diffs that return false positives. Meaning there is '-' and '+' characters that interfere with the diffing algorithm. 
I think in a later fix when we support conflicts we can annotate the conflicted lines in a smarter way

yuja: Yeah, I agree about that. I just mean materialized conflict will be okay as a temporary solution, and should be better than disabling annotation at all.

CHANGELOG.md
ilyagr: +1

Also, the section about `annotate` now needs to be moved up into the Unreleased section.

lib/src/annotate.rs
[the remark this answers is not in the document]
allonsy: Updated the code to now use let Some. I think it looks better now actually, thanks for the suggestions

CHANGELOG.md
allonsy: updated. I also changed my editor to not automatically format that file (source of the issue) going forward.

cli/src/commands/annotate.rs
allonsy: done

cli/src/commands/annotate.rs
allonsy: added

cli/src/config/templates.toml
allonsy: done

lib/src/annotate.rs
allonsy: Yeah, initially since I used BufRead/lines, everything had to be a string. After changing to split instead, I reverted to Vec<u8>

lib/src/annotate.rs
allonsy: yeah, this goes with your other comments. Now, when the expected file is not a file, we just treat it as absent

lib/src/annotate.rs
allonsy: renamed

lib/src/annotate.rs
allonsy: agreed

lib/src/annotate.rs
allonsy: I implemented conflicts now according to your suggestions. It now renders the file conflict dressings as added in the current commit

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,91 @@
...
+    let repo = workspace_command.repo();
+    let starting_commit =
+        workspace_command.resolve_single_rev(args.revision.as_ref().unwrap_or(&RevisionArg::AT))?;
+    let file_path = workspace_command.parse_file_path(&args.path)?;
+    let file_value = starting_commit.tree()?.path_value(&file_path)?;
+    let ui_path = workspace_command.format_file_path(&file_path);
+    if file_value.is_absent() {
+        return Err(user_error(format!("No such path: {ui_path}")));
+    }
+    if let Some(TreeValue::Tree(_)) = file_value.as_normal() {
```
yuja, line 58: nit: maybe we can use `file_value.is_tree()` here? It's not exactly the same, but I think it's okay to start annotation from conflicts.

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,91 @@
...
+        return Err(user_error(format!("file path is not a file: {ui_path}")));
```
yuja, line 59: nit: capitalize error message. Copying from `jj file show`, it could be rephrased as `"Path exists but is not a file"`

cli/src/commands/annotate.rs
```diff
@@ -0,0 +1,91 @@
...
+        .map_err(|e| CommandError::new(CommandErrorKind::Internal, e))?;
+
+    render_annotations(repo.as_ref(), ui, template, &annotations)?;
+    Ok(())
+}
+
+fn render_annotations(
+    repo: &dyn Repo,
+    ui: &mut Ui,
+    template_render: TemplateRenderer<Commit>,
```
yuja, line 78: nit: pass by reference

lib/src/annotate.rs
```diff
@@ -0,0 +1,382 @@
...
+    #[error("Unable to locate file: {0}")]
+    FileNotFound(String),
+    /// the file type is incorrect. Usually a directory was given but a regular
+    /// file is required
+    #[error("File {0} must be a regular file, not a directory")]
+    UnsupportedFileType(String),
+    /// the file is in a conflicted state and can therefore not be annotated
+    /// properly
+    #[error("File {0} is conflicted at commit: {1}")]
+    Conflicted(String, String),
```
yuja, line 51: nit: `UnsupportedFileType` and `Conflicted` are no longer used. Maybe `FileNotFound` can also replaced with an empty result.

lib/src/annotate.rs
yuja: Just in case this comment was buried, the intent is to remove `local_line_map.get_mut(old_commit_id)`/`local_line_map.get_mut(new_commit_id)` hash map lookup inside the lines loop.

allonsy: I'm happy to try out this change but I don't really get the reasoning behind it. Is it an issue to have the double mutation?

lib/src/annotate.rs
allonsy: Removing the unused errors. I'm hesitant about removing FileNotFound because this is a library function. 
From the UI, it already checks that the file exists, but, if other functions call this with a non-existent file, I think it makes sense to throw an error rather than empty results right?

cli/src/commands/mod.rs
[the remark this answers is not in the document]
allonsy: So are we set on moving this command to jj file annotate with an alias of jj annotate?

lib/src/annotate.rs
yuja: Yes, that's what I meant, thanks.

It might be better to add `struct { line_map: HashMap<usize, usize>, file_content: Vec<u8> }`. What `process_files_in_commits()` will do is basically forwarding `current_commit.line_map` to `parent_commit.line_map` by comparing `current_commit.file_content` and `parent_commit.file_content`.

allonsy: Ok, I pushed another iteration. Along your lines, PartialResults has been split out since it was slowly disappearing. 
In this iteration, I removed many instances where was pass &mut references and instead the return the results of each commit's computation which are then compiled into hashmaps close to the top level. 
With these changes, I removed Partial Results completely and just pass the necessary objects. Type aliases in the code help differentiate all the different HashMap types floating around to make it more clear what is being consumed and returned.

How does it look now? A step in the right direction or should I revert?

lib/src/annotate.rs
allonsy: We already special case the starting path in the cli component (if it doesn't exist we error out before going to jj_lib). 
With that in mind, I've taken your suggestion and removed FileNotFound (returning empty results if we can't find the starting file) and therefore AnnotateError in general (since there was nothing left). It now returns a BackendError like other functions in jj_lib.

I also integrated your latest changes to materialize_merge_results which impacted my code. Take a look and let me know if I integrated your changes correctly.

lib/src/annotate.rs
```diff
@@ -0,0 +1,394 @@
...
+    pub file_annotations: Vec<(CommitId, Vec<u8>)>,
+}
+
+/// A map from commits to line mappings.
+/// Namely, for a given commit A, the value is the mapping of lines in the file
+/// at commit A to line numbers in the original file
+type CommitLineMap = HashMap<CommitId, HashMap<usize, usize>>;
+
+/// Memoizes the file contents for a given version to save time
+type FileCache = HashMap<CommitId, Vec<u8>>;
```
yuja, line 60: It might be simpler to combine `CommitLineMap` and `FileCache`. They are conceptually separate, but their lifetimes are the same?

Doing that will probably help remove `.unwrap()` from `process_files_in_commits()`. If the cache entry of the current commit has been removed from the `HashMap`, new entry can be inserted to the map without dropping the previous borrow.

lib/src/annotate.rs
```diff
@@ -0,0 +1,394 @@
...
+            let (new_original_lines, new_commit_lines) = process_commit(
+                repo,
+                file_name,
+                &current_commit,
+                &mut file_cache,
+                current_commit_line_map,
+                &edge_list,
+            )?;
+            original_line_map.extend(new_original_lines);
+            commit_line_map.extend(new_commit_lines);
```
yuja, line 218: better to pass `&mut commit_line_map` and `&mut origianl_line_map` in to `process_commit()` instead of allocating temporary `HashMap`.

lib/src/annotate.rs
```diff
@@ -0,0 +1,394 @@
...
+            let parent_commit = repo.store().get_commit(&parent_edge.target)?;
+            let same_line_map = process_files_in_commits(
+                repo.store(),
+                file_name,
+                file_cache,
+                current_commit,
+                &parent_commit,
+            )?;
+
+            for (current_line_number, parent_line_number) in same_line_map {
```
yuja, line 258: resolve `commit_line_map[parent_commit.id()]` out of the loop. You can use `.entry(parent_id).or_default()`.
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry

Perhaps, `remap_line()` and `mark_lines_from_original()` can be inlined.

lib/src/annotate.rs
yuja: Thanks. I've added a couple of nits as inline comments.

I'm not sure about `PartialResults`. It might be useful to pass around global data as `&mut PartialResults` (see also inline comments about temporary `HashMap`), but that might be unnecessary if we can simplify and inline some of the functions.

lib/src/annotate.rs
yuja: Oh, and maybe we shouldn't replace existing `commit_line_map` at fork point?

lib/src/annotate.rs
allonsy: I played around with inlining quite a bit and the function just ended up a long spaghetti. In general I prefer smaller functions and breaking it out when things get big. 
I think same thing with entry. I tries to move that inside but the mutable borrows got really messy to the point where I was losing readability (or cloning excessively). 
If you can find a clean way to do what you are suggesting, let me know.

lib/src/annotate.rs
allonsy: I reverted the changes to original line map because that was excessive. I actually kind of like consuming the commit_line_map for the current commit map and returning the new parent's commit line maps so I left that in (so far). Let me know what you think.

lib/src/annotate.rs
yuja: Here we can initialize the parent line map first. It's nice that we move hash map lookup out of the loop. (And that's the reason I suggested to `.remove()` the `current_commit_line_map` first.)
```rust
            // Perhaps, the initialization can be combined with load_file_into_cache()??
            // let (parent_contents, parent_commit_line_map) = get_or_load(&mut ..)?;
            let parent_commit_line_map = commit_line_map
                .entry(parent_commit.id().clone())
                .or_default();
            for (current_line_number, parent_line_number) in same_line_map {
                if let Some(original_line_number) =
                    current_commit_line_map.remove(&current_line_number)
                {
                    parent_commit_line_map.insert(parent_line_number, original_line_number);
                }
            }
            // TODO: remove entry if parent_commit_line_map was empty?
            // or return/continue early if same_line_map...peek().is_none()
```

lib/src/annotate.rs
yuja: > I actually kind of like consuming the commit_line_map for the current commit map and returning the new parent's commit line maps 

I generally agree about that. However, in this case:
* the caller would have to merge existing `commit_line_map[id]` and `new_commit_lines[id]` line by line (at fork point, if I understand it correctly.)
* it's silly to hold temporary results in a `HashMap`

PhilipMetzger:
What are the remaining blockers here? As I'd really like to finally have a `annotate` command, even if it is not perfect yet.

yuja:
> What are the remaining blockers here? As I'd really like to finally have a `annotate` command, even if it is not perfect yet.

This and some related changes?
https://github.com/martinvonz/jj/pull/4176#discussion_r1735521093

I can rewrite the implementation as a follow up if he's okay.

allonsy:
I was working on a few different implementations, none of which I was
really happy with so I'm going to fix it up to a state that I liked over
the next two days and then we can go over final review

lib/src/annotate.rs
allonsy: thanks, that helped a bit simplify things, uploaded a new version addressing this

allonsy:
I think the last big issue remaining is command name. 
Its between raw `jj annotate` or `jj file annotate` I think file annotate makes the most sense at this point right?
@yuja 
@PhilipMetzger

yuja:
> I think the last big issue remaining is command name. Its between raw `jj annotate` or `jj file annotate` I think file annotate makes the most sense at this point right?

Yes, I would go with `file annotate` (and add `annotate` -> `file annotate` alias if we find it's the common requirement.)

lib/src/annotate.rs
```diff
@@ -0,0 +1,375 @@
...
+    load_file_into_cache(
+        file_cache,
+        store,
+        parent_commit.id(),
+        file_name,
+        &parent_commit.tree()?,
+    )?;
+
+    let current_contents = file_cache.get(current_commit.id()).unwrap();
+    let parent_contents = file_cache.get(parent_commit.id()).unwrap();
```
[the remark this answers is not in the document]
allonsy: Yes, I think if someone renames a directory in history to a regular file, this would panic. For now, I've added the empty vecs to get this across. I can come back later to combine some of the temporary caches created

lib/src/annotate.rs
```diff
@@ -0,0 +1,375 @@
...
+            let new_commit_lines = process_commit(
+                repo,
+                file_name,
+                &mut original_line_map,
+                &current_commit,
+                &mut file_cache,
+                current_commit_line_map,
+                &edge_list,
+            )?;
+            commit_line_map.extend(new_commit_lines);
```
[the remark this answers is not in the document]
allonsy: Yes, I think in more complicated topologies (like diamonds or pentagonal histories) the current implementation will lose the mappings. I've adjusted to merge the results instead, as suggested

cli/src/commands/mod.rs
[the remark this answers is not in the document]
PhilipMetzger: I see this as resolved with https://github.com/martinvonz/jj/pull/4176#issuecomment-2409845274.

CHANGELOG.md
```diff
@@ -172,12 +177,10 @@ Thanks to the people who made this release happen!
...
-
-* `next/prev` will no longer infer when to go into edit mode when moving from
-  commit to commit. It now either follows the flags `--edit|--no-edit` or it
-  gets the mode from `ui.movement.edit`.
-
-### Deprecations
+* New command `jj annotate` that annotates files line by line. This is similar
+  in functionality to git's blame. Invoke the command with `jj annotate <file_path>`.
+  The output can be customized via the `templates.annotate_commit_summary`
+  config variable.
```
yuja, line 183: bad edits

lib/src/annotate.rs
```diff
@@ -0,0 +1,374 @@
...
+    let mut result_lines = Vec::new();
+    original_contents
+        .split_inclusive(|b| *b == b'\n')
+        .enumerate()
+        .for_each(|(idx, line)| {
+            result_lines.push((
+                original_line_map.get(&idx).unwrap().clone(),
+                line.to_owned(),
+            ));
+        });
```
yuja, line 105: nit: `.map(..).collect()` instead of `.for_each(.. push(..))`.

lib/src/annotate.rs
```diff
@@ -0,0 +1,374 @@
...
+/// that parent Y has line x. The line mappings for all parents are returned
+/// along with any lines originated in the current commit
+fn process_commit(
+    repo: &dyn Repo,
+    file_name: &RepoPath,
+    original_line_map: &mut HashMap<usize, CommitId>,
+    commit_line_map: &mut CommitLineMap,
+    current_commit: &Commit,
+    file_cache: &mut FileCache,
+    edges: &Vec<GraphEdge<CommitId>>,
```
yuja, line 221: nit: `edges: &[GraphEdge<CommitId>]`

lib/src/annotate.rs
```diff
@@ -0,0 +1,374 @@
...
+    let parent_contents = file_cache.get(parent_commit.id()).unwrap();
+
+    Ok(get_same_line_map(current_contents, parent_contents))
+}
+
+/// For two files, get a map of all lines in common (e.g. line 8 maps to line 9)
+fn get_same_line_map(current_contents: &[u8], parent_contents: &[u8]) -> HashMap<usize, usize> {
+    let mut result_map = HashMap::new();
+    let inputs = vec![current_contents, parent_contents];
+    let diff = Diff::by_line(&inputs);
```
yuja, line 306: nit: `Diff::by_line([current_contents, parent_contents])` (Vec is no longer needed)

lib/src/annotate.rs
```diff
@@ -0,0 +1,374 @@
...
+                if !current_output.is_empty() {
+                    for _ in current_output.split_inclusive(|b| *b == b'\n') {
+                        current_line_counter += 1;
+                    }
+                }
+                if !parent_output.is_empty() {
+                    for _ in parent_output.split_inclusive(|b| *b == b'\n') {
+                        parent_line_counter += 1;
+                    }
+                }
```
yuja, line 330: nit: `line_counter += ..split_inclusive(..).count()`
(`if .is_empty()` is redundant)

lib/src/annotate.rs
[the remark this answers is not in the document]
yuja: It's relatively easy to trigger:
* base commit containing `A\nB\n`
* one side `A\n`
* other side `B\n`
* merge them as `A\nB\n`

That said, I won't suggest adding more edge cases to the CLI test. We should probably add library tests (as a follow up.)

lib/src/annotate.rs
```diff
@@ -0,0 +1,374 @@
...
+) -> Result<(), BackendError> {
+    if file_cache.contains_key(commit_id) {
+        return Ok(());
+    }
+
+    if let Some(file_contents) = get_file_contents(store, file_path, tree)? {
+        file_cache.insert(commit_id.clone(), file_contents);
+    } else {
+        // file is missing or is not a regular file (e.g. directory)
+        file_cache.insert(commit_id.clone(), Vec::new());
```
yuja, line 128: nit: (at your option) `get_file_contents()` can be changed to return `Vec<u8>` instead of `Option<_>`. An absent/non-file entry is equivalent to an empty file during annotation computation.

yuja (approved):
Thanks!

allonsy:
@yuja addressed the last round of quick fixes. Anything more before merging?

yuja:
> Anything more before merging?

No. Thanks for updating it quickly.

PR #4185, opened 2024-07-31
polish/4147: Change conflict hint if decendants resolves conflict.  

### LINKED ISSUE #4147: "jj status" output while fixing conflicts is misleading and offers bad advice
In my scenario:
* `jj rebase` - get some conflicts.
* `jj new changeid` to start resolving them.
* fix the conflicts, run `jj status` to check I've got everything.

The output from the `jj status` is shown:
image

However, I think the text is misleading: it's telling me to `jj new change_id` - but that change is already my parent, and my current change has resolved the conflicts. If I follow that advice, I'll effectively lose the fixes to the conflicts and need to start again. Once I re-resolve them, `jj status` will again advise me to drop the fixes and start again. This will continue until I learn to ignore the given advice, and I'll later discover I have a number of almost identical heads, all fixing the same conflicts.

essiene:
Fixes #4147 

Change behaviour to consider the state of the working commit, instead of the state of the tree. In particular:

* If working commit has conflicts: Print the existing message we print today.                                                                                                                                                                                
* If working commit has no conflicts:                                                                                                             
  * If any parent has conflicts, print: "Conflict in parent is resolved in working copy"                                                                 
  * If no parent has any conflicts: no message  

This behaviour resolves the issue raised in #4147.                                                                                                  
                                                                                                                                                 
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* `jj status` will show different messages in a conflicted tree, depending
+  on the state of the working commit. In particular, if a child commit fixes
+  a conflict in the parent, this will be reflected in the hint provided
+  by `jj status`
+
 * `jj diff --git` no longer shows the contents of binary files.
 
 * Windows binaries no longer require `vcruntime140.dll` to be installed
--- a/cli/src/commands/status.rs
+++ b/cli/src/commands/status.rs
@@ -92,21 +92,36 @@ pub(crate) fn cmd_status(
             writeln!(formatter)?;
         }
 
-        let wc_revset = RevsetExpression::commit(wc_commit.id().clone());
-        // Ancestors with conflicts, excluding the current working copy commit.
-        let ancestors_conflicts = workspace_command
-            .attach_revset_evaluator(
-                wc_revset
-                    .parents()
-                    .ancestors()
-                    .filtered(RevsetFilterPredicate::HasConflict)
-                    .minus(&revset_util::parse_immutable_expression(
-                        &workspace_command.revset_parse_context(),
-                    )?),
-            )?
-            .evaluate_to_commit_ids()?
-            .collect();
-        workspace_command.report_repo_conflicts(formatter, repo, ancestors_conflicts)?;
+        if wc_commit.has_conflict()? {
+            let wc_revset = RevsetExpression::commit(wc_commit.id().clone());
+
+            // Ancestors with conflicts, excluding the current working copy commit.
+            let ancestors_conflicts = workspace_command
+                .attach_revset_evaluator(
+                    wc_revset
+                        .parents()
+                        .ancestors()
+                        .filtered(RevsetFilterPredicate::HasConflict)
+                        .minus(&revset_util::parse_immutable_expression(
+                            &workspace_command.revset_parse_context(),
+                        )?),
+                )?
+                .evaluate_to_commit_ids()?
+                .collect();
+
+            workspace_command.report_repo_conflicts(formatter, repo, ancestors_conflicts)?;
+        } else {
+            for parent in wc_commit.parents() {
+                let parent = parent?;
+                if parent.has_conflict()? {
+                    writeln!(
+                        formatter.labeled("hint"),
+                        "Conflict in parent commit has been resolved in working copy"
+                    )?;
+                    break;
+                }
+            }
+        }
     } else {
         writeln!(formatter, "No working copy")?;
     }
--- a/cli/tests/test_status_command.rs
+++ b/cli/tests/test_status_command.rs
@@ -105,8 +105,9 @@ fn test_status_filtered() {
 }
 
 // See <https://github.com/martinvonz/jj/issues/3108>
+// See <https://github.com/martinvonz/jj/issues/4147>
 #[test]
-fn test_status_display_rebase_instructions() {
+fn test_status_display_relevant_working_commit_conflict_hints() {
     let test_env = TestEnvironment::default();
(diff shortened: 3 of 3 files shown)


lib/src/default_index/revset_engine.rs
```diff
@@ -1092,6 +1092,11 @@ fn build_predicate_fn(
                     .unwrap()
             })
         }
+        RevsetFilterPredicate::HasNoConflict => box_pure_predicate_fn(move |index, pos| {
+            let entry = index.entry_by_pos(pos);
+            let commit = store.get_commit(&entry.commit_id()).unwrap();
+            !commit.has_conflict().unwrap()
+        }),
```
yuja, line 1099: Maybe you miss the fact that `.filtered()` is a short for `.intersection(expr)`. You can instead negate the filter expression.
```rust
let has_no_conflict = RevsetExpression::filter(RevsetFilterPredicate::HasConflict).negated();
main_expr.intersection(&has_no_conflict)
```

cli/src/commands/status.rs
```diff
@@ -91,21 +91,54 @@ pub(crate) fn cmd_status(
...
+                .attach_revset_evaluator(
+                    wc_revset
+                        .children()
+                        .descendants()
+                        .filtered(RevsetFilterPredicate::HasNoConflict),
+                )?
+                .evaluate_to_commits()?
+                .collect::<Vec<_>>()
+                .pop()
+            {
```
yuja, line 107: Does it make sense to check wc children (which usually doesn't exist)? The user could `edit` a conflict commit, but in that case, I don't think they would be interested in whether descendants have a conflict or not.

essiene: The rationale here is so that in the case  where a descendant has already resolved the conflict, printing out the same instructions to resolve the conflict again would be misleading.

This could happen if:

```
# wc has conflict 
$ jj new

# 1. resolve conflict, but don't squash yet 
# 2. now step back to previous commit for whatever reason 

$ jj prev --edit
$ jj status

# informing to resolve the conflict all over again here, is a bit confusing, since we know it has already been resolved. Probably better to state clearly that it's been resolved in a descendant. Though current message in the PR could be improved?
```

wdyt?

yuja: I don't think it's a common mistake we need to catch. It's also unclear whether the wc child contains only conflict resolution. If a rebased commit happened to resolve conflict in parent for example, it may contain other unrelated changes, and the user will have to resolve the root conflict anyway.

lib/src/default_index/revset_engine.rs
essiene: Ah thanks. I indeed didn't read that far. That said, this is no longer needed as I'm going with your advice and ignoring descendants. So I've reverted the commit that added this.

cli/src/commands/status.rs
essiene: Makes sense. I've removed this. So all we're doing now is:

* If wc is conflict free:
  * if parent has a conflict, mention that wc resolves the conflict in the parent.
  * if parent has no conflict. do nothing.
* if wc has conflicts: print the same message we printed before, regardless of the state of the children.

This behaviour resolves the main complain in #4147 so I believe this is good enough.

yuja (approved):
Thanks. We might want to handle `wc_parent == root_conflict` better, but this PR looks strictly better than the current state.

cli/src/commands/status.rs
```diff
@@ -92,21 +92,36 @@ pub(crate) fn cmd_status(
...
-                    )?),
-            )?
-            .evaluate_to_commit_ids()?
-            .collect();
-        workspace_command.report_repo_conflicts(formatter, repo, ancestors_conflicts)?;
+        if wc_commit.has_conflict()? {
+            let wc_revset = RevsetExpression::commit(wc_commit.id().clone());
+
+            // Ancestors with conflicts, excluding the current working copy commit.
+            let ancestors_conflicts: Vec<_> = workspace_command
```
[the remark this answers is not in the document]
essiene: Residue. Thanks for catching. Removed.

cli/src/commands/status.rs
```diff
@@ -92,21 +92,36 @@ pub(crate) fn cmd_status(
...
+                .evaluate_to_commit_ids()?
+                .collect();
+
+            workspace_command.report_repo_conflicts(formatter, repo, ancestors_conflicts)?;
+        } else {
+            for parent in wc_commit.parents() {
+                let parent = parent?;
+                if parent.has_conflict()? {
+                    writeln!(
+                        formatter.labeled("conflict"),
```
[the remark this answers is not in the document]
essiene: Done.

CHANGELOG.md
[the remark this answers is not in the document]
essiene: Done.

cli/tests/test_status_command.rs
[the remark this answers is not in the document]
essiene: Done. I moved everything back into the original test. I've added comments at each stage to show what's being tested.

cli/tests/test_status_command.rs
```diff
@@ -143,7 +143,7 @@ fn test_status_display_rebase_instructions() {
     test_env.jj_cmd_ok(&repo_path, &["new", "--message", "boom-cont"]);
     test_env.jj_cmd_ok(&repo_path, &["new", "--message", "boom-cont-2"]);
 
-    let stdout = test_env.jj_cmd_success(&repo_path, &["log", "-r", "::@"]);
+    let stdout = test_env.jj_cmd_success(&repo_path, &["log", "-r", "root()::"]);
```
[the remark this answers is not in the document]
essiene: Thanks. Updated everywhere.

yuja (approved):

PR #4187, opened 2024-07-31
cli: drop built-in `pager_handle` when appropriate

ErichDonGubler:
The `:builtin` pager currently has some undesirable behavior when a user is attempting to quit `minus` paging via keybindings (i.e., `q` by default). This happens whenever `jj` has not yet sent all output to the pager thread; it finishes generating all output before heeding the user's request to stop the command. For commands like `jj log --revision "..trunk()"` in enormous repositories like [`mozilla-central`] (where Firefox is hosted), this can take a nontrivial amount of time, and is almost certainly not what the user wants.

[`mozilla-central`]: https://hg.mozilla.org/mozilla-central

Notably, this is a major difference between the built-in and the external pager implementations. Exiting before `jj` sends all output to `less` exits more-or-less immediately, with `jj` also exiting more-or-less immediately with `BROKEN_PIPE_EXIT_CODE`. This makes sense, as child process input is indeed a broken pipe. This seems like the behavior we should use for the built-in pager, too.

Solution: interrupt `jj`'s loop that sends output to the built-in pager by placing the existing `BuiltinPager::pager_handle` behind `Arc<RwLock<Option<…>>>` layers, and:

1. Discarding `pager_handle` in contexts where we know we should stop sending output. To wit, these are:

	1. In an exit callback registered with the `minus` pager.
	2. When the `minus` paging thread ends. This is technically redundant with (1), but doesn't seem to hurt anything.
	3. When `BuiltinPager::finalize` is called.

2. Forcing `<BuiltinPager as Write::write` to gain a read lock on `pager_handle`'s contents, otherwise returning a broken pipe error.

# Checklist

If applicable:

### DIFF
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -15,6 +15,7 @@
 use std::io::{IsTerminal as _, Stderr, StderrLock, Stdout, StdoutLock, Write};
 use std::process::{Child, ChildStdin, Stdio};
 use std::str::FromStr;
+use std::sync::{Arc, RwLock};
 use std::thread::JoinHandle;
 use std::{env, fmt, io, mem};
 
@@ -45,7 +46,7 @@ enum UiOutput {
 
 /// A builtin pager
 pub struct BuiltinPager {
-    pager: MinusPager,
+    pager: Arc<RwLock<Option<MinusPager>>>,
     dynamic_pager_thread: JoinHandle<()>,
 }
 
@@ -56,8 +57,14 @@ impl Write for &BuiltinPager {
     }
 
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let pager = self
+            .pager
+            .read()
+            .unwrap()
+            .clone()
+            .ok_or_else(|| io::Error::from(io::ErrorKind::BrokenPipe))?;
         let string = std::str::from_utf8(buf).map_err(io::Error::other)?;
-        self.pager.push_str(string).map_err(io::Error::other)?;
+        pager.push_str(string).map_err(io::Error::other)?;
         Ok(buf.len())
     }
 }
@@ -70,6 +77,7 @@ impl Default for BuiltinPager {
 
 impl BuiltinPager {
     pub fn finalize(self) {
+        self.pager.write().unwrap().take();
         let dynamic_pager_thread = self.dynamic_pager_thread;
         dynamic_pager_thread.join().unwrap();
     }
@@ -81,13 +89,24 @@ impl BuiltinPager {
         pager
             .set_exit_strategy(minus::ExitStrategy::PagerQuit)
             .expect("Able to set the exit strategy");
-        let pager_handle = pager.clone();
+        let pager_handle = Arc::new(RwLock::new(Some(pager.clone())));
+
+        pager
+            .add_exit_callback(Box::new({
+                let pager_handle = pager_handle.clone();
+                move || {
+                    pager_handle.write().unwrap().take();
+                }
+            }))
+            .unwrap();
 
         BuiltinPager {
-            pager,
+            pager: pager_handle.clone(),
             dynamic_pager_thread: std::thread::spawn(move || {
                 // This thread handles the actual paging.
-                minus::dynamic_paging(pager_handle).unwrap();
+                let res = minus::dynamic_paging(pager);
+                pager_handle.write().unwrap().take();
+                res.unwrap();
             }),
         }
     }


ErichDonGubler:
I'm happy to sign a CLA and check the remaining boxes in the PR template, just want to validate that this fix is interesting from maintainers before I go to the effort of doing so.

ilyagr:
This sounds useful to me. I'm not an expert on multi-threading, but I'll take a look at the code.

I'm also hoping that your approach will solve https://github.com/martinvonz/jj/issues/4182#issuecomment-2259315065. If `dynamic_pager` fails, it might allow `jj` to not panic but instead notice the error relatively early, right?

Signing the CLA would help, especially for the Googler contributors.

ErichDonGubler:
CLA has been signed.

cli/src/ui.rs
```diff
@@ -45,7 +46,7 @@ enum UiOutput {
 
 /// A builtin pager
 pub struct BuiltinPager {
-    pager: MinusPager,
+    pager: Arc<RwLock<Option<MinusPager>>>,
```
ilyagr, line 49: I'd add a comment:

> `Option<MinusPager>` is set to None after the pager quits (e.g. the user quits the pager early).

ilyagr (commented):
This LGTM. Thank you!

I'd feel mildly more comfortable if somebody took a second look, since I'm mildly afraid of multi-threading, but if everybody else is busy I'll approve this in a day or two.

cli/src/ui.rs
```diff
@@ -81,13 +89,24 @@ impl BuiltinPager {
...
             .set_exit_strategy(minus::ExitStrategy::PagerQuit)
             .expect("Able to set the exit strategy");
-        let pager_handle = pager.clone();
+        let pager_handle = Arc::new(RwLock::new(Some(pager.clone())));
+
+        pager
+            .add_exit_callback(Box::new({
+                let pager_handle = pager_handle.clone();
+                move || {
+                    pager_handle.write().unwrap().take();
```
yuja, line 98: In order to disable late `write()`s, we don't have to drop `pager` object. Instead, maybe we can use `AtomicBool` to be set on pager quit.

BTW, this PR revealed that the current builtin pager is fundamentally broken. Because the backing channel is unbounded, `.write()` will never stuck and burn CPU time.

ilyagr: > Because the backing channel is unbounded, .write() will never stuck and burn CPU time.

What do you mean? Are you talking about the problem this PR solves or about a separate problem? Are you talking about something we do or `minus`'s design?

ilyagr: Also, AtomicBool sounds fine to me as far as this PR goes, but I'll mention that I was considering replacing the `Option` here with a `Result` later for https://github.com/martinvonz/jj/issues/4182#issuecomment-2259315065. I am not 100% sure that's the best solution to that problem, but it would need preserving more information than just one AtomicBool.

yuja: It's kinda the same problem under the hood, but user might not notice. Basically, if the pager is minus, `jj log -r.. -p` will scan all ancestors, materialize diffs, and the whole output will be buffered in memory.

Our expectation is that `.write()` into the pager will stuck if the pipe is full, but `minus` doesn't because the underlying channel is unbounded.

ilyagr: What would be symptoms of this? Running out of memory? Something more serious?

Would anything short of redesigning minus help? Though, they might accept a PR.

I think the pager Sapling used, streampager, was more carefully designed for huge inputs, but it seemed abandoned last I checked.

yuja: Wasted CPU time is probably more important. It will surely reach the memory limit, but the output speed wouldn't be that high.

There's a related issue in minus. I guess this wouldn't be an easy issue to fix.
https://github.com/AMythicDev/minus/issues/106

cli/src/ui.rs
ErichDonGubler: Agreed! This is one of the things I was hoping to do before taking this out of draft. 🙂 Will attend to this soon.

cli/src/ui.rs
ErichDonGubler: @yuja: It's true that we _could_ use `AtomicBool` for signaling here, but I see no advantage to doing so, and a few advantages in keeping the current `Option` container:

1. If we don't intend to use the pager anymore, this makes it inaccessible by design. It matches our intent; I think it will be far less surprising for those unfamiliar with this code to see this handle being discarded than an `AtomicBool` as a peer.
2. Because `minus`' pager handle is implemented as a sender and receiver channel pair for duplix comms. with the paging thread, dropping `pager_handle`'s contents should free the CLI thread's `Receiver`'s resource (which is very relevant to the above discussion about resource usage with lines that haven't been displayed yet). It's unclear to me what comms. might be queued up in return from the paging thread, but this sidesteps any question of whether they're significant.

yuja: Hmm, I would say `AtomicBool` is simpler and that's the advantage.

As far as I understand the implementation, `minus::dynamic_paging()` doesn't drop their `tx` end, so dropping our `pager` instance wouldn't signal the termination. Ideally, all `rx` ends should be dropped on pager quit, so any further `pager.push_str()` automatically fails. However, `minus` API appears not designed as such use case in mind.

ilyagr: Is the point that `AtomicBool` avoids the possibility of deadlocks? While it's more primitive, that makes its docs more complicated (https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html, for example), so I couldn't immediately figure that out.

Like Erich, I liked the idea of making invalid states inexpressible, but I'm not sure about the downsides of RwLock and whether they are worth it.

yuja: > Is the point that `AtomicBool` avoids the possibility of deadlocks?

No? I just mean boolean flag is easier to follow than heavy lifting on interior mutability.

BTW, judging from the discussion on Discord, it might be better to just switch to streampager.

yuja:
Sent simpler version as https://github.com/martinvonz/jj/pull/4209

ErichDonGubler:
Dropping in favor of https://github.com/martinvonz/jj/pull/4209.

cli/src/ui.rs
ilyagr: If anyone is reading this, here is what seems like a nice book that discusses Atomic types:

https://marabos.nl/atomics/atomics.html#atomic-load-and-store-operations

PR #4188, opened 2024-07-31
fixes for paths on Windows

ilyagr:
This is my attempt to replace #4184 with something more principled. It is, however, not so principled as to replace `fs::canonicalize` with `dunce::canonicalize` everywhere.

I tried to improve (what used to be) `absolute_git_source` in general, but I may have introduced some bugs.

It feels a bit like whack-a-mole. I couldn't find rules on what kind of Windows paths Git or libgit2 accept in clone URLs so far. I'm also introducing more and more ways the same path can look in a test.

I'm still looking into it, but I am running out of patience (though perhaps I'll get some better ideas later). Advice is welcome. Do you think using `dunce` everywhere would help, for example? Also, my Windows setup is awkward; I'm running it in a VM and it's not very fast and runs out of memory or causes the main system to run out of memory occasionally. Perhaps because of that, or perhaps it's on a USB drive, it also freezes occasionally.

At the moment, it fails a *different* test when merged with `git2` upgrade that #4080 does, see https://github.com/martinvonz/jj/pull/4183. (There are also more trivial but annoying test failures with TEST_ENV)

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -1165,6 +1165,7 @@ pub fn sanitize_git_url_if_path(cwd: &Path, source: &str) -> String {
         // TODO: This won't work for Windows UNC path or (less importantly) if the
         // original source is a non-UTF Windows path. For Windows UNC paths, we
         // could use dunce, though see also https://gitlab.com/kornelski/dunce/-/issues/7
+        // TODO: double-check about UNC paths; what does Git do?
         cwd.join(source).to_slash().map_or_else(
             // It's less likely that cwd isn't utf-8, so just fall back to original source.
             || source.to_owned(),
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -18,11 +18,12 @@ use std::borrow::Cow;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::default::Default;
 use std::io::Read;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::{fmt, str};
 
 use git2::Oid;
 use itertools::Itertools;
+use path_slash::PathBufExt;
 use tempfile::NamedTempFile;
 use thiserror::Error;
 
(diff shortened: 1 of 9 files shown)


lib/src/git.rs
```diff
@@ -1164,7 +1198,7 @@ pub fn set_remote_url(
     })?;
 
     git_repo
-        .remote_set_url(remote_name, new_remote_url)
+        .remote_set_url(remote_name, &sanitize_git_url_if_path(cwd, new_remote_url))
```
yuja, line 1201: nit: this should be handled by CLI because it's UI issue to resolve relative path.

(BTW, I think it's good idea to move the path conversion function to library.)

lib/src/git.rs
```diff
@@ -1143,10 +1144,43 @@ pub fn rename_remote(
...
+    // exits, and fails because '$PWD/https' is unsupported protocol. Since it would
+    // be tedious to copy the exact git (or libgit2) behavior, we simply assume a
+    // source containing ':' is a URL, SSH remote, or absolute path with Windows
+    // drive letter.
+    if looks_like_a_path_to_git(source) && Path::new(source).exists() {
+        // TODO: This won't work for Windows UNC path or (less importantly) if the
+        // original source is a non-UTF Windows path. For Windows UNC paths, we
+        // could use dunce, though see also https://gitlab.com/kornelski/dunce/-/issues/7
+        // TODO: double-check about UNC paths; what does Git do?
+        cwd.join(source).to_slash().map_or_else(
```
yuja, line 1169: Can't we simply do replace `MAIN_SEPARATOR` to slash (if `cfg!(windows)`)? `path-slash` might be optimized if we have to deal with `Path`/`OsStr`, but here we build `String` anyway.

ilyagr: This is on my list to try and see if it makes things saner. My main worry is about UNC paths, especially the ones appearing in our tests. I would have to find out experimentally whether I can get rid of them via `dunce`, and/or I'd have to find out how libgit2 deals with paths like:

- `\\?\C:/Path/`
- `//?/C:/Path/`

Neither of these is correct according to Microsoft, but I think `libgit2` definitely chokes on the correct `\\?\C:\Path\`. I wish I knew what Git did; I think we'd have to ask or read the source to find out.

(As for dependency on path-slash, that library hasn't been updated in years and it's probably pretty simple, so we could re-implement it if needed)

PR #4189, opened 2024-07-31
Expose some internal constructors

torquestomp:
Many places in jj_cli assume a working std::fs for interrogating repo storage types before loading the repos.  Rather than abstract all of that away and introduce complexity, let's expose a couple simple constructor methods to allow a determined client to bypass std::fs interactions in environments that don't support them (for instance, when the repo is stored in a cloud filesystem that isn't mounted locally).

# Checklist

If applicable:

### DIFF
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -604,6 +604,26 @@ pub struct RepoLoader {
 }
 
 impl RepoLoader {
+    pub fn new(
+        repo_path: PathBuf,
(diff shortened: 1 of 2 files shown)


lib/src/repo.rs
PhilipMetzger, line 1: Nit: This commit needs to be formatted

PhilipMetzger (approved):
LGTM

PR #4195, opened 2024-08-01
draft: FR: Add new revset `builtin_immutable_heads()`

### LINKED ISSUE #4162: FR: Add `builtin_immutable_heads()`

When redefining `immutable_heads()`, there is no way to include the default set. Since the [default set occasionally changes](https://github.com/martinvonz/jj/commit/ea3a574e36ece4a3772765ef6fcc81094915a068), the user has to change their config to get the new settings.

Define the default set in `builtin_immutable_heads()` and redefine `immutable_heads()` like so:

```toml
[revset-aliases]
'builtin_immutable_heads()' = 'trunk() | tags() | untracked_remote_branches()'
'immutable_heads()' = 'builtin_immutable_heads()'
```

If a user overrides `immutable_heads()` and includes `builtin_immutable_heads()`, updates to jj's default will automatically be used.

If the user _doesn't_ want to include the default set, they simply omit `builtin_immutable_heads()`. This means this change wouldn't adversely affect anyone until they opt-in by adding `builtin_immutable_heads()` to their config.

If we wanted the user to opt out of the default set rather than opting in, the definition could be like so:

```toml
[revset-aliases]
'immutable_heads()' = 'none()'

# This becomes the revset function that jj uses
'real_immutable_heads()' = 'immutable_heads() | trunk() | tags() | untracked_remote_branches()'
```

The user could still opt out by defining `real_immutable_heads()` instead.

However, I think this alternative is too opinionated, not backwards compatible, and harder to explain:

> If you want to include the default set of immutable heads, define `immutable_heads()`. If you want to completely replace the set, use `real_immutable_heads()`.

Versus:

> Define `immutable_heads()` as desired. The default revset can be included as `builtin_immutable_heads()`.

**Additional context**

This was [discussed in another pull request](https://github.com/martinvonz/jj/pull/4121#discussion_r1691351101) with @yuja, @scott2000, and @ilyagr.

essiene:
Fixes #4162 

* Implement `builtin_immutable_heads()` in code, so it can't be overwritten by the user.
* This needs us to define `trunk()` in code to avoid duplication and drift.
* Redefine `immutable_heads()` in terms of `builtin_immutable_heads()`

# Checklist

If applicable:


essiene (commented):
> SELF REMINDER: squash commits before submitting.

Done.

martinvonz (commented):
> I think we need a way to override `trunk()` since some repos have main branches called something other than `main`, `master` or `trunk`.

Done.

essiene:
> I think we need a way to override `trunk()` since some repos have main branches called something other than `main`, `master` or `trunk`.

I've changed my approach. Instead of defining this in jj-lib, I'm now defining it as an alias in a new config source layer that takes precedence over all other config sources: builtins.toml

So since it's just a regular alias, it can be defined in terms of `trunk()` with no issues, as opposed to my previous approach.

wdyt?

If this looks good, I'll go ahead to add some tests.

cli/src/config.rs
```diff
@@ -217,6 +220,7 @@ impl LayeredConfigs {
             repo: None,
             env_overrides: env_overrides(),
             arg_overrides: None,
+            builtin_overrides: builtin_overrides(),
```
yuja, line 223: Can't we just let user override the builtin if they really want? That's harmless anyway.

Even if we want to be strict about that, I (as a user) would expect some error/warning instead of silently shadowing the user-defined value.

essiene: > Can't we just let user override the builtin if they really want? That's harmless anyway.

It feels like we'd end up in exactly the same place we started when #4162 was filed. No? Won't it be better to actually do nothing instead of that?

> 
> Even if we want to be strict about that, I (as a user) would expect some error/warning instead of silently shadowing the user-defined value.

I went this route. Could you take a look at where I hooked in the check? I feel like I just hit it on the head until it worked, but I don't know if there's a better place to insert this check? I also thought it may be a good idea to check for any user defined revset-aliases that start with `builtin_` instead of this particular alias.

Thoughts please! :-D

I still have to add tests and maybe some minor cleanup, but I want to check that there's no better way to do this first.

Thanks

yuja: > > Can't we just let user override the builtin if they really want? That's harmless anyway.
> 
> It feels like we'd end up in exactly the same place we started when #4162 was filed. No? Won't it be better to actually do nothing instead of that?

It's not recommended to override builtins, but doing that wouldn't make the situation worse. That's my point. And silent failure is the worst thing IMHO because the user can't see why their configuration doesn't work.

> > Even if we want to be strict about that, I (as a user) would expect some error/warning instead of silently shadowing the user-defined value.
> 
> I went this route. Could you take a look at where I hooked in the check?

I think warnings can be emitted by `revset_util::load_revset_aliases()`.

BTW, if we implement this, `immutable()` and `mutable()` are eligible for warnings.

essiene: > > > Can't we just let user override the builtin if they really want? That's harmless anyway.
> > 
> > 
> > It feels like we'd end up in exactly the same place we started when #4162 was filed. No? Won't it be better to actually do nothing instead of that?
> 
> It's not recommended to override builtins, but doing that wouldn't make the situation worse. That's my point. And silent failure is the worst thing IMHO because the user can't see why their configuration doesn't work.
> 
> > > Even if we want to be strict about that, I (as a user) would expect some error/warning instead of silently shadowing the user-defined value.
> > 
> > 
> > I went this route. Could you take a look at where I hooked in the check?
> 
> I think warnings can be emitted by `revset_util::load_revset_aliases()`.
> 
> BTW, if we implement this, `immutable()` and `mutable()` are eligible for warnings.

will a simple warning like: 'redefining `immutable()` and `mutable()` is not advised. redefine `immutable_heads()` instead' ?

Or does it make sense to also add `builtin_` versions of these as well?

```toml
'builtin_immutable()' = '::(immutable_heads() | root())'
'builtin_mutable()' = '~builtin_immutable()'
'immutable()' = 'builtin_immutable()'
'mutable' = 'builtin_mutable()'
```

cli/src/config.rs
```diff
@@ -216,6 +218,7 @@ pub struct LayeredConfigs {
     repo: Option<config::Config>,
     env_overrides: config::Config,
     arg_overrides: Option<config::Config>,
+    builtin_overrides: config::Config,
```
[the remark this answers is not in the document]
essiene: As you pointed out earlier, it doesn't break anything if we don't have it.

The two possible outcomes are:
  - No builtin layer: User sees warning, `jj` behaviour is inconsistent. 
  - Builtin layer: User sees warning, but `jj` behaves consistently.

I do think consistent behaviour is a good thing to strive for as it removes a class bugs and simplifies the user experience.

All that said, I'm starting to wonder if the better thing here is to remove the possibility of shadowing/redefining arbitrary builtins completely?

The main problem is that all these layers populate a single namespace.

Supposing we reserved the anonymous namespace for builtins that can't be shadowed, and introduced a 'user' namespace that the user is allowed to shadow/redefine.

We would provide default definitions (same as today) for things like 'user.trunk()' and 'user.immutable_heads()', but 'mutable()' and 'immutable()' would be in the anonymous namespace and not overwritable.

The thing that would make this all work though is that name resolution doesn't need the namespace and we search the anonymous namespace before we search the user namespace. Also, users can still disambiguate if they need to by specifying the fully qualified name.

What I'm doing here with the builtin layer is a hack, which I believe is why it feels wrong to you. Stepping back and introducing namespaces could be a better approach.

wdyt?

I could always submit this, and while we figure out a better approach or I could drop this if a better approach is obvious.

cli/src/config.rs
essiene: Done. PTAL.

cli/src/config.rs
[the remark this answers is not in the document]
yuja: > The two possible outcomes are:
> 
> * No builtin layer: User sees warning, `jj` behaviour is inconsistent.
> * Builtin layer: User sees warning, but `jj` behaves consistently.

To me, the latter is less consistent because defaults should be overridden by user, and builtins are more or less defaults.

> All that said, I'm starting to wonder if the better thing here is to remove the possibility of shadowing/redefining arbitrary builtins completely?

The major problem of disabling shadowing is that revset symbols include tags and branches, so the "builtin" symbols set can be updated dynamically by `jj git fetch`. It's weird if my "foo" alias got overridden by "foo" tag crated by someone else.

Function aliases doesn't have this problem, but they can still get broken if we add new builtin functions. Warning seems okay in that situation, but silent behavior change is bad.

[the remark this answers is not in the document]
essiene: Ok. I think I have a much better understanding now. Thanks for the explanations.

I've done the following:

- Defined `builtin_immutable_heads()` in the `default` layer. which means it can indeed be redefined.
- Added warnings when the user redefines `builtin_immutable_heads()`, `immutable()` or `mutable()`. The warning points them at `immutable_heads()` as the recommended function to redefine.

cli/src/config.rs
essiene: warnings added.

CHANGELOG.md
```diff
@@ -65,6 +65,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* Define `immutable_heads()` revset alias in terms of a new `builtin_immutable_heads()`.
+  This enables users to redefine `immutable_heads()` as they wish, but still
+  have `builtin_immutable_heads()` which can should not be redefined. Also 
+  issue warnings if any of `mutable()`, `immutable()` or `builtin_immutable_heads()`
+  are redefined by the user.
```
yuja, line 72: nit: `which can should` -> `which should`

And I would remove the last "Also issue warnings ..." sentence because user wouldn't have to know that beforehand.

cli/src/revset_util.rs
```diff
@@ -107,6 +107,31 @@ impl<'repo> RevsetExpressionEvaluator<'repo> {
...
+        ConfigSource::Env | ConfigSource::User | ConfigSource::Repo | ConfigSource::CommandArg => {
+            let checked_mutability_builtins =
+                ["mutable()", "immutable()", "builtin_immutable_heads()"];
+
+            if checked_mutability_builtins.contains(&name) {
+                writeln!(
+                    ui.warning_default(),
+                    "redefining `revset-aliases.{}` is not recommended; redefine \
+                     `immutable_heads()` instead",
+                    name
```
yuja, line 125: nit: capitalize "Redefining", and `name` can be inlined as `revset-aliases.{name}`.

cli/src/revset_util.rs
```diff
@@ -107,6 +107,31 @@ impl<'repo> RevsetExpressionEvaluator<'repo> {
     }
 }
 
+fn warn_user_redefined_builtin(
+    ui: &Ui,
+    source: &ConfigSource,
+    name: &str,
+) -> Result<(), CommandError> {
+    match source {
+        ConfigSource::Env | ConfigSource::User | ConfigSource::Repo | ConfigSource::CommandArg => {
```
yuja, line 116: nit: (at your option) I would add early return `if source == ConfigSource::Default`, but exhaustive match seems also good.

cli/src/revset_util.rs
```diff
@@ -107,6 +107,31 @@ impl<'repo> RevsetExpressionEvaluator<'repo> {
...
+            let checked_mutability_builtins =
+                ["mutable()", "immutable()", "builtin_immutable_heads()"];
```
yuja, line 118: nit: technically, these aliases could be overridden by `mutable(    )`, etc., but it's unlikely, and missed warning isn't problem.

(no action needed)

yuja (approved):
Thanks!

cli/src/revset_util.rs
essiene: Acknowledged.

cli/src/revset_util.rs
essiene: Done.

cli/src/revset_util.rs
essiene: I've left the exhaustive match, but I've moved the `ConfigSource::Default` case to the top, so it's easier to see that we bail there with a noop.

CHANGELOG.md
essiene: Whoops. Thanks for catching. Also removed last sentence.

essiene:
> Thanks!

Yw, and thanks again for the patient explanations. Much appreciated.

ilyagr:
We probably want `builtin_revsets_log` as well, right? [Somebody on Discord](https://discord.com/channels/968932220549103686/968932220549103689/1273515348825608202) seems to want to show some commits on the main branch in their default revset, and I think being able to set that up in terms of `builtin_revsets_log` should be good.

Issue #4182, opened 2024-07-30
Builtin pager does not work when using Git Bash on Windows and `jj` panics


jennings:
## Description

When using the builtin pager on the MinTTY Git Bash on Windows, pagination panics:

```
$ RUST_BACKTRACE=1 jj
Hint: Use `jj -h` for a list of available commands.
Run `jj config set --user ui.default-command log` to disable this message.
thread '<unnamed>' panicked at cli\src\ui.rs:90:53:
called `Result::unwrap()` on an `Err` value: Setup(InvalidTerminal)
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
thread 'main' panicked at cli\src\ui.rs:74:37:
called `Result::unwrap()` on an `Err` value: Any { .. }
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

## Steps to Reproduce the Problem

1. Open Git Bash from the start menu
2. `jj log`

## Suggestions

The failure seems to occur because stdout's `is_tty()` returns false, so Minus returns an `Err(SetupError(InvalidTerminal))`.

If possible, we can work around it, but we should also document that the user can avoid the problem by:

- Adding Git Bash to Windows Terminal. The configuration is:
  ```
  {
    "profiles": {
      "list": [
        ...
        {
          "guid": "{00000000-0000-0000-ba54-000000000002}",
          "commandline": "%PROGRAMFILES%/Git/bin/bash.exe -i -l",
          "icon": "%PROGRAMFILES%/Git/mingw64/share/git/git-for-windows.ico",
          "name" : "Bash",
          "startingDirectory" : "%USERPROFILE%"
        }
      ]
    }
  }
  ```
- Using PowerShell or cmd (preferrably in Windows Terminal, but the standard host will do as well)
- Disabling pagination (though who knows if another error will occur somewhere else)

ilyagr:
On discord, you suggested `winpty`. Does that work?

> Disabling pagination (though who knows if another error will occur somewhere else)

If the problem is just with our internal pager, we could disable it on that error.

zschreur:
fwiw when I went and followed these steps it fixed it for me: https://martinvonz.github.io/jj/latest/windows/#pagination

jennings:
> On discord, you suggested `winpty`. Does that work?

[...]

PR #4197, opened 2024-08-02
built-in pager: write a message to the user if pager doesn't start

ilyagr:
This will hopefully improve the experience of users who encounter #4182.

# Checklist

If applicable:

### DIFF
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -18,7 +18,8 @@ use std::str::FromStr;
 use std::thread::JoinHandle;
 use std::{env, fmt, io, mem};
 
-use minus::Pager as MinusPager;
+use indoc::indoc;
+use minus::{MinusError, Pager as MinusPager};
 use tracing::instrument;
 
 use crate::command_error::{config_error_with_message, CommandError};
@@ -46,7 +47,7 @@ enum UiOutput {
 /// A builtin pager
 pub struct BuiltinPager {
     pager: MinusPager,
-    dynamic_pager_thread: JoinHandle<()>,
+    dynamic_pager_thread: JoinHandle<Result<(), MinusError>>,
 }
 
 impl Write for &BuiltinPager {
@@ -69,9 +70,9 @@ impl Default for BuiltinPager {
 }
 
 impl BuiltinPager {
-    pub fn finalize(self) {
+    pub fn finalize(self) -> Result<(), MinusError> {
         let dynamic_pager_thread = self.dynamic_pager_thread;
-        dynamic_pager_thread.join().unwrap();
+        dynamic_pager_thread.join().unwrap()
     }
 
     pub fn new() -> Self {
@@ -87,7 +88,7 @@ impl BuiltinPager {
             pager,
             dynamic_pager_thread: std::thread::spawn(move || {
                 // This thread handles the actual paging.
-                minus::dynamic_paging(pager_handle).unwrap();
+                minus::dynamic_paging(pager_handle)
             }),
         }
     }
@@ -130,7 +131,44 @@ impl UiOutput {
                 }
             }
             UiOutput::BuiltinPaged { pager } => {
-                pager.finalize();
+                if let Err(minus_error) = pager.finalize() {
+                    writeln!(
+                        ui.warning_default(),
+                        "Built-in pager failed to start: {minus_error}",
+                    )
+                    .ok();
+                    writeln!(ui.warning_default(), "The output of this command is lost.").ok();
+                    if matches!(
+                        minus_error,
+                        MinusError::Setup(minus::error::SetupError::InvalidTerminal)
+                    ) {
+                        if cfg!(windows) {
+                            writeln!(
+                                ui.hint_default(),
+                                indoc! {r#"
+                                    jj's builtin pager is likely incompatible with this terminal
+
+                                    This is known to happen with `mintty`, the default Git Bash terminal on Windows.
+
+                                    Possible workarounds:
+                                    - Use `jj --no-pager`
+                                    - Configure a different pager, see https://martinvonz.github.io/jj/latest/windows/#pagination for Git Bash on Windows
+                                    - Use a different terminal (e.g. Windows Terminal or the Command Prompt)
+                                    - Use `winpty jj ...`; `winpty` comes with Git Bash or Cygwin."#
+                                }
+                            ).ok();
+                        } else {
+                            writeln!(
+                                ui.hint_default(),
+                                "Try `jj --no-pager`. You can also try `TERM=xterm` or setting up \
+                                 a different pager."
+                            )
+                            .ok();
+                        }
+                    }
+                    // We do not cause a panic or another error so that the exit
+                    // code of the command is preserved.
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


cli/src/ui.rs
```diff
@@ -87,7 +88,32 @@ impl BuiltinPager {
...
+                                - Use `jj --no-pager`
+                                - Use a different terminal (e.g. Windows Terminal or the Command Prompt)
+                                - Configure a different pager, see https://martinvonz.github.io/jj/latest/windows/#pagination for Git Bash on Windows
+                                - Use `winpty jj ...` on Windows."#
+                                })
+                        } else {
+                            eprintln!("jj ERROR: built-in pager failed to start, will panic.")
+                        }
+                        eprintln!();
+                    })
```
yuja, line 115: nit: Alternatively, you can send the error back through the join handle, and print it at `.finalize()` where `ui` is available.

https://github.com/yuja/jj/commit/dfb5251a6a996c3e99040c13eea547069c9216d4#diff-f72142370d81ec8e718184bed135927f6fb90a77c5f7f0195a99c40f48a4ee6eR106-R113
(This is an experimental patch to use streampager. It works, but brings some old crates as dependency.)

In any case, we don't have to `.unwrap()` the error.

ilyagr: As I was complaining in https://github.com/martinvonz/jj/issues/4182#issuecomment-2259315065, the panic tends to happen early and abort any mutating operations. The only way I can think of easily preserving the error (which seems to match your suggestion), it will only be detected *after* the `jj` operation completes. So, jj could do some mutations, but all user output will be lost at that point. This seemed problematic to me, so while a panic is not great, it seems like a better alternative.

ilyagr: As for switching to streampager, that would hopefully solve this problem as well.

However, I am not sure about it in its currently abandoned state. Despite the Discord discussion, I think somebody should run `cargo update` on it, and I think I remember some bugs when I was testing it out (though I don't remember whether I was just really picky; at that point I was wondering if I should use it instead of `less`).

ilyagr: With https://github.com/martinvonz/jj/pull/4187, we could detect the error at the first `write` to the pager, but at that point, it's unclear to me how to best handle it and print a nice message to the user if the problem is with their terminal.

yuja: > the panic tends to happen early and abort any mutating operations.

Well, the end result is basically the same behavior as `ui.pager="false"`. Panicking in the pager thread doesn't stop the main thread, and we'll need to fix #4187 anyway so the main thread can notice the problem early.

And, it's pager. Mutable operations usually happen before the pager is spawned.

ilyagr: This contradicts a few of my assumptions/beliefs:

- I thought that the pager is initialized when the UI is, and I thought the UI is initialized in the very beginning of `jj` execution, before `cmd_whatever` is called.
- I thought a panic stops all threads. If not, we could replace it with something to stop all threads, perhaps?
- I thought that if `minus` fails, it becomes a black hole in which any data that has been written to it before or is written after is lost. Does it re-print its data store without paging? That would be nice.

Now, I'm not quite sure what's right or what's best; I think you are saying that at least the first two of my assumptions were wrong. If you can give me pointers to what's going on, I'd appreciate it, otherwise I'll look myself.

If we can get to a state where the behavior is the same as `ui.pager="false"`, with a warning, that would be ideal. Considering my last belief above, this seemed hard to achieve.

-------------

My main concern is that some sort of explanation is shown to people trying to use `jj` on Git Bash explaining to them that disabling the pager will cause things to start working.

yuja: > * I thought that the pager is initialized when the UI is,

No. Pager starts as late as possible (so that early argument errors wouldn't be paged for example.)

> * I thought a panic stops all threads. If not, we could replace it with something to stop all threads, perhaps?

No. I'm not sure if we can inject panic to other running threads, but maybe we could abort the entire process if that's what we want.

> * I thought that if `minus` fails, it becomes a black hole in which any data that has been written to it before or is written after is lost. Does it re-print its data store without paging? That would be nice.

Yes and no. All data written to `ui` would be lost. New data can be printed after `ui.finalize_pager()`.

ilyagr: Ok, I changed the whole thing to match the plan you suggested. I was in a mild rush, so there might be silly mistakes.

ilyagr:
(Changed the whole PR design after the discussion with Yuya)

cli/src/ui.rs
```diff
@@ -130,7 +131,35 @@ impl UiOutput {
                 }
             }
             UiOutput::BuiltinPaged { pager } => {
-                pager.finalize();
+                if let Err(minus_error) = pager.finalize() {
+                    writeln!(
+                        ui.error_with_heading("Error"),
```
yuja, line 136: nit: use `ui.warning_default()` (and `warning_no_heading()`) as it's not a fatal error

cli/src/ui.rs
```diff
@@ -130,7 +131,35 @@ impl UiOutput {
...
+                        "Built-in pager failed to start: {minus_error}",
+                    )
+                    .ok();
+                    writeln!(ui.error_no_heading(), "The output of this command is lost.").ok();
+                    if let MinusError::Setup(minus::error::SetupError::InvalidTerminal) =
+                        minus_error
```
yuja, line 142: nit: `matches!(..)` instead of `let`?

And maybe we can add `cfg!(windows)` because the added hint is very Windows specific.

cli/src/ui.rs
```diff
@@ -130,7 +131,35 @@ impl UiOutput {
...
+                    {
+                        writeln!(
+                            ui.hint_with_heading("Hint"),
```
yuja, line 145: nit: `hint_default()`

yuja (approved):
FYI, I've uploaded streampager PR. https://github.com/martinvonz/jj/pull/4203

cli/src/ui.rs
ilyagr: I might use two separate `warning_default()`-s, since they warn about different things.

cli/src/ui.rs
ilyagr: Well, the first three workarounds still would work if this error shows up not on Windows. I could write two different hints for Windows and non-Windows, but I'd rather do that if we hear about somebody seeing this error in a different situation.

cli/src/ui.rs
ilyagr: Also, thanks!

cli/src/ui.rs
yuja: I prefer a shorter hint in general, and I would add "Hint: Try `jj --no-pager`" or something on non-Windows if we need to show something.

The builtin pager isn't default on non-Windows, and it's less likely that a terminal emulator has compatibility issue on non-Windows. (If it had, a better suggestion would be to try `TERM=xterm`.)

---

Unrelated: maybe better to not shout "POSSIBLE WORKAROUNDS:"? I think "Possible workarounds:" is our style.

ilyagr: I simplified the hint a bit and set up a shorter hint for non-Windows. I'll probably merge this shortly, we can keep editing the hint after that as well.

PR #4202, opened 2024-08-02
cargo-deny: remove uses of unsupported config keys

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -9,24 +9,6 @@
 # The values provided in this template are the default values that will be used
 # when any section or field is not specified in your own configuration
 
-# If 1 or more target triples (and optionally, target_features) are specified,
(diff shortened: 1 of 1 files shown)


thoughtpolice:
This seems fine if it makes the build pass, but I wonder what the alternatives for some of these are now that they're gone. I looked at [the history of `deny.toml`](https://github.com/martinvonz/jj/commits/main/deny.toml) and it's very short; maybe we could just re-generate a default `deny.toml` (via `cargo deny init`) and see if it "fixes" this beyond just deleting everything? It might at least be insightful even if we apply this patch anyway.

ilyagr:
I tried to do a quick-and-dirty version of Austin's suggestion in #4204.

martinvonz:
Dropping in favor of #4204

PR #4203, opened 2024-08-03
switch builtin pager to streampager

yuja:
# Checklist

If applicable:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ rpassword = "7.3.1"
 rustix = { version = "0.38.43", features = ["fs"] }
 same-file = "1.0.6"
 sapling-renderdag = "0.1.0"
+sapling-streampager = "0.10.3"
 scm-record = "0.5.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.135"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ chrono = { version = "0.4.39", default-features = false, features = [
 chrono-english = { version = "0.1.7" }
 clru = "0.6.2"
 criterion = "0.5.1"
-crossterm = { version = "0.28", default-features = false }
+crossterm = { version = "0.28", default-features = false, features = ["windows"] }
 digest = "0.10.7"
 dirs = "5.0.1"
 dunce = "1.0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,7 +72,6 @@ insta = { version = "1.42.0", features = ["filters"] }
 itertools = "0.13.0"
 libc = { version = "0.2.169" }
 maplit = "1.0.2"
-minus = { version = "5.6.1", features = ["dynamic_output", "search"] }
 num_cpus = "1.16.0"
 once_cell = "1.20.2"
 os_pipe = "1.2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ maplit = "1.0.2"
 minus = { version = "5.6.1", features = ["dynamic_output", "search"] }
 num_cpus = "1.16.0"
 once_cell = "1.20.2"
+os_pipe = "1.2.1"
 pest = "2.7.15"
 pest_derive = "2.7.15"
 pollster = "0.3.0"
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -28,14 +28,13 @@ use std::process::Child;
 use std::process::ChildStdin;
 use std::process::Stdio;
 use std::str::FromStr;
+use std::thread;
 use std::thread::JoinHandle;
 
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -101,6 +101,7 @@ smallvec = { version = "1.13.2", features = [
     "const_new",
     "union",
 ] }
+streampager = "0.10.3"
```
yuja, line 104: Perhaps, "CC0-1.0" can be allowed, but I haven't updated the list yet.
https://github.com/martinvonz/jj/actions/runs/10224603116/job/28292547146?pr=4203#step:4:83

ilyagr:
I would really like a version of `streampager` with its dependencies updated for this. It's annoying to have to link an old version of `clap` (see the diff in Cargo.toml).

IIRC, streampager's interface seemed a bit annoying in parts by default, so we might want to change its default config (or try it out at least).

Apart from that, I think this might be great.

yuja:
> I would really like a version of `streampager` with its dependencies updated for this 

Totally agree, and the stripped-down version exists in Sapling tree. Asked question at https://github.com/markbt/streampager/issues/61

yuja:
Updated dependency to sapling-streampager.

thoughtpolice (approved):

cli/src/ui.rs
```diff
@@ -140,6 +81,26 @@ impl UiOutput {
         Ok(UiOutput::Paged { child, child_stdin })
     }
 
+    fn new_builtin_paged() -> streampager::Result<UiOutput> {
+        let mut pager = streampager::Pager::new_using_stdio()?;
+        // TODO: should we set the interface mode to be "less -FRX" like?
+        // It will override the user-configured values.
```
ilyagr, line 87: Could you elaborate on "It will override the user-configured values."? Can Streampager be configured by the user outside `jj`? 

I think we should default to the ["Hybrid" behavior](https://docs.rs/sapling-streampager/latest/streampager/config/enum.InterfaceMode.html#variant.Hybrid), at least if the user hasn't configured anything.

ilyagr:
Also, yay! 🎉

cli/src/ui.rs
yuja: > Streampager be configured by the user outside jj?

Yes. https://github.com/markbt/streampager?tab=readme-ov-file#configuration

And setting non-default value here means user config wouldn't be respected afaik. I don't have an opinion about the default behavior, but it should be configurable somehow.

ilyagr: > Yes. https://github.com/markbt/streampager?tab=readme-ov-file#configuration

Thank you!

> And setting non-default value here means user config wouldn't be respected afaik. 

Do you think it'd make sense to check whether Streampager's config file exists, and use our default if it doesn't?

ilyagr: Actually, I don't think `jj`'s streampager respects `$CONFIG_DIR/streampager/streampager.toml`. I have both `~/.config/streampager/streampager.toml` and `~/Library/Application Support/streampager/streampager.toml`, and it doesn't seem to be respected. Is it for you?

Sapling also seems to duplicate streampager config. I put an excerpt from `sl help config` in https://gist.github.com/ilyagr/8760765932c2b6f4b70ea87a0bf5c804 .

ilyagr: As a complete aside, the `pager.ignore` setting from that Gist is an interesting idea.

yuja: > Do you think it'd make sense to check whether Streampager's config file exists, and use our default if it doesn't?

Sounds like a bad idea. Maybe better to add our config section for the builtin pager.

> Actually, I don't think jj's streampager respects $CONFIG_DIR/streampager/streampager.toml

`strace` shows the file is read. Appears that `interface_mode` isn't parsed however, maybe because of serde or toml API change?

ilyagr: Yes, it's supposed to be read, and some environment variables are supposed to affect it as well.

https://docs.rs/crate/sapling-streampager/0.10.3/source/src/config.rs#201
https://docs.rs/crate/sapling-streampager/0.10.3/source/src/pager.rs#108

Clearly, it's buggy.

Unfortunately, Streampager does not seem to have a `new_with_config` function. I wonder whether we could disable all of that streampager native config by setting `pager.config` ourselves after `pager` is created. Then, we could have our own config for the builtin pager, just as you said.

This would mean we'd be uselessly reading that file. We could contribute some form of `new_with_config` to streampager if it makes sense for them and their API.

ilyagr: So, the best option might be to copy Sapling: set the three options we can in `jj` config, overriding streampager's defaults with our own. Without any config, Sapling defaults to the Hybrid mode, which is the default for the Sapling config setting, but not the default for streampager's.

ilyagr: I'll do that.

PR #4205, opened 2024-08-03
cli: make "op abandon" not fail with multiple op heads

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/operation/abandon.rs
+++ b/cli/src/commands/operation/abandon.rs
@@ -13,15 +13,14 @@
 // limitations under the License.
 
 use std::io::Write as _;
-use std::slice;
+use std::{iter, slice};
 
 use itertools::Itertools as _;
-use jj_lib::op_store::OperationId;
 use jj_lib::op_walk;
 use jj_lib::operation::Operation;
 
 use crate::cli_util::{short_operation_hash, CommandHelper};
-use crate::command_error::{cli_error, user_error, user_error_with_hint, CommandError};
+use crate::command_error::{cli_error, user_error, CommandError};
 use crate::ui::Ui;
 
 /// Abandon operation history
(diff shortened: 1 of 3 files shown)


cli/tests/test_operations.rs
```diff
@@ -508,6 +508,84 @@ fn test_op_abandon_without_updating_working_copy() {
...
+    "###);
+
+    // Can't abandon the other head operation.
+    let stderr = test_env.jj_cmd_failure(&repo_path, &["op", "abandon", other_head_op_id]);
+    insta::assert_snapshot!(stderr, @r###"
+    Error: Cannot abandon the current operation 603773cdd351
+    Hint: Run `jj undo` to revert the current operation, then use `jj op abandon`
+    "###);
+
+    // Can abandon the operation in one branch, and the other head should be
```
martinvonz, line 556: What do you mean by this? It looks like we abandon the operation *before* the branch point. And what does "the other head should be unchanged" mean?

cli/src/commands/operation/abandon.rs
```diff
@@ -82,25 +81,36 @@ pub fn cmd_op_abandon(
...
+        .iter()
+        .find(|op| current_head_ops.contains(op))
+    {
         return Err(user_error_with_hint(
-            "Cannot abandon the current operation",
+            format!(
+                "Cannot abandon the current operation {}",
+                short_operation_hash(op.id())
+            ),
             "Run `jj undo` to revert the current operation, then use `jj op abandon`",
```
martinvonz, line 96: I suppose this isn't entirely valid when there are multiple op heads, right? I think it would attempt to undo the merge operation, which I think fails. Should we just not show the hint if there are multiple op heads?

cli/tests/test_operations.rs
yuja: Suppose the following graph, `prev_op` is abandoned and `head_op` is reparented. `other_head_op` is also reparented (to the same base operation), but that should be noop.
```
  o other_head_op
  |
o | head_op
| |
o | prev_op
|/
o
```

Perhaps, ", and the other head should be unchanged." can be just removed. This is the test for "would crash if we attempted to remap the unchanged op."

cli/src/commands/operation/abandon.rs
yuja: > Should we just not show the hint if there are multiple op heads?

Make sense.

cli/tests/test_operations.rs
yuja: Updated to
```
    // Can abandon the operation which is not an ancestor of the other head.
    // This would crash if we attempted to remap the unchanged op in the op
    // heads store.
```

martinvonz (approved):

PR #4206, opened 2024-08-04
ui: include error sources in pager warning

yuja:
Extracted from #4203

# Checklist

If applicable:


ilyagr (approved):
I guess you are doing this for parity with other errors `jj` prints? Or do you know a case where it makes a difference?

In any case, it can't hurt.

yuja:
> I guess you are doing this for parity with other errors `jj` prints? Or do you know a case where it makes a difference?

`MinusError` is now printed as "Failed to initialize the terminal: The standard output is not a valid terminal", etc. It's not super important here, but sometimes the outermost error message is useless.

PR #4209, opened 2024-08-04
ui: do not write() to channel if builtin pager has terminated

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -58,6 +58,11 @@ impl Write for &BuiltinPager {
     }
 
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        // Since minus::Pager holds the rx end internally, push_str() doesn't
+        // fail even after the paging thread gets terminated.
+        if self.dynamic_pager_thread.is_finished() {
+            return Err(io::ErrorKind::BrokenPipe.into());
+        }
         let string = std::str::from_utf8(buf).map_err(io::Error::other)?;
         self.pager.push_str(string).map_err(io::Error::other)?;
         Ok(buf.len())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Windows binaries no longer require `vcruntime140.dll` to be installed
   (normally through Visual Studio.)
 
+* On quit, the builtin pager no longer waits for all outputs to be discarded.
+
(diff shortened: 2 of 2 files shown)


cli/src/ui.rs
```diff
@@ -57,6 +57,11 @@ impl Write for &BuiltinPager {
     }
 
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        // Since minus::Pager holds the rx end internally, push_str() doesn't
```
ilyagr, line 60: Do you want to add a quick note to the changelog?

ilyagr (commented):
Nice! Seems to me like this should work. @ErichDonGubler , WDYT?

In principle, an approach closer to the one from #4187 would give us more control and maybe allow us not to lose output in #4197. I'm not sure it's worth it, especially if we might jump over to `streampager` anyway.

ilyagr (approved):
I'm hoping a fix to this can go in before the release, but let's wait a day before merging to give people time to look at this.

cli/src/ui.rs
```diff
@@ -57,6 +57,11 @@ impl Write for &BuiltinPager {
...
+        // fail even after the paging thread gets terminated.
+        if self.dynamic_pager_thread.is_finished() {
+            return Err(io::Error::new(io::ErrorKind::BrokenPipe, "pager quit"));
```
ErichDonGubler, line 63: suggestion: Unless you want the error message here (I don't think it's used at any point?), you can avoid an allocation for the error message by using `io::Error::from(io::ErrorKind::BrokenPipe)`.

ErichDonGubler (approved):
Ahh, I like this much better than <https://github.com/martinvonz/jj/pull/4187>. I believe it's more robust, even. Cheers!

cli/src/ui.rs
```diff
@@ -57,6 +57,11 @@ impl Write for &BuiltinPager {
     }
 
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        // Since minus::Pager holds the rx end internally, push_str() doesn't
+        // fail even after the paging thread gets terminated.
+        if self.dynamic_pager_thread.is_finished() {
```
ErichDonGubler, line 62: praise: I'm feeling silly that I had completely forgotten about `JoinHandle::is_finished`, and hadn't considered this approach in https://github.com/martinvonz/jj/pull/4187. 😆😭

cli/src/ui.rs
yuja: Allocation isn't important, but I like it. Switched to from/into.

PR #4212, opened 2024-08-05
tutorial: link to Steve Klabnik's tutorial, mention `jj log -r ::`

ilyagr:
I was thinking of updating the tutorial in #4168, but conversations like [this one on
Discord](https://discord.com/channels/968932220549103686/968932220549103689/1269803837070377076) suggest that many people aren't reading it anyway.

So, this is an attempt to quickly patch up the most relevant things before the upcoming release.

Let me know whether you think this is a good idea or needs more editing.

Cc @steveklabnik

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -1,5 +1,9 @@
 # Tutorial
 
+>  **Hint:** This tutorial has become somewhat out of date. Many people find
+>   the alternative (not quite finished) [tutorial by Steve
+>   Klabnik](https://steveklabnik.github.io/jujutsu-tutorial/) helpful. 
+
 This text assumes that the reader is familiar with Git.
 
 ## Preparation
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -150,15 +156,14 @@ commit ID, which changes when you rewrite the commit. You can give either ID
 to commands that take revisions as arguments. We will generally prefer change
 IDs because they stay the same when the commit is rewritten.
 
-By default, `jj log` lists your local commits, with some remote commits added
-for context.  The `~` indicates that the commit has parents that are not
```
martinvonz, line 154: The "with some remote commits added for context" was meant to be about the `ancestors(x, 2)` part (`orrkosyo` above), so that's still true

docs/tutorial.md
```diff
@@ -1,5 +1,11 @@
 # Tutorial
 
+!!! note
```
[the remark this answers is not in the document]
ilyagr: OK, how about this? For comparison, the new version looks like this in MkDocs.

image

The old version looked like this:

image

I also found https://oprypin.github.io/markdown-callouts/ that would make the GitHub version look like:

------

HINT: something something

------

and might allow

> HINT: something something.

docs/tutorial.md
ilyagr: OK, reverted. (I thought it referred to remote-tracking branches, as you probably guessed)

martinvonz (approved):

steveklabnik:
Thank you!

Some stuff in my personal life has been shifting lately, and I'm suspecting I'm actually going to have the time and energy to finish this off. I know I've been saying it, maybe it will happen! And we can talk about actually integrating it upstream, finally.

PR #4226, opened 2024-08-06
jj run: describe it as a stub in the help text

ilyagr:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/run.rs
+++ b/cli/src/commands/run.rs
@@ -20,7 +20,7 @@ use crate::cli_util::{CommandHelper, RevisionArg};
 use crate::command_error::{user_error, CommandError};
 use crate::ui::Ui;
 
-/// Run a command across a set of revisions.
+/// (**Stub**, does not work yet) Run a command across a set of revisions.
 ///
 ///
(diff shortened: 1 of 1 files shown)


thoughtpolice (approved):

PhilipMetzger:
I dont think this makes a difference, since its hidden anyway.

mhammond:
FWIW, I recently struck this. A colleague wished `jj run` wasn't a stub so they could use it. I thought that it was enabled by default in 0.19, so run `jj help run` and saw no evidence it was a stub, so started a conversation about "maybe you need to upgrade" etc. Joke was on me :)

ilyagr:
Yeah, I wasn't sure this was necessary either, but I followed the logic of "well, it confused me, and I can fix it".

PhilipMetzger:
If you want `jj run` you should just follow #4021, so it finally gets a implementation. My in-tree implementation is struggling no disagreement there though.

PhilipMetzger:
I would've accepted it anyway since @mhammond opinion is valid and since we've seen people run into it already but since it was past midnight for me I was not the reviewer.

Issue #4222, opened 2024-08-05
Weird corner-case issues with .gitignore in git init'ed repos


juanmeleiro:
## Description

So, for context: I keep my `~/.config` on version control. It started on `git`. Just for aesthetic purposes, my `.gitignore` file for this repo is actually `~/.config/gitignore` (note the lack of a dot). This was done using the following command:

```sh
git config core.excludesFile gitignore
```

All was well and good. But also, I didn't like to do `cd .config` every time, so I aliased as `config` this command:

```sh
git --work-tree=$XDG_CONFIG_HOME --git-dir=$XDG_CONFIG_HOME/.git
```

All was well and good. And then, I discovered `jj` and, evidently, made the switch (`jj git init --colocate` and all that jazz). I changed the alias:

```sh
jj --repository $XDG_CONFIG_HOME
```

And all was well and good, *until* I did `config status`, when all hell broke loose. All those large files those anoying applications keep in my beautiful config directory started resurfacing from the depths of the `gitignore`, `jj` trying to index them all just to fail because of some megabyte-sized monster. Oh the horrors.

But more than that, I was confused. Weren't they ignored? And `jj` knows it, as the following works quite well, thank you:

```sh
$ cd .config
$ jj status
```

## Steps to Reproduce the Problem

Go to some directory in your system and run the following:

```console
$ mkdir testrepo
$ cd testrepo
$ git config core.excludesFile gitignore
$ jj git init --colocate
$ echo a > gitignore
$ touch a
$ jj status
Working copy changes:
A gitignore
Working copy : ntsqttkw da1a1604 (no description set)
Parent commit: zzzzzzzz 00000000 (empty) (no description set)
$ cd ..
$ jj --repository testrepo/
Working copy changes:
A testrepo/a
A testrepo/gitignore
Working copy : ntsqttkw 2c759af3 (no description set)
Parent commit: zzzzzzzz 00000000 (empty) (no description set)
```

## Expected Behavior

The two outputs of the `status` subcommand should be the same. The same files should be ignored.

## Actual Behavior

They aren't.

## Specifications

```console
$ uname -mrsv
Linux 6.6.42-1-lts #2-Alpine SMP PREEMPT_DYNAMIC 2024-07-25 15:20:22 x86_64
```

## Comments

I'm not sure what the fix should be, or even if it should be fixed. Anyway, I'd love suggestions to deal with the situation. Specifically, I would like to be able to

1. Keep the file as `gitignore` (without the dot)
[...]

PR #4227, opened 2024-08-07
git: resolve relative core.excludesFile path at workspace root

yuja:
#4222

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -800,17 +800,20 @@ impl WorkspaceCommandHelper {
 
     #[instrument(skip_all)]
     pub fn base_ignores(&self) -> Result<Arc<GitIgnoreFile>, GitIgnoreError> {
-        fn get_excludes_file_path(config: &gix::config::File) -> Option<PathBuf> {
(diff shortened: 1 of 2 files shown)


cli/src/cli_util.rs
```diff
@@ -821,16 +821,25 @@ impl WorkspaceCommandHelper {
             std::env::var("HOME").map(|x| Path::new(&x).join(".config"))
         }
 
+        let resolve_excludes_file_path = |config: &gix::config::File| {
+            let path = get_excludes_file_path(config)?;
+            // The configured path is usually absolute, but if it's relative,
+            // the "git" command would read the file at the work-tree directory.
+            Some(self.workspace_root().join(path))
+        };
```
martinvonz, line 829: Why not bake this into `get_excludes_file_path()`? And should it only be done in the `core.excludesFile` case? Maybe it doesn't matter, assuming `xdg_config_home()` is always absolute.

yuja: I assumed `get_excludes_file_path()` could be replaced with some gix function, but apparently not. Inlined it.

martinvonz (approved):

PR #4229, opened 2024-08-07
merged_tree: remove .diff() method in favor of .diff_stream()

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/default_index/revset_engine.rs
+++ b/lib/src/default_index/revset_engine.rs
@@ -1144,7 +1144,15 @@ fn has_diff_from_parent(
     }
     let from_tree = rewrite::merge_commit_trees_without_repo(store, &index, &parents)?;
     let to_tree = commit.tree()?;
-    Ok(from_tree.diff(&to_tree, matcher).next().is_some())
+    let mut tree_diff = from_tree.diff_stream(&to_tree, matcher);
+    async {
+        match tree_diff.next().await {
+            Some((_, Ok(_))) => Ok(true),
+            Some((_, Err(err))) => Err(err),
+            None => Ok(false),
+        }
+    }
+    .block_on()
 }
 
 fn matches_diff_from_parent(
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -308,19 +308,10 @@ impl MergedTree {
         TreeEntriesIterator::new(self.clone(), matcher)
     }
 
-    /// Iterate over the differences between this tree and another tree.
+    /// Stream of the differences between this tree and another tree.
     ///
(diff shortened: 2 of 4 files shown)


lib/src/default_index/revset_engine.rs
```diff
@@ -1144,7 +1144,15 @@ fn has_diff_from_parent(
...
-    Ok(from_tree.diff(&to_tree, matcher).next().is_some())
+    let mut tree_diff = from_tree.diff_stream(&to_tree, matcher);
+    async {
+        match tree_diff.next().await {
+            Some((_, Ok(_))) => Ok(true),
+            Some((_, Err(err))) => Err(err),
+            None => Ok(false),
+        }
+    }
+    .block_on()
```
martinvonz, line 1155: For later: I'm not sure if this code even needs to compare the trees. Most backends are probably going to use content-based tree ids, so we could simply compare the tree ids. However, I'm not sure we should require backends to do that.

Also, for now, we have the legacy trees to worry about. I was actually planning to send a PR to promote legacy trees to modern trees on read. I've just been waiting to do that early in the next release (i.e. today).

martinvonz (approved):
Good idea! This removes the risk of calling the slower `diff()`. I had somehow not considered that myself (probably incorrectly thinking that we need to keep `TreeDiffIterator` so we need to keep `diff()` too).

lib/src/default_index/revset_engine.rs
yuja: It's fallback path for merges and filtered matches. Maybe we can add fast path for prefix matches, but at least a simple `file(.)` query won't trigger tree diffing.

martinvonz: Good point about filtered matches. But for merges, I would think we should be able to compare the tree id of the result of `merge_commit_trees_without_repo()` with the commit's own tree id (ignoring the other potential issues I mentioned), no?

yuja: > But for merges (without filtering), I would think we should be able to compare the tree id of the result of `merge_commit_trees_without_repo()` with the commit's own tree id 

Correct. I assumed the diff iterator/stream has fast path for `self.id == other.id`, but apparently it doesn't, and even if it had, `self.id != other.id` can't be short-circuited during diffing. That said, we might want more generic fast path for `.`, `prefix/dir`, `exact/file` queries.

yuja: I'll address this separately.

martinvonz: Oh, definitely. I didn't mean to say that we need to change anything at all here.

PR #4235, opened 2024-08-08
copy-tracking: update design doc

fowles:


PhilipMetzger:
cc @torquestomp

martinvonz:
I think the general opinion in #4118 was to that this design won't work for conflicts and merges (and I'm not sure the old design will either). @yuja argued that we should just let design doc remain as is rather than documenting what's implemented since we're pretty confident the new API is not sufficient. So I would suggest closing this PR and then I or someone else will update the docs once we've figured something out for conflicts and merges.

PR #4241, opened 2024-08-08
merged_tree: do not re-look up non-conflicting tree values by name

yuja:
While measuring file(path) query, I noticed BTreeMap lookup appears in perf. It actually has a measurable cost if the history is linear and parent trees don't have to be merged dynamically. For merge-heavy history, the cost of tree merges is more significant. I'll address that separately.

```
% hyperfine --sort command --warmup 3 --runs 50 -L bin jj-1,jj-2 \
  'target/release-with-debug/{bin} -R ~/mirrors/git --ignore-working-copy \
   log -r "::trunk() & ~merges() & file(root:builtin)" --no-graph -n100'
Benchmark 1: target/release-with-debug/jj-1 ..
  Time (mean ± σ):     239.7 ms ±   7.1 ms    [User: 192.1 ms, System: 46.5 ms]
  Range (min … max):   222.2 ms … 249.7 ms    50 runs

Benchmark 2: target/release-with-debug/jj-2 ..
  Time (mean ± σ):     201.7 ms ±   6.9 ms    [User: 153.7 ms, System: 46.6 ms]
  Range (min … max):   184.2 ms … 211.1 ms    50 runs

Relative speed comparison
        1.19 ±  0.05  target/release-with-debug/jj-1 ..
        1.00          target/release-with-debug/jj-2 ..
```

# Checklist

If applicable:

### DIFF
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -22,10 +22,11 @@ use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::{iter, vec};
 
+use either::Either;
 use futures::future::BoxFuture;
 use futures::stream::{BoxStream, StreamExt};
 use futures::{Stream, TryStreamExt};
-use itertools::Itertools;
+use itertools::{EitherOrBoth, Itertools};
 
 use crate::backend;
 use crate::backend::{BackendResult, MergedTreeId, TreeId, TreeValue};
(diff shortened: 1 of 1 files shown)


lib/src/merged_tree.rs
```diff
@@ -333,28 +334,36 @@ fn all_tree_basenames(trees: &Merge<Tree>) -> impl Iterator<Item = &RepoPathComp
...
-    ) -> Box<dyn Iterator<Item = &'a RepoPathComponent> + 'a> {
-        Box::new(iter1.merge(iter2).dedup())
+fn all_tree_entries(
+    trees: &Merge<Tree>,
+) -> impl Iterator<Item = (&RepoPathComponent, MergedTreeVal<'_>)> {
+    if let Some(tree) = trees.as_resolved() {
+        let iter = tree
+            .entries_non_recursive()
+            .map(|entry| (entry.name(), MergedTreeVal::Resolved(Some(entry.value()))));
+        Either::Left(iter)
```
martinvonz, line 344: What does this line do?

yuja: It's cheaper version of `Box<dyn ..>`, but I haven't measured which one is fast.

martinvonz: Interesting trick. I had not seen it before.

martinvonz (approved):

PR #4243, opened 2024-08-09
github: build on macos-13 for x86_64

thoughtpolice:
We all noticed that x86 macOS binaries are no longer being provided on release, due to `macos-11` runners going the way of the Dodo a while back. Nobody alterted us to this, funny enough.

After some quick discussion, we concluded some things:

- x86 macOS runners are likely oversubscribed, and hurt CI latency badly
- `macos-12` is also deprecated; `macos-13` is the best x86 runner available
- GitHub probably isn't going to expand macOS runner capacity; `macos-13` will one day go away
- Some people are still using `jj` on Intel Macs. We didn't get alerted because they do their own builds for now, but may not always do that.
- We can just try to build on `macos-13` and make it optional for merges.

So that's what this does. It might be mergeable outright, but we can also use it to measure build latency impacts.

# Checklist

If applicable:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-14, windows-latest]
+        # macos-13 is x86; macos-14 is ARM
+        os: [ubuntu-latest, macos-13, macos-14, windows-latest]
         cargo_flags: [""]
         include:
         - os: ubuntu-latest
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
 
       matrix:
-        build: [linux-x86_64-musl, linux-x86_64-gnu, linux-aarch64-musl, linux-aarch64-gnu, macos-aarch64, win-x86_64]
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -15,7 +15,8 @@ jobs:
     strategy:
```
ilyagr, line 15: Perhaps add a commit updating the runner in the release job?

ilyagr:
Thank you Austin!

It looks like the macos-13 runners are not marked as required and ran in 6 minutes. I think we can just merge this?

.github/workflows/build.yml
ilyagr: (Separate PR is also fine if you prefer)

ilyagr (approved):

ilyagr:
> But we probably want to mark them as required once this is in, right?

I'm not sure. Austin was worried that these tests might be slow to get runners. If that's the case, it might make sense to keep them optional.

For me, the important thing is that they run regularly so that we eventually become aware if they stop working. For example, we'll now know when GitHub removes these runners before a release.

Cretezy (approved):

thoughtpolice:
Let's just go ahead and move forward with adding `macos-13` everywhere and see how it goes. I'll remind Martin to update the "required" list in a week or two if nothing goes wrong.

PR #4245, opened 2024-08-09
copy-tracking: improve --summary and add --stat

fowles:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### New features
 
 * The following diff formats now include information about copies and moves:
-  `--color-words`, `--summary`
+  `--color-words`, `--stat`, `--summary`
 
 * A tilde (`~`) at the start of the path will now be expanded to the user's home
   directory when configuring a `signing.key` for SSH commit signing.
--- a/cli/src/commit_templater.rs
+++ b/cli/src/commit_templater.rs
@@ -1394,14 +1394,18 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
             let path_converter = language.path_converter;
             let template = self_property
                 .map(move |diff| {
+                    // TODO: don't pass separate copies of from_tree/to_tree/matcher
+                    let from_tree = diff.from_tree.clone();
                     let to_tree = diff.to_tree.clone();
-                    diff.into_formatted(move |formatter, _store, tree_diff| {
+                    let matcher = diff.matcher.clone();
+                    diff.into_formatted(move |formatter, _store, _tree_diff| {
                         diff_util::show_diff_summary(
                             formatter,
-                            tree_diff,
                             path_converter,
-                            &Default::default(),
+                            &from_tree,
                             &to_tree,
+                            matcher.as_ref(),
+                            &Default::default(),
                         )
                     })
                 })
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -280,12 +280,17 @@ impl<'a> DiffRenderer<'a> {
         for format in &self.formats {
             match format {
                 DiffFormat::Summary => {
-                    let tree_diff = from_tree.diff_stream(to_tree, matcher, copy_records);
-                    show_diff_summary(formatter, tree_diff, path_converter, copy_records, to_tree)?;
+                    show_diff_summary(
+                        formatter,
+                        path_converter,
+                        from_tree,
+                        to_tree,
+                        matcher,
+                        copy_records,
+                    )?;
                 }
                 DiffFormat::Stat => {
-                    let no_copy_tracking = Default::default();
-                    let tree_diff = from_tree.diff_stream(to_tree, matcher, &no_copy_tracking);
+                    let tree_diff = from_tree.diff_stream(to_tree, matcher, copy_records);
                     show_diff_stat(formatter, store, tree_diff, path_converter, width)?;
                 }
                 DiffFormat::Types => {
@@ -1146,18 +1151,26 @@ pub fn show_git_diff(
     .block_on()
 }
 
-// TODO rework this signature to pass both from_tree and to_tree explicitly
 #[instrument(skip_all)]
 pub fn show_diff_summary(
     formatter: &mut dyn Formatter,
-    mut tree_diff: TreeDiffStream,
     path_converter: &RepoPathUiConverter,
-    copy_records: &CopyRecords,
+    from_tree: &MergedTree,
     to_tree: &MergedTree,
+    matcher: &dyn Matcher,
+    copy_records: &CopyRecords,
 ) -> io::Result<()> {
+    let mut tree_diff = from_tree.diff_stream(to_tree, matcher, copy_records);
+
     let copied_sources: HashSet<&RepoPath> = copy_records
         .iter()
-        .map(|record| record.source.as_ref())
+        .filter_map(|record| {
+            if matcher.matches(&record.target) {
+                Some(record.source.as_ref())
+            } else {
+                None
+            }
+        })
         .collect();
 
     async {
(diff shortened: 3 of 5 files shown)


ilyagr:
How is this related to #4118 (and is it)? I haven't been following that discussion as closely as I would have liked, so I'm wondering whether to look there or here first when I finally get to trying to understand what you are doing (no need to wait for me though).

fowles:
> How is this related to #4118 (and is it)? I haven't been following that discussion as closely as I would have liked, so I'm wondering whether to look there or here first when I finally get to trying to understand how copy-tracking works (no need to wait for me though).

it builds on it.  You only need to look at the last commit on this one.

CHANGELOG.md
```diff
@@ -14,8 +14,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+<<<<<<< Conflict 1 of 1
++++++++ Contents of side #1
```
PhilipMetzger, line 18: Nit: Remove the partially commited conflicted resolution.

cli/src/text_util.rs
```diff
@@ -35,6 +37,33 @@ pub fn split_email(email: &str) -> (&str, Option<&str>) {
...
+        .last()
+        .map(|((s, _), (t, _))| (s, t))
+        .unwrap_or((0, 0));
+    let suffix = UnicodeSegmentation::split_word_bound_indices(source)
+        .rev()
+        .zip(UnicodeSegmentation::split_word_bound_indices(target).rev())
+        .take_while(|((_, s), (_, t))| s == t)
+        .last()
+        .map(|((s, _), (t, _))| (s, t))
+        .unwrap_or((source.len(), target.len()));
```
PhilipMetzger, line 56: Nit: the whole block needs readable names, iiuc `s` => `source`, `t` => `target`

cli/src/text_util.rs
```diff
@@ -35,6 +37,33 @@ pub fn split_email(email: &str) -> (&str, Option<&str>) {
     }
 }
 
+pub fn render_copied_path(source: &str, target: &str) -> String {
+    if source == target {
+        return target.into();
+    }
```
PhilipMetzger, line 43: Nit: move this into the caller for now.

PhilipMetzger (changes requested):
This is great, there are just some minor nits.

cli/src/text_util.rs
fowles: done

cli/src/text_util.rs
fowles: I don't like that.  As it stands you can always pass strings to this and it will generate the right thing.  If you don't have this check, it will do something weird for the of equal strings.

CHANGELOG.md
fowles: done

cli/src/text_util.rs
```diff
@@ -629,4 +658,30 @@ mod tests {
...
+            render_copied_path("one/two/three", "one/three"),
+            "one/{two => }/three"
+        );
+        assert_eq!(
+            render_copied_path("one/two", "one/four"),
+            "one/{two => four}"
+        );
+        assert_eq!(render_copied_path("two", "four"), "{two => four}");
+        assert_eq!(render_copied_path("file1", "file2"), "{file1 => file2}");
+    }
```
PhilipMetzger, line 686: Nit: Missing test for the equal path handling, if you want to keep it.

cli/src/diff_util.rs
```diff
@@ -1244,21 +1254,28 @@ pub fn show_diff_stat(
     display_width: usize,
 ) -> Result<(), DiffRenderError> {
     let mut stats: Vec<DiffStat> = vec![];
+    let mut unresolved_renames = HashSet::<String>::new();
```
PhilipMetzger, line 1257: This is a stylistic nit, this can just be `HashSet::new` as Rust will infer it on line 1276.

cli/src/text_util.rs
PhilipMetzger: Basically the caller can handle the no-op or it should be documented somewhere, as it currently is a bit magic.

PhilipMetzger (approved):
LG, maybe Yuya or Martin should take a look too

cli/src/text_util.rs
```diff
@@ -35,6 +37,33 @@ pub fn split_email(email: &str) -> (&str, Option<&str>) {
...
+        .last()
+        .map(|((source_index, _), (target_index, _))| (source_index, target_index))
+        .unwrap_or((0, 0));
+    let suffix = UnicodeSegmentation::split_word_bound_indices(source)
+        .rev()
+        .zip(UnicodeSegmentation::split_word_bound_indices(target).rev())
+        .take_while(|((_, source_word), (_, target_word))| source_word == target_word)
+        .last()
+        .map(|((source_index, _), (target_index, _))| (source_index, target_index))
+        .unwrap_or((source.len(), target.len()));
```
yuja, line 56: Is it intentional that `file-1 => file-2` is rendered as `file-{1 => 2}`? I personally find it's hard to follow, and would rather want to split by path separator, not at word boundary.

BTW, a generic implementation of this function would be something like
```rust
Diff::for_tokenizer([source, target], find_path_component_ranges)
    .map(|hunk| match hunk {
        DiffHunk::Matching(common) => common,
        DiffHunk::Different([source, target]) => "{source} => {target}", // XXX handle trailing sep?
     })
     .collect()
```

cli/src/commit_templater.rs
```diff
@@ -1394,14 +1394,17 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
             let path_converter = language.path_converter;
             let template = self_property
                 .map(move |diff| {
+                    let from_tree = diff.from_tree.clone();
                     let to_tree = diff.to_tree.clone();
-                    diff.into_formatted(move |formatter, _store, tree_diff| {
+                    let matcher = diff.matcher.clone();
+                    diff.into_formatted(move |formatter, _store, _tree_diff| {
```
yuja, line 1400: nit: can you leave a TODO comment?
`// TODO: don't pass separate copies of from_tree/to_tree/matcher`

cli/src/text_util.rs
fowles: added docs, there was already a test for this behavior.

cli/src/diff_util.rs
fowles: done

cli/src/commit_templater.rs
fowles: Done

cli/src/text_util.rs
fowles: That is a good point.  I have moved this into `repo_path` so it sits with the other format function for changing paths into UI strings.  I have also update the implementation to be component based (which means we no long have a dep on unicode).

lib/src/repo_path.rs
```diff
@@ -502,6 +503,70 @@ impl RepoPathUiConverter {
         }
     }
 
+    /// Format a copy from `source` to `target` for display in the UI by
+    /// extracting common components and producing something like
+    /// "common/prefix/{source => target}/common/suffix".
+    ///
+    /// If `source == target`, returns `format_file_path(source)`.
```
martinvonz, line 510: I'm not sure this belongs in the lib crate. It feels a bit specific to a CLI. It seems likely that e.g. a GUI or Web UI would want to display it differently. Actually, even the CLI may want to style the renamed part differently from the rest of the string. So I feel like this should be made more generic or be moved into CLI. It could be made more generic by returning the `(prefix, (before, after), suffix)` or something like that. Maybe a proper struct or enum is better (an enum could have a `Unchanged(String)` variant, but it could instead be a struct wrapped in an `Option`). What do you think?

fowles: do you have a suggestion for where to move it?  One of the reasons it fits here in my mind is that `format_file_path` is also doing a bunch of UI specific work and they do share behavior that core part of things.

martinvonz: I think `format_file_path()` is generic enough that it makes sense in a web UI too, for example, but a web UI would probably not want to display the `common/prefix/{source => target}/common/suffix` form. Maybe `diff_util.rs` is a good place since this way of displaying paths is specific to diffs.

martinvonz: Oh, but maybe the use of `RepoPathUiConverter` means that it should be here. Let me take a look at the code to see how this works.

lib/src/repo_path.rs
```diff
@@ -858,4 +923,29 @@ mod tests {
...
+                .replace("\\", "/")
+        };
+
+        assert_eq!(format("one/two/three", "one/two/three"), "one/two/three");
+        assert_eq!(format("two/three", "four/three"), "{two => four}/three");
+        assert_eq!(
+            format("one/two/three", "one/four/three"),
+            "one/{two => four}/three"
+        );
+        assert_eq!(format("one/two/three", "one/three"), "one/{two => }/three");
```
martinvonz, line 945: Could you add tests like the following too?
```
        assert_eq!(format("one/two", "one/two/three"), "one/{two => two/three}");
        assert_eq!(format("one/two", "zero/one/two"), "{one/ => zero/one/}two");
        assert_eq!(format("one/two/three", "one/two"), "one/{two/three => three}");
        assert_eq!(format("zero/one/two", "one/two"), "{zero/one/ => one/}two");
        assert_eq!(format("one/two", "one/two/three/one/two"), "one/{two => two/three/one/two}");
```

I think that's how I would expect the result to look but I'm not sure. You can check what git does if you feel like it.

lib/src/repo_path.rs
martinvonz: I'm still not sure where this code should live but we can always move it later. Let's leave it here for now unless someone else has a good idea for how to structure it. Sorry about my indecision.

lib/src/repo_path.rs
fowles: done

PR #4247, opened 2024-08-09
merge_tools.toml: add VSCodium as a merge tool

NyCodeGHG:
[VSCodium](https://vscodium.com/) is a free/libre distribution of Microsoft's Visual Studio Code editor, it's functionally more or less the same, but distributed under a FOSS license, unlike VS Code.

This adds VSCodium as a merge tool.

# Checklist

If applicable:
  I'm not sure if this is worth a changelog entry, it's not really a "new feature", is it?

### DIFF
--- a/cli/src/config/merge_tools.toml
+++ b/cli/src/config/merge_tools.toml
@@ -44,6 +44,7 @@ merge-tool-edits-conflict-markers = true
 # https://gist.github.com/ilyagr/5d6339fb7dac5e7ab06fe1561ec62d45
 edit-args = ["-f", "-d", "$left", "$right"]
 
+# if you change the settings for vscode, please do the same for vscodium
 [merge-tools.vscode]
 program = "code"
 merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
@@ -51,3 +52,10 @@ merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
 # markers. Unfortunately, it does not seem to be able to output conflict markers when
 # the user only resolves some of the conflicts.
 merge-tool-edits-conflict-markers = true
+
+# free/libre distribution of vscode, functionally more or less the same
+[merge-tools.vscodium]
+program = "codium"
+merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
+merge-tool-edits-conflict-markers = true
+
--- a/docs/config.md
+++ b/docs/config.md
@@ -621,12 +621,12 @@ by `jj resolve`. For example:
 
 ```toml
 # Use merge-tools.meld.merge-args
-ui.merge-editor = "meld"  # Or "vscode" or "kdiff3" or "vimdiff"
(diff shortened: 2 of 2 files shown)


ilyagr:
Thank you for the PR!

I am not sure about this. On one hand, this might be a pragmatic solution. On the other, I wonder if we should have a better way to slightly modify existing merge tools. (Whether or not we merge this, its a great reminder of the issue regardless).

For background, currently, the options are setting `merge-tools.vscode.program = "vscoduim"` or the user copying the config as this PR does. Perhaps `jj config` can help automate the copying? Of course, that wouldn't keep the new config up to date.

Another aside: a similar issue occurs is the vimdiff config. Some people prefer Neovim. I have a copy of the config for `gvim` (which on my machine launches Macvim). There are more possible variations.

NyCodeGHG:
I see your point that this is kind of redundant, but it would be a way better user experience imo if it would just work with VS Code as well as VS Codium. Not so sure about this.

martinvonz:
I think this PR makes sense if it's just a few lines that are duplicated in the config (looks like it), and if VSCodium is a somewhat popular tool ([looks like it](https://github.com/VSCodium/vscodium)). The duplication is small enough that I don't even see much need to add a feature that allows better reuse, but if @ilyagr or others do, it seems like we can think about that feature later?

ilyagr:
I'm not strongly against it, especially if other people feel this is worthwhile.

There are a couple reasons for my uncertainty, but indeed this PR won't make a huge difference to them. In my mind, if we do this PR, we should probably also do `gvim` and `neovim` at least, and there will be other variations. If we have a lot of copying in our configs, it'll become more annoying to maintain. This makes me slightly worried about the longer term; in my mind, the "feature for better reuse" is a bit unclear and not imminent.

cli/src/config/merge_tools.toml
```diff
@@ -51,3 +51,9 @@ merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
 # markers. Unfortunately, it does not seem to be able to output conflict markers when
 # the user only resolves some of the conflicts.
 merge-tool-edits-conflict-markers = true
+
+[merge-tools.vscodium]
```
yuja, line 55: nit: I would add some comment so we'll copy new settings from vscode to vscodium.
```
# free/libre distribution of vscode, functionally is more or less the same
```

yuja (commented):
> if we do this PR, we should probably also do `gvim` and `neovim` at least, and there will be other variations.

I think copying a few lines of settings is okay. We'll need it for `gvim` if we add `gui = true` flag. We can add some abstraction to copy/inherit config values, but I'm not sure if that's better.

ilyagr (approved):
OK, the consensus is clear. Thank you again for the PR! PTAL Yuya's suggestion, though.

cli/src/config/merge_tools.toml
NyCodeGHG: good idea, i added a comment

ilyagr:
Thank you!

Issue #4152, opened 2024-07-25
Consider not partially resolving merged trees


martinvonz:
When we merge trees and we run into conflicts, we still resolve other parts of the tree and update all the input trees with the resolves parts. The reasoning was that it saves us from redoing those merges later. I've been thinking that it may still be better to leave the input trees unchanged.

Having the unchanged trees available means that we can often simplify away some terms from the merge when we do repeated rebases. We currently often end up with more complex conflicts as a commit gets rebased multiple times. That happens if the commit has multiple files that conflict with different files in the destination commits. That's much less of a problem since we started simplifying conflicts when materializing them, but it's still a bit ugly.

While keeping the partially resolved trees can save time, it can also cost time in other cases. For example, when determining if a merge commit is empty, we currently have to redo the resolution of the parent trees (but we don't recurse into files). If we instead aborted the merge process and stored the unchanged trees as soon as we ran into a conflict, it would be cheaper to find the parent state and determine if it matches. Similarly, being able to simplify away terms when rebasing a commit repeatedly, as mentioned above, should make it cheaper.

If we make this change, we'll need to make sure that snapshotting doesn't update unchanged paths, even if the file has been `touch`ed. We would still update all trees if a file changed (whether or not if it was previously resolved), but I don't think that should happen without an explicit action (i.e. editing the file) from the user. We'll also need to make sure that `ConflictIterator` doesn't [assume that there are no resolvable conflicts](https://github.com/martinvonz/jj/blob/fa7f4e68c5183cba9a41f497907aae381c73b563/lib/src/merged_tree.rs#L704-L706).

I've been thinking about this for a while. #4151 is what made me post it now. #4062 is also related.

scott2000:
I think this would be a good change to make, especially now that we have conflict labels for each tree. It would also probably make #8377 display more accurate labels in some cases. Partially resolving conflicts causes the trees to no longer match the content of the commit they came from, which can lead to confusing scenarios where two terms have the same label but different contents.

martinvonz:
I have some draft commits for it, so let me know if you want me to push them for you to take a look so you don't need to redo the work.

trast-google:
I stumbled into very high root tree count while rebasing through a branch that uses .golden testcase files. Due to how things are structured, every substantial change will end up having to change all the .golden, then resulting in conflicts in any subsequent change that is being rebased.

This turns out to be reproducible as follows:

```shell
#!/bin/sh
set +x
[...]

PR #4250, opened 2024-08-10
merged_tree: split function that merges trees without resolving conflicts

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/default_index/revset_engine.rs
+++ b/lib/src/default_index/revset_engine.rs
@@ -1142,7 +1142,8 @@ fn has_diff_from_parent(
             return Ok(false);
         }
     }
-    let from_tree = rewrite::merge_commit_trees_without_repo(store, &index, &parents)?;
+    let from_tree =
+        rewrite::merge_commit_trees_no_resolve_without_repo(store, &index, &parents)?.resolve()?;
     let to_tree = commit.tree()?;
     let mut tree_diff = from_tree.diff_stream(&to_tree, matcher);
     async {
@@ -1163,7 +1164,8 @@ fn matches_diff_from_parent(
     files_matcher: &dyn Matcher,
 ) -> BackendResult<bool> {
     let parents: Vec<_> = commit.parents().try_collect()?;
-    let from_tree = rewrite::merge_commit_trees_without_repo(store, &index, &parents)?;
+    let from_tree =
+        rewrite::merge_commit_trees_no_resolve_without_repo(store, &index, &parents)?.resolve()?;
     let to_tree = commit.tree()?;
     let tree_diff = from_tree.diff_stream(&to_tree, files_matcher);
     // Conflicts are compared in materialized form. Alternatively, conflict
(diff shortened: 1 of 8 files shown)


lib/src/rewrite.rs
```diff
@@ -58,9 +64,10 @@ pub fn merge_commit_trees_without_repo(
                 .iter()
                 .map(|id| store.get_commit(id))
                 .try_collect()?;
-            let ancestor_tree = merge_commit_trees_without_repo(store, index, &ancestors)?;
+            let ancestor_tree =
+                merge_commit_trees_no_resolve_without_repo(store, index, &ancestors)?;
             let other_tree = other_commit.tree()?;
-            new_tree = new_tree.merge(&ancestor_tree, &other_tree)?;
+            new_tree = new_tree.merge_no_resolve(&ancestor_tree, &other_tree);
```
yuja, line 70: @martinvonz Is this assumption valid?

> Since "ancestor_tree" isn't resolved by itself, this patch has subtle behavior
> change. For example, "jj diff -r9eaef582" in the "git" repository is no longer
> empty. I think the new behavior is also technically correct, but I'm not pretty
> sure.

martinvonz: I think that may be due to the `A+(A-B)=A` rule resulting being applied in the ancestor before this patch (for those not familiar, that's https://github.com/martinvonz/jj/blob/bf2cb55ea7475adb36a93c8632979f956bbdedbf/lib/src/merge.rs#L92-L105 and also line 65 in that file). I've confirmed that removing that rule results in the same diff as with this patch (plus a ton of no-op "Resolved conflict" diffs). So I think in some sense the new behavior is more correct.

lib/src/merged_tree.rs
martinvonz, line 1: > This allows us to resolve conflicts of matching files only.

What do you mean by this?

FWIW, I've considered making the same change for #4152. Or maybe it was just in the context of that - that change will make it so we can more often simplify whole trees away.

Another advantage of it is that it reduces roundtrips to the backend by walking more trees in parallel. Without this change, we first walk from the roots to the differing leaves in the ancestor merge, then we walk again from the roots when merging the descendants (or other siblings in the case of octopus).

lib/src/rewrite.rs
martinvonz: Good find, btw!

lib/tests/test_revset.rs
martinvonz, line 1: nit: split this to a separate commit since it doesn't seem to depend on the changes in this commit

lib/src/default_index/revset_engine.rs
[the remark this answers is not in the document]
yuja: Updated, thanks.

lib/src/merged_tree.rs
yuja: Added code snippets. It's about `file()` revset.
```
merged_tree: allow to postpone resolution of intermediate trees

This allows us to diff trees without fully resolving conflicts:

    let from_tree = merge_no_resolve(..);
    for (path, (from, to)) in from_tree.diff(to_tree, matcher) {
        let from = resolve_conflicts(from);
        if from == to {
            continue; // resolved file may be identical
        ...
```

martinvonz (approved):

PR #4254, opened 2024-08-12
rewrite: migrate `move_commits` function from `rebase` command

bnjmnt4n:
# Checklist

If applicable:


yuja (approved):
Thanks!

lib/src/rewrite.rs
```diff
@@ -437,3 +440,361 @@ impl<'settings, 'repo> DescendantRebaser<'settings, 'repo> {
...
+
+pub struct MoveCommitsStats {
+    /// The number of commits in the target set which were rebased.
+    pub num_rebased_targets: u32,
+    /// The number of descendant commits which were rebased.
+    pub num_rebased_descendants: u32,
+    /// The number of commits for which rebase was skipped, due to the commit
+    /// already being in place.
+    pub num_skipped_rebases: u32,
+}
```
bnjmnt4n, line 452: Forgot to mention before merging: now that this is in `jj-lib`, perhaps returning just the number of commits which have been rebased might not be sufficient for consumers. In the future this could perhaps be changed to return the actual commit/commit IDs.

PR #4261, opened 2024-08-13
diff_util: add copy records tracking to `DiffRenderer::show_patch`

bnjmnt4n:
This allow `jj show --summary` and other commands to include copy tracking information.

# Checklist

If applicable:

### DIFF
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -349,13 +349,21 @@ impl<'a> DiffRenderer<'a> {
     ) -> Result<(), DiffRenderError> {
         let from_tree = commit.parent_tree(self.repo)?;
         let to_tree = commit.tree()?;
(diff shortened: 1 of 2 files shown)


cli/src/diff_util.rs
yuja, line 1: Can you add `log`/`show --summary` test to ensure that this code path is working?

yuja (approved):

cli/src/diff_util.rs
bnjmnt4n: I've added some tests for `jj show` based on `jj diff`'s tests. Do you think that's sufficient?

yuja: Yes, thanks. I just meant to add one or two test invocations so future refactoring wouldn't break `show_patch()` callers.

PR #4264, opened 2024-08-14
`rewrite::move_commits()`: preserve order of parent commits

bnjmnt4n:
# Checklist

If applicable:

### DIFF
--- a/lib/src/rewrite.rs
+++ b/lib/src/rewrite.rs
@@ -643,33 +643,39 @@ pub fn move_commits(
         new_children
             .iter()
             .map(|child_commit| {
-                let mut new_child_parent_ids: IndexSet<_> = child_commit
-                    .parent_ids()
-                    .iter()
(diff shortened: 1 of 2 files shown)


lib/src/rewrite.rs
```diff
@@ -654,17 +654,24 @@ pub fn move_commits(
...
+                    // new children.
+                    .flat_map(|id| {
+                        if new_parent_ids
                             .iter()
-                            .any(|new_parent_id| new_parent_id == id)
+                            .any(|new_parent_id| *new_parent_id == id)
+                        {
+                            target_heads.clone()
+                        } else {
+                            [id.clone()].to_vec()
```
yuja, line 668: nit: `vec![id]`

btw, this `.collect()`  would be better expressed as imperative `for` loop.

yuja (approved):
Looks good, thanks!

PR #4269, opened 2024-08-14
copy-tracking: add support for `diff --git`

fowles:
# Checklist

If applicable:

### DIFF
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -308,9 +308,15 @@ impl<'a> DiffRenderer<'a> {
                     show_names(formatter, tree_diff, path_converter)?;
                 }
                 DiffFormat::Git { context } => {
-                    let no_copy_tracking = Default::default();
-                    let tree_diff = from_tree.diff_stream(to_tree, matcher, &no_copy_tracking);
-                    show_git_diff(formatter, store, tree_diff, *context)?;
+                    show_git_diff(
+                        formatter,
+                        store,
+                        from_tree,
+                        to_tree,
+                        matcher,
+                        copy_records,
+                        *context,
+                    )?;
                 }
                 DiffFormat::ColorWords { context } => {
                     let tree_diff = from_tree.diff_stream(to_tree, matcher, copy_records);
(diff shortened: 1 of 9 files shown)


cli/src/diff_util.rs
```diff
@@ -1098,23 +1104,40 @@ fn show_unified_diff_hunks(
...
-            let left_part = git_diff_part(&path, left_value)?;
-            let right_part = git_diff_part(&path, right_value)?;
+
+            let left_part = git_diff_part(&left_path, left_value)?;
+            let right_part = git_diff_part(&right_path, right_value)?;
+
+            // Skip the "delete" entry when there is a rename.
+            if right_part.mode.is_none() && copied_sources.contains(left_path.as_ref()) {
+                continue;
+            }
```
yuja, line 1134: Slightly off topic, but I think this should be handled by `TreeDiffStream`/`Iterator`. From API standpoint, it's odd that source files are adjusted but the destinations aren't.

(Not meant for this PR)

cli/src/diff_util.rs
```diff
@@ -1127,6 +1150,16 @@ pub fn show_git_diff(
                         writeln!(formatter, "index {left_hash}..{right_hash}")?;
                     }
                     (Some(left_mode), Some(right_mode)) => {
+                        if left_path != right_path {
+                            let operation = if to_tree.path_value(&left_path).unwrap().is_absent() {
```
yuja, line 1154: nit: `path_value(..)?`

The return type of this closure will have to be `Result<(), DiffRenderError>`.

cli/tests/test_diff_command.rs
yuja, line 1: nit: Can you add "copy from"/"to" samples for `--git` and `--summary`?

yuja (approved):
Thanks!

cli/src/diff_util.rs
fowles: I agree it is odd.  The core issue is that a copy is defined by

source -> still exists
source -> target

and a rename is defined by

source -> removed
source -> target

so the first case doesn't necessarily show up in a diff of the trees at all.

cli/src/diff_util.rs
fowles: Done

cli/src/diff_util.rs
yuja: I'm not sure if we'll need copy/rename distinction at store layer, but I think we can at least mask "source -> removed" entry by diff stream.

Maybe leave a TODO comment?

cli/tests/test_diff_command.rs
fowles: boom!  This found a misconfig in my git tracking stuff.  Nice one!!

cli/src/diff_util.rs
fowles: I think I would prefer to wait for martin and google folks to have there meeting about how to redo the API here, cause the TODO would end up sprinkled in several places.

PR #4274, opened 2024-08-15
cli: propagate BackendError from tree diffs

yuja:
# Checklist

If applicable:


martinvonz (approved):
Thanks! I didn't remember we had that convenient error type so I thought this would involve something slightly more complicated.

yuja:
I assume this was the problem that `formatter.with_label()` can only return `io::Error`.

Issue #3947, opened 2024-06-22
FR: Add a always `edit` mode to `prev/next`


PhilipMetzger:
Some people don't use the `new` + `squash` workflow and always want to edit their commits which would match the behavior of `sl next/prev`.

martinvonz:
Reminder to address https://github.com/martinvonz/jj/pull/3935#discussion_r1649520967 if we add a config for edit mode (and presumably stop making it implied in some cases then).

essiene:
I don't really use the `new+squash` workflow myself.

What would be a good config flag name for this?

```
[ui]
prev-always-edit = true|false
next-always-edit = true|false
```

thoughts?

essiene:
> Would this conflict with the decision of not allowing default configuration for commands to avoid inconsistent behavior in #1509?

Hmmm... I'm not certain, though my own reading of the conclusion of #1509 is that allowing arbitrary defaults for any commands is the undesirable thing. This issue, IIUC, doesn't change the default for the `--edit` flag, it sets a default behaviour which should hold in the absence of the flag and will still respect the flag if set in either direction.

so if `ui.prev-always-edit = true`, but I run `jj prev --edit=false` I expect this to respect the commandline flag. Conversely `ui.prev-always-edit = false` (the intended default) with `jj prev --edit` should behave as expected too.

I'd like to get other opinions though, before proceeding in case I've interpreted that wrong.

yuja:
> Would this conflict with the decision of not allowing default configuration for commands

Suppose edit/squash workflow is user preference, config knob seems good. It might be also okay to remove or deprecate the `--edit` flag.

I personally want a config to never continue editing by `jj new --insert-before && jj prev`.

joyously:
Since there is an `--edit` flag, wouldn't the solution be an alias, so that it follows 1509?

PhilipMetzger:
Since this a FR based on this comment https://github.com/martinvonz/jj/issues/2126#issuecomment-1707527036, I always thought of having a `ui.movement.always_edit` flag, which responds to passing `jj next/prev --edit`. 

> > Would this conflict with the decision of not allowing default configuration for commands
> 
> Suppose edit/squash workflow is user preference, config knob seems good.

Yes, that was always my goal here.

> It might be also okay to remove or deprecate the `--edit` flag.

But I don't think we should deprecate it, as usually use `edit` to move along the stack and then use the `new` + `squash` version for dealing with conflicts.

essiene:
[...]

PR #4282, opened 2024-08-15
next/prev: refactor movement utilities into cli/src/movement_utils.rs

essiene:
The code in both cli/src/commands/{next,prev}.rs is identical except for the direction of movement. This PR pull the parts that make sense out into cli/src/movement_util.rs so it's easier to see the differences. No new behavious is introduced.

Part of #3947

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/next.rs
+++ b/cli/src/commands/next.rs
@@ -12,15 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::io::Write;
-
-use itertools::Itertools;
-use jj_lib::commit::Commit;
-use jj_lib::repo::Repo;
-use jj_lib::revset::{RevsetExpression, RevsetFilterPredicate, RevsetIteratorExt};
-
-use crate::cli_util::{short_commit_hash, CommandHelper, WorkspaceCommandHelper};
-use crate::command_error::{user_error, CommandError};
+use crate::cli_util::CommandHelper;
+use crate::command_error::CommandError;
+use crate::movement_util::{move_to_commit, Direction};
 use crate::ui::Ui;
 
 /// Move the working-copy commit to the child revision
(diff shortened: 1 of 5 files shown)


cli/src/movement_util.rs
```diff
@@ -0,0 +1,170 @@
...
+        [target] => target,
+        [] => {
+            // We found no descendant.
+            return Err(user_error(format!(
+                "No {} found {} commit{} {}",
+                direction.next_node_type(),
+                change_offset,
+                if change_offset > 1 { "s" } else { "" },
+                direction.direction(),
+            )));
```
martinvonz, line 128: This message is hard to read and hard to internationalize. I would prefer if you make this two separate messages depending on direction instead (i.e. something like `let message = if direction == Direction::Next { ... } else { ... };`).

(I don't like the `if change_offset > 1 { "s" } else { "" }` bit either but that was already there.)

cli/src/movement_util.rs
```diff
@@ -0,0 +1,170 @@
...
+    let targets: Vec<Commit> = target_revset
+        .evaluate_programmatic(workspace_command.repo().as_ref())?
+        .iter()
+        .commits(workspace_command.repo().store())
+        .try_collect()?;
+
+    let target = match targets.as_slice() {
+        [target] => target,
+        [] => {
+            // We found no descendant.
```
martinvonz, line 121: nit: "descendant" => "descendant/ancestor"

cli/src/movement_util.rs
```diff
@@ -0,0 +1,170 @@
...
+                direction.direction(),
+            )));
+        }
+        commits => choose_commit(ui, workspace_command, &direction, commits)?,
+    };
+
+    Ok(target.clone())
+}
+
+pub fn choose_commit<'a>(
```
martinvonz, line 136: nit: doesn't need to be public?

cli/src/commands/next.rs
```diff
@@ -117,52 +79,24 @@ pub(crate) fn cmd_next(
...
-        start_revset.descendants_at(args.offset)
-    }
-    .minus(&wc_revset);
-
-    let targets: Vec<Commit> = target_revset
-        .evaluate_programmatic(workspace_command.repo().as_ref())?
-        .iter()
-        .commits(workspace_command.repo().store())
-        .try_collect()?;
+    let target = get_target_commit(
```
martinvonz, line 82: The rest of `cmd_next()` and `cmd_prev()` look very similar. Also move that into the new module?

martinvonz (commented):
nit: I think "no-op" and "Refactor" is redundant. Refactoring always has (or should have) no visible effect on the behavior. So e.g. "next/prev: refactor ..." would be slightly clearer to me, and "This no-op PR" sounds better as "This refactoring commit/patch" sounds better to me (it's clearer about the sense in which it's a no-op).

PhilipMetzger:
I had a approach here: https://github.com/PhilipMetzger/jj/commit/9c37e18e5db1d5da0a7a75133624d1ab8b56e8d6, although I renamed the directions to just `Direction::Ancestors` and `Direction::Descendants`.

cli/src/movement_util.rs
essiene: > This message is hard to read and hard to internationalize. I would prefer if you make this two separate messages depending on direction instead (i.e. something like `let message = if direction == Direction::Next { ... } else { ... };`).
> 

Makes sense. Done.

> (I don't like the `if change_offset > 1 { "s" } else { "" }` bit either but that was already there.)

Since I've touched this anyway, I've refactored this out also, so I spell out "commit" or "commits" completely, so it should be friendly to internalization.

Also, there's the age old parenthesized plural like: `No ancestor found 1 commit(s) back`. We could go with that as well?

cli/src/movement_util.rs
essiene: fixed. copy+pasta fallout.

cli/src/commands/next.rs
essiene: Thanks for the nudge. Moved them all over now.

essiene:
> nit: I think "no-op" and "Refactor" is redundant. Refactoring always has (or should have) no visible effect on the behavior. So e.g. "next/prev: refactor ..." would be slightly clearer to me, and "This no-op PR" sounds better as "This refactoring commit/patch" sounds better to me (it's clearer about the sense in which it's a no-op).

Makes sense. Reworded.

cli/src/movement_util.rs
essiene: Fixed.

cli/src/movement_util.rs
martinvonz: > so I spell out "commit" or "commits" completely, so it should be friendly to internalization.

Not necessarily, actually. I've heard Polish (maybe other languages too) has a different form for 1, 2, and 3 :) I've personally been going with just "commits" (i.e. including "1 commits") until we do i18n for real. I don't really care what we do until then.

cli/src/movement_util.rs
```diff
@@ -0,0 +1,217 @@
...
+        tx.finish(
+            ui,
+            format!("next: {current_short} -> editing {target_short}"),
+        )?;
+        return Ok(());
+    }
+    let mut tx = workspace_command.start_transaction();
+    // Move the working-copy commit to the new parent.
+    tx.check_out(&target)?;
+    tx.finish(ui, format!("next: {current_short} -> {target_short}"))?;
```
martinvonz, line 215: the `next` in these two places should be `direction.cmd()`, right?

martinvonz (approved):

cli/src/movement_util.rs
essiene: lol! This is the real world we're talking about. It's all exceptions and a few rules ;-)

`commit(s)` it is!

cli/src/movement_util.rs
essiene: Whoops! Fixed.

cli/src/movement_util.rs
```diff
@@ -0,0 +1,210 @@
...
+use jj_lib::revset::{RevsetExpression, RevsetFilterPredicate, RevsetIteratorExt};
+
+use crate::cli_util::{short_commit_hash, WorkspaceCommandHelper};
+use crate::command_error::{user_error, CommandError};
+use crate::ui::Ui;
+
+pub enum Direction {
+    Next,
+    Prev,
+}
```
yuja, line 31: nit: add `#[derive(Clone, Copy, Debug, Eq, PartialEq)]` as `Direction` is a simple value?

cli/src/movement_util.rs
```diff
@@ -0,0 +1,210 @@
...
+
+impl Direction {
+    fn cmd(&self) -> String {
+        match self {
+            Direction::Next => "next".to_string(),
+            Direction::Prev => "prev".to_string(),
+        }
```
yuja, line 38: nit: `-> &'static str` makes sense here.

cli/src/movement_util.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+fn get_target_commit(
+    ui: &mut Ui,
+    workspace_command: &WorkspaceCommandHelper,
+    direction: &Direction,
+    working_commit_id: &CommitId,
+    edit: bool,
+    has_conflict: bool,
+    change_offset: u64,
```
yuja, line 102: nit: I would extract `edit, has_conflict, change_offset` to `struct MovementOptions` or something. 

(It could be `MovementArgs` flattened to `Next`/`PrevArgs`, but we would need to rephrase the help messages.)

cli/src/movement_util.rs
essiene: See #4297

cli/src/movement_util.rs
essiene: See #4297

cli/src/movement_util.rs
essiene: See #4297

Issue #4238, opened 2024-08-08
FR: Better support the edit workflow


matts1:
**Is your feature request related to a problem? Please describe.**

The creation of new empty commits when the current commit is emptied *is magic*, and magic is bad as it's hard to explain, hard to understand, and often requires special-casing. I believe (and feel free to correct me if I'm wrong) that the only reason it exists in `jj` is that it maps well to the concept of `@-` being the "current commit". Is the magic worth it? For a squash workflow user, yes, absolutely, so I'm not advocating for the removal of the feature.

For an edit workflow user though, this means that `jj squash` actually is equivalent to `jj squash --from @- --to @--; jj commit -m ""`. This is completely nonsensical, and fights what any edit user wants to do.

***Why edit users should be first-class citizens***
I personally believe that combining `jj` and the [edit workflow](https://steveklabnik.github.io/jujutsu-tutorial/real-world-workflows/the-edit-workflow.html) provides probably the simplest abstraction of a version control system that I've ever seen. To me, when I was first taught `jj`, I used the squash workflow, and tried to understand it as "fig (mercurial), but it runs `hg evolve` after every operation".

However, by the time I'd used it for a few days, I found that trying to think of it in terms of mercurial was just holding me back, and I'd be better off thinking about it in terms of what it is - `jj`. The key thing to that understanding was realizing that I needed to think of a repository as a tree of commits, and that I needed to think of each commit as *both* a tree and a patch, not just one or the other. When I thought about it this way, everything became *much* simpler, and the edit workflow came naturally to me (I didn't read documentation on the edit workflow until after starting to use it).

The edit workflow is simpler than the squash workflow, although it's probably wierder for people who try to think of things in terms of another version control system.
* `@` isn't special. Nor is `@-`. They're commits like any other.
* I don't need special-cases like "squash merges two commits into one, except if it's your currently checked out commit"
* I no longer need a concept for "amend". I just edit it and it's already in there.
  * If I get told, "hey, can you fix commit foo", I `jj edit foo`, make my changes, then `jj upload foo`. No amend
  * This is a blessing for some, and a curse for others. I personally love it, but this is a valid reason why people might prefer the squash workflow even if they find this mental model simpler.
    * When I'm making a change I'm not sure about, I often create a new commit for the changes, then run squash later, similar to how someone using the squash workflow would use it. I still think about it mentally as the edit workflow though.

[...]

Issue #1509, opened 2023-04-12
FR: add config for default options


fowles:
I would like to be able to set default arguments that apply to commands if no other arguments are provided.  I know that something similar can be accomplished with aliases, but I like being able to use commands like `jj log` with some customization.

I tried to implement it in

https://github.com/fowles/jj/commit/c4c8040488d6eb776e9d4a3fb7aa0129f9313a36

but I got lost trying to work with the clap module.  If you would be receptive to this feature and can point me in the right direction, I can try to finish a PR, but if you don't like it I would rather learn that then continue spelunking through clap docs.

martinvonz:
Mercurial has this feature, and I remember they deprecated it, but I couldn't remember why. After some quick digging, I found https://lists.mercurial-scm.org/pipermail/mercurial-devel/2009-October/016095.html. For reference, the arguments there are:

* It can be hard to override the defaults. If you make `--reversed` the default log order, we'd need to add a `--no-reversed` option for this case.
* It's likely to confuse scripts.
* It's likely to confuse your coworkers when they're trying to help you with something on your machine.

The last two points also apply to more specific config options, like `diff.format` (which can be set to "git", for example). Mercurial has its `$HGPLAIN` environment variable for the script case: scripts are supposed to set that variable, which disables most UI customizations.

So I'm not sure if we want to add this feature, but I'm also not sure that shouldn't.

fowles:
If a real project had cause to deprecate it, then I will listen to that wisdom.  It was probably painfully learned.

durin42:
@martinvonz summarized the issues correctly. It __might__ be less of an issue iff you could convince people to use a proper API to interact with `jj` working copies, but as a practical matter I've known too many programmers and they'll absolutely write skeevy shell scripts that will be broken by changes in default values. In general it's been significantly more sustainable to use a config section like `alias` to allow users to define their own custom commands which don't break out-of-the-box scriptability.

Git has tried to deal with this by having commands be "plumbing" and "porcelain" and then insisting that you only write scripts with the former. In practice...nobody listens until they get burned, and then the horse is already out of the barn and in the next county.

Hyrum's Law always finds you.

PR #4283, opened 2024-08-15
Add config flag to control `prev/next` edit behaviour.

essiene:
* We started with a tristate flag where:
    - Auto - Maintain current behaviour. This edits if
      the wc parent is not a head commit. Else, it will
      create a new commit on the parent of the wc in
      the direction of movement.
    - Always - Always edit
    - Never - Never edit, prefer the new+squash workflow.
  However, consensus the review thread is that `auto` mode where we try to infer when to
  switch to `edit mode`, should be removed. So `ui.movement.edit` is a boolean flag now.
    - true: edit mode
    - false: new+squash mode
* Also add a `--no-edit` flag as the explicit inverse of `--edit` and
  ensure both flags take precedence over the config.
* Update tests that assumed edit mode inference, to specify `--edit` explicitly.

NOTE: #4302 was squashed into this commit, so see that closed PR for review history.

Part of https://github.com/martinvonz/jj/issues/3947

### DIFF
--- a/cli/src/commands/next.rs
+++ b/cli/src/commands/next.rs
@@ -51,15 +51,24 @@ use crate::ui::Ui;
 #[command(verbatim_doc_comment)]
 pub(crate) struct NextArgs {
     /// How many revisions to move forward. Advances to the next child by
-    /// default.
+    /// default
     #[arg(default_value = "1")]
     offset: u64,
     /// Instead of creating a new working-copy commit on top of the target
     /// commit (like `jj new`), edit the target commit directly (like `jj
-    /// edit`).
+    /// edit`)
+    ///
+    /// Takes precedence over config in `ui.movement.edit`; i.e.
+    /// will negate `ui.movement.edit = false`
     #[arg(long, short)]
     edit: bool,
-    /// Jump to the next conflicted descendant.
+    /// The inverse of `--edit`
+    ///
+    /// Takes precedence over config in `ui.movement.edit`; i.e.
+    /// will negate `ui.movement.edit = true`
+    #[arg(long, short, conflicts_with = "edit")]
+    no_edit: bool,
+    /// Jump to the next conflicted descendant
     #[arg(long, conflicts_with = "offset")]
     conflict: bool,
 }
@@ -69,6 +78,7 @@ impl From<&NextArgs> for MovementArgs {
         MovementArgs {
             offset: val.offset,
             edit: val.edit,
+            no_edit: val.no_edit,
             conflict: val.conflict,
         }
     }
@@ -79,11 +89,5 @@ pub(crate) fn cmd_next(
     command: &CommandHelper,
     args: &NextArgs,
 ) -> Result<(), CommandError> {
-    let mut workspace_command = command.workspace_helper(ui)?;
-    move_to_commit(
-        ui,
-        &mut workspace_command,
-        Direction::Next,
-        &MovementArgs::from(args),
-    )
+    move_to_commit(ui, command, Direction::Next, &MovementArgs::from(args))
 }
--- a/cli/src/movement_util.rs
+++ b/cli/src/movement_util.rs
@@ -21,7 +21,7 @@ use jj_lib::commit::Commit;
 use jj_lib::repo::Repo;
 use jj_lib::revset::{RevsetExpression, RevsetFilterPredicate, RevsetIteratorExt};
 
-use crate::cli_util::{short_commit_hash, WorkspaceCommandHelper};
+use crate::cli_util::{short_commit_hash, CommandHelper, WorkspaceCommandHelper};
 use crate::command_error::{user_error, CommandError};
 use crate::ui::Ui;
 
@@ -29,7 +29,15 @@ use crate::ui::Ui;
 pub(crate) struct MovementArgs {
     pub offset: u64,
     pub edit: bool,
+    pub no_edit: bool,
     pub conflict: bool,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct MovementArgsInternal {
+    offset: u64,
+    should_edit: bool,
+    conflict: bool,
 }
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
@@ -56,12 +64,12 @@ impl Direction {
     fn get_target_revset(
         &self,
         working_commit_id: &CommitId,
-        args: &MovementArgs,
+        args: &MovementArgsInternal,
     ) -> Result<Rc<RevsetExpression>, CommandError> {
         let wc_revset = RevsetExpression::commit(working_commit_id.clone());
         // If we're editing, start at the working-copy commit. Otherwise, start from
         // its direct parent(s).
-        let start_revset = if args.edit {
+        let start_revset = if args.should_edit {
             wc_revset.clone()
         } else {
             wc_revset.parents()
@@ -103,7 +111,7 @@ fn get_target_commit(
     workspace_command: &WorkspaceCommandHelper,
     direction: Direction,
     working_commit_id: &CommitId,
-    args: &MovementArgs,
+    args: &MovementArgsInternal,
 ) -> Result<Commit, CommandError> {
     let target_revset = direction.get_target_revset(working_commit_id, args)?;
     let targets: Vec<Commit> = target_revset
@@ -162,29 +170,29 @@ fn choose_commit<'a>(
 
 pub(crate) fn move_to_commit(
     ui: &mut Ui,
-    workspace_command: &mut WorkspaceCommandHelper,
+    command: &CommandHelper,
     direction: Direction,
     args: &MovementArgs,
 ) -> Result<(), CommandError> {
+    let mut workspace_command = command.workspace_helper(ui)?;
+
     let current_wc_id = workspace_command
         .get_wc_commit_id()
         .ok_or_else(|| user_error("This command requires a working copy"))?;
 
-    let mut args = args.clone();
-    args.edit = args.edit
-        || !&workspace_command
-            .repo()
-            .view()
-            .heads()
-            .contains(current_wc_id);
+    let config_edit_flag = command.settings().config().get_bool("ui.movement.edit")?;
+    let args = MovementArgsInternal {
+        should_edit: args.edit || (!args.no_edit && config_edit_flag),
+        offset: args.offset,
+        conflict: args.conflict,
+    };
 
-    let target = get_target_commit(ui, workspace_command, direction, current_wc_id, &args)?;
+    let target = get_target_commit(ui, &workspace_command, direction, current_wc_id, &args)?;
     let current_short = short_commit_hash(current_wc_id);
     let target_short = short_commit_hash(target.id());
     let cmd = direction.cmd();
-
     // We're editing, just move to the target commit.
-    if args.edit {
+    if args.should_edit {
         // We're editing, the target must be rewritable.
         workspace_command.check_rewritable([target.id()])?;
         let mut tx = workspace_command.start_transaction();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,17 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* `next/prev` will no longer infer when to go into edit mode when moving from
+  commit to commit. It now either follows the flags `--edit|--no-edit` or it
+  gets the mode from `ui.movement.edit`.
+
 ### Deprecations
 
 ### New features
+
+* Add new boolean config knob, `ui.movement.edit` for controlling the behaviour
+  of `prev/next`. The flag turns `edit` mode `on` and `off` permanently when set
+  respectively to `true` or `false`.
 
 * The following diff formats now include information about copies and moves:
   `--color-words`, `--git`, `--stat`, `--summary`, `--types`, and external diff
--- a/cli/src/commands/prev.rs
+++ b/cli/src/commands/prev.rs
@@ -45,17 +45,26 @@ use crate::ui::Ui;
 /// A        A
 /// ```
 /// If the working copy revision already has visible children, then `--edit` is
-/// implied.
+/// implied
 #[derive(clap::Args, Clone, Debug)]
 #[command(verbatim_doc_comment)]
 pub(crate) struct PrevArgs {
-    /// How many revisions to move backward. Moves to the parent by default.
+    /// How many revisions to move backward. Moves to the parent by default
     #[arg(default_value = "1")]
     offset: u64,
-    /// Edit the parent directly, instead of moving the working-copy commit.
+    /// Edit the parent directly, instead of moving the working-copy commit
+    ///
+    /// Takes precedence over config in `ui.movement.edit`; i.e.
+    /// will negate `ui.movement.edit = false`
     #[arg(long, short)]
     edit: bool,
-    /// Jump to the previous conflicted ancestor.
+    /// The inverse of `--edit`
+    ///
+    /// Takes precedence over config in `ui.movement.edit`; i.e.
+    /// will negate `ui.movement.edit = true`
+    #[arg(long, short, conflicts_with = "edit")]
+    no_edit: bool,
+    /// Jump to the previous conflicted ancestor
     #[arg(long, conflicts_with = "offset")]
     conflict: bool,
 }
@@ -65,6 +74,7 @@ impl From<&PrevArgs> for MovementArgs {
         MovementArgs {
             offset: val.offset,
             edit: val.edit,
+            no_edit: val.no_edit,
             conflict: val.conflict,
         }
     }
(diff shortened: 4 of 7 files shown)


cli/src/movement_util.rs
```diff
@@ -179,12 +179,17 @@ pub fn move_to_commit(
...
+        MovementEditMode::Always => true,
+        MovementEditMode::Never => false,
+        MovementEditMode::Auto => {
+            edit || !&workspace_command
+                .repo()
+                .view()
+                .heads()
+                .contains(current_wc_id)
+        }
+    };
```
martinvonz, line 192: Looks like `--edit` is now ignored. I think we should either remove the flag or let the flag override the config (and provide `--no-edit` to override in the other direction)

PhilipMetzger: Or just respect the flag in all cases?

martinvonz: Yes, but how is that different from what I said?

PhilipMetzger: My bad I only saw the last part of the message.

cli/src/ui.rs
```diff
@@ -397,6 +421,10 @@ impl Ui {
         }
     }
 
+    pub fn movement_edit_mode(&self) -> MovementEditMode {
+        self.movement.edit.to_owned()
+    }
```
yuja, line 426: Despite the config name, it's unrelated to the `Ui` object. Move it to movement_util.rs?

cli/src/movement_util.rs
essiene: Fixed properly now. Thanks.

cli/src/ui.rs
essiene: Done.

OT: Btw, is there a bug filed to make config loading single homed? `UserSettings` seems like the correct location. If there is already an issue filed for this? If nobody owns it, I'd like to take it after this issue is closed, else, I'll file one and assign to myself.

yuja: I don't remember if there was the exact issue, but iirc we'd rather want separate config definitions/structs (per feature or something.) A single big config definition doesn't work because extensions might need their own settings.

essiene: I see. This makes sense. Especially since commands have access to config via `command.settings().config()`, so it's not really a big deal.

Thanks for the context.

cli/src/commands/next.rs
```diff
@@ -56,9 +56,15 @@ pub(crate) struct NextArgs {
...
-    #[arg(long, short)]
+    /// edit`). Takes precedence over config in `ui.movement.edit`; i.e.
+    /// will negate `ui.movement.edit = "never"`
+    #[arg(long, short, conflicts_with = "no_edit")]
     edit: bool,
+    /// The inverse of `--edit`.
+    /// Takes precedence over config in `ui.movement.edit`; i.e.
+    /// will negate `ui.movement.edit = "always"`
+    #[arg(long, short, conflicts_with = "edit")]
+    no_edit: bool,
```
yuja, line 67: Another option is to allow `--edit` and `--edit=true|false`, but I don't know if we like it.
https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_missing_value

This will probably apply to `git init --no-colocate`.

cli/src/movement_util.rs
```diff
@@ -201,3 +220,34 @@ pub(crate) fn move_to_commit(
...
+    }
+}
+
+impl TryFrom<&Config> for MovementSettings {
+    type Error = CommandError;
+
+    fn try_from(config: &Config) -> Result<Self, CommandError> {
+        config
+            .get::<Self>("ui.movement")
+            .map_err(|err| config_error_with_message("Invalid `ui.movement`", err))
```
yuja, line 251: nit: you can inline this to caller. If we drop support for `auto`, this will be just `config.get_bool("ui.movement")?`.

yuja (commented):
The code looks good to me, thanks. I'm not sure about:
* config naming: `ui.movement.edit`
* should we remove "auto"? -> maybe yes?
* `--no-edit` or `--edit[=false]`

cli/src/commands/next.rs
martinvonz: I'm not sure I like it either, but an advantage of that style is that we can presumably extend it later to support e.g. `--edit=auto` (I don't want that particular option; I'm just saying that it's useful in general).

martinvonz:
> The code looks good to me, thanks. I'm not sure about:
> 
> * config naming: `ui.movement.edit`

I don't know either. Maybe we can find something more generic that applies also to other scenarios (like whether to edit the/a parent when abandoning the working copy as in #4238). I don't have a better proposal right now, though.

> * should we remove "auto"? -> maybe yes?

I also vote for removing it.

> * `--no-edit` or `--edit[=false]`

I wonder if we can even get rid of the flag. Do you think people will want to set the config to one mode and then once in a while use the other mode? If we don't expect people to do that, we could add the config (as this PR does) and also deprecate `--edit`.

PhilipMetzger:
> > * should we remove "auto"? -> maybe yes?
> 
> I also vote for removing it.

Yes, I'm also in favor of removing it as the inference didn't really help from the feedback it received. 

> > * `--no-edit` or `--edit[=false]`
> 
> I wonder if we can even get rid of the flag. Do you think people will want to set the config to one mode and then once in a while use the other mode? If we don't expect people to do that, we could add the config (as this PR does) and also deprecate `--edit`.

I use both modes and won't opt-in to `always_edit` mode, so removing it would at least hurt my workflow.

cli/src/commands/next.rs
essiene: So today we have the following states:

```
--edit, --no-edit - illegal
--edit,   --      - edit mode
   -- , --no-edit - new+squash mode
   -- ,   --      - get mode from config
```

if we go with a purely boolean flag, the config will have no effect:

```
--edit=true                 - edit mode
--edit=false                - new+squash mode
--edit (implies true)       - edit mode
unspecified (implies false) - new+squash mode
```

What we'll actually want is `--edit=true|false|config` with `default_value="config", default_missing_value="true"`
```

--edit=true                 - edit mode
--edit=false                - new+squash mode
--edit (implies true)       - edit mode
unspecified (implies config)- get mode from config
```

I personally find this tri-state, but mostly boolean flag to be weird and surprising behaviour. `--edit |--no-edit` is more expected behaviour.

That said, if the consensus is `--edit=true|false|config`, I'll go ahead to implement that.

cli/src/movement_util.rs
essiene: Done in #4302

essiene:
> > > * should we remove "auto"? -> maybe yes?
> > 
> > 
> > I also vote for removing it.
> 
> Yes, I'm also in favor of removing it as the inference didn't really help from the feedback it received.
> 

Removing it in #4302 to keep test churn to a minimum in this PR.

> > > * `--no-edit` or `--edit[=false]`
> > 
> > 
> > I wonder if we can even get rid of the flag. Do you think people will want to set the config to one mode and then once in a while use the other mode? If we don't expect people to do that, we could add the config (as this PR does) and also deprecate `--edit`.
> 
> I use both modes and won't opt-in to `always_edit` mode, so removing it would at least hurt my workflow.

Noted. Keeping the flag(s).

cli/src/commands/next.rs
joyously: I still think you are describing what 1509 says to avoid -- commands behaving differently for different people because the config changes the default.

essiene: > I still think you are describing what 1509 says to avoid -- commands behaving differently for different people because the config changes the default.

I do think there's an important difference. #1509 says not to allow making arbitrary cli flags into defaults as that leads differing behaviours for different users. The reason for that is because anyone can make a flag that should be explicitly enabled into an implicit default.

This PR doesn't do that. It adds a very narrow and specific default that is available for everyone, and importantly, provides the same behaviour, if the same config value is set by different users.

martinvonz: > --edit=true                 - edit mode
> --edit=false                - new+squash mode
> --edit (implies true)       - edit mode
> unspecified (implies config)- get mode from config

I was hoping that `--edit` (without a value) would be disallowed. `--edit=` with an empty string, or anything that's not "true", "false", or "config", would be an error. So I think that means a string-valued flag with a default of "config". Is that possible? Would you find that behavior clearer?

yuja: Unless we need an explicit `--edit=<USE_CONFIG_VALUE>`, it can be implemented as `Option<bool>` argument.

FWIW, I found `jj new` has `--no-edit`, so `--edit`/`--no-edit` might make sense.

martinvonz: True, we already have some `--no-*` flags, so `--no-edit` seems least surprising then.

essiene: Nice. So I'll leave it with `--edit|--no-edit` then. Resolving this. Do re-open if I've missed something. Thanks.

PhilipMetzger:
> How about we keep this in `ui.movement.*` for now, and whenever #4238 is settled, we can retire that for a flag provided with #4238 ?

While I'm fine with the flag name, I would keep it separate from #4238 and its implementation until there's an overarching design and convergence on it.

essiene:
> > How about we keep this in `ui.movement.*` for now, and whenever #4238 is settled, we can retire that for a flag provided with #4238 ?
> 
> While I'm fine with the flag name, I would keep it separate from #4238 and its implementation until there's an overarching design and convergence on it. 

+1 - I was more making the point that the current config name was fine for it's limited purpose and whenever we arrive at a better overall plan, we can subsume it then.

cli/src/commands/next.rs
```diff
@@ -56,9 +56,15 @@ pub(crate) struct NextArgs {
     offset: u64,
     /// Instead of creating a new working-copy commit on top of the target
     /// commit (like `jj new`), edit the target commit directly (like `jj
-    /// edit`).
-    #[arg(long, short)]
+    /// edit`). Takes precedence over config in `ui.movement.edit`; i.e.
+    /// will negate `ui.movement.edit = "never"`
```
yuja, line 60: nit: split paragraphs and omit period from the summary line. `jj next -h` only shows the summary line.
```
    /// Instead of creating a new working-copy commit on top of the target
    /// commit (like `jj new`), edit the target commit directly (like `jj
    /// edit`)
    ///
    /// Takes precedence over config in `ui.movement.edit`; i.e.  will negate
    /// `ui.movement.edit = "never"`.
```

fix the other help messages as well.

cli/src/commands/next.rs
```diff
@@ -56,9 +56,15 @@ pub(crate) struct NextArgs {
...
+    #[arg(long, short, conflicts_with = "no_edit")]
```
yuja, line 61: nit: remove `, conflicts_with = "no_edit"`?

It's implied by `no_edit`'s attribute, and we don't have the one in `PrevArgs`.

cli/src/movement_util.rs
```diff
@@ -15,23 +15,32 @@
...
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub(crate) struct MovementArgs {
     pub offset: u64,
     pub edit: bool,
+    pub no_edit: bool,
     pub conflict: bool,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+struct MovementArgsInternal {
```
yuja, line 38: nit: remove `Copy`

yuja (approved):
Thanks.

I feel "ui.movement" is a bit vague, but I don't have a suggestion. Maybe we can find a better name or abstraction later.

cli/src/movement_util.rs
essiene: Done. Thanks.

cli/src/commands/next.rs
essiene: Ah, these two files were inconsistent with other commands. I've spot checked a few other commands like `new`, `init`, etc, and they don't have periods. I've removed all the periods at the end of sentences in `next` and `prev` as well. Thanks for catching.

cli/src/commands/next.rs
essiene: Done. The `conflicts_with` attribute is only set on `no_edit` now, in both {next,prev}.rs.

sammko:
Should this note be removed then also? Got quite confused by it https://github.com/martinvonz/jj/blob/86de9139b34734b7dc90d8502c7789818a42e29d/cli/src/commands/next.rs#L50-L51

Same for `prev`.

PhilipMetzger:
Yes, as the behavior no longer exists.

martinvonz:
Thanks for letting us know. #4805 should fix it.

emesterhazy:
I guess this has been decided, but I wanted to mention that I just noticed that the implicit edit was removed today and found it to be a bit surprising.

If I'm editing an interior revision like this:

```
○  vqrkrnzultun
@  ylzwpwovqzoz
│ ○  nkvywtpqkysm
├─╯
○  xxyrxwnurtpu
◆  zzzzzzzzzzzz
```

And type `jj next`, I end up with a new commit on to of a sibling of the commit I was editing:

```
@  lktkwmpvvxlw
○  nkvywtpqkysm
│ ○  vqrkrnzultun
│ ○  ylzwpwovqzoz
├─╯
○  xxyrxwnurtpu
◆  zzzzzzzzzzzz
```

What I really wanted was to move forwards on the commit chain that I'm currently on, which is to say I wanted to edit `vqrkrnzultun` or at least add a new commit on top of it.

The workaround is to type `jj new` and then `jj next` in this situation, I suppose, but I thought the implied edit handled this situation more elegantly. Maybe I'm just a simple guy, but when I type `next` and `@` moves sideways, it makes me wonder what I did wrong.

martinvonz:
You can [configure which behavior you want](https://jj-vcs.github.io/jj/latest/config/#behavior-of-prev-and-next-commands), but the implicit version is not available.

I agree that these commands are confusing but I don't know how to improve it.

emesterhazy:
Is there any strong objection to having the `auto` option? I see that you, Yuya and Philip were in favor of removing it, but it didn't seem like anyone had a particularly strong opinion.

I'm generally in favor of consistency, but I don't think it's bad to give users the option. Here's another scenario that's a bit surprising without the auto edit:

```
○  snpmrywokukw
@  npmzzrornlul
○  xxyrxwnurtpu
◆  zzzzzzzzzzzz
```

If you run `jj next`, it will complain:
```
Error: No other descendant found 1 commit(s) forward from the working copy parent(s)
```

I almost never actually edit commits, and I'm pretty sure that the three of you don't either, which is probably why this didn't feel like a loss. But, it's still occasionally useful to move up and down the commit graph to run `jj diff` without needing to type out the full `jj diff -f xyz -t abc` to check diffs, or `jj edit xyz`. I guess we can use the `-e` flag, but there was also discussion about removing it, which I think would be a mistake.

PhilipMetzger:
@matts1 also was against it, when it was discussed in the Discord. 

> I guess at the end of the day, I think it would have been nice to leave the `auto` config option, but I'm not sure it's worth going back to re-add it. Let's please leave the `-e` flag though.

I don't think that is going away though.

matts1:
> @matts1 also was against it, when it was discussed in the Discord.

To clarify further, I'm not against the idea of being able to configure it. I just think that we should be consistent about it. I'd personally love to add a `ui.movement` flag, but if that's done, IMO it should also:
* Ensure that if you run `jj squash`, you're now editing the squashed commit, not a new commit on top
* Remove the feature that if you're ever left with an empty commit, it creates a new empty commit
* Basically just ensure that the semantics of `jj` align with the edit workflow instead of the squash workflow

What I wanted was to be able to edit the default options for a command (eg. `prev = ["prev", "--edit"]`). However, a decision seems  to have been made to not allow that, in the interest of ensuring that a jj user on one computer can have their basic commands work the same on another computer. I'm personally not a fan of this, as I think that it just means every user will implement their own aliases (eg. I have `rb = ["rebase", "--skip-emptied"]`). I'd personally love it if we could revert that decision, but given that the decision has been made, I think that this flag is inconsistent with the decision.

yuja:
> Is there any strong objection to having the `auto` option? I see that you, Yuya and Philip were in favor of removing it, but it didn't seem like anyone had a particularly strong opinion.
> 
> I'm generally in favor of consistency, but I don't think it's bad to give users the option. Here's another scenario that's a bit surprising without the auto edit:
> 
> ```
> ○  snpmrywokukw
> @  npmzzrornlul
> ○  xxyrxwnurtpu
> ◆  zzzzzzzzzzzz
> ```
> 
> If you run `jj next`

I occasionally do that, but `auto` wouldn't help because I would want to get out of the editing state by `jj next`. `jj next` with `auto` (= `--edit`) would edit `snpmrywokukw`, but that's surprising to me because the editing state is temporary (which I don't expect to persist.)

emesterhazy:
Yeah, but in this case you can't even use `jj next` since it will throw an error. With auto-edit you can at least move forward and then type `jj new` if you want a new commit, or remain in the edit state on `snpmrywokukw` if you want to edit it, or in my case run `jj diff`, `jj desc`, or something else where having `@` on the commit is convenient.

yuja:
To me, an error is preferable than implicit editing. IOW, I don't like stateful `next`/`prev`. (I'm not against adding such behavior if that's useful to many users, though.)

PR #4284, opened 2024-08-15
Update error message when no movement target is found.

essiene:
If movement commands don't find a target commit, they fail. However,                                                                              
it's usually not intuitive why they fail because in non-edit mode the                                                                             
start commit is the parent of the working commit.                                                                                                 
                                                                                                                                                  
Adding the start commit change hash to the error message makes it easier                                                                          
for the user to figure out what is going on. 

Part of #3947 

# Checklist

If applicable:

### DIFF
--- a/cli/src/movement_util.rs
+++ b/cli/src/movement_util.rs
@@ -59,52 +59,89 @@ impl Direction {
         }
     }
 
-    fn target_not_found_message(&self, change_offset: u64) -> String {
-        match self {
-            Direction::Next => format!("No descendant found {} commit(s) forward", change_offset),
-            Direction::Prev => format!("No ancestor found {} commit(s) back", change_offset),
-        }
-    }
-
-    fn get_target_revset(
+    fn target_not_found_error(
         &self,
-        working_commit_id: &CommitId,
+        workspace_command: &WorkspaceCommandHelper,
         args: &MovementArgsInternal,
-    ) -> Result<Rc<RevsetExpression>, CommandError> {
-        let wc_revset = RevsetExpression::commit(working_commit_id.clone());
-        // If we're editing, start at the working-copy commit. Otherwise, start from
-        // its direct parent(s).
-        let start_revset = if args.should_edit {
-            wc_revset.clone()
-        } else {
-            wc_revset.parents()
+        commits: &[Commit],
+    ) -> CommandError {
+        let offset = args.offset;
+        let err_msg = match (self, args.should_edit, args.conflict) {
+            // in edit mode, start_revset is the WC, so we only look for direct descendants.
+            (Direction::Next, true, true) => {
+                String::from("The working copy has no descendants with conflicts")
+            }
+            (Direction::Next, true, false) => {
+                format!("No descendant found {offset} commit(s) forward from the working copy",)
+            }
+            // in non-edit mode, start_revset is the parent of WC, so we look for other descendants
+            // of start_revset.
+            (Direction::Next, false, true) => {
+                String::from("The working copy parent(s) have no other descendants with conflicts")
+            }
+            (Direction::Next, false, false) => format!(
+                "No other descendant found {offset} commit(s) forward from the working copy \
+                 parent(s)",
+            ),
+            // The WC can never be an ancestor of the start_revset since start_revset is either
+            // itself or it's parent.
+            (Direction::Prev, true, true) => {
+                String::from("The working copy has no ancestors with conflicts")
+            }
+            (Direction::Prev, true, false) => {
+                format!("No ancestor found {offset} commit(s) back from the working copy",)
+            }
+            (Direction::Prev, false, true) => {
+                String::from("The working copy parent(s) have no ancestors with conflicts")
+            }
+            (Direction::Prev, false, false) => format!(
+                "No ancestor found {offset} commit(s) back from the working copy parents(s)",
+            ),
         };
 
-        let target_revset = match self {
-            Direction::Next => if args.conflict {
-                start_revset
-                    .children()
-                    .descendants()
-                    .filtered(RevsetFilterPredicate::HasConflict)
-                    .roots()
-            } else {
-                start_revset.descendants_at(args.offset)
-            }
-            .minus(&wc_revset),
-
-            Direction::Prev => {
-                if args.conflict {
-                    // If people desire to move to the root conflict, replace the `heads()` below
-                    // with `roots(). But let's wait for feedback.
-                    start_revset
-                        .parents()
-                        .ancestors()
-                        .filtered(RevsetFilterPredicate::HasConflict)
-                        .heads()
+        let template = workspace_command.commit_summary_template();
+        let mut cmd_err = user_error(err_msg);
+        commits.iter().for_each(|commit| {
+            cmd_err.add_formatted_hint_with(|formatter| {
+                if args.should_edit {
+                    write!(formatter, "Working copy: ")?;
                 } else {
-                    start_revset.ancestors_at(args.offset)
+                    write!(formatter, "Working copy parent: ")?;
                 }
+                template.format(commit, formatter)
+            })
+        });
+
+        cmd_err
+    }
+
+    fn build_target_revset(
+        &self,
+        working_revset: &Rc<RevsetExpression>,
+        start_revset: &Rc<RevsetExpression>,
+        args: &MovementArgsInternal,
+    ) -> Result<Rc<RevsetExpression>, CommandError> {
+        let target_revset = match (self, args.conflict) {
+            (Direction::Next, true) => start_revset
+                .children()
+                .descendants()
+                .filtered(RevsetFilterPredicate::HasConflict)
+                .roots()
+                .minus(working_revset),
+            (Direction::Next, false) => start_revset
+                .descendants_at(args.offset)
+                .minus(working_revset),
+            (Direction::Prev, true) =>
+            // If people desire to move to the root conflict, replace the `heads()` below
+            // with `roots(). But let's wait for feedback.
+            {
+                start_revset
+                    .parents()
+                    .ancestors()
+                    .filtered(RevsetFilterPredicate::HasConflict)
+                    .heads()
             }
+            (Direction::Prev, false) => start_revset.ancestors_at(args.offset),
         };
 
         Ok(target_revset)
@@ -118,7 +155,17 @@ fn get_target_commit(
     working_commit_id: &CommitId,
     args: &MovementArgsInternal,
 ) -> Result<Commit, CommandError> {
-    let target_revset = direction.get_target_revset(working_commit_id, args)?;
+    let wc_revset = RevsetExpression::commit(working_commit_id.clone());
+    // If we're editing, start at the working-copy commit. Otherwise, start from
+    // its direct parent(s).
+    let start_revset = if args.should_edit {
+        wc_revset.clone()
+    } else {
+        wc_revset.parents()
+    };
+
+    let target_revset = direction.build_target_revset(&wc_revset, &start_revset, args)?;
+
     let targets: Vec<Commit> = target_revset
         .evaluate_programmatic(workspace_command.repo().as_ref())?
         .iter()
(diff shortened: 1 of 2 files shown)


cli/src/movement_util.rs
```diff
@@ -38,38 +38,55 @@ impl Direction {
...
+                if edit {
+                    // in edit mode, start_revset is the WC, so
+                    // we only look for direct descendants.
+
+                    format!(
+                        "No descendant found {} {} forward from {}",
+                        change_offset, commit_text, change_hash
+                    )
+                } else {
+                    // in none edit mode, start_revset is the parent
```
martinvonz, line 64: nit: "none edit" => "non-edit"

cli/src/movement_util.rs
```diff
@@ -122,9 +158,11 @@ fn get_target_commit(
         [target] => target,
         [] => {
             // We found no ancestor/descendant.
-            return Err(user_error(
-                direction.target_not_found_message(change_offset),
-            ));
+            return Err(user_error(direction.target_not_found_message(
+                edit,
+                change_offset,
+                &short_change_hash(start_commit.change_id()),
```
martinvonz, line 164: nit: how about passing the `&ChangeId` into the function? that makes the signature clearer

martinvonz (approved):

cli/src/movement_util.rs
```diff
@@ -110,8 +127,27 @@ fn get_target_commit(
...
+    };
+
+    let start_commit: Commit = start_revset
+        .clone()
+        .evaluate_programmatic(workspace_command.repo().as_ref())?
+        .iter()
+        .commits(workspace_command.repo().store())
+        .collect::<Result<Vec<_>, BackendError>>()?
+        .pop()
+        .unwrap();
```
PhilipMetzger, line 146: Nit: import itertools and convert into `try_collect` or whatever the function is called.

cli/src/movement_util.rs
```diff
@@ -110,8 +127,27 @@ fn get_target_commit(
...
+        wc_revset.parents()
+    };
+
+    let start_commit: Commit = start_revset
+        .clone()
+        .evaluate_programmatic(workspace_command.repo().as_ref())?
+        .iter()
+        .commits(workspace_command.repo().store())
+        .collect::<Result<Vec<_>, BackendError>>()?
+        .pop()
```
martinvonz, line 145: Better to do `.first()` than to collect into a vector and take only the first argument.

It's also a bit weird that we ignore any other parents than the first. Makes me wonder if we should just say "from the parents" instead of the specific change id.

cli/src/movement_util.rs
yuja: or use commit summary template?

essiene: > nit: how about passing the &ChangeId into the function? that makes the signature clearer

I've done this.

> or use commit summary template?
I don't know what this is. Any pointers? Thanks

cli/src/movement_util.rs
essiene: Done.

cli/src/movement_util.rs
essiene: Resolving. Do re-open if something is not quite right. Thanks.

> > nit: how about passing the &ChangeId into the function? that makes the signature clearer
> 
> I've done this.
> 
> > or use commit summary template?

Whoops! My previous comment quoted wrongly and hid my question:

> > or use commit summary template?

I don't know what this is. Any pointers? Thanks

martinvonz: It's these functions: https://github.com/martinvonz/jj/blob/bb018a54c311cc7d881dc8352d6ef100902b9096/cli/src/cli_util.rs#L1119-L1147

martinvonz: I think this PR already seems like an improvement, but I agree with Yuya that printing the commit summary seems even better, as long as we can lay it out in reasonable way when there are multiple parents. We'll presumably want to print one per line, which means we might want something like this:
```
Error: No other descendant found 1 commit(s) forward from:
  xxx aaa | parent 1
  yyy bbb | parent 2
```

FYI, the `ErrorHint` type has a `Formatted` variant for capturing color.

essiene:
```
$ jj next
Error: No other descendant found 1 commit(s) forward from:
Hint: ynsslwsp ea0238fc (empty) b2-c3
Hint: skrvwlnk 4609a180 (empty) b1-c3
```

> It's these functions:
> 
> https://github.com/martinvonz/jj/blob/bb018a54c311cc7d881dc8352d6ef100902b9096/cli/src/cli_util.rs#L1119-L1147

Thanks!

essiene: > I think this PR already seems like an improvement, but I agree with Yuya that printing the commit summary seems even better, as long as we can lay it out in reasonable way when there are multiple parents. We'll presumably want to print one per line, which means we might want something like this:
> 
> ```
> Error: No other descendant found 1 commit(s) forward from:
>   xxx aaa | parent 1
>   yyy bbb | parent 2
> ```
> 
> FYI, the `ErrorHint` type has a `Formatted` variant for capturing color.

So this actually has a 'Hint:' prefix. is that okay? So it looks like so:

 ```
 Error: No other descendant found 1 commit(s) forward from:
   Hint: xxx aaa | parent 1
   Hint: yyy bbb | parent 2
 ```

See the updated test output for details.

cli/src/movement_util.rs
```diff
@@ -118,7 +146,24 @@ fn get_target_commit(
...
+    };
+
+    let start_commits: Vec<Commit> = start_revset
+        .clone()
+        .evaluate_programmatic(workspace_command.repo().as_ref())?
+        .iter()
+        .commits(workspace_command.repo().store())
+        .try_collect()?;
+
+    let target_revset = direction.get_target_revset(&wc_revset, &start_revset, args)?;
```
yuja, line 165: nit: wc commit can be looked up without using revset.

```rust
    let wc_commit = repo.store().get_commit(&working_commit_id)?;
    wc_commit.parents()?;
```

and the resulting commit ids can be passed in to `get_target_revset()` in place of `start_revset` and `wc_revset`.

cli/src/movement_util.rs
yuja: Perhaps, the error message can be split to the error and hint lines? something like
```
Error: No other descendant found N commits forward from the working copy parent
Hint: Working copy parent: xxx aaa | ...
```

cli/src/movement_util.rs
```diff
@@ -129,7 +174,12 @@ fn get_target_commit(
         [target] => target,
         [] => {
             // We found no ancestor/descendant.
-            return Err(user_error(direction.target_not_found_message(args.offset)));
+            return Err(direction.target_not_found_error(
+                workspace_command,
+                args.should_edit,
+                args.offset,
```
yuja, line 180: nit: maybe this is an existing bug, but `args.offset` shouldn't be mentioned if `args.conflict` is set?

cli/src/movement_util.rs
essiene: Much better. I've gone ahead to structure things like this. Thanks!

cli/src/movement_util.rs
PhilipMetzger: Yes, it'd be better if this function would take `args.offset` as an optional parameter, since it's always 1 for `prev/next --conflict`.

essiene: > nit: maybe this is an existing bug, but args.offset shouldn't be mentioned if args.conflict is set?

I'm pretty sure I introduced the bug. Thanks for catching.

>Yes, it'd be better if this function would take args.offset as an optional parameter, since it's always 1 for prev/next --conflict.

I'm now also checking for the conflict arg, and replaced the if/else's with a compound match which is easier to read.

cli/src/movement_util.rs
essiene: Hmmm... `get_target_revset` needs the `start_revset` and `wc_revset` so we can run expressions that return conflicted or non-conflicted ancestors/descendants.

The API suggest we can't do that with raw `Commit`.

Or did I completely misunderstand your suggestion?

essiene: So, I'm still not getting it, I'm probably just slow. That said, I've gone ahead to rename `get_target_revset` -> `build_target_revset` and I've moved `start_commits` to be built just before we need it. I think this communicates the intents clearer. Maybe that helps us converge?

Based on that, we're not resolving `wc_revset` in `get_target_revset/build_target_revset`, we just use it when filtering children of the start_commits.

Also, I guess I could indeed resolve `start_commits`, then pass in the commits to `build_target_revset`, but then I would immediately construct a revset expression again. Though, I really don't need `start_commits` to be resolved anywhere else except for the `target_not_found_error` error message builder, so I've moved it there just before it's called.

So to check my understanding of the suggestion:
- `wc_revset` is never actually resolved, so I'm thinking no change there?
- `start_commit` is only resolved for the error message. Do you still think I should resolve, pass in the commits to `build_target_revset` and reconstruct the `start_revset` expression in there?

yuja: > I've moved `start_commits` to be built just before we need it. I think this communicates the intents clearer. Maybe that helps us converge?
> 
> Based on that, we're not resolving `wc_revset` in `get_target_revset/build_target_revset`, we just use it when filtering children of the start_commits.

Yep, new code looks better.

BTW, we might want to make `target_not_found_error()` and `build_target_revset()` look similar because these two functions tests the same conditions `(Direction, should_edit, conflict)` to build revset to be evaluated and error message of the evaluation respectively, but I don't have a good idea.

essiene: Great observation!

I've gone ahead to change `build_target_revset()` to look similar to `target_not_found_error()`, but there's a slight difference, since we use `should_edit` to get the `start_revset` outside `build_target_revset()`, before passing only `start_revset` in, so we no longer need `should_edit` in `build_target_revset()`.

I considered moving `should_edit` inside, only passing `wc_revset` alone and then having different revset chains, but it felt a bit too forced, especially since the chains look exactly the same, so I didn't go that route.

This definitely looks better though.

Thanks

cli/src/movement_util.rs
```diff
@@ -59,52 +59,110 @@ impl Direction {
...
+        let err_msg = match (self, args.should_edit, args.conflict) {
+            // in edit mode, start_revset is the WC, so
+            // we only look for direct descendants.
+
+            // edit, conflict
+            (Direction::Next, true, true) => {
+                String::from("The working copy has no descendants with conflicts")
+            }
+
+            // edit, no-conflict
```
yuja, line 77: nit: I would remove these `(no-)edit, (no-)conflict` comments and blank lines so the match source expression are placed closer.

You can also assign `args.offset` to temporary so `"{offset}"` can be inlined.

cli/src/movement_util.rs
```diff
@@ -129,7 +197,13 @@ fn get_target_commit(
         [target] => target,
         [] => {
             // We found no ancestor/descendant.
-            return Err(user_error(direction.target_not_found_message(args.offset)));
+            let start_commits: Vec<Commit> = start_revset
+                .clone()
```
yuja, line 201: nit: `.clone()` is redundant

cli/tests/test_next_prev_commands.rs
```diff
@@ -569,9 +572,45 @@ fn test_prev_prompts_on_multiple_parents() {
...
+    ╭─┼─╮
+    ○ │ │  qpvuntsmwlqt first
+    │ │ ○  kkmpptxzrspx second
+    ├───╯
+    │ ○  mzvwutvlkqwt third
+    ├─╯
+    ◆  zzzzzzzzzzzz
+    "###);
+
+    _ = test_env.jj_cmd_ok(&repo_path, &["next"]);
```
yuja, line 592: nit: `_ = ` can be omitted

CHANGELOG.md
```diff
@@ -36,6 +36,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* Fix confusing/unclear error message, when next/prev can not find a target commit
+  by explicitly specifying the commit where the search for the target started from.
```
yuja, line 40: nit: I wouldn't add changelog entry because users don't have to know error/warning messages are improved.

yuja (approved):
Thanks!

CHANGELOG.md
essiene: Removed.

cli/tests/test_next_prev_commands.rs
essiene: Removed.

cli/src/movement_util.rs
essiene: Ahh! Removed.

cli/src/movement_util.rs
essiene: Done.

> nit: I would remove these `(no-)edit, (no-)conflict` comments and blank lines so the match source expression are placed closer.
> 

Done.

> You can also assign `args.offset` to temporary so `"{offset}"` can be inlined.

Done.

PR #4286, opened 2024-08-16
next/prev: Fetch descendants with more correctness

essiene:
See context in [this discussion](https://github.com/martinvonz/jj/pull/3935#discussion_r1649520967)

This is the last planned change for this issue, so marking this as the fixing PR.

Fixes https://github.com/martinvonz/jj/issues/3947

# Checklist

If applicable:

### DIFF
--- a/cli/src/movement_util.rs
+++ b/cli/src/movement_util.rs
@@ -121,27 +121,27 @@ impl Direction {
         start_revset: &Rc<RevsetExpression>,
         args: &MovementArgsInternal,
     ) -> Result<Rc<RevsetExpression>, CommandError> {
-        let target_revset = match (self, args.conflict) {
-            (Direction::Next, true) => start_revset
+        let nth = match (self, args.should_edit) {
+            (Direction::Next, true) => start_revset.descendants_at(args.offset),
+            (Direction::Next, false) => start_revset
                 .children()
+                .minus(working_revset)
+                .descendants_at(args.offset - 1),
+            (Direction::Prev, _) => start_revset.ancestors_at(args.offset),
+        };
+
+        let target_revset = match (self, args.conflict) {
+            (_, false) => nth,
+            (Direction::Next, true) => nth
                 .descendants()
                 .filtered(RevsetFilterPredicate::HasConflict)
-                .roots()
-                .minus(working_revset),
-            (Direction::Next, false) => start_revset
-                .descendants_at(args.offset)
-                .minus(working_revset),
-            (Direction::Prev, true) =>
+                .roots(),
             // If people desire to move to the root conflict, replace the `heads()` below
             // with `roots(). But let's wait for feedback.
-            {
-                start_revset
-                    .parents()
-                    .ancestors()
-                    .filtered(RevsetFilterPredicate::HasConflict)
-                    .heads()
-            }
-            (Direction::Prev, false) => start_revset.ancestors_at(args.offset),
+            (Direction::Prev, true) => nth
+                .ancestors()
+                .filtered(RevsetFilterPredicate::HasConflict)
+                .heads(),
         };
 
         Ok(target_revset)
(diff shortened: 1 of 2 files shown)


cli/src/movement_util.rs
yuja, line 1: Can you also add some tests?

cli/src/movement_util.rs
```diff
@@ -109,16 +109,20 @@ impl Direction {
...
+                        .children()
+                        .minus(working_revset)
+                        .descendants()
+                        .filtered(RevsetFilterPredicate::HasConflict)
+                        .roots()
+                } else {
+                    start_revset
+                        .descendants_at(args.offset)
+                        .minus(working_revset)
+                }
```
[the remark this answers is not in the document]
yuja: BTW, I think the query can be reorganized as
```rust
let next = if !args.edit {
    start.children().minus(wc)
} else {
    start.children()
};
if args.conflict {
    next.descendants().filtered(HasConflict).roots()
} else {
    next.descendants_at(N - 1)
}
```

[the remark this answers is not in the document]
essiene: Thanks, this was helpful. I'm finally getting used to revsets!

I've folded this in, though for consistency, I've used to match expressions so that it's clearer how `Direction::Prev` is treated. 

Using only one match expressions ends up with too much duplicated code that is just so slightly different, but a single match seems just right.

cli/src/movement_util.rs
essiene: Done. I think I got all the interesting corner cases.

cli/tests/test_next_prev_commands.rs
```diff
@@ -1112,6 +1112,405 @@ fn test_movement_edit_mode_false() {
     "###);
 }
 
+#[test]
+fn test_movement_target_revset() {
```
yuja, line 1116: Some of the added tests look redundant. What we need to test in this patch is `movement.edit = false` but the wc commit has children (i.e. currently editing), right?

yuja (commented):
The code change looks good to me, thanks.

cli/tests/test_next_prev_commands.rs
essiene: Good point. I got carried away and added test that would already be covered by other tests. D'oh!

I've pruned most of them now. In particular, I've remove `edit = true` and any `prev` related stanzas.

PTAL.

yuja: You'll need to add tests in which `@` isn't a head. Try run the added tests against the previous commit to see if they fail.

essiene: The only case I've found which isn't covered existing tests is the case where all the following hold.

* edit=false
* wc is not a head
* no conflict
* offset > 1 (I still added a test, but it's not really interesting)

I've added a test for that and confirm that it fails without the code change in this PR. See #4484 

The conflict case doesn't have this problem in trunk; when `edit=false` since we end up taking the `root()` conflicted descendants and then removing the wc.

Let me know if there's any other edge case that would make sense to explicitly test for.

cli/src/movement_util.rs
yuja, line 1: nit: can you add `next/prev: fetch descendants ...` to the commit description header?

cli/tests/test_next_prev_commands.rs
yuja: Thanks. I think we could add test in which both sides have conflicts 
(you'll get prompt without this patch), but that's not so important. I think the added test is good enough.

yuja (approved):
Thanks!

cli/tests/test_next_prev_commands.rs
essiene: Thanks. Resolving now.

cli/src/movement_util.rs
essiene: Done.

PR #4287, opened 2024-08-16
Fix macOS x86_64 release builds

Cretezy:
Seems like this broke for the v0.19 release and later.

`macos-11` is no longer supported: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

`macos-13` is the last x86_64 macOS version

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
           os: ubuntu-24.04
           target: aarch64-unknown-linux-musl
         - build: macos-x86_64
-          os: macos-11
+          os: macos-13
           target: x86_64-apple-darwin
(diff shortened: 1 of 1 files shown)


ilyagr:
See also #4243. @thoughtpolice, is there any reason not to merge that PR? We could also have somebody else merge it if it's easier.

thoughtpolice:
I just didn't have time to fix the other workflows to consistently use macos-13, which is why it was still draft. I'll try to get to it this morning

thoughtpolice:
Fixed by https://github.com/martinvonz/jj/pull/4243, thanks for nudging me on this topic!

PR #4289, opened 2024-08-16
tests: restore snapshots of line/word-oriented diff hunks

yuja:
# Checklist

If applicable:


martinvonz (approved):
I was a bit concerned that we were losing test coverage when I was reviewing some of those PRs but I didn't look closely enough. Thanks.

PR #4296, opened 2024-08-17
drop "support" for legacy tree config

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -124,8 +124,6 @@ pub struct GitBackend {
     empty_tree_id: TreeId,
     extra_metadata_store: TableStore,
     cached_extra_metadata: Mutex<Option<Arc<ReadonlyTable>>>,
-    /// Whether tree of imported commit should be promoted to non-legacy format.
-    imported_commit_uses_tree_conflict_format: bool,
 }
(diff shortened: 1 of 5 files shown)


lib/src/git_backend.rs
```diff
@@ -190,14 +190,23 @@ impl GitBackend {
...
         git_repo_path: &Path,
+    ) -> Result<Self, Box<GitBackendInitError>> {
+        Self::init_external_with_tree_conflicts(settings, store_path, git_repo_path, true)
+    }
+
+    fn init_external_with_tree_conflicts(
+        settings: &UserSettings,
+        store_path: &Path,
+        git_repo_path: &Path,
+        imported_commit_uses_tree_conflict_format: bool,
```
yuja, line 209: nit: maybe we can instead extract `git_backend.import_head_commits_with_tree_conflicts(.., uses_tree_conflict_format: bool)` for tests.

(can be addressed separately)

yuja (approved):

lib/src/git_backend.rs
martinvonz: Ah, that's much cleaner, thanks! Done in this PR.

PR #4297, opened 2024-08-18
Follow ups on post #4282 submission comments.

essiene:
Follow up on post-submission comments in #4282.

* Derive a bunch of standard and useful traits for `movement_util::Direction`
  as it is a simple type. Importantly `Copy`.
* Return `&'static str` from Direction.cmd()
* Refactor out `MovementArgs` to reduce the number of arguments
  to `movement_util::move_to_commit`.
* Implement `From<&NextArgs/&PrevArgs>` for MovementArgs

Part of #3947

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/next.rs
+++ b/cli/src/commands/next.rs
@@ -14,7 +14,7 @@
 
 use crate::cli_util::CommandHelper;
 use crate::command_error::CommandError;
-use crate::movement_util::{move_to_commit, Direction};
+use crate::movement_util::{move_to_commit, Direction, MovementArgs};
 use crate::ui::Ui;
 
 /// Move the working-copy commit to the child revision
@@ -64,6 +64,16 @@ pub(crate) struct NextArgs {
     conflict: bool,
 }
 
+impl From<&NextArgs> for MovementArgs {
+    fn from(val: &NextArgs) -> Self {
+        MovementArgs {
+            offset: val.offset,
+            edit: val.edit,
+            conflict: val.conflict,
+        }
+    }
+}
+
 pub(crate) fn cmd_next(
     ui: &mut Ui,
     command: &CommandHelper,
@@ -73,9 +83,7 @@ pub(crate) fn cmd_next(
     move_to_commit(
         ui,
         &mut workspace_command,
-        &Direction::Next,
-        args.edit,
-        args.conflict,
-        args.offset,
+        Direction::Next,
+        &MovementArgs::from(args),
     )
 }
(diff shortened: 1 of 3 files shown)


cli/src/movement_util.rs
```diff
@@ -159,39 +160,29 @@ fn choose_commit<'a>(
...
 
-pub fn move_to_commit(
+pub(crate) fn move_to_commit(
     ui: &mut Ui,
     workspace_command: &mut WorkspaceCommandHelper,
     direction: &Direction,
-    edit: bool,
-    has_conflict: bool,
-    change_offset: u64,
+    args: &mut MovementArgs,
```
yuja, line 167: nit: It's weird that this function takes input parameter as `&mut`. Let's clone `args` in this function instead. This `.clone()` will be removed if we drop support for the "auto" behavior.

cli/src/commands/next.rs
```diff
@@ -74,8 +84,6 @@ pub(crate) fn cmd_next(
         ui,
         &mut workspace_command,
         &Direction::Next,
```
yuja, line 86: nit: pass `Direction::Next` by value (applies to all callers)

cli/src/movement_util.rs
```diff
@@ -25,16 +25,24 @@ use crate::cli_util::{short_commit_hash, WorkspaceCommandHelper};
 use crate::command_error::{user_error, CommandError};
 use crate::ui::Ui;
 
-pub enum Direction {
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) struct MovementArgs {
```
yuja, line 29: nit: I wouldn't add `Copy` here as we might add more parameters to this struct.

cli/src/movement_util.rs
```diff
@@ -25,16 +25,24 @@ use crate::cli_util::{short_commit_hash, WorkspaceCommandHelper};
...
+    pub(crate) offset: u64,
+    pub(crate) edit: bool,
+    pub(crate) conflict: bool,
```
yuja, line 32: nit: `(crate)` can be omitted since outer type is `pub(crate)`.

yuja (approved):
Thanks!

cli/src/movement_util.rs
essiene: Ah thanks. This is better. Done.

cli/src/commands/next.rs
essiene: Lol... I'd forgotten I'd derived Copy. Thanks. Fixed.

cli/src/movement_util.rs
essiene: Makes sense. Done. Thanks.

cli/src/movement_util.rs
essiene: I didn't know that. Done. Thanks.

PR #4298, opened 2024-08-18
diff: add option to display complex color-words diffs without inlining

yuja:
# Checklist

If applicable:

---

original (inline everything):
image

`max-inline-alternation = 3`:
image

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -200,6 +200,29 @@ can override the default style with the following keys:
 ui.diff.format = "git"
 ```
 
+#### Color-words diff options
+
+In color-words diffs, changed words are displayed inline by default. Because
+it's difficult to read a diff line with many removed/added words, there's a
+threshold to switch to traditional separate-line format.
+
+* `max-inline-alternation`: Maximum number of removed/added word alternation to
+  inline. For example, `<added> ... <added>` sequence has 1 alternation, so the
+  line will be inline if `max-inline-alternation >= 1`. `<added> ... <removed>
+  ... <added>` sequence has 3 alternation.
+
+  * `0`: disable inlining, making `--color-words` more similar to `--git`
+  * `1`: inline removes-only or adds-only lines
+  * `2`, `3`, ..: inline up to `2`, `3`, .. alternation
+  * `-1`: inline all lines (default)
+
+  **This parameter is experimental.** The definition is subject to change.
+
+```toml
+[diff.color-words]
+max-inline-alternation = 3
+```
+
 ### Generating diffs by external command
 
 If `ui.diff.tool` is set, the specified diff command will be called instead of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `--color-words`, `--git`, `--stat`, `--summary`, `--types`, and external diff
   tools in file-by-file mode.
 
+* Color-words diff has gained [an option to display complex changes as separate
+  lines](docs/config.md#color-words-diff-options).
+
 * A tilde (`~`) at the start of the path will now be expanded to the user's home
   directory when configuring a `signing.key` for SSH commit signing.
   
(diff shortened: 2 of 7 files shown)


ilyagr:
This looks really cool!

I'm worried it might look worse with the default colors (that have underlines). WDYT? Could you make a screenshot?

If we think this is a problem, one possible workaround would be to mention suggested 256color config in the docs next to the new option. (Unfortunately, I am still worried about using 256 colors by default, since absolute color values would look bad for some terminal color schemes, unless we have some ideas about that. For example, if we'd want to test them with a bunch of terminals' default light/dark color schemes, there are quite a few, and it'd be an annoying amount of work. Off the top of my head, xterm, GNOME, and VS Code terminal are likely all different; I'm not sure about Terminal.app, Windows Terminal, iTerm2, ???. I suppose we could collect the list of "all the themes we care about", but I don't want to be the person doing it.)

If it's not a problem, we should consider making the new option the default in a separate PR.

docs/config.md
```diff
@@ -200,6 +200,27 @@ can override the default style with the following keys:
...
 ```
 
+#### Color-words diff options
+
+In color-words diffs, changed words are displayed inline by default. Because
+it's difficult to read a diff line with many removed/added words, there's a
+threshold to switch to traditional separate-line format.
+
+* `max-inline-alternation`: Maximum number of removed/added word alternation to
+  inline. For example, `<added> <added>` sequence has 1 alternation, so the line
```
ilyagr, line 210: Does `<added> <added>` make sense? Did you mean `<added> unchanged <added>`?

docs/config.md
```diff
@@ -200,6 +200,27 @@ can override the default style with the following keys:
...
+  sequence has 3 alternation.
+
+  * `0`: disable inlining
+  * `1`: inline removes-only or adds-only lines
+  * `2`, `3`, ..: inline up to `2`, `3`, .. alternation
+  * `-1`: inline all lines (default)
+
+```toml
+[diff.color-words]
+max-inline-alternation = 3
```
ilyagr, line 221: I'm wondering if this config could be something simpler.

Perhaps

```toml
inline = true
inline-max-complexity = 3
```

For the docs, we could first explain the intuition (your 0,1,... examples), and then have a section with the details.

Another thought:

docs/config.md
```diff
@@ -200,6 +200,27 @@ can override the default style with the following keys:
...
+In color-words diffs, changed words are displayed inline by default. Because
+it's difficult to read a diff line with many removed/added words, there's a
+threshold to switch to traditional separate-line format.
+
+* `max-inline-alternation`: Maximum number of removed/added word alternation to
+  inline. For example, `<added> <added>` sequence has 1 alternation, so the line
+  will be inline if `max-inline-alternation >= 1`. `<added> <removed> <added>`
+  sequence has 3 alternation.
+
+  * `0`: disable inlining
```
ilyagr, line 214: Perhaps add:

> ..., making `diff --color-words` more similar to `diff --git`.

ilyagr:
I am also slightly worried about diffs where lines that reach the complexity threshold alternate with lines that don't. For example, what if in your example, some of the shorter lines in the long changed comment happened to only have additions? Would it still look clear?

(I haven't checked whether you have logic to avoid this problem. If this is a problem, it can probably be considered something to refine in a future PR, unless we can think of some other algorithm that avoids the problem entirely)

ilyagr:
In general, my comments so far are not meant to be a formal review, just thoughts. I'm not sure which, if any, of them are blockers; we could always go with any version of this with the intention of trying it as an experiment, seeing how it goes, and adjusting things later.

yuja:
> I'm worried it might look worse with the default colors (that have underlines). WDYT? Could you make a screenshot?

Could look worse, but the result is basically the same as `diff --git` (where leading "-" and "+" are replaced with line numbers.)

> I am also slightly worried about diffs where lines that reach the complexity threshold alternate with lines that don't. For example, what if in your example, some of the shorter lines in the long changed comment happened to only have additions? Would it still look clear?

Inline-or-not is determined per hunk, so contiguous changed lines won't be rendered in different styles.

docs/config.md
yuja: I assume we'll settle on some better default, and users rarely customize this parameter. It's hard to explain what the "complexity" (or "alternation") means.

martinvonz (approved):
This looks very useful to me

docs/config.md
yuja: Inserted ellipses, which I think can be read as repetition (plus unchanged if the user is aware of that.)

docs/config.md
yuja: Marked this parameter as experimental. If we find a better default and formalization, we can deprecate the current one.

PR #4300, opened 2024-08-18
Add separate stream item type with copy info

martinvonz:
@fowles, I think this is a step on the way to making the behavior around deleted rename sources more consistent.

# Checklist

If applicable:

### DIFF
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -18,6 +18,7 @@ use std::ops::Range;
 use std::path::{Path, PathBuf};
 use std::{io, mem};
 
+use futures::stream::BoxStream;
 use futures::StreamExt;
 use itertools::Itertools;
 use jj_lib::backend::{BackendError, TreeValue};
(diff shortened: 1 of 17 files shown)


cli/src/diff_util.rs
```diff
@@ -303,7 +306,8 @@ impl<'a> DiffRenderer<'a> {
                     )?;
                 }
                 DiffFormat::NameOnly => {
-                    let tree_diff = from_tree.diff_stream(to_tree, matcher, copy_records);
+                    let tree_diff: BoxStream<CopiesTreeDiffEntry> =
```
fowles, line 309: do you need the type here?

lib/src/local_working_copy.rs
```diff
@@ -1353,23 +1352,16 @@ impl TreeState {
...
+                    match diff {
+                        Ok((before, after)) => {
+                            let result = materialize_tree_value(&self.store, &path, after).await;
+                            (path, result.map(|value| (before.is_present(), value)))
                         }
-                    },
-                )
+                        Err(err) => (path, Err(err)),
+                    }
+                })
```
fowles, line 1364: both sides of this map construct a pair with `(path, ...)`.  I suspect you can implement this without using `match` at all and instead use `diff.map(|(before, after)| ...)`

lib/src/local_working_copy.rs
```diff
@@ -1452,16 +1444,10 @@ impl TreeState {
...
         let mut changed_file_states = Vec::new();
         let mut deleted_files = HashSet::new();
-        let mut diff_stream = old_tree.diff_stream(new_tree, matcher.as_ref(), &copy_records);
-        while let Some(TreeDiffEntry {
-            source: _, // TODO handle copy tracking
-            target: path,
-            value: diff,
-        }) = diff_stream.next().await
-        {
+        let mut diff_stream = old_tree.diff_stream(new_tree, matcher.as_ref());
```
fowles, line 1449: the TODO went away here. Not sure if you wanted to keep it or not

fowles (approved):

cli/src/diff_util.rs
martinvonz: no, and i don't know why i added it. removed now.

lib/src/local_working_copy.rs
martinvonz: I removed it because i don't think the working copy needs to care about copies. We *could* implement updating across a rename as a move in the file system, but it adds complexity and probably makes things slower in most cases.

lib/src/local_working_copy.rs
martinvonz: actually not because of the `.await` (this is in an `async` block)

PR #4308, opened 2024-08-20
Added docs about installation under Gentoo Linux

Raniz85:
There is an ebuild available in the Gentoo GURU repository (the official project for user-contributed packages).

I added information about how to get Jujutsu via the GURU repo to the installation guide.


martinvonz (approved):
Thanks

PR #4313, opened 2024-08-21
copies: filter rename source entries by CopiesTreeDiffStream

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/copies.rs
+++ b/lib/src/copies.rs
@@ -29,8 +29,9 @@ use crate::repo_path::{RepoPath, RepoPathBuf};
 #[derive(Default, Debug)]
 pub struct CopyRecords {
     records: Vec<CopyRecord>,
-    // Maps from `target` to the index of the target in `records`.  Conflicts
-    // are excluded by keeping an out of range value.
+    // Maps from `source` or `target` to the index of the entry in `records`.
+    // Conflicts are excluded by keeping an out of range value.
+    sources: HashMap<RepoPathBuf, usize>,
     targets: HashMap<RepoPathBuf, usize>,
 }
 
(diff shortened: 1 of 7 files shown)


lib/src/copies.rs
```diff
@@ -121,28 +121,26 @@ impl Stream for CopiesTreeDiffStream<'_> {
...
-                    target: diff_entry.path,
-                    value: diff_entry.value.and_then(|(_, target_value)| {
-                        self.source_tree
-                            .path_value(source)
-                            .map(|source_value| (source_value, target_value))
-                    }),
-                }
-            })
-        })
+        let Some(diff_entry) = ready!(self.inner.as_mut().poll_next(cx)) else {
```
martinvonz: This looked quite confusing to me until I read that `ready!` can return early. That's a bit surprising to me (it feels like it should be called `return_unless_ready!` or something. I suppose I'll get used to it, but it's very confusing until you know what it does.

So I wonder if it's worth it, which I guess depends on what this commit does. I didn't understand that either. I figured the two `.map()` calls (on `task::Poll` and on `Option`) would have the same effect. What's the difference?

lib/src/copies.rs
```diff
@@ -121,26 +121,33 @@ impl Stream for CopiesTreeDiffStream<'_> {
...
-            return Poll::Ready(None);
-        };
+        while let Some(diff_entry) = ready!(self.inner.as_mut().poll_next(cx)) {
+            let Some(CopyRecord { source, .. }) = self.copy_records.for_target(&diff_entry.path)
+            else {
+                let target_deleted =
+                    matches!(&diff_entry.value, Ok((_, target_value)) if target_value.is_absent());
+                if target_deleted && self.copy_records.has_source(&diff_entry.path) {
+                    // Skip the "delete" entry when there is a rename.
+                    continue;
```
martinvonz, line 131: FYI (i.e. not requesting any changes), I was going to make `CopiesTreeDiffEntry` an enum with `Regular`, `Copied`, `Renamed`, and `SourceDeleted` variants. I guess we don't even need the last variant, or not yet at least.

yuja: I was thinking of a similar change: `Regular(path)|Copied(source, dest)|Renamed(source, dest)`.

lib/src/copies.rs
yuja: The reason behind this change is to wrap the whole block within a `while let _ = ready!(..)` loop. Alternatively, we can use `loop { match self.inner.as_mut().poll_next(cx) { .. return; .. }; .. }`.

Perhaps, `ready!()` would make sense before when `try!(..)` was a thing.

martinvonz: > The reason behind this change is to wrap the whole block within a `while let _ = ready!(..)` loop. Alternatively, we can use `loop { match self.inner.as_mut().poll_next(cx) { .. return; .. }; .. }`.

I see. The "return early" made me think we currently did some unnecessary processing.

> Perhaps, `ready!()` would make sense before when `try!(..)` was a thing.

Yes, that's what I thought after I learnt about the hidden early return.

martinvonz (approved):

lib/src/copies.rs
yuja: Updated commit message and switched to `std::task::ready!()`.

PR #4320, opened 2024-08-22
docs: Add FAQ for keeping changes to tracked files

jennings:
(I'm fine with having this whole PR changed or even rejected. Since this is a docs change, I figured it was easiest to just make the PR rather than trying to pre-approve it in chat.)

The question "How do I avoid committing changes to files?" comes up a lot in chat, and the solution is not obvious. It will be useful to have a description with an example we can link to.

The wording of the similar question "How can I keep my scratch files in the repository?" was tweaked to emphasize the difference between keeping untracked files in the workspace and keeping changes tracked files out of published history.

# Checklist

If applicable:

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -75,6 +75,10 @@ Then future edits will go into a new working-copy commit on top of the now
 former working-copy commit. Whenever you are happy with another set of edits,
 use `jj squash` to amend the previous commit.
 
+If you have changes you _never_ want to put in a public commit, see: [How can I
+keep my scratch files in the repository without committing
+them?](#how-can-i-keep-my-scratch-files-in-the-repository-without-committing-them)
+
 For more options see the next question.
 
 ### Can I interactively create a new commit from only some of the changes in the working copy, like `git add -p && git commit` or `hg commit -i`?
@@ -95,12 +99,16 @@ which may require multiple invocations of `jj rebase -r` or `jj rebase -s`.
 
 To squash or split commits, use `jj squash` and `jj split`.
 
-### How can I keep my scratch files in the repository?
+### How can I keep my scratch files in the repository without committing them?
 
 You can keep your notes and other scratch files in the repository, if you add
 a wildcard pattern to either the repo's `gitignore` or your global `gitignore`.
 Something like `*.scratch` or `*.scratchpad` should do, after that rename the
 files you want to keep around to match the pattern.
+
+If you keep your scratch files in their own directory with no tracked files, you
+can create a `.gitignore` file in that directory containing only `*`. This will
+ignore everything in the directory including the `.gitignore` file itself.
 
 If `$EDITOR` integration is important, something like `scratchpad.*` may be more
 helpful, as you can keep the filename extension intact (it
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -111,6 +111,71 @@ store arbitrary files in `<your-git-repo>/scratch/`.
 
 You can find more details on `gitignore` files [here][gitignore].
 
+### How can I avoid committing my local-only changes to tracked files?
```
PhilipMetzger, line 114: Nit: Shouldn't this be "How can I keep local changes private?"

And then the introduction should mention the use-cases of configs and/or build files.

docs/FAQ.md
```diff
@@ -95,7 +95,7 @@ which may require multiple invocations of `jj rebase -r` or `jj rebase -s`.
 
 To squash or split commits, use `jj squash` and `jj split`.
 
-### How can I keep my scratch files in the repository?
+### How can I create scratch files without comitting them to the repository?
```
PhilipMetzger, line 98: Nit: I think the previous title was clearer, since it implied that it works _also_ for adding files.

docs/FAQ.md
```diff
@@ -111,6 +111,71 @@ store arbitrary files in `<your-git-repo>/scratch/`.
...
 
+### How can I avoid committing my local-only changes to tracked files?
+
+Suppose your repository tracks a file like `config.json`, and you make some
+changes to that file to work locally. Since Jujutsu automatically commits the
+working copy, there's no way to Jujutsu from committing changes to the file.
+But, you never want to push those changes to the remote repository. 
+
+One solution is to keep these changes in a separate commit branched from the
+trunk and _merge_ the private commit into your branch to use those changes.
```
PhilipMetzger, line 122: Nit: If we use a merge commit here, we need to discuss that the _private commit_ is the second parent of your work. If we just say, that this should just be a parent, I think the whole section could be simplified.

PhilipMetzger (changes requested):
While I don't think this will be rejected, as it _certainly_ got popular as a question. There's just some word forgery and simplifications left.

docs/FAQ.md
```diff
@@ -111,6 +111,71 @@ store arbitrary files in `<your-git-repo>/scratch/`.
 
 You can find more details on `gitignore` files [here][gitignore].
 
+### How can I avoid committing my local-only changes to tracked files?
+
+Suppose your repository tracks a file like `config.json`, and you make some
+changes to that file to work locally. Since Jujutsu automatically commits the
+working copy, there's no way to Jujutsu from committing changes to the file.
```
ilyagr, line 118: typo: Need a word like `prevent` between `to` and `Jujutsu`

docs/FAQ.md
```diff
@@ -111,6 +111,71 @@ store arbitrary files in `<your-git-repo>/scratch/`.
 
 You can find more details on `gitignore` files [here][gitignore].
 
+### How can I avoid committing my local-only changes to tracked files?
+
+Suppose your repository tracks a file like `config.json`, and you make some
```
ilyagr, line 116: Maybe `secret_config.json` is slightly better?

docs/FAQ.md
```diff
@@ -111,6 +111,71 @@ store arbitrary files in `<your-git-repo>/scratch/`.
...
+│ │  private: my credentials
+◉ │  mmzpqlum me@example.com 2024-08-21 11:13:21 ef612875
+├─╯  Add new feature
+◉  tvxsnxqm me@example.com 2024-08-21 11:13:09 main b624cf12
+│  Existing work
+~
+```
+
+The revision _ttsqqnrx_ contains changes from both _mmzpqlum_ and _ttsqqnrx_. As
+you make changes you want to keep squash them into change _mmzpqlum_ using `jj
```
ilyagr, line 169: The word `keep` is out of place.

You could also write 

> As you make changes, you can squash them using `jj squash --into mmzpqlum`. Or, you can `jj rebase -r @ -d mmzpqlum`, `jj describe`, and then `jj new @ ttsqqnrx` to create a new merge commit to work in.

docs/FAQ.md
ilyagr: What do you mean by adding files?

At a glance, either title seems OK to me, but I also like the idea of combining them:

> How can I keep my scratch files in the working copy without committing them to the repository?

ilyagr, line 98: Aside: This is not strictly related to this PR, but there is another trick for this I found. We're actually using it to not track `.jj`. I found it useful to not track `.vscode` in the `jj` repo without ignoring it in all repos (some repos put important workflow config in .vscode). 

You can create any dir with .gitignore containing `*` and it won't be tracked.

Or, if you want to be really surgical, you can probably put a `.gitignore` in any of the existing dirs that don't already have one and put

```
.gitignore
my_scratch_file
```

in it.

ilyagr:
Thank you! I think this is useful to add. 

I was mildly worried it might be distracting (and if others people feel stronger about this, feel free to chime in), but if the FAQ gets too long, we could always split it into sections with these two questions in its own section.

docs/FAQ.md
jennings: I think phrasing questions as someone unfamiliar with jj would ask them might make the FAQ easiest to navigate. I only pondered it for a couple seconds while writing this commit, but my concern with "working copy" was:

- I don't _think_ most developers use that term. They may recognize it in context, but wouldn't use it themselves.
- It might be unclear that we're referring to "the directory tree of the workspace" rather than "the working copy commit".

I'm not attached to any particular phrasing though.

docs/FAQ.md
PhilipMetzger: Here's a suggestion: 
"As long as your private commit is a parent of your work, it's changes will be reflected in the current commit and the working copy. As long as you keep this commit up to date, you can add any local changes there. "

docs/FAQ.md
```diff
@@ -95,7 +95,7 @@ which may require multiple invocations of `jj rebase -r` or `jj rebase -s`.
 
 To squash or split commits, use `jj squash` and `jj split`.
 
-### How can I keep my scratch files in the repository?
+### How can I create scratch files without committing them to the repository?
```
PhilipMetzger, line 98: Nit: "keep or create [...]", since initially it was all about keeping Git's untracked files available.

docs/FAQ.md
```diff
@@ -111,6 +111,79 @@ store arbitrary files in `<your-git-repo>/scratch/`.
...
+│ ◉  ttsqqnrx me@example.com 2024-08-22 08:57:40 2106921e
+│ │  private: my credentials
+◉ │  mmzpqlum me@example.com 2024-08-21 11:13:21 ef612875
+├─╯  Add new feature
+◉  tvxsnxqm me@example.com 2024-08-21 11:13:09 main b624cf12
+│  Existing work
+~
+```
+
+The revision _ttsqqnrx_ contains changes from both _mmzpqlum_ and _ttsqqnrx_. As
```
PhilipMetzger, line 168: Nit: This is imo, to short of a explainer, see my suggestion for clarification.

PhilipMetzger (approved):
LGTM, I'd wait for another round of feedback from Ilya though.

docs/FAQ.md
PhilipMetzger: My opinion on that is similar, see https://github.com/martinvonz/jj/pull/4320#discussion_r1729209626

ilyagr:
This PR might become out of date once/if https://github.com/martinvonz/jj/pull/4338/files goes in.

ilyagr:
Actually, on second thought, only the tiny part of this PR that has to do with the "scratch files" question overlaps with #4338, so I may have spoken too soon..

docs/FAQ.md
```diff
@@ -111,6 +111,79 @@ store arbitrary files in `<your-git-repo>/scratch/`.
 
 You can find more details on `gitignore` files [here][gitignore].
 
+### How can I avoid committing my local-only changes to tracked files?
```
ilyagr, line 114: I wonder if the "Can I prevent Jujutsu from recording my unfinished work? I'm not ready to commit it." question should be combined with this one or at least link to this one.

docs/FAQ.md
jennings: I worry that might be too much information before the reader sees what we're talking about.

I split this paragraph into two simpler sentences that are hopefully easier to read. At the bottom of the demo, I added some bullet points talking about the first and second parents. What do you think? Does this help?

docs/FAQ.md
jennings: I added a link in that question, thanks

docs/FAQ.md
jennings: I added some bullet points here, does that help?

I also edited the change IDs to _yyyyyyy_, _xxxxxxx_, etc., to help visually distinguish them in the paragraphs.

jennings:
I just noticed that the _next_ question already in the FAQ is "How can I keep local changes around, but not use them for Pull Requests?". That answer recommends doing either:

1. Rebasing the branch on top of the private commit, then rebasing back on top of main before pushing
2. Rebase the private commit on top of the branch, then `jj new`

This PR gives a third option, creating a merge commit. I think this is too many options for someone relatively new to Jujutsu, I think we should be more prescriptive until they are comfortable. So I _removed_ the original question entirely. I'm open to changing this decision if people don't like it.

docs/FAQ.md
PhilipMetzger: > What do you think? Does this help?

It seems good enough that I no longer see a reason to contest it.

ilyagr (approved):
This looks pretty good to me. Thank you!

> I just noticed that the next question already in the FAQ is "How can I keep local changes around, but not use them for Pull Requests?". 

Replacing that question seems to me like it works quite well.

docs/FAQ.md
jennings: I added a paragraph about this, thanks.

PR #4327, opened 2024-08-23
Support copy/renames in `jj status`

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/status.rs
+++ b/cli/src/commands/status.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use itertools::Itertools;
+use jj_lib::copies::CopyRecords;
 use jj_lib::repo::Repo;
 use jj_lib::revset::RevsetExpression;
 use jj_lib::revset::RevsetFilterPredicate;
@@ -21,6 +22,7 @@ use tracing::instrument;
 use crate::cli_util::print_conflicted_paths;
 use crate::cli_util::CommandHelper;
 use crate::command_error::CommandError;
+use crate::diff_util::get_copy_records;
 use crate::diff_util::DiffFormat;
 use crate::revset_util;
 use crate::ui::Ui;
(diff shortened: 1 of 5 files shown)


cli/src/commands/status.rs
```diff
@@ -67,6 +69,12 @@ pub(crate) fn cmd_status(
             writeln!(formatter, "The working copy is clean")?;
         } else {
             writeln!(formatter, "Working copy changes:")?;
+            let mut copy_records = CopyRecords::default();
+            for parent in wc_commit.parents() {
```
bnjmnt4n, line 73: This could probably be `wc_commit.parent_ids()`, unless the intent was to cache the commit?

bnjmnt4n (approved):

cli/tests/test_status_command.rs
```diff
@@ -36,6 +36,33 @@ fn create_commit(
...
+    std::fs::write(repo_path.join("rename-target1"), "rename").unwrap();
+    std::fs::write(repo_path.join("rename-target2"), "rename").unwrap();
+
+    let stdout = test_env.jj_cmd_success(&repo_path, &["status"]);
+    insta::assert_snapshot!(stdout, @r###"
+    Working copy changes:
+    A copy-target1
+    A copy-target2
+    R {rename-source => rename-target1}
+    A rename-target2
```
yuja, line 60: nit: might be better to simplify the test case to not heavily depend on the git backend behavior.

"copy-target"s aren't detected because the copy-source is unmodified, but ideally they should be.

yuja (approved):

fowles (approved):

cli/tests/test_status_command.rs
yuja: I just meant the test is good so long as it contains at least one rename or copy entry, but the change looks also good to me.

BTW, it might be a bug (in gix or our backend) that `rename-target2` isn't detected as a copy.

martinvonz: I simplified the test a bit to only have one rename and one copy.

I agree that `rename-target2` seems like a bug somewhere. I haven't looked into where it is.

PR #4332, opened 2024-08-24
docs: add a roadmap

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -1,0 +1,91 @@
+# Roadmap
+
+This documents some of the goals we have. Many of them are quite independent.
+
+> **Note:** Most people contributing to Jujutsu do so in their spare time, which
+>  means that we cannot attach any target dates to any of the goals below.
+
+## Support for copies and renames
+
+We want to support copy tracing in a way that leaves it up to the commit backend
+to either record or detect copies. That should let us work with existing Git
+repos (Git does not record copies, it detects them on the fly) as well as with
+very large repos where detection would be too slow. See
+[design doc][copy-design-doc].
+
+## Forge integrations
+
+We would like to make it easier to work with various popular forges by providing
+something like `jj github submit`, `jj gitlab submit`, and `jj gerrit send`. For
+popular forges, we might include that support by default in the standard `jj`
+binary.
+
+## Submodule support
+
+Git submodules are used frequently enough in large Git repos that we will
+probably need to [support them][submodules]. There are still big open
+questions around UX.
+
+## Better Rust API for UIs
+
+UIs like [gg] currently have to duplicate quite a bit of logic from `jj-cli`. We
+need to make this code not specific to the CLI (e.g. return status objects
+instead of printing messages) and move it into `jj-lib`.
+
+## RPC API
+
+One problem with writing tools using the Rust API is that they will only work
+with the backends they were compiled with. For example, a regular [gg] build
+will not work on Google repos because it doesn't have the backends necessary to
+load them. We want to provide an RPC API for tools that want to work with an
+unknown build of `jj` by having the tool run something like `jj api` to give it
+an address to talk to.
+
+In addition to helping with the problem of unknown backends, having an RPC API
+should make it easier for tools like VS Code that are not written in Rust. The
+RPC API will probably be at a higher abstraction level than the Rust API.
+
+See [design doc][api-design-doc].
+
+## Open-source cloud-based repos (server and daemon process)
+
+Google has an internal Jujutsu server backed by a database. This server allows
+commits and repos (operation logs) to be stored in the cloud (i.e. the database).
+Working copies can still be stored locally.
+
+In order to reduce latency, there is a local daemon process that caches reads
+and writes. It also prefetches of objects it thinks the client might as for
+next. In also helps with write latency by optimistically answering write
+requests (it therefore needs to know the server's hashing scheme so it can
+return the right IDs).
+
+We (the project, not necessarily Google) want to provide a similar experience
+for all users. We would therefore like to create a similar server and daemon.
+The daemon might be the same process as for the RPC API mentioned above.
+
+## Virtual file system (VFS)
+
+For very large projects and/or large files, it can be expensive to update the
+working copy. We want to provide a VFS to help with that. Updating the working
+copy to another commit can then be done simply by telling the VFS to use the
+other commit as base, without needing to download any large files in the target
+commit until the user asks for them via the file system. A VFS can also make it
+cheap to snapshot the working copy by keeping track of all changes compared to
+the base commit.
+
+Having a VFS can also be very benefial for [`jj run`][jj-run], since we can then
+cheaply create temporary working copies for the commands to run in.
+
+## Better support for large files
+
+We have talked about somehow using content-defined chunking (CDC) to reduce
+storage and transfer costs for large files. Maybe we will store files in our
+future cloud-based server using the same model as [XetHub][xet-storage].
+
+
+[api-design-doc]: https://docs.google.com/document/d/1rOKvutee5TVYpFhh_UDNZDxfUKyrJ8rjCNpFaNHOHwU/edit?usp=sharing&resourcekey=0-922ApyoAjuXN_uTKqmCqjg
+[copy-design-doc]: design/copy-tracking.md
+[gg]: https://github.com/gulbanana/gg
+[jj-run]: https://github.com/martinvonz/jj/issues/1869
+[submodules]: https://github.com/martinvonz/jj/issues/494
+[xet-storage]: https://xethub.com/assets/docs/concepts/xet-storage
--- a/README.md
+++ b/README.md
@@ -11,11 +11,13 @@
 
 **[Homepage] &nbsp;&nbsp;&bull;&nbsp;&nbsp;**
 **[Installation] &nbsp;&nbsp;&bull;&nbsp;&nbsp;**
-**[Getting Started]**
+**[Getting Started] &nbsp;&nbsp;&bull;&nbsp;&nbsp;**
+**[Roadmap]**
 
 [Homepage]: https://martinvonz.github.io/jj
 [Installation]: https://martinvonz.github.io/jj/latest/install-and-setup
 [Getting Started]: https://martinvonz.github.io/jj/latest/tutorial
+[Roadmap]: https://martinvonz.github.io/jj/latest/roadmap
 
 </div>
 


docs/roadmap.md
```diff
@@ -0,0 +1,83 @@
...
+
+We want to support copy tracing in a way that leaves it up to the commit backend
+to either record or detect copies. That should let us work with existing Git
+repos (Git does not record copies, it detects them on the fly) as well as with
+very large repos where detection would be too slow. See
+[design doc][copy-design-doc].
+
+## Forge integrations
+
+We would like to make it easier to work with various popular forges by providing
```
rdamazio, line 18: Am I hearing "dynamically loadable extensions"? :) 'cause that sounds much better than trying to bake all this in.

PhilipMetzger: Please not, Rust the language has a bad stance on it and dealing with inconsistent `.so`'s on user machines distributed by distributors instead of us will make it a huge pain. See https://github.com/martinvonz/jj/issues/3575#issuecomment-2088812446 for a good explainer.

docs/roadmap.md
```diff
@@ -0,0 +1,83 @@
...
+should make it easier for tools like VS Code that are not written in Rust. The
+RPC API will probably be at a higher abstraction level than the Rust API.
+
+See [design doc][api-design-doc].
+
+## Open-source cloud-based repos (server and daemon process)
+
+Google has an internal Jujutsu server backed by a database. This server allows
+commits and repos (operation logs) to be stored in the cloud (i.e. the database).
+Working copies can still be stored locally.
```
PhilipMetzger, line 47: While this part implies _some_ kind of object storage independent from Git, it isn't explicitly written as that. Any reason why? As I think having a BLAKE3 derived storage will help the later goals in the Roadmap.

docs/roadmap.md
PhilipMetzger: I think we should manage that by making them Rust feature dependent, which works but makes code as brittle as C++/C #ifdefs if not managed correctly.

PhilipMetzger (approved):
This is a really nice Roadmap from all the discourse in the Discord from the last ~1.5 years.

docs/roadmap.md
emilazy: Rust can load dynamic code fine as long as you define a stable C ABI. You can do this by having a glue layer that matches up a rich Rust API on the host and a rich Rust API in the plugin through a C ABI that the plugin author never has to see. There are existing libraries to accomplish this, including loading plugins with WebAssembly which skips a lot of the OS‐level `dlopen` headache.

PhilipMetzger: I'm well aware that Rust can do that, but I still do not like it and Austin has put it nicely in the comment I linked to. It also sounds painful to maintain either way with or without the C ABI and its woes or WASM and its ecosystem.

thoughtpolice: "Dynamic extension loading for forges" is largely a red herring I'm afraid. I'll explain my reasoning briefly but, first off: no, there are no plans to do dynamic linking for anything, much less forge integrations, and I wouldn't read too much into this one particular bullet point on that note, or about this changing anytime soon. Furthermore, unless you specifically mean some other mechanism than the system dynamic linker, I'm going to assume that's what we're talking about.

In the past dynamic linking was useful for a variety of things. People always talk about stuff like "vendoring copies of zlib everywhere" or whatever and security patches, but originally it was actually just a massive time *and* space optimization. It turns out in the 1990s computers were slow and hard drives were slower; if you linked in a copy of a library into two programs, then executed them both, they were slower to start because they had more code pages that needed to be read from disk! Hard drive spindles are slow! Think about the boot process of a Unix machine; it's typically hundreds of short-lived program invocations. It was actually a massive wall clock time optimization to put everything into shared objects; code that would be read from disk 100 times now only needed to be loaded once. I remember reading some ancient hieroglyphics that once told a tale where boot time for a SunOS was slashed from minutes to seconds due to the introduction of shared object linking.

At the same time, in other places like Windows-land, it's also historically been a cultural consideration to allow dynamic loading of DLLs which are popular as a method of extending programs. Extensions are often useful in places because they allow things like:

- Decoupled development velocity: The main program may be aging but the extensions may thrive and add new functionality. The developers are not the same set of people, typically.
- Political/technical differences: some extensions add things that aren't seen as part of the project's focus or even go against the design criteria of the maintainers.
- Different approaches to a desired outcome. It's often useful to have some people explore new approaches.

Dynamic linking also requires:

- Careful technical design: you need a robust ABI/API design up front, because if you just spend time iterating it constantly and breaking plugins, what's the point? There is obviously some give and take here, but broadly speaking it requires a lot of investment in this ABI barrier to make it all work out.
- Cultural considerations: developers *and* extension authors have to actually respect existing ABIs. This is cultural! Not technical! Windows developers and Linux developers treat this completely differently in practice (more on this in a second)
- Particular distribution methods so that things do not explode fantastically because it turns out it's a *really* fiddly and user-unfriendly method of distributing things if it doesn't go just right.

And now I've got the bad news: honestly I feel none of these really apply to us or are so much work as to dwarf the actual tasks we want to accomplish.

- We don't want to decouple development! We're actively thriving. We want to *increase* the velocity of our own development if anything and keep things going. We actively want people to contribute upstream, and it's much easier to get things like deeper code review and shared tooling that way.
- On that note, we do a release every month so we're keen to get things into users hands quickly. So there isn't much overall velocity improvement that can come from having dynamic loading; we could just do releases more frequently too!
- There aren't political or technical difficulties -- everyone wants these features! We all think they're really important, even if we disagree about exact technical procedures to achieve them.
- While there is utility to having different approaches, we have strong opinions about topics like code review and patch submission, so I feel this is an opportunity for us to actually "double down" and really embrace Jujutsu's strengths. For example I don't really know how much different you can actually make a Gerrit integration, when you really get down to it.

And:

- Spending a lot of time designing a robust stable C-ified ABI for all our interfaces is a lot of work! It's tricky! There are platform specific things that suck! Dynamic linking was always a useful but massively complex feature in my experience working in the Haskell world, and it seems similarly "sticky" in Rust. Most importantly, we can spend this time actually improving the Rust codebase more extensively rather than spending time in the weeds.
- I'll be frank, Linux as a community, with a culture, just doesn't give a damn about ABI stability. Windows and even macOS in contrast deeply commit to at least years' worth of ABI/binary stability. In the Linux world, there's a massive amount of cultural inertia which boils down to "just recompile the thing" which completely eliminates the value in having a stable ABI, and isn't a good user experience. The distribution methods are mostly centralized too, so the decoupled development is often not really achieved because your distro has to do the updates for you anyway. We aren't packaged in distros or anything either, for the most part, so largely users just end up compiling things anyway. We shouldn't encourage this, IMO.
- And on the note of distribution methods: well, it's too much to write, but it requires everyone to do the right thing, anytime the wrong thing happens we'll get inscrutible bug reports, and due to the aforementioned bits about recompiling all the time, you often just nerf many benefits to dynamic loading.

I will also finally note two things to close this out:

- **We can still extend Jujutsu, today, without dynamic linking, and without upstream patches**! Just import the crates and add custom commands. Because of this many of the "experimentation" or "go your own way" points above are mitigated. No, it's not using the super cool dynamic linker, just boring Rust compilation. Yes, that has its own downsides. But it's there and largely fills most of the same practical roles. And: we want these features upstream, and at least in my case I'm already an upstream developer &mdash; so it makes sense for PRs like the Gerrit one to go here first. But I could have done all that out of tree, too.
- I like to think of this discussion as a good example of "top down" versus "bottom up" design dynamics, IMO. Dynamic linking is a very "bottom up" concern, it's a concern of building blocks, which is very easy for us to all latch on to (it's a difficult and technically interesting topic.) While "github integration for users" is a very "top-down" concern; it starts from the UX and moves from there. You can't design the UX without a list of features which is where we should start; beginning at dynamic linking doesn't tell us much because it has no relation to the features available! I think top-down design is very important for us as we seek to evolve because it will give us a better chance at a great UX that will make users happier, even if we don't always use some particular technical approach.

This is more than I expected to write, but that's to be expected from me, but I suspect this should probably address most of the talking points people have.

thoughtpolice: Oh, but I will say this: if we *do* want Jujutsu to be extendable in a more open-ended way, which I think is important &mdash; I think a much better "bang for buck" approach is to make the CLI scriptable instead of dynamic loading. I don't have much more to say on this topic for now, but it's something I want to explore sometime in the next year. So maybe keep your ears open (or bug me elsewhere on the topic)

emilazy: Sorry to make you write all of that; I just wanted to address the specific technical point :)

I think that there is value to people being able to add extensions without recompiling all of Jujutsu, in the long run, but I agree that loading code directly, whether via `dlopen` or WebAssembly, is probably not the best approach, and that just adding upstream support for a bunch of forges is the correct approach for most things.

docs/roadmap.md
```diff
@@ -0,0 +1,83 @@
+# Roadmap
```
thoughtpolice, line 1: This is just a general nit, but I wonder if there's value in just adding sections with our specific names/list of pet projects. Literally just a heading named `## Austin` or `## Yuya` for example. I think I have at least 2 things I'd like to put down in here for the forseeable future, but it's probably useful to also know they're my bailiwick so people can ask or contact me about it.

martinvonz: IMO: Get buy-in from a few others and add your ideas to a "Undecided" section instead.

Related: I meant this roadmap to have only items I felt like we have consensus on. If there are things here that seem controversial, we should remove them or move them to the "Undecided" section.

PhilipMetzger: I don't think adding the people responsible to it makes sense, at these are the common project goals. IMO, if you want a feature and responsibility for it, create an issue and assign to you. 

Because adding yourself to some long-term communicated goal can badly missfire.

docs/roadmap.md
```diff
@@ -0,0 +1,83 @@
...
+[design doc][copy-design-doc].
+
+## Forge integrations
+
+We would like to make it easier to work with various popular forges by providing
+something like `jj github submit`, `jj gitlab submit`, and `jj gerrit send`.
+
+## Better Rust API for UIs
+
+UIs like [gg] currently have to duplicate quite
```
sheremetyev, line 23: 👍 same story in VisualJJ - had to copy jj-cli and then modify

docs/roadmap.md
martinvonz: Do you think I should make any changes in this doc? It's already unspecified how we're going to implement it and I'm fine with that. Should I explicitly say that it's not yet decided? Should I say that we're likely going to add bundled support for a few popular forges (that's my preference)?

docs/roadmap.md
martinvonz: I'm not sure I understand your comment. I would prefer to not decide now what hash algorithm we're going to use, if that's what you're saying. Or are you saying that I should explicitly say that the server would use a hashing scheme that is *not* compatible with Git? It might actually be nice if we can have an option to be compatible with Git's hashing scheme so you can import a Git repo into our server.

emilazy: Maintaining compatibility with existing Git repositories would require allowing SHA‐1 into the system, which is going to be a can of worms. That’s not to say it might be the right trade‐off (especially since Git actually uses a hacked SHA‐1 that dances around the known easy collisions), but you’re sacrificing most of your potential guarantees right off the bat. (I do agree that committing to a hash function now makes no sense, but I don’t think that @PhilipMetzger meant that. (But I also agree with @PhilipMetzger that BLAKE3 is clearly the correct choice!))

docs/roadmap.md
```diff
@@ -0,0 +1,83 @@
...
+One problem with writing tools using the Rust API is that they will only work
+with the backends they were compiled with. For example, a regular [gg] build
+will not work on Google repos because it doesn't have the backends necessary to
+load them. We want to provide an RPC API for tools that want to work with an
+unknown build of `jj` by having the tool run something like `jj api` to give it
+an address to talk to.
+
+In addition to helping with the problem of unknown backends, having an RPC API
+should make it easier for tools like VS Code that are not written in Rust. The
+RPC API will probably be at a higher abstraction level than the Rust API.
```
sheremetyev, line 39: I wonder if this section addresses the same problem as previous one - providing good API for third-party tools (not necessarily UI), where definition of "good" API includes working well with alternative backends. This would probably imply RPC. Not sure there is a need for separate Rust API in addition to RPC.

docs/roadmap.md
PhilipMetzger: > I'm not sure I understand your comment.

Sorry for that.

> Or are you saying that I should explicitly say that the server would use a hashing scheme that is _not_ compatible with Git?

The intention was to get an answer if we could move on from Git as a commit storage model (it's our object store) and potentially build our own based on a BLAKE3 or even better hash in the future. This thought popped into my head while reading it, since Sapling is transitioning to BLAKE3 to unify the commit and CAS lookups, which probably is a major internal benefit to them. 

>  It might actually be nice if we can have an option to be compatible with Git's hashing scheme so you can import a Git repo into our server.

Ack, this makes definitely sense since a goal is to provide a seamless transition. But that shouldn't mean we'd need to support everything from Git.

docs/roadmap.md
PhilipMetzger: FYI, I have a nice summary here https://github.com/martinvonz/jj/issues/3034#issuecomment-2030300465.

docs/roadmap.md
PhilipMetzger: I think the latter is a good formulation, since I left a +1 w/o clearly saying which option it was for.

docs/roadmap.md
martinvonz: To summarize, the RPC API will probably be at a higher abstraction level, so you won't be able to create a new command that does a totally different thing, for example. Custom `jj` binaries (like the one at Google) will also need access to the low-level parts of the API.

sheremetyev: Thanks, that makes sense! Do you expect other use cases where that would be needed, beyond custom `jj` binary for custom backend at Google? E.g. do you foresee alternative backends implemented outside this codebase? (Just trying to understand motivation for Rust APIs correctly.)

Not sure if that's a good direction, but potentially `jj` CLI could use RPC API instead of linking directly and Google's internal version of CLI could use additional RPC API provided by Google backend. Are there performance concerns maybe that imply the need for direct linking?

docs/roadmap.md
rdamazio: I don't feel strongly about dynamic loading specifically, but I do think there'd be value to users if they can just `apt-get install jj; apt-get install some-jj-extension` (or equivalent) and then get updates for those without having to deal with cargo, keeping track of versions, etc. (e.g. I'd hate if a lot of users miss out on critical/security updates because they were forced to compile their own, or if they don't even try out jj because it's too much hassle to get it installed).
As others have pointed out there's many ways to achieve that though.

> Do you think I should make any changes in this doc? 

I'd be fine if no changes are made here.

docs/roadmap.md
martinvonz: > E.g. do you foresee alternative backends implemented outside this codebase?

I think it would make sense to add a backend for Meta's [Mononoke server[(https://github.com/facebook/sapling/tree/main/eden/mononoke) for example. Or maybe XetHub (now under Hugging Face) could implement a backend for their storage.

> Not sure if that's a good direction, but potentially `jj` CLI could use RPC API instead of linking directly and Google's internal version of CLI could use additional RPC API provided by Google backend. Are there performance concerns maybe that imply the need for direct linking?

Yes, that's a possibility.

martinvonz: FYI, @ylow. I have no idea if you'd have any interest in this but I'm happy to talk about it if you want. It seems like a good match to me.

docs/roadmap.md
martinvonz: Done.

docs/roadmap.md
martinvonz: > The intention was to get an answer if we could move on from Git as a commit storage model

Yes, that has always been a goal. The local backend uses a different hashing scheme, for example. So does the Google backend. I'll close this but let me know if you think I should change anything here.

martinvonz (commented):
I added a link to the roadmap from the top of the readme

docs/roadmap.md
```diff
@@ -0,0 +1,91 @@
+# Roadmap
+
+This is a list of goals we have. Many of them are quite independent.
```
ilyagr, line 3: (Optional:) Perhaps say

"This document lists some of the goals we have"

to make it clear this list is incomplete?

martinvonz: Makes sense. Done.

ilyagr:
Thank you, the roadmap looks nice to me!

This is very much non-blocking, but I would consider linking a Google Doc with the same roadmap, so that people can more easily start discussions about different parts of it. I was initially going to suggest only having a Google Doc, but on second thought, it's probably nice to occasionally to have more major updates this roadmap to go through a PR review and to have a commit description for them.

docs/roadmap.md
ilyagr: I was thinking also of how to communicate more of the sentiment from https://github.com/martinvonz/jj/pull/4332#discussion_r1730065786 about the intention of this roadmap, but this is the best I could come up so far.

Adding such a section could be good for a later update if you feel up to it.

martinvonz:
> it's probably nice to occasionally to have more major updates this roadmap to go through a PR review and to have a commit description for them.

Yes, I think I would prefer that.

I'll probably merge in about a day. It sounds like there are no major objections to what's in here, so that's good. We can add more goals later as we think of them or as we reach a consensus on them.

ilyagr (approved):

Issue #4122, opened 2024-07-19
FR: Rename `conflict()` revset function to `conflicts()`


cenviity:
**Is your feature request related to a problem? Please describe.**

The `conflict()` revset function may return multiple commits with conflicts, so the plural form `conflicts()` should be used to better indicate this. There is currently no `conflicts()` function, so an error occurs if the user tries to use it in a revset.

**Describe the solution you'd like**

- Rename `conflict()` to `conflicts()`.
- Perhaps soft-deprecate `conflict()` and keep it as an alias of `conflicts()` until some future version when it can be removed entirely, if deemed necessary. In the meantime a hint can be shown to the user suggesting use of `conflicts()` over `conflict()`.

**Describe alternatives you've considered**

We could remain with the status quo, but this seems inconsistent with the naming of other revset functions:

- Functions that may return multiple commits and are in plural form:
  - `parents(...)`
  - `children(...)`
  - `ancestors(...)`
  - `descendants(...)`
  - `*branches(...)`
  - `tags()`
  - `git_refs()`
  - `*heads(...)`
  - `roots(...)`
  - `merges()`
  - `working_copies()`
  - *`conflicts()` would belong in this group*
- Functions that return a single canonical commit (if it exists) and are in singular form:
  - `git_head()`
  - `root()`

yuja:
Our revset function naming is somewhat inconsistent, but I think `conflict()` belongs to the group that the function name represents a property of matching commits:

* is_/has_ `conflict()`
* is_ `empty()`
* matches_ `file()`
etc.

OTOH, `parents()`, `children()`, etc. are short for "parent" revisions, "child" revisions, ...

cenviity:
I see your reasoning, although since a commit may contain multiple conflicts I would read it as 'has_ `conflicts()`' rather. Or perhaps 'is_ `conflicted()`'?

yuja:
I'm not sure if adjective use of "conflict" is okay. I'm not against renaming to conflicted/conflicts/conflicting? if that sounds strictly better. We'll also need to update templater and some other internal APIs.

cenviity:
In fact, we could consider the functions with no parameters as taking an implicit repo and returning revsets that are almost like attributes of the repo. For example:

- `tags()` returns tagged commits of the repo.
- `merges()` returns merge commits of the repo.
- `root()` returns root commit of the repo.

In that view, `conflicts()` returns conflict commits of the repo.

[...]

PR #4334, opened 2024-08-24
Rename `conflict` revset to `conflicts`.

essiene:
See discussion thread in linked issue.

With this PR, all revset functions in [BUILTIN_FUNCTION_MAP](https://github.com/martinvonz/jj/blob/8d166c764251042287626ab37a1148386cef8371/lib/src/revset.rs#L570)
that return multiple values are either named in plural or the naming is hard to misunderstand (e.g. `reachable`)

Fixes: #4122
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   commit to commit. It now either follows the flags `--edit|--no-edit` or it
   gets the mode from `ui.movement.edit`.
 
+* `conflict` revset has been renamed to `conflicts`. With this, all revsets that
+  could potentially return multiple results are named in the plural.
+
 ### Deprecations
 
 * `jj untrack` has been renamed to `jj file untrack`.
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -789,7 +789,14 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
         function.expect_no_arguments()?;
         Ok(RevsetExpression::is_empty())
     });
-    map.insert("file", |diagnostics, function, context| {
+    map.insert("files", |diagnostics, function, context| {
+        // TODO: Remove in jj 0.28+
+        if function.name != "files" {
+            diagnostics.add_warning(RevsetParseError::expression(
+                "file() is deprecated; use files() instead",
+                function.name_span,
+            ));
+        }
         let ctx = context.workspace.as_ref().ok_or_else(|| {
             RevsetParseError::with_span(
                 RevsetParseErrorKind::FsPathWithoutWorkspace,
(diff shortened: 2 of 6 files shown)


martinvonz:
I'm for this change, so I'll let someone else review this PR. (Also, Essien and I work for the same company, so there's also that potential conflict of interest.)

martinvonz:
> I'm for this change, so I'll let someone else review this PR.

To clarify, I don't mean that only someone who's against this should review it :) I just mean that I've already said elsewhere that I'm for it, so I want to give others a chance to say what they think, in particular Yuya who seemed ambivalent.

thoughtpolice (approved):
I'm completely OK with this, though maybe an alias with a warning wouldn't hurt. (Someone else can hold this PR back if they think that's important enough.)

lib/src/revset.rs
```diff
@@ -773,7 +773,7 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
             RevsetFilterPredicate::DiffContains { text, files },
         ))
     });
-    map.insert("conflict", |function, _context| {
+    map.insert("conflicts", |function, _context| {
```
yuja, line 776: I think it's better to leave the original "conflict" function for a while (with TODO comment about deprecation warning.)

We'll probably need to rename `commit.conflict()` template function for consistency, and it's more likely to break user's stock templates.

I have no idea whether `--conflict`/`--conflicted` should be renamed to `--conflicts`, btw.

CHANGELOG.md
>  With this PR, all revset functions in [BUILTIN_FUNCTION_MAP](https://github.com/martinvonz/jj/blob/8d166c764251042287626ab37a114838>
> that return multiple values are either named in plural or named in a difficult-to-misused manner (e.g. `reachable`)

nit: this is not true. `file()` for example returns multiple revisions, but is not plural. I'm not saying it should be renamed to `files()`, but I mean `conflicts()` in this context can be read as "any revision has one or more conflicts."

lib/src/revset.rs
martinvonz: > I think it's better to leave the original "conflict" function for a while (with TODO comment about deprecation warning.)

Yes, good point (you and @thoughtpolice), especially since we know there are some tools that depend on the CLI (was it lazyjj? jj_tui?).

essiene: A good sould implemented diagnostics while I wasn't looking :-D
So I've gone ahead to add the deprecation warnings.

CHANGELOG.md
essiene: thanks for pointing that out.

I've actually gone ahead to also rename `file()` to `files()`, by the same logic, so it is consistent.

essiene: Thanks for checking and confirming. I'll go ahead to submit.

CHANGELOG.md
```diff
@@ -96,6 +96,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   commit to commit. It now either follows the flags `--edit|--no-edit` or it
   gets the mode from `ui.movement.edit`.
 
+* `conflict` revset has been renamed to `conflicts`. With this, all revsets that
+  could potentially return multiple results are named in the plural.
```
martinvonz, line 100: Should `file` have been mentioned here?

essiene: My bad. I forgot to come back to check this.

I've just sent #4529 to update the CHANGELOG.

CHANGELOG.md
yuja: Oh, I mean the rationale described in changelog and commit message doesn't make sense:
> With this, all revsets that could potentially return multiple results are named in the plural.

The change looks okay to me (though I don't think `files()` is better than `file()`.)

essiene: I completely misunderstood that. My bad.

PR #4335, opened 2024-08-24
Color signature names yellow within the log

qubitz:
Included the details in the commit message. There is probably a better way to do this, but I've been running with these changes in my local config.toml and seem to be working well with no other color conflicts. Suggestions welcome.

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/colors.toml
+++ b/cli/src/config/colors.toml
@@ -21,8 +21,8 @@
 "divergent prefix" = { fg = "red", underline = true }
 "hidden prefix" = "default"
 
-"email" = "yellow"
-"username" = "yellow"
+"author" = "yellow"
+"committer" = "yellow"
 "timestamp" = "cyan"
 "working_copies" = "green"
 "bookmark" = "magenta"
@@ -54,7 +54,8 @@
 "working_copy change_id" = "bright magenta"
 # We do not use bright yellow because of how it looks on xterm's default theme.
 # https://github.com/martinvonz/jj/issues/528
-"working_copy email" = "yellow"
+"working_copy author" = "yellow"
+"working_copy committer" = "yellow"
 "working_copy timestamp" = "bright cyan"
 "working_copy working_copies" = "bright green"
 "working_copy bookmark" = "bright magenta"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   to keeping all changes into the first commit while keeping the current commit
   description for the second commit (the newly created empty one).
 
+* Color author and committer names yellow
+
 ### Fixed bugs
 
 * Update working copy before reporting changes. This prevents errors during reporting
--- a/cli/tests/test_commit_template.rs
+++ b/cli/tests/test_commit_template.rs
@@ -346,8 +346,8 @@ fn test_log_builtin_templates_colored() {
     │
     ○  Commit ID: [38;5;4m230dd059e1b059aefc0da06a2e5a7dbf22362f22[39m
     │  Change ID: [38;5;5mqpvuntsmwlqtpsluzzsnyyzlmlwvmlnu[39m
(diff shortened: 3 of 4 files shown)


cli/src/config/colors.toml
```diff
@@ -23,6 +23,7 @@
 
 "email" = "yellow"
 "username" = "yellow"
+"log author name" = "yellow"
```
yuja, line 26: I'm not pretty sure, but maybe it's okay to colorize `author` and `committer`, and override the `timestamp` field?
```toml
"author" = "yellow"
"committer" = "yellow"
"timestamp" = "cyan"
```

This means author/committer names in `jj show` are colorized, but I think it's more consistent.

qubitz: Looks much better ❤️. I'll give that a shot.

qubitz:
Good call @yuja with the suggestion. Looks more consistent:

`jj show`
image

`jj log`
image

PR is ready for merging.

cli/src/config/colors.toml
```diff
@@ -23,6 +23,8 @@
 
 "email" = "yellow"
 "username" = "yellow"
+"author" = "yellow"
+"committer" = "yellow"
```
yuja, line 27: nit: perhaps, we can remove `"email" = ` and `"username" =` now.

And replace `"working_copy email" =` with `"working_copy author" =` and `"working_copy committer" =`? I have no idea why there's no `"working_copy username"`. That might be a bug.

yuja (approved):
Thanks.

You'll need to update test snapshots by `cargo insta test && cargo insta review`.
https://martinvonz.github.io/jj/latest/contributing/#summary

cli/src/config/colors.toml
qubitz: > nit: perhaps, we can remove `"email" = ` and `"username" =` now.

I'm not very experienced with the templating language. I'll take your word on it 👍. Doesn't seem make a difference in show or log.

> And replace `"working_copy email" =` with `"working_copy author" =` and `"working_copy committer" =`? I have no idea why there's no `"working_copy username"`. That might be a bug.

I'm more hesitate to make this change. If I'm understanding correctly you suggest:
```diff
-"working_copy email" = "yellow"
+"working_copy author" = "yellow"
+"working_copy committer" = "yellow"
+"working_copy username" = "yellow"
"working_copy timestamp" = "bright cyan"
```
Which is symmetric with the previous change which makes a kind of sense. However according to [this](https://martinvonz.github.io/jj/latest/templates/#signature-type), `name` , `email`, `username`, and `timestamp` are all related. Why is it then that you're suggesting that `name` and `email` be removed to but `username` be added? 

Thanks for taking the time to improve this PR 😊.

yuja: Oh, I meant there should be `"working_copy username"` but isn't. All `"<prefix> email"` and `"<prefix> username"` pairs can be replaced with `"<prefix> author"` and `"<prefix> committer"`.
```diff
@@ -21,8 +21,8 @@
 "divergent prefix" = { fg = "red", underline = true }
 "hidden prefix" = "default"
 
-"email" = "yellow"
-"username" = "yellow"
+"author" = "yellow"
+"committer" = "yellow"
 "timestamp" = "cyan"
 "working_copies" = "green"
 "bookmark" = "magenta"
@@ -54,7 +54,8 @@
 "working_copy change_id" = "bright magenta"
 # We do not use bright yellow because of how it looks on xterm's default theme.
 # https://github.com/martinvonz/jj/issues/528
-"working_copy email" = "yellow"
+"working_copy author" = "yellow"
+"working_copy committer" = "yellow"
 "working_copy timestamp" = "bright cyan"
 "working_copy working_copies" = "bright green"
 "working_copy bookmark" = "bright magenta"
```

qubitz: Ok I made the changes as you suggested. Please let me know if I misunderstood your intentions. 

I'm working to understand the failing tests and will rebase on top of main once I do.

PR #1997, opened 2023-08-07
EXPERIMENT: build jujutsu with buck2

thoughtpolice:
This is an experimental and fully un-tested branch, which attempts to add build rules for building `jj` with [Buck2](https://github.com/facebook/buck2), AKA `BUCK` files. The goal of such a technology would be to have a fully unified and integrated system for development of Jujutsu &mdash; no matter the build platform or client of the build system &mdash; even as the project expands and includes more and more features, more developers, and possibly languages and tools.

Please note that I don't think this is a branch that should in its current state be merged (or ever will? Assuming all goes well?) While I was partially motivated to unify things in `workspace.dependencies` recently because it would help here, that wasn't really the main goal. Part of it was to see if it could even remotely *work at all*; I am happily using buck2 for my own (private) Rust projects, so it was also an attempt to see if I could bring some of that magic over. I'm happy to say it *seems promising* but I've run into a (big) issue. In the meantime, I am publishing an interm report and PR on it.

If something like this is truly needed one day, I assume that's because Jujutsu will be wildly successful beyond our dreams, to the point that this sort of tech is a requirement, for reasons I'll go over. But maybe there's more to it than just that? This is above all an experiment, but I'll elaborate.

Because this kind of wide-reaching impact this kind of thing can have (on project direction, on developers, on support, et cetera) the remainder of this description is a very long novel about why you might be interested in this (or maybe why you would hate it.)

Note: This patch series shouldn't be taken as any kind of endorsement or commitment to its maintenance. If you're not interested, don't worry, it's not going anywhere. It may break, get rewritten, or otherwise behave incorrectly. Do not taunt Happy Fun Ball. 

> **NOTE**: Jujutsu has many Google employees who work with or on it, or are otherwise in touch with the development; they'll be familiar with much of what I'm about to discuss thanks to the use of Blaze/Bazel for their own needs, but for the purposes of pedagogy I'll assume the reader isn't familiar at all. Perhaps Bazel users will find something interesting or even learn something new, too. Bazel specifically will be discussed later.

## Prologue

In the beginning &mdash; literally, since the dawn of time &mdash; there was `make`. `make` is actually not one program but many, but all variants of `make` have a similar idea: to represent the compilation of software, or software artifacts, as a directed acyclic graph &mdash; where each node represents an "action" to perform that results in outputs, and where most actions (non `.PHONY` ones) can be cached in the filesystem.

[...]

PR #4336, opened 2024-08-25
cli, lib: inline grammar definitions into `.rs` files

thoughtpolice:
When building Jujutsu crates as a dependency in a Buck project, the usage of `#[grammar]` requires the `.pest` files in the code to be in the `$PWD` relative to where the Rust compiler is invoked on the crate; which isn't identical to how it is done by Cargo, causing a build failure ("file not found").

Rather than special case a Buck build path into the codebase for this (only used by downstream consumers in other projects), inlining the grammar definition into a new Rust file makes this go away.

# Checklist

If applicable:

### DIFF
--- a/cli/src/grammars/mod.rs
+++ b/cli/src/grammars/mod.rs
@@ -1,0 +1,17 @@
+// Copyright 2021-2024 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Grammars for parsing Jujutsu formats.
+
+pub mod template;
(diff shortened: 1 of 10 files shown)


thoughtpolice:
> When building Jujutsu crates as a dependency in a Buck project,

To be clear: I mean when I try to use `jj-cli` and `jj-lib` in a Buck project separate from this repository. This is *not* from the Buck branch in #1997 (though coincidentally it does help that branch too.)

thoughtpolice:
@yuja I know you work a lot on the parsers, more than anyone else, so I'm particularly keen on your feedback. I can maybe keep trying to find workarounds on my side so we don't need this patch, but if this doesn't make your life too much worse or you're ok with it, it would be a nice.

yuja:
For ease of maintenance, I prefer a separate `.pest` file. Can't we fix pest-derive somehow?

> When building Jujutsu crates as a dependency in a Buck project, the usage of
> `#[grammar]` requires the `.pest` files in the code to be in the `$PWD` relative
> to where the Rust compiler is invoked on the crate;

Apparently the search paths are:
1. `$CARGO_MANIFEST_DIR`
2. `$CARGO_MANIFEST_DIR/src`

https://github.com/pest-parser/pest/blob/master/generator/src/lib.rs#L68

I don't know buck or bazel, but maybe cargo environment variables could be faked?

thoughtpolice:
Aha, I was in fact holding it wrong and was able to diagnose it; there is a flag for setting `CARGO_MANIFEST_DIR` but I hadn't set it. Nevermind!

Issue #3525, opened 2024-04-17
FR: Support `.ignore`/`.jjignore`, a Git Independent way to ignore files. 


PhilipMetzger:
**Is your feature request related to a problem? Please describe.**

At some point we won't have Git as a base system anymore, which means we'll some way to migrate to a new `.ignore` file. It either could be a custom `.jjignore` or `.ignore` if it doesn't conflict with other tools you're using. This is to progress in 

**Describe the solution you'd like**
Add support for a `.jjignore`/`.ignore` file which can be adjacent to the `.gitignore` and a way to migrate the `.gitignore` to it (should be hopefully just a name change).  

**Describe alternatives you've considered**
Support `.gitignore` forever, which also could work, but I'd really like Git to be gone.

thoughtpolice:
> While adjacent, the `jj` one would have `.git` in it while the `git` one would have `.jj` in it.

Note that you don't need to add `.jj` to a `.gitignore` today because when jj uses a colocated repo, it automatically adds `.jj/*` to the [special `.git/info/exclude`](https://luisdalmolin.dev/blog/ignoring-files-in-git-without-gitignore/) file.

necauqua:
> A plain `.ignore` is not as future proof.

I kind of.. did not understand your message here. Why?

I've mentioned previously that creator of the [`ignore`](https://github.com/BurntSushi/ripgrep/tree/master/crates/ignore) crate (and ripgrep) is pushing for a unified .ignore (which is just a .gitignore that's not specific to git) file that's recognized by various tools and I like the idea a lot. So I'd be in favor of `.ignore` file compliant with that instead of making yet another `.somethingignore` file.

joyously:
> I've mentioned previously that creator of the [`ignore`](https://github.com/BurntSushi/ripgrep/tree/master/crates/ignore) crate (and ripgrep) is pushing for a unified .ignore

I haven't seen that remark. I suppose it doesn't matter that all the VCSs would be listed in one file. That's what would be intended anyway. I haven't thought about it as much as the crate author, but when I wrote that about future proofing, I was thinking in terms of a future maintainer that doesn't really know the old tool so wouldn't know that generic `.ignore` goes with it, and also mixing ignore files from VCS and some other kind of tool like backups or encryption or something else not written yet.

dbarnett:
I see two possible approaches here:
1. jj only respects .ignore in its own operations
[...]

PR #4338, opened 2024-08-25
cli: make paths to auto-track configurable, add `jj track`

martinvonz:
#323

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* The new config option `snapshot.auto-track` lets you automatically track only
+  the specified paths (all paths by default). Use the new `jj file track`
+  command to manually tracks path that were not automatically tracked. There is
+  no way to list untracked files yet. Use `git status` in a colocated workspace
+  as a workaround.
+  [#323](https://github.com/martinvonz/jj/issues/323)
+
 * `jj fix` now allows fixing unchanged files with the `--include-unchanged-files` flag. This
   can be used to more easily introduce automatic formatting changes in a new
   commit separate from other changes.
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -878,6 +878,18 @@ impl WorkspaceCommandHelper {
         Ok(FilesetExpression::union_all(expressions))
     }
 
+    pub fn auto_tracking_matcher(&self) -> Result<Box<dyn Matcher>, CommandError> {
+        let pattern = self.settings().config().get_string("snapshot.auto-track")?;
+        let expression = fileset::parse(
+            &pattern,
+            &RepoPathUiConverter::Fs {
+                cwd: "".into(),
+                base: "".into(),
+            },
+        )?;
+        Ok(expression.to_matcher())
+    }
+
     pub(crate) fn path_converter(&self) -> &RepoPathUiConverter {
         &self.path_converter
     }
@@ -1316,6 +1328,7 @@ impl WorkspaceCommandHelper {
             return Ok(());
         };
         let base_ignores = self.base_ignores()?;
+        let auto_tracking_matcher = self.auto_tracking_matcher()?;
 
         // Compare working-copy tree and operation with repo's, and reload as needed.
         let fsmonitor_settings = self.settings().fsmonitor_settings()?;
@@ -1371,6 +1384,7 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
             base_ignores,
             fsmonitor_settings,
             progress: progress.as_ref().map(|x| x as _),
+            start_tracking_matcher: &auto_tracking_matcher,
             max_new_file_size,
         })?;
         drop(progress);
--- a/cli/src/commands/file/track.rs
+++ b/cli/src/commands/file/track.rs
@@ -1,0 +1,68 @@
+// Copyright 2024 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::io::Write;
+
+use jj_lib::working_copy::SnapshotOptions;
+use tracing::instrument;
+
+use crate::cli_util::CommandHelper;
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Start tracking specified paths in the working copy
+///
+/// Without arguments, all paths that are not ignored will be tracked.
+///
+/// New files in the working copy can be automatically tracked.  
+/// You can configure which paths to automatically track by setting
+/// `snapshot.auto-track` (e.g. to `"none()"` or `"glob:**/*.rs"`). Files that
+/// don't match the pattern can be manually tracked using this command. The
+/// default pattern is `all()` and this command has no effect.
+#[derive(clap::Args, Clone, Debug)]
+pub(crate) struct FileTrackArgs {
+    /// Paths to track
+    #[arg(required = true, value_hint = clap::ValueHint::AnyPath)]
+    paths: Vec<String>,
+}
+
+#[instrument(skip_all)]
+pub(crate) fn cmd_file_track(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &FileTrackArgs,
+) -> Result<(), CommandError> {
+    let mut workspace_command = command.workspace_helper(ui)?;
+    let matcher = workspace_command
+        .parse_file_patterns(&args.paths)?
+        .to_matcher();
+
+    let mut tx = workspace_command.start_transaction().into_inner();
+    let base_ignores = workspace_command.base_ignores()?;
+    let (mut locked_ws, _wc_commit) = workspace_command.start_working_copy_mutation()?;
+    locked_ws.locked_wc().snapshot(&SnapshotOptions {
+        base_ignores,
+        fsmonitor_settings: command.settings().fsmonitor_settings()?,
+        progress: None,
+        start_tracking_matcher: &matcher,
+        max_new_file_size: command.settings().max_new_file_size()?,
+    })?;
+    let num_rebased = tx.repo_mut().rebase_descendants(command.settings())?;
+    if num_rebased > 0 {
+        writeln!(ui.status(), "Rebased {num_rebased} descendant commits")?;
+    }
+    let repo = tx.commit("track paths");
+    locked_ws.finish(repo.op_id().clone())?;
+    Ok(())
+}
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -40,6 +40,7 @@ This document contains the help content for the `jj` command-line program.
 * [`jj file chmod`↴](#jj-file-chmod)
 * [`jj file list`↴](#jj-file-list)
 * [`jj file show`↴](#jj-file-show)
+* [`jj file track`↴](#jj-file-track)
 * [`jj file untrack`↴](#jj-file-untrack)
 * [`jj fix`↴](#jj-fix)
 * [`jj git`↴](#jj-git)
@@ -739,6 +740,7 @@ File operations
 * `chmod` — Sets or removes the executable bit for paths in the repo
 * `list` — List files in a revision
 * `show` — Print contents of files in a revision
+* `track` — Start tracking specified paths in the working copy
 * `untrack` — Stop tracking specified paths in the working copy
 
 
@@ -806,6 +808,22 @@ If the given path is a directory, files in the directory will be visited recursi
 * `-r`, `--revision <REVISION>` — The revision to get the file contents from
 
   Default value: `@`
+
+
+
+## `jj file track`
+
+Start tracking specified paths in the working copy
+
+Without arguments, all paths that are not ignored will be tracked.
+
+New files in the working copy can be automatically tracked. You can configure which paths to automatically track by setting `snapshot.auto-track` (e.g. to `"none()"` or `"glob:**/*.rs"`). Files that don't match the pattern can be manually tracked using this command. The default pattern is `all()` and this command has no effect.
+
+**Usage:** `jj file track <PATHS>...`
+
+###### **Arguments:**
+
+* `<PATHS>` — Paths to track
 
 
 
--- a/cli/tests/test_file_track_untrack_commands.rs
+++ b/cli/tests/test_file_track_untrack_commands.rs
@@ -17,7 +17,7 @@ use std::path::PathBuf;
 use crate::common::TestEnvironment;
 
 #[test]
-fn test_untrack() {
+fn test_track_untrack() {
     let test_env = TestEnvironment::default();
     test_env.add_config(r#"ui.allow-init-native = true"#);
     test_env.jj_cmd_ok(test_env.env_root(), &["init", "repo"]);
@@ -103,7 +103,7 @@ fn test_untrack() {
 }
 
 #[test]
-fn test_untrack_sparse() {
+fn test_track_untrack_sparse() {
     let test_env = TestEnvironment::default();
     test_env.add_config(r#"ui.allow-init-native = true"#);
     test_env.jj_cmd_ok(test_env.env_root(), &["init", "repo"]);
@@ -124,6 +124,97 @@ fn test_untrack_sparse() {
     let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["file", "untrack", "file2"]);
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @"");
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "list"]);
+    insta::assert_snapshot!(stdout, @r###"
+    file1
+    "###);
+    // Trying to manually track a file that's not included in the sparse working has
+    // no effect. TODO: At least a warning would be useful
+    let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["file", "track", "file2"]);
+    insta::assert_snapshot!(stdout, @"");
+    insta::assert_snapshot!(stderr, @"");
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "list"]);
+    insta::assert_snapshot!(stdout, @r###"
+    file1
+    "###);
+}
+
+#[test]
+fn test_auto_track() {
+    let test_env = TestEnvironment::default();
+    test_env.add_config(r#"snapshot.auto-track = 'glob:*.rs'"#);
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    std::fs::write(repo_path.join("file1.rs"), "initial").unwrap();
+    std::fs::write(repo_path.join("file2.md"), "initial").unwrap();
+    std::fs::write(repo_path.join("file3.md"), "initial").unwrap();
+
+    // Only configured paths get auto-tracked
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "list"]);
+    insta::assert_snapshot!(stdout, @r###"
+    file1.rs
+    "###);
+
+    // Can manually track paths
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "track", "file3.md"]);
+    insta::assert_snapshot!(stdout, @"");
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "list"]);
+    insta::assert_snapshot!(stdout, @r###"
+    file1.rs
+    file3.md
+    "###);
+
+    // Can manually untrack paths
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "untrack", "file3.md"]);
+    insta::assert_snapshot!(stdout, @"");
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "list"]);
+    insta::assert_snapshot!(stdout, @r###"
+    file1.rs
+    "###);
+
+    // CWD-relative paths in `snapshot.auto-track` are evaluated from the repo root
+    let subdir = repo_path.join("sub");
+    std::fs::create_dir(&subdir).unwrap();
+    std::fs::write(subdir.join("file1.rs"), "initial").unwrap();
+    let stdout = test_env.jj_cmd_success(&subdir, &["file", "list"]);
+    insta::assert_snapshot!(stdout.replace('\\', "/"), @r###"
+    ../file1.rs
+    "###);
+
+    // But `jj file track` wants CWD-relative paths
+    let stdout = test_env.jj_cmd_success(&subdir, &["file", "track", "file1.rs"]);
+    insta::assert_snapshot!(stdout, @"");
+    let stdout = test_env.jj_cmd_success(&subdir, &["file", "list"]);
+    insta::assert_snapshot!(stdout.replace('\\', "/"), @r###"
+    ../file1.rs
+    file1.rs
+    "###);
+}
+
+#[test]
+fn test_track_ignored() {
+    let test_env = TestEnvironment::default();
+    test_env.add_config(r#"snapshot.auto-track = 'none()'"#);
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    std::fs::write(repo_path.join(".gitignore"), "*.bak\n").unwrap();
+    std::fs::write(repo_path.join("file1"), "initial").unwrap();
+    std::fs::write(repo_path.join("file1.bak"), "initial").unwrap();
+
+    // Track an unignored path
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "track", "file1"]);
+    insta::assert_snapshot!(stdout, @"");
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "list"]);
+    insta::assert_snapshot!(stdout, @r###"
+    file1
+    "###);
+    // Track an ignored path
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "track", "file1.bak"]);
+    insta::assert_snapshot!(stdout, @"");
+    // TODO: We should teach `jj file track` to track ignored paths (possibly
+    // requiring a flag)
     let stdout = test_env.jj_cmd_success(&repo_path, &["file", "list"]);
     insta::assert_snapshot!(stdout, @r###"
     file1
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -101,6 +101,10 @@ To squash or split commits, use `jj squash` and `jj split`.
 
 ### How can I keep my scratch files in the repository without committing them?
 
+You can set `snapshot.auto-track` to only start tracking new files matching the
+configured pattern (e.g. `"none()"`). Changes to already tracked files will
+still be snapshotted by every command.
+
 You can keep your notes and other scratch files in the repository, if you add
 a wildcard pattern to either the repo's `gitignore` or your global `gitignore`.
 Something like `*.scratch` or `*.scratchpad` should do, after that rename the
--- a/docs/working-copy.md
+++ b/docs/working-copy.md
@@ -12,12 +12,19 @@ working-copy contents when they have changed. Most `jj` commands you run will
 commit the working-copy changes if they have changed. The resulting revision
 will replace the previous working-copy revision.
 
-Also unlike most other VCSs, added files are implicitly tracked. That means that
-if you add a new file to the working copy, it will be automatically committed
-once you run e.g. `jj st`. Similarly, if you remove a file from the working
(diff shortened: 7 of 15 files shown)


cli/src/cli_util.rs
```diff
@@ -856,6 +857,18 @@ impl WorkspaceCommandHelper {
...
 
+    pub fn auto_tracking_matcher(&self) -> Result<Box<dyn Matcher>, CommandError> {
+        let pattern = self.settings.config().get_string("snapshot.auto-track")?;
+        // This special case is to support users who have disabled filesets
+        // TODO: once we delete allow-filesets, we can default to "all()" in misc.toml
+        // and drop this special case
+        if pattern.is_empty() {
+            return Ok(Box::new(EverythingMatcher));
+        }
+        let expression = self.parse_file_patterns(&[pattern])?;
```
yuja, line 868: nit: I think it's okay to disregard `allow-filesets` here. It's not path arguments, and `fix` patterns are parsed as filesets for example.

lib/src/local_working_copy.rs
```diff
@@ -1049,7 +1061,12 @@ impl TreeState {
                                 max_size: HumanByteSize(max_new_file_size),
                             });
                         }
-                        if let Some(new_file_state) = file_state(&metadata) {
+                        if maybe_current_file_state.is_none()
+                            && !start_tracking_matcher.matches(&path)
```
yuja, line 1065: maybe move this before file size check?

cli/src/cli_util.rs
martinvonz: This was specifically to support the default value from `misc.toml`. I've currently made that an empty string. If we make it `all()` instead, then we're going to break users who disable filesets. Setting it `.` doesn't work either, because it's relative to CWD by default, right?

yuja: `ui.allow-filesets` controls whether command arguments are parsed as filesets (mainly for backward compatibility.) I don't think config values would have to respect this flag.

And, we'll have to use low-level `fileset::parse()` function to not resolve the auto-track patterns relative to cwd.

docs/working-copy.md
```diff
@@ -12,12 +12,14 @@ working-copy contents when they have changed. Most `jj` commands you run will
...
-the working copy, first make sure it's [ignored](#ignored-files) and then run
-`jj untrack <path>`.
+Also unlike most other VCSs, added files are implicitly tracked by default. That
+means that if you add a new file to the working copy, it will be automatically
+committed once you run e.g. `jj st`. Similarly, if you remove a file from the
+working copy, it will implicitly be untracked. The `snapshot.auto-track` config
+option controls which paths get automatically tracked when they're added. See
+the [fileset documentation](filesets.md) for the syntax. To untrack a file while
+keeping it in the working copy, first make sure it's [ignored](#ignored-files)
+and then run `jj untrack <path>`.
```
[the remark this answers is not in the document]
joyously: > The content is not lost. 

So, both `track` and `untrack` do a snapshot?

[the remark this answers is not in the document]
martinvonz: Yes. The only thing that's lost is whether the file was tracked. The content is not lost.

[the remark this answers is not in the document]
marc-h38: > I think we should do away with the ambiguous states of tracked/ignored and untracked/not ignored.

Every version control system before `jj` had these states well defined. This does not mean they were a good idea but they were certainly not "ambiguous".

> The consequence would be that you would be reminded of the ignored files by every jj status command.

... or reminded of none. Either way this would get rid of the distinction between "temporary files with random names" versus "build directory with predictable name" which is probably the main use case discussed in #323. @joyously has been repeatedly trying to push back against this distinction in multiple  ways across multiple comments; this is just the latest attempt.
- https://github.com/martinvonz/jj/pull/4338#discussion_r1730438561 (previous comment)
- https://github.com/martinvonz/jj/pull/4338#discussion_r1730392319
- https://github.com/martinvonz/jj/pull/4338#discussion_r1730388156
- https://github.com/martinvonz/jj/issues/323#issuecomment-2308609293
- https://github.com/martinvonz/jj/issues/323#issuecomment-2218074254
- https://github.com/martinvonz/jj/issues/323#issuecomment-2308041154
- https://github.com/martinvonz/jj/issues/323#issuecomment-2308084601

CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* The new config option `snapshot.auto-track` lets you automatically track only
+  the specified paths (all paths by default). Use the new `jj track` command to
```
yuja, line 24: nit: `jj file track`

docs/working-copy.md
```diff
@@ -12,12 +12,16 @@ working-copy contents when they have changed. Most `jj` commands you run will
...
-the working copy, first make sure it's [ignored](#ignored-files) and then run
-`jj file untrack <path>`.
+Also unlike most other VCSs, added files are implicitly tracked by default. That
+means that if you add a new file to the working copy, it will be automatically
+committed once you run e.g. `jj st`. Similarly, if you remove a file from the
+working copy, it will implicitly be untracked. The `snapshot.auto-track` config
+option controls which paths get automatically tracked when they're added. See
+the [fileset documentation](filesets.md) for the syntax.
+
+You can use `jj untrack` to untrack a file while keeping it in the working copy.
```
yuja, line 22: nit: `jj file untrack`

docs/working-copy.md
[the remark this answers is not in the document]
emilazy: I don’t think it’s true that every historical version control system has had Git’s exact tracking/ignore system, but please be civil about the differences of opinion here rather than turning the discussion into a personal dispute…

cli/src/commands/file/track.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Start tracking specified paths in the working copy
+///
+/// Without arguments, all paths that are not ignored will be tracked.
+///
+/// By default, Jujutsu starts tracking all new files automatically. This
+/// command is not useful then. You can configure which paths to automatically
+/// track by setting e.g. `snapshot.auto-track = 'none()'`. You will then need
+/// to run this command to start tracking new files.
+#[derive(clap::Args, Clone, Debug)]
+pub(crate) struct TrackArgs {
```
yuja, line 33: nit: `FileTrackArgs`

cli/src/commands/file/track.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// Paths to track
+    #[arg(value_hint = clap::ValueHint::AnyPath)]
+    paths: Vec<String>,
```
yuja, line 36: nit: might be better to require at least one path. If the user set `auto-track = 'none()'`, they might have random build files in the working copy.

lib/src/local_working_copy.rs
```diff
@@ -1008,6 +1011,14 @@ impl TreeState {
                                 }
                             }
                         }
+                    } else if start_tracking_matcher.visit(&path).is_nothing()
+                        && current_tree.path_value(&path)?.is_absent()
```
yuja, line 1015: nit: maybe this can be `file_states.is_empty()`? It's a bit odd that we test `current_tree` here whereas scan `file_states` for git-ignored directory.

Maybe we can also combine these two to optimize directory walk.

lib/src/local_working_copy.rs
```diff
@@ -1034,15 +1046,22 @@ impl TreeState {
                         && git_ignore.matches(path.as_internal_file_string())
                     {
                         // If it wasn't already tracked and it matches
-                        // the ignored paths, then
-                        // ignore it.
+                        // the ignored paths, then ignore it.
                     } else {
+                        if maybe_current_file_state.is_none()
+                            && !start_tracking_matcher.matches(&path)
+                        {
```
yuja, line 1053: nit: I would move this one level up as `} else if ..`. This is quite similar to `git_ignore.matches()`.

yuja (approved):

docs/working-copy.md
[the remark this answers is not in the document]
marc-h38: Recommending at least 7 times "don't work like this" in more or less subtle and different ways may look very civil on the surface, and I did assume good faith most of the time: you can check the record. I'm afraid it's not possible to assume good faith anymore. 

I don't know whether `jj` can support the usage we described in #323 without affecting existing users (and I'm extremely grateful to @martinvonz for trying - on the weekend!) If that's not possible then fine: people like me will have to either change or not use `jj`. Disappointing but we'll all survive. My problem is: pretending to help while actually doing the opposite. This is time-consuming, not productive and not respectful.

Please propose a better and more civil way I should express this so I can learn.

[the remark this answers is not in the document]
joyously: Two conversations is not "repeatedly trying to push back". I ask questions to explore the consequences that might not have been considered. It really doesn't matter to me which way it goes, but this topic does point out a few ingrained workflows that might benefit from the slight changes that `jj` makes to how a VCS works.

lib/src/local_working_copy.rs
martinvonz: > nit: maybe this can be `file_states.is_empty()`? It's a bit odd that we test `current_tree` here whereas scan `file_states` for git-ignored directory.

Good idea. Done.

> Maybe we can also combine these two to optimize directory walk.

The gitignore matcher is a different kind of matcher, so I don't see how. Or do you just mean that we could skip asking the gitignore matcher if `file_states.is_empty()`? That makes sense.

docs/working-copy.md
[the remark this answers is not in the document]
marc-h38: > > > I think we should do away with the ambiguous states of tracked/ignored and untracked/not ignored.

> > At least outside jj, tracked + ignored is not so rare. For instance it's pretty common to ignore *.o and yet have a couple of binaries or other generated files tracked - because they were generated outside that project.

> Well, it could be that you keep ignores as is, but make the track and untrack command update the ignore file.

The `.gitignore` format is surprisingly complex: wildcards, negations, precedence rules... Should that complexity be duplicated in `jj`? It's not even just one file: you can have one `.gitignore` file in each subdirectory; which one should be updated? Also, what happens when the user modifies their HOME `~/.gitconfig` file? Or is this about some `jj`-specific ignore file, combined somehow with git's?

Also, when and where do you show untracked == ignored files and which ones? Always? Never? Depending on the `jj` command? Depending on some `jj cmd --ignored` option? A mix of the above? You've simplified the data model which always looks good on paper, but that increased the distance with real-world use cases from #323, which at best complexifies the user interface in order to bridge the gap between the two.


> this topic does point out a few ingrained workflows that might benefit from the slight changes that jj makes to how a VCS works.

Thanks: please be more direct and clearly explicit these "ingrained workflows" and usages that you think should simply not be supported at all rather than demonstrating how to avoid them in #323 and suggesting designs here that make them obviously harder or impossible to support. Then we can just agree to disagree on use cases and move on.

Brainstorming and prototyping is great and useful, but at some point a formally defined, high-level design that lists  what's possible versus not will be required; at the very least for documentation purposes if nothing else. Some back and forth is inevitable but right now I find it quite hard to discuss both implementation details and possibly shifting requirements at the same time.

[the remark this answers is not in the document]
marc-h38: > I don’t think it’s true that every historical version control system has had Git’s exact tracking/ignore system,

I don't know whether they were all similar (probably not) but for sure every system had to have some "unambiguous" definition of those states: defined at least by their respective implementation if not documented formally/properly.  I don't remember track/ignore definitions fundamentally different from the git ones and it looks like these stood the test of time. Recent tools like git have the huge benefit of hindsight (especially from BitKeeper but I digress).

lib/src/local_working_copy.rs
yuja: > The gitignore matcher is a different kind of matcher, so I don't see how. Or do you just mean that we could skip asking the gitignore matcher if `file_states.is_empty()`? That makes sense.

I mean these two `if`s can be combined.
```rust
                    if git_ignore.matches(&path.to_internal_dir_string())
                        || start_tracking_matcher.visit(&path).is_nothing()
                    {
                        // If the whole directory is ignored, visit only paths we're already
                        // tracking.
                        for (tracked_path, current_file_state) in file_states {
```

btw, in order to warn untracked paths, we'll need to visit subdirectory even if `file_states.is_empty()`. An empty untracked directory or directories containing only ignored files shouldn't be warned.

docs/working-copy.md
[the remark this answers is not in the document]
PhilipMetzger: > > I think we should do away with the ambiguous states of tracked/ignored and untracked/not ignored.
> 
> Interesting idea! That basically means dropping support for ignores and having just the auto-tracking config.
> > That basically means dropping support for ignores and having just the auto-tracking config.
> 
> Well, it could be that you keep ignores as is, but make the `track` and `untrack` command update the ignore file.

Thanks for bringing it up,  as I really like the idea, @joyously could you expand on it in #3525? While I have not deeply thought about the implications yet, I'd like to try such a approach for a native backend.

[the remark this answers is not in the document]
scott2000: If we want to show untracked files in `jj status`, then I feel like it would be better to have them be separate concepts, because ignored files shouldn't show up in `jj status`.

docs/FAQ.md
```diff
@@ -63,8 +63,9 @@ easy way to see the evolution of the commit's contents.
 
 ### Can I prevent Jujutsu from recording my unfinished work? I'm not ready to commit it.
 
-Jujutsu automatically records new files in the current working-copy commit and
-doesn't provide a way to prevent that.
+You can set `snapshot.auto-track` to only start tracking new files matching the
```
ilyagr, line 66: I think this passage might make more sense for the question about scratch files below.

For *this* question, you could change the paragraph to something like:

> Jujutsu automatically records changes to tracked files in the current working copy commit...

I think we could not mention "new" files at all.

Cc #4320, which might need to be adjusted once this PR goes in.

ilyagr (commented):
Nice! I only glanced at the docs, but this seems like a nice idea to try.

CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* The new config option `snapshot.auto-track` lets you automatically track only
+  the specified paths (all paths by default). Use the new `jj file track`
+  command to manually tracks path that were not automatically tracked. There is
+  no way to list untracked files yet.
```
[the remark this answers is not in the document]
ilyagr: (Idle thinking) I think we'd want an option to `jj file list` that debugs why files are tracked (it could now be the `auto-track` pattern, explicit `track` command; we could also distinguish .gitignore patterns that start with `!`). Perhaps that'd make sense to combine that with a command that lists untracked files.

cli/src/commands/file/track.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::cli_util::CommandHelper;
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Start tracking specified paths in the working copy
+///
+/// Without arguments, all paths that are not ignored will be tracked.
+///
+/// By default, Jujutsu starts tracking all new files automatically. This
+/// command is not useful then. You can configure which paths to automatically
```
ilyagr, line 29: Re "This command is not useful then", would

`jj file track ignored-file`

start tracking a file in `.gitignore`? I think it probably should, given that you can already end up tracking a `.gitignore`-d file if you edit `.gitignore`, but I'm not entirely sure.

docs/working-copy.md
```diff
@@ -12,12 +12,16 @@ working-copy contents when they have changed. Most `jj` commands you run will
...
 
-Also unlike most other VCSs, added files are implicitly tracked. That means that
-if you add a new file to the working copy, it will be automatically committed
-once you run e.g. `jj st`. Similarly, if you remove a file from the working
-copy, it will implicitly be untracked. To untrack a file while keeping it in
-the working copy, first make sure it's [ignored](#ignored-files) and then run
-`jj file untrack <path>`.
+Also unlike most other VCSs, added files are implicitly tracked by default. That
+means that if you add a new file to the working copy, it will be automatically
+committed once you run e.g. `jj st`. Similarly, if you remove a file from the
```
ilyagr, line 17: (Optional)

I think the "Similarly, if you remove a file..." sentence is a bit confusing. Perhaps we should delete it? I am tempted to say that the path is still tracked, but there is no file there. I think what happens when you delete a file depends also on whether the path is in `.gitignore`.

docs/working-copy.md
```diff
@@ -12,12 +12,16 @@ working-copy contents when they have changed. Most `jj` commands you run will
...
+working copy, it will implicitly be untracked. The `snapshot.auto-track` config
+option controls which paths get automatically tracked when they're added. See
```
ilyagr, line 19: (Optional) I would put a paragraph break before the sentence about `snapshot.auto-track`. If we'd like to connect it to the previous paragraph, perhaps we can split it: "The default can be changed with the `snapshot.auto-track` option. This option controls...`

I'm also wondering how to make this more correct in regards to ignore files. Perhaps just add a sentence: "Files with paths matching [ignore files] are never tracked automatically".

We could also add more details to the ignored files section. Something like:

> Ignored files are never tracked automatically (regardless of the value of `snapshot.auto-track`), but they can still end up being tracked for a few reasons:
>
> - if they were tracked in the parent commit
> - if they did not match the `.gitignore` previously, but the ignore file changed since
> - because of an explicit `jj file track` command
>
> You can untrack such files with the `jj file untrack` command.

(This might benefit from further editing)

cli/tests/test_file_track_untrack_commands.rs
```diff
@@ -128,4 +128,66 @@ fn test_untrack_sparse() {
...
+    insta::assert_snapshot!(stdout, @"");
+    insta::assert_snapshot!(stderr, @"");
+    let stdout = test_env.jj_cmd_success(&repo_path, &["file", "list"]);
+    insta::assert_snapshot!(stdout, @r###"
+    file1
+    "###);
+}
+
+#[test]
+fn test_auto_track() {
```
ilyagr, line 143: Whatever we decide the behavior should be, I think we should have a test for trying to `jj file track` an ignored file.

ilyagr:
I'll also Cc #4188 for the Windows error. In the case of this PR, the solution might be simpler, though.

docs/working-copy.md
```diff
@@ -12,12 +12,16 @@ working-copy contents when they have changed. Most `jj` commands you run will
...
+the [fileset documentation](filesets.md) for the syntax.
+
+You can use `jj file untrack` to untrack a file while keeping it in the working
+copy. However, first [ignore](#ignored-files) them or remove them from the
+`snapshot.auto-track` patterns; otherwise they will be immediately tracked again.
```
joyously, line 24: Shouldn't `track` and `untrack` bring up the ignore file or the config file in the editor, so it is more of an atomic action?

cli/tests/cli-reference@.md.snap
```diff
@@ -773,6 +775,24 @@ If the given path is a directory, files in the directory will be visited recursi
 
 
 
+## `jj file track`
+
+Start tracking specified paths in the working copy
+
+Without arguments, all paths that are not ignored will be tracked.
+
+By default, Jujutsu starts tracking all new files automatically. This command is not useful then. You can configure which paths to automatically track by setting e.g. `snapshot.auto-track = 'none()'`. You will then need to run this command to start tracking new files.
```
joyously, line 784: Does this say the same thing?
By default, Jujutsu tracks all files in the working copy automatically and this command does nothing (<--fix this?). If you manually configure the paths to automatically track by setting e.g. `snapshot.auto-track = 'none()'` then you will need to run this command to start tracking files.

cli/tests/test_file_track_untrack_commands.rs
joyously: How about tracking the ignore file? (and undoing it?)

ilyagr: > How about tracking the ignore file? (and undoing it?)

Are you asking about what happens when the `.gitignore` file itself is untracked?

I'm not sure whether we have tests about that, but that question is relevant independently of this PR. What Git does, what I believe to be the correct behavior, and what I *believe* jj does, is to respect `.gitignore` regardless of whether it is tracked or not. This allows me, for example, to have a `.vscode/.gitignore` file in my `jj` repo that contains `*`, thus preventing itself and the entire `.vscode` dir from being tracked. I learned this trick from `jj` itself, which creates a `.jj/.gitignore` file in co-located dirs for the same reason.

docs/working-copy.md
ilyagr: My current opinion is that something like this can probably be made to work and is a good suggestion in general, but there are enough devils sitting in the details that we might not want to implement that right now and right here.

So, I'd put this discussion outside the scope of this PR.

---------

That said, I thought a bit about either putting the list of ignored files in an exclude file or putting it into a config file, and here is a proposal for the latter (which seemed somewhat easier to me).

If we put the list of (un)tracked files in a TOML config file, the main questions in my mind are:
- Should `jj file (un)track` modify the repo-level config file or workspace-level, like https://github.com/martinvonz/jj/pull/3670? If it's the latter, we should implement it.
- Do we attempt to sync this config with Git's `.git/info/excludes` file? (Fortunately, I think we can always just say no to this)
- How do we deal with the possible situation if these configs are specified manually (as opposed to with `jj file track` which wouldn't do this) in both the repo-level and the user-level config files (and perhaps also in the command-line args or workspace-level). 
 
For the last question, I believe we currently have the more specific config fully override the less specific config. In particular, if both are TOML lists, they are *not* concatenated, the latter replaces the former. We could change that, but then it would be unclear how and whether to make it possible to remove an item from the list in the more specific config.

One trick that comes to mind is this. Have the config have the form:

```toml
[ignore_patterns]
global_list1=["pattern1", "pattern2", ]
global_list2=["pattern3"]
```

The names here are arbitrary. If that is the repo-level config, the workspace-level config could add a new list by adding `ignored_patterns.local_list1=["pattern5"]`. Or, it could adjust the top-level config by redefining `global_list1`.

docs/working-copy.md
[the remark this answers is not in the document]
ilyagr: > If we want to show untracked files in jj status, then I feel like it would be better to have them be separate concepts, because ignored files shouldn't show up in jj status.

I don't think we want that, usually, do we?

There are some exceptional case where we would want these to show up in `jj status`, such as files that weren't tracked because they are too large (see e.g. https://github.com/martinvonz/jj/issues/3912) or perhaps files that aren't tracked because `jj` detected some secrets that shouldn't be committed in them (once `jj` becomes all-too-smart). But in the normal case, if the user is aware these files aren't tracked, I don't think they should show up in `status`.

More generally, this discussion made me follow a few rabbit holes, see https://github.com/martinvonz/jj/pull/4338#discussion_r1741304289 for one of them.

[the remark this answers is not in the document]
scott2000: Personally, I think it would be better to encourage the user to make a explicit decision regarding untracked files (either to explicitly start tracking them or explicitly mark them as ignored), since that would prevent mistakes in either direction (accidentally committing files or forgetting to commit files). So based on that, I think it would be nice if untracked files showed up in `jj status` as a reminder to the user that they should make a decision.

I also think it would be simpler to make all untracked-but-not-explicitly-ignored files show up in `jj status` than to have special cases for large files or secrets (I could even imagine a future where there are `larger_than(1MiB)` and `contains_secret()` filesets, and these checks are implemented by excluding those filesets from `snapshot.auto-track`).

Another alternative would be to just give a message the first time an untracked file is snapshotted (even though it isn't added to the working copy commit), similar to how `jj git fetch` notifies about new untracked branches.

I could also see an argument that setting `snapshot.auto-track` is already making the final decision not to track any files that aren't included in it, but in that case I feel like the name would be misleading, and it should be called something like `snapshot.ignore` instead.

But I think having any config option at all is nice, and any UX improvements/bikeshedding could be done later.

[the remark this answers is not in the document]
ilyagr: > But I think having any config option at all is nice, and any UX improvements/bikeshedding could be done later.

I generally agree, I wouldn't block this PR on fully resolving these questions.

------------------

I think there are now at least two orthogonal features being discussed. One is to have a category of files that the user is reminded about in `jj status`. I do not very much like the name of "untracked but not ignored files" for this concept, but substantially having such a concept might make a lot of sense. I'd try to make it more like a tri-state: a path could be definitely tracked, definitely untracked, or pending a user decision. In my mind, this is closely related to the question of how to handle the `larger_than(1MiB)` files. 

The previous paragraph's feature would be a different feature than the feature of having a per-repo (or per-workspace) list of ignored files that is not shared with the other people working on the repo. I *thought* the latter is closer to what https://github.com/martinvonz/jj/issues/323 and this PR are about, though I might be wrong.

It's a confusing topic, since there are quite a few orthogonal features that can be thought of as related to ignored/untracked files.

> I could also see an argument that setting snapshot.auto-track is already making the final decision not to track any files that aren't included in it,

Yes, that's what I thought, though (again) I'm not entirely certain whether that is what Martin intended or what most people from #323 were thinking.

> but in that case I feel like the name would be misleading, and it should be called something like snapshot.ignore instead.

It didn't seem misleading to me, but that doesn't necessarily mean it isn't.

Either way, it is a good question of whether it's easier for the user to think in terms of ignore patterns or in terms of the opposite, patterns for files to track. In a vacuum, I'd go with the latter like Martin did, but perhaps people are so used to ignore patterns that the other way is better? I imagine people would get confused if we did something like https://github.com/martinvonz/jj/pull/4338#discussion_r1741304289 in terms of positive patterns rather than ignore patterns.

CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* The new config option `snapshot.auto-track` lets you automatically track only
```
ilyagr, line 23: I just had this possibly crazy idea: could we replace this option with a sparse pattern? Do we need separate concepts for sparse patterns and auto-tracked files?

`jj file (un)track` could modify sparse patterns for the working copy. We could rename "sparse patterns" if that's useful, much of the other discussion in this PR would still apply.

I had this thought after writing https://github.com/martinvonz/jj/pull/4338#discussion_r1741300502, this also follows some other assumptions I made about the goals of this that might be wrong.

ilyagr: The main issue that I can think of with spare files is that it's possible that the file X is in the repo and outside the sparse set, and there is a different X in the working copy. I'm not sure what is best to do in this case, but complaining loudly and treating this like a conflict of sorts seems like an option.

ilyagr: `jj file untrack` would remove the file from the repo and from the sparse pattern. There would also be a way (as there probably already exists) to change the sparse pattern without changing the repo, which would remove the file from the working copy.

One issue is whether we'd still try to warn people if they `jj file untrack` a file that's not `.gitignore`-d. This would be more difficult with the sparse files approach as opposed to the `snapshot.auto-track` approach.

ilyagr: Another issue here is whether sparse patterns can handle (currently, or can be sensibly changed to handle) gitignored-but-still-tracked files.

docs/working-copy.md
PhilipMetzger: > My current opinion is that something like this can probably be made to work and is a good suggestion in general, but there are enough devils sitting in the details that we might not want to implement that right now and right here.

I agree with this but think that we also should consider this as a potential solution for our own `.ignore` in #3525. Because it just eliminates one state.

docs/FAQ.md
martinvonz: I moved it to that section.

(But the current text here ("Jujutsu automatically records new files in the current working-copy commit and
doesn't provide a way to prevent that.") does sound very much related to this feature too, doesn't it? I'm not sure I understand the difference between the two entries.)

docs/working-copy.md
martinvonz: I don't find it confusing, and I don't think I changed that part anyway, so I'll leave it as is.

docs/working-copy.md
martinvonz: Makes sense. I dropped the middle bullet because it doesn't sound correct to me (they won't be tracked regardless of the value of `snapshot.auto-track`), and because it's already covered by "Ignored files" at the start of the paragraph.

docs/working-copy.md
martinvonz: Closing this

lib/src/local_working_copy.rs
martinvonz: > btw, in order to warn untracked paths, we'll need to visit subdirectory even if `file_states.is_empty()`. An empty untracked directory or directories containing only ignored files shouldn't be warned.

Good point. I've merged the conditions and updated the comment.

CHANGELOG.md
```diff
@@ -48,6 +48,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+  the specified paths (all paths by default). Use the new `jj file track`
+  command to manually tracks path that were not automatically tracked. There is
+  no way to list untracked files yet. Use `git status` in a colocated workspace
+  as a workaround.
+  [#323](https://github.com/martinvonz/jj/issues/323)
+
```
yuja, line 57: nit: move changelog entry to the unreleased section

martinvonz: Done

CHANGELOG.md
ilyagr: Making this function properly might be a bit trickier than I initially thought, but I think it might still be worth thinking about.

The analogue to `snapshot.auto-track = "none()"` would be to have the sparse patterns exclude any files that are not in the working copy commit. If this set was constant, that would be it; the "trickier than I initially thought" part is that currently the sparse set is the property of the working copy, while the set of files in the commit depends on the commit.

Even less conclusively, perhaps it's inevitable to have two separate configs for patters that are explicitly added to the sparse files, and patterns that are explicitly subtracted, but even then it might be useful to think in terms of these as methods to determine a single final set of files in the working copy that are tracked + snapshotted.

PR #4340, opened 2024-08-25
cli: move "untrack" to "file" subcommand

yuja:
# Checklist

If applicable:


PhilipMetzger (approved):

martinvonz:
An alternative would be to move `untrack` and `sparse` to some place for commands working with working copies. But `file untrack` also makes sense. Do we want to move `sparse` under `file` too?

martinvonz:
> Do we want to move `sparse` under `file` too?

But `sparse` *only* affects the working copy, so maybe it deserves to be separate in its top-level place?

PhilipMetzger:
Didn't we want to move `sparse` under `workspace`?

martinvonz:
> Didn't we want to move `sparse` under `workspace`?

Maybe. That's what I meant by "some place for commands working with working copies", while also leaving it open to be a separate namespace.

yuja:
If we move `sparse` under `workspace`, it might make sense to move `untrack` there (i.e. `jj workspace untrack`.) OTOH, `jj file list --untracked` might be the command to list untracked files, which isn't technically correct, but seems good for discoverability.

PhilipMetzger:
Yes, this makes sense if we view the _current working-copy_ addressed by its workspace as a subset of a `(default) workspace`  mutation.

martinvonz:
> OTOH, `jj file list --untracked` might be the command to list untracked files, which isn't technically correct, but seems good for discoverability.

I was thinking that listing them in `jj status` would be sufficient. Do you think we'll need a more specialized listing like `jj file list --untracked`?

martinvonz:
> If we move `sparse` under `workspace`, it might make sense to move `untrack` there (i.e. `jj workspace untrack`.)

Except that `jj workspace untrack` sounds like it somehow stops tracking the workspace (similar to `jj workspace forget`). So even though it makes sense in theory, I think it would be too confusing. Your proposed `jj file untrack` in this PR seems better to me.

yuja:
> I was thinking that listing them in `jj status` would be sufficient. Do you think we'll need a more specialized listing like `jj file list --untracked`?

No. It's just a possible reason to justify `jj file untrack`/`track`.

Oh, `jj status` could be considered an alias of `jj workspace status` though it's not just about the workspace. (To be clear, I'm not suggesting this change.)

> Except that `jj workspace untrack` sounds like it somehow stops tracking the workspace

Yes.

martinvonz (approved):

PR #4350, opened 2024-08-27
feat: Use JJ_LOG instead of RUST_LOG for tracing configuration

### LINKED ISSUE #4219: FR: Switch from RUST_LOG to JJ_LOG

I work on a lot of Rust projects, and I often set the RUST_LOG env var, either manually or automatically with direnv.

This means that I end up seeing `jj` logs , even if I absolutely don't care about them.

Rust itself switched from `RUST_LOG` over to `RUSTC_LOG` quite a while ago, with a similar motivation.

Since `jj` will be used in (Rust) development environments, it makes sense to :

* Stop respecting the `RUST_LOG` env var
* use `JJ_LOG` for logging configuration instead

Happy to open a PR for this if agreed upon.

theduke:
Prevents clashes when working on other Rust projects.

Closes #4219

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -189,11 +189,14 @@ pub struct TracingSubscription {
 }
 
 impl TracingSubscription {
+    const ENV_VAR_NAME: &str = "JJ_LOG";
+
     /// Initializes tracing with the default configuration. This should be
     /// called as early as possible.
     pub fn init() -> Self {
         let filter = tracing_subscriber::EnvFilter::builder()
             .with_default_directive(tracing::metadata::LevelFilter::ERROR.into())
+            .with_env_var(Self::ENV_VAR_NAME)
             .from_env_lossy();
         let (filter, reload_log_filter) = tracing_subscriber::reload::Layer::new(filter);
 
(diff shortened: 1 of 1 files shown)


theduke:
This should probably be documented somewhere, but not sure where the best place is.

martinvonz:
> This should probably be documented somewhere, but not sure where the best place is.

Perhaps docs/contributing.md is the best place we have right now

cli/src/cli_util.rs
```diff
@@ -189,7 +189,7 @@ pub struct TracingSubscription {
 }
 
 impl TracingSubscription {
-    const ENV_VAR_NAME: &str = "JJ_LOG";
+    const ENV_VAR_NAME: &'static str = "JJ_LOG";
```
martinvonz, line 192: please squash this into the parent commit

thoughtpolice:
I think this would be nice to get in before the 0.21 release. @theduke Can you squash these commits and re-push? If not I'm happy to do so (and retain your authorship!) and refile them myself, since we disallow the "Squash" button on GitHub.

samueltardieu:
Hi @theduke. Do you still plan to work on this?

theduke:
Sorry I dropped the ball here, will get it fixed today.

arxanas:
Superseded by https://github.com/martinvonz/jj/pull/5112.

PR #4355, opened 2024-08-28
Remove `ReadonlyRepo::repo_path`

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -21,7 +21,6 @@ use std::fmt::Debug;
 use std::fmt::Formatter;
 use std::fs;
 use std::path::Path;
-use std::path::PathBuf;
 use std::slice;
 use std::sync::Arc;
 
@@ -122,7 +121,6 @@ pub trait Repo {
 }
 
 pub struct ReadonlyRepo {
-    repo_path: PathBuf,
     store: Arc<Store>,
     op_store: Arc<dyn OpStore>,
     op_heads_store: Arc<dyn OpHeadsStore>,
@@ -138,10 +136,7 @@ pub struct ReadonlyRepo {
 
 impl Debug for ReadonlyRepo {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
-        f.debug_struct("Repo")
-            .field("repo_path", &self.repo_path)
-            .field("store", &self.store)
-            .finish()
+        f.debug_struct("Repo").field("store", &self.store).finish()
     }
 }
 
@@ -239,7 +234,6 @@ impl ReadonlyRepo {
             // be initialized properly.
             .map_err(|err| BackendInitError(err.into()))?;
         let repo = Arc::new(ReadonlyRepo {
-            repo_path,
             store,
             op_store,
             op_heads_store,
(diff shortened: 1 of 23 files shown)


martinvonz:
Looks like I messed up canonicalization. Will look into it later

martinvonz:
> Looks like I messed up canonicalization. Will look into it later

That should be done now. I added canonicalization in `TestBackend`.

lib/src/repo.rs
yuja, line 1: nit: typo in commit message: `Repoload::` -> `RepoLoader::`

lib/testutils/src/test_backend.rs
yuja, line 1: nit: maybe better to add inline comment? It's subtle problem and uneasy to figure out why `.canonicalize()` is needed here.

lib/tests/test_git.rs
```diff
@@ -159,14 +159,14 @@ fn test_import_refs() {
         view.get_remote_branch("main", "git"),
         &RemoteRef {
             target: RefTarget::normal(jj_id(&commit2)),
-            state: RemoteRefState::Tracking,
+            state: RemoteRefState::Tracking
```
yuja, line 162: nit: unrelated formatting change in `test_git.rs`.

lib/testutils/src/lib.rs
```diff
@@ -238,11 +245,13 @@ impl TestWorkspace {
             signer,
         )
         .unwrap();
+        let repo_path = workspace_root.join(".jj").join("repo");
```
yuja, line 248: nit: might be better to use `workspace.workspace_root()` or just forward to `workspace.repo_path()`. `Workspace::init*()` does path canonicalization.

lib/src/workspace.rs
```diff
@@ -150,24 +151,28 @@ fn init_working_copy(
...
     pub fn new(
         workspace_root: &Path,
+        repo_path: PathBuf,
         working_copy: Box<dyn WorkingCopy>,
         repo_loader: RepoLoader,
     ) -> Result<Workspace, PathError> {
         let workspace_root = workspace_root.canonicalize().context(workspace_root)?;
         Ok(Self::new_no_canonicalize(
             workspace_root,
+            repo_path,
```
yuja, line 161: Maybe it's safer to do `repo_path.canonicalize()` here? It was previously obtained from `repo.loader()`, which I think was canonicalized.

lib/tests/test_init.rs
```diff
@@ -42,7 +42,6 @@ fn test_init_local() {
         .backend_impl()
         .downcast_ref::<GitBackend>()
         .is_none());
-    assert_eq!(repo.repo_path(), &canonical.join(".jj").join("repo"));
```
yuja, line 45: nit: `workspace.repo_path()` could be tested instead

yuja (approved):

lib/src/repo.rs
martinvonz: Done

lib/testutils/src/test_backend.rs
martinvonz: Good idea. I added a comment on the map.

lib/src/workspace.rs
yuja: Oh, but `load()`-ed workspace wasn't? That's odd.

lib/tests/test_git.rs
martinvonz: Fixed. Not sure how that happened.

lib/testutils/src/lib.rs
martinvonz: Ah, thanks for catching that. Fixed

PR #4356, opened 2024-08-28
RFC: Formalize our commit style to mention that we don't use conventi…

PhilipMetzger:
…onal commits.

It's the simple _when in Rome, do as the Romans do_ rule.

cc @martinvonz 

# Checklist

If applicable:

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -36,9 +36,17 @@ commit and the new feature in a different commit. If the refactoring itself
 consists of many parts, try to separate out those into separate commits. You can
 use `jj split` to do it if you didn't realize ahead of time how it should be
 split up. Include tests and documentation in the same commit as the code they
-test and document. The commit message should describe the changes in the commit;
+test and document. 
+
+The commit message should describe the changes in the commit;
 the PR description can even be empty, but feel free to include a personal
-message.
+message. We write commit messages in an affected component style and don't use 
+[conventional commits](www.conventionalcommits.org/en/v1.0.0/). This means if 
+you modified a command in the CLI, use its name as the topic, e.g 
+`next/prev: <your-modification>` or `conflicts: <your-modification>`. We don't
+currently have a specific guidelines on what to write in the topic field, but 
+the reviewers will help you provide a topic if you have difficulties choosing 
+it.
 
 When you address comments on a PR, don't make the changes in a commit on top (as
 is typical on GitHub). Instead, please make the changes in the appropriate


martinvonz:
I'm personally fine with not including a topic and because "foo: " saves a few characters compared to "in foo, ". I use the topic style mostly out of habit. But I'm fine with requiring it if that's the consensus.

ilyagr:
I'm mildly opposed to this in its present form. It feels like formalization purely for the sake of formalization.

Such guidelines could be helpful if they actually decreased the cognitive load of people making commits. For example, I find topics helpful, but I have trouble figuring out exactly what should go in one. If we could have a good list of topics to solve this problem, that could be worth writing down (though I'm not sure this is easy to do). Even then, I'd say something like "follow this guideline if it helps, don't worry about it if it doesn't; we'll suggest something if needed"

As is, I think most people already have the instinct of "when in Rome...", so stating that explicitly only has a small benefit for the cost of increasing people's fear of making a mistake. We can always tell people to update their commit message if it becomes clearer with a topic.

ilyagr:
I don't mean to say that I'm opposed to something like conventional commits. We could switch to using them if it had a significant benefit, e.g. if we want to generate some parts of changelogs automatically. Though, I haven't seen it working perfectly, I don't know of a project that works this way that has an easy-to-follow changelog.

PhilipMetzger:
This PR was sparked by the conversation in https://github.com/martinvonz/jj/pull/4319#pullrequestreview-2255306636 and other PRs which I looked over during my time on the project. See the 3rd comment to a Discord link with some rough statistics for the current commit descriptions.

ilyagr:
Thanks for the link, it's quite helpful!

I'm not opposed to the goal of having a clearer explanation somewhere of how we write our commit messages, and your link mentions a few somewhat specific things we'd like to clarify. It's just tricky, and hopefully not super-urgent.

If it is urgent, by the way, I think this PR would be OK with the clause about "when in Rome..." removed. Or we could replace that sentence with: "We don't currently have specific guidelines on what to write in the topic field.", which puts less responsibility on the contributor.

fowles (approved):

PhilipMetzger:
> f it is urgent, by the way, I think this PR would be OK with the clause about "when in Rome..." removed. Or we could replace that sentence with: "We don't currently have specific guidelines on what to write in the topic field.", which puts less responsibility on the contributor.

I'm fine to reword it as that, it just was the first English idiom which fell into my mind for it.

docs/contributing.md
```diff
@@ -38,7 +38,13 @@ use `jj split` to do it if you didn't realize ahead of time how it should be
...
 test and document. The commit message should describe the changes in the commit;
 the PR description can even be empty, but feel free to include a personal
-message.
+message. We write commit messages in a affected component style and don't use 
+[conventional commits](www.conventionalcommits.org/en/v1.0.0/), this means if 
+you modified a command in the cli, use it's name as the topic, e.g 
+`next/prev: <your-modification>` or `conflicts: <your-modification>`. We don't
+currently have a specific guidelines on what to write in the topic field, but 
+the reviewers will help you provide a topic if you have difficulties choosing 
+it.
```
PhilipMetzger, line 47: @ilyagr PTAL

cenviity: nit:
a affected → an affected
, this means → . This means
cli → CLI
it's name → its name

Thanks for initiating this PR! I've always wondered about the 'convention' around here.

martinvonz: It seems like it would be useful if someone could state the benefit of having a convention. Maybe it's what you're kind of saying, @cenviity, that having written down that there's convention means that people don't have to wonder whether there is a convention. If that's the reason, I suppose an alternative is to explicitly state that you don't need to follow any convention (if that's the consensus).

I personally usually use a `<topic>: ` prefix. I think I do that because I'm used to it from the Git and Mercurial projects, but I'm not sure anyone has explained to me why I was supposed to do it 🤷

cenviity: Yes, mentioning this 'convention' in the contribution guidelines would stop people from having to wonder indeed – even if it's the fact that, ironically, a convention of whether to include a topic does _not_ exist. That's where the lowering of cognitive load comes in.

I don't feel too strongly one way or the other re: including a topic, though I can see one benefit of including a topic is that it helps with quickly glancing through the commit log after fetching repo changes.

ilyagr: Sorry it took me a while to get back to this.

Apart from cenviity 's helpful nits (https://github.com/martinvonz/jj/pull/4356#discussion_r1753108052) above, this text looks fine to me. Thank you for the edits! I'll approve this based on the fact that I think it describes our status-quo well.

> having written down that there's convention means that people don't have to wonder whether there is a convention. If that's the reason, I suppose an alternative is to explicitly state that you don't need to follow any convention (if that's the consensus).

This is a good reason to write down something like this.

> I personally usually use a <topic>: prefix. I think I do that because I'm used to it from the Git and Mercurial projects, but I'm not sure anyone has explained to me why I was supposed to do it 🤷

I don't necessarily have an answer, but I generally like the topic prefix. It allows me to quickly look at the list of recent `jj` commits and judge what kind of things have changed and (for example) how likely I am to see conflicts if I rebase.

ilyagr (approved):
I am not sure there is a consensus, but this LGTM.

docs/contributing.md
martinvonz: Good point about using it for scanning the changelog. I have sometimes mixed topics in a PR (like https://github.com/martinvonz/jj/pull/4300/commits). Maybe I should try to more consistently use the same topic in a PR/series. I'll try to remember to do that next time.

PhilipMetzger: I also think that stating that we have a convention is beneficial for contributors, instead of stating that it is a free for all. And the `topic:` convention will scale nicely to further projects like the Web UI for `jj web`, native server and the different CR providers.

docs/contributing.md
```diff
@@ -38,7 +38,13 @@ use `jj split` to do it if you didn't realize ahead of time how it should be
 split up. Include tests and documentation in the same commit as the code they
 test and document. The commit message should describe the changes in the commit;
 the PR description can even be empty, but feel free to include a personal
-message.
+message. We write commit messages in an affected component style and don't use
```
martinvonz, line 41: This paragraph is getting long. Maybe break before "The commit message should ..."?

PhilipMetzger: added a break.

PR #4362, opened 2024-08-29
workspace: turn WorkspaceLoader into a trait

torquestomp:
Similar to other changes, this small API change allows extensions to take over repo and working copy loading entirely when the factory types cannot be obtained through the local filesystem.

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -113,11 +113,14 @@ use jj_lib::working_copy::SnapshotOptions;
 use jj_lib::working_copy::WorkingCopy;
 use jj_lib::working_copy::WorkingCopyFactory;
 use jj_lib::workspace::default_working_copy_factories;
+use jj_lib::workspace::get_working_copy_factory;
+use jj_lib::workspace::DefaultWorkspaceLoaderFactory;
 use jj_lib::workspace::LockedWorkspace;
 use jj_lib::workspace::WorkingCopyFactories;
 use jj_lib::workspace::Workspace;
 use jj_lib::workspace::WorkspaceLoadError;
 use jj_lib::workspace::WorkspaceLoader;
+use jj_lib::workspace::WorkspaceLoaderFactory;
 use once_cell::unsync::OnceCell;
 use tracing::instrument;
 use tracing_chrome::ChromeLayerBuilder;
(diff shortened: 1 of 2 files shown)


cli/src/cli_util.rs
martinvonz, line 1: > Similar to other changes, this small API change allows extensions to take over repo and working copy loading entirely when the factory types cannot be obtained through the local filesystem.

nit: please move this into the commit description

Also, can you elaborate a bit? For a long time I thought this was related to the server use case and I couldn't figure out why the server wouldn't just instantiate a `Workspace` manually. Then I remembered that you're working on testing now, so I guess this has something to do with that? Could you explain a bit more in the commit description what problem this solves?

cli/src/cli_util.rs
```diff
@@ -341,8 +344,11 @@ impl CommandHelper {
         &self.operation_template_extensions
     }
 
-    pub fn workspace_loader(&self) -> Result<&WorkspaceLoader, CommandError> {
-        self.maybe_workspace_loader.as_ref().map_err(Clone::clone)
+    pub fn workspace_loader(&self) -> Result<&dyn WorkspaceLoader, CommandError> {
+        self.maybe_workspace_loader
+            .as_ref()
+            .map(|l| l.as_ref())
```
martinvonz, line 350: does `.as_deref()` work?

martinvonz (commented):
I'll let a non-googler approve as usual

cli/src/cli_util.rs
torquestomp: Added description to the commit and elaborated a bit there.  Short answer, yes this has to do with testing, because tests do a ton of sub-processing to jj as a binary (as opposed the server where it's invoked as a library).

lib/src/workspace.rs
```diff
@@ -460,17 +460,70 @@ impl<'a> LockedWorkspace<'a> {
...
+        Ok(factory.as_ref())
+    } else {
+        Err(StoreLoadError::UnsupportedType {
+            store: "working copy",
+            store_type: working_copy_type.to_string(),
+        })
+    }
+}
+
+pub trait WorkspaceLoader {
```
PhilipMetzger, line 484: Nit: Missing documentation for the whole trait.

cli/src/cli_util.rs
PhilipMetzger: > Short answer, yes this has to do with testing, because tests do a ton of sub-processing to jj as a binary (as opposed the server where it's invoked as a library).

Could also add a version of this to the commit message as it makes it complete.

lib/src/workspace.rs
```diff
@@ -460,17 +460,70 @@ impl<'a> LockedWorkspace<'a> {
...
 
+pub trait WorkspaceLoaderFactory {
+    fn create(&self, workspace_root: &Path)
+        -> Result<Box<dyn WorkspaceLoader>, WorkspaceLoadError>;
+}
+
+pub fn get_working_copy_factory<'a>(
+    workspace_loader: &dyn WorkspaceLoader,
+    working_copy_factories: &'a WorkingCopyFactories,
+) -> Result<&'a dyn WorkingCopyFactory, StoreLoadError> {
```
PhilipMetzger, line 471: It's not possible to elide this lifetime?

PhilipMetzger (approved):
LGTM

lib/src/workspace.rs
torquestomp: I don't think so, because there are two input lifetimes.

PR #4366, opened 2024-08-30
cli: fix: add --include-unchanged-files flag to allow fixing as yet unchanged files

hooper:
This enables workflows like "insert a commit that reformats the code in one of my project directories".

`jj fix --include-unchanged-files` is an easy way to fix everything in the repo.

`jj fix --include-unchanged-files <file...>` fixes all of the `<files>` even if they are unchanged.

This is mostly orthogonal to other features, so not many tests are added.

This is a significant and simple enough improvement that I think it's appropriate to make it here instead of waiting for a `jj run`-based solution.

# Checklist

If applicable:

`jj diff -r tumtymnsotvq` shows the change a bit more clearly than Github.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj fix` now allows fixing unchanged files with the `--include-unchanged-files` flag. This
+  can be used to more easily introduce automatic formatting changes in a new
+  commit separate from other changes.
+
 ### Fixed bugs
 
 ## [0.21.0] - 2024-09-04
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -28,6 +28,7 @@ use jj_lib::fileset;
 use jj_lib::fileset::FilesetExpression;
 use jj_lib::matchers::EverythingMatcher;
 use jj_lib::matchers::Matcher;
+use jj_lib::merged_tree::MergedTree;
 use jj_lib::merged_tree::MergedTreeBuilder;
 use jj_lib::merged_tree::TreeDiffEntry;
 use jj_lib::repo::Repo;
@@ -36,6 +37,7 @@ use jj_lib::repo_path::RepoPathUiConverter;
 use jj_lib::revset::RevsetExpression;
 use jj_lib::revset::RevsetIteratorExt;
 use jj_lib::store::Store;
+use jj_lib::tree::Tree;
 use pollster::FutureExt;
 use rayon::iter::IntoParallelIterator;
 use rayon::prelude::ParallelIterator;
@@ -125,6 +127,10 @@ pub(crate) struct FixArgs {
     /// Fix only these paths
     #[arg(value_hint = clap::ValueHint::AnyPath)]
     paths: Vec<String>,
+    /// Fix unchanged files in addition to changed ones. If no paths are
+    /// specified, all files in the repo will be fixed.
+    #[arg(long)]
+    include_unchanged_files: bool,
 }
 
 #[instrument(skip_all)]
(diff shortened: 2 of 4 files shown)


cli/src/commands/fix.rs
```diff
@@ -125,6 +126,11 @@ pub(crate) struct FixArgs {
     /// Fix only these paths
     #[arg(value_hint = clap::ValueHint::AnyPath)]
     paths: Vec<String>,
+    /// Fix unchanged files in addition to changed ones. If no files are
+    /// specified, all files in the repo will be fixed.
+    // TODO: use -L for --all-lines, and avoid -A
+    #[arg(long, short = 'F')]
+    all_files: bool,
```
yuja, line 133: nit: maybe `--include-unchanged-files` or something is better?

`fix --all-files PATH` looks a bit odd.

cli/src/commands/fix.rs
```diff
@@ -185,39 +191,49 @@ pub(crate) fn cmd_fix(
...
+            async {
+                parent_tree
+                    .diff_stream(&tree, &matcher)
+                    .map(|entry| entry.values.map(|(_before, after)| (entry.path, after)))
+                    .try_collect()
+                    .await
+            }
+            .block_on()?
+        };
+        tree_values.into_iter().try_for_each(|(path, terms)| {
```
yuja, line 220: nit: simply use `for (path, terms) in tree_values`?

cli/src/commands/fix.rs
```diff
@@ -185,39 +191,49 @@ pub(crate) fn cmd_fix(
             }
         }
 
-        // Also fix any new paths that were changed in this commit.
+        // There are four behaviors for the combinations of specifying or omitting the
+        // --all-files flag or file arguments. The flag determines whether we diff the
+        // revision against its parents or just fix the entire tree. In either case, the
+        // file arguments filter that set of files.
```
martinvonz, line 197: It could be implemented as diffing from the empty tree. I expect the performance difference not to be noticeable.

hooper: That is more like what `hg fix` does, but I would rather not rework this right now.  The "baserevs" logic for doing the diff in the `include_unchanged_files == false` case is not very complicated, but it's more complicated than what we have here right now.

cli/src/commands/fix.rs
hooper: Thanks, that was probably left over from it being chained off the previous expression at some point.

cli/src/commands/fix.rs
martinvonz: > I expect the performance difference not to be noticeable.

Actually, I take that back. Using `entries_matching` can be a lot slower at Google because it doesn't do the parallel tree walk that `diff_stream()` does. We should fix that, but that shouldn't block this PR.

This is what I had in mind:

```rust
let base_tree = if args.include_unchanged_files {
  MergedTree::resolved(Tree::empty(tx.repo().store().clone(), RepoPathBuf::root()))
} else {
  commit.parent_tree(tx.repo())?
};
let diff_stream = base_tree.diff_stream(&tree, &matcher);
/// as before
```

I don't think it's more complicated than what we have right now, but I'm fine with either form. I think using `entries_matching()` is conceptually simpler because it's more direct. Again, up to you.

hooper: Diffing against an empty tree seems to use fewer Rust concepts, so I'll go with that.

CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   
 * When reconfiguring the author, warn that the working copy won't be updated
 
+* `jj fix` now allows fixing unchanged files with the `--all-files` flag. This
+  can be used to more easily introduce automatic formatting changes in a new
+  commit separate from other changes.
+
```
martinvonz, line 43: Don't forget to move this to the new "Unreleased" section after syncing this commit with main

cli/src/commands/fix.rs
```diff
@@ -185,39 +191,47 @@ pub(crate) fn cmd_fix(
...
-                            file_id: id.clone(),
-                            repo_path: repo_path.clone(),
-                        };
-                        unique_tool_inputs.insert(tool_input.clone());
-                        paths.insert(repo_path.clone());
-                    }
+                entry.values.map(|(_before, after)| (entry.path, after))
+            })
+            .try_collect()
+            .await
```
martinvonz, line 216: Looks like this async block can be removed since you're just awaiting a single value and immediately blocking on it.

cli/src/commands/fix.rs
```diff
@@ -185,39 +191,47 @@ pub(crate) fn cmd_fix(
...
+                commit.parent_tree(tx.repo())?
+            }
+            .diff_stream(&tree, &matcher)
+            .map(|entry| {
+                // TODO: handle copy tracking by fixing all destination files, and only the
+                // source files that were otherwise modified. This will also need to extend to
+                // changed line computation once we support that.
                 // Deleted files have no file content to fix, and they have no terms in `after`,
                 // so we don't add any tool inputs for them. Conflicted files produce one tool
                 // input for each side of the conflict.
```
martinvonz, line 212: I think this comment belongs just before the `.flatten()` call below

martinvonz (commented):
LGTM, but I'll let a non-Googler give the approval as usual

cli/src/commands/fix.rs
```diff
@@ -185,39 +191,45 @@ pub(crate) fn cmd_fix(
...
-                        unique_tool_inputs.insert(tool_input.clone());
-                        paths.insert(repo_path.clone());
-                    }
+        let tree_values: Vec<(RepoPathBuf, Merge<Option<TreeValue>>)> =
+            if args.include_unchanged_files {
+                MergedTree::resolved(Tree::empty(tx.repo().store().clone(), RepoPathBuf::root()))
+            } else {
+                commit.parent_tree(tx.repo())?
+            }
+            .diff_stream(&tree, &matcher)
```
yuja, line 205: nit: I don't have preference about async use, but I think the actual diff here is just:
```rust
let parent_tree = if include_unchanged_files {
    MergedTree::resolved(Tree::empty(..))
} else {
    commit.parent_tree(tx.repo())?
};
```

yuja (approved):

cli/src/commands/fix.rs
martinvonz: Yes, I was also wondering about the other changes. It seems they should at least go in a separate commit with some justification. Maybe ideally a separate PR so it doesn't have to block the feature change.

martinvonz (commented):
LGTM

PR #4367, opened 2024-08-31
rewrite: do not resolve transitive parents repeatedly when updating refs

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1032,6 +1032,40 @@ impl MutableRepo {
         new_ids
     }
 
+    /// Fully resolves transitive replacements in `parent_mapping`.
+    ///
+    /// If `parent_mapping` contains cycles, this function will panic.
+    fn resolve_rewrite_mapping_with(
+        &self,
+        mut predicate: impl FnMut(&Rewrite) -> bool,
+    ) -> HashMap<CommitId, Vec<CommitId>> {
+        let sorted_ids = dag_walk::topo_order_forward(
+            self.parent_mapping.keys(),
+            |&id| id,
+            |&id| match self.parent_mapping.get(id).filter(|&v| predicate(v)) {
+                None => &[],
+                Some(rewrite) => rewrite.new_parent_ids(),
+            },
+        );
+        let mut new_mapping: HashMap<CommitId, Vec<CommitId>> = HashMap::new();
+        for old_id in sorted_ids {
+            let Some(rewrite) = self.parent_mapping.get(old_id).filter(|&v| predicate(v)) else {
+                continue;
+            };
+            let lookup = |id| new_mapping.get(id).map_or(slice::from_ref(id), |ids| ids);
+            let new_ids = match rewrite.new_parent_ids() {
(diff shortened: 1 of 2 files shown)


lib/src/repo.rs
```diff
@@ -1041,12 +1041,12 @@ impl MutableRepo {
...
-        for (old_parent_id, rewrite) in self.parent_mapping.clone() {
-            // Call `new_parents()` here since `parent_mapping` only contains direct
+        for (old_parent_id, _rewrite) in self.parent_mapping.clone() {
+            // Call `rewritten_ids_with()` here since `parent_mapping` only contains direct
             // mappings, not transitive ones.
             // TODO: keep parent_mapping updated with transitive mappings so we don't need
-            // to call `new_parents()` here.
-            let new_parent_ids = self.new_parents(rewrite.new_parent_ids());
+            // to call `rewritten_ids_with()` here.
+            let new_parent_ids = self.rewritten_ids_with(slice::from_ref(&old_parent_id), |_| true);
```
martinvonz: Seems useful to have a test for this, especially since I've apparently broken it once already. IIUC, `test_rebase_descendants_update_branches_after_divergent_rewrite` is close but in order to trigger this bug, we would need to add further rewrites (not necessarily divergent) after the first divergent rewrite.

lib/src/repo.rs
```diff
@@ -1032,6 +1032,40 @@ impl MutableRepo {
         new_ids
     }
 
+    /// Fully resolves transitive replacements in `parent_mapping`.
+    ///
+    /// If `parent_mapping` contains cycles, this function will panic.
+    fn resolve_rewrite_mapping_with(
+        &self,
+        mut predicate: impl FnMut(&Rewrite) -> bool,
```
martinvonz, line 1040: Does this function take a predicate because you expect to add another caller that will use a different predicate?

lib/src/repo.rs
```diff
@@ -1032,6 +1032,40 @@ impl MutableRepo {
...
+                None => &[],
+                Some(rewrite) => rewrite.new_parent_ids(),
+            },
+        );
+        let mut new_mapping: HashMap<CommitId, Vec<CommitId>> = HashMap::new();
+        for old_id in sorted_ids {
+            let Some(rewrite) = self.parent_mapping.get(old_id).filter(|&v| predicate(v)) else {
+                continue;
+            };
+            let lookup = |id| new_mapping.get(id).map_or(slice::from_ref(id), |ids| ids);
```
martinvonz, line 1055: Does `unwrap_or()` work here?

yuja: No. We need to coerce `&Vec<_>` to `&[_]`.

lib/src/repo.rs
yuja: Added tests.

The behavior isn't exactly the same because intermediate rewrite doesn't appear in negative terms of the conflict, but I think that's okay.

martinvonz (approved):

lib/src/repo.rs
martinvonz: Looks good. Thanks!

PR #4374, opened 2024-09-03
API code and evaluating RevsetExpression - RevsetExpressionEvaluator: extract resolve() from evaluate()

jonathantanmy:
This is probably part of a greater conversation about how API code (that in theory, doesn't need the `RevsetExtensions` and `IdPrefixContext` that CLI code needs) should manipulate and evaluate revision set expressions. Currently, we have `RevsetExpression` (widely used within jj code and has a lot of convenience functions) and `ResolvedExpression` (not widely used and does not have a lot of convenience functions - at work, in order to construct one from an input `ResolvedExpression`, I wrote out the enum instances directly and used `Box::new` a lot).

Should we change this so that it's easier to write API code? I can think of a few ways to do it:

- Declare that API code must support `RevsetExtensions` and `IdPrefixContext` (and maybe make a way to easily instantiate `RevsetExpressionEvaluator` with defaults for both)
- Keep both `RevsetExpression` and `ResolvedExpression`, and add convenience methods to construct `ResolvedExpression` as we need them
- Add a phantom type to `RevsetExpression` that indicates whether it's resolved or not, and gradually remove uses of `ResolvedExpression`. Convenience methods to construct instances can operate on either phantom type, and other methods will be specialized to the relevant type.

It seems to me that the last option is best, but perhaps I'm missing something.

(commit message follows)

This was noticed at work while implementing an API, when I needed to resolve an expression without evaluating it (because the result of resolution needs to be part of another expression, and is not used directly as an iterator of commit IDs).

# Checklist

If applicable:
- n/a I have updated `CHANGELOG.md`
- n/a I have updated the documentation (README.md, docs/, demos/)
- n/a I have added tests to cover my changes

### DIFF
--- a/cli/src/revset_util.rs
+++ b/cli/src/revset_util.rs
@@ -25,6 +25,7 @@ use jj_lib::id_prefix::IdPrefixContext;
 use jj_lib::repo::Repo;
 use jj_lib::revset;
 use jj_lib::revset::DefaultSymbolResolver;
+use jj_lib::revset::ResolvedExpression;
 use jj_lib::revset::Revset;
 use jj_lib::revset::RevsetAliasesMap;
 use jj_lib::revset::RevsetCommitRef;
@@ -90,14 +91,19 @@ impl<'repo> RevsetExpressionEvaluator<'repo> {
         self.expression = self.expression.intersection(other);
     }
 
-    /// Evaluates the expression.
-    pub fn evaluate(&self) -> Result<Box<dyn Revset + 'repo>, UserRevsetEvaluationError> {
+    /// Resolves the expression.
+    pub fn resolve(&self) -> Result<ResolvedExpression, UserRevsetEvaluationError> {
         let symbol_resolver = default_symbol_resolver(
             self.repo,
             self.extensions.symbol_resolvers(),
             self.id_prefix_context,
         );
-        evaluate(self.repo, &symbol_resolver, self.expression.clone())
+        resolve(self.repo, &symbol_resolver, self.expression.clone())
+    }
+
+    /// Evaluates the expression.
+    pub fn evaluate(&self) -> Result<Box<dyn Revset + 'repo>, UserRevsetEvaluationError> {
+        evaluate_resolved(self.repo, &self.resolve()?)
     }
 
     /// Evaluates the expression to an iterator over commit ids. Entries are
@@ -175,17 +181,31 @@ pub fn load_revset_aliases(
     Ok(aliases_map)
 }
 
+fn resolve(
+    repo: &dyn Repo,
+    symbol_resolver: &DefaultSymbolResolver,
+    expression: Rc<RevsetExpression>,
+) -> Result<ResolvedExpression, UserRevsetEvaluationError> {
+    revset::optimize(expression)
+        .resolve_user_expression(repo, symbol_resolver)
+        .map_err(UserRevsetEvaluationError::Resolution)
+}
+
+fn evaluate_resolved<'a>(
+    repo: &'a dyn Repo,
+    resolved_expression: &ResolvedExpression,
+) -> Result<Box<dyn Revset + 'a>, UserRevsetEvaluationError> {
+    resolved_expression
+        .evaluate(repo)
+        .map_err(UserRevsetEvaluationError::Evaluation)
+}
+
 pub fn evaluate<'a>(
     repo: &'a dyn Repo,
     symbol_resolver: &DefaultSymbolResolver,
     expression: Rc<RevsetExpression>,
 ) -> Result<Box<dyn Revset + 'a>, UserRevsetEvaluationError> {
-    let resolved = revset::optimize(expression)
-        .resolve_user_expression(repo, symbol_resolver)
-        .map_err(UserRevsetEvaluationError::Resolution)?;
-    resolved
-        .evaluate(repo)
-        .map_err(UserRevsetEvaluationError::Evaluation)
+    evaluate_resolved(repo, &resolve(repo, symbol_resolver, expression)?)
 }
 
 /// Wraps the given `IdPrefixContext` in `SymbolResolver` to be passed in to


yuja:
> * Declare that API code must support `RevsetExtensions` and `IdPrefixContext` (and maybe make a way to easily instantiate `RevsetExpressionEvaluator` with defaults for both)
> 
> * Keep both `RevsetExpression` and `ResolvedExpression`, and add convenience methods to construct `ResolvedExpression` as we need them
> 
> * Add a phantom type to `RevsetExpression` that indicates whether it's resolved or not, and gradually remove uses of `ResolvedExpression`. Convenience methods to construct instances can operate on either phantom type, and other methods will be specialized to the relevant type.
> 
> It seems to me that the last option is best, but perhaps I'm missing something.

I think `RevsetExpression` could have some type parameter that represents whether the expression is resolved. (I assume the last option is similar to this, where `T` is a opaque token type rather than `Unresolved(T)`.)
```rust
enum RevsetExpressionInner<T> {
    ..
    Unresolved(T),
}
enum Unresolved {
    Symbol(String),
    ..
}
type RevsetExpression = RevsetExpressionInner<Unresolved>;
type ResolvedExpression = RevsetExpressionInner<()>;
```

> This was noticed at work while implementing an API, when I needed to resolve an expression without evaluating it (because the result of resolution needs to be part of another expression, and is not used directly as an iterator of commit IDs).

Is that related to the extension API? Could you elaborate why an expression has to be partially resolved?

jonathantanmy:
> I think `RevsetExpression` could have some type parameter that represents whether the expression is resolved. (I assume the last option is similar to this, where `T` is a opaque token type rather than `Unresolved(T)`.)
> 
> ```rust
> enum RevsetExpressionInner<T> {
>     ..
>     Unresolved(T),
> }
> enum Unresolved {
>     Symbol(String),
>     ..
> }
> type RevsetExpression = RevsetExpressionInner<Unresolved>;
> type ResolvedExpression = RevsetExpressionInner<()>;
> ```

Yes, exactly. https://doc.rust-lang.org/rust-by-example/generics/phantom.html - see the PhantomStruct part. The documentation uses `f32` and `f64`, but we can use `Unresolved` and `()` (like in your example) or maybe `Unresolved` and `Resolved`.

I was thinking of `RevsetExpression<Unresolved>` and `RevsetExpression<Resolved>` but don't feel very strongly about the naming.

> > This was noticed at work while implementing an API, when I needed to resolve an expression without evaluating it (because the result of resolution needs to be part of another expression, and is not used directly as an iterator of commit IDs).
> 
> Is that related to the extension API? Could you elaborate why an expression has to be partially resolved?

By extension API, do you mean the revset extension API (the `extensions` parameter in the constructor of `RevsetExpressionEvaluator`)? If so, then I think the answer is "no" and "yes". "No" because the API I'm thinking of is the interface that jj presents to code that wants to use its basic features like manipulating operations, changes, and commits (as opposed to the interface that jj presents to the end user who runs it through the `jj` CLI), not the interface jj presents to whoever/whatever that wants a more flexible way of evaluating revset expressions. "Yes" because a user of this jj API would probably not want to do things like use the revset extension API.

As for why an expression has to be partially resolved, I'm not sure if "partially resolved" is the right term - we are fully resolving it; it's just that we are not evaluating it (I'm using the terms "resolve" and "evaluate" like the functions in `RevsetExpression`). Our use case is we want to refactor some code to be usable both by a CLI interface and an internal API. We want CLI users to be able to specify commits through the existing revset expression language, but our API users already have commit IDs (so they don't need evaluation of revset expressions). But whether we have revset expressions or commit IDs, we are not operating only on those commits - as an implementation detail, we look at all mutable commits reachable from them (the exact algorithm is subject to change, but the main point is that we are looking at more commits than what's specified as input). So I'm planning to resolve the user-provided expression (parsed into a RevsetExpression) into a ResolvedExpression and only use ResolvedExpression in the common code.

yuja:
> As for why an expression has to be partially resolved, I'm not sure if "partially resolved" is the right term - we are fully resolving it; it's just that we are not evaluating it (I'm using the terms "resolve" and "evaluate" like the functions in `RevsetExpression`). Our use case is we want to refactor some code to be usable both by a CLI interface and an internal API. We want CLI users to be able to specify commits through the existing revset expression language, but our API users already have commit IDs (so they don't need evaluation of revset expressions). But whether we have revset expressions or commit IDs, we are not operating only on those commits - as an implementation detail, we look at all mutable commits reachable from them (the exact algorithm is subject to change, but the main point is that we are looking at more commits than what's specified as input). So I'm planning to resolve the user-provided expression (parsed into a RevsetExpression) into a ResolvedExpression and only use ResolvedExpression in the common code.

Thanks for the details. I think what we need is a `RevsetExpression` that has no `RevsetExpression::CommitRef(_)`?

The current expression transformation stages are as follows:

1. parse user expression or construct `RevsetExpression` programmatically
2. apply basic tree substitution (called `optimize()` in revset.rs)
3. resolve `RevsetExpression::CommitRef(_)` (`resolve_symbols()` in revset.rs)
4. insert `visible_heads()` or `all()` where implied (`resolve_visibility()` in revset.rs)

`ResolvedExpression` is constructed at 4. I think internal API callers would want to do 2 and 4.

```rust
// In CLI, revset text is parsed
let user_expr: Rc<RevsetExpression> = ..;
// and symbols have to be resolved
let expr: Rc<RevsetExpression> = user_expr.resolve_symbol(repo, symbol_resolver)?;
// Non-CLI users can bypass symbol resolution, and pass expression containing no user symbols
expr.evaluate_programmatic() // will do basic tree substitution and resolve_visibility()
```

The last part isn't type safe, and we might want to introduce `RevsetExpression<SymbolResolved|Unresolved>` to ensure that `evaluate_programmatic()` is never called without symbol resolution.

jonathantanmy:
> Thanks for the details. I think what we need is a `RevsetExpression` that has no `RevsetExpression::CommitRef(_)`?

Yes. And also no `RevsetExpression::Filter(_)` with `RevsetFilterPredicate::Extension(_)`, I believe.

> The current expression transformation stages are as follows:
> 
> 1. parse user expression or construct `RevsetExpression` programmatically
> 2. apply basic tree substitution (called `optimize()` in revset.rs)
> 3. resolve `RevsetExpression::CommitRef(_)` (`resolve_symbols()` in revset.rs)
> 4. insert `visible_heads()` or `all()` where implied (`resolve_visibility()` in revset.rs)
> 
> `ResolvedExpression` is constructed at 4. I think internal API callers would want to do 2 and 4.

Ah, that's true.

> ```rust
> // In CLI, revset text is parsed
> let user_expr: Rc<RevsetExpression> = ..;
> // and symbols have to be resolved
> let expr: Rc<RevsetExpression> = user_expr.resolve_symbol(repo, symbol_resolver)?;
> // Non-CLI users can bypass symbol resolution, and pass expression containing no user symbols
> expr.evaluate_programmatic() // will do basic tree substitution and resolve_visibility()
> ```
> 
> The last part isn't type safe, and we might want to introduce `RevsetExpression<SymbolResolved|Unresolved>` to ensure that `evaluate_programmatic()` is never called without symbol resolution.

Yes, if we used phantom types, functions will need to receive or return `RevsetExpression<SymbolResolved>`, `RevsetExpression<Unresolved>`, or `RevsetExpression<a>`, depending on what kind of expression they are working with.

Also, what do you or anyone else think of the code in this PR? I understand that eventually we will need a new `RevsetExpression`, but we will still need a `resolve` function in the future. So this code is useful now, and will be useful (with modifications) in the future.

yuja:
> > Thanks for the details. I think what we need is a `RevsetExpression` that has no `RevsetExpression::CommitRef(_)`?
> 
> Yes. And also no `RevsetExpression::Filter(_)` with `RevsetFilterPredicate::Extension(_)`, I believe.

Maybe you mean a symbol resolved by extension? I don't think `RevsetFilterPredicate::Extension(_)` can be substituted at this stage.

> Also, what do you or anyone else think of the code in this PR? I understand that eventually we will need a new `RevsetExpression`, but we will still need a `resolve` function in the future. So this code is useful now, and will be useful (with modifications) in the future.

Suppose the goal is to resolve user symbols, I think it's better to add `RevsetExpression::resolve_user_symbols()` or something, and wrap it by CLI helper as needed. The returned expression is still `RevsetExpression`, so all helper methods are available.

PR #4377, opened 2024-09-04
config: add color for renamed and copied in summary

scott2000:
Right now, renamed and copied files don't have any color in the output of `jj status`, and it makes them stand out. I think it's reasonable to color renamed files the same as modified files, since renaming is like modifying the path, and to color copied files the same as added files, since they're basically just added files that happen to have similar contents to an existing file.

Before this change:

No color

After this change (I couldn't get copied to show up):

Color on renamed

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/colors.toml
+++ b/cli/src/config/colors.toml
@@ -82,6 +82,8 @@
 "diff added" = { fg = "green" }
 "diff token" = { underline = true }
 "diff modified" = "cyan"
+"diff renamed" = "cyan"
+"diff copied" = "green"
 "diff access-denied" = { bg = "red" }
 
 "op_log id" = "blue"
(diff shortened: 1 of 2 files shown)


martinvonz (approved):
That makes sense to me. Thanks!

ilyagr:
This is only tangentially related and not blocking, but I wish `jj diff -s` could distinguish between a file that was renamed without changes and a file that was renamed and modified. If it could, we could only use color if the file was modified.

Git presents renamed+modified as follows:

```console
$ git s
## HEAD (no branch)
RM CHANGELOG.md -> CHANGELOG2.md
```

The R is green, the M is red, and the rest is uncolored:

image


I'm not sure how much I like this though. We could also have two lines, one about the rename and one about modifying CHANGELOG2.md.

PR #4379, opened 2024-09-04
cli/build-rs: remove metadata command dependency in favor of env vars

kalebpace:
When building this project with [Nix/Crane](https://github.com/ipetkov/crane/discussions/693), if the `jj-cli` dependency is specified in `Cargo.toml` as a git-based crate, `cargo vendor` splits this workspace up into sub-crate directories, which causes `cargo metadata` to fail when searching for relative deps in the workspace root. 

This PR simply changes how the crate version is determined, using Cargo's built-in environment variable [`CARGO_PKG_VERSION`](https://doc.rust-lang.org/cargo/reference/environment-variables.html)

Let me know if this is problematic change! 

# Checklist

If applicable:

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -47,9 +47,6 @@ required-features = ["test-fakes"]
 [[test]]
 name = "runner"
 
-[build-dependencies]
-cargo_metadata = { workspace = true }
-
 [dependencies]
 bstr = { workspace = true }
 chrono = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,38 +267,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "camino"
-version = "1.1.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b96ec4966b5813e2c0507c1f86115c8c5abaadc3980879c3424042a02fd1ad3"
-dependencies = [
- "serde",
-]
-
-[[package]]
-name = "cargo-platform"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "24b1f0365a6c6bb4020cd05806fd0d33c44d38046b8bd7f0e40814b9763cabfc"
-dependencies = [
- "serde",
-]
-
-[[package]]
-name = "cargo_metadata"
-version = "0.17.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e7daec1a2a2129eeba1644b220b4647ec537b0b5d4bfd6876fcc5a540056b592"
-dependencies = [
- "camino",
- "cargo-platform",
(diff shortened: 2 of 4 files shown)


martinvonz:
Nice! 

> When building this project with [Nix/Crane](https://github.com/ipetkov/crane/discussions/693), if the `jj-cli` dependency is specified in `Cargo.toml` as a git-based crate, `cargo vendor` splits this workspace up into sub-crate directories, which causes `cargo metadata` to fail when searching for relative deps in the workspace root.
> 
> This PR simply changes how the crate version is determined, using Cargo's built-in environment variable [`CARGO_PKG_VERSION`](https://doc.rust-lang.org/cargo/reference/environment-variables.html)

Can you include this part in the commit message, please? (See https://martinvonz.github.io/jj/latest/contributing/#code-reviews for reasoning.)

> Let me know if this is problematic change!

I don't think so. It think the author (I) probably just didn't know better.

kalebpace:
> Can you include this part in the commit message, please? (See https://martinvonz.github.io/jj/latest/contributing/#code-reviews for reasoning.)

Done! 

> I don't think so. It think the author (I) probably just didn't know better.

😅 Was a TIL for me if that helps

yuja:
We can also remove `cargo_metadata` from build-dependencies.

martinvonz:
> We can also remove `cargo_metadata` from build-dependencies.

Good point. Want to fix that too, @kalebpace?

cli/Cargo.toml
```diff
@@ -47,9 +47,6 @@ required-features = ["test-fakes"]
 [[test]]
 name = "runner"
 
-[build-dependencies]
-cargo_metadata = { workspace = true }
```
martinvonz, line 51: Ah, thanks! Can probably go from the root `Cargo.toml` too

kalebpace: Woops, removed from root as well now 👍

thoughtpolice: note: I do have a script I wrote to catch unused dependencies since `ouroboros` slipped through the cracks at one point, but it's on one of my branches: https://github.com/martinvonz/jj/blob/push-qvrktsnwpkzu/tools/scripts/unused_workspace_deps.py

I wonder if I should submit it upstream as a separate check or something.

thoughtpolice (approved):
Negative diffstat. lgtm

cli/Cargo.toml
martinvonz: Another option might be https://github.com/est31/cargo-udeps. Anyway, I'll resolve this comment now so this can be merged.

kalebpace:
So close. I think the [`windows-latest` job](https://github.com/martinvonz/jj/actions/runs/10695375862/job/29648703701?pr=4379) needs to be re-run?

kalebpace:
tyvm 🙏

PR #4381, opened 2024-09-04
Make backend methods async

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/store.rs
+++ b/lib/src/store.rs
@@ -262,8 +262,8 @@ impl Store {
         self.backend.read_symlink(path, id).await
     }
 
-    pub fn write_symlink(&self, path: &RepoPath, contents: &str) -> BackendResult<SymlinkId> {
-        self.backend.write_symlink(path, contents).block_on()
+    pub async fn write_symlink(&self, path: &RepoPath, contents: &str) -> BackendResult<SymlinkId> {
+        self.backend.write_symlink(path, contents).await
     }
 
     pub fn read_conflict(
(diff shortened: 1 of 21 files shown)


martinvonz:
Seems harmless but I'll wait until after 0.22 to merge this just in case

lib/src/store.rs
```diff
@@ -262,7 +262,7 @@ impl Store {
         self.backend.read_symlink(path, id).await
     }
 
-    pub fn write_symlink(&self, path: &RepoPath, contents: &str) -> BackendResult<SymlinkId> {
+    pub async fn write_symlink(&self, path: &RepoPath, contents: &str) -> BackendResult<SymlinkId> {
         self.backend.write_symlink(path, contents).block_on()
```
yuja, line 266: `.await`

yuja (approved):
Do we plan to remove sync version of `store.read`/`get_<thing>()` and drop `_async` suffix? The current function naming is inconsistent.

lib/src/store.rs
martinvonz: Done.

martinvonz:
> Do we plan to remove sync version of `store.read`/`get_<thing>()` and drop `_async` suffix? The current function naming is inconsistent.

Yes, I think we should. I also noticed that as I was working on this PR.

PhilipMetzger:
I think it's worthwhile to convert the `get_commit()` to return a future and instead add a `get_commit_sync()` helper for those who still need a sync method. The second method is only needed, if we have enough callers for it.

PR #4399, opened 2024-09-05
Fix panic when parsing conflict markers

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+ * Fixed panic when parsing invalid conflict markers of a particular form.
+   ([#2611](https://github.com/martinvonz/jj/pull/2611))
+
 ## [0.21.0] - 2024-09-04
 
 ### Breaking changes
--- a/lib/src/conflicts.rs
+++ b/lib/src/conflicts.rs
@@ -456,7 +456,7 @@ fn parse_conflict_hunk(input: &[u8]) -> Merge<BString> {
                     removes.last_mut().unwrap().extend_from_slice(rest);
                     adds.last_mut().unwrap().extend_from_slice(rest);
                 } else {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+ * Fixed panic when parsing conflict markers of a particular form.
```
ilyagr, line 20: I'd add the word "invalid" before conflict markers

lib/src/conflicts.rs
```diff
@@ -473,7 +473,12 @@ fn parse_conflict_hunk(input: &[u8]) -> Merge<BString> {
         }
     }
 
-    Merge::from_removes_adds(removes, adds)
+    if adds.len() == removes.len() + 1 {
+        Merge::from_removes_adds(removes, adds)
+    } else {
+        // Doesn't look like a conflict
```
ilyagr, line 479: I'd say something like "We cannot parse these conflict markers" or "Doesn't look like a valid conflict"

lib/tests/test_conflicts.rs
```diff
@@ -748,6 +748,28 @@ fn test_parse_conflict_different_wrong_arity() {
     )
 }
 
+#[test]
+// TODO: Should *not* panic
+#[should_panic]
+fn test_parse_conflict_missing_removes() {
+    assert_eq!(
```
ilyagr, line 753: Assuming this is correct, I'd add a comment here:

> `jj` should never produce conflict markers like this, but we don't want to panic while parsing them.

ilyagr (approved):
Great, thank you!

(All of my comments are optional)

CHANGELOG.md
martinvonz: Done

lib/src/conflicts.rs
martinvonz: Done (here and above)

PR #4401, opened 2024-09-05
fix: Add `enabled` config for fix tools

jennings:
As inspired by @ilyagr: https://discord.com/channels/968932220549103686/969291218347524238/1276278608729608243

Adds an optional `fix.tools.TOOL.enabled` config that disables use of a fix tool (if omitted, the tool is enabled). This is useful for defining tools in the user's configuration without enabling them for all repositories:

```toml
# ~/.jjconfig.toml
[fix.tools.rustfmt]
enabled = false
command = ["rustfmt", "--emit", "stdout"]
patterns = ["glob:'**/*.rs'"]
```

Then to enable it in a repository:

```shell
$ jj config set --repo fix.tools.rustfmt.enabled true
```

This also allows Jujutsu to ship some built-in tool configurations that are disabled by default.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New `git.sign-on-push` config option to automatically sign commits which are being
   pushed to a Git remote.
 
+* New `fix.tools.TOOL.enabled` config option to enable/disable tools. This is
+  useful for defining disabled tools in user configuration that can be enabled
+  in individual repositories with one config setting.
+
 ### Fixed bugs
 
 * Fixed diff selection by external tools with `jj split`/`commit -i FILESETS`.
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -85,6 +85,9 @@ use crate::ui::Ui;
 ///    empty, no files will be affected by the tool. If there are multiple
 ///    patterns, the tool is applied only once to each file in the union of the
 ///    patterns.
+///  - `enabled`: Enables or disables the tool. If omitted, the tool is enabled.
+///    This is useful for defining disabled tools in user configuration that can
+///    be enabled in individual repositories with one config setting.
 ///
 /// For example, the following configuration defines how two code formatters
 /// (`clang-format` and `black`) will apply to three different file extensions
@@ -407,6 +410,8 @@ struct ToolConfig {
     command: CommandNameAndArgs,
     /// The matcher that determines if this tool matches a file.
     matcher: Box<dyn Matcher>,
+    /// Whether the tool is enabled
+    enabled: bool,
     // TODO: Store the `name` field here and print it with the command's stderr, to clearly
     // associate any errors/warnings with the tool and its configuration entry.
 }
@@ -424,6 +429,12 @@ struct ToolsConfig {
 struct RawToolConfig {
     command: CommandNameAndArgs,
     patterns: Vec<String>,
+    #[serde(default = "default_tool_enabled")]
+    enabled: bool,
+}
+
+fn default_tool_enabled() -> bool {
+    true
 }
 
 /// Parses the `fix.tools` config table.
@@ -432,7 +443,7 @@ struct RawToolConfig {
 /// not check for issues that might still occur later like missing executables.
 /// This is a place where we could fail earlier in some cases, though.
 fn get_tools_config(ui: &mut Ui, settings: &UserSettings) -> Result<ToolsConfig, CommandError> {
-    let tools: Vec<ToolConfig> = settings
+    let mut tools: Vec<ToolConfig> = settings
         .table_keys("fix.tools")
         // Sort keys early so errors are deterministic.
         .sorted()
@@ -458,11 +469,18 @@ fn get_tools_config(ui: &mut Ui, settings: &UserSettings) -> Result<ToolsConfig,
             Ok(ToolConfig {
                 command: tool.command,
                 matcher: expression.to_matcher(),
+                enabled: tool.enabled,
             })
         })
         .try_collect()?;
     if tools.is_empty() {
-        Err(config_error("No `fix.tools` are configured"))
+        return Err(config_error("No `fix.tools` are configured"));
+    }
+    tools.retain(|t| t.enabled);
+    if tools.is_empty() {
+        Err(config_error(
+            "At least one entry of `fix.tools` must be enabled.".to_string(),
+        ))
     } else {
         Ok(ToolsConfig { tools })
     }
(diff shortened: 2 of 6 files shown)


martinvonz:
Sounds good. Mercurial has a similar feature. @hooper might want to review this

cli/src/commands/fix.rs
```diff
@@ -465,8 +471,15 @@ fn get_tools_config(ui: &mut Ui, config: &config::Config) -> Result<ToolsConfig,
...
             .sorted_by(|a, b| a.0.cmp(&b.0))
-            .map(|(_name, value)| -> Result<ToolConfig, CommandError> {
-                let tool: RawToolConfig = value.try_deserialize()?;
+            .filter_map(
+                |(_name, value): (_, _)| match value.try_deserialize::<RawToolConfig>() {
+                    // tool is disabled, omit it entirely
+                    Ok(tool) if !tool.enabled => None,
+                    result => Some(result),
+                },
+            )
```
yuja, line 480: nit: you can do `.map(|(_name, value)| value.try_deserialize..).filter_ok(|tool| tool.enabled)`.

docs/config.md
```diff
@@ -769,6 +769,27 @@ command = ["head", "-n", "10"]
...
+repositories.
+
+In the user configuration, define a disabled tool for running rustfmt:
+
+```toml
+[fix.tools.rustfmt]
+enabled = false
+command = ["rustfmt", "--emit", "stdout"]
+patterns = ["glob:'**/*.rs'"]
+```
```
yuja, line 785: nit: can you also update `config-schema.json` and doc string of `FixArgs`?

yuja:
> This also allows Jujutsu to ship some built-in tool configurations that are disabled by default.

Since `enabled` is optional field, adding builtin config with `enabled = false` can break user-defined tool of the same name. We have to be careful.

jennings:
> Since `enabled` is optional field, adding builtin config with `enabled = false` can break user-defined tool of the same name. We have to be careful.

Oh, that's a very good point, I didn't think of that possibility.

cli/src/commands/fix.rs
jennings: Thank you, I was struggling trying to find the most readable way to do this.

docs/config.md
jennings: Whoops, I didn't realize that's what config-schema.json was for (it's obvious in retrospect). Updated.

ilyagr:
> > Since `enabled` is optional field, adding builtin config with `enabled = false` can break user-defined tool of the same name. We have to be careful.
> 
> Oh, that's a very good point, I didn't think of that possibility.

Me neither. Very good point!

The best solution I came up with so far would be to split `fix.tools` into `fix.default-tools` and `fix.tools`, and encourage users to only add tools to the latter. In fact, we could make `enabled` default to `true` in `fix.tools` (or we could not even allow `enabled` in there) and have it default to `false` in `fix.default-tools`.

We could, of course, also default `enabled` to `false` in general, but the only way I can think of making this not too annoying is if we encourage people to use `jj fix` as `jj fix --tool rustfmt --tool another_tool` (so only a few tools are enabled by default). I don't think we want that, though.

yuja:
This might be a rejected idea, but we can add a list of tool names to be used, and the default list will be constructed from `fix.tools` of `config_source != Default`.

Oh, the default `fix.tools.<name>.enabled` could be derived from `config_source != Default`, but that might be too cryptic?

ilyagr:
> Oh, the default fix.tools.<name>.enabled could be derived from config_source != Default, but that might be too cryptic?

I didn't know that was an option. If we stick with Stephen's current approach, we could use this to print a warning if it looks like our `enabled=false` overrides the user's config.

jennings:
Maybe the simplest solution would be for any built-in fix tools to be named `builtin-*`: `builtin-prettier`, `builtin-rustfmt`, etc.? Then it's reasonable to assume we wouldn't conflict with anything the user configured themselves.

yuja:
> Maybe the simplest solution would be for any built-in fix tools to be named `builtin-*`: `builtin-prettier`, `builtin-rustfmt`, etc.? Then it's reasonable to assume we wouldn't conflict with anything the user configured themselves.

Yes. It's not fancy, but works.

@hooper any comments?

I'm also interested in whether or not something like the following was rejected. It will solve the ordering issue, but it's not fancy either.
> we can add a list of tool names to be used, and the default list will be constructed from fix.tools of config_source != Default.

jennings:
> I'm also interested in whether or not something like the following was rejected. It will solve the ordering issue, but it's not fancy either.
> 
> > we can add a list of tool names to be used, and the default list will be constructed from fix.tools of config_source != Default.

I think it’s a good idea, but was hoping to defer the work when/if we actually want to ship built-in tools.

yuja:
> was hoping to defer the work when/if we actually want to ship built-in tools.

That's reasonable, but the list (of enabled tool names) will replace `<tool>.enabled` flag, so I'm asking.

cli/src/commands/fix.rs
```diff
@@ -484,9 +493,18 @@ fn get_tools_config(ui: &mut Ui, settings: &UserSettings) -> Result<ToolsConfig,
         .table_keys("fix.tools")
         // Sort keys early so errors are deterministic.
         .sorted()
-        .map(|name| -> Result<ToolConfig, CommandError> {
+        .filter_map(
+            |name| -> Option<Result<(RawToolConfig, &str), CommandError>> {
+                match settings.get::<RawToolConfig>(["fix", "tools", name]) {
+                    Ok(tool) if tool.enabled => Some(Ok((tool, name))),
+                    Ok(_) => None,
+                    Err(err) => Some(Err(err.into())),
```
yuja, line 501: nit: type annotation can be omitted if `err` wasn't converted here

That said, it might be better to parse tools including disabled ones. The error message for `tools_config.tools.is_empty()` would be a bit confusing if all tools are disabled.

cli/tests/test_fix_command.rs
```diff
@@ -187,6 +187,48 @@ fn test_config_multiple_tools_with_same_name() {
...
+        enabled = true
+        command = ["{formatter}", "--lowercase"]
+        patterns = ["bar"]
+
+        [fix.tools.tool-3]
+        enabled = false
+        command = ["{formatter}", "--lowercase"]
+        patterns = ["baz"]
+        "###,
+        formatter = escaped_formatter_path.as_str()
```
yuja, line 215: nit: `to_toml_value()` can be used instead

CHANGELOG.md
> This also allows Jujutsu to ship some built-in tool configurations that are
> disabled by default.

I think this is no longer valid. Can you remove or update?

yuja (approved):
Thanks. There's no comment from @hooper, so I marked this as accepted.

cli/src/commands/fix.rs
jennings: I added an `enabled` property to `ToolConfig` so I could check for all tools being disabled after checking for no tools being configured.

PR #4402, opened 2024-09-06
signing: allow disabling by setting backend to empty string

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -776,6 +776,8 @@ GnuPG or SSH signing keys.
 
 To do this you need to configure a signing backend.
 
+Setting the backend to `"none"` disables signing.
+
 ### GnuPG Signing
 
 ```toml
--- a/lib/src/settings.rs
+++ b/lib/src/settings.rs
@@ -267,7 +267,8 @@ impl UserSettings {
     // separate from sign_settings as those two are needed in pretty different
     // places
     pub fn signing_backend(&self) -> Option<String> {
-        self.config.get_string("signing.backend").ok()
+        let backend = self.config.get_string("signing.backend").ok()?;
+        (backend.as_str() != "none").then_some(backend)
     }
 
     pub fn sign_settings(&self) -> SignSettings {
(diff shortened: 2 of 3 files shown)


martinvonz:
I should make `signing.backend` also default to `""` and document that. I'll make that change

martinvonz:
> I should make `signing.backend` also default to `""` and document that. I'll make that change

Done

lib/src/settings.rs
```diff
@@ -267,7 +267,11 @@ impl UserSettings {
     // separate from sign_settings as those two are needed in pretty different
     // places
     pub fn signing_backend(&self) -> Option<String> {
-        self.config.get_string("signing.backend").ok()
+        let backend = self
+            .config
+            .get_string("signing.backend")
```
ilyagr, line 272: I wonder whether

   .get_string(...)?;

would be clearer than `unwrap_or_default` (assuming it works, of course).

ilyagr (approved):

lib/src/settings.rs
martinvonz: I'm not sure it's clearer but it's at least shorter (and not less clear). Good idea.

docs/config.md
```diff
@@ -776,6 +776,8 @@ GnuPG or SSH signing keys.
 
 To do this you need to configure a signing backend.
 
+Setting the backend to `""` (empty string) disables signing.
```
yuja, line 779: nit: might be better to use `"none"`. `signing.backend` is enum-like, and is similar to `core.fsmonitor`.

lib/src/settings.rs
ilyagr: Another option is `.get_string(...).ok().filter(|x| x != "none")`.

I'll unresolve the conversation, but feel free to use it or not.

martinvonz: I went with `.ok()?` and a subsequent `.then_some()`

docs/config.md
martinvonz: Makes sense. Done.

PR #4403, opened 2024-09-06
demos: fixup, working_copy.png wasn't regenerated properly

ilyagr:
The "redacted output" wasn't properly redacted.

Also removed extraneous comment, and added a comment since
the new PNGs are larger than I expected.


martinvonz (approved):

ilyagr:
Thank you!

I am annoyed that using the new ImageMagick command made the PNGs larger than they should be, but it's too late now.

PR #4408, opened 2024-09-06
cli: implement `workspace add --sparse-patterns`

thoughtpolice:
This flag implements three modes:

- `copy`: copy sparse patterns from parent
- `none`: do not copy sparse patterns from parent
- `empty`: clear all paths, equal to `set --clear`

This is useful for various tooling like tools that want to run a parallel
process that queries the build system (without running into locks/blocking.)

I think continuing to copy sparse patterns makes sense as the default behavior.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   can be used to more easily introduce automatic formatting changes in a new
   commit separate from other changes.
 
+* `jj workspace add` now accepts a `--sparse-patterns=<MODE>` option, which
+  allows control of the sparse patterns for a newly created workspace: `copy`
+  (inherit from parent; default), `full` (full working copy), or `empty` (the
+  empty working copy).
+
 ### Fixed bugs
 
  * Fixed panic when parsing invalid conflict markers of a particular form.
--- a/cli/src/commands/workspace/add.rs
+++ b/cli/src/commands/workspace/add.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::fmt::Display;
 use std::fs;
 
 use itertools::Itertools;
@@ -31,9 +32,31 @@ use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
+/// How to handle sparse patterns when creating a new workspace.
+#[derive(clap::ValueEnum, Clone, Debug, Eq, PartialEq)]
+enum SparseInheritance {
+    /// Copy all sparse patterns from the current workspace.
+    Copy,
+    /// Include all files in the new workspace.
+    Full,
+    /// Clear all files from the workspace (it will be empty).
+    Empty,
+}
+
+impl Display for SparseInheritance {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            SparseInheritance::Copy => write!(f, "copy"),
+            SparseInheritance::Full => write!(f, "full"),
+            SparseInheritance::Empty => write!(f, "empty"),
+        }
+    }
+}
+
 /// Add a workspace
 ///
-/// Sparse patterns will be copied over from the current workspace.
+/// By default, the new workspace inherits the sparse patterns of the current
+/// workspace. You can override this with the `--sparse-patterns` option.
 #[derive(clap::Args, Clone, Debug)]
 pub struct WorkspaceAddArgs {
     /// Where to create the new workspace
@@ -58,6 +81,9 @@ pub struct WorkspaceAddArgs {
     /// new r1 r2 r3 ...`.
     #[arg(long, short)]
     revision: Vec<RevisionArg>,
+    /// How to handle sparse patterns when creating a new workspace.
+    #[arg(long, default_value_t=SparseInheritance::Copy)]
+    sparse_patterns: SparseInheritance,
 }
 
 #[instrument(skip_all)]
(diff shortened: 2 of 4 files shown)


cli/src/commands/workspace/add.rs
```diff
@@ -114,19 +119,22 @@ pub fn cmd_workspace_add(
...
+            .working_copy()
+            .sparse_patterns()?
+            .to_vec();
+        locked_ws
+            .locked_wc()
+            .set_sparse_patterns(sparse_patterns)
+            .map_err(|err| internal_error_with_message("Failed to set sparse patterns", err))?;
+        let operation_id = locked_ws.locked_wc().old_operation_id().clone();
+        locked_ws.finish(operation_id)?;
+    }
```
PhilipMetzger, line 137: Nit: use the if expression pattern here, as it makes it in my opinion more readable. 

```rust
  let sparse_patterns = if args.no_sparse_pattern {
    Vec::new()
  } else {
   // old code here
  };
```

PhilipMetzger (approved):
LG, minor nit remaining tho.

cli/src/commands/workspace/add.rs
```diff
@@ -58,6 +59,10 @@ pub struct WorkspaceAddArgs {
     /// new r1 r2 r3 ...`.
     #[arg(long, short)]
     revision: Vec<RevisionArg>,
+    /// Do not copy sparse patterns from the current workspace into the new
+    /// workspace.
+    #[arg(long)]
+    no_sparse_patterns: bool,
```
martinvonz, line 65: I would actually guess that this flag would give you an empty list of sparse patterns. Maybe we want to make it `--sparse-patterns=copy/full/empty` or something like that?

cli/src/commands/workspace/add.rs
PhilipMetzger: which would make it a really small diff.

thoughtpolice: So this doesn't work; my testsuite actually fails with this. I think the problem is that the empty sparse pattern set is equivalent to `--clear`, while a reset state i.e. no patterns at all is equivalent to `.`. That's probably a bit too much to change for now.

PhilipMetzger: Shouldn't that be `RepoPathBuf::root()`? I think it should work with that.

martinvonz: Yes, that's correct.

> while a reset state i.e. no patterns at all

It's actually not "no patterns at all". The full stop you see on line 105 represents the root directory (and implicitly everything under it).

martinvonz: I think it's better to keep it the way you have it. The new working copy includes all patterns by default. Seems unnecessary to lock the working copy again just to write the same value to it.

PhilipMetzger: To be clear my suggested diff would be: 

```suggestion
    
        let (mut locked_ws, _wc_commit) = new_workspace_command.start_working_copy_mutation()?;
        let sparse_patterns = if args.no_sparse_patterns { 
           // A root pattern is equivalent to not setting a pattern at all. 
           // TODO: improve this?
            vec![RepoPathBuf::root()]
        } else { 
             old_workspace_command
            .working_copy()
            .sparse_patterns()?
            .to_vec();
       };
            locked_ws
            .locked_wc()
            .set_sparse_patterns(sparse_patterns)
            .map_err(|err| internal_error_with_message("Failed to set sparse patterns", err))?;
        let operation_id = locked_ws.locked_wc().old_operation_id().clone();
        locked_ws.finish(operation_id)?;
    }
```

martinvonz: Yes. My argument was that if `args.no_sparse_patterns` is true, then lines 126 and 137-142 are unnecessary since they're just setting the sparse patterns to what they already were. Right?

That would change, however, if we always created new working copies without any sparse patterns, which could be a reasonable change to make. Or we could say that the caller always has to specify the sparse patterns when they create the working copy.

PhilipMetzger: > Yes. My argument was that if `args.no_sparse_patterns` is true, then lines 126 and 137-142 are unnecessary since they're just setting the sparse patterns to what they already were. Right?

Yes, you're correct and I'm not contesting it, I just wanted to show the full example. 

> That would change, however, if we always created new working copies without any sparse patterns, which could be a reasonable change to make. Or we could say that the caller always has to specify the sparse patterns when they create the working copy.

My suggestion would definitely be better for a later change which you suggested here. It probably is out of scope for this PR though.

cli/src/commands/workspace/add.rs
thoughtpolice: Good idea. I've got a patch incoming that will do all this.

thoughtpolice:
I quite like the new `--sparse-patterns` solution, though we could keep going further. Why can't I specify patterns at `add` time, after all? And if I can do that with workspaces, why not clone? But, let's go one step at a time...

CHANGELOG.md
```diff
@@ -24,6 +24,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   can be used to more easily introduce automatic formatting changes in a new
   commit separate from other changes.
 
+* `jj workspace add` now accepts a `--sparse-patterns=<MODE>` option, which
+  allows control of the sparse patterns for a newly created workspace: `copy`
+  (inherit from parent; default), `none` (no patterns), or `empty` (the empty
+  working copy pattern).
```
martinvonz, line 30: There's actually no "empty working copy pattern"; sparse patterns don't support general filesets. That's instead represented by an empty list of paths to include. Relatedly, I still find `none` confusing. I interpret it the same was as `empty`. I think `full` is clearer (what it means in practice is that we use `[RepoPathBuf::root()]`).

cli/src/commands/workspace/add.rs
```diff
@@ -31,9 +31,34 @@ use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
+/// How to handle sparse patterns when creating a new workspace.
+#[derive(clap::ValueEnum, Clone, Debug, Eq, PartialEq)]
+enum SparseInheritance {
+    /// Copy sparse patterns from the current workspace.
+    Copy,
+    /// Do not copy sparse patterns from the current workspace, i.e. the
+    /// workspace will have a full working copy.
```
martinvonz, line 41: I'd drop the "Do not copy sparse patterns... " bit. That's true for all but `Copy` after all.

martinvonz:
> I quite like the new `--sparse-patterns` solution, though we could keep going further. Why can't I specify patterns at `add` time, after all? And if I can do that with workspaces, why not clone? But, let's go one step at a time...

I think the most important feature to have is a way of not including more files than necessary to start with because that can't be worked around. That's important if your working copy is huge, and especially if your files are fetched from a remote. So I think the most important feature is the new `--sparse-patterns=empty` version. Well, except that we already had that functionality with `-r 'root()'`. Still, `--sparse-patterns=empty` seems like a much more natural solution to look for.

martinvonz:
Sorry, my mouse go stuck causing me to click "Close with comment" by mistake

thoughtpolice:
@martinvonz Good input, I changed the name from `none` to `full`, which makes much more sense versus `empty`. PTAL

cli/src/commands/workspace/add.rs
```diff
@@ -31,9 +32,33 @@ use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
+/// How to handle sparse patterns when creating a new workspace.
+#[derive(clap::ValueEnum, Clone, Debug, Eq, PartialEq)]
+enum SparseInheritance {
+    /// Copy sparse patterns from the current workspace.
+    Copy,
+    /// Use the default sparse pattern for a "full" repository, including all
+    /// files in the workspace.
```
martinvonz, line 41: nit: maybe s/repository/working copy/

Also, it might be better to say simply say "Include all files in the working copy". That way it's still true even if we change the default.

martinvonz (approved):
Thanks!

cli/src/commands/workspace/add.rs
thoughtpolice: I reworded to make it all a little simpler. Thanks!

PR #4413, opened 2024-09-06
gitignore: add `/buck-out/` to ignore list

thoughtpolice:
While working on the Buck branch, I have been bitten about a million times by the following setup:

- Hack away
- Switch back to main
- Run `jj st` after editing a file
- `/buck-out` gets snapshotted, because it isn't ignored
- Trip max snapshot filesize error
- Pain

There are a few ways to skin the cat but this is simplest. It's an incredibly small addition and it may be all for naught in the long run, but this will stem the bleeding, at least.

# Checklist

If applicable:


istudyatuni:
You can just add to `.git/info/exclude`

thoughtpolice:
> You can just add to `.git/info/exclude`

Yes, but on every single repository I have to clone on all my machines everywhere, forever. Please see the previous discussion in https://github.com/martinvonz/jj/pull/4409 &mdash; I will admit don't like this solution, to a degree that is almost inexplicable (why add a thing this branch doesn't need, I guess?) but at the end of the day I want to *not* snapshot thousands of files on accident, every other day, in a foolproof way, and this change perhaps is uncontroversial.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target/
+/buck-out/
```
ilyagr, line 2: I would put this on the bottom and add a comment: `# To make working on ... easier`, perhaps with an issue id.

ilyagr (approved):

.gitignore
thoughtpolice: Done. I also went ahead and made the capitalization in comments here consistent.

Issue #3201, opened 2024-03-03
If `@` turns immutable, it can still be mutated


mlcui-corp:
## Description

If `@` somehow turns immutable - for example, if `immutable_heads()` contains `remote_branches()`, and `@` is pushed - `@` can still be mutated.

GitHub's "[Adding new commits](https://martinvonz.github.io/jj/prerelease/github/#adding-new-commits)" workflow discourages force-pushing new commits, so for those repositories I've added `remote_branches()` to `immutable_heads()`.

## Steps to Reproduce the Problem

1. Have `remote_branches()` in `immutable_heads()`.
1. `jj git push -c @ --remote mlcui`
1. Change the working copy.

## Expected Behavior

Changing the working copy does not work. I'm not too sure what the nicest way of doing so would be:

- Automatically creating a new empty change on top if `@` becomes immutable, then committing things to that.
- Having a divergent `@` change.

## Actual Behavior

`@` is modified, even though it is immutable.

## Specifications

- Platform: Ubuntu 22.04.2 LTS running under WSL
- Version: jj 0.14.0-82b3017fdaabbb60823ad133e37d980223054381

PhilipMetzger:
I think this is expected, as I wrote in the section: 
> Warning We strongly suggest to jj new after the example below, as all further edits still get amended to the previous commit.
>
> $ # Create a new commit on top of the `your-feature` branch from above.
> $ jj new your-feature
> $ # Address the comments by updating the code. Then review the changes.
> $ jj diff
> $ # Give the fix a description.
> $ jj describe -m 'address pr comments'
> $ # Update the branch to point to the current commit.
> $ jj branch set your-feature -r @
> $ # Push it to your remote
> $ jj git push

the commit will only be considered immutable after a `new` following the `fetch`. This is because there is no `sync` in-between.

martinvonz:
Yes, this is expected, but we can probably do better. I suppose one option is to check at the end of each command to see if any working-copy commit is now immutable and then create a new working-copy commit for it.

PhilipMetzger:
Yes, this seems to be a good choice, although I'd expect that we print a message before creating a new wc-commit.

martinvonz:
> I'd expect that we print a message before creating a new wc-commit.

Yes, I would also expect that.

mlcui-corp:
> I suppose one option is to check at the end of each command to see if any working-copy commit is now immutable

[...]

PR #4421, opened 2024-09-08
Make other working copies immutable

martinvonz:
I'm a bit uneasy about this because of the test changes in the last commit. It's clearly a corner case there (undoing `jj workspace forget` causing some commits to become immutable), but there may be similar cases. I was wondering if we should create the new working-copy commits when snapshotting instead. I thought we had talked about that in the context of #3201 but I don't see any mention of it there. Maybe we talked about it on Discord. Anyone remembers?

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* The working-copy commits of other workspaces are now considered immutable by
+  default. `builtin_immutable_heads()` now includes `other_working_copies()`,
+  and the default `jj log` revset shows all workspaces' working copies. Use
+  `--ignore-immutable` or redefine `immutable_heads()` to rewrite another
+  workspace's checked-out commit.
+
 ### Deprecations
 
 ### New features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   jj workspace can have its own Git HEAD. Existing repositories are migrated
   automatically.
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
yuja, line 1: > `working_copies() ~ @` is close, but it's not correct if the
> same commit is being edited in the current workspace and in another
> workspace.

Wouldn't it simpler to allow mutation in that case? If the wc commit is already editing, I wouldn't expect it to become immutable by editing the exact commit in other workspace.

yuja: > Since the goal is to avoid accidentally creating stale workspaces, we should avoid that also if another workspace is editing the same commit (or a descendant).

That's correct, but in order to edit the same commit, the user would have to use `--ignore-immutable edit` in another workspace, so I (as a user) wouldn't mind if it made the other workspace ignore immutability of the wc commit. (There might be other ways to enter such situation, but that's rare?)

martinvonz: It could also happen if you do `jj edit <parent>` in one workspace and `jj edit <child>` in another workspace. Going back to the first workspace would then result in it becoming immutable. That seems consistent to me. However, since we don't prevent snapshotting even if the commit is immutable, it's already not going to work very well for preventing stale workspaces in that case, so I think it makes sense to simply add `working_copies() ~ @` as I think you're suggesting. I'll update this PR to do just that.

seh:
There's some [related discussion in the "chat" channel](https://discord.com/channels/968932220549103686/968932220549103689/1363926916276687079) of the Discord server.

drieber:
What is the status of this PR, is it moving forward? Several multi-workspace scenarios run into divergence without this. Squash is one of those. Other things run into surprising issues that could be fixed by making the common ancestors of working copy commits immutable, for example `jj fix`.

PR #4422, opened 2024-09-08
cli: extract non-repo part from `WorkspaceCommandHelper`, prep for `op log --op-diff`

yuja:
# Checklist

If applicable:


martinvonz (approved):
Nice.

> We might want to extract some parts to helper function, but I don't have a good idea right now.

Yes, there was a bit of duplication in that patch, but hopefully it's clearer later how to extract some common parts.

PR #4427, opened 2024-09-09
Update `jj edit <commit>` to add commit into view heads if not already

kevincliao:
`jj new <commit>` automatically adds the checked out commits into the view head ids. However, `jj edit` does not.

To reproduce:
```
jj git init test
cd test
jj commit -m "my commit"
jj log -r @- -T commit_id # Save the id
jj abandon -r @-
jj edit <saved_id>

jj log -r :: # Does not show the currently editing commit 
```

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
  * Fixed panic when parsing invalid conflict markers of a particular form.
    ([#2611](https://github.com/martinvonz/jj/pull/2611))
 
+ * Editing a hidden commit now makes it visible.
+
 ## [0.21.0] - 2024-09-04
 
 ### Breaking changes
--- a/lib/tests/test_mut_repo.rs
+++ b/lib/tests/test_mut_repo.rs
@@ -319,6 +319,28 @@ fn test_edit_initial() {
 }
 
 #[test]
+fn test_edit_hidden_commit() {
+    // Test that MutableRepo::edit() edits a hidden commit and updates
+    // the view head ids.
+    let settings = testutils::user_settings();
+    let test_repo = TestRepo::init();
+    let repo = &test_repo.repo;
+
+    let mut tx = repo.start_transaction(&settings);
+    let wc_commit = write_random_commit(tx.repo_mut(), &settings);
+
+    // Intentionally not doing tx.commit, so the commit id is not tracked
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
  * Fixed panic when parsing invalid conflict markers of a particular form.
    ([#2611](https://github.com/martinvonz/jj/pull/2611))
+ * Fixed edit command sometimes not adding the working copy commit into visible heads.
```
martinvonz, line 43: I think we can describe more exactly when it happens without this becoming too verbose. Something like this:

> Editing a hidden commit now makes it visible

Do you think that's clear enough?

kevincliao: Ah, that's much better. Done.

martinvonz (commented):
LGTM, but as usual, I'll let a non-Googler approve it.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
  * Fixed panic when parsing invalid conflict markers of a particular form.
    ([#2611](https://github.com/martinvonz/jj/pull/2611))
+ * Editing a hidden commit now makes it visible.
```
yuja, line 43: nit: insert blank line before

lib/tests/test_mut_repo.rs
```diff
@@ -318,6 +318,33 @@ fn test_edit_initial() {
...
 
+#[test]
+fn test_edit_hidden_commit() {
+    // Test that MutableRepo::edit() edits a hidden commit and updates
+    // the view head ids.
+    let settings = testutils::user_settings();
+    let test_repo = TestRepo::init();
+    let repo = &test_repo.repo;
+
+    dbg!(repo.view().heads());
```
yuja, line 329: nit: remove `dbg!()`?

yuja (approved):

CHANGELOG.md
kevincliao: Done.

lib/tests/test_mut_repo.rs
kevincliao: Ah, thanks for spotting that. It would be nightmare to me if this was accidentally submitted.

PR #4429, opened 2024-09-10
Be more specific when warning the user about missing identity configs

Veykril:
Typoed `user.name` as `user.nane` and got confused why my entire user didn't get recognized, warning on each command. 

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1613,13 +1613,31 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
         }
 
         let settings = self.settings();
-        if settings.user_name().is_empty() || settings.user_email().is_empty() {
+        let missing_user_name = settings.user_name().is_empty();
+        let missing_user_mail = settings.user_email().is_empty();
+        if missing_user_name || missing_user_mail {
+            let mut writer = ui.warning_default();
+            let not_configured_msg = match (missing_user_name, missing_user_mail) {
+                (true, true) => "Name and email not configured.",
+                (true, false) => "Name not configured.",
+                (false, true) => "Email not configured.",
+                _ => unreachable!(),
+            };
+            write!(writer, "{not_configured_msg} ")?;
             writeln!(
-                ui.warning_default(),
-                r#"Name and email not configured. Until configured, your commits will be created with the empty identity, and can't be pushed to remotes. To configure, run:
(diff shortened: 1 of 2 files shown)


Veykril:
Any thoughts on this? Feels like a very simple QoL change

cli/src/cli_util.rs
```diff
@@ -1613,13 +1613,30 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
...
-  jj config set --user user.name "Some One"
-  jj config set --user user.email "someone@example.com""#
+                writer,
+                "{} not configured. Until configured, your commits will be created with the empty \
+                 identity, and can't be pushed to remotes. To configure, run:",
+                match (missing_user_name, missing_user_mail) {
+                    (true, true) => "Name and email",
+                    (true, false) => "Name",
+                    (false, true) => "Email",
+                    _ => unreachable!(),
```
yuja, line 1628: nit: can you split by sentence (to help grepping and possible i18n support)?
```rust
match .. {
    (true, true) => "Name and email not configured.",
    ...
}
```
btw, I prefer not inline large code block as expression.

yuja (approved):
Thanks!

cli/src/cli_util.rs
PhilipMetzger: My suggestion would've been to extract it into a variable above the usage.

cli/src/cli_util.rs
```diff
@@ -1613,13 +1613,31 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
...
+            };
+            write!(writer, "{not_configured_msg} ")?;
             writeln!(
-                ui.warning_default(),
-                r#"Name and email not configured. Until configured, your commits will be created with the empty identity, and can't be pushed to remotes. To configure, run:
-  jj config set --user user.name "Some One"
-  jj config set --user user.email "someone@example.com""#
+                writer,
+                "Until configured, your commits will be created with the empty identity, and \
+                 can't be pushed to remotes. To configure, run:",
```
PhilipMetzger, line 1630: This can be reverted back to a raw string right?

Veykril: This isn't a multiline string as was the case with the raw string before which contained two line breaks, so I don't quite see why this should be a raw string

PR #4432, opened 2024-09-10
config: introduce default `at = @` revset alias

thoughtpolice:
When using Windows, `@` is a bit of an annoyance to type in the terminal due to Windows Terminal escape rules, so you actually have to write it with a backtick in front. It's annoying to remember this and not possible to configure a different behavior, as far as I can see.

However, there is a smart trick you can do to make this a lot easier: just alias `@` to the name `at` and be happy. Because:

- It's short: sweet and easy to read, and requires no backticks.
- It's unambiguous: `a` is from commit alphabet, but `z` is from the change alphabet
- It can be typed without a huge loss in efficiency (and one hand, assuming QWERTY).

I believe it was Stephen Jennings who came up with this trick first in Discord, and I have it in my aliases, but I run into it every time I try to use JJ on a fresh Windows VM to test stuff. Let's make life a little easier for everyone and just ship this by default.

# Checklist

If applicable:


thoughtpolice:
There's no changelog or docs yet but I'll get to it shortly assuming nobody has strong feelings on this.

PhilipMetzger (approved):

jennings:
I am honored!

> but `z` is from the change alphabet

"z"? 🤔

cli/src/config/revsets.toml
```diff
@@ -22,3 +22,6 @@ latest(
 'immutable_heads()' = 'builtin_immutable_heads()'
 'immutable()' = '::(immutable_heads() | root())'
 'mutable()' = '~immutable()'
+
+# NOTE: for Windows users, who don't want to type `@ all the time
+'at' = '@'
```
yuja, line 27: Symbol aliases overrides tags and branches, so builtin aliases have to be all functions. And `at()` wouldn't be easy to type.

Maybe add doc example instead?

yuja (changes requested):

cli/src/config/revsets.toml
thoughtpolice: Well, that's unfortunately a pretty big hole in the side of the ship, so to speak. A doc example is at least in order though.

I also feel like this is something we should should have tests for i.e. iff there exists any keys in `[revset-aliases]` in any of the builtin configs, and they are not functions, then panic &mdash; since I wouldn't have thought of this without your suggestion. So maybe I should fix that, too.

emilazy: Is there a more verbose way to specify a branch/tag specifically? I feel like `at` will not be a common name for those, so as long as there’s an override it might be okay.

yuja: `"at"` or `tags(exact:at)` works, but less special case is better. It would be weird if `trunk` was a builtin alias and not pointing to the `trunk` branch. The `at` alias looks similar.

I also think that, to work around PowerShell, it's better to find another symbol (e.g. `.`) to replace `@`, instead of adding a builtin alias which only PowerShell users would practically use.

martinvonz: One reason I didn't pick `.` for the working copy was that it seemed like it would be confusing to former Mercurial users (`.` is the *parent* of the working copy there). Also, the `@` symbol works well in the graph log output. I'm not sure `.` would work as well there. It might be hard to notice, and it might be hard to realize the correspondence with the revset symbol. Were you thinking of shipping `.` as an alias or to eventually have it replace `@`?

Shipping `at` by default on Windows might be okay if we think only (a subset of) Windows users would use it.

thoughtpolice: I have also filed #4437 in order to prevent this from happening again, or until we decide otherwise.

yuja: > Shipping `at` by default on Windows might be okay if we think only (a subset of) Windows users would use it.

I think it's rather annoying for people who use both Windows and Mac.

ilyagr: I wonder if we should have some syntax for "tag or branch `at`" if `at` is aliased, whether by us or by the user.

For example, `#at` could represent that and `##at` would represent a branch named literally `#at` (if that's legal). Or, we could be more fancy and have something along the lines of `r"at"` or `r##"at"##`. I also considered `.at` (which I like, but conflicts with `..at`) and `,at`, which wouldn't require quoting (though I don't think quoting is harmful to have here).

It'd be also be nice if we could then make most of our revset functions accessible without parentheses, e.g. `conflicts` instead of `conflicts()`.

ilyagr: > I also think that, to work around PowerShell, it's better to find another symbol (e.g. `.`) to replace `@`, instead of adding a builtin alias which only PowerShell users would practically use.

I think using `.` for this would make `...` ambiguous with our current syntax: `(.)..` vs `..(.)`.

Also, I like `@` quite a bit, though I'd prefer a world where `at` also worked.

jennings: Suppose it were possible to unset a configuration.

    # hypothetical syntax
    revset-aliases.at = null

Would this make shipping `at` in the default Windows config acceptable? Then the few people who do have a branch or tag with that name that they need to regularly refer to can unset the alias.

yuja: > I wonder if we should have some syntax for "tag or branch `at`" if `at` is aliased,

`"at"` (quoted)

It also works within function arguments, e.g. `description("at")` to disable `at` -> `@` substitution.

> It'd be also be nice if we could then make most of our revset functions accessible without parentheses, e.g. `conflicts` instead of `conflicts()`.

If we disable unquoted branch/tag symbols, that's probably doable. I think it would be messy to mix builtin functions and user ref namespaces because ref namespace is often updated by fetching remote changes.

> Suppose it were possible to unset a configuration.
> [snip]
> Would this make shipping `at` in the default Windows config acceptable?

I would say no. It's confusing if `at` were interpreted differently across platforms. It wouldn't be uncommon that Windows devs use PowerShell and WSL.

I think the `at` alias is nice if it is an opt-in feature, and that basically means documenting how to add the alias. And the user might want to make it even shorter (e.g. `_ = @`.)

martinvonz: What do people think about Yuya's proposal to simply document the alias? Seems fine to at least start with that and see how many reports we get after that?

Btw, if the `@` symbol is annoying for PowerShell users, then our use of it in remote bookmarks and workspace identifiers will also be a problem, although a much less frequent one. I suppose one possible alternative syntax that works particularly well for remote bookmarks is `origin/main`, but I'm not sure I like `/` for the working copy and `/other` for the working copy in workspace `other` as much.

jennings: It’s okay in those contexts because it isn’t the first character in a token. `main@origin` isn’t a problem because PowerShell doesn’t interpret that as the start of an array since it’s after a word character.

martinvonz: I see. `@other` (which refers to the working-copy in workspace `other`) is still a problem then.

yuja: Oh, so `default@`, `..@`, etc. work. The workspace id syntax is `<id>@`, btw.

ilyagr: > What do people think about Yuya's proposal to simply document the alias?

We actually already document it, sort of, in https://martinvonz.github.io/jj/latest/windows/#typing-in-powershell. That recommends `HEAD`, which is not as good as `at` IMO.

Should we make it more prominent? I think it'd make sense to explain it in the main config doc and then link to it again from the Windows doc.

Separately (and not blocking the more obvious improvements), should we explain the `"at"` syntax for a branch/tag next to where we describe revset aliases?

Separately again, I still wonder about the larger issue. While we reduce the problem of overlaps with branch names if we don't define `at` ourselves, it'd be nice to have a solution that also works for users who configured it themselves (and maybe forget). But we can ponder and/or discuss this question long after we fix the easy-to-fix things.

PR #4437, opened 2024-09-11
cli: add test to ensure the default config has no symbol aliases

thoughtpolice:
In #4432 I submitted a patch to ship a symbol alias in the default configuration. However, Yuya quickly pointed out that this isn't acceptable because symbols override all branch and trunk names, meaning that the name `at` would have at best been taken  away from users, and at worst been very confusing for them down the road had they ever tried to use it.

I should have thought of this myself (I encountered it long ago when first using Jujutsu, actually) but didn't. Such is life.

But this didn't cause any tests to fail, as it was not immediately clear this would impede anything or cause downstream behavioral changes; the problems it adds are actually latent and it's very possible someone will want to make the same mistake in the future. And Yuya might not be able to stop them (me) in time; requiring him to go out of his way and then handle the fallout. I'm not a fan of this.

Instead, let's patch this latent hole for good, or until we decide otherwise, by applying the Beyonce Rule: "if you liked that behavior, then you shoulda put a test on it."

# Checklist

If applicable:


thoughtpolice:
@yuja The "is this a symbol" check is a complete hack, but maybe that's OK. Let me know what you think.

cli/src/config.rs
```diff
@@ -1270,6 +1270,38 @@ mod tests {
...
+            //
+            // but can never be something like:
+            //
+            //     'foo)' = ...
+            //     '(foo)' = ...
+            //
+            // and so testing for the presence of a closing parenthesis is
+            // sufficient to ensure we're not introducing a symbol.
+            if !alias.ends_with(")") {
+                anyhow::bail!(
```
thoughtpolice, line 1293: Should just be `return`

yuja:
> @yuja The "is this a symbol" check is a complete hack, but maybe that's OK. Let me know what you think.

That's good. We could also use `RevsetAliasesMap::insert()` and `function_names()`.

That said, I think a comment like `# Don't add default symbol aliases because ...` in `config/revsets.toml` is good enough.

PR #4442, opened 2024-09-12
Fix broken revset formatting

matts1:
The indentation has been breaking the previews on github, since it treats it as if the whole example, including the dot points, is a code block.

# Checklist

If applicable:

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -78,96 +78,96 @@ only symbols.
 You can use parentheses to control evaluation order, such as `(x & y) | z` or
 `x & (y | z)`.
 
-??? examples
+### Examples
 
-    Given this history:
-    ```
-    D
-    |\
-    | o C
-    | |
-    o | B
-    |/
-    o A
-    |
-    o root()
-    ```
+Given this history:
+```
+D
+|\
+| o C
+| |
+o | B
+|/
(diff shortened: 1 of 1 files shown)


matts1:
Ah, never mind, looknig at the blame layer it appears this was intentional ([commit](https://github.com/martinvonz/jj/commit/a6d470dbc302ca3c0930b4a0315ea068d9d42026)).

"This makes the examples in
https://martinvonz.github.io/jj/prerelease/revsets/#operators render
properly at the cost of worse (but readable) rendering on GitHub."

martinvonz (commented):
@ilyagr: Will MkDocs be happy with the new format?

matts1:
My commit ended up being, roughly speaking, a revert of the original change, so I'd suspect not. It would certainly be nice if there was something we could do to have it preview nicely on both.

ilyagr:
The main advantage of the current way is that the Examples are collapsed by default. If we're willing to sacrifice that, Matt's formatting will look fine in both, but scrolling down the document will be more exhausting.

I have an unfinished commit that turns those examples into a table, which is more compact. However, they don't fit into a table perfectly since there's some variation of which revsets are nice to show for which example. So, I'm not sure about it.

PR #4444, opened 2024-09-12
Highlight change ids in rebase advice

Veykril:
Fixes https://github.com/martinvonz/jj/issues/3561

# Checklist

If applicable:
  - I don't know if that needs mention as a new feature or not, let me know if it does
- [x] I have ~~added~~updated tests to cover my changes

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1290,6 +1290,11 @@ impl WorkspaceCommandHelper {
             .expect("parse error should be confined by WorkspaceCommandHelper::new()")
     }
 
+    pub fn short_change_id_template(&self) -> TemplateRenderer<'_, Commit> {
+        self.parse_commit_template("format_short_change_id(self.change_id())")
+            .expect("parse error should be confined by WorkspaceCommandHelper::new()")
+    }
+
     /// Returns one-line summary of the given `commit`.
     ///
     /// Use `write_commit_summary()` to get colorized output. Use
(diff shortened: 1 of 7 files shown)


martinvonz (approved):
Thanks! This should make my demo at Git Merge next week a tiny bit prettier :)

cli/src/cli_util.rs
```diff
@@ -1290,6 +1290,11 @@ impl WorkspaceCommandHelper {
             .expect("parse error should be confined by WorkspaceCommandHelper::new()")
     }
 
+    pub fn format_short_change_id(&self) -> TemplateRenderer<'_, Commit> {
```
yuja, line 1293: nit: `format_()` sounds like a function that returns formatted object. maybe rename to `short_change_id_template()`?

cli/src/cli_util.rs
```diff
@@ -1290,6 +1290,11 @@ impl WorkspaceCommandHelper {
...
+        self.parse_commit_template("format_short_change_id(self.change_id())")
+            .expect("syntax is valid")
```
yuja, line 1295: `format_short_change_id()` can be overridden by user, and may have syntax error.

Veykril: Ah okay, I thought it only parsed this snippet specifically (anything else being caught upfront)

martinvonz:
> I don't know if that needs mention as a new feature or not, let me know if it does

Nah, I don't think so. I think in general, it makes sense to mention things that fit at least one of:

* Would break or surprise existing users' workflows
* Users would appreciate but would perhaps not notice if we don't mention it
* Are significant enough that users who don't yet use jj are encouraged to finally try it (or is enough to convince them to try it again)

There may be other good reasons, and that's just my opinion anyway.

PR #4445, opened 2024-09-12
Report what operation has been undone in `jj op undo`

Veykril:
Fixes https://github.com/martinvonz/jj/issues/4431

# Checklist

If applicable:
- [x] I have ~~added~~updated tests to cover my changes

### DIFF
--- a/cli/src/commands/operation/diff.rs
+++ b/cli/src/commands/operation/diff.rs
@@ -36,7 +36,6 @@ use jj_lib::revset;
 use jj_lib::revset::RevsetIteratorExt as _;
 
 use crate::cli_util::short_change_hash;
-use crate::cli_util::short_operation_hash;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::LogContentFormat;
 use crate::command_error::CommandError;
@@ -123,42 +122,15 @@ pub fn cmd_op_diff(
         workspace_env.parse_template(&language, &text, CommitTemplateLanguage::wrap_commit)?
     };
 
+    let op_summary_template = workspace_command.operation_summary_template();
     ui.request_pager();
     let mut formatter = ui.stdout_formatter();
     formatter.with_label("op_log", |formatter| {
         write!(formatter, "From operation ")?;
-        write!(
-            formatter.labeled("id"),
-            "{}",
-            short_operation_hash(from_op.id()),
-        )?;
-        write!(formatter, ": ")?;
-        write!(
-            formatter.labeled("description"),
-            "{}",
-            if from_op.id() == from_op.op_store().root_operation_id() {
-                "root()"
-            } else {
-                &from_op.metadata().description
-            }
-        )?;
+        op_summary_template.format(&from_op, &mut *formatter)?;
         writeln!(formatter)?;
         write!(formatter, "  To operation ")?;
-        write!(
-            formatter.labeled("id"),
-            "{}",
-            short_operation_hash(to_op.id()),
-        )?;
-        write!(formatter, ": ")?;
-        write!(
-            formatter.labeled("description"),
-            "{}",
-            if to_op.id() == to_op.op_store().root_operation_id() {
-                "root()"
-            } else {
-                &to_op.metadata().description
-            }
-        )?;
+        op_summary_template.format(&to_op, &mut *formatter)?;
         writeln!(formatter)
     })?;
 
--- a/cli/src/commands/operation/undo.rs
+++ b/cli/src/commands/operation/undo.rs
@@ -21,6 +21,7 @@ use super::DEFAULT_UNDO_WHAT;
 use crate::cli_util::CommandHelper;
 use crate::command_error::user_error;
 use crate::command_error::CommandError;
+use crate::operation_templater::OperationTemplateLanguage;
 use crate::ui::Ui;
 
 /// Create a new operation that undoes an earlier operation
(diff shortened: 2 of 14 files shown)


PhilipMetzger (changes requested):
We now have a commit style, so please adhere to it with by prefixing the commit message with `undo:` or `cli:`. The rest are minor nits.

PhilipMetzger (approved):
LGTM

cli/tests/test_duplicate_command.rs
```diff
@@ -75,7 +75,12 @@ fn test_duplicate() {
 
     let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["undo"]);
     insta::assert_snapshot!(stdout, @"");
-    insta::assert_snapshot!(stderr, @"");
+    insta::assert_snapshot!(stderr, @r#"
+    Undid operation c4b0b2a977fee749cf7d857e915ba5ae70e48e10d2a7d53d76d27efab3184c342eaee24e2bc007c2e6714cdb53f703ae6d20912bd5f541e1b3fcb535bfaa62e6
+    c4b0b2a977fe test-username@host.example.com 2001-02-03 04:05:17.000 +07:00 - 2001-02-03 04:05:17.000 +07:00
+    duplicate 1 commit(s)
+    args: jj duplicate a
```
martinvonz, line 82: I think we may want a shorter format. I was going to suggest a single-line format similar to `commit_summary`. My concern is just that maybe the `args:` tag is useful enough that we can't really do that? What do you think? I'm thinking something like this (you'll have to imagine colors to separate the different fields):

> Undid operation c4b0b2a9 10 seconds ago, lasted 13 milliseconds duplicate 1 commit(s)

Veykril: I was wondering about that as well. The first line here is quite a mouthful at least (and I just realized this is now printing both the full op id and shortened, so thats redundant either way)

cli/src/commands/operation/undo.rs
```diff
@@ -70,5 +71,25 @@ pub fn cmd_op_undo(
...
+        let workspace_env = workspace_command.env();
+        let language = OperationTemplateLanguage::new(
+            bad_op.op_store().root_operation_id(),
+            Some(bad_op.id()),
+            workspace_env.operation_template_extensions(),
+        );
+        let text = command
+            .settings()
+            .config()
+            .get_string("template-aliases.builtin_op_undo")?;
```
yuja, line 85: nit: `[templates]`, not `[template-aliases]`

cli/tests/test_duplicate_command.rs
yuja: If we introduce single-line "operation_summary" (or "op_summary"), we can also update `cmd_op_diff()` to use that. If not, reusing "op_log" template is okay for now.

cli/src/commands/operation/undo.rs
Veykril: Why do the two exists? Is the former newer?

martinvonz: The thinking was that `[template-aliases]` is an open namespace where the user can define whatever alias they want but `[templates]` has a predefined set of keys, which means we can introduce a new key there later without having to worry that the user has already defined an alias with that name. However, we now the the `builtin-` aliases that are also predefined and implicitly not for the user to add to. So I think we have said that we could remove the `[templates]` section. Right, @yuja?

yuja: Since user shouldn't (conceptually) override `builtin_*` aliases, we'll need `<name> = builtin_<name>` indirection somewhere. That's `templates.<name> = builtin_<name>_<style>` right now.

iirc, `[templates]`/`[template-aliases]` distinction isn't super important. It's just copied from revsets where symbol aliases shadow tag/branch names, so no builtin symbol aliases should be added to `[revset-aliases]`.

Veykril:
Added an `op_summary` template now

cli/src/commands/operation/undo.rs
```diff
@@ -70,5 +71,26 @@ pub fn cmd_op_undo(
     tx.repo_mut().set_view(new_view);
     tx.finish(ui, format!("undo operation {}", bad_op.id().hex()))?;
 
+    if let Some(mut formatter) = ui.status_formatter() {
+        write!(formatter, "Undid operation ")?;
+        let workspace_env = workspace_command.env();
+        let language = OperationTemplateLanguage::new(
+            bad_op.op_store().root_operation_id(),
+            Some(bad_op.id()),
```
yuja, line 79: this should be `workspace_command.repo().op_id()`. It's usually the operation specified by `--at-op`, and here we've already committed the transaction, the operation after undo is the current operation.

cli/src/commands/operation/diff.rs
```diff
@@ -124,42 +124,40 @@ pub fn cmd_op_diff(
         workspace_env.parse_template(&language, &text, CommitTemplateLanguage::wrap_commit)?
     };
 
+    let from_op_summary_template = {
+        let workspace_env = workspace_command.env();
+        let language = OperationTemplateLanguage::new(
+            from_op.op_store().root_operation_id(),
+            Some(from_op.id()),
```
yuja, line 131: same here. `repo.op_id()`, and we don't need two separate template instances.

It might be better to add `workspace_command.operation_template_language()` as there are 3 callers.

yuja (approved):
Thanks!

PhilipMetzger (approved):

cli/src/commands/operation/undo.rs
Veykril: `workspace_command.repo().op_id()` is not a thing (`op_id()` on the repo specifically). `bad_op` should be correct though here no? It is the result parsed from the input `    let bad_op = workspace_command.resolve_single_op(&args.operation)?;`

cli/src/commands/operation/diff.rs
yuja: https://github.com/martinvonz/jj/pull/4471

cli/src/commands/operation/undo.rs
yuja: No, it's not. The `current_op_id` argument will be used in order to test if the context operation (= `bad_op`) is the head of the operation at the currently loaded operation (= `repo.operation()`).

`bad_op` is the operation to be reverted.

Veykril: Oh right

PR #4448, opened 2024-09-12
git_backend: Support shallow git repositories

Veykril:
Fixes https://github.com/martinvonz/jj/issues/675

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/clone.rs
+++ b/cli/src/commands/git/clone.rs
@@ -15,6 +15,7 @@
 use std::fs;
 use std::io;
 use std::io::Write;
+use std::num::NonZeroU32;
 use std::path::Path;
 use std::path::PathBuf;
 
@@ -60,6 +61,9 @@ pub struct GitCloneArgs {
     /// Whether or not to colocate the Jujutsu repo with the git repo
     #[arg(long)]
     colocate: bool,
+    /// Create a shallow clone of the given depth
+    #[arg(long)]
+    depth: Option<NonZeroU32>,
 }
 
 fn absolute_git_source(cwd: &Path, source: &str) -> String {
@@ -132,6 +136,7 @@ pub fn cmd_git_clone(
         ui,
         command,
         args.colocate,
+        args.depth,
         remote_name,
         &source,
         &canonical_wc_path,
@@ -195,6 +200,7 @@ fn do_git_clone(
     ui: &mut Ui,
     command: &CommandHelper,
     colocate: bool,
+    depth: Option<NonZeroU32>,
     remote_name: &str,
     source: &str,
     wc_path: &Path,
@@ -223,6 +229,7 @@ fn do_git_clone(
             &[StringPattern::everything()],
             cb,
             &command.settings().git_settings(),
+            depth,
         )
     })
     .map_err(|err| match err {
--- a/docs/git-compatibility.md
+++ b/docs/git-compatibility.md
@@ -57,8 +57,9 @@ a comparison with Git, including how workflows are different, see the
   not be lost either.
 * **Partial clones: No.** We use the [libgit2](https://libgit2.org/) library,
   which [doesn't have support for partial clones](https://github.com/libgit2/libgit2/issues/5564).
-* **Shallow clones: No.** We use the [libgit2](https://libgit2.org/) library,
-  which [doesn't have support for shallow clones](https://github.com/libgit2/libgit2/issues/3058).
+* **Shallow clones: Kind of.** Shallow commits all have the virtual root commit as
+  their parent. However, deepening or fully unshallowing a repository is currently not yet
+  supported and will cause issues.
 * **git-worktree: No.** However, there's native support for multiple working
   copies backed by a single repo. See the `jj workspace` family of commands.
 * **Sparse checkouts: No.** However, there's native support for sparse
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -21,6 +21,7 @@ use std::collections::HashSet;
 use std::default::Default;
 use std::fmt;
 use std::io::Read;
+use std::num::NonZeroU32;
 use std::path::PathBuf;
 use std::str;
 
@@ -1244,6 +1245,7 @@ pub fn fetch(
     branch_names: &[StringPattern],
     callbacks: RemoteCallbacks<'_>,
     git_settings: &GitSettings,
+    depth: Option<NonZeroU32>,
 ) -> Result<GitFetchStats, GitFetchError> {
     // Perform a `git fetch` on the local git repo, updating the remote-tracking
     // branches in the git repo.
@@ -1260,6 +1262,9 @@ pub fn fetch(
     fetch_options.proxy_options(proxy_options);
     let callbacks = callbacks.into_git();
     fetch_options.remote_callbacks(callbacks);
+    if let Some(depth) = depth {
+        fetch_options.depth(depth.get().try_into().unwrap_or(i32::MAX));
+    }
     // At this point, we are only updating Git's remote tracking branches, not the
     // local branches.
     let refspecs: Vec<_> = branch_names
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -519,6 +519,7 @@ fn commit_from_git_without_root_parent(
     id: &CommitId,
     git_object: &gix::Object,
     uses_tree_conflict_format: bool,
+    is_shallow: bool,
 ) -> BackendResult<Commit> {
     let commit = git_object
         .try_to_commit_ref()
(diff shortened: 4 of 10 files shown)


Veykril:
Have not tested it with partial repos yet (also only did some light manual testing). Unsure how to write a proper test for this, so if anyone has any ideas I am all ears :)

cli/src/commands/git/clone.rs
```diff
@@ -57,6 +57,9 @@ pub struct GitCloneArgs {
     /// Whether or not to colocate the Jujutsu repo with the git repo
     #[arg(long)]
     colocate: bool,
+    /// Create a shallow clone of the given depth.
+    #[arg(long)]
+    depth: Option<usize>,
```
PhilipMetzger, line 62: Nit: We don't end help message sentences with a period. And this could be 
```suggestion
   #[arg(long, default = 0)]
   depth: usize,
```

lib/src/git.rs
PhilipMetzger, line 1: Nit: I think this belongs to a separate `lib:`/`git:` commit or it could be squashed into the parent. This is for all touched files in `lib/`.

lib/src/git.rs
```diff
@@ -1240,6 +1240,7 @@ pub struct GitFetchStats {
 pub fn fetch(
     mut_repo: &mut MutableRepo,
     git_repo: &git2::Repository,
+    depth: usize,
```
PhilipMetzger, line 1243: Nit: Isn't a `Option<usize>` more  convienent here?

lib/src/git.rs
```diff
@@ -1260,6 +1261,7 @@ pub fn fetch(
     fetch_options.proxy_options(proxy_options);
     let callbacks = callbacks.into_git();
     fetch_options.remote_callbacks(callbacks);
+    fetch_options.depth(depth.try_into().unwrap_or(0));
     // At this point, we are only updating Git's remote tracking branches, not the
```
PhilipMetzger, line 1265: Then this could be wrapped into a if let block.

lib/src/git_backend.rs
PhilipMetzger, line 1: Optional: This commit could have the topic `git_backend:` if you want to be consistent with the files history.

docs/git-compatibility.md
```diff
@@ -57,8 +57,7 @@ a comparison with Git, including how workflows are different, see the
   not be lost either.
 * **Partial clones: No.** We use the [libgit2](https://libgit2.org/) library,
   which [doesn't have support for partial clones](https://github.com/libgit2/libgit2/issues/5564).
-* **Shallow clones: No.** We use the [libgit2](https://libgit2.org/) library,
-  which [doesn't have support for shallow clones](https://github.com/libgit2/libgit2/issues/3058).
+* **Shallow clones: Yes.** Shallow commits all have the virtual root commit as their parent.
```
PhilipMetzger, line 60: Nit: This still needs to wrap at 80.

docs/git-compatibility.md
PhilipMetzger, line 1: Nit: Also contains some unrelated changes again.

PhilipMetzger (changes requested):
Minor nits and some points for further discussion. I also think the initial PR description was good, but we don't really have guidelines here.

lib/src/git_backend.rs
```diff
@@ -537,10 +538,16 @@ fn commit_from_git_without_root_parent(
...
     );
-    let parents = commit
-        .parents()
-        .map(|oid| CommitId::from_bytes(oid.as_bytes()))
-        .collect_vec();
+    // shallow commits don't have parents their parents actually fetched, so we
+    // discard them here
+    let parents = if is_shallow {
+        vec![]
+    } else {
```
yuja, line 545: Maybe this means that the shallow-boundary commit was rewritten, its parents would be permanently lost?

That might be okay for experimental support of shallow clones, but we'll need to find a better workaround later.

docs/git-compatibility.md
Veykril: Dammit you have a lot of trailing whitespace in your repo 😅

lib/src/git_backend.rs
Veykril: As in, if someone tries to unshallow their repository we won't pick up those changes? I'll have to check but I imagine we can work around that by marking shallow commits or so.

Either way I'd also say for initial support that would be be probably fine.

docs/git-compatibility.md
martinvonz: IMO, cleanups are welcome, btw, but they should ideally be separate commits.

The better solution is to set up some markdown formatter that runs as a GitHub Action.

lib/src/git_backend.rs
martinvonz: Yes, not crashing is at least better than crashing.

If the repo is later unshallowed (or just deepened a bit), we would somehow want to detect that so we can rebuild our commit index. That seems to suggest some kind of weird interaction between the commit backend and the index backend that doesn't currently exist (the index backend is agnostic to the kind of commit backend, it doesn't treat the Git backend specially in any way).

yuja: > if someone tries to unshallow their repository we won't pick up those changes?

We might get panic somewhere because parents aren't indexed.

> Either way I'd also say for initial support that would be be probably fine.

+1. TODO comments will also help.

Veykril: Okay I just checked, we do in fact panic if the repo is deepened (`lib\src\default_index\composite.rs:472:49` in my testing just now)

Veykril:
Okay this is ready for proper review now. I'm leaving deepening/unshallowing for a separate issue (will open an issue for that after this is merged)

lib/src/git_backend.rs
```diff
@@ -859,6 +867,8 @@ fn import_extra_metadata_entries_from_heads(
     head_ids: &HashSet<&CommitId>,
     uses_tree_conflict_format: bool,
 ) -> BackendResult<()> {
+    let shallow_commits = git_repo.shallow_commits().ok().flatten();
```
yuja, line 870: nit: better to map error to `BackendError::Other`, but that can be addressed later. We'll get another read error anyway.

lib/tests/test_git.rs
```diff
@@ -3468,3 +3468,99 @@ ignoreThisSection = foo
...
+    let e = empty_git_commit(&git_repo, "refs/heads/main", &[&c]);
+
+    git_repo.set_head("refs/heads/main").unwrap();
+
+    let make_shallow = |repo, mut shallow_commits: Vec<_>| {
+        let shallow_file = get_git_backend(repo).git_repo().shallow_file();
+        shallow_commits.sort();
+        let mut buf = Vec::<u8>::new();
+        for commit in shallow_commits {
+            std::io::Write::write_fmt(&mut buf, format_args!("{commit}\n")).unwrap();
```
yuja, line 3513: nit: `use std::io::Write` and `writeln!()`.

lib/tests/test_git.rs
```diff
@@ -3468,3 +3468,99 @@ ignoreThisSection = foo
...
+    };
+    assert_eq!(*repo.view().heads(), expected_heads);
+
+    let parents = |store: &Arc<jj_lib::store::Store>, commit| {
+        let commit = store.get_commit(&jj_id(commit)).unwrap();
+        commit
+            .parents()
+            .map(Result::unwrap)
+            .map(|c| c.id().clone())
+            .collect_vec()
```
yuja, line 3537: nit: `commit.parent_ids().to_vec()`

lib/src/git.rs
```diff
@@ -1240,6 +1241,7 @@ pub struct GitFetchStats {
 pub fn fetch(
     mut_repo: &mut MutableRepo,
     git_repo: &git2::Repository,
+    depth: Option<NonZeroU32>,
```
yuja, line 1244: nit: I would insert `depth` argument somewhere around `callbacks` and `git_settings`. It's not a mandatory option, and we might want to extract a parameter struct later.

lib/src/git.rs
```diff
@@ -1260,6 +1262,9 @@ pub fn fetch(
     fetch_options.proxy_options(proxy_options);
     let callbacks = callbacks.into_git();
     fetch_options.remote_callbacks(callbacks);
+    if let Some(depth) = depth.and_then(|depth| depth.get().try_into().ok()) {
```
yuja, line 1265: nit: better to clamp at `i32::MAX` than ignoring overflow

cli/src/commands/git/clone.rs
```diff
@@ -60,6 +61,11 @@ pub struct GitCloneArgs {
     /// Whether or not to colocate the Jujutsu repo with the git repo
     #[arg(long)]
     colocate: bool,
+    /// Create a shallow clone of the given depth
+    // We accept a depth of 0 for parity with git's --depth option.
+    // The option is here to not render a default = 0 in the help message.
```
yuja, line 66: nit: I don't think we need to copy the exact git CLI behavior. If `--depth 0` doesn't make sense, I would make it an error.

cli/src/commands/git/clone.rs
yuja, line 1: Can you add CLI tests of `jj git clone --depth`?

yuja (commented):
LGTM, thanks!

PhilipMetzger (approved):
LG after addressing Yuya's nits.

cli/src/commands/git/clone.rs
Veykril: It looks like local clones do not support shallow clones in git2, yielding
> shallow fetch is not supported by the local transport

yuja: Ok, it wouldn't be easy to write tests, then. (we could test that non-zero `--depth` is passed to libgit2, though.)

Thanks for checking.

lib/src/git.rs
PhilipMetzger, line 1: Nit: squash into previous.

Veykril: Squash which into which, all 3 commits touch this file

PhilipMetzger: These changes address Yuya's nits from the second commit, so move them there.

PR #4450, opened 2024-09-12
cli: Support renaming workspaces

### LINKED ISSUE #4342: FR: Allow renaming of existing workspaces
# Motivation

The _jj workspace add_ accepts a `--name` flag to specify a name other than the basename of the directory. If I create a workspace without passing the `--name`, or I do so but later decide that I no longer like that name, I'd like to be able to **rename the existing workspace**.

The purpose that this workspace serves may change over time, or my preferred way of remembering it may change, but at present I have to create a new workspace in order to get one with the name that I desire.

# Proposed Solution

Consider introducing a _jj workspace rename_ command that requires two names: the existing workspace name and the proposed workspace name. _jj_ could treat the two names being equal as acceptable, requiring no further action, or it could reject such an invocation.

_jj_ would also have to ensure that no workspace already exists with the proposed workspace name.

# Alternatives Considered

It's possible to use _jj workspace forget_ and then _jj workspace add_—or even skip the "forgetting" step altogether—to wind up with a workspace with the desired name, assuming that one doesn't mind placing this new workspace in different directory than an existing workspace.

The proposed subcommand may be too specific. I considered whether _jj workspace edit_ might be useful, accepting flags like `--new-name` to specify _how_ one wishes to edit an existing workspace. However, there aren't enough facets of a workspace today to warrant this generality.

# Context

I first [asked about this capability gap in the "questions" channel](https://discord.com/channels/968932220549103686/1277066793973710919/1277066793973710919) of the "JJ" Discord server.

kevincliao:
fixes #4342

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/workspace/rename.rs
+++ b/cli/src/commands/workspace/rename.rs
@@ -1,0 +1,78 @@
+// Copyright 2020 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use jj_lib::op_store::WorkspaceId;
+use tracing::instrument;
+
+use crate::cli_util::CommandHelper;
+use crate::command_error::user_error;
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Renames the current workspace
+#[derive(clap::Args, Clone, Debug)]
+pub struct WorkspaceRenameArgs {
+    /// The name of the workspace to update to.
+    new_workspace_name: String,
+}
+
+#[instrument(skip_all)]
+pub fn cmd_workspace_rename(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &WorkspaceRenameArgs,
+) -> Result<(), CommandError> {
+    if args.new_workspace_name.is_empty() {
+        return Err(user_error("New workspace name cannot be empty"));
+    }
+
+    let mut workspace_command = command.workspace_helper(ui)?;
+
+    let old_workspace_id = workspace_command.working_copy().workspace_id().clone();
+    let new_workspace_id = WorkspaceId::new(args.new_workspace_name.clone());
+    if new_workspace_id == old_workspace_id {
+        writeln!(ui.status(), "Nothing changed.")?;
+        return Ok(());
+    }
+
+    if workspace_command
+        .repo()
+        .view()
+        .get_wc_commit_id(&old_workspace_id)
+        .is_none()
+    {
+        return Err(user_error(format!(
+            "The current workspace '{}' is not tracked in the repo.",
+            old_workspace_id.as_str()
+        )));
+    }
+
+    let mut tx = workspace_command.start_transaction().into_inner();
+    let (mut locked_ws, _wc_commit) = workspace_command.start_working_copy_mutation()?;
+
+    locked_ws
+        .locked_wc()
+        .rename_workspace(new_workspace_id.clone());
+
+    tx.repo_mut()
+        .rename_workspace(&old_workspace_id, new_workspace_id)?;
+    let repo = tx.commit(format!(
+        "Renamed workspace '{}' to '{}'",
+        old_workspace_id.as_str(),
+        args.new_workspace_name
+    ));
+    locked_ws.finish(repo.op_id().clone())?;
+
+    Ok(())
+}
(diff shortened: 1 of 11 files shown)


cli/src/commands/workspace/rename.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    let view = workspace_command.repo().view();
+    if view.get_wc_commit_id(&new_workspace_id).is_some() {
+        return Err(user_error(format!(
+            "Workspace named '{}' already exists",
+            args.new_workspace_name
+        )));
+    }
+    let wc_commit_id = view.get_wc_commit_id(&old_workspace_id).unwrap().clone();
```
yuja, line 56: this can fail if the workspace was orphaned by `workspace forget` or `op restore`.

cli/src/commands/workspace/rename.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    let op_id = workspace_command.repo().op_id().clone();
+    let (mut locked_ws, _) = workspace_command.start_working_copy_mutation()?;
+    locked_ws
+        .locked_wc()
+        .rename_workspace(new_workspace_id.clone());
+    locked_ws.finish(op_id)?;
```
yuja, line 63: the lock should be released after `tx.finish()`. see `cmd_file_track`/`untrack()` for example.

cli/src/commands/workspace/rename.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    let mut tx = workspace_command.start_transaction();
+    tx.repo_mut()
+        .set_wc_commit(new_workspace_id.clone(), wc_commit_id)?;
+    tx.repo_mut().remove_wc_commit(&old_workspace_id)?;
```
yuja, line 68: nit: might be better to add `rename_workspace()` function. We might add `git_heads: HashMap<WorkspaceId, RefTarget>` to the view.

kevincliao: Done.

cli/src/commands/workspace/rename.rs
kevincliao: Done. Thanks for the pointer.

cli/src/commands/workspace/rename.rs
kevincliao: Oh I haven't thought about this. Updated the code and added a test case about that.

cli/src/commands/workspace/rename.rs
```diff
@@ -0,0 +1,83 @@
...
+        .rename_workspace(new_workspace_id.clone());
+
+    tx.repo_mut()
+        .rename_workspace(&old_workspace_id, new_workspace_id, wc_commit_id);
+    tx.commit(format!(
+        "Renamed workspace '{}' to '{}'",
+        old_workspace_id.as_str(),
+        args.new_workspace_name
+    ));
+    locked_ws.finish(op_id)?;
```
yuja, line 80: `tx.commit()` creates new repo/operation, and its `op_id()` should be recorded here.

lib/src/repo.rs
```diff
@@ -1353,6 +1353,17 @@ impl MutableRepo {
         Ok(())
     }
 
+    pub fn rename_workspace(
+        &mut self,
+        old_workspace_id: &WorkspaceId,
+        new_workspace_id: WorkspaceId,
+        wc_commit_id: CommitId,
```
yuja, line 1360: I think `wc_commit_id` argument is redundant. We just need to `wc_commit_ids.remove()`/`remove_entry()` and `.insert()` the removed entry.

yuja (approved):
Thanks!

cli/src/commands/workspace/rename.rs
kevincliao: Done

lib/src/repo.rs
kevincliao: Good idea. I added a `RenameWorkspaceError` type to handle the case where wc_commit_ids does not contain the workspace id. Let me know if there's a better way to do this.

lib/src/repo.rs
```diff
@@ -1353,6 +1353,22 @@ impl MutableRepo {
...
     }
 
+    pub fn rename_workspace(
+        &mut self,
+        old_workspace_id: &WorkspaceId,
+        new_workspace_id: WorkspaceId,
+    ) -> Result<(), RenameWorkspaceError> {
+        let wc_commit_id = self
+            .view_mut()
+            .remove_wc_commit(old_workspace_id)
```
yuja, line 1363: nit: maybe better to add `view.rename_workspace()` instead of making `.remove_wc_commit()` return old data?

lib/src/repo.rs
yuja: `.rename_remote()` ignores error, but I like the explicit error. Maybe we should also check `new_workspace_id` does not exist (which will replace precondition in CLI.)

kevincliao: Done.

lib/src/repo.rs
kevincliao: Sounds good. Done. (also moved `RenameWorkspaceError` to view.rs)

seh:
Thank you, Kevin!

Issue #2562, opened 2023-11-09
FR: `jj util stale diff` and `jj util stale status`


necauqua:
**Is your feature request related to a problem? Please describe.**

There are two points for this:
  - Some things, like startship or maybe the intellij plugin I was experimenting with, need to know the state of the jj repo constantly, and I don't want to snapshot the repo every second - I know in some situations that's fine, but often you want more control, especially in cases when editing .gitignore, etc
  - I will never get over the auto-amend happenning on every single command including readonly ones, when often I feel like I don't need it. `--ignore-working-copy` _could've_ helped (at least in log it does), but the whole point is that I want to see changes I just did but not amend them, leaving the repo stale until I run some meaningful command and not obsess over running diff/status every second making snapshots.
  For now I often run `git status` and `git diff` (and also starship shows git numbers, but there are different numbers between untracked and staged which is wonky), but I'd love to use jj only.

**Describe the solution you'd like**

`jj util stale diff [-r @]` - `jj diff` but between the @ (or other) and the physical state of the files in working copy, `--ignore-working-copy` (meaning no snapshot made) implied.

It is under the util subcommand (and not, say, `jj diff --stale`) because it essentially is git staging, which I feel like you wouldn't want to have among first-class commands, this is an "advanced", "low-level-ish" feature (useful, again, in things like starship) - although if you're ok with `jj diff --stale`, so am I.
Maaaybe there could be a `jj stale [verb]` set of commands, which is "officially discouraged" as it goes against the amend-everything principle :)
Or maybe I'll write a `jj-stale` program that will just do all of that and be unofficial, and could be called by `jj stale` when we enable external subcommands

Another thing would be `jj util stale status` (`jj stale status`? :) ) which is just like `jj status`, but does not snapshot - the difference from `jj status --ignore-working-copy` is that here the latter ignores the stale changes (it just works with the wc commit which wasn't amended), the stale-status should show the same thing `jj status` would've shown if the snapshot _did_ happen - except it shouldn't

martinvonz:
I think the best way of doing that would be with a global `--no-commit-transaction` flag. We would still create the commit and add the `refs/jj/keep-*` ref, but if we don't commit the transaction, it won't be visible afterwards. We would need to implement GC (#12) for the `refs/jj/keep-*` to get GC'd. Alternatively, if we implement #50, we could postpone creating the `refs/jj-keep*` refs until the transaction commits (which means we wouldn't create the refs with `--no-commit-transaction`).

[...]

PR #4457, opened 2024-09-13
cli: add global option to not commit transaction

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj op log` gained an option to include operation diffs.
 
+* A new global flag `--no-commit-transaction` lets you run a command without
+  impacting the repo state or the working copy.
+
 ### Fixed bugs
 
  * Fixed panic when parsing invalid conflict markers of a particular form.
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -346,6 +346,23 @@ impl CommandHelper {
         )?)
     }
 
+    pub fn should_commit_transaction(&self) -> bool {
+        !self.global_args().no_commit_transaction
+    }
+
+    pub fn maybe_commit_transaction(
+        &self,
+        tx: Transaction,
+        description: impl Into<String>,
+    ) -> Arc<ReadonlyRepo> {
+        let unpublished_op = tx.write(description);
+        if self.should_commit_transaction() {
+            unpublished_op.publish()
+        } else {
+            unpublished_op.leave_unpublished()
+        }
+    }
+
     pub fn workspace_loader(&self) -> Result<&dyn WorkspaceLoader, CommandError> {
         self.data
             .maybe_workspace_loader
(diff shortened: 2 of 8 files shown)


cli/src/cli_util.rs
```diff
@@ -839,7 +856,11 @@ impl WorkspaceCommandHelper {
             // state to it without updating working copy files.
             locked_ws.locked_wc().reset(&new_git_head_commit)?;
             tx.repo_mut().rebase_descendants(command.settings())?;
-            self.user_repo = ReadonlyUserRepo::new(tx.commit("import git head"));
+            self.user_repo = ReadonlyUserRepo::new(
+                self.env
+                    .command
+                    .maybe_commit_transaction(tx, "import git head"),
+            );
             locked_ws.finish(self.user_repo.repo.op_id().clone())?;
```
yuja, line 864: Is it safe to finalize working-copy modification here?

cli/src/cli_util.rs
```diff
@@ -1606,10 +1633,11 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
             print_failed_git_export(ui, &refs)?;
         }
 
-        self.user_repo = ReadonlyUserRepo::new(tx.commit(description));
```
yuja, line 1609: Perhaps, we'll also need to disable git exports because the operation is discarded. (There's another `git::export_refs()` caller in `snapshot_working_copy()`.)

It might make sense to split `may_update_working_copy`/`working_copy_shared_with_git` flags into `may_snapshot`/`may_update`/`may_import`/`may_export`, but I'm not sure if that works out.

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj op log` gained an option to include operation diffs.
 
+* A new global flag `--no-commit-transaction` lets you run a command without
```
ilyagr, line 52: (Non-blocking) I'm thinking of alternative names.

`--no-publish-operation` might be a little better (?)

`--transient-operation`? In the same vein, `--disposable-operation` or `--provisional-operation` (but I prefer transient). `--ephemeral-operation` (thanks, thesaurus). 

`--dry-run` would be appropriate for many operations, but not for `fetch`, `push`, `git export`, and such.


-------------

We can absolutely decide or change our mind later, but I like "transient" and "ephemeral", I think.

bryceberger:
This pr doesn't seem to have a link to #2562 -- adding with this message

YPares:
I have a couple questions, based on my interpretation of the intent of this PR:

Besides a global flag for most jj commands, could it be possible to make this behaviour controllable via an env var?
Because in most use cases (again, if I understand correctly the goal here), you are likely to want to bunch together quite a few jj operations in the same "transaction block". Exposing an env var to the whole block would make sure you don't forget to add `--no-commit-transaction` to one operation.

Is the goal to also add a new command like `jj op commit-transaction` to actually commit? Or should we just run the _last_ operation without the `--no-commit-transaction` flag? The latter API seems less convenient to use in practice in scripts, as you'd need to special case the last operation just for the sake of compliance with this API.

yuja:
> Besides a global flag for most jj commands, could it be possible to make this behaviour controllable via an env var?

I don't think env var will help. You'll need to run `jj` with `--at-op <prev-op-id>` to chain unpublished commands anyway. I assume a convenient interface will be added by `jj api` or something.

> Is the goal to also add a new command like `jj op commit-transaction` to actually commit?

That could be.

sbarfurth:
I cloned this PR and rebased it on HEAD in #8882. I addressed some of the comments as well, including renaming the option. I asked @martinvonz directly before doing this.

josephlou5:
> I cloned this PR and rebased it on HEAD in #8882. I addressed some of the comments as well, including renaming the option. I asked @martinvonz directly before doing this.

Since that PR was merged, can this one be closed?

PR #3906, opened 2024-06-17
commit_builder: reset author timestamp on discardable commits

scott2000:
This is my first time contributing, but I thought this might be a useful feature. This PR resolves #2000 by adding a revset to configure which commits have their author timestamps updated when they are rewritten. This gives a lot of flexibility for configuration, but it's more complex than having a single hard-coded rule. Another alternative might be to have a configuration option with a few fixed enum variants instead of a revset. I would appreciate any comments and feedback!

One thing which I was unsure about is whether all of the existing tests should be updated (it caused a large number of changes since many commit IDs are different due to the different timestamps), or if this feature should be disabled by default in tests except for a few new tests created to test this feature.

# Checklist

If applicable:
- [X] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [x] I have added tests to cover my changes

Issue #2000, opened 2023-08-07
FR: Better handling of author dates (as opposed to committer dates)


ilyagr:
This is an attempt to summarize a [Discord discussion](https://discord.com/channels/968932220549103686/968932220549103689/1137951649269284954) started by @chriskrycho.

**Is your feature request related to a problem? Please describe.**

Currently, `jj` sets the author date of a commit on `jj new` and does not change it when the commit is rewritten. This matches the letter git's behavior, but can be surprising.

For example, one might do `jj new main` today, then come three weeks later and start working on something. The author date for that commit will be three weeks before any work was done on it.

**Describe alternatives you've considered**

**Workaround:** @yuja pointed out the workaround: a user can run `jj describe --reset-author --no-edit` at any time to update the author date manually.

We have two proposed changes to this behavior:

1. @yuja suggested to record the author date when the commit first becomes non-"discardable". AFAIK, this means either non-empty, or with children, or with more than one parent.

    This would match the spirit of the current behavior: the author date would better represent when the work on a change was *started*.

     IMO, we could alternatively simply record the moment when the commit became non-empty, which might be a bit simpler and seems to make little difference.

2. @arxanas explained that `git-branchless` updates the author date on every edit to the commit *except* rebasing of the commit.

      This seems very sensible to me: the author date would represent when the work on a commit was *finished*.

3. (Rejected?) In git, people tend to commit when a change first works (e.g. compiles). We could aim for a design that tries to make the author date represent that, but it seems difficult since this is very ambiguous from jj's perspective.

**Describe the solution you'd like**

We have to decide whether we prefer behavior 1, or 2, or an option to switch between them. Having an option is currently my preference.

necauqua:
Isn't 1 and 2 like literally the git author date and committer date?
Except "committer date" does not change on rebases which is actually sensible

Storing both "start of the work" and "end of the work" seems beneficial to me

ilyagr:
I believe Github changes the committer date when a PR gets merged. I'm guessing that's what the committer date is for: when was the change merged to the trunk?

necauqua:
It does it because it doesn't fast-forward :(

arxanas:
> Isn't 1 and 2 like literally the git author date and committer date?
> Except "committer date" does not change on rebases which is actually sensible

The first case seems like an author date, but I wouldn't call the second case a committer date. The committer date is updated on rebases (and the author date isn't), and it's a different thing than the "end of work" date.

[...]

PR #4466, opened 2024-09-14
cli: add `--author` argument for `commit` and `describe`

mati865:
# Checklist

If applicable:

Fixes https://github.com/martinvonz/jj/issues/4170

### DIFF
--- a/cli/src/commands/commit.rs
+++ b/cli/src/commands/commit.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use jj_lib::backend::Signature;
 use jj_lib::object_id::ObjectId;
 use jj_lib::repo::Repo;
 use tracing::instrument;
@@ -22,6 +23,7 @@ use crate::command_error::CommandError;
 use crate::description_util::description_template;
 use crate::description_util::edit_description;
 use crate::description_util::join_message_paragraphs;
+use crate::text_util::parse_author;
 use crate::ui::Ui;
 
 /// Update the description and create a new change on top.
@@ -50,6 +52,16 @@ pub(crate) struct CommitArgs {
     /// $ JJ_USER='Foo Bar' JJ_EMAIL=foo@bar.com jj commit --reset-author
     #[arg(long)]
     reset_author: bool,
+    /// Set author to the provided string
+    ///
+    /// This changes author name and email while retaining author
+    /// timestamp for non-discardable commits.
+    #[arg(
+        long,
+        conflicts_with = "reset_author",
+        value_parser = parse_author
+    )]
+    author: Option<(String, String)>,
 }
 
 #[instrument(skip_all)]
@@ -105,6 +117,14 @@ new working-copy commit.
     commit_builder.set_tree_id(tree_id);
     if args.reset_author {
         commit_builder.set_author(commit_builder.committer().clone());
+    }
+    if let Some((name, email)) = args.author.clone() {
+        let new_author = Signature {
+            name,
+            email,
+            timestamp: commit_builder.author().timestamp.clone(),
+        };
+        commit_builder.set_author(new_author);
     }
 
     let description = if !args.message_paragraphs.is_empty() {
--- a/cli/src/commands/describe.rs
+++ b/cli/src/commands/describe.rs
@@ -17,6 +17,7 @@ use std::io;
 use std::io::Read;
 
 use itertools::Itertools;
+use jj_lib::backend::Signature;
 use jj_lib::commit::CommitIteratorExt;
 use jj_lib::object_id::ObjectId;
 use tracing::instrument;
@@ -30,6 +31,7 @@ use crate::description_util::edit_description;
 use crate::description_util::edit_multiple_descriptions;
 use crate::description_util::join_message_paragraphs;
 use crate::description_util::ParsedBulkEditMessage;
+use crate::text_util::parse_author;
 use crate::ui::Ui;
 
 /// Update the change description or other metadata
@@ -72,6 +74,16 @@ pub(crate) struct DescribeArgs {
     /// $ JJ_USER='Foo Bar' JJ_EMAIL=foo@bar.com jj describe --reset-author
     #[arg(long)]
     reset_author: bool,
+    /// Set author to the provided string
+    ///
+    /// This changes author name and email while retaining author
+    /// timestamp for non-discardable commits.
+    #[arg(
+        long,
+        conflicts_with = "reset_author",
+        value_parser = parse_author
+    )]
+    author: Option<(String, String)>,
 }
 
 #[instrument(skip_all)]
@@ -139,6 +151,14 @@ pub(crate) fn cmd_describe(
                     let new_author = commit_builder.committer().clone();
                     commit_builder.set_author(new_author);
                 }
+                if let Some((name, email)) = args.author.clone() {
+                    let new_author = Signature {
+                        name,
+                        email,
+                        timestamp: commit.author().timestamp.clone(),
+                    };
+                    commit_builder.set_author(new_author);
+                }
                 let temp_commit = commit_builder.write_hidden()?;
                 Ok((commit.id(), temp_commit))
             })
@@ -194,13 +214,14 @@ pub(crate) fn cmd_describe(
     // `transform_descendants` below unnecessarily.
     let commit_descriptions: HashMap<_, _> = commit_descriptions
         .into_iter()
-        .filter_map(|(commit, new_description)| {
-            if *new_description == *commit.description() && !args.reset_author {
-                None
-            } else {
-                Some((commit.id(), new_description))
-            }
+        .filter(|(commit, new_description)| {
+            new_description != commit.description()
+                || args.reset_author
+                || args.author.as_ref().is_some_and(|(name, email)| {
+                    name != &commit.author().name || email != &commit.author().email
+                })
         })
+        .map(|(commit, new_description)| (commit.id(), new_description))
         .collect();
 
     let mut num_described = 0;
@@ -223,6 +244,14 @@ pub(crate) fn cmd_describe(
                 commit_builder = commit_builder.set_description(description);
                 if args.reset_author {
                     let new_author = commit_builder.committer().clone();
+                    commit_builder = commit_builder.set_author(new_author);
+                }
+                if let Some((name, email)) = args.author.clone() {
+                    let new_author = Signature {
+                        name,
+                        email,
+                        timestamp: commit_builder.author().timestamp.clone(),
+                    };
                     commit_builder = commit_builder.set_author(new_author);
                 }
                 num_described += 1;
--- a/cli/src/text_util.rs
+++ b/cli/src/text_util.rs
@@ -261,6 +261,12 @@ pub fn write_wrapped(
     })
 }
 
+pub fn parse_author(author: &str) -> Result<(String, String), &'static str> {
+    let re = regex::Regex::new(r"(?<name>.*?)\s*<(?<email>.+)>$").unwrap();
+    let captures = re.captures(author).ok_or("Invalid author string")?;
+    Ok((captures["name"].to_string(), captures["email"].to_string()))
+}
+
 #[cfg(test)]
 mod tests {
     use std::io::Write as _;
@@ -631,5 +637,34 @@ mod tests {
             format_plain_text(|formatter| write_wrapped(formatter, &recorder, 10)),
             "foo\n",
         );
+    }
+
+    #[test]
+    fn test_parse_author() {
+        let expected_name = "Example";
+        let expected_email = "example@example.com";
+        let parsed = parse_author(&format!("{expected_name} <{expected_email}>")).unwrap();
+        assert_eq!(
+            (expected_name.to_string(), expected_email.to_string()),
+            parsed
+        );
+    }
+
+    #[test]
+    fn test_parse_author_with_utf8() {
+        let expected_name = "Ąćęłńóśżź";
+        let expected_email = "example@example.com";
+        let parsed = parse_author(&format!("{expected_name} <{expected_email}>")).unwrap();
+        assert_eq!(
+            (expected_name.to_string(), expected_email.to_string()),
+            parsed
+        );
+    }
+
+    #[test]
+    fn test_parse_author_without_name() {
+        let expected_email = "example@example.com";
+        let parsed = parse_author(&format!("<{expected_email}>")).unwrap();
+        assert_eq!(("".to_string(), expected_email.to_string()), parsed);
     }
 }
--- a/cli/tests/test_describe_command.rs
+++ b/cli/tests/test_describe_command.rs
@@ -525,6 +525,52 @@ fn test_describe_author() {
     ~
     "###);
 
+    // Change the author for the latest commit (the committer is always reset)
+    test_env.jj_cmd_ok(
+        &repo_path,
+        &[
+            "describe",
+            "--no-edit",
+            "--author",
+            "Super Seeder <super.seeder@example.com>",
+        ],
+    );
+    insta::assert_snapshot!(get_signatures(), @r#"
+    @  Super Seeder super.seeder@example.com 2001-02-03 04:05:12.000 +07:00
+    │  Test User test.user@example.com 2001-02-03 04:05:12.000 +07:00
+    ○  Test User test.user@example.com 2001-02-03 04:05:09.000 +07:00
+    │  Test User test.user@example.com 2001-02-03 04:05:09.000 +07:00
+    ○  Test User test.user@example.com 2001-02-03 04:05:08.000 +07:00
+    │  Test User test.user@example.com 2001-02-03 04:05:08.000 +07:00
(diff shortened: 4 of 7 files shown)


cli/src/commands/commit.rs
```diff
@@ -50,6 +51,11 @@ pub(crate) struct CommitArgs {
     /// $ JJ_USER='Foo Bar' JJ_EMAIL=foo@bar.com jj commit --reset-author
     #[arg(long)]
     reset_author: bool,
+    /// Set author to the provided string
+    ///
+    /// This changes author name and email while retaining timestamp.
+    #[arg(long, conflicts_with = "reset_author")]
```
mati865, line 57: In the issues author said:
> If more than a single --author option is specified, the operation should fail, instructing the user to specify, at most, one.

I'm not sure if this is doable with Clap.

PhilipMetzger: Since the author field is not a `Vec`, you cannot pass it multiple times so it is fine as is.

PhilipMetzger (approved):
LGTM, it's only missing a UI test in `cli/tests/test_describe.rs` for the feature.

cli/src/commands/commit.rs
mati865: I thought Clap would override an argument with the last instance by default, but maybe I have confused it with another crate.

mati865:
Sorry, I've looked for these tests but couldn't find them previously. I'll update them tomorrow.

cli/src/commands/describe.rs
```diff
@@ -218,13 +242,21 @@ pub(crate) fn cmd_describe(
             .collect_vec(),
         |rewriter| {
             let old_commit_id = rewriter.old_commit().id().clone();
+            let new_author = author_arg.clone().map(|(name, email)| Signature {
+                name,
+                email,
+                timestamp: rewriter.old_commit().author().timestamp.clone(),
```
mati865, line 248: https://github.com/martinvonz/jj/blob/56dbbb8fc629a80173b1253e6baec225bc83b8f6/lib/src/commit_builder.rs#L208 modifies the timestamp, so I'm reading it before `rewritter.rebase` is called.

cli/src/commands/describe.rs
```diff
@@ -112,6 +119,12 @@ pub(crate) fn cmd_describe(
         None
     };
 
+    let author_arg = args
+        .author
+        .as_deref()
+        .map(parse_user_name_email)
+        .transpose()?;
+
     let commit_descriptions: Vec<(_, _)> = if args.no_edit || shared_description.is_some() {
```
mati865, line 128: `else` branch is lacking tests for both `--reset-author` and `--author`

cli/src/description_util.rs
```diff
@@ -251,13 +252,22 @@ pub fn description_template(
     Ok(output.into_string_lossy())
 }
 
+pub fn parse_user_name_email(author: &str) -> Result<(String, String), CommandError> {
```
mati865, line 255: I hadn't thought about that earlier, but it could be moved to Clap `https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser`

mati865 (commented):
The tests revealed a few issues that should be addressed now. I'll move value parsing to Clap and rebase the conflict later.

cli/src/commands/describe.rs
yuja: I think the updated timestamp should be preserved in that case. For consistency, `jj describe` should update author timestamp if the commit was empty.

Maybe adjust the doc string instead?

mati865: That would be inconsistent with Git and also lead to strange-looking history that has nonchronological timestamps.

cli/src/description_util.rs
mati865: Done

cli/src/commands/describe.rs
PhilipMetzger: `jj` always had that, see #3906 and it's associated bug (#2000). There's no reason that we have to be consistent with Git here.

mati865: I fully agree with regard to resetting the date of `jj new` commit, not so much about resetting timestamp of other commits but updated PR to that version.
Since this option is now the same as `JJ_USER='Foo Bar' JJ_EMAIL=foo@bar.com jj describe --reset-author` I'm not sure if  there is much benefit from it, shorter and more intuitive syntax seems like the only selling point.

PhilipMetzger: I think it was intended as that, so to my understanding this is also fine.

yuja: > Since this option is now the same as `JJ_USER='Foo Bar' JJ_EMAIL=foo@bar.com jj describe --reset-author`

`--reset-author` always updates author timestamp, but `--author` doesn't, right? I don't think Git consistency is important here, but it's somewhat consistent because in Git, we don't (usually) have a commit without description and content.

mati865: Sorry, my bad.

I had based this entirely on my misunderstanding of the unit test. Commits there have timestamps rewritten because they have empty descriptions, right?

Running on the real repository shows the author timestamp intact and committer timestamp updated as intended:
```
/tmp/author-test
❯ ~/Projects/jj/target/debug/jj log -T "description ++ ' ' ++ author.timestamp() ++ ' ' ++ committer.timestamp()"
@   2024-09-17 01:41:42.000 +02:00 2024-09-17 01:41:42.000 +02:00
○  bar
│   2024-09-17 01:41:42.000 +02:00 2024-09-17 01:41:42.000 +02:00
○  foo
│   2024-09-17 01:41:32.000 +02:00 2024-09-17 01:41:33.000 +02:00
◆   1970-01-01 00:00:00.000 +00:00 1970-01-01 00:00:00.000 +00:00

/tmp/author-test
❯ jj log
@  vmrplqvo mati865@gmail.com 2024-09-17 01:41:42 4da73e4d
○  ztxlzwmq mati865@gmail.com 2024-09-17 01:41:42 2919a8bd
│  bar
○  xonnomqr mati865@gmail.com 2024-09-17 01:41:33 6359b7c7
│  foo
◆  zzzzzzzz root() 00000000

/tmp/author-test
❯ ~/Projects/jj/target/debug/jj desc @-- --no-edit --author "foo bar <foo@example.com>"
Rebased 2 descendant commits
Working copy now at: vmrplqvo 7bacbeed (empty) (no description set)
Parent commit      : ztxlzwmq 9c1a7191 bar

/tmp/author-test
❯ ~/Projects/jj/target/debug/jj log -T "description ++ ' ' ++ author.timestamp() ++ ' ' ++ committer.timestamp()"
@   2024-09-17 01:47:25.000 +02:00 2024-09-17 01:47:25.000 +02:00
○  bar
│   2024-09-17 01:41:42.000 +02:00 2024-09-17 01:47:25.000 +02:00
○  foo
│   2024-09-17 01:41:32.000 +02:00 2024-09-17 01:47:25.000 +02:00
◆   1970-01-01 00:00:00.000 +00:00 1970-01-01 00:00:00.000 +00:00
```

I have updated the docstrings to indicate timestamp preservation if the description is nonempty.

cli/src/cli_util.rs
```diff
@@ -3298,3 +3298,48 @@ fn format_template_aliases_hint(template_aliases: &TemplateAliasesMap) -> String
     );
     hint
 }
+
+pub fn parse_author_arg(author: &str) -> Result<(String, String), clap::Error> {
```
yuja, line 3302: nit: we don't have to use `clap::Error`. `&'static str` or `String` also works.

`text_util` or `description_util` might be better place to host this function. `cli_util` is bloated. We have `split_email()` in `text_util.rs` btw.

cli/src/cli_util.rs
```diff
@@ -3298,3 +3298,48 @@ fn format_template_aliases_hint(template_aliases: &TemplateAliasesMap) -> String
...
+    let re = regex::Regex::new(r"(?<name>.*?)\s*<(?<email>.+)>").unwrap();
```
yuja, line 3303: needs `$` anchor to not discard trailing text?

I think using `.rsplit_once()`, `.strip_suffix()`, `.trim_ascii()`, etc. is also good.

cli/src/commands/commit.rs
```diff
@@ -50,6 +51,16 @@ pub(crate) struct CommitArgs {
     /// $ JJ_USER='Foo Bar' JJ_EMAIL=foo@bar.com jj commit --reset-author
     #[arg(long)]
     reset_author: bool,
+    /// Set author to the provided string
+    ///
+    /// This changes author name and email while retaining author
+    /// timestamp if the description is filled.
```
yuja, line 57: nit: "if the description is filled" isn't accurate. Maybe we can say "discardable" or "empty commit with no description" (copied from 1eebbe57c080568eec96e548c02ab8e43d8b1d8a)

(also applies to `cmd_describe()`)

cli/src/commands/commit.rs
```diff
@@ -106,6 +117,14 @@ new working-copy commit.
     if args.reset_author {
         commit_builder.set_author(commit_builder.committer().clone());
     }
+    if let Some((name, email)) = args.author.clone() {
+        let new_author = jj_lib::backend::Signature {
+            name,
+            email,
+            timestamp: commit_builder.committer().timestamp.clone(),
```
yuja, line 124: you mean `author().timestamp`?

cli/src/commands/commit.rs
```diff
@@ -106,6 +117,14 @@ new working-copy commit.
     if args.reset_author {
         commit_builder.set_author(commit_builder.committer().clone());
     }
+    if let Some((name, email)) = args.author.clone() {
+        let new_author = jj_lib::backend::Signature {
```
yuja, line 121: nit: we usually import `jj_lib::backend::Signature`

cli/src/commands/describe.rs
```diff
@@ -195,7 +215,10 @@ pub(crate) fn cmd_describe(
     let commit_descriptions: HashMap<_, _> = commit_descriptions
         .into_iter()
         .filter_map(|(commit, new_description)| {
-            if *new_description == *commit.description() && !args.reset_author {
+            if *new_description == *commit.description()
+                && !args.reset_author
+                && args.author.is_none()
```
yuja, line 220: nit: strictly speaking, we should test `new_author == commit.author()` because new author might be unchanged (unlike `--reset-author`.) Perhaps, we can refactor `commit_descriptions` to a list of `(old_commit, new_author, new_description)`.

This isn't a blocker. Can be addressed later.

cli/tests/test_describe_command.rs
```diff
@@ -525,29 +525,75 @@ fn test_describe_author() {
...
         &repo_path,
         &[
             "describe",
-            "--config-toml",
-            r#"user.name = "Ove Ridder"
-            user.email = "ove.ridder@example.com""#,
             "--no-edit",
-            "--reset-author",
+            "--author",
+            r#""Super Seeder <super.seeder@example.com>""#,
```
yuja, line 535: Did you intentionally include quotes in user name?

yuja (approved):
LGTM, thanks!

cli/src/cli_util.rs
mati865: Right, I've got so used to `anyhow`/`thiserror` in recent years I tried to create `&'static str` error but with one `Err()` too many over the weekend.
Changed to `&'static str`.

cli/src/cli_util.rs
mati865: `$` makes sense, if there is something after `>` then it's improper syntax.
Added.

cli/src/commands/commit.rs
mati865: I think you meant negation of these statements? Author timestamps of discardable commits have no value.
Changed to `non-discardable commits`.

cli/src/commands/commit.rs
mati865: This is intentional. using author's timestamp gives weird results:
```
-- STDOUT:              jj-cli::runner test_commit_command::test_commit_with_description_template ---

running 1 test
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: commit_with_description_template-2
Source: cli/tests/test_commit_command.rs:244
────────────────────────────────────────────────────────────────────────────────
Expression: std::fs::read_to_string(test_env.env_root().join("editor")).unwrap()
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    0       │-JJ: Author: Another User <another.user@example.com> (2001-02-03 08:05:09)
          0 │+JJ: Author: Another User <another.user@example.com> (2001-02-03 08:05:08)
    1     1 │ JJ: Committer: Test User <test.user@example.com> (2001-02-03 08:05:09)
    2     2 │
    3     3 │ JJ: file2 | 1 +
    4     4 │ JJ: 1 file changed, 1 insertion(+), 0 deletions(-)
────────────┴───────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
test test_commit_command::test_commit_with_description_template ... FAILED

failures:

failures:
    test_commit_command::test_commit_with_description_template

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 598 filtered out; finished in 0.89s
```
This would be inconsistent with other cases in that test.

cli/src/commands/describe.rs
mati865: This is more complicated, we need to test whether the author is none or equal, so it's either:
```rust
                && !args.author.as_ref().is_some_and(|(name, email)| {
                    name != &commit.author().name || email != &commit.author().email
                })
```
or
```rust
                && args.author.is_none()
                && args.author.as_ref().is_some_and(|(name, email)| {
                    name == &commit.author().name && email == &commit.author().email
                })
```

cli/tests/test_describe_command.rs
mati865: 🤦🏻‍♂️ I was fixing a script and forgot to change my mindset that I'm using sane language again.

cli/src/commands/commit.rs
yuja: Well, if you don't pass `--author` here, you'll get the author timestamp of "2001-02-03 08:05:08", right? (Beware that the "file2" exists when the first commit is created (or split by `jj commit file1`.) Since `--author` shouldn't affect how author timestamp is updated, the tested behavior is wrong.

cli/src/text_util.rs
```diff
@@ -261,6 +261,12 @@ pub fn write_wrapped(
     })
 }
 
+pub fn parse_author_arg(author: &str) -> Result<(String, String), &'static str> {
```
yuja, line 264: nit: let's remove `_arg` as it's no longer return `clap::Error`.

cli/src/commands/describe.rs
yuja: Yep, we might want to refactor it later.

You could use `.map_or(true, ..)`, btw, but it might be slightly simpler if we invert the condition.
```rust
        .filter(|(commit, new_description)| {
            new_description != commit.description()
                || args.reset_author
                || args.author.as_ref().is_some_and(|(name, email)| {
                    name != &commit.author().name || email != &commit.author().email
                })
        })
        .map(|(commit, new_description)| (commit.id(), new_description))
```

cli/src/commands/commit.rs
yuja: Looks good, thanks.

cli/src/commands/commit.rs
mati865: Ah, that explains it.
Changed to author timestamp.

mati865:
Thank you for bearing with me through the whole process!
I'll consider how to refactor problematic code in `describe.rs` but probably not until next weekend.

cli/src/commands/describe.rs
yuja: > I'll consider how to refactor problematic code in describe.rs but probably not until next weekend.

Oops, I didn't mean this was a blocker. I think the current state is good, and this PR can be merged if you're okay.

mati865: Oh, sorry. I have badly worded that sentence.
I mean I'll investigate it but likely not before next weekend, and I've enabled auto-merge. The last time I used GH auto merge, the conversations weren't a blocker, so I'll resolve them.

mati865:
I couldn't find the way to refactor it in a way that would improve it, felt like I'm only making things worse.

omentic:
This flag appears to be undocumented.

mati865:
@omentic could you be more verbose about the issue? I think (cannot test ATM) it should show up in the `--help` for the mentioned args.
I'm happy to extend the jj  docs if that's what you mean.

bnjmnt4n:
The feature has been deprecated in favor of `jj metaedit`: https://github.com/jj-vcs/jj/commit/51224ed675dfc6430b35f6cdf0c2fb085b3107be.

omentic:
@mati865 It doesn't show up in `--help`, yeah. Thanks for the info @bnjmnt4n -- I was playing around with flags and noticed this `--author` existed, but I guess I didn't actually *run* it to get the deprecation warning (which appears).

PR #4469, opened 2024-09-15
split: Add optional `--message` arg

Veykril:
I tried using that earlier today as it felt natural (given `describe`) only to realize this not being a thing. Given the implementation is simple I figured to open a PR for the discussion directly. 

Also because I only searched for related issues after, https://github.com/martinvonz/jj/issues/3060, and the discussion in there seems to have derailed somewhat.

# Checklist

If applicable:


yuja:
> The main question is, which change will receive the new description, what happens to the description of the other part and whether this is then intuitive to the majority of users. In my opinion it makes the most sense to adjust the parent change and keep the previous description for the child.

While I agree that `--message` should apply to the first commit if `split` had `--message` option, I don't think it's intuitive. If the goal is to suppress interactive editor, I would add a flag to turn it off.

arxanas:
> In my opinion it makes the most sense to adjust the parent change and keep the previous description for the child.

I didn't follow this. Is it the case that `jj split` creates a child with a new change ID? If so, shouldn't it be the child/new change that gets the provided message?

Veykril:
Hah, I assumed it was splitting out the parent and keeping the current as a child. You are right in that case it would make more sense for it to apply to the new child instead.

Imo having to pass a flag to make it non-interactive feels weird (or well, probably annoying rather) to me I'd expect that to be flipped instead, having to pass a flag/arg to make it interactive then

PR #4470, opened 2024-09-15
refactor: Roll our own hex utils

Veykril:
`hex` does not seem to be maintained any longer and we already transitively depend on `faster_hex` anyways.

# Checklist

If applicable:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,7 @@ gix = { version = "0.66.0", default-features = false, features = [
 ] }
 gix-filter = "0.13.0"
 glob = "0.3.1"
-hex = "0.4.3"
+faster-hex = { version = "0.9.0", default-features = false, features = ["std"]}
 ignore = "0.4.23"
 indexmap = "2.5.0"
 indoc = "2.0.4"
--- a/lib/src/hex_util.rs
+++ b/lib/src/hex_util.rs
@@ -52,7 +52,10 @@ pub fn to_reverse_hex(forward_hex: &str) -> Option<String> {
 }
 
 pub fn decode_hex_string(hex: &str) -> Option<Vec<u8>> {
-    hex::decode(hex.as_bytes()).ok()
+    let mut dst = vec![0; hex.len() / 2];
+    faster_hex::hex_decode(hex.as_bytes(), &mut dst)
+        .ok()
+        .map(|()| dst)
 }
 
 /// Calculates common prefix length of two byte sequences. The length
--- a/lib/src/hex_util.rs
+++ b/lib/src/hex_util.rs
@@ -51,6 +51,10 @@ pub fn to_reverse_hex(forward_hex: &str) -> Option<String> {
         .collect()
 }
 
+pub fn decode_hex_string(hex: &str) -> Option<Vec<u8>> {
+    hex::decode(hex.as_bytes()).ok()
+}
+
 /// Calculates common prefix length of two byte sequences. The length
 /// to be returned is a number of hexadecimal digits.
 pub fn common_hex_len(bytes_a: &[u8], bytes_b: &[u8]) -> usize {
--- a/lib/src/hex_util.rs
+++ b/lib/src/hex_util.rs
@@ -14,27 +14,23 @@
 
 #![allow(missing_docs)]
 
-pub fn to_reverse_hex_digit(b: u8) -> u8 {
-    let offset = match b {
-        b'0'..=b'9' => b - b'0',
-        b'A'..=b'F' => b - b'A' + 10,
-        b'a'..=b'f' => b - b'a' + 10,
-        b => return b,
-    };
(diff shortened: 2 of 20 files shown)


lib/src/hex_util.rs
```diff
@@ -14,7 +14,17 @@
 
 #![allow(missing_docs)]
 
-fn to_reverse_hex_digit(b: u8) -> Option<u8> {
+pub fn to_reverse_hex_digit(b: u8) -> u8 {
+    let offset = match b {
+        b'0'..=b'9' => b - b'0',
+        b'A'..=b'F' => b - b'A' + 10,
+        b'a'..=b'f' => b - b'a' + 10,
+        b => return b,
```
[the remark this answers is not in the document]
emilazy: I’m guessing this can result in more efficient code (in terms of code size and runtime performance), since it means the function never panics.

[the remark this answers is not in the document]
samueltardieu: > I’m guessing this can result in more efficient code (in terms of code size and runtime performance), since it means the function never panics.

When the input is valid (the fast path) the code stays the same. Using `unreachable!()` here would prevent risking returning an invalid result in case the input is invalid.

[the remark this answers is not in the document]
emilazy: Right. I just mean that having to handle the potential of a panic in a function call can potentially result in worse code generation. I don’t know if it’s meaningful in this case, though.

lib/src/hex_util.rs
```diff
@@ -51,6 +51,13 @@ pub fn to_reverse_hex(forward_hex: &str) -> Option<String> {
         .collect()
 }
 
+pub fn decode_hex_string(hex: &str) -> Option<Vec<u8>> {
```
yuja, line 64: Can you reorder (or squash) the first two patches? It's easier to follow if we had a drop-in replacement for `hex::decode()`.

lib/src/hex_util.rs
```diff
@@ -51,6 +51,13 @@ pub fn to_reverse_hex(forward_hex: &str) -> Option<String> {
...
+    let mut dst = vec![0; hex.len() / 2];
+    faster_hex::hex_decode(hex.as_bytes(), &mut dst)
+        .ok()
+        .map(|()| dst)
```
yuja, line 68: nit: (at your option) I think `hex_decode(..).ok()?; Some(dst)` is simpler.

Cargo.toml
```diff
@@ -66,7 +66,7 @@ gix = { version = "0.66.0", default-features = false, features = [
 ] }
 gix-filter = "0.13.0"
 glob = "0.3.1"
-hex = "0.4.3"
+faster-hex = { version = "0.9.0", default-features = false, features = ["std"]}
```
yuja, line 69: nit: apparently, 0.10.0 is released yesterday

and the dependency list is sorted alphabetically.

lib/src/hex_util.rs
```diff
@@ -14,7 +14,17 @@
 
 #![allow(missing_docs)]
 
-fn to_reverse_hex_digit(b: u8) -> Option<u8> {
+pub fn to_reverse_hex_digit(b: u8) -> u8 {
```
yuja, line 17: nit: might be simpler to do hex encoding by ourselves. It won't gain SIMD optimization, but the performance wouldn't matter.

lib/src/hex_util.rs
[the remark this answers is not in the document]
Veykril: Yes, having a panicking path here results in unwinding codegen that propagates through callers. Its not really relevant for us here.

Cargo.toml
Veykril: I didnt wanna depend on that due to us already depending on 0.9.0 transitively and those dependencies aren't bumpable

lib/src/hex_util.rs
Veykril: As in roll all the hex stuff ourselves and put it into this module? Fine by me

Veykril:
Why is CI running nightly rustfmt 😕

samueltardieu:
> Why is CI running nightly rustfmt 😕

The current state of the repository seems compatible with `cargo fmt` from the stable, beta and nightly toolchains.

lib/src/hex_util.rs
yuja: Oh, I just meant `to_reverse_hex(&str) -> Option<String>` could be reimplemented as `encode_reverse_hex(&[u8]) -> String`.

It might be okay to self-host all hex encoding functions if that helps reduce dependency, though.

Veykril:
Ah wait, this is rustfmt working differently depending on how its invoked again I think (bare `fmt` invocation vs stdin). Nevermind then.

Veykril:
Closing this as I don't really have the drive for this refactoring right now

PR #4471, opened 2024-09-15
cli: add helper to set up op template environment for the current repo state

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -156,6 +156,7 @@ use crate::git_util::print_git_import_stats;
 use crate::merge_tools::DiffEditor;
 use crate::merge_tools::MergeEditor;
 use crate::merge_tools::MergeToolConfigError;
+use crate::operation_templater::OperationTemplateLanguage;
 use crate::operation_templater::OperationTemplateLanguageExtension;
 use crate::revset_util;
 use crate::revset_util::RevsetExpressionEvaluator;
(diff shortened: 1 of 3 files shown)


cli/src/cli_util.rs
```diff
@@ -1293,6 +1294,15 @@ impl WorkspaceCommandHelper {
             .commit_template_language(self.repo().as_ref(), self.id_prefix_context())
     }
 
+    /// Creates operation template language environment for this workspace.
+    pub fn operation_template_language(&self) -> OperationTemplateLanguage {
```
Veykril, line 1298: This won't be usable for the the diff command for https://github.com/martinvonz/jj/pull/4445 as that (if I understand things correctly) can in theory deal with multiple repositories that differ from the one in the `WorkspaceCommandHelper` no?

Veykril (approved):

cli/src/cli_util.rs
yuja: `cmd_op_diff()` loads multiple repos in order to show refs at these operations. OTOH, "From operation"/"To operation"  are headers displayed at a higher level. If `op_sumamry` template contained `if(current_operation, "(current)")`, `jj op diff --from @- --to @` should show "From operation .., To operation .. (current)".

Veykril: Yup, your comment on #4445 clarified things so this should be fine. Feel free to merge then I'll rebase mine on this

PR #4472, opened 2024-09-15
squash: preserve the working copy position in the tree

### LINKED ISSUE #4468: FR: when abandoning the working copy, keep the children in the new change

After creating a fix-up commit in the middle of a tree and working on it, I squashed my changes into the parent commit because some more tests passed, but I still intended to continue working on the fix. However, the new commit created after `jj squash` didn't have the same children as I expected: I was now branching out of the tree instead of working in the middle of it.

I would have liked my new commit to have the same children (same child in my case) as the one who just got emptied, so that I could continue working "at the same place" in the tree.

I could have used `jj squash --keep-emptied`, but I like the idea of abandoning a change when emptying it via squashing and starting a new one.

**Additional context**

The newly created change after abandoning the working copy keeps the same parents as the previous working copy. Also preserving the children would be more consistent.

**Backward compatiblity concerns**

The change would change the behavior of jj only in one particular scenario:

- the working copy is not "in the middle" of the tree (it has at least one child)
- it is abandoned in place, for example via squashing its content in another change
- the user continues editing the new working copy

I think it is unlikely that a user would want to:

- develop a change purposely _inserted_ in the middle of the tree (for example using `jj new --after <CHANGE ID>`, opposed to `jj new <CHANGE ID>` to start a new head from this change)
- squash the change into its parent after working on it
- then continue developing as a new head from this newly squashed change

samueltardieu:
When a squashed working copy commit is emptied and abandoned, the newly created working copy will keep the same parents and children as the previous one.

Note that this does not change the result of any existing test.

# Checklist

If applicable:

Fix #4468


martinvonz:
I think we would first need to address @yuja's [comment on the tracking issue](https://github.com/martinvonz/jj/issues/4468#issuecomment-2351261460) about abandoning the working copy when you move off of it.

samueltardieu:
> I think we would first need to address @yuja's [comment on the tracking issue](https://github.com/martinvonz/jj/issues/4468#issuecomment-2351261460) about abandoning the working copy when you move off of it.

Agreed, which is why I put it in draft mode right now. I've yet to come with an acceptable solution.

PR #4475, opened 2024-09-15
cli: Color more change and commit ids in cli outputs

Veykril:
# Checklist

If applicable:


cli/src/commands/git/push.rs
```diff
@@ -396,46 +406,33 @@ fn print_commits_ready_to_push(
...
+        }
+        if let Some(old_target) = &update.old_target {
+            write!(formatter, " from ")?;
+            format_commit_shortest_prefix(formatter, repo, id_prefix_context, old_target)?;
+        }
+        if let Some(new_target) = &update.new_target {
+            write!(formatter, " to ")?;
+            format_commit_shortest_prefix(formatter, repo, id_prefix_context, new_target)?;
+        }
+        writeln!(formatter)?;
```
yuja, line 435: Perhaps, this function (and `print_git_import_stats()`) can be extracted to templates similar to `bookmark_list`.

Each `(String, BookmarkPushUpdate)` could be encoded as `commit_templater::RefName`, but it's probably better to add `RefDiff` or `RefChange` template type. In `print_git_import_stats()`, we'll need to represent a diff of remote-ref to remote-ref.

cli/src/commands/git/push.rs
```diff
@@ -266,8 +268,15 @@ pub fn cmd_git_push(
 
     validate_commits_ready_to_push(&bookmark_updates, &remote, &tx, command, args)?;
     if let Some(mut formatter) = ui.status_formatter() {
-        writeln!(formatter, "Changes to push to {remote}:")?;
-        print_commits_ready_to_push(formatter.as_mut(), repo.as_ref(), &bookmark_updates)?;
+        write!(formatter, "Changes to push to ")?;
+        write!(formatter.labeled("remote_branches"), "{remote}")?;
```
PhilipMetzger, line 272: Nit: Use the new name `remote_bookmarks`.

cli/src/commands/git/push.rs
Veykril: Yes they ought to be all templated I noticed that after I was done with this current version (hence the draft).

PR #4476, opened 2024-09-15
lib: optimize common prefix computation of two hex strings

samueltardieu:
Comparing each byte before comparing the nibbles is more efficient. A benchmark comparing the old and new implementations with various common prefix lengths shows:

```
Common hex len/old/3    time:   [7.5444 ns 7.5807 ns 7.6140 ns]
Common hex len/new/3    time:   [1.2100 ns 1.2144 ns 1.2192 ns]
Common hex len/old/6    time:   [11.849 ns 11.879 ns 11.910 ns]
Common hex len/new/6    time:   [1.9950 ns 2.0046 ns 2.0156 ns]
Common hex len/old/32   time:   [63.030 ns 63.345 ns 63.718 ns]
Common hex len/new/32   time:   [6.4647 ns 6.4800 ns 6.4999 ns]
```

[Benchmark source](https://git.sr.ht/~rfc1149/common-hex-len-benchmark/tree/main/item/benches/hex.rs) (I know that internal hex strings are not ASCII but a byte is a byte)

# Checklist

If applicable:

### DIFF
--- a/lib/src/hex_util.rs
+++ b/lib/src/hex_util.rs
@@ -51,24 +51,23 @@ pub fn to_reverse_hex(forward_hex: &str) -> Option<String> {
         .collect()
 }
 
-/// Calculates common prefix length of two bytes. The length to be returned is
-/// a number of hexadecimal digits.
+/// Calculates common prefix length of two byte sequences. The length
+/// to be returned is a number of hexadecimal digits.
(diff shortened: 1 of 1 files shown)


martinvonz (approved):
Nice! I'm not sure it matters in practice, but maybe? I don't think this makes the code harder to read anyway.

PR #4478, opened 2024-09-15
Add config for ignoring executable bit changes

### LINKED ISSUE #3949: Annoying executable reporting on NTFS drives in Linux
I have a shared NTFS drive between Linux and Windows. In Linux I mount the NTFS drive with the 'exec' option (as default) so I can compile and run binaries. However 'exec' makes *all* files within the drive executable.

When I try to colocate a git repo on this drive, jj reports every file as being modified to be executable. While they all *are* executable, I don't want this represented as a change in the repo. Running `jj chmod normal <some_file>` does nothing because the drive is *mounted* with all files executable, and jj continues to report everything as modified to be executable.

Git has an option for this as `core.fileMode`.

Here is git's documentation:
https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefileMode

`core.fileMode`
> Tells Git if the executable bit of files in the working tree is to be honored.
>
> Some filesystems lose the executable bit when a file that is marked as executable is checked out, or checks out a non-executable file with executable bit on. [git-clone[1]](https://git-scm.com/docs/git-clone) or [git-init[1]](https://git-scm.com/docs/git-init) probe the filesystem to see if it handles the executable bit correctly and this variable is automatically set as necessary.
>
> A repository, however, may be on a filesystem that handles the filemode correctly, and this variable is set to `true` when created, but later may be made accessible from another environment that loses the filemode (e.g. exporting ext4 via CIFS mount, visiting a Cygwin created repository with Git for Windows or Eclipse). In such a case it may be necessary to set this variable to `false`. See [git-update-index[1]](https://git-scm.com/docs/git-update-index).
>
> The default is true (when core.filemode is not specified in the config file).

I'm not sure how an option like this might integrate with JJ's use of executable bits in windows, but it would be nice to add some solution for filesystems on Linux that can't easily control the executable bit.

wrzian:
Starting a PR to close #3949 . This is my first time contributing, so I would appreciate feedback on the code. I'm certain that there are some places that can be integrated better. I've left a few TODO markers scattered around that I would appreciate help with.

~~Also: **I have not run these changes locally yet.** They build and pass `cargo test`, but I'm still not certain how to run these myself.~~

~~I also haven't added any tests yet. I'd appreciate feedback on a direction to go for those and where it would be best to test these changes.~~

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -237,6 +237,11 @@ Thanks to the people who made this release happen!
 * `jj resolve` now accepts new built-in merge tools `:ours` and `:theirs`.
   These merge tools accept side #1 and side #2 of the conflict respectively.
 
+* Updated how we handle executable bits in the working copy to allow Unix to
+  ignore executable bit changes when using a nonstandard filesystem. We try to
+  detect the filesystem's behavior on Unix by default, but it can be overridden
+  manually by setting `core.executable-bit = "ignore" | "respect"`.
+
 ### Fixed bugs
 
 * `jj log -p --stat` now shows diff stats as well as the default color-words/git
--- a/docs/config.md
+++ b/docs/config.md
@@ -1536,6 +1536,37 @@ snapshots on filesystem changes by setting
 You can check whether Watchman is enabled and whether it is installed correctly
 using `jj debug watchman status`.
 
+## Respect or ignore executable bit permissions
+
+Whether to respect or ignore changes to the executable bit for files on Unix.
+This option is unused on Windows.
+
+`core.executable-bit = "respect" | "ignore"`
+
+On Unix systems, files have a permission bit for whether they are executable as
+scripts or binary code. Jujutsu stores this state in each commit and will update
+it for files as you operate on a repository. If you set your working copy to a
+commit where a file is recorded as executable or not, `jj` will adjust the
+permission of that file. If you change a file's executable bit through the
+filesystem, `jj` will record that change when taking a snapshot.
+
+Setting this to `"ignore"` will make Jujutsu ignore the executable bit on the
+filesystem and only store executable bit state internally. In addition, `jj` will
+not attempt to modify a file's executable bit and will add new files as "not
+executable." This is already the behavior on Windows, and having the option to
+enable this behavior is useful for Unix users dual-booting Windows, Windows
+users accessing files from WSL, or anyone experimenting with filesystem
+configurations.
+
+On Unix, Jujutsu will try to detect whether the filesystem supports executable
+bits to choose a default value (it should be `"respect"` on most systems), but a
+configured value will always override the default. On Windows there is no
+executable bit, and this option is not used.
+
+You can always use `jj file chmod` to update the recorded executable bit for a
+file in the commit manually. If this option is `"respect"`, `jj` will attempt to
+propagate that change to the filesystem.
+
 ## Snapshot settings
 
 ### Paths to automatically track
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -2090,6 +2090,8 @@ impl LocalWorkingCopy {
         })
     }
 
+    /// "Load" the working copy lazily to avoid work. This uses an empty
+    /// `OnceCell` to load the states in the future as needed.
     pub fn load(
         store: Arc<Store>,
         working_copy_path: PathBuf,
@@ -2129,8 +2131,9 @@ impl LocalWorkingCopy {
             })
     }
 
-    fn tree_state_mut(&mut self) -> Result<&mut TreeState, WorkingCopyStateError> {
-        self.tree_state()?; // ensure loaded
+    fn tree_state_mut(&mut self) -> Result<&mut TreeState, (String, Box<dyn Error + Send + Sync>)> {
+        self.tree_state() // Ensure loaded.
+            .map_err(|WorkingCopyStateError { message, err }| (message, err))?;
         Ok(self.tree_state.get_mut().unwrap())
     }
 
@@ -2235,10 +2238,7 @@ impl LockedWorkingCopy for LockedLocalWorkingCopy {
         let tree_state = self
             .wc
             .tree_state_mut()
-            .map_err(|err| SnapshotError::Other {
-                message: "Failed to read the working copy state".to_string(),
-                err: err.into(),
-            })?;
+            .map_err(|(message, err)| SnapshotError::Other { message, err })?;
         let (is_dirty, stats) = tree_state.snapshot(options)?;
         self.tree_state_dirty |= is_dirty;
         Ok((tree_state.current_tree_id().clone(), stats))
@@ -2255,10 +2255,7 @@ impl LockedWorkingCopy for LockedLocalWorkingCopy {
         let tree_state = self
             .wc
             .tree_state_mut()
-            .map_err(|err| CheckoutError::Other {
-                message: "Failed to load the working copy state".to_string(),
-                err: err.into(),
-            })?;
+            .map_err(|(message, err)| CheckoutError::Other { message, err })?;
         if tree_state.tree_id != *commit.tree_id() {
             let stats = tree_state.check_out(&new_tree, options)?;
             self.tree_state_dirty = true;
@@ -2276,10 +2273,7 @@ impl LockedWorkingCopy for LockedLocalWorkingCopy {
         let new_tree = commit.tree()?;
         self.wc
             .tree_state_mut()
-            .map_err(|err| ResetError::Other {
-                message: "Failed to read the working copy state".to_string(),
-                err: err.into(),
-            })?
+            .map_err(|(message, err)| ResetError::Other { message, err })?
             .reset(&new_tree)
             .block_on()?;
         self.tree_state_dirty = true;
@@ -2290,10 +2284,7 @@ impl LockedWorkingCopy for LockedLocalWorkingCopy {
         let new_tree = commit.tree()?;
         self.wc
             .tree_state_mut()
-            .map_err(|err| ResetError::Other {
-                message: "Failed to read the working copy state".to_string(),
-                err: err.into(),
-            })?
+            .map_err(|(message, err)| ResetError::Other { message, err })?
             .recover(&new_tree)
             .block_on()?;
         self.tree_state_dirty = true;
@@ -2314,10 +2305,7 @@ impl LockedWorkingCopy for LockedLocalWorkingCopy {
         let stats = self
             .wc
             .tree_state_mut()
-            .map_err(|err| CheckoutError::Other {
-                message: "Failed to load the working copy state".to_string(),
-                err: err.into(),
-            })?
+            .map_err(|(message, err)| CheckoutError::Other { message, err })?
             .set_sparse_patterns(new_sparse_patterns, options)?;
         self.tree_state_dirty = true;
         Ok(stats)
@@ -2331,7 +2319,8 @@ impl LockedWorkingCopy for LockedLocalWorkingCopy {
         assert!(self.tree_state_dirty || &self.old_tree_id == self.wc.tree_id()?);
         if self.tree_state_dirty {
             self.wc
-                .tree_state_mut()?
+                .tree_state_mut()
+                .map_err(|(message, err)| WorkingCopyStateError { message, err })?
                 .save()
                 .map_err(|err| WorkingCopyStateError {
                     message: "Failed to write working copy state".to_string(),
@@ -2359,10 +2348,7 @@ impl LockedLocalWorkingCopy {
     pub fn reset_watchman(&mut self) -> Result<(), SnapshotError> {
         self.wc
             .tree_state_mut()
-            .map_err(|err| SnapshotError::Other {
-                message: "Failed to read the working copy state".to_string(),
-                err: err.into(),
-            })?
+            .map_err(|(message, err)| SnapshotError::Other { message, err })?
             .reset_watchman();
         self.tree_state_dirty = true;
         Ok(())
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -1011,6 +1011,8 @@ pub struct WcTreeMutator<'a> {
 pub struct WcTreeConfig {
     pub conflict_marker_style: ConflictMarkerStyle,
     pub exec_config: ExecConfig,
+    pub fsmonitor_settings: FsmonitorSettings,
+    pub max_new_file_size: u64,
 }
 
 impl WcTreeMutator<'_> {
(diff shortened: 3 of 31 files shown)


martinvonz (commented):
Not a full review, just a few random comments

martinvonz:
Thanks, btw! It looks good overall

lib/src/settings.rs
```diff
@@ -258,6 +258,25 @@ impl UserSettings {
...
+///
+/// On unix setting this to `false` will disable executable bit reporting and
+/// setting it to `true` will enable it regardless of our filesystem check. This
+/// can be useful for users dual-booting windows or experimenting with
+/// nonstandard filesystems.
+///
+/// This is not a method on UserSettings since it needs to be callable even when
+/// only having the config itself (currently only used in `Ui::with_config`).
+pub fn report_executable_bit(config: &config::Config) -> Option<bool> {
+    config.get_bool("core.report-executable-bit").ok()
```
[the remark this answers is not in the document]
wrzian: Really apreciate this insight, will update. Ignore is a much better term. I struggled with the name a bunch when writing this.

lib/src/local_working_copy.rs
```diff
@@ -104,19 +107,130 @@ use crate::working_copy::WorkingCopy;
 use crate::working_copy::WorkingCopyFactory;
 use crate::working_copy::WorkingCopyStateError;
 
+/// None means we are ignoring the executable bit and don't care what it is.
```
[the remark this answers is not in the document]
wrzian: Yep. Fixing

lib/src/settings.rs
```diff
@@ -258,6 +258,25 @@ impl UserSettings {
...
+/// update files if they are manually changed via `jj file chmod`.
+///
+/// On unix setting this to `false` will disable executable bit reporting and
+/// setting it to `true` will enable it regardless of our filesystem check. This
+/// can be useful for users dual-booting windows or experimenting with
+/// nonstandard filesystems.
+///
+/// This is not a method on UserSettings since it needs to be callable even when
+/// only having the config itself (currently only used in `Ui::with_config`).
+pub fn report_executable_bit(config: &config::Config) -> Option<bool> {
```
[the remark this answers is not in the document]
wrzian: Nope :upside_down_face: It's also used in `cli/src/ui.rs`. See the added `exec_config` arguments in the `merge_tools/{mod,diff_working_copies}.rs` files that originate from the ui.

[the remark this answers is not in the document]
wrzian: Though I'm still not sure where to put this. If we choose `local_working_copy.rs` does that have any effect on other users of the library? I'm fine moving it there if you recommend

lib/src/local_working_copy.rs
```diff
@@ -104,19 +107,130 @@ use crate::working_copy::WorkingCopy;
...
+/// On Windows there is no executable bit, so this will always be None. On Unix
+/// it will usually be Some(_), but may be disabled by a config value for users
+/// with non-standard filesystems or who use a nonstandard filesystem on unix.
+///
+/// This represents *whether* a file is executable, but not whether that state
+/// is to be reported.
+#[derive(Debug, Clone, Copy)]
+pub enum ExecFlag {
+    Exec(bool),
+    DontCare,
```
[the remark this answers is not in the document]
wrzian: Going with `Ignore`.

lib/src/settings.rs
```diff
@@ -258,23 +258,23 @@ impl UserSettings {
...
-/// On unix setting this to `false` will disable executable bit reporting and
-/// setting it to `true` will enable it regardless of our filesystem check. This
-/// can be useful for users dual-booting windows or experimenting with
+/// On unix setting this to `true` will disable automatic executable bit updates
+/// and setting it to `false` will enable it regardless of the filesystem check.
+/// This can be useful for users dual-booting windows or experimenting with
 /// nonstandard filesystems.
 ///
 /// This is not a method on UserSettings since it needs to be callable even when
 /// only having the config itself (currently only used in `Ui::with_config`).
```
martinvonz, line 275: There's a lot of duplication between this text and `config.md`. I think we can keep this description short. Developers should know where they can go to read about the config option.

wrzian: Can do. I will make it reference the config documentation explicitly though.

docs/config.md
```diff
@@ -951,24 +951,25 @@ snapshots on filestem changes by setting
...
-there is no executable bit and this config is completely ignored.
+Whether to ignore changes to the executable bit for files on unix. On windows
+there is no executable bit and this config is unused.
 
 On unix we determine the default value by querying the filesystem as to whether
-executable bits are supported. If they are, we will report changes to files'
-executable bit and update the repository. Otherwise we will only update files if
-they are manually changed via `jj file chmod`.
+executable bits are supported. If they are, we will not ignore them and changing
+a files' executable bit will update the repository. If they are not supported we
```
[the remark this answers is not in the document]
wrzian: I think the short description at the top and the explicit line with the available options is really helpful to let a user who is just skimming the docs get what they need and be able to choose whether to read the rest or not.
I made your intro paragraph a bit more explicit about how the working-copy commit reads from the filesystem.

lib/src/settings.rs
PhilipMetzger, line 1: Nit: Squash this into the first commit.

CHANGELOG.md
PhilipMetzger, line 1: Nit: this should probably have the topic of `lib: Capitalize "Unix" and "Windows"`.

PhilipMetzger, line 1: Nit: Squash this into the first commit.

PhilipMetzger, line 1: Nit: The first commit  probably should have the title of `loca_working_copy:` since it contains a majority of the changes.

PhilipMetzger (commented):
Just some nits to conform to our code review style here: https://github.com/martinvonz/jj/blob/main/docs/contributing.md#code-reviews

lib/src/local_working_copy.rs
```diff
@@ -571,14 +683,21 @@ impl TreeState {
...
         wc.save()?;
         Ok(wc)
     }
 
-    fn empty(store: Arc<Store>, working_copy_path: PathBuf, state_path: PathBuf) -> TreeState {
+    fn empty(
+        store: Arc<Store>,
+        working_copy_path: PathBuf,
+        state_path: PathBuf,
+        exec_config: Option<bool>,
```
yuja, line 697: Perhaps, `exec_config` can be carried by `SnapshotOptions`? When materializing working-copy files, it's probably okay to try `chmod()` even if the filesystem doesn't support exec bit.

It's a bit odd that `TreeState::empty()` constructor needs it.

lib/src/local_working_copy.rs
```diff
@@ -1644,6 +1777,7 @@ impl LocalWorkingCopy {
         state_path: PathBuf,
         operation_id: OperationId,
         workspace_id: WorkspaceId,
+        config: &config::Config,
```
yuja, line 1780: nit: we typically use `&UserSeettings` object.

CHANGELOG.md
wrzian: This didn't change anything but this PR so I'm going to just merge it back into the main commit.

docs/config.md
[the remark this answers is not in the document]
wrzian: I feel like ignore works fine, it's far enough from any gitignore files that I don't really care. However I'm fine with either "ignore" or "respect" as the main words.
Should I change it again or not?

lib/src/local_working_copy.rs
wrzian: I can see how `SnapshotOptions` might work, but I'm not sure that it would preserve the correct behavior when doing a checkout/reset/recover and then calling `LockedWorkingCopy::finish()`, since those save to the repo. If we're ok with sending the permission changes regardless of if they might work then I could see it working, but it feels somewhat 'less correct'? Not sure.

lib/src/local_working_copy.rs
wrzian: I'm only using config because the `Ui` object only gets initialized with a  config and it felt odd to pass user_settings into everything and then only use it by calling: `ignore_exec_bit(user_settings.config())`. I'm ok with changing it but I don't really see a reason to pass the full user settings?

lib/src/local_working_copy.rs
yuja: (assuming `fchmod()` doesn't fail,) checkout is repo->working-copy operation, so it won't have to read exec bit from disk. `reset()` and `recover()` are similar?

lib/src/local_working_copy.rs
yuja: > I don't really see a reason to pass the full user settings?

I don't have a concrete plan, but we'll probably replace `config::Config` with our config module that supports layering better. Then, `config::Config` APIs might be rehosted on `UserSettings`.

lib/src/local_working_copy.rs
wrzian: Yes, but then its the call to `finish()` afterwards that tries to write to the repo. And `finish` doesn't currently take any other configuration. I think it makes sense to keep the config with the tree state.

yuja: Just to be celar, `LockedLocalWorkingCopy::finish()` doesn't write commits to repo, but writes working-copy/tree states to disk. It's more or less a serialization of in-memory data.

That said, it might be okay to keep a copy of filesystem capability options in `TreeState`. We already have `symlink_support: bool` (and `store: Arc<Store>` which isn't state data but a shared resource.) If we make `symlink_support` configurable, we'll have to resolve it somewhere in `LocalWorkingCopy`, and pass in to `TreeState` constructor (or methods.)

wrzian:
Pushed some small fixes in the local_working_copy changes, also added a new commit that refactors the check_out functions in `diff_working_copy.rs`. Let me know if this needs any work.

I've also started using the update on my machine, and it seems to work correctly on my dual-boot repo with the config, but I haven't really tested all of the behavior.
(Still need to add actual tests for the new behavior)

One other note is that I don't think my current changes work correctly if you update the config value. That is, if we alter the config value to stop respecting executable bits, the tree should refresh itself, but I think it still keeps the stored changes instead of refreshing even if I do `jj status`. Not sure what to do for that.

lib/src/local_working_copy.rs
```diff
@@ -104,19 +107,117 @@ use crate::working_copy::WorkingCopy;
...
+/// The executable bit for a filetype, potentially ignored.
+///
+/// On Windows there is no executable bit, so this will always be Ignore. On
+/// Unix it will usually be Exec(true|false), but may be ignored by a config
+/// value or by a check we run when we load the config if not specified.
+#[derive(Debug, Clone, Copy)]
+pub enum ExecFlag {
+    Exec(bool),
+    Ignore,
+}
```
martinvonz, line 119: Is it possible to always use a bool instead of this type and only change how we used that bool based on the `IgnoreExec`?

lib/src/local_working_copy.rs
```diff
@@ -104,19 +107,117 @@ use crate::working_copy::WorkingCopy;
...
+/// is ignored if we find that the filesystem doesn't support it or by user
+/// configuration.
 #[cfg(unix)]
-type FileExecutableFlag = bool;
+#[derive(Debug, Clone, Copy)]
+struct IgnoreExec(bool);
 #[cfg(windows)]
-type FileExecutableFlag = ();
+#[derive(Debug, Clone, Copy)]
+struct IgnoreExec();
```
martinvonz, line 151: nit: you can drop the parentheses, but it's also fine if you prefer to keep them

lib/src/local_working_copy.rs
wrzian: Ah, that makes more sense for`finish`.

How would you feel about some sort of `FileCapability` or `Platform` trait/module? It seems like it effectively already exists for symlink and executable bits in `file_util.rs`, and it would be nice to add helpers like a `from_config()` function.

lib/src/local_working_copy.rs
wrzian: Ok, then I will move to passing UserSettings

lib/src/local_working_copy.rs
wrzian: We could, the main difference is just whether the tree state stores that a file is not executable at a given moment. I assume this would affect users who modify the configuration between snapshots, but I don't really know what that would imply.

I can convert it to just a bool if you'd like.

wrzian: Alternatively, it might be better to use a bool-equivalent enum, e.g. `enum { Exec, NotExec }`.

martinvonz: > We could, the main difference is just whether the tree state stores that a file is not executable at a given moment. I assume this would affect users who modify the configuration between snapshots, but I don't really know what that would imply.

Good point. I think that means it's better to do our best to keep the stored flag up to date even the configuration currently says to ignore the executable bit in the file system. Otherwise we might clear the flag in the file and not notice if the user then sets the config to respect the config and also has cleared the executable bit on some file in the file system.

> Alternatively, it might be better to use a bool-equivalent enum, e.g. `enum { Exec, NotExec }`.

I would go with just a `bool` until we find that it's hard to read or likely to cause bugs for some reason (e.g. because we confuse that boolean for some other boolean value).

martinvonz:
> What are the current blockers that are precluding a merge?

I think it's just the unresolved comments. Plus rebasing it, of course. Since we haven't heard from @wrzian here in a while, I think you should feel free to take it over (and I assume Ian will let us know if he wants to get back to it soon instead).

wrzian:
I have some time to work on this now, so I'll try to get this finished in the next few weeks. Marking as draft while I finish cleaning up the rebase from main and loading context into my head.

docs/config.md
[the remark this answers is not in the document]
wrzian: In the new updates I'm using an enum with `Ignore` and `Respect` variants corresponding to string keys for the config.

wrzian:
Alright, I've got this mostly put together now. I've added some basic tests in a new file, and I've massively reorganized the commit structure: things are broken out much more atomically. LMK what things can be improved :)

This is now ready for review!

wrzian:
Resolved a bunch of personal nits when reviewing my own code :grin:

lib/src/local_working_copy.rs
```diff
@@ -588,7 +588,7 @@ fn remove_old_file(disk_path: &Path) -> Result<bool, CheckoutError> {
     match fs::remove_file(disk_path) {
         Ok(()) => Ok(true),
         Err(err) if err.kind() == io::ErrorKind::NotFound => Ok(false),
-        // TODO: Use io::ErrorKind::IsADirectory if it gets stabilized
+        Err(err) if err.kind() == io::ErrorKind::IsADirectory => Ok(false),
         Err(_) if disk_path.symlink_metadata().is_ok_and(|m| m.is_dir()) => Ok(false),
```
yuja, line 646: nit: the `m.is_dir()` line should be removed

yuja (commented):
(Just quickly scanned random changes. I haven't reviewed the commits thoroughly.)

Since this PR is quite large, and GitHub doesn't allow partial acceptance, I suggest splitting this into three or four PRs. For example, pure refactoring like the first two patches can be easily approved. Things like the fourth patch "diff_working_copies: Update `DiffSide` enum and simplify tree checkout functions" include non-trivial changes, but they look pretty independent and well documented, so can be reviewed separately.

Thanks.

lib/src/local_working_copy.rs
wrzian: Doing so causes test failures on Mac and Windows. Let's discuss more in PR 1:  #6442

wrzian:
I'm closing this because I'm moving to a new account, @isuffix. I'll restart this PR soon.

PR #4480, opened 2024-09-16
docs: redirect `branches` to `bookmarks`, fixup URL

ilyagr:


samueltardieu (changes requested):
I think this needs to be repeated in `mkdocs-offline.yml` as well, as plugin config is not inherited.

ilyagr:
Good point, thank you! Fixed.

samueltardieu (approved):

PR #4483, opened 2024-09-16
bookmark: add "b" alias

samueltardieu:
`jj bookmark` is a frequently used command. Its subcommands already have one letter aliases. Defining `jj b` as an alias for `jj bookmarks` make bookmarks really easy to use.

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_alias.rs
+++ b/cli/tests/test_alias.rs
@@ -24,9 +24,9 @@ fn test_alias_basic() {
     test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
     let repo_path = test_env.env_root().join("repo");
 
-    test_env.add_config(r#"aliases.b = ["log", "-r", "@", "-T", "bookmarks"]"#);
+    test_env.add_config(r#"aliases.bk = ["log", "-r", "@", "-T", "bookmarks"]"#);
     test_env.jj_cmd_ok(&repo_path, &["bookmark", "create", "my-bookmark"]);
-    let stdout = test_env.jj_cmd_success(&repo_path, &["b"]);
+    let stdout = test_env.jj_cmd_success(&repo_path, &["bk"]);
     insta::assert_snapshot!(stdout, @r###"
     @  my-bookmark
     │
@@ -41,9 +41,9 @@ fn test_alias_legacy_section() {
     let repo_path = test_env.env_root().join("repo");
 
     // Can define aliases in [alias] section
-    test_env.add_config(r#"alias.b = ["log", "-r", "@", "-T", "bookmarks"]"#);
+    test_env.add_config(r#"alias.bk = ["log", "-r", "@", "-T", "bookmarks"]"#);
     test_env.jj_cmd_ok(&repo_path, &["bookmark", "create", "my-bookmark"]);
-    let stdout = test_env.jj_cmd_success(&repo_path, &["b"]);
+    let stdout = test_env.jj_cmd_success(&repo_path, &["bk"]);
     insta::assert_snapshot!(stdout, @r###"
     @  my-bookmark
     │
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -21,7 +21,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj obslog` is now called `jj evolution-log`/`jj evolog`. `jj obslog` remains
   as an alias.
 
-* `jj branch` has been deprecated in favor of `jj bookmark`.
+* `jj branch` has been deprecated in favor of `jj bookmark`. `jj b` is available
+  as an alias.
```
PhilipMetzger, line 25: Nit: This needs no CHANGELOG entry.

docs/bookmarks.md
```diff
@@ -218,4 +218,11 @@ To resolve a conflicted state in a remote bookmark (e.g. `main@origin`), simply
 pull from the remote (e.g. `jj git fetch`). The conflict resolution will also
 propagate to the local bookmark (which was presumably also conflicted).
 
+## Ease of use
+
+The use of bookmarks is so frequent in some workflows that one-letter
+shortcuts have been implemented, both for the `jj bookmark` command
+itself (as `jj b`) and for its subcommands. For example, `jj bookmark
+create BRANCH-NAME` can be abbreviated as `jj b c BRANCH-NAME`.
+
```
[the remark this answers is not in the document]
samueltardieu: Even though the whole file needs a rework, notably to include a clear explanation of how jj bookmarks are not similar to git branches or git tags, I have underline that bookmarks are commonly used to interact with git.

yuja:
Can't we instead add `aliases.b` so the user can override it to whatever they want?

FWIW, I personally don't like one-character command aliases such as `jj b c`.

docs/bookmarks.md
```diff
@@ -218,4 +218,13 @@ To resolve a conflicted state in a remote bookmark (e.g. `main@origin`), simply
 pull from the remote (e.g. `jj git fetch`). The conflict resolution will also
 propagate to the local bookmark (which was presumably also conflicted).
 
+## Ease of use
+
+The use of bookmarks is frequent in some workflows, for example, when
+interacting with git repositories containing branches. To this end,
```
martinvonz, line 224: nit: I think we try to write "Git" when talking about the tool (and `git` when talking about the command).

docs/bookmarks.md
```diff
@@ -218,4 +218,13 @@ To resolve a conflicted state in a remote bookmark (e.g. `main@origin`), simply
...
+one-letter shortcuts have been implemented, both for the `jj bookmark`
+command itself (as `jj b`) and for its subcommands. For example, `jj
+bookmark create BRANCH-NAME` can be abbreviated as `jj b c
+BRANCH-NAME`.
```
martinvonz, line 228: Should be `BOOKMARK-NAME` here :)

samueltardieu: Ooops!

samueltardieu:
> Can't we instead add `aliases.b` so the user can override it to whatever they want?

I hadn't realized we could do that. It is a better idea indeed. Done.

docs/bookmarks.md
samueltardieu: Fixed

martinvonz (approved):

cli/tests/test_alias.rs
```diff
@@ -41,20 +41,20 @@ fn test_alias_legacy_section() {
...
     "###);
 
     // The same alias (name) in both [alias] and [aliases] sections is an error
-    test_env.add_config(r#"aliases.b = ["bookmark", "list"]"#);
-    let stderr = test_env.jj_cmd_failure(&repo_path, &["b"]);
+    test_env.add_config(r#"aliases.bk = ["bookmark", "list"]"#);
+    let stderr = test_env.jj_cmd_failure(&repo_path, &["bk"]);
     insta::assert_snapshot!(stderr, @r###"
-    Error: Alias "b" is defined in both [aliases] and [alias]
+    Error: Alias "bk" is defined in both [aliases] and [alias]
```
martinvonz, line 57: Btw, these tests probably didn't have to change now that the alias is defined in the config file, but it still seems reasonable to use a different name here.

PhilipMetzger (approved):
LG

PR #4485, opened 2024-09-16
squash doc: --keep-emptied let the description unchanged

### LINKED ISSUE #4473: `jj squash --keep-emptied` does not copy the description of the source commit
## Description

`jj squash` help page says that the description of the source commit, if any, will be copied or merged into the target commit description. This is not the case when using `--keep-emptied`.

```
$ jj new
$ jj new -m "Some description"
$ echo foo > bar
$ jj squash --keep-emptied
$ jj log -r @-::
@  rtnywryn sam@rfc1149.net 2024-09-15 14:36:12 84bf696b
│  (empty) Some description
○  nsknwpvm sam@rfc1149.net 2024-09-15 14:36:12 5c65418c
│  (no description set)
```

## Expected Behavior

The description should have been moved to the target commit.

## Actual Behavior

No description copying or merging takes place.

## Specifications

- Platform: Linux
- Version: main

samueltardieu:
Fix #4473

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/squash.rs
+++ b/cli/src/commands/squash.rs
@@ -48,7 +48,7 @@ use crate::ui::Ui;
 /// parent(s), and `--keep-emptied` is not set, it will be abandoned. Without
 /// `--interactive` or paths, the source revision will always be empty.
 ///
-/// If the source became empty and both the source and destination had a
+/// If the source was abandoned and both the source and destination had a
 /// non-empty description, you will be asked for the combined description. If
 /// either was empty, then the other one will be used.
 ///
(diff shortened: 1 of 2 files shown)


samueltardieu:
Technically, each "change" would keep its own description, not each "commit". However, using "changes" instead of "commits" here would be confusing (even if technically more accurate) because the text talks about the "changes" (as in "diffs") present in the initial commits.

matts1 (approved):

cli/src/commands/squash.rs
```diff
@@ -50,7 +50,8 @@ use crate::ui::Ui;
 ///
 /// If the source became empty and both the source and destination had a
 /// non-empty description, you will be asked for the combined description. If
-/// either was empty, then the other one will be used.
+/// either was empty, then the other one will be used. As an exception, when
+/// `--keep-emptied` is used, each commit keeps its own description.
```
martinvonz, line 54: An alternative solution is to update the paragraph with "If the source became empty" -> "If the source was abandoned" (or perhaps "If the source was abandoned (because it became empty and `--keep-emptied` was not requested)"). Thoughts?

matts1: I like that phrasing, I think it's much clearer since there's no special-casing.

samueltardieu: Agreed.

PR #4479, opened 2024-09-15
unsquash: deprecated in favor of squash

samueltardieu:
Everything can be done with `jj squash` that could be done with `jj unsquash`.

# Checklist

If applicable:
- [X] I have updated `CHANGELOG.md`
- [X] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #4489, opened 2024-09-17
diffedit: add --restore-descendants flag

samueltardieu:
This adds to `diffedit` something that `unsquash` could do and `squash` could not. This would ease the deprecation of `unsquash` (see #4479).

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/diffedit.rs
+++ b/cli/src/commands/diffedit.rs
@@ -38,8 +38,8 @@ use crate::ui::Ui;
 ///
 /// Edit the right side of the diff until it looks the way you want. Once you
 /// close the editor, the revision specified with `-r` or `--to` will be
-/// updated. Descendants will be rebased on top as usual, which may result in
-/// conflicts.
+/// updated. Unless `--restore-descendants` is used, descendants will be
+/// rebased on top as usual, which may result in conflicts.
 ///
 /// See `jj restore` if you want to move entire files from one revision to
 /// another. See `jj squash -i` or `jj unsquash -i` if you instead want to move
@@ -64,6 +64,14 @@ pub(crate) struct DiffeditArgs {
     /// Specify diff editor to be used
     #[arg(long, value_name = "NAME")]
     tool: Option<String>,
+    /// Preserve the content (not the diff) when rebasing descendants
+    ///
+    /// When rebasing a descendant on top of the rewritten revision, its diff
+    /// compared to its parent(s) is normally preserved, i.e. the same way that
+    /// descendants are always rebased. This flag makes it so the content/state
+    /// is preserved instead of preserving the diff.
+    #[arg(long)]
+    restore_descendants: bool,
 }
 
 #[instrument(skip_all)]
@@ -119,13 +127,23 @@ don't make any changes, then the operation will be aborted.",
             .write()?;
         // rebase_descendants early; otherwise `new_commit` would always have
         // a conflicted change id at this point.
-        let num_rebased = tx.repo_mut().rebase_descendants(command.settings())?;
+        let (num_rebased, extra_msg) = if args.restore_descendants {
+            (
+                tx.repo_mut().reparent_descendants(command.settings())?,
+                " (while preserving their content)",
+            )
+        } else {
+            (tx.repo_mut().rebase_descendants(command.settings())?, "")
+        };
         if let Some(mut formatter) = ui.status_formatter() {
             write!(formatter, "Created ")?;
             tx.write_commit_summary(formatter.as_mut(), &new_commit)?;
             writeln!(formatter)?;
             if num_rebased > 0 {
-                writeln!(formatter, "Rebased {num_rebased} descendant commits")?;
+                writeln!(
+                    formatter,
+                    "Rebased {num_rebased} descendant commits{extra_msg}"
+                )?;
             }
         }
         tx.finish(ui, format!("edit commit {}", target_commit.id().hex()))?;
--- a/cli/tests/test_diffedit_command.rs
+++ b/cli/tests/test_diffedit_command.rs
@@ -514,3 +514,42 @@ fn test_diffedit_old_restore_interactive_tests() {
     +unrelated
     "###);
 }
+
+#[test]
+fn test_diffedit_restore_descendants() {
+    let mut test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    std::fs::write(repo_path.join("file"), "println!(\"foo\")\n").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["new"]);
+    std::fs::write(repo_path.join("file"), "println!(\"bar\")\n").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["new"]);
+    std::fs::write(repo_path.join("file"), "println!(\"baz\");\n").unwrap();
+
+    let edit_script = test_env.set_up_fake_diff_editor();
+
+    // Add a ";" after the line with "bar". There should be no conflict.
+    std::fs::write(edit_script, "write file\nprintln!(\"bar\");\n").unwrap();
+    let (stdout, stderr) = test_env.jj_cmd_ok(
+        &repo_path,
+        &["diffedit", "-r", "@-", "--restore-descendants"],
+    );
+    insta::assert_snapshot!(stdout, @"");
+    insta::assert_snapshot!(stderr, @r#"
+    Created rlvkpnrz 62b8c2ce (no description set)
+    Rebased 1 descendant commits (while preserving their content)
+    Working copy now at: kkmpptxz 321d1cd1 (no description set)
+    Parent commit      : rlvkpnrz 62b8c2ce (no description set)
+    "#);
+    let stdout = test_env.jj_cmd_success(&repo_path, &["diff", "--git"]);
+    insta::assert_snapshot!(stdout, @r#"
+    diff --git a/file b/file
+    index 1a598a8fc9..7b6a85ab5a 100644
+    --- a/file
+    +++ b/file
+    @@ -1,1 +1,1 @@
+    -println!("bar");
+    +println!("baz");
+    "#);
+}
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1314,6 +1314,11 @@ impl MutableRepo {
         result
     }
 
+    /// Rebase descendants of the rewritten commits.
+    ///
+    /// The descendants of the commits registered in `self.parent_mappings` will
+    /// be recursively rebased onto the new version of their parents.
+    /// Returns the number of rebased descendants.
     pub fn rebase_descendants(&mut self, settings: &UserSettings) -> BackendResult<usize> {
         let roots = self.parent_mapping.keys().cloned().collect_vec();
         let mut num_rebased = 0;
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1334,6 +1334,27 @@ impl MutableRepo {
         Ok(num_rebased)
     }
 
+    /// Reparent descendants of the rewritten commits.
+    ///
+    /// The descendants of the commits registered in `self.parent_mappings` will
+    /// be recursively reparented onto the new version of their parents.
+    /// The content of those descendants will remain untouched.
+    /// Returns the number of reparented descendants.
+    pub fn reparent_descendants(&mut self, settings: &UserSettings) -> BackendResult<usize> {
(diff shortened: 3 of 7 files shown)


martinvonz:
See #1027 for some discussion about how we may want to make this more generic. I don't think we should have to duplicate implementation in each command where it's useful.

samueltardieu:
> I don't think we should have to duplicate implementation in each command where it's useful.

I have moved the reparenting code into `jj_lib::repo`, and simplified it a lot, thanks to `Repo::transform_descendants()`. Now, the part that is specific to `diffedit` is reduced to calling either `rebase_children` or `reparent_children`.

cli/src/commands/diffedit.rs
```diff
@@ -59,6 +59,9 @@ pub(crate) struct DiffeditArgs {
     /// Specify diff editor to be used
     #[arg(long, value_name = "NAME")]
     tool: Option<String>,
+    /// Do not modify the content of the revision specified with `-r` or `--to`.
```
PhilipMetzger, line 62: Nit: No trailing period in doc strings.

samueltardieu: Thanks. I've also cleaned up the rest of the help strings for `diffedit` in #4497.

cli/src/commands/diffedit.rs
```diff
@@ -59,6 +59,9 @@ pub(crate) struct DiffeditArgs {
     /// Specify diff editor to be used
     #[arg(long, value_name = "NAME")]
     tool: Option<String>,
+    /// Do not modify the content of the revision specified with `-r` or `--to`
+    #[arg(long)]
+    preserve_descendants: bool,
```
yuja, line 64: nit: `--preserve-descendants` sounds like the default behavior is to somehow destroy descendants. `--reparent-descendants` or `--preserve-descendant-snapshots` ("snapshots" compared to "diffs") might be slightly better.

cli/src/commands/diffedit.rs
```diff
@@ -125,6 +132,6 @@ don't make any changes, then the operation will be aborted.",
             }
```
yuja, line 132: nit: "Reparented .."?

cli/src/commands/diffedit.rs
```diff
@@ -125,6 +132,6 @@ don't make any changes, then the operation will be aborted.",
...
         }
         tx.finish(ui, format!("edit commit {}", target_commit.id().hex()))?;
-    }
+    };
```
yuja, line 135: nit: redundant `;`

cli/src/commands/diffedit.rs
samueltardieu: I feel like `--preserve-descendant-snapshots` is too long to be usable and too technical. Even `content` would be clearer than `snapshots` for the end-user I think.

cli/src/commands/diffedit.rs
samueltardieu: I'm not sure I understand what you're suggesting here, the suggestion seems to be on the "the operation will be aborted" line.

cli/src/commands/diffedit.rs
samueltardieu: Ooops. Wasn't caught by clippy, I don't even see an applicable lint, I'll investigate.

cli/src/commands/diffedit.rs
yuja:
```rust
if args.preserve_descendants {
    writeln!(formatter, "Reparented ...")
} else {
    writeln!(formatter, "Rebased ...")
}
```
(no idea why I can't add comment to unchanged line on GitHub.)

samueltardieu: Let me try this. I'm not sure this level of details is warranted, but it might be. I'll add this as a separate commit to test the idea for the time being.

samueltardieu: "reorganize"?

samueltardieu: Martin has suggested "rewrite" on Discord, I'll give it a try.

cli/src/commands/diffedit.rs
martinvonz: I agree that `--preserve-descendant-snapshots` is too technical (not sure I think it's too long because users will probably set up shell completion). I also agree with Yuya's concern. I'm not sure what's a better name. Another way of viewing it is that we apply the reverse change in the descendants, but I suspect it would be even more confusing to explain it that way (`--undo/backout-in-descendants`). Maybe `--restore-descendants` would work? The idea is that it matches the `jj restore` semantics.

samueltardieu: Another point I had read was the `--restore-descendants` could not be later abbreviated as `-r` or `-R` since this is used already, but nothing prevents us from using another letter, that is not a rare practice.

I'll update the PR with `--restore-descendants` and see how it feels.

lib/src/repo.rs
```diff
@@ -1329,6 +1329,21 @@ impl MutableRepo {
         Ok(num_rebased)
     }
 
+    pub fn reparent_descendants(&mut self, settings: &UserSettings) -> BackendResult<usize> {
```
martinvonz, line 1332: Can you add a doc comment?

Sorry that I forgot to ask before, and sorry that `rebase_descendants()` (which was probably added by me) doesn't have it. I hope to some day find time to go through and add missing doc comments so we can remove `#![allow(missing_docs)]` everywhere.

cli/src/commands/diffedit.rs
```diff
@@ -64,6 +64,9 @@ pub(crate) struct DiffeditArgs {
     /// Specify diff editor to be used
     #[arg(long, value_name = "NAME")]
     tool: Option<String>,
+    /// Do not modify the content of the revision specified with `-r` or `--to`
```
martinvonz, line 67: Let's try to clarify this. I'm not sure most readers will understand what it means. (It also seems incorrect in that it says that the target commit itself is unchanged). Maybe something like this:

> Preserve the content (not diff) when rebasing descendants
> 
> When rebasing a descendant on top of the rewritten revision, its diff compared to its parent(s) is normally
> preserved, i.e. the same way that descendants are always rebased. This flag makes it so the content/state
> is preserved instead of preserving the diff.

Do you think that's clear?

cli/src/commands/diffedit.rs
```diff
@@ -119,13 +122,17 @@ don't make any changes, then the operation will be aborted.",
...
+            tx.repo_mut().rebase_descendants(command.settings())?
+        };
         if let Some(mut formatter) = ui.status_formatter() {
             write!(formatter, "Created ")?;
             tx.write_commit_summary(formatter.as_mut(), &new_commit)?;
             writeln!(formatter)?;
-            if num_rebased > 0 {
-                writeln!(formatter, "Rebased {num_rebased} descendant commits")?;
+            if num_rewritten > 0 {
+                writeln!(formatter, "Rewritten {num_rewritten} descendant commits")?;
```
martinvonz, line 135: nit: Rewritten -> Rewrote

Also - and I know we talked about this on Discord, but I didn't think about it until now - I'm worried that users who don't pass `--restore-descendants` will be confused that this command says "Rewrote 5 descendant commits" when all other commands say "Rebased 5 descendant commits". I'm leaning towards  using "Rebased" for both cases or maybe "Rebased 5 descendant commits" and "Rebased 5 descendant commits (while preserving their content)" or something like that.

cli/tests/test_diffedit_command.rs
```diff
@@ -514,3 +516,119 @@ fn test_diffedit_old_restore_interactive_tests() {
...
+    --- a/file
+    +++ b/file
+    @@ -1,1 +1,1 @@
+    -println!("bar");
+    +println!("baz");
+    "#);
+}
+
+#[test]
+fn test_diffedit_restore_descendants_with_multiple_parents() {
```
martinvonz, line 560: This test should ideally go into the lib crate, testing the new function in the first commit in this PR. It makes sense to also have a test of the CLI feature, but we should not have to test more than just the basics here. The lib tests are faster, and they mean that we test the reusable functionality, so we don't have to test it for every command that gains a `--restore-descendants` flag. Sorry that I didn't notice this earlier.

lib/src/repo.rs
samueltardieu: I've added an initial commit with documentation for `rebase_descendants()`, and added the documentation for `reparent_descendants()`.

cli/src/commands/diffedit.rs
samueltardieu: The "(not diff)" looked a bit strange to me at first glance, but with the full explanation it makes sense.

cli/src/commands/diffedit.rs
samueltardieu: I've added your proposed phrasing. If we add this option to more commands, it will make sense to also factor out the message printing in order not to repeat the same text (and code) everywhere.

cli/tests/test_diffedit_command.rs
samueltardieu: I won't be able to do it right away, but I'll keep this on my list.

cli/src/commands/diffedit.rs
martinvonz: not sure if it actually helps, but "(not the diff)" sounds better upon rereading it

samueltardieu: Yes a bit indeed. Updated.

cli/tests/test_diffedit_command.rs
martinvonz: No problem. If you think you'll get to it soon, please update the PR. If it might take a bit longer (than a few days, says), feel free to send another PR instead.

martinvonz (approved):
LGTM, but let's give Yuya a chance to comment on the option name (and the code too).

cli/tests/test_diffedit_command.rs
samueltardieu: I've added a new test in its own commit

lib/tests/test_mut_repo.rs
```diff
@@ -676,3 +676,75 @@ fn test_remove_wc_commit_previous_discardable() {
     mut_repo.rebase_descendants(&settings).unwrap();
     assert!(!mut_repo.view().heads().contains(old_wc_commit.id()));
 }
+
+#[test]
+fn test_reparent_descendants() {
```
martinvonz, line 681: nit: squash the test into the parent commit (see https://martinvonz.github.io/jj/latest/contributing/#code-reviews)

cli/tests/test_diffedit_command.rs
martinvonz: Thanks! I think you can delete this test now, unless you think it covers some case that the jj-lib test and the test above don't cover.

martinvonz (approved):

cli/src/commands/diffedit.rs
samueltardieu: @martinvonz What about using "Rebased 5 descendant commits (no content change)" instead of "Rebased 5 descendant commits (while preserving their content)"?

martinvonz: I think that sounds a bit more like something that happened to be the case given some circumstance than something that's guaranteed to be the case. Perhaps the user will think that it was something about those commits that made it so there was no content change. Does that make sense?

samueltardieu: It does! Leaving as-is.

lib/tests/test_mut_repo.rs
```diff
@@ -676,3 +676,76 @@ fn test_remove_wc_commit_previous_discardable() {
...
+    }
+    let repo = tx.commit("test");
+
+    // Rewrite "commit_a" by removing its content.
+    let mut tx = repo.start_transaction(&settings);
+    let mut_repo = tx.repo_mut();
+    let empty_commit_tree_id = mut_repo.store().empty_merged_tree_id().clone();
+    mut_repo
+        .rewrite_commit(&settings, &commit_a)
+        .set_tree_id(empty_commit_tree_id)
```
yuja, line 714: Maybe better to use another random tree (or file content updates) here.

If a file deleted from both merge sides, the result is deletion. This test passes even if we did `rebase_descendants()`.

yuja (approved):

lib/src/repo.rs
```diff
@@ -1314,6 +1314,11 @@ impl MutableRepo {
         result
     }
 
+    /// Reparent descendants of the rewritten commits.
```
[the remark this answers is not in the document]
samueltardieu: Ooops, cut-n-paste went wrong. Fixed, thanks.

lib/tests/test_mut_repo.rs
samueltardieu: Good catch. Fixed.

PR #4492, opened 2024-09-17
simplify-parents: add a command to remove redundant parents

torquestomp:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Added diff options to ignore whitespace when comparing lines. Whitespace
   changes are still highlighted.
 
+* New command `jj simplify-parents` will remove redundant parent edges.
+
 ### Fixed bugs
 
 
@@ -106,6 +108,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   content.
 
 * `jj git fetch -b <remote-git-branch-name>` will now warn if the branch(es)
+
+ ### Fixed bugs
+ 
+  * Update working copy before reporting changes. This prevents errors during reporting
+    from leaving the working copy in a stale state.
+ 
+  * `jj git fetch -b <remote-git-branch-name>` will now warn if the branch(es)
    can not be found in any of the specified/configured remotes.
 
 * `jj split` now lets the user select all changes in interactive mode. This may be used
--- a/cli/src/commands/simplify_parents.rs
+++ b/cli/src/commands/simplify_parents.rs
@@ -1,0 +1,119 @@
+use std::collections::HashSet;
+
+use itertools::Itertools;
+use jj_lib::backend::BackendResult;
+use jj_lib::revset::RevsetExpression;
+use jj_lib::rewrite::CommitRewriter;
+use jj_lib::settings::UserSettings;
+
+use crate::cli_util::CommandHelper;
+use crate::cli_util::RevisionArg;
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Simplify parent edges for the specified revision(s).
+///
+/// Removes all parents of each of the specified revisions that are also
+/// indirect ancestors of the same revisions through other parents. This has no
+/// effect on any revision's contents, including the working copy.
+///
+/// In other words, for all (A, B, C) where A has (B, C) as parents and C is an
+/// ancestor of B, A will be rewritten to have only B as a parent instead of
+/// B+C.
+#[derive(clap::Args, Clone, Debug)]
+#[command(group = clap::ArgGroup::new("revision-args").multiple(true).required(true))]
+pub(crate) struct SimplifyParentsArgs {
+    /// Simplify specified revision(s) together with their trees of descendants
+    /// (can be repeated)
+    #[arg(long, short, group = "revision-args")]
+    source: Vec<RevisionArg>,
+    /// Simplify specified revision(s) (can be repeated)
+    #[arg(long, short, group = "revision-args")]
+    revisions: Vec<RevisionArg>,
+}
+
+pub(crate) fn cmd_simplify_parents(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &SimplifyParentsArgs,
+) -> Result<(), CommandError> {
+    let mut workspace_command = command.workspace_helper(ui)?;
+    let revs = RevsetExpression::descendants(
+        workspace_command
+            .parse_union_revsets(ui, &args.source)?
+            .expression(),
+    )
+    .union(
+        workspace_command
+            .parse_union_revsets(ui, &args.revisions)?
+            .expression(),
+    );
+    let commit_ids = workspace_command
+        .attach_revset_evaluator(revs)
+        .evaluate_to_commit_ids()?
+        .collect_vec();
+    workspace_command.check_rewritable(&commit_ids)?;
+    let commit_ids_set: HashSet<_> = commit_ids.iter().cloned().collect();
+    let num_orig_commits = commit_ids.len();
+
+    let mut tx = workspace_command.start_transaction();
+    let mut stats = SimplifyStats::default();
+    tx.repo_mut()
+        .transform_descendants(command.settings(), commit_ids, |rewriter| {
+            if commit_ids_set.contains(rewriter.old_commit().id()) {
+                simplify_commit_parents(command.settings(), rewriter, &mut stats)?;
+            }
+
+            Ok(())
+        })?;
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        if !stats.is_empty() {
+            writeln!(
+                formatter,
+                "Removed {} edges from {} out of {} commits.",
+                stats.edges, stats.commits, num_orig_commits
+            )?;
+        }
+    }
+    tx.finish(ui, format!("simplify {num_orig_commits} commits"))?;
+
+    Ok(())
+}
+
+#[derive(Default)]
+struct SimplifyStats {
+    commits: usize,
+    edges: usize,
+}
+
+impl SimplifyStats {
+    fn is_empty(&self) -> bool {
+        self.commits == 0 && self.edges == 0
+    }
+}
+
+fn simplify_commit_parents(
+    settings: &UserSettings,
+    mut rewriter: CommitRewriter,
+    stats: &mut SimplifyStats,
+) -> BackendResult<()> {
+    if rewriter.old_commit().parent_ids().len() <= 1 {
+        return Ok(());
+    }
+
+    let num_old_heads = rewriter.new_parents().len();
+    rewriter.simplify_ancestor_merge();
+    let num_new_heads = rewriter.new_parents().len();
+
+    if rewriter.parents_changed() {
+        rewriter.reparent(settings)?.write()?;
+
+        if num_new_heads < num_old_heads {
+            stats.commits += 1;
+            stats.edges += num_old_heads - num_new_heads;
+        }
+    }
+
+    Ok(())
+}
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -75,6 +75,7 @@ This document contains the help content for the `jj` command-line program.
 * [`jj restore`↴](#jj-restore)
 * [`jj root`↴](#jj-root)
 * [`jj show`↴](#jj-show)
+* [`jj simplify-parents`↴](#jj-simplify-parents)
 * [`jj sparse`↴](#jj-sparse)
 * [`jj sparse edit`↴](#jj-sparse-edit)
 * [`jj sparse list`↴](#jj-sparse-list)
@@ -138,6 +139,7 @@ To get started, see the tutorial at https://martinvonz.github.io/jj/latest/tutor
 * `restore` — Restore paths from another revision
 * `root` — Show the current workspace root directory
 * `show` — Show commit description and changes in a revision
+* `simplify-parents` — Simplify parent edges for the specified revision(s)
 * `sparse` — Manage which paths from the working-copy commit are present in the working copy
 * `split` — Split a revision in two
 * `squash` — Move changes from a revision into another revision
@@ -1841,6 +1843,23 @@ Show commit description and changes in a revision
 * `--context <CONTEXT>` — Number of lines of context to show
 * `--ignore-all-space` — Ignore whitespace when comparing lines
 * `--ignore-space-change` — Ignore changes in amount of whitespace when comparing lines
+
+
+
+## `jj simplify-parents`
+
+Simplify parent edges for the specified revision(s).
+
+Removes all parents of each of the specified revisions that are also indirect ancestors of the same revisions through other parents. This has no effect on any revision's contents, including the working copy.
+
+In other words, for all (A, B, C) where A has (B, C) as parents and C is an ancestor of B, A will be rewritten to have only B as a parent instead of B+C.
+
+**Usage:** `jj simplify-parents <--source <SOURCE>|--revisions <REVISIONS>>`
+
+###### **Options:**
+
+* `-s`, `--source <SOURCE>` — Simplify specified revision(s) together with their trees of descendants (can be repeated)
+* `-r`, `--revisions <REVISIONS>` — Simplify specified revision(s) (can be repeated)
 
 
 
--- a/cli/tests/test_simplify_parents_command.rs
+++ b/cli/tests/test_simplify_parents_command.rs
@@ -1,0 +1,180 @@
+// Copyright 2024 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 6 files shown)


cli/src/commands/simplify.rs
```diff
@@ -0,0 +1,133 @@
...
+        .into_iter()
+        .collect();
+    if new_heads == old_heads {
+        return Ok(SimplifyStats::default());
+    }
+    let removed = old_heads.len() - new_heads.len();
+
+    tx.repo_mut()
+        .rewrite_commit(settings, commit)
+        .set_parents(new_heads.into_iter().collect())
```
samueltardieu, line 129: I think you might reorder the untouched parents in an arbitrary order. Order should probably be preserved.

cli/tests/cli-reference@.md.snap
```diff
@@ -1811,6 +1813,23 @@ Show commit description and changes in a revision
 
 
 
+## `jj simplify`
+
+Simplify parent edges for the specified revision(s).
+
+Removes any and all redundant parent edges going into any of the specified revisions. By definition, this has no effect on any commit contents and should not cause any working copy changes or conflicts.
+
+Immutable commits are automatically filtered from the specified inputs.
```
samueltardieu, line 1822: Shouldn't `--ignore-immutable` be honored instead?

cli/tests/cli-reference@.md.snap
```diff
@@ -1811,6 +1813,23 @@ Show commit description and changes in a revision
 
 
 
+## `jj simplify`
+
+Simplify parent edges for the specified revision(s).
+
+Removes any and all redundant parent edges going into any of the specified revisions. By definition, this has no effect on any commit contents and should not cause any working copy changes or conflicts.
```
samueltardieu, line 1820: nit: "all" covers "any and all"

samueltardieu, line 1820: It was not clear to me what a redundant parent edge was before reading the code and playing with the code. Maybe a textual example would be good ("For example, a commit with multiple parents descending from one another will only depend on its closest ancestors." or something like that, even though it is clear what "closest ancestors" mean here).

torquestomp: Added a paragraph here.

samueltardieu:
I don't want to approve this change myself because I've not followed closely enough the discussions to know the policy about adding new commands. In particular, is "simplify" clear enough, or should it be "simplify-tree", or whatever?

But I would have had a use for it as recently as today, when one of my PR (change A) was merged into `main` today: one of my development subtree was based both on this change A and another unmerged one (change B). After rebasing B onto main, I ended up for my dev subtree parents with a merge between a descendant of main (B) and what was now a parent of main (A). A `jj simplify` here would have saved me the manual rebasing of my subtree on B only.

In this case it was only one subtree with only two parents, and a simple operation, but with a more complex development tree I can see how it would have been even more useful.

cli/src/commands/simplify.rs
```diff
@@ -0,0 +1,139 @@
...
+        .evaluate_to_commits()?
+        .try_collect()?;
+    workspace_command.check_rewritable(commits.iter().map(|c| c.id()))?;
+    let orig_commits = commits.len();
+
+    let mut tx = workspace_command.start_transaction();
+    let mut stats = SimplifyStats::default();
+    for commit in commits {
+        stats.add(simplify_commit(command.settings(), &mut tx, &commit)?);
+    }
```
yuja, line 67: Perhaps, we can use `repo.transform_descendants()` to rewrite commits in topological order. It will save interim predecessor commits.

Maybe we can also build some cache for fast heads computation, but I'm not sure if doing that makes sense. If the history is mostly linear, `index.heads()` might be fast enough.

cli/src/commands/simplify.rs
```diff
@@ -0,0 +1,139 @@
...
+    if let Some(mut formatter) = ui.status_formatter() {
+        if !stats.is_empty() {
+            writeln!(
+                formatter,
+                "Removed {} edges from {} out of {} commits.",
+                stats.edges, stats.commits, orig_commits
+            )?;
+        }
+    }
+    tx.finish(ui, format!("simplified {} commits", orig_commits))?;
```
yuja, line 78: nit: "simplify .."

yuja (commented):
> In particular, is "simplify" clear enough, or should it be "simplify-tree", or whatever?

I don't have a good suggestion, but I feel "simplify" is too generic. If we use "simplify-<something>", "simplify-parents" is probably better.

torquestomp:
> > In particular, is "simplify" clear enough, or should it be "simplify-tree", or whatever?
> 
> I don't have a good suggestion, but I feel "simplify" is too generic. If we use "simplify-", "simplify-parents" is probably better.

Done, changed to simplify-parents for now.

Idk if the file should still be named 'simplify.rs' to leave room for others, but I changed it to simplify_parents.rs also for now

cli/src/commands/simplify.rs
torquestomp: This also lets me re-use the 'simplify_ancestor_merge()' function

CHANGELOG.md
```diff
@@ -64,6 +64,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * CommitId / ChangeId template types now support `.normal_hex()`.
 
+* New command `jj simplify` will remove redundant parent edges.
```
thoughtpolice, line 67: `simplify-parents`

samueltardieu (commented):
nit: the commit message still contains `simplify` instead of `simplify-parents`

torquestomp:
> nit: the commit message still contains `simplify` instead of `simplify-parents`

Find and replace, my greatest weakness

cli/src/commands/simplify_parents.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    let num_old_heads = rewriter.new_parents().len();
+    rewriter.simplify_ancestor_merge();
+    let num_new_heads = rewriter.new_parents().len();
+
+    if num_new_heads < num_old_heads {
+        rewriter.reparent(settings)?.write()?;
+
+        let new_stats = SimplifyStats::for_commit(num_old_heads - num_new_heads);
```
yuja, line 114: nit: maybe simpler to just do `stats.commits += 1; ...`?

cli/src/commands/simplify_parents.rs
```diff
@@ -0,0 +1,119 @@
...
+    if rewriter.old_commit().parent_ids().len() <= 1 {
+        return Ok(());
+    }
+
+    let num_old_heads = rewriter.new_parents().len();
+    rewriter.simplify_ancestor_merge();
+    let num_new_heads = rewriter.new_parents().len();
+
+    if num_new_heads < num_old_heads {
+        rewriter.reparent(settings)?.write()?;
```
yuja, line 112: probably better to do `.reparent()` if `.parents_changed()` so the descendants won't be rebased later by `tx.finish()`.

yuja:
> `simplify-parents` seems rather wordy to be in the top-level namespace, but other than that I think it's fine. I suspect it might be best to go under a `utils` subcommand or something like we've talked about before, but I'm interested in what other people think.

Yeah. If we had a generic graph reshaping command or command namespace, `simplify` can be moved there. This could also be `rebase --simplify-ancestor-merge`, but it's inconvenient if the user case is to eliminate redundant edges created by autorebase.

@ilyagr any thoughts?

ilyagr:
I think `simplify-parents` is fine for now, unless we get a better idea.

One alternative I can think of is to have a subcommand, e.g. `jj graph simplify-parents` or `jj graph-util simplify-parents`. 

Yet another option would be to make `jj rebase` a namespace with `jj rebase source` instead of `jj rebase -s`, and then we could have `jj rebase simplify-parents`.

Another command that could go into this namespace (or be an option to rebase) is `jj reparent-children --from X --to Y` (or `move-children`) that replaces `X` with `Y` in the list of any commit's children (without breaking merge commits).

torquestomp:
I like the idea of `jj graph simplify-parents`, it would probably make sense to also move `jj parallelize` into this subgroup?  (Also is anyone making `jj linearize` yet, that would also fit here?)

yuja:
Operation history is also a "graph", so I think `jj graph` is still vague.

> it would probably make sense to also move `jj parallelize` into this subgroup?

It might make sense, but I tend to prefer flattened command structure. `jj parallelize` isn't so ambiguous compared to `jj simplify` (without `-parents`/`-merge`.)

martinvonz:
If we add a `graph` command category, then it seems like even `rebase` would belong under it, but if we do that, then basically everyone would create an alias for it anyway.

ilyagr:
> Operation history is also a "graph", so I think jj graph is still vague.

This is a good point. `jj commits` or `jj commitgraph` or `jj cgraph` seem like obvious alternatives, but I'm not sure they are great. Or we could have `jj graph`, `jj evo graph` (or `evo-graph` or `evograph`), and `jj op graph`.

> If we add a graph command category, then it seems like even rebase would belong under it, but if we do that, then basically everyone would create an alias for it anyway.

I wouldn't mind `jj rebase` being an alias for `jj graph rebase`. On the other hand, if we just want the uncommon commands to be in a subcommand, it could be `jj graph-utils simplify-parents` or `jj util graph simplify-parents`.

----

Another idea, based on Yuya's comment is `jj simplify parents`. But, in any case, we could start with `jj simplify-parents`.

cli/src/commands/simplify_parents.rs
```diff
@@ -0,0 +1,119 @@
...
+use jj_lib::settings::UserSettings;
+
+use crate::cli_util::CommandHelper;
+use crate::cli_util::RevisionArg;
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Simplify parent edges for the specified revision(s).
+///
+/// Removes all redundant parent edges going into any of the specified
```
ilyagr, line 16: (Optional) The following seems a bit clearer to me for the first sentence:

> Removes any parents of each of the the specified revisions that are also indirect ancestors of the same revision

For the second sentence, I'd also be more decisive:

> This has no effect on any revision's contents or on the working copy.

I'm also tempted to write "This should have no effect on any...", and I'm OK with that if you prefer it, but I think style guides or Orwell's "Politics and the English language" would suggest dropping the "should"

cli/src/commands/simplify_parents.rs
```diff
@@ -0,0 +1,119 @@
...
+/// In other words, for all (A, B, C) where A has (B, C) as parents and C is an
+/// ancestor of B, A will be rewritten to have only B as a parent instead of
+/// B+C.
+#[derive(clap::Args, Clone, Debug)]
+#[command(group = clap::ArgGroup::new("revision-args").multiple(true).required(true))]
+pub(crate) struct SimplifyParentsArgs {
+    /// Simplify specified revision(s) together with their trees of descendants
+    /// (can be repeated)
+    #[arg(long, short, group = "revision-args")]
+    sources: Vec<RevisionArg>,
```
ilyagr, line 29: I think we use singular "source" in `rebase`, and that singular is also better (since to the user it looks like `--source a --source b`, where `a` and `b` usually expand to a single revision).

cli/src/commands/simplify_parents.rs
```diff
@@ -0,0 +1,119 @@
...
+
+pub(crate) fn cmd_simplify_parents(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &SimplifyParentsArgs,
+) -> Result<(), CommandError> {
+    let mut workspace_command = command.workspace_helper(ui)?;
+    let revs = RevsetExpression::descendants(
+        workspace_command
+            .parse_union_revsets(ui, &args.sources)?
```
ilyagr, line 43: Relatedly to the previous comment, I think this should use `resolve_some_revsets_default_single`.

ilyagr (commented):
For now, here are a few bikesheddy comments. I hope they are uncontroversial, but I thought I'd send them out early in just in case there are other opinions.

cli/src/commands/simplify_parents.rs
```diff
@@ -0,0 +1,119 @@
...
+            .expression(),
+    )
+    .union(
+        workspace_command
+            .parse_union_revsets(ui, &args.revisions)?
```
ilyagr, line 48: (No action needed) For "revisions", `rebase` currently uses plural and `parse_union_revsets`, so I think things are correct as they are.

ilyagr:
By the way, I agree with this approach of dealing with redundant ancestry overall. It seems more predictable to me than trying to do simplification automatically in some situations. (Well, I suppose, if we have some great ideas about auto-simplification, this PR doesn't prevent us trying to do that too, but I'm not optimistic about that at the moment).

cli/tests/test_simplify_parents_command.rs
```diff
@@ -0,0 +1,142 @@
...
+    let stdout = test_env.jj_cmd_success(&repo_path, &["log", "-r", "all()", "-T", "description"]);
+    insta::allow_duplicates! {
+    insta::assert_snapshot!(stdout, @r###"
+    @    c
+    ├─╮
+    │ ○  b
+    ├─╯
+    ○  a
+    ◆
+    "###);
```
ilyagr, line 120: Another interesting case to me (that I think we should test) is `jj simplify-parents -r 'all()'` for something like

```
    @    d
    ├─╮
    ○ │  c
    │ ○  b
    ├─╯
    ○  a
    ◆
```

I think it should be a no-op, and I haven't read the code carefully enough to check whether it will be.

cli/src/commands/simplify_parents.rs
yuja: A bit off topic, but do you remember why `rebase -s`/`-b` requires `all:`? I don't think `-s multiple-roots` is a footgun. `-s`/`-b` are similar to `-r`.

ilyagr: `-s multiple-roots` can be a footgun, at least if it acts in the way that makes sense to me.

I'd like `-s A -s B -d X` to be equivalent to `-s all:A|B -d X` and result in both `A` and `B` becoming direct children of `X`, *independently* of whether one of `A` or `B` is a descendant of the other one.

If so, `rebase -s A::` would give a very confusing result, and so I'd prefer that gave an error, and the user did `rebase -s all:A::` if that's what they wanted.

There might be other ways to negotiate this problem, which would involve changing what `-s multiple_revision` means (e.g. automatically rewrite it to `-s roots(multiple_revisions)`, making `-s A|B` result in different thing if `B` is a child of `A`). However, the behavior I described seems sensible to me and seems to require `all:`.

ilyagr: Though, this makes me think that all of my arguments for `rebase` do not apply to `simplify-parents`. So, maybe we could keep using `parse_union_revset` here. I'm not sure whether or not this slight extra flexibility is worth making the UI subtly different from `rebase`'s.

I think, for `simplify-parents`, I still mildly prefer making the UI the same as `rebase`'s, but I no longer think it's particularly important.

yuja: > I'd like `-s A -s B -d X` to be equivalent to `-s all:A|B -d X` and result in both `A` and `B` becoming direct children of `X`, _independently_ of whether one of `A` or `B` is a descendant of the other one.

Ah, good point. For `-s A|B`, the user might expect `roots(A|B)` to be the destination children (if `-s` accepted revision ranges by default.)

> Though, this makes me think that all of my arguments for `rebase` do not apply to `simplify-parents`.

Yes. `simplify-parents` is similar to `jj fix -s`.

ilyagr: >  For -s A|B, the user might expect roots(A|B) to be the destination children

Yes, or `roots(connected(A|B)`. Not having to choose between these slightly different behaviors is one advantage of the current behavior.

cli/src/commands/simplify_parents.rs
torquestomp: Rewored this mostly according to the suggestions.

CHANGELOG.md
```diff
@@ -101,6 +101,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+
+* New command `jj simplify-parents` will remove redundant parent edges.
+
+ ### Fixed bugs
+ 
+  * Update working copy before reporting changes. This prevents errors during reporting
+    from leaving the working copy in a stale state.
+ 
+  * `jj git fetch -b <remote-git-branch-name>` will now warn if the branch(es)
+>>>>>>>
```
ilyagr, line 113: There's a forgotten conflict marker here.

cli/tests/test_simplify_parents_command.rs
ilyagr: Thanks for adding! I hadn't noticed when you did it. I'll hopefully take one last look at this soon.

CHANGELOG.md
```diff
@@ -101,6 +101,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   content.
 
 * `jj git fetch -b <remote-git-branch-name>` will now warn if the branch(es)
+
+* New command `jj simplify-parents` will remove redundant parent edges.
```
samueltardieu, line 105: It looks like this will be inserted in the 0.22 release section, this should be moved up to the unreleased section.

ilyagr (approved):
This all LGTM. The new clippy warning should be easy to fix. Thank you!

bnjmnt4n:
Would it be a good idea to set a default revset for `jj simplify-parents` (like [`jj fix`](https://github.com/martinvonz/jj/pull/3866))?

Issue #4293, opened 2024-08-16
`jj git fetch` succeeds in pulling a non-existent branch name


rdamazio:
## Description

If I specify a non-existent branch name to `jj git fetch`, it doesn't fail.

## Steps to Reproduce the Problem

1. `jj git clone https://github.com/rdamazio/efis-editor.git`
1. `cd efis-editor`
1. `jj git remote add zyv https://github.com/zyv/efis-editor.git`
2. `jj git fetch -b features/foreflight-support --remote zyv` (misspelled branch name)

## Expected Behavior

It'd tell me that the branch I asked to fetch does not exist.

## Actual Behavior

`Nothing changed`

## Specifications

- Platform: macOS 14.6.1 (23G93)
- Version: `jj 0.17.1-a17292317c63df7aad14c3203046f683b847fa69`

yuja:
Perhaps, warning can be displayed if a branch glob doesn't match any of the changed or known remote branches.

PR #4494, opened 2024-09-17
cli: Warn if specified remote branch not found for `jj git fetch`

essiene:
* First fetch from remote.
* Then check tx.{base_repo(),repo}.view().remote_bookmarks_matching(<branch>, <remote>).
  This has to happen after the fetch has been done so the tx.repo() is updated.
* Warn if a branch is not found in any of the remotes used in the fetch. Note that the remotes
  used in the fetch can be a subset of the remotes configured for the repo, so the language
  of the warning tries to point that out.

Fixes: #4293

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/fetch.rs
+++ b/cli/src/commands/git/fetch.rs
@@ -21,6 +21,7 @@ use jj_lib::settings::UserSettings;
 use jj_lib::str_util::StringPattern;
 
 use crate::cli_util::CommandHelper;
+use crate::cli_util::WorkspaceCommandTransaction;
 use crate::command_error::user_error;
 use crate::command_error::user_error_with_hint;
 use crate::command_error::CommandError;
@@ -100,6 +101,12 @@ pub fn cmd_git_fetch(
         })?;
         print_git_import_stats(ui, tx.repo(), &stats.import_stats, true)?;
     }
+    warn_if_branches_not_found(
+        ui,
+        &tx,
+        &args.branch,
+        &remotes.iter().map(StringPattern::exact).collect_vec(),
+    )?;
     tx.finish(
         ui,
         format!("fetch from git remote(s) {}", remotes.iter().join(",")),
@@ -139,4 +146,35 @@ fn get_all_remotes(git_repo: &git2::Repository) -> Result<Vec<String>, CommandEr
         .iter()
         .filter_map(|x| x.map(ToOwned::to_owned))
         .collect())
+}
+
+fn warn_if_branches_not_found(
+    ui: &mut Ui,
+    tx: &WorkspaceCommandTransaction,
+    branches: &[StringPattern],
+    remotes: &[StringPattern],
+) -> Result<(), CommandError> {
+    for branch in branches {
+        let matches = remotes.iter().any(|remote| {
+            tx.repo()
+                .view()
+                .remote_bookmarks_matching(branch, remote)
+                .next()
+                .is_some()
+                || tx
+                    .base_repo()
+                    .view()
+                    .remote_bookmarks_matching(branch, remote)
+                    .next()
+                    .is_some()
+        });
+        if !matches {
+            writeln!(
+                ui.warning_default(),
+                "No branch matching `{branch}` found on any specified/configured remote",
+            )?;
+        }
+    }
+
+    Ok(())
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+ * `jj git fetch -b <remote-git-branch-name>` will now warn if the branch(es)
+   can not be found in any of the specified/configured remotes.
+
  * Fixed panic when parsing invalid conflict markers of a particular form.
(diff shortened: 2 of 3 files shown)


cli/src/commands/git/fetch.rs
```diff
@@ -98,6 +100,15 @@ pub fn cmd_git_fetch(
             GitFetchError::InternalGitError(err) => map_git_error(err),
             _ => user_error(err),
         })?;
+        for pattern in &args.branch {
```
essiene, line 103: Open question: This will issue a warning for *each* remote. So if a user has multiple remotes configured in `git.fetch` in the config file, this will issue a warning for a remote where the branch was not found, even if it finds it for the other remote(s).

thoughts?

samueltardieu: I would have thought that being in one remote would be enough to not get a warning, as it means this is probably not a typo.

samueltardieu: I'm not sure I understand your answer @PhilipMetzger. If `git.fetch = ["origin", "upstream"]` and the user issues `git fetch -b my-topic-branch`, do you think it warrants a warning that `my-topic-branch` is not found on upstream, even if it is found in `origin`? IMO the intention of the user is clear here.

essiene: > IMO the intention of the user is clear here.

I'm not so sure it's that clear. I think any of the below could be true:

* fetch `my-topic-branch` from only `origin`?
* fetch `my-topic-branch` from both `origin` and `upstream`, so they can diff/cherrypick locally?
* actually meant to fetch from `upstream` but didn't realize it existed on `origin`. Whoops!

Life would be much simpler if remote branches were always specified as <branch>@<remote>, alas...

yuja: > being in one remote would be enough to not get a warning, as it means this is probably not a typo.

+1

> if a branch glob doesn't match any of the changed or known remote branches

I just said that the implementation would have to scan changed refs as well as the known `repo.view()` (or `old_repo.view()` + `new_repo.view()` might be simpler.)

samueltardieu: This is a matter of taste I guess, but I usually display warnings only when both conditions are true:
- There is a high probability that the user did something wrong (otherwise, there would be too many false positive, and the warning becomes annoying) – for example using `jj log XXX` where `XXX` doesn't match any file, and is probably meant to be a revset.
- I cannot prove that they did something wrong (otherwise, that would be an error, not a warning) – the user might have really wanted to check if any revision had touched file `XXX` in `jj log XXX`.

If the glob expands to at least one branch on one remote, I'm afraid I couldn't determine with a high probability that it is indeed a mistake, as such use would be valid. I, for one, would not warn in this case.

essiene: > Thanks for laying out your thoughts. This is convincing to me. Especially as you and @yuja think the same. I'll go ahead and change the warning to only fail if missing from _all_ remotes.

Done.

> I just said that the implementation would have to scan changed refs as well as the known repo.view()

Switched the implementation to this. Thanks for the hint. So there are no library changes now, only cli changes.

cli/src/commands/git/fetch.rs
```diff
@@ -98,8 +106,24 @@ pub fn cmd_git_fetch(
...
+                continue;
+            }
+        }
         print_git_import_stats(ui, tx.repo(), &stats.import_stats, true)?;
     }
+    for branch in args
+        .branch
+        .iter()
+        .filter(|b| !found_branches.contains(b.as_str()))
+    {
```
yuja, line 121: Perhaps, we don't have to clone `repo.view()` (which is expensive), build `found_branches` set, etc.

Here we can access to `tx.repo()` (= new state) and `tx.base_repo()` (= old state), and we'll need to emit warning if the branch pattern doesn't match either of these repos? Something like (untested):
```rust
let remote_patterns = remotes.iter().map(StringPattern::exact).collect_vec();
for branch in &args.branch {
    // or [views..].iter().cartesian_product(&remote_patterns) ?
    let matched = remote_patterns.iter().any(|remote| tx.repo().view().remote_bookmarks_matching(branch, remote).next().is_some()) || ...;
    if !matched {
        warn ...
    }
}
```

essiene: Done. This is simpler and I think better.

It would have been nice to issue the warning before the fetch is done, instead of after, but life can't be perfect.

Thanks for the code pointers. Much appreciated as usual.

cli/src/commands/git/fetch.rs
```diff
@@ -100,6 +101,12 @@ pub fn cmd_git_fetch(
         })?;
         print_git_import_stats(ui, tx.repo(), &stats.import_stats, true)?;
     }
+    warn_if_branches_not_found(
+        ui,
+        &tx,
+        &args.branch[..],
+        &remotes.iter().map(StringPattern::exact).collect_vec()[..],
```
yuja, line 108: nit: `[..]` can be omitted

yuja (approved):
Thanks

cli/src/commands/git/fetch.rs
essiene: Removed. Thanks.

PR #4341, opened 2024-08-25
everything: Rename `branches` to `bookmarks`

PhilipMetzger:
Jujutsu's branches do not behave like Git branches, which is a major hurdle for people adopting it from Git. They rather behave like Mercurial's (hg) bookmarks. 

We've had multiple discussions about it in the last ~1.5 years about this rename in the Discord,  where multiple people agreed that this _false_ familiarity does not help anyone. Initially we were  reluctant to do it but overtime, more and more users agreed that `bookmark` was a better for name  the current mechanism. This may be hard break for current `jj branch` users, but it will immensly  help Jujutsu's future, by defining it as our first own term. The `[experimental-moving-branches]`  config option is currently left alone, to force not another large config update for users, since the last time this happened was when `jj log -T show` was removed, which immediately  resulted in breaking users and introduced soft deprecations.

This name change will also make it easier to introduce Topics (#3402) as _topological branches_  with a easier model. 

This is a pure mechanical change done via LSP, ripgrep and sed with some manual fixups.

cc @thoughtpolice for help writing the deprecation text in the CHANGELOG.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [x] I have updated the documentation (README.md, docs/, demos/)
- [x] I have updated the config schema (cli/src/config-schema.json)

PR #4495, opened 2024-09-17
docs: Rework `bookmarks.md`

PhilipMetzger:
This applies some suggestions from Ilya from #4341 and starts to differentiate in terminology.

cc @ilyagr for edits. 

# Checklist

If applicable:

### DIFF
--- a/docs/bookmarks.md
+++ b/docs/bookmarks.md
@@ -11,9 +11,8 @@ pass a bookmark's name to commands that want a revision as argument. For example
 `jj bookmark list` to list bookmarks and `jj bookmark` to create, move, or delete
 bookmarks. There is currently no concept of an active/current/checked-out bookmark.
 
-Currently Jujutsu maps its Bookmarks to Git Branches and stores them as that 
-in the Git backend. This means that all Bookmarks will be reflected as 
-Git Branches, this may change in the future. 
+In a Jujutsu repo that's colocated with Git, Jujutsu's bookmarks currently map 
+to Git branches, this may change in the future. 
 
 ## Remotes and tracked bookmarks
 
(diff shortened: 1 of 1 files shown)


docs/bookmarks.md
```diff
@@ -11,8 +11,8 @@ pass a bookmark's name to commands that want a revision as argument. For example
 `jj bookmark list` to list bookmarks and `jj bookmark` to create, move, or delete
 bookmarks. There is currently no concept of an active/current/checked-out bookmark.
 
-Currently Jujutsu maps its Bookmarks to Git Branches and stores them as that 
-in the Git backend. This means that all Bookmarks will be reflected as 
+In a Jujutsu repo that's colocated with Git, Jujutsu's bookmarks map to Git 
+branches. This means that all Bookmarks will be reflected as
```
samueltardieu, line 15: I thought that it was true as soon as git is used as a backend, even in non-colocated repositories. Am I wrong?

PhilipMetzger: its based on this comment https://github.com/martinvonz/jj/pull/4341#discussion_r1747911931.

martinvonz: > I thought that it was true as soon as git is used as a backend, even in non-colocated repositories. Am I wrong?

You're right. Well, it was more correct before than after at least, but the old form also sounds a bit incorrect to me. I think "stores them as that in the Git backend" makes it sound like we actually store bookmarks as branches in the Git backend. We store bookmarks in the operation store (which has no Git backend). We have `jj git export/import` to export/import from/to our operation store to Git. The export/import requires a Git backend, but it is not the source of truth for storage of bookmarks.

Issue #3419, opened 2024-04-01
Should `jj split` leave branches on the first commit instead of moving to the second?


emesterhazy:
I'm forking this off of https://github.com/martinvonz/jj/issues/2274#issuecomment-2028818012 where Martin wrote:

>> I don't think that checking out the second commit is an issue (presumably we have to pick one), but perhaps we should leave any branches in a conflicted state. Thoughts?

> Good question. I think it would make more sense to leave the branch in a conflicted state, but it seems like it's practically quite useful to leave it on one of the commits instead. That way it requires action from the user in some of the cases instead of all of the cases. I'd say we should leave it that way for now and change it later if we hear that it's confusing - or worse, that it leads people to push the wrong commit to a remote.

> By the way, we have had a discussion internally at Google about where regular (parent/child) jj split should put the branch. It currently leaves it on the child, but it lets the parent keep the change id. We use branches for tracking code reviews at Google (I know you're at Google, @emesterhazy; I'm just providing context for others). For that use case, it's confusing that the branch doesn't follow the change id. It makes sense for the branch to move to the child in more Git-like workflows, since branches typically point to the tip of a chain of commits in Git. However, I don't want us to design a behavior that makes less sense just because it better matches Git. What do people think?

I'm going to think about this a bit more before giving an opinion, but I wanted to give this its own issue for discussion.

emesterhazy:
I've been thinking about this a bit more and would like to propose that we change the current behavior.

Currently, the ~branch~ bookmark follows the child revision, moving forward the way branches usually do in a git workflow.

- This doesn't make a lot of sense from the jj centric viewpoint though where bookmarks are aliases for a revision / change id, and it causes issues for workflows where each change id is reviewed separately (stacked PRs and Google's internal workflow).

Instead of moving the bookmark to the child revision, I think split should leave bookmarks in place and attached to the same change id before and after the split. If users want to move the bookmark, they can. This no-implicit-move behavior aligns the change that makes abandon drop bookmarks instead of moving them to the parent revision. 

@PhilipMetzger, @martinvonz What do you think?

PhilipMetzger:
I also think it makes sense, since it aligns with the proposed model for `jj gerrit send` (#2845), It also explicitly requires a `jj bookmark move` which I also find great.

martinvonz:
[...]

PR #4496, opened 2024-09-17
WIP: docs/design: Topics

PhilipMetzger:
Continues the conversation started in Discord and hopefully concludes it with a final design.

### DIFF
--- a/docs/design/topics.md
+++ b/docs/design/topics.md
@@ -1,0 +1,90 @@
+# Topics
+
+Authors: [Philip Metzger](mailto:philipmetzger@bluewin.ch), [Noah Mayr](mailto:dev@noahmayr.com)
+ [Anton Bulakh](mailto:him@necaq.ua)
+
+## Summary
+
+Introduce Topics as a truly Jujutsu native way for topological branches, which 
+also replace the current bookmark concept for Git interop. As they have been
+documented to be confusing users coming from Git. They also replace the 
+`[experimental-advance-branches]` config for those who currently use it, as 
+such a behavior will be built-in for Topics.
+
+## Prior work 
+
+Currently there only is Mercurial which has a implementation of 
+[Topics][hg-topic]. There also is the [Topic feature][gerrit-topics] in Gerrit,
+which groups commits with a single identifier.
+
+
+## Goals and non-goals
+
+### Goals
+
+The goals for this Project are small, see below.
+
+* Introduce the concept of topological branches for Jujutsu.
+* Simplify Git interop by reducing the burden on `jj bookmark`.
+* Add Change metadata as a concept.
+* Remove the awkward `bookmark` to Git `branch` mapping.
+
+### Non-Goals
+
+## Overview
+
+A detailed overview of the project and the improvements it brings.
+
+### Detailed Design
+
+
+#### Storage
+
+We should store `Topic` metadata as pure commit metadata in the serialized proto:
+
+
+```protobuf
+message Commit {
+  //...
+  map<string, string> topics = N;
+}
+```
+
+while the actual code should look like this:
+
+```rust
+#[derive(ContentHash, ...)]
+struct Commit {
+  //...
+  #[ContentHash(ignore = true)]
+  topics: HashMap<String, String>
+}
+```
+
+#### Backend implications
+
+If Topics were stored as commit metadata, it would allow backends to drop 
+the metadata if necessary. 
+
+TODO: Write it
+
+## Alternatives considered 
+
+
+### Store them not on the Commit
+
+See [#1889][prototype] for another idea of keeping the  
+
+Other alternatives to your suggested approach, and why they fall short.
+
+### Single Head Topics
+
+While these are conceptually simpler, they wouldn't help with ...
+
+## Future Possibilities
+
+The section for things which could be added to it or deemed out of scope during
+the discussion.
+
+[hg-topics]:
+[gerrit-topics]: 
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -141,3 +141,4 @@ nav:
     - 'Sparse Patterns v2': 'design/sparse-v2.md'
     - 'Tracking branches': 'design/tracking-branches.md'
     - 'Copy tracking and tracing': 'design/copy-tracking.md'
+    - 'Topics': 'design/topics.md'


docs/design/topics.md
```diff
@@ -0,0 +1,168 @@
...
+
+
+TODO: Example here
+
+#### Major Use-Cases 
+
+
+##### Obsoletion of `[experimental.auto-advance-bookmarks]`
+
+Since Topics are infectious by nature, they can perfectly map to a Git branch.
```
sheremetyev, line 70: how would a topic with multiple heads map to a single Git branch (or maybe multiple branches)?

BatmanAoD: This sentence is unclear to me as well. I'm guessing what it's supposed to mean is that topics "perfectly map to" what people _expect from_ Git branches, rather than that they'd actually be implemented as a 1:1 equivalent for Git branches. But since Git branches don't actually work the way Topics are described (i.e. they are not merely "infectious" but actually implicitly apply to all ancestor commits, and they cannot have multiple heads), they don't actually "map perfectly" to Git branches as understood by people who deeply understand the Git model (which is probably the minority of Git users, but certainly a nontrivial contingent).

Certainly this sentence _cannot_ mean that topics are always represented in the git storage backend as branches, since that would be incompatible with both of the differences cited above.

docs/design/topics.md
```diff
@@ -0,0 +1,168 @@
...
+revisions with a `archive:` description as archived. For this use-case the 
+non-contiguous property of Topics really shine.
+
+##### Checkout history
+
+### Detailed Design
+
+#### Git Interop
+
+For all continuous Topics we can just simply export them as Git Branches. For
```
BatmanAoD, line 91: Guessing this is a typo, especially since it's contrasted with "non-contiguous":

```suggestion
For all contiguous Topics we can just simply export them as Git Branches. For 
```

docs/design/topics.md
```diff
@@ -0,0 +1,168 @@
...
+
+##### Checkout history
+
+### Detailed Design
+
+#### Git Interop
+
+For all continuous Topics we can just simply export them as Git Branches. For 
+Topics which contain non-contiguous parts, we should allow exporting them 
+one-by-one or with a generated name, such as `<topic-name>-1` for each part.
```
BatmanAoD, line 93: What would "one-by-one" mean, and why is it distinct from having a generated name? I'm guessing it means that the CLI would let the user decide to export just one contiguous "section" of a topic without exporting the others, but [a] what would the actual benefit to the user be, and [b] wouldn't there still need to be an auto-generated name in order to prevent future conflict when exporting another "section" of the same topic?

docs/design/topics.md
```diff
@@ -0,0 +1,168 @@
...
+important. So having a 
+`jj topic create -r 'trunk()..@ & description("archive:")' should mark them all
+revisions with a `archive:` description as archived. For this use-case the 
+non-contiguous property of Topics really shine.
+
+##### Checkout history
+
+### Detailed Design
+
+#### Git Interop
```
BatmanAoD, line 89: There's a part of this that seems to be missing: where does a topic "end"? I'm guessing that you would export one git branch for each head commit in the topic, rather than one git branch for _every commit_ in the topic (since that wouldn't really have much advantage compared to `jj git push -c`). But git doesn't have a concept of where a branch "starts" except relative to other branches, and then only in the context of commands (such as rebase) that start by finding common ancestors, which implicitly "prune" the branch. So for the "change-archival" use-case, what commits would be included in the `archive` topic?

I'm guessing the answer is something like: `topic(x) := ::heads(topic) ~ ::trunk()`, i.e. topic `x` contains the heads plus all ancestors _minus_ everything on the `trunk` "branch". I think this is consistent with the property of `hg` topics that they "disappear" as soon as they are merged (though I haven't used `hg` and am not totally sure I followed the linked `topic` documentation). But whatever the rule is, it should be explicit.

docs/design/topics.md
PhilipMetzger: > how would a topic with multiple heads map to a single Git branch (or maybe multiple branches)?

Each separate head will get its own Git branch. There's also the variant of connecting them, but I do not think that is desirable, as the different heads should represent different kind of logical work.  

> This sentence is unclear to me as well. I'm guessing what it's supposed to mean is that topics "perfectly map to" what people _expect from_ Git branches, rather than that they'd actually be implemented as a 1:1 equivalent for Git branches. But since Git branches don't actually work the way Topics are described (i.e. they are not merely "infectious" but actually implicitly apply to all ancestor commits, and they cannot have multiple heads), they don't actually "map perfectly" to Git branches as understood by people who deeply understand the Git model (which is probably the minority of Git users, but certainly a nontrivial contingent).

That is correct. But sharing the conceptual behavior never meant that we as Jujutsu VCS need to represent them the same internally.

docs/design/topics.md
PhilipMetzger: > What would "one-by-one" mean, and why is it distinct from having a generated name?

The distinction is mostly manual. 

> I'm guessing it means that the CLI would let the user decide to export just one contiguous "section" of a topic without exporting the others, but [a] what would the actual benefit to the user be, and [b] wouldn't there still need to be an auto-generated name in order to prevent future conflict when exporting another "section" of the same topic?

On point A, we give the user the choice of when they want to export another slice of a topic and on B the previous slice's branch may have already been deleted, so we'd be free to choose the name again.

docs/design/topics.md
PhilipMetzger: > There's a part of this that seems to be missing: where does a topic "end"? I'm guessing that you would export one git branch for each head commit in the topic, rather than one git branch for _every commit_ in the topic (since that wouldn't really have much advantage compared to `jj git push -c`).

Yes, this is also correct. 

> So for the "change-archival" use-case, what commits would be included in the `archive` topic?

This is @avamsi's use-case from #4180. This means that you locally add a description like `archive:` to your commits and then let `jj` infer it in someway or occasionally do it manually with a command. 

> I'm guessing the answer is something like: `topic(x) := ::heads(topic) ~ ::trunk()`, i.e. topic `x` contains the heads plus all ancestors _minus_ everything on the `trunk` "branch".

Yes, that is correct and thanks for the formulation. 

> But whatever the rule is, it should be explicit.

Yes, I'll write it down.

docs/design/topics.md
jennings: Pushing the multiple heads of a topic as `foo-1`, `foo-2`, etc., is surprising to me. I was expecting topics to work roughly like this:

- A commit can manually be labeled a part of topic "foo".
- A new commit created on top of that commit will automatically be labeled "foo".
- If `heads(topic(foo))` has a single revision, the topic may be pushed as a git branch named "foo" with a command like `jj git push -t foo`
- If `heads(topic(foo))` has more than one revision, pushing the topic fails with an error (similar to hg refusing to create multiple remote heads)

I'm not sure why I would ever want to push a topic and get branches `foo-1`, `foo-2`, etc. The only reason might be to create a remote backup of my work, but I don't expect topics to fix this problem (it's something to solve for any anonymous head).

I think forges like Gerrit would support pushing multiple heads in the same topic, but I would expect `jj gerrit` to handle this.

I also don't know how creating multiple branches for a topic works as the graph changes over time. Suppose I had commits B and C in a single topic, both descending from A, and I pushed, creating branches `foo-1` and `foo-2`. Then I create a merge commit D also in the topic. Does pushing delete `foo-1` and `foo-2`, and replace them with just `foo`?

```
D
|\
B |  foo-1@origin
| C  foo-2@origin
|/
A
```

docs/design/topics.md
jennings: nit: This is something to handle whenever the topic has multiple heads, not just if the topic is disconnected. For example, the following topography is connected but can't be exported as one git branch:

```
B    topic=foo
|
| C  topic=foo
|/
A    topic=foo
```

PhilipMetzger:
Also cc @necauqua and @noahmayr since they were heavily involved in the first discussion.

docs/design/topics.md
```diff
@@ -0,0 +1,168 @@
...
+while the actual code should look like this:
+
+```rust
+#[derive(ContentHash, ...)]
+struct Commit {
+  //...
+  //
+  // This avoids rewriting the Change-ID, but must be implemented.
+  #[ContentHash(ignore)]
+  topics: HashMap<String, String>
```
yuja, line 124: What do `key` and `value` represent here? `key = topic_name, value = ??`? Isn't it just a set of topic names?

FWIW, I assume topics are commit metadata, hence the commit id changes when topics are updated.

docs/design/topics.md
PhilipMetzger: There will be export restrictions on Topics, as we cannot represent this as a Git Branch. This will hit the auto-generated name  I mentioned previously.

docs/design/topics.md
PhilipMetzger: > What do `key` and `value` represent here? `key = topic_name, value = ??`? Isn't it just a set of topic names?

How would a hashset implement a Topic on two different remotes then? It may also be worth it to just rename it to metadata. 

> WIW, I assume topics are commit metadata, hence the commit id changes when topics are updated.

I choose this option because of https://github.com/martinvonz/jj/pull/3613#issuecomment-2094280350, as you should be able to rewrite those lightweight tags easily. If we can change the underlying commit-id w/o rewriting the Change-ID it should be unnecessary.

docs/design/topics.md
martinvonz: FWIW, we currently skip export of bookmarks when they're conflicted, so it wouldn't be that different to skip exporting topics when they have multiple heads.

PhilipMetzger: > FWIW, we currently skip export of bookmarks when they're conflicted, so it wouldn't be that different to skip exporting topics when they have multiple heads.

Yes, that would also would be a variant of dealing with the issue.

docs/design/topics.md
martinvonz: Oh, are you saying that topics will not be stored in the commit metadata? I think that needs to be clarified then. It wasn't clear at all to me. (The doc says that topics will be stored in a metadata map, but it doesn't say where that map is stored.)

PhilipMetzger: > Oh, are you saying that topics will not be stored in the commit metadata?

Where did I say that? 

> The doc says that topics will be stored in a metadata map, but it doesn't say where that map is stored.)

On the Commit which makes this dance I think necessary.

martinvonz: > Where did I say that?

I was just asking if you did because it wasn't clear what you meant. You said "I choose this option" without saying which option you meant. It was in reply to Yuya's "I assume topics are commit metadata", which made me think it would be stored in the commit, but then you linked to a comment that said "After some discussion with @PhilipMetzger we came to the conclusion that we can't actually store topics on commits", which made me think that you meant that it should not be stored on commits.

> On the Commit which makes this dance I think necessary.

I see. I think that information needs to go in the doc too. Thanks.

necauqua:
> While this works it falls short of having the metadata synced by multiple clients, which is something desirable. The prototype thus also avoids rewriting the Change-ID which is a good thing, but makes them only locally available.

Uh, the part where this was discussed completely escaped me?.
Why is "having the metadata synced by multiple clients" something desirable? I mean it surely would be nice, with an actuall jj remote that's yet to exist, but without that we have to resort to annoying and intrusive things like rewriting commit messages or similar.

I thought we were settled on them being locally available (like the oplog, for example) and for easy editing of topics without actually rewriting commits.

And for sharing them through a git remote, roundtripping through git branches while only really works nice for single-head topics that branch off the trunk, is plenty imo.

Interoping with some gerrit features would be then the job of the gerrit integration etc.

docs/design/topics.md
yuja: > How would a hashset implement a Topic on two different remotes then? It may also be worth it to just rename it to metadata.

Oh, so `topics` is a map of `remote_name: topic_name`? I was just wondering why it was a map at all, because I just thought topic is a local thing (and mirrored to Git branch at remote.)

docs/design/topics.md
```diff
@@ -0,0 +1,168 @@
...
+* Remove the awkward `bookmark` to Git `branch` mapping.
+
+### Non-Goals
+
+* Making bookmarks unnecessary.
+
+## Overview
+
+Until now, Jujutsu had no native set of topological branches, just 
+[Bookmarks][bm] which interact poorly with Git's expectation of branches.
```
BatmanAoD, line 44: I'm not sure if it makes sense to refer to "Git's expectation" separately from the expectations of Git _users_, but if we do want to call out poor interaction between bookmarks and branches, then I think we should be more specific about where the tooling actually diverges in behavior. Git has a concept of a "current branch" that `jj` completely lacks, and it auto-advances the current branch, whereas `jj` doesn't auto-advance bookmarks except using the experimental feature that this design hopes to deprecate. 

```suggestion
[Bookmarks][bm] which interact poorly with Git's expectation of a "current branch" that is advanced automatically as commits are added. 
```

docs/design/topics.md
BatmanAoD: It looks like this section now disagrees with the backend-implementation section below:

> For the Git backend, we could either embed them in the message, like Arcanist or Gerrit do or store them as Git Notes, if necessary.

Or am I misunderstanding, and topics will be exported to Git as notes _and_ branches?

emilazy: I kinda like the idea of exporting them as both. Notes will avoid modifying the commit hashes while supporting faithfully exporting discontiguous topics. We could perhaps use notes as the source of truth for syncing and treat branches as purely for compatibiltiy?

scott2000:
I have a few questions which don't seem to be answered yet in the design doc:

* Will jj bookmarks still be able to track remote git branches?
  * If so, can a bookmark and a topic track the same remote git branch?
* Will jj bookmarks still be pushed as git branches?
* Will git branches still be fetched as jj remote bookmarks?
* Can git branches be fetched as topics even if they were created by someone else?
  * If so, how does jj know what the base branch is?
    * Is there a command to set the base branch for fetching a topic?
    * If it defaults to `trunk()`, is it impossible to fetch `trunk()` (or a branch that's an ancestor of `trunk()`) as a topic since there's zero commits in that range?
    * How does this work for projects with multiple long-lived branches?
* Is there such a thing as a "remote topic"?
  * If so, what determines whether a git branch is fetched as a remote bookmark or a remote topic?
  * Or are there only remote bookmarks still, and there's an explicit command to start tracking a remote bookmark as a local topic?
* Can a topic and a bookmark have the same name?
  * If so, which takes precedence? Do they have different syntax?
* Can a topic be conflicted?
  * If so, how are topic conflicts resolved?
* If there's a series of commits `trunk(), A, B, C` and I set a topic as `B, C` and push it, I assume the remote git branch will be pointing to `C`. What happens if a coworker rebases the branch, creating the new series of commits `trunk(), X, A', B', C'` with the remote git branch pointing to `C'` now? When I fetch, does my topic get expanded to `X, A', B', C'`, or does it become `B', C'` somehow? Or does it become conflicted?

PhilipMetzger:
I'm going to create threads for top-level questions, since GitHub's UI for them isn't great and I expect that this will receive more than a handful of comments. Since this is isn't nearly done yet.

docs/design/topics.md
```diff
@@ -0,0 +1,168 @@
...
+or Gerrit do or store them as Git Notes, if necessary. 
+
+## Alternatives considered 
+
+### Storing Topics out-of-band 
+
+See [Noah's prototype][prototype] for the variant of keeping them out of band.
+While this works it falls short of having the metadata synced by multiple 
+clients, which is something desirable. The prototype thus also avoids rewriting
+the Change-ID which is a good thing, but makes them only locally available.
```
PhilipMetzger, line 144: @necauqua 
> Uh, the part where this was discussed completely escaped me?.
Why is "having the metadata synced by multiple clients" something desirable? I mean it surely would be nice, with an actuall jj remote that's yet to exist, but without that we have to resort to annoying and intrusive things like rewriting commit messages or similar.


This Design Doc contains two features which I desire for `jj`: 
1. Topics (Git Interop, a tagging mechanism)
2. Commit Metadata (Topics are a part of them)

And we always should design for the native future and something which is compatible with the Google backend. 

I only remembered the drawback of keeping them out of band from an earlier Design Doc, but I am sadly unable to find that conversation with Torquestomp. 

>I thought we were settled on them being locally available (like the oplog, for example) and for easy editing of topics without actually rewriting commits.

Yes, we agreed on that for Topics. But for pure Commit Metadata we never decided on it and it is feasible without to much cost, in my opinion. 

>And for sharing them through a git remote, roundtripping through git branches - while only really works nice for single-head topics that branch off the trunk - is plenty imo.

Agreed.
 
>Interoping with some gerrit features would be then the job of the gerrit integration etc.

Where did you get that from, I never mentioned something like it? AFAIK, Gerrit Topics are once again some special refs on the Gerrit/Git Server.

docs/design/topics.md
PhilipMetzger, line 1: @scott2000 
> I have a few questions which don't seem to be answered yet in the design doc:
> 
> * Will jj bookmarks still be able to track remote git branches (meaning they'll be updated on fetch)?
>   
>   * If so, can a bookmark and a topic track the same remote git branch?

The intent of this work is to remove this behavior and take over the necessary parts from [Tracking Branches](https://github.com/martinvonz/jj/blob/main/docs/design/tracking-branches.md). Maybe some parts will be left intact to have something equivalent to a Git Tag, but I haven't thought deeply about it.

> * Will jj bookmarks still be pushed as git branches?

No, see the first answer.

> * Will git branches still be fetched as jj remote bookmarks?

Ditto. 
> * Can git branches be fetched as topics even if they were created by someone else?

Yes, otherwise this feature would not help with Git interop.
>   * If so, how does jj know what the base branch is?

This will continue to use Git tooling as a implementation, so it will be the merge-base. 

>    Is there a command to set the base branch for fetching a topic?

I don't understand your question, please elaborate. 

>  * If it defaults to `trunk()`, is it impossible to fetch `trunk()` (or a branch that's an ancestor of `trunk()`) as a topic since there's zero commits in that range?

Thanks for this question, this will need some thought. Since the intended design is to allow free rewriting of commits, they can be preserved by the respective backend/server implementation. 

>   * How does this work for projects with multiple long-lived branches?

I don't understand your question here, since I have no base of your expectations here. Can you elaborate?

>  * Is there such a thing as a "remote topic"?

See my first answer. 

>   * If so, what determines whether a git branch is fetched as a remote bookmark or a remote topic?

You'll hopefully only be able to fetch a Git Branch as a Topic, so this shouldn't occur.

> Or are there only remote bookmarks still, and there's an explicit command to start tracking a remote bookmark as a local topic?

See my first answer for my current plan. 

> * Can a topic and a bookmark have the same name?

Maybe? This is to be determined on how they reflect on a backend. 
>   * If so, which takes precedence? Do they have different syntax?

This behavior is to be determined here. 

> * Can a topic be conflicted?

I don't think so, since introducing conflicting metadata seems to be a problem I don't understand.
>   * If so, how are topic conflicts resolved?

As above.

> * If there's a series of commits `trunk(), A, B, C` and I set a topic as `B, C` and push it, I assume the remote git branch will be pointing to `C`. What happens if a coworker rebases the branch, creating the new series of commits `trunk(), X, A', B', C'` with the remote git branch pointing to `C'` now? When I fetch, does my topic get expanded to `X, A', B', C'`, or does it become `B', C'` somehow? Or does it become conflicted?

I expect the first behavior I think, but that depends on how we want to integrate it.

docs/design/topics.md
PhilipMetzger: > > For the Git backend, we could either embed them in the message, like Arcanist or Gerrit do or store them as Git Notes, if necessary.
> 
> Or am I misunderstanding, and topics will be exported to Git as notes _and_ branches?

These are options beside exporting them as branches, which is _the desired implementation_ for Git interop. We should enumerate all options to fully consider the whole design space.

> We could perhaps use notes as the source of truth for syncing and treat branches as purely for compatibiltiy?

That's a good idea. I just don't know what that encompasses.

docs/design/topics.md
scott2000: > > Is there a command to set the base branch for fetching a topic?
>
> I don't understand your question, please elaborate.

Since a Git branch is just a pointer to a head commit, I believe it's unclear which commits are logically part of a branch unless you're comparing against some base branch/commit, so for fetching new remote branches from Git, I was just wondering how `jj` will determine which commits to include in the corresponding topic and whether it's user-configurable. It seems like there's a lot of options, such as:

1. Topic contains `..branch_head`
2. Topic contains `trunk()..branch_head`
3. Add a new configurable revset alias like `git_topic_base()`, then topic contains `git_topic_base()..branch_head`
4. Add a new configurable revset alias like `git_topic_revisions(branch_head)`, then topic just contains `git_topic_revisions(branch_head)`
5. Store a base commit as metadata with every remote topic, then topic contains `base_commit..branch_head`

My concern for options 2 and 3 was that it's possible for the imported topic to be resolved to an empty set (if the branch head is an ancestor of the base branch/commit), which would need to be resolved somehow.

> > How does this work for projects with multiple long-lived branches?
>
> I don't understand your question here, since I have no base of your expectations here. Can you elaborate?

My concern for option 2 is that some repos have multiple long-lived branches (e.g. branches A and B), and then there are branches off of those branches (e.g. C branches off from A, and D branches off from B). Ideally, it would be nice to be able to import C as containing `A..C` and D as containing `B..D`, but for option 2 they would be imported as `trunk()..C` and `trunk()..D` respectively.

BatmanAoD: I share the concern about branch pointers not being sufficient to indicate which commits are part of a topic.

docs/design/topics.md
necauqua: > This Design Doc contains two features

Well the confusing part is then that the title says "design for topics", while as you just clarified it's actually for this generic maybe-shareable metadata that topics could be an instance of.

I think combining the two into one doc was a mistake - there were a lot of discussions and some sort of vision for the out-of-band topics, but this is first time I'm hearing about the other thing and the doc doesn't seem very actionable because of it, at least at the moment.

> Where did you get that from

Honestly don't remember :joy:
Someone somewhere at some point mentioned somethings like that idk

necauqua: I skimmed through the doc again - you mostly talk about topics specifically in it and a high-level results of the discussions that happened - but then starting on the storage paragraph you suddently derail into this generic commit metadata protobuf thing and it gets confusing - right after saying they could be exported to git as branches, no less

docs/design/topics.md
scott2000: Actually, I've been thinking about it more, and it seems like there could be another problem with the current approach. Currently, `trunk()` is generally defined in terms of remote bookmarks. If we get rid of remote bookmarks and replace them with remote topics, then `trunk()` would have to be defined based on a remote topic instead. However, from what I've heard, the plan is to exclude commits from a topic if they are present in `trunk()`. This leads to a circular definition, where `trunk()` depends on remote topics, but remote topics depend on `trunk()`.

Maybe it would be better to keep remote bookmarks as-is, but just allow tracking a remote bookmark as a topic with a command like `jj topic track <bookmark@remote>`. Then `trunk()` could continue to be defined in terms of remote bookmarks, and there would be no circular dependency. This would also allow a command like `jj topic track <bookmark@remote> --base <revset>` to specify that the topic should include only commits in `<revset>..<bookmark@remote>` instead of the default `trunk()..<bookmark@remote>`. I don't think it would even be necessary to store `<revset>` anywhere, because for future fetches, the local topic could just be updated to `roots(local_topic)-..updated_remote_bookmark`.

Remote topics still could be supported, but they just wouldn't be used for the Git backend. It's conceivable that a native backend might want to support both remote bookmarks and remote topics, just like Git has both tags and branches.

docs/design/topics.md
emesterhazy: I also have concerns about this. From the doc:

> Single Head Topics - While these are conceptually simpler, they wouldn't help with Git interop where it is useful to map a single underlying Topic to multiple Git branches.

Why is it useful to map a single topic to multiple Git branches? Git doesn't have any similar concept as far as I know, and Git branches always have a single head.

docs/design/topics.md
emesterhazy: Why do topics need to support non-contiguous commits? What compelling use case does it enable? It seems like non-contiguous topics are just a way to add additional metadata to commits which could be useful for selecting commits later. If that's the use case, why don't we just add a separate and more flexible metadata system which topics can integrate with by automatically setting some metadata on commits belonging to the topic?

docs/design/topics.md
PhilipMetzger: > Why is it useful to map a single topic to multiple Git branches?

Since Topics actually can be disjoint, each subpart can be exported as a branch. Otherwise we're unnecessary restricting the capabilities of Topics. 

```text
I    G 'foo
|    | 
H    F 'foo
|  / 
E
~ (large distance of commits is elided)
C 'foo 
| 
B 'foo
| 
A 'foo
```
this tries to give a answer to the graph above. As they're all on Topic "foo" here but represent two distinct branches with two heads.

emesterhazy:
I think the doc needs to explain the details of how topics are propagated infectiously. What happens, for example, if I rebase a commit that belongs to a topic onto another topic (including the anonymous topic). Does it cease to be part of that topic, or do I end up with a single disjoint commit belonging to the original topic?

docs/design/topics.md
emesterhazy: But why is it useful for Topics to be disjoint, and why is it useful for Topics to have multiple heads? Some examples of why this is a useful property would be helpful. I understand the desire to attach metadata to disjoint commits (i.e. these commits are all related to feature xyz), but it's unclear to me why we need to optimize for that in a design for topological branches.

PhilipMetzger:
I wanted to give an update on this thing, I have largely rewritten this doc in last 6 months but to me it's still not _good enough_ to go out of the WIP state. I will update it soonish though.

PhilipMetzger:
Thanks for the feedback @jennings. 

> I think the most important thing to document is _what topics are meant to accomplish_.

Agreed, that always is/was the goal. It just happened to be that this design doc got attention in an incomplete state where I rolled with the punches which changed the contents. It (and I) never were ready for that so the state pretty much reflects that.

> I think different people expect them to solve different problems, and it's not clear to me what we're trying to do.

I think there are at least five different versions of Topics floating around, where one is mine which made it onto a design doc others pretty much only exist as "vibes" in active discussions. Whether the project wants a design doc for it or just an implementation like #3419 is pretty much up to the community and the maintainers.  

> > * Introduce the concept of native topological branches for Jujutsu.
> 
> Why do we want this concept? Doesn't this already exist just by nature of the commit graph being a DAG? We just don't name the branches, but why do we want to?

I think this is addressed in my local version, but I think the named part has been missing the whole time. 

> > * Simplify Git interop by reducing the burden on Bookmarks.
> 
> What's painful about it? I think this is specifically to address advancing bookmarks, but we should explain the use case explicitly.

👍 

> > * Add Change metadata as a storage concept.
> 
> Why do we want to do this? Isn't this already handled by the new `change-id` header? (It's possible this simply predates the header and is no longer necessary)

It is a separate thing I want, which is reconcilable with my ideal implementation, someone else probably could make Topics work without that since it then would only encompass 10% of the current design.  

> > * Remove the awkward bookmark to Git branch mapping.
> 
> What's awkward about it?

The concepts don't match all in my opinion. Bookmarks give a name to a single change which doesn't really fit the expectations of Git branches. 

I will push the revised v2 with your points addressed and open the discussion again but after that it's unlikely that I will spend further time on it, since the last months can be considered a waste.   

Let me explain my motivations here, until about a year ago there were constant complaints that `jj` is to tedious to use for feature branches and long-living Git branches and that `jj tug` or its variants are inadequate.  Topics should hopefully solve that, since it gives those users what they want. It definitely isn't my problem so I think another person which is affected by it should step up. Like I could punch this through (the willpower is here), but I am no longer motivated to do so. 

Anyway, I've recently realized that my time as a contributor and community member in the project has an expiry date which is likely sooner rather than later.

docs/design/topics.md
```diff
@@ -0,0 +1,168 @@
...
+// A simple Key-Value pair. 
+message StringPair {
+  string key = 1;
+  string value = 2;
+  // Could be extended by a protobuf.Any see the future possibilities section.
+}
+
+message Commit {
+  //...
+  repeated StringPair metadata = N;
```
martinvonz, line 111: For context, Google would probably use this for shipping things like "submitted_piper_change_number: 123" from the server to the client

Issue #2866, opened 2024-01-22
`jj <alias> --help` works, but `jj help <alias>` does not


torquestomp:
Example: Add an alias `fetch = git fetch` in your config, run `jj help fetch`.  You get `error: unrecognized subcommand 'fetch'`

ilyagr:
I wouldn't say that `jj <alias> --help` works properly, either.

E.g., if I have an aliase `d = ["diff", "-s"]`, `jj d --help` is the same as `jj diff --help`. I'd prefer if `jj d --help` told me that `d` is an alias. We could *also* make it print `jj diff --help` afterwards, or we could not do that.

torquestomp:
I was hoping this would be a quick in-and-out fix, but it appears that the way `help` is actually invoked is deep in the internals of clap, so it's not as simple as I hoped.  I may come back to this later but I'll unassign for now.

martinvonz:
> but it appears that the way `help` is actually invoked is deep in the internals of clap

Yes, I suspect we'll need to take over handling of all help messages in order to fix this.

jyn514:
Rather than implementing `help` manually, it might make sense to go through `visible_alias`, which makes help work out of the box without having to reimplement it: https://docs.rs/clap/latest/clap/struct.Command.html#method.visible_alias
(i took a look at reimplementing help and it looks decidedly non-trivial, there's lots of digging through clap's reflection API.)

jyn514:
ah, i see, aliases can be multiple commands, they don't correspond 1-1 with a subcommand. so visible_alias won't work.

jyn514:
for anyone picking this up in the future - the way clap does this internally is https://docs.rs/clap_builder/4.4.18/src/clap_builder/builder/command.rs.html#4474-4487

joyously:
Is there a place where the "not found" error can be checked, and a list of aliases could be listed instead?

jyn514:
@joyously do you mean the "unrecognized subcommand" error? yes, but that place is in clap itself: https://github.com/jyn514/clap/blob/a7e04a53e4a9e310663a0cdb5bfb0753d84b82bd/clap_builder/src/error/format.rs#L529

LucasWagler:
Related: `jj b --help` displays the top-level help text, instead of the `bookmark` help text. Other aliases seem to behave similarly.

This seems to contradict behavior mentioned in this previous comment, as the help page now differs between the full subcommand and any aliases to that subcommand:
[...]

PR #4518, opened 2024-09-21
Accept early args after help command

Grillo-0:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/help.rs
+++ b/cli/src/commands/help.rs
@@ -1,0 +1,49 @@
+// Copyright 2024 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use tracing::instrument;
+
+use crate::cli_util::CommandHelper;
+use crate::command_error;
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Print this message or the help of the given subcommand(s)
+#[derive(clap::Args, Clone, Debug)]
+pub(crate) struct HelpArgs {
+    /// Print help for the subcommand(s)
+    pub(crate) command: Vec<String>,
+}
+
+#[instrument(skip_all)]
+pub(crate) fn cmd_help(
+    _ui: &mut Ui,
+    command: &CommandHelper,
+    args: &HelpArgs,
+) -> Result<(), CommandError> {
+    let mut args_to_show_help = vec![command.app().get_name()];
+    args_to_show_help.extend(args.command.iter().map(|s| s.as_str()));
+    args_to_show_help.push("--help");
+
+    // TODO: `help log -- -r` will gives an cryptic error, ideally, it should state
+    // that the subcommand `log -r` doesn't exist.
+    let help_err = command
+        .app()
+        .clone()
+        .subcommand_required(true)
+        .try_get_matches_from(args_to_show_help)
+        .expect_err("Clap library should return a DisplayHelp error in this context");
+
+    Err(command_error::cli_error(help_err))
+}
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -30,6 +30,7 @@ mod evolog;
 mod file;
 mod fix;
 mod git;
+mod help;
 mod init;
 mod interdiff;
 mod log;
@@ -71,6 +72,7 @@ use crate::command_error::CommandError;
 use crate::ui::Ui;
 
 #[derive(clap::Parser, Clone, Debug)]
+#[command(disable_help_subcommand = true)]
 enum Command {
     Abandon(abandon::AbandonArgs),
     Backout(backout::BackoutArgs),
@@ -110,6 +112,7 @@ enum Command {
     Fix(fix::FixArgs),
     #[command(subcommand)]
     Git(git::GitCommand),
+    Help(help::HelpArgs),
     Init(init::InitArgs),
     Interdiff(interdiff::InterdiffArgs),
     Log(log::LogArgs),
@@ -217,6 +220,7 @@ pub fn run_command(ui: &mut Ui, command_helper: &CommandHelper) -> Result<(), Co
         }
         Command::Fix(args) => fix::cmd_fix(ui, command_helper, args),
         Command::Git(args) => git::cmd_git(ui, command_helper, args),
+        Command::Help(args) => help::cmd_help(ui, command_helper, args),
         Command::Init(args) => init::cmd_init(ui, command_helper, args),
         Command::Interdiff(args) => interdiff::cmd_interdiff(ui, command_helper, args),
         Command::Log(args) => log::cmd_log(ui, command_helper, args),
--- a/cli/tests/test_help_command.rs
+++ b/cli/tests/test_help_command.rs
@@ -1,0 +1,86 @@
+// Copyright 2024 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::common::TestEnvironment;
+
+#[test]
+fn test_help() {
+    let test_env = TestEnvironment::default();
+
+    let help_cmd_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help"]);
+    // The help command output should be equal to the long --help flag
+    let help_flag_stdout = test_env.jj_cmd_success(test_env.env_root(), &["--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    // Help command should work with commands
+    let help_cmd_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help", "log"]);
+    let help_flag_stdout = test_env.jj_cmd_success(test_env.env_root(), &["log", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    // Help command should work with subcommands
+    let help_cmd_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["help", "workspace", "root"]);
+    let help_flag_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["workspace", "root", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    // Help command should not work recursively
+    let stderr = test_env.jj_cmd_cli_error(test_env.env_root(), &["workspace", "help", "root"]);
+    insta::assert_snapshot!(stderr, @r#"
+    error: unrecognized subcommand 'help'
+
+    Usage: jj workspace [OPTIONS] <COMMAND>
+
+    For more information, try '--help'.
+    "#);
+
+    let stderr = test_env.jj_cmd_failure(test_env.env_root(), &["workspace", "add", "help"]);
+    insta::assert_snapshot!(stderr, @r#"
+    Error: There is no jj repo in "."
+    "#);
+
+    let stderr = test_env.jj_cmd_failure(test_env.env_root(), &["new", "help", "main"]);
+    insta::assert_snapshot!(stderr, @r#"
+    Error: There is no jj repo in "."
+    "#);
+
+    // Help command should output the same as --help for nonexistent commands
+    let help_cmd_stderr = test_env.jj_cmd_cli_error(test_env.env_root(), &["help", "nonexistent"]);
+    let help_flag_stderr =
+        test_env.jj_cmd_cli_error(test_env.env_root(), &["nonexistent", "--help"]);
+    assert_eq!(help_cmd_stderr, help_flag_stderr);
+
+    // Some edge cases
+    let help_cmd_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help", "help"]);
+    let help_flag_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    let stderr = test_env.jj_cmd_cli_error(test_env.env_root(), &["help", "unknown"]);
+    insta::assert_snapshot!(stderr, @r#"
+    error: unrecognized subcommand 'unknown'
+
+      tip: a similar subcommand exists: 'undo'
+
+    Usage: jj [OPTIONS] <COMMAND>
+
+    For more information, try '--help'.
+    "#);
+
+    let stderr = test_env.jj_cmd_cli_error(test_env.env_root(), &["help", "log", "--", "-r"]);
+    insta::assert_snapshot!(stderr, @r#"
+    error: a value is required for '--revisions <REVISIONS>' but none was supplied
+
+    For more information, try '--help'.
+    "#);
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Evaluation error of `revsets.short-prefixes` configuration is now reported.
 
+* Help command doesn't work recursively anymore, i.e. `jj workspace help root`
+  doesn't work anymore.
+
 ### Deprecations
 
 ### New features
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2842,7 +2842,7 @@ pub struct EarlyArgs {
     // Option<bool>.
     pub quiet: Option<bool>,
     /// Disable the pager
-    #[arg(long, value_name = "WHEN", global = true, action = ArgAction::SetTrue)]
+    #[arg(long, global = true, action = ArgAction::SetTrue)]
     // Parsing with ignore_errors will crash if this is bool, so use
     // Option<bool>.
     pub no_pager: Option<bool>,
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -56,6 +56,7 @@ This document contains the help content for the `jj` command-line program.
 * [`jj git remote remove`↴](#jj-git-remote-remove)
 * [`jj git remote rename`↴](#jj-git-remote-rename)
 * [`jj git remote set-url`↴](#jj-git-remote-set-url)
+* [`jj help`↴](#jj-help)
 * [`jj init`↴](#jj-init)
 * [`jj interdiff`↴](#jj-interdiff)
 * [`jj log`↴](#jj-log)
@@ -126,6 +127,7 @@ To get started, see the tutorial at https://martinvonz.github.io/jj/latest/tutor
 * `file` — File operations
 * `fix` — Update files with formatting fixes or other changes
 * `git` — Commands for working with Git remotes and the underlying Git repo
+* `help` — Print this message or the help of the given subcommand(s)
 * `init` — Create a new repo in the given directory
 * `interdiff` — Compare the changes of two commits
 * `log` — Show revision history
@@ -1161,6 +1163,18 @@ Set the URL of a Git remote
 
 * `<REMOTE>` — The remote's name
 * `<URL>` — The desired url for `remote`
+
+
+
+## `jj help`
+
+Print this message or the help of the given subcommand(s)
+
+**Usage:** `jj help [COMMAND]...`
+
+###### **Arguments:**
+
+* `<COMMAND>` — Print help for the subcommand(s)
 
 
 
--- a/cli/tests/runner.rs
+++ b/cli/tests/runner.rs
@@ -44,6 +44,7 @@ mod test_git_remotes;
 mod test_git_submodule;
 mod test_gitignores;
 mod test_global_opts;
+mod test_help_command;
 mod test_immutable_commits;
 mod test_init_command;
 mod test_interdiff_command;
--- a/cli/tests/test_global_opts.rs
+++ b/cli/tests/test_global_opts.rs
@@ -555,6 +555,10 @@ fn test_early_args() {
     let stdout = test_env.jj_cmd_success(test_env.env_root(), &["--color=always", "help"]);
     insta::assert_snapshot!(stdout.lines().find(|l| l.contains("Commands:")).unwrap(), @"[1m[4mCommands:[0m");
 
+    // Check that early args are accepted after the help command
+    let stdout = test_env.jj_cmd_success(test_env.env_root(), &["help", "--color=always"]);
+    insta::assert_snapshot!(stdout.lines().find(|l| l.contains("Commands:")).unwrap(), @"[1m[4mCommands:[0m");
+
     // Early args are parsed with clap's ignore_errors(), but there is a known
     // bug that causes defaults to be unpopulated. Test that the early args are
     // tolerant of this bug and don't cause a crash.


cli/src/commands/help.rs
```diff
@@ -0,0 +1,58 @@
+// Copyright 2020 The Jujutsu Authors
```
martinvonz, line 1: 2024

cli/src/commands/help.rs
```diff
@@ -0,0 +1,58 @@
...
+            .render_help()
+    } else {
+        command.app().clone().render_help()
+    };
+
+    let help_str = if ui.color() {
+        help_str.ansi().to_string()
+    } else {
+        help_str.to_string()
+    };
```
martinvonz, line 54: It's probably better to call `handle_clap_error()` here. That reduces some duplication and means that pager support still works (when `--no-pager` is not passed).

cli/src/commands/help.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use crate::cli_util::CommandHelper;
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Print this message or the help of the given subcommand(s)
+#[derive(clap::Args, Clone, Debug)]
+pub(crate) struct HelpArgs {
+    /// The command to show help
+    command: Option<String>,
```
martinvonz, line 27: The current clap-provided help supports things like `jj help git push`, so this probably need to be a `Vec<String>` to support that. Also add tests for that case and anything similar you can think of. Oh, `jj git help` and `jj git help push` also work, apparently. I don't know if users expect those combinations to work. It might be fine to ignore some of them. What do you and others think? Should we at least keep things like `jj git push help`? But if we do, it's probably not much harder to support `jj git help push` too.

martinvonz (commented):
Thanks!

By the way, this is a good step towards #2866, in case you want to work on that next :)

cli/src/commands/help.rs
Grillo-0: Maybe even better, would be to just return with a `CommandError::Cli` with the `DisplayHelp` inside it?

martinvonz: Yes, that does seem better. Good point :)

cli/src/commands/help.rs
Grillo-0: > The current clap-provided help supports things like `jj help git push`, so this probably need to be a `Vec<String>` to support that. Also add tests for that case and anything similar you can think of.

I agree, already working on that.

> Oh, `jj git help` and `jj git help push` also work, apparently. I don't know if users expect those combinations to work. It might be fine to ignore some of them. What do you and others think? Should we at least keep things like `jj git push help`? 
 
For me, `jj git help` is expected, but `jj git push help` not that much, but, as you said, if one is implemented, the other most probably will be too. I will think how to tackle this.

Grillo-0:
> Thanks!
> 
> By the way, this is a good step towards #2866, in case you want to work on that next :)

Interesting, I'll check it out ;)

cli/src/commands/help.rs
Grillo-0: > Should we at least keep things like `jj git push help`

Actually `jj git push help` doesn't work

cli/src/commands/mod.rs
```diff
@@ -172,13 +174,80 @@ struct DummyCommandArgs {
...
 #[instrument(skip_all)]
 pub fn run_command(ui: &mut Ui, command_helper: &CommandHelper) -> Result<(), CommandError> {
-    let subcommand = Command::from_arg_matches(command_helper.matches()).unwrap();
+    let subcommand = if has_help_subcmd(command_helper.matches()) {
+        Command::Help(help::HelpArgs {
+            command: get_help_arg_cmd(command_helper.matches()),
+        })
+    } else {
+        Command::from_arg_matches(command_helper.matches()).unwrap()
+    };
```
martinvonz, line 249: Do you think it would be simpler to remove `has_help_subcmd()` and let `get_help_arg_cmd()` return `Option<Vec<String>>`?

yuja:
Just an idea. Maybe we could use a hack similar to `resolve_aliases()`? Something like:
1. walk subcommands with `allow_external_subcommands(true).disable_help_subcommand(true)`
2. if `command_name` is `"help"`, stop there and find `"--no-pager"`
3. move `"--no-pager"` earlier and reparse args normally

Grillo-0:
I used `git add -f` to add the new snapshots. Please tell me if this is not the intend way to do this.

cli/tests/test_help_command.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    // Help command should work recursively
+    let help_cmd_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["workspace", "help", "add"]);
+    insta::assert_snapshot!(help_cmd_stdout);
+    let help_flag_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["workspace", "add", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    let _ = test_env.jj_cmd_failure(test_env.env_root(), &["workspace", "add", "help"]);
```
Grillo-0, line 49: I need some input here. I just copied the current behavior, but maybe it would be good to accept this now?

Grillo-0:
> Just an idea. Maybe we could use a hack similar to `resolve_aliases()`? Something like:
> 
>     1. walk subcommands with `allow_external_subcommands(true).disable_help_subcommand(true)`
> 
>     2. if `command_name` is `"help"`, stop there and find `"--no-pager"`
> 
>     3. move `"--no-pager"` earlier and reparse args normally

This could work, it is a smaller change I think. I can change to this if you guys think it's better.

cli/src/commands/mod.rs
```diff
@@ -172,13 +174,80 @@ struct DummyCommandArgs {
...
+        });
+    }
+
+    ret
+}
+
 pub fn default_app() -> clap::Command {
-    Command::augment_subcommands(Args::command())
+    let app = Command::augment_subcommands(Args::command()).disable_help_subcommand(true);
+    add_help_cmd_to_subcommands(
```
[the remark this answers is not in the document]
Grillo-0: Is there a way to add a test for this?

martinvonz:
> I used `git add -f` to add the new snapshots. Please tell me if this is not the intend way to do this.

Those should not have to be committed. You should be able to just delete them.

> This could work, it is a smaller change I think. I can change to this if you guys think it's better.

Sounds good to me, thanks.

Grillo-0:
> > I used `git add -f` to add the new snapshots. Please tell me if this is not the intend way to do this.
> 
> Those should not have to be committed. You should be able to just delete them.
> 

Strange, I did this because the CI was failing due to the lack of those snapshots

martinvonz:
> Strange, I did this because the CI was failing due to the lack of those snapshots

Ah, it's the `insta::assert_snapshot!()`. We use inline snapshots instead. Search for `, @"` and you'll find lots of examples. However, maybe we don't actually need to test the full output? You already test that the output matches the `--help` output. Maybe that's sufficient?

martinvonz:
Sorry about the delay! I blame it on the travel and week full of meetings. I'll review this now.

cli/tests/test_help_command.rs
```diff
@@ -0,0 +1,46 @@
...
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    // Help command should work recursively
+    let help_cmd_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["workspace", "help", "add"]);
+    let help_flag_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["workspace", "add", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    let _ = test_env.jj_cmd_failure(test_env.env_root(), &["workspace", "add", "help"]);
```
martinvonz, line 45: It might be good to test this output to make sure it's failing for the right reason (that the command is actually run):

```rust
    let stderr = test_env.jj_cmd_failure(test_env.env_root(), &["workspace", "add", "help"]);
    insta::assert_snapshot!(stderr, @r###"
    Error: There is no jj repo in "."
    "###);
```

cli/tests/test_help_command.rs
```diff
@@ -0,0 +1,46 @@
...
+    let help_cmd_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help", "log"]);
+    let help_flag_stdout = test_env.jj_cmd_success(test_env.env_root(), &["log", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    // Help command should work with subcommands
+    let help_cmd_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["help", "workspace", "add"]);
+    let help_flag_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["workspace", "add", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
```
martinvonz, line 36: We should have a test for e.g. `jj new help main` (which is supposed to create a merge commit on top of the `help` and `main` commits. That doesn't work with the current version of this PR, so the PR needs to be fixed. Maybe the best way of fixing it is revert to the version with the real `help` command?

yuja: Oh, maybe we'll have to set `.disable_help_subcommand(true)`, etc. to all subcommands recursively? That's a bit more than I thought.

BTW, if we take this approach, we can enable the hack only after clap returned a `DisplayHelp` error.

Grillo-0: > BTW, if we take this approach, we can enable the hack only after clap returned a `DisplayHelp` error.

The issue is that when `--no-pager` is used after a `help` command, a `DisplayHelp` is not returned.

Grillo-0: > We should have a test for e.g. `jj new help main` (which is supposed to create a merge commit on top of the `help` and `main` commits). That doesn't work with the current version of this PR, so the PR needs to be fixed. Maybe the best way of fixing it is revert to the version with the real `help` command?

I think maybe there are ways to fix that in the current version, but I also think that new bugs like this might appear. So I think the version with a real `help` command is a more solid choice.

cli/tests/test_help_command.rs
Grillo-0: Sure, will add this.

cli/tests/test_help_command.rs
yuja: > The issue is that when `--no-pager` is used after a `help` command, a `DisplayHelp` is not returned.

Got it. It's unfortunate that `help` doesn't ignore unknown args as `--help` would do.

> I think maybe there are ways to fix that in the current version, but I also think that new bugs like this might appear. So I think the version with a real `help` command is a more solid choice.

Suppose we have to modify all subcommands recursively anyway, I agree it might be simpler to just duplicate `help` commands.

martinvonz: Having our own `help` command will also be useful for implementing the `jj help <category>` if we ever want something like that. The `<category>` there might be e.g. `revset`.

Oh, because I figured I should explain what I meant by `<category>` I also happened remind myself of the other type of help categories we want, which is that the top-level `jj help` would be easier to read if we grouped commands into categories.

yuja: I agree we'll eventually need our own `jj help`. I don't think `jj <subcommand> help` should support all the arguments (such as `jj <subcommand> help revset`), though.

Grillo-0: > Having our own `help` command will also be useful for implementing the `jj help <category>` if we ever want something like that. The `<category>` there might be e.g. `revset`.
> 
> Oh, because I figured I should explain what I meant by `<category>` I also happened remind myself of the other type of help categories we want, which is that the top-level `jj help` would be easier to read if we grouped commands into categories.

Interesting, I like the idea to not have to go to the docs for information on these sorts of things. I might try to work on something along those lines after this is accepted.

martinvonz: > I don't think `jj <subcommand> help` should support all the arguments (such as `jj <subcommand> help revset`), though.

I agree. I was surprised to see that that currently works. Dropping that support seems like a relatively minor breaking change.

> Interesting, I like the idea to not have to go to the docs for information on these sorts of things. I might try to work on something along those lines after this is accepted.

Thanks! Just keep in mind that we want a way of not having to duplicate documentation. We've talked about this somewhere in the past but I'm not sure if it was on GitHub or on Discord. I'd rather just point to the website than to duplicate content, but it would be nice if we can find a way to share the content.

cli/tests/snapshots/runner__test_help_command__help-2.snap
yuja, line 1: use inlined `insta::assert_snapshot(.. @"")` instead?

cli/tests/test_help_command.rs
yuja: > > I don't think `jj <subcommand> help` should support all the arguments (such as `jj <subcommand> help revset`), though.
> 
> I agree. I was surprised to see that that currently works. Dropping that support seems like a relatively minor breaking change.

Do you mean we can deprecate/remove `jj <subcommand> help` in favor of `jj help <subcommand>`? I agree that as we can remove hacks from this PR.

Grillo-0: We would deprecate `jj <subcommand> help` but keep `jj <subcommand> help <subsubcommand>`? Or we would just have a primary `help` command and no sub ones?

Grillo-0: Basically only letting the user get help for a sub-subcommand with `jj help <subcommand> <sub-subcommand>`

Grillo-0: Ohhhh, that makes sense. I like this, it reduces code complexity, too ;).

cli/tests/snapshots/runner__test_help_command__help-2.snap
Grillo-0: These `.snap`'s shouldn't exist, fixed now, thx.

cli/src/commands/mod.rs
```diff
@@ -175,12 +177,49 @@ struct DummyCommandArgs {
...
 
 #[instrument(skip_all)]
 pub fn run_command(ui: &mut Ui, command_helper: &CommandHelper) -> Result<(), CommandError> {
-    let subcommand = Command::from_arg_matches(command_helper.matches()).unwrap();
+    let subcommand = if let Some(command) = get_help_arg_cmd(command_helper.matches()) {
+        Command::Help(help::HelpArgs { command })
+    } else {
+        Command::from_arg_matches(command_helper.matches()).unwrap()
+    };
+
```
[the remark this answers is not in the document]
Grillo-0: Oh, totally forgot to remove that, custom commands work without this too.

cli/tests/test_help_command.rs
```diff
@@ -0,0 +1,57 @@
...
+    let help_cmd_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help", "log"]);
+    let help_flag_stdout = test_env.jj_cmd_success(test_env.env_root(), &["log", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
+
+    // Help command should work with subcommands
+    let help_cmd_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["help", "workspace", "root"]);
+    let help_flag_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["workspace", "root", "--help"]);
+    assert_eq!(help_cmd_stdout, help_flag_stdout);
```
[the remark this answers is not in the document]
Grillo-0: For me, the `--help` flag should work like a global flag but `jj --help nonexistent` should not work, as you are asking help for a nonexistent subcommand.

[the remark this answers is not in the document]
martinvonz: I agree. The only reason I said "probably" is that it's currently supported. I'm not sure why it is (deliberate behavior by clap? bug in how we use it?), but that doesn't matter.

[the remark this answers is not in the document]
Grillo-0: Yeah, I checked that it works after I read your comment. Bug or not, for me, it's an estrange behavior. Is there something to do here?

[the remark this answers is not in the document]
martinvonz: I think it would be good to have tests of the `--help` flag in various positions. And it would be good to have tests with non-existent commands too, both `jj help nonexistent` and `jj nonexistent --help`. I assume the latter will be handled by clap, so just check check that the output is the same, as with the existing tests. Thanks.

cli/tests/test_help_command.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let stderr = test_env.jj_cmd_failure(test_env.env_root(), &["workspace", "add", "help"]);
+    insta::assert_snapshot!(stderr, @r###"
+                            Error: There is no jj repo in "."
+                            "###);
+
+    let stderr = test_env.jj_cmd_failure(test_env.env_root(), &["new", "help", "main"]);
+    insta::assert_snapshot!(stderr, @r###"
+                            Error: There is no jj repo in "."
+                            "###);
```
yuja, line 56: nit: delete snapshot outputs and rerun `cargo insta test; cargo insta accept` to fix indent.

cli/tests/test_help_command.rs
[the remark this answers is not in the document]
Grillo-0: Hum, actually the current behavior is not what I think is the ideal.

`jj workspace root --help` and `jj workspace --help root` give different outputs, the first one gives the help of `jj workspace root` and the latter gives the help for `jj workspace`. For me, both should give the help for `jj workspace root`. What you guys think?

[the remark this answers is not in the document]
yuja: iirc, that's because clap ignores arguments after `--help`. I don't know the exact reason, but it might be useful if `--help` can be inserted in failing command arguments without removing bad parts.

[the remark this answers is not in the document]
martinvonz: Maybe @epage can tell us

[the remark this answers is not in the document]
epage: imo the expectation is that `--help` is an argument for the command that is being applied to.

Some more concrete thoughts

The current approach of clap's allows you to edit an existing command and add `--help` anywhere to get details on that part of it.

`--help` bails out immediately.  If we treated it as a global flag that runs against whatever subcommands are specified, then any kind of validation error could prevent you from seeing `--help`.  We could go into an `ignore_errors` mode but that is dicey.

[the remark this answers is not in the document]
martinvonz: Do you mean that something like `jj log --help nonexistent` or `jj log --help --nonexistent` would prevent help for `log` from being displayed, and instead resulting in the top-level help being displayed? I actually think it sounds reasonable to do that. `jj help nonexistent` and `jj nonexistent --help` currently error out. I would actually prefer for `jj --help nonexistent` to error out in the same way.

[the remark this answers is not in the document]
epage: I'm not just talking about `jj --help non-existent` but any other parse errors.

cli/src/commands/mod.rs
```diff
@@ -175,7 +177,7 @@ struct DummyCommandArgs {
 }
 
 pub fn default_app() -> clap::Command {
-    Command::augment_subcommands(Args::command())
+    Command::augment_subcommands(Args::command()).disable_help_subcommand(true)
```
yuja, line 180: nit: this can be moved to `Command` attribute (if you like).
```rust
#[command(disable_help_subcommand = true)]
enum Command {
```

cli/tests/test_help_command.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    // Help command should not work recursively
+    let stderr = test_env.jj_cmd_cli_error(test_env.env_root(), &["workspace", "help", "root"]);
+    insta::assert_snapshot!(stderr, @r###"
+                            error: unrecognized subcommand 'help'
+
+                            Usage: jj workspace [OPTIONS] <COMMAND>
+
+                            For more information, try '--help'.
+                            "###);
```
yuja, line 46: nit: can you rerun insta to fix indent?

cli/tests/test_help_command.rs
```diff
@@ -0,0 +1,57 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::common::TestEnvironment;
+
+#[test]
+fn test_help() {
```
yuja, line 18: nit: some other edge cases that might be interesting:

* `["help", "help"]`
* `["help", "unknown"]`

Maybe we can also insert a workaround for `["help", "log", "--", "-r"]`, but I personally don't care.

yuja (approved):
The current state looks good to me, thanks!

cli/tests/test_help_command.rs
Grillo-0: maybe rem

> Maybe we can also insert a workaround for `["help", "log", "--", "-r"]`, but I personally don't care.

Maybe removing all the args that are not GlobalArgs? But, meh, is it worth it?

yuja: > Maybe we can also insert a workaround for `["help", "log", "--", "-r"]`, but I personally don't care.

Alternatively, we can do subcommand lookup (as you did in earlier version), and print help message there instead of manipulating command arguments. It will require a bit more code, but I think it's less hacky.
https://docs.rs/clap/latest/clap/struct.Command.html#method.render_long_help

Since `jj help` is our own help command, we wouldn't have to copy the exact behavior on `jj help unknown subcommand`.

Grillo-0: But the previous subcommand lookup would break if you put some flags in the middle, as it would treat the flag as a subcommand.

yuja: Just to be sure, with the current implementation, `help log -- -r` will be transformed to `log -r --help`. The error message is a bit cryptic. Instead, `help log -- -r` should ideally error out saying that the subcommand "log -r" doesn't exist.

martinvonz: I agree with Yuya, but I also think the current state of the PR already seems like an improvement, so I'm personally fine with merging it and improving it in a follow-up if you prefer, Grillo-0. But let's also hear what Yuya thinks.

Grillo-0: A test for `help log -- r`?

yuja: Yes, `help log -- -r` or something that produces a weird command arguments.

martinvonz (approved):

Grillo-0:
Can I merge this?

ilyagr:
I think this may have broken `jj git help push`. `jj help git push` still works.

If there is no easy fix, that's probably OK. If there is, it might be worth fixing. (I haven't read the long discussion in here yet, so I don't know if it was already discussed)

I'm used to this working from other clap programs, I don't know how many people would feel this way. OTOH, it shouldn't be that hard to get used to it.

ilyagr:
I'll also add a reference to a slightly more problematic bug (IMO), #4746. Yuya pointed me to read the discussion here for that; I have not done that yet, but I'm guessing there might not be an easy fix.

ilyagr:
Another problem is that `jj help <tab>` completion no longer works, at least in `fish`. I wonder if it'd be possible to have a `clap` option to keep the command, but only for completion purposes.

Grillo-0:
> **Update:** I see it in the changelog, so I'm guessing it was discussed. So, never mind, I'll post something else if I have more thoughts.


Yeah, it was discussed, here is the discussion:
https://github.com/martinvonz/jj/pull/4518#discussion_r1786242353

Grillo-0:
> Another problem is that `jj help <tab>` completion no longer works, at least in `fish`. I wonder if it'd be possible to have a `clap` option to keep the command, but only for completion purposes.
> 
> I'm not sure whether to promote this to its own issue as a bug/regression. I do like `jj help <tab>` for exploring docs. #4737 could also help with this.

Good catch! For me, it is a regression.

Another way to solve this would be to list all the possible commands to show help with [value_parser](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser) (like it was suggest by @yuja on https://github.com/martinvonz/jj/pull/4630#discussion_r1817961319).

Grillo-0:
> I'll also add a reference to a slightly more problematic bug (IMO), #4746. Yuya pointed me to read the discussion here for that; I have not done that yet, but I'm guessing there might not be an easy fix for that.

Maybe we could make the help flags boolean flags that call the new help command if they are present?

yuja:
> Another way to solve this would be to list all the possible commands to show help with [value_parser](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser) (like it was suggested by @yuja on [#4630 (comment)](https://github.com/martinvonz/jj/pull/4630#discussion_r1817961319)).

I guess it can't handle multi-level subcommands?

I think dynamic completion is the way to go. Maybe we could also cheat that clap's help command exists only in "util completion", but that would disable completion for "help -k <KEYWORD>".

> > I'll also add a reference to a slightly more problematic bug (IMO), #4746. Yuya pointed me to read the discussion here for that; I have not done that yet, but I'm guessing there might not be an easy fix for that.
> 
> Maybe we could make the help flags boolean flags that call the new help command if they are present?

I don't know if clap's `-h`/`--help` handling is intuitive, but I wouldn't add messy workaround for this small problem. `-h` terminates parsing, which I think is okay?

ilyagr:
> I don't know if clap's `-h`/`--help` handling is intuitive, but I wouldn't add messy workaround for this small problem.

I am still not up to date on exactly how messy different workarounds are, and whether this can be fixed in `jj` as opposed to `clap`, but I put a couple of thoughts about why I think this problem matters in https://github.com/martinvonz/jj/issues/4746#issuecomment-2452746205 . I wouldn't call it a major issue that we should panic about, but I wouldn't call it a minor issue either.

For example, in my mind, it is more serious than the issue of tab-completing help, which is inconvenient but is not confusing to beginners.

>  -h terminates parsing, which I think is okay?

I think it'd be OK if we taught people about it. For example, we could say in the beginning or end of the help text:

```console
$ jj rebase --help --no-pager -d abracadabra
Warning. the following arguments were ignored: `--no-pager -d abracadabra`.
<Normal help text follows>
```

This would address the beginner problem (people would still be able to learn that the order of arguments is meant to be irrelevant, and that jj will warn them in the exceptional cases it's relevant), but it feels a bit too annoying in general.

PR #4519, opened 2024-09-21
lib: Supporting building for `wasm32-wasip1`

Earthmark:
It's really strange not putting a description of changes in this message.

This is my first set of JJ managed changes, I hope this works.

### DIFF
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -106,7 +106,8 @@ use crate::working_copy::WorkingCopyStateError;
 
 #[cfg(unix)]
 type FileExecutableFlag = bool;
-#[cfg(windows)]
+// Wasi doesn't have a concept of an executable file.
+#[cfg(any(windows, target_os = "wasi"))]
 type FileExecutableFlag = ();
 
 #[derive(Debug, PartialEq, Eq, Clone)]
@@ -132,7 +133,7 @@ impl FileState {
     fn placeholder() -> Self {
         #[cfg(unix)]
         let executable = false;
-        #[cfg(windows)]
+        #[cfg(any(windows, target_os = "wasi"))]
         let executable = ();
         FileState {
             file_type: FileType::Normal { executable },
@@ -142,7 +143,7 @@ impl FileState {
     }
 
     fn for_file(executable: bool, size: u64, metadata: &Metadata) -> Self {
-        #[cfg(windows)]
+        #[cfg(any(windows, target_os = "wasi"))]
         let executable = {
             // Windows doesn't support executable bit.
             let _ = executable;
@@ -349,7 +350,7 @@ fn file_state_from_proto(proto: &crate::protos::working_copy::FileState) -> File
         #[cfg(unix)]
         crate::protos::working_copy::FileType::Executable => FileType::Normal { executable: true },
         // can exist in files written by older versions of jj
-        #[cfg(windows)]
+        #[cfg(any(windows, target_os = "wasi"))]
         crate::protos::working_copy::FileType::Executable => FileType::Normal { executable: () },
         crate::protos::working_copy::FileType::Symlink => FileType::Symlink,
         crate::protos::working_copy::FileType::Conflict => FileType::Normal {
@@ -371,7 +372,7 @@ fn file_state_to_proto(file_state: &FileState) -> crate::protos::working_copy::F
         FileType::Normal { executable: false } => crate::protos::working_copy::FileType::Normal,
         #[cfg(unix)]
         FileType::Normal { executable: true } => crate::protos::working_copy::FileType::Executable,
-        #[cfg(windows)]
+        #[cfg(any(windows, target_os = "wasi"))]
         FileType::Normal { executable: () } => crate::protos::working_copy::FileType::Normal,
         FileType::Symlink => crate::protos::working_copy::FileType::Symlink,
         FileType::GitSubmodule => crate::protos::working_copy::FileType::GitSubmodule,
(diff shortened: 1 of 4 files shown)


lib/src/local_working_copy.rs
```diff
@@ -106,7 +106,7 @@ use crate::working_copy::WorkingCopyStateError;
 
 #[cfg(unix)]
 type FileExecutableFlag = bool;
-#[cfg(windows)]
+#[cfg(any(windows, target_os = "wasi"))]
```
Earthmark, line 109: I can comment this in the repo if wanted, but wasi doesn't have a concept of file permission flags right now https://github.com/WebAssembly/wasi-filesystem/blob/main/wit/types.wit

martinvonz (approved):
Thanks. I suppose it would make sense to add this build to CI. Do you want to follow up with that?

martinvonz (commented):
If anyone here knows more about WebAssembly (more than ~nothing, that is), please review this. Otherwise I think we can merge this in a day or so.

PhilipMetzger:
Can you please adjust the commit message to `lib: Supporting building for wasm32-wasip1`, see https://github.com/martinvonz/jj/blob/main/docs/contributing.md#code-reviews. And I'd like to know the motivation behind supporting it, if there are any.  And maybe it'd be nice to coordinate the changes with #4478, since that PR is a major rework for `lib/src/local_working_copy.rs`?

lib/src/local_working_copy.rs
thoughtpolice: Yes, please do leave a comment. Is there a ticket or some kind of ideas for this? It would be enough to just leave a comment to a relevant ticket i.e. `// wasi symlink support: https://github.com/webassembly/wasi-filesystem/issue/...`

Earthmark:
Oh something went wrong, I'll see what happened locally and get that resolved.

thoughtpolice (commented):
Seems fine, but let's add a build matrix entry too. I suspect we'll probably need to remove git support for that to work? I would be surprised if `libgit2` and the needed deps all build cleanly.

If we could actually build the CLI without git support, we would in theory throw away most of the external deps, and be able to write a WASI backend that used the browser filesystem APIs. Then we could host a version of JJ in the browser as an interactive tutorial. :)

Earthmark:
> Thanks. I suppose it would make sense to add this build to CI. Do you want to follow up with that?

Step added, it should verify `lib` builds for wasi.

lib/src/local_working_copy.rs
Earthmark: I haven't found any workstreams trying to add executable flag support for wasi.

I would assume executable files would be added in a separate interface. I'm pretty sure wasi doesn't have the ability to execute another process right now in the current spec, marking a file as executable wouldn't be meaningful.

I'll add a comment as well for this line, if I do run into an issue link I'll append that.

Earthmark:
This should probably wait until https://github.com/martinvonz/jj/pull/4478 goes through, I feel that change is likely a higher precedence and it'd be easy to rebuild these changes.

Earthmark:
> Can you please adjust the commit message to `lib: Supporting building for wasm32-wasip1`, see https://github.com/martinvonz/jj/blob/main/docs/contributing.md#code-reviews. And I'd like to know the motivation behind supporting it, if there are any. And maybe it'd be nice to coordinate the changes with #4478, since that PR is a major rework for `lib/src/local_working_copy.rs`?

Commit message adjusted.

The use case for this is building out very trivial vfs-wasi support to https://github.com/RobertBaruch/dergwasm to represent entities in the game world as entries in a filesystem, and running jj as a wasm executable to version objects inside the VR environment, from inside the environment itself.

@thoughtpolice 's use case of a web tool on an example filesystem is a much more realizeable use case. 

I said this in another comment, but I feel #4478 should probably go through first, as it's a much more complex change.

Earthmark:
Turns out the executable bit of every file got set on accident.

My main machine is windows, but I worked on this repo in a linux dev container with the folder mounted into the container. I'm pretty sure it exposes mounted windows files as 755, which got committed to the repo by the dev container. In my initial change I ran jj from the windows env, for the follow up I ran it in the container itself, which is why the issue only came up in the follow up.

martinvonz:
Oops :) Sounds like #4478 will be useful for you too then.

PhilipMetzger:
> > Can you please adjust the commit message to `lib: Supporting building for wasm32-wasip1`, see https://github.com/martinvonz/jj/blob/main/docs/contributing.md#code-reviews. And I'd like to know the motivation behind supporting it, if there are any. And maybe it'd be nice to coordinate the changes with #4478, since that PR is a major rework for `lib/src/local_working_copy.rs`?
> 
> Commit message adjusted.

Thanks. 

> The use case for this is building out very trivial vfs-wasi support to https://github.com/RobertBaruch/dergwasm to represent entities in the game world as entries in a filesystem, and running jj as a wasm executable to version objects inside the VR environment, from inside the environment itself.

That sounds like a really interesting idea, although I think that just having the auto-snapshotting mechanism could be enough for you.

Thank you for enabling the support for it.

Earthmark:
The ability to restructure changes is very useful as well. I'm happy to talk about it over discord if you're curious!

Velociraptor115:
Did this go anywhere?
I was considering using a wasm build of jj-lib from a bun project

EDIT: Oh, is this waiting on https://github.com/jj-vcs/jj/pull/7484?

Earthmark:
I did not follow up on this. I may poke at it again, but don't hold your breath. This should probably just be closed.

pksunkara:
Superseded by #9849

PR #4522, opened 2024-09-23
diff: handle "matching" regions of different lengths, minor cleanup

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/diff.rs
+++ b/lib/src/diff.rs
@@ -342,7 +342,7 @@ fn intersect_unchanged_words(
 
 #[derive(Clone, PartialEq, Eq, Debug)]
 struct UnchangedRange {
-    base_range: Range<usize>,
+    base: Range<usize>,
     offsets: Vec<isize>,
 }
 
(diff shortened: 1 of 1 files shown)


lib/src/diff.rs
```diff
@@ -313,23 +313,28 @@ fn unchanged_ranges_lcs(
...
+struct RangeOffset<T> {
+    pub start: T,
+    pub end: T,
+}
+
 #[derive(Clone, PartialEq, Eq, Debug)]
 struct UnchangedRange {
     base_range: Range<usize>,
-    offsets: Vec<isize>,
+    offsets: Vec<RangeOffset<isize>>,
```
martinvonz, line 326: Now that the base ranges and the other ranges are more independent because we don't enforce that their lengths match, is it simpler to have a single `ranges: Vec<Range<usize>>`?

yuja: Good point, and it revealed a bigger problem that `intersect_regions()` can split existing other ranges at non-token boundary. Perhaps, it has to be based on token indices, but that might be more complicated than preprocessing inputs.

yuja: > it has to be based on token indices, but that might be more complicated than preprocessing inputs.

Actually, it is simpler because intersection of ranges will be a basic merge join of indices. I'll send new PR separately.

Extracted some patches to https://github.com/martinvonz/jj/pull/4527. This PR also contains non-controversial changes to prepare for index-based range intersection.

yuja: Updated to use absolute other ranges. `compact_unchanged_ranges()` should also be fixed.

martinvonz (approved):

PR #4536, opened 2024-09-25
ui: redirect prompt messages to stderr

torquestomp:
# Checklist

If applicable:

### DIFF
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -560,7 +560,7 @@ impl Ui {
     }
 
     pub fn can_prompt() -> bool {
-        io::stdout().is_terminal()
+        io::stderr().is_terminal()
             || env::var("JJ_INTERACTIVE")
                 .map(|v| v == "1")
                 .unwrap_or(false)
@@ -573,8 +573,8 @@ impl Ui {
                 "Cannot prompt for input since the output is not connected to a terminal",
             ));
         }
-        write!(self.stdout(), "{prompt}: ")?;
-        self.stdout().flush()?;
+        write!(self.stderr(), "{prompt}: ")?;
+        self.stderr().flush()?;
         let mut buf = String::new();
         io::stdin().read_line(&mut buf)?;
 
@@ -601,7 +601,7 @@ impl Ui {
         if !Self::can_prompt() {
             if let Some(default) = default {
                 // Choose the default automatically without waiting.
-                writeln!(self.stdout(), "{prompt}: {default}")?;
+                writeln!(self.stderr(), "{prompt}: {default}")?;
                 return Ok(default.to_owned());
             }
         }
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -566,8 +566,6 @@ impl Ui {
                 .unwrap_or(false)
     }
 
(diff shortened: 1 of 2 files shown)


cli/src/ui.rs
```diff
@@ -580,16 +578,15 @@ impl Ui {
         let mut buf = String::new();
         io::stdin().read_line(&mut buf)?;
 
-        if let Some(trimmed) = buf.strip_suffix('\n') {
-            buf = trimmed.to_owned();
-        } else if buf.is_empty() {
-            return Err(io::Error::new(
+        let buf = buf.strip_suffix('\n').map(|s| s.to_owned()).unwrap_or(buf);
```
yuja, line 581: nit: `buf.truncate(trimmed.len())` could also be used.

btw, if `buf == "\n"`, an empty string should be returned. It's not an EOF.

cli/src/ui.rs
```diff
@@ -566,15 +566,15 @@ impl Ui {
                 .unwrap_or(false)
     }
 
-    pub fn prompt(&self, prompt: &str) -> io::Result<String> {
+    pub fn prompt(&self, writer: &mut dyn Write, prompt: &str) -> io::Result<String> {
         if !Self::can_prompt() {
```
yuja, line 570: Can you elaborate the intent? It's weird that `can_prompt()` tests `stdout`, but we use an arbitrary write stream.

If the goal was to handle the case where `stdout` is redirected, it might be okay to always use `stderr`. Maybe we can also use some library that can obtain tty even if stdout/err was redirected. I assume `rpassword` does something like that.

PhilipMetzger: I assume its for testing Google's custom binary. But I agree that its missing the intent in the commit message.

torquestomp: Updated the commit description to be clearer, and refactored the change to fix the disconnect between `can_prompt()` and the actual output channel.

The internal use case I'm trying to support here is running `jj` in a shell function environment where we need to parse `stdout` and can't have it polluted with other messaging.  It's not really possible to use some sort of escaping mechanism to work around the problem of `prompt` printing to stdout because then either the user doesn't see the prompt message (if we hide stdout), or else we print the parsed output the user isn't supposed to see in addition to the prompt message (if we don't hide stdout).

cli/src/ui.rs
```diff
@@ -581,15 +579,20 @@ impl Ui {
...
-            buf = trimmed.to_owned();
-        } else if buf.is_empty() {
-            return Err(io::Error::new(
+            buf.truncate(trimmed.len());
+            if buf.is_empty() {
+                return Ok(buf);
+            }
+        }
+
+        if !buf.is_empty() {
```
yuja, line 610: better to test `buf.is_empty()` before stripping

cli/src/ui.rs
yuja: Thanks. Maybe we can just always use `stderr` then? I don't think there is a reason we have to use `stdout`.

torquestomp: Sure, it's certainly an easier change.

yuja (approved):
Thanks!

PR #4539, opened 2024-09-26
cli: Introduce `write_labeled!` template formatter macro

Veykril:
Mainly an experiment to see if this looks more readable / uniform.

# Checklist

If applicable:

### DIFF
--- a/cli/src/command_error.rs
+++ b/cli/src/command_error.rs
@@ -65,6 +65,7 @@ use crate::merge_tools::MergeToolConfigError;
 use crate::revset_util::UserRevsetEvaluationError;
 use crate::template_parser::TemplateParseError;
 use crate::template_parser::TemplateParseErrorKind;
+use crate::templater::write_labeled;
 use crate::ui::Ui;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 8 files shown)


cli/src/commands/operation/diff.rs
```diff
@@ -384,14 +385,14 @@ fn write_modified_change_summary(
 ) -> Result<(), std::io::Error> {
     writeln!(formatter, "Change {}", short_change_hash(change_id))?;
     for commit in modified_change.added_commits.iter() {
-        formatter.with_label("diff", |formatter| write!(formatter.labeled("added"), "+"))?;
+        formatter.with_label("diff", |formatter| write_labeled!(formatter, "added", "+"))?;
```
Veykril, line 388: Interesting one being nested. Could make the macro accept something like
```suggestion
        write_labeled!(formatter, "diff" & "added", "+")?;
```

cli/src/command_error.rs
```diff
@@ -775,12 +776,12 @@ fn print_error_sources(ui: &Ui, source: Option<&dyn error::Error>) -> io::Result
     ui.stderr_formatter()
         .with_label("error_source", |formatter| {
             if err.source().is_none() {
-                write!(formatter.labeled("heading"), "Caused by: ")?;
+                write_labeled!(formatter, "heading", "Caused by: ")?;
```
yuja, line 779: I like the original code because it's clearer which string literal is a label.

cli/src/commands/operation/diff.rs
yuja: label string could be separated by whitespace (and split by `formatter.labeled()`), but I don't know if we like it.

PhilipMetzger:
I personally don't like the macro variant, as it doesn't improve the readability of the code.

Veykril:
Fair, I am indifferent to the change so mainly opened it to see opinions.

PR #4541, opened 2024-09-26
changelog: warning on `jj git fetch` is a new feature, not a bugfix

samueltardieu:
@essiene I think that the new warning you implemented is a new feature, it does not fix a bug in jj, it enhances its usage. What do you think?


essiene:
> @essiene I think that the new warning you implemented is a new feature, it does not fix a bug in jj, it enhances its usage. What do you think?

Hmmm... I guess it could fit into any of the two categories. I may have worded that changelog sentence ambigously. The original report was a "bug" because `jj` completed silently without notifying the user that the requested branch didn't exist.

I don't really have a strong opinion on which is more apt.

essiene (approved):

PR #1168, opened 2023-01-31
split: give warning when given paths don't match anything

samueltardieu:
# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [x] I have added tests to cover my changes

PR #4542, opened 2024-09-26
split: do not prevent all changes from going into the parent commit, and a fix

### LINKED ISSUE #4533: Split behavior differs when using diff editor and passing path
## Description

Just ran into the following scenario.

```
> jj new
> jj describe -m 'Exciting feature'
... Run into a bug.
... Fix the bug instead of working on the feature.
```

Ideally, what I want to do from here is "shove" the changes down one commit and specify a good commit message.

I expected `jj split` to be the tool for this but selecting all changes in the built-in diff editor quits the command with `Nothing changed.`

Sidenote: I understand that creation of an empty commit is probably usually undesirable but an empty commit with a commit message is not entirely empty. If this limitation is intentional, I hope there can be a flag to lift the restriction.

Two ways to achieve what I wanted came to mind.

```
> jj describe
... Copy the message to clipboard and edit
> jj new
> jj describe
... Paste clipboard
```

```
... Make a token change that will belong to the commit with the current message
> jj split
... Dig into the changes to find the token change
```

Neither of these feel like great UX. Luckily the [PATHS]... argument in the help for split caught my eye and I only had a single file with changes (manually listing a dozen files would be tedious). `jj split -- path/to/changed/file` to the rescue, adding all the changes to the first commit and allowing me to edit the message and copying the original message to the new empty commit for me.

In the end I got exactly what I wanted but the diff editor (especially the built-in one) making it impossible is unexpected and IMO a bug.

I did look for existing issues that express a desire to have split result in empty commits, #1155 is the closest I found and #3772 is somewhat similar but neither has quite the same expectations.

1. `jj split`
1. Select all changes in the built-in diff editor and confirm

## Expected Behavior

Add all the changes to the first commit and create a new empty commit, offering to edit the original message for both.

## Actual Behavior

Outputs `Nothing changed.`, doesn't create any new commits.

## Specifications

- Platform:
  ```
  > uname -a
  Linux yorp 6.6.44 #1-NixOS SMP PREEMPT_DYNAMIC Sat Aug  3 06:54:42 UTC 2024 x86_64 GNU/Linux
  ```

- Version:
  ```
  > jj --version
  jj 0.17.1
  ```

samueltardieu:
Allowing to put everything into the first commit closes #4533.

**Question:** should we suggest using `jj undo` if everything was put into one of the two commits in case the user didn't want to do this? Or, do we want this only if, in addition, the empty commit has an empty description, because there should be no reason to create an empty commit?

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj git fetch -b <remote-git-branch-name>` will now warn if the branch(es)
    can not be found in any of the specified/configured remotes.
 
+* `jj split` now lets the user select all changes in interactive mode. This may be used
+  to keeping all changes into the first commit while keeping the current commit
+  description for the second commit (the newly created empty one).
+
 ### Fixed bugs
 
 * Update working copy before reporting changes. This prevents errors during reporting
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -99,8 +99,7 @@ You are splitting a commit into two: {}
 The diff initially shows the changes in the commit you're splitting.
 
 Adjust the right side until it shows the contents you want for the first commit.
-The remainder will be in the second commit. If you don't make any changes, then
-the operation will be aborted.
+The remainder will be in the second commit.
 ",
             tx.format_commit_summary(&commit)
         )
@@ -109,12 +108,13 @@ the operation will be aborted.
     // Prompt the user to select the changes they want for the first commit.
     let selected_tree_id =
         diff_selector.select(&base_tree, &end_tree, matcher.as_ref(), format_instructions)?;
-    if &selected_tree_id == commit.tree_id() && diff_selector.is_interactive() {
+    if &selected_tree_id == commit.tree_id() {
         // The user selected everything from the original commit.
-        writeln!(ui.status(), "Nothing changed.")?;
-        return Ok(());
-    }
-    if selected_tree_id == base_tree.id() {
+        writeln!(
+            ui.warning_default(),
+            "All changes have been selected, so the second commit will be empty"
+        )?;
+    } else if selected_tree_id == base_tree.id() {
         // The user selected nothing, so the first commit will be empty.
         writeln!(
             ui.warning_default(),
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -118,8 +118,7 @@ the operation will be aborted.
         // The user selected nothing, so the first commit will be empty.
         writeln!(
             ui.warning_default(),
-            "The given paths do not match any file: {}",
-            args.paths.join(" ")
+            "No changes have been selected, so the first commit will be empty"
         )?;
     }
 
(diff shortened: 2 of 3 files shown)


martinvonz:
Should most of the PR description be moved into the commit description (with s/PR/commit/ or similar)? The question makes sense in the PR description.

samueltardieu:
> Should most of the PR description be moved into the commit description (with s/PR/commit/ or similar)? The question makes sense in the PR description.

It even led me to split the commits (using `split`, this is appropriate).

cli/tests/test_split_command.rs
```diff
@@ -129,14 +129,14 @@ fn test_split_by_paths() {
     test_env.set_up_fake_editor();
     let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["split", "-r", "@-", "nonexistent"]);
     insta::assert_snapshot!(stdout, @"");
-    insta::assert_snapshot!(stderr, @r###"
-    Warning: The given paths do not match any file: nonexistent
+    insta::assert_snapshot!(stderr, @r#"
+    Warning: No changes have been selected, the first commit will be empty
```
martinvonz, line 134: The current message was added by you in #1168. I wonder if you included it as a reminder to the user. Perhaps they did `jj split -r X *.rs` and the `*.rs` gets resolved by their shell and doesn't include a file that was added in X. But maybe that's a low-enough risk that we don't need to worry about it.

CHANGELOG.md
```diff
@@ -101,6 +101,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Editing a hidden commit now makes it visible.
 
+* `jj split` will no longer reference the paths given on the command line when
+  none have been given in interactive mode.
```
martinvonz, line 109: I don't think this needs to be mentioned. I can't think of a user group who would care to learn this. It also doesn't seem like it should be considered a fixed bug.

cli/src/commands/split.rs
```diff
@@ -109,12 +108,13 @@ the operation will be aborted.
...
-    if &selected_tree_id == commit.tree_id() && diff_selector.is_interactive() {
+    if &selected_tree_id == commit.tree_id() {
         // The user selected everything from the original commit.
-        writeln!(ui.status(), "Nothing changed.")?;
-        return Ok(());
-    }
-    if selected_tree_id == base_tree.id() {
+        writeln!(
+            ui.warning_default(),
+            "All changes have been selected, the second commit will be empty"
```
martinvonz, line 115: nit: maybe use semicolon instead of comma since these are two full sentences. Oh, but this matches the other message below, so never mind

cli/src/commands/split.rs
```diff
@@ -109,12 +108,13 @@ the operation will be aborted.
     // Prompt the user to select the changes they want for the first commit.
     let selected_tree_id =
         diff_selector.select(&base_tree, &end_tree, matcher.as_ref(), format_instructions)?;
-    if &selected_tree_id == commit.tree_id() && diff_selector.is_interactive() {
+    if &selected_tree_id == commit.tree_id() {
         // The user selected everything from the original commit.
```
martinvonz, line 112: nit: might be nice to add ", so the second commit will be empty" to match the comment from the case below

martinvonz (approved):

cli/tests/test_split_command.rs
samueltardieu: Yeah, I don't remember why I added this, it was a long time ago.

PR #4547, opened 2024-09-27
Add new command `jj config unset`

### LINKED ISSUE #4458: FR: `jj config unset [OPTIONS] <--user|--repo> <NAME>`
Removing a configuration currently requires editing the file manually, having this command makes the removal of configs more convenient.

Have a `jj config` subcommand that removes any set config by the given name.

None

**Additional context**
This would give parity with git's config command, which likewise offers an `unset` subcommand

pylbrecht:
Allow unsetting config values similar to `git config unset`.

```bash
$ jj config set --user some-key some-val
$ jj config get some-key
some-val
$ jj config unset --user some-key
$ jj config get some-key
Config error: configuration property "some-key" not found
For help, see https://martinvonz.github.io/jj/latest/config/.
```

Unsetting a key, which is part of a table, might leave that table empty. For now we do not delete such empty tables, as there may be cases where an empty table is semantically meaningful (https://github.com/martinvonz/jj/issues/4458#issuecomment-2407109269).

For example:

```toml
[table]
key = "value"

[another-table]
key = "value"
```

Running `jj config unset --user table.key` will leave us with `table` being empty:
```toml
[table]

[another-table]
key = "value"
```

---

Closes #4458

# Checklist
If applicable:

- [x] I have updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Add a new template alias `bultin_log_compact_full_description()`.
 
+* New command `jj config unset` that unsets config values. For example,
+  `jj config unset --user user.name`.
+
 ### Fixed bugs
 
 * Error on `trunk()` revset resolution is now handled gracefully.
--- a/cli/src/commands/config/mod.rs
+++ b/cli/src/commands/config/mod.rs
@@ -17,6 +17,7 @@ mod get;
 mod list;
 mod path;
 mod set;
+mod unset;
 
 use tracing::instrument;
 
@@ -30,6 +31,8 @@ use self::path::cmd_config_path;
 use self::path::ConfigPathArgs;
 use self::set::cmd_config_set;
 use self::set::ConfigSetArgs;
+use self::unset::cmd_config_unset;
+use self::unset::ConfigUnsetArgs;
 use crate::cli_util::CommandHelper;
 use crate::command_error::CommandError;
 use crate::config::ConfigSource;
@@ -82,6 +85,8 @@ pub(crate) enum ConfigCommand {
     Path(ConfigPathArgs),
     #[command(visible_alias("s"))]
     Set(ConfigSetArgs),
+    #[command(visible_alias("u"))]
+    Unset(ConfigUnsetArgs),
 }
 
 #[instrument(skip_all)]
@@ -96,5 +101,6 @@ pub(crate) fn cmd_config(
         ConfigCommand::List(args) => cmd_config_list(ui, command, args),
         ConfigCommand::Path(args) => cmd_config_path(ui, command, args),
         ConfigCommand::Set(args) => cmd_config_set(ui, command, args),
+        ConfigCommand::Unset(args) => cmd_config_unset(ui, command, args),
     }
 }
--- a/cli/src/commands/config/unset.rs
+++ b/cli/src/commands/config/unset.rs
@@ -1,0 +1,50 @@
+// Copyright 2024 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use tracing::instrument;
+
+use super::ConfigLevelArgs;
+use crate::cli_util::get_new_config_file_path;
+use crate::cli_util::CommandHelper;
+use crate::command_error::user_error;
+use crate::command_error::CommandError;
+use crate::config::remove_config_value_from_file;
+use crate::config::ConfigNamePathBuf;
+use crate::ui::Ui;
+
+/// Update config file to unset the given option.
+#[derive(clap::Args, Clone, Debug)]
+pub struct ConfigUnsetArgs {
+    #[arg(required = true)]
+    name: ConfigNamePathBuf,
+    #[command(flatten)]
+    level: ConfigLevelArgs,
+}
+
+#[instrument(skip_all)]
+pub fn cmd_config_unset(
+    _ui: &mut Ui,
+    command: &CommandHelper,
+    args: &ConfigUnsetArgs,
+) -> Result<(), CommandError> {
+    let config_path = get_new_config_file_path(&args.level.expect_source_kind(), command)?;
+    if config_path.is_dir() {
+        return Err(user_error(format!(
+            "Can't set config in path {path} (dirs not supported)",
+            path = config_path.display()
+        )));
+    }
+
+    remove_config_value_from_file(&args.name, &config_path)
+}
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -583,12 +583,7 @@ fn read_config_path(config_path: &Path) -> Result<config::Config, config::Config
         .build()
 }
 
-pub fn write_config_value_to_file(
-    key: &ConfigNamePathBuf,
-    value: toml_edit::Value,
-    path: &Path,
-) -> Result<(), CommandError> {
-    // Read config
+fn read_config(path: &Path) -> Result<toml_edit::Document, CommandError> {
     let config_toml = std::fs::read_to_string(path).or_else(|err| {
         match err.kind() {
             // If config doesn't exist yet, read as empty and we'll write one.
@@ -599,12 +594,29 @@ pub fn write_config_value_to_file(
             )),
         }
     })?;
-    let mut doc: toml_edit::Document = config_toml.parse().map_err(|err| {
+    config_toml.parse().map_err(|err| {
         user_error_with_message(
             format!("Failed to parse file {path}", path = path.display()),
             err,
         )
-    })?;
+    })
+}
+
+fn write_config(path: &Path, doc: &toml_edit::Document) -> Result<(), CommandError> {
+    std::fs::write(path, doc.to_string()).map_err(|err| {
+        user_error_with_message(
+            format!("Failed to write file {path}", path = path.display()),
+            err,
+        )
+    })
+}
+
+pub fn write_config_value_to_file(
+    key: &ConfigNamePathBuf,
+    value: toml_edit::Value,
+    path: &Path,
+) -> Result<(), CommandError> {
+    let mut doc = read_config(path)?;
 
     // Apply config value
     let mut target_table = doc.as_table_mut();
@@ -633,13 +645,7 @@ pub fn write_config_value_to_file(
     }
     target_table[last_key_part] = toml_edit::Item::Value(value);
 
-    // Write config back
-    std::fs::write(path, doc.to_string()).map_err(|err| {
-        user_error_with_message(
-            format!("Failed to write file {path}", path = path.display()),
-            err,
-        )
-    })
+    write_config(path, &doc)
 }
 
 /// Command name and arguments specified by config.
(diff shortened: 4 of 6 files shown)


cli/src/commands/config/unset.rs
```diff
@@ -0,0 +1,36 @@
...
+) -> Result<(), CommandError> {
+    let config_path = get_new_config_file_path(&args.level.expect_source_kind(), command)?;
+    if config_path.is_dir() {
+        return Err(user_error(format!(
+            "Can't set config in path {path} (dirs not supported)",
+            path = config_path.display()
+        )));
+    }
+
+    // TODO(pylbrecht): do we need to check_wc_author() here?
```
pylbrecht, line 33: I saw we have this check here:
https://github.com/martinvonz/jj/blob/f4ec653e12ddaeb39c98124dad7d2de5845321d2/cli/src/commands/config/set.rs#L66-L72

Do we need to inform the user in the same way for `jj config unset`?

cli/src/commands/config/unset.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use super::ConfigLevelArgs;
+use crate::cli_util::get_new_config_file_path;
+use crate::cli_util::CommandHelper;
+use crate::command_error::user_error;
+use crate::command_error::CommandError;
+use crate::config::remove_config_value_from_file;
+use crate::config::ConfigNamePathBuf;
+
+/// Update config file to unset the given option.
```
pylbrecht, line 11: Should we document for users that unsetting keys, which are part of a table, might result in empty tables?

cli/tests/test_config_command.rs
```diff
@@ -602,6 +602,96 @@ fn test_config_set_nontable_parent() {
...
+            "set",
+            "--user",
+            "table.inline-table",
+            "{ foo = true }",
+        ],
+    );
+    test_env.jj_cmd_ok(
+        &repo_path,
+        &["config", "unset", "--user", "table.inline-table.foo"],
+    );
```
pylbrecht, line 661: I am unsure how extensive we want to be in terms of coverage here. I am fine with adding more cases, if you feel it's beneficial.

cli/tests/test_config_command.rs
```diff
@@ -602,6 +602,96 @@ fn test_config_set_nontable_parent() {
...
+fn test_config_unset_for_repo() {
+    let test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    test_env.jj_cmd_ok(
+        &repo_path,
+        &["config", "set", "--repo", "test-key", "test-val"],
+    );
+    test_env.jj_cmd_ok(&repo_path, &["config", "unset", "--repo", "test-key"]);
```
pylbrecht, line 682: I just covered the very simple happy path for `--repo`, since the implementation covers both `--user` and `--repo`. However, I don't mind copy/pasting the cases from `test_config_unset_for_user()` if that is beneficial.

(I am not sure how likely it is that the implementations for `--user` and `--repo` diverge in the future, such that having extensive tests for both becomes necessary.)

pylbrecht:
I would like to get some feedback on my changes. I left some open questions as inline comments.
Also, since writing Rust still feels pretty wonky, I am happy for any Rust specific feedback.

cli/src/commands/config/mod.rs
```diff
@@ -17,8 +17,11 @@ mod get;
 mod list;
 mod path;
 mod set;
+mod unset;
 
 use tracing::instrument;
+use unset::cmd_config_unset;
+use unset::ConfigUnsetArgs;
```
yuja, line 24: nit: `self::unset` for consistency (and clarity)

cli/src/commands/config/unset.rs
yuja: I don't think explicit doc is needed.

cli/src/commands/config/unset.rs
yuja: Maybe no. It's technically correct that unsetting repo config may change the user name/email, but I expect such situation is rare.

cli/src/config.rs
```diff
@@ -583,6 +584,58 @@ fn read_config_path(config_path: &Path) -> Result<config::Config, config::Config
...
+                err,
+            )),
+        }
+    })?;
+    let mut doc: toml_edit::Document = config_toml.parse().map_err(|err| {
+        user_error_with_message(
+            format!("Failed to parse file {path}", path = path.display()),
+            err,
+        )
+    })?;
```
yuja, line 607: nit: Perhaps, we can extract read/write helper functions or a function that takes callback `update_doc: impl FnOnce(&mut Document) -> Result<..>`.

cli/src/config.rs
```diff
@@ -583,6 +584,58 @@ fn read_config_path(config_path: &Path) -> Result<config::Config, config::Config
...
+    let mut key_iter = key.components();
+    let last_key = key_iter.next_back().expect("key must not be empty");
+    let root: &mut dyn TableLike = doc.as_table_mut();
+    let target_table = key_iter.try_fold(root, |table, key| {
+        table
+            .get_mut(key)
+            .ok_or(config::ConfigError::NotFound(key.to_string()))
+            .and_then(|table| {
+                table
+                    .as_table_like_mut()
```
yuja, line 618: nit: better to use `.as_table_mut()` for consistency. Think inline table as a single value.

cli/tests/test_config_command.rs
yuja: I think these tests are good enough. CLI tests are slow.

cli/tests/test_config_command.rs
```diff
@@ -602,6 +602,96 @@ fn test_config_set_nontable_parent() {
...
+            "{ foo = true }",
+        ],
+    );
+    test_env.jj_cmd_ok(
+        &repo_path,
+        &["config", "unset", "--user", "table.inline-table.foo"],
+    );
+
+    let user_config_toml = std::fs::read_to_string(&user_config_path)
+        .unwrap_or_else(|_| panic!("Failed to read file {}", user_config_path.display()));
```
yuja, line 664: nit: just do `.unwrap()` in tests unless there's a special consideration

yuja (commented):
Looks generally good to me, thanks.

pylbrecht:
@yuja, thanks for your quick feedback!

I will address your comments over the next couple of days and let you know when I'm done.

cli/src/commands/config/mod.rs
pylbrecht: Thanks, my LSP took care of that auto-import and I wasn't aware that it didn't do the right thing here.

cli/tests/test_config_command.rs
pylbrecht: I just copy/pasted that from another test in the same file and haven't really thought about it.
Now that I put some thought into it, I guess the "custom" panic here serves a better error message than a plain `.unwrap()`, so I am inclined to leave it as is.

`.unwrap_or_else(...)`
```
---- test_config_command::test_config_unset_for_user stdout ----
thread 'test_config_command::test_config_unset_for_user' panicked at cli/tests/test_config_command.rs:665:29:
Failed to read file /private/var/folders/hb/zr1tp04d29b8wdh14lczb3c40000gn/T/jj-test-IkBUbD/config/config.toml
```

vs.

`.unwrap()`
```
---- test_config_command::test_config_unset_for_user stdout ----
thread 'test_config_command::test_config_unset_for_user' panicked at cli/tests/test_config_command.rs:664:64:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```
---
Another option would be to use `.expect()`
```suggestion
        .expect(&format!("Failed to read file {}", user_config_path.display()));
```
I don't know if that is "better", because of `&format!(...)`. What's your take on this?

yuja: Well, if `std::fs::read_to_string(&user_config_path)` is likely to fail, it makes sense to provide a better error message. However, we're testing "config unset" (which isn't the command to create new config file), so I would just use `.unwrap()`.

pylbrecht: Thanks for elaborating. That makes a lot of sense now, since I copied that part from a `jj config set` test case. 💡 
I will replace it with `.unwrap()` then.

cli/src/config.rs
pylbrecht: Good point, thanks!

Should we give people a hint in the error message if they try to unset a key of an inline table then?
Or is `"{key} is not a table"` good enough?

(Maybe I am overthinking this.)

cli/tests/test_config_command.rs
joyously: Do you test unsetting it twice in a row? What would be the expected result for the second time?

pylbrecht: > Do you test unsetting it twice in a row? What would be the expected result for the second time?

Right now there is no test case, which covers running `jj config unset` twice in a row for the same key.

With the current implementation, given this user config:
```toml
[table]
foo = "bar"
```
..running `jj config --user unset table.foo` twice in a row will result in the following error:
```
Config error: configuration property "table.foo" not found
For help, see https://martinvonz.github.io/jj/latest/config/.
```
..because the first time running `jj config --user unset table.foo` removes `table.foo`. So the second execution can't find the key any longer.

(The same applies for `jj config --repo unset table.foo`)

---

@joyously, your question got me thinking. I added a comment in the issue: https://github.com/martinvonz/jj/issues/4458#issuecomment-2438613512

yuja: I don't know whether it should be an error or a warning, but it's nice that `jj config unset` reports unknown key to unset (i.e. noop change).

cli/src/config.rs
pylbrecht: I really liked the idea of an `update_doc` callback, but it turned out to be a little bumpy. `write_config_value_to_file()` takes a `value` argument, whereas `remove_config_value_from_file()` does not. I tried making `value` optional, but that made the interface awkward.  

That's when I decided to go for `read_config()` and `write_config()` helpers, despite not being too happy about it. 

Since I do not consider myself an advanced Rustacean, feel free to give it a shot yourself; I might be missing something.

pylbrecht: I just realized I baked the refactoring and the feature into one change. I will split these into two separate ones.

cli/src/config.rs
pylbrecht: Great, happy to leave the `"{key} is not a table"`. In case people get confused, we can special-case it then.

pylbrecht:
@yuja, thanks again for taking the time to look through my changes. I addressed your comments, so the changes are ready for another review. For your sake, I hope this is the last round trip. 🤞 

I also added a test case for unsetting keys of inline tables, making it more explicit that we do not support this at the moment.
(When it comes to tests, I am biased towards "better too much than too little", so let me know if the overhead of another test is not worth it in this case.)

CHANGELOG.md
```diff
@@ -78,6 +78,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Add a new template alias `bultin_log_compact_full_description()`.
 
+* New command `jj config unset` that unsets config values. For example,
+  `jj config unset --user user.name`.
+
+
```
yuja, line 84: nit: one blank line, not two

cli/src/commands/config/unset.rs
yuja, line 1: nit: can you add "Copyright 2024 ..." comment to this file?

cli/src/config.rs
```diff
@@ -633,13 +645,35 @@ pub fn write_config_value_to_file(
...
+    // Find target table
+    let mut key_iter = key.components();
+    let last_key = key_iter.next_back().expect("key must not be empty");
+    let target_table = key_iter.try_fold(doc.as_table_mut(), |table, key| {
+        table
+            .get_mut(key)
+            .ok_or_else(|| config::ConfigError::NotFound(key.to_string()))
+            .and_then(|table| {
+                table.as_table_mut().ok_or_else(|| {
+                    config::ConfigError::Message(format!("\"{key}\" is not a table",))
```
yuja, line 666: nit: remove trailing comma

(at your option) we can use `r#".."` to remove `\"` escapes.

yuja (approved):
LGTM, thanks.

PR #4551, opened 2024-09-29
cli `new`: Allow `--before/--after` without a value to default to `@`

ilyagr:
Without this, I find it a bit jarring that `jj new` works but `jj new --before` does not. By contrast, since `jj rebase` does not currently work, I don't think `jj rebase --before` should either.

Note that `jj new --before @ another_revision` is invalid, so `jj new --before another_revision` can only be parsed correctly in one way. I am slightly concerned that `clap` might forbids this in the future even in the cases where a human can tell there is no ambiguity, but I'm hoping for the best.

# Checklist

If applicable:


martinvonz:
> Without this, I find it a bit jarring that jj new works but jj new --before does not.

I would not expect `jj new --before` to work. Maybe it's just because I'm used to Mercurial's argument parsing. Either way, I didn't think we have and existing flags with optional values, so I think we should think carefully before we add one. I assume it will only be possible to do it for commands that don't take any positional arguments, which limits the commands where we can use it even if we wanted to.

ilyagr:
By the way, `jj new --before` used to work before we added the ability to use both `--before` and `--after` in the same command, since it was a boolean flag back then. I used that often, and I still haven't gotten used to it going away. I don't know whether I was the only one.

I do not propose we use these very often, but `jj new` seems pretty safe to me as I explained above and as the tests show. Still, I can understand it if people are worried, and perhaps I will get used to it eventually.

algmyr:
I'm guessing this is way obsolete now and you got used to things? Closing as part of some janitorial work.

ilyagr:
I still miss not having this, but I mostly gave up on it. It's a pretty minor issue.

In terms of getting used to it, the pangs of annoyance became a reminder to type "space @" :)

PR #4557, opened 2024-10-01
revset: fix crash on "log --at-op 00000000 -r 'root()'"

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_revset.rs
+++ b/lib/tests/test_revset.rs
@@ -34,6 +34,7 @@ use jj_lib::op_store::RefTarget;
 use jj_lib::op_store::RemoteRef;
 use jj_lib::op_store::RemoteRefState;
 use jj_lib::op_store::WorkspaceId;
+use jj_lib::operation::Operation;
 use jj_lib::repo::Repo;
 use jj_lib::repo_path::RepoPath;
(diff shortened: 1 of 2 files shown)


lib/tests/test_revset.rs
```diff
@@ -954,6 +963,14 @@ fn test_evaluate_expression_root_and_checkout() {
         vec![root_commit.id().clone()]
     );
 
+    // but not in the root operation. It might be okay to pretend that the root
+    // commit exists in the root operation, but queries like "root()" shouldn't
+    // panic in any case.
```
martinvonz, line 968: Yes, seems to me like it the root operation should have the root commit. I guess that would mean that the op store would have to learn about the root commit id, which I don't think it currently knows about.

martinvonz (approved):
Fun case

lib/tests/test_revset.rs
martinvonz: But, to be clear, I'm not asking for any changes to this PR

yuja: Yes, it's a bit odd that `OpStore` knows about the root commit. If we had an abstraction layer on top of the backend `OpStore`, an empty view might be handled there.

Issue #4509, opened 2024-09-20
`jj untrack foo~` fails


0-wiz-0:
## Description

emacs writes backup files with a `~` (tilde) at the end. jj automatically picks them up if you forgot to ignore them in the `.gitignore` file.
I can't untrack them in the latest version any longer. (I think this worked when it was `jj untrack`, but now it doesn't with `jj file untrack`, but perhaps I misremember.)

## Steps to Reproduce the Problem

1. jj git init
1. touch foo~
1. jj
1. jj file untrack foo~
```
jj file untrack foo\~
Error: Failed to parse fileset: Syntax error
Caused by:  --> 1:5
  |
1 | foo~
  |     ^---
  |
  = expected `~` or <primary>
```

## Specifications

- Platform: NetBSD/10.99.12/amd64
- Version: 0.21.0

yuja:
Since fileset is enabled by default, `~` has to be quoted `'"foo~"'` or `'glob:"*~"'`.
https://martinvonz.github.io/jj/latest/filesets/

Maybe we can add a hint if the expression looks like an existing path.

0-wiz-0:
From a quick read of the linked page: `~` seems only to be supported as a separate word, or at the start of a word, so perhaps making that an explicit requirement would also solve this.

yuja:
Yeah, trailing `~` could be special cased, but I don't think it's worth the added oddity. `*~` is usually a backup file and so often ignored by `.gitignore` or user ignore.

`x~y` and `~x` is a valid syntax, btw.

samueltardieu:
The message is now:

```
Error: Failed to parse fileset: Syntax error
Caused by:  --> 1:5
  |
1 | foo~
  |     ^---
  |
  = expected `~` or <primary>
Hint: See https://martinvonz.github.io/jj/latest/filesets/ for filesets syntax, or for how to match file paths.
```

Hopefully this is a clearer message. If this is, might we close this issue?

0-wiz-0:
I accept that I'm the only one annoyed by the change from 'jj untrack' to 'jj file untrack' and have worked around it by having my global git config ignore "*~" files.

As for the actual change - I think this needs improvements in the documentation.
I read `expected `~` or <primary>` as that `jj` would expect a second tilde or a <primary> after the first tilde.
Both alternatives are unclear to me.
[...]

PR #4562, opened 2024-10-02
cli: add pointer to the documentation in case of fileset syntax error

samueltardieu:
Inspired by #4509.

# Checklist

If applicable:

### DIFF
--- a/cli/src/command_error.rs
+++ b/cli/src/command_error.rs
@@ -613,6 +613,10 @@ fn file_pattern_parse_error_hint(err: &FilePatternParseError) -> Option<String> 
 
 fn fileset_parse_error_hint(err: &FilesetParseError) -> Option<String> {
     match err.kind() {
+        FilesetParseErrorKind::SyntaxError => Some(String::from(
+            "See https://martinvonz.github.io/jj/latest/filesets/ for filesets syntax, or for how \
+             to match file paths.",
+        )),
         FilesetParseErrorKind::NoSuchFunction {
             name: _,
             candidates,
@@ -620,7 +624,6 @@ fn fileset_parse_error_hint(err: &FilesetParseError) -> Option<String> {
         FilesetParseErrorKind::InvalidArguments { .. } | FilesetParseErrorKind::Expression(_) => {
             find_source_parse_error_hint(&err)
         }
-        _ => None,
     }
 }
 
--- a/cli/tests/test_global_opts.rs
+++ b/cli/tests/test_global_opts.rs
@@ -292,6 +292,31 @@ fn test_bad_path() {
 }
 
 #[test]
+fn test_invalid_filesets_looking_like_filepaths() {
+    let test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    let stderr = test_env.jj_cmd_failure(&repo_path, &["file", "show", "abc~"]);
+    insta::assert_snapshot!(stderr, @r#"
+    Error: Failed to parse fileset: Syntax error
+    Caused by:  --> 1:5
+      |
+    1 | abc~
+      |     ^---
+      |
+      = expected `~` or <primary>
+    Hint: See https://martinvonz.github.io/jj/latest/filesets/ for filesets syntax, or for how to match file paths.
+    "#);
+
+    test_env.add_config(r#"ui.allow-filesets=false"#);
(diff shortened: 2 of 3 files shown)


cli/src/cli_util.rs
```diff
@@ -1093,17 +1093,37 @@ impl WorkspaceCommandHelper {
         if values.is_empty() {
             Ok(FilesetExpression::all())
         } else if self.settings().config().get_bool("ui.allow-filesets")? {
-            self.parse_union_filesets(ui, values)
+            self.parse_union_filesets(ui, values).or_else(|e| {
+                // If the arguments could not be parsed as valid filesets but could still be
+                // valid paths, warn the user about it.
+                if self.parse_file_path_expressions(values).is_ok() {
```
yuja, line 1099: I don't think parsing as a file path would help reduce false positives. Many fileset expressions are valid file paths.

Instead, maybe we can just attach hint to `FilesetParseError` unconditionally (in command_error.rs)?

samueltardieu: Using a hint is a good idea, I've switched the code to use them, thanks for the suggestion.

For the rest: attaching to the hint in `command_error.rs` without checking that the arguments look like valid file paths is risky. For example, using "../foobar" as a fileset when we are at the root of the directory gives a `FilesetParserError` but it makes no sense to attach this hint, as the problem is that the file is out of the repository. This is why checking that file paths look valid using `self.parse_file_path_expressions(values)` is important to reduce the number of false positive for attaching this hint.

With this change, we get:

```shell
$ jj file show ../foobar
Error: Failed to parse fileset: Invalid file pattern
Caused by:
1:  --> 1:1
  |
1 | ../foobar
  | ^-------^
  |
  = Invalid file pattern
2: Path "../foobar" is not in the repo "."
3: Invalid component ".." in repo-relative path "../foobar"
```

If I attach the hint inconditionally, I get:

```shell
$jj file show ../foobar
Error: Failed to parse fileset: Invalid file pattern
Caused by:
1:  --> 1:1
  |
1 | ../foobar
  | ^-------^
  |
  = Invalid file pattern
2: Path "../foobar" is not in the repo "."
3: Invalid component ".." in repo-relative path "../foobar"
Hint: Filesets are enabled by default. If you want to match file paths containing characters used in fileset expressions (':', or '~'), escape them, or prefix them with "glob:".
```

yuja: Ah, yes. Perhaps, the hint should be added only to `SyntaxError`?

fwiw, there are more metacharacters in addition to `:` and `~`. It might be better to show a doc link https://martinvonz.github.io/jj/latest/filesets/ instead.

samueltardieu: I've updated the message to point to the documentation.

Also @yuja, when you mean attaching the hint to `SyntaxError`, would you prefer the implementation in https://github.com/martinvonz/jj/commit/dd70bd34193bf7b9a1eaef40e98dfd76968ef2b8?

yuja: I meant the hint can be added by `command_error::fileset_parse_error_hint()` or `impl From<FilesetParseError> for CommandError`. I assume it wouldn't be important for user to know whether the fileset is enabled by default or not.

Maybe we can also add a similar hint to revset, to mention that a tag or branch name might need quoting, see this URL (or `jj help revset` or something) for the syntax...

samueltardieu: I see. It makes the code much more compact indeed. I haven't touched revsets though, this would have to be done in another PR.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* Commands accepting filesets hint to look at the documentation in presence of
+  a syntax error.
```
yuja, line 23: nit: delete changelog entry? I don't think users have to know the error message is improved.

cli/src/command_error.rs
```diff
@@ -620,7 +620,10 @@ fn fileset_parse_error_hint(err: &FilesetParseError) -> Option<String> {
         FilesetParseErrorKind::InvalidArguments { .. } | FilesetParseErrorKind::Expression(_) => {
             find_source_parse_error_hint(&err)
         }
-        _ => None,
+        FilesetParseErrorKind::SyntaxError => Some(String::from(
+            "See https://martinvonz.github.io/jj/latest/filesets/ for filesets syntax, or for how \
+             to match file paths.",
+        )),
```
yuja, line 626: nit: can you sort them in definition order (i.e. insert `SyntaxError` first)?

yuja (approved):
Thanks!

PR #4563, opened 2024-10-02
diff: add machinery to ignore whitespace changes

yuja:
(Not meant for inclusion in the October release)

# Checklist

If applicable:

### DIFF
--- a/lib/src/diff.rs
+++ b/lib/src/diff.rs
@@ -16,7 +16,6 @@
 
 use std::cmp::Ordering;
 use std::collections::BTreeMap;
-use std::collections::HashMap;
 use std::hash::BuildHasher;
 use std::hash::Hash;
 use std::hash::Hasher;
@@ -26,6 +25,7 @@ use std::ops::Range;
 use std::slice;
 
 use bstr::BStr;
+use hashbrown::HashTable;
 use itertools::Itertools;
 
 pub fn find_line_ranges(text: &[u8]) -> Vec<Range<usize>> {
@@ -189,7 +189,6 @@ struct WordComparator<C, S> {
     hash_builder: S,
 }
 
-#[allow(unused)] // TODO
 impl<C: CompareBytes> WordComparator<C, RandomState> {
     fn new(compare: C) -> Self {
         WordComparator {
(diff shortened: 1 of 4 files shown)


lib/src/diff.rs
```diff
@@ -71,6 +75,73 @@ pub fn find_nonword_ranges(text: &[u8]) -> Vec<Range<usize>> {
         .collect()
 }
 
+/// Compares byte sequences based on a certain equivalence property.
+pub trait CompareBytes {
```
PhilipMetzger, line 111: Nit: I would add a comment here, _why_ we are generic instead of a wrapper type or a taking a callback. So noone sweeps in trying to "improve" it.

PhilipMetzger (approved):
LGTM, but @martinvonz  should also take a look. 

> This could be implemented as a newtype `Wrapper<'a>(&'a [u8])`, but a lifetime
>of the wrap function couldn't be specified correctly:
>
 > fn diff(left: &[u8], right: &[u8], wrap_fn: F, ..)
  >where
    F: for<'a> Fn(&'a [u8]) -> W<'a>, // F::Output<'a> can't be specified
    W: Copy + Eq + Hash
>
>If the wrapper were of `&Wrapper([u8])` type, `Fn(&[u8]) -> &W` works. However,
>it means we can no longer set comparison parameter (such as Regex) dynamically. Another idea is to add some filter function of `Fn(&[u8]) -> Cow<'_, [u8]>`
type, but I don't think we would want to pay the allocation cost in
hashing/comparison code. `Fn(&[u8]) -> impl Iterator<Item = &[u8]>` might work,
but it would be equally complex.

It seems like being generic, is the only way which could potentially avoid allocations as both other variants fall short of Rust being Rust.

lib/src/diff.rs
yuja: Good point. Expanded doc comment.

lib/src/diff.rs
```diff
@@ -71,6 +75,83 @@ pub fn find_nonword_ranges(text: &[u8]) -> Vec<Range<usize>> {
         .collect()
 }
 
+/// Compares byte sequences based on a certain equivalence property.
+///
+/// This isn't a newtype `Wrapper<'a>(&'a [u8])` but an external comparison
+/// object for the following reasons:
+///
+/// a. If it were newtype, a generic `wrap` function would be need. It couldn't
```
PhilipMetzger, line 115: s/need/needed

PhilipMetzger (commented):
Minor typo

martinvonz (approved):

PR #4566, opened 2024-10-02
doc: add comment to explain the syntax of folded examples

samueltardieu:
# Checklist

If applicable:

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -78,6 +78,9 @@ only symbols.
 You can use parentheses to control evaluation order, such as `(x & y) | z` or
 `x & (y | z)`.
 
+<!-- The following format will be understood by the web site generator, and will
+ generate a folded section that can be unfolded at will. -->
+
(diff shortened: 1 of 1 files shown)


docs/revsets.md
```diff
@@ -78,96 +78,96 @@ only symbols.
 You can use parentheses to control evaluation order, such as `(x & y) | z` or
 `x & (y | z)`.
 
-??? examples
+### Examples
```
martinvonz, line 81: I think the current form is a compromise to make it render well on the web page (using MkDocs, iirc). I think I may have sent a similar PR once. Maybe we need to add a comment here so people don't waste time trying to "fix" it.

samueltardieu: Good idea. I'll propose a comment instead.

samueltardieu: Done.

martinvonz (approved):

PR #4570, opened 2024-10-02
Enforce selected clippy lints, and small style fixes

samueltardieu:
Right now, we only use the basic Clippy lints. While many software opt to use `clippy::pedantic`, with sometimes individually disabled restrictions, this would be a bit extreme to do in one step (or to do at all, some lints might be intrusive or opiniated).

This PR proposes the activation of selected Clippy lints. Those I have tried so far are:
- `explicit_iter_loop`: prefer iterating over `&collection` rather than `collection.iter()`, this is more compact and idiomatic
- `flat_map_option`: using `.flat_map(f)` when `f` returns an `Option` is equivalent to using `.filter_map(f)`, except that the latter better underlines the filtering aspect
- `implicit_clone`: using `.to_vec()` on a `Vec`, or `.to_owned()` on an owned object can be replaced by `.clone()`, which makes it clearer that we already have something of the right type; at least one `.clone()` call is even superfluous after doing this transformation
- `needless_for_each`: for simple loops (one occurrence currently in the code), using an explicit `for` loop is clearer than using `.for_each()`
- ~~`redundant_else`: we learn very early that using `else` after a block which returns early is useless. This might be the most contentious lint in this list, as it changes the style a bit.~~ *(removed after discussion, see below)*
- `semicolon_if_nothing_returned`: using `;` at the end of expressions returning `()` when we do not explicitly need this unit value better expresses that we have a statement, not just an expression. It increases style consistency.
- `unlined_format_args`: inline variable names in `{}` when interpolating, also for consistency as this is done in 90%+ of the code.

Other lints might be of interest of course, but I gave it a first run in order to test the idea, and find some lints that other developers might feel comfortable with.

If I'm not wrong, all those lints are available in the stable compiler used to run Clippy in the CI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,6 +134,7 @@ testutils = { path = "lib/testutils" }
 
 [workspace.lints.clippy]
 explicit_iter_loop = "warn"
+flat_map_option = "warn"
 implicit_clone = "warn"
 semicolon_if_nothing_returned = "warn"
 uninlined_format_args = "warn"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,7 +134,8 @@ testutils = { path = "lib/testutils" }
 
 [workspace.lints.clippy]
 explicit_iter_loop = "warn"
-
+uninlined_format_args = "warn"
+ 
 # Insta suggests compiling these packages in opt mode for faster testing.
 # See https://docs.rs/insta/latest/insta/#optional-faster-runs.
 [profile.dev.package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -136,6 +136,7 @@ testutils = { path = "lib/testutils" }
 explicit_iter_loop = "warn"
 flat_map_option = "warn"
 implicit_clone = "warn"
+needless_for_each = "warn"
 semicolon_if_nothing_returned = "warn"
 uninlined_format_args = "warn"
  
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -132,6 +132,8 @@ jj-lib = { path = "lib", version = "0.22.0" }
 jj-lib-proc-macros = { path = "lib/proc-macros", version = "0.22.0" }
 testutils = { path = "lib/testutils" }
 
+[workspace.lints.clippy]
+
 # Insta suggests compiling these packages in opt mode for faster testing.
 # See https://docs.rs/insta/latest/insta/#optional-faster-runs.
 [profile.dev.package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,6 +134,7 @@ testutils = { path = "lib/testutils" }
 
 [workspace.lints.clippy]
 explicit_iter_loop = "warn"
+semicolon_if_nothing_returned = "warn"
 uninlined_format_args = "warn"
  
 # Insta suggests compiling these packages in opt mode for faster testing.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,6 +133,7 @@ jj-lib-proc-macros = { path = "lib/proc-macros", version = "0.22.0" }
 testutils = { path = "lib/testutils" }
 
 [workspace.lints.clippy]
+explicit_iter_loop = "warn"
 
 # Insta suggests compiling these packages in opt mode for faster testing.
 # See https://docs.rs/insta/latest/insta/#optional-faster-runs.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,6 +134,7 @@ testutils = { path = "lib/testutils" }
 
 [workspace.lints.clippy]
 explicit_iter_loop = "warn"
+implicit_clone = "warn"
 semicolon_if_nothing_returned = "warn"
 uninlined_format_args = "warn"
  
--- a/cli/src/command_error.rs
+++ b/cli/src/command_error.rs
@@ -352,7 +352,7 @@ impl From<SnapshotError> for CommandError {
                         size_diff, max_size.0, max_size,
                     )
                 } else {
-                    format!("it is {}; the maximum size allowed is ~{}.", size, max_size,)
+                    format!("it is {size}; the maximum size allowed is ~{max_size}.")
                 };
 
                 user_error(format!(
--- a/cli/tests/test_operations.rs
+++ b/cli/tests/test_operations.rs
@@ -738,7 +738,7 @@ fn test_op_recover_from_bad_gc() {
     let repo_path = test_env.env_root().join("repo");
     let git_object_path = |hex: &str| {
         let (shard, file_name) = hex.split_at(2);
-        let mut file_path = repo_path.to_owned();
+        let mut file_path = repo_path.clone();
         file_path.extend([".git", "objects", shard, file_name]);
         file_path
     };
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -956,8 +956,7 @@ impl Backend for GitBackend {
             .try_into_blob()
             .map_err(|err| to_read_object_err(err, id))?;
         let target = String::from_utf8(blob.take_data())
-            .map_err(|err| to_invalid_utf8_err(err.utf8_error(), id))?
-            .to_owned();
+            .map_err(|err| to_invalid_utf8_err(err.utf8_error(), id))?;
         Ok(target)
     }
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -117,3 +117,6 @@ watchman = ["jj-lib/watchman"]
 # The archive name is jj, not jj-cli. Also, `cargo binstall` gets
 # confused by the `v` before versions in archive name.
 pkg-url = "{ repo }/releases/download/v{ version }/jj-v{ version }-{ target }.{ archive-format }"
+
+[lints]
+workspace = true
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -33,9 +33,9 @@ fn main() {
     println!("cargo:rerun-if-env-changed=NIX_JJ_GIT_HASH");
 
     if let Some(git_hash) = get_git_hash() {
-        println!("cargo:rustc-env=JJ_VERSION={}-{}", version, git_hash);
+        println!("cargo:rustc-env=JJ_VERSION={version}-{git_hash}");
     } else {
-        println!("cargo:rustc-env=JJ_VERSION={}", version);
+        println!("cargo:rustc-env=JJ_VERSION={version}");
     }
 }
 
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -19,7 +19,7 @@ use std::str;
 const GIT_HEAD_PATH: &str = "../.git/HEAD";
 const JJ_OP_HEADS_PATH: &str = "../.jj/repo/op_heads/heads";
 
-fn main() -> std::io::Result<()> {
+fn main() {
     let version = std::env::var("CARGO_PKG_VERSION").unwrap();
 
     if Path::new(GIT_HEAD_PATH).exists() {
@@ -37,8 +37,6 @@ fn main() -> std::io::Result<()> {
     } else {
         println!("cargo:rustc-env=JJ_VERSION={}", version);
     }
-
-    Ok(())
 }
 
 fn get_git_hash() -> Option<String> {
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -254,7 +254,7 @@ impl TracingSubscription {
             .modify(|filter| {
                 *filter = tracing_subscriber::EnvFilter::builder()
                     .with_default_directive(tracing::metadata::LevelFilter::DEBUG.into())
-                    .from_env_lossy()
+                    .from_env_lossy();
             })
             .map_err(|err| internal_error_with_message("failed to enable debug logging", err))?;
         tracing::info!("debug logging enabled");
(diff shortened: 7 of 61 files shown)


ilyagr:
This is not necessarily related to this PR, but `cargo +nightly clippy` currently has numerous errors of the type  https://rust-lang.github.io/rust-clippy/master/index.html#too_long_first_doc_paragraph . I ended up using `cargo cranky` to ignore it, but if you feel like fixing some lints :)

lib/src/file_util.rs
```diff
@@ -92,13 +92,13 @@ pub fn relative_path(from: &Path, to: &Path) -> PathBuf {
     // Find common prefix.
     for (i, base) in from.ancestors().enumerate() {
         if let Ok(suffix) = to.strip_prefix(base) {
-            if i == 0 && suffix.as_os_str().is_empty() {
-                return ".".into();
+            return if i == 0 && suffix.as_os_str().is_empty() {
```
ilyagr, line 95: I don't like this change.

lib/src/simple_op_store.rs
```diff
@@ -212,12 +212,12 @@ impl OpStore for SimpleOpStore {
             let hex_prefix = prefix.hex();
             if hex_prefix.len() == OPERATION_ID_LENGTH * 2 {
                 // Fast path for full-length ID
-                if matches_root || op_dir.join(hex_prefix).try_exists()? {
+                return Ok(if matches_root || op_dir.join(hex_prefix).try_exists()? {
```
ilyagr, line 215: I'm ~Ok with this here, but I like the original a bit better, and I don't want this to be enforced by clippy.

ilyagr (commented):
I like most of these, except for the `redundant_else` commit. I highlighted the two that seemed to get a lot worse. The others in that commit seem OKish to change, though I often am not sure about them and I'm not sure it's good to force those changes.

So, I'd prefer dropping that entire commit, but I don't insist on it (especially if the two cases I highlighted can be avoided)

I'm happy with the others, though I'd wait for others to take a look.

ilyagr:
Another thing to consider is that this will add to headache of us being affected by clippy false-positive bugs. I imagine these checks are less well-tested than the default checks.

On the bright side, these used to occur previously, and haven't for the past ~3 months, so perhaps Clippy changed something about their testing. But again, they might be more lax about less frequently used checks.

I don't think this is a blocker; if it becomes a problem, we can revert parts of this PR.

samueltardieu:
> This is not necessarily related to this PR, but `cargo +nightly clippy` currently has numerous errors of the type https://rust-lang.github.io/rust-clippy/master/index.html#too_long_first_doc_paragraph . I ended up using `cargo cranky` to ignore it, but if you feel like fixing some lints :)

Agreed. The first version of this PR had a `#![expect(clippy::too_long_first_doc_paragraph)]` in several of our crates, but this is not supported by older rustc unfortunately. This would have been a relief, as it conveys that we expect this warning (and prevents it from being displayed) and hope to have it fixed in the future.

samueltardieu:
> Another thing to consider is that this will add to headache of us being affected by clippy false-positive bugs. I imagine these checks are less well-tested than the default checks.

They are well-tested nontheless: many projects use `#![deny(clippy::pedantic)]` and `#[allow(clippy::A_PARTICULAR_LINT)]`.

> On the bright side, these used to occur previously, and haven't for the past ~3 months, so perhaps Clippy changed something about their testing. But again, they might be more lax about less frequently used checks.

New lints must now go (since a few weeks ago) through a FCP (final comment period) before being integrated into Clippy, and must be explicitly approved by one extra member of the Clippy team (in addition to the one approving the PR), and any member of the Clippy team may use a veto. This should lead to an even higher lint quality.
 
> I don't think this is a blocker; if it becomes a problem, we can revert parts of this PR.

Yes, removing a denying lint is done by removing the corresponding line in the top `Cargo.toml`, as lints are inherited by the various crates.

samueltardieu:
> I like most of these, except for the `redundant_else` commit (as you foresaw). I highlighted the two that seemed to get a lot worse. The others in that commit seem OKish to change, though I often am not sure about them and I'm not sure it's good to force those changes.
> 
> So, I'd prefer dropping that entire commit, but I don't insist on it (especially if the two cases I highlighted can be avoided)

Agreed. As I expected, this was the most controversial lint to enable. I'll remove it.

Cargo.toml
```diff
@@ -132,6 +132,14 @@ jj-lib = { path = "lib", version = "0.22.0" }
 jj-lib-proc-macros = { path = "lib/proc-macros", version = "0.22.0" }
 testutils = { path = "lib/testutils" }
 
+[workspace.lints.clippy]
+explicit_iter_loop = "deny"
+flat_map_option = "deny"
+implicit_clone = "deny"
+needless_for_each = "deny"
+semicolon_if_nothing_returned = "deny"
+uninlined_format_args = "deny"
```
yuja, line 141: Perhaps, these rules should be all `"warn"`ed? It doesn't make sense that the default rules are warned, whereas extra rules are denied.

cli/src/command_error.rs
```diff
@@ -352,7 +352,7 @@ impl From<SnapshotError> for CommandError {
                         size_diff, max_size.0, max_size,
                     )
                 } else {
-                    format!("it is {}; the maximum size allowed is ~{}.", size, max_size,)
+                    format!("it is {size}; the maximum size allowed is ~{max_size}.",)
```
yuja, line 355: nit: remove trailing comma

lib/src/git_backend.rs
```diff
@@ -957,7 +957,7 @@ impl Backend for GitBackend {
             .map_err(|err| to_read_object_err(err, id))?;
         let target = String::from_utf8(blob.take_data())
             .map_err(|err| to_invalid_utf8_err(err.utf8_error(), id))?
-            .to_owned();
+            .clone();
```
yuja, line 960: nit: this `.clone()` can be removed at all

cli/tests/test_operations.rs
```diff
@@ -738,7 +738,7 @@ fn test_op_recover_from_bad_gc() {
     let repo_path = test_env.env_root().join("repo");
     let git_object_path = |hex: &str| {
         let (shard, file_name) = hex.split_at(2);
-        let mut file_path = repo_path.to_owned();
+        let mut file_path = repo_path.clone();
```
yuja, line 741: I'm not against this lint rule, but I tend to use `.to_owned()` if the owned type has a corresponding slice type, and an owned type isn't needed at call site.

yuja (approved):

Cargo.toml
samueltardieu: Warnings are denied in the CI, or they would appear green:

```
cargo +stable clippy --all-features --workspace --all-targets -- -D warnings
```

It would rather make sense do deny them as well during development. I've added two patches to the serie:

- One near the beginning, which allows `clippy::too_long_first_doc_paragraph` with a higher priority, to avoid being bothered by this one until we have fixed the doc comments.
- One at the end, which sets the default clippy severity to deny, to match what is done in the CI.

cli/tests/test_operations.rs
samueltardieu: The idea behind this lint is that sometimes you may remove the `.clone()` completely as you already have the type you want. When you call `.to_owned()`, it does not appear as clearly, and you are not warned by the compiler or clippy about it.

lib/src/git_backend.rs
samueltardieu: Done. One other example where seeing `.clone()` makes us question whether this is really needed or not, while seeing `to_owned()` might be legitimate if we had a slice.

Cargo.toml
yuja: > It would rather make sense do deny them as well during development.

I disagree. Leaving e.g. unused variables should be okay when testing some ideas. It would be annoying if compilation failed because of minor nits.

martinvonz: I was thinking the same, but I thought these checks would be done by `cargo clippy` only. Will they also be done as part of the regular build/test?

samueltardieu: All those lints are for Clippy, they will not make the compilation fail in any way, only runs of `cargo clippy` that you would use only when your code is ready to be shared.

lib/src/git_backend.rs
yuja: Just for clarification, I spotted it just because there was a diff. It doesn't matter (to me) whether it was `.clone()` or not.

Cargo.toml
samueltardieu: Also note that I'm not opposed to making them all warnings if people prefer, as long as the CI deny warnings it will keep the codebase just as clean.

Now that I think more about it, it might make sense if people use `cargo clippy` as a linter in their IDE (instead of `cargo check`, I know that VScode [that I don't use] can be configured that way).

yuja: It's technically correct. However, compile and lint errors are both highlighted as errors in editor, so I think their error levels are conceptually the same.

martinvonz: Making them all warnings and just having CI promote warnings to errors sounds good to me too.

cli/tests/test_operations.rs
samueltardieu: I've made them warnings.

Cargo.toml
samueltardieu: I've made them warnings, the CI will turn them into errors.

Cargo.toml
```diff
@@ -132,6 +132,17 @@ jj-lib = { path = "lib", version = "0.22.0" }
 jj-lib-proc-macros = { path = "lib/proc-macros", version = "0.22.0" }
 testutils = { path = "lib/testutils" }
 
+[workspace.lints.clippy]
+all = "warn"
```
yuja, line 136: this appears to demote default-denied lints to warning.

Cargo.toml
```diff
@@ -132,6 +132,17 @@ jj-lib = { path = "lib", version = "0.22.0" }
...
+explicit_iter_loop = "warn"
+flat_map_option = "warn"
+implicit_clone = "warn"
+needless_for_each = "warn"
+semicolon_if_nothing_returned = "warn"
+# Until all documentation comments are fixed, do not warn about them
+too_long_first_doc_paragraph = { level = "allow", priority = 1 }
```
yuja, line 143: got unknown lint warning on stable.

Cargo.toml
samueltardieu: You're right, moreover this is not needed anymore since I want to let them in the default state.

Cargo.toml
samueltardieu: Darn. Added a temporary `allow` for unknown lint names. That should take care of it.

PhilipMetzger:
I don't have a problem with doing this, although some discussionin the Discord beforehand would've been valuable, since this affects all pending PRs. This is similar in style to Matt's one import per line PR.

martinvonz:
> I don't have a problem with doing this, although some discussion in the Discord beforehand would've been valuable, since this affects all pending PRs. This is similar in style to Matt's one import per line PR.

That's fair. A heads-up on Discord would have been good, but I also don't think contributors should have to join Discord, so I think it's more important that the PR stays or for at least a day or two to get people a chance to comment. Does 24 hours seem like a reasonable time to you too or do you think we should require a longer time?

martinvonz:
I think @PhilipMetzger's concern was not with merge conflicts but that we should give people chime in on style guide changes.

PhilipMetzger:
> That's fair. A heads-up on Discord would have been good, but I also don't think contributors should have to join Discord, so I think it's more important that the PR stays or for at least a day or two to get people a chance to comment. Does 24 hours seem like a reasonable time to you too or do you think we should require a longer time?

I think we have a implicit three day period on RFC style changes, so that should be fine. 

> > I don't have a problem with doing this, although some discussion in the Discord beforehand would've been valuable, since this affects all pending PRs. This is similar in style to Matt's one import per line PR.
> 
> Many files are touched by this PR, but changes are very sparse. I am not sure there will be many conflicts.

Its not about conflicts, as you essentially made a [LSC][lsc] without informing anyone and getting appropriate buy-ins, even if the technical steering for the project isn't defined yet. It also makes landing currently approved PRs harder, as we now _enforce_ new clippy checks, which weren't enforced before. 

[lsc]: https://chromium.googlesource.com/chromium/src/+/HEAD/docs/process/lsc/lsc_workflow.md

PhilipMetzger:
> I wasn't actually aware of the 3-day period on RFC-style changes (same as design docs?), but I think you're saying that 24 hours is fine for a style change like this PR?

Sorry, it seems I was unclear. I meant that usually Authors let their RFC like proposals open for discussion for three days, before a decision was made (again, I am recalling this). 

The no unsafe in `cli` change was discussed beforehand and the import change was also brought up before. 

I'd to wait another day, so we give everyone time to react though.

ilyagr:
(Not really specific to this PR) As an experiment, I added a label for this.

I've occasionally missed having a way to say that a PR LGTM but I want to give others a chance to look at it. Perhaps a label could work.

We could also have a convention for how long the label should stay on a PR before it's merged. We could mention this in the label description. For now, I'll leave this to everyone's judgement with the author having the final say.

Cargo.toml
yuja: If there aren't many places to annotate, it might be better to add scoped `#[allow(too_long_first_doc_paragprah)] #[allow(unknown_lints)]`.

samueltardieu: In fact, adding summaries (for a better documentation) will be even better: #4581

PR #4578, opened 2024-10-04
Fix: Disallow revset names starting with a number.

matts1:
Currently, `5` is a valid function name. This seems incorrect

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* Revset function names can no longer start with a number.
+
 ### Deprecations
 
 ### New features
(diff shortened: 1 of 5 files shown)


lib/src/revset.pest
```diff
@@ -64,7 +64,7 @@ compat_sub_op = { "-" }
 infix_op = _{ union_op | intersection_op | difference_op | compat_add_op | compat_sub_op }
 
 function = { function_name ~ "(" ~ whitespace* ~ function_arguments ~ whitespace* ~ ")" }
-function_name = @{ (ASCII_ALPHANUMERIC | "_")+ }
+function_name = @{ (ASCII_ALPHA | "_") ~ (ASCII_ALPHANUMERIC | "_")* }
```
yuja, line 67: Can you also update fileset.pest and add a few parsing examples to revset/fileset_parser.rs tests?

Since this could break user aliases, it might be better to add changelog entry.

yuja (commented):
Thanks!

martinvonz (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* Revset and Fileset function names can no longer start with a number.
```
yuja, line 14: nit: "fileset" can be removed since we don't have fileset aliases yet.

yuja (approved):

Issue #3673, opened 2024-05-12
FR: Enhance aliases to run multiple commands with conditional logic


kuchta:
If aliases would be able to run multiple commands with conditional logic, many workflow specific commands (porcelain in git jargon) could be implemented by the user or his/hers company.

Many current commands expect some idiomatic workflow which could be implemented as a default aliases if aliases allow for conditional logic and new command eg. `is-same-commit <revision> <revision>`

This change would externalize such workflow specific commands (and discussions) to user configurable defaults and be potentionaly replaces by user/company prefered workflow...

PhilipMetzger:
This FR is probably a subset of #3262, as it is along the static/dynamic configuration discussion, which has been brought up many times in Discord.

fowles:
I actually just ran across this myself.  I would very much like an alias that lets me do `jj branch set main -r @- && jj git push -b main`.  Bonus points for allowing it to take the branch name as an argument or to figure it out from the log.

matts1:
This feels somewhat related to #3575 and #3577.

I think we could do it with minimal changes to the templating language. We'd just have to swap out the builtins for a few different ones, and require the user to create a template returning a `List<List<String>>`. For example, I could write `jj upload` as:
```toml
[template-aliases]
'stack(x)' = 'surround("mutable() && ::(", ")", x)'

[aliases]
upload = {
   args = {
         "fix": {
             "short": "-f",
             "long": "--fix",
             "type": "bool"
         },
         "lint": {
             "short": "-l",
             "long": "--lint",
             "type": "bool"
         },
         "revision": {
             "short": "-r",
             "long": "--revision",
             "type": "revset"
             "default": "@"
         }
   },
   "positional": True,
   commands = '
      [
          if(args.fix,  ["fix", "-r", stack("opts.revision)]),
          if(args.lint, ["lint", "-r", stack(opts.revision)])
          ["git", "push", "-r", opts.revision] ++ positional
      ]
   '
}
```

What do people think of this? I personally have a lot of use-cases that I'd be able to solve with something like this.

yuja:
> ```toml
> [template-aliases]
> 'stack(x)' = 'surround("mutable() && ::(", ")", x)'
> 
> [aliases]
> upload = {
>    args = {
>          "fix": {
>              "short": "-f",
>              "long": "--fix",
>              "type": "bool"
>          },
>          "lint": {
>              "short": "-l",
>              "long": "--lint",
>              "type": "bool"
>          },
[...]

PR #4580, opened 2024-10-04
refactor: Allow the aliases map to map to arbitrary types.

matts1:
For #3673, we will have aliases such as:
```toml
'upload(revision)' = [
  ["fix", "-r", "$revision"],
  ["lint", "-r", "$revision"],
  ["git", "push", "-r", "$revision"],
]
```

Template aliases:
1) Start as Config::Value
2) Are converted to String
3) Are placed in the alias map
4) Expand to a TemplateExpression type via expand_defn.

However, command aliases:
1) Start as Config::Value
2) Are converted to Vec<Vec<String>>
3) Are placed in an alias map
4) Do not expand

Thus, AliasesMap will need to support non-string values.

# Checklist

If applicable:

### DIFF
--- a/lib/src/dsl_util.rs
+++ b/lib/src/dsl_util.rs
@@ -424,28 +424,32 @@ impl<R: RuleType> StringLiteralParser<R> {
 
 /// Map of symbol and function aliases.
 #[derive(Clone, Debug, Default)]
-pub struct AliasesMap<P> {
-    symbol_aliases: HashMap<String, String>,
+pub struct AliasesMap<P, V> {
+    symbol_aliases: HashMap<String, V>,
     // name: [(params, defn)] (sorted by arity)
-    function_aliases: HashMap<String, Vec<(Vec<String>, String)>>,
+    function_aliases: HashMap<String, Vec<(Vec<String>, V)>>,
     // Parser type P helps prevent misuse of AliasesMap of different language.
     parser: P,
 }
 
-impl<P> AliasesMap<P> {
+impl<P, V> AliasesMap<P, V> {
     /// Creates an empty aliases map with default-constructed parser.
     pub fn new() -> Self
     where
         P: Default,
     {
-        Self::default()
+        Self {
+            symbol_aliases: Default::default(),
+            function_aliases: Default::default(),
+            parser: Default::default(),
+        }
     }
 
     /// Adds new substitution rule `decl = defn`.
     ///
     /// Returns error if `decl` is invalid. The `defn` part isn't checked. A bad
     /// `defn` will be reported when the alias is substituted.
-    pub fn insert(&mut self, decl: impl AsRef<str>, defn: impl Into<String>) -> Result<(), P::Error>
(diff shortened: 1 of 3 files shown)


yuja (commented):
I'm skeptical how many alias handling helpers can be reused for command aliases. Command aliases are so different. If we can't leverage `dsl_util::expand_aliases()` for example, it wouldn't make much sense to use `AliasesMap`. It's just a wrapper of `HashMap`.

matts1:
I've found it to be useful primarily because of the storing multiple overloads by arity, and the `AliasDeclarationParser` is pretty much word-for-word the same as in revset and template aliases. There's a few things that can't be reused, but I believe that AliasesMap is generally a good abstraction for whenever you want to do function overloading.

lib/src/dsl_util.rs
```diff
@@ -475,46 +488,53 @@ impl<P> AliasesMap<P> {
     }
 
     /// Looks up symbol alias by name. Returns identifier and definition text.
-    pub fn get_symbol(&self, name: &str) -> Option<(AliasId<'_>, &str)> {
+    pub fn get_symbol(&self, name: &str) -> Option<(AliasId<'_>, &<V as Deref>::Target)> {
```
yuja, line 491: Maybe we can just make it return `&String` (= `&V`) instead of `&str`. For a generic container type, there wouldn't be much reason to convert ref to slice type.

yuja:
> I've found it to be useful primarily because of the storing multiple overloads by arity, and the `AliasDeclarationParser` is pretty much word-for-word the same as in revset and template aliases. There's a few things that can't be reused, but I believe that AliasesMap is generally a good abstraction for whenever you want to do function overloading.

Hmm, I'm afraid that there might be some weird requirement specific to command aliases, but I don't have a strong reason to against generalizing `AliasesMap`.

lib/src/dsl_util.rs
```diff
@@ -475,46 +487,46 @@ impl<P> AliasesMap<P> {
...
 
     /// Looks up function aliases by name.
-    fn get_function_overloads(&self, name: &str) -> Option<AliasFunctionOverloads<'_>> {
+    fn get_function_overloads(&self, name: &str) -> Option<AliasFunctionOverloads<'_, V>> {
         let (name, overloads) = self.function_aliases.get_key_value(name)?;
         Some(AliasFunctionOverloads { name, overloads })
     }
 }
 
 #[derive(Clone, Copy, Debug)]
```
yuja, line 510: nit: remove `Copy` as it can no longer be derived?

lib/src/dsl_util.rs
```diff
@@ -423,16 +423,28 @@ impl<R: RuleType> StringLiteralParser<R> {
 }
 
 /// Map of symbol and function aliases.
-#[derive(Clone, Debug, Default)]
-pub struct AliasesMap<P> {
-    symbol_aliases: HashMap<String, String>,
+#[derive(Clone, Debug)]
+pub struct AliasesMap<P, V = String> {
```
yuja, line 427: nit: remove default type or use it consistently

(I prefer not adding `V = String`.)

yuja (approved):
Thanks.

PR #4581, opened 2024-10-04
Short first doc paragraph

samueltardieu:
This removes the need for ignoring the `too_long_first_doc_paragraph` lint.

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -522,13 +522,15 @@ impl ReadonlyUserRepo {
     }
 }
 
-/// A bookmark that should be advanced to satisfy the "advance-bookmarks"
-/// feature. This is a helper for `WorkspaceCommandTransaction`. It provides a
-/// type-safe way to separate the work of checking whether a bookmark can be
-/// advanced and actually advancing it. Advancing the bookmark never fails, but
-/// can't be done until the new `CommitId` is available. Splitting the work in
-/// this way also allows us to identify eligible bookmarks without actually
-/// moving them and return config errors to the user early.
+/// A advanceable bookmark to satisfy the "advance-bookmarks" feature.
+///
+/// This is a helper for `WorkspaceCommandTransaction`. It provides a
+/// type-safe way to separate the work of checking whether a bookmark
+/// can be advanced and actually advancing it. Advancing the bookmark
+/// never fails, but can't be done until the new `CommitId` is
+/// available. Splitting the work in this way also allows us to
+/// identify eligible bookmarks without actually moving them and
+/// return config errors to the user early.
 pub struct AdvanceableBookmark {
     name: String,
     old_commit_id: CommitId,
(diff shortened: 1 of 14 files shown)


lib/proc-macros/src/lib.rs
```diff
@@ -6,6 +6,8 @@ use quote::quote;
 use syn::parse_macro_input;
 use syn::DeriveInput;
 
+/// Sequentially hash struct members to derive `ContentHash` trait
+///
```
yuja, line 10: nit: it also supports enum types. Maybe just copy the std doc phrase?
https://doc.rust-lang.org/std/fmt/derive.Debug.html

lib/src/conflicts.rs
```diff
@@ -367,10 +367,12 @@ pub fn materialized_diff_stream<'a>(
         .buffered((store.concurrency() / 2).max(1))
 }
 
-/// Parses conflict markers from a slice. Returns None if there were no valid
-/// conflict markers. The caller has to provide the expected number of merge
-/// sides (adds). Conflict markers that are otherwise valid will be considered
-/// invalid if they don't have the expected arity.
+/// Parse conflict markers from a slice
```
yuja, line 370: nit: "Parse**s**" and terminate with period? I think that's the convention of Rust std doc.

(applies to other doc changes)

lib/src/rewrite.rs
```diff
@@ -462,6 +462,9 @@ pub struct MoveCommitsStats {
     pub num_skipped_rebases: u32,
 }
 
+/// Rebase and move `target_commits` in graph based on new parent and child
+/// nodes
+///
 /// Moves `target_commits` from their current location to a new location in the
 /// graph, given by the set of `new_parent_ids` and `new_children`.
```
yuja, line 469: nit: I think the first paragraph repeats the same thing. Maybe we can just insert blank line here?

lib/src/stacked_table.rs
```diff
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! A persistent table of fixed-size keys to variable-size values. The keys are
-//! stored in sorted order, with each key followed by an integer offset into the
-//! list of values. The values are concatenated after the keys. A file may have
-//! a parent file, and the parent may have its own parent, and so on. The child
-//! file then represents the union of the entries.
+//! A oersistent table of fixed-size keys to variable-size values
```
yuja, line 15: nit: typo "persistent"

cli/src/cli_util.rs
```diff
@@ -522,6 +522,8 @@ impl ReadonlyUserRepo {
     }
 }
 
+/// Type-safe helper for conditionally advancing bookmarks
+///
 /// A bookmark that should be advanced to satisfy the "advance-bookmarks"
 /// feature. This is a helper for `WorkspaceCommandTransaction`. It provides a
```
yuja, line 528: nit: insert blank line after "... feature." instead?

"helper for ..." sounds like an object that advances bookmark, but this is just a data object.

yuja (approved):

PR #4585, opened 2024-10-04
docs: mention `snapshot.auto-track` in `config.md` too

martinvonz:
I forgot to add the `snapshot.auto-track` config option to `config.md` when I added it. This patch copies it from `working-copy.md` and modifies it slightly.

# Checklist

If applicable:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -954,6 +954,19 @@ using `jj debug watchman status`.
 
 ## Snapshot settings
 
+### Paths to automatically track
+
+All new files in the working copy that don't match the ignore patterns are
+tracked by default. You can set the `snapshot.auto-track` to set which paths
+get automatically tracked when they're added to the working copy. See the
+[fileset documentation](filesets.md) for the syntax. Files with paths matching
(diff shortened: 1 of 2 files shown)


PhilipMetzger (approved):

tim-janik:
Your docs say:
> You can use `jj file untrack` to untrack a file while keeping it in the working
> copy. However, first [ignore](working-copy.md#ignored-files) them or remove them
> from the `snapshot.auto-track` patterns; otherwise they will be immediately
> tracked again.

Thanks for explicitly explaining the need to ignore files to properly make use of `jj file untrack`.
Wouldn't it make sense to add an option `--add-ignore` or similar, so `jj file untrack --add-ignore foo.txt` would add `foo.txt` to `./.gitignore` ?

martinvonz:
> Wouldn't it make sense to add an option `--add-ignore` or similar, so `jj file untrack --add-ignore foo.txt` would add `foo.txt` to `./.gitignore` ?

You're not alone: https://github.com/martinvonz/jj/issues/3493

PR #4588, opened 2024-10-05
Colocated workspaces using git worktrees

### LINKED ISSUE #4436: FR: workspaces should also be collocated in --colocate mode

From https://github.com/martinvonz/jj/discussions/4424#discussioncomment-10609454

Currently if you create a workspace in a collocated repo, the new workspace is not recognisable as a Git repository by any git tools. It does not have a .git directory.

The most painful tool, as usual, is Nix flakes, which silently drops back to "no git repo" mode and stops looking at your .gitignore file, and then just repeatedly copies your enormous build artifacts folder into the nix store. I actually only discovered this issue because Nix caused my computer to run out of space. (As usual this Nix's fault!)

I think we should be able to create a corresponding git worktree for each workspace. (We might consider supporting importing worktrees you created with git, but not a priority really.)

Git worktrees are pretty simple and really just involve a couple of files and a ref. So I don't think it'll be that hard.

Unfortunately I don't think you can use libgit2 for this because there are no APIs that don't also do a full checkout into a newly created worktree directory. There are no options that prevent checkout. Nor does gitoxide help, it has some worktree code but not useful either (search docs.rs for gix-worktree).

So I recommend just copying roughly these bits of libgit2's `git_worktree_add` function (unless you think that much would violate GPL, in which case just read https://git-scm.com/docs/git-worktree sections Refs, and Details):

https://github.com/libgit2/libgit2/blob/main/src/libgit2/worktree.c#L355-L366

https://github.com/libgit2/libgit2/blob/main/src/libgit2/worktree.c#L389-L409

All it really needs to do is write a couple of files in the .git directory in the main workspace and in the newly created worktree, and then set an appropriate HEAD on the new worktree. It may be easier to add this in gitoxide, but this is a pretty niche use case.

None, but see the linked discussion

cormacrelf:
Fixes #4436

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1749,9 +1749,7 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
             .transpose()?;
 
         if self.working_copy_shared_with_git {
-            // FIXME: should have a GIT_DIR pointing to the relevant worktree,
-            // if we're in a workspace in a colocated repo
-            let git_repo = self.git_backend().unwrap().open_git_repo()?;
+            let git_repo = git2::Repository::open(self.workspace().workspace_root())?;
             if let Some(wc_commit) = &maybe_new_wc_commit {
                 git::reset_head(tx.repo_mut(), &git_repo, wc_commit)?;
             }
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -382,7 +382,12 @@ impl CommandHelper {
         let op_head = self.resolve_operation(ui, workspace.repo_loader())?;
         let repo = workspace.repo_loader().load_at(&op_head)?;
         let env = self.workspace_environment(ui, &workspace)?;
-        revset_util::warn_unresolvable_trunk(ui, repo.as_ref(), &env.revset_parse_context())?;
+        revset_util::warn_unresolvable_trunk(
+            ui,
+            repo.as_ref(),
+            &env.revset_parse_context(),
+            workspace.workspace_id(),
+        )?;
         WorkspaceCommandHelper::new(ui, workspace, repo, env, self.is_at_head_operation())
     }
 
@@ -645,7 +650,7 @@ impl WorkspaceCommandEnvironment {
         &self.workspace_id
     }
 
-    pub(crate) fn revset_parse_context(&self) -> RevsetParseContext {
+    pub fn revset_parse_context(&self) -> RevsetParseContext {
         let workspace_context = RevsetWorkspaceContext {
             path_converter: &self.path_converter,
             workspace_id: &self.workspace_id,
@@ -669,7 +674,10 @@ impl WorkspaceCommandEnvironment {
     /// Creates fresh new context which manages cache of short commit/change ID
     /// prefixes. New context should be created per repo view (or operation.)
     pub fn new_id_prefix_context(&self) -> IdPrefixContext {
-        let context = IdPrefixContext::new(self.command.revset_extensions().clone());
+        let context = IdPrefixContext::new(
+            self.command.revset_extensions().clone(),
+            Some(self.workspace_id().clone()),
+        );
         match &self.short_prefixes_expression {
             None => context,
             Some(expression) => context.disambiguate_within(expression.clone()),
@@ -747,7 +755,10 @@ impl WorkspaceCommandEnvironment {
         // Not using self.id_prefix_context() because the disambiguation data
         // must not be calculated and cached against arbitrary repo. It's also
         // unlikely that the immutable expression contains short hashes.
-        let id_prefix_context = IdPrefixContext::new(self.command.revset_extensions().clone());
+        let id_prefix_context = IdPrefixContext::new(
+            self.command.revset_extensions().clone(),
+            Some(self.workspace_id.clone()),
+        );
         let to_rewrite_revset =
             RevsetExpression::commits(commits.into_iter().cloned().collect_vec());
         let mut expression = RevsetExpressionEvaluator::new(
@@ -928,8 +939,10 @@ impl WorkspaceCommandHelper {
     fn import_git_head(&mut self, ui: &Ui) -> Result<(), CommandError> {
         assert!(self.may_update_working_copy);
         let command = self.env.command.clone();
+        let workspace_id = self.workspace_id().clone();
+        let git_repo = self.git_either_colocated_or_backend()?;
         let mut tx = self.start_transaction();
-        git::import_head(tx.repo_mut())?;
+        git::import_head(tx.repo_mut(), &git_repo, &workspace_id)?;
         if !tx.repo().has_changes() {
             return Ok(());
         }
@@ -945,8 +958,9 @@ impl WorkspaceCommandHelper {
         //   out yet.
 
         let mut tx = tx.into_inner();
-        let old_git_head = self.repo().view().git_head().clone();
-        let new_git_head = tx.repo().view().git_head().clone();
+        let workspace_id = self.workspace_id();
+        let old_git_head = self.repo().view().git_head(workspace_id).clone();
+        let new_git_head = tx.repo().view().git_head(workspace_id).clone();
         if let Some(new_git_head_id) = new_git_head.as_normal() {
             let workspace_id = self.workspace_id().to_owned();
             let new_git_head_commit = tx.repo().store().get_commit(new_git_head_id)?;
@@ -1074,11 +1088,34 @@ impl WorkspaceCommandHelper {
         self.working_copy_shared_with_git
     }
 
-    pub fn open_colocated_git_worktree(&self) -> Result<Option<git2::Repository>, git2::Error> {
+    pub fn open_colocated_git_worktree_git2(
+        &self,
+    ) -> Result<Option<git2::Repository>, git2::Error> {
         self.working_copy_shared_with_git()
             .then_some(self.workspace_root())
             .map(git2::Repository::open)
             .transpose()
+    }
+
+    pub fn open_colocated_git_worktree_gix(
+        &self,
+    ) -> Result<Option<gix::Repository>, Box<gix::open::Error>> {
+        self.working_copy_shared_with_git()
+            .then_some(self.workspace_root())
+            .map(gix::open)
+            .transpose()
+            .map_err(Box::new)
+    }
+
+    pub fn git_either_colocated_or_backend(&self) -> Result<gix::Repository, CommandError> {
+        self.open_colocated_git_worktree_gix()
+            .map_err(user_error)
+            .transpose()
+            .unwrap_or_else(|| {
+                self.git_backend()
+                    .ok_or_else(|| user_error("JJ repo is not backed by git"))
+                    .map(|g| g.git_repo())
+            })
     }
 
     pub fn format_file_path(&self, file: &RepoPath) -> String {
@@ -1755,9 +1792,9 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
             .map(|commit_id| tx.repo().store().get_commit(commit_id))
             .transpose()?;
 
-        if let Some(git_worktree) = self.open_colocated_git_worktree()? {
+        if let Some(git_worktree) = self.open_colocated_git_worktree_git2()? {
             if let Some(wc_commit) = &maybe_new_wc_commit {
-                git::reset_head(tx.repo_mut(), &git_worktree, wc_commit)?;
+                git::reset_head(tx.repo_mut(), &git_worktree, self.workspace_id(), wc_commit)?;
             }
             let refs = git::export_refs(tx.repo_mut())?;
             print_failed_git_export(ui, &refs)?;
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1074,6 +1074,13 @@ impl WorkspaceCommandHelper {
         self.working_copy_shared_with_git
     }
 
+    pub fn open_colocated_git_worktree(&self) -> Result<Option<git2::Repository>, git2::Error> {
+        self.working_copy_shared_with_git()
+            .then_some(self.workspace_root())
+            .map(git2::Repository::open)
+            .transpose()
+    }
+
     pub fn format_file_path(&self, file: &RepoPath) -> String {
         self.path_converter().format_file_path(file)
     }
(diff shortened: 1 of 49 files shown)


cli/src/cli_util.rs
```diff
@@ -941,6 +941,19 @@ impl WorkspaceCommandHelper {
         self.workspace.workspace_id()
     }
 
+    pub fn workspace_contains_actual_repo(&self) -> bool {
+        // we won't assume your default workspace is called "default"
+        //
+        // repo_path is not canonicalized (and may be relative) whereas workspace_root is.
+        self.repo_path()
+            .canonicalize()
```
samueltardieu, line 955: Instead of trying to canonicalize paths for comparing them, wouldn't it suffice to check whether the current workspace as a `.jj/repo` directory as its root? Something like:

```rust
    fs::metadata(self.workspace_root().join(".jj").join("repo")).is_ok_and(|s| s.is_dir())
```

cli/src/commands/workspace/add.rs
```diff
@@ -117,6 +120,33 @@ pub fn cmd_workspace_add(
             .unwrap()
             .to_string()
     };
+
+    if let Some(git_backend) = old_workspace_command.git_backend() {
+        let should_add_worktree = if git_backend.git_workdir().is_some() {
+            !args.no_colocate
+        } else {
+            args.colocate
+        };
```
PhilipMetzger, line 129: Nit: Can this move into a config flag? I dislike that we do it implicitly.

PhilipMetzger: as I'm thinking about the use-case here https://github.com/martinvonz/jj/pull/2512

cli/src/cli_util.rs
cormacrelf: Ah, this commit was intended to be be landed separately as it will need a bit of chat about when it's okay to delete the default workspace, and whether we can implement turning the repo back from a bare repo.

cli/src/commands/workspace/add.rs
cormacrelf: I was generally trying to make it follow what you wanted in that repo -- for me the decision to use collocation is at least repo-wide. E.g. if you use Nix in a repo, it needs a .git, all the time, as everything breaks apart silently if you forget. See the issue linked at the top.

All these UI choices would be much easier if you could change your mind about collocation after you had created or cloned the repo or created a workspace. It just wouldn't matter so much. I don't think it would be that hard to write.

cormacrelf: For the time being we can just follow `jj git init` and make it entirely dependent on the `--colocate` flag and just require people to remember to use it. Ok?

lib/src/simple_op_store.rs
```diff
@@ -495,10 +518,16 @@ fn view_from_proto(proto: crate::protos::op_store::View) -> View {
     }
 
     #[allow(deprecated)]
-    if proto.git_head.is_some() {
-        view.git_head = ref_target_from_proto(proto.git_head);
+    if !proto.git_heads.is_empty() {
+        view.git_heads = git_heads_from_proto(proto.git_heads);
+    } else if proto.git_head.is_some() {
+        view.git_heads = hashmap! {
+            WorkspaceId::default() => ref_target_from_proto(proto.git_head),
```
yuja, line 525: Since the "default" workspace could have been renamed, we might have to guess the workspace id based on `wc_commit_ids`.

lib/src/revset.rs
```diff
@@ -1630,6 +1657,7 @@ impl<'a> DefaultSymbolResolver<'a> {
                 .iter()
                 .flat_map(|ext| ext.as_ref().new_resolvers(repo))
                 .collect(),
+            workspace_ctx,
```
yuja, line 1671: `HEAD@git` would have to be translated to `CommitRef::GitHead(WorkspaceId)` or something at `lower_expression()` (who knows the current workspace.)

martinvonz:
> workspace: Refuse to implicitly delete the default workspace
This tripped me up personally. Didn't know what `jj workspace forget`
did, ran it, and had to `jj clone` again.

You should be able to undo that operation. Maybe it's enough that we print a hint about that? Or was there some other consequence I'm missing?

lib/src/revset.rs
yuja: BTW, it might be okay to deprecate `HEAD@git` symbol since it can no longer be abstracted as a remote ref of the local `@git` remote. It serves as an indicator of colocated repo, but I've never used the `HEAD@git` symbol, which should usually point to `@-`.

lib/src/simple_op_store.rs
cormacrelf: Can you show me how to rename the default workspace? There is no `jj workspace rename` and we don't care about the case where people run `jj workspace forget default` and then use another one, because until this PR, no other workspace than default could have had a git_head created. Does this have to be through using JJ as a library? Is there a config flag I'm not seeing?

I wanted to create a test, and I'm struggling.

yuja: `jj workspace rename` was added recently. That said, it's good point that non-"default" workspace wouldn't have a colocated git repo.

I was thinking of using the workspace name if `wc_commit_ids.len() == 1` or something, but it might be better to just check if `wc_commit_ids.contains_key("default")`. It could go wrong, but new `git_heads[id]` will be set at some point later.

cormacrelf: Would it be okay to do a really hacky thing like this in the `view_from_proto` function, on the basis that the default workspace could not actually be in any other location before this PR?

```rust
            let guess_default_workspace_id = store_path
                .and_then(|sp| sp.parent())
                .and_then(|repo| repo.parent())
                .map(|jj| jj.join("working_copy").join("checkout"))
                .and_then(|checkout| fs::read(checkout).ok())
                .and_then(|buf| crate::protos::working_copy::Checkout::decode(buf.as_slice()).ok())
                .map(|checkout| checkout.workspace_id)
                .map(WorkspaceId::new)
                .filter(|workspace_id| view.wc_commit_ids.contains_key(workspace_id));
```

cormacrelf: In any case it's going to be great to have a CLI test that manipulates the views to make them use deprecated fields and sees what JJ can do with them.

yuja: > Would it be okay to do a really hacky thing like this in the `view_from_proto` function,

I would avoid that so long as the `git_heads` entry can be recovered by e.g. checking out new commit. And as you said, it's difficult to test.

cli/src/commands/workspace/add.rs
cormacrelf: Done, explicit --colocate only.

cli/src/git_util.rs
```diff
@@ -51,22 +51,84 @@ pub fn get_git_repo(store: &Store) -> Result<git2::Repository, CommandError> {
...
+    //     /repo/.git
+    // But for a JJ workspace of a colocated repo, this will be
+    //     /repo/.git/worktrees/second
+    // ... or, if the JJ repo was not originally colocated:
+    //     /repo/.jj/repo/store/git/worktrees/second
+    //
+    // ... and the regular file /second/.git will direct libgit2 to that location.
+    //
+    // So try to open the workspace root (/second) as a git repository.
+    let worktree_repo = git2::Repository::open(workspace.workspace_root()).ok()?;
```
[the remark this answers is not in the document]
cormacrelf: Done (used gix::open and Repository::common_dir to check this stuff) but didn't use the worktrees() method in this function. But I did also use it for the git_worktree_remove code, which benefitted from gix telling us if the worktree was locked.

cli/src/cli_util.rs
```diff
@@ -1591,7 +1593,9 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
             .map(|commit_id| tx.repo().store().get_commit(commit_id))
             .transpose()?;
 
-        if self.working_copy_shared_with_git {
+        if self.working_copy_shared_with_git() {
```
PhilipMetzger: Nit: you already have `working_copy_shared_with_git` initialized with the value above, so reuse it here.

cli/src/cli_util.rs
```diff
@@ -1591,7 +1593,9 @@ See https://martinvonz.github.io/jj/latest/working-copy/#stale-working-copy \
...
+            // FIXME: should have a GIT_DIR pointing to the relevant worktree,
+            // if we're in a workspace in a colocated repo
```
PhilipMetzger: Nit: Move the FIXME in the commit that modifies it, so the next one.

cli/src/cli_util.rs
PhilipMetzger, line 1: As a TODO until the PR is ready but this should have a good motivation for the change. See https://github.com/martinvonz/jj/blob/main/docs/contributing.md#code-reviews particularly this section 
> We are not particularly strict about the style, but please do explain the reason for the change unless it's obvious.

PhilipMetzger (commented):
Here's another batch of mostly minor nits.

lib/src/protos/op_store.proto
```diff
@@ -85,10 +85,13 @@ message View {
...
   // This field is just for historical reasons (before we had the RefTarget
-  // type). New Views have (only) the target field.
+  // type). New Views have (only) the git_head or git_heads fields.
   // TODO: Delete support for the old format.
   bytes git_head_legacy = 7 [deprecated = true];
-  RefTarget git_head = 9;
+  // Equivalent to one value of git_heads with a workspace id of "default"
+  RefTarget git_head = 9 [deprecated = true];
+  // We now track a git head per workspace.
+  map<string, RefTarget> git_heads = 11;
```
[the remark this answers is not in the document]
cormacrelf: Isn't `map<string, RefTarget>` simply syntax sugar for a repeated field?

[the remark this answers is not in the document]
PhilipMetzger: Yes, but having the option to not immediately getting a `HashMap` in gencode makes it more flexible. We always can store them internally as that before conversion.

cormacrelf:
The main substantive question I have left is, when we import and export from git, should this always do it for every single worktree? Or just the current one?

The approach I have taken so far is to do it for the current one. But I think this is spreading out the import over too many operations when the worktrees are, in fact, all right there, readable, and iterable.

lib/src/protos/op_store.proto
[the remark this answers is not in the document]
cormacrelf: Ah, and you can have a nice named repeated field type like WorkspaceView that we can extend later. And potentially absorb the wc_commit_ids field into.

lib/src/simple_op_store.rs
cormacrelf: I did manage to write a test for this. Possibly the most long-winded test in the whole repo. But I think we needed something to exercise the deprecated fields.

cormacrelf:
Ah, it would be quite complicated to import all the HEADs at once. If a git HEAD has changed in .git, then `WorkspaceCommandHelper::import_git_head` will check it out as the working copy, basically `jj new` it and write the tree to disk. To actually do that in all workspaces, we would need all other workspaces to be accessible at known paths all the time from any other workspace. And we do not currently store, in .jj/repo, where the workspaces' working copies are located, in order to do such a thing. And storing that data would imply we need a `jj workspace repair` to fix up the mapping of WorkspaceId -> path when stuff has moved on disk, which is perhaps the most annoying thing about git worktrees.

yuja:
> Ah, it would be quite complicated to import all the HEADs at once. If a git HEAD has changed in .git, then `WorkspaceCommandHelper::import_git_head` will check it out as the working copy, basically `jj new` it and write the tree to disk. To actually do that in all workspaces, we would need all other workspaces to be accessible at known paths all the time from any other workspace. And we do not currently store, in .jj/repo, where the workspaces' working copies are located, in order to do such a thing. And storing that data would imply we need a `jj workspace repair` to fix up the mapping of WorkspaceId -> path when stuff has moved on disk, which is perhaps the most annoying thing about git worktrees.

And workspaces might not always be accessible. They might be on USB stick or some network drive. So no, not all worktrees, but just the current one.

cormacrelf:
Is there anyone who can test on windows? I don't have a machine handy.

```
command=`"D:\\a\\jj\\jj\\target\\debug\\jj.exe" "workspace" "add" "--colocate" "../second"`
code=1
stdout=""
stderr=```
Created workspace in \"..\\second\"
Error: Git operation failed
Caused by: could not find repository at \'\\\\?\\C:\\Users\\runneradmin\\AppData\\Local\\Temp\\jj-test-fy26a2\\second\'; class=Repository (6); code=NotFound (-3)
```

That path, de-backslashed, is

```
\\?\C:\Users\runneradmin\AppData\Local\Temp\jj-test-fy26a2\second
```

which is a [Verbatim](https://doc.rust-lang.org/std/path/enum.Prefix.html#variant.Verbatim) path prefix.

I suspect Git likes its paths in the gitfile (`gitdir: <path>\n`) in a different format, but I don't know which one. It could even be Unix-style slashed format. If someone could run this on a windows machine that would help:

```sh
; git init repo
; cd repo
; touch file.txt && git add file.txt && git commit -m 'initial commit'
; git workspace add ../abc
; cat ../abc/.git
```

cli/src/commands/workspace/add.rs
```diff
@@ -73,6 +76,23 @@ pub struct WorkspaceAddArgs {
...
+    ///
+    /// This has a similar effect to `jj git init --colocate`,
+    /// but for new workspaces. It allows you to use Git tooling,
+    /// by creating a git worktree that links to JJ's git backend.
+    /// Git tools see a regular git worktree and a HEAD that is
+    /// kept updated to reflect `@-`. You should be able to run
+    /// `git worktree list` in the new workspace and see it there.
+    ///
+    /// This flag works even if a repository was not
+    /// originally initialised with `--colocate`.
```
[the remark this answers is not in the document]
cormacrelf: Updated, following a bit of the wording from `jj git init --help`.

lib/src/revset.rs
cormacrelf: I had to add the workspace id to the resolver anyway, because the resolver needs to be able to generate HEAD@git as a did-you-mean suggestion, and must do that only when the current workspace is colocated. Since it had to be added to the resolver anyway, it was simple to resolve `GitHead` without changing CommitRef.

yuja: Hmm, good point. I lean towards deprecating `HEAD@git` then.

As I said, it's no longer a valid abstraction, and the HEAD revision can be queried by `git_head()`. IIRC, `HEAD@git` resolution was added just because `git_head` template is rendered as such, not because we have a real use case.

yuja:
Could you somehow split this PR into two or three units? This looks too big to run multiple review rounds, and GitHub UI doesn't support per-patch acceptance.

One idea (not sure if this is the easiest/simplest option):
1. add minimal command to set up colocated workspace
2. migrate view
3. add `jj workspace` command integration

lib/src/revset.rs
yuja: https://github.com/martinvonz/jj/pull/4632

cormacrelf:
Closing to split in half, next is #4644

PR #4594, opened 2024-10-07
repo: pretend root commit exists in the root operation

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_commit_concurrent.rs
+++ b/lib/tests/test_commit_concurrent.rs
@@ -69,9 +69,9 @@ fn test_commit_parallel(backend: TestRepoBackend) {
     // the root commit
     assert_eq!(repo.view().heads().len(), num_threads + 1);
 
-    // One additional operation for the root commit, one for initializing the repo,
-    // one for checking out the initial commit.
-    assert_eq!(count_non_merge_operations(&repo), num_threads + 3);
+    // One additional operation for the root operation, one for checking out the
+    // initial commit.
+    assert_eq!(count_non_merge_operations(&repo), num_threads + 2);
 }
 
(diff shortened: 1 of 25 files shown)


lib/tests/test_commit_concurrent.rs
```diff
@@ -69,9 +69,9 @@ fn test_commit_parallel(backend: TestRepoBackend) {
     // the root commit
     assert_eq!(repo.view().heads().len(), num_threads + 1);
 
-    // One additional operation for the root commit, one for initializing the repo,
-    // one for checking out the initial commit.
-    assert_eq!(count_non_merge_operations(&repo), num_threads + 3);
+    // One additional operation for the root commit, one for checking out the
+    // initial commit.
```
martinvonz, line 73: Isn't it the operation "for the the root commit" that's no longer there? I.e., I would think this should look like this (but reflowed):

> // One additional operation for initializing the repo,
> // one for checking out the initial commit.

(Same below)

martinvonz (approved):
Thanks!

lib/tests/test_commit_concurrent.rs
yuja: My understanding is that this patch removes the "initialize repo" operation, so the count is decremented accordingly. It counts the number of ancestors, not the number of diffs between these.

martinvonz: Oh, I think what confused me is the existing error that "One additional operation for the root commit" should say "... for the root *operation*", right?

yuja: Right, fixed it.

PR #4599, opened 2024-10-07
squash: move the core functionality to jj_lib

torquestomp:
This allows 'squash' to be executed more easily from a programmatic context

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/squash.rs
+++ b/cli/src/commands/squash.rs
@@ -308,7 +308,7 @@ from the source will be moved into the destination.
                 .filter_map(|source| source.abandon.then_some(source.commit))
                 .collect_vec();
             combine_messages(
-                tx.base_workspace_helper(),
+                tx.base_workspace_helper().repo_path(),
                 &abandoned_commits,
                 destination,
                 settings,
--- a/cli/src/commands/squash.rs
+++ b/cli/src/commands/squash.rs
@@ -16,9 +16,9 @@ use itertools::Itertools as _;
 use jj_lib::commit::Commit;
 use jj_lib::commit::CommitIteratorExt;
 use jj_lib::matchers::Matcher;
-use jj_lib::merged_tree::MergedTree;
 use jj_lib::object_id::ObjectId;
 use jj_lib::repo::Repo;
+use jj_lib::rewrite;
 use jj_lib::settings::UserSettings;
 use tracing::instrument;
 
@@ -196,12 +196,6 @@ pub fn move_diff(
     tx.base_workspace_helper()
         .check_rewritable(sources.iter().chain(std::iter::once(destination)).ids())?;
 
-    struct SourceCommit<'a> {
-        commit: &'a Commit,
-        parent_tree: MergedTree,
-        selected_tree: MergedTree,
-        abandon: bool,
-    }
     let mut source_commits = vec![];
     for source in sources {
         let parent_tree = source.parent_tree(tx.repo())?;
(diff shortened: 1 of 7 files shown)


cli/src/commands/squash.rs
```diff
@@ -227,105 +222,65 @@ from the source will be moved into the destination.
...
-        let abandon = !keep_emptied && selected_tree.id() == source_tree.id();
-        if !abandon && selected_tree_id == parent_tree.id() {
-            // Nothing selected from this commit. If it's abandoned (i.e. already empty), we
-            // still include it so `jj squash` can be used for abandoning an empty commit in
-            // the middle of a stack.
-            continue;
+
+        let commit = rewrite::CommitToSquash::new(tx.repo(), source.clone(), selected_tree)?;
+        if !keep_emptied && commit.is_full_selection() {
+            abandoned_commits.push(source);
```
yuja, line 228: It's weird that we have to test the same condition by caller and callee. Maybe we can extract description handling as a callback? It might be also good to make `rewrite::squash_commits()` return `CommitBuilder` and let caller do postprocessing as needed.

lib/src/rewrite.rs
```diff
@@ -830,3 +830,133 @@ pub fn move_commits(
...
+    // from the commit).
+    //
+    // Both `is_full_selection()` and `is_empty_selection()`
+    // can be true if the commit is itself empty.
+    pub fn is_empty_selection(&self) -> bool {
+        self.selected_tree.id() == self.parent_tree.id()
+    }
+}
+
+pub struct SquashOptions {
```
yuja, line 865: nit: add `#[derive(Clone, Debug)]`?

lib/src/rewrite.rs
```diff
@@ -830,3 +830,133 @@ pub fn move_commits(
...
+pub enum SquashResult {
+    /// No inputs contained actual changes.
+    NoChanges,
+    /// Destination was rewritten.
+    NewCommit(Commit),
+}
+
+// Squash `sources` into `destination` and return a CommitBuilder for the
+// resulting commit. Caller is responsible for setting the description and
+// finishing the commit.
```
yuja, line 879: nit: use `///` to make it doc comment (applies to other function docs)

cli/src/commands/squash.rs
torquestomp: I tried both of these approaches first but couldn't get them to work due to lifetime issues.  `combine_descriptions()` requires an immutable borrow of `tx.base_workspace_helper()` which can't overlap with `CommitBuilder` since it borrows the repo mutably.

torquestomp: Ah, that's convenient.  Added a commit, done.

lib/src/rewrite.rs
torquestomp: Not possible anymore with the callback.

yuja: It's probably easier to pass a callback by function argument. `Box<dyn` tends to make things complicated.

torquestomp: Done.

lib/src/rewrite.rs
```diff
@@ -830,3 +830,150 @@ pub fn move_commits(
...
+
+pub struct CommitToSquash {
+    commit: Commit,
+    selected_tree: MergedTree,
+    parent_tree: MergedTree,
+}
+
+impl CommitToSquash {
+    pub fn new(repo: &dyn Repo, commit: Commit, selected_tree: MergedTree) -> BackendResult<Self> {
+        let parent_tree = commit.parent_tree(repo)?;
```
yuja, line 842: nit: `parent_tree()` might have to merge trees. Maybe it's okay to just make all `CommitToSquash` fields public and let callers fill out?

lib/src/rewrite.rs
```diff
@@ -830,3 +830,150 @@ pub fn move_commits(
...
+    }
+}
+
+pub type SquashedDescriptionFn<'a, E> =
+    Box<dyn FnOnce(&[&CommitToSquash]) -> Result<String, E> + 'a>;
+
+pub struct SquashOptions<'a, E> {
+    pub keep_emptied: bool,
+    pub description_fn: SquashedDescriptionFn<'a, E>,
+}
```
yuja, line 879: nit: unused types

yuja (approved):
LGTM, thanks.

PR #4602, opened 2024-10-07
templates: make `op_summary` in e.g. `jj undo` output more readable and colorful

ilyagr:
This is meant to be a quick fix before I update the demos in the README. The primary reason to update them is the branch -> bookmark transition, but I noticed that `jj undo`'s output is currently not very pretty.

I believe printing both the start and end times is excessive for a summary. For a long-running operation like `jj describe` with the editor open for a long time, the ending time seems more meaningful.

I was a bit torn on whether to use `format_timestamp(self.time().end())` or `self.time().end().ago()`. The latter looks better (`finished 2 seconds ago` instead of `finished 2024-10-07 15:46:03`), but is less configurable and worse for dates long ago. In the future, we could add a `format_op_summary_timestamp` function and/or a template function that uses `.ago()` for recent dates and absolute dates for old dates.

Previous look:
Pasted image 20241006191123

Or in the demo:
image

**New look**:

image

<details><summary>Previous version</summary>

image

Pasted image 20241007143128

</details>

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -41,10 +41,20 @@ tag_list = '''
 label("tag", name) ++ format_ref_targets(self) ++ "\n"
 '''
 
+# TODO: Here and elsewhere, the end time for operations is more meaningful than
+# the start time. https://github.com/martinvonz/jj/pull/4602#discussion_r1791207491
 op_summary = '''
-if(root,
-  separate(" ", self.id().short(), label("root", "root()")),
-  separate(" ", self.id().short(), format_time_range(self.time()), self.description().first_line())
+separate(" ",
+  self.id().short(),
+  if(root,
+    label("root", "root()"),
+    surround(
+       "(",
+       ")",
+       format_timestamp(self.time().start())
+    )
+  ),
+  self.description().first_line()
 )
 '''
 
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1492,6 +1492,12 @@ impl WorkspaceCommandHelper {
             &self.op_summary_template_text,
             OperationTemplateLanguage::wrap_operation,
         )
+        // TODO: "op_log" is not a great label name for op_summary. We could
+        // replace it with `operation summary` and use `operation` in
+        // colors.toml. TODO: It would be better to put labels like these
+        // consistently in the same place, e.g. `templates.toml` instead of the
+        // rust code.
+        .labeled("op_log")
     }
 
     pub fn short_change_id_template(&self) -> TemplateRenderer<'_, Commit> {
--- a/cli/src/commands/operation/diff.rs
+++ b/cli/src/commands/operation/diff.rs
@@ -125,14 +125,12 @@ pub fn cmd_op_diff(
     let op_summary_template = workspace_command.operation_summary_template();
     ui.request_pager();
     let mut formatter = ui.stdout_formatter();
-    formatter.with_label("op_log", |formatter| {
-        write!(formatter, "From operation ")?;
-        op_summary_template.format(&from_op, &mut *formatter)?;
-        writeln!(formatter)?;
-        write!(formatter, "  To operation ")?;
-        op_summary_template.format(&to_op, &mut *formatter)?;
-        writeln!(formatter)
-    })?;
+    write!(formatter, "From operation ")?;
+    op_summary_template.format(&from_op, &mut *formatter)?;
+    writeln!(formatter)?;
+    write!(formatter, "  To operation ")?;
+    op_summary_template.format(&to_op, &mut *formatter)?;
+    writeln!(formatter)?;
 
     show_op_diff(
         ui,
--- a/cli/src/commands/operation/diff.rs
+++ b/cli/src/commands/operation/diff.rs
@@ -125,10 +125,10 @@ pub fn cmd_op_diff(
     let op_summary_template = workspace_command.operation_summary_template();
     ui.request_pager();
     let mut formatter = ui.stdout_formatter();
-    write!(formatter, "From operation ")?;
+    write!(formatter, "From operation: ")?;
     op_summary_template.format(&from_op, &mut *formatter)?;
     writeln!(formatter)?;
-    write!(formatter, "  To operation ")?;
+    write!(formatter, "  To operation: ")?;
     op_summary_template.format(&to_op, &mut *formatter)?;
     writeln!(formatter)?;
 
--- a/cli/src/commands/operation/restore.rs
+++ b/cli/src/commands/operation/restore.rs
@@ -56,7 +56,7 @@ pub fn cmd_op_restore(
     );
     tx.repo_mut().set_view(new_view);
     if let Some(mut formatter) = ui.status_formatter() {
-        write!(formatter, "Restored to operation ")?;
+        write!(formatter, "Restored to operation: ")?;
         let template = tx.base_workspace_helper().operation_summary_template();
         template.format(&target_op, formatter.as_mut())?;
         writeln!(formatter)?;
--- a/cli/src/commands/operation/undo.rs
+++ b/cli/src/commands/operation/undo.rs
@@ -69,7 +69,7 @@ pub fn cmd_op_undo(
     );
     tx.repo_mut().set_view(new_view);
     if let Some(mut formatter) = ui.status_formatter() {
-        write!(formatter, "Undid operation ")?;
+        write!(formatter, "Undid operation: ")?;
         let template = tx.base_workspace_helper().operation_summary_template();
         template.format(&bad_op, formatter.as_mut())?;
         writeln!(formatter)?;
--- a/cli/tests/test_duplicate_command.rs
+++ b/cli/tests/test_duplicate_command.rs
@@ -76,7 +76,7 @@ fn test_duplicate() {
     let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["undo"]);
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r#"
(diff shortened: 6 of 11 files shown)


cli/src/config/templates.toml
```diff
@@ -41,10 +41,24 @@ tag_list = '''
 label("tag", name) ++ format_ref_targets(self) ++ "\n"
 '''
 
+# TODO: "op_log" is not a great label name for op_summary. We could replace it
+# with `operation` and insert this new label throughout the op log templates.
```
yuja, line 45: Yes. Maybe we can insert `op_summary`/`commit_summary` labels if needed. That's the current convention.

cli/src/config/templates.toml
```diff
@@ -41,10 +41,24 @@ tag_list = '''
...
-  separate(" ", self.id().short(), format_time_range(self.time()), self.description().first_line())
+label(
+  "op_log",
+  concat(
+    self.id().short(),
+    if(!root, concat(
+       ", ",
+       label("time", "finished "),
+       format_timestamp(self.time().end())
+    )),
```
yuja, line 55: nit: I would omit "finished " because it's shorter and I don't care whether the timestamp is start or end. As I don't care, it might be better to use the start time. I think the start time is displayed in `op log` if you disabled relative timestamps.

ilyagr: >  I don't care whether the timestamp is start or end. As I don't care, it might be better to use the start time. I think the start time is displayed in op log if you disabled relative timestamps.

Is there a situation you are thinking of where the start is more useful than the end?

My current feeling is that I do care somewhat, and I think that when it matters, the end is more important. It's the moment where the repository was actually modified with the result of the operation. For example, as I mentioned above, if I was writing a description in `jj describe` for an hour, saved it, and then undid it, it'd be easier for me to remember the end moment. Another way to see that it's important is that aborted operations have a start time but no end time, and lacking the end time is related to the fact that they are not important.

That could be considered a separate issue from this PR, though. I thought we display both the start and end time in the op log, but it might make sense to display either the time ago + duration or the ending time + duration.

>  I would omit "finished " because it's shorter

I could say "ended" or "from" if you prefer shorter. This brings us back to the recurring argument where I like using English words and you do not like it.

Not using English would make more sense to me if we didn't have "Undid operation" before the summary, but something like "Operation undone:". This would require making changes to several commands, which I'd prefer to consider outside of the scope of this PR.

---------------------

If none of these work for you, I'll just adjust this to your liking tomorrow (I'm not finishing it today in any case). We can change it later, ultimately I just want something that looks OK in the demo.

cli/src/config/templates.toml
ilyagr: That sounds good. I'd prefer to leave this cleanup to some later time, though.

ilyagr: You mean that `format_operation` and related macros would also add the `op_summary` label? I thought we'd want one label to be used both there and in `op_summary` so that we don't need to add many more colors to `colors.toml`.

Ultimately, either is fine by me, but if I understand the plan, I can adjust the TODO accordingly.

cli/src/config/templates.toml
yuja: > Is there a situation you are thinking of where the start is more useful than the end?

No. I'm just saying that using start time is more consistent because it's displayed in "op log". Maybe we can change "op log" to use the end time?

> > I would omit "finished " because it's shorter
> 
> I could say "ended" or "from" if you prefer shorter. This brings us back to the recurring argument where I like using English words and you do not like it.
> 
> Not using English would make more sense to me if we didn't have "Undid operation" before the summary, but something like "Operation undone:". This would require making changes to several commands, which I'd prefer to consider outside of the scope of this PR.

I prefer not using English word there because the summary template can be embedded in various places. I think the intent of "Undid operation" message is `Undid operation <quote>the-operation</quote>`.

cli/src/config/templates.toml
yuja: I mean the label should be inserted by caller, not by template. (See `cmd_op_log()` for instance.)

Anyway, if we remove `"finished "`, I think we can remove `label("op_log", ...)` too.

cli/src/config/templates.toml
ilyagr: > Maybe we can change "op log" to use the end time?

Yeah, I think that might be good. Might be a separate PR.

> I think the intent of "Undid operation" message is Undid operation <quote>the-operation</quote>.

I'll experiment with `Undid operation: op_summary_template` tomorrow.

cli/src/config/templates.toml
ilyagr: > Anyway, if we remove "finished ", I think we can remove label("op_log", ...) too.

I don't follow. Without `op_log`, the labels from `colors.toml` aren't triggered and there is no color at all. Somehow, there has to be a tag that says that we have an *operation* id here, as opposed to a *change* id or a *commit* id.

yuja: > > Anyway, if we remove "finished ", I think we can remove label("op_log", ...) too.
> 
> I don't follow. Without `op_log`, the labels from `colors.toml` aren't triggered and there is no color at all.

Oh, I didn't notice this was the fix for the existing problem. Inserting `operation` label somewhere sounds also good to me (to denote the `self` type is `Operation`.)

ilyagr: Yeah, I had a hard time stopping at the bug fix when I saw other things that could be better.

I belatedly split this into two commits.

cli/src/config/templates.toml
ilyagr: "tomorrow" turned out to be optimistic, but here it is.

I changed the times to the start time for now, but I intend to change it in a separate PR.

cli/src/config/templates.toml
yuja, line 1: nit: can you add some `--color=always`|`debug` snapshot to ensure this won't break?

cli/src/config/templates.toml
```diff
@@ -41,10 +41,15 @@ tag_list = '''
...
 op_summary = '''
-if(root,
-  separate(" ", self.id().short(), label("root", "root()")),
-  separate(" ", self.id().short(), format_time_range(self.time()), self.description().first_line())
+label(
+  "op_log",
+  if(root,
+    separate(" ", self.id().short(), label("root", "root()")),
+    separate(" ", self.id().short(), format_time_range(self.time()), self.description().first_line())
+  )
```
yuja, line 64: nit: This can be added to `WorkspaceCommandHelper::operation_summary_template()` instead.
```rust
self.reparse_valid_template(...).labeled("op_log")
```

cli/src/config/templates.toml
```diff
@@ -42,13 +42,25 @@ label("tag", name) ++ format_ref_targets(self) ++ "\n"
...
-  if(root,
-    separate(" ", self.id().short(), label("root", "root()")),
-    separate(" ", self.id().short(), format_time_range(self.time()), self.description().first_line())
+  concat(
+    self.id().short(),
+    if(!root, concat(
+       " (",
+       format_timestamp(self.time().start()),
+       ")"
+    )),
```
yuja, line 58: nit: perhaps, we can use `separate()` and/or `surround()` to deal with whitespace
```
separate(" ",
  self.id().short(),
  if(root, .., ..),
  self.description().first_line(),
)
```

yuja (approved):
Thanks!

cli/src/config/templates.toml
ilyagr: Sure.

In the longer term, I think it'd be nice to move all of these into `templates.toml` so that all the labels are in one place. Right now, I wouldn't know where to look (the debug output, I suppose).

cli/src/config/templates.toml
ilyagr: Ah, forgot that `surround` existed, great!

cli/src/config/templates.toml
ilyagr: It seems that `op diff` already inserted the `op_log` template. I'll make it so that it only gets inserted once, but this reinforces my desire to put all the labels in one place.

ilyagr:
I'll keep this unmerged for a few hours in case somebody has a comment or wants me to wait.

cli/src/config/templates.toml
```diff
@@ -41,10 +41,20 @@ tag_list = '''
...
-  separate(" ", self.id().short(), format_time_range(self.time()), self.description().first_line())
+separate(" ",
+  self.id().short(),
+  if(!root, surround(
+     "(",
+     ")",
+     format_timestamp(self.time().start())
+  )),
+  if(root, 
+     label("description", "root()"),
```
yuja, line 55: nit: I would add `"root"` label because `"root()"` isn't a description. I'll also move `description.first_line()` out of `if()`.

cli/src/config/templates.toml
yuja: > In the longer term, I think it'd be nice to move all of these into templates.toml

The idea is that templates can be customized by user, but the outermost label shouldn't overridden by such changes.

yuja (commented):
Thanks.

cli/src/config/templates.toml
ilyagr: I see. Maybe it's possible to consistently move all the labels to functions like `operation_summary_template` where they'd all be in the same place. Or, we could have `_operation_summary_template_internal` template that wraps `operation_summary_template` in the label.

cli/src/config/templates.toml
ilyagr: Good point, I missed the label. Seems like root has an empty description, so that sounds good.

cli/src/config/templates.toml
yuja: I would avoid adding abstraction layers in the template configuration. It's easier to do in code so long as the requirement isn't dynamic.

PR #4605, opened 2024-10-08
feat: Add support for *args to AliasMap

matts1:
This is required in order to support command aliases as functions, because they will require the following to be supported:
```toml
'upload()' = ["upload", "@"]
'upload(r)' = [["fix", "$r"], ["git", "push", "--change", "$r"]]
```

Here, `jj upload` would need to map to the first, `jj upload foo` would need to map to the second, and `jj upload foo --bar` would also need to map to the second.

# Checklist

If applicable:

### DIFF
--- a/lib/src/dsl_util.rs
+++ b/lib/src/dsl_util.rs
@@ -458,14 +458,18 @@ impl<P, V> AliasesMap<P, V> {
                 self.symbol_aliases.insert(name, defn.into());
             }
             AliasDeclaration::Function(name, params) => {
-                let overloads = self.function_aliases.entry(name).or_default();
-                match overloads.binary_search_by_key(&params.len(), |(params, _)| params.len()) {
-                    Ok(i) => overloads[i] = (params, defn.into()),
-                    Err(i) => overloads.insert(i, (params, defn.into())),
-                }
+                self.insert_function(name, params, defn.into());
             }
         }
         Ok(())
+    }
+
+    fn insert_function(&mut self, name: String, params: Vec<String>, defn: V) {
+        let overloads = self.function_aliases.entry(name).or_default();
+        match overloads.binary_search_by_key(&params.len(), |(params, _)| params.len()) {
+            Ok(i) => overloads[i] = (params, defn),
+            Err(i) => overloads.insert(i, (params, defn)),
+        }
     }
 
     /// Iterates symbol names in arbitrary order.
@@ -490,6 +494,18 @@ impl<P, V> AliasesMap<P, V> {
     pub fn get_function(&self, name: &str, arity: usize) -> Option<(AliasId<'_>, &[String], &V)> {
         let overloads = self.get_function_overloads(name)?;
         overloads.find_by_arity(arity)
+    }
+
+    /// Looks up a function alias by name and arity, assuming that the function
+    /// can take extra parameters (eg. *args in python).
+    /// Returns identifier, list of parameter names, and definition text.
+    pub fn get_function_with_leftovers(
+        &self,
+        name: &str,
+        arity: usize,
+    ) -> Option<(AliasId<'_>, &[String], &V)> {
+        let overloads = self.get_function_overloads(name)?;
+        overloads.find_by_arity_with_leftovers(arity)
     }
 
     /// Looks up function aliases by name.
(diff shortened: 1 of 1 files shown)


yuja:
> This is required in order to support command aliases as functions, because they will require the following to be supported:
> 
> ```toml
> 'upload()' = ["upload", "@"]
> 'upload(r)' = [["fix", "$r"], ["git", "push", "--change", "$r"]]
> ```
> 
> Here, `jj upload` would need to map to the first, `jj upload foo` would need to map to the second, and `jj upload foo --bar` would also need to map to the second.

These examples don't look like `*args`. (If they were, `upload(*args)` should conflict with `upload(r, *args)`.)

Perhaps, we'll need a public API to get `AliasFunctionOverloads`?

```rust
let overloads = map.get_function_overloads("upload")?;
// for `jj upload foo --bar`, max_parameters = ["foo"].len()
// XXX should we pick upload/1 for `jj upload foo bar`?? It might be rather confusing.
let arity = max(overloads.max_arity(), max_parameters);
// If there were upload/0, /1, /3, and max_parameters is 2, I think it's better
// to error out (as ambiguous alias definitions) than falling back to upload/1.
let function = overloads.find_by_arity(arity)?;
```

yuja:
> ```toml
> 'upload()' = ["upload", "@"]
> 'upload(r)' = [["fix", "$r"], ["git", "push", "--change", "$r"]]
> ```
> 
> Here, `jj upload` would need to map to the first, `jj upload foo` would need to map to the second, and `jj upload foo --bar` would also need to map to the second.

Oh, but `jj upload foo --bar` will have to be rewritten to `[["fix", "foo"], ["git", "push", "--change", "foo", "--bar"]]`? Then, we'll need an explicit `*args` syntax to specify which command list to copy the remainder arguments to, and only one of the overloads should have `*args`.

PhilipMetzger:
just a minor nit from my side, "feature" is not a known topic, it could be `dsl_util:` or `aliases:`.

matts1:
> Oh, but `jj upload foo --bar` will have to be rewritten to `[["fix", "foo"], ["git", "push", "--change", "foo", "--bar"]]`? Then, we'll need an explicit `*args` syntax to specify which command list to copy the remainder arguments to, and only one of the overloads should have `*args`.

My original intention was that every overload would have an implicit *args, so given, and:
```
'upload()' = ["upload", "@"]
'upload(r, blah)' = [["fix", "$r"], ["git", "push", "--change", "$r", "$blah", "$args"]]
```

```
upload() ==> upload/0(args=[])
upload(first) ==> upload/0(args=[first])
upload(first, second) ==> upload/2(first, second, args=[])
upload(first, second, third) ==> upload/2(first, second, args=[third])
```

My logic was going to be:
* If the overload goes to a single command, add an implicit `$args` at the end if it doesn't exist
* If the overload goes to multiple commands, `$args` was non-empty, and none of the commands ended with `$args`, then error out.

Your proposal does work as well, I mostly wanted to avoid the `*args` in the interest of slimming down the syntax. However, if we're ok with it, then I can implement it in the following way:
```
'upload()' = ["upload", "@"]
'upload(r, *args)' = [["fix", "$r"], ["git", "push", "--change", "$r" "$args"]]
```

This would generate an `upload/0` and an `upload/1+`.

yuja:
> My original intention was that every overload would have an implicit `*args` and we perform a longest prefix match, so given, and:
> 
> ```
> 'upload()' = ["upload", "@"]
> 'upload(r, blah)' = [["fix", "$r"], ["git", "push", "--change", "$r", "$blah", "$args"]]
> ```
> 
> ```
> upload() ==> upload/0(args=[])
> upload(first) ==> upload/0(args=[first])
> upload(first, second) ==> upload/2(first, second, args=[])
> upload(first, second, third) ==> upload/2(first, second, args=[third])
> ```

My feeling is that choosing a prefix match would lead to random unexpected result. I would rather want to see an error if overloads set is somewhat inconsistent.

BTW, the example above has recursion? `upload/0` -> (`upload/1` ->) `upload/0`
(`["upload"]` might be substituted to `["upload", "@", "@"]`, but I'm not sure.)

> Your proposal does work as well, I mostly wanted to avoid the `*args` in the interest of slimming down the syntax since concern was expressed over making the DSL more complex. However, if we're ok with it, then I can implement it in the following way:
> 
> ```
> 'upload()' = ["upload", "@"]
> 'upload(r, *args)' = [["fix", "$r"], ["git", "push", "--change", "$r" "$args"]]
> ```
> 
> This would generate an `upload/0` and an `upload/1+`.

I don't have a strong feeling about whether `*args` can be implicit or not, but if it can, I would add it only to the function having the maximum arity.

PR #4613, opened 2024-10-09
build: add `rust-toolchain.toml`

arxanas:
Follow-up from discussion at https://discord.com/channels/968932220549103686/1288926971719323762

I don't think we achieved consensus in that thread. We could use `stable` or `nightly` here instead if we prefer. Note that user `rustup` overrides are still respected in the presence of a `rust-toolchain.toml`.

# Checklist

If applicable:

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -133,9 +133,7 @@ recommended steps.
 One-time setup:
 
     rustup toolchain add nightly  # wanted for 'rustfmt'
-    cargo install cargo-insta
-    cargo install cargo-watch
-    cargo install cargo-nextest
+    cargo install cargo-insta cargo-watch cargo-nextest
 
 During development (adapt according to your preference):
 


yuja:
For reference, there's a previous attempt: https://github.com/martinvonz/jj/pull/1913

(I personally don't like `rust-toolchain.toml` because it forces `rustup` to install the specified toolchain if not exist.)

arxanas:
One reason in favor of `rust-toolchain.toml` (failed to mention it in the commit message) is that otherwise you might have to reconfigure the version of Rust to use for every new worktree.

Besides worktrees, I often use different machines, including ephemeral cloud machines that don't have my previous overrides configured. It's annoying to have to manually configure the Rust version every time I freshly check out the repository. I'm open to other solutions.

(Of course, if you specifically _do_ want to use the default Rust toolchain for every worktree, then this gets in the way of that. Maybe setting `stable` as the version helps to avoid forcing a download unnecessarily?)

ilyagr:
If we go with this, could you add an explanation somewhere of how to override it?

Is there a way to make `cargo fmt` to default to `+nightly`?

> It's annoying to have to manually configure the Rust version every time I freshly check out the repository. I'm open to other solutions.

I'd just do `cargo +nightly` (or stable, but you still need nightly for `cargo fmt`), `cargo +1.76`, `rustup default nightly`, or `rustup default 1.76`.

I wish there was a way to define a `+ci` version for the project in `rust-version.toml` instead of making it the default (so that people don't have to look up the 1.76). I'm guessing this is not possible?

arxanas:
> If we go with this, could you add an explanation somewhere of how to override it? Update: More generally, I think a few updates to contributing.md would have to go with it, e.g. https://martinvonz.github.io/jj/prerelease/contributing/#summary.

👍

Looking at the docs now, I think adding `rust-toolchain.toml` simplifies the general instructions slightly.

> I wish there was a way to define a +ci version for the project in rust-version.toml instead of making it the default (so that people don't have to look up the 1.76). I'm guessing this is not possible?

I think this would be ideal but I couldn't find a way to do it.

> I'd just do cargo +nightly (or stable, but you still need nightly for cargo fmt), cargo +1.76, rustup default nightly, or rustup default 1.76.

Hmm, but then I'd either download new nightly toolchains as they come out, or I'd have to update all invocations of Rust commands with the new version.
- [assessment: annoying] Having to update all the possible call-sites seems particularly annoying/difficult (as I'm sure I have many aliases, plus any IDE support needs to be updated).
- [assessment: less annoying] Downloading the new toolchain whenever it's released is maybe less bad, but I'd still prefer to download it only once when I start working on the project on a given machine.
  - [assessment: not a serious concern] If we wanted to support Codespaces/etc., it might be easier in practice to use a static Rust toolchain version to boostrap the image, rather than dynamically provision the image with nightly or defer the download to development-time.

ilyagr:
A couple of links to relevant issues. It doesn't look like either of them is going to be solved imminently, which is a shame, it would have helped a lot.

https://github.com/rust-lang/rustup/issues/3546#issuecomment-1931848506

https://github.com/rust-lang/rustup/issues/1484

arxanas:
Need to update docs

arxanas:
Now that I'm revisiting this PR, I'm still in favor of adding `rust-toolchain.toml` as per [my previous comment in the thread](https://github.com/martinvonz/jj/pull/4613#issuecomment-2402420853) (unlike https://github.com/martinvonz/jj/pull/1913 and https://github.com/martinvonz/jj/pull/4749, which have been closed for other reasons).

> @ilyagr: If we go with this, could you add an explanation somewhere of how to override it? Update: More generally, I think a few updates to contributing.md would have to go with it, e.g. https://martinvonz.github.io/jj/prerelease/contributing/#summary.

I updated the docs. It becomes simpler in the common case (no need to select a toolchain manually), and then I added a section detailing how to explicitly override it.

> @yuja: (I personally don't like rust-toolchain.toml because it forces rustup to install the specified toolchain if not exist.)

I think that's a valid concern, but I still much prefer the Bazel workflow where the build tool downloads everything (and it's annoying), but then everything works much more reliably after that. You can still do `rustup override set` if you want to use your local version. (I added this to the docs, as per above.)

Besides my previous personal reasoning, I think it's also likely that some drive-by contributors won't even read `contributing.md`, but simply clone and start running `cargo` commands, so it's nice to guide them into most reliable configuration by default.

docs/contributing.md
```diff
@@ -133,24 +133,23 @@ recommended steps.
...
-    cargo install cargo-watch
-    cargo install cargo-nextest
+    cargo install cargo-insta cargo-watch cargo-nextest
 
 During development (adapt according to your preference):
 
     cargo watch --ignore '.jj/**' -s \
       'cargo clippy --workspace --all-targets \
-       && cargo +1.76 check --workspace --all-targets'
+       && cargo check --workspace --all-targets'
```
arxanas, line 142: comment (non-blocking): It occurs to me that the old instructions here were probably wrong or suboptimal since `cargo clippy` and `cargo check` were potentially running with different compiler versions, so they might not have reused the same build artifacts, or you could be getting too-old or too-new `clippy` diagnostics.

senekor:
I'm not against this per se, I opened [one of the other PRs](https://github.com/martinvonz/jj/pull/4749) to add this but closed it again. I guess what's missing here for me is the justification why the MSRV is the preferrable default for development. As stated in [this comment](https://github.com/martinvonz/jj/pull/1913#issuecomment-2037636332), it's not necessary to run an outdated toolchain to catch MSRV violations earlier than CI.

The one additional argument I can see in this thread is:

> that otherwise you might have to reconfigure the version of Rust to use for every new worktree.

Which is true, but it's still missing an initial motivation for why MSRV should be the preferred default in the first place.

senekor:
> I think it's valuable to have "deterministic" builds. [...] Is this an ideal you share?

Not really, but this is a reason I can accept.

Deterministic builds are nice when shipping to users, but the idealist in me would prefer to solve the problems you describe at their root. The Rust compiler is supposed to be backwards compatible. Your build is not supposed to fail when you upgrade to a new stable release. (Sometimes new versions reject unsound programs that were previously accepted erroneously, that's actually a situation where you _want_ your build to fail.) If libraries are causing the issue, it can be nice to report these issues upstream. I realize that it's not _our_ main goal to contribute to our libraries, but I personally think it's a matter of courtesy in the open source community. And anyway, if one of our libraries doesn't have its engineering practices under control, wouldn't we _want_ to know about it and maybe try to improve that?

I'm always a little saddened by blunt force solutions to hard engineering problems.
- security -> put everyhing behind a VPN
- messy dev environment or not platform agnostic -> stuff everything in container
- flaky builds -> pin versions of everything

But those are just my "feelings". I'm fine with this change.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.76"
+profile = "default"
```
senekor, line 3:
```suggestion
[toolchain]
channel = "1.76"
profile = "default"
components = ["rust-analyzer"]
```

Otherwise LSP will be broken for people who use rust-analyzer provided by rustup.

arxanas:
> I realize that it's not our main goal to contribute to our libraries, but I personally think it's a matter of courtesy in the open source community. And anyway, if one of our libraries doesn't have its engineering practices under control, wouldn't we want to know about it and maybe try to improve that?

[additional context, not a specific argument] One common situation is not even a "problem" or "mistake" situation, but rather the upstream library has deliberately started using features from a later MSRV than ours (either directly or transitively), and we need to be careful to stay at an older version until we can upgrade our own MSRV.
- IIUC, the `cargo` dependency resolver doesn't yet take into account dependency MSRV. Hopefully, once that's implemented, it'll significantly alleviate this category of problem. But, in the meantime, I would definitely like to avoid the above situation.
- It's incredibly frustrating to have to rework the resolved dependencies later (oftentimes by hand), so surfacing that kind of error as early as possible in the development cycle can save a lot of time.
- In the worst case, I'll have started using and building on top of a critical feature from a dependency that's not available in the most recent MSRV-compatible version of the library.

rust-toolchain.toml
arxanas: Added it, thanks for letting me know. I'm assuming that this results in unnecessary downloads for people who use `rust-analyzer` from other sources (I think my VS Code extension provisions it itself?), but that's still in line with the trade-off I'm trying to make in this PR.

yuja:
Just to be clear, I don't like `rust-toolchain.toml` (because it forces me to install random toolchains without consent), but I'm not against adding it if that's a common practice.

BTW, does it help if we add `rust-toolchain.toml.recommended` (which the user can add symlink at their option)?

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
...
+components = ["rust-analyzer"]
```
ilyagr, line 4: Is it possible to make "nightly" the default toolchain for `rustfmt`/`cargo fmt`?

ilyagr: Looking at https://github.com/rust-lang/rustup/issues/3546#issuecomment-1931848506 again (I already linked to it in https://github.com/martinvonz/jj/pull/4613#issuecomment-2405901050), probably not :(

ilyagr:
Another option that I would mildly prefer would be to bump our MSRV to at least 1.78 and try enabling the lint from https://github.com/martinvonz/jj/pull/1913#issuecomment-2037636332 first.

I'm having a hard time telling what's good for other people here. I personally would prefer if the rust ecosystem provided [a better solution to this](https://github.com/martinvonz/jj/pull/4613#issuecomment-2405901050), the current capabilities of `rust-toolchain.toml` are not a great fit for my workflow. For me, this swaps one set of inconveniences for another set. But perhaps it would help new people or most people. I don't think any of my inconveniences are bad enough to be blocking.

rust-toolchain.toml
senekor: One thing we could do is add editor configuration to the repo which overrides the formatting command. That would be a bunch more files though and I assume some people won't like it.

senekor:
> BTW, does it help if we add rust-toolchain.toml.recommended (which the user can add symlink at their option)?

I don't think that solves the issue with workspaces. The symlink would have to be added manually to each.

A bit of a tangent: Where do people put their workspaces, usually? I have my repo organization scheme `~/repos/$host/$owner/$repo/` and I dont know how workspaces fit into that. I've recently considered something like `.jj_workspaces/` right in the repo and put that in my global gitconfig. That would actually work with rustup directory overrides, right? If you override the toolchain for a repo once, the workspaces should pick it up as well, since they are in a subdirectory.

Second tangent: Why do we have an MSRV in the first place? I don't remember the reason being documented anywhere. Are there known downstream users who rely on this? I'm asking because it seems to add a bunch more friction than I would've imagined.

arxanas:
> A bit of a tangent: Where do people put their workspaces, usually? I have my repo organization scheme ~/repos/$host/$owner/$repo/ and I dont know how workspaces fit into that. I've recently considered something like .jj_workspaces/ right in the repo and put that in my global gitconfig. That would actually work with rustup directory overrides, right? If you override the toolchain for a repo once, the workspaces should pick it up as well, since they are in a subdirectory.

I usually put new worktrees in the same directory (so everything ends up going into `~/Workspace` for me, including worktrees for unrelated proejcts). But I could probably modify my workflow to put all the project worktrees into a project-specific directory. (I didn't know/realize that you could `rustup override` for arbitrary directories that don't have a Rust project/`Cargo.toml` in them.)

> Second tangent: Why do we have an MSRV in the first place? I don't remember the reason being documented anywhere. Are there known downstream users who rely on this? I'm asking because it seems to add a bunch more friction than I would've imagined.

I believe it's primarily to support package maintainers, because they're limited to using the `rustc` version that's part of their distribution, so it's a problem when newer versions of jj can't be built by the older `rustc` in their distribution. I don't know if we have actually enumerated the package maintainers/distributions we want to support (which would also determine when we can upgrade the MSRV).

yuja:
> "$HOME" = "stable-x86_64-unknown-linux-gnu"

I wouldn't want to ignore toolchain settings for peculiar environments (e.g. cortex-m4a), but yeah, it's good to know.

>  if we add rust-toolchain.toml.recommended (which the user can add symlink at their option)?

FWIW, what I meant here is that user can do `ln -s rust-toolchain.toml.recommended rust-toolchain.toml` in jj repo to opt in to the toolchain configuration (just like we have .envrc.recommended.)

arxanas:
> Just to be clear, I don't like rust-toolchain.toml (because it forces me to install random toolchains without consent), but I'm not against adding it if that's a common practice.

Re common practice, from the [top 30 Rust GitHub repositories ordered by stars](https://github.com/search?q=language%3Arust++stars%3A%3E1000+&type=repositories&s=stars&o=desc) (arbitrary sample size), it looks like 14 of them have a `rust-toolchain.toml` file:

- https://github.com/denoland/deno
- https://github.com/FuelLabs/fuel-core
- https://github.com/zed-industries/zed
- https://github.com/meilisearch/meilisearch
- https://github.com/dani-garcia/vaultwarden
- https://github.com/lapce/lapce
- https://github.com/helix-editor/helix
- https://github.com/astral-sh/ruff
- https://github.com/nushell/nushell
- https://github.com/swc-project/swc
- https://github.com/pola-rs/polars
- https://github.com/astral-sh/uv
- https://github.com/influxdata/influxdb
- https://github.com/servo/servo

rust-toolchain.toml
arxanas: I'm not personally opposed to adding per-editor configuration to the repo, but I'll leave it as a potential follow-up.

arxanas:
For future readers: This was reverted in https://github.com/jj-vcs/jj/pull/5090 due to causing an issue with building musl binaries in the release (don't know why).

ilyagr:
MSRV-aware Cargo is available starting with Rust 1.84: <https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html> (it does not strictly require using the 2024 edition, but is implied by using the 2024 edition). Hopefully, we'll move to that in a few months.

Issue #1283, opened 2023-02-18
FR: revset syntax to resolve an expression at an old operation


ilyagr:
It would be nice to have something like `xyz#a5b` to represent the commit that the change `xyz` was at operation `a5b`. So, one could do `jj log --at-op a5b` and follow that up with `jj restore --from xyz#a5b`.

martinvonz:
`xyz@a5b` is another option, but then we'd probably need to replace the use of `@` for `branch@remote` syntax.

martinvonz:
If we choose `@` for the operator, we'd probably also want to remove it from the allowed characters in an identifier: https://github.com/martinvonz/jj/blob/92f9fe5a1b12035e52866fd8abcae60b9e86bea9/lib/src/revset.pest#L15-L19

martinvonz:
But since we use `@` for the working-copy commit, that's doesn't seem like a good idea :) So never mind, `#` seems much better

ilyagr:
I also don't particularly like `#`, but I haven't thought of a better option.

One ambiguity is whether @#abc6 should be the commit that was current at abc6 or the commit with the same change id as @ at abc6. I would probably want the former.

ilyagr:
This would also be quite useful for branches.

yuja:
So `expr#op` will be the operator to evaluate `expr` within the environment at `op`. If we can't find a better operator character, we can try with a function syntax first.

ilyagr:
A related FR would be a "universal reflog", a way to see or use the "previous value" of something like "main@origin" as something like "main@origin#1". This would ignore any past operations that didn't change the value of "main@origin". (of course, we'd need a different symbol or syntax if "main@origin#1" treats 1 as an operation id).

I think filtering by revset could already be implemented as an option to `jj op log` (similarly to filtering by file for `jj log`), but that's a separate feature and would be more awkward to use.

PR #4614, opened 2024-10-09
revset: add at_operation(op, expression)

yuja:
#1283

# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_revset.rs
+++ b/lib/tests/test_revset.rs
@@ -949,6 +949,13 @@ fn test_resolve_symbol_git_refs() {
 }
 
 fn resolve_commit_ids(repo: &dyn Repo, revset_str: &str) -> Vec<CommitId> {
+    try_resolve_commit_ids(repo, revset_str).unwrap()
+}
+
+fn try_resolve_commit_ids(
+    repo: &dyn Repo,
+    revset_str: &str,
+) -> Result<Vec<CommitId>, RevsetResolutionError> {
     let settings = testutils::user_settings();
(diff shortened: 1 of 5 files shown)


lib/tests/test_revset.rs
```diff
@@ -2894,6 +2899,134 @@ fn test_evaluate_expression_committer() {
...
+    );
+
+    // Operation is resolved relative to the outer ReadonlyRepo.
+    assert_eq!(
+        resolve_commit_ids(repo2.as_ref(), "at_operation(@-, at_operation(@-, all()))"),
+        resolve_commit_ids(repo2.as_ref(), "at_operation(@--, all())")
+    );
+
+    // It's unclear whether "@" should point to the uncommitted operation or
+    // not, but it's resolved to the base operation right now.
```
martinvonz, line 2986: It seems to me that `at_operation(@, x)` should be equivalent to `x` as far as possible

martinvonz (approved):
Nice! Thanks for all your work on this feature.

lib/tests/test_revset.rs
yuja: I tend to agree, but I said "it's unclear" because a `MutableRepo` doesn't represent an `Operation`. Implementation-wise, it's not easy to resolve `@` (or `@-+`) to a `MutableRepo` because of that. In practice, it's unlikely to evaluate `at_operation()` against `MutableRepo`.

martinvonz: Sorry, i didn't mean that as blocking.

Issue #4616, opened 2024-10-09
jj handles renames of the remote trunk poorly


tonyfinn:
## Steps to Reproduce the Problem

1. Create a git repo with a `master` branch
2. Push the git repo to a remote
3. Run `jj git init --colocate` in this copy
4. Run `jj branch track master@origin` in this copy
5. From a different checkout of the git repo (or a different device/user), rename the branch `master` to `main`, push the new `main` branch to the remote, delete `master` on the remote
6. Run `jj git fetch` on the first checkout
7. Attempt to switch to the new `main` branch with `jj new main`

## Expected Behavior

Ideally: You're able to continue developing the on local repo with the new name for the trunk branch. 

At a minimum: Warn the target that the `trunk()` alias refers to has been deleted and you need to update the `trunk()` alias.

## Actual Behavior

Output from `jj git fetch` in step 6:

```
> jj git fetch
bookmark: main@origin   [updated] tracked
bookmark: master@origin [deleted] untracked
```

Most `jj` commands fail after this `jj git fetch`.

```
> jj new main
Config error: Invalid `revset-aliases.immutable_heads()`
Caused by: Revision "master@origin" doesn't exist
```

```
> jj log
Error: Revision "master@origin" doesn't exist
```

```
> jj log -r main@origin
Error: Failed to parse template: In alias "builtin_log_node"
Caused by:
1:  --> 1:1
  |
1 | builtin_log_node
  | ^--------------^
  |
  = In alias "builtin_log_node"
2:  --> 6:10
  |
6 |       if(immutable, "immutable"),
  |          ^-------^
  |
  = Failed to evaluate revset
3: Revision "master@origin" doesn't exist
Hint: Did you mean "main@origin"
```

```
> jj bookmark untrack master@origin
Error: No such remote bookmark: master@origin
```

It's not clear how to recover from this situation other than `rm -R .jj && jj git init --colocate`

## Specifications

- Platform: NixOS
- Version: 0.22.0

yuja:
You'll ened to update `trunk()` saved in .jj/repo/config.toml, which was probably generated by `jj git init --colocate`.

I agree the error message isn't actionable.

PR #4617, opened 2024-10-10
cli: warn if trunk() alias cannot be resolved, fall back to none()

yuja:
#4616

# Checklist

If applicable:


cli/src/revset_util.rs
```diff
@@ -213,6 +214,37 @@ pub fn parse_immutable_heads_expression(
...
+/// generated by `jj git init`/`clone`.
+pub(super) fn warn_unresolvable_trunk(
+    ui: &Ui,
+    repo: &dyn Repo,
+    context: &RevsetParseContext,
+) -> io::Result<()> {
+    let (_, _, revset_str) = context
+        .aliases_map()
+        .get_function("trunk", 0)
+        .expect("trunk() should be defined by default");
```
PhilipMetzger, line 227: Nit: Can we not crash here? Since some people surely will try to remove or misconfigure the `trunk()` alias.

cli/src/revset_util.rs
```diff
@@ -213,6 +214,37 @@ pub fn parse_immutable_heads_expression(
...
+    // prefixes.
+    let symbol_resolver = DefaultSymbolResolver::new(repo, context.symbol_resolvers());
+    if let Err(err) = expression.resolve_user_expression(repo, &symbol_resolver) {
+        writeln!(
+            ui.warning_default(),
+            "Failed to resolve `revset-aliases.trunk()`: {err}"
+        )?;
+        writeln!(
+            ui.hint_default(),
+            "Use `jj config edit --repo` to update or delete `trunk()` alias."
```
PhilipMetzger, line 242: Nit: s/update or delete/adjust the/ since that implies an action of either updating or deleting. But that is mostly preference.

PhilipMetzger (approved):
LG, just two minor nits.

cli/src/config/revsets.toml
```diff
@@ -18,7 +22,8 @@ latest(
 )
 '''
 
-'builtin_immutable_heads()' = 'trunk() | tags() | untracked_remote_bookmarks()'
+# If immutable_heads() failed to evaluate, many jj commands wouldn't work.
```
martinvonz, line 25: nit: say something about `present()` here so it doesn't sound like generic explaining what `builtin_immutable_heads()` is for. Same for `revsets.log` above

cli/src/revset_util.rs
martinvonz: We don't have a way of removing config (and `trunk()` is defined by default), but can it happen if you redefine `trunk() = 1` (an integer) or something like that?

martinvonz (approved):
Thanks for fixing.

cli/src/config/revsets.toml
yuja: Does this work?
```
# If immutable_heads() failed to evaluate, many jj commands wouldn't work. Use
# present(expr) to suppress symbol resolution error.
```

cli/src/revset_util.rs
yuja: Aliases are loaded from all config layers (to get around parameter name differences.) Invalid `trunk() = 1` will never override the default.

cli/src/config/revsets.toml
martinvonz: Yes, SGTM, thanks

PR #4623, opened 2024-10-11
cli: use singular form when only 1 commit is abandoned after Git import

cenviity:
# Checklist

If applicable:
~- [ ] I have updated `CHANGELOG.md`~
~- [ ] I have updated the documentation (README.md, docs/, demos/)~
~- [ ] I have updated the config schema (cli/src/config-schema.json)~
~- [ ] I have added tests to cover my changes~


yuja:
Thanks, but the current consensus is to not add ad-hoc singular/plural handling, and revisit it when we introduce i18n support.

PR #4626, opened 2024-10-11
Revset: allow returning errors during iteration

torquestomp:
# Checklist

If applicable:

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -381,6 +381,7 @@ fn abandon_unreachable_commits(
         .evaluate_programmatic(mut_repo)
         .unwrap()
         .iter()
+        .map(Result::unwrap) // TODO: Return error to caller
         .collect_vec();
     for abandoned_commit in &abandoned_commits {
         mut_repo.record_abandoned_commit(abandoned_commit.clone());
--- a/lib/src/graph.rs
+++ b/lib/src/graph.rs
@@ -19,11 +19,15 @@ use std::collections::HashSet;
 use std::collections::VecDeque;
 use std::hash::Hash;
 
+use crate::revset::RevsetEvaluationError;
+
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
 pub struct GraphEdge<N> {
     pub target: N,
     pub edge_type: GraphEdgeType,
 }
+
+pub type GraphNodeResult<N, E> = Result<(N, Vec<GraphEdge<N>>), E>;
 
 impl<N> GraphEdge<N> {
     pub fn missing(target: N) -> Self {
@@ -77,10 +81,13 @@ impl<N> ReverseGraphIterator<N>
 where
     N: Hash + Eq + Clone,
 {
-    pub fn new(input: impl IntoIterator<Item = (N, Vec<GraphEdge<N>>)>) -> Self {
+    pub fn new(
+        input: impl Iterator<Item = Result<(N, Vec<GraphEdge<N>>), RevsetEvaluationError>>,
+    ) -> Result<Self, RevsetEvaluationError> {
         let mut entries = vec![];
         let mut reverse_edges: HashMap<N, Vec<GraphEdge<N>>> = HashMap::new();
-        for (node, edges) in input {
+        for item in input {
+            let (node, edges) = item?;
             for GraphEdge { target, edge_type } in edges {
                 reverse_edges.entry(target).or_default().push(GraphEdge {
                     target: node.clone(),
@@ -95,15 +102,15 @@ where
             let edges = reverse_edges.get(&node).cloned().unwrap_or_default();
             items.push((node, edges));
         }
-        Self { items }
+        Ok(Self { items })
     }
 }
 
 impl<N> Iterator for ReverseGraphIterator<N> {
-    type Item = (N, Vec<GraphEdge<N>>);
+    type Item = Result<(N, Vec<GraphEdge<N>>), RevsetEvaluationError>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.items.pop()
+        self.items.pop().map(Ok)
     }
 }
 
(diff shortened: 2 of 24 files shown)


lib/src/graph.rs
```diff
@@ -148,10 +154,12 @@ impl<N> Default for TopoGroupedGraphNode<N> {
     }
 }
 
+type GraphEdgeResult<N> = Result<Option<(N, Vec<GraphEdge<N>>)>, RevsetEvaluationError>;
+
 impl<N, I> TopoGroupedGraphIterator<N, I>
 where
     N: Clone + Hash + Eq,
-    I: Iterator<Item = (N, Vec<GraphEdge<N>>)>,
+    I: RevsetIterator<(N, Vec<GraphEdge<N>>)>,
```
yuja, line 162: `TopoGroupedGraphIterator` is used in other contexts, so the error type would have to be generic. For example, `show_op_diff()` can use `E = Infallible`.

lib/src/revset.rs
```diff
@@ -2065,18 +2065,47 @@ impl VisibilityResolutionContext<'_> {
...
+        T: 'a;
+}
+
+impl<T, I: Iterator<Item = Result<T, RevsetEvaluationError>>> RevsetIterator<T> for I {
+    fn unwrapped<'a>(self: Box<Self>) -> Box<dyn Iterator<Item = T> + 'a>
+    where
+        Self: 'a,
+        T: 'a,
+    {
+        Box::new(UnwrappedRevsetIterator { iter: self })
```
yuja, line 2081: It's bad to suggest using `.unwrap()`. Maybe we can use `.try_collect().unwrap()`, `.map_ok()`, etc. instead?
https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.map_ok

lib/src/revset.rs
```diff
@@ -2123,13 +2151,14 @@ pub struct RevsetCommitIterator<I> {
     iter: I,
 }
 
-impl<I: Iterator<Item = CommitId>> Iterator for RevsetCommitIterator<I> {
+impl<I: RevsetIterator<CommitId>> Iterator for RevsetCommitIterator<I> {
     type Item = BackendResult<Commit>;
```
yuja, line 2155: The error type would have to be changed to `RevsetEvaluationError`. If that involves huge code change, please leave a TODO comment.

yuja (commented):
Thanks!

lib/src/revset.rs
torquestomp: I inlined `map(Result::unwrap)` at the call sites instead.

lib/src/revset.rs
torquestomp: Added a commit for it; it was a relatively small change.

lib/src/revset.rs
```diff
@@ -98,6 +97,15 @@ pub enum RevsetEvaluationError {
     Other(Box<dyn std::error::Error + Send + Sync>),
 }
 
+impl RevsetEvaluationError {
+    pub fn into_backend_error(self) -> BackendError {
+        match self {
+            Self::StoreError(err) => err,
+            Self::Other(err) => BackendError::Other(err),
```
yuja, line 104: It's probably better to do `panic!()` as before because that's the caller's expectation. The function can be renamed to `.expect_backend_error()`?

lib/src/git.rs
```diff
@@ -380,6 +380,7 @@ fn abandon_unreachable_commits(
         .evaluate_programmatic(mut_repo)
         .unwrap()
         .iter()
+        .map(Result::unwrap)
```
yuja, line 383: nit: the error should be propagated to caller. Can you leave a TODO comment?

lib/src/repo.rs
```diff
@@ -1169,7 +1169,8 @@ impl MutableRepo {
         let heads_to_add = heads_to_add_expression
             .evaluate_programmatic(self)
             .unwrap()
-            .iter();
+            .iter()
+            .map(Result::unwrap);
```
yuja, line 1173: nit: same here (and two more changes in repo.rs)
The error should have to be returned to caller. TODO comment?

lib/src/id_prefix.rs
```diff
@@ -40,9 +40,17 @@ use crate::revset::SymbolResolverExtension;
...
+    Resolution(
+        #[from]
+        #[source]
+        RevsetResolutionError,
+    ),
     #[error("Failed to evaluate short-prefixes disambiguation revset")]
-    Evaluation(#[source] RevsetEvaluationError),
+    Evaluation(
+        #[from]
+        #[source]
```
yuja, line 51: nit: `#[source]` is implied by `#[from]`, can be deleted

lib/src/id_prefix.rs
```diff
@@ -73,7 +81,7 @@ impl DisambiguationData {
                 .evaluate(repo)
                 .map_err(IdPrefixIndexLoadError::Evaluation)?;
```
yuja, line 82: nit: can you also remove these two `.map_err()`s?

lib/src/revset.rs
```diff
@@ -2158,18 +2158,21 @@ impl VisibilityResolutionContext<'_> {
     }
 }
 
+pub trait RevsetIterator<T>: Iterator<Item = Result<T, RevsetEvaluationError>> {}
+impl<T, I: Iterator<Item = Result<T, RevsetEvaluationError>>> RevsetIterator<T> for I {}
```
yuja, line 2162: nit: it no longer make much sense to define new `RevsetIterator` trait. Just inline `Item = Result<..>`?

yuja (approved):
Thanks!

lib/src/revset.rs
torquestomp: This somewhat defeats the point of the PR; the goal of being able to return errors from the Revset iterator is as a safer and cleaner alternative to panics.  If the client panics anyway this achieves a lot less.

I'll leave it like this for now since I'm starting to accrue conflicts but I'd like to address this in a future PR.

yuja: If revset engine can return `Other(err)` for these ancestry queries, callers would have to propagate it as `RevsetEvaluationError`. That's the point ,and leaving TODO looks good to me. Thanks.

PR #4627, opened 2024-10-11
backout: partial reversion of 3f0703ca2cf

thoughtpolice:
All of these toml files already have a `[lints]` key, it just exited earlier on in the file and were missed.

However, the occurrence of duplicate keys breaks TOML parsers that have stricter standard rules and reject these files (rather than merging the values). In particular, `tomllib` in Python 3.11 will reject these cargo files.

# Checklist

If applicable:

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -117,6 +117,3 @@ watchman = ["jj-lib/watchman"]
 # The archive name is jj, not jj-cli. Also, `cargo binstall` gets
 # confused by the `v` before versions in archive name.
 pkg-url = "{ repo }/releases/download/v{ version }/jj-v{ version }-{ target }.{ archive-format }"
-
-[lints]
-workspace = true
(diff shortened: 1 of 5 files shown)


PhilipMetzger:
This is missing a note to #1997 as an explainer in the description.

martinvonz (commented):
> All of these toml files already have a `[lints]` key, it just exited earlier on
in the file and were missed.

What do you mean by that? I don't see any existing `[lints]` tables elsewhere in these files. And what exited earlier?

thoughtpolice:
Oh, you're right! OK, this was introduced on my end by a merge error, I think, and when I rebased off of my branch onto trunk, it applied the diff naturally (because my other branch introduced the lint keys separately.) I think this can be closed, but let me confirm that....

thoughtpolice:
Yes, PEBKAC. Sorry about that!

PR #4628, opened 2024-10-12
revset: add `coalesce(revsets...)`

bnjmnt4n:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   The output can be customized via the `templates.annotate_commit_summary`
   config variable.
 
+* New `coalesce(revsets...)` revset which returns commits in the first revset
+  in the `revsets` list that does not evaluate to `none()`.
+
 ### Fixed bugs
 
 * Error on `trunk()` revset resolution is now handled gracefully.
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -304,6 +304,10 @@ given [string pattern](#string-patterns).
 * `present(x)`: Same as `x`, but evaluated to `none()` if any of the commits
   in `x` doesn't exist (e.g. is an unknown bookmark name.)
 
+* `coalesce(revsets...)`: Commits in the first revset in the list of `revsets`
+  which does not evaluate to `none()`. If all revsets evaluate to `none()`, then
+  the result of `coalesce` will also be `none()`.
+
 * `working_copies()`: The working copy commits across all the workspaces.
 
 * `at_operation(op, x)`: Evaluates `x` at the specified [operation][]. For
--- a/lib/src/default_index/revset_engine.rs
+++ b/lib/src/default_index/revset_engine.rs
@@ -919,6 +919,14 @@ impl<'index> EvaluationContext<'index> {
                     self.take_latest_revset(candidate_set.as_ref(), *count),
                 ))
             }
+            ResolvedExpression::Coalesce(expression1, expression2) => {
+                let set1 = self.evaluate(expression1)?;
+                if set1.positions().attach(index).next().is_some() {
+                    Ok(set1)
+                } else {
+                    self.evaluate(expression2)
+                }
+            }
             ResolvedExpression::Union(expression1, expression2) => {
                 let set1 = self.evaluate(expression1)?;
                 let set2 = self.evaluate(expression2)?;
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -218,6 +218,7 @@ pub enum RevsetExpression {
         /// Copy of `repo.view().heads()`, should be set by `resolve_symbols()`.
         visible_heads: Option<Vec<CommitId>>,
     },
+    Coalesce(Rc<Self>, Rc<Self>),
     Present(Rc<Self>),
     NotIn(Rc<Self>),
     Union(Rc<Self>, Rc<Self>),
@@ -439,6 +440,20 @@ impl RevsetExpression {
         Rc::new(Self::Difference(self.clone(), other.clone()))
     }
 
+    /// Commits that are in the first expression in `expressions` that is not
+    /// `none()`.
+    pub fn coalesce(expressions: &[Rc<Self>]) -> Rc<Self> {
+        match expressions {
+            [] => Self::none(),
+            [expression] => expression.clone(),
+            _ => {
+                // Build balanced tree to minimize the recursion depth.
+                let (left, right) = expressions.split_at(expressions.len() / 2);
+                Rc::new(Self::Coalesce(Self::coalesce(left), Self::coalesce(right)))
+            }
+        }
+    }
+
     /// Resolve a programmatically created revset expression.
     ///
     /// In particular, the expression must not contain any symbols (bookmarks,
@@ -528,6 +543,7 @@ pub enum ResolvedExpression {
         candidates: Box<Self>,
         count: usize,
     },
+    Coalesce(Box<Self>, Box<Self>),
     Union(Box<Self>, Box<Self>),
     /// Intersects `candidates` with `predicate` by filtering.
     FilterWithin {
@@ -853,6 +869,14 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
             visible_heads: None,
         }))
     });
+    map.insert("coalesce", |diagnostics, function, context| {
+        let ([], args) = function.expect_some_arguments()?;
+        let expressions: Vec<_> = args
+            .iter()
+            .map(|arg| lower_expression(diagnostics, arg, context))
+            .try_collect()?;
+        Ok(RevsetExpression::coalesce(&expressions))
+    });
     map
 });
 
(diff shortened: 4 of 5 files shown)


lib/src/default_index/revset_engine.rs
```diff
@@ -919,6 +919,23 @@ impl<'index> EvaluationContext<'index> {
...
+                condition_set.positions().attach(index).next().map_or_else(
+                    || -> Result<_, _> {
+                        if let Some(alternate) = alternate {
+                            self.evaluate(alternate)
+                        } else {
+                            Ok(Box::new(EagerRevset::empty()))
+                        }
+                    },
+                    |_pos| -> Result<_, _> { self.evaluate(consequent) },
+                )
```
yuja, line 937: nit: can be expressed as:
```rust
if condition_set.positions().attach(index).next().is_some() {
    ...
} else {
    ...
}
```

lib/src/revset.rs
```diff
@@ -218,6 +218,11 @@ pub enum RevsetExpression {
         /// Copy of `repo.view().heads()`, should be set by `resolve_symbols()`.
         visible_heads: Option<Vec<CommitId>>,
     },
+    If {
+        condition: Rc<Self>,
+        consequent: Rc<Self>,
+        alternate: Option<Rc<Self>>,
```
yuja, line 224: nit: can be set to `alternate: Rc<Self> = Self::none()` instead of `None`.

CHANGELOG.md
yuja, line 1: As I said in Discord, I think `coalesce(x, y)` is simpler (and is probably less confusing.)

`if(x, y, z)` could be misinterpreted as `source_revisions.filter(|rev| if x(rev) { y(rev) } else { z(rev) })`.

Can you also include some use case in the commit message?

emilazy: Is there a nice way to express `if` in terms of `coalesce`? Right now we can already express `coalesce` and `if`, but not in any O(1) way, which is awkward. Perhaps `if_empty` would be a better name?

yuja: I don't think `if` can be expressed in terms of `coalesce`. If there's a real use case of `if`, we'll have to add it. `if(x, x, y)` could be transformed to `coalesce(x, y)` to not evaluate `x` twice, but I don't think that's important.

> Perhaps `if_empty` would be a better name?

We have `empty()` function for a different thing, so the function will have to be named `if_none()` or `if_not_none()`.

bnjmnt4n: > Can you also include some use case in the commit message?

Personally I don't have a use case for this. I added this because it seems like it's more expressive than `coalesce`, and there did seem to be some discussion about it.

Currently `if` behaves like `if_not_none` instead of `if_none`, so I'd rename it to `if_not_none` but that might be a bit confusing as well. Perhaps inverting the condition to `if_none` instead might be better?

bnjmnt4n:
I've updated to add a `coalesce` function as well. I'm not very familiar with the revset engine for now, so I'm not sure if I should attempt to create an `InternalRevset` for these two functions versus evaluating it in the `evaluate` function.

lib/src/revset.rs
```diff
@@ -223,6 +223,7 @@ pub enum RevsetExpression {
         consequent: Rc<Self>,
         alternate: Rc<Self>,
     },
+    Coalesce(Vec<Rc<Self>>),
```
yuja, line 226: I would transform `coalesce(x, y, z..)` to binary form `Coalesce(x, Coalesce(y, z))` because it's easier to process in later pass (and the chain wouldn't be that long.)

CHANGELOG.md
yuja: Yeah, `if_not_none` might be confusing, too. That's another reason I think `coalesce()` is nicer than generic `if()`s. In any case, if we add `If`, I don't think we'll need less-generic `RevsetExpression::Coalesce(..)`. We might want `coalesce(..)` function for convenience, but it can be transformed to `If`.

yuja (commented):
> I'm not very familiar with the revset engine for now, so I'm not sure if I should attempt to create an `InternalRevset` for these two functions versus evaluating it in the `evaluate` function.

I think the current implementation is good. We could add `InternalRevset` to make iteration lazy, but it doesn't mean the evaluation can always be lazy. For example, `coalesce(foo, heads(bar))` would have to "evaluate" `bar` because `heads(candidates)` evaluates `candidates` eagerly. With the current implementation, `heads(bar)` evaluation can be skipped if `foo` is not empty.

CHANGELOG.md
yuja: To be clear, my opinion is:
1. just add `coalesce()` function (and binary `Coalesce(x, y)` as an internal representation.)
2. revisit `If` things if we find a real use case.

docs/revsets.md
```diff
@@ -310,6 +310,10 @@ given [string pattern](#string-patterns).
   example, `at_operation(@-, visible_heads())` will return all heads which were
   visible at the previous operation.
 
+* `coalesce(revsets...)`: Commits in the first revset in the list of `revsets`
+  which does not evaluate to `none()`. If all revsets evaluate to `none()`, then
+  the result of `coalesce` will also be `none()`.
```
yuja, line 315: nit: I think `coalesce()` can be placed somewhere close to `present(x)`. `at_opeartion()` is quite different from the other functions.

lib/src/revset.rs
```diff
@@ -853,6 +855,17 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
...
         }))
     });
+    map.insert("coalesce", |diagnostics, function, context| {
+        let ([], args) = function.expect_some_arguments()?;
+        let mut revset = RevsetExpression::none();
+        for arg in args.iter().rev() {
+            revset = Rc::new(RevsetExpression::Coalesce(
+                lower_expression(diagnostics, arg, context)?,
+                revset,
+            ));
```
yuja, line 865: nit: can you add a few `RevsetExpression` construction tests (similar to e.g. `union_all()`)? It might be easier if this function is extracted as `RevsetExpression` method.

btw, this function will leave a redundant `none()` at the bottom expression. I think that's okay, but if we want to simplify the constructed expression, we can either special case `[]` or do recursion in the same way as `union_all()`.

docs/revsets.md
yuja, line 1: nit: for future reference, can you include some use case (or reason why we want this) in the commit message? Maybe it can be copied from the Discord discussion.

yuja (approved):
Thanks!

docs/revsets.md
yuja: Mine is something like `coalesce(present(user_configured_trunk), builtin_trunk)`.

@emilazy do you have a better example?

lib/src/revset.rs
```diff
@@ -439,6 +440,19 @@ impl RevsetExpression {
...
+    /// Commits that are in the first expression in `expressions` that is not
+    /// `none()`.
+    pub fn coalesce(expressions: &[Rc<Self>]) -> Rc<Self> {
+        match expressions {
+            [] => Self::none(),
+            [expression] => expression.clone(),
+            [expression, remainder @ ..] => Rc::new(Self::Coalesce(
+                expression.clone(),
+                Self::coalesce(remainder),
+            )),
```
yuja, line 452: nit: If we use recursion, it might be better to split at middle so that long sequence of expressions can be processed.

To build a right-recursive tree, maybe `rfold()` can be used instead.
https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html#method.rfold

docs/revsets.md
emilazy: @lilyball’s motivating use‐case [from the Discord](https://discord.com/channels/968932220549103686/968932220549103689/1293722881968312433):

> My original use-case here was for a `jj u` alias that rebases the current stack onto the appropriate base, where it tries to figure out what the most appropriate base is for the stack. My original attempt considered trunk and untracked remote branches, and then I discovered that this failed if a coworker pushed a PR to the repo that branched off of trunk from the same commit that I branched off, because that remote branch is now just as suitable a base as trunk. So I've now changed it to consider trunk first, and only check remote branches if trunk is unsuitable
>
> here's my current config for that
>
> ```nix
>       revset-aliases = {
>         "trunks()" = "trunk()"; # override point for repos with multiple trunks
>         "immutable_heads()" = "trunks() | builtin_immutable_heads()";
>         "immutable_movable_heads()" = "trunks() | untracked_remote_bookmarks()";
> 
>         "stack_base()" = "if_nonempty(stack_base_trunks(), stack_base_all())";
>         "stack_base_trunks()" = "heads(reachable(@, mutable())-::trunks())";
>         "stack_base_all()" = "heads(reachable(@, mutable())-::immutable_movable_heads())";
>         "stack_roots()" = "roots(reachable(@, mutable()) & children(::stack_base()) & ~merges())";
> 
>         # horrible hack
>         "if_nonempty(a, b)" = "a | (b ~ (a|(b..a)-)::b)";
>       };
> ```

I’m not sure if anyone has come up with a concrete use for full `if`.

lib/src/revset.rs
bnjmnt4n: Opted to just use recursion for now instead.

Issue #3224, opened 2024-03-06
FR: explain command


khionu:
**Is your feature request related to a problem? Please describe.**
As jj promotes a different mindset than will be familiar to most people who try it, it might be useful to have a `jj explain <ident>` command. Explanations would be a catch-all place to add in-depth documentation to help people understand various concepts and issues. This would be a place that deprecation reasonings could be outlined.

**Describe the solution you'd like**
- A directory added to the repo that contains text files, each its own explanation entry
- Each entry would be named in the pattern of `E000_short_description.txt`
- Each entry would start with a header that would contain:
  - The ID of the entry
  - A renderable short description
  - A list of links relevant to the evolution of the document (for tracking the history of the context)
- After the header, each entry would contain a plain text body with annotations for bold, italics, underline, and [links](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
  - It's possible that variable interpolation may be useful, eg passing `--rev-a XX --rev-b YY` could allow examples to be filled in with commands that are ready for direct usage.

**Describe alternatives you've considered**
The most minimal implementation of this would infer the ID and short description from the file name, then use nothing but plain text, without annotations, for the body.

joyously:
Is this like #3130?

khionu:
Hrmm... maybe similar enough to where it could be implemented the same way? Maybe similar enough to be redundant? Though there is a distinction between docs and explanations.

Explanations would be more for providing nuance for specific contexts. An error might return a message long enough for a more experienced user, and say "run `jj explain E1234` for more details" to help those that haven't run into the error before.

mhammond:
> Is this like https://github.com/martinvonz/jj/pull/3130?

I think it's complimentary - the `<ident>` here is interesting. I just noticed in the cli:

> If the source became empty and both the source and destination had a non-empty description, you will be asked for the combined description. If either was empty, then the other
one will be used.
>
> If a working-copy commit gets abandoned, it will be given a new, empty commit. This is true in general; it is not specific to this command.

I can imagine, eg,

> You might be asked for a description - see 'jj docs edits-descriptions'
>
> The usual working-copy semantics apply - see 'jj docs edits-and-working-copy'

Something like this would make "help" be both more concise and informative.

joyously:
I don't see why we need more commands (`docs` or `explain`). Both Git and Breezy use `help` for commands and other topics which have pages of text (Git's are longer).

khionu:
[...]

PR #3130, opened 2024-02-24
wip: implement `jj docs` command

thoughtpolice:
This is a rudimentary approach to embedding docs in the binary. It uses `termimad` to render markdown into the terminal. It still needs a lot of work, but something like `jj docs FAQ` can in fact work right now.

Getting this in a submittable state probably isn't too much work, but polishing the docs so they look good will take a bit of effort.

Closes #1294.

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #4630, opened 2024-10-13
help: Add a keyword feature

Grillo-0:
It would be nice to not need to go the documentation website. This aims to solve that by introducing the concept of categories to the help command.

Basically, categories are things that we want to add help messages to, but they don't necessarily have an associated subcommand.

For now we only have two categories:
	- `revsets`: Shows the docs for revsets
	- `tutorial`: Shows the tutorial that is on the documentation

You get the category content by tipping `jj help revsets`.

It would be nice to have all the documentation on the categories, maybe a next commit could do it.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New command `jj config unset` that unsets config values. For example,
   `jj config unset --user user.name`.
 
+* `jj help` now has the flag `--keyword` (shorthand `-k`), which can give help
+  for some keywords (e.g. `jj help -k revsets`). To see a list of the available
+  keywords you can do `jj help --help`.
+
 ### Fixed bugs
 
 * Error on `trunk()` revset resolution is now handled gracefully.
--- a/cli/src/commands/help.rs
+++ b/cli/src/commands/help.rs
@@ -12,6 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::fmt::Write as _;
+use std::io::Write;
+
+use clap::builder::PossibleValue;
+use clap::builder::StyledStr;
+use crossterm::style::Stylize;
+use itertools::Itertools;
 use tracing::instrument;
 
 use crate::cli_util::CommandHelper;
@@ -24,14 +31,33 @@ use crate::ui::Ui;
 pub(crate) struct HelpArgs {
     /// Print help for the subcommand(s)
     pub(crate) command: Vec<String>,
+    /// Show help for keywords instead of commands
+    #[arg(
+        long,
+        short = 'k',
+        conflicts_with = "command",
+        value_parser = KEYWORDS
+            .iter()
+            .map(|k| PossibleValue::new(k.name).help(k.description))
+            .collect_vec()
+    )]
+    pub(crate) keyword: Option<String>,
 }
 
 #[instrument(skip_all)]
 pub(crate) fn cmd_help(
-    _ui: &mut Ui,
+    ui: &mut Ui,
     command: &CommandHelper,
     args: &HelpArgs,
 ) -> Result<(), CommandError> {
+    if let Some(name) = &args.keyword {
+        let keyword = find_keyword(name).expect("clap should check this with `value_parser`");
+        ui.request_pager();
+        write!(ui.stdout(), "{}", keyword.content)?;
+
+        return Ok(());
+    }
+
     let mut args_to_show_help = vec![command.app().get_name()];
     args_to_show_help.extend(args.command.iter().map(|s| s.as_str()));
     args_to_show_help.push("--help");
@@ -46,4 +72,53 @@ pub(crate) fn cmd_help(
         .expect_err("Clap library should return a DisplayHelp error in this context");
 
     Err(command_error::cli_error(help_err))
+}
+
+#[derive(Clone)]
+struct Keyword {
+    name: &'static str,
+    description: &'static str,
+    content: &'static str,
+}
+
+// TODO: Add all documentation to keywords
+//
+// Maybe adding some code to build.rs to find all the docs files and build the
+// `KEYWORDS` at compile time.
+//
+// It would be cool to follow the docs hierarchy somehow.
+//
+// One of the problems would be `config.md`, as it has the same name as a
+// subcommand.
+//
+// TODO: Find a way to render markdown using ANSI escape codes.
+//
+// Maybe we can steal some ideas from https://github.com/martinvonz/jj/pull/3130
+const KEYWORDS: &[Keyword] = &[
+    Keyword {
+        name: "revsets",
+        description: "A functional language for selecting a set of revision",
+        content: include_str!("../../../docs/revsets.md"),
+    },
+    Keyword {
+        name: "tutorial",
+        description: "Show a tutorial to get started with jj",
+        content: include_str!("../../../docs/tutorial.md"),
+    },
+];
+
+fn find_keyword(name: &str) -> Option<&Keyword> {
+    KEYWORDS.iter().find(|keyword| keyword.name == name)
+}
+
+pub fn show_keyword_hint_after_help() -> StyledStr {
+    let mut ret = StyledStr::new();
+    writeln!(
+        ret,
+        "{} list available keywords. Use {} to show help for one of these keywords.",
+        "'jj help --help'".bold(),
+        "'jj help -k'".bold(),
+    )
+    .unwrap();
+    ret
 }
--- a/cli/tests/test_help_command.rs
+++ b/cli/tests/test_help_command.rs
@@ -84,3 +84,62 @@ fn test_help() {
     For more information, try '--help'.
     "#);
 }
+
+#[test]
+fn test_help_keyword() {
+    let test_env = TestEnvironment::default();
+
+    // It should show help for a certain keyword if the `--keyword` flag is present
+    let help_cmd_stdout =
+        test_env.jj_cmd_success(test_env.env_root(), &["help", "--keyword", "revsets"]);
+    // It should be equal to the docs
+    assert_eq!(help_cmd_stdout, include_str!("../../docs/revsets.md"));
+
+    // It should show help for a certain keyword if the `-k` flag is present
+    let help_cmd_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help", "-k", "revsets"]);
+    // It should be equal to the docs
+    assert_eq!(help_cmd_stdout, include_str!("../../docs/revsets.md"));
+
+    // It should give hints if a similar keyword is present
+    let help_cmd_stderr = test_env.jj_cmd_cli_error(test_env.env_root(), &["help", "-k", "rev"]);
+    insta::assert_snapshot!(help_cmd_stderr, @r#"
+    error: invalid value 'rev' for '--keyword <KEYWORD>'
+      [possible values: revsets, tutorial]
+
+      tip: a similar value exists: 'revsets'
+
+    For more information, try '--help'.
+    "#);
+
+    // It should give error with a hint if no similar keyword is found
+    let help_cmd_stderr =
+        test_env.jj_cmd_cli_error(test_env.env_root(), &["help", "-k", "<no-similar-keyword>"]);
+    insta::assert_snapshot!(help_cmd_stderr, @r#"
+    error: invalid value '<no-similar-keyword>' for '--keyword <KEYWORD>'
+      [possible values: revsets, tutorial]
+
+    For more information, try '--help'.
+    "#);
+
+    // The keyword flag with no argument should error with a hint
+    let help_cmd_stderr = test_env.jj_cmd_cli_error(test_env.env_root(), &["help", "-k"]);
+    insta::assert_snapshot!(help_cmd_stderr, @r#"
+    error: a value is required for '--keyword <KEYWORD>' but none was supplied
+      [possible values: revsets, tutorial]
+
+    For more information, try '--help'.
+    "#);
+
+    // It shouldn't show help for a certain keyword if the `--keyword` is not
+    // present
+    let help_cmd_stderr = test_env.jj_cmd_cli_error(test_env.env_root(), &["help", "revsets"]);
+    insta::assert_snapshot!(help_cmd_stderr, @r#"
+    error: unrecognized subcommand 'revsets'
+
+      tip: some similar subcommands exist: 'resolve', 'prev', 'restore', 'rebase', 'revert'
+
+    Usage: jj [OPTIONS] <COMMAND>
+
+    For more information, try '--help'.
+    "#);
+}
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -73,6 +73,7 @@ use crate::ui::Ui;
 
 #[derive(clap::Parser, Clone, Debug)]
 #[command(disable_help_subcommand = true)]
+#[command(after_long_help = help::show_keyword_hint_after_help())]
 enum Command {
     Abandon(abandon::AbandonArgs),
     Backout(backout::BackoutArgs),
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -113,6 +113,9 @@ To get started, see the tutorial at https://martinvonz.github.io/jj/latest/tutor
 
 **Usage:** `jj [OPTIONS] [COMMAND]`
 
+'jj help --help' list available keywords. Use 'jj help -k' to show help for one of these keywords.
+
+
 ###### **Subcommands:**
 
 * `abandon` — Abandon a revision
@@ -1215,11 +1218,22 @@ Set the URL of a Git remote
 
 Print this message or the help of the given subcommand(s)
 
-**Usage:** `jj help [COMMAND]...`
+**Usage:** `jj help [OPTIONS] [COMMAND]...`
 
 ###### **Arguments:**
 
 * `<COMMAND>` — Print help for the subcommand(s)
+
+###### **Options:**
+
+* `-k`, `--keyword <KEYWORD>` — Show help for keywords instead of commands
+
+  Possible values:
+  - `revsets`:
+    A functional language for selecting a set of revision
+  - `tutorial`:
+    Show a tutorial to get started with jj
+
 
 
 


joyously:
Is this associated with #3224?

cli/src/commands/help.rs
```diff
@@ -26,21 +31,99 @@ pub(crate) struct HelpArgs {
...
+        .get_subcommands()
+        .map(|cmd| cmd.get_name().len())
+        .max()
+        .unwrap();
+
+    let help_template_categories = format!(
+        "\x1b[1m\x1b[4mCategories:\x1b[0m\n{}",
+        &categories
+            .iter()
+            .map(|(&name, category)| format!(
```
yuja, line 102: Change `categories` to `Vec` or sort the `HashMap` here.

cli/src/commands/help.rs
```diff
@@ -26,21 +31,99 @@ pub(crate) struct HelpArgs {
...
+
+    let subcommand_max_len = command
+        .app()
+        .get_subcommands()
+        .map(|cmd| cmd.get_name().len())
+        .max()
+        .unwrap();
+
+    let help_template_categories = format!(
+        "\x1b[1m\x1b[4mCategories:\x1b[0m\n{}",
```
yuja, line 99: I think it's confusing to include `jj help <CATEGORY>` list in the `jj <COMMAND>` help. Maybe it can be listed in `jj help help`? Alternatively, the header can be renamed to "Help Categories" or something.

Grillo-0:
> Is this associated with #3224?

Oh! I didn't know about that! But yeah, I don't see why this couldn't be extended after to explain errors. This PR started from discussion on #4518: https://github.com/martinvonz/jj/pull/4518#discussion_r1784521735.

cli/src/commands/help.rs
Grillo-0: Yeah, I agree that now it's confusing, people may want to do `jj revsets`. But I don't think that putting in `jj help help`, for me, seems a little hidden. For now, I'm going to rename to "Help Categories".

Maybe, as I'm already changing the help template, I could but a `usage: jj help <CATEGORY>` and some text explaining what a category are?

cli/src/commands/help.rs
Grillo-0: Could you explain why?

Grillo-0: Oh, to sort the category names correctly?

yuja: and stabilize the output. We should add some test to capture the output.

Grillo-0: Added a test for that. I changed to an BTreeMap, this should do it.

cli/src/commands/help.rs
Grillo-0: For the moment, I just changed to "Help Categories". Now, I'm not so sure that going with the `jj help help` is all that bad.

I think this needs more discussion.

cli/src/commands/help.rs
```diff
@@ -26,21 +30,94 @@ pub(crate) struct HelpArgs {
...
 pub(crate) fn cmd_help(
-    _ui: &mut Ui,
+    ui: &mut Ui,
     command: &CommandHelper,
     args: &HelpArgs,
 ) -> Result<(), CommandError> {
+    // TODO: Add all documentation to categories
+    //
+    // Maybe adding some code to build.rs to find all the docs files and build the
+    // HashMap at compile time.
```
yuja, line 47: nit: `HashMap` -> `BTreeMap`

btw, it might make sense to extract the inner array as file-local `const CATEGORIES: &[Category] = ..`, and add helper function that constructs `help_template_categories`.

cli/src/commands/help.rs
```diff
@@ -26,21 +30,94 @@ pub(crate) struct HelpArgs {
     pub(crate) command: Vec<String>,
 }
 
+struct Category {
+    description: &'static str,
+    content: &'static str,
+}
```
yuja, line 36: nit: I would move this type after `cmd_help()` so `HelpArgs` is placed closer to the consumer function.

cli/src/commands/help.rs
```diff
@@ -26,21 +30,94 @@ pub(crate) struct HelpArgs {
...
+        (
+            "tutorial",
+            Category {
+                description: "Show a tutorial to get started with jj",
+                content: include_str!("../../../docs/tutorial.md"),
+            },
+        ),
+    ]);
+
+    if let Some(category) = categories.get(args.command.join(" ").as_str()) {
```
yuja, line 70: nit: it's probably better to check if `args.command.len() == 1` instead of concatenating `args`.
```rust
if let [name] = &*args.command {
    ...
}
```

cli/src/commands/help.rs
```diff
@@ -26,21 +30,94 @@ pub(crate) struct HelpArgs {
...
+        &categories
+            .iter()
+            .map(|(&name, category)| format!(
+                "{{tab}}\x1b[1m{:<width$}\x1b[0m{{tab}}{}",
+                name,
+                category.description,
+                width = subcommand_max_len
+            ))
+            .join("\n")
+    );
```
yuja, line 103: nit: do category stuff only if `args.command.is_empty()`? With the current implementation, it's not obvious that the categories list is inserted only to the top-level help.

cli/src/commands/help.rs
```diff
@@ -26,21 +30,94 @@ pub(crate) struct HelpArgs {
...
+
+    let help_template_after_categories = "\
+    \x1b[1m\x1b[4mOptions:\x1b[0m\n{options}
+    ";
+
     // TODO: `help log -- -r` will gives an cryptic error, ideally, it should state
     // that the subcommand `log -r` doesn't exist.
+    #[allow(clippy::uninlined_format_args)]
```
yuja, line 111: nit: any reason not to inline args? Maybe you can also use `.join("\n\n")` instead of `format!()`.

cli/tests/test_help_command.rs
```diff
@@ -84,3 +79,145 @@ fn test_help() {
...
+      sparse            Manage which paths from the working-copy commit are present in the working copy
+      split             Split a revision in two
+      squash            Move changes from a revision into another revision
+      status            Show high-level repo status [aliases: st]
+      tag               Manage tags
+      util              Infrequently used commands such as for generating shell completions
+      undo              Undo an operation (shortcut for `jj op undo`)
+      version           Display version information
+      workspace         Commands for working with workspaces
+
```
yuja, line 134: nit: is it intentional that we have two blank lines here?

cli/tests/test_help_command.rs
```diff
@@ -84,3 +79,145 @@ fn test_help() {
...
+      tag               Manage tags
+      util              Infrequently used commands such as for generating shell completions
+      undo              Undo an operation (shortcut for `jj op undo`)
+      version           Display version information
+      workspace         Commands for working with workspaces
+        
+
+    Help Categories:
+      revsets           Documentation about revsets
+      tutorial          Show a tutorial to get started with jj
```
yuja, line 138: I would move "Help Categories" to bottom because it's not related to the command.

I found the default template in clap source:
https://github.com/clap-rs/clap/blob/master/clap_builder/src/output/help_template.rs#L64

If we don't have to insert "Help Categories" here, maybe we can just use it?

cli/src/commands/help.rs
yuja: I agree about the discoverability. Let's try with "Help Categories" then.

cli/tests/test_help_command.rs
yuja: or `.after_long_help()`?
https://docs.rs/clap/latest/clap/struct.Command.html#method.after_long_help

cli/src/commands/help.rs
```diff
@@ -26,21 +30,94 @@ pub(crate) struct HelpArgs {
...
+    //
+    // It would be cool to follow the docs hierarchy somehow.
+    //
+    // One of the problems would be `config.md`, as it has the same name as a
+    // subcommand.
+    let categories = BTreeMap::from([
+        (
+            "revsets",
+            Category {
+                description: "Documentation about revsets",
```
joyously, line 57: How about an actual description?
"a functional language for selecting a set of revisions"

cli/tests/test_help_command.rs
```diff
@@ -84,3 +79,145 @@ fn test_help() {
...
+              Additional configuration options (can be repeated)
+
+      -h, --help
+              Print help (see a summary with '-h')
+
+      -V, --version
+              Print version
+    "#);
+
+    // Help command should work category commands
```
joyously, line 219: The comment "Help command should work category commands"
doesn't make sense.

cli/tests/test_help_command.rs
```diff
@@ -84,3 +79,145 @@ fn test_help() {
...
     "#);
 }
+
+#[test]
+fn test_help_category() {
+    let test_env = TestEnvironment::default();
+    // Now that we have a custom help output, capture it
+    let help_cmd_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help"]);
+    insta::assert_snapshot!(help_cmd_stdout, @r#"
+    Jujutsu (An experimental VCS)
```
joyously, line 89: I think it's a bad idea to hard code the help output into the test.

yuja: It's not so bad because the snapshot can be updated relatively easily, but I would add some regex pattern to omit commands and options lists.
https://insta.rs/docs/filters/

thoughtpolice:
Related: https://github.com/martinvonz/jj/pull/3130 by me, which actually adds a `docs` subcommand that just includes *all* documentation from the website into the binary; it uses a terminal display library to make the markdown more readable. But I haven't had any time to really polish it and make it merge worthy.

I don't think it should block this or anything, just putting it out there. In the future I imagine some merger of these two ideas would occur, really.

cli/tests/test_help_command.rs
Grillo-0: Oh, actually no. It seems that `{subcommands}` comes with a `\n` at the end

cli/tests/test_help_command.rs
Grillo-0: Oh, much better

Grillo-0: One small thing. With `after_long_help` we can't use `{tab}`.

cli/tests/test_help_command.rs
Grillo-0: Ops, missed a "with"

cli/src/commands/help.rs
```diff
@@ -28,22 +32,98 @@ pub(crate) struct HelpArgs {
...
+    // It would be cool to follow the docs hierarchy somehow.
+    //
+    // One of the problems would be `config.md`, as it has the same name as a
+    // subcommand.
+    let categories = BTreeMap::from([
+        (
+            "revsets",
+            Category {
+                description: "A functional language for selecting a set of revision",
+                content: include_str!("../../../docs/revsets.md"),
```
martinvonz, line 53: I had hoped that we would be able to render the markdown in the terminal using ANSI escape codes somehow. I guess the source is fine to start with. Add a TODO about rendering it?

It's unfortunate that there are so many flavors of markdown and that we'll have to find something that's able to format the same markdown source as MkDocs is able to format for the website.

(Related to @thoughtpolice's PR-level comment)

yuja: OT: the markdown renderer could also be a template function so that user can do `markdown(description)`.

martinvonz: Interesting. So the raw template output (before formatting) would include color codes then? That seems a bit weird, but I guess it works (we don't escape escapes in the formatter input, do we?).

yuja: We do escapes. I assumed the renderer would emit labels, not ANSI escapes.

martinvonz: Ah, of course. Good idea.

cli/src/commands/help.rs
```diff
@@ -28,22 +32,98 @@ pub(crate) struct HelpArgs {
...
+
+    let subcommand_max_len = command
+        .app()
+        .get_subcommands()
+        .map(|cmd| cmd.get_name().len())
+        .max()
+        .unwrap();
+
+    let help_template_categories = format!(
+        "\x1b[1m\x1b[4mHelp Categories:\x1b[0m\n{}",
```
martinvonz, line 90: This is pretty hard to read. It's making the text bold, right? Can we use the formatter (`ui.stdout_formatter()`) instead? Then it will also respect `NO_COLOR` and `ui.color`.

cli/src/commands/help.rs
Grillo-0: Is that what you meant?

cli/src/commands/help.rs
Grillo-0: With my testing, it seems to work only on the first level, but I couldn't find a statement saying that it will, or will not, apply it recursively through the subcommands.

yuja: I don't think the template option applies recursively, but it's clearer if we do that explicitly.

Anyway, I would remove all templating stuff and just use `after_long_help`.

cli/src/commands/help.rs
Grillo-0: Wouldn't be better to use `ui.new_formatter()`?

cli/src/commands/help.rs
yuja: I meant something like:
```rust
const CATEGORIES: .. = &[
];

// will be embedded as #[command(after_long_help = help::format_categories_list())]
pub fn format_categories_list() -> String { // or clap's StyledStr?
    ...
}

// could be inilined
fn find_category(name) -> Option<&Category> {
    CATEGORIES.iter().find(..)
}
```

cli/src/commands/help.rs
martinvonz: Oh, you just want a string/bytes here? Yes, `ui.new_formatter()` is better then.

yuja: It's probably easier to use crossterm API (or clap's styling functions if available.) clap will strip ANSI escapes when color is disabled.
https://docs.rs/crossterm/latest/crossterm/style/trait.Stylize.html

cli/src/commands/help.rs
Grillo-0: Added a TODO for that, referencing #3130. I think that the work of @thoughtpolice has some great ideas. As he said, I also think that might be a future PR improving this with some of his ideas.

CHANGELOG.md
```diff
@@ -62,6 +62,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Timestamp objects in templates now have `after(date) -> Boolean` and
   `before(date) -> Boolean` methods for comparing timestamps to other dates.
 
+* `jj help` now can give help for some categories, like `jj help revsets`.
```
joyously, line 65: I don't think that the categories should be at the first level of help, especially since some of the names will be the same as revset functions or template functions or even aliases.
If you look at Git, some commands are listed and at the bottom is
```
'git help -a' and 'git help -g' list available subcommands and some
concept guides. See 'git help <command>' or 'git help <concept>'
to read about a specific subcommand or concept.
See 'git help git' for an overview of the system.
```
If you look at Breezy, a few commands are listed and at the bottom is
```
  brz help init      more help on e.g. init command
  brz help commands  list all commands
  brz help topics    list all help topics
```

cli/src/commands/help.rs
```diff
@@ -42,8 +55,70 @@ pub(crate) fn cmd_help(
...
+        .find(|(cat_name, _)| *cat_name == name)
+        .map(|tuple| &tuple.1)
+}
+
+fn format_categories(command: &clap::Command) -> String {
+    let subcommand_max_len = command
+        .get_subcommands()
+        .map(|cmd| cmd.get_name().len())
+        .max()
+        .unwrap();
```
yuja, line 113: The `Categories`list doesn't have to be aligned with `Commands` list. I would use hard-coded width or max length of `cat_name`.

CHANGELOG.md
yuja: I don't have a strong opinion, but it seems also good to mention `jj help --help` or `jj help --list-categories` in the top-level help. Then, the categories list can be removed from the top-level help.

cli/src/commands/help.rs
```diff
@@ -42,8 +55,70 @@ pub(crate) fn cmd_help(
         .app()
         .clone()
         .subcommand_required(true)
+        .after_help(format_categories(command.app()))
```
yuja, line 58: I would move this to `enum Command` attribute so `jj --help` shows the identical content.

yuja (approved):
Thanks.

CHANGELOG.md
martinvonz: I don't have a strong opinion either, but we should probably decide before we add support for topics and add a top-level `jj help topics` about that. We should decide if we want `jj help topics` to be about the "topics" feature or if we want `jj help topics topics` to be for that.

yuja: I feel `jj help topics <TOPIC>` is too verbose. I prefer `jj help <--command|--topic> NAME..` (like `man [PAGE] NAME`) or `jj help <--topic=NAME|COMMAND..>` instead.

Grillo-0: I prefer the `--topic` flag too. Just to be clear, we prefer `topic` over `category`?

emilazy: I think “topic” is going to be really confusing once topics are a thing.

Grillo-0: Are topics different from categories? I'm a bit confused here XD

yuja: We're going to introduce "topic" branches concept.

`--category` sounds also good, but it's short form would conflict with `-c|--command` if we add it. Mercurial uses `-k|--keyword`.

Grillo-0: Ohhh, I see. Where can learn more about this feature?

I'm fine changing to `keyword` to have the flag.

joyously: Git uses `--g` for `guides`.

yuja: > Where can learn more about this feature?
https://github.com/martinvonz/jj/issues/3402

Grillo-0: I think `keyword` is more generic.

Grillo-0: I made big changes, `s/categories/keywords/` and now we have a boolean flag '--keyword' to select between help for commands or keywords.

cli/src/commands/help.rs
```diff
@@ -12,38 +12,158 @@
...
+
+use clap::builder::StyledStr;
+use crossterm::style::Stylize;
+use itertools::Itertools;
 use tracing::instrument;
 
 use crate::cli_util::CommandHelper;
-use crate::command_error;
 use crate::command_error::CommandError;
+use crate::command_error::{self};
```
yuja, line 25: nit: bad formatting change

cli/src/commands/help.rs
```diff
@@ -12,38 +12,158 @@
...
 use crate::ui::Ui;
 
 /// Print this message or the help of the given subcommand(s)
 #[derive(clap::Args, Clone, Debug)]
 pub(crate) struct HelpArgs {
     /// Print help for the subcommand(s)
     pub(crate) command: Vec<String>,
+    // Show help for keywords instead of commands
+    #[arg(long, short = 'k')]
+    keyword: bool,
```
yuja, line 35: Since `-k` is required in order to show docs, it probably makes sense to add `--keyword=KEYWORD` option instead of boolean `-k`.

i.e. `keyword: Option<String>` or `keyword: Option<Option<String>>` (if `-k` should be a flag to show list of keywords.) I personally prefer a separate `--list-keywords` flag than overloading `--keyword`, but I don't have a strong opinion about that.

Then, we can use `arg(.., value_parser = ["foo", "bar", ..])` to let clap provide suggestion.
https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser

cli/src/commands/help.rs
```diff
@@ -12,38 +12,158 @@
...
+    ret.into()
+}
+
+pub fn format_keywords_for_after_help() -> StyledStr {
+    let mut ret = StyledStr::new();
+    writeln!(
+        ret,
+        "Use {} to show help related to these keywords",
+        "'jj help -k'".bold()
+    )
```
yuja, line 165: Since we have a flag to show the list, maybe we can remove it from the top-level help? I think "Use jj help -k ..." is good enough for discoverability.

Then, we can use our `ui.formatter()` API to print the list of keywords.

cli/src/commands/help.rs
Grillo-0: I thought about adding the flag as `--keyword=KEYWORD`, but then I thought that the user could do something like `jj help --keyword=revsets log` which I found ambiguous.

yuja: Invalid combinations can be rejected by `conflicts_with = "command"` or `ArgGroup`.

martinvonz:
@Grillo-0: I see that you requested my review. I find it hard to find enough time to review many PRs lately. I'm happy to trust Yuya's review. Let me know if you really want me to review it too.

cli/src/command_error.rs
```diff
@@ -207,7 +207,7 @@ pub fn internal_error_with_message(
     CommandError::with_message(CommandErrorKind::Internal, message, source)
 }
 
-fn format_similarity_hint<S: AsRef<str>>(candidates: &[S]) -> Option<String> {
+pub fn format_similarity_hint<S: AsRef<str>>(candidates: &[S]) -> Option<String> {
```
yuja, line 210: nit: this change is no longer needed

cli/src/commands/help.rs
```diff
@@ -47,3 +79,74 @@ pub(crate) fn cmd_help(
...
+        description: "Show a tutorial to get started with jj",
+        content: include_str!("../../../docs/tutorial.md"),
+    },
+];
+
+fn find_keyword(name: &str) -> Option<&Keyword> {
+    KEYWORDS.iter().find(|keyword| keyword.name == name)
+}
+
+fn write_keyword_list(mut formatter: Box<dyn Formatter + '_>) -> io::Result<()> {
```
yuja, line 120: nit: `&mut dyn Formatter` because we don't need an owned formatter

cli/src/commands/help.rs
```diff
@@ -47,3 +79,74 @@ pub(crate) fn cmd_help(
...
+
+    formatter.push_label("bold_text")?;
+    formatter.push_label("underlined_text")?;
+    writeln!(formatter, "Help Keywords:")?;
+    formatter.pop_label()?;
+    formatter.pop_label()?;
+    for keyword in KEYWORDS {
+        formatter.push_label("bold_text")?;
+        write!(formatter, "  {:<keyword_name_max_len$}  ", keyword.name)?;
+        formatter.pop_label()?;
```
yuja, line 135: `formatter.labeled("help_keyword")` or `formatter.with_label("help", |formatter| { ... formatter.labeled("keyword") .. })`?

Labels usually describe the content, not the style.

cli/src/commands/help.rs
```diff
@@ -47,3 +79,74 @@ pub(crate) fn cmd_help(
...
+        .iter()
+        .map(|keyword| keyword.name.len())
+        .max()
+        .unwrap();
+
+    formatter.push_label("bold_text")?;
+    formatter.push_label("underlined_text")?;
+    writeln!(formatter, "Help Keywords:")?;
+    formatter.pop_label()?;
+    formatter.pop_label()?;
```
yuja, line 131: Maybe we can remove "Help Keywords:" header? Since we no longer embed the list in `jj help`, we don't have to copy the style of the clap output.

cli/tests/test_help_command.rs
```diff
@@ -84,3 +84,82 @@ fn test_help() {
...
+    // Capture 'Help Keywords' part
+    let help_cmd_stdout = test_env.jj_cmd_success(test_env.env_root(), &["help"]);
+    insta::with_settings!({filters => vec![
+        (r"(?s).+\n\n(?<h>.+)", "$h"),
+    ]}, {
+        insta::assert_snapshot!(help_cmd_stdout, @r#"
+        'jj help --list-keywords' list available keywords. Use 'jj help -k' to show help for one of these
+        keywords.
+        "#);
+    });
```
yuja, line 101: nit: I think this test can be removed because the list isn't included here.

cli/src/commands/help.rs
```diff
@@ -12,26 +12,58 @@
...
 /// Print this message or the help of the given subcommand(s)
 #[derive(clap::Args, Clone, Debug)]
 pub(crate) struct HelpArgs {
     /// Print help for the subcommand(s)
     pub(crate) command: Vec<String>,
+    /// Show help for keywords instead of commands
+    #[arg(
+        long,
+        short = 'k',
+        exclusive = true,
```
yuja, line 39: `exclusive` conflicts with any other flags including `--color=always`, etc.

cli/src/commands/help.rs
yuja: BTW, another idea is to use `PossibleValue::new(k.name).help(k.description)` and suggest `jj help --help` instead of implementing `jj help --list-keywords`.
https://docs.rs/clap/latest/clap/builder/struct.PossibleValue.html

CHANGELOG.md
```diff
@@ -81,6 +81,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New command `jj config unset` that unsets config values. For example,
   `jj config unset --user user.name`.
 
+* `jj help` now has the flag `--keyword`(shorthand `-k`), which can give help
+  for some keywords(e.g. `jj help -k revsets`). To see a list of the available
```
yuja, line 85: nit: missing space after `for some keywords` ?

yuja (approved):
Thanks.

Grillo-0:
I will merge this

cli/src/commands/help.rs
```diff
@@ -47,3 +73,52 @@ pub(crate) fn cmd_help(
...
+// subcommand.
+//
+// TODO: Find a way to render markdown using ANSI escape codes.
+//
+// Maybe we can steal some ideas from https://github.com/martinvonz/jj/pull/3130
+const KEYWORDS: &[Keyword] = &[
+    Keyword {
+        name: "revsets",
+        description: "A functional language for selecting a set of revision",
+        content: include_str!("../../../docs/revsets.md"),
```
martinvonz, line 101: Does anyone know if this pointer outside of the crate will work when we publish to crates.io (as I'm supposed to do on wednesday)?

yuja: Good catch. It could be worked around by a symlink (`ln -s ../docs cli`), but symlink has its own problem, so I have no idea.

thoughtpolice: I somewhat doubt it will, I found a rather [confusing thread](https://users.rust-lang.org/t/include-str-does-not-work-when-releasing-because-of-changed-pathes/15551) about it.

We should probably instead use `build.rs` in order to get access to `OUT_DIR` and just copy the files during build time somewhere. A minor annoyance but nothing too horrid, I guess (though I'd need to figure out how to port the Buck changes.)

martinvonz: Do you mean that we should make `build.rs` copy the files from `docs/` into e.g. `cli/srcs/docs/` so we have two copies of the files?

yuja: Perhaps, what we can do with `build.rs` is to fall back to `../docs` if `./docs` doesn't exist? We'll need `docs` symlink or something in order to include `docs/*` in the release package.

PR #4632, opened 2024-10-13
revset: drop support for HEAD@git symbol resolution

yuja:
# Checklist

If applicable:


CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Evaluation error of `revsets.short-prefixes` configuration is now reported.
 
 * The `HEAD@git` symbol no longer resolves to the Git HEAD revision. Use
-  `git_head()` or `@-` revset expression instead.
+  `git_head()` or `@-` revset expression instead. The `git_head` template
+  keyword now returns a boolean.
```
martinvonz, line 20: From what I understand of #4588, we'll soon replace `git_head()` by `git_heads()`. It would be nice to not change `HEAD@git` to `git_head()` between 0.22 and 0.23 and then from `git_head()` to `git_heads()` between 0.23 an 0.24. I don't think it's a big deal for the log output, but for inputs in revsets, it would be good to not make users update any configs they have twice. So I think this PR should wait until at least the `git_head()` -> `git_heads()` rename from @cormacrelf's PR is ready (then this PR can land and the `git_head()` -> `git_head()` rename can land soon thereafter). Makes sense?

martinvonz (approved):

CHANGELOG.md
cormacrelf: It's not a rename, I just added git_heads() in order to test the changes to the view storage. The singular version is still there.

yuja: As far as I know, `git_head()` won't be deprecated. It returns the `HEAD` of the current workspace. `git_heads()` will resolve to `HEAD`s of the all workspaces?

martinvonz: I see. Sounds like the subject line of https://github.com/martinvonz/jj/pull/4588/commits/0aca8bea554246df53a05aee189fe3a557100a9e should be updated then.

martinvonz: Oh, wait, that's the *template*, not the revset. Sorry!

yuja: It's `view.git_head: RefTarget`. FWIW, it might be better to leave this PR open until we're close to merge the colocated worktrees feature.

martinvonz: If we're going to keep `git_head()` (and maybe add a `git_heads()` too?), I don't think this PR needs to wait for the other PR(s) to be ready. Feel free to merge it as far as I'm concerned.

PR #4638, opened 2024-10-14
docs/FAQ: fix incorrect change ID for commit removed as parent

cenviity:
# Checklist

If applicable:
~- [ ] I have updated `CHANGELOG.md`~
~- [ ] I have updated the documentation (README.md, docs/, demos/)~
~- [ ] I have updated the config schema (cli/src/config-schema.json)~
~- [ ] I have added tests to cover my changes~

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -219,8 +219,9 @@ Now you're ready to work:
 - The private commit _wwwwwwww_ is the second parent of the merge commit.
 - The working copy (_vvvvvvvv_) contains changes from both.
 
-As you work, squash your changes using `jj squash --into xxxxxxxx`. Or, you can
-keep your changes in a separate commit and remove _ttsqqnrx_ as a parent:
+As you work, squash your changes using `jj squash --into xxxxxxxx`. Or you can
+keep your changes in a separate commit without the private commit _wwwwwwww_ as a
+parent:
 
 ```shell
 # Remove the private commit as a parent


docs/FAQ.md
```diff
@@ -219,8 +219,9 @@ Now you're ready to work:
 - The private commit _wwwwwwww_ is the second parent of the merge commit.
 - The working copy (_vvvvvvvv_) contains changes from both.
 
-As you work, squash your changes using `jj squash --into xxxxxxxx`. Or, you can
-keep your changes in a separate commit and remove _ttsqqnrx_ as a parent:
+As you work, squash your changes using `jj squash --into xxxxxxxx`. Or you can
+keep your changes in a separate commit without the private commit _wwwwwwww_ as a
+parent:
```
martinvonz, line 224: @jennings, should the `jj new vvvvvvvv wwwwwwww` also be changed to `jj new xxxxxxxx vvvvvvvv`? Otherwise I don't see the point of the command. Maybe the description should also be changed to "Recreate the merge commit to work in"?

jennings: I don't think so? After the rebase command below, `vvvvvv` and `wwwwww` are the heads, and `vvvvvv` has work you want to publish.

So, `jj new vvvvvvv wwwwwww` gets you a new merge commit to work in.

Maybe I'm reading it wrong? I thought these single-letter change IDs would make it easier to follow, but clearly it can be improved. :)

martinvonz: Oh, I see what you mean now. I thought the below was if you decided to not include the private changes anymore, but that's not what it's about. It's describing how to move your current working-copy changes to a new commit on top of `xxxxxxxx`. Another option for that is `jj new; jj rebase -r @- -A xxxxxxxx`. That seems a bit more intuitive to me. Anyway, this is off topic for this PR, so feel free to merge, @cenviity.

martinvonz (approved):

docs/FAQ.md
cenviity: Thanks!

PR #4640, opened 2024-10-14
 rebase -s: add support for `--insert-after` and `--insert-before`

bnjmnt4n:
Supercedes https://github.com/martinvonz/jj/pull/3607.

# Checklist

If applicable:


martinvonz (commented):
Neat! Looks good to me, but I'll leave approval to Yuya since he reviewed the previous incarnation of this PR.

lib/src/rewrite.rs
```diff
@@ -467,18 +467,21 @@ pub struct MoveCommitsStats {
...
 ///
-/// The roots of `target_commits` are rebased onto the new parents
+/// Commits in `target_roots` are rebased onto the new parents
 /// given by `new_parent_ids`, while the `new_children` commits are
-/// rebased onto the heads of `target_commits`. This assumes that
-/// `target_commits` and `new_children` can be rewritten, and there
-/// will be no cycles in the resulting graph. `target_commits` should
-/// be in reverse topological order.
+/// rebased onto the heads of `target_commits`. If `target_roots` is
+/// empty, it will be computed as the roots of the connected set of
```
yuja, line 473: nit: empty -> `None`

Do we have other expectations to document (such as `target_roots` should be a subset of `target_commits`)?

cli/src/commands/rebase.rs
```diff
@@ -241,60 +241,16 @@ pub(crate) fn cmd_rebase(
...
-            )?;
-        }
+        rebase_revisions(
+            ui,
+            command.settings(),
+            &mut workspace_command,
+            &args.revisions,
+            &args.destination,
+            &args.insert_after,
+            &args.insert_before,
```
yuja, line 249: nit: it's probably better to pass `args` or its subset in to `rebase_revisions()` and `compute_rebase_destination()`.

For example, (maybe as a follow up,) `destination`, `insert_after`, and `insert_before` can be extracted to `sturct RebaseDestinationArgs` and flattened into `RebaseArgs`.

cli/src/commands/rebase.rs
```diff
@@ -451,164 +453,74 @@ fn rebase_descendants_transaction(
...
-    let before_commit_ids = before_commits.iter().ids().cloned().collect_vec();
-    workspace_command.check_rewritable(&before_commit_ids)?;
-
-    let after_commit_ids = after_commits.iter().ids().cloned().collect_vec();
-    let new_children_expression = RevsetExpression::commits(before_commit_ids);
-    let new_parents_expression = RevsetExpression::commits(after_commit_ids.clone());
+    };
+    let destination_commits = resolve_revisions(destination)?;
+    let after_commits = resolve_revisions(insert_after)?;
+    let before_commits = resolve_revisions(insert_before)?;
```
yuja, line 506: nit: (as a follow up) these args can be changed to `Option<Vec<Revisionarg>>` type. Then, `.is_empty()` can be replaced with match of `Some(_)`s.

yuja (approved):
Thanks!

cli/src/commands/rebase.rs
bnjmnt4n: I added a new commit to address this.

cli/src/commands/rebase.rs
```diff
@@ -481,28 +475,27 @@ fn rebase_descendants_transaction(
...
+                    .resolve_some_revsets_default_single(ui, revisions)?
+                    .into_iter()
+                    .collect_vec())
+            } else {
+                Ok(vec![])
+            }
+        };
+    let destination_commits = resolve_revisions(&rebase_destination.destination)?;
+    let after_commits = resolve_revisions(&rebase_destination.insert_after)?;
+    let before_commits = resolve_revisions(&rebase_destination.insert_before)?;
```
yuja, line 498: nit: I meant these `Option<&[RevisionArg]>`s can be mapped to `Option<Vec<Commit>>`s so the following `if` block can be a single `match` statement.

Issue #4624, opened 2024-10-11
FR: Add a `jj git colocate` command 


PhilipMetzger:
**Is your feature request related to a problem? Please describe.**

We've repeatedly had the request in the Discord for a command which eases the transition from a bare Git backed repo, to a colocated one. This command should be responsible for doing that. 

**Implementation Steps**

Add a new file under named `colocate.rs` under `cli/src/commands/git/` (don't forget to add `mod colocate` to the `mod.rs`). 
Implement these steps with code: https://martinvonz.github.io/jj/latest/git-compatibility/#converting-a-repo-into-a-co-located-repo

estk:
@PhilipMetzger how does this differ from `jj git init --colocate`?

essiene:
> @PhilipMetzger how does this differ from `jj git init --colocate`?

IIUC, this takes a *jj* repo that is backed by git and create with `jj git init` and turns it into a colocated repo.

It wasn't very clear from the initial description until I followed the link.

PhilipMetzger:
> > @PhilipMetzger how does this differ from `jj git init --colocate`?
> 
> IIUC, this takes a **jj git backed repo** presumably created by `jj git init` and turns it into a colocated repo.

This is the correct, I'll edit the description.

PhilipMetzger:
This was completed by #7392 but GH didn't do its thing again.

PR #2026, opened 2023-08-10
cli: detect .git symlink as a colocated workspace

yuja:
#2011

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (src/config-schema.json)
- [x] I have added tests to cover my changes

PR #4644, opened 2024-10-14
Colocated workspaces, minimal

cormacrelf:
Only the worktree recognition and HEAD reading/writing parts of #4588.

In other PRs:
- The CLI changes: create/destroy git worktrees during `jj workspace add`/`forget`
- The view changes to track multiple git heads and make the `git_head()` revset work correctly

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -835,7 +835,7 @@ impl WorkspaceCommandHelper {
         let op_summary_template_text = settings.config().get_string("templates.op_summary")?;
         let may_update_working_copy =
             loaded_at_head && !env.command.global_args().ignore_working_copy;
-        let working_copy_shared_with_git = is_colocated_git_workspace(&workspace, &repo);
+        let working_copy_shared_with_git = is_colocated_git_workspace(Some(ui), &workspace, &repo);
         let helper = Self {
             workspace,
             user_repo: ReadonlyUserRepo::new(repo),
@@ -878,7 +878,7 @@ impl WorkspaceCommandHelper {
     }
 
     /// Snapshot the working copy if allowed, and import Git refs if the working
-    /// copy is collocated with Git.
+    /// copy is colocated with Git.
     #[instrument(skip_all)]
     pub fn maybe_snapshot(&mut self, ui: &Ui) -> Result<(), CommandError> {
         if self.may_update_working_copy {
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -386,6 +386,10 @@ impl CommandHelper {
         WorkspaceCommandHelper::new(ui, workspace, repo, env, self.is_at_head_operation())
     }
 
+    pub fn get_store_factories(&self) -> &StoreFactories {
+        &self.data.store_factories
+    }
+
     pub fn get_working_copy_factory(&self) -> Result<&dyn WorkingCopyFactory, CommandError> {
         let loader = self.workspace_loader()?;
 
--- a/cli/src/git_util.rs
+++ b/cli/src/git_util.rs
@@ -73,12 +73,72 @@ pub fn get_git_repo(store: &Store) -> Result<git2::Repository, CommandError> {
     }
 }
 
-pub fn is_colocated_git_workspace(_workspace: &Workspace, repo: &ReadonlyRepo) -> bool {
+pub fn is_colocated_git_workspace(
+    ui: Option<&Ui>,
+    workspace: &Workspace,
+    repo: &ReadonlyRepo,
+) -> bool {
     let Some(git_backend) = repo.store().backend_impl().downcast_ref::<GitBackend>() else {
         return false;
     };
 
-    git_backend.is_colocated()
+    if git_backend.is_colocated() {
+        return true;
+    }
+
+    if let Some(ui) = ui {
+        let workspace_dot_git = workspace.workspace_root().join(".git");
+        let Ok(meta) = workspace_dot_git.symlink_metadata() else {
+            return git_backend.is_colocated();
+        };
+        // Check symlink first as is_dir follows symlinks
+        if meta.is_symlink() {
+            let target = std::fs::read_link(&workspace_dot_git)
+                .unwrap_or_else(|_| Path::new("<could not read link>").to_path_buf());
+            writeln!(
+                ui.warning_default(),
+                "Workspace has a .git symlink (pointing to {}), that isn't pointing to JJ's git \
+                 repo",
+                target.display()
+            )
+            .ok();
+        } else if meta.is_dir() {
+            writeln!(
+                ui.warning_default(),
+                "Workspace has a .git directory that is not managed by JJ"
+            )
+            .ok();
+        } else if let Ok(_worktree_repo) = gix::open(workspace.workspace_root()) {
+            writeln!(
+                ui.warning_default(),
+                "Workspace is also a Git worktree that is not managed by JJ",
+            )
+            .ok();
+        } else {
+            let gitfile_contents = std::fs::read_to_string(&workspace_dot_git)
+                .unwrap_or_else(|_| "<could not read .git file>".to_string());
+            let gitdir = gitfile_contents
+                .strip_prefix("gitdir: ")
+                .unwrap_or(&gitfile_contents)
+                .trim();
+            writeln!(ui.warning_default(), "Workspace is a broken Git worktree").ok();
+            writeln!(ui.warning_no_heading(), "The .git file points at: {gitdir}").ok();
+            // work_dir may return Some even if colocated = false, because we may be in some
+            // secondary workspace, and we may have just opened the colocated primary
+            // workspace with a .git directory.
+            let git_repo = git_backend.git_repo();
+            let repair_dir = git_repo.work_dir().unwrap_or(git_repo.common_dir());
+            writeln!(
+                ui.hint_default(),
+                "If this is meant to be a colocated JJ workspace, you may like to try `git -C {} \
+                 worktree repair`",
+                repair_dir.display()
+            )
+            .ok();
+        }
+    }
+
+    false
 }
 
 fn terminal_get_username(ui: &Ui, url: &str) -> Option<String> {
--- a/cli/src/git_util.rs
+++ b/cli/src/git_util.rs
@@ -73,22 +73,12 @@ pub fn get_git_repo(store: &Store) -> Result<git2::Repository, CommandError> {
     }
 }
 
-pub fn is_colocated_git_workspace(workspace: &Workspace, repo: &ReadonlyRepo) -> bool {
+pub fn is_colocated_git_workspace(_workspace: &Workspace, repo: &ReadonlyRepo) -> bool {
     let Some(git_backend) = repo.store().backend_impl().downcast_ref::<GitBackend>() else {
         return false;
     };
-    let Some(git_workdir) = git_backend.git_workdir() else {
-        return false; // Bare repository
-    };
-    if git_workdir == workspace.workspace_root() {
-        return true;
-    }
-    // Colocated workspace should have ".git" directory, file, or symlink. Compare
-    // its parent as the git_workdir might be resolved from the real ".git" path.
-    let Ok(dot_git_path) = workspace.workspace_root().join(".git").canonicalize() else {
-        return false;
-    };
-    git_workdir.canonicalize().ok().as_deref() == dot_git_path.parent()
+
+    git_backend.is_colocated()
 }
 
 fn terminal_get_username(ui: &Ui, url: &str) -> Option<String> {
--- a/cli/tests/test_git_colocated.rs
+++ b/cli/tests/test_git_colocated.rs
@@ -791,6 +791,7 @@ fn get_log_output(test_env: &TestEnvironment, workspace_root: &Path) -> String {
       commit_id,
       bookmarks,
       if(git_head, "git_head()"),
+      working_copies,
       description,
     )
     "#;
@@ -806,6 +807,7 @@ fn get_log_output_with_stderr(
       commit_id,
       bookmarks,
       if(git_head, "git_head()"),
+      working_copies,
       description,
     )
     "#;
@@ -887,4 +889,35 @@ fn test_git_colocated_unreachable_commits() {
 fn get_bookmark_output(test_env: &TestEnvironment, repo_path: &Path) -> String {
     // --quiet to suppress deleted bookmarks hint
     test_env.jj_cmd_success(repo_path, &["bookmark", "list", "--all-remotes", "--quiet"])
+}
+
+#[test]
+fn test_git_colocated_create_workspace_moving_wc() {
+    let test_env = TestEnvironment::default();
+    let repo_path = test_env.env_root().join("repo");
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "--colocate", "repo"]);
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "second_wc_parent"]);
+    let second_wc_parent =
+        test_env.jj_cmd_success(&repo_path, &["log", "-Tcommit_id", "-r@-", "--no-graph"]);
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "should be git head"]);
+    test_env.jj_cmd_ok(
+        &repo_path,
+        &["workspace", "add", "../second", "-r", &second_wc_parent],
+    );
+
+    // The second workspace is not colocated. So creating it should not
+    // move git_head(). This tests whether we managed to reload the workspace
+    // command effectively in the middle of `jj workspace add`; if we did not,
+    // it will still think it's colocated in the default workspace.
+    //
+    // This test should survive switching to multi-git-head views, where each
+    // workspace gets its own git_head().
+    insta::assert_snapshot!(get_log_output(&test_env, &repo_path), @r"
+    ○  3ee64208816f8264850a1cfcb99ac75e878c207a second@
+    │ @  076d128fc75cb291f908b77ffe9ff24c80de4fd7 default@
+    │ ○  005cf6c5b5e5fd032b804d98d11212a5317b4f3a git_head() should be git head
+    ├─╯
+    ○  ad5ff99a9ee8726ffe511430e4569f1d57bd550e second_wc_parent
+    ◆  0000000000000000000000000000000000000000
+    ");
 }
--- a/cli/tests/test_git_colocated.rs
+++ b/cli/tests/test_git_colocated.rs
@@ -975,3 +975,354 @@ fn stopgap_workspace_colocate(
         .assert()
         .success();
 }
+
+#[test]
+fn test_colocated_workspace_in_bare_repo() {
+    // TODO: Remove when this stops requiring git (stopgap_workspace_colocate)
+    if Command::new("git").arg("--version").status().is_err() {
+        eprintln!("Skipping because git command might fail to run");
+        return;
+    }
+
+    let test_env = TestEnvironment::default();
+    let repo_path = test_env.env_root().join("repo");
+    let second_path = test_env.env_root().join("second");
+    //
+    // git init without --colocate creates a bare repo
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    std::fs::write(repo_path.join("file"), "contents").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "initial commit"]);
+    let (initial_commit, _) = test_env.jj_cmd_ok(
+        &repo_path,
+        &["log", "--no-graph", "-T", "commit_id", "-r", "@-"],
+    );
+    // TODO: replace with workspace add, when it can create worktrees
+    stopgap_workspace_colocate(&test_env, &repo_path, false, "../second", &initial_commit);
+
+    insta::assert_snapshot!(get_log_output(&test_env, &second_path), @r#"
+    @  baf7f13355a30ddd3aa6476317fcbc9c65239b0c second@
+    │ ○  45c9d8477181a2b9c077ff1b724694fe0969b301 default@
+    ├─╯
+    ○  046d74c8ab0a4730e58488508a5398b7a91e54a2 git_head() initial commit
+    ◆  0000000000000000000000000000000000000000
+    "#);
+
+    test_env.jj_cmd_ok(
+        &second_path,
+        &["commit", "-m", "commit in second workspace"],
+    );
+    insta::assert_snapshot!(get_log_output(&test_env, &second_path), @r#"
+    @  fca81879c29229d0097cb7d32fc8a661ee80c6e4 second@
+    ○  220827d1ceb632ec7dd4cb2f5110b496977d14c2 git_head() commit in second workspace
+    │ ○  45c9d8477181a2b9c077ff1b724694fe0969b301 default@
+    ├─╯
+    ○  046d74c8ab0a4730e58488508a5398b7a91e54a2 initial commit
+    ◆  0000000000000000000000000000000000000000
+    "#);
+
+    // FIXME: There should still be no git HEAD in the default workspace, which
+    // is not colocated. However, git_head() is a property of the view. And
+    // currently, all colocated workspaces read and write from the same
+    // entry of the common view.
+    //
+    // let stdout = test_env.jj_cmd_success(&repo_path, &["log", "--no-graph",
+    // "-r", "git_head()"]); insta::assert_snapshot!(stdout, @r#""#);
+
+    let stdout = test_env.jj_cmd_success(
+        &second_path,
+        &["op", "log", "-Tself.description().first_line()"],
+    );
+    insta::assert_snapshot!(stdout, @r#"
+    @  commit baf7f13355a30ddd3aa6476317fcbc9c65239b0c
+    ○  import git head
+    ○  create initial working-copy commit in workspace second
+    ○  add workspace 'second'
+    ○  commit 4e8f9d2be039994f589b4e57ac5e9488703e604d
+    ○  snapshot working copy
+    ○  add workspace 'default'
+    ○
+    "#);
+}
+
+#[test]
+fn test_colocated_workspace_moved_original_on_disk() {
+    if Command::new("git").arg("--version").status().is_err() {
+        eprintln!("Skipping because git command might fail to run");
+        return;
+    }
+
+    let test_env = TestEnvironment::default();
+    let repo_path = test_env.env_root().join("repo");
+    let second_path = test_env.env_root().join("second");
+    let new_repo_path = test_env.env_root().join("repo-moved");
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "--colocate", "repo"]);
+    std::fs::write(repo_path.join("file"), "contents").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "initial commit"]);
+    let (initial_commit, _) = test_env.jj_cmd_ok(
+        &repo_path,
+        &["log", "--no-graph", "-T", "commit_id", "-r", "@-"],
+    );
+    // TODO: replace with workspace add, when it can create worktrees
+    stopgap_workspace_colocate(&test_env, &repo_path, true, "../second", &initial_commit);
+
+    // Break our worktree by moving the original repo on disk
+    std::fs::rename(&repo_path, &new_repo_path).unwrap();
+    // imagine JJ were able to do this
+    std::fs::write(
+        second_path.join(".jj/repo"),
+        new_repo_path
+            .join(".jj/repo")
+            .as_os_str()
+            .as_encoded_bytes(),
+    )
+    .unwrap();
+
+    let (_, stderr) = test_env.jj_cmd_ok(&second_path, &["status"]);
+    // hack for windows paths
+    let gitfile_contents = std::fs::read_to_string(second_path.join(".git"))
+        .unwrap()
+        .strip_prefix("gitdir: ")
+        .unwrap()
+        .trim()
+        .to_owned();
+    let stderr = stderr.replace(&gitfile_contents, "$TEST_ENV/repo/.git/worktrees/second");
+    insta::assert_snapshot!(stderr, @r"
+    Warning: Workspace is a broken Git worktree
+    The .git file points at: $TEST_ENV/repo/.git/worktrees/second
+    Hint: If this is meant to be a colocated JJ workspace, you may like to try `git -C $TEST_ENV/repo-moved worktree repair`
+    ");
+
+    Command::new("git")
+        .args(["worktree", "repair"])
+        .current_dir(&new_repo_path)
+        .assert()
+        .success();
+    insta::assert_snapshot!(get_log_output(&test_env, &second_path), @r#"
+    @  05530a3e0f9d581260343e273d66c381e76957df second@
+    │ ○  45c9d8477181a2b9c077ff1b724694fe0969b301 default@
+    ├─╯
+    ○  046d74c8ab0a4730e58488508a5398b7a91e54a2 git_head() initial commit
+    ◆  0000000000000000000000000000000000000000
+    "#);
+}
+
+#[test]
+fn test_colocated_workspace_wrong_gitdir() {
+    // TODO: Remove when this stops requiring git (stopgap_workspace_colocate)
+    if Command::new("git").arg("--version").status().is_err() {
+        eprintln!("Skipping because git command might fail to run");
+        return;
+    }
+
+    let test_env = TestEnvironment::default();
+    let repo_path = test_env.env_root().join("repo");
+    let second_path = test_env.env_root().join("second");
+    let other_path = test_env.env_root().join("other");
+    let other_second_path = test_env.env_root().join("other_second");
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "--colocate", "repo"]);
+    std::fs::write(repo_path.join("file"), "contents").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "initial commit"]);
+    let (initial_commit, _) = test_env.jj_cmd_ok(
+        &repo_path,
+        &["log", "--no-graph", "-T", "commit_id", "-r", "@-"],
+    );
+    // TODO: replace with workspace add, when it can create worktrees
+    stopgap_workspace_colocate(&test_env, &repo_path, true, "../second", &initial_commit);
+
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "--colocate", "other"]);
+    std::fs::write(other_path.join("file"), "contents2").unwrap();
+    test_env.jj_cmd_ok(&other_path, &["commit", "-m", "initial commit"]);
+    let (ic_other, _) = test_env.jj_cmd_ok(
+        &other_path,
+        &["log", "--no-graph", "-T", "commit_id", "-r", "@-"],
+    );
+    // TODO: replace with workspace add, when it can create worktrees
+    stopgap_workspace_colocate(&test_env, &other_path, true, "../other_second", &ic_other);
+
+    // Break one of our worktrees
+    std::fs::copy(other_second_path.join(".git"), second_path.join(".git")).unwrap();
+
+    let (_, stderr) = test_env.jj_cmd_ok(&second_path, &["status"]);
+    insta::assert_snapshot!(stderr, @"Warning: Workspace is also a Git worktree that is not managed by JJ");
+}
+
+#[test]
+fn test_colocated_workspace_invalid_gitdir() {
+    // TODO: Remove when this stops requiring git (stopgap_workspace_colocate)
+    if Command::new("git").arg("--version").status().is_err() {
+        eprintln!("Skipping because git command might fail to run");
+        return;
+    }
+
+    let test_env = TestEnvironment::default();
+    let repo_path = test_env.env_root().join("repo");
+    let second_path = test_env.env_root().join("second");
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "--colocate", "repo"]);
+    std::fs::write(repo_path.join("file"), "contents").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "initial commit"]);
+    let (initial_commit, _) = test_env.jj_cmd_ok(
+        &repo_path,
+        &["log", "--no-graph", "-T", "commit_id", "-r", "@-"],
+    );
+    // TODO: replace with workspace add, when it can create worktrees
+    stopgap_workspace_colocate(&test_env, &repo_path, true, "../second", &initial_commit);
+
+    // Break one of our worktrees
+    std::fs::write(second_path.join(".git"), "invalid").unwrap();
+
+    let (_, stderr) = test_env.jj_cmd_ok(&second_path, &["status"]);
+    insta::assert_snapshot!(stderr, @r"
+    Warning: Workspace is a broken Git worktree
+    The .git file points at: invalid
+    Hint: If this is meant to be a colocated JJ workspace, you may like to try `git -C $TEST_ENV/repo worktree repair`
+    ");
+}
+
+#[test]
+fn test_colocated_workspace_independent_heads() {
+    // TODO: Remove when this stops requiring git (stopgap_workspace_colocate)
+    if Command::new("git").arg("--version").status().is_err() {
+        eprintln!("Skipping because git command might fail to run");
+        return;
+    }
+
+    let test_env = TestEnvironment::default();
+    let repo_path = test_env.env_root().join("repo");
+    let second_path = test_env.env_root().join("second");
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "--colocate", "repo"]);
+    // create a commit so that git can have a HEAD
+    std::fs::write(repo_path.join("file"), "contents").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "initial commit"]);
+    let (initial_commit, _) = test_env.jj_cmd_ok(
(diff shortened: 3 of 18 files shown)


lib/src/git.rs
```diff
@@ -546,13 +546,15 @@ fn remotely_pinned_commit_ids(view: &View) -> Vec<CommitId> {
 ///
 /// Unlike `reset_head()`, this function doesn't move the working-copy commit to
 /// the child of the new `HEAD@git` revision.
-pub fn import_head(mut_repo: &mut MutableRepo) -> Result<(), GitImportError> {
+pub fn import_head(
+    mut_repo: &mut MutableRepo,
+    git_worktree: &gix::Repository,
+) -> Result<(), GitImportError> {
```
yuja, line 969: @martinvonz, Is it acceptable to teach `RepoLoader` about `workspace_root: Option<&Path>`? I know it's layering violation, but will simplify the mess of backend Git repo and workspace's Git repo.

The idea is roughly as follows:
```
RepoLoader::init_from_file_system(user_settings, &self.repo_path, Some(self.workspace_root()), store_factories)
  store_factories.load_backend(user_settings, &repo_path.join("store"), maybe_workspace_root)
    GitBackend::load(settings, store_path, maybe_workspace_root)
      # check if the workspace_root points to the Git repo of the store.
      # if it does, load it from there
```

yuja (commented):
Reviewed the first half. Thanks for splitting the PRs, but I think this is still big. I would extract patches after "git: Implement git_worktree_add" to new PR.

cli/tests/test_git_colocated.rs
```diff
@@ -874,3 +874,72 @@ fn get_bookmark_output(test_env: &TestEnvironment, repo_path: &Path) -> String {
...
+        let first_git = git2::Repository::open(&repo_path).unwrap();
+        assert!(first_git.head_detached().unwrap());
+
+        let second_git = git2::Repository::open(&second_path).unwrap();
+        assert!(second_git.head_detached().unwrap());
+        let second_head = second_git.head().unwrap();
+
+        let commit = second_head.peel_to_commit().unwrap().id();
+        assert_eq!(commit.to_string(), initial_commit);
+    }
```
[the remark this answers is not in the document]
cormacrelf: Sure, if that's the style. I usually add a failing test separately first, so I can verify later that the "fix" commit actually does something differently. It's useful when you're reordering commits a lot :) But I can collapse them for you.

cli/src/git_util.rs
```diff
@@ -55,18 +55,83 @@ pub fn is_colocated_git_workspace(workspace: &Workspace, repo: &ReadonlyRepo) ->
...
         return false;
     };
-    let Some(git_workdir) = git_backend.git_workdir() else {
-        return false; // Bare repository
-    };
-    if git_workdir == workspace.workspace_root() {
+
+    // The git backend may be a bare repository if we created it without --colocate
+    // but then created a workspace using `jj workspace add --colocate ../second`
+    let git_backend_workdir = git_backend.git_workdir();
```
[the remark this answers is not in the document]
cormacrelf: Also if it's only allowed when the backing repo is colocated, then assuming #4624 is implemented, there would be no point having a `workspace add --colocate` flag at all. You would just want workspaces to be colocated if and only if the backing repo is. And then when you `jj git colocate` a whole repo, it would add worktrees for all the workspaces, and hypothetically `jj git un-colocate` (?) would remove them too. If you did go down that route, we would still need a CLI flag in this PR, because the feature is not ready to be turned on all the time.

The way this function works now, we can choose any of these UI options later. So I think we should keep it as-is.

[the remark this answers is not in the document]
cormacrelf: That reminds me, we need to have a test for creating worktrees where workspace ids have slashes. Browsing https://git-scm.com/docs/git-worktree#_details it looks like git doesn't support slashes in worktree ids:

> Each linked worktree has a private sub-directory in the repository’s $GIT_DIR/worktrees directory. The private sub-directory’s name is usually the base name of the linked worktree’s path, possibly appended with a number to make it unique

So if your workspace is called `hotfix/thingo`, we need to mangle the slashes I think.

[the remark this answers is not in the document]
yuja: > Until we add https://github.com/martinvonz/jj/issues/4624, being able to quickly use git (in a temporary workspace) sounds pretty convenient. I generally colocate everything myself though.

Understood, but we should at least ensure that the repo is not bare? I might be wrong, but `git config --unset core.bare` was needed in order to fix weird `HEAD` behavior, and jj's internal Git repo should be a bare repo.

> Also if it's only allowed when the backing repo is colocated, then assuming https://github.com/martinvonz/jj/issues/4624 is implemented, there would be no point having a workspace add --colocate flag at all.

I assumed we want to create a second colocated workspace. If you just want to convert a non-colocated repo to colocated one, #4624 is indeed the way to go, and is probably easier to implement.

cli/tests/test_git_colocated.rs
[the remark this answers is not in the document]
yuja: Just for reference, "tests and documentation in the same commit as the code they test and document."
https://martinvonz.github.io/jj/latest/contributing/#code-reviews

cli/src/git_util.rs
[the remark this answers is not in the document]
cormacrelf: > we should at least ensure the repo is not bare

Some facts
- `jj git init` leaves `core.bare = true`
- `jj git init --colocate` leaves `core.bare = false`
- [core.bare](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corebare) basically disables things that require a working directory. Like `git add`.

- [From here](https://git-scm.com/docs/git-worktree#_configuration_file):
   > By default, the repository config file is shared across all worktrees. If the config variables core.bare or core.worktree are present in the common config file and extensions.worktreeConfig is disabled, then they will be applied to the main worktree only.

Therefore worktrees are never actually bare. So this is not a problem.

You can have a worktree that points to a bare repo and it works just fine. Indeed, you could even implement `jj git init --colocate` using only worktrees, keeping the git repo in `.jj/repo/store/git`, leaving core.bare=true as it doesn't matter, and writing `gitdir: .jj/repo/store/git/worktrees/default` into a workspace root .git file. So after landing the `git_worktree_add` code, you can implement #4624 very easily.

I'm happy to add a test showing that `jj workspace add --colocate` in a non-colocated repo works perfectly, because it does.

> I assumed we want to create a second colocated workspace. If you just want to convert a non-colocated repo to colocated one, https://github.com/martinvonz/jj/issues/4624 is indeed the way to go, and is probably easier to implement.

Of course I want secondary colocated workspaces! I'm describing an extension of #4624 to colocate all the existing secondary workspaces, using git worktrees, as well as the main repo itself. So you can only be in two states -- all workspaces colocated, or all workspaces un-colocated.

lib/src/git.rs
```diff
@@ -348,6 +349,192 @@ pub fn git_worktree_add(
...
+/// when the last path segment of the worktree location matches the worktree
+/// name.
+pub fn git_worktree_remove(stat: WorktreeStat) -> Result<(), WorktreeRemovalValidationError> {
+    let WorktreeStat {
+        working_directory_state,
+        worktree_data,
+    } = stat;
+    fs::remove_dir_all(&worktree_data).map_err(WorktreeRemovalValidationError::RemoveData)?;
+    let worktrees = worktree_data.parent().unwrap();
+    let other_dentries = worktrees
```
PhilipMetzger, line 615: Nit: Other directories is probably a better name, than dentry.

lib/src/git.rs
```diff
@@ -268,6 +268,98 @@ impl CreateWorktreeError {
...
+                Component::Normal(s) => buf.push_str(&s.to_string_lossy()),
+            }
+            buf.push('/');
+        }
+
+        if !windows::ends_with_main_sep(path) && buf != "/" && buf.ends_with('/') {
+            buf.pop(); // Pop last '/'
+        }
+
+        Cow::Owned(buf)
```
PhilipMetzger, line 359: @martinvonz Does this need to be a dep for Google, since its under a separate license?

PhilipMetzger (commented):
I left some minor stuff as Yuya already has begun thoroughly reviewing.

lib/src/git.rs
martinvonz: Thanks for asking. No, but `lib/LICENSE` should be updated to say that the MIT license applies to this file. See https://github.com/libgit2/libgit2/pull/6413 and https://github.com/mitsuhiko/sha1-smol/pull/47 for reference.

cormacrelf: cc @ilyagr ref https://github.com/martinvonz/jj/pull/4188 re path-slash

lib/src/git.rs
martinvonz: I would prefer not. I recently removed `RepoLoader::repo_path` because paths don't make sense when `RepoLoader` is used on a server (4fb7dba73df0bb0241a51d35f04d66f2ae450be6). I haven't started reviewing this PR yet so I don't know if there's another good solution.

cli/src/git_util.rs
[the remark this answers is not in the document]
yuja: Thanks for the detail. I didn't know it's valid to create worktrees for bare repo.

It's odd that internal `.jj/repo/store/git` is exposed to user as a plain Git repo, but I don't have a good reason to prevent that still supporting other bare-repo setup (created by `jj git init --git-repo`.)

> I'm happy to add a test showing that `jj workspace add --colocate` in a non-colocated repo works perfectly, because it does.

That will help. Git worktree handling is often source of subtle bugs.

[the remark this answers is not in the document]
cormacrelf: Lots of new tests added in `Detect... ` commit.

cli/src/git_util.rs
```diff
@@ -55,18 +55,83 @@ pub fn is_colocated_git_workspace(workspace: &Workspace, repo: &ReadonlyRepo) ->
...
+    if worktree_common_dir != backend_repo_path {
+        tracing::warn!(
+            "This workspace has a .git directory that isn't managed by JJ; it points to a Git \
+             repo at {}. You may wish to try `git worktree repair` if you have moved the repo or \
+             worktree around.",
+            worktree_common_dir.display()
+        );
+        return false;
+    }
+    true
```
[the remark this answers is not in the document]
cormacrelf: Lots of new tests added in `Detect... ` commit.  Covered most of the cases including when the main repo moves on disk (which JJ doesn't even cope with yet).

cli/src/cli_util.rs
```diff
@@ -1074,13 +1074,39 @@ impl WorkspaceCommandHelper {
         self.working_copy_shared_with_git
     }
 
-    pub fn open_colocated_git_worktree(&self) -> Result<Option<git2::Repository>, git2::Error> {
+    pub fn open_colocated_git_worktree_git2(
```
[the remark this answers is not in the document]
cormacrelf: Split into one commit to add the helper functions and one to use them in many places.

lib/src/git.rs
cormacrelf: Fixed but no longer part of this batch of changes

cli/src/git_util.rs
```diff
@@ -55,18 +55,83 @@ pub fn is_colocated_git_workspace(workspace: &Workspace, repo: &ReadonlyRepo) ->
...
+    // /repo/.git/worktrees/second should have the git backend's .git directory as a
+    // prefix. Check -- the .git file could somehow be pointing elsewhere, or be
+    // its own .git directory
+    if worktree_common_dir != backend_repo_path {
+        tracing::warn!(
+            "This workspace has a .git directory that isn't managed by JJ; it points to a Git \
+             repo at {}. You may wish to try `git worktree repair` if you have moved the repo or \
+             worktree around.",
+            worktree_common_dir.display()
+        );
```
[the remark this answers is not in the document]
cormacrelf: Done and tested but possibly a bit messy

lib/src/git.rs
cormacrelf: Moved to lib/LICENSE, but this will be in a followup PR.

lib/src/git.rs
yuja: @martinvonz gentle ping. I expect the decision will affect fundamental design.

@cormacrelf do you like the idea of loading (or reinitializing) GitBackend for the workspace/worktree path (instead of working around the problem at CLI layer)?

cormacrelf: I think it would be good to have it in `jj_lib`. Currently it is CLI-only so other lib-only projects are not going to be able to use it. If `libgit2` supports worktrees, then so should `jj_lib` I think.

 I don't have an opinion on which is better, but it seems like two options are:

-  reinitialize in someplace like `DefaultWorkspaceLoader::load`
-  or have it pass down a `workspace_root: Option<&Path>` into the repo loader.

cormacrelf: I tried implementing this, easy until you get to `workspace add`. Even if you pass `workspace_root: Option<&Path>` down through RepoLoader, the code for adding a workspace needs to reinitialise an existing repo with a different workspace root. The backend is stored behind multiple Arc references and quite difficult to modify.

martinvonz: > @martinvonz gentle ping. I expect the decision will affect fundamental design.

Sorry, I had not understood that you were expecting my input. I don't know what the issue here is, so I'll try to review this PR now.

cli/tests/test_git_colocated.rs
martinvonz, line 1: > To create a colocated workspace in an existing JJ repo, 
we need to add a corresponding git _worktree_.

Do you mean if you have a regular non-colocated jj repo that uses the Git backend? So there's a `.jj/repo/store/git/` directory, which is a bare Git repo, and we create a `.git/` directory for the new worktree? That sounds reasonable as an alternative solution to #4624.

lib/src/git.rs
yuja: > I tried implementing this, easy until you get to `workspace add`. Even if you pass `workspace_root: Option<&Path>` down through RepoLoader, the code for adding a workspace needs to reinitialise an existing GitBackend with a different workspace root.

Good point. Maybe we can work around it by reloading workspace/repo pair, but it's yeah a bit ugly.

> > @martinvonz gentle ping. I expect the decision will affect fundamental design.
> 
> Sorry, I had not understood that you were expecting my input. I don't know what the issue here is, so I'll try to review this PR now.

It's basically whether we accept this PR (adding workarounds to CLI) or not (to GitBackend and its loaders.) For example, https://github.com/martinvonz/jj/pull/4644/commits/70dae549c85451c31d9f8a774c01636939f9191c wouldn't be needed if `GitBackend::git_workdir()` points to the jj's workspace root.

yuja: Another idea:
* remove `GitBackend::git_workdir()`, etc.
* rename `GitBackend::git_repo()` to `::git_store_repo()`? (ideally it should be detached from the workdir)
* add `GitBackend::git_repo(workspace_root: &Path) -> Option_or_Result<_>`, etc. which check if the workdir points to the same repository, load+cache or reuse repository instance (which shares in-memory cache)

cli/tests/test_git_colocated.rs
cormacrelf: Yes! I described the layout to colocate the default workspace this way here: https://github.com/martinvonz/jj/pull/4644#discussion_r1801370345

The code to make the worktree is in https://github.com/martinvonz/jj/pull/4678 and it should not be difficult to implement 4624 on top.

Only downside is, if you are using it for the default workspace that houses the JJ repo, Git does not like relative (../../ etc) paths in the `.jj/repo/store/git/worktrees/xxx/gitdir` file. If you put a relative path in there, it treats it relative to PWD, which is silly if you ask me. It doesn't break everything, it just reports the worktree as missing and therefore prunable, and git will eventually remove the worktree. So you have to lock the worktree, which prevents pruning, and may want to have JJ automatically update the file to be an absolute path. But we don't have a great story on being able to move workspaces and main repos around without breaking things, in general.

martinvonz (commented):
Sorry that I'm so slow to review this PR. I haven't finished reviewing it but what I've seen looks good to me.

lib/src/git.rs
yuja: The problem is that we have to be careful to not use `git_backend.git_repo()` when `HEAD` matters. IIRC, this patch re-instantiates git2/gix repo by CLI, and passes it to `git::import_head()`/`reset_head()`/`export_refs()` because `git_backend.git_repo()` points to the worktree of the main/default workspace.

If `GitBackend::load()` could set up the internal `gix::ThreadSafeRepository` properly, all problems would go away.

cli/tests/test_git_colocated.rs
cormacrelf: It's definitely used for:

- `git worktree list` etc
- pruning prunable worktrees during `git gc`, tuned by `gc.worktreePruneExpire`

I ran a system call trace on a few other git commands:

- `git status` in the worktree  -> doesn't
- `git log` in the worktree -> doesn't
- `git log --all` in the worktree **does look at gitdir**.
    - But it was traced in the sequence `refs/heads`, `refs/heads/master`, `worktrees`, `worktrees/two/gitdir`, `worktrees/git/commondir`.
    - I put a path like `/tmp/__invalid__` in the gitdir, and it did not appear in the trace. So git didn't actually use the contents. Maybe it just called stat() on it. (I'm running ktrace on a Mac.)
    - The behaviour of `git log --all` requires enumerating all the `refs/heads`. So I think it was just scanning the worktrees for other commondirs to discover more refs/heads directories. Or something like that.

lib/src/git.rs
cormacrelf: I sketched out what having GitBackend open the right repo would look like in https://github.com/martinvonz/jj/pull/4801. The code supporting `jj workspace add`'s reinitialization needs a better idea than using the arc-swap crate, which I just whacked in there to make it go. But it is feasible.

yuja: Thanks. I assume `is_colocated_git_workspace()` and `GitBackend` constructors will become much simpler if we remove colocation warning and refactor the code. I think it's okay to ignore non-colocated `.git` directory.

For `jj workspace add`, I would just make it reload the entire repo at the new workspace.

cormacrelf: Reloading RepoLoader from disk was easy. Neat!

I refactored the code but kept the colocation warnings by stuffing them in an enum. The GitBackend code ended up not getting messy at all.

yuja: Thanks. I'll take a look tomorrow. btw, warning could be moved to CLI if that's important. Something like, if workspace_root + "/.git" exists, and if not git_repo.is_colocated(), emit warning.

lib/src/workspace.rs
```diff
@@ -366,20 +381,37 @@ impl Workspace {
...
+        // Now load a repo at the current operation. We go through
+        // OperationId as the Operation struct contains a reference
+        // to the old workspace's Store.
+        let current_op = repo_loader
+            .load_operation(existing_repo.operation().id())
+            .map_err(WorkspaceInitExistingError::LoadOperation)?;
+        let repo = repo_loader
+            .load_at(&current_op)
+            .map_err(WorkspaceInitExistingError::LoadRepo)?;
+
```
yuja, line 405: I meant this could be worked around by the caller by recreating the `WorkspaceCommandHelper` after minimal workspace setup has been done. Can you check if that's doable?

lib/src/git.rs
```diff
@@ -1812,3 +1813,145 @@ pub fn parse_gitmodules(
...
+}
+
+impl GitRepoColocationType {
+    pub(crate) const fn internal() -> Self {
+        Self::NotColocated { warning: None }
+    }
+    pub fn is_colocated(&self) -> bool {
+        matches!(self, Self::Colocated)
+    }
+    pub fn warning(&self) -> Option<&ColocatedWorkspaceWarning> {
```
yuja, line 1881: nit: as I said in other thread, I think warning stuff can be moved to CLI. The `GitBackend` doesn't have to know that there was a `.git` directory that looked like a colocated repo but it wasn't.

lib/src/git.rs
```diff
@@ -1951,7 +1978,78 @@ impl MaybeColocatedGitRepo {
...
+        // But for a JJ workspace of a colocated repo, this will be
+        //     /repo/.git/worktrees/second
+        // ... or, if the JJ repo was not originally colocated:
+        //     /repo/.jj/repo/store/git/worktrees/second
+        //
+        // ... and the regular file /second/.git will direct git to that location.
+        //
+        // So try to open the workspace root (/second) as a git repository.
+        let worktree_repo =
+            match gix::ThreadSafeRepository::open_opts(workspace_root.join(".git"), open_opts) {
```
yuja, line 1997: Can we use the main repo to iterate worktrees?
https://docs.rs/gix/latest/gix/struct.Repository.html#method.worktrees

lib/src/git.rs
yuja: I generally like the new patches, many thanks.

@martinvonz, are you okay with this change?
https://github.com/martinvonz/jj/pull/4644/commits/8b586b0c54ac5bd84ff5e9f5e2871252d457e27d

lib/src/workspace.rs
cormacrelf: This code is the minimal workspace setup being done, and the caller goes on to recreate the WorkspaceCommandHelper using the Workspace & repo returned from this function.

You need to reload the store *before* you call `init_working_copy` so that the store returned in the repo from this function is the same as the rest of the places, and then the changes to the repo can be written out (i.e. the tx.finish at the end of workspace/add.rs can call `git::reset_head()`). If you instead ignore the return values of this function and reload the workspace from disk, the `independent_heads` tests fails.

And you can't do a full WorkspaceCommandHelper load from scratch if you haven't called `init_working_copy`, because in that case there is no working copy to load. So this is the only way.

lib/src/git.rs
cormacrelf: Yes, but it comes with disadvantages. You would have to:
 
1. Either
      - Plumb WorkspaceId down through the store factories as well as the workspace root (we would need both, as we already need the workspace root to detect non-worktree colocated repos)
      - Decide on a naming convention for JJ-managed worktree identifiers, which we should be loath to change later as it will break people's repos
      - Canonicalize and compare with workspace root anyway, to detect if someone has messed with the worktrees or their names
2. Or
      - Read the worktree gitdir file for O(N) worktrees. Pretty sure there will be people out there with dozens, who make one per branch they ever work on and never clean up. Or people who keep worktrees on network drives.
      - Canonicalize and compare with workspace root anyway, as part of the filtering

The way it is now is as simple and fast as it gets, really, modulo the warnings. We already have the benefits of gix checking all the worktree stuff, so iterating worktrees would not be any more correct. As I noted in another thread, git does not iterate its worktrees in normal operation, probably for similar reasons.

If you really want me to do it, I will.

lib/src/workspace.rs
yuja: I meant something like
```rust
let (new_workspace, ..) = Workspace::init_workspace_with_existing_repo(..);
let loader_factory = ..; // somehow get workspace loader factory
let new_workspace = loader_factory.create(new_workspace.workspace_root())?.load(..);
```

But yeah, it looks uglier than I thought. It's better to add a bit of mess to the library code.

lib/src/git.rs
yuja: > Canonicalize and compare with workspace root anyway, to detect if someone has messed with the worktrees or their names

I meant this. My feeling is that it's slightly better than instantiating an arbitrary repo at `<workspace_root>/.git` (which might be unrelated), but I'm not sure if that makes the code shorter.

martinvonz:
> @martinvonz, are you okay with this change?
https://github.com/martinvonz/jj/commit/8b586b0c54ac5bd84ff5e9f5e2871252d457e27d

Yes, I don't have a better suggestion at least, and I don't want to block these PRs any more than I already have (sorry that it took me two days to even notice your question). Thank you both! I'm happy we're getting this feature soon.

lib/src/git.rs
cormacrelf: I actually think that, with the strategy of just opening `workspace root/.git`, we can probably just delete most of the previous colocation checking code that I ported over to MaybeColocatedGitRepo. I just wanted to keep these changes clean and reviewable, so I didn't touch it. But I actually think if you delete everything above the gix open call, it will all still work.

lib/src/git.rs
```diff
@@ -1863,46 +1863,45 @@ impl MaybeColocatedGitRepo {
...
+            return self.with_colocated(false);
+        };
 
-        // Colocated workspace should have ".git" directory, file, or symlink. Compare
-        // its parent as the git_workdir might be resolved from the real ".git" path.
-        let workspace_dot_git = workspace_root.join(".git");
-        let Ok(workspace_dot_git_canonical) = workspace_dot_git.canonicalize() else {
+        // Especially for worktrees, common_dir() returns paths with ../.. in them,
+        // usually. Must canonicalize.
+        let Ok(workspace_common_dir) = workspace_repo.to_thread_local().common_dir().canonicalize()
```
yuja, line 1890: Maybe we can use `ThreadSafeRepository::common_dir`?
https://docs.rs/gix/latest/gix/struct.ThreadSafeRepository.html#structfield.common_dir

It's `Option<_>`, so I think we can return early if the repo isn't a worktree.

cli/tests/test_git_colocated.rs
```diff
@@ -975,3 +975,354 @@ fn stopgap_workspace_colocate(
...
+        );
+
+        // Now we import the new HEAD in the commit_in workspace, so it's up to date.
+        let (check_git_head, stderr) = test_env.jj_cmd_ok(
+            commit_in,
+            &["log", "--no-graph", "-r", "git_head()", "-T", "commit_id"],
+        );
+        assert_eq!(
+            stderr,
+            "Reset the working copy parent to the new Git HEAD.\n",
```
yuja, line 1316: nit: better to use `insta::allow_duplicates!`? It's tedious to update the expectation on wording changes.

cli/tests/test_git_colocated.rs
```diff
@@ -975,3 +975,354 @@ fn stopgap_workspace_colocate(
...
+    // Write in default, expect second not to import it
+    let new_commit = test_independent_import(&test_env, &repo_path, &second_path, &followup_commit);
+    // Write in second, expect default not to import it
+    test_independent_import(&test_env, &second_path, &repo_path, &new_commit);
+
+    fn test_independent_import(
+        test_env: &TestEnvironment,
+        commit_in: &Path,
+        no_import_in_workspace: &Path,
+        workspace_at: &str,
```
yuja, line 1265: nit: `workspace_at` sounds like a path. Rename to `wc_commit` or something?

cli/tests/test_git_init.rs
```diff
@@ -731,6 +731,237 @@ fn test_git_init_external_but_git_dir_exists() {
...
+
+    assert_eq!(
+        PathBuf::from(read_git_target(&workspace_root))
+            .canonicalize()
+            .unwrap(),
+        worktree_path.join(".git")
+    );
+
+    // This is similar to a normal `jj git init --git-repo=` -- we import the
+    // commits, but in this case our HEAD@git comes from the worktree.
```
yuja, line 793: nit: `HEAD@git` -> `Git HEAD`

cli/tests/test_git_init.rs
```diff
@@ -731,6 +731,237 @@ fn test_git_init_external_but_git_dir_exists() {
...
+    │  args: jj git init --git-repo .
+    ○  eac759b9ab75 add workspace 'default'
+    ○  000000000000
+    "#);
+}
+
+const OP_LOG_COMPACT: &[&str] = &[
+    "op",
+    "log",
+    "-Tself.id().short() ++ ' ' ++ separate(\"\\n\", self.description().first_line(), self.tags())",
```
yuja, line 885: nit: `r#".."` to simplify escape

cli/tests/test_git_init.rs
```diff
@@ -731,6 +731,237 @@ fn test_git_init_external_but_git_dir_exists() {
...
+];
+
+/// This one is a bit weird, but technically you can do it. Should be roughly
+/// equivalent to the --git-repo=. case, but with a different git_target file.
```
yuja, line 889: Indeed. It might be eligible for warning, but we don't have to within this PR.

(No action needed)

yuja (approved):
Thanks for these works.

lib/src/git.rs
cormacrelf: The implementation of common_dir() on Repository is

```
    /// Returns the main git repository if this is a repository on a linked work-tree, or the `git_dir` itself.
    pub fn common_dir(&self) -> &std::path::Path {
        self.common_dir.as_deref().unwrap_or_else(|| self.git_dir())
    }
```

So really gix should just have a copy of this function on ThreadLocalRepository. I will inline this.

> It's Option<_>, so I think we can return early if the repo isn't a worktree.

Not now that we are using this to check normal .git dirs as well. But the function above will work.

lib/src/git.rs
```diff
@@ -1863,46 +1863,45 @@ impl MaybeColocatedGitRepo {
...
-        // 1. Check if we are in a colocated workspace, specifically the one that has
-        //    both .git and .jj/repo.
-        // --------------------------------------------------------------------------
-
-        // Fast path -- the paths are the same, without looking through symlinks.
-        if git_backend_workdir == Some(workspace_root) {
-            // We are in a colocated workspace that's home to the real .git directory.
-            // e.g. /repo with /repo/.git
-            return self.with_colocated(true);
-        }
```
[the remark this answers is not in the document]
cormacrelf: I don't even know if the fast path was ever being activated. I'll try with some debug logs.

cli/src/git_util.rs
```diff
@@ -73,12 +73,72 @@ pub fn get_git_repo(store: &Store) -> Result<git2::Repository, CommandError> {
...
+                target.display()
+            )
+            .ok();
+        } else if meta.is_dir() {
+            writeln!(
+                ui.warning_default(),
+                "Workspace has a .git directory that is not managed by JJ"
+            )
+            .ok();
+        } else if let Ok(_worktree_repo) = gix::open(workspace.workspace_root()) {
```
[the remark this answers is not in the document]
cormacrelf: True, I had it like this because there are more details you can add to the warning if you don't have to make the tests work on windows with its myriad path formats, and I had to remove them.

lib/src/git.rs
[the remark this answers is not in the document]
cormacrelf: Right, but these paths aren't necessarily canonicalized. It might well be doing absolutely nothing in almost every case.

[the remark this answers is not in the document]
yuja: I meant the fast path where we don't have to construct new `gix::ThreadSafeRepository`, with/without path canonicalization.

If it failed (and maybe if the store repo has common_dir?), we'll need to instantiate new repo at `<workspace_root>.git`.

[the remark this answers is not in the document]
cormacrelf: It's defeated by `cd subdirectory; jj -R ..`. But most of the time it works. Worth keeping.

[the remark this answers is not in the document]
yuja: > It's defeated by `cd subdirectory; jj -R ..`.

Oh, really? I have no idea how that was working (or broken) before this PR.

fwiw, it might be worth adding some `tracing::trace!` or `::debug!` there.

PhilipMetzger (approved):
LG, left some last minor nits.

cli/tests/test_git_colocated.rs
```diff
@@ -975,3 +975,354 @@ fn stopgap_workspace_colocate(
...
+    ○  046d74c8ab0a4730e58488508a5398b7a91e54a2 git_head() initial commit
+    ◆  0000000000000000000000000000000000000000
+    "#);
+}
+
+#[test]
+fn test_colocated_workspace_wrong_gitdir() {
+    // TODO: Remove when this stops requiring git (stopgap_workspace_colocate)
+    if Command::new("git").arg("--version").status().is_err() {
+        eprintln!("Skipping because git command might fail to run");
```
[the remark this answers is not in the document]
cormacrelf: I am just copying what's there already, so someone can grep for all uses.

cli/src/git_util.rs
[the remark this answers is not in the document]
yuja: If it can be verbose, I would remove the warning at all. Maybe the warning can be inserted to initialization/debug commands?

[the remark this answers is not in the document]
PhilipMetzger: > Want me to add a flag in this PR? `git.suppress-colocation-warning` or something?

No, I think landing it as is or with Yuya's suggestion is fine, but it should definitely move behind the auto-colocate workspace flag in the follow-up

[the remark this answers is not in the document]
cormacrelf: 1. We currently hard error if you try to `jj git init` in an existing Git repo. To accomplish that you have to manually move files around. `git init` in an existing JJ repo is what this is most likely to warn about.
2. I think we should have a config flag, **and** I think the flag should also control whether jj reads/writes HEAD and the index after mutating jj/git operations. If you want jj to stop touching the HEAD and the index for a while, because you urgently need to push to production using git and some script using jj keeps messing with everything (e.g. your prompt uses jj but doesn't ignore wc), then I think you should be able to do that by temporarily disabling the flag.
3. I would call it `git.sync-with-colocated` or similar. I think that flag should be separate from `git.auto-colocate` because of the use case I just described.
4. It's also very simple to implement, all the HEAD mutation stuff is already contingent on colocation being discovered, so just don't discover.
    - I will move MaybeColocatedGitRepo to git_backend.rs because that's indeed where it will need to end up.
6. However I would probably want to disable these warnings specifically for .git *directories* / *symlinks* for now, just in case this PR makes it to 0.24 and we're still debating the name. Worktree warnings can stay because nobody is doing that yet, and I have written so many tests for it anyway.

cli/src/git_util.rs
```diff
@@ -73,12 +73,72 @@ pub fn get_git_repo(store: &Store) -> Result<git2::Repository, CommandError> {
...
+        if meta.is_symlink() {
+            let target = std::fs::read_link(&workspace_dot_git)
+                .unwrap_or_else(|_| Path::new("<could not read link>").to_path_buf());
+            writeln!(
+                ui.warning_default(),
+                "Workspace has a .git symlink (pointing to {}), that isn't pointing to JJ's git \
+                 repo",
+                target.display()
+            )
+            .ok();
```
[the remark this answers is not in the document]
cormacrelf: What would you do instead? If we can't write a warning to stderr, we aren't going to be able to print a panic message either. So I figured there was no point panicking. All other instances of `writeln!(ui.xxx())` in the codebase either use `?` or if not in a ->Result function `.ok()`, so I followed that.

lib/src/git.rs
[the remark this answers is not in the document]
cormacrelf: You actually fixed this bug with `-R` in #2026. It has been working because the symlink-proof code below it still works if the fast path isn't taken.

I looked into why the mismatch, and at the start of jj execution, we canonicalize CWD but don't canonicalize the join of CWD + the `-R` flag. That's all. I can fix that too.

cli/src/git_util.rs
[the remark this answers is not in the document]
yuja: What I meant was just make this function return `Result<(), CommandError>`. All callers can propagate the error.

lib/src/git.rs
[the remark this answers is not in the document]
yuja: Just to be clear, I called the original `is_colocated_git_workspace()` code a fast path. So long as it can detect colocated workspace, we don't have to instantiate new gix repository.

cli/src/git_util.rs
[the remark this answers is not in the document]
yuja: I don't think we would want a flag to disable import/export in colocated workspace, but that's a separate issue, so let's not discuss here.

Regarding the warning, suppose we have some concerns, I would remove it at least from this PR. I personally don't think we'll need these warnings, and there are no bug report about that iirc.

ejpcmac:
Hello, is there something I could to to help landing this? From reading the comments, it seems it was mostly good except a few nits to fix. Is just time / energy lacking? If yes I can provide some of mine.

Context: I’m just switching to Jujutsu and have an extensive use of Git worktrees and Nix Flakes, so I’m very interested in this feature, even if it’s not 100% polished.

teto:
if there is a rebase I can help testdrive this as well. Like the previous commenter, heavy user of nix flakes and git worktrees.

zzJinux:
I am also a heavy git worktree user and really wish it merged. Are there any blockers left, @cormacrelf ?

yanshay:
> I am also a heavy git worktree user and really wish it merged. Are there any blockers left, @cormacrelf ?

I moved from using worktrees to local clones and so far it's been working reasonably well for my needs.

junkblocker:
> > I am also a heavy git worktree user and really wish it merged. Are there any blockers left, @cormacrelf ?
> 
> I moved from using worktrees to local clones and so far it's been working reasonably well for my needs.

All the non-jj tools, build automation etc. want things to recognized as a git workspace or worktree. Without that it remains a blocker for my jj adoption for work and thus for personal use too.

ejpcmac:
Could we collectively stay focused on the matter, that is:

* either waiting silently for a reply from a maintainer or the author of the PR,
* concrete actions to try moving this forward (rebasing, testing)?

Other comments won’t add to the discussion and just add pressure where there’s no need to—it’s open-source, people are working on this on their spare time and energy.

I’ll myself try to rebase this on top of the latest `master` in July, when I’ll have much more time to spend on open source, if no one has done it before.

cormacrelf:
I stopped working on it because I didn't have the energy to get through review, which is quite exhaustive. I don't mind if someone else rebases and resubmits.

ejpcmac:
Thank you @cormacrelf for all the work already done on this subject :)

I’ll then rebase / test and proceed with the review comments in July, except if someone wants to do it before that.

WaffleLapkin:
@ejpcmac did you end up having time for this? 👀

ejpcmac:
> @ejpcmac did you end up having time for this? 👀

TL;DR: I’m only on it since this afternoon. So far I’ve been able to setup a test repo with colocated workspaces working, with some quirks, using the code from this PR.

## Done / observed so far

I’ve left my full-time job at the end of June to become a freelancer, and I initially thought I’d start working on OSS on the 1st of July. Actually I have taken a huge break and I’m back on Serious Things™ since yesterday.

Today I’ve setup a development environment and test repository to work on this PR. I am now able to create a colocated workspace using the method from the tests (i.e. creating a Git worktree, renaming it, creating a Jujutsu workspace at the same place, moving the `.git` file in and fixing the worktree).

After my first experiments, I’m seeing three things I’d like to fix:

1. when a workspace is not colocated, `jj log` still shows the `git_head()` from the `@default` workspace, which is IMO misleading. `git_head()` is properly updated once the workspace is colocated;
2. when switching back and forth between workspaces, Jujutsu seems to reimport the state from Git, which creates a lot of duplicate changes;
3. this one is certainly related to (1) and (2): if we create a new non-colocated workspace, `git_head()` points to the Git HEAD from the latest visited colocated worktree, which does not give a consistent result.

Also, given:

1. we have a Jujutsu / Git colocated repo,
2. we create a new Git worktree,
3. we run `jj git init --colocate` in this new Git worktree;

the command happily creates a *new* Jujutsu repo without any warning. Maybe we don’t want to fix this now, but I think we should issue a warning at some point to avoid some UX pitfall.

I haven’t looked at the code in details yet, but I’m seeing some TODOs / FIXMEs that are related to these quirks.

## Next steps

Before changing anything, I’ll both look more in details at the code and rebase it on the latest `main`. I bet that I’ll need to open a new PR after the rebase, where we can continue the discussion.

The main question I am anticipating is: should we merge this with these `git_head()` quirks? Should we fix them in a follow-up, or maybe better, prior refactoring PR? I’m up to do any needed work to have a clean, unsurprising UX.

yuja:
> The main question I am anticipating is: should we merge this with these `git_head()` quirks? Should we fix them in a follow-up, or maybe better, prior refactoring PR? I’m up to do any needed work to have a clean, unsurprising UX.

I don't remember the details, but I think the `git_head()` stuff was split from this because this PR is pretty big by itself. It's probably better to  work on `git_head()` handling separately. iirc, this PR can get smaller by removing changes that aren't strictly necessary for the minimal functional implementation, but I'm not pretty sure.

PhilipMetzger:
> > The main question I am anticipating is: should we merge this with these `git_head()` quirks? Should we fix them in a follow-up, or maybe better, prior refactoring PR? I’m up to do any needed work to have a clean, unsurprising UX.
> 
> I don't remember the details, but I think the `git_head()` stuff was split from this because this PR is pretty big by itself.

Yes, that's correct see https://github.com/jj-vcs/jj/pull/4588#issuecomment-2408789687 in the original big PR.

ejpcmac:
Hello all, little status update here.

I’ve tried to rebase the PR during the summer, but it turned out a little bit more difficult than anticipated, notably because the added code uses in some functions a parameter that has been removed. As I had zero knowledge of the code base and could not take any decision on what to do, I’ve been stuck and worked on other subjects.

I’m now taking another approach, by first getting used to the code base as a whole in term of concepts and architecture. As I prefer to think upfront and act after, I’ll also write a plan of the different steps we want to follow to enable the support of colocated workspaces. This will give us a clear view with milestones. I’ve notably seen the recent addition of `jj git colocate` which will likely be useful to integrate.

Would a new tracking issue on top of #4436 be a good place to discuss the plan and design? I like having a centralised place like an issue where I can edit the description to see were I’m going.

I will allocate a few hours per week on Jujutsu until the end of the year at least.

ejpcmac:
I have created a tracking issue for this feature and its associated work: https://github.com/jj-vcs/jj/issues/8052. I will post updates in this tracking issue from now on.

sjawhar:
@cormacrelf sorry for accidentally rebasing this branch. I was working on the plane and wasn't paying very close attention.

martinvonz:
Should this PR be closed if #8667 replaces it? Or maybe #8667 can be closed so we can keep the review here?

calebdw:
Closing, as https://github.com/jj-vcs/jj/pull/9941 introduced colocated workspaces

PR #4647, opened 2024-10-14
formatter & colors.toml: replace `op_log` color label

ilyagr:
Cleans up after 7051effa8f

It's split into "operation log" and "operation summary". The color labels use "operation".

# Checklist

If applicable:
- [?] It doesn't seem to be mentioned in the docs

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1492,12 +1492,7 @@ impl WorkspaceCommandHelper {
             &self.op_summary_template_text,
             OperationTemplateLanguage::wrap_operation,
         )
-        // TODO: "op_log" is not a great label name for op_summary. We could
-        // replace it with `operation summary` and use `operation` in
-        // colors.toml. TODO: It would be better to put labels like these
-        // consistently in the same place, e.g. `templates.toml` instead of the
-        // rust code.
-        .labeled("op_log")
+        .labeled("operation")
     }
 
     pub fn short_change_id_template(&self) -> TemplateRenderer<'_, Commit> {
--- a/cli/src/commands/operation/log.rs
+++ b/cli/src/commands/operation/log.rs
@@ -130,6 +130,7 @@ fn do_op_log(
                 &text,
                 OperationTemplateLanguage::wrap_operation,
             )?
+            .labeled("operation")
             .labeled("op_log");
         op_node_template = workspace_env
             .parse_template(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Help command doesn't work recursively anymore, i.e. `jj workspace help root`
   doesn't work anymore.
 
+* The color label `op_log` from the `[colors]` config section now **only**
+  applies to the op log and not to the other places operations are displayed. In
+  almost all cases, if you configured `op_log` before, you should use the new
+  `operation` label instead.
+
 * Default operation log template now shows end times of operations instead of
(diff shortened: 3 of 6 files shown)


cli/src/cli_util.rs
```diff
@@ -1492,12 +1492,8 @@ impl WorkspaceCommandHelper {
...
             OperationTemplateLanguage::wrap_operation,
         )
-        // TODO: "op_log" is not a great label name for op_summary. We could
-        // replace it with `operation summary` and use `operation` in
-        // colors.toml. TODO: It would be better to put labels like these
-        // consistently in the same place, e.g. `templates.toml` instead of the
-        // rust code.
-        .labeled("op_log")
+        .labeled("summary")
+        .labeled("operation")
```
yuja, line 1496: nit: shouldn't it be just `"operation"` or `"op_summary" "operation"`?

cli/src/commands/operation/log.rs
```diff
@@ -130,7 +130,8 @@ fn do_op_log(
                 &text,
                 OperationTemplateLanguage::wrap_operation,
             )?
-            .labeled("op_log");
+            .labeled("log")
+            .labeled("operation");
```
yuja, line 134: nit: should be `.labeled("operation").labeled("op_log")` (= `["op_log", "operation"]`)? (read: the command is "op_log", and the context object is an "operation".)

If we apply the same rule to `jj log`, the labels will be `["log", "commit"]`.

yuja (approved):

cli/src/cli_util.rs
ilyagr: To the degree that this was a question: I thought that "operation op_summary" is a bit redundant, and "operation summary" is better. Same for "operation op_log". Perhaps one day we'll need to formatting for all summaries, or all logs (log, op_log, evolog).

But the usefulness of that does seem a bit remote. I'm happy to do adjust it to your preference, especially since you have the rest of the formatter code fresher in your mind. The confusion about the organizing principle is part of the reason I wish all these labels were in one place.

yuja: Well, "operation summary" sounds natural as English, but labels represent a document structure. I believe "operation id" makes more sense than ("operation ")"summary id".

cli/src/commands/operation/log.rs
yuja: Are we going to remove other `<command_name>` labels (e.g. "log", "bookmark_list")? If not, I would avoid derailing the existing convention here.

ilyagr: It sounds like you're OK with the new state of this, so I'll merge it.

I agree that it'd good to keep in mind the entire convention and change the whole thing at once, if somebody feels up to it (which I don't at the moment). So, thank you for keeping track of what the current convention is and how often it's used.

I'd also change the convention, ideally, but I'm not going to do it in the short-term. Also, I think it'd be more important to make the labels more easily discoverable, though you did point out some downsides of putting them all in `templates.toml` and I'm not sure what other options are viable. Neither of these is an urgent task.

ilyagr: Actually, I'll wait a bit, since this is not urgent, and since it's possible that you actually meant for me to change something (though I don't think so). I'll merge this in a few hours or tomorrow if I don't hear otherwise.

yuja: Wrong order? should be "op_log" -> "operation". Other than that, it looks good to me, thanks.

Regarding discoverability, maybe we should document the current convention in templates.md, and suggest using `--color=debug`.

PR #4649, opened 2024-10-15
cli: allow (and ignore) `-d` flag for `jj new`

martinvonz:
We were discussing whether `jj backout` and `jj duplicate` should support `-d/-A/-B` just like `jj rebase` does. `jj new` already accepts `-A/-B` but it does not accept `-d`. It does support `-r`, however. It seems like `-d` is a better match for `jj new` since it creates a commit on top. So this patch adds support for that flag too. I now think `-r` sounds misleading for `jj new`, but I left it in for now.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/new.rs
+++ b/cli/src/commands/new.rs
@@ -51,9 +51,10 @@ pub(crate) struct NewArgs {
     /// Parent(s) of the new change
     #[arg(default_value = "@")]
     pub(crate) revisions: Vec<RevisionArg>,
-    /// Ignored (but lets you pass `-r` for consistency with other commands)
-    #[arg(short = 'r', hide = true, action = clap::ArgAction::Count)]
-    unused_revision: u8,
+    /// Ignored (but lets you pass `-d`/`-r` for consistency with other
+    /// commands)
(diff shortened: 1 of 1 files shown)


cli/src/commands/new.rs
```diff
@@ -51,8 +51,9 @@ pub(crate) struct NewArgs {
     /// Parent(s) of the new change
     #[arg(default_value = "@")]
     pub(crate) revisions: Vec<RevisionArg>,
-    /// Ignored (but lets you pass `-r` for consistency with other commands)
-    #[arg(short = 'r', hide = true, action = clap::ArgAction::Count)]
+    /// Ignored (but lets you pass `-d`/`-r` for consistency with other
+    /// commands)
+    #[arg(short = 'd', hide = true, short_alias = 'r',  action = clap::ArgAction::Count)]
     unused_revision: u8,
```
ilyagr, line 57: You could call this `unused_destination`.

ilyagr (approved):
SGTM.

I don't see any harm in keeping `-r`. I don't think it's even documented anywhere, and I can imagine people using it.

cli/src/commands/new.rs
martinvonz: Done

yuja (approved):

PR #4653, opened 2024-10-16
cli: remove author name from default annotate template, reorder fields

yuja:
# Checklist

If applicable:


cli/src/config/templates.toml
```diff
@@ -17,9 +17,8 @@ commit_summary = 'format_commit_summary_with_refs(self, bookmarks)'
 annotate_commit_summary = '''
 separate(" ",
   format_short_id(change_id),
-  format_short_id(commit_id),
-  format_short_signature(author),
   format_timestamp(committer.timestamp()),
+  format_short_id(commit_id),
```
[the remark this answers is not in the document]
yuja: Ok, let's revert the change then.

cli/src/config/templates.toml
```diff
@@ -17,9 +17,8 @@ commit_summary = 'format_commit_summary_with_refs(self, bookmarks)'
 annotate_commit_summary = '''
 separate(" ",
   format_short_id(change_id),
-  format_short_id(commit_id),
-  format_short_signature(author),
   format_timestamp(committer.timestamp()),
```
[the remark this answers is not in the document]
yuja: and to people who use `.shortest(0)`. I assume they'll probably want to customize the annotation template at all.

yuja (commented):
> If this is the easiest fix in the short term, I think it's fine. 

I believe so. I just want to make the default usable.

> How hard would it be to implement padding?

I don't think it's difficult to implement a padding-only function, but we'll need to come up with a nicer syntax.

A padding function would have to support optional min/max widths, padding character, left/right/center alignment. For readability, it's nice if multiple texts can be inlined in a string literal (just like Rust's `format!()`.) This means padding function will take another template mini language.

cli/src/config/templates.toml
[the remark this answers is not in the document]
ilyagr: Yes. Maybe, until padding exists, we should just hard-code everything here and leave a note to use the normal macros once the padding is available?

(This is still optional)

ilyagr (approved):
I'll just approve it then. But let's wait until tomorrow to merge.

cli/src/config/templates.toml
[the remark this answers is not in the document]
yuja: >  we should just hard-code everything here

@allonsy, was that the reason you didn't use `format_*()` aliases in the original patch?

I don't have a strong opinion, but I thought it would be nice to provide some customization points.

cli/tests/test_file_annotate_command.rs
ilyagr, line 1: (Non-blocking and off-topic, just to have a thread on this)

> This means padding function will take another template mini language.

You clearly have something nice in mind, but I don't entirely follow the need for a mini-language. It seems to me at the moment that we could go for a simple padding function first, without a mini-language, even if having one was ultimately better.

At first, I was thinking that the major problem the mini-language would solve is the lack of keyword arguments in our templating language, so that invocations of the padding function are not too cryptic because of its many arguments, but don't we have those already (or is that just the revset language?)?

> For readability, it's nice if multiple texts can be inlined in a string literal (just like Rust's format!().) 

Could this be separate from a padding function?

----------------

Just to have something concrete in mind, here's a possible design. It's already more complicated than it has to be, but I included all the nice-to-have things I'd want from it in the short term. (Replacing it with something better designed later is certainly also an option).

We could have:

```
pad(
  text: Template,
  padding: ProbablyTemplate,
  min_len=0 : int,
  max_len = INT_MAX : int,
  align: "left" | "right" | "center",
  trim_marker = "": ProbablyTemplate
)
```

Here, `ProbablyTemplate` could be a `Template`, but it's also fine if it's an `str` or a `char` for simplicity, or if we had to require it to be a single-width character. (But I think you've already solved most of these problems in the templating language)

`trim_marker` is optional, but I was thinking that we'd require `len(trim_marker) < max_len`, and the end of the trimmed string is replaced with it.

-----

There's also my crazy de-duplication idea, but that can wait (I guess that'd have to take the list of all possible `text`-s and the index in that list, conceptually).

yuja: > Could this be separate from a padding function?

Yes. I just mentioned it because there would be multiple columns if user wants to do padding, and writing `pad()` for each column would be super noisy. Under the hood, each column would be translated to a `pad()` call you illustrated here.
https://github.com/martinvonz/jj/issues/3183

(FWIW, I would call `width` instead of `len` because it should be visual width, not bytes/character length.)

ilyagr: Oh, I see, it'd include a table-formatting language.

> FWIW, I would call width instead of len because it should be visual width, not bytes/character length

Good point.

ilyagr: In my mind, the hard part might be to implement the function trimming a template to a width. Hopefully there's a library for that, though dealing with our formatting labels as you trim would also be annoying.

I don't know what happens if you change the label and then print some Unicode combining symbol, but we'd need to know the width of whatever that ends up being. I guess you must have dealt with that when you were implementing `fill`, since it must already compute a width.

yuja: Labels are sent to separate channel, so it's not difficult to process. My biggest concern (about `pad()` function) is API consistency. Other formatting functions take `content: Template` as the last argument, but `pad()` will have to be `pad(content, optional_args..)`.

ilyagr: I wouldn't mind switching `content` to be the first argument everywhere. It'd be an annoying deprecation, though.

ilyagr: I would make an exception for `separate` and functions with a variable number of `content` arguments, though. I think it's better to have the separator come first (though it wouldn't be the end of the world to me if it came last).

cli/src/config/templates.toml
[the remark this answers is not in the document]
allonsy: I think you give me too much credit :)
I didn't include the format_* because I wasn't as aware of them as I am now when I started work on file annotation

allonsy:
Just to catch myself up, is the purpose of this PR to remove the author because it messes with alignment?
I do like having the author included in some way in the annotate and its an important part of the change metadata. 
However, it does make formatting a bit of a nightmare.
The issue with padding is it requires global context (meaning it needs to know all the input lines in order to format properly). We are also formatting information in a table like way meaning alignment is important for each column. This is a challenge, especially if the user inputs a custom template for file annotating. This is one of the reasons I decided not to deal with this in the original PR and save that for a later day

yuja:
> is the purpose of this PR to remove the author because it messes with alignment?

Yes, that's the main reason. Another reason is that author name can be long enough to make file contents invisible. So even if we include author, I want to truncate it up to 4-8 characters.

BTW, I've implemented padding/truncating function. If we'd like to include author in the default template, it can be something like `pad(format_short_signature(author), min_width=8, max_width=8)`.

allonsy (approved):

allonsy:
> > is the purpose of this PR to remove the author because it messes with alignment?
> 
> Yes, that's the main reason. Another reason is that author name can be long enough to make file contents invisible. So even if we include author, I want to truncate it up to 4-8 characters.
> 
> BTW, I've implemented padding/truncating function. If we'd like to include author in the default template, it can be something like `pad(format_short_signature(author), min_width=8, max_width=8)`.

Nice, I see no reason not to merge this PR as is. I can either add author manually with padding as mentioned or discuss in an issue if people feel like it is default-worthy

yuja:
Do we like to include author name in the default annotation template?
Since padding function is landed, I can rewrite the template to be fully fixed widths.

joyously:
I vote yes for showing the author and if anything needs to be cut, omit the commit ID.

ilyagr:
My personal preference would be to include the email, trimmed to (say) 8-10 chars. We could try the trimmed name, but that will be annoying if we have more than one Alexander or Jean-Jacques one day.

We could also consider various ways to abbreviate names, but probably not in the first version. It'd also be cool to consider de-duplication, as in https://github.com/martinvonz/jj/pull/4653#discussion_r1802259553, but that is also complicated.

yuja:
Ok, updated the template to include hard-coded (change_id, commit_id, author, timestamp). PTAL.

martinvonz:
> I vote yes for showing the author and if anything needs to be cut, omit the commit ID.

+1. What do the rest of you think about removing the commit id?

ilyagr:
I'm torn about the commit id. The change id is a bit problematic if there are multiple commits with the same commit id, or if we blame hidden revisions. I'm not sure how to address this in a blame view.

I think it would furthermore be convenient for me to keep the change id, since I'd guess that if I'm blaming, I'm likely to be using other tools at the same time (GitHub, VS Code) that might be more familiar with commit ids.

That said, I agree that it's the least important of the things we currently show.

cli/src/config/templates.toml
```diff
@@ -16,10 +16,10 @@ commit_summary = 'format_commit_summary_with_refs(self, bookmarks)'
 
 annotate_commit_summary = '''
 separate(" ",
-  format_short_id(change_id),
-  format_short_id(commit_id),
-  format_short_signature(author),
-  format_timestamp(committer.timestamp()),
+  change_id.shortest(8),
+  commit_id.shortest(8),
+  pad_end(8, truncate_end(8, author.username())),
```
ilyagr, line 21: Perhaps `author.email()` would be better (even though I had just the username in my screenshot), in case we have somebody with `me@one-cool-username.com` and another person with `me@another-cool-username`.

If you disagree, I'd use `pad(truncate(author.username() ++ '@'))`, to make it clearer what this field is.

More questionably, after looking at the tests, I'm not sure whether 8 characters is enough, I think we could try 10. We can also keep it as is and change it later, if needed.

cli/src/config/templates.toml
```diff
@@ -16,10 +16,10 @@ commit_summary = 'format_commit_summary_with_refs(self, bookmarks)'
...
+  committer.timestamp().local().format("%Y-%m-%d"),
```
ilyagr, line 22: I would include the hour and minute, in case you are looking at a draft PR stack you recently wrote.

ilyagr:
> Ok, updated the template to include hard-coded (change_id, commit_id, author, timestamp). PTAL.

LGTM. I had some minor suggestions, but we can experiment with them later as well. This is certainly better than what we had before.

cli/src/config/templates.toml
emilazy: This is not solely related to this PR, but I dislike how we have an entire field in commits that represents an appropriate human‐readable identifier to use for their authors and committers but ignore it to just show email addresses instead, and then run into situations like this where we’re hoping that we can use the local‐part as a sensible global identifier. It makes sense in a corporate context where everyone is `‹username›@google.com`, but not for collaborative projects in general.

Of course I’m biased since the local‐parts of my email addresses don’t identify me at all.

cli/src/config/templates.toml
```diff
@@ -16,10 +16,10 @@ commit_summary = 'format_commit_summary_with_refs(self, bookmarks)'
 
 annotate_commit_summary = '''
 separate(" ",
-  format_short_id(change_id),
-  format_short_id(commit_id),
-  format_short_signature(author),
-  format_timestamp(committer.timestamp()),
+  change_id.shortest(8),
```
ilyagr, line 19: This will not show whether the commit is hidden or divergent. 

I'm not sure what to do about it. The simplest things would be to change the color, but it would make the template complicated. Adding `?` for divergent makes sense, but we don't have a short syntax for hidden. 

So, this might not be actionable, but I thought I'd put it out there in case people have ideas.

joyously: Is the goal of `annotate` to give all the gory details, or to show the last change that modified the line?

ilyagr: If the revision is hidden or divergent, it can't be referred to easily by its change id. This seems worth indicating somehow when you show this (potentially useless) change id.

It might be OK to keep things as they are if we also show the commit id. Maybe I'd add color to the change id at some point.

BTW, this keeps reminding me that we might want a syntax like `zzx^a` to refer to a revision with change id zzx (that could be divergent) and commit id starting with `a`. I'm not sure how appropriate this is for hidden commits, but if it is *and* if we decided to usually omit commit ids, it could be a way out. But again, at least for now, I think just showing the commit id might be easier and more convenient (though I'm not certain about it).

martinvonz: I feel like it's going to be very rare to want to annotate a hidden or divergent file. Probably rare enough that the user can override the template when they want to do that. I can't remember ever running `hg blame` or `git blame` on a hidden commit.

cli/src/config/templates.toml
yuja: `username()` (which is sometimes shorter than 8 chars) looked nicer than full 8 or 10 `email()` characters truncated at arbitrary place. I don't have a strong opinion, though.

I wouldn't add `@` suffix unless we do the same for `builtin_log_onenline`.

cli/src/config/templates.toml
yuja: I would add `%H:%M:%S` then. Consistency is more important than saving 3 characters.

ilyagr: I'm not sure. Would you ever want to see the seconds in the blame view? I'd only want that if I did several commits in less than a minute, which is rare.

But I'll leave it up to you and others.

cli/src/config/templates.toml
ilyagr: For reference, here is the log with emails trimmed to 10 chars. I also make the `@` white somewhat indulgently (it'd be easier if we supported `signature.domain()`).

image

cli/src/config/templates.toml
yuja: I personally don't want to see `%H:%M:%S` at all. I wouldn't even be interested in `%d` (and author as I said before.) I'm just saying that we should stick to the default formatting than adding a couple of variants that look quite similar.

cli/src/config/templates.toml
ilyagr: I'm not sure what people other than me prefer. Is a trimmed email a worthwhile compromise between a trimmed username and a full name trimmed to, say, 20 characters?

> I wouldn't add @ suffix unless we do the same for builtin_log_onenline.

I wouldn't mind adding `@` to `builtin_log_onenline` as well, especially if we don't go with the full email here and adding it to the username was conditional on adding it to that template. But again, this might be my personal aesthetics.

ilyagr: I also don't feel very strongly about this, by the way. I can always configure it and, more importantly, we can try something and see how/whether people react.

cli/src/config/templates.toml
martinvonz: It's an interesting question about showing the author. We decided to not call it `jj [file] blame` because we don't want it to seem like a tool for blaming people (at least I think that's why). Removing the author from the output seems compatible with that line of thinking.

OTOH, it does seem a bit useful to help remind you of the context in which a line change. I personally often remember (think I) remember who added a feature, so filtering by user is sometimes helpful at least for finding a commit I'm looking for. But it probably is much less helpful in `jj file annotate` output. I think I just look for the change id (or commit id in git/hg) and date (in addition to the line contents, of course).

So, I think I would personally be happy to try including only change id and date (not even time). But that's just my feeling.

ilyagr:
By the way, there is nothing stopping us from providing a few prefabricated annotate template aliases that the user could swap between, right? (Probably in a separate PR) Or do we not want the user to configure `templates.annotate_commit_summary`, even to choose between a few options?

Also, we probably want to allow something like `jj file annotate -T minimal` or even `jj file annotate -T commit_id` eventually, right? It seems pretty clear that options to show just the commit id and the line number and/or content would be good for scripting. We could also have a git-like template.

The one issue I see is that, ideally, the whole line including the line number and content would be templated, but I don't think it's a huge deal if it isn't.

cli/src/config/templates.toml
ilyagr: (Non-actionable) The actionable part of your comments is pretty clear, but I am a little confused by some details in your comments. I guess I don't follow how you use the annotations?

Yuya, you just want to see the year and the month? So, you only care whether some lines are much older than others? 

Martin, I'm not sure what you are saying in your second paragraph. Where are you "filtering by user" if not in the `jj file annotate` output? The log? I find showing the author useful for precisely the reason you described, for remembering what a commit was about.

Possibly more actionable, but also fanciful and not for this PR: This discussion makes me wonder whether we should include 20-40 characters of the commit description instead of the author and date. I don't know if it would work in practice; I wish there was a way to shorten commit descriptions to X characters. Perhaps this will be `jj`'s first use of ChatGPT 😛 .

martinvonz: > Where are you "filtering by user" if not in the jj file annotate output? The log?

Yes, that's what I meant. Sorry that I wasn't clear.

martinvonz: > BTW, author name is more visually distinct than change/commit ids and dates, so it will help separate lines originating from different commits.

OTOH, that may make it harder to notice when some nearby lines are from different commits if they have the same author.

joyously: If this command has machine output, all the details can be output (JSON ?).

ilyagr: > If this command has machine output, all the details can be output (JSON ?).

No, the output is meant for humans. It could also be used by scripts, but they probably only need the commit id and the line number.

> It could be a TUI opened by e.g. jj file annotate --interactive.

Yes, a TUI would be very nice. Among other nice things, it could have a binding to hide/show author or date, which would make the defaults less important.

I generally use `tig blame` for annotations, which is quite a nice UI, though I'd change a bunch of things about it (easier and more discoverable key bindings, add the show/hide author or date binding, ...).

cli/src/config/templates.toml
ilyagr: Tig supports an abbreviated author name that seems to split the name by spaces, use initials, and then the full last name. Examples from `tig blame Cargo.toml`:

- ASeipp
- MvZweigbergk
- EMesterhazy
- IGrigoriev
- db   (from `dependabot[bot]`)
- SJennings
- YNishihara
- BTan
- JOkoński 
- STardieu

I'm not sure we're lacking in options, but I thought this is worth mentioning.

cli/src/config/templates.toml
yuja: Which one do we like as the default?
(I'm not going to discuss about the truncation width and other minor things, which can be changed later as needed.)

```toml
# A
annotate_commit_summary = '''
separate(" ",
  change_id.shortest(8),
  committer.timestamp().local().format('%Y-%m-%d'),
)
'''

# B
annotate_commit_summary = '''
separate(" ",
  change_id.shortest(8),
  pad_end(8, truncate_end(8, author.username())),
  committer.timestamp().local().format('%Y-%m-%d'),
)
'''

# C (this patch)
annotate_commit_summary = '''
separate(" ",
  change_id.shortest(8),
  commit_id.shortest(8),
  pad_end(8, truncate_end(8, author.username())),
  committer.timestamp().local().format('%Y-%m-%d'),
)
'''

# D
annotate_commit_summary = '''
separate(" ",
  change_id.shortest(8),
  commit_id.shortest(8),
  pad_end(8, truncate_end(8, author.email())),
  committer.timestamp().local().format('%Y-%m-%d %H:%M:%S'),
)
'''
```

I personally like A or B.

cli/src/config/templates.toml
yuja: Discussion continues to https://github.com/martinvonz/jj/pull/4653#discussion_r1816545416, closing as resolved.

cli/src/config/templates.toml
joyously: I like B with the time added.

yuja: Ok, updated to B with '%Y-%m-%d %H:%M:%S'.

If there aren't strong complaints, I'll merge this.

PR #4654, opened 2024-10-16
docs glossary & git-comparison: fixup & update bookmark->branch transition

ilyagr:
This fixes up git-comparison.md. The changes to the glossary could be moved to a separate commit/PR.

### DIFF
--- a/docs/git-comparison.md
+++ b/docs/git-comparison.md
@@ -22,19 +22,21 @@ various use cases.
   to an intermediate commit between `HEAD` and the working copy, so workflows
   that depend on it can be modeled using proper commits instead. Jujutsu has
   excellent support for moving changes between commits. [Details](#the-index).
-* **No need for bookmark names (but they are supported).** Git lets you check out
-  a commit without attaching a bookmark. It calls this state "detached HEAD". This
-  is the normal state in Jujutsu (there's actually no way -- yet, at least -- to
-  have an active bookmark). However, Jujutsu keeps track of all visible heads
+* **No need for branch names (but they are supported as
+  [bookmarks](glossary.md#bookmark)).** Git lets you check out a commit without
+  attaching a branch to it. It calls this state "detached HEAD". This is the
+  normal state in Jujutsu (there's actually no way -- yet, at least -- to have
+  an active branch/bookmark). However, Jujutsu keeps track of all visible heads
   (leaves) of the commit graph, so the commits won't get lost or
   garbage-collected.
-* **No current bookmark.** Git lets you check out a bookmark, making it the 'current
-  bookmark', and new commits will automatically update the bookmark. This is
+* **No current branch.** Git lets you check out a branch, making it the 'current
+  branch', and new commits will automatically update the branch. This is
   necessary in Git because Git might otherwise lose track of the new commits.
-  Jujutsu does not have a 'current bookmark'; instead, you update bookmarks
-  manually. For example, if you start work on top of a commit with a bookmark,
-  new commits are created on top of the bookmark, then you issue a later command
-  to update the bookmark.
+
+    Jujutsu does not have a corresponding concept of a 'current bookmark';
+  instead, you update bookmarks manually. For example, if you start work on top
+  of a commit with a bookmark, new commits are created on top of the bookmark,
+  then you issue a later command to update the bookmark.
 * **Conflicts can be committed.** No commands fail because of merge conflicts.
   The conflicts are instead recorded in commits and you can resolve them later.
   [Details](conflicts.md).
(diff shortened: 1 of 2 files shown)


docs/git-comparison.md
```diff
@@ -110,13 +111,15 @@ parent.
       <td><code>git clone &lt;source&gt; &lt;destination&gt; [--origin &lt;remote name&gt;]</code></td>
     </tr>
     <tr>
-      <td>Update the local repo with all bookmarks from a remote</td>
+      <!-- TODO: Mention that you might need to do a `jj branch track branch@remote` -->
```
emilazy, line 114: I’m not sure where this TODO is from, but it seems outdated.

docs/glossary.md
```diff
@@ -22,13 +22,24 @@ such as the "operation store backend" for storing
...
-A bookmark is a named pointer to a [commit](#commit). They automatically follow
-the commit if it gets [rewritten](#rewrite). Bookmarks are sometimes called
-"named branches" to distinguish them from [anonymous branches](#anonymous-branch).
-They are similar to Git's branches and even more similar to
+A bookmark is a named pointer to a [commit](#commit). They are similar to Git's [branches](#branch) and even more similar to
 [Mercurial's bookmarks](https://wiki.mercurial-scm.org/Bookmarks). See
 [here](bookmarks.md) for details.
 
+Unlike in Git, bookmarks automatically follow the commit if it gets
+[rewritten](#rewrite). However, bookmarks do *not* move when you create a new commit;
```
emilazy, line 30: What are some practical cases where you rewrite a commit in Git and the corresponding branch doesn’t get rewritten? I worry this is a bit misleading as‐is.

emilazy (commented):
Broadly LGTM. We might want to explicitly discuss the bookmark–branch distinction in the comparison, but this is an improvement over the current wrong state.

docs/git-comparison.md
```diff
@@ -44,10 +46,9 @@ various use cases.
   updated, and so will the working copy if it points to any of the rebased
   commits.
 * **Branches are identified by their names (across remotes).** For example, if
```
martinvonz, line 48: maybe "Branches/bookmarks" here

docs/git-comparison.md
```diff
@@ -324,29 +327,29 @@ parent.
...
     <tr>
-      <td>Move a bookmark backward or sideways</td>
+      <td>Move a bookmark/branch backward or sideways</td>
       <td><code>jj bookmark set &lt;name&gt; -r &lt;revision&gt; --allow-backwards</code></td>
-      <td><code>git bookmark -f &lt;name&gt; &lt;revision&gt;</code></td>
+      <td><code>git branch -f &lt;name&gt; &lt;revision&gt;</code></td>
     </tr>
     <tr>
-      <td>Delete a bookmark</td>
+      <td>Delete a bookmar/branchk</td>
```
martinvonz, line 350: fix the typos here

docs/git-comparison.md
ilyagr: I just added it :) It's obviously not clear enough, but honestly it's unclear to me what to do here.

The problem is that if you `jj git fetch` from a remote, you'll get all the bookmarks, but you won't see them in `jj log` until you `jj bookmark track` something. I think new people are likely to be confused by that, but I'm not sure how to help. (I already inserted an explanation in a few places, it's not 100% sure this has to be one of them, but it felt relevant).

ilyagr: I'll update the TODO in a moment.

docs/glossary.md
ilyagr: The most important case (which I think has been addressed to a degree recently) is if you have

```
x -> branch_b -> branch_a
 \-> main
```

and you rebase `branch_a` onto `main`, `branch_b` normally does not move.

----

But I think you have a point, I'll reword this a bit to trigger people's sense of "um, actually" a bit less.

docs/git-comparison.md
emilazy: To be clear the outdated part was just that it’s talking about `jj branch` :)

docs/glossary.md
martinvonz: Good example, but please mention the `--update-refs` flag (which `git rebase` learned about two years ago) so readers don't think we don't know about or that we pretend that it doesn't exist.

docs/git-comparison.md
ilyagr: Ah, great point, thanks!

docs/git-comparison.md
```diff
@@ -243,17 +249,17 @@ parent.
...
       <td><code>jj new main</code></td>
       <td><code>git switch -c topic main</code> or
         <code>git checkout -b topic main</code> (may need to stash or commit
         first)</td>
     </tr>
     <tr>
-      <td>Move bookmark A onto bookmark B</td>
+      <td>Move bookmark/branch A onto bookmark/branch B</td>
       <td><code>jj rebase -b A -d B</code></td>
       <td><code>git rebase B A</code>
```
ilyagr, line 261: @martinvonz: Would it be correct to say `git rebase B A --update-refs` here and remove the comment below? Or would you prefer to mention `--update-refs` in the comment? I'd probably put it into a separate commit.

docs/glossary.md
ilyagr: I just removed "Unlike in Git" from this sentence (and moved it to another sentence).

But see also https://github.com/martinvonz/jj/pull/4654#discussion_r1802344909, which we could also fix.

docs/git-comparison.md
martinvonz: No, I don't think that would be correct because it would still not rebase descendants. If you had a linear chain of bookmarks A..C, then `git rebase D C` would rebase A, B, and C. But it would still fail if it was not linear, so I think it makes sense to leave the comment here.

docs/glossary.md
martinvonz: I think it makes sense to leave it for another commit/PR.

docs/git-comparison.md
ilyagr: OK, I'll leave it be then. For some strange reason, I stopped following new Git features a few years ago :)

PR #4656, opened 2024-10-16
rewrite: move_commits: add `MoveCommitsTarget` enum to specify roots or commits to move

bnjmnt4n:
This also allows some minor optimizations to be performed, such as avoiding recomputation of the connected target set when `MoveCommitsTarget::Roots` is used since the connected target set is identical to the target set (all descendants of the roots).

# Checklist

If applicable:


lib/src/rewrite.rs
```diff
@@ -846,6 +831,35 @@ pub fn move_commits(
     })
 }
 
+/// Computes the parents of all commits in the connected target set, allowing
```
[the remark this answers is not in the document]
bnjmnt4n: Yeah, I guess the naming/description can be a bit ambiguous. This computes what should serve as the parents of the commits if parents not in the target set are removed and replaced with their ancestors. So it's computing a new set of parents, but these parents can definitely include thier current ancestors.

lib/src/rewrite.rs
```diff
@@ -846,6 +831,35 @@ pub fn move_commits(
...
+/// only commits in the target set as parents. The parents of each commit are
+/// identical to the ones found using a preorder DFS of the node's ancestors,
+/// starting from the node itself, and avoiding traversing an edge if the
+/// parent is in the target set.
+fn compute_connected_target_commits_internal_parents(
```
yuja, line 876: nit: perhaps, this function can be renamed to `compute_internal_parents_within()` or something? I don't think "connected" is important here. If I understand it, this is the function to remap parents not included in `target_commit_ids` to ancestors by using the graph `connected_target_commits`.

lib/src/rewrite.rs
```diff
@@ -464,70 +464,107 @@ pub struct MoveCommitsStats {
...
+/// `new_children` can be rewritten, and there will be no cycles in the
+/// resulting graph. Commits in `target` should be in reverse topological order.
 pub fn move_commits(
     settings: &UserSettings,
     mut_repo: &mut MutableRepo,
     new_parent_ids: &[CommitId],
     new_children: &[Commit],
-    target_commits: &[Commit],
-    target_roots: Option<&[CommitId]>,
+    target: MoveCommitsTarget,
```
yuja, line 489: nit: `target: &MoveCommitsTarget` or `: MoveCommitsTarget<'_>` (`Commits(&'a [Commit])`)

lib/src/rewrite.rs
```diff
@@ -464,70 +464,107 @@ pub struct MoveCommitsStats {
...
+
+    match target {
+        MoveCommitsTarget::Commits(commits) => {
+            if commits.is_empty() {
+                return Ok(MoveCommitsStats {
+                    num_rebased_targets: 0,
+                    num_rebased_descendants: 0,
+                    num_skipped_rebases: 0,
+                    num_abandoned: 0,
+                });
```
yuja, line 506: nit: `#[derive(Default)]` could be added to `MoveCommitStats` to make this shorter.

lib/src/rewrite.rs
```diff
@@ -464,70 +464,107 @@ pub struct MoveCommitsStats {
...
+                .commits(mut_repo.store())
+                .try_collect()?;
+            target_commit_ids = target_commits.iter().ids().cloned().collect();
+
+            connected_target_commits = target_commits.iter().cloned().collect_vec();
+            connected_target_commits_internal_parents =
+                compute_connected_target_commits_internal_parents(
+                    &target_commit_ids,
+                    &connected_target_commits,
+                );
```
yuja, line 564: We don't have to resolve internal parents because targets are identical to connected targets?

yuja (approved):
Thanks.

lib/src/rewrite.rs
```diff
@@ -486,12 +487,7 @@ pub fn move_commits(
     options: &RebaseOptions,
 ) -> BackendResult<MoveCommitsStats> {
     if target_commits.is_empty() {
-        return Ok(MoveCommitsStats {
-            num_rebased_targets: 0,
-            num_rebased_descendants: 0,
-            num_skipped_rebases: 0,
-            num_abandoned: 0,
-        });
+        return Ok(Default::default());
```
samueltardieu: nit: even though this is used inconsistently throughout jj's code base, it is usually considered a better practice to write `MoveCommitsStats::default()` to make it clear what type is built here.

bnjmnt4n: Ah, thanks for the pointer. As you might be able to tell, I don't really write a lot of Rust, so I'm not familiar with quite a lot of these common practices.

bnjmnt4n:
@yuja @samueltardieu I've updated some of the function names and made the suggested tweaks. Do you think there's anything else which should be changed?

lib/src/rewrite.rs
```diff
@@ -471,85 +472,91 @@ pub struct MoveCommitsStats {
     pub num_abandoned: u32,
 }
 
+pub enum MoveCommitsTarget {
+    /// The commits to be moved. Commits should be mutable and in reverse
+    /// topological order.
+    Commits(Vec<Commit>),
+    /// The root commits to be moved, along with all their descendants.
+    /// Root commits should be mutable and in reverse topological order.
```
yuja, line 480: nit: I don't think roots have to be in topological order. CLI wouldn't guarantee that.

yuja (approved):
These looks good to me, thanks.

PR #4662, opened 2024-10-17
style: remove unneeded lifetimes from trait implementations

samueltardieu:
Clippy 1.83 (currently in beta) detects more cases of unneeded lifetimes, namely in trait implementation declarations. Since this lint is warn by default (this is not something we opted in specifically), we will have to fix those instances to get a clean CI.


yuja (approved):
> Clippy 1.83 (currently in beta) detects more cases of unneeded lifetimes, namely in trait implementation declarations. Since this lint is warn by default (this is not something we opted in specifically), we will have to fix those instances to get a clean CI.

Can you add this to the commit description?

PR #4675, opened 2024-10-19
rewrite: don't resolve intermediate parent tree when rebasing

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_merge_trees.rs
+++ b/lib/tests/test_merge_trees.rs
@@ -645,3 +645,70 @@ fn test_simplify_conflict_after_resolving_parent() {
 
 // TODO: Add tests for simplification of multi-way conflicts. Both the content
 // and the executable bit need testing.
+
+#[test]
+fn test_rebase_on_lossy_merge() {
+    let settings = testutils::user_settings();
+    let test_repo = TestRepo::init();
+    let repo = &test_repo.repo;
+
+    // Test this rebase:
+    // D    foo=2          D'   foo=3
+    // |\                  |\
+    // | C  foo=2          | C' foo=3
+    // | |           =>    | |
+    // B |  foo=2          B |  foo=2
+    // |/                  |/
+    // A    foo=1          A    foo=1
+    //
+    // Commit D effectively discarded a change from "1" to "2", so one
+    // reasonable result in D' is "3". That's what the result would be if we
+    // didn't have the "A+(A-B)=A" rule. It's also what the result currently
+    // is because we don't attempt to resolve the auto-merged parents (if we
+    // had, it would have been resolved to just "2" before the rebase and we
+    // get a conflict after the rebase).
+    let path = RepoPath::from_internal_string("foo");
+    let mut tx = repo.start_transaction(&settings);
+    let repo_mut = tx.repo_mut();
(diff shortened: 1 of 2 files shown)


yuja (approved):
Out of curiosity, is this change for correctness, performance reason, or both?

martinvonz:
> Out of curiosity, is this change for correctness, performance reason, or both?

I'm not sure :) I wrote it many months ago (just before your similar improvements to avoid intermediate resolution in octopus merges and whatever else it was).

I think it was while I was thinking of a related problem, which is that conflicts can accrue many sides as they get rebased. I think that only happens if different files have conflicts in different rebases, so I think you can get at most N+1 sides if you have N files. I've been thinking that perhaps we should store the unmodified trees when a conflict can't be automatically resolved. That would lead to having to redo some work when you edit, diff, etc., but it would also lead to less work then you rebases repeatedly, thanks to earlier conflict simplification.

Anyway, I have not noticed in practice the performance problem this PR addresses (but it seems like it shouldn't be hard to produce work a somewhat complex merge).

lib/tests/test_merge_trees.rs
```diff
@@ -645,3 +645,65 @@ fn test_simplify_conflict_after_resolving_parent() {
...
+    // Test this rebase:
+    // D    foo=2          D'   foo=3
+    // |\                  |\
+    // | C  foo=2          | C' foo=3
+    // | |           =>    | |
+    // B |  foo=2          B |  foo=2
+    // |/                  |/
+    // A    foo=1          A    foo=1
+    // Commit D effectively discarded a change from "1" to "2", so the result should
+    // be "3".
```
ilyagr, line 664: Somewhat optional: I would explain this in more detail. I don't think "the result should be 3", I think the result of a conflict or 3 is equally right and equally problematic (if we take `A + A - B = A` seriously), and it might change later. In other words, it depends on whether we think of `D` as the "empty commit" or as a snapshot that tries to follow normal patch theory rules (without `A + A - B = A`).

This is not necessarily blocking, I can write something up later if I remember to do it.

ilyagr: In any case, good job finding this example, it's very illustrative as an example of A + A - B = A confusion.

martinvonz: I've added more explanation to the code comment.

joyously: The additional explanation doesn't help. Is the original `D` supposed to be "foo=3", so that your comment
> Commit D effectively discarded a change from "1" to "2"

actually makes sense, or should it be referring to `D'`?
Maybe instead of just saying "Test this rebase", you state what "this" is.

martinvonz: `A` starts with contents "1" in `foo`. `B` and `C` both represent a patch that changes "1" to "2" (in `foo`). So the merge of them should strictly speaking be a conflict (and some VCSs like Darcs and Pijul would consider it a conflict). Still, the state in `D` is just "2" (not a conflict). That's what I mean when I said it effectively discards a change from "1" to "2".

I think it's clear enough for those likely to work on this code. We've talked about the conflict algebra often enough that I think they will understand.

PR #4680, opened 2024-10-20
Default template alias to show full description

Grillo-0:
# Checklist

If applicable:

### DIFF
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -91,22 +91,13 @@ if(root,
   )
 )
 '''
+
 builtin_log_compact = '''
 if(root,
   format_root_commit(self),
   label(if(current_working_copy, "working_copy"),
     concat(
-      separate(" ",
-        format_short_change_id_with_hidden_and_divergent_info(self),
-        format_short_signature(author),
-        format_timestamp(committer.timestamp()),
-        bookmarks,
-        tags,
-        working_copies,
-        if(git_head, label("git_head", "git_head()")),
-        format_short_commit_id(commit_id),
-        if(conflict, label("conflict", "conflict")),
-      ) ++ "\n",
+      format_short_commit_header(self) ++ "\n",
       separate(" ",
         if(empty, label("empty", "(empty)")),
         if(description,
@@ -118,6 +109,25 @@ if(root,
   )
 )
 '''
+
+builtin_log_compact_full_description = '''
+if(root,
+  format_root_commit(self),
+  label(if(current_working_copy, "working_copy"),
+    concat(
+      format_short_commit_header(self) ++ "\n",
+      separate(" ",
+        if(empty, label("empty", "(empty)")),
+        if(description,
+          description,
+          label(if(empty, "empty"), description_placeholder),
+        ),
+      ) ++ "\n",
+    ),
+  )
+)
+'''
+
 builtin_log_comfortable = 'builtin_log_compact ++ "\n"'
 builtin_log_detailed = '''
 concat(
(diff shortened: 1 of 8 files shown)


docs/config.md
```diff
@@ -299,6 +299,24 @@ revsets.log = "main@origin.."
...
+```toml
+[templates]
+# Use builtin log template
+log = "builtin_log_compact"
+# Use builtin op log template
+op_log = "builtin_log_compact"
+# Use builtin show template
+show = "builtin_log_detailed"
+```
+
```
martinvonz, line 319: If we think it's pretty common to want the full description (as we presumably do since we're adding this feature), it might be good to include an example using `builtin_log_compact_full_description` here.

martinvonz (commented):
LGTM, but let's give others a chance to review it too.

cli/src/config/templates.toml
```diff
@@ -90,8 +90,9 @@ if(root,
     ),
   )
 )
+
 '''
-builtin_log_compact = '''
+'builtin_log_compact_desc(full_description)' = '''
```
yuja, line 95: nit: maybe better to extract the first line (= `separate(" ", .. if(conflict ..))`) as `format_compact_commit_header(commit)`? I don't know if we want to reuse this function for other templates, but user might want to choose where to put "(empty)" marker.

PhilipMetzger (approved):
LG, thanks

cli/src/config/templates.toml
```diff
@@ -91,12 +91,9 @@ if(root,
...
 )
 '''
-builtin_log_compact = '''
-if(root,
-  format_root_commit(self),
-  label(if(current_working_copy, "working_copy"),
-    concat(
-      separate(" ",
+
+builtin_log_compact_commit_header = '''
```
yuja, line 95: nit: `format_...(commit)` for naming consistency and hide it from the list of builtin aliases.

cli/src/config/templates.toml
```diff
@@ -243,6 +238,20 @@ if(commit.hidden(),
 )
 '''
 
+'format_commit_header(commit)' = '''
```
yuja, line 241: nit: rename to `format_compact_commit_header` or `format_short_commit_header`?

If we split `builtin_log_detailed`, the header part would be called "format_detailed_commit_header" or something.

cli/src/config/templates.toml
```diff
@@ -243,6 +238,20 @@ if(commit.hidden(),
...
+separate(" ",
+        format_short_change_id_with_hidden_and_divergent_info(commit),
+        format_short_signature(author),
+        format_timestamp(committer.timestamp()),
+        bookmarks,
+        tags,
+        working_copies,
+        if(git_head, label("git_head", "git_head()")),
+        format_short_commit_id(commit.commit_id()),
+        if(conflict, label("conflict", "conflict")),
```
yuja, line 251: nit: could you update keywords "author", etc. to method calls `commit.author()`, etc.?

cli/src/config/templates.toml
```diff
@@ -91,33 +91,28 @@ if(root,
...
-        if(git_head, label("git_head", "git_head()")),
-        format_short_commit_id(commit_id),
-        if(conflict, label("conflict", "conflict")),
-      ) ++ "\n",
+      format_commit_header(self) ++ "\n",
       separate(" ",
         if(empty, label("empty", "(empty)")),
         if(description,
-          description.first_line(),
+          if(full_description, description, description.first_line()),
```
yuja, line 104: nit: I wouldn't add `builtin_log_compact_desc()` function (which seems not so useful), and just copy-paste the second line.

yuja (approved):
Thanks!

cli/src/config/templates.toml
Grillo-0: Is that what you meant?

yuja: Yes, thanks.

Grillo-0:
Will merge this

PR #4687, opened 2024-10-22
templates: add `.email_domain()` method to `Signature`s

ilyagr:
This is slightly less pretty than making an Email type, but I think is much easier.

I would use instead of the last argument of:

```
  concat(
    label("email", signature.username()),
    "@",
    label("email",
      signature.email().remove_prefix(
        signature.username()
      ).remove_prefix("@")
    )
  )
```

(Aside for more motivation: I initially had a bug above by using `.remove_prefix(signature.username() ++ "@")`. This is wrong since an email might not contain an `@`, as I discovered while looking at the code. The bug is now fixed, but `signature.email_domain()` would avoid the problem)

The effect I was going for (though I'm still not sure whether and how many characters I want to trim to):

image

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Timestamp objects in templates now have `after(date) -> Boolean` and
   `before(date) -> Boolean` methods for comparing timestamps to other dates.
 
+* Signature objects in templates now have a `email_domain() -> String` method.
+
 * New template functions `pad_start()`, `pad_end()`, `truncate_start()`, and
   `truncate_end()` are added.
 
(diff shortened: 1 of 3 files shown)


yuja:
Maybe it's time to add an `Email` template type (and deprecate `.username()`)? Suppose the goal is to label `@` differently, an `Email` object could be rendered to `<local>foo</local>@<domain>example.org</domain>` by default.

ilyagr:
That would be better, but also a lot more work. As I mentioned, this is giving me pause at the moment TBH. I haven't decided whether/when I want to do it.

bnjmnt4n:
Closing as https://github.com/jj-vcs/jj/pull/5079 has been merged.

PR #4710, opened 2024-10-25
diff: Allow setting the default level of context in config.

Mossop:
I'm used to more context by default.

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/misc.toml
+++ b/cli/src/config/misc.toml
@@ -8,6 +8,10 @@ unamend = ["unsquash"]
 
 [diff.color-words]
 max-inline-alternation = 3
+context = 3
+
+[diff.git]
+context = 3
 
 [ui]
 # TODO: delete ui.allow-filesets in jj 0.26+
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -184,7 +184,7 @@ fn diff_formats_from_args(
         formats.push(DiffFormat::NameOnly);
     }
     if args.git {
-        let options = UnifiedDiffOptions::from_args(args);
+        let options = UnifiedDiffOptions::from_settings_and_args(settings, args)?;
         formats.push(DiffFormat::Git(Box::new(options)));
     }
     if args.color_words {
@@ -230,7 +230,7 @@ fn default_diff_format(
         "types" => Ok(DiffFormat::Types),
         "name-only" => Ok(DiffFormat::NameOnly),
         "git" => {
-            let options = UnifiedDiffOptions::from_args(args);
+            let options = UnifiedDiffOptions::from_settings_and_args(settings, args)?;
             Ok(DiffFormat::Git(Box::new(options)))
         }
         "color-words" => {
@@ -520,8 +520,11 @@ impl ColorWordsDiffOptions {
                 })?),
             }
         };
+        let context = args
+            .context
+            .map_or_else(|| config.get("diff.color-words.context"), Ok)?;
         Ok(ColorWordsDiffOptions {
-            context: args.context.unwrap_or(DEFAULT_CONTEXT_LINES),
+            context,
             line_diff: LineDiffOptions::from_args(args),
             max_inline_alternation,
         })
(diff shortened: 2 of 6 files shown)


cli/src/config/misc.toml
```diff
@@ -6,6 +6,9 @@ b = ["bookmark"]
 co = ["checkout"]
 unamend = ["unsquash"]
 
+[diff]
+context = 3
```
yuja, line 10: Since we have `[diff.<format>]` namespace, I think it's better to add `diff.color-words.context` and `diff.git.context` separately.

cli/src/diff_util.rs
```diff
@@ -520,8 +520,21 @@ impl ColorWordsDiffOptions {
...
+        } else {
+            let key = "diff.context";
+            let value = config.get_int(key)?;
+            if value < 0 {
+                return Err(config::ConfigError::Message(format!(
+                    "context must be at least 0, got {value}"
+                )));
+            }
+            usize::try_from(value)
+                .map_err(|err| config::ConfigError::Message(format!("invalid {key}: {err}")))?
```
yuja, line 534: nit: `config.get(key)?` (i.e. `config.get::<usize>()`) should produce a good-enough error message.

cli/tests/test_diff_command.rs
```diff
@@ -1736,12 +1736,209 @@ fn test_diff_skipped_context() {
...
+    test_env.jj_cmd_ok(&repo_path, &["new", "@-", "-m", "=== No gap to skip"]);
+    std::fs::write(repo_path.join("file1"), "a\nB\nc\nd\ne\nf\ng\nh\nI\nj").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["new", "@-", "-m", "=== No gap to skip"]);
+    std::fs::write(repo_path.join("file1"), "a\nb\nC\nd\ne\nf\ng\nh\nI\nj").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["new", "@-", "-m", "=== 1 line at start"]);
+    std::fs::write(repo_path.join("file1"), "a\nb\nc\nd\nE\nf\ng\nh\ni\nj").unwrap();
+    test_env.jj_cmd_ok(&repo_path, &["new", "@-", "-m", "=== 1 line at end"]);
+    std::fs::write(repo_path.join("file1"), "a\nb\nc\nd\ne\nF\ng\nh\ni\nj").unwrap();
+
+    let stdout = test_env.jj_cmd_success(
```
yuja, line 1768: Can you simplify the test? We only need to ensure that the config value is applied.

cli/tests/test_diff_command.rs
```diff
@@ -1736,12 +1736,209 @@ fn test_diff_skipped_context() {
...
     test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
     let repo_path = test_env.env_root().join("repo");
 
+    // Verify that the command line overrides a user setting.
+    test_env.add_config(
+        r#"
+[diff]
+context = 4
+        "#,
+    );
```
yuja, line 1940: nit: the default config includes `diff.context = 3`, so we don't have to override it?

yuja (commented):
Thanks!

Mossop:
Updated with the requested changes.

cli/src/diff_util.rs
```diff
@@ -520,8 +520,13 @@ impl ColorWordsDiffOptions {
                 })?),
             }
         };
+        let context = if let Some(c) = args.context {
+            c
+        } else {
+            config.get::<usize>("diff.color-words.context")?
```
yuja, line 526: nit: `::<usize>` can be omitted. (I would add `let context: usize` if I need to make the type clear.)

We can also use `map_or_else()`, but I don't have any preference about that.
https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.map_or_else

cli/tests/test_diff_command.rs
```diff
@@ -1736,6 +1736,45 @@ fn test_diff_skipped_context() {
     "###);
 }
 
+#[test]
+fn test_diff_skipped_context_fromsettings() {
```
yuja, line 1740: nit: `_from_settings`

cli/tests/test_diff_command.rs
```diff
@@ -1736,6 +1736,45 @@ fn test_diff_skipped_context() {
...
+            4: d
+            5: e
+    === Must show 0 context
+    Modified regular file file1:
+        ...
+       3    3: cC
+        ...
+    "#);
+}
+
```
yuja, line 1777: can you add test for `diff.git.context`?

yuja (approved):
Thanks for quick updates.

cli/src/diff_util.rs
Mossop: What an interesting pattern, thanks for pointing it out.

Mossop:
Updated.

PR #4714, opened 2024-10-25
templates: respect format_short_signature() in builtin_log_oneline

neongreen:
I noticed that `jj log -T builtin_log_oneline` wasn't respecting my `format_short_signature()` settings.

This is a slightly breaking change -- people's `oneline` logs will become longer if they use the default `format_short_signature()`. Perhaps there should be a new override, eg. `format_tiny_signature()`?

# Checklist

*not updating tests/changelog yet since idk if the PR will be merged*

If applicable:

### DIFF
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -73,7 +73,7 @@ if(root,
     concat(
       separate(" ",
         format_short_change_id_with_hidden_and_divergent_info(self),
-        if(author.email(), author.username(), email_placeholder),
+        format_short_signature(author),
         format_timestamp(committer.timestamp()),
         bookmarks,
         tags,


neongreen:
@ilyagr I just noticed that you also want to have `format_very_short_signature()`. If it existed, I suppose it would be used both in `jj annotate` and in `builtin_log_oneline`?

ilyagr:
I don't have a strong opinion, but whether or not we decide to show the author in the default annotation template (as discussed in #4653), Emily's idea of a `format_very_short_signature` that's used here and in a (possibly non-default) annotation template seems good to me. We could do this at the same time as or shortly after such an annotation template exists.

For example, in https://github.com/martinvonz/jj/pull/4653#discussion_r1809341388, I was thinking that `username@` might look better than just `username`, and `format_very_short_signature` would make that consistent (again, regardless of whether this is a new default or a user's customization).

Another possible value for `format_very_short_signature` might be `truncate_end(email, 12)`, though then you'd have to adjust the annotation template to use the same padding.

-------

I like `format_very_short_signature` slightly better than this PR's original idea using `format_short_signature` in the oneline template, since the default will become verbose, but either could make sense, and there is some downside for having one more signature template (which we effectively already do have, but this PR would remove). I don't use the oneline template enough to have a strong opinion.

neongreen:
I’m in the same boat as @emilazy:

> I dislike how we have an entire field in commits that represents an appropriate human‐readable identifier to use for their authors and committers but ignore it to just show email addresses instead, […] Of course I’m biased since the local‐parts of my email addresses don’t identify me at all.

So, it’s not the default I particularly care about — just the ability to customize the name display, without having to copy templates into my config and keep up with future jj changes.

ilyagr:
In that case, it sounds like creating a "very short signature" template alias will help you, work for a future annotation template, and be a noop for everybody else. That sounds good to me, even if we later decide to change our mind and, say, unify two of the signature slides.

So, let's do that unless we hear another opinion?

emilazy:
At this rate, perhaps we could simply hard‐code it to output `"Emily"`.

ilyagr:
Just to be clear, I'm waiting for you to adjust the PR and I'm assuming you just haven't gotten around to it yet (which is totally fine!). But let me know if you're waiting on something instead.

neongreen:
Yeah, just haven’t gotten around to it; I’ll convert it into draft for now

PR #4720, opened 2024-10-27
docs: Replace "first make sure that nothing is staged" comment with equivalent `--only` flag

rmacklin:
This part of the git-comparison table was implying you need to manually check and clear staged changes in order to only update the previous commit's message. That isn't the case, though. As mentioned in the [`git-commit` documentation][1], this can be directly achieved by using the `--only` flag with `--amend`:

> If this option is specified together with --amend, then no paths need
> to be specified, which can be used to amend the last commit without
> committing changes that have already been staged.

[1]: https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--o

### DIFF
--- a/docs/git-comparison.md
+++ b/docs/git-comparison.md
@@ -239,8 +239,7 @@ parent.
     <tr>
       <td>Edit description (commit message) of the previous change</td>
       <td><code>jj describe @-</code></td>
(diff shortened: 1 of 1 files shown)


martinvonz (approved):
Thanks. I was not aware of that flag (turns out it's ancient, so maybe I had just forgotten about it).

PR #4723, opened 2024-10-27
docs for diff-editor: change KDiff3 -> Meld, advertise Meld

ilyagr:
I am not sure whether it's related, but I notice more and more people not knowing about using external diff editors, which are quite different from the built-in one. I'm not sure, but perhaps this will help with e.g. https://github.com/martinvonz/jj/discussions/1905.

# Checklist

If applicable:


docs/config.md
```diff
@@ -566,37 +566,47 @@ Obviously, you would only set one line, don't copy them all in!
...
+to edit the diff in your terminal.
 
 [scm-diff-editor]: https://github.com/arxanas/scm-record?tab=readme-ov-file#scm-diff-editor
 
+You can try a different tool temporarily by doing e.g. `jj split --tool meld` or
+you can set the option to change the default. This requires that you have an
+appropriate tool installed, e.g. [Meld](https://meldmerge.org/) to use the
+`meld` diff editor.
+
+If `ui.diff-editor` consists of a single word, e.g. `"meld"`, the arguments
```
yuja, line 590: nit: "consists of a single word" -> "is a string"

Tool names don't usually contain whitespace, but it's not a hard requirement.

yuja (approved):

docs/config.md
ilyagr: Thanks! This is quite an important distinction, one could wrongly assume that spaces split arguments.

yuja: String argument is split by whitespace, but table lookup is prioritized.

```
"name in the merge-tools" 
| "command_name with arg .." 
| ["command_name", "with", "arg", ..]
# ["name in the merge-tools"]  # This is not allowed
```

ilyagr: Thanks. As far as this PR is concerned, I won't do anything about this.

More generally, there is one case I can imagine this being useful, and it doesn't work in this case: `ui.diff-editor = "meld --new-window"`. It doesn't work because `$left` and `$right` are not added automatically in this case (as I found out by using `echo`).

I could fix it, but it's probably better to remove this special case entirely, right?

Otherwise, I could briefly document it and/or fix it.

yuja: There isn't any special case. "meld --new-window" doesn't work because there isn't any tool named "meld --new-window".

Well, we could say that there is one special case that a string may be parsed as a list of command arguments. We can deprecate this syntax in favor of `["name", "arg", ..]`, but I'm not sure if people like this idea.

ilyagr: I meant that, if we want to have the split-by-whitespace syntax, we could have "meld new-window" interpreted as `["meld", "--new-window", "$left", "$right"]` (if there is not tool named "meld --new-window", of course). This would be analogous to how "binary" is interpreted as `["binary", "$left", $right"]`.

Currently, you'd have to write out `ui.diff-editor = "meld new-window $left $right"`, which doesn't seem very useful to me. So, I think deprecating the space-separated syntax might be even better.

BTW, while we allow space-separated syntax, I don't think `ui.diff-editor="C:\Program Files\Meld\meld.exe"` is valid, you'd have to use the list syntax.

yuja: > if we want to have the split-by-whitespace syntax (which is what I meant by the "special case"), we could have "meld new-window" interpreted as `["meld", "--new-window", "$left", "$right"]` (if there is not tool named "meld --new-window", of course). This would be analogous to how "binary" is interpreted as `["binary", "$left", $right"]`.

It's rather confusing. If `"<TOOL_NAME> additional args"` were allowed, `["<TOOL_NAME>", "additional", "args"]` would also have to be translated to `tools[TOOL_NAME].args + ["additional", "args"]`, for consistency. And we might need a flag whether additional args should be prepended or appended.

PR #4724, opened 2024-10-28
annotate: refactor, fix propagation of duplicated lines at fork/merge point

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_annotate.rs
+++ b/lib/tests/test_annotate.rs
@@ -183,7 +183,59 @@ fn test_annotate_merge_split() {
 }
 
 #[test]
-#[should_panic] // FIXME: shouldn't panic because of duplicated "1"s
+fn test_annotate_merge_split_interleaved() {
+    let settings = testutils::user_settings();
+    let test_repo = TestRepo::init();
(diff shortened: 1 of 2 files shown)


lib/tests/test_annotate.rs
```diff
@@ -215,6 +267,9 @@ fn test_annotate_merge_dup() {
     let commit4 = create_commit("commit4", &[commit2.id(), commit3.id()], tree4.id());
     drop(create_commit);
 
+    // Both "1"s can be propagated to commit1 through commit2 and commit3.
+    // Alternatively, it's also good to interpret that one of the "1"s was
+    // produced at either commit2 or commit3.
```
martinvonz, line 272: Or even produced by commit 4? If you just merge 2 and 3, you'll get "2 1 3", so it seems reasonable to say that one of the 1's came from commit 4.

martinvonz (approved):

lib/tests/test_annotate.rs
yuja: Correct. If we had a step to diff against auto-merge parent, "1" would be left there. Updated the comment.

PR #4733, opened 2024-10-30
bookmarks: add support for `git.auto-local-bookmark` (to match docs)

### LINKED ISSUE #4732: Bookmark auto-tracking is broken (as documented)
## Description

[The docs](https://martinvonz.github.io/jj/latest/config/#automatic-local-bookmark-creation) mention the `git.auto-local-bookmark` setting and is called that throughout the docs. The codebase uses both that name as well as `git.auto-local-bookmark` somewhat inconsistently. The line where that matters most is the following: https://github.com/martinvonz/jj/blob/e464c0e607fd9b7308c8c0666f5d5b2c5c9596c9/lib/src/settings.rs#L54

This results in the unfortunate effect that any setting `git.auto-local-bookmark = true` will be ignored leading to unexpecting behavior when following the documented approach

1. Remove any config for `git.auto-local-branch = true` - branches are now called bookmarks after all
1. Add `git.auto-local-bookmark = true` to your config
1. Fetch changes containing a new bookmark

## Expected Behavior

`git.auto-local-branch` should be defunct (or emit a deprecation warning perhaps?) and `git.auto-local-bookmark` should take over that functionality.

Alternatively: If `git.auto-local-branch` is the intended name for that setting going forward, the field in the Settings struct should be renamed to match the name and the documentation should be updated to reflect it as well.

## Actual Behavior

Following the steps above newly fetched bookmarks aren't tracked

## Specifications

- Platform: macOS Sequoia 15.0.1
- Version: 0.22.0 (also present on the current main, see permalink above for exact revision)

martinvonz:
Closes #4732

# Checklist

If applicable:

### DIFF
--- a/lib/src/settings.rs
+++ b/lib/src/settings.rs
@@ -50,11 +50,16 @@ pub struct GitSettings {
 
 impl GitSettings {
     pub fn from_config(config: &config::Config) -> Self {
+        let auto_local_bookmark = config
+            .get_bool("git.auto-local-bookmark")
+            .or_else(|_| config.get_bool("git.auto-local-branch"))
+            .unwrap_or(false);
+        let abandon_unreachable_commits = config
+            .get_bool("git.abandon-unreachable-commits")
+            .unwrap_or(true);
         GitSettings {
-            auto_local_bookmark: config.get_bool("git.auto-local-branch").unwrap_or(false),
-            abandon_unreachable_commits: config
-                .get_bool("git.abandon-unreachable-commits")
-                .unwrap_or(true),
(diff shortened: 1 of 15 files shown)


lib/src/settings.rs
yuja, line 1: need to update cli/src/config-schema.json

yuja (approved):

lib/src/settings.rs
martinvonz: Ah, thanks. I've fixed that and two other instances (in test comments) that I had missed by searching for the full config name.

PR #4737, opened 2024-10-30
add dynamic completions

senekor:
This is a second attempt to solve https://github.com/martinvonz/jj/issues/1217.

Currently, it's just a proof of concept to gather feedback. What it does is complete local bookmark names for `jj bookmark rename`.

To test it yourself, source the new completions scripts in your shell. For bash, as example:
```sh
# cd jj
# cargo build
source <(COMPLETE=bash ./target/debug/jj)
```
Instructions for each shell are [here](https://docs.rs/clap_complete/latest/clap_complete/env/index.html).

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * New command `jj util exec` that can be used for arbitrary aliases.
 
+* A preview of improved shell completions was added. Please refer to the
+  [documentation](https://martinvonz.github.io/jj/latest/install-and-setup/#command-line-completion)
+  to activate them.
+
 ### Fixed bugs
 
 ## [0.23.0] - 2024-11-06
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -394,6 +394,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "11611dca53440593f38e6b25ec629de50b14cdfa63adc0fb856115a2c6d97595"
 dependencies = [
  "clap",
+ "clap_lex",
+ "is_executable",
+ "shlex",
 ]
 
 [[package]]
@@ -1752,6 +1755,15 @@ dependencies = [
  "hermit-abi 0.4.0",
  "libc",
  "windows-sys 0.52.0",
+]
+
+[[package]]
+name = "is_executable"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d4a1b5bad6f9072935961dfbf1cced2f3d129963d091b6f69f007fe04e758ae2"
+dependencies = [
+ "winapi",
 ]
 
 [[package]]
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -17,7 +17,6 @@ use std::cell::OnceCell;
 use std::collections::BTreeMap;
 use std::collections::HashSet;
 use std::env;
-use std::env::ArgsOs;
 use std::env::VarError;
 use std::ffi::OsString;
 use std::fmt;
@@ -2128,7 +2127,7 @@ impl WorkspaceCommandTransaction<'_> {
     }
 }
 
-fn find_workspace_dir(cwd: &Path) -> &Path {
+pub fn find_workspace_dir(cwd: &Path) -> &Path {
     cwd.ancestors()
         .find(|path| path.join(".jj").is_dir())
         .unwrap_or(cwd)
@@ -3057,7 +3056,7 @@ fn handle_early_args(
 pub fn expand_args(
     ui: &Ui,
     app: &Command,
-    args_os: ArgsOs,
+    args_os: impl IntoIterator<Item = OsString>,
     config: &config::Config,
 ) -> Result<Vec<String>, CommandError> {
     let mut string_args: Vec<String> = vec![];
@@ -3379,6 +3378,16 @@ impl CliRunner {
     #[must_use]
     #[instrument(skip(self))]
     pub fn run(mut self) -> ExitCode {
+        match clap_complete::CompleteEnv::with_factory(|| self.app.clone())
+            .try_complete(env::args_os(), None)
+        {
+            Ok(true) => return ExitCode::SUCCESS,
+            Err(e) => {
+                eprintln!("failed to generate completions: {e}");
+                return ExitCode::FAILURE;
+            }
+            Ok(false) => {}
+        };
         let builder = config::Config::builder().add_source(crate::config::default_config());
         let config = self
             .extra_configs
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -1,0 +1,126 @@
+// Copyright 2024 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::FromArgMatches as _;
+use clap_complete::CompletionCandidate;
+use jj_lib::workspace::DefaultWorkspaceLoaderFactory;
+use jj_lib::workspace::WorkspaceLoaderFactory as _;
+
+use crate::cli_util::expand_args;
+use crate::cli_util::find_workspace_dir;
+use crate::cli_util::GlobalArgs;
+use crate::command_error::user_error;
+use crate::command_error::CommandError;
+use crate::config::default_config;
+use crate::config::LayeredConfigs;
+use crate::ui::Ui;
+
+pub fn local_bookmarks() -> Vec<CompletionCandidate> {
+    with_jj(|mut jj| {
+        let output = jj
+            .arg("bookmark")
+            .arg("list")
+            .arg("--template")
+            .arg(r#"if(!remote, name ++ "\n")"#)
+            .output()
+            .map_err(user_error)?;
+
+        Ok(String::from_utf8_lossy(&output.stdout)
+            .lines()
+            .map(CompletionCandidate::new)
+            .collect())
+    })
+}
+
+/// Shell out to jj during dynamic completion generation
+///
+/// In case of errors, print them and early return an empty vector.
+fn with_jj<F>(completion_fn: F) -> Vec<CompletionCandidate>
+where
+    F: FnOnce(std::process::Command) -> Result<Vec<CompletionCandidate>, CommandError>,
+{
+    get_jj_command()
+        .and_then(completion_fn)
+        .unwrap_or_else(|e| {
+            eprintln!("{}", e.error);
+            Vec::new()
+        })
+}
+
+/// Shell out to jj during dynamic completion generation
+///
+/// This is necessary because dynamic completion code needs to be aware of
+/// global configuration like custom storage backends. Dynamic completion
+/// code via clap_complete doesn't accept arguments, so they cannot be passed
+/// that way. Another solution would've been to use global mutable state, to
+/// give completion code access to custom backends. Shelling out was chosen as
+/// the preferred method, because it's more maintainable and the performance
+/// requirements of completions aren't very high.
+fn get_jj_command() -> Result<std::process::Command, CommandError> {
+    let current_exe = std::env::current_exe().map_err(user_error)?;
+    let mut command = std::process::Command::new(current_exe);
+
+    // Snapshotting could make completions much slower in some situations
+    // and be undesired by the user.
+    command.arg("--ignore-working-copy");
+    command.arg("--color=never");
+    command.arg("--no-pager");
+
+    // Parse some of the global args we care about for passing along to the
+    // child process. This shouldn't fail, since none of the global args are
+    // required.
+    let app = crate::commands::default_app();
+    let config = config::Config::builder()
+        .add_source(default_config())
+        .build()
+        .expect("default config should be valid");
+    let mut layered_configs = LayeredConfigs::from_environment(config);
+    let ui = Ui::with_config(&layered_configs.merge()).expect("default config should be valid");
+    let cwd = std::env::current_dir()
+        .and_then(|cwd| cwd.canonicalize())
+        .map_err(user_error)?;
+    let maybe_cwd_workspace_loader = DefaultWorkspaceLoaderFactory.create(find_workspace_dir(&cwd));
+    layered_configs.read_user_config().map_err(user_error)?;
+    if let Ok(loader) = &maybe_cwd_workspace_loader {
+        layered_configs
+            .read_repo_config(loader.repo_path())
+            .map_err(user_error)?;
+    }
+    let config = layered_configs.merge();
+    // skip 2 because of the clap_complete prelude: jj -- jj <actual args...>
+    let args = std::env::args_os().skip(2);
+    let args = expand_args(&ui, &app, args, &config)?;
+    let args = app
+        .clone()
+        .disable_version_flag(true)
+        .disable_help_flag(true)
+        .ignore_errors(true)
+        .try_get_matches_from(args)?;
+    let args: GlobalArgs = GlobalArgs::from_arg_matches(&args)?;
+
+    if let Some(repository) = args.repository {
+        command.arg("--repository");
+        command.arg(repository);
+    }
+    if let Some(at_operation) = args.at_operation {
+        command.arg("--at-operation");
+        command.arg(at_operation);
+    }
+    for config_toml in args.early_args.config_toml {
+        command.arg("--config-toml");
+        command.arg(config_toml);
+    }
+
+    Ok(command)
+}
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -210,10 +210,23 @@ To set up command-line completion, source the output of
 `jj util completion bash/zsh/fish`. Exactly how to source it
 depends on your shell.
 
+Improved completions are currently in the works, these will complete things
+like bookmark names as well. You can activate them with the alternative "dynamic"
+instructions below. Please let us know if you encounter any issues, so we can
+ensure a smooth transition once we default to these new completions. If you
+have ideas about specific completions that could be added, please share them
+[here](https://github.com/martinvonz/jj/issues/4763).
+
 ### Bash
 
 ```shell
 source <(jj util completion bash)
+```
+
+dynamic:
+
+```shell
+echo "source <(COMPLETE=bash jj)" >> ~/.bashrc
 ```
 
 ### Zsh
@@ -224,10 +237,22 @@ compinit
 source <(jj util completion zsh)
 ```
 
+dynamic:
+
+```shell
+echo "source <(COMPLETE=zsh jj)" >> ~/.zshrc
+```
+
 ### Fish
 
 ```shell
 jj util completion fish | source
+```
+
+dynamic:
+
+```shell
+echo "source (COMPLETE=fish jj | psub)" >> ~/.config/fish/config.fish
 ```
 
 ### Nushell
@@ -237,8 +262,12 @@ jj util completion nushell | save completions-jj.nu
 use completions-jj.nu *  # Or `source completions-jj.nu`
 ```
 
+(dynamic completions not available yet)
+
 ### Xonsh
 
 ```shell
 source-bash $(jj util completion)
 ```
+
+(dynamic completions not available yet)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ clap = { version = "4.5.20", features = [
     "wrap_help",
     "string",
 ] }
-clap_complete = "4.5.37"
+clap_complete = { version = "4.5.37", features = ["unstable-dynamic"] }
 clap_complete_nushell = "4.5.4"
 clap-markdown = "0.1.4"
 clap_mangen = "0.2.10"
--- a/cli/src/commands/bookmark/rename.rs
+++ b/cli/src/commands/bookmark/rename.rs
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use clap_complete::ArgValueCandidates;
 use jj_lib::op_store::RefTarget;
 
 use super::has_tracked_remote_bookmarks;
 use crate::cli_util::CommandHelper;
 use crate::command_error::user_error;
 use crate::command_error::CommandError;
+use crate::complete;
 use crate::ui::Ui;
 
 /// Rename `old` bookmark name to `new` bookmark name
@@ -26,6 +28,7 @@ use crate::ui::Ui;
 #[derive(clap::Args, Clone, Debug)]
 pub struct BookmarkRenameArgs {
     /// The old name of the bookmark
+    #[arg(add = ArgValueCandidates::new(complete::local_bookmarks))]
     old: String,
 
     /// The new name of the bookmark
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -19,6 +19,7 @@ pub mod cli_util;
 pub mod command_error;
 pub mod commands;
 pub mod commit_templater;
+pub mod complete;
 pub mod config;
 pub mod description_util;
 pub mod diff_util;
--- a/cli/tests/runner.rs
+++ b/cli/tests/runner.rs
@@ -18,6 +18,7 @@ mod test_builtin_aliases;
 mod test_checkout;
 mod test_commit_command;
 mod test_commit_template;
+mod test_completion;
 mod test_concurrent_operations;
 mod test_config_command;
 mod test_copy_detection;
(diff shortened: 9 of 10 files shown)


Cargo.lock
```diff
@@ -1754,6 +1757,15 @@ dependencies = [
  "windows-sys 0.52.0",
 ]
 
+[[package]]
+name = "is_executable"
```
senekor, line 1761: This is owned by Nick Fitzgerald, so trustworthy IMO.
https://github.com/fitzgen/is_executable

cli/src/cli_util.rs
```diff
@@ -3396,6 +3415,182 @@ impl CliRunner {
...
+///
+/// fn main() {
+///     let (ui, command_helper) = prepare_stuff();
+///     CliRunner::run(ui, command_helper);
+/// }
+///
+/// fn custom_completer() {
+///     let (ui, command_helper) = prepare_stuff();
+///     // do the thing
+/// }
```
senekor, line 3444: This is what I'm most unsure about. How to go about sharing this initialization logic between normal run path and completion logic?

It turns out I couldn't even copy-paste the initialization logic as-is. The completion logic cannot actually parse the CLI args, because the might still be invalid while the user/shell is requesting dynamic completions.

It seems tricky to share this logic in a nice way.

yuja: I have no idea how `ArgValueCompleter` is supposed to access global resources. We might have to store them in static variables if completion is requested. Something like

1. process args normally until somewhere in `run_internal()`
2. if `$COMPLETE` is set, stop there and copy resources to static variables
3. `try_complete()`

It's ugly, but we'll need some static to handle extensions registered by caller?

senekor: > I have no idea how `ArgValueCompleter` is supposed to access global resources.

From the perspective of `clap`, I don't think it's supposed to do that _at all_. The idea is for the completer functions to run on a completely separate code path from normal operation, which makes sense for most if not all applications.

Whenever I chose to use a static, I usually regretted it. What is the problem with extracting the shared logic into a function?

I'm also now thinking I might be approaching it completely wrong. I'm trying to access `ui` and `command_helper` while simply trying to get the list of local branches. Maybe I should use slightly lower-level APIs to implement completion logic? I'm not too familiar with the codebase yet, but theres a library part, maybe I should use that?

senekor: > we'll need some static to handle extensions registered by caller?

What do you mean by that?

yuja: > What extensions, registered how?

https://github.com/martinvonz/jj/blob/main/cli/examples/custom-backend/main.rs#L89

Maybe it can be registered at link time (with some static registration mechanisms), but it's currently constructed in call stack.

> Whenever I chose to use a static, I usually regretted it. 

I agree, but accessing `env::args()` is conceptually the same. They're bad, but work.

yuja: > In what sense do such extensions interact with completions?

Yes. The `repo` would fail to load if the store backend weren't registered, for example. Google implements in-house backend for their monorepo.

senekor: Aah, now I understand. The custom subcommands aren't the problem, but the storage backend is. I also can't see a different solution than a static variable in that case. I will try to implement that.

senekor: Alright, done. Thanks for taking the time to review :bow:

ilyagr:
cargo now does something similar with dynamic completions. So, I guess, the time has come. 🎉 

https://blog.rust-lang.org/inside-rust/2024/10/31/this-development-cycle-in-cargo-1.83.html#shell-completions

Many people looking at this probably already know, but it was news to me.

Thank you for starting this work!

senekor:
> https://blog.rust-lang.org/inside-rust/2024/10/31/this-development-cycle-in-cargo-1.83.html#shell-completions

This is really valuable to see how they did it. They prefer `ArgValueCandidates` over `ArgValueCompleter`, which I used previously. I assume `ArgValueCandidates` can be called fewer times and is therefore more efficient, so I use it now as well.

Also, they shell out to `rustc` and `rustup` for target triples. Maybe it would be an option for us as well, to simply shell out to ourselves to generate completion candidates? It will be less efficient, but we could remove the yucky global state. The custom completion code could become quite simple, e.g. calling `jj bookmark list -T 'if(!remote, name ++ "\n")'` for branches. On the other hand, we lose type safety. If there are breaking changes to the CLI, this will break without any help from the compiler.

So, not a complete win, but maybe worth considering. I lean towards whatever's more maintainable, completions don't have to be blazingly fast IMO.

yuja:
> Also, they shell out to `rustc` and `rustup` for target triples. Maybe it would be an option for us as well, to simply shell out to ourselves to generate completion candidates?

Sounds good to me. We'll still need to parse global args (such as `-R`, `--at-op`, and maybe have to do alias expansion to extract global args from there?), but that should be doable from static context.

> On the other hand, we lose type safety. If there are breaking changes to the CLI, this will break without any help from the compiler.

Maybe we can add snapshot tests?

senekor:
> Maybe we can add snapshot tests?

Great idea, I've added one. That should reduce the risk of completions breaking to a minimum. They are pretty easy to write that way too, feels like shell scripting. I like the current approach.

> We'll still need to parse global args (such as -R, --at-op, and maybe have to do alias expansion to extract global args from there?)

I haven't done that part yet, will get to it later. If stuff from a different repository than the CWD is completed correctly, that would feel pretty magical!

senekor:
[CI failure](https://github.com/martinvonz/jj/actions/runs/11649079354/job/32436209422#step:7:1844) seems unrelated to me? `test_git::test_concurrent_read_write_commit` is failing on Windows only... might be flaky if it's testing concurrency stuff.

docs/install-and-setup.md
```diff
@@ -206,6 +206,25 @@ $ jj config set --user user.email "martinvonz@google.com"
...
+(see the recommendation here: https://docs.rs/clap_complete/latest/clap_complete/env/index.html)
+
+Nushell is not yet supported for dynamic completions, I presume we can simply
+continue to emit the normal, static completions for Nushell.
+
+Another option would be to have some kind of beta-phase for the dynamic
+completions, where we tell users to manually add `COMPLETE=fish jj | source`
+to their shell init files if they want to opt into the beta. If there are no
+issues found, we change `jj util completion` to emit the dynamic completion
+hooks instead.
```
senekor, line 226: What is people's opinion on this? should we have a beta phase or is that unnecessary?

cli/src/complete.rs
```diff
@@ -0,0 +1,104 @@
...
+        command.arg(repository);
+    }
+    if let Some(at_operation) = args.at_operation {
+        command.arg("--at-operation");
+        command.arg(at_operation);
+    }
+    for config_toml in args.early_args.config_toml {
+        command.arg("--config-toml");
+        command.arg(config_toml);
+    }
```
senekor, line 101: I believe these should be the only global args relevant for completions. I only added a test for the `--repository` flag for now, since the other two are not relevant for renaming bookmarks.

docs/install-and-setup.md
yuja: I wouldn't change `jj util completion` at least for now. If we fully migrated to dynamic completion, maybe we can just deprecate `jj util completion`? I don't think we'll need a command that prints static a few lines content.

senekor: Nushell isn't supported yet by the dynamic completions and I assume Xonsh wouldn't work either. So I think we should keep the command if we want to continue supporting these shells.

senekor:
All open questions from my side are answered, so marking as ready for review.

cli/src/cli_util.rs
```diff
@@ -3387,6 +3387,7 @@ impl CliRunner {
     #[must_use]
     #[instrument(skip(self))]
     pub fn run(mut self) -> ExitCode {
+        clap_complete::CompleteEnv::with_factory(crate::commands::default_app).complete();
```
yuja, line 3390: nit: `|| self.app.clone()` is probably better. I think it can support custom commands.

cli/src/complete.rs
```diff
@@ -0,0 +1,107 @@
+// Copyright 2020-2023 The Jujutsu Authors
```
yuja, line 1: nit: 2024

cli/src/cli_util.rs
yuja, line 3390: nit: `.try_complete()` and return `ExitCode`? There might be custom resource cleanup after `run()`.

cli/src/complete.rs
```diff
@@ -0,0 +1,107 @@
...
+/// This is necessary because dynamic completion code needs to be aware of
+/// global configuration like custom storage backends. Dynamic completion
+/// code via clap_complete doesn't accept arguments, so they cannot be passed
+/// that way. Another solution would've been to use global mutable state, to
+/// give completion code access to custom backends. Shelling out was chosen as
+/// the preferred method, because it's more maintainable and the performance
+/// requirements of completions aren't very high.
+fn jj() -> std::process::Command {
+    let mut command = std::process::Command::new(
+        std::env::current_exe().unwrap_or(std::path::PathBuf::from("jj")),
```
yuja, line 52: nit: it's probably better to print error message to stderr instead of assuming the executable name is `jj`.

cli/src/complete.rs
```diff
@@ -0,0 +1,107 @@
...
+    command.arg("--ignore-working-copy");
+
+    // Parse some of the global args we care about for passing along to the
+    // child process. This shouldn't fail, since none of the global args are
+    // required.
+    let app = crate::commands::default_app();
+    let config = config::Config::builder()
+        .add_source(default_config())
+        .build()
+        .unwrap();
```
yuja, line 66: nit: better to print error and return empty completion instead of panicking?
(applies to other `.unwrap()`s)

cli/src/complete.rs
```diff
@@ -0,0 +1,107 @@
...
+    let cwd = std::env::current_dir()
+        .and_then(|cwd| cwd.canonicalize())
+        .unwrap();
+    let maybe_cwd_workspace_loader = DefaultWorkspaceLoaderFactory.create(find_workspace_dir(&cwd));
+    layered_configs.read_user_config().unwrap();
+    if let Ok(loader) = &maybe_cwd_workspace_loader {
+        layered_configs
+            .read_repo_config(loader.repo_path())
+            .unwrap();
+    }
```
yuja, line 78: This can't handle custom workspace implementation, but that's probably okay. This only means that aliases stored in repository configuration wouldn't be expanded.

(No action needed)

cli/src/complete.rs
```diff
@@ -0,0 +1,107 @@
...
+    let config = layered_configs.merge();
+    // skip 2 because of the clap_complete prelude: jj -- jj <actual args...>
+    let args = std::env::args().skip(2).collect();
+    let args = resolve_default_command(&ui, &config, &app, args).unwrap();
+    let args = resolve_aliases(&ui, &config, &app, args).unwrap();
```
yuja, line 83: nit: make expand_args() generic instead of making these functions public?

cli/src/complete.rs
```diff
@@ -0,0 +1,107 @@
...
+    let args = app
+        .clone()
+        .disable_version_flag(true)
+        .disable_help_flag(true)
+        .disable_help_subcommand(true)
```
yuja, line 88: nit: `disable_help_subcommand(true)` can be removed, which is now the default.

cli/src/complete.rs
```diff
@@ -0,0 +1,107 @@
...
+        .ignore_errors(true)
+        .get_matches_from(args);
```
yuja, line 90: nit: `try_get_matches_from()`?

yuja (commented):
Looks good to me, thanks!

Do we want to include this in the November release?

cli/src/complete.rs
senekor: I guess it depends on the shell, but `fish` swallows `stderr` of completion code _unless it panics_. So, if we want users to see error messages (such that they can report it to us), we would have to intentionally crash in case of an error.

yuja: Perhaps, we can use `Result` type internally, and transform it to empty result (+ error message) or panic (in case we need to exit with non-zero status.)

senekor: Yes, I was just about to code something like that. What would be a situation where we need to exit with a non-zero exit status in the completion code? Not panicking does have the benefit of running custom cleanup (together with the above suggestion to use `try_complete`).

senekor:
> Do we want to include this in the November release?

Good question. I think in this state, no. It's not yet useful. I'd like to get at least the low-hanging fruit of the [wishlist](https://github.com/martinvonz/jj/issues/4763) merged before having this in a release.

How much time do we have?

senekor:
Oh I just noticed the releases are pretty regular at the beginning of each month. So let's hold off until the December release, that should be fine.

bnjmnt4n:
> Also, they shell out to `rustc` and `rustup` for target triples. Maybe it would be an option for us as well, to simply shell out to ourselves to generate completion candidates? It will be less efficient, but we could remove the yucky global state. The custom completion code could become quite simple, e.g. calling `jj bookmark list -T 'if(!remote, name ++ "\n")'` for branches. On the other hand, we lose type safety. If there are breaking changes to the CLI, this will break without any help from the compiler.

I guess this might be convenient in terms of implementation, but it might make certain things slightly trickier. E.g. if I want to get a list of change IDs with revision names, we'd need to make sure we output the data from `jj` with an appropriate template, then parse it back before providing it to clap.

senekor:
True. Do you have something in mind that's not easily achievable with templates at the moment? Revisions and branches are pretty flexible. I haven't looked into files and operations yet, you mentioned those in your feature list.

CHANGELOG.md
PhilipMetzger, line 1: Nit: Adhere to our commit style here.

docs/install-and-setup.md
```diff
@@ -210,12 +210,25 @@ To set up command-line completion, source the output of
...
 ### Bash
 
 ```shell
 source <(jj util completion bash)
 ```
 
+dynamic:
+
+```shell
+echo "source <(COMPLETE=bash your_program)" >> ~/.bashrc
```
martinvonz, line 229: "your_program" seems to be a placeholder? Supposed to be "jj"?

docs/install-and-setup.md
```diff
@@ -210,12 +210,25 @@ To set up command-line completion, source the output of
 `jj util completion bash/zsh/fish`. Exactly how to source it
 depends on your shell.
 
+Improved completions are currently in the works, these will complete things
+like branch names as well. You can activate them with the alternative "dynamic"
```
martinvonz, line 214: "branch" -> "bookmark"

CHANGELOG.md
PhilipMetzger: yes, although the second part could just be "... teach rename about local bookmarks"

senekor: great :+1:

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Default operation log template now shows end times of operations instead of
   start times.
 
+* The minimum supported Rust version (MSRV) is now 1.80.0.
```
yuja: nit: this entry should be relocated by the first commit, not by the next commit

cli/src/complete.rs
```diff
@@ -0,0 +1,148 @@
...
+            Ok(jj) => jj,
+            Err((ctx, err)) => {
+                eprintln!("{ctx}:");
+                eprintln!("{err}");
+                return Vec::new();
+            }
+        };
+        jj_cmd
+    }};
+}
```
yuja, line 43: I prefer a normal function instead of macro if possible. Something like:
```rust
with_jj(|ui, cmd| {
    cmd.arg(..)...output()?...
})
```

cli/src/complete.rs
```diff
@@ -0,0 +1,148 @@
...
+    for config_toml in args.early_args.config_toml {
+        command.arg("--config-toml");
+        command.arg(config_toml);
+    }
+
+    Ok(command)
+}
+
+// a small hand-rolled error handling utility
+type Error = (&'static str, Box<dyn std::error::Error>);
```
yuja, line 129: nit: might be simpler to just use `CommandError` and `handle_command_result()`

.github/workflows/build.yml
PhilipMetzger, line 1: Nit: "branch" => "bookmark" in the commit message.

ilyagr:
Very much optional: if we think it's likely we'll merge this before the next release, I think it might make sense to merge the MSRV update now. This is so that we have more time for people to adjust to it and resolve any minor issues (and find out about any major issues, which I think and hope are quite unlikely).

But if you prefer not to for some reason, that's fine too.

senekor:
I forgot to actually update the PR here, but I created an issue upstream (clap's MSRV is currently at 1.74) and it was fixed. So there's actually no need anymore for the MSRV bump. I removed the commit.

cli/src/complete.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::config::LayeredConfigs;
+use crate::ui::Ui;
+
+pub fn local_bookmarks() -> Vec<CompletionCandidate> {
+    with_jj(|mut jj| {
+        jj.arg("bookmark")
+            .arg("list")
+            .arg("--template")
+            .arg(r#"if(!remote, name ++ "\n")"#)
+            .output()
```
yuja, line 35: nit: propagate error for consistency

cli/src/complete.rs
```diff
@@ -0,0 +1,124 @@
...
+    command.arg("--ignore-working-copy");
+
+    // Parse some of the global args we care about for passing along to the
+    // child process. This shouldn't fail, since none of the global args are
+    // required.
+    let app = crate::commands::default_app();
+    let config = config::Config::builder()
+        .add_source(default_config())
+        .build()
+        .map_err(internal_error)?;
```
yuja, line 84: nit: `.unwrap()` or `.expect()` here. The default config should be valid.

cli/src/complete.rs
```diff
@@ -0,0 +1,124 @@
...
+    let mut layered_configs = LayeredConfigs::from_environment(config);
+    let ui = Ui::with_config(&layered_configs.merge())?;
```
yuja, line 86: nit: same here. `CliRunner::run()` expects the default config is valid.

cli/src/complete.rs
```diff
@@ -0,0 +1,124 @@
...
+    let cwd = std::env::current_dir()
+        .and_then(|cwd| cwd.canonicalize())
+        .map_err(internal_error)?;
```
yuja, line 89: nit: `user_error()` seems more appropriate (because it's an environment error.) Applies to other `internal_error()`s.

cli/src/complete.rs
```diff
@@ -0,0 +1,124 @@
...
+    // skip 2 because of the clap_complete prelude: jj -- jj <actual args...>
+    let args = std::env::args_os().skip(2);
+    let args = expand_args(&ui, &app, args, &config)?;
+    let args = app
+        .clone()
+        .disable_version_flag(true)
+        .disable_help_flag(true)
+        .ignore_errors(true)
+        .try_get_matches_from(args)
+        .map_err(internal_error)?;
```
yuja, line 107: nit: just `?` to propagate clap::Error?

cli/src/complete.rs
```diff
@@ -0,0 +1,124 @@
...
+    let args: GlobalArgs = GlobalArgs::from_arg_matches(&args).map_err(internal_error)?;
```
yuja, line 108: nit: same here. `?`

yuja (approved):
LGTM, thanks!

cli/src/complete.rs
senekor: I turned a few internal errors into user errors, but left some that I think the users shouldn'b be able to trigger. Please take another look if it's correct now.

cli/src/complete.rs
```diff
@@ -0,0 +1,122 @@
...
+pub fn local_bookmarks() -> Vec<CompletionCandidate> {
+    with_jj(|mut jj| {
+        jj.arg("bookmark")
+            .arg("list")
+            .arg("--template")
+            .arg(r#"if(!remote, name ++ "\n")"#)
+            .output()
+            .map(|output| output.stdout)
+            .map_err(user_error)
+            .and_then(|stdout| String::from_utf8(stdout).map_err(internal_error))
```
yuja, line 39: nit: I don't know if this should be user/internal error, but maybe we can just use `from_utf8_lossy()`? I think it's better for completion.

btw, do you prefer method chaining over `?`? I would split this expression at `.output()`.

cli/src/complete.rs
```diff
@@ -0,0 +1,122 @@
...
+/// the preferred method, because it's more maintainable and the performance
+/// requirements of completions aren't very high.
+fn get_jj_command() -> Result<std::process::Command, CommandError> {
+    let current_exe = std::env::current_exe().map_err(user_error)?;
+    let mut command = std::process::Command::new(current_exe);
+
+    // Snapshotting could make completions much slower in some situations
+    // and be undesired by the user.
+    command.arg("--ignore-working-copy");
+
```
yuja, line 75: nit: it's probably safer to add `--color=never` and `--no-pager`. `ui.color=always` can be set by user config.

yuja (approved):
Thanks.

cli/src/complete.rs
senekor: Splitting the expression does look neater in this case :+1:

docs/install-and-setup.md
```diff
@@ -210,12 +210,25 @@ To set up command-line completion, source the output of
...
 ### Bash
 
 ```shell
 source <(jj util completion bash)
 ```
 
+dynamic:
+
+```shell
+echo "source <(COMPLETE=bash jj)" >> ~/.bashrc
```
martinvonz, line 229: Is there a reason the user would want non-dynamic completions only in the current shell but dynamic completions in future shells too by updating their shell config? If there wasn't a real reason for that, I think we should align the two. I don't feel strongly about which form we should choose. Updating the shell config is probably what most users want. Do we need to worry about users running this line multiple times and ending up with a cluttered config?

docs/install-and-setup.md
```diff
@@ -224,21 +237,37 @@ compinit
...
 ### Fish
 
 ```shell
 jj util completion fish | source
 ```
 
+dynamic:
+
+```shell
+echo "source (COMPLETE=fish jj | psub)" >> ~/.config/fish/config.fish
```
ilyagr, line 255: In `fish`, `COMPLETE=fish jj | source` is more idiomatic than using `psub`.

I was going to submit a PR, but following Martin's comment, perhaps it's worth collecting them all in one place.

docs/install-and-setup.md
ilyagr: Another thing that it'd be nice to make clearer is whether the user wants both (which would be my guess) or just the dynamic one.

Assuming it's both, I'd do something like:

> Add these lines to `~/.bashrc`:
>
> ```shell
> source <(jj util completion bash)
> source <(COMPLETE=bash jj)  # If you also want dynamic completions
> ```

I think adding the line about adding these to .bashrc is a good enough and less noisy substitute compared to including `echo ... > ~/.bashrc`, but there's also an argument of having a copy-pastable command; either is OK by me. It's more important to me that the two are aligned.

martinvonz: > Another thing that it'd be nice to make clearer is whether the user wants both (which would be my guess) or just the dynamic one.

That's what I thought too, but the instructions above say this (emphasis mine):

> You can activate them with the *alternative* "dynamic" instructions below.

@senekor, is there any reason to use the static version? Is it just that we're not confident enough in the dynamic ones yet?

docs/install-and-setup.md
ilyagr: (I'm still happy to do a PR if that's easier)

docs/install-and-setup.md
yuja: I think they're just copied from https://docs.rs/clap_complete/latest/clap_complete/env/index.html

Dropping `echo ".." >> ..` makes sense.

senekor: I opened a [PR](https://github.com/martinvonz/jj/pull/4847/files) to address this.

I did indeed copy these without thinking too much. People may have different preferences about organizing their shell configs. When I used to use zsh, I had the init script configured as `~/.config/zsh/.zshrc` or something, to declutter my home directory. For fish I like to split up my config in many small files. So I agree with removing that part.

> Another thing that it'd be nice to make clearer is whether the user wants both (which would be my guess) or just the dynamic one.

The dynamic completions should already do all the things the static ones did. So there's no need to have both. I added a sentence in the PR mentioning that.

> is there any reason to use the static version? Is it just that we're not confident enough in the dynamic ones yet?

Exactly. Because the dynamic completions are supposed to _replace_ the static ones, I'm not entirely sure if there are some regressions which don't apply to my workflow.

docs/install-and-setup.md
senekor: I also addressed this here: https://github.com/martinvonz/jj/pull/4847

docs/install-and-setup.md
yuja: > we're not confident enough in the dynamic ones yet?

I tried dynamic completion on zsh, and noticed that `jj <TAB>` only lists "log". Apparently, zsh completion script relies on `_CLAP_COMPLETE_INDEX=$(expr $CURRENT - 1)`, which would get out of sync after alias/default-command substitution?

martinvonz: Thanks for checking. Sounds like we should not recommend only dynamic completions yet then.

PR #4740, opened 2024-10-30
cli: make some visible aliases hidden

martinvonz:
# Checklist

If applicable:


CHANGELOG.md
ilyagr, line 1: The commit description is out of date about `jj st`.

ilyagr (approved):
I'm approving this, but I also think we should perhaps mention these aliases and, especially, the `b` alias in the help text. Then, it will be prominent without interfering with shell completion.

This can be a separate PR, though.

CHANGELOG.md
martinvonz: I actually meant to leave `jj st` as a visible alias. I've updated the code and changelog instead.

cli/src/commands/commit.rs
```diff
@@ -28,7 +28,7 @@ use crate::ui::Ui;
 
 /// Update the description and create a new change on top.
 #[derive(clap::Args, Clone, Debug)]
-#[command(visible_aliases=&["ci"])]
+#[command(alias = "ci")]
```
yuja, line 31: nit: maybe we don't have to hide `ci` because it doesn't conflict with any other commands?

yuja (approved):

cli/src/commands/commit.rs
martinvonz: True. I feel like we should still hide it, move it to config, or remove it completely as we've done with many other aliases. But that would be for different reasons, so should be a separate PR.

PhilipMetzger:
As a heavy user of `jj desc` I'm not the biggest fan from hiding from completion, cc @honglooker who added it.

honglooker:
What does `desc` conflict with?

martinvonz:
> What does `desc` conflict with?

`description`

martinvonz:
> As a heavy user of `jj desc` I'm not the biggest fan from hiding from completion, cc @honglooker who added it.

Why do you want it in completion? Do you mean as a way of learning that the alias exists?

ilyagr:
Repeating myself a bit, I think the best of all worlds is to use hidden aliases and manually add some of the aliases to the help text. For example,

```diff
diff --git a/cli/src/commands/status.rs b/cli/src/commands/status.rs
index 98d1090fcd..4008794a8e 100644
--- a/cli/src/commands/status.rs
+++ b/cli/src/commands/status.rs
@@ -26,7 +26,7 @@
 use crate::diff_util::DiffFormat;
 use crate::ui::Ui;
 
-/// Show high-level repo status
+/// Show high-level repo status (alias: st)
 ///
 /// This includes:
 ///
@@ -34,7 +34,8 @@
 ///    changes between them
 ///  * Conflicted bookmarks (see https://martinvonz.github.io/jj/latest/bookmarks/)
 #[derive(clap::Args, Clone, Debug)]
-#[command(visible_alias = "st")]
+// Use hidden aliases so that they do not affect completion
+#[command(alias = "st")]
 pub(crate) struct StatusArgs {
     /// Restrict the status display to these paths
     #[arg(value_hint = clap::ValueHint::AnyPath)]
```

PhilipMetzger:
> Why do you want it in completion?

I basically got used to `jj de<tab>` for `desc` and have been using `jj st` since I downloaded `jj` almost two years ago, so its definitely a loss for me. I'll probably revert to using `describe` with that workflow now.  

> Do you mean as a way of learning that the alias exists?

No, just of out convenience. All-in all I think hiding them is not solving the correct problem and most of them should've been in the default config instead with a special mention in the documentation as @ilyagr suggests.

PhilipMetzger:
And with dynamic completions, we should be able to teach `jj` about user-configured aliases, which is a better solution in my book.

honglooker:
No objection from me.

martinvonz:
> I basically got used to `jj de<tab>` for `desc`

That will still work. It's just that it will complete to `jj describe ` (with a trailing space) instead of completing to `jj desc`.

> have been using `jj st` since I downloaded `jj` almost two years ago, so its definitely a loss for me.

This PR is about making aliases *hidden*, not about removing them. So you can continue to use `jj desc` and `jj st` etc.

CHANGELOG.md
martinvonz: Actually, I ended up making both `st` and `desc` hidden in clap's sense and instead added it to the help text so the `jj help` text is the same, but the `jj describe/status --help` output now also mentions the name of the alias. Maybe we should do that consistently for all commands (e.g. `jj b l`).

ilyagr: SGTM (obviously, given the other things I said here).

PR #3142, opened 2024-02-25
Add `jj sign` command

julienvincent:
This is a continuation from #3007 and #2728 which adds in the `jj sign` command that was split out of the original PR.

---

Some outstanding comments I am carrying over from #3007:

- [ ] https://github.com/martinvonz/jj/pull/3007#discussion_r1492155652
- [ ] https://github.com/martinvonz/jj/pull/3007#discussion_r1492161844
- [ ] https://github.com/martinvonz/jj/pull/3007#discussion_r1492180855

# Checklist

- [x] I have updated `CHANGELOG.md`
- [x] I have updated the documentation (README.md, docs/, demos/)
- [x] I have updated the config schema (cli/src/config-schema.json)
- [x] I have added tests to cover my changes

PR #4747, opened 2024-11-01
Implement `jj sign` and `jj unsign` commands

### LINKED ISSUE #4712: FR: Add a command for signing commits (when `signing.sign-all = false`)

We have had support for automatically signing all commits since #3007.

We want a `jj sign` command for when `signing.sign-all = false`.

None.

**Additional context**

This ticket was extracted from #58 so we could close that one and not make people think that we don't have any support for signing at all.

pylbrecht:
Closes #4712

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,7 +91,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Commands that used to suggest `--ignore-immutable` now print the number of
   immutable commits that would be rewritten if used and a link to the docs.
 
-* The new `jj sign` command allows signing commits.
+* The new `jj sign` and `jj unsign` commands allow for signing/unsigning commits.
 
 ### Fixed bugs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Commands that used to suggest `--ignore-immutable` now print the number of
   immutable commits that would be rewritten if used and a link to the docs.
 
+* The new `jj sign` command allows signing commits.
+
 ### Fixed bugs
 
 * `jj status` now shows untracked files under untracked directories.
--- a/cli/src/commands/sign.rs
+++ b/cli/src/commands/sign.rs
@@ -1,0 +1,130 @@
+// Copyright 2023 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap_complete::ArgValueCandidates;
+use indexmap::IndexSet;
+use itertools::Itertools;
+use jj_lib::commit::Commit;
+use jj_lib::commit::CommitIteratorExt;
+use jj_lib::repo::Repo;
+use jj_lib::signing::SignBehavior;
+
+use crate::cli_util::CommandHelper;
+use crate::cli_util::RevisionArg;
+use crate::command_error::user_error_with_hint;
+use crate::command_error::CommandError;
+use crate::complete;
+use crate::ui::Ui;
+
+/// Cryptographically sign a revision
+#[derive(clap::Args, Clone, Debug)]
+pub struct SignArgs {
+    /// What revision(s) to sign
+    ///
+    /// Note that revisions are always re-signed.
+    ///
+    /// While that leads to discomfort for users, which sign with hardware
+    /// devices, as of now we cannot reliably check if a commit is already
+    /// signed by the user without creating a signature (see https://github.com/jj-vcs/jj/issues/5786).
+    #[arg(
+        long, short,
+        value_name = "REVSETS",
+        add = ArgValueCandidates::new(complete::mutable_revisions),
+    )]
+    revisions: Vec<RevisionArg>,
+}
+
+pub fn cmd_sign(ui: &mut Ui, command: &CommandHelper, args: &SignArgs) -> Result<(), CommandError> {
+    let mut workspace_command = command.workspace_helper(ui)?;
+
+    if !workspace_command.repo().store().signer().can_sign() {
+        return Err(user_error_with_hint(
+            "No signing backend configured",
+            "For configuring a signing backend, see https://jj-vcs.github.io/jj/latest/config/#commit-signing",
+        ));
+    }
+
+    let to_sign: IndexSet<Commit> = workspace_command
+        .parse_union_revsets(ui, &args.revisions)?
+        .evaluate_to_commits()?
+        .try_collect()?;
+
+    workspace_command.check_rewritable(to_sign.iter().ids())?;
+
+    let mut tx = workspace_command.start_transaction();
+
+    let mut signed_commits = vec![];
+    let mut num_reparented = 0;
+
+    tx.repo_mut().transform_descendants(
+        to_sign.iter().ids().cloned().collect_vec(),
+        |rewriter| {
+            let old_commit = rewriter.old_commit().clone();
+            let commit_builder = rewriter.reparent();
+
+            if to_sign.contains(&old_commit) {
+                let new_commit = commit_builder
+                    .set_sign_behavior(SignBehavior::Force)
+                    .write()?;
+
+                signed_commits.push(new_commit);
+            } else {
+                commit_builder.write()?;
+                num_reparented += 1;
+            }
+
+            Ok(())
+        },
+    )?;
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        if !signed_commits.is_empty() {
+            let template = tx.commit_summary_template();
+            writeln!(formatter, "Signed {} commits:", signed_commits.len())?;
+            for commit in &signed_commits {
+                write!(formatter, "  ")?;
+                template.format(commit, formatter.as_mut())?;
+                writeln!(formatter)?;
+            }
+        }
+    }
+
+    let num_not_authored_by_me = signed_commits
+        .iter()
+        .filter(|commit| commit.author().email != tx.settings().user_email())
+        .count();
+    if num_not_authored_by_me > 0 {
+        writeln!(
+            ui.warning_default(),
+            "{num_not_authored_by_me} of these commits are not authored by you",
+        )?;
+    }
+
+    if num_reparented > 0 {
+        writeln!(ui.status(), "Rebased {num_reparented} descendant commits")?;
+    }
+
+    let transaction_description = match &*signed_commits {
+        [] => "".to_string(),
+        [commit] => format!("sign commit {}", commit.id()),
+        commits => format!(
+            "sign commit {} and {} more",
+            commits[0].id(),
+            commits.len() - 1
+        ),
+    };
+    tx.finish(ui, transaction_description)?;
+
+    Ok(())
+}
--- a/cli/src/commands/unsign.rs
+++ b/cli/src/commands/unsign.rs
@@ -1,0 +1,123 @@
+// Copyright 2023 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap_complete::ArgValueCandidates;
+use indexmap::IndexSet;
+use itertools::Itertools;
+use jj_lib::commit::Commit;
+use jj_lib::commit::CommitIteratorExt;
+use jj_lib::signing::SignBehavior;
+
+use crate::cli_util::CommandHelper;
+use crate::cli_util::RevisionArg;
+use crate::command_error::CommandError;
+use crate::complete;
+use crate::ui::Ui;
+
+/// Drop a cryptographic signature
+#[derive(clap::Args, Clone, Debug)]
+pub struct UnsignArgs {
+    /// What revision(s) to unsign
+    #[arg(
+        long, short,
+        value_name = "REVSETS",
+        add = ArgValueCandidates::new(complete::mutable_revisions),
+    )]
+    revisions: Vec<RevisionArg>,
+}
+
+pub fn cmd_unsign(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &UnsignArgs,
+) -> Result<(), CommandError> {
+    let mut workspace_command = command.workspace_helper(ui)?;
+
+    let commits: IndexSet<Commit> = workspace_command
+        .parse_union_revsets(ui, &args.revisions)?
+        .evaluate_to_commits()?
+        .try_collect()?;
+
+    workspace_command.check_rewritable(commits.iter().ids())?;
+
+    let to_unsign: IndexSet<Commit> = commits
+        .into_iter()
+        .filter(|commit| commit.is_signed())
+        .collect();
+
+    let mut tx = workspace_command.start_transaction();
+
+    let mut unsigned_commits = vec![];
+    let mut num_reparented = 0;
+
+    tx.repo_mut().transform_descendants(
+        to_unsign.iter().ids().cloned().collect_vec(),
+        |rewriter| {
+            let old_commit = rewriter.old_commit().clone();
+            let commit_builder = rewriter.reparent();
+
+            if to_unsign.contains(&old_commit) {
+                let new_commit = commit_builder
+                    .set_sign_behavior(SignBehavior::Drop)
+                    .write()?;
+
+                unsigned_commits.push(new_commit);
+            } else {
+                commit_builder.write()?;
+                num_reparented += 1;
+            }
+            Ok(())
+        },
+    )?;
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        if !unsigned_commits.is_empty() {
+            let template = tx.commit_summary_template();
+            writeln!(formatter, "Unsigned {} commits:", unsigned_commits.len())?;
+            for commit in &unsigned_commits {
+                write!(formatter, "  ")?;
+                template.format(commit, formatter.as_mut())?;
+                writeln!(formatter)?;
+            }
+        }
+    }
+
+    let num_not_authored_by_me = unsigned_commits
+        .iter()
+        .filter(|commit| commit.author().email != tx.settings().user_email())
+        .count();
+    if num_not_authored_by_me > 0 {
+        writeln!(
+            ui.warning_default(),
+            "{num_not_authored_by_me} of these commits are not authored by you",
+        )?;
+    }
+
+    if num_reparented > 0 {
+        writeln!(ui.status(), "Rebased {num_reparented} descendant commits")?;
+    }
+
+    let transaction_description = match &*unsigned_commits {
+        [] => "".to_string(),
+        [commit] => format!("unsign commit {}", commit.id()),
+        commits => format!(
+            "unsign commit {} and {} more",
+            commits[0].id(),
+            commits.len() - 1
+        ),
+    };
+    tx.finish(ui, transaction_description)?;
+
+    Ok(())
+}
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -79,6 +79,7 @@ This document contains the help content for the `jj` command-line program.
 * [`jj restore`↴](#jj-restore)
 * [`jj root`↴](#jj-root)
 * [`jj show`↴](#jj-show)
+* [`jj sign`↴](#jj-sign)
 * [`jj simplify-parents`↴](#jj-simplify-parents)
 * [`jj sparse`↴](#jj-sparse)
 * [`jj sparse edit`↴](#jj-sparse-edit)
@@ -149,6 +150,7 @@ To get started, see the tutorial at https://jj-vcs.github.io/jj/latest/tutorial/
 * `restore` — Restore paths from another revision
 * `root` — Show the current workspace root directory
 * `show` — Show commit description and changes in a revision
+* `sign` — Cryptographically sign a revision
 * `simplify-parents` — Simplify parent edges for the specified revision(s)
 * `sparse` — Manage which paths from the working-copy commit are present in the working copy
 * `split` — Split a revision in two
@@ -2210,6 +2212,22 @@ Show commit description and changes in a revision
 * `--context <CONTEXT>` — Number of lines of context to show
 * `--ignore-all-space` — Ignore whitespace when comparing lines
 * `--ignore-space-change` — Ignore changes in amount of whitespace when comparing lines
+
+
+
+## `jj sign`
+
+Cryptographically sign a revision
+
+**Usage:** `jj sign [OPTIONS]`
+
+###### **Options:**
+
+* `-r`, `--revisions <REVSETS>` — What revision(s) to sign
+
+   Note that revisions are always re-signed.
+
+   While that leads to discomfort for users, which sign with hardware devices, as of now we cannot reliably check if a commit is already signed by the user without creating a signature (see https://github.com/jj-vcs/jj/issues/5786).
 
 
 
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -99,6 +99,7 @@ This document contains the help content for the `jj` command-line program.
 * [`jj util install-man-pages`↴](#jj-util-install-man-pages)
 * [`jj util markdown-help`↴](#jj-util-markdown-help)
 * [`jj undo`↴](#jj-undo)
+* [`jj unsign`↴](#jj-unsign)
 * [`jj version`↴](#jj-version)
 * [`jj workspace`↴](#jj-workspace)
 * [`jj workspace add`↴](#jj-workspace-add)
@@ -159,6 +160,7 @@ To get started, see the tutorial at https://jj-vcs.github.io/jj/latest/tutorial/
 * `tag` — Manage tags
 * `util` — Infrequently used commands such as for generating shell completions
 * `undo` — Undo an operation (shortcut for `jj op undo`)
+* `unsign` — Drop a cryptographic signature
 * `version` — Display version information
 * `workspace` — Commands for working with workspaces
 
@@ -2604,6 +2606,18 @@ Undo an operation (shortcut for `jj op undo`)
   - `remote-tracking`:
     The remote-tracking bookmarks. Do not restore these if you'd like to push after the undo
 
+
+
+
+## `jj unsign`
+
+Drop a cryptographic signature
+
+**Usage:** `jj unsign [OPTIONS]`
+
+###### **Options:**
+
+* `-r`, `--revisions <REVSETS>` — What revision(s) to unsign
 
 
 
--- a/cli/tests/test_immutable_commits.rs
+++ b/cli/tests/test_immutable_commits.rs
@@ -538,4 +538,18 @@ fn test_rewrite_immutable_commands() {
     [EOF]
     [exit status: 1]
     "#);
+    // unsign
+    let output = test_env.run_jj_in(&repo_path, &["unsign", "-r=main"]);
+    insta::assert_snapshot!(output, @r#"
+    ------- stderr -------
+    Error: Commit bcab555fc80e is immutable
+    Hint: Could not modify commit: mzvwutvl bcab555f main | (conflict) merge
+    Hint: Immutable commits are used to protect shared history.
+    Hint: For more information, see:
+          - https://jj-vcs.github.io/jj/latest/config/#set-of-immutable-commits
+          - `jj help -k config`, "Set of immutable commits"
+    Hint: This operation would rewrite 1 immutable commits.
+    [EOF]
+    [exit status: 1]
+    "#);
 }
--- a/cli/tests/test_immutable_commits.rs
+++ b/cli/tests/test_immutable_commits.rs
@@ -521,4 +521,21 @@ fn test_rewrite_immutable_commands() {
     [EOF]
     [exit status: 1]
     "##);
+    // sign
+    let output = test_env.run_jj_in(
+        &repo_path,
+        &["sign", "-r=main", "--config=signing.backend=test"],
+    );
+    insta::assert_snapshot!(output, @r#"
+    ------- stderr -------
+    Error: Commit bcab555fc80e is immutable
+    Hint: Could not modify commit: mzvwutvl bcab555f main | (conflict) merge
+    Hint: Immutable commits are used to protect shared history.
+    Hint: For more information, see:
+          - https://jj-vcs.github.io/jj/latest/config/#set-of-immutable-commits
+          - `jj help -k config`, "Set of immutable commits"
+    Hint: This operation would rewrite 1 immutable commits.
+    [EOF]
+    [exit status: 1]
+    "#);
 }
--- a/cli/tests/test_sign_command.rs
+++ b/cli/tests/test_sign_command.rs
@@ -1,0 +1,252 @@
+// Copyright 2023 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::common::TestEnvironment;
+
+#[test]
+fn test_sign() {
+    let test_env = TestEnvironment::default();
+
+    test_env.add_config(
+        r#"
+[ui]
+show-cryptographic-signatures = true
+
+[signing]
+behavior = "keep"
+backend = "test"
+"#,
+    );
+
+    test_env
+        .run_jj_in(test_env.env_root(), ["git", "init", "repo"])
+        .success();
+    let repo_path = test_env.env_root().join("repo");
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "one"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "two"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "three"])
+        .success();
+
+    let output = test_env.run_jj_in(&repo_path, ["log", "-r", "all()"]);
+    insta::assert_snapshot!(output, @r"
+    @  zsuskuln test.user@example.com 2001-02-03 08:05:10 7acb64be
+    │  (empty) (no description set)
+    ○  kkmpptxz test.user@example.com 2001-02-03 08:05:10 8bdfe4fb
+    │  (empty) three
+    ○  rlvkpnrz test.user@example.com 2001-02-03 08:05:09 b0e11728
+    │  (empty) two
+    ○  qpvuntsm test.user@example.com 2001-02-03 08:05:08 876f4b7e
+    │  (empty) one
+    ◆  zzzzzzzz root() 00000000
+    [EOF]
+    ");
+
+    let output = test_env.run_jj_in(&repo_path, ["sign", "-r", "..@"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Signed 4 commits:
+      qpvuntsm 8174ec98 (empty) one
+      rlvkpnrz 6500b275 (empty) two
+      kkmpptxz bcfaa4c3 (empty) three
+      zsuskuln 4947c6dd (empty) (no description set)
+    Working copy now at: zsuskuln 4947c6dd (empty) (no description set)
+    Parent commit      : kkmpptxz bcfaa4c3 (empty) three
+    [EOF]
+    ");
+
+    let output = test_env.run_jj_in(&repo_path, ["log", "-r", "all()"]);
+    insta::assert_snapshot!(output, @r"
+    @  zsuskuln test.user@example.com 2001-02-03 08:05:12 4947c6dd [✓︎]
+    │  (empty) (no description set)
+    ○  kkmpptxz test.user@example.com 2001-02-03 08:05:12 bcfaa4c3 [✓︎]
+    │  (empty) three
+    ○  rlvkpnrz test.user@example.com 2001-02-03 08:05:12 6500b275 [✓︎]
+    │  (empty) two
+    ○  qpvuntsm test.user@example.com 2001-02-03 08:05:12 8174ec98 [✓︎]
+    │  (empty) one
+    ◆  zzzzzzzz root() 00000000
+    [EOF]
+    ");
+
+    // Commits already always signed, even if they are already signed by me.
+    // https://github.com/jj-vcs/jj/issues/5786
+    let output = test_env.run_jj_in(&repo_path, ["sign", "-r", "..@"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Signed 4 commits:
+      qpvuntsm dabebf30 (empty) one
+      rlvkpnrz 2085a464 (empty) two
+      kkmpptxz 227f5e15 (empty) three
+      zsuskuln 15d1b128 (empty) (no description set)
+    Working copy now at: zsuskuln 15d1b128 (empty) (no description set)
+    Parent commit      : kkmpptxz 227f5e15 (empty) three
+    [EOF]
+    ");
+}
+
+#[test]
+fn test_warn_about_signing_commits_not_authored_by_me() {
+    let test_env = TestEnvironment::default();
+
+    test_env.add_config(
+        r#"
+[ui]
+show-cryptographic-signatures = true
+
+[signing]
+behavior = "keep"
+backend = "test"
+"#,
+    );
+
+    test_env
+        .run_jj_in(test_env.env_root(), ["git", "init", "repo"])
+        .success();
+    let repo_path = test_env.env_root().join("repo");
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "one"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "two"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "three"])
+        .success();
+
+    test_env
+        .run_jj_in(
+            &repo_path,
+            &[
+                "desc",
+                "--author",
+                "Someone Else <someone@else.com>",
+                "--no-edit",
+                "..@-",
+            ],
+        )
+        .success();
+    let output = test_env.run_jj_in(&repo_path, ["sign", "-r", "..@-"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Signed 3 commits:
+      qpvuntsm 254d1a64 (empty) one
+      rlvkpnrz caa78d30 (empty) two
+      kkmpptxz c2bc0eb0 (empty) three
+    Warning: 3 of these commits are not authored by you
+    Rebased 1 descendant commits
+    Working copy now at: zsuskuln ede04d15 (empty) (no description set)
+    Parent commit      : kkmpptxz c2bc0eb0 (empty) three
+    [EOF]
+    ");
+
+    let output = test_env.run_jj_in(&repo_path, ["log", "-r", "all()"]);
+    insta::assert_snapshot!(output, @r"
+    @  zsuskuln test.user@example.com 2001-02-03 08:05:12 ede04d15
+    │  (empty) (no description set)
+    ○  kkmpptxz someone@else.com 2001-02-03 08:05:12 c2bc0eb0 [✓︎]
+    │  (empty) three
+    ○  rlvkpnrz someone@else.com 2001-02-03 08:05:12 caa78d30 [✓︎]
+    │  (empty) two
+    ○  qpvuntsm someone@else.com 2001-02-03 08:05:12 254d1a64 [✓︎]
+    │  (empty) one
+    ◆  zzzzzzzz root() 00000000
+    [EOF]
+    ");
+}
+
+#[test]
+fn test_keep_signatures_in_rebased_descendants() {
+    let test_env = TestEnvironment::default();
+
+    test_env.add_config(
+        r#"
+[ui]
+show-cryptographic-signatures = true
+
+[signing]
+behavior = "drop"
+backend = "test"
+"#,
+    );
+
+    test_env
+        .run_jj_in(test_env.env_root(), ["git", "init", "repo"])
+        .success();
+    let repo_path = test_env.env_root().join("repo");
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "A"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "B"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["desc", "-m", "C"])
+        .success();
+
+    let output = test_env.run_jj_in(&repo_path, ["sign", "-r", "@|@--"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Signed 2 commits:
+      qpvuntsm 034b975d (empty) A
+      kkmpptxz 29dc7928 (empty) C
+    Rebased 1 descendant commits
+    Working copy now at: kkmpptxz 29dc7928 (empty) C
+    Parent commit      : rlvkpnrz 014c011c (empty) B
+    [EOF]
+    ");
+
+    let output = test_env.run_jj_in(&repo_path, ["log", "-r", "all()"]);
+    insta::assert_snapshot!(output, @r"
+    @  kkmpptxz test.user@example.com 2001-02-03 08:05:11 29dc7928 [✓︎]
+    │  (empty) C
+    ○  rlvkpnrz test.user@example.com 2001-02-03 08:05:11 014c011c
+    │  (empty) B
+    ○  qpvuntsm test.user@example.com 2001-02-03 08:05:11 034b975d [✓︎]
+    │  (empty) A
+    ◆  zzzzzzzz root() 00000000
+    [EOF]
+    ");
+}
+
+#[test]
+fn test_abort_with_error_if_no_signing_backend_is_configured() {
+    let test_env = TestEnvironment::default();
+
+    test_env.add_config(
+        r#"
+[signing]
+backend = "none"
+"#,
+    );
+
+    test_env
+        .run_jj_in(test_env.env_root(), ["git", "init", "repo"])
+        .success();
+    let repo_path = test_env.env_root().join("repo");
+
+    let output = test_env.run_jj_in(&repo_path, ["sign", "-r", "@-"]);
+
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Error: No signing backend configured
+    Hint: For configuring a signing backend, see https://jj-vcs.github.io/jj/latest/config/#commit-signing
+    [EOF]
+    [exit status: 1]
+    ");
+}
--- a/cli/tests/test_sign_unsign_commands.rs
+++ b/cli/tests/test_sign_unsign_commands.rs
@@ -250,3 +250,149 @@ backend = "none"
     [exit status: 1]
     ");
 }
+
+#[test]
+fn test_unsign() {
+    let test_env = TestEnvironment::default();
+
+    test_env.add_config(
+        r#"
+[ui]
+show-cryptographic-signatures = true
+
+[signing]
+behavior = "keep"
+backend = "test"
+"#,
+    );
+
+    test_env
+        .run_jj_in(test_env.env_root(), ["git", "init", "repo"])
+        .success();
+    let repo_path = test_env.env_root().join("repo");
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "one"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "two"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "three"])
+        .success();
+
+    test_env
+        .run_jj_in(&repo_path, ["sign", "-r", "..@"])
+        .success();
+
+    let output = test_env.run_jj_in(&repo_path, ["log", "-r", "all()"]);
+    insta::assert_snapshot!(output, @r"
+    @  zsuskuln test.user@example.com 2001-02-03 08:05:11 7aa7dcdf [✓︎]
+    │  (empty) (no description set)
+    ○  kkmpptxz test.user@example.com 2001-02-03 08:05:11 0413d103 [✓︎]
+    │  (empty) three
+    ○  rlvkpnrz test.user@example.com 2001-02-03 08:05:11 c8768375 [✓︎]
+    │  (empty) two
+    ○  qpvuntsm test.user@example.com 2001-02-03 08:05:11 b90f5370 [✓︎]
+    │  (empty) one
+    ◆  zzzzzzzz root() 00000000
+    [EOF]
+    ");
+
+    let output = test_env.run_jj_in(&repo_path, ["unsign", "-r", "..@"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Unsigned 4 commits:
+      qpvuntsm cb05440c (empty) one
+      rlvkpnrz deb0db4b (empty) two
+      kkmpptxz 7c11ee12 (empty) three
+      zsuskuln be9daa4d (empty) (no description set)
+    Working copy now at: zsuskuln be9daa4d (empty) (no description set)
+    Parent commit      : kkmpptxz 7c11ee12 (empty) three
+    [EOF]
+    ");
+
+    let output = test_env.run_jj_in(&repo_path, ["log", "-r", "all()"]);
+    insta::assert_snapshot!(output, @r"
+    @  zsuskuln test.user@example.com 2001-02-03 08:05:13 be9daa4d
+    │  (empty) (no description set)
+    ○  kkmpptxz test.user@example.com 2001-02-03 08:05:13 7c11ee12
+    │  (empty) three
+    ○  rlvkpnrz test.user@example.com 2001-02-03 08:05:13 deb0db4b
+    │  (empty) two
+    ○  qpvuntsm test.user@example.com 2001-02-03 08:05:13 cb05440c
+    │  (empty) one
+    ◆  zzzzzzzz root() 00000000
+    [EOF]
+    ");
+}
+
+#[test]
+fn test_warn_about_unsigning_commits_not_authored_by_me() {
+    let test_env = TestEnvironment::default();
+
+    test_env.add_config(
+        r#"
+[ui]
+show-cryptographic-signatures = true
+
+[signing]
+behavior = "keep"
+backend = "test"
+"#,
+    );
+
+    test_env
+        .run_jj_in(test_env.env_root(), ["git", "init", "repo"])
+        .success();
+    let repo_path = test_env.env_root().join("repo");
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "one"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "two"])
+        .success();
+    test_env
+        .run_jj_in(&repo_path, ["commit", "-m", "three"])
+        .success();
+
+    test_env
+        .run_jj_in(&repo_path, ["sign", "-r", "..@"])
+        .success();
+
+    let run_jj_as_someone_else = |args: &[&str]| {
+        let output = test_env.run_jj_with(|cmd| {
+            cmd.current_dir(&repo_path)
+                .env("JJ_USER", "Someone Else")
+                .env("JJ_EMAIL", "someone@else.com")
+                .args(args)
+        });
+        (output.stdout, output.stderr)
+    };
+
+    let (_, stderr) = run_jj_as_someone_else(&["unsign", "-r", "..@"]);
+    insta::assert_snapshot!(stderr, @r"
+    Unsigned 4 commits:
+      qpvuntsm 757aba72 (empty) one
+      rlvkpnrz 49a6eeeb (empty) two
+      kkmpptxz 8859969b (empty) three
+      zsuskuln 8cea2d75 (empty) (no description set)
+    Warning: 4 of these commits are not authored by you
+    Working copy now at: zsuskuln 8cea2d75 (empty) (no description set)
+    Parent commit      : kkmpptxz 8859969b (empty) three
+    [EOF]
+    ");
+
+    let output = test_env.run_jj_in(&repo_path, ["log", "-r", "all()"]);
+    insta::assert_snapshot!(output, @r"
+    @  zsuskuln test.user@example.com 2001-02-03 08:05:12 8cea2d75
+    │  (empty) (no description set)
+    ○  kkmpptxz test.user@example.com 2001-02-03 08:05:12 8859969b
+    │  (empty) three
+    ○  rlvkpnrz test.user@example.com 2001-02-03 08:05:12 49a6eeeb
+    │  (empty) two
+    ○  qpvuntsm test.user@example.com 2001-02-03 08:05:12 757aba72
+    │  (empty) one
+    ◆  zzzzzzzz root() 00000000
+    [EOF]
+    ");
+}
--- a/docs/config.md
+++ b/docs/config.md
@@ -1211,6 +1211,17 @@ key = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIGj+J6N6SO+4P8dOZqfR1oiay2yxhhHnagH52
 sign-on-push = true
 ```
 
+### Manually signing commits
+
+You can use [`jj sign`](./cli-reference.md#jj-sign) to manually sign commits.
+
+
+!!! warning
+
+    `jj sign` is always signing commits, even if they are already signed by the
+    user. While this is cumbersome for users signing via hardware devices, we
+    cannot reliably check if a commit is already signed without creating a
+    signature (see [this issue](https://github.com/jj-vcs/jj/issues/5786)).
 
 ## Commit Signature Verification
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -1213,7 +1213,8 @@ sign-on-push = true
 
 ### Manually signing commits
 
-You can use [`jj sign`](./cli-reference.md#jj-sign) to manually sign commits.
+You can use [`jj sign`](./cli-reference.md#jj-sign)/[`jj unsign`](./cli-reference.md#jj-unsign)
+to sign/unsign commits manually.
 
 
 !!! warning
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -53,6 +53,7 @@ mod split;
 mod squash;
 mod status;
 mod tag;
+mod unsign;
 mod unsquash;
 mod util;
 mod version;
@@ -153,6 +154,7 @@ enum Command {
     Util(util::UtilCommand),
     /// Undo an operation (shortcut for `jj op undo`)
     Undo(operation::undo::OperationUndoArgs),
+    Unsign(unsign::UnsignArgs),
     // TODO: Delete `unsquash` in jj 0.28+
     #[command(hide = true)]
     Unsquash(unsquash::UnsquashArgs),
@@ -228,6 +230,7 @@ pub fn run_command(ui: &mut Ui, command_helper: &CommandHelper) -> Result<(), Co
         Command::Status(args) => status::cmd_status(ui, command_helper, args),
         Command::Tag(args) => tag::cmd_tag(ui, command_helper, args),
         Command::Undo(args) => operation::undo::cmd_op_undo(ui, command_helper, args),
+        Command::Unsign(args) => unsign::cmd_unsign(ui, command_helper, args),
         Command::Unsquash(args) => unsquash::cmd_unsquash(ui, command_helper, args),
         Command::Untrack(args) => {
             let cmd = renamed_cmd("untrack", "file untrack", file::untrack::cmd_file_untrack);
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -46,6 +46,7 @@ mod restore;
 mod root;
 mod run;
 mod show;
+mod sign;
 mod simplify_parents;
 mod sparse;
 mod split;
@@ -139,6 +140,7 @@ enum Command {
     // TODO: Flesh out.
     Run(run::RunArgs),
     Show(show::ShowArgs),
+    Sign(sign::SignArgs),
     SimplifyParents(simplify_parents::SimplifyParentsArgs),
     #[command(subcommand)]
     Sparse(sparse::SparseCommand),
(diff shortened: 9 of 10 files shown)


cli/src/commands/sign.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// Cryptographically sign a revision
+#[derive(clap::Args, Clone, Debug)]
+pub struct SignArgs {
+    /// What key to use, depends on the configured signing backend.
+    #[arg()]
+    key: Option<String>,
+    /// What revision to sign
+    #[arg(long, short, default_value = "@")]
+    revision: RevisionArg,
```
jalil-salame, line 28: Wouldn't it make sense to have this be `revisions: Vec<RevisionArg>`?

This way you can do `jj sign -r main..bookmark` to sign a whole branch.

martinvonz: That can still be supported with a single argument. The whole `main..bookmark` string gets passed as a single argument from your shell to the `jj` binary. The binary then decides how to interpret it, and the `jj sign` command will presumably want to accept multiple revisions, as most commands do. Commands that want a single revision still accept revsets; they just validate that the revset contains exactly one revision.

It may still be a good idea to allow multiple revision *arguments* here, so the user can do `jj sign -r X -r Y -r Z` to sign those three revsets.

jalil-salame: Thanks for the clarification c: and yes `jj sign -r X -r Y -r Z` sounds logical

necauqua: A reminder about an [identical](https://github.com/jj-vcs/jj/pull/3142#discussion_r1503634258) discussion - maybe the issue with rebasing transformations got better since then?

If implemented naively, signing `main..bookmark` is quadratic - signing an ancestor rebases all of it's descendants, and then we sign the child of that ancesrot rebasing all of _its_ descendants and so on

yuja: > maybe the issue with rebasing transformations got better since then?

Yes, we have `.transform_descendants(roots, |rewriter| ..)` API now.

lib/src/commit_builder.rs
```diff
@@ -332,7 +332,7 @@ impl DetachedCommitBuilder {
     }
 
     pub fn set_sign_key(&mut self, sign_key: Option<String>) -> &mut Self {
-        self.sign_settings.key = sign_key;
+        self.sign_settings.key = sign_key.or(self.sign_settings.key.clone());
```
necauqua, line 335:
```suggestion
        if sign_key.is_some() {
            self.sign_settings.key = sign_key;
        }
```

necauqua: ah wait it's a draft with cmd_sign not implemented yet, oh well

pylbrecht:
This is still WIP, but I would love to get some feedback on the general approach and if this has any design flaws that I need to address before polishing.
@yuja, @necauqua, since you have been involved in my last PR, maybe you have some suggestions?
(Of course anyone else is also welcome to leave feedback!)

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+            if !args.drop && commit.is_signed() {
+                return Err(user_error(
+                    "Commit is already signed, use --force to sign anyway",
+                ));
+            }
+            if commit.author().email != command.settings().user_email() {
+                return Err(user_error(
+                    "Commit is not authored by you, use --force to sign anyway",
+                ));
+            }
```
necauqua, line 71: Now that this version can work with multiple commits, those checks should show the offending one.
Or one***s***, actually, so they should eagerly collect lists of bad commits and print them all at once.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+    let behavior = if args.drop {
+        SignBehavior::Drop
+    } else if args.force {
+        SignBehavior::Force
+    } else {
+        SignBehavior::Own
+    };
+    let mut signed_commits = vec![];
+    tx.repo_mut().transform_descendants(
+        commits.iter().ids().cloned().collect_vec(),
```
necauqua, line 86: Use `commit_ids` here?

Hmm, if `check_rewritable` consumes the vector that sounds meh, it should accept `&[CommitId]`.
If it really does need to consume the vector for some reason ignore this comment then 🤷
(Or maybe it accepts any iterable and you just need to borrow there, I haven't looked at the code in a year already)

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+        |rewriter| {
+            if commits.contains(rewriter.old_commit()) {
```
necauqua, line 88: For the unlikely scenario of signing thousands of commits at once should maybe use a set?

If this check is needed at all, I'm not thinking too deep about the underlying logic here, but linearly checking for `commits.contains` in a rebase sounds a bit too quadratic

lib/src/commit_builder.rs
```diff
@@ -332,7 +332,9 @@ impl DetachedCommitBuilder {
     }
 
     pub fn set_sign_key(&mut self, sign_key: Option<String>) -> &mut Self {
-        self.sign_settings.key = sign_key;
+        if sign_key.is_some() {
+            self.sign_settings.key = sign_key;
+        }
         self
     }
```
necauqua, line 339: nit: looking at the overall method now (my previous nit was a mechanical refactor to avoid an unnecessary clone) - `set_sign_key` does nothing if given option is None, which is a bit.. meh, idk - could change it to accept `sign_key: String` directly and make the callers do the `if let Some(..) = ..`?

necauqua (commented):
Actually left a couple of comments.

Dropped signatures hint and builtin templates can be separate PRs, yes.

Will reiterate my idea for the templates here:

There should be a `some_consistent_name_for_crypt_sig` template that's used in the default log template and _by default_ it's empty.
Then there should be `some_other_name` template that's _unused_, but is present by default, with my pretty checkmarks or whatever.
And then in FAQ or in docs, there should be a tip "you can run this jj config command to set that first template to equal that second one so that you'll get checkmarks in all the default templates".

"Or you can set that template to `if(signature, "[•]", "")` to just differentiate signed commits from unsigned in the logs, which is very fast compared to actually verifying them"

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
+// Copyright 2023 The Jujutsu Authors
```
martinvonz, line 1: nit: maybe 2024 or 2025? or maybe this work was started in 2023 so it should be 2023?

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Drop the signature, explicitly "un-signing" the commit.
+    #[arg(long, short = 'D', conflicts_with = "force")]
+    drop: bool,
+}
+
+pub fn cmd_sign(ui: &mut Ui, command: &CommandHelper, args: &SignArgs) -> Result<(), CommandError> {
+    let mut workspace_command = command.workspace_helper(ui)?;
+
+    let commits = workspace_command
+        .resolve_some_revsets_default_single(ui, &args.revisions)?
```
martinvonz, line 53: I think we want `parse_union_revsets()` here. I think it's pretty expected to want to sign multiple commits at once.

cli/tests/test_sign_command.rs
martinvonz, line 1: Make most of these tests sign multiple commits to make sure that that works as expected

cli/src/commands/sign.rs
martinvonz: `commits` is an `IndexSet`, so it should be fast to check for membership in

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+    let Some(mut formatter) = ui.status_formatter() else {
+        return Ok(());
+    };
+    let template = workspace_command.commit_summary_template();
+    for commit in &signed_commits {
+        if args.drop {
+            write!(formatter, "Signature was dropped: ")?;
+        } else {
+            write!(formatter, "Commit was signed: ")?;
+        }
```
martinvonz, line 111: I would expect these messages to depend on the commit's previous signature, especially "Signature was dropped" is confusing for commits that were not signed to start with.

pylbrecht:
Wow, thanks a lot for your feedback, @necauqua and @martinvonz!
This helps a lot. Being new to the codebase (and the VCS domain), I struggle with *what* is expected a fair bit. It's great to receive some guidance here.

Now I am all set to return to my dungeon and work on the *how*.

cli/src/commands/sign.rs
necauqua: ah in my head I've seen the `collect_vec` and assumed `commits` was a vec for some reason, my bad 😅

cli/src/commands/sign.rs
necauqua: nit: Also `if args.force` can be moved outside of the loop

cli/src/commands/sign.rs
necauqua: This made sense with a singular commit, since pre-check would short-circuit the situation where args.drop=true and the commit was not signed

cli/tests/test_sign_command.rs
necauqua: random suggestion, ignore if it doesn't make sense:
Snapshot the evolog to ensure no quadratic rebases that I feared?.

necauqua:
Very offtopic, sorry, but there's no chance in hell I'd have contributed to jj if is was cpp or something.
Even if it was C (that I don't _dislike_ per se) every single project has it's own very specific patterns etc. and the build systems are plentiful.
Here in Rust we have cargo that just works, we have rustfmt, and we have the borrowchecker to not be scared of your code, and to not learn _again_ how the particular project handles memory.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+/// Cryptographically sign a revision
+#[derive(clap::Args, Clone, Debug)]
+pub struct SignArgs {
+    /// What key to use, depends on the configured signing backend.
+    #[arg()]
+    key: Option<String>,
+    /// What revision(s) to sign
+    #[arg(
+        long, short,
+        default_value = "@",
```
yuja, line 36: nit: it might be better to make the default configurable. I think `jj sign` will be used in the context where we do `jj fix && jj sign .. && jj git push`.

(can be addressed separately)

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+        value_name = "REVSETS",
+        add = ArgValueCandidates::new(complete::all_revisions),
+    )]
+    revisions: Vec<RevisionArg>,
+    /// Sign a commit that is not authored by you or was already signed.
+    #[arg(long, short)]
+    force: bool,
```
yuja, line 43: nit: `--force` tends to mean arbitrary things. Maybe `--allow-not-mine` or something is better?

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Drop the signature, explicitly "un-signing" the commit.
+    #[arg(long, short = 'D', conflicts_with = "force")]
+    drop: bool,
```
yuja, line 46: nit: just dropping idea: `jj unsign` or `jj signature drop`. 
(It's unlikely we'll add more sub commands, so namespaced command might be overkill.)

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+        |rewriter| {
+            if commits.contains(rewriter.old_commit()) {
+                let commit_builder = rewriter.rebase()?;
+                let new_commit = commit_builder
+                    .set_sign_key(args.key.clone())
+                    .set_sign_behavior(behavior)
+                    .write()?;
+                signed_commits.push(new_commit);
+            }
+            Ok(())
```
yuja, line 96: maybe you'll need to do `rewriter.reparent()`. `cmd_describe()` does `rewriter.rebase()`, but I think `.reparent()` works there, too.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,117 @@
...
+pub struct SignArgs {
+    /// What key to use, depends on the configured signing backend.
+    #[arg()]
+    key: Option<String>,
+    /// What revision(s) to sign
+    #[arg(
+        long, short,
+        default_value = "@",
+        value_name = "REVSETS",
+        add = ArgValueCandidates::new(complete::all_revisions),
```
yuja, line 38: nit: `complete::mutable_revisions`?

pylbrecht:
Thanks @yuja!

cli/src/commands/sign.rs
pylbrecht: I like `jj unsign`.
Splitting off the `drop` from `jj sign` would make `cmd_sign()` a little simpler, as well.

cli/src/commands/sign.rs
pylbrecht: I `jj duplicate`d the commit, so I suppose the work started in 2023.

pylbrecht:
Things have been a bit hectic at work recently, but I expect to make some progress on this in the next couple of days.

cli/src/commands/sign.rs
pylbrecht: Right now we have two cases, where `--force` can be applied:
* commit is not authored by me
* commit is already signed

Would we also want to introduce another option for already signed commits then (e.g. `--allow-signed`)?
I interpret `--force` here as "sign this commit no matter what" and it feels reasonable to me. However I am not too confident about my judgement here (maybe I am already poisoned by having gotten used to CLIs using `--force` 🤔 ), so I am happy to lean on your opinion/experience.

pylbrecht: Now that I think about it, maybe 
* commit is not authored by me

is the only case we actually care about.
Is there any risk involved in (re)signing a commit that is already signed by me?

jalil-salame: For people using hardware devices for signatures, resigning commits is a pain. So I would prefer it if it would not replace any signatures by default.

Resigning a commit is also strictly more work as you have to rebase all descendants.

yuja: In any case, these two are different requirements, and shouldn't be controlled by a single `--force` flag, I think.

pylbrecht: > For people using hardware devices for signatures, resigning commits is a pain. So I would prefer it if it would not replace any signatures by default.

Thanks a lot for that feedback. That would have been a major annoyance in terms of usability.

cli/src/commands/sign.rs
pylbrecht: Looks like `check_rewritable()` is consuming the iterator:
```rust
    pub fn check_rewritable<'a>(
        &self,
        commits: impl IntoIterator<Item = &'a CommitId>,
    ) -> Result<(), CommandError>
```
But maybe I am missing something?

jalil-salame: If it receives `impl IntoIterator` you can just give it a `vec.iter()` you might need to do `iter().copied()` if you have errors due to double references (`&&Commit`).

cli/src/commands/sign.rs
yuja: Kinda yes. I have no idea whether `revsets.sign` should default to `@` or `reachable(@, mutable())`. I don't think it makes sense to sign the working-copy commit, so the latter might be a better default?

For `unsign`, I wouldn't add a default or configurable revset. I don't expect the user would want to casually remove signatures from a certain set of commits.

In any case, this isn't a blocker. We can add a revset configuration as a follow up patch.

pylbrecht: > Kinda yes. I have no idea whether revsets.sign should default to @ or reachable(@, mutable()). I don't think it makes sense to sign the working-copy commit, so the latter might be a better default?

`reachable(@, mutable())` does feel better than `@`.

> For unsign, I wouldn't add a default or configurable revset. I don't expect the user would want to casually remove signatures from a certain set of commits.

Yeah, that makes more sense.

> In any case, this isn't a blocker. We can add a revset configuration as a follow up patch.

Let me see how much effort it is to include that. If I am starting to block myself, because I have no idea what I am doing, I will address it in a follow-up.

cli/src/commands/sign.rs
pylbrecht: I think using `copied()` is not possible, because `CommitId` does not implement `Copy`:

```diff
diff --git a/cli/src/commands/sign.rs b/cli/src/commands/sign.rs
index 8da754b3b5..94b90033ea 100644
--- a/cli/src/commands/sign.rs
+++ b/cli/src/commands/sign.rs
@@ -81,7 +81,8 @@
         .evaluate_to_commits()?
         .try_collect()?;
 
-    workspace_command.check_rewritable(commits.iter().ids())?;
+    let commit_ids = commits.iter().ids();
+    workspace_command.check_rewritable(commit_ids)?;
 
     if !args.allow_not_mine {
         check_commits_authored_by(&commits, command.settings().user_email())?;
@@ -95,9 +96,8 @@
         SignBehavior::Own
     };
     let mut signed_commits = vec![];
-    tx.repo_mut().transform_descendants(
-        commits.iter().ids().cloned().collect_vec(),
-        |rewriter| {
+    tx.repo_mut()
+        .transform_descendants(commit_ids.copied().collect_vec(), |rewriter| {
             if rewriter.old_commit().is_signed() {
                 // Don't resign commits, which are already signed by me. For people using
                 // hardware devices for signatures, resigning commits is
@@ -114,8 +114,7 @@
                 signed_commits.push(new_commit);
             }
             Ok(())
-        },
-    )?;
+        })?;
 
     tx.finish(ui, format!("signed {} commits", signed_commits.len()))?;
 ```

<details><summary><code>cargo check</code> output</summary>
<p>

```
    Checking jj-cli v0.25.0 (/Users/palbrecht/dev/personal/jj/cli)
error[E0277]: the trait bound `jj_lib::backend::CommitId: std::marker::Copy` is not satisfied
    --> cli/src/commands/sign.rs:100:43
     |
100  |         .transform_descendants(commit_ids.copied().collect_vec(), |rewriter| {
     |                                           ^^^^^^ the trait `std::marker::Copy` is not implemented for `jj_lib::backend::CommitId`
     |
note: required by a bound in `std::iter::Iterator::copied`
    --> /Users/palbrecht/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:3364:12
     |
3361 |     fn copied<'a, T: 'a>(self) -> Copied<Self>
     |        ------ required by a bound in this associated function
...
3364 |         T: Copy,
     |            ^^^^ required by this bound in `Iterator::copied`

error[E0599]: the method `collect_vec` exists for struct `Copied<impl Iterator<Item = &CommitId>>`, but its trait bounds were not satisfied
   --> cli/src/commands/sign.rs:100:52
    |
100 |         .transform_descendants(commit_ids.copied().collect_vec(), |rewriter| {
    |                                                    ^^^^^^^^^^^ method cannot be called on `Copied<impl Iterator<Item = &CommitId>>` due to unsatisfied trait bounds
    |
   ::: /Users/palbrecht/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/adapters/copied.rs:19:1
    |
19  | pub struct Copied<I> {
    | -------------------- doesn't satisfy `_: Iterator` or `_: Itertools`
    |
    = note: the following trait bounds were not satisfied:
            `std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: std::iter::Iterator`
            which is required by `std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: itertools::Itertools`
            `&std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: std::iter::Iterator`
            which is required by `&std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: itertools::Itertools`
            `&mut std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: std::iter::Iterator`
            which is required by `&mut std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: itertools::Itertools`

Some errors have detailed explanations: E0277, E0599.
For more information about an error, try `rustc --explain E0277`.
error: could not compile `jj-cli` (lib) due to 2 previous errors
```

</p>
</details> 

---

How would I handle this idiomatically?

File: lib/src/commit_builder.rs (hunk 6 above)
Comment (pylbrecht) [reply lib/src/commit_builder.rs:339]: [Since `sign_settings.key` is `Option<String>`](https://github.com/pylbrecht/jj/blob/3edbfa34c327a0f7e16f95a1ae0aebe4381f39c6/lib/src/settings.rs#L90), wouldn't we eliminate the case, where the caller wants to explicitly set `sign_settings.key = None`?
Or is this case nonsense and `sign_settings.key` should never be set to `None` after initialization ([`SignSettings::from_settings()`](https://github.com/pylbrecht/jj/blob/3edbfa34c327a0f7e16f95a1ae0aebe4381f39c6/lib/src/settings.rs#L104))?

File: cli/src/commands/sign.rs (hunk 4 above)
Comment (jalil-salame) [reply cli/src/commands/sign.rs:99]: There seems to be a misunderstanding/miscommunication between us, the current way is fine, but you can also do this:

```rust
let commit_ids = commits.iter().ids().copied().collect_vec();
workspace_command.check_rewritable(&commit_ids)?;

// ...

tx.repo_mut().transform_descendants(commit_ids, ...)
```

Is this better? Maybe, it depends:
- The previous way only allocated memory before `transform_descendants`, but iterated over the commits (`commits.iter().ids()`) twice.
- This way allocates memory before `check_rewritable`, but only goes through the commits once.

I don't believe this makes any real difference.

The `copied` comes from my misconception that you might have a double reference: `&&CommitId`, now, references are `Copy`, so `&&T -> &T // always Copy`, but `&T -> T // only Copy/Clone if T is Copy/Clone`.

Either way is idiomatical.

necauqua: Yeah my comment was a nit mostly.
And yup, &T is always Copy, so as [&T].iter() will yield &&T you'd often use [&T].iter().copied() to unwrap that

pylbrecht: All right, thanks for elaborating!
I will leave it as is then.

So this doesn't work, because we do not have a double reference:
```diff
diff --git a/cli/src/commands/sign.rs b/cli/src/commands/sign.rs
index 8da754b3b5..00775aef7a 100644
--- a/cli/src/commands/sign.rs
+++ b/cli/src/commands/sign.rs
@@ -81,7 +81,8 @@
         .evaluate_to_commits()?
         .try_collect()?;
 
-    workspace_command.check_rewritable(commits.iter().ids())?;
+    let commit_ids = commits.iter().ids().copied().collect_vec();
+    workspace_command.check_rewritable(&commit_ids)?;
 
     if !args.allow_not_mine {
         check_commits_authored_by(&commits, command.settings().user_email())?;
```

<details><summary><code>cargo check</code> output</summary>
<p>

```
[Running 'cargo clippy --workspace --all-targets && cargo +1.76 check --workspace --all-targets']
    Checking jj-cli v0.25.0 (/Users/palbrecht/dev/personal/jj/cli)
error[E0277]: the trait bound `jj_lib::backend::CommitId: std::marker::Copy` is not satisfied
    --> cli/src/commands/sign.rs:84:43
     |
84   |     let commit_ids = commits.iter().ids().copied().collect_vec();
     |                                           ^^^^^^ the trait `std::marker::Copy` is not implemented for `jj_lib::backend::CommitId`
     |
note: required by a bound in `std::iter::Iterator::copied`
    --> /Users/palbrecht/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:3364:12
     |
3361 |     fn copied<'a, T: 'a>(self) -> Copied<Self>
     |        ------ required by a bound in this associated function
...
3364 |         T: Copy,
     |            ^^^^ required by this bound in `Iterator::copied`

error[E0599]: the method `collect_vec` exists for struct `Copied<impl Iterator<Item = &CommitId>>`, but its trait bounds were not satisfied
  --> cli/src/commands/sign.rs:84:52
   |
84 |     let commit_ids = commits.iter().ids().copied().collect_vec();
   |                                                    ^^^^^^^^^^^ method cannot be called on `Copied<impl Iterator<Item = &CommitId>>` due to unsatisfied trait bounds
   |
  ::: /Users/palbrecht/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/adapters/copied.rs:19:1
   |
19 | pub struct Copied<I> {
   | -------------------- doesn't satisfy `_: Iterator` or `_: Itertools`
   |
   = note: the following trait bounds were not satisfied:
           `std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: std::iter::Iterator`
           which is required by `std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: itertools::Itertools`
           `&std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: std::iter::Iterator`
           which is required by `&std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: itertools::Itertools`
           `&mut std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: std::iter::Iterator`
           which is required by `&mut std::iter::Copied<impl std::iter::Iterator<Item = &jj_lib::backend::CommitId>>: itertools::Itertools`

Some errors have detailed explanations: E0277, E0599.
For more information about an error, try `rustc --explain E0277`.
error: could not compile `jj-cli` (lib) due to 2 previous errors
```

</p>
</details>

docs/config.md
```diff
@@ -1095,6 +1095,10 @@ key = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIGj+J6N6SO+4P8dOZqfR1oiay2yxhhHnagH52
 sign-on-push = true
 ```
 
+### Manually signing commits
+
+You can use [`jj sign`](./cli-reference.md#jj-sign)/[`jj
+unsign`](./cli-reference.md#jj-unsign)to sign/unsign commits manually.
```
pylbrecht, line 1101: @yuja, we can extend this section with an example for the workflow you mentioned in https://github.com/jj-vcs/jj/pull/4747#discussion_r1904854027, when making `-r` configurable.

CHANGELOG.md
```diff
@@ -116,6 +116,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New `subject(pattern)` revset function that matches first line of commit
   descriptions.
 
+* The new `jj sign` command allows signing commits.
+
+* The new `jj unsign` command allows unsigning commits.
+
```
pylbrecht, line 122: These do feel a bit scarce, but maybe that's ok? I am open to suggestions of things to add here.

cli/src/commands/sign.rs
pylbrecht: For the sake of getting this merged rather sooner than later, I will address that in #5428.

lib/src/commit_builder.rs
pylbrecht: @necauqua, I reverted the change as I honestly do not recall why I did it in the first place. 🤔 
Let me know if you have any objections.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Cryptographically sign a revision
+#[derive(clap::Args, Clone, Debug)]
+pub struct SignArgs {
+    /// What key to use, depends on the configured signing backend.
+    #[arg()]
+    key: Option<String>,
+    /// What revision(s) to sign
+    #[arg(
+        long, short,
+        default_value = "@",
```
yuja, line 39: nit: omit the default value for now? I don't think the current default is useful. (the same for `unsign`)

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,161 @@
...
+        value_name = "REVSETS",
+        add = ArgValueCandidates::new(complete::mutable_revisions),
+    )]
+    revisions: Vec<RevisionArg>,
+    /// Sign a commit that is not authored by you.
```
yuja, line 44: nit: "Sign revisions that are .."?

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+pub fn check_commits_authored_by<'a>(
+    commits: impl IntoIterator<Item = &'a Commit>,
+    user_email: &str,
+    hint: &str,
+) -> Result<(), CommandError> {
+    let bad_commits = commits
+        .into_iter()
+        .filter(|commit| commit.author().email != user_email)
```
yuja, line 56: nit: Can't we use `SignSettings::should_sign()`?

BTW, I'm not sure if this should be an error. If a warning is good enough, warnings can be emitted within/after `transform_descendants()`.

@necauqua any thoughts?

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,161 @@
...
+        |rewriter| {
+            if rewriter.old_commit().is_signed() {
+                // Don't resign commits, which are already signed by me. For people using
+                // hardware devices for signatures, resigning commits is
+                // cumbersome.
+                return Ok(());
+            }
+
+            if commits.contains(rewriter.old_commit()) {
+                let commit_builder = rewriter.reparent();
```
yuja, line 119: We'll probably need to do always `.reparent()`. See `cmd_describe()` for example.

`.reparent()` can be skipped when `!rewriter.parents_changed() && no_need_to_sign_the_current_commit`.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,161 @@
...
+        },
+    )?;
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        match signed_commits.len() {
+            0 => (),
+            1 => {
+                write!(formatter, "Signed commit ")?;
+                tx.base_workspace_helper()
+                    .write_commit_summary(formatter.as_mut(), &signed_commits[0])?;
```
yuja, line 136: use `tx.write_commit_summary()` and `tx.commit_summary_template()` because `signed_commits[N]` is rewritten commits. (also applies to unsign)

cli/tests/cli-reference@.md.snap
```diff
@@ -1,6 +1,7 @@
 ---
 source: cli/tests/test_generate_md_cli_help.rs
 description: "AUTO-GENERATED FILE, DO NOT EDIT. This cli reference is generated by a test as an `insta` snapshot. MkDocs includes this snapshot from docs/cli-reference.md."
+snapshot_kind: text
```
yuja, line 4: nit: you'll need to upgrade `cargo-insta`

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,161 @@
...
+                signed_commits.push(new_commit);
+            }
+            Ok(())
+        },
+    )?;
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        match signed_commits.len() {
+            0 => (),
+            1 => {
```
yuja, line 133: nit: you can also use slice pattern
```rust
match &*signed_commits {
    [] => {}
    [commit] => ...
    commits => ...
}
```

cli/src/commands/unsign.rs
```diff
@@ -0,0 +1,116 @@
...
+    #[arg(
+        long, short,
+        default_value = "@",
+        value_name = "REVSETS",
+        add = ArgValueCandidates::new(complete::mutable_revisions),
+    )]
+    revisions: Vec<RevisionArg>,
+    /// Unsign a commit that is not authored by you.
+    #[arg(long, short)]
+    allow_not_mine: bool,
```
yuja, line 42: @necauqua Do we need `--allow-not-mine` for unsign?

cli/tests/test_sign_command.rs
yuja, line 1: Oh, can you add `sign`/`unsign` to `test_rewrite_immutable_commands()`?

cli/src/commands/sign.rs
pylbrecht: Oh, neat! Thanks.

lib/src/commit_builder.rs
necauqua: Well, I never loaded the entire context of this specific PR into my head, so I don't have any objections whatsoever
If it works, it works, and if it also happens to make sense, the better)

cli/src/commands/sign.rs
necauqua: My thoughts about jj philosophy are that you can always `jj undo` any rewrite - at least if jj gabe you a warning/notice that something unusual has happened.

So, in that case, error => warning is good imo

//offtop
I guess the same applies to removing all: btw, even though I like it and didn't like the idea of getting rid of it - if jj brings your attention to the fact that rebase happened to affect multiple revs so that you can undo it if it was not what you expected - the all: stumble it not needed

cli/src/commands/unsign.rs
necauqua: Actually, it's the same thing with `jj undo` - I now think that we don't need this flag at all in both subcommands.

If jj prints a warning that you messed with someone's signature, you can just undo that.

Or, if we keep it, it's better to have it in both ones for consistency

pylbrecht: > Actually, it's the same thing with jj undo - I now think that we don't need this flag at all in both subcommands.
>
> If jj prints a warning that you messed with someone's signature, you can just undo that.

Couldn't agree more on removing the flag entirely. That also leaves us with a simpler interface.

cli/src/commands/sign.rs
pylbrecht: I really like the idea of embracing `jj undo`. Warnings it is then!

pylbrecht: I need some guidance on how to place (and format) these warnings.
I found `ui.warning_default()`, but it doesn't format commits as nicely as `ui.status_formatter()`.

I started with something like this:
```diff
diff --git a/cli/src/commands/sign.rs b/cli/src/commands/sign.rs
index 19853838d7..5fa3991425 100644
--- a/cli/src/commands/sign.rs
+++ b/cli/src/commands/sign.rs
@@ -19,6 +19,7 @@
 use jj_lib::commit::CommitIteratorExt;
 use jj_lib::signing::SignBehavior;
 
+use crate::cli_util::short_commit_hash;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::RevisionArg;
 use crate::command_error::CommandError;
@@ -58,6 +59,15 @@
         |rewriter| {
             let authored_by_me =
                 rewriter.old_commit().author().email == command.settings().user_email();
+
+            if !authored_by_me {
+                writeln!(
+                    ui.warning_default(),
+                    "signing commit {}, which was not authored by you",
+                    short_commit_hash(rewriter.old_commit().id())
+                )?;
+            }
+
             if rewriter.old_commit().is_signed() && authored_by_me {
                 // Don't resign commits, which are already signed by me. For people using
                 // hardware devices for signatures, resigning commits is
```
Besides being unsure what id to display (commit hash or change hash?), I couldn't write the warning within `transform_descendants()`:
```
error[E0277]: `?` couldn't convert the error to `jj_lib::backend::BackendError`
  --> cli/src/commands/sign.rs:68:18
   |
68 |                 )?;
   |                  ^ the trait `std::convert::From<std::io::Error>` is not implemented for `jj_lib::backend::BackendError`
   |
   = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
   = help: the trait `From<std::io::Error>` is not implemented for `jj_lib::backend::BackendError`
           but trait `From<jj_lib::git_backend::GitBackendError>` is implemented for it
   = help: for that trait implementation, expected `jj_lib::git_backend::GitBackendError`, found `std::io::Error`
   = note: required for `std::result::Result<(), jj_lib::backend::BackendError>` to implement `std::ops::FromResidual<std::result::Result<std::convert::Infallible, std::io::Error>>`
```

All right, then let's do it outside of `transform_descendants()` and deal with mapping errors later when polishing a working solution:
```diff
diff --git a/cli/src/commands/sign.rs b/cli/src/commands/sign.rs
index 19853838d7..28d561f0d1 100644
--- a/cli/src/commands/sign.rs
+++ b/cli/src/commands/sign.rs
@@ -19,6 +19,7 @@
 use jj_lib::commit::CommitIteratorExt;
 use jj_lib::signing::SignBehavior;
 
+use crate::cli_util::short_commit_hash;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::RevisionArg;
 use crate::command_error::CommandError;
@@ -53,11 +54,17 @@
     let mut tx = workspace_command.start_transaction();
 
     let mut signed_commits = vec![];
+    let mut foreign_commits = vec![];
     tx.repo_mut().transform_descendants(
         commits.iter().ids().cloned().collect_vec(),
         |rewriter| {
             let authored_by_me =
                 rewriter.old_commit().author().email == command.settings().user_email();
+
+            if !authored_by_me {
+                foreign_commits.push(rewriter.old_commit().clone());
+            }
+
             if rewriter.old_commit().is_signed() && authored_by_me {
                 // Don't resign commits, which are already signed by me. For people using
                 // hardware devices for signatures, resigning commits is
@@ -77,6 +84,24 @@
         },
     )?;
 
+    match &*foreign_commits {
+        [] => {}
+        [commit] => {
+            writeln!(
+                ui.warning_default(),
+                "signed commit {}, which was not authored by you",
+                short_commit_hash(commit.id())
+            )?;
+        }
+        commits => {
+            writeln!(
+                ui.warning_default(),
+                "signed the following commits, which were not authored by you:\n  {}",
+                commits.iter().ids().map(short_commit_hash).join("\n  ")
+            )?;
+        }
+    }
+
     if let Some(mut formatter) = ui.status_formatter() {
         match &*signed_commits {
             [] => {}
```
(keep in mind I am still in the stage of getting something working; `clone()` being a sign of that)

Great, it compiles! Let's get a basic test going, so we can see the actual output in a snapshot diff:
```diff
diff --git a/cli/tests/test_sign_command.rs b/cli/tests/test_sign_command.rs
index 5086b3ae03..f5dec988c7 100644
--- a/cli/tests/test_sign_command.rs
+++ b/cli/tests/test_sign_command.rs
@@ -72,7 +72,33 @@
 }
 
 #[test]
-#[should_panic]
 fn test_warn_about_signing_commits_not_authored_by_me() {
-    todo!()
+    let test_env = TestEnvironment::default();
+
+    test_env.add_config(
+        r#"
+[signing]
+sign-all = false
+backend = "test"
+"#,
+    );
+
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "one"]);
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "two"]);
+    test_env.jj_cmd_ok(&repo_path, &["commit", "-m", "three"]);
+
+    test_env.jj_cmd_ok(
+        &repo_path,
+        &[
+            "desc",
+            "--author",
+            "Someone Else <someone@else.com>",
+            "--no-edit",
+            "..@-",
+        ],
+    );
+    let (_, stderr) = test_env.jj_cmd_ok(&repo_path, &["sign", "-r", "..@-"]);
+    insta::assert_snapshot!(stderr, @r"");
 }

```
```
-old snapshot
+new results
────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
          0 │+Warning: signed the following commits, which were not authored by you:
          1 │+  4bcea9a68424
          2 │+  e14d78359449
          3 │+  ef6b36d10836
          4 │+Signed the following commits:
          5 │+  qpvuntsm hidden 82f99921 (empty) one
          6 │+  rlvkpnrz hidden 715131ae (empty) two
          7 │+  kkmpptxz hidden 60618621 (empty) three
          8 │+Rebased 1 descendant commits
          9 │+Working copy now at: zsuskuln 5a1d05b3 (empty) (no description set)
         10 │+Parent commit      : kkmpptxz 60618621 (empty) three
────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```
While it does the job, I feel it would be nice to format the warning like the output of signed commits.

But I am not sure if (ab)using `ui.status_formatter()` is the way to go or if I am overthinking this. Having spent enough time in that rabbit hole already, I felt it's time to reach out.

Any suggestions or some prevalent examples in the code you can point me to?

yuja: I think warnings can be printed later (or earlier) with the short summary line as warning, and the list of commit summaries as status output.
```
Warning: Signed N commits ... not authored by you
  <commit summary #0>
  <commit summary #1>
  ...
```

cli/src/commands/sign.rs
pylbrecht: Resolving this as obsolete, since we no longer have `--allow-not-mine`.

cli/src/commands/sign.rs
pylbrecht: While we can reduce the roots passed to `transform_descendants()`, I think we still need to apply the same checks (e.g. `no_need_to_sign_the_current_commit`) in `transform_descendants()`'s callback.

yuja: Yes, we'll need to check if the commit should be signed or not. What we can remove is the check for reparent or not.

pylbrecht: Mhh, I am not sure I fully grokked what you mean by "the check for reparent or not".
Why would we always need to reparent?

Maybe you could have a sneak peek into the recently pushed changes I made in `transform_descendants()`'s callback. I do not always reparent, but the test snapshots look reasonable to me (at least with my somewhat superficial knowledge).

For context, if we always reparent, we end up with this instead:
```diff
diff --git a/cli/tests/test_sign_command.rs b/cli/tests/test_sign_command.rs
index 18e3862265..16322afcef 100644
--- a/cli/tests/test_sign_command.rs
+++ b/cli/tests/test_sign_command.rs
@@ -52,7 +52,6 @@
       qpvuntsm hidden 8174ec98 (empty) one
       rlvkpnrz hidden 6500b275 (empty) two
       kkmpptxz hidden bcfaa4c3 (empty) three
-    Rebased 1 descendant commits
     Working copy now at: zsuskuln eeb8c985 (empty) (no description set)
     Parent commit      : kkmpptxz bcfaa4c3 (empty) three
     ");
@@ -109,7 +108,6 @@
       qpvuntsm hidden 82f99921 (empty) one
       rlvkpnrz hidden 715131ae (empty) two
       kkmpptxz hidden 60618621 (empty) three
-    Rebased 1 descendant commits
     Working copy now at: zsuskuln 5a1d05b3 (empty) (no description set)
     Parent commit      : kkmpptxz 60618621 (empty) three
     ");
```
Maybe that's what we actually want? I am leaning onto your judgement here.

yuja: See `cmd_describe()` for why. In short, it will save extra cost of redundant rewrites. I think it's more important here, because signature will be invalidated if the signed commit is rebased thereafter.

pylbrecht: Thanks, that helps!

I will play with incorporating into a test (even if only locally and for the sake of my own understanding), that we do not invalidate signatures if the signed commit is rebased thereafter.

pylbrecht:
I wanted to finish this PR this week. However, life got in the way: daycare center is understaffed this week and I have to spend the night time I have left on my dayjob responsibilities.
I won't make any promises, but I try to squeeze in a little time whenever I can. 🤞

pylbrecht:
Things are slowly getting back to normal, so I should be able to pick things up again in the next couple of days. 🎉

cli/src/commands/sign.rs
pylbrecht: Turns out writing such a test is not as easy as I imagined it to be. I am currently trying to figure out how to sign two different commits with different identities, but my brain is too tired.
(I spent the last 15 minutes butchering `unsign.rs` and wondering why my `jj unsign` tests keep passing, only to realize I have been executing the `jj sign` tests. 🙄)

Though, it would be nice to get signing with different identities in tests to work. It would enable us to write another test: warning when `jj unsign` dropped signatures of others.

yuja: I don't remember the detail, but the minimal test that will have to be added is something that could trigger automated rebase after signing, with auto-signing off?

#5707 might help craft such situation.

pylbrecht: Thanks, I believe I got a test for `jj sign` working now.
Still have to play with the one for `jj unsign` though, as I am not seeing expected warnings.

cli/tests/test_sign_command.rs
```diff
@@ -0,0 +1,195 @@
...
+      rlvkpnrz hidden 715131ae (empty) two
+      kkmpptxz hidden 60618621 (empty) three
+    Rebased 1 descendant commits
+    Working copy now at: zsuskuln 5a1d05b3 (empty) (no description set)
+    Parent commit      : kkmpptxz 60618621 (empty) three
+    ");
+}
+
+#[test]
+fn test_keep_signatures_in_rebased_descendants() {
```
pylbrecht, line 138: Following up on https://github.com/jj-vcs/jj/pull/4747#discussion_r1957269978.

I am not confident in this test case, yet.

Doesn't `signing.behavior` (#5707) dictate whether to keep or drop signatures during rebase, regardless of using `commit_builder.rebase()` or `commit_builder.reparent()`?

yuja: > Doesn't `signing.behavior` (#5707) dictate whether to keep or drop signatures during rebase, regardless of using `commit_builder.rebase()` or `commit_builder.reparent()`?

Correct. It configures the default behavior of the commit builder, afaik.

Suppose we have the following history, and we're going to sign `A` and `C` explicitly. If we didn't reparent/rebase C *before* signing, a temporary signed commit `C'` would be created, and later `C'` would be rebased. If the signing behavior was `"drop"` (or `C` was authored by someone else), `C''` wouldn't be signed. I think that's the problem to be tested.

```
C       C C'     C''
|       |/       |
B   =>  B    =>  B'
|       |        |
A       A A'     A'
|       |/       |
```

pylbrecht: Ohhh, now I get what you meant by "we *always* need to reparent".
The whole time I was assuming it's impossible to change a commit (e.g. by signing it) without rewriting its parents (or is it?). At least I couldn't find any methods on `rewriter` that would give me an instance of `CommitBuilder`, which in turn allows me to set signing behavior and rewrite the commit.

pylbrecht: I think the test case should cover this scenario now.

With `signing.behavior = drop`,
```
C
|
B
|
A
```
executing
```bash
$ jj sign -r C
$ jj sign -r A
```
should result in:
```
C (signed)
|
B
|
A (signed)
```
However, we not only need to always reparent, but also [set signing behavior to `SigningBehavior::Keep`](https://github.com/pylbrecht/jj/blob/09fcf8ba6230c1ae682aa2057dc673f0125a1f91/cli/src/commands/sign.rs#L84-L85) to not lose signatures in descendants.
Is this what we want?
(Now I am questioning if I *really* got what you meant or if I just fooled myself.)

Or are there any cases, where we actually want to drop signatures in descendants during a rebase (e.g. by honoring `signing.behavior = "drop"`)?

I can't think of one and I would at least expect a warning if `jj sign` dropped signatures in descendants during a rebase. But then again, if I purposefully configure `signing.behavior = "drop"`, I am intentionally telling `jj` to drop signatures, so maybe I *want* these signatures to be dropped.. 🤔 

I keep running in circles.

yuja: > However, we not only need to always reparent, but also [set signing behavior to SigningBehavior::Keep](https://github.com/pylbrecht/jj/blob/09fcf8ba6230c1ae682aa2057dc673f0125a1f91/cli/src/commands/sign.rs#L84-L85) to not lose signatures in descendants.
Is this what we want?

No. If the user explicitly set `signing.behavior="drop"`, rewritten commits that are not included in the `jj sign -r` revsets won't be signed.

The test I suggested was `jj sign -rC|A`, btw. If `B` wasn't rebased/reparented, the signed `C` would be rewritten immediately and therefore unsigned.

pylbrecht: Thank you for hand-holding me through the process.

> The test I suggested was jj sign -rC|A, btw. If B wasn't rebased/reparented, the signed C would be rewritten immediately and therefore unsigned.

I adapted the test accordingly and it's passing. However, I would feel more confident in the test case, if I could see it fail first. In `transform_descendants()`'s callback, how would I sign `A` *without* reparenting/rebasing `B`?

I tried the following, which should do nothing with `B`, but `C` keeps getting signed:
```diff
diff --git a/cli/src/commands/sign.rs b/cli/src/commands/sign.rs
index 3483a2a416..f14ec67ea9 100644
--- a/cli/src/commands/sign.rs
+++ b/cli/src/commands/sign.rs
@@ -66,9 +66,9 @@
         to_sign.iter().ids().cloned().collect_vec(),
         |rewriter| {
             let old_commit = rewriter.old_commit().clone();
-            let commit_builder = rewriter.reparent();
 
             if to_sign.contains(&old_commit) {
+                let commit_builder = rewriter.reparent();
                 let new_commit = commit_builder
                     .set_sign_key(args.key.clone())
                     .set_sign_behavior(SignBehavior::Own)
@@ -80,8 +80,8 @@
 
                 signed_commits.push(new_commit);
             } else {
-                commit_builder.write()?;
-                num_reparented += 1;
+                // commit_builder.write()?;
+                // num_reparented += 1;
             }
 
             Ok(())
```

yuja: Commenting out `commit_builder.write()` should be good enough. Maybe you forgot to disable auto-signing? With `signing.behavior="keep"`, `A` will be re-signed on rebase because the temporary commit has a signature.

cli/tests/test_unsign_command.rs
yuja, line 1: nit: Since `sign` and `unsign` are closely related, I would add `test_sign_unsign_commands.rs`.

cli/tests/test_sign_command.rs
pylbrecht: Finally, with `signing.behavior = "drop"` I am getting the test to fail as expected.
```diff
diff --git a/cli/src/commands/sign.rs b/cli/src/commands/sign.rs
index 3483a2a416..7f9c987fcd 100644
--- a/cli/src/commands/sign.rs
+++ b/cli/src/commands/sign.rs
@@ -80,8 +80,8 @@
 
                 signed_commits.push(new_commit);
             } else {
-                commit_builder.write()?;
-                num_reparented += 1;
+                // commit_builder.write()?;
+                // num_reparented += 1;
             }
 
             Ok(())
diff --git a/cli/tests/test_sign_command.rs b/cli/tests/test_sign_command.rs
index 949ad15a39..ccca32e126 100644
--- a/cli/tests/test_sign_command.rs
+++ b/cli/tests/test_sign_command.rs
@@ -144,7 +144,7 @@
 show-cryptographic-signatures = true
 
 [signing]
-behavior = "keep"
+behavior = "drop"
 backend = "test"
 "#,
```
And it's passing when I uncomment `commit_builder.write()?;`.

Thank you for restoring my sanity, @yuja.

cli/src/commands/sign.rs
pylbrecht: Closing this thread, as it was continued and resolved in https://github.com/jj-vcs/jj/pull/4747#discussion_r1957273685.

cli/src/commands/unsign.rs
```diff
@@ -0,0 +1,151 @@
...
+    let is_authored_by_me =
+        |commit: &Commit| -> bool { commit.author().email == command.settings().user_email() };
+
+    tx.repo_mut().transform_descendants(
+        commits.iter().ids().cloned().collect_vec(),
+        |rewriter| {
+            let old_commit = rewriter.old_commit().clone();
+            let commit_builder = rewriter.reparent();
+
+            if commits.contains(&old_commit) && old_commit.is_signed() {
```
pylbrecht, line 68: I had to move that `is_signed()` check into `transform_descendants()`.

At first I filtered when evaluating the revsets to commits:
```diff
diff --git a/cli/src/commands/unsign.rs b/cli/src/commands/unsign.rs
index 0cf7f83a06..a992604923 100644
--- a/cli/src/commands/unsign.rs
+++ b/cli/src/commands/unsign.rs
@@ -47,6 +47,7 @@
     let commits: IndexSet<Commit> = workspace_command
         .parse_union_revsets(ui, &args.revisions)?
         .evaluate_to_commits()?
+        .filter_ok(|commit| commit.is_signed())
         .try_collect()?;
 
     workspace_command.check_rewritable(commits.iter().ids())?;
```
But that failed [test_immutable_commits](https://github.com/pylbrecht/jj/blob/819a46c56abc346276713398b4c0aada8f408461/cli/tests/test_immutable_commits.rs#L394-L401) with
```
Nothing changed.
```
Since `main` is not signed, `jj unsign` filters out `main` and there is nothing left to unsign, which results in exit code `0`, which in turn fails the test.

Given that the test setup in `test_immutable_commits` for that scenario is fairly complex compared to moving the `is_signed()` check into `transform_descendants()`, I went for the latter.

pylbrecht:
@yuja, docs (and commit messages) still need some love, but other than that I would appreciate another look from you.
I had to extend `TestEnvironment` a little (819a46c56abc346276713398b4c0aada8f408461) to get a test for `jj unsign` working. Let me know what you think!

cli/tests/common/test_environment.rs
```diff
@@ -262,6 +266,16 @@ impl TestEnvironment {
...
 
+    #[allow(dead_code)]
+    pub fn set_user_name(&mut self, user_name: String) {
+        self.user_name = user_name;
+    }
+
+    #[allow(dead_code)]
+    pub fn set_user_email(&mut self, user_email: String) {
+        self.user_email = user_email;
+    }
```
yuja, line 275: You can override `JJ_USER`/`JJ_EMAIL` instead of configuring `TestEnvironment`.

```rust
test_env.run_jj_with(|cmd| cmd.current_dir(repo_root).env(..))
``

File: cli/src/commands/sign.rs [hunk 28]
```diff
@@ -0,0 +1,161 @@
...
+
+    let is_authored_by_me =
+        |commit: &Commit| -> bool { commit.author().email == command.settings().user_email() };
+    let is_signed_by_me =
+        |commit: &Commit| -> bool { commit.is_signed() && is_authored_by_me(commit) };
+
+    let to_sign: IndexSet<Commit> = workspace_command
+        .parse_union_revsets(ui, &args.revisions)?
+        .evaluate_to_commits()?
+        .filter_ok(|commit| !is_signed_by_me(commit))
```
yuja, line 62: I think we decided to not add `--allow*` flags and just sign all commits?

https://github.com/jj-vcs/jj/pull/4747#discussion_r1926358502

BTW, we might want to exclude `is_signed_by_me()` commits from the roots of `transform_descendants()`. Root commits that are already signed wouldn't have to be re-signed. This is for optimization, so can be addressed later.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,161 @@
...
+            [] => {}
+            [commit] => {
+                writeln!(ui.warning_default(), "Signed 1 commit not authored by you")?;
+                write!(formatter, "  ")?;
+                tx.base_workspace_helper()
+                    .write_commit_summary(formatter.as_mut(), commit)?;
+                writeln!(ui.status())?;
+            }
+            commits => {
+                let template = tx.base_workspace_helper().commit_summary_template();
```
yuja, line 110: You'll probably need to use `tx.commit_summary_template()` because `commits` are the list of the rewritten commits.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,161 @@
...
+                writeln!(
+                    ui.warning_default(),
+                    "Signed {} commits not authored by you",
+                    commits.len()
+                )?;
+                for commit in commits {
+                    write!(formatter, "  ")?;
+                    template.format(commit, formatter.as_mut())?;
+                    writeln!(formatter)?;
+                }
```
yuja, line 120: Since these outputs are mostly identical, I would combine them. We usually use plural form.

cli/src/commands/unsign.rs
yuja: You can filter `commits` after testing `check_rewritable()`. That's probably easier than skipping unneeded `.reparent()` in `transform_descendants()`.

cli/src/commands/sign.rs
pylbrecht: > I think we decided to not add --allow* flags and just sign all commits?

Yes, we did!

> BTW, we might want to exclude is_signed_by_me() commits from the roots of transform_descendants(). Root commits that are already signed wouldn't have to be re-signed. This is for optimization, so can be addressed later.

That's exactly what I intended to do here, since resigning is cumbersome for people using hardware devices (https://github.com/jj-vcs/jj/pull/4747#discussion_r1916030580).

Just for my understanding, why do we need to do the filtering later on the roots of `transform_descendants()` and not here?

cli/src/commands/unsign.rs
pylbrecht: Thanks. That seems so obvious, now that I read it.
It's about time that I get this PR over the finish line. I have been looking at that code for far too long.

cli/tests/common/test_environment.rs
pylbrecht: Neat, thanks for pointing that out!

cli/src/commands/sign.rs
pylbrecht: All right, I wasn't really sure how "perfect" we want the phrasing to be.
Happy to go with the plural form then.

cli/src/commands/sign.rs
pylbrecht: Deja vu: https://github.com/jj-vcs/jj/pull/4747#discussion_r1925330331
Sorry for missing that.

pylbrecht: Ohhh, so that makes the `(hidden)` go away. Nice!
Maybe you can help me fill my knowledge gaps here.

What's the difference between using
```rust
tx.base_workspace_helper().commit_summary_template()
```
and
```rust
tx.commit_summary_template()
```
?

I found some pieces of the puzzle in the glossary ([View](https://jj-vcs.github.io/jj/latest/glossary/#view)) and architecture docs ([ReadOnlyRepo](https://jj-vcs.github.io/jj/latest/technical/architecture/#readonlyrepo), [MutableRepo](https://jj-vcs.github.io/jj/latest/technical/architecture/#mutablerepo)).

It seems `tx.base_workspace_helper().commit_summary_template()` provides a `ReadOnlyRepo` view, whereas `tx.commit_summary_template()` a `MutableRepo` view.

The commits show up as hidden from the `ReadOnlyRepo` view, because they have been rewritten *and* we haven't finished the transaction, yet. That's why these rewritten commits are only visible from the `MutableRepo` view.
After finishing the transaction, a `ReadOnlyRepo` view would also show these commit as visible, correct?

(I hope I didn't butcher the concepts and terms too much.)

martinvonz: > After finishing the transaction, a `ReadOnlyRepo` view would also show these commit as visible, correct?

Close. `ReadonlyRepo` is immutable, so it won't change when the transaction commits. However, when [the transaction is committed](https://github.com/jj-vcs/jj/blob/35936463ed89dadbccb3ab743b345981e57e917e/lib/src/transaction.rs#L105-L111), you get a new `ReadonlyRepo` back.

martinvonz: Oh, let me also answer this part more directly:

> The commits show up as hidden from the `ReadOnlyRepo` view, because they have been rewritten _and_ we haven't finished the transaction, yet.

`ReadonlyRepo` has an immutable view of the repo at a particular operation. The view includes the set of visible head commits. By definition, those commits and their ancestors are considered visible. Any other commit is considered hidden. So the new commits are hidden from the `ReadonlyRepo`'s point of view because they were not visible at that operation.

cli/src/commands/sign.rs
yuja: Ah, I misunderstood the filtering logic, sorry.

I think the condition will have to be `is_signed() && signature == <current_configured_key>`, but I have no idea if that's doable.

> why do we need to do the filtering later on the roots of `transform_descendants()` and not here?

If the signed commit is descendant of other commits to be signed, it will have to be re-signed after reparenting/rebasing.

cli/src/commands/sign.rs
pylbrecht: Thanks a lot for elaborating, @martinvonz!

cli/src/commands/sign.rs
pylbrecht: Ah, true. `signed_by_me != (signed && authored_by_me)`

> I think the condition will have to be is_signed() && signature == <current_configured_key>, but I have no idea if that's doable.

Off the top of my head, I think we would need to create a new signature on the fly to compare against. That's too bad, since it won't avoid the resigning issue.

> It's probably okay to approximate the condition as `signed_by_someone_including_me && authored_by_me`.

Now I am having second thoughts. 😅 
I am wondering how likely the scenario of having a commit, which is `signed_by_someone_else && authored_by_me`, actually is..
I recall `jj desc` dropping signatures, which would make it very unlikely, but I may be misremembering.

Anyway, to avoid more back and forth I would leave the appromixation in place and add some disclaimer to the command's docs and/or code. 

> If the signed commit is descendant of other commits to be signed, it will have to be re-signed after reparenting/rebasing.

Oh yeah, I missed that fact. Thanks for clarifying!

pylbrecht: > Anyway, to avoid more back and forth I would leave the approximation in place and add some disclaimer to the command's docs and/or code.

I changed my mind. I will go for always resigning and open another issue for addressing the optimization.
While it's more painful for users with hardware devices (myself included), it's clearer to communicate and probably leads to less surprises than using the approximation.

(So much for avoiding back and forth.)

CHANGELOG.md
```diff
@@ -259,6 +259,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   Please [see the documentation](/docs/config.md#json-schema-support) for more
   on this.
 
+* The new `jj sign` command allows signing commits.
```
yuja, line 262: nit: you'll need to move this to the unreleased section.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,144 @@
...
+use crate::complete;
+use crate::ui::Ui;
+
+/// Cryptographically sign a revision
+///
+/// Note that revisions, which are already signed by you, will be resigned.
+///
+/// While that leads to discomfort for users, which sign with hardware devices,
+/// as of now we cannot reliably check if a commit is already signed by the user
+/// without creating a signature (see https://github.com/jj-vcs/jj/issues/5786).
```
yuja, line 35: nit: maybe this can be added to the help of the `--revisions` argument? It's odd that the comment starts with `revisions` without specifying the details.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,144 @@
...
+        value_name = "REVSETS",
+        add = ArgValueCandidates::new(complete::mutable_revisions),
+    )]
+    revisions: Vec<RevisionArg>,
+}
+
+pub fn cmd_sign(ui: &mut Ui, command: &CommandHelper, args: &SignArgs) -> Result<(), CommandError> {
+    let mut workspace_command = command.workspace_helper(ui)?;
+
+    if command.settings().signing_backend()?.is_none() {
```
yuja, line 53: nit: maybe better to use `repo.store().signer().can_sign()`?

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,144 @@
...
+        return Err(user_error_with_hint(
+            "No signing backend configured.",
```
yuja, line 55: nit: error message doesn't usually include period.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,144 @@
...
+            "For configuring a signing backend, see https://jj-vcs.github.io/jj/latest/config/#commit-signing",
+        ));
+    }
+
+    let is_authored_by_me =
+        |commit: &Commit| -> bool { commit.author().email == command.settings().user_email() };
```
yuja, line 61: nit: can you use `workspace_command.settings()` or `tx.settings()`? They should be identical to `command.settings()`, but it's safer to get the settings object from the workspace.

It might be easier to inline this closure.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,144 @@
...
+    tx.repo_mut().transform_descendants(
+        to_sign.iter().ids().cloned().collect_vec(),
+        |rewriter| {
+            let old_commit = rewriter.old_commit().clone();
+            let commit_builder = rewriter.reparent();
+
+            if to_sign.contains(&old_commit) {
+                let new_commit = commit_builder
+                    .set_sign_key(args.key.clone())
+                    .set_sign_behavior(SignBehavior::Own)
```
yuja, line 85: Shouldn't this be `Force`?

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,144 @@
...
+    )?;
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        if !foreign_commits.is_empty() {
+            let template = tx.commit_summary_template();
+            writeln!(
+                ui.warning_default(),
+                "Signed {} commit(s) not authored by you",
+                foreign_commits.len()
+            )?;
```
yuja, line 109: nit: warnings should be printed even if the status output is suppressed by `--quiet`.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        if !signed_commits.is_empty() {
+            let template = tx.commit_summary_template();
+            writeln!(formatter, "Signed {} commit(s):", signed_commits.len())?;
+            for commit in &signed_commits {
+                write!(formatter, "  ")?;
+                template.format(commit, formatter.as_mut())?;
+                writeln!(formatter)?;
+            }
+        }
```
yuja, line 125: It seems odd that "foreign" commits are printed twice. Maybe we can reorganize the status/warning messages something like this?
```
Signed N commits:
  ...
Warning: M of these commits are not authored by you.
```

cli/tests/test_sign_command.rs
```diff
@@ -0,0 +1,207 @@
...
+    Signed 3 commit(s):
+      qpvuntsm 82f99921 (empty) one
+      rlvkpnrz 715131ae (empty) two
+      kkmpptxz 60618621 (empty) three
+    Rebased 1 descendant commits
+    Working copy now at: zsuskuln 5a1d05b3 (empty) (no description set)
+    Parent commit      : kkmpptxz 60618621 (empty) three
+    [EOF]
+    ");
+}
```
yuja: Maybe we need to snapshot `jj log` to ensure that these commits are signed.

cli/tests/test_immutable_commits.rs
```diff
@@ -521,4 +521,20 @@ fn test_rewrite_immutable_commands() {
     [EOF]
     [exit status: 1]
     "##);
+    // sign
+    test_env.add_config(r#"signing.backend = "test""#);
+    let output = test_env.run_jj_in(&repo_path, &["sign", "-r=main"]);
```
yuja, line 526: nit: use `--config=signing.backend=test` instead.

cli/tests/test_sign_unsign_commands.rs
yuja, line 1: Can you squash the last commit?

cli/src/commands/sign.rs
pylbrecht: I used `Force` in the very beginning, but can't recall why I changed it to `Own` anymore.
Will change it back to `Force`.

cli/tests/test_immutable_commits.rs
pylbrecht: That is much nicer, thanks!

cli/src/commands/sign.rs
pylbrecht: Thanks, having it on the `--revisions` argument does put it into context.

CHANGELOG.md
pylbrecht: Ah thanks. Another sign that I need to get this PR done. 😅

cli/src/commands/sign.rs
yuja: Your code snippet looks good to me because we just need to test the explicitly-specified commits. Maybe we can also count that later, `signed_commits.iter().filter(..).count()`.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::command_error::user_error_with_hint;
+use crate::command_error::CommandError;
+use crate::complete;
+use crate::ui::Ui;
+
+/// Cryptographically sign a revision
+#[derive(clap::Args, Clone, Debug)]
+pub struct SignArgs {
+    /// What key to use, depends on the configured signing backend.
+    #[arg()]
```
yuja, line 34: nit: remove empty `arg()`

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,134 @@
...
+    key: Option<String>,
+    /// What revision(s) to sign
+    ///
+    /// Note that revisions, which are already signed by you, will be resigned.
```
yuja, line 38: nit: I think `by you` can be removed since all target revisions will be re-signed.

(fwiw, it might be better to say "re-sign" instead of "resign", but I'm not native.)

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::cli_util::RevisionArg;
+use crate::command_error::user_error_with_hint;
+use crate::command_error::CommandError;
+use crate::complete;
+use crate::ui::Ui;
+
+/// Cryptographically sign a revision
+#[derive(clap::Args, Clone, Debug)]
+pub struct SignArgs {
+    /// What key to use, depends on the configured signing backend.
```
yuja, line 33: nit: remove period from summary line of clap help.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    tx.repo_mut().transform_descendants(
+        to_sign.iter().ids().cloned().collect_vec(),
+        |rewriter| {
+            let old_commit = rewriter.old_commit().clone();
+            let commit_builder = rewriter.reparent();
+
+            if to_sign.contains(&old_commit) {
+                let new_commit = commit_builder
+                    .set_sign_key(args.key.clone())
```
yuja, line 81: Doesn't it override the default key no matter if `--key` is specified or not?

We should test this behavior, but I assume we'll need to update the test backend to do so. Maybe better to drop `--key` argument from the initial implementation?

I think the backend API can be changed to `.set_sign_key(String)` and `.clear_sign_key()` to prevent misuse.

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            Ok(())
+        },
+    )?;
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        if !signed_commits.is_empty() {
+            let template = tx.commit_summary_template();
+            writeln!(formatter, "Signed {} commit(s):", signed_commits.len())?;
```
yuja, line 98: nit: just say `commits` (plural) for consistency with rebase messages?

(same for `unsign`)

cli/src/commands/sign.rs
```diff
@@ -0,0 +1,134 @@
...
+            for commit in &signed_commits {
+                write!(formatter, "  ")?;
+                template.format(commit, formatter.as_mut())?;
+                writeln!(formatter)?;
+            }
+        }
+    }
+
+    let num_not_authored_by_me = to_sign
```
yuja, line 107: nit: I think `signed_commits` is more correct here.

cli/tests/test_sign_unsign_commands.rs
```diff
@@ -250,3 +250,135 @@ backend = "none"
...
+      qpvuntsm 757aba72 (empty) one
+      rlvkpnrz 49a6eeeb (empty) two
+      kkmpptxz 8859969b (empty) three
+      zsuskuln 8cea2d75 (empty) (no description set)
+    Warning: 4 of these commits are not authored by you
+    Working copy now at: zsuskuln 8cea2d75 (empty) (no description set)
+    Parent commit      : kkmpptxz 8859969b (empty) three
+    [EOF]
+    ");
+}
```
yuja, line 384: nit: snapshot `jj log` output to ensure that these commits are unsigned?

cli/tests/test_sign_command.rs
```diff
@@ -0,0 +1,252 @@
...
+show-cryptographic-signatures = true
+
+[signing]
+behavior = "keep"
+backend = "test"
+"#,
+    );
+
+    test_env
+        .run_jj_in(test_env.env_root(), &["git", "init", "repo"])
```
yuja: nit: `&` can be omitted (applies to all `.run_jj_in()` callers)

yuja (approved):

cli/src/commands/sign.rs
pylbrecht: Oh yeah, absolutely. Thanks for spotting!

cli/src/commands/sign.rs
pylbrecht: Oh wow, good catch.

We indeed always override the key. That's bad.
I would have expected that if we pass in `None`, `set_sign_key()` would just do nothing.

> I think the backend API can be changed to .set_sign_key(String) and .clear_sign_key() to prevent misuse.

👍

cli/src/commands/sign.rs
pylbrecht: I suppose this comment does no longer apply, since move `--key` to https://github.com/jj-vcs/jj/issues/5804?

pylbrecht: For the sake of not dragging this out any further from my side, I will resolve this conversation to merge.

PR #4749, opened 2024-11-01
add rust-toolchain.toml

senekor:
CI is pinned to our MSRV. This file makes rustup automatically select that same version for people developing locally with rustup. MSRV violations can therefore be caught earilier during development, instead of in CI.

I had some MSRV related broken CI [here](https://github.com/martinvonz/jj/pull/4737), which I didn't notice locally. I don't think this is extremely important, just nice to have if there are no downsides. I don't know of any downsides, but maybe someone else does.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ members = ["cli", "lib", "lib/gen-protos", "lib/proc-macros", "lib/testutils"]
 [workspace.package]
 version = "0.22.0"
 license = "Apache-2.0"
-rust-version = "1.76" # NOTE: remember to update CI, contributing.md, changelog.md, install-and-setup.md, and flake.nix
+rust-version = "1.76" # NOTE: remember to update CI, contributing.md, changelog.md, install-and-setup.md, and rust-toolchain.toml
 edition = "2021"
 readme = "README.md"
 homepage = "https://github.com/martinvonz/jj"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -7,7 +7,7 @@ members = ["cli", "lib", "lib/gen-protos", "lib/proc-macros", "lib/testutils"]
 [workspace.package]
 version = "0.22.0"
 license = "Apache-2.0"
-rust-version = "1.76" # NOTE: remember to update CI, contributing.md, changelog.md, install-and-setup.md, and flake.nix
```
senekor, line 10: flake.nix seems to use nightly Rust since 749481e5527c092da9096504e08c6aa065f598bb

yuja:
See https://github.com/martinvonz/jj/pull/4613 and https://github.com/martinvonz/jj/pull/1913 for the previous discussions.

senekor:
Great, thanks for the links. The new `incompatible_msrv` didn't trigger for me because it was actually code of the `clap_complete` dependency with the `unstable-dynamic` flag turned on that violated clap's MSRV. clap doesn't check this feature in CI, I opened an issue about that.

I agree with [this comment](https://github.com/martinvonz/jj/pull/1913#issuecomment-2037636332) that the link makes `rust-toolchain.toml` unnecessary.

PR #4754, opened 2024-11-02
Nix devshell fixes

JoaquinTrinanes:
This PR addresses a couple of rough edges I found while setting up my local development environment for the project using the included flake.

The first thing I noticed is that I had no rust LSP because `rust-analyzer` was not included in the shell, which led me to do a small refactor of the flake:

- Included the toolchain's `rust-analyzer`in the devShell packages
- Added the formatter as well
- Formatted the flake
- Instantiate the shell using `packages` instead of `buildInputs`

I thought about doing a more opinionated and aggressive refactor (removing the dependency on `flake-utils`, explicitly adding `nixpkgs` as inputs...) but I feel like these are enough changes for my first PR to the repo 😅

# Checklist

If applicable:


thoughtpolice:
The first commit is a regression from https://github.com/martinvonz/jj/commit/4a2983b77ec10da66b27fd5625ec4da3451c6a0a which landed recently, which I thought I would have caught, but apparently, I didn't 🤷🏻 Maybe my editor wasn't fully restarted.

However, I'm sorry (not sorry) to say that I totally despise some of nix-fmt's choices, so I'd rather not take the third commit. A few of the changes for things like `overrideAttrs` the parens are fine. Otherwise, it seems fine to me.

JoaquinTrinanes:
Agree, I overall don't like `nixpkgs-fmt`, but because it was set as the formatter I assumed it was what I was supposed to use. I use `nixfmt-rfc-style` for my personal projects, and I'd be happy to open a follow-up PR switching to it if you like the results, or removing the formatter field altogether if none are desired.

For now, I just dropped that commit :)

emilazy:
I think `nixfmt-rfc-style` is the one Austin hates, but I guess maybe he hates `nixpkgs-fmt` too 😅

thoughtpolice (approved):

thoughtpolice:
It might be both. Can't remember. My hatred is unlimited and knows no bounds.

Issue #4746, opened 2024-10-31
`jj -h --color=never` and `jj -h --no-pager` do not work with options in this order


ilyagr:
`jj --color=never -h` and `jj  --no-pager -h` work as expected. `jj -h --color=never` and `jj -h --no-pager` ignore the extra option.

## Specifications

- Platform: ARM Mac OS
- Version: Today's nightly, jj 0.22.0-f2acb98ef2a0b33e665d28d58b59bd3d713b0fbd

yuja:
Just for reference, see this thread for how `-h`/`--help` is interpreted by clap: https://github.com/martinvonz/jj/pull/4518#discussion_r1786655967

ilyagr:
From a certain perspective, this could seem quite minor, but I think it is worth addressing (if we can) because:

- this will happen to beginners and people looking for help
- this messes up the principle that the order of `jj` arguments is usually irrelevant
- these two points combined mean that it may take new people a long time to realize this is an exception from a rule. They will instead wonder every time whether reordering arguments will trip them up, and might abandon `jj` thinking the UI is confusing.

Of course, this does not mean that any amount of tortured code is acceptable to solve this problem.

PR #4755, opened 2024-11-02
cli: insert dummy -h/--help flag when parsing early args

yuja:
#4746

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2908,7 +2908,7 @@ fn resolve_default_command(
     app: &Command,
     mut string_args: Vec<String>,
 ) -> Result<Vec<String>, CommandError> {
-    const PRIORITY_FLAGS: &[&str] = &["help", "--help", "-h", "--version", "-V"];
+    const PRIORITY_FLAGS: &[&str] = &["--help", "-h", "--version", "-V"];
 
     let has_priority_flag = string_args
         .iter()
@@ -3033,7 +3033,6 @@ fn handle_early_args(
         .clone()
         .disable_version_flag(true)
         .disable_help_flag(true)
-        .disable_help_subcommand(true)
         .ignore_errors(true)
         .try_get_matches_from(args)?;
     let mut args: EarlyArgs = EarlyArgs::from_arg_matches(&early_matches).unwrap();
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3032,7 +3032,16 @@ fn handle_early_args(
     let early_matches = app
         .clone()
         .disable_version_flag(true)
+        // Do not emit DisplayHelp error
         .disable_help_flag(true)
+        // Do not stop parsing at -h/--help
+        .arg(
+            clap::Arg::new("help")
+                .short('h')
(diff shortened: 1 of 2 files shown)


cli/tests/test_global_opts.rs
```diff
@@ -559,6 +559,16 @@ fn test_early_args() {
     let stdout = test_env.jj_cmd_success(test_env.env_root(), &["help", "--color=always"]);
     insta::assert_snapshot!(stdout.lines().find(|l| l.contains("Commands:")).unwrap(), @"[1m[4mCommands:[0m");
 
+    // Check that early args are accepted after -h/--help
```
ilyagr, line 562: (Optional) Should we also test `--color=always -h`?Or is that completely equivalent to `--color=always --help`?

ilyagr (approved):
Thank you very much! This looks good to me, I'll also ping @Grillo-0 in case they have any thoughts.

cli/src/cli_util.rs
```diff
@@ -3032,8 +3032,15 @@ fn handle_early_args(
     let early_matches = app
         .clone()
         .disable_version_flag(true)
+        // Ignore any appearance of -h/--help and continue parsing
```
ilyagr, line 3035: I am a bit confused about why this works. Shouldn't `.disable_help_flag().ignore_errors(true)` have been sufficient to ignore `-h` while parsing?

cli/tests/test_global_opts.rs
yuja: They're equivalent in implementation wise.

cli/src/cli_util.rs
yuja: In clap, help is an "error" (or the thing that triggers exceptional code path.) That's my understanding.

ilyagr: yes, but shouldn't `ignore_errors` ignore this error? I guess it clearly doesn't. Maybe it's a documentation bug in clap.

Thanks again for figuring it out.

ilyagr: If this is true, maybe the comment should be moved to the next line and say:

> Without this, parsing will stop on the first `-h` or `--help`, despite .disable_help_flag() and .ignore_errors().

yuja: > yes, but shouldn't ignore_errors ignore this error, especially since the help flag is disabled? 

If the default `-h` is disabled, `-h` triggers a normal parse error. If it's enabled, `-h` triggers a "help requested error". In both cases, any remaining args won't be parsed. We do need to skip over `-h`. And that's why I put the comment there.

yuja: Ok, that appears wrong. `.ignore_errors(true)` doesn't ignore "help error" (maybe because it isn't a parse error?) `.disable_help_flag(true)` disables the help error as well as parsing of `-h`, so we need a rule to parse `-h`.

Grillo-0 (approved):
LGTM :)

cli/src/cli_util.rs
yuja: Updated comment.

PR #4758, opened 2024-11-02
docs/FAQ.md: answer monitoring `jj log` with watch and TUIs

tim-janik:
As discussed on [Discord](https://discord.com/channels/968932220549103686/968932220549103689/1302302710393208943), this describes ways to monitor `jj log` output as it evolves in the FAQ.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -37,6 +37,23 @@ See [revsets] and [templates] for further guidance.
 
 Use `jj log -r ..`. The `..` [operator] lists all visible commits in the repo, excluding the root (which is never interesting and is shared by all repos).
 
+### Can I monitor how `jj log` evolves?
+
+The simplest way to monitor how the history as shown by `jj log` evolves is by using the [watch(1)](https://man7.org/linux/man-pages/man1/watch.1.html) command (or [hwatch](https://github.com/blacknon/hwatch?tab=readme-ov-file#configuration) or [viddy](https://github.com/sachaos/viddy)).
+For example:
+
+```sh
+watch --color jj --ignore-working-copy log --color=always
+```
+
+This will continuously update the (colored) log output in the terminal.
+The `--ignore-working-copy` option avoids conflicts with manual operations during the creation of snapshots.
+Martin used watch in a [tmux](https://github.com/tmux/tmux/wiki) pane during his presentation [Jujutsu - A Git-compatible VCS](https://www.youtube.com/watch?v=LV0JzI8IcCY).
+
+Alternatively, you can use [jj-fzf](https://github.com/tim-janik/jj-fzf), where the central piece is the `jj log` view and common operations can be carried out via key bindings while the log view updates.
+
+The wiki lists additional TUIs and GUIs beyond the terminal: [GUI-and-TUI](https://github.com/martinvonz/jj/wiki/GUI-and-TUI)
+
 ### Should I co-locate my repository?
 
 Co-locating a Jujutsu repository allows you to use both Jujutsu and Git in the


PhilipMetzger (approved):
LG, although the title could be either "docs:", "faq:" or "docs/faq:".

docs/FAQ.md
```diff
@@ -37,6 +37,23 @@ See [revsets] and [templates] for further guidance.
 
 Use `jj log -r ..`. The `..` [operator] lists all visible commits in the repo, excluding the root (which is never interesting and is shared by all repos).
 
+### Can I monitor how `jj log` evolves?
+
+The simplest way to monitor how the history as shown by `jj log` evolves is by using the `watch(1)` command (or [hwatch](https://github.com/blacknon/hwatch) or [viddy](https://github.com/sachaos/viddy)).
```
ilyagr, line 42: Optional: I always give a link to https://github.com/blacknon/hwatch?tab=readme-ov-file#configuration when mentioning hwatch, since its defaults are quite weird.

You could also link to https://man7.org/linux/man-pages/man1/watch.1.html and/or https://gitlab.com/procps-ng/procps for `watch`.

docs/FAQ.md
```diff
@@ -37,6 +37,23 @@ See [revsets] and [templates] for further guidance.
...
+For example:
+
+```sh
+watch jj --ignore-working-copy log --color=always
+```
+
+This will continuously update the (colored) log output in the terminal.
+The `--ignore-working-copy` option avoids conflicts with manual operations during the creation of snapshots.
+Martin used this in a [tmux](https://github.com/tmux/tmux/wiki) pane during his presentation [Jujutsu - A Git-compatible VCS](https://www.youtube.com/watch?v=LV0JzI8IcCY).
```
ilyagr, line 51: Minor: I'd put a newline before this sentence and say `watch` instead of "this".

docs/FAQ.md
```diff
@@ -37,6 +37,23 @@ See [revsets] and [templates] for further guidance.
 
 Use `jj log -r ..`. The `..` [operator] lists all visible commits in the repo, excluding the root (which is never interesting and is shared by all repos).
 
+### Can I monitor how `jj log` evolves?
+
+The simplest way to monitor how the history as shown by `jj log` evolves is by using the `watch(1)` command (or [hwatch](https://github.com/blacknon/hwatch) or [viddy](https://github.com/sachaos/viddy)).
+For example:
+
+```sh
+watch jj --ignore-working-copy log --color=always
```
ilyagr, line 46: Important: You need `--color`. I'd make this:

     watch --color -- jj --ignore-working-copy --color=always log

docs/FAQ.md
```diff
@@ -37,6 +37,23 @@ See [revsets] and [templates] for further guidance.
...
+```
+
+This will continuously update the (colored) log output in the terminal.
+The `--ignore-working-copy` option avoids conflicts with manual operations during the creation of snapshots.
+Martin used this in a [tmux](https://github.com/tmux/tmux/wiki) pane during his presentation [Jujutsu - A Git-compatible VCS](https://www.youtube.com/watch?v=LV0JzI8IcCY).
+
```
ilyagr, line 52: Optional (or I could add it): I'd add an example of a more complicated command.

> Here is an example of a more complicated `watch` invocation, using some Unix tools:
>
> ```shell
> watch --color -t -- "
>      jj --ignore-working-copy --color=always diff --stat | grep -Ev '^\d files'
>      jj --ignore-working-copy log --color=always"
> ```

ilyagr (approved):
There's one mistake that needs fixing (--color). I also added a bunch of take-it-or-leave-it suggestions.

docs/FAQ.md
ilyagr: On second thought, maybe this is not such a great idea.

The problem is that `diff --stat` would only be updated after `jj` commands, which would be noticeable. So, we'd have to explain the tradeoff of keeping it that way, removing one of the `--ignore-working-copy`-ies, or using `core.watchman.register_snapshot_trigger=true`.

The downside of the last two is the increased chance of concurrency causing divergent change ids or stale workspaces. Also, using both of them makes the chance larger.

docs/FAQ.md
tim-janik: I don't know the tool. Make a call, do you want the change or not?

tim-janik: Had another edit to make, so I added the links now.

docs/FAQ.md
tim-janik: Ok fixed, I have not added the newline though, I think it makes sense to keep the watch related paragraph in once piece.

Issue #3082, opened 2024-02-18
FR: Support shell command aliases


yjyao:
**Is your feature request related to a problem? Please describe.**
the `aliases` in `jj`'s config can only run `jj` subcommands.

**Describe the solution you'd like**
to be able to configure an alias that runs a shell command. both git and mercurial support this. with this, one can chain multiple `jj` commands with one alias, which i don't think is possible with today's `jj` aliases.

yuja:
Related discussion in https://github.com/martinvonz/jj/issues/3001

PR #4759, opened 2024-11-03
util: add exec command for arbitrary aliases

### LINKED ISSUE #3001: FR: run `jj-*` commands in PATH
i want to add an alias that involves multiple jj commands (specifically `jj branch set main -r @- && jj git push`). currently, it's not possible to run multiple commands in aliases, because there's no equivalent of git's `!` feature to run things in the shell.

that's ok, i'm not asking for it to be changed, but it would be nice if instead `jj push-head` ran `jj-push-head` in PATH, so i don't have to remember which commands are aliases and which are shell helpers.

add support for shell execution in aliases

**Additional context**
jj 0.13.0-81b0e3bf3bf6f230740e93532b9a9767e8a4ef6e

senekor:
closes #3001 

I also believe that this PR, if accepted, makes #3673 redundant. The discussion there revolves around some kind of custom, lightweight, embedded scripting language. But the original use case is just streamlining workflows, which is fully covered by this PR. Nothing gives the user more flexibility and is less of a maintenance burden to us than letting users bring their own scripting language.

# Checklist

If applicable:


senekor:
```
thread 'test_util_command::test_util_exec' panicked
stderr="Error: No such file or directory (os error 2)\n"
```

Nix doesn't have `echo` installed by default? :flushed: lol. What do I do? :thinking:

cli/src/commands/util.rs
```diff
@@ -99,6 +101,45 @@ pub(crate) struct UtilMarkdownHelp {}
...
+# This makes sure that flags are passed to your script instead of parsed by jj.
+```
+
+If you don't want to have to manage your script as a separate file, you can
+even inline a script into your config file:
+
+```toml
+[aliases]
+my-inline-script = ["util", "exec", "--", "bash", "-c", """
+#!/bin/bash
```
DD5HT, line 124: don't use bin/bash directly always use `#!/usr/bin/env bash`

cli/src/commands/util.rs
```diff
@@ -99,6 +101,45 @@ pub(crate) struct UtilMarkdownHelp {}
...
+set -euo pipefail
+echo "Look Ma, everything in one file!"
+echo "args: $@"
+""", ""]
+#    ^^
+# This last empty string will become "$0" in bash, so your actual arguments
+# are all included in "$@" and start at "$1" as expected.
+```
+"#]
```
PhilipMetzger, line 133: Nit: use a normal doc comment and use the `verbatim_doc_comment` clap attribute.

cli/tests/test_util_command.rs
```diff
@@ -112,3 +112,13 @@ fn test_shell_completions() {
     test("nushell");
     test("zsh");
 }
+
+#[cfg(unix)]
```
PhilipMetzger, line 116: Nit: Cross-platform tests, since we cannot just implement the feature purely for Unix.

PhilipMetzger (changes requested):
Blocking on missing consensus.

cli/tests/test_util_command.rs
senekor: Do you have an idea for a command that's always available on Windows? I don't use Windows and I don't want to debug with a CI that takes over 5mins to run.

PhilipMetzger: Use test tools introduced by `jj fix`.

senekor:
also closes #3082

cli/src/commands/util.rs
```diff
@@ -35,6 +36,7 @@ pub(crate) enum UtilCommand {
     Mangen(UtilMangenArgs),
     MarkdownHelp(UtilMarkdownHelp),
     ConfigSchema(UtilConfigSchemaArgs),
+    Exec(UtilExecArgs),
```
yuja, line 39: nit: the list should (have been) sorted. I noticed util.rs isn't split up to per-command files yet, I'll send a patch separately.

cli/src/commands/util.rs
```diff
@@ -230,6 +271,24 @@ fn cmd_util_config_schema(
     Ok(())
 }
 
+fn cmd_util_exec(args: &UtilExecArgs) -> Result<(), CommandError> {
+    let status = std::process::Command::new(&args.command)
+        .args(&args.args)
+        .spawn()
+        .map_err(|e| CommandError::new(CommandErrorKind::User, e))?
+        .wait()
+        .unwrap();
```
yuja, line 280: nit: `.status()` can be used
https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.status

cli/src/commands/util.rs
```diff
@@ -230,6 +271,24 @@ fn cmd_util_config_schema(
...
+    if !status.success() {
+        let error_msg = if let Some(code) = status.code() {
+            format!("external command failed with exit code {code}")
+        } else {
+            "external command was terminated by signal".into()
+        };
```
yuja, line 286: nit: `ExitStatus` implements `Display`. Something like this should work: `"External command exited with {exit_status}"`

cli/src/commands/util.rs
```diff
@@ -122,6 +162,7 @@ pub(crate) fn cmd_util(
         UtilCommand::Mangen(args) => cmd_util_mangen(ui, command, args),
         UtilCommand::MarkdownHelp(args) => cmd_util_markdownhelp(ui, command, args),
         UtilCommand::ConfigSchema(args) => cmd_util_config_schema(ui, command, args),
+        UtilCommand::Exec(args) => cmd_util_exec(args),
```
yuja, line 165: nit: pass `ui, command` consistently

cli/src/commands/util.rs
```diff
@@ -230,6 +271,24 @@ fn cmd_util_config_schema(
...
+        .map_err(|e| CommandError::new(CommandErrorKind::User, e))?
+        .wait()
+        .unwrap();
+    if !status.success() {
+        let error_msg = if let Some(code) = status.code() {
+            format!("external command failed with exit code {code}")
+        } else {
+            "external command was terminated by signal".into()
+        };
+        return Err(CommandError::new(CommandErrorKind::User, error_msg));
```
yuja, line 287: nit: use `user_error()`

PhilipMetzger:
> @PhilipMetzger who should give their approval for consensus to be achieved? Are there remaining open questions?

We have no final decision if we want to support a `jj util exec` command or pursue @matts1's work in #3673 (#4605) further. While this is definitely a step-up, I'd rather avoid such a capability until we have a agreed upon solution. There's a reason why the discussion in the issue became dormant. 

> The discussion there revolves around some kind of custom, lightweight, embedded scripting language. But the original use case is just streamlining workflows, which is fully covered by this PR. Nothing gives the user more flexibility and is less of a maintenance burden to us than letting users bring their own scripting language.

We've also had a few talks about having a embeddedable scripting language with `jj script`, so I think it is the favored approach.

senekor:
> if we want to support `jj util exec` command

There's not much to support, the essence of this feature is a one-liner.

> I'd rather avoid such a capability until we have a agreed upon solution

What's the harm in having both? Even if an embedded scripting language ends up materializing at some point, this trivial feature does not impose any additional maintenance burden.

jj is not at 1.0 yet, this feature could even be removed if an alternative, preferred solution has become mature enough.

I'm personally not in favor of an embedded scripting language. I don't see any benefit compared to this simple approach. On the contrary, it is a huge amount of work and maintenance burden. It will never be as flexible as shelling out. Even if we embed the most-amazing-and-objectively-superior-to-any-other-scripting-language TM (unlikely), we will still be forcing people to learn yet-another-language instead of letting them use the tools they know and are good at.

I don't want to have to argue against it though, that's no fun. I'd rather just we have both, to which there are no downsides.

thoughtpolice:
> I don't see any benefit compared to this simple approach.

The most important one are the subtle downstream consequences. Scripts can be pure and deterministic by default, so you can be sure that rerunning commands is either idempotent or fully reversible through `jj undo`; many things that e.g. reword commits, check commit messages, or any number of things that are pure operations on the commit graph are great for this. Scripts can be used on every platform, because they don't need any configuration changes and can be portable by default; a subtle and non-obvious consequence of your example above is that `/path/to/foo` on your Linux machine is called `C:\\path\\to\\foo.exe` on my machine. This means users can no longer use a trivial single configuration file but instead has to have multiple configurations for each platform (or we need to introduce like platform-conditional maps in our TOML files, `[platform.x86_64-unknown-linux-gnu]` like cargo... Ugh.)[1]

But also, scripts can be much more secure by default; they can for example be given no access to the network or any other number of features, can invoke UI prompts or wizards to the user, etc. That also means scripts can easily be checked into the source repository without fear that random things will get run without review. In the future, I also hope that a `jj-script` mechanism could work on the server side, and also inside your IDE. That means that you could write a script that shows a prompt *in visual studio code* instead, or one that runs a script on the server-side for your code review tool, etc. "Randomly invoke thing out of `$PATH`" will never clear a security review for many cases and is a complete non-starter anyway in things like a server-side component.

So, there are many conceptual advantages. I have a private prototype but it's very slow going because I'm busy a lot right now. But I want to be clear that an unintended consequence of just giving people access to the lowest-common-denominator fast-access dopamine-hit tool &mdash; in this case, random shell scripts &mdash; is that they often will suck the air out of the room for alternative tools, and they will often miss many subtle things like the ones I listed above (e.g. requiring some monstrous Git bash installation, or just not working on Windows at all.)

[1] Actually, the cross-platform thing extends to many other examples, but this is the most obvious. The reason I did not just write a wrapper `jj-gerrit` command is exactly because I wanted to use my config and this tool on all my machines, and I use Windows too. Therefore, I have to either write the tool in a full blown language anyway, completely negating the fast scripting advantage, or write it twice, etc etc...

senekor:
> Scripts can be pure and deterministic by default, so you can be sure that rerunning commands is either idempotent or fully reversible through `jj undo`

That sounds legitimately neat. But framed this way, there is really very little overlap between these two features and therefore even less reason to pit them against each other. A scripting language that only performs pure operations on the commit graph may have a bunch of benefits, but it is also extremely limited by definition in the context of use cases that are fundamentally impure. What if I want a command `jj lfs-sync` or something? That's not pure. But it fits into the workflow of users of a VCS. It's useful. Let's have both.

> `/path/to/foo` on your Linux machine is called `C:\\path\\to\\foo.exe` on my machine
>
> I wanted to use my config and this tool on all my machines, and I use Windows too. Therefore, I have to either write the tool in a full blown language anyway, completely negating the fast scripting advantage, or write it twice, etc etc...

The fact that `jj util exec` doesn't single-handedly resolve all differences between unix and windows doesn't mean that it isn't useful or doesn't pull its weight in maintenance burden. A pure, jj-specific scripting language will not solve these problems either, so it's simply an irrelevant point.

The main use case for this feature is for individuals to streamline their own workflows. Few people use unix and windows regularly. Which doesn't mean we should dismiss their needs, but there are purpose-built tools to solve these problems. For example, I use [chezmoi](https://www.chezmoi.io/) to manage my dotfiles, including my jj-config and therefore any such aliases. Chezmoi has a templating system, which I use for differences between workstations & servers, private & company machines. You could use it to paper over differences between unix and windows, like path syntax. I _assume_ your `jj-gerrit` command is not the only piece of workflow automation you would like to share across your machines. In that case, `jj-script` won't solve your problem and you'll have to reach for another tool like chezmoi anyway (or just live with the inconvenience)

That's for individuals, what about teams working on a project? Chezmoi is not fit to solve that problem. But neither is a jj-specific scripting language. Every project has tools, scripts, automation and whatnot that go along with the development process. Here again, there are purpose-built tools to solve these problems. Some may specifically seek out platform-independent tools like `just` and `nushell` or `python` instead of `make` and `bash`. Some may tell their windows coworkers to use WSL. Some may stuff all their tooling in docker containers. Whatever a project ends up doing, `jj util exec` will be perfectly compatible with it.

My main point here is that "solving the problem of platform incompatibility" and "streamlining VCS workflows" are two orthogonal problems and `jj util exec` shouldn't be arbitrarily held to the standard of solving both of them.

> [scripts] can for example be given no access to the network or any other number of features
>
> That also means scripts can easily be checked into the source repository without fear that random things will get run without review. In the future, I also hope that a `jj-script` mechanism could work on the server side, and also inside your IDE.
>
> "Randomly invoke thing out of $PATH" will never clear a security review for many cases and is a complete non-starter anyway in things like a server-side component.

Like purity, these secure-by-default features sound legitimiately useful. This definitely makes me more excited to see how the scripting endeavor develops.

But I don't think it has anything to do with `jj util exec`. Just don't let alisas be checked into version control like they already can't. That doesn't prevent pure scripts from being checked in VC in the future under a different system than aliases. These things have nothing to do with each other and I don't think they should be pitted against each other.

> we will still be forcing people to learn yet-another-language instead of letting them use the tools they know and are good at.

I agree that by your description of a pure, secure scripting language we would have to make an opinionated choice anyway so we can precisely control the API. So I don't mind "forcing" a specific language in that case. I _would_ mind if the scripting language ended up having all the capabilities of a regular shell language and I would be forced to use that specific one for no reason. But that doesn't seem to be the plan, so my original point is mute.

---

I'll repeat my main point, which _isn't_ that we shouldn't do the scripting language. The scripting language sounds legitimately useful and I'm excited for it now. My main point is that there is no reason for these two features to be mutually exclusive. They are different features, serving different purposes, and they can live side-by-side without any issues.

yuja:
I'm not against adding `jj util exec` or something similar even if we're going to add a full-blown scripting feature at some point. It's simple and wouldn't conflict with any future alias syntax changes.

thoughtpolice:
> The scripting language sounds legitimately useful and I'm excited for it now

Just to be clear, my only point was to emphasize that a scripting solution can and does have conceptual benefits over just calling out to random scripts in `$YOUR_LANGUAGE`. I don't disagree that the `util exec` feature is complementary (though I stand by my point it sucks some air out of the room because it's deceptively simple but has some flaws, but there are worse things in life and it does the job.)

Actually, this approach of requiring a subcommand does fix at least one of my gripes about "just execute things from `$PATH`", which is that you don't really know what anything is. But this provides a way to do something like `jj util exec my-thing` and then set an alias in your *user config* that sets `aliases.my-thing = ["util", "exec", "my-thing.exe"]`, which is 99% of what people want. In other words, `jj` isn't executing random commands out of thin air, it only does what you ask, and some aliases can make the syntax shorter. Which I think is an OK compromise.

docs/config.md
```diff
@@ -528,6 +528,28 @@ You can define aliases for commands, including their arguments. For example:
 aliases.l = ["log", "-r", "(main..@):: | (main..@)-"]
 ```
 
+If you would like to create an alias that's more complicated than a short form
+of a single jj command, you can use `jj util exec` in your alias to execute
+arbitrary external commands. This is analogous to `git`'s feature where
+executables in you `$PATH` starting with `git-*` can be called implicitly with
+`git *`. Here are some examples:
```
PhilipMetzger, line 535: I would really like to mention some caveats here, as some scripts may break `jj op undo` and can cause irreparable damage to the repo. 

Another point to mention here, that we may subsume this feature with an actual scripting language in the future.

cli/src/commands/util/exec.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Execute an external command via jj
+///
+/// This is useful for arbitrary aliases. For example, assume you have a script
+/// called "my-jj-script" in you $PATH and you would like to execute it as "jj
+/// my-script". You would add the following line to your configuration file to
+/// achieve that:
+///
```
PhilipMetzger, line 26: Add a warning section here, which mentions the caveats from my other comments.

PhilipMetzger (approved):
> > if we want to support `jj util exec` command
> 
> There's not much to support, the essence of this feature is a one-liner.

There's is more to support, beyond the technical implementation such as user expectations, which is the reason why I'm so split on the feature. 

> > I don't see any benefit compared to this simple approach.
> 
> The most important one are the subtle downstream consequences. Scripts can be pure and deterministic by default, so you can be sure that rerunning commands is either idempotent or fully reversible through `jj undo`; many things that e.g. reword commits, check commit messages, or any number of things that are pure operations on the commit graph are great for this. Scripts can be used on every platform, because they don't need any configuration changes and can be portable by default; a subtle and non-obvious consequence of your example above is that `/path/to/foo` on your Linux machine is called `C:\\path\\to\\foo.exe` on my machine.

Austin also really has a great point with this, and we may allow a checked in config file in the future (#3684) which will make this a major pitfall. 

> > `/path/to/foo` on your Linux machine is called `C:\\path\\to\\foo.exe` on my machine
> > I wanted to use my config and this tool on all my machines, and I use Windows too. Therefore, I have to either write the tool in a full blown language anyway, completely negating the fast scripting advantage, or write it twice, etc etc...
> 
> The fact that `jj util exec` doesn't single-handedly resolve all differences between unix and windows doesn't mean that it isn't useful or doesn't pull its weight in maintenance burden. [...]
> 
> The main use case for this feature is for individuals to streamline their own workflows. [...] For example, I use [chezmoi](https://www.chezmoi.io/) to manage my dotfiles, including my jj-config and therefore any such aliases. Chezmoi has a templating system, which I use for differences between workstations & servers, private & company machines. You could use it to paper over differences between unix and windows, like path syntax. I _assume_ your `jj-gerrit` command is not the only piece of workflow automation you would like to share across your machines. In that case, `jj-script` won't solve your problem and you'll have to reach for another tool like chezmoi anyway (or just live with the inconvenience)
> 
> That's for individuals, what about teams working on a project? Chezmoi is not fit to solve that problem. But neither is a jj-specific scripting language. Every project has tools, scripts, automation and whatnot that go along with the development process. Here again, there are purpose-built tools to solve these problems. Some may specifically seek out platform-independent tools like `just` and `nushell` or `python` instead of `make` and `bash`. Some may tell their windows coworkers to use WSL. Some may stuff all their tooling in docker containers. Whatever a project ends up doing, `jj util exec` will be perfectly compatible with it.

I disagree with your point here, as giving the capability to `jj` makes it _implicitly supported_ which is something we cannot control. A general purpose scripting language built on the base Austin mentioned, lets us _control the behavior_ which is desirable, as we can move users off of bad patterns and deprecate and remove bad APIs and assumptions. So it is **definitely** not a orthogonal feature for a potential feature, which I'm once again weary of. 

> > The scripting language sounds legitimately useful and I'm excited for it now
> 
> Just to be clear, my only point was to emphasize that a scripting solution can and does have conceptual benefits over just calling out to random scripts in `$YOUR_LANGUAGE`. I don't disagree that the `util exec` feature is complementary (though I stand by my point it sucks some air out of the room because it's deceptively simple but has some flaws, but there are worse things in life and it does the job.)

Otherwise I agree with both Austin and Yuya and will accept the feature with some caveats and a warning in the documentation. This was not my favorite decision 😐.

senekor:
In case I came across as too pushy, my goal is not to get this PR merged _before_ reaching concensus. Thank you for stating your concerns, I don't want to merge this before they are addressed.

> > [...] Scripts can be used on every platform, because they don't need any configuration changes and can be portable by default; a subtle and non-obvious consequence of your example above is that `/path/to/foo` on your Linux machine is called `C:\\path\\to\\foo.exe` on my machine.
> 
> Austin also really has a great point with this, and we may allow a checked in config file in the future (#3684) which will make this a major pitfall.

I tried to address this with the following:

> Just don't let alisas be checked into version control like they already can't. That doesn't prevent pure scripts from being checked in VC in the future under a different system than aliases.

Whatever we allow to be checked into VC in the future, it won't make sense to check in an entire jj config, right? Are we going to allow checking in `user.email`? Admittedly, even if that doesn't make sense, it's definitely a smaller problem than arbitrary code execution.

Either way, do you think it would be problematic to differentiate explicitly about what kind of configuration can and cannot be checked into VC?

> > Whatever a project ends up doing, `jj util exec` will be perfectly compatible with it.
>
> I disagree with your point here, as giving the capability to `jj` makes it _implicitly supported_ which is something we cannot control.

What do you think we would be _implicitly supporting_ with this feature? In my mind, what we support would be a convenient syntax to shell out to the environment. What the environment does with that is not our concern. It could only turn into a problem for us if users are somehow confused about this, but I'm having a hard time imagining what would cause this confusing.

I guess this goes into your concern about user expectations. I haven't thought about the expectation of jj commands always being `jj undo`-able. Aren't there already commands that break this assumption? I can't undo `jj git push`. What about `jj abandon -r '..' --ignore-immutable` + `jj util gc --expire now`? I checked, and it doesn't actually nuke the op log. But judging from the docs, I expect it can and will do that in the future. `jj util exec -- rm -rf .jj` is just a little more reliable, and I can't imagine a user being surprised by it.

I'm happy to make it very clear in the documenatation that this is just a convenient way to run arbitrary code on your system (including all the danger) which may not even be related to jj.

> A general purpose scripting language built on the base Austin mentioned, lets us _control the behavior_ which is desirable, as we can move users off of bad patterns and deprecate and remove bad APIs and assumptions. So it is **definitely** not a orthogonal feature for a potential feature, which I'm once again weary of.

I still feel like this hypothetical scripting language we talk about changes its promised feature set depending on the context of the conversation. We need to be on the same page about what we're talking about.

a) A **pure and secure** scripting language. This has several potential benefits over shelling out. I hope this will become reality. However, this _is_ orthogonal to shelling out, because it doesn't cover many use cases that shelling out does. (e.g. `jj lfs-sync`) I stand by my point that these two features would complement each other, they aren't mutually exclusive.

b) A **general purpose** scripting language that can do most if not all the things bash can do. Certainly, such a feature would make `jj util exec` redundant. But it has none of the benefits of a pure and secure scripting language while imposing a massive maintenance burden compared to shelling out as well as forcing users to learn and use a specific thing unnecessarily. If the language let's me make HTTP requests, I will run a server on localhost that shells out to whatever it finds in the query string. If the language let's me read from the file system, I will mount my own FUSE that calls any command encoded in a file path that my script attempts to read from. (Is that possible? Just came up with it. Have to check.) My point being, the _secure_ part of the scripting language is extremely limiting in terms of functionality and use cases.

On a side note, are you also concerned about [extension support](https://github.com/martinvonz/jj/issues/3575) for the same reasons? As I pointed out in [this comment](https://github.com/martinvonz/jj/issues/3575#issuecomment-2455434624), the proposal is essentially a superset of `jj util exec`.

---

We can always remove `jj util exec` in the future. The worst that will happen is that we'll have to tell users to type `jj-lfs-sync` instead of `jj lfs-sync`. I don't think we'll need to do that, but it wouldn't be the end of the world.

matts1:
As the person who filed #3673, I'd like to express my support for `jj util exec` as an alternative to my proposal. I do think that my proposal had issues with either not being expressive enough, or adding too much complexity into the DSL.

FWIW, I'm a big fan of the starlark choice. I'm not a huge fan of arbitrary binaries, especially given that if it's a compiled language, it'll be platform-specific.

> a) A **pure and secure** scripting language. This has several potential benefits over shelling out. I hope this will become reality. However, this _is_ orthogonal to shelling out, because it doesn't cover many use cases that shelling out does. (e.g. `jj lfs-sync`) I stand by my point that these two features would complement each other, they aren't mutually exclusive.
> 
> b) A **general purpose** scripting language that can do most if not all the things bash can do. Certainly, such a feature would make `jj util exec` redundant. But it has none of the benefits of a pure and secure scripting language while imposing a massive maintenance burden compared to shelling out as well as forcing users to learn and use a specific thing unnecessarily. If the language let's me make HTTP requests, I will run a server on localhost that shells out to whatever it finds in the query string. If the language let's me read from the file system, I will mount my own FUSE that calls any command encoded in a file path that my script attempts to read from. (Is that possible? Just came up with it. Have to check.) My point being, the _secure_ part of the scripting language is extremely limiting in terms of functionality and use cases.

Personally, I'd be in favor of a solution where only support pure and secure scripting language, and similar to bazel's `repository_ctx.execute`, we provide some kind of mechanism to do those general purpose things. This would allow us to guarantee "if this function isn't called, then it's hermetic".

cli/src/commands/util/exec.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright 2020 The Jujutsu Authors
```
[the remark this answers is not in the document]
PhilipMetzger: this is a new file, so the copyright statement should only contain 2024

cli/src/commands/util/exec.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// Execute an external command via jj
+///
+/// This is useful for arbitrary aliases.
+///
+/// !! WARNING !!
+///
+/// The following technique just provides a convenient syntax for running
+/// arbitrary code on your system. Using it irresponsibly may cause damage
+/// ranging from breaking the behavior of `jj op undo` to wiping your file
```
PhilipMetzger, line 28: Nit: Most users only know about `jj undo`, so use it here.

docs/config.md
```diff
@@ -528,6 +528,38 @@ You can define aliases for commands, including their arguments. For example:
...
+This alias syntax can only run a single jj command. However, you may want to
+execute multiple jj commands with a single alias, or run arbitrary scripts that
+complement your version control workflow. This can be done, but be aware of the
+danger:
+
+!!! warning
+
+    The following technique just provides a convenient syntax for running
+    arbitrary code on your system. Using it irresponsibly may cause damage
+    ranging from breaking the behavior of `jj op undo` to wiping your file
```
PhilipMetzger, line 540: Ditto

PhilipMetzger (commented):
Another round of minor nits.

PhilipMetzger:
Sorry for the late response here. Now onto my thoughts:

> > > [...] Scripts can be used on every platform, because they don't need any configuration changes and can be portable by default; a subtle and non-obvious consequence of your example above is that `/path/to/foo` on your Linux machine is called `C:\\path\\to\\foo.exe` on my machine.
> > 
> > 
> > Austin also really has a great point with this, and we may allow a checked in config file in the future (#3684) which will make this a major pitfall.
> 
> I tried to address this with the following:
> 
> > Just don't let alisas be checked into version control like they already can't. That doesn't prevent pure scripts from being checked in VC in the future under a different system than aliases.
> 
> Whatever we allow to be checked into VC in the future, it won't make sense to check in an entire jj config, right?

There probably won't be a full config checked in, but a _common_ set of aliases, which then leads to a chicken and egg problem if you have any alias depending on `jj util exec` without a checked in script. 

> Are we going to allow checking in `user.email`?

Depends on how you look at it. If we only want a common configuration then it could be disallowed but I assume some smart users will use the feature to configure a bot and its config. 

> Admittedly, even if that doesn't make sense, it's definitely a smaller problem than arbitrary code execution.

Yes, which `jj util exec` gracefully avoids. 

> Either way, do you think it would be problematic to differentiate explicitly about what kind of configuration can and cannot be checked into VC?

Any suggestion we make in the docs will be understood as a Guideline not something we actively prevent, so as soon as you have a user which ignores it, you're bound by Hyrum's Law. Preventing it in code will also be hard as long as there is no IFTT (If-This-Then-That) linter for it. 

> > > Whatever a project ends up doing, `jj util exec` will be perfectly compatible with it.
> > 
> > 
> > I disagree with your point here, as giving the capability to `jj` makes it _implicitly supported_ which is something we cannot control.
> 
> What do you think we would be _implicitly supporting_ with this feature?

Basically anything, as long as it runs. As it is "supported" now, a user will come and scream at us, when we remove their favorite plaything. 

> In my mind, what we support would be a convenient syntax to shell out to the environment. What the environment does with that is not our concern. It could only turn into a problem for us if users are somehow confused about this, but I'm having a hard time imagining what would cause this confusing.

This is not entirely wrong but does not represent the user expectation that `jj util exec` for a custom alias must continue to work until the end of eternity, which hits the issue of #3549 and its removal/supersedence with #4496 even though we advertised the feature as purely experimental (relevant [Discord Thread](https://discord.com/channels/968932220549103686/1293592969676460104)). 

>[..] I haven't thought about the expectation of jj commands always being `jj undo`-able. Aren't there already commands that break this assumption? I can't undo `jj git push`. What about `jj abandon -r '..' --ignore-immutable` + `jj util gc --expire now`?

Currently the Git backend/store and related functionality break that currently. In the future with a native store and server for it, we should strive for making every comand/operation cleanly revertible. 

> I checked, and it doesn't actually nuke the op log.

The op log is a separate thing, see https://github.com/martinvonz/jj/issues/3700#issuecomment-2170519215 for a nice explanation of it. And as you noted at some point `util gc` may garbage collect it too.  

> > A general purpose scripting language built on the base Austin mentioned, lets us _control the behavior_ which is desirable, as we can move users off of bad patterns and deprecate and remove bad APIs and assumptions. So it is **definitely** not a orthogonal feature for a potential feature, which I'm once again weary of.
> 
> I still feel like this hypothetical scripting language we talk about changes its promised feature set depending on the context of the conversation. We need to be on the same page about what we're talking about.
> 
> a) A **pure and secure** scripting language. This has several potential benefits over shelling out. I hope this will become reality. However, this _is_ orthogonal to shelling out, because it doesn't cover many use cases that shelling out does. (e.g. `jj lfs-sync`) I stand by my point that these two features would complement each other, they aren't mutually exclusive.
> 
> b) A **general purpose** scripting language that can do most if not all the things bash can do. Certainly, such a feature would make `jj util exec` redundant. But it has none of the benefits of a pure and secure scripting language while imposing a massive maintenance burden compared to shelling out as well as forcing users to learn and use a specific thing unnecessarily. If the language let's me make HTTP requests, I will run a server on localhost that shells out to whatever it finds in the query string. If the language let's me read from the file system, I will mount my own FUSE that calls any command encoded in a file path that my script attempts to read from. (Is that possible? Just came up with it. Have to check. (Edit: [it works](https://git.buenzli.dev/remo/fush))) My point being, the _secure_ part of the scripting language is extremely limiting in terms of functionality and use cases.

Sorry for badly expressing myself here, I just meant that more "generally purpose" functionality can be built on top of the Starlark language for Revsets which we then could use to run something like a CI on the server. Otherwise I agree with Matt's point and your definitions. 

> On a side note, are you also concerned about [extension support](https://github.com/martinvonz/jj/issues/3575) for the same reasons?

Yes, although they only operate in a universe where we call the shots (as we define the extension points).

senekor:
> > Whatever we allow to be checked into VC in the future, it won't make sense to check in an entire jj config, right?
> 
> There probably won't be a full config checked in, but a _common_ set of aliases, which then leads to a chicken and egg problem if you have any alias depending on `jj util exec` without a checked in script.

I was thinking along the lines of the `[aliases]` section in the config being completely forbidden from being checked in, whereas a new section `[scripts]` could be checked in, because it is safe. I'm assuming the scripting language would be at least as powerful as current aliases, so there would be no need to check in the `[aliases]` section.

(We'd have to make a decision abouot precedence if there is an alias and a script with the same name, but that doesn't seem like too big of an issue.)

> > Either way, do you think it would be problematic to differentiate explicitly about what kind of configuration can and cannot be checked into VC?
> 
> Any suggestion we make in the docs will be understood as a Guideline not something we actively prevent, so as soon as you have a user which ignores it, you're bound by Hyrum's Law. Preventing it in code will also be hard as long as there is no IFTT (If-This-Then-That) linter for it.

I agree that documentation is not good enough, I was thinking about preventing it in code. What is difficult about this approach? If the config is checked into VCS and it contains a `[aliases]` section, ignore it and emit a warning or even hard error.

> > What do you think we would be _implicitly supporting_ with this feature?
> 
> Basically anything, as long as it runs. As it is "supported" now, a user will come and scream at us, when we remove their favorite plaything.

> the user expectation that `jj util exec` for a custom alias must continue to work until the end of eternity

I don't think we have to worry about that. As mentioned, transitioning from `jj my-script` to `jj-my-script` is a pretty mild breaking change.

> > On a side note, are you also concerned about [extension support](https://github.com/martinvonz/jj/issues/3575) for the same reasons?
> 
> Yes, although they only operate in a universe where we call the shots (as we define the extension points).

What do you mean by "we call the shots"? If the method of invoking these extensions is shelling out to them, they are going to have all the capabilities of a bash script _plus_ any extension points we provide.

docs/config.md
PhilipMetzger: > I forgot to mention that the feature may be removed, added that now:
> 
> > This feature may be removed entirely in the future. It may be replaced by an embedded scripting language with limited capabilities for safety.

This can be simplified to "This feature may be removed or replaced by an embedded scripting language in the future".

PhilipMetzger:
> > > Whatever we allow to be checked into VC in the future, it won't make sense to check in an entire jj config, right?
> > 
> > 
> > There probably won't be a full config checked in, but a _common_ set of aliases, which then leads to a chicken and egg problem if you have any alias depending on `jj util exec` without a checked in script.
> 
> I was thinking along the lines of the `[aliases]` section in the config being completely forbidden from being checked in, whereas a new section `[scripts]` could be checked in, because it is safe. I'm assuming the scripting language would be at least as powerful as current aliases, so there would be no need to check in the `[aliases]` section.

That definitely could work but at this point we're already far into a hypothetical bikeshed so I'd rather drop it and move on. 

> > > Either way, do you think it would be problematic to differentiate explicitly about what kind of configuration can and cannot be checked into VC?
> > 
> > 
> > Any suggestion we make in the docs will be understood as a Guideline not something we actively prevent, so as soon as you have a user which ignores it, you're bound by Hyrum's Law. Preventing it in code will also be hard as long as there is no IFTT (If-This-Then-That) linter for it.
> 
> I agree that documentation is not good enough, I was thinking about preventing it in code. What is difficult about this approach?

You need to catch it at Codereview time, which isn't easy if you don't have a good infra for it. This also needs to encompass renames and more, with the thought in mind that we'll always add more config flags to `jj` keeping in mind that we're at the start of the project. So it's a social rather than technical issue.  

> > > What do you think we would be _implicitly supporting_ with this feature?
> > 
> > 
> > Basically anything, as long as it runs. As it is "supported" now, a user will come and scream at us, when we remove their favorite plaything.
> 
> > the user expectation that `jj util exec` for a custom alias must continue to work until the end of eternity
> 
> I don't think we have to worry about that. As mentioned, transitioning from `jj my-script` to `jj-my-script` is a pretty mild breaking change.

I still respectfully disagree and even gave you a example. 

> > > On a side note, are you also concerned about [extension support](https://github.com/martinvonz/jj/issues/3575) for the same reasons?
> > 
> > 
> > Yes, although they only operate in a universe where we call the shots (as we define the extension points).
> 
> What do you mean by "we call the shots"? If the method of invoking these extensions is shelling out to them, they are going to have all the capabilities of a bash script _plus_ any extension points we provide.

See #3869 (visit the Design Doc) and #3577 for the long discussions around the actual suggested feature here, where I already spent many hours discussing a agreeable design . I do not have the capacity to give you a TL;DR right now, as it's just to much content.

At this point I'm ready to drop my concerns, as we discussed a potential future and migration plan for the feature, its major drawbacks and warned users about it, which is a major step-up from other existing systems. And since it solves a bunch of user problems we should accept it on the ground of "perfect is the enemy of good" while a better design is in the works.

And on a side-note, as I forgot that part in my last reply:

> In case I came across as too pushy, my goal is not to get this PR merged _before_ reaching concensus. Thank you for stating your concerns, I don't want to merge this before they are addressed.

While it was a bit pushy, the discussion never got heated which I really value after having a bunch of heated discussions in other issues or with other users. 

And feel free to solve conversations after you addressed them.

cli/src/commands/util/exec.rs
```diff
@@ -0,0 +1,88 @@
...
+
+pub fn cmd_util_exec(
+    _ui: &mut Ui,
+    _command: &CommandHelper,
+    args: &UtilExecArgs,
+) -> Result<(), CommandError> {
+    let status = std::process::Command::new(&args.command)
+        .args(&args.args)
+        .status()
+        .map_err(user_error)?;
```
yuja, line 78: nit: `user_error_with_message("Failed to execute external command {..}", err)`? The default `io::Error` message lacks context.

cli/src/commands/util/exec.rs
```diff
@@ -0,0 +1,88 @@
...
+    if !status.success() {
+        let error_msg = if let Some(exit_code) = status.code() {
+            format!("External command exited with {exit_code}")
+        } else {
+            "External command was terminated by signal".into()
```
yuja, line 83: nit: use `{status}` which includes signal number, etc.

cli/tests/test_util_command.rs
```diff
@@ -112,3 +112,33 @@ fn test_shell_completions() {
...
+}
+
+#[test]
+fn test_util_exec_fail() {
+    let test_env = TestEnvironment::default();
+    let err = test_env.jj_cmd_failure(
+        test_env.env_root(),
+        &["util", "exec", "--", "missing-program"],
+    );
+    assert!(!err.is_empty());
```
yuja, line 143: nit: use `insta::assert_snapshot!()` with `strip_last_line()`? It's easier to follow in case of an error.

yuja (approved):

Issue #1039, opened 2023-01-14
Implement `jj sync` command


dbarnett:
jj should have a command to sync all of my branches (or some subset) with a remote via merge or rebase, in a single operation.

Since that could cause a lot of big changes at once, it'd be nice to show a preview of changes that will be performed and prompt to confirm and/or output a very clear summary of what all changed from that sync operation.

dbarnett:
I could also imagine some related curation tools for e.g. moving a local branch to "draft mode" to not be published or synced yet.

dbarnett:
This is also one of the areas jj was dinged on in the comparison table at https://blog.waleedkhan.name/git-ui-features/#data.

MForster:
IMHO `sync` is ambiguous about the direction. Does it only pull or also push?

PhilipMetzger:
I think `jj sync` should only pull, like its [Perforce predecessor](https://www.perforce.com/manuals/cmdref/Content/CmdRef/p4_sync.html). If that isn't clear enough, we can alias the command for Git users to `jj pull`

martinvonz:
> IMHO `sync` is ambiguous about the direction. Does it only pull or also push?

Hmm, that's a good point. On the other hand, I don't think we've had any confusion about that at Google with our `hg sync` command, which does pull+rebase. I'm not sure what a better name would be.

> we can alias the command for Git users to `jj pull`

I think Git's use of "pull" for "fetch+rebase/merge" is quite confusing. I prefer hg's use of `hg pull` for just pulling (like `git fetch`). That is more symmetric with `hg/git push`.

MForster:
> I don't think we've had any confusion about that at Google with our `hg sync` command

That's not representative, because ~everybody at Google has used Perforce/Piper.

FWIW, VSCode has a "Sync" button that pulls and pushes...

martinvonz:
> That's not representative, because ~everybody at Google has used Perforce/Piper.

True. (I have practically not used it, but I still have run `p4 sync` several times.)

> FWIW, VSCode has a "Sync" button that pulls and pushes...

Oh, that's a good data point!

We'd still have to think of a better name then... There's `hg fetch` that does this. I don't know if they named it that as a joke, given that it means that `hg pull` == `git fetch` and `hg fetch` == `git pull`. So I don't really think we should call it `jj fetch`.

PhilipMetzger:
[...]

PR #4767, opened 2024-11-04
git fetch: Refactor a reusable `git_fetch` function from the command.

essiene:
We will need this in the `jj git sync` command to perform a fetch that behaves the same as the `jj git fetch` command.

Fetch extract the function in this commit, so diffs are easier to review.

A later commit will move this function (and associated functions) out to cli/src/git_util.rs

Part of: #1039

# Checklist

If applicable:


cli/src/git_util.rs
```diff
@@ -433,3 +456,82 @@ export or their "parent" bookmarks."#,
...
+) -> Result<(), CommandError> {
+    for remote in remotes {
+        let stats = with_remote_git_callbacks(ui, None, |cb| {
+            git::fetch(
+                tx.repo_mut(),
+                git_repo,
+                remote,
+                branch,
+                cb,
+                &settings.git_settings(),
```
yuja, line 476: nit: `settings` could be obtained from `tx.settings()` to reduce number of arguments.

yuja (approved):

cli/src/git_util.rs
essiene: Ah nice. Done. Thanks.

PR #4768, opened 2024-11-04
cli: Refactor out `get_fetch_remotes` into git_util.rs

essiene:
Similarly to `git_fetch`, this will be used in `jj git sync`, so moving this out to maintain the same behaviour across `fetch` command and `sync`.

Issue: #1039


cli/src/git_util.rs
```diff
@@ -535,3 +544,40 @@ fn warn_if_branches_not_found(
...
+    use_all_remotes: bool,
+) -> Result<Vec<String>, CommandError> {
+    if use_all_remotes {
+        Ok(git_repo
+            .remotes()?
+            .iter()
+            .filter_map(|x| x.map(ToOwned::to_owned))
+            .collect())
+    } else if !remotes.is_empty() {
+        Ok(remotes.iter().cloned().collect_vec())
```
yuja, line 563: nit: `.to_vec()`

cli/src/git_util.rs
```diff
@@ -535,3 +544,40 @@ fn warn_if_branches_not_found(
...
     Ok(())
 }
+
+const DEFAULT_REMOTE: &str = "origin";
+pub fn get_fetch_remotes(
+    ui: &Ui,
+    settings: &UserSettings,
+    git_repo: &git2::Repository,
+    remotes: &[String],
+    use_all_remotes: bool,
```
yuja, line 554: If both `jj git sync` and `jj git fetch` share command parameters, it might be better to extract common `Args` struct. Then, we can pass it in to helper functions.

yuja (approved):

cli/src/git_util.rs
essiene: Done.

cli/src/git_util.rs
essiene: Makes sense.

I'm thinking this could actually collapse down to something like:

```rust
let fetcher = GitFetcher::new(ui, tx, git_repo, branch, remotes, use_all_remotes);
fetcher.fetch()?;
```

Inside `fetcher.fetch` we'd grab the remotes and then do the fetch. Mainly because I can't think of a good reason why to expose to function calls to the commands instead of just one, as the intermediate result is not used by callers.

Does that feel like overkill?

In the meantime, I'm thinking to merge this PR and make changes in a follow-up to keep things clearer.

Let me know if you think otherwise.

yuja: I'm not sure. Do we have more methods than `fetch()`?

My concern about the current API is that it's hard to tell which `String`/`bool` argument represents which variable.

essiene: > I'm not sure. Do we have more methods than `fetch()`?

I just realized we do need to expose `get_fetch_remotes`, since we use the result to build a message after the fetch succeeds further down.

> 
> My concern about the current API is that it's hard to tell which `String`/`bool` argument represents which variable.

I've added a FetchArgs struct that should make this better and doesn't end up feeling over designed.

wdyt?

cli/src/git_util.rs
```diff
@@ -456,30 +466,36 @@ export or their "parent" bookmarks."#,
     Ok(())
 }
 
+pub struct FetchArgs<'a> {
+    pub repo: &'a git2::Repository,
```
yuja, line 470: I wouldn't include `git2::Repository` in args struct because it's not a user parameter and has different lifetime.

BTW, what I meant was to extract common `clap::Args` struct from `GitFetchArgs`. If `jj git sync` shares the same help messages, we can "flatten" the common Args struct into `GitFetch`/`SyncArgs`. If not, we'll need a separate parameter struct as you did.

essiene: > I wouldn't include `git2::Repository` in args struct because it's not a user parameter and has different lifetime.

Removed.

> 
> BTW, what I meant was to extract common `clap::Args` struct from `GitFetchArgs`. If `jj git sync` shares the same help messages, we can "flatten" the common Args struct into `GitFetch`/`SyncArgs`. If not, we'll need a separate parameter struct as you did.

I've gone to flesh out a first draft of the sync command to be sure we need to share the help messages and yeah, it makes sense to share them. So I've extracted these out and flattened them back into `GitFetchArgs`.

Is this what you meant?

cli/src/commands/git/fetch.rs
```diff
@@ -54,52 +41,22 @@ pub fn cmd_git_fetch(
...
-    git_fetch(ui, &mut tx, &git_repo, &remotes, &args.branch)?;
+    git_fetch(
+        ui,
+        &mut tx,
+        &git_repo,
+        &FetchArgs {
+            branch: args.fetch.branch.clone(),
+            remotes: remotes.clone(),
+            all_remotes: args.fetch.all_remotes,
+        },
```
yuja, line 55: Oh, we can't use the original `args.fetch`? That makes `args.fetch` useless..

yuja: Maybe `FetchRemoteArgs` can be extracted instead of all-in `FetchArgs`, but I'm not pretty sure if that makes sense. Sorry for the noise.

essiene: No worries at all, I may have jumped the gun on the Refactor before finalizing how it will be used. For `sync` it turns out we need to fetch all branches for all configured remotes[1], so the help text will be different after all, and I don't need to flatten them in. I've reverted that last change, back to just passing FetchArgs{} around with branch, remotes and all_remotes as the fields.


[1] - https://docs.google.com/document/d/1K4j1xQPFxtINhyOyKuCMTZ_iNrw8g-5_giBFMGoLjaI/edit?tab=t.0#bookmark=id.y91cr92o9czf

yuja: Let's not add `branch` to the args struct then. It doesn't make sense to pass `branch` in to `get_fetch_remotes()`, and `all_remotes` to `git_fetch()`.

It might also be simpler to just make both `get_all_remotes()` and `get_default_fetch_remotes()` public.

essiene: So I think I finally have something that looks descent:

* `git_fetch` takes `FetchArgs`; which only has fields for `branch` and `remotes`. This is useful since FetchArgs disambiguated the fields and makes it easier to know what goes where.
* `get_fetch_remotes` doesn't need `FetchArgs` and since `remotes` and `all_remotes` are different types completely, it doesn't look terrible if I just pass them in directly

wdyt?

yuja: Seems okay.

FWIW, I wonder if `git_fetch()` can actually be reused in "git sync". Suppose "git sync" does rebase based on the stats returned by `git::fetch()`, "git sync" will have to do lower-level API.

Regarding `args.all_remotes`, copy-pasting `if args.all_remotes { ... }` block wouldn't be so bad.

PR #4773, opened 2024-11-05
build: fix `cargo publish` by symlinking cli/docs->docs

martinvonz:
# Checklist

If applicable:


thoughtpolice:
FWIW, my other comment on #4630 was meant to say we should:

1. Write a `build.rs` that copies the docs directory to `$OUT_DIR`, which is a variable provided to the build script: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
2. Then, change the `include_str!()` call to instead do something like:

```rust
const DOCS_DIR = concat!(env!("OUT_DIR"), "/docs")
```

since `$OUT_DIR` will also be set during the build process, too.

However, I'm still not sure this will solve the problem, because I don't know if the `.crate` itself will get those files. So maybe we do have to symlink things which... Ugh. Well, maybe it's all we can do...

thoughtpolice:
I also wonder if there's a way to test `cargo publish` before we actually merge a PR...

martinvonz:
> However, I'm still not sure this will solve the problem, because I don't know if the `.crate` itself will get those files. So maybe we do have to symlink things which... Ugh. Well, maybe it's all we can do...

Yes, I suspect so.

> I also wonder if there's a way to test `cargo publish` before we actually merge a PR...

I tested it with `cargo publish --dry-run`, which fails with lots of compilation errors because the jj-cli points to the published 0.22.0 version of jj-lib, but it also seems to show that the `include_str!` does not work. The `include_str!` error goes away with this PR.

cli/build.rs
```diff
@@ -37,6 +38,15 @@ fn main() {
     } else {
         println!("cargo:rustc-env=JJ_VERSION={version}");
     }
+
+    let manifest_dir = PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap());
+    let docs_symlink_path = manifest_dir.join("docs");
```
yuja, line 43: nit: maybe we can assume that cwd is `CARGO_MANIFEST_DIR`? `GIT_HEAD_PATH` already relies on that.

yuja (approved):
Thanks!

cli/build.rs
martinvonz: Good point that we already rely on that. Done.

PR #4777, opened 2024-11-06
docs: switch all toml examples to heading notation

robinst:
I was a bit confused why adding this to the end of my config didn't do anything:

```toml
revset-aliases."immutable_heads()" = "..."
```

It turns out that I had a `[snapshot]` earlier in the file, so either I had to add that key at the beginning or it needs its own table header:

```toml
[revset-aliases]
"immutable_heads()" = "..."
```

IMO the config examples should all use that syntax to be unambiguous. If that sounds good I can go through all the examples in this file.

# Checklist

If applicable:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -58,10 +58,10 @@ colors."commit_id prefix".bold = true
 "commit_id prefix" = { bold = true }
 ```
 
-Jujutsu favors the dotted style in these instructions, if only because it's
-easier to write down in an unconfusing way. If you are confident with TOML
+The docs below refer to keys in text using dotted notation, but example
+blocks will use heading notation to be unambiguous. If you are confident with TOML
 then use whichever suits you in your config. If you mix dotted keys and headings,
-**put the dotted keys before the first heading**.
+**you must put the dotted keys before the first heading**.
 
 That's probably enough TOML to keep you out of trouble but the [syntax guide] is
 very short if you ever need to check.
@@ -70,8 +70,9 @@ very short if you ever need to check.
 ## User settings
 
 ```toml
-user.name = "YOUR NAME"
-user.email = "YOUR_EMAIL@example.com"
+[user]
+name = "YOUR NAME"
+email = "YOUR_EMAIL@example.com"
 ```
 
 Don't forget to change these to your own details!
@@ -87,7 +88,8 @@ active labels alongside the regular colorized output.
 This setting overrides the `NO_COLOR` environment variable (if set).
 
 ```toml
-ui.color = "never" # Turn off color
+[ui]
+color = "never" # Turn off color
 ```
 
 ### Custom colors and styles
@@ -95,7 +97,8 @@ ui.color = "never" # Turn off color
 You can customize the colors used for various elements of the UI. For example:
 
 ```toml
-colors.commit_id = "green"
+[colors]
+commit_id = "green"
 ```
 
 The following colors are available:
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -285,15 +285,17 @@ You can configure the set of immutable commits via
 also consider the `release@origin` bookmark immutable:
 
 ```toml
-revset-aliases."immutable_heads()" = "builtin_immutable_heads() | release@origin"
+[revset-aliases]
+"immutable_heads()" = "builtin_immutable_heads() | release@origin"
```
martinvonz, line 289: Line 61 says that we prefer dotted notation but it doesn't say clearly why. I also see that our documentation is already very inconsistent about it.

One reason to prefer dotted notation is that you can then often take a single line and paste into your config. I think that's especially useful in error message and such. However, it only works to actually paste those strings as long as you consistently use dotted notation in your config file. TOML doesn't have a way to escape out of table syntax (so if there's a `[foo]` table, then you can't put a `bar.baz = 1` config after it; you have to write it `[bar]\nbaz = 1`).

So perhaps we should consistently use table notation instead. Let's hear what others think. Does anyone know what other projects do? I just checked The Cargo Book and they seem to consistently use table notation.

PhilipMetzger: I suggested the dotted notation back then as it made documenting them easier. But since we've gained a bunch of users which it may confuse, so its definitely a negative now. 

> So perhaps we should consistently use table notation instead.

Yeah, definitely in favor of that now, as our developer shortcut no longer scales.

robinst: > Line 61 says that we prefer dotted notation

Ah yeah I see that now, but the problem is that people (e.g. me) don't sequentially read through the docs, they might have landed on a particular section via a link or a search.

Should we change the recommendation on line 61 then and change all the examples to use table notation? Happy to change this PR to do that.

martinvonz: > Ah yeah I see that now, but the problem is that people (e.g. me) don't sequentially read through the docs, they might have landed on a particular section via a link or a search.

Yes, I didn't mean that you should have seen that when reading the doc :) I just meant that we apparently had a documented preference.

> Should we change the recommendation on line 61 then and change all the examples to use table notation? Happy to change this PR to do that.

Given the three thumbs up on my comment above (plus my implicit thumbs up), I think we should. Thanks!

robinst: No worries. Done now!

PhilipMetzger (approved):

martinvonz (approved):
I would squash the two commits since they do conceptually the same thing. Also, our convention is to start the first line with "<topic>: ", so maybe "docs: switch all toml examples to heading notation" in this case

PhilipMetzger:
LGTM, but please adhere to our commit style: 
> The commit message should describe the changes in the commit; the PR description can even be empty, but feel free to include a personal message. We start the commit message with <topic>:  and don't use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). This means if you modified a command in the CLI, use its name as the topic, e.g. next/prev: <your-modification> or conflicts: <your-modification>. We don't currently have a specific guidelines on what to write in the topic field, but the reviewers will help you provide a topic if you have difficulties choosing it. [How to Write a Git Commit Message](https://cbea.ms/git-commit/) is a good guide if you're new to writing good commit messages. We are not particularly strict about the style, but please do explain the reason for the change unless it's obvious.

robinst:
> I would squash the two commits since they do conceptually the same thing. Also, our convention is to start the first line with ": ", so maybe "docs: switch all toml examples to heading notation" in this case

Thanks, done. (I was expecting a squash merge to be used.)

PhilipMetzger:
can rebase this PR, so it can land?

simonmichael:
I hit the same problem.

PR #4778, opened 2024-11-06
github: cancel in-progress builds on stale PRs

thoughtpolice:
When a pull request is pushed multiple times in sequence (e.g. to fix small errors that are noticed post-submission), a backlog of builds ends up needing to be cleared out before your new update can be built. If you push N times, N builds are queued and need to clear out first.

This can cause higher latency for *every* PR since the pool of public runners is shared, and in general seems uneconomical and inefficient since 99% of the time you want to build the new version ASAP.

Luckily GHA has a universal solution to this: use the `concurrency` directive to group all builds for a PR under a name, and when a new build appears in that group, cancel all builds in the group that are in-progress.

Taken from this useful blog post: "Simple trick to save environment and money when using GitHub Actions"

https://turso.tech/blog/simple-trick-to-save-environment-and-money-when-using-github-actions

# Checklist

If applicable:

### DIFF
--- a/.github/workflows/build-nix.yml
+++ b/.github/workflows/build-nix.yml
@@ -6,6 +6,10 @@ on:
       - main
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 permissions: read-all
 
 jobs:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,10 @@ on:
   push:
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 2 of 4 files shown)


thoughtpolice:
Unfortunately this falls short of what I think would be the *best* solution, which is to cancel runs from previous pushes, but build and test every commit individually. Alas, we're still stuck on that.

martinvonz:
> Unfortunately this falls short of what I think would be the _best_ solution, which is to cancel runs from previous pushes, but build and test every commit individually. Alas, we're still stuck on that.

I have done that maybe once or twice ever (i.e. push commit A to trigger CI, then add commit B on top and push again, etc.). I think it's fine if we break that use case. Mayeb we'll add better support for stacked PRs some day anyway.

martinvonz (approved):

martinvonz:
Thanks for this PR btw. I'm happy to waste a bit less resources when e.g. the codespell check notices a type in my PR (as it just did).

PR #4781, opened 2024-11-06
template: add support for logical equality operator

bnjmnt4n:
# Checklist

If applicable:

### DIFF
--- a/cli/src/template_builder.rs
+++ b/cli/src/template_builder.rs
@@ -605,6 +605,41 @@ fn build_binary_operation<'a, L: TemplateLanguage<'a> + ?Sized>(
             let out = lhs.and_then(move |l| Ok(l && rhs.extract()?));
             Ok(L::wrap_boolean(out))
         }
+        BinaryOp::LogicalEq => {
+            let lhs = build_expression(language, diagnostics, build_ctx, lhs_node)?;
+            let rhs = build_expression(language, diagnostics, build_ctx, rhs_node)?;
+            // TODO: Implement as a method on Property type
+            // e.g., lhs.try_eq(rhs) -> Option<Box<dyn TemplateProperty<Output = bool>>>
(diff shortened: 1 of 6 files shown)


cli/src/template_builder.rs
```diff
@@ -605,6 +605,39 @@ fn build_binary_operation<'a, L: TemplateLanguage<'a> + ?Sized>(
...
+                }
+                (lhs_type @ ("Boolean" | "Integer" | "String"), rhs_type) => Err(
+                    TemplateParseError::expected_type(lhs_type, rhs_type, rhs_node.span),
+                ),
+                (lhs_type, _) => {
+                    let message = format!(
+                        r#"Expected expression of type "Boolean", "Integer", or "String", but actual type is "{lhs_type}""#
+                    );
+                    Err(TemplateParseError::expression(message, lhs_node.span))
+                }
```
yuja, line 638: nit: Perhaps, this will be a method of `Property` type (something like `lhs.try_eq(rhs) -> Option<Box<dyn TemplateProperty<Output = bool>>>`), but it can be implemented later.

Can you leave a TODO comment?

cli/src/template_builder.rs
```diff
@@ -1728,6 +1761,54 @@ mod tests {
...
+          = Expected expression of type "String", but actual type is "Boolean"
+        "#);
+        insta::assert_snapshot!(env.parse_err(r#"'a' == 1"#), @r#"
+         --> 1:8
+          |
+        1 | 'a' == 1
+          |        ^
+          |
+          = Expected expression of type "String", but actual type is "Integer"
+        "#);
```
yuja, line 1811: nit: can you add a test for `String == Template`? It should fail, but the distinction between string and template is sometimes unclear.

yuja (approved):
Thanks!

PR #4794, opened 2024-11-07
rebase: allow `-r` with `--skip-emptied`

bnjmnt4n:
# Checklist

If applicable:

### DIFF
--- a/lib/src/rewrite.rs
+++ b/lib/src/rewrite.rs
@@ -826,6 +826,12 @@ pub fn move_commits(
     let mut num_skipped_rebases = 0;
     let mut num_abandoned = 0;
 
+    // Always keep empty commits when rebasing descendants.
+    let rebase_descendant_options = &RebaseOptions {
+        empty: EmptyBehaviour::Keep,
+        simplify_ancestor_merge: options.simplify_ancestor_merge,
+    };
+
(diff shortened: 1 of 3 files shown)


lib/src/rewrite.rs
```diff
@@ -826,6 +826,12 @@ pub fn move_commits(
     let mut num_skipped_rebases = 0;
     let mut num_abandoned = 0;
 
+    // Always keep empty commits when rebasing descendants.
+    let rebase_descendant_options = &RebaseOptions {
+        empty: EmptyBehaviour::Keep,
+        simplify_ancestor_merge: options.simplify_ancestor_merge,
```
martinvonz, line 832: Out of scope for this PR: We should probably change this too. If there already was an ancestor merge, then I think we should leave it when we rebase the descendants. However, it's a bit more complicated here. Let's say we have this:

```
  E
 / \
C   |
|   D
B   |
 \ /
  A
```
Assuming `rebase` had a `--simplify-parents` flag or config and it was enabled here, I would probably expect `jj rebase -r D -d A` to simplify `E`, but I'm not sure what to expect for `jj rebase -r C -d A` or `jj rebase -r B -d A`. I guess it's the same question for `jj abandon` of any of these commits.

martinvonz (approved):
Thanks!

PR #4795, opened 2024-11-07
revset: add `fork_point` function

bnjmnt4n:
This can be used to find the best common ancestors of a revset with an arbitrary number of commits, which cannot be done currently.

# Checklist

If applicable:

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -253,6 +253,12 @@ revsets (expressions) as arguments.
 * `latest(x[, count])`: Latest `count` commits in `x`, based on committer
   timestamp. The default `count` is 1.
 
+* `fork_point(x)`: The fork point of all commits in `x`. The fork point is the
+  common ancestor(s) of all commits in `x` which do not have any descendants
+  that are also common ancestors of all commits in `x`. It is equivalent to
+  the revset `heads(::x_1 & ::x_2 & ... & ::x_N)`, where `x_{1..N}` are commits
+  in `x`. If `x` resolves to a single commit, `fork_point(x)` resolves to `x`.
+
 * `merges()`: Merge commits.
 
 * `description(pattern)`: Commits that have a description matching the given
@@ -361,6 +367,18 @@ given [string pattern](#string-patterns).
     * `roots(E|B)` ⇒ `{B}`
     * `roots(E|A)` ⇒ `{A}`
     * `roots(A)` ⇒ `{A}`
+
+    **function** `fork_point()`
+
+    * `fork_point(E|D)` ⇒ `{A}`
+    * `fork_point(E|C)` ⇒ `{A}`
+    * `fork_point(E|B)` ⇒ `{B}`
+    * `fork_point(E|A)` ⇒ `{A}`
+    * `fork_point(D|C)` ⇒ `{C}`
+    * `fork_point(D|B)` ⇒ `{A}`
+    * `fork_point(B|C)` ⇒ `{A}`
+    * `fork_point(A)` ⇒ `{A}`
+    * `fork_point(none())` ⇒ `{}`
 
 ## String patterns
 
--- a/lib/src/default_index/revset_engine.rs
+++ b/lib/src/default_index/revset_engine.rs
@@ -953,6 +953,22 @@ impl EvaluationContext<'_> {
                 });
                 Ok(Box::new(EagerRevset { positions }))
             }
+            ResolvedExpression::ForkPoint(expression) => {
+                let expression_set = self.evaluate(expression)?;
+                let mut expression_positions_iter = expression_set.positions().attach(index);
+                let Some(position) = expression_positions_iter.next() else {
+                    return Ok(Box::new(EagerRevset::empty()));
+                };
+                let mut positions = vec![position?];
+                for position in expression_positions_iter {
+                    positions = index
+                        .common_ancestors_pos(&positions, [position?].as_slice())
+                        .into_iter()
+                        .collect_vec();
+                }
+                positions.reverse();
+                Ok(Box::new(EagerRevset { positions }))
+            }
             ResolvedExpression::Latest { candidates, count } => {
                 let candidate_set = self.evaluate(candidates)?;
                 Ok(Box::new(self.take_latest_revset(&*candidate_set, *count)?))
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -249,6 +249,7 @@ pub enum RevsetExpression<St: ExpressionState> {
     },
     Heads(Rc<Self>),
     Roots(Rc<Self>),
+    ForkPoint(Rc<Self>),
     Latest {
         candidates: Rc<Self>,
         count: usize,
@@ -425,6 +426,11 @@ impl<St: ExpressionState> RevsetExpression<St> {
             roots: self.clone(),
             generation: generation_range,
         })
+    }
+
+    /// Fork point (best common ancestors) of `self`.
+    pub fn fork_point(self: &Rc<Self>) -> Rc<Self> {
+        Rc::new(Self::ForkPoint(self.clone()))
     }
 
     /// Filter all commits by `predicate` in `self`.
@@ -615,6 +621,7 @@ pub enum ResolvedExpression {
     },
     Heads(Box<Self>),
     Roots(Box<Self>),
+    ForkPoint(Box<Self>),
     Latest {
         candidates: Box<Self>,
         count: usize,
@@ -799,6 +806,11 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
             1
         };
         Ok(candidates.latest(count))
+    });
+    map.insert("fork_point", |diagnostics, function, context| {
+        let [expression_arg] = function.expect_exact_arguments()?;
+        let expression = lower_expression(diagnostics, expression_arg, context)?;
+        Ok(RevsetExpression::fork_point(&expression))
     });
     map.insert("merges", |_diagnostics, function, _context| {
         function.expect_no_arguments()?;
@@ -1247,6 +1259,9 @@ fn try_transform_expression<St: ExpressionState, E>(
             RevsetExpression::Roots(candidates) => {
                 transform_rec(candidates, pre, post)?.map(RevsetExpression::Roots)
             }
+            RevsetExpression::ForkPoint(expression) => {
+                transform_rec(expression, pre, post)?.map(RevsetExpression::ForkPoint)
+            }
             RevsetExpression::Latest { candidates, count } => transform_rec(candidates, pre, post)?
                 .map(|candidates| RevsetExpression::Latest {
                     candidates,
@@ -1436,6 +1451,10 @@ where
         RevsetExpression::Roots(roots) => {
             let roots = folder.fold_expression(roots)?;
             RevsetExpression::Roots(roots).into()
+        }
+        RevsetExpression::ForkPoint(expression) => {
+            let expression = folder.fold_expression(expression)?;
+            RevsetExpression::ForkPoint(expression).into()
         }
         RevsetExpression::Latest { candidates, count } => {
             let candidates = folder.fold_expression(candidates)?;
@@ -2327,6 +2346,9 @@ impl VisibilityResolutionContext<'_> {
             RevsetExpression::Roots(candidates) => {
                 ResolvedExpression::Roots(self.resolve(candidates).into())
             }
+            RevsetExpression::ForkPoint(expression) => {
+                ResolvedExpression::ForkPoint(self.resolve(expression).into())
+            }
             RevsetExpression::Latest { candidates, count } => ResolvedExpression::Latest {
                 candidates: self.resolve(candidates).into(),
                 count: *count,
@@ -2431,6 +2453,7 @@ impl VisibilityResolutionContext<'_> {
             | RevsetExpression::Reachable { .. }
             | RevsetExpression::Heads(_)
             | RevsetExpression::Roots(_)
+            | RevsetExpression::ForkPoint(_)
             | RevsetExpression::Latest { .. } => {
                 ResolvedPredicateExpression::Set(self.resolve(expression).into())
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `--insert-before` options to customize the location of the duplicated
   revisions.
 
+* New `fork_point()` revset function can be used to obtain the fork point
+  of multiple commits.
+
 ### Fixed bugs
 
 ## [0.23.0] - 2024-11-06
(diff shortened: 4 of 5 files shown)


algmyr:
I don't think GCA is the term used for this, it's [lowest (or sometimes last) common ancestor](https://en.wikipedia.org/wiki/Lowest_common_ancestor). I would probably be fine with `lca`, but maybe others might not be as used to the abbreviations. Expanding it fully feels quite wordy though. Maybe `common_ancestor(revset)` is precise/descriptive enough?

lib/src/default_index/revset_engine.rs
```diff
@@ -957,6 +957,21 @@ impl EvaluationContext<'_> {
...
+                    return Ok(Box::new(EagerRevset::empty()));
+                };
+                let mut positions = vec![position?];
+                for position in expression_positions_iter {
+                    positions = index
+                        .common_ancestors_pos(&positions, [position?].as_slice())
+                        .into_iter()
+                        .collect_vec();
+                }
+                Ok(Box::new(EagerRevset { positions }))
```
yuja, line 973: Perhaps, the order is reversed because `common_ancestors_pos()` returns `BTreeSet`?

lib/src/revset.rs
```diff
@@ -253,6 +253,7 @@ pub enum RevsetExpression<St: ExpressionState> {
         candidates: Rc<Self>,
         count: usize,
     },
+    GreatestCommonAncestors(Rc<Self>),
```
yuja, line 256: nit: I would move it closer to heads/roots because "gca" is DAG-based operation.

yuja (commented):
Perhaps, it's the "greatest" in terms of generation numbers? I feel both `gca` and `lca` are equally cryptic.

`common_ancestor()`, `best_ancestor()`, `closest_ancestor()`?

bnjmnt4n:
Git calls it best common ancestor (via [`git merge-base`](https://git-scm.com/docs/git-merge-base)):

> git merge-base finds the best common ancestor(s) between two commits to use in a three-way merge. One common ancestor is better than another common ancestor if the latter is an ancestor of the former. A common ancestor that does not have any better common ancestor is a best common ancestor, i.e. a merge base. Note that there can be more than one merge base for a pair of commits.

lib/src/default_index/revset_engine.rs
bnjmnt4n: Ah, because all revsets must return items in reverse index position?

bnjmnt4n:
Does anyone else have any thoughts on the naming? Also, should the name be in singular or plural form (`common_ancestor` vs `common_ancestors`)? I went with plural because the function can return multiple results if there are multiple common ancestors.

yuja:
If we don't say "best", "greatest", etc., a singular form would suggest that it won't include all common ancestors. So I prefer singular `common_ancestor()`.

bnjmnt4n:
Now that I think about it, I agree that `common_ancestors` seems to indicate `::a & ::b & ::c & ...`. What about `best_common_ancestors`?

yuja:
> What about `best_common_ancestors`?

That seems good, but isn't it too wordy? I'm okay with either `common_ancestor()` (singular), `best_common_ancestor[s]()` (singular/plural). The return value is usually singular.

emilazy:
Would `ancestor` be too much? I feel the fact that it’s common is implicit to the fact that you’re passing multiple arguments. Also, I might be misreading, but `ancestors()` should be `root()` (it looks like it might currently be `none()`?).

bnjmnt4n:
> Would `ancestor` be too much? I feel the fact that it’s common is implicit to the fact that you’re passing multiple arguments.

`ancestor` is way to close to `ancestors`, which obtains the ancestors of all commits in the given revset. Additionally, both functions technically accept only a single revset argument, but the revset can resolve to multiple commits instead.

> Also, I might be misreading, but `ancestor()` should be `root()` (it looks like it might currently be `none()`?).

This is an interesting point, I'm not familiar enough with graph theory (?) to say if the best common ancestors of an empty set of nodes is the root of the graph.

algmyr:
I think the best common ancestor of a single commit should be itself, and the best common ancestor of an empty set ought to be none (it doesn't exist)

emilazy:
Hmm, you might actually be right. We want the property that `ancestor(A | B)` is `ancestor(ancestor(A) | ancestor(B))`, right? Therefore, yes, `ancestor(A)` should be `A`, and `ancestor(none())` should be the identity element, and I guess `none()` is that identity?

martinvonz:
Something like `fork_point()` might also work. I'm not sure I like it better, but figured I'd share it at least.

martinvonz:
> How can a commit be its own ancestor?

We simply define it such that a commits ancestor includes itself. We do that because it's often useful in practice. Mercurial does the same.

emilazy:
Or, to motivate it directly in terms of a defining property: the greatest common ancestor of a set of commits is the commit that all elements of the set are descendents of, such that there is no descendent of the ancestor that also satisfies that property. For a single commit, that’s the commit itself. (This still leaves the choice of the ancestor of an empty set somewhat arbitrary, though it already makes it clear that there can be no single unique choice for many commit graphs, which motivates `none()` as a result.)

joyously:
I think if it includes itself, it shouldn't be called ancestor. Commits don't appear out of thin air, do they? Wouldn't there at least be `root()`?

algmyr:
> I think if it includes itself, it shouldn't be called ancestor.

In graph theory, when talking about things like [lowest common ancestor (LCA)](https://en.wikipedia.org/wiki/Lowest_common_ancestor) you consider nodes to be descendants of themselves.

emilazy:
It’s also the same case as the greatest common divisor of a single number being itself (which Python’s `math.gcd` implements).

joyously:
> In graph theory, when talking about things like [lowest common ancestor (LCA)](https://en.wikipedia.org/wiki/Lowest_common_ancestor) you consider nodes to be descendants of themselves.

If this is simply for the name of a function in code, call it whatever and explain with a comment.
If it's something a user will see, don't confuse us with "you are your own ancestor", even if another VCS did it.

emilazy:
It’s the standard graph theory terminology, not a VCS‐specific quirk, and the mathematically obvious definition. People anyway aren’t likely to call `ancestor` on single commits except as a result of computed revsets, where the single‐commit case will provide correct results that are consistent with its behaviour on revsets of other sizes.

algmyr:
> Hmm, you might actually be right. We want the property that `ancestor(As | Bs)` is `ancestor(ancestor(As) | ancestor(Bs))`, right? Therefore, yes, `ancestor(A)` (where `A` is a single commit) should be `A`, and `ancestor(none())` should be the identity element, and I guess `none()` is that identity?

This is mostly a tangent, but the empty case is a bit iffy and you end up having to be careful in how you define what a common ancestor is (to avoid issues of vacuous truths).

E.g. for a set of nodes `S` and we want to define the set of all common ancestors (ignoring the lowest/best part), it's tempting to try

"`common_ancestors(S)` are nodes with all nodes in `S` as descendants"

But this behaves poorly if `S` is empty. It's vacuously true that any node is an ancestor of all nodes in the empty set!

If you make the (sensible) constraint that common ancestors need to be an ancestors of some node in `S`, then you should end with nicer behavior that gives you the `none` result.

emilazy:
Some more motivation of the base cases here:

If we desire that `ancestor(as | bs)` equals `ancestor(ancestor(as) | ancestor(bs))`, as it does for the >1 commit case, then we immediately get `ancestor(as)` = `ancestor(as | none())` (because `as | none()` = `none()`) = `ancestor(ancestor(none()) | ancestor(bs))`, which can only (as far as I can see) be equal to `ancestor(none() | bs)` if `ancestor(none())` = `none()` and `ancestor(ancestor(bs))` = `ancestor(bs)`, which gives us both the zero and one commit cases (assuming I understand correctly that `ancestor()` should never return a revset of more than one commit – although I wonder if that’s actually a desirable property, as I think it differs from `heads(::a & ::b)`?).

More pragmatically, a common practical use of this operation – Git’s `git merge-base` – is to answer the question “what’s the most ‘recent’ commit I can base my work on if I need it to be able to cleanly merge into these branches without pulling in additional commits only present in some of them?”. For the single‐commit case, the answer is that you can base that work on the commit itself, since there’s nothing “missing” from the non‐existent other branches. (Although, in that case I think it doesn’t give us a clear answer for `ancestor(none())`, since I think `root()` would be a valid answer there; it’s other properties that rule that case out.)

Base cases are hard. It wasn’t that long ago that people found the idea of zero very alien, and rejected formulas like *n* + 0 = n, *n* × 0 = 0, or the idea that you could sum up no integers at all and get 0, or take the product of no integers at all and get 1. But once you work out the rules it’s clear that there’s only one correct answer, and that skipping defining it because it seems weird and unintuitive at first reduces the power and convenience of your operations and forces you to introduce special cases elsewhere in the system.

yuja:
> assuming I understand correctly that `ancestor()` should never return a revset of more than one commit

There's an edge case that `ancestor()` is resolved to multiple revisions, so I'm afraid that we cannot define `ancestor(*xs)` based on `ancestor(a | b | ...)`.
https://github.com/martinvonz/jj/pull/4795/commits/a1ced5b280f2f0af1c00261f5998faba51340981#diff-6bd5d4c0fd459e2c1cb3f1e0d880d6f23c4e3b11863adc76db3931e4b94d194dR2524-R2542

> Something like `fork_point()` might also work.

Sounds also good. It would reflect user's intent.

martinvonz:
> > Something like `fork_point()` might also work.
> 
> Sounds also good. It would reflect user's intent.

Actually, maybe even just `forks()` since we have `merges()` (not `merge_point()`). "Forks" and "merges" seem to work well as analogies to roads forking and merging.

yuja:
> EDIT: But an important difference between them is that `merges()` doesn't take an input set, so maybe `forks()` (and `fork_points()`?) would make it sound like `forks()` returns all commits with more than one child.

Yes, maybe we can add `merge_point(xs)` that finds the best common descendants, but I don't know if this name makes sense.

martinvonz:
There doesn't seem to be a clear consensus and no one seems to feel strongly. To avoid this just getting stalled, I'm fine with @bnjmnt4n making the final decision on the name. Sounds good? Or do other feel like there's a name that's clearly best (and or names that they really don't want)?

bnjmnt4n:
I like the suggested `fork_point`, which is relatively short and doesn't have any immediately ambiguous/conflicting meanings. I'm not too sure about the preferred definition of `fork_point` in `docs/revsets.md`: would anyone have any suggestions for the definition?

algmyr:
Maybe it makes sense to give the example of the two commit case (which I think is `heads(::x & ::y)`) and just say fork_points generalizes this to an arbitrary number of commits?

yuja:
We have "examples" section in the doc. `fork_point()` should probably be added there.

docs/revsets.md
```diff
@@ -362,6 +368,18 @@ given [string pattern](#string-patterns).
...
+
+    * `fork_point(E|D)` ⇒ `{A}`
+    * `fork_point(E|C)` ⇒ `{A}`
+    * `fork_point(E|B)` ⇒ `{B}`
+    * `fork_point(E|A)` ⇒ `{A}`
+    * `fork_point(D|C)` ⇒ `{C}`
+    * `fork_point(D|B)` ⇒ `{A}`
+    * `fork_point(B|C)` ⇒ `{A}`
+    * `fork_point(A)` ⇒ `{A}`
+    * `fork_point()` ⇒ `{}`
```
martinvonz, line 381: nit: maybe `fork_point(none())` so users don't think it can be called with no arguments

martinvonz (approved):

arxanas:
"Fork point" already has a specific meaning in Git that involves traversing both the commit ancestry and the reflog. Is it intentional to have adopted the same term here? I was quite confused by the request in #4915 referencing the "fork point" because I didn't see why "three-dot diffs" would use (Git) fork points.

arxanas:
> We talked about possibly using merge_point() for the opposite here: https://github.com/martinvonz/jj/pull/4795#issuecomment-2475119832

That makes sense to me.

---

Since I'm already here, here's some other thoughts, but I'm not specifically proposing anything:

- [not in favor] I agree that all the terms in the family "nearest/closest/least/greatest/best" common ancestor are unintuitive and wordy.
- [not in favor] I didn't see mentioned in this thread the terms from order theory: https://en.wikipedia.org/wiki/Join_and_meet
  - "join" and "meet" are quite concise and probably used extensively in the field, but I could never remember which was which. The names are not really intuitive, so I would not use them here.
  - "supremum" and "infimum" are... actually not the worst, in my opinion, because many users can leverage their existing etymological intuition to remember what they mean. But they're not "real" words outside of the mathematical sphere, and they're slightly longer, so I could easily see people failing to remember them, or how to spell them. (`sup` and `inf` are real abbreviations, as I recall, and I think would be better than e.g. `gca`, but still not ideal.)
  - "least upper bound" and "greatest lower bound" *sound* like they have the same problem as "least/greatest common ancestor", but I've seen them used much more consistently and much less confusingly. (`lub`/`glb` are also usable abbreviations and sometimes used in other contexts, like type systems, but worse than `sup`/`inf` in my opinion.) They're still fairly wordy and, while more intuitive than "join"/"meet", maybe still not sufficiently intuitive for users?
- [neutral] I think "fork" is a fine term overall, particularly as it has a natural analogue in trees, etc., similar to the "branch" metaphor. It just has historical baggage and ambiguity:
  - The meaning of "fork point" specifically as in Git is confusing, and, honestly, I question whether people really use this feature much. I think noting that our `fork_point` is unrelated to that "fork point" in the docs would be fine.
  - "fork" itself has a specific meaning in OSS, and "forks" in the sense that we're now describing them *sometimes* agree with the OSS "fork" meaning, which is unfortunate because some users might only think they understand what it means. I'm not convinced that we can't overcome this problem in documentation too, though.
  - One could argue that the natural dual to "fork" is ["join"](https://en.wikipedia.org/wiki/Fork%E2%80%93join_model) rather than "merge".
- [not in favor] We could also work backwards and ask "what's the opposite of a 'merge'?".
  - Maybe it would be a "split" point or a "divide"/"division" point? "Diverge"/"divergence" point? They're not the worst terms, in my opinion, but probably don't offer a significant advantage over "fork".
  - In the tree metaphor, besides the term "fork", I'm sure some people have called it a "branch" point or a "branching" point, but of course it seems dangerous to use "branch" as a term due to the confusion with specifically Git-style branches.
- [question] Is there a reason not to adopt the existing terminology of "merge-base" from Git?
  - It's not necessarily *obvious*,  but it seems precise, and probably has a reasonable degree of penetration as a widespread term.
  - We could also shorten it to "base"? It's a little more concise, but still feels fairly precise and descriptive. I could see people casually discussing "what's the base commit for this?" and have it be both intuitive and technically correct, which I think are strong factors in favor of the term.

emilazy:
I was actually thinking of proposing `base()` for the bikeshed, so consider this a seconding of that proposal :) I can see people confusing it with `roots()`, but that seems okay; there’s probably no terminological choice that won’t seem ambiguous in some contexts.

martinvonz:
> One could argue that the natural dual to "fork" is ["join"](https://en.wikipedia.org/wiki/Fork%E2%80%93join_model) rather than "merge".

When talking about road, I think it's "fork" and "merge".

PR #4796, opened 2024-11-08
revset: reorganize expression resolution/evaluation methods

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -531,27 +531,40 @@ impl<St: ExpressionState<CommitRef = RevsetCommitRef>> RevsetExpression<St> {
 impl UserRevsetExpression {
     /// Resolve a user-provided expression. Symbols will be resolved using the
     /// provided `SymbolResolver`.
-    // TODO: return ResolvedRevsetExpression which can still be combined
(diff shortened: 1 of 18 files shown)


lib/src/revset.rs
```diff
@@ -531,27 +531,40 @@ impl<St: ExpressionState<CommitRef = RevsetCommitRef>> RevsetExpression<St> {
...
     }
 }
 
 impl ResolvedRevsetExpression {
-    /// Resolve a programmatically created revset expression and evaluate it in
-    /// the repo.
-    // TODO: rename to .evaluate(), and add .evaluate_unoptimized() for
-    // testing/debugging purpose?
-    pub fn evaluate_programmatic<'index>(
+    /// Evaluates this expression after optimizing it.
```
martinvonz, line 544: nit: "Optimizes and evaluates this expression" is clearer to me. The "after optimizing it" could otherwise be interpreted as a requirement on the input (i.e. that you have to optimize it before calling this function).

martinvonz (approved):

lib/src/revset.rs
yuja: Updated.

Issue #3820, opened 2024-06-03
FR: Automatically `jj workspace update-stale`


torquestomp:
**Is your feature request related to a problem? Please describe.**
When working with a repo that has multiple working copies, it is common to sync/fetch/rebase in one wc before switching to another.  This often requires running `jj workspace update-stale` to update the working copy, but this is generally busy work and usually does not materially impact the thing being worked on in the other wc.

**Describe the solution you'd like**
A new config option, e.g. "ui.always-update-stale" which causes jj to automatically run `jj workspace update-stale` for me whenever it's necessary, printing a status message about it and proceeding with the original command rather than halting execution, making me run `jj workspace update-stale` manually and then re-running the original command.

PhilipMetzger:
done as of 49890fa

PR #4807, opened 2024-11-08
workspace-update-stale: add a config knob to run automatically

torquestomp:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   [documentation](https://martinvonz.github.io/jj/latest/install-and-setup/#command-line-completion)
   to activate them.
 
+* Added the config setting `snapshot.auto-update-stale` for automatically
+  running `jj workspace update-stale` when applicable.
+
 ### Fixed bugs
 
 ## [0.23.0] - 2024-11-06
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -54,7 +54,6 @@ use jj_lib::backend::CommitId;
 use jj_lib::backend::MergedTreeId;
 use jj_lib::backend::TreeValue;
 use jj_lib::commit::Commit;
-use jj_lib::dag_walk;
 use jj_lib::file_util;
 use jj_lib::fileset;
 use jj_lib::fileset::FilesetDiagnostics;
@@ -111,10 +110,10 @@ use jj_lib::str_util::StringPattern;
 use jj_lib::transaction::Transaction;
 use jj_lib::view::View;
 use jj_lib::working_copy::CheckoutStats;
-use jj_lib::working_copy::LockedWorkingCopy;
 use jj_lib::working_copy::SnapshotOptions;
 use jj_lib::working_copy::WorkingCopy;
 use jj_lib::working_copy::WorkingCopyFactory;
+use jj_lib::working_copy::WorkingCopyFreshness;
 use jj_lib::workspace::default_working_copy_factories;
 use jj_lib::workspace::get_working_copy_factory;
 use jj_lib::workspace::DefaultWorkspaceLoaderFactory;
@@ -1558,7 +1557,7 @@ impl WorkspaceCommandHelper {
         let mut locked_ws = self.workspace.start_working_copy_mutation()?;
         let old_op_id = locked_ws.locked_wc().old_operation_id().clone();
         let (repo, wc_commit) =
-            match check_stale_working_copy(locked_ws.locked_wc(), &wc_commit, &repo) {
+            match WorkingCopyFreshness::check_stale(locked_ws.locked_wc(), &wc_commit, &repo) {
                 Ok(WorkingCopyFreshness::Fresh) => (repo, wc_commit),
                 Ok(WorkingCopyFreshness::Updated(wc_operation)) => {
                     let repo = repo.reload_at(&wc_operation)?;
@@ -2212,55 +2211,6 @@ pub fn start_repo_transaction(
     quoted_strings.extend(string_args.iter().skip(1).map(shell_escape));
     tx.set_tag("args".to_string(), quoted_strings.join(" "));
     tx
-}
-
-/// Whether the working copy is stale or not.
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub enum WorkingCopyFreshness {
-    /// The working copy isn't stale, and no need to reload the repo.
-    Fresh,
-    /// The working copy was updated since we loaded the repo. The repo must be
-    /// reloaded at the working copy's operation.
-    Updated(Box<Operation>),
-    /// The working copy is behind the latest operation.
-    WorkingCopyStale,
-    /// The working copy is a sibling of the latest operation.
-    SiblingOperation,
-}
-
-#[instrument(skip_all)]
-pub fn check_stale_working_copy(
-    locked_wc: &dyn LockedWorkingCopy,
-    wc_commit: &Commit,
-    repo: &ReadonlyRepo,
-) -> Result<WorkingCopyFreshness, OpStoreError> {
-    // Check if the working copy's tree matches the repo's view
-    let wc_tree_id = locked_wc.old_tree_id();
-    if wc_commit.tree_id() == wc_tree_id {
-        // The working copy isn't stale, and no need to reload the repo.
-        Ok(WorkingCopyFreshness::Fresh)
-    } else {
-        let wc_operation = repo.loader().load_operation(locked_wc.old_operation_id())?;
-        let repo_operation = repo.operation();
-        let ancestor_op = dag_walk::closest_common_node_ok(
-            [Ok(wc_operation.clone())],
-            [Ok(repo_operation.clone())],
-            |op: &Operation| op.id().clone(),
-            |op: &Operation| op.parents().collect_vec(),
-        )?
-        .expect("unrelated operations");
-        if ancestor_op.id() == repo_operation.id() {
-            // The working copy was updated since we loaded the repo. The repo must be
-            // reloaded at the working copy's operation.
-            Ok(WorkingCopyFreshness::Updated(Box::new(wc_operation)))
-        } else if ancestor_op.id() == wc_operation.id() {
-            // The working copy was not updated when some repo operation committed,
-            // meaning that it's stale compared to the repo view.
-            Ok(WorkingCopyFreshness::WorkingCopyStale)
-        } else {
-            Ok(WorkingCopyFreshness::SiblingOperation)
-        }
-    }
 }
 
 #[instrument(skip_all)]
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -372,7 +372,49 @@ impl CommandHelper {
     #[instrument(skip(self, ui))]
     pub fn workspace_helper(&self, ui: &Ui) -> Result<WorkspaceCommandHelper, CommandError> {
         let mut workspace_command = self.workspace_helper_no_snapshot(ui)?;
-        workspace_command.maybe_snapshot(ui)?;
+
+        let workspace_command = match workspace_command.maybe_snapshot_impl(ui) {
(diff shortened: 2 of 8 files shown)


torquestomp:
Fixes issue #3820

PhilipMetzger:
In my opinion the PR contains three unrelated topics in the stack. As the first commit is a docs cleanup (1), the second two are moving the recovering commit logic into lib (2) (which makes it usable for the Google server) and then the actually advertised PR (3) which depends on (2). I think at least the first commit should be split out, since it has no relation to rest, the other two would be nice to have but can stay together. 

And all commits are missing a motivation, as it definitely is not obvious what they want to achieve. Quoted from contributing.md:
> We are not particularly strict about the style, but please do explain the reason for the change unless it's obvious.

torquestomp:
> In my opinion the PR contains three unrelated topics in the stack. As the first commit is a docs cleanup (1), the second two are moving the recovering commit logic into lib (2) (which makes it usable for the Google server) and then the actually advertised PR (3) which depends on (2). I think at least the first commit should be split out, since it has no relation to rest, the other two would be nice to have but can stay together.

The only difference between (2) and (3) is one moves stuff to lib, the other to cli_util, I don't think they're very different.  I split (1) to its own PR but it's still here too, I'm not sure how to set up dependent PRs with jj.

> 
> And all commits are missing a motivation, as it definitely is not obvious what they want to achieve. Quoted from contributing.md:
> 
> > We are not particularly strict about the style, but please do explain the reason for the change unless it's obvious.

Added some more text to the commit descriptions.

lib/src/workspace.rs
```diff
@@ -461,11 +464,64 @@ pub struct LockedWorkspace<'a> {
...
+    }
+}
+
 impl LockedWorkspace<'_> {
     pub fn locked_wc(&mut self) -> &mut dyn LockedWorkingCopy {
         self.locked_wc.as_mut()
     }
 
+    // Recover this workspace to its last known checkout.
+    pub fn create_and_check_out_recovery_commit(
```
yuja, line 489: It's odd that `LockedWorkspace` has a method that does very specific thing. Maybe this can be moved to cli_util?

cli/src/cli_util.rs
```diff
@@ -429,12 +429,13 @@ impl CommandHelper {
...
+                let update_stale = false;
+                workspace_command.maybe_snapshot_and_update_stale(ui, update_stale)?;
 
                 let wc_commit_id = workspace_command.get_wc_commit_id().unwrap();
-                Ok(StaleWorkingCopy::Snapshotted(
-                    workspace_command.repo().store().get_commit(wc_commit_id)?,
-                ))
+                let repo = workspace_command.repo().clone();
+                let commit = repo.store().get_commit(wc_commit_id)?;
+                Ok(StaleWorkingCopy::Snapshotted((repo, commit)))
```
yuja, line 438: Perhaps, this function can return `WorkspaceCommandHelper` instead of `(repo, wc_commit)` pair?

For reference, `for_stale_working_copy()` was a `CommandHelper` method before. 4d0f1b974600fd9c1425b68b5400d51e94515939

cli/src/cli_util.rs
```diff
@@ -924,7 +932,35 @@ impl WorkspaceCommandHelper {
...
+                            )?;
+                            writeln!(
+                                ui.status(),
+                                "Automatically updated to fresh commit {}",
+                                short_commit_hash(stale_commit.id())
+                            )?;
+                            self.write_stale_commit_stats(ui, &new_commit, stats)?;
+
+                            self.user_repo = ReadonlyUserRepo::new(repo);
+                        }
```
yuja, line 959: I think this can be simplified if auto-update-stale is handled by `CommandHelper::workspace_helper()`. Something like:
```rust
let maybe_stale = workspace_command.maybe_snapshot(ui)?;
let Some(stale_err) = maybe_stale else { return Ok(workspace_command); };
if !auto_update_stale {
   return Err(stale_err.into());
}
Self::load_stale_working_copy_and_recover_from_it(..) // this function will reload workspace and repo anyway
```

cli/src/cli_util.rs
```diff
@@ -1604,7 +1653,11 @@ impl WorkspaceCommandHelper {
     }
 
     #[instrument(skip_all)]
-    fn snapshot_working_copy(&mut self, ui: &Ui) -> Result<(), CommandError> {
+    fn snapshot_working_copy(
+        &mut self,
+        ui: &Ui,
+        update_stale: bool,
```
yuja, line 1659: Instead of pass `update_stale` in to this function, `SnapshotWorkingCopyResult` can be an error type to be converted to `CommandError` by caller?

CHANGELOG.md
```diff
@@ -111,6 +111,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Added the config settings `diff.color-words.context` and `diff.git.context` to
   control the default number of lines of context shown.
 
+* Added the config setting `snapshot.auto-update-stale` for automatically
+  running `jj workspace update-stale` when applicable.
+
```
yuja, line 116: nit: move changelog entry to the unreleased section.

lib/src/workspace.rs
torquestomp: I'd prefer to keep this in lib if possible for use by extensions.  How about moving it to working_copy.rs as a free function?

yuja: A free function sounds good to me, but can you add some commit description about why it should be in the lib? Ah, "the extensions" means extension commands/UI that uses jj-lib?

Also, iirc, we generally avoid adding commit description (that might have to be translated) to lib, so `"RECOVERY COMMIT FROM ...` might have to be a function parameter.

torquestomp: Done, yes, and done.

cli/src/cli_util.rs
```diff
@@ -1535,6 +1535,22 @@ to the current parents may contain changes from multiple commits.
         self.commit_summary_template().format(commit, formatter)
     }
 
+    pub fn write_stale_commit_stats(
```
yuja, line 1648: nit: perhaps, we can extract `print_update_stats()` or something from `update_working_copy()`. It's not important that this function doesn't print the wc parents.

(can be addressed later)

cli/src/cli_util.rs
```diff
@@ -413,6 +418,40 @@ impl CommandHelper {
             })
     }
 
+    pub fn load_stale_working_copy_commit(
```
yuja, line 463: I don't think we need to rewrite `for_stale_working_copy()` not to return `WorkspaceCommandHelper` in this patch. If it returned `WorkspaceCommandHelper`, the caller can just reuse it in the next patch, I think.

If `Recovered`, the `workspace_command` is instantiated for the current op, so it can be reused. If `Snapshotted`, the caller will create new `workspace_command` anyway, so it can be done in this function.

cli/src/cli_util.rs
```diff
@@ -372,7 +372,49 @@ impl CommandHelper {
...
+                            workspace_command.unchecked_start_working_copy_mutation()?;
+                        let stats = update_stale_working_copy(
+                            locked_ws,
+                            repo.op_id().clone(),
+                            &stale_commit,
+                            &new_commit,
+                        )?;
+                        writeln!(
+                            ui.status(),
+                            "Automatically updated to fresh commit {}",
```
yuja, line 406: nit: maybe warning is better here? I think this is an usual event.

cli/src/cli_util.rs
```diff
@@ -855,6 +896,29 @@ pub struct WorkspaceCommandHelper {
...
+    }
+}
+
+impl<E> From<E> for SnapshotWorkingCopyError
+where
+    E: Into<CommandError>,
+{
+    fn from(value: E) -> Self {
+        Self::Command(value.into())
+    }
```
yuja, line 919: nit: I wouldn't add automatic conversion because the caller should be careful to not map stale wc error to `Command(_)`.

cli/src/cli_util.rs
```diff
@@ -372,7 +372,49 @@ impl CommandHelper {
     #[instrument(skip(self, ui))]
     pub fn workspace_helper(&self, ui: &Ui) -> Result<WorkspaceCommandHelper, CommandError> {
         let mut workspace_command = self.workspace_helper_no_snapshot(ui)?;
-        workspace_command.maybe_snapshot(ui)?;
+
+        let workspace_command = match workspace_command.maybe_snapshot_impl(ui) {
```
yuja, line 376: Just an idea, but maybe we can move almost all codes from `cmd_workspace_update_stale()`?

```rust
fn cmd_workspace_update_stale() {
    let (workspace_command, was_stale) = command.recover(ui)?;
    if !was_stale {
        writeln!(ui.status(), "Nothing to do...")?;
    }
}

impl WorkspaceCommand {
    fn workspace_helper() {
        if auto_update_stale {
            let (..) = self.recover(ui)?;
            if was_stale {
               writeln!(ui.warning_default(), "Automatically recovered ...")?;
            }
        } else {
            ..
        }
    }
```

lib/src/working_copy.rs
```diff
@@ -369,6 +372,55 @@ impl WorkingCopyFreshness {
...
+pub enum RecoverWorkspaceError {
+    /// Backend error.
+    #[error(transparent)]
+    Backend(#[from] BackendError),
+    /// Error during checkout.
+    #[error(transparent)]
+    Reset(#[from] ResetError),
+    /// Checkout attempted to modify the root commit.
+    #[error("Cannot rewrite the root commit")]
+    RewriteRootCommit,
```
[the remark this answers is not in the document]
torquestomp: Done. I thought it was weird for a struct to hold another struct which holds nothing but it doesn't really hurt anything.

cli/src/cli_util.rs
torquestomp: Left a TODO for now.

cli/src/cli_util.rs
torquestomp: After the rest of the refactoring it's fairly straightforward to consolidate this; done.  Did it in an extra commit at the end of the PR

cli/src/cli_util.rs
```diff
@@ -367,7 +373,40 @@ impl CommandHelper {
...
+                // lower level (e.g. inside snapshot_working_copy()) to avoid recursive locking
+                // of the working copy.
+                let (workspace_command, status) = self.recover_stale_working_copy(ui)?;
+                match status {
+                    StaleWorkingCopyStatus::NotStale
+                    | StaleWorkingCopyStatus::RecoveredMissingOp => {}
+                    StaleWorkingCopyStatus::RecoveredSnapshot(id) => {
+                        writeln!(
+                            ui.warning_default(),
+                            "Automatically updated to fresh commit {}",
```
yuja, line 400: nit: It's odd that warning is printed only for (less scary) `RecoveredSnapshot(_)` case, though I understand that's probably because we've printed lengthy status message for the `RecoveredMissingOp` case.

Maybe we can move all status/warning messages to `recover_stale_working_copy()` function instead of returning `StaleWorkingCopyStatus`? I thought `workspace_helper()` would call `recover_stale_working_copy()` unconditionally, but actually we check the snapshot error first. So I think it's safe to print `"Nothing to do .."` there.

cli/src/cli_util.rs
yuja: Ok, I don't think it's terrible to insert `.map_err(to_other_err)?`, but I agree it would look a bit unusual. We can extract `StaleWorkingCopyError` that isn't convertible to `CommandError`, but I don't have a strong opinion about that. We can refactor it later.

(No action requested)

yuja: I like it, thanks.

cli/src/cli_util.rs
yuja: Looks better, thanks. Maybe we can insert a warning saying "recovering from stale working copy automatically..." or something, but it might be too verbose.

torquestomp: I'll leave this as-is for now, we can iterate on it after there's some usage.

PR #4809, opened 2024-11-09
docs: beginnings of an 'editor setup' section

neongreen:
# Checklist

If applicable:

---

An alternative to https://github.com/martinvonz/jj/pull/4294.

I agree with not checking in editor folders, b/c then it's harder for people to set up their local settings. For VSCode, there is an extension AFAIK, but for Zed it's impossible.

I can't contribute anything about the actual Rust setup in VSCode, but perhaps somebody else can.


docs/contributing.md
```diff
@@ -216,6 +216,48 @@ to use" instructions](https://github.com/rui314/mold#how-to-use).
...
+        // If you are working on docs and don't need `cargo check`, uncomment
+        // this option:
+        // 
+        //   "checkOnSave": false,
+        
+        // Use nightly `rustfmt`, equivalent to `cargo +nightly fmt`
+        "rustfmt": { "extraArgs": ["+nightly"] }
+      }
+    }
+  }
```
yuja, line 258: nit: can you clean up trailing whitespaces in this block?

yuja (approved):
I don't know vscode/zed settings, but these changes generally look good to me, thanks.

docs/contributing.md
neongreen: Done. It turns out Zed respects .editorconfig as of last week, and we have an .editorconfig that disables whitespace trimming. I redefined it so that it's for .rs files only.

neongreen: @yuja I've requested rereview b/c the .editorconfig changes weren't in this PR before

docs/contributing.md
```diff
@@ -216,6 +216,56 @@ to use" instructions](https://github.com/rui314/mold#how-to-use).
 On recent versions of MacOS, the default linker Rust uses is already
 multi-threaded. It should use all the CPU cores without any configuration.
 
+### Editor setup
+
+#### Visual Studio Code
+
+We recommend at least these settings:
+
+```js
```
martinvonz, line 225: nit: does github support `json` (which i guess this is)? maybe it gets rendered in a nicer way then somehow

martinvonz (approved):

docs/contributing.md
neongreen: For this specific block -- we could use `json`.

The next block though is json with comments, where were have several options:

* `json` will look good in the docs but bad on github
* `jsonc` does not work in the docs
* `js` looks a bit worse but is acceptable

If we prioritize docs over github, then I'd use `json`, otherwise `js` as a compromise

yuja (approved):

PR #4811, opened 2024-11-09
cli: make command aliases not interfere with completion

ilyagr:
Follows up on #4740

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/bookmark/create.rs
+++ b/cli/src/commands/bookmark/create.rs
@@ -23,7 +23,7 @@ use crate::command_error::user_error_with_hint;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
-/// Create a new bookmark
+/// Create a new bookmark [aliases: c]
 #[derive(clap::Args, Clone, Debug)]
 pub struct BookmarkCreateArgs {
     /// The bookmark's target revision
--- a/cli/src/commands/bookmark/delete.rs
+++ b/cli/src/commands/bookmark/delete.rs
@@ -22,7 +22,7 @@ use crate::command_error::CommandError;
 use crate::ui::Ui;
 
 /// Delete an existing bookmark and propagate the deletion to remotes on the
-/// next push
+/// next push [aliases: d]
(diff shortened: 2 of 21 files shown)


cli/src/commands/bookmark/create.rs
```diff
@@ -23,7 +23,7 @@ use crate::command_error::user_error_with_hint;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
-/// Create a new bookmark
+/// Create a new bookmark [aliases: c]
```
yuja, line 26: nit: I think these changes can be source of future bugs because we need to update two different places when we add/remove aliases. We can instead make `jj util completion` to remove prefix aliases.

yuja (commented):
I found this bug report in clap. https://github.com/clap-rs/clap/issues/4265

Indeed, `jj desc --` no longer list arguments. I think that's worse than `jj desc<TAB>` doesn't insert space.

cli/src/commands/bookmark/create.rs
ilyagr: Yes, I complained about this in the commit description. I am not sure how easy it is to have `jj util completion` remove prefix aliases, it should probably be a `clap` feature.

But I agree that, together with your observation that this breaks completion after `jj desc`, this is problematic. Do we want to revert #4740 as well?

Anyway, the part of this that I actually care about is documenting `b` for bookmark. I'll split that off into another PR.

martinvonz: > Do we want to revert #4740 as well?

Makes sense to me

ilyagr: Also, IMO, if that bug in `clap` was fixed, I'd be OK with duplication. It's not ideal, but might be worth it, especially since most ways I can think of adding support to `clap` for suppressing some completions would be somewhat ugly.

I guess there could be a `#[command(long_completions=["every_allowed_completion"]))]` or `omit_long_completions()`. I may or may not write an FR over at clap.

Issue #170, opened 2022-03-22
Add a `jj absorb` command


martinvonz:
## Description

As suggested by @unrelentingtech in #64, we should have a `jj absorb` command. There's [`git absorb`](https://github.com/tummychow/git-absorb), which was inspired by [`hg absorb`](https://www.mercurial-scm.org/repo/hg/rev/5111d11b8719). The two are implemented quite differently. Mercurial's is based on [weaves](https://en.wikipedia.org/wiki/Interleaved_deltas), while Git's is based on ["commuting" diff hunks](https://github.com/tummychow/git-absorb#how-it-works-roughly). @durin42 tells me weaves are superior. I'd like to understand better what the difference is in practice (i.e. some concrete examples and how they behave with each method). Maybe @quark-zju (the original author of `hg absorb`) also has some thoughts.

quark-zju:
I think the superior cases are covered by the tests. If you can port the tests then it's easier to see the differences. Basically weave should behave better when adjacent/context lines are changed.

martinvonz:
I can confirm that. Here's a commit created by `git absorb`:

```
commit 15d9908f911a5eed7bd4b2a5880f52154eb3dc87 (HEAD -> master)
Author: Martin von Zweigbergk <martinvonz@google.com>
Date:   Tue Mar 22 15:18:42 2022 -0700

    fixup! fourth

diff --git a/file b/file
index 46876d0..48ffd66 100644
--- a/file
+++ b/file
@@ -1,7 +1,7 @@
-a
-b2
-c3
-d4
+a5
+b5
+c5
+d5
 e
 f
 g3
```

The number on each line is the commit it was added in (e.g. "c3" was added in commit 3, replacing the previous "c*" line). The lines ending in "5" were added in the index. As you can see, all the fixup would move all the edits in the fourth commit instead of spreading the changes out over commits 1-4. It makes sense when you think about how it works, of course. Thanks for explaining.

martinvonz:
While not automatic like `hg/git absorb`, I've found `jj move` flexible and easy to use, and even more so now that it accepts path arguments. For example, I had some changes in the working copy just now that should go into two different commits. So I ran this:

```
# Move most changes into the parent
$ jj squash CHANGELOG.md src/commands.rs tests/test_describe_command.rs
# Move remaining changes into the grandparent
$ jj move --to @--
```

I still think we should have a `jj absorb`, but I don't miss it as much as would miss `hg absorb`.

[...]

PR #4812, opened 2024-11-09
cli: add `jj absorb`

yuja:
#170

# Checklist

If applicable:


lib/src/annotate.rs
```diff
@@ -116,10 +120,38 @@ pub fn get_annotation_for_file(
...
+    compute_file_annotation(repo, starting_commit.id(), domain, file_path, source)
+}
+
+/// Get line by line annotations for a specific file path starting with the
+/// given content.
+///
+/// The file content at the `starting_commit` is set to `starting_text`. This is
+/// typically one of the file contents in the conflict or merged-parent tree.
+///
+/// See [`get_annotation_for_file()`] for the other arguments.
```
martinvonz, line 172: Out of scope for this PR, but perhaps an alternative is to add an option to continue annotating older version, so we'd record information about every commit where a line was modified, kind of like a weave. That would also be useful for a `jj annotate --interactive/--tui` that lets you view the annotated file at a previous commit.

cli/src/commands/absorb.rs
```diff
@@ -0,0 +1,1172 @@
...
+use jj_lib::settings::UserSettings;
+use pollster::FutureExt as _;
+use tracing::instrument;
+
+use crate::cli_util::CommandHelper;
+use crate::cli_util::RevisionArg;
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+/// Move changes from a revision into the stack of draft revisions
```
martinvonz, line 54: nit: "draft" -> "mutable" (same in the changelog)

cli/src/commands/absorb.rs
```diff
@@ -0,0 +1,1172 @@
...
+///
+/// This command splits changes in the source revision and moves each change to
+/// the revision where the corresponding lines modified last. If the destination
```
martinvonz, line 57: nit: "lines modififed last" -> "lines were modified last"

martinvonz, line 57: nit: "the revision" -> "the closest mutable ancestor" or something like that to be more specific

martinvonz (commented):
Thank you!

I'm not done reviewing yet but I don't have more time right now. Will try to get back to it soon.

lib/src/annotate.rs
yuja: Yes. Even with the current implementation, annotation process can be resumed if the result contained remainder of `CommitSourceMap`, but I'm not sure if that's sufficient for interactive use case or weave-like merging.

cli/src/commands/absorb.rs
```diff
@@ -0,0 +1,1172 @@
...
+        })),
+        MaterializedTreeValue::Symlink { .. } => Err("Is a symlink".into()),
+        MaterializedTreeValue::FileConflict { .. }
+        | MaterializedTreeValue::OtherConflict { .. } => Err("Is a conflict".into()),
+        MaterializedTreeValue::GitSubmodule(_) => Err("Is a Git submodule".into()),
+        MaterializedTreeValue::Tree(_) => panic!("diff should not contain trees"),
+    }
+}
+
+#[cfg(test)]
```
martinvonz, line 407: Wow, this is very thorough testing! Thank you!

martinvonz (approved):
Looking forward to using this!

PR #4817, opened 2024-11-10
completion: teach remote about existing names

senekor:
# Checklist

If applicable:

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -43,6 +43,25 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
     })
 }
 
+pub fn git_remotes() -> Vec<CompletionCandidate> {
+    with_jj(|mut jj| {
+        let output = jj
+            .arg("git")
+            .arg("remote")
+            .arg("list")
+            .output()
+            .map_err(user_error)?;
+
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        Ok(stdout
+            .lines()
+            .filter_map(|line| line.split_once(' ').map(|(name, _url)| name))
+            .map(CompletionCandidate::new)
+            .collect())
+    })
+}
+
 /// Shell out to jj during dynamic completion generation
 ///
 /// In case of errors, print them and early return an empty vector.
--- a/cli/src/commands/bookmark/list.rs
+++ b/cli/src/commands/bookmark/list.rs
@@ -14,6 +14,7 @@
 
 use std::collections::HashSet;
 
+use clap_complete::ArgValueCandidates;
 use itertools::Itertools;
 use jj_lib::git;
 use jj_lib::revset::RevsetExpression;
@@ -24,6 +25,7 @@ use crate::cli_util::RevisionArg;
 use crate::command_error::CommandError;
 use crate::commit_templater::CommitTemplateLanguage;
 use crate::commit_templater::RefName;
+use crate::complete;
 use crate::ui::Ui;
 
 /// List bookmarks and their targets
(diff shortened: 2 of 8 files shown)


cli/src/complete.rs
```diff
@@ -43,6 +43,34 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+        let existing_remotes: Vec<_> = stdout
+            .lines()
+            .map(|line| {
+                line.split_once(' ')
+                    .map(|(name, _url)| name)
+                    .unwrap_or_default()
+            })
+            .collect();
+
+        Ok(["origin", "upstream"]
```
yuja, line 66: I personally don't think it's useful to suggest these names, and I wouldn't even mind if they contained existing remotes.

Just use static list if that's useful?

senekor: You're right, a static list is more than enough.

bnjmnt4n: > I personally don't think it's useful to suggest these names, and I wouldn't even mind if they contained existing remotes.

Personally, I feel completions won't be very useful if they're going to suggest things that don't actually work. I'm not very convinced about completions for new remote names. `origin` is already the default remote name, so there's not much value suggesting that. Perhaps `upstream` is slightly more valuable.

If we do want to add completions for `jj git remote add`, I do think we should filter out existing remote names, because providing a completion which causes the resulting command to fail doesn't add any value to the user. In that case we might as well not provide completions.

senekor: I'm fine with either. It's not a high-stages situation.

The reason I came up with it is because I often create little throw-away repos locally to try something out. These don't have a remote to start. But then I might change my mind and want to add a remote anyway. So I ended up typing "origin" a bunch of times, which isn't a big deal, that's just where the idea came from.

senekor:
I have added more thorough completion of remote names across all `git remote` subcommands.

cli/src/complete.rs
senekor: @yuja are you OK with keeping the logic of not suggesting existing remote names for new remotes? It's just a 10 line function at this point.

yuja: My vote (in order of preference):
1. no completion for new remote names
2. simple static list of "common" names
3. dynamic list of "common" names that doesn't exist

In any case, I don't strongly oppose to adding (3) if people like it.

senekor: I like (3) :slightly_smiling_face:

martinvonz: I vote for (1). I'm also fine with (3) if that's the consensus.

senekor: Alright, I implemented (1).

senekor:
I also added the completion of existing remotes to flags of some other commands:
- `jj git push --remote`
- `jj git fetch --remote`
- `jj bookmark list --remote`

cli/src/complete.rs
```diff
@@ -43,6 +43,29 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
     })
 }
 
+pub fn remotes() -> Vec<CompletionCandidate> {
```
yuja, line 46: nit: rename to `git_remotes()` (because there might be other forges)

cli/src/complete.rs
```diff
@@ -43,6 +43,29 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+            .map_err(user_error)?;
+
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        Ok(stdout
+            .lines()
+            .map(|line| {
+                line.split_once(' ')
+                    .map(|(name, _url)| name)
+                    .unwrap_or_default()
```
yuja, line 62: nit: `.filter_map()` instead of adding `""` as a candidate?

yuja (approved):

PR #4821, opened 2024-11-10
Remove `DescendantRebaser` and cleanup `MutableRepo::rebase_descendants_*` functions

bnjmnt4n:
# Checklist

If applicable:


lib/src/repo.rs
```diff
@@ -1274,16 +1274,6 @@ impl MutableRepo {
     // TODO(ilyagr): Either document that this also moves bookmarks (rename the
     // function and the related functions?) or change things so that this only
     // rebases descendants.
```
yuja, line 1276: nit: orphaned inline comment?

lib/src/repo.rs
```diff
@@ -1275,19 +1275,28 @@ impl MutableRepo {
...
+    ///
+    /// If `options.empty` is the default ([`EmptyBehaviour::Keep`]), all
+    /// rebased descendant commits will be preserved even if they were
+    /// emptied following the rebase operation. A map of newly rebased
+    /// commit ID to original commit ID will be returned.
     ///
-    /// If the `options.empty` is the default, this function will only
-    /// rebase commits, and the return value is what you'd expect it to be.
+    /// Otherwise, this function may rebase some commits and abandon others,
+    /// based on the given [`EmptyBehaviour`]. The behavior is such that only
```
yuja, line 1289: nit `EmptyBehavior` isn't in scope, so this link syntax doesn't work.

lib/src/repo.rs
```diff
@@ -1312,13 +1321,31 @@ impl MutableRepo {
...
 
+    /// Rebase descendants of the rewritten commits.
+    ///
+    /// The descendants of the commits registered in `self.parent_mappings` will
+    /// be recursively rebased onto the new version of their parents.
+    /// Returns the number of rebased descendants.
+    ///
+    /// This function is similar to [`MutableRepo::rebase_descendants`], but
+    /// allows for rebase behavior to be customized via [`RebaseOptions`].
     pub fn rebase_descendants_with_options(
```
yuja, line 1349: fwiw, we have no callers of `rebase_descendants_with_options()`, and one for `rebase_descendants_return_map()` (except for tests.)

Maybe we can remove these and migrate all callers to `rebase_descendants()` and `rebase_descendants_with_options()` (`_return_map`)?

(no action needed within this PR)

yuja (approved):
Nice.

lib/src/repo.rs
bnjmnt4n: I guess those are convenience functions for the default behaviour, and perhaps they should be kept in the lib?

yuja: My feeling is that they're too many. The simplest `rebase_descendants() -> usize` shouldn't be removed, but there aren't many callers of the other variants.

martinvonz (approved):
Thanks! I agree with Yuya that we should remove the infrequently used versions of the functions.

PR #4824, opened 2024-11-10
PoC: Implement `jj git sync` - v1

essiene:
This is not meant to be merged in this form.
I am bashing everything on the head with a large bat, until it works. Once everything works nicely in this draft, I will break it up into smaller tested pieces that are intended to be merged.

* Grab current heads and build a set of [(Parent, Child)...]
* Start a transaction.
* Fetch in the transaction.
* Grab current heads on transaction:
  * Transaction MutableRepo should be updated after fetch.
  * Build a Map where old heads are the keys and the new heads are values.
    * If old == new, skip it.
    * This way if nothing changed, we end up with an empty map and avoid bugs downstream (index.is_ancestor returns true if old == new).
  * Relationship is figured out by index.is_ancestor(old_head_id, new_head_id) in a loop.
* Check if rebase is needed.
  * old_heads.set_diffence(new_heads) gives old heads that are no longer heads, which would mean, their descendants need to be rebased.
* Find children needing rebase:
  * Build list of rebase specs with: (commit, old_parent, new_parent)
  * IMPORTANT: Not YET handling parent commits deleted in remotes.
* Perform simple rebase which abandons emptied commits.

Part of: #1039

# Checklist

If applicable:


martinvonz:
> I'm not sure if there's a better way to handle the splits though, I'm happy to adjust in that case.

I prefer to see it split up so the `jj-lib` changes come first. Speaking of that, I hope it's not too late to say that ideally most of the changes and tests should go in `jj-lib`. The CLI crate should just do the UI-level stuff like handling arguments and presenting the result. Tests are *much* faster in `jj-lib` than in `jj-cli`, so we prefer to test what's possible there.

essiene:
> > I'm not sure if there's a better way to handle the splits though, I'm happy to adjust in that case.
> 
> I prefer to see it split up so the `jj-lib` changes come first. Speaking of that, I hope it's not too late to say that ideally most of the changes and tests should go in `jj-lib`. The CLI crate should just do the UI-level stuff like handling arguments and presenting the result. Tests are _much_ faster in `jj-lib` than in `jj-cli`, so we prefer to test what's possible there.

This makes sense. Let me see how best to rework it to end up with most changes in `jj-lib`

martinvonz:
> The chain of commits is long, so I'm splitting them up for ease of review, starting
> with this simple scaffolding PR.

GitHub allows us to review the individual commits in a PR, so I think it's simpler to combine a few commits into one PR. It makes it easier to navigate the stack of commits. I think separating into multiple PRs makes sense if you want to get some commits merged early.

PR #4826, opened 2024-11-11
cli: document `b` as a "default alias"

ilyagr:
Unlike other documented aliases, it can be redefined by the user. However, I think it's important for people to be informed that `b` exists.

I left `amend`, `unamend`, and `ci` undocumented, since I think we were considering removing or at least discouraging them.

We could also instead make b a normal built-in alias, now that it won't interfere with completion if we define them as in the previous commit (**Update:** This is no longer the case). I am not sure if letting people redefine it is worth the special case.

### DIFF
--- a/cli/src/commands/bookmark/mod.rs
+++ b/cli/src/commands/bookmark/mod.rs
@@ -56,7 +56,10 @@ use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
-/// Manage bookmarks
+// Unlike most other aliases, `b` is defined in the config and can be overridden
+// by the user.
+
+/// Manage bookmarks [default alias: b]
 ///
 /// For information about bookmarks, see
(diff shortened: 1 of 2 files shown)


yuja (approved):
Thanks.

> We could also instead make `b` a normal built-in alias,

I would say no because builtins cannot be overridden, and `jj <whatever-character>` doesn't make sense to me as a reserved command name.

ilyagr:
I wonder whether the `[default alias: b]` docs will be sufficient to not confuse the communication of people who redefined the `b` alias and others.

I think it's probably a good thing that this help text is printed regardless of whether the user redefined `b` to mean something else (at first, I thought this was a downside, but I think I changed my mind).

In any case, this is/was my main concern for keeping the possibility of redefining `b`. We'll probably naturally revisit this if it becomes an issue. In any case, now we'd also have the completion problem again if we went the other way.

Thanks for the review!

PR #4828, opened 2024-11-11
completion: teach commands about bookmark names

senekor:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/bookmark/forget.rs
+++ b/cli/src/commands/bookmark/forget.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use clap_complete::ArgValueCandidates;
 use itertools::Itertools as _;
 use jj_lib::op_store::BookmarkTarget;
 use jj_lib::op_store::RefTarget;
@@ -22,6 +23,7 @@ use jj_lib::view::View;
 use super::find_bookmarks_with;
 use crate::cli_util::CommandHelper;
 use crate::command_error::CommandError;
+use crate::complete;
 use crate::ui::Ui;
 
 /// Forget everything about a bookmark, including its local and remote
@@ -36,7 +38,11 @@ pub struct BookmarkForgetArgs {
     /// By default, the specified name matches exactly. Use `glob:` prefix to
     /// select bookmarks by wildcard pattern. For details, see
     /// https://martinvonz.github.io/jj/latest/revsets/#string-patterns.    
-    #[arg(required = true, value_parser = StringPattern::parse)]
+    #[arg(
+        required = true,
+        value_parser = StringPattern::parse,
+        add = ArgValueCandidates::new(complete::bookmarks),
+    )]
     names: Vec<StringPattern>,
 }
 
--- a/cli/src/commands/bookmark/list.rs
+++ b/cli/src/commands/bookmark/list.rs
@@ -77,7 +77,7 @@ pub struct BookmarkListArgs {
     /// By default, the specified name matches exactly. Use `glob:` prefix to
     /// select bookmarks by wildcard pattern. For details, see
     /// https://martinvonz.github.io/jj/latest/revsets/#string-patterns.
-    #[arg(value_parser = StringPattern::parse)]
+    #[arg(value_parser = StringPattern::parse, add = ArgValueCandidates::new(complete::bookmarks))]
     names: Option<Vec<StringPattern>>,
 
     /// Show bookmarks whose local targets are in the given revisions
--- a/cli/src/commands/bookmark/untrack.rs
+++ b/cli/src/commands/bookmark/untrack.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use clap_complete::ArgValueCandidates;
 use itertools::Itertools as _;
 use jj_lib::git;
 
@@ -19,6 +20,7 @@ use super::find_remote_bookmarks;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::RemoteBookmarkNamePattern;
 use crate::command_error::CommandError;
+use crate::complete;
 use crate::ui::Ui;
 
 /// Stop tracking given remote bookmarks
@@ -34,7 +36,11 @@ pub struct BookmarkUntrackArgs {
     /// https://martinvonz.github.io/jj/latest/revsets/#string-patterns.
     ///
     /// Examples: bookmark@remote, glob:main@*, glob:jjfan-*@upstream
-    #[arg(required = true, value_name = "BOOKMARK@REMOTE")]
+    #[arg(
+        required = true,
+        value_name = "BOOKMARK@REMOTE",
+        add = ArgValueCandidates::new(complete::tracked_bookmarks)
+    )]
     names: Vec<RemoteBookmarkNamePattern>,
 }
 
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -15,6 +15,7 @@
 use clap::FromArgMatches as _;
 use clap_complete::CompletionCandidate;
 use config::Config;
+use itertools::Itertools;
 use jj_lib::workspace::DefaultWorkspaceLoaderFactory;
 use jj_lib::workspace::WorkspaceLoaderFactory as _;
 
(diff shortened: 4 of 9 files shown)


senekor:
Note that I'm using `git.push-bookmark-prefix` as a heuristic for "own" bookmarks, which may not be a good idea. See this comment [here](https://github.com/martinvonz/jj/pull/4823#issuecomment-2467044200):

> I don't think it's good idea to suggest `git.push-bookmark-prefix` as a new name. It's the prefix for automatically-generated bookmarks.

senekor:
@yuja I removed the push prefix stuff from the other PR, but I still think it should make sense here. If a bookmark starts with the prefix I specifically configured, it's probably more relevant to me than other bookmarks, right?

cli/src/commands/bookmark/forget.rs
```diff
@@ -36,7 +38,11 @@ pub struct BookmarkForgetArgs {
     /// By default, the specified name matches exactly. Use `glob:` prefix to
     /// select bookmarks by wildcard pattern. For details, see
     /// https://martinvonz.github.io/jj/latest/revsets/#string-patterns.    
-    #[arg(required = true, value_parser = StringPattern::parse)]
+    #[arg(
+        required = true,
+        value_parser = StringPattern::parse,
+        add = ArgValueCandidates::new(complete::bookmarks),
```
yuja, line 44: "forget" expects a local part of local/remote bookmarks

cli/src/commands/bookmark/list.rs
```diff
@@ -74,7 +76,7 @@ pub struct BookmarkListArgs {
     /// By default, the specified name matches exactly. Use `glob:` prefix to
     /// select bookmarks by wildcard pattern. For details, see
     /// https://martinvonz.github.io/jj/latest/revsets/#string-patterns.
-    #[arg(value_parser = StringPattern::parse)]
+    #[arg(value_parser = StringPattern::parse, add = ArgValueCandidates::new(complete::bookmarks))]
```
yuja, line 79: "list" expects a local part of local/remote bookmarks

cli/src/commands/bookmark/untrack.rs
```diff
@@ -34,7 +36,11 @@ pub struct BookmarkUntrackArgs {
     /// https://martinvonz.github.io/jj/latest/revsets/#string-patterns.
     ///
     /// Examples: bookmark@remote, glob:main@*, glob:jjfan-*@upstream
-    #[arg(required = true, value_name = "BOOKMARK@REMOTE")]
+    #[arg(
+        required = true,
+        value_name = "BOOKMARK@REMOTE",
+        add = ArgValueCandidates::new(complete::tracked_bookmarks)
```
yuja, line 42: "untrack" expects `<name>@<remote>`, not `<name>`

cli/src/complete.rs
```diff
@@ -43,15 +45,113 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+
+        let output = jj
+            .build()
+            .arg("bookmark")
+            .arg("list")
+            .arg("--all-remotes")
+            .arg("--template")
+            .arg(r#"if(remote, name ++ "@" ++ remote ++ "\n")"#)
+            .output()
+            .map_err(user_error)?;
```
yuja, line 90: maybe we can abuse template `if(remote && !tracked, ..)` to filter out tracked bookmarks

cli/src/complete.rs
```diff
@@ -43,15 +45,113 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+            .arg("list")
+            .arg("--tracked")
+            .arg("--template")
+            .arg(r#"if(!remote, name ++ "\n")"#)
+            .output()
+            .map_err(user_error)?;
+
+        Ok(String::from_utf8_lossy(&output.stdout)
+            .lines()
+            .filter(|bookmark| !bookmark.ends_with("@git"))
```
yuja, line 62: nit: `remote == "git"` is excluded by `--tracked`, I think.

cli/src/complete.rs
```diff
@@ -48,10 +49,10 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
 /// In case of errors, print them and early return an empty vector.
 fn with_jj<F>(completion_fn: F) -> Vec<CompletionCandidate>
 where
-    F: FnOnce(std::process::Command) -> Result<Vec<CompletionCandidate>, CommandError>,
+    F: FnOnce(std::process::Command, Config) -> Result<Vec<CompletionCandidate>, CommandError>,
```
yuja: nit: pass `&Config` by reference

cli/src/complete.rs
```diff
@@ -110,6 +111,11 @@ fn get_jj_command() -> Result<std::process::Command, CommandError> {
     let args: GlobalArgs = GlobalArgs::from_arg_matches(&args)?;
 
     if let Some(repository) = args.repository {
+        // Try to update repo-specific config on a best-effort basis.
+        if let Ok(loader) = DefaultWorkspaceLoaderFactory.create(&cwd.join(&repository)) {
+            let _ = layered_configs.read_repo_config(loader.repo_path());
```
yuja, line 215: nit: propagate config error consistently?

I think it's also good to suppress all `user`/`repo_config` errors.

cli/src/complete.rs
```diff
@@ -116,17 +117,42 @@ fn get_jj_command() -> Result<(std::process::Command, Config), CommandError> {
...
+
+/// `std::process::Command` can be reused to spawn multiple processes, but only
+/// with the same arguments or additional arguments. Completion functions may
+/// need to spawn multiple jj processes with completely different arguments, so
+/// they need the ability to recreate the pre-configured command from scratch.
+struct JjBuilder {
+    current_exe: std::path::PathBuf,
+    args: Vec<String>,
+}
+impl JjBuilder {
```
yuja, line 241: nit: we usually insert blank lines between `fn`, `struct`, `impl`s.

cli/src/complete.rs
senekor: I suppressed these additional errors:
```rust
let maybe_cwd_workspace_loader = DefaultWorkspaceLoaderFactory.create(find_workspace_dir(&cwd));
let _ = layered_configs.read_user_config();
if let Ok(loader) = &maybe_cwd_workspace_loader {
    let _ = layered_configs.read_repo_config(loader.repo_path());
}
```
Let me know if there are more you think should be suppressed.

cli/src/complete.rs
```diff
@@ -92,11 +200,9 @@ fn get_jj_command() -> Result<(std::process::Command, Config), CommandError> {
         .and_then(|cwd| cwd.canonicalize())
         .map_err(user_error)?;
     let maybe_cwd_workspace_loader = DefaultWorkspaceLoaderFactory.create(find_workspace_dir(&cwd));
-    layered_configs.read_user_config().map_err(user_error)?;
+    let _ = layered_configs.read_user_config();
     if let Ok(loader) = &maybe_cwd_workspace_loader {
-        layered_configs
-            .read_repo_config(loader.repo_path())
-            .map_err(user_error)?;
+        let _ = layered_configs.read_repo_config(loader.repo_path());
```
yuja, line 205: nit: can you move these hunks to the first (or separate) commit?

cli/src/complete.rs
```diff
@@ -44,6 +46,112 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+            .arg(r#"if(remote && !tracked, name ++ "@" ++ remote ++ "\n")"#)
+            .output()
+            .map_err(user_error)?;
+
+        Ok(String::from_utf8_lossy(&output.stdout)
+            .lines()
+            .filter(|bookmark| !bookmark.ends_with("@git"))
+            .map(|bookmark| {
+                let mut display_order = 0;
+                if let Ok(prefix) = config.get::<String>("git.push-bookmark-prefix") {
```
yuja, line 83: nit: maybe better to resolve config value out of the loop

cli/src/complete.rs
```diff
@@ -44,6 +46,112 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+                    if !bookmark.starts_with(&prefix) {
+                        // other people's bookmarks are less interesting
+                        display_order += 1;
```
yuja, line 86: nit: do we need to assign unique display order?

(Just asking because `bookmarks()` appears to assign static weights.)

cli/src/complete.rs
```diff
@@ -44,6 +46,112 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+            .map_err(user_error)?;
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        let prefix = config.get::<String>("git.push-bookmark-prefix").ok();
+        let mut bookmarks = HashMap::new();
+
+        // To determine the importance of each bookmark, we figure out if it is
+        // tracked locally and if its push prefix matches our configured one.
+        for line in stdout.lines() {
+            let (bookmark, _) = line.split_once('@').unwrap_or((line, ""));
```
yuja, line 113: nit: better to use the split result instead of testing `line.contains('@')` again.

cli/src/complete.rs
```diff
@@ -44,6 +46,112 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+        // display order (0 is highest priority).
+        let mut ordered_bookmarks = vec![vec![]; 4];
+        for (bookmark, (local, mine)) in bookmarks.into_iter() {
+            match (local, mine) {
+                (true, true) => ordered_bookmarks[0].push(bookmark),
+                (true, false) => ordered_bookmarks[1].push(bookmark),
+                (false, true) => ordered_bookmarks[2].push(bookmark),
+                (false, false) => ordered_bookmarks[3].push(bookmark),
+            }
+        }
```
yuja, line 139: nit: Perhaps, we can use `chunk_by()` to group bookmarks by name, and map to `(name, weight)`.
https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.chunk_by

cli/src/complete.rs
```diff
@@ -44,6 +46,112 @@ pub fn local_bookmarks() -> Vec<CompletionCandidate> {
...
+            let (mut local, mut mine) = *properties;
+
+            if !line.contains('@') {
+                local = true;
+            }
+            if let Some(prefix) = prefix.as_ref() {
+                if bookmark.starts_with(prefix) {
+                    mine = true;
+                }
+            }
```
yuja, line 124: nit: `.is_soem_and(|prefix| ...)` also works

yuja (approved):
Thanks.

cli/src/complete.rs
senekor: I thought about something like this, but I was unsure if it's absolutely guaranteed that `jj bookmark list` will return them grouped like that? I guess not grouping them would be considered worse UX, so it probably won't happen.

yuja: It's unlikely we'll change the output order because the default `jj bookmark list` output relies on that (local name followed by remotes). Also, other completion functions assume that the output is sorted by name.

PR #4822, opened 2024-11-10
docs: migrate from Poetry to uv

neongreen:
# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [x] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

---

`uv` is a fast Python project manager written in Rust. It can automatically install Python, or use a global version if it satisfies the projects requirements.

After the switch, contributors don't need to install Python to build the docs. The single command `uv run mkdocs serve` will work.

I also removed the Poetry-specific keyring workarounds. if someone uses the keyring to build the docs, I'd like them to try out this PR and report any issues.

PR #4829, opened 2024-11-11
infra: automatically post documentation previews for PRs

neongreen:
This adds a new workflow that renders docs, publishes them to https://jj-preview.github.io/docs/, and posts a link to the PR thread.

Ignore all commits except for the last one. (uv changes are #4822)

How it looks in action: https://github.com/neongreen/jj/pull/1

image

# Checklist

If applicable:


martinvonz:
@ilyagr, you know much more about the doc generation than I do. Will you have time to review this?

ilyagr:
I will take a look, yes.

.github/workflows/docs-preview-deploy.yml
```diff
@@ -0,0 +1,93 @@
...
+          folder: rendered-docs
+          repository-name: jj-preview/docs
+          branch: main
+          clean: true
+          target-folder: 'pr-${{ env.PR_NUMBER }}'
+          # This key allows pushing to jj-preview/docs. How to set it up:
+          #
+          #   1. ssh-keygen -t ed25519 -C "" -N "" -f <where to put the keypair>
+          #   2. Add the public key to https://github.com/jj-preview/docs/settings/keys/new
+          #   3. Add the private key to https://github.com/martinvonz/jj/settings/secrets/actions/new
```
ilyagr, line 77: I'm guessing only @martinvonz can do this, right?

Speaking of which, I just invited Martin to the jj-preview org. It's a party now. :)

.github/workflows/docs-preview-cleanup.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Deploy
+        uses: JamesIves/github-pages-deploy-action@62fec3add6773ec5dbbf18d2ee4260911aa35cf4
+        with:
+          folder: empty-dir
+          repository-name: jj-preview/docs
+          branch: 'main'
+          clean: true
+          target-folder: 'pr-${{ github.event.pull_request.number }}'
+          ssh-key: ${{ secrets.DOCS_DEPLOY_KEY }}
```
ilyagr, line 28: Might make sense to add a comment: `# Creating this is documented in docs-preview-deploy.yml`.

.github/workflows/docs-preview-cleanup.yml
ilyagr, line 1: I'm not sure this is possible, but could we put all of these in their own dir, so that this be `.github/workflows/docs-preview/cleanup.yml`?

.github/workflows/docs-preview-deploy.yml
```diff
@@ -0,0 +1,93 @@
...
+          commit-message: "Update preview for PR ${{ env.PR_NUMBER }}"
+          single-commit: true
+
+      - name: Post a link to the docs in the PR
+        uses: thollander/actions-comment-pull-request@65f9e5c9a1f2cd378bd74b2e057c9736982a8e74
+        with:
+          message: |
+            **📖 Documentation preview is published to https://jj-preview.github.io/docs/pr-${{ env.PR_NUMBER }}.**
+            Thanks for working on the docs!
+
```
ilyagr, line 90: I'd add:

> This feature is experimental, feel free to open a bug or tell us in a chat if something is not working.
>
> You can always build and serve the docs on your own machine by following the instructions at <https://martinvonz.github.io/jj/prerelease/contributing/#previewing-the-html-documentation>

.github/workflows/docs-preview-build.yml
```diff
@@ -0,0 +1,49 @@
...
+#
+# The split into "build" and "deploy" follows GitHub's guide at
+# https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/.
+# The build workflow has no access to the repo secrets. The deploy workflow runs
+# in the context of the main repo rather than the PR branch.
+
+name: docs-preview-build
+
+on:
+  pull_request:
```
ilyagr, line 16: Will this trigger when a pull request changes? If not, is it possible to provide a way to trigger it manually?

.github/workflows/docs-preview-build.yml
```diff
@@ -0,0 +1,49 @@
...
+# This workflow builds the docs for each PR (including from forks), and uploads
+# the rendered docs as an artifact.
+#
+# The artifact will be picked up by docs-preview-deploy.yml and uploaded to
+# https://jj-preview.github.io/docs/.
+#
+# The split into "build" and "deploy" follows GitHub's guide at
+# https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/.
+# The build workflow has no access to the repo secrets. The deploy workflow runs
+# in the context of the main repo rather than the PR branch.
```
ilyagr, line 11: Thanks, I learned a lot about GitHub actions from this PR.

neongreen: tbh after making this PR I am motivated to learn about other CIs. I used to think they were all roughly the same nowadays, and the biggest annoyance was yaml; now I think yaml is the smallest annoyance. I wish for a feature-complete CI 🥲

ilyagr: Somewhere in Docusaurus docs, they recommend using Netlify or Vercel instead of GitHub pages since (they say) it's easier to do PR previews with them.

No idea whether it's better now that you've done the work. Also, we probably should wait to have a `jj` domain before jumping around providers, so that it doesn't affect URLs too much.

neongreen: Netlify and Vercel are both infinitely easier than GH Pages. But there are some downsides:

* Vercel's free plan does not support teams
* Netlify is reported to be costly in general nowadays
* Cloudflare Pages has a better free plan, but they only give you 200 builds per month

And all of those would require setting up a new account, getting someone (Martin?) to add a payment method, etc.

The jj domain thing doesn't apply here, though — we don't want previews to be on the same domain anyway. So jumping between providers is fine; Vercel/Netlify would give us is an autogenerated preview domain and all new previews would land there.

.github/workflows/docs-preview-deploy.yml
neongreen: Yeah, Martin needs to add a GH secret. Or give someone a "CI/CD admin" role:

> https://github.blog/changelog/2024-09-25-introducing-ci-cd-admin-a-new-pre-defined-organization-role-for-github-actions/
>
> With the new CI/CD Admin role, organization owners and teams can now delegate comprehensive CI/CD management to individuals without the need to maintain a custom role

.github/workflows/docs-preview-cleanup.yml
neongreen: I think it's not possible — https://github.com/orgs/community/discussions/15935

.github/workflows/docs-preview-deploy.yml
martinvonz: I've added a secret called `DOCS_DEPLOY_KEY`. Its public key is:
```
ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIH5kNau/89p02CNJy0I+ijuE0qZUp+HjhZgZLtJ64m9y
```

.github/workflows/docs-preview-cleanup.yml
ilyagr: Oh well. Thanks for checking.

.github/workflows/docs-preview-cleanup.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+
+      - run: mkdir -p empty-dir
+
+      - name: Deploy
+        uses: JamesIves/github-pages-deploy-action@62fec3add6773ec5dbbf18d2ee4260911aa35cf4
+        with:
+          folder: empty-dir
```
ilyagr, line 23: (Thinking out loud, we can fiddle with this later if necessary) This might create a lot of empty dirs.

I don't think it matters much, but perhaps a better approach would be:

- To add a file called `PR_CLOSED` to the dir here
- Have an action in the docs repo that deletes all dirs with PR_CLOSED marker if it's over 7 days old.

I'm not sure how well that will work if that action happens to happen while another action is deploying to the docs. But I'm also not sure it'd be worse than with this PR as is; I'm not sure how this PR will do if the cleanup action on one PR is running at the same time as the deploy action acts on another PR.

It is something we can discover experimentally, though.

neongreen: > This might create a lot of empty dirs.

It actually deletes the dir entirely (I checked).

> I'm not sure how well that will work if that action happens to happen while another action is deploying to the docs.

Good catch. In fact, GHA doesn't support a "workflow queue" at all — if you use `concurrency:`, it will only keep one pending job in the queue.

I will fix that by disabling `force` and `single-commit` and switching to per-PR concurrency groups.

neongreen:
I just spent several hours reading about GHA security and learned (among other things) that all workflows have access to repo secrets.

We can mitigate that by creating a "deployment environment", scoping it to the `main` branch, and moving the secret there.

- Move the deploy key (can call the environment eg. "preview") — https://github.com/martinvonz/jj/settings/environments/new
- Configure the environment to only allow deploys from the main branch
- I will change the deploy/cleanup workflows to require `environment: preview`

.github/workflows/docs-preview-cleanup.yml
neongreen: Update: I will also delete the cleanup entirely b/c the extra complexity is not worth it

martinvonz:
> I just spent several hours reading about GHA security and learned (among other things) that all workflows have access to repo secrets.

Thank you! I've moved the private key to the new `preview` environment now (and removed it from the project's top level).

.github/workflows/docs-preview-cleanup.yml
ilyagr: (Thinking out loud) Would just adding the `PR_CLOSED` file (or something like that) be too complex as well? Then we could occasionally clean up manually.

Unless we nuke the repo history, garbage collection won't save much git storage, but it might save on the amount of traffic needed to deploy the website. OTOH, GitHub seems to take care of that, so maybe it's not a problem (or we could assume that they are smart about it and use `rsync`).

Ultimately, this is up to you.

neongreen:
GitHub closed this PR b/c I renamed the branch in my fork. 

I’ll make a new PR after fixing a few papercuts in the new version.

PR #4831, opened 2024-11-11
cli: rebase: extract out common `print_move_commits_stats` function

bnjmnt4n:
This commit extracts out the common code printing out the `MoveCommitsStats` information into a shared function. The printed output was also inconsistent between `-r` and `-s`/`-b` code paths, so I standardized it to say "Rebased ? commits onto destination" for both cases.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/rebase.rs
+++ b/cli/src/commands/rebase.rs
@@ -409,12 +409,7 @@ fn rebase_descendants_transaction(
         )
     };
 
-    let MoveCommitsStats {
-        num_rebased_targets,
-        num_rebased_descendants,
-        num_skipped_rebases,
-        num_abandoned,
-    } = move_commits(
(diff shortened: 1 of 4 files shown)


cli/src/commands/rebase.rs
```diff
@@ -618,3 +565,38 @@ fn check_rebase_destinations(
     }
     Ok(())
 }
+
+/// Print details about the provided [`MoveCommitsStats`].
+fn print_move_commits_stats(ui: &Ui, stats: MoveCommitsStats) -> std::io::Result<()> {
```
yuja, line 570: nit: pass `&MoveCommitsStats` by reference (and destructure as `let &MoveCommitsStatus ..`

yuja (approved):

PR #4832, opened 2024-11-11
completion: teach jj about aliases

senekor:
# Checklist

If applicable:


cli/src/complete.rs
```diff
@@ -106,10 +122,23 @@ fn get_jj_command() -> Result<std::process::Command, CommandError> {
...
         .disable_help_flag(true)
         .ignore_errors(true)
+        // Here, allow_external_subcommands fixes a weird issue. Without it,
+        // parsing GlobalArgs will fail with the message that a required arg
+        // is missing, where the required arg is a boolean flag. This seems
+        // unexpected, because missing boolean flags are usually treated as
+        // false. It is also not clear to me why allow_external_subcommands
+        // changes this behavior. See the discussion in the clap repo:
+        // https://github.com/clap-rs/clap/discussions/5812
+        .allow_external_subcommands(true)
```
senekor, line 132: There's an incoming patch for this in clap:
https://github.com/clap-rs/clap/pull/5813

Let's wait with this PR until we can get rid of this workaround.

ilyagr:
Re commit descriptions: at a glance, it's hard to tell how "teach jj about alias" is different from "resolve aliases" from 325402d . I'm sure it'd be clearer if I read the code, but I'd still appreciate a short explanation in the description (assuming this PR indeed does something different).

senekor:
I added an explanation to the description, let me know if it can still be improved.

cli/src/complete.rs
senekor: The fix in clap is merged and released.

cli/tests/test_completion.rs
```diff
@@ -242,3 +242,47 @@ fn test_remote_names() {
...
+    test_env.jj_cmd_ok(
+        &repo_path,
+        &[
+            "config",
+            "set",
+            "--repo",
+            "aliases.repo-alias",
+            "['bookmark']",
+        ],
+    );
```
martinvonz, line 264: no action needed: I don't think we really need to test both user- and repo-level configs. The code doesn't seem to care about it, and it seems unlikely to regress.

martinvonz (approved):

cli/tests/test_completion.rs
senekor: Makes sense, I did remove that test.

PR #4833, opened 2024-11-11
`jj git colocate`

### LINKED ISSUE #4624: FR: Add a `jj git colocate` command

cormacrelf:
Fixes #4624 

Open to bikesheds on what to call the inverse command. `jj git delocate`?

# Checklist

If applicable:


joyously:
> Open to bikesheds on what to call the inverse command. `jj git delocate`?

Would it be `isolate`?

neongreen:
`jj git dismantle`? `jj git purify`?

cormacrelf:
If you're turning off git, you are really turning a page on what once was. One might type it with a renewed sense of purpose, confidence in your knowledge of the JJ way, and independence of the crutch you once needed. In that vein I suggest

- `jj git renounce`, my favourite; or
- `jj git disavow`

martinvonz:
`jj git internalize` might be okay. I think of the operation as making the `.git/` dir internal to `.jj/`. But it's not as cheeky as some of the other options :)

emilazy:
`jj git conceal` keeps everything nicely under a co‐ namespace. (Not serious. Or am I?)

joyously:
`separate` comes to mind. 
Or `jj git colocate --reverse`
I think of most of the `git` subcommands as actual Git commands, but this is not, so it seems a bit different. (both `colocate` and its reverse)

lib/src/workspace.rs
```diff
@@ -357,6 +357,30 @@ impl Workspace {
...
     }
 
+    pub fn reload_store(
+        &self,
+        workspace_loader: &dyn WorkspaceLoader,
+        user_settings: &UserSettings,
+        working_copy_factory: &dyn WorkingCopyFactory,
+        store_factories: &StoreFactories,
+    ) -> Result<Self, WorkspaceLoadError> {
+        // Load the new repo from the file we just wrote.
```
cormacrelf, line 367: xxx

essiene:
> `separate` comes to mind. Or `jj git colocate --reverse` I think of most of the `git` subcommands as actual Git commands, but this is not, so it seems a bit different. (both `colocate` and its reverse)

I really like this since unlike other suggestions, I don't have to think about what this does.

If there *must* be a command under `jj git`, a close second of `colocate --reverse` would be `jj git decolocate`, which is not a real word to be sure, but also communicates intent clearly.

fwiw ;-)

cli/src/commands/git/colocate.rs
```diff
@@ -0,0 +1,142 @@
...
+            .open(&config_path)
+            .context(config_path)?;
+        config
+            .write_to(&mut file)
+            .map_err(|e| internal_error(format!("Could not write git config file: {e}")))?;
+        drop(file);
+
+        let old_path = workspace_command.repo_path().join("store").join("git");
+        let new_path = workspace.workspace_root().join(".git");
+        fs::rename(&old_path, &new_path).context(old_path)?;
```
cormacrelf, line 85: Ah, sorry, we have to check that old_path exists (and is a directory, and is not a symlink, and also that git_target points to it) before we commit to writing out core.bare=false.

As long as we check this, I don't think we need to handle the large variety of git_target scenarios outside of this, because there will soon be an alternative, which is to just add a worktree. Which will work in many more scenarios. git_target can point into outer space for all we care.

The idea is to have `--prefer-worktree` switch to using worktrees in the primary workspace, as well as secondary ones.

cormacrelf: Done

PhilipMetzger:
> Open to bikesheds on what to call the inverse command. `jj git delocate`?

I'd like to have the inverse in the native backend with something like `adopt` but that can wait. Otherwise `internalize` is a fine name.

cli/src/commands/git/colocate.rs
```diff
@@ -0,0 +1,171 @@
...
+    let dotgit_path = workspace.workspace_root().join(".git");
+    if dotgit_path.exists() {
+        return Err(user_error(format!(
+            "Path {} already exists, cannot colocate this workspace.",
+            dotgit_path.display()
+        )));
+    }
+
+    let git_repo = git_backend.git_repo();
+    if workspace.is_primary_workspace() {
```
yuja, line 76: Can we detect this somehow by caller or git layer? 

I don't think "is primary workspace" can be a common concept across working-copy/workspace-loader backends.

cli/src/commands/git/colocate.rs
```diff
@@ -0,0 +1,171 @@
...
+    ui: &mut Ui,
+    command: &CommandHelper,
+    _args: &GitColocateArgs,
+) -> Result<(), CommandError> {
+    let workspace_command = command.workspace_helper(ui)?;
+    let workspace = workspace_command.workspace();
+
+    if workspace_command.working_copy_shared_with_git() {
+        return Err(user_error(format!(
+            "Workspace '{}' is already colocated.",
```
yuja, line 56: nit: error message doesn't usually contain period

cli/src/commands/git/colocate.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    let git_repo = git_backend.git_repo();
+    if workspace.is_primary_workspace() {
+        let common_dir = git_repo.common_dir();
+        let old_path = workspace_command.repo_path().join("store").join("git");
+        let common_dir_canon = common_dir.canonicalize().context(common_dir)?;
+        let old_path_canon = old_path.canonicalize().ok();
+
+        let mut config =
+            gix::config::File::from_git_dir(common_dir.to_path_buf()).map_err(internal_error)?;
```
yuja, line 83: nit: maybe this should be a user error (assuming it would do some I/O)?

BTW, the gix' fetch/clone implementation might help understand their config API. I'm not familiar with that.
https://github.com/GitoxideLabs/gitoxide/blob/main/gix/src/clone/fetch/util.rs

cli/src/commands/git/colocate.rs
```diff
@@ -0,0 +1,171 @@
...
+            "Unimplemented: colocating a secondary workspace",
+        ));
+    }
+
+    // Both the Workspace (i.e. the store) and the WorkspaceCommandHelper
+    // need to be reloaded to pick up changes to colocation.
+    //
+    // This way we end up with a git HEAD written immediately, rather than
+    // next time @ moves. And you can immediately start running git commands.
+    let workspace = workspace
```
yuja, line 132: Maybe we can use `command.workspace_helper_no_snapshot()`?

cli/src/commands/git/colocate.rs
```diff
@@ -0,0 +1,171 @@
...
+            return Ok(());
+        }
+    }
+
+    let git2_repo = git_backend.open_git_repo()?;
+    let mut tx = command.start_transaction();
+
+    git::reset_head(tx.repo_mut(), &git2_repo, &wc_commit)?;
+
+    tx.finish(ui, format!("Colocated existing workspace {name}"))?;
```
yuja, line 168: nit: `Colocated` -> `colocate`

cli/src/commands/git/colocate.rs
```diff
@@ -0,0 +1,171 @@
...
+        let old_path_canon = old_path.canonicalize().ok();
+
+        let mut config =
+            gix::config::File::from_git_dir(common_dir.to_path_buf()).map_err(internal_error)?;
+        let is_bare = config
+            .value::<gix::config::Boolean>("core.bare")
+            .map_or(false, |x| x.is_true());
+
+        if old_path.is_symlink()
+            || !old_path.is_dir()
```
yuja, line 89: nit: use `.symlink_metadata()` instead of `stat()`ing twice?

yuja:
> `jj git colocate --reverse`

I don't like the name `--reverse`, but it's nice that we don't have to add two separate commands which would be rarely used.

camsteffen:
subsume?

bryango:
I would suggest `jj git relocate` with flags `--colocate` and the reverse, perhaps `--conceal` or whatever. The `--colocate` flag then coincides with the flag for `jj git init`.

KingMob:
> Open to bikesheds on what to call the inverse command

As joyously pointed out, many of the existing `jj git *` commands mirror existing git commands, and colocate is an oddball. If we wanted to strengthen the perception that `jj git` is for existing git commands, what about moving colocation under `util`?

It could be `jj util dotgit --colocate` or `jj util dotgit --decolocate` (or show/hide/conceal/relocate/etc). Just my $.02.

----

Otherwise, I agree that such a rare operation shouldn't have a dedicated word. I'm in favor of `jj git colocate --reverse`.

PhilipMetzger:
> If we wanted to strengthen the perception that `jj git` is for existing git commands, what about moving colocation under `util`?

I disagree on that, as colocating a Git repo should be visible to many users. The `util` namespace is for _our_ utilities. 

> It could be `jj util dotgit --colocate` or `jj util dotgit --decolocate` (or show/hide/conceal/relocate/etc). Just my $.02.

If there were a transparent `.git` mode, then it also could have a place there. But that only came up once a while ago in the Discord.

tim-janik:
Wouldn't it be simplest to just have:

```sh
jj git init --colocate
jj git init --bare
```

And instead of throwing an error if the repo exists, ensure it becomes bare/colocated?

jennings:
Maybe this is a dumb suggestion, but what if we called this `jj git reinit` instead?

- `jj git reinit --colocate` - Convert to colocated
- `jj git reinit --no-colocate` - Convert to non-colocated 

Or, if we're feeling _really wild_, we could overload `jj git init` to do this if we're already in a repo.

KingMob:
A new user might worry that running `jj git reinit` or rerunning `jj git init` would do other, destructive things that maybe they didn't want, like erasing all history and resetting the repo. I think `init` implies too many other things.

nrbray:
> > `jj git colocate --reverse`
> 
> I don't like the name `--reverse`, but it's nice that we don't have to add two separate commands which would be rarely used.

```jj git colocate --remove``` would be my suggestion, as it removes the presence of git from any existing git aware tooling and is the most significant effect from my point of view

**Comment by nrbray** [comment]
> > > `jj git colocate --reverse`
> > 
> > 
> > I don't like the name `--reverse`, but it's nice that we don't have to add two separate commands which would be rarely used.
> 
> `jj git colocate --remove` would be my suggestion, as it removes the presence of git from any existing git aware tooling and is the most significant effect from my point of view

Or alternatively

```jj git locate --legacy```
```jj git locate --internal```

**Comment by nrbray** [comment]
> I really like `colocate --reverse` since unlike other suggestions, I don't have to think about what this does.

I think you made the most relevant comment.  Also the effect is to remove the presence of git from the perspective of legacy tools.  So you inspired me to suggest:

```
jj git locate --legacy
jj git locate --internal
```

samueltardieu:
Back in October, I had a need to colocate/"uncolocate" a lot of repos, so I wrote a [half-backed implementation](https://github.com/samueltardieu/jj/blob/git-colocate/cli/src/commands/git/colocate.rs) of it which does not deal with workspaces (more exactly which hasn't checked if it works or not in secondary workspaces), so I haven't finished or submitted it.

I chose `jj git colocate` and `jj git colocate --undo` which seemed clear enough to me.

necauqua:
Can't believe nobody suggested `jj git dislocate`

PhilipMetzger:
@cormacrelf are you willing to rebase and land this? I think the bikeshedding for the other command name should come whenever that is created.

neandrake:
For the inverse command I suggest `jj git vacate`.

AngelEzquerra:
FYI, I was unaware of this PR so I created a similar one in #7392 . The main difference compared to this one is that my PR also allows you to undo a colocation which according to some comments in jujutsu's discord people find valuable. However, my PR does not really try to support workspaces.
I'd like at least one of these 2 PRs to be accepted. @cormacrelf, are you going to try to push this one through?

algmyr:
Obsoleted by https://github.com/jj-vcs/jj/pull/7392

Issue #4784, opened 2024-11-06
jujutsu test failure on Nix install 


mitchellh:
## Steps to Reproduce the Problem

I've added the jujutsu flake and overlay to my flake as described.

This previously has worked for many versions, it only recently started failing. I'm sorry I haven't bisected it.

## Actual Behavior

```
error: builder for '/nix/store/q0g3dzx7qdx2vz6rw9qx8sc6b7prfkap-jujutsu-unstable-0a73245.drv' failed with exit code 100;
       last 25 log lines:
       >  right: [CommitId("0000000000000000000000000000000000000000")]
       > stack backtrace:
       >    0: rust_begin_unwind
       >              at /rustc/e92993dbb43f0a5d17fe56e2d82f90435d6521c8/library/std/src/panicking.rs:665:5
       >    1: core::panicking::panic_fmt
       >              at /rustc/e92993dbb43f0a5d17fe56e2d82f90435d6521c8/library/core/src/panicking.rs:74:14
       >    2: core::panicking::assert_failed_inner
       >    3: core::panicking::assert_failed
       >    4: core::ops::function::FnOnce::call_once
       >    5: core::ops::function::FnOnce::call_once
       >              at /rustc/e92993dbb43f0a5d17fe56e2d82f90435d6521c8/library/core/src/ops/function.rs:250:5
       > note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
       >
       >    Canceling due to test failure: 5 tests still running;1mpassed, 0 skipped
       >         PASS [   0.016s] jj-lib::runner test_gpg::invalid_signaturefailed, 0 skipped
       >         PASS [   0.020s] jj-lib::runner test_gpg::unknown_keyiled, 0 skipped
       >         PASS [   0.018s] jj-lib::runner test_id_prefix::test_id_prefix 0 skipped
       >         PASS [   0.012s] jj-lib::runner test_id_prefix::test_id_prefix_divergent skipped
       >         PASS [   0.024s] jj-lib::runner test_gpg::gpg_signing_roundtrip0 skipped
       >         PASS [   0.025s] jj-lib::runner test_gpg::gpg_signing_roundtrip_explicit_key3;1mskipped
       > ------------[0m [ 00:00:08] [] 1189/1421: 1 running, 1188 passed, 1 failed, 0 skipped
       >      Summary [   8.474s] 1190/1421 tests run: 1189 passed, 1 failed, 0 skipped
[...]

PR #4839, opened 2024-11-12
nix: skip `test_shallow_commits_lack_parents` under Nix builds

thoughtpolice:
While this isn't a real solution to #4784 it does stem the bleeding for anyone like Mitchell (or myself) who have to install the flake from the Git repository. This patch also probably needs to be backported to upstream nixpkgs, too.

The `flake.lock` update is so that we can get a newer version of `cargo-nextest` that supports the `--skip` flag (there is a range of nextest versions that don't support `--skip` after the migration to use 'filtersets' was introduced.)

# Checklist

If applicable:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -77,6 +77,14 @@
           buildFeatures = [ "packaging" ];
           cargoBuildFlags = [ "--bin" "jj" ]; # don't build and install the fake editors
           useNextest = true;
+
+          checkFlags = [
+            # https://github.com/martinvonz/jj/issues/4784
+            # seems to fail reliably under Nix/nix-shell only, but otherwise
+            # is an unknown anomaly?
+            "--skip" "test_shallow_commits_lack_parents"
+          ];
+          
           src = filterSrc ./. [
             ".*\\.nix$"
             "^.jj/"
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1728538411,
-        "narHash": "sha256-f0SBJz1eZ2yOuKUr5CA9BHULGXVSn6miBuUWdTyhUhU=",
+        "lastModified": 1730958623,
+        "narHash": "sha256-JwQZIGSYnRNOgDDoIgqKITrPVil+RMWHsZH1eE1VGN0=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "b69de56fac8c2b6f8fd27f2eca01dcda8e0a4221",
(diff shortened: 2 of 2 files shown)


thoughtpolice:
I have spent some time on #4784 but haven't had anything conclusive and I was out of town this weekend. I hope to have more time to look at this if Yuya doesn't in the mean time, but otherwise this at least unbreaks people's flakes/overlays.

flake.nix
```diff
@@ -77,6 +77,14 @@
           buildFeatures = [ "packaging" ];
           cargoBuildFlags = [ "--bin" "jj" ]; # don't build and install the fake editors
           useNextest = true;
+
+          checkFlags = [
+            # https://github.com/martinvonz/jj/issues/4784
+            # seems to fail reliably under Nix/nix-shell only, but otherwise
+            # is an unknown anomaly?
+            "--skip" "test_shallow_commits_lack_parents"
```
emilazy, line 85: The nextest native syntax would involve `--filterset`, and AIUI you need a `--` for `--skip` – does this actually work to skip it?

thoughtpolice: I'm aware of the filterset expression, but I chose this in part because I think `--skip` reads much more naturally (part of the reason it was presumably added back to nextest, along with the fact it's backwards compatible with older versions; when I originally tried diagnosing this I had to spend 5 minutes figuring out why `--skip` was reworked and the right way to skip tests, which involved [a nice bug report](https://github.com/nextest-rs/nextest/issues/1816).) Regarding the `--` yes, it's because the actual Nix harness around `cargo test` adds it for you. I've never seen `--skip` used in any other way in any Rust+Nix project FWIW, so I've assumed this was always the default behavior.

emilazy (approved):

bnjmnt4n:
This test does seem somewhat flaky even in non-Nix environments - example of a failing GitHub actions run: https://github.com/martinvonz/jj/actions/runs/11792819629/job/32847132523. Perhaps we should disable it on Rust's end?

yuja:
> This test does seem somewhat flaky even in non-Nix environments - example of a failing GitHub actions run: https://github.com/martinvonz/jj/actions/runs/11792819629/job/32847132523. Perhaps we should disable it on Rust's end?

Maybe we need to reload the `repo` after `make_shallow()`? `gix` holds in-memory cache which is invalidated by file mtime.

yuja:
https://github.com/martinvonz/jj/pull/4842

thoughtpolice:
> #4842

Yes, I figured it was something like that just like `test_gc` was flaky, but I've been a bit sick and busy all weekend so I didn't get time to ensure that it was the same problem. I'll test this out.

PR #4847, opened 2024-11-13
completion: improve docs for activation

senekor:
motivated by [this discussion](https://github.com/martinvonz/jj/pull/4737#discussion_r1839402140)

# Checklist

If applicable:

### DIFF
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -210,12 +210,13 @@ To set up command-line completion, source the output of
 `jj util completion bash/zsh/fish`. Exactly how to source it
 depends on your shell.
 
-Improved completions are currently in the works, these will complete things
-like bookmark names as well. You can activate them with the alternative "dynamic"
-instructions below. Please let us know if you encounter any issues, so we can
-ensure a smooth transition once we default to these new completions. If you
-have ideas about specific completions that could be added, please share them
-[here](https://github.com/martinvonz/jj/issues/4763).
+Improved completions are currently in the works, these will complete things like
+bookmark names as well. You can activate them with the alternative "dynamic"
+instructions below. They should still complete everything the static completions
+did, so only activate one of them. Please let us know if you encounter any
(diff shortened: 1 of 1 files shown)


docs/install-and-setup.md
```diff
@@ -210,12 +210,13 @@ To set up command-line completion, source the output of
...
-like bookmark names as well. You can activate them with the alternative "dynamic"
-instructions below. Please let us know if you encounter any issues, so we can
-ensure a smooth transition once we default to these new completions. If you
-have ideas about specific completions that could be added, please share them
-[here](https://github.com/martinvonz/jj/issues/4763).
+Improved completions are currently in the works, these will complete things like
+bookmark names as well. You can activate them with the alternative "dynamic"
+instructions below. They should still complete everything the static completions
+did, so only activate one of them. Please let us know if you encounter any
+issues, so we can ensure a smooth transition once we default to these new
```
martinvonz, line 217: When should I (as a user) prefer to use the static completions?

senekor: - You don't care about dynamic completions, so you have no desire to be beta-tester for them.
- You already encountered some issue with the dynamic ones, so you switch back to fix your workflow. (hopefully you also report the issue upstream)

senekor: I have obviously been using the dynamic completions myself the last couple of days and I haven't experienced any regressions with respect to the static ones. (using fish)

Do you think we should advertise the dynamic completions as _default_, while still providing the static ones as fallback if people run into issues? That would probably lead to more people trying it out.

martinvonz: > Do you think we should advertise the dynamic completions as _default_, while still providing the static ones as fallback if people run into issues? That would probably lead to more people trying it out.

Yes, that's what I was thinking, but Yuya [reported](https://github.com/martinvonz/jj/pull/4737#discussion_r1839702085) a bug in the Zsh version already, so we probably should not do that yet.

martinvonz (approved):

PR #3141, opened 2024-02-25
Signature verification

julienvincent:
This is a continuation from #3007 and #2728 which adds in the templater changes for verifying and displaying commit signatures.

---

There were some outstanding comments on #3007 which I am listing below:

- [x] https://github.com/martinvonz/jj/pull/3007#discussion_r1492197725
- [ ] https://github.com/martinvonz/jj/pull/3007#discussion_r1492225610 - I understand the conclusion to be that this could be addressed later.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [x] I have updated the documentation (README.md, docs/, demos/)
- [x] I have updated the config schema (cli/src/config-schema.json)
- [x] I have added tests to cover my changes

PR #4853, opened 2024-11-13
Add templater support for rendering commit signatures

pylbrecht:
These changes are continuing the work of #3141.
I took over its changes and rebased them on latest `main`.

This work is required for #4747, as it lays the foundation for writing CLI tests for `jj sign`.

# Checklist

If applicable:


bnjmnt4n:
Is there anything blocking this PR? It seems to be in a functional state.

pylbrecht:
> Is there anything blocking this PR? It seems to be in a functional state.

Well, since I simply materialized the changes from @julienvincent and @necauqua, there are still two things missing:
* printing hints about dropped commit signatures https://github.com/jj-vcs/jj/pull/3141/commits/9eaea83e1a16c230d7dc9181faacf7b2776302da
* docs https://github.com/jj-vcs/jj/pull/3141/commits/739ddf246b183a1835b8b9b6d30f59c603c36d08

Regarding printing hints about dropped signatures, I reached out for help (https://github.com/jj-vcs/jj/issues/4712#issuecomment-2537139727, https://github.com/jj-vcs/jj/issues/4712#issuecomment-2538012701).

Maybe we *could* merge without the hints (and create a new issue to track that missing piece to be implemented later)? Then I would focus on getting the docs ready.

bnjmnt4n:
Based on the PR title, I would presume that this PR is focused mainly on being able to display commit signatures. So it seems fine to me if printing hints about dropped commit signatures is shifted to a different issue/PR, since it isn't directly related to displaying commit signatures.

pylbrecht:
Sounds good to me.
I will get the PR ready and if nobody objects until then, we can merge without the hints.

pylbrecht:
I pushed a slightly modified version of https://github.com/jj-vcs/jj/pull/3141/commits/739ddf246b183a1835b8b9b6d30f59c603c36d08. However, I believe it was planned to add a `builtin_log_compact_with_sig`, which looks like this for good signatures (fancy checkmarks):
```
    Rebased 1 descendant commits
    Working copy now at: rlvkpnrz b162855d (empty) (no description set)
    Parent commit      : qpvuntsm [✓︎] 5aab9df2 (empty) init
    Commit was signed: qpvuntsm [✓︎] 5aab9df2 (empty) init
```
https://github.com/jj-vcs/jj/pull/3142/files#diff-edf6d251f71a7d4a3b3b8cf73345ea16e6dc7691ed1b98a1811d5709cbe011d2R47-R50)

@necauqua, originally this was your work, I think. Was it planned to add `builtin_log_compact_with_sig`?
(I cross-read a bunch of issues / PRs in the past, but I am too tired right now to recollect the facts coming out of all the discussions.)

pylbrecht:
@julienvincent, I added you as co-author to the commits.
However the email from your Github profile does not seem to be the one you used for Google's CLA, which [makes the cla/google job fail on CI](https://github.com/jj-vcs/jj/pull/4853/checks?check_run_id=34826126610).

Would you like to provide the email address you used to sign the CLA, so I can add you as a co-author?

julienvincent:
Thanks for attribution - I just went ahead and signed the CLA for the email address you used.

I reran the check which is now passing 👍🏻

julienvincent:
> Was it planned to add builtin_log_compact_with_sig?

As far as I remember - originally this was implemented by modifying the default template to add commit signatures.

This was then removed from the default template into a new, named one.

This was done for two reasons:

1. To avoid making unexpected changes to the default view.
2. Commit signature verification can be slow especially when viewing a large number of commits.

necauqua:
Yeah, mostly the second point.
We debated about adding a separate `--show-signatures` flag, and maybe it still makes sense to add to have your personal template either show checkmarks or not.

Also, I vaguely recall that I suggested having something like `[●]` for just showing that some signature is present by default (that is very cheap), so that you know that you might want to look at the signatures, but @julienvincent hated that I think 😅

pylbrecht:
Thanks for elaborating on past discussions.

> Was it planned to add `builtin_log_compact_with_sig`?

I was wondering about the dormant [`builtin_sig_status` template](https://github.com/pylbrecht/jj/blob/8a8e8b5ddb4ccdc314f6729f7eb91e2f151d307c/cli/src/config/templates.toml#L335-L353) plus the test cases I found in https://github.com/jj-vcs/jj/pull/3142/files#diff-edf6d251f71a7d4a3b3b8cf73345ea16e6dc7691ed1b98a1811d5709cbe011d2 asserting output with these fancy checkmarks.

While it is not actively used at the moment, should we keep `builtin_sig_status` as a building block?
 
On a general note, I would like to get this merged rather sooner than later, dealing with additional work in follow-ups wherever reasonable. I think @bnjmnt4n would be happy to see this merged, as #4712 depends on parts of this PR.

yuja:
I would suggest splitting builtin template thingy to separate PR (because people have their tastes, and we might have to think about compatibility if the default template contained some emojis.)

Regarding template types and methods, we now have comparison operators `==`, so it might make sense to use string/enum instead of `is_<state>() -> bool`.

pylbrecht:
> I would suggest splitting builtin template thingy to separate PR (because people have their tastes, and we might have to think about compatibility if the default template contained some emojis.)

Good point. I am happy to move that to a separate PR/issue.

Just to be clear, you are suggesting to move `builting_log_detailed_with_sig` to a separate PR, correct?
Right now the test in `cli/tests/test_signature_templates.rs` is using `builtin_log_detailed_with_sig`.
With `builtin_log_detailed_with_sig` being gone, we would need to write a template from scratch and pass it via `-T` in the test. (We could use a very minimal template to solely verify the commit signature part.)

Does that sound like a reasonable approach to you?

> Regarding template types and methods, we now have comparison operators ==, so it might make sense to use string/enum instead of is_<state>() -> bool.

I suppose you are referring to methods on [CommitSignature](https://github.com/pylbrecht/jj/blob/8a8e8b5ddb4ccdc314f6729f7eb91e2f151d307c/cli/src/commit_templater.rs#L1728-L1764) (exposed via [`build_commit_signature_method()`](https://github.com/pylbrecht/jj/blob/8a8e8b5ddb4ccdc314f6729f7eb91e2f151d307c/cli/src/commit_templater.rs#L1766-L1840))?
We could replace
```
.present() -> Boolean
.good() -> Boolean
.bad() -> Boolean
.unknown() -> Boolean
```
with
```
.state() -> Option<String>  # returns "good", "bad", "unknown" if signature is present
```

Having this would have been nice:
```
signature() -> Option<String> # returns "good", "bad", "unknown" if signature is present 
```
However, we still have to take care of
```
.key() -> String  # Signing backend specific key id
.display() -> String  # Signing backend specific display string.
```

We could move these to a separate type:
```
signing_backend() -> Option<SigningBackend>

# SigningBackend
.key()
.display()
```
I might be overthinking it, though.

What do you think?

yuja:
> Just to be clear, you are suggesting to move `builting_log_detailed_with_sig` to a separate PR, correct? Right now the test in `cli/tests/test_signature_templates.rs` is using `builtin_log_detailed_with_sig`. With `builtin_log_detailed_with_sig` being gone, we would need to write a template from scratch and pass it via `-T` in the test. (We could use a very minimal template to solely verify the commit signature part.)

Correct. Just implement the core templating logic in this PR, and do `builtin_*` templates and color stuff in separate PR. The latter tends to lead to bikeshedding discussion.

> ```
> .present() -> Boolean
> .good() -> Boolean
> .bad() -> Boolean
> .unknown() -> Boolean
> ```
> 
> with
> 
> ```
> .state() -> Option<String>  # returns "good", "bad", "unknown" if signature is present
> ```

nit: return value needs to be `String` (both `Option<_>` and `String` can be implicitly casted to boolean in template language, so `Option<String>` would be ambiguous.)

> 
> Having this would have been nice:
> 
> ```
> signature() -> Option<String> # returns "good", "bad", "unknown" if signature is present 
> ```
> 
> However, we still have to take care of
> 
> ```
> .key() -> String  # Signing backend specific key id
> .display() -> String  # Signing backend specific display string.
> ```
> 
> We could move these to a separate type:
> 
> ```
> signing_backend() -> Option<SigningBackend>
> 
> # SigningBackend
> .key()
> .display()
> ```

Maybe we can add intermediate (unverified) signature type?

```
commit.signature() -> Option<CryptographicSignature>
sig.verify() (or .status()?) -> Good|Bad|Unknown(|Invalid?)
sig.display()
sig.key()
```

iirc, signature verification is the expensive part, so it might make sense to test the existence of signature without verifying it. The `CryptographicSignature` type might have to run verification and cache the result internally. @necauqua might have a better idea about API surface.

pylbrecht:
Thanks for your input, @yuja.

I will remove `builtin_log_detailed_with_sig` from this PR, then.

> nit: return value needs to be String (both Option<_> and String can be implicitly casted to boolean in template language, so Option<String> would be ambiguous.)

Thanks, I will take that into account.

> Maybe we can add intermediate (unverified) signature type?
> ```
> commit.signature() -> Option<CryptographicSignature>
> sig.verify() (or .status()?) -> Good|Bad|Unknown(|Invalid?)
> sig.display()
> sig.key()
> ```

I really like that design. Having an isolated `verify()` fits well with the "verifying signatures can be slow" constraint.

> The CryptographicSignature type might have to run verification and cache the result internally.

Not entirely sure what "cache the result" would mean for the code, yet, but I will look into it.

necauqua:
I just realised that since we have `--ignore-immutable` and `--color=` as global parameters, which are basically shortcuts for `--config=appropriate config`, and similar almost could be said for things like `--debug` or `--no-pager`.

I feel like `--show-signatures` could be a shortcut for `--config=templates.builtin_log_signature="builtin_log_signature_show"` where `builtin_log_signature` is empty by default and is present within default log templates and the show variant is normally-unused template that does show the fancy checkmarks.

This whole idea could be extended to "custom flag switches that are aliases for config overrides", or even (just realised) literally "flag switch command aliases"

pylbrecht:
@yuja, I pushed a first draft implementation of your suggested design (https://github.com/jj-vcs/jj/pull/4853#issuecomment-2561663222):
> ```
> commit.signature() -> Option<CryptographicSignature>
> sig.verify() (or .status()?) -> Good|Bad|Unknown(|Invalid?)
> sig.display()
> sig.key()
> ```
What I did not know is that `.key()` and `.display()` both require the slow (but cached) verification. Now [the implementation](https://github.com/pylbrecht/jj/blob/f55244e77265d4f09216821b04896082aa3b6d7e/cli/src/commit_templater.rs#L1739-L1769) feels a little repetitive, but maybe it has to be?

Any feedback/suggestions are more than welcome!

yuja:
> What I did not know is that `.key()` and `.display()` both require the slow (but cached) verification. Now [the implementation](https://github.com/pylbrecht/jj/blob/f55244e77265d4f09216821b04896082aa3b6d7e/cli/src/commit_templater.rs#L1739-L1769) feels a little repetitive, but maybe it has to be?

Yeah, there might be a better way to organize the verification API, but I have no idea. @necauqua ?

I think that's good as a start, btw.

cli/src/commit_templater.rs
```diff
@@ -319,6 +336,13 @@ impl<'repo> CommitTemplateLanguage<'repo> {
     ) -> CommitTemplatePropertyKind<'repo> {
         CommitTemplatePropertyKind::TreeDiff(Box::new(property))
     }
+
+    fn wrap_cryptographic_signature(
```
yuja, line 340: nit: `_opt`

cli/src/commit_templater.rs
```diff
@@ -1671,3 +1730,83 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
     // TODO: add files() or map() to support custom summary-like formatting?
     map
 }
+
+#[derive(Clone, Debug, Eq, PartialEq)]
```
yuja, line 1734: nit: better to not implement `Eq, PartialEq` because it doesn't compare the signature.

cli/src/commit_templater.rs
```diff
@@ -1671,3 +1730,83 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
...
+pub struct CryptographicSignature {
+    commit: Commit,
+}
+
+impl CryptographicSignature {
+    /// Panics if the commit is not signed.
+    fn new(commit: Commit) -> Self {
+        if !commit.is_signed() {
+            panic!("commit must be signed");
```
yuja, line 1743: nit: `assert!(..)`?

cli/src/commit_templater.rs
```diff
@@ -1671,3 +1730,83 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
...
+    let mut map = CommitTemplateBuildMethodFnMap::<CryptographicSignature>::new();
+    map.insert(
+        "verify",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            function.expect_no_arguments()?;
+            let out_property = TemplateFunction::new(self_property, |sig| match sig.verify() {
+                Ok(verification) => match verification.status {
+                    SigStatus::Good => Ok("Good".into()),
+                    SigStatus::Bad => Ok("Bad".into()),
+                    SigStatus::Unknown => Ok("Unknown".into()),
```
yuja, line 1785: nit: lowercase "good"/"bad"/"unknown" might be better here. We tend to use kebab-case in config-like context.

cli/src/commit_templater.rs
```diff
@@ -1671,3 +1730,83 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
...
+            });
+            Ok(L::wrap_string(out_property))
+        },
+    );
+    map.insert(
+        "key",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            function.expect_no_arguments()?;
+            let out_property =
+                TemplateFunction::new(self_property, |sig| Ok(sig.key().unwrap_or("".into())));
```
yuja, line 1798: nit: `.unwrap_or_default()` or `.unwrap_or_else(String::new)`?

necauqua:
> @necauqua ?

With gpg you have to verify to get any info, not sure about ssh, but probably same

The only thing you (kind of obviously) get for free is checking if there's a signature at all

cli/src/commit_templater.rs
pylbrecht: That went away with the implementation.

pylbrecht:
Thanks for your feedback, @yuja and @necauqua.

I implemented it, as suggested, with the exception of using `commit.signed()` instead of `commit.is_signed()`. There were no other `is_...()` template functions, so I felt `.signed()` would be more consistent with what we already have (e.g. `divergent()`, `immutable()`, etc.).

If you disagree, I am happy to change it back to `.is_signed()`. No strong feelings whatsoever!

yuja:
> commit.signed() instead of commit.is_signed()

Isn't `if(commit.signed(), ..)` equivalent to `if(commit.signature(), ..)`? About the naming, I agree.

I'll take a look maybe tomorrow, thanks.

necauqua:
> equivalent

The point is that if there *is* a signature then the `.signature()` has to do full verification to get the signature object, I did see that we're duplicating the meaning a bit with this separate boolean.

Technically we could probably make the Option => Boolean coersion avoid computing the content of the option somehow, that would be a cool optimisation.

OTOH having a separate boolean thing that's guaranteed to be cheap without magic may be nicer.
But I love the lazy option coersion idea.

pylbrecht:
> The point is that if there is a signature then the .signature() has to do full verification to get the signature object, I did see that we're duplicating the meaning a bit with this separate boolean.

Exactly this.

> Technically we could probably make the Option => Boolean coersion avoid computing the content of the option somehow, that would be a cool optimisation.

That's what I have been thinking about for the past hour. But now my head hurts and I still have no idea how to implement that.
It would certainly be the way I would like it to be in terms of ergonomics.

yuja:
> > equivalent
> 
> The point is that if there _is_ a signature then the `.signature()` has to do full verification to get the signature object, I did see that we're duplicating the meaning a bit with this separate boolean.

Oh, I thought `.signature()` would return a lazy wrapper `CryptographicSignature(Commit)`, and `.signature().<whatever>()` would invoke the underlying `verify()` method.

Maybe I misread https://github.com/jj-vcs/jj/pull/4853#issuecomment-2564956648 that we were discussing about the library API.

pylbrecht:
> Oh, I thought `.signature()` would return a lazy wrapper `CryptographicSignature(Commit)`, and `.signature().<whatever>()` would invoke the underlying `verify()` method.

That's how I implemented it at first, but that resulted in the repitition I mentioned in https://github.com/jj-vcs/jj/pull/4853#issuecomment-2564744675 (i.e. `signature().verify()`, `signature().key()`, and `signature().display()` all having to call [`Commit::verification()`](https://github.com/pylbrecht/jj/blob/8a379f048191156541ce5c2b110f77129f4b2318/lib/src/commit.rs#L172)).

If it's fine that every `signature().<whatever>()` method calls [`Commit::verification()`](https://github.com/pylbrecht/jj/blob/8a379f048191156541ce5c2b110f77129f4b2318/lib/src/commit.rs#L172), I am happy to go back to the lazy wrapper `CryptographicSignature(Commit)`.

yuja:
> > Oh, I thought `.signature()` would return a lazy wrapper `CryptographicSignature(Commit)`, and `.signature().<whatever>()` would invoke the underlying `verify()` method.
> 
> That's how I implemented it at first, but that resulted in the repitition I mentioned in [#4853 (comment)](https://github.com/jj-vcs/jj/pull/4853#issuecomment-2564744675) (i.e. `signature().verify()`, `signature().key()`, and `signature().display()` all having to call [`Commit::verification()`](https://github.com/pylbrecht/jj/blob/8a379f048191156541ce5c2b110f77129f4b2318/lib/src/commit.rs#L172)).

Yes, and I think that's okay as a start. (Regarding the method name, `.status()` might be better because they all do the verification under the hood.)

We could cache the result in that object (by `OnceCell`), but it wouldn't help because templater doesn't provide syntax to call multiple methods on the same object. Since the verification result is cached anyway by the backend, we wouldn't have to pay extra effort in templater part.

pylbrecht:
> > > Oh, I thought `.signature()` would return a lazy wrapper `CryptographicSignature(Commit)`, and `.signature().<whatever>()` would invoke the underlying `verify()` method.
> > 
> > 
> > That's how I implemented it at first, but that resulted in the repitition I mentioned in [#4853 (comment)](https://github.com/jj-vcs/jj/pull/4853#issuecomment-2564744675) (i.e. `signature().verify()`, `signature().key()`, and `signature().display()` all having to call [`Commit::verification()`](https://github.com/pylbrecht/jj/blob/8a379f048191156541ce5c2b110f77129f4b2318/lib/src/commit.rs#L172)).
> 
> Yes, and I think that's okay as a start. (Regarding the method name, `.status()` might be better because they all do the verification under the hood.)

Great, then I will "rollback" to that implementation (probably later today).

> We could cache the result in that object (by `OnceCell`), but it wouldn't help because templater doesn't provide syntax to call multiple methods on the same object. Since the verification result is cached anyway by the backend, we wouldn't have to pay extra effort in templater part.

Thanks for elaborating on this!
I also thought about caching the verification within `CryptographicSignature`, but wasn't sure whether the templater wouldn't just create a new instance every time we call `commit.signature()`.

necauqua:
I didn't look at the current code much, so maybe I was adding a bit to the confusion because of that, sorry :upside_down_face: 

Yeah I wanted to mention (but never did) that I do recall the backend doing caching of it's own, so it's ok to call verify repeatedly and does, in fact, give purpose to that caching lol

cli/src/commit_templater.rs
```diff
@@ -1671,3 +1732,81 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
...
+    map.insert(
+        "status",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            function.expect_no_arguments()?;
+            let out_property = TemplateFunction::new(self_property, |sig| {
+                let status = match sig.status() {
+                    Ok(SigStatus::Good) => "good",
+                    Ok(SigStatus::Bad) => "bad",
+                    Ok(SigStatus::Unknown) => "unknown",
+                    Err(_) => "invalid",
```
pylbrecht, line 1786: I am not sure if we want to steamroller every error into "invalid", or if we want something more fine-grained.
[`SignError` defines two variants](https://github.com/pylbrecht/jj/blob/e3aebadb8a86d5d9a6e44eb1d359f2dafe588390/lib/src/signing.rs#L124-L131):
```rust
/// An error type for the signing/verifying operations
#[derive(Debug, Error)]
pub enum SignError {
    /// The verification failed because the signature *format* was invalid.
    #[error("Invalid signature")]
    InvalidSignatureFormat,
    /// A generic error from the backend impl.
    #[error("Signing error")]
    Backend(#[source] Box<dyn std::error::Error + Send + Sync>),
}
```

Maybe we only want to convert `InvalidSignatureFormat` to "invalid" and bubble up the other?

necauqua: I'm pretty sure that was exactly the point of separating out the explicitly invalid variant

yuja: If it rarely occurs, I would just propagate all errors. The way it was implemented would be just because we didn't have a mechanism to propagate error from template functions.

lib/src/signing.rs
```diff
@@ -168,6 +170,9 @@ impl Signer {
             // Box::new(X509Backend::from_settings(settings).map_err(..)?),
         ];
 
+        #[cfg(feature = "testing")]
+        backends.push(Box::new(TestSigningBackend) as Box<dyn SigningBackend>);
```
yuja, line 191: nit: type annotation can be omitted

and maybe it can be inlined in `vec![..]`

lib/src/test_signing_backend.rs
```diff
@@ -63,13 +63,13 @@ impl SigningBackend for TestSigningBackend {
             Ok(Verification {
                 status: SigStatus::Good,
                 key,
-                display: None,
+                display: Some(self.name().into()),
```
yuja: nit: maybe better to hardcode "test" here because `display` is NOT the backend name. (applies to the other changes)

CHANGELOG.md
yuja, line 1: Re: commit message

iirc, `libgit2` appends trailing newline. See bfb7613d5d192d3c4dc533afa4f2ff0d6b9016c5 for the workaround.

cli/src/commit_templater.rs
```diff
@@ -319,6 +336,13 @@ impl<'repo> CommitTemplateLanguage<'repo> {
     ) -> CommitTemplatePropertyKind<'repo> {
         CommitTemplatePropertyKind::TreeDiff(Box::new(property))
     }
+
+    fn wrap_cryptographic_signature_opt(
+        &self,
```
yuja, line 341: remove `&self`

cli/src/commit_templater.rs
```diff
@@ -621,6 +666,22 @@ fn builtin_commit_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, Comm
...
+                language.wrap_cryptographic_signature_opt(TemplateFunction::new(
+                    self_property,
+                    |commit| {
+                        Ok(commit
+                            .is_signed()
+                            .then(|| CryptographicSignature::new(commit)))
+                    },
+                )),
+            )
+        },
```
yuja, line 683: `self_property.map()` can be used here
```rust
            let out_property = self_property.map(|commit| {
                commit
                    .is_signed()
                    .then_some(CryptographicSignature::new(commit))
            });
            Ok(L::wrap_cryptographic_signature_opt(out_property))
```

cli/src/commit_templater.rs
```diff
@@ -1671,3 +1732,81 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
...
+    }
+
+    fn key(&self) -> Option<String> {
+        self.verify().map_or(None, |verification| verification.key)
+    }
+
+    fn display(&self) -> Option<String> {
+        self.verify()
+            .map_or(None, |verification| verification.display)
+    }
```
yuja, line 1768: nit: maybe better to propagate error?

cli/src/commit_templater.rs
```diff
@@ -1671,3 +1732,81 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
...
+) -> CommitTemplateBuildMethodFnMap<'repo, CryptographicSignature> {
+    type L<'repo> = CommitTemplateLanguage<'repo>;
+    // Not using maplit::hashmap!{} or custom declarative macro here because
+    // code completion inside macro is quite restricted.
+    let mut map = CommitTemplateBuildMethodFnMap::<CryptographicSignature>::new();
+    map.insert(
+        "status",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            function.expect_no_arguments()?;
+            let out_property = TemplateFunction::new(self_property, |sig| {
```
yuja, line 1781: nit: maybe `self_property.map()` or `.and_then()` can be used here (applies to the other changes)

docs/templates.md
```diff
@@ -93,6 +93,7 @@ This type cannot be printed. The following methods are defined.
 * `parents() -> List<Commit>`
 * `author() -> Signature`
 * `committer() -> Signature`
+* <a name="commit.signature"></a>`signature() -> Option<CryptographicSignature>`
```
yuja, line 96: nit: not sure if it's common to add anchor in markdown. Maybe we can link to the `CryptographicSignature type` section instead?

lib/src/signing.rs
```diff
@@ -53,6 +53,11 @@ pub struct Verification {
     /// A display string, if available. For GPG, this will be formatted primary
     /// user ID.
     pub display: Option<String>,
+    /// The name of the backend that provided this verification.
+    /// Is `None` when no backend was found that could read the signature.
+    ///
+    /// Always set by the signer.
+    backend: Option<String>,
```
yuja, line 60: If this is supposed to be set by `Signer`, I don't think `Verification::new()` should have `backend` argument.

I personally don't mind if it's set by the signing backend, and the field is `pub backend: String`.

yuja (commented):
Thanks!

docs/templates.md
pylbrecht: I don't understand.
We are linking from the [Commit Signature Verification section in config.md](https://github.com/pylbrecht/jj/blob/e3aebadb8a86d5d9a6e44eb1d359f2dafe588390/docs/config.md#L1030) to here. That paragraph in `config.md` also links to the `CryptographicSignature` type.
That way you effortlessly get from the [Commit Signature Verification section](https://github.com/pylbrecht/jj/blob/e3aebadb8a86d5d9a6e44eb1d359f2dafe588390/docs/config.md#L1030) to the template functions.

Or are you suggesting to remove the `commit.signature` anchor entirely and only link to the `CryptographicSignature` type?

CHANGELOG.md
pylbrecht: Thanks, that makes much more sense now. I will edit the commit message accordingly.

cli/src/commit_templater.rs
pylbrecht: Addressed.
However we can't use `then_some()`, because it evaluates eagerly and we [panic in `CryptographicSignature::new()`](https://github.com/pylbrecht/jj/blob/1d3c3b8ab207e66c978ed026a7393682989111c5/cli/src/commit_templater.rs#L1740-L1742) for unsigned commits. I left a comment in the code.

Let me know if that works for you.

cli/src/commit_templater.rs
pylbrecht: > If it rarely occurs, I would just propagate all errors. The way it was implemented would be just because we didn't have a mechanism to propagate error from template functions.

That would mean removing `"invalid"` entirely, correct?

lib/src/signing.rs
pylbrecht: Ah yeah, I recall that I was a bit unsure about this part.

If the `Signer` should set `backend`, wouldn't we need some sort of setter method then?
Or how would we set it from "outside" `Verification`?

@necauqua, @julienvincent, maybe you remember your intentions? (I think the original changes came from you?)

docs/templates.md
joyously: If you leave the anchor, you should not be using a deprecated attribute (`name`). The `id` attribute is the standard, and the id should be unique on a page.

lib/src/signing.rs
necauqua: Afair no specific reason, Yujas suggestion is totally ok

docs/templates.md
yuja: > Or are you suggesting to remove the commit.signature anchor entirely and only link to the CryptographicSignature type?

Yes, but I'm not entirely sure. I'm just questioning because we don't have this pattern in our markdown docs. (fwiw, if we use `id`, maybe `<span id=..>signature()..</span>` is semantically correct.)

cli/src/commit_templater.rs
yuja: Ah, true. Another option is make `CryptographicSignature::new` return `Option<Self>`.

cli/src/commit_templater.rs
yuja: > That would mean removing "invalid" entirely, correct?

Correct. A template property error will be rendered inline as `<Error: ..>`. If `"invalid"` is non-error, it might be too verbose. So the decision would depend on whether the error is unusual or not.

cli/src/commit_templater.rs
pylbrecht: Oh, that's neat! It never occurred to me to return an `Option` from `new()` (seems like my head is still stuck in Python's way of constructors).

lib/tests/test_ssh_signing.rs
```diff
@@ -37,14 +37,14 @@ y2yxhhHnagH52avUqw5hAAAAAAECAwQF
...
     _keys: tempfile::TempDir,
-    private_key_path: PathBuf,
-    allowed_signers: Option<tempfile::TempPath>,
+    pub private_key_path: PathBuf,
+    pub allowed_signers: Option<tempfile::TempPath>,
 }
 
 impl SshEnvironment {
-    fn new() -> Result<Self, std::process::Output> {
+    pub fn new() -> Result<Self, std::process::Output> {
```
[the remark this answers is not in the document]
pylbrecht: Could be! 
I started out with `jj duplicate`ing commits from #3141 and might have lost track of what should go where during editing.
I will remove that change and see what happens.

docs/templates.md
pylbrecht: Fair enough.
I will leave out the anchor for now. We can think about introducing anchors in general and then update the existing docs, I suppose.

lib/src/signing.rs
pylbrecht: The `Signer` does set `backend` [here](https://github.com/pylbrecht/jj/blob/d76a54d32f499ba995be16410e243e3ff0a936e5/lib/src/signing.rs#L257). However, removing `backend` from `Verification::new()` makes some tests in `lib/tests/test_signing.rs` fail:
```
running 6 tests
test test_signing::manual::local_backend ... FAILED
test test_signing::forced::git_backend ... FAILED
test test_signing::configured::git_backend ... FAILED
test test_signing::manual::git_backend ... FAILED
test test_signing::keep_on_rewrite::git_backend ... FAILED
test test_signing::manual_drop_on_rewrite::git_backend ... ok

failures:

---- test_signing::manual::local_backend stdout ----
thread 'test_signing::manual::local_backend' panicked at lib/tests/test_signing.rs:80:5:
assertion `left == right` failed
  left: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: Some("test") })
 right: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: None })

---- test_signing::forced::git_backend stdout ----
thread 'test_signing::forced::git_backend' panicked at lib/tests/test_signing.rs:162:5:
assertion `left == right` failed
  left: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: Some("test") })
 right: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: None })

---- test_signing::configured::git_backend stdout ----
thread 'test_signing::configured::git_backend' panicked at lib/tests/test_signing.rs:181:5:
assertion `left == right` failed
  left: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: Some("test") })
 right: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: None })

---- test_signing::manual::git_backend stdout ----
thread 'test_signing::manual::git_backend' panicked at lib/tests/test_signing.rs:80:5:
assertion `left == right` failed
  left: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: Some("test") })
 right: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: None })

---- test_signing::keep_on_rewrite::git_backend stdout ----
thread 'test_signing::keep_on_rewrite::git_backend' panicked at lib/tests/test_signing.rs:109:5:
assertion `left == right` failed
  left: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: Some("test") })
 right: Some(Verification { status: Good, key: Some("impeccable"), display: Some("test"), backend: None })


failures:
    test_signing::configured::git_backend
    test_signing::forced::git_backend
    test_signing::keep_on_rewrite::git_backend
    test_signing::manual::git_backend
    test_signing::manual::local_backend

test result: FAILED. 1 passed; 5 failed; 0 ignored; 0 measured; 396 filtered out; finished in 0.20s

error: test failed, to rerun pass `-p jj-lib --test runner`
```

We can no longer construct a `Verification` instance to assert against (see [`good_signature()`](https://github.com/pylbrecht/jj/blob/d76a54d32f499ba995be16410e243e3ff0a936e5/lib/tests/test_signing.rs#L47-L54)).

Since nobody raised any major concerns with it, I will just go for @yuja's suggestion of making `Verification::backend` public since that is easy to do.

We could convert the failing `test_signing` tests to snapshot tests, which would let us assert without having to construct a `Verification` instance ([like we do in `test_gpg`](https://github.com/pylbrecht/jj/blob/d76a54d32f499ba995be16410e243e3ff0a936e5/lib/tests/test_gpg.rs#L137-L150)). I am happy to go that extra mile, if you confirm that it's worth it! 😁

docs/templates.md
```diff
@@ -129,6 +130,20 @@ The following methods are defined.
...
+
+* `.status() -> String`: The signature's status ("good", "bad", "unknown", "invalid").
+* `.key() -> String`: The signature's key id representation (for GPG, this is the key fingerprint).
+* `.display() -> String`: The signature's display string (for GPG this is the formatted primary user ID.
+
+!!! warning
+
+    Calling any of `.status()`, `.key()`, or `.display()` is slow, as it incurs
+    the performance cost of verifying the signature. Though consecutive calls
+    will be faster, because the backend caches the verification result.
```
necauqua, line 145: (actually looking through the code top-to-bottom finally)

imo you should mention that just checking for signature presence via the boolean coersion (`if(commit.signature(), "commit has a signature", "commit is unsigned")`) is very fast and cheap, as opposed to the verification part

Also, `[..] verifying the signature. Though [..]` -> `[..] verifying the signature (for example, shelling out to gpg or ssh-keygen). Though [..]`, just giving an example of why it is slow.

fyi creating a single commit with a gpg signature made through yubikey takes ~200ms for me, a bit annoying :)

fyi² it is (by design) possible to implement another gpg backend (for example) that'll be faster by using some rust gpg implementation or something instead of the lazy way of just calling the gpg binary.

lib/src/signing.rs
necauqua: eh, you could fix good_signature with the field being public, or you could convert the tests if you want, not a huge difference in terms of worth idk

CHANGELOG.md
```diff
@@ -82,6 +82,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   below it, similar to a "scissor line" in git. When editing multiple commits,
   only ignore until the next `JJ: describe` line.
 
+* Add templater support for rendering commit signatures.
+
```
yuja, line 86: reminder: you'll need to relocate the changelog entry after the release

lib/src/test_signing_backend.rs
yuja: nit: squash `Some("test".into())` from the next commit

docs/templates.md
pylbrecht: Let me know if I am overdoing it with [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/). I *love* admonitions. 😅

pylbrecht:
@yuja, @necauqua, everything should be addressed now, so this is ready for a (hopefully) final round of review.
Thank you already for patiently playing comment ping-pong with me 😉

cli/src/commit_templater.rs
```diff
@@ -1671,3 +1723,76 @@ fn builtin_tree_diff_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, T
...
+    fn status(&self) -> SignResult<SigStatus> {
+        self.verify().map(|verification| verification.status)
+    }
+
+    /// Defaults to empty string if key is not present.
+    fn key(&self) -> Result<String, TemplatePropertyError> {
+        self.verify()
+            .map_err(|e| TemplatePropertyError(e.into()))
+            .map(|verification| verification.key.unwrap_or_default())
+    }
```
yuja, line 1753: nit: for consistency, `key()` and `display()` can return `SignResult<String>`. The error can be mapped by caller as `Ok(sig.key()?)`.

cli/tests/test_signature_templates.rs
```diff
@@ -0,0 +1,39 @@
...
+                      ) ++ " signature""#;
+
+    let stdout = test_env.jj_cmd_success(&repo_path, &["log", "-T", template]);
+    insta::assert_snapshot!(stdout, @r"
+        @  good test signature
+        ◆  no signature");
+
+    let stdout = test_env.jj_cmd_success(&repo_path, &["show", "-T", template]);
+    insta::assert_snapshot!(stdout, @"good test signature");
+}
```
yuja, line 39: nit: If we aren't going to more tests, I would move it to test_commit_template.rs.

cli/tests/test_signature_templates.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright 2022 The Jujutsu Authors
```
yuja, line 1: nit: 2025?

lib/src/signing.rs
```diff
@@ -39,6 +41,16 @@ pub enum SigStatus {
...
 
+impl From<SigStatus> for String {
+    fn from(status: SigStatus) -> Self {
+        match status {
+            SigStatus::Good => "good".to_string(),
+            SigStatus::Unknown => "unknown".to_string(),
+            SigStatus::Bad => "bad".to_string(),
+        }
+    }
+}
```
yuja, line 52: nit: A common practice for default string representation is to implement `std::fmt::Display`. (see `ColorChoice` for example.) It's also good to inline it in templater.

lib/src/signing.rs
```diff
@@ -71,8 +87,14 @@ impl Verification {
...
             key,
             display,
+            backend: None,
         }
     }
+
+    /// The name of the backend that provided this verification.
+    pub fn backend(&self) -> Option<&str> {
+        self.backend.as_deref()
+    }
```
yuja, line 97: nit: this method is no longer needed since `backend` is public.

Oh, btw, the `backend` field is unused right now. Do you remember where it will be used?

lib/src/signing.rs
```diff
@@ -233,7 +257,10 @@ impl Signer {
             .find_map(|backend| match backend.verify(data, signature) {
                 Ok(check) if check.status == SigStatus::Unknown => None,
                 Err(SignError::InvalidSignatureFormat) => None,
-                e => Some(e),
+                e => Some(e.map(|mut v| {
+                    v.backend = Some(backend.name().to_owned());
+                    v
+                })),
```
yuja, line 263: nit: Since the backend filed is public, I thought it would now be set by the backend implementation, not by Signer.

lib/src/test_signing_backend.rs
```diff
@@ -42,17 +60,9 @@ impl SigningBackend for TestSigningBackend {
 
         let sig = self.sign(data, key.as_deref())?;
         if sig == signature {
-            Ok(Verification {
-                status: SigStatus::Good,
-                key,
-                display: None,
-            })
+            Ok(Verification::new(SigStatus::Good, key, Some("test".into())))
```
yuja, line 63: nit: since `backend` field is public, we don't have to use `::new()`. It's nice if field names are explicitly listed.

yuja (approved):
Thanks!

lib/src/signing.rs
necauqua: oh, actually

I think it should be in the templater?.
Like `commit.signature().backend()`
That's the only thing I recall about the backend field, maybe there was something else..
Anyway it's definitely not a core part/dealbreaker, so if it's unused and we dont need it in the templater we can get rid of it then

lib/src/signing.rs
pylbrecht: Addressed, but inlining it in `commit_templater.rs` gives me:
```
error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate
    --> cli/src/commit_templater.rs:1764:1
     |
1764 | impl Display for SigStatus {
     | ^^^^^^^^^^^^^^^^^---------
     | |                |
     | |                `jj_lib::signing::SigStatus` is not defined in the current crate
     | impl doesn't use only types from inside the current crate
     |
     = note: define and implement a trait or new type instead
```

cli/tests/test_signature_templates.rs
pylbrecht: Oh wow.. copy/paste is not my friend 😬

lib/src/signing.rs
pylbrecht: Neat, I wasn't aware that implementing `Display` gives you `.to_string()` 💡 
Thanks!

lib/src/signing.rs
pylbrecht: Oh true, in #3141 we had `signature.backend()` in the templater!
I must have lost that.. I will bring it back. Thanks for spotting that!

lib/src/test_signing_backend.rs
pylbrecht: I suppose that comment applies to all occurrences of `Verification::new()` then?
(We have a bunch more in `lib/src/gpg_signing.rs` and `lib/src/ssh_signing.rs`, for example.)

lib/src/signing.rs
pylbrecht: Ah yeah, of course.
I will look into that (probably later today).

lib/src/signing.rs
yuja: Oh, I just meant `Display` OR inlining (into "status" method), not both. I don't have strong feeling about these.

lib/src/test_signing_backend.rs
yuja: Yes, but I wouldn't rewrite the existing callers within this commit. They're unrelated change.

pylbrecht: I will take some time and play with both approaches (`backend` private vs. public). I have this itchy feeling that leaving it private and adjusting the failing tests might be less work.

yuja: or just remove `backend` for now? I don't think it's super important to know whether the signature was verified by `gpg` or `ssh-keygen`.

pylbrecht: It turned out to be very straightforward to update the tests with `backend` being private.
[insta's inline snapshots](https://insta.rs/docs/snapshot-types/#inline-snapshots) are *awesome*!

lib/tests/test_gpg.rs
```diff
@@ -143,8 +143,9 @@ fn gpg_signing_roundtrip_explicit_key() {
         display: Some(
             "Someone Else (jj test signing key) <someone-else@example.com>",
         ),
+        backend: None,
```
pylbrecht, line 146: I suppose it's alright that `backend` is `None` here, since the `Signer` is setting it and not the backend?

lib/src/test_signing_backend.rs
pylbrecht: We do introduce a little duplication, because we inline `good_signature()`, but I think that's just unavoidable with inline snapshots. We might be able to use [unnamed snapshots](https://insta.rs/docs/snapshot-types/#file-snapshots) as an alternative, but I liked that one can see the expected value when looking at the code.

pylbrecht:
All right, not  daring to mention the words "final" or "last", but this is ready for review again.
Also, feel free to just react with a 👍 to this comment. Then I will merge it as is.

lib/tests/test_gpg.rs
pylbrecht: I did not question if there is a reason that (or if at all) the backend should always be set by the `Signer`.
@necauqua, do you know?

necauqua: Nope
As I said, the only recollection I have of the backend field at all is that it could be used in the templater?.
Like the `builtin_log_detailed` would have another line looking like `Signature: A good **GPG** signature by Test User <user@example.com>` etc.

So if it is not obvious why the backend field is even needed it can be ditched (I think Yuja suggested that, we can think if it's really needed later), the whole "setting the field from the outside" seems dodgy anyway

The answer to the direct question "why the backend shouldn't set its own name" I think was something along the lines of "reducing things backend impl should care about" - so nothing special.

yuja: So if we prefer strictness, the `Verification` type would be `Unknown|Verified(BackendVerification, BackendName)`. I don't have a strong feeling about that, so I suggested either making `backend` public or removing it for now.

> I think Yuja suggested that, we can think if it's really needed later

Yes.

pylbrecht: All right, thanks again for clarifying. I didn't really get it the first time.
I will remove the `backend` field then and we can see how we deal want to deal with it later.

pylbrecht: > So if we prefer strictness, the Verification type would be Unknown|Verified(BackendVerification, BackendName).

That pseudo code made it click for me.

Now, with `backend` removed, the diff is also much more focused and concise. 👍

docs/templates.md
```diff
@@ -129,6 +130,31 @@ The following methods are defined.
...
 * `.shortest([min_len: Integer]) -> ShortestIdPrefix`: Shortest unique prefix.
 
+### CryptographicSignature type
+
+The following methods are defined.
+
+* `.status() -> String`: The signature's status (`"good"`, `"bad"`, `"unknown"`, `"invalid"`).
+* `.key() -> String`: The signature's key id representation (for GPG, this is the key fingerprint).
+* `.display() -> String`: The signature's display string (for GPG this is the formatted primary user ID).
+* `.backend() -> String`: The signature backend (e.g. `"gpg"`).
```
yuja, line 140: nit: remove `.backend()` (from here and the warnings)

yuja (approved):

pylbrecht:
@yuja, thanks again!
I seem to no longer have write access to the repo; not sure it that came with the recent maintainer announcement. In any case, you have the honors of merging.

yuja:
> I seem to no longer have write access to the repo; not sure it that came with the recent maintainer announcement.

Oh, the invitation email might be missing or filtered as spam. Mine was received on "Dec 20, 2024", so it would be expired anyway.

cc @martinvonz

martinvonz:
I've sent a new invitation now, @pylbrecht

PR #4867, opened 2024-11-15
cli: Make fix tools run from the repo root.

### LINKED ISSUE #4866: `jj fix` uses incorrect paths when executed from a non-root directory.

Add a tool to your config
```
[fix.tools.gofmt]
command = ["gofmt", "$path"]
patterns = ["glob:'**/*.go'", "glob:'**/go.mod'"]
```

```
$ cd <workspace root>/subdirectory
subdirectory$ jj fix
```

## Expected Behavior

It should be able to fix it

## Actual Behavior

```
stat subdirectory/foo.go: no such file or directory
```

## Specifications

- Platform:
- Version:

matts1:
Fixes #4866

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed bugs
 
 * `jj config unset <TABLE-NAME>` no longer removes a table (such as `[ui]`.)
+* Formatters called by `jj fix` now always run from the repo root
+  ([#4616](https://github.com/martinvonz/jj/issues/4616))
 
 ## [0.23.0] - 2024-11-06
 
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -15,6 +15,7 @@
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::io::Write;
+use std::path::Path;
 use std::process::Stdio;
 use std::sync::mpsc::channel;
 
@@ -158,6 +159,11 @@ pub(crate) fn cmd_fix(
         .parse_file_patterns(ui, &args.paths)?
         .to_matcher();
 
+    // See #4866.
+    // Fix commands must run from the repo root, as it may read files such as
+    // .clang-format that depend on the working directory being correct.
+    let working_dir = workspace_command.repo_path().to_path_buf();
+
     let mut tx = workspace_command.start_transaction();
 
     // Collect all of the unique `ToolInput`s we're going to use. Tools should be
@@ -239,6 +245,7 @@ pub(crate) fn cmd_fix(
         tx.repo().store().as_ref(),
         &tools_config,
         &unique_tool_inputs,
+        &working_dir,
     )?;
 
     // Substitute the fixed file IDs into all of the affected commits. Currently,
@@ -326,6 +333,7 @@ fn fix_file_ids<'a>(
     store: &Store,
     tools_config: &ToolsConfig,
     tool_inputs: &'a HashSet<ToolInput>,
+    working_dir: &Path,
 ) -> Result<HashMap<&'a ToolInput, FileId>, CommandError> {
     let (updates_tx, updates_rx) = channel();
     // TODO: Switch to futures, or document the decision not to. We don't need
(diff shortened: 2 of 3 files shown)


yuja:
Doesn't it break formatter configuration? Ideally config files should be loaded from the project root.

matts1:
Would I be correct in saying that the following failure, which occurs only on windows, is a bug in jj itself?

```
---- test_fix_command::test_executed_but_nothing_changed stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: executed_but_nothing_changed-2
Source: D:\a\jj\jj:816
───────────────────────────────────────────────────────────────────────────────
Expression: redact(&stderr.replace(logfile_path_str, "$logfile"))
───────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬──────────────────────────────────────────────────────────────────
    0     0 │ Warning: The `fix.tool-command` config option is deprecated and will be removed in a future version.
    1     1 │ Hint: Replace it with the following:
    2     2 │             [fix.tools.legacy-tool-command]
    3       │-            command = [<redacted formatter path>, "--tee", "$logfile"]
          3 │+            command = [<redacted formatter path>, "--tee", '$logfile']
    4     4 │             patterns = ["all()"]
    5     5 │             
    6     6 │ Fixed 0 commits of 1 checked.
    7     7 │ Nothing changed.
────────────┴──────────────────────────────────────────────────────────────────
```

It would seem that on windows, the output of the deprecation command is `command = ["formatter", "--tee", '$logfile']`. I don't have a a windows computer to test on, so I can't really debug, but that seems incorrect. This seems wierd because it was previously working.

martinvonz:
> Would I be correct in saying that the following failure, which occurs only on windows, is a bug in jj itself?

The message comes from here:
https://github.com/martinvonz/jj/blob/c5b93a0c366f438186bd6cb16fafd0f5f10b6438/cli/src/commands/fix.rs#L465-L473

The `command = ...` line in particular comes from here:
https://github.com/martinvonz/jj/blob/c5b93a0c366f438186bd6cb16fafd0f5f10b6438/cli/src/commands/fix.rs#L472

`to_toml_value()` is defined here:
https://github.com/martinvonz/jj/blob/c5b93a0c366f438186bd6cb16fafd0f5f10b6438/cli/src/config.rs#L52-L76

That doesn't seem to have any platform-specific code and just returns a `toml_edit::Value`. That type presumably doesn't contain the quotation marks, so maybe the problem is in how we convert it to a string in the second snippet above. Hmm, I can't find anything platform-dependent there either.

Oh, somehow I had not looked at the `redact()` function until now. It has this comment:
https://github.com/martinvonz/jj/blob/068fa0f37e87038b4345d689b0555ce9b2255c03/cli/tests/test_fix_command.rs#L47-L54

That took me quite a while to find. So maybe the `redact()` closure should be updated to also replace the workspace root by a placeholder?

martinvonz (commented):
LGTM, but we work for the same company, so I'll leave actual approval to someone else.

CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed bugs
 
 * `jj config unset <TABLE-NAME>` no longer removes a table (such as `[ui]`.)
+* Formatters called by `jj fix` now always run from the repo root
```
yuja, line 52: nit: insert blank line before

cli/src/commands/fix.rs
```diff
@@ -158,6 +159,11 @@ pub(crate) fn cmd_fix(
         .parse_file_patterns(ui, &args.paths)?
         .to_matcher();
 
+    // See #4866.
+    // Fix commands must run from the repo root, as it may read files such as
+    // .clang-format that depend on the working directory being correct.
+    let working_dir = workspace_command.repo_path().to_path_buf();
```
yuja, line 165: Shouldn't it be `workspace_root()`? Maybe we can add a fake-formatter option that stat/read a file to test the behavior.

fwiw, this can be inlined as `tx.base_workspace_helper()...`.

bryceberger:
#5755

PR #4869, opened 2024-11-15
completion: teach forget about workspaces

senekor:
# Checklist

If applicable:

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -271,6 +271,27 @@ pub fn operations() -> Vec<CompletionCandidate> {
     })
 }
 
+pub fn workspaces() -> Vec<CompletionCandidate> {
+    with_jj(|mut jj, _| {
+        let output = jj
+            .arg("--config-toml")
+            .arg(r#"templates.commit_summary = 'if(description, description.first_line(), "(no description set)")'"#)
+            .arg("workspace")
+            .arg("list")
+            .output()
+            .map_err(user_error)?;
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        Ok(stdout
+            .lines()
+            .map(|line| {
+                let (name, desc) = line.split_once(": ").unwrap_or((line, ""));
+                CompletionCandidate::new(name).help(Some(desc.to_string().into()))
+            })
+            .collect())
+    })
+}
+
 /// Shell out to jj during dynamic completion generation
 ///
 /// In case of errors, print them and early return an empty vector.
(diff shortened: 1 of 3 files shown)


cli/src/complete.rs
```diff
@@ -159,6 +159,37 @@ pub fn aliases() -> Vec<CompletionCandidate> {
...
+            .arg("workspace")
+            .arg("list")
+            .output()
+            .map_err(user_error)?;
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        // example 'workspace list' output for reference:
+        //
+        // default: qpzoy 3ea66ab9 (no description set)
+        // test-ws: powxr b19e946c (empty) (no description set)
```
bnjmnt4n, line 174: `jj workspace list` uses the `templates.commit_summary` template to render the workspace's current working copy commit. This means that the output format listed here might not be the actual output of `jj workspace list` if the user has modified the commit summary template.

A possible workaround is to specify `jj --config-toml templates.commit_summary "[custom template]"` or add a `jj workspace list --template "[custom template]" argument to ensure that the output format matches what we expect.

yuja: Maybe we can use `jj log -r 'working_copies()' -T..`?

senekor: I applied the `--config-toml` approach and it seems to work well. Is there an advantage to using `log`? I think it could complicate some cases, for example when two workspaces point at the same revision.

yuja: No. I just meant `log` can be used as a tool to query this kind of things. (And I wouldn't want to add `workspace list -T` that only affects the summary part.)

cli/src/complete.rs
```diff
@@ -271,6 +271,31 @@ pub fn operations() -> Vec<CompletionCandidate> {
...
+            .arg("workspace")
+            .arg("list")
+            .output()
+            .map_err(user_error)?;
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        Ok(stdout
+            .lines()
+            .map(|line| {
+                let (name, desc) = line.split_once(": ").expect("output should contain ': '");
```
yuja, line 288: nit: I would make it not panic because we're parsing command output.

cli/src/complete.rs
```diff
@@ -271,6 +271,31 @@ pub fn operations() -> Vec<CompletionCandidate> {
...
+                let help = match desc {
+                    "" => Some("(no description set)".to_string().into()),
```
yuja, line 290: nit: this could be moved to template `if(description, description.first_line(), description_placeholder)` for consistency, though the result is mostly identical.

yuja (approved):

PR #4871, opened 2024-11-15
cli: git push: do not push new bookmarks by default

yuja:
This basically implements https://github.com/martinvonz/jj/pull/4730#issuecomment-2451082883

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj move` has been removed. It was deprecated in 0.16.0.
 
+* `jj git push` no longer pushes new bookmarks by default. Use `--allow-new` to
+  bypass this restriction.
+
 ### Deprecations
 
 ### New features
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -57,7 +57,7 @@ use crate::ui::Ui;
 
 /// Push to a Git remote
 ///
-/// By default, pushes any bookmarks pointing to
+/// By default, pushes tracking bookmarks pointing to
 /// `remote_bookmarks(remote=<remote>)..@`. Use `--bookmark` to push specific
 /// bookmarks. Use `--all` to push all bookmarks. Use `--change` to generate
 /// bookmark names based on the change IDs of specific commits.
@@ -98,7 +98,7 @@ pub struct GitPushArgs {
         add = ArgValueCandidates::new(complete::local_bookmarks),
     )]
     bookmark: Vec<StringPattern>,
-    /// Push all bookmarks (including deleted bookmarks)
+    /// Push all bookmarks (including new and deleted bookmarks)
     #[arg(long)]
     all: bool,
     /// Push all tracked bookmarks (including deleted bookmarks)
@@ -115,6 +115,11 @@ pub struct GitPushArgs {
     /// correspond to missing local bookmarks.
     #[arg(long)]
     deleted: bool,
+    /// Allow pushing new bookmarks
+    ///
+    /// Newly-created remote bookmarks will be tracked automatically.
+    #[arg(long, conflicts_with = "what")]
+    allow_new: bool,
     /// Allow pushing commits with empty descriptions
     #[arg(long)]
     allow_empty_description: bool,
@@ -127,8 +132,9 @@ pub struct GitPushArgs {
     /// Push this commit by creating a bookmark based on its change ID (can be
     /// repeated)
     ///
-    /// Use the `git.push-bookmark-prefix` setting to change the prefix for
-    /// generated names.
+    /// The created bookmark will be tracked automatically. Use the
+    /// `git.push-bookmark-prefix` setting to change the prefix for generated
+    /// names.
     #[arg(long, short)]
     change: Vec<RevisionArg>,
     /// Only display what will change on the remote
@@ -172,7 +178,8 @@ pub fn cmd_git_push(
     let mut bookmark_updates = vec![];
     if args.all {
         for (bookmark_name, targets) in view.local_remote_bookmarks(&remote) {
-            match classify_bookmark_update(bookmark_name, &remote, targets) {
+            let allow_new = true; // implied by --all
+            match classify_bookmark_update(bookmark_name, &remote, targets, allow_new) {
                 Ok(Some(update)) => bookmark_updates.push((bookmark_name.to_owned(), update)),
                 Ok(None) => {}
                 Err(reason) => reason.print(ui)?,
@@ -184,7 +191,8 @@ pub fn cmd_git_push(
             if !targets.remote_ref.is_tracking() {
                 continue;
             }
-            match classify_bookmark_update(bookmark_name, &remote, targets) {
+            let allow_new = false; // doesn't matter
+            match classify_bookmark_update(bookmark_name, &remote, targets, allow_new) {
                 Ok(Some(update)) => bookmark_updates.push((bookmark_name.to_owned(), update)),
                 Ok(None) => {}
                 Err(reason) => reason.print(ui)?,
@@ -196,7 +204,8 @@ pub fn cmd_git_push(
             if targets.local_target.is_present() {
                 continue;
             }
-            match classify_bookmark_update(bookmark_name, &remote, targets) {
+            let allow_new = false; // doesn't matter
+            match classify_bookmark_update(bookmark_name, &remote, targets, allow_new) {
                 Ok(Some(update)) => bookmark_updates.push((bookmark_name.to_owned(), update)),
                 Ok(None) => {}
                 Err(reason) => reason.print(ui)?,
@@ -228,12 +237,27 @@ pub fn cmd_git_push(
             (bookmark_name.as_ref(), targets)
         });
         let view = tx.repo().view();
-        let bookmarks_by_name = find_bookmarks_to_push(view, &args.bookmark, &remote)?;
-        for (bookmark_name, targets) in change_bookmarks.chain(bookmarks_by_name.iter().copied()) {
+        for (bookmark_name, targets) in change_bookmarks {
             if !seen_bookmarks.insert(bookmark_name) {
                 continue;
             }
-            match classify_bookmark_update(bookmark_name, &remote, targets) {
+            let allow_new = true; // --change implies creation of remote bookmark
+            match classify_bookmark_update(bookmark_name, &remote, targets, allow_new) {
+                Ok(Some(update)) => bookmark_updates.push((bookmark_name.to_owned(), update)),
+                Ok(None) => writeln!(
+                    ui.status(),
+                    "Bookmark {bookmark_name}@{remote} already matches {bookmark_name}",
+                )?,
+                Err(reason) => return Err(reason.into()),
+            }
+        }
+
+        let bookmarks_by_name = find_bookmarks_to_push(view, &args.bookmark, &remote)?;
+        for &(bookmark_name, targets) in &bookmarks_by_name {
+            if !seen_bookmarks.insert(bookmark_name) {
+                continue;
+            }
+            match classify_bookmark_update(bookmark_name, &remote, targets, args.allow_new) {
                 Ok(Some(update)) => bookmark_updates.push((bookmark_name.to_owned(), update)),
                 Ok(None) => writeln!(
                     ui.status(),
@@ -256,7 +280,7 @@ pub fn cmd_git_push(
             if !seen_bookmarks.insert(bookmark_name) {
                 continue;
             }
-            match classify_bookmark_update(bookmark_name, &remote, targets) {
+            match classify_bookmark_update(bookmark_name, &remote, targets, args.allow_new) {
                 Ok(Some(update)) => bookmark_updates.push((bookmark_name.to_owned(), update)),
                 Ok(None) => {}
                 Err(reason) => reason.print(ui)?,
@@ -504,6 +528,7 @@ fn classify_bookmark_update(
     bookmark_name: &str,
     remote_name: &str,
     targets: LocalAndRemoteRef,
+    allow_new: bool,
 ) -> Result<Option<BookmarkPushUpdate>, RejectedBookmarkUpdateReason> {
     let push_action = classify_bookmark_push_action(targets);
     match push_action {
@@ -526,6 +551,18 @@ fn classify_bookmark_update(
                  bookmark."
             )),
         }),
+        BookmarkPushAction::Update(update) if update.old_target.is_none() && !allow_new => {
+            Err(RejectedBookmarkUpdateReason {
+                message: format!(
+                    "Refusing to create new remote bookmark {bookmark_name}@{remote_name}"
+                ),
+                hint: Some(
+                    "Use --allow-new to push new bookmark. Use --remote to specify the remote to \
+                     push to."
+                        .to_owned(),
+                ),
+            })
+        }
         BookmarkPushAction::Update(update) => Ok(Some(update)),
     }
 }
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -1148,7 +1148,7 @@ Create a new Git backed repo
 
 Push to a Git remote
 
-By default, pushes any bookmarks pointing to `remote_bookmarks(remote=<remote>)..@`. Use `--bookmark` to push specific bookmarks. Use `--all` to push all bookmarks. Use `--change` to generate bookmark names based on the change IDs of specific commits.
+By default, pushes tracking bookmarks pointing to `remote_bookmarks(remote=<remote>)..@`. Use `--bookmark` to push specific bookmarks. Use `--all` to push all bookmarks. Use `--change` to generate bookmark names based on the change IDs of specific commits.
 
 Before the command actually moves, creates, or deletes a remote bookmark, it makes several [safety checks]. If there is a problem, you may need to run `jj git fetch --remote <remote name>` and/or resolve some [bookmark conflicts].
 
@@ -1166,19 +1166,22 @@ Before the command actually moves, creates, or deletes a remote bookmark, it mak
 * `-b`, `--bookmark <BOOKMARK>` — Push only this bookmark, or bookmarks matching a pattern (can be repeated)
 
    By default, the specified name matches exactly. Use `glob:` prefix to select bookmarks by wildcard pattern. For details, see https://martinvonz.github.io/jj/latest/revsets#string-patterns.
-* `--all` — Push all bookmarks (including deleted bookmarks)
+* `--all` — Push all bookmarks (including new and deleted bookmarks)
 * `--tracked` — Push all tracked bookmarks (including deleted bookmarks)
 
    This usually means that the bookmark was already pushed to or fetched from the relevant remote. For details, see https://martinvonz.github.io/jj/latest/bookmarks#remotes-and-tracked-bookmarks
 * `--deleted` — Push all deleted bookmarks
 
    Only tracked bookmarks can be successfully deleted on the remote. A warning will be printed if any untracked bookmarks on the remote correspond to missing local bookmarks.
+* `--allow-new` — Allow pushing new bookmarks
+
+   Newly-created remote bookmarks will be tracked automatically.
 * `--allow-empty-description` — Allow pushing commits with empty descriptions
 * `--allow-private` — Allow pushing commits that are private
 * `-r`, `--revisions <REVISIONS>` — Push bookmarks pointing to these commits (can be repeated)
 * `-c`, `--change <CHANGE>` — Push this commit by creating a bookmark based on its change ID (can be repeated)
 
-   Use the `git.push-bookmark-prefix` setting to change the prefix for generated names.
+   The created bookmark will be tracked automatically. Use the `git.push-bookmark-prefix` setting to change the prefix for generated names.
 * `--dry-run` — Only display what will change on the remote
 
 
--- a/cli/tests/test_bookmark_command.rs
+++ b/cli/tests/test_bookmark_command.rs
@@ -196,7 +196,7 @@ fn test_bookmark_move() {
 
     // Delete bookmark locally, but is still tracking remote
     test_env.jj_cmd_ok(&repo_path, &["describe", "@-", "-mcommit"]);
-    test_env.jj_cmd_ok(&repo_path, &["git", "push", "-r@-"]);
+    test_env.jj_cmd_ok(&repo_path, &["git", "push", "--allow-new", "-r@-"]);
     test_env.jj_cmd_ok(&repo_path, &["bookmark", "delete", "foo"]);
     insta::assert_snapshot!(get_bookmark_output(&test_env, &repo_path), @r###"
     foo (deleted)
@@ -441,7 +441,7 @@ fn test_bookmark_rename() {
     test_env.jj_cmd_ok(&repo_path, &["new"]);
     test_env.jj_cmd_ok(&repo_path, &["describe", "-m=commit-2"]);
     test_env.jj_cmd_ok(&repo_path, &["bookmark", "create", "bremote"]);
-    test_env.jj_cmd_ok(&repo_path, &["git", "push", "-b=bremote"]);
+    test_env.jj_cmd_ok(&repo_path, &["git", "push", "--allow-new", "-b=bremote"]);
     let (_stdout, stderr) =
         test_env.jj_cmd_ok(&repo_path, &["bookmark", "rename", "bremote", "bremote2"]);
     insta::assert_snapshot!(stderr, @r###"
@@ -1081,7 +1081,7 @@ fn test_bookmark_track_conflict() {
     );
     test_env.jj_cmd_ok(&repo_path, &["bookmark", "create", "main"]);
     test_env.jj_cmd_ok(&repo_path, &["describe", "-m", "a"]);
-    test_env.jj_cmd_ok(&repo_path, &["git", "push", "-b", "main"]);
+    test_env.jj_cmd_ok(&repo_path, &["git", "push", "--allow-new", "-b", "main"]);
     test_env.jj_cmd_ok(&repo_path, &["bookmark", "untrack", "main@origin"]);
     test_env.jj_cmd_ok(
         &repo_path,
@@ -1757,6 +1757,7 @@ fn test_bookmark_list_tracked() {
         &[
             "git",
             "push",
+            "--allow-new",
             "--remote",
             "upstream",
             "--bookmark",
--- a/cli/tests/test_completion.rs
+++ b/cli/tests/test_completion.rs
@@ -49,7 +49,10 @@ fn test_bookmark_names() {
     test_env.jj_cmd_ok(&repo_path, &["desc", "-r", "aaa-tracked", "-m", "x"]);
     test_env.jj_cmd_ok(&repo_path, &["bookmark", "create", "bbb-tracked"]);
     test_env.jj_cmd_ok(&repo_path, &["desc", "-r", "bbb-tracked", "-m", "x"]);
-    test_env.jj_cmd_ok(&repo_path, &["git", "push", "--bookmark", "glob:*-tracked"]);
+    test_env.jj_cmd_ok(
+        &repo_path,
+        &["git", "push", "--allow-new", "--bookmark", "glob:*-tracked"],
+    );
 
     test_env.jj_cmd_ok(&origin_path, &["bookmark", "create", "aaa-untracked"]);
     test_env.jj_cmd_ok(&origin_path, &["desc", "-r", "aaa-untracked", "-m", "x"]);
--- a/cli/tests/test_git_private_commits.rs
+++ b/cli/tests/test_git_private_commits.rs
@@ -67,7 +67,14 @@ fn set_up_remote_at_main(test_env: &TestEnvironment, workspace_root: &Path, remo
     );
     test_env.jj_cmd_ok(
         workspace_root,
-        &["git", "push", "--remote", remote_name, "-b=main"],
+        &[
+            "git",
+            "push",
+            "--allow-new",
+            "--remote",
+            remote_name,
+            "-b=main",
+        ],
     );
 }
 
@@ -155,7 +162,10 @@ fn test_git_private_commits_not_directly_in_line_block_pushing() {
     test_env.jj_cmd_ok(&workspace_root, &["bookmark", "create", "bookmark1"]);
 
     test_env.add_config(r#"git.private-commits = "description(glob:'private*')""#);
-    let stderr = test_env.jj_cmd_failure(&workspace_root, &["git", "push", "-b=bookmark1"]);
+    let stderr = test_env.jj_cmd_failure(
+        &workspace_root,
+        &["git", "push", "--allow-new", "-b=bookmark1"],
+    );
     insta::assert_snapshot!(stderr, @r###"
     Error: Won't push commit f1253a9b1ea9 since it is private
     "###);
@@ -192,8 +202,10 @@ fn test_git_private_commits_already_on_the_remote_do_not_block_push() {
     test_env.jj_cmd_ok(&workspace_root, &["new", "main", "-m=private 1"]);
     test_env.jj_cmd_ok(&workspace_root, &["new", "-m=public 3"]);
     test_env.jj_cmd_ok(&workspace_root, &["bookmark", "set", "main"]);
-    let (_, stderr) =
-        test_env.jj_cmd_ok(&workspace_root, &["git", "push", "-b=main", "-b=bookmark1"]);
+    let (_, stderr) = test_env.jj_cmd_ok(
+        &workspace_root,
+        &["git", "push", "--allow-new", "-b=main", "-b=bookmark1"],
+    );
     insta::assert_snapshot!(stderr, @r#"
     Changes to push to origin:
       Move forward bookmark main from 7eb97bf230ad to fbb352762352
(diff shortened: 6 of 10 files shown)


martinvonz (approved):
Thanks. This sounds good to me.

jennings:
I just updated to a version that includes this, and I find it intrusive because:

- Most of my repos have a single remote, so I don't have the "push to the wrong remote" problem
- I don't create bookmarks until I want to push them, so I don't have bookmarks I intended to remain private

@emilazy [said in Discord](https://discord.com/channels/968932220549103686/968932220549103689/1308844845636259910):

> having to keep track of bookmark states feels almost like a staging area

This makes sense to me, I liked that before it was "stateless". I knew exactly which bookmarks were going to push because either I specified `-b`/`-c`/`-r`, or I know the push rules that bare `jj git push` follows.

If this behavior is kept, I would at least want a config setting to disable it. I think I would also want to disable the behavior added in !4730 for similar reasons.

yuja:
> is the issue that newly‐created bookmarks are automatically pushed to all remotes?

Kind of yes. Any bookmarks can be pushed to any remotes as long as the remote doesn't have conflicting bookmarks of the same name.

> What if instead newly‐created bookmarks didn’t track anything and you had to explicitly `jj bookmark track foo@origin` to get it to be pushed at all?

That's totally doable, but I'm not sure if it will improve the UX. We'll have to do `jj bookmark create foo && jj bookmark track foo@origin && jj git push` instead of `jj bookmark create foo && jj git push --allow-new`.

> I just updated to a version that includes this, and I find it intrusive because:
>
> * Most of my repos have a single remote, so I don't have the "push to the wrong remote" problem
> * I don't create bookmarks until I want to push them, so I don't have bookmarks I intended to remain private

My workflow is basically the same. I didn't notice it was scary until I had to work on a security fix which had to be pushed to a temporary remote.

emilazy:
> That's totally doable, but I'm not sure if it will improve the UX. We'll have to do `jj bookmark create foo && jj bookmark track foo@origin && jj git push` instead of `jj bookmark create foo && jj git push --allow-new`.

I personally find it better, because the model already involves keeping track of the state of “what bookmarks on what remotes are tracked”, and this is something we have existing UI to expose and manipulate. It’s one more command, but it situates the state of how pushing treats the bookmark where it already is (bookmark tracking) rather than making it an additional consideration (whether it has been pushed before, essentially).

My impression is that most new bookmarks to push are created with `jj git push -c`, which would of course continue automatically tracking it. We could also have an option like `jj bookmark create foo --track origin` or maybe even `jj bookmark create foo@origin` if the extra command seems too annoying.

I’d be interested in hearing what @jennings thinks of my proposal, though, since if it doesn’t sound promising to him it might not be a good solution to the problem.

ilyagr:
It might make sense to allow both `git push --allow-new` and `jj branch track branch@remote-where-it-does-not-yet-exist` (with a friendly message in the latter case).

Much like with `jj commit` vs `jj describe`, sometimes the user would want to mark a branch as allowed-to-be-pushed-to-this-remote when you are pushing, and sometimes the user might prefer to make this designation when creating the branch (or at some intermediate time).

jennings:
> `jj branch track branch@remote-where-it-does-not-yet-exist`

This hypothetical command didn't make sense to me at first, because I thought, How can you track a remote branch that doesn't exist yet? But then I realized this is probably due to the internal implementation of bookmarks:

```rust
struct Bookmark {
  name: String,
  remote_bookmarks: Vec<RemoteBookmark>,
}
```

So I'm guessing `jj bookmark track branch@remote` adds `branch@remote` into the `remote_bookmarks` vector, even if the remote branch doesn't actually exist yet? And thanks to #4730, once the branch tracks one remote, jj won't push the bookmark to any new remotes without being explicit?

I like the result, but I think "track" is a confusing verb for this because it only makes sense if you know this implementation detail. I think of this action as putting a restriction on the branch `foo`, so the verbs that make more sense to me are:

```
jj bookmark restrict foo --remote bar
jj bookmark limit foo --remote bar
jj bookmark config foo --restrict-to-remote bar
```

emilazy:
Seeing it as a restriction is strange to me. I think of it as a request: “I want `foo` and `foo@remote` to be synchronized”. Then the semantics of a new bookmark not tracking anything by default, and opting in to which remotes you want it to be reflected on, gives us the “no surprise pushing of new bookmarks” behaviour without introducing a new kind of state to keep track of (“is the bookmark new or not”). Perhaps “track” is the wrong verb, though, but I don’t think “restrict” or “limit” are correct – every track command on the same local bookmark would allow it to be pushed to increasingly more remotes.

martinvonz:
What's the conclusion here? I always use `jj git push -c` myself, and almost always with only one remote, so I'm not affected much myself, and I find it hard to imagine how annoying it is for others.

yuja:
My feeling is that `jj bookmark track new-bookmark-to-push@origin` is technically correct, but it wouldn't help unless the user understands the underlying model. Maybe we can add a config knob akin to `git.auto-local-bookmark`, but I'm not sure if that's common enough to add such option. FWIW, Ilyagr added `-N` for `--allow-new`, which I think is good.

chriskrycho:
Following up on this since I first hit it today after the 0.24 release, and I am similarly not a fan, at least as it is configured. Not pushing to *all remotes* by default seems reasonable to me, but if there is only one remote, it seems like needless ceremony. “Yeah, I just said explicitly to push a new branch, and you’re telling me I can’t because… reasons?” I strongly suspect that for most users, this will quickly turn into `jj git push -N the-branch` or just straight-up aliasing `jj git push` to do it automatically, rather like many Git users end up defaulting to `git ci -a` because they don’t want to bother with a staging area—and if so, it won’t help in precisely the scenarios where it’s a concern!

My own sense of the best flow here would be something like this:

- If there is only a single remote, just push to it with no extra ceremony.
- If the user has a *default* remote (a concept I am not sure whether jj currently understands? I haven’t checked!), only push to that remote by default.
- Pushes to multiple remotes (including *all* remotes) or to non-default remotes should require being explicit.

That would solve the safety issue highlighted here, but would also make the normal workflow “just work”. It’s classic “progressive disclosure of complexity”: if the situation you’re working in is simple, the interaction should be simple; and if the situation is more complex, it is okay for the interaction to be more complex.

yuja:
I'm not a fan of making the check conditional based on the number of the registered remotes. The UX would be suddenly broken if the user added new remote to e.g. fetch changes to review.

The original PR (which requires some flag if local bookmark is tracking another remote) might be a better compromise. It wouldn't help users who want to use bookmarks for local stuff, but that's probably rare?

senekor:
I strongly agree with Chris' opinion. The `--allow-new` flag should not be required if there is only one remote. The rationale for requiring the `--allow-new` flag only applies to situations with multiple remotes.

> I'm not a fan of making the check conditional based on the number of the registered remotes. The UX would be suddenly broken if the user added new remote to e.g. fetch changes to review.

I don't agree that this UX is broken, it makes perfect sense to me:
1. Pushing a new bookmark with only one remote: unambiguous situation, go right ahead.
2. Pushing a new bookmark with multiple remotes: ambiguous situation, please clarify and retry.

That is way more intuitive and understandable than: "Oh, you're making a request that's completely unambiguous? Can you please just repeat your intent a second time mindlessly? Because there is this other situation that has nothing to do with you that requires a confirmation dialog and I want to be consistent so I'm giving you a useless confirmation dialog as well."

I'm currently writing a tutorial for VCS beginners and I have no idea how I'm supposed to explain the status quo to people. Do I have to introduce the concept of tracking bookmarks across multiple remotes before they can even push the `main` branch they just created for the first time? I try _very hard_ not to force readers to mindlessly memorize commands and flags which they have _no hope_ of understanding at the level they're currently at.

---

This second point is less important to me, but I think the `--allow-new` flag doesn't make sense in any situation, only `--remote` does. If I have multiple remotes, it makes sense that Jujutsu refuses a normal `jj git push --bookmark new-bookmark`, because it's ambiguous which remote I want to push to. However, the flag `--allow-new` doesn't resolve that ambiguity. I haven't actually specified which remote I want to push to! So in my opinion, it would make the most sense to get rid of `--allow-new` entirely and recommend `--remote origin` instead.

yuja:
> I strongly agree with Chris' opinion. The `--allow-new` flag should not be required if there is only one remote. The rationale for requiring the `--allow-new` flag only applies to situations with multiple remotes.

There's another scenario in which the user wants the current behavior. See https://github.com/jj-vcs/jj/issues/5053

To me, creation of remote bookmarks isn't the same as updating existing remote bookmarks, so it makes sense that explicit flag is required. This check can be turned off by config knob, so I don't think we need another "--allow-new-if-there-are-multiple-remotes" behavior.

ilyagr:
I haven't thought about this situation carefully in a bit, but I personally enjoy the interface with `--allow-new`. Creating a new branch on the remote is a big deal in my mind, and my current workflow in the `jj` repo depends on pushing different branches to upstream and origin (the upstream ones are prefixed with `ig/`). Even in single-remote repos sometimes I even have local-only branches. This seems to be similar to #5053.

OTOH, I can also see how new users could be confused by it, especially if they don't want to do anything fancy and have a single remote.

Perhaps we should advertise the config knob more as something one might want to turn on in personal repositories (such as test repositories for tutorials)? You could suggest the `jj config set --repo` command for this, and we could document it more.

Other than that, I'm not sure what to do. Explicitly setting the option does seem better to me than magic changes in behavior when the number of remotes changes.

senekor:
> There's another scenario in which the user wants the current behavior. See https://github.com/jj-vcs/jj/issues/5053

This has nothing to do with the topic here? The linked issue is about private commits. IIUC the private/public commit distinction is orthogonal to the new/tracked bookmark distinction. Private commits cannot be pushed, it doesn't matter if the pushed bookmark that contains the private commits is new or tracked.

> To me, creation of remote bookmarks isn't the same as updating existing remote bookmarks

I think the "upsert" semantics make sense for pushing bookmarks. This reminds me of the discussion around getting rid of `bookmark set`. In that case, there was an actual reason why the upsert semantics could be harmful: Having a typo in the bookmark name would create a new bookmark instead of moving an existing one. That is simply not the case for `push`. Making a typo in `jj git push --bookmark <name>` does not result in a misspelled bookmark being created and pushed to the remote.

So, while I guess it's fine to think of create and update as different operations, to me that's still not a convincing reason why the two can't be accomplished by the same command.

Creating a regular commit and creating a merge commit can be considered two conceptually different operations. But they can both be achieved with the `jj new` command without any downsides. Nobody complains about that.

And lastly, if "create and update are different" was truly the reason why the CLI must be unnecessarily inconvenient, the `--allow-new` flag wouldn't be the solution. `--allow-new` doesn't have "create" semantics, it has "upsert" semantics. So an actual solution would be to replace `jj git push` with two new commands `jj git push-new` and `jj git push-update` or something like that.

While I wouldn't like that approach at all, at least I might have _some_ chance of explaining that in my tutorial.

> This check can be turned off by config knob

I think this is possibly the biggest sign that the current design is completely broken. The first reaction of many users when they encountered this new "feature" was clearly "Oh god, how can I turn this off??".

And people don't turn this feature off because they don't want to be protected from accidentally pushing to the wrong remote! Allowing new bookmarks to be pushed by default _if there is only one remote_ would greatly alleviate the need for this config option. We could deactivate and deprecate it. Which means the actual safety purpose of the flag would help many more users that would otherwise have turned it off completely just for convenience.

> I don't think we need another "--allow-new-if-there-are-multiple-remotes" behavior.

You make it sound like what I'm proposing is going to increase the surface area of the CLI, make it more complicated. The opposite is the case. The last paragraph of my previous comment argues that the `--allow-new` flag is completely redundant and I have working changes which remove (deprecate) it entirely. Just need to clean it up a little for a PR.

The semantics are very intuitive: To push new bookmarks, the remote to push to must be *unambiguous*. If there is only one remote, it is unambiguous, no further information needed. If there are multiple remotes, the desired one to push to must be specified with `--remote`. Way simpler and intuitive than the current soup of `--allow-new`, `--remote` _and a config option_.

---

> You could suggest the `jj config set --repo` command for this, and we could document it more.

No no no! That would be horrible! I would be telling my readers: "Jujutsu's default behavior is confusing and inconvenient, so just run these magical incantations to fix what upstream couldn't."

And then what? When my readers clone their first real project, they will find that **I lied to them** and `jj git push --bookmark name` doesn't work the way I explained it to them.

If anything, I'd have to tell them to set the config globally. And I don't see how anyone could be OK with that. If the default behavior of Jujutsu is so bad that people start telling new users unconditionally to configure it, we should **fix the default behavior instead**. We aren't 1.0 yet, we don't have to carry around bad defaults for eternity because of backwards-compatibility.

ilyagr:
> Jujutsu's default behavior is confusing and inconvenient

I think the use-case of a personal repo (one where only one person ever pushes to them) is fundamentally different than the use-case of a repo one works with other people on. There are several instances where Jujutsu's behavior is more confusing and less convenient for the personal repo use-case (**Update:** in order to work better for the shared repo use-case). Another example is the notion of tracked bookmarks.

\[**Aside:** In my mind, the fact that `jj` doesn't auto-advance branches, and the focus on editing commits, also has to do with optimizing for working with other people (who want to review easily understandable sequences of commits as opposed to the history of how the sausage was made).\]

So, I think it might be OK to suggest different settings for the personal repo use-case. I'm hoping this is possible to communicate in such a way that people understand both workflows (though I don't think we do this very well -- or maybe at all -- at the moment).

-------------

But your concerns are certainly warranted (and I'm aware I probably haven't addressed all of them; this reply is meant to be quick rather than comprehensive). If we could figure out one data model and one workflow that worked perfectly for both personal and collaborative repos, that would of course be better. The trade-offs we're making are not pleasant, and we might be able to choose better ones.

I think the fact that you're writing a tutorial makes the warts all the clearer. It might also lead to better ideas.

I'm mainly trying to pass along the point of view and the use-cases for which the current design makes sense. These are also the use-cases we'd have to consider carefully if we change the design, since they are the easiest to break.

senekor:
> If we could figure out one data model and one workflow that worked perfectly for both personal and collaborative repos, that would of course be better.

This is my proposal:

> To push new bookmarks, the remote to push to must be _unambiguous_. If there is only one remote, it is unambiguous, no further information needed. If there are multiple remotes, the desired one to push to must be specified with `--remote`.

I don't see any downsides of this model for the collaborative use case. If you do, I'd love to hear them.

ilyagr:
The main problem I see with this is that the behavior of `jj push` changes unexpectedly when you add or remove a remote. You might be used to a multi-remote behavior, do a `jj git remote remove`, then do a `jj git push` and push too much.

This could also happen in a script that would suddenly be buggy for (say) repos that have 2 remotes. To be fair, I guess a script would need to fix the config settings that affect its behavior anyway.

At the moment this seems like a severe problem to me. This is despite the fact that, if what I wrote above isn't convincing enough, I'm not sure I have a more convincing argument at the moment.

yuja:
> > There's another scenario in which the user wants the current behavior. See #5053
> 
> This has nothing to do with the topic here? The linked issue is about private commits.

#5053 was solved by `--allow-new`. The user wants local "bookmarks" and they don't want to push them accidentally. That makes kind of sense.

> The main problem I see with this is that the behavior of jj push changes unexpectedly when you add or remove a remote.

Yes, that's another concern.

senekor:
> The main problem I see with this is that the behavior of `jj` push changes unexpectedly when you add or remove a remote.

I don't understand how this is unexpected. When I have two remotes and I tell Jujutsu to "push this bookmark to the remote", I expect Jujutsu to respond with the question "which one?" If I then remove a remote so I'm left with only one and I repeat my request, I expect Jujutsu to complete my now-unambiguous request.

Here's another example. Assume `@` is a merge commit with two parents. `jj show @-` will fail, because `@-` resolves to two commits. If I abandon one of them and rerun `jj show @-`, it will succeed. Should we conclude from this that the behavior of `jj show` is inconsistent and unexpected? Because it fails or succeeds "magically" based on how many parents the working copy has? No! Ambiguous requests fail and unambiguous requests succeed. This is perfectly expected.

> #5053 was solved by `--allow-new`.

This is simply incorrect. `--allow-new` has nothing to do with what the issue is describing. Based on the reproduction, you can confirm it like this:

```sh
jj git push --bookmark a -N # succeeds, because not pointing at private commits
jj bookmark move a --to local/a
jj git push --bookmark a # fails, because now pointing at private commits
```

ilyagr:
> The user wants local "bookmarks" and they don't want to push them accidentally. That makes kind of sense.

I certainly like being able to have local bookmarks this way.

If we changed the behavior of `--allow-new`, we'd probably want to support this use-case with some sort of "keep this bookmark local" config, or by making things work the way #5053 asked.

Aside: In my mind, this is the connection between `--allow-new` and #5053. It's not direct, and I'm not 100% sure this is what Yuya meant to by linking this bug, but in any case I think the use-case matters more than that specific bug.

------

So, the two currently written down[^where] objections to https://github.com/jj-vcs/jj/pull/4871#issuecomment-3067629013 are the inconsistency in https://github.com/jj-vcs/jj/pull/4871#issuecomment-3067683382 and the local bookmark use-case (which might be fixable, though the fix could also end up being too ugly, e.g. if it also needs to work differently depending on the number of remotes).

I don't think either of these objections is trivial, though they also both seem potentially surpassable maybe. At least, I feel like neither is perfectly understood by me nor is absolute.

[^where]: Only considering our latest discussion today, I haven't re-read the whole thread nor do I remember the earlier discussion very well.

senekor:
...I think I'm finally starting to understand.

You are abusing the property of tracked/untracked as an additional filter on the default push revset. While I can see the appeal and convenience of that, you must agree that this is a **horrible** hack.

There are many problems with this:
- You are forcing this filter on everybody else who doesn't want it.
- The way it's implemented, the filter also applies if the flags `--bookmark` or `--revisions` are used. This makes **zero** sense, because you're already explicitly specifiying which bookmarks to push. Why impose an arbitrary filter on top of that which must be opted-out of with a non-sensical flag like `--allow-new`?

This is just working around the fact that the [configurable push revset](https://github.com/jj-vcs/jj/pull/5985) hasn't landed yet.

But why? Can't you already do this today very easily? Just make an alias for `jj git push --revisions "<insert arbitrary revset where you can distinguish between local and non-local bookmark however you like>"`.

senekor:
Wait, why am I overcompliating this. This is literally just `jj git push --tracked`. Why not use that for the use case of treating untracked bookmarks as "local"?

senekor:
And applied to the issue you keep referencing:

```sh
jj git push -r 'bookmarks(regex:"^[^(local/)]")'
```

yuja:
> You are abusing the property of tracked/untracked as an additional filter on the default push revset.

No, it's not abuse. A local bookmark which isn't associated with the remote shouldn't be pushed implicitly. If you want this implicit behavior, you can set `git.push-new-bookmarks = true`. The word "tracked" might be misleading, but it basically means the bookmark is not local-only.

senekor:
How is it implicit? The default push revset is well-defined. The command `jj git push` means "push all bookmarks in the default push revset". If that contains local-only bookmarks, then those should obviously be included. And there's nothing implicit about that.

FWIW if you just think that local-only bookmarks should not be part of the default push revset, then I have no strong opinion about that. Fine by me to restrict it in that way.

Even if we accept that argument, that does **not** apply to `--bookmark` and `--revisions`. In those cases, the user is specifying the bookmarks they want to push even more explicitly than with `jj git push`. It doesn't make any sense to require the `--allow-new` flag in those cases.

yuja:
> How is it implicit?

Creating "new" remote bookmarks implicitly.

Perhaps, you have different mental model than me. Some forks believe that `jj git push` should create new remote bookmarks because it is the command to do so, and local bookmarks are primarily meant for pushing. Others (including me) consider that `jj git push` should just synchronize local state with the remote, and creation (or tracking) of new remote bookmarks should be explicit (because doing so would result in new state.)

Maybe you can find previous discussion in this or related issues/PRs/discord, but I don't have a pointer.

senekor:
> Others (including me) consider that jj git push should just synchronize local state with the remote, and creation (or tracking) of new remote bookmarks should be explicit

I realize that I write a lot and I can't expect people to read & respond to everything. But it is frustrating to be met with the same arguments that I have already addressed _but didn't get a response to myself_. I'll just copy-paste what I've already said about this, maybe you can respond this time:

I think the "upsert" semantics make sense for pushing bookmarks. This reminds me of the discussion around getting rid of bookmark set. In that case, there was an actual reason why the upsert semantics could be harmful: Having a typo in the bookmark name would create a new bookmark instead of moving an existing one. That is simply not the case for push. Making a typo in `jj git push --bookmark <name>` does not result in a misspelled bookmark being created and pushed to the remote.

So, while I guess it's fine to think of create and update as different operations, to me that's still not a convincing reason why the two can't be accomplished by the same command.

Creating a regular commit and creating a merge commit can be considered two conceptually different operations. But they can both be achieved with the `jj new` command without any downsides. Nobody complains about that.

And lastly, if "create and update are different" was truly the reason why the CLI must be unnecessarily inconvenient, the `--allow-new` flag wouldn't be the solution. `--allow-new` doesn't have "create" semantics, it has "upsert" semantics. So an actual solution would be to replace `jj git push` with two new commands `jj git push-new` and `jj git push-update` or something like that.

While I wouldn't like that approach at all, at least I might have some chance of explaining that in my tutorial.

yuja:
> I realize that I write a lot and I can't expect people to read & respond to everything. But it is frustrating to be met with the same arguments that I have already addressed _but didn't get a response to myself_. I'll just copy-paste what I've already said about this, maybe you can respond this time:

FWIW, it's frustrating that I have to repeat this discussion despite the problem was addressed by the config knob.

> I think the "upsert" semantics make sense for pushing bookmarks.

I disagree, but I understand it makes sense for some forks.

> This reminds me of the discussion around getting rid of bookmark set. In that case, there was an actual reason why the upsert semantics could be harmful: Having a typo in the bookmark name would create a new bookmark instead of moving an existing one. That is simply not the case for push.

No. There may be local bookmarks which the user isn't intended to push.

> Making a typo in `jj git push --bookmark <name>` does not result in a misspelled bookmark being created and pushed to the remote.

That's true. If the bookmark name is explicitly specified, there would be high chance that the user wants to create that bookmark. To me, it's not a strong reason to introduce inconsistent `--allow-new` handling, though.

> And lastly, if "create and update are different" was truly the reason why the CLI must be unnecessarily inconvenient, the `--allow-new` flag wouldn't be the solution. `--allow-new` doesn't have "create" semantics, it has "upsert" semantics. So an actual solution would be to replace `jj git push` with two new commands `jj git push-new` and `jj git push-update` or something like that.

There was an idea to allow `jj bookmark track new-local-bookmark@origin && jj git push`. Naming aside, it is the most technically correct choice. However, I think it's hard to discover without understanding the data model, and it's inconvenient for people who just want to push new bookmarks, so `--allow-new` and `git.push-new-bookmarks` was added as a compromise.

martinvonz:
It seems like this discussion is getting heated. Perhaps it's best to take a break and discuss it more another day? I have some thoughts but I'll wait until another day so I don't share something some of you may disagree with and then tell you to not respond to it :)

senekor:
Good idea, thanks :) (I don't mind not responding for the moment if you'd like to get your ideas out now)

For reference, I've implemented my proposal mostly to make sure I understand it myself: #6981 I'm happy to throw it away if the disussion comes to a different conclusion.

ilyagr:
> There was an idea to allow jj bookmark track new-local-bookmark@origin && jj git push. Naming aside, it is the most technically correct choice. However, I think it's hard to discover without understanding the data model, and it's inconvenient for people who just want to push new bookmarks, so --allow-new and git.push-new-bookmarks was added as a compromise.

Another compromise we added was `jj git push --named new-local-bookmark=@` to create and push a bookmark at the same time. I use it all the time (which is no surprise, I suggested it), though I don't know how popular it is in general or how easily people discover it.

Hopefully I didn't just do what Martin tried to avoid doing :). I'm adding this for completeness, context, and perhaps as an example for later; I don't think this greatly affects anybody's fundamental points.

senekor:
## Summary of discussion so far

It's been a couple days and the thread is getting long, I'll try to summarize by arguments in favor of the status quo. Please tell me if you think I misrepresented an argument.

- Pushing new bookmarks when there are multiple remotes is dangerous, because secret commits can be published accidentally. (That's the actual justification in the commit message.)
  - This is fully addressed by the proposal to require `--allow-new` only when there are multiple remotes.
    - Different behavior for single vs. multiple remotes is inconsistent.
      - It's not inconsistent because one request is ambiguous (succeeds) and the other is unambiguous (fails). An analogy I gave was `jj show @-` succeeding when there's one parent but failing when there are two.

- Pushing new bookmarks is dangerous even with one remote. Users may want to keep some bookmarks local.
  - Users always specify the bookmarks they want to push, so they have already expressed their intent to push the bookmark. If that wasn't actually the case, that's a user error.
    - Users do not specify explicitly what to push with `jj git push`. The default revset is implicit.
      - That may be an argument in favor of restricting the default push revset to only include already-tracked bookmarks. Interpreted from the perspective of this argument, the `--allow-new` flag is a hacky subset of proper push revset customization, a feature that's currently in development.
      - If `jj git push` is not explicit enough, then `--bookmark` should not require `--allow-new`.
        - That may make sense but it's not worth making `--allow-new` inconsistent for that.
          - It wouldn't be inconsistent because `--named` and `--change` already don't require the flag for the same reason.

- Pushing new bookmarks is conceptually different from syncing the state of tracked bookmarks and therefore warrants a different UI.
  - That's not sufficient justification for a different UI. If two concepts can be achieved with one command, there's no reason to turn it into two commands unnecessarily. Positive example: Regular commits and merge commits can both be created with `jj new` with no downsides, though the operations are conceptually different. Negative example: `jj bookmark set` can lead to issues when the user makes a typo.
    - The downside of combining creating and updating remote bookmarks is that users may push local-only bookmarks accidentally.
      - See argument above about restricting default push revset to exclude local-only bookmarks.
  - If this mindset is accepted, then the UI should actually be made to consistently implement it, which it currently does not.
    - One suggestion in this direction was `jj bookmark track new-bookmark@origin` before pushing.
      - I agree that this makes sense as an incremental change, but IMO more work would be needed to make the whole UI consistent around this idea.

- There is a configuration option to get rid of `--allow-new`.
  - We should still care about the default experience making sense and being consistent. I'm realizing how important this actually is, now that I'm writing a tutorial for VCS beginners. I don't want to teach readers "my preferred flavor" of Jujutsu.
  - Every configuration option should be justified by multiple competing use cases that are worth supporting. Otherwise, configuration becomes an escape hatch to avoid the effort of finding consensus about actually good UI. In this case, I don't think there is sufficient justification yet that there are two valid use cases that cannot be consolidated.

## Proposals to move forward

- We drop the `--allow-new` requirement when there's only one remote. I still find the arguments against this proposal unconvincing.

- The above, plus we change the default push revset to exclude local-only bookmarks. Fine by me.

- We rethink the UI around a consistent separation of creating and updating remote bookmarks, with the goals of consistency first and convenience a close second. I have some ideas and am happy to help with the design if that's the route we choose. Still, I think it would be good to check first that there's actually maintainer consensus on this. Until now, I've only read Yuya say explicitly that the UI should be designed around this mindset.

- We drop the `--allow-new` requirement for `--bookmark`. This is consistent with `--named` and `--change` also not requiring it. While I think this would be a missed opportunity to find a better solution, it's enough to unblock my tutorial. I could tell readers to push new bookmarks with `--bookmark` every time. (Until I introduce `--change` later.) If they explore the CLI on their own and bump into the remaining inconsistencies... not my problem 🤷

(@martinvonz I'm still eager to hear your thoughts...)

emilazy:
Hi, I’ve been meaning to write up the proposal I think is the best way forward for this for now. I don’t remember if it was my idea originally and I think it dovetails with what @yuja has been talking about, but I think it hasn’t been written up in detail on GitHub and should address many of the UX concerns here. I’ve been procrastinating on it, but here’s the idea, happy to hear what people think and elaborate on request:

1. We extend the concept of tracking a bookmark on a remote to apply to ones that don’t yet exist on that remote. `jj git push` will only push bookmarks that are tracked, whether they are new or pre‐existing. This means that the default behaviour will remain as currently, i.e. no `--allow-new` by default.

2. `--allow-new` is eliminated and replaced by the more orthogonal `jj bookmark track ⟨bookmark⟩@⟨remote⟩` before pushing. If you `jj git push` and the lack of tracking is the only thing preventing a push, it can print a hint to this effct.

3. It’s fairly likely we’d want to add a `--track ⟨remote⟩` option to `jj bookmark create` that sets up the tracking for you.

   This has overlap with `jj git push --named`; unclear to me whether we’d want to eliminate it in favour of this functionality or if we value having a one‐shot create‐and‐push command that takes custom names and sets up the tracking. If we want to keep it, `jj git push --named` would presumably set up the tracking according to the remotes you’re pushing the new bookmark to, so `jj git push --remote ⟨remote⟩ --named ⟨bookmark⟩=⟨rev⟩` would be short for `jj bookmark create --track ⟨remote⟩ -r ⟨rev⟩ ⟨bookmark⟩ && jj git push -b ⟨bookmark⟩`, which would be short for `jj bookmark create -r ⟨rev⟩ ⟨bookmark⟩ && jj bookmark track ⟨bookmark⟩@⟨remote⟩ && jj git push -b ⟨bookmark⟩`.

3. To make the UX more convenient for cases like the single‐remote one where people are usually creating new bookmarks to push to one obvious remote, we introduce an `auto-track-bookmark` option, which takes a (list of) pattern(s) like `glob:*@origin`, or possibly even a revset. Three possible behaviours here:

   1. The first time we consider a bookmark – whether new locally or fetched from a remote – we store its tracking status according to the current `auto-track-bookmark`.
   2. If a bookmark has not been explicitly tracked or untracked, then operations that care about its tracking status check `auto-track-bookmark`.
   3. `auto-track-bookmark` takes precedence over manual tracking; you cannot manually untrack a bookmark it matches.

   (iii) seems like the most conceptually elegant option, but I’m mostly ambivalent. (i) and (ii) are only distinguished by what happens when you change your `auto-track-bookmark` setting. With (ii) you can’t explicitly “forget” having expressed an opinion about whether a bookmark should be tracked, which seems kind of weird.

4. `git.auto-local-bookmark = true` and `git.push-new-bookmarks = true` can both be deprecated and replaced with something like `auto-track-bookmark = "glob:*@origin"` or `auto-track-bookmark = "glob:*@*"`. I think this matches the semantics people want for a “one‐remote repository”.

   This doesn’t handle the case where you want to push all new bookmarks to that remote, but don’t want bookmarks pulled down from it to be tracked. If that’s critically important, that would be a flaw of this plan, but that’s not the impression I get – it seems like these use cases only use bookmarks for exchange with a single remote, and therefore have no use for untracked bookmarks on that remote. (In any case, the current tracking system doesn’t handle that kind of asymmetry well in general.)

Advantages I see:

1. This eliminates the weird thing where new local bookmarks “want to be pushed to every remote by default” but are held back by a flag/setting, whereas bookmarks that come down from a remote are associated with it and you need to opt in to syncing them with a local one. I find the current model very strange, and believe the mental model here is much simpler: newly‐created bookmarks are local (…because you’re creating them locally), and tracking is the sole thing that determines whether a local bookmark is synced with a remote on fetch and push.

2. Your tutorial can introduce bookmark tracking up front as the way we tell Jujutsu which bookmarks we want to be synced with which remotes, which is important to learn and a common source of confusion for new users. This would still mean telling them to set a configuration setting if you don’t want people to have to explicitly specify `--track origin` for new bookmarks, but hopefully the setting makes a lot more sense than the current “disable some safety behaviour to work around (what I see as) an exception to the general tracking model” one.

3. Automatic tracking of a subset of bookmarks is useful in other cases, too. In Nixpkgs, I often have to push manual merges to `staging` or `staging-next` to resolve conflicts. However, we also create versions of these for each stable release, i.e. `staging-25.05` and `staging-next-25.05`. Being able to set `auto-track-bookmark = "glob:staging*@origin"` would be handy for this.

I don’t think this solution is perfect because bookmark tracking in general has dodgy semantics IMO (being used for both the fetch and push end leads to weirdness that @arxanas has pointed out in the past and that we work around with things like `trunk()` in the default immutable heads). It should probably be split into multiple separate concepts for the “direction of travel”. But I feel this proposal would be better than the status quo, the configuration knobs should make it comparably convenient for the important use cases, and hopefully the result would feel less inelegant than the current situation does to me and seemingly many others.

joyously:
This discussion seems bogged down in the implementation details of how to direct one storage backend, instead of how the user can maintain a simple mental model and let the VCS do the heavy lifting. 
Is there a future `sync` command? How should it work so the human doesn't have to juggle both the content of their changes and the details of how to tell the VCS what to do? 
I say any `jj git` command should accept flags that get passed directly to Git, and be done with it. Work on the native `jj` experience as an improvement on what came before. The novice should get good results without having to set any config or use lots of parameters.

emilazy:
Some discussion on Discord of my above proposal that I don’t feel qualified/don’t have the energy to summarize right now, so I hope people can comment here as applicable. But my own additional thoughts that arose:

1. Maybe `--track-on` is nicer than `--track` for `jj bookmark create`; maybe we want to move away from `⟨bookmark⟩@⟨remote⟩` syntax for tracking if people find the idea of tracking a bookmark that doesn’t yet exist on a remote to be confusing (I personally don’t really find it confusing).

2. This might pave the way to the request to support tracking multiple bookmarks on different remotes that have the same name but are otherwise unrelated, because the model would be “local bookmarks have a set of remote bookmarks they sync with” and so you could add UI to track a remote bookmark under a different local name.

3. In a shared repository where you only create bookmarks to push your work but you don’t want everyone else’s work to be mutable, you could `auto-track-bookmark = "glob:emily/*@origin"`. In this scenario, if you don’t have a consistent format (everyone just uses the same namespace but only some bookmarks are owned by you), then the options fall back to the more explicit `jj bookmark crate --track(-on)` or `jj bookmark track` or `jj git push --named`.

senekor:
I support Emily's proposal. Eliminating `--allow-new` and `git.push-new-bookmarks` removes a lot of inconsistency from the UI. The configurable `auto-track-bookmark` pattern is very convenient and still explicit in intent. It even covers both use cases of single and multiple remotes elegantly (which none of my proposals did).

I think I will be able to explain this UI well in my tutorial. (Both the verbose, conceptually clean commands as well as the convenience options.)

yuja:
@emilazy Thank you for writing the proposal. Since this is a closed PR, can you file it as a new issue? FWIW, I generally like the idea as you might know.

emilazy:
Yep, done here: https://github.com/jj-vcs/jj/issues/7072.

PR #4876, opened 2024-11-15
cli: simplify-parents: avoid multiple rebases of the same commits

bnjmnt4n:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/simplify_parents.rs
+++ b/cli/src/commands/simplify_parents.rs
@@ -2,10 +2,7 @@ use std::collections::HashSet;
 
 use clap_complete::ArgValueCandidates;
 use itertools::Itertools;
-use jj_lib::backend::BackendResult;
 use jj_lib::revset::RevsetExpression;
-use jj_lib::rewrite::CommitRewriter;
-use jj_lib::settings::UserSettings;
 
 use crate::cli_util::CommandHelper;
 use crate::cli_util::RevisionArg;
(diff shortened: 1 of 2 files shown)


martinvonz:
I assigned to @torquestomp since they wrote the current version, but they're out today, so don't expect a review until Monday, @bnjmnt4n. (I'm guessing there's no rush, but let us know if you want me to review it sooner.)

cli/src/commands/simplify_parents.rs
```diff
@@ -57,63 +54,53 @@ pub(crate) fn cmd_simplify_parents(
...
+            {
+                if rewriter.parents_changed() {
+                    rewriter.reparent(command.settings())?.write()?;
+                    reparented_descendants += 1;
+                }
+                return Ok(());
             }
 
+            let num_old_heads = rewriter.new_parents().len();
+            rewriter.simplify_ancestor_merge();
```
yuja, line 74: nit: may be the `if ..` condition can be inserted here? We'll do `reparent()` anyway if ancestors were simplified.

bnjmnt4n: The intent of putting the `if` block at the start was to avoid calling `rewriter.simplify_ancestor_merge()` when unneeded.

yuja: Yes. I just meant
```rust
let num_old_heads = ..;
if target_rev && num_old_heads > 1 {
    rewriter.simplify_ancestor_merge();
}
```

torquestomp (approved):
Thanks!

PR #4882, opened 2024-11-16
README: Remove the "build passing" badge

ilyagr:
It feels extraneous. People can just look at whether the GitHub UI shows a checkmark next to the commit.

This creates more space for the IRC badge.

This was originally part of #4881, but it might be simpler to merge it separately.


martinvonz (approved):
And the build should basically always be passing, so I agree that it doesn't add much value.

ilyagr:
Thanks!

PR #4886, opened 2024-11-16
cli: provide short flags for rev ranges

senekor:
Some commands already have short versions `-f` and `-t` for `--from` and `--to`. These include:
- `jj move` (deprecated)
- `jj squash`
- `jj absorb`

Another group of commands doesn't have the short flags yet and these would be relevant for dynamic completion of files I'm working on. I think it would be nice to provide the short flags here for consistency, unless there are reasons against it. 
- `jj diff`
- `jj interdiff`
- `jj restore`

Other commands are not relevant to my completions, but I could add them there as well if it's considered desirable:
- `jj diffedit`
- `jj op diff`

And lastly, `jj bookmark move` doesn't have the short flag, but there is an explicit reason documented why:
> We intentionally do not support the short `-f` for `--from` since it
> could be confused with a shorthand for `--force`, and people might not
> realize they need `-B`/`--allow-backwards` instead.

# Checklist

If applicable:


avamsi (approved):
Please update the CHANGELOG?

Also, you might've already seen it, but there's some discussion in https://github.com/martinvonz/jj/pull/4612 -- please wait a few days for any objections.

ilyagr:
> please wait a few days for any objections.

I added a label to indicate this. Feel free to remove it whenever it feels right (or now, if it doesn't seem right).

senekor:
Any opinions on adding the short flags to `diffedit` and `op diff` for consistency?

avamsi:
I'm okay with those too.

avamsi:
Sorry about the wait, @senekor -- there doesn't seem to be any objections, so feel free to sync and submit.

senekor:
No worries, it was good to wait.

PR #4888, opened 2024-11-16
log: describe the different graph node symbols in help text

martinvonz:
# Checklist

If applicable:


cli/src/commands/log.rs
```diff
@@ -47,6 +47,10 @@ use crate::ui::Ui;
 ///
 /// Spans of revisions that are not included in the graph per `--revisions` are
 /// rendered as a synthetic node labeled "(elided revisions)".
+///
+/// The working-copy commit is indicated by a `@` symbol in the graph. Immutable
```
bnjmnt4n, line 51: Should we also describe that these are the defaults and can be customized?

martinvonz: Can't hurt. Done.

PR #4891, opened 2024-11-16
completion: teach config about keys

senekor:
There are two known limitations right now:

- Only statically known keys are suggested.

- Keys that the user did not set are still suggested for `jj config get`. Running that suggestion may result in an error. The error message will be appropriate though and there is some value in letting the user know that this config value theoretically exists. Some users may try to explore what configurations are available via the completions.

# Checklist

If applicable:

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -80,6 +80,7 @@ rpassword = { workspace = true }
 sapling-renderdag = { workspace = true }
 scm-record = { workspace = true }
 serde = { workspace = true }
+serde_json = { workspace = true }
 slab = { workspace = true }
 strsim = { workspace = true }
 tempfile = { workspace = true }
--- a/cli/src/commands/config/get.rs
+++ b/cli/src/commands/config/get.rs
@@ -14,11 +14,13 @@
 
 use std::io::Write as _;
 
+use clap_complete::ArgValueCandidates;
 use tracing::instrument;
 
 use crate::cli_util::CommandHelper;
 use crate::command_error::config_error;
 use crate::command_error::CommandError;
+use crate::complete;
 use crate::config::ConfigNamePathBuf;
 use crate::ui::Ui;
 
@@ -34,7 +36,7 @@ use crate::ui::Ui;
 #[derive(clap::Args, Clone, Debug)]
 #[command(verbatim_doc_comment)]
 pub struct ConfigGetArgs {
-    #[arg(required = true)]
+    #[arg(required = true, add = ArgValueCandidates::new(complete::leaf_config_keys))]
     name: ConfigNamePathBuf,
 }
 
--- a/cli/src/commands/config/unset.rs
+++ b/cli/src/commands/config/unset.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use clap_complete::ArgValueCandidates;
 use tracing::instrument;
 
 use super::ConfigLevelArgs;
@@ -19,6 +20,7 @@ use crate::cli_util::get_new_config_file_path;
 use crate::cli_util::CommandHelper;
 use crate::command_error::user_error;
 use crate::command_error::CommandError;
+use crate::complete;
 use crate::config::remove_config_value_from_file;
 use crate::config::ConfigNamePathBuf;
 use crate::ui::Ui;
@@ -26,7 +28,7 @@ use crate::ui::Ui;
 /// Update config file to unset the given option.
 #[derive(clap::Args, Clone, Debug)]
 pub struct ConfigUnsetArgs {
-    #[arg(required = true)]
+    #[arg(required = true, add = ArgValueCandidates::new(complete::leaf_config_keys))]
     name: ConfigNamePathBuf,
     #[command(flatten)]
     level: ConfigLevelArgs,
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -26,7 +26,9 @@ use crate::cli_util::GlobalArgs;
 use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::config::default_config;
+use crate::config::ConfigNamePathBuf;
 use crate::config::LayeredConfigs;
+use crate::config::CONFIG_SCHEMA;
 use crate::ui::Ui;
 
 const BOOKMARK_HELP_TEMPLATE: &str = r#"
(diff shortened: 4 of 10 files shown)


cli/Cargo.toml
```diff
@@ -80,6 +80,7 @@ rpassword = { workspace = true }
 sapling-renderdag = { workspace = true }
 scm-record = { workspace = true }
 serde = { workspace = true }
+serde_json.workspace = true
```
yuja, line 83: nit: can you copy the current style `{ workspace = true }` for consistency? (or add a separate patch that updates everything.)

cli/src/commands/config/unset.rs
```diff
@@ -12,21 +12,23 @@
...
+use crate::complete;
 use crate::config::remove_config_value_from_file;
 use crate::config::ConfigNamePathBuf;
 use crate::ui::Ui;
 
 /// Update config file to unset the given option.
 #[derive(clap::Args, Clone, Debug)]
 pub struct ConfigUnsetArgs {
-    #[arg(required = true)]
+    #[arg(required = true, add = ArgValueCandidates::new(complete::config_keys))]
```
yuja, line 31: maybe this should also be a "leaf_config_keys"?

cli/src/complete.rs
```diff
@@ -271,6 +271,73 @@ pub fn operations() -> Vec<CompletionCandidate> {
...
+            continue;
+        }
+        let owned_key;
+        let key = if key.contains('(') {
+            // Parenthesis can occur in revset aliases, e.g. immutable_heads().
+            // These must be escaped.
+            owned_key = format!("'{key}'");
+            &owned_key
+        } else {
+            key
```
yuja, line 292: `ConfigNamePathBuf` can be used to escape TOML keys properly.

cli/src/complete.rs
```diff
@@ -271,6 +271,73 @@ pub fn operations() -> Vec<CompletionCandidate> {
...
+                    .get("description")
+                    .map(|desc| desc.as_str().unwrap().to_string().into());
+                acc.push(CompletionCandidate::new(format!("{prefix}{key}")).help(help));
+            }
+        }
+    }
+}
+
+fn config_keys_impl(only_leaves: bool) -> Vec<CompletionCandidate> {
+    let schema = include_str!("config-schema.json");
```
yuja, line 317: nit: better to extract the constant to the config module than including blob at two places.

cli/src/complete.rs
```diff
@@ -271,6 +271,73 @@ pub fn operations() -> Vec<CompletionCandidate> {
...
+
+pub fn config_keys() -> Vec<CompletionCandidate> {
+    config_keys_impl(false)
+}
+
+pub fn leaf_config_keys() -> Vec<CompletionCandidate> {
+    config_keys_impl(true)
+}
+
+#[test]
```
yuja, line 335: nit: add `mod tests` at the end of the file

cli/src/commands/config/get.rs
yuja, line 1: nit: missing snapshot tests?

cli/src/commands/config/unset.rs
senekor: I checked, and unset seems to work with objects as well. If I do:
```sh
jj config unset --user aliases
```
The entire `aliases` table is removed from my config. While it could be a bit of a footgun, the current behavior makes sense I think.

cli/Cargo.toml
senekor: Oops, I think I added it with `cargo add` and didn't pay attention.

cli/src/commands/config/unset.rs
yuja: Indeed. I believe it's a bug. The comment says "remove config value".

senekor: Alright, I'll change it to `leaf_config_keys` in that case. I'll leave fixing the bug to someone else :slightly_smiling_face:

cli/src/config_schema.rs
yuja, line 1: Can you add copyright header to this file?

I think it's also good to move this constant to config.rs

cli/src/config_schema.rs
```diff
@@ -0,0 +1,2 @@
+// TODO(#879): Consider generating entire schema dynamically vs. static file.
+pub static CONFIG_SCHEMA: &str = include_str!("config-schema.json");
```
yuja, line 2: nit: just use `const`?

cli/tests/test_completion.rs
```diff
@@ -400,3 +400,164 @@ fn test_operations() {
...
+    let mut test_env = TestEnvironment::default();
+    test_env.add_env_var("COMPLETE", "fish");
+    let dir = test_env.env_root();
+
+    let stdout = test_env.jj_cmd_success(dir, &["--", "jj", "config", "get", ""]);
+    insta::assert_snapshot!(stdout, @r#"
+    core.fsmonitor	Whether to use an external filesystem monitor, useful for large repos
+    core.watchman.register_snapshot_trigger	Whether to use triggers to monitor for changes in the background.
+    diff.color-words.context	Number of lines of context to show
+    diff.color-words.max-inline-alternation	Maximum number of removed/added word alternation to inline
```
yuja, line 415: Can you strip the output to e.g. 3-5 lines? We wouldn't want to update the snapshot every time we add new config key.

It might be also good to filter by `"get", "<some-prefix-char>"`.

cli/src/complete.rs
```diff
@@ -271,6 +273,67 @@ pub fn operations() -> Vec<CompletionCandidate> {
...
+        }
+        let mut prefix = prefix.clone();
+        prefix.push(key);
+
+        let value = value.as_object().unwrap();
+        match value.get("type").and_then(|v| v.as_str()) {
+            Some("object") => {
+                if !only_leaves {
+                    let escaped_key = prefix.to_string();
+                    acc.push(CompletionCandidate::new(escaped_key));
```
yuja, line 295: nit: we can add a hint to non-leaf keys?

cli/src/complete.rs
```diff
@@ -271,6 +273,67 @@ pub fn operations() -> Vec<CompletionCandidate> {
...
+    prefix: ConfigNamePathBuf,
+    properties: &serde_json::Map<String, serde_json::Value>,
+    acc: &mut Vec<CompletionCandidate>,
+    only_leaves: bool,
+) {
+    for (key, value) in properties {
+        if key == "type" {
+            // used by "additionalProperties", ignore
+            continue;
+        }
```
yuja, line 286: nit: do you remember which `"additionalProperties"` caused problems? If I understand it, only `"properties"` are scanned.

cli/src/complete.rs
```diff
@@ -271,6 +273,67 @@ pub fn operations() -> Vec<CompletionCandidate> {
...
+
+        let value = value.as_object().unwrap();
+        match value.get("type").and_then(|v| v.as_str()) {
+            Some("object") => {
+                if !only_leaves {
+                    let escaped_key = prefix.to_string();
+                    acc.push(CompletionCandidate::new(escaped_key));
+                }
+                let Some(properties) = value.get("properties") else {
+                    continue;
```
yuja, line 298: nit: might be better to suggest `<prefix>." if no child properties are defined. (can be added later)

yuja (approved):
Thanks!

cli/src/complete.rs
senekor: You're right, I think I was trying to scan "additionalProperties" too at some point, but decided against it. This can be removed.

cli/src/complete.rs
senekor: You mean with a trailing dot? These non-leaf keys are currently only suggested for `config list`, where omitting the dot makes sense.

It would make sense to suggest the trailing dot for the other commands, though. I agree that this can be added later.

yuja: > You mean with a trailing dot?

Yes. I think it's nice if `aliases.` (without trailing space) is suggested for `jj config set a<TAB>`, but I'm not sure if that's doable.

PR #4892, opened 2024-11-16
completion: teach git about remote urls

senekor:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/clone.rs
+++ b/cli/src/commands/git/clone.rs
@@ -19,6 +19,7 @@ use std::num::NonZeroU32;
 use std::path::Path;
 use std::path::PathBuf;
 
+use clap_complete::ArgValueCompleter;
 use jj_lib::git;
 use jj_lib::git::GitFetchError;
 use jj_lib::git::GitFetchStats;
@@ -33,6 +34,7 @@ use crate::command_error::user_error;
 use crate::command_error::user_error_with_message;
 use crate::command_error::CommandError;
 use crate::commands::git::maybe_add_gitignore;
+use crate::complete;
 use crate::config::write_config_value_to_file;
 use crate::config::ConfigNamePathBuf;
 use crate::git_util::get_git_repo;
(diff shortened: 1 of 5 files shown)


cli/src/commands/git/clone.rs
```diff
@@ -47,7 +49,7 @@ use crate::ui::Ui;
 #[derive(clap::Args, Clone, Debug)]
 pub struct GitCloneArgs {
     /// URL or path of the Git repo to clone
-    #[arg(value_hint = clap::ValueHint::DirPath)]
+    #[arg(add = ArgValueCompleter::new(complete::new_git_remote_url))]
```
senekor, line 52: This is a little unfortunate. I tried to keep both, but that wouldn't work, the directory path value hint would just be ingored.

This means that file paths are not completed anymore with this change, a slight regression. The only way I see to fix this is to reimplement the logic in our own value completer. (or maybe ask clap to enable this use case)

cli/src/complete.rs
```diff
@@ -193,6 +193,62 @@ pub fn git_remotes() -> Vec<CompletionCandidate> {
...
+    if current.contains(domain_terminator) {
+        // Completing the path after the domain as well would be cool, but it's
+        // more difficult. One approach could be to cache the URLs that were
+        // cloned previously. Users probably clone repositories from the same
+        // user / organisations frequently.
+        return Vec::new();
+    }
+
+    let Some(known_hosts) = dirs::home_dir()
+        .and_then(|home| std::fs::read_to_string(home.join(".ssh/known_hosts")).ok())
```
senekor, line 232: I had the idea to scrape aliases in `~/.ssh/config` as well. I might add that to this PR later if I find the time.

senekor:
I don't know why it doesn't work on windows. I suppose it has something to do with paths? Maybe the home directory discovery? The test environment sets the `HOME` variable, maybe that is not detected properly?

cli/src/complete.rs
senekor: I guess known_hosts should be sufficient. The difference is only typing the url manually about once per new computer setup.

senekor:
don't wanna deal with windows

PR #4895, opened 2024-11-17
completion: suggest file paths incrementally

senekor:
stacked on #4887 (teach commands about files)

If there are multiple files in a subdirectory that are candidates for completion, only complete the common directory prefix to reduce the number of completion candidates shown at once.

This matches the normal shell completion of file paths.

# Checklist

If applicable:

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -448,7 +448,23 @@ pub fn leaf_config_keys() -> Vec<CompletionCandidate> {
     config_keys_impl(true)
 }
 
-fn all_files_from_rev(rev: String) -> Vec<CompletionCandidate> {
+fn dir_prefix_from<'a>(path: &'a str, current: &str) -> Option<&'a str> {
+    path[current.len()..]
+        .split_once(std::path::MAIN_SEPARATOR)
+        .map(|(next, _)| path.split_at(current.len() + next.len() + 1).0)
+}
+
+fn current_prefix_to_fileset(current: &str) -> String {
+    let cur_esc = glob::Pattern::escape(current);
+    let dir_pat = format!("{cur_esc}*/**");
+    let path_pat = format!("{cur_esc}*");
+    format!("glob:{dir_pat:?} | glob:{path_pat:?}")
+}
+
+fn all_files_from_rev(rev: String, current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
+    let Some(current) = current.to_str() else {
+        return Vec::new();
+    };
     with_jj(|jj, _| {
         let mut child = jj
             .build()
@@ -456,6 +472,9 @@ fn all_files_from_rev(rev: String) -> Vec<CompletionCandidate> {
             .arg("list")
             .arg("--revision")
             .arg(rev)
+            .arg("--config-toml")
+            .arg("ui.allow-filesets = true")
+            .arg(current_prefix_to_fileset(current))
             .stdout(std::process::Stdio::piped())
             .spawn()
             .map_err(user_error)?;
@@ -465,7 +484,13 @@ fn all_files_from_rev(rev: String) -> Vec<CompletionCandidate> {
             .lines()
             .take(1_000)
             .map_while(Result::ok)
-            .map(CompletionCandidate::new)
+            .map(|path| {
+                if let Some(dir_path) = dir_prefix_from(&path, current) {
+                    return CompletionCandidate::new(dir_path);
+                }
+                CompletionCandidate::new(path)
+            })
+            .dedup() // directories may occur multiple times
             .collect())
     })
 }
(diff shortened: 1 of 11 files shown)


cli/src/complete.rs
```diff
@@ -407,14 +416,31 @@ fn all_files_from_rev(rev: String) -> Vec<CompletionCandidate> {
             .map_err(user_error)?;
         let stdout = String::from_utf8_lossy(&output.stdout);
 
-        Ok(stdout.lines().map(CompletionCandidate::new).collect())
+        Ok(stdout
+            .lines()
+            .filter_map(|path| {
+                if !path.starts_with(current) {
```
martinvonz, line 422: Instead of filtering afterwards here, we should pass the current path as an argument to `jj file list`. That way we can avoid walking parts of the tree that doesn't match. We also need to give `jj file list` (or `jj debug tree`?) an option to also list directory entries so we don't have to list all files in subdirectories. But that can be fixed later.

cli/src/complete.rs
```diff
@@ -396,7 +396,16 @@ mod parse {
...
 }
 
-fn all_files_from_rev(rev: String) -> Vec<CompletionCandidate> {
+fn dir_prefix_from<'a>(path: &'a str, current: &str) -> Option<&'a str> {
+    path.strip_prefix(current)?
+        .split_once(std::path::MAIN_SEPARATOR)
+        .map(|(next, _)| path.split_at(current.len() + next.len() + 1).0)
+}
+
+fn all_files_from_rev(rev: String, current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
```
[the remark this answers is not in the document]
senekor: This has been addressed in https://github.com/martinvonz/jj/pull/4887

cli/src/complete.rs
senekor: There is a small problem with passing `current` as an argument to `jj file list`. It only works properly if `current` is a correct directory path, not a partial one. For example, in the jj repo:

- `jj file list cli` returns all files in `cli/` as expected
- `jj file list cl` doesn't do that. It doesn't return any files, because there is no directory called `cl/`.

Do you think I should implement some logic to figure out if `current` is a correct directory path and can be passed to `jj file list` as-is? Or if it needs to be truncated to the next parent or something?

yuja: I think you can turn the path into glob pattern, something like `glob:{escape(path)}*/**`.

cli/src/complete.rs
```diff
@@ -448,14 +448,27 @@ pub fn leaf_config_keys() -> Vec<CompletionCandidate> {
     config_keys_impl(true)
 }
 
-fn all_files_from_rev(rev: String) -> Vec<CompletionCandidate> {
+fn dir_prefix_from<'a>(path: &'a str, current: &str) -> Option<&'a str> {
+    path.strip_prefix(current)?
```
yuja, line 452: nit: just find separator from `current.len()`? Callers should exclude paths not under the current path.

cli/src/complete.rs
```diff
@@ -448,14 +448,27 @@ pub fn leaf_config_keys() -> Vec<CompletionCandidate> {
...
     with_jj(|jj, _| {
         let mut child = jj
             .build()
             .arg("file")
             .arg("list")
             .arg("--revision")
             .arg(rev)
+            .arg("--config-toml")
+            .arg("ui.allow-filesets = true")
+            .arg(format!(r#"glob:"{cur_esc}*/**" | glob:"{cur_esc}*""#))
```
yuja, line 471: nit: might be a bit safer to abuse string debug format to escape double quotes in `cur_esc`: `glob:{pat:?}`

cli/src/complete.rs
```diff
@@ -484,10 +507,18 @@ fn modified_files_from_rev_with_jj_cmd(
...
     Ok(stdout
         .lines()
-        .map(|line| {
+        .filter_map(|line| {
             let (mode, path) = line
                 .split_once(' ')
                 .expect("diff --summary should contain a space between mode and path");
+
+            if !path.starts_with(current) {
+                return None;
```
yuja, line 516: nit: specify glob pattern instead of post-filtering?

cli/src/complete.rs
```diff
@@ -519,62 +557,74 @@ fn conflicted_files_from_rev(rev: &str) -> Vec<CompletionCandidate> {
 
         Ok(stdout
             .lines()
-            .filter_map(|line| line.split_whitespace().next())
-            .map(CompletionCandidate::new)
+            .filter_map(|line| {
+                let path = line.split_whitespace().next()?;
+
+                if !path.starts_with(current) {
```
yuja, line 563: nit: same here. `resolve --list` appears to accept path patterns

yuja (approved):
Thanks.

PR #4896, opened 2024-11-17
cli: fix "config unset" to not delete a whole table

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -669,9 +669,17 @@ pub fn remove_config_value_from_file(
     })?;
 
     // Remove config value
-    target_table
-        .remove(last_key)
-        .ok_or_else(|| config::ConfigError::NotFound(key.to_string()))?;
+    match target_table.entry(last_key) {
+        toml_edit::Entry::Occupied(entry) => {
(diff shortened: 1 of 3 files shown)


cli/src/config.rs
```diff
@@ -669,9 +669,17 @@ pub fn remove_config_value_from_file(
     })?;
 
     // Remove config value
-    target_table
-        .remove(last_key)
-        .ok_or_else(|| config::ConfigError::NotFound(key.to_string()))?;
+    match target_table.entry(last_key) {
+        toml_edit::Entry::Occupied(entry) => {
+            if entry.get().is_table() {
+                return Err(user_error(format!("Cannot remove {key}: is a table")));
```
martinvonz, line 675: I think "Cannot remove {key} because it is a table" sounds more natural. Or maybe "Won't remove non-inline table {key}" to try to clarify that it's not for lack of support but a concious decision to not remove it

martinvonz (approved):

cli/src/config.rs
yuja: Updated to "Won't remove table {key}".

Issue #3975, opened 2024-06-26
FR: Materialize files with conflict markers (or any files) in a way that can be parsed


ilyagr:
This is a tracking issue for this bug/FR. I'm promoting this discussion from https://github.com/martinvonz/jj/issues/3968#issuecomment-2192516971. I think I also have a plan now that could work, so I am posting an outline it here in case people have comments.

The alternative would be to declare that it is not a goal to be able to materialize any text file conflict, and show a materialization error instead. I was leaning towards this, but I think the following approach is simple enough to work.

-------------------------

-------------------------

_Originally posted by @ilyagr in https://github.com/martinvonz/jj/issues/3968#issuecomment-2192516971_

> Wouldn't you want to preserve the invariant that you can reconstruct any of the file sides directly from the conflicted contents, rather than having to consult an out-of-band store?

This issue has been nerd-sniping me for a while; in addition to trailing newlines (**Update:** in non-empty files; we actually support files that are 0 or more lines, each ending with a newline), it's also a matter of encoding conflict markers inside files 

------

TODO: find link. I remember discussing it with Martin recently in some PR, haven't found it yet. TLDR: `jj` currently can't handle conflicts in out own `tutorial.md` because of

https://github.com/martinvonz/jj/blob/638649b435795bbb991d840e9ad1750648d9b74a/docs/tutorial.md?plain=1#L284-L291

----

In the shortest term, I'll probably ignore both issues and just add newlines to the end of files.

However, I think I finally figured out a syntax that will work, will not require a complicated parser, and even be somewhat readable. Before computing conflicts (and only if a file is conflicted), we'd encode file contents as follows:

1. A mildly pathological file

```
JJ Verbatim Line:<<<<<<<<<<<
blahblah
JJ EOF: No newline at the end of file
```

2. A moderately pathological file

```
JJ Verbatim Line:<<<<<<<<<<<
JJ Verbatim Line:JJ Verbatim Line:<<<<<<<<<<<
blahblah
JJ Verbatim Line:JJ EOF: No newline at the end of file
```

3. Another fun file

```
blah
JJ Verbatim Line:JJ EOF: No newline at the end of file
JJ EOF: No newline at the end of file
```

Fun, huh?

-------

[...]

PR #4897, opened 2024-11-17
conflicts: add "ui.conflict-marker-style" config

### LINKED ISSUE #823: Allow conflicts to be materialized in Git style
## Description

This improves interop with tools that only understand Git-style conflict markers, e.g. IDEs, `npm`.

We would probably make a config option for this, but it might also be nice to have a command that explicitly re-materializes the conflicted file (e.g. a user who prefers jj-style conflict markers and occasionally needs Git-style conflict markers for tooling).

scott2000:
Resolves #823.

Adds support for "diff" (current default), "snapshot", and "git-diff3" (falls back to "diff" if more than 2 sides) options.

**Example of "diff"**

```
<<<<<<< Conflict 1 of 1
+++++++ Contents of side #1
fn example(word: String) {
    println!("word is {word}");
%%%%%%% Changes from base to side #2
-fn example(w: String) {
+fn example(w: &str) {
     println!("word is {w}");
>>>>>>> Conflict 1 of 1 ends
}
```

**Example of "snapshot"**

```
<<<<<<< Conflict 1 of 1
+++++++ Contents of side #1
fn example(word: String) {
    println!("word is {word}");
------- Contents of base
fn example(w: String) {
    println!("word is {w}");
+++++++ Contents of side #2
fn example(w: &str) {
    println!("word is {w}");
>>>>>>> Conflict 1 of 1 ends
}
```

**Example of "git"**

```
<<<<<<< Side #1 (Conflict 1 of 1)
fn example(word: String) {
    println!("word is {word}");
||||||| Base
fn example(w: String) {
    println!("word is {w}");
=======
fn example(w: &str) {
    println!("word is {w}");
>>>>>>> Side #2 (Conflict 1 of 1 ends)
}
```

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   materialized in the working copy. The default option ("diff") renders
   conflicts as a snapshot with a list of diffs to apply to the snapshot.
   The new "snapshot" option renders conflicts as a series of snapshots, showing
-  each side and base of the conflict.
+  each side and base of the conflict. The new "git" option replicates Git's
+  "diff3" conflict style, meaning it is more likely to work with external tools,
+  but it doesn't support conflicts with more than 2 sides.
 
 ### Fixed bugs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New `ui.conflict-marker-style` config option to change how conflicts are
   materialized in the working copy. The default option ("diff") renders
   conflicts as a snapshot with a list of diffs to apply to the snapshot.
+  The new "snapshot" option renders conflicts as a series of snapshots, showing
+  each side and base of the conflict.
 
 ### Fixed bugs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   - `op diff`
   - `restore`
 
+* New `ui.conflict-marker-style` config option to change how conflicts are
+  materialized in the working copy. The default option ("diff") renders
+  conflicts as a snapshot with a list of diffs to apply to the snapshot.
+
 ### Fixed bugs
 
 * `jj config unset <TABLE-NAME>` no longer removes a table (such as `[ui]`.)
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -158,6 +158,14 @@
                 "merge-editor": {
                     "type": "string",
                     "description": "Tool to use for resolving three-way merges. Behavior for a given tool name can be configured in merge-tools.TOOL tables"
+                },
+                "conflict-marker-style": {
+                    "type": "string",
+                    "description": "Conflict marker style to use when materializing conflicts in the working copy",
+                    "enum": [
+                        "diff"
+                    ],
+                    "default": "diff"
                 }
             }
         },
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -163,7 +163,8 @@
                     "type": "string",
                     "description": "Conflict marker style to use when materializing conflicts in the working copy",
                     "enum": [
-                        "diff"
+                        "diff",
+                        "snapshot"
                     ],
                     "default": "diff"
                 }
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -164,7 +164,8 @@
                     "description": "Conflict marker style to use when materializing conflicts in the working copy",
                     "enum": [
                         "diff",
-                        "snapshot"
+                        "snapshot",
+                        "git"
                     ],
                     "default": "diff"
                 }
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -35,6 +35,7 @@ use jj_lib::commit::Commit;
 use jj_lib::config::ConfigError;
 use jj_lib::conflicts::materialize_merge_result_to_bytes;
 use jj_lib::conflicts::materialized_diff_stream;
+use jj_lib::conflicts::ConflictMarkerStyle;
 use jj_lib::conflicts::MaterializedTreeDiffEntry;
 use jj_lib::conflicts::MaterializedTreeValue;
 use jj_lib::copies::CopiesTreeDiffEntry;
@@ -267,6 +268,7 @@ pub enum DiffRenderError {
 pub struct DiffRenderer<'a> {
     repo: &'a dyn Repo,
     path_converter: &'a RepoPathUiConverter,
+    conflict_marker_style: ConflictMarkerStyle,
     formats: Vec<DiffFormat>,
 }
 
@@ -274,12 +276,14 @@ impl<'a> DiffRenderer<'a> {
     pub fn new(
         repo: &'a dyn Repo,
         path_converter: &'a RepoPathUiConverter,
+        conflict_marker_style: ConflictMarkerStyle,
         formats: Vec<DiffFormat>,
     ) -> Self {
         DiffRenderer {
             repo,
-            formats,
             path_converter,
+            conflict_marker_style,
+            formats,
         }
     }
 
@@ -331,7 +335,15 @@ impl<'a> DiffRenderer<'a> {
                 DiffFormat::Stat(options) => {
                     let tree_diff =
                         from_tree.diff_stream_with_copies(to_tree, matcher, copy_records);
-                    show_diff_stat(formatter, store, tree_diff, path_converter, options, width)?;
+                    show_diff_stat(
+                        formatter,
+                        store,
+                        tree_diff,
+                        path_converter,
+                        options,
+                        width,
+                        self.conflict_marker_style,
+                    )?;
                 }
                 DiffFormat::Types => {
                     let tree_diff =
@@ -346,12 +358,25 @@ impl<'a> DiffRenderer<'a> {
                 DiffFormat::Git(options) => {
                     let tree_diff =
                         from_tree.diff_stream_with_copies(to_tree, matcher, copy_records);
-                    show_git_diff(formatter, store, tree_diff, options)?;
+                    show_git_diff(
+                        formatter,
+                        store,
+                        tree_diff,
+                        options,
+                        self.conflict_marker_style,
+                    )?;
                 }
                 DiffFormat::ColorWords(options) => {
                     let tree_diff =
                         from_tree.diff_stream_with_copies(to_tree, matcher, copy_records);
-                    show_color_words_diff(formatter, store, tree_diff, path_converter, options)?;
+                    show_color_words_diff(
+                        formatter,
+                        store,
+                        tree_diff,
+                        path_converter,
+                        options,
+                        self.conflict_marker_style,
+                    )?;
                 }
                 DiffFormat::Tool(tool) => {
                     match tool.diff_invocation_mode {
@@ -365,12 +390,21 @@ impl<'a> DiffRenderer<'a> {
                                 tree_diff,
                                 path_converter,
                                 tool,
+                                self.conflict_marker_style,
                             )
                         }
                         DiffToolMode::Dir => {
                             let mut writer = formatter.raw()?;
-                            generate_diff(ui, writer.as_mut(), from_tree, to_tree, matcher, tool)
-                                .map_err(DiffRenderError::DiffGenerate)
+                            generate_diff(
+                                ui,
+                                writer.as_mut(),
+                                from_tree,
+                                to_tree,
+                                matcher,
+                                tool,
+                                self.conflict_marker_style,
+                            )
+                            .map_err(DiffRenderError::DiffGenerate)
                         }
                     }?;
                 }
@@ -837,7 +871,11 @@ fn file_content_for_diff(reader: &mut dyn io::Read) -> io::Result<FileContent> {
     })
 }
 
-fn diff_content(path: &RepoPath, value: MaterializedTreeValue) -> io::Result<FileContent> {
+fn diff_content(
+    path: &RepoPath,
+    value: MaterializedTreeValue,
+    conflict_marker_style: ConflictMarkerStyle,
+) -> io::Result<FileContent> {
     match value {
         MaterializedTreeValue::Absent => Ok(FileContent::empty()),
         MaterializedTreeValue::AccessDenied(err) => Ok(FileContent {
@@ -863,7 +901,7 @@ fn diff_content(path: &RepoPath, value: MaterializedTreeValue) -> io::Result<Fil
             executable: _,
         } => Ok(FileContent {
             is_binary: false,
-            contents: materialize_merge_result_to_bytes(&contents).into(),
+            contents: materialize_merge_result_to_bytes(&contents, conflict_marker_style).into(),
         }),
         MaterializedTreeValue::OtherConflict { id } => Ok(FileContent {
             is_binary: false,
@@ -902,6 +940,7 @@ pub fn show_color_words_diff(
     tree_diff: BoxStream<CopiesTreeDiffEntry>,
     path_converter: &RepoPathUiConverter,
     options: &ColorWordsDiffOptions,
+    conflict_marker_style: ConflictMarkerStyle,
 ) -> Result<(), DiffRenderError> {
     let mut diff_stream = materialized_diff_stream(store, tree_diff);
     async {
@@ -937,7 +976,7 @@ pub fn show_color_words_diff(
                     formatter.labeled("header"),
                     "Added {description} {right_ui_path}:"
                 )?;
-                let right_content = diff_content(right_path, right_value)?;
+                let right_content = diff_content(right_path, right_value, conflict_marker_style)?;
                 if right_content.is_empty() {
                     writeln!(formatter.labeled("empty"), "    (empty)")?;
                 } else if right_content.is_binary {
@@ -999,8 +1038,8 @@ pub fn show_color_words_diff(
                         )
                     }
                 };
-                let left_content = diff_content(left_path, left_value)?;
-                let right_content = diff_content(right_path, right_value)?;
+                let left_content = diff_content(left_path, left_value, conflict_marker_style)?;
+                let right_content = diff_content(right_path, right_value, conflict_marker_style)?;
                 if left_path == right_path {
                     writeln!(
                         formatter.labeled("header"),
@@ -1028,7 +1067,7 @@ pub fn show_color_words_diff(
                     formatter.labeled("header"),
                     "Removed {description} {right_ui_path}:"
                 )?;
-                let left_content = diff_content(left_path, left_value)?;
+                let left_content = diff_content(left_path, left_value, conflict_marker_style)?;
                 if left_content.is_empty() {
                     writeln!(formatter.labeled("empty"), "    (empty)")?;
                 } else if left_content.is_binary {
@@ -1050,18 +1089,18 @@ pub fn show_file_by_file_diff(
     tree_diff: BoxStream<CopiesTreeDiffEntry>,
     path_converter: &RepoPathUiConverter,
     tool: &ExternalMergeTool,
+    conflict_marker_style: ConflictMarkerStyle,
 ) -> Result<(), DiffRenderError> {
-    fn create_file(
-        path: &RepoPath,
-        wc_dir: &Path,
-        value: MaterializedTreeValue,
-    ) -> Result<PathBuf, DiffRenderError> {
+    let create_file = |path: &RepoPath,
+                       wc_dir: &Path,
+                       value: MaterializedTreeValue|
+     -> Result<PathBuf, DiffRenderError> {
         let fs_path = path.to_fs_path(wc_dir)?;
         std::fs::create_dir_all(fs_path.parent().unwrap())?;
-        let content = diff_content(path, value)?;
+        let content = diff_content(path, value, conflict_marker_style)?;
         std::fs::write(&fs_path, content.contents)?;
         Ok(fs_path)
-    }
+    };
 
     let temp_dir = new_utf8_temp_dir("jj-diff-")?;
     let left_wc_dir = temp_dir.path().join("left");
@@ -1124,6 +1163,7 @@ struct GitDiffPart {
 fn git_diff_part(
     path: &RepoPath,
     value: MaterializedTreeValue,
+    conflict_marker_style: ConflictMarkerStyle,
 ) -> Result<GitDiffPart, DiffRenderError> {
     const DUMMY_HASH: &str = "0000000000";
     let mode;
@@ -1176,7 +1216,8 @@ fn git_diff_part(
             hash = DUMMY_HASH.to_owned();
             content = FileContent {
                 is_binary: false, // TODO: are we sure this is never binary?
-                contents: materialize_merge_result_to_bytes(&contents).into(),
+                contents: materialize_merge_result_to_bytes(&contents, conflict_marker_style)
+                    .into(),
             };
         }
         MaterializedTreeValue::OtherConflict { id } => {
@@ -1440,6 +1481,7 @@ pub fn show_git_diff(
     store: &Store,
     tree_diff: BoxStream<CopiesTreeDiffEntry>,
     options: &UnifiedDiffOptions,
+    conflict_marker_style: ConflictMarkerStyle,
 ) -> Result<(), DiffRenderError> {
     let mut diff_stream = materialized_diff_stream(store, tree_diff);
     async {
@@ -1450,8 +1492,8 @@ pub fn show_git_diff(
             let right_path_string = right_path.as_internal_file_string();
             let (left_value, right_value) = values?;
 
-            let left_part = git_diff_part(left_path, left_value)?;
-            let right_part = git_diff_part(right_path, right_value)?;
+            let left_part = git_diff_part(left_path, left_value, conflict_marker_style)?;
+            let right_part = git_diff_part(right_path, right_value, conflict_marker_style)?;
 
             formatter.with_label("file_header", |formatter| {
                 writeln!(
@@ -1625,6 +1667,7 @@ pub fn show_diff_stat(
     path_converter: &RepoPathUiConverter,
     options: &DiffStatOptions,
     display_width: usize,
+    conflict_marker_style: ConflictMarkerStyle,
 ) -> Result<(), DiffRenderError> {
     let mut stats: Vec<DiffStat> = vec![];
     let mut unresolved_renames = HashSet::new();
@@ -1637,8 +1680,8 @@ pub fn show_diff_stat(
             let (left, right) = values?;
             let left_path = path.source();
             let right_path = path.target();
-            let left_content = diff_content(left_path, left)?;
-            let right_content = diff_content(right_path, right)?;
+            let left_content = diff_content(left_path, left, conflict_marker_style)?;
+            let right_content = diff_content(right_path, right, conflict_marker_style)?;
 
             let left_ui_path = path_converter.format_file_path(left_path);
             let path = if left_path == right_path {
--- a/docs/conflicts.md
+++ b/docs/conflicts.md
@@ -3,6 +3,12 @@
 
 ## Introduction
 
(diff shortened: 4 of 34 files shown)


tim-janik:
> **Example of "snapshot"**

+1 for supporting the "snapshot" markers.

lib/src/conflicts.rs
```diff
@@ -269,9 +270,9 @@ pub fn materialize_merge_result(
                         // The vast majority of conflicts one actually tries to
                         // resolve manually have 1 base.
                         let base_str = if hunk.removes().len() == 1 {
-                            "base".to_string()
+                            Cow::Borrowed("base")
                         } else {
-                            format!("base #{}", base_index + 1)
+                            Cow::Owned(format!("base #{}", base_index + 1))
```
martinvonz: This feels like a premature optimization to me. Does it make a measurable difference?

CHANGELOG.md
```diff
@@ -41,6 +41,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New `fork_point()` revset function can be used to obtain the fork point
   of multiple commits.
 
+* New `ui.conflict-marker-style` config option to change how conflicts are
+  materialized in the working copy. The default option ("diff") renders
+  conflicts as a snapshot with a list of diffs to apply to the snapshot.
```
martinvonz, line 46: nit: don't forget to update `config-schema.json`

lib/src/conflicts.rs
```diff
@@ -141,6 +141,8 @@ pub async fn extract_as_single_hunk(
 pub enum ConflictMarkerStyle {
     /// Style which shows a snapshot and a series of diffs to apply
     Diff,
+    /// Style which only shows a snapshot for each base and side
```
martinvonz, line 148: nit: drop "only"

CHANGELOG.md
```diff
@@ -45,7 +45,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   materialized in the working copy. The default option ("diff") renders
   conflicts as a snapshot with a list of diffs to apply to the snapshot.
   The new "snapshot" option just renders conflicts as a series of snapshots,
-  showing each side of the conflict and the base(s).
+  showing each side and base of the conflict. The new "git-diff3" option
+  replicates Git's "diff3" conflict style, meaning it is more likely to work
+  with external tools, but it doesn't support conflicts with more than 2 sides.
```
martinvonz, line 50: We could use the ancestor line for each base and the separator line for each side. I don't know if that would be more or less surprising for users and tools.

martinvonz (commented):
I haven't reviewd the last two commits, but I've read the commit message for them and it sounds like you didn't miss anything (in particular, it would have been easy to miss that we want support for parsing either style regardless of configured style).

lib/src/conflicts.rs
scott2000: No, I don't think it makes a big difference, so I'll remove it.

CHANGELOG.md
scott2000: I was thinking it would be easier to implement the basic version first, and then if support for more than 2 sides is required, it could be added in a later version. One reason I think the current behavior might be more desirable is that it seems very difficult to manually resolve many-sided conflicts just by looking at the sides/bases, so it seems like providing diffs in that case would be more helpful, but I'm not sure.

martinvonz:
Let me know when this is ready for review. (I don't think I get a notification when a PR is marked ready.)

cli/src/diff_util.rs
```diff
@@ -296,6 +297,7 @@ impl<'a> DiffRenderer<'a> {
         matcher: &dyn Matcher,
         copy_records: &CopyRecords,
         width: usize,
+        conflict_marker_style: ConflictMarkerStyle,
```
yuja, line 300: nit: Perhaps, the style can be captured by `DiffRenderer::new()` for now.

If we add native support for conflict diffs, the `conflict_marker_style` might be moved to `DiffFormat` variants. And `[diff.color-words]` might have `materialize-conflict = true|false` and `conflict-marker-style = <style>`, for example.

scott2000:
Ok, I think it's ready for review now! Let me know if I should split some of these commits into separate PRs.

cli/src/config-schema.json
```diff
@@ -164,7 +164,8 @@
                     "description": "Conflict marker style to use when materializing conflicts in the working copy",
                     "enum": [
                         "diff",
-                        "snapshot"
+                        "snapshot",
```
ilyagr: (For discussion) For the snapshot style, we could also do something more similar to `git` style, something like

```
<<<<<<< Side 1 (Conflict 1 of 1 begins)
fn example(word: String) {
    println!("word is {word}");
||||||| Base
fn example(w: String) {
    println!("word is {w}");
======= Side 2
fn example(w: &str) {
    println!("word is {w}");
>>>>>>> Conflict 1 of 1 ends
}
```

One downside of this is that, beyond the number of sides (which could be larger than 2 here) and the placement of comments, there would be another subtle difference between this style and git's style once we fix https://github.com/martinvonz/jj/issues/3975, since the separators could get longer.

If we go with this, I'd go with a different name, e.g. "git-extended".

cli/src/config-schema.json
ilyagr, line 1: (Probably no action needed) I was thinking about how this PR relates to https://github.com/martinvonz/jj/issues/3975. Making parsing more flexible means that more kinds of text would need special encoding, but I *think* (but am not sure) that the solution we are discussing there would still work.

The approach we seemed to settle on in that issue would make the conflict markers longer if the original text of the file contains conflict markers.

cli/src/config-schema.json
```diff
@@ -164,7 +164,8 @@
...
+                        "git-diff3"
```
ilyagr: (Bikeshedding) For the name, I'd prefer if "git" and "diff3" were synonyms. Either is easier to remember that "git-diff3".

cli/src/config-schema.json
```diff
@@ -164,7 +164,8 @@
...
                     ],
```
ilyagr: (For discussion) I checked, and the `diff3` format is not meant for being perfectly decodable in the sense of #3975. Git and `diff3` don't do anything special if the text contains conflict markers, and generate text that cannot be parsed unambiguously.

I wonder whether, in these cases we should fall back to other conflict styles or fail to materialize (it could, for example, output to something similar as what `jj` does when symlinks conflict). We *probably* should do the same thing in this case as in the case when there are too many sides.

It's fine to just do a fallback for now and maybe implement something else later (e.g. a "git-strict" style that fails on error), as long as we have a sense that it won't be a hopeless mess.

ilyagr (commented):
This is a clever approach, it's simpler than what I was thinking (allowing other conflict styles only in a `jj file edit` command). I was worried for a bit that this would make #3975 worse, but I currently think that it's probably manageable and might be worth the complexity. I wrote down some related thoughts in case they spark anybody else's thinking.

cli/src/config-schema.json
scott2000: I personally think it would be nice to keep the "diff" and "snapshot" styles as similar as possible, because I feel like it gives a nice sense of consistency. In my mind, "diff" and "snapshot" would be the preferred Jujutsu-style conflict markers, and any other conflict marker styles would be added mainly for compatibility with other tools.

On the other hand, similarity to Git could be good for new users because it's less for them to learn.

cli/src/config-schema.json
scott2000: Maybe we could just have it be called "git", since it seems unlikely we'd support anything other than "diff3" from Git. If we were to use "diff3" as-is, we'd probably want to rename "diff" to something else to avoid confusion since they're pretty different.

ilyagr: We could also consider renaming "diff" to "default" or "jj" or "side-and-diffs".  Git supports several merge styles, so I think supporting "diff3" is good if we can (the [other ones](https://git-scm.com/docs/git-merge#_how_conflicts_are_presented) are either lossy or confusing, we shouldn't support them, but they do exist). Supporting "git" as a synonym makes sense to me as well, though I could also do without it.

I'd love it if we came up with better names before merging this, it might be hard to change later. It is, however, tricky to come up with something better. If we don't have a better idea, we should probably just merge instead of bikeshedding forever.

scott2000: I suppose another argument in favor of calling it "diff3" rather than "git" is because those conflict markers originally come from the "diff3" program (as far as I know).

Right now, I think I like "default" more than "jj" because calling it "jj" seems to imply that the other conflict marker styles aren't Jujutsu, even though "snapshot" is also Jujutsu-specific currently.

If we replaced "snapshot" with an extended version of "diff3", then we could probably get away with having "jj", "diff3", and "diff3+" as you suggested in your other comment.

lib/src/local_working_copy.rs
```diff
@@ -335,6 +336,7 @@ pub struct TreeState {
     sparse_patterns: Vec<RepoPathBuf>,
     own_mtime: MillisSinceEpoch,
     symlink_support: bool,
+    conflict_marker_style: ConflictMarkerStyle,
```
yuja, line 339: Perhaps, this can be passed as `CheckOutOptions { conflict_marker_style }`?

It's a bit weird that we have to pass the same value to `snapshot()` (to reconstruct the old file content), but I think it can be simplified later. For example, tree state could remember a hash of conflict file content for comparison.

lib/src/conflicts.rs
yuja, line 1: I have a similar patch locally. https://github.com/martinvonz/jj/compare/main...yuja:jj:push-puzorutuyrzw

If you're okay with my change, I'll send it.

lib/src/conflicts.rs
```diff
@@ -339,6 +316,33 @@ pub fn materialize_merge_result(
     Ok(())
 }
 
+/// Write a positive snapshot (side) of a conflict
+fn write_side(add_index: usize, data: &BString, output: &mut dyn Write) -> std::io::Result<()> {
```
yuja, line 373: nit: `&BStr` or `&[u8]` instead of `&BString`

lib/src/conflicts.rs
scott2000: Looks good to me!

cli/src/config-schema.json
martinvonz: I agree that we should keep them as similar as possible.

cli/src/config-schema.json
martinvonz: "snapshot", "diff", and "git" sound good to, but I don't feel strongly

cli/src/config-schema.json
ilyagr: Do you mean `snapshot` and the current jj style or `snapshot` and `git`? I'm guessing it's the first one, and I think I'm also leaning towards it, though I'm not sure.

martinvonz: Ah, yes, that's what I mean.

I think new tools should not care too much about being similar to old tools. That just benefits existing users of the old tool. It should not be at the expense of what we think is best for new users. I think we should at least hope to have more future users than migrating users in the long run (though that's a pretty ambitious when we're talking about a tool with ~98% market or whatever it is Git has).

cli/src/config-schema.json
scott2000: > "snapshot", "diff", and "git" sound good to, but I don't feel strongly

I think this is my current favorite option. After thinking about it a bit more, I don't love "diff3" as a name because it sounds like it should show a diff, but it doesn't actually show any diffs, which is confusing. Since we have a format that actually does show diffs, it seems like users could get confused about which is which.

lib/src/conflicts.rs
```diff
@@ -302,6 +312,26 @@ fn materialize_conflict_hunks(
     Ok(())
 }
 
+fn materialize_git_diff3_conflict(
```
scott2000, line 315: Right now, it's a bit unusual that the "git"/"diff3" style labels the "base" section but not either of the sides' sections, so I'm wondering if instead of this format:

```
<<<<<<< Conflict 1 of 1
fn example(word: String) {
    println!("word is {word}");
||||||| Contents of base
fn example(w: String) {
    println!("word is {w}");
=======
fn example(w: &str) {
    println!("word is {w}");
>>>>>>> Conflict 1 of 1 ends
}
```

We should have this format (inspired by @ilyagr's comment):

```
<<<<<<< Side #1 (Conflict 1 of 1)
fn example(word: String) {
    println!("word is {word}");
||||||| Base
fn example(w: String) {
    println!("word is {w}");
=======
fn example(w: &str) {
    println!("word is {w}");
>>>>>>> Side #2 (Conflict 1 of 1 ends)
}
```

Thoughts?

martinvonz: Weak vote for the second format.

Btw, if we plan to support more than 2 sides, then the second side will have to go on the `=======` line instead of on the `>>>>>>>` line (yes, different from Git). I guess there's not much reason to support more than 2 sides using this format, and the fact that the labels with more than 2 sides will either be inconsistent with Git or with how they look for 2-sides conflicts seems like enough of a reason to say that we simply won't support more than 2 sides ever (falling back to "snapshot" instead).

scott2000: > Btw, if we plan to support more than 2 sides, then the second side will have to go on the `=======` line instead of on the `>>>>>>>` line (yes, different from Git). I guess there's not much reason to support more than 2 sides using this format, and the fact that the labels with more than 2 sides will either be inconsistent with Git or with how they look for 2-sides conflicts seems like enough of a reason to say that we simply won't support more than 2 sides ever (falling back to "snapshot" instead).

Yeah, I agree with all of that. I updated it to the second format since it matches the Git format a bit better, and it seems like the point of having this option is to match Git as closely as possible. I also updated the fallback to "snapshot" since it's more similar of a format.

lib/tests/test_conflicts.rs
```diff
@@ -1065,6 +1189,249 @@ fn test_update_conflict_from_content_simplified_conflict() {
...
+                    .unwrap();
+
+            assert_eq!(
+                parsed, conflict,
+                "parse {conflict_marker_style:?} conflict markers with {other_style:?}"
+            );
+        }
+
+        content
+    };
```
martinvonz: It looks like we update from an unmodified conflict here, right? It looks like it's only the format that changed. If that's right, I think it would be clearer if we made one of two changes:

1. Pass in a different conflict on line 1239. It could be a completely different conflict as long as it has the right arity.
2. Use `materialize_conflict_string` to produce the materialized content. The test could then become a double loop like this:
```rs
for old_style in [ConflictMarkerStyle::Diff, ConflictMarkerStyle::Snapshot] {
  let content = materialize_conflict_string(store, path, &conflict, conflict_marker_style);
  for new_style in [ConflictMarkerStyle::Diff, ConflictMarkerStyle::Snapshot] {
    let parsed = ...;
    assert_eq!(...);
  }
}
```

I'd probably go with 2 because it's simpler.

The same applies to the next test case.

Does that make sense?

lib/tests/test_conflicts.rs
```diff
@@ -944,6 +1100,26 @@ fn test_parse_conflict_malformed_diff() {
     );
 }
 
+#[test]
+fn test_parse_conflict_wrong_git_style() {
```
martinvonz, line 1104: Might also want one for this (i.e. "snapshot" format with the first `+++++++` line missing):

```
line 1
<<<<<<<
left
-------
base
+++++++
right
>>>>>>>
line 5
```

That seems like something that could easily be allowed when maybe it shouldn't. Might also want one test for a mix of "snapshot" and "git" delimiters. It seems unlikely that these cases would happen in practice but might be best to be strict to start? It's hard to make it more strict later if find a reason to want to make it more strict.

lib/src/conflicts.rs
ilyagr: If we really want to make this format compatible with Git tools, we might want to eventually have a "yours:filename" and "ours:filename" side, like https://git-scm.com/docs/git-merge#_how_conflicts_are_presented . I have no idea how many of Git tools rely on those markers being in that format, we could first check whether anybody requests it.

For now it'd be difficult to consistently pick an ours/theirs side, we'd have to pick arbitrarily. So, there's less reason to follow Git here.

As long as we're not following Git, I also think the second format is a bit nicer.

lib/src/conflicts.rs
```diff
@@ -473,38 +515,71 @@ pub fn parse_conflict(input: &[u8], num_sides: usize) -> Option<Vec<Merge<BStrin
...
+                    state = State::Minus;
+                    removes.push(BString::new(vec![]));
+                    continue;
+                }
+                // ======= indicates Git-style add section
+                CONFLICT_GIT_SEPARATOR_LINE_CHAR => {
+                    // Take lines before the first marker and treat them as an add
+                    if let State::BeforeMarkers = state {
+                        adds.push(std::mem::take(&mut before_markers));
+                    }
```
martinvonz, line 575: Does this mean that `========` is allowed before `|||||||`? Oh, this is for the "empty side 1 and reordered sections" test? I would consider that invalid instead, but I don't feel strongly.

I wouldn't mind if we disallowed reordered sections in the "snapshot" format either.

cli/src/config/merge_tools.toml
```diff
@@ -52,10 +52,13 @@ merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
 # markers. Unfortunately, it does not seem to be able to output conflict markers when
 # the user only resolves some of the conflicts.
 merge-tool-edits-conflict-markers = true
+# VS Code prefers Git-style conflict markers
+conflict-marker-style = "git-diff3"
```
martinvonz, line 56: Set "snapshot" for vimdiff? I don't use vimdiff but I'm pretty sure I've heard that vimdiff users would prefer that. Maybe some vimdiff user can confirm?

docs/conflicts.md
```diff
@@ -50,61 +50,117 @@ For information about how conflicts are handled in the working copy, see
 
 ## Conflict markers
 
+Conflicts happen when Jujutsu can't figure out how to merge different changes
+made to the same file. For instance, this can happen if two people are working
+on the same file and make different changes to the same part of the file, and
+then their commits are merged together with `jj new` (or one is rebased onto the
+other with `jj rebase`).
```
martinvonz, line 57: I feel like this paragraph belongs at the start of "Introduction" (before what's currently on line 6). What do you think?

docs/conflicts.md
```diff
@@ -50,61 +50,117 @@ For information about how conflicts are handled in the working copy, see
...
-In this example, the left side changed "grape" to "grapefruit", and the right
-side made all lines uppercase. To resolve the conflict, we would presumably keep
-the right side (the third section) and replace "GRAPE" by "GRAPEFRUIT". This way
-of visually finding the changes between the base and one side and then applying
-them to the other side is a common way of resolving conflicts when using Git's
-"diff3" style.
+One person replaces the word "grape" with "grapefruit" in commit A, while
+another person changes every line to uppercase in commit B. If you merge the
+changes together with `jj new A B`, the resulting commit will have a conflict
+since Jujutsu can't figure out how to combine these changes together. Therefore,
```
martinvonz, line 77: nit: drop "together" because it's redundant with "combine"?

martinvonz (commented):
Thanks a lot for this PR! I know a lot of people want this feature.

lib/src/annotate.rs
```diff
@@ -358,9 +391,9 @@ fn get_file_contents(
                 })?;
             Ok(file_contents.into())
         }
-        MaterializedTreeValue::FileConflict { contents, .. } => {
-            Ok(materialize_merge_result_to_bytes(&contents))
-        }
+        MaterializedTreeValue::FileConflict { contents, .. } => Ok(
+            materialize_merge_result_to_bytes(&contents, conflict_marker_style),
```
yuja, line 395: nit: just hardcode the default here?

This is just a workaround until we implement a proper tracking of conflict files.

lib/src/default_index/revset_engine.rs
```diff
@@ -1346,7 +1353,7 @@ fn to_file_content(path: &RepoPath, value: MaterializedTreeValue) -> BackendResu
         MaterializedTreeValue::Symlink { id: _, target } => Ok(target.into_bytes()),
         MaterializedTreeValue::GitSubmodule(_) => Ok(vec![]),
         MaterializedTreeValue::FileConflict { contents, .. } => {
-            Ok(materialize_merge_result_to_bytes(&contents).into())
+            Ok(materialize_merge_result_to_bytes(&contents, conflict_marker_style).into())
```
yuja, line 1356: nit: move the hardcoded value here. This is also a temporary hack.

lib/src/settings.rs
```diff
@@ -261,6 +262,14 @@ impl UserSettings {
     pub fn sign_settings(&self) -> SignSettings {
         SignSettings::from_settings(self)
     }
+
+    pub fn conflict_marker_style(&self) -> Result<ConflictMarkerStyle, config::ConfigError> {
+        Ok(self
+            .config
+            .get::<ConflictMarkerStyle>("ui.conflict-marker-style")
```
yuja, line 269: nit: the type can be deduced

lib/src/local_working_copy.rs
yuja: The plumbing code looks good to me, thanks.

lib/tests/test_conflicts.rs
scott2000: I was originally trying to do two different things in the test: checking that it materialized in the correct format, and checking that parsing the materialized conflict works regardless of the configured conflict marker style. I agree that it's a bit unclear, so I converted that test to your "option 2", and I added separate tests for the format of materialized conflicts. I think it's a bit cleaner, thanks!

lib/tests/test_conflicts.rs
scott2000: I added tests for the cases you mentioned, and I made the parsing logic a bit more strict. Now it determines whether it's a Git-style conflict or a JJ-style conflict based on the first line of the hunk, and then parses it using the appropriate function. When parsing a JJ-style conflict, Git-style conflict markers are ignored; and when parsing a Git-style conflict, JJ-style conflict markers are ignored. Does this make sense to you?

lib/src/conflicts.rs
scott2000: Yes, it was allowed previously. I updated it to have a stricter parsing logic for Git-style conflicts that requires the 3 sections to be in order. Since the parsing for "snapshot" is the same as "diff", and "diff" currently allows sections to be reordered, I left it as-is for "snapshot".

cli/src/config/merge_tools.toml
scott2000: Per @yuja's suggestion, I'll be moving the merge tool changes to a separate PR: #4951.

docs/conflicts.md
scott2000: Makes sense to me! I moved it.

lib/src/annotate.rs
scott2000: Ok, that makes sense to me! I updated it.

lib/src/default_index/revset_engine.rs
scott2000: Ok, makes sense!

lib/src/conflicts.rs
```diff
@@ -355,6 +338,33 @@ fn materialize_conflict_hunks(
     Ok(())
 }
 
+/// Write a positive snapshot (side) of a conflict
+fn write_side(add_index: usize, data: &[u8], output: &mut dyn Write) -> std::io::Result<()> {
+    output.write_all(CONFLICT_PLUS_LINE)?;
+    output.write_all(format!(" Contents of side #{}\n", add_index + 1).as_bytes())?;
```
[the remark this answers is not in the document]
scott2000: Yeah, that makes things a lot easier to read actually, thanks! I had to make it be `&str` because I couldn't find a way to construct a `&BStr` in a `const` context and I didn't want to use a `LazyLock`.

lib/src/conflicts.rs
```diff
@@ -355,6 +338,33 @@ fn materialize_conflict_hunks(
...
+    base_str: &str,
+    add_index: usize,
+    diff: &[DiffHunk],
+    output: &mut dyn Write,
+) -> std::io::Result<()> {
+    output.write_all(CONFLICT_DIFF_LINE)?;
+    output
+        .write_all(format!(" Changes from {base_str} to side #{}\n", add_index + 1).as_bytes())?;
+    write_diff_hunks(diff, output)
+}
```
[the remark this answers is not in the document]
scott2000: Makes sense, I updated it.

lib/src/conflicts.rs
```diff
@@ -473,38 +515,71 @@ pub fn parse_conflict(input: &[u8], num_sides: usize) -> Option<Vec<Merge<BStrin
...
 
+/// This method handles parsing both JJ-style and Git-style conflict markers,
+/// meaning that switching conflict marker styles won't prevent existing files
+/// with other conflict marker styles from being parsed successfully.
+///
+/// The Git "|||||||" and "=======" sections are treated as "removes" and "adds"
+/// respectively. If they are the first marker present in the conflict, then any
+/// lines before them are treated as an "add" implicitly (because Git doesn't
+/// put any marker before the first "add").
 fn parse_conflict_hunk(input: &[u8]) -> Merge<BString> {
```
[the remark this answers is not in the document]
scott2000: I updated the logic to first determine which style is being used, and then dispatch to separate functions to handle each style individually. Could you take another look when you get a chance?

cli/tests/test_config_command.rs
```diff
@@ -1018,6 +1018,124 @@ fn test_config_author_change_warning_root_env() {
     );
 }
 
+#[test]
+fn test_config_change_conflict_marker_style() {
```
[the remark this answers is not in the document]
scott2000: Yeah, I was struggling to find a good place to put it. That seems like a much better place, thanks!

martinvonz (approved):

lib/src/conflicts.rs
scott2000: I think the format actually depends on the version of Git, the user's config options, and/or which particular command produced the conflict, because this is what I get when running `git merge branch-name` on my machine:

```
<<<<<<< HEAD
fn example(word: String) {
    println!("word is {word}");
|||||||
fn example(w: String) {
    println!("word is {w}");
=======
fn example(w: &str) {
    println!("word is {w}");
>>>>>>> branch-name
}
```

For `git rebase branch-name`, the conflict end marker is slightly different, with `<short commit ID> (<description>)` for the text. Anecdotally, I've seen Git produce a few different formats of text next to these markers, so I don't think we should be too concerned about matching exactly. If someone runs into a case where a tool doesn't parse our format, we can always update it later.

lib/src/conflicts.rs
```diff
@@ -547,6 +611,55 @@ fn parse_conflict_hunk(input: &[u8]) -> Merge<BString> {
     }
 }
 
+fn parse_git_style_conflict_hunk(input: &[u8]) -> Merge<BString> {
+    enum State {
+        Left,
+        Base,
+        Right,
+    }
```
yuja, line 619: nit: derive Eq instead of `if let ..`?

lib/src/conflicts.rs
```diff
@@ -474,7 +512,33 @@ pub fn parse_conflict(input: &[u8], num_sides: usize) -> Option<Vec<Merge<BStrin
...
+        // JJ-style conflicts must start with one of these 3 conflict marker lines
+        Some(CONFLICT_DIFF_LINE_CHAR | CONFLICT_MINUS_LINE_CHAR | CONFLICT_PLUS_LINE_CHAR) => {
+            parse_jj_style_conflict_hunk(input)
+        }
+        // Git-style conflicts either must not start with a conflict marker line, or must start with
+        // the "|||||||" conflict marker line (if the first side was empty)
+        None | Some(CONFLICT_GIT_ANCESTOR_LINE_CHAR) => parse_git_style_conflict_hunk(input),
+        // No other conflict markers are allowed at the start of a hunk
+        Some(_) => Merge::resolved(BString::new(vec![])),
+    }
```
yuja, line 538: nit: we might want to handle cases that git-style content had a line identical to jj-style marker and vice versa, but that's not new problem.

(No action requested)

yuja (approved):

cli/src/config-schema.json
ilyagr: Turns out Git sort of supports longer conflict markers, see https://github.com/martinvonz/jj/pull/4969#discussion_r1859088488.

PR #4898, opened 2024-11-17
completion: teach log about files

senekor:
stacked on #4895 (suggest file paths incrementally)

# Checklist

If applicable:

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -638,6 +638,17 @@ pub fn interdiff_files(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
     })
 }
 
+/// Specific function for completing file paths for `jj log`
+pub fn log_files(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
+    let mut rev = parse::log_revisions().join(")|(");
+    if rev.is_empty() {
+        rev = "@".into();
+    } else {
+        rev = format!("latest(heads(({rev})))"); // limit to one
+    };
+    all_files_from_rev(rev, current)
+}
+
 /// Shell out to jj during dynamic completion generation
 ///
 /// In case of errors, print them and early return an empty vector.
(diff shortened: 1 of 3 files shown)


cli/src/complete.rs
```diff
@@ -394,6 +394,44 @@ mod parse {
...
+        while let Some(arg) = args.next() {
+            if candidates.contains(&arg.as_ref()) {
+                match args.next() {
+                    Some(rev) => revisions.push(rev),
+                    None => continue,
+                }
+            }
+            let Some((flag, rev)) = arg.split_once('=') else {
+                continue;
+            };
```
martinvonz, line 419: Share some code with `parse_flag()` here? Perhaps a function that takes candidates and a mutable reference to an iterator would work?

thoughtpolice (approved):

cli/src/complete.rs
```diff
@@ -472,10 +472,35 @@ mod parse {
...
+        let mut args = std::env::args();
+
+        let union = std::iter::from_fn(|| parse_flag(candidates, &mut args))
+            // multiple -r arguments are interpreted as a union
+            .fold("none()".into(), |mut buf: String, rev| {
+                buf.push_str("|(");
+                buf.push_str(&rev);
+                buf.push(')');
+                buf
+            });
```
yuja, line 499: nit: just return `Vec<String>` since `log` can accept multiple revisions?

cli/src/complete.rs
```diff
@@ -641,6 +666,43 @@ pub fn interdiff_files(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
...
+        let output = jj
+            .arg("log")
+            .arg("--no-graph")
+            .arg("--template=")
+            .arg("--summary")
+            .arg("--revisions")
+            .arg(rev)
+            .output()
+            .map_err(user_error)?;
+        let stdout = String::from_utf8_lossy(&output.stdout);
```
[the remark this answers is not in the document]
senekor: What do you mean by most recent revision? And what sparse pattern?

I do think it would've been nice if the completions included files that were deleted (/ renamed) in one revision, but were present in another. But it's a small compromise if the performance is much better.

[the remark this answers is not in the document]
yuja: > What do you mean by most recent revision?

`--limit 1` or `heads(..)`. It's not uncommon to query `jj log -r::main <PATH>`, and the `<PATH>` is usually contained in the `main` revision.

> And what sparse pattern?

It's related to the stuff Martin pointed out. Listing all files can be super expensive, and it might make sense to filter paths by the working-copy's sparse pattern in addition to the current path prefix.

cli/src/complete.rs
```diff
@@ -638,6 +638,25 @@ pub fn interdiff_files(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
...
+    let mut rev = parse::log_revisions()
+        // multiple -r arguments are interpreted as a union
+        .into_iter()
+        .fold("none()".into(), |mut buf: String, rev| {
+            buf.push_str("|(");
+            buf.push_str(&rev);
+            buf.push(')');
+            buf
+        });
+    if rev == "none()" {
```
yuja, line 652: nit: perhaps, it's simpler to just `.join(")|(")` and test `.is_empty()` here.

cli/src/complete.rs
```diff
@@ -638,6 +638,25 @@ pub fn interdiff_files(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
...
+        rev = "@".into();
+    } else {
+        rev = format!("latest({rev})"); // limit to one
```
yuja, line 655: It's kinda implementation detail, but `latest()` is expensive if the source set is large. We can instead use `latest(heads(..))` for now.

cli/src/complete.rs
```diff
@@ -837,10 +856,18 @@ mod parse {
...
-        parse_flag(&["-f", "--from"], std::env::args())
+        parse_flag(&["-f", "--from"], &mut std::env::args())
+    }
+
+    // Special parse function only for `jj log`. It has a --revisions flag,
+    // instead of the usual --revision, and it can be supplied multiple times.
+    pub fn log_revisions() -> Vec<String> {
+        let candidates = &["-r", "--revisions"];
+        let mut args = std::env::args();
+        std::iter::from_fn(|| parse_flag(candidates, &mut args)).collect()
```
yuja, line 870: nit: I would make `parse_flag()` return an iterator. It's not so obvious where the input iterator points after `parse_flag()` returned `Some(_)`. It's also easier to write a unit test.

cli/src/complete.rs
```diff
@@ -778,36 +789,41 @@ impl JjBuilder {
...
 /// actual execution of the command would fail.
 mod parse {
-    fn parse_flag(candidates: &[&str], mut args: impl Iterator<Item = String>) -> Option<String> {
-        for arg in args.by_ref() {
-            // -r REV syntax
-            if candidates.contains(&arg.as_ref()) {
-                match args.next() {
-                    Some(val) if !val.starts_with('-') => return Some(val),
-                    _ => return None,
+    fn parse_flag<'a>(
```
yuja, line 792: nit: it's nice if we had a parsing test for multiple revisions.

yuja (approved):
Thanks.

PR #4900, opened 2024-11-17
completion: include bookmarks and tags in revisions

senekor:
# Checklist

If applicable:

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -216,8 +216,90 @@ pub fn aliases() -> Vec<CompletionCandidate> {
     })
 }
 
-fn revisions(revisions: &str) -> Vec<CompletionCandidate> {
-    with_jj(|jj, _| {
+fn revisions(revisions: Option<&str>) -> Vec<CompletionCandidate> {
+    with_jj(|jj, config| {
+        // display order
+        const LOCAL_BOOKMARK_MINE: usize = 0;
+        const LOCAL_BOOKMARK: usize = 1;
+        const TAG: usize = 2;
+        const CHANGE_ID: usize = 3;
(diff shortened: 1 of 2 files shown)


martinvonz (approved):
Thanks!

cli/src/complete.rs
```diff
@@ -218,18 +218,58 @@ fn revisions(revisions: &str) -> Vec<CompletionCandidate> {
...
             .arg("--template")
-            .arg(r#"change_id.shortest() ++ " " ++ if(description, description.first_line(), "(no description set)") ++ "\n""#)
+            .arg(
+                r#"
+                change_id.shortest() ++ "\t" ++
+                bookmarks.map(|b| separate("@", b.name(), b.remote())).join(",") ++ "\t" ++
+                tags.map(|t| t.name()).join(",") ++ " " ++
+                if(description, description.first_line(), "(no description set)") ++ "\n"
+                "#,
+            )
```
yuja, line 228: Should we omit old tags and branches? I don't think it's uncommon to query changes between some old tagged revision (specified by tag) and recent commits (specified by `@` or change id.)

senekor: That's true. The revisions are limited to 100, any bookmarks / tags not pointing to this limited list are not shown. I'll have to shell out twice to improve that.

yuja: Yeah, we'll need a helper struct.

BTW, for "all_revisions" query, it might make sense to use `revsets.short-prefixes` (instead of `--limit 100`) to collect change ids which the user would be interested in. For "mutable_revisions", using `mutable()` seems good.

senekor:
Splitting the commands resulted in quite a few changes, I'll wait for another approving review before merging.

yuja (approved):
Thanks.

PR #4907, opened 2024-11-18
revset: allow tags() to take a pattern for an argument

lukerandall:
# Checklist

If applicable:

### DIFF
--- a/lib/src/view.rs
+++ b/lib/src/view.rs
@@ -309,6 +309,17 @@ impl View {
         self.data.tags.get(name).flatten()
     }
 
+    /// Iterates tags `(name, target)`s matching the given pattern. Entries
+    /// are sorted by `name`.
+    pub fn tags_matching<'a: 'b, 'b>(
+        &'a self,
+        pattern: &'b StringPattern,
+    ) -> impl Iterator<Item = (&'a str, &'a RefTarget)> + 'b {
+        pattern
+            .filter_btree_map(&self.data.tags)
+            .map(|(name, target)| (name.as_ref(), target))
+    }
+
     /// Sets tag to point to the given target. If the target is absent, the tag
     /// will be removed.
     pub fn set_tag_target(&mut self, name: &str, target: RefTarget) {
--- a/lib/tests/test_revset.rs
+++ b/lib/tests/test_revset.rs
@@ -2306,6 +2306,91 @@ fn test_evaluate_expression_remote_bookmarks() {
 }
 
 #[test]
+fn test_evaluate_expression_tags() {
+    let settings = testutils::user_settings();
+    let test_repo = TestRepo::init();
+    let repo = &test_repo.repo;
+
+    let mut tx = repo.start_transaction(&settings);
+    let mut_repo = tx.repo_mut();
+
+    let commit1 = write_random_commit(mut_repo, &settings);
+    let commit2 = write_random_commit(mut_repo, &settings);
+    let commit3 = write_random_commit(mut_repo, &settings);
+    let commit4 = write_random_commit(mut_repo, &settings);
+
+    // Can get tags when there are none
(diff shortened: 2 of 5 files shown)


thoughtpolice (commented):
Looks fine, but would you mind adding a couple tests (just 2 or 3 very simple ones) inside the `cat cli/tests/test_tag_command.rs` suite? That will ensure that not only is the revset engine parsing and handling patterns correctly, but that from a user POV the command behaves as expected. You can just add a new `fn test_tag_patterns()` and copy/paste the boilerplate from the other existing test. Otherwise LGTM.

cli/tests/test_tag_command.rs
```diff
@@ -117,3 +117,92 @@ fn test_tag_list() {
     added_targets: commit2
     "###);
 }
+
+#[test]
+fn test_tag_patterns() {
```
yuja, line 122: nit: can you add tests to lib/tests/test_revset.rs instead? Library tests is much faster than CLI tests, and I don't think e2e CLI test is needed for this change.

I think you can copy setup code from `test_resolve_symbol_tags()` and `test_evaluate_expression_bookmarks()`. There was no library test for `tags()` expression yet.

lib/src/view.rs
```diff
@@ -309,6 +309,17 @@ impl View {
         self.data.tags.get(name).flatten()
     }
 
+    /// Iterates local tag `(name, target)`s matching the given pattern. Entries
+    /// are sorted by `name`.
+    pub fn local_tags_matching<'a: 'b, 'b>(
```
yuja, line 314: nit: `tags_matching()` (without `local_`)? We don't have remote tags yet.

yuja (approved):
Thanks!

lib/src/view.rs
thoughtpolice: FWIW this was a quick copy paste job from me, so I take all the blame.

cli/tests/test_tag_command.rs
lukerandall: Have added these. The bookmark tests seem fairly exhaustive, so just copied what they had.

lib/src/view.rs
lukerandall: In fairness if I steal your code I have to take the bad with the good :) have updated this to `tags_matching`.

lukerandall:
@yuja I'm not sure what PR etiquette is here, in general do you prefer resolving your conversations once you're satisfied with the resolution, or do I do it once I've addressed the issue?

martinvonz:
> @yuja I'm not sure what PR etiquette is here, in general do you prefer resolving your conversations once you're satisfied with the resolution, or do I do it once I've addressed the issue?

It's the middle of the night in Japan, so I'll answer. Feel free to mark the issue resolved if you're reasonably confident you resolved it. If you still missed something (e.g. because you misunderstood a reviewer's request), we (collectively, including you) can always fix it after it's merged. I interpret Yuya's approval stamp to mean that the remaining changes were small enough that he trusted you to address them. Still, in a case like this where the change was fairly large (moving the tests to the lib crate), I would personally give the reviewer another chance to review the new state before I merged.

lib/tests/test_revset.rs
```diff
@@ -2305,6 +2305,91 @@ fn test_evaluate_expression_remote_bookmarks() {
     );
 }
 
+#[test]
+fn test_evaluate_expression_tags() {
```
martinvonz, line 2309: Not for this PR: @yuja, to avoid having to duplicate so much test code here, what do you think about moving most of them to `str_util.rs` and then trusting that code and just testing the integration here?

yuja: Sounds good. We don't need to test every detail of string patterns in test_revset.rs.

yuja (approved):

PR #4908, opened 2024-11-19
Ignore .cargo/insta.yaml

essiene:
Ignore .cargo/insta.yaml so folks to customize this locally.

See https://insta.rs/docs/settings/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ result
 *.pending-snap
 *.snap*
 !cli/tests/cli-reference@.md.snap
+# Per user insta settings.
+# See https://insta.rs/docs/settings/#tool-config-file for details.
+.config/insta.yaml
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -8,6 +8,7 @@ result
 *.pending-snap
 *.snap*
 !cli/tests/cli-reference@.md.snap
+.cargo/insta.yaml
```
yuja, line 11: Can you add some comment saying this is per-user configuration?

yuja, line 11: You mean `.config/insta.yaml`?
https://insta.rs/docs/settings/#tool-config-file

essiene: Done.

essiene: Whoops! Fixed. Thanks

yuja (approved):

PR #4909, opened 2024-11-19
git sync: Get branch heads and fetch from remote

essiene:
## Summary
* [X] Get branch pre-fetch heads
* [ ] Build candidates from prefetch heads
* [X] Fetch from remotes
* [X] Get branch post-fetch heads
* [ ] Rebase

## Details
* For branch heads, we grab local branches matching the pattern
  specified in args.
* Pre-fetch heads query tx.base_repo() while post-fetch heads
  query tx.repo(); tx.repo() is the in memory Mutable repo which
  is updated after a fetch in a transaction.
* We fetch from all branches from all configured remotes.
  --all_remotes affects if we use `git.fetch` or all repo remotes.

Issue: https://github.com/martinvonz/jj/issues/1039

### DIFF
--- a/cli/src/commands/git/sync.rs
+++ b/cli/src/commands/git/sync.rs
@@ -13,11 +13,23 @@
 // limitations under the License.
 
 use clap_complete::ArgValueCandidates;
+use itertools::Itertools;
+use jj_lib::backend::CommitId;
+use jj_lib::commit::Commit;
+use jj_lib::repo::Repo;
+use jj_lib::revset::FailingSymbolResolver;
+use jj_lib::revset::RevsetExpression;
+use jj_lib::revset::RevsetIteratorExt;
 use jj_lib::str_util::StringPattern;
 
 use crate::cli_util::CommandHelper;
+use crate::cli_util::WorkspaceCommandTransaction;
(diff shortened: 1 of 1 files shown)


essiene:
I'm not sure if I should add unit tests tests for these or just add integration tests for `jj git sync` directly; I'm planning to do that in the later PRs.

yuja (commented):
I'm not sure what we're going to achieve, but my gut feeling is that this and the subsequent PRs duplicate `git::import_refs()`? I would instead try to extend `git:import_refs()` to return information needed to rebase commits.

And the core "git sync" function might be implemented in lib/src/git.rs.

essiene:
> I'm not sure what we're going to achieve, but my gut feeling is that this and the subsequent PRs duplicate `git::import_refs()`? I would instead try to extend `git:import_refs()` to return information needed to rebase commits.
> 
> And the core "git sync" function might be implemented in lib/src/git.rs.

Thanks for the suggestion. Based on this and comments from @martinvonz, I've gone to actually read `git::import_some_refs()` and I understand it better now. 

Abandoning this PR now and will send a new one that changes `jj-lib` to introduce core functionality for sync to be built atop of.

PR #4922, opened 2024-11-19
cli: delete deprecated `merge` and `checkout` commands

thoughtpolice:
Following in the footsteps of the `jj move` deletion.

# Checklist

If applicable:


cli/src/commands/mod.rs
ilyagr, line 1: Should we remove the `co` alias from `cli/src/config/misc.toml` or change it to "new". This PR should do one or the other.

thoughtpolice: Good catch. I figure it's more likely people use the alias than the actual command post-deprecation, but then again, I have nothing to back this up at all. One thing worth considering is that if we delete it the error message will (probably?) be terrible. So I think I'm in soft-favor of just keeping an alias.

ilyagr: I don't have a strong opinion. Thanks for making a choice, we'll see if anybody complains about it.

martinvonz (approved):
Thanks!

I was wondering if we should instead replace the commands with stubs that just print an error telling the user to go read the tutorial or something. My concern was that new users might just run these commands (even though they're hidden) and maybe they would be annoyed when they don't exist. As I'm sure you saw, I asked users on Discord if they remembered running `jj checkout` before reading about the proper way of doing it. We didn't hear from anyone who did. So I think we should just merge this.

cli/src/commands/mod.rs
yuja: > I suspect most people will just do it themselves if they want it.

That's true. I already have `co = ["new"]` alias to suppress the deprecation warning. Removing the default alias seems fine.

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj move` has been removed. It was deprecated in 0.16.0.
 
+* `jj checkout` and the built-in alias `jj co` have been removed.
+  It was deprecated in 0.14.0.
+
+* `jj merge` has been removed. It was deprecated in 0.14.0.
+
```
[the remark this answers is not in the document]
thoughtpolice: Eh, I think it's fine. We can do some copy editing when the release rolls around if we want.

Issue #4920, opened 2024-11-19
`jj git fetch` with multiple remotes can abandon reachable commits (and rebase reachable descendants)


mateon1:
## Description
jj abandons work from pruned PR branches and mangles history if the pr repo is fetched before the origin.

## Steps to Reproduce the Problem

Minimal reproduction:
1. Add two git remotes b, c, with only an initial commit
2. In b: create a pr branch, commit some work
3. fetch into jj repo
4. In c: do some other work on master, merge b/pr
5. In b: delete pr branch
6. Current state should have b/master pointing to initial commit, c/master to merge commit
7. In jj repo: jj git fetch --all-remotes

## Expected Behavior
Bookmarks are updated, repo has state consistent with both remotes.
Fetch never, under any circumstances modifies any commits. Abandons should only happen for commits that are unreachable after all fetches.

## Actual Behavior

b is pulled first, the commit is marked to be abandoned
c is pulled, referencing the lost commit
jj blindly performs the abandon, modifying history such that 'work' is lost.

```
[mateon@ceres:/tmp/a]$ jj git fetch --all-remotes
bookmark: pr@b [deleted] untracked
Abandoned 1 commits that are no longer reachable.
bookmark: master@c [updated] untracked
Rebased 1 descendant commits

[mateon@ceres:/tmp/a]$ jj log -r 'all()'
○    xqknuwot matin1111@wp.pl 2024-11-19 17:24:42 ba93e3ed
├─╮  (empty) Merge remote-tracking branch 'b/pr'
◆ │  qrmnvrwp matin1111@wp.pl 2024-11-19 17:15:43 8ac29516
├─╯  other
◆  yrpzrykw matin1111@wp.pl 2024-11-19 17:06:57 master@b 103e2828
│  (empty) Initial commit
│ @  pznkzwlt matin1111@wp.pl 2024-11-19 17:05:48 47efbbb4
├─╯  (empty) (no description set)
◆  zzzzzzzz root() 00000000

[mateon@ceres:/tmp/a]$ jj op show --patch d255b75cda76
d255b75cda76 mateon@ceres 1 minute ago, lasted 298 milliseconds
fetch from git remote(s) b,c
args: jj git fetch --all-remotes

Changed commits:
○  Change xqknuwottovp
   + xqknuwot ba93e3ed (empty) Merge remote-tracking branch 'b/pr'
○  Change mszwrokvtxot
   - mszwrokv hidden 1780e38a work
   Added regular file work:
[...]

PR #4923, opened 2024-11-20
Improve fetch from multiple remotes

mateon1:
See #4920 for rationale.

# Checklist

If applicable:
- - [x] I have updated existing tests affected by this change

### DIFF
--- a/cli/src/commands/git/clone.rs
+++ b/cli/src/commands/git/clone.rs
@@ -225,7 +225,7 @@ fn do_git_clone(
         git::fetch(
             fetch_tx.repo_mut(),
             &git_repo,
-            remote_name,
+            &[remote_name],
             &[StringPattern::everything()],
             cb,
             &command.settings().git_settings(),
--- a/cli/src/git_util.rs
+++ b/cli/src/git_util.rs
@@ -465,38 +465,36 @@ pub fn git_fetch(
 ) -> Result<(), CommandError> {
     let git_settings = tx.settings().git_settings();
 
-    for remote in remotes {
-        let stats = with_remote_git_callbacks(ui, None, |cb| {
-            git::fetch(
-                tx.repo_mut(),
-                git_repo,
-                remote,
-                branch,
-                cb,
-                &git_settings,
-                None,
-            )
-        })
-        .map_err(|err| match err {
-            GitFetchError::InvalidBranchPattern => {
-                if branch
-                    .iter()
-                    .any(|pattern| pattern.as_exact().is_some_and(|s| s.contains('*')))
-                {
-                    user_error_with_hint(
-                        err,
-                        "Prefix the pattern with `glob:` to expand `*` as a glob",
-                    )
-                } else {
-                    user_error(err)
-                }
+    let stats = with_remote_git_callbacks(ui, None, |cb| {
+        git::fetch(
+            tx.repo_mut(),
+            git_repo,
+            &remotes.iter().map(String::as_str).collect::<Vec<_>>(),
+            branch,
+            cb,
+            &git_settings,
+            None,
+        )
+    })
+    .map_err(|err| match err {
+        GitFetchError::InvalidBranchPattern => {
+            if branch
+                .iter()
+                .any(|pattern| pattern.as_exact().is_some_and(|s| s.contains('*')))
+            {
+                user_error_with_hint(
+                    err,
+                    "Prefix the pattern with `glob:` to expand `*` as a glob",
+                )
+            } else {
+                user_error(err)
             }
-            GitFetchError::GitImportError(err) => err.into(),
-            GitFetchError::InternalGitError(err) => map_git_error(err),
-            _ => user_error(err),
-        })?;
-        print_git_import_stats(ui, tx.repo(), &stats.import_stats, true)?;
-    }
+        }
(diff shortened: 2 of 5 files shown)


yuja:
Perhaps, the easiest workaround is to do `rebase_descendants()` per remote? Does this patch solve the problem?
```diff
diff --git a/cli/src/git_util.rs b/cli/src/git_util.rs
index 74ebe2d2fa..8949218232 100644
--- a/cli/src/git_util.rs
+++ b/cli/src/git_util.rs
@@ -496,6 +496,11 @@
             _ => user_error(err),
         })?;
         print_git_import_stats(ui, tx.repo(), &stats.import_stats, true)?;
+        let settings = tx.settings().clone(); // XXX silly clone
+        let num_rebased = tx.repo_mut().rebase_descendants(&settings)?;
+        if num_rebased > 0 {
+            writeln!(ui.status(), "Rebased {num_rebased} descendant commits")?;
+        }
     }
     warn_if_branches_not_found(
         ui,
```

It might be cleaner to do import at the end, but it's weird that `git:fetch()` accepts multiple remotes (and returns one `default_branch`.)

mateon1:
Yuja's patch does work, with this result:
```
[mateon@ceres:/tmp/a]$ jj git fetch --all-remotes 
bookmark: pr@b [deleted] untracked
Abandoned 1 commits that are no longer reachable.
bookmark: master@c [updated] untracked

[mateon@ceres:/tmp/a]$ jj op show 63915faff2bc
63915faff2bc mateon@ceres 31 seconds ago, lasted 11 milliseconds
fetch from git remote(s) b,c
args: jj git fetch --all-remotes

Changed commits:
○  Change xqknuwottovp
   + xqknuwot bdd92bb1 master@c | (empty) Merge remote-tracking branch 'b/pr'

Changed remote bookmarks:
master@c:
+ untracked xqknuwot bdd92bb1 master@c | (empty) Merge remote-tracking branch 'b/pr'
- untracked qrmnvrwp 8ac29516 other
pr@b:
+ untracked (absent)
- untracked mszwrokv 1780e38a work
```

So it's a direct fix to the current bug while keeping existing behavior.
Personally I find it rather unintuitive that branch changes on one remote can mess with your local work even if the commits still exist in the shared history.
Although frankly I find it surprising that any kind of fetch can rebase your local commits - my expectation as a new jujutsu user was having to rebase your work yourself when a fetch makes your commits stale. I also expected fetch from multiple remotes to be significantly more atomic than it is now.

Re: default_branch, that's also a bit weird. It's only used from jj git clone, which only supports a single remote. Even if it did support multiple, how could you possibly know which remote is 'most authoritative' for choosing the default branch? What if you clone two forks of the same repo? What if you clone two completely independent repos?
I'm not sure what the most correct thing to do here would be. Returning a Vec of default_branches corresponding to the passed remotes? Making a 'clone' implementation that supports only one remote, handles default_branch, and ripping out that code from fetch?

essiene:
> > I find it surprising that any kind of fetch can rebase your local commits - my expectation as a new jujutsu user was having to rebase your work yourself when a fetch makes your commits stale.
> 
> You can turn it off by `git.abandon-unreachable-commits = false`, but you'll need to clean up old branches manually (if the remote does rebase/squash merges.)
> 
> > Re: default_branch, that's also a bit weird. It's only used from jj git clone, which only supports a single remote. Even if it did support multiple, how could you possibly know which remote is 'most authoritative' for choosing the default branch?
> 
> Perhaps, the API can be refactored to something like this:
> 
> ```rust
> let mut peer = GitFetch::new(..);
> let default_branch = peer.fetch(remote_name, branch_names, ..);
> ...
> let stats = peer.import_refs();
> ```


This seems like a good approach.

We could call peer.fetch() in a loop for all remotes and at the end, call peer.import_refs just once.
> 
> (and `git::fetch()` will be a thin wrapper.)

We could also have a thin wrapper for `git::clone` which returns the default branch as `git::fetch` doesn't need to. And I'll be able to add `git::sync` that does the extra stuff I need.

If there are no objections, I go ahead and send out a few PRs that do this. 
> 
> @essiene might be interested.

Indeed. Thanks for looping me in.

yuja:
> We could also have a thin wrapper for `git::clone` which returns the default branch as `git::fetch` doesn't need to. And I'll be able to add `git::sync` that does the extra stuff I need.

Sounds good, thanks.

BTW, I used `peer` in the example, but it's probably a bad word. I expect `peer` to be associated with a single remote.

essiene:
> > We could also have a thin wrapper for `git::clone` which returns the default branch as `git::fetch` doesn't need to. And I'll be able to add `git::sync` that does the extra stuff I need.
> 
> Sounds good, thanks.
> 
> BTW, I used `peer` in the example, but it's probably a bad word. I expect `peer` to be associated with a single remote.

Yeah, makes sense.

essiene:
> > > We could also have a thin wrapper for `git::clone` which returns the default branch as `git::fetch` doesn't need to. And I'll be able to add `git::sync` that does the extra stuff I need.
> > 
> > 
> > Sounds good, thanks.
> > BTW, I used `peer` in the example, but it's probably a bad word. I expect `peer` to be associated with a single remote.
> 
> Yeah, makes sense.

#4959 & #4960

mateon1:
Closed due to being superseded by #5498

PR #4925, opened 2024-11-20
`git fetch`: clarify errors with bad characters in globs and branch names

ilyagr:
The changes are minor, but hopefully will make it clearer that fetch branch patterns
are unlike other bookmark patterns.

# Checklist

If applicable:


cli/src/git_util.rs
```diff
@@ -484,7 +484,7 @@ pub fn git_fetch(
                     .any(|pattern| pattern.as_exact().is_some_and(|s| s.contains('*')))
                 {
                     user_error_with_hint(
-                        err,
+                        GitFetchError::StarInBranchName,
```
ilyagr, line 487: I realized there's no reason for this to be a `GitFetchError`, which could simplify the commit. However, I don't know what to use instead. I'll look what happens in other similar situations tomorrow, unless somebody lets me know.

lib/src/git.rs
```diff
@@ -1273,7 +1273,10 @@ pub fn fetch(
         .map(|pattern| {
             pattern
                 .to_glob()
-                .filter(|glob| !glob.contains(INVALID_REFSPEC_CHARS))
+                .filter(
+                    /* This check will fail if `to_glob()` escapes a literal `*` */
```
yuja, line 1277: nit: "check will fail" sounds like the check doesn't work. Remove the comment? or say something like "to_glob() should escape * as [*] if the pattern wasn't a glob"?

ilyagr: I edited it a bit, hope it's better.

PR #4929, opened 2024-11-20
cargo: enable zlib-ng and fast sha1 in gix

yuja:
# Checklist

If applicable:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ gix = { version = "0.68.0", default-features = false, features = [
     "attributes",
     "blob-diff",
     "index",
-    "max-performance-safe",
+    "max-performance",
 ] }
 glob = "0.3.1"
 hashbrown = { version = "0.15.2", default-features = false, features = ["inline-more"] }
--- a/flake.nix
+++ b/flake.nix
@@ -100,8 +100,12 @@
             makeWrapper
             pkg-config
 
+            # for libz-ng-sys (zlib-ng)
+            # TODO: switch to the packaged zlib-ng and drop this dependency
+            cmake
+
             # for signing tests
-            gnupg 
+            gnupg
             openssh
           ] ++ linuxNativeDeps;
           buildInputs = with pkgs; [
@@ -184,6 +188,10 @@
 
           # In case you need to run `cargo run --bin gen-protos`
           protobuf
+
+          # for libz-ng-sys (zlib-ng)
+          # TODO: switch to the packaged zlib-ng and drop this dependency
+          cmake
 
           # To run the signing tests
           gnupg
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -430,6 +430,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cbd0f76e066e64fdc5631e3bb46381254deab9ef1158292f27c8c57e3bf3fe59"
 
 [[package]]
+name = "cmake"
+version = "0.1.52"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c682c223677e0e5b6b7f63a64b9351844c3f1b1678a68b7ee617e30fb082620e"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "colorchoice"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -756,6 +765,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "324a1be68054ef05ad64b861cc9eaf1d623d2d8cb25b4bf2cb9cdd902b4bf253"
 dependencies = [
  "crc32fast",
+ "libz-ng-sys",
  "miniz_oxide 0.8.0",
 ]
 
@@ -1139,6 +1149,7 @@ dependencies = [
  "once_cell",
  "parking_lot",
  "prodash",
+ "sha1",
  "sha1_smol",
  "thiserror 2.0.4",
  "walkdir",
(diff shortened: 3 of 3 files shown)


flake.nix
```diff
@@ -91,8 +91,11 @@
             makeWrapper
             pkg-config
 
+            # for libz-ng-sys (zlib-ng)
+            cmake
```
thoughtpolice, line 95: Ugh, I'm not a huge fan of this. It looks like by default it will actually cause a full build of the library in the background, based on my reading of the build file? We already have `zlib-ng` packaged fully in Nixpkgs, so I'd rather avoid that if possible. I'll have to take a closer look later today.

What's the performance impact from this? I'm figure it's reasonably significant (zlib-ng really is much faster) but I'm curious.

yuja: I don't know nix, so it'll be nice if you can figure out what we should do here.

> What's the performance impact from this?

See the commit message. It has some measurable difference in file/annotate query.

emilazy: It seems like `libz-ng-sys`’s build script is tragically unwilling to use a system package, even though `libz-sys` does the right thing and checks `pkg-config`.

@Byron Would you accept a patch to `libz-ng-sys` to move the `zlib-ng` logic in the build script so that it can use system packages for that library as well where available?

thoughtpolice: Yes, if we can get it to just use the system libraries like we do with `zstd` and `libssh2` (`export ZSTD_SYS_USE_PKG_CONFIG=1` for example), then that should be sufficient to just drop the `cmake` input and use `zlib-ng` instead.

emilazy: `libz-sys` doesn’t support an environment variable in general, unfortunately (or we’d export it from our Rust setup hook). But it does look for system `zlib` first, and `libz-ng-sys` is maintained in the same repository with the same build script, so I am hoping it will be a simple matter of sending a PR for the `build.rs` and bumping to a new Gitoxide release. Then we can use the Nixpkgs `zlib-ng` package and everyone is happy.

Byron: > @Byron Would you accept a patch to `libz-ng-sys` to move the `zlib-ng` logic in the build script so that it can use system packages for that library as well where available?

It seems like this should be done, even though I don't know if there is any reason this wasn't done.
Ideally, the current behaviour remains, and a flag of sorts is added to enable looking for the system package. 

From what I know, `libz-ng-sys` uses its own build-script as well.

emilazy: `build.rs` calls out to some `libz-ng-sys`‐specific code, but the actual file itself is shared, and that’s the file that contains the `pkg-config` logic, so I figured it would just need some moving around and conditional names.

Would you be amenable to an environment variable that forces the use of `pkg-config`, for both `libz-sys` and `libz-ng-sys`? That would solve the current problem and be convenient in general for Nixpkgs, as most other packages that vendor libraries support it, and the current default behaviour of proactively searching `pkg-config` for `libz-sys` but not for `libz-ng-sys` could be retained for backwards compatibility.

Byron: Yes, environment variables seem to be the most common way to do that and keep it backwards compatible with it. Thank you.

emilazy: Thanks! I’ll try to send a PR off today.

Cargo.toml
```diff
@@ -61,7 +61,7 @@ gix = { version = "0.67.0", default-features = false, features = [
     "attributes",
     "blob-diff",
     "index",
-    "max-performance-safe",
+    "max-performance",
```
ilyagr, line 64: I'm slightly worried about whether this will compile correctly as a static linux-musl binary, like our release build. Did you check?

Perhaps we should have a presubmit check that checks that directly. I'm not sure the `--all-features` check is sufficient (though the build that would happen after the PR is merged would probably check it). Or do we have one that I forgot about?

yuja: Oh, I didn't know that release builds aren't triggered by PRs.

The x86_64 musl binary works fine.

yuja: I think that's a general concern about including C dependency.

FWIW, the latest libz-ng-sys has a linking issue with static msvcrt, which was detected by CI.

thoughtpolice: Yeah, I see your note about downgrading it. I wasn't really too worried about musl in particular since I think `zlib-ng` is used in that part of the world, but the MSVC one is more interesting.

Well, I'll see if I can try to fix that static linking issue, but the changes between the current version and the working one will need a bit of review and investigation (lots of "commit soup" it seems...)

flake.nix
yuja: Is this flake.nix issue a blocker? Can we address it later? or add feature flag to disable zlib-ng on nix?

thoughtpolice: Let's just merge it for now; I'd rather not have a true "parity" distinction between Nix users and others. We can look at fixing `zlib-ng` in a proper way this release cycle.

thoughtpolice (approved):

thoughtpolice:
@yuja Actually I just tested this, and you need to add `cmake` in one more spot: please add it inside the `devShell` dependencies at the very end of the `flake.nix`. Otherwise, it breaks my ability to compile interactively. I can also do this as a follow up (and I should deduplicate this code a bit so it doesn't happen again...)

yuja:
@thoughtpolice
> you need to add `cmake` in one more spot: please add it inside the `devShell` dependencies at the very end of the `flake.nix`.

Fixed. Can you take another look? I also added a TODO comment.

thoughtpolice:
> @thoughtpolice
> 
> > you need to add `cmake` in one more spot: please add it inside the `devShell` dependencies at the very end of the `flake.nix`.
> 
> Fixed. Can you take another look? I also added a TODO comment.

WFM. Land when you're ready.

PR #4944, opened 2024-11-22
conflicts: whitespace parsing fixes

scott2000:
I noticed these while working on #4897 and I thought I'd make a PR for them.

# Checklist

If applicable:

### DIFF
--- a/lib/src/conflicts.rs
+++ b/lib/src/conflicts.rs
@@ -467,6 +467,12 @@ fn parse_conflict_hunk(input: &[u8]) -> Merge<BString> {
                 } else if let Some(rest) = line.strip_prefix(b" ") {
                     removes.last_mut().unwrap().extend_from_slice(rest);
                     adds.last_mut().unwrap().extend_from_slice(rest);
+                } else if line == b"\n" || line == b"\r\n" {
+                    // Some editors strip trailing whitespace, so " \n" might become "\n". It would
+                    // be unfortunate if this prevented the conflict from being parsed, so we add
+                    // the empty line to the "remove" and "add" as if there was a space in front
+                    removes.last_mut().unwrap().extend_from_slice(line);
+                    adds.last_mut().unwrap().extend_from_slice(line);
                 } else {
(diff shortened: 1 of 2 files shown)


lib/tests/test_conflicts.rs
```diff
@@ -768,8 +768,8 @@ fn test_parse_conflict_crlf_markers() {
 
 #[test]
 fn test_parse_conflict_diff_stripped_whitespace() {
-    // Conflict parsing fails since diff contains empty line without leading space
-    assert_eq!(
+    // Should be able to parse conflict even if diff contains empty line
```
ilyagr, line 771: I'd add "(without even a leading space, which are sometimes stripped by text editors)" to this comment.

ilyagr (commented):
This looks fine to me. I'll wait a bit before approving in case there are other comments.

lib/src/conflicts.rs
```diff
@@ -69,6 +69,7 @@ const CONFLICT_PLUS_LINE_CHAR: u8 = CONFLICT_PLUS_LINE[0];
 static CONFLICT_MARKER_REGEX: once_cell::sync::Lazy<Regex> = once_cell::sync::Lazy::new(|| {
     RegexBuilder::new(r"^(<{7}|>{7}|%{7}|\-{7}|\+{7})( .*)?$")
         .multi_line(true)
+        .crlf(true)
```
yuja, line 72: nit: might be safer to trim ascii whitespace at end. `ByteSlice::trim_end_with()` can be used for that purpose (because `.trim_ascii_end()` requires rustc 1.80.) 
https://docs.rs/bstr/latest/bstr/trait.ByteSlice.html#method.trim_end_with

It might also make sense to rewrite the regex with simple bytes functions. It's harder to spot edge cases of regex pattern.

yuja (approved):

lib/src/conflicts.rs
scott2000: Ok, makes sense to me. I went with your first suggestion, since it seemed like the smaller change.

PR #4947, opened 2024-11-22
config: move some utility types to lib, replace `settings.config()` callers

yuja:
# Checklist

If applicable:


cli/src/config.rs
```diff
@@ -135,19 +135,19 @@ impl LayeredConfigs {
     }
 
     #[instrument]
-    pub fn read_user_config(&mut self) -> Result<(), ConfigError> {
+    pub fn read_user_config(&mut self) -> Result<(), ConfigEnvError> {
```
martinvonz, line 139: Optional: Could make all these return `Self` instead. Probably a good idea in general.

lib/src/settings.rs
```diff
@@ -265,17 +265,17 @@ impl UserSettings {
 
     /// Looks up string value by `key`.
     pub fn get_string(&self, key: &str) -> Result<String, ConfigError> {
-        self.get(key).and_then(config::Value::into_string)
+        self.get(key)
```
martinvonz, line 268: Why didn't the code in the second-to-last patch use `self.get_string(key)` instead? That seems closer to what we used before, so it would seem more natural to me.

martinvonz (approved):
Thanks for working on this!

cli/src/config.rs
yuja: `LayeredConfigs` isn't used in a builder-like way, so `Self -> Self` method is inconvenient. `&mut Self -> &mut Self` works, but the caller would have to allocate `Self` somewhere to get back owned object.

lib/src/settings.rs
yuja: I just wanted to remove `self.config.` references because they'll be replaced with new config API. It's not important because of the last patch, though.

PR #4951, opened 2024-11-22
merge_tools: allow setting conflict marker style per-tool

scott2000:
Follow-up to #4897.

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/merge_tools.toml
+++ b/cli/src/config/merge_tools.toml
@@ -39,6 +39,7 @@ program = "vim"
 merge-args = ["-f", "-d", "$output", "-M", "$left", "$base", "$right",
               "-c", "wincmd J", "-c", "set modifiable", "-c", "set write"]
 merge-tool-edits-conflict-markers = true
+conflict-marker-style = "snapshot"
 # Using vimdiff as a diff editor is not recommended. For instructions on configuring
 # the DirDiff Vim plugin for a better experience, see
 # https://gist.github.com/ilyagr/5d6339fb7dac5e7ab06fe1561ec62d45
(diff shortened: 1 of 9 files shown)


cli/src/config/merge_tools.toml
```diff
@@ -52,10 +52,13 @@ merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
 # markers. Unfortunately, it does not seem to be able to output conflict markers when
 # the user only resolves some of the conflicts.
 merge-tool-edits-conflict-markers = true
+# VS Code prefers Git-style conflict markers
+conflict-marker-style = "git"
```
scott2000, line 56: https://github.com/martinvonz/jj/pull/4897#discussion_r1853311074

> Set "snapshot" for vimdiff? I don't use vimdiff but I'm pretty sure I've heard that vimdiff users would prefer that. Maybe some vimdiff user can confirm?

scott2000: This makes sense to me. I think it gives better diff highlighting, since it allows vim to find more matching lines.

cli/src/merge_tools/external.rs
```diff
@@ -61,13 +61,15 @@ pub struct ExternalMergeTool {
...
+    /// the merge tool is done, any remaining conflict markers in the
+    /// file are parsed and taken to mean that the conflict was only partially
     /// resolved.
-    // TODO: Instead of a boolean, this could denote the flavor of conflict markers to put in
-    // the file (`jj` or `diff3` for example).
     pub merge_tool_edits_conflict_markers: bool,
+    /// If provided, overrides the normal conflict marker style setting. This is
+    /// useful if a merge tool parses conflict markers, and so it requires a
+    /// specific format.
+    pub conflict_marker_style: Option<ConflictMarkerStyle>,
```
yuja, line 72: Are we going to apply this style to diff generation (i.e. `jj diff` of external tools)? If not, I would rename it to `merge_tool_conflict_marker_style`.

Regarding `merge-tool-edits-conflict-markers`, I'm not sure whether we should merge them into one. If we prefer merging, `merge-tool-edits-conflict-markers: true` can be translated to the default style (maybe with a deprecation warning.)

scott2000: I think it makes sense to apply it for diff generation too, since it seems like a useful option to have. For instance, a diff tool might have syntax highlighting, and it might know how to recognize Git-style conflict markers and highlight them appropriately. I don't feel too strongly about it though, since I think it's less common for diff tools to care about conflict markers than merge tools.

PR #4953, opened 2024-11-23
`util gc`: document the fact that it does not abandon operations

ilyagr:


cli/src/commands/util/gc.rs
```diff
@@ -24,6 +24,9 @@ use crate::command_error::CommandError;
 use crate::ui::Ui;
 
 /// Run backend-dependent garbage collection.
+///
+/// To abandon old operations, run `jj op abandon ..<some old operation>` before
+/// `jj util gc`.
```
martinvonz, line 29: nit: Should we say "To garbage-collect old operations, ..." instead? Because otherwise the "before ..." part isn't needed. Or maybe it's already clear and I'm just being overly pedantic (very possible).

yuja: Or "To garbage-collect old hidden commits"? If the user knows reachable commits cannot be GC-ed, they would do `jj op abandon` without a hint.

ilyagr: Good point! How about this?

ilyagr:
Two vaguely related points:

- On [Discord](https://discord.com/channels/968932220549103686/1309593022043914340/1309884243773751367), @scott2000 aksed about the evolog, and Yuya said (IIUC) that old evolog entries are never garbage collected. Should this be changed? It seems like most commits from old operations will be in the evolog, except for revisions that are `jj abandon`-ed.

- I'm wondering about renaming `jj util gc` to `jj repo gc` (perhaps not now). Another operation for that subcommand I'm thinking of is `jj repo duplicate --ops ..<some_operation> some_dir/` operation that would copy some operations from jj repo into a new repo in a new dir. If nothing else, this would be nice for reporting bugs in jj, so that you can cheaply clone the state of the repo at the buggy op.

joyously:
> * Another operation for that subcommand I'm thinking of is `jj repo duplicate --ops ..<some_operation> some_dir/` operation

Would that be something like `jj git clone ../repo-dir --at-op <op ID>` ?

ilyagr:
> > * Another operation for that subcommand I'm thinking of is `jj repo duplicate --ops ..<some_operation> some_dir/` operation
> 
> Would that be something like `jj git clone ../repo-dir --at-op <op ID>` ?

No, `jj git clone` clones a *git* repo and `jj git clone --at-op` is not valid.

It's a good point that these operations would be close; the one I described could be called `jj repo clone`, but it'd only work locally.

Another possibility this brought to mind is to have a `jj repo export` command that would generate some sort of an archive (of possibly just a subset of the operations) that could be read by `jj repo import`. Though, unless one wants to get a subset of operations, `jj repo export` would be very similar to `tar`.

yuja (approved):

cli/src/commands/util/gc.rs
yuja: Looks good, thanks.

yuja:
> old evolog entries are never garbage collected. Should this be changed?

We'll surely need something, but I'm not sure how we can achieve that. Predecessor IDs are conceptually stored in commit metadata, so stripping them means the commit ID (should) changes. In Git, it doesn't, but that breaks the data model.

ilyagr:
Thanks!

Maybe commit ids in the evolog should not be assumed to always point to existing commits? It could be shown in the evolog as `(Missing commit)` or `(Garbage-collected commit)`.

yuja:
> Maybe commit ids in the evolog should not be assumed to always point to existing commits?

That can be an option, but I would want to pretty much avoid it (because we can no longer detect data corruption.)

ilyagr:
> > Maybe commit ids in the evolog should not be assumed to always point to existing commits?
> 
> That can be an option, but I would want to pretty much avoid it (because we can no longer detect data corruption.)

Is your "shallow boundary" idea something different?

> or predecessors shouldn't be a commit metadata

I actually assumed they weren't, not even conceptually. What do you mean when you say that they have to conceptually be in the commit metadata or it breaks the data model?

(Unless I'm very confused, technically they aren't part of the commit metadata, in the sense that they don't affect the commit hash; the git remote needs to compute the commit id and has no way to know about our evolog.)

cli/src/commands/util/gc.rs
```diff
@@ -24,6 +24,9 @@ use crate::command_error::CommandError;
 use crate::ui::Ui;
 
 /// Run backend-dependent garbage collection.
+///
+/// To garbage-collect old operations and the commits/objects referenced by
+/// then, run `jj op abandon ..<some old operation>` before `jj util gc`.
```
ilyagr, line 29: I'm now thinking of adding something like:

> At present, jj's garbage collection is only partially effective. Commits in the evolution log are not yet garbage collected.

WDYT?

ilyagr: Or maybe "If a commit is preserved by the garbage collection, objects reachable from its evolution log cannot be garbage collected."

yuja:
> > > Maybe commit ids in the evolog should not be assumed to always point to existing commits?
> > 
> > 
> > That can be an option, but I would want to pretty much avoid it (because we can no longer detect data corruption.)
> 
> Is your "shallow boundary" idea something different?

Yes. Shallow boundary is explicit.

> > or predecessors shouldn't be a commit metadata
> 
> I actually assumed they weren't, not even conceptually. What do you mean when you say that they have to conceptually be in the commit metadata or it breaks the data model?

It's stored in extras table, which is a commit metadata, and "conceptually" content-hashed in a way that there are no more than one `Commit` (Git commit object + extras) pointed to by one `CommitId` (at least locally.) That's the model. We don't store extras in the Git commit object just because we don't want make a Git commit created by jj different.

cli/src/commands/util/gc.rs
martinvonz: I prefer your second proposal.

ilyagr: I think I'll merge this as is and make another tiny PR, to make the discussion a bit more orderly.

PR #4959, opened 2024-11-24
jj-lib: abstract a lower-level api for fetching from git.

essiene:
* GitFetch{} separates `fetch()` from `import_refs()`.
* This allows more control over the stages of the fetch so multiple fetches
  can happen before `import_refs` resolves the changes in the local jj repo.
* Implement `git::fetch` in terms of the new api.
* Add a test case for initial fetch from a repo without `HEAD` set. This tests
  the default branch retrieving behaviour.

Issue: #4923

# Checklist

If applicable:

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -1229,6 +1229,162 @@ pub enum GitFetchError {
     InternalGitError(#[from] git2::Error),
 }
 
+fn fetch_options(
+    callbacks: RemoteCallbacks<'_>,
+    depth: Option<NonZeroU32>,
+) -> git2::FetchOptions<'_> {
+    let mut proxy_options = git2::ProxyOptions::new();
+    proxy_options.auto();
+
+    let mut fetch_options = git2::FetchOptions::new();
+    fetch_options.proxy_options(proxy_options);
+    fetch_options.remote_callbacks(callbacks.into_git());
+    if let Some(depth) = depth {
+        fetch_options.depth(depth.get().try_into().unwrap_or(i32::MAX));
+    }
+
+    fetch_options
+}
+
+struct FetchedBranches {
+    branches: Vec<StringPattern>,
+    remote: String,
+}
+
+struct GitFetch<'a> {
+    mut_repo: &'a mut MutableRepo,
+    git_repo: &'a git2::Repository,
+    git_settings: &'a GitSettings,
+    fetch_options: git2::FetchOptions<'a>,
+    fetched: Vec<FetchedBranches>,
+}
+
+impl<'a> GitFetch<'a> {
+    fn new(
+        mut_repo: &'a mut MutableRepo,
+        git_repo: &'a git2::Repository,
+        git_settings: &'a GitSettings,
+        fetch_options: git2::FetchOptions<'a>,
+    ) -> Self {
+        GitFetch {
+            mut_repo,
+            git_repo,
+            git_settings,
+            fetch_options,
+            fetched: vec![],
+        }
+    }
+
+    /// Perform a `git fetch` on the local git repo, updating the
+    /// remote-tracking branches in the git repo.
+    ///
+    /// Keeps track of the {branch_names, remote_name} pair the refs can be
+    /// subsequently imported into the `jj` repo by calling `import_refs()`.
+    fn fetch(
+        &mut self,
+        branch_names: &[StringPattern],
+        remote_name: &str,
+    ) -> Result<Option<String>, GitFetchError> {
+        let mut remote = self.git_repo.find_remote(remote_name).map_err(|err| {
+            if is_remote_not_found_err(&err) {
+                GitFetchError::NoSuchRemote(remote_name.to_string())
+            } else {
+                GitFetchError::InternalGitError(err)
+            }
+        })?;
+        // At this point, we are only updating Git's remote tracking branches, not the
+        // local branches.
+        let refspecs: Vec<_> = branch_names
+            .iter()
+            .map(|pattern| {
+                pattern
+                    .to_glob()
+                    .filter(
+                        /* This triggered by non-glob `*`s in addition to INVALID_REFSPEC_CHARS
+                         * because `to_glob()` escapes such `*`s as `[*]`. */
+                        |glob| !glob.contains(INVALID_REFSPEC_CHARS),
+                    )
+                    .map(|glob| format!("+refs/heads/{glob}:refs/remotes/{remote_name}/{glob}"))
+            })
+            .collect::<Option<_>>()
+            .ok_or(GitFetchError::InvalidBranchPattern)?;
+        if refspecs.is_empty() {
+            // Don't fall back to the base refspecs.
+            return Ok(None);
+        }
+
+        tracing::debug!("remote.download");
+        remote.download(&refspecs, Some(&mut self.fetch_options))?;
+        tracing::debug!("remote.prune");
+        remote.prune(None)?;
+        tracing::debug!("remote.update_tips");
+        remote.update_tips(
+            None,
+            git2::RemoteUpdateFlags::empty(),
+            git2::AutotagOption::Unspecified,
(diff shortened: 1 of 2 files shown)


lib/src/git.rs
```diff
@@ -1229,6 +1229,126 @@ pub enum GitFetchError {
...
+    ) -> Self {
+        GitFetch {
+            mut_repo,
+            git_repo,
+            git_settings,
+            fetch_options,
+        }
+    }
+
+    fn fetch_options(
```
yuja, line 1254: nit: make it a free function?

lib/src/git.rs
```diff
@@ -1239,6 +1359,30 @@ pub struct GitFetchStats {
...
+    let default_branch = git_fetch.fetch(branch_names, remote_name)?;
+    let import_stats = git_fetch.import_refs(branch_names, &[remote_name])?;
+    let stats = GitFetchStats {
+        default_branch,
+        import_stats,
+    };
+    Ok(stats)
+}
+
 pub fn fetch(
```
yuja, line 1386: Can you port `fetch()` within this patch so the new `GitFetch` iplementation is covered by tests?

lib/src/git.rs
```diff
@@ -1239,6 +1359,30 @@ pub struct GitFetchStats {
 }
 
 #[tracing::instrument(skip(mut_repo, git_repo, callbacks))]
+pub fn clone(
```
yuja, line 1362: nit: `clone()` sounds like it would create new repository. I can't think of a better function name, but it might be `fetch_<something>()` where `<something>` is e.g. `new_remote`.

lib/src/git.rs
```diff
@@ -1229,6 +1229,126 @@ pub enum GitFetchError {
...
+        tracing::debug!("remote.disconnect");
+        remote.disconnect()?;
+        Ok(default_branch)
+    }
+
+    pub fn import_refs(
+        &mut self,
+        branch_names: &[StringPattern],
+        remote_names: &[&str],
+    ) -> Result<GitImportStats, GitImportError> {
```
yuja, line 1334: It doesn't make sense that `import_refs()` can import arbitrary refs specified by caller. My original idea was:
1. `fetch()` pushes `(branch, remote)` to `fetched: Vec<_>`
2. `import_refs()` runs `import_some_refs()` for each `fetched` elements, then clears `fetched`.

lib/src/git.rs
essiene: Done.

lib/src/git.rs
essiene: I wasn't sure how to think about this initially, so thanks for confirming. I've updated this now. Even better, since it track {branches, remote} together, import refs maintains that mapping and doesn't do arbitrary permutations which the previous implementation was doing.

lib/src/git.rs
essiene: I went with `fetch_and_get_default_branch`, which is a mouthful!

However, I'm going ahead to simplify cli/src/commands/git/clone.rs and cli/src/commands/git/fetch.rs directly in the next commit(s). I'm going to get rid of these `git::fetch*` functions and use the new `git::GitFetch` API directly.

lib/src/git.rs
essiene: Done.

cli/src/commands/git/clone.rs
```diff
@@ -1,4 +1,3 @@
-// Copyright 2020-2023 The Jujutsu Authors
```
yuja, line 1: bad edit

cli/src/commands/git/clone.rs
```diff
@@ -222,7 +221,7 @@ fn do_git_clone(
     let mut fetch_tx = workspace_command.start_transaction();
 
     let stats = with_remote_git_callbacks(ui, None, |cb| {
-        git::fetch(
+        git::fetch_and_get_default_branch(
```
yuja, line 224: Perhaps, the `git::fetch()` function doesn't have to be renamed within this commit? `git::fetch()` can still return the default branch.

lib/src/git.rs
```diff
@@ -1229,6 +1229,146 @@ pub enum GitFetchError {
...
+        }
+    }
+
+    fn fetch(
+        &mut self,
+        branch_names: &[StringPattern],
+        remote_name: &str,
+    ) -> Result<Option<String>, GitFetchError> {
+        // Perform a `git fetch` on the local git repo, updating the remote-tracking
+        // branches in the git repo.
```
yuja, line 1284: nit: perhaps, this can be a doc comment for `fn fetch`

lib/src/git.rs
```diff
@@ -1229,6 +1229,146 @@ pub enum GitFetchError {
...
+                .filter_map(|fetched| {
+                    to_remote_branch(ref_name, &fetched.remote).map(|branch| {
+                        fetched
+                            .branches
+                            .iter()
+                            .any(|pattern| pattern.matches(branch))
+                    })
+                })
+                .next()
+                .unwrap_or(matches!(ref_name, RefName::Tag(_)))
```
yuja, line 1363: nit: It's probably better to do `match ref_name` first (instead of using `to_remote_branch()`. Something like:
```rust
match ref_name {
    RefName::LocalBranch(_) => false,
    RefName::RemoteBranch { branch, remote } => {
        self.fetched.get(remote).is_some_and(|patterns| ...)
    }
    RefName::Tag(_) => true,
}
```

lib/src/git.rs
```diff
@@ -1229,6 +1229,146 @@ pub enum GitFetchError {
...
+        })?;
+
+        self.fetched = vec![];
```
yuja, line 1366: nit: `self.fetched.clear()`

lib/src/git.rs
```diff
@@ -1229,6 +1229,146 @@ pub enum GitFetchError {
...
+        }
+        tracing::debug!("remote.disconnect");
+        remote.disconnect()?;
+        Ok(default_branch)
+    }
+
+    pub fn import_refs(&mut self) -> Result<GitImportStats, GitImportError> {
+        // Import the remote-tracking branches into the jj repo and update jj's
+        // local branches. We also import local tags since remote tags should have
+        // been merged by Git.
```
yuja, line 1349: nit: can be a doc comment of `import_refs()`?

lib/src/git.rs
```diff
@@ -1248,88 +1388,45 @@ pub fn fetch(
...
-    remote.prune(None)?;
-    tracing::debug!("remote.update_tips");
-    remote.update_tips(
-        None,
-        git2::RemoteUpdateFlags::empty(),
-        git2::AutotagOption::Unspecified,
-        None,
-    )?;
-    // TODO: We could make it optional to get the default branch since we only care
-    // about it on clone.
```
yuja, line 1302: nit: this TODO comment should be preserved

lib/tests/test_git.rs
```diff
@@ -2330,6 +2330,68 @@ fn test_fetch_initial_commit() {
...
+    let commit_target = RefTarget::normal(jj_id(&new_git_commit));
+    let commit_remote_ref = RemoteRef {
+        target: commit_target.clone(),
+        state: RemoteRefState::Tracking,
+    };
+    assert_eq!(
+        *view.git_refs(),
+        btreemap! {
+            "refs/remotes/origin/main".to_string() => commit_target.clone(),
+            "refs/tags/v1.0".to_string() => commit_target.clone(),
```
yuja, line 2378: nit: maybe we can omit uninteresting assertions and refs? I think here we just need to test `default_branch` and do some basic sanity checks.

yuja (approved):

lib/tests/test_git.rs
essiene: Makes sense, since they're already verified in the test above. I gone ahead to whack the asserts after verifying `default_branch`.

PR #4960, opened 2024-11-24
git: update `jj git clone|fetch` to use new GitFetch api directly.

essiene:
* Make the new `GitFetch` api public.
* Rewrite all tests that used `git::fetch*` to use the new API.
  One interesting artifact of the new API is that it holds onto a reference
  of the `tx.repo_mut()`, so the GitFetch object needs to fall go of scope
  completely before we can make any other mutable calls to `tx`.
* Delete the `git::fetch`
* Update `jj git clone` and `git_fetch` in `cli/src/git_utils.rs` to use
  the new api directly. Removing one redundant layer of indirection.
* This fixes https://github.com/martinvonz/jj/issues/4920 as it first fetches from all remotes before `import_refs()`
  is called, so there is no race condition if the same commit is treated
  differently in different remotes specified in the same command.

Fixes: https://github.com/martinvonz/jj/issues/4920

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,1063 +12,1072 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
-* Configuration variables are no longer "stringly" typed. For example, `true` is
-  not converted to a string `"true"`, and vice versa.
+-   Configuration variables are no longer "stringly" typed. For example, `true` is
+    not converted to a string `"true"`, and vice versa.
 
-* The following configuration variables are now parsed strictly:
-  `colors.<labels>`, `ui.color`, `ui.progress-indicator`, `ui.quiet`
+-   The following configuration variables are now parsed strictly:
+    `colors.<labels>`, `ui.color`, `ui.progress-indicator`, `ui.quiet`
 
-* `jj config list` now prints inline tables `{ key = value, .. }` literally.
-  Inner items of inline tables are no longer merged across configuration files.
-  See [the table syntax
-  documentation](docs/config.md#dotted-style-headings-and-inline-tables) for
-  details.
+-   `jj config list` now prints inline tables `{ key = value, .. }` literally.
+    Inner items of inline tables are no longer merged across configuration files.
+    See [the table syntax
+    documentation](docs/config.md#dotted-style-headings-and-inline-tables) for
+    details.
 
-* `jj config edit --user` now opens a file even if `$JJ_CONFIG` points to a
-  directory. If there are multiple config files, the command will fail.
+-   `jj config edit --user` now opens a file even if `$JJ_CONFIG` points to a
+    directory. If there are multiple config files, the command will fail.
 
-* The deprecated `[alias]` config section is no longer respected. Move command
-  aliases to the `[aliases]` section.
+-   The deprecated `[alias]` config section is no longer respected. Move command
+    aliases to the `[aliases]` section.
 
 ### Deprecations
 
-* The `Signature.username()` template method is deprecated for
-  `Signature().email().local()`.
+-   The `Signature.username()` template method is deprecated for
+    `Signature().email().local()`.
 
 ### New features
 
-* `jj` command no longer fails due to new working-copy files larger than the
-  `snapshot.max-new-file-size` config option. It will print a warning and large
-  files will be left untracked.
+-   `jj` command no longer fails due to new working-copy files larger than the
+    `snapshot.max-new-file-size` config option. It will print a warning and large
+    files will be left untracked.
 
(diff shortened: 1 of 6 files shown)


lib/src/git.rs
```diff
@@ -1383,88 +1383,29 @@ pub fn clone(
     Ok(stats)
 }
 
+#[tracing::instrument(skip(mut_repo, git_repo, callbacks))]
 pub fn fetch(
     mut_repo: &mut MutableRepo,
     git_repo: &git2::Repository,
-    remote_name: &str,
+    remote_names: &[&str],
```
yuja, line 1390: I feel it's a bit odd that `fetch()` takes multiple `remote_names`. (It doesn't make much sense to fetch the same `branch_names` from multiple remotes. It's kinda a limitation of the CLI.) So I think this can be moved to caller by using new `GitFetch` API.

essiene: I didn't think of that. Thanks.

I've gone ahead to directly use the new `GitFetch` API in both the `clone` and `fetch` commands. The multiple remotes is handled in `git_fetch` in the cli.

lib/src/git.rs
```diff
@@ -1394,31 +1394,6 @@ pub struct GitFetchStats {
...
-        fetch_options(callbacks, depth),
-    );
-    let default_branch = git_fetch.fetch(branch_names, remote_name)?;
-    let import_stats = git_fetch.import_refs()?;
-    let stats = GitFetchStats {
-        default_branch,
-        import_stats,
-    };
-    Ok(stats)
-}
```
yuja, line 1420: nit: I would copy this function to test_git.rs to minimize the change.

essiene: Good idea. Done.

CHANGELOG.md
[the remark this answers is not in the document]
essiene: I couldn't come up with anything better either, so `git` it is.

cli/src/commands/git/clone.rs
```diff
@@ -221,27 +222,34 @@ fn do_git_clone(
...
-        GitFetchError::InternalGitError(err) => map_git_error(err),
-        GitFetchError::InvalidBranchPattern => {
-            unreachable!("we didn't provide any globs")
-        }
+            &git_settings,
+            git::fetch_options(cb, depth),
+        );
+
+        let default_branch = git_fetch
+            .fetch(&[StringPattern::everything()], remote_name)
```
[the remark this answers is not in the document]
essiene: Hmmm... I see. Thanks. I've reversed the ordering.

martinvonz (commented):
> git: refactor jj git clone|fetch to use new GitFetch api directly.
> [...]
> This fixes https://github.com/martinvonz/jj/issues/4920 as it first fetches from all remotes before `import_refs()`

nit: I think "refactor" is supposed to mean that it has no impact on functionality, so just say "git: use new GitFetch api directly in `jj git clone|fetch`" or something instead

cli/src/git_util.rs
```diff
@@ -456,47 +457,49 @@ export or their "parent" bookmarks."#,
...
     remotes: &[String],
     branch: &[StringPattern],
 ) -> Result<(), CommandError> {
-    let git_settings = tx.settings().git_settings();
-
-    for remote in remotes {
-        let stats = with_remote_git_callbacks(ui, None, |cb| {
-            git::fetch(
+    let import_stats =
+        with_remote_git_callbacks(ui, None, |cb| -> Result<GitImportStats, CommandError> {
```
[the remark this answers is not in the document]
essiene: Good call. The callback is only initialized once when we cast from RemoteCallback into git2::RemoteCallback. I've gone ahead to rework this so that each `.fetch()` takes a callback and depth, so it doesn't break the progress indicator or any other thing. Shame though, as apart from the progress, other functionality would have been fine.

essiene:
> > git: refactor jj git clone|fetch to use new GitFetch api directly.
> > [...]
> > This fixes #4920 as it first fetches from all remotes before `import_refs()`
> 
> nit: I think "refactor" is supposed to mean that it has no impact on functionality, so just say "git: use new GitFetch api directly in `jj git clone|fetch`" or something instead

makes sense. Updated.

CHANGELOG.md
yuja, line 1: Restore formatting changes in CHANGELOG.md

yuja (commented):
Thanks.

bsdinis:
Hey @essiene! Super excited for this patch! 

Was wondering if you are having time to work on this, cause I'd be happy to pick this up

bsdinis:
Opened a new version of this PR (#5444), since the conflicts were a bit too hairy for me

emilazy:
Closing since https://github.com/jj-vcs/jj/pull/5498 was merged.

PR #4961, opened 2024-11-24
cli: diff: add `--exclude` flag

bryceberger:
First contribution, please let me know if this should be an issue / something else first.

Often, while iterating, I end up with a state where I've updated some dependencies (`cargo update`, etc) and some code. I want to inspect the diff, but `jj diff` (and `jj show`, etc) only accept an allowlist. This works ok when all the relevant changes are in a common directory, but is annoying when working in multiple (such as with workspaces).

This adds an `--exclude` flag to `jj diff`, which does what it says on the tin. `jj diff -x Cargo.lock`

Potential pitfalls: `jj diff -x *.lock` may not work as expected, as it could shell expand to something like `jj diff -x Cargo.lock flake.lock`, which would **ex**clude `Cargo.lock` but **in**clude `flake.lock`. Unsure if this motivates a section in the docs.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   "diff3" conflict style, meaning it is more likely to work with external tools,
   but it doesn't support conflicts with more than 2 sides.
 
+* `jj diff` now accepts a `-x` / `--exclude` flag (can be repeated), to remove
+  files from the view. Explicitly excluded files take precedence over included
+  files.
+
 ### Fixed bugs
 
 * `jj config unset <TABLE-NAME>` no longer removes a table (such as `[ui]`.)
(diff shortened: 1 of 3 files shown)


yuja:
I'm not sure if we want to add `--exclude` in addition to `PATH` patterns. Is it super inconvenient to specify `'~Cargo.lock'`?

ilyagr:
Another possible approach for this would be to have a `default-diff-filepattern` config.

As I wrote this, I also realized you might be able to instead do

```toml
[aliases]
diffc = ["diff", "~root-file:Cargo.lock"]
```

I also ran into this problem, and I think it's worth addressing, but I'm not sure which way is best.

One downside of the alias approach is that I already have a `jj difft` and `jj delta` aliases to pass the diff to difftastic/delta. It'd be annoying to create a "clean diff" variant for each alias.

bryceberger:
> I'm not sure if we want to add `--exclude` in addition to `PATH` patterns. Is it super inconvenient to specify `'~Cargo.lock'`?

No, it's not, just didn't realize that was an option, whoops.

> Another possible approach for this would be to have a `default-diff-filepattern` config.

It would (probably) be a more involved change, but if this route is taken it would be nice to have something like the template system. `jj diff -P no_nix` to expand to `jj diff '~root-path:(flake.lock | flake.nix)'`

ilyagr:
`[fileset-aliases]` sound like an interesting idea to me.

PR #4962, opened 2024-11-24
cli: diff: add fileset-alias option

bryceberger:
Following discussion in #4961.

The original problem was wanting to hide paths from the diff view. I failed to read the docs and realize that `~patterns` are available, and already solve that.

However, there are a few ergonomic issues with this:
- `~...` may be interpreted by the shell, so the excluded paths should be quoted
- Some paths are common to ignore, and annoying to type repeatedly (`Cargo.lock`, etc)
- Path is even longer if you're not at the root of the repository (`~root-file:Cargo.lock`)

The template system provides a nice solution for frequently used commands. It would be nice if there was a similar system for filesets.

# Example
```toml
[fileset-aliases]
no-lock = "~root-glob:*.lock"
```
```sh
jj diff -P no-lock
```

# Open questions:
- Should combining aliases and user-provided paths be allowed?

# Checklist

If applicable:


yuja:
The `fileset-aliases` idea is to substitute parsed tree of fileset (or path) expression. You can see how this is implemented in cli/src/template_parser.rs and lib/src/revset.rs. (It's a bit odd that alias substitution isn't made in revset_parser.rs, but that's just for a historical reason.) We'll need to write some plumbing code.

bryceberger:
Ah, so to have something like `jj diff no_lock` directly use the `no_lock` alias? In that case, what would happen if the user wants to interact with a file whose name is shadowed by an alias?

martinvonz:
> Ah, so to have something like `jj diff no_lock` directly use the `no_lock` alias? In that case, what would happen if the user wants to interact with a file whose name is shadowed by an alias?

Then they would probably have to escape it with e.g. `cwd:no_lock`

bryceberger:
Copied implementations from `fileset_parser.rs` until it worked. The implementation in 6b1e2d338f644def95f9b3de879ae91dce515a45 adds a `dsl_util::expand_aliases()` step to `fileset::parse()`.

Since this tries to expands all `ExpressionKind::Identifier`, `jj diff <path>` no longer always matches `<path>`, depending on user configuration. Workarounds:
- Quote the path: `jj diff '"path"'` (double quoted because of shell expansions)
- Add a file pattern: `jj diff file:path`

yuja:
> Since this tries to expands all `ExpressionKind::Identifier`, `jj diff <path>` no longer always matches `<path>`, depending on user configuration. Workarounds:

Correct. It's up to user to choose sane alias symbols such as `NO_LOCK` (assuming file names are usually lower case), `no_lock()` (function call), etc.

The change generally looks good to me, thanks.

lib/src/fileset.pest
```diff
@@ -89,3 +89,5 @@ program_or_bare_string = _{
         | bare_string_pattern ~ EOI
         | bare_string ~ EOI )
 }
+
+alias_declaration = _{ SOI ~ identifier ~ EOI }
```
yuja, line 93: It's probably better to use `strict_identifier` here, because `identifier` rule in fileset is lax.

We can also add function aliases, but that can be implemented by a separate patch.

lib/src/revset.rs
```diff
@@ -894,7 +895,9 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
             ));
         }
         let file_expressions = itertools::chain([arg], args)
-            .map(|arg| expect_fileset_expression(diagnostics, arg, ctx.path_converter))
+            .map(|arg| {
+                expect_fileset_expression(diagnostics, arg, ctx.path_converter, &Default::default())
```
yuja, line 899: You'll need to pass around the aliases map. (leave a TODO if it wasn't easy.)

lib/src/fileset.pest
bryceberger: > `strict_identifier`
Makes sense, implemented

bryceberger:
Implementation (of non-function) aliases should be completed, feel free to review under that assumption.

Todo:
- docs
- tests

cli/src/cli_util.rs
```diff
@@ -338,6 +341,14 @@ impl CommandHelper {
         load_template_aliases(ui, &self.data.layered_configs)
     }
 
+    /// Loads fileset aliases from the configs.
+    ///
+    /// For most commands that depend on a loaded repo, you should use
+    /// `WorkspaceCommandHelper::fileset_aliases_map()` instead.
+    fn load_fileset_aliases(&self, ui: &Ui) -> Result<FilesetAliasesMap, CommandError> {
+        load_fileset_aliases(ui, &self.data.layered_configs)
+    }
```
yuja, line 350: nit: just inline this because it's unlikely we'll need to parse file paths without loading a workspace.

lib/src/fileset.rs
```diff
@@ -473,9 +485,11 @@ pub fn parse(
     diagnostics: &mut FilesetDiagnostics,
     text: &str,
     path_converter: &RepoPathUiConverter,
+    aliases_map: &FilesetAliasesMap,
 ) -> FilesetParseResult<FilesetExpression> {
     let node = fileset_parser::parse_program(text)?;
     // TODO: add basic tree substitution pass to eliminate redundant expressions
+    let node = dsl_util::expand_aliases(node, aliases_map)?;
```
yuja, line 492: nit: in order to write parsing tests, it might be better to add `fileset_parser::parse(text, aliases_map)` (and make parse_program private.)

PhilipMetzger:
Closing this in favor of https://github.com/jj-vcs/jj/pull/8957 which was implemented a while ago.

PR #4964, opened 2024-11-24
cli `util gc`: document the fact that evolution log limits gc

ilyagr:
This is a followup to the discussion in
https://github.com/martinvonz/jj/pull/4953#discussion_r1855328098

### DIFF
--- a/cli/src/commands/util/gc.rs
+++ b/cli/src/commands/util/gc.rs
@@ -27,6 +27,9 @@ use crate::ui::Ui;
 ///
 /// To garbage-collect old operations and the commits/objects referenced by
 /// then, run `jj op abandon ..<some old operation>` before `jj util gc`.
+///
+/// Previous versions of a change that are reachable via the evolution log are
+/// not garbage-collected.
 #[derive(clap::Args, Clone, Debug)]
 pub struct UtilGcArgs {
     /// Time threshold
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -2268,6 +2268,8 @@ Run backend-dependent garbage collection.
 
 To garbage-collect old operations and the commits/objects referenced by then, run `jj op abandon ..<some old operation>` before `jj util gc`.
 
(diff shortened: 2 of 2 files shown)


cli/src/commands/util/gc.rs
```diff
@@ -27,6 +27,10 @@ use crate::ui::Ui;
 ///
 /// To garbage-collect old operations and the commits/objects referenced by
 /// then, run `jj op abandon ..<some old operation>` before `jj util gc`.
+///
+/// At present, jj's garbage collection is not comprehensive. If a commit is
```
martinvonz, line 31: I think that's the right thing to do so we don't leave dangling pointers, so I wouldn't call it "not comprehensive".  I'd just drop that sentence.

ilyagr: I'm not sure everyone will understand the motivation for second sentence on its own. How about something like:

> At present, jj's garbage collector may clean up less space than you'd expect. If a commit...

But, if you prefer, we can just go with the second sentence.

martinvonz: Maybe something like this?

> Note that previous versions of a change are reachable via the evolution log and are therefore not garbage-collected.

ilyagr: Sounds pretty good. I did edit it slightly, PTAL.

martinvonz: I prefer to not say "currently" because it seems to imply that it will change in the future and I'm not sure it will. Maybe we'll even remove the evolution log completely, but I feel like it's better to not say anything about it yet.

Also, I think there's an extra "that" in there.

ilyagr: OK, sure. I didn't realize this might be permanent, but if we're not sure, it is better to not say anything.

ilyagr:
Is this still blocked by something? It's fine if we consciously decide it needs more work, but otherwise it'd be nice to merge it before the next release.

martinvonz (approved):

martinvonz:
> Is this still blocked by something?

Not really. I just forgot about this PR, and I was on vacation for a few days too.

PR #4968, opened 2024-11-25
conflicts: remove use of Regex from conflict marker parsing

yuja:
# Checklist

If applicable:


scott2000:
I've done a similar refactoring for #4969, but using an enum instead to allow different lengths of conflict markers.

yuja:
> I've done a similar refactoring for #4969, but using an enum instead to allow different lengths of conflict markers.

Looks good, closing mine.

PR #4969, opened 2024-11-25
conflicts: escape conflict markers by making them longer

scott2000:
Related to #3975 and #4088.

# Checklist

If applicable:


joyously:
Perhaps this should be tested by correctly parsing its own source code.

scott2000:
> Perhaps this should be tested by correctly parsing its own source code.

Surprisingly, the source code for conflicts.rs doesn't actually contain any lines that look like conflict markers! The only two files that have lines which can be confused for conflict markers currently are tutorial.md and conflicts.md it looks like.

lib/src/conflicts.rs
scott2000, line 1: I'm not really sure it's necessary to disable the long markers for Git-style conflicts. It's actually simpler if we just allow long markers for all conflict marker styles in my opinion. Git technically does support longer conflict markers as well [through gitattributes](https://git-scm.com/docs/gitattributes#_conflict_marker_size), and it seems like some merge drivers could support [an argument for conflict marker size](https://git-scm.com/docs/gitattributes#_defining_a_custom_merge_driver) (using `%L` in Git). Instead of disallowing long markers for Git, we could maybe add a `"$markerLength"` argument to be used by `jj resolve` merge tools so that they produce conflicts that we can parse? But that would be a separate feature after this PR gets merged.

lib/src/conflicts.rs
```diff
@@ -250,6 +225,78 @@ pub enum ConflictMarkerStyle {
...
+
+/// Write a conflict marker to an output file.
+fn write_conflict_marker(
+    output: &mut dyn Write,
+    kind: ConflictMarkerKind,
+    suffix_text: &str,
+) -> io::Result<()> {
+    let conflict_marker: BString = iter::repeat(kind.to_byte())
+        .take(CONFLICT_MARKER_LEN)
+        .collect();
```
yuja: nit: `vec![kind.to_byte(); CONFLICT_MARKER_LEN]` also works

lib/src/conflicts.rs
```diff
@@ -480,24 +543,28 @@ pub fn parse_conflict(input: &[u8], num_sides: usize) -> Option<Vec<Merge<BStrin
...
-            } else if conflict_start.is_some() && line[0] == CONFLICT_END_LINE_CHAR {
-                let conflict_body = &input[conflict_start.unwrap() + conflict_start_len..pos];
-                let hunk = parse_conflict_hunk(conflict_body);
-                if hunk.num_sides() == num_sides {
-                    let resolved_slice = &input[resolved_start..conflict_start.unwrap()];
-                    if !resolved_slice.is_empty() {
-                        hunks.push(Merge::resolved(BString::from(resolved_slice)));
+            }
+            Some(ConflictMarkerKind::ConflictEnd) => {
+                if let Some(conflict_start_index) = conflict_start {
```
yuja, line 660: nit: you can do `conflict_start.take()` here to nullify it.

lib/src/conflicts.rs
```diff
@@ -50,8 +50,14 @@ use crate::merge::MergedTreeValue;
...
-/// Length of conflict markers.
-pub const CONFLICT_MARKER_LEN: usize = 7;
+/// Minimum length of conflict markers.
+pub const MIN_CONFLICT_MARKER_LEN: usize = 7;
+
+/// If a file already contains lines which look like conflict markers of length
+/// N, then the conflict markers we add will be of length (N + increment). This
+/// number is chosen to make the conflict markers noticeably longer than the
+/// existing markers.
+pub const CONFLICT_MARKER_LEN_INCREMENT: usize = 4;
```
yuja, line 60: nit: doesn't have to be `pub`?

lib/src/conflicts.rs
```diff
@@ -259,15 +265,23 @@ impl ConflictMarkerKind {
...
 fn write_conflict_marker(
     output: &mut dyn Write,
     kind: ConflictMarkerKind,
+    len: usize,
     suffix_text: &str,
 ) -> io::Result<()> {
-    let conflict_marker: BString = iter::repeat(kind.to_byte())
-        .take(CONFLICT_MARKER_LEN)
-        .collect();
+    debug_assert!(len >= MIN_CONFLICT_MARKER_LEN);
```
yuja, line 290: nit: I would make it `assert!()`, or remove the assertion. This expression is cheap. It's harmless to call this function with `len < MIN_CONFLICT_MARKER_LEN`.

lib/src/conflicts.rs
```diff
@@ -276,37 +290,80 @@ fn write_conflict_marker(
...
+    Some(ConflictMarker { kind, len })
+}
+
+/// Parse a conflict marker, expecting it to have a certain length. Any conflict
+/// markers which have a different length are ignored.
+fn parse_conflict_marker(line: &[u8], expected_len: usize) -> Option<ConflictMarkerKind> {
+    debug_assert!(expected_len >= MIN_CONFLICT_MARKER_LEN);
+
+    parse_conflict_marker_any_len(line)
+        .filter(|marker| marker.len == expected_len)
```
yuja, line 330: Maybe we need to accept `marker.len >= expected_len`? 

Suppose working-copy content was materialized with `N = 10`, and user removed a literal "<<<<<<<<<" (N = 9) without resolving any other conflicts, the new `expected_len` would drop. iirc, the working-copy content isn't updated to use the new `expected_len`.

lib/src/conflicts.rs
```diff
@@ -276,37 +290,80 @@ fn write_conflict_marker(
...
+/// marker length will be at least MIN_CONFLICT_MARKER_LENGTH, and is guaranteed
+/// to be longer than any existing markers in the file.
+pub fn choose_conflict_marker_len<T: AsRef<[u8]>>(single_hunk: &Merge<T>) -> usize {
+    let max_existing_marker_len = single_hunk
+        .iter()
+        .map(|term| find_max_conflict_marker_len(term.as_ref()))
+        .max()
+        .unwrap_or_default();
+
+    if max_existing_marker_len < MIN_CONFLICT_MARKER_LEN {
```
yuja, line 353: nit: perhaps, this should be `max(max_existing_marker_len + INCREMENT, MIN_LEN)`?

If existing literal were `<<<<<<`, marker of the length 7 would be hard to distinguish.

lib/src/conflicts.rs
```diff
@@ -276,37 +290,80 @@ fn write_conflict_marker(
...
     let merge_result = files::merge(single_hunk);
     match &merge_result {
-        MergeResult::Resolved(content) => output.write_all(content),
+        MergeResult::Resolved(content) => {
+            output.write_all(content)?;
+            Ok(None)
+        }
         MergeResult::Conflict(hunks) => {
-            materialize_conflict_hunks(hunks, conflict_marker_style, output)
+            let conflict_marker_len = choose_conflict_marker_len(single_hunk);
```
yuja, line 378: nit: I would move this to caller, and add a local function that materializes conflicts with the given marker length.

Most external callers aren't interested in the marker length.

lib/src/conflicts.rs
```diff
@@ -743,21 +843,46 @@ pub async fn update_from_content(
...
+
+    // Only accept the longest conflict markers in the file, also requiring that
+    // they are at least as long as the materialized conflict markers. We need to
+    // allow conflict markers to be longer than the materialized conflict markers,
+    // since the user could have removed the "fake" conflict markers from the file
+    // while editing, in which case their existing conflict markers will be
+    // considered longer than necessary the next time the file is snapshotted.
+    let expected_marker_len = materialized_marker_len
+        .unwrap_or_else(|| choose_conflict_marker_len(&merge_hunk))
+        .max(max_content_marker_len);
```
yuja, line 878: Should we trust the marker length of the new `content`? Doesn't it mean any existing conflicts can be invalidated if user added long `<<<<<<<<<<<<<` literal by mistake?

lib/src/conflicts.rs
yuja: > I'm not really sure it's necessary to disable the long markers for Git-style conflicts.

I wouldn't make it conditional, and simply let all styles follow the same rule.

lib/src/conflicts.rs
scott2000: That's what I had originally, but I think it's too permissive actually. If we just had a minimum, and didn't enforce an exact number, then this could happen:

1. Conflict materialized with length 12 markers since file contains length 8 marker
2. User deletes length 8 marker
3. Conflict is parsed expecting length >=12 markers successfully
4. User adds length 8 marker back, since they made a mistake while editing
5. Conflict is parsed expecting length >=7 markers now, so both the length 8 and length 12 markers are parsed, even though the length 8 marker isn't a real marker

To solve this, I made it so that only the longest length markers in a file are parsed. So the example would look like this:

1. Conflict materialized with length 12 markers since file contains length 8 marker
2. User deletes length 8 marker
3. Conflict is parsed expecting length 12 markers successfully
4. User adds length 8 marker back, since they made a mistake while editing
5. Conflict is parsed expecting length 12 markers (since 12 is the longest marker present, and 12 >= 7), so the length 8 markers are correctly ignored

lib/src/conflicts.rs
scott2000: Yeah, they would be invalidated if the user adds a marker with length greater than or equal to the expected marker length. I'm not sure we can work around that though, since there will always be some length of marker that's invalid for them to add.

My goal was to make it so that if a file initially had fake markers of length N when it was materialized, then the user can always add/remove as many markers as they want with length <= N without breaking parsing, even if multiple snapshot operations happen during the process. I think it makes sense that adding markers of length > N could interfere with conflict resolution, since it makes it ambiguous whether the marker is a real one or not.

lib/src/conflicts.rs
yuja: Hmm, we can't know whether the re-added marker is a literal or conflict, right? My feeling is that it's safer to assume that a parsable marker is a conflict rather than invalidating any other conflicts by single addition of marker-looking line.

If we prefer stricter behavior, we'll probably need to store the materialized marker length to `TreeState` on checkout.

scott2000: Yeah, that's a good point. Sorry for the long response; I was just trying to think through all the reasonable options there might be and summarize them. I think these are the variables we can use for setting a bound:

* R = length of real conflict markers in original materialization (would need to be stored in `TreeState`)
* N = length of longest fake conflict markers in last snapshot's `Merge<FileId>`
* K = increment added to longest fake marker to get materialized marker length
* C = length of longest conflict marker (real or fake) in the raw file contents being snapshotted

I think it's important that we have a simple rule that the user can follow while editing a file to ensure that the real conflict markers will continue to be parsed successfully. These are the 4 options I can think of, along with the expressions that we could use to determine if a marker is parsable:

<details>
<summary>Possible options</summary>

**Option 1:** "Don't add any new fake markers which are longer than any fake markers in the last snapshotted version of the file"

Either `len >= max(7, N+K)` or `len >= max(7, N+1)` would work here.

**Option 2:** "Don't add any new fake markers which are longer than any fake markers in the original materialized file"

I think `len == max(7, N+K, C)` matches this rule, which is what I currently have implemented. The main difference with this and option 1 is that the user is allowed to add back fake conflict markers that they removed, and they still won't be parsed.

**Option 3:** "Don't add any new fake markers with length greater than or equal to the markers that `jj` added"

I think `len >= R` is the obvious implementation of this rule. It does require adding it to the `TreeState` though.

I think `len == max(7, N+1, C)` would also fall into this category, since the user could still add any fake markers that are shorter than the original materialized markers, since that would keep the original materialized markers as being the longest in the file, so it would continue parsing them successfully. This is different from `len == max(7, N+K, C)`, because that rule would cause snapshotting to fail the next time after adding any fake markers of length `N+1` through `N+K-1`, since that would result in the real conflict markers becoming considered shorter than allowed in the following snapshot.

**Option 4:** "Don't add any new fake markers with the exact same length as the markers that `jj` added"

This is probably the easiest rule to follow for the user, and it is what we currently have in `jj 0.23.0`, except that the materialized length is always 7. This could be achieved with `len == R`, which also requires adding it to the `TreeState`.

</details>

---

If we don't store the materialized marker length in the `TreeState`, then I believe adding a fake marker which is longer than the real materialized markers will **always** cause the real markers to not be parsed correctly. With some implementations it might fail to parse in the current snapshot (because it picked the wrong markers to parse), or with other implementations it might parse successfully in the current snapshot, but then fail in the next snapshot (because the inferred minimum conflict marker length increased as a result of there being new fake conflict markers added), or it could parse successfully but just have the wrong content (since some fake markers were treated as real ones and happened to form a valid conflict).

With that in mind, I feel like these are the 3 best options, each with different tradeoffs:

* `len >= max(7, N+1)`, since it is the simplest to implement, and it handles all of the common cases reasonably well.
* `len == max(7, N+1, C)`, since it allows adding any fake markers that are shorter than the real markers, which the user might expect to work.
* `len == R` since it relies on the fewest assumptions (but it requires storing marker length in the `TreeState`).

I think I'll look into adding it to the `TreeState` to see how difficult it would be. If it's a small change, then probably that would be the best option?

yuja: > I think I'll look into adding it to the TreeState to see how difficult it would be. If it's a small change, then probably that would be the best option?

It wouldn't be difficult, but I'm not pretty sure if that's better because we're adding a new persistent state invisible to user. (BTW, the materialized conflict style could be stored in the TreeState if supported styles had ambiguous markers.)

@ilyagr, @martinvonz, any comments?

martinvonz (commented):
Sorry, I thought I had already sent these comments last week

lib/src/conflicts.rs
ilyagr: These links are a good find. I'd put them into a comment or at least in a commit description.

lib/src/conflicts.rs
scott2000: > That's a good point about the invisible state. I think that's enough to convince me that one of the options is better.

Yeah, I agree. It would also make the behavior less predictable for merge tools, since the checkout state is even more hidden there (since the user might not even realize there's any state at all).

> Maybe another option is to use option 1 but additionally require that all markers have the same length, or otherwise interpret the file as resolved. I don't strongly which of these three options we go with.

I hadn't thought about that, that's a good idea. I think it feels unintuitive if we allow mixing different marker lengths within the same file, because it doesn't look like it should work visually. Based on that, I feel like my favorite options are currently:

* Parse any markers which are at least length `max(max_existing_marker_len + 1, MIN_CONFLICT_MARKER_LEN)`, but treat the file as resolved if there are multiple lengths of marker which could be parsed.
* Parse only the longest set of markers in the file which are at least length `max(max_existing_marker_len + 1, MIN_CONFLICT_MARKER_LEN)`, and ignore any other markers which are shorter.

lib/src/conflicts.rs
```diff
@@ -250,6 +225,78 @@ pub enum ConflictMarkerStyle {
     Git,
 }
 
+/// Represents a kind of conflict marker which can be materialized and parsed.
+#[derive(Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+enum ConflictMarkerKind {
```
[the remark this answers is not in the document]
scott2000: I'm not sure how I feel about "separator". I personally have always thought of each individual line like "<<<<<<<" as being markers, rather than just being a part of a marker. Maybe `ConflictMarkerLine` and `ConflictMarkerLineChar` would be clearer? I don't feel strongly though.

lib/src/conflicts.rs
yuja: I prefer leaving conflict-looking content as conflict as possible. If the file were marked as resolved, user wouldn't notice that it had materialized conflict markers.

martinvonz: True, that's also a good point. Does that mean you prefer Scott's option 1? I'm fine with that option.

scott2000: > I prefer leaving conflict-looking content as conflict as possible. If the file were marked as resolved, user wouldn't notice that it had materialized conflict markers.

If I'm understanding correctly, you're suggesting we allow mixing conflict markers sizes, and just parse any markers with length `>= expected_marker_len`, right?

My main issue with this right now is that I believe it can result in a situation where doing `touch <file>` will cause the file to change from conflicted to resolved on the next snapshot. Like if the current conflict markers are length 12, and somebody adds a `-------` conflict marker of length 15 outside of any existing conflict hunks, then it would be parsed correctly the first snapshot (since the `-------` line is outside of any hunk, it would be treated as plain text), but then after `touch <file>`, since the length 15 marker was added to the file in the working copy commit's tree, any conflict markers of length <= 15 will be ignored, so the file will be considered resolved.

Maybe this could be solved by marking the file as resolved immediately if any conflict marker lines are encountered outside of conflict hunks? Like if we see any other markers in the loop that searches for `<<<<<<<` and `>>>>>>>`, then we could mark the file as resolved. We would also probably have to mark JJ-style conflicts as resolved if they contain Git-style markers and vice versa.

martinvonz: >  Like if the current conflict markers are length 12, and somebody adds a [...]

By the way, if you find time to add tests for cases like this one and the one you mentioned in https://github.com/martinvonz/jj/pull/4969#discussion_r1859717434, that would be very useful. Even if the behavior in those cases end up being something of a compromise, it's still useful to keep track of what the behavior is in a test case.

yuja: > Does that mean you prefer Scott's option 1?

Yes. 1 or 3. I prefer 1 because it's also simpler to implement.

> My main issue with this right now is that I believe it can result in a situation where doing touch <file> will cause the file to change from conflicted to resolved on the next snapshot.

That's also a good point. I just assumed it would be less likely to insert longer conflict marker than editing conflict marker by mistake.

scott2000: > EDIT: btw, `touch <file>` should be safe because we check unmodified content explicitly (which I think can be replaced with content hash to be stored in file state.)

I tested it, and it's actually not safe with my current implementation, since the current check is based on the re-materialized conflict markers, and the re-materialized conflict markers would be longer than the actual conflict markers in the file, causing them not to match. I think using a content hash for the check would be a good idea to prevent this case at least.

---

Overall, I think these properties are all desirable, but we can only pick at most 2 of them (unless there's some other solution I'm missing):
1. There should be no hidden state (i.e. we shouldn't store marker length in `TreeState`)
2. Making a change that doesn't touch any conflict markers (e.g. `touch file` or `echo line >> file`) should never cause a file to become resolved if it is currently conflicted
3. Making a single conflict hunk invalid (e.g. wrong number of sides, invalid markers in hunk, etc.) shouldn't cause the file to become resolved if there are still other valid conflict hunks in the file

If we have 1 and 3, then we can't have 2, because when one hunk becomes resolved but still contains markers, it would cause the file to contain new markers, which would cause the existing markers to stop parsing on the next snapshot (e.g. after `touch file` or `echo line >> file`). If we have 1 and 2, then we can't have 3, since the way that we guarantee 2 would be by making parsing more strict, which sacrifices 3. We can get 2 and 3, but it would require sacrificing 1 by adding hidden state so that we know which conflict markers to parse always.

If this is the case, which two properties do we want to have? I'm personally leaning towards picking 1 and 2, since it seems like it gives the most predictable behavior to the user, but I can also see 2 and 3 being good because I'm not sure that hidden state would be an issue in practice. I personally would rather not sacrifice 2, since it seems like it would be surprising to the user that making a change unrelated to conflict markers could cause the file to become considered resolved.

martinvonz: > * R = length of real conflict markers in original materialization (would need to be stored in `TreeState`)

Do we need to store it in the `TreeState`? Can we not get it from the previous content in the tree object? That might not help make the state visible to the user but it's nice if we avoid depending on `TreeState` for correctness.

yuja: > > * R = length of real conflict markers in original materialization (would need to be stored in `TreeState`)
> 
> Do we need to store it in the `TreeState`? Can we not get it from the previous content in the tree object?

Do you mean getting it from the tree of the working-copy parent? It might work, but I'm not sure.

FWIW, I think it makes sense to trust `TreeState` here because the materialization parameter is kinda working-copy state, not a state of the committed tree.

scott2000: I updated the PR to first implement parsing if `>= materialized_marker_len`, and then I added `conflict_marker_len` to the file state in another commit to see how it could work. If we decide not to add it to the file state, I can always drop the last commit from the PR.

lib/src/conflicts.rs
```diff
@@ -273,37 +287,73 @@ fn write_conflict_marker(
...
 }
 
 pub fn materialize_merge_result<T: AsRef<[u8]>>(
     single_hunk: &Merge<T>,
     conflict_marker_style: ConflictMarkerStyle,
     output: &mut dyn Write,
+) -> io::Result<()> {
+    let conflict_marker_len = choose_materialized_conflict_marker_len(single_hunk);
+
+    materialize_merge_result_with_marker_len(
```
yuja: nit: perhaps `match &merge._result { .. }` can be inlined here like other `materialize_*` functions?

lib/tests/test_conflicts.rs
```diff
@@ -1810,6 +1816,328 @@ fn test_update_conflict_from_content_simplified_conflict() {
...
+    "##
+    );
+}
+
+#[test]
+fn test_update_from_content_malformed_conflict() {
+    let test_repo = TestRepo::init();
+    let store = test_repo.repo.store();
+
+    // Create conflicts which contain conflict markers of varying lengths
```
yuja, line 2030: nit: copy-paste error?

lib/tests/test_conflicts.rs
```diff
@@ -1810,6 +1816,328 @@ fn test_update_conflict_from_content_simplified_conflict() {
...
+        %%%%%%% Changes from base to side #1
+        -line 4
+        +line 4 left
+        line 4 right
+        >>>>>>> Conflict 2 of 2 ends
+        line 5
+    "};
+    // On the first snapshot, it will parse as a conflict containing conflict
+    // markers as text
+    let new_conflict = parse(&conflict, new_conflict_contents.as_bytes());
```
yuja, line 2136: Indeed. This one is scary, and would probably provide a worse UX than the current state.

We'll have to either remember the materialized marker length or parse unpaired markers into some form of conflict.

cli/src/diff_util.rs
```diff
@@ -906,7 +907,12 @@ fn diff_content(
             executable: _,
         } => Ok(FileContent {
             is_binary: false,
-            contents: materialize_merge_result_to_bytes(&contents, conflict_marker_style).into(),
+            contents: materialize_merge_result_to_bytes(
+                &contents,
+                conflict_marker_style,
+                choose_materialized_conflict_marker_len(&contents),
```
yuja, line 913: nit: maybe this change is a merge/squash error? I don't think callers usually want to specify the marker length.

lib/src/conflicts.rs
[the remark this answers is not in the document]
yuja: I think `ConflictMarkerChar`, `LineChar`, `LineKind` are good, if the original concern was that `ConflictMarkerKind` sounds like a `ConflictMarkerStyle`.

lib/tests/test_conflicts.rs
```diff
@@ -1027,26 +1036,6 @@ fn test_parse_conflict_simple() {
     )
     "#
     );
-    // The conflict markers are too long and shouldn't parse (though we may
```
martinvonz, line 1030: Should we keep this or do we have new tests that cover the case of parsing a conflict that's longer than necessary for no good reason?

scott2000: I think it would be good to keep it; I'll add it back.

lib/src/conflicts.rs
scott2000: > I was thinking that it's good if `TreeState` is just a cache that can we can discard without impact other than in the corner case where a file has been modified in a way that the cache prevents us from detecting (currently if the mtime and size is preserved but e.g. inode is changed).

Currently, I have it set to default to parsing length 7 conflict markers if there's no materialized conflict data in the `TreeState`, so this should still work in many cases since it will parse all conflict markers of length `>= 7` if the `TreeState` is reset, and the conflict markers should usually be length 7 anyways.

If we want to make it possible to reset the `TreeState` in more cases, we could have the conflict marker length default to `choose_materialized_conflict_marker_len(current_tree_merge)` instead maybe? There is a tradeoff though, because doing this could break already-materialized conflicts from older versions of `jj` if a user upgrades and the new materialized markers are longer than the actual markers in the file (which would be length 7 always in older versions).

martinvonz: Sounds good to keep it simple for now.

lib/src/local_working_copy.rs
scott2000, line 1: Hey @martinvonz and @yuja, do you have any more comments regarding this PR? I think I've addressed the current comments unless I've missed anything. Do you think we should decide to go with the current approach of storing the materialized conflict marker length in the working copy state?

martinvonz: Yes, I think that seems like the least bad option. If you agrees, I think that means this PR is ready to merge.

lib/src/local_working_copy.rs
```diff
@@ -676,6 +706,7 @@ fn file_state(metadata: &Metadata) -> Option<FileState> {
             file_type,
             mtime,
             size,
+            materialized_conflict_data,
```
yuja, line 709: nit: maybe better to copy conflict data only if the new file type is a normal file? 

(but see also the other comment. this could be handled by `process_present_file()`.)

lib/src/local_working_copy.rs
```diff
@@ -1305,13 +1342,13 @@ impl FileSnapshotter<'_> {
...
+        mut new_file_state: FileState,
     ) -> Result<(), SnapshotError> {
         let update = self.get_updated_tree_value(
             &path,
             disk_path,
             maybe_current_file_state,
-            &new_file_state,
+            &mut new_file_state,
         )?;
         if let Some(tree_value) = update {
```
yuja, line 1353: I think `new_file_state` can be updated here instead of passing mutable ref around. Something like `if normal_file && (None | !is_resolved()) { new_file_state.materialized_conflict_data = .. }`.

The downstream functions shouldn't be responsible for the new `FileState`.

(We might have to rewrite some `current_file_state == new_file_state` to not compare uninteresting fields)

lib/src/local_working_copy.rs
```diff
@@ -1449,6 +1493,11 @@ impl FileSnapshotter<'_> {
                 repo_path,
                 &content,
                 self.conflict_marker_style,
+                new_file_state
```
yuja, line 1496: nit: I think the marker length can be obtained from the current file state, not from the new file state.

lib/src/local_working_copy.rs
yuja: We'll need to either make parsing strict about malformed markers (= somehow keep them as conflicts) or about the materialized marker length (= remember the length.) I have no idea which is better overall, but the latter would be easier to implement.

So I agree it's the least bad option.

yuja (approved):
The code generally looks good to me, thanks.

lib/src/local_working_copy.rs
scott2000: That's a good idea, thanks! I updated it based on your comments. Could you take another look and let me know if there's anything I should change?

lib/src/local_working_copy.rs
```diff
@@ -1391,7 +1434,13 @@ impl FileSnapshotter<'_> {
             };
             let new_tree_values = match new_file_type {
                 FileType::Normal { executable } => self
-                    .write_path_to_store(repo_path, disk_path, &current_tree_values, executable)
+                    .write_path_to_store(
+                        repo_path,
+                        disk_path,
+                        &current_tree_values,
+                        maybe_current_file_state,
```
yuja, line 1441: nit: at your option, the state could be mapped here to `Option<MaterializedConflictData>`, but I have no idea which is better overall.

lib/src/local_working_copy.rs
yuja: LG, thanks!

jonathantanmy:
I think that we should make `conflict_marker_len` in `conflicts::update_from_content` optional. I've started a discussion at #5243 , including my reasons.

PR #4981, opened 2024-11-26
merge_tools: add merge-tool-conflict-exit-codes option

scott2000:
Some Git merge drivers can partially resolve conflicts, leaving some conflict markers in the output file. In that case, they exit with a code between 1 and 127 to indicate that there was a conflict remaining in the file (since Git uses a shell to run the merge drivers, and shells often use exit codes above 127 for special meanings such as exiting due to a signal):

https://git-scm.com/docs/gitattributes#_defining_a_custom_merge_driver

We should support this convention as well, since it will allow many Git merge drivers to be used with Jujutsu, but we don't run our merge tools through a shell, so there is no reason to treat exit codes 1 through 127 specially. Instead, we let the user specify which exact exit codes should indicate conflicts. This is also better for cross-platform support, since Windows may use different exit codes than Linux for instance.

# Checklist

If applicable:


cli/src/merge_tools/external.rs
```diff
@@ -58,6 +58,11 @@ pub struct ExternalMergeTool {
     /// `$left`, `$right`, `$base`, and `$output` are replaced with
     /// paths to the corresponding files.
     pub merge_args: Vec<String>,
+    /// If false (default), any exit status from the merge tool other than 0
+    /// will cause `jj resolve` to fail. If true, then any status between 1 and
+    /// 127 can be used to indicate that the conflict was not fully resolved,
+    /// and the output file should contain conflict markers.
+    pub merge_tool_error_status_on_conflict: bool,
```
yuja, line 65: The choice `1 to 127` sounds a bit arbitrary. Maybe better to define a set or range of expected exit codes per tool? I assume merge drivers wouldn't use full range of status codes from 1 to 127.

The Git doc says "When the driver crashes (e.g. killed by SEGV), it is expected to exit with non-zero status that are higher than 128." This implies that the driver is executed through a system shell (i.e. `system()` instead of `execv()`.) If the driver is directly executed, SEGV wouldn't be mapped to an exit status code.

scott2000: I replaced it with an array of exit codes to indicate a conflict. Do you think that works better?

cli/src/merge_tools/external.rs
```diff
@@ -58,6 +58,14 @@ pub struct ExternalMergeTool {
...
     /// paths to the corresponding files.
     pub merge_args: Vec<String>,
+    /// By default, if a merge tool exits with a non-zero exit code, then the
+    /// merge will be cancelled. Some merge tools allow leaving some conflicts
+    /// unresolved, in which case they will be left as conflict markers in the
+    /// output file. In that case, the merge tool may exit with a non-zero exit
+    /// code to indicate that not all conflicts were resolved. Adding an exit
+    /// code to this array will tell `jj` to interpret that exit code as
+    /// indicating that the `$output` file should contain conflict markers.
+    pub merge_tool_conflict_exit_codes: Vec<i32>,
```
yuja, line 68: nit: the flag naming is already a bit inconsistent, but I think `merge_conflict_exit_codes` (without `tool`) is good. All parameters defined here are tool configuration.

cli/src/merge_tools/external.rs
```diff
@@ -246,6 +266,17 @@ pub fn run_mergetool_external(
...
         Merge::normal(new_file_id)
     };
+
+    // If the exit status indicated there should be conflict markers but there
+    // weren't any, it's likely that the tool generated invalid conflict markers, so
+    // we need to inform the user. If we didn't treat this as an error, the user
+    // might think the conflict was resolved successfully.
+    if exit_status_implies_conflict && new_file_ids.is_resolved() {
+        return Err(ConflictResolveError::ExternalTool(
+            ExternalToolError::InvalidConflictMarkers { exit_status },
```
yuja, line 276: nit: can you add a test for this case?

yuja (approved):
Thanks!

PR #4982, opened 2024-11-26
nix: fix usage of new parallel linker in Darwin devShell

thoughtpolice:
As reported by @lilyball on Discord, the recent nixpkgs update this week brought with it an entirely new Darwin stdenv, which had a behavioral change versus the prior one, causing builds inside the developer shell to fail, as the new super-duper parallel linker could not be used.

However, rather than giving up and throwing away *billions* of precious CPU cycles, @emilazy and @lilyball instead doubled down and diagnosed the issue. The result is a new impenetrable line of Nix code, which is fully explained by the comments within.

# Checklist

If applicable:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -60,7 +60,14 @@
         if pkgs.stdenv.isLinux then
           [ "-fuse-ld=mold" "-Wl,--compress-debug-sections=zstd" ]
         else if pkgs.stdenv.isDarwin then
-          [ "-fuse-ld=/usr/bin/ld" "-ld_new" ]
+          # on darwin, /usr/bin/ld actually looks at the environment variable
+          # $DEVELOPER_DIR, which is set by the nix stdenv, and if set,
+          # automatically uses it to route the `ld` invocation to the binary
+          # within. in the devShell though, that isn't what we want; it's
+          # functional, but Xcode's linker as of ~v15 (not yet open source)
+          # is ultra-fast and very shiny; it is enabled via -ld_new, and on by
+          # default as of v16+
+          [ "--ld-path=$(unset DEVELOPER_DIR; /usr/bin/xcrun --find ld)" "-ld_new" ]
         else
           [ ];
 


flake.nix
```diff
@@ -60,7 +60,14 @@
...
           [ "-fuse-ld=mold" "-Wl,--compress-debug-sections=zstd" ]
         else if pkgs.stdenv.isDarwin then
-          [ "-fuse-ld=/usr/bin/ld" "-ld_new" ]
+          # on darwin, /usr/bin/ld actually looks at the environment variable
+          # $DEVELOPER_DIR, which is set by the nix stdenv, and if set,
+          # automatically uses it to route the `ld` invocation to the binary
+          # within. in the devShell though, that isn't what we want; it's
+          # functional, but we actually want to use the newer Xcode linker
+          # instead (unfortunately not yet open source) which is ultra-fast and
+          # very shiny; it is enabled via -ld_new
```
emilazy, line 69: FWIW it is the default these days.

flake.nix
```diff
@@ -60,7 +60,14 @@
...
+          [ "-fuse-ld=ld" "--ld-path=$(unset DEVELOPER_DIR; /usr/bin/xcrun -find ld)" "-ld_new" ]
```
emilazy, line 70: The option is documented as `--find` so I’d suggest that syntax. (It also won’t be necessary to resort to `/usr/bin/xcrun` once the Nix `xcrun` fallback logic is refined a bit.)

emilazy: Also people can have Nix installed without other toolchains so I’d recommend `2>&1 || which ld` as a fallback here.

thoughtpolice: That might be a leftover error, but I'll check.

thoughtpolice: > Also people can have Nix installed without other toolchains so I’d recommend `2>/dev/null || command -v ld` as a fallback here.

Do we really expect people to have Nix but *not* Xcode *and* wanting to compile `jj`? I can make this change, but I'm surprised this is even a factor (then again maybe this was one of the million things motivating all the recent Darwin changes.)

joyously: > Do we really expect people to have Nix but _not Xcode and_ wanting to compile jj? 

What is Xcode? I've installed Nix before (not sure if it's still there), and I've wanted to compile `jj` (it compiles when I install from the crate).

flake.nix
thoughtpolice: I put a better comment here, lmk.

flake.nix
thoughtpolice: Xcode is the development environment used to develop things on macOS. This includes a big IDE, and tools for writing fancy and expensive iOS and macOS apps &mdash; but it also contains really important core tools like the system C++ compiler and the "SDK" used for many downstream projects so they can compile on Apple machines. You can get it from the App Store.

If you run something like `xcodebuild -version` in a terminal, does that work? I bet it will since you are using `cargo`; many tools will prompt you to install Xcode automatically if you don't have it, so it's possible you just installed it a long time ago and forgot :)

Also, I should have been more clear previously: this patch is *only* for compiling `jj` from source code using the Nix development shell *and* cloning the repository from GitHub. If you're not familiar with Nix, then you definitely aren't using it.

(I'm actually surprised Nix-on-Darwin can even *work* without using Xcode, though I guess all the recent stuff the project has been doing has been to support exactly these kinds of things.)

joyously: Ah, I'm not on an Apple machine...
but I have wanted to get into Nix, and I do compile `jj` sometimes.

emilazy: > Do we really expect people to have Nix but _not_ Xcode _and_ wanting to compile `jj`?

I know there’s at least some people out there who use Nix toolchains exclusively, which I admittedly consider kind of hardcore :)

emilazy (approved):
LGTM. We can skip the fallback since it wasn’t already present. (The better fallback would be `lld`, anyway; I might send another PR for that as I use it to get faster builds in my system flake.)

flake.nix
lilyball: > (Also is `-fuse-ld=ld` actually doing anything here?)

Yes it tells clang what the linker flavor is. In particular, it tells clang that it isn't lld.

PR #4986, opened 2024-11-27
config: replace LayeredConfigs with StackedConfig

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2773,7 +2773,7 @@ pub fn get_new_config_file_path(
         ConfigSource::User => command
             .config_env()
             .new_user_config_path()?
-            .ok_or_else(|| user_error("No repo config path found to edit"))?
+            .ok_or_else(|| user_error("No user config path found to edit"))?
             .to_owned(),
         ConfigSource::Repo => command.workspace_loader()?.repo_path().join("config.toml"),
         _ => {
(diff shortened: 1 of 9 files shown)


PhilipMetzger (approved):

cli/src/cli_util.rs
```diff
@@ -2767,15 +2767,16 @@ impl LogContentFormat {
 pub fn get_new_config_file_path(
     config_source: ConfigSource,
     command: &CommandHelper,
-) -> Result<PathBuf, CommandError> {
+) -> Result<&Path, CommandError> {
```
martinvonz, line 2772: Since it returns a value borrowed from `command`, do you think that should become the first argument? Should the function even be moved onto `CommandHelper` or maybe onto `ConfigEnv`? I'm leaning towards moving it onto `ConfigEnv`.

martinvonz (approved):

cli/src/cli_util.rs
yuja: Yeah, moving to `ConfigEnv` sounds good. It can also be moved to `commands::config::ConfigLevelArgs`. iirc, there are some google-internal use of this function, but I assume they'll just need `new_<user|repo>_config_path()`, not a mapping from `ConfigSource` to file path?

(I'll send a separate PR, so marked as resolved.)

martinvonz: Oh, good memory! I had completely forgotten that. We use it for writing a user-level config file based on their corp username and email.

PR #4988, opened 2024-11-27
Updated copy-tracing proposal to handle conflicts

martinvonz:
# Checklist

If applicable:

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209241374115289

### DIFF
--- a/docs/design/copy-tracking.md
+++ b/docs/design/copy-tracking.md
@@ -1,6 +1,6 @@
 # Copy Tracking and Tracing Design
 
-Authors: [Daniel Ploch](mailto:dploch@google.com)
+Authors: [Daniel Ploch](mailto:dploch@google.com), [Martin von Zweigbergk](mailto:martinvonz@google.com)
 
 **Summary:** This Document documents an approach to tracking and detecting copy
 information in jj repos, in a way that is compatible with both Git's detection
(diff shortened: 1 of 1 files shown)


martinvonz:
I'm getting tired of working on this doc, but I think it's good enough that it's reasonable clear how it will work. So please take a look if you're interested. If you spot problems, it's good to hear about them before I spend time polishing it.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+#### Example: Divergent copy and rename (best rename target)
+
+```
+N copy baz->qux
+|
+M rename bar->baz
+|
+| L rename foo->bar
+|/
+K add foo
```
yuja, line 283: nit: maybe something missing in `K` or between `K` and `M`? Since `L` renames to `bar`, `bar` doesn't exist in `L`'s parent `K`.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+If we use the file name as only input to the ID, then we get deterministic tree
+ids. On the other hand, if we add a salt to the copy graph node, then we can
+represent that a file was rewritten from scratch. That makes logical sense, but
+I'm not sure how useful it will be.
+
+Do we want to allow an edge in the copy graph to be able to say that changes
+should not be propagated across it? It might be useful to still have such edges
+for the log and annotation use cases.
+
+Should we support copy tracking for symlinks? Their history is not very for
```
torquestomp, line 190: "There history is not very for" seems incomplete

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+
+Propagating changes to copies means
+that rebasing a commit and then rebasing it back is no longer a no-op even when
+ignoring the "same-change rule". For example, if your commit modifies file `foo`
+and you rebase that commit onto a commit that copied `foo` to `bar`, and then
+you rebase it back, the same change will be applied twice to `foo`. However,
+thanks to the same-change rule, we won't consider it a conflict, so maybe it
+actually works well in practice thanks.
+
+For example:
```
torquestomp, line 344: These examples appear incomplete - what is their relation, do they even have one?

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+name: foo, id: K, copy_id: 1:foo
+
+Commit M:
+name: bar, id: M, copy_id: 4:bar
+name: baz, id: M, copy_id: 5:baz->1:foo
+```
+
+The `2:bar->1:foo` syntax above means that copy ID 2 has file `bar`, which was
+copied from copy ID `1`, where it was called `foo`.
+
```
ilyagr, line 250: Presentation: Perhaps I'd add something like:

> Let's consider the diff from `L` to `M`. This is the same as the diff of the commit `M2` we'd get by running `jj new L; jj bookmark create M2; jj restore --from M --to M2` (which would result in the commit M2 having the same tree as M).

docs/design/copy-tracking.md
ilyagr: Perhaps M could rename foo->baz?

ilyagr (commented):
I've read about half of the doc so far. It is very interesting; I think it's assembling into a sensible theory in my head (though my understanding is still slowly evolving).

There are a few things that I think could be clearer;  I think I did manage to answer most of my own questions though. I still point them out; feel free to tell me if you'd like me to work on clarifying them.It might be a good exercise. I might try to get to the end of the doc first, though.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+Diffing from `L` to `M` finds 2,3,4,5 as changed copy IDs. By walking their
+graphs, we find that 1,2, and 5 are related, while 3 and 4 are not.
+
+Because the `bar` and `baz` files have unrelated copy graphs, we break up their
```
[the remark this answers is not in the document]
ilyagr:
```mermaid
graph LR

    subgraph L["Commit L"]
        2["2:bar"]
        3["3:baz"]
    subgraph K["Commit K"]
        1["1:foo"]
    end
    end
    
    subgraph M["Commit M"]
        4["4:bar"]
        5["5:baz"]
    end
    
    2 --> 1
    5 --> 1
```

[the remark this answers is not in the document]
ilyagr: The type of diff (whether it's a rename, merge, or a split) seem to depend on which way the arrows in the copy ID DAG point, probably. These might get pretty complicated to describe, if you get something liken the following path between A and D in the DAG. I'm not sure what to do in these cases.

```
A <---- B ----> C <-----D
```

Another weird case: what if we care about the diff between A and D in

```
        B ----> C <---- D
A <---- B               D
        B <---- E --- ->D
```

Here, B and D are single nodes, each with one incoming and one outgoing node. This is still a DAG.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+name: foo, id: K, copy_id: 1:foo
+
+Commit L:
+name: bar, id: L, copy_id: 2:bar->1:foo
+name: baz, id: L, copy_id: 3:baz
+name: foo, id: K, copy_id: 1:foo
+
+Commit M:
+name: bar, id: M, copy_id: 4:bar
+name: baz, id: M, copy_id: 5:baz->1:foo
```
ilyagr, line 245: These arrows could be considered "backwards" by analogy to commits. `baz` has a pointer to `foo`, but this represents `foo` becoming `baz`. In the descriptions in the beginning of the section, the arrows go the other way.

My guess is that it's best to point the arrows the other way from the beginning (so, this arrow should be `foo`->`baz`, and the text should say that `baz` has a pointer to `foo`), but it's of course arguable.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+saying that `bar` was copied from `foo`.
+
+When diffing from `M` to `K`, we will instead present one record that says that
+`bar` was created, and one record that says that `bar` was merged into `foo`.
+
+### Merging
+
+When merging trees, we start by rewriting each diff to match any different names
+in the destination tree. For example, if the tree conflict is `A+(B-C)+(D-E)`,
+then we will rewtite the `(B-C)` diff and the `(D-E)` diff to the paths in `A`.
```
PhilipMetzger, line 314: Typo: "rewtite" => "rewrite"

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+
+### Log
+
+The copy graph contains all past paths and copy IDs of a file, so when doing
+`jj log <filename>`, we might want to translate that to a revset that's similar
+to `files()` but matches specific (path, copy ID) pairs instead of specific
+paths.
+
+### Annotate
+
```
PhilipMetzger, line 557: Incomplete

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+modify/delete conflicts in the first file and you will have to manually copy
+over the changes to the copied file.
+
+Propagating changes to copies means
+that rebasing a commit and then rebasing it back is no longer a no-op even when
+ignoring the "same-change rule". For example, if your commit modifies file `foo`
+and you rebase that commit onto a commit that copied `foo` to `bar`, and then
+you rebase it back, the same change will be applied twice to `foo`. However,
+thanks to the same-change rule, we won't consider it a conflict, so maybe it
+actually works well in practice thanks.
```
PhilipMetzger, line 342: unfinished sentence, probably should be "thanks to \<property\>".

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+paths.
+
+### Annotate
+
+
+### Representation in Git
+
+Do we ever want to record renames in the Git backend? If we do, we would
+presumably store it outside the Git object, similar to how we store the change
+id for commits.
```
PhilipMetzger, line 563: Some thoughts, it'd be nice to support precise renames and conflicts in a best effort basis. But since the bar to entry here already is quite steep, which you point out nicely in the next paragraph, I'd rather go for a set of special commands to deal with it in the Git backend, when we communicate with a native backend.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+
+What do we use for trees where we *don't* have any copy graph recorded? If we
+simply create a new copy graph based on the current path, then the caller will
+never find any copies. Do we need an indexing pass to detect all renames in a
+repo when running `jj git init`? That can be very expensive for large repos.
```
PhilipMetzger, line 568: I think doing it on `jj git init` or on repository upgrade will be a large regression to the users in the big repositories I mention in my next comment. So there could be a way to "chunk" the import with some kind of background job system which incrementally indexes the renames (which makes the bar to entry even higher, as I don't think it will be a small effort).

PhilipMetzger (commented):
I just have some minor thougths and some typo fixes, I also don't consider supporting Git and its backend worth it.

docs/design/copy-tracking.md
martinvonz: Yes, background-indexing is one possibility. I added a little paragraph about that.

docs/design/copy-tracking.md
martinvonz: I think if we're going to index copy information for Git repos anyway, then it doesn't seem like much extra work to make it possible for a user to modify that state themselves.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+
+One option might be for it to keep an index of all copy targets for a given copy
+source. Then it can look up the possible targets for the requested source files
+and then check which of them were actually copied from the source tree to the
+destination tree.
+
+A weakness of this solution is that the search gets expensive if there are very
+many related files. That's probably not much of a problem in practice. The
+server might want to populate the the index only for public/immutable commits.
+Otherwise, a user could poison the index by creating tons of copyies
```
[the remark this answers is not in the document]
martinvonz: Thanks, fixed.

Btw, you had a typo in your typo (you spelled it correctly) :)

martinvonz (commented):
I've addressed many of the comments. There are still many left to address.

FYI, @steadmon will be working on implementing the feature.

docs/design/copy-tracking.md
PhilipMetzger: > I think if we're going to index copy information for Git repos anyway, then it doesn't seem like much extra work to make it possible for a user to modify that state themselves.

Yes, that's a good point. I think I previously over-indexed on the best effort use-case.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,649 @@ results of rebasing across remote copies.
...
+
+Commit M:
+name: bar, id: M, copy_id: 4:bar
+name: baz, id: M, copy_id: 5:baz->1:foo
+```
+
+The `2:bar->1:foo` syntax above means that copy ID 2 has file `bar`, which was
+copied from copy ID `1`, where it was called `foo`.
+
+Diffing from `L` to `M` finds 2,3,4,5 as changed copy IDs. By walking their
```
steadmon, line 261: I think I got tripped up over something silly here. You say "Diffing from L to M", which makes me think L is the source, and M is the destination. However, later you say "foo does not exist on the source side", which implies the opposite.

martinvonz (commented):
Sorry it took so long for me to update this. I've been pretty burned out on this topic, but now @steadmon and @jonathantanmy are starting to work on implementing copy tracking, so I needed to get this updated.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,613 @@ results of rebasing across remote copies.
...
+Do we ever want to record renames in the Git backend? If we do, we would
+presumably store it outside the Git object, similar to how we store the change
+id for commits.
+
+What do we use for trees where we *don't* have any copy graph recorded? If we
+simply create a new copy graph based on the current path, then the caller will
+never find any copies. Do we need an indexing pass to detect all renames in a
+repo when running `jj git init`? That can be very expensive for large repos.
+For reference, `git log --summary --find-copies-harder` takes about 165 seconds
+in the git.git repo on my computer.
```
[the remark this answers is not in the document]
martinvonz: I'm not sure it will help much but I'm running that command in the Nixpkgs repo now anyway. I'll try to remember to check on it in a few hours so I can update this paragraph.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,686 @@ results of rebasing across remote copies.
...
+For example:
+```
+M set foo="bye"
+|
+| L rename foo->bar
+|/
+K add foo="hello"
+```
+
+When rebasing `L` onto `M`,
```
drieber, line 399: It looks like you did not finish this sentence: When rebasing `L` onto `M`, .... ?

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,686 @@ results of rebasing across remote copies.
...
+
+```
+M foo="M"
+|
+| L copy foo->bar
+|/
+K add foo="K"
+```
+
+Let's say we rebase `M` onto `L`. If we decide to propagate changes to copies,
```
drieber, line 445: I am not getting this example. Should this say "let's say we rebase L onto M" ?

ilyagr:
>  He convinced me that we should not automatically propagate changes to copies and instead leave them conflicted.

TLDR: I think this relates closely to my general worries about tracking copies, and that having a satisfying solution here might not be practical. Do you have something in mind?

Here is how it arranges in my mind.

-----------------------

I think this might be a very different kind of conflict to our usual content conflicts with two sides and a base. If you treated it like a normal conflict with unchanged tree as base, change to first copy as one side, and change to the other copy as the second side, it would get simplified immediately and both copies would be modified.

So, either we'd have to force the user to resolve the conflict immediately (like Git does for content conflicts), or we'd need to learn to store a whole new kind of conflicts in the repo.

One option that comes to mind and seems natural is, in a `Tree`, for every path, in addition to its blob, we could store an "annotation" that is a set of patches that may or may not need to be applied to this path. Then, if you have

```
  change f
A---------->B
\  copy f to g and h
 \-------------------->X
```

and you merge `X` and `B` (or rebase `X` onto `B`), you'd get an unconflicted `Merge<Tree>` (that is, only one tree), where both `g` and `h` (and maybe `f` as well) in this tree would have their content from A, but both of them would also have the annotation that the patch "change f" might need to be applied to it. When presenting this to the user, we *would* present it as a conflict.

I'm not fully sure whether there's a sensible algebra for these kind of annotations, but hopefully? I am apprehensive that it might require building a whole new patch-based VCS on top of `jj`, since these "annotations" might need to compose. E.g. if you have `A --> B --> C` where both B and C change `f`, rebasing X onto C should result in the same annotations as if you rebased it onto B first, and then rebased the result onto C.

------

I also wonder what "unsatisfying" but less involved solutions to this would look like. One option might be to restrict what people can do with revisions that have this kind of conflicts, e.g. forbid rebasing them in ways that would affect the problematic files.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,686 @@ results of rebasing across remote copies.
...
+
+For example:
+```
+M set foo="bye"
+|
+| L rename foo->bar
+|/
+K add foo="hello"
+```
+
```
ilyagr, line 398: This example has nothing to do with the two paragraphs about propagating changes to copies, right? I think it should be moved before those paragraphs, or (perhaps better) those two paragraphs should be moved to a separate subsection together with " Example: Propagate changes to copied file, then rebase back" below.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,686 @@ results of rebasing across remote copies.
...
+| M rename bar->bar2, modify baz
+| |
+| | L rename foo->foo2, modify bar
+| |/
+|/
+K add foo, bar, baz
+$ jj new L M N
+```
+
+#### Example: Propagate changes to copied file, then rebase back
```
ilyagr, line 435: Here's a similar example that I find useful (though I don't have a resolution).

Let's start with

```
M foo="M" bar="K"
|
L copy foo->bar
|
K add foo="K"
```

Now, let's rebase M onto K to get `M'`.

In our current conflict theory, there is literally no value we could assign to `M'` that would have the property that rebasing `M'` back onto `L` would result in `M`.

For example, let's say we want `foo="M"` in `M'`. Then, rebasing `M'` back onto `L` would result in a different commit that `M`, one where both `foo` and `bar` have value "M". To illustrate all of these commits together:


```
M foo="M" bar="K"
|
| M'' foo="M" bar="M"
|/
L copy foo->bar
|
| M' foo = "M"
|/
K add foo="K"
```

ilyagr (commented):
I literally can't find one of the very old pending comments I'm about to post in the GitHub UI. I'll delete or edit it if I can find it later. The other two are relevant, though.

docs/design/copy-tracking.md
ilyagr: Ah, I hadn't realized the comment I wrote in Jan was never sent, sorry.

docs/design/copy-tracking.md
ilyagr: > I think the proposed data model has enough information that we could decide to consider it a conflict if we wanted to. That would be done by simply recording a conflict between the three trees.

> So that would mean that M'' would be left in a conflicted state.

I think we should flesh out what kind of conflict you mean here. As I mentioned in https://github.com/jj-vcs/jj/pull/4988#issuecomment-2837322963, I'm not sure the kind of conflicted trees we are used to are appropriate here, at least as long as we follow the usual conflict simplification rules.

To recap, the issue as I see it is that if we express a conflict (for one of the paths) as a `Merge<Blob>` the way I imagine, the base would be the same as one of the sides, so it'd get simplified immediately. For `Merge<Tree>`s, it'd get simplified immediately as long as we allow simplifying those one path at a time.

I'm not saying it's hopeless; I'm still wondering whether we can get away with only changing "a little bit" of our conflict theory or conflict simplification rules. (And perhaps you also have some idea in mind.) As I explained in the previous comment, I'm thinking of annotating individual paths with patches that "may or may not need to be applied", but I'm now more optimistic that this might be feasible with small changes to our theory.

Let me know what you are thinking.

martinvonz: I mean that we would leave the trees completely unchanged and then whenever you look at that commit in the future, we redo the copy tracking. So in the scenario above, we might have trees like this:

```
K:
  name: "foo", content: "K", copy_id: 1,

L:
  name: "bar", content: "K", copy_id: 2
  name: "foo", content: "K", copy_id: 1

M:
  name: "bar", content: "K", copy_id: 2
  name: "foo", content: "M", copy_id: 1

M':
  name: "foo", content: "M", copy_id: 1
```

The copy IDs would be these:
```
1:
  name: "foo", parents: []

2:
  name:  "bar", parents: [1]
```

When we then rebase `M'` back onto `L`, we see that `foo` was modified in the commit we're rebasing, so we look at its copy history. It has copy id 1, which is related to copy id 2, and copy id 2 appears in the destination `L`. Since `foo` also still exists in `L` and the copy id matches, we consider `bar` a copy. And, since we said that we should not automatically propagate changes to copies, we stop at this point. Next time we inspect the commit with `jj status` etc, we redo the conflict resolution and come up with the same conclusion, namely that there are changes to `foo` that should possibly be propagated to `bar`. We hint to the user that they should run `jj resolve` to resolve that conflict. We'll also need to also modify things like `jj diff` to include information about copy conflicts (and divergent renames etc.).

Does that make sense?

ilyagr: Less technically, I'll elaborate on what I think is good about the UI you are proposing, and what we might want to watch out for (to the degree we can).

### The problem and the goal

From the example in the beginning on this thread, it seems that asking for a rebasing back-and-forth cycle to be a no-op in the presence of copy tracking is just too much to ask for, at least unless we come up with some breakthrough VCS conceptual model.

This means that rebasing A to X and then to Y will, in general, be different than rebasing A straight to Y.

If that's the case, the goal in my mind is that the user should have some mental model of what they could get if they rebase A to Y indirectly (with intermediate steps) and the lower the number of possibilities, the better.

#### Example

One example where this is important is that this indirect rebasing will be happening if the user has some tree 

```
X->maybe many commits->A
\-->@
```

If the user edits `@` and keeps squashing changes into `X` repeatedly, `A` will be rebased over and over again, and might end up with a different contents than if the user edited `@` once and squashed into `X` once.

If there was no limit to how complex `A` could get or what it could become depending on the intermediate squashes, that would not be great.

(the "maybe many commits" in between `A` and `X` aren't essential to the example, they just illustrate that A could be far away from anything the user is actively thinking about as they edit `@` and `X`)

### Why conflicts might help

I'm hoping that this will still be manageable if one of these operations that should-be-equivalent-but-aren't results in a conflict (in the UI sense, not necessarily in the `Merge<Tree>` sense) for which the other result is a natural resolution (in some sense that is not perfectly clear to me yet, perhaps the other result is one of the sides in the conflict).

For it to stay manageable, my hope is also that we could keep the complexity of this conflict from growing too much. Ideally, and informally, if we rebased A `n` times to different places and then rebased it to Y, the complexity of the conflict would only increase a bounded number of times and would not keep growing with `n`. For example, perhaps we could limit this complexity to the number of times files in A's diff are copied between A and Y. (This could also be hard, if each of the `n` intermediate commits copies files in A's diff, but it's something to aim for)

drieber: I have been thinking about copy/rename tracking for just a couple of weeks so my reasoning is perhaps basic. At a high level I see this way: in today's jj a file has a name and content, and the content may be in conflict. With copy tracking we are changing jj: files not have a name, content and "CopyId" (aside: perhaps we can pick a better name, see below). The way I see, the model should allow a file's CopyId to be in a conflicted state. I think it should be easy to represent this state (CopyId is in conflict). It is up to the user to resolve that conflict, perhaps directly with `jj resolve`, perhaps by doing `undo`, perhaps by rebasing back...

Can a file have both a content conflict AND a CopyId conflict? Maybe... But if that is the case maybe the user has to first resolve the CopyId conflict and then the content conflict? Or not.... I don't know.

I hope some of this makes sense.

ilyagr: > When we then rebase `M'` back onto L (to get `M''` - Ilya) ... since we said that we should not automatically propagate changes to copies, we stop at this point.

The question in my mind: how do we think of `M''` at this point? How do we store it? What do we do if, instead of running `jj status`, the user rebases it soemwhere else instead, or rebases other commits on top of it?

I think we need a good model for a commit in this "transient" state, similarly to how we have a model for regular content conflicts.

As in the example from the last comment, this new rebase could happen because the user is actually editing a distant ancestor of `M''`, so interrupting their work and forcing them to think about `M''` might not be very welcome.

--------------

@drieber Absolutely, conflicted copy ids might be a good way of thinking about it. This is certainly the kind of model that would address the question in my mind.

The devil is in the details: how do conflicted copy ids interact with conflicted contents? Is it clear how the conflicted copy ids should behave when the commit containing them is rebased before the conflict is resolved? It's possible some of these questions are already figured out, or are already in this doc, or seem obvious to somebody who have thought about the copy-tracking ids more than I have.

The model I vaguely proposed above is different. I think I proposed it because it seemed a bit clearer to me how to resolve these questions for it. This doesn't mean it's necessarily better, in fact it would probably be more limited than the "conflicted copy id" model (if it can be made to work).

Also very vaguely, my worry about the "conflicted copy id" model is that it might require building a whole patch-based VCS on top of jj's model. This is a hunch, I'm not going to try to make it precise right now, and this worry might go away once I think about it more.

martinvonz: > The question in my mind: how do we think of `M''` at this point?

I think of it as a commit with three trees like any other conflicted commit. So `M''` would be `L+(M'-K)`. Again, since resolving that tree involves propagating changes across a copy, we don't do it automatically. So if we then rebase it back onto `K`, we get `M''' = K+(M''-L) = K+((L+(M'-K))-L) = K+(M'-K) = M'`.

ilyagr: Thanks! I think I now have a picture of what you are thinking. It's interesting, and seems to have a lot of the nice properties I wanted.

Expanding your explanation, following your comment from far above, we have trees:

```
K:
  name: "foo", content: "K", copy_id: 1,

L:
  name: "bar", content: "K", copy_id: 2
  name: "foo", content: "K", copy_id: 1

M':
  name: "foo", content: "M", copy_id: 1
```

These are the values for the conflict `L+(M'-K)`. These come with the copy ID graph:

```
1:
  name: "foo", parents: []

2:
  name:  "bar", parents: [1]
```

Quoting you some more,

> `foo` has copy id 1, which is related to copy id 2, and copy id 2 appears in the destination L. Since foo also still exists in L and the copy id matches, we consider bar a copy. And, since we said that we should not automatically propagate changes to copies, we stop at this point.

In other words, based on the copy id graph and this deduction, we leave `L+(M'-K)` unsimplified. (Before copy tracking, or if we forgot the copy-tracking data, we'd try to simplify the conflict per-file-path and every contents conflict for both "foo" and "bar" would be simplified to `foo="M"` and `bar="K"`; it's also instructive to go through what would have happened if `L` renamed "foo" to "baz" in addition to copying "foo" to "bar").

joyously:
> He convinced me that we should not automatically propagate changes to copies and instead leave them conflicted. Then it becomes a simple yes/no question for the user whether to propagate the changes (`jj resolve` can prompt the user).

I'm struggling to see how the user would know whether to propagate the changes, especially if there is a lot of history. Unless there are restrictions on copies, I think you have to base the design on the most complicated scenarios, like circular references (copy one to two, make a change, copy two to one, rebase any of those), or a series of copies (copy one to two, copy two to three, copy three to four...). 

The more I think about it, the more I think a command to copy files doesn't belong in `jj`. I can't think of one benefit of having it, and it certainly makes a lot of complications. I say leave the editing and file commands to external programs, and let the VCS record the snapshots. (I think of `fix` this way also.)

docs/design/copy-tracking.md
martinvonz: I think this is obsolete now with the change to say that we don't automatically propagate changes to copies

docs/design/copy-tracking.md
martinvonz: Done.

martinvonz:
@ilyagr: I've made the updates to say that we don't automatically propagate changes to copies. Do you think this is good enough to merge? I would very much like it to be merged because I'm tired to updating it :) Once merged, it's going to be easier for e.g. @steadmon and @jonathantanmy to make further edits.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,671 @@ results of rebasing across remote copies.
...
+| L copy foo->bar
+|/
+K add foo="K"
+```
+
+Let's say we rebase `M` onto `L`. Since we decided to not automatically
+propagate changes to copies, we will leave the `M+(L-K)` tree unresolved (i.e.
+without making any changes to the three trees). If the user does not resolve
+the conflict, and instead rebases `L` back onto `K`, the conflict will be
+resolved automatically per the usual conflict simplification.
```
ilyagr, line 453: I don't think this is urgent, but I think we should eventually add more details here, explaining the conflict with copy ids that you get, and how we decide to keep it a conflict. For example, it is critical that we are thinking of conflicts as conflicts of trees for this to work, IIRC.

There is already a "Conflicts" section below, but it feels a bit out of date and doesn't discuss this.

We could add a TODO and/or the example from the above thread and/or a link to that thread. 

But, in time, this might need a whole new section rather than a quick fix, even just for the parts of the algorithm that we seem to have some handle on (based on that thread). As I said in some other comments, I think there are also open questions we'd have to eventually decide on as well.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,671 @@ results of rebasing across remote copies.
...
+K add foo="K"
+```
+
+Let's say we rebase `M` onto `N`. The changes to `foo`, `foo2`, and` foo3` will
+then all apply to `foo`, which means we get a 4-sided conflict.
+
+#### Example: Convergent renames
+
+Consider this "convergent copy/rename" scenario:
+```
```
ilyagr, line 473: Minor: It's a bit too easy to miss `jj new B C` here. This will look clearer in some renderings if you use the "console" syntax highlighting here. Perhaps also add a new line before `$ jj new`?

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,671 @@ results of rebasing across remote copies.
...
+C | rename bar->baz
+| |
+| B rename foo->baz
+|/
+A add foo="foo" and bar="bar"
+$ jj rebase -r E -d C
+$ jj rebase new D E -m F
+```
+
+If F is empty (auto-merged), it should have the same state as E before.
```
ilyagr, line 605: (No action needed) This sounds very nice.

docs/design/copy-tracking.md
```diff
@@ -176,17 +833,29 @@ A well known and thorny problem in Mercurial occurs in the following scenario:
...
+    rename lives ephemerally in memory and has no persistence after the rebase
+    completes.
+1.  A to-be-determined algorithm diffs the copy records between [D]->[B] and
+    [D]->[C] in order to make changes to the rebased commit. This results in
+    edits to renamed files being propagated to those renamed files, and avoiding
+    conflicts on the deletion of their sources. A copy/move may also be undone
+    in this way; abandoning a commit which renames A->B should move all
+    descendant edits of B back into A.
 
 ### Conflicts
```
ilyagr, line 860: Is this section still relevant? It feels a bit out of date.

I'd either delete it or say explicitly that this section doesn't discuss the conflicts from copy-tracking + content changes. (For example, you could add a heading for that and mark it as TODO).

ilyagr (approved):
Overall, I think this seems feasible. It's an impressive idea!

Apart from a few copy-editing comments, I keep coming back to two non-blocking issues for later:

- we should at some point transfer some of what we discussed in the long thread about copy-tracking conflicts into the docs. 
- there are still a bunch of open (at least in my mind) questions about how the details of that will interact. It might be nice to make a list. They all seem to be issues we can experiment with and change later.

ilyagr:
> I'm struggling to see how the user would know whether to propagate the changes, especially if there is a lot of history

I'm also a bit worried about this, but I think we'll have some freedom to experiment with what cases, exactly, we call a conflict.


-------

Non-actionable: Thinking about this leads me to a philosophical issue that bothers me a bit, where a user might want to be notified about potential copies even if there is no rebase going on. I wonder whether a VCS could or should help with that, it's unclear.

Let's say I'm working of a bugfix to file `foo` in commit A while somebody from another department of my company copied `foo` to `bar` in their codebase (which I'm unaware of) in commit X.

**Case 1, good:** Now, if I started and finished my work before `X` happened, the other department will get the bugfix.

**Case 2, good:** If I started my work before `X` happened and merged to `main` after `X` happened, I would pressumably get a conflict when I rebase `A` onto `main`. This seems good, I can reason about whether the other department needs my bugfix.

**Case 3, not so good:** However, if I started my work after `X` happened, there is no rebase and there is no conflict, the other department just doesn't get my bugfix at all.

Is there some way a VCS could help and notify me about potential other files I'd want to clone my changes to?

martinvonz:
Thanks for reviewing! I'll try to update the doc within the next few days.

> Is there some way a VCS could help and notify me about potential other files I'd want to clone my changes to?

It shouldn't be hard to provide a feature which tells you which files have been copied from a given file. We could also provide a `jj recopy` command to copy over the change that happened to the source file since it was copied into the destination file. However, the model proposed in this document doesn't provide a way to record that the destination is now in sync with the source file again, so we cannot figure out what a *subsequent* `jj recopy` should do. For that to work, I suppose we would have to record not only the copy source in the copy graph node but also the `FileId` of the copy source. I have not thought about what the consequences of that would be.

joyously:
> **Case 2, good:** If I started my work before `X` happened and merged to `main` after `X` happened, I would presumably get a conflict when I rebase `A` onto `main`. This seems good, I can reason about whether the other department needs my bugfix.

I don't agree that you could reason what the other department needs. I was reading this doc to mean that the copy was traced, not the original. Perhaps this is semantics, but just because someone copied my file shouldn't obligate me to fix their subsequent problems. Rebasing `A` shouldn't be a conflict. I can see how the person copying might want to know, but I think there are more cases when they wouldn't want to know ( a folder of templates everyone starts from). 

I'm still waiting to hear a case where tracing a copy is a thing the VCS should do.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,671 @@ results of rebasing across remote copies.
...
+|/
+K add foo="K"
+```
+
+When rebasing `M` onto `N`, we find the `foo->bar` rename in `N`, but since it
+is unrelated to the `foo` file in `M` (assuming the `foo` file created in `M`
+used a different salt), we will not perform any renames. The new `foo` file
+is then simply created in the rebased `M` just like it was before the rebase.
+
+```
```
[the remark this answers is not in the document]
martinvonz: Deleted

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,671 @@ results of rebasing across remote copies.
...
+|
+| L rename foo->bar
+|/
+K add foo="hello"
+```
+
+When rebasing `M` onto `L`, we apply the `foo->bar` rename to the trees in `M`
+and its parent.
+
+```
```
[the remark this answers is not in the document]
martinvonz: Done

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,671 @@ results of rebasing across remote copies.
...
+
+### Merging
+
+When merging trees, we start by rewriting each diff to match any different names
+in the destination tree. For example, if the tree conflict is `A+(B-C)+(D-E)`,
+then we will rewrite the `(B-C)` diff and the `(D-E)` diff to the paths in `A`.
+To translate the `(B-C)` diff, we calculate renames from `C` to `A` and then we
+apply those renames to both `C` and `B`. This may result in conflicts.
+
+If a file has a conflict in the copy ID, it will appear as if it doesn't exist
```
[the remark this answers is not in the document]
martinvonz: I think the paragraph makes sense here. I've rewritten this sections quite a bit. We can adjust further later.

docs/design/copy-tracking.md
martinvonz: > I don't think this is urgent, but I think we should eventually add more details here, explaining the conflict with copy ids that you get, and how we decide to keep it a conflict. For example, it is critical that we are thinking of conflicts as conflicts of trees for this to work, IIRC.

I agree. I added a bit about this directly under the "Merging" heading.

> There is already a "Conflicts" section below, but it feels a bit out of date and doesn't discuss this.

Yes, that was obsolete. I've removed a lot of obsolete contents now.

docs/design/copy-tracking.md
```diff
@@ -10,17 +10,686 @@ results of rebasing across remote copies.
...
+to another idea I've had in the past: we could replace the current tree-level
+conflict representation where we have full snapshots by having a single snapshot
+and then diffs represented by sparse trees with only relevant changes. However,
+specifically for copy/rename propagation, it can be useful to have the whole
+old tree in order to detect if a whole directory was renamed.
+
+I don't see a solution that makes the "rebase -> rebase-back" cycle a no-op,
+unless we decide to not propagate modifications to copies in the first place.
+
+#### Example: Multiple copies
```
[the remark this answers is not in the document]
martinvonz: > In the following graph, what happens if we rebase P to N?

We would first find that `foo` and `bar` were modified in P, so we need to consider copies related to them. We then look them up in N and find that both files are related to `bar` there. Since `foo` was copied from `bar`, we ask the user if they want to propagate the change to `foo` into `bar`.

> What if we rebase Q to N?

This is the same case, except that if the user says they want to propagate the changes, the resulting contents is a conflict.

docs/design/copy-tracking.md
martinvonz: Done

spundun:
@joyously Hey I haven't thought about this deeply to the level of writing a design doc, but I do think VCS should track copies and moves.

Here's my 2 minute case for it: Often copies and moves are meant for uncomplicating things, reducing complexity, reducing the burden on the reader of the code in trying to understand it. Features like blame and commit log are crucial in being able to understand why the code is the way it is, in what context was a line of code written, etc. If renaming a file(moving) breaks the commit log,(also breaking blame as a result), then it's a huge setback in providing the context behind the state of the code to the reader. This is counter productive to the very thing that the coder was trying to do by renaming the file, simplify and organize things into being more comprehensible.

Same thing applies to copy mutation(sometimes even more acutely). One common case of refactoring is when a file of code keeps blowing up and the file is just too big, now I want to split it into multiple smaller files, again with the purpose of simplifying things and making it easier for the reader to understand the code. But it is absolutely crucial that the code doesn't lose it's history of when, how and why it was checked in, in the first place. The best way I've found to preserve this information, is to track the file across copies. With hg, I'll often duplicate a file and remove a portion of the code from it, and from the other copy, I'll remove all the code except the portion that I removed from the other file. I've effectively split the file into two without adding any lines of code(except for some boilerplate to preserve the scope of the code). Because hg tracks the copies, I can still see the original commits when each line of code was checked in, in BOTH splits of the file...

I hope I've been able to illustrate the use cases and provide what might be a first case for copy-tracing in a VCS to you.

Cheers.
PS: Sorry if I misunderstood this conversation that I just jumped in the middle of.

Edit: I for one, have been desperately waiting for this feature to be included in jj

Another Edit: Ultimately what matters to me is that when I copy or rename files, if I try to see the commit log of that file or see the blame annotations, then all the commits show up appropriately in both those views, even the commits from before I copied or renamed or moved the file.

joyously:
@spundun Thanks for the use case for tracking a copy. (I already agree that renames should be tracked.) I don't think your use case is a good use of copy; it seems that tracking a split would be better. The discussion above was about the VCS showing conflicts in copies, and making the same changes in copies as in the original, when a rebase happens. This is the part where I think the user can't possibly know if it should as it can get really complicated having to know all the history of the code.
If the tracking of copy was only for being able to show `blame` correctly, that seems okay, although it seems that `blame` could show the fact of the copy and that would be sufficient.

PR #4991, opened 2024-11-28
merged_tree: minor refactor of TreeDiffStreamImpl

yuja:
This is remainder of #4240.

# Checklist

If applicable:


lib/src/merged_tree.rs
```diff
@@ -952,37 +952,36 @@ impl<'matcher> TreeDiffStreamImpl<'matcher> {
     }
 
     /// Gets the given tree if `value` is a tree, otherwise an empty tree.
```
martinvonz, line 954: nit: Maybe "tree" -> "trees" now, and "value" should have been "values" from before

martinvonz (approved):

lib/src/merged_tree.rs
yuja: Inserted new commit to fix the doc comments. There are two instances.

PR #4997, opened 2024-11-29
docs: clarify dynamic completion features

senekor:
# Checklist

If applicable:

### DIFF
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -210,13 +210,17 @@ To set up command-line completion, source the output of
 `jj util completion bash/zsh/fish`. Exactly how to source it
 depends on your shell.
 
-Improved completions are currently in the works, these will complete things like
-bookmark names as well. You can activate them with the alternative "dynamic"
-instructions below. They should still complete everything the static completions
-did, so only activate one of them. Please let us know if you encounter any
-issues, so we can ensure a smooth transition once we default to these new
-completions. If you have ideas about specific completions that could be added,
-please share them [here](https://github.com/martinvonz/jj/issues/4763).
+Improved completions are also available. They will complete things like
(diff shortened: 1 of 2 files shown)


docs/install-and-setup.md
```diff
@@ -210,13 +210,17 @@ To set up command-line completion, source the output of
 `jj util completion bash/zsh/fish`. Exactly how to source it
 depends on your shell.
 
-Improved completions are currently in the works, these will complete things like
-bookmark names as well. You can activate them with the alternative "dynamic"
+Improved completions are also available, they will complete things like
```
martinvonz, line 213: Nit: "they will ... " is a full sentence, so use full stop and capital letter

docs/install-and-setup.md
```diff
@@ -210,13 +210,17 @@ To set up command-line completion, source the output of
...
+bookmarks, aliases, revisions, operations and files as well. They can be
```
martinvonz, line 214: It took me a few seconds to figure out what "as well" refers to. I assume it's "as well as what the static completions do". I'm not sure we need to say that. What do you think about just deleting "as well"? Oh, we already say further down that they also support the static completions.

martinvonz (approved):

senekor:
Btw. I hard wrapped these lines because that seems to be the convention. I personally prefer [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line).

martinvonz:
I had not heard of that convention until very recently. It was mentioned in a recent PR for the tutorial (I'll let you search for it). As I said there, I would be fine with changing to that convention. Whichever convention we choose, we should usually find a formatter that can automate it for us.

PR #5000, opened 2024-11-29
settings: own StackedConfig by UserSettings, migrate tests to use config layer

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_commit_builder.rs
+++ b/lib/tests/test_commit_builder.rs
@@ -13,11 +13,15 @@
 // limitations under the License.
 
 use futures::StreamExt as _;
+use indoc::indoc;
 use itertools::Itertools;
 use jj_lib::backend::ChangeId;
 use jj_lib::backend::MillisSinceEpoch;
 use jj_lib::backend::Signature;
 use jj_lib::backend::Timestamp;
+use jj_lib::config::ConfigLayer;
+use jj_lib::config::ConfigSource;
+use jj_lib::config::StackedConfig;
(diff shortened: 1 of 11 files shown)


lib/tests/test_commit_builder.rs
```diff
@@ -257,11 +268,15 @@ fn test_rewrite_resets_author_timestamp(backend: TestRepoBackend) {
...
-        .unwrap();
+    let mut config = testutils::base_user_config();
+    config.add_layer(ConfigLayer::with_data(
+        ConfigSource::User,
+        config::Config::builder()
+            .set_override("debug.commit-timestamp", initial_timestamp)
+            .unwrap()
+            .build()
+            .unwrap(),
+    ));
```
martinvonz: > Not all tests are ported to ConfigLayer::parse() because it seemed a bit odd
to format!() a TOML document and parse it to build a table of configuration
variables.

Maybe we'll want a `ConfigLayer::set_value()` so this can be something like this:

```
let mut layer = ConfigLayer::empty(ConfigSource::User);
layer.set_value("debug.commit-timestamp", initial_timestamp);
config.add_layer(layer);
```

Just an idea. Not asking for any changes in this PR.

martinvonz (approved):

lib/tests/test_commit_builder.rs
yuja: Yes, that's one option. Once migrated to `toml_edit`, we can initialize the variables in a similar way to `HashMap`. Maybe there will be some helper function to set a value with dotted name path.

Issue #5002, opened 2024-11-29
FR: Optionally emit more details when refusing to push private commits


shanesveller:
**Is your feature request related to a problem? Please describe.**

When I have commits that match my `private` settings, and I attempt to bulk push using commands like `jj git push [--all | --tracked]` or similar, the default error sort of immediately forces me to `jj show` a regular commit ID in order to recall what the change was, or what its description was. Or I could use `jj log` to also see it in the context of its lineage.

Sample from the test suite code:
```
Error: Won't push commit aa3058ff8663 since it is private
```

Given how `jj` has adjusted my workflows, I'm less likely than ever to recognize a given raw commit SHA at a glance, and if my push syntax didn't reference a specific bookmark or change I don't have that to contextualize it with either.

**Describe the solution you'd like**
An early improvement would be to emit the description and/or change ID as part of the existing command output:

```
# same sample as above
Error: Won't push commit aa3058ff8663 since it is private
commit aa3058ff8663, description: 'private 1'
```

This could be opt-in as a flag if preserving the terse output as default is desired.

I'd love even richer output that collated multiple errors if I had a bulk push attempt that included more than one private commit as I believe it will currently fail-fast.

**Describe alternatives you've considered**

A suffix output with something spiritually like:
```
Reminder: jj log -r 'private() & {the-original-push-revset}' for context
```
Would help steer me as well, although `private()` is not currently a revset/alias or defined directly as one, it's a separate `git.private-commits` settings so on 0.23.0 I'd actually have to write:
```bash
jj log -r "$(jj config get git.private-commits) ..."
```

**Additional context**
I'm open to any solutions that improve the UX here, but I'll also open a PR shortly with an attempt at the "include the description via opt-in flag" tactic above, in case it passes muster.

yuja:
I think this is implemented by https://github.com/martinvonz/jj/commit/5553baf540e203aed072cbbfaec8815938283f3d, closing.

shanesveller:
Fine with me! The only passage that I considered unfulfilled was:

> I'd love even richer output that collated multiple errors if I had a bulk push attempt that included more than one [private or otherwise failed] commit as I believe it will currently fail-fast.

PR #5003, opened 2024-11-29
cli: git push: optionally describe commits that would fail push

shanesveller:
First time-contributor, so feedback at any level of pedantry is fine while I adjust to the project norms!

Related to #5002.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -402,15 +402,30 @@ fn validate_commits_ready_to_push(
         if commit.has_conflict()? {
             reasons.push("it has conflicts");
         }
-        if !args.allow_private && is_private(commit.id())? {
+        let is_private = is_private(commit.id())?;
+        if !args.allow_private && is_private {
             reasons.push("it is private");
         }
         if !reasons.is_empty() {
-            return Err(user_error(format!(
+            let mut error = user_error(format!(
                 "Won't push commit {} since {}",
                 short_commit_hash(commit.id()),
                 reasons.join(" and ")
-            )));
+            ));
+            error.add_formatted_hint_with(|formatter| {
+                write!(formatter, "Rejected commit: ")?;
+                workspace_helper.write_commit_summary(formatter, &commit)?;
+                Ok(())
+            });
+            if !args.allow_private && is_private {
+                error.add_hint(format!(
+                    "Configured git.private-commits: '{}'",
+                    settings
+                        .get_string("git.private-commits")
+                        .expect("should have private-commits setting")
+                ));
+            }
+            return Err(error);
         }
     }
     Ok(())
--- a/cli/tests/test_git_private_commits.rs
+++ b/cli/tests/test_git_private_commits.rs
@@ -88,9 +88,11 @@ fn test_git_private_commits_block_pushing() {
     // Will not push when a pushed commit is contained in git.private-commits
     test_env.add_config(r#"git.private-commits = "description(glob:'private*')""#);
     let stderr = test_env.jj_cmd_failure(&workspace_root, &["git", "push", "--all"]);
-    insta::assert_snapshot!(stderr, @r###"
+    insta::assert_snapshot!(stderr, @r"
     Error: Won't push commit aa3058ff8663 since it is private
-    "###);
+    Hint: Rejected commit: yqosqzyt aa3058ff main* | (empty) private 1
+    Hint: Configured git.private-commits: 'description(glob:'private*')'
+    ");
 
     // May push when the commit is removed from git.private-commits
     test_env.add_config(r#"git.private-commits = "none()""#);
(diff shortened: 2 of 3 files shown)


cli/src/commands/git/push.rs
```diff
@@ -140,6 +140,9 @@ pub struct GitPushArgs {
     /// Only display what will change on the remote
     #[arg(long)]
     dry_run: bool,
+    /// Display per-commit reasons for failed pushes
+    #[arg(long)]
+    explain_denied: bool,
```
shanesveller, line 145: Feedback about the user-facing phrasing is especially welcome, I don't have any real attachment to this one. Same with all naming, really.

cli/src/commands/git/push.rs
```diff
@@ -402,15 +405,33 @@ fn validate_commits_ready_to_push(
...
                 "Won't push commit {} since {}",
                 short_commit_hash(commit.id()),
                 reasons.join(" and ")
-            )));
+            );
+            if args.explain_denied {
+                message.push('\n');
+                message.push_str(&exposition.join("\n"));
+            }
+            return Err(user_error(message));
```
yuja, line 434: I personally don't think we need a flag to enable detailed error message.

For formatting, we can use commit_summary_template and `error.added_formatted_hint_with()`. See `check_rewritable()` in cmd_util.rs for example.

shanesveller: Much obliged, will revise tomorrow or the next day!

cli/tests/test_git_private_commits.rs
```diff
@@ -88,9 +88,11 @@ fn test_git_private_commits_block_pushing() {
     // Will not push when a pushed commit is contained in git.private-commits
     test_env.add_config(r#"git.private-commits = "description(glob:'private*')""#);
     let stderr = test_env.jj_cmd_failure(&workspace_root, &["git", "push", "--all"]);
-    insta::assert_snapshot!(stderr, @r###"
```
shanesveller, line 91: ℹ️  The triple octothorpes `###` were removed automatically during my run of `cargo insta test --review --workspace --test-runner nextest`. I'm using `cargo-insta 1.41.1` and `cargo-nextest 0.9.82`. Please let me know if I should restore that part of the diff, or adjust my command invocation in some way.

cli/src/commands/git/push.rs
shanesveller: I've adopted this feedback during my last push, thank you for teaching me about those features of the codebase!

cli/src/commands/git/push.rs
```diff
@@ -402,15 +402,35 @@ fn validate_commits_ready_to_push(
...
-            return Err(user_error(format!(
+            let short = short_commit_hash(commit.id());
+            let mut error = user_error(format!(
                 "Won't push commit {} since {}",
-                short_commit_hash(commit.id()),
+                short,
                 reasons.join(" and ")
-            )));
+            ));
+            if !args.allow_private && is_private {
```
yuja, line 416: nit: isn't it better to show the commit detail for whatever reasons? I think it's useful to know which conflicted commit prevented pushing for example.

cli/tests/test_git_private_commits.rs
yuja: It's new `cargo insta` behavior. No need to preserve the old style.

cli/src/commands/git/push.rs
shanesveller: That seems right to me, and wasn't a particularly noisy diff. I've included this suggestion in my last push as 608bb6ea5e8e2a1b7c4c3dc4379f54fba617c4d0 .

shanesveller: The main motive for the static prefix on my side was in case someone has reconfigured their `templates.commit_summary` in such a way that doesn't include enough of the default details, especially the change/commit IDs. For now I've left it alone, fixed the `short` variable and squashed the commit.

Repeating "Won't push commit" would also be fine with me and I'm not at all perturbed if the phrasing gets touched up in follow-up work. I'll think on it for a while in case I can come up something that feels better.

martinvonz: Another option is "Rejected commit: " or "Problematic commit: " or something like that. I don't feel strongly either, but I do think "Commit summary: " sounds too generic.

shanesveller: I do like "rejected" better here, so I've switched to that. I am mildly hesitant about wording that might give the impression that the *remote* side had a problem with it instead, i.e. tripping GitHub branch protection rules.

cli/src/commands/git/push.rs
```diff
@@ -402,15 +402,34 @@ fn validate_commits_ready_to_push(
...
+                error.add_formatted_hint_with(|formatter| {
+                    write!(
+                        formatter,
+                        "Configured git.private-commits: '{}'",
+                        settings
+                            .get_string("git.private-commits")
+                            .expect("should have private-commits setting")
+                    )?;
+                    Ok(())
```
yuja, line 429: nit: the latter hint can be `error.add_hint(format!(..))`

yuja (approved):

cli/src/commands/git/push.rs
shanesveller: Fixed, thank you!

martinvonz:
Looks like all comments have been addressed , so feel free to mark them resolved and then merge this PR

PR #5007, opened 2024-12-01
config: add StackedConfig method that looks up value from all layers, migrate settings API

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/merge_tools/mod.rs
+++ b/cli/src/merge_tools/mod.rs
@@ -159,7 +159,7 @@ pub fn get_external_tool_config(
     name: &str,
 ) -> Result<Option<ExternalMergeTool>, ConfigError> {
     const TABLE_KEY: &str = "merge-tools";
-    let tools_table = settings.raw_config().get_table(TABLE_KEY)?;
+    let tools_table = settings.get_table(TABLE_KEY)?;
     if let Some(v) = tools_table.get(name) {
         let mut result: ExternalMergeTool = v
             .clone()
--- a/cli/src/merge_tools/mod.rs
+++ b/cli/src/merge_tools/mod.rs
@@ -125,7 +125,7 @@ impl MergeTool {
 fn editor_args_from_settings(
     ui: &Ui,
     settings: &UserSettings,
-    key: &str,
+    key: &'static str,
 ) -> Result<CommandNameAndArgs, ConfigError> {
     // TODO: Make this configuration have a table of possible editors and detect the
     // best one here.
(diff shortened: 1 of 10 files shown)


lib/src/config.rs
```diff
@@ -381,10 +407,108 @@ impl StackedConfig {
...
+        }
+    }
+
+    // Simply merge tables from the bottom layer. Upper items should override
+    // the lower items (including their children) no matter if the upper items
+    // are shadowed by the other upper items.
+    let (mut merged, mut top_index) = match to_merge.pop() {
+        Some((item, index)) => (item.clone(), index),
+        None => return None,
+    };
```
martinvonz, line 478: optional: I suppose this could be written using `?` for `Option`:

```rust
    let (item, mut top_index) = to_merge.pop()?;
    let mut merged = item.clone();
```

cli/src/merge_tools/mod.rs
```diff
@@ -158,22 +159,19 @@ pub fn get_external_tool_config(
...
-        if result.program.is_empty() {
-            result.program.clone_from(&name.to_string());
-        };
-        Ok(Some(result))
-    } else {
-        Ok(None)
-    }
+    let full_name = ConfigNamePathBuf::from_iter(["merge-tools", name]);
+    let Some(mut tool) = settings
+        .get::<ExternalMergeTool>(&full_name)
```
martinvonz, line 164: From commit message:
> cli: leverage new settings.get(dotted_name) in merge-tools lookup

We don't seem to use the dotted_name form here. Is the commit message stale?

martinvonz (approved):

lib/src/config.rs
```diff
@@ -289,6 +290,31 @@ impl ConfigLayer {
             .map(|path| Self::load_from_file(source, path))
             .try_collect()
     }
+
+    /// Looks up item by the `name` path. Returns `Some(item)` if an item
+    /// found at the path. Returns `Err(item)` if middle node wasn't a table.
+    fn lookup_item(&self, name: &ConfigNamePathBuf) -> Result<Option<&ConfigValue>, &ConfigValue> {
```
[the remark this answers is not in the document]
yuja: Renamed.

lib/src/config.rs
yuja: Good catch. Maybe it was implemented that way because the return type was originally `Result<_>`.

cli/src/merge_tools/mod.rs
yuja: Fixed as s/dotted_name/path/. Maybe I just called the name path type as dotted_name.

PR #5028, opened 2024-12-04
release: release version 0.24.0

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+### Deprecations
+
+### New features
+
+### Fixed bugs
+
+## [0.24.0] - 2024-12-04
+
+### Breaking changes
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+### Deprecations
+
+### New features
+
+### Fixed bugs
+
+## [0.24.0] - 2024-12-04
```
thoughtpolice, line 19: I wonder if we should put the new dynamic completions support in a "Highlights" section or something, since it's been a huge ask for so long. Either way, LGTM.

thoughtpolice (approved):

CHANGELOG.md
martinvonz: Makes sense, but I think I remember reading that dynamic completions only work well with Fish so far, so we'll probably want to fix the bugs (in jj, clap, or elsewhere) and test it a bit more first.

PR #5030, opened 2024-12-05
local_working_copy: spawn snapshot job per directory,  filter deleted files per job

yuja:
# Checklist

If applicable:

---

Benchmark:
1. original
2. per-directory spawn
3. per-directory deleted files
4. shorter path comparison

gecko-dev (~357k files, ~25k dirs)
```
% JJ_CONFIG=/dev/null hyperfine --sort command --warmup 3 --runs 30 ..
Benchmark 1: target/release-with-debug/jj-1 -R ~/mirrors/gecko-dev debug snapshot
  Time (mean ± σ):     764.9 ms ±  16.7 ms    [User: 3274.7 ms, System: 2183.3 ms]
  Range (min … max):   731.9 ms … 814.2 ms    30 runs

Benchmark 2: target/release-with-debug/jj-2 -R ~/mirrors/gecko-dev debug snapshot
  Time (mean ± σ):     710.7 ms ±   9.1 ms    [User: 3070.7 ms, System: 2142.6 ms]
  Range (min … max):   695.9 ms … 740.1 ms    30 runs

Benchmark 3: target/release-with-debug/jj-3 -R ~/mirrors/gecko-dev debug snapshot
  Time (mean ± σ):     480.1 ms ±   8.8 ms    [User: 3190.5 ms, System: 2127.2 ms]
  Range (min … max):   471.2 ms … 509.8 ms    30 runs

Benchmark 4: target/release-with-debug/jj-4 -R ~/mirrors/gecko-dev debug snapshot
  Time (mean ± σ):     404.0 ms ±   4.4 ms    [User: 1933.4 ms, System: 2148.8 ms]
  Range (min … max):   396.4 ms … 416.9 ms    30 runs

Relative speed comparison
        1.89 ±  0.05  target/release-with-debug/jj-1 -R ~/mirrors/gecko-dev debug snapshot
        1.76 ±  0.03  target/release-with-debug/jj-2 -R ~/mirrors/gecko-dev debug snapshot
        1.19 ±  0.03  target/release-with-debug/jj-3 -R ~/mirrors/gecko-dev debug snapshot
        1.00          target/release-with-debug/jj-4 -R ~/mirrors/gecko-dev debug snapshot
```

linux (~87k files, ~6k dirs)
```
% JJ_CONFIG=/dev/null hyperfine --sort command --warmup 3 --runs 30 ..
Benchmark 1: target/release-with-debug/jj-1 -R ~/mirrors/linux debug snapshot
  Time (mean ± σ):     268.2 ms ±  11.3 ms    [User: 636.6 ms, System: 518.5 ms]
  Range (min … max):   247.5 ms … 295.2 ms    30 runs

Benchmark 2: target/release-with-debug/jj-2 -R ~/mirrors/linux debug snapshot
  Time (mean ± σ):     242.3 ms ±   3.3 ms    [User: 656.8 ms, System: 538.0 ms]
  Range (min … max):   236.9 ms … 252.3 ms    30 runs

Benchmark 3: target/release-with-debug/jj-3 -R ~/mirrors/linux debug snapshot
  Time (mean ± σ):     204.2 ms ±   3.0 ms    [User: 667.3 ms, System: 545.6 ms]
  Range (min … max):   197.1 ms … 209.2 ms    30 runs

Benchmark 4: target/release-with-debug/jj-4 -R ~/mirrors/linux debug snapshot
  Time (mean ± σ):     191.3 ms ±   3.3 ms    [User: 467.4 ms, System: 542.2 ms]
  Range (min … max):   186.1 ms … 200.6 ms    30 runs

Relative speed comparison
        1.40 ±  0.06  target/release-with-debug/jj-1 -R ~/mirrors/linux debug snapshot
        1.27 ±  0.03  target/release-with-debug/jj-2 -R ~/mirrors/linux debug snapshot
        1.07 ±  0.02  target/release-with-debug/jj-3 -R ~/mirrors/linux debug snapshot
        1.00          target/release-with-debug/jj-4 -R ~/mirrors/linux debug snapshot
```

nixpkgs (~45k files, ~31k dirs)
```
% JJ_CONFIG=/dev/null hyperfine --sort command --warmup 3 --runs 30 ..
Benchmark 1: target/release-with-debug/jj-1 -R ~/mirrors/nixpkgs debug snapshot
  Time (mean ± σ):     201.0 ms ±   8.5 ms    [User: 929.3 ms, System: 917.6 ms]
  Range (min … max):   170.3 ms … 218.5 ms    30 runs

Benchmark 2: target/release-with-debug/jj-2 -R ~/mirrors/nixpkgs debug snapshot
  Time (mean ± σ):     190.7 ms ±   4.1 ms    [User: 859.3 ms, System: 881.1 ms]
  Range (min … max):   184.6 ms … 202.4 ms    30 runs

Benchmark 3: target/release-with-debug/jj-3 -R ~/mirrors/nixpkgs debug snapshot
  Time (mean ± σ):     173.3 ms ±   6.7 ms    [User: 899.4 ms, System: 889.0 ms]
  Range (min … max):   166.5 ms … 197.9 ms    30 runs

Benchmark 4: target/release-with-debug/jj-4 -R ~/mirrors/nixpkgs debug snapshot
  Time (mean ± σ):     161.7 ms ±   2.5 ms    [User: 739.1 ms, System: 881.7 ms]
  Range (min … max):   156.5 ms … 166.4 ms    30 runs

Relative speed comparison
        1.24 ±  0.06  target/release-with-debug/jj-1 -R ~/mirrors/nixpkgs debug snapshot
        1.18 ±  0.03  target/release-with-debug/jj-2 -R ~/mirrors/nixpkgs debug snapshot
        1.07 ±  0.04  target/release-with-debug/jj-3 -R ~/mirrors/nixpkgs debug snapshot
        1.00          target/release-with-debug/jj-4 -R ~/mirrors/nixpkgs debug snapshot
```

git (~4.5k files, 0.2k dirs)
```
% JJ_CONFIG=/dev/null hyperfine --sort command --warmup 30 --runs 50 ..
Benchmark 1: target/release-with-debug/jj-1 -R ~/mirrors/git debug snapshot
  Time (mean ± σ):      30.3 ms ±   1.1 ms    [User: 40.5 ms, System: 39.4 ms]
  Range (min … max):    28.3 ms …  35.7 ms    50 runs

Benchmark 2: target/release-with-debug/jj-2 -R ~/mirrors/git debug snapshot
  Time (mean ± σ):      30.6 ms ±   1.1 ms    [User: 33.8 ms, System: 39.0 ms]
  Range (min … max):    29.0 ms …  35.0 ms    50 runs

Benchmark 3: target/release-with-debug/jj-3 -R ~/mirrors/git debug snapshot
  Time (mean ± σ):      28.8 ms ±   1.0 ms    [User: 33.0 ms, System: 37.6 ms]
  Range (min … max):    26.8 ms …  31.3 ms    50 runs

Benchmark 4: target/release-with-debug/jj-4 -R ~/mirrors/git debug snapshot
  Time (mean ± σ):      28.8 ms ±   1.9 ms    [User: 30.3 ms, System: 36.5 ms]
  Range (min … max):    26.0 ms …  39.2 ms    50 runs

Relative speed comparison
        1.05 ±  0.05  target/release-with-debug/jj-1 -R ~/mirrors/git debug snapshot
        1.06 ±  0.05  target/release-with-debug/jj-2 -R ~/mirrors/git debug snapshot
        1.00          target/release-with-debug/jj-3 -R ~/mirrors/git debug snapshot
        1.00 ±  0.08  target/release-with-debug/jj-4 -R ~/mirrors/git debug snapshot
```

### DIFF
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -1059,15 +1059,9 @@ impl FileSnapshotter<'_> {
                 message: format!("Failed to read directory {}", disk_dir.display()),
                 err: err.into(),
             })?;
-        dir_entries.into_par_iter().try_for_each_with(
-            (
-                self.tree_entries_tx.clone(),
-                self.file_states_tx.clone(),
-                self.present_files_tx.clone(),
-            ),
-            |(tree_entries_tx, file_states_tx, present_files_tx),
-             entry|
-             -> Result<(), SnapshotError> {
+        dir_entries
+            .into_par_iter()
+            .try_for_each(|entry| -> Result<(), SnapshotError> {
                 let file_type = entry.file_type().unwrap();
                 let file_name = entry.file_name();
                 let name = file_name
@@ -1097,36 +1091,7 @@ impl FileSnapshotter<'_> {
 
                         // If the whole directory is ignored, visit only paths we're already
                         // tracking.
-                        for (tracked_path, current_file_state) in file_states {
-                            if !self.matcher.matches(tracked_path) {
-                                continue;
-                            }
-                            let disk_path =
-                                tracked_path.to_fs_path(&self.tree_state.working_copy_path)?;
-                            let metadata = match disk_path.symlink_metadata() {
-                                Ok(metadata) => metadata,
-                                Err(err) if err.kind() == io::ErrorKind::NotFound => {
-                                    continue;
-                                }
-                                Err(err) => {
-                                    return Err(SnapshotError::Other {
-                                        message: format!(
-                                            "Failed to stat file {}",
-                                            disk_path.display()
-                                        ),
-                                        err: err.into(),
-                                    });
-                                }
-                            };
-                            if let Some(new_file_state) = file_state(&metadata) {
-                                self.process_present_file(
-                                    tracked_path.to_owned(),
-                                    &disk_path,
-                                    Some(&current_file_state),
-                                    new_file_state,
-                                )?;
-                            }
-                        }
+                        self.visit_tracked_files(file_states)?;
                     } else {
                         let directory_to_visit = DirectoryToVisit {
                             dir: path,
@@ -1176,8 +1141,38 @@ impl FileSnapshotter<'_> {
                     }
                 }
                 Ok(())
-            },
-        )?;
+            })?;
+        Ok(())
+    }
+
+    /// Visits only paths we're already tracking.
+    fn visit_tracked_files(&self, file_states: FileStates<'_>) -> Result<(), SnapshotError> {
+        for (tracked_path, current_file_state) in file_states {
+            if !self.matcher.matches(tracked_path) {
+                continue;
+            }
+            let disk_path = tracked_path.to_fs_path(&self.tree_state.working_copy_path)?;
+            let metadata = match disk_path.symlink_metadata() {
+                Ok(metadata) => metadata,
+                Err(err) if err.kind() == io::ErrorKind::NotFound => {
+                    continue;
+                }
+                Err(err) => {
+                    return Err(SnapshotError::Other {
+                        message: format!("Failed to stat file {}", disk_path.display()),
+                        err: err.into(),
+                    });
+                }
+            };
+            if let Some(new_file_state) = file_state(&metadata) {
+                self.process_present_file(
+                    tracked_path.to_owned(),
+                    &disk_path,
+                    Some(&current_file_state),
+                    new_file_state,
+                )?;
+            }
+        }
         Ok(())
     }
 
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -37,6 +37,7 @@ use std::slice;
 use std::sync::mpsc::channel;
 use std::sync::mpsc::Sender;
 use std::sync::Arc;
+use std::sync::OnceLock;
 use std::time::UNIX_EPOCH;
 
 use futures::StreamExt;
@@ -46,6 +47,7 @@ use once_cell::unsync::OnceCell;
 use pollster::FutureExt;
 use prost::Message;
 use rayon::iter::IntoParallelIterator;
+use rayon::prelude::IndexedParallelIterator;
 use rayon::prelude::ParallelIterator;
 use tempfile::NamedTempFile;
 use thiserror::Error;
@@ -900,6 +902,7 @@ impl TreeState {
                 tree_entries_tx,
                 file_states_tx,
                 present_files_tx,
+                error: OnceLock::new(),
                 progress,
                 max_new_file_size,
                 conflict_marker_style,
@@ -910,7 +913,13 @@ impl TreeState {
                 git_ignore: base_ignores.clone(),
                 file_states: self.file_states.all(),
             };
-            snapshotter.visit_directory(directory_to_visit)
+            // Here we use scope as a queue of per-directory jobs.
+            rayon::scope(|scope| {
+                snapshotter.spawn_ok(scope, |scope| {
+                    snapshotter.visit_directory(directory_to_visit, scope)
+                });
+            });
+            snapshotter.into_result()
         })?;
 
         let mut tree_builder = MergedTreeBuilder::new(self.tree_id.clone());
@@ -1033,23 +1042,49 @@ struct FileSnapshotter<'a> {
     tree_entries_tx: Sender<(RepoPathBuf, MergedTreeValue)>,
     file_states_tx: Sender<(RepoPathBuf, FileState)>,
     present_files_tx: Sender<RepoPathBuf>,
+    error: OnceLock<SnapshotError>,
     progress: Option<&'a SnapshotProgress<'a>>,
     max_new_file_size: u64,
     conflict_marker_style: ConflictMarkerStyle,
 }
 
 impl FileSnapshotter<'_> {
-    fn visit_directory(&self, directory_to_visit: DirectoryToVisit) -> Result<(), SnapshotError> {
+    fn spawn_ok<'scope, F>(&'scope self, scope: &rayon::Scope<'scope>, body: F)
+    where
+        F: FnOnce(&rayon::Scope<'scope>) -> Result<(), SnapshotError> + Send + 'scope,
+    {
+        scope.spawn(|scope| {
+            if self.error.get().is_some() {
+                return;
+            }
+            match body(scope) {
+                Ok(()) => {}
+                Err(err) => self.error.set(err).unwrap_or(()),
+            };
+        });
+    }
+
+    /// Extracts the result of the snapshot.
+    fn into_result(self) -> Result<(), SnapshotError> {
+        match self.error.into_inner() {
+            Some(err) => Err(err),
+            None => Ok(()),
+        }
+    }
+
+    /// Visits the directory entries, spawns jobs to recurse into sub
+    /// directories.
+    fn visit_directory<'scope>(
+        &'scope self,
+        directory_to_visit: DirectoryToVisit<'scope>,
+        scope: &rayon::Scope<'scope>,
+    ) -> Result<(), SnapshotError> {
         let DirectoryToVisit {
             dir,
             disk_dir,
             git_ignore,
             file_states,
         } = directory_to_visit;
-
-        if self.matcher.visit(&dir).is_nothing() {
-            return Ok(());
-        }
 
         let git_ignore = git_ignore
             .chain_with_file(&dir.to_internal_dir_string(), disk_dir.join(".gitignore"))?;
@@ -1062,16 +1097,22 @@ impl FileSnapshotter<'_> {
             })?;
         dir_entries
             .into_par_iter()
-            .try_for_each(|entry| self.process_dir_entry(&dir, &git_ignore, file_states, &entry))?;
+            // Don't split into too many small jobs. For a small directory,
+            // sequential scan should be fast enough.
+            .with_min_len(100)
+            .try_for_each(|entry| {
+                self.process_dir_entry(&dir, &git_ignore, file_states, &entry, scope)
+            })?;
         Ok(())
     }
 
-    fn process_dir_entry(
-        &self,
+    fn process_dir_entry<'scope>(
+        &'scope self,
         dir: &RepoPath,
         git_ignore: &Arc<GitIgnoreFile>,
-        file_states: FileStates<'_>,
+        file_states: FileStates<'scope>,
         entry: &DirEntry,
+        scope: &rayon::Scope<'scope>,
     ) -> Result<(), SnapshotError> {
         let file_type = entry.file_type().unwrap();
         let file_name = entry.file_name();
@@ -1102,15 +1143,17 @@ impl FileSnapshotter<'_> {
 
                 // If the whole directory is ignored, visit only paths we're already
                 // tracking.
-                self.visit_tracked_files(file_states)?;
-            } else {
+                self.spawn_ok(scope, move |_| self.visit_tracked_files(file_states));
+            } else if !self.matcher.visit(&path).is_nothing() {
                 let directory_to_visit = DirectoryToVisit {
                     dir: path,
                     disk_dir: entry.path(),
                     git_ignore: git_ignore.clone(),
                     file_states,
                 };
-                self.visit_directory(directory_to_visit)?;
+                self.spawn_ok(scope, |scope| {
+                    self.visit_directory(directory_to_visit, scope)
+                });
             }
         } else if self.matcher.matches(&path) {
             if let Some(progress) = self.progress {
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -19,6 +19,7 @@ use std::any::Any;
 use std::collections::HashSet;
 use std::error::Error;
 use std::fs;
+use std::fs::DirEntry;
 use std::fs::File;
 use std::fs::Metadata;
 use std::fs::OpenOptions;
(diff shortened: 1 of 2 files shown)


lib/src/local_working_copy.rs
```diff
@@ -1030,24 +1033,47 @@ struct FileSnapshotter<'a> {
...
+            match body(scope) {
+                Ok(()) => {}
+                Err(err) => self.error.set(err).unwrap_or(()),
+            };
+        });
+    }
+
+    /// Extracts the result of the snapshot.
+    fn into_result(self) -> Result<(), SnapshotError> {
+        self.error.into_inner().map_or(Ok(()), Err)
```
arxanas, line 1050: nit: It's hard for me to read the `map_or` here, can you write it as a `match`?

lib/src/local_working_copy.rs
```diff
@@ -1085,15 +1111,17 @@ impl FileSnapshotter<'_> {
 
                     // If the whole directory is ignored, visit only paths we're already
                     // tracking.
-                    self.visit_tracked_files(file_states)?;
-                } else {
+                    self.spawn_ok(scope, move |_| self.visit_tracked_files(file_states));
```
arxanas, line 1107: comment (non-blocking): Nice, I completely neglected/didn't consider that we're blocking on the sub-directory job here.

lib/src/local_working_copy.rs
```diff
@@ -1218,6 +1212,44 @@ impl FileSnapshotter<'_> {
...
+        enum Kind {
+            Dir,
+            File,
+        }
+
+        let file_state_chunks = file_states.iter().chunk_by(|(path, _state)| {
+            // Extract <name> from <dir>, <dir>/<name>, or <dir>/<name>/**.
+            // (file_states may contain <dir> file on file->dir transition.)
+            debug_assert!(path.starts_with(dir));
+            let len = dir.as_internal_file_string().len() + (!dir.is_root() as usize);
```
arxanas, line 1232: question: Regarding `(!dir.is_root() as usize)`, is this safe to do on Windows? In those cases, could the root be e.g. `C:\` and have length >1?

lib/src/local_working_copy.rs
```diff
@@ -1152,6 +1145,7 @@ impl FileSnapshotter<'_> {
                         });
                     }
                     if let Some(new_file_state) = file_state(&metadata) {
+                        present_entries.files.insert(name);
```
arxanas, line 1148: question: Shouldn't this be inside `process_present_file`? If not, can you add a comment explaining why?

lib/src/local_working_copy.rs
```diff
@@ -1162,6 +1156,7 @@ impl FileSnapshotter<'_> {
                 }
             }
         }
+        self.emit_deleted_files(&dir, file_states, &present_entries);
```
arxanas, line 1159: comment (non-blocking): It's a little unfortunate how this is all structured:

- It feels like you have to very carefully sprinkle additions to `present_entries` around the functions here and think very hard about which code paths need it.
- You have to worry about the correctness of early exiting from this function without calling `emit_deleted_files`. I think it's correct for now because we only exit in the case of error, which should terminate the snapshotting altogether?
  - I guess it would be ideal if this function returned `PresentDirEntries` (and marked it as `#[must_use]`) so that it could be enforced that every code path either returned with a value that must be consumed or an error, but maybe that's a little too much ceremony in the name of type-safety.
- For example: If we start emitting warnings for too-large files rather than returning fatal errors, then we might need to accumulate the warnings but also be careful to mark them as deleted for the snapshot.

I don't have concrete ideas on how to improve it, though.

lib/src/local_working_copy.rs
```diff
@@ -1218,6 +1212,44 @@ impl FileSnapshotter<'_> {
...
+            Dir,
+            File,
+        }
+
+        let file_state_chunks = file_states.iter().chunk_by(|(path, _state)| {
+            // Extract <name> from <dir>, <dir>/<name>, or <dir>/<name>/**.
+            // (file_states may contain <dir> file on file->dir transition.)
+            debug_assert!(path.starts_with(dir));
+            let len = dir.as_internal_file_string().len() + (!dir.is_root() as usize);
+            let tail = path.as_internal_file_string().get(len..).unwrap_or("");
```
arxanas, line 1233: suggestion: Maybe a variable name like `suffix` or `relative_path` is more helpful than `tail` (if that's correct)?

lib/src/local_working_copy.rs
```diff
@@ -1218,6 +1212,44 @@ impl FileSnapshotter<'_> {
...
+            match tail.split_once('/') {
+                Some((name, _)) => (Kind::Dir, name),
+                None => (Kind::File, tail),
+            }
```
arxanas, line 1237: question: Is there a specific invariant here like that the file path has at most one slash in it at this point?

lib/src/local_working_copy.rs
```diff
@@ -1218,6 +1212,44 @@ impl FileSnapshotter<'_> {
         Ok(())
     }
 
+    /// Emits file paths that don't exist in the `present_entries`.
+    fn emit_deleted_files(
+        &self,
+        dir: &RepoPath,
+        file_states: FileStates<'_>,
+        present_entries: &PresentDirEntries,
+    ) {
```
arxanas, line 1221: issue: Does this work when using Watchman? In particular: I thought `file_states` was strictly an optimization, and we previously didn't rely on the presence or absence of a file as a canonical data source, but now it seems that we're checking `file_states` to determine if a file was deleted. But under Watchman, we don't populate `file_states`, so I'm worried.

Alternatives to sending all the files over a channel (I've tried these out at a previous workplace and they both resulted in speedups):
- Store all files in thread-local storage, then collect them after the full traversal is finished.
  - I think we already block on the full traversal before we can start processing the present files?
- Introduce a `BufferedSender`/`BufferedReceiver` wrapper around channels.
  - I suspect there's unnecessary contention when we try to send to many individual items over a channel, and we don't really need to optimize for the latency of individual items.

arxanas (approved):
Main issue I'm concerned about is that I don't think we can rely on `file_states` for correctness; see my comment on the last commit. I'm approving to unblock, you can re-request review later if you want.

comment (non-blocking): There's some very interesting implementation comments in the walk implementation for `ignore`. At my last workplace, `ignore` was faster than both my custom Rust walk implementation and our previous C++ walk implementation at traversing the filesystem. You might be interested to read through it sometime.

For example: There's this comment about DFS vs BFS on wide directory trees: https://github.com/BurntSushi/ripgrep/blob/79cbe89deb1151e703f4d91b19af9cdcc128b765/crates/ignore/src/walk.rs#L1405-L1408 (relevant if you need to accumulate lots of `.gitignore`s).

For example: IIRC it uses a static work-stealing approach with a fixed buffer size (32 items?), unlike `rayon`, which dynamically sizes the amount of work to steal (aiming for half the remaining work in the queue). (I don't know if it makes a significant performance difference.)

suggestion: For the future, I recommend you use 30 runs as a default count for benchmarking rather than 10. At my last workplace, I noticed that times didn't tend to converge unless I provided a higher amount. It aligns with the "conventional wisdom" (the stuff I vaguely remember from statistics class) where they suggested you needed 15–30 samples for the kinds of tests they covered.

question (non-blocking): In the commit message for https://github.com/martinvonz/jj/pull/5030/commits/ef6d9ef43115937aa6bb71790b0ad2a0466a53b9, is it also supposed to show "Benchmark 1"? Does that disappear because you provided `--sort command`? Some of the other commit message benchmark results are missing `jj-0` or `jj-2`, so I had some trouble following the exact setup. (If it was intentionally or unintentionally removed, I don't think there's a need for you to regenerate those benchmarks.)

thoughtpolice:
Philip brought this up on Discord and was wondering what the impact on Nixpkgs would be; results are as follows:

```
austin@GANON:~/src/nixpkgs$ JJ_CONFIG=/dev/null hyperfine --sort command --warmup 3 --runs 10 -L bin jj,/home/austin/src/jj/target/release/jj '{bin} -R /home/austin/src/nixpkgs debug snapshot'
Benchmark 1: jj -R /home/austin/src/nixpkgs debug snapshot
  Time (mean ± σ):     823.7 ms ±  14.2 ms    [User: 813.8 ms, System: 96.9 ms]
  Range (min … max):   810.7 ms … 853.1 ms    10 runs

Benchmark 2: /home/austin/src/jj/target/release/jj -R /home/austin/src/nixpkgs debug snapshot
  Time (mean ± σ):     863.2 ms ±  26.4 ms    [User: 825.6 ms, System: 132.2 ms]
  Range (min … max):   821.0 ms … 897.2 ms    10 runs

Relative speed comparison
        1.00          jj -R /home/austin/src/nixpkgs debug snapshot
        1.05 ±  0.04  /home/austin/src/jj/target/release/jj -R /home/austin/src/nixpkgs debug snapshot
```

This seems reasonably consistent after a few runs, so it's 5% slower or somesuch. It's likely the depth and distribution of directories to files aren't quite as "fat" as some other repos; many directories in nixpkgs just have 1 file, even if they are deeply nested.

Note that I use watchman for this repo anyway however, so this change would not meaningfully impact my usage of Nixpkgs.

lib/src/local_working_copy.rs
yuja: It shouldn't because `visit_tracked_files()` doesn't need a set of present files to detect deletion.

lib/src/local_working_copy.rs
yuja: Yeah, that's the main reason I tried to make `visit_directory()` smaller. `present_files` could be returned to caller (with `#[must_use]`), but I don't think it would make huge difference.

lib/src/local_working_copy.rs
yuja: It relied on `file_states.all()` before, so there should be no change conceptually.

BTW, I noticed `deleted: HashSet<_> = file_states.all()...` (in the original impl) has measurable cost in gecko-dev benchmark. I think it's better to move filtering to per-directory jobs, and reduce string length to be hashed (full path -> file name).

lib/src/local_working_copy.rs
yuja: It's a (normalized) `RepoPath`, so there's no platform difference.

lib/src/local_working_copy.rs
yuja: I avoided `suffix` because it sounded like a file suffix. It could be called a `relative_path`, but it's a bit odd that `path` is `RepoPath`, but `relative_path` is `str`.

lib/src/local_working_copy.rs
yuja: The path is normalized `RepoPath`, so it never starts/ends with `/`.

yuja (commented):
> question (non-blocking): In the commit message for [ef6d9ef](https://github.com/martinvonz/jj/commit/ef6d9ef43115937aa6bb71790b0ad2a0466a53b9), is it also supposed to show "Benchmark 1"? Does that disappear because you provided `--sort command`?

I ran 3 benchmarks and split into two commit descriptions. That's why.

yuja:
> ```
> austin@GANON:~/src/nixpkgs$ JJ_CONFIG=/dev/null hyperfine --sort command --warmup 3 --runs 10 -L bin jj,/home/austin/src/jj/target/release/jj '{bin} -R /home/austin/src/nixpkgs debug snapshot'
> Benchmark 1: jj -R /home/austin/src/nixpkgs debug snapshot
>   Time (mean ± σ):     823.7 ms ±  14.2 ms    [User: 813.8 ms, System: 96.9 ms]
>   Range (min … max):   810.7 ms … 853.1 ms    10 runs
> 
> Benchmark 2: /home/austin/src/jj/target/release/jj -R /home/austin/src/nixpkgs debug snapshot
>   Time (mean ± σ):     863.2 ms ±  26.4 ms    [User: 825.6 ms, System: 132.2 ms]
>   Range (min … max):   821.0 ms … 897.2 ms    10 runs
> 
> Relative speed comparison
>         1.00          jj -R /home/austin/src/nixpkgs debug snapshot
>         1.05 ±  0.04  /home/austin/src/jj/target/release/jj -R /home/austin/src/nixpkgs debug snapshot
> ```

Yeah, the result in nixpkgs is bad. Appears that we need both per-directory `spawn()` (to not block the current directory walk) and `entries.par_iter()` with some threshold like `.with_min_len(100)` (to distribute large directory walk.)

thoughtpolice:
> Yeah, the result in nixpkgs is bad. Appears that we need both per-directory `spawn()` (to not block the current directory walk) and `entries.par_iter()` with some threshold like `.with_min_len(100)` (to distribute large directory walk.)

Sounds good; a simple cutoff for file count may be enough. Note that like everything this is probably filesystem sensitive; I'll get around to testing on XFS sometime this weekend. The above results are ext4.

yuja:
> we need both per-directory `spawn()` (to not block the current directory walk) and `entries.par_iter()` with some threshold like `.with_min_len(100)` (to distribute large directory walk.)

Uploaded this version. It's pretty okay on my Linux machine, ext4. The last patch is new, which I originally planned to send as a follow up PR. I'll probably split the first half instead.

lib/src/local_working_copy.rs
arxanas: I looked at my previous code and it seems like I misremembered, and it was always traversing `file_states.all()` even in the Watchman case, sorry. I'm guessing I tried to implement it without materializing the full `file_states` map, but didn't succeed/have time.

> BTW, I noticed deleted: HashSet<_> = file_states.all()... (in the original impl) has measurable cost in gecko-dev benchmark. I think it's better to move filtering to per-directory jobs, and reduce string length to be hashed (full path -> file name).

Agreed, but we might fundamentally struggle with the performance as long as we're always loading and deserializing O(repo) data from disk.

[just an idea] For engineering practicality, we could potentially say "for repos over a certain size, we expect you to use fsmonitor, and will optimize the fsmonitor snapshotting specifically" rather than trying to optimize both the non-fsmonitor and fsmonitor snapshotting cases.

[aside] For the Watchman case, I suspect I may have tried to load the previous tree for comparison rather than checking `file_states`. The idea would be that we only need to check O(change) entries instead of O(repo) entries. But it's possible that either I didn't really try it, or, if I tried it, then it was too slow to fetch too many arbitrary objects from disk, whereas even the O(repo) sequential loading of `file_states` is still faster than many random accesses.

lib/src/local_working_copy.rs
arxanas: nit (non-blocking): We could write `if dir.is_root() { "/".len() } else { 0 }` to make it clearer what's happening here, but not a big deal regardless.

lib/src/local_working_copy.rs
yuja: > we might fundamentally struggle with the performance as long as we're always loading and deserializing O(repo) data from disk.

We might need to redesign the tree state format (to be e.g. `mmap()`-able), but I don't have any particular idea in mind. It might be also good to use a committed tree object as you said.

yuja:
Updated the benchmark. It now includes `nixpkgs` result.

If needed, I'll extract the first 3 patches to separate PR. They just move codes around. The last patch wasn't included in the original PR. It can also be split to new PR.

lib/src/local_working_copy.rs
```diff
@@ -1097,36 +1091,7 @@ impl FileSnapshotter<'_> {
...
-                            if let Some(new_file_state) = file_state(&metadata) {
-                                self.process_present_file(
-                                    tracked_path.to_owned(),
-                                    &disk_path,
-                                    Some(&current_file_state),
-                                    new_file_state,
-                                )?;
-                            }
-                        }
+                        self.visit_tracked_files(file_states)?;
```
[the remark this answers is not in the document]
yuja: I've copied "visit only paths we're already tracking" to doc comment, but the original comment isn't removed from the call site. It seemed better to mention that ignored-directory is handled differently.

lib/src/local_working_copy.rs
```diff
@@ -304,6 +336,25 @@ impl<'a> FileStates<'a> {
...
+        let to_name = coerce(|path| {
+            let tail = path.get(prefix_len..).unwrap_or("");
+            tail.split_once('/').map_or(tail, |(name, _)| name)
+        });
+        let start = self
+            .data
+            .partition_point(|entry| to_name(&entry.path) < base.as_internal_str());
+        let len = self.data[start..]
+            .partition_point(|entry| to_name(&entry.path) == base.as_internal_str());
+        start..(start + len)
```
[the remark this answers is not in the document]
yuja: Ok, I've inlined it. There are only two callers, and the function is relatively short.

lib/src/local_working_copy.rs
```diff
@@ -272,6 +273,13 @@ impl<'a> FileStates<'a> {
         Self::from_sorted(&self.data[range])
     }
 
+    /// Suppose all entries share the same prefix `dir`, returns file states
+    /// under the `<dir>/<base>` path.
```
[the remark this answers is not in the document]
yuja: Good point. Updated the doc, thanks!

PR #5032, opened 2024-12-05
docs: add "Release highlights" section to latest release (v0.24.0)

arxanas:
See commit message.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Release highlights
+
 ### Breaking changes
 
 ### Deprecations
@@ -17,6 +19,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed bugs
 
 ## [0.24.0] - 2024-12-04
+
+### Release highlights
+
+* New [`jj absorb`](https://martinvonz.github.io/jj/latest/cli-reference/#jj-absorb) command automatically squashes changes from the current commit into relevant ancestor commits.
+
+* Experimental dynamic shell completions have been added; see [the docs](https://martinvonz.github.io/jj/latest/install-and-setup/#command-line-completion) for configuration.
+
+* [`jj duplicate`](https://martinvonz.github.io/jj/latest/cli-reference/#jj-duplicate) now accepts `--destination`/`--insert-before`/`--insert-after`.
+
+* Some deprecated commands have been removed (`jj move`, `jj checkout`, `jj merge`).
 
 ### Breaking changes
 


CHANGELOG.md
```diff
@@ -18,6 +18,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.24.0] - 2024-12-04
 
+### Release highlights
+
+* New [`jj absorb`](https://martinvonz.github.io/jj/latest/cli-reference/#jj-absorb) command automatically moves changes from the current commit into relevant ancestor commits.
```
martinvonz, line 23: Should we say "squashes" instead of "moves" to hopefully make it clearer to user who are familiar with the `jj squash` command?

CHANGELOG.md
```diff
@@ -18,6 +18,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.24.0] - 2024-12-04
 
+### Release highlights
```
martinvonz, line 21: Should we add this heading under "[Unreleased]" above too? I'm not sure. Pros: I won't forget to add it. Cons: Contributors might think they're supposed to decide whether to update it in their own PR (they shouldn't IMO).

arxanas: I think if I were a new contributor, I would just ignore the "release highlights" section if I were adding an entry to the changelog, not having the hubris to think that my change must be a release highlight 🤣

CHANGELOG.md
arxanas: Sure, but FYI I copied the description from the `jj absorb` command itself, which says "moves" rather than "squashes". (I think the `jj absorb` docs should be improved separately because it doesn't say anything like that it tries to move lines/hunks to the most recent commit that touched the same lines/hunks, so it's a bit mysterious.)

CHANGELOG.md
arxanas: Added it to the unreleased template.

CHANGELOG.md
arxanas: Changed "moves" -> "squashes".

martinvonz (approved):

thoughtpolice (approved):

martinvonz:
Will you update https://github.com/martinvonz/jj/releases/tag/v0.24.0 to match or should I?

arxanas:
@martinvonz I updated the release and the associated GitHub Discussions post. However, I kept all the headings at `h3` (whereas my previous proposal put `About`, `Release highlights`, `Changelog`, and `Contributors` at `h2`), since I wanted to make it easy to be uniform in the future. Since I'm assuming you'll be doing the releases for the foreseeable future, you can adjust the formatting as desired.

PR #5037, opened 2024-12-06
config: don't use ConfigBuilder to merge tables, clean up one-line Ui functions

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/config.rs
+++ b/lib/src/config.rs
@@ -17,6 +17,7 @@
 use std::borrow::Borrow;
 use std::convert::Infallible;
 use std::fmt;
+use std::mem;
 use std::ops::Range;
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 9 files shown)


PhilipMetzger (approved):
LG

> Some of Vec<ConfigLayer> could be changed to Vec<config::Config> (or
Vec<ConfigTable>), but I'm not sure which is better.

What would be the major difference here?

lib/src/config.rs
```diff
@@ -314,6 +325,21 @@ impl ConfigLayer {
     ) -> Result<Option<&ConfigValue>, &ConfigValue> {
         look_up_item(&self.data.cache, name.into_name_path().borrow())
     }
+
+    /// Inserts new `value` at the `name` path.
+    pub fn insert_value(
```
martinvonz, line 330: Would it make sense to call it `set_value()`? The goal is to make it not sound like it's going to insert it in any particular position. Because there's no ordering implied within a layer, right? I'm guessing you used "insert" in the `HashMap::insert` sense (I'm not sure why that isn't called  "set").

martinvonz (approved):

yuja:
> LG
> 
> > Some of Vec could be changed to Vecconfig::Config (or
> > Vec), but I'm not sure which is better.
> 
> What would be the major difference here?

Whether or not caller can/has to specify the `layer` parameter.

lib/src/config.rs
yuja: Yes, I just copied `insert` because the underlying function is called as such. `set` also works.

PR #5043, opened 2024-12-07
Better respect `$NO_COLOR`

### LINKED ISSUE #5041: NO_COLOR is not correctly overridden by ui.color config
According to the docs, having NO_COLOR set in the environment, but having `ui.color` set to something like `always` should cause the NO_COLOR environment variable to get overridden, but it doesn't appear to work.

1. `export NO_COLOR=1`
1. `jj diff` shows no color as expected
1. add `color="always"` to `[ui]` section of `~/.config/jj/config.toml`
1. `jj diff` still shows no color. interestingly, if `ui.diff.format` is set to `git`, the output shows some **bold** characters, but still no color. Perhaps this is a hint to where the issue lies.

- Platform: Tested on Mac + Linux
- Version: 0.24.0

martinvonz:
Closes #5041 

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* The `$NO_COLOR` environment variable must now be non-empty to be respected.
+
 ## [0.24.0] - 2024-12-04
 
 ### Release highlights
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3627,6 +3627,8 @@ impl CliRunner {
             .fold(builder, |builder, config| builder.add_source(config))
             .build()
             .unwrap();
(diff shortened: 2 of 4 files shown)


PhilipMetzger (approved):
Thanks

cli/src/formatter.rs
```diff
@@ -161,6 +161,7 @@ impl FormatterFactory {
     }
 
     pub fn color(config: &StackedConfig, debug: bool) -> Result<Self, ConfigGetError> {
+        crossterm::style::force_color_output(true);
```
yuja, line 164: nit: Since this is a global flag, it might be better to do this in `CliRunner` or somewhere closer to `main()`.

yuja (approved):

cli/src/formatter.rs
martinvonz: Yeah, I wasn't sure where the best place would be. I moved it to `CliRunner::run()`. PTAL.

yuja (approved):

PR #5047, opened 2024-12-07
build: switch back to `gix/max-performance-safe` feature by default

arxanas:
See commit message, plus discussions at
https://discord.com/channels/968932220549103686/1314478258636718081 and
https://discord.com/channels/968932220549103686/969291218347524238/1315041069032804383.

Before committing, I'll need to update the docs to clarify the `cmake` dependency situation.

# Checklist

If applicable:


cli/Cargo.toml
```diff
@@ -111,7 +111,7 @@ jj-cli = { path = ".", features = ["test-fakes"], default-features = false }
 [features]
 default = ["watchman"]
 bench = ["dep:criterion"]
-packaging = []
+packaging = ["jj-lib/gix-max-performance"]
```
yuja, line 114: nit: might be better to alias `gix-max-performance` through cli/Cargo.toml (like `watchman`)?

yuja (approved):
iirc, @thoughtpolice's idea was to drop cmake dependency from lib-nz-sys. If that works, we won't need the feature flag.

We'll also need to revert the doc change b7ba3fc0.

cli/Cargo.toml
arxanas: Makes sense, I added the top-level `gix-max-performance` feature to `cli/Cargo.toml` and had `packaging` reference it.

arxanas:
> iirc, @thoughtpolice's idea was to drop cmake dependency from lib-nz-sys. If that works, we won't need the feature flag.

That would be great 😄 

> We'll also need to revert the doc change https://github.com/martinvonz/jj/commit/b7ba3fc0be734cfa39bb7d6cfc50ed67eee15dce.

Reverted in this PR.

PR #5050, opened 2024-12-07
bookmark: Make `bookmark {create, set, move}` unhide hidden commits

PhilipMetzger:
This was [discussed in Discord][dc] a while ago,  and this is the logical and consistent
conclusion. Implementing it as such makes it consistent with both `jj edit` and `jj new`
which unhide any predecessor.

# Checklist

If applicable:

[dc]: https://discord.com/channels/968932220549103686/968932220549103689/1298729954670022676

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj bookmark create`, `jj bookmark set` and `jj bookmark move` onto a hidden
+   commit make it visible.
+
 * `jj undo` now shows a hint when undoing an undo operation that the user may
    be looking for `jj op restore` instead.
 
--- a/cli/src/commit_templater.rs
+++ b/cli/src/commit_templater.rs
@@ -984,10 +984,7 @@ fn builtin_commit_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, Comm
         |language, _diagnostics, _build_ctx, self_property, function| {
             function.expect_no_arguments()?;
             let repo = language.repo;
-            let out_property = self_property.map(|commit| {
-                let maybe_entries = repo.resolve_change_id(commit.change_id());
-                maybe_entries.map_or(true, |entries| !entries.contains(commit.id()))
-            });
+            let out_property = self_property.map(|commit| commit.is_hidden(repo));
             Ok(L::wrap_boolean(out_property))
         },
     );
--- a/lib/src/commit.rs
+++ b/lib/src/commit.rs
@@ -157,6 +157,12 @@ impl Commit {
         &self.data.committer
     }
 
+    ///  A commit is hidden if its commit id is not in the change id index.
+    pub fn is_hidden(&self, repo: &dyn Repo) -> bool {
+        let maybe_entries = repo.resolve_change_id(self.change_id());
+        maybe_entries.map_or(true, |entries| !entries.contains(&self.id))
+    }
+
     /// A commit is discardable if it has no change from its parent, and an
     /// empty description.
     pub fn is_discardable(&self, repo: &dyn Repo) -> BackendResult<bool> {
--- a/lib/tests/test_mut_repo.rs
+++ b/lib/tests/test_mut_repo.rs
@@ -503,6 +503,7 @@ fn test_has_changed() {
     mut_repo.set_local_bookmark_target("main", RefTarget::normal(commit2.id().clone()));
     assert!(mut_repo.has_changes());
     mut_repo.set_local_bookmark_target("main", RefTarget::normal(commit1.id().clone()));
+    mut_repo.remove_head(commit2.id());
     assert!(!mut_repo.has_changes());
 
     mut_repo.set_remote_bookmark("main", "origin", normal_remote_ref(commit2.id()));
@@ -714,4 +715,28 @@ fn test_reparent_descendants() {
             assert_ne!(parent_ids, rewritten_parent_ids);
         }
     }
+}
+
+#[test]
+fn test_bookmark_hidden_commit() {
+    // Test that MutableRepo::set_local_bookmark_target() on a hidden commit makes
+    // it visible.
+    let test_repo = TestRepo::init();
+    let repo = &test_repo.repo;
+    let root_commit = repo.store().root_commit();
+
+    let mut tx = repo.start_transaction();
+    let wc_commit = write_random_commit(tx.repo_mut());
+
+    // Intentionally not doing tx.commit, so the commit id is not tracked
+    // in the view head ids.
+
+    let mut tx = repo.start_transaction();
+    tx.repo_mut()
+        .set_local_bookmark_target("b", RefTarget::normal(wc_commit.id().clone()));
+    let repo = tx.commit("test").unwrap();
+    assert_eq!(
+        *repo.view().heads(),
+        hashset! {wc_commit.id().clone(), root_commit.id().clone()}
+    );
 }
--- a/cli/tests/test_bookmark_command.rs
+++ b/cli/tests/test_bookmark_command.rs
@@ -1079,9 +1079,9 @@ fn test_bookmark_track_conflict() {
     insta::assert_snapshot!(stderr, @r###"
     Started tracking 1 remote bookmarks.
     main (conflicted):
-      + qpvuntsm e802c4f8 (empty) b
-      + qpvuntsm hidden 427890ea (empty) a
-      @origin (behind by 1 commits): qpvuntsm hidden 427890ea (empty) a
+      + qpvuntsm?? e802c4f8 (empty) b
+      + qpvuntsm?? 427890ea (empty) a
+      @origin (behind by 1 commits): qpvuntsm?? 427890ea (empty) a
     "###);
 }
 
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -877,6 +877,7 @@ impl MutableRepo {
     }
 
     pub fn has_changes(&self) -> bool {
+        self.view.ensure_clean(|v| self.enforce_view_invariants(v));
         !(self.parent_mapping.is_empty() && self.view() == &self.base_repo.view)
     }
 
@@ -1521,7 +1522,11 @@ impl MutableRepo {
     }
 
     pub fn set_local_bookmark_target(&mut self, name: &str, target: RefTarget) {
-        self.view_mut().set_local_bookmark_target(name, target);
+        let view = self.view_mut();
+        for id in target.added_ids() {
+            view.add_head(id);
+        }
+        view.set_local_bookmark_target(name, target);
     }
 
     pub fn merge_local_bookmark(
@@ -1534,7 +1539,7 @@ impl MutableRepo {
         let index = self.index.as_index();
         let self_target = view.get_local_bookmark(name);
         let new_target = merge_ref_targets(index, self_target, base_target, other_target);
-        view.set_local_bookmark_target(name, new_target);
+        self.set_local_bookmark_target(name, new_target);
     }
 
     pub fn get_remote_bookmark(&self, name: &str, remote_name: &str) -> RemoteRef {


tim-janik:
> This was [discussed in Discord](https://discord.com/channels/968932220549103686/968932220549103689/1298729954670022676) a while ago, while I'm still in favor of making it a hard error down the line, it currently is just implemented as a warning to get user feedback.
> 
> TODO: write tests and update CHANGELOG.md
> # Checklist
> 
> If applicable:
> 
>     * [ ]  I have updated `CHANGELOG.md`
> 
>     * [ ]  I have updated the documentation (README.md, docs/, demos/)
> 
>     * [ ]  I have updated the config schema (cli/src/config-schema.json)
> 
>     * [ ]  I have added tests to cover my changes

IMHO, adding a tag or a bookmark to a hidden commit should make it visible.
It is one of the first things newbies try to resurrect a hidden commit.
I wonder what the rationale is for issuing a warning or an error in this case.
As far as my experience goes, adding tags or bookmarks to hidden commits is normally a deliberate action, do you know of cases where this would frequently happen in error?
And even if so, what is the rationale for `jj new <hiddencommit>` being OK and `jj b c <hiddencommit>` not?
(Honestly asking, I really don't get it)

PhilipMetzger:
> IMHO, adding a tag or a bookmark to a hidden commit should make it visible.
> It is one of the first things newbies try to resurrect a hidden commit.

Have a reference for that, beside you? 

> I wonder what the rationale is for issuing a warning or an error in this case.
> As far as my experience goes, adding tags or bookmarks to hidden commits is normally a deliberate action, do you know of cases where this would frequently happen in error?

While it works, it shouldn't really be a option since those commits should be addressable in the _perfect model_ anyway. It is a mistake to address a predecessor as existing in the repo. 

> And even if so, what is the rationale for `jj new <hiddencommit>` being OK and `jj b c <hiddencommit>` not?
> (Honestly asking, I really don't get it)

`jj new` resurfaces it back in the log, while any `bookmark` command won't.

cli/src/commands/bookmark/create.rs
```diff
@@ -46,7 +48,9 @@ pub fn cmd_bookmark_create(
     let mut workspace_command = command.workspace_helper(ui)?;
     let target_commit = workspace_command
         .resolve_single_rev(ui, args.revision.as_ref().unwrap_or(&RevisionArg::AT))?;
+    let repo = workspace_command.repo();
     let view = workspace_command.repo().view();
+    let is_hidden = target_commit.is_hidden(repo.deref())?;
```
yuja, line 53: nit: `.as_ref()` (or something like `&*repo.)

yuja (commented):
> adding a tag or a bookmark to a hidden commit should make it visible.

That's a good point. I'm not sure it's useful in practice, but the behavior is consistent.
(Un-hidden commits can be warned, but that would also apply to `jj new`.)

tim-janik:
> > IMHO, adding a tag or a bookmark to a hidden commit should make it visible.
> > It is one of the first things newbies try to resurrect a hidden commit.
> 
> Have a reference for that, beside you?

You might find some if you look for "branch" and "hidden" on discord.

> > And even if so, what is the rationale for `jj new <hiddencommit>` being OK and `jj b c <hiddencommit>` not?
> > (Honestly asking, I really don't get it)
> 
> `jj new` resurfaces it back in the log, while any `bookmark` command won't.

I know the current state of affairs.
I am asking for a reason, because I don't see the logic behind the current inconsistency.
Can you give any, please?

joyously:
I'm unsure what makes a commit hidden, and if there could be confusion between "commit" and "change" in this context.
>  It is a mistake to address a predecessor as existing in the repo.

Would this apply to the use case of taking the current `jj` repo and making a bookmark for each release that has come out?

martinvonz:
> I am asking for a reason, because I don't see the logic behind the current inconsistency.
> Can you give any, please?

The set of visible commits is defined by a set of visible heads. Some pretty low-level library code adds new commits to that set because we generally want new commits to be visible. Until #4427, `jj edit` wouldn't even update the set (because it doesn't create a new commit). So I think the inconsistency is not because we decided that that's best for the user as it is because we haven't thought much about it. I think it makes sense to make a commit visible if we move a bookmark to it.

PhilipMetzger:
> > > IMHO, adding a tag or a bookmark to a hidden commit should make it visible.
> > > It is one of the first things newbies try to resurrect a hidden commit.
> > 
> > 
> > Have a reference for that, beside you?
> 
> You might find some if you look for "branch" and "hidden" on discord.

The only conversation which talks about this behavior is already mentioned, so this clearly doesn't qualify for a "reference beside you". 

> So I think the inconsistency is not because we decided that that's best for the user as it is because we haven't thought much about it. I think it makes sense to make a commit visible if we move a bookmark to it.

That also works for me,

I actually made this patchset explicitly a warning to see how many complaints it will receive when it lands, since we don't know how many people actually use the workflow @tim-janik argues for.

tim-janik:
> > > Have a reference for that, beside you?
> > 
> > You might find some if you look for "branch" and "hidden" on discord.
> 
> The only conversation which talks about this behavior is already mentioned, so this clearly doesn't qualify for a "reference beside you".

There are several discussions if you bother searching for "hidden branch" or "branch hidden" on the jj discord server, but AFAIK there is no way to link to searches on discord. You could also search on GH and might find #4537, etc, but really, this point is nothing more than a distraction.

PhilipMetzger:
> There are several discussions if you bother searching for "hidden branch" or "branch hidden" on the jj discord server, but AFAIK there is no way to link to searches on discord.

You can link the individual messages, so provide them if you care so much instead calling my argument a shallow dismissal.

cli/src/commands/bookmark/create.rs
PhilipMetzger: done

cli/src/commit_templater.rs
```diff
@@ -774,8 +774,8 @@ fn builtin_commit_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, Comm
             function.expect_no_arguments()?;
             let repo = language.repo;
             let out_property = self_property.map(|commit| {
-                let maybe_entries = repo.resolve_change_id(commit.change_id());
-                maybe_entries.map_or(true, |entries| !entries.contains(commit.id()))
+                let value = commit.is_hidden(repo);
+                value.unwrap_or_default()
```
martinvonz, line 778: nit: inline `value`? It's so short now that I think it's easier to read as one expression

cli/src/commands/bookmark/create.rs
martinvonz, line 1: > Implementing it as such makes it consistent with both `jj edit` and `jj new`
which unhide any predecessor.

I find "predecessor" a bit confusing the unhidden commit doesn't need to be a predecessor of the current commit, or a predecessor of any commit at all for that matter (maybe it was just abandoned).

cli/src/commands/bookmark/create.rs
```diff
@@ -53,7 +53,9 @@ pub fn cmd_bookmark_create(
     let mut workspace_command = command.workspace_helper(ui)?;
     let target_commit = workspace_command
         .resolve_single_rev(ui, args.revision.as_ref().unwrap_or(&RevisionArg::AT))?;
+    let repo = workspace_command.repo();
     let view = workspace_command.repo().view();
```
martinvonz, line 57: nit: use `repo` here

martinvonz (commented):
I wonder if we should make the commit at a lower level so we don't risk missing it. Like this: https://github.com/jj-vcs/jj/compare/main...push-uulxprvmmtsw

The drawback is that if some caller really wants to point a bookmark to a hidden commit, then they'll have to set the bookmark and then make the commit hidden afterwards (and be careful to only do that if it was already hidden).

cli/src/commands/bookmark/create.rs
```diff
@@ -53,7 +53,9 @@ pub fn cmd_bookmark_create(
...
+    let is_hidden = target_commit.is_hidden(repo.as_ref())?;
```
yuja, line 58: nit: I think `is_hidden()` test is redundant (in all commands.) iirc, these commands check if bookmarks are moved/created, so we can just do `.add_head()` if there are any bookmark changes.

yuja:
> I wonder if we should make the commit at a lower level so we don't risk missing it. Like this: [main...push-uulxprvmmtsw](https://github.com/jj-vcs/jj/compare/main...push-uulxprvmmtsw)

It might be good to add repo method that updates bookmark refs and makes targets visible, but I would expect there's an escape hatch to do low-level thing.

cli/src/commands/bookmark/create.rs
martinvonz: True, the repo seems like a better level to do it. I've updated my branch. I should also add tests for it in jj-lib. Or do you want to take my commit and add tests, @PhilipMetzger?

PhilipMetzger: > I think `is_hidden()` test is redundant (in all commands.) iirc, these commands check if bookmarks are moved/created, so we can just do `.add_head()` if there are any bookmark changes.

I agree, but it was the simplest way to implement it. 

> True, the repo seems like a better level to do it. I've updated my branch. I should also add tests for it in jj-lib. Or do you want to take my commit and add tests, @PhilipMetzger?

I in general agree with the direction we take here, so I probably should take over Martin's branch and drop the first commit, if it otherwise doesn't provide value.

PhilipMetzger:
> The drawback is that if some caller really wants to point a bookmark to a hidden commit, then they'll have to set the bookmark and then make the commit hidden afterwards (and be careful to only do that if it was already hidden).

Is that a thing we actually want? As it doesn't make sense from the viewpoint of the CLI, like we can provide a API for that but it should be a error otherwise.

yuja:
> > The drawback is that if some caller really wants to point a bookmark to a hidden commit, then they'll have to set the bookmark and then make the commit hidden afterwards (and be careful to only do that if it was already hidden).
> 
> Is that a thing we actually want? As it doesn't make sense from the viewpoint of the CLI, like we can provide a API for that but it should be a error otherwise.

Callers might depends on property that `set()` + `unset()` is noop. It's probably better to rename the function if it does `set()` + `add_heads()`.

cli/src/commands/bookmark/create.rs
PhilipMetzger: dropped the commit, so this is no longer necessary

cli/src/commit_templater.rs
PhilipMetzger: done

lib/tests/test_mut_repo.rs
```diff
@@ -715,3 +716,52 @@ fn test_reparent_descendants() {
         }
     }
 }
+
+#[test]
+fn test_bookmark_hidden_commit() {
+    // Test that MutableRepo::set_local_bookmark_target() on a hidden commit updates
+    // the view head ids as setting it makes the commit visible.
```
martinvonz, line 723: "updates the view head ids as setting it makes the commit visible" sounds backwards (it becomes visible because we make it a visible head). I think just "makes it visible" is enough

lib/tests/test_mut_repo.rs
```diff
@@ -715,3 +716,52 @@ fn test_reparent_descendants() {
...
+        .set_local_bookmark_target("b", RefTarget::normal(wc_commit.id().clone()));
+    let repo = tx.commit("test").unwrap();
+    assert_eq!(
+        *repo.view().heads(),
+        hashset! {wc_commit.id().clone(), root_commit.id().clone()}
+    );
+}
+
+#[test]
+fn test_bookmark_hidden_predecessor() {
```
martinvonz, line 745: This looks like effectively the same test case as the one above (you just create an extra `wc_id` variable that's unused AFAICT). I don't think we need a separate test for unhiding a predecessor, so I would just delete this test.

lib/src/commit.rs
```diff
@@ -157,6 +157,12 @@ impl Commit {
         &self.data.committer
     }
 
+    ///  A commit is hidden, if its commit id is not in the predecessor set.
```
martinvonz, line 160: nit: "not in the predecessor set" is not correct. Maybe "not in the change id index". (I think the comma should also be removed.)

lib/tests/test_mut_repo.rs
PhilipMetzger: done

lib/tests/test_mut_repo.rs
PhilipMetzger: I copied it because I wanted to write a second test, but it is gone now.

lib/src/commit.rs
PhilipMetzger: done

lib/tests/test_mut_repo.rs
PhilipMetzger: should be fixed now

CHANGELOG.md
```diff
@@ -308,6 +308,9 @@ changes. Two select improvements:
 
 ### New features
 
+* `jj bookmark create`, `jj bookmark set` and `jj bookmark move` onto a hidden
+  commit make it visible, making the behavior consistent across commands.
```
martinvonz, line 312: nit: I'm not sure users will understand what "making the behavior consistent across commands" means. I would just remove it. I don't think many users are going to wonder why, so I don't think we need to explain the rationale like we've done a few times before.

martinvonz (approved):

ilyagr:
Thanks Philip! I imagine the current behavior is confusing a lot of people.

---------

Unrelated to this PR: I'm not sure what the long-term behavior should be. As history becomes more accessible, I hope people will want to refer to hidden commits (and thus maybe label them) more and more often.

One long-term possibility I like would be to forbid bookmarks from pointing to hidden commits entirely, but have good support for tags pointing to hidden commits (as well as pushing/fetching tags). Perhaps a bookmark could even become a pointer to a change id, while tags would continue pointing to commit ids. This would extend the common use of tags for labeling releases. Obviously, this would require a lot more redesign and work (especially if we also introduce the third concept of topics in the meantime).

In the shorter term, I'd consider allowing `jj abandon` to keep bookmarks at the hidden commits instead of deleting (or sliding) them. However, I don't think this is urgent; for now people seem to refer to hidden commits quite rarely.

PhilipMetzger:
> Unrelated to this PR: I'm not sure what the long-term behavior should be. As history becomes more accessible, I hope people will want to refer to hidden commits (and thus maybe label them) more and more often.
> 
> One long-term possibility I like would be to forbid bookmarks from pointing to hidden commits entirely, but have good support for tags pointing to hidden commits (as well as pushing/fetching tags). Perhaps a bookmark could even become a pointer to a change id, while tags would continue pointing to commit ids. This would extend the common use of tags for labeling releases. Obviously, this would require a lot more redesign and work (especially if we also introduce the third concept of topics in the meantime).

I think long-term the goal should be to make it (bookmarking a hidden commit) an error, like I outlined in the Discord discussion and initial version of this patch. I am also of the opinion that a native Jujutsu has no need for tags, see [this Discord message][dc]. Since most of this is unrelated to this patch, we should move the discussion to a new issue. 

> In the shorter term, I'd consider allowing `jj abandon` to keep bookmarks at the hidden commits instead of deleting (or sliding) them. However, I don't think this is urgent; for now people seem to refer to hidden commits quite rarely.

That wouldn't match my model, so I'd disallow relatively soon. 

[dc]: https://discord.com/channels/968932220549103686/968932220549103689/1314277847937847356

lib/src/commit.rs
```diff
@@ -157,6 +157,12 @@ impl Commit {
         &self.data.committer
     }
 
+    ///  A commit is hidden if its commit id is not in the change id index.
+    pub fn is_hidden(&self, repo: &dyn Repo) -> BackendResult<bool> {
```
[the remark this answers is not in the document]
PhilipMetzger: done

CHANGELOG.md
PhilipMetzger: done

PR #5051, opened 2024-12-08
diff_util: Add the short alias `S` for `--stat`

PhilipMetzger:
This was a suggestion from sm in the libera #jujutsu IRC chat.

# Checklist

If applicable:

### DIFF
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -92,7 +92,7 @@ pub struct DiffFormatArgs {
     #[arg(long, short)]
     pub summary: bool,
     /// Show a histogram of the changes
-    #[arg(long)]
+    #[arg(long, short_alias = 'S')]
     pub stat: bool,
     /// For each path, show only its type before and after
     ///
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* The `--stat` option for `jj log`, `jj show` and `jj diff` now has a
+  short alias `-S`
+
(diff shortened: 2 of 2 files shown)


simonmichael:
More context, in case it helps:
```
what would you think of adding -S as short flag for --stat (for jj diff, eg)
"like -s, but louder"
the value could be having a discoverable convenient shortcut for all.. if it's not too clashy
```
I was repeatedly running diff (or log) with successively more detail:
```
jj diff, jj diff -s, jj diff --stat, jj diff -p
and using terminal history to edit the previous command. The --stat is just a little awkward to type each time. 
```

cli/src/diff_util.rs
```diff
@@ -92,7 +92,7 @@ pub struct DiffFormatArgs {
     #[arg(long, short)]
     pub summary: bool,
     /// Show a histogram of the changes
-    #[arg(long)]
+    #[arg(long, short_alias = 'S')]
```
martinvonz, line 95: I think this should be `short` instead of `short_alias` (it's strange to define an alias for something that doesn't exist, which is what I think this does).

martinvonz (commented):
There are only ~50 possible short flags, so we need to be careful about adding new ones. I don't think we've heard the request often enough that we should add a flag.

PhilipMetzger:
> There are only ~50 possible short flags, so we need to be care.  I don't think we've heard the request often enough that we should add a flag.

Agreed, so I'll leave it up for @simonmichael until we hear more demand for it,

simonmichael:
Waiting to see if this is worthwhile is quite understandable.

cli/src/diff_util.rs
PhilipMetzger: done

PR #5052, opened 2024-12-08
docs: revert string formatting change of "jj help util exec"

yuja:
cc @arxanas 

# Checklist

If applicable:


arxanas (approved):
Interesting, I'm surprised that CI didn't fail before?

yuja:
I don't think we rely on doc tests. I wouldn't notice the problem unless I ran `cargo insta test`.

arxanas:
> I don't think we rely on doc tests. I wouldn't notice the problem unless I ran `cargo insta test`.

I see. I opened #5056 to check docs in CI.

Issue #5049, opened 2024-12-07
git diff output incompatibility with BSD patch for new files


botovq:
## Description

The diff hunk header between two pairs of `@` for a newly created file indicates that it applies to the first line of `/dev/null` file and confuses the patch implementations on FreeBSD and OpenBSD which then ask for a filename to be specified when one tries to apply the patch.

## Steps to Reproduce the Problem

Create and commit a file containing a single line, generate patches from that commit and try to apply the patches in an empty directory.

Git's output has a hunk indicator `@@ -0,0 +1 @@`:
  
  ```console
  $ git show
  [...]
  diff --git a/hello b/hello
  new file mode 100644
  index 0000000..ce01362
  --- /dev/null
  +++ b/hello
  @@ -0,0 +1, @@
  +hello
  ```

 whereas the equivalent output of jj has `@@ -1,0 +1,1 @@`:

```console
$ jj show --git -r @-
[...]
diff --git a/hello b/hello
new file mode 100644
index 0000000000..ce01362503
--- /dev/null
+++ b/hello
@@ -1,0 +1,1 @@
+hello
```

The problem is the first `-1` here. The BSD's patch implementations (derived from Larry Wall's patch) special case `@@ 0` and use that as a hint that a new file needs to be created. Whether that's a great approach is a different question, but it has been this way for more than thirty years. If it's not `@@ 0`, it doesn't find the file and enters interactive mode, which doesn't work.

A simlar problem exists for the reverted diff, but it doesn't confuse BSD patch.

I *think* the following passage in the [POSIX 2024 specification][1] indicates that git's behavior is the expected one:

> `"%1d,%1d", <beginning line number>, <number of lines>`
>
> otherwise. If a range is empty, its beginning line number shall be the number of the line just before the range, or 0 if the empty range starts the file.

## Expected Behavior

The diffs generated by git and jujutsu apply cleanly for the system-provided patch and create files as needed.

## Actual Behavior

Attempting to apply the diffs in an empty directory results in the following ouptut on FreeBSD (tested on 14.2) and OpenBSD (tested on 7.6):

```console
$ patch < /tmp/jj.diff
Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|Commit ID: 58258525c7bc0bb5afa9c672df4570600da27803
|Change ID: nzylvuozztzlpxkovltwqukupmmzwmlw
|Bookmarks: master master@git
[...]

PR #5054, opened 2024-12-08
diff: fix unified diff to emit one-lower start line number for empty hunks

yuja:
#5049

# Checklist

If applicable:


botovq:
FWIW, this is along the lines I expected the fix to look like and works for me in the scenarios where I hit this bug. The test output changes all look reasonable to me. Many thanks!

martinvonz (approved):

PR #5056, opened 2024-12-08
ci: check Rust documentation

arxanas:
Follow-up from https://github.com/martinvonz/jj/pull/5052. This isn't critical, so we can drop this PR if we think it's not important or not worth resolving the outstanding questions for now.

Questions:

- What is our policy/strategy for new top-level checks vs embedding steps into existing checks?
  - I chose to make new top-level checks since I figured the feedback loop would be faster that way:
    - When a problem occurs, it's obvious at the top-level display that it's not a "real" test failure, just a rustdoc issue.
    - When you're fixing the problem, you don't have to wait the entire test suite.
  - However, it probably induces additional overall load to compile/process the codebase again?
- What is our policy for setting the Rust toolchain for the new checks?
  - I chose MSRV for the doctests themselves, since I imagined that they could demonstrate the API boundary, which needs to adhere to MSRV.
    - See below: we don't have any "real" doctests right now. So it's mostly immaterial for now.
  - I chose `stable` for `cargo doc` in case new lints get added.
    - It could be annoying to have to resolve lints that don't appear unless you have the latest `stable` toolchain.
    - We're already doing this for `cargo clippy`, so I don't think it should be much worse than the status quo?

Notes:

- These checks are mostly unrelated to the doc website generation.
  - Feel free to suggest any naming changes to the jobs/steps/etc. that make it clearer.
  - The exception is that website might process the same information from the CLI doc-comments, similarly to the original issue in #5052, but this is incidental.
- We don't actually have any doctests at present.
  - So this mainly just detects "breaking the doctest build".
  - We might add some in the future to demonstrate how to use the library?
  - I confirmed that `cargo test --doc` catches the issue in #5052.
- I fixed the few existing rustdoc issues.
- I highly recommend linking symbols in docs when possible (with square brackets, like <code>[\`Foo\`]</code>):
  - It makes the docs easier to navigate (both in the IDE and the crate documentation website).
  - With this PR, we can check validity of those references automatically to ensure the docs are up-to-date (via `rustdoc::broken_intra_doc_links`).
- In the current setup, I added new non-required top-level checks.
  - They're not "required" checks unless/until somebody goes into the GitHub repo configuration and adds them.

# Checklist

If applicable:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -118,6 +118,26 @@ jobs:
           components: rustfmt
       - run: cargo +nightly fmt --all -- --check
 
+  check-doctests:
+    name: Run doctests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - uses: dtolnay/rust-toolchain@1482605bfc5719782e1267fd0c0cc350fe7646b8
+        with:
+          toolchain: 1.76
+      # NOTE: We need to run `cargo test --doc` separately from normal tests:
+      # - `cargo build --all-targets` specifies: "Build all targets"
+      # - `cargo test --all-targets` specifies: "Test all targets (does not include doctests)"
+      - name: Run doctests
+        run: cargo test --workspace --doc
+        env:
+          RUST_BACKTRACE: 1
+      - name: Check `cargo doc` for lint issues
+        env:
+          RUSTDOCFLAGS: "--deny warnings"
+        run: cargo doc --workspace --no-deps
+
   mkdocs:
     name: Check that MkDocs can build the docs
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -118,6 +118,35 @@ jobs:
...
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        - uses: dtolnay/rust-toolchain@1482605bfc5719782e1267fd0c0cc350fe7646b8
+          with:
+            toolchain: stable
+        - name: Check `cargo doc` for warnings
+          env:
+            RUSTDOCFLAGS: "--deny warnings"
+          run: cargo doc --workspace --no-deps
```
martinvonz, line 148: What do you think about combining these two since they're closely related?

arxanas: I don't feel strongly. Updated it to combine them.

martinvonz (approved):

PR #5058, opened 2024-12-09
Fix performance of progress bar in Git fetch

### LINKED ISSUE #5057: The progress bar causes massive traffic and high CPU usage to the terminal running `jj`
## Description

```
$ jj git clone https://github.com/torvalds/linux.git
Fetching into new repo in "/tmp/linux"
  9% 578.6 MiB at   2.4 MiB/s [███████████                                                               ]
```

The progress bar showed above causes massive traffic and high CPU usage to the terminal running `jj`, if `jj` is run in the remote server, it also causes massive traffic between local machine and remote server.

Ssh login to a remote server, and run `jj git clone https://github.com/torvalds/linux.git /tmp/linux` on the remote server.

Run `sudo nethogs -a` on the remote server to measure the network traffic.

### The progress bar causes massive traffic between local machine and remote server

`jj git clone https://github.com/torvalds/linux.git /tmp/linux`:

Screenshot 2024-12-09 at 10 46 01

### Git does not have this behavior

`git clone https://github.com/torvalds/linux.git /tmp/linux`:

Screenshot 2024-12-09 at 10 47 08

### When the progress bar is redirect to `/dev/null`, the massive traffic is gone

`jj git clone https://github.com/torvalds/linux.git /tmp/linux 2>/dev/null`:

Screenshot 2024-12-09 at 10 50 13

### The progress bar also causes `tmux` high CPU usage when running `jj` in `tmux`:

Screenshot 2024-12-09 at 11 50 26

When running `git clone ..` in `tmux`, you can't even see `tmux` in `top`:

Screenshot 2024-12-09 at 11 46 20

## Expected Behavior

The progress bar do not cause massive traffic and high CPU usage to the terminal.

## Actual Behavior

The progress bar causes massive traffic and high CPU usage to the terminal.

## Specifications

- Platform: Linux and macOS
- Version: 0.23.0

martinvonz:
@Ralith, can you take a look and see if this preserves the behavior you intended?

Closes #5057.

# Checklist

If applicable:

### DIFF
--- a/cli/src/progress.rs
+++ b/cli/src/progress.rs
@@ -33,11 +33,11 @@ impl Progress {
         }
     }
 
-    pub fn update(
+    pub fn update<W: std::io::Write>(
         &mut self,
         now: Instant,
         progress: &git::Progress,
-        output: &mut ProgressOutput,
+        output: &mut ProgressOutput<W>,
     ) -> io::Result<()> {
         use std::fmt::Write as _;
 
@@ -166,7 +166,7 @@ impl RateEstimateState {
 pub fn snapshot_progress(ui: &Ui) -> Option<impl Fn(&RepoPath) + '_> {
     struct State {
         guard: Option<OutputGuard>,
-        output: ProgressOutput,
+        output: ProgressOutput<std::io::Stderr>,
         next_display_time: Instant,
     }
 
@@ -215,6 +215,8 @@ pub fn snapshot_progress(ui: &Ui) -> Option<impl Fn(&RepoPath) + '_> {
 
 #[cfg(test)]
 mod tests {
+    use insta::assert_snapshot;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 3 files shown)


cli/src/progress.rs
```diff
@@ -61,7 +61,15 @@ impl Progress {
...
         }
-        self.next_print = now.min(self.next_print + Duration::from_secs(1) / UPDATE_HZ);
+        let period = Duration::from_secs(1) / UPDATE_HZ;
+        // Try to print periodically by setting the new threshold exactly one period
+        // later
+        self.next_print = self.next_print + period;
+        // But if it's been long enough since last update that we'd effectively print
+        // two updates now, reset the threshold to now.
+        if now >= self.next_print {
+            self.next_print = now + period;
```
Ralith, line 71: Might be simpler to just do this unconditionally? Means more drift, but it's not really important that we exactly nail the update rate.

Ralith (approved):
Seems fine. I guess the issue here is that sometimes it takes much longer than `INITIAL_DELAY` for us to start receiving updates, particularly when cloning huge repos?

martinvonz:
> I guess the issue here is that sometimes it takes much longer than `INITIAL_DELAY` for us to start receiving updates, particularly when cloning huge repos?

No, the issue in the existing code is that the `self.next_print = now.min(...);` is effectively `self.next_print = now;`, so we end up waiting the initial delay and then we never wait again (i.e. we print every update).

cli/src/progress.rs
martinvonz: I'm fine with simplifying it to `self.next_print = now + period;`

Ralith:
Oh, wow, no idea what I(?) was thinking there. Nice catch.

cli/src/progress.rs
```diff
@@ -61,7 +61,15 @@ impl Progress {
             _ = write!(output, "{}", crossterm::cursor::Hide);
             self.guard = Some(guard);
         }
-        self.next_print = now.min(self.next_print + Duration::from_secs(1) / UPDATE_HZ);
+        let period = Duration::from_secs(1) / UPDATE_HZ;
```
istudyatuni, line 64: Probably this can be const?

Ralith: It'll get optimized regardless, I wouldn't overthink it.

cli/src/progress.rs
martinvonz: Done. I also moved this line up a bit to be closer to the `if now < self.next_print {` line.

cli/src/progress.rs
martinvonz: I left it as is (except that I inlined it again now that it's only used in one place).

Issue #3912, opened 2024-06-18
FR: Add `--ignore-working-copy` hint to large file snapshot error 


fynnsu:
**Is your feature request related to a problem? Please describe.**
I recently ran into this issue when I temporarily added a large (>500 GB) untracked file to my repo. I wanted to run some tests, make a change to an upstream commit, switch back to my commit and run the tests again. Instead I found that I couldn't run any `jj` commands (even stuff like `jj log`) because it was trying to snapshot the large file.

My workaround was to add the file to `.gitignore` switch to the upstream commit, realize that I also need to change the upstream `.gitignore` just to switch back.

**Describe the solution you'd like**
While this doesn't fully solve my problem (since I can't record the change to the upstream commit), it would've been helpful to have `--ignore-working-copy` hinted as an option to allow me to run `jj log` and switch between commits. The current experience is that `jj` stops working entirely and its suggestions (increasing max file size to >500 GB or adding to `.gitignore`) aren't helpful (in my case).

**Additional context**
```
Error: Failed to snapshot the working copy
The file 'MY_LARGE_FILE' is too large to be snapshotted: it is 660.0MiB; the maximum size allowed is ~1.0MiB.
Hint: This is to prevent large files from being added on accident. You can fix this error by:
  - Adding the file to `.gitignore`
  - Run `jj config set --repo snapshot.max-new-file-size 692064256`
    This will increase the maximum file size allowed for new files, in this repository only.
  - Run `jj --config-toml 'snapshot.max-new-file-size=692064256' st`
    This will increase the maximum file size allowed for new files, for this command only.
```

**I think it would be helpful to add something like**
```
- Run the command with `--ignore-working-copy`
  This will ignore changes to the working copy that have not been previously snapshotted.
```

Then approaching this again, I would likely use `--ignore-working-copy` to list the commits, choose a commit high enough upstream, switch to the commit (again using `--ignore-working-copy`) and add the file to the `.gitignore` upstream.

PhilipMetzger:
Ideally we shouldn't even fail in `jj st`, see https://github.com/martinvonz/jj/issues/3616#issuecomment-2096748218 but this seems to be nice improvement until that is done.

cole-h:
Now that https://github.com/jj-vcs/jj/pull/5069 has merged (which changes the behavior to warn and continue instead of error when encountering large files), is this still desirable?

thoughtpolice:
[...]

Issue #3616, opened 2024-05-03
FR: allow restoring repo without snapshot


0xdeafbeef:
**Is your feature request related to a problem? Please describe.**
Your app accidentally created a billion files, and you want to restore the repository to a clean state. 
If one of the files is larger than `snapshot.max-new-file-size`, you're in trouble.
You can either snapshot it to delete or do it by hand, both of which are cumbersome. 

```sh
 cd $(mktemp -d)
 jj git init .
 echo a > a
 jj describe -m 'init'
 jj new
 dd if=/dev/urandom of=rand_10m bs=1M count=10
 jj restore .
#Error: Failed to snapshot the working copy
#The file '/tmp/tmp.CQHWzRWxd6/rand_10m' is too large to be snapshotted: it is 10.0MiB; the maximum size allowed #is ~1.0MiB.
#Hint: This is to prevent large files from being added on accident. You can fix this error by:
#  - Adding the file to `.gitignore`
#  - Run `jj config set --repo snapshot.max-new-file-size 10485760`
#    This will increase the maximum file size allowed for new files, in this repository only.
#  - Run `jj --config-toml 'snapshot.max-new-file-size=10485760' st`
#    This will increase the maximum file size allowed for new files, for this command only.
```

**Describe the solution you'd like**
`jj restore --force`, which will skip snapshotting of repo. Maybe we can introduce new `jj reset` subcommand.

noahmayr:
Have you tried `jj restore . --ignore-working-copy`?

yuja:
iirc, there are some discussion about demoting the max-new-file-size error to warning. It will allow `jj restore` in that situation. I'm not sure if that's a good idea because doing that can be a footgun.

0xdeafbeef:
> iirc, there are some discussion about demoting the max-new-file-size error to warning. It will allow `jj restore` in that situation. I'm not sure if that's a good idea because doing that can be a footgun.

Perhaps it would be feasible to actually remove newly created files when `--ignore-working-copy` is set?

yuja:
> Perhaps it would be feasible to actually remove newly created files when `--ignore-working-copy` is set?

`--ignore-working-copy` is the option to not touch working-copy files. It seems scarier than allowing `jj restore` in dirty working copy.

noahmayr:
maybe `jj restore --untracked` which would get rid of all files that are not tracked or ignored?

yuja:
[...]

PR #5069, opened 2024-12-10
working_copy: on snapshot, warn new large files and continue

yuja:
#3616, #3912

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -50,6 +50,7 @@ use clap::FromArgMatches;
 use clap_complete::ArgValueCandidates;
 use indexmap::IndexMap;
 use indexmap::IndexSet;
+use indoc::writedoc;
 use itertools::Itertools;
 use jj_lib::backend::BackendResult;
 use jj_lib::backend::ChangeId;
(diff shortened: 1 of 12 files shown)


cli/src/cli_util.rs
```diff
@@ -2555,6 +2560,68 @@ pub fn print_conflicted_paths(
...
+        for (path, reason) in &stats.untracked_paths {
+            let ui_path = path_converter.format_file_path(path);
+            let message = match reason {
+                UntrackedReason::FileTooLarge { size, max_size } => {
+                    // if the size difference is < 1KiB, then show exact bytes.
+                    // otherwise, show in human-readable form; this avoids weird cases
+                    // where a file is 400 bytes too large but the error says something
+                    // like '1.0MiB, maximum size allowed is ~1.0MiB'
+                    let size_diff = size - max_size;
+                    if size_diff <= 1024 {
```
martinvonz, line 2583: Could we compare the humanized form instead? That seems more direct (and it better handles unlikely cases like max_size being 1KiB or 1GiB)

martinvonz (approved):

cli/src/cli_util.rs
yuja: Do you mean something like `HumanByteSize(size).to_string() == HumanByteSize(max_size).to_string()`? It will work, but I don't know if user want to see size_diff for GiB case.

martinvonz: Yes, that's what I mean. We could could humanize the diff too, perhaps? So something like "42 KiB too large; the maximum size allowed is 1073741824 bytes (1GiB)".

yuja: I simply removed the condition to print diff + exact sizes. New output is something like
```
Warning: Refused to snapshot some files:
  file.img: 100.0MiB (104857600 bytes); the maximum size allowed is 1.0MiB (1048576 bytes)
```

PTAL.

martinvonz: LGTM, thanks

Issue #2483, opened 2023-10-31
FR: fully automate the release process


thoughtpolice:
This is mostly for me, because I brought it up on Discord a while ago. We should get the release process to be fully automated, not only so that Martin doesn't have to do it, but so that the whole lifecycle is more secure and foolproof.

I think the idea was just to do a release every 4 weeks, on the 1st of every month. This is probably OK for now and we could adjust it in the future depending on how users feel.

There actually isn't that much to do in the grand scheme of things. But breaking it down, we at least need to:

- Tweak [CHANGELOG.md](./CHANGELOG.md) automatically
  - We may, frankly, want to just change this process entirely rather than doing egregious `sed` hacks or something. Or at least do it right (e.g. an actual markdown parser that can insert changes in the right places.) 
- Bump `workspace.package.version` in [Cargo.toml](./Cargo.toml)
- Update `Cargo.lock` with those version changes.
- Commit those changes to `main` if they build correctly, with a tag
- Have a workflow that creates a release directly from the tag

There are a lot of finnicky little details to sort out with this, but it can all reasonably be achieved with some elbow grease. In the long run it will drastically simplify things, from setting deprecation deadlines to ensuring continuity.

I'll probably experiment in a separate repository with doing the release of a small rust application. There are in fact some workflows to do this automatically but depending on those things might (or might not!) have their own issues.

ilyagr:
I can think of two tangentially related issues (neither is a prerequisite for this to work):

- https://github.com/martinvonz/jj/pull/2258 is blocked (at least) on increasing the release workload. If it can be automated, I still think it's a good idea.
- The demos, as discussed in https://github.com/martinvonz/jj/pull/2463#discussion_r1375509939, are not currently an issue for automating the release, but would be if we wanted to version them to the "correct" version of `jj`.

PR #5076, opened 2024-12-11
github: scope/narrow permissions, prevent template injection via GHA, enable zizmor workflow

thoughtpolice:
This adds a series of changes to the workflow files suggested by the [`zizmor`](https://woodruffw.github.io/zizmor/) tool, an automated security scanner for GitHub Actions Workflows. At the very end, it also adds a new workflow which will automatically run `zizmor` on all changes going forward.

Context: there was [a recent compromise of a very popular pypi package named `ultralytics`](https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection) that occurred due to template injection through a pushed tag. GHA has also had a long string of security vulnerabilities

---

Almost all of the suggested changes were low-priority, related to:

1) Overly broad permissions (we granted `read-all` when we either didn't need it, or only needed it for extremely few steps), and
2) Persisting GitHub credentials into `$HOME/.git` storage when running checkout &mdash; which could be exploited by steps that read and exfiltrated them.

In our case, neither of these features were necessary, so removing them is easy. However...

It also found **real** template injection vulnerabilities in the `release.yml` workflow, which in theory could be exploited by contributors who pushed tags (I am not sure if anyone except Martin is capable of pushing tags now, but it's easier to assume the answer is "yes").

I don't believe these could be exploited in a meaningful way *today* by any (compromised) contributor because the only thing you could exfiltrate is a narrowly scoped `$GITHUB_TOKEN` credential. However, I filed #2483 last year to automate the release workflow, which would necessarily expose a https://crates.io API token to the release workflow &mdash; which could have then been potentially exfiltrated via this vulnerability, resulting in absolute catastrophe.

# Checklist

If applicable:

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
 
-permissions: read-all
+permissions: {}
 
 jobs:
   binaries:
@@ -42,6 +42,8 @@ jobs:
     name: Build binary artifacts
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          persist-credentials: false
       - name: Install packages (Ubuntu)
         if: startsWith(matrix.os, 'ubuntu')
         run: |
@@ -61,7 +63,7 @@ jobs:
         run: |
           outdir="target/${{ matrix.target }}/release"
           BIN=$outdir/jj
-          [[ "${{ matrix.os }}" == "windows-latest" ]] && BIN+=".exe" 
+          [[ "${{ matrix.os }}" == "windows-latest" ]] && BIN+=".exe"
 
           mkdir -p target/out
           cp $BIN target/out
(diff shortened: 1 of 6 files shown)


thoughtpolice:
I haven't tested the release workflow in my thoughtpolice/jj repository yet, so this is currently in draft until I can make sure nothing broke due to a missed permission.

ilyagr:
This might be interesting to @neongreen or relevant to <https://github.com/jj-vcs/jj/pull/5117>.

neongreen:
Lgtm (assuming it works). I will rebase my docs pipeline when this is merged

ilyagr (approved):
Thanks Austin, and thanks Emily for the help with reviewing!

I don't follow all of these issues, but I trust your research, and we can revert this if it causes problems (e.g. if `zizmor` turns out to be hard to live with).

thoughtpolice:
Yeah, I think we can just get this in and fix any minor issues that come up in post. Thanks!

ilyagr:
This PR: https://www.youtube.com/watch?v=2Pox3loU9lQ

PR #5079, opened 2024-12-12
templater: add Email template type, deprecate `Signature.username()`

bnjmnt4n:
The `Signature.email()` method is also updated to return the new Email type. The `Signature.username()` method is deprecated for `Signature.email().local()`.

This is an alternative to #4687.

# Checklist

If applicable:


cli/src/template_builder.rs
```diff
@@ -217,6 +222,9 @@ impl<'a> IntoTemplateProperty<'a> for CoreTemplatePropertyKind<'a> {
             CoreTemplatePropertyKind::StringList(property) => {
                 Some(Box::new(property.map(|l| !l.is_empty())))
             }
+            CoreTemplatePropertyKind::Email(property) => Some(Box::new(
+                property.map(|e| !e.local.is_empty() || e.domain.is_some()),
+            )),
```
yuja, line 227: nit: sort in definition order?

cli/src/template_builder.rs
```diff
@@ -820,14 +840,19 @@ fn builtin_signature_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
         "email",
         |_language, _diagnostics, _build_ctx, self_property, function| {
             function.expect_no_arguments()?;
-            let out_property = self_property.map(|signature| signature.email);
-            Ok(L::wrap_string(out_property))
+            let out_property = self_property.map(|signature| signature.email.clone().into());
```
yuja, line 843: nit: `.clone()` isn't needed

cli/src/templater.rs
```diff
@@ -75,13 +76,41 @@ impl Template for Signature {
...
         }
         Ok(())
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Email {
+    pub local: String,
+    pub domain: Option<String>,
+}
```
yuja, line 91: nit: might be simpler to implement as a new type `Email(pub String)` (because the type implements implicit cast to boolean based on string)

docs/templates.md
```diff
@@ -210,10 +210,16 @@ The following methods are defined.
 The following methods are defined.
 
 * `.name() -> String`
-* `.email() -> String`
-* `.username() -> String`
+* `.email() -> Email`
 * `.timestamp() -> Timestamp`
 
+### Email type
```
yuja, line 216: nit: types are sorted alphabetically in doc

yuja (approved):
Thanks.

cli/src/templater.rs
bnjmnt4n: Does this mean performing the splitting to local/domain on demand in the `Template::format` function and the `Email.local()` and `Email.domain()` function?

yuja: Yes. Since `split_email()` never fails, `Email` doesn't have to be parsed at the construction.

bnjmnt4n:
A potential issue of this change is that you can no longer simply compare `author.email() == "x@y.com"`, unless we implement that as well. Any thoughts on whether we should do that?

cli/src/template_builder.rs
```diff
@@ -280,18 +289,28 @@ impl<'a> IntoTemplateProperty<'a> for CoreTemplatePropertyKind<'a> {
...
             }
             (CoreTemplatePropertyKind::Integer(lhs), CoreTemplatePropertyKind::Integer(rhs)) => {
                 Some(Box::new((lhs, rhs).map(|(l, r)| l == r)))
             }
+            (CoreTemplatePropertyKind::Email(lhs), CoreTemplatePropertyKind::Email(rhs)) => {
+                Some(Box::new((lhs, rhs).map(|(l, r)| l == r)))
+            }
+            (CoreTemplatePropertyKind::Email(lhs), CoreTemplatePropertyKind::String(rhs)) => {
+                Some(Box::new((lhs, rhs).map(|(l, r)| l.0 == r)))
+            }
```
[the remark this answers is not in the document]
bnjmnt4n: Done!

cli/src/template_builder.rs
```diff
@@ -2141,6 +2221,10 @@ mod tests {
         insta::assert_snapshot!(env.render_ok(r#"'a' == 'b'"#), @"false");
         insta::assert_snapshot!(env.render_ok(r#"'a' != 'a'"#), @"false");
         insta::assert_snapshot!(env.render_ok(r#"'a' != 'b'"#), @"true");
+        insta::assert_snapshot!(env.render_ok(r#"email1 == email1"#), @"true");
+        insta::assert_snapshot!(env.render_ok(r#"email1 == email2"#), @"false");
+        insta::assert_snapshot!(env.render_ok(r#"email1 == 'local-1@domain'"#), @"true");
+        insta::assert_snapshot!(env.render_ok(r#"email1 != 'local-2@domain'"#), @"true");
```
yuja, line 2227: nit: add `str <op> email`

yuja (approved):

PR #5082, opened 2024-12-12
cli: add --config-file=PATH argument

yuja:
# Checklist

If applicable:


yuja:
> > If we want to add more source variants (such as fd number), it might be better
> > to add --config-from=:<path|fd|..>.
> 
> This also seems like a really good idea.

I'm not yet sure if we'll add some of these, so I'll take `--config-file=PATH` syntax for now.

PR #5090, opened 2024-12-12
Back out 89d57ffb: "build: add `rust-toolchain.toml`"

thoughtpolice:
This backs out commit 89d57ffb29577c99f0f5187b76bff369f19c5886.

This is causing a CI failure because we can't build musl binaries, presumably because the rust-toolchain file overriding the chosen musl toolchain for some reason. Backout until we can reapply a proper fix.

# Checklist

If applicable:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ members = ["cli", "lib", "lib/gen-protos", "lib/proc-macros", "lib/testutils"]
 [workspace.package]
 version = "0.24.0"
 license = "Apache-2.0"
-rust-version = "1.76" # NOTE: remember to update CI, rust-toolchain.toml, contributing.md, changelog.md, install-and-setup.md, and flake.nix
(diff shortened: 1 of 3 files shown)


arxanas (approved):
Thanks for fixing this.

- Do you understand why the job is failing? It's not clear to me why setting the Rust toolchain would affect this; surely we can build a release on our MSRV as well as latest `stable`?
- Do you know if there's a way to test/trigger the build-releases job as part of a PR to validate this behavior in the future?

thoughtpolice:
Yeah, we can build binaries on PRs to catch this in the future, I might submit a follow up change for that even if they mostly won't be used. As for why it's failing, no, I don't yet have a fully clear idea, but I also haven't isolated it to whether it's just a problem with dtolnay/rust-toolchain, or the existince of the .toml file is overriding something. I might hopefully have some time this weekend to do so...

PR #5096, opened 2024-12-13
Docs: swap order of `@` and `main` in `jj new` doc comment

david-crespo:
I followed the recommendation in the `jj new` doc to use `jj new main @` to make a merge commit and ended up with a merge commit that GitHub did not like. The PR diff included both the relevant changes from that branch plus everything I merged in. @papertigers pointed out that swapping the two args produces a merge commit GitHub understands better. Happy to add a line explaining that the order matters, but it might be too much detail for this spot. The linked doc https://martinvonz.github.io/jj/latest/working-copy/ also does not explain this.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/new.rs
+++ b/cli/src/commands/new.rs
@@ -44,8 +44,8 @@ use crate::ui::Ui;
 /// the new commit. This can be avoided with `--no-edit`.
 ///
 /// Note that you can create a merge commit by specifying multiple revisions as
-/// argument. For example, `jj new main @` will create a new commit with the
-/// `main` bookmark and the working copy as parents.
+/// argument. For example, `jj new @ main` will create a new commit with the
+/// working copy and the `main` bookmark as parents.
 ///
 /// For more information, see
 /// https://martinvonz.github.io/jj/latest/working-copy/.
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -1392,7 +1392,7 @@ Create a new, empty change and (by default) edit it in the working copy
 
 By default, `jj` will edit the new change, making the working copy represent the new commit. This can be avoided with `--no-edit`.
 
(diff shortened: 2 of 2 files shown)


cli/src/commands/new.rs
PhilipMetzger, line 1: The _commit message_ not the PR should be in the topic format, so this should be `new: swap order of @ and main in jj new doc comment`

PhilipMetzger (approved):
LG, but can you comply with our commit style here and maybe add the PR text as a motivation too.

cli/src/commands/new.rs
```diff
@@ -44,7 +44,7 @@ use crate::ui::Ui;
 /// the new commit. This can be avoided with `--no-edit`.
 ///
 /// Note that you can create a merge commit by specifying multiple revisions as
-/// argument. For example, `jj new main @` will create a new commit with the
+/// argument. For example, `jj new @ main` will create a new commit with the
 /// `main` bookmark and the working copy as parents.
```
cenviity, line 48: nit: Swap the order of ‘the `main` bookmark’ and ‘the working copy’ as well for consistency.

cli/tests/cli-reference@.md.snap
```diff
@@ -1392,7 +1392,7 @@ Create a new, empty change and (by default) edit it in the working copy
 
 By default, `jj` will edit the new change, making the working copy represent the new commit. This can be avoided with `--no-edit`.
 
-Note that you can create a merge commit by specifying multiple revisions as argument. For example, `jj new main @` will create a new commit with the `main` bookmark and the working copy as parents.
+Note that you can create a merge commit by specifying multiple revisions as argument. For example, `jj new @ main` will create a new commit with the `main` bookmark and the working copy as parents.
```
cenviity, line 1395: nit: Same as above. Swap the order of ‘the `main` bookmark’ and ‘the working copy’ as well for consistency.

cenviity (changes requested):

PhilipMetzger:
and feel free to resolve all comments that you've adressed

cenviity (approved):

PR #3867, opened 2024-06-12
cli: deprecate `--config-toml` in favor of `--config`

martinvonz:
# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #5102, opened 2024-12-15
cli: add simpler --config=NAME=VALUE argument

yuja:
#3867

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3103,8 +3103,13 @@ pub struct EarlyArgs {
     // Option<bool>.
     pub no_pager: Option<bool>,
     /// Additional configuration options (can be repeated)
-    //  TODO: Introduce a `--config` option with simpler syntax for simple
-    //  cases, designed so that `--config ui.color=auto` works
+    ///
+    /// The name should be specified as TOML dotted keys. The value should be
+    /// specified as a TOML expression. If string value doesn't contain any TOML
+    /// constructs (such as array notation), quotes can be omitted.
+    #[arg(long, value_name = "NAME=VALUE", global = true)]
+    pub config: Vec<String>,
+    /// Additional configuration options (can be repeated)
     #[arg(long, value_name = "TOML", global = true)]
     pub config_toml: Vec<String>,
     /// Additional configuration files (can be repeated)
@@ -3117,10 +3122,12 @@ impl EarlyArgs {
         merge_args_with(
             matches,
             &[
+                ("config", &self.config),
                 ("config_toml", &self.config_toml),
                 ("config_file", &self.config_file),
             ],
             |id, value| match id {
+                "config" => (ConfigArgKind::Item, value.as_ref()),
                 "config_toml" => (ConfigArgKind::Toml, value.as_ref()),
                 "config_file" => (ConfigArgKind::File, value.as_ref()),
                 _ => unreachable!("unexpected id {id:?}"),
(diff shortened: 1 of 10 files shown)


cli/src/commands/config/set.rs
```diff
@@ -26,16 +27,21 @@ use crate::cli_util::WorkspaceCommandHelper;
...
 #[derive(clap::Args, Clone, Debug)]
 pub struct ConfigSetArgs {
     #[arg(required = true, add = ArgValueCandidates::new(complete::leaf_config_keys))]
     name: ConfigNamePathBuf,
-    #[arg(required = true)]
-    value: String,
+    /// New value to set
+    ///
+    /// The value should be specified in either TOML syntax or string. If a
+    /// string value contained TOML constructs (such as array notation), it must
```
PhilipMetzger, line 41: Nit: Should be "The value should be specified in either TOML syntax or as a string. If string value contains TOML constructs (such as array notation), it must be quoted".

cli/src/cli_util.rs
```diff
@@ -3099,8 +3099,13 @@ pub struct EarlyArgs {
     // Option<bool>.
     pub no_pager: Option<bool>,
     /// Additional configuration options (can be repeated)
-    //  TODO: Introduce a `--config` option with simpler syntax for simple
-    //  cases, designed so that `--config ui.color=auto` works
+    ///
+    /// The name should be specified as TOML dotted keys. The value should be in
+    /// either TOML syntax or string. If a string value contained TOML
+    /// constructs (such as array notation), it must be quoted.
```
PhilipMetzger, line 3105: Ditto for the correction.

cli/src/commands/config/set.rs
PhilipMetzger, line 1: Nit: This may break some user scripts, so it should be soft deprecated and get a CHANGELOG entry.

PhilipMetzger (approved):
LG, just a minor correction and a actual nit.

cli/src/commands/config/set.rs
yuja: Thanks. Updated the doc.

cli/src/commands/config/set.rs
yuja: Hmm, I assumed scripts don't usually set config values permanently. I've added changelog entry, but no code change (to suppress parse error.)

cli/src/cli_util.rs
```diff
@@ -3099,8 +3099,13 @@ pub struct EarlyArgs {
     // Option<bool>.
     pub no_pager: Option<bool>,
     /// Additional configuration options (can be repeated)
-    //  TODO: Introduce a `--config` option with simpler syntax for simple
-    //  cases, designed so that `--config ui.color=auto` works
+    ///
+    /// The name should be specified as TOML dotted keys. The value should be
+    /// specified in either TOML syntax or as a string. If string value contains
+    /// TOML constructs (such as array notation), it must be quoted.
```
martinvonz, line 3105: I think it would be good explicitly say that TOML syntax is attempted first. That means it's safe for scripts to use this option with valid TOML and be sure that it won't be interpreted as a string.

martinvonz (approved):
> I'll probably replace all uses of --config-toml and mark --config-toml as deprecated.

SGTM

cli/src/commands/config/set.rs
PhilipMetzger: I err on the side of caution, but let's just land it and see what happens.

cli/src/cli_util.rs
yuja: Does this work?
```
    /// The value should be specified as a TOML expression. If string value
    /// doesn't contain any TOML constructs (such as array notation), quotes can
    /// be omitted.
```

martinvonz: Yes, that makes sense. Thanks.

PR #5110, opened 2024-12-16
Useful Revsets in docs/github.md: fix ::@ == ancestors and exclude bookmark ancestry

tim-janik:
First, describe that ::@ lists ancestors.
Second, to make the example useful, remote bookmarks need to be excluded with their ancestry, otherwise we would list a bunch of immutable commits without their bookmarks.

# Checklist

If applicable:
- [ x] I have updated the documentation (README.md, docs/, demos/)

### DIFF
--- a/docs/github.md
+++ b/docs/github.md
@@ -222,10 +222,10 @@ Log all remote bookmarks that you authored or committed to:
 $ jj log -r 'remote_bookmarks() & (mine() | committer(your@email.com))'
 ```
 
-Log all descendants of the current working copy that aren't on any remote:
+Log all ancestors of the current working copy that aren't on any remote:
 
 ```shell
-$ jj log -r '::@ & ~remote_bookmarks()'
+$ jj log -r 'remote_bookmarks()..@'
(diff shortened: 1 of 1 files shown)


docs/github.md
```diff
@@ -222,10 +222,10 @@ Log all remote bookmarks that you authored or committed to:
 $ jj log -r 'remote_bookmarks() & (mine() | committer(your@email.com))'
 ```
 
-Log all descendants of the current working copy that aren't on any remote:
+Log all ancestors of the current working copy that aren't on any remote:
 
 ```shell
-$ jj log -r '::@ & ~remote_bookmarks()'
+$ jj log -r '::@ & ~::remote_bookmarks()'
```
martinvonz, line 228: This can also be written using the difference operator instead of the negation operator (same symbol but infix instead of prefix): `::@ ~ ::remote_bookmarks()`. That can also be written as `remote_bookmarks()..@`. Let's switch to that?

martinvonz (approved):

PR #5111, opened 2024-12-16
resolve: try to resolve all conflicted files in fileset

scott2000:
If many files are conflicted, it would be nice to be able to resolve all conflicts at once without having to run `jj resolve` multiple times. This is especially nice for merge tools which try to automatically resolve conflicts without user input, but it can also be used for regular merge editors like VS Code.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/resolve.rs
+++ b/cli/src/commands/resolve.rs
@@ -28,10 +28,13 @@ use crate::command_error::CommandError;
 use crate::complete;
 use crate::ui::Ui;
 
-/// Resolve a conflicted file with an external merge tool
+/// Resolve conflicted files with an external merge tool
 ///
 /// Only conflicts that can be resolved with a 3-way merge are supported. See
-/// docs for merge tool configuration instructions.
+/// docs for merge tool configuration instructions. External merge tools will be
+/// invoked for each conflicted file one-by-one until all conflicts are
+/// resolved. To stop resolving conflicts, exit the merge tool without making
+/// any changes.
 ///
 /// Note that conflicts can also be resolved without using this command. You may
 /// edit the conflict markers in the conflicted file directly with a text
@@ -52,7 +55,7 @@ pub(crate) struct ResolveArgs {
         add = ArgValueCandidates::new(complete::mutable_revisions),
     )]
     revision: RevisionArg,
-    /// Instead of resolving one conflict, list all the conflicts
+    /// Instead of resolving conflicts, list all the conflicts
     // TODO: Also have a `--summary` option. `--list` currently acts like
     // `diff --summary`, but should be more verbose.
     #[arg(long, short)]
@@ -60,10 +63,8 @@ pub(crate) struct ResolveArgs {
     /// Specify 3-way merge tool to be used
     #[arg(long, conflicts_with = "list", value_name = "NAME")]
     tool: Option<String>,
-    /// Restrict to these paths when searching for a conflict to resolve. We
-    /// will attempt to resolve the first conflict we can find. You can use
-    /// the `--list` argument to find paths to use here.
-    // TODO: Find the conflict we can resolve even if it's not the first one.
+    /// Only resolve conflicts in these paths. You can use the `--list` argument
+    /// to find paths to use here.
     #[arg(
         value_name = "FILESETS",
         value_hint = clap::ValueHint::AnyPath,
@@ -103,16 +104,15 @@ pub(crate) fn cmd_resolve(
         );
     };
 
-    let (repo_path, _) = conflicts.first().unwrap();
+    let repo_paths = conflicts
+        .iter()
+        .map(|(path, _)| path.as_ref())
+        .collect_vec();
     workspace_command.check_rewritable([commit.id()])?;
     let merge_editor = workspace_command.merge_editor(ui, args.tool.as_deref())?;
-    writeln!(
-        ui.status(),
-        "Resolving conflicts in: {}",
-        workspace_command.format_file_path(repo_path)
-    )?;
     let mut tx = workspace_command.start_transaction();
-    let new_tree_id = merge_editor.edit_file(&tree, repo_path)?;
+    let (new_tree_id, partial_resolution_error) =
+        merge_editor.edit_files(ui, &tree, &repo_paths)?;
     let new_commit = tx
         .repo_mut()
         .rewrite_commit(&commit)
@@ -138,6 +138,10 @@ pub(crate) fn cmd_resolve(
                 print_conflicted_paths(new_conflicts, formatter.as_mut(), &workspace_command)?;
             }
         }
+    }
+
+    if let Some(err) = partial_resolution_error {
+        return Err(err.into());
     }
     Ok(())
 }
--- a/cli/src/merge_tools/external.rs
+++ b/cli/src/merge_tools/external.rs
@@ -8,7 +8,6 @@ use std::sync::Arc;
 
 use bstr::BString;
 use itertools::Itertools;
-use jj_lib::backend::FileId;
 use jj_lib::backend::MergedTreeId;
 use jj_lib::backend::TreeValue;
 use jj_lib::conflicts;
@@ -19,10 +18,8 @@ use jj_lib::conflicts::MIN_CONFLICT_MARKER_LEN;
 use jj_lib::gitignore::GitIgnoreFile;
 use jj_lib::matchers::Matcher;
 use jj_lib::merge::Merge;
-use jj_lib::merge::MergedTreeValue;
 use jj_lib::merged_tree::MergedTree;
 use jj_lib::merged_tree::MergedTreeBuilder;
-use jj_lib::repo_path::RepoPath;
 use jj_lib::working_copy::CheckoutOptions;
 use pollster::FutureExt;
 use thiserror::Error;
@@ -35,6 +32,7 @@ use super::diff_working_copies::DiffSide;
 use super::ConflictResolveError;
 use super::DiffEditError;
 use super::DiffGenerateError;
+use super::MergeToolFile;
 use crate::config::find_all_variables;
 use crate::config::interpolate_variables;
 use crate::config::CommandNameAndArgs;
@@ -175,13 +173,17 @@ pub enum ExternalToolError {
 
 pub fn run_mergetool_external(
     editor: &ExternalMergeTool,
-    file_merge: Merge<Option<FileId>>,
-    content: Merge<BString>,
-    repo_path: &RepoPath,
-    conflict: MergedTreeValue,
     tree: &MergedTree,
+    merge_tool_file: &MergeToolFile,
     default_conflict_marker_style: ConflictMarkerStyle,
 ) -> Result<MergedTreeId, ConflictResolveError> {
+    let MergeToolFile {
+        repo_path,
+        conflict,
+        file_merge,
+        content,
+    } = merge_tool_file;
+
     let conflict_marker_style = editor
         .conflict_marker_style
         .unwrap_or(default_conflict_marker_style);
@@ -193,13 +195,13 @@ pub fn run_mergetool_external(
     // MIN_CONFLICT_MARKER_LEN since the merge tool can't know about our rules for
     // conflict marker length.
     let conflict_marker_len = if editor.merge_tool_edits_conflict_markers || uses_marker_length {
-        choose_materialized_conflict_marker_len(&content)
+        choose_materialized_conflict_marker_len(content)
     } else {
         MIN_CONFLICT_MARKER_LEN
     };
     let initial_output_content = if editor.merge_tool_edits_conflict_markers {
         materialize_merge_result_to_bytes_with_marker_len(
-            &content,
+            content,
             conflict_marker_style,
             conflict_marker_len,
         )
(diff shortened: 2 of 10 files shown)


arxanas:
I recently learned that some commands support [filesets](https://martinvonz.github.io/jj/latest/filesets/). Is `jj resolve` one of those? Could we support this use-case by extending the fileset language, rather than adding a flag specifically to `jj resolve`? (Does `jj resolve 'all()'` work today?)

martinvonz:
We actually already support filsets in this command. It's just that the command would always stop after the first conflict before this PR (so even `jj resolve 'all()'` would just pick the first conflict it runs into).

scott2000:
> I recently learned that some commands support [filesets](https://martinvonz.github.io/jj/latest/filesets/). Is `jj resolve` one of those? Could we support this use-case by extending the fileset language, rather than adding a flag specifically to `jj resolve`? (Does `jj resolve 'all()'` work today?)

Yes, it already does support filesets, and the default is `all()` currently already. The issue is that it always picks a single file from the fileset, rather than using the whole fileset. The new `--all` flag just tells it to use the whole fileset, rather than picking a single file from the fileset. We could maybe make it so that if the user explicitly passes a fileset, it uses the whole fileset? I'm not sure if anyone relies on the existing behavior though.

arxanas:
> We actually already support filsets in this command. It's just that the command would always stop after the first conflict before this PR (so even jj resolve 'all()' would just pick the first conflict it runs into).

Is this because existing merge editor work on a single file at a time, so we have no choice but to resolve conflicts individually? (Is there a warning in the case that subsequent conflicts are ignored?) 

For uniformity, should we have an `all:` modifier (like for revsets) for filesets?
- Assuming that it's a limitation of the merge editor interface, and you don't inadvertently want to resolve multiple files at once, perhaps it makes sense to warn before trying to resolve conflicts for multiple files, in the same way that we warn for certain multiple-revision operations?
- I could be convinced either way, but by default, I would be inclined to keep the DSLs similar.

arxanas:
> If we want the default behavior of picking a single file to be expressable with filesets, then we'd probably want a `first(fileset)` function to pick a single file from a fileset. I'm not sure if that would make the implementation of filesets more complicated though.

Makes sense to me, although implementing that is its whole own matter.

> TBH, I'm not sure I've ever actually wanted the default behavior of only resolving a single, seemingly-random file though. Usually, I want to resolve every conflict, or I want to resolve a conflict in a specific file I specify. So maybe we could just change `jj resolve` to just resolve every file in the fileset and require the user to name a specific file if they only want to resolve a single file?

In both your workflow and my workflow, it sounds like resolving a single arbitrary file is probably not desirable:
- In my own workflow, I typically get scared and confused when I try to finish up and I wasn't expecting the tool to launch more editors.
  - It used to be particularly difficult to resolve in cases where I was using a CLI editor to resolve conflicts, so there was no way to just go back to my terminal and type <kbd>ctrl-c</kbd> to cancel the whole process.
  - These days, I know how to make Vim exit with an error code, but there may be other users might share my old workflow of trying to trying to quit and then press <kbd>ctrl-c</kbd> really fast before the caller can launch another editor 👀.
- Thererfore: I would prefer an error in the case that `jj resolve` would operate on multiple files, while you would prefer to just start resolving multiple files, but neither of us would want the current behavior.
- But, in fairness: I don't use `jj resolve` that much vs just opening the conflicted files in my editor and letting them get auto-resolved, so I'll defer to anyone who specifically uses `jj resolve`.

scott2000:
> These days, I know how to make Vim exit with an error code, but there may be other users might share my old workflow of trying to trying to quit and then press ctrl-c really fast before the caller can launch another editor 👀.

I think in practice, this shouldn't be an issue because we treat an unmodified or blank output file as an error, so it would stop launching more editors if someone closes the editor without making any changes. That being said, it does feel a bit strange to launch the editor multiple times without the user requesting it explicitly, so it might be nice to not launch multiple editors by default anyway.

At some point, I'm thinking that we probably will want to support merge tools which can merge whole directories at once instead of being invoked with each file individually.

martinvonz:
> So maybe we could just change `jj resolve` to just resolve every file in the fileset and require the user to name a specific file if they only want to resolve a single file?

I'm also for that.

martinvonz:
Oh, @ilyagr might also have thoughts. He wrote most of the current implementation.

cli/src/merge_tools/external.rs
```diff
@@ -286,8 +289,49 @@ pub fn run_mergetool_external(
...
+            }
+            Err(err) => {
+                // Since some conflicts were already successfully resolved, just print a warning
+                // and stop resolving conflicts
+                writeln!(
+                    ui.warning_default(),
+                    "Stopping due to error after resolving {i} conflicts"
+                )?;
+                print_error_sources(ui, Some(&err))?;
+                break;
```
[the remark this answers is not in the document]
scott2000: Yeah, I was considering that as well. I mainly chose to make it a warning instead of an error because I didn't want any previous successful resolutions to be discarded (so returning an error immediately isn't an option), and I thought it looked strange having the error message be at the end of the command after the transaction commits successfully instead of being before the transaction commits. Maybe I could make it so that it prints the error immediately, but doesn't actually exit with the error code until after the transaction commits?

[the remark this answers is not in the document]
martinvonz: > Maybe I could make it so that it prints the error immediately, but doesn't actually exit with the error code until after the transaction commits?

SGTM, thanks.

cli/src/merge_tools/mod.rs
```diff
@@ -324,44 +328,58 @@ impl MergeEditor {
     /// Starts a merge editor for the specified file.
```
martinvonz, line 328: Should be plural now. The function name too

cli/src/commands/resolve.rs
```diff
@@ -103,16 +105,32 @@ pub(crate) fn cmd_resolve(
...
-    let new_tree_id = merge_editor.edit_file(&tree, repo_path)?;
+    let merge_editor = tx
+        .base_workspace_helper()
+        .merge_editor(ui, args.tool.as_deref())?;
+    let mut suppressed_error = None;
+    let new_tree_id = match merge_editor.edit_files(ui, &tree, &repo_paths) {
+        Ok(new_tree_id) => new_tree_id,
+        Err(err) => {
+            // If resolution failed completely, we can return immediately since there's
+            // nothing to commit
```
martinvonz, line 122: nit: put this comment inside the `else` block since it applies only to that?

cli/src/commands/resolve.rs
```diff
@@ -103,16 +105,32 @@ pub(crate) fn cmd_resolve(
...
+            let ConflictResolveError::PartialResolution { resolved_tree, .. } = &err else {
+                return Err(err.into());
+            };
+
+            // Some files were resolved successfully, so we want to commit the transaction
+            // before returning the error
+            let resolved_tree = resolved_tree.clone();
+            suppressed_error = Some(err.into());
+            resolved_tree
+        }
```
martinvonz, line 132: Probably clearer if we handle `ConflictResolveError::PartialResolution` in the surrounding `match`?


```rust
    let new_tree_id = match merge_editor.edit_files(ui, &tree, &repo_paths) {
        Ok(new_tree_id) => new_tree_id,
        Err(ConflictResolveError::PartialResolution { resolved_tree, .. }) => {
            // Some files were resolved successfully, so we want to commit the transaction
            // before returning the error
            suppressed_error = Some(err.into());
            resolved_tree
        }
        Err(err) => {
            // If resolution failed completely, we can return immediately since there's
            // nothing to commit
            return Err(err.into());
        }
    };
```

Or maybe like this if you prefer:

```rust
    let new_tree_id = match merge_editor.edit_files(ui, &tree, &repo_paths) {
        Err(ConflictResolveError::PartialResolution { resolved_tree, .. }) => {
            // Some files were resolved successfully, so we want to commit the transaction
            // before returning the error
            suppressed_error = Some(err.into());
            resolved_tree
        }
        result => result?
    };
```

cli/src/merge_tools/mod.rs
```diff
@@ -324,51 +338,65 @@ impl MergeEditor {
...
+                };
+                let content =
+                    extract_as_single_hunk(&simplified_file_merge, tree.store(), repo_path)
+                        .block_on()?;
+                Ok(MergeToolFile {
+                    repo_path: repo_path.to_owned(),
+                    conflict,
+                    file_merge,
+                    content,
+                })
```
martinvonz, line 385: Worth extracting a function for this?

cli/src/commands/resolve.rs
scott2000: I actually originally had it like that, but I couldn't make it typecheck. The main issue is that I want to have `err` be owned so I can do `err.into()`, but I also want to pattern match `ConflictResolveError::PartialResolution { resolved_tree, .. }` so I can return the resolved tree, and it seems like `err @ ConflictResolveError::PartialResolution { resolved_tree, .. }` doesn't allow this. I think I would have to do something like this to make it typecheck:

```rust
    let new_tree_id = match merge_editor.edit_files(ui, &tree, &repo_paths) {
        Ok(new_tree_id) => new_tree_id,
        Err(ConflictResolveError::PartialResolution {
            source,
            resolved_tree,
            resolved_count,
        }) => {
            // Some files were resolved successfully, so we want to commit the transaction
            // before returning the error
            suppressed_error = Some(
                ConflictResolveError::PartialResolution {
                    source,
                    resolved_tree: resolved_tree.clone(),
                    resolved_count,
                }
                .into(),
            );
            resolved_tree
        }
        Err(err) => {
            // If resolution failed completely, we can return immediately since there's
            // nothing to commit
            return Err(err.into());
        }
    };
```

And I thought it was a bit too lengthy.

yuja: Perhaps, `edit_files()` should return some stats of resolution failure? something like `Result<(MergedTreeId, Vec<(RepoPathBuf, UnresolvedReason)>), ..>` or `Result<(MergedTreeId, Option<UnresolvedReason>), ..>`? It's probably better to distinguish expected resolution failure and unexpected backend/IO error.

scott2000: That seems a lot nicer, thanks!

martinvonz (approved):

cli/src/merge_tools/external.rs
```diff
@@ -308,10 +313,53 @@ pub fn run_mergetool_external(
...
+            Ok(()) => {}
+            Err(err) if i == 0 => {
+                // If the first resolution fails, just return the error normally
+                return Err(err);
+            }
+            Err(err) => {
+                // Some conflicts were already resolved, so we should return an error with the
+                // partially-resolved tree so that the caller can save the resolved files.
+                partial_resolution_error = Some(MergeToolPartialResolutionError {
+                    source: Box::new(err),
```
yuja, line 354: nit: Boxing might not be needed anymore since the type isn't recursive?

cli/src/merge_tools/external.rs
```diff
@@ -308,10 +313,53 @@ pub fn run_mergetool_external(
...
+                    resolved_count: i,
+                });
+                break;
```
yuja, line 357: Not for this PR, but I don't think exiting at the first unresolved conflict provides good UX.

If the merge tool can automatically resolve conflicts, user will probably want to process all resolvable files. If the tool is interactive, user might want to terminate in the middle of the process at arbitrary timing.

cli/src/merge_tools/mod.rs
```diff
@@ -97,8 +103,17 @@ pub enum ConflictResolveError {
...
+    #[error(transparent)]
     Backend(#[from] jj_lib::backend::BackendError),
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+}
+
+#[derive(Debug, Error)]
+#[error("Stopped due to error after resolving {resolved_count} conflicts")]
+pub struct MergeToolPartialResolutionError {
+    pub source: Box<ConflictResolveError>,
```
yuja, line 115: nit: it's probably better to extract error of "conflict resolution"-type (i.e. split `ExternalToolError` and `BuiltinToolError`), but that will require large refactoring.

(no action required)

cli/src/commands/resolve.rs
```diff
@@ -103,16 +104,17 @@ pub(crate) fn cmd_resolve(
...
-    writeln!(
-        ui.status(),
-        "Resolving conflicts in: {}",
-        workspace_command.format_file_path(repo_path)
-    )?;
     let mut tx = workspace_command.start_transaction();
-    let new_tree_id = merge_editor.edit_file(&tree, repo_path)?;
+    let merge_editor = tx
+        .base_workspace_helper()
+        .merge_editor(ui, args.tool.as_deref())?;
```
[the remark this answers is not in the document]
scott2000: I don't think this is possible because `MergeEditor<'_>` is now tied to the lifetime of `workspace_command` (since it contains a reference to the path converter), and `workspace_command.start_transaction()` takes a mutable reference.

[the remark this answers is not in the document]
yuja: Oh, can't we clone path converter? There editor types are to avoid borrowing issues.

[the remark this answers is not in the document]
scott2000: Oh, I didn't think of that. Let me change it.

PR #5112, opened 2024-12-16
feat: use `JJ_LOG` to specify log configuration

### LINKED ISSUE #4219: FR: Switch from RUST_LOG to JJ_LOG

mrnossiom:
Using the default `RUST_LOG` environment variable conflicts when working on unrelated Rust projects.

Supersedes #4350, which is stale.

Closes #4219 #4350

I am unsure whether this has to be listed as a breaking change in the CHANGELOG.

# Checklist

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -455,6 +455,16 @@ you can submit a PR based on the `gh-pages` bookmark of
  The `.rs` files generated from `.proto` files are included in the repository,
  and there is a GitHub CI check that will complain if they do not match.
 
+## Logging
+
+You can print internal jj logs using `JJ_LOG`. It acts like the `RUST_LOG`
+environment variable, frequent in Rust codebases, and accepts one or more
+[directives]. You can also use the `--debug` global option that sets
+`debug` log level for all targets by default. `JJ_LOG` is still respected when
+using `--debug`.
+
+[directives]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
+
 ## Profiling
 
 One easy-to-use sampling profiler
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -212,11 +212,14 @@ pub struct TracingSubscription {
 }
 
 impl TracingSubscription {
+    const ENV_VAR_NAME: &'static str = "JJ_LOG";
+
     /// Initializes tracing with the default configuration. This should be
     /// called as early as possible.
     pub fn init() -> Self {
         let filter = tracing_subscriber::EnvFilter::builder()
             .with_default_directive(tracing::metadata::LevelFilter::ERROR.into())
+            .with_env_var(Self::ENV_VAR_NAME)
             .from_env_lossy();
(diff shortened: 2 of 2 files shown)


PhilipMetzger (changes requested):
Nit: the topic of first commit should be either `cli:` or `cli_util: <rest-of-description`, as feat is not a known topic in the project, otherwise LG.

PhilipMetzger (approved):
LG

docs/contributing.md
```diff
@@ -455,6 +455,16 @@ you can submit a PR based on the `gh-pages` bookmark of
...
 
+## Logging
+
+You can print internal jj logs using `JJ_LOG`. It acts like the `RUST_LOG`
+environment variable, frequent in Rust codebases, and accepts one or more
+[directives]. You can also use the debug `--debug` global option that sets
+`debug` log level for all targets by default. `JJ_LOG` is still respected when
+using `--debug`.
+
+[directives]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
```
arxanas, line 466:
```suggestion
You can print internal jj logs using `JJ_LOG`. It acts like [the `RUST_LOG`
environment variable][rust-log-env-var] and accepts one or more
[directives]. You can also use the `--debug` global option that sets
`debug` log level for all targets by default. `JJ_LOG` is still respected when
using `--debug`.

[rust-log-env-var]: https://docs.rs/env_logger/latest/env_logger/#enabling-logging
[directives]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
```

suggestion (non-blocking): I think the `EnvLogger` page on `RUST_LOG` is helpful since it has several concrete examples, so I added a link there as well.
suggestion (non-blocking): I changed "debug `--debug`" to just "`--debug`".
comment: The commit message says that it documents the relation of `JJ_TRACE`, but I don't think it mentions `JJ_TRACE` here? We could add `JJ_TRACE` to the discussion or just remove that from the commit message.

arxanas (approved):
Seems good to me 👍. I had some suggestions, but they're not terribly important.

docs/contributing.md
mrnossiom: - `tracing-subscriber` documentation also has a few examples, adding the `env_logger` link feels kind of like a duplicate to me.
- debug repetition is just a typo, good catch
- `JJ_TRACE` is explained just below in the "Profiling" section of "Contributing". I originally took the issue as is in the commit message

istudyatuni:
Linked PR not closed

PR #5113, opened 2024-12-16
gitignore: add more detail to errors on invalid pattern

scott2000:
From Discord question: https://discord.com/channels/968932220549103686/1317492522712436826

# Checklist

If applicable:

### DIFF
--- a/lib/src/gitignore.rs
+++ b/lib/src/gitignore.rs
@@ -17,6 +17,7 @@
 use std::fs;
 use std::io;
 use std::iter;
+use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


martinvonz (approved):

lib/src/gitignore.rs
```diff
@@ -60,22 +64,31 @@ impl GitIgnoreFile {
...
             let line =
                 std::str::from_utf8(input_line).map_err(|err| GitIgnoreError::InvalidUtf8 {
-                    path: PathBuf::from(prefix),
+                    path: ignore_path.to_path_buf(),
                     line_num_for_display: i + 1,
                     line: String::from_utf8_lossy(input_line).to_string(),
                     source: err,
                 })?;
             // FIXME: do we need to provide the `from` argument? Is it for providing
             // diagnostics or correctness?
```
yuja, line 80: Maybe this comment can be removed? (or maybe we can pass it to `.add_line(Some(ignore_path), ..)`?

scott2000: I looked through the `ignore` crate, and it looks like it's neither for diagnostics nor correctness. It just stores it in the `Glob` so that when you call `GitIgnore::matched` or `GitIgnore::matched_path_or_any_parents`, it can tell you which file the matching glob came from. We ignore the `Glob` returned from these methods entirely, so I think it should be safe to remove the comment. I don't think we should pass it the path, since it causes an unnecessary clone of the string.

PR #5118, opened 2024-12-17
gitignore: remove FIXME comment about passing path

scott2000:
I looked through the code for the `ignore` crate, and this optional path is not used anywhere. The only reason to pass it would be to be able to get the path from the `Glob` when we call `Gitignore::matched` or `Gitignore::matched_path_or_any_parents`, but we ignore the returned `Glob` completely anyway. Passing the path would require an unnecessary clone of the path for each line in every .gitignore file, so it's better not to pass it since we don't need it.

https://github.com/martinvonz/jj/pull/5113#discussion_r1887769952

# Checklist

If applicable:

### DIFF
--- a/lib/src/gitignore.rs
+++ b/lib/src/gitignore.rs
@@ -76,8 +76,9 @@ impl GitIgnoreFile {
                     line: String::from_utf8_lossy(input_line).to_string(),
                     source: err,
                 })?;
-            // FIXME: do we need to provide the `from` argument? Is it for providing
-            // diagnostics or correctness?
+            // The `from` argument doesn't provide any diagnostics or correctness, so it is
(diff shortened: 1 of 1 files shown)


lib/src/gitignore.rs
```diff
@@ -76,8 +76,6 @@ impl GitIgnoreFile {
                     line: String::from_utf8_lossy(input_line).to_string(),
                     source: err,
                 })?;
-            // FIXME: do we need to provide the `from` argument? Is it for providing
-            // diagnostics or correctness?
             builder
```
yuja, line 79: nit: maybe we can leave a comment saying that "from argument doesn't provide any diagnostics or correctness"?

yuja (approved):
Thanks!

PR #5128, opened 2024-12-17
docs/cli: update martinvonz.github.io URLs to jj-vcs.github.io

martinvonz:
We just migrated to the jj-vcs GitHub org, so we should point to the new GitHub pages.

# Checklist

If applicable:

### DIFF
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -769,7 +769,7 @@ fn recreate_no_gc_refs(
         // TODO: might be better to switch to a dummy merge, where new no-gc ref
         // will always have a unique name. Doing that with the current
         // ref-per-head strategy would increase the number of the no-gc refs.
-        // https://github.com/martinvonz/jj/pull/2659#issuecomment-1837057782
+        // https://github.com/jj-vcs/jj/pull/2659#issuecomment-1837057782
         let loose_ref_path = git_repo.path().join(git_ref.name.to_path());
         if let Ok(metadata) = loose_ref_path.metadata() {
             let mtime = metadata.modified().expect("unsupported platform?");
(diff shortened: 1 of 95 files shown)


PhilipMetzger (approved):

arxanas (approved):
comment (non-blocking): You're no doubt aware, but the root URL <https://jj-vcs.github.io> produces a 404, which we should try to fix at some point.

comment (non-blocking): It's weird that all these tests are failing 👀

scott2000:
> comment (non-blocking): It's weird that all these tests are failing 👀

I think it's cause the README file added to conflict commits changed, so all conflicted commits have a different hash now.

lib/src/git_backend.rs
```diff
@@ -1430,7 +1430,7 @@ Git or other tools that are unfamiliar with jj.
 
 The .jjconflict-* directories represent the different inputs to the conflict.
 For details, see
-https://martinvonz.github.io/jj/prerelease/git-compatibility/#format-mapping-details
+https://jj-vcs.github.io/jj/prerelease/git-compatibility/#format-mapping-details
```
scott2000, line 1433: This file also contains a link to `https://github.com/martinvonz/jj` which should probably be updated.

martinvonz: Oh, good point! There are tons of those links. I'll fix all of them, including the ones in the changelog (it feels a bit weird to update the changelog, but I think it makes sense in this case).

martinvonz:
> comment (non-blocking): You're no doubt aware, but the root URL https://jj-vcs.github.io produces a 404, which we should try to fix at some point.

I was not, but it makes sense that it fails. That's the user-/org-level GitHub Pages URL, so we'll need to set up a separate little repo for that (like https://github.com/martinvonz/martinvonz.github.io)

lib/src/git_backend.rs
martinvonz: Done.

PR #5133, opened 2024-12-18
Use `FILESETS`/`REVSETS` for placeholders in command synopsis

martinvonz:
Thanks to @fowles and @yuja for the inspiration.

# Checklist

If applicable:


cenviity:
Should we also change the accompanying doc comments then, e.g. path(s) / file(s) → fileset(s), revision(s) → revset(s)?

fowles (approved):

cli/src/commands/file/show.rs
```diff
@@ -57,6 +58,7 @@ pub(crate) struct FileShowArgs {
         value_hint = clap::ValueHint::FilePath,
         add = ArgValueCompleter::new(complete::all_revision_files),
     )]
+    #[arg(value_name = "FILESETS")]
```
yuja, line 61: nit: combine `arg()`s for consistency?

(applies to file/show.rs, file/untrack.rs, interdiff.rs, log.rs, restore.rs, squash.rs)

cli/src/commands/absorb.rs
```diff
@@ -39,6 +39,7 @@ pub(crate) struct AbsorbArgs {
     #[arg(
         long, short,
         default_value = "@",
+        value_name="REVSET",
```
yuja, line 42: nit: insert space around `=`? (there are a few more instances)

yuja (approved):

cli/src/commands/absorb.rs
martinvonz: Done.

cli/src/commands/file/show.rs
martinvonz: Oops, done. Thanks.

martinvonz:
> Should we also change the accompanying doc comments then, e.g. path(s) / file(s) → fileset(s), revision(s) → revset(s)?

Yes, we should probably basically replace all "paths" by "paths matching these filesets" and "revisions" by "revisions matching these revsets". I'll try to remember to do that in a followup. I think this PR is an improvement on its own, and I should get back to the copy-tracking design.

PR #5138, opened 2024-12-18
show untracked files in jj status

cstoitner:
This adds the plumbing to get the untracked files from the snapshot to the status command and displays them similar to mercurial. I am not sure if this is the right way to do this.

This might be the last thing to fix https://github.com/jj-vcs/jj/issues/323.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/status.rs
+++ b/cli/src/commands/status.rs
@@ -47,7 +47,7 @@ pub(crate) fn cmd_status(
     command: &CommandHelper,
     args: &StatusArgs,
 ) -> Result<(), CommandError> {
-    let workspace_command = command.workspace_helper(ui)?;
+    let (workspace_command, _snapshot_stats) = command.workspace_helper_with_stats(ui)?;
     let repo = workspace_command.repo();
     let maybe_wc_commit = workspace_command
         .get_wc_commit_id()
--- a/cli/src/commands/status.rs
+++ b/cli/src/commands/status.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::io;
+
 use itertools::Itertools;
 use jj_lib::copies::CopyRecords;
 use jj_lib::repo::Repo;
@@ -47,7 +49,7 @@ pub(crate) fn cmd_status(
     command: &CommandHelper,
     args: &StatusArgs,
 ) -> Result<(), CommandError> {
-    let (workspace_command, _snapshot_stats) = command.workspace_helper_with_stats(ui)?;
+    let (workspace_command, snapshot_stats) = command.workspace_helper_with_stats(ui)?;
     let repo = workspace_command.repo();
     let maybe_wc_commit = workspace_command
         .get_wc_commit_id()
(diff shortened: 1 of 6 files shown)


yuja (commented):
Thanks! The implementation generally looks good to me.

cli/src/commands/status.rs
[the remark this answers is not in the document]
cstoitner: I will be out of town over the weekend and I'm not sure how much time a will have next week, so it might be a while before I get to that.

lib/src/working_copy.rs
```diff
@@ -258,6 +258,8 @@ pub enum UntrackedReason {
         /// Maximum allowed size.
         max_size: u64,
     },
+    /// Setting snapshot.auto-track did not match this file.
```
martinvonz, line 261: nit: I initially parsed "Setting" as a verb. Maybe "Setting" -> "Config"?

cstoitner: How about: "File does not match the fileset specified in snapshot.auto-track."?

cli/src/commands/status.rs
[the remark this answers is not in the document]
cstoitner: I have added a test for the status command output.

cli/tests/test_status_command.rs
```diff
@@ -368,3 +368,67 @@ fn test_status_simplify_conflict_sides() {
...
+
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    std::fs::write(repo_path.join("initially-untracked-file"), "...").unwrap();
+    std::fs::write(repo_path.join("always-untracked-file"), "...").unwrap();
+
+    let stdout = test_env.jj_cmd_success(&repo_path, &["status"]);
+    insta::assert_snapshot!(stdout, @r"
+    The working copy is clean
```
yuja, line 385: maybe better to suppress "The working copy is clean" if untracked paths exist?

cli/tests/test_status_command.rs
```diff
@@ -368,3 +368,67 @@ fn test_status_simplify_conflict_sides() {
...
+    Parent commit: zzzzzzzz 00000000 (empty) (no description set)
+    ");
+
+    test_env.jj_cmd_success(&repo_path, &["file", "track", "initially-untracked-file"]);
+
+    let stdout = test_env.jj_cmd_success(&repo_path, &["status"]);
+    insta::assert_snapshot!(stdout, @r"
+    Working copy changes:
+    A initially-untracked-file
+    ? always-untracked-file
```
yuja, line 398: nit: might be better to insert `Untracked paths:` header. These two lists are sorted separately.

yuja (approved):
Thanks!

yuja:
Just in case you didn't know, "when your first PR has been approved, we typically give you contributor access, so you can address any remaining minor comments and then merge the PR yourself when you're ready."
https://jj-vcs.github.io/jj/latest/contributing/#code-reviews

martinvonz:
I think this feature definitely deserves an entry in CHANGELOG.md. Could you send a PR for that?

PR #5145, opened 2024-12-19
config: remove check for relative path patterns, add doc about path equivalence

yuja:
# Checklist

If applicable:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1256,4 +1256,13 @@ Condition keys:
 
 * `--when.repositories`: List of paths to match the repository path prefix.
 
+Paths should be absolute. Each path component (directory or file name, drive
+letter, etc.) is compared case-sensitively on all platforms. A path starting
+with `~` is expanded to the home directory. On Windows, directory separator may
+be either `\` or `/`. (Beware that `\` needs escape in double-quoted strings.)
+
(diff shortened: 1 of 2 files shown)


docs/config.md
```diff
@@ -1256,4 +1256,12 @@ Condition keys:
 
 * `--when.repositories`: List of paths to match the repository path prefix.
 
+Paths should be absolute. Each path component (directory or file name, drive
+letter, etc.) is compared case-sensitively on all platforms. A path starting
+with `~` is expanded to the home directory. On Windows, directory separator may
+be either `\` or `/`. (Beware that `\` needs escape in double-quoted strings.)
+
+Use `jj root` to see the workspace root directory. The repository path is
+usually `<workspace_root>/.jj/repo`.
```
martinvonz, line 1265: Why mention the `.jj/repo` part? Is the point that the repo is not under `jj root` if you're in secondary workspace? Maybe something like "Note that the repository path is in the main workspace if you're using multiple workspaces with `jj workspace`."?

martinvonz (approved):

docs/config.md
yuja: I just needed to explain "the repository path" somehow. Copied your suggestion, thanks.

PR #5149, opened 2024-12-19
Fixed auto-detected typos

gtsiam:
I discovered `jj` a few days ago, and it seems pretty neat!

While I familiarize myself with the codebase (there will be an actual PR at some point, I promise), here's a few typos `typos` was screaming at me for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -274,7 +274,7 @@ Thanks to the people who made this release happen!
 * New template functions `pad_start()`, `pad_end()`, `truncate_start()`, and
   `truncate_end()` are added.
 
-* Add a new template alias `bultin_log_compact_full_description()`.
+* Add a new template alias `builtin_log_compact_full_description()`.
 
 * Added the config settings `diff.color-words.context` and `diff.git.context` to
   control the default number of lines of context shown.
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -556,7 +556,7 @@ impl CommandHelper {
     /// Returns true if the current operation is considered to be the head.
     pub fn is_at_head_operation(&self) -> bool {
         // TODO: should we accept --at-op=<head_id> as the head op? or should we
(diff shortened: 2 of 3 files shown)


gtsiam:
> Grant of Copyright License. Subject to the terms and conditions of this Agreement, You hereby grant to Google and to recipients of software distributed by Google a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable copyright license to reproduce, prepare derivative works of, publicly display, publicly perform, sublicense, and distribute Your Contributions and such derivative works.

WTF?

So I'm granting google specifically the right to do whatever the f\*\*k it wants with my contributions irrespective of the project license now and forever? Uhh... No? I'd much rather start my own fork than this.

You're claiming that this is a community project, even though google is involved. That rings hollow if I have to sign this shit.

arxanas:
See https://github.com/jj-vcs/jj/discussions/4849 for additional discussion.

I am not a lawyer, but I'm not sure that the specific text you're quoting is as unusual as you're remarking.

- Google probably wants the right to distribute the project code for fairly mundane reasons. As an example, [GitHub includes similar terms here](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#d-user-generated-content), although much more readable than the Google CLA. (I personally recall being shocked by seeing similar terms in the Facebook terms until somebody explained to me that they need the right to reproduce and redistribute my photos for free so that my friends can actually see them.)
- I'm not sure that the text actually grants Google rights regardless of the future license of the project (in that I think 'sublicense' doesn't let you change the license arbitrarily, only add restrictions, and you would expect a consumer to be able to use the original project code and license as an alternative), but I can't say authoritatively.
- There's commentary in the above discussion noting that copyright is still held by the original authors.

gtsiam:
@arxanas Yeah, I just finished diving into the CLA. You (people at google) should really have a word with your legal department. IMHO this is all very unnecessary (See discord).

FYI, I ended up signing the CLA. Granting google an apache license plus a DCO seems reasonable.

martinvonz (approved):

PR #5152, opened 2024-12-19
docs: Git doesn't use bookmarks, jj does

homeworkprod:
# Checklist

If applicable:

### DIFF
--- a/docs/github.md
+++ b/docs/github.md
@@ -67,8 +67,10 @@ changes.
 ## Working in a Git co-located repository
 
 After doing `jj git init --colocate`, Git will be in a [detached HEAD
-state][detached], which is unusual, as Git mainly works with bookmarks. In a
-co-located repository, every `jj` command will automatically synchronize
+state][detached], which is unusual, as Git mainly works with named branches; jj
+does not.
+
+In a co-located repository, every `jj` command will automatically synchronize
 Jujutsu's view of the repo with Git's view. For example, `jj commit` updates the
 HEAD of the Git repository, enabling an incremental migration.
 


docs/github.md
```diff
@@ -67,8 +67,10 @@ changes.
 ## Working in a Git co-located repository
 
 After doing `jj git init --colocate`, Git will be in a [detached HEAD
-state][detached], which is unusual, as Git mainly works with bookmarks. In a
-co-located repository, every `jj` command will automatically synchronize
+state][detached], which is unusual; this is because jj mainly works with
+bookmarks, not branches.
```
arxanas, line 71: comment: I don't know if this is emphasizing the right thing. I would say that this is unusual because jj mainly works with commits (or heads) rather than because it mainly works with bookmarks, so a "detached HEAD" state is unusual in Git but common in jj. You could also say it's unusual because Git strongly encourages that all work be reachable from some branch, while jj does not, or somehow express that.

docs/github.md
```diff
@@ -67,7 +67,7 @@ changes.
 ## Working in a Git co-located repository
 
 After doing `jj git init --colocate`, Git will be in a [detached HEAD
-state][detached], which is unusual, as Git mainly works with bookmarks. In a
+state][detached], which is unusual, as jj mainly works with bookmarks. In a
```
arxanas: Considering the first commit in isolation, I think this ought to be `as Git mainly works with branches`, rather than `as jj mainly works with bookmarks`. I believe this sentence is meant to expand on the fact that "detached HEAD state" is unusual for Git itself.

jennings: Someone who has used Git might have learned that being in a detached HEAD state is dangerous or a mistake. I think this section's goal is to reassure that person that it's okay, and jj will record any commits you make in Git as long as you run `jj` before you do any `git checkout` or `git switch` commands (is that right? I never mutate repos with git tooling so I don't actually know).

homeworkprod: @arxanas Good observation. I have updated the changes with that in mind.

@jennings Good point. May I suggest that tackling that is out of scope here? At this point I'm just trying to get the facts right :)

docs/github.md
homeworkprod: As suggested on Discord, among other things, I have left out mentioning bookmarks altogether. That also goes along with your other comment about the intended meaning of the original sentence likely being that Git uses (named) branches, not jj using bookmarks.

arxanas (approved):
Discussed in Discord: the PR fixes the immediate correctness issue, but discussion regarding the PR has also made the fact apparent that this particular documentation should probably also discuss what the alternative to using branches/bookmarks is (as a separate matter).

martinvonz:
I'll merge this but I agree that the text could be improved.

Issue #4210, opened 2024-08-04
FR: Automatically delete all lines below (and including) a "scissor line" in description buffers


gpanders:
**Is your feature request related to a problem? Please describe.**

Git supports something they call a "scissor line" in git commit messages. This line and anything below it will be automatically deleted by git before applying the commit message. This allows the commit message buffer to contain text without being commented out. In particular, this is useful for including diffs into commit message buffers.

`jj` now supports including diffs in description message buffers with the `draft_commit_description` template and the `diff.git()` function, e.g.:

```toml
[templates]
draft_commit_description = '''
concat(
  description,
  surround(
    "\nJJ: This commit contains the following changes:\n", "",
    indent("JJ:     ", diff.summary()),
  ),
  "\n",
  diff.git(),
)
'''
```

However, because the diff is not commented with the `JJ: ` comment prefix, the diff becomes part of the description message. Prepending `JJ: ` to the range of the diff does not work because editors lose the ability to syntax highlight the diff (it also makes the diff much harder to read).

I would like jj to support a "scissor line" concept similar to git, where any line below the scissor line is automatically removed before applying the description message, even if it does not contain the `JJ: ` comment prefix.

**Describe the solution you'd like**

Add a new template function called `scissor_line()` (or `cut_line()` or similar, I am indifferent to the name). This function inserts a line according to another config option (`ui.scissor-line` perhaps) which defaults to `JJ: ----------- >8 -----------` (the `>8` looks like a pair of scissors, hence the name). This is the same value used by git: again, I'm indifferent to the actual value here, but this works well enough.

When the user exits their editor after writing a description message, `jj` removes the scissor line and all lines after it.

If/when `jj` supports writing/modifying multiple descriptions in the same file, then `jj` would only delete lines between 2 successive scissor lines (or perhaps we would need two separate markers: "start scissor/cut region" and  "end scissor/cut region".

**Describe alternatives you've considered**

This is possible to solve already today with some custom editor integration. I am adding a scissor line "manually" to my `draft_commit_description` template and am using a hook in my editor (Neovim) to modify the temporary file to remove the scissor line. This works pretty well, but it'd be nice to be able to only use `jj` for this (and if it does not rely on custom editor integrations it would be easier to include things like diffs in default description messages).

yuja:
No template function would be needed. We can just inline `"----------- >8 -----------"` or make it a template alias for readability. We'll instead need code changes in parsing/cleanup part.

[...]

Issue #1946, opened 2023-08-02
FR: `git commit --verbose` equivalent for `jj describe`


sullyj3:
Often when writing a commit message, it's useful to be able to see the diff at a glance. Git has `git commit --verbose`, which displays the diff inline in your editor while you're writing the message. This saves having to try to remember what changes you made. 
I personally use this all the time. I have 
```
alias gca='git commit --all --verbose'
```
which I use for committing almost exclusively.

It would be cool if `jj` supported this as well! I'm not wedded to the name `--verbose`, which I think could be clearer (and in fact looking at the help it conflicts with an existing flag with different functionality). Maybe `--show-diff`?

ilyagr:
This could apply to quite a few commands (`describe`, `split`, `squash`, etc)

martinvonz:
Maybe we can have a configurable template for producing that. We currently include something like this:

```
JJ: This commit contains the following changes:
JJ:     M lib/src/working_copy.rs
JJ:     M lib/tests/test_working_copy.rs
```

Maybe that could be replaced by a template like `"This commit contains the following changes:\n" ++ indent("    ", diffsummary)`, and you could replace the `diffsummary` by `diff` (or `gitdiff`, perhaps) to get the full diff.

We don't yet have support for including a diff in templates, so we'd need to implement that first.

#1354 is slightly related.

9999years:
I went looking for this feature, remembered `jj` has templates, and was a little surprised that `jj describe` didn't have a `--template` option for this yet, so I think templates would be a natural place to implement this. (Although I suspect this is a common enough use-case that a dedicated flag might be worthwhile, even if it's just an alias for `--template=commit-with-diff` or something.)

Cretezy:
Maybe `jj describe --diff` should be the command? `jj split --diff` and `jj squash --diff` too.

Cretezy:
Looking for consensus before implementing this:

1. Should we use `--show-diff` for `diff`/`squash`/`split`?
2. Should we instead use a template feature for this?
  1. If using template, how will this be used? Will jj ship with a built-in template for this?

yuja:
> 1. Should we use `--show-diff` for `diff`/`squash`/`split`?

I think config knob is better. It was discussed previously in
https://github.com/martinvonz/jj/pull/3470#issuecomment-2053014846

> 2. Should we instead use a template feature for this?

[...]

PR #5155, opened 2024-12-20
describe: add ignore-rest directive

bryceberger:
Implements #4210

~~Instead of a scissor _line_, this implements a scissor _section_. Everything between a pair of scissor lines is ignored. This allows it to work with #3828, as long as you put the ending line. Currently the beginning and ending lines are the same, so only one alias is required. Could easily be separated.~~

Implements a single scissor line, since `JJ: describe` lines are parsed first. Everything after a scissor line, before a `JJ: describe` line is ignored.

Open to other ideas on the style of the scissor lines. Additionally, the `scissor` template alias does *not* insert the "Do not modify or remove the line above." line from git --- should it? I lean softly towards no, since the user would have to seek out the alias to end up with it.

# Example usage
```toml
[templates]
draft_commit_description = '''
concat(
  description,
  "\n",
  scissors,
  diff.stat(72),
  diff.git(),
  scissors,
)
'''
```

# Checklist

If applicable:

### DIFF
--- a/cli/src/description_util.rs
+++ b/cli/src/description_util.rs
@@ -5,6 +5,7 @@ use std::path::Path;
 use bstr::ByteVec as _;
 use indexmap::IndexMap;
 use indoc::indoc;
+use itertools::FoldWhile;
 use itertools::Itertools;
 use jj_lib::backend::CommitId;
 use jj_lib::commit::Commit;
@@ -28,8 +29,17 @@ where
 {
     let description = lines
         .into_iter()
-        .filter(|line| !line.as_ref().starts_with("JJ:"))
-        .fold(String::new(), |acc, line| acc + line.as_ref() + "\n");
+        .fold_while(String::new(), |acc, line| {
+            let line = line.as_ref();
+            if line.strip_prefix("JJ: ignore-rest").is_some() {
+                FoldWhile::Done(acc)
+            } else if line.starts_with("JJ:") {
+                FoldWhile::Continue(acc)
+            } else {
+                FoldWhile::Continue(acc + line + "\n")
+            }
+        })
+        .into_inner();
     text_util::complete_newline(description.trim_matches('\n'))
 }
 
--- a/cli/tests/test_describe_command.rs
+++ b/cli/tests/test_describe_command.rs
@@ -140,6 +140,35 @@ fn test_describe() {
     std::fs::write(&edit_script, "fail").unwrap();
     let stderr = test_env.jj_cmd_failure(&repo_path, &["describe"]);
     assert!(stderr.contains("exited with an error"));
+
+    // ignore everything after the first ignore-rest line
+    std::fs::write(
+        &edit_script,
+        indoc! {"
+            write
+            description from editor
+
+            content of message from editor
+            JJ: ignore-rest
+            content after ignore line should not be included
+            JJ: ignore-rest
+            ignore everything until EOF or next description
+        "},
+    )
+    .unwrap();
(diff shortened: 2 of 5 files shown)


bryceberger:
Updated the `jjdescription` tree-sitter grammar at https://github.com/kareigu/tree-sitter-jjdescription/pull/2.

todo: tests, docs

yuja:
> Instead of a scissor line, this implements a scissor section. Everything between a pair of scissor lines is ignored. This allows it to work with https://github.com/jj-vcs/jj/pull/3828

Just fyi, #3828 should be compatible with scissors. The "JJ: describe " line should be parsed first.

(BTW, we tend to include details in commit messages than in PR description.)

bryceberger:
> Just fyi, https://github.com/jj-vcs/jj/pull/3828 should be compatible with scissors. The "JJ: describe " line should be parsed first.

Oh, didn't notice that. Updated to only require one line.

bryceberger:
Paging @ilyagr and @arxanas, as you both had comments about the design of a scissor line in #4210

martinvonz:
I think @arxanas had a good point about discoverability. Something like "JJ: ignore-after" or "JJ: ignore-rest" is a lot clearer and more searchable.

bryceberger:
Changed to `JJ: ignore-rest`. Slightly less visible, but if the user wants to have a more visible marker they can easily insert it in their default template.

I do have a concern about docs. There didn't seem to be a good place to put it, so right now it's only mentioned in the changelog and "default description" config. If we keep adding similar directives (`JJ: describe ...`, `JJ: ignore-rest`, ..?), there should probably be a dedicated spot for them.

bryceberger:
Just found the line in the default config referencing #1946, which this also provides the ability to solve.

cli/src/description_util.rs
```diff
@@ -28,8 +29,17 @@ where
 {
     let description = lines
         .into_iter()
-        .filter(|line| !line.as_ref().starts_with("JJ:"))
-        .fold(String::new(), |acc, line| acc + line.as_ref() + "\n");
+        .fold_while(String::new(), |acc, line| {
+            let line = line.as_ref();
+            if line == "JJ: ignore-rest" {
```
yuja, line 34: nit: maybe better to trim trailing whitespace and `-` as we do for `JJ: describe "` line?

cli/tests/test_describe_command.rs
yuja, line 1: Can you squash this into the implementation patch?

It might be also better to update the existing unit (and/or CLI) tests instead of adding new CLI tests. CLI tests are slow.

yuja (approved):
Thanks!

cli/src/description_util.rs
bryceberger: The detection of the `"JJ: describe "` is just `line.strip_prefix(...)`, so I changed it to that.

yuja: It's a bit stricter in that "JJ: ignore-rest<SPACE>" is required (and would fail to parse if trailing argument contained junk.)

BTW, `.strip_prefix()` can be replaced with `.starts_with()`.

arxanas:
Sorry for the late reply. The overall approach and naming seems good to me. If we add more `JJ: ` directives, we should systematically document them and consider their interactions (such as whether `JJ: new-directive` can appear after `JJ: ignore-rest`, in the same way as the implementation currently appears to allow a `JJ: describe` directive to follow `JJ: ignore-rest`).

PR #5160, opened 2024-12-20
Unify alias and default command parsing

gtsiam:
I'm trying to pull config parsing out of the cli. But to do that, we need to untangle config parsing and argument parsing. This is part of that.

As a bonus, you can now specify aliases some of the simpler aliases/default commands as a space-delimited string.

# Checklist

If applicable:


bnjmnt4n:
> I'm trying to pull config parsing out of the cli. But to do that, we need to untangle config parsing and argument parsing. This is part of that.

Any reason for this?

> As a bonus, you can now specify aliases some of the simpler aliases/default commands as a space-delimited string.

This wouldn't correctly handle cases where one of the arguments is a quoted string with whitespace within, would it? I'm not sure if there's a good enough justification for this then.

cli/src/cli_util.rs
```diff
@@ -3252,70 +3250,88 @@ fn resolve_default_command(
...
     for command in app.get_subcommands() {
         real_commands.insert(command.get_name());
         for alias in command.get_all_aliases() {
             real_commands.insert(alias);
         }
     }
-    for alias in defined_aliases.intersection(&real_commands).sorted() {
+
+    // Collect all the aliases defined by the user.
+    let defined_aliases: HashMap<String, Vec<String>> = config.get("aliases")?;
```
yuja: > The only significant one I'm aware of is that jj will now always complain about
malformed aliases, instead of waiting for first use.

It might be okay to warn all bad aliases, but I think it's better to expand valid aliases instead of rejecting everything.

CHANGELOG.md
```diff
@@ -53,6 +53,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Configuration files now support [conditional
   variables](docs/config.md#conditional-variables).
 
+* Command aliases and default commands can now be defined as space-delimited strings.
```
yuja, line 56: I wouldn't add this. We've overloaded some external command parameters, but that's for compatibility reasons.

gtsiam: I don't see why not? We already support this for default-command, so why not for aliases as well?

Besides, I'd expect this to be possible coming from git.

gtsiam: Also from my other comment:

> we could forbid quotes (", ' and `)

cli/src/cli_util.rs
gtsiam: On second thought, keeping the original behaviour is probably better.

CHANGELOG.md
yuja: ah yes, default-command works that way for compatibility reasons. Aliases doesn't have this problem.

> I don't see why not?

because it's not difficult to write a TOML array expression? A string version might also provide a false implication that it supports full shell syntax.

yuja:
cc @matts1 as he was working on https://github.com/jj-vcs/jj/issues/3673

gtsiam:
I'll remove whitespace splitting and restore the old behaviour for this

gtsiam:
Ok, this is just the refactor now. I tried to do too much in one pr I think.

Only real change now should be that you can specify single-command (no argument) aliases with a string. This is already possible for default-command, but now it is also possible in aliases.

cli/src/cli_util.rs
martinvonz, line 1: > cli: simplify argument expansion code

Could you add some more detail? This patch adds 39 lines of code, so it's not an obvious simplification. It would be good to know why you're doing it (if it's not only to make the code simpler). Since it's a non-trivial patch, it's also good to explain how you're refactoring the code so the readers don't have to spend so much time to try to figure out what's going on.

PR #5164, opened 2024-12-20
cli: git push: allow signing commits on push

bnjmnt4n:
This adds the `git.sign-on-push` configuration which can be used to
automatically sign unsigned commits before pushing to a Git remote.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fmt;
 use std::io;
@@ -19,8 +20,11 @@ use std::io::Write;
 
 use clap::ArgGroup;
 use clap_complete::ArgValueCandidates;
+use indexmap::IndexSet;
 use itertools::Itertools;
 use jj_lib::backend::CommitId;
+use jj_lib::commit::Commit;
+use jj_lib::commit::CommitIteratorExt as _;
 use jj_lib::config::ConfigGetResultExt as _;
 use jj_lib::git;
 use jj_lib::git::GitBranchPushTargets;
@@ -34,6 +38,7 @@ use jj_lib::refs::LocalAndRemoteRef;
 use jj_lib::repo::Repo;
 use jj_lib::revset::RevsetExpression;
 use jj_lib::settings::UserSettings;
+use jj_lib::signing::SignBehavior;
 use jj_lib::str_util::StringPattern;
 use jj_lib::view::View;
 
@@ -295,7 +300,38 @@ pub fn cmd_git_push(
         return Ok(());
     }
 
-    validate_commits_ready_to_push(ui, &bookmark_updates, &remote, &tx, args)?;
+    let sign_behavior = if tx.settings().get_bool("git.sign-on-push")? {
+        Some(SignBehavior::Own)
+    } else {
+        None
+    };
+    let commits_to_sign =
+        validate_commits_ready_to_push(ui, &bookmark_updates, &remote, &tx, args, sign_behavior)?;
+    if !args.dry_run && !commits_to_sign.is_empty() {
+        if let Some(sign_behavior) = sign_behavior {
+            let num_updated_signatures = commits_to_sign.len();
+            let num_rebased_descendants;
+            (num_rebased_descendants, bookmark_updates) = sign_commits_before_push(
+                &mut tx,
+                commits_to_sign,
+                sign_behavior,
+                bookmark_updates,
+            )?;
+            if let Some(mut formatter) = ui.status_formatter() {
+                writeln!(
+                    formatter,
+                    "Updated signatures of {num_updated_signatures} commits"
+                )?;
+                if num_rebased_descendants > 0 {
+                    writeln!(
+                        formatter,
+                        "Rebased {num_rebased_descendants} descendant commits"
+                    )?;
+                }
+            }
+        }
+    }
+
     if let Some(mut formatter) = ui.status_formatter() {
         writeln!(formatter, "Changes to push to {remote}:")?;
         print_commits_ready_to_push(formatter.as_mut(), tx.repo(), &bookmark_updates)?;
@@ -335,14 +371,17 @@ pub fn cmd_git_push(
 }
 
 /// Validates that the commits that will be pushed are ready (have authorship
-/// information, are not conflicted, etc.)
+/// information, are not conflicted, etc.).
+///
+/// Returns the list of commits which need to be signed.
 fn validate_commits_ready_to_push(
     ui: &Ui,
     bookmark_updates: &[(String, BookmarkPushUpdate)],
     remote: &str,
     tx: &WorkspaceCommandTransaction,
     args: &GitPushArgs,
-) -> Result<(), CommandError> {
+    sign_behavior: Option<SignBehavior>,
+) -> Result<Vec<Commit>, CommandError> {
     let workspace_helper = tx.base_workspace_helper();
     let repo = workspace_helper.repo();
 
@@ -369,6 +408,13 @@ fn validate_commits_ready_to_push(
     } else {
         Box::new(|_: &CommitId| Ok(false))
     };
+    let sign_settings = sign_behavior.map(|sign_behavior| {
+        let mut sign_settings = settings.sign_settings();
+        sign_settings.behavior = sign_behavior;
+        sign_settings
+    });
+
+    let mut commits_to_sign = vec![];
 
     for commit in workspace_helper
         .attach_revset_evaluator(commits_to_push)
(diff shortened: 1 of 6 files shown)


pylbrecht:
Heads-up: with #4853 hopefully soon to be merged, you may have a conflict-inducing rebase ahead.

We removed `builtin_sig_detailed` and only kept templating primitives for signatures. This means you will have to write your own template in `cli/tests/test_git_push.rs` to assert against signature presence.

cli/src/commands/git/push.rs
```diff
@@ -295,8 +300,39 @@ pub fn cmd_git_push(
...
+                    "Dry-run requested, skipped updating signatures of {num_updated_signatures} \
+                     commits"
+                )?;
+                if num_rebased_descendants > 0 {
+                    writeln!(
+                        formatter,
+                        "Dry-run requested, skipped rebasing of {num_rebased_descendants} \
+                         descendant commits"
+                    )?;
+                }
```
martinvonz, line 322: I'm not sure this "skipped rebasing" message is very useful but I don't feel strongly at all. To me, the "rebased N descendant commits" messages are useful mostly so you can notice if there are many more rebased commits than you expected, which may indicate that you passed the wrong commit as argument, for example. In this case, it seems like the "Changes to push to origin:" stuff is much more likely to help you spot an error.

So you have a weak vote for removing it from me, but feel free to keep it.

Actually, do you think we can skip calling `sign_commits_before_push()` completely when `--dry-run` is passed? Maybe it then becomes simple enough that it can be shared with the `jj sign` command @pylbrecht is working on in #4747.

cli/src/commands/git/push.rs
```diff
@@ -369,6 +407,13 @@ fn validate_commits_ready_to_push(
     } else {
         Box::new(|_: &CommitId| Ok(false))
     };
+    let mut sign_settings = settings.sign_settings();
+    sign_settings.behavior = if settings.get_bool("git.sign-on-push").unwrap_or(false) {
```
martinvonz, line 411: nit: we usually put the default value in `misc.toml` (or other config file) so we can expect a valid value to be available and do `get_bool(...)?` when we read it

cli/src/commands/git/push.rs
```diff
@@ -418,8 +463,72 @@ fn validate_commits_ready_to_push(
             }
             return Err(error);
         }
+        if commit.is_signed() != sign_settings.should_sign(commit.store_commit()) {
```
martinvonz, line 466: nit: `!commit.is_signed() && sign_settings.should_sign(commit.store_commit())` is slightly easier to read to me, and makes it clearer that we don't worry about dropping signatures here (we can never have it `is_signed()` and not `should_sign()`)

cli/src/commands/git/push.rs
```diff
@@ -418,8 +463,72 @@ fn validate_commits_ready_to_push(
...
+fn sign_commits_before_push(
+    tx: &mut WorkspaceCommandTransaction,
+    bookmark_updates: Vec<(String, BookmarkPushUpdate)>,
+    commits_to_sign: Vec<Commit>,
+    dry_run: bool,
+) -> Result<(usize, Vec<(String, BookmarkPushUpdate)>), CommandError> {
+    let sign_behavior = if tx.settings().get_bool("git.sign-on-push").unwrap_or(false) {
+        SignBehavior::Own
+    } else {
+        SignBehavior::Keep
```
martinvonz, line 486: When we're just keeping signatures, what's the point of even transforming them? I.e. can. we skip calling `sign_commits_before_push()` completely when `git.sign-on-push` is off?

cli/src/commands/git/push.rs
```diff
@@ -418,8 +463,72 @@ fn validate_commits_ready_to_push(
...
+///
+/// Returns the number of commits with rebased descendants and the updated list
+/// of bookmark names and corresponding [`BookmarkPushUpdate`]s.
+fn sign_commits_before_push(
+    tx: &mut WorkspaceCommandTransaction,
+    bookmark_updates: Vec<(String, BookmarkPushUpdate)>,
+    commits_to_sign: Vec<Commit>,
+    dry_run: bool,
+) -> Result<(usize, Vec<(String, BookmarkPushUpdate)>), CommandError> {
+    let sign_behavior = if tx.settings().get_bool("git.sign-on-push").unwrap_or(false) {
```
martinvonz, line 483: nit: can we reuse the `sign_behavior` value from before by passing it in?

martinvonz (commented):
Thanks! Signing is really slow for me so I'm looking forward to this feature

cli/src/commands/git/push.rs
bnjmnt4n: Yeah, I was initially thinking we might want to check the total number of descendants rebased as part of the dry run, but I agree that it's probably not the most useful and just the changes to push (or the number of commits to sign is enough).

cli/src/commands/git/push.rs
```diff
@@ -295,17 +300,56 @@ pub fn cmd_git_push(
...
     if let Some(mut formatter) = ui.status_formatter() {
+        if num_updated_signatures > 0 {
+            writeln!(
+                formatter,
+                "Updated signatures of {num_updated_signatures} commits"
+            )?;
+            if num_rebased_descendants > 0 {
+                writeln!(
+                    formatter,
+                    "Rebased {num_rebased_descendants} descendant commits"
```
bnjmnt4n, line 345: Should we still call this "Rebased X descendant commits"? Technically these have been reparented, but I suspect most users won't know/need to know. the difference

bnjmnt4n (commented):
Updated to address most of the comments.

cli/src/commands/git/push.rs
```diff
@@ -295,17 +300,56 @@ pub fn cmd_git_push(
...
+            sign_commits_before_push(&mut tx, commits_to_sign, sign_behavior, bookmark_updates)?;
+    } else {
+        num_rebased_descendants = 0;
+    }
+
     if let Some(mut formatter) = ui.status_formatter() {
+        if num_updated_signatures > 0 {
+            writeln!(
+                formatter,
+                "Updated signatures of {num_updated_signatures} commits"
```
bnjmnt4n, line 340: Do we need to be thorough about describing this? E.g. listing the number of newly signed commits, number of signatures dropped, for both the commits being pushed and the rebased descendants.

cli/src/commands/git/push.rs
```diff
@@ -335,14 +379,17 @@ pub fn cmd_git_push(
...
+///
+/// Returns the list of commits which need to be signed.
 fn validate_commits_ready_to_push(
     ui: &Ui,
     bookmark_updates: &[(String, BookmarkPushUpdate)],
     remote: &str,
     tx: &WorkspaceCommandTransaction,
     args: &GitPushArgs,
-) -> Result<(), CommandError> {
+    sign_behavior: &Option<SignBehavior>,
```
yuja, line 391: nit: just pass by value. `SignBehavior` is `Copy`.

cli/src/commands/git/push.rs
```diff
@@ -295,17 +300,56 @@ pub fn cmd_git_push(
...
-    validate_commits_ready_to_push(ui, &bookmark_updates, &remote, &tx, args)?;
+    let sign_behavior = if tx.settings().get_bool("git.sign-on-push")? {
+        Some(SignBehavior::Own)
+    } else {
+        None
+    };
+    let commits_to_sign =
+        validate_commits_ready_to_push(ui, &bookmark_updates, &remote, &tx, args, &sign_behavior)?;
+    let num_updated_signatures = commits_to_sign.len();
+
```
yuja, line 311: Maybe we'll need `.check_rewritable()` somewhere (or exclude immutable commits)?

yuja: Oh, never mind. It appears to be excluded by `.union(workspace_helper.env().immutable_heads_expression())`.

bnjmnt4n: Hmm, is it actually? I thought given it's just a union and not an intersection with the old heads, there might be some items which are descendants of the old heads but still immutable.

cli/src/commands/git/push.rs
martinvonz: I'm not sure. This seems fine to me. We can adjust later if people think it's not sufficient.

cli/src/commands/git/push.rs
```diff
@@ -295,17 +300,56 @@ pub fn cmd_git_push(
...
+    if args.dry_run {
+        if let Some(mut formatter) = ui.status_formatter() {
+            if num_updated_signatures > 0 {
+                writeln!(
+                    formatter,
+                    "Dry-run requested, skipped updating signatures of {num_updated_signatures} \
+                     commits"
+                )?;
+            }
```
martinvonz, line 320: I think we can skip this message. What do you think? Then we should not have to duplicate the next two lines either, as long as we skip the call to `sign_commits_before_push()` in dry-run mode. I think the `sign_behavior` variable can be replaced by a `should_sign: bool` now that we only use `SignBehavior::Own`. Just an idea, possibly a bad one, so you decide.

cli/src/commands/git/push.rs
yuja: It's `(old_heads | immutable_heads)..new_heads` so I think `commits_to_push` are actually `mutable_commits_to_push`. In any case, adding tests will help.

There was inline comment in the original patch 6ae4f458, but apparently it was lost.

cli/src/commands/git/push.rs
```diff
@@ -295,8 +300,44 @@ pub fn cmd_git_push(
...
+            writeln!(
+                formatter,
+                "Updated signatures of {num_updated_signatures} commits"
+            )?;
+            if num_rebased_descendants > 0 {
+                writeln!(
+                    formatter,
+                    "Rebased {num_rebased_descendants} descendant commits"
+                )?;
+            }
```
martinvonz, line 339: optional: Move this messages inside the `match` block above and reduce scope of `num_updated_signatures` and `num_rebased_descendants` since they can only be non-zero there. Then the `match` can also become an `if let`. The `if let Some(mut formatter) = ui.status_formatter()` will be duplicated, though.

martinvonz (approved):

cli/src/commands/git/push.rs
martinvonz: I don't think users will be confused. The reparenting is the same as rebasing in this case anyway.

Issue #5088, opened 2024-12-12
FR: Add `--reversed` flag to `evolution-log`


zx8:
**Describe the solution you'd like**

Related to https://github.com/martinvonz/jj/issues/4190, I just tried to use `jj evolog --reversed` like I do with `jj log` and noticed it was missing.

**Describe alternatives you've considered**

`jj evolog --color=always | tac`

JDSeiler:
I've not contributed before but this seems like something I could wrap my head around. Any objections to me giving this a shot?

martinvonz:
No objections! Thanks :)

I haven't looked carefully but it looks like it should be straightforward to copy how it's done in `jj log --reversed`.

JDSeiler:
Perfect! I started looking through the code and getting set up last night. I'll report back if I get stuck / derailed.

PR #5170, opened 2024-12-21
Add `--reversed` flag to `evolog`

JDSeiler:
Resolves https://github.com/jj-vcs/jj/issues/5088

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/evolog.rs
+++ b/cli/src/commands/evolog.rs
@@ -12,10 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::convert::Infallible;
+
 use clap_complete::ArgValueCandidates;
 use itertools::Itertools;
+use jj_lib::backend::BackendError;
+use jj_lib::backend::CommitId;
 use jj_lib::commit::Commit;
 use jj_lib::dag_walk::topo_order_reverse_ok;
+use jj_lib::graph::reverse_graph;
+use jj_lib::graph::GraphEdge;
+use jj_lib::graph::GraphNode;
 use jj_lib::matchers::EverythingMatcher;
 use tracing::instrument;
 
@@ -28,6 +35,7 @@ use crate::command_error::CommandError;
 use crate::commit_templater::CommitTemplateLanguage;
 use crate::complete;
 use crate::diff_util::DiffFormatArgs;
+use crate::graphlog::convert_edge_type;
 use crate::graphlog::get_graphlog;
 use crate::graphlog::Edge;
 use crate::graphlog::GraphStyle;
@@ -57,6 +65,9 @@ pub(crate) struct EvologArgs {
         value_name = "LIMIT"
     )]
     deprecated_limit: Option<usize>,
+    /// Show revisions in the opposite order (older revisions first)
+    #[arg(long)]
+    reversed: bool,
     /// Don't show the graph, show a flat list of revisions
     #[arg(long)]
     no_graph: bool,
@@ -151,14 +162,31 @@ pub(crate) fn cmd_evolog(
     if !args.no_graph {
         let mut raw_output = formatter.raw()?;
         let mut graph = get_graphlog(graph_style, raw_output.as_mut());
-        for commit in commits {
-            let edges = commit
-                .predecessor_ids()
-                .iter()
-                .map(|id| Edge::Direct(id.clone()))
-                .collect_vec();
+
+        let commit_dag: Vec<GraphNode<Commit>> = commits
+            .into_iter()
+            .map(|c| -> Result<_, BackendError> {
+                let edges = c.predecessors().map_ok(GraphEdge::direct).try_collect()?;
+                Ok((c, edges))
+            })
+            .try_collect()?;
+
+        let iter_nodes = if args.reversed {
+            reverse_graph(commit_dag.into_iter().map(Result::<_, Infallible>::Ok)).unwrap()
+        } else {
+            commit_dag
+        };
+
+        for node in iter_nodes {
+            let (commit, edges) = node;
+            // Slightly different edge type used for display.
+            let graphlog_edges: Vec<Edge<CommitId>> = edges
+                .into_iter()
+                .map(|e| convert_edge_type(e).map(|e| e.id().clone()))
+                .collect();
             let mut buffer = vec![];
-            let within_graph = with_content_format.sub_width(graph.width(commit.id(), &edges));
+            let within_graph =
+                with_content_format.sub_width(graph.width(commit.id(), &graphlog_edges));
             within_graph.write(ui.new_formatter(&mut buffer).as_mut(), |formatter| {
                 template.format(&commit, formatter)
             })?;
@@ -180,12 +208,16 @@ pub(crate) fn cmd_evolog(
             let node_symbol = format_template(ui, &Some(commit.clone()), &node_template);
             graph.add_node(
                 commit.id(),
-                &edges,
+                &graphlog_edges,
                 &node_symbol,
                 &String::from_utf8_lossy(&buffer),
             )?;
         }
     } else {
+        if args.reversed {
+            commits.reverse();
+        }
+
         for commit in commits {
             with_content_format
                 .write(formatter, |formatter| template.format(&commit, formatter))?;
--- a/cli/src/graphlog.rs
+++ b/cli/src/graphlog.rs
@@ -18,6 +18,8 @@ use std::io::Write;
 
 use itertools::Itertools;
 use jj_lib::config::ConfigGetError;
+use jj_lib::graph::GraphEdge;
+use jj_lib::graph::GraphEdgeType;
 use jj_lib::settings::UserSettings;
 use renderdag::Ancestor;
 use renderdag::GraphRowRenderer;
@@ -29,6 +31,28 @@ pub enum Edge<T> {
     Direct(T),
     Indirect(T),
     Missing,
+}
+
+impl<T> Edge<T> {
+    pub fn map<U>(self, f: impl FnOnce(T) -> U) -> Edge<U> {
+        match self {
+            Edge::Direct(inner) => Edge::Direct(f(inner)),
+            Edge::Indirect(inner) => Edge::Indirect(f(inner)),
+            Edge::Missing => Edge::Missing,
+        }
+    }
+}
+
+/// Used to convert a jj_lib GraphEdge into a graphlog Edge.
+pub fn convert_edge_type<N>(e: GraphEdge<N>) -> Edge<N> {
+    // TODO: Is it possible for the ClI to use the GraphEdge type directly?
+    // These two types seem to correspond exactly, only differing in minor
+    // implementation details.
+    match e.edge_type {
+        GraphEdgeType::Missing => Edge::Missing,
+        GraphEdgeType::Direct => Edge::Direct(e.target),
+        GraphEdgeType::Indirect => Edge::Indirect(e.target),
+    }
 }
 
 pub trait GraphLog<K: Clone + Eq + Hash> {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj evolog` now accepts `--reversed`.
+
 ### Fixed bugs
 
 ## [0.25.0] - 2025-01-01
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -797,6 +797,7 @@ Lists the previous commits which a change has pointed to. The current commit of 
 
   Default value: `@`
 * `-n`, `--limit <LIMIT>` — Limit number of revisions to show
+* `--reversed` — Show revisions in the opposite order (older revisions first)
 * `--no-graph` — Don't show the graph, show a flat list of revisions
 * `-T`, `--template <TEMPLATE>` — Render each revision using the given template
 
--- a/cli/tests/test_evolog_command.rs
+++ b/cli/tests/test_evolog_command.rs
@@ -348,3 +348,70 @@ fn test_evolog_with_no_template() {
     - name_placeholder
     "#);
 }
+
+#[test]
+fn test_evolog_reversed_no_graph() {
+    let test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    test_env.jj_cmd_ok(&repo_path, &["describe", "-m", "a"]);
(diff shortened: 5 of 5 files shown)


PhilipMetzger:
This can be at most two commits until Yuya shows up for the `ReverseGraphIterator` problem, so squash commits 1, 2 and 4 into the first commit. Also thanks for implementing it.

JDSeiler:
> This can be at most two commits until Yuya shows up for the `ReverseGraphIterator` problem, so squash commits 1, 2 and 4 into the first commit. Also thanks for implementing it.

To make sure I understand: You'd like the second commit to be what is currently https://github.com/jj-vcs/jj/pull/5170/commits/32c4449db4011f308444166a91adecadeb875080 and then the first commit to be everything else? Some of the clippy lints apply to code from 32c444 so I might have to split it.

PhilipMetzger:
> To make sure I understand: You'd like the second commit to be what is currently [32c4449](https://github.com/jj-vcs/jj/commit/32c4449db4011f308444166a91adecadeb875080) and then the first commit to be everything else?

No, you can add the flag, most of the implementation, test and CHANGELOG entry in a single commit while keeping a second commit (https://github.com/jj-vcs/jj/pull/5170/commits/32c4449db4011f308444166a91adecadeb875080) with the fixed clippy lints.

JDSeiler:
> > To make sure I understand: You'd like the second commit to be what is currently [32c4449](https://github.com/jj-vcs/jj/commit/32c4449db4011f308444166a91adecadeb875080) and then the first commit to be everything else?
> 
> No, you can add the flag, most of the implementation, test and CHANGELOG entry in a single commit while keeping a second commit ([32c4449](https://github.com/jj-vcs/jj/commit/32c4449db4011f308444166a91adecadeb875080)) with the fixed clippy lints, where Yuya will probably know the solution to whenever he shows up.

Ok I think I've got it all compressed. Let me know if it's not quite what you're looking for 👍

cli/src/commands/evolog.rs
```diff
@@ -151,14 +158,53 @@ pub(crate) fn cmd_evolog(
...
+                    .collect_vec();
+                Ok((c.clone(), edges))
+            })
+            .collect_vec();
+
+        // TODO: This exists to strip out the Result types because ReverseGraphIterator
+        // assumes that the error type of its input iterator is RevsetEvaluationError
+        // when in this case, the input iterator is infallible (E is `()`), or at the
+        // very least a different error type. This is clunky but monkeying around in
+        // lib/src/graph proved to be a bit much for me.
```
yuja, line 178: I don't follow the underlying problem, but `ReverseGraphIterator` isn't really a (lazy) iterator. Maybe we can extract a function that consumes inputs and turns them into `Vec<GraphNode>>`.

JDSeiler: To try and explain where I got stuck more clearly: I made the items in `graph_nodes` be `Result<GraphNode<Commit>, ()>` so that I could pass it to `ReverseGraphIterator::new`. But, because of the `Iterator` implementation for `ReverseGraphIterator`, its `Item` type is `Result<GraphNode<Commit>, RevsetEvaluationError>`.

Since `()` (or whatever other error it might be) doesn't match `RevsetEvaluationError`, I can't have a single value that could be either the normal or reversed iterator (like how I did when `--no-graph` is passed).

I tried typing `graph_nodes` as `Vec<Result<GraphNode<Commit>, RevsetEvaluationError>`, but then I ran into issues where `RevsetEvaluationError` doesn't implement `Clone`.

Alternately, I tried re-tooling the `Iterator` implementation so that `RevsetEvaluationError` isn't hard-coded, but I couldn't figure out how to do it. At least without making more changes than I was comfortable with as a 1st time contributor.

---

Maybe I'll take a fresh look at this code tomorrow / in the next few days and try to clean it up again. If I get stuck, I can come back with more specific blockers.

yuja: > extract a function that consumes inputs and turns them into `Vec<GraphNode>>`
https://github.com/jj-vcs/jj/pull/5172

JDSeiler: This makes sense to me! I'm not sure what my availability will be like going into Christmas but I'll report back with updated changes when I have time.

JDSeiler:
@yuja I've pushed up some new changes based off your refactoring work. I hadn't realized the `reverse_graph` function had been merged, so there are some small changes from where I re-typed it. Happy to back out all/part of those.

cli/src/commands/log.rs
PhilipMetzger, line 1: Nit: Undo the change here.

PhilipMetzger (commented):
Minor nit, I'll let Yuya do the final review.

cli/src/commands/evolog.rs
```diff
@@ -151,14 +159,36 @@ pub(crate) fn cmd_evolog(
...
+            .map(|c| {
+                let edges = c
+                    .predecessors()
+                    .map(|p| GraphEdge::direct(p.unwrap().clone()))
+                    .collect_vec();
+                Ok((c.clone(), edges))
+            })
+            .collect_vec();
+
+        let iter_nodes: Box<dyn Iterator<Item = _>> = if args.reversed {
```
yuja, line 174: Boxed iterator is no longer needed since `reversed_graph()` returns a `Vec`. You can collect `Vec<GraphNode<Commit>>` first, and reverse it as needed.
```rust
if reversed {
    reverse_graph(commit_dag.into_iter().map(Result::<_, Infallible>::Ok)).unwrap()
...
```
(https://doc.rust-lang.org/stable/std/convert/enum.Infallible.html is preferred because the type system ensures that `Err` is never constructed. In recent version of Rust, we can use `let Ok(_) = ..` instead of `unwrap()`.)

cli/src/commands/evolog.rs
```diff
@@ -180,23 +210,28 @@ pub(crate) fn cmd_evolog(
...
         }
     } else {
-        for commit in commits {
-            with_content_format
-                .write(formatter, |formatter| template.format(&commit, formatter))?;
+        let commits_iter: Box<dyn Iterator<Item = _>> = if args.reversed {
+            Box::new(commits.iter().rev())
+        } else {
+            Box::new(commits.iter())
+        };
```
yuja, line 223: nit: same here. just reverse the `Vec`.

cli/src/graphlog.rs
```diff
@@ -31,6 +33,28 @@ pub enum Edge<T> {
...
+            Edge::Missing => Edge::Missing,
+        }
+    }
+}
+
+/// Used to convert a jj_lib GraphEdge into a graphlog Edge.
+pub fn convert_edge_type<N>(e: GraphEdge<N>) -> Edge<N> {
+    // TODO: Is it possible for the ClI to use the GraphEdge type directly?
+    // These two types seem to correspond exactly, only differing in minor
+    // implementation details.
```
yuja, line 50: Probably yes. I think the `graphlog::Edge` type exists just for historical reasons. If you're okay, can you add a refactoring patch first?

lib/src/graph.rs
```diff
@@ -86,15 +87,17 @@ pub fn reverse_graph<N: Clone + Eq + Hash, E>(
...
+    // with their reverse edges.
+    let mut reverse_graph = vec![];
+    while let Some(node) = original_nodes.pop() {
         let edges = reverse_edges.get(&node).cloned().unwrap_or_default();
-        items.push((node, edges));
+        reverse_graph.push((node, edges));
     }
-    Ok(items)
+    Ok(reverse_graph)
 }
```
yuja, line 101: nit: maybe merge error? the change looks identical.

yuja (approved):
Thanks.

cli/src/commands/evolog.rs
JDSeiler: I didn't know about `Infallible`! Thanks for pointing it out. I touched up this code a bit more. I'm not convinced I used `Infallible` correctly, but I got rid of the boxed iterators, so it's at least simpler / less noisy in that way.

cli/src/graphlog.rs
JDSeiler: If it's alright, I'd prefer to do a follow-up PR that just contains the `Edge -> GraphEdge` refactoring? If it's important to you for it to all be in this PR, that's fine. It's just a bit easier for me personally to open a new PR after this one.

cli/src/commands/evolog.rs
```diff
@@ -151,14 +161,35 @@ pub(crate) fn cmd_evolog(
     if !args.no_graph {
         let mut raw_output = formatter.raw()?;
         let mut graph = get_graphlog(graph_style, raw_output.as_mut());
-        for commit in commits {
-            let edges = commit
-                .predecessor_ids()
+
+        let commit_dag: Vec<Result<GraphNode<Commit>, Infallible>> = commits
```
yuja, line 165: We don't need `Result<..>` here, right? It's only required by `reverse_graph()`.

cli/src/commands/evolog.rs
```diff
@@ -151,14 +161,35 @@ pub(crate) fn cmd_evolog(
...
+            .iter()
```
yuja, line 166: nit: `.into_iter()` instead of `.iter()` + `.clone()`

cli/src/commands/evolog.rs
```diff
@@ -151,14 +161,35 @@ pub(crate) fn cmd_evolog(
...
+            .map(|c| {
+                let edges = c
+                    .predecessors()
+                    .map(|p| GraphEdge::direct(p.unwrap().clone()))
```
yuja, line 170: nit: this `.unwrap()` can panic. We can use `.map_ok()` and `.try_collect()` instead.
```rust
                c.predecessors()
                    .map_ok(GraphEdge::direct)
                    .try_collect()
                    .map(|edges| (c, edges)) // or .try_collect()?; Ok((c, edges)) with type annotation
```

Another idea is to construct `Vec<GraphNode<Commit>>` upfront at `topo_order_reverse_ok()`.

cli/src/commands/evolog.rs
```diff
@@ -151,14 +161,35 @@ pub(crate) fn cmd_evolog(
...
+        };
+
+        for node in iter_nodes {
+            let (commit, edges) = node;
+            // Slightly different edge type used for display.
+            let graphlog_edges: Vec<Edge<CommitId>> = edges
                 .iter()
-                .map(|id| Edge::Direct(id.clone()))
-                .collect_vec();
+                .cloned()
```
yuja, line 187: nit: `.into_iter()`

cli/src/graphlog.rs
yuja: No problem. These functions can be easily removed later.

cli/src/commands/evolog.rs
JDSeiler: If this is what you're getting at: Yes, the code in `evolog` doesn't care about the `Result` but it's required by the signature of `reverse_graph`.

From skimming the other uses of `reverse_graph` it seems possible to remove the requirement for a `Result` in the signature but it would require more refactoring changes across several unrelated files. So, perhaps that's another good candidate for a follow-up if we'd like to change that.

cli/src/commands/evolog.rs
JDSeiler: To be honest this particular suggestion has me a bit lost. I tried using `map_ok` and `try_collect` but it ended up pushing the potential `BackendError` up a level, so that `commit_dag` became `Vec<Result<_, BackendError>`, and I'd need to unsafely unwrap it somewhere else (like on line 179).

What's the desired behavior if we fail to fetch predecessors? I'd assume the whole command should fail with an error message, but I'm not sure what the mechanics of making that happen are.

cli/src/commands/evolog.rs
JDSeiler: I changed this instance and the other one you pointed out, thanks!

cli/src/commands/evolog.rs
yuja: Try something liek this:
```rust
        let commit_dag: Vec<GraphNode<Commit>> = commits
            .into_iter()
            .map(|c| -> Result<_, BackendError> {
                let edges = c.predecessors().map_ok(GraphEdge::direct).try_collect()?;
                Ok((c, edges))
            })
            .try_collect()?;

        // XXX this isn't actually an "iter"
        let iter_nodes = if args.reversed {
            reverse_graph(commit_dag.into_iter().map(Result::<_, Infallible>::Ok)).unwrap()
        } else {
            commit_dag
        };
```

Another option:
```rust
    let to_graph_node = |commit: Commit| -> Result<GraphNode<Commit>, BackendError> {
        let edges = commit.predecessors().map_ok(GraphEdge::direct).try_collect()?;
        Ok((commit, edges))
    };
    let commit_dag = topo_order_reverse_ok(
        [to_graph_node(start_commit)],
        |(commit, _)| commit.id().clone(),
        |(_, edges)| {
            // Predecessors don't need to follow any defined order. However in
            // practice, if there are multiple predecessors, then usually the
            // first predecessor is the previous version of the same change, and
            // the other predecessors are commits that were squashed into it. If
            // multiple commits are squashed at once, then they are usually
            // recorded in chronological order. We want to show squashed commits
            // in reverse chronological order, and we also want to show squashed
            // commits before the squash destination (since the destination's
            // subgraph may contain earlier squashed commits as well), so we
            // visit the predecessors in reverse order.
            edges.iter().rev().map(|edge| to_graph_node(edge.target.clone())).collect_vec()
        },
    )?;
```

JDSeiler:
@yuja I think everything is resolved now

cli/src/commands/evolog.rs
```diff
@@ -151,14 +162,31 @@ pub(crate) fn cmd_evolog(
...
+
+        let commit_dag: Vec<GraphNode<Commit>> = commits
+            .into_iter()
+            .map(|c| -> Result<_, BackendError> {
+                let edges = c.predecessors().map_ok(GraphEdge::direct).try_collect()?;
+                Ok((c, edges))
+            })
+            .try_collect()?;
+
+        let iter_nodes = if args.reversed {
```
yuja, line 174: nit: can you rename this to `commit_dag` or something? It's not an iterator.

cli/src/commands/evolog.rs
```diff
@@ -151,14 +162,31 @@ pub(crate) fn cmd_evolog(
...
+            reverse_graph(commit_dag.into_iter().map(Result::<_, Infallible>::Ok)).unwrap()
+        } else {
+            commit_dag
+        };
+
+        for node in iter_nodes {
+            let (commit, edges) = node;
```
yuja, line 181: nit: you can inline this as `for (commit, edges) in ..`

yuja (approved):
LG, thanks!

cli/src/commands/evolog.rs
JDSeiler: Maybe this nit and the next I can wrap in as miscellaneous refactors if/when I take a second look at `GraphEdge`.

PR #5171, opened 2024-12-22
merge_tools: add `"$marker_length"` variable

scott2000:
Discussed previously here: https://github.com/jj-vcs/jj/pull/4969#discussion_r1859088488

Git supports passing the conflict marker length to merge drivers using "%L". It would be useful if we also had a way to pass the marker length to merge tools, since it would allow Git merge drivers to be used with `jj resolve` in more cases. Without this variable, any merge tool that parses or generates conflict markers could fail on files which require conflict markers longer than 7 characters.

https://git-scm.com/docs/gitattributes#_defining_a_custom_merge_driver

# Checklist

If applicable:


yuja (approved):
LG, thanks.

PR #5181, opened 2024-12-24
conflicts: allow missing newlines at end of file (based on `[noeol]` marker)

### LINKED ISSUE #3968: Conflicts at the end of files that don't end in a newline are not materialized correctly
```
jj new
echo -n "aa" > qq
jj new
echo -n "bb" > qq
jj new @-
echo -n "cc" > qq
jj new all:@-+
cat qq
```

## Actual Result

```
<<<<<<< Conflict 1 of 1
%%%%%%% Changes from base to side #1
-aa+cc+++++++ Contents of side #2
bb>>>>>>> Conflict 1 of 1 ends
```

If the file is subsequently edited, `jj` cannot parse the conflict correctly.

## Expected Behavior

I'd materialize the conflict as though the files ended in newlines, if we decide this is acceptable. We could perhaps notify the user about this in the labels:

```
<<<<<<< Conflict 1 of 1 (extra newlines inserted)
%%%%%%% Changes from base to side #1
-aa
+cc
+++++++ Contents of side #2
bb
>>>>>>> Conflict 1 of 1 ends
```

If this is not acceptable, I'm not sure what is best to do.

scott2000:
Fixes #3968 (this is an alternative to #4088). This is a rough draft of one possible solution. It's similar to what @yuja proposed with adding the "noeol" indicator to the comment part of the conflict marker, but I only added it to the `<<<<<<<` start marker. I think this makes the materialization and parsing easier since we don't need to handle adding markers for both the removes and adds in `%%%%%%%` diff sections, but I'm open to comments and other approaches.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,6 +141,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `[diff.<format>]` configuration now applies to `.diff().<format>()` commit
   template methods.
 
+* Conflicts at the end of files which don't end with a newline character are
+  now materialized in a way that can be parsed correctly.
+  [#3968](https://github.com/jj-vcs/jj/issues/3968)
+
 ## [0.25.0] - 2025-01-01
 
 ### Release highlights
--- a/docs/conflicts.md
+++ b/docs/conflicts.md
@@ -184,3 +184,33 @@ New Heading
 ===========
 >>>>>>>>>>>>>>> Conflict 1 of 1 ends
 ```
+
+## Conflict with no newline at end of file
+
+On POSIX systems, text files are generally assumed to be a series of lines, with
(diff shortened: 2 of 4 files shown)


yuja:
Another option is to reconstruct the missing eol state from the original tree. Something like, if the last parsed hunk is a conflict, check if the corresponding file content had newline at end, if not, strip newline from the parsed content.

scott2000:
> Another option is to reconstruct the missing eol state from the original tree. Something like, if the last parsed hunk is a conflict, check if the corresponding file content had newline at end, if not, strip newline from the parsed content.

Yeah, that could also work. I made a separate PR with a possible implementation of that here: #5186. It's definitely a lot simpler, but it does have the downside that it's no longer possible to compare the changes to the EOL state between the sides by looking at the materialized file.

PR #5186, opened 2024-12-25
conflicts: allow missing newlines at end of file (based on previous content)

### LINKED ISSUE #3968: Conflicts at the end of files that don't end in a newline are not materialized correctly

scott2000:
Fixes #3968. Alternative to #5181 and #4088.

# Checklist

If applicable:

### DIFF
--- a/lib/src/conflicts.rs
+++ b/lib/src/conflicts.rs
@@ -65,17 +65,17 @@ fn write_diff_hunks(hunks: &[DiffHunk], file: &mut dyn Write) -> io::Result<()> 
                 debug_assert!(hunk.contents.iter().all_equal());
                 for line in hunk.contents[0].lines_with_terminator() {
                     file.write_all(b" ")?;
-                    file.write_all(line)?;
+                    write_and_ensure_newline(file, line)?;
                 }
             }
             DiffHunkKind::Different => {
                 for line in hunk.contents[0].lines_with_terminator() {
                     file.write_all(b"-")?;
-                    file.write_all(line)?;
+                    write_and_ensure_newline(file, line)?;
                 }
                 for line in hunk.contents[1].lines_with_terminator() {
                     file.write_all(b"+")?;
-                    file.write_all(line)?;
+                    write_and_ensure_newline(file, line)?;
                 }
             }
         }
@@ -462,14 +462,16 @@ fn materialize_git_style_conflict(
         conflict_marker_len,
         &format!("Side #1 ({conflict_info})"),
     )?;
-    output.write_all(left)?;
+    write_and_ensure_newline(output, left)?;
+
     write_conflict_marker(
         output,
         ConflictMarkerLineChar::GitAncestor,
         conflict_marker_len,
         "Base",
     )?;
-    output.write_all(base)?;
+    write_and_ensure_newline(output, base)?;
+
     // VS Code doesn't seem to support any trailing text on the separator line
     write_conflict_marker(
         output,
@@ -477,7 +479,8 @@ fn materialize_git_style_conflict(
         conflict_marker_len,
         "",
     )?;
-    output.write_all(right)?;
+
+    write_and_ensure_newline(output, right)?;
     write_conflict_marker(
         output,
         ConflictMarkerLineChar::ConflictEnd,
@@ -503,7 +506,7 @@ fn materialize_jj_style_conflict(
             conflict_marker_len,
             &format!("Contents of side #{}", add_index + 1),
         )?;
-        output.write_all(data)
+        write_and_ensure_newline(output, data)
     };
 
     // Write a negative snapshot (base) of a conflict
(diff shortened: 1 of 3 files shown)


scott2000:
One possible issue with this approach is that some conflicts become unreadable. For instance, a conflict with a base of `base`, a left side of `left`, and a right side of `base\n` would be rendered as:

```
<<<<<<< Conflict 1 of 1
+++++++ Contents of side #1
left
%%%%%%% Changes from base to side #2
 base
>>>>>>> Conflict 1 of 1 ends
```

yuja:
> One possible issue with this approach is that some conflicts become unreadable. For instance, a conflict with a base of `base`, a left side of `left`, and a right side of `base\n`

I think `base` and `base\n` can still be rendered as different contents (though these two are visually identical.) Maybe we can add some comment to the marker description?

scott2000:
Ok, I updated it so that it would render as this:

```
<<<<<<< Conflict 1 of 1
+++++++ Contents of side #1 [noeol]
left
%%%%%%% Changes from base to side #2 [-noeol]
 base
>>>>>>> Conflict 1 of 1 ends
```

lib/src/conflicts.rs
```diff
@@ -460,14 +482,17 @@ fn materialize_git_style_conflict(
         output,
         ConflictMarkerLineChar::ConflictStart,
         conflict_marker_len,
-        &format!("Side #1 ({conflict_info})"),
+        &format!(
+            "Side #1 ({conflict_info}){}",
+            no_eol_marker(no_eol.get_add(0))
+        ),
     )?;
     output.write_all(left)?;
```
yuja, line 490: I think missing `\n` should be padded here. (applies to the other materialization styles and sides)

It's the matter of rendering.

scott2000:
I'll mark this as ready for review since there hasn't been any more discussion, and I think this is the simplest solution.

lib/tests/test_conflicts.rs
```diff
@@ -2007,6 +2024,214 @@ fn test_update_conflict_from_content_with_long_markers() {
...
+#[test]
+fn test_update_conflict_from_content_no_eol() {
+    let test_repo = TestRepo::init();
+    let store = test_repo.repo.store();
+
+    let path = RepoPath::from_internal_string("file");
+    let base_id = testutils::write_file(store, path, "line 1\nline 2\nline 3\nbase");
+    let left_empty_id =
+        testutils::write_file(store, path, "line 1\nline 2 - left\nline 3\nbase\nleft\n");
+    let right_id = testutils::write_file(store, path, "line 1\nline 2 - right\nline 3\nright");
```
yuja, line 2036: nit: better to not use ` - ` as separator? The diff output looks like newline char is missing.

lib/src/conflicts.rs
```diff
@@ -58,6 +58,15 @@ pub const MIN_CONFLICT_MARKER_LEN: usize = 7;
...
 const CONFLICT_MARKER_LEN_INCREMENT: usize = 4;
 
+/// Marker for missing terminating newline in a term of a conflict.
+const NO_EOL_MARKER: &str = " [noeol]";
+
+/// Marker for missing terminating newline in the "add" side of a diff.
+const ADD_NO_EOL_MARKER: &str = " [+noeol]";
+
+/// Marker for missing terminating newline in the "remove" side of a diff.
+const REMOVE_NO_EOL_MARKER: &str = " [-noeol]";
```
yuja, line 68: nit: since these strings are not meant to be parsed, I wouldn't call them as `MARKER`. Maybe `COMMENT` or `LABEL` is better?

yuja (approved):
Looks good to me, thanks!

lib/tests/test_conflicts.rs
yuja, line 1: Can you also add a test in which only the last newline char differs? If we did some resolution before fixing up newline chars, it would fail.

scott2000: I added a test for this, but I had to make the base different from both of the sides because otherwise the conflict would be trivially resolvable even with the difference in newline characters (e.g. `["a\n", "a\n", "a"]` trivially simplifies to `["a"]`). Does the test I added work for testing the case you were thinking of?

Issue #3979, opened 2024-06-27
Include nonconflicted files in root tree of file-conflict Git commits


amiryal:
With the aim of better compatibility with tools that look for the state of working-copy files in the colocated Git’s `HEAD`, I propose to **include all the would-be working copy files in the commit**, _in addition_ to the `.jj-conflict-*` trees that are already there for preventing GC.

This came up in a discussion about Nix flakes, and the way it breaks when `HEAD@git` has file conflicts. It was pointed out that:
> Nix flakes actually use file contents from the working copy, but the index to determine which files to *include*.
> 
> _Originally posted by @emilazy in https://github.com/martinvonz/jj/discussions/3970#discussioncomment-9878676_

In other parts of the discussion, there has been reluctance towards updating Git’s index with the would-be working tree of `HEAD@git` (that is the parent of the jj working copy). When I suggested to instead change the way file-conflict commits are prepared in Git, it was pointed out that the barrier to do that is not high:
> It used to be more friendly to colocated repos until commit 006c764694a2. I don't think there's much harm in changing it to better match the files in the working copy, as long as we *also* include the `.jj-conflict-*` subdirectories (they're needed in order to prevent GC).
> 
> _Originally posted by @martinvonz in https://github.com/martinvonz/jj/discussions/3970#discussioncomment-9887040_

yuja:
If this is the issue of Git index, maybe we can add materialized file blobs to the index?

`git::reset_head()` currently resets Git index to the contents of the raw Git commit of the first parent. This wouldn't make sense if the working-copy commit was a merge or the parent had conflicts.

Of course, `git commit` shouldn't be used in that situation, which is probably the same for `.jj-conflict-*` trees.

martinvonz:
Good point! I think you're right that the issue is only with the Git index, but maybe @amiryal can confirm.

I know we have a similar problem when the user uses `git switch` to update to a conflicted commit, but hopefully that doesn't confuse tooling in the same way.

joyously:
Are the Git special cases scattered in the code, or in one place?

yuja:
> but don't remove the .jj-conflict dirs.

Hmm, it might be also good to leave a `README` (with scalier name), or some fake conflict marker if `git commit` has a sanity check for that.

> Also, any such change makes it harder to understand how jj affects the staging area now that the answer is not "never".

Just to be clear, `jj new` resets Git index, and my idea is to make it be consistent with the working directory contents materialized by `jj new`.

[...]

Issue #3786, opened 2024-05-30
Bug: Git index has all-zero entries when updating HEAD@git


mlcui-corp:
## Description

When updating `HEAD@git`, such as by `jj new`ing a change that has a different tree than `@-`, the changed entries in the Git index get set to all-zeroes.

This affects shell prompts which use [romkatv/gitstatus](https://github.com/romkatv/gitstatus), such as [romkatv/powerlevel10k](https://github.com/romkatv/powerlevel10k).

I have confirmed this also occurred before a075a5c6cabd493759fc973d99d3fd88feb74735.

## Steps to Reproduce the Problem

1. Create a new empty working copy change in a colocated repository.
1. `jj new` to another change, which changes files.
1. Observe the Git index for the changed files has its cache entry set to an all-zero entry.

<details>

<summary>Example</summary>

```
~/git/jj @3090adfd ❯ jj log -r "main::wz | @"
@  lpom mlcui@google.com 2024-05-30 10:29:18 75ad
│  (empty) (no description set)
│ ○  wzzo mlcui@google.com 2024-05-30 10:02:28 10da
├─╯  test index update
◆  sqww martinvonz@google.com 2024-05-29 23:54:30 main* HEAD@git 3090
│  repo: consider empty and undescribed merge commits as discardable
~

~/git/jj @3090adfd ❯ jj show -s wz
Commit ID: 10da93209b109ca8789a5e5af511a6fad635ec10
Change ID: wzzoqkqrstsxxxztxupwyquwntktqsuo
Author: mlcui <mlcui@google.com> (2024-05-29 17:10:32)
Committer: mlcui <mlcui@google.com> (2024-05-30 10:02:28)

    test index update

M cli/src/cli_util.rs
M lib/src/git.rs

~/git/jj @3090adfd ❯ git status
HEAD detached at 3090adfd5
nothing to commit, working tree clean

~/git/jj @3090adfd ❯ jj new wz
Working copy now at: oolo e29f (empty) (no description set)
Parent commit      : wzzo 10da test index update
Added 0 files, modified 2 files, removed 0 files

~/git/jj @10da9320 !2 ❯ git ls-files --debug > /tmp/status1
~/git/jj @10da9320 !2 ❯ git update-index --refresh
~/git/jj @10da9320 ❯ git ls-files --debug > /tmp/status2
~/git/jj @10da9320 ❯ git diff --no-index /tmp/status1 /tmp/status2 | cat      
[...]

Issue #3754, opened 2024-05-25
concurrent `git::reset_head()` can rewind new `HEAD` exported by other


yuja:
## Description

https://github.com/martinvonz/jj/pull/3749#issuecomment-2130404116

Repo: https://ilya.gr/share/jj-bug-2024-05-24.tar.gz (50MB)

```
E    (@: wc1 -> wc0, HEAD: wc0)
|
D    (@: wc1, HEAD: wc0) "--ignore-working-copy"
|\
| C  (@: wc0, HEAD: wc1 -> wc0)
| |
B |  (@: wc0 -> wc1, HEAD: wc0 -> wc1) "squash"
|/
A    (@: wc0, HEAD: wc0)
```

1. The operation `B` squashes wc commit, and exports new `HEAD`.
2. Concurrently, the operation `C` observes uncommitted refs exported by `B` (but not the `HEAD` change), imports them, and commits the operation. It triggers `git::reset_head()` -> `git_repo.reset(..)`, and overwrites `HEAD` back to `wc0`.
3. These operations are merged at `D`.
4. At `E`, old `HEAD` overwritten by `C` is imported as the "new" `HEAD`.

## Expected Behavior

Perhaps, `HEAD` shouldn't be overwritten at `C`, or it should be re-exported at `D`?

I think the problem can be mitigated by:
a. not exporting unchanged `HEAD`
b. use `PreviousValue::MustExistAndMatch(..)` on exporting `HEAD` (but we'll need to deal with symbolic ref)

## Specifications

- Version: somewhere around 4478055e

PR #5204, opened 2024-12-28
git: use merged parent tree for git index

scott2000:
I had some free time over the holidays, so I thought I'd take a shot at #3979. I think this matches what was agreed upon as the best "step 1" in the issue discussion.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * The deprecated `--siblings` options for `jj split` has been removed.
   `jj split --parallel` can be used instead.
 
+* In colocated repos, the Git index now contains the changes from all parents
+  of the working copy instead of just the first parent (`HEAD`). 2-sided
+  conflicts from the merged parents are now added to the Git index as conflicts
+  as well.
+
 ### Deprecations
 
 ### New features
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -25,13 +25,13 @@ use std::num::NonZeroU32;
 use std::path::PathBuf;
 use std::str;
 
-use git2::Oid;
 use itertools::Itertools;
 use tempfile::NamedTempFile;
 use thiserror::Error;
 
 use crate::backend::BackendError;
 use crate::backend::CommitId;
+use crate::backend::TreeId;
 use crate::commit::Commit;
 use crate::git_backend::GitBackend;
 use crate::index::Index;
@@ -592,6 +592,8 @@ pub enum GitExportError {
     InternalGitError(#[source] Box<dyn std::error::Error + Send + Sync>),
     #[error("The repo is not backed by a Git repo")]
     UnexpectedBackend,
+    #[error(transparent)]
+    Backend(#[from] BackendError),
 }
 
 impl GitExportError {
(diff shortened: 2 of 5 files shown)


yuja:
Thanks for attacking this. I haven't reviewed the code, but we need to be aware that our tests don't cover most of Git index (or staging area) changes. For example, my last attempt to migrate `reset_head()` to `gix` had timestamp issue.
https://github.com/jj-vcs/jj/compare/main...yuja:jj:push-vpsmwmxpokyx

See https://github.com/jj-vcs/jj/issues/3786 for the timestamp issue.

We'll probably need to 1. add basic tests for Git index, 2. reimplement index updates by using gix, 3. switch to merged parent.

scott2000:
> Thanks for attacking this. I haven't reviewed the code, but we need to be aware that our tests don't cover most of Git index (or staging area) changes. For example, my last attempt to migrate `reset_head()` to `gix` had timestamp issue. [main...yuja:jj:push-vpsmwmxpokyx](https://github.com/jj-vcs/jj/compare/main...yuja:jj:push-vpsmwmxpokyx)
>
> See #3786 for the timestamp issue.

Ahh I hadn't seen that. It seems more complicated than I had thought.

> We'll probably need to 1. add basic tests for Git index, 2. reimplement index updates by using gix, 3. switch to merged parent.

Makes sense to me.

scott2000:
Ok I added tests for the current behavior with git2 and added a commit that migrates to gix without changing the behavior before the commit that changes it to use the merged parent tree.

This doesn't solve #3786, since we still use a "0:0" timestamp for newly added files, but I think it at least doesn't make it any worse than the old implementation (it still preserves timestamps for files which already had timestamps in the index).

I think it's a slight improvement actually, since previously we would lose all stat information when checking out a commit with conflicts (since the index would only contain the `.jjconflict` files), but now it will at least be preserved for files which aren't conflicted.

CHANGELOG.md
> 1. If the working copy commit contains conflicted files which weren't
>   conflicted in the merged parent tree, then those files won't appear
>   as conflicts in the index (since the index is based on the parents).

In this case, I think user would have to do `git add` to commit the conflict by `git`. Is there any safety check in `git` that prevents accidental `git add`? (I'm just asking because I'm not familiar with the git behavior.)

yuja (commented):
Looks generally good, and the added tests are nice, thanks.

lib/src/git.rs
```diff
@@ -1080,20 +1056,37 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
...
+        let entry = entry?.simplify();
+        if let [left, base, right] = entry.as_slice() {
+            // 2-sided conflicts can be represented in the Git index
+            push_index_entry(&path, left, Stage::Ours);
+            push_index_entry(&path, base, Stage::Base);
+            push_index_entry(&path, right, Stage::Theirs);
+        } else {
+            // We can't represent many-sided conflicts in the Git index, so just add the
+            // first side as staged even if there are conflicts.
+            push_index_entry(&path, entry.first(), Stage::Unconflicted);
```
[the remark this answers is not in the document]
scott2000: Some IDEs (e.g. IntelliJ IDEA) use the index for their conflict resolution interfaces, and they ignore the materialized markers entirely. So if we put the first 3 as the conflict, then the user might accidentally resolve it using IntelliJ, which would lead to an incorrect resolution. So I think it's safer that we just don't add it as a conflict at all, since the user might think it's safe to use Git-based tools to resolve it if we do.

CHANGELOG.md
scott2000: Yeah, that's correct; I think the safety check is only for `git commit`. Even though there's no risk of the user committing the conflict markers in that case without running `git add`, I still think we should add the dummy conflict. Mainly because if the user runs `git commit`, it will cause the unstaged files with conflict markers to be carried over to their new working copy, but the conflicted status won't carry over to the new working copy on the `jj` side, so they would end up with `@` having conflict markers without being conflicted.

yuja: > Even though there's no risk of the user committing the conflict markers in that case without running `git add`, I still think we should add the dummy conflict. Mainly because if the user runs `git commit`, it will cause the unstaged files with conflict markers to be carried over to their new working copy, but the conflicted status won't carry over to the new working copy on the `jj` side,

True, but I don't think it's worth the complexity we have to scan both wc tree and parent tree. Maybe there would be other edge cases which can't be handled (something like user created new commit by `git commit`, then made conflicting state by using `git`, then ran `jj` which imports new wc-parent commit and conflicting wc)?

(BTW, if a committed tree had a file named `INDEX_DUMMY_CONFLICT_FILE`, maybe it should be indexed as a plain file.)

scott2000: Hmm yeah that's true. I'll change it.

lib/src/git.rs
```diff
@@ -1080,20 +1056,37 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
...
             mode,
             path,
         );
+    };
+
+    // Use the merged parent tree as the Git index, allowing `git diff` to show the
+    // same changes as `jj diff`. If the merged parent tree has 2-sided conflicts,
+    // then the Git index will also be conflicted.
+    for (path, entry) in wc_commit.parent_tree(mut_repo)?.entries() {
+        let entry = entry?.simplify();
```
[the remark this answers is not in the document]
scott2000: > nit: it might be better to handle resolved case early. I'm not sure if it has measurable performance difference, though.

I'm not sure I understand what you mean by handling the resolved case early. Do you mean to remove the `.simplify()` somehow?

> Can you also check if the gix version doesn't regress the fast path [a075a5c](https://github.com/jj-vcs/jj/commit/a075a5c6cabd493759fc973d99d3fd88feb74735) ? (I assume it wouldn't, but just in case.)

Hmm it does look like it has a small regression when the tree didn't change (from ~580ms to ~615ms for `jj new` on an empty commit in nixpkgs on my machine). I did some profiling, and it looks like `TreeEntriesIterator::next` is 24% of the runtime of the command (75% of which is just reading the trees from the Git repo), `push_index_entry` is 2.6%, and writing the index to disk is 2.8%, leading to a total of 30% of the runtime of the command being spent in `reset_head`.

I'm not sure if it would be possible to reduce it by much, since 80% of the time in `reset_head` comes from iterating over the tree entries rather than anything to do with the index, and it seems like we'd have to iterate over the entries no matter what. Do you have any ideas? Or does a 6% regression seem acceptable for this case?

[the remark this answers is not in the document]
yuja: > > nit: it might be better to handle resolved case early. I'm not sure if it has measurable performance difference, though.
> 
> I'm not sure I understand what you mean by handling the resolved case early. Do you mean to remove the `.simplify()` somehow?

Yes. I meant testing `.as_resolved()`/`.into_resolved()` first. It's probably better to handle resolved case differently from n-way conflicts (n > 3) anyway, for explicitness. Oh, btw, what would happen if n-way conflict were labeled as Ours (without Base and Theirs)? Does IntelliJ resolve it by taking the Ours?

> > Can you also check if the gix version doesn't regress the fast path [a075a5c](https://github.com/jj-vcs/jj/commit/a075a5c6cabd493759fc973d99d3fd88feb74735) ? (I assume it wouldn't, but just in case.)
> 
> Hmm it does look like it has a small regression when the tree didn't change (from ~580ms to ~615ms for `jj new` on an empty commit in nixpkgs on my machine).

Suppose nixpkgs has unusual structure (many directories compared to files), I think that's okay.

> For the non-fast-path case, it seems to have around the same performance as trunk for me.
> 
> I'm not sure if it would be possible to reduce it by much, since 80% of the time in `reset_head` comes from iterating over the tree entries rather than anything to do with the index, and it seems like we'd have to iterate over the entries no matter what. Do you have any ideas?

Maybe we could use gix API if the parent commit isn't a merge? I think gix one is faster compared to jj's native API (because of fewer abstraction layers.)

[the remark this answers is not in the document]
scott2000: I think I got it working with the `gix::Repository::index_from_tree` function when there's no conflict by copying the stat information afterwards. It seems to bring it down to about the same latency as the current trunk `jj` build for the fast path.

lib/tests/test_git.rs
```diff
@@ -2184,21 +2196,184 @@ fn test_reset_head_with_index() {
...
+
+    let mut tx = repo.start_transaction();
+    let mut_repo = tx.repo_mut();
+
+    let file_contents = repo
+        .store()
+        .write_file(RepoPath::root(), &mut b"file".as_slice())
+        .block_on()
+        .unwrap();
+
```
yuja, line 2237: nit: maybe we can use `testutils::write_*()` helpers

lib/src/git.rs
```diff
@@ -987,73 +989,94 @@ fn update_git_head(
...
-            false
-        };
-        if !skip_reset {
-            git_repo.reset(new_git_commit.as_object(), git2::ResetType::Mixed, None)?;
-        }
+    // If the first parent of the working copy has changed, reset the Git HEAD.
+    if mut_repo.git_head() != first_parent {
+        update_git_head(
+            &git_repo,
+            gix::refs::transaction::PreviousValue::Any,
```
yuja, line 1013: nit: can you leave a TODO comment about #3754? Ideally we should ensure that the existing HEAD ref points to the known object id (but we have to somehow deal with non-detached state.)

fwiw, maybe we can use `PreviousValue::MustExist` for a little bit of safety.

lib/src/git.rs
```diff
@@ -987,73 +989,94 @@ fn update_git_head(
...
+                .tree_id()
+                .map_err(GitExportError::from_git)?
+                .as_bytes()
+                .to_owned(),
+        )
+    };
+
+    let mut index = if &parent_tree_id == mut_repo.store().empty_tree_id() {
+        // If the tree is empty, gix fails to load the object, so we just use an empty
+        // index directly.
```
yuja: nit: "gix fails" -> "gix can fail"?

It might be our fault that we don't write an empty tree object to the store even if we create a real Git commit with an empty tree.

lib/src/git.rs
```diff
@@ -1022,38 +1026,28 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
...
-    let parent_tree_id = if first_parent_id == mut_repo.store().root_commit_id() {
-        mut_repo.store().empty_tree_id().clone()
-    } else {
-        TreeId::new(
+    let parent_tree = wc_commit.parent_tree(mut_repo)?;
+
+    // Use the merged parent tree as the Git index, allowing `git diff` to show the
+    // same changes as `jj diff`. If the merged parent tree has 2-sided conflicts,
+    // then the Git index will also be conflicted.
+    let mut index = if let Some(tree) = parent_tree.as_merge().as_resolved() {
```
yuja, line 1037: nit: could you add lib test that covers both resolved and merged cases?

lib/src/git.rs
```diff
@@ -1080,6 +1074,106 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
...
+                    (id.as_bytes(), gix::index::entry::Mode::FILE)
+                }
+            }
+            TreeValue::Symlink(id) => (id.as_bytes(), gix::index::entry::Mode::SYMLINK),
+            TreeValue::Tree(tree_id) => {
+                for (path, entry) in store.get_tree(path.to_owned(), tree_id)?.entries() {
+                    push_index_entry(index, store, &path, &Some(entry), stage)?;
+                }
+                return Ok(());
+            }
```
yuja, line 1110: This seems odd because `.entries()` is a recursive iterator.

I think it handles the case where tree conflicts with non-tree object. In that case, we would probably treat the path as non-tree. I don't remember what would be materialized in the working copy, but I don't think it's correct to recursively insert Git index entries.

yuja (approved):
LG, thanks!

lib/src/git.rs
scott2000: Yeah, I added this to handle the case of conflicts between tree and non-tree. When we materialize it, it looks like this (for conflict between "test" and "test/file"):

```
Conflict:
  Adding file with id f73f3093ff865c514c6c51f867e35f693487d0d3
  Adding tree with id c3e9a16324d8ebd732b8a6aa79fa6881238c4ea6
```

Currently, I add all of the files in the tree to the index (which is why it needs to be recursive), and it produces this `git status` output:

```
Unmerged paths:
  (use "git restore --staged <file>..." to unstage)
  (use "git add <file>..." to mark resolution)
        added by us:     test
        added by them:   test/file
```

But it looks like Git does something different in this case; `git merge` keeps the tree and renames the file to something else so that it doesn't conflict with the name of the tree, so in Git the merge would produce something like:

```
You have unmerged paths.
  (fix conflicts and run "git commit")
  (use "git merge --abort" to abort the merge)

Changes to be committed:
        deleted:    test
        new file:   test/file

Unmerged paths:
  (use "git add <file>..." to mark resolution)
        added by us:     test~HEAD
```

Since this is an uncommon case, maybe we could just skip adding the tree to the index if this case happens? Then it would look like this:

```
Unmerged paths:
  (use "git restore --staged <file>..." to unstage)
  (use "git add <file>..." to mark resolution)
        added by us:     test
```

What do you think?

martinvonz: > When we materialize it, it looks like this (for conflict between "test" and "test/file"):

That gets written as a file called `test`, right? If we write a file like that to the working copy, then I think we should also just write a single matching entry to the git index. Does that make sense?

scott2000: > That gets written as a file called `test`, right?

Yes, that's correct.

> If we write a file like that to the working copy, then I think we should also just write a single matching entry to the git index. Does that make sense?

That makes sense to me. I think we can just skip adding the tree then, so only `test` would be added to the index.

Issue #5085, opened 2024-12-12
FR: Allow specifying `tail` for `truncate_*` template functions


zx8:
**Describe the solution you'd like**

This is a very minor nice-to-have, but would it be possible to add an optional `tail` to the `truncate_*` template functions, to indicate that truncation has occurred? Sometimes truncation occurs at a space character, and it's not obvious that there was a longer description available.

**Describe alternatives you've considered**

Currently, it's possible to do something like:

```
if(
  description.first_line().substr(0, 80).starts_with(description.first_line()),
  pad_end(80, description.first_line()),
  truncate_end(79, description.first_line()) ++ "…"
)
```

_(After https://github.com/martinvonz/jj/pull/5072 is released, I can replace this workaround with a length check)_

yuja:
fwiw, I would call the parameter as `ellipsis` or something, than `fill_char`. It shouldn't be repeated to fill the remaining space.

PR #5214, opened 2024-12-30
templates: add `tail` argument to truncate_*

bryceberger:
#5085 

The way this is implemented, the tail length is *not* taken into account when finding the truncation position. That means that `truncate_end(2, "abcdef", tail="...")` will result in `"ab..."`, a 5 character string.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+* `truncate_start` and `truncate_end` now accept an optional `tail` named
+  argument, to indicate when truncation has occurred.
+
 ### Release highlights
 
 ### Breaking changes
--- a/cli/src/template_builder.rs
+++ b/cli/src/template_builder.rs
@@ -1318,22 +1318,32 @@ fn builtin_functions<'a, L: TemplateLanguage<'a> + ?Sized>() -> TemplateBuildFun
     map.insert(
         "truncate_start",
         |language, diagnostics, build_ctx, function| {
-            let [width_node, content_node] = function.expect_exact_arguments()?;
+            let ([width_node, content_node], [tail_node]) =
+                function.expect_named_arguments(&["", "", "tail"])?;
             let width = expect_usize_expression(language, diagnostics, build_ctx, width_node)?;
             let content =
                 expect_template_expression(language, diagnostics, build_ctx, content_node)?;
-            let template = new_truncate_template(content, width, text_util::write_truncated_start);
+            let tail = tail_node
+                .map(|node| expect_plain_text_expression(language, diagnostics, build_ctx, node))
+                .transpose()?;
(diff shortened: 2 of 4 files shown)


yuja:
> The way this is implemented, the tail length is _not_ taken into account when finding the truncation position. That means that `truncate_end(2, "abcdef", tail="...")` will result in `"ab..."`, a 5 character string.

I don't think that's desired behavior because user would want to align columns. Can you check if we can copy the logic of `elide_start()`/`elide_end()`? We'll need a formatter version of these functions. (BTW, in order to just append ellipsis, we wouldn't need native support for `tail`. Something like `truncate(...) ++ if(width(...) > 2, "...")` should work.)

FWIW, I'm not sure if `tail` is a good name. It might be okay for `truncate_end()`, but it's definitely wrong name for `truncate_start()` (where the head is truncated.)

bryceberger:
> > ... length of tail not accounted for
> I don't think that's desired behavior because user would want to align columns. Can you check if we can copy the logic of `elide_start()`/`elide_end()`? We'll need a formatter version of these functions.
Yeah, can take a look into that. Did it this way because it looked like less code to touch. I think `elide_*` is only used for some non-user-contrallable text, unsure what we should do if the user inputs a longer indicator than the truncate length.

> (BTW, in order to just append ellipsis, we wouldn't need native support for `tail`. Something like `truncate(...) ++ if(width(...) > 2, "...")` should work.)
Hmm, so is this wanted at all? Doesn't require repeating the item to be truncated, but otherwise that looks like a simpler solution.

> FWIW, I'm not sure if `tail` is a good name. It might be okay for `truncate_end()`, but it's definitely wrong name for `truncate_start()` (where the head is truncated.)
If we want to keep the same name for both, maybe `indicator`? Otherwise could do `head` and `tail`.

yuja:
> Yeah, can take a look into that. Did it this way because it looked like less code to touch. I think `elide_*` is only used for some non-user-contrallable text, unsure what we should do if the user inputs a longer indicator than the truncate length.

iirc, it trims ellipsis/indicator to the maximum width, but it wouldn't be super important.

> > (BTW, in order to just append ellipsis, we wouldn't need native support for `tail`. Something like `truncate(...) ++ if(width(...) > 2, "...")` should work.)
> 
> Hmm, so is this wanted at all? Doesn't require repeating the item to be truncated, but otherwise that looks like a simpler solution.

I would say we'll want this feature, but not in the way it is implemented in this PR.

> > FWIW, I'm not sure if `tail` is a good name. It might be okay for `truncate_end()`, but it's definitely wrong name for `truncate_start()` (where the head is truncated.)
> 
> If we want to keep the same name for both, maybe `indicator`? Otherwise could do `head` and `tail`.

I tend to use `ellipsis` (which was the name used in Qt), but maybe I'm biased. I'm not native either.

bryceberger:
https://github.com/jj-vcs/jj/pull/5584

PR #5221, opened 2025-01-01
nix: use dynamic completions in installPhase

bryceberger:
Have been used in nixpkgs since the update to 0.24. This makes dynamic completions the default for those who use jj directly from the jj repo as well.

# Checklist

If applicable:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -127,9 +127,9 @@
             installManPage ./jj.1
 
             installShellCompletion --cmd jj \
-              --bash <($out/bin/jj util completion bash) \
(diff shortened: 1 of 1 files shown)


bnjmnt4n (approved):

bnjmnt4n:
Just wanted to point out that I'm fine with this since I've been using this daily without much issues, but the dynamic completions are in fact still labelled "experimental". People using the flake are probably fine with using an experimental feature (given that they are living on the unreleased trunk), but I'm wondering if it was a little premature for the nixpkgs derivation to switch to the dynamic completions.

PR #5236, opened 2025-01-02
Add a paragraph on how to install jj completions into Powershell.

keltia:
# Checklist

If applicable:

### DIFF
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -276,3 +276,14 @@ source-bash $(jj util completion)
 ```
 
 (dynamic completions not available yet)
+
+### Powershell
+
+Insert this line in your $PROFILE file:
+(usually $HOME\Documents\PowerShell\Microsoft.PowerShell_profile.ps1)
+
+```shell
+Invoke-Expression (& { (jj util completion power-shell | Out-String) })
+```
+
+(dynamic completions not available yet)


docs/install-and-setup.md
```diff
@@ -276,3 +276,14 @@ source-bash $(jj util completion)
 ```
 
 (dynamic completions not available yet)
+
+### Powershell
+
+Insert this line in your $PROFILE file:
+(usually C:\Users\USER\Documents\PowerShell\Microsoft.PowerShell_profile.ps1)
```
martinvonz, line 283: nit: would it be clearer to replace "USER" by "$USER" or "<users>"? (i definitely think it would for linux users, but i'm not sure about windows users)

martinvonz: (I edited the message to treat the "&lt;user&gt;" as an html tag)

keltia: might be clearer to have `%USER%` ?

keltia: or even `(%HOME%\Documents\PowerShell\Microsoft.PowerShell_profile.ps1)`.  Not sure what a typical Windows user expect.

martinvonz: Sure, I'll let you decide.

martinvonz (approved):

PR #5245, opened 2025-01-03
cli: drop support for `jj init --{git, git_repo}`

thoughtpolice:
These were deprecated early last year in favor of `jj git init`.

# Checklist

If applicable:


cli/tests/test_init_command.rs
```diff
@@ -71,11 +71,9 @@ fn read_git_target(workspace_root: &Path) -> String {
 #[test]
 fn test_init_git_internal() {
```
yuja, line 72: These tests appear to be copied to test_git_init.rs. Maybe we can just remove them?

yuja (approved):

cli/tests/test_init_command.rs
thoughtpolice: It looks like all of them were dupes, so I just deleted them. Good catch!

Issue #5231, opened 2025-01-02
Unable to configure user name


Tiwalun:
## Description

I've just downloaded jujutsu on a new machine, but I'm unable to configure the username, or to interact with the config in any way. All operations involving config (`jj config`) seem to return an error about an invalid type or value.

## Steps to Reproduce the Problem

1. Install jujustu using `cargo binstall --strategies crate-meta-data jj-cli`
2. Try to set  user name using `jj config set --user user.name "Martin von Zweigbergk"`

The steps were performed on a machine where jujutsu was not installed, so no existing config was present.

## Expected Behavior

I can set the username

## Actual Behavior

The following error message is shown:

```
❯ jj config set --user user.name "Martin von Zweigbergk"
Config error: Invalid type or value for operation.username
Caused by: Null record
For help, see https://jj-vcs.github.io/jj/latest/config/.
```

I've tried setting the user name manually in ~/.config/jj/config.toml, but that doesn't seem to make a difference.

## Specifications

- Platform: Ubuntu 24.04
- Version: jj 0.25.0-041c4fecb77434dd6720e7d7f1ce48d9575ac5f7

arxanas:
Thanks for reporting. Does the issue also occur with jj v0.24?

ilyagr:
Weird, I can't immediately reproduce it. What does `jj config path --user` say? Is there anything in that file? Does deleting (or, more safely, moving) it help?

yuja:
Oh, do you have invalid (or maybe empty) name in your /etc/passwd entry? The old behavior was to fall back to "Unknown" or something. If that's common, the change should be reverted.

As a workaround, please add `operation.username = "whatever" to config.toml.

Tiwalun:
> Thanks for reporting. Does the issue also occur with jj v0.24?

No, it seems to work fine with version v.024.

> Oh, do you have invalid (or maybe empty) name in your /etc/passwd entry? The old behavior was to fall back to "Unknown" or something. If that's common, the change should be reverted.

I can't see anything obviously invalid in `/etc/passwd`, but I'm also not sure what exactly I should be looking for.

> As a workaround, please add `operation.username = "whatever"` to config.toml.

Thanks, this seems to work.

yuja:
> I can't see anything obviously invalid in `/etc/passwd`, but I'm also not sure what exactly I should be looking for.

Can you run these commands? I expect the pwd entry would be missing, but I have no idea why. (Is it some container environment?)
```sh
[...]

PR #5247, opened 2025-01-03
config: fall back to $USER if username couldn't be obtained by libc

yuja:
#5231

# Checklist

If applicable:

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -438,9 +438,23 @@ pub fn config_from_environment(default_layers: impl IntoIterator<Item = ConfigLa
     RawConfig(config)
 }
 
+const OP_HOSTNAME: &str = "operation.hostname";
+const OP_USERNAME: &str = "operation.username";
+
 /// Environment variables that should be overridden by config values
 fn env_base_layer() -> ConfigLayer {
     let mut layer = ConfigLayer::empty(ConfigSource::EnvBase);
+    // TODO: warn if hostname/username is empty after loading config files?
+    if let Ok(value) = whoami::fallible::hostname()
+        .inspect_err(|err| tracing::warn!(?err, "failed to get hostname"))
+    {
+        layer.set_value(OP_HOSTNAME, value).unwrap();
+    }
+    if let Ok(value) = whoami::fallible::username()
+        .inspect_err(|err| tracing::warn!(?err, "failed to get username"))
+    {
+        layer.set_value(OP_USERNAME, value).unwrap();
+    }
     if !env::var("NO_COLOR").unwrap_or_default().is_empty() {
         // "User-level configuration files and per-instance command-line arguments
         // should override $NO_COLOR." https://no-color.org/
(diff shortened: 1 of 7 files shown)


cli/tests/test_global_opts.rs
```diff
@@ -774,6 +777,105 @@ fn test_conditional_config() {
...
+        "<date-time>",
+    );
+    insta_settings.add_filter(r"\b[k-z]{8,12}\b", "<change-id>");
+    insta_settings.add_filter(r"\b[0-9a-f]{8,12}\b", "<id>");
+    let _guard = insta_settings.bind_to_scope();
+
+    let is_maskable_char = |c: u8| c.is_ascii_alphanumeric() | matches!(c, b'-' | b'.' | b'_');
+    let is_maskable_str = |s: &str| s.as_bytes().iter().all(|&c| is_maskable_char(c));
+    let maskable_op_user = whoami::fallible::hostname().is_ok_and(|s| is_maskable_str(&s))
+        && whoami::fallible::username().is_ok_and(|s| is_maskable_str(&s));
```
samueltardieu, line 832: Couldn't you concatenate `user`, `@`, and `host`, and use the same regexp as in the `insta` setting to check if it is maskable to ensure consistency?

cli/src/config.rs
```diff
@@ -451,6 +450,10 @@ fn env_base_layer() -> ConfigLayer {
         .inspect_err(|err| tracing::warn!(?err, "failed to get username"))
     {
         layer.set_value(OP_USERNAME, value).unwrap();
+    } else if let Ok(value) = env::var("USER") {
+        // On Unix, $USER is set by login(1). Use it as a fallback because
+        // getpwuid() of musl libc appears not (fully?) supporting nsswitch.
+        layer.set_value(OP_USERNAME, value).unwrap();
```
samueltardieu, line 456: If this is a last resort, shouldn't we even use something like

```rust
  if let Some(value) = env::var_os("USER") {
    layer.set_value(OP_USERNAME, value.to_string_lossy()).unwrap();
  }
```

?

samueltardieu (approved):

cli/src/config.rs
yuja: could be, but I don't have a strong reason to not copy the existing style. (All `env::var()` could be changed to lossy conversion because ignoring non-utf-8 values doesn't make sense either.)

cli/tests/test_global_opts.rs
yuja: It also works. If insta API accepted `Regex` object, I would do that way. I don't have strong preference over these.

cli/src/config.rs
yuja: This is unchanged for consistency with the other `env::var()` calls.

cli/tests/test_global_opts.rs
yuja: Updated to use regex, but the pattern is slightly different. `^..$` vs `\b..\b`.

Issue #5250, opened 2025-01-03
External diff tool exiting with code 1 should not cause a warning


dotdash:
## Description

diff uses exit code 0 to signal that there were no changes, 1 to signal that there were changes, and an exit code >1 to signal an error. Tools like `delta` do the same, but jj treats exit code 1 as a problem.

## Steps to Reproduce the Problem

1. Change something in your working copy
1. Run `jj diff --config ui.diff.tool=diff` or `jj diff --config ui.diff.tool=delta`

## Expected Behavior

No warnings show up.

## Actual Behavior

The output ends with:
```
Warning: Tool exited with exit status: 1 (run with --debug to see the exact invocation)
```

## Specifications

- Platform: Linux Debian Sid
- Version: jj 0.25.0-041c4fecb77434dd6720e7d7f1ce48d9575ac5f7

yuja:
It's not an error but a warning for that reason. Maybe we can add a per-tool list of expected exit codes to suppress the warning.

fwiw, you might want to set up `delta` as a pager, not a diff generator.

bryceberger:
Comment from PR (https://github.com/jj-vcs/jj/pull/5257#pullrequestreview-2529694650):

> Perhaps it would be better to reduce complexity in jj by having the user write their own wrapper for the tool? Then it can handle arbitrary exit conditions.

I'm in favor of a wrapper. Minimal example:
```bash
#!/usr/bin/env bash
diff "$@"
status=$?
case $status in
0 | 1) ;;
*)
    # custom handling, etc
    exit $status
    ;;
esac
```

dotdash:
> fwiw, you might want to set up `delta` as a pager, not a diff generator.

That works for e.g. ``jj diff` but not for `jj log -p` because `delta` doesn't recognize the diff format due to the vertical bar left of the diff (it fails the same way with `git log --graph -p`). Though using `ui.diff.tool` the output in `jj log -p` is still broken WRT to file path, because `delta` gets the raw paths instead of a git style diff 🤔 

Should I file a separate issue for that?

arxanas:
> > fwiw, you might want to set up `delta` as a pager, not a diff generator.
> 
[...]

PR #5257, opened 2025-01-03
diff: add `merge-tools.*.diff-expected-exit-codes`

bryceberger:
#5250

Certain tools (`diff`, `delta`) exit with code 1 to indicate there was
a difference. This allows selectively suppressing the "Tool exited with
... status" warning from jj when generating a diff.

example:
```toml
[merge.tools.delta]
diff-expected-exit-codes = [1]
```

# Checklist

If applicable:

### DIFF
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -380,6 +380,14 @@
                             "type": "string"
                         }
                     },
+                    "diff-expected-exit-codes": {
+                        "type": "array",
+                        "items": {
+                            "type": "integer"
+                        },
+                        "description": "Array of exit codes that do not indicate tool failure, i.e. [0, 1] for unix diff.",
+                        "default": [0]
+                    },
                     "diff-invocation-mode": {
                       "description": "Invoke the tool with directories or individual files",
                       "enum": [
--- a/cli/src/merge_tools/external.rs
+++ b/cli/src/merge_tools/external.rs
@@ -50,6 +50,8 @@ pub struct ExternalMergeTool {
     /// Arguments to pass to the program when generating diffs.
     /// `$left` and `$right` are replaced with the corresponding directories.
     pub diff_args: Vec<String>,
+    /// Exit codes to be treated as success when generating diffs.
+    pub diff_expected_exit_codes: Vec<i32>,
     /// Whether to execute the tool with a pair of directories or individual
     /// files.
     pub diff_invocation_mode: DiffToolMode,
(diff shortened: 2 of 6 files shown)


docs/config.md
```diff
@@ -295,6 +295,15 @@ diff.tool = "vimdiff"
 diff-invocation-mode = "file-by-file"
 ```
 
+By default `jj` will display a warning when the command exits with a non-success
+error code. The `diff-expected-exit-codes` config can suppress this warning
+message for specific exit codes:
+
+```toml
+[merge-tools.delta]
+expected-exit-codes = [1]
```
arxanas, line 304: question: Shouldn't this include `0` as an expected exit code?

arxanas (commented):
comment (non-blocking): Perhaps it would be better to reduce complexity in jj by having the user write their own wrapper for the tool? Then it can handle arbitrary exit conditions.

docs/config.md
bryceberger: `0` is covered by `exit.code.success()`. This is for allowed non-success error code. Perhaps docs should clarify that more? (0 is allowed, just redundant. No way to set 0 as an unexpected code.)

yuja: If it's named as "expected"/"success"/"ok" exit codes, I agree it should include `0`, and the default should be `[0]`.

We already have `merge-conflict-exit-codes`, so I wouldn't mind the added complexity. The list could be `diff-different-exit-codes`, but we wouldn't have any use case other than suppressing the warning.

bryceberger: To clarify, the new behavior is:

- `merge-tools.*.diff-expected-exit-codes` defaults to `[0]`.
- `exit_code.success()` is never considered.
- If the config value is set to `[]`, the warning will be printed on every invocation.
- If the exit code can't be determined (??), the warning will be printed.

cli/src/config-schema.json
```diff
@@ -380,6 +380,12 @@
                             "type": "string"
                         }
                     },
+                    "diff-expected-exit-codes": {
+                        "type": "array",
+                        "items": {
+                            "type": "integer"
+                        }
```
yuja, line 387: nit: can you add "description" and "default"? I think they can be copied from `merge-conflict-exit-codes`.

cli/src/merge_tools/external.rs
```diff
@@ -411,7 +414,10 @@ pub fn invoke_external_diff(
     // will exit with 1 if inputs are different.
     let exit_status = child.wait().map_err(ExternalToolError::Io)?;
     tracing::info!(?cmd, ?exit_status, "The external diff generator exited:");
-    if !exit_status.success() {
+    let exit_ok = exit_status
+        .code()
+        .is_some_and(|status| tool.diff_expected_exit_codes.contains(&status));
```
yuja, line 419: Can you add non-zero exit code test to test_diff_command.rs?

docs/config.md
yuja: Thanks. An empty value is useless, but I think that's the correct behavior.

yuja (approved):

docs/config.md
yuja, line 304: Oh, and update the doc example to include `0`.

PR #5259, opened 2025-01-03
cli: make git support optional

thoughtpolice:
There are some experiments to try and compile `jj` to WebAssembly, so that we might be able to do things like interactive web tutorials. One step for that is making `git` support in `jj-cli` optional, because we can stub it out for something more appropriate and it's otherwise a lot of porting annoyance for both gitoxide and libgit2.

(On top of that, it might be a useful build configuration for other experiments of mine where removing the need for the large libgit2 depchain is useful.)

As part of this, we need to mark `jj-lib` as having `default-features = false` in the workspace dependency configuration; otherwise, the default behavior for Cargo is to compile with all its default features, i.e. with git support enabled, ignoring the `jj-cli` features clauses.

Other than that, it is fairly straightforward junk; it largely just sprinkles some `#[cfg]` around liberally in order to make things work. It also adjusts the CI pipeline so this is tested there, too, so we can progressively clean it up.

# Checklist

If applicable:

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -23,7 +23,7 @@ include = [
     "/tests/",
     "!*.pending-snap",
     "!*.snap*",
-    "/tests/cli-reference@.md.snap"
+    "/tests/cli-reference@.md.snap",
 ]
 
 [[bin]]
@@ -61,8 +61,8 @@ crossterm = { workspace = true }
 dirs = { workspace = true }
 dunce = { workspace = true }
 futures = { workspace = true }
-git2 = { workspace = true }
-gix = { workspace = true }
+git2 = { workspace = true, optional = true }
+gix = { workspace = true, optional = true }
 glob = { workspace = true }
 indexmap = { workspace = true }
 indoc = { workspace = true }
@@ -109,8 +109,9 @@ testutils = { workspace = true }
 jj-cli = { path = ".", features = ["test-fakes"], default-features = false }
 
 [features]
-default = ["watchman"]
+default = ["watchman", "git"]
 bench = ["dep:criterion"]
+git = ["jj-lib/git", "dep:git2", "dep:gix"]
 gix-max-performance = ["jj-lib/gix-max-performance"]
 packaging = ["gix-max-performance"]
 test-fakes = ["jj-lib/testing"]
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -17,7 +17,6 @@ use std::cell::OnceCell;
 use std::collections::BTreeMap;
 use std::collections::HashSet;
 use std::env;
-use std::env::VarError;
 use std::ffi::OsString;
 use std::fmt;
 use std::fmt::Debug;
@@ -64,12 +63,9 @@ use jj_lib::config::ConfigNamePathBuf;
 use jj_lib::config::ConfigSource;
 use jj_lib::config::StackedConfig;
 use jj_lib::conflicts::ConflictMarkerStyle;
-use jj_lib::file_util;
 use jj_lib::fileset;
 use jj_lib::fileset::FilesetDiagnostics;
 use jj_lib::fileset::FilesetExpression;
-use jj_lib::git;
-use jj_lib::git_backend::GitBackend;
 use jj_lib::gitignore::GitIgnoreError;
 use jj_lib::gitignore::GitIgnoreFile;
 use jj_lib::id_prefix::IdPrefixContext;
@@ -165,9 +161,6 @@ use crate::diff_util::DiffRenderer;
 use crate::formatter::FormatRecorder;
 use crate::formatter::Formatter;
 use crate::formatter::PlainTextFormatter;
-use crate::git_util::is_colocated_git_workspace;
-use crate::git_util::print_failed_git_export;
-use crate::git_util::print_git_import_stats;
 use crate::merge_tools::DiffEditor;
 use crate::merge_tools::MergeEditor;
 use crate::merge_tools::MergeToolConfigError;
@@ -662,10 +655,6 @@ impl ReadonlyUserRepo {
             id_prefix_context: OnceCell::new(),
         }
     }
-
-    pub fn git_backend(&self) -> Option<&GitBackend> {
-        self.repo.store().backend_impl().downcast_ref()
-    }
 }
 
 /// A advanceable bookmark to satisfy the "advance-bookmarks" feature.
@@ -999,7 +988,9 @@ impl WorkspaceCommandHelper {
         let op_summary_template_text = settings.get_string("templates.op_summary")?;
         let may_update_working_copy =
             loaded_at_head && !env.command.global_args().ignore_working_copy;
-        let working_copy_shared_with_git = is_colocated_git_workspace(&workspace, &repo);
+        let working_copy_shared_with_git =
+            crate::git_util::is_colocated_git_workspace(&workspace, &repo);
+
         let helper = Self {
             workspace,
             user_repo: ReadonlyUserRepo::new(repo),
@@ -1020,10 +1011,6 @@ impl WorkspaceCommandHelper {
     /// Settings for this workspace.
     pub fn settings(&self) -> &UserSettings {
         self.workspace.settings()
-    }
-
-    pub fn git_backend(&self) -> Option<&GitBackend> {
-        self.user_repo.git_backend()
     }
 
     pub fn check_working_copy_writable(&self) -> Result<(), CommandError> {
@@ -1048,6 +1035,7 @@ impl WorkspaceCommandHelper {
             return Ok(SnapshotStats::default());
         }
 
+        #[cfg(feature = "git")]
         if self.working_copy_shared_with_git {
             self.import_git_head(ui).map_err(snapshot_command_error)?;
         }
@@ -1058,6 +1046,7 @@ impl WorkspaceCommandHelper {
         let stats = self.snapshot_working_copy(ui)?;
 
         // import_git_refs() can rebase the working-copy commit.
+        #[cfg(feature = "git")]
         if self.working_copy_shared_with_git {
             self.import_git_refs(ui).map_err(snapshot_command_error)?;
         }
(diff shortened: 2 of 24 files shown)


thoughtpolice:
This is basically a huge hack right now, and the testsuite doesn't work in a no-git configuration of course, but it does make everything otherwise work together.

cli/src/cli_util.rs
```diff
@@ -1837,13 +1861,23 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
...
+                    .snapshot(&options)
+                    .map_err(snapshot_command_error)?
+            }
+            #[cfg(not(feature = "git"))]
+            {
+                locked_ws
+                    .locked_wc()
+                    .snapshot(&options)
+                    .map_err(snapshot_command_error)?
+            }
```
thoughtpolice, line 1880: I couldn't quickly think of a way to write this that didn't have warnings or weird code, but suggestions welcome.

lib/src/lib.rs
```diff
@@ -53,6 +53,13 @@ pub mod fsmonitor;
 pub mod git;
 #[cfg(feature = "git")]
 pub mod git_backend;
+#[cfg(not(feature = "git"))]
+/// Stub module for when the `git` feature is disabled.
+pub mod git {
+    // TODO FIXME: hack
+    /// Reserved remote name for the backing Git repo.
+    pub const REMOTE_NAME_FOR_LOCAL_GIT_REPO: &str = "git";
+}
```
thoughtpolice, line 62: Removing this probably needs a little discussion, I guess, this was just a hack. We probably just want to not use it at all in the given set of commands (like `diff`), but I didn't want to get into a bigger change at the moment, `#[cfg]` is otherwise a little awkward e.g. with chain method syntax.

thoughtpolice:
I also removed use of `zstd` in the native backend since it doesn't serve much purpose beyond adding dependencies. I think we can bring back some other compression crate later.

cli/src/cli_util.rs
thoughtpolice: I improved this a bit, but I'm not sure I like it.

lib/src/lib.rs
martinvonz: Perhaps some helper (`is_git_remote(&str) -> bool`?) can be moved to `git_util`

cli/Cargo.toml
```diff
@@ -61,13 +61,13 @@ crossterm = { workspace = true }
...
-git2 = { workspace = true }
-gix = { workspace = true }
+git2 = { workspace = true, optional = true }
+gix = { workspace = true, optional = true }
 glob = { workspace = true }
 indexmap = { workspace = true }
 indoc = { workspace = true }
 itertools = { workspace = true }
-jj-lib = { workspace = true }
+jj-lib = { workspace = true, default-features = false }
```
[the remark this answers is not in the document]
thoughtpolice: No, leftover debugging tool. If you leave this line in but *remove* it from the workspace `Cargo.toml` file you get a warning, which is what tipped me off to the fix in the first place.

cli/src/progress.rs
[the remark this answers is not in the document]
thoughtpolice: No, another artifact of just doing "fix every error that comes by in order" and being mercy to the order of compiler errors. Good catch.

lib/src/lib.rs
thoughtpolice: I thought about that. But you'd still need to `#[cfg]` the use sites in all the command files. But then again, I was wondering if we should even special case the use of the `git` name at all if git support isn't compiled into the binary.

thoughtpolice: I reworked it a bit, but I'm still not fully satisfied.  I put something in `cli_utils` and in `lib` to try and minimize `#[cfg]` clauses, so let me know what you think (maybe make a new comment and resolve this one.)

yuja: I don't think `#[cfg]` is needed at call sites. I think it's conceptually a `Backend` property like `reserved_remote_name() -> Option<&str>` (but isn't general enough to add trait method.)

cli/src/cli_util.rs
```diff
@@ -634,6 +641,7 @@ impl ReadonlyUserRepo {
         }
     }
 
+    #[cfg(feature = "git")]
```
yuja, line 644: nit: does it help if we had a dummy `git_utils` or something that stubs out various git functions and types?

It might be also good to extract git-related functions to separate `impl` block, and feature-gate it.

cli/src/command_error.rs
```diff
@@ -463,12 +467,14 @@ impl From<TempTextEditError> for CommandError {
     }
 }
 
+#[cfg(feature = "git")]
```
yuja, line 470: nit: maybe we can extract these `impl`s to inner module, and add `#[cfg(feature = "git")]` there.

cli/src/cli_util.rs
thoughtpolice: I tried this approach and was able to reduce the amount of `#[cfg]` junk a bit. I think we need to just keep cleaning things up. It seems like we're probably even pretty close to being able to put the `git`/`git_util` code into a separate crate, and then feed it into `jj-cli`? Maybe I can take a stab at it.

yuja: Looks better, thanks.

> It seems like we're probably even pretty close to being able to put the `git`/`git_util` code into a separate crate, and then feed it into `jj-cli`?

I think many feature-gated jj-lib modules can be extracted to jj-lib-git or something (but we might have to extract core/utility types first to re-export `jj-lib-git::*` from jj-lib and/or run jj-lib tests.) I have no idea about jj-cli part.

btw, dummy type could be an empty enum to ensure that it'll never be constructed.

lib/src/lib.rs
```diff
@@ -51,8 +51,27 @@ pub mod fmt_util;
...
 #[cfg(feature = "git")]
 pub mod git_backend;
+#[cfg(not(feature = "git"))]
+/// A dummy module that provides a no-op implementation of the functions in the
+/// `git_backend` module.
+pub mod git_backend {
+    /// A dummy struct that represents the Git backend.
+    pub struct GitBackend;
+}
```
martinvonz, line 74: Looks like this can be removed if we define the two `git_backend()` methods in `cli_util` only when `git` is active. We can actually inline `ReadonlyUserRepo::git_backend()` too

cli/src/cli_util.rs
thoughtpolice: I tried a little last night and was able to make a start on a `jj-git` crate that moves some of the code in there and some tests there. We'll probably have to do a series of moves like you mention, for a types-based crate. I'll submit another change later.

lib/src/lib.rs
thoughtpolice: I think this is resolved for now with the current changes and seems to be the least of all evils.

thoughtpolice:
@yuja I also went ahead and included a commit from my other stack that removes an old migration hack from `jj-lib`, reducing the `#[cfg]` count by one more. Let me know if you want to keep this around and I can split it out.

lib/src/lib.rs
```diff
@@ -51,8 +51,27 @@ pub mod fmt_util;
 pub mod fsmonitor;
 #[cfg(feature = "git")]
 pub mod git;
+#[cfg(not(feature = "git"))]
+/// A dummy module that provides a no-op implementation of the functions in the
```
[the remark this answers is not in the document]
martinvonz: Not done here

cli/src/progress.rs
```diff
@@ -1,128 +1,32 @@
...
-    }
-}
-
-const UPDATE_HZ: u32 = 30;
-const INITIAL_DELAY: Duration = Duration::from_millis(250);
+pub const UPDATE_HZ: u32 = 30;
+pub const INITIAL_DELAY: Duration = Duration::from_millis(250);
 
-struct RateEstimate {
+pub struct RateEstimate {
```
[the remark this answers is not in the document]
thoughtpolice: Ah, I thought there was still a dependency on that for some reason. Moved.

martinvonz (approved):

martinvonz:
The last commit seems to remove support for a format from ~2 years ago. That seems okay to me to remove.

cli/Cargo.toml
```diff
@@ -109,10 +109,11 @@ testutils = { workspace = true }
...
 
 [features]
-default = ["watchman"]
+default = ["watchman", "git"]
+packaging = ["gix-max-performance"]
+git = ["jj-lib/git", "dep:git2", "dep:gix"]
 bench = ["dep:criterion"]
 gix-max-performance = ["jj-lib/gix-max-performance"]
-packaging = ["gix-max-performance"]
 test-fakes = ["jj-lib/testing"]
```
yuja, line 117: nit: sort lexicographically as it was (mostly) before?

cli/src/cli_util.rs
```diff
@@ -3767,6 +3773,14 @@ fn format_template_aliases_hint(template_aliases: &TemplateAliasesMap) -> String
     hint
 }
 
+/// Determine, by its name, if a remote refers to the special local-only "git"
+/// remote that is used in the Git backend.
+///
+/// This function always returns false if the "git" feature is not enabled.
+pub fn is_special_git_remote(_remote: &str) -> bool {
```
yuja, line 3780: nit: `_remote` argument is used, and maybe this stub function is no longer needed?

cli/src/git_util.rs
```diff
@@ -326,6 +332,150 @@ pub fn print_git_import_stats(
...
+            rate: RateEstimate::new(),
+            buffer: String::new(),
+            guard: None,
+        }
+    }
+
+    pub fn update<W: std::io::Write>(
+        &mut self,
+        now: Instant,
+        progress: &git::Progress,
```
yuja, line 355: nit: might be better to extract `git::Progress` type to somewhere else. The implementation looks general for file downloading/uploading workload.

thoughtpolice: I agree with this, but for now it's only used in `git_util` anyway, so I think inlining it here is fine as it is right now. I can extract it back out into `progress` later, since I might need it anyway.

yuja: To minimize the change, I would rather not move codes back and forth. I don't have strong feeling, though.

martinvonz: Do we think it will be reused by non-git commands? I kind of feel like it won't.

yuja: If we add other backend, it might be. For the moment, no, I don't think it will be reused.

martinvonz: I mean that even if we add another backend, the current struct (`bytes_downloaded: Option<u64>, overall: f32`) seems too specific to how Git does things that I'm not sure we will reuse it. Perhaps it can be made into a more generic progress bar, however.

yuja: Yes, the current data struct might have to be adjusted. My feeling is that the UI impl is mostly backend agnostic.

Issue #5144, opened 2024-12-19
Conditional config doesn't apply when initializing or cloning a repository


jennings:
## Description

When initializing or cloning a new repository, conditional config is not considered when creating the first empty commit.

## Steps to Reproduce the Problem

1. Add the following conditional config (Windows, but probably not relevant to this issue):
   ```toml
   [user]
   name = "Stephen"
   [[--scope]]
   --when.repositories = ["\\\\?\\C:\\src\\github.com"]
   [--scope.user]
   email = "stephen@example.com"
   ```
2. Execute: `jj git clone git@github.com:jj-vcs/jj C:\src\github.com\jj`

## Expected Behavior

The working copy commit created by jj should have the email `stephen@example.com`.

## Actual Behavior

The working copy commit has no email address set.

## Specifications

- Platform: Windows
- Version: jj 0.24.0-f5d450d7c3a8f8a49e30aef0fd18d8295ffc8cb6

yuja:
It's working as intended (because the repository to load configuration for doesn't exist), but yeah, it's a bit inconvenient. We might need some way to specify configuration for the cwd, not for the repository/workspace directory.

yuja:
> Suppose we had both `--when.repositories` and `--when.cwd`. Is there a scenario when you'd want the `repositories` behavior? If not, I think special case might make the most sense.

In most scenarios, `--when.repositories` (or `--when.workspaces`) would be preferred. The workspace path can be specified by `-R` (which doesn't change the cwd.) That said, I agree that special case is probably better. It might look ugly, but initialization code is a bit weird in some way.

PR #5268, opened 2025-01-05
cli: resolve settings for newly initialized/cloned workspace

yuja:
#5144

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -346,6 +346,20 @@ impl CommandHelper {
         &self.data.settings
     }
 
+    /// Resolves configuration for new workspace located at the specified path.
+    pub fn settings_for_new_workspace(
+        &self,
+        workspace_root: &Path,
+    ) -> Result<UserSettings, CommandError> {
+        let mut config_env = self.data.config_env.clone();
+        let mut raw_config = self.data.raw_config.clone();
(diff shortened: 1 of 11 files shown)


joyously:
Is it possible that this could use enough defaults that you can run read-only `jj` commands without having done `init`?
Yesterday I was rereading[ a comment I made](https://github.com/jj-vcs/jj/discussions/2065#discussioncomment-8673556) about how Breezy commands work in a Git repo without doing `init`.

martinvonz:
> Is it possible that this could use enough defaults that you can run read-only `jj` commands without having done `init`?

`jj git init` also walks all commits and writes some information about them in a persistent index, so it's not just a matter of configuration variables.

cli/src/cli_util.rs
```diff
@@ -345,6 +345,20 @@ impl CommandHelper {
         &self.data.settings
     }
 
+    /// Resolves configuration for new workspace located at the specified path.
```
martinvonz, line 348: Maybe add comment explaining that it may be different due to conditional config?

martinvonz (approved):

cli/src/cli_util.rs
yuja: Updated the doc to mention conditional variables and repo config file.

PR #5269, opened 2025-01-05
added changelog entry for showing untracked files in `jj status`

cstoitner:
I am not sure this belongs in breaking changes or new features. I have it as breaking right now because it changes the output of the status command.

# Checklist

If applicable:


martinvonz:
> I am not sure this belongs in breaking changes or new features. I have it as breaking right now because it changes the output of the status command.

The output is not specified, so I think we don't need to consider it a breaking change.

cstoitner:
I move the line to new features and reworded it slightly.

martinvonz (approved):
Thanks!

PR #5279, opened 2025-01-06
git init: add option to track local bookmarks

bryceberger:
Uses the same list of bookmarks that is otherwise printed to the user as local bookmarks not associated with their remote. Essentially, just automatically runs the hinted `jj bookmark track ...`.

I use an external tool that is not jj-aware to manage my repositories (ghq). As such, I find myself running `jj git init --colocate` fairly often. Almost always, I immediately follow the hint and copy and paste the given `jj bookmark track ...` command. Would like to just skip that step.

# Checklist

If applicable:


cli/src/commands/git/init.rs
```diff
@@ -183,7 +184,13 @@ fn do_init(
                     tx.finish(ui, "import git head")?;
                 }
             }
-            print_trackable_remote_bookmarks(ui, workspace_command.repo().view())?;
+            if settings.get_bool("git.track-on-init")? {
+                let mut checkout_tx = workspace_command.start_transaction();
```
martinvonz, line 188: Find a better name :)

Actually, maybe we should just use the same transaction as for the import above? Should be able to move the `let mut tx = ...` line outside the block and also move the `if tx.repo().has_changes()` block later.

cli/src/commands/git/init.rs
```diff
@@ -252,3 +259,32 @@ pub fn maybe_set_repository_level_trunk_alias(
 
     Ok(())
 }
+
+fn track_local_bookmarks(ui: &mut Ui, repo: &mut MutableRepo) -> Result<(), CommandError> {
```
martinvonz, line 263: Don't forget to add tests

flake.nix
martinvonz, line 1: Unrelated changes? Please revert (move to a separate PR if you want to keep them)

martinvonz (commented):
Sounds like a reasonable feature, thanks

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+* New `git.init-track-local-bookmarks` config option to automatically track
+  local bookmarks during `jj git init --colocate`.
```
martinvonz, line 12: This should go under the "New features" heading

cli/src/commands/git/init.rs
```diff
@@ -252,3 +258,32 @@ pub fn maybe_set_repository_level_trunk_alias(
...
+    let pairs: Vec<(String, String)> = repo
+        .view()
+        .bookmarks()
+        .filter(|(_, bookmark_target)| bookmark_target.local_target.is_present())
+        .flat_map(|(name, bookmark_target)| {
+            bookmark_target
+                .remote_refs
+                .into_iter()
+                .filter(|&(remote_name, _)| remote_name != "git")
```
martinvonz, line 271: nit: Use `git::is_special_git_remote(remote_name)` (after rebasing past e15429b)

cli/src/commands/git/init.rs
```diff
@@ -252,3 +258,32 @@ pub fn maybe_set_repository_level_trunk_alias(
...
+                .map(|(remote_name, _)| (name.into(), remote_name.into()))
```
martinvonz, line 272: nit: Use the more specific (and therefore clearer) `.to_owned()` or `to_string()` here. Feel free to ignore if you like `.into()` better.

cli/src/commands/git/init.rs
```diff
@@ -172,18 +173,23 @@ fn do_init(
...
                 }
-                if tx.repo().has_changes() {
-                    tx.finish(ui, "import git head")?;
-                }
             }
-            print_trackable_remote_bookmarks(ui, workspace_command.repo().view())?;
+            if settings.get_bool("git.init-track-local-bookmarks")? {
+                track_local_bookmarks(ui, tx.repo_mut())?;
+            } else {
+                print_trackable_remote_bookmarks(ui, tx.repo().view())?;
```
martinvonz, line 188: Seems like this function could be moved into this file. It seems better to have `track_local_bookmarks()` and `print_trackable_remote_bookmarks()` near each other, don't you think?

cli/src/commands/git/init.rs
```diff
@@ -252,3 +258,32 @@ pub fn maybe_set_repository_level_trunk_alias(
 
     Ok(())
 }
+
+fn track_local_bookmarks(ui: &mut Ui, repo: &mut MutableRepo) -> Result<(), CommandError> {
+    let pairs: Vec<(String, String)> = repo
```
martinvonz, line 263: nit: `bookmarks` or `remote_bookmarks` seems clearer than `pairs`

cli/tests/test_git_init.rs
```diff
@@ -743,6 +743,79 @@ fn test_git_init_colocated_via_flag_git_dir_exists() {
...
+    insta::assert_snapshot!(stdout, @"");
+    insta::assert_snapshot!(stderr, @r#"
+    Done importing changes from the underlying Git repo.
+    Setting the revset alias "trunk()" to "my-bookmark@origin"
+    Tracking the following remote bookmarks:
+      my-bookmark@origin
+    Initialized repo in "repo_track"
+    "#);
+
+    // Check that the bookmark is checked out
```
martinvonz, line 777: nit: "checked out" -> "tracked"? same on line 810

cli/src/commands/git/init.rs
```diff
@@ -252,3 +258,32 @@ pub fn maybe_set_repository_level_trunk_alias(
 
     Ok(())
 }
+
+fn track_local_bookmarks(ui: &mut Ui, repo: &mut MutableRepo) -> Result<(), CommandError> {
```
yuja, line 262: nit: same here `import_tracking_local_bookmarks()` or something?

cli/src/commands/git/init.rs
```diff
@@ -252,3 +258,32 @@ pub fn maybe_set_repository_level_trunk_alias(
...
+        writeln!(formatter, "Tracking the following remote bookmarks:")?;
+        for (name, remote_name) in &pairs {
+            write!(formatter, "  ")?;
+            writeln!(formatter.labeled("bookmark"), "{name}@{remote_name}")?;
+        }
+    }
+
+    for (name, remote_name) in &pairs {
+        repo.track_remote_bookmark(name, remote_name);
+    }
```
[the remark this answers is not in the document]
bryceberger: I think it would make more sense to only track if local and remote are the same, and print the current hint text if they're different.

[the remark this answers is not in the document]
yuja: or if local is ahead of remote? `local == remote` might be too strict.

cli/src/commands/git/init.rs
```diff
@@ -252,3 +265,123 @@ pub fn maybe_set_repository_level_trunk_alias(
...
+        .evaluate(repo)
+        .map(|r| r.containing_fn())
+    else {
+        return Some(Err(format!("{name}@{remote_name}")));
+    };
+
+    if local_target
+        .added_ids()
+        .any(|add| ahead_of_remote(add).unwrap_or(false))
+    {
```
bryceberger, line 382: @yuja would like you to look at this. My intent was:
- get descendants of remote ref
- check if local ref is contained in above set

Unsure if this is the best way to do that, or if that's exactly what's wanted.

Also, this triggers if any of the `added_ids()` are a descendant. _Pretty_ sure this is what's wanted, as it will trigger even if local has merged in another branch.

yuja: Perhaps, you can use `repo.index().is_ancestor()`. It's also good to use `refs::merge_ref_targets()` and check the merge result, something like:
```rust
let base_target = remote_ref.tracking_target();
let merged_target = merge_ref_targets(index, local_target, base_target, &remote_ref.target);
merged_target == local_target
```

PR #5283, opened 2025-01-06
github: migrate all actions to `ubuntu-24.04`

thoughtpolice:
This gets rid of an annoying warning in the Actions tab, but we already rely on this for the build-binaries workflow, and it's probably better to be explicit about this anyway.

# Checklist

If applicable:

### DIFF
--- a/.github/workflows/build-nix.yml
+++ b/.github/workflows/build-nix.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-14]
+        os: [ubuntu-24.04, macos-14]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15 # NOTE (aseipp): keep in-sync with the build.yml timeout limit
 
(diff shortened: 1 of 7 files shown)


.github/workflows/build-nix.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-14]
+        os: [ubuntu-24.04, macos-14]
```
martinvonz, line 20: How do we remember to keep up to date? I guess we'll just have to remember to check once a year or something and decide if we want to upgrade?

Speaking of that, we should we switch from macos-14 to macos-15? And maybe from macos-13 to macos-14?

thoughtpolice: Ubuntu does an update to their LTS once every two years, so I don't think keeping up with it is much of a problem.

I think we want to keep macOS 13 because it's the last x86_64 macOS platform. macos-15 is still in Public Preview; 14 is just what the default is, and we're already fairly explicit about that.

martinvonz (approved):

PR #5284, opened 2025-01-06
Roll back 9d4a973 (resolve intermediate parent tree when rebasing merge commits)

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_merge_trees.rs
+++ b/lib/tests/test_merge_trees.rs
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 use itertools::Itertools;
+use jj_lib::backend::MergedTreeId;
 use jj_lib::backend::TreeValue;
+use jj_lib::merge::Merge;
 use jj_lib::repo::Repo;
 use jj_lib::repo_path::RepoPath;
 use jj_lib::repo_path::RepoPathBuf;
@@ -675,8 +677,7 @@ fn test_rebase_linearize_lossy_merge() {
 
     let commit_d2 = rebase_commit(repo_mut, commit_d, vec![commit_b.id().clone()]).unwrap();
 
-    // TODO: Should be tree 2
-    assert_eq!(*commit_d2.tree_id(), tree_1.id());
+    assert_eq!(*commit_d2.tree_id(), tree_2.id());
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


lib/tests/test_merge_trees.rs
```diff
@@ -627,13 +629,64 @@ fn test_simplify_conflict_after_resolving_parent() {
...
+    // Test this rebase:
+    // D    foo=2          D' foo=2
+    // |\                  |
+    // | C  foo=2          |
+    // | |           =>    B  foo=2
+    // B |  foo=2          |
+    // |/                  |
+    // A    foo=1          A  foo=1
+    //
+    // Commit D effectively discarded a change from "1" to "2", so one
```
arxanas, line 646: question: I didn't understand this. In what sense did this effectively discard a change from `1` to `2`? Isn't the value `2` in all parent commits? Why would `D` become `1` when rebased onto `B`?

lib/tests/test_merge_trees.rs
```diff
@@ -627,13 +629,64 @@ fn test_simplify_conflict_after_resolving_parent() {
...
+    let repo = &test_repo.repo;
+
+    // Test this rebase:
+    // D    foo=2          D' foo=2
+    // |\                  |
+    // | C  foo=2          |
+    // | |           =>    B  foo=2
+    // B |  foo=2          |
+    // |/                  |
+    // A    foo=1          A  foo=1
```
arxanas, line 644: comment (non-blocking): Fun idea: We could use [DrawDag](https://sapling-scm.com/docs/dev/internals/drawdag/) to start drawing out our test cases (with the crate being [`sapling-drawdag`](https://crates.io/crates/sapling-drawdag)).

martinvonz: I'm familiar with the `.t` version of that from the hg project. It was quite useful there at least. I think it's less annoying to produce graphs using the test libraries we already have, so I'm not sure it's worth using DrawDag in Rust tests. (Not saying I'm strongly against it either, in case someone else wants to experiment with it.)

lib/tests/test_merge_trees.rs
martinvonz: I've updated the description. Let me know if it's still unclear.

yuja (approved):

arxanas (approved):

Issue #5255, opened 2025-01-03
`core.watchman.register_snapshot_trigger` in a config. section causes an error on all CLI invocations


ErichDonGubler:
## Description

Setting `core.fsmonitor` to `"watchman"` and `core.watchman.register_snapshot_trigger` using a table section in configuration:

```toml
# Can be placed anywhere in configuration, stopped working with 0.25.0.
[core]
fsmonitor = "watchman"
watchman = { register_snapshot_trigger = true }
```

…rather than fully qualified key paths, as [documentation suggests](https://github.com/jj-vcs/jj/blob/337663c312e55282ce5ddbe10c83f7fe64ed8c19/docs/config.md?plain=1#L1142-L1153):

```toml
# Must be placed before any config. sections, still works.
core.fsmonitor = "watchman"
core.watchman.register_snapshot_trigger = true
```

…now causes a configuration error that blocks most invocations of `jj`. 😱 This is a problem, because these are supposed to be equivalent in TOML, but `jj` apparently treats them differently.

I've bisected this issue to e5361c6c (CC @yuja).

## Steps to Reproduce the Problem

1. With `jj` 0.25.0, use a configuration with the following `core` section:
	
	```toml
	[core]
	fsmonitor = "watchman"
	watchman = { register_snapshot_trigger = true }
	```

1. Run a `jj` command that loads configuration as one of its steps, i.e., `jj log`. Observe that only the following message is printed, with a non-zero exit code:

	```
	Config error: Value not found for core.watchman.register_snapshot_trigger
	For help, see https://jj-vcs.github.io/jj/latest/config/.
	```

1. Comment out configuration of one of these `core` keys. Observe that the error goes away, and `jj` works as before.

## Expected Behavior

I expect "bare" key paths and key paths in sections to have the same config. behavior, in general.

## Specifications

Reproduced on macOS 15.1.1 and Windows 11.

yuja:
It's actually "broken" by 0de36918e4c020e0e54816f29c47cb57cc9cfbf5, and the change was intentional. We have to merge multiple TOML documents into a single table, which is something that is TOML but duplicated keys are allowed. I decided to not take inline table as a merge-able table because it would probably make sense to override the default `colors.<name> = { style .. }` by user-specified value. (And for consistency reasons, inner values of inline table aren't addressable by dotted key.)

That said, if the new behavior is super confusing, maybe we should revert the change and take inline table as mergeable item.

emilazy:
[...]

PR #5297, opened 2025-01-08
config: allow inline table syntax in mutation and conditional scope API

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_config_command.rs
+++ b/cli/tests/test_config_command.rs
@@ -691,7 +691,10 @@ fn test_config_unset_non_existent_key() {
 
 #[test]
 fn test_config_unset_inline_table_key() {
-    let test_env = TestEnvironment::default();
+    let mut test_env = TestEnvironment::default();
+    // Test with fresh new config file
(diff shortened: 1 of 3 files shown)


martinvonz (commented):
@ErichDonGubler, FYI since you had opinions about TOML handling in #5255. IIUC, this PR makes it so if you have `foo = {a = 1}` in your user-level config and then `foo = {b = 2}` in your repo-level config, then the resulting config in the repo will be `foo = {a = 1, b = 2}`. I think that makes sense, and it matches what would happen if you instead had defined those configs using dotted notation instead of inline tables.

yuja:
> IIUC, this PR makes it so if you have `foo = {a = 1}` in your user-level config and then `foo = {b = 2}` in your repo-level config, then the resulting config in the repo will be `foo = {a = 1, b = 2}`.

It's implemented (or reverted) by d001291a. What this PR does is to allow `jj config set --user foo.b 2` to get `foo = {a = 1, b = 2}` as it would do for non-inline table `foo.a = 1`.

martinvonz (approved):

PR #5298, opened 2025-01-08
config: add simple migration logic to update deprecated variables

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -59,6 +59,7 @@ use jj_lib::commit::Commit;
 use jj_lib::config::ConfigGetError;
 use jj_lib::config::ConfigGetResultExt as _;
 use jj_lib::config::ConfigLayer;
+use jj_lib::config::ConfigMigrationRule;
 use jj_lib::config::ConfigNamePathBuf;
 use jj_lib::config::ConfigSource;
 use jj_lib::config::StackedConfig;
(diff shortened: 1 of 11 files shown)


lib/src/config.rs
```diff
@@ -203,7 +213,7 @@ impl fmt::Display for ConfigNamePathBuf {
 /// constrained by the source type.
 pub trait ToConfigNamePath: Sized {
     /// Path type to be converted from `Self`.
-    type Output: Borrow<ConfigNamePathBuf>;
+    type Output: Borrow<ConfigNamePathBuf> + Into<ConfigNamePathBuf>;
```
martinvonz, line 216: Is `ToOwned` more appropriate here? (I don't know the answer.)

cli/src/cli_util.rs
martinvonz, line 1: > The list of migration rules is managed by CliRunner. I don't know if that's
needed, but in theory, an extension may insert migration rules as well as
default config layers.

Thanks! I don't think we have any need for it yet at Google, but maybe we will in the future, or maybe some other extension will.

martinvonz (approved):

lib/src/config.rs
yuja: No. It's micro optimization to move owned (or parsed) `ConfigNamePathBuf` without cloning. `.to_owned()` doesn't consume `self`, but `.into()` can.

PR #5308, opened 2025-01-09
templates: add cryptographic_signature display to default formats

bryceberger:
Cryptographic signature support in templates was added in c99c97c6467fee3f3269d2934c10c758e041f834, but has to be manually configured. This adds some defaults to the built-in config.

Instead of having separate `builtin_log_*_with_sig` aliases, this adds to the aliases that actually format commits. Since signature verification is slow, this is disabled by default. To enable it, override the `should_show_cryptographic_signature()` template alias like so:

    [template-aliases]
    'should_show_cryptographic_signature()' = 'true'
    'format_short_cryptographic_signature(signature)' = ...
    'format_detailed_cryptographic_signature(signature)' = ...

Note that the two formatting functions take `Option<Signature>`, not `Signature`. This allows you to display a custom message if a signature is not found, but will emit an error if you do not check for signature presence.

    [template-aliases]
    'format_detailed_cryptographic_signature(signature)' = '''
      if(signature,
        "message if present",
        "message if missing",
      )
    '''

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/colors.toml
+++ b/cli/src/config/colors.toml
@@ -111,3 +111,10 @@
 "node current_operation" = { fg = "green", bold = true }
 "node immutable" = { fg = "bright cyan", bold = true }
 "node conflict" = { fg = "red", bold = true }
+
+"signature display" = "yellow"
+"signature key" = "cyan"
+"signature status good" = "green"
+"signature status unknown" = "yellow"
+"signature status bad" = "red"
+"signature status invalid" = "red"
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -78,6 +78,8 @@ if(root,
         if(git_head, label("git_head", "git_head()")),
         format_short_commit_id(commit_id),
         if(conflict, label("conflict", "conflict")),
+        if(config("ui.show-cryptographic-signatures").as_boolean(),
+          format_short_cryptographic_signature(signature)),
         if(empty, label("empty", "(empty)")),
         if(description,
           description.first_line(),
@@ -134,6 +136,8 @@ concat(
   surround("Tags     : ", "\n", tags),
   "Author   : " ++ format_detailed_signature(author) ++ "\n",
   "Committer: " ++ format_detailed_signature(committer)  ++ "\n",
+  if(config("ui.show-cryptographic-signatures").as_boolean(),
+    "Signature: " ++ format_detailed_cryptographic_signature(signature) ++ "\n"),
   "\n",
   indent("    ",
     coalesce(description, label(if(empty, "empty"), description_placeholder) ++ "\n")),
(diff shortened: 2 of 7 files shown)


bryceberger:
This uses the `template-aliases."should_xyz()"` pattern from #5290. Alternatively, it could be implemented by having the default `format_*_cryptographic_signature()` methods be empty, but that falls back to forcing the user to define a format.

cli/src/config/colors.toml
```diff
@@ -111,3 +111,10 @@
 "node current_operation" = { fg = "green", bold = true }
 "node immutable" = { fg = "bright cyan", bold = true }
 "node conflict" = { fg = "red", bold = true }
+
+"cryptographic_signature" = "yellow"
+"cryptographic_signature key" = "cyan"
+"cryptographic_signature good" = "green"
+"cryptographic_signature unknown" = "yellow"
+"cryptographic_signature bad" = "red"
+"cryptographic_signature invalid" = "red"
```
yuja, line 120: nit: `signature.<method>()` is automatically labeled as `"signature <method>"`. It's better to not deviate from this naming convention unless necessary.

yuja (commented):
cc @pylbrecht, @necauqua

cli/src/config/colors.toml
bryceberger: So `"signature display"`, `"signature key"`, and.. `"signature status (good|bad|...)"`?

necauqua:
Overall LGTM, but I don't like the default short template - just randomly having a green "good" at the end of commit header is not self-explanatory enough, like for example "conflict" is.

Personally I'm a fan of my symbols, with green checkmarks we piggy-back on association with github ui.

Or maybe at least just append " sig" to the status if people are against having more sigils idk

docs/config.md
```diff
@@ -500,7 +500,30 @@ formatted with `format_timestamp()`.
...
+
+Note that the formatting functions take an `Option<CryptographicSignature>`.
+This allows you to emit a custom message if a signature is not present, but
+will raise an error if you try to access methods on a signature that is not
+available.
+
+```toml
+[template-aliases]
+# default is "false", change to "true" to enable
+'should_show_cryptographic_signature()' = 'true'
```
necauqua, line 517: forgot to update this doc

cli/src/config/templates.toml
```diff
@@ -1,3 +1,6 @@
+[ui]
+show-cryptographic-signatures = false
```
yuja, line 2: nit: I would move this to misc.toml where other `[ui]` variables are set.

cli/src/config/templates.toml
```diff
@@ -281,9 +288,39 @@ separate(" ",
         if(commit.git_head(), label("git_head", "git_head()")),
         format_short_commit_id(commit.commit_id()),
         if(commit.conflict(), label("conflict", "conflict")),
+        if(config("ui.show-cryptographic-signatures").as_boolean(),
+          format_short_cryptographic_signature(commit.signature())),
       )
 '''
 
+# signature verification is slow, disable by default
```
yuja, line 296: nit: maybe this comment should be attached to `ui.show-cryptographic-signatures`?

cli/src/config/templates.toml
```diff
@@ -281,9 +288,39 @@ separate(" ",
...
+'format_detailed_cryptographic_signature(signature)' = '''
+if(signature,
+  separate(" ",
+    format_cryptographic_signature_status(signature),
+    label("signature display", signature.display()),
+    label("signature key", signature.key()),
```
yuja, line 302: nit: these `label()`s are redundant. `signature.display()` is transformed to `label("signature display", signature_display_value)`.

cli/src/config/templates.toml
```diff
@@ -281,9 +288,39 @@ separate(" ",
...
+        if(config("ui.show-cryptographic-signatures").as_boolean(),
+          format_short_cryptographic_signature(commit.signature())),
       )
 '''
 
+# signature verification is slow, disable by default
+'format_detailed_cryptographic_signature(signature)' = '''
+if(signature,
+  separate(" ",
+    format_cryptographic_signature_status(signature),
```
yuja, line 300: nit: might be better to use less fancy output in detailed template. just print `"[" ++ signature.status() ++ "]"`?

yuja (approved):
Thanks.

cli/src/config/templates.toml
necauqua: +1, the detailed template should be more of a plain english type of thing probably

"Good <backend if/when we ever re-add the backend field> signature by <display>"

necauqua:
Forgot to check that when making a commend about outdated doc, but the commit message is also outdated :upside_down_face:

Not like it matters that much, but I'd hate for my commit message to be forever incorrect in history

Oh wait you actually did update the config example in the commit message, but not the preceeding sentence sadly

necauqua (approved):
Hey, so this is good to go? lgtm, and Yuja approved it awhile ago

Feel free to push the merge button :)

PR #5310, opened 2025-01-09
contributing: recommend bacon over cargo-watch

bryceberger:
As of https://github.com/watchexec/cargo-watch/commit/9f27bc1c96d93ff85347619ec4a2bf72fc90fe68, `cargo-watch` is on minimal development and officially recommends `bacon` as a successor.

# Checklist

If applicable:

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -134,15 +134,13 @@ One-time setup:
 
     rustup toolchain add nightly  # wanted for 'rustfmt'
     rustup toolchain add 1.76     # also specified in Cargo.toml
-    cargo install cargo-insta
-    cargo install cargo-watch
-    cargo install cargo-nextest
+    cargo install --locked bacon
+    cargo install --locked cargo-insta
+    cargo install --locked cargo-nextest
 
 During development (adapt according to your preference):
 
-    cargo watch --ignore '.jj/**' -s \
-      'cargo clippy --workspace --all-targets \
-       && cargo +1.76 check --workspace --all-targets'
+    bacon clippy-all
     cargo +nightly fmt # Occasionally
     cargo nextest run --workspace # Occasionally
     cargo insta test --workspace --test-runner nextest # Occasionally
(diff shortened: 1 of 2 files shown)


docs/contributing.md
```diff
@@ -178,9 +176,8 @@ These are listed roughly in order of decreasing importance.
 4. Your code needs to pass `cargo clippy`. You can also
    use `cargo +nightly clippy` if you wish to see more warnings.
 
-5. You may also want to install and use `cargo-watch`. In this case, you should
-   exclude `.jj`. directory from the filesystem watcher, as it gets updated on
-   every `jj log`.
+5. You may also want to install and use `bacon`, to automatically build, check,
```
shanesveller, line 179: Nit: consider linking out to https://dystroy.org/bacon/ since the name is a lot less descriptive/compelling by itself.

Do we have insta snapshots or anything similar? If so those might warrant an extra suggestion based on https://github.com/Canop/bacon/issues/138#issuecomment-2198829660 .

flake.nix
```diff
@@ -181,7 +181,7 @@
           cargo-deny
           cargo-insta
           cargo-nextest
-          cargo-watch
+          bacon
```
shanesveller, line 184: Nit: these look like they may have been alphabetized, please consider preserving the order if so.

docs/contributing.md
```diff
@@ -135,14 +135,12 @@ One-time setup:
     rustup toolchain add nightly  # wanted for 'rustfmt'
     rustup toolchain add 1.76     # also specified in Cargo.toml
     cargo install cargo-insta
-    cargo install cargo-watch
+    cargo install bacon
```
shanesveller, line 138: Subjective and intended as discussion rather than a nit, but for software rather than dependencies I usually encourage `cargo install --locked`. It's not unambiguously better, though.

shanesveller (approved):

docs/contributing.md
bryceberger: `*.pending-snap` and `*.snap*` are in the gitignore, so they won't be picked up by bacon.

docs/contributing.md
bryceberger: `cargo-insta` and `bacon` both suggest installing with `cargo install --locked`, so I think it's worth it to add that to all of them.

`cargo-nextest` encourages using `cargo-binstall`, but since the other two don't seem to be available through binstall I don't think it's worth a note here.

thoughtpolice (approved):

PR #5311, opened 2025-01-09
config: add `--when.commands` scope

### LINKED ISSUE #5217: FR: Make the pager configurable per subcommand (e.g. enable for diff, disable for status, etc).
From the current documentation I can only globally disable the use of a pager. I can set it to `auto` so jj automatically uses it, when it deems it necessary.

Having the pager used for things like a 4-line `jj st` is annoying, while it is useful for things like e.g. a diff.

Currently the workaround seems to be to use `less -SFRX` as pager command.

If this is useful to others, it would be great to allow enabling/disabling the use of the pager per subcommand. Have a configuration option for `status.pager` to disable it, while `log.pager` enables it for logs, etc. pp.

Currently I am trying the workaround (using `less -SFRX` as pager command) recommended in #4744.

bryceberger:
Closes #5217 

Adds a new config scope resolution `--when.commands`. This is a list of space-separated commands:
```toml
--when.command = ["file"]        # matches `jj file show`, `jj file list`, etc
--when.command = ["file show"]   # matches `jj file show`, but *NOT* `jj file list`
--when.command = ["file", "log"] # matches `jj file` *OR* `jj log` (or subcommand of either)
```

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3456,6 +3456,18 @@ fn parse_args(
     Ok((matches, args))
 }
 
+fn command_name(mut matches: &ArgMatches) -> String {
+    let mut command = String::new();
+    while let Some((subcommand, new_matches)) = matches.subcommand() {
+        if !command.is_empty() {
+            command.push(' ');
+        }
+        command.push_str(subcommand);
+        matches = new_matches;
+    }
+    command
+}
+
 pub fn format_template<C: Clone>(ui: &Ui, arg: &C, template: &TemplateRenderer<C>) -> String {
     let mut output = vec![];
     template
@@ -3700,6 +3712,7 @@ impl CliRunner {
         for process_global_args_fn in self.process_global_args_fns {
             process_global_args_fn(ui, &matches)?;
         }
+        config_env.set_command_name(command_name(&matches));
 
         let maybe_workspace_loader = if let Some(path) = &args.global_args.repository {
             // TODO: maybe path should be canonicalized by WorkspaceLoader?
@@ -3712,15 +3725,16 @@ impl CliRunner {
                 .map_err(|err| map_workspace_load_error(err, Some(path)))?;
             config_env.reset_repo_path(loader.repo_path());
             config_env.reload_repo_config(&mut raw_config)?;
-            config = config_env.resolve_config(&raw_config)?;
-            migrate_config(&mut config)?;
             Ok(loader)
         } else {
             maybe_cwd_workspace_loader
         };
 
-        // Apply workspace configs and --config arguments.
+        // Apply workspace configs, --config arguments, and --when.commands.
+        config = config_env.resolve_config(&raw_config)?;
+        migrate_config(&mut config)?;
         ui.reset(&config)?;
+
         // Print only the last migration messages to omit duplicates.
         for desc in &last_config_migration_descriptions {
             writeln!(ui.warning_default(), "Deprecated config: {desc}")?;
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -263,6 +263,7 @@ pub struct ConfigEnv {
     repo_path: Option<PathBuf>,
     user_config_path: ConfigPath,
     repo_config_path: ConfigPath,
+    command: Option<String>,
 }
 
 impl ConfigEnv {
@@ -281,7 +282,12 @@ impl ConfigEnv {
             repo_path: None,
             user_config_path: env.resolve()?,
             repo_config_path: ConfigPath::Unavailable,
+            command: None,
         })
+    }
+
+    pub fn set_command_name(&mut self, command: String) {
+        self.command = Some(command);
     }
 
     /// Returns a path to the user-specific config file or directory.
@@ -398,6 +404,7 @@ impl ConfigEnv {
         let context = ConfigResolutionContext {
             home_dir: self.home_dir.as_deref(),
             repo_path: self.repo_path.as_deref(),
+            command: self.command.as_deref(),
         };
         jj_lib::config::resolve(config.as_ref(), &context)
     }
@@ -1369,6 +1376,7 @@ mod tests {
                 repo_path: None,
                 user_config_path: env.resolve()?,
                 repo_config_path: ConfigPath::Unavailable,
+                command: None,
             })
         }
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -1371,6 +1371,9 @@ You can conditionally enable config variables by using `--when` and
 `[[--scope]]` tables. Variables defined in `[[--scope]]` tables are expanded to
 the root table. `--when` specifies the condition to enable the scope table.
 
+If no conditions are specified, table is always enabled. If multiple conditions
+are specified, the intersection is used.
+
 ```toml
 [user]
 name = "YOUR NAME"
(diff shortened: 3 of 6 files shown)


bryceberger:
Unsure how to update `config-schema.json` for this. The current entry for pager just says `"type": "string"`, which isn't true.

Also unsure if / where there are tests for pagers. I was unable to find any on a cursory glance.

yuja:
> ```toml
> [ui]
> pager.status = ":none"
> pager.log = { command = ["less", "-FRX"], env = { LESSCHARSET = "utf-8" } }
> pager.diff = ["delta", "..."]
> pager.default = "..."
> ```

I think it's better to not overuse ":<type>" syntax. How about adding a separate table for per-table pager configuration?

```toml
[ui]
paginate = "auto"  # default
pager = "..."  # default

[<table-for-per-command-pager-options>]
status.paginate = "never"
diff.pager = ["delta", "..."]

# or generic per-command table like this?
[commands.status]
ui.paginate = "never"

# or extend the conditional table?
[[--scope]]
--when.commands = ["status"]
ui.paginate = "never"
```

> The full name of the highest level subcommand is used, i.e. `jj operation diff` uses `ui.pager.operation`.

I think "file show" and "file list" are quite different, and the user might want to enable pager only for one of them.

bryceberger:
I'm intrigued by extending the `[[--scope]]` rules.

The rule for `--when.repositories = ["a", "b"]` is to match `"a" | "b"` --- would have to put some thought into how to match `operation`, `operation diff`, and `diff` separately. Maybe `"operation.diff"`?

Currently, `jj_lib::config_resolver::resolve()` throws away the layer if its `--when` doesn't match. Thinking of changing `ScopeCondition::matches()` to return `Option<bool>`, with `None` meaning "not enough information to resolve", i.e. resolving `--when.repository` during `cli_util::run_internal` but waiting until `commands::run_command` to resolve `--when.commands`.

yuja:
> The rule for `--when.repositories = ["a", "b"]` is to match `"a" | "b"` --- would have to put some thought into how to match `operation`, `operation diff`, and `diff` separately. Maybe `"operation.diff"`?

or `"operation diff"`? The `[colors]` table uses space-separated string.

> Currently, `jj_lib::config_resolver::resolve()` throws away the layer if its `--when` doesn't match. Thinking of changing `ScopeCondition::matches()` to return `Option<bool>`, with `None` meaning "not enough information to resolve", i.e. resolving `--when.repository` during `cli_util::run_internal` but waiting until `commands::run_command` to resolve `--when.commands`.

`run_command()` calls `resolve()` multiple times, so I think it's okay to omit `--when.commands` conditionals until the command name becomes available. That's how `--repository` path is resolved.

ilyagr:
Just saw this, nice! 🎉 

FYI, I'm planning to add config for the built-in pager. My original thought was to simply create a `ui.builtin-pager` table. I'm not sure how it'll look if we make it configurable per-command, but we could figure it out later.

yuja:
> simply create a `ui.builtin-pager` table.

nit: might be better to include "streampager" in the key (e.g. `stream-pager`, `builtin-stream-pager`, `builtin-pager-sp`) so we can freely switch the underlying implementations in future.

> I'm not sure how it'll look if we make it configurable per-command,

The current idea is to extend the scoped table, so any parameters (except for `[aliases]`) can be set per-command.

ilyagr:
> > simply create a `ui.builtin-pager` table.
> 
> nit: might be better to include "streampager" in the key (e.g. `stream-pager`, `builtin-stream-pager`, `builtin-pager-sp`) so we can freely switch the underlying implementations in future.

I went back and forth on this. I went into the direction of making the options more generic, see https://github.com/jj-vcs/jj/pull/5415 . Your suggestion might still be better, I'm not sure. If we go with that, perhaps `:builtin` should be an alias for `:builtin-sp`.

> > I'm not sure how it'll look if we make it configurable per-command,
> 
> The current idea is to extend the scoped table, so any parameters (except for `[aliases]`) can be set per-command.

That sounds great!

bryceberger:
Completely changed the implementation, as reflected in the title. Now implements `--when.commands = [...]`.

Also changed how config resolution works --- a condition can either be "resolved true", "resolved false", or "unknown". If any field is "unknown", the entire expression is unknown, and kept around. If any field is "resolved false", the entire expression is false, and discarded. Otherwise, added to current config.

This means that if a config layer still contains a `--when` key after resolution, it was resolved to "unknown". These layers are skipped in `config::get_merged_item()`.

yuja:
> Also changed how config resolution works --- a condition can either be "resolved true", "resolved false", or "unknown". If any field is "unknown", the entire expression is unknown, and kept around. If any field is "resolved false", the entire expression is false, and discarded. Otherwise, added to current config.

Do we really need this? As I said, config layers are resolved multiple times in `run_internal()`, and the last resolution wins. If conditional variables have `--when.commands`, I don't think they should be disabled until the command name becomes available.

cli/src/cli_util.rs
```diff
@@ -3701,6 +3701,23 @@ impl CliRunner {
...
+                    subcommand_combined.push_str(subcommand);
+                    subcommand_matches = new_matches;
+                }
+                config = config_env
+                    .resolve_config_with_command(&raw_config, Some(&subcommand_combined))?;
+                migrate_config(&mut config)?;
+                ui.reset(&config)?;
+                Some(subcommand_combined)
+            } else {
+                None
```
yuja, line 3718: nit: perhaps, actual command name can be set to `""` if no subcommands exist. `None` means the command name is unknown, whereas the command is known to be unspecified here.

cli/src/config.rs
```diff
@@ -395,9 +395,18 @@ impl ConfigEnv {
     /// Resolves conditional scopes within the current environment. Returns new
     /// resolved config.
     pub fn resolve_config(&self, config: &RawConfig) -> Result<StackedConfig, ConfigGetError> {
+        self.resolve_config_with_command(config, None)
+    }
+
+    pub fn resolve_config_with_command(
```
yuja, line 401: set the command name in `ConfigEnv`, and do `resolve_config()`. `config_env` can be reused later in order to resolve settings for cloned/inited workspace.

lib/src/config_resolver.rs
```diff
@@ -98,11 +117,32 @@ fn expand_home(path: &Path, home_dir: Option<&Path>) -> Result<Option<PathBuf>,
...
+
+fn matches_command(candidates: Option<&[String]>, actual: Option<&str>) -> Option<bool> {
+    match (candidates, actual) {
+        // empty condition never matches
+        (Some([]), _) => Some(false),
+        (Some(candidates), Some(actual)) => {
+            let actual = actual.split(' ');
+            let any_match = candidates.iter().any(|candidate| {
+                let candidate = candidate.split(' ');
+                candidate.zip_longest(actual.clone()).all(|e| match e {
```
yuja, line 136: Maybe `.strip_prefix()` and check the first character?

bryceberger:
> Do we really need this? As I said, config layers are resolved multiple times in `run_internal()`, and the last resolution wins. If conditional variables have `--when.commands`, I ~don't~ think they should be disabled until the command name becomes available.

... oooh, the `RawConfig` is kept unchanged through resolution. I had it in my mind that `pop_scope_condition()` would destructively remove the table, didn't realize it was clone-on-write.

lib/src/config_resolver.rs
bryceberger: Ah, that is cleaner. Not sure why you would need to check the first character --- `"file show".strip_prefix("file") == Some(" show")`, `"file show".strip_prefix("file show") == Some("")`.

bryceberger: Oh, so `jj abcdef` doesn't accidentally match `--when.commands = ["abc"]`.

lib/src/config_resolver.rs
```diff
@@ -106,6 +119,18 @@ fn matches_path_prefix(candidates: Option<&[PathBuf]>, actual: Option<&Path>) ->
     }
 }
 
+fn matches_command(candidates: Option<&[String]>, actual: Option<&str>) -> bool {
+    match (candidates, actual) {
+        // empty condition never matches
+        (Some([]), _) => false,
```
yuja, line 125: nit: I don't think optimization is needed for empty condition.

cli/src/cli_util.rs
```diff
@@ -3701,6 +3701,25 @@ impl CliRunner {
...
         }
 
+        config_env.set_command(
+            if let Some((subcommand, mut subcommand_matches)) = matches.subcommand() {
+                let mut subcommand_combined = String::from(subcommand);
+                while let Some((subcommand, new_matches)) = subcommand_matches.subcommand() {
+                    subcommand_combined.push(' ');
+                    subcommand_combined.push_str(subcommand);
+                    subcommand_matches = new_matches;
+                }
```
yuja, line 3711: nit: I would extract this part to a function that returns string command name.

cli/src/cli_util.rs
```diff
@@ -3701,6 +3701,25 @@ impl CliRunner {
...
+                config = config_env.resolve_config(&raw_config)?;
+                migrate_config(&mut config)?;
+                ui.reset(&config)?;
```
yuja, line 3714: nit: deduplicate this. Perhaps, `ui.reset()` can be omitted as we do that later.

cli/src/config.rs
```diff
@@ -281,9 +282,14 @@ impl ConfigEnv {
             repo_path: None,
             user_config_path: env.resolve()?,
             repo_config_path: ConfigPath::Unavailable,
+            command: None,
         })
     }
 
+    pub fn set_command(&mut self, command: Option<String>) {
```
yuja, line 289: nits: 
* `set_command_name()` for clarity? (Perhaps, context variable can also be renamed.)
* just take `String`? It's unlikely that we need to clear the command name.

docs/config.md
```diff
@@ -1371,6 +1371,9 @@ You can conditionally enable config variables by using `--when` and
 `[[--scope]]` tables. Variables defined in `[[--scope]]` tables are expanded to
 the root table. `--when` specifies the condition to enable the scope table.
 
+If no conditions are specified, table is always enabled. If multiple conditions
+are specified, the intersection is used.
+
 ```toml
 [user]
```
yuja, line 1378: nit: can you add pager example here?

martinvonz:
I just noticed that [the documentation](https://jj-vcs.github.io/jj/latest/config/#conditional-variables) has not been updated. Are you able to fix that, @bryceberger?

ilyagr:
There [prerelease docs](https://jj-vcs.github.io/jj/prerelease/config/#conditional-variables) have some information. I'm not sure whether something is missing in there as well.

martinvonz:
No, that looks good. Sorry, I don't know why I went to the 0.25 docs.

PR #5315, opened 2025-01-09
chore(deps): update git2 to use libgit2 1.9

chenrui333:
# Checklist

If applicable:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -913,9 +913,9 @@ checksum = "40ecd4077b5ae9fd2e9e169b102c6c330d0605168eb0e8bf79952b256dbefffd"
 
 [[package]]
 name = "git2"
-version = "0.19.0"
+version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b903b73e45dc0c6c596f2d37eccece7c1c8bb6e4407b001096387c63d0d93724"
+checksum = "3fda788993cc341f69012feba8bf45c0ba4f3291fcc08e214b4d5a7332d88aff"
 dependencies = [
  "bitflags",
  "libc",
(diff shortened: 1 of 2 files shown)


thoughtpolice:
There isn't a build of libgit2 1.9 for Nix yet, so I'm not sure we want to pull the trigger on this just yet. Or we need to update the Flake to build libgit2 out-of-band.

chenrui333:
I did not know that nix did not have a libgit2 1.9 build yet, but yeah, this PR can certainly wait for libgit2 1.9 on the nix side.

alerque:
For the flake you should be able to just temporarily the `LIBGIT2_NO_VENDOR=1` so that the *libgit2-sys* crate vendors the build, then when nixpkgs can be update to something with *libgit2-1.9* you can go back to forcing it to not vendor it.

PhilipMetzger:
Can you also please adhere to the commit style outlined by this document here: https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#code-reviews. Thanks, my suggestion is either `cargo:` like dependabot or `deps:` which you already have.

emilazy:
> There isn't a build of libgit2 1.9 for Nix yet, so I'm not sure we want to pull the trigger on this just yet. Or we need to update the Flake to build libgit2 out-of-band.

There’s a bit of a catch‐22 here: https://github.com/NixOS/nixpkgs/pull/373096 updates `libgit2` to 1.9.0, but of course by the time it hits the channels it’ll break our Jujutsu package.

I guess we can apply this patch on the Nixpkgs end for now to break the cycle.

alerque:
The `git2` crate ever only supporting one release cycle at a time is really rather a pain.

As the Arch Linux packager working through the libgit2-1.9 todo list, I can say we're probably going to have to provide a 1.8 compat package for a while. Nix may end up having to do the same, there are a few upstream projects that are not easily patched one direction or the other. Jujutsu was pretty easy (we just back ported this bump and everything worked) but some other tools are not playing so nice.

emilazy:
Continued as https://github.com/jj-vcs/jj/pull/5938.

chenrui333:
I am a bit lost, isn't #5938 the same PR as this one?

chenrui333:
any way, the other PR got merged, close this one now.

emilazy:
Sorry about that! This seemed to have been blocked on https://github.com/jj-vcs/jj/pull/5315#issuecomment-2590565096 after we sorted things out on the Nixpkgs end and I didn’t want to bother you about it :)

PR #5316, opened 2025-01-09
cli: op log: add `--bookmark` option to filter operations by bookmark

bnjmnt4n:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/operation/log.rs
+++ b/cli/src/commands/operation/log.rs
@@ -130,7 +130,7 @@ fn do_op_log(
                 ui,
                 &language,
                 &get_node_template(graph_style, settings)?,
-                OperationTemplateLanguage::wrap_operation,
+                OperationTemplateLanguage::wrap_operation_opt,
             )?
             .labeled("node");
     }
@@ -212,7 +212,7 @@ fn do_op_log(
                 let mut formatter = ui.new_formatter(&mut buffer);
                 show(ui, formatter.as_mut(), &op, &within_graph)?;
             }
-            let node_symbol = format_template(ui, &op, &op_node_template);
+            let node_symbol = format_template(ui, &Some(op), &op_node_template);
             graph.add_node(
                 op.id(),
                 &edges,
--- a/cli/src/commands/operation/log.rs
+++ b/cli/src/commands/operation/log.rs
@@ -12,15 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
 use std::slice;
 
+use clap_complete::ArgValueCandidates;
+use indexmap::IndexMap;
 use itertools::Itertools as _;
 use jj_lib::config::ConfigGetError;
 use jj_lib::config::ConfigGetResultExt as _;
+use jj_lib::op_store::OperationId;
 use jj_lib::op_walk;
 use jj_lib::operation::Operation;
 use jj_lib::repo::RepoLoader;
 use jj_lib::settings::UserSettings;
+use jj_lib::str_util::StringPattern;
+use jj_lib::view::View;
 
 use super::diff::show_op_diff;
 use crate::cli_util::format_template;
@@ -29,6 +35,7 @@ use crate::cli_util::LogContentFormat;
 use crate::cli_util::WorkspaceCommandEnvironment;
 use crate::command_error::CommandError;
 use crate::commit_templater::CommitTemplateLanguage;
+use crate::complete;
 use crate::diff_util::diff_formats_for_log;
 use crate::diff_util::DiffFormatArgs;
 use crate::diff_util::DiffRenderer;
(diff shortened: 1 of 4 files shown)


bnjmnt4n:
I've wrote a preliminary version of this feature, which allows filtering op log entries by bookmark. Unfortunately it's not very performant, because it requires doing a toposort of the whole op log before filtering out irrelevant entries. In contrast, the plain (unfiltered) op log has a lazy iterator to go through linear portions of the graph, which I think would be large portions of the graph for most.

I'm not sure if there is a way to correctly construct the graph with elided revisions in a lazy manner? My understanding (do correct me if I'm wrong) is that the regular `jj log` is able to construct the graph lazily due to it already knowing the items in the "input set", as well as the use of the index and generation count.

I'm also not that familiar with graphs, so perhaps there are better algorithms for this?

yuja:
Maybe we can apply the same workaround as the lazy iterator? It just splits the graph into linear and non-linear parts, and processes non-linear part eagerly.

FWIW, we'll need opset language at some point, but adding `--bookmark` flag seems okay until then.

bnjmnt4n:
> Maybe we can apply the same workaround as the lazy iterator? It just splits the graph into linear and non-linear parts, and processes non-linear part eagerly.

I was thinking the same, but I'll probably need a bit more time to go through the code to understand and adapt the algorithm.

> FWIW, we'll need opset language at some point, but adding `--bookmark` flag seems okay until then.

I also wanted to add the ability to filter by change ID, so perhaps the opset language might come in sooner rather than later?

yuja:
> I was thinking the same, but I'll probably need a bit more time to go through the code to understand and adapt the algorithm.

Since op log is usually liner, it's probably okay to not implement `skip_transitive_edges`. It might help reuse the existing code by somehow pre-processing the input op -> parents chains.

> I also wanted to add the ability to filter by change ID, so perhaps the opset language might come in sooner rather than later?

It might be also okay to add minimal opset parser (something like `strip_prefix("bookmarks(") ..`.)

bnjmnt4n:
Is the eventual intention to get something like `jj op log --op "@ | id1..id2` working?

> Since op log is usually linear, it's probably okay to not implement `skip_transitive_edges`. It might help reuse the existing code by somehow pre-processing the input op -> parents chains.

This is referring to the code in `RevsetGraphWalk` right? Without an op store index, I don't think it's possible to easily reuse that code because it depends on the input being already sorted?

Unless we intend to create an operation store index, I was thinking we could attempt to resolve any resolvable ops in the opset first (apart from `bookmarks(x)` for now), then iterate through the op graph and filter ops if they are not in the input set and if the bookmarks predicate doesn't apply. I think I might be able to convert the filtering to a lazy method with some time.

yuja:
> Is the eventual intention to get something like `jj op log --op "@ | id1..id2` working?

Yes, it's nice if `op log` supports arbitrary range/filter query.

> > Since op log is usually linear, it's probably okay to not implement `skip_transitive_edges`. It might help reuse the existing code by somehow pre-processing the input op -> parents chains.
> 
> This is referring to the code in `RevsetGraphWalk` right? Without an op store index, I don't think it's possible to easily reuse that code because it depends on the input being already sorted?

I meant `dag_walk::topo_order_reverse_lazy_ok()` might be reusable by filtering the inputs. To achieve `skip_transitive_edges = true`, the graph iterator would have to know whether the edge is direct or not, but that wouldn't be needed for op log where the graph is usually linear so it's rare that filtered graph had redundant edges.

> Unless we intend to create an operation store index, I was thinking we could attempt to resolve any resolvable ops in the opset first (apart from `bookmarks(x)` for now), then iterate through the op graph and filter ops if they are not in the input set and if the bookmarks predicate doesn't apply. I think I might be able to convert the filtering to a lazy method with some time.

I think most part of "opset" expression will be transformed to `Fn(&Opration) -> bool` function. Some common range expression (like `::<head> & ..`) might be extracted to optimize range query.

ilyagr:
I think that generalizing this to filtering by an arbitrary revset instead just by a bookmark might be worthwhile (assuming it's not much harder). Two other revsets that come to mind are `@`, `bookmark@remote`, and a change id.

I can't immediately come up with a useful revset that will often expand to multiple commits (the best I can think of is debugging a conflicted branch/divergent change id/etc), but that could be treated by diffing the set of commit id. Or it could be unsupported for now.

bnjmnt4n:
> (Not necessarily for this PR) I think that generalizing this to filtering by an arbitrary revset instead just by a bookmark might be worthwhile (assuming it's not much harder). Two other revsets that come to mind are `@`, `bookmark@remote`, and a change id, evaluated at each operation.

I would definitely prefer this. I think this would be blocked on implementing the opset language.

For now, I think this is blocked on implementing a lazy version of the graph iterator. I'm not sure when I'll have the time to come back to visit that.

PR #5328, opened 2025-01-10
templater: add config(name) function

yuja:
Do we want this?

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj util mangen` is replaced with `jj util install-man-pages`, which can
   install man pages for all `jj` subcommands to a given path.
 
+* In `jj config list` template, `value` is now typed as `ConfigValue`, not as
+  `String` serialized in TOML syntax.
+
 ### Deprecations
 
 ### New features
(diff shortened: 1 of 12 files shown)


bnjmnt4n (approved):
I think this is useful, and will definitely use it in my own configuration.

PhilipMetzger (approved):
I also much prefer this variant.

jakobhellermann:
I think it would be valuable to have a variant of this which does not fail in case the config value does not exist.

I wonder if it should be handled via a general `try(config("custom.config"))` function that evaluates to null on any error (similar to how `present(x)` is used in revsets), or maybe just `config(name: string, allow_missing: bool)`.

bryceberger:
Regarding #5328 and #5308, is there a consensus on where the `"should_xyz"` config values go? Probably not `templates.should_xyz`, because that could be confusing with `jj _ -T should_xyz`. Maybe `ui.templates`?

martinvonz:
> Regarding #5290 and #5308, is there a consensus on where the `"should_xyz"` config values go? Probably not `templates.should_xyz`, because that could be confusing with `jj _ -T should_xyz`. Maybe `ui.templates`?

I think even `ui.templates` is a bit confusing if the value is not a template. Maybe something like `ui.show_cryptographic_signature`?

yuja:
> I think it would be valuable to have a variant of this which does not fail in case the config value does not exist.

Do you have any use case in mind (something like getting value from free-form table e.g. `revset-aliases`)?

My current idea is that we should define all config values by default. If it's user addition, they can easily add one to their config. `config(name)` could return `Option<ConfigValue>`, but I'm not sure if that's good. It might be confusing if boolean value is expected.

PR #5329, opened 2025-01-10
cli: op log: add `--reversed` option

### LINKED ISSUE #4190: FR: Add `--reversed` option to `op log`
Align `op log` output with `log` output.

Add option `--reversed` to `op log` (same as `log`).

n/a

**Additional context**
n/a

bnjmnt4n:
Closes #4190.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/operation/log.rs
+++ b/cli/src/commands/operation/log.rs
@@ -17,6 +17,7 @@ use std::slice;
 use itertools::Itertools as _;
 use jj_lib::config::ConfigGetError;
 use jj_lib::config::ConfigGetResultExt as _;
+use jj_lib::graph::GraphEdge;
 use jj_lib::op_walk;
 use jj_lib::operation::Operation;
 use jj_lib::repo::RepoLoader;
@@ -34,7 +35,6 @@ use crate::diff_util::DiffFormatArgs;
 use crate::diff_util::DiffRenderer;
 use crate::formatter::Formatter;
 use crate::graphlog::get_graphlog;
-use crate::graphlog::Edge;
 use crate::graphlog::GraphStyle;
 use crate::operation_templater::OperationTemplateLanguage;
 use crate::ui::Ui;
@@ -198,7 +198,7 @@ fn do_op_log(
             let op = op?;
             let mut edges = vec![];
             for id in op.parent_ids() {
-                edges.push(Edge::Direct(id.clone()));
+                edges.push(GraphEdge::direct(id.clone()));
             }
             let mut buffer = vec![];
             let within_graph = with_content_format.sub_width(graph.width(op.id(), &edges));
--- a/cli/src/commands/operation/log.rs
+++ b/cli/src/commands/operation/log.rs
@@ -17,7 +17,9 @@ use std::slice;
 use itertools::Itertools as _;
 use jj_lib::config::ConfigGetError;
 use jj_lib::config::ConfigGetResultExt as _;
+use jj_lib::graph::reverse_graph;
 use jj_lib::graph::GraphEdge;
+use jj_lib::op_store::OpStoreError;
 use jj_lib::op_walk;
 use jj_lib::operation::Operation;
 use jj_lib::repo::RepoLoader;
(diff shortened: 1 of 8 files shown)


cli/src/commands/operation/log.rs
```diff
@@ -190,18 +196,29 @@ fn do_op_log(
...
         let mut raw_output = formatter.raw()?;
         let mut graph = get_graphlog(graph_style, raw_output.as_mut());
-        for op in iter {
+        let iter = iter.map(|op| -> Result<_, OpStoreError> {
             let op = op?;
-            let mut edges = vec![];
-            for id in op.parent_ids() {
-                edges.push(Edge::Direct(id.clone()));
-            }
+            let edges = op.parents().map_ok(GraphEdge::direct).try_collect()?;
```
bnjmnt4n, line 206: Loading `op.parents()` here isn't really required. I'm wondering if we should update `reverse_graph` to add an `id_fn`? Although we will still end up having to load most operations in most cases, so perhaps it's not worth optimizing.

cli/src/commands/operation/log.rs
```diff
@@ -49,6 +52,9 @@ pub struct OperationLogArgs {
     /// Limit number of operations to show
```
yuja, line 52: nit: Can you also update the doc of `--limit`?
```
    /// Applied after operations are reordered.
```

cli/src/commands/operation/log.rs
yuja: Yeah, maybe that's good idea, but I also agree that the extra I/O wouldn't matter in practice.

cli/src/commands/operation/log.rs
```diff
@@ -190,18 +196,29 @@ fn do_op_log(
...
+        let iter_nodes: Box<dyn Iterator<Item = _>> = if args.reversed {
+            Box::new(reverse_graph(iter)?.into_iter().map(Ok))
+        } else {
+            Box::new(iter)
+        };
+        for node in iter_nodes.take(limit) {
+            let (op, edges) = node?;
+            let graphlog_edges = edges
+                .into_iter()
+                .map(|e| convert_edge_type(e).map(|e| e.id().clone()))
```
yuja, line 218: nit: If you're interested, can you take a look at https://github.com/jj-vcs/jj/pull/5170#discussion_r1898770980 ?

cli/tests/test_operations.rs
```diff
@@ -203,6 +203,106 @@ fn test_op_log_no_graph() {
...
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+    test_env.jj_cmd_ok(&repo_path, &["describe", "-m", "description 0"]);
+
+    let stdout = test_env.jj_cmd_success(
+        &repo_path,
+        &[
+            "op",
+            "log",
+            "--config=template-aliases.'format_time_range(x)'=x",
```
yuja, line 218: nit: I think this alias isn't needed anymore. Relative timestamp is disabled globally in tests.

yuja (approved):
Thanks.

cli/src/commands/operation/log.rs
bnjmnt4n: So in this case, I should just remove the `jj_cli::graphlog::Edge` type and manually map from `jj_lib::graph::GraphEdge` to `renderdag::Ancestor` for rendering?

yuja: I think all users of `graphlog::Edge` type can be migrated to `GraphEdge`. iirc, it existed because we had our own graph renderer based on `graphlog::Edge`.

bnjmnt4n: @yuja I've updated the PR to remove `graphlog::Edge`. Could you see if there's anything I missed?

cli/src/graphlog.rs
```diff
@@ -25,60 +24,28 @@ use renderdag::Ancestor;
...
 }
 
-impl<K: Clone> From<&Edge<K>> for Ancestor<K> {
-    fn from(e: &Edge<K>) -> Self {
-        match e {
-            Edge::Direct(target) => Ancestor::Parent(target.clone()),
-            Edge::Indirect(target) => Ancestor::Ancestor(target.clone()),
-            Edge::Missing => Ancestor::Anonymous,
-        }
+pub fn convert_graph_edge_into_ancestor<K: Clone>(e: &GraphEdge<K>) -> Ancestor<K> {
```
yuja, line 44: nit: remove `pub` since the `Ancestor` type is kinda implementation detail?

yuja (approved):
Looks good, thanks!

PR #5330, opened 2025-01-10
op diff: remove redundant "Change {change_id}" from changed commits graph

yuja:
# Checklist

If applicable:


bnjmnt4n (approved):
I don't really have a strong opinion on whether to keep the header or not. Perhaps others would have more thoughts?

martinvonz:
> I don't really have a strong opinion on whether to keep the header or not. Perhaps others would have more thoughts?

I don't feel strongly either way. We can try removing it and we'll see if people find it confusing.

PR #5337, opened 2025-01-10
cli: autocomplete: complete `--config` with possible configuration keys

jakobhellermann:
image

If applicable:

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -390,6 +390,7 @@ fn config_keys_rec(
     properties: &serde_json::Map<String, serde_json::Value>,
     acc: &mut Vec<CompletionCandidate>,
     only_leaves: bool,
+    suffix: &str,
 ) {
     for (key, value) in properties {
         let mut prefix = prefix.clone();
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj util mangen` is replaced with `jj util install-man-pages`, which can
   install man pages for all `jj` subcommands to a given path.
 
+* Add dynamic autocompletion for the `jj --config key=value` argument
```
yuja, line 48: nit: omit the changelog entry? I don't think changelog entry is needed for improvement of completion. User doesn't have to know that beforehand.

cli/src/complete.rs
```diff
@@ -409,20 +410,23 @@ fn config_keys_rec(
...
+                    escaped_key.push('=');
+                }
                 acc.push(CompletionCandidate::new(escaped_key).help(help));
             }
         }
     }
 }
 
-fn config_keys_impl(only_leaves: bool) -> Vec<CompletionCandidate> {
+fn config_keys_impl(only_leaves: bool, include_eq: bool) -> Vec<CompletionCandidate> {
```
yuja, line 429: nit: (at your option) `suffix: &str` could be added instead (to avoid `(bool, bool)`), but that's not important. One minor good thing is that we can use `format!("{name}{suffix}")` to construct an escaped key.

yuja (approved):
Thanks.

cli/src/complete.rs
jakobhellermann: good idea

jakobhellermann:
I've addressed the comments and added a test. So I can just go ahead and enable the automerge, right? Since the PR has been approved and the changes afterwards were minor (test and clippy fix).

PR #5341, opened 2025-01-10
cli: complete: fix stderr during autocomplete of revsets

jakobhellermann:
When running `cmd.spawn()` rust will by default inherit the stderr of the parent, so `jj log test<TAB>`, would print `There is no jj repo in "."` into the prompt if you aren't in a repository.


thoughtpolice (approved):
I had noticed this occasionally; thanks! I'm not sure if we need a test for this tbh, but I'll let someone else give their thoughts.

martinvonz:
We have tests in [cli/tests/test_completion.rs](https://github.com/jj-vcs/jj/blob/main/cli/tests/test_completion.rs). I haven't checked but it seems like it shouldn't be hard to add a test for this case.

PR #5342, opened 2025-01-10
docs: Explain what elided revisions are

jennings:
A common question is asking how to show elided revisions. This FAQ explains why revisions are elided and how to display them.

# Checklist

If applicable:

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -33,6 +33,42 @@ revision visible again.
 
 See [revsets] and [templates] for further guidance.
 
+### What are elided revisions in the output of `jj log`? How can I display them?
+
+"Elided revisions" appears in the log when one revision descends from another,
+both are in the revset, but the revisions connecting them are _not_ in the
+revset.
+
+For example, suppose you log the revset `tyl|mus` which contains exactly two
+revisions:
+
+```sh
+$  jj log -r 'tyl|mus'
+○  musnqzvt me@example.com 1 minute ago 9a09f8a5
+│  Revision C
+~  (elided revisions)
+○  tylynnzk me@example.com 1 minute ago f26967c8
+│  Revision A
+```
+
+Only the two revisions in the revset are displayed. The text "(elided
+revisions)" is shown to indicate that `musnqzvt` descends from `tylynnzk`, but
+the nodes connecting them are not in the revset.
+
+To view the elided revisions, change the [revset expression](revsets.md) so it
+includes the connecting revisions.  The `connected()` revset function does
+exactly this:
+
+```sh
+$ jj log -r 'connected(tyl|mus)'
+○  musnqzvt me@example.com 43 seconds ago 9a09f8a5
+│  Revision C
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -33,6 +33,41 @@ revision visible again.
...
+~  (elided revisions)
+○  tylynnzk me@example.com 1 minute ago f26967c8
+│  Revision A
+```
+
+Only the two revisions in the revset are displayed. "Elided revisions" is shown
+to indicate that `musnqzvt` descends from `tylynnzk`, but the nodes connecting
+them are not in the revset.
+
+To view the elided revisions, change the [revset expression](./revsets.md) so it
```
PhilipMetzger, line 58: Nit: the link does not need a `./`.

docs/FAQ.md
```diff
@@ -33,6 +33,41 @@ revision visible again.
...
+includes the connecting revisions.  One option is `tyl::mus`:
```
PhilipMetzger, line 59: Nit: It may also be valuable to mention `ty..mus` as another example.

docs/FAQ.md
```diff
@@ -33,6 +33,41 @@ revision visible again.
...
+○  musnqzvt me@example.com 1 minute ago 9a09f8a5
+│  Revision C
+~  (elided revisions)
+○  tylynnzk me@example.com 1 minute ago f26967c8
+│  Revision A
+```
+
+Only the two revisions in the revset are displayed. "Elided revisions" is shown
+to indicate that `musnqzvt` descends from `tylynnzk`, but the nodes connecting
+them are not in the revset.
```
PhilipMetzger, line 56: Nit: I think it also is shown when the distance between is to large in cases where 1000 commits are between both revisions.

PhilipMetzger (approved):
LG with minor nits.

docs/FAQ.md
jennings: That would then exclude `tyl`, so the output would change in _two_ ways. I think that would be distracting.

However, maybe I should change it to `connected(tyl|mus)`, because it is the same expression as before, just wrapped in something that un-elides anything that was elided.

docs/FAQ.md
jennings: Are you sure that's true? In the microsoft/winget-pkgs repository, I have 170,000 commits, but "elided" doesn't appear in the output at all .

```powershell
PS> jj log -T '1 ++ "\n"' -r 'all()' | measure -Line

 Lines Words Characters Property
 ----- ----- ---------- --------
170887

PS> jj log -T '1 ++ "\n"' -r 'all()' | sls elided
PS> jj log -T '1 ++ "\n"' -r 'root() | master' | sls elided

~  (elided revisions)

PS>
```

PhilipMetzger: Sorry, I didn't remember it correctly as they were introduced for it because the default log revset contains a variaton of `root() | @`. And you cannot see it when using `all()` as it disables log elision by forcing all nodes to be rendered anyway.

PR #5347, opened 2025-01-12
docs: Expose `config-schema.json` in the docs site

rwjblue:
When landed and published, this will expose the config file as:

https://jj-vcs.github.io/jj/latest/config-schema.json

Exposing the schema like that will allow users to reference in their `~/.config/jj/config.toml` like:

```toml
"$schema" = 'https://jj-vcs.github.io/jj/latest/config-schema.json'
```

At which point any user with a configured LSP for TOML files will get inline documentation, suggestions on valid keys, &c.

### Examples

Some example screenshots of this working in my local NeoVim setup (using a local `mkdocs serve` of these changes), but the same works in any editor with an LSP setup for TOML files.

Screenshot 2025-01-11 at 10 04 32 PM
Screenshot 2025-01-11 at 10 06 41 PM
Screenshot 2025-01-11 at 10 05 32 PM

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   useful for defining disabled tools in user configuration that can be enabled
   in individual repositories with one config setting.
 
+* The Jujutsu documentation site now publishes a schema for the official
+  configuration file, which can be integrated into your editor for autocomplete,
+  inline errors, and more. Please [see the documentation](/docs/config.md#json-schema-support)
+  for more on this.
+
 ### Fixed bugs
 
 * Fixed diff selection by external tools with `jj split`/`commit -i FILESETS`.
--- a/docs/config-schema.json
+++ b/docs/config-schema.json
@@ -1,0 +1,1 @@
+../cli/src/config-schema.json
--- a/docs/config.md
+++ b/docs/config.md
@@ -23,12 +23,17 @@ These are listed in the order they are loaded; the settings from earlier items
 in the list are overridden by the settings from later items if they disagree.
 Every type of config except for the built-in settings is optional.
 
+You can enable JSON Schema validation in your editor by adding a `$schema`
+reference at the top of your TOML config files. See [JSON Schema
+Support] for details.
+
 See the [TOML site] and the [syntax guide] for a detailed description of the
 syntax. We cover some of the basics below.
 
 [the user config file]: #user-config-file
 [TOML site]: https://toml.io/en/
 [syntax guide]: https://toml.io/en/v1.0.0
+[JSON Schema Support]: #json-schema-support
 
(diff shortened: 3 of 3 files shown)


rwjblue:
There are a few other ways to do this (by adding other mkdocss plugins that would copy the file into place after the build), but this is the simplest path forward and I figured it'd give a good starting point for conversation. Let me know if we don't want a symlink in the repo like this, and I can work another version.

rwjblue:
OK, just updated and marked as ready for review. I added a brief blurb to the CHANGELOG and also added a section to `docs/config.md`.

CHANGELOG.md
```diff
@@ -77,6 +77,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New `git.sign-on-push` config option to automatically sign commits which are being
   pushed to a Git remote.
 
+* Publish `config-schema.json` in the documentation site. Users can now reference
+  the schema in their `~/.config/jj/config.toml` file:
+
+```toml
+# Get editor completions based on the config schema
+"$schema" = "https://jj-vcs.github.io/jj/latest/config-schema.json"
```
thoughtpolice, line 85: Ideally, users would be able to target the version of the tool they have, but that's probably not always possible (i.e. git compiled versions?) We could figure that out later on, in any case!

CHANGELOG.md
```diff
@@ -77,6 +77,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New `git.sign-on-push` config option to automatically sign commits which are being
   pushed to a Git remote.
 
+* Publish `config-schema.json` in the documentation site. Users can now reference
```
thoughtpolice, line 80: I'd probably say:

> The Jujutsu documentation site now publishes a schema for the official configuration file, which can be integrated into your editor for autocomplete, inline errors, and more. Please [see the documentation](...) for more on this.

i.e. we normally would link to the docs elsewhere, rather than including the example inline.

thoughtpolice (commented):
This is a really cool feature I didn't know existed, thanks!

CHANGELOG.md
rwjblue: Once we have it deployed for a tagged release we can start updating to use the version (the same way the rest of the docs are versioned), it just won't work for any versions prior (unless we have a system for back porting that kind of thing?).

CHANGELOG.md
rwjblue: Sounds good, thank you!! I wasn't quite sure about how specific to be here. Will update shortly.

rwjblue: Done! Thanks for your help!

thoughtpolice (approved):
Thanks! I resolved the review nits I left earlier, but normally you can do that. I'll approve this but leave it here for a little while today in case anyone else wants to do a review.

PR #5356, opened 2025-01-14
docs config: an advanced example of `--config` syntax

ilyagr:
This seems non-trivial to discover and could be useful for people transitioning form `--config-toml`.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1319,11 +1319,20 @@ to specify additional configuration settings. This overrides settings defined in
 config files or environment variables. For example,
 
 ```shell
-jj --config=ui.color=always --config=ui.diff-editor=kdiff3 split
+# Must not have spaces around the `=`
+jj --config ui.color=always --config ui.diff-editor=kdiff3 split
 ```
 
 Config value should be specified as a TOML expression. If string value doesn't
 contain any TOML constructs (such as array notation), quotes can be omitted.
+Here is an example with more advanced TOML constructs:
+
+```shell
+# Single quotes and the '\' are interpreted by the shell and assume a Unix shell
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -1279,11 +1279,20 @@ to specify additional configuration settings. This overrides settings defined in
...
 ```
 
 Config value should be specified as a TOML expression. If string value doesn't
 contain any TOML constructs (such as array notation), quotes can be omitted.
+Here is an example with more advanced TOML constructs:
+
+```shell
+# Single quotes and the first '\' are interpreted by the shell and assume a Unix shell
+# Double quotes and the remaining '\'s are parsed as TOML syntax
+jj log -config \
```
yuja, line 1293: typo: `--config`

docs/config.md
```diff
@@ -1279,11 +1279,20 @@ to specify additional configuration settings. This overrides settings defined in
...
+  'template-aliases."format_timestamp(timestamp)"="timestamp.format(\"%Y-%m-%d %H:%M %:::z\")"'
```
yuja, line 1294: nit: we can use triple-quotes instead `"""timestamp.format("...")"""` if the purpose wasn't to demonstrate backslash mess.

yuja (approved):

docs/config.md
thoughtpolice: We should add a triple quote example too, then. I had no idea that existed!

ilyagr: I think I'll just have a triple-quoted example. It seems cleaner and more educational, though we could also have an example to show backslash mess as well.

PR #5360, opened 2025-01-14
index: Return Result from heads

AM5800:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/simplify_parents.rs
+++ b/cli/src/commands/simplify_parents.rs
@@ -2,6 +2,7 @@ use std::collections::HashSet;
 
 use clap_complete::ArgValueCandidates;
 use itertools::Itertools;
+use jj_lib::backend::BackendError;
 use jj_lib::revset::RevsetExpression;
 
 use crate::cli_util::CommandHelper;
@@ -87,7 +88,11 @@ pub(crate) fn cmd_simplify_parents(
         .transform_descendants(commit_ids, |mut rewriter| {
             let num_old_heads = rewriter.new_parents().len();
             if commit_ids_set.contains(rewriter.old_commit().id()) && num_old_heads > 1 {
-                rewriter.simplify_ancestor_merge();
+                // TODO: BackendError is not the right error here because
+                // the error does not come from `Backend`, but `Index`.
+                rewriter
+                    .simplify_ancestor_merge()
+                    .map_err(|err| BackendError::Other(err.into()))?;
             }
             let num_new_heads = rewriter.new_parents().len();
 
--- a/lib/src/default_index/composite.rs
+++ b/lib/src/default_index/composite.rs
@@ -43,6 +43,7 @@ use crate::hex_util;
 use crate::index::AllHeadsForGcUnsupported;
 use crate::index::ChangeIdIndex;
 use crate::index::Index;
+use crate::index::IndexError;
 use crate::object_id::HexPrefix;
 use crate::object_id::ObjectId;
 use crate::object_id::PrefixResolution;
@@ -487,15 +488,19 @@ impl Index for &CompositeIndex {
         Ok(Box::new(self.all_heads()))
     }
 
-    fn heads(&self, candidate_ids: &mut dyn Iterator<Item = &CommitId>) -> Vec<CommitId> {
+    fn heads(
+        &self,
+        candidate_ids: &mut dyn Iterator<Item = &CommitId>,
+    ) -> Result<Vec<CommitId>, IndexError> {
         let candidate_positions: BTreeSet<_> = candidate_ids
             .map(|id| self.commit_id_to_pos(id).unwrap())
             .collect();
 
-        self.heads_pos(candidate_positions)
+        Ok(self
+            .heads_pos(candidate_positions)
             .iter()
             .map(|pos| self.entry_by_pos(*pos).commit_id())
-            .collect()
+            .collect())
     }
 
     fn evaluate_revset<'index>(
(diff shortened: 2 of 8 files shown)


lib/src/repo.rs
```diff
@@ -1428,6 +1428,7 @@ impl MutableRepo {
             view.head_ids = self
                 .index()
                 .heads(&mut view.head_ids.iter())
+                .unwrap()
```
PhilipMetzger, line 1431: nit: Add a comment here that we know that the commit should exist, as the default index impl guarantees it.

PhilipMetzger (approved):
LG

lib/src/repo.rs
martinvonz: I'm not sure the default implementation can't fail, but that still doesn't help non-default implementations. I think it's better to leave a TODO here (or update this function and its callers to propagate the error).

lib/src/rewrite.rs
```diff
@@ -197,14 +197,16 @@ impl<'repo> CommitRewriter<'repo> {
 
     /// If a merge commit would end up with one parent being an ancestor of the
     /// other, then filter out the ancestor.
-    pub fn simplify_ancestor_merge(&mut self) {
+    pub fn simplify_ancestor_merge(&mut self) -> Result<(), BackendError> {
         let head_set: HashSet<_> = self
             .mut_repo
             .index()
             .heads(&mut self.new_parents.iter())
+            .map_err(|err| err.expect_backend_error())?
```
martinvonz, line 205: It seems better to let `simplify_ancestor_merge()` return the same error as `heads()` returns instead of mapping it here

lib/src/default_index/composite.rs
```diff
@@ -487,15 +487,19 @@ impl Index for &CompositeIndex {
         Ok(Box::new(self.all_heads()))
     }
 
-    fn heads(&self, candidate_ids: &mut dyn Iterator<Item = &CommitId>) -> Vec<CommitId> {
+    fn heads(
+        &self,
+        candidate_ids: &mut dyn Iterator<Item = &CommitId>,
+    ) -> Result<Vec<CommitId>, RevsetEvaluationError> {
```
martinvonz, line 493: It seems a little strange to return `RevsetEvaluationError` here because there's no revset visibly involved. Maybe we should introduce a new error like this (basically the same as the `IndexReadError` and `IndexWriteError` errors associated with the `IndexStore` trait):

```
#[derive(Debug, Error)]
#[error(transparent)]
pub struct IndexError(pub Box<dyn std::error::Error + Send + Sync>);
```

lib/src/repo.rs
yuja: The default implementation wouldn't fail (because all input commits must be valid), but I don't know if that applies to the Google backend. It's probably better to use different error type to clarify what kind of errors are expected.

I don't think it would make sense to propagate error from this function, at leas with the current design. `enforce_view_invariants()` exists for lazy computation, so it can be called from arbitrary places where the method call is supposed to be immutable & infallible.

lib/src/default_index/composite.rs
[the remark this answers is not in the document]
yuja: Isn't that coding error?

I'm not against allowing invalid/unindexed commit ids, but if that's expected, I would add `::CommitNotFound` or something.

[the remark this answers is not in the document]
martinvonz: Good point. I was thinking that we could get here if we had a broken index with some commits missing, but I guess that wouldn't really happen (it would probably fail at an earlier point). Sounds good to keep the `unwrap()`.

lib/src/repo.rs
martinvonz: The Google backend does RPCs so it could fail. I think @AM5800 wanted to avoid crashing server jobs if that happens (even the server jobs do RPCs to the database). Maybe it's unlikely enough that we don't have to worry about it for now.

PhilipMetzger: After this conversation, I think also adding a TODO comment is approriate to maybe redesign the model to allow intermediate index failures and the comment should mention that we assume it for _all_ index implementations. 

> The default implementation wouldn't fail (because all input commits must be valid), but I don't know if that applies to the Google backend

That's what I assumed, so thank you for confirming it.

AM5800: Added a TODO

lib/src/rewrite.rs
AM5800: Done, thanks!

lib/src/default_index/composite.rs
AM5800: Added IndexError as you suggested above, thanks!

yuja (approved):

cli/src/commands/simplify_parents.rs
```diff
@@ -87,7 +88,9 @@ pub(crate) fn cmd_simplify_parents(
         .transform_descendants(commit_ids, |mut rewriter| {
             let num_old_heads = rewriter.new_parents().len();
             if commit_ids_set.contains(rewriter.old_commit().id()) && num_old_heads > 1 {
-                rewriter.simplify_ancestor_merge();
+                rewriter
+                    .simplify_ancestor_merge()
+                    .map_err(|err| BackendError::Other(err.into()))?;
```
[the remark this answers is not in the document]
AM5800: Added TODO

lib/src/rewrite.rs
```diff
@@ -304,7 +306,9 @@ pub fn rebase_commit_with_options(
 ) -> BackendResult<RebasedCommit> {
     // If specified, don't create commit where one parent is an ancestor of another.
     if options.simplify_ancestor_merge {
-        rewriter.simplify_ancestor_merge();
+        rewriter
+            .simplify_ancestor_merge()
+            .map_err(|err| BackendError::Other(err.into()))?;
```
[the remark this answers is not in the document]
AM5800: Added a TODO

PR #5375, opened 2025-01-15
docs: Move language references to a Reference section

jennings:
I think the reference pages for the revset, fileset, and templating languages are different concerns from the configuration reference pages.

As [suggested by @arxanas](https://discord.com/channels/968932220549103686/1307814358906179694/1326971715406008350), I've moved the language descriptions into a new Reference section.

Then I created new pages `revset-aliases.md` and `template-aliases.md` in the Configuration section for the well-known aliases like `immutable_heads()`, `format_timestamp()`, etc.


jennings:
The "Check that MkDocs..." build failure seems caused by this issue: https://github.com/astral-sh/uv/issues/10654

martinvonz:
Seems like @ilyagr should review this (others can of course also review). Will you have time to look at this?

ilyagr:
Could you rebase? I'd like to know whether CI will pass now that uv 0.5.20 was released. Since 0.5.19 was the newest version for less than a day, perhaps we can ignore its existence...

docs/config.md
```diff
@@ -457,54 +434,6 @@ To get shorter prefixes for certain revisions, set `revsets.short-prefixes`:
...
-will need to modify the `format_time_range()` template alias.
-
-```toml
-[template-aliases]
-'format_time_range(time_range)' = 'time_range.start() ++ " - " ++ time_range.end()'
-```
-
-### Author format
-
-Can be customized by the `format_short_signature()` template alias.
```
ilyagr, line 482: Ditto.

docs/config.md
```diff
@@ -457,54 +434,6 @@ To get shorter prefixes for certain revisions, set `revsets.short-prefixes`:
...
-'format_short_signature(signature)' = 'signature.email()'
-# Both name and email address
-'format_short_signature(signature)' = 'signature'
-# Username part of the email address
-'format_short_signature(signature)' = 'signature.username()'
-```
-
-### Commit timestamp
-
-Commits have both an "author timestamp" and "committer timestamp". By default,
```
ilyagr, line 496: Ditto. I'm wondering if this paragraph can also be shortened.

docs/template-aliases.md
```diff
@@ -0,0 +1,101 @@
...
+[template-aliases]
+# Full email address (default)
+'format_short_signature(signature)' = 'signature.email()'
+# Both name and email address
+'format_short_signature(signature)' = 'signature'
+# Username part of the email address
+'format_short_signature(signature)' = 'signature.username()'
+```
+
+## `commit_timestamp(commit)`
```
ilyagr, line 85: Perhaps this could be moved after `format_time_range`?

docs/template-aliases.md
```diff
@@ -0,0 +1,101 @@
+# Template aliases
+
+Some template aliases are built-in to Jujutsu and can be used to customize its output.
+
+See [Templating language](templates.md) for information about the templating
+syntax.
+
+## `format_short_change_id(id)`
```
ilyagr, line 8: We could split this document into three sections:

```
## Display of change and commit ids

### format_short_change_id

### format_short_commit_id

### format_short_change

## Display of times and dates

....

## Display of commit authors
```

docs/revset-aliases.md
```diff
@@ -0,0 +1,51 @@
...
+
+When Jujutsu clones a Git repository, it uses the remote's HEAD to set this
+revset aliases in the repo's config.
+
+```toml
+[revset-aliases]
+"trunk()" = "dev@origin"
+```
+
+## `immutable_heads()`
```
ilyagr, line 23: This seems like it would be cryptic to a new reader.

Perhaps something like "`immutable_heads()`: changing the set of immutable commits"? Or maybe have a redundant subheading:

```md
## Changing the set of immutable commits

### `immutable_heads()`
```

docs/revset-aliases.md
```diff
@@ -0,0 +1,51 @@
...
+Some revset aliases are built-in to Jujutsu and used to control its behavior.
+
+See [Revset language](revsets.md) for information about the revset syntax.
+
+## `trunk()`
+
+Most teams have a main line of development, usually named `main`, `master`, or
+`trunk`. The `trunk()` alias resolves to the bookmark for this branch.
+
+This alias is used in other built-in aliases, it is not used directly by
```
ilyagr, line 12: I'd be more specific: 

```md
this alias is used in the default values for the [log revset]
and for [`immutable_heads()`].
```

docs/revset-aliases.md
```diff
@@ -0,0 +1,51 @@
+# Revset aliases
```
ilyagr, line 1: (Open-ended confusion)

I wonder whether we should reorganize this PR somehow because of the existence of `[revsets]` or `[templates]`. The reorganization would probably end up being relatively slight, but it's not entirely clear to me what exactly it would look like.

For example, as things currently are, it's a bit confusing that the default log revset is not in this doc (it is documented right below the diff of config.md). To a lesser degree, this might also apply to default templates.

It doesn't help that I'm still a bit uncomfortable about the distinction (beyond the sense that `[revsets]` are more built-in than `[revset-aliases]`; not sure whether this is meaningful to a user).

I'm not sure whether resolving this is in scope of this PR, but either way I wonder if this URL should be renamed, and if so -- to what?

ilyagr (commented):
Thanks, splitting `config.md` sounds great to me! A reference section also sounds great.

I had a bunch of minor comments. There's also a comment I marked as "Open-ended confusion" about `[revsets]` and `[templates]` that I'm not sure what to do about.

docs/config.md
```diff
@@ -426,29 +426,6 @@ log-word-wrap = true
 
 ### Display of commit and change ids
```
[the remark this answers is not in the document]
jennings: I also changed the second paragraph to this, which I _think_ made it clearer what this setting actually does?

> To control which revisions get priority for shorter prefixes, set `revsets.short-prefixes`:

uv.lock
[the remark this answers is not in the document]
jennings: I removed this commit when I rebased and it looks like it builds fine now 👍

docs/revset-aliases.md
jennings: I was concerned about breaking links if we rephrase it. This anchor is `#immutable_heads`, but adding more text would lead to `#immutable_heads-changing-the-set-of-immutable-commits`, which would break if we rephrase it to, say, "immutable_heads(): Set of immutable commits".

I didn't see a way to control what anchor MkDocs uses. Ideally, the anchor would be `#immutable_heads` but the visible text would be more descriptive.

What do you think?

docs/revset-aliases.md
yuja: I think it's better to document `[revsets]` and `[revset-aliases]` in the same page (maybe in "Revset configuration"?) They're closely related, and their distinction is kinda implementation detail.

jennings: - Renamed the new files to revset-config.md and template-config.md
- Moved `revsets.*` and `templates.*` setting descriptions to those files.

docs/template-config.md
```diff
@@ -0,0 +1,202 @@
+# Template configuration
+
+## Templates
+
+This configuration section configures the default. These values can use [template
+aliases](#template-aliases).
+
```
yuja, line 7: I think it's better to briefly describe the overview of the `[templates]` section. For example, we had this in the original doc:

```
You can configure the template used when no `-T` is specified.

- `templates.log` for `jj log`
- `templates.op_log` for `jj op log`
- `templates.show` for `jj show`

[templates]
# Use builtin log template
log = "builtin_log_compact"
# Use builtin op log template
op_log = "builtin_op_log_compact"
# Use builtin show template
show = "builtin_log_detailed"
```

I also think we should omit detailed section unless there are things to be documented. For example, the `### Show` section is pretty much empty.

mkdocs.yml
```diff
@@ -111,16 +111,18 @@ nav:
...
 - 'Comparisons':
       - 'Git comparison': 'git-comparison.md'
       - 'Git compatibility': 'git-compatibility.md'
       - 'Sapling': 'sapling-comparison.md'
       - 'Other related work': 'related-work.md'
 
+- 'Reference':
+      - 'Revset language': 'revsets.md'
+      - 'Fileset language': 'filesets.md'
+      - 'Templating language': 'templates.md'
```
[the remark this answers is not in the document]
jennings: Sure. I was thinking to sort them by "importance" but there's only three so it really doesn't matter.

docs/revset-aliases.md
ilyagr: I was also worried a bit, perhaps the "redundant subheading" option might help. If we can't figure out an ideal solution, I think readability is more important than potential changes to heading anchors.

I also don't know a good way to affect MkDocs anchors or add new ones. We could use HTML, problably, but that'd be terrible for reading the docs in the terminal.

jennings: Ah! [I found how to set anchors with just markdown](https://www.mkdocs.org/about/release-notes/#anchors):

```md
### `immutable_heads()`: Set of immutable commits {#immutable_heads}
```

I've added more descriptive text to both `trunk()` and `immutable_heads()`.

docs/config.md
```diff
@@ -459,51 +392,19 @@ short-prefixes = "(main..@)::"
...
-
-`jj op log` defaults to relative timestamps. To use absolute timestamps, you
-will need to modify the `format_time_range()` template alias.
-
-```toml
-[template-aliases]
-'format_time_range(time_range)' = 'time_range.start() ++ " - " ++ time_range.end()'
-```
+You can use relative timestamps by adjusting the [`format_timestamp()` template
+alias](template-config.md#format_timestamp).
```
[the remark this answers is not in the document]
jennings: This is a request to remove headings like "Relative timestamps", "Author format", etc., and just have one blanket section "Template and style" that links over to the template-config.md page. Do I understand correctly?

@ilyagr suggested keeping these headings so incoming links don't break, like the link to `#node-style` that `jj log --help` outputs.

But, now that I've found how to add custom anchors to the page, I could make all those anchors go to the relevant blanket section.

```markdown

### Template and style

[](){#relative-timestamps}[](){#author-format}...

See [Log template configuration](template-config.md) for more information.
```

What do you think?

ilyagr:
>
> Ah! I found how to set anchors with just markdown
> <https://www.mkdocs.org/about/release-notes/#anchors>:
>

I'm a bit frozen on this point. I might write more later, but TLDR:

- This syntax is *not* part of CommonMark, and is not supported by GitHub.
It's probably not supported by VS Code, etc.
- This syntax is supported by MkDocs and [Docusaurus](
https://docusaurus.io/docs/markdown-features/toc#heading-ids) (which is my
best guess at a CommonMark-based doc builder we could use that has
versioning, even though I don't like its heavy reliance on React&JS magic)
- This syntax would look OK if we printed the files unformatted to stdout
(like we might be doing with `jj help --topic`).

The more HTML `<a>` syntax, also explained in Stephen's links,

- should be supported by any HTML-destined Markdown system.
- We'd probably have to find a way to remove it if we print the Markdown to
stdout or if we ever want to convert it via Pandoc.
- I haven't tested it, but hopefully MkDocs will recognize these anchors
and won't label links to them as broken.
- Not sure about VS Code

I'm not sure whether we should use one of these, or just grit our teeth and
only use the anchors given by headings, which everything understand, but
would mean that we should try to get the heading structure right the first
time (though we shouldn't overdo it either, it's not the end of the world
if people have to use their eyes to search through the page, especially if
we make it shorter).

Sorry for a sleepy comment with runon sentences.

steveklabnik:
> This syntax is *not* part of CommonMark, and is not supported by GitHub.

If I type `<http://example.com>` here: <http://example.com> it does link on GitHub.

While I don't see it in the CommonMark specification, their "dingus" does in fact render properly too, you can try it here https://spec.commonmark.org/dingus/

In my Windows Terminal, running bash in Ubuntu, links like this are clickable, when I hover over this, it gets underlined, and when I click it, it does open the link in my browser:

image


> even though I don't like its heavy reliance on React&JS magic

A nice thing about Docasaurus is that it's primarily server-side rendered, so I'm not sure *which* parts you object to, but it's all at build time, not in the browser.

ilyagr:
@steveklabnik I meant the "specifying the anchor" syntax.

**Aside:** As a test, here is a link to two ids below: [first id](#my-custom-heading-id), [second id](#my-very-custom-heading-id). The first one does not work, the second one does

You can see the GitHub markdown rendering here failing here for `{#trunk}` (it should not be shown):

image

You can also see it below.

## Here are the targets of the above links

The source is:

```
### This anchor does not work {#my-custom-heading-id}

<a id="my-very-custom-heading-id"></a> `<<<` There is an invisible anchor here
```


### This anchor does not work {#my-custom-heading-id}

<a id="my-very-custom-heading-id"></a>  `<<<` There is an invisible anchor here

PR #5391, opened 2025-01-18
style/cleanup: remove a bunch of Default implementations

ilyagr:
I noticed that some config structures do not use their "Default" implementation, and have their default specified in a TOML file. I think specifying defaults in TOML is great, and it'd be good to delete those Default implementations, so that it does not diverge from the default in the TOML file.

I then kept going with a few more structures that seem better off without Default implemented.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/debug/watchman.rs
+++ b/cli/src/commands/debug/watchman.rs
@@ -50,6 +50,10 @@ pub fn cmd_debug_watchman(
 
     let mut workspace_command = command.workspace_helper(ui)?;
     let repo = workspace_command.repo().clone();
+    let watchman_config = WatchmanConfig {
+        // The value is likely irrelevant here. TODO(ilyagr): confirm
+        register_trigger: false,
+    };
     match subcommand {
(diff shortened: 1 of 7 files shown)


cli/src/commands/debug/watchman.rs
```diff
@@ -50,6 +50,9 @@ pub fn cmd_debug_watchman(
 
     let mut workspace_command = command.workspace_helper(ui)?;
     let repo = workspace_command.repo().clone();
+    let dummy_watchman_config = WatchmanConfig {
```
martinvonz, line 53: nit: drop the `dummy_` prefix in case we later add configs to the `WatchmanConfig` (it's possible their value will matter to the debug command so it's not a dummy config anymore then)

martinvonz (approved):

cli/src/commands/debug/watchman.rs
ilyagr: I replaced it with a comment. I realized I'm only 90% sure the value is irrelevant.

PR #5392, opened 2025-01-18
cli: abandon: delete bookmarks pointing to abandoned commits

### LINKED ISSUE #3505: FR: Stop branches from sliding back on `jj abandon`, delete them instead
Branches currently are designed to slide back when a change is deleted. @arxanas wrote:

> I don't think branches should slide backwards to begin with. git-branchless deletes the branch, and I'm actually surprised to hear that Mercurial slides it backwards. Sapling also deletes the branch (regardless of whether you address the commit via hash or branch):

I have yet to see a use case which benefits from the sliding mechanism, but see many use cases that break because of it.

Consider:
```
◉  pvlpykyo msta 5 days ago foo 13bbdfc2
│  My local commit
◉  vzokzlpu msta 5 days ago main a2016214
│  immutable head here
```
If I were to write `jj abandon foo`, then the branch `foo` would instead move to `main`. The same thing would happen if I were to submit foo, then later run `jj rebase --skip-empty`.

This is a problem for multiple use cases. The first is when you want to associate each commit with a change (eg. make one commit branch be `crrev.com/c/123`)

I made #3482 earlier. However, this would be my preferred solution.

**Additional context**

yuja:
Closes #3505

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* `jj abandon` now deletes bookmarks pointing to the revisions to be abandoned.
+  Use `--retain-bookmarks` to move bookmarks backwards. If deleted bookmarks
+  were tracking remote bookmarks, the associated bookmarks (or branches) will be
+  deleted from the remote on `jj git push --all`.
+  [https://github.com/jj-vcs/jj/issues/3505](#3505)
+
 * `jj init --git` and `jj init --git-repo` have been removed. They were
   deprecated in early 2024. Use `jj git init` instead.
 
--- a/lib/tests/test_rewrite.rs
+++ b/lib/tests/test_rewrite.rs
@@ -1049,31 +1049,51 @@ fn test_rebase_descendants_update_bookmark_after_abandon() {
     let test_repo = TestRepo::init();
     let repo = &test_repo.repo;
 
-    // Bookmark "main" points to commit B. B is then abandoned. Bookmark main should
-    // be updated to point to A.
+    // Commit B is abandoned. Local bookmarks should be deleted or moved
+    // accordingly, whereas remote bookmarks should not get updated.
     //
-    // B main
-    // |          =>   A main
-    // A
+    // C other
+    // |
+    // B main main@origin        C2 other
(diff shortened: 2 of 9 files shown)


lib/tests/test_rewrite.rs
```diff
@@ -1077,6 +1078,56 @@ fn test_rebase_descendants_update_bookmark_after_abandon() {
     );
 }
 
+#[test]
+fn test_rebase_descendants_delete_bookmark_after_abandon() {
```
martinvonz, line 1082: Would be good to also have tests showing what happens when one or all targets of a conflicted bookmark get abandoned. It's not obvious what we should do. Let's say the bookmark target is a conflict like `A+(B-C)+(D-E)`. Perhaps we should apply a diff that removes the target, i.e. by replacing it by `absent()`. So if you abandon the `A` commit, the new conflict is `A+(B-C)+(D-E)+(∅-A)= ∅+(B-C)+(D-E)`. What do you think?

yuja: Added (and updated) tests. I also reverted the doc change since bookmark conflicts can't always be resolved by `jj abandon`.

CHANGELOG.md
```diff
@@ -38,6 +38,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
     Inner items of inline tables are no longer merged across configuration
     files.
 
+* `jj abandon` now deletes bookmarks pointing to the revisions to be abandoned.
+  Use `--retain-bookmarks` to move bookmarks backwards.
+  [https://github.com/jj-vcs/jj/issues/3505](#3505)
```
martinvonz, line 43: This might be the most important breaking change, so put it first? (We sometime reorder items when we do the release, but sometimes we forget.)

martinvonz (approved):
Thanks!

CHANGELOG.md
martinvonz: I wonder if we should mention somewhere that a subsequent `jj git push --all` will delete the branch from the remote. Do you think a note here is good? And/or in `jj abandon --help`?

yuja: Updated the changelog entry. I don't think it's surprising that deleted bookmarks are propagated on push, so I think changelog entry is good.

Issue #5218, opened 2024-12-31
help text for rebase is inconsistent


joyously:
## Description
My reading of the `jj help rebase` text is perhaps tainted by seeing `-r` used on other commands as plural, so I equate it with "revset". I got confused by the first paragraph saying that `-s` is for "a revision and its descendants" followed immediately by `-r` for "a single commit".  They seemed backward to me, so the examples were a bit much especially since they take up enough room that I can't fit the first paragraph explanation on the same screen as the option descriptions. In addition, the example graphs have visible Markdown taking up space. 
The option description says that `-s` is for source and `-r` is for revisions (plural).

This is one explanation where using both "commit" and "revision" is very confusing.
The help needs to say a little more about rebase than simply "move revisions to different parents", to give at least a hint of what changes and/or what consequences or limits there are since novice users need that.

## Steps to Reproduce the Problem

1. `jj help rebase`
1. compare first paragraph to option definitions

## Expected Behavior
I expect option letters to make sense.

## Actual Behavior
Option letters represent renamed concepts ( -b for branch ), no match for the letter ( -s for "a revision and its descendants" ), and different plurality ( -r is "single commit" versus "--revisions" )

## Specifications

- Platform: Ubuntu Studio 24.04
- Version: jj 0.23.0

arxanas:
Link for reference: https://jj-vcs.github.io/jj/v0.24.0/cli-reference/#jj-rebase

> My reading of the jj help rebase text is perhaps tainted by seeing -r used on other commands as plural, so I equate it with "revset".

I think the documentation is misleading and I believe `-r` can take an arbitrary revset, not just a single revision, as you expected.
- This is useful in case you want to "splice out" a set of commits and rebase it.
- It's corroborated later in the docs when it refers to "specified revisions".

> I got confused by the first paragraph saying that -s is for "a revision and its descendants" followed immediately by -r for "a single commit". They seemed backward to me, so the examples were a bit much especially since they take up enough room that I can't fit the first paragraph explanation on the same screen as the option descriptions. In addition, the example graphs have visible Markdown taking up space.

[historical note] I suspect the `-b` and `-s` options were taken directly from Mercurial, and `-r` may not have existed at first, hence the documentation might have been written in the "wrong" order. 

I agree that, from a conceptual perspective, it probably makes sense to discuss the `-r` case first, and define `-s` and `-b` flags in terms of that behavior.

I personally don't like `-b`/`--branch` and `-s`/`--source` as names that much.
[...]

PR #5400, opened 2025-01-19
cli: rebase: say that `-r` allows set, clarify `-s X` vs `-r X::`

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/rebase.rs
+++ b/cli/src/commands/rebase.rs
@@ -47,10 +47,16 @@ use crate::ui::Ui;
 
 /// Move revisions to different parent(s)
 ///
+/// This command moves revisions to different parent(s) while preserving the
+/// changes (diff) in the revisions.
+///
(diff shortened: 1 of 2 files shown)


arxanas:
Didn't look at PR yet but this issue was also reported in https://github.com/jj-vcs/jj/issues/5218 btw, might be worth a look if we're touching this area

yuja (approved):

martinvonz:
> Didn't look at PR yet but this issue was also reported in #5218 btw, might be worth a look if we're touching this area

Good point. I've update this PR to consistently use "revision" instead of using a mix of "revision" and "commit". I'm not sure if we want to do that in all our documentation. Our glossary currently say that "revision" is a synonym for "commit".

ilyagr:
LGTM, though I only looked it over briefly.

Issue #5394, opened 2025-01-18
`jj restore --to` is easy to mix up with `jj restore --from`


v-gb:
When I want to reset the working copy to whatever some other revision contained, I frequently use both `jj restore --from x` and  `jj restore --to x`. Alas, the former is correct, but the latter creates crazy conflicts, until `jj undo` saves the day once again.

I think the reason for my recurring confusion is that, in a sentence, `restore from` is clear but mostly used in "restore from backup" and similar phrasing, while "restore to" is more frequent and often means the same thing ("I restored the config to the previous version", "I restored the painting to its 18th century state").

Anyway, maybe it's just me getting confused by this, or maybe there's nothing worth doing here, but I thought I'd bring it up just in case.

In terms of potential ways to change jj:
- maybe  `--src` and `--dst` would be clearer, but I imagine a number of places use the `--to`/`--from` terminology, so probably a non starter
- maybe `--to` could be renamed `--into`, but same problem as above
- maybe specifying `--to` without `--from` could be rejected. You'd have to write `--from @` explicitly if that's what you mean (this operation seems very niche, so it seems fine to add a hurdle there)

ilyagr:
The idea of requiring `--from` if you specify `--to` sounds good to me.

I was also surprised recently that we haven't renamed `--to` to `--into` for `jj restore`. I don't remember why, even though I may have been involved in the discussion. I think that would help. If we also require `--from` with `--into`, I think it might be OK to keep `--to` as an alias; if we don't, then we'd require people to write `--into`.

v-gb:
`--onto` would have to mean "on top of" rather than "into", so it wouldn't work for `restore`. It could work for `new`/`rebase`, but they already have they terminology `--insert-after`.
 
Ah, I didn't even know there were `--into` flags already. Surveying the code:

- `squash`/`absorb` use `--from`/`--into`. https://github.com/jj-vcs/jj/discussions/2882 is apparently the source of the `--into` for `squash`, and then I suppose `absorb` did the same thing. No discussion of `restore`.
- `diffedit`/`interdiff`/`diff`/`restore` use `--from`/`--to`
- `bookmark move` uses `--to`
- `new`/`rebase`/`duplicate` do more involved things

I think it would make sense to say:
- commands that modify a commit based on some other ones use `--from`/`--into`
- commands that compare two commits use `--from`/`--to`
- commands that create new commits at arbitrary locations have a different need (designate positions between commits, rather than commits), so no reason to rethink that here

[...]

PR #5401, opened 2025-01-19
restore: add --into flag, make --to an alias to it

v-gb:
This is a part of #5394.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/restore.rs
+++ b/cli/src/commands/restore.rs
@@ -63,24 +63,24 @@ pub(crate) struct RestoreArgs {
     from: Option<RevisionArg>,
     /// Revision to restore into (destination)
     #[arg(
-        long,
-        short,
+        long, short = 't',
+        visible_alias = "to",
         value_name = "REVSETS",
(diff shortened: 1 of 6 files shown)


cli/src/commands/restore.rs
```diff
@@ -116,12 +116,12 @@ pub(crate) fn cmd_restore(
         return Err(user_error(
             "`jj restore` does not have a `--revision`/`-r` option. If you'd like to modify\nthe \
              *current* revision, use `--from`. If you'd like to modify a *different* \
-             revision,\nuse `--to` or `--changes-in`.",
+             revision,\nuse `--into` or `--changes-in`.",
```
martinvonz, line 119: Thanks for not missing this instance. That reminds me that there might be docs to update too. Looks like only one instance in FAQ.md. Can you update that too? Thanks

v-gb: Indeed. I fixed the handful of matches of `rg 'restore.*--to'`, but apparently that was the only instance in the doc.

martinvonz (approved):

PR #5402, opened 2025-01-19
make `jj bookmark *` aliases non-visible

jakobhellermann:
This goes makes completions go from
```
c                                                               (Create a new bookmark)
d  (Delete an existing bookmark and propagate the deletion to remotes on the next push)
f          (Forget everything about a bookmark, including its local and remote targets)
l                                                    (List bookmarks and their targets)
m                                          (Move existing bookmarks to target revision)
r                                   (Rename `old` bookmark name to `new` bookmark name)
s                            (Create or update a bookmark to point to a certain commit)
t                                               (Start tracking given remote bookmarks)
untrack                                          (Stop tracking given remote bookmarks)
```

To
```
create                                                               (Create a new bookmark)
delete  (Delete an existing bookmark and propagate the deletion to remotes on the next push)
forget          (Forget everything about a bookmark, including its local and remote targets)
list                                                      (List bookmarks and their targets)
move                                            (Move existing bookmarks to target revision)
rename                                   (Rename `old` bookmark name to `new` bookmark name)
set                               (Create or update a bookmark to point to a certain commit)
track                                                (Start tracking given remote bookmarks)
untrack                                               (Stop tracking given remote bookmarks)
```
which is much more readable.

I think this is worth more than the discoverability of these short aliases.
The bookmark commands also all start with distinct characters, so `jj
bookmark c<TAB>` etc. all work.

I also added a `u = untrack` alias for consistency, it was the only subcommand that didn't have any.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/bookmark/mod.rs
+++ b/cli/src/commands/bookmark/mod.rs
@@ -80,6 +80,7 @@ pub enum BookmarkCommand {
     Set(BookmarkSetArgs),
     #[command(alias("t"))]
     Track(BookmarkTrackArgs),
+    #[command(alias("u"))]
     Untrack(BookmarkUntrackArgs),
 }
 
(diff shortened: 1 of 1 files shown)


yuja:
iirc, this breaks completion of arguments because hidden aliases aren't included in the generated completion script. (It might work in dynamic completion, but I haven't tried.)

jakobhellermann:
I see, that's annoying. I guess this is blocked on better clap support for hidden aliases then.

Issue #4561, opened 2024-10-02
`sign-all=true` doesn't sign commits authored by someone else


Wuestengecko:
## Description

With `signing.sign-all = true`, jj does not sign commits that someone else originally authored, after I made changes to them (like editing the description).

## Steps to Reproduce the Problem

1. `jj config set --user signing.sign-all true`
1. Edit a commit that someone else originally authored. Now Author ≠ Committer.
1. Verify the commit signature with e.g.: \
   `` git log -1 --format=fuller --show-signature `jj log --no-graph -r@ -Tcommit_id` ``

## Expected Behavior

The commit is signed (with my key, as I am the Committer)

## Actual Behavior

The commit is not signed

## Specifications

- Platform: Arch (btw)
- Version: 0.21.0

yuja:
I think the current behavior is by design, and this use case will be handled by `jj sign --force` or something. https://github.com/martinvonz/jj/pull/3142

Wuestengecko:
> I think the current behavior is by design

That's a bit surprising. First, git with `commit.gpgsign = true` behaves like I described (which is what led me to think this was a bug). Second, the setting is called "sign-***all***", and the docs also don't mention this restriction, as far as I can see.

https://github.com/martinvonz/jj/blob/main/docs/config.md#commit-signing

> `jj` can be configured to sign and verify the commits it creates using either GnuPG or SSH signing keys.

I can see how you could _maybe_ interpret this to mean it only signs new commits (as in, created with `jj new`, but not ones fetched from a remote and then edited). But then again, it will verify everything that has a signature, and not re-signing after making changes also kind of defeats the purpose - whether a signature is trustworthy is an entirely different discussion.

<sub>Also, [the argument was made](https://github.com/martinvonz/jj/issues/58#issuecomment-1068701465) that people are gonna want the "Verified" chip on Github. I'm one of those people. :)</sub>

yuja:
cc @necauqua

(I agree the doc is unclear.)

necauqua:
Huh, I'm pretty sure that by that design point 2 should've failed to complete, sonce the idea was that automatically rewriting *and* resigning someone elses commits may be undesirable.

Oh, maybe that other commit was not signed, and we just don't sign it too then?. That does sound weird. I need to revisit it all, was a while, I'll come back later

necauqua:
Hey I just noticed I haven't touched this in 3 months.
[...]

PR #5408, opened 2025-01-20
Allow signing commits by other authors when rewriting them

Wuestengecko:
Hi! I want to try and get the ball rolling on a fix for #4561, so here's a first shot at it. This simply adds another boolean flag `sign-foreign`, which upgrades `SignBehavior::Own` to `Force`. As such, it's a very non-invasive change.

The code in this PR is working fine for me locally, but I haven't really looked into automated tests yet.

[One proposed alternative](https://github.com/jj-vcs/jj/issues/4561#issuecomment-2559422465) is to change it to a `sign` variable that takes one of these behavior names. This would require migration or deprecation of the previous `sign-all` flag, which I haven't figured out how it works yet :)

Please let me know what you think about this. I'm also happy to change to a different approach (although it might take me a while, as I'm new to this codebase).

# Checklist

If applicable:

### DIFF
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -565,6 +565,11 @@
                     "description": "Whether to sign all commits by default. Overridden by global `--no-sign` option",
                     "default": false
                 },
+                "sign-foreign": {
+                    "type": "boolean",
+                    "description": "Whether to sign commits of other authors with your own key when rewriting them. Only applicable with `sign-all = true`.",
+                    "default": false
+                },
                 "backends": {
                     "type": "object",
                     "description": "Tables of options to pass to specific signing backends",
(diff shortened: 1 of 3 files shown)


martinvonz:
Thanks! 

> [One proposed alternative](https://github.com/jj-vcs/jj/issues/4561#issuecomment-2559422465) is to change it to a `sign` variable that takes one of these behavior names.

I just added [more support for that idea](https://github.com/jj-vcs/jj/issues/4561#issuecomment-2603030412
) (as you've probably seen already).

>  This would require migration or deprecation of the previous `sign-all` flag, which I haven't figured out how it works yet :)

Yuya just added better support for that in #5298, so you can probably look at the changes in that PR to figure out what to do.

yuja:
> #5298

I have unsent patches to rename+update config values. Maybe I need to push them first.

bnjmnt4n:
Closing since #5707 was merged.

PR #5414, opened 2025-01-21
dynamic completion: add revision completion for more options

ilyagr:


cli/src/commands/git/push.rs
```diff
@@ -141,15 +141,25 @@ pub struct GitPushArgs {
...
     ///
     /// The created bookmark will be tracked automatically. Use the
     /// `git.push-bookmark-prefix` setting to change the prefix for generated
     /// names.
-    #[arg(long, short, value_name = "REVSETS")]
+    #[arg(
+        long,
+        short,
+        value_name = "REVSETS",
+        add = ArgValueCandidates::new(complete::mutable_revisions)
```
yuja, line 161: nit: I won't disagree, but it's better to add inline comment why `-c` suggests mutable revisions only.

yuja (approved):

ilyagr:
Thank you very much for the review & comments!

PR #5415, opened 2025-01-21
builtin pager: add configuration options, change default options

ilyagr:
I've been going back and forth on naming of the `alternate-screen` option. It has many subtle effects that seem hard to describe without thinking of the alternate screen (e.g. if we call it `clear-screen`), at least if we want to allow the `if-long-or-slow` behavior. If we are OK with forbidding it, we could have a simple `quit-if-one-page` boolean, but I think it might just be useful enough to keep.

Streampager's naming of the corresponding [`interface_mode` option](https://github.com/facebook/sapling/blob/0280ed4ab533abbaa4836abc0854b66ac13a10fa/eden/scm/lib/third-party/streampager/src/config.rs#L14) seems more confusing to me: the Direct mode seems useless, there are two hybrid modes (one of which is called Hybrid) that are hard to distinguish unless you think in terms of the alternate screen terminal mode.

# TODOs

- [x] Describe streampager keybindings (here or another PR)
- [ ] (not planned) I have added tests to cover my changes

### DIFF
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -107,6 +107,30 @@
                     "description": "Pager to use for displaying command output",
                     "default": "less -FRX"
                 },
+                "streampager": {
+                    "type": "object",
+                    "description": "':builtin' (streampager-based) pager configuration",
+                    "properties": {
+                        "interface": {
+                            "description": "Whether to quit automatically, whether to clear screen on startup/exit",
+                            "enum": [
+                                "quit-if-one-page",
+                                "full-screen-clear-output",
+                                "quit-quickly-or-clear-output"
+                            ],
+                            "default": "never"
+                        },
+                        "wrapping": {
+                            "description": "Whether to wrap long lines",
+                            "enum": [
+                                "anywhere",
+                                "word",
+                                "none"
+                            ],
+                            "default": "anywhere"
+                        }
+                    }
+                },
                 "diff": {
                     "type": "object",
                     "description": "Options for how diffs are displayed",
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -79,10 +79,19 @@ impl UiOutput {
         Ok(UiOutput::Paged { child, child_stdin })
     }
 
-    fn new_builtin_paged() -> streampager::Result<UiOutput> {
+    fn new_builtin_paged(config: &StreampagerConfig) -> streampager::Result<UiOutput> {
+        // This uselessly reads ~/.config/streampager/streampager.toml, even
+        // though we then override the important options.
+        // TODO(ilyagr): Fix this once a version of streampager with
+        // https://github.com/facebook/sapling/pull/1011 is released.
         let mut pager = streampager::Pager::new_using_stdio()?;
-        // TODO: should we set the interface mode to be "less -FRX" like?
-        // It will override the user-configured values.
+        pager.set_wrapping_mode(config.wrapping);
+        pager.set_interface_mode(config.streampager_interface_mode());
+        // We could make scroll-past-eof configurable, but I'm guessing people
+        // will not miss it. If we do make it configurable, we should mention
+        // that it's a bad idea to turn this on if `interface=quit-if-one-page`,
+        // as it can leave a lot of empty lines on the screen after exiting.
+        pager.set_scroll_past_eof(false);
 
         // Use native pipe, which can be attached to child process. The stdout
         // stream could be an in-process channel, but the cost of extra syscalls
@@ -257,9 +266,59 @@ pub enum PaginationChoice {
     Auto,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, serde::Deserialize)]
+#[serde(rename_all(deserialize = "kebab-case"))]
+pub enum StreampagerAlternateScreenMode {
+    QuitIfOnePage,
+    FullScreenClearOutput,
+    QuitQuicklyOrClearOutput,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, serde::Deserialize)]
+#[serde(rename_all(deserialize = "kebab-case"))]
+enum StreampagerWrappingMode {
+    None,
+    Word,
+    Anywhere,
+}
+
+impl From<StreampagerWrappingMode> for streampager::config::WrappingMode {
+    fn from(val: StreampagerWrappingMode) -> Self {
+        use streampager::config::WrappingMode;
+        match val {
+            StreampagerWrappingMode::None => WrappingMode::Unwrapped,
+            StreampagerWrappingMode::Word => WrappingMode::WordBoundary,
+            StreampagerWrappingMode::Anywhere => WrappingMode::GraphemeBoundary,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, serde::Deserialize)]
+#[serde(rename_all(deserialize = "kebab-case"))]
+struct StreampagerConfig {
+    interface: StreampagerAlternateScreenMode,
+    wrapping: StreampagerWrappingMode,
+    // TODO: Add an `quit-quickly-delay-seconds` floating point option or a
+    // `quit-quickly-delay` option that takes a 's' or 'ms' suffix. Note that as
+    // of this writing, floating point numbers do not work with `--config`
+}
+
+impl StreampagerConfig {
+    fn streampager_interface_mode(&self) -> streampager::config::InterfaceMode {
+        use streampager::config::InterfaceMode;
+        use StreampagerAlternateScreenMode::*;
+        match self.interface {
+            // InterfaceMode::Direct not implemented
+            FullScreenClearOutput => InterfaceMode::FullScreen,
+            QuitIfOnePage => InterfaceMode::Hybrid,
+            QuitQuicklyOrClearOutput => InterfaceMode::Delayed(std::time::Duration::from_secs(2)),
+        }
+    }
+}
+
 enum PagerConfig {
     Disabled,
-    Builtin,
+    Builtin(StreampagerConfig),
     External(CommandNameAndArgs),
 }
 
@@ -270,7 +329,7 @@ impl PagerConfig {
         };
         match config.get("ui.pager")? {
             CommandNameAndArgs::String(name) if name == BUILTIN_PAGER_NAME => {
-                Ok(PagerConfig::Builtin)
+                Ok(PagerConfig::Builtin(config.get("ui.streampager")?))
             }
             _ => Ok(PagerConfig::External(config.get("ui.pager")?)),
         }
@@ -318,16 +377,18 @@ impl Ui {
             PagerConfig::Disabled => {
                 return;
             }
-            PagerConfig::Builtin => UiOutput::new_builtin_paged()
-                .inspect_err(|err| {
-                    writeln!(
-                        self.warning_default(),
-                        "Failed to set up builtin pager: {err}",
-                        err = format_error_with_sources(err),
-                    )
-                    .ok();
-                })
-                .ok(),
+            PagerConfig::Builtin(streampager_config) => {
+                UiOutput::new_builtin_paged(streampager_config)
+                    .inspect_err(|err| {
+                        writeln!(
+                            self.warning_default(),
+                            "Failed to set up builtin pager: {err}",
+                            err = format_error_with_sources(err),
+                        )
+                        .ok();
+                    })
+                    .ok()
+            }
             PagerConfig::External(command_name_and_args) => {
                 UiOutput::new_paged(command_name_and_args)
                     .inspect_err(|err| {
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -201,8 +201,7 @@ impl Write for UiStderr<'_> {
 
 pub struct Ui {
     quiet: bool,
-    pager_cmd: CommandNameAndArgs,
-    paginate: PaginationChoice,
+    pager: PagerConfig,
     progress_indicator: bool,
     formatter_factory: FormatterFactory,
     output: UiOutput,
@@ -258,12 +257,31 @@ pub enum PaginationChoice {
     Auto,
 }
 
+enum PagerConfig {
+    Disabled,
+    Builtin,
+    External(CommandNameAndArgs),
+}
+
+impl PagerConfig {
+    fn from_config(config: &StackedConfig) -> Result<PagerConfig, ConfigGetError> {
+        if matches!(config.get("ui.paginate")?, PaginationChoice::Never) {
+            return Ok(PagerConfig::Disabled);
+        };
+        match config.get("ui.pager")? {
+            CommandNameAndArgs::String(name) if name == BUILTIN_PAGER_NAME => {
+                Ok(PagerConfig::Builtin)
+            }
+            _ => Ok(PagerConfig::External(config.get("ui.pager")?)),
+        }
+    }
+}
+
 impl Ui {
     pub fn null() -> Ui {
         Ui {
             quiet: true,
-            pager_cmd: CommandNameAndArgs::String(String::new()),
-            paginate: PaginationChoice::Never,
+            pager: PagerConfig::Disabled,
             progress_indicator: false,
             formatter_factory: FormatterFactory::plain_text(),
             output: UiOutput::Null,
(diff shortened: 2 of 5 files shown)


cli/src/ui.rs
```diff
@@ -267,43 +266,63 @@ pub enum PaginationChoice {
...
+            return Ok(PagerConfig::Disabled);
+        };
+        let pager_cmd: CommandNameAndArgs = config.get("ui.pager")?;
+        match pager_cmd {
+            CommandNameAndArgs::String(name) if name == BUILTIN_PAGER_NAME => {
+                Ok(PagerConfig::Builtin)
+            }
+            _ => Ok(PagerConfig::Command(pager_cmd)),
+        }
+    }
```
yuja: nit: maybe move to `PagerConfig::from_config() -> Result<Self, ConfigGetError>`?

cli/src/ui.rs
```diff
@@ -267,43 +266,63 @@ pub enum PaginationChoice {
...
-        let use_builtin_pager = matches!(
-            &self.pager_cmd, CommandNameAndArgs::String(name) if name == BUILTIN_PAGER_NAME);
-        let new_output = if use_builtin_pager {
-            UiOutput::new_builtin_paged()
+        let new_output = match &self.pager {
+            PagerConfig::Disabled => {
+                unreachable!(
+                    "earlier check avoids possible syscall when pager disabled, may not be worth \
+                     it"
+                )
```
yuja, line 381: nit: `return` here? It's probably okay to test `is_terminal()` unconditionally.

cli/src/ui.rs
```diff
@@ -267,43 +266,63 @@ pub enum PaginationChoice {
     Auto,
 }
 
+enum PagerConfig {
+    Disabled,
+    Builtin,
+    Command(CommandNameAndArgs),
```
yuja, line 330: nit: naming: `External(CommandNameAndArgs)`?

cli/src/config-schema.json
```diff
@@ -108,6 +108,35 @@
                     "description": "Pager to use for displaying command output",
                     "default": "less -FRX"
                 },
+                "builtin-pager": {
+                    "type": "object",
+                    "description": "':builtin' pager configuration",
+                    "properties": {
+                        "alternate-screen": {
```
yuja, line 115: nit: it sounds like "(use) alternate screen" (noun). How about `reset-screen` or `clear-screen`?

Oh, but `never` means "no init" + "quit automatically". I feel "interface-mode" is better in that case.

cli/src/ui.rs
```diff
@@ -81,10 +81,18 @@ impl UiOutput {
         Ok(UiOutput::Paged { child, child_stdin })
     }
 
-    fn new_builtin_paged() -> streampager::Result<UiOutput> {
+    fn new_builtin_paged(config: &StreampagerConfig) -> streampager::Result<UiOutput> {
+        // This uselessly reads ~/.config/streampager/streampager.toml, even
+        // though we then override the important options. TODO(ilyagr): Fix this
+        // once/if https://github.com/facebook/sapling/pull/1011 is merged.
         let mut pager = streampager::Pager::new_using_stdio()?;
```
yuja, line 88: I think it still make sense to load the config file. There may be key bindings, etc.?

cli/src/ui.rs
```diff
@@ -266,9 +274,59 @@ pub enum PaginationChoice {
...
+impl StreampagerConfig {
+    fn streampager_interface_mode(&self) -> streampager::config::InterfaceMode {
+        use streampager::config::InterfaceMode;
+        use StreampagerAlternateScreenMode::*;
+        match self.alternate_screen {
+            // InterfaceMode::Direct not implemented
+            Always => InterfaceMode::FullScreen,
+            Never => InterfaceMode::Hybrid,
+            IfLongOrSlow => InterfaceMode::Delayed(std::time::Duration::from_millis(
+                self.long_or_slow_delay_millis,
```
yuja, line 321: nit: no idea if we can generalize this mode. If it's not common, maybe we can omit the `Delayed` mode?

cli/src/ui.rs
ilyagr: I would prefer to copy any such config into `jj`, especially since streampager's current docs at https://github.com/markbt/streampager are both incomplete and seem to be incorrect (as we discovered in the other thread, setting `streampager.toml` to the values shown there didn't seem to work).

It seems difficult at the moment to improve streampager's own configuration and docs. If the situation changes later, and if more programs start using streampager, we could change our mind.

cli/src/ui.rs
ilyagr: Thanks, I was guessing you might know what's best here.

cli/src/ui.rs
ilyagr: I'm undecided. 

One concern I have is that even if we remove it now, if we also simplify the config option names accordingly, it would be difficult to add the Delayed mode back in if we need it. Perhaps this could be solved somehow.

For me, the upshot of the Delayed mode is that it addresses the downside of the default mode when using it to read long texts (e.g. manual pages) where it leaves the last page of what you were reading on on the screen, while still being useable for short outputs. I am not sure how many people would like/want this tradeoff.

A part of this problem could be addressed by #5311, making the Delayed mode less important.

cli/src/ui.rs
ilyagr: Sure, why not.

I could also implement https://doc.rust-lang.org/std/convert/trait.TryFrom.html, but I'm not sure that'd be much more helpful.

ilyagr:
Thank you for looking this over!

cli/src/ui.rs
yuja: > I would prefer to copy any such config into jj, 

In that case, I would add config table specific to "streampager", not generic "builtin". It's unlikely all config knobs can be generalized.

I don't have a strong opinion about that. The current config loading at streampager side is broken, but we can send a fix to upstream. OTOH, if we fixed the upstream config deserialization implementation, maybe we can reuse it to load parameters from our TOML table.

ilyagr: I changed the key to `ui.streampager`. Perhaps we should also rename the pager to `:builtin-streampager`, with `:builtin` being an alias.

>  OTOH, if we fixed the upstream config deserialization implementation, maybe we can reuse it to load parameters from our TOML table.

That's what I was thinking.

cli/src/config-schema.json
ilyagr: I changed it to `clear-screen`.

If we went with `interface-mode`, I'm not sure what the options should be so that they are not too long and clear. I don't like the choices Streampager made for these names.

But I'm unsure what's best here.

cli/src/ui.rs
yuja: > I changed the key to `ui.streampager`. Perhaps we should also rename the pager to `:builtin-streampager`, with `:builtin` being an alias.

I have no strong opinion about that. It's unlikely we would add more than one builtin pagers. FWIW, the syntax is `:<internal-name>`, so the canonical name would be `:streampager`.

> > OTOH, if we fixed the upstream config deserialization implementation, maybe we can reuse it to load parameters from our TOML table.
> 
> That's what I was thinking.

In that case, shouldn't we follow the upstream naming convention (e.g. `streampager.interface_mode = "delayed"`)? Otherwise, we'll need to reimplement deserializer.

https://github.com/markbt/streampager?tab=readme-ov-file#configuration

ilyagr: I was thinking of doing it only in those cases where Streampager's names seem good for us. For example, keybindings maybe.

Re Streampager's names: the names "interface_mode", "delayed", and "hybrid" all seem unclear to me. What's delayed? Is "hybrid" more hybrid than "delayed"? (Also, to explain "hybrid", you have to explain the "direct" mode, which is otherwise pretty useless) And `interface_mode` could be a setting for changing anything at all.

Let me know if you prefer Streampager's choices, or can think of something better. I'm not 100% happy with my choices, but together with the text in the documentation, I'm hoping they will be easy for people to wrap their head around.

yuja: >  doing it only in those cases where Streampager's names seem good for us. For example, keybindings maybe.

Ok.

> Re Streampager's names:

I agree it's unclear, but I think it's more correct than `clear-screen = never|always|if-long-or-slow`. It's hybrid in a way that the screen is not cleared *and* the pager exits immediately depending on the content length. (BTW, this part is what I don't like about streampager. I just want to turn off the screen reset as `less -RX` would do.)

ilyagr: I thought `clear-screen="never"` is better than `full-screen="never"`, because, for long files, you do get a full-screen interface even if you choose "never" (but the screen is not cleared).

The way I think of "clear-screen" is that it determines when the screen is cleared, and then there is the additional behavior that whenever the screen is *not* cleared, streampager exits on short inputs. I tried explaining this in config.md, I'm not sure whether it's enough to get people to not get confused.

I find `echo aa | LESS= less -RX` a bit weird, especially if you press `up` and `q` after it, which leaves a ton of `~`s on the screen. So, I sort of understand why streampager chose its set of tradeoffs instead of less's, and from this perspective behaving as I described in the previous paragraph sort-of makes sense to me. From other perspectives, it's a headache.

yuja: I misunderstood "never" would be mapped to `Direct`. So, yes, it's kinda `full-screen="if-long"`, and there's no "never".

To me, "clear/reset screen" sounds tightly related to terminal reset behavior, and I wouldn't expect it affects the interactive UI. Maybe it can be `interactive=always|if-long|if-long-or-slow` or the inverse `auto-quit=never|if-one-screen|if-one-screen-and-quick`?

ilyagr: I like these better, but they do not indicate the fact that "always" always clears the screen on exit, "if-long" never clears it on exit, and "if-long-or-slow" clears it only if the pager becomes interactive. When expressed in this way, I find it hard to wrap my head around.

There might be some way to present this information in a memorable way, similar to my attempt described in https://github.com/jj-vcs/jj/pull/5415#discussion_r1934515205 (that relied on focusing the user's attention on whether the screen is cleared). Then, I'd be happy to switch. I don't have it in mind right now, but we have a few days to think about it.

One way that comes to mind, but is probably not great, is `interactive=always|if-long-and-never-clear-screen|if-long-or-slow`.

yuja: > but they do not indicate the fact that "always" always clears the screen on exit,

Correct. I just assumed UI difference is more important to users than screen reset.

> "if-long" never clears it on exit, and "if-long-or-slow" clears it only if the pager becomes interactive.

Oh, weird. Maybe better to provide two separate flags then (and error out on unsupported combination)?

(I also think copying the upstream `interface-mode` is okay given that this flag is super abstract and not possible to describe in one word.)

ilyagr: I've also been thinking about providing two flags, but I'm leaning against it. It would make the delay flag hard to name. Here's what I'm getting for this option (that I don't really like):

```toml
[ui.streampager]
quit-if-one-screen = true  # default
clear-screen-on-exit = false # default false, has no effect if quit-if-one-screen = false

# Only relevant if quit-if-one-screen and clear-screen-on-exit are both true
quit-if-one-screen-delay-millis = 2000 # (defaults to 2 seconds)
```

We could also report an error if `quit-if-one-screen=false` and `clear-screen-on-exit=false`, but this would be annoying if you tried to use `jj --config ui.streampager.quit-if-one-screen=false` and got an error.

-----------------

Here's the best option I have at this point:

```toml
[ui.streampager]
# Do not clear screen on exit. Use a full-screen interface for
# long output only. Like `less -FX`.
interface = "quit-if-one-page"  # (default).
# Always use a full-screen interface, ask the terminal to
# clear the screen on exit. Like `less +F +X`.
interface = "full-screen-clear-output"
# Use the alternate screen if the input is either long or takes
# more than `quit-quickly-delay-millis` to finish. Similar but not
# identical to `less -F +X`
interface = "quit-quickly-or-clear-output"

# Only relevant in the `quit-quickly-or-clear-output` mode
quit-quickly-delay-millis = 2000 # (defaults to 2 seconds)
```

I prefer this to the upstream's version (where the above options would correspond to, in order, "hybrid", "full-screen", and "delayed").

WDYT?

yuja: > Here's the best option I have at this point:

Looks good, but to be clear, I don't think it's any better than the streampager's naming. I'll probably need to read the documentation to understand the behavior anyway. (and I tend to use whatever the upstream provides in that case.)

fwiw, I would change the duration unit to "seconds" by default, and allow optional unit suffix or floating-point number.

ilyagr: > Looks good, but to be clear, I don't think it's any better than the streampager's naming.

Well, it's the best I could do. I think it's a little better than Streampager's.

> I would change the duration unit to "seconds" by default, and allow optional unit suffix or floating-point number.

This is a bit of a struggle. I could make it non-configurable for now.

Do we parse such suffixes anywhere?

One option is to do it manually and allow `ms` and `s` suffixes.

For floating points, I tried https://github.com/ilyagr/jj/commit/serde-float, but it results in a weird error when I try to use it with `--config`.

ilyagr: I made the duration non-configurable for now.

yuja: > I made the duration non-configurable for now.

Sounds good.

For parsing, we can use `config.get_value_with(name, parsing_fn)`. See `snapshot.max-new-file-size` handling.

cli/src/ui.rs
```diff
@@ -258,12 +257,32 @@ pub enum PaginationChoice {
...
+enum PagerConfig {
+    Disabled,
+    Builtin,
+    External(CommandNameAndArgs),
+}
+
+impl PagerConfig {
+    fn from_config(config: &StackedConfig) -> Result<PagerConfig, ConfigGetError> {
+        let PaginationChoice::Auto = config.get("ui.paginate")? else {
+            return Ok(PagerConfig::Disabled);
```
yuja, line 328: nit: `if matches!(.., PaginationChoice::Auto)` looks slightly better.

cli/src/ui.rs
```diff
@@ -258,12 +257,32 @@ pub enum PaginationChoice {
...
+        };
+        let pager_cmd: CommandNameAndArgs = config.get("ui.pager")?;
+        match pager_cmd {
```
yuja, line 331: nit: this can be inlined as `match config.get("ui.pager")? { ... pager_cmd => Ok(..) }`

cli/src/ui.rs
```diff
@@ -257,9 +266,59 @@ pub enum PaginationChoice {
     Auto,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, serde::Deserialize)]
+#[serde(rename_all(deserialize = "kebab-case"))]
+pub enum StreampagerAlternateScreenMode {
```
yuja, line 271: nit: `StreampagerInterface` or `StreampagerInterfaceMode`?

docs/config.md
```diff
@@ -598,6 +598,46 @@ paginate = "auto"
...
+features like word-wrapping are disabled.
+
+
+```toml
+[ui.streampager]
+# Do not clear screen on exit. Use a full-screen interface for long
+# output only. Like `less -FX`.
+interface = "quit-if-one-page"  # (default).
+# Always use a full-screen interface, ask the terminal to clear the
+# screen on exit. Like `less +F +X`.
```
yuja, line 649: Maybe you meant `less` or `less -+FX`? Appears that `+` takes command name.

docs/config.md
```diff
@@ -598,6 +598,46 @@ paginate = "auto"
...
+                       # left and right like `less -S`
+```
+
+#### Auto-exit, clearing the screen on startup or exit
+
+You can configure whether the pager clears the screen on startup or exit, and
+whether it quits automatically on short inputs. When the pager auto-quits,
+features like word-wrapping are disabled.
+
+
```
yuja, line 642: nit: double blank lines, remove one

yuja (approved):
Thanks.

docs/config.md
ilyagr: Thank you! TIL. I was wondering why `less -F +X` didn't do *quite* what I thought it should, but I convinced myself this was because of some weird precedence issue between my `LESS` and the arguments. 😆 

This should indeed be `less -+FX`. I'm not a huge fan of this syntax, but at least it works.

cli/src/ui.rs
ilyagr: OK, it's now

```
        if !matches!(config.get("ui.paginate")?, PaginationChoice::Auto) {
```

ilyagr:
Thank you Yuya! I'll merge this in an hour or two unless I hear otherwise.

cli/src/ui.rs
yuja: Oh, I should say `if matches!(.., Never)`. If you don't have any preference, I would use `if never => disabled`.

ilyagr: Sure, SG.

PR #5419, opened 2025-01-21
Add a --edit flag

### LINKED ISSUE #5379: FR: --edit flag for describe
I use script that creates an AI summary from the diff for `jj describe`. Currently I'm doing it like this:  `jujutsu-ai-summarize-diff | jj describe --stdin`. But it would be great that I would have an option to go through the generated change description in the editor. Currently we only have a flag `--no-edit` but no flag for situation where I want to force the editor to open.

I would like to have a flag `--edit` or similar to force the change description to open in the editor. 

I checked could I fill the `draft_commit_description` template with stdin input but seems not to be possible(?)

MrJohz:
The --edit flag forces the editor to be shown, even if it would have been hidden due to the --no-edit flag or another flag that implies it (such as --message or --stdin).

This fixes #5379  

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/describe.rs
+++ b/cli/src/commands/describe.rs
@@ -55,7 +55,12 @@ pub(crate) struct DescribeArgs {
     ///
     /// If multiple revisions are specified, the same description will be used
     /// for all of them.
-    #[arg(long = "message", short, value_name = "MESSAGE")]
+    #[arg(
+        long = "message",
+        short,
+        value_name = "MESSAGE",
+        conflicts_with = "stdin"
+    )]
     message_paragraphs: Vec<String>,
     /// Read the change description from stdin
     ///
--- a/cli/src/commands/describe.rs
+++ b/cli/src/commands/describe.rs
@@ -66,8 +66,14 @@ pub(crate) struct DescribeArgs {
     /// Don't open an editor
     ///
     /// This is mainly useful in combination with e.g. `--reset-author`.
-    #[arg(long)]
+    #[arg(long, conflicts_with = "edit")]
     no_edit: bool,
+    /// Open an editor
+    ///
+    /// Forces an editor to open when using `--stdin` or `--message` to
+    /// allow the message to be edited afterwards.
+    #[arg(long)]
+    edit: bool,
     /// Reset the author to the configured user
     ///
     /// This resets the author name, email, and timestamp.
@@ -133,7 +139,12 @@ pub(crate) fn cmd_describe(
         None
     };
 
-    let commit_descriptions: Vec<(_, _)> = if args.no_edit || shared_description.is_some() {
+    // edit and no_edit are conflicting arguments and therefore it should not
+    // be possible for both to be true at the same time.
+    assert!(!(args.edit && args.no_edit));
+    let use_editor = args.edit || (shared_description.is_none() && !args.no_edit);
+
+    let commit_descriptions: Vec<(_, _)> = if !use_editor {
         commits
             .iter()
             .map(|commit| {
@@ -150,7 +161,9 @@ pub(crate) fn cmd_describe(
             .rev()
             .map(|commit| -> Result<_, CommandError> {
                 let mut commit_builder = tx.repo_mut().rewrite_commit(commit).detach();
-                if commit_builder.description().is_empty() {
+                if let Some(description) = &shared_description {
+                    commit_builder.set_description(description);
+                } else if commit_builder.description().is_empty() {
                     commit_builder
                         .set_description(tx.settings().get_string("ui.default-description")?);
                 }
--- a/cli/tests/test_describe_command.rs
+++ b/cli/tests/test_describe_command.rs
@@ -781,6 +781,74 @@ fn test_describe_avoids_unc() {
     assert_eq!(edited_path, dunce::simplified(&edited_path));
 }
 
+#[test]
+fn test_describe_with_edit_and_message_args_opens_editor() {
+    let mut test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let workspace_path = test_env.env_root().join("repo");
+
+    let edit_script = test_env.set_up_fake_editor();
+    std::fs::write(edit_script, ["dump editor"].join("\0")).unwrap();
+    let (stdout, stderr) = test_env.jj_cmd_ok(
+        &workspace_path,
+        &["describe", "-m", "message from command line", "--edit"],
+    );
+    insta::assert_snapshot!(stdout, @"");
+    insta::assert_snapshot!(stderr, @r"
+    Working copy now at: qpvuntsm 61ece7a9 (empty) message from command line
+    Parent commit      : zzzzzzzz 00000000 (empty) (no description set)
+    ");
+    insta::assert_snapshot!(
+        std::fs::read_to_string(test_env.env_root().join("editor")).unwrap(), @r#"
+    message from command line
+
+    JJ: Lines starting with "JJ:" (like this one) will be removed.
+    "#);
+}
+
+#[test]
+fn test_describe_change_with_existing_message_with_edit_and_message_args_opens_editor() {
+    let mut test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let workspace_path = test_env.env_root().join("repo");
+
+    test_env.jj_cmd_ok(&workspace_path, &["describe", "-m", "original message"]);
+
+    let edit_script = test_env.set_up_fake_editor();
+    std::fs::write(edit_script, ["dump editor"].join("\0")).unwrap();
+    let (stdout, stderr) = test_env.jj_cmd_ok(
+        &workspace_path,
+        &["describe", "-m", "new message", "--edit"],
+    );
+    insta::assert_snapshot!(stdout, @"");
+    insta::assert_snapshot!(stderr, @r"
+    Working copy now at: qpvuntsm de694560 (empty) new message
+    Parent commit      : zzzzzzzz 00000000 (empty) (no description set)
+    ");
+    insta::assert_snapshot!(
+        std::fs::read_to_string(test_env.env_root().join("editor")).unwrap(), @r#"
+    new message
+
+    JJ: Lines starting with "JJ:" (like this one) will be removed.
+    "#);
+}
+
+#[test]
+fn test_edit_cannot_be_used_with_no_edit() {
+    let test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let workspace_path = test_env.env_root().join("repo");
+
+    let stderr = test_env.jj_cmd_cli_error(&workspace_path, &["describe", "--no-edit", "--edit"]);
+    insta::assert_snapshot!(stderr, @r"
+    error: the argument '--no-edit' cannot be used with '--edit'
+
+    Usage: jj describe --no-edit [REVSETS]...
+
+    For more information, try '--help'.
+    ");
+}
+
 fn get_log_output(test_env: &TestEnvironment, repo_path: &Path) -> String {
     let template = r#"commit_id.short() ++ " " ++ description"#;
     test_env.jj_cmd_success(repo_path, &["log", "-T", template])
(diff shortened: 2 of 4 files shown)


arxanas:
In other contexts, `--edit` means "switch to the commit as the new working copy and start editing it", which is inconsistent with its usage here (and the current usage of `--no-edit`). Maybe it should be something like `--editor`/`--no-editor`?

MrJohz:
That makes sense to me and I'm happy to make the change.  To briefly play devil's advocate:

* That makes this a breaking change for anyone who's already using `--no-edit` in scripts etc
* Both Git and Mercurial use `--edit` (and `--no-edit` for Git) to invoke/avoid invoking the editor.

But internal consistency is more important than consistency with those other tools, so I'm happy to make the change if you/other people would prefer it that way.

PhilipMetzger:
What about a `--open-editor`? Also adhere to the commit style outlined here: https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#code-reviews

cli/src/commands/describe.rs
```diff
@@ -68,6 +68,12 @@ pub(crate) struct DescribeArgs {
     /// This is mainly useful in combination with e.g. `--reset-author`.
     #[arg(long)]
     no_edit: bool,
+    /// Open an editor
+    ///
+    /// This is mainly useful to reverse the effects of --no-edit or another
+    /// flag/option that implies --no-edit e.g. --message
+    #[arg(long)]
```
arxanas, line 75: suggestion: Should this conflict with `--no-edit`? (See e.g. https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with, or there's probably examples in other places we have conflicting arguments).

suggestion (non-blocking): Hmm, it seems like `--message` and `--stdin` should also be marked as conflicting, since one is silently ignored if the other is passed, so if you have free time, that could be a good change to make as well 🙂.

cli/src/commands/describe.rs
```diff
@@ -133,7 +139,13 @@ pub(crate) fn cmd_describe(
         None
     };
 
-    let commit_descriptions: Vec<(_, _)> = if args.no_edit || shared_description.is_some() {
+    let skip_editor = if args.edit {
+        false
+    } else {
+        args.no_edit || shared_description.is_some()
+    };
```
arxanas, line 146: suggestion: After marking `--edit`/`--no-edit` as conflicting as per above, we might want to add an assertion here to ensure that both bools didn't somehow get set.

suggestion: It might be better to use a "positive" form of the name rather than a "negative" form, such as `use_editor` (and then either check `!use_editor` instead of `skip_editor` or reorder conditions as appropriate).

cli/src/commands/describe.rs
```diff
@@ -151,8 +163,12 @@ pub(crate) fn cmd_describe(
...
                 let mut commit_builder = tx.repo_mut().rewrite_commit(commit).detach();
                 if commit_builder.description().is_empty() {
-                    commit_builder
-                        .set_description(tx.settings().get_string("ui.default-description")?);
+                    let description = if let Some(description) = &shared_description {
+                        description.into()
+                    } else {
+                        tx.settings().get_string("ui.default-description")?
+                    };
+                    commit_builder.set_description(description);
```
arxanas, line 171:
```suggestion
                    commit_builder.set_description(match shared_description {
                        Some(description) => description.to_owned(),
                        None => tx.settings().get_string("ui.default-description")?,
                    });
```

suggestion: I recommend not using `into` since it's not obvious (without IDE support) what type conversion is happening, and it can result in unexpected behavior in future refactorings. If this is meant to convert from `&str` to `String`, then maybe `.to_owned()` is best here.

nit: My preferred way of writing this might be via an inline `match`, but not a big deal if you prefer an explicit variable and `if`.

cli/tests/test_describe_command.rs
```diff
@@ -781,6 +781,63 @@ fn test_describe_avoids_unc() {
...
+    JJ: This commit contains the following changes:
+    JJ:     A file1
+    JJ:     A file2
+
+    JJ: Lines starting with "JJ:" (like this one) will be removed.
+    "#);
+}
+
+#[test]
+fn test_edit_flag_overrides_no_edit_flag() {
```
arxanas, line 816: comment (non-blocking): I suggested above that the arguments should conflict (which will cause `clap` to reject them for us). Let me know if you think `--edit` or `--no-edit` should actually override each other.

cli/tests/test_describe_command.rs
```diff
@@ -781,6 +781,63 @@ fn test_describe_avoids_unc() {
     assert_eq!(edited_path, dunce::simplified(&edited_path));
 }
 
+#[test]
+fn test_describe_opens_editor_if_edit_flag_overrides_default_behaviour() {
```
arxanas, line 785: question: What "default" behavior is this overriding?

arxanas (approved):
Looks good to me overall, thanks for implementing this!

Re `--editor` vs `--no-editor`, I also asked in [this Discord message](https://discord.com/channels/968932220549103686/968932220549103689/1331303202565525545). It seems like there's many factors to consider, so let's not block this PR on that. I'll open a separate discussion.

cli/src/commands/describe.rs
MrJohz: 1. I thought about that, but decided against it because it's not uncommon to have `--x`/`--no-x` flags that can both be used and can override each other.  I also implemented the change such that `--edit` explicitly overrides `--no-edit` (regardless of order), and added a test to that effect.  I could be convinced otherwise though!
2. That's a good idea, I'll add that.  I should add that in a separate commit in this PR, right?

cli/tests/test_describe_command.rs
MrJohz: The default behaviour of not opening the editor if e.g. `--message` is passed.  But I agree that the name isn't great, what would you think about: `test_describe_with_edit_and_message_args_opens_editor()`?  That describes the contents of the test more accurately.

cli/tests/test_describe_command.rs
MrJohz: Yeah, let's maybe keep that discussion here for now.  I see four options:

1. The user can provide both flags, but `--edit` always "wins".  (This is the current state of the PR.)
2. The user can provide both flags, but the final one passed always "wins".  (This requires futzing with clap, but a quick Google suggests it should be possible.)
3. The user can provide either `--edit` or `--no-edit`, never both.  (I'd then updated the clap definition, change the tests, add an assertion etc.)
4. There is an `--editor [skip|open|auto]` flag, which can only be handled once.  (This is the breaking change option.)

The more I think about it, the more I kind of like option (4), because it's nice and explicit, even if it's more verbose.  Of the other options, I'm leaning somewhat towards (2) because that's Git's behaviour and it allows setting defaults as aliases, but I'm easy.  I'll sleep on this tonight and if nobody has strong opinions any way, I'll implement (2) tomorrow at some point.

cli/src/commands/describe.rs
MrJohz: --stdin and --message are now marked as conflicting.  I'll resolve this given that the rest of the comment is being discussed elsewhere.

cli/tests/test_describe_command.rs
arxanas: What's the workflow that option 2 supports? I would be fairly concerned that it was user error. We can also just use the strictest option for now and loosen it later without breaking any user workflows, if it turns out to be annoying in practice.

I don't think we need to be overly concerned with the Git equivalent UI:
- For this case, I can see that users may have some muscle memory like `git commit --amend --no-edit`, but it doesn't seem like it would carry over to `jj describe` to begin with.
- For this case, where Git has last-wins conflicting arguments, is there a common workflow where it's relevant?
- Generally: There are many places where our CLI diverges from Git's already.
  - Example: rebase's `--keep-emptied` vs `--keep-empty`, which IIRC was changed specifically to resolve a certain ambiguity in the flag name.
  - So we can continue to make small adjustments if there's a good reason. (Although I think this case is a little weaker.)
- Generally: The Git UI is often inconsistent or dangerous, so not always worth modeling 👀

MrJohz: Okay, given that the flags might well change at some point anyway, and given that we can add the "keep latest" concept later on, I'll implement option (3).

cli/tests/test_describe_command.rs
MrJohz: Now using `test_describe_with_edit_and_message_args_opens_editor`

cli/tests/test_describe_command.rs
MrJohz: Option (3) is now implemented

cli/src/commands/describe.rs
MrJohz: I went with `use_editor` in the end, because I think it makes the calculation a bit easier to read (it's now `use_editor = args.edit || (shared_description.is_none() && !args.no_edit)`, which feels roughly like how a human would describe the condition in natural langauge).  The if-condition is now negated, which isn't great, but the alternative would be to swap around the if block and the else block, and the else block is so large that that would be unreadable.

I've added an assertion, plus a comment to clarify what exactly the assertion is there.

MrJohz:
What's the convention here for who merges, when that happens, etc?  I think this is ready to go now.  @arxanas

cli/tests/test_describe_command.rs
```diff
@@ -781,6 +781,92 @@ fn test_describe_avoids_unc() {
...
 }
 
+#[test]
+fn test_describe_with_edit_and_message_args_opens_editor() {
+    let mut test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let workspace_path = test_env.env_root().join("repo");
+
+    std::fs::write(workspace_path.join("file1"), "foo\n").unwrap();
+    std::fs::write(workspace_path.join("file2"), "bar\n").unwrap();
```
martinvonz, line 791: nit: these don't seem to be needed for the test. same in the next test case

cli/tests/test_describe_command.rs
```diff
@@ -781,6 +781,92 @@ fn test_describe_avoids_unc() {
...
+    JJ: This commit contains the following changes:
+    JJ:     A file1
+    JJ:     A file2
+
+    JJ: Lines starting with "JJ:" (like this one) will be removed.
+    "#);
+}
+
+#[test]
+fn test_describe_change_with_existing_message_with_edit_and_message_args_opens_editor() {
```
martinvonz, line 816: nit: I don't think we need both this test and the previous one. They're basically the same, but this one uses "original message" as initial description and the other test case uses "" (empty string) as initial description. They seem to take the same code paths too. But I don't feel strongly - feel free to leave both tests if you prefer.

cli/tests/test_describe_command.rs
```diff
@@ -781,6 +781,92 @@ fn test_describe_avoids_unc() {
...
+#[test]
+fn test_edit_cannot_be_used_with_no_edit() {
+    let mut test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let workspace_path = test_env.env_root().join("repo");
+
+    std::fs::write(workspace_path.join("file1"), "foo\n").unwrap();
+    std::fs::write(workspace_path.join("file2"), "bar\n").unwrap();
+    let edit_script = test_env.set_up_fake_editor();
+    std::fs::write(&edit_script, "dump editor").unwrap();
```
martinvonz, line 857: Can delete

cli/tests/test_describe_command.rs
```diff
@@ -781,6 +781,92 @@ fn test_describe_avoids_unc() {
...
+    let stderr = test_env.jj_cmd_cli_error(&workspace_path, &["describe", "--no-edit", "--edit"]);
+    insta::assert_snapshot!(stderr, @r"
+    error: the argument '--no-edit' cannot be used with '--edit'
+
+    Usage: jj describe --no-edit [REVSETS]...
+
+    For more information, try '--help'.
+    ");
+
+    assert!(!test_env.env_root().join("editor").exists());
```
martinvonz, line 867: I don't think we need to check this. The message above is a strong indication that the error happened while parsing arguments, so we should not have gotten to the point where we call an editor.

cli/tests/test_describe_command.rs
MrJohz: I've kept both tests.  When I first implemented this, I made a mistake and only covered one of the cases (the empty string case).  When I fixed that, I made it so that the code behaves the same in both cases, which means that theoretically you should only need one of the tests, but I figure leaving both helps in case someone else makes the same mistake I did.

PR #5423, opened 2025-01-21
cli: warn user if they try to `jj bookmark set bookmark@remote`

0xdeafbeef:
I'm not sure, maybe there is a legit use case for having `@` in branch name, but most likely it's a mistake.

It's also not very consistent with other commands.
You can `jj rebase -d br@origin`, but you can't` jj bookmark set br@origin` after rebasing(I was lazy importing bookmark).

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * New `subject(pattern)` revset function that matches first line of commit
   descriptions.
+* Warn user if they try to use `jj boookmark set branch@remote`
 
 ### Fixed bugs
 
--- a/cli/src/commands/bookmark/mod.rs
+++ b/cli/src/commands/bookmark/mod.rs
@@ -27,6 +27,7 @@ use jj_lib::backend::CommitId;
 use jj_lib::op_store::RefTarget;
 use jj_lib::op_store::RemoteRef;
 use jj_lib::repo::Repo;
+use jj_lib::revset::ExpressionKind;
 use jj_lib::str_util::StringPattern;
 use jj_lib::view::View;
 
@@ -51,6 +52,7 @@ use self::untrack::BookmarkUntrackArgs;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::RemoteBookmarkName;
 use crate::cli_util::RemoteBookmarkNamePattern;
+use crate::cli_util::WorkspaceCommandHelper;
 use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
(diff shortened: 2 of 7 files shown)


PhilipMetzger:
Thanks for doing this:  It would be nice if you also the `{set, create}` errors as discussed on [Discord][dc1] and [here][dc2].


[dc1]: https://discord.com/channels/968932220549103686/968932220549103689/1324399998066098196
[dc2]: https://discord.com/channels/968932220549103686/968932220549103689/1324430885990432769

cli/src/commands/bookmark/set.rs
```diff
@@ -124,3 +128,43 @@ pub fn cmd_bookmark_set(
...
+        } else {
+            "Suspicious bookmark name:"
+        });
+        if has_at.len() > 1 {
+            writeln!(writer)?;
+        }
+        for name in &has_at {
+            writeln!(writer, "  {name}")?;
+        }
+        drop(writer);
```
yuja, line 156: nits: 
* just use `ui.warning_default()` to print "Warning: "?
* we generally avoid ad-hoc singular/plural handling. just use plural form.
* print bookmarks in line (with `.join(", ")`)? it's unlikely there are bunch of bookmarks to be created

cli/src/commands/bookmark/set.rs
```diff
@@ -124,3 +128,43 @@ pub fn cmd_bookmark_set(
...
+
+        let mut hint = ui.hint_default();
+        writeln!(
+            hint,
+            "Bookmark name{name} should not contain '@'.\nTo track remote bookmark{bm}, use 'jj \
```
yuja, line 161: nit: "should" might be too strong. It's not wrong to include `@` in bookmark.

cli/tests/test_bookmark_at.rs
yuja, line 1: nit: I would merge the tests into `test_bookmark_command.rs`

yuja (commented):
LG.

Another option is to parse bookmark name as revset symbol, so the user will need to quote bookmark name as `'"foo@bar"' if they really want to create a local bookmark of that name.

martinvonz:
> Another option is to parse bookmark name as revset symbol, so the user will need to quote bookmark name as `'"foo@bar"' if they really want to create a local bookmark of that name.

That would also catch cases like `jj bookmark set foo-` (which also creates a bookmark that's hard to refer to in revsets). That would be good. Feel like trying that @0xdeafbeef?

0xdeafbeef:
> > Another option is to parse bookmark name as revset symbol, so the user will need to quote bookmark name as `'"foo@bar"' if they really want to create a local bookmark of that name.
> 
> That would also catch cases like `jj bookmark set foo-` (which also creates a bookmark that's hard to refer to in revsets). That would be good. Feel like trying that @0xdeafbeef?

i've added revset parsing

cli/src/commands/bookmark/set.rs
0xdeafbeef: done

cli/src/commands/bookmark/mod.rs
```diff
@@ -201,3 +203,92 @@ fn is_fast_forward(repo: &dyn Repo, old_target: &RefTarget, new_target_id: &Comm
...
+
+enum BookMarkKind {
+    RevsetLike,
+    RemoteLike,
+    Normal,
+}
+
+impl BookMarkKind {
+    fn parse(name: &str, helper: &WorkspaceCommandHelper) -> Result<Self, CommandError> {
+        let node = helper.parse_revset_expression(name)?;
```
yuja, line 276: Perhaps, we can add public `revset::parse_symbol() -> Result<String, ..>` function that does `parse_program()` and `expect_literal()` (or reexport these functions.) Revset aliases shouldn't be expanded, and the other context parameters are unneeded.

Maybe we can add error hint by inspecting the source string? It might be easier if parsed tree (= result of `parse_program()`) is available, but I'm not sure.

cli/src/commands/bookmark/mod.rs
```diff
@@ -201,3 +203,92 @@ fn is_fast_forward(repo: &dyn Repo, old_target: &RefTarget, new_target_id: &Comm
...
+            writeln!(writer, "  {name}")?;
+        }
+        drop(writer);
+
+        let mut hint = ui.hint_default();
+        writeln!(
+            hint,
+            "To track remote bookmarks, use: jj bookmark --track {}",
+            remote_like.join(" ")
+        )?;
```
yuja, line 238: Unparsable name can be just an error, not a warning. I think that's the idea of using revset parser. User can use escape syntax to specify weird bookmark name.

CHANGELOG.md
```diff
@@ -115,6 +115,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * New `subject(pattern)` revset function that matches first line of commit
   descriptions.
+* Warn user if he tries to use `jj boookmark set branch@remote`
```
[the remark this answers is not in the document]
0xdeafbeef: done

cli/src/commands/bookmark/mod.rs
```diff
@@ -201,3 +203,92 @@ fn is_fast_forward(repo: &dyn Repo, old_target: &RefTarget, new_target_id: &Comm
...
+    }
+
+    if !remote_like.is_empty() {
+        let mut writer = ui.warning_with_heading("Bookmarks containing '@':");
+        for name in &remote_like {
+            writeln!(writer, "  {name}")?;
+        }
+        drop(writer);
+
+        let mut hint = ui.hint_default();
```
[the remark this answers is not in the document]
0xdeafbeef: done

cli/src/commands/bookmark/mod.rs
0xdeafbeef: > Unparsable name can be just an error, not a warning. I think that's the idea of using revset parser. User can use escape syntax to specify weird bookmark name.

Sorry, I didn't get it. The only reason to use a revset parser is not to reinvent my own :)
You want to return an error for cases when name was sucessfully parsed as revset(except legit cases) and when it was not parsed?

martinvonz: Yes, I think Yuya just meant the since we're using the real revset parser, we can make it an error. If we had some naive check instead, we might have wanted to make it a warning so we don't prevent valid names. But I'm not 100% sure that's what he meant.

yuja: Yes.

without shell escaping:
* `foo@bar` -> error
* `"foo@bar"` -> local bookmark `foo@bar`
* `foo-` -> error
* `"foo-"` -> local bookmark `foo-`
* `(foo)` -> could be an error (but might be okay to parse as local bookmark `foo`)

0xdeafbeef:
I've updated `validate_bookmark_names`, but it looks a bit of clunky. 
Seems like we should call `validate_bookmark_names` in all bookmark subcommands, and it also should strip dquotes if they are present. If you think that it's vaible aproach i'll implement it. Sorry for slow responses, I'll ve zero time next few weeks

yuja:
> Seems like we should call `validate_bookmark_names` in all bookmark subcommands,

One thing we can do is to add some newtype `BookmarkName(String)` and let clap handle parsing. That's basically the same as `RemoteBookmarkNamePattern`, and we'll need to fix pattern parsing later. There's a TODO comment about that.

PR #5425, opened 2025-01-21
actions: fix codespell CI failure

ilyagr:
It is having a fit for some reason, see e.g.
https://github.com/jj-vcs/jj/actions/runs/12897590111/job/35963026619?pr=5414

Not sure what changed, but here is a quick fix


ilyagr:
The reason might be https://github.com/codespell-project/codespell/releases/tag/v2.4.0 . It'd be cool to fix the version of `codespell` the action uses, but I couldn't find a way to do it quickly.

martinvonz (approved):

PR #5441, opened 2025-01-23
insta: run --force-update-snapshots

bsdinis:
This was after:
 run `cargo insta test --force-update-snapshots`
 fix the raw `\r`

### DIFF
--- a/cli/tests/test_git_fetch.rs
+++ b/cli/tests/test_git_fetch.rs
@@ -87,9 +87,7 @@ fn test_git_fetch_with_default_config(subprocess: bool) {
 
     test_env.jj_cmd_ok(&repo_path, &["git", "fetch"]);
     insta::allow_duplicates! {
-    insta::assert_snapshot!(get_bookmark_output(&test_env, &repo_path), @r###"
-    origin@origin: oputwtnw ffecd2d6 message
-    "###);
+    insta::assert_snapshot!(get_bookmark_output(&test_env, &repo_path), @"origin@origin: oputwtnw ffecd2d6 message");
     }
 }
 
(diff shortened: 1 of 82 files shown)


ilyagr:
I filed https://github.com/mitsuhiko/insta/issues/712.

bsdinis:
Was running on the latest `1.42.0`

bsdinis:
@ilyagr thanks for opening the issue upstream

cli/tests/test_git_fetch.rs
```diff
@@ -330,7 +328,10 @@ fn test_git_fetch_nonexistent_remote_from_config(subprocess: bool) {
     Error: No git remote named 'rem2'
     "###);
     // No remote should have been fetched as part of the failing transaction
-    insta::assert_snapshot!(get_bookmark_output(&test_env, &repo_path), @"");
```
ilyagr, line 333: I find this diff very mysterious. How could the tests possibly be succeeding on main if the output of this test is not correct?

martinvonz: Maybe the same as #5314

ilyagr: Yes, it seems likely that splitting the `allow_duplicates` section in two would fix this. Still, if it's possible, it would be nice if it was fixed in insta.

bsdinis: Was really baffling to get such differences

bsdinis: I think I'll wait for insta to fix it.
This PR was more to share the error I was seeing, was good to get awareness.
Thanks for the debugging!

PR #5443, opened 2025-01-23
git: remove git2 from git subprocessing code paths in lib

bsdinis:
Currently, the subprocessing code paths in the library were relying on git2 to figure out if a remote exists. This is because when git errors out for a remote not existing it outputs the same error as when the repository is not found.

As the cli tool makes a distinction between the two (e.g., after cloning, the remote must exist and as such we cannot send out an error that is shared by `no remote found` and `no repository found`), this was a hack put in place to sort out the difference.

It calls out to `git remote` to list out the remotes.

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -1601,15 +1601,14 @@ impl<'a> GitFetch<'a> {
         branch_names: &[StringPattern],
         remote_name: &str,
     ) -> Result<Option<String>, GitFetchError> {
-        // check the remote exists
-        // TODO: we should ideally find a way to do this without git2
-        let _remote = self.git_repo.find_remote(remote_name).map_err(|err| {
-            if is_remote_not_found_err(&err) {
-                GitFetchError::NoSuchRemote(remote_name.to_string())
-            } else {
-                GitFetchError::InternalGitError(err)
-            }
-        })?;
+        let git_ctx =
+            GitSubprocessContext::from_git2(self.git_repo, &self.git_settings.executable_path);
+
+        let remotes = git_ctx.spawn_remote()?;
+        if !remotes.contains(remote_name) {
+            return Err(GitFetchError::NoSuchRemote(remote_name.to_string()));
+        }
+
         // At this point, we are only updating Git's remote tracking branches, not the
         // local branches.
         let mut remaining_refspecs: Vec<_> = Self::expand_refspecs(remote_name, branch_names)?;
@@ -1617,9 +1616,6 @@ impl<'a> GitFetch<'a> {
             // Don't fall back to the base refspecs.
             return Ok(None);
         }
-
-        let git_ctx =
-            GitSubprocessContext::from_git2(self.git_repo, &self.git_settings.executable_path);
 
         let mut branches_to_prune = Vec::new();
         // git unfortunately errors out if one of the many refspecs is not found
@@ -2019,16 +2015,12 @@ fn subprocess_push_refs(
     qualified_remote_refs_expected_locations: &HashMap<&str, Option<&CommitId>>,
     refspecs: &[RefSpec],
 ) -> Result<(), GitPushError> {
-    // check the remote exists
-    // TODO: we should ideally find a way to do this without git2
-    let _remote = git_repo.find_remote(remote_name).map_err(|err| {
-        if is_remote_not_found_err(&err) {
-            GitPushError::NoSuchRemote(remote_name.to_string())
-        } else {
-            GitPushError::InternalGitError(err)
-        }
-    })?;
     let git_ctx = GitSubprocessContext::from_git2(git_repo, git_executable_path);
+    // check the remote exists
+    let remotes = git_ctx.spawn_remote()?;
+    if !remotes.contains(remote_name) {
+        return Err(GitPushError::NoSuchRemote(remote_name.to_string()));
+    }
 
     let mut remaining_remote_refs: HashSet<_> = qualified_remote_refs_expected_locations
         .keys()
--- a/lib/src/git_subprocess.rs
+++ b/lib/src/git_subprocess.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+use std::collections::HashSet;
 use std::num::NonZeroU32;
 use std::path::Path;
 use std::path::PathBuf;
@@ -154,6 +155,24 @@ impl<'a> GitSubprocessContext<'a> {
         let () = parse_git_branch_prune_output(output)?;
 
         Ok(())
+    }
+
+    // TODO: we should remove this in lieu of using the gix API.
+    // It is not trivial at this point because gix caches the config in memory
+    // (which is where the configured remotes live). This means that the
+    // externally modified remote is not found on clone.
+    /// List the configured remotes
+    ///
+    /// `git remote`
+    ///
+    /// Prints a remote per line
+    pub(crate) fn spawn_remote(&self) -> Result<HashSet<String>, GitSubprocessError> {
+        let mut command = self.create_command();
+        command.stdout(Stdio::piped());
+        command.arg("remote");
+        let output = self.spawn_cmd(command)?;
+
+        parse_git_remote_output(output)
     }
 
     /// How we retrieve the remote's default branch:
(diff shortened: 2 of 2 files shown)


yuja:
Can't we use `gix` API instead?
https://docs.rs/gix/latest/gix/struct.Repository.html#method.try_find_remote

It's unlikely we'll drop `gix` dependency, and the `gix` API is usually nicer and more reliable than shelling out `git` and parsing the output.

Perhaps, `GitFetch::new()` should obtain git2/gix repo instance from `mut_repo` (as we do in `import_refs()`.

bsdinis:
> EDIT: fwiw, if there are a few related small patches, I prefer including them in one PR. One PR per patch is conceptually nice, but is uneasy to manage without external tooling.
Happy to close this one and compare the top of the stack to main

bsdinis:
So on trying to use gix, I'm running into a weird thing.
When cloning an empty repo (as in https://github.com/jj-vcs/jj/blob/f3bbc50453e5674805e505e372e07963a4fa08ed/cli/tests/test_git_clone.rs#L57) `gix` reports no remotes, while suprocessing does find the correct remote

```
fetching origin
gix: {} # output of gix_repo.remote_names()
git: {"origin"} # output of `git remote`
```

bsdinis:
So on trying to use gix, I'm running into a weird thing.
When cloning an empty repo (as in https://github.com/jj-vcs/jj/blob/f3bbc50453e5674805e505e372e07963a4fa08ed/cli/tests/test_git_clone.rs#L57) `gix` reports no remotes, while suprocessing does find the correct remote

```
fetching origin
gix: {} # output of gix_repo.remote_names()
git: {"origin"} # output of `git remote`
```

Will look more into this later, but it's definitely confusing2

bsdinis:
So on trying to use gix, I'm running into a weird thing.
When cloning an empty repo (as in https://github.com/jj-vcs/jj/blob/f3bbc50453e5674805e505e372e07963a4fa08ed/cli/tests/test_git_clone.rs#L57) `gix` reports no remotes, while suprocessing does find the correct remote

```
fetching origin
gix: {} # output of gix_repo.remote_names()
git: {"origin"} # output of `git remote`
```

Will look more into this later, but it's definitely confusing

martinvonz:
Without knowing what your code looks, that sounds like it might be a bug in Gitoxide. GitHub user Byron is usually very responsive, so we can probably just loop him in here if/when you feel confident that Gitoxide is behaving strangely.

yuja:
`gix::Repository` caches config file, but we updates config file externally by using `git2` API. That's probably why.
https://docs.rs/gix/latest/gix/struct.Repository.html#method.config_snapshot

We might have to reload our `Arc<ReadonlyRepo>` after adding the remote configuration.

```rust
git::add_remote(..)?;
let repo = repo.reload_at(repo.operation())?;
command.for_workable_repo(ui, workspace, repo)
```

emilazy:
IIRC remote/config management, fetches, and pulls were the only things remaining on `git2` a year+ ago. Is that still the case? If so I could try to revive my old patch to migrate the former to Gitoxide which might help with this and pave the way to a `git2`-less build.

yuja:
> ```rust
> git::add_remote(..)?;
> let repo = repo.reload_at(repo.operation())?;
> command.for_workable_repo(ui, workspace, repo)
> ```

Oh, but the store is still reused iirc, so we'll need to reload the workspace (or `RepoLoader`)?

> IIRC remote/config management, fetches, and pulls were the only things remaining on `git2` a year+ ago. Is that still the case? If so I could try to revive my old patch to migrate the former to Gitoxide which might help with this and pave the way to a `git2`-less build.

+1

bsdinis:
I tried a very hacky way to get this to work (it didn't). 
@yuja how do you feel about keeping with the subprocessing to `git remote` in the meantime. 
It is honestly the simplest of them all to parse (it outputs a list of remote names, nothing else), and I'll add a note to remove when the `gix` problem is hopefully solved by @emilazy 's patch. 

I only ask because this is blocking a bunch of the next refactorings, which require a strict separation between the git2 and subprocessing path (which is the direction I feel we'd like to move towards)

yuja:
> how do you feel about keeping with the subprocessing to `git remote` in the meantime.

I feel this patch is going backwards because git2 API just works.

> remove when the gix problem is hopefully solved by @emilazy 's patch.

I don't think it will solve the problem. Porting remote management API to `gix` is nice, but it wouldn't mean the remote configuration can be updated without reloading a repo instance.

> this is blocking a bunch of the next refactorings, which require a strict separation between the git2 and subprocessing path (which is the direction I feel we'd like to move towards)

Oh, are you going to move git2/subprocess handling to caller? I would rather not, and leave it in library code. For example, `GitFetch` might look like:
```rust
struct GitFetch {
    inner: enum Git2(..)|Subprocess(..),
    ...
}

impl GitFetch {
    pub fn new(mut_repo, git_settings) -> Result<Self, ..> {
        get_git_backend(mut_repo.store()).map_err(..)
        ...
    }
```

yuja:
Can you try this PR? https://github.com/jj-vcs/jj/pull/5476

bsdinis:
> Oh, are you going to move git2/subprocess handling to caller? I would rather not, and leave it in library code. 

There are several good reasons to do this: 
 1. For the UNC handling, if we want the GitSubprocessContext to be constructed with the workspace to have the root and the path to git handled correctly, we need to be aware of it at the caller.
  2. `git2` is being used directly in the caller, would be good (IMO) to make it clear when it stops being needed.
  
 Overall, I think a side by side implementation needs to holistically separate, so turning one of them off should be simple

bsdinis:
> This also applies to git2 (and gix) implementation if we add one later. The problem is not UNC paths, but is that relative remote paths are resolved differently from the git CLI. If git2 provided an API to resolve configured remote by caller, we could do workspace_root.join(configured_remote_path) somewhere in the git2 code path.

The UNC problem I think is hard to solve with lib is unrelated to remotes. It's more that we have to pass in `--git-dir` to `git` and to avoid UNC problems we want that path to be relative to the workspace dir (which we would pass as `Command::current_dir`). This is the hard bit to figure out from the lib alone.

> Hmm, I have feeling that we should hide git2 use in library code so we can phase it out easily by replacing the implementation by gix or subprocessing. FWIW, I think the main reason we do instantiate git2 repo by caller is that, unlike git_repo() -> gix::Repository, open_git_repo() -> git2::Repository does non-trivial work, so the instance should be reused (by caller) if possible.

From looking at the code, `cli` is using a `git2::Repository` for doing a bunch of work. My idea with separating it at the caller is to make it explicit that, from that point on, the `git2` repository is not being used again. It would also future proof the Subprocessing path against using `git2` in the future (which I think is desirable). Then we can work on migrating the `cli` code to `gitoxide` independently.

yuja:
> > This also applies to git2 (and gix) implementation if we add one later. The problem is not UNC paths, but is that relative remote paths are resolved differently from the git CLI. If git2 provided an API to resolve configured remote by caller, we could do workspace_root.join(configured_remote_path) somewhere in the git2 code path.
> 
> The UNC problem I think is hard to solve with lib is unrelated to remotes. It's more that we have to pass in `--git-dir` to `git` and to avoid UNC problems we want that path to be relative to the workspace dir (which we would pass as `Command::current_dir`). This is the hard bit to figure out from the lib alone.

If we don't have relative configured remote paths, we can simply run `git` at the `<git-dir>`, right?
My point is that, the `workspace_root` path is needed by all implementations in order to make them behave the same. (`git2` doesn't need it just because we have no way to fix the problem.)

> > Hmm, I have feeling that we should hide git2 use in library code so we can phase it out easily by replacing the implementation by gix or subprocessing. FWIW, I think the main reason we do instantiate git2 repo by caller is that, unlike git_repo() -> gix::Repository, open_git_repo() -> git2::Repository does non-trivial work, so the instance should be reused (by caller) if possible.
> 
> From looking at the code, `cli` is using a `git2::Repository` for doing a bunch of work. My idea with separating it at the caller is to make it explicit that, from that point on, the `git2` repository is not being used again. It would also future proof the Subprocessing path against using `git2` in the future (which I think is desirable). Then we can work on migrating the `cli` code to `gitoxide` independently.

If @emilazy's patch gets merged, I expect most of `git2` uses are removed from CLI (except for tests.) Remaining parts would be just fetch and pull.

bsdinis:
Okay, I see your point. If that's the case, I think I might abandon most of the chain of work I've been doing, will rebase the relevant changes and get them out

bsdinis:
Closing

PR #5445, opened 2025-01-23
cli: relocate git fetch code to cli/src/commands/git/fetch.rs

bsdinis:
This code was living in `cli/src/git_utils.rs` but no one else was using it


martinvonz (commented):
@essiene, do you still plan to use this function from `jj git sync`?

bsdinis:
Closing here because this commit is now part of #5498 .
FWIW there is a comment in @essiene 's original PR about making this function move.

Issue #4756, opened 2024-11-02
FR: `jj diff --remote`


Grillo-0:
**Is your feature request related to a problem? Please describe.**
Sometimes I change my local branch and I want to see what changed on the remote before pushing. Currently, I've being doing `jj diff --from <bookmark-name>@<remote> --to <bookmark-name>`. I would find it much easier if I don't need to remember the bookmark name and type it twice.

**Describe the solution you'd like**
I think that a new flag called `--remote` could solve this. If the head is on a tracked bookmark, it would diff the remote with the local.

**Describe alternatives you've considered**
Maybe this could be solved with new revsets functions? Maybe a `bookmark_name()` to return the bookmark name of a revision, so you could use with `remote_branches()`.

yuja:
I feel `--remote` is too specific. It only works if the revision to be diffed has one tracking branch. Maybe better to add some revset function/parameter that looks up tracking remote by local revision.

david-crespo:
I ran into this yesterday and came up with a goofy workaround. I'm sure there's a way to clean it up, and I'm sure there are lots of edge cases around, e.g., multiple bookmarks between `main` and `@`, but it is working for me in simple cases.

```sh
function curr_bookmark { jj bookmark list --tracked -r 'trunk()..@' -T 'name++"\n"' | head -1 }
alias jdr='jj diff -f "$(curr_bookmark)@origin"'
```

arxanas:
> This is even more relevant for me with jj, given the auto-snapshotting behaviour. Is there absolutely no way (e.g. alias) to do this currently?

Since jj doesn't have a "currently-checked-out branch" (yet?), there's fundamentally no way to ask "what is the current branch" so that you can compare it to its remote counterpart. I think @david-crespo's solution above is the best you can do for now, where you have to heuristically determine what branch you're on.

0xdeafbeef:
> > This is even more relevant for me with jj, given the auto-snapshotting behaviour. Is there absolutely no way (e.g. alias) to do this currently?
> 
> Since jj doesn't have a "currently-checked-out branch" (yet?), there's fundamentally no way to ask "what is the current branch" so that you can compare it to its remote counterpart. I think @david-crespo's solution above is the best you can do for now, where you have to heuristically determine what branch you're on.

What you think about` jj diff - -branch name  - - remote`
You will need to type less and you don't need any heuristics. Just diff branch with `branch@remote`

david-crespo:
[...]

Issue #5307, opened 2025-01-09
FR: Add a language for selecting a tree


martinvonz:
**Is your feature request related to a problem? Please describe.**

It's always bothered me that `jj diff -r X` cannot be written as `jj diff --to X --from <something>` because `X-` might resolve to multiple revisions. There are other such examples like in `jj restore`.

A related to case that came up on Discord was that `jj restore` could be used for resolving a conflict in favor of a particular side of the conflict if we had a way of selecting a given tree from a conflict.

**Describe the solution you'd like**

Commands that currently take a revision argument only to resolve it to a tree (e.g. `jj diff --from/--to`, `jj restore --from`, `jj file show -r`) could take a new kind of expression instead of a revset. The new expression would allow regular revset but will resolve to a tree instead. By default, the revset is expected to resolve to a single revision and we use that revision's tree (merged if a conflict). But the user can also do something like `jj diff --from 'merge_of(X-)' --to Y` to merge the commits in`X-`, or `jj restore --from 'conflict_side(X, 1)'`. 

**Describe alternatives you've considered**

We've talked about making revsets themselves able to calculate merges of commits. I don't like that idea because it complicates revsets, and it means that the merged result would presumably be assigned to a commit (with some placeholder description etc) to fit into the revset language.

arxanas:
> A related to case that came up on Discord was that `jj restore` could be used for resolving a conflict in favor of a particular side of the conflict if we had a way of selecting a given tree from a conflict.

Link to the Discord question thread: ["Resolve conflict by picking a side for all"](https://discord.com/channels/968932220549103686/1319328750403321976).

PR #5450, opened 2025-01-23
revset: add `remote` function

Grillo-0:
This can be used to find the remote bookmark tracked by a revset.

It can be useful for diffing with the remote:
`jj diff --from remote(@) --to @`

Fixes: #4756

# Checklist

If applicable:


lib/src/revset.rs
```diff
@@ -365,6 +370,17 @@ impl<St: ExpressionState<CommitRef = RevsetCommitRef>> RevsetExpression<St> {
     pub fn git_head() -> Rc<Self> {
         Rc::new(Self::CommitRef(RevsetCommitRef::GitHead))
     }
+
+    pub fn remote(commit_ref: Rc<Self>, remote_pattern: StringPattern) -> Option<Rc<Self>> {
+        if let Self::CommitRef(commit_ref) = commit_ref.borrow() {
+            Some(Rc::new(Self::CommitRef(RevsetCommitRef::Remote {
+                commit_ref: Box::new(commit_ref.clone()),
+                remote_pattern,
+            })))
```
yuja, line 379: Ah, yes, it's correct that we can't support full range of revset expressions at the name resolution stage, but it's a big limitation I didn't notice before. We'll probably want to do `remote(@-)` or something, but `x-` can't be evaluated here. I have no idea how we can solve this problem.

We might need a language selecting bookmarks. Related: https://github.com/jj-vcs/jj/issues/5307

martinvonz: Agreed. It would still be useful in that it allows you to create an alias that you can pass a specific bookmark name to, right?

Grillo-0: Hey @yuja! I know this is a bit late XD, but do you remember why you added a 'Related' to the tree language issue? I don't see the connection between bookmarks and trees.

yuja: I think I meant they would be similar in that evaluated revisions are transformed by user expression.

* `some_tree_func(revset_expr)`: `revisions -> trees`
* `some_bookmark_func(revset_expr)`: `revisions -> remote/local bookmarks`?

PR #5451, opened 2025-01-24
codespell action: use `uv` to run codespell

ilyagr:
The primary goal is to control the version of the `codespell` Python
package that we run via `uv.lock`. See also
https://github.com/jj-vcs/jj/pull/5425.

Also move the action next to the `cargo fmt` action. It might make sense
to split both of them out of `build.yml` if it gets too long, but I
think they should be next to each other since they are so similar in
spirit.

--------

I kept the name, so the action is still considered "required". I also checked that the action can still fail, see https://github.com/jj-vcs/jj/actions/runs/12942072678/job/36099176842?pr=5451.


pyproject.toml
```diff
@@ -24,4 +24,16 @@ dev-dependencies = [
...
 ]
+
+[tool.codespell]
+check-filenames=true
+check-hidden=true
+# Codespell does not respect .gitignore. It may be necessary to add to
+# this list by running e.g. `uv run codespell --skip=./rendered-docs`
+# if you have less common ignored files or globally ignored files present.
+# Alternatively, try `uv run codespell $(jj file list)`.
+skip="./target,./.jj,*.lock,./.git,./.venv"
```
bryceberger, line 37: Should also include `./.direnv`

thoughtpolice (approved):
LGTM, along with Bryce's nit

ilyagr:
Thank you, Austin and Bryce!

Issue #63, opened 2022-02-20
git: SSH config not respected


uhthomas:
## Expected Behaviour

`jj git clone` should respect the SSH config (`~/.ssh/config`).

## Actual Behaviour

The SSH config is not respected, which may be problematic in some cases. In this case, `ProxyCommand` is to communicate with a private Git server.

## Steps to Reproduce the Problem

1. Use an SSH environment which requires additional configuration via `~/.ssh/config` to work.
2. Observe `jj git clone` hang.

## Specifications

- Version: `0.2.0.r917.2916cb2-1`
- Platform: Linux x86_64

martinvonz:
Thanks for your report. I'll be away from my computer for about a week from now. If anyone feels like looking into this while I'm away, I think the right place to start is [here](https://github.com/martinvonz/jj/blob/2916cb2d9f654313727b539b6991bc09d5af560b/lib/src/git.rs#L433-L452).

unrelentingtech:
libgit2 uses libssh2, not OpenSSH, so *of course* it doesn't use OpenSSH's configuration. And unfortunately there is no support for proxying at all. See https://github.com/libgit2/libgit2/issues/4872 (one linked PR does something in that direction)

> the right place to start is [here](https://github.com/martinvonz/jj/blob/2916cb2d9f654313727b539b6991bc09d5af560b/lib/src/git.rs#L433-L452)

Hopefully the hardcoded `id_rsa` won't remain the only supported key path for non-agent operation there for long… :)

bergkvist:
Is it possible to use OpenSSH instead of libssh2?

martinvonz:
> Is it possible to use OpenSSH instead of libssh2?

I was going to say that I think we'd have to ask for such support from libgit2 (or switch to gitoxide or something) but I see [you've already asked there](libgit2/libgit2/issues/4872) (thanks). I have no idea how feasible it is.

bergkvist:
@arxanas Do you have to rerun `jj init --git-repo .` every time you do a git pull? Or does it "keep in sync" after you do it once?

yuja:
`jj init --git-repo .` is needed only once.

`jj git fetch` is basically the same as `git fetch --prune && jj git import`, and for "colocated" repo created by `--git-repo .`, `jj git import` runs implicitly every time you do `jj <subcommand>`.

martinvonz:
[...]

PR #5228, opened 2025-01-02
git: spawn a separate git process for network operations

bsdinis:
# Reasoning:

`jj` fails to push/fetch over ssh depending on the system.
Issue #4979 lists over 20 related issues on this and proposes spawning
a `git` subprocess for tasks related to the network (in fact, just push/fetch
are enough).

This PR implements this.
Users can either enable shelling out to git in a config file:

```toml
[git]
subprocess = true
```

# Implementation Details:

This PR implements shelling out to `git` via `std::process::Command`.
There are 2 sharp edges with the patch:
 - it relies on having to parse out git errors to match the error codes
   (and parsing git2's errors in one particular instance to match the
   error behaviour). This seems mostly unavoidable

 - to ensure matching behaviour with git2, the tests are maintained across the
   two implementations. This is done using test_case, as with the rest
   of the codebase

# Testing:

Run the rust tests:
```
$ cargo test
```

# Build:
```
$ cargo build
```

# Clone a private repo:
```
$ path/to/jj git clone --config='git.subprocess=true' <REPO_SSH_URL>
```

# Create new commit and push
```
$ echo "TEST" > this_is_a_test_file.txt
$ path/to/jj describe -m 'test commit'
$ path/to/jj git push --config='git.subprocess=true' -b <branch>
```

# Issues Closed

With a grain of salt, but most of these problems should be fixed (or at least checked if they are fixed). They are the ones listed in #4979 .

SSH:
- https://github.com/martinvonz/jj/issues/63
- https://github.com/martinvonz/jj/issues/440
- https://github.com/martinvonz/jj/issues/1455
- https://github.com/martinvonz/jj/issues/1507
- https://github.com/martinvonz/jj/issues/2931
- https://github.com/martinvonz/jj/issues/2958
- https://github.com/martinvonz/jj/issues/3322
- https://github.com/martinvonz/jj/issues/4101
- https://github.com/martinvonz/jj/issues/4333
- https://github.com/martinvonz/jj/issues/4386
- https://github.com/martinvonz/jj/issues/4488
- https://github.com/martinvonz/jj/issues/4591
- https://github.com/martinvonz/jj/issues/4802
- https://github.com/martinvonz/jj/issues/4870
- https://github.com/martinvonz/jj/issues/4937
- https://github.com/martinvonz/jj/issues/4978
- https://github.com/jj-vcs/jj/issues/5120
- https://github.com/jj-vcs/jj/issues/5166

Clone/fetch/push/pull:
[...]

PR #5458, opened 2025-01-24
Add `git.ssh-key-file` setting

dconeybe:
This PR addresses a specific problem resulting from `~/.ssh/config` being ignored by jj (https://github.com/jj-vcs/jj/issues/63). Namely, without this PR, jj hardcodes 3 possible ssh key files that it considers: `~/.ssh/id_ed25519_sk`, `~/.ssh/id_ed25519`, and `~/.ssh/id_rsa`: https://github.com/jj-vcs/jj/blob/2a57a6bdd257262059dc62d021e8324a929001c6/cli/src/git_util.rs#L188

Personally, I also have a file called `~/.ssh/id_ed25519_personal` that I use for personal projects, but there is no way to get jj to use it. As a result, my `jj git push` commands always fail with this error:

```
$ jj git push
Changes to push to origin:
  Add bookmark tmp to d2f817604110
Error: failed to authenticate SSH session: Unable to extract public key from private key file: Wrong passphrase or invalid/unrecognized private key file format; class=Ssh (23)
Hint: Jujutsu uses libssh2, which doesn't respect ~/.ssh/config. Does `ssh -F /dev/null` to the host work?
```

With the patch in this PR, I can run this command to tell jj to use the correct ssh key when contacting the git server:

```
jj config set --repo git.ssh-key-file $HOME/.ssh/id_ed25519_personal
```

It may also make sense to add this configuration setting _per remote_, or possibly a different one for _pushing_ and _fetcing_.

Disclaimer: I am a novice rust programmer and this is my first contribution to jj; if this idea is accepted I will also add proper unit testing and changelog entries.

### DIFF
--- a/lib/src/settings.rs
+++ b/lib/src/settings.rs
@@ -61,6 +61,7 @@ pub struct GitSettings {
     pub abandon_unreachable_commits: bool,
     pub subprocess: bool,
     pub executable_path: PathBuf,
+    pub ssh_key_file: Option<PathBuf>,
 }
 
 impl GitSettings {
@@ -70,6 +71,17 @@ impl GitSettings {
             abandon_unreachable_commits: settings.get_bool("git.abandon-unreachable-commits")?,
             subprocess: settings.get_bool("git.subprocess")?,
             executable_path: settings.get("git.executable-path")?,
+            ssh_key_file: match settings.get_string("git.ssh-key-file") {
+                Ok(ssh_key_file) => {
+                    let trimmed_ssh_key_file = ssh_key_file.trim();
+                    if trimmed_ssh_key_file.is_empty() {
+                        None
+                    } else {
+                        Some(crate::file_util::expand_home_path(trimmed_ssh_key_file))
+                    }
+                }
+                Err(_) => None,
+            },
         })
     }
 }
@@ -81,6 +93,7 @@ impl Default for GitSettings {
             abandon_unreachable_commits: true,
             subprocess: false,
             executable_path: PathBuf::from("git"),
+            ssh_key_file: None,
         }
     }
 }
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -390,6 +390,11 @@
                     "type": "string",
                     "description": "Path to the git executable",
                     "default": "git"
+                },
+                "ssh-key-file": {
+                    "type": "string",
+                    "description": "Path to SSH key file to use; by default, uses the following files in ~/.ssh: id_ed25519_sk, id_ed25519, id_rsa",
+                    "default": ""
                 }
             }
         },
--- a/cli/src/git_util.rs
+++ b/cli/src/git_util.rs
@@ -311,8 +311,13 @@ pub fn with_remote_git_callbacks<T>(
             .map(|x| x as &mut dyn FnMut(&git::Progress));
         callbacks.sideband_progress =
             sideband_progress_callback.map(|x| x as &mut dyn FnMut(&[u8]));
-        let mut get_ssh_keys = get_ssh_keys; // Coerce to unit fn type
-        callbacks.get_ssh_keys = Some(&mut get_ssh_keys);
+        let get_ssh_keys: &mut dyn FnMut(&_) -> _ = &mut |username| {
+            git_settings
+                .ssh_key_file
+                .as_ref()
+                .map_or_else(|| get_ssh_keys(username), |key_file| vec![key_file.clone()])
+        };
+        callbacks.get_ssh_keys = Some(get_ssh_keys);
         let mut get_pw =
             |url: &str, _username: &str| pinentry_get_pw(url).or_else(|| terminal_get_pw(ui, url));
         callbacks.get_password = Some(&mut get_pw);


dconeybe:
I will be OOO for a week, so I will respond to any feedback when I get back on Feb 03, 2025.

bryceberger:
The build failures in CI are due to https://github.com/rust-lang/rust/pull/121346/. CI is on 1.76, presumably locally you are running 1.79+.

lib/src/settings.rs
```diff
@@ -70,6 +71,17 @@ impl GitSettings {
             abandon_unreachable_commits: settings.get_bool("git.abandon-unreachable-commits")?,
             subprocess: settings.get_bool("git.subprocess")?,
             executable_path: settings.get("git.executable-path")?,
+            ssh_key_file: match settings.get_string("git.ssh-key-file") {
+                Ok(ssh_key_file) => {
+                    let trimmed_ssh_key_file = ssh_key_file.trim();
+                    if trimmed_ssh_key_file.is_empty() {
+                        None
+                    } else {
+                        Some(PathBuf::from(trimmed_ssh_key_file))
```
bryceberger, line 80: This should probably use `crate::file_util::expand_home_path()`

dconeybe: Will change.

dconeybe:
> The build failures in CI are due to [rust-lang/rust#121346](https://github.com/rust-lang/rust/pull/121346). CI is on 1.76, presumably locally you are running 1.79+.

Yes I downgraded to 1.76 and I can repro the failures locally. I'm seeing what I can do for a fix.

bryceberger:
Think fixing it on 1.76 would require moving the condition into the closure, ex:
```diff
diff --git a/cli/src/git_util.rs b/cli/src/git_util.rs
index f3a84c5099..192d00b017 100644
--- a/cli/src/git_util.rs
+++ b/cli/src/git_util.rs
@@ -311,11 +311,12 @@
             .map(|x| x as &mut dyn FnMut(&git::Progress));
         callbacks.sideband_progress =
             sideband_progress_callback.map(|x| x as &mut dyn FnMut(&[u8]));
-        let mut get_ssh_keys: &mut dyn FnMut(&str) -> Vec<PathBuf> =
-            match &git_settings.ssh_key_file {
-                None => &mut get_ssh_keys, // Coerce to unit fn type
-                Some(ssh_key_file) => &mut |_: &str| -> Vec<PathBuf> { vec![ssh_key_file.clone()] },
-            };
+        let mut get_ssh_keys: &mut dyn FnMut(&_) -> _ = &mut |username| {
+            git_settings
+                .ssh_key_file
+                .as_ref()
+                .map_or_else(|| get_ssh_keys(username), |key_file| vec![key_file.clone()])
+        };
         callbacks.get_ssh_keys = Some(&mut get_ssh_keys);
         let mut get_pw =
             |url: &str, _username: &str| pinentry_get_pw(url).or_else(|| terminal_get_pw(ui, url));
```

That said, we might be bumping the MSRV soon anyway ([discord discussion](https://discord.com/channels/968932220549103686/969291218347524238/1332484516983603392)).

martinvonz:
I'm not against this PR, but you probably don't need it if you build after #5228 and set `git.subprocess = true`

lib/src/settings.rs
dconeybe: Done.

dconeybe:
> Think fixing it on 1.76 would require moving the condition into the closure, ex:

I've applied your patch. The code looks pretty good either way IMO, so may as well get it to build :)

dconeybe:
> I'm not against this PR, but you probably don't need it if you build after #5228 and set `git.subprocess = true`

I'll keep my eye out for that!

lib/src/settings.rs
```diff
@@ -70,6 +71,17 @@ impl GitSettings {
             abandon_unreachable_commits: settings.get_bool("git.abandon-unreachable-commits")?,
             subprocess: settings.get_bool("git.subprocess")?,
             executable_path: settings.get("git.executable-path")?,
+            ssh_key_file: match settings.get_string("git.ssh-key-file") {
```
yuja, line 74: If we add config knob, I would expect it would be a list of paths which defaults to `["~/.ssh/id_..", ..]`.

And it's probably better to move config handling to CLI. `jj_lib::expand_home_path()` is half-baked. Ideally, library code shouldn't depend on environment.

dconeybe: Duly noted. If nobody gets to it I'll take a look on the week of Feb 03 when I return from vacation.

dconeybe: Update: I've looked into modifying the new `ssh_key_file` configuration option to support a _list_ of values (rather than a single string value); however, I got lost in the complexities of rust (I am a total rust noob). It looks like a list-value configuration option would be breaking some new ground.

What if we keep this new config option as a single string value for now. In the future (or in a follow-up PR) it can be expanded to support a list of string values. This updated implementation could treat a string value in the same manner that it would treat a list of one string element, to maintain backwards compatibility.

martinvonz: #5228 has been merged and will be part of the release that goes out today. I suspect we'll make subprocessing the default from the release after.

dconeybe: Ok I'll close this PR and revisit it if `git.subprocess` doesn't address my specific issue. Thank you for your engagement.

bryceberger: > It looks like a list-value configuration option would be breaking some new ground.

Because `StackedConfig::get` returns any `T` such that `T: Deserialize`, you can just `config.get::<Vec<String>>("some.key")` to get a `Result<Vec<String>, ConfigGetError>`.

PR #5460, opened 2025-01-25
github: migrate to GHA aarch64 runners

thoughtpolice:
We no longer need to do cross compilation, and these instances seem pretty fast. Enable them everywhere across the board.

# Checklist

If applicable:

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -22,10 +22,10 @@ jobs:
           os: ubuntu-24.04
           target: x86_64-unknown-linux-gnu
         - build: linux-aarch64-musl
-          os: ubuntu-24.04
+          os: ubuntu-24.04-arm
           target: aarch64-unknown-linux-musl
         - build: linux-aarch64-gnu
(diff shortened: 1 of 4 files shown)


emilazy:
Perhaps should move all the Macs to AArch64 too? Should be just as easy to build binaries for x86, and the AArch64 runners have Rosetta 2 and thus can handle the tests too; and of course the speed improvement should be even larger.

thoughtpolice:
How do we cross compile on macOS? We don't even need `cross`, right? The main reason I haven't done it is just bc I haven't swapped over to my Mac in a while to debug issues.

emilazy:
It can just be used as a native x86 Darwin machine – could even install a native x86 Rust toolchain. I’d assume `cargo --target` and so on also just works, though (and would be faster).

thoughtpolice:
Alright, I'll try that, but I'll do a follow up for it in a separate PR since I might have to hit it with a wrench a few times. I think this is good to go, though.

bsdinis (approved):

PR #5464, opened 2025-01-25
derive `clap::ValueEnum` for `ColorChoice`

jakobhellermann:
# Checklist

If applicable:


PhilipMetzger:
Please still adhere to the commit style outlined in the docs. Also I assume that this is prep for more dynamic completion work?

jakobhellermann:
> Please still adhere to the commit style outlined in the docs. 

Oops, forgot about that.

> Also I assume that this is prep for more dynamic completion work?

No, this is actually enough on its own for clap to suggest `always never debug auto` for `jj --color <TAB>`. Looking at the expanded macro it does some kind of autoref specialization trick:
```rust
{
    use clap_builder::builder::impl_prelude::*;
    let auto = clap_builder::builder::_infer_ValueParser_for::<ColorChoice>::new();
    (&&&&&&auto).value_parser()
}
```

cli/tests/cli-reference@.md.snap
```diff
@@ -187,8 +187,9 @@ To get started, see the tutorial at https://jj-vcs.github.io/jj/latest/tutorial/
 
    It is possible to run mutating commands when loading the repo at an earlier operation. Doing that is equivalent to having run concurrent commands starting at the earlier operation. There's rarely a reason to do that, but it is possible.
 * `--debug` — Enable debug logging
-* `--color <WHEN>` — When to colorize output (always, never, debug, auto)
-* `--quiet` — Silence non-primary command output
+* `--color <WHEN>` — When to colorize output (always, never, debug, auto)uto`
+
+quiet` — Silence non-primary command output
```
PhilipMetzger, line 192: what happened here?

cli/src/cli_util.rs
```diff
@@ -3099,7 +3099,7 @@ pub struct GlobalArgs {
 
 #[derive(clap::Args, Clone, Debug)]
 pub struct EarlyArgs {
-    /// When to colorize output (always, never, debug, auto)
+    /// When to colorize output
```
PhilipMetzger, line 3102: I still think displaying the options has some value.

PhilipMetzger (approved):

cli/tests/cli-reference@.md.snap
jakobhellermann: No idea honestly, I thought I just did `cargo nextest` and `cargo insta review`. Should be fixed now.

cli/src/cli_util.rs
jakobhellermann: Without deleting this, `jj -h` would result in
```
      --color <WHEN>                 When to colorize output (always, never, debug, auto) [possible values: always, never, debug, auto]
```

PhilipMetzger: ok, then it is fine to remove.

yuja:
FWIW, it implements `FromStr` and `Display` instead of `clap::ValueParser` because I thought it would be a bit weird that `ui` depends on `clap`. I don't have strong feeling though.

jakobhellermann:
> FWIW, it implements `FromStr` and `Display` instead of `clap::ValueParser` because I thought it would be a bit weird that `ui` depends on `clap`. I don't have strong feeling though.

Hm. The same end result could be accomplished by removing the `clap::ValueParser` derive and instead supplying
```rust
        value_parser = vec![PossibleValue::new(ColorChoice::Always.to_string(), ColorChoice::Never.to_string(), ColorChoice::Debug.to_string(), ColorChoice::Auto.to_string()]
```
in `EarlyArgs`.

I don't have a strong opinion but my gut reaction says the boilerplate (and possibility to miss changes) is not worth the extra separation of `clap` from `ui`.

martinvonz:
Maybe it can be implemented by leveraging the `FromStr` implementation? Something similar to what we do here: 

https://github.com/jj-vcs/jj/blob/e8620c31ae493830dfd903f6d609611c7abb7314/cli/src/cli_util.rs#L3219-L3225

jakobhellermann:
I could do that, but what's the difference between implementing `ValueParserFactory` or `clap::ValueEnum`? If the motivation against the latter is that `ui.rs` shouldn't concern itself with cli logic, I don't see the advantage of the former.

martinvonz:
Would the `ValueParserFactory` impl have to be in `ui.rs`? Maybe we could have a `ColorChoiceArg` that's not in `ui.rs`.

jakobhellermann:
> Would the `ValueParserFactory` impl have to be in `ui.rs`? Maybe we could have a `ColorChoiceArg` that's not in `ui.rs`.

The impl could live in `cli_util.rs`. You wouldn't be able to move `ui.rs` into another crate without changes, but `ui.rs` would contain no mention of clap.
At that point you could also do `impl ValueEnum for ColorChoiceArg` in `cli_util.rs`.

So the options are
1. `#[derive(clap::ValueEnum)]` in `ui.rs`
2. `impl ValueParserFactory for ColorChoiceArg` or `impl ValueEnum for ColorChoiceArg` in `cli_util.rs`
3. list out `PossibleValue`s in `EarlyArgs` explicitly.

Im fine with either option, with a slight tendency towards 1. for the simplicity.

yuja:
> So the options are
>
> 1. `#[derive(clap::ValueEnum)]` in `ui.rs`
> 2. `impl ValueParserFactory for ColorChoiceArg` or `impl ValueEnum for ColorChoiceArg` in `cli_util.rs`
> 3. list out `PossibleValue`s in `EarlyArgs` explicitly.

I wouldn't do 2 because it would be complicated. Does 3 really work? I think we need to somehow specify the parse function. To do 3, we might have to use `ArgValueCandidates` instead. The list of constants (of type `[&str]` or `[ColorChoice]`) can be defined in ui.

As I said, I don't have strong feeling against 1. I'm not sure if future `jj-cli-lib` will have `Ui` in it either. If we do 1, maybe we can remove some of trait impls.

martinvonz:
I don't feel strongly either way. Option 1 sounds good because it's simplest. I'm also unclear about how we'll restructure this code in the future.

jakobhellermann:
> If we do 1, maybe we can remove some of trait impls.

That's right, I could remove the `FromStr` now.

> Option 1 sounds good because it's simplest. I'm also unclear about how we'll restructure this code in the future.

Okay, sounds like I can go ahead merging this as is, and if/when ui/jj-cli is restructered in the future we can reevaluate.

PR #5468, opened 2025-01-26
tests: keep user's `$PATH` while running `jj` commands

scott2000:
Currently, the Git subprocess tests only work on Linux due to a default path being used for the `git` executable when `$PATH` is unset. This can break if Git isn't installed at the expected path. Also, I believe it is currently necessary to set the `$TEST_GIT_EXECUTABLE_PATH` environment variable on Windows for tests to pass. Instead, we should use the user's `$PATH` to locate the `git` executable, as well as any other executables that are needed. This also makes `$TEST_GIT_EXECUTABLE_PATH` no longer necessary, so it can be removed.

Discussed in #5463.

# Checklist

If applicable:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,20 +86,10 @@ jobs:
         tool: nextest
     - name: Build
       run: cargo build --workspace --all-targets --verbose ${{ matrix.cargo_flags }}
-    - name: Test [non-windows]
-      if: ${{ matrix.os != 'windows-latest' }}
+    - name: Test
       run: |
         cargo nextest run --workspace --all-targets --verbose ${{ matrix.cargo_flags }}
       env:
-        RUST_BACKTRACE: 1
-        CARGO_TERM_COLOR: always
-    - name: Test [windows]
-      if: ${{ matrix.os == 'windows-latest' }}
-      run: cargo nextest run --workspace --all-targets --verbose ${{ matrix.cargo_flags }}
-      env:
-        # tests clear the PATH variable (but is only felt on windows),
-        # so we need to set the git executable
-        TEST_GIT_EXECUTABLE_PATH: 'C:\Program Files\Git\bin\git.exe'
         RUST_BACKTRACE: 1
         CARGO_TERM_COLOR: always
 
--- a/cli/tests/common/mod.rs
+++ b/cli/tests/common/mod.rs
@@ -18,7 +18,6 @@ use std::path::Path;
 use std::path::PathBuf;
 
 use indoc::formatdoc;
-use itertools::Itertools as _;
 use regex::Captures;
 use regex::Regex;
 use tempfile::TempDir;
@@ -96,6 +95,9 @@ impl TestEnvironment {
             cmd.env(key, value);
         }
         cmd.env("RUST_BACKTRACE", "1");
+        // We want to keep the "PATH" environment variable to allow accessing
+        // executables like `git` from the PATH.
+        cmd.env("PATH", std::env::var_os("PATH").unwrap_or_default());
         cmd.env("HOME", self.home_dir.to_str().unwrap());
         cmd.env("JJ_CONFIG", self.config_path.to_str().unwrap());
         cmd.env("JJ_USER", "Test User");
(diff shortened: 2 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -97,9 +97,6 @@ jobs:
       if: ${{ matrix.os == 'windows-latest' }}
```
martinvonz, line 97: Looks like the Windows and non-Windows cases can be merged now, right?

cli/tests/common/mod.rs
```diff
@@ -92,6 +91,23 @@ impl TestEnvironment {
...
+        if cfg!(windows) {
+            // On Windows, "Path" may be set instead of (or in addition to) "PATH"
+            if let Ok(path_var) = std::env::var("Path") {
+                path_components.extend(std::env::split_paths(&path_var));
+            }
+        }
+        cmd.env(
+            "PATH",
+            std::env::join_paths(path_components).expect("invalid PATH variable"),
+        );
```
martinvonz, line 110: Will it work to just copy over both `PATH` and - on Windows -`Path`? Something like this in `TestEnvironment::default()`:

```rust
let mut env_vars = HashMap::new();
env_vars.insert("PATH".to_string(), std::env::var("PATH").unwrap_or_default());
if cfg!(windows) {
    env_vars.insert("Path".to_string(), std::env::var("Path").unwrap_or_default());
}
```

scott2000: I wasn't sure about this because we do have a couple tests that explicitly clear `PATH` to test a missing `git` executable using this command:

```rust
test_env.add_env_var("PATH", "");
```

So we'd probably have to add another command there to clear `Path` for Windows, and so I thought it was simpler to just always use `PATH`. I'm open to changing it though, what do you think is cleaner?

martinvonz: I think we can update those tests a little. We always define a value for `git.executable-path` (it has a default value in `misc.toml`), so we don't need a special test for when it isn't set at all.

The point of the first test is to check what happens when `git.executable-path` is the name of a binary that doesn't exist on `$PATH`. It doesn't need to be called exactly `git`, so we can set `git.executable-path = "missing-program"` instead of setting `PATH=`.

The second test provides an absolute path, so it doesn't need set set `$PATH` at all.

yuja: iirc, environment variables on Windows are case insensitive. I don't know if `Command` builder handles env vars in platform-dependent manner, but I assume `std::env` would do (because these variables are read from OS.) So I assume either `PATH` or `Path` should exist on Windows, not both.

martinvonz: So something like this?

```rust
let mut env_vars = HashMap::new();
let path = if cfg!(unix) {
    std::env::var("PATH").unwrap_or_default()
} else {
  std::env::var("PATH").or(std::env::var("Path")).unwrap_or_default()
};
env_vars.insert("PATH".to_string(), path);
```

scott2000: It looks like on the [git-branchless issue](https://github.com/arxanas/git-branchless/issues/1256#issue-2161396773), the problem was actually that the code collected the environment variables into a map and used `map.get("PATH")`, which is case sensitive (since the map used regular case-sensitive lookup). Apparently `std::env::var("PATH")` is case insensitive (as the user tested in that issue), so it should automatically work for both possible cases.

bsdinis:
This looks great! Thank you for doing this, getting a hang of it was very annoying to begin with and this is much cleaner!

I think we should definitely do the `LC_ALL` bit, but definitely in another PR

cli/tests/test_git_clone.rs
```diff
@@ -1029,14 +1029,13 @@ fn test_git_clone_no_git_executable() {
 
 #[test]
 fn test_git_clone_no_git_executable_with_path() {
-    let mut test_env = TestEnvironment::default();
+    let test_env = TestEnvironment::default();
```
martinvonz, line 1032: Did it not work to set `git.executable-path = "missing-program"` in the test case above? I think that's slightly better because it avoids resetting `$PATH`, which seems to be bad to do on MinGW.

scott2000: It worked well, but I just removed the change when we noticed that there was no issue with using `PATH` always since I thought it was unnecessary. That's a good point though, I'll add it back.

bsdinis: The test above is specifically testing for a failure on PATH resolution, so it looks good to me

martinvonz: I think it's probably fine because the test case fails very quickly, so maybe we haven't gotten to the point where we need to run any MinGW executables. I don't know which executables the comment is about. The code was added in #1566. @71, can you tell us which executables we're talking about? Is it things like the fake diff editors? If so, it should be fine because we don't use those in this test.

I still prefer setting `git.executable-path = "missing-program"` because it's simpler to think about.

scott2000: I do think it's a bit nicer to use `git.executable-path`, since it's a bit strange to be clearing the `PATH` only for some tests and not others. I updated the PR, but I used `jj-test-missing-program` as the name since it's even more unlikely that someone would have that as the name of a real program on their machine.

martinvonz (approved):
Thanks!

Sorry, I should have given approval yesterday. I didn't mean to blackmail you into changing the `$PATH` vs `git.executable-path` to my preference.

PR #5470, opened 2025-01-26
docs: fix typo in env variable `$COMPLETE`

### LINKED ISSUE #5466: docs: two environment variables mentioned for shell completions
Reading the section on "Command-line completion", there is a note that seems to contradict the instructions for the dynamic completions. It also has a link to an explanation which doesn't explain anything.
> To get the improved completions it is required to set the `COMPLETION` env variable to your current shell before using the completion command, as they're not the default yet.

The instructions use a different variable. For instance, for bash:
> `source <(COMPLETE=bash jj)`

and for fish:
> `COMPLETE=fish jj | source`

martinvonz:
Closes #5466.

# Checklist

If applicable:

### DIFF
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -220,10 +220,9 @@ transition once we default to these new completions.
 
 !!! info "Why are the improved completions not the default?"
 
-    To get the improved completions it is required to set the `COMPLETION` env
-    variable to your current shell before using the completion command, as
-    they're not the default yet. See the upstream clap issue [#3166][clap]
-    for the explanation.
+    To generate the dynamic completion script, execute `jj` with `$COMPLETE` set
+    to the name of your shell (see examples below). See the upstream clap issue
+    [#3166][clap] for the explanation.
 
 
 ### Bash


docs/install-and-setup.md
```diff
@@ -220,7 +220,7 @@ transition once we default to these new completions.
 
 !!! info "Why are the improved completions not the default?"
 
-    To get the improved completions it is required to set the `COMPLETION` env
+    To get the improved completions it is required to set the `COMPLETE` env
```
yuja, line 223: nit: (unrelated to this PR), I think this should say "To generate improved completion script, ... execute `jj` with `COMPLETE` env variable" or something. With the original wording, I would add `COMPLETE=` to my `~/.zshrc`.

yuja (approved):

docs/install-and-setup.md
martinvonz: Good point. I reworded it a bit more. Hopefully it's clearer.

yuja (approved):

Issue #5359, opened 2025-01-14
Error using a repository with a unicode primary bookmark name


Nemo157:
## Description

Attempting to use a repository that uses a unicode primary bookmark name fails. Commenting out the `trunk()` alias in the local config avoids the issue, and afaict all commands work correctly with the bookmark name (except where they need to know the head is immutable or primary).

## Steps to Reproduce the Problem

```console
> jj git clone https://github.com/Nemo157/ennead
Fetching into new repo in "/tmp/scratch.2025-01-14T14-08.do3GxX/ennead"
bookmark: κεφάλαιος@origin [new] untracked
Setting the revset alias "trunk()" to "κεφάλαιος@origin"
Working copy now at: rowmrups e527c90d (empty) (no description set)
Parent commit      : zsqrywow cf4d6273 κεφάλαιος | fix dithering functions for non-target size images
Added 23 files, modified 0 files, removed 0 files
> cd ennead
> jj log
Config error: Invalid `revset-aliases.immutable_heads()`
Caused by:
1:  --> 1:1
  |
1 | builtin_immutable_heads()
  | ^-----------------------^
  |
  = In alias "builtin_immutable_heads()"
2:  --> 1:9
  |
1 | present(trunk()) | tags() | untracked_remote_bookmarks()
  |         ^-----^
  |
  = In alias "trunk()"
3:  --> 1:1
  |
1 | κεφάλαιος@origin
  | ^---
  |
  = expected <identifier> or <expression>
For help, see https://martinvonz.github.io/jj/latest/config/.
> cat .jj/repo/config.toml
[revset-aliases]
"trunk()" = "κεφάλαιος@origin"
```

## Specifications

- Platform: linux
- Version: jj 0.24.0-70e796a6b622dfdf135c353e6765b2066bb6cc62 (running from a somewhat recent version of https://github.com/jj-vcs/jj/pull/3191)

Nemo157:
Figured out the actual solution, the config should be:

```toml
[revset-aliases]
"trunk()" = "'κεφάλαιος'@origin"
```

So the error is more that when cloning the repo it doesn't correctly serialize the bookmark name into revset syntax taking into account which bookmark names need escaping.

martinvonz:
Fixed by #5475

PR #5475, opened 2025-01-26
escape bookmark and remote names in `trunk()` config

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_git_clone.rs
+++ b/cli/tests/test_git_clone.rs
@@ -601,6 +601,49 @@ fn test_git_clone_remote_default_bookmark(subprocess: bool) {
     }
 }
 
+// A branch with a strange name should get quoted in the config. Windows doesn't
+// like the strange name, so we don't run the test there.
+#[cfg(unix)]
+#[test_case(false; "use git2 for remote calls")]
+#[test_case(true; "spawn a git subprocess for remote calls")]
+fn test_git_clone_remote_default_bookmark_with_escape(subprocess: bool) {
+    let test_env = TestEnvironment::default();
+    if subprocess {
+        test_env.add_config("git.subprocess = true");
+    }
+    let git_repo_path = test_env.env_root().join("source");
+    let git_repo = git2::Repository::init(git_repo_path).unwrap();
+    set_up_non_empty_git_repo(&git_repo);
+    // Rename the main branch to something that needs to be escaped
+    git_repo
+        .find_reference("refs/heads/main")
+        .unwrap()
+        .rename("refs/heads/\"", false, "")
+        .unwrap();
+
+    let (_stdout, stderr) =
+        test_env.jj_cmd_ok(test_env.env_root(), &["git", "clone", "source", "clone"]);
+    insta::allow_duplicates! {
+    insta::assert_snapshot!(stderr, @r#"
+    Fetching into new repo in "$TEST_ENV/clone"
+    bookmark: "@origin [new] untracked
+    Setting the revset alias "trunk()" to ""\""@origin"
+    Working copy now at: sqpuoqvx cad212e1 (empty) (no description set)
+    Parent commit      : mzyxwzks 9f01a0e0 " | message
+    Added 1 files, modified 0 files, removed 0 files
+    "#);
+    }
+
+    // "trunk()" alias should be escaped and quoted
+    let stdout = test_env.jj_cmd_success(
+        &test_env.env_root().join("clone"),
+        &["config", "list", "--repo", "revset-aliases.'trunk()'"],
+    );
+    insta::allow_duplicates! {
+    insta::assert_snapshot!(stdout, @r#"revset-aliases.'trunk()' = '"\""@origin'"#);
+    }
+}
+
 #[test_case(false; "use git2 for remote calls")]
 #[test_case(true; "spawn a git subprocess for remote calls")]
 fn test_git_clone_ignore_working_copy(subprocess: bool) {
--- a/cli/tests/test_git_clone.rs
+++ b/cli/tests/test_git_clone.rs
@@ -582,11 +582,11 @@ fn test_git_clone_remote_default_bookmark(subprocess: bool) {
     }
     insta::allow_duplicates! {
     insta::assert_snapshot!(
-        get_bookmark_output(&test_env, &test_env.env_root().join("clone2")), @r###"
-    feature1@origin: mzyxwzks 9f01a0e0 message
-    main: mzyxwzks 9f01a0e0 message
+        get_bookmark_output(&test_env, &test_env.env_root().join("clone3")), @r"
+    feature1: mzyxwzks 9f01a0e0 message
       @origin: mzyxwzks 9f01a0e0 message
-    "###);
+    main@origin: mzyxwzks 9f01a0e0 message
+    ");
     }
 
     // "trunk()" alias should be set to new default bookmark "feature1"
--- a/lib/src/dsl_util.rs
+++ b/lib/src/dsl_util.rs
@@ -15,6 +15,7 @@
 //! Domain-specific language helpers.
 
 use std::array;
+use std::ascii;
 use std::collections::HashMap;
 use std::fmt;
 use std::slice;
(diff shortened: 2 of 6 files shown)


lib/src/revset.rs
```diff
@@ -2704,6 +2704,21 @@ pub struct RevsetWorkspaceContext<'a> {
     pub workspace_id: &'a WorkspaceId,
 }
 
+/// Formats a string as symbol by quoting and escaping it if necessary
+pub fn format_symbol(literal: &str) -> String {
+    let needs_escaping = if let Ok(node) = revset_parser::parse_program(literal) {
+        !matches!(node.kind, ExpressionKind::Identifier(_))
```
yuja, line 2710: This doesn't work if the literal contains parentheses (e.g. `(foo)`.) We'll need to compare `node.span`.

It might be also good to add a test function by using low-level pest API, something like `RevsetParser::parse(Rule::identifier, s)` and compare span.

lib/src/revset.rs
```diff
@@ -2704,6 +2704,21 @@ pub struct RevsetWorkspaceContext<'a> {
...
+    } else {
+        true
+    };
+    if needs_escaping {
+        let escaped = literal.replace('\\', r#"\\"#).replace('"', r#"\""#);
```
yuja, line 2715: Maybe we can apply `char::escape_default()` to chars lower than `0x80`?
https://doc.rust-lang.org/stable/std/primitive.char.html#method.escape_default

`str::escape_debug()` might also work, but it might escape some unicode characters as `\u....` (which our parser doesn't support.)

btw, I would add `dsl_util::escape_string()` where string parser is defined.

lib/src/revset.rs
yuja: > Good point about those silent parsing rules. I update the code to compare the span. Just for my understanding, would it be equivalent to compare the parsed identifier (in `ExpressionKind::Identifier(identifier)`)?

I think it'll work right now (because identifier doesn't support escape syntax.) It might break in future if we change the grammar.

> > It might be also good to add a test function by using low-level pest API, something like `RevsetParser::parse(Rule::identifier, s)` and compare span.
> 
> I didn't follow this part.

I just meant something like this would work (because we don't have to parse the full revset expression):
```rust
fn is_identifier(text: &str) -> bool {
    match RevsetParser::parse(Rule::identifier, text) {
        Ok(pairs) => pairs.next().unwrap().as_span().end() == text.len(),
        Err(_) => false,
    }
}
```
(untested)

lib/src/revset.rs
yuja: > > Maybe we can apply `char::escape_default()` to chars lower than `0x80`? https://doc.rust-lang.org/stable/std/primitive.char.html#method.escape_default
> 
> Why? Would it prevent some bugs? Or would it be to make the stored string not rely on unicode in order to make sure it's readable even if you have a simple terminal or font?

Just because it would be easier to implement. Suppose we want to reuse `escape_string()` for displaying purpose, I expect control characters in ASCII range are escaped.

yuja: I assumed `char::escape_default()` would use '\xNN` for lower values, but apparently not. It is the behavior of `ascii::escape_default()`.

lib/src/revset.rs
martinvonz: Ah, makes sense. That's much cleaner. Done.

lib/src/revset.rs
martinvonz: I think it's nice if we can allow non-ascii values without escaping, so bookmark names like the one in #5359 are easy to read.

yuja: > we can allow non-ascii values without escaping,

Yes, I agree. I just mean it's better to escape ASCII control characters (i.e. 0x00..=0x1f and 0x7f.)

cli/tests/test_git_clone.rs
```diff
@@ -599,6 +599,36 @@ fn test_git_clone_remote_default_bookmark(subprocess: bool) {
...
+        insta::assert_snapshot!(stderr, @r#"
+    Fetching into new repo in "$TEST_ENV/clone4"
+    bookmark: "@origin        [new] untracked
+    bookmark: feature1@origin [new] untracked
+    bookmark: main@origin     [new] untracked
+    Setting the revset alias "trunk()" to ""\""@origin"
+    Working copy now at: wmwvqwsz abb7c50b (empty) (no description set)
+    Parent commit      : mzyxwzks 9f01a0e0 " feature1@origin main@origin | message
+    Added 1 files, modified 0 files, removed 0 files
+    "#);
```
yuja, line 620: nit: rerun `insta` to update the indent (or split this to separate test function)?

lib/src/dsl_util.rs
```diff
@@ -429,6 +429,24 @@ impl<R: RuleType> StringLiteralParser<R> {
...
+    let mut escaped = String::with_capacity(unescaped.len());
+    for c in unescaped.chars() {
+        match c {
+            '"' => escaped.push_str(r#"\""#),
+            '\\' => escaped.push_str(r#"\\"#),
+            '\t' => escaped.push_str(r#"\t"#),
+            '\r' => escaped.push_str(r#"\r"#),
+            '\n' => escaped.push_str(r#"\n"#),
+            '\0' => escaped.push_str(r#"\0"#),
+            '\x1b' => escaped.push_str(r#"\e"#),
```
yuja, line 443: nit: I would escape `\x1b` in hex form since `\e` isn't common escape sequence.

lib/src/revset.rs
```diff
@@ -4327,4 +4336,29 @@ mod tests {
...
+        assert_eq!(format_symbol(r#"foo\bar"#), r#""foo\\bar""#);
+        assert_eq!(format_symbol(r#"foo\"bar"#), r#""foo\\\"bar""#);
+        assert_eq!(format_symbol("foo\nbar"), r#""foo\nbar""#);
+
+        // Some characters don't technically need escaping, but we escape them for
+        // clarity
+        assert_eq!(format_symbol(r#"foo"bar"#), r#""foo\"bar""#);
+        assert_eq!(format_symbol(r#"foo\bar"#), r#""foo\\bar""#);
+        assert_eq!(format_symbol(r#"foo\"bar"#), r#""foo\\\"bar""#);
+    }
```
yuja, line 4363: nit: can you add test for symbol starts/ends with space? It could be an edge case if our grammar had implicit whitespace rule.

yuja (approved):
LG, thanks.

cli/tests/test_git_clone.rs
martinvonz: Good idea to split it into a separate function.

lib/src/revset.rs
martinvonz: Oh, I see. Yes, I agree. I'll do that.

martinvonz: Done now.

lib/src/dsl_util.rs
```diff
@@ -429,6 +430,28 @@ impl<R: RuleType> StringLiteralParser<R> {
...
+    let mut escaped = String::with_capacity(unescaped.len());
+    for c in unescaped.chars() {
+        match c {
+            '"' => escaped.push_str(r#"\""#),
+            '\\' => escaped.push_str(r#"\\"#),
+            '\t' => escaped.push_str(r#"\t"#),
+            '\r' => escaped.push_str(r#"\r"#),
+            '\n' => escaped.push_str(r#"\n"#),
+            '\0' => escaped.push_str(r#"\0"#),
+            c if c < '\x20' || c == '\x7f' => {
```
yuja, line 444: nit: `.is_ascii_contro()` or range pattern can be used

lib/src/dsl_util.rs
```diff
@@ -14,6 +14,7 @@
 
 //! Domain-specific language helpers.
 
+use core::ascii;
```
yuja, line 17: nit: just import from `std::`?

yuja (approved):

lib/src/dsl_util.rs
martinvonz: Ah, nice. Switched to `.is_ascii_control()`

lib/src/dsl_util.rs
martinvonz: Done. (VS code does the imports and I usually don't check.)

PR #5476, opened 2025-01-26
cli: git clone: reload workspace after configuring new remote

yuja:
# Checklist

If applicable:


emilazy:
If my Gitoxide migration PR can cause the changes to be reflected in the underlying `ThreadSafeRepository` would this patch be unnecessary? (I haven't confirmed if that's possible/easy or not, just checking.)

yuja:
> If my Gitoxide migration PR can cause the changes to be reflected in the underlying `ThreadSafeRepository` would this patch be unnecessary?

Yes, but I don't think that's easy. The config cache doesn't support interior mutability, iirc.

Issue #80, opened 2022-02-24
Add support for Git LFS


martinvonz:
Git LFS seems to be used frequently enough that it may be worth adding support for it. I don't think it'll be a priority for me very soon, but I guess that depends on how many people want it.

[The specification](https://github.com/git-lfs/git-lfs/blob/main/docs/spec.md) says that it uses clean/smudge filters. We don't have anything like that yet. So the first step is probably to add support for that. We could make the filters only available internally (i.e. in Rust code) to start with to keep it simple. On the other hand, it might not be hard to make them user-configurable.

Another option is to add a separate file type in the data model for LFS entries. For reference, we currently have [files, symlinks, trees, conflicts, and gitmodules](https://github.com/martinvonz/jj/blob/f9298582a78a4d06e6b9c5b51c67e034e7406739/lib/src/backend.rs#L291). I haven't thought through the consequences yet. There should be no difference to the user and no difference in the representation when using the Git backend. However, clean/smudge filters are probably useful to have anyway. Oh, one *possible* advantage of representing LFS entries in the model is that we can decide to always leave merged LFS files as conflicts, without downloading the files until the user checks them out or looks at the diff etc.

I don't yet know what other aspects of Git LFS we need to consider.

Originally requested in #77.

ilyagr:
Filters are supported in libgit2, but not in rust's git2. It's also apparently not very hard to implement the logic ourselves. See https://github.com/rust-lang/git2-rs/issues/442.

martinvonz:
I'm worried about clean/smudge in general because it seems expensive to make it behave consistently. https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes says that smudging happens just before checkout and cleaning happens just before staging. If that's correct, then that seems to mean that `git diff` on a commit that changes an LFS file would just show the changed metadata, which is not very user-friendly. On the other hand, if it's not correct, and smudging happens whenever you need to present the file to the user, then there are ugly corner cases to deal with, like where `.gitattributes` changes but subtrees remain the same (then you'd technically need to diff the whole tree recursively to tell if it actually changed). I *think* I asked someone on the Git team at Google about this and they said Git just ignores the corner case(s).

ilyagr:
>  there are ugly corner cases to deal with, like where .gitattributes changes but subtrees remain the same

[...]

PR #5480, opened 2025-01-26
Ignore Git LFS files

weiznich:
This PR adds support for ignoring files handled by git lfs. `jj` previously detected these files as changed as it expected the original file content before the `git lfs` checkout to be present. With this change it's possible to opt in (via the `git.ignore-lfs-files` setting) into just ignoring this files from `jj`. They then need to be manually handled via e.g. git if they are changed.

From a technical point of view this change parses all `.gitattributes` files in the current repository and constructs a ignore list from that. This ignore list is then applied through all subcommands.

This is still missing tests, as I'm really unsure what to test about and how to write these tests. I likely won't have the capacity to figure that out on my own, so I would appreciate if someone could point to the correct places.

Somewhat addresses #80 

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* Add support for just ignoring Git LFS files by parsing `.gitattribute` files and automatically
+  adding the relevant files to the internal ignore list
+
 ### Fixed bugs
 
 ## [0.27.0] - 2025-03-05
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1338,6 +1338,32 @@ to the current parents may contain changes from multiple commits.
         self.path_converter().parse_file_path(input)
     }
 
+    /// Parses the given strings as file patterns, convert it to a matcher and
+    /// apply Git LFS exclusions if requested
+    pub fn file_matcher(
+        &self,
+        ui: &Ui,
+        values: &[String],
+    ) -> Result<Box<dyn Matcher>, CommandError> {
+        let mut matcher = self.parse_file_patterns(ui, values)?.to_matcher();
+        if self.settings().git_settings()?.ignore_lfs_files {
+            let root = self.workspace().workspace_root();
+            if let Some(wc_commit) = self
+                .get_wc_commit_id()
+                .map(|id| self.repo().store().get_commit(id))
+                .transpose()?
+            {
+                let tree = wc_commit.tree()?;
+                matcher = Box::new(jj_lib::matchers::DifferenceMatcher::new(
+                    matcher,
+                    jj_lib::matchers::GitAttributesMatcher::new(tree.as_merge().first(), root)
+                        .expect("gitattributes matcher failed"),
+                ));
+            }
+        }
+        Ok(matcher)
+    }
+
     /// Parses the given strings as file patterns.
     pub fn parse_file_patterns(
         &self,
@@ -1404,6 +1430,7 @@ to the current parents may contain changes from multiple commits.
             start_tracking_matcher,
             max_new_file_size,
             conflict_marker_style,
+            skip_git_lfs_files: self.settings().git_settings()?.ignore_lfs_files,
         })
     }
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -1371,6 +1371,17 @@ subprocess = false
 Note that `libgit2` support will likely be removed in the future, so you are
 encouraged to report any issues you experience with the default configuration.
 
+### Ignore Git LFS Files
+
+By default Git LFS files are **not** handled by `jj`. This will result in `jj` showing
+changes in these files, even if they are unchanged. You can configure `jj` to ignore these
+files by instructing it to parse the relevant `.gitattributes` files
+
+```toml
+[git]
+ignore-lfs-files = true
+```
+
 ## Filesystem monitor
 
 In large repositories, it may be beneficial to use a "filesystem monitor" to
--- a/lib/src/matchers.rs
+++ b/lib/src/matchers.rs
@@ -18,13 +18,20 @@ use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fmt;
 use std::fmt::Debug;
+use std::fs::File;
+use std::io;
+use std::io::BufRead;
+use std::io::BufReader;
 use std::iter;
+use std::path::Path;
 
 use itertools::Itertools as _;
+use thiserror::Error;
 use tracing::instrument;
 
 use crate::repo_path::RepoPath;
 use crate::repo_path::RepoPathComponentBuf;
+use crate::tree::Tree;
 
 #[derive(PartialEq, Eq, Debug)]
 pub enum Visit {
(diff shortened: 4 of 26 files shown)


docs/config.md
```diff
@@ -1258,6 +1258,17 @@ particular binary, you can:
...
 
+### Ignore Git LFS Files
+
+By default Git LFS files are **not** handled by `jj`. This will result in `jj` showing
+changes in these files, even if they are unchanged. You can configure `jj` to ignore these
+files by instructing it to parse the relevant `.gitattributes` files
+
+```tom
+[git]
+ignore-lfs-files = -rue
```
istudyatuni, line 1269:
```suggestion
ignore-lfs-files = true
```

docs/config.md
```diff
@@ -1258,6 +1258,17 @@ particular binary, you can:
 executable-path = "/path/to/git"
 ```
 
+### Ignore Git LFS Files
+
+By default Git LFS files are **not** handled by `jj`. This will result in `jj` showing
+changes in these files, even if they are unchanged. You can configure `jj` to ignore these
+files by instructing it to parse the relevant `.gitattributes` files
+
+```tom
```
istudyatuni, line 1267:
```suggestion
```toml
```

lib/src/matchers.rs
```diff
@@ -506,6 +513,132 @@ impl<V: Debug> Debug for RepoPathTree<V> {
...
+pub enum GitAttributesMatcherError {
+    #[error("failed to parse .gitattributes file")]
+    GitIgnoreError(#[from] ignore::Error),
+    #[error("failed to read .gitattributes file")]
+    ReadError(#[from] io::Error),
+    #[error("invalid pattern in .gitattributes: {0}")]
+    InvalidPattern(String),
+}
+
+fn parse_gitattributes(
```
istudyatuni, line 545: There is a [gix-attributes](https://crates.io/crates/gix-attributes) crate, probably it's better to use it?

PhilipMetzger:
Please adhere to the commit style guidelines outlined here: https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#code-reviews

weiznich:
I'm sorry, but it's unclear for me to which part of the guide lines you exactly refer and also how that interacts with this kind on commit structure with commits from several authors. For example I fear it's not possible to just squash contributions from other persons into my commits.

PhilipMetzger:
> I'm sorry, but it's unclear for me to which part of the guide lines you exactly refer and also how that interacts with this kind on commit structure with commits from several authors.

I don't think matching the `topic:` commit description structure and explaining what your commit does is a hard ask, as the project just doesn't squash these commits. 

> For example I fear it's not possible to just squash contributions from other persons into my commits.

For squashing others commits you can manually add the `co-authored-by:` trailer to still the credit the original author. This should also be no issue as long as the original author is fine with it, which I assume. cc @bcspragu

.ignore
PhilipMetzger, line 1: nit: remove

bcspragu:
> For squashing others commits you can manually add the co-authored-by: trailer to still the credit the original author. This should also be no issue as long as the original author is fine with it, which I assume. cc @bcspragu

100%, please do!

cli/src/cli_util.rs
```diff
@@ -1277,6 +1277,32 @@ to the current parents may contain changes from multiple commits.
         self.path_converter().parse_file_path(input)
     }
 
+    /// Parses the given strings as file patterns, convert it to a matcher and
+    /// apply GIT LFS exclusions if requested
```
PhilipMetzger, line 1281: Nit: Its Git LFS

CHANGELOG.md
PhilipMetzger, line 1: Nit: squash this into the first commit which successfully implements this.

PhilipMetzger (changes requested):
Initial minor nits, also cc @yuja as he knows the matcher internals better and should give the final approval.

ilyagr:
Are you able to sign the CLA?

yuja:
> he knows the matcher internals better

I haven't reviewed the code, but I wonder if gitattributes could really be abstracted as a `Matcher`. We deal with `.gitignore` differently because ignore files have to be loaded from sub trees, and I assume `.gitattributes` are similar?

If the attributes patterns can be converted to a `Matcher`, it might make sense to add `load_git_attributes(path) -> Result<Matcher, _>`, but I wouldn't expect `GitAttributesMatcher` that does parsing and other heavy lifting within the matcher.

weiznich:
> I haven't reviewed the code, but I wonder if gitattributes could really be abstracted as a Matcher. We deal with .gitignore differently because ignore files have to be loaded from sub trees, and I assume .gitattributes are similar?
>
>If the attributes patterns can be converted to a Matcher, it might make sense to add load_git_attributes(path) -> Result<Matcher, _>, but I wouldn't expect GitAttributesMatcher that does parsing and other heavy lifting within the matcher.

Yes `.gitattributes` files are very similar in that regard to `.gitignore` files. They can be "everywhere" in the tree. This change don't handle the parsing and other heavy lifting as you call it in the matcher itself, but while creating the matcher. If that's not the expected solution I would like to get some more details on what's the expected solution. Depending on the required effort I might find the time to look into that or I just stay on my fork and leave that be as the implemented solution works well for me on a large repo with several nested gitignore files.

lib/src/matchers.rs
yuja: We already use `gix`, so enabling some additional features should be fine.

yuja:
> Yes `.gitattributes` files are very similar in that regard to `.gitignore` files. They can be "everywhere" in the tree. This change don't handle the parsing and other heavy lifting as you call it in the matcher itself, but while creating the matcher. If that's not the expected solution I would like to get some more details on what's the expected solution.

I expect something that's similar to how we handle `.gitignore`.

```rust
    fn visit_directory<'scope>(
        &'scope self,
        directory_to_visit: DirectoryToVisit<'scope>,
        scope: &rayon::Scope<'scope>,
    ) -> Result<(), SnapshotError> {
        let DirectoryToVisit {
            dir,
            disk_dir,
            git_ignore,
            git_attributes, // ADD some object that accumulates attributes?
            file_states,
        } = directory_to_visit;

        let git_ignore = git_ignore
            .chain_with_file(&dir.to_internal_dir_string(), disk_dir.join(".gitignore"))?;
        let git_attributes = git_attributes...(disk_dir.join(".gitattributes"))?;
            // UPDATE it for the current directory
```

bcspragu:
@weiznich Thanks for all your work on it, I'm happy to pick up the torch (and continue our relay race :smile:) if it's becoming too time-consuming.  After all, the current undesirable architecture is from my lack of knowledge of the internals anyway, so with @yuja's pointers from above, I can probably get it to the desired state.

PhilipMetzger:
@bcspragu are you able to finish this work? As, it seems the author no longer has any time for it. That would be nice.

bcspragu:
Yup! I'll poke at it this week and see where that leads

PhilipMetzger:
Thank you.

kejadlen:
@bcspragu I did some preliminary poking at it: https://github.com/jj-vcs/jj/compare/main...kejadlen:jj:lfs

There's a fair amount of cleanup work to do between adding configuration, documentation, removing unused parameters, and probably a bit of refactoring as well, but it seems to work?

bcspragu:
Thanks for the poking @kejadlen, it looks great! I've started making (small) tweaks to that in https://github.com/jj-vcs/jj/compare/main...bcspragu:jj:lfs

There's a bunch of duplication between `GitIgnore` and `GitAttributes` now, but that's to be expected given the architecture. I think it's fine, but in a parallel universe where there were other similar `.git*` config files to ignore, we'd likely want to refactor that a bit.

I'm hoping to get this finished up tomorrow: all the assorted cleaning/refactoring and (hopefully!) adding a few tests. Once that's ready, I'll fix up the commit structure and send out a PR

bcspragu:
Did some more work on this today, added some tests. Some are failing, and I need to spend more time understanding the `gix_attributes` stuff to wrap my head around how to use it correctly. I created https://github.com/jj-vcs/jj/pull/6036 as a draft to stop abusing this closed PR, I'll put future updates there

PR #5491, opened 2025-01-27
Build Nix flake checks in `test` mode

thoughtpolice:
Also includes an extra tidbit to track down slow tests in CI.

# Checklist

If applicable:

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,3 @@
+[profile.ci]
+slow-timeout = { period = "5s", terminate-after = 20 }
+fail-fast = false
--- a/flake.nix
+++ b/flake.nix
@@ -100,7 +100,6 @@
       };
     in {
       formatter = pkgs.alejandra;
-      checks.jujutsu = self.packages.${system}.jujutsu;
 
       packages = {
         jujutsu = rustMinimalPlatform.buildRustPackage {
@@ -149,6 +148,21 @@
         };
         default = self.packages.${system}.jujutsu;
       };
+
+      checks.jujutsu = self.packages.${system}.jujutsu.overrideAttrs ({...}: {
+        # The default Rust infrastructure runs all builds in the release
+        # profile, which is significantly slower. Run this under the `test`
+        # profile instead, which matches all our other CI systems, Cargo, etc.
+        cargoBuildType = "test";
+        cargoCheckType = "test";
+
+        # By default, `flake check` will want to run the install phase, but
+        # because we override the cargoBuildType, it fails to find the proper
+        # binary. But we don't even care about the binary or even the buildPhase
+        # in this case; just remove them both.
+        buildPhase = "true";
+        installPhase = "touch $out";
+      });
 
       devShells.default = let
         packages = with pkgs; [
--- a/flake.nix
+++ b/flake.nix
@@ -40,12 +40,25 @@
             pkgs.lib.all (re: builtins.match re relPath == null) regexes;
         };
 
-      ourRustVersion = pkgs.rust-bin.selectLatestNightlyWith (toolchain: toolchain.default);
-
-      ourRustPlatform = pkgs.makeRustPlatform {
-        rustc = ourRustVersion;
-        cargo = ourRustVersion;
+      # When we're running in the shell, we want to use rustc with a bunch
+      # of extra junk to ensure that rust-analyzer works, clippy etc are all
+      # installed.
+      rustShellToolchain = (pkgs.rust-bin.selectLatestNightlyWith (t: t.default)).override {
+        # NOTE (aseipp): explicitly add rust-src to the rustc compiler only in
+        # devShell. this in turn causes a dependency on the rust compiler src,
+        # which bloats the closure size by several GiB. but doing this here and
+        # not by default avoids the default flake install from including that
+        # dependency, so it's worth it
+        #
+        # relevant PR: https://github.com/rust-lang/rust/pull/129687
+        extensions = ["rust-src" "rust-analyzer"];
       };
+
+      # But, whenever we are running CI builds or checks, we want to use a
+      # smaller closure. This reduces the CI impact on fresh clones/VMs, etc.
+      rustMinimalPlatform =
+        let platform = pkgs.rust-bin.selectLatestNightlyWith (t: t.minimal);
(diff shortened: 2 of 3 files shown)


.config/nextest.toml
```diff
@@ -0,0 +1,3 @@
+[profile.ci]
+slow-timeout = { period = "8s", terminate-after = 20 }
```
ilyagr, line 2: Will it retry the test if `terminate-after` happens? If not, I'd use a longer timeout here, say 60 seconds, since we don't want to increase flakyness.

flake.nix
```diff
@@ -136,6 +136,21 @@
...
+        # profile instead, which matches all our other CI systems, Cargo, etc.
+        cargoBuildType = "test";
+        cargoCheckType = "test";
+
+        # By default, `flake check` will want to run the install phase, but
+        # because we override the cargoBuildType, it fails to find the proper
+        # binary. But we don't even care about the binary or even the buildPhase
+        # in this case; just remove them both.
+        buildPhase = "true";
+        installPhase = "touch $out";
```
ilyagr, line 151: If I run `nix profile install` on this, or do `nix shell`, etc, will it install an optimized version of `jj`? I'm guessing this is what people want, though I guess installing a debug version wouldn't be the end of the world.

ilyagr (approved):
I'm mostly approving this based on the CI, in the hopes it will decrease flakyness. We can revert things if they don't work. Or, if some more Nix-y person wants to review this before it's merged, please leave a comment.

I also did check that `nix flake check`, `nix build`, and `nix shell` work with this, though I'm not sure how optimized they are.

.config/nextest.toml
bryceberger: It terminates without retry. However, the 20 is 20 _periods_, so 160  seconds.

flake.nix
bryceberger: `nix build` shows `jujutsu> cargoBuildHook flags: -j 16 --target x86_64-unknown-linux-gnu --offline --profile release --features=packaging --bin jj`, so it should be a release build

.config/nextest.toml
ilyagr: I see, thanks! Optionally, it might be worth adding a comment about that. I think Austin's commit description also mentioned 20s, or I misread it.

ilyagr: I'm still wondering about whether it's better to leave out `terminate-after` entirely, but it doesn't seem so important now. In my mind, it depends on the probability that, in CI, tests will be waiting on each other instead of running in parallel.

flake.nix
```diff
@@ -136,6 +136,21 @@
         default = self.packages.${system}.jujutsu;
       };
 
+      checks.jujutsu = self.packages.${system}.jujutsu.overrideAttrs ({ ...}: {
```
bryceberger, line 139: Could we also override the toolchain here? Use `toolchain.minimal` instead of `toolchain.default` to not install clippy, rustfmt, etc

bryceberger: Actually, instead of overriding, we should probably just use `toolchain.minimal` for builds, and `toolchain.default` only in the shell?

bryceberger:
```diff
Commit ID: 5682c3c836b6cbaea3878ecaa3b169d6ed0b40df
Change ID: ypxnoyzsmlvkkusyvqqnmylwozlvtryq
Author   : Bryce Berger <bryce.z.berger@gmail.com> (2025-01-27 19:59:18)
Committer: Bryce Berger <bryce.z.berger@gmail.com> (2025-01-27 20:51:43)
Signature: good signature by Bryce Berger <bryce.z.berger@gmail.com> 58CA4F9FEF8F4296

    (no description set)

diff --git a/flake.nix b/flake.nix
index f2bd9dc7e6..1940db62dd 100644
--- a/flake.nix
+++ b/flake.nix
@@ -40,12 +40,13 @@
             pkgs.lib.all (re: builtins.match re relPath == null) regexes;
         };
 
-      ourRustVersion = pkgs.rust-bin.selectLatestNightlyWith (toolchain: toolchain.default);
-
-      ourRustPlatform = pkgs.makeRustPlatform {
-        rustc = ourRustVersion;
-        cargo = ourRustVersion;
-      };
+      rustDefault = pkgs.rust-bin.selectLatestNightlyWith (toolchain: toolchain.default);
+      rustMinimal = pkgs.rust-bin.selectLatestNightlyWith (toolchain: toolchain.minimal);
+      makeRustPlatform = platform:
+        pkgs.makeRustPlatform {
+          rustc = platform;
+          cargo = platform;
+        };
 
       nativeBuildInputs = with pkgs;
         [
@@ -89,7 +90,7 @@
       formatter = pkgs.alejandra;
 
       packages = {
-        jujutsu = ourRustPlatform.buildRustPackage {
+        jujutsu = (makeRustPlatform rustMinimal).buildRustPackage {
           pname = "jujutsu";
           version = "unstable-${self.shortRev or "dirty"}";
 
@@ -136,7 +137,7 @@
         default = self.packages.${system}.jujutsu;
       };
 
-      checks.jujutsu = self.packages.${system}.jujutsu.overrideAttrs ({ ...}: {
+      checks.jujutsu = self.packages.${system}.jujutsu.overrideAttrs ({...}: {
         # The default Rust infrastructure runs all builds in the release
         # profile, which is significantly slower. Run this under the `test`
         # profile instead, which matches all our other CI systems, Cargo, etc.
@@ -160,7 +161,7 @@
           # that dependency, so it's worth it
           #
           # relevant PR: https://github.com/rust-lang/rust/pull/129687
-          (ourRustVersion.override {
+          (rustDefault.override {
             extensions = ["rust-src" "rust-analyzer"];
           })
 
```

.config/nextest.toml
thoughtpolice: FWIW, my hope was to just make it gracious enough that even really highly loaded CI runners can probably make progress. If something is so bad that a test can't finish in ~3 minutes, I think it should probably fail so we can investigate it a bit.

flake.nix
thoughtpolice: This change only takes effect when someone runs `nix flake check` on the source cod directly. `nix profile install`, `nix shell`, etc are all entirely unaffected.

PR #5497, opened 2025-01-27
docs: explain `[noeol]` conflict marker comment

scott2000:
The meaning of `[noeol]` might not be immediately clear to a user, so it would be good to document it on our documentation page for conflicts. We may also want to improve the conflict marker comments we use for this case in the future (possibly after receiving user feedback).

Follow up to #5186.

# Checklist

If applicable:

### DIFF
--- a/docs/conflicts.md
+++ b/docs/conflicts.md
@@ -184,3 +184,36 @@ New Heading
 ===========
 >>>>>>>>>>>>>>> Conflict 1 of 1 ends
 ```
+
+## Conflicts with missing terminating newline (`[noeol]`)
+
+When materializing conflicts, `jj` outputs them in a line-based format. This
+format is easiest to interpret for text files that consist of a series of lines,
+with each line terminated by a newline character (`\n`). This means that a text
+file should either be empty, or it should end with a newline character.
+
+While most text files follow this convention, some do not. To handle this case,
+when `jj` encounters a missing terminating newline character in a conflict, it
+will add a `[noeol]` comment to the conflict markers to indicate that the
+"end-of-line" (EOL) character was missing. If you don't care about whether your
+file ends with a terminating newline character, you can generally ignore this
+comment and resolve the conflict normally.
+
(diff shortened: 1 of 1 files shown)


docs/conflicts.md
```diff
@@ -184,3 +184,32 @@ New Heading
 ===========
 >>>>>>>>>>>>>>> Conflict 1 of 1 ends
 ```
+
+## Conflicts with missing terminating newline (`[noeol]`)
+
+On POSIX systems, text files are generally assumed to be a series of lines with
```
ilyagr, line 190: Suggestion: (feel free to reword it further or not to use it)

> `jj` relies on line-based diff algorithms, which work best on text files that consist of a series of lines with each line terminated by a newline character. This means that a text file would either be empty or it would terminate in a newline.
>
> While most text files follow this convention, some do not. When `jj`....

ilyagr (approved):
This LGTM.

Thank you also for continuing to work on this. I'm sorry I haven't paid as much attention to this as I would have liked.

Please also give this a bit of time before merging, in case Yuya, Waleed, or others have comments.

docs/conflicts.md
ilyagr: Thanks, LGTM.

arxanas (approved):
Docs look good to me overall, thanks for adding this.

[no action required] Looking at the docs now, I think the square brackets and abbreviated "keyword'-ish name make me think that the conflict marker has semantic meaning. If it had just written `(no trailing newline)` like normal prose then I would have assumed that it was primarily for my benefit as a reader, rather than machine-meaningful content.

scott2000:
> [no action required] Looking at the docs now, I think the square brackets and abbreviated "keyword'-ish name make me think that the conflict marker has semantic meaning. If it had just written `(no trailing newline)` like normal prose then I would have assumed that it was primarily for my benefit as a reader, rather than machine-meaningful content.

That's a good point. I think I'll merge this PR, and then make a separate PR that makes the wording a bit more clear as you suggested.

PR #5498, opened 2025-01-28
git: update jj git clone|fetch to use new GitFetch api directly

bsdinis:
Make the new GitFetch api public.
   - Move git::fetch to lib/tests/test_git.rs as git_fetch, to minimize churn in the tests. Update test call sites to use git_fetch
   - Delete the git::fetch from lib/src/git.rs.
   - Update jj git clone and git_fetch in cli/src/git_utils.rs to use the new api directly. Removing one redundant layer of indirection.
   - This fixes https://github.com/jj-vcs/jj/issues/4920 as it first fetches from all remotes before import_refs() is called, so there is no race condition if the same commit is treated differently in different remotes specified in the same command.

Original commit by @essiene https://github.com/jj-vcs/jj/pull/4960

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_git_fetch.rs
+++ b/cli/tests/test_git_fetch.rs
@@ -17,14 +17,10 @@ use test_case::test_case;
 
 use crate::common::TestEnvironment;
 
-/// Creates a remote Git repo containing a bookmark with the same name
-fn init_git_remote(test_env: &TestEnvironment, remote: &str) {
-    let git_repo_path = test_env.env_root().join(remote);
-    let git_repo = git2::Repository::init(git_repo_path).unwrap();
-    let signature =
-        git2::Signature::new("Some One", "some.one@example.com", &git2::Time::new(0, 0)).unwrap();
+fn add_commit_to_branch(git_repo: &git2::Repository, branch: &str) -> git2::Oid {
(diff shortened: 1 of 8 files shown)


bsdinis:
@yuja had to abandon the other PR (the target branch was not going to be merged).
I think I've addressed your comments in this version though. Also added more refactory commits.

cli/tests/test_git_fetch.rs
```diff
@@ -1764,3 +1788,111 @@ fn test_git_fetch_remote_only_bookmark(subprocess: bool) {
     "###);
     }
 }
+
+#[test_case(false; "use git2 for remote calls")]
+#[test_case(true; "spawn a git subprocess for remote calls")]
+fn test_git_fetch_preserve_commits_across_repos(subprocess: bool) {
+    let test_env = TestEnvironment::default();
+    if subprocess {
+        test_env.add_config("git.subprocessing = true");
```
yuja, line 1797: typo: `git.subprocess`

yuja (approved):
Thanks.

cli/tests/test_git_fetch.rs
bsdinis: done

PR #5501, opened 2025-01-28
docs: add new logo as favicon

martinvonz:
# Checklist

If applicable:


docs/favicon.png
ilyagr, line 1: I would put it in `docs/images/favicon.png`.

It's fine for now, but I think 271KB is a bit large for a favicon. We probably will want to update it after we get an uncompressed version.

ilyagr (commented):
We can test this out. It will only affect the `prerelease` site.

To affect everything, we could try putting it as `favicon.png` in https://github.com/jj-vcs/jj-vcs.github.io. (We might have to experiment, I'm not sure whether it needs to be ICO or can be PNG). The comment about the size applies to that too.

docs/favicon.png
ilyagr: For now, let's just use the 40KB JPEG as is. I think it's the best version we have, and it'll blow up the size of the repo by only 40KB.

ilyagr: We can also call it something like `jj-logo.jpg`, since I think we'll want to reference it from  the README and a few more places.

martinvonz: I shrunk it to 192x192. I also added the license to the LICENSE file in the repo root.

ilyagr: Is there any reason *not* to use the JPG? I assume you got the PNG from the JPG, ultimately.

ilyagr: Also, if we wait a bit, we might get the original.

martinvonz: I got the png from the webp file. I don't know if it's been through a conversion to jpg somewhere on the way (e.g on Discord's servers).

LICENSE
```diff
@@ -200,3 +200,9 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
```
ilyagr, line 203: (Optional) I'm not sure what the official advice is, but I would do the following:

In the beginning of this doc, say:

> See docs/images/LICENSE for the license that applies to the files in the docs/images directory.
>
> The rest of `jj` is licensed under the terms below.

Then, I'd create the docs/images/LICENSE file with what you wrote here.

jennings:
I’ll get you a bigger PNG and SVG tomorrow, the file sent over Discord will probably be artifacty. I just need to make sure there’s no gross metadata in the files before I put them on the internet.

I think [this is the favicon generator I’ve used in the past](https://realfavicongenerator.net/), I recall there being weird platform quirks about getting the high quality versions to show up everywhere. This may be overkill for the docs site though.

docs/favicon.png
ilyagr: I'm guessing the webp file was obtained *from* the JPG file, making it strictly worse. But I might be wrong. In any case, it's not really important anymore.

LICENSE
martinvonz: I don't know what the best practice is either. I moved it to `docs/images/LICENSE`.

ilyagr: If we get better advice, we can always adjust it.

jennings:
I don't trust Discord so I pushed a [full size image here to a branch](https://github.com/jennings/jj/blob/jujutsu-logo/docs/images/jujutsu-logo.png).

martinvonz:
Thanks! I uploaded that image to the favicon generator you linked to. I added the same background color as we use elsewhere. Then I took the generated `favicon.ico` file and put in the repo. I hope that's what I was supposed to do. The GitHub UI doesn't render the favicon file but hopefully most browsers understand the format.

jennings:
I think ico is deprecated for favicons, it’s there as a last resort. All modern browsers will use the higher quality PNG versions if the link tags exist for it.

Let me give it a try soon and post a suggested here?

I suppose it would be find to merge as is and improve it later.

martinvonz:
> Let me give it a try soon and post a suggested here?

Thanks. I'll wait.

jennings:
Oh I see, MkDocs doesn't have a great way to add extra `<link>` elements to the page header without extending the theme, which we don't do yet.

The only change I would probably make, then, is to use the 96x96 PNG that the favicon generator spits out, instead of an ICO file. Comparing the quality:

image

martinvonz:
Done

martinvonz:
@ilyagr and @jennings, are you okay with the current version?

ilyagr:
I think what you have now is fine, further changes could go to another PR.

I am wondering about changing the background, though. Currently, this favicon (visible via mkdocs serve) looks like this compared to others:

image

I'd consider either keeping the background transparent or rounding the corners of the colored background (leaving transparent color where the sharp corners are removed). In the first case, the birds should become larger.

Do you want to experiment with this now or to postpone this to future work and PRs?

martinvonz:
> Do you want to experiment with this now or to postpone this to future work (possibly by others) and PRs?

I think we can change that later if we feel like it. FWIW, I have a few icons among my current tabs with a square favicon. They all seem to be for internal Google tools, though, so maybe it is indeed unconventional outside of Google.

ilyagr:
OK.

I quickly tried a transparent version, and it's not better. We might need a different icon for the favicon, either with simpler birds, or just one bird, or I'm not sure.

image

and

image


But for now, we can just go with this IMO.

ilyagr (approved):

PR #5502, opened 2025-01-28
SECURITY.md: insert an explicit URL

ilyagr:
The previous "on this page" statement is wrong more often than not. Unfortunately there is no "Report a vulnerability" button on https://github.com/jj-vcs/jj/security/policy, and looking for such a button from https://github.com/jj-vcs/jj?tab=security-ov-file leads to confusion.

This is not the end of the world, but I don't see much security downside to clarifying it (that is, I don't think *not* having a link protects against phishing in any real way).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,7 +1,10 @@
 To report a security issue, please use the "Report a vulnerability" button on
-this page. Our vulnerability management team will respond within 3 working days
-of your report. If the issue is confirmed as a vulnerability, we will open a
-Security Advisory. This project follows a 90 day disclosure timeline.
+GitHub's Security tab for `jj`'s main repo, under
+[Advisories](https://github.com/jj-vcs/jj/security/advisories).
+
+Our vulnerability management team will respond within 3 working days of your
+report. If the issue is confirmed as a vulnerability, we will open a Security
+Advisory. This project follows a 90 day disclosure timeline.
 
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -1,7 +1,9 @@
 To report a security issue, please use the "Report a vulnerability" button on
-this page. Our vulnerability management team will respond within 3 working days
-of your report. If the issue is confirmed as a vulnerability, we will open a
-Security Advisory. This project follows a 90 day disclosure timeline.
+GitHub's Security tab for `jj`'s main repo, also accessible at
+<https://github.com/jj-vcs/jj/security/advisories>. Our vulnerability management
```
martinvonz: What do you think about simplifying to "... button under [Advisories](https://github.com/jj-vcs/jj/security/advisories)."?

ilyagr: How about this?

Because of my (armchair) concern about phishing, I also wanted to give people enough breadcrumbs to be able to find the right place without using the link, in case they feel (rightly or wrongly) worried about using the link.

martinvonz: I was thinking that the reader was accessed this doc via https://github.com/jj-vcs/jj/security/policy. I think you're saying that it may be useful to give the full instructions in case they find this doc in the source tree. Makes sense.

martinvonz:
I agree. Sorry, I forgot that I had not approved this.

ilyagr:
No problem, thanks!

PR #5508, opened 2025-01-29
git: forward "remote: " lines when they arrive

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/git_subprocess.rs
+++ b/lib/src/git_subprocess.rs
@@ -11,7 +11,10 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-//
+
+use std::io;
+use std::io::BufReader;
+use std::io::Read;
 use std::num::NonZeroU32;
 use std::path::Path;
 use std::path::PathBuf;
@@ -19,6 +22,7 @@ use std::process::Child;
 use std::process::Command;
 use std::process::Output;
 use std::process::Stdio;
+use std::thread;
 
 use bstr::ByteSlice;
 use thiserror::Error;
@@ -55,8 +59,6 @@ pub(crate) struct GitSubprocessContext<'a> {
     git_dir: PathBuf,
     git_executable_path: &'a Path,
 }
-
-pub(crate) type SidebandProgressCallback<'a> = &'a mut dyn FnMut(&[u8]);
 
 impl<'a> GitSubprocessContext<'a> {
     pub(crate) fn new(git_dir: impl Into<PathBuf>, git_executable_path: &'a Path) -> Self {
@@ -131,9 +133,9 @@ impl<'a> GitSubprocessContext<'a> {
         command.arg("--").arg(remote_name);
         command.args(refspecs.iter().map(|x| x.to_git_format()));
 
-        let output = wait_with_output(self.spawn_cmd(command)?)?;
+        let output = wait_with_progress(self.spawn_cmd(command)?, callbacks)?;
 
-        parse_git_fetch_output(output, &mut callbacks.sideband_progress)
+        parse_git_fetch_output(output)
     }
 
     /// Prune particular branches
@@ -190,7 +192,7 @@ impl<'a> GitSubprocessContext<'a> {
         &self,
         remote_name: &str,
         references: &[RefToPush],
-        mut callbacks: RemoteCallbacks<'_>,
+        callbacks: &mut RemoteCallbacks<'_>,
     ) -> Result<(Vec<String>, Vec<String>), GitSubprocessError> {
         let mut command = self.create_command();
         command.stdout(Stdio::piped());
@@ -209,9 +211,9 @@ impl<'a> GitSubprocessContext<'a> {
                 .map(|r| r.refspec.to_git_format_not_forced()),
         );
 
-        let output = wait_with_output(self.spawn_cmd(command)?)?;
+        let output = wait_with_progress(self.spawn_cmd(command)?, callbacks)?;
 
-        parse_git_push_output(output, &mut callbacks.sideband_progress)
+        parse_git_push_output(output)
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/src/git_subprocess.rs
```diff
@@ -497,27 +504,72 @@ fn parse_git_push_output(
...
+    let (stdout, stderr) = thread::scope(|s| -> io::Result<_> {
+        drop(child.stdin.take());
+        let mut child_stdout = child.stdout.take().expect("stdout should be piped");
+        let mut child_stderr = child.stderr.take().expect("stderr should be piped");
+        let thread = s.spawn(move || -> io::Result<_> {
+            let mut buf = Vec::new();
+            child_stdout.read_to_end(&mut buf)?;
+            Ok(buf)
+        });
+        let stderr = read_to_end_with_sideband(&mut child_stderr, BUF_SIZE, sideband_cb)?;
```
bsdinis, line 521: nit: although in the terminal `stderr` looks small, the actual byte representation is big (there are a lot of `\r` to update the progress bits).

lib/src/git_subprocess.rs
```diff
@@ -131,9 +132,14 @@ impl<'a> GitSubprocessContext<'a> {
         command.arg("--").arg(remote_name);
         command.args(refspecs.iter().map(|x| x.to_git_format()));
 
-        let output = wait_with_output(self.spawn_cmd(command)?)?;
+        let child = self.spawn_cmd(command)?;
+        let output = if let Some(cb) = callbacks.sideband_progress.as_mut() {
```
bsdinis, line 136: I have some reservations about this.
Mainly that we get a small combinatorial explosion with `callbacks.progress`, which is also an `Option<&mut Fn...>`. 

Maybe replacing this `if let` with a `wait_with_callbacks(child, callbacks)` which internally can even `unwrap_or` the `Option`s with no-ops?

bsdinis (commented):
Thanks for this one!
It overall looks awesome and much cleaner.
I'm pretty sure it conflicts with #5507 but we should merge this before as it gives a more solid foundation

lib/src/git_subprocess.rs
bsdinis: Note: this doesn't have to be handled in this PR

lib/src/git_subprocess.rs
yuja: Yes, if we want to parse and reproduce progress message, it might be better to add separate scratch buffer for unparsed chunk of data. It's unlikely we would want to extract error message from buffer containing random junk of progress information.

lib/src/git_subprocess.rs
yuja: At this point, unwrapping Option here was the easiest. It might make sense to pass Option or map None to null callback to support multiple callbacks.

lib/src/git_subprocess.rs
yuja: Add `BufReader` version on top of the old patch. I'll squash if it looks good.

yuja: Maybe I misunderstood the original point. If `--progress` is passed to `git`, remote sends back progress messages by abusing `\r`, so the sideband message content can be big. Indeed. We'll need a separate buffer for that.

yuja: Squashed `BufReader` version, and added handling of `\r`. Sideband lines are now excluded from the `stderr` content.

lib/src/git_subprocess.rs
yuja: Changed to pass `callbacks` down to the wait function. Since `wait_with_progress()` now filters out `remote: ` lines, it doesn't make sense to use `wait_with_output()` conditionally.

lib/src/git_subprocess.rs
```diff
@@ -497,27 +495,84 @@ fn parse_git_push_output(
...
+            }
+            data.truncate(start);
+        }
     }
+    Ok(data)
 }
 
-fn wait_with_output(child: Child) -> Result<Output, GitSubprocessError> {
-    child.wait_with_output().map_err(GitSubprocessError::Wait)
+fn read_until_cr_or_lf<R: io::BufRead + ?Sized>(
```
bsdinis, line 550: nit: I think it would be cleaner for this function to return a `Vec<u8>` or handle the line directly.

If the function returns a `Vec<u8>`, the caller would only `extend` or `append` the line if the line is not a sideband progress line (or an actual progress line).

Alternatively, we could pass a line processing closure here, something like
```rust
fn read_until_cr_or_lf<R>(reader: &mut R, dest_buf: &mut Vec<u8>, line_handler: F) -> io::Result<()> 
  where F: FnMut(&[u8]) -> bool
{
    loop {
        let data = match reader.fill_buf() {
            Ok(data) => data,
            Err(err) if err.kind() == io::ErrorKind::Interrupted => continue,
            Err(err) => return Err(err),
        };
        let (n, found) = match data.iter().position(|&b| matches!(b, b'\r' | b'\n')) {
            Some(i) => (i + 1, true),
            None => (data.len(), false),
        };

        if found || n == 0 {
           if !line_handler(&data[..n]) {
              dest_buf.extend_from_slice(&data[..n]);
           }
           reader.consume(n);
           return Ok(());
        }
    }    
}
```

The upside here would be to avoid `extend` + `truncate` while also avoiding an extra allocation to be returned to the caller (although, since the buffer here should never be more than 100 or so characters (line width) I wouldn't be too worried about it)

yuja: > nit: I think it would be cleaner for this function to return a `Vec<u8>` or handle the line directly.
> 
> If the function returns a `Vec<u8>`, the caller would only `extend` or `append` the line if the line is not a sideband progress line (or an actual progress line).

One reason is to avoid extra allocation, though it wouldn't be super important here. Another reason is that I just copied the signature of library function `read_until()` minus the number of read bytes. The caller can pass in an empty `&mut Vec` if they want a separate line buffer.

> Alternatively, we could pass a line processing closure here, something like
> 
> ```rust
> fn read_until_cr_or_lf<R>(reader: &mut R, dest_buf: &mut Vec<u8>, line_handler: F) -> io::Result<()> 
>   where F: FnMut(&[u8]) -> bool

Yes, it also works. It also means `dest_buf` can be removed since the `line_handler` can append to `dest_buf` conditionally. I personally think the current `&mut Vec` version is simpler than callback, though.

lib/src/git_subprocess.rs
```diff
@@ -497,27 +504,73 @@ fn parse_git_push_output(
...
+    buf_size: usize,
+    mut sideband_cb: impl FnMut(&[u8]),
+) -> io::Result<Vec<u8>> {
+    let mut emit_sideband_data = |data: &[u8]| {
+        let lines = data.lines_with_terminator();
+        for line in lines.filter_map(|line| line.strip_prefix(b"remote: ")) {
+            sideband_cb(line);
+        }
+    };
+    let mut buf = vec![b'0'; buf_size];
```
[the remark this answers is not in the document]
yuja: Good catch. It was definitely wrong.

lib/src/git_subprocess.rs
bsdinis: sounds good to me. again, this is more for the case of having to parse out the progress info, but that doesn't have to be done in this PR

bsdinis (approved):

Issue #5516, opened 2025-01-30
Comma in a filename causes an error


mrzv:
## Description

Putting a comma in a filename produces an error when it's parsed as a fileset.

## Steps to Reproduce the Problem

1. `touch a,b`
1. `jj file untrack a,b`

## Expected Behavior

Untracking `a,b`

## Actual Behavior

```
Error: Failed to parse fileset: Syntax error
Caused by:  --> 1:2
  |
1 | a,b
  |  ^---
  |
  = expected <EOI>, `:`, `|`, `&`, or `~`
Hint: See https://jj-vcs.github.io/jj/latest/filesets/ for filesets syntax, or for how to match file paths.
```

## Specifications

- Platform: macOS
- Version: 0.25.0

ilyagr:
Does quoting the file, e.g. `jj file untrack \"a,b\"` (in Unix syntax) work?

The linked docs say:

> File names passed to these commands [must be quoted](https://jj-vcs.github.io/jj/latest/templates/#string-literals) if they contain whitespace or meta characters. However, as a special case, quotes can be omitted if the expression has no operators nor function calls.

I wonder if there's a way we can make this more prominent.

yuja:
(Closing as a dup)

mrzv:
@ilyagr Yes, quoting works. It's just not clear what "special symbols" means (in the commit you just added). When I ran into this, my shell autocompletion escaped a bunch of other characters, so I thought it should work. The unintuitive part is that jj needs quotes on top of that.

ilyagr:
I replied in https://github.com/jj-vcs/jj/pull/5517#issuecomment-2623559580, in the hope of not multiplying the number of threads for this discussion too much.

martinvonz:
> my shell autocompletion escaped a bunch of other characters

I suspect that was just shell escaping. For example, `echo \&` can be used in at least Bash and Fish to print a '&' character (instead of backgrounding the task).

mrzv:
Oh, it definitely was. The part that confused me is that I thought special symbols in jj documentation were referring to the characters that shell needs to escape. But it's actually something on top of it.

PR #5517, opened 2025-01-30
fileset syntax error hint: explain quoting briefly

ilyagr:
This might help with #4509, #5516. I am not sure it will help, but seems worth an experiment.

Cc @yuja

### DIFF
--- a/cli/tests/test_global_opts.rs
+++ b/cli/tests/test_global_opts.rs
@@ -304,7 +304,7 @@ fn test_invalid_filesets_looking_like_filepaths() {
     let repo_path = test_env.env_root().join("repo");
 
     let stderr = test_env.jj_cmd_failure(&repo_path, &["file", "show", "abc~"]);
-    insta::assert_snapshot!(stderr, @r#"
+    insta::assert_snapshot!(stderr, @r"
     Error: Failed to parse fileset: Syntax error
     Caused by:  --> 1:5
       |
@@ -312,8 +312,8 @@ fn test_invalid_filesets_looking_like_filepaths() {
       |     ^---
       |
       = expected `~` or <primary>
-    Hint: See https://jj-vcs.github.io/jj/latest/filesets/ for filesets syntax, or for how to match file paths.
-    "#);
+    Hint: See https://jj-vcs.github.io/jj/latest/filesets/ for filesets syntax. Names with special symbols must be quoted.
+    ");
 
     test_env.add_config(r#"ui.allow-filesets=false"#);
     let stderr = test_env.jj_cmd_failure(&repo_path, &["file", "show", "abc~"]);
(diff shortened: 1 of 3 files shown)


ilyagr:
Quoting @mrzv :

> [@ilyagr](https://github.com/ilyagr) Yes, quoting works. It's just not clear what "special symbols" means (in the commit you just added). When I ran into this, my shell autocompletion escaped a bunch of other characters, so I thought it should work. The unintuitive part is that jj needs quotes on top of that.

## My reply

> It's just not clear what "special symbols" means (in the commit you just added).

My hope was that the hint would be enough to get people to look into the docs. I would appreciate suggestions. (Or, perhaps, improving the hint is not the way to go).

> The unintuitive part is that jj needs quotes on top of that.

I agree this is unintuitive. We could explicitly say that shell quoting is not enough in the filesets docs, but that wouldn't help if people don't look there.

Did you look there and find it unhelpful or did you never look there, BTW?

ilyagr:
Or perhaps it might be OK to have a significantly longer hint?

mrzv:
@ilyagr I looked at the docs before filing an issue, but only skimmed through them, and so decided that only `~`, `&`, and `|` were special characters. That's on me.

Is putting filesets behind a flag and defaulting to just filenames off the table? I'd personally find that most intuitive.

ilyagr:
> looked at the docs before filing an issue, but only skimmed through them, and so decided that only `~`, `&`, and `|` were special characters. That's on me

That also means there is room for improvements in the docs, if we can figure out how to make it clearer.

Also, the error, since mentioning those symbols in that context was clearly not helpful. (Though if you *meant* to be writing a fileset and got this error, seeing what the syntax expected would be more helpful).

> Is putting filesets behind a flag and defaulting to just filenames off the table?

I'm not sure a flag (in the sense of `--use-filesets`) feels like a great solution. But perhaps we could make only `:` a special symbol and assume that something is a literal path unless it starts with `exact:`, `prefix:`, or something (or `set:`).

Additionally, there are ideas that seem helpful to me in the other thread, https://github.com/jj-vcs/jj/issues/4509#issuecomment-2398938140 and https://github.com/jj-vcs/jj/issues/4509#issuecomment-2398952703.

cli/tests/test_global_opts.rs
```diff
@@ -304,16 +304,16 @@ fn test_invalid_filesets_looking_like_filepaths() {
...
     Error: Failed to parse fileset: Syntax error
     Caused by:  --> 1:5
       |
     1 | abc~
       |     ^---
       |
       = expected `~` or <primary>
-    Hint: See https://jj-vcs.github.io/jj/latest/filesets/ for filesets syntax, or for how to match file paths.
-    "#);
+    Hint: See https://jj-vcs.github.io/jj/latest/filesets/ for filesets syntax. Names with special symbols must be quoted.
```
yuja, line 315: I don't think this change will help because the user might not know why literal path doesn't work. If they knew, they can easily parse the error message.

Maybe we can instead print an escaped version of user-provided literal path. Since quoted syntax may be shell-dependent, it would be easier if we had `literal:` prefix or something.
https://github.com/jj-vcs/jj/issues/4509#issuecomment-2465963426

ilyagr:
I'll close the PR, we can continue the discussion in #4509.

PR #5519, opened 2025-01-30
git: show fetch progress with `git.subprocess = true`

bsdinis:
# Checklist

If applicable:

### DIFF
--- a/lib/src/git_subprocess.rs
+++ b/lib/src/git_subprocess.rs
@@ -27,6 +27,7 @@ use std::thread;
 use bstr::ByteSlice;
 use thiserror::Error;
 
+use crate::git::Progress;
 use crate::git::RefSpec;
 use crate::git::RefToPush;
 use crate::git::RemoteCallbacks;
@@ -123,10 +124,11 @@ impl<'a> GitSubprocessContext<'a> {
         }
         let mut command = self.create_command();
         command.stdout(Stdio::piped());
-        command
-            .arg("fetch")
-            // attempt to prune stale refs
-            .arg("--prune");
+        // attempt to prune stale refs with --prune
+        command.args(["fetch", "--prune"]);
+        if callbacks.progress.is_some() {
+            command.arg("--progress");
+        }
         if let Some(d) = depth {
             command.arg(format!("--depth={d}"));
         }
@@ -523,12 +525,40 @@ fn wait_with_progress(
     })
 }
 
+#[derive(Default)]
+struct GitProgress {
+    // (frac, total)
+    deltas: (u64, u64),
+    objects: (u64, u64),
+    counted_objects: (u64, u64),
+    compressed_objects: (u64, u64),
+}
+
+impl GitProgress {
+    fn to_progress(&self) -> Progress {
+        Progress {
+            bytes_downloaded: None,
+            overall: self.fraction() as f32 / self.total() as f32,
+        }
+    }
+
+    fn fraction(&self) -> u64 {
+        self.objects.0 + self.deltas.0 + self.counted_objects.0 + self.compressed_objects.0
+    }
+
+    fn total(&self) -> u64 {
+        self.objects.1 + self.deltas.1 + self.counted_objects.1 + self.compressed_objects.1
+    }
+}
+
 fn read_to_end_with_progress<R: Read>(
     src: R,
     callbacks: &mut RemoteCallbacks<'_>,
 ) -> io::Result<Vec<u8>> {
     let mut reader = BufReader::new(src);
     let mut data = Vec::new();
+    let mut git_progress = GitProgress::default();
+
     loop {
         // progress sent through sideband channel may be terminated by \r
         let start = data.len();
@@ -537,7 +567,25 @@ fn read_to_end_with_progress<R: Read>(
         if line.is_empty() {
             break;
         }
-        if let Some(message) = line.strip_prefix(b"remote: ") {
+
+        if update_progress(line, &mut git_progress.objects, b"Receiving objects:")
+            || update_progress(line, &mut git_progress.deltas, b"Resolving deltas:")
+            || update_progress(
+                line,
+                &mut git_progress.counted_objects,
+                b"remote: Counting objects:",
+            )
+            || update_progress(
+                line,
+                &mut git_progress.compressed_objects,
+                b"remote: Compressing objects:",
+            )
+        {
+            if let Some(cb) = callbacks.progress.as_mut() {
+                cb(&git_progress.to_progress());
+            }
+            data.truncate(start);
+        } else if let Some(message) = line.strip_prefix(b"remote: ") {
             if let Some(cb) = callbacks.sideband_progress.as_mut() {
                 cb(message);
             }
@@ -545,6 +593,18 @@ fn read_to_end_with_progress<R: Read>(
         }
     }
     Ok(data)
+}
+
+fn update_progress(line: &[u8], progress: &mut (u64, u64), prefix: &[u8]) -> bool {
+    if let Some(line) = line.strip_prefix(prefix) {
+        if let Some((frac, total)) = read_progress_line(line) {
+            *progress = (frac, total);
+        }
+
+        true
+    } else {
+        false
+    }
 }
 
 fn read_until_cr_or_lf<R: io::BufRead + ?Sized>(
@@ -561,12 +621,30 @@ fn read_until_cr_or_lf<R: io::BufRead + ?Sized>(
             Some(i) => (i + 1, true),
             None => (data.len(), false),
         };
+
         dest_buf.extend_from_slice(&data[..n]);
         reader.consume(n);
+
         if found || n == 0 {
             return Ok(());
         }
     }
+}
+
+/// Read progress lines of the form: `<text> (<frac>/<total>)`
+/// Ensures that frac < total
+fn read_progress_line(line: &[u8]) -> Option<(u64, u64)> {
+    // isolate the part between parenthesis
+    let (_prefix, suffix) = line.split_once_str("(")?;
+    let (fraction, _suffix) = suffix.split_once_str(")")?;
+
+    // split over the '/'
+    let (frac_str, total_str) = fraction.split_once_str("/")?;
+
+    // parse to integers
+    let frac = frac_str.to_str().ok()?.parse().ok()?;
+    let total = total_str.to_str().ok()?.parse().ok()?;
+    (frac <= total).then_some((frac, total))
 }
 
 #[cfg(test)]
@@ -699,6 +777,7 @@ Done";
             blah blah
             remote: line2.0\rremote: line2.1
             remote: line3
+            Resolving deltas: (12/42)
             some error message
         "};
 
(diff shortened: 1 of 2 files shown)


lib/src/git_subprocess.rs
```diff
@@ -123,10 +124,7 @@ impl<'a> GitSubprocessContext<'a> {
         }
         let mut command = self.create_command();
         command.stdout(Stdio::piped());
-        command
-            .arg("fetch")
-            // attempt to prune stale refs
-            .arg("--prune");
+        command.args(["fetch", "--prune", "--progress"]);
```
yuja, line 127: `--progress` will have to be enabled conditionally.
(see https://github.com/jj-vcs/jj/pull/5521)

lib/src/git_subprocess.rs
```diff
@@ -561,14 +600,26 @@ fn read_until_cr_or_lf<R: io::BufRead + ?Sized>(
...
         };
-        dest_buf.extend_from_slice(&data[..n]);
-        reader.consume(n);
-        if found || n == 0 {
-            return Ok(());
+
+        if found || n == last_n {
+            let result = handler(&data[..n]);
+            reader.consume(n);
+            return Ok(result);
```
yuja, line 607: Since `.fill_buf()` never appends new data to non-empty buffer, and `BufReader`'s buffer size is capped, this doesn't work if `data` is partially received. (You can test that with smaller capacity.)

The buffer content has to be copied and consumed within this iteration. I think the original code is pretty okay since we need a line/content buffer anyway?

bsdinis: I see. will do

lib/src/git_subprocess.rs
bsdinis: wouldn't it be discarded when the progress callback doesn't exist?

lib/src/git_subprocess.rs
bsdinis: done

lib/src/git_subprocess.rs
bsdinis: done

bsdinis:
@yuja there is some weird interplay here with #5521 

Basically, in the case that the remote sends a message while the progress bar is updating, the progress bar will be split and not disappear, as it should.

In particular, there is one message, like this:
```
remote: Total 72578 (delta 173), reused 124 (delta 79), pack-reused 72267 (from 2)
```

bsdinis:
How it looked before:
```
Fetching into new repo in "/tmp/tmp.usqsXuL1dA/jj"
remote: Enumerating objects: 56033, done.
remote: Counting objects: 100% (7/7), done.
remote: Compressing objects: 100% (4/4), done.
 98% [████████████████████████████████████████████████████████████████████████████████████████████████████████████▊  ]remote: Total 56033 (delta 5), reused 3 (delta 3), pack-reused 56026 (from 1)
```

How it looks now (bar disappears when it's complete, as it should):
```
remote: Enumerating objects: 56034, done.
remote: Total 56034 (delta 6), reused 3 (delta 3), pack-reused 56026 (from 1)
```

yuja:
> Basically, in the case that the remote sends a message while the progress bar is updating, the progress bar will be split and not disappear, as it should.

Doesn't it an issue of the frontend? I think the UI will have to hide progress bar when sideband message is received, or something like that.

bsdinis:
> > Basically, in the case that the remote sends a message while the progress bar is updating, the progress bar will be split and not disappear, as it should.
> 
> Doesn't it an issue of the frontend? I think the UI will have to hide progress bar when sideband message is received, or something like that.

I mean, it would require a more involved frontend here. The progress bar is hidden in the end by issuing a `\r`. 
I tried looking for a clean solution there but couldn't find one. To be fair, there is a single message here. 
Happy to leave a comment about it, but unless I'm missing something really obvious we could use, it seems like that should be its own PR

yuja:
> > Doesn't it an issue of the frontend? I think the UI will have to hide progress bar when sideband message is received, or something like that.
> 
> I mean, it would require a more involved frontend here. The progress bar is hidden in the end by issuing a `\r`. I tried looking for a clean solution there but couldn't find one.

Yeah, the frontend will have to do something based on both prgoress and sideband messages, but I think that's the correct way to fix the problem. It's weird that two separate machinery is writing to the same output stream.

> it seems like that should be its own PR

Yes. If the current state is annoying, I would revert my patch.

bsdinis:
What do you think of the solution here of holding the sideband messages while progress is going on? It seems to be the best compromise (send sideband immediately when available, hold them for later when not).

Agree that it's not the cleanest, but sounds like it wouldn't be to bad.

yuja:
> What do you think of the solution here of holding the sideband messages while progress is going on? It seems to be the best compromise (send sideband immediately when available, hold them for later when not).

I suspect that the git2 impl would be already broken (if progress and sideband messages are processed in the same way as `git` command.) I haven't test that.

bsdinis:
```
% ~/dev/jj/target/debug/jj --config 'git.subprocess=false' git clone https://github.com/jj-vcs/jj
Fetching into new repo in "/tmp/tmp.xjeT5o8aH9/jj"
remote: Enumerating objects: 72578, done.
remote: Counting objects: 100% (315/315), done.
remote: Compressing objects: 100% (227/227), done.
 22%  35.3 MiB at  11.9 MiB/s [████████████████████████████████████████████▉                                                                                                                                                                 ]remote: Total 72578 (delta 176), reused 135 (delta 86), pack-reused 72263 (from 2)
 31% [████████████████████████████████████████████████████████████████████████▏                                                                                                                                                              ]
```

It is

yuja:
> I suspect that the git2 impl would be already broken (if progress and sideband messages are processed in the same way as `git` command.) I haven't test that.

It is. I'll take a look later.

bsdinis:
Now that I remember, the sideband callback for git fetch operations were always set to none.

yuja:
> I suspect that the git2 impl would be already broken

https://github.com/jj-vcs/jj/pull/5530

bsdinis:
Tested both this PR and `git2` on top of #5530 .
Works like a charm.
Thanks

lib/src/git_subprocess.rs
```diff
@@ -537,7 +566,25 @@ fn read_to_end_with_progress<R: Read>(
         if line.is_empty() {
             break;
         }
-        if let Some(message) = line.strip_prefix(b"remote: ") {
+
+        if update_progress(line, &mut git_progress.objects, "Receiving objects:")
+            || update_progress(line, &mut git_progress.deltas, "Receiving objects:")
```
steadmon, line 571: Is this a copy/paste error? Shouldn't it be "Resolving deltas:"?

bsdinis: Correct. Fixed

lib/src/git_subprocess.rs
```diff
@@ -123,10 +124,10 @@ impl<'a> GitSubprocessContext<'a> {
         }
         let mut command = self.create_command();
         command.stdout(Stdio::piped());
-        command
-            .arg("fetch")
-            // attempt to prune stale refs
```
yuja, line 128: nit: I think this comment is still valid. revert the change?

lib/src/git_subprocess.rs
```diff
@@ -537,7 +566,25 @@ fn read_to_end_with_progress<R: Read>(
...
+            || update_progress(
+                line,
+                &mut git_progress.counted_objects,
+                "remote: Counting objects:",
+            )
+            || update_progress(
+                line,
+                &mut git_progress.compressed_objects,
+                "remote: Compressing objects:",
+            )
```
yuja, line 581: nit: do we really need to merge remote progress into local ones? I feel the current local progress is too obscure, so it's better to print what's going on in the remote.

lib/src/git_subprocess.rs
```diff
@@ -547,6 +594,18 @@ fn read_to_end_with_progress<R: Read>(
     Ok(data)
 }
 
+fn update_progress(line: &[u8], progress: &mut (u64, u64), prefix: &str) -> bool {
+    if line.starts_with_str(prefix) {
```
yuja, line 598: nit: `.strip_prefix(b"...")` is slightly better (than rescanning uninteresting part)

lib/src/git_subprocess.rs
```diff
@@ -561,14 +620,27 @@ fn read_until_cr_or_lf<R: io::BufRead + ?Sized>(
...
 
+/// Read progress lines of the form: `<text> (<frac>/<total>)`
+/// Ensures that frac < total
+fn read_progress_line(line: &[u8]) -> Option<(u64, u64)> {
+    let (_prefix, suffix) = line.split_once_str("(")?;
+    let (fraction, _suffix) = suffix.split_once_str(")")?;
+    let (frac_str, total_str) = fraction.split_once_str("/")?;
+    let frac = frac_str.to_str().ok()?.parse().ok()?;
+    let total = total_str.to_str().ok()?.parse().ok()?;
+    (frac <= total).then_some((frac, total))
```
yuja, line 641: nit: if you like, maybe we can use regex. I don't have strong feeling which would be readable overall.

lib/src/git_subprocess.rs
```diff
@@ -717,4 +789,22 @@ Done";
...
+            read_progress_line(b"Resolving deltas: (0/1000)\r"),
+            Some((0, 1000))
+        );
+        assert_eq!(read_progress_line(b"Receiving objects: (420/100)\r"), None);
+        assert_eq!(
+            read_progress_line(b"remote: this is something else\n"),
+            None
+        );
+        assert_eq!(read_progress_line(b"fatal: this is a git error\n"), None);
+    }
```
yuja, line 809: nit: can you insert some progress line to `test_read_to_end_with_progress()` as well?

yuja (approved):
Thanks.

lib/src/git_subprocess.rs
bsdinis: So the reasoning here was as follows:
  - the main thing about the progress bar is showing the user progress is happening.
  - without the remote lines, the progress bar was very very jumpy, seems prefereable to show it all being progress. 
  - long term, we might want to have a bar per item (i.e., a bar for received objects, another for resolved deltas, etc.)

lib/src/git_subprocess.rs
bsdinis: done

lib/src/git_subprocess.rs
bsdinis: i don't think this simple use case warrants regex

lib/src/git_subprocess.rs
bsdinis: done

lib/src/git_subprocess.rs
bsdinis: done

lib/src/git_subprocess.rs
yuja: I don't feel it's jumpy, but I might have lower expectation about console UI.

> we might want to have a bar per item (i.e., a bar for received objects, another for resolved deltas, etc.)

Yeah, I think we can revisit this when we add that. It's weird that all progress items of different units/weights are packed into one. It's also odd that some remote progress items are discarded when it's done, but some others don't.

bsdinis: yup. I like how git-branchless seems to do it, but i guess that'd be an extra dependency

yuja: Oh, I meant to remove `remote: <..>` progress interception until we add proper progress support. Sorry I didn't clearly say that.

Issue #5374, opened 2025-01-15
Warn/require confirmation on bookmarking empty changes?


drieber:
### Discussed in https://github.com/jj-vcs/jj/discussions/5363

<div type='discussions-op-text'>

<sup>Originally posted by **sourcefrog** January 14, 2025</sup>
In getting to know jj, when I am working on a feature branch that I want to push to github, I've fairly often run a flow like

    jj commit -m Whatever
    jj bookmark move featurebranch

The trap of course is that the bookmark is set to the new empty revision following the thing that I committed. I suppose pushing the empty rev to a feature branch on github is not too harmful but it does seem a bit messy. 

The right answer I guess is that I should either

1. Move the bookmark _before_ committing, which seems a bit weird coming from other VCS
2. `bookmark move -r@-`
3. Just `jj desc` then bookmark it but then don't forget to `jj new`

I see there is a good bit of existing discussion about various types of auto-moving bookmarks, which would perhaps be a more fundamental answer for this kind of workflow:

 * https://github.com/jj-vcs/jj/discussions/3549 
 * https://github.com/jj-vcs/jj/discussions/2425
 * https://github.com/jj-vcs/jj/discussions/5333
 * https://github.com/jj-vcs/jj/issues/3402
 * https://github.com/jj-vcs/jj/issues/2338

As a more concrete and tactical (perhaps excessively tactical) fix I wonder if the UI should by default prevent moving a bookmark to an empty change, the same way it by default prevents you pushing new bookmarks and moving bookmarks backwards.

That is, `jj new; jj bookmark set bbb` would fail unless you said `--allow-empty`. 

Alternatively, rather than failing it could say 

    warning: bookmark bbb is set to an empty revision
    hint: to set it to the parent revision use `jj bookmark move bbb -rrstuvwxyz --allow-backwards`

</div>

martinvonz:
I agree with @yuja's suggestion [here](https://github.com/jj-vcs/jj/discussions/5363#discussioncomment-11839865), i.e. that we should just make `jj bookmark create/set/move` require a revision argument instead of defaulting to `@`.

drieber:
> I agree with [@yuja](https://github.com/yuja)'s suggestion [here](https://github.com/jj-vcs/jj/discussions/5363#discussioncomment-11839865), i.e. that we should just make `jj bookmark create/set/move` require a revision argument instead of defaulting to `@`.

That sounds good to me. There are two practical things to consider:
[...]

PR #5527, opened 2025-01-30
bookmarks: Make target revision a required argument to create/move/set

drieber:
This change is not backwards compatible, but it should be very easy to adapt to it.

This fixes
https://github.com/jj-vcs/jj/issues/5374
See also
https://github.com/jj-vcs/jj/discussions/5363

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Deprecations
 
+* This release takes the first steps to make target revision required in
+  `bookmark create`, `bookmark move` and `bookmark set`. Those commands will display
+  a warning if the user does not specify target revision  explicitly. In the near
+  future those commands will fail if target revision is not specified.
+
 ### New features
 
 * `jj undo` now shows a hint when undoing an undo operation that the user may
--- a/cli/src/commands/bookmark/create.rs
+++ b/cli/src/commands/bookmark/create.rs
@@ -28,10 +28,13 @@ use crate::ui::Ui;
 /// Create a new bookmark
 #[derive(clap::Args, Clone, Debug)]
 pub struct BookmarkCreateArgs {
+    // TODO(#5374): Make required in jj 0.32+
     /// The bookmark's target revision
     //
     // The `--to` alias exists for making it easier for the user to switch
-    // between `bookmark create`, `bookmark move`, and `bookmark set`.
+    // between `bookmark create`, `bookmark move`, and `bookmark set`. Currently target revision
+    // defaults to the working copy if not specified, but in the near future it will be required to
+    // explicitly specify it.
     #[arg(
         long, short,
         visible_alias = "to",
@@ -51,6 +54,13 @@ pub fn cmd_bookmark_create(
     args: &BookmarkCreateArgs,
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
+    if args.revision.is_none() {
+        writeln!(
+            ui.warning_default(),
+            "Target revision was not specified, defaulting to the working copy (-r@). In the near \
+             future it will be required to explicitly specify target revision."
+        )?;
+    }
     let target_commit = workspace_command
         .resolve_single_rev(ui, args.revision.as_ref().unwrap_or(&RevisionArg::AT))?;
     let view = workspace_command.repo().view();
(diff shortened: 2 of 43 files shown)


yuja (commented):
Thanks.

cli/tests/test_git_fetch.rs~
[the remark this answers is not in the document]
drieber: Sorry about that. Done.

arxanas:
Should we make this a soft deprecation in the next version and a hard error in the subsequent version? It seems like it might break real interactive and non-interactive workflows in a frustrating way. It looks like the hard work of migrating all the tests is already done, so hopefully it's a small additional change?

PhilipMetzger:
> Should we make this a soft deprecation in the next version and a hard error in the subsequent version? It seems like it might break real interactive and non-interactive workflows in a frustrating way

yes, that should be done and actually deprecated after 6 releases, like the rest of the user-visible command changes.

drieber:
> > Should we make this a soft deprecation in the next version and a hard error in the subsequent version? It seems like it might break real interactive and non-interactive workflows in a frustrating way
> 
> yes, that should be done and actually deprecated after 6 releases, like the rest of the user-visible command changes.

drieber:
Sorry, I accidentally closed this PR. Reopening....

cli/src/commands/bookmark/create.rs
```diff
@@ -51,6 +53,13 @@ pub fn cmd_bookmark_create(
     args: &BookmarkCreateArgs,
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
+    if args.revision.is_none() {
+        writeln!(
+            ui.warning_default(),
+            "Target revision was not specified, defaulting to the working copy (-r@). In the near \
+             future it will be required to explicitly specify target revision."
```
PhilipMetzger, line 60: Nit: use the same message everywhere

cli/src/commands/bookmark/create.rs
```diff
@@ -51,6 +53,13 @@ pub fn cmd_bookmark_create(
     args: &BookmarkCreateArgs,
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
+    if args.revision.is_none() {
```
PhilipMetzger, line 56: nit: missing TODO to make it required in jj 0.32+

PhilipMetzger (approved):
last minor nits from me

cli/src/commands/bookmark/create.rs
drieber: Done

CHANGELOG.md
```diff
@@ -51,6 +51,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj log` can now show cryptographic commit signatures. The output can be
   controlled by the `ui.show-cryptographic-signatures=true` config knob.
 
+* This release takes the first steps make target revision a required argument of
+  `bookmark create`, `bookmark move` and `bookmark set`. Those commands will
+  display a warning if the user does not specify target revision  explicitly. In
+  the near future those commands will fail if target revision is not specified.
+
```
martinvonz, line 58: This should be moved to "[Unreleased]" above. I think "Deprecations" is the right place for it.

cli/src/commands/bookmark/move.rs
```diff
@@ -56,14 +56,14 @@ pub struct BookmarkMoveArgs {
 
     /// Move bookmarks to this revision
     // We intentionally do not support the short `-t` for `--to` since we don't
-    // support `-f` for `--from`.
+    // support `-f` for `--from`. Currently this defaults to the working copy, but in the near
+    // future it will be required to explicitly specify it.
```
martinvonz, line 60: nit: Reflow text. I think if you simply join the lines, then `rustfmt` will take for wrapping them correctly.

cli/src/commands/bookmark/create.rs
```diff
@@ -51,6 +53,14 @@ pub fn cmd_bookmark_create(
     args: &BookmarkCreateArgs,
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
+    // TODO: Make args.revision required in jj 0.32+. See https://github.com/jj-vcs/jj/issues/5374.
```
PhilipMetzger, line 56: Sorry, I miss-placed my nit before but could you move the comment below the doc comment for the specific arg. Thanks.

cli/src/commands/bookmark/move.rs
drieber: Done, I think. Neither `jj fix` nor `cargo +nightly fmt` wrapped the text, so I wrapped it manually. I did not try invoking `rustfmt` directly though. Should I do that instead?

CHANGELOG.md
drieber: Done. I also tweaked the wording.

cli/src/commands/bookmark/create.rs
drieber: Done.

PR #5528, opened 2025-01-31
cli: add error stub for `jj util mangen`

### LINKED ISSUE #5525: `jj util mangen` will be broken for packagers in v0.26
For package maintainers, here's an example of how to adjust to the new interface, since it no longer writes to stdout:

https://github.com/jj-vcs/jj/pull/5258/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0

---

> Not sure if this is the right place to highlight it, but this broke `--HEAD` builds in Homebrew:
> 
> * https://github.com/Homebrew/homebrew-core/blob/488e84fdb8e9ea309dff4b507e170edd6e92f38e/Formula/j/jj.rb#L46
>
> ```console
> $ brew install --HEAD jj
> ==> Fetching jj
> ==> Downloading https://raw.githubusercontent.com/Homebrew/homebrew-core/2c92fccaa0f234d8d34a2a2e68c12d438b4c42b7/Formula/j/jj.rb
> Already downloaded: /Users/zx8/Library/Caches/Homebrew/downloads/629dbff562ae32e7c6f7ef688c739356c1488fc61a597432e69c8b194e268be7--jj.rb
> ==> Downloading https://github.com/jj-vcs/jj/commit/b4f936ac302ee835aa274e4dd186b436781d5d2f.patch?full_index=1
> Already downloaded: /Users/zx8/Library/Caches/Homebrew/downloads/4f0b715f9452bf91255eb771e1f8462f949705770cd700fb93f6b4fdbac4dd65--b4f936ac302ee835aa274e4dd186b436781d5d2f.patch
> ==> Cloning https://github.com/martinvonz/jj.git
> Updating /Users/zx8/Library/Caches/Homebrew/jj--git
> ==> Checking out branch main
> Already on 'main'
> Your branch is up to date with 'origin/main'.
> HEAD is now at d3f56c0ac git: forward "remote: " lines when they arrive
> ==> Patching
> ==> Applying b4f936ac302ee835aa274e4dd186b436781d5d2f.patch
> ==> cargo install --path=cli
> Error: Failure while executing; `/opt/homebrew/Cellar/jj/HEAD-d3f56c0_2/bin/jj util mangen` exited with 2.
> ```

— @zx8 in https://github.com/jj-vcs/jj/pull/5258#issuecomment-2624693830

> [...] every package maintainer who currently runs `jj util mangen` will try to upgrade jj, see that their packaging no longer works, and will need to debug.
> - Currently, I guess their best option is to trawl the changelog, or even bisect.
> - Unlike most commands: I suspect that deprecating the command probably won't help, because the deprecation notice will probably be silently swallowed? This is in contrast to commands for interactive human use.
> - So the best option to inform package maintainers is probably to make the command produce a hard error, but with a helpful message to tell them how to fix it?
> 
> We should see if we can do this before the next release to make it easier for package maintainers to upgrade.

— @arxanas in https://github.com/jj-vcs/jj/issues/5258#issuecomment-2625134509

arxanas:
Closes #5525.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/util/mod.rs
+++ b/cli/src/commands/util/mod.rs
@@ -35,6 +35,7 @@ use self::install_man_pages::UtilInstallManPagesArgs;
 use self::markdown_help::cmd_util_markdown_help;
 use self::markdown_help::UtilMarkdownHelp;
 use crate::cli_util::CommandHelper;
+use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
@@ -45,6 +46,9 @@ pub(crate) enum UtilCommand {
     ConfigSchema(UtilConfigSchemaArgs),
     Exec(UtilExecArgs),
     Gc(UtilGcArgs),
+    /// No longer supported; use `jj util install-man-pages` instead. This stub
+    /// will be removed in a future version.
+    Mangen,
     InstallManPages(UtilInstallManPagesArgs),
     MarkdownHelp(UtilMarkdownHelp),
 }
@@ -60,6 +64,7 @@ pub(crate) fn cmd_util(
         UtilCommand::ConfigSchema(args) => cmd_util_config_schema(ui, command, args),
         UtilCommand::Exec(args) => cmd_util_exec(ui, command, args),
         UtilCommand::Gc(args) => cmd_util_gc(ui, command, args),
+        UtilCommand::Mangen => Err(user_error("`jj util mangen` has been replaced by `jj util install-man-pages`; see https://github.com/jj-vcs/jj/issues/5525")),
         UtilCommand::InstallManPages(args) => cmd_util_install_man_pages(ui, command, args),
         UtilCommand::MarkdownHelp(args) => cmd_util_markdown_help(ui, command, args),
     }
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -95,6 +95,7 @@ This document contains the help content for the `jj` command-line program.
 * [`jj util config-schema`↴](#jj-util-config-schema)
 * [`jj util exec`↴](#jj-util-exec)
 * [`jj util gc`↴](#jj-util-gc)
+* [`jj util mangen`↴](#jj-util-mangen)
 * [`jj util install-man-pages`↴](#jj-util-install-man-pages)
 * [`jj util markdown-help`↴](#jj-util-markdown-help)
 * [`jj undo`↴](#jj-undo)
@@ -2182,6 +2183,7 @@ Infrequently used commands such as for generating shell completions
 * `config-schema` — Print the JSON schema for the jj TOML config format
 * `exec` — Execute an external command via jj
 * `gc` — Run backend-dependent garbage collection
+* `mangen` — No longer supported; use `jj util install-man-pages` instead. This stub will be removed in a future version
 * `install-man-pages` — Install Jujutsu's manpages to the provided path
 * `markdown-help` — Print the CLI help for all subcommands in Markdown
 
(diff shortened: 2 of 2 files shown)


PhilipMetzger (changes requested):
I think a full restore should be in state, since the actual PR didn't really follow the "normal" deprecation and removal procedure.

cli/src/commands/util/mod.rs
```diff
@@ -45,6 +46,9 @@ pub(crate) enum UtilCommand {
     ConfigSchema(UtilConfigSchemaArgs),
     Exec(UtilExecArgs),
     Gc(UtilGcArgs),
+    /// No longer supported; use `jj util install-man-pages` instead. This stub
+    /// will be removed in a future version.
+    Mangen,
```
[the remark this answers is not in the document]
arxanas: Thanks, will add (also I didn't know the next jj version 🤣).

arxanas:
> I think a full restore should be in state.

@PhilipMetzger What does "in state" mean? Does this mean the same as "in order", as in "warranted in this situation"?

> the actual PR didn't really follow the "normal" deprecation and removal procedure.

See the [commit message](https://github.com/jj-vcs/jj/pull/5528/commits/02ecc3ad7ab4ed2830fdabea6bb56ee72de42c8f) for why I think it's not useful to follow the normal deprecation feature. I was going to do that first, but

PhilipMetzger:
> > I think a full restore should be in state.
> 
> @PhilipMetzger What does "in state" mean? Does this mean the same as "in order", as in "warranted in this situation"?

Yes, that should've been a "a full restore in order".

> See the [commit message](https://github.com/jj-vcs/jj/pull/5528/commits/02ecc3ad7ab4ed2830fdabea6bb56ee72de42c8f) for why I think it's not useful to follow the normal deprecation feature. I was going to do that first, but I don't think it improves anything for package maintainers.
 
While it happened with someone who was a packager, we have no data on actual command utilization in the wild. So carving out such a exception for packagers seems really unwise. It also casually breaks our social contract.

emilazy:
This is a packaging command so I think applying norms for packagers makes sense? (Even if someone is “packaging by hand”, they’ll do it once per install, so I don’t think this could break someone’s day‐to‐day workflow.)

FWIW as a downstream packager of Jujutsu I would find a hard error for this much more useful than a restored deprecated command.

(Is the long deprecation period for every backwards‐incompatible change written down anywhere?)

PhilipMetzger (approved):

cli/src/commands/util/mod.rs
```diff
@@ -45,6 +46,9 @@ pub(crate) enum UtilCommand {
     ConfigSchema(UtilConfigSchemaArgs),
     Exec(UtilExecArgs),
     Gc(UtilGcArgs),
+    /// No longer supported; use `jj util install-man-pages` instead. This stub
```
ilyagr, line 49: I feel like this command should be hidden in the Clap sense. This should also remove it from the docs. TODO: Find the right syntax.

ilyagr (approved):
Thank you!

cli/src/commands/util/mod.rs
ilyagr: I think addign `#[command(hide = true)]` here will do it.

arxanas: Yes, good point, thanks for raising.

arxanas:
Apologies: This should have gone out in the most recent release, but I didn't realize that this item was pending when we were talking about doing the release. Since we've already done the release and packagers have had to adapt, I think this doesn't warrant merging anymore.

As per https://github.com/jj-vcs/jj/pull/5258#issuecomment-2639784685, Homebrew is working again in https://github.com/Homebrew/homebrew-core/pull/206710/files#diff-94b0b9da218994a111c379a601b2e2ac1a84ce60db8159d564d0b76d315a2838R43, so at least one packager has successfully updated. (The code for installing man-pages also becomes simpler on their end.)

Issue #3594, opened 2024-04-29
FR: simplify overriding of default colors


noahmayr:
**Is your feature request related to a problem? Please describe.**
When wanting to override a color from `cli/src/config/colors.toml`, e.g. setting the description color to blue whenever there is no description (`description placeholder`) regardless of whether the commit is the working copy or if it's empty you would have to add the following to your config:
```toml
[colors]
"description placeholder" = "blue"
"empty description placeholder" = "blue"
"working_copy description placeholder" = "blue"
"working_copy empty description placeholder" = "blue"
```

**Describe the solution you'd like**
Since I've heavily customized my log colors, something like `ui.use_builtin_colors = false` would work very well for me.

**Describe alternatives you've considered**
Alternatively a flag like `!important` from CSS that allows overriding the default style from the example above with just the following would be ok as well:
```toml
[colors]
"description placeholder" = { fg ="blue", override = true }
```

**Additional context**

yuja:
> ui.use_builtin_colors = false

It could be `ui.color-theme = <theme-name>` (and `[color-themes.default]`, etc.), but I don't know if we'll need this kind of indirection.

noahmayr:
That's interesting, it could actually allow not only themes with different colors for the given section but even shipping 2-3 default themes which each highlight the information provided differently.
1. Default (basically what we have now)
2. Minimal (same as default with only the most important fields colorized and less important fields dimmed)
3. Contextual (something similar to what I'm using, which colorizes most fields of a change in the same color and gives a unique color to working copy changes, immutable changes, changes with conflicts, etc.)

![image](https://github.com/martinvonz/jj/assets/1039040/17104db9-070b-4270-a88d-c977130f950b)

jlombera:
I'm also interested in an easy way to ignore all the default colors and just set ones I want to afterwards.  In my case, besides the diff hunks coloring, I don't want any colors at all, I just want to use bold and underline (I wish italics was supported as well) for formatting.  Right now my my `[colors]` configs section is large consisting of mostly overriding colors do `"default"`.

yuja:
> Right now my [colors] config section is large consisting of mostly overriding colors do "default".

fwiw, I'm going to replace the config backend to `toml_edit`. If the migration gets done, an inline table `key = { bg = .. }` will overwrite the entire `key` table, so `{ bg = .., color = "default" }` won't be needed in order to remove the color field from the key. We'll still need `key = {}` or `key = "default"` to override the existing style, though.

[...]

PR #5538, opened 2025-01-31
cli: ability to disable default colorscheme

dpc:
My motivation is mostly to address #5098 , but in a lot of software there's a flag to disable sourcing default configs, which is useful in a lot contexts.

A more granular (or just atlernative/better) approaches would require more design and debate, while this seems simple and immediately useful (at least to some).

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* New `JJ_NO_DEFAULT_COLORS` environment variable disables setting built-in default colors.
+
 * `jj show` now supports `--reversed` flag.
 
 * `jj` now looks for config files in `/etc/jj`.
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -768,14 +768,24 @@ pub fn default_config_layers() -> Vec<ConfigLayer> {
     // Syntax error in default config isn't a user error. That's why defaults are
     // loaded by separate builder.
     let parse = |text: &'static str| ConfigLayer::parse(ConfigSource::Default, text).unwrap();
-    let mut layers = vec![
-        parse(include_str!("config/colors.toml")),
-        parse(include_str!("config/hints.toml")),
-        parse(include_str!("config/merge_tools.toml")),
-        parse(include_str!("config/misc.toml")),
-        parse(include_str!("config/revsets.toml")),
-        parse(include_str!("config/templates.toml")),
-    ];
+    let mut layers = {
+        let maybe_colors = if env::var("JJ_NO_DEFAULT_COLORS").is_ok() {
+            vec![]
+        } else {
+            vec![parse(include_str!("config/colors.toml"))]
+        };
+        [
+            maybe_colors,
+            vec![
+                parse(include_str!("config/hints.toml")),
+                parse(include_str!("config/merge_tools.toml")),
+                parse(include_str!("config/misc.toml")),
+                parse(include_str!("config/revsets.toml")),
+                parse(include_str!("config/templates.toml")),
+            ],
+        ]
+        .concat()
+    };
     if cfg!(unix) {
         layers.push(parse(include_str!("config/unix.toml")));
     }


martinvonz:
Thanks, but `JJ_CONFIG= jj` should already have this effect

dpc:
@martinvonz The `JJ_CONFIG` controls the user's config, no? While I'm trying to disable the default configs shipped with jj.

martinvonz:
Oh, I'm sorry. Interesting. We have a lot of code that depends on the default configs existing, so I think it's going to be more complicated to support than this PR's current state.

dpc:
@martinvonz I just noticed. Oops.


As a first user who'd want this, I don't mind just adding all the required keys to my config, and thus being sure I am in full control.

Also, I could add `unwrap_or_default()` on most obvious ones.

Also (or/and) I could add `JJ_CONFIG_DEFAULT_COLORS`.

dpc:
image

This seems to have no issues.

ilyagr:
For the colors issue, I wonder if it'd be better to introduce a notion of themes. So, the colors we have now would become the default theme, and you could reset them all by using another theme.

Settings like `templates.log_node` could also be set by a theme, though I don't think there's any reasonable "undefined" value that can have by default.

There are many other such settings; in my mind, any `jj` command is allowed to crash if the default config is not loaded. The first example that comes to mind is not merged yet, but https://github.com/jj-vcs/jj/pull/5415 would crash if you deleted the default config and tried using `:builtin` pager.

Recently, [scoped config](https://jj-vcs.github.io/jj/prerelease/config/#conditional-variables) was introduced. I'm not 100% sure about that direction, but perhaps it's sufficient to create a "default theme" scope that contains all the color config and is activated by default. (This would hopefully be relatively easy, but it's not trivial, it still needs some design work on how these theme scopes work)

jakobhellermann:
> Recently, [scoped config](https://jj-vcs.github.io/jj/prerelease/config/#conditional-variables) was introduced. I'm not 100% sure about that direction, but perhaps it's sufficient to create a "default theme" scope that contains all the color config and is activated by default. (This would hopefully be relatively easy, but it's not trivial, it still needs some design work on how these theme scopes work)

Instead of having a "theme scope", we could introduce a generic "config scope", that is enabled when a config variable is set to a certain value: 

```toml
[[--scope]]
--when.config = ["ui.enable-colors"] # equivalent to ["ui.enable-colors", "true"]
[--scope.colors]
"error" = { fg = "default", bold = true }
"error_source" = { fg = "default" }
...
```
or
```toml
[[--scope]]
--when.config = ["ui.theme", ":builtin"]
[--scope.colors]
"error" = { fg = "default", bold = true }
"error_source" = { fg = "default" }
...
```

yuja:
I wouldn't add any "conditional" thingy that also depends on config values (and/or any other dynamic variables.) That would make resolution complicated.

For color overrides, it's probably better to add some indirection dedicated for `[colors]` table. https://github.com/jj-vcs/jj/issues/3594

FWIW, the default colors table can be reset by abusing invalid value, but please don't expect this will work forever.
```toml
[[--scope]]
colors = 0  # reset colors table by assigning non-table value
[[--scope]]
colors.commit_id = "red"
```

ilyagr:
I agree that this request seems like it would be covered by a "theme"-like config that deals with the "colors" table only. It makes sense to me to consider this a dupe of #3594 .

At the same time, the contrarian in me started wondering about how general such a mechanism could be. I think we could have, at an extreme, a special table `[config-toggles]` that none of the `--scope` configs are allowed to modify. Then, I think it might be OK (though not necessarily a great idea) to allow scopes that depend on values inside the `[config-toggles]` table and are allowed to modify (almost?) anything other than the `config-toggles` table.

dpc:
So far I've disabled the colors and moved on. The problem with colors is that it's impossible to reset them after the fact due to way the selector priority works. Everything else seems easily overrideable, so not as important.

To me this is a stop-gap measure. Not sure how long will it take to debate on the more proper system. Might be worth to land a env var to disable just the colors to unblock people like me and focus on whatever more pressing things.

ilyagr:
My preference would be not to merge the stop-gap measure as is. It feels a bit hacky, and it feels like this problem would only be urgent for a few people. Are you unable to use a custom version of `jj` compiled with this patch? If you're willing to go this far to get `jj` customized just the way you want it, what's one more step?

I do think that this is a real issue that would be nice to solve, just not so urgent to be worth exposing everybody to a temporary stop-gap that will be very likely to lead to a breaking change later.

See also #5571 (though if we go with something like themes, it might not actually be that relevant).

yuja:
As a workaround, `colors` table can be emptied by setting non-table value. It's ugly, but works for now. Custom binary isn't needed.
https://github.com/jj-vcs/jj/pull/5538#issuecomment-2628894822

emilazy:
It might be nice for Jujutsu to respect [`$NO_COLOR`](https://no-color.org/).

ilyagr:
> It might be nice for Jujutsu to respect [`$NO_COLOR`](https://no-color.org/).

It does, though it might not be obvious due to

https://github.com/jj-vcs/jj/blob/0bcddff473698c8f6a080c3958398aaf7c490ebb/CHANGELOG.md?plain=1#L2641-L2642

(I am not sure if I agree with that design, it certainly makes NO_COLOR not very useful for me, but apparently that's how it's supposed to work. I've never relied on NO_COLOR, so I'm OK with it not being very useful to me.)

But I think the OP wants color, they just want color for exactly the things they specify and nothing else.

dpc:
> But I think the OP wants color, they just want color for exactly the things they specify and nothing else.

Exactly. I'm a very hard to please, opinionated and particular about everything. Just the worst kind of user really.

cli/src/config.rs
```diff
@@ -768,14 +768,24 @@ pub fn default_config_layers() -> Vec<ConfigLayer> {
...
+            vec![
+                parse(include_str!("config/hints.toml")),
+                parse(include_str!("config/merge_tools.toml")),
+                parse(include_str!("config/misc.toml")),
+                parse(include_str!("config/revsets.toml")),
+                parse(include_str!("config/templates.toml")),
+            ],
+        ]
+        .concat()
+    };
```
martinvonz, line 788: nit: since order shouldn't matter, conditionally add `colors.toml` below like the platform-specific ones instead

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* New `JJ_NO_DEFAULT_COLORS` environment variable disables setting built-in default colors.
+
```
martinvonz, line 32: If we're going to drop support for this environment variable once #3594 has been implemented (if we decide to do that), then perhaps it's better to not announce this feature now (it'll basically be a secret feature for you).

martinvonz (commented):
I'm not sure it's worth adding this before #3594.

dpc:
I'm not expecting this to be merged as is. It works fine for me as a fork, though I'm looking forward to not needing it eventually, hopefully.

PR #5542, opened 2025-01-31
templater: add pad_center template function

### LINKED ISSUE #5066: FR: Add `pad_center` template function

Similar to the other issue I recently raised, I'm currently working through the surface-level features of `jj` to see if I can bring things in line with my git setup.

It does not seem possible to align text to the center of a column in the log output at the moment. Git has support for this via:

```
%><( <N> ), %><|( <M> )
similar to %<( <N> ), %<|( <M> ) respectively, but padding both sides (i.e. the text is centered)
```

ref: https://git-scm.com/docs/pretty-formats

It would be nice to have a `pad_center` function, similar to `pad_start` and `pad_end` that middle-aligned text.

I'm not sure if there are any alternatives.

steadmon:
I started out trying to implement a `pad_between` function as suggested on the feature request, but got bogged down because all of the template wrapping expects a single content argument. If you think it's still worth pursuing that approach, I can spend Monday working on it, but I thought it might be more useful to get a simpler fix out faster.

Fixes #5066.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * New template function `config(name)` to access to configuration variable from
   template.
 
+* New template function `pad_centered()` to center content within a minimum width.
+
 * Templater now supports `list.filter(|x| ..)` method.
 
 * The `diff` commit template keyword now supports custom formatting via
--- a/cli/src/template_builder.rs
+++ b/cli/src/template_builder.rs
@@ -1476,6 +1476,22 @@ fn builtin_functions<'a, L: TemplateLanguage<'a> + ?Sized>() -> TemplateBuildFun
         Ok(L::wrap_template(template))
     });
     map.insert(
+        "pad_centered",
+        |language, diagnostics, build_ctx, function| {
+            let ([width_node, content_node], [fill_char_node]) =
+                function.expect_named_arguments(&["", "", "fill_char"])?;
+            let width = expect_usize_expression(language, diagnostics, build_ctx, width_node)?;
+            let content =
+                expect_template_expression(language, diagnostics, build_ctx, content_node)?;
+            let fill_char = fill_char_node
+                .map(|node| expect_template_expression(language, diagnostics, build_ctx, node))
+                .transpose()?;
+            let template =
+                new_pad_template(content, fill_char, width, text_util::write_padded_centered);
+            Ok(L::wrap_template(template))
+        },
+    );
+    map.insert(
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -100,7 +100,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Add a new template alias `builtin_op_log_oneline` along with `format_operation_oneline` and `format_snapshot_operation_oneline`
 
 * New template function `config(name)` to access to configuration variable from
-  template.
+  template, and `pad_center()` to center content within a minimum width.
```
yuja, line 103: nit: split to new bullet point since these features are unrelated?

cli/src/template_builder.rs
```diff
@@ -1475,6 +1475,22 @@ fn builtin_functions<'a, L: TemplateLanguage<'a> + ?Sized>() -> TemplateBuildFun
         let template = new_pad_template(content, fill_char, width, text_util::write_padded_end);
         Ok(L::wrap_template(template))
     });
+    map.insert(
+        "pad_center",
```
yuja, line 1479: nit: I'm not sure if "pad center" is correct term, but I don't think it's confusing either.

cli/src/text_util.rs
```diff
@@ -302,6 +302,28 @@ pub fn write_padded_end(
...
+    formatter: &mut dyn Formatter,
+    recorded_content: &FormatRecorder,
+    recorded_fill_char: &FormatRecorder,
+    min_width: usize,
+) -> io::Result<()> {
+    // We don't care about the width of non-UTF-8 bytes, but should not panic.
+    let width = String::from_utf8_lossy(recorded_content.data()).width();
+    let fill_width = min_width.saturating_sub(width);
+    let fill_left = fill_width / 2;
+    let fill_right = fill_left + (fill_width % 2);
```
yuja, line 320: nit: I think `fill_width - fill_left` is clearer

yuja (approved):
Looks good, thanks.

CHANGELOG.md
steadmon: Done.

cli/src/text_util.rs
steadmon: Done.

cli/src/template_builder.rs
steadmon: Switched to "pad_centered", does that sound better?

yuja: Sounds good, thanks.

PR #5543, opened 2025-01-31
docs: Update the contributing docs to mention the new `contributors` …

PhilipMetzger:
…team.

I hope this makes the procedure clearer to new contributors like jakobhellerman.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -58,7 +58,8 @@ commit. You can do that by creating a new commit on top of the initial commit
 `jj git push`
 will automatically force-push the bookmark.
 
-When your first PR has been approved, we typically give you contributor access,
+When your first PR has been approved, we typically invite you to the
+`jj-vcs/contributors` team to give you contributor access,
 so you can address any remaining minor comments and then merge the PR yourself
 when you're ready. If you realize that some comments require non-trivial
 changes, please ask your reviewer to take another look.


docs/contributing.md
```diff
@@ -58,7 +58,8 @@ commit. You can do that by creating a new commit on top of the initial commit
 `jj git push`
 will automatically force-push the bookmark.
 
-When your first PR has been approved, we typically give you contributor access,
+When your first PR has been approved, we invite you to the
```
martinvonz, line 61: nit: Keep "typically"? That was supposed to indicate that it's not guaranteed that we give them contributor access. In particular, I usually don't do it for a simple documentation update.

martinvonz (approved):

docs/contributing.md
```diff
@@ -58,7 +58,8 @@ commit. You can do that by creating a new commit on top of the initial commit
...
+`jj-vcs/contributors` team to give you contributor access,
```
martinvonz, line 62: You could make it a link to https://github.com/orgs/jj-vcs/teams/contributors if you think that's useful

PhilipMetzger: I don't think you can see that if you're not within the org, so probably not worth it.

docs/contributing.md
PhilipMetzger: done. 

> In particular, I usually don't do it for a simple documentation update.

Yes, I also started to land doc updates from non-longterm contributors like you.

PR #5545, opened 2025-01-31
docs: Document the existing deprecation policy

PhilipMetzger:
We recently renamed `jj util mangen` to `jj install-manpages` without a deprecation warning, as it was missed in the initial PR. Waleed then fixed it in a follow-up, which led to a longer discussion in the Discord which revealed that not everyone was aware of it and _what_ requires a warning based deprecation.

This initially categorizes the policy into user-visible UX changes and small impact renames, like for a small usergroup like packagers and third-party dependencies. If we have more hickups with users, this can be expanded.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -87,6 +87,14 @@ existing component, we require an architecture review from  multiple
 stakeholders, which we do with [Design Docs](design_docs.md), see the
 [process here](design_docs.md#process).
 
+## Deprecating and removing commands
+
+Before removing a required argument, or making a optional argument
+required or renaming a command, make yourself familiar with our
+[Deprecation policy](deprecation.md). The deprecation policy is designed to
+allow us to make progress quickly without breaking workflows to users too
+frequently.
+
 ## Contributing to the documentation
 
 We appreciate [bug
--- a/docs/deprecation.md
+++ b/docs/deprecation.md
@@ -1,0 +1,46 @@
+# Deprecation Policy & Breaking Changes
+
+This documentation gives a short outline of our deprecation strategy in the
+project. The general policy is here to ensure that we don't break our users
+workflows when renaming a command or making an argument required. The basis of
+this policy was introduced in [PR #1911] when we removed
+`--allow-large-revsets` in favor of `all:`.
+
+Binary format changes are currently not mentioned, as they currently need
+to be supported for a year or longer to allow the tooling ecosystem to adjust
+to it.
+
+## User-facing commands and their arguments
+
+When we rename a command or make a previously optional argument required,
+we usually try to keep the old command invocations working for 6
+months (so 6 releases, since we currently release monthly) with a deprecation
+message. The message should inform the user that the previous workflow is
+deprecated and to be removed in the future.
+
+## Niche commands
+
+For commands with a niche user audience or something we assume is rarely used
+(we sadly have no data), we take the liberty to remove the old behavior within
+two releases. This means that you can change the old command to immediately
+return a error. A example is if we want to rename `jj debug reindex` to
+`jj util reindex` in a release, then we make `jj debug reindex` an error in the
+same patchset. Since `util` and `debug` commands generally shouldn't be used as
+large API surface, it still should be fine to extend them without a required
+deprecation period.
+
+## Third-party dependencies
+
+For third-party dependencies which previously were used for a core functionality
+like `libgit2` was before the `[git.subprocess]` option was added, we're free
+to remove most codepaths and move it to a `cargo` feature which we support
+up to 6 releases, this is to ease transition for package managers. This is done
+on a case-by-case basis to allow the project to remove third-party dependencies
+as soon as possible.
+
+As an example, the `git.subprocess` option was implemented in 0.26 and the
+`libgit2` path was finally removed in 0.30, which was four months later after
+the community, maintainers and packagers agreed on it (within the aforementioned
+6 months).
+
+[PR #1911]: https://github.com/jj-vcs/jj/pull/1911
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -158,6 +158,7 @@ nav: # This lists all the files that become part of the documentation
       - Style guide: 'style_guide.md'
       - Design docs: 'design_docs.md'
       - Design doc blueprint: 'design_doc_blueprint.md'
+      - Deprecations: 'deprecation.md'
       - Releasing: 'releasing.md'
       - Temporary voting for governance: 'governance/temporary-voting.md'
       - Governance: 'governance/GOVERNANCE.md'


docs/contributing.md
```diff
@@ -82,6 +82,14 @@ existing component, we require an architecture review from  multiple
 stakeholders, which we do with [Design Docs](design_docs.md), see the
 [process here](design_docs.md#process).
 
+## Deprecating and removing commands
+
+Before removing a until now required argument, or making a optional argument
+required or even a user-visible command, make yourself familiar with our
```
ilyagr, line 88: "or even a user-visible command" seems out of place.

docs/deprecation.md
```diff
@@ -0,0 +1,23 @@
+# Deprecation Policy
+
+## User-facing commands and their arguments
+
+Whenever you rename a command or make a previously optional argument required,
+we require that you preserve the old command invocations keeps on working for 6
+months (so 6 releases, since we release monthly) with a deprecation message.
```
ilyagr, line 7: (For discussion) This might be a mostly a stylistic difference, but I'd prefer to make this document softer.

For example, I'd say,

> When we rename commands or make previously optional arguments required, we usually try to keep the old command invocations working for 6 months...

(there are also a few unrelated fixups in the above)


We could also insert

> When we rename *commonly used* commands...

docs/deprecation.md
```diff
@@ -0,0 +1,23 @@
...
+The message should inform the user that the previous workflow is deprecated
+and to be removed in the future.
+
+## Packaging commands and niche commands
+
+For commands with a niche user audience or something we assume is rarely used
+(we sadly have no data), we take the liberity to remove the old behavior within
+two releases. This means that you can change the old command to immediately
+error.
```
ilyagr, line 16: What do you mean about "within two releases"? This seems to contradict making the command to immediately error.

docs/deprecation.md
```diff
@@ -0,0 +1,23 @@
...
+
+## User-facing commands and their arguments
+
+Whenever you rename a command or make a previously optional argument required,
+we require that you preserve the old command invocations keeps on working for 6
+months (so 6 releases, since we release monthly) with a deprecation message.
+The message should inform the user that the previous workflow is deprecated
+and to be removed in the future.
+
+## Packaging commands and niche commands
```
ilyagr, line 11: (For discussion) You don't really discuss packaging commands in here. I think it'd make sense to have a separate section about how we treat packager-facing deprecations, but I'm not sure what it should say.

docs/deprecation.md
```diff
@@ -0,0 +1,23 @@
...
+
+For commands with a niche user audience or something we assume is rarely used
+(we sadly have no data), we take the liberity to remove the old behavior within
+two releases. This means that you can change the old command to immediately
+error.
+
+## Third-Party dependencies
+
+For third-party dependencies which previously were used for a core functionality
```
ilyagr, line 20: (For discussion) This section doesn't make a lot of sense to me. Perhaps it's OK to just remove it, in favor of a possible future "packager-facing deprecations" section.

docs/deprecation.md
```diff
@@ -0,0 +1,23 @@
...
+Whenever you rename a command or make a previously optional argument required,
+we require that you preserve the old command invocations keeps on working for 6
+months (so 6 releases, since we release monthly) with a deprecation message.
+The message should inform the user that the previous workflow is deprecated
+and to be removed in the future.
+
+## Packaging commands and niche commands
+
+For commands with a niche user audience or something we assume is rarely used
+(we sadly have no data), we take the liberity to remove the old behavior within
```
martinvonz, line 14: typo in "liberity"

docs/deprecation.md
```diff
@@ -0,0 +1,23 @@
...
+two releases. This means that you can change the old command to immediately
+error.
+
+## Third-Party dependencies
```
martinvonz, line 18: nit: lowercase "party" for consistency

docs/deprecation.md
```diff
@@ -0,0 +1,23 @@
+# Deprecation Policy
```
martinvonz, line 1: It feels natural to talk about breaking changes together with deprecations. I don't want to expand the scope too much. I'm just mentioning this in case you think it should affect the name of the file or the heading. I personally think it's fine to cover breaking changes without changing either.

docs/deprecation.md
martinvonz, line 16: Maybe we can mention `jj debug` commands specifically. I think breaking changes in them is always fine. But we can still try to balance it against how hard it is to keep compatibility.

docs/deprecation.md
martinvonz: Yes, IMO, these are more like rules of thumb than something we should strictly enforce. I think we should always balance how hard it is to keep compatibility vs how hard it is for users to adapt. I mentioned on Discord the example of file-level vs tree-level conflicts. In that case, dropping compatibility completely means that users have to re-clone their repo. As a result, we have been much more conservative with that deprecation. We still support file-level conflicts in old repos over a year later.

I think (but I may be wrong) when we originally started talking about the 6-month policy, it was about storage formats. I think it's fair to have a strict rule in that case saying that a repo created with a jj version less than 6 months old should be readable by the latest version.

docs/deprecation.md
PhilipMetzger: > Maybe we can mention `jj debug` commands specifically. I think breaking changes in them is always fine. But we can still try to balance it against how hard it is to keep compatibility.

Since the `jj debug` commands actually are internal APIs, I think they don't even should get mentioned.

docs/deprecation.md
PhilipMetzger: cc @emilazy since she brought it up in the [Discord thread]. I'm also happy to combine it with the next section if that works for you. 

[Discord thread]: https://discord.com/channels/968932220549103686/1334974861067686004

docs/deprecation.md
PhilipMetzger: I've added "& Breaking Changes" to the title

docs/deprecation.md
PhilipMetzger: done

docs/deprecation.md
PhilipMetzger: fixed

docs/contributing.md
```diff
@@ -82,6 +82,14 @@ existing component, we require an architecture review from  multiple
...
+[Deprecation policy](deprecation.md). This makes the tool pleasant to use even
+if the hasn't reached a 1.0 yet. It allows changes like the `branch` to
+`bookmark` rename or the removal of `checkout` and `merge` to happen.
```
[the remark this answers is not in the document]
PhilipMetzger: done

docs/contributing.md
```diff
@@ -82,6 +82,14 @@ existing component, we require an architecture review from  multiple
 stakeholders, which we do with [Design Docs](design_docs.md), see the
 [process here](design_docs.md#process).
 
+## Deprecating and removing commands
+
+Before removing a until now required argument, or making a optional argument
```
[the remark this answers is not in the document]
PhilipMetzger: deleted

docs/contributing.md
```diff
@@ -82,6 +82,14 @@ existing component, we require an architecture review from  multiple
...
+required or even a user-visible command, make yourself familiar with our
+[Deprecation policy](deprecation.md). This makes the tool pleasant to use even
+if the hasn't reached a 1.0 yet. It allows changes like the `branch` to
```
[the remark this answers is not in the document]
PhilipMetzger: removed in favor of your suggestion below

docs/deprecation.md
PhilipMetzger: I renamed it to just "Niche dommands", maybe that fits better?

docs/deprecation.md
PhilipMetzger: I've added a example with a debug command

docs/deprecation.md
PhilipMetzger: Yes, fair points. I have left the storage formats unmentioned since I had no good formulation of it.

docs/contributing.md
PhilipMetzger: rephrased

docs/contributing.md
```diff
@@ -85,6 +85,14 @@ existing component, we require an architecture review from  multiple
 stakeholders, which we do with [Design Docs](design_docs.md), see the
 [process here](design_docs.md#process).
 
+## Deprecating and removing commands
+
+Before removing a required argument, or making a optional argument
+required or renaming a command, make yourself familiar with our
+[Deprecation policy](deprecation.md). The deprecation policy is designed to
+allow us to make progress quickly without breaking workflows to users to
```
martinvonz, line 93: nit: the last "to" should be "too"

docs/deprecation.md
```diff
@@ -0,0 +1,28 @@
...
+and to be removed in the future.
+
+## Niche commands
+
+For commands with a niche user audience or something we assume is rarely used
+(we sadly have no data), we take the liberty to remove the old behavior within
+two releases. This means that you can change the old command to immediately
+return a error. A example is if we want to rename `jj debug reindex` to
+`jj util reindex` in a release, then we make `jj debug reindex` an error in the
+same patchset.
```
martinvonz, line 21: Does this mean that even for niche commands, it's not okay to make breaking changes from one release to the next? I'm personally fine with making changes like #3983 in the future too.

docs/deprecation.md
martinvonz: I agree. I think there are two different things involved in this transition. One is the user-visible changes. That's what I think we need to be careful with. That's not specific to third-party dependencies. The other thing is the actual removal of the dependency. I don't think we should have a deprecation policy for that (we should remove dependencies as soon as they're unused).

docs/deprecation.md
```diff
@@ -0,0 +1,28 @@
+# Deprecation Policy & Breaking Changes
+
+This documentation gives a short outline of our deprecation strategy in the
+project.
```
martinvonz, line 4: I think it's useful to write something about the general policy here.

docs/deprecation.md
PhilipMetzger: > Does this mean that even for niche commands, it's not okay to make breaking changes from one release to the next?

I don't think so, as it currently only mentions renames or removals not extending basic functionality.

docs/deprecation.md
PhilipMetzger: I've written something but am not really happy with it.

PhilipMetzger:
any interest left to bring this over the line?

docs/deprecation.md
```diff
@@ -0,0 +1,46 @@
+# Deprecation Policy & Breaking Changes
+
+This documentation gives a short outline of our deprecation strategy in the
+project. The general policy is here to ensure that we don't break our users
+workflows when renaming a command or making an argument required. The basis of
+this policy was introduced in [PR #1911] when we removed
+`--allow-large-revsets` in favor of `all:`.
+
+Binary format changes are currently not mentioned, as they currently need
```
ilyagr, line 9: I would get rid of this paragraph (but my other comment takes precedence)

docs/deprecation.md
ilyagr: Now this section seems even more obsolete. (but my other comment takes precedence)

ilyagr (commented):
I gave this another look, and I'm currently leaning towards just replacing this with a paragraph inside the Contributing doc, something like:

> ### Deprecating and removing commands
>
> When we rename a command or make a previously optional argument required,
we usually try to keep the old command invocations working for 6
months (so 6 releases, since we currently release monthly) with a deprecation
message. The message should inform the user that the previous workflow is
deprecated and to be removed in the future.
>
> For commands with a niche user audience or something we assume is rarely used
(we sadly have no data), we take the liberty to remove the old behavior within
two releases.
>
> Other exceptions can be made, if necessary.

If we remove the "third-party dependencies" section and reword the other paragraphs to be more of suggestions, not that much is left of that doc: just the idea of "6 releases". OTOH, that idea does seem worth writing down somewhere.

----------

We could also do the opposite, and expand this document. I'm guessing this might be your plan, but while I have a few examples in mind of stories we could add, I'm not sure they are actionable or conclusive. (Some stories that come to mind are the effect on plugins of repo format changes on `gg`/VisualJJ or of moving the `config.toml` location on Macs. However, I don't think what we did in those situations was clearly the right thing or clearly the wrong thing. So, if I wrote them up, the conclusion would be unclear.)

PhilipMetzger:
> I gave this another look, and I'm currently leaning towards just replacing this with a paragraph inside the Contributing doc, something like:
> 
> > ### Deprecating and removing commands
> > When we rename a command or make a previously optional argument required,
> > we usually try to keep the old command invocations working for 6
> > months (so 6 releases, since we currently release monthly) with a deprecation
> > message. The message should inform the user that the previous workflow is
> > deprecated and to be removed in the future.
> > For commands with a niche user audience or something we assume is rarely used
> > (we sadly have no data), we take the liberty to remove the old behavior within
> > two releases.
> > Other exceptions can be made, if necessary.
> 
> If we remove the "third-party dependencies" section and reword the other paragraphs to be more of suggestions, not that much is left of that doc: just the idea of "6 releases". OTOH, that idea does seem worth writing down somewhere.

I still disagree with you here, while this _essentially_ achieves the same thing, it isn't publicly visible. This document gives a nice warning to the users (distro packagers included). The goal of this document is to give the project some leeway to continuously make breaking changes until the eventual stabilization and not to scare away people with constant breakages, it also allows publicly visible policy changes. See it as a PR move more than anything.  

OT: I don't it like that you seemingly dismiss my work here, which also stems from a Discord discussion where at least some people found it a good thing (feel free to DM me in Discord, if you want to discuss this). 

> We could also do the opposite, and expand this document. I'm guessing this might be your plan, but while I have a few examples in mind of stories we could add, I'm not sure they are actionable or conclusive. (Some stories that come to mind are the effect on plugins of repo format changes on `gg`/VisualJJ or of moving the `config.toml` location on Macs. However, I don't think what we did in those situations was clearly the right thing or clearly the wrong thing. So, if I wrote them up, the conclusion would be unclear.)

👍 , I'd much rather spend time on discussing these things, since there's now an actual ecosystem to take care of while still keeping our current speed (even though it has slowed down considerably).

ilyagr:
> OT: I don't it like that you seemingly dismiss my work here, which also stems from a Discord discussion

I didn't mean to; it probably comes from my uncertainty about what I should do here. This is a bit of an experiment for me, as I'm trying to figure out how to make some aspects of my work on `jj` function better.

The puzzle I find myself in is:

- I feel bad not saying anything here, which leaves you and this PR on the hook, and shifts this weight onto the other maintainers. This PR has already been in limbo for a long time.
- While portions of this PR and its intention seem good, I find other portions not very good in their current state
- I don't immediately have an idea of how to modify this PR to achieve its intentions fully. Trying to think about that further currently doesn't feel as productive or entertaining or helpful to me as other work I could be doing. Still, I don't want to reject the possibility that someone else might have one or might want to.
- I am unsure whether we have a good enough idea of what our deprecation policy should be to have one. I also think that it's dangerous to write down imperfect policies; they can be difficult to change later.

What I tried to do with my previous comments was to suggest a small subset of this work that, I think, would be a clear improvement over the status quo.

We could also (or instead) keep this PR open for discussion. I tried to paint a picture of what direction that could take, while also saying that I'm not sure how to make it really happen.

One way or another, I was hoping to get this to a point where I've done what I can for now and I can move it off my personal TODO list in the near future, until something changes in this equation. If I don't learn to do so, the amount of things on this TODO list that feel like they should be actionable but that I don't actually know what to do about becomes overwhelming.

I hope this makes my thoughts feel less dismissive, though I'm not sure. A part of this experiment is whether I can dismiss something from my TODO list without dismissing the effort other people have put into their work; feedback is appreciated (privately via Discord is fine, unless you think other people's input might also be helpful).

PhilipMetzger:
I have no further interest in pursuing this, as I think I'm the wrong person to write processes down for the project. As the maintainers have generally not shown any interest in it.

PR #5546, opened 2025-02-01
cli: add `--untrack` option to `bookmark forget`

### LINKED ISSUE #5524: FR: `jj bookmark untrack` should delete the local bookmark
Sometimes I want to work on a remote bookmark, so I run `jj bookmark track name@remote`, and then I make some changes and push them, and then I want to stop tracking the bookmark since I'm done working on it, so I run `jj bookmark untrack name@remote`. I would expect `untrack` to do the opposite of `track` (meaning it deletes the local bookmark that `track` created), but currently it doesn't do this, so I have to run `jj bookmark delete name` afterwards.

When a remote bookmark is untracked by `jj bookmark untrack` and it was the last tracked remote bookmark with that name (to handle the case of tracking from multiple remotes), then any local bookmark with that name should be automatically deleted (but nothing should be abandoned). To me, it seems uncommon to want to untrack a remote bookmark while keeping the local bookmark, but there could be a `--keep-local` flag to keep the local bookmark if this functionality is still desired.

We could continue to require the user to run two commands in this case. We could also keep the default the same, but add a `--delete-local` flag to `jj bookmark untrack` instead.

scott2000:
Fixes #5524. I went with `--untrack` for the naming because it seemed easier to explain in the help text than `--local`.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/bookmark/untrack.rs
+++ b/cli/src/commands/bookmark/untrack.rs
@@ -26,6 +26,9 @@ use crate::ui::Ui;
 ///
 /// A non-tracking remote bookmark is just a pointer to the last-fetched remote
 /// bookmark. It won't be imported as a local bookmark on future pulls.
+///
+/// If you want to forget a local bookmark while also untracking the
+/// corresponding remote bookmarks, use `jj bookmark forget --untrack` instead.
 #[derive(clap::Args, Clone, Debug)]
 pub struct BookmarkUntrackArgs {
     /// Remote bookmarks to untrack
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,6 +143,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Conditional configuration now supports `--when.commands` to change configuration
   based on subcommand.
 
+* `jj bookmark forget` now accepts a `--untrack` flag to untrack remote
+  bookmarks instead of forgetting them.
+
 ### Fixed bugs
 
 * `jj git fetch` with multiple remotes will now fetch from all remotes before
--- a/cli/src/commands/bookmark/forget.rs
+++ b/cli/src/commands/bookmark/forget.rs
@@ -33,6 +33,10 @@ use crate::ui::Ui;
 /// recreated on future pulls if it still exists in the remote.
 #[derive(clap::Args, Clone, Debug)]
 pub struct BookmarkForgetArgs {
+    /// Untrack remote bookmarks instead of forgetting them
+    #[arg(long, short)]
+    untrack: bool,
+
     /// The bookmarks to forget
     ///
     /// By default, the specified name matches exactly. Use `glob:` prefix to
@@ -61,8 +65,12 @@ pub fn cmd_bookmark_forget(
         tx.repo_mut()
             .set_local_bookmark_target(name, RefTarget::absent());
         for (remote_name, _) in &bookmark_target.remote_refs {
-            tx.repo_mut()
-                .set_remote_bookmark(name, remote_name, RemoteRef::absent());
+            if args.untrack {
+                tx.repo_mut().untrack_remote_bookmark(name, remote_name);
+            } else {
+                tx.repo_mut()
+                    .set_remote_bookmark(name, remote_name, RemoteRef::absent());
+            }
         }
     }
     writeln!(ui.status(), "Forgot {} bookmarks.", matched_bookmarks.len())?;
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -338,7 +338,7 @@ Forget everything about a bookmark, including its local and remote targets
 
 A forgotten bookmark will not impact remotes on future pushes. It will be recreated on future pulls if it still exists in the remote.
 
-**Usage:** `jj bookmark forget <NAMES>...`
+**Usage:** `jj bookmark forget [OPTIONS] <NAMES>...`
 
 ###### **Arguments:**
 
@@ -347,6 +347,10 @@ A forgotten bookmark will not impact remotes on future pushes. It will be recrea
    By default, the specified name matches exactly. Use `glob:` prefix to select bookmarks by [wildcard pattern].
 
    [wildcard pattern]: https://jj-vcs.github.io/jj/latest/revsets/#string-patterns
+
+###### **Options:**
+
+* `-u`, `--untrack` — Untrack remote bookmarks instead of forgetting them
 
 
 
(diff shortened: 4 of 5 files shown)


yuja:
iirc, @ilyagr had some idea about `jj branch forget` options.

My feeling is that `jj branch forget --untrack` sounds like an option to do more things than the default, but actually it is an option to restrict the target scope, which is surprising.

ilyagr:
Thank you for working on this!

I have thought of this operation as `jj forget --local`, since you forget only the local bookmark. We [once discussed this on Discord](https://discord.com/channels/968932220549103686/969291218347524238/1256137371561234442) and Yuya preferred `untrack --all-remotes` back then. I think I still prefer `forget --local`, but either that or `untrack --all-remotes` seem better than `forget --untrack`.

-------------

The reason I prefer `forget --local` is that I also want `jj forget --remote branch@remote` that would forget a remote-tracking branch without forgetting the local branch (`jj forget --remote 'glob:branch@*'` would also be allowed). One use-case for this is when I fetched more branches than I wanted from `upstream` and want to undo this mistake.

We don't have to implement it in this PR, but this UI with `--local` and `--remote` makes sense to me . This results in the trichotomy of three modes for `jj forget`:

1. Forget the local branch and all remote-tracking branches
2. Forget some or all of the remote-tracking branches, but keep the local one.
3. Forget only the local branch without marking the remote-tracking branches for deletion.
 
The confusion is that "forget the local branch and untrack all the remote-tracking branches" doesn't fit into this pattern quite as well, but the fact that it's equivalent to 3. as I described above is good enough for me.

cli/src/commands/bookmark/untrack.rs
```diff
@@ -26,6 +26,9 @@ use crate::ui::Ui;
 ///
 /// A non-tracking remote bookmark is just a pointer to the last-fetched remote
 /// bookmark. It won't be imported as a local bookmark on future pulls.
+///
+/// If you want to forget a local bookmark while also untracking the
+/// corresponding remote bookmarks, use `jj bookmark forget --untrack` instead.
```
ilyagr, line 31: Nice idea to mention it here! (No action needed, beyond updating the name if we change it)

yuja:
> I have thought of this operation as `jj forget --local`, since you forget only the local bookmark. We [once discussed this on Discord](https://discord.com/channels/968932220549103686/969291218347524238/1256137371561234442) and Yuya preferred `untrack --all-remotes` back then. I think I still prefer `forget --local`, but either that or `untrack --all-remotes` seem better than `forget --untrack`.

(Perhaps, `untrack --all-remotes` wouldn't delete local bookmark, but I don't remember why `forget --local` sounded weird.)

If I redesign `forget` from scratch, I would probably make:

* `forget NAME` deletes matching local bookmarks and disassociates them from tracking remotes (i.e. `forget --local`/`--untrack`)
* `forget NAME@REMOTE` deletes matching remote bookmarks
* `forget --all[-remotes] NAME` deletes matching local and remote bookmarks (the current behavior)

I think the tricky part of `forget --local` is that it actually updates per-remote state.

scott2000:
> The reason I prefer `forget --local` is that I also want `jj forget --remote branch@remote` that would forget a remote-tracking branch without forgetting the local branch (`jj forget --remote 'glob:branch@*'` would also be allowed). One use-case for this is when I fetched more branches than I wanted from `upstream` and want to undo this mistake.
> 
> We don't have to implement it in this PR, but this UI with `--local` and `--remote` makes sense to me.

Yeah `--local` does make a lot more sense if there's also going to be a `--remote` option as well.

> If I redesign `forget` from scratch, I would probably make:
> 
>   * `forget NAME` deletes matching local bookmarks and disassociates them from tracking remotes (i.e. `forget --local`/`--untrack`)
> 
>   * `forget NAME@REMOTE` deletes matching remote bookmarks
> 
>   * `forget --all[-remotes] NAME` deletes matching local and remote bookmarks (the current behavior)

I like this a lot better than the current behavior. It's always felt strange to me that if I pass the name of a local bookmark, it forgets the remote bookmarks too. I would be in favor of changing it to this or something like this.

> I think the tricky part of `forget --local` is that it actually updates per-remote state.

Yeah, I think that's what made me want to go with `--untrack` instead of `--local`, since it seems weird to me that `--local` would change the behavior of the command for remote bookmarks without changing the behavior for local bookmarks.

martinvonz:
Yuya's proposal makes sense to me too. I'm not at all attached to how it currently behaves.

ilyagr:
I also like Yuya's proposal. I think the state the repo is in after `jj forget --local` is usually more desireable than the state after the current `jj forget`, since it is not messed up by a subsequent `fetch`. So, it makes sense to make that the default.

My only nit (that I'm sure we can figure out) is that I don't like the `--all-remotes` name for the option to represent the current behavior, since I'd expect that to only forget the remote-tracking branches. The best name I came up with so far is `--with-all-remotes` (the short can still be `--all`) or `--also-forget-remotes`, or maybe `--forget-remotes`[^1]. Another option that came to mind is `--everywhere`, but that's slightly optimistic since it doesn't forget the branch on actual remote servers.

[^1]: I'm conflicted about this one. In theory, it shares the same problem as `--all-remotes`, but it's somehow easier for me to imagine `--forget-remotes` meaning "do the usual thing and also forget remotes".

ilyagr:
There's also the question of how to make this transition. Do you think it's OK to make this breaking change directly?

At some point, I had the plan to require at least one of `--local` and `--remote` for `jj forget` in a transitional period. Then, I thought we'd make `--local` the default.

At that point, I was thinking of the syntax as being `jj forget br --remote someremote` instead of `jj forget br@someremote` or `jj forget --remote br@someremote`.

I think we could still have an intermediate period where we'd require an argument when specifying the local branch, either `jj forget --local br` or `jj forget --with-all-remotes br`. Instead of --remote, we can simply allow `jj forget br@someremote`.

joyously:
> I'm really struggling to think of any use case where it's necessary to forget both local and remote bookmark at the same time, and also untracking isn't sufficient.

I was just going to ask for all the scenarios this is trying to cover, so I could get a better feel for the option that would make sense.
I think there are different cases when you have only one remote versus multiple remotes. Also a factor is your workflow (stacking changes on main, or pushing single revisions to your fork). And if you fetch other authors' work, you want to forget more bookmarks.

yuja:
> I don't like the `--all-remotes` name for the option to represent the current behavior, since I'd expect that to only forget the remote-tracking branches.

Agreed. It could also mean `forget --all-remotes` (without name patterns).

How about `--include-remotes`? I don't think this option is useful, but something like this would be needed for smoother migration. I think breaking change is acceptable. `bookmark forget` wouldn't be widely used.

scott2000:
> How about `--include-remotes`? I don't think this option is useful, but something like this would be needed for smoother migration. I think breaking change is acceptable. `bookmark forget` wouldn't be widely used.

I like it! I can take a shot at implementing it sometime this week, but if someone else would like to do it instead that's fine with me too.

PR #5552, opened 2025-02-02
favicon: make corners round, birds a bit larger

ilyagr:
Follow up to #5501.

~~I haven't been able to make the PNG quite as small, hints welcome. When zoomed in, the birds are also now a bit blurry; making them larger resulted in fractional scaling. However, I don't think this matters at favicon sizes.~~

**Update:** I think I managed to fix what I was talking about above. Inkscape FTW. The rendering of the sharper birds is not perfect on Firefox (actually worse than for fuzzy ones), but I guess I'll just blame Firefox for that.

I still wonder whether we should eventually have a simpler design for the favicon (so that the birds are more obvious), but this should be fine for now. Also, nobody will be confused about where the walrus is 🦭.

image

image

Compared to ("before" state): 

image

and 

image


ilyagr:
Another option, especially if we like the smaller birds:

image

thoughtpolice (approved):
LGTM, thanks for this attention to detail!

PR #5557, opened 2025-02-02
revset: escape symbols in "did you mean" hint

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -2686,13 +2686,29 @@ pub struct RevsetWorkspaceContext<'a> {
     pub workspace_id: &'a WorkspaceId,
 }
 
-/// Formats a string as symbol by quoting and escaping it if necessary
+/// Formats a string as symbol by quoting and escaping it if necessary.
+///
+/// Note that symbols may be substituted to user aliases. Use
+/// [`format_string()`] to ensure that the provided string is resolved as a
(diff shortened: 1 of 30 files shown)


lib/src/revset.rs
```diff
@@ -1885,15 +1885,14 @@ fn resolve_remote_bookmark(repo: &dyn Repo, name: &str, remote: &str) -> Option<
...
 }
 
-fn all_bookmark_symbols(
+fn all_formatted_bookmark_symbols(
     repo: &dyn Repo,
     include_synced_remotes: bool,
 ) -> impl Iterator<Item = String> + '_ {
     let view = repo.view();
     view.bookmarks().flat_map(move |(name, bookmark_target)| {
-        // Remote bookmark "x"@"y" may conflict with local "x@y" in unquoted form.
```
martinvonz, line 1894: Can the `.dedup()` in `collect_similar()` be dropped now? I would expect the candidates passed to that function to not have duplicates, and I'm guessing it did only because of the conflict fixed by this patch.

martinvonz (approved):
Good idea!

lib/src/revset.rs
yuja: Hmm, yes. Templater could pass duplicated `"self"` entries in very rare cases, but they'll be deduplicated later.

Removed `.dedup()`.

PR #5100, opened 2024-12-14
feat(cli): add buzz command

InCogNiTo124:
`jj buzz` is an easter egg command that sends a bell character to the terminal. Running `jj buzz` is expected to make a bell sound

The code was adapted from the simplest subcommand, `jj version`.

https://discord.com/channels/968932220549103686/1167230861448581167/1317610036733673684

# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

<hr/>

NOTE: not a serious proposal

PR #5561, opened 2025-02-02
cli: jj accuse!

ccmtaylor:
all credit to @skade, with apologies to Émile Zola

fixes https://hachyderm.io/@skade/113873652877876436

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -79,6 +79,9 @@ use crate::ui::Ui;
 enum Command {
     Abandon(abandon::AbandonArgs),
     Absorb(absorb::AbsorbArgs),
+    // with apologies to Émile Zola: https://en.wikipedia.org/wiki/J%27Accuse...!
+    #[command(hide = true)]
+    Accuse(file::annotate::FileAnnotateArgs),
     Backout(backout::BackoutArgs),
     #[cfg(feature = "bench")]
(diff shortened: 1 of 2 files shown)


ilyagr:
It's like the speaker has a stutter.

I'm OK with this.

cli/src/commands/mod.rs
```diff
@@ -79,6 +79,8 @@ use crate::ui::Ui;
 enum Command {
     Abandon(abandon::AbandonArgs),
     Absorb(absorb::AbsorbArgs),
+    #[command(hide = true)]
+    Accuse(file::annotate::FileAnnotateArgs),
```
ilyagr, line 83: You can add a comment pointing to <https://en.wikipedia.org/wiki/J%27Accuse...!>, to emphasize that we're being *historical* as opposed to merely mean.

martinvonz:
See also #5100. I'm concerned that it's hard to decide where to draw the line if we start accepting Easter eggs. (I'm guessing this is an Easter egg and not something we expect people will use much, but maybe they will?)

PhilipMetzger:
Even if we're doing this, it should be kept a config alias not something builtin.

ccmtaylor:
totally fair; I can see this turning into a [spacebar heating](https://xkcd.com/1172/). I'll close the PR

PR #5565, opened 2025-02-03
[feat]: Add oldest() function that complements the latest() function in the revset language 

AjithPanneerselvam:
# Checklist

If applicable:


PhilipMetzger:
Hello, welcome to the project. This can be all one commit and should be titled something like `revset:`, per https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#code-reviews

martinvonz:
nit: it seems like it should be `lastest()` and `earliest()` (or `newest()` and `oldest()`). Or maybe the inconsistency is fine.

AjithPanneerselvam:
> nit: it seems like it should be `lastest()` and `earliest()` (or `newest()` and `oldest()`). Or maybe the inconsistency is fine.

Yeah, it makes sense! I'll go with `earliest` to stay consistent with the existing `latest` function.

emilazy:
I like the rename to `newest()`/`oldest()`, because currently people often use `latest()` when they want `heads()`.

AjithPanneerselvam:
> I like the rename to `newest()`/`oldest()`, because currently people often use `latest()` when they want `heads()`. The renames seem less confusable with the topological notion that is usually more relevant.

Ah, I just [renamed the references in this PR to earliest()](https://github.com/jj-vcs/jj/pull/5565/commits/6400d1dbb6f2342ece4527046210ad41a1bca5c3) 😅 That's okay I can revert this commit. 

@emilazy I reckon the renaming from `latest()` to `newest()` is a breaking change. Can I rename it in a separate PR?

emilazy:
Yeah not trying to hold up this PR, just saying that I think renaming them both actually seems like a good idea in general. We’d want to keep a backwards compatibility alias etc. for a while.

PhilipMetzger:
Sorry, it is the _commit message_ which should contain `revset:` not the PR title.

yuja:
> I like the rename to `newest()`/`oldest()`, because currently people often use `latest()` when they want `heads()`. The renames seem less confusable with the topological notion that is usually more relevant.

fwiw, `latest()` could have some topological notion assuming that all ancestors commits are (usually) older than the commit itself. The current implementation scans all commits, which is inefficient.

jonathantanmy:
The GitHub UI on my computer says "AjithPanneerselvam marked this pull request as ready for review 2 weeks ago" - is this still true? If yes...

Taking a step back, I'm not fully convinced that an oldest()/earliest() function is as useful as the currently existing latest() function. Having said that, it is strange that we have latest() without earliest(). So I'm neutral about whether we want this feature in the first place. Assuming we want this feature:

- Are you going with "oldest" or "earliest"? As far as I can tell, your commit still contains "oldest". (I think it should be "earliest" - there might be an argument that we should use the names "newest" and "oldest" instead, but I personally have no preference, so I'm fine with merging "earliest" and worrying later about whether we should have "newest"/"oldest". But if we decide now that we want "newest", then we should make that change first and then use "oldest" in this PR.)
- There is substantial code duplication in `lib/src/default_index/revset_engine.rs`. You probably should refactor the existing code into a function that takes a comparison function, and use that function both for latest() and earliest().
- Add at least a test in `lib/tests/test_revset.rs` similar to the existing ones that test latest().
- Update the commit message, not the PR title, as PhilipMetzger said.

AjithPanneerselvam:
Well, I was going with `oldest` based on [this comment](https://github.com/jj-vcs/jj/pull/5565#issuecomment-2634273593). But, regardless of what you suggest (earliest) or `oldest`, we need to rename the existing `latest` to `newest` anyway. Before, the idea was to do the rename after merging this PR, but I'm fine creating a PR for that rename first and come back to this after. Thanks for your input! 

> * Are you going with "oldest" or "earliest"? As far as I can tell, your commit still contains "oldest". (I think it should be "earliest" - there might be an argument that we should use the names "newest" and "oldest" instead, but I personally have no preference, so I'm fine with merging "earliest" and worrying later about whether we should have "newest"/"oldest". But if we decide now that we want "newest", then we should make that change first and then use "oldest" in this PR.)

As there are many conversation regarding the naming, to summarize
- We are renaming `latest` to `newest` (Action for me - will create a PR). 
- For now, I'm sticking with `oldest` as in this PR implementation.

yuja:
> * We are renaming `latest` to `newest` (Action for me - will create a PR).
> * For now, I'm sticking with `oldest` as in this PR implementation.

Do we agree on the rename? I kinda feel negative because `newest()` seems to imply more about timestamp-based filtering. It might be nice if `latest(heads, n)` is optimized in a way that it scans at most `n` ancestor commits from the `heads`.

emilazy:
Isn’t `latest()` exactly about timestamp ordering? If we don’t want to do timestamp ordering (I agree that it usually doesn’t make much sense), then maybe we should just drop it. But currently the name `latest()` is ambiguous and makes people use it when they might want `heads()` or similar, whereas `newest()` would be accurate to (my understanding of) its actual semantics.

yuja:
> Isn’t `latest()` exactly about timestamp ordering? If we don’t want to do timestamp ordering (I agree that it usually doesn’t make much sense), then maybe we should just drop it.

It is because the implementation is naive. I think `latest(x, 1)` can be relaxed as a function that disambiguates `heads(x)` based on timestamp, and `latest(x, 2)` will be `latest(x, 1) | latest(x ~ latest(x, 1), 1)`.

PR #5573, opened 2025-02-04
pyproject.toml: update to newer syntax

ilyagr:
We've been using `uv 0.4` syntax, but we only support `uv` 0.5.1+. The newer `uv 0.5` syntax follows an approved PEP, and the older syntax will probably be removed at some point.

The CI verifies that this works with uv 0.5.1.


martinvonz (approved):

ilyagr:
Thanks!

Issue #3700, opened 2024-05-16
FR: jj undo ergonomics


benbrittain:
**Is your feature request related to a problem? Please describe.**
The current behavior of the jj undo command does not seem to be what users intuitively expect. Currently if you run `jj undo` twice in a row it has the same behavior as `jj op undo`, it returns you to the same place you were when you started. undo + undo = no change.

**Describe the solution you'd like**
I'd like to preserve the ergonomics of the underlying `jj op undo` command (as suggested by @PhilipMetzger on Discord), but have the higher level `jj undo` command track if a user immediately undoes an undo and warn them.

**Describe alternatives you've considered**
Alternatively, the jj undo command could keep track of where you are in the undo log ctrl-Z style. That probably implies a redo command as well?

martinvonz:
> I'd like to preserve the ergonomics of the underlying `jj op undo` command

Why make them different? Oh, I suppose the top-level `jj undo` would lose its optional operation argument?

benbrittain:
I'm actually starting to question if `jj op undo` is needed at all. I'd originally been modeling it as a lower level operation, but it's really just a special case of `jj op restore`, no? I can't imagine I'd ever use `jj op undo` if jj undo had `ctrl-z` semantics

martinvonz:
> it's really just a special case of `jj op restore`, no?

No, `jj undo @-` undoes the second-to-last operation while leaving the changes from the last operation. For example, if you abandoned some commit, and then did a bunch of unrelated changes, you can still recover that commit with `jj undo <old operation>`. It can be hard to reason about, however, because it also brings back ancestors of the old commit when it makes the old abandoned commit visible.

martinvonz:
Think of it like `jj backout` but for operations (while `jj op restore` is `jj restore` for operations).

PhilipMetzger:
To reword the FR a bit: 

Improve the ergonomics of `jj undo` by making it a separate command, which can do `undo` tracking like editors. Currently running `undo` twice in a row, surprises a lot of users as it just rolls back the last operation (not surprising if you know that `jj undo` is an alias of `jj op undo`). This would then pave a way for a smart `redo`, which has the opposite behavior. 

> **Describe alternatives you've considered**
> Alternatively, the jj undo command could keep track of where you are in the undo log ctrl-Z style. That probably implies a redo command as well?

[...]

PR #5580, opened 2025-02-04
Show a warning when undoing an undo operation

MrJohz:
One common issue that users run into with `jj undo` is that they expect it to behave like an "undo stack", in which every consecutive `jj undo` execution moves the current state further back in history. This is (currently) an incorrect intuition - running multiple `jj undo` commands back-to-back will only result in the previous non-undo change being reverted, then un-reverted, then reverted, and so on.

This change adds a hint when the user runs `jj undo` multiple times consecutively, suggesting that they may be looking for `jj restore`, which allows them to revert the whole state of the repository back to a previous snapshot.

See also #3700

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj undo` now shows a hint when undoing an undo operation that the user may
+   be looking for `jj op restore` instead.
+
 * Template functions `truncate_start()` and `truncate_end()` gained an optional
   `ellipsis` parameter; passing this prepends or appends the ellipsis to the
   content if it is truncated to fit the maximum width.
--- a/cli/src/commands/operation/undo.rs
+++ b/cli/src/commands/operation/undo.rs
@@ -70,6 +70,7 @@ pub fn cmd_op_undo(
         &args.what,
     );
     tx.repo_mut().set_view(new_view);
+    let was_undo_op = *tx.repo().view() == parent_op.view()?;
     if let Some(mut formatter) = ui.status_formatter() {
         write!(formatter, "Undid operation: ")?;
         let template = tx.base_workspace_helper().operation_summary_template();
@@ -77,6 +78,19 @@ pub fn cmd_op_undo(
         writeln!(formatter)?;
     }
     tx.finish(ui, format!("undo operation {}", bad_op.id().hex()))?;
+
+    if args.operation == "@" && was_undo_op {
+        writeln!(
+            ui.hint_default(),
+            "This action reverted an 'undo' operation. The repository is now in the same state as \
+             it was before the original 'undo'."
+        )?;
+        writeln!(
+            ui.hint_default(),
+            "If your goal is to undo multiple operations, consider using `jj op log` to see past \
+             states, and `jj op restore` to restore one of these states."
+        )?;
+    }
 
     Ok(())
 }
--- a/cli/tests/test_undo.rs
+++ b/cli/tests/test_undo.rs
@@ -385,6 +385,44 @@ fn test_bookmark_track_untrack_undo() {
     "###);
 }
 
+#[test]
+fn test_shows_a_warning_when_undoing_an_undo_operation_as_bare_jj_undo() {
+    let test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    test_env.jj_cmd_ok(&repo_path, &["new"]);
+    test_env.jj_cmd_ok(&repo_path, &["undo"]);
+    let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["undo"]);
+    insta::assert_snapshot!(stdout, @"");
+    insta::assert_snapshot!(stderr, @r"
+    Undid operation: 2d5b73a97567 (2001-02-03 08:05:09) undo operation 289cb69a8458456474a77cc432e8009b99f039cdcaf19ba4526753e97d70fee3fd0f410ff2b7c1d10cf0c2501702e7a85d58f9d813cdca567c377431ec4d2b97
+    Working copy now at: rlvkpnrz 65b6b74e (empty) (no description set)
+    Parent commit      : qpvuntsm 230dd059 (empty) (no description set)
+    Hint: This action reverted an 'undo' operation. The repository is now in the same state as it was before the original 'undo'.
+    Hint: If your goal is to undo multiple operations, consider using `jj op log` to see past states, and `jj op restore` to restore one of these states.
+    ");
+}
+
+#[test]
+fn test_shows_no_warning_when_undoing_a_specific_undo_change() {
+    let test_env = TestEnvironment::default();
+    test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
+    let repo_path = test_env.env_root().join("repo");
+
+    test_env.jj_cmd_ok(&repo_path, &["new"]);
+    test_env.jj_cmd_ok(&repo_path, &["undo"]);
+    let stdout = test_env.jj_cmd_success(&repo_path, &["op", "log"]);
+    let op_id_hex = stdout[3..15].to_string();
+    let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["undo", &op_id_hex]);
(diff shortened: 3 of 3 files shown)


cli/src/commands/operation/undo.rs
```diff
@@ -75,6 +76,16 @@ pub fn cmd_op_undo(
         let template = tx.base_workspace_helper().operation_summary_template();
         template.format(&bad_op, formatter.as_mut())?;
         writeln!(formatter)?;
+
+        let is_bare_undo =
+            command.matches().subcommand_matches("operation").is_none() && args.operation == "@";
+        let was_undo_operation = bad_op.metadata().description.starts_with("undo operation");
```
MrJohz, line 82: I'm not entirely happy with this check here, but it does work (as long as the operation description remains consistent).

A better alternative might be to attach metadata to each operation that indicates whether that operation is an undo.  I see there is the space for arbitrary key-value pairs in the metadata, presumably I could use that?  For now I've chosen to take the easier option, though, to keep the scope of this PR small.

martinvonz: Yes, we've talked about adding metadata. For now, I think we can instead compare the view objects:

```rust
let was_undo_operation = tx.repo().view() == parent_op.view();
```

CHANGELOG.md
```diff
@@ -77,6 +77,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj undo` now shows a hint when undoing an undo operation that the user may be looking for `jj restore` instead.
```
PhilipMetzger, line 80: Nit: wrap the `jj restore` part to match the rest.

PhilipMetzger (commented):
This works but I'd rather not see it here, like I've already talked about this morning in Discord. I won't block it, but it still should imo be in the new "freestanding" undo, instead of in the lower-level plumbing.

MrJohz:
> This works but I'd rather not see it here, like I've already talked about this morning in Discord. I won't block it, but it still should imo be in the new "freestanding" undo, instead of in the lower-level plumbing.

I kind of agree, this feels a bit ugly right now because of the "is this `jj undo` or `jj op undo`?" check.  However, it feels worse to just duplicate most of the code between the two commands, because then they'll probably diverge more and they really should be doing the same thing.

Is there an existing way to delegate from one command to some other command?  If there is, and it's not too complicated to get this working with model, I'm happy to try that approach.  But if the alternative is just duplicating the code, then I'd rather just make the small change here.

PhilipMetzger:
> Is there an existing way to delegate from one command to some other command? If there is, and it's not too complicated to get this working with model, I'm happy to try that approach. But if the alternative is just duplicating the code, then I'd rather just make the small change here.

Beside extracting the core out of the current `jj op undo` and calling it from both "freestanding undo" and "op undo" there are not many variants, as making it dependent on a single parameter is probably quite ugly.

arxanas:
I think this is overall a good hint to have (while we don't have a more ergonomic undo solution). It would be fine with me if it were implemented for both `jj undo` and `jj op undo`, but I don't feel strongly.

martinvonz:
Oh, that was a deliberate attempt to change only `jj undo` without changing `jj op undo`? I'd rather keep the same for now. I'd make it a separate change to rename `jj op undo` to `jj op backout`.

arxanas (commented):
(Oops, my inline comment got stuck, something seems wrong with the mobile app.)

cli/src/commands/operation/undo.rs
```diff
@@ -75,6 +75,16 @@ pub fn cmd_op_undo(
...
         template.format(&bad_op, formatter.as_mut())?;
         writeln!(formatter)?;
+
+        let is_bare_undo =
+            command.matches().subcommand_matches("operation").is_none() && args.operation == "@";
+        let was_undo_operation = *tx.repo().view() == parent_op.view()?;
+        if is_bare_undo && was_undo_operation {
+            writeln!(
+                ui.hint_default(),
+                "Consider using `jj restore` if your intention was to undo multiple operations"
```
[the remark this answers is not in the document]
MrJohz: I agree, I struggled to find a good concise way of explaining the problem.  Therefore I've given up on conciseness, let's see if that helps!  ;)  The new text is as follows:

```
Hint: This action reverted an 'undo' operation. The repository is now in the same state as it was before the original 'undo'.
Hint: If your goal is to undo multiple operations, consider using `jj op log` to see past states, and `jj op restore` to restore one of these states.

```

I have some colleagues who haven't really used JJ much before, I'd like to see how they react to this message and if it points them in the right direction.  I should get a chance to do that today or tomorrow.

> comment (non-blocking): I don't know if we have a general framework for defining hints and the configuration to silence them. If we do, then it would be good to use it.

I think the `ui.hint_default()` is the framework as it stands.  This handles e.g. `--quiet` correctly (although this disables most output, as opposed to just disabling hints).  Joining these together, and potentially allowing users to selectively disable them could be a good follow-up task, though.

cli/src/commands/operation/undo.rs
```diff
@@ -75,6 +75,20 @@ pub fn cmd_op_undo(
         let template = tx.base_workspace_helper().operation_summary_template();
         template.format(&bad_op, formatter.as_mut())?;
         writeln!(formatter)?;
+
+        let was_undo_op = *tx.repo().view() == parent_op.view()?;
+        if args.operation == "@" && was_undo_op {
```
MrJohz, line 80: This is a review to discuss the `op undo` vs `undo` question, but I think at this point we're fairly decided: `op undo` and `undo` now behave identically, and both will display the hint.

The hint will still not be displayed if the user attempts to undo a specific operation (e.g. `jj [op] undo 1234abc`), with the assumption that they know what they're doing in that case.

CHANGELOG.md
```diff
@@ -77,6 +77,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj undo` now shows a hint when undoing an undo operation that the user may
+   be looking for `jj restore` instead.
```
cenviity, line 81: `jj restore` restores filesets from a revision. It is probably `jj op restore` instead that we want to suggest here, which restores the repo to an earlier state.

cenviity (changes requested):

cenviity (changes requested):
The commit message should mention `jj op restore` instead of `jj restore` as well.

cli/src/commands/operation/undo.rs
PhilipMetzger: I don't think there's much left to discuss, before either embarking on a new `jj undo` command (#3700), as both Martin and Yuya are fine with the current solutio.

cli/src/commands/operation/undo.rs
```diff
@@ -75,6 +75,20 @@ pub fn cmd_op_undo(
...
+                ui.hint_default(),
+                "This action reverted an 'undo' operation. The repository is now in the same \
+                 state as it was before the original 'undo'."
+            )?;
+            writeln!(
+                ui.hint_default(),
+                "If your goal is to undo multiple operations, consider using `jj op log` to see \
+                 past states, and `jj op restore` to restore one of these states."
+            )?;
+        }
```
yuja, line 91: nit: I would move these hints after `tx.finish()`. Hints are often printed last. It's also better to not make both `ui.status_formatter()` and `ui.hint_default()` alive, while that wouldn't cause problems.

CHANGELOG.md
MrJohz: Yeah, I had a bit of a brainfart with the combination of `jj undo`, `jj op undo` and `jj op restore`.  I think this is now correct in all places.

martinvonz (approved):

CHANGELOG.md
```diff
@@ -77,6 +77,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj undo` now shows a hint when undoing an undo operation that the user may
+   be looking for `jj op restore` instead.
+
```
martinvonz, line 82: Don't forget to move this to "[Unreleased]". I actually don't think we even need to mention it. Users are not going to need to know about it before they run into it.

MrJohz: I've moved the message, but left it in, because I think it's valuable to point out these minor UX changes.  Yes, most users won't notice it until they need it, but I think it's helpful for someone who's already run into this issue to be able to read the changelog and see that if they recommend JJ to someone else, that person _won't_ run into this issue.

cenviity (approved):
LGTM. Thanks!

PR #5584, opened 2025-02-04
templater: add optional ellipsis arg to truncate template functions

### LINKED ISSUE #5085: FR: Allow specifying `tail` for `truncate_*` template functions

steadmon:
If an ellipsis arg is given to the truncate_* template functions, append (or prepend) the ellipsis when the template content is truncated to fit the maximum width.

Fixes #5085.

I went with "ellipsis" rather than "tail", because we already use "ellipsis" for the very similar `elide_*` functions in `text_util.rs`.

I did not use the `elide_*` functions directly, because they assume `&str` content, whereas IIUC, templates can be arbitrary bytes.

To determine whether or not the content should be truncated, I followed the example of the `write_padded_*` functions, which lossily convert the template args to strings to check their width; I assume that's OK, but would appreciate a confirmation.

As you can see in some of the tests, the prepended/appended ellipsis is not colored, even when the truncated output is. I *think* that's OK, but again would appreciate thoughts on the matter. When I tried to make the ellipsis share the coloring of the original content, I ended up duplicating the ellipsis when multiple labels were involved.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* Template functions `truncate_start()` and `truncate_end()` gained an optional
+  `ellipsis` parameter; passing this prepends or appends the ellipsis to the
+  content if it is truncated to fit the maximum width.
+
 ### Fixed bugs
 
 * `jj status` now shows untracked files under untracked directories.
--- a/cli/src/template_builder.rs
+++ b/cli/src/template_builder.rs
@@ -1494,22 +1494,32 @@ fn builtin_functions<'a, L: TemplateLanguage<'a> + ?Sized>() -> TemplateBuildFun
     map.insert(
         "truncate_start",
         |language, diagnostics, build_ctx, function| {
-            let [width_node, content_node] = function.expect_exact_arguments()?;
+            let ([width_node, content_node], [ellipsis_node]) =
+                function.expect_named_arguments(&["", "", "ellipsis"])?;
             let width = expect_usize_expression(language, diagnostics, build_ctx, width_node)?;
             let content =
                 expect_template_expression(language, diagnostics, build_ctx, content_node)?;
-            let template = new_truncate_template(content, width, text_util::write_truncated_start);
+            let ellipsis = ellipsis_node
+                .map(|node| expect_template_expression(language, diagnostics, build_ctx, node))
+                .transpose()?;
+            let template =
+                new_truncate_template(content, ellipsis, width, text_util::write_truncated_start);
             Ok(L::wrap_template(template))
         },
     );
     map.insert(
         "truncate_end",
         |language, diagnostics, build_ctx, function| {
-            let [width_node, content_node] = function.expect_exact_arguments()?;
+            let ([width_node, content_node], [ellipsis_node]) =
+                function.expect_named_arguments(&["", "", "ellipsis"])?;
             let width = expect_usize_expression(language, diagnostics, build_ctx, width_node)?;
             let content =
                 expect_template_expression(language, diagnostics, build_ctx, content_node)?;
-            let template = new_truncate_template(content, width, text_util::write_truncated_end);
+            let ellipsis = ellipsis_node
+                .map(|node| expect_template_expression(language, diagnostics, build_ctx, node))
+                .transpose()?;
+            let template =
+                new_truncate_template(content, ellipsis, width, text_util::write_truncated_end);
             Ok(L::wrap_template(template))
         },
     );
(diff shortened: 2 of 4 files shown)


yuja (commented):
Looks generally good, thanks.

> lossily convert the template args to strings to check their width; I assume that's OK

Yes.

> the prepended/appended ellipsis is not colored, even when the truncated output is. I _think_ that's OK,

Sounds also good. User can apply labels to ellipsis part as needed.

cli/src/template_builder.rs
```diff
@@ -1643,18 +1661,29 @@ where
 
 fn new_truncate_template<'a, W>(
     content: Box<dyn Template + 'a>,
+    ellipsis_char: Option<Box<dyn Template + 'a>>,
```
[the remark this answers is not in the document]
steadmon: Done.

cli/src/text_util.rs
```diff
@@ -230,10 +230,33 @@ fn count_start_zero_width_chars_bytes(text: &[u8]) -> usize {
...
-    let (start, truncated_width) = truncate_start_pos_bytes(data, max_width);
+    let data_width = String::from_utf8_lossy(data).width();
+    let ellipsis_data = recorded_ellipsis_char.data();
+    let ellipsis_width = String::from_utf8_lossy(ellipsis_data).width();
+
+    let (start, truncated_width, truncated) = if data_width > max_width {
+        truncate_start_pos_bytes(data, max_width.saturating_sub(ellipsis_width))
+    } else {
+        truncate_start_pos_bytes(data, max_width)
+    };
```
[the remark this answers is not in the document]
steadmon: Done.

cli/src/text_util.rs
```diff
@@ -230,10 +230,33 @@ fn count_start_zero_width_chars_bytes(text: &[u8]) -> usize {
...
+        // The ellipsis itself may be larger than max_width, so maybe truncate it too.
+        let (start, _, _) = truncate_start_pos_bytes(ellipsis_data, max_width);
+        let truncated_start = start + count_start_zero_width_chars_bytes(&ellipsis_data[start..]);
+        recorded_ellipsis_char.replay_with(formatter, |formatter, range| {
+            let start = cmp::max(range.start, truncated_start);
+            if start < range.end {
+                formatter.write_all(&ellipsis_data[start..range.end])?;
+            }
+            Ok(())
+        })?;
```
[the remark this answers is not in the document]
steadmon: Done.

CHANGELOG.md
```diff
@@ -104,6 +104,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * New template function `pad_centered()` to center content within a minimum width.
 
+* Template functions `truncate_start()` and `truncate_end()` gained an optional
+  `ellipsis` parameter; passing this prepends or appends the ellipsis to the
+  content if it is truncated to fit the maximum width.
+
```
yuja, line 110: reminder: you'll need to relocate changelog entry after the release.

cli/src/text_util.rs
```diff
@@ -230,18 +230,38 @@ fn count_start_zero_width_chars_bytes(text: &[u8]) -> usize {
...
     let data = recorded_content.data();
-    let (start, truncated_width) = truncate_start_pos_bytes(data, max_width);
+    let data_width = String::from_utf8_lossy(data).width();
+    let ellipsis_data = recorded_ellipsis.data();
+    let ellipsis_width = String::from_utf8_lossy(ellipsis_data).width();
+
+    let (start, truncated_width) = if data_width > max_width {
+        truncate_start_pos_bytes(data, max_width.saturating_sub(ellipsis_width))
+    } else {
+        truncate_start_pos_bytes(data, max_width)
```
yuja, line 244: nit: `(0, data_width)` instead of recomputing the width?

cli/src/text_util.rs
```diff
@@ -252,17 +272,36 @@ pub fn write_truncated_start(
...
-        Ok(())
-    })?;
+    let data_width = String::from_utf8_lossy(data).width();
+    let ellipsis_data = recorded_ellipsis.data();
+    let ellipsis_width = String::from_utf8_lossy(ellipsis_data).width();
+
+    let (truncated_end, truncated_width) = if data_width > max_width {
+        truncate_end_pos_bytes(data, max_width.saturating_sub(ellipsis_width))
+    } else {
+        truncate_end_pos_bytes(data, max_width)
```
yuja, line 286: nit: same here. `(data.len(), data_width)`?

cli/src/text_util.rs
```diff
@@ -702,128 +742,356 @@ mod tests {
...
+        insta::assert_snapshot!(
+            format_colored(|formatter| {
+                write_truncated_end(formatter, &recorder, &ellipsis_recorder, 1).map(|_| ())
+            }),
+            @""
+        );
+    }
+
+    #[test]
+    fn test_write_truncated_ellipsis_labeled() {
```
yuja, line 926: Can you add test for truncated ellipsis text (something like `text_elide_start`/`end`)?

BTW, "…" U+2026 is a bit tricky regarding character width. It's unlikely we'll add support for "East Asian Width" compatibility, but it's simpler to not use this character in tests.
https://www.unicode.org/reports/tr11/

yuja (approved):
Thanks.

cli/src/text_util.rs
```diff
@@ -230,18 +230,38 @@ fn count_start_zero_width_chars_bytes(text: &[u8]) -> usize {
...
     let truncated_start = start + count_start_zero_width_chars_bytes(&data[start..]);
-    recorded_content.replay_with(formatter, |formatter, range| {
-        let start = cmp::max(range.start, truncated_start);
-        if start < range.end {
-            formatter.write_all(&data[start..range.end])?;
-        }
-        Ok(())
-    })?;
+    replay_truncated(recorded_content, truncated_start)?;
     Ok(truncated_width)
```
yuja, line 265: nit: oh, and this function is supposed to return the width of the printed characters including ellipsis. Since we don't have any users of this value right now, it's also okay to remove the return value.

CHANGELOG.md
steadmon: Done.

cli/src/text_util.rs
steadmon: Done.

cli/src/text_util.rs
steadmon: Done.

cli/src/text_util.rs
steadmon: Done. Replaced U+2026 with "..", and then the truncation of the ellipsis itself can be handled with the same test cases as before.

cli/src/text_util.rs
steadmon: Done (by incrementing truncated_width by the width of the maybe-truncated ellipsis, rather than by dropping the return value).

steadmon: I thought about adding test cases to check the return value, but it looks like I'd have to either modify or duplicate `format_colored()` to add some assertions, and I'm not sure it's worth the trouble.

Issue #5389, opened 2025-01-18
`jj st` does not show untracked files in directories


necauqua:
## Description

With `snapshot.auto-track` set to `none()` `jj st` seems to only show untracked files at the root of the repository.

## Steps to Reproduce the Problem

```bash
λ jj git init test
λ cd test
λ jj config set --repo snapshot.auto-track 'none()'
λ touch test-file
λ jj st
Untracked paths:
? test-file
Working copy : (m) b91c73c (empty) (no description set)
Parent commit: (z) 0000000 (empty) (no description set)
λ mkdir test-dir
λ touch test-dir/something
λ jj st
Untracked paths:
? test-file
Working copy : (m) b91c73c (empty) (no description set)
Parent commit: (z) 0000000 (empty) (no description set)
```

## Expected Behavior
Second `jj st` should've shown a `? test-dir/something` line, or maybe a `? test-dir` line if the entire dir is untracked.

## Actual Behavior
It doesn't show that, and in fact if `test-file` was not there it would say "The working copy is clean"

## Specifications

- Platform: NixOS
- Version: jj 0.25.0-f7431650bea45f22421951058949ebacabb1de9a

necauqua:
If we do
```bash
λ touch test-dir/tracked
λ jj file track test-dir/tracked
```
The test-dir/something still wont show up

And if we commit the test-dir/tracked it wont show up either

yuja:
`auto-track = none()` makes all directories untracked (or non-auto-tracked to be precise.) New untracked files in those directories aren't caught right now.

BTW, there's another issue that `jj status` won't notice untracked files if watchman is enabled.

martinvonz:
@cstoitner, are you interested in fixing this?

cstoitner:
I had planned to get back to this.

Recursing into untracked directories is not implemented yet. Maybe I can find some time to look at this today.

0xAdk:
Messing around with this I think

```toml
[snapshot]
auto-track = '~glob:**/*'
```

Does what you'd expect `none()` to do, tracking all directories and no files.

PR #5595, opened 2025-02-05
release: 0.26.0

yuja:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,25 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+### Deprecations
+
+### New features
+
+### Fixed bugs
+
+## [0.26.0] - 2025-02-05
+
+### Release highlights
+
+* Improved Git push/fetch compatibility by spawning an external `git` process.
+  This can be enabled by the `git.subprocess=true` config knob, and will be the
+  default in a future release.
+
+* `jj log` can now show cryptographic commit signatures. The output can be
+  controlled by the `ui.show-cryptographic-signatures=true` config knob.
+
+### Breaking changes
+
 * `jj abandon` now deletes bookmarks pointing to the revisions to be abandoned.
   Use `--retain-bookmarks` to move bookmarks backwards. If deleted bookmarks
   were tracking remote bookmarks, the associated bookmarks (or branches) will be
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -112,40 +125,50 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * The `diff.stat()` template method now provides methods to get summary values.
 
-* New `git.sign-on-push` config option to automatically sign commits which are being
-  pushed to a Git remote.
+* Add templater support for rendering commit signatures.
+
+* Add a config option `ui.show-cryptographic-signatures`. When set to `true`,
+  the builtin templates will show signataure information if available. The
```
martinvonz, line 131: nit: typo in "signataure"

CHANGELOG.md
```diff
@@ -88,16 +107,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
 * `jj file list` now supports templating.
 
-* `jj status` now shows untracked files when they reside directly under a tracked directory.
-
-* New `merge-tools.<TOOL>.diff-expected-exit-codes` config option to suppress
-  warnings from tools exiting with non-zero exit codes.
-
-* Add a new template alias `builtin_op_log_oneline` along with `format_operation_oneline` and `format_snapshot_operation_oneline`
+* Add a new template alias `builtin_op_log_oneline` along with
+  `format_operation_oneline` and `format_snapshot_operation_oneline`
```
martinvonz, line 113: nit: The imperative mood here (and on lines 128-134) feels out of place with the rest of the items. We're not telling the reader to add the template alias.

I'm also not sure it's clear to the reader what this means. Maybe something like:

> There is a new `builtin_op_log_oneline` template alias you can pass to `jj op log -T` for a more compact output. You can use  `format_operation_oneline` and `format_snapshot_operation_oneline` to customize parts of it.

CHANGELOG.md
```diff
@@ -112,40 +125,50 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * The `diff.stat()` template method now provides methods to get summary values.
 
-* New `git.sign-on-push` config option to automatically sign commits which are being
-  pushed to a Git remote.
+* Add templater support for rendering commit signatures.
```
martinvonz, line 128: Is this the same as the bullet you added under "Release highlights"? I find that version easier to read. Can we use that here too (instead of the current text)? Or maybe this is redundant with the next bullet so we should drop this one?

CHANGELOG.md
```diff
@@ -112,40 +125,50 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+  `format_short_cryptographic_signature(signature)`.
+
+* New `git.sign-on-push` config option to automatically sign commits which are
+  being pushed to a Git remote.
 
 * New `git.push-new-bookmarks` config option to push new bookmarks without
   `--allow-new`.
 
+* `jj status` now shows untracked files when they reside directly under a
+  tracked directory.
```
martinvonz, line 143: It's probably helpful link to #5389 so the reader doesn't wonder if the described behavior is intentional

CHANGELOG.md
```diff
@@ -171,6 +194,42 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+* Philipp Albrecht (@pylbrecht)
+* Robert Jackson (@rwjblue)
+* Samuel Tardieu (@samueltardieu)
+* Scott Taylor (@scott2000)
+* Stephen Jennings (@jennings)
+* Valentin Gatien-Baron (@v-gb)
+* Vincent Ging Ho Yim (@cenviity)
+* Waleed Khan (@arxanas)
+* Yuya Nishihara (@yuja)
+* blinry (@blinry)
```
martinvonz, line 231: nit: I have sorted names case-insensitively in the past, so I would have put this after Benjamin

martinvonz (approved):
Thank you!

CHANGELOG.md
yuja: Updated, thanks.

CHANGELOG.md
yuja: Copied the highlight line and merged two "new features" items into one.

CHANGELOG.md
yuja: Added `There's still an issue that files under untracked
  directories aren't listed. [#5389](https://github.com/jj-vcs/jj/issues/5389)`

CHANGELOG.md
yuja: Done.

thoughtpolice (approved):

PR #5598, opened 2025-02-05
github: clean up and consolidate the build matrix

thoughtpolice:
This makes the existing build matrix much more "regular" and makes it easier to navigate by consolidating everything under the main build workflow. Here's an example of the results: https://github.com/jj-vcs/jj/actions/runs/13180977716

This requires a mass rename of all the required checks, so @martinvonz will have to take over.

# Checklist

If applicable:


thoughtpolice:
This also contains an update to use macOS-14 on ARM to build x86 binaries to keep the build matrix more regular. It's still faster than Windows. :') I can pull that out, but I think migrating off macos-13 is probably a good idea in general.

.github/workflows/build.yml
```diff
@@ -28,12 +28,28 @@ jobs:
...
+          target: x86_64-unknown-linux-gnu
           cargo_flags: "--all-features"
+        - build: linux-aarch64-gnu
+          os: ubuntu-24.04-arm
+          target: aarch64-unknown-linux-gnu
+          cargo_flags: "--all-features"
+        - build: macos-x86_64
+          os: macos-14
+          target: x86_64-apple-darwin
+          cargo_flags: "--features vendored-openssl"
```
emilazy, line 44: What makes this necessary? Does GitHub Actions ship with a Homebrew‐installed OpenSSL by default or something? (macOS doesn’t have a system OpenSSL that you can actually use without hacks, so I’m wondering about the x86‐64 vs. AArch64 difference here.)

It might be possible to use `/usr/local/bin/brew` to install an x86‐64 OpenSSL, depending on how the GitHub Actions environment is set up.

.github/workflows/build.yml
```diff
@@ -51,64 +67,55 @@ jobs:
...
     - name: Build
       run: cargo build -p jj-cli --no-default-features --verbose
 
+  build-nix:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04, macos-14]
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 15 # NOTE (aseipp): keep in-sync with the build.yml timeout limit
```
emilazy, line 107: Comment could probably use adjusting if this is going in `build.yml`.

.github/workflows/build.yml
thoughtpolice: Basically, there's no x86_64 library to link against on an M1 host, so vendoring it is just the easiest workaround. I actually don't know if OpenSSL comes from an ARM-based brew install, or if the `openssl` crate does some funky shit in order to work with the system-provided libressl, to be fair.

emilazy: It looks like we specifically don’t have Jujutsu depending on OpenSSL at all on Darwin in Nixpkgs, so I wonder why it’s getting pulled in.

thoughtpolice: Yes, that looks like it might be an oversight especially since `libgit2` depends on `openssl`, which is actually why we depend on it at all (so we can in theory get rid of it *entirely* if we drop libgit2, I think!) So the nixpkgs expression might just be a bit wrong.

.github/workflows/build.yml
```diff
@@ -203,8 +234,8 @@ jobs:
         run: uv run -- mkdocs build --strict
 
   # An optional job to alert us when uv updates break the build
-  mkdocs-latest:
-    name: Check that MkDocs can build the docs with latest Python and uv
+  check-mkdocs-latest:
+    name: check (mkdocs latest)
```
ilyagr, line 231: I feel like this rename lost a lot of info. I'm not sure what's best, but let's keep `Check that MkDocs can build the docs` and `Check that MkDocs can build the docs with latest Python and uv` as a comment, at least.

Perhaps there's a better way to share this info, e.g. put it into some sort of "description" field or print it out, but I think a comment is sufficient as a minimal solution.

.github/workflows/build.yml
```diff
@@ -148,8 +146,41 @@ jobs:
           components: rustfmt
       - run: cargo +nightly fmt --all -- --check
 
-  codespell:
-    name: Codespell
+  check-clippy:
```
ilyagr, line 142: (Optional) If you are doing this much reorganizing, perhaps you could split the `check-` checks into a separate file, `.github/workflow/check.yml`? This file is already getting to long, and the `build-` prefix before this jobs doesn't help much.

image

ilyagr (commented):
Thank you for working on this!

.github/workflows/build.yml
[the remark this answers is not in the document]
thoughtpolice: Yes, this is maybe a good idea. I'll think about it...

.github/workflows/build.yml
thoughtpolice: I think that this goes against what I want, though. I want all the information exactly on one page; often when I'm looking at things like build times or errors from builds I want to flip between different things, or I accidentally click on the wrong thing, and that's because they get put in separate workflows (so GitHub's UX shows them as different "groups".)

These checks haven't changed drastically in quite a while and expand slowly; keeping all the information exactly on one page is what I want here. I think renaming things to `test` or something is a good idea though; I'll try to think of something more informative maybe.

.github/workflows/build.yml
thoughtpolice: What even is this testing? What's the backstory on this entry? Could we just always build with the latest version and drop *the other* version?

.github/workflows/build.yml
thoughtpolice: I renamed this to `ci` actually, since I think that's much more on-point for what we're actually trying to convey with this workflow (all of these need to pass to continuously integrate changes into trunk.) I think that's actually quite good and it keeps the build matrix entries shorter.

.github/workflows/build.yml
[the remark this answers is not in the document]
thoughtpolice: See previous comment; this is now `ci`.

.github/workflows/build.yml
ilyagr: There are two reasons for this, the first one being the main one:

- if `uv` (or Python, I guess) has a breaking change, I want to know about it, but I do not want PR authors to be blocked by it. So, I want the "required" check to use a fixed version of everything, and have a non-required check (this one) that checks the latest versions of everything.
- We say we support `uv 0.5.1+` in contributing.md, and I want there to be a check verifying this. As time goes by, I imagine more and more people will be on a non-latest version of `uv`.

.github/workflows/build.yml
ilyagr: ci is definitely better than build, thanks! I'll trust you on not splitting them, my instinct is different, but I haven't thought about it as much.

.github/workflows/build.yml
ilyagr: Perhaps this will be a bit clearer if you add "non-required" to the name of the check. This will also make it easier for Martin (or for future updates of the rules by anyone) to make the correct check required and leave this one optional.

thoughtpolice: Fixed. TBH I feel like having two checks is honestly a bit redundant in a way, but maybe we can address that later; I'd like to get this in sooner than that, though :)

ilyagr (approved):
Looks good, at a glance. Thank you!

PR #5604, opened 2025-02-06
local_working_copy: scan directory entries excluded by auto-track pattern

### LINKED ISSUE #5389: `jj st` does not show untracked files in directories

yuja:
Fixes #5389

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* `jj status` now shows untracked files under untracked directories.
+  [#5389](https://github.com/jj-vcs/jj/issues/5389)
+
 ## [0.26.0] - 2025-02-05
 
 ### Release highlights
--- a/cli/tests/test_status_command.rs
+++ b/cli/tests/test_status_command.rs
@@ -377,60 +377,89 @@ fn test_status_untracked_files() {
     test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "repo"]);
     let repo_path = test_env.env_root().join("repo");
 
-    std::fs::write(repo_path.join("initially-untracked-file"), "...").unwrap();
     std::fs::write(repo_path.join("always-untracked-file"), "...").unwrap();
+    std::fs::write(repo_path.join("initially-untracked-file"), "...").unwrap();
+    std::fs::create_dir(repo_path.join("sub")).unwrap();
+    std::fs::write(repo_path.join("sub").join("always-untracked"), "...").unwrap();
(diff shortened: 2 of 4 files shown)


thoughtpolice (approved):

jakobhellermann:
I have my home dir as a jj repo with `snapshot.auto-track = 'none()'`, and this change made it so that running jj starts `Snapshotting ...` for a bunch of folders in my `~` and eventually fails with

```
Internal error: Failed to snapshot the working copy
Caused by:
1: Failed to read directory /home/jakob/.local/share/containers/storage/volumes/861e72b0da83ff1cee691c51d408c5ca4e2c205a96dc3cc2a530d5a94e43dfb0
2: Permission denied (os error 13)
```

I suppose it is impossible to support both
- listing untracked files in untracked folders
- not walking the entire universe of folders

so you have to decide which one you value more.

I don't thinking having a `*` in `~/.gitignore` is a solution, because you can't override `jj file track` a gitignore'd file.

martinvonz:
Maybe we could report `.local/` as untracked without walking it recursively?

yuja:
If there are no tracking entries starting with `.local/`, yes, we can short-circuit and report all files in `.local` are untracked. I didn't add such optimization because the purpose of `auto-track` option is afaik to avoid auto-add of temporary files, etc., not to avoid permission errors or ignore files permanently.

Maybe you can add `~/.local/.gitignore` to ignore all files in `~/.local`?

> because you can't override `jj file track` a gitignore'd file.

We'll need some command flag to track ignored files.

martinvonz:
> I didn't add such optimization because the purpose of `auto-track` option is afaik to avoid auto-add of temporary files, etc., not to avoid permission errors or ignore files permanently.

Right, I should have remembered that. I agree with you.

Issue #405, opened 2022-06-30
Integrate with https://pre-commit.com/ 


martinvonz:
## Description

I don't think I had seen https://pre-commit.com/ until @chandlerc [linked to it](https://github.com/martinvonz/jj/discussions/403#discussioncomment-3052651). It would be really nice to integrate with that somehow. I currently have a long command for running Clippy and `rustfmt` on my commits. It would be better if we could use pre-commit.com for sharing that between users and making it easy to run. I don't yet know how it would work.

Mercurial's [fix extension](https://www.mercurial-scm.org/repo/hg/file/tip/hgext/fix.py) seems closely related. IIUC, pre-commit.com runs all its checks in the working copy (possibly after stashing any unstaged changes), so that's a big difference compared to `hg fix`. I think it would be nice if we could run the pre-commits on all commits in parallel just like `hg fix` does, but that's slow to do if the working copy is large because we'd need to create a working copy for each commit (unless/until we can expose a commit in a virtual file system). That's why `hg fix` doesn't do it. A good start might be if we can figure out a way of running the pre-commit.com checks in the working copy.

chandlerc:
In case its useful, a collection of useful https://pre-commit.com/ hooks is here: https://github.com/google/pre-commit-tool-hooks

Maybe some even useful for working with this repo! =D

madjar:
I looked into this a bit, and I've found two potential ways to interface with pre-commit.

The first one is to use the `--from-ref` and `--to-ref` arguments of `pre-commit run` to specify a range of commits to use.

The other one is to pass it an explicit list of files with `--files`. Using this, one can run pre-commit on the file from the last revision (just after a `jj commit` for example), with this monstrosity (the python part extracts the list of files):
```
jj show @- -s| python3 -c 'import sys; print(" ".join(l[2:] for l in sys.stdin.read().split("\n\n")[-1].split("\n")))' | xargs pre-commit run --files
```

---
There's one hook that's not behaving well (darker), but that's because the program itself is looking at the git index, so I might have to find another trick to convince it to do what I want.

martinvonz:
I think `--from-ref` and `--to-ref` implies an underlying Git repo. I think it would be nice if we could make it independent of storage backend. That's not terribly important in practice since no one uses jj's native backend yet (and they shouldn't).

[...]

PR #5612, opened 2025-02-06
git: skip running pre-push hooks when subprocessing to match git2 behaviour

### LINKED ISSUE #5610: Git operations with `git.subprocess=true` fail if hook prints unexpected text
JJ should ignore hooks when `git.subprocess=true`, since some of my Git hooks output information.

Example:

```
$ jj git push
Changes to push to origin:
  Move sideways bookmark zx8/ntmyzyrvtzwu from 7de2273ab713 to 1febf55ed2b2
remote: Resolving deltas: 100% (3/3), completed with 3 local objects.
Error: Git process failed: Git push output unfamiliar:

 ␛[1;97mhttps://github.com/zx8/jj/commit/1febf55␛[0m

To github.com:zx8/jj.git
+ 1febf55ed2b2a07dc1aaed40bcaed10f672ec532:refs/heads/zx8/ntmyzyrvtzwu 7de2273...1febf55 (forced update)
Done

```

I think disabling Git hooks via the following might help:

```
git -c core.hooksPath=/dev/null
```

bsdinis:
This fixes #5610

### DIFF
--- a/lib/src/git_subprocess.rs
+++ b/lib/src/git_subprocess.rs
@@ -205,7 +205,12 @@ impl<'a> GitSubprocessContext<'a> {
     ) -> Result<(Vec<String>, Vec<String>), GitSubprocessError> {
         let mut command = self.create_command();
         command.stdout(Stdio::piped());
-        command.args(["push", "--porcelain"]);
+        // Currently jj does not support commit hooks, so we prevent git from running
+        // them
+        //
+        // https://github.com/jj-vcs/jj/issues/3577 and https://github.com/jj-vcs/jj/issues/405
+        // offer more context
+        command.args(["push", "--porcelain", "--no-verify"]);
         if callbacks.progress.is_some() {
             command.arg("--progress");
         }


lib/src/git_subprocess.rs
```diff
@@ -205,7 +205,8 @@ impl<'a> GitSubprocessContext<'a> {
     ) -> Result<(Vec<String>, Vec<String>), GitSubprocessError> {
         let mut command = self.create_command();
         command.stdout(Stdio::piped());
-        command.args(["push", "--porcelain"]);
+        // we use --no-verify to skip the pre-push hooks
```
arxanas, line 208: suggestion: Do we actually _want_ to skip the pre-push hooks in the long-term? I had imagined that we did want to run them. Let's update the comment to explain whether or not this is the end goal.

PhilipMetzger: See the discussion in Discord.

PhilipMetzger: I argue against this until #405 and #3577 are somewhat solved.

bsdinis: I think we do want to run pre-push hooks in the long term. However, it seems that there are some design considerations to take into account.

I am happy to expand on this in the comment, but since this is a bug that has hit our users, we should merge the solution that matches the behaviour that exists elsewhere

martinvonz (approved):
LGTM, but let's give @arxanas a chance to object before merging.

arxanas (approved):
To be clear, for this PR, I just wanted to update the comment to specify that it's an intermediate state of indecision rather than the ultimate state, rather than make any specific decision about `pre-push` hooks right now. It looks good to me now that the comment includes references to the relevant issues.

arxanas:
> Let's update the comment to explain whether or not this is the end goal.

Oops, I see now that I didn't write what I wanted to write. Originally, I wrote something like "let's update the comment to say that this is not the end goal", but then I realized we technically did not have agreement on whether it was or wasn't the end goal, so I tried to adjust the wording, and did so wrongly.

Issue #3402, opened 2024-03-30
FR: Topics


noahmayr:
**Is your feature request related to a problem? Please describe.**
I read through several issues and saw some plans/discussions (https://github.com/martinvonz/jj/discussions/2425#discussioncomment-7376935,  about a "topics" feature. However I wasn't able to find a dedicated issue for this. This issue contains my thoughts on how topics could work from the perspective of a user that has no prior experience with mercurial's topics.

**Describe the solution you'd like**
There's several things that would go into this feature in my opinion:

1. Working on a topic
`jj new --topic my_feature 'trunk()'` / `jj new -t my_feature` will create a new change associated with the topic `my_feature`
Running `jj new` , `jj commit`, etc. on a revision that is already associated with a topic, will result in the next change and also being associated with the topic.

2. Pushing changes made on the topic
`jj git push --topic my_feature` if currently working on a topic, it should suffice to pass `--topic`/'-t'
This would then create a branch with a name based on that topic or move that branch to the head of the topic and push it

3. Being able to specify whether to push a single branch for a topic or one per change
`jj git push --topic --mode=single` would work like `jj git push change=<head of topic>` with the branch name being set to the name of the topic
`jj git push --topic --mode=stacked` would create individual branches for each change of the topic named `<topic>-<change_id>`
I think we can come up with a better name than `mode` for the flag, but this would work with and without specifying a topic. When no topic is specified, we can use a revset instead and potentially default that to `trunk()..@-`.

4. A better way to define what automatically created branches will be named
Instead of using `git.push-branch-prefix`, we could use a configurable template for auto created branches:
`git.branch_name_template = 'if(mode == 'stacked', separate("-", if(topic, topic, "push"), change_id), topic)'`

5. Being able to associate an existing (consecutive?) set of changes with a topic:
`jj topic set my_feature -r 'trunk()..@'` (not sure if there's a good existing command to add this into)

6. Extending the revset language to include topics
Topics would most likely be a first class symbol in the revset language. Not sure where exactly they'd end up in the priority order though.
Apart from that, a `topics([pattern])` function analog to the one for branches should be added.

**Describe alternatives you've considered**
Alternatives would be #2338 for being able to work with branches more effectively, but I think topics could provide a more idiomatic jj approach while still providing great interop with branches.

Since we may want a topic command anyways (see 5.), there could be several other sub commands of it:
- `jj topic start my_feature -r 'trunk()'` (the revision to start from could either default to the current head or trunk).
[...]

Issue #2338, opened 2023-10-05
FR: `--advance-branches` flag for `jj commit`


emilykfox:
**Is your feature request related to a problem? Please describe.**
 I often work directly on `main` for personal projects, but to successfully push my changes to GitHub, I have to run something like `jj commit -m [message] && jj branch set main -r @-` or `jj branch set main && jj commit -m [message]`. The former is error prone as it's easy to forget the `-r` flag. The latter feels slightly unnatural as there's a brief moment where the `main` branch is pointing to something without a description. They're both pretty verbose.

**Describe the solution you'd like**
In Discord, @martinvonz  suggested adding a `--advance-branches` flag to `jj commit` which could move any branches pointing to the old `@-` to point to the new `@-`.

**Describe alternatives you've considered**
None

**Additional context**
See the brief conversation starting at https://discord.com/channels/968932220549103686/969829516539228222/1159510245849186335.

Probably more important is that this feature would also improve workflows that involve adding several commits in sequence to a single feature branch.

ilyagr:
I'd like this too.

Some miscellaneous thoughts:

1. I think `jj new` could also have such a flag, for moving branches from whichever commit `jj new` makes the parent of the new commit (usually `@`).
2. This is minor, but I think `jj commit --advance-branches` should also move branches from old `@` to new `@` (if any).
3. (Not for the first version) I wish there was some way to specify the branches that define immutable commits (e.g. `main`), so that `jj commit --advance-branches` would know not to move these branches. This would make defining the immutable commits revset harder, though, since it's actually defined by `main@origin` and `main@upstream`.

necauqua:
I think I've advocated in the past that `jj commit` should just do that - makes sense as commit is "an alias to be familiar for git users, as well as just a shortcut for new+describe"

Maybe there should be an option for `jj new`, and then we describe commit as an alias for new with such flag+describe - again, makes total sense as it is just for more git-like workflow

martinvonz:
I'm concerned that not all users who do `jj new main; <edit files>; jj commit` want to move `main` forward. Some prefer to work with anonymous branches (aka detached heads) but still like to use `jj commit` as a `describe+new` shortcut.

necauqua:
Well, I now realised that if we just have the flag (probably on both commands), then I can alias it anyway

Especially since in personal repos I do anonymous heads and at work I have to maintain feature branches :)

ilyagr:
[...]

PR #5618, opened 2025-02-07
cli split: Add a config option controlling how bookmarks move during splits

### LINKED ISSUE #3419: Should `jj split` leave branches on the first commit instead of moving to the second?

emesterhazy:
Currently, `jj split` moves bookmarks from the target revision to the second
revision created by the split. Since the first revision inherits the change id
of the target revision, moving the bookmarks to the first revision is less
surprising (i.e. the bookmarks stay with the change id). This no-implicit-move
behavior also aligns with how `jj abandon` drops bookmarks instead of moving
them to the parent revision.

Two releases from now, `jj split` will no longer move bookmarks to the second
revision created by the split. Instead, local bookmarks associated with the
target revision will move to the first revision created by the split (which
inherits the target revision's change id). You can opt out of this change by
setting `split.legacy-bookmark-behavior = true`, but this will likely be
removed in a future release. You can also try the new behavior now by setting
`split.legacy-bookmark-behavior = false`.

Users who have not set the new config option will see a warning when they run
`jj split` informing them about the change.

The `jj split` tests for bookmarks are updated to run in all three configurations:

- Config setting enabled
- Config setting disabled
- Config setting unset

# Checklist

If applicable:
    - N/A: *I didn't see any docs describing the bookmark semantics.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,16 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   a warning if the user does not specify target revision  explicitly. In the near
   future those commands will fail if target revision is not specified.
 
+* In the next release, `jj split` will no longer move bookmarks to the second
+  revision created by the split. Instead, local bookmarks associated with the
+  target revision will move to the first revision created by the split (which
+  inherits the target revision's change id). You can opt out of this change by
+  setting `split.legacy-bookmark-behavior = true`, but this will likely be
+  removed in a future release. You can also try the new behavior now by setting
+  `split.legacy-bookmark-behavior = false`.
+  [#3419](https://github.com/jj-vcs/jj/issues/3419)
+
+
 ### New features
 
 * `jj undo` now shows a hint when undoing an undo operation that the user may
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -15,6 +15,7 @@ use std::io::Write;
 
 use clap_complete::ArgValueCandidates;
 use clap_complete::ArgValueCompleter;
+use jj_lib::config::ConfigGetResultExt;
 use jj_lib::object_id::ObjectId;
 use jj_lib::repo::Repo;
 use tracing::instrument;
@@ -196,26 +197,37 @@ The remainder will be in the second commit.
         commit_builder.write(tx.repo_mut())?
     };
 
-    // Mark the commit being split as rewritten to the second commit. As a
-    // result, if @ points to the commit being split, it will point to the
-    // second commit after the command finishes. This also means that any
-    // bookmarks pointing to the commit being split are moved to the second
-    // commit.
-    tx.repo_mut()
-        .set_rewritten_commit(commit.id().clone(), second_commit.id().clone());
+    let legacy_bookmark_behavior = read_legacy_bookmark_behavior_setting(tx.settings(), ui)?;
+    if legacy_bookmark_behavior {
+        // Mark the commit being split as rewritten to the second commit. This
+        // moves any bookmarks pointing to the target commit to the second
+        // commit.
+        tx.repo_mut()
+            .set_rewritten_commit(commit.id().clone(), second_commit.id().clone());
+    }
     let mut num_rebased = 0;
     tx.repo_mut()
         .transform_descendants(vec![commit.id().clone()], |mut rewriter| {
             num_rebased += 1;
-            if args.parallel {
+            if args.parallel && legacy_bookmark_behavior {
+                // The old_parent is the second commit due to the rewrite above.
                 rewriter
                     .replace_parent(second_commit.id(), [first_commit.id(), second_commit.id()]);
+            } else if args.parallel {
+                rewriter.replace_parent(first_commit.id(), [first_commit.id(), second_commit.id()]);
+            } else {
+                rewriter.replace_parent(first_commit.id(), [second_commit.id()]);
             }
-            // We don't need to do anything special for the non-parallel case
-            // since we already marked the original commit as rewritten.
             rewriter.rebase()?.write()?;
             Ok(())
         })?;
+    // Move the working copy commit (@) to the second commit for any workspaces
+    // where the target commit is the working copy commit.
+    for (workspace_id, working_copy_commit) in tx.base_repo().clone().view().wc_commit_ids() {
+        if working_copy_commit == commit.id() {
+            tx.repo_mut().edit(workspace_id.clone(), &second_commit)?;
+        }
+    }
 
     if let Some(mut formatter) = ui.status_formatter() {
         if num_rebased > 0 {
@@ -229,4 +241,37 @@ The remainder will be in the second commit.
     }
     tx.finish(ui, format!("split commit {}", commit.id().hex()))?;
     Ok(())
+}
+
+/// Reads 'split.legacy-bookmark-behavior' from settings and prints a warning if
+/// the value is unset to alert the user to the behavior change.
+fn read_legacy_bookmark_behavior_setting(
+    settings: &jj_lib::settings::UserSettings,
+    ui: &Ui,
+) -> Result<bool, CommandError> {
+    match settings
+        .get_bool("split.legacy-bookmark-behavior")
+        .optional()?
+    {
+        // Use the new behavior.
+        Some(false) => Ok(false),
+        // Use the legacy behavior.
+        Some(true) | None => {
+            writeln!(
+                ui.warning_default(),
+                "`jj split` will leave bookmarks on the first commit in the next release."
+            )?;
+            writeln!(
+                ui.warning_default(),
+                "Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this \
+                 message and use the new behavior."
+            )?;
+            writeln!(
+                ui.warning_default(),
+                "See https://github.com/jj-vcs/jj/issues/3419"
+            )?;
+            // TODO: https://github.com/jj-vcs/jj/issues/3419 - Return false by default in v0.28.
+            Ok(true)
+        }
+    }
 }
--- a/cli/tests/test_split_command.rs
+++ b/cli/tests/test_split_command.rs
@@ -15,6 +15,8 @@
 use std::path::Path;
 use std::path::PathBuf;
 
+use test_case::test_case;
+
 use crate::common::TestEnvironment;
 
 fn get_log_output(test_env: &TestEnvironment, cwd: &Path) -> String {
@@ -60,6 +62,9 @@ fn test_split_by_paths() {
     let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["split", "file2"]);
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     First part: qpvuntsm 65569ca7 (no description set)
     Second part: zsuskuln 709756f0 (no description set)
     Working copy now at: zsuskuln 709756f0 (no description set)
@@ -107,14 +112,17 @@ fn test_split_by_paths() {
     test_env.set_up_fake_editor();
     let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["split", "-r", "@-", "."]);
     insta::assert_snapshot!(stdout, @"");
-    insta::assert_snapshot!(stderr, @r#"
+    insta::assert_snapshot!(stderr, @r###"
     Warning: All changes have been selected, so the second commit will be empty
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     Rebased 1 descendant commits
     First part: qpvuntsm 9da0eea0 (no description set)
     Second part: znkkpsqq 5b5714a3 (empty) (no description set)
     Working copy now at: zsuskuln 0c798ee7 (no description set)
     Parent commit      : znkkpsqq 5b5714a3 (empty) (no description set)
-    "#);
+    "###);
 
     insta::assert_snapshot!(get_log_output(&test_env, &repo_path), @r###"
     @  zsuskulnrvyr false
@@ -135,14 +143,17 @@ fn test_split_by_paths() {
     test_env.set_up_fake_editor();
     let (stdout, stderr) = test_env.jj_cmd_ok(&repo_path, &["split", "-r", "@-", "nonexistent"]);
     insta::assert_snapshot!(stdout, @"");
-    insta::assert_snapshot!(stderr, @r#"
+    insta::assert_snapshot!(stderr, @r###"
     Warning: No changes have been selected, so the first commit will be empty
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     Rebased 1 descendant commits
     First part: qpvuntsm bd42f95a (empty) (no description set)
     Second part: lylxulpl ed55c86b (no description set)
     Working copy now at: zsuskuln 1e1ed741 (no description set)
     Parent commit      : lylxulpl ed55c86b (no description set)
-    "#);
+    "###);
 
     insta::assert_snapshot!(get_log_output(&test_env, &repo_path), @r###"
     @  zsuskulnrvyr false
@@ -183,6 +194,9 @@ fn test_split_with_non_empty_description() {
     let (stdout, stderr) = test_env.jj_cmd_ok(&workspace_path, &["split", "file1"]);
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     First part: qpvuntsm 231a3c00 part 1
     Second part: kkmpptxz e96291aa part 2
     Working copy now at: kkmpptxz e96291aa part 2
@@ -235,6 +249,9 @@ fn test_split_with_default_description() {
     let (stdout, stderr) = test_env.jj_cmd_ok(&workspace_path, &["split", "file1"]);
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     First part: qpvuntsm 02ee5d60 TESTED=TODO
     Second part: rlvkpnrz 33cd046b (no description set)
     Working copy now at: rlvkpnrz 33cd046b (no description set)
@@ -300,6 +317,9 @@ fn test_split_with_merge_child() {
         test_env.jj_cmd_ok(&workspace_path, &["split", "-r", "description(a)", "file1"]);
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     Rebased 1 descendant commits
     First part: kkmpptxz e8006b47 Add file1
     Second part: royxmykx 5e1b793d Add file2
@@ -344,6 +364,9 @@ fn test_split_siblings_no_descendants() {
     let (stdout, stderr) = test_env.jj_cmd_ok(&workspace_path, &["split", "--parallel", "file1"]);
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     First part: qpvuntsm 48018df6 TESTED=TODO
     Second part: kkmpptxz 7eddbf93 (no description set)
     Working copy now at: kkmpptxz 7eddbf93 (no description set)
@@ -422,6 +445,9 @@ fn test_split_siblings_with_descendants() {
     let (stdout, stderr) = test_env.jj_cmd_ok(&workspace_path, &["split", "--parallel", "file1"]);
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     Rebased 2 descendant commits
     First part: qpvuntsm 84df941d Add file1
     Second part: vruxwmqv 94753be3 Add file2
@@ -500,6 +526,9 @@ fn test_split_siblings_with_merge_child() {
     );
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     Rebased 1 descendant commits
     First part: kkmpptxz e8006b47 Add file1
     Second part: royxmykx 2cc60f3d Add file2
@@ -596,6 +625,9 @@ fn test_split_interactive() {
 
     insta::assert_snapshot!(stdout, @"");
     insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     First part: qpvuntsm 0e15949e (no description set)
     Second part: rlvkpnrz 9ed12e4c (no description set)
     Working copy now at: rlvkpnrz 9ed12e4c (no description set)
@@ -640,12 +672,15 @@ fn test_split_interactive_with_paths() {
 
     // Select file1 and file2 by args, then select file1 interactively
     let (_stdout, stderr) = test_env.jj_cmd_ok(&workspace_path, &["split", "-i", "file1", "file2"]);
-    insta::assert_snapshot!(stderr, @r"
+    insta::assert_snapshot!(stderr, @r###"
+    Warning: `jj split` will leave bookmarks on the first commit in the next release.
+    Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+    Warning: See https://github.com/jj-vcs/jj/issues/3419
     First part: rlvkpnrz e3d766b8 (no description set)
     Second part: kkmpptxz 4cf22d3b (no description set)
     Working copy now at: kkmpptxz 4cf22d3b (no description set)
     Parent commit      : rlvkpnrz e3d766b8 (no description set)
-    ");
+    "###);
 
     insta::assert_snapshot!(
         std::fs::read_to_string(test_env.env_root().join("editor")).unwrap(), @r###"
@@ -797,21 +832,42 @@ fn test_split_with_multiple_workspaces_different_working_copy() {
     "###);
 }
 
-#[test]
-fn test_split_with_bookmarks() {
+enum BookmarkBehavior {
+    Default,
+    Legacy,
+    Modern,
+}
+
+// TODO: https://github.com/jj-vcs/jj/issues/3419 - Delete params when the config is removed.
+#[test_case(BookmarkBehavior::Default; "default_behavior")]
+#[test_case(BookmarkBehavior::Legacy; "legacy_behavior")]
+#[test_case(BookmarkBehavior::Modern; "modern_behavior")]
+fn test_split_with_bookmarks(bookmark_behavior: BookmarkBehavior) {
     let mut test_env = TestEnvironment::default();
     test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "main"]);
     let main_path = test_env.env_root().join("main");
+
+    match bookmark_behavior {
+        BookmarkBehavior::Modern => {
+            test_env.add_config("split.legacy-bookmark-behavior=false");
+        }
+        BookmarkBehavior::Legacy => {
+            test_env.add_config("split.legacy-bookmark-behavior=true");
+        }
+        BookmarkBehavior::Default => (),
+    }
 
     // Setup.
     test_env.jj_cmd_ok(&main_path, &["desc", "-m", "first-commit"]);
     std::fs::write(main_path.join("file1"), "foo").unwrap();
     std::fs::write(main_path.join("file2"), "foo").unwrap();
     test_env.jj_cmd_ok(&main_path, &["bookmark", "set", "*le-signet*", "-r", "@"]);
+    insta::allow_duplicates! {
     insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r###"
     @  qpvuntsmwlqt false *le-signet* first-commit
     ◆  zzzzzzzzzzzz true
     "###);
+    }
 
     // Do the split.
     std::fs::write(
@@ -819,13 +875,46 @@ fn test_split_with_bookmarks() {
         ["", "next invocation\n", "write\nsecond-commit"].join("\0"),
     )
     .unwrap();
-    test_env.jj_cmd_ok(&main_path, &["split", "file2"]);
-    // The bookmark moves to the second commit.
-    insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r###"
-    @  mzvwutvlkqwt false *le-signet* second-commit
-    ○  qpvuntsmwlqt false first-commit
-    ◆  zzzzzzzzzzzz true
-    "###);
+    let (_, stderr) = test_env.jj_cmd_ok(&main_path, &["split", "file2"]);
+    match bookmark_behavior {
+        BookmarkBehavior::Modern => {
+            insta::allow_duplicates! {
+            insta::assert_snapshot!(stderr, @r###"
+            First part: qpvuntsm 63d0c5ed *le-signet* | first-commit
+            Second part: mzvwutvl a9f5665f second-commit
+            Working copy now at: mzvwutvl a9f5665f second-commit
+            Parent commit      : qpvuntsm 63d0c5ed *le-signet* | first-commit
+            "###);
+            }
+            insta::allow_duplicates! {
+            insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r###"
+            @  mzvwutvlkqwt false second-commit
+            ○  qpvuntsmwlqt false *le-signet* first-commit
+            ◆  zzzzzzzzzzzz true
+            "###);
+            }
+        }
+        BookmarkBehavior::Default | BookmarkBehavior::Legacy => {
+            insta::allow_duplicates! {
+            insta::assert_snapshot!(stderr, @r###"
+            Warning: `jj split` will leave bookmarks on the first commit in the next release.
+            Warning: Run `jj config set --repo split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
+            Warning: See https://github.com/jj-vcs/jj/issues/3419
+            First part: qpvuntsm 63d0c5ed first-commit
+            Second part: mzvwutvl a9f5665f *le-signet* | second-commit
+            Working copy now at: mzvwutvl a9f5665f *le-signet* | second-commit
+            Parent commit      : qpvuntsm 63d0c5ed first-commit
+            "###);
+            }
+            insta::allow_duplicates! {
+            insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r###"
+            @  mzvwutvlkqwt false *le-signet* second-commit
+            ○  qpvuntsmwlqt false first-commit
+            ◆  zzzzzzzzzzzz true
+            "###);
+            }
+        }
+    }
 
     // Test again with a --parallel split.
     test_env.jj_cmd_ok(&main_path, &["undo"]);
@@ -835,11 +924,26 @@ fn test_split_with_bookmarks() {
     )
     .unwrap();
     test_env.jj_cmd_ok(&main_path, &["split", "file2", "--parallel"]);
-    // The bookmark moves to the second commit.
-    insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r###"
-    @  vruxwmqvtpmx false *le-signet* second-commit
-    │ ○  qpvuntsmwlqt false first-commit
-    ├─╯
-    ◆  zzzzzzzzzzzz true
-    "###);
+    match bookmark_behavior {
+        BookmarkBehavior::Modern => {
+            insta::allow_duplicates! {
+            insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r###"
+            @  vruxwmqvtpmx false second-commit
+            │ ○  qpvuntsmwlqt false *le-signet* first-commit
+            ├─╯
+            ◆  zzzzzzzzzzzz true
+            "###);
+            }
+        }
+        BookmarkBehavior::Default | BookmarkBehavior::Legacy => {
+            insta::allow_duplicates! {
+            insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r###"
+            @  vruxwmqvtpmx false *le-signet* second-commit
+            │ ○  qpvuntsmwlqt false first-commit
+            ├─╯
+            ◆  zzzzzzzzzzzz true
+            "###);
+            }
+        }
+    }
 }
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -684,6 +684,17 @@
                     "description": "Settings for tools run by jj fix"
                 }
             }
+        },
+        "split": {
+            "type": "object",
+            "description": "Settings for jj split",
+            "properties": {
+                "legacy-bookmark-behavior": {
+                    "type": "boolean",
+                    "description": "If true, bookmarks will move to the second commit instead of the first.",
+                    "default": false
+                }
+            }
         }
     }
 }


cli/src/commands/split.rs
```diff
@@ -227,6 +229,10 @@ The remainder will be in the second commit.
         tx.write_commit_summary(formatter.as_mut(), &second_commit)?;
         writeln!(formatter)?;
     }
+    // If @ refers to the target commit, move it to the second commit.
+    if at.id() == commit.id() {
+        tx.edit(&second_commit)?;
+    }
```
martinvonz, line 235: This will update working copies in other workspaces. I think we should update the code here to use `rewrite::move_commits()` in order to move the children of `commit` (i.e. before splitting) to be rebased on top of `second_commit`.

PhilipMetzger: it would also be nice if you spelled out "the working copy commit (@)" in the comment

PhilipMetzger (approved):
Thanks for doing this, lgtm. I will leave the last call to Martin.

cli/src/commands/split.rs
emesterhazy: I'm not sure I understand. How will using `rewrite::move_commits()` avoid the issue of `edit` updating the working copy in other workspaces? I need to read up on how workspaces work, but I'm a little surprised that `tx.edit` updates workspaces other than the current one.

The code suggests it only acts on the current workspace via the `workspace_id`?

```rust
    pub fn edit(&mut self, commit: &Commit) -> Result<(), EditCommitError> {
        let workspace_id = self.helper.workspace_id().to_owned();
        self.id_prefix_context.take(); // invalidate
        self.tx.repo_mut().edit(workspace_id, commit)
    }
```

Separately, what should I pass to the `targets` argument of `rewrite::move_commits()`? I don't think I have a list of the children of the commit being split. Do I need to evaluate a revset, or is there another way to do it?

martinvonz: > How will using `rewrite::move_commits()` avoid the issue of `edit` updating the working copy in other workspaces?

Oh, you're right, it won't. I was thinking that this was a workaround for how we rewrite commits in a strange way above, but it's just a consequence of wanting the working copy to move to the second commit.

> but I'm a little surprised that `tx.edit` updates workspaces other than the current one.

It doesn't. That's the problem, kind of. If you had split a commit checked out by another workspace, then that workspace should now point to the rewritten commit.

I think the problem is that `split` is a bit unusual in how it rewrites the graph:

1. The change id should stay on the first commit
2. Bookmarks should stay on the first commit
3. Working copies should move to the second commit
4. Descendants should move onto the second commit

Normally they all move in the same way, and that's what building blocks like `rebase_descendants()` expects. But because we don't want that here, we have to do something more manually. Moving the working-copy pointers manually as you did makes sense. We should just update the code to update all working copies instead of just the current working copy.

Sorry about the confusion.

emesterhazy: Thanks for the explanation. I think I have everything working as intended now. I added two new tests for splitting with multiple workspaces, one where the workspaces have the same working copy commit which is also the target of the split, and one where they don't.

Please take another look when able :)

CHANGELOG.md
ilyagr, line 1: If we want to make this more consistent, I would much rather have the change id and the bookmark stay with the second revision. I would very much like to have ways to make a new commit and have branches slide down to it, and currently jj split is that way (either after making some changes, or `jj split 'all()'` to keep all files in the first commit).

Or have people already decided they prefer this way?

ilyagr: Defaulting to keeping bookmarks at the first revision might make sense when we have a good solution for topics. Until then, I'd rather default to sliding them down. Philip said somewhere that this is bad for `jj gerrit push`, which I don't fully follow, but sliding down is what I want almost always for `jj git push` and feature branches.

cli/src/commands/split.rs
```diff
@@ -196,26 +200,37 @@ The remainder will be in the second commit.
...
-    // second commit after the command finishes. This also means that any
-    // bookmarks pointing to the commit being split are moved to the second
-    // commit.
+    // Mark the first commit as the rewrite of the target commit. Since the
+    // first commit inherits the change id of the target commit, this ensures
+    // bookmarks continue pointing to the same change id before and after the
+    // split.
     tx.repo_mut()
-        .set_rewritten_commit(commit.id().clone(), second_commit.id().clone());
+        .set_rewritten_commit(commit.id().clone(), first_commit.id().clone());
```
yuja, line 208: nit: this can be removed. It's the default behavior of`.rewrite_commit()`

cli/src/commands/split.rs
```diff
@@ -196,26 +200,37 @@ The remainder will be in the second commit.
...
             }
-            // We don't need to do anything special for the non-parallel case
-            // since we already marked the original commit as rewritten.
             rewriter.rebase()?.write()?;
             Ok(())
         })?;
+    // Move the working copy commit (@) to the second commit for any workspaces
+    // where the target commit is the working copy commit.
+    for workspace_id in tx
+        .base_repo()
```
yuja, line 224: nit: `.base_repo().clone()` also works in place of `.to_owned()` + `.collect_vec()`. Cloning `Arc` should be cheap.

CHANGELOG.md
ilyagr: It is unfortunate that the better default seems to be backend-dependent. I see how sliding bookmarks is inconvenient if bookmarks represent CL numbers or whatever the gerrit equivalent is.

We could also ship and advertise a built-in alias for the non-default option. If we can come up with a good name, it matters less which is the default.

But really, we need different bottoms for the different kinds of backend ids.

cli/src/commands/split.rs
emesterhazy: Thanks, fixed!

CHANGELOG.md
emesterhazy: IMO this isn't really about the backend/forge you use, it's about the workflow. The workflow is of course influenced by the backend/forge you use, but the current behavior is problematic even on GitHub if you use a stacked PR workflow.

I think the UX will be more consistent if we move to a place where bookmarks are always explicitly moved by the user. I agree though that we need to provide a convenient way to slide bookmarks forwards. It seems like some people are doing this with a `jj bookmark move` alias (see https://github.com/jj-vcs/jj/issues/2338#issuecomment-2477654445). I wonder if we could turn this into a first-party command and smooth out the rough edges (what to do if there are multiple bookmarks equidistant from the target). That would offer a low friction way for users to slide bookmarks if they want to before pushing to a git forge.

Another option is to add `jj split` to the `advance-branches` feature and move to a place where branches either never move implicitly or often move implicitly, depending on whether you're using that feature or not.

I think I would personally prefer to make `jj tug/slide` a first-party command and delete the `advance-branches` feature since it gives the user more flexibility.

PhilipMetzger: In my opinion we have known that bookmarks are not the solution we want for Git branch like behavior, as most of the behavior was retrofitted on them not considering the impact on the Jujutsu side. Longterm there should be Topics which solve it considerably better. 

> It is unfortunate that the better default seems to be backend-dependent. I see how sliding bookmarks is inconvenient if bookmarks represent CL numbers or whatever the gerrit equivalent is.

I don't think the behavior is backend-dependent, as it just was a bad decision to ever allow that. I also raised this point when Evan initially implemented the advancing bookmarks. Also the requirement to make the movement explicit is something which signals that movment shouldn't be a default anyway.

cli/tests/test_split_command.rs
```diff
@@ -675,3 +680,199 @@ fn test_split_interactive_with_paths() {
...
+        ["", "next invocation\n", "write\nsecond-commit"].join("\0"),
+    )
+    .unwrap();
+    test_env.jj_cmd_ok(&main_path, &["split", "file2", "--parallel"]);
+    insta::assert_snapshot!(stdout, @r###"
+    default: vruxwmqv 97490ec0 second-commit
+    second: vruxwmqv 97490ec0 second-commit
+    "###);
+    insta::assert_snapshot!(get_workspace_log_output(&test_env, &main_path), @r###"
+    @  wqnwkozpkust default@ second@ second-commit
```
martinvonz, line 762: Btw, i used `split --parallel` a day or two ago and wished it had left me on the first commit. Maybe we should change that behavior too. That's obviously out of scope for this PR. I just thought I'd mention it here.

cli/tests/test_split_command.rs
```diff
@@ -675,3 +680,199 @@ fn test_split_interactive_with_paths() {
...
+    std::fs::write(main_path.join("file1"), "foo").unwrap();
+    std::fs::write(main_path.join("file2"), "foo").unwrap();
+
+    // Create the second workspace and change its working copy commit to match
+    // the default workspace.
+    test_env.jj_cmd_ok(
+        &main_path,
+        &["workspace", "add", "--name", "second", "../secondary"],
+    );
+    // Change the working copy in the second workspace.
```
martinvonz, line 704: nit: Seems redundant with line 700

cli/tests/test_split_command.rs
```diff
@@ -675,3 +680,199 @@ fn test_split_interactive_with_paths() {
...
+    // Check the working-copy commit in each workspace in the log output. The "@"
+    // node in the graph indicates the current workspace's working-copy commit.
+    insta::assert_snapshot!(get_workspace_log_output(&test_env, &main_path), @r###"
+    @  qpvuntsmwlqt default@ second@ first-commit
+    ◆  zzzzzzzzzzzz
+    "###);
+    insta::assert_snapshot!(get_workspace_log_output(&test_env, &secondary_path), @r###"
+    @  qpvuntsmwlqt default@ second@ first-commit
+    ◆  zzzzzzzzzzzz
+    "###);
```
martinvonz, line 723: nit: This seems unnecessary. It should be tested in `test_workspaces.rs` (I didn't check, but I assume it is already tested there).

cli/tests/test_split_command.rs
```diff
@@ -675,3 +680,199 @@ fn test_split_interactive_with_paths() {
...
+    insta::assert_snapshot!(get_workspace_log_output(&test_env, &main_path), @r###"
+    @  vruxwmqvtpmx default@ second@ second-commit
+    ○  qpvuntsmwlqt first-commit
+    ◆  zzzzzzzzzzzz
+    "###);
+    insta::assert_snapshot!(get_workspace_log_output(&test_env, &secondary_path), @r###"
+    @  vruxwmqvtpmx default@ second@ second-commit
+    ○  qpvuntsmwlqt first-commit
+    ◆  zzzzzzzzzzzz
+    "###);
```
martinvonz, line 747: nit: this seems redundant with the check just above unless I'm missing something

cli/tests/test_split_command.rs
```diff
@@ -675,3 +680,199 @@ fn test_split_interactive_with_paths() {
...
+    Hint: Run `jj workspace update-stale` to update it.
+    See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy for more information.
+    "###);
+    test_env.jj_cmd_ok(&secondary_path, &["workspace", "update-stale"]);
+    insta::assert_snapshot!(get_workspace_log_output(&test_env, &secondary_path), @r###"
+    @  wqnwkozpkust default@ second@ second-commit
+    │ ○  qpvuntsmwlqt first-commit
+    ├─╯
+    ◆  zzzzzzzzzzzz
+    "###);
```
martinvonz, line 780: nit: this should already be covered by `test_workspaces.rs`

cli/tests/test_split_command.rs
```diff
@@ -675,3 +680,199 @@ fn test_split_interactive_with_paths() {
...
+    │ ○  pmmvwywvzvvn second@
+    ├─╯
+    ◆  zzzzzzzzzzzz
+    "###);
+    insta::assert_snapshot!(get_workspace_log_output(&test_env, &secondary_path), @r###"
+    @  pmmvwywvzvvn second@
+    │ ○  qpvuntsmwlqt default@ first-commit
+    ├─╯
+    ◆  zzzzzzzzzzzz
+    "###);
```
martinvonz, line 819: nit: same comments as above apply here and below

cli/tests/test_split_command.rs
```diff
@@ -226,8 +226,7 @@ fn test_split_with_default_description() {
     std::fs::write(workspace_path.join("file1"), "foo\n").unwrap();
     std::fs::write(workspace_path.join("file2"), "bar\n").unwrap();
 
-    // Create a bookmark pointing to the commit. It will be moved to the second
-    // commit after the split.
+    // Create a bookmark pointing to the commit. It will not move during the split.
```
martinvonz, line 229: nit: technically, it does move from the old commit to the first split commit. Feel free to either clarify that or leave as is

cli/tests/test_split_command.rs
emesterhazy: Yeah, I agree and had the same thought while I was working on this. I can update it in a follow up PR.

cli/tests/test_split_command.rs
emesterhazy: Agreed. I think we can delete at least one of these log calls. I think it is nice to show the log output at least once though since it makes it easier to reason about how the graph looks after the split.

cli/tests/test_split_command.rs
emesterhazy: Yeah, it's redundant. I think it would be better to remove the `jj workspace list` call and just leave a single log call since the log call shows the commit graph. Having the graph makes the test easier to follow, at least IMO.

cli/tests/test_split_command.rs
martinvonz: Yes, I also like to have one for testing the setup and making it easier to follow. If you keep the log output here, I think you can drop the `jj workspace list` output above instead.

cli/tests/test_split_command.rs
emesterhazy: There aren't any tests that split a commit in `test_workspaces.rs`, but there are tests for the working copy becoming stale after a commit is rewritten. Maybe that's good enough. I'll delete this.

cli/tests/test_split_command.rs
emesterhazy: You're right about the technicality. It gets a bit wordy to write it out, but probably better to be precise. Fixed.

cli/tests/test_split_command.rs
emesterhazy: Ack

cli/tests/test_split_command.rs
emesterhazy: Yep, thanks. I restructured the test a bit and forgot to delete this.

cli/tests/test_split_command.rs
emesterhazy: Yep, I slimmed down the tests based on your feedback. Let me know if I misunderstood anything.

cli/src/commands/split.rs
emesterhazy: Resolving since I think we've addressed everything being discussed here w/r/t workspaces.

CHANGELOG.md
martinvonz: I think that looks good but some of the changes ended up in the second commit

emesterhazy: Oof, my bad. Funny enough, in splitting the changes to fix this I fell victim to my `push-sksyopkrnqrs` bookmark moving off of the revision with the `sksyopkrnqrs` change id.

This also makes me wonder, has there been any consideration to allowing `jj git push -b` to match on the shortest unambiguous bookmark name? I had to type out the full `jj git push -b push-sksyopkrnqrs` to upload these changes, but `jj git push -b push-sk` would have been unambiguous.

cli/tests/test_split_command.rs
```diff
@@ -263,8 +264,9 @@ fn test_split_with_default_description() {
     "###);
     assert!(!test_env.env_root().join("editor2").exists());
     insta::assert_snapshot!(get_log_output(&test_env, &workspace_path), @r###"
-    @  kkmpptxzrspx false test_bookmark
+    @  kkmpptxzrspx false
     ○  qpvuntsmwlqt false TESTED=TODO
+    │   test_bookmark
```
martinvonz, line 269: Why does the bookmark end up rendered on the second line? Oh, because it's after the description, and the description has a newline. It looks a bit strange. Do you think we should put bookmarks before the description in the `get_log_output ()` template?

cli/tests/test_split_command.rs
martinvonz: This line is still here. Should be squashed from the child commit.

martinvonz (commented):
LGTM, but I'll leave it for a non-googler to approve

CHANGELOG.md
martinvonz: [Dynamic shell completions](https://jj-vcs.github.io/jj/latest/install-and-setup/#command-line-completion) should help with the typing. We've also talked loosely about having some "branchset" language for selecting branches.

cli/tests/test_split_command.rs
emesterhazy: Yeah, it's not a bad idea to change the template. I'll push another commit that does that.

RE: approval, it's not clear to me if @ilyagr has a strong objection to this change. Let me know if we need more discussion before submitting.

CHANGELOG.md
martinvonz: > I would much rather have the change id and the bookmark both stay with the second revision.

That's an interesting idea that I don't think we've talked much about. It would mean that when splitting the working copy, the newly created parent commit will get a new change id. That seemed strange at first but maybe it's not. I'm think I usually don't care which commit gets to keep the change id in that case.

When splitting an older (non-working-copy) commit, I'm also not sure which commit I would typically want to keep the change id and bookmark.

emesterhazy: I usually think of (non-parallel) split in terms of moving changes out of the target commit and into a child commit. Using a stacked PR workflow, that separate commit might end up in a different pull request, so I don't want the bookmark to move off of the target commit. If I do want the bookmark to move, it's easier to move a bookmark forwards than backwards, since `jj` prevents backwards moves by default.

It also seems odd to me to move the change id to the second commit. If I split change id `b` below, I'd find it a bit surprising if the history ends up as `a->c->b`.
```
              b
              |
b   -split->  c
|             |
a             a
```

emesterhazy: I feel like I should add that I think we should also work on better support for the classic git sliding branches style of workflow, which is I think what Ilya is asking for when he says:

> I would very much like to have ways to make a new commit and have branches slide down to it, and currently jj split is that way

I also agree with Philip's comment:

> In my opinion we have known that bookmarks are not the solution we want for Git branch like behavior, as most of the behavior was retrofitted on them not considering the impact on the Jujutsu side.

The current situation isn't ideal for either type of workflow. Things like split moving bookmarks are awkward for stacked PR style workloads, and are inconsistent with other commands that don't move bookmarks. And for users of the sliding workflow we don't have consistent support across the various commands that create new commits.

Maybe the solution for better git workflow support is topics, or perhaps it's improving and stabilizing the `advance-branches` feature. Discussion about topics (https://github.com/jj-vcs/jj/issues/3402) has been quiet since the middle of last year, but I'd be happy to work with Philip or anyone else on a solution that's responsive to the needs of users using the git style "sliding branches" workflow.

PhilipMetzger: Just to mention, I have used the last 6 months to rewrite #4496. But it still lacks polish in my opinion.

emesterhazy: Ah very cool, I'll take a look.

ilyagr: One possible design I'm thinking of is something like this: have `jj split --child` behave roughly like `jj split` works today, splitting out the changes you list into a new child commit. The parent commit would keep the change id, and it would make sense for it to then keep the branches too. Then, we could have `jj split --parent` to work in the opposite way, creating a new parent commit. This would seem to suggest that `jj split --parent file` would keep `file` in the parent commit.

I am not 100% this is convenient. We'd need to decide between that and having `jj split` work as it does now, but with an option of where the change id and any branches go (e.g. `--child-inherits-id` vs `--parent-inherits-id`; note that shortening this to `--parent` could make it confusing to distingush its behavior from `--parent`'s from the previous paragraph).

Or, perhaps, we can come up with some third option.

I think that, outside Google, git-branch-style development is very popular. Removing the sliding down of branches without changing anything else would inconvenience most outside-google people excited about `jj`, without giving them much in return. So, I'd prefer figuring out the "other" split as a prerequisite to merging something like this. We can then add an alias, e.g. `jj sparent`.

This seems to me to be a practical solution in that it's relatively simple and it's not awfully inconsistent with everything else we do. I was initially thinking that it could also wait, but I'm now realizing that it is likely relatively urgent to fix this inside Google, where the current `split` behavior must be quite inconvenient.

On second thought, I care less about what the default is in the end, especially if we are careful about providing or documenting a built-in alias for the "other" behavior. My feeling is that if there is an option to keep sliding the branches down with an option, the change will not be as anger-inducing to people. (Again, my main objection to the current version of this change is based on my guesses about the psychology of jj users)

------------------------

Evan's points in https://github.com/jj-vcs/jj/pull/5618#discussion_r1948125575 are quite good; we could try to perfect some `jj tug`-like apporach instead of adding functionality to `split`. I think that is harder, both to design and to document so that this information reaches the right people at the right time.

As I think most of us agree, the real problem is that bookmarks are doing too many different jobs. Unfortunately, I'm worried that figuring out the right concepts to make sense of bookmarks-as-PRs, bookmarks-as-tags, and Git tags might not be easy. In particular, I haven't looked at #4496 very carefully, but I'm worried there might be a conflict between "topics" being good at being topics, and them being good at representing GitHub PRs (which is, again, is currently a very important usecase; though hopefully its importance will decrease eventually).

For the previous two paragraphs, I might also be too pessimistic; while I think these are harder, this could change quickly if somebody has a good idea. But currently, thinking about how to cover all the important use-case with those ideas gives me a bit of a headache.

emesterhazy: I should probably say that although I work at Google and use jj there now, I don't work on jj as part of my job and actually started contributing before we could use it with the monorepo. jj's interop with git is one of the things that got me excited about it in the first place, and I think it's important to support the (presumably) large number of users who are familiar with git and interested in using jj with the standard GitHub PR workflow. That's actually why I worked on adding the `advance-branches` config option, even though I don't use it myself.

---

> One possible design I'm thinking of is something like this: have jj split --child behave roughly like jj split works today, splitting out the changes you list into a new child commit.

I'm pretty sure this is the opposite of how split works today. When you run `jj split`, the changes you select stay in the first commit, and everything else moves to the second commit. That aside, I think it's reasonable to have a `--child` flag that moves the changes you select into the second commit instead of leaving them in the first. Sometimes you want to split a small change out of a large commit and are stuck selecting everything except that change in order to keep the bulk of the commit in the parent.

However, this seems orthogonal to whether or not split should add a new parent instead of a new child. If we set aside moving the bookmark, is there a compelling reason to prefer inserting a commit with a new change id as the parent of the target commit instead of adding a child of the target?

---

I also want to push back ever so slightly on the idea that this change to split will significantly inconvenience git-branch-workflow users. IMO they are already very inconvenienced by the fact that `jj new` and `jj commit` don't move branches forwards. I think we should focus effort on providing comprehensive support for that workflow instead of holding onto `jj split` as a workaround.

Here are a few ideas (some reiterated), starting with the simpler ones:

1. We could add a `jj slide/tug` function that allows users to slide the closest bookmark forwards towards a target commit. I like this option a lot since it is simple and consistent. Bookmarks only move off of a change id when the user moves them. This is also a command that people using a stacked PR workflow may find useful from time to time (i.e. maybe you're just updating your dotfiles and need to slide the "main" bookmark up).
1. **We could add a `--slide` flag to `jj split`** that explicitly moves bookmarks from the target commit to the second commit created by the split, preserving the current behavior. I don't think this should be the default, but I'm not opposed to adding it if this is a sticking point.

Longer term:

1. Add topics support and delete the `advance-branches` feature. I need to read more about topics to form an opinion about whether they solve the issue for git-branch-workflow users, but I think it's been suggested that they will.
1. Improve and release `advance-branches` (with a new name). This feature works for `jj new` and `jj commit`, but it's in an "experimental" state where we never publicized it so most users probably don't know that it exists. People who have tried it have also made suggestions for how to improve it, so we could incorporate that feedback and add support for `jj split` as well. Maybe this manifests as `--slide` and `--no-slide` flags added to the key functions (i.e. `new`, etc), along with a config option to make `--slide` the default.
1. Something else entirely?

cli/tests/test_split_command.rs
emesterhazy: It seems like we need to have a bit more discussion about this functionality change, so I moved the new tests in the base commit into a new PR: https://github.com/jj-vcs/jj/pull/5638

CHANGELOG.md
ilyagr: This isn't really a full response, but let me try to explain better why I think sliding down bookmarks is important for the PR workflow. I don't think I did before, or perhaps I did not understand my gut feeling very well myself and so haven't given the right reason. I regard this quite differently than the `advance-branches` feature.

I think the main reason in my mind applies mostly to multi-commit PRs. If a bookmark represents a PR, then splitting a commit in the middle of the PR keeps both halves inside the PR (no matter which option we choose for split behavior). Currently, splitting the leaf commit of the PR works the same way. Changing split to not slide down branches means that splitting a commit in a PR has a substantially different effect depending on whether it's a leaf commit or a commit in the middle.

So, in my mind, whenever one thinks of bookmarks as representing PRs, it is less mentally taxing to use the version of split that slides bookmarks down (whether it's the default or an alias or an option).

I'm not sure whether it's immediately obvious why this feels so important to me. It feels like a tax on mental models; the difference between having to keep in mind different workflows for single-commit and multi-commit PRs, leaf- and  non-leaf commits, versus having one workflow. (Of course, I still need different workflows for, say, reordering leaf- and non-leaf commits, which would be a nice problem to solve; but with split, the problem is currently solved and the workflow is already nice; changing the behavior feels like un-solving the problem)

OTOH, I'm much less enthused about `advance-branches`, since `jj new` on a commit in the middle of a PR creates a commit outside that PR. *Without* `advance-branches`, the same is true for the leaf commit of a PR. I also feel that `advance-branches` doesn't do the right thing if a commit has multiple branches at it, while `split` does. So, in my mind, without some additional ideas, `advance-branches` is not a clear win for the PR workflow in the way that sliding down branches on `split` is (or feel like it is to my gut at the moment).

The fact that `jj split` works as poor man's `advance-branches` is nice, but I think might actually be secondary.

martinvonz: I agree that moving bookmarks to the second commit is useful when working with PRs. However, I think we should try to first design the feature without considering GitHub PRs (or GitLab MRs, etc.), and only then try to see to map it to Git. It may be that the mapping simply doesn't work, and then it may be worth reconsidering our model.

@ilyagr, if we ignore PRs/MRs, do you think the bookmark should end up on the first or second commit? IMO, it's at least clear that it should end up on the commit that inherited the change id. It's less clear if that should be the first or the second commit.

Disclosure: I also work for Google and having bookmarks stay associated with the same change id is helpful for our users. I don't think it's terribly important, though; users can always swap the bookmarks and export to Piper again if `jj split` did the opposite of what they wanted.

> I see how sliding bookmarks is inconvenient if bookmarks represent CL numbers or whatever the gerrit equivalent is.

I think they're called "changes" there too. The Gerrit change id is stored in the commit message in a footer called `Change-Id`.

joyously: > do you think the bookmark should end up on the first or second commit?

If the split commits were siblings, would it be harder to decide where the bookmark goes? And in the case where someone is intending to split into 3 or more, is the question still answered the same way?
What if there are 2 bookmarks there before the split?
>  I think we should try to first design the feature without considering GitHub PRs (or GitLab MRs, etc.)

I agree with this also, but have never seen the usage of a bookmark described as anything other than a way to refer to remote Git branches. They seem to be like Git tags, but not. Topics seems to be the illusive solution to everything, but not actually described either. 

If you don't think about the storage, what do users need? Labels, obviously, but in what context? What is missing here:
- labels in the log (graph) output 
- labels to represent a release (deterministic state of repo) to clone/test/merge.
- labels to categorize changes
- labels to use in commands to navigate repo or feed to revsets or align with colleagues

We can't have this tool without the storage, however, so that does need to be taken into account, but the interop should happen in that specific command (`jj git`) in order to get it right for that backend.
The name "bookmark" evokes a placeholder feel, used for navigation and orientation in the graph output. I think `split` shouldn't move the bookmark, so the question is whether the split commit is a child, parent, or sibling.

PhilipMetzger: > Evan's points in [#5618 (comment)](https://github.com/jj-vcs/jj/pull/5618#discussion_r1948125575) are quite good; we could try to perfect some `jj tug`-like apporach instead of adding functionality to `split`. I think that is harder, both to design and to document so that this information reaches the right people at the right time.
> 
> As I think most of us agree, the real problem is that bookmarks are doing too many different jobs. Unfortunately, I'm worried that figuring out the right concepts to make sense of bookmarks-as-PRs, bookmarks-as-tags, and Git tags might not be easy. In particular, I haven't looked at #4496 very carefully, but I'm worried there might be a conflict between "topics" being good at being topics, and them being good at representing GitHub PRs (which is, again, is currently a very important usecase; though hopefully its importance will decrease eventually).
> 
> For the previous two paragraphs, I might also be too pessimistic; while I think these are harder, this could change quickly if somebody has a good idea. But currently, thinking about how to cover all the important use-case with those ideas gives me a bit of a headache.

I'm of the opinion that Topics will require no awkward dance, as they naturally map to Git branches, this leaves bookmarks only as tags on the Jujutsu side which definitely fit better into the current concept. 

> I agree that moving bookmarks to the second commit is useful when working with PRs. However, I think we should try to first design the feature without considering GitHub PRs (or GitLab MRs, etc.), and only then try to see to map it to Git.

I heavily agree. 

> f you don't think about the storage, what do users need? Labels, obviously, but in what context? What is missing here:
> 
> * labels in the log (graph) output
> * labels to represent a release (deterministic state of repo) to clone/test/merge.
> * labels to categorize changes
> * labels to use in commands to navigate repo or feed to revsets or align with colleagues

Users need somekind of lightweight tag (named metadata) which moves along their Changes, which then can be reflected as a Git branch (this is basically what the Topics design doc tries to propose). This then will also be reflected in `jj split` bringing back the advancing bookmarks somewhat, which fit into the Git-PR workflow. 

Just saying it as a bystander which hasn't used a Gerrit or Critique like system where the new bookmark behavior will excel.

emilazy: Probably should have commented here instead of https://github.com/jj-vcs/jj/issues/3419#issuecomment-2649371680, so quoting:

> The proposed behaviour can be better even for a PR workflow: it lets you split out changes that shouldn’t be in that PR (because they should be postponed to a later PR, or kept in your local stack of changes, or whatever). That is something I actually do sometimes! So I feel like the PR workflow is a bit of a red herring here; I’m not convinced there’s an obvious choice based solely on it.
> 
> Given that, I think that it makes sense to keep bookmarks on the change that keeps the original change ID, based on the general principle of what keeping the original change ID _means_ and the fact that it makes more sense with the branch names `jj git push -c` creates. I think having an option to let the “other side” keep the change ID could make sense, and in that case that commit should get the bookmarks instead. But the current behaviour seems generally wrong to me and I support changing it.

emilazy: (BTW, I’m not totally convinced that topics map *naturally* to Git branches, since even putting aside the ability to have multiple heads, Git branches have no real way of representing discontiguous sets of commits? That doesn’t mean they might not work okay in practice for interop, but I think it’s not completely certain that they are going to be a simple solution to all our branch‐interoperability‐related woes.)

ilyagr: > The pragmatic thing to do at this point if we can't abandon the current behavior is to add a flag that controls which commit inherits the change id and add a config option to control the default behavior. We can remove the flag and config option once Topics or some other solution provides a comprehensive answer to git branch interop.

I might to write a longer reply to the many good thoughts in here later, but in short, I think a flag would be good enough. I'm not sure we need a config option, perhaps a suggested (or built-in) alias `aliases.spslide=["split", "--slide"]` is enough (especially since, I think, dynamic completion now works with aliases 🎉 ). "slip-and-slide :)" 

There are a few remaining issues of polish in my mind, but I think we could postpone resolving these questions to later, and just leave it at whatever point Evan has the energy and desire for. Some examples of these remaining issues: is `--slide` a good name for the long-term? Should we slide down the change id as well (probably, but should it be part of the first version)? Should `jj split --slide file` move `file` to the child commit or should it keep it in the parent; do we like the current behavior of `jj split file` in this respect?

emesterhazy: > Should jj split --slide file move file to the child commit or should it keep it in the parent; do we like the current behavior of jj split file in this respect?

I think it's bad for the bookmark to move off of the change id, so we should probably move the change id to the child commit along with the bookmark.

I'm not sure if `--slide` is the best name for the flag or not. It's kinda pithy for a normal split, but doesn't really make sense for `--parallel` split. This is hopefully not a permanent flag, so maybe we don't need to stress over the name. If anyone has suggestions let me know.

Likewise, if someone is strongly opposed to adding this flag, please comment. If not, I'll rework the PR to add the flag (but no config option).

martinvonz: There's been so much discussion that I'm a bit lost. Can you provide a summary of the changes you plan to make in this PR compared to before this PR?

emesterhazy: 1. The changes originally proposed in this PR will be submitted.

2. In addition, we will add a flag to `jj split` that moves the change id and bookmarks to the second commit created by the split instead of the first. The difference between this and the current behavior in jj 0.26 is that the change id will move to the second commit, not just the bookmarks. This is to support users who want bookmarks to behave more like branches do in Git. Hopefully once there's a better git branch interop story we can remove the flag entirely.

I think [ilyagr](https://github.com/ilyagr) is the only person advocating for the flag in this discussion, but we know anecdotally that there are other Git users which find it helpful for the bookmark to move to the second commit (https://github.com/jj-vcs/jj/issues/3419#issuecomment-2649278042).

emilazy: I personally feel like we should split the flag discussion into a separate issue/PR and implement this change as‐is. The current bookmark behaviour just feels wrong to me; I know bookmarks point to commits and not changes, but if you have two possible candidates to move a bookmark to and one of them matches the change it was previously on, picking the other one seems incorrect. Wanting to be able to make `jj split` move the change ID to the other side for workflow reasons is a sensible desire, but feels orthogonal.

martinvonz: Even with the flag, we'll break the workflow for some users (like in the comment Evan linked to), right? I feel like we should have a config for the new behavior so we can roll it out more slowly and so users can roll it back quickly even once it's enabled by default. Hopefully that's not too much extra work for you, @emesterhazy. What do you think?

emesterhazy: I don't mind adding a config option. Just to clarify:

1. Should we leave the default as-is and hide the new behavior behind the config option?
    * If we change the default, should we print a warning on `jj split` unless the user has set the config option?
    * If we don't change the default, should we print a warning that the default will change in the future?
2. Do we still need a flag if we have a config option? I vote no, for simplicity.
3. What is the deprecation criteria for the config option (I'm going to add a TODO in the code). Is it Topics or something like Topics that has a better usability story with Git branches?

Thanks!

martinvonz: IMO:

1. Yes, for one release cycle
   - n/a
   - yes
2. No
3. Flip the default after one month if early adopters are happy. Delete the config after six months or so

emesterhazy: I'm going to send separate PRs to make these changes in stages. I'll attach them all to https://github.com/jj-vcs/jj/issues/3419 so that there's a single list of them somewhere.

emesterhazy: Resolving this thread.
- I've pushed a new revision that includes a config option controlling the change.
- The default behavior will not change for now.
- New users will see a warning when they run `jj split` until they set the config option to opt in or out of the change.

cli/src/commands/split.rs
```diff
@@ -230,3 +242,26 @@ The remainder will be in the second commit.
...
+
+/// Reads 'split.legacy-bookmark-behavior' from settings and prints a warning if
+/// the value is unset to alert the user to the behavior change.
+fn read_legacy_bookmark_behavior_setting(
+    settings: &jj_lib::settings::UserSettings,
+    ui: &Ui,
+) -> Result<bool, CommandError> {
+    match settings.get_bool("split.legacy-bookmark-behavior") {
+        Err(ConfigGetError::NotFound { name: _ }) => {
+            writeln!(
```
emesterhazy, line 254: I didn't think super hard about the message we should show. Let me know if you have other ideas.

CHANGELOG.md
ilyagr: wow, things are changing quickly. I'm having trouble keeping up with everything.

The idea about the config option seems fine in the short-term (so let's merge it!), but I'm worried that it leaves nobody happy in the medium term.

I think it would be quite helpful to allow both behaviors with a flag. I would occasionally use both; I don't see any theoretical reason to not have such an option. 

We can continue the discussion... somewhere. I'm not sure what the most relevant points are in different people's minds at the moment.

------------

I'll add the following as a reply to @martinvonz 's <https://github.com/jj-vcs/jj/pull/5618#discussion_r1948490921>, since I already had a draft of it written down. I'm not sure how relevant it is at this point.

> @ilyagr, if we ignore PRs/MRs, do you think the bookmark should end up on the first or second commit? 

If we ignore PRs/MRs, I'm not sure where the bookmark should end up, or whether it should become conflicted[^factoid]. This is part of the reason that I think it's good to offer both options.

[^factoid]: OK, one possibly relevant factoid is that, if both sides of the conflict are descendants of the conflict base, `jj git fetch` currently resolves conflicted bookmarks in favor of the descendant of the other side of the conflict. (This behavior also only makes sense in terms of the Git model of relying on named branches to remember/push/pull commits, I don't know whether we'd do it otherwise). This would give some credence to saying that "well, after a split the bookmark should become conflcited, but then we immediately resolve the conflict in favor of the child"

I agree with the philosophy of "first designing the feature without considering GitHub PRs" to a point, as a tool for coming up with good design for everyone. I am not sure it serves us very well here. To a large degree, the purpose of bookmarks is still to interact with GitHub and Git remotes, so it is difficult to unlink them from the model GitHub assumes until we find a better way to interact with GitHub's model.

ilyagr: Just to make it very clear: I'm fine with considering this conversation resolved for PR merging purposes.

ilyagr: Also, thank you Evan for your patience with this!

cli/src/commands/split.rs
```diff
@@ -230,3 +242,26 @@ The remainder will be in the second commit.
...
     Ok(())
 }
+
+/// Reads 'split.legacy-bookmark-behavior' from settings and prints a warning if
+/// the value is unset to alert the user to the behavior change.
+fn read_legacy_bookmark_behavior_setting(
+    settings: &jj_lib::settings::UserSettings,
+    ui: &Ui,
+) -> Result<bool, CommandError> {
+    match settings.get_bool("split.legacy-bookmark-behavior") {
```
yuja, line 252: nit: you can use `.optional()?` to get `Option<bool>`.

fwiw, it might also be good to set the default value in `config/misc.toml`. Suppose this is a temporary flag, we wouldn't need an option to suppress the warning without changing the behavior.

cli/tests/test_split_command.rs
```diff
@@ -797,21 +832,42 @@ fn test_split_with_multiple_workspaces_different_working_copy() {
...
     let mut test_env = TestEnvironment::default();
     test_env.jj_cmd_ok(test_env.env_root(), &["git", "init", "main"]);
     let main_path = test_env.env_root().join("main");
 
+    match bookmark_behavior {
+        BookmarkBehavior::Modern => {
+            test_env.add_config("[split]\nlegacy-bookmark-behavior=false");
+        }
+        BookmarkBehavior::Legacy => {
+            test_env.add_config("[split]\nlegacy-bookmark-behavior=true");
```
yuja, line 855: nit: just use dotted path `split.legacy-bookmark-behavoir`?

emesterhazy: Done

cli/src/commands/split.rs
emesterhazy: Thanks for the `.optional()` suggestion.

> Suppose this is a temporary flag, we wouldn't need an option to suppress the warning without changing the behavior.

I'm not sure I understand. Adding the default to `config/misc.toml` suppresses the warning. I think we want users to  see the warning and update their configs according to their preference. Are you suggesting that we print the warning until the user disables the legacy behavior? Based on the previous discussion on this PR I have a feeling that will be controversial. Maybe we can do that in a few months before actually removing the config option?

CHANGELOG.md
emesterhazy: Let's continue the discussion on the issue tracker in https://github.com/jj-vcs/jj/issues/3419. A few quick comments:

1. I'm not opposed to adding a flag personally, as long as we move the change id to the second commit as well. It does complicate the code a bit though.
2. Leaving bookmarks conflicted after a split isn't a pragmatic solution in my opinion, even if it's theoretically reasonable. I'm not sure I understand the example in the footnote though. Is there a test demonstrating it?

cli/src/commands/split.rs
yuja: > Are you suggesting that we print the warning until the user disables the legacy behavior?

Yes. If the warning could be suppressed by explicitly enabling the legacy behavior, the user would forget about that.

> Maybe we can do that in a few months before actually removing the config option?

That sounds also good. I would take this option if there are high chance we'll revert the deprecation.

emilazy: > Are you suggesting that we print the warning until the user disables the legacy behavior? Based on the previous discussion on this PR I have a feeling that will be controversial.

I think this is the correct path. We're just including the option to follow our convention on depending for breaking changes. If it's a significant impediment to someone's workflow, hopefully the option and warning will get them to speak up, and we can prioritize a longer-term solution like a flag to switch which commit inherits the change ID.

emesterhazy: That's fine with me. @ilyagr @martinvonz

cli/src/commands/split.rs
```diff
@@ -230,3 +242,28 @@ The remainder will be in the second commit.
...
+    {
+        Some(value) => Ok(value),
+        None => {
+            writeln!(
+                ui.warning_default(),
+                "`jj split` will stop moving bookmarks in the next release.\n\
+                     Run `jj config set --repo split.legacy-bookmark-behavior true` \
+                     to silence this message and keep the current behavior.\n\
+                     See https://github.com/jj-vcs/jj/issues/3419"
+            )?;
```
PhilipMetzger, line 264: nit: use a raw string here `r#...#;` as it makes the intended indentation clearer and less error-prone.

PhilipMetzger (approved):
still looks great.

cli/src/commands/split.rs
ilyagr: SGTM, I agree with Emily and Yuya. I'm not sure we gain much by waiting a few months.

I'm also OK with moving slower; IMO the main reason for slowing down would be if people want to keep working on this invisibly to the users. My sense is that this is not where people are at, though.

emesterhazy: Ok, then it's decided. We will print the warning unless the user enables the new behavior. After the new behavior becomes the default, we will print the warning if the user is opting out of the new behavior.

martinvonz (commented):
I had these two draft comments that I meant to send. (But don't let that change your decision about the warning.)

cli/src/commands/split.rs
emesterhazy: Thanks good point. I decided to use three `writeln!` calls instead so that each line is prefixed by `Warning:`. I think this makes it a little easier to distinguish the warning message from the rest of the info printed by `jj split`. I briefly searched through the codebase and didn't see much precedence for multi-line warning messages, so this approach seems reasonable to me.

cli/src/commands/split.rs
```diff
@@ -230,3 +242,26 @@ The remainder will be in the second commit.
...
+/// the value is unset to alert the user to the behavior change.
+fn read_legacy_bookmark_behavior_setting(
+    settings: &jj_lib::settings::UserSettings,
+    ui: &Ui,
+) -> Result<bool, CommandError> {
+    match settings.get_bool("split.legacy-bookmark-behavior") {
+        Err(ConfigGetError::NotFound { name: _ }) => {
+            writeln!(
+                ui.warning_default(),
+                "`jj split` will stop moving bookmarks in the next release.\n\
```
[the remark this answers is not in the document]
emesterhazy: I changed it to the copy below to make it clear what the new behavior is.

> `jj split` will leave bookmarks on the first commit in the next release.

cli/src/commands/split.rs
emesterhazy: We've tweaked this a bit, and I think it's good now.

cli/src/commands/split.rs
emilazy: Wait I’m a bit confused – I thought surely we’d use the new behaviour by default? `jj split` warning out of the box seems bad. Only users who rely on the old behaviour should have to adjust their config, right? Maybe there was a miscommunication here.

emesterhazy:
I am going to submit this since at this point we are just going back and forth on how and when the warning should be printed, and there's still time to adjust that.

For what it's worth, I agree with @emilazy that it's not great to force users that want the new behavior to adjust their config to silence the warning. However, I thought the reason we added the warning in the first place was to avoid confusing users by changing the behavior silently (not everyone reads the patch notes).

I think the current plan is:

1. In the next release `n`, users will get a warning every time they `jj split` unless they opt into the new behavior.
2. In release `n+1` the behavior will be enabled by default, and users will only get a warning if they opt out of the new behavior.
3. In some future release, the opt-out will be removed and users will get an error during `jj split`, or it will just silently do nothing.

Please open a new PR if anyone thinks we should change the plan.

PR #5619, opened 2025-02-07
cli: describe how to set a config value with apostrophes

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/config/set.rs
+++ b/cli/src/commands/config/set.rs
@@ -38,8 +38,15 @@ pub struct ConfigSetArgs {
     /// New value to set
     ///
     /// The value should be specified as a TOML expression. If string value
-    /// doesn't contain any TOML constructs (such as array notation), quotes can
-    /// be omitted.
+    /// doesn't contain any TOML constructs (such as apostrophes or array
+    /// notation), quotes can be omitted. Note that the value may also need
+    /// shell quoting. TOML multi-line strings can be useful if the value
+    /// contains apostrophes. For example, to set `foo.bar` to the string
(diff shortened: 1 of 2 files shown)


cli/src/commands/config/set.rs
```diff
@@ -38,8 +38,13 @@ pub struct ConfigSetArgs {
     /// New value to set
     ///
     /// The value should be specified as a TOML expression. If string value
-    /// doesn't contain any TOML constructs (such as array notation), quotes can
-    /// be omitted.
+    /// doesn't contain any TOML constructs (such as apostrophes or array
+    /// notation), quotes can be omitted. Note that the value may also need
+    /// shell quoting. The exact syntax depends on your shell. For example,
+    /// to set `foo.bar` to the string "don't" in Bash, use 
+    /// `jj config set --user foo.bar '"don'\''t"'`. In Fish, use
```
yuja, line 45: nit: triple quotes can be used in order to avoid closing/reopening shell quotes:
```
"'''don't'''"
```

yuja (approved):

cli/src/commands/config/set.rs
martinvonz: That produces a different result in the TOML source. I don't like to give different examples for Bash and Fish that produce different results even if they represent the same string in the end. We could use the TOML triple quotes for both shells. Do you prefer that?

yuja: Yes, I would rewrite both. The bash example is hard to parse without further explanation.

yuja: Looks good, thanks.

Issue #5312, opened 2025-01-09
FR: Allow splitting into more than just two changes


kastl-ars:
**Is your feature request related to a problem? Please describe.**
Say I have a large change, which needs to be split. With `jj split` currently I would need to rerun the command multiple times to split this up into more than just two changes.

**Describe the solution you'd like**
Adding a new CLI option for `jj split`, that just ask the user after the first selection of changes, if she wants to split the change any further.

Calling `jj split` without this option gets the current behaviour of only splitting once.

Calling `jj split` with the new option starts a split as usual, but after selecting the changes it pops up a question like "Do you want to split up the change any further?". If the user selects no, it is done.

If the user selects yes, split is being called again on the "originating" change (the one the changes are being split from) and the procedure starts again.

**Describe alternatives you've considered**
One alternative would be to just keep the current implementation and have the user rerun the command multiple times to achieve the same.

Another, more intrusive implementation could be to not just select changes (using the space key) but rather select them using numbers (0, 1, 2, 3) to put them into the respective commits. Not sure how intuitive this would be (which commit is 0?) and if it is worth the effort. 

Another alternative would be some kind of TUI that allows making the decisions completely in the TUI. I think this is the most effort and I do not think it is worth it.

martinvonz:
> Calling `jj split` with the new option starts a split as usual, but after selecting the changes it pops up a question like "Do you want to split up the change any further?". If the user selects no, it is done.

Another option is to do what `hg split` does: if the user selects all changes, then it means they don't want to split further.

arxanas:
See also arxanas/scm-record#73 for implementing it in the TUI for the built-in diff editor.

arxanas:
> Maybe I am mixing this up with the squash workflow or I am missing something, but during the "first" split I can select/deselect changes. Selected ones go into a new changes, deselected ones stay (or the other way round). How would I show that I want the changes to go to more than two changes?

In this case, you would make a selection for the first change, and then the "remainder" is what could go into the second *or later* changes. You continue until the "remainder" is empty.
- To indicate that you're "done", you have to actually select all of the changes so that there's no "remainder".
- Therefore, at each step, you make a small commit with a subset of the changes, until you eventually would create an empty commit, at which point you stop instead.

emesterhazy:
This is something I've missed coming from Mercurial, and I'd be happy to see it implemented.

[...]

PR #5620, opened 2025-02-07
split: describe both sides, add `--no-edit`

### LINKED ISSUE #3060: FR: `jj split --message`, like `jj commit --message`
i want to create a commit without context switching between my terminal and my editor. i tried `jj split PATH -m 'make a change'`. that errored because `jj split` doesn't support a `--message` argument.

support `--message` in any command that creates a new commit, not just in `jj commit`

- add `--message` only to `jj split`
- add a way to run jj split without creating a message, so i can run `jj split && jj describe @- --message 'make a change'`

**Additional context**
```
; jj --version
jj 0.14.0-8e4d1af98eccf1ed8a79a7ddf9500467aa7684f5
```

### LINKED ISSUE #5432: FR: Describe both sides of `jj split` in one editor
`jj split` opens an editor up to twice: for the first commit, and optionally for the second commit (if the original commit's description was not empty). Since #3828, it's been possible to describe multiple commits in the same editor.

`jj split` should (configurable?) open an editor with contents like:
```
JJ: describe aaaaaa -----
JJ: describe the first commit
{templates.draft_commit_description}

JJ: describe bbbbbb -----
JJ: describe the second commit
{templates.draft_commit_description}
```

This could be configurable. The current behavior of one commit = one $EDITOR could be useful for workflows that override $EDITOR to do configuration more advanced than available in templates.

**Additional context**
This comes from a discord discussion where the user's normal split workflow involved only setting a message for the first commit. In that case, having any comment is an annoyance, since only the first commit message should be modified. However, that workflow is equivalent to `jj commit`. I am of the opinion that `jj commit [-i]` should be preferred in this case, and `jj split` should always show `JJ: describe` lines for both commits.

bryceberger:
closes #3060, with the solution of "don't open an editor at all"
closes #5432

These could be submitted as separate patches, but they don't merge nicely when separate.

This changes the behavior of split, without offering an option to use the old behavior. `jj split` will always have the UI of `jj describe first second`, unless `--no-edit` or `--only-first` are passed. There is no way to use the old behavior (~ `jj describe first && jj describe second`).

Do we want a config option to control this? I could see something like `ui.split.default-describe = "both|first|second|none|separate"`.

Do we want inverse flags? `--edit`, `--only-second`?

# Checklist

If applicable:


arxanas:
I have two concerns:

- interaction with multi-way splits
- CLI surface area

---

**Interaction with multi-way splits**: I had understood that we probably wanted to support multi-way splits, perhaps by "running `jj split` in a loop":
- you would first split some changes out,
- then set the commit message,
- and if anything "remains", repeat the process.

I think editing all the commit messages simultaneously would make this workflow worse, because you'll have to figure out the commit message for something you split a long time ago. cc @martinvonz, who I think was describing this?

A possible solution might be to surface a synthetic `JJ-MESSAGE` file as part of each split, which would be editable and contain the commit message to apply. Then it should solve both problems:

- you can (?) edit the commit message for both sides simultaneously
- you can keep context in your head about which commit message applies to which changes

[technical note] This should work for external diff editors. For the built-in TUI, it doesn't support editing arbitrary files, but it actually has specific functionality for editing the commit message, which is currently unused in jj.

[off-topic] Lately, for certain workflows, I've really been wishing that the commit message were literally embedded as content in the commit somehow, so that you could e.g. apply 3-way merge to a commit message just like its contents.

---

**CLI surface area**: Overall, I'm inclined to minimize the CLI surface area. It's a lot easier to add a missing flag later than to remove a flag that users have come to depend on. Therefore, if we went through with the PR in this form, my own preference would be:

- No `--edit`/`--no-edit` flag.
  - If we wanted to support this, I would also be inclined to control *both* the diff editor and the commit editor via the same `--interactive`/`--no-interactive` flag, which should basically mean "don't launch the editor".
  - I think users can restore this workflow via `--config ui.editor=true` if they need to.
- No `--only-first` (or `--only-side=1`) flag.
  - It seems like users would be able to adjust their workflows to always edit both sides (if only by scripting something to remove one of the sides).
  - But the whole affair might not play well with multi-way splitting.

[tangential] Oftentimes I kind of want to support features that also print out warning messages that say "this feature is experimental and will get deleted unless you post about your workflow at this link ..." to see if people are actually using the feature and how 🤣.

emesterhazy:
The multiway splits issue is https://github.com/jj-vcs/jj/issues/5312. I commented recently that I'd like to see it in `jj`, but I could also live without it. I do think that this PR will complicate multi-way splits though for the reason @arxanas mentioned.

If we decide that multi-way splits won't be supported, editing both messages in the same editor seems reasonable, but not essential. I do find myself deleting the message for the second commit more often than not (but not always), so I probably would use a flag that drops the message in the second commit.

The concern about CLI surface area is legitimate, but it's important to find a good balance. For people who use the CLI all day long, small efficiency gains can add up.

PR #5621, opened 2025-02-07
cli: untrack remote bookmarks on `bookmark forget`

### LINKED ISSUE #5524: FR: `jj bookmark untrack` should delete the local bookmark

scott2000:
Fixes #5524. This solution was suggested by @yuja in #5546.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/bookmark/forget.rs
+++ b/cli/src/commands/bookmark/forget.rs
@@ -26,13 +26,22 @@ use crate::command_error::CommandError;
 use crate::complete;
 use crate::ui::Ui;
 
-/// Forget everything about a bookmark, including its local and remote
-/// targets
+/// Forget a bookmark without marking it as a deletion to be pushed
 ///
-/// A forgotten bookmark will not impact remotes on future pushes. It will be
-/// recreated on future pulls if it still exists in the remote.
+/// If a local bookmark is forgotten, any corresponding remote bookmarks will
+/// become untracked to ensure that the forgotten bookmark will not impact
+/// remotes on future pushes.
 #[derive(clap::Args, Clone, Debug)]
 pub struct BookmarkForgetArgs {
+    /// When forgetting a local bookmark, also forget any corresponding remote
+    /// bookmarks
+    ///
+    /// A forgotten remote bookmark will not impact remotes on future pushes. It
+    /// will be recreated on future fetches if it still exists on the remote. If
+    /// there is a corresponding Git-tracking remote bookmark, it will also be
+    /// forgotten.
+    #[arg(long)]
+    include_remotes: bool,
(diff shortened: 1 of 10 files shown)


cli/src/commands/bookmark/forget.rs
```diff
@@ -57,22 +64,46 @@ pub fn cmd_bookmark_forget(
...
-        ),
-    )?;
+    if untracked == 0 {
+        writeln!(ui.status(), "Forgot {forgotten} bookmarks.")?;
+    } else {
+        writeln!(
+            ui.status(),
+            "Forgot {forgotten} bookmarks and untracked {untracked} remote bookmarks."
+        )?;
+    }
```
yuja: nit: simply print these as separate messages?

It might be also better to print "N local bookmarks" and "M remote bookmarks" (instead of forgotten + untracked) as we kind of define "forget local bookmarks" is the operation to delete + untrack.

yuja (commented):
I would split the first patch because I'm not sure which is better: `NAME@REMOTE` or `--remote=REMOTE NAME`.

cli/src/commands/bookmark/forget.rs
```diff
@@ -57,22 +66,39 @@ pub fn cmd_bookmark_forget(
...
+        ui.status(),
+        "Forgot {} local bookmarks.",
+        matched_bookmarks.len()
     )?;
+    if forgotten_remote != 0 {
+        writeln!(ui.status(), "Forgot {forgotten_remote} remote bookmarks.")?;
+    }
+    let forgotten_bookmarks = matched_bookmarks
+        .iter()
+        .map(|(name, _)| name.to_string())
```
yuja, line 99: nit: `.to_string()` can be omitted

yuja (approved):

PR #5622, opened 2025-02-08
cli: remove `ui.allow-filesets` configuration option

bnjmnt4n:
The "fileset" language has been enabled by default since v0.20.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* The `ui.allow-filesets` configuration option has been removed.
+  [The "fileset" language](docs/filesets.md) has been enabled by default since v0.20.
+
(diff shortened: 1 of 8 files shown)


bnjmnt4n:
I'm not sure if we should deprecate the option first, which I imagine means we would want to further delay the removal of the option?

PhilipMetzger:
> I'm not sure if we should deprecate the option first, which I imagine means we would want to further delay the removal of the option?

I'm not sure I understand you correctly, as the config option `ui.allow-filesets` was added to help with the gradual rollout from Yuya. There's another similarly named option which is called `allow-large-filesets` which _should_ get a proper deprecation window, as it actually changes behavior in revset selection (which I have configured for myself).

yuja:
I think okay to delete `ui.allow-filesets` since it existed just in case filesets had a major breakage.

thoughtpolice (approved):

PR #5623, opened 2025-02-08
repo: Change name for unused closure argument

emesterhazy:
The previous name violated the "snake case" convention for variable names. I suspect the intention was to use `RewriteRootCommit` as a type declaration, not the argument name. Since the argument isn't used, we can use an underscore as the name and leave the type declaration.

#cleanup

### DIFF
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1386,7 +1386,7 @@ impl MutableRepo {
         self.maybe_abandon_wc_commit(&workspace_id)?;
         self.add_head(commit)?;
         self.set_wc_commit(workspace_id, commit.id().clone())
-            .map_err(|RewriteRootCommit| EditCommitError::RewriteRootCommit)
+            .map_err(|_: RewriteRootCommit| EditCommitError::RewriteRootCommit)
     }
 
     fn maybe_abandon_wc_commit(


martinvonz (approved):
(Approving since it's clearly not Google-specific)

lib/src/repo.rs
```diff
@@ -1386,7 +1386,7 @@ impl MutableRepo {
         self.maybe_abandon_wc_commit(&workspace_id)?;
         self.add_head(commit)?;
         self.set_wc_commit(workspace_id, commit.id().clone())
-            .map_err(|RewriteRootCommit| EditCommitError::RewriteRootCommit)
+            .map_err(|_: RewriteRootCommit| EditCommitError::RewriteRootCommit)
```
yuja, line 1389: I think the intent here is to ensure that `RewriteRootCommit` is a unit struct. It could be expressed as `|RewriteRootCommit: RewriteRootCommit|`, but that seems too verbose.

martinvonz: I was also wondering if that might have been the intent. Seems unlikely that `RewriteRootCommit` will gain other variants or additional data, though.

I'm also curious to see how the merge queue deals with this unresolved comment :)

yuja: Still it doesn't make much sense to relax the requirement. If `RewriteRootCommit` had a value, it would have to be propagated.

martinvonz:
> The previous name violated the "snake case" convention for variable names.

Actually, where did you see that? I think the compiler actually treats it as a destructing variable binding that's binding no variables (I haven't checked the official terms for these things but hopefully it's clear what I mean). So if some tool told you that it was a variable, then maybe we should file a bug report for that tool. Or was it just that it looked odd to you? In that case, I think we should roll back this PR.

emesterhazy:
I'll check later when I'm back at the computer. I'm using Rust Rover (Jet Brains IDE), which surfaced the lint. I'm not 100% sure about the parsing rules for destructing a unit tuple, but this seems ambiguous since the identifier could be an argument name.

lib/src/repo.rs
martinvonz: How about `RewriteRootCommit: _` to make it a pattern-match and still rely on the compiler to infer the type?

emesterhazy:
Ok, so I tried a few things.

1. First, without this change, Rust Rover shows the snake_case identifier lint, but Clippy and Rust Analyzer in VS Code do not. I think this is indeed being parsed as destructuring the (empty) `RewriteRootCommit` struct.
2. I reverted this commit, and then added a member to `RewriteRootCommit`. After doing that, the `RewriteRootCommit` is parsed as a closure variable, and `cargo build` shows two warnings:

```
warning: unused variable: `RewriteRootCommit`
    --> lib/src/repo.rs:1389:23
     |
1389 |             .map_err(|RewriteRootCommit| EditCommitError::RewriteRootCommit)
     |                       ^^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_RewriteRootCommit`
     |
     = note: `#[warn(unused_variables)]` on by default

warning: variable `RewriteRootCommit` should have a snake case name
    --> lib/src/repo.rs:1389:23
     |
1389 |             .map_err(|RewriteRootCommit| EditCommitError::RewriteRootCommit)
     |                       ^^^^^^^^^^^^^^^^^ help: convert the identifier to snake case: `rewrite_root_commit`
     |
     = note: `#[warn(non_snake_case)]` on by default

warning: `jj-lib` (lib) generated 2 warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 11.46s
```

**tl;dr**: The original lint seems to be a mistake from Rust Rover, but the original code is also brittle since adding a member to `RewriteRootCommit` doesn't cause a compiler error at the `map_err` call.

If we really want to make this foolproof, we should add a custom `impl From<RewriteRootCommit> for EditCommitError`:

```rust
diff --git a/lib/src/repo.rs b/lib/src/repo.rs
index 947da81cb3..909378f8d0 100644
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1385,8 +1385,7 @@
     ) -> Result<(), EditCommitError> {
         self.maybe_abandon_wc_commit(&workspace_id)?;
         self.add_head(commit)?;
-        self.set_wc_commit(workspace_id, commit.id().clone())
-            .map_err(|RewriteRootCommit| EditCommitError::RewriteRootCommit)
+        Ok(self.set_wc_commit(workspace_id, commit.id().clone())?)
     }

     fn maybe_abandon_wc_commit(
@@ -1855,6 +1854,15 @@
     BackendError(#[from] BackendError),
 }

+impl From<RewriteRootCommit> for EditCommitError {
+    fn from(_: RewriteRootCommit) -> Self {
+        // This let binding guarantees that RewriteRootCommit is a unit struct.
+        // If members are added, update this function to propagate them.
+        let _ = RewriteRootCommit;
+        EditCommitError::RewriteRootCommit
+    }
+}
+
```

lib/src/repo.rs
emesterhazy: > How about RewriteRootCommit: _ to make it a pattern-match and still rely on the compiler to infer the type?

This doesn't work because as soon as you add a member to the struct the compiler starts treating `RewriteRootCommit` as a variable name.

emesterhazy:
I mailed https://github.com/jj-vcs/jj/pull/5624 which adds the custom `From` impl I described in my last comment. I think this is the only way to guarantee a compiler error during the conversion if members are added to `struct RewriteRootCommit` later.

If we don't want to add the `From` for some reason, then I agree we should revert this commit.

emilazy:
I think `RewriteRootCommit {}` will work and ensure an error if the type changes.

emesterhazy:
Oh interesting. I assumed that wouldn't work and didn't try it. It would be better if Rust only allowed that syntax for empty structs IMO, since it catches new members being added.

Anyways, I still kinda think that using `From` is better. It's cleaner than needing `map_err`, and hopefully prevents other "bad" `map_err` calls from being introduced in the future. I'll leave it up to you all collectively though, so please comment on https://github.com/jj-vcs/jj/pull/5624 if you disagree.

Sorry for creating this noise!

PR #5624, opened 2025-02-08
repo: Use thiserror #[from] with EditCommitError::RewriteRootCommit

emesterhazy:
See the discussion in #5623. This prevents future accidents if member variables are added to RewriteRootCommit which need to be propagated to EditCommitError during conversion.

### DIFF
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1385,8 +1385,7 @@ impl MutableRepo {
     ) -> Result<(), EditCommitError> {
         self.maybe_abandon_wc_commit(&workspace_id)?;
         self.add_head(commit)?;
-        self.set_wc_commit(workspace_id, commit.id().clone())
-            .map_err(|_: RewriteRootCommit| EditCommitError::RewriteRootCommit)
+        Ok(self.set_wc_commit(workspace_id, commit.id().clone())?)
     }
 
     fn maybe_abandon_wc_commit(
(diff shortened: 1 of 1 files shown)


lib/src/repo.rs
```diff
@@ -1855,6 +1854,15 @@ pub enum EditCommitError {
     BackendError(#[from] BackendError),
 }
 
+impl From<RewriteRootCommit> for EditCommitError {
```
yuja, line 1857: This could be handled by thiserror:
```rust
    #[error(transparent)]
    RewriteRootCommit(#[from] RewriteRootCommit),
```

I think https://github.com/jj-vcs/jj/pull/5623#issuecomment-2645992444 is also a good option. I didn't know the match pattern `Foo` is (roughly?) equivalent to `Foo { .. }`.

emilazy: > I didn't know the match pattern `Foo` is (roughly?) equivalent to `Foo { .. }`.

I believe what actually happens is that `Foo` is parsed as a unit struct pattern if there is such a unit struct, but if there is not, then it’s a variable name. So it changes syntactic form depending on the shape of the type in scope, which is weird. `Foo {}` should avoid that, since it can only parse as a struct pattern. (Conversely, `Foo { .. }` would fail if `Foo` was removed or became something else that can’t match that way, I think.)

emesterhazy: I also didn't realize `Foo {}` and `Foo` were equivalent for empty structs. It would be better if only `Foo {}` worked I think to avoid this entire issue.

I thought about using thiserror, wasn't sure why it wasn't used in the first place. I guess the downside is that the `EditCommitError::RewriteRootCommit` enum will have an empty struct in it.

Overall I think it's better to handle this with `From`, either using thiserror or by hand, since hopefully avoids "bad" `map_err` calls being added in the future. I don't have an opinion on which is better though. Whatever you prefer is fine by me.

yuja: I don't have strong feeling. There's only one caller of `From`, so I would choose anything that's easier to implement (i.e. `#[from]` or `.map_err()`.)

martinvonz (approved):

PR #5626, opened 2025-02-09
sign: Support ~/ path expansion for allowed-signers

### LINKED ISSUE #5625: `signing.backends.ssh.allowed-signers` doesn't support ~ expansion (as `signing.key` does)
## Description

I was trying out the new signature verification and all of my changes were marked `[?]`, even though `git log --show-signature` reported valid signatures. On a whim, I changed my `allowed-signers` to use an absolute path instead of `~/`, which fixed the signature verification.

1. Set up commit signing using `~/` paths and signature verification:
```
[signing]
backend = "ssh"
backends.ssh.allowed-signers = "~/.ssh/signers"
key = "~/.ssh/<yourkey>.pub"

[ui]
show-cryptographic-signatures = true
```
1. Check that `jj log` shows `[?]`
1. Change `signing.backends.ssh.allowed-signers` to use an absolute path
1. Check that `jj log` shows `[✓]`

## Expected Behavior

`signing.backends.ssh.allowed-signers` to accept a `~` path, similar to `signing.key`.

## Actual Behavior

`signing.backends.ssh.allowed-signers` fails to find the signer file

## Specifications

- Platform: macOS
- Version: `jj 0.26.0`

JacobHayes:
I'm very new to rust so not sure this is the best approach - happy to hear any suggestions. I'll add tests and a changelog entry in a bit.

Fixes #5625

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj status` now shows untracked files under untracked directories.
   [#5389](https://github.com/jj-vcs/jj/issues/5389)
 
+* The `signing.backends.ssh.allowed-signers` configuration option will now
+  expand `~/` to `$HOME/`.
+  [#5626](https://github.com/jj-vcs/jj/pull/5626)
+
 ## [0.26.0] - 2025-02-05
 
 ### Release highlights
--- a/lib/src/ssh_signing.rs
+++ b/lib/src/ssh_signing.rs
@@ -123,7 +123,8 @@ impl SshBackend {
         let program = settings.get_string("signing.backends.ssh.program")?;
         let allowed_signers = settings
             .get_string("signing.backends.ssh.allowed-signers")
-            .optional()?;
+            .optional()?
+            .map(|v| crate::file_util::expand_home_path(v.as_str()));
         Ok(Self::new(program.into(), allowed_signers.map(|v| v.into())))
     }
 


lib/src/ssh_signing.rs
```diff
@@ -115,7 +115,8 @@ impl SshBackend {
     pub fn new(program: OsString, allowed_signers: Option<OsString>) -> Self {
         Self {
             program,
-            allowed_signers,
+            allowed_signers: allowed_signers
+                .map(|s| crate::file_util::expand_home_path(s.to_str().unwrap()).into_os_string()),
```
yuja, line 119: It's probably better to do path resolution in `from_settings()` because `"~"` may be a valid path in other context. It will also help remove `.unwrap()`.

yuja (commented):
Thanks.

JacobHayes:
Is it worth adding a test for this? A couple things I noticed:
- I don't see a similar test for the existing `~/` expansion in `signing.key`
- none of the current ssh signing tests create a `UserSetting`, so will need a bit of extra scaffolding since we only added the logic to `.from_settings`
- to avoid `std::env::set_var` (which seems understandably rare in the test suite) for `HOME`, it might be easiest to just check that `allowed_signers` does _not_ start with `~/` after expansion

Happy to dig in if you think it's worth it, but figured I'd check first.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `ellipsis` parameter; passing this prepends or appends the ellipsis to the
   content if it is truncated to fit the maximum width.
 
+* The `signing.backends.ssh.allowed-signers` configuration option will now
+  expand `~/` to `$HOME/`.
```
yuja, line 30: nit: I think this is more like a bug fix than new feature.

yuja (approved):
> Is it worth adding a test for this?

I would say no. It's nice to have some e2e tests for SSH signing, but we'll need to write a fake-ssh-keygen stub or something.

PR #5629, opened 2025-02-09
annotate: make AnnotationLine template type

bryceberger:
Discord discussion: https://discord.com/channels/968932220549103686/1337969561793990757

Allows:
* self.commit()
* self.line_number()
* self.first_line_in_hunk()

Certain pagers (like `delta`), when used for `git blame`, only show the
commit information for the first line in a hunk. This would be a nice
addition to `jj file annotate`.

`jj file annotate` already uses a template to control the rendering of
commit information --- `templates.annotate_commit_summary`. Instead of
a custom CLI flag, the tools necessary to do this should be available in
the template language.

If `1 % 2` or `1.is_even()` was available in the template language, this
would also allow alternating colors (using `raw_escape_sequence`).

Example:

```toml
[templates]
# only show commit info for the first line of each hunk
annotate_commit_summary = '''
if(first_line_in_hunk,
  show_commit_info(commit),
  pad_end(20, " "),
)
'''
```

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/file/annotate.rs
+++ b/cli/src/commands/file/annotate.rs
@@ -16,7 +16,6 @@ use clap_complete::ArgValueCandidates;
 use clap_complete::ArgValueCompleter;
 use jj_lib::annotate::get_annotation_for_file;
 use jj_lib::annotate::FileAnnotation;
-use jj_lib::commit::Commit;
 use jj_lib::repo::Repo;
 use jj_lib::revset::RevsetExpression;
 use tracing::instrument;
@@ -25,6 +24,8 @@ use crate::cli_util::CommandHelper;
 use crate::cli_util::RevisionArg;
 use crate::command_error::user_error;
 use crate::command_error::CommandError;
+use crate::commit_templater::AnnotationLine;
+use crate::commit_templater::CommitTemplateLanguage;
 use crate::complete;
 use crate::templater::TemplateRenderer;
 use crate::ui::Ui;
@@ -33,8 +34,6 @@ use crate::ui::Ui;
 ///
 /// Annotates a revision line by line. Each line includes the source change that
 /// introduced the associated line. A path to the desired file must be provided.
-/// The per-line prefix for each line can be customized via
-/// template with the `templates.annotate_commit_summary` config variable.
 #[derive(clap::Args, Clone, Debug)]
 pub(crate) struct FileAnnotateArgs {
     /// the file to annotate
@@ -51,6 +50,21 @@ pub(crate) struct FileAnnotateArgs {
         add = ArgValueCandidates::new(complete::all_revisions)
     )]
     revision: Option<RevisionArg>,
+    /// Render a prefix for each line using the given template
+    ///
+    /// All 0-argument methods of the [`AnnotationLine` type] are available as
+    /// keywords in the [template expression].
+    ///
+    /// If not specified, this defaults to the
+    /// `templates.file_annotate` setting.
+    ///
+    /// [template expression]:
+    ///     https://jj-vcs.github.io/jj/latest/templates/
+    ///
+    /// [`AnnotationLine` type]:
+    ///     https://jj-vcs.github.io/jj/latest/templates/#annotationline-type
+    #[arg(long, short = 'T', add = ArgValueCandidates::new(complete::template_aliases))]
+    template: Option<String>,
 }
 
 #[instrument(skip_all)]
@@ -75,10 +89,19 @@ pub(crate) fn cmd_file_annotate(
         )));
     }
 
-    let annotate_commit_summary_text = workspace_command
-        .settings()
-        .get_string("templates.annotate_commit_summary")?;
-    let template = workspace_command.parse_commit_template(ui, &annotate_commit_summary_text)?;
+    let template_text = match &args.template {
+        Some(value) => value.clone(),
+        None => workspace_command
+            .settings()
+            .get_string("templates.file_annotate")?,
+    };
+    let language = workspace_command.commit_template_language();
+    let template = workspace_command.parse_template(
+        ui,
+        &language,
+        &template_text,
+        CommitTemplateLanguage::wrap_annotation_line,
+    )?;
 
     // TODO: Should we add an option to limit the domain to e.g. recent commits?
     // Note that this is probably different from "--skip REVS", which won't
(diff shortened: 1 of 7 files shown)


cli/src/commands/file/annotate.rs
```diff
@@ -94,17 +101,27 @@ pub(crate) fn cmd_file_annotate(
...
+            None => true,
+            Some(last) => last != commit_id,
+        };
+        let annotate_data = AnnotateData {
+            commit,
+            line_no: line_no + 1,
+            is_first_line,
+        };
+        template_render.format(&annotate_data, formatter.as_mut())?;
         formatter.write_all(line)?;
```
yuja, line 123: Perhaps, we should also move `line` to template, but we don't have a `BString` template type right now. This can be addressed later.

cli/src/commit_templater.rs
```diff
@@ -2197,3 +2220,41 @@ pub fn builtin_cryptographic_signature_methods<'repo>(
...
+    map.insert(
+        "commit",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            function.expect_no_arguments()?;
+            let out_property = self_property.and_then(|data| Ok(data.commit.clone()));
+            Ok(L::wrap_commit(out_property))
+        },
+    );
+    map.insert(
+        "line_no",
```
yuja, line 2244: nit: rename to `line_number`? We use that in diff output.

cli/src/commit_templater.rs
```diff
@@ -2197,3 +2220,41 @@ pub fn builtin_cryptographic_signature_methods<'repo>(
...
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            function.expect_no_arguments()?;
+            let out_property = self_property.and_then(|data| Ok(data.line_no.try_into()?));
+            Ok(L::wrap_integer(out_property))
+        },
+    );
+    map.insert(
+        "is_first_line",
```
yuja, line 2252: "first line" is confusing, and template methods don't have `is_` prefix. Maybe it can be "first_line_in_hunk" or "first_hunk_line"? It's not technically a "hunk", but I don't have a better suggestion.

cli/src/commit_templater.rs
```diff
@@ -2197,3 +2220,41 @@ pub fn builtin_cryptographic_signature_methods<'repo>(
     );
     map
 }
+
+#[derive(Debug, Clone)]
+pub struct AnnotateData {
```
yuja, line 2225: nit: suppose we'll add line content filed, `AnnotationLine` or `AnnotatedLine` might be a better name.

cli/src/commit_templater.rs
```diff
@@ -2197,3 +2220,41 @@ pub fn builtin_cryptographic_signature_methods<'repo>(
...
+    pub commit: Commit,
+    pub line_no: usize,
+    pub is_first_line: bool,
+}
+
+pub fn builtin_annotate_data_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, AnnotateData>
```
yuja, line 2231: Can you also add docs and templating tests?

martinvonz:
@yuja, it looks like this is ready for another review whenever you have time. All the discussions are marked resolved but there were no text responses to any of them, so you might not have noticed.

cli/src/commands/file/annotate.rs
```diff
@@ -51,6 +52,22 @@ pub(crate) struct FileAnnotateArgs {
         add = ArgValueCandidates::new(complete::all_revisions)
     )]
     revision: Option<RevisionArg>,
+    /// Render each line using the given template
+    ///
+    /// Run `jj log -T` to list the built-in templates.
```
yuja, line 57: I would copy help text from non-`Commit` template (such as `cmd_file_list()`.) The built-in templates are mostly useless in this context.

cli/src/commands/file/annotate.rs
```diff
@@ -51,6 +52,22 @@ pub(crate) struct FileAnnotateArgs {
...
+    /// You can also specify arbitrary [template expressions] using the
+    /// [built-in keywords].
+    ///
+    /// If not specified, this defaults to the
+    /// `templates.annotate_commit_summary` setting.
+    ///
+    /// [template expression]:
+    ///     https://jj-vcs.github.io/jj/latest/templates/
+    /// [built-in keywords]:
+    ///     https://jj-vcs.github.io/jj/latest/templates/#commit-keywords
```
yuja, line 68: nit: insert blank line so these link anchors won't be reformatted into a single line

cli/src/commands/file/annotate.rs
```diff
@@ -75,10 +92,19 @@ pub(crate) fn cmd_file_annotate(
...
 
-    let annotate_commit_summary_text = workspace_command
-        .settings()
-        .get_string("templates.annotate_commit_summary")?;
-    let template = workspace_command.parse_commit_template(ui, &annotate_commit_summary_text)?;
+    let annotate_commit_summary_text = match &args.template {
+        Some(value) => value.clone(),
+        None => workspace_command
+            .settings()
+            .get_string("templates.annotate_commit_summary")?,
```
yuja, line 99: Since it is now template for the annotation type, it's probably better to rename to `templates.<command>` (i.e. `templates.file_annotate`).

It's breaking change, but the original summary template can't be used anyway.

cli/src/commands/file/annotate.rs
```diff
@@ -94,17 +120,27 @@ pub(crate) fn cmd_file_annotate(
...
+    let mut last_id = None;
+    for (line_number, (commit_id, line)) in annotation.lines().enumerate() {
         let commit_id = commit_id.expect("should reached to the empty ancestor");
         let commit = repo.store().get_commit(commit_id)?;
-        template_render.format(&commit, formatter.as_mut())?;
-        write!(formatter, " {:>4}: ", line_no + 1)?;
+        let first_line_in_hunk = match last_id {
+            None => true,
+            Some(last) => last != commit_id,
+        };
```
yuja, line 135: nit: `last_id.as_ref() != Some(&commit_id)`?

cli/src/commit_templater.rs
```diff
@@ -2197,3 +2220,41 @@ pub fn builtin_cryptographic_signature_methods<'repo>(
...
+
+pub fn builtin_annotation_line_methods<'repo>(
+) -> CommitTemplateBuildMethodFnMap<'repo, AnnotationLine> {
+    type L<'repo> = CommitTemplateLanguage<'repo>;
+    let mut map = CommitTemplateBuildMethodFnMap::<AnnotationLine>::new();
+    map.insert(
+        "commit",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            function.expect_no_arguments()?;
+            let out_property = self_property.and_then(|data| Ok(data.commit.clone()));
```
yuja, line 2239: nit: redundant `.clone()`

docs/templates.md
```diff
@@ -401,6 +401,15 @@ This type cannot be printed. The following methods are defined.
   `"git-submodule"`, or `"conflict"`.
 * `.executable() -> Boolean`: True if the entry is an executable file.
 
+### AnnotationLine type
```
yuja, line 404: nit: can you move this section above? Types are sorted lexicographically.

docs/templates.md
```diff
@@ -401,6 +401,15 @@ This type cannot be printed. The following methods are defined.
...
+
+The following methods are defined.
+
+* `.line_number() -> Integer`: 1-based line number.
+* `.commit() -> Commit`: Commit responsible for changing the relevant line.
+* `.first_line_in_hunk() -> Boolean`: False when the directly preceding line
```
yuja, line 410: nit: I would reorder them as `commit, line_number, first_line_in_hunk` because `commit` is probably the most important, and methods are defined in that order.

yuja (approved):
Thanks.

(Please also feel free to ping me if you get no response in a couple of days.)

cli/src/commands/file/annotate.rs
bryceberger: Should there be a `ConfigMigrationRule`? Leaning towards no, since the migration seems like it would be difficult / impossible to write. Could add something like `Rule::rewrite_no_auto_update()`, but that feels like a bit much for this.

yuja: Agreed. It's not easy to rewrite template expression.

On thing we could do is to add custom migration that does nothing except for warning, but I don't have strong feeling about that.

PR #5635, opened 2025-02-10
docs: add diagrams for jj new -A and -B

neongreen:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -1513,7 +1513,60 @@ Note that you can create a merge commit by specifying multiple revisions as argu
 * `-m`, `--message <MESSAGE>` — The change description to use
 * `--no-edit` — Do not edit the newly created change
 * `-A`, `--insert-after <REVSETS>` — Insert the new change after the given commit(s)
+
+   Example: `jj new --after A` creates a new change between `A` and its
+   children:
+
+   ```text
+                   B   C
+                    \ /
+       B   C   =>    @
+        \ /          |
+         A           A
+   ```
+
+   Specifying `--after` multiple times will relocate all children of the
+   given commits.
+
+   Example: `jj new --after A --after X` creates a change with `A` and `X`
+   as parents, and rebases all children on top of the new change:
+
+   ```text
+                   B   Y
+                    \ /
+       B  Y    =>    @
+       |  |         / \
+       A  X        A   X
+   ```
 * `-B`, `--insert-before <REVSETS>` — Insert the new change before the given commit(s)
+
+   Example: `jj new --before C` creates a new change between `C` and its
+   parents:
+
+   ```text
+                      C
+                      |
+         C     =>     @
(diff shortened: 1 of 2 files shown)


neongreen:
image

neongreen:
I realize now that `jj prev` and `jj parallelize` show graphs vertically and not horizontally, but also I think horizontal graphs are nicer and save space in the CLI output. I don't mind converting horizontal->vertical or vice versa, but I also don't mind leaving them in an inconsistent state and seeing which will eventually win.

yuja:
Can you rewrite the graph without using non-ASCII characters? Unicode drawings might be rendered badly. I don't have any preference over horizontal/vertical.

PhilipMetzger:
I agree with Yuya and think that the graph should be kept vertical to fit into the rest of the documentation as consistency is key.

martinvonz:
I agree with Philip that the graphs should be consistent with our existing graphs to make it easier for users to read them.

I don't have much opinion between horizontal vs vertical per se (maybe a slight preference for vertical since it matches `jj log`). I care more about the consistency.

neongreen:
I removed unicode and transmuted the diagrams

cli/tests/cli-reference@.md.snap
```diff
@@ -1511,8 +1511,66 @@ Note that you can create a merge commit by specifying multiple revisions as argu
...
+   Example: `jj new -A 1 -B 2` creates a change between `1` and `2`,
+   but does not touch `3`:
+
+   ```text
+
+                  2
+                  |
+     2   3   =>   @   3
+      \ /          \ /
+       1            1
```
martinvonz, line 1568: `jj new -B 2` would have been sufficient in this case, so I think we should give another example where both the `-A` and the `-B` are needed.

PhilipMetzger:
last concistency nit from me, use the letters instead of numbers in the graph

cli/tests/cli-reference@.md.snap
neongreen: Done, I changed it to a stitching thing

```rust
    /// Example: `jj new --after A --before D`:
    ///
    /// ```text
    /// 
    ///     D            D
    ///     |           / \
    ///     C          |   C
    ///     |    =>    @   |
    ///     B          |   B
    ///     |           \ /
    ///     A            A
    /// ```
```

neongreen: (empty line b/c otherwise it glues to "text" unpleasantly)

neongreen:
> last concistency nit from me, use the letters instead of numbers in the graph

done, and I switched to `--after` and `--before` because `-A A` isn't great

martinvonz (approved):

neongreen:
Hmmm. Anything I need to do to start the CI? Or is it just that the merge queue is long?

emilazy:
You need to rebase this, it’s from before the CI changes so the mandatory job is never going to happen.

PR #3757, opened 2024-05-25
Format many kinds of files

arxanas:
In particular, this formats documentation and configuration in a consistent way.

I hate to take a dependency on Node just for this, but `prettier` seems like the minimal amount of tooling to adopt to format all these kinds of files.

# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #5636, opened 2025-02-10
Add a Justfile

neongreen:
This lets us run common commands as `just <foo>` and reference them in the docs.

`just` is a popular alternative to managing project-specific commands in Makefiles. The nice parts are:

- Fewer syntax footguns
- Command descriptions and groups

Example:

```
❯ just
Available recipes:
    format               # Format all code
    update-cli-reference # Update generated CLI help (cli/tests/cli-reference@.md.snap)

    [docs]
    build-docs           # Build documentation into rendered-docs/ for offline use
    serve-docs           # Preview documentation with live reloading
```

If there is agreement, I will also update the docs to reference `just` for builds-docs, serve-docs, and update-cli-reference.

# Checklist

If applicable:

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,21 @@
+[private]
+default:
+    @just --list
+  
+# Format all code
+format:
+    cargo +nightly fmt
+
+# Update generated CLI help (cli/tests/cli-reference@.md.snap)
+update-cli-reference:
+    cargo insta test --accept --workspace -- test_generate_md_cli_help
+
+# Preview documentation with live reloading
+[group('docs')]
+serve-docs:
+    uv run mkdocs serve
+
+# Build documentation into rendered-docs/ for offline use
+[group('docs')]
+build-docs:
+    uv run mkdocs build -f mkdocs-offline.yml


arxanas:
How about [mise](https://mise.jdx.dev/) instead? In addition to functioning as a task runner, it should also solve certain problems of installing cross-language tooling (particularly formatters and doc website tooling). This was a problem in https://github.com/jj-vcs/jj/pull/3757 (cc @ilyagr).

In this case it would also be able to provision `uv` for the user. We could probably leverage it to also install `dprint`/`treefmt` and/or any language-specific formatters. It seems to work straightforwardly in CI as well.

PhilipMetzger:
I'm no particular fan of doing this, since it is similar in spirit to adding a editor-specific configuration folder or supporting a Buck2 (#1997) or Bazel build. I don't intend to block it if a majority likes it.

ilyagr:
Indeed, as Waleed said, there was a related conversation around <https://github.com/jj-vcs/jj/pull/3757#issuecomment-2132073794>.

The trouble with `just` is that it doesn't help the user install any of the tools it relies on. I'm also not sure it's easy to set up on Windows. (It might need to be pointed to Git Bash's `sh`?)

I would not change `contributing.md` to rely on `just`, because that would make `just` one more tool that the user needs to install.

There were a few discussions on Discord about possible tools that could do a better job. I don't think we've discussed `mise`, but it might be great, especially if Waleed had some good experience with it. Other possible tools are `proto` or `moon` from https://moonrepo.dev, or trying to overload `uv` to do everything. My hope is that one of these tools would help install everything else, including `uv` and any future tools we might want to use a dev tools (NodeJS-based, especially). One of these tools could be `just`, though any of `mise`/`moon`/`uv` might provide their own command runners for the tools they install.

Justfile
```diff
@@ -0,0 +1,21 @@
+[private]
+default:
+    @just --list
+  
+# Format all code
+format:
+    cargo +nightly fmt
+
```
emesterhazy, line 8: A clippy alias matching CI would also be helpful:

`cargo +stable clippy --all-features --workspace --all-targets`

emesterhazy:
I'm generally in favor of something like this and don't have an opinion about the specific tool we use. It would be nice to have simple ways to run cargo commands (fmt, clippy) using the same flags as the CI pipeline, assuming we can't accomplish that with `.rustfmt.toml` and `.clippy.toml`.

emesterhazy:
Actually, I do have an opinion about what tool to use :D

Even if `just` has simpler syntax and command descriptions, everyone already has `make` installed, so from that perspective `make` is simpler than `just`. With that in mind, I'd suggest either using something that solves the tool provisioning issue that [ilyagr](https://github.com/ilyagr) and [arxanas](https://github.com/arxanas) mentioned, or just using `make` for now. Hopefully we aren't adding so many commands that command descriptions become necessary.

PhilipMetzger:
> Actually, I do have an opinion about what tool to use :D
> 
> Even if `just` has simpler syntax and command descriptions, everyone already has `make` installed, so from that perspective `make` is simpler than `just`. With that in mind, I'd suggest either using something that solves the tool provisioning issue that [ilyagr](https://github.com/ilyagr) and [arxanas](https://github.com/arxanas) mentioned, or just using `make` for now. Hopefully we aren't adding so many commands that command descriptions become necessary.


Make doesn't exist on Windows, so it would hinder portability between OSes. Also I have a heavy personal dislike it.

neongreen:
I ended up trying mise a couple days ago and I’m sold. I will close this PR and create a new one for mise. 

A rather neat thing is that mise tasks can just be .sh scripts, which mise will autodetect.

ilyagr:
> A rather neat thing is that mise tasks can just be .sh scripts, which mise will autodetect.

Does this work on Windows? Can mise install bash or something? 

We really need help with polishing jj on Windows, so I think it's important to make developing there as simple as possible and to not have Windows devs feel like they are an afterthought.

steveklabnik:
Yes, I would really encourage minimizing the use of shell scripts. mingw and friends really make me as a windows user feel like a second class citizen. I don't have them on my computer even.

Normally I would say that `Just` would make the most sense for all of these tools, but mise seems like a huge advantage so that you can maintain other tools' environments too, since jj depends on python as well.

neongreen:
> Does this work on Windows? Can mise install mingw or git for windows or something?

I assume it could be some kind of "sh or powershell" situation, where each script is present as a sh version and a powershell version and mise can choose between them.

Ie. what I'd like is "if you choose, you can run things without needing to install just/mise/etc at all", which can hopefully be orthogonal to the shell choice

PR #5638, opened 2025-02-10
cli split: Test how the working copy commit changes across multiple workspaces

emesterhazy:
When a commit is split, the second commit produced by the split becomes the working copy commit for all workspaces whose working copy commit was the target of the split.

This commit adds two tests for this behavior:

1. Split a commit which is the working copy commit for two workspaces.
2. Split a commit which is the working copy commit for only one of the two workspaces.

These tests check that the working copy commits for the workspaces are updated correctly. Both parallel and non-parallel splits are tested.


emesterhazy:
These new tests are split out of https://github.com/jj-vcs/jj/pull/5618.

martinvonz (commented):
LGTM but I'll leave approval to a non-googler

PR #5575, opened 2025-02-04
cli/tests: move a lot of the cli tests to `gitoxide`

bsdinis:
In keeping with #5548 , this moves a lot of the current `cli` tests to gitoxide.
This doesn't do it all. but thought it best to send this as is, as there is a lot of code to review already (even though most of it is relatively straightforward). Nevertheless, this is what is missing:
 - `push/fetch/clone/remote` require adding git remotes, which is not easy. Maybe when #5553 is merged we can use that machinery directly here
 - `test_git_colocated`: i have it mostly done locally, but not quite
 - `test_git_init`: ditto
 
 This PR also consolidates a lot of what was copy pasted code (namely creating new repos and adding commits to it). This, plus gitoxide's idiosyncrasies make most of the git hashes in the tests change, hence `cargo-insta` was run in a separate commit to avoid the noise. 
 
 The new `gitoxide` helpers in `cli/tests/common/gitoxide.rs` help centralize a lot of common functionality, and make sure we don't run into https://github.com/GitoxideLabs/gitoxide/issues/1829.

Overall, this PR has a lot of commits to help reviewing, but at the end of the day, I think the commit structure should be:
1. comment code commit
2. move test environment to its own file
3. add gitoxide helpers
4. cli/tests: migrate (some) tests to gitoxide

Issue #5548, opened 2025-02-01
Tracking Issue for `git2` deprecation


bsdinis:
There has been a latent sentiment amongst contributors that `jj` has outgrown its need to depend on `libgit2`, and as such we are moving to replace it and eventually deprecate it. This issue tracks the reasoning for this move, as well as the steps needed to make it happen and a deprecation plan.

# Steps
- [x] Implement shelling out to a git subprocess for `push/fetch`. #5228 
- [x] Migrate remote operations with `gitoxide`.
- [x] Migrate git tests to `gitoxide` 
  - #5639 
  - #5757 
  - #5754 
  - #5655 
  - #5813 
  - #5803 
  - #5770 
  - #5846 
- [x] Switch the default `push/fetch` code path to use the `git` subprocess #5652
- [x] Barricade the `git2` code behind a (enabled by default) feature flag #5956
- [x] Enable a deprecation warning for `git2` https://github.com/jj-vcs/jj/pull/6047
- [ ] Remove `git2` https://github.com/jj-vcs/jj/pull/6048

# Outstanding bugs

We list here bugs that have arisen with git subprocessing. Although sometimes `git2` might not be the solution (some of them were use-cases that were wholly blocked by `git2`), it is good to have them registered and try to figure them out before we fully deprecate `git2`.

- authentication problems
  - [ ] #5924 
  - [ ] #5760 
  - [x] #5968 
- git-push ref failures
  - [x] #5978 

# Rationale
## Issues with `git2`

`libgit2` (with `git2` being its Rust bindings) is a project to implement `git` as a library so that external applications can interact programmatically with `git` repos without having to shell out to git and parse its output.

This effort is commendable and `git2` has been crucial in making `jj` git compatible from the get go. I'd like to emphasize that this is by no means a bash on `libgit2`. However, we have found some pain points with its usage:
- Missing support for `SSH`. This regularly comes up when users try to push to Github, with over 20 issues opened about it for `jj` plus multiple reports a week.
- Packaging problems. `libgit2` only supports one version at a time, meaning packagers have a problematic window of time while `jj` has not updated `git2` to its latest version.
- Performance. For the main (current) usage of `libgit2`, which is interacting with `git` remotes, `libgit2` has been reported to be slow.

## Change to what

`jj` actually uses another `git` library to interact with repositories: [`gitoxide`](https://github.com/GitoxideLabs/gitoxide). This means that, as it stands, we have two libraries for interacting with `git`, which is arguably one too many.
[...]

PR #5639, opened 2025-02-10
 cli/tests: move a lot of the cli tests to gitoxide #5575 

bsdinis:
In keeping with #5548 , this moves a lot of the current `cli` tests to `gitoxide`.
This doesn't do it all. but thought it best to send this as is, as there is a lot of code to review already (even though most of it is relatively straightforward). Nevertheless, this is what is missing:

   - push/fetch/clone/remote require adding git remotes, which is not easy. Maybe when     #5553  is merged we can use that machinery directly here

This PR also consolidates a lot of what was copy pasted code (namely creating new repos and adding commits to it). This, plus `gitoxide`'s idiosyncrasies make most of the git hashes in the tests change, hence `cargo-insta` was run in a separate commit to avoid the noise.

The new gitoxide helpers in `cli/tests/common/git.rs` help centralize a lot of common functionality, and make sure we don't run into https://github.com/GitoxideLabs/gitoxide/issues/1829.

Overall, this PR has a lot of commits to help reviewing, but at the end of the day, I think the commit structure should be:

1.   comment code commit
 2.  move test environment to its own file
 3. add gitoxide helpers
  4.  cli/tests: migrate (some) tests to gitoxide

This PR was originally #5575 but Github exists to make my life miserable and I unwittingly removed the branch the PR was based on


cli/tests/test_git_colocated.rs
```diff
@@ -145,7 +120,6 @@ fn test_git_colocated_unborn_bookmark() {
     ◆  0000000000000000000000000000000000000000
     "###);
     // Staged change shouldn't persist.
-    checkout_index();
```
bsdinis, line 148: From @yuja 

We need to either check out the index content to working copy or assert the index contents to ensure that the Git index is reset by jj new root().

yuja: If you're happy to split first few good patches, I'll review and approve them separately.

bsdinis: you mean splice out `git_init`, `git_colocated` and `gitignore`? Can do it

bsdinis: > We need to either check out the index content to working copy or assert the index contents to ensure that the Git index is reset by jj new root().

I think I got this fixed. The way I go about it is by walking the index and remove the added entries. Note that I think this is being double tested. The commit hashes in the snapshots already mostly test this behaviour

bsdinis: Ok, I'm going to open a new PR with the inits colocated and gitignore. This one now has been spliced out of those

bsdinis: #5655

cli/Cargo.toml
```diff
@@ -106,6 +106,12 @@ async-trait = { workspace = true }
 insta = { workspace = true }
 test-case = { workspace = true }
 testutils = { workspace = true }
+gix = { workspace = true, features = [
```
yuja, line 109: nit: insert before `insta`. The list appears to be sorted (except for internal deps.)

cli/tests/test_git_import_export.rs
```diff
@@ -261,13 +272,22 @@ fn get_bookmark_output(test_env: &TestEnvironment, repo_path: &Path) -> String {
...
+        .filter_ok(|git_ref| {
+            matches!(
+                git_ref.name().category(),
+                Some(gix::reference::Category::Tag)
+                    | Some(gix::reference::Category::LocalBranch)
+                    | Some(gix::reference::Category::RemoteBranch),
+            )
+        })
+        .filter_map_ok(|mut git_ref| {
+            let full_name = git_ref.name().as_bstr().to_string();
```
yuja, line 290: nit: this could be `.as_bstr().to_owned()` (i.e. return `BString`) instead of doing lossy conversion here.

yuja (approved):
Thanks!

PR #5640, opened 2025-02-10
cli: status: replace "is clean" by "has no changes"

martinvonz:
"The working copy has no changes" seems much easier to understand for new users. Thanks to @Wilfred for the suggestion.

# Checklist

If applicable:


martinvonz:
We also discussed on Discord whether we should update `(empty)` to `(no changes)` and rename the `empty()` revset and the `empty` template keyword to `has_changes()` and `has_changes` for consistency, but those are larger changes with other concerns (e.g. that "changes" in some of those contexts might be misunderstood to be about change ids).

cli/src/commands/status.rs
```diff
@@ -72,7 +72,7 @@ pub(crate) fn cmd_status(
         let wc_has_changes = tree.id() != parent_tree.id();
         let wc_has_untracked = !snapshot_stats.untracked_paths.is_empty();
         if !wc_has_changes && !wc_has_untracked {
-            writeln!(formatter, "The working copy is clean")?;
+            writeln!(formatter, "The working copy has no changes")?;
```
yuja, line 75: nit: maybe add period as well? I think status/warning messages usually ends with `.` or `:`.

demos/demo_working_copy.sh
```diff
@@ -37,7 +37,7 @@ comment "Note that we were told the working copy is now empty (AKA clean). The
 
 run_command "jj log"
 comment "Let's do a sanity check: 'jj status' should tell us that
-the working copy is clean."
+The working copy has no changes."
```
yuja, line 40: bad edit

lib/tests/test_local_working_copy.rs
```diff
@@ -105,7 +105,7 @@ fn tree_entries(tree: &MergedTree) -> Vec<(RepoPathBuf, Option<MergedTreeValue>)
 
 #[test]
 fn test_root() {
-    // Test that the working copy is clean and empty after init.
+    // Test that The working copy has no changes and empty after init.
```
yuja, line 108: bad edits in this file

yuja (approved):

cli/src/commands/status.rs
martinvonz: Done

PR #5650, opened 2025-02-10
git: pipe `stdin` for authentication

bsdinis:
If `git` requires interactive authentication (e.g., pw protected private keys), it's necessary to passthrough stdin to allow the user to fill in the passphrase.


arxanas:
For git-branchless, I found that Git is not actually asking for the passphrase on stdin, so forwarding stdin didn't seem to work. See https://github.com/arxanas/git-branchless/issues/433

PR #5651, opened 2025-02-10
config-schema: expand options for `ui.editor`, `ui.diff.tool`, and `ui.pager`

### LINKED ISSUE #5617: config schema raises an error when using an array for ui.editor
## Description

As a workaround to #3753 I've configured ui.editor using an array of strings:

```toml
[ui]
editor = ["nvim", "--cmd", "let g:launched_by_shell=1"]
```

But after adding the schema to the top of the config file:

```toml
"$schema" = "https://jj-vcs.github.io/jj/latest/config-schema.json"
```

[Taplo](https://github.com/tamasfe/taplo) raises an error on the `editor` property:
> Even Better TOML: ["nvim","--cmd","let g:launched_by_shell=1"] is not of type "string"

## Specifications

- Platform: macOS 15.3
- Version: jj 0.26.0

steadmon:
Fixes #5617

# Checklist

If applicable:

### DIFF
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -45,6 +45,20 @@
                         "git"
                     ],
                     "default": "diff"
+                },
+                "command-env": {
+                    "type": "object",
+                    "properties": {
+                        "command": {
+                            "type": ["array", "string"],
+                            "items": {
+                                "type": "string"
+                            }
+                        },
+                        "env": {
+                            "type": "object"
+                        }
+                    }
                 }
             },
             "properties": {
@@ -98,7 +112,20 @@
                     "default": "auto"
                 },
                 "pager": {
-                    "type": "string",
+                    "oneOf": [
+                        {
+                            "type": "string"
+                        },
+                        {
+                            "type": "array",
+                            "items": {
+                                "type": "string"
+                            }
+                        },
+                        {
+                            "$ref": "#/properties/ui/definitions/command-env"
+                        }
+                    ],
                     "description": "Pager to use for displaying command output",
                     "default": "less -FRX"
                 },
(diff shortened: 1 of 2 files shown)


cli/src/config-schema.json
```diff
@@ -140,7 +140,10 @@
                             "default": "color-words"
                         },
                         "tool": {
-                            "type": "string",
```
ilyagr, line 143: This seems to work, thanks!

Could you do the same for "pager"? It actually also accepts a third config (an object with `command` and `env` fields; I'm not sure whether or not diff-tool and editor accept these), but we could omit that for now.

steadmon: Done, although I'm a bit iffy about the `command` and `env` fields. From looking at `CommandNameAndArgs` in `cli/src/config.rs` I think we can have either a string or array once again for `command`. `env` is just a hash of strings, which I'm assuming maps to a Table in TOML and an object in json-schema.

It might be nice to have a test that checks the schema actually validates some reasonable example config. Do you think that's worth doing in this PR, or should I file an issue for it?

ilyagr: Thank you!

I thought you would just make the `pager` variable the same as the others. I imagine few people outside jj's builtin config use the object notation.

I think what you have is OK. (Very optionally) you could probably make it better by setting up a schema for the pager object notation in the definitions of the schema and using `oneOf`, there are a few examples of that inside the schema.

> It might be nice to have a test that checks the schema actually validates some reasonable example config. Do you think that's worth doing in this PR, or should I file an issue for it?

That would be wonderful, but I think it can be a separate PR.

ilyagr (approved):

cli/src/config-schema.json
steadmon: Done, thanks!

PR #5665, opened 2025-02-12
cli split: Move bookmark testing to a separate test

emesterhazy:
We are planning to add a config option that controls how bookmarks and change ids move during `jj split` based on feedback in https://github.com/jj-vcs/jj/pull/5618. I think the tests will be more readable after the config option is added if we move the bookmark testing to its own test.

#3419

### DIFF
--- a/cli/tests/test_split_command.rs
+++ b/cli/tests/test_split_command.rs
@@ -18,7 +18,7 @@ use std::path::PathBuf;
 use crate::common::TestEnvironment;
 
 fn get_log_output(test_env: &TestEnvironment, cwd: &Path) -> String {
-    let template = r#"separate(" ", change_id.short(), empty, description, local_bookmarks)"#;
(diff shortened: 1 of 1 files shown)


steadmon (approved):
LGTM

martinvonz:
FYI, @steadmon, @emesterhazy is also a googler and our [policy](https://jj-vcs.github.io/jj/latest/contributing/#code-reviews) is to not approve PRs from other users from the same company unless they're clearly unrelated to the company's interests. I think that is the case here, so there's no issue here. It just reminded me to mention the policy to you, and to let you know that Even is a googler. Which brings up a good question: are you supposed to try to figure out which users are googlers? :) We have an internal page (go/github) where you can usually check, but I also kind of feel like it doesn't matter if they're a googler if you don't even know they are (no idea if you and Even knew of each other). 🤷

emesterhazy:
Sorry about that - I didn't realize steadmon is also from Google!

steadmon:
Ahhh sorry, same here.

PR #5674, opened 2025-02-12
docs: Change the docs color scheme to match the new logo better

emesterhazy:
This is totally unnecessary, but the current indigo color scheme doesn't really match the blue in the new blue jay logo.

Setting totally custom colors in mkdocs is a pain since you need to override lots of different css properties, but just using the built in 'blue' color works pretty well. The 'light blue' color actually matches the logo better, but I was a little concerned about readability for hyperlinks.

The accent color is overridden to 'blue' as well to match better. Note that even though the primary and accent have the same name, they are actually slightly different colors due to the way the material theme works.

Before:
Screenshot 2025-02-12 at 17 09 47

After:
Screenshot 2025-02-12 at 17 07 15

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -20,13 +20,15 @@ theme:
         name: Switch to system preference
     - media: "(prefers-color-scheme: light)"
       scheme: default
-      primary: !ENV [MKDOCS_PRIMARY_COLOR, 'indigo']
+      primary: !ENV [MKDOCS_PRIMARY_COLOR, 'blue']
+      accent: 'blue'
       toggle:
         icon: material/brightness-7
         name: Switch to light mode
     - media: "(prefers-color-scheme: dark)"
       scheme: slate
-      primary: !ENV [MKDOCS_PRIMARY_COLOR, 'indigo']
+      primary: !ENV [MKDOCS_PRIMARY_COLOR, 'blue']
+      accent: 'blue'
       toggle:
         icon: material/brightness-4
         name: Switch to dark mode


ilyagr (approved):
Thanks Evan for thinking of this!

While I appreciate this line of thought, I am conflicted about the PR (I feel a bit bad for saying so). Since it's a stylistic choice, I don't feel like I should impose my feeling on everybody; it's *our* website after all. We can try it and see how people feel (they will only see it after the next release).

-------

For reference, my concerns are:

The new shade of blue is very close to the blue in the logo, but not quite exactly right, which is a bit of an uncanny valley situation. For indigo, it's far enough that it doesn't seem like we're trying to match colors, so the difference is more OK.

Also, I find the color distractingly bright, especially on dark background. It might be a bit better if the text in the header was black, but it's white.

------

My preference would be to try matching the darker shade of blue used for the feathers of the birds exactly. We could also match the bright blue color exactly. But this might be a lot of work. It requires fiddling with much finer CSS settings, and might affect things like what the accent color should be. Ideally, a proper graphics designer would make sure that all the colors are working out. (I assume the defaults for Material theme were chosen in some careful way; they certainly get adjusted when the primary color is adjusted)

emesterhazy:
> My preference would be to try matching the darker shade of blue used for the feathers of the birds exactly. We could also match the bright blue color exactly.

An exact color match would be great, but as you said it's potentially a lot of work. I think just overriding the primary color and accent colors with css gets you most of the way there.

One of the reasons I abandoned the custom CSS though is that I wasn't sure how to make it play nicely with the `MKDOCS_PRIMARY_COLOR` change you added recently. Maybe we can figure something out there, I didn't look closely.

ilyagr:
> One of the reasons I abandoned the custom CSS though is that I wasn't sure how to make it play nicely with the MKDOCS_PRIMARY_COLOR change you added recently.

Makes sense.

-----------

At this point, I'm not sure how much of my opinion is due to a preference for darker colors on my screen, or how common this preference is. It's also possible I'd change my mind after getting used to it.

I think we have a few options:

- Just merge this (I'd still prefer reverting the accent color change, unless there's something I missed about the implications of it)
- Ask more people for their opinion (if you think it's worth the effort and everybody's time)
- Abandon this (this seems only appropriate if people changed their mind after a few days or after reading what I wrote above; at the moment it seems that at least 3 people like this quite a bit)

I think I'll leave it up to you, Evan, to decide (as long as you're OK with making some decision), either between these options or something else. Your guess about what's best is at least as good as mine at this point.

martinvonz:
One option might be to use the new color only for prereleases. That way we can see how we like it without affecting most users.

emesterhazy:
I'm going to revert this since there wasn't a clear consensus about it being better than the current color scheme due to the slight mismatch between the logo colors and the title bar color.

Maybe someone with better design skills will want to take a crack at a totally custom theme that matches the logo colors.

PR #5675, opened 2025-02-13
json schema: Fix bad indentation for `jj fix` settings

emesterhazy:
The fix settings weren't indented consistent with the rest of the settings, causing them to appear in the redacted snapshot in the test_util_config_schema test.

### DIFF
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -617,7 +617,7 @@
                 }
             }
         },
-    "fix": {
+        "fix": {
             "type": "object",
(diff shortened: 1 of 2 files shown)


ilyagr (approved):
Thank you!

cli/src/config-schema.json
martinvonz, line 1: nit on commit message: typo in "intentation"

emesterhazy: Oops! Fixed.

PR #5677, opened 2025-02-13
cli: improve hint message when suggesting `--ignore-immutable`

bryceberger:
There have been a number of users confused about why their commits are immutable, or what to do about it, ex. #5659.

Separately, I feel that the cli is too quick to suggest `--ignore-immutable`, without context of the consequences. A new user could see that the command is failing, see a helpful hint to make it not fail, apply it and move on. This has wildly different consequences, from `jj squash --into someone_elses_branch@origin` rewriting a single commit, to `jj edit 'root()+'` rewriting your entire history.

This commit does two things for any commit that suggests `--ignore-immutable`:

* Adds a short description of what immutable commits are, and a link to the relevant docs, to the hint message.
* Shows the number of immutable commits that would be rewritten if passed.

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -897,14 +897,19 @@ impl WorkspaceCommandEnvironment {
         }
     }
 
+    /// Returns first immutable commit + lower and upper bounds on number of
+    /// immutable commits.
     fn find_immutable_commit<'a>(
         &self,
         repo: &dyn Repo,
         commits: impl IntoIterator<Item = &'a CommitId>,
-    ) -> Result<Option<CommitId>, CommandError> {
+    ) -> Result<Option<(CommitId, usize, Option<usize>)>, CommandError> {
         if self.command.global_args().ignore_immutable {
             let root_id = repo.store().root_commit_id();
-            return Ok(commits.into_iter().find(|id| *id == root_id).cloned());
+            return Ok(commits
+                .into_iter()
+                .find(|id| *id == root_id)
+                .map(|root| (root.clone(), 1, None)));
         }
 
         // Not using self.id_prefix_context() because the disambiguation data
@@ -924,7 +929,21 @@ impl WorkspaceCommandEnvironment {
         let mut commit_id_iter = expression.evaluate_to_commit_ids().map_err(|e| {
             config_error_with_message("Invalid `revset-aliases.immutable_heads()`", e)
         })?;
-        Ok(commit_id_iter.next().transpose()?)
+
+        let Some(first_immutable) = commit_id_iter.next().transpose()? else {
+            return Ok(None);
+        };
+
+        let mut bounds = RevsetExpressionEvaluator::new(
+            repo,
+            self.command.revset_extensions().clone(),
+            &id_prefix_context,
+            self.immutable_expression(),
+        );
+        bounds.intersect_with(&to_rewrite_revset.descendants());
+        let (lower, upper) = bounds.evaluate()?.count_estimate()?;
+
+        Ok(Some((first_immutable, lower, upper)))
     }
 
     /// Parses template of the given language into evaluation tree.
@@ -1798,7 +1817,7 @@ to the current parents may contain changes from multiple commits.
         &self,
         commits: impl IntoIterator<Item = &'a CommitId>,
     ) -> Result<(), CommandError> {
-        let Some(commit_id) = self
+        let Some((commit_id, lower_bound, upper_bound)) = self
             .env
             .find_immutable_commit(self.repo().as_ref(), commits)?
         else {
@@ -1814,10 +1833,18 @@ to the current parents may contain changes from multiple commits.
                 self.write_commit_summary(formatter, &commit)?;
                 Ok(())
             });
-            error.add_hint(
-                "Pass `--ignore-immutable` or configure the set of immutable commits via \
-                 `revset-aliases.immutable_heads()`.",
-            );
+            error.add_hint("Immutable commits are used to protect shared history.");
+            error.add_hint(indoc::indoc! {"
+                For more information, see:
+                      - https://jj-vcs.github.io/jj/latest/config/#set-of-immutable-commits
+                      - `jj help -k config`, \"Set of immutable commits\""});
+
+            let exact = upper_bound == Some(lower_bound);
+            let or_more = if exact { "" } else { " or more" };
+            error.add_hint(format!(
+                "This operation would rewrite {lower_bound}{or_more} immutable commits."
+            ));
+
             error
         };
         Err(error)
--- a/docs/config.md
+++ b/docs/config.md
@@ -343,6 +343,15 @@ To prevent rewriting commits authored by other users:
 Ancestors of the configured set are also immutable. The root commit is always
 immutable even if the set is empty.
 
+Immutable commits (other than the root commit) can be rewritten using the
+`--ignore-immutable` CLI flag.
+
+!!! warning
+
+    Using `--ignore-immutable` will allow you to rewrite any commit in the
+    history, and all descendants, without warning. Use this power wisely, and
+    remember `jj undo`.
+
 ### Behavior of prev and next commands
 
 If you prefer using an "edit-based" workflow, rather than squashing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Help text is now colored (when stdout is a terminal).
 
+* Commands that used to suggest `--ignore-immutable` now print the number of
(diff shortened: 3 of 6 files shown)


cli/src/cli_util.rs
```diff
@@ -1814,10 +1814,32 @@ to the current parents may contain changes from multiple commits.
...
+            );
             error.add_hint(
                 "Pass `--ignore-immutable` or configure the set of immutable commits via \
                  `revset-aliases.immutable_heads()`.",
             );
+
+            let mut expression = self.attach_revset_evaluator(self.env.immutable_expression());
+            expression.intersect_with(&RevsetExpression::commits(vec![commit_id]).descendants());
+            if let Ok(to_visit_revset) = expression.evaluate() {
+                let max_evaluate = 10_000;
```
necauqua, line 1830: This looks like it absolutely must be a config value

cli/src/cli_util.rs
```diff
@@ -1814,10 +1814,32 @@ to the current parents may contain changes from multiple commits.
...
+                let num_descendants = to_visit_revset.iter().take(max_evaluate).count();
+                let or_more = if num_descendants >= max_evaluate {
+                    "+"
+                } else {
+                    ""
+                };
```
necauqua, line 1836: Should be `.take(max_evaluate + 1)`, and the message can be more like "would rewrite 1234 commits"/"would rewrite over {max_evaluate} commits"

cli/src/cli_util.rs
```diff
@@ -1814,10 +1814,32 @@ to the current parents may contain changes from multiple commits.
...
+                error.add_hint(format!(
+                    "Using `--ignore-immutable` would rewrite {num_descendants}{or_more} \
+                     immutable commit(s)."
```
necauqua, line 1839: The `(s)` part is only ever relevant if it would rewrite a single commit? So I'd either check for that and print commit/commits, or (optionally) dont even print the hint for small enough number, maybe 1, maybe also configurable.

cli/src/cli_util.rs
yuja: Perhaps, `count_estimate()` can be used instead.

FWIW, I think we'll need to intersect with all `commits`, not the first immutable commit.

cli/src/cli_util.rs
```diff
@@ -1814,10 +1814,32 @@ to the current parents may contain changes from multiple commits.
...
             });
+            error.add_hint("Immutable commits are used to protect shared history.");
+            error.add_hint(
+                "See https://jj-vcs.github.io/jj/latest/config/#set-of-immutable-commits for more \
+                 information.",
+            );
             error.add_hint(
                 "Pass `--ignore-immutable` or configure the set of immutable commits via \
                  `revset-aliases.immutable_heads()`.",
             );
```
martinvonz, line 1825: I wouldn't mind removing the hint. I think at least @emilazy said the same thing.

cli/src/cli_util.rs
martinvonz: We often simply use plural form even if the number might be 1. I personally don't care

cli/src/cli_util.rs
bryceberger: Discussion on discord is ~[here](https://discord.com/channels/968932220549103686/1307814358906179694/1339667373225148416). I'm in favor of removing it, but if so the linked docs should at least mention the flag.

If we remove the hint, should we also remove the (new) count? If no, will rewrite it to something like:

```
Hint: <description, link, help text>
Hint: Rewriting this commit would rewrite 1234 immutable commits
```

cli/src/cli_util.rs
bryceberger: Note about `count_estimates()`: due to feature unification, when you `cargo run -- ...`, `jj-cli` will be built with the `testing` feature, causing `count_estimate()` to return a max of 10.

This can be "fixed" by running `cargo run -p jj-cli -- ...` instead, which skips building the fake editors that require the `testing` feature.

bryceberger:
I know the second commit doesn't fit the commit guidelines, will squash if accepted. Kept as commit for now to discuss removing the suggestion to use `--ignore-immutable` in the hint text.

cli/src/cli_util.rs
```diff
@@ -1814,10 +1814,32 @@ to the current parents may contain changes from multiple commits.
                 self.write_commit_summary(formatter, &commit)?;
                 Ok(())
             });
+            error.add_hint("Immutable commits are used to protect shared history.");
+            error.add_hint(
+                "See https://jj-vcs.github.io/jj/latest/config/#set-of-immutable-commits for more \
```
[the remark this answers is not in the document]
bryceberger: Put it in a list, because one line was too long. Rendered:
```
Hint: For more information, see:
      - https://jj-vcs.github.io/jj/latest/config/#set-of-immutable-commits
      - `jj help -k config`, "Set of immutable commits"
```

cli/src/cli_util.rs
bryceberger: I like `count_estimate()`.

If it was to be a config value, perhaps `count_estimate()` could use that? `ui.hint-max-revset-evaluate = ...` or something. Sounds like a separate PR. Currently `count_estimate()` is only used for this and the tracking ahead / behind count.

cli/src/cli_util.rs
bryceberger: Change to `count_estimate()` gives a lower and upper bound instead, so new messages are `"would rewrite 1234 immutable commits"` if exact known, `"would rewrite 1234 or more immutable commits"` otherwise.

cli/src/cli_util.rs
yuja: > Note about `count_estimates()`: due to feature unification, when you `cargo run -- ...`, `jj-cli` will be built with the `testing` feature, causing `count_estimate()` to return a max of 10.

That's intentional. For this PR, we can add 10 more immutable commits to test `10 or more` case.

cli/src/cli_util.rs
martinvonz: Agreed, the linked docs should mention the flag. Sorry, I meant to say that before not I forgot.

Updating the hint as you suggest also makes sense.

bryceberger:
No one raised objections about removing the message, so I squashed it into the commit.

docs/config.md
```diff
@@ -343,6 +343,15 @@ To prevent rewriting commits authored by other users:
 Ancestors of the configured set are also immutable. The root commit is always
 immutable even if the set is empty.
 
+Immutable commits (other than the root commit) can be rewritten using the
+`--ignore-immutable` cli flag.
```
martinvonz, line 347: nit: "cli" -> "CLI" (IMO)

cli/src/cli_util.rs
```diff
@@ -924,7 +929,25 @@ impl WorkspaceCommandEnvironment {
...
+        let mut bounds = RevsetExpressionEvaluator::new(
+            repo,
+            self.command.revset_extensions().clone(),
+            &id_prefix_context,
+            self.immutable_expression(),
+        );
+        bounds.intersect_with(&to_rewrite_revset.descendants());
+        let (lower, upper) = bounds
+            .evaluate()
+            .ok()
```
martinvonz, line 946: Should we not propagate an error instead? Same with the `.ok()` below

cli/src/cli_util.rs
```diff
@@ -1814,10 +1837,19 @@ to the current parents may contain changes from multiple commits.
...
+                For more information, see:
+                      - https://jj-vcs.github.io/jj/latest/config/#set-of-immutable-commits
+                      - `jj help -k config`, \"Set of immutable commits\""});
+
+            let exact = upper_bound == Some(lower_bound);
+            let or_more = if exact { "" } else { " or more" };
+            let s = if exact && lower_bound == 1 { "" } else { "s" };
+            error.add_hint(format!(
+                "This operation would rewrite {lower_bound}{or_more} immutable commit{s}."
+            ));
```
martinvonz, line 1851: nit: I prefer to simplify it and always use plural. That's what we generally do. I don't think "rewrite 1 commits" looks strange, perhaps because it's written with a numeric "1" (not "one"), or maybe it's because I'm used to that from Mercurial (I'm also not a native English speaker but my language also has plurals).

Gluing together strings like this is bad for i18n, if we ever want to support that.

cli/src/cli_util.rs
bryceberger: My reasoning for this was that the lower / upper bounds are hints, so it made sense to supply a reasonable default instead of a hard error.

That said, this _shouldn't_ be able to fail, since `self.immutable_expression()` was shown to resolve earlier in the function, so changing it to propagate might reveal a bug instead of hiding it. Will change to propagate

martinvonz (approved):
Thanks!

PR #5690, opened 2025-02-13
cli split: Make `jj split --parallel` set @ to the first revision

emesterhazy:
`jj split --parallel` now sets the working copy revision (`@`) to the first revision created by the split instead of the second. As a result, `@` remains on the same change id before and after the split.

There is no good reason to choose the second commit as the working copy commit after a --parallel split. The only reason it works that way today is because doing that didn't require any code changes compared to the non parallel split.

Martin proposed this change in the review for https://github.com/jj-vcs/jj/pull/5618.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * The `ui.allow-filesets` configuration option has been removed.
   [The "fileset" language](docs/filesets.md) has been enabled by default since v0.20.
 
+* `jj split --parallel` now sets the working copy revision (`@`) to the first
+  revision created by the split instead of the second. As a result, `@` remains on
+  the same change id before and after the split.
+  
+
 ### Deprecations
 
 * This release takes the first steps to make target revision required in
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -221,11 +221,20 @@ The remainder will be in the second commit.
             rewriter.rebase()?.write()?;
             Ok(())
         })?;
-    // Move the working copy commit (@) to the second commit for any workspaces
-    // where the target commit is the working copy commit.
+    // Change the working copy commit as needed.
     for (workspace_id, working_copy_commit) in tx.base_repo().clone().view().wc_commit_ids() {
         if working_copy_commit == commit.id() {
-            tx.repo_mut().edit(workspace_id.clone(), &second_commit)?;
+            // TODO: https://github.com/jj-vcs/jj/issues/3419 - Delete this
+            //   branch when the config setting is removed.
+            if args.parallel && legacy_bookmark_behavior {
+                // We need to do this manually since the legacy behavior
+                // rewrites the target to the second commit.
+                tx.repo_mut().edit(workspace_id.clone(), &first_commit)?;
+            } else if !args.parallel {
+                // Move the working copy commit (@) to the second commit for any
+                // workspaces where the target commit is the working copy commit.
+                tx.repo_mut().edit(workspace_id.clone(), &second_commit)?;
+            }
         }
     }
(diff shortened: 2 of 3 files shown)


yuja:
Since non-parallel `split` bring us to the second commit (= the state before split), it might make sense to move the working copy to a merge of split commits.

emesterhazy:
> Since non-parallel `split` bring us to the second commit (= the state before split), it might make sense to move the working copy to a merge of split commits.

Yeah I actually thought about that as well, but I'm not sure if that's more useful or not. If you're splitting to stash the changes out of line,  sort of like git stash, then it's probably not helpful. If you just want to parallelize the commits then it is.

Curious to hear what other people think. I do like the consistency of preserving the working copy across the split.

ilyagr:
I assumed that if we're talking about keeping the change id and the bookmarks at the first commit for non-parallel split, the working copy would stay there too. It seems I was wrong re #5618 , but would that make sense?

martinvonz:
> Since non-parallel `split` bring us to the second commit (= the state before split), it might make sense to move the working copy to a merge of split commits.

Yeah, I was also wondering if we should do that, but it's also kind of weird that it creates an additional commit. I guess it would be this:

```
$ jj log
o C
|
@ B
|
o A
$ jj split B
$ jj log
C
|\
| | @
+-+-'
| |
o | B2
| |
| o B1
|/
o A
```

It's a bit weird that we end up in a empty commit.

If we think of `jj split --parallel` as `jj split` followed by `jj parallelize`, we would end up with the working copy on the same commit as `jj split` does (which is currently the second, but we're planning to make it end up on the first commit).

yuja:
> It's a bit weird that we end up in a empty commit.

Right.

I think non-parallel `split` (of the working copy) has implication that the user wants to split out unfinished work, so the new commit is edited. I have no idea what is expected for parallel split of `@`.

martinvonz:
> I have no idea what is expected for parallel split of `@`.

I have used it for moving out changes that were in the working copy into a separate change.

emesterhazy:
> If we think of `jj split --parallel` as `jj split` followed by `jj parallelize`, we would end up with the working copy on the same commit as `jj split` does (which is currently the second, but we're planning to make it end up on the first commit).

Yeah, thinking about this again, the current behavior is actually fairly consistent. It might be suboptimal if people are using `jj split --parallel` as an analogue for `git stash` though.

> we would end up with the working copy on the same commit as `jj split` does (which is currently the second, but we're planning to make it end up on the first commit).

Wait are you saying that we are planning to make `jj split` (non --parallel) end up on the first commit? I don't really think that's a great idea for the reason Yuja mentioned - `jj split` sort of implies that the second commit has some unfinished work that the user wants do edit.

martinvonz:
> Wait are you saying that we are planning to make `jj split` (non --parallel) end up on the first commit?

Yes, but I was just confused about all the versions we've discussed. Sorry

emesterhazy:
If we want it to behave like `jj split` followed by `jj parallelize`, then maybe we shouldn't move `@` at all since that's how `jj parallelize` behaves.

I agree that with Yuja that moving `@` to the second commit seems useful for a normal `jj split`, but I'm having trouble justifying it for `jj split --parallel`.

As far as adding an empty child of both the first and second commit, I think it's unnecessary. If the user wants to end up in this state, they can just create the new commit before splitting its parent. As far as other workspaces go, their working copy will remain intact until update-stale is run.

If we decide to create a new empty commit to avoid updating the working copy after `jj split --parallelize`, then we should probably do the same for `jj parallelize` for consistency, and I think it's going to be more involved in that case since the working copy for a workspace can be any of the commits hit by the parallelize.

emesterhazy:
Going to close this PR since there's ample discussion about this and other split related topics in #3419.

PR #5691, opened 2025-02-13
cli split: Recommend `jj config set --user` instead of `--repo` in deprecation warning

emesterhazy:
It only occurred to me after I rebuilt jj and actually had to run the command that we should probably recommend a change to the user config instead of the repo config.

#3419

If applicable:


martinvonz (commented):
LGTM

ilyagr (approved):

PR #5692, opened 2025-02-13
config-schema: validate default configs in tests with taplo

steadmon:
An example failure (after intentionally breaking `cli/src/config/windows.toml`) can be found here:
https://github.com/steadmon/jj/actions/runs/13318732695/job/37198937640#step:7:323

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_config_schema.rs
+++ b/cli/tests/test_config_schema.rs
@@ -1,0 +1,71 @@
+use std::process::Command;
+use std::process::Stdio;
+
+fn taplo_check_config(file: &str) {
+    if Command::new("taplo")
+        .arg("--version")
+        .stdout(Stdio::null())
+        .status()
+        .is_err()
+    {
+        eprintln!("Skipping test because taplo is not installed on the system");
+        return;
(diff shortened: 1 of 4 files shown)


cli/tests/test_config_schema.rs
```diff
@@ -0,0 +1,71 @@
+use std::process::Command;
+use std::process::Stdio;
+
+fn taplo_check_config(file: &str) {
+    if Command::new("taplo")
```
ilyagr, line 5: (No action needed) I would prefer for this to also error out if the CI env var is set and taplo is not found, but it seems like we don't do this for `gpg`. So, I guess we should add it to both in a separate PR.

ilyagr (approved):
LGTM.

I think we should add a few more files purely for testing, with e.g. lists for editors or pagers, but this seems like a good start.

cli/tests/test_config_schema.rs
ilyagr: Good idea, thank you!

PR #5693, opened 2025-02-14
fix: add simple debugging logs

bryceberger:
https://discord.com/channels/968932220549103686/1339741938324340927/1339741938324340927

There's currently no way to tell what `jj fix` is doing. This adds some traces such that `jj fix --debug` will at least tell you what's being run and what it's exit code is, similar to other subprocessing code.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -383,8 +383,9 @@ fn run_tool(
     // associated with.
     let mut vars: HashMap<&str, &str> = HashMap::new();
     vars.insert("path", tool_input.repo_path.as_internal_file_string());
-    let mut child = tool_command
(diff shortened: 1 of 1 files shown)


cli/src/commands/fix.rs
```diff
@@ -397,6 +398,7 @@ fn run_tool(
         Some(child.wait_with_output().or(Err(())))
     })
     .unwrap()?;
+    tracing::debug!(status = %output.status, ?cmd, "fix tool exited:");
```
yuja, line 401: nit: just use debug output? `%output.status` would print something like `"exit status: N"`.

yuja (approved):

Issue #5282, opened 2025-01-06
Some config doesn't work when loaded via --config-file


teatwig:
## Description

It seems like some config such as aliases and ui.default-command don't work if they're loaded via `--config-file`, but do work if they're specified in the regular config file.

Other options like
```toml
[colors]
working_copy = { bold = false }
```
work without issues.

## Steps to Reproduce the Problem

1. Create a new file `jjconfig2.toml` with the following contents:
```toml
[aliases]
l = ["log"]
```
1. Run `jj --config-file=jjconfig2.toml l`

## Expected Behavior

The log is displayed.

## Actual Behavior

```
error: unrecognized subcommand 'l'
```

## Specifications

- Platform: Linux
- Version: 0.25.0

yuja:
Since aliases have to be expanded before parsing arguments, they cannot be defined by `--config-file`/`--config` args. Best we can do here is to show warning if `--config-file` would affect the expanded command.

teatwig:
That makes sense and explains why it's only those options. 

Maybe some background of why I'm trying to do this:

With git I'm always using two config files:

1. My shared config file in `~/.config/git/config` that's synced between all machines 
2. My machine specific config file in `~/.gitconfig` (which can easily be changed with the CLI since it's the default)

Since jj doesn't allow both `~/.jjconfig.toml` and `~/.config/jj/config.toml` to exist at the same time I created an alias that always loads the shared config file as well. 

For now I could probably just switch the files (since not being able to set them via the CLI is not a super big deal), but it would be nice if there was some kind of include mechanism that also supports alias expansion.

yuja:
See https://github.com/jj-vcs/jj/issues/4926#issuecomment-2489837552 for the multiple config files issue. Maybe we can change the default to `~/.config/jj/*.toml`, but for now, you'll need to set `JJ_CONFIG=` env.

PR #5694, opened 2025-02-14
cli: warn when command differs after --config, including when subcommand not found

bryceberger:
Aliases and `ui.default-command` are loaded before `--config` and `--config-file` arguments are parsed (#5282).

There is supposed to be a warning when the result of these arguments would change the parsed args: #5286.

However, that warning was not being printed if the arguments failed to parse due to an unrecognized subcommand. Example:

```bash
# correct:
$ jj --config ui.default-command=nonsense
Warning: Command aliases cannot be loaded from -R/--repository path or --config/--config-file arguments.

# previous, confusing:
$ jj --config aliases.some-alias=nonsense some-alias
error: unrecognized subcommand 'some-alias'

# now:
$ jj --config aliases.some-alias=nonsense some-alias
Warning: Command aliases cannot be loaded from -R/--repository path or --config/--config-file arguments.
error: unrecognized subcommand 'some-alias'
```

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3737,6 +3737,11 @@ impl CliRunner {
                 "--config-toml is deprecated; use --config or --config-file instead."
             )?;
         }
+
+        if args.has_config_args() {
+            warn_if_args_mismatch(ui, &self.app, &config, &string_args)?;
+        }
+
         let (matches, args) = parse_args(&self.app, &self.tracing_subscription, &string_args)
             .map_err(|err| map_clap_cli_error(err, ui, &config))?;
         for process_global_args_fn in self.process_global_args_fns {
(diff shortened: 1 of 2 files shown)


cli/src/cli_util.rs
yuja, line 1: Can you add test for the fixed case?

yuja (approved):
Thanks.

ilyagr:
This is probably not for this PR, but I find this message a bit confusing.

```
$ jj --config ui.default-command=nonsense
Warning: Command aliases cannot be loaded from -R/--repository path or --config/--config-file arguments.
```

Should I file this as an issue, is there some clear way to improve it, or am I confused somehow?

bryceberger:
> This is probably not for this PR, but I find this message a bit confusing.

I think it would be fine to resolve here. Do you think it would be better as a warning and a hint? Something like:

```
Warning: Command parsing differed after finalizing configuration
Hint: Aliases and the default command cannot be set by --config/--config-toml
```

As I'm writing this, I'm having a hard time figuring out what the note about `-R` actually means (I assume conditional config?), so the message probably needs improved somehow.

ilyagr:
That SGTM!

ilyagr:
Should we merge some version of this? I think it's fine to merge it as-is and then have a separate PR clarifying the message in the (rarer) `default-command` case.

PR #5697, opened 2025-02-14
cli split: Move bookmarks to the first commit instead of the second

emesterhazy:
After even more discussion on Discord, we decided to use the new bookmark behavior immediately and only print a warning during `jj split` if the user has opted out of the new behavior using the `split.legacy-bookmark-behavior` config setting.

The reasoning is that if the behavior change breaks someone's workflow or is very disruptive, they are likely to check the changelog and learn about the config option. For users that are not adversely impacted, printing a warning that can only be silenced by changing their config is also disruptive.

#3419

If applicable:

### DIFF
--- a/cli/src/config/misc.toml
+++ b/cli/src/config/misc.toml
@@ -47,3 +47,7 @@ wrapping = "anywhere"
 max-new-file-size = "1MiB"
 auto-track = "all()"
 auto-update-stale = false
+
+# TODO: https://github.com/jj-vcs/jj/issues/3419 - Remove when fully deprecated.
+[split]
+legacy-bookmark-behavior = false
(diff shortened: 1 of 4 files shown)


cli/src/config/misc.toml
```diff
@@ -47,3 +47,7 @@ wrapping = "anywhere"
 max-new-file-size = "1MiB"
 auto-track = "all()"
 auto-update-stale = false
+
+# TODO: https://github.com/jj-vcs/jj/issues/3419 - Remove when the fully deprecated.
```
martinvonz, line 51: nit: delete "the", i guess (or just mark this resolved if you prefer - it doesn't really matter)

martinvonz (commented):
Thanks for the patch and for your patience! This went through a lot of confusing discussion :)

LGTM, but I'm a googler...

PhilipMetzger (approved):
LG, you just need to address Martin's nit.

cli/src/config/misc.toml
emesterhazy: Ah, how aggravating. Fixed now.

Issue #4708, opened 2024-10-24
FR: Generalizable and reusable "rebase behavior"


arxanas:
**Is your feature request related to a problem? Please describe.**

There's been a proliferation of flags added to various commands that create or rebase commits. For example, `--insert-before`/`--insert-after` should be added to many commands, but hasn't been yet. Rebasing is a core competency and reason to use jj. I think we should invest in a unified flexible design to control "rebase behavior" rather than add more flags.

Most commands that create or rebase commits could reuse some of these options, including:

- `jj abandon`
- `jj commit`
- `jj describe`
- `jj duplicate`
- `jj new`
- `jj parallelize`
- `jj rebase`
- `jj split`
- `jj squash`

**Describe the solution you'd like**

Roughly speaking `jj rebase --dest X` (and similar for other commands) should accept an expression representing "rebase behavior" (and also accept a revset for backward compatibility and the common case).

As an example, suppose you're using a "mega-merge" strategy. You have a bunch of commits in `X` that you want to insert before the mega-merge commit as standalone commits. You want to take this opportunity to also write some helpful messages for each commit. If this wouldn't merge cleanly, then you want to abort the operation rather than produce conflicts. An example invocation might look like this:

```console
$ jj rebase -r X -d 'behavior(after=trunk(), before=@, topology="horizontal", empty="keep", message="prompt", merge="abort")'
```

However, if you have a commit in `X` and you need to split it as part of doing the mega-merge, then maybe you'll issue an invocation like this:

```console
$ jj split -r X -d 'behavior(after=trunk(), before=@, topology="horizontal", message="prompt", merge="ours")'
```

To demonstrate the general applicability of this design, see the issues linked below.

Given a commit set `X` that's being rebased, the following are some relevant axes to control rebase behavior:
- **Location**: one or more of the following:
  - "after" revset `Y`: commit set `X` is rebased to be children of `Y`. This is the current behavior of `jj rebase -r X -d Y`.
  - "before" revset `Z`: after rebasing the commit set, the commits in `Z` will be rebased as children of it. This is the current behavior of `jj rebase -r X --insert-before Z`.
  - See: https://github.com/martinvonz/jj/issues/1188, https://github.com/martinvonz/jj/issues/1531, https://github.com/martinvonz/jj/issues/3518, https://github.com/martinvonz/jj/issues/3772, https://github.com/martinvonz/jj/issues/4468
- **Topology**: exactly one of the following:
[...]

Issue #4451, opened 2024-09-13
FR: Command for resetting change ID


scott2000:
**Is your feature request related to a problem? Please describe.**
If there are multiple commits with the same change ID and I want to keep both versions of them for some reason, it can be difficult to resolve the divergent change ID. I think it would be nice if there was a command that just replaces all commits in a revset with identical commits, except with new change IDs.

**Describe the solution you'd like**
I'm not sure what a good name for such a command would be, since `jj reset-change-id` seems too long. Maybe it would be good to have a general command `jj reset` that takes different flags depending on what needs to be reset:

```shell
# Reset change IDs of commits
$ jj reset --change-id <REVSET>
# Reset author name and timestamp of commits
$ jj reset --author <REVSET>
```

This might be confusing with `git reset`, so it might not be a good idea. I do feel like this would be a better place to put "reset author" though, since the current `jj describe --reset-author` feels like an unintuitive place to put it. Another option might be a `change-id` subcommand, but I'm not sure if there's any other operations on change IDs that would be useful.

**Describe alternatives you've considered**
`jj duplicate` followed by `jj abandon` of the original commits works. This could maybe be done with an alias if aliases could execute multiple commands and take arguments to pass to each.

mati865:
How about `jj rehash`, `jj reroll` (this one might not be so obvious), or perhaps something else that intuitively means SHA/ID?
I suppose it should be configurable with flags to reroll only ID, only the hash or both.

scott2000:
I think it is technically not possible to modify the change ID without also modifying the commit ID because of how the change ID is stored for the Git backend (I think if you try it, `jj` will decrement the committer timestamp to cause it to have a new commit ID, but I could be wrong), so it would probably have to update the committer timestamp like the other commands which modify commits do.

mati865:
I don't know jj code, but I've assumed it stores references to Git commits and it'd be possible to only recrate the reference.
Having to change both ID likely won't be a problem; I just imagined it'd be neater if somebody wanted to change only one of them, but I likely overengineered it, and nobody would care.

joyously:
> If there are multiple commits with the same change ID and I want to keep both versions of them for some reason, it can be difficult to resolve the divergent change ID.

Is it really the change ID that is divergent? It seems like the change ID is the thing being managed by the tool, so there are commands to manipulate commits that comprise the change.

yuja:
> I think it is technically not possible to modify the change ID without also modifying the commit ID

[...]

PR #5715, opened 2025-02-15
cli: util: add `clear-predecessors` command

bnjmnt4n:
This will be useful in removing old versions of a revision from the evolog, and eventually from the repository after the operation is abandoned and the repository is garbage collected.

# Checklist

If applicable:


cli/src/commands/util/clear_predecessors.rs
```diff
@@ -0,0 +1,100 @@
...
+    let mut tx = workspace_command.start_transaction();
+    let mut num_rebased = 0;
+    tx.repo_mut().transform_descendants(
+        target_set.iter().copied().cloned().collect(),
+        |rewriter| {
+            if target_set.contains(rewriter.old_commit().id()) {
+                rewriter.reparent().set_predecessors(vec![]).write()?;
+            } else {
+                rewriter.rebase()?.write()?;
+                num_rebased += 1;
```
bnjmnt4n, line 71: I'm not sure if we want to allow preserving predecessors for rebased descendants? It doesn't seem very useful to me (I'm guessing the only reason why people would want this command is for garbage collection), so I can remove it.

yuja:
I think this and resetting change ids are similar, so it's better to add a single command for these requests. Technically they belong to `jj describe`, but I'm not sure if `jj describe --reset-{change-id,predecessors}` makes sense.

https://github.com/jj-vcs/jj/issues/4451

bnjmnt4n:
Those options also seem like they could be part of this list of generalizable rebase options: https://github.com/jj-vcs/jj/issues/4708.

ilyagr:
I think that clearing the predecessors should reset the change id, but resetting the change id doesn't have to necessarily imply clearing the predecessors. `jj duplicate` and `jj split` both preserve the evolog currently.

This seems to be compatible with what everybody is saying, but I wanted to mention it explicitly just in case.

cli/src/commands/util/clear_predecessors.rs
ilyagr: I think it might be nice for `jj clear-predecessors x` to be different from `jj clear-predecessors x::`, but it's also not super-important.

emilazy:
Will it be possible to clear the predecessors of an immutable change? It seems weird if that would change its change ID.

yuja:
> Will it be possible to clear the predecessors of an immutable change?

No, with the current data model. We might want to move predecessors from `Commit` to `Operation` so we can freely discard (and gc) predecessor chains.

yuja:
Adding some debug command seems good to me. It'll be useful in tests to e.g. craft weird predecessor chains or diverged change ids.

yuja:
fwiw, I'm experimenting on storing `predecessors: HashMap<CommitId, Vec<CommitId>>` in operation object. `jj evolog` will have to scan operation history, but I assume the performance would be okay without an index #6405. `jj op log --no-graph -T ''` isn't unacceptably slow (which doesn't include the cost of deserializing `predecessors` table, though.)

bnjmnt4n:
> fwiw, I'm experimenting on storing `predecessors: HashMap<CommitId, Vec<CommitId>>` in operation object. `jj evolog` will have to scan operation history, but I assume the performance would be okay without an index #6405. `jj op log --no-graph -T ''` isn't unacceptably slow (which doesn't include the cost of deserializing `predecessors` table, though.)

Do you have an estimate on when this is ready/what else must be done to land this? If you think it could land in the near future, I won't bother to continue this PR. Otherwise, I think this PR would still be useful.

yuja:
> Do you have an estimate on when this is ready/what else must be done to land this?

I think I can send a PR next week. However, I haven't implemented `jj evolog` part for benchmarking. If the performance was bad, we'll need to find a different design.

Issue #4126, opened 2024-07-19
jj throws a config error after setting "ui.paginate" incorrectly, refuses to correct this setting


CSDUMMI:
## Steps to Reproduce the Problem

1. Run `jj config set --user ui.paginate always`
1. Get <code>Config error: Invalid `ui.paginate`</code>
1. Run `jj config set --user ui.paginate auto`

## Expected Behavior
The configuration is corrected and set to auto, making it possible to use jj again.

## Actual Behavior
All jj commands I tried (`jj log` and `jj config set`) fail with the `Config error` ̇

## Specifications

- Platform: GNU/Linux
- Version:

bnjmnt4n:
This is improved in https://github.com/martinvonz/jj/pull/3852 (available in v0.19): https://github.com/martinvonz/jj/issues/3317#issuecomment-2166606998.

steveklabnik:
This has improved in several ways, but still is ultimately broken:

```text
steveklabnik@Mac work % jj config set --user ui.paginate always
steveklabnik@Mac work % jj log
Config error: Invalid type or value for ui.paginate
Caused by: unknown variant `always`, expected `never` or `auto`

Hint: Check the config file: /Users/steveklabnik/.config/jj/config.toml
For help, see https://jj-vcs.github.io/jj/latest/config/ or use `jj help -k config`.
steveklabnik@Mac work % jj config set --user ui.paginate auto
Config error: Invalid type or value for ui.paginate
Caused by: unknown variant `always`, expected `never` or `auto`

Hint: Check the config file: /Users/steveklabnik/.config/jj/config.toml
For help, see https://jj-vcs.github.io/jj/latest/config/ or use `jj help -k config`.
```

It's nice that this shows the config file location so it can be fixed, but it would be good if we could use `jj config set` to fix this error, and the message is confusing: I'm trying to set it to `auto`, why is that invalid if it's expected?

yuja:
> I'm trying to set it to `auto`, why is that invalid if it's expected?

The error means that the existing config has an invalid value, so jj cannot continue command processing. In this particular case, `Ui` object couldn't be loaded.

It would be nice if `jj config set ui.paginate invalid` can be rejected.

steveklabnik:
Oh yeah, sorry that question was purely rhetorical. Full agreement, it would be nice to reject setting something in an invalid way, and it would be even nicer if you could fix an invalid config by setting it to a valid one.

PR #3234, opened 2024-03-06
cli: use a different symbol for immutable nodes in log

zummenix:
This is a part of #3111. I've introduced a different symbol for immutable nodes. I'm not planning doing configurable node symbols in this PR (mentioned in the issue). And I'm not sure about coloring the commit id differently. I'm open to color suggestions :)

I didn't update tests because there are a lot of them failed and also to avoid opening a massive PR as a draft. I'll update them when the main part of the PR is resolved.

Here are some screenshots with different symbols I've tried:

| `◈` diamond inside (PR) | `✦` four pointed star |
| --- | --- |
| Screenshot 2024-03-06 at 22 26 27 | Screenshot 2024-03-06 at 22 28 38 |

| `◆` diamond | `·` middle dot |
| --- | --- |
| Screenshot 2024-03-06 at 22 29 56 | Screenshot 2024-03-06 at 22 31 09 |

> Using `jj log -r ..` on the repo with `revset-aliases.immutable_heads()="main@upstream | gh-pages@upstream"` in the config.

While diamond symbols look great, with my vision I'd prefer `✦`  or `·` because they are more distinct than the default node symbol.

# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #5716, opened 2025-02-15
ui: add color to help text

gechr:
I appreciate that this might not be to everyone's taste, but I find help text output more readable with colours.

I chose the [default clap v3 colours](https://docs.rs/clap_builder/4.5.29/clap_builder/builder/styling/struct.Styles.html#example) since that's what I've become used to with Rust tools, but happy to tweak as desired, or close this PR entirely if the plain appearance is the `jj` maintainers' preference.

## Before

image

## After

image

## Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Templates now have `format_short_operation_id(id)` function for users to
   customize the default operation id representation.
 
+* Help text is now colored (when stdout is a terminal).
+
 ### Fixed bugs
 
 * `jj status` now shows untracked files under untracked directories.
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -59,6 +59,8 @@ mod workspace;
 
 use std::fmt::Debug;
 
+use clap::builder::styling::AnsiColor;
+use clap::builder::Styles;
 use clap::CommandFactory;
 use clap::FromArgMatches;
 use clap::Subcommand;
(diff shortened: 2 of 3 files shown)


martinvonz:
I personally like the colored styled. Can you remind me why clap4 switched away from it?

martinvonz:
Thanks. I think that's quite well justified, actually. For reference, here's what it says, with some comments from me interleaved:

> The greens and yellows are gone, instead using bold, underline, and ~dimmed~: [PR 4117](https://github.com/clap-rs/clap/pull/4117) though [some more work might be needed to enable colored help, depending on the needs](https://github.com/clap-rs/clap/issues/3108)
> * **See PR for screenshot**
> * Dimmed was removed from placeholders in [fix(help): Don't dim placeholders #4126](https://github.com/clap-rs/clap/pull/4126)
> * Some liked the colors ([example](https://github.com/clap-rs/clap/issues/4132#issuecomment-1229099561)), some hated them ([example](https://github.com/clap-rs/clap/discussions/2906))

We use a lot of colors too and some people hate it. So I don't think this argument applies to jj. (We may want to define some kinds of "color scheme" or way of resetting colors without disabling them, but that's mostly unrelated to this PR.)

> * Color makes it easier to scan

+1

> * Some applications have specific branding that it clashed with ([Make colored help output more compatible with an applications overall theme #2963](https://github.com/clap-rs/clap/issues/2963))
> * True colors would be prettier but requires getting into terminal theme detection to make sure it plays nice and hesistant to do that at the moment
> * **Users can't rollback until theming support is in ([Allow good/warning/error/hint color to be customized #3234](https://github.com/clap-rs/clap/issues/3234)) though they can just disable the coloring with `Command::disable_colored_help(true)`**

These are a bit relevant to us. I don't think the default clap colors conflict much with our color scheme. I don't think users are going to interpret the green commands as good and the yellow headings as warnings, for example. It seems reasonable for us to let the user define the colors in `[colors]` (see https://jj-vcs.github.io/jj/latest/config/#custom-colors-and-styles), however. I don't think that needs to be part of this PR.

So this PR looks good to me, but I'm not going to approve it yet because I think we should give others a chance to comment first.

cli/src/ui.rs
```diff
@@ -46,6 +48,12 @@ use crate::formatter::PlainTextFormatter;
 
 const BUILTIN_PAGER_NAME: &str = ":builtin";
 
+pub const STYLES: Styles = Styles::styled()
+    .header(AnsiColor::Yellow.on_default().bold())
+    .usage(AnsiColor::Yellow.on_default().bold())
+    .literal(AnsiColor::Green.on_default().bold())
+    .placeholder(AnsiColor::Green.on_default());
```
yuja, line 55: nit: I would move this to `commands/mod.rs` or `cli_util.rs`. `Ui` has no dependency on `clap`.

yuja (commented):
I don't have strong feeling against the clap's default nor this PR.

martinvonz (approved):
We heard no objections, so I think we can merge this. Thanks!

PR #5721, opened 2025-02-16
git: remove --bare flag from git commands

### LINKED ISSUE #5700: Unexpected rebase after `jj git fetch` when `git.subprocess=true`

1. Working-copy is based on one of remote branches.
2. `jj git fetch`
3. Working-copy is rebased on the latest `git_head()`.

## Expected Behavior
No rebase happens after `jj git fetch`

## Actual Behavior
Working-copy is rebased on the latest `git_head()`.

## Specifications
Reproduce steps on my local environment:

```
~/data/code/xx> cat .git/HEAD
c7d0c233fa822a621a94eda34eee4c21cd4e9ca7                                                                                                           

~/data/code/xx> jj --config 'git.subprocess=true' git fetch
Nothing changed.

~/data/code/xx> cat .git/HEAD
ref: refs/heads/develop

~/data/code/xx> git rev-parse refs/heads/develop
feb29c4f7bd658a25f11275e7796ce597bde3ab2

~/data/code/xx> jj log
Reset the working copy parent to the new Git HEAD.
...

```

- Platform: `linux`
- Version: `jj 0.26.0-37992412b64f828d078789b2db84d1d291412ab0`

bsdinis:
Fixes #5700

### DIFF
--- a/lib/src/git_subprocess.rs
+++ b/lib/src/git_subprocess.rs
@@ -85,7 +85,6 @@ impl<'a> GitSubprocessContext<'a> {
         // root to Command::current_dir and then pass a relative path to the git
         // dir
         git_cmd
-            .arg("--bare")
             .arg("--git-dir")
             .arg(&self.git_dir)
             // Disable translation and other locale-dependent behavior so we can


martinvonz:
> git: don't rebase automatically after jj git fetch with subprocessing

This worried me until I looked at the commit messages :) I thought it was saying that we simply skip the usual rebasing after `jj git fetch` with subprocessing, but the fix is much more targeted. Good.

bsdinis:
Sorry, didn't mean to alarm. Maybe `git: don't unexpectedly rebase after an empty `jj git fetch` with subprocessing?

lib/src/git_subprocess.rs
```diff
@@ -85,7 +85,6 @@ impl<'a> GitSubprocessContext<'a> {
         // root to Command::current_dir and then pass a relative path to the git
         // dir
         git_cmd
-            .arg("--bare")
```
martinvonz, line 88: The documentation for `--bare` says:

> Treat the repository as a bare repository. If GIT_DIR environment is not set, it is set to the current working directory.

Do we know what "treating it as a bare repository" means? I thought it was just about not having an associated working copy, but it's clearly more than that.

bsdinis: I  don't quite understand it, honestlly.

martinvonz: Also, maybe we can pass `--work-tree=<an empty temporary directory>` or something to make sure that Git doesn't care about working copy? Or are we already sure that it doesn't?

martinvonz: I don't understand these things either. I think we should wait until we've heard from someone who knows. Hopefully one of the people looped in on the bug can tell us.

bsdinis: According to the tips from @tw4452852 , on a bare repository `git fetch` might rewrite HEAD instead of `refs/remotes/<remote>/HEAD`: https://github.com/git/git/blob/e2067b49ecaef9b7f51a17ce251f9207f72ef52d/builtin/fetch.c#L1660-L1662

martinvonz: Yeah, I saw. I just don't know if that's intentional or not.

bsdinis: should we loop in some of the git experts? (I don't actually know who they are)

ilyagr:
Thank you!

I'm not yet completely sure, but I think this PR happens immensely with concurrency mess I was getting when running CLI `jj` commands and VisualJJ at the same time. I thought that got much worse recently, but I didn't realize it had anything to do with using native Git.

I'm guessing that the update to fetch-HEAD caused either VisualJJ or VS Code to reload the repo, which caused and "import git refs" jj operation somehow.

yuja (commented):
I think we agree that removing `--bare` is least bad. Thanks for working on it.

PR #5722, opened 2025-02-16
working_copy: adjust max-size hint to fit all configurations

### LINKED ISSUE #5718: max-new-file-size hint suggests command not applicable to all configurations
## Description

I have `snapshot.auto-track = "none()"` set in my config file. This means I chose to track all files by hand using `jj file track my_file`. Recently I ran into a case where I tried tracking a "large" (3Mb) file, and got this hint:

```
Warning: Refused to snapshot some files:
  mots.txt: 3.4MiB (3615451 bytes); the maximum size allowed is 1.0MiB (1048576 bytes)
Hint: This is to prevent large files from being added by accident. You can fix this by:
  - Adding the file to `.gitignore`
  - Run `jj config set --repo snapshot.max-new-file-size 3615451`
    This will increase the maximum file size allowed for new files, in this repository only.
  - Run `jj --config snapshot.max-new-file-size=3615451 st`
    This will increase the maximum file size allowed for new files, for this command only.
```

Supposing this is a one-time occurrence in this repo, I want to use the last suggestion, which has no effect for me because the file doesn't get auto-tracked!

1. Have `snapshot.auto-track = "none()"` set in your config (or any other fileset different from `all()`)
1. Create a larger than 1Mb file in a `jj` repo, that doesn't match the `snapshot.auto-track` fileset.
1. Run `jj file track big_file`
1. Run the command suggested in the hint
2. The file did not get tracked

## Expected Behavior

I see two possibilities here:

- The hint suggests a command like `jj --config snapshot.max-new-file-size=big_size file track list_of all_concerned files`, which should work for any user regardless of whether they configured `snapshot.auto-track` or not.
- Considering that the user ran `jj file track big_file`, this could be considered as an explicit override of the anti-footgun option. Maybe in that case the file should get tracked either way? In other words, the `max-new-file-size` setting only applies to auto-tracked files, not explicitly tracked ones?

## Actual Behavior

The hint suggests a command that has no effect on my repository:

```
$ jj --config snapshot.max-new-file-size=3615451 st
Untracked paths:
? mots.txt
Working copy : tmtwtukt b54c3e59 (empty) (no description set)
Parent commit: tmnxolkr 17d8a4c5 Initial import
```

## Specifications

- Platform: Linux
- Version: `0.26.0`

alarsyo:
Fixes #5718

This is one of the proposed fixes, the other option needs more discussion I think.

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1957,8 +1957,13 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
         locked_ws
             .finish(self.user_repo.repo.op_id().clone())
             .map_err(snapshot_command_error)?;
-        print_snapshot_stats(ui, &stats, &self.env.path_converter)
-            .map_err(snapshot_command_error)?;
+        print_snapshot_stats(
+            ui,
+            &stats,
+            &self.env.path_converter,
+            SnapshotContext::Automatic,
+        )
+        .map_err(snapshot_command_error)?;
         Ok(stats)
     }
 
@@ -2645,10 +2650,17 @@ pub fn print_conflicted_paths(
     Ok(())
 }
 
+pub enum SnapshotContext {
+    Automatic,
+    Track,
+    Untrack,
+}
+
 pub fn print_snapshot_stats(
     ui: &Ui,
     stats: &SnapshotStats,
     path_converter: &RepoPathUiConverter,
+    context: SnapshotContext,
 ) -> io::Result<()> {
     // Paths with UntrackedReason::FileNotAutoTracked shouldn't be warned about
     // every time we make a snapshot. These paths will be printed by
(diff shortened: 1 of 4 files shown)


cli/src/cli_util.rs
```diff
@@ -2697,14 +2697,19 @@ pub fn print_snapshot_stats(
         })
         .max()
     {
+        let large_files = stats
+            .untracked_paths
+            .keys()
+            .map(|p| path_converter.format_file_path(p))
+            .join(" ");
```
PhilipMetzger, line 2704: Nit: I think the `jj file track` hint should only be showed when `snapshot.autotrack != "all()"` as otherwise `jj st` still is applicable.

PhilipMetzger (commented):
there's a minor discussion point otherwise it looks good to me.

cli/src/cli_util.rs
martinvonz: I don't think we should get a `FileTooLarge` for paths that are also `FileNotAutoTracked`, so if a path is too large, it means the file will be auto-tracked.

martinvonz: Wait, I was confused. The whole reason for the bug report and this patch is that we display this hint also when you are not auto-tracking all paths. This hint gets printed because we temporarily try to snapshot just the specified files. Maybe we should print a different message in `src/commands/file/track.rs` instead?

alarsyo: Should replace the call to `print_snapshot_stats` with custom logic in `track.rs`? Or pass a new argument to `print_snapshot_stats` to specify when it gets called from `track`?

martinvonz: Good question. I was thinking of replacing it, but it looks like most of it is relevant also for `jj file track`. Passing a flag is a bit ugly but it seems good enough to me. What do you think?

alarsyo: I've added a flag to `print_snapshot_stats`. I'm not really satisfied with how it looks, most of the hint message is duplicated, maybe I should print it in several parts, gated by the flag?

cli/src/cli_util.rs
```diff
@@ -2697,14 +2697,19 @@ pub fn print_snapshot_stats(
         })
         .max()
     {
+        let large_files = stats
```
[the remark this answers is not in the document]
alarsyo: Yes, I'd forgotten the filtering part, that should be fixed now!

cli/tests/test_working_copy.rs
```diff
@@ -30,33 +30,57 @@ fn test_snapshot_large_file() {
...
+      - Run `jj --config snapshot.max-new-file-size=11264 st
+        This will increase the maximum file size allowed for new files, for this command only.
+    Warning: Refused to snapshot some files:
+      large: 11.0KiB (11264 bytes); the maximum size allowed is 10.0KiB (10240 bytes)
+      large2: 11.0KiB (11264 bytes); the maximum size allowed is 10.0KiB (10240 bytes)
+    Hint: This is to prevent large files from being added by accident. You can fix this by:
+      - Run `jj config set --repo snapshot.max-new-file-size 11264`
+        This will increase the maximum file size allowed for new files, in this repository only.
+      - Run `jj --config snapshot.max-new-file-size=11264 file track large large2`
+        This will increase the maximum file size allowed for new files, for this command only.
```
alarsyo, line 82: This shows a weird edge-case: calling `file track` on a large file when it's not needed (i.e., the file is in the `auto-track` fileset and would've been tracked anyways) will result in a double output like that, one for the "auto snapshotting" which occurs for all commands, and one specific to `file track`

martinvonz: Seems fine to leave as is for now. Maybe `track.rs` should subtract `auto_tracking_matcher()` from the matcher it passes into `snapshot_options_with_start_tracking_matcher()`

martinvonz (approved):
Thanks!

martinvonz:
You should have an invitation to join the jj-vcs team in your in email. Maybe it ended up in spam? Once you've joined the team and resolved the comments, you can merge the PR.

alarsyo:
Thanks! Had my GitHub email settings a bit messed up, it landed on an old address :)

PR #5724, opened 2025-02-17
op_store: propagate initialization error

yuja:
# Checklist

If applicable:


lib/src/simple_op_store.rs
```diff
@@ -93,10 +105,13 @@ impl SimpleOpStore {
     }
 
     /// Creates an empty OpStore, panics if it already exists
```
bnjmnt4n, line 107: Minor nit: this comment can be updated since it no longer panics

bnjmnt4n (approved):

lib/src/simple_op_store.rs
yuja: Good catch. Fixed.

PR #5725, opened 2025-02-17
cli: parse bookmark name to be created/updated as revset symbol

### LINKED ISSUE #5705: `jj bookmark create` allows creating unreferenceable bookmarks
Using https://github.com/jj-vcs/jj/commit/514a009675c59588ac4b2eb6abc4d21bd48e2b52

---

Repro:

```
$ jj bookmark create foo:bar
Warning: Failed to export some bookmarks:
  foo:bar: Failed to set: A reference must be a valid tag name as well: A ref must not
  contain invalid bytes or ascii control characters: ":"
```

At this point, the bookmark is still created, despite the error:

```
○  yxlrwnxw zx8@example.com 2025-02-14 15:54:52 foo:bar 7d98d8bc
```

If I try to reference that bookmark, it does not work:

```
$ jj show foo:bar
Error: Failed to parse revset: Modifier `foo` doesn't exist
Caused by:  --> 1:1
  |
1 | foo:bar
  | ^-^
  |
  = Modifier `foo` doesn't exist
```

If I try to delete the bookmark:

```
$ jj bookmark delete foo:bar
error: invalid value 'foo:bar' for '<NAMES>...': Invalid string pattern kind `foo:`
```

If I use `exact:foo:bar`, it does work.

yuja:
Closes #5705

cc @0xdeafbeef 

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `Commit`. All methods on `Commit` can be accessed with `commit.method()`, or
   `self.commit().method()`.
 
+* Bookmark name to be created/updated is now parsed as [a revset
+  symbol](docs/revsets.md#symbols). Quotation may be needed in addition to shell
+  quotes. Example: `jj bookmark create -r@- "'name with space'"`
+
 * `jj bookmark forget` now untracks any corresponding remote bookmarks instead
   of forgetting them, since forgetting a remote bookmark can be unintuitive.
   The old behavior is still available with the new `--include-remotes` flag.
- 
+
 * `jj split` no longer moves bookmarks to the second revision created by the
   split. Instead, local bookmarks associated with the target revision will move
   to the first revision created by the split (which inherits the target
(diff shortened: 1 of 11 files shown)


ilyagr:
(Thinking out loud, not sure whether this is actionable) I think such a bookmark could also appear when fetched from a git remote (unless all valid git branch names are valid revset symbols). So, if the goal is to make the user aware about the necessity of quoting, this is not a perfect solution.

Another approach would be to warn when listing/printing such bookmarks, or (perhaps even better) to print them in their quoted form. We'd then have to be consistent and *always* print bookmark names in their quoted forms, though.

yuja:
> or (perhaps even better) to print them in their quoted form.

Yes, we need to fix the default template output and various status messages to use revset-compatible form.

jonathantanmy (approved):
For what it's worth, looks good to me. I don't know of a good way to make sure that we always display bookmark names as quoted whenever quotation is needed, but I don't think that needs to block this PR.

yuja:
> My version had tips for `bookmark@remote` cases. Maybe it's valuable for people migrated from git

I'll address this separately.

PR #5733, opened 2025-02-17
cli: accept remote bookmark patterns in `bookmark forget`

scott2000:
Follow-up to #5621. Allows `BOOKMARK@REMOTE` patterns in `bookmark forget`. Another alternative might be `--remote REMOTE`.

TODO: add tests for forgetting `BOOKMARK@REMOTE` without forgetting `BOOKMARK`.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/bookmark/forget.rs
+++ b/cli/src/commands/bookmark/forget.rs
@@ -18,7 +18,10 @@ use jj_lib::op_store::LocalRemoteRefTarget;
 use jj_lib::op_store::RefTarget;
 use jj_lib::op_store::RemoteRef;
 use jj_lib::ref_name::RefName;
+use jj_lib::ref_name::RemoteRefSymbol;
+use jj_lib::ref_name::RemoteRefSymbolBuf;
 use jj_lib::repo::Repo as _;
+use jj_lib::str_util::StringExpression;
 use jj_lib::view::View;
 
 use super::warn_unmatched_local_or_remote_bookmarks;
@@ -26,7 +29,7 @@ use crate::cli_util::CommandHelper;
 use crate::cli_util::default_ignored_remote_name;
 use crate::command_error::CommandError;
 use crate::complete;
-use crate::revset_util::parse_union_name_patterns;
+use crate::revset_util::parse_name_patterns_or_remote_symbols;
 use crate::ui::Ui;
 
 /// Forget a bookmark without marking it as a deletion to be pushed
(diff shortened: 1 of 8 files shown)


cli/src/commands/bookmark/forget.rs
scott2000, line 1: I think allowing `BOOKMARK@REMOTE` is a better option than `--remote REMOTE`, since the meaning of `jj bookmark forget BOOKMARK --remote REMOTE` seems less clear to me:

* If the local bookmark `BOOKMARK` exists, it's unclear whether this command would delete both `BOOKMARK` and `BOOKMARK@REMOTE`, or only `BOOKMARK@REMOTE`.
* Since the default is to untrack all remote bookmarks, a user might expect `--remote REMOTE` to mean "only untrack the remote bookmark from REMOTE" (i.e. they might expect it to be the same as `jj bookmark delete BOOKMARK` followed by `jj bookmark untrack BOOKMARK@REMOTE`, when really it's an entirely different meaning).

I also think being able to `jj bookmark forget BOOKMARK@REMOTE` would be convenient in cases where a remote bookmark was deleted, and you want to skip fetching the deletion since it would cause some commits to be abandoned, so it seems useful to support this feature.

We could also decide that this feature isn't important to implement for now, since it seems like it would mainly be useful for advanced users and it could confuse regular users.

yuja: > * If the local bookmark `BOOKMARK` exists, it's unclear whether this command would delete both `BOOKMARK` and `BOOKMARK@REMOTE`, or only `BOOKMARK@REMOTE`.
> * Since the default is to untrack all remote bookmarks, a user might expect `--remote REMOTE` to mean "only untrack the remote bookmark from REMOTE" (i.e. they might expect it to be the same as `jj bookmark delete BOOKMARK` followed by `jj bookmark untrack BOOKMARK@REMOTE`, when really it's an entirely different meaning).

I agree about these, and I'm not against adding `forget BOOKMARK@REMOTE`.

What I'm not certain about is how the pattern syntax will be evolved. Suppose we add compound string expression something like `bookmarks("foo" & ~glob:"bar/*")`, it makes sense to parse CLI name patterns in the same way. If the pattern argument accepts both local and remote symbols, they can be combined as `"foo | bar@baz .."`. That's technically doable, but seems weird.

scott2000:
I'm closing this because I don't think it's relevant anymore now that `jj bookmark` commands no longer support `bookmark@remote` syntax. We can always add `--remote` to `jj bookmark forget` instead if it's requested.

0WD0:
I think we should reopen this pr.

> I'm closing this because I don't think it's relevant anymore now that `jj bookmark` commands no longer support `bookmark@remote` syntax.

Because https://github.com/jj-vcs/jj/pull/9641 and https://github.com/jj-vcs/jj/pull/9874 reverted this.

PR #5734, opened 2025-02-17
track: only print snapshot stats and hints once

alarsyo:
Since every command first runs an automatic snapshot, we don't need to explicitly snapshot a file that would've been matched by the auto-matcher previously. This would in some cases lead to double hints being printed (see snapshot test case).

This is a quick followup to https://github.com/jj-vcs/jj/pull/5722#discussion_r1958525279

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -413,11 +413,22 @@ impl CommandHelper {
     /// Loads workspace and repo, then snapshots the working copy if allowed.
     #[instrument(skip(self, ui))]
     pub fn workspace_helper(&self, ui: &Ui) -> Result<WorkspaceCommandHelper, CommandError> {
-        Ok(self.workspace_helper_with_stats(ui)?.0)
+        let (workspace_command, stats) = self.workspace_helper_with_stats(ui)?;
+        print_snapshot_stats(
+            ui,
+            &stats,
+            workspace_command.env().path_converter(),
+            SnapshotContext::Automatic,
+        )?;
+        Ok(workspace_command)
     }
 
     /// Loads workspace and repo, then snapshots the working copy if allowed and
     /// returns the SnapshotStats.
+    ///
+    /// Note that unless you have a good reason not to do so, you should always
+    /// call [`print_snapshot_stats`] with the [`SnapshotStats`] returned by
+    /// this function to present possible untracked files to the user.
     #[instrument(skip(self, ui))]
     pub fn workspace_helper_with_stats(
         &self,
@@ -505,6 +516,9 @@ impl CommandHelper {
             .map_err(|err| map_workspace_load_error(err, None))
     }
 
+    /// Note that unless you have a good reason not to do so, you should always
+    /// call [`print_snapshot_stats`] with the [`SnapshotStats`] returned by
+    /// this function to present possible untracked files to the user.
     pub fn recover_stale_working_copy(
         &self,
         ui: &Ui,
@@ -521,7 +535,9 @@ impl CommandHelper {
                 // operation, then merge the divergent operations. The wc_commit_id of the
                 // merged repo wouldn't change because the old one wins, but it's probably
                 // fine if we picked the new wc_commit_id.
-                let stats = workspace_command.maybe_snapshot(ui)?;
+                let stats = workspace_command
+                    .maybe_snapshot_impl(ui)
+                    .map_err(|err| err.into_command_error())?;
 
                 let wc_commit_id = workspace_command.get_wc_commit_id().unwrap();
                 let repo = workspace_command.repo().clone();
@@ -1063,6 +1079,9 @@ impl WorkspaceCommandHelper {
         }
     }
 
+    /// Note that unless you have a good reason not to do so, you should always
+    /// call [`print_snapshot_stats`] with the [`SnapshotStats`] returned by
+    /// this function to present possible untracked files to the user.
     #[instrument(skip_all)]
     fn maybe_snapshot_impl(&mut self, ui: &Ui) -> Result<SnapshotStats, SnapshotWorkingCopyError> {
         if !self.may_update_working_copy {
@@ -1090,9 +1109,17 @@ impl WorkspaceCommandHelper {
     /// Snapshot the working copy if allowed, and import Git refs if the working
     /// copy is collocated with Git.
     #[instrument(skip_all)]
-    pub fn maybe_snapshot(&mut self, ui: &Ui) -> Result<SnapshotStats, CommandError> {
-        self.maybe_snapshot_impl(ui)
-            .map_err(|err| err.into_command_error())
+    pub fn maybe_snapshot(&mut self, ui: &Ui) -> Result<(), CommandError> {
+        let stats = self
+            .maybe_snapshot_impl(ui)
+            .map_err(|err| err.into_command_error())?;
+        print_snapshot_stats(
+            ui,
+            &stats,
+            self.env().path_converter(),
+            SnapshotContext::Automatic,
+        )?;
+        Ok(())
     }
 
     /// Imports new HEAD from the colocated Git repo.
@@ -1272,7 +1299,8 @@ to the current parents may contain changes from multiple commits.
         locked_ws.finish(repo.op_id().clone())?;
         self.user_repo = ReadonlyUserRepo::new(repo);
 
-        self.maybe_snapshot(ui)
+        self.maybe_snapshot_impl(ui)
+            .map_err(|err| err.into_command_error())
     }
 
     pub fn workspace_root(&self) -> &Path {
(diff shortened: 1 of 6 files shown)


alarsyo:
I'm not sure if this deserves its own new test case, how exhaustive do we want snapshot tests to be for CLI/UI stuff?

cli/src/commands/file/track.rs
```diff
@@ -45,9 +46,12 @@ pub(crate) fn cmd_file_track(
     args: &FileTrackArgs,
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
```
yuja, line 48: Maybe we can instead use `workspace_helper_with_stats()` and combine/suppress the stats. We'll also need to move `print_snapshot_stats()` to callers.

alarsyo: That seems cleaner, I'll work on that!

`print_snapshot_stats` now shows a different hint depending on whether the original command was an automatic snapshot or an explicit `track` command (See #5722). It does so thanks to a `SnapshotContext` passed to it. If we combine the stats, this context will need to be associated to each `untracked_path` instead of the entire stats. 
However this means moving `SnapshotContext` into jj-lib rather than jj-cli, and kind of leaks CLI concepts (`SnapshotContext::Track`, `SnapshotContext::Untrack`).

The other solution would be splitting the "standard" `print_snapshot_stats` implementation: keep one function for general automated snapshots, have a new function specific to the `track` command, and that one can handle showing a more specific hint depending on the reason why the file is tracked.

I feel like the second solution is cleaner. What do you think?

alarsyo: Reading your comment again I think my second solution is exactly what you meant here 

> We'll also need to move print_snapshot_stats() to callers.

alarsyo: I've now moved the responsibility of calling `print_snapshot_stats` to callers of `workspace_helper_with_stats` :)

cli/src/commands/file/track.rs
```diff
@@ -44,26 +52,90 @@ pub(crate) fn cmd_file_track(
...
+        .into_iter()
+        // focus on files that are now tracked with `file track`
+        .filter(|(_, reason)| matches!(reason, UntrackedReason::FileNotAutoTracked).not())
+    {
+        // if the path was previously rejected because it wasn't tracked, update its
+        // reason
+        merged_untracked_paths
+            .entry(path)
+            .and_modify(|other_reason| *other_reason = reason.clone())
+            .or_insert(reason);
```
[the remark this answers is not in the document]
alarsyo: Ohhh my bad I thought `insert()` did *not* update the value if one was already present. Fixing!

cli/src/commands/file/track.rs
```diff
@@ -44,26 +52,90 @@ pub(crate) fn cmd_file_track(
...
+        // care about
+        return print_snapshot_stats(ui, &auto_stats, path_converter);
+    }
+
+    let mut merged_untracked_paths = auto_stats.untracked_paths;
+    for (path, reason) in track_stats
+        .untracked_paths
+        .into_iter()
+        // focus on files that are now tracked with `file track`
+        .filter(|(_, reason)| matches!(reason, UntrackedReason::FileNotAutoTracked).not())
```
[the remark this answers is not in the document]
alarsyo: Oh I thought it would be more readable this way actually, the `!matches!` looked weird to me. I'll change it if that's the convention here :)

cli/src/cli_util.rs
```diff
@@ -1091,8 +1098,16 @@ impl WorkspaceCommandHelper {
     /// copy is collocated with Git.
     #[instrument(skip_all)]
     pub fn maybe_snapshot(&mut self, ui: &Ui) -> Result<SnapshotStats, CommandError> {
-        self.maybe_snapshot_impl(ui)
-            .map_err(|err| err.into_command_error())
+        let stats = self
+            .maybe_snapshot_impl(ui)
+            .map_err(|err| err.into_command_error())?;
+        print_snapshot_stats(
```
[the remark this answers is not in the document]
alarsyo: Ah yes you're right. I guess this means callers of `maybe_snapshot() need to explicitly print the stats as well.

I wanted to avoid this as much as possible as it means anyone can write a code path that never prints the stats if they forget to call `print_snapshot_stats`...

I've pushed a new patch with the changes. Perhaps we could use `#[must_use]` on these functions?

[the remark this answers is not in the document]
yuja: Perhaps, we can instead replace internal callers of `maybe_snapshot()` with `maybe_snapshot_impl()`, and add public `maybe_snapshot()` that doesn't return `SnapshotStats`?

It'll clarify whether the caller has responsibility to print the stats.

> Perhaps we could use `#[must_use]` on these functions?

Unfortunately, we can't attribute the inner value of `Result<_, _>` must be used.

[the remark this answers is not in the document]
alarsyo: Done!

cli/src/commands/file/track.rs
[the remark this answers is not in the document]
alarsyo: Fixed!

cli/src/cli_util.rs
```diff
@@ -413,11 +413,23 @@ impl CommandHelper {
...
+            workspace_command.env().path_converter(),
+            SnapshotContext::Automatic,
+        )?;
+        Ok(workspace_command)
     }
 
     /// Loads workspace and repo, then snapshots the working copy if allowed and
     /// returns the SnapshotStats.
+    ///
+    /// Note that unless you have a good reason to do so, you should always call
```
martinvonz, line 424: nit: missing a "not" somewhere (e.g. "unless you have a good reason not to")

alarsyo: Thanks, fixed :)

cli/src/cli_util.rs
```diff
@@ -1272,7 +1299,16 @@ to the current parents may contain changes from multiple commits.
...
+        let stats = self
+            .maybe_snapshot_impl(ui)
+            .map_err(|err| err.into_command_error())?;
+        print_snapshot_stats(
+            ui,
+            &stats,
+            self.env().path_converter(),
+            SnapshotContext::Automatic,
+        )?;
+        Ok(stats)
```
yuja: Perhaps, this `stats` will be printed twice?

cli/src/cli_util.rs
```diff
@@ -2731,48 +2710,29 @@ pub fn print_snapshot_stats(
...
-                  - Run `jj --config snapshot.max-new-file-size={size} st`
-                    This will increase the maximum file size allowed for new files, for this command only.
-                "
-            )?,
-        };
+    let large_files_sizes =
+        stats
+            .untracked_paths
+            .iter()
+            .filter_map(|(_path, reason)| match reason {
```
yuja, line 2720: nit: `.values().filter_map(|reason|`

cli/src/commands/file/track.rs
```diff
@@ -12,13 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::io;
 use std::io::Write;
+use std::ops::Not;
```
yuja, line 17: nit: remove `Not`

cli/src/commands/file/track.rs
```diff
@@ -44,26 +52,87 @@ pub(crate) fn cmd_file_track(
...
+    if track_stats
+        .untracked_paths
+        .iter()
+        .any(|(path, _)| track_matcher.matches(path))
+        .not()
+    {
+        // we can use the "normal" report path, nothing special happened to a file we
+        // care about
+        return print_snapshot_stats(ui, &auto_stats, path_converter);
+    }
```
yuja, line 97: Remove this? I don't think we need to use the "normal" message here because the user intentionally tried to start tracking some new files.

cli/src/commands/file/track.rs
```diff
@@ -44,26 +52,87 @@ pub(crate) fn cmd_file_track(
...
+
+    print_untracked_files(ui, &merged_untracked_paths, path_converter)?;
+
+    let large_files = merged_untracked_paths
+        .iter()
+        .filter_map(|(path, reason)| match reason {
+            UntrackedReason::FileTooLarge { size, .. } => Some((path, size)),
+            UntrackedReason::FileNotAutoTracked => None,
+        })
+        .collect_vec();
```
yuja, line 119: nit: `.unzip()` also works, but I don't have any preference.

yuja (approved):
Thanks.

cli/src/cli_util.rs
alarsyo: Ah yes you're right! Nice catch

alarsyo: Fixed!

alarsyo:
Thanks for the helpful review :)

PR #5735, opened 2025-02-17
merge_tools: use `ExternalToolError` for external tools

scott2000:
I think this is what was suggested in https://github.com/jj-vcs/jj/pull/5111#discussion_r1904820097

# Checklist

If applicable:

### DIFF
--- a/cli/src/merge_tools/external.rs
+++ b/cli/src/merge_tools/external.rs
@@ -20,6 +20,7 @@ use jj_lib::matchers::Matcher;
 use jj_lib::merge::Merge;
 use jj_lib::merged_tree::MergedTree;
 use jj_lib::merged_tree::MergedTreeBuilder;
+use jj_lib::repo_path::InvalidRepoPathError;
 use jj_lib::repo_path::RepoPathUiConverter;
 use jj_lib::store::Store;
 use jj_lib::working_copy::CheckoutOptions;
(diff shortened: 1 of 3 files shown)


cli/src/merge_tools/external.rs
```diff
@@ -170,8 +170,17 @@ pub enum ExternalToolError {
...
+         see the exact invocation)."
+    )]
+    EmptyOrUnchanged,
+    #[error(transparent)]
+    InvalidRepoPath(#[from] InvalidRepoPathError),
+    #[error(transparent)]
+    Backend(#[from] jj_lib::backend::BackendError),
     #[error("I/O error")]
-    Io(#[source] std::io::Error),
+    Io(#[from] std::io::Error),
```
yuja, line 183: This is source of bugs because there are other `io::Error` variants. I'm not pretty sure what this `Io` error represents, but it's not the error of console I/O.

yuja (commented):
> I think this is what was suggested in [#5111 (comment)](https://github.com/jj-vcs/jj/pull/5111#discussion_r1904820097)

What I meant was to split recoverable (or expected) and non-recoverable errors. The former will be mapped to `MergeToolPartialResolutionError`. For example, `FailedToExecute` is non-recoverable because it doesn't make sense to partially accept the merge result on configuration error. `ToolAborted` is recoverable error as it means merge failed.

FWIW, I think it's better (depending on the use case) to continue merging on recoverable error, and print a list of failed merges.

PR #5736, opened 2025-02-18
git: handle multi-component remotes

### LINKED ISSUE #5731: `jj git fetch` can’t handle remotes with slashes in the name
## Description

As issue title. This is with the (newly‐default) `git.subprocess`; I don’t know whether `git2` is affected or not.

```
yuyuko:/v/f/y/m/T/tmp.nn8ESRTV6z
❭ jj git init --colocate
Initialized repo in "."

yuyuko:/v/f/y/m/T/tmp.nn8ESRTV6z  (empty) (no description set)
❭ jj git remote add test/jj https://github.com/jj-vcs/jj.git

yuyuko:/v/f/y/m/T/tmp.Dba7YWIbpU  (empty) (no description set)
❭ jj git fetch --remote test/jj --bookmark main --debug
2025-02-17T17:02:34.661492Z  INFO jj_cli::cli_util: debug logging enabled
2025-02-17T17:02:34.668605Z DEBUG run_command:cmd_git_fetch{args=GitFetchArgs { branch: [Exact("main")], remotes: ["test/jj"], all_remotes: false }}:workspace_helper:workspace_helper_with_stats:maybe_snapshot_impl:snapshot_working_copy:base_ignores: globset: glob converted to regex: Glob { glob: "**/._*", re: "(?-u)^(?:/?|.*/)\\._[^/]*$", opts: GlobOptions { case_insensitive: false, literal_separator: true, backslash_escape: true, empty_alternates: false }, tokens: Tokens([RecursivePrefix, Literal('.'), Literal('_'), ZeroOrMore]) }
2025-02-17T17:02:34.668633Z DEBUG run_command:cmd_git_fetch{args=GitFetchArgs { branch: [Exact("main")], remotes: ["test/jj"], all_remotes: false }}:workspace_helper:workspace_helper_with_stats:maybe_snapshot_impl:snapshot_working_copy:base_ignores: globset: glob converted to regex: Glob { glob: "**/.*.kak.*", re: "(?-u)^(?:/?|.*/)\\.[^/]*\\.kak\\.[^/]*$", opts: GlobOptions { case_insensitive: false, literal_separator: true, backslash_escape: true, empty_alternates: false }, tokens: Tokens([RecursivePrefix, Literal('.'), ZeroOrMore, Literal('.'), Literal('k'), Literal('a'), Literal('k'), Literal('.'), ZeroOrMore]) }
2025-02-17T17:02:34.668654Z DEBUG run_command:cmd_git_fetch{args=GitFetchArgs { branch: [Exact("main")], remotes: ["test/jj"], all_remotes: false }}:workspace_helper:workspace_helper_with_stats:maybe_snapshot_impl:snapshot_working_copy:base_ignores: globset: built glob set; 0 literals, 16 basenames, 0 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 2 regexes
2025-02-17T17:02:34.670841Z DEBUG run_command:cmd_git_fetch{args=GitFetchArgs { branch: [Exact("main")], remotes: ["test/jj"], all_remotes: false }}:fetch{remote_name="test/jj" branch_names=[Exact("main")] depth=None}: jj_lib::git_subprocess: spawning a git subprocess cmd=LC_ALL="C" "git" "--bare" "--git-dir" "/private/var/folders/yd/mh726b5d2vqfyp132jtzq9t80000gn/T/tmp.Dba7YWIbpU/.git" "fetch" "--prune" "--progress" "--" "test/jj" "+refs/heads/main:refs/remotes/test/jj/main"
remote: Enumerating objects: 59317, done.
remote: Total 59317 (delta 97), reused 67 (delta 59), pack-reused 59156 (from 2)
2025-02-17T17:02:40.527972Z DEBUG run_command:cmd_git_fetch{args=GitFetchArgs { branch: [Exact("main")], remotes: ["test/jj"], all_remotes: false }}:import_refs: jj_lib::git: import_refs
2025-02-17T17:02:40.575574Z DEBUG run_command:cmd_git_fetch{args=GitFetchArgs { branch: [Exact("main")], remotes: ["test/jj"], all_remotes: false }}:import_refs:import_head_commits: jj_lib::git_backend: import extra metadata entries heads_count=31
tag: v0.10.0 [new]
tag: v0.11.0 [new]
tag: v0.12.0 [new]
tag: v0.13.0 [new]
tag: v0.14.0 [new]
tag: v0.15.0 [new]
tag: v0.15.1 [new]
tag: v0.16.0 [new]
tag: v0.17.0 [new]
tag: v0.17.1 [new]
tag: v0.18.0 [new]
tag: v0.19.0 [new]
tag: v0.20.0 [new]
tag: v0.21.0 [new]
tag: v0.22.0 [new]
tag: v0.23.0 [new]
tag: v0.24.0 [new]
tag: v0.25.0 [new]
tag: v0.26.0 [new]
tag: v0.3.0  [new]
tag: v0.3.1  [new]
tag: v0.3.2  [new]
tag: v0.3.3  [new]
tag: v0.4.0  [new]
tag: v0.5.0  [new]
tag: v0.5.1  [new]
tag: v0.6.0  [new]
tag: v0.6.1  [new]
tag: v0.7.0  [new]
tag: v0.8.0  [new]
tag: v0.9.0  [new]
Warning: No branch matching `main` found on any specified/configured remote
```

## Expected Behavior

With Git:

```
yuyuko:/v/f/y/m/T/tmp.zjfjFoHqKn
❭ git init
Initialized empty Git repository in /private/var/folders/yd/mh726b5d2vqfyp132jtzq9t80000gn/T/tmp.zjfjFoHqKn/.git/

yuyuko:/v/f/y/m/T/tmp.zjfjFoHqKn
❭ git remote add test/jj https://github.com/jj-vcs/jj.git

yuyuko:/v/f/y/m/T/tmp.zjfjFoHqKn
❭ git fetch test/jj main
remote: Enumerating objects: 59316, done.
remote: Counting objects: 100% (157/157), done.
remote: Compressing objects: 100% (100/100), done.
remote: Total 59316 (delta 95), reused 64 (delta 57), pack-reused 59159 (from 2
Receiving objects: 100% (59316/59316), 24.61 MiB | 4.28 MiB/s, done.
Resolving deltas: 100% (48773/48773), done.
From https://github.com/jj-vcs/jj
 * branch              main       -> FETCH_HEAD
 * [new branch]        main       -> test/jj/main

yuyuko:/v/f/y/m/T/tmp.zjfjFoHqKn
❭ git show-ref refs/remotes/test/jj/main
4ed0e1330a7ac8cf6ca3cc70db70ffb78ddb76c4 refs/remotes/test/jj/main
```

## Actual Behavior

As above.

## Specifications

- Platform: macOS 15.3.1
- Version: 539cd75f with https://github.com/jj-vcs/jj/pull/5553 and a couple other changes merged in. (The `.git/config` matches what `git(1)` produces, so I don’t think my PR is relevant here.)

bsdinis:
`jj` assumes that remotes don't have slashes in them Hence, a reference like `refs/remotes/test/jj/some/branch` will always be parsed into `jj/some/branch@test`, even if the user explicitely created the remote via `jj git remote add test/jj <URL>`.

This does not match the git behaviour, where such remotes accepted, and as such creates confusion with the `jj git` commands.

This commit fixes this problem, by adding the current known remotes to the ref parsing function. Note that if the remote reference doesn't match any of the known remotes, then the remote is parsed assuming it has no slashes.

Fixes #5731

# Checklist

If applicable:

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -16,6 +16,7 @@
 
 use std::borrow::Cow;
 use std::collections::BTreeMap;
+use std::collections::BTreeSet;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::default::Default;
@@ -165,19 +166,46 @@ impl<'a> RefToPush<'a> {
     }
 }
 
-pub fn parse_git_ref(ref_name: &str) -> Option<RefName> {
+/// Parse a git reference
+///
+/// There are three types of git references:
+///  - `refs/heads/<branch-name>`
+///  - `refs/remotes/<remote-name>/<branch-name>`
+///  - `refs/tags/<tag>`
+///
+/// Here, both `<remote-name>`, `<branch-name>` and `<tag>` are path like
+/// strings. In the case of remote references, this means that it's impossible
+/// to distinguish where the remote name ends and the branch name begins
+pub fn parse_git_ref(ref_name: &str, remote_names: &BTreeSet<impl AsRef<BStr>>) -> Option<RefName> {
     if let Some(branch_name) = ref_name.strip_prefix("refs/heads/") {
         // Git CLI says 'HEAD' is not a valid branch name
         (branch_name != "HEAD").then(|| RefName::LocalBranch(branch_name.to_string()))
     } else if let Some(remote_and_branch) = ref_name.strip_prefix("refs/remotes/") {
-        remote_and_branch
-            .split_once('/')
-            // "refs/remotes/origin/HEAD" isn't a real remote-tracking branch
-            .filter(|&(_, branch)| branch != "HEAD")
-            .map(|(remote, branch)| RefName::RemoteBranch {
-                remote: remote.to_string(),
-                branch: branch.to_string(),
-            })
+        let mut it = remote_and_branch.split("/");
+
+        let mut remote = String::new();
+        for component in it.by_ref() {
+            if !remote.is_empty() {
+                remote.push('/');
+            }
+            remote.push_str(component);
+
+            if remote_names.iter().any(|x| x.as_ref() == remote) {
+                break;
+            }
+        }
+
+        // if the remote is unknown, we assume it's a single
+        // component remote as there is no way to differentiate
+        let branch = if remote == remote_and_branch {
+            let (r, b) = remote_and_branch.split_once('/')?;
+            remote = r.to_string();
+            b.to_string()
+        } else {
+            it.join("/")
+        };
+
+        (branch != "HEAD").then_some(RefName::RemoteBranch { branch, remote })
     } else {
         ref_name
             .strip_prefix("refs/tags/")
@@ -499,12 +527,14 @@ fn diff_refs_to_import(
     git_repo: &gix::Repository,
     git_ref_filter: impl Fn(&RefName) -> bool,
 ) -> Result<RefsToImport, GitImportError> {
+    let git_remotes = git_repo.remote_names();
     let mut known_git_refs: HashMap<&str, &RefTarget> = view
         .git_refs()
         .iter()
         .filter_map(|(full_name, target)| {
             // TODO: or clean up invalid ref in case it was stored due to historical bug?
-            let ref_name = parse_git_ref(full_name).expect("stored git ref should be parsable");
+            let ref_name =
+                parse_git_ref(full_name, &git_remotes).expect("stored git ref should be parsable");
             git_ref_filter(&ref_name).then_some((full_name.as_ref(), target))
         })
         .collect();
@@ -552,7 +582,7 @@ fn diff_refs_to_import(
             // Skip non-utf8 refs.
             continue;
         };
-        let Some(ref_name) = parse_git_ref(full_name) else {
+        let Some(ref_name) = parse_git_ref(full_name, &git_remotes) else {
             // Skip other refs (such as notes) and symbolic refs.
             continue;
         };
@@ -770,6 +800,7 @@ pub fn export_some_refs(
     git_ref_filter: impl Fn(&RefName) -> bool,
 ) -> Result<Vec<FailedRefExport>, GitExportError> {
     let git_repo = get_git_repo(mut_repo.store())?;
+    let git_remotes = git_repo.remote_names();
 
     let RefsToExport {
         branches_to_update,
@@ -777,6 +808,7 @@ pub fn export_some_refs(
         mut failed_branches,
     } = diff_refs_to_export(
         mut_repo.view(),
+        &git_repo,
         mut_repo.store().root_commit_id(),
         &git_ref_filter,
     );
@@ -787,7 +819,7 @@ pub fn export_some_refs(
             .target()
             .try_name()
             .and_then(|name| str::from_utf8(name.as_bstr()).ok())
-            .and_then(parse_git_ref)
+            .and_then(|s| parse_git_ref(s, &git_remotes))
         {
             let old_target = head_ref.inner.target.clone();
             let current_oid = match head_ref.into_fully_peeled_id() {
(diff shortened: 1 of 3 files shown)


lib/src/git.rs
```diff
@@ -165,19 +166,46 @@ impl<'a> RefToPush<'a> {
...
+        let mut remote = String::new();
+        for component in it.by_ref() {
+            if !remote.is_empty() {
+                remote.push('/');
+            }
+            remote.push_str(component);
+
+            if remote_names.iter().any(|x| x.as_ref() == remote) {
+                break;
+            }
```
emilazy, line 195: This somewhat arbitrarily handles the case of `refs/remotes/a/b/c` where there are remotes names both `a` and `a/b`. I think if we want to support these kinds of shenanigans in general we should at least test for this, and probably have a defined way of resolving the mess (I guess we would prefer `b/c@a` with this code due to the use of an ordered `BTreeSet`?)

lib/src/git.rs
```diff
@@ -1588,27 +1623,28 @@ impl<'a> GitFetch<'a> {
...
+            match ref_name {
+                RefName::LocalBranch(_) => false,
+                RefName::Tag(_) => true,
+                RefName::RemoteBranch { branch, remote } => self.fetched.iter().any(|fetched| {
+                    /*
+                    eprintln!(
+                        "checking if {:?}@{:?} matches {branch}@{remote}",
+                        fetched.branches, fetched.remote
+                    );
+                    */
```
emilazy, line 1636: Left over debug code.

lib/src/git.rs
emilazy: (The fact that there is probably no foolproof way of handling this case – at least none that isn’t a huge pain – is why I think it might be okay to just error out on remotes with `/` in the name.)

lib/src/git.rs
bsdinis: done

lib/src/git.rs
bsdinis: > This somewhat arbitrarily handles the case of refs/remotes/a/b/c where there are remotes names both a and a/b. I think if we want to support these kinds of shenanigans in general we should at least test for this, and probably have a defined way of resolving the mess (I guess we would prefer b/c@a with this code due to the use of an ordered BTreeSet?)

We would prefer the shortest remote at all times, due to the checking after each component being added.

As context, `git` also does not differentiate (also because they never have to distinguish implicitly -- either a remote is the default or it's being explicitly given)

bsdinis: > (The fact that there is probably no foolproof way of handling this case – at least none that isn’t a huge pain – is why I think it might be okay to just error out on remotes with `/` in the name.)

Note, I don't even know if we can do it, at least not here. We can try to curb it, by parsing it out on `jj git remote *` commands. But on a colocated repo `git` will allow it

emilazy: Yeah but Git also allows invalid UTF‐8 and stuff and we just refuse to deal with that ourselves. So complaining to the user and giving up is always an option.

OTOH non‐UTF‐8 is always evil but I ran into this bug when trying to structure my remotes in “good faith”.

yuja: I agree that it's okay (and is probably better) to reject remote name containing `/`. Things wouldn't work if there were remotes named `foo` and `foo/bar`, and bookmarks `bar/baz`, ...

> But on a colocated repo git will allow it

We can still error out on fetch/push.

bsdinis: Yeah, but at this point (looking at remotes that already exist), we cannot distinguish something from being a remote with a `/` and a branch with a `/`

joyously: _and the docs have a suggestion for `push -c` using a name with `/`_

bsdinis: > _and the docs have a suggestion for `push -c` using a name with `/`_

I think here we are discussing _remotes_ with `/` not bookmarks with `/`. In my experience, bookmarks with `/` are much much more common.

yuja: > Yeah, but at this point (looking at remotes that already exist), we cannot distinguish something from being a remote with a `/` and a branch with a `/`

I think it's okay to assume there are no remotes named `foo/bar` (given such remotes are rejected by `jj git fetch`/`clone`/`remote`/..)

Maybe we can also make `import_refs()` be stricter by enumerating configured remotes, but that can be implemented separately.

bsdinis: > I think it's okay to assume there are no remotes named foo/bar (given such remotes are rejected by jj git fetch/clone/remote/..)

Just to be clear, this does not happen right now. I have another PR which we can use to decide

bsdinis:
Implemented an alternative in #5737

bsdinis:
Closing in favour of #5737

PR #5745, opened 2025-02-18
cli: remove error hint for invalid remote names

bsdinis:
Existing hint suggests using `jj git remote rename`, but that's not applicable to `jj git clone`.

This commit removes that hint

# Checklist

If applicable:

### DIFF
--- a/cli/src/command_error.rs
+++ b/cli/src/command_error.rs
@@ -585,16 +585,7 @@ jj currently does not support partial clones. To use jj with this repository, tr
 
     impl From<GitRemoteManagementError> for CommandError {
         fn from(err: GitRemoteManagementError) -> Self {
-            match err {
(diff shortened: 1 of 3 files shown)


yuja (commented):
I would simply remove the hint because:
1. `rename` doesn't apply to most `jj git remote` commands (it might be valid on `set-url`, but that would rarely happen.)
2. I think it's obvious that the user needs to give a different name.

bsdinis:
> Need to update the commit message too

nice catch, thanks

cli/src/command_error.rs
yuja, line 1: nit: in commit message
> cli: remove error message for invalid remote names

error message -> error hint

btw, the hint doesn't apply to `jj git remote add`, etc.

yuja (approved):

PR #5749, opened 2025-02-19
tests: add helper to snapshot command output and exit status all at once

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/common/test_environment.rs
+++ b/cli/tests/common/test_environment.rs
@@ -14,10 +14,12 @@
 
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::ffi::OsStr;
 use std::fmt;
 use std::fmt::Display;
 use std::path::Path;
 use std::path::PathBuf;
+use std::process::ExitStatus;
 
 use indoc::formatdoc;
 use regex::Captures;
@@ -94,6 +96,32 @@ impl Default for TestEnvironment {
 }
 
 impl TestEnvironment {
+    /// Runs `jj args..` in the `current_dir`, returns the output.
+    #[must_use = "either snapshot the output or assert the exit status with .success()"]
+    pub fn run_jj_in<I>(&self, current_dir: &Path, args: I) -> CommandOutput
+    where
+        I: IntoIterator,
+        I::Item: AsRef<OsStr>,
+    {
+        self.run_jj_with(|cmd| cmd.current_dir(current_dir).args(args))
+    }
+
+    /// Runs `jj` command with additional configuration, returns the output.
+    #[must_use = "either snapshot the output or assert the exit status with .success()"]
+    pub fn run_jj_with(
+        &self,
+        configure: impl FnOnce(&mut assert_cmd::Command) -> &mut assert_cmd::Command,
+    ) -> CommandOutput {
+        let mut cmd = self.jj_cmd(&self.env_root, &[]);
+        let output = configure(&mut cmd).output().unwrap();
+        CommandOutput {
+            stdout: self.normalize_output(String::from_utf8(output.stdout).unwrap()),
+            stderr: self.normalize_output(String::from_utf8(output.stderr).unwrap()),
+            status: output.status,
+        }
+    }
+
+    // TODO: rename to new_jj_cmd, remove arguments that can be set later
     #[must_use]
     pub fn jj_cmd(&self, current_dir: &Path, args: &[&str]) -> assert_cmd::Command {
         let mut cmd = assert_cmd::Command::cargo_bin("jj").unwrap();
@@ -169,6 +197,7 @@ impl TestEnvironment {
 
     /// Run a `jj` command, check that it was successful, and return its
     /// `(stdout, stderr)`.
+    // TODO: remove jj_cmd_*() in favor of run_jj_*()
     pub fn jj_cmd_ok(
         &self,
         current_dir: &Path,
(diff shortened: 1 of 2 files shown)


cli/tests/common/test_environment.rs
```diff
@@ -94,6 +96,31 @@ impl Default for TestEnvironment {
 }
 
 impl TestEnvironment {
+    /// Runs `jj args..` at the `current_dir`, returns the output.
+    #[must_use = "either snapshot the output or assert the exit status with .success()"]
+    pub fn jj_at<I>(&self, current_dir: &Path, args: I) -> CommandOutput
```
martinvonz, line 101: It seems like this is going to become the main entry point, right? `jj_at()` is not very intuitive to me. I guess it's not `jj_cmd()` just because that's already taken. How about `run_jj()` (and `run_jj_with()` for the other one)?

cli/tests/common/test_environment.rs
```diff
@@ -338,6 +365,71 @@ impl TestEnvironment {
...
+        }
+        if !stderr.is_empty() {
+            writeln!(f, "------- stderr -------")?;
+            write!(f, "{stderr}")?;
+        }
+        if !status.success() {
+            if let Some(code) = status.code() {
+                writeln!(f, "[exit status: {code}]")?;
+            } else {
+                // On Windows, this would print "exit code: N".
```
martinvonz, line 425: nit: do you mean "code" -> "status"?

cli/tests/test_global_opts.rs
```diff
@@ -68,55 +63,74 @@ fn test_version() {
...
+    test_env
+        .jj_at(&repo_path, ["bookmark", "create", "-r@", "help"])
+        .success();
+    test_env
+        .jj_at(&repo_path, ["bookmark", "create", "-r@", "log"])
+        .success();
+    test_env
+        .jj_at(&repo_path, ["bookmark", "create", "-r@", "show"])
+        .success();
+    // TODO: test_env.jj(&repo_path, ["-r", "help"]).success()
```
martinvonz, line 124: nit: fix name of method (currently `jj`)

martinvonz (commented):
I like it

cli/tests/common/test_environment.rs
yuja: Yes, and we might add something like `.at(dir) -> TestRepo<'_>; test_repo.jj(args)` later. `.jj_cmd()` won't be removed since there will be a few callers who don't wanna run command, but just get the command builder object.

I'm okay with renaming to `run_jj_at()`, `run_jj_with()`, and `new_jj_cmd()`.

cli/tests/common/test_environment.rs
```diff
@@ -338,6 +365,71 @@ impl TestEnvironment {
...
+
+impl Display for CommandOutput {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let CommandOutput {
+            stdout,
+            stderr,
+            status,
+        } = self;
+        if !stdout.is_empty() {
+            writeln!(f, "------- stdout -------")?;
```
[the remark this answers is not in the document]
yuja: That's doable. Maybe it's also good to remove the `stdout` header at all.

cli/tests/common/test_environment.rs
yuja: No. The output is platform-dependent, so we normalize it by not using the default formatting.

cli/tests/common/test_environment.rs
martinvonz: Sounds good. But `run_jj_in()` seems more more intuitive than `run_jj_at()`. At least I think of it as running in a directory more than running at a path.

cli/tests/common/test_environment.rs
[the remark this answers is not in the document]
martinvonz: Good point, never printing it sounds good to me.

cli/tests/common/test_environment.rs
martinvonz: Sorry, I still don't follow. `{status}` is the default formatting, right? And what does "this" in "this would print" refer to?

cli/tests/common/test_environment.rs
yuja: Perhaps, I inherited it from `*at()` POSIX functions. I'm okay with either way.

cli/tests/common/test_environment.rs
yuja: I mean `{status}` would print `exit code: N` on Windows, so don't remove the  `if` block above.

martinvonz: Ah, I see. Mayne something like this then?

```rust
// If there is an exit code, `{status}` would get rendered as "exit code: N"
// on Windows, so we render it ourselves for compatibility.
if let Some(code) = status.code() {
    writeln!(f, "[exit status: {code}]")?;
} else {
    writeln!(f, "[{status}]")?;
}
```

cli/tests/common/test_environment.rs
martinvonz: I prefer `run_jj_in()`. The `_at()` made me think it was about running at a certain operation id. I'd get used to it quickly, of course, but it seems less intuitive.

martinvonz (approved):

cli/tests/common/test_environment.rs
yuja: Maybe `_in` has implication that the environment is isolated within the directory? That wouldn't matter in this context, so `_in` seems also good.

cli/tests/common/test_environment.rs
ilyagr: If <https://github.com/jj-vcs/jj/pull/5749#discussion_r1962702592> is correct, does the first branch of the `if` need to print `[exit code...` instead of `[exit status...`?

ilyagr:
Did you see my comments? They could be resolved in another PR if you prefer, but if so let it be a conscious decision.

cli/tests/common/test_environment.rs
yuja: The plan is to add helper type that holds `repo_root`, and we'll set the root once and do `.run_jj(args)`. That's why I added `_in` suffix to the function that takes `cwd`.

cli/tests/common/test_environment.rs
yuja: No. I tend to use the Unix convention over Windows one.

ilyagr: I'm not sure I follow, but the tests seem to pass on both Unix on Windows, so I don't need to delay you further.

ilyagr:
OK, thanks for the replies!

cli/tests/common/test_environment.rs
yuja: If we simply do `writeln!(f, "[{status}]")`, the test would fail. (I haven't tried, but I know that `status` is rendered differently on Windows.)

martinvonz: @ilyagr, if there is an exit code (`status.code()` is `Some`), then `{status}` gets rendered as "exit status: N" on Linux and "exit code: N" on Windows.

ilyagr: I didn't notice the `normalize_exit_status` function. I think the comment here is still confusing, but it's not a big deal; I might fix it up if I remember.

I actually like Yuya's original version of having the comment in the `if` branch, but I'd say something like

> This branch gets triggered on Windows, where it prints "status code: N". The difference with Unix gets normalized away in `normalize_exit_status`.

martinvonz: I don't think that's correct. The branch does not depend on OS, IIUC. The lack of exit code instead signals (pun not intended) that the process was killed by a signal.

yuja: Yeah, the `else` path is for unusual cases in which it's okay to emit platform-dependent output.

PR #5750, opened 2025-02-19
docs: explain that `jj root` is synonymous with `jj workspace root`

joliss:


cli/src/commands/root.rs
```diff
@@ -22,6 +22,8 @@ use crate::command_error::CommandError;
 use crate::ui::Ui;
 
 /// Show the current workspace root directory
+///
+/// Synonymous with `jj workspace root`.
```
martinvonz, line 26: nit: `jj undo` says "Undo an operation (shortcut for jj op undo)", so it would be nice to be consistent with that. Putting it on a separate line as you have done works better if the first line is long. So I would suggest updating `jj undo` to match.

cli/src/commands/root.rs
yuja, line 1: iirc, the idea was that `jj root` will have flags to show other paths. I don't know if that will be implemented anytime soon, though.

If we document that `jj root` is identical to `jj workspace root`, it's probably better to unify the implementations so they'll never diverge.

cli/src/commands/root.rs
joliss: Done!

cli/src/commands/root.rs
joliss: FWIW, as a downstream consumer of the CLI, I wouldn't be taken by surprise if flags specific to `jj root` get added later, and it stops being a perfect shortcut -- just as long as shared options keep working.

cli/src/commands/root.rs
joliss: Oh, I misread, and I updated `jj root` to match `jj undo`. But I think that's fine as well? I like the `undo` wording. Or feel free to change it to whatever wording you prefer.

martinvonz: Yeah, that's also fine, as long as it fit on one line (but I don't know exactly how we define that).

joliss: Great, I force pushed, so I think this can probably be merged then.

martinvonz: There are failing tests. Could you update those too? You can use `cargo insta test --review --test-runner nextest`  or just `cargo insta test --review`.

joliss: @martinvonz Sorry about that -- I didn't even think to run the tests or check the CI failures. 😅 I've pushed again so this should be good to merge now.

martinvonz (approved):

PR #5755, opened 2025-02-20
fix: invoke tools from the workspace root

bryceberger:
Motivating discord discussion: https://discord.com/channels/968932220549103686/1339369275915501599

Previously, tools invoked by `jj fix` did not have their `.current_dir()` set, and would just run from whatever directory the user was in.

Now, the tools will always be invoked from the same directory that `jj root` gives.

As a motivating example, consider a configuration path that's always at the workspace root:

```toml
[fix.tools.something]
# previous:
command = ["cmd", "--config", "$root/tool.toml"]
#                              ^^^^^^ not possible
# now:
command = ["cmd", "--config", "./tool.toml"]
#                              ^^ now, just use a relative path
```

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `split.legacy-bookmark-behavior = true`, but this will likely be removed in a
   future release. [#3419](https://github.com/jj-vcs/jj/issues/3419)
 
+* `jj fix` now always sets the working directory of invoked tools to be the
+  workspace root, instead of the working directory of the `jj fix`.
+
 ### Deprecations
 
 * This release takes the first steps to make target revision required in
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -15,6 +15,7 @@
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::io::Write;
+use std::path::Path;
 use std::process::Stdio;
 use std::sync::mpsc::channel;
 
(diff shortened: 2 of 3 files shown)


bryceberger:
The current implementation is a breaking change, and does not have a way to restore the old behavior. Is there a desire for the old behavior? I don't think there is, given most fix tools either a) don't care about the working directory, or b) only care in order to set a config file or similar.

yuja:
Just fyi, https://github.com/jj-vcs/jj/pull/4867

I thought this would have been merged, but apparently not.

bryceberger:
Ah, thanks for pointing me to that. It's somewhat difficult to search for issues / prs about "fix" :laughing:

martinvonz:
Thanks for finding that (I also unsuccessfully tried searching for `fix` and `"fix:"`). I was sure it had been merged already (and I was also - incorrectly - sure that I had authored the PR :) ).

I think making the path always be the workspace root is reasonable, but I think it's also a bit weird because the formatting tools shouldn't really depend on the current working copy. No action requested.

bryceberger:
> it always goes off of the cwd

According to https://biomejs.dev/reference/cli/, you should be able to `biome --config-path ./bar/config.toml`. That's what I was expecting --- locate a known path (workspace root) with jj, then rely on the tool if it cares about a file other than the root.

bryceberger:
Since this branch merges cleanly, do we want to just merge this one?

martinvonz (approved):

martinvonz:
> Since this branch merges cleanly, do we want to just merge this one?

Yes, I think so. I hope you don't mind, @matts1. You can close #4867 now.

PR #5756, opened 2025-02-20
tests: move gitoxide test helpers to testutils

bsdinis:
These helpers are going to be needed to port the git2 code in the lib tests to gitoxide. Since the cli tests already depend on testutils, this helps with avoiding duplicating the code

# Checklist

If applicable:


yuja (approved):
Sounds good.

> These helpers are going to be needed to port the git2 code in the lib
> tests to gitoxide.

Be careful to not add circular deps. It's good to use `testutils` in `lib/tests/*.rs`, but things would get messy if we add `testutils` dependency in `lib/src/git_backend.rs` tests. The latter compiles, but would create `jj-lib (cfg tests) -> testutils -> jj-lib` dependency.

bsdinis:
> Be careful to not add circular deps. It's good to use `testutils` in `lib/tests/*.rs`, but things would get messy if we add `testutils` dependency in `lib/src/git_backend.rs` tests. The latter compiles, but would create `jj-lib (cfg tests) -> testutils -> jj-lib` dependency.

Looking at `lib/src/git_backend.rs` there is already a testutils dependency there, e.g.: https://github.com/jj-vcs/jj/blob/c8ee6b70885dfa48db215fcd9518918f1381968b/lib/src/git_backend.rs#L1547C1-L1547C50

yuja:
> Looking at `lib/src/git_backend.rs` there is already a testutils dependency there, e.g.: https://github.com/jj-vcs/jj/blob/c8ee6b70885dfa48db215fcd9518918f1381968b/lib/src/git_backend.rs#L1547C1-L1547C50

Correct, and we have several PRs to either remove it or move everything to jj-lib. It's better to not add more mess.

FWIW, one of the problems that would be caused by circular dependency is that there would be two different types of the same name. (See `user_settings()` in `git_backend.rs` tests.)

bsdinis:
> > Looking at `lib/src/git_backend.rs` there is already a testutils dependency there, e.g.: https://github.com/jj-vcs/jj/blob/c8ee6b70885dfa48db215fcd9518918f1381968b/lib/src/git_backend.rs#L1547C1-L1547C50
> 
> Correct, and we have several PRs to either remove it or move everything to jj-lib. It's better to not add more mess.
> 
> FWIW, one of the problems that would be caused by circular dependency is that there would be two different types of the same name. (See `user_settings()` in `git_backend.rs` tests.)

Gotcha, thanks for letting me know. I guess in that case we'll have to go raw

PR #5757, opened 2025-02-20
lib/tests: port lib tests to gitoxide

bsdinis:
As per #5548, this moves the majority of the tests in lib/test over to gitoxide

# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_git.rs
+++ b/lib/tests/test_git.rs
@@ -25,7 +25,6 @@ use std::sync::Barrier;
 use std::thread;
 
 use assert_matches::assert_matches;
-use git2::Oid;
 use itertools::Itertools;
 use jj_lib::backend::BackendError;
 use jj_lib::backend::ChangeId;
@@ -84,33 +83,27 @@ struct GitFetchStats {
     pub import_stats: git::GitImportStats,
 }
 
-fn empty_git_commit<'r>(
-    git_repo: &'r git2::Repository,
+fn empty_git_commit(
+    git_repo: &gix::Repository,
     ref_name: &str,
-    parents: &[&git2::Commit],
-) -> git2::Commit<'r> {
-    let signature = git2::Signature::now("Someone", "someone@example.com").unwrap();
-    let empty_tree_id = Oid::from_str("4b825dc642cb6eb9a060e54bf8d69288fbee4904").unwrap();
-    let empty_tree = git_repo.find_tree(empty_tree_id).unwrap();
-    let oid = git_repo
-        .commit(
-            Some(ref_name),
-            &signature,
-            &signature,
-            &format!("random commit {}", rand::random::<u32>()),
-            &empty_tree,
-            parents,
-        )
-        .unwrap();
-    git_repo.find_commit(oid).unwrap()
+    parents: &[gix::ObjectId],
+) -> gix::ObjectId {
+    let empty_tree_id = git_repo.empty_tree().id().detach();
+    testutils::git::write_commit(
+        git_repo,
+        ref_name,
+        empty_tree_id,
+        &format!("random commit {}", rand::random::<u32>()),
+        parents,
+    )
 }
 
-fn jj_id(commit: &git2::Commit) -> CommitId {
-    CommitId::from_bytes(commit.id().as_bytes())
+fn jj_id(id: gix::ObjectId) -> CommitId {
+    CommitId::from_bytes(id.as_bytes())
 }
 
-fn git_id(commit: &Commit) -> Oid {
-    Oid::from_bytes(commit.id().as_bytes()).unwrap()
+fn git_id(commit: &Commit) -> gix::ObjectId {
+    gix::ObjectId::from_bytes_or_panic(commit.id().as_bytes())
 }
 
 fn get_git_backend(repo: &Arc<ReadonlyRepo>) -> &GitBackend {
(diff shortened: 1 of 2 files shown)


lib/tests/test_git.rs
```diff
@@ -71,6 +70,7 @@ use tempfile::TempDir;
 use test_case::test_case;
 use testutils::commit_transactions;
 use testutils::create_random_commit;
+use testutils::git as tgit;
```
yuja, line 73: nit: I would use `testutils::git::` instead. `tgit` looks like a typo.

lib/tests/test_git.rs
```diff
@@ -84,33 +84,30 @@ struct GitFetchStats {
...
-            &empty_tree,
-            parents,
-        )
+    parents: &[gix::ObjectId],
+) -> gix::ObjectId {
+    let mut empty_tree_editor = git_repo
+        .edit_tree(gix::ObjectId::empty_tree(git_repo.object_hash()))
         .unwrap();
-    git_repo.find_commit(oid).unwrap()
+    let empty_tree_id = empty_tree_editor.write().unwrap().detach();
```
yuja, line 95: I think the empty tree id can be used without writing the tree object. iirc, gix has a fast path for empty-tree lookup.

lib/tests/test_git.rs
```diff
@@ -1566,7 +1608,20 @@ fn test_export_refs_bookmark_changed() {
...
+            .peel_to_commit_in_place()
+            .unwrap()
+            .id(),
+        git_repo
+            .find_reference("refs/heads/feature")
+            .unwrap()
+            .peel_to_commit()
+            .unwrap()
+            .id()
+    );
```
yuja, line 1624: Here we need to assert that the `HEAD` is unchanged (i.e. a symbolic ref.)

btw, maybe we can use `git_repo.head_id()`, `.head_name()`, etc.

lib/tests/test_git.rs
```diff
@@ -1614,20 +1673,28 @@ fn test_export_refs_unborn_git_bookmark(move_placeholder_ref: bool) {
...
-    git_repo.set_head("refs/heads/main").unwrap();
+    tgit::set_symbolic_reference(&git_repo, "HEAD", "refs/heads/main");
     let mut tx = test_data.repo.start_transaction();
     let mut_repo = tx.repo_mut();
     git::import_head(mut_repo).unwrap();
     git::import_refs(mut_repo, &git_settings).unwrap();
     mut_repo.rebase_descendants().unwrap();
     assert!(git::export_refs(mut_repo).unwrap().is_empty());
-    assert!(git_repo.head().is_err(), "HEAD is unborn");
+    assert!(matches!(
```
yuja, line 1683: nit: use `assert_matches!(..)`?

lib/tests/test_revset.rs
```diff
@@ -236,61 +236,48 @@ fn test_resolve_symbol_change_id(readonly: bool) {
...
-        git_commit_ids.push(git_commit_id);
+
+    for i in &[133, 320, 1020, 1885] {
+        let commit_id = testutils::git::write_commit(
+            &git_repo,
+            &format!("refs/heads/bookmark{i}"),
+            tree_id,
+            &format!("test {i}"),
+            &[],
+        );
```
yuja, line 254: nit: maybe we can instead use the same `Signature` values so the test outputs will be identical?

yuja (commented):
Thanks.

lib/tests/test_git.rs
bsdinis: couldn't, because git_repo.head() does not impl Debug

lib/tests/test_git.rs
bsdinis: done

lib/tests/test_git.rs
bsdinis: done

lib/tests/test_git.rs
bsdinis: found a better day to do it: with `is_unborn()`

lib/tests/test_revset.rs
bsdinis: I guess we could try, but what would be the upside? It would complicate the helper code for very little gain, maybe?

yuja: https://github.com/jj-vcs/jj/pull/5772

> what would be the upside?

Just because it would be easier to review ;)

bsdinis: very nice! do you think we should maybe merge that PR into this branch instead?

bsdinis: Or I can just implement it here

bsdinis: removed this file from the diff, resolving

lib/tests/test_git.rs
```diff
@@ -1601,10 +1632,14 @@ fn test_export_refs_current_bookmark_changed() {
...
-            .id(),
+            .id()
+            .detach(),
         git_id(&new_commit)
     );
-    assert!(git_repo.head_detached().unwrap());
+    assert!(matches!(
+        git_repo.head().unwrap().kind,
+        gix::head::Kind::Detached { .. }
+    ));
```
yuja, line 1642: nit: `.is_detached()`?

lib/tests/test_git.rs
```diff
@@ -2175,23 +2233,31 @@ fn test_reset_head_to_root() {
...
         RefTarget::normal(commit1.id().clone())
     );
 
     // Set Git HEAD back to root
     git::reset_head(tx.repo_mut(), &commit1).unwrap();
-    assert!(git_repo.head().is_err());
+    assert!(matches!(
+        git_repo.head().unwrap().kind,
+        gix::head::Kind::Unborn(_)
+    ));
```
yuja, line 2247: nit: `.is_unborn()`

lib/tests/test_git.rs
```diff
@@ -2200,7 +2266,10 @@ fn test_reset_head_to_root() {
...
```
yuja, line 2272: nit: `.is_unborn()`

btw, I think `.head_id().is_ok()` in this test can be replaced with `!is_unborn()` or `is_detached()`.

yuja (approved):

PR #5761, opened 2025-02-20
tests: port some non-trivial `jj_cmd_*()` callers

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_diffedit_command.rs
+++ b/cli/tests/test_diffedit_command.rs
@@ -111,11 +111,13 @@ fn test_diffedit() {
 
     // Nothing happens if the diff-editor exits with an error
     std::fs::write(&edit_script, "rm file2\0fail").unwrap();
-    let stderr = test_env.jj_cmd_failure(&repo_path, &["diffedit"]);
-    insta::assert_snapshot!(stderr.normalize_exit_status(), @r"
+    let output = test_env.run_jj_in(&repo_path, ["diffedit"]);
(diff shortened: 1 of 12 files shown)


cli/tests/test_diffedit_command.rs
```diff
@@ -111,11 +111,13 @@ fn test_diffedit() {
...
     std::fs::write(&edit_script, "rm file2\0fail").unwrap();
-    let stderr = test_env.jj_cmd_failure(&repo_path, &["diffedit"]);
-    insta::assert_snapshot!(stderr.normalize_exit_status(), @r"
+    let output = test_env.run_jj_in(&repo_path, ["diffedit"]);
+    insta::assert_snapshot!(output.normalize_stderr_exit_status(), @r"
+    ------- stderr -------
     Error: Failed to edit diff
     Caused by: Tool exited with exit status: 1 (run with --debug to see the exact invocation)
     [EOF]
+    [exit status: 1]
```
martinvonz, line 120: Btw, and off topic for this PR, we could always decide to print the exit status line even on success. Then we wouldn't need the `[EOF]` line anymore.

martinvonz (approved):

PhilipMetzger (approved):

cli/tests/test_diffedit_command.rs
yuja: I thought that would be noisy. `[EOF]` could be removed, but that means `[exit status` or `------- stderr -------` line might be concatenated with the previous line.

PR #5770, opened 2025-02-20
lib: move git_backend tests to gitoxide

bsdinis:
This forces some code duplication to avoid depending on the testutils.

# Checklist

If applicable:

### DIFF
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -1531,7 +1531,6 @@ fn bytes_vec_from_json(value: &serde_json::Value) -> Vec<u8> {
 #[cfg(test)]
 mod tests {
     use assert_matches::assert_matches;
-    use git2::Oid;
     use hex::ToHex;
     use pollster::FutureExt;
     use test_case::test_case;
@@ -1541,6 +1540,32 @@ mod tests {
     use crate::content_hash::blake2b_hash;
     use crate::tests::new_temp_dir;
 
+    const GIT_USER: &str = "Someone";
+    const GIT_EMAIL: &str = "someone@example.com";
+
+    fn git_config() -> Vec<bstr::BString> {
+        vec![
+            format!("user.name = {GIT_USER}").into(),
+            format!("user.email = {GIT_EMAIL}").into(),
+            "init.defaultBranch = master".into(),
+        ]
+    }
+
+    fn open_options() -> gix::open::Options {
+        gix::open::Options::isolated().config_overrides(git_config())
+    }
+
+    fn git_init(directory: impl AsRef<Path>) -> gix::Repository {
+        gix::ThreadSafeRepository::init_opts(
+            directory,
+            gix::create::Kind::WithWorktree,
+            gix::create::Options::default(),
+            open_options(),
+        )
+        .unwrap()
+        .to_thread_local()
+    }
+
     #[test_case(false; "legacy tree format")]
     #[test_case(true; "tree-level conflict format")]
     fn read_plain_git_commit(uses_tree_conflict_format: bool) {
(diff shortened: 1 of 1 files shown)


bsdinis:
There is a test missing because it relies on forging signatures, which I'm not sure how to do

lib/src/git_backend.rs
```diff
@@ -1540,67 +1539,100 @@ mod tests {
...
-        let git_committer = git2::Signature::new(
-            "git committer",
-            "git.committer@example.com",
-            &git2::Time::new(2000, -480),
-        )
-        .unwrap();
-        let git_tree = git_repo.find_tree(root_tree_id).unwrap();
+        let root_tree_id = root_tree_builder.write().unwrap().detach();
+        let git_author = gix::actor::Signature {
+            name: bstr::BString::from("git author"),
```
yuja, line 1594: nit: we usually import `BString` by name. `"..".into()` also works.

lib/src/git_backend.rs
```diff
@@ -1540,67 +1539,100 @@ mod tests {
...
-                &git_author,
+            .commit_as(
                 &git_committer,
+                &git_author,
+                "refs/heads/dummy",
                 "git commit message",
-                &git_tree,
-                &[],
+                root_tree_id,
+                None::<gix::ObjectId>,
```
yuja, line 1610: nit: `[] as [gix::ObjectId; 0]` or something? `None` is confusing here because `parents` is a list.

lib/src/git_backend.rs
```diff
@@ -1540,67 +1539,100 @@ mod tests {
...
+            time: gix::date::Time::new(2000, -480 * 60),
+        };
         let git_commit_id = git_repo
-            .commit(
-                None,
-                &git_author,
+            .commit_as(
                 &git_committer,
+                &git_author,
+                "refs/heads/dummy",
```
yuja, line 1607: nit: maybe better to delete the dummy ref because we're testing core `GitBackend` functionality.

lib/src/git_backend.rs
```diff
@@ -1540,67 +1539,100 @@ mod tests {
...
-            "git committer",
-            "git.committer@example.com",
-            &git2::Time::new(2000, -480),
-        )
-        .unwrap();
-        let git_tree = git_repo.find_tree(root_tree_id).unwrap();
+        let root_tree_id = root_tree_builder.write().unwrap().detach();
+        let git_author = gix::actor::Signature {
+            name: bstr::BString::from("git author"),
+            email: bstr::BString::from("git.author@example.org"),
```
yuja, line 1595: nit: unneeded change `.com` -> `.org` (there are more instances)

lib/src/git_backend.rs
```diff
@@ -2168,17 +2233,17 @@ mod tests {
         let obj = git_repo
             .find_object(gix::ObjectId::try_from(id.as_bytes()).unwrap())
             .unwrap();
-        insta::assert_snapshot!(std::str::from_utf8(&obj.data).unwrap(), @r###"
+        insta::assert_snapshot!(std::str::from_utf8(&obj.data).unwrap().lines().map(|x| x.trim()).join("\n"), @r"
```
yuja, line 2236: nit: unnecessary normalization? better to test the exact value.

bsdinis: so this was causing me all kinds of trouble, for weird reasons. mainly trailing whistespace on the empty lines (everytime i opened the file it would change the snapshot). Do you have any other options here?

yuja: Oh, so your editor strips trailing whitespaces in string literals? That's weird.

Maybe we can remove blank lines from fake signer output, but I don't remember why we added that. @necauqua do you have any idea?

bsdinis: Done

lib/src/git_backend.rs
```diff
@@ -1750,31 +1797,39 @@ mod tests {
...
-                &signature,
-                &signature,
-                "git commit message",
-                &empty_tree,
-                &[],
-            )
-            .unwrap();
+        let secure_sig: BString = "here are some ASCII bytes to be used as a test \
+                                   signature\n\ndefinitely not PGP\n"
+            .into();
```
yuja, line 1812: nit: type annotation can be removed by moving `.into()` to `.extra_headers.push(..)`

lib/src/git_backend.rs
```diff
@@ -1912,14 +1967,20 @@ mod tests {
...
         let git_repo_path = temp_dir.path().join("git");
-        let git_repo = git2::Repository::init(git_repo_path).unwrap();
+        let git_repo = git_init(&git_repo_path);
 
         let backend = GitBackend::init_external(&settings, store_path, git_repo.path()).unwrap();
         let create_tree = |i| {
-            let blob_id = git_repo.blob(b"content {i}").unwrap();
-            let mut tree_builder = git_repo.treebuilder(None).unwrap();
+            let blob_id = git_repo
+                .write_blob(format!("content {i}").as_bytes())
```
yuja, line 1975: nit: `.as_bytes()` can be omitted

lib/src/git_backend.rs
```diff
@@ -1949,47 +2010,55 @@ mod tests {
...
+                read_commit_id.as_bytes(),
+            ))
             .unwrap();
-        let git_tree = git_repo.find_tree(git_commit.tree_id()).unwrap();
+        let git_tree = git_repo.find_tree(git_commit.tree_id().unwrap()).unwrap();
         assert!(git_tree
             .iter()
-            .filter(|entry| entry.name() != Some("README"))
-            .all(|entry| entry.filemode() == 0o040000));
+            .filter(|entry| match entry {
```
yuja, line 2020: nit: it's probably better to insert `.map(Result::unwrap)` here instead of filtering out errors.

lib/src/git_backend.rs
```diff
@@ -1949,47 +2010,55 @@ mod tests {
...
+                Ok(e) => e.filename() != b"README",
+                Err(_) => false,
+            })
+            .all(|entry| match entry {
+                Ok(e) => e.mode().0 == 0o040000,
+                Err(_) => false,
+            }));
         let mut iter = git_tree.iter();
-        let entry = iter.next().unwrap();
```
yuja, line 1960: nit: maybe we can also map `iter` here as `git_tree.iter().map(Result::unwrap)`

lib/src/git_backend.rs
```diff
@@ -2032,65 +2103,70 @@ mod tests {
...
+            .references()
             .unwrap()
-            .try_collect()
-            .unwrap();
-        assert!(git_refs
-            .iter()
-            .any(|git_ref| git_ref.target().unwrap() == git_commit_id));
+            .prefixed("refs/jj/keep/")
+            .unwrap()
+            .any(|git_ref| git_ref.unwrap().into_fully_peeled_id().unwrap() == git_commit_id));
```
yuja, line 2169: nit: do you have any reason to use various peeling functions, `.into_fully_peeled_id()`, `.peel_to_commit()`, and `.peel_to_id_in_place()`?

It's better to stick to one, and I think `.id()` works because `jj/keep` refs shouldn't be a symbolic ref.

yuja (approved):
Thanks.

lib/src/git_backend.rs
bsdinis: done

lib/src/git_backend.rs
bsdinis: done

PR #5772, opened 2025-02-21
tests: rewrite change ID resolution test to not rely on git::import_refs()

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/tests/test_revset.rs
+++ b/lib/tests/test_revset.rs
@@ -18,6 +18,7 @@ use std::path::Path;
 use assert_matches::assert_matches;
 use chrono::DateTime;
 use itertools::Itertools;
+use jj_lib::backend::ChangeId;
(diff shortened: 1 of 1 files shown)


bsdinis (approved):
Very cool!

martinvonz (approved):
I have no idea why the test looked that way. Having distinct author and committer also seems unnecessary.

bsdinis:
> I have no idea why the test looked that way. Having distinct author and committer also seems unnecessary.

Agree, but I have exhausted jujutsu's Bitcoin mining budget trying brute-forcing the change ids to match yesterday, don't think we can spend more carbon credits for yuja to streamline that part /jk

yuja:
Yep, and we don't have to generate commit IDs sharing the same prefix with change IDs anymore. I just tried to minimize the output changes.

PR #5780, opened 2025-02-21
templater: add trim methods to `String` type

### LINKED ISSUE #5086: FR: Add `trim` template functions

Would it be possible to add the `trim` family of functions (`trim`, `trim_start`, `trim_end`) to allow stripping off unwanted characters at the beginning and/or end of a string?

I don't think there are any alternatives currently.

**Additional context**

I have noticed my log output sometimes truncates with uninteresting trailing characters (like whitespace), and it'd be neat to be able to trim those off before printing the log output.

gechr:
Resolves #5086

### DIFF
--- a/cli/src/template_builder.rs
+++ b/cli/src/template_builder.rs
@@ -845,6 +845,30 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
         },
     );
     map.insert(
+        "trim",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            function.expect_no_arguments()?;
+            let out_property = self_property.map(|s| s.trim().to_owned());
+            Ok(L::wrap_string(out_property))
+        },
+    );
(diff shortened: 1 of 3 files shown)


cli/src/template_builder.rs
```diff
@@ -2647,6 +2671,15 @@ mod tests {
             env.render_ok(r#""bar@other.example.com".remove_suffix("@other.example.com")"#),
             @"bar");
 
+        insta::assert_snapshot!(env.render_ok(r#"" \n \r    \t \r ".trim()"#), @"");
+        insta::assert_snapshot!(env.render_ok(r#"" \n \r testing \t \r ".trim()"#), @"testing");
```
martinvonz, line 2675: nit: here and below, replace the single "testing" by two "foo bar" or something to make sure that we don't remove internal spaces (the implementation is straight-forward so I'm not worried we actually do)

martinvonz (approved):

PR #5785, opened 2025-02-22
merge_tools: add "ours" and "theirs" merge tools

scott2000:
This is a bit of a hack, but it would be useful to have these available for users until a proper solution is implemented.

Related to #1027.

# Checklist

If applicable:

### DIFF
--- a/cli/src/merge_tools/mod.rs
+++ b/cli/src/merge_tools/mod.rs
@@ -20,6 +20,7 @@ use std::sync::Arc;
 
 use bstr::BString;
 use itertools::Itertools as _;
+use jj_lib::backend::BackendError;
 use jj_lib::backend::FileId;
 use jj_lib::backend::MergedTreeId;
 use jj_lib::config::ConfigGetError;
@@ -32,6 +33,7 @@ use jj_lib::matchers::Matcher;
 use jj_lib::merge::Merge;
 use jj_lib::merge::MergedTreeValue;
 use jj_lib::merged_tree::MergedTree;
+use jj_lib::merged_tree::MergedTreeBuilder;
 use jj_lib::repo_path::InvalidRepoPathError;
 use jj_lib::repo_path::RepoPath;
 use jj_lib::repo_path::RepoPathBuf;
@@ -56,6 +58,8 @@ use crate::config::CommandNameAndArgs;
 use crate::ui::Ui;
 
 const BUILTIN_EDITOR_NAME: &str = ":builtin";
+const OURS_TOOL_NAME: &str = ":ours";
+const THEIRS_TOOL_NAME: &str = ":theirs";
 
 #[derive(Debug, Error)]
 pub enum DiffEditError {
@@ -127,6 +131,8 @@ pub enum MergeToolConfigError {
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum MergeTool {
     Builtin,
+    Ours,
+    Theirs,
     // Boxed because ExternalMergeTool is big compared to the Builtin variant.
     External(Box<ExternalMergeTool>),
 }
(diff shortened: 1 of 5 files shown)


cli/src/config/merge_tools.toml
```diff
@@ -1,3 +1,11 @@
+[merge-tools.ours]
+program = "cp"
+merge-args = ["$left", "$output"]
+
+[merge-tools.theirs]
+program = "cp"
+merge-args = ["$right", "$output"]
+
```
martinvonz, line 8: Does `cp` work on Windows too? Otherwise this can go into `unix.toml`.

scott2000: It looks like the equivalent command on Windows is `xcopy /y SRC DEST`. It looks like there's already an override for the command for a merge tool in `windows.toml`, so maybe I can put it there?

yuja: Is it difficult to implement these in code (something like `:ours`/`:theirs`)? I think we can deprecate them relatively easily, but handling deprecation in user-configurable tools table might be a bit harder.

scott2000: Hmm that's a good point, that might be better. Let me take a look.

cli/tests/test_resolve_command.rs
```diff
@@ -553,6 +553,83 @@ fn test_resolution() {
...
+    -<<<<<<< Conflict 1 of 1
+    -+++++++ Contents of side #1
+     a
+    -------- Contents of base
+    -base
+    -+++++++ Contents of side #2
+    -b
+    ->>>>>>> Conflict 1 of 1 ends
+    [EOF]
+    "##);
```
martinvonz, line 585: I think it's easier to see the result by just displaying the file, i.e. something like this:

```rust
insta::assert_snapshot!(std::fs::read_to_string(repo_path.join("file")).unwrap(), "a
");
```

martinvonz (commented):
It looks like this is ready for review. Don't hesitate to let the reviewers know next time. I don't typically interpret a new push as a signal that I should take a look. I just happened to notice that @yuja's comment seems to have been addressed.

cli/src/merge_tools/mod.rs
```diff
@@ -69,6 +73,8 @@ pub enum DiffEditError {
     Snapshot(#[from] SnapshotError),
     #[error(transparent)]
     Config(#[from] ConfigGetError),
+    #[error("The tool `{tool_name}` cannot be used as a diff editor")]
+    DiffEditNotSupported { tool_name: String },
```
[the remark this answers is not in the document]
yuja: Perhaps, `enum MergeTool` can be split to `MergeTool` and `DiffTool`. It's better to handle configuration error early.

scott2000:
> It looks like this is ready for review. Don't hesitate to let the reviewers know next time. I don't typically interpret a new push as a signal that I should take a look.

Ah yeah, I forgot about that :)

cli/src/merge_tools/mod.rs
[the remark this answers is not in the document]
scott2000: > nit: Can use `tool_name: &'static str`?

I moved it to `MergeToolConfigError` now, so I think it would be good to leave it as `tool_name: String` since `MergeToolConfigError::MergeArgsNotConfigured` also takes `tool_name: String`, and this new variant could be used to handle custom merge tools as well in the future (e.g. if a custom merge tool isn't suitable for diff editing).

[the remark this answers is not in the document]
yuja: Since `:ours`/`:theirs` don't exist in `DiffTool`, I think we can simply remove the error handling. (Alternatively, any unknown `:<name>`s could be mapped to error.)

scott2000:
I think I addressed all of the comments now, could you take another look?

cli/tests/test_resolve_command.rs
```diff
@@ -1871,3 +1871,97 @@ fn test_multiple_conflicts_with_error() {
     [EOF]
     ");
 }
+
+#[test]
+fn test_resolve_with_contents_of_side() {
```
yuja, line 1876: nit: can you apply the same change as 3ba2ea7ace728170ae360aa1f6011d200cd9bc8b?

yuja (approved):
Thanks.

cli/tests/test_resolve_command.rs
scott2000: That's a lot nicer, thanks!

PR #5788, opened 2025-02-23
tests: port `jj_cmd_success()` to `run_jj_in()`

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_file_track_untrack_commands.rs
+++ b/cli/tests/test_file_track_untrack_commands.rs
@@ -34,7 +34,9 @@ fn test_track_untrack() {
     test_env.jj_cmd_ok(&repo_path, &["st"]);
     std::fs::write(repo_path.join(".gitignore"), "*.bak\n").unwrap();
     let files_before = test_env
-        .jj_cmd_success(&repo_path, &["file", "list"])
+        .run_jj_in(&repo_path, ["file", "list"])
+        .success()
+        .stdout
         .into_raw();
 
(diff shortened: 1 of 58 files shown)


cli/tests/test_file_track_untrack_commands.rs
```diff
@@ -34,7 +34,9 @@ fn test_track_untrack() {
     test_env.jj_cmd_ok(&repo_path, &["st"]);
     std::fs::write(repo_path.join(".gitignore"), "*.bak\n").unwrap();
     let files_before = test_env
-        .jj_cmd_success(&repo_path, &["file", "list"])
+        .run_jj_in(&repo_path, ["file", "list"])
+        .success()
+        .stdout
         .into_raw();
```
martinvonz, line 40: The `.success().stdout.into_raw()` pattern seems pretty common. Worth adding a `CommandOutput::raw_stdout()` or something for? Maybe you didn't because naming it is hard?

martinvonz (approved):
Thanks!

cli/tests/test_file_track_untrack_commands.rs
yuja: It might be even better to add helper functions that get commit id, tree content, etc. without executing `jj` command.

Another idea is to implement `Eq` that compares `raw` outputs and `status` only.

martinvonz: Makes sense. Can be done after this PR either way.

PR #5800, opened 2025-02-24
Drop scary red color from immutable divergent commits

martinvonz:
# Checklist

If applicable:


cli/src/config/colors.toml
```diff
@@ -38,8 +38,8 @@
 "tags" = "magenta"
 "git_refs" = "green"
 "git_head" = "green"
-"divergent" = "red"
-"divergent change_id" = "red"
+"mutable divergent" = "red"
+"mutable divergent change_id" = "red"
```
scott2000, line 42: I wonder if it would be good to also change mutable divergent change IDs to be yellow or gray or some other less-scary color than red to make it clear that it's not something that absolutely needs to be resolved. Then maybe immutable and mutable commits could use the same styling?

Also, if immutable divergent change IDs look like normal change IDs and the `??` isn't a different color to make it stand out, then it might make it difficult to see in `jj log` whether to use the change ID or commit ID to refer to the commit. Maybe immutable divergent change IDs could be styled similar to a hidden commit instead to make it clear to users that they should avoid using the change ID?

cli/tests/test_commit_template.rs
```diff
@@ -456,59 +456,13 @@ fn test_log_builtin_templates_colored_debug() {
...
...○>>  \u{1b}[1m\u{1b}[38;5;5m<<log mutable change_id shortest prefix::q>>\u{1b}[0m\u{1b}[38;5;8m<<log mutable change_id shortest rest::pvuntsm>>\u{1b}[39m<<log mutable:: >>\u{1b}[38;5;3m<<log mutable author email local::test.user>>\u{1b}[39m<<log mutable:: >>\u{1b}[38;5;6m<<log mutable committer timestamp local format::2001-02-03 08:05:07>>\u{1b}[39m<<log mutable:: >>\u{1b}[1m\u{1b}[38;5;4m<<log mutable commit_id shortest prefix::2>>\u{1b}[0m\u{1b}[38;5;8m<<log mutable commit_id shortest rest::30dd059>>\u{1b}[39m<<log mutable:: >>\u{1b}[38;5;2m<<log mutable empty::(empty)>>\u{1b}[39m<<log mutable:: >>\u{1b}[38;5;2m<<log mutable empty description placeholder::(no description set)>>\u{1b}[39m<<log mutable::>>\n\u{1b}[1m\u{1b}[38;5;14m<<node immutable::◆>>\u{1b}[0m  \u{1b}[1m\u{1b}[38;5;5m<<log immutable change_id shortest prefix::z>>\u{1b}[0m\u{1b}[38;5;8m<<log immutable change_id shortest rest::zzzzzzz>>\u{1b}[39m<<log immutable:: >>\u{1b}[38;5;2m<<log immutable root::root()>>\u{1b}[39m<<log immutable:: >>\u{1b}[1m\u{1b}[38;5;4m<<log immutable commit_id shortest prefix::0>>\u{1b}[0m\u{1b}[38;5;8m<<log immutable commit_id shortest rest::0000000>>\u{1b}[39m<<log immutable::>>\n[EOF]");
```
yuja, line 460: nit: these formatting change will be reverted in the next `cargo-insta` release.

CHANGELOG.md
```diff
@@ -109,6 +109,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `config-schema.json` now allows an array of strings or nested table for the
   `ui.pager` setting.
 
+* Divergent changes are no longer marked red in immutable revisions. Since the
+  revision is immutable, the user shouldn't take any action, so the red color
+  was unnecessarily alarming.
+
```
calebdw, line 115: I would recommend that the [docs](https://jj-vcs.github.io/jj/latest/FAQ/#how-do-i-deal-with-divergent-changes-after-the-change-id) be updated to let the user know that no action should be taken in this case

ilyagr:
I'm not entirely certain about this, but I think it's worth a try. Currently, I intend to approve this unless others disagree.

For people who understand what's going on, the idea of making just the mutable divergent commits "scary" seems great to me. (We could also later make this highlighting less scary looking, but the point is that these are the actionable situations). My main worry if that some users might not understand why divergent commits -- even with the same change id -- are only *sometimes* highlighted. This would be easier if we formatted change ids of immutable commits differently in general, but I'm not sure how.

----------

Aside: I've thought of putting a symbol like `🔒` or `⚓` next to the change id, but I'm not sure how they are rendered in different terminals. Also, they are wide.

Screenshots:

Ghostty on Mac 
image

Github on Firefox on Mac:
image


**Update:** On Termux (Android), the `⚓` symbol is unfortunately invisible. :( `🔒` is fine. Though, Google's Android Terminal app might be more popular soon (or is it still Pixel-only?), so maybe Termux is less relevant. (I haven't yet tested that app)

martinvonz:
> This would be easier if we formatted change ids of immutable commits differently in general, but I'm not sure how.

Yes, enabling that was another reason for my recent PR that added the `mutable/immutable` labels. We already color immutable commits differently at Google (using a  label based on a custom template function). I think it makes sense to de-emphasize most of the information on immutable commits.

ilyagr:
If you already have a way of doing it at Google, it might be worth making a PR so that we can see whether we can immediately have a consensus on it or not.

(**Update:** Myself, I'm not immediately sure whether I agree with de-emphasizing most of the info on them or not, especially for smaller-than-Google-scale repos, but a picture might be worth 1000 words. Now that I thought of that, perhaps that's exactly the reason you haven't shared the PR yet.)

If we get no strong objection, the main question now in my mind is whether we want to do just this PR for the upcoming release, whether we want both this and something about formatting of immutable commits before the upcoming release, or whether we want to wait with both until after the release.

ilyagr (approved):
OK, I'll approve it as planned.

I'm still not sure whether we should merge this before or after the upcoming release (depending on how we feel about changing the immutable formatting more generally: whether we think we can do it quickly, whether we think it's important to do both together). But your judgement here is at least as good as mine (I don't have a clear plan for immutable commit formatting in mind).

Even if we don't update the immutable commit formatting, merging this now and to see whether this will actually confuse anyone might be a risk worth taking, there's a decent chance it'll be fine.

CHANGELOG.md
martinvonz: This was taken care of in #7061

cli/src/config/colors.toml
martinvonz: I've actually considered not even showing the change id for immutable commits. Or at least we could deemphasize them by using the default color or dim them ("bright black").

joyously: > I've actually considered not even showing the change id for immutable commits.

So, people couldn't see them to use them on options like `-A`, `-B`, `-d` ?

> we could deemphasize them by using the default color or dim them ("bright black").

If it's in the log output, it should be legible. Don't make it more difficult for those of us with poor eyesight.

martinvonz: > So, people couldn't see them to use them on options like `-A`, `-B`, `-d` ?

Correct. The idea is that commit ids are better identifiers for immutable commits. Just a thought.

scott2000: > I've actually considered not even showing the change id for immutable commits. Or at least we could deemphasize them by using the default color or dim them ("bright black").

It might be nice if the commit ID is in the far-left position then (where the change ID normally would be), so that it's still easy to find the correct ID to use. The same might apply for hidden commits as well.

martinvonz: Yes, I was thinking the same. I'm not sure it's better. Good point about hidden commits.

glehmann: Showing the commit-id in place of the change-id for immutable commits would also avoid showing the divergence where it doesn't matter for the user. We wouldn't have seen a divergence on main yesterday, for example (https://discord.com/channels/968932220549103686/968932220549103689/1410735286018768988).
That may be a big change on how change-id are presented to the new users though.

yuja: > The idea is that commit ids are better identifiers for immutable commits.

I tend to use change ids because alphabets are easier to type than numbers.

ilyagr: I also tend to use change ids, and I think moving things around too much for immutable commits might be confusing.

I like better the idea of making immutable change ids use the default color. Using bright black with underlines to show short prefixes might work too, though it wouldn't be my first choice. I might try one or the other for my config.

martinvonz:
Looks like I forgot to merge this after the September release. We're not close to the November release yet, so I think now is fine. I'll merge it.

PR #5803, opened 2025-02-25
cli/tests: move test_git_colocated to gitoxide

bsdinis:
As per #5548

### DIFF
--- a/cli/tests/test_git_colocated.rs
+++ b/cli/tests/test_git_colocated.rs
@@ -15,7 +15,7 @@
 use std::fmt::Write;
 use std::path::Path;
 
-use git2::Oid;
+use testutils::git;
 
 use crate::common::CommandOutput;
 use crate::common::TestEnvironment;
@@ -24,36 +24,26 @@ use crate::common::TestEnvironment;
 fn test_git_colocated() {
     let test_env = TestEnvironment::default();
     let workspace_root = test_env.env_root().join("repo");
-    let git_repo = git2::Repository::init(&workspace_root).unwrap();
+    let git_repo = git::init(&workspace_root);
 
     // Create an initial commit in Git
-    std::fs::write(workspace_root.join("file"), "contents").unwrap();
-    git_repo
-        .index()
-        .unwrap()
-        .add_path(Path::new("file"))
-        .unwrap();
-    let tree1_oid = git_repo.index().unwrap().write_tree().unwrap();
-    let tree1 = git_repo.find_tree(tree1_oid).unwrap();
-    let signature = git2::Signature::new(
-        "Someone",
-        "someone@example.com",
-        &git2::Time::new(1234567890, 60),
+    let tree_id = git::add_commit(
+        &git_repo,
+        "refs/heads/master",
+        "file",
+        b"contents",
+        "initial",
+        &[],
     )
-    .unwrap();
-    git_repo
-        .commit(
-            Some("refs/heads/master"),
-            &signature,
-            &signature,
-            "initial",
-            &tree1,
-            &[],
-        )
-        .unwrap();
+    .tree_id;
+    git::checkout_tree_index(&git_repo, tree_id);
+    assert_eq!(
+        std::fs::read(workspace_root.join("file")).unwrap(),
+        b"contents"
+    );
     insta::assert_snapshot!(
-        git_repo.head().unwrap().peel_to_commit().unwrap().id().to_string(),
-        @"e61b6729ff4292870702f2f72b2a60165679ef37"
+        git_repo.head_id().unwrap().to_string(),
+        @"97358f54806c7cd005ed5ade68a779595efbae7e"
     );
 
     // Import the repo
@@ -61,42 +51,43 @@ fn test_git_colocated() {
         .run_jj_in(&workspace_root, ["git", "init", "--git-repo", "."])
         .success();
     insta::assert_snapshot!(get_log_output(&test_env, &workspace_root), @r"
-    @  3e9369cd54227eb88455e1834dbc08aad6a16ac4
-    ○  e61b6729ff4292870702f2f72b2a60165679ef37 master git_head() initial
+    @  c3a12656d5027825bc69f40e11dc0bb381d7c277
+    ○  97358f54806c7cd005ed5ade68a779595efbae7e master git_head() initial
     ◆  0000000000000000000000000000000000000000
     [EOF]
     ");
     insta::assert_snapshot!(
-        git_repo.head().unwrap().peel_to_commit().unwrap().id().to_string(),
-        @"e61b6729ff4292870702f2f72b2a60165679ef37"
+        git_repo.head_id().unwrap().to_string(),
+        @"97358f54806c7cd005ed5ade68a779595efbae7e"
     );
 
     // Modify the working copy. The working-copy commit should changed, but the Git
     // HEAD commit should not
     std::fs::write(workspace_root.join("file"), "modified").unwrap();
     insta::assert_snapshot!(get_log_output(&test_env, &workspace_root), @r"
-    @  4f546c80f30abc0803fb83e5032a4d49fede4d68
-    ○  e61b6729ff4292870702f2f72b2a60165679ef37 master git_head() initial
+    @  59642000f061cf23eb37ab6eecce428afe7824da
+    ○  97358f54806c7cd005ed5ade68a779595efbae7e master git_head() initial
     ◆  0000000000000000000000000000000000000000
     [EOF]
     ");
     insta::assert_snapshot!(
-        git_repo.head().unwrap().peel_to_commit().unwrap().id().to_string(),
-        @"e61b6729ff4292870702f2f72b2a60165679ef37"
+        git_repo.head_id().unwrap().to_string(),
+        @"97358f54806c7cd005ed5ade68a779595efbae7e"
     );
 
     // Create a new change from jj and check that it's reflected in Git
     test_env.run_jj_in(&workspace_root, ["new"]).success();
     insta::assert_snapshot!(get_log_output(&test_env, &workspace_root), @r"
-    @  0e2301a42b288b9568344e32cfdd8c76d1e56a83
-    ○  4f546c80f30abc0803fb83e5032a4d49fede4d68 git_head()
-    ○  e61b6729ff4292870702f2f72b2a60165679ef37 master initial
+    @  fd4e130e43068d76ec6eb2c6df01653ea12eccb4
+    ○  59642000f061cf23eb37ab6eecce428afe7824da git_head()
+    ○  97358f54806c7cd005ed5ade68a779595efbae7e master initial
     ◆  0000000000000000000000000000000000000000
     [EOF]
     ");
+    assert!(git_repo.head().unwrap().is_detached());
     insta::assert_snapshot!(
-        git_repo.head().unwrap().target().unwrap().to_string(),
-        @"4f546c80f30abc0803fb83e5032a4d49fede4d68"
+        git_repo.head_id().unwrap().to_string(),
+        @"59642000f061cf23eb37ab6eecce428afe7824da"
     );
 }
 
@@ -261,7 +252,7 @@ fn test_git_colocated_export_bookmarks_on_snapshot() {
 
     let test_env = TestEnvironment::default();
     let workspace_root = test_env.env_root().join("repo");
-    let git_repo = git2::Repository::init(&workspace_root).unwrap();
+    let git_repo = git::init(&workspace_root);
     test_env
         .run_jj_in(&workspace_root, ["git", "init", "--git-repo", "."])
         .success();
@@ -288,8 +279,7 @@ fn test_git_colocated_export_bookmarks_on_snapshot() {
     insta::assert_snapshot!(git_repo
         .find_reference("refs/heads/foo")
         .unwrap()
-        .target()
-        .unwrap()
+        .id()
         .to_string(), @"4d2c49a8f8e2f1ba61f48ba79e5f4a5faa6512cf");
 }
 
@@ -297,7 +287,7 @@ fn test_git_colocated_export_bookmarks_on_snapshot() {
 fn test_git_colocated_rebase_on_import() {
     let test_env = TestEnvironment::default();
     let workspace_root = test_env.env_root().join("repo");
-    let git_repo = git2::Repository::init(&workspace_root).unwrap();
+    let git_repo = git::init(&workspace_root);
     test_env
         .run_jj_in(&workspace_root, ["git", "init", "--git-repo", "."])
         .success();
@@ -320,15 +310,23 @@ fn test_git_colocated_rebase_on_import() {
 
     // Move `master` backwards, which should result in commit2 getting hidden,
     // and the working-copy commit rebased.
-    let commit2_oid = git_repo
-        .find_branch("master", git2::BranchType::Local)
+    let parent_commit = git_repo
+        .find_reference("refs/heads/master")
         .unwrap()
-        .get()
-        .target()
+        .peel_to_commit()
+        .unwrap()
+        .parent_ids()
+        .next()
+        .unwrap()
+        .detach();
+    git_repo
+        .reference(
+            "refs/heads/master",
+            parent_commit,
+            gix::refs::transaction::PreviousValue::Any,
+            "update ref",
+        )
         .unwrap();
-    let commit2 = git_repo.find_commit(commit2_oid).unwrap();
-    let commit1 = commit2.parents().next().unwrap();
-    git_repo.branch("master", &commit1, true).unwrap();
     insta::assert_snapshot!(get_log_output(&test_env, &workspace_root), @r"
     @  15b1d70c5e33b5d2b18383292b85324d5153ffed
     ○  47fe984daf66f7bf3ebf31b9cb3513c995afb857 master git_head() add a file
@@ -349,7 +347,7 @@ fn test_git_colocated_rebase_on_import() {
 fn test_git_colocated_bookmarks() {
     let test_env = TestEnvironment::default();
     let workspace_root = test_env.env_root().join("repo");
-    let git_repo = git2::Repository::init(&workspace_root).unwrap();
+    let git_repo = git::init(&workspace_root);
     test_env
         .run_jj_in(&workspace_root, ["git", "init", "--git-repo", "."])
         .success();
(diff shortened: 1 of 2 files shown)


bsdinis:
There are a couple of notes/questions I have with this:
 - In the second commit, using gitoxide changes what `jj` considers to be mutable in the cloned repo. Not sure what to make of this, maybe it's something to do with the default branch?
 - I've isolated the third commit with the most complex test (involves the index)
 
 Both commits should be squashed before merging

cli/tests/test_git_colocated.rs
```diff
@@ -24,79 +24,63 @@ use crate::common::TestEnvironment;
...
+    @  5b2eb7bbbedf5a888cc8810b5d74bd8eb0b2d470
+    ○  10a2a8c6b7264e07dd8220775a512527474950f7 git_head()
+    ○  5154199a23c1964598f6245ef8a0b3112510945a master initial
     ◆  0000000000000000000000000000000000000000
     [EOF]
     ");
     insta::assert_snapshot!(
-        git_repo.head().unwrap().target().unwrap().to_string(),
-        @"4f546c80f30abc0803fb83e5032a4d49fede4d68"
+        git_repo.head().unwrap().into_peeled_id().unwrap().to_string(),
```
yuja, line 82: nit: might be better to check that the `HEAD` ref is NOT symbolic ref here.

btw, maybe you can use `git_repo.head_id()` instead of `into_peeled_id()`.

cli/tests/test_git_colocated.rs
```diff
@@ -528,40 +509,24 @@ fn test_git_colocated_conflicting_git_refs() {
...
     let workspace_root = test_env.env_root().join("repo");
-    let git_repo = git2::Repository::init(&workspace_root).unwrap();
+    let git_repo = git::init(&workspace_root);
     test_env
         .run_jj_in(&workspace_root, ["git", "init", "--git-repo", "."])
         .success();
 
     // Create an initial commit in Git
     // We use this to set HEAD to master
-    std::fs::write(workspace_root.join("file"), "contents").unwrap();
```
yuja, line 538: same here. can you check if the working directory is clean?

cli/tests/test_git_colocated.rs
```diff
@@ -601,7 +601,7 @@ fn test_git_colocated_fetch_deleted_or_moved_bookmark() {
     ├─╯
     │ ○  c49ec4fb50844d0e693f1609da970b11878772ee B_to_delete B_to_delete
     ├─╯
-    ◆  a7e4cec4256b7995129b9d1e1bda7e1df6e60678 A git_head() A
+    ○  a7e4cec4256b7995129b9d1e1bda7e1df6e60678 A git_head() A
```
yuja, line 641: > In the second commit, using gitoxide changes what jj considers to be mutable in the cloned repo. Not sure what to make of this, maybe it's something to do with the default branch?

Perhaps, `git::clone()` doesn't update `refs/remotes/origin/HEAD`?

lib/testutils/src/git.rs
```diff
@@ -155,6 +155,19 @@ pub fn write_commit(
     .detach()
 }
 
+pub fn set_ref_to_id(repo: &gix::Repository, reference: &str, target: gix::ObjectId) {
```
yuja, line 158: Are we going to replace existing `repo.reference()` callers?

cli/tests/test_git_colocated.rs
```diff
@@ -86,30 +86,60 @@ fn test_git_colocated() {
 
 #[test]
 fn test_git_colocated_unborn_bookmark() {
+    use std::sync::Arc;
+    use std::sync::Mutex;
```
yuja, line 90: maybe we can just reload the index from disk?

cli/tests/test_git_colocated.rs
```diff
@@ -86,30 +86,60 @@ fn test_git_colocated() {
...
     };
+
+    // checkout index (i.e., drop the in-memory changes)
     let checkout_index = || {
-        let mut index = git_repo.index().unwrap();
-        index.read(true).unwrap(); // discard in-memory cache
-        git_repo.checkout_index(Some(&mut index), None).unwrap();
+        *index.lock().unwrap() = gix::index::File::from_state(
+            gix::index::State::new(git_repo.object_hash()),
+            git_repo.index_path(),
```
yuja, line 125: Does this really update the working copy to match the index?

yuja (commented):
> Both commits should be squashed before merging

I like the current state because the last patch is complex. I think the first two patches could be squashed if you like.

cli/tests/test_git_colocated.rs
bsdinis: done

cli/tests/test_git_colocated.rs
```diff
@@ -1333,11 +1305,22 @@ fn get_log_output(test_env: &TestEnvironment, workspace_root: &Path) -> CommandO
...
+        .status(gix::progress::Discard)
         .unwrap()
-        .diff_index_to_workdir(None, Some(git2::DiffOptions::new().update_index(true)))
+        .into_index_worktree_iter(None)
+        .unwrap();
+
+    // need to explicitly iterate over the changes to recreate the index
+
+    for item in iter.by_ref() {
+        let _item = item.unwrap();
```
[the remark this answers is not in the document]
bsdinis: done

lib/testutils/src/git.rs
bsdinis: removed

cli/tests/test_git_colocated.rs
```diff
@@ -24,79 +24,63 @@ use crate::common::TestEnvironment;
 fn test_git_colocated() {
     let test_env = TestEnvironment::default();
     let workspace_root = test_env.env_root().join("repo");
-    let git_repo = git2::Repository::init(&workspace_root).unwrap();
+    let git_repo = git::init(&workspace_root);
 
     // Create an initial commit in Git
-    std::fs::write(workspace_root.join("file"), "contents").unwrap();
```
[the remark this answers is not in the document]
yuja: `jj` just imports `.git/HEAD`, and then snapshots working-copy changes on top.

cli/tests/test_git_colocated.rs
bsdinis: that's not it: it comes down to the default branch

cli/tests/test_git_colocated.rs
bsdinis: dnoe

cli/tests/test_git_colocated.rs
bsdinis: so I fixed this by changing the name of the A bookmark to "main"
this way it gets automatically tracked as immutable.

My understanding here is that git2 probably figured out the default branch as the one that was written to first, and gitoxide doesn't have the same mechanism

cli/tests/test_git_colocated.rs
bsdinis: ok, I think I got it

cli/tests/test_git_colocated.rs
bsdinis: now it does

cli/tests/test_git_colocated.rs
bsdinis: I looked into the tempdir and `refs/remotes/origin/HEAD` was a symbolic link to `refs/remotes/origin/A`, is there something else that needs to be done?

bsdinis: Might be a gitoxide problem: https://github.com/GitoxideLabs/gitoxide/issues/1865

yuja: We'll probably need to create `refs/remotes/origin/HEAD` manually. I have no idea whether `gix` API is supposed to do because it's called a `fetch`, not `clone`.

cli/tests/test_git_colocated.rs
bsdinis: So, I'm a bit confused on what the way this is supposed to do. I've tried instead of checking out the head checking out the index corresponding to the latest `add_file_to_index`. This fails because after the first `checkout_index` the point there is that, according to the comment: `// Staged change shouldn't persist`.

I'm unsure what this is supposed to do, if not checking out the head index. 
Note that I'm unsure if with `git_repo.index()` doesn't load the HEAD index

cli/tests/test_git_colocated.rs
bsdinis: I'll wait for Byron to weigh in, they usually have a quick turnaround

cli/tests/test_git_colocated.rs
yuja: This is the test for the following scenario:

1. user add some change to staging area by using `git add`
2. user checks out sibling commit by using `jj new`
3. (2) shouldn't clobber symbolic `HEAD` (because the working-copy parent is unchanged)
4. (2) should discard the staged changes (because new working-copy commit is checked out)

We can either (a) assert that the index contents are identical to working directory files or (b) snapshot working-directory content after `checkout_index()`. The current implementation does (b) because it was easier iirc.

bsdinis: Doesn't `jj` move the HEAD (on `jj new`) and as such move the index?

bsdinis: Another way of looking at it, all the existing code is doing is the following:
```
let mut index = git_repo.index().unwrap();
index.read(true).unwrap(); // discard in memory cache 
git_repo.checkout_index(Some(&mut index), None).unwrap();
```

So, with this code, this is just checking out the index associated with the `HEAD` commit, which seems equivalent to the `gitoxide` code.

yuja: No, it checks out the `index` (staged changes), not the `HEAD` tree. They are different objects. (To be precise, "index" isn't a Git object.)

> Doesn't jj move the HEAD (on jj new) and as such move the index?

`jj new @-` wouldn't, for example.

bsdinis: makes sense, i think i figured it out with the latest revision

bsdinis: okay, gave up on gitoxide for this, shelling out to `git`

cli/tests/test_git_colocated.rs
```diff
@@ -374,11 +375,15 @@ fn test_git_colocated_bookmarks() {
         .run_jj_in(&workspace_root, ["bookmark", "create", "-r@", "master"])
         .success();
     insta::assert_snapshot!(
-        git_repo.find_reference("refs/heads/master").unwrap().target().unwrap().to_string(),
+        git_repo.find_reference("refs/heads/master").unwrap().target().id().to_string(),
         @"3560559274ab431feea00b7b7e0b9250ecce951f"
     );
+    assert!(
+        git_repo.head().unwrap().try_into_referent().is_none(),
```
yuja, line 419: nit: `.is_detached()

cli/tests/test_git_colocated.rs
```diff
@@ -844,15 +846,19 @@ fn test_git_colocated_squash_undo() {
...
-    let git_repo = git2::Repository::init(&repo_path).unwrap();
+    let git_repo = git::init(&repo_path);
     test_env
         .run_jj_in(&repo_path, ["git", "init", "--git-repo=."])
         .success();
 
     // Create new HEAD
     test_env.run_jj_in(&repo_path, ["new"]).success();
+    assert!(
+        git_repo.head().unwrap().try_into_referent().is_none(),
```
yuja, line 894: nit: `.is_detached()`

cli/tests/test_git_colocated.rs
```diff
@@ -880,8 +886,12 @@ fn test_git_colocated_undo_head_move() {
     ◆  0000000000000000000000000000000000000000
     [EOF]
     ");
+    assert!(
+        git_repo.head().unwrap().try_into_referent().is_none(),
```
yuja, line 927: nit: `.is_detached()`

lib/testutils/src/git.rs
```diff
@@ -56,23 +56,24 @@ pub fn init_bare(directory: impl AsRef<Path>) -> gix::Repository {
...
-        .main_worktree(gix::progress::Discard, &gix::interrupt::IS_INTERRUPTED)
-        .unwrap();
+pub fn clone(dest_path: &Path, repo_url: &str) -> gix::Repository {
+    // gitoxide doesn't seem to write the `refs/remotes` to disk on clone
+    // (https://github.com/GitoxideLabs/gitoxide/issues/1865)
+    //
+    // it'd be nice not to have the hard-dependency on `git` at test time, but it
+    // seems unavoidable
+    assert!(std::process::Command::new("git")
+        .args(["clone", repo_url, dest_path.to_str().unwrap()])
```
yuja, line 66: nit: `.arg(dest_path)` also works

lib/testutils/src/git.rs
```diff
@@ -56,23 +56,24 @@ pub fn init_bare(directory: impl AsRef<Path>) -> gix::Repository {
...
+        .stdin(std::process::Stdio::null())
+        .stdout(std::process::Stdio::null())
+        .stderr(std::process::Stdio::null())
```
yuja, line 69: nit: might be better to not redirect stdout and stderr so we can debug problem on error.

Another option is to use `.output()` and include the output in assertion message.
https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.output

lib/testutils/src/git.rs
```diff
@@ -56,23 +56,24 @@ pub fn init_bare(directory: impl AsRef<Path>) -> gix::Repository {
...
+        .spawn()
+        .unwrap()
+        .wait()
+        .unwrap()
```
yuja, line 73: nit: `.status()` or `.output()`

cli/tests/test_git_colocated.rs
```diff
@@ -98,28 +98,59 @@ fn test_git_colocated() {
...
+            git_repo
+                .edit_tree(gix::ObjectId::empty_tree(git_repo.object_hash()))
+                .unwrap(),
+        );
+        let blob_oid = git_repo.write_blob(data).unwrap();
+        tree_editor
+            .upsert(name, gix::object::tree::EntryKind::Blob, blob_oid)
+            .unwrap();
+        let tree_id = tree_editor.write().unwrap().detach();
+        git::checkout_tree_index(&git_repo, tree_id);
```
yuja, line 115: You need to write the index as well.

I think we can use `git::IndexManager` after 63f8743. It should be better to not touch files other than `name`.

cli/tests/test_git_colocated.rs
```diff
@@ -98,28 +98,59 @@ fn test_git_colocated() {
...
-        let mut index = git_repo.index().unwrap();
-        index.read(true).unwrap(); // discard in-memory cache
-        git_repo.checkout_index(Some(&mut index), None).unwrap();
+        let mut index = gix::index::File::at_or_default(
+            git_repo.index_path(),
+            git_repo.object_hash(),
+            true,
+            gix::index::decode::Options::default(),
+        )
+        .unwrap();
```
yuja, line 126: nit: `git_repo.open_index()`?

cli/tests/test_git_colocated.rs
```diff
@@ -98,28 +98,59 @@ fn test_git_colocated() {
...
+    assert!(git_repo.head().unwrap().is_unborn());
     assert_eq!(
-        git_repo.find_reference("HEAD").unwrap().symbolic_target(),
-        Some("refs/heads/master")
+        git_repo
+            .find_reference("HEAD")
+            .unwrap()
+            .target()
+            .try_name()
+            .unwrap()
```
yuja, line 151: nit: `head_name()` also works. iirc, we had to use `.find_reference("HEAD")` in git2 because `repo.head()` would fail on unborn case.

cli/tests/test_git_colocated.rs
```diff
@@ -24,79 +24,73 @@ use crate::common::TestEnvironment;
...
-    ○  4f546c80f30abc0803fb83e5032a4d49fede4d68 git_head()
-    ○  e61b6729ff4292870702f2f72b2a60165679ef37 master initial
+    @  fd4e130e43068d76ec6eb2c6df01653ea12eccb4
+    ○  59642000f061cf23eb37ab6eecce428afe7824da git_head()
+    ○  97358f54806c7cd005ed5ade68a779595efbae7e master initial
     ◆  0000000000000000000000000000000000000000
     [EOF]
     ");
+    assert!(
+        git_repo.head().unwrap().try_into_referent().is_none(),
```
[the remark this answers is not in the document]
bsdinis: here we actually want to check that it's not a symbolic reference, but I agree that `is_detached() || is_unborn()` is cleaner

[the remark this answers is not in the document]
yuja: iirc, `is_detached()` means "not a symbolic ref." Unborn ref is a symbolic ref that points to missing target (i.e. `head_id()` would fail in that case.)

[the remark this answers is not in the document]
bsdinis: correct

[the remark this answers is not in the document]
bsdinis: done

lib/testutils/src/git.rs
```diff
@@ -56,23 +56,26 @@ pub fn init_bare(directory: impl AsRef<Path>) -> gix::Repository {
...
-        .fetch_then_checkout(gix::progress::Discard, &gix::interrupt::IS_INTERRUPTED)
-        .unwrap();
-    let (repo, _outcome) = prepare_checkout
-        .main_worktree(gix::progress::Discard, &gix::interrupt::IS_INTERRUPTED)
+pub fn clone(dest_path: &Path, repo_url: &str) -> gix::Repository {
+    // gitoxide doesn't seem to write the `refs/remotes` to disk on clone
+    // (https://github.com/GitoxideLabs/gitoxide/issues/1865)
+    //
+    // it'd be nice not to have the hard-dependency on `git` at test time, but it
+    // seems unavoidable
```
yuja, line 64: nit: can you update the comment? I misunderstood the problem. gix does write `remotes/origin/HEAD`, but not as a symbolic ref.

fwiw, I think it's good to use the `git` command in tests if there are no high-level API in gix. If we write non-trivial amount of code to replicate the git behavior, we'll have to write tests for that.

lib/testutils/src/git.rs
```diff
@@ -56,23 +56,26 @@ pub fn init_bare(directory: impl AsRef<Path>) -> gix::Repository {
...
+        .args(["clone", repo_url])
+        .arg(dest_path)
+        .output()
         .unwrap();
-
-    repo
+    assert!(
+        output.status.success(),
+        "git cloning failed with exit code {:?}:\n{}\n----- stderr -----\n{}",
+        output.status.code(),
```
yuja, line 73: nit: `output.status` implements `Display`, so this could be `"git cloning failed with {} ...", output.status, ...`

yuja (approved):
Thanks.

lib/testutils/src/git.rs
bsdinis: will do, and generally agree with the sentiment going forward

PR #5806, opened 2025-02-25
config: Allow hiding the 'how to resolve conflicts' hint

### LINKED ISSUE #5766: FR: Allow hiding certain hints
Various operations in `jj` produce hint text. This is obviously very useful but after using it for a while, some of them are not particularly useful.

One such example is the following hint, which appears when new conflicts do:

```
Hint: To resolve the conflicts, start by updating to the first one:
  jj new <changeid>
Then use `jj resolve`, or edit the conflict markers in the file directly.
Once the conflicts are resolved, you may want to inspect the result with `jj diff`.
Then run `jj squash` to move the resolution into the conflicted commit.
```

Once you're fairly familiar with `jj`, this is ~5 lines of terminal output that don't need to be read each time.

### Git's approach
`git` has a (fairly extensive) set of config options under `[advice]` ([docs](https://git-scm.com/docs/git-config#Documentation/git-config.txt-advice)), each of which allows you to disable a certain hint message. Personally I quite like this design since it lets you silence hints for workflows you're very familiar with, while keeping them around for ones you run into less often. It also means new hints are shown by default.

jgilchrist:
I've started as minimal as I could as I'm sure there will be discussions around both approach and naming of the config options. I also haven't updated `config-schema.json` or the changelog but will do if there's general agreement on the approach.

Currently you can disable the following hint:

```
Hint: To resolve the conflicts, start by updating to the first one:
  jj new <changeid>
Then use `jj resolve`, or edit the conflict markers in the file directly.
Once the conflicts are resolved, you may want to inspect the result with `jj diff`.
Then run `jj squash` to move the resolution into the conflicted commit.
```

by setting:

```toml
[hints]
howto-resolve-conflicts = false
```

I've gone with `[hints]` instead of `[advice]` since jj explicitly calls them hints in the CLI output.

Fixes #5766 

# Checklist

If applicable:


cli/src/cli_util.rs
```diff
@@ -2293,7 +2293,12 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
             .commits(repo.store())
             .try_collect()?;
 
-        if !root_conflict_commits.is_empty() {
+        let should_show_resolve_conflicts_hint = self
+            .settings()
+            .get_bool("hints.howto-resolve-conflicts")
+            .unwrap_or(true);
```
yuja, line 2299: nit: can you add the default value to `config/*.toml` instead of ignoring error? Suppose the list will grow, it might make sense to add `config/hints.toml`.

yuja (commented):
It might be better to add the hint key to the message so user doesn't have to grep the codebase or docs.
```
Hint: To resolve the conflicts, start by updating to the first one:
  ..
Then run `jj squash` to move the resolution into the conflicted commit. [how-to-resolve-conflicts]
```

I don't have strong feeling about naming.

jgilchrist:
> It might be better to add the hint key to the message so user doesn't have to grep the codebase or docs.

I can certainly see the argument for that. On the other hand, it does increase the noise for folks who aren't going to turn them off. I suppose I saw this as being likely used by power users who are going to expend a bit more effort to get their terminal output 'just right', and I suspect that those users might be more comfortable hunting around in the docs.

No strong feelings one way or another, happy to go with the consensus!

cli/src/cli_util.rs
jgilchrist: That's done - I've gone with `hints.toml`

cli/src/config/hints.toml
```diff
@@ -0,0 +1,2 @@
+[hints]
+howto-resolve-conflicts = true
```
yuja, line 2: nit: maybe it should be `how-to-`?

cli/src/cli_util.rs
```diff
@@ -2293,7 +2293,10 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
             .commits(repo.store())
             .try_collect()?;
 
-        if !root_conflict_commits.is_empty() {
+        let should_show_resolve_conflicts_hint =
```
yuja, line 2296: nit: I would move this to the function start since we don't have to compute the revset if disabled. This function is actually `show_hint_on_how_to_resolve_conflicts()`.

yuja (commented):
The code looks good to me, but I'll leave it open for a while to gather comments.

cli/src/config/hints.toml
jgilchrist: I did think about that. FWIW my reasoning was that I was going for 'how-to' (i.e. a how-to on resolving conflicts) rather than 'how to' and given it's already hyphenated I dropped the hyphen. I'm happy to add it.

Perhaps just `hints.resolving-conflicts` is better though?

jgilchrist:
> Some ideas around how `git` handles disabling hints if you need some inspiration: https://stackoverflow.com/questions/55463863/how-to-turn-off-the-help-hints-in-git-output

@zx8 Is there anything there you'd particularly like to see? The method I went with is already inspired by Git's implementation. I haven't added any switches for turning off everything (e.g. Git's `--no-advice` or `GIT_ADVICE=0` but they could easily be added later.

cli/src/config/hints.toml
jgilchrist: I've dropped the 'how to' part - I suppose there's quite a good chance that it would be a prefix on all of the hints since a lot of hints are going to point out how to do something so I'm not sure it adds anything.

cli/src/cli_util.rs
```diff
@@ -2283,6 +2283,13 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
         repo: &ReadonlyRepo,
         conflicted_commits: Vec<CommitId>,
     ) -> Result<(), CommandError> {
+        let should_show_resolve_conflicts_hint =
```
yuja, line 2286: nit: since the expression is simple, I would inline it.

cli/src/config/hints.toml
```diff
@@ -0,0 +1,2 @@
+[hints]
+resolving-conflicts = true
```
yuja, line 2: Can you also update `config-schema.json`?

I don't think we'll need detailed documentation per hint key (because it would be tedious to update.) Something like `"aliases"` or `"template-aliases"` should be good.

yuja (approved):
Since there are no other comments, I think this is good to go. Thanks.

cli/src/config/hints.toml
jgilchrist: I've done that. But presumably it's best to document the list of hint keys that you can hide somewhere, similar to [git's docs?](https://git-scm.com/docs/git-config#Documentation/git-config.txt-advice).

cli/src/config-schema.json
```diff
@@ -700,6 +700,13 @@
                     "default": false
                 }
             }
+        },
+        "hints": {
+            "type": "object",
+            "description": "Various hints in jj's UI that can be disabled",
+            "additionalProperties": {
+                "type": "string"
```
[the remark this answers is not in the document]
ilyagr: I think `boolean` is right.

I'd also recommend making a specific boolean for the `resolving-conflict` key instead of using `additionalProperties`. This can then serve as docs for what hints are possible. You can add some help to a "description" field. (If this feels onerous, you can leave it to others)

We currently don't have a perfect way to surface these kinds of docs, but I wonder if we should/could start auto-generating a reference from the schema at some point.

cli/src/cli_util.rs
```diff
@@ -2283,6 +2283,10 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
         repo: &ReadonlyRepo,
         conflicted_commits: Vec<CommitId>,
     ) -> Result<(), CommandError> {
+        if !self.settings().get_bool("hints.resolving-conflicts")? {
```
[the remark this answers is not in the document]
ilyagr: (But if you want to merge this before the release tomorrow, it could be a separate PR)

cli/src/config-schema.json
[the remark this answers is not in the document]
ilyagr: > I'd also recommend making a specific boolean for the resolving-conflict key instead of using additionalProperties.

I realized that Yuya suggested the opposite. It's fine to keep it as-is.

cli/src/config/hints.toml
ilyagr: > That's one reason I suggested `Hint: {message} [{key}]` suffix.

I like this idea. I think I'd ultimately want something more structured (so that we can e.g. generate the json schema from the same source of truth), but until then, this would work. Even if the source of truth becomes more structured, printing hints in this way might still make sense.

Yuya, correct me if I'm wrong, but I don't think you meant this to be a blocking commend, did you? I think of this as a plan for the future; once there is more than one hint config, the message can be changed to this form.

yuja: No blocker. Things have to be fixed are just "string" -> "boolean".

cli/src/config-schema.json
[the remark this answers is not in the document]
jgilchrist: Thanks @yuja @ilyagr - just a plain oversight here on my part. Happy to leave it as-is for now with a view to improving the docs here when there are more examples of hints to hide.

cli/src/cli_util.rs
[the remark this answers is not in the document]
jgilchrist: I've added those - for `test_repo_change_report.rs` it was a little easier to add a separate test but let me know if you were thinking differently

cli/tests/test_repo_change_report.rs
```diff
@@ -224,3 +224,32 @@ fn test_report_conflicts_with_divergent_commits() {
...
+    let repo_path = test_env.env_root().join("repo");
+
+    std::fs::write(repo_path.join("file"), "A\n").unwrap();
+    test_env.run_jj_in(&repo_path, ["commit", "-m=A"]).success();
+    std::fs::write(repo_path.join("file"), "B\n").unwrap();
+    test_env.run_jj_in(&repo_path, ["commit", "-m=B"]).success();
+    std::fs::write(repo_path.join("file"), "C\n").unwrap();
+    test_env.run_jj_in(&repo_path, ["commit", "-m=C"]).success();
+
+    let output = test_env.run_jj_in(&repo_path, ["rebase", "-s=description(B)", "-d=root()", "--config", "hints.resolving-conflicts=false"]);
```
yuja, line 241: nit: you'll need to run `cargo fmt`, and `"--config=hints..."` will save vertical space.

yuja (approved):

cli/tests/test_repo_change_report.rs
jgilchrist: Strange, I did try putting it in a single argument and got an error. I just tried again and it's working fine. I think perhaps I missed the `=` the first time around. I've changed that and run `cargo fmt`

ilyagr (approved):
LGTM, thank you!

jgilchrist:
Thanks all!

PR #5811, opened 2025-02-25
cli `git fetch`: support string pattern syntax in `remote` option

unexge:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * The new `jj sign` and `jj unsign` commands allow for signing/unsigning commits.
 
+* `jj git fetch` now supports [string pattern syntax](docs/revsets.md#string-patterns)
+  on `--remote` option and `git.fetch` configuration.
+
 ### Fixed bugs
 
 * `jj status` now shows untracked files under untracked directories.
--- a/cli/src/commands/git/fetch.rs
+++ b/cli/src/commands/git/fetch.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashSet;
+
 use clap_complete::ArgValueCandidates;
 use itertools::Itertools;
 use jj_lib::config::ConfigGetResultExt as _;
@@ -23,6 +25,8 @@ use jj_lib::str_util::StringPattern;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::WorkspaceCommandHelper;
 use crate::cli_util::WorkspaceCommandTransaction;
+use crate::command_error::config_error;
+use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::commands::git::get_single_remote;
 use crate::complete;
@@ -54,12 +58,20 @@ pub struct GitFetchArgs {
     ///
     /// This defaults to the `git.fetch` setting. If that is not configured, and
     /// if there are multiple remotes, the remote named "origin" will be used.
+    ///
+    /// By default, the specified remote names matches exactly. Use a [string
+    /// pattern], e.g. `--remote 'glob:*'`, to select remotes using
+    /// patterns.
+    ///
+    /// [string pattern]:
+    ///     https://jj-vcs.github.io/jj/latest/revsets#string-patterns
     #[arg(
         long = "remote",
         value_name = "REMOTE",
+        value_parser = StringPattern::parse,
         add = ArgValueCandidates::new(complete::git_remotes),
     )]
-    remotes: Vec<String>,
+    remotes: Vec<StringPattern>,
     /// Fetch from all remotes
     #[arg(long, conflicts_with = "remotes")]
     all_remotes: bool,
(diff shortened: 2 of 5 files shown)


cli/src/commands/git/fetch.rs
```diff
@@ -72,13 +82,32 @@ pub fn cmd_git_fetch(
...
+        });
+
+    if !remote_patterns.is_empty() {
+        let all_remotes = git::get_all_remote_names(workspace_command.repo().store())?;
+        remotes.extend(
+            all_remotes
+                .into_iter()
+                .filter(|r| remote_patterns.iter().any(|p| p.matches(r))),
+        );
+    }
```
yuja, line 109: In `jj bookmark` subcommands, we report patterns not matching anything as error, instead of handling exact patterns differently. (see `find_bookmarks_with()`.) I don't have preference over these, but for consistency, I think it's better to process remote name patterns in the same way.

cli/tests/test_git_fetch.rs
```diff
@@ -297,6 +297,182 @@ fn test_git_fetch_multiple_remotes(subprocess: bool) {
...
+    test_env
+        .run_jj_in(
+            &repo_path,
+            [
+                "git",
+                "fetch",
+                "--remote",
+                "glob:rem*",
+                "--remote",
+                "glob:upstream*",
```
yuja, line 353: nit: `"--remote=glob:..."` to save vertical space?

cli/src/commands/git/fetch.rs
```diff
@@ -108,12 +140,16 @@ fn get_default_fetch_remotes(
...
-        Ok(vec![remote])
+        Ok(vec![StringPattern::exact(remote)])
     } else {
-        Ok(vec![DEFAULT_REMOTE.to_owned()])
+        Ok(vec![StringPattern::exact(DEFAULT_REMOTE)])
     }
 }
 
+fn parse_remote_pattern(remote: &str) -> Result<StringPattern, CommandError> {
+    StringPattern::parse(remote).map_err(user_error)
```
yuja, line 150: nit: `config_error()` might be more appropriate here

CHANGELOG.md
yuja, line 1: nit: can you also update `docs/config.md`?

yuja (commented):
Thanks!

cli/src/commands/git/fetch.rs
```diff
@@ -72,13 +83,50 @@ pub fn cmd_git_fetch(
...
+            let matching_remotes = all_remotes
+                .iter()
+                .filter(|r| pattern.matches(r))
+                .collect::<Vec<_>>();
+            if matching_remotes.is_empty() {
+                Either::Left(pattern)
+            } else {
+                Either::Right(matching_remotes)
+            }
+        });
```
yuja, line 107: I think it's simpler to use plain `for` loop instead of building nested `remotes: Vec<Vec<&String>>`. Maybe we can also use `HashSet` or `IndexSet` for deduplication, but I'm not sure if that makes the code more readable. The performance wouldn't matter here.
https://docs.rs/indexmap/2.7.1/indexmap/set/struct.IndexSet.html

cli/src/commands/git/fetch.rs
```diff
@@ -93,13 +141,16 @@ const DEFAULT_REMOTE: &str = "origin";
...
+) -> Result<Vec<StringPattern>, CommandError> {
     const KEY: &str = "git.fetch";
     let settings = workspace_command.settings();
-    if let Ok(remotes) = settings.get(KEY) {
-        Ok(remotes)
+    if let Ok(remotes) = settings.get::<Vec<String>>(KEY) {
+        remotes
+            .into_iter()
+            .map(|r| parse_remote_pattern(&r))
+            .collect::<Result<Vec<_>, _>>()
```
yuja, line 151: nit: `.try_collect()` also works.

cli/tests/test_git_fetch.rs
```diff
@@ -297,6 +297,168 @@ fn test_git_fetch_multiple_remotes(subprocess: bool) {
...
+    test_env.add_config("git.auto-local-bookmark = true");
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let repo_path = test_env.env_root().join("repo");
+    add_git_remote(&test_env, &repo_path, "rem1");
+    add_git_remote(&test_env, &repo_path, "rem2");
+    add_git_remote(&test_env, &repo_path, "upstream");
+
+    test_env
+        .run_jj_in(&repo_path, ["git", "fetch", "--remote", "glob:rem*"])
+        .success();
```
yuja, line 316: Maybe we can instead snapshot the `git fetch` output? It's not so important to see how remote changes were merged to local bookmarks.

(Also applies to the other tests)

yuja (approved):
Thanks.

cli/tests/test_git_fetch.rs
unexge: Updated to snapshot `git fetch`'s output

cli/tests/test_git_fetch.rs
```diff
@@ -297,6 +297,168 @@ fn test_git_fetch_multiple_remotes(subprocess: bool) {
...
+    insta::assert_snapshot!(get_bookmark_output(&test_env, &repo_path), @r"
+    rem1: qxosxrvv 6a211027 message
+      @rem1: qxosxrvv 6a211027 message
+    rem2: yszkquru 2497a8a0 message
+      @rem2: yszkquru 2497a8a0 message
+    [EOF]
+    ");
+    }
+}
+
```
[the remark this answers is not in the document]
unexge: Added test case for no matching git remotes and reduced number of test cases

cli/src/commands/git/fetch.rs
```diff
@@ -72,13 +84,45 @@ pub fn cmd_git_fetch(
...
+        }
+        patterns => {
+            return Err(user_error(format!(
+                "No matching git remotes for patterns: {}",
+                patterns.iter().join(", ")
+            )))
+        }
+    }
+
+    let remotes = remotes.into_iter().map(|s| s.as_str()).collect_vec();
```
yuja, line 124: The order is not deterministic (though the stats output is sorted.) I think we can use `all_remotes.retain(|r| remotes.contains(r))` or something.

cli/tests/test_git_fetch.rs
```diff
@@ -297,6 +297,114 @@ fn test_git_fetch_multiple_remotes(subprocess: bool) {
...
+        test_env.add_config("git.subprocess = false");
+    }
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let repo_path = test_env.env_root().join("repo");
+    add_git_remote(&test_env, &repo_path, "rem1");
+    add_git_remote(&test_env, &repo_path, "rem2");
+
+    let output = test_env
+        .run_jj_in(&repo_path, ["git", "fetch", "--remote", "glob:*"])
+        .success();
```
yuja, line 314: nit: `.success()` can be omitted when the output is snapshotted. (applies to the other tests.)

yuja (approved):

cli/tests/test_git_fetch.rs
unexge: Dropped `.success()` call from snapshotted outputs

PR #5819, opened 2025-02-27
docs workflow: try to fix it again

ilyagr:
(See also the commit description)

Some notes:

  - I'm not sure whether the env var should be `GH_PAGES` or `GITHUB_PAGES`, but several places in the docs suggested using the former. We can try this as is, or we can set both.
  - It bothers me that there is no easy way I can think of to test these things after 514a009, or to experiment with it outside of production. WDYT, @thoughtpolice ?

Sample error: https://github.com/jj-vcs/jj/actions/runs/13548628896/job/37866477622

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -36,5 +36,7 @@ jobs:
           export MKDOCS_SITE_NAME="Jujutsu docs (prerelease)"
           export MKDOCS_PRIMARY_COLOR="blue grey"
           .github/scripts/docs-build-deploy prerelease --push
+        env:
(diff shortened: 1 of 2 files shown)


ilyagr:
Cc @neongreen if you have any ideas for how to deal with this better. I'll also start a Discord thread.

martinvonz (approved):
Thanks!

ilyagr:
FYI, this is how the `checkout` action sets up `git` auth and (currently, with `persist-credentials:false`) removes it:

image

You need to unfold the appropriate triangles in <https://github.com/jj-vcs/jj/actions/runs/13559508098/job/37899985506> to see this.

PR #5820, opened 2025-02-27
docs workflow: try to fix it a third time

ilyagr:
See commit description for more.

cc @thoughtpolice @neongreen @martinvonz

See https://github.com/jj-vcs/jj/actions/runs/13559508098/job/37899985506 . The `GH_PAGES` variable is populated, but we get the same error.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,6 +9,8 @@ permissions: {}
 
 jobs:
   prerelease-docs-build-deploy:
+    # IMPORTANT: this workflow also functions as a test for `docs-deploy-website-latest-release` in
+    # releases.yml. Any fixes here should probably be duplicated there.
     permissions:
       contents: write
     if: github.repository_owner == 'jj-vcs' # Stops this job from running on forks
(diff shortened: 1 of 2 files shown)


ilyagr:
I think I figured out what the problem actually is. It's `persist-credetials: false` in the workflow in the checkout action. I'm not sure what the correct way to get around that is except for setting `persist-credentials: true`.

I won't merge this until I research it a bit more. For example, will fixing `persist-credentials` upset zizmor?

ilyagr:
Yes, I think this is a proper fix now. I think the security implications are OK, but if there's doubt, we don't have to rush to merge this. (Though we should figure something out a few days before the upcoming release in a week)

ilyagr:
FYI, this is how the `checkout` action sets up `git` auth and (currently, with `persist-credentials:false`) removes it:

image

You need to unfold the appropriate triangles in <https://github.com/jj-vcs/jj/actions/runs/13559508098/job/37899985506> to see this.

martinvonz (approved):
Sure, let's try it

ilyagr:
I added one more comment here. I'll probably still start the merge process soon, but let me know if you don't like it. The comment is:

```
# IMPORTANT: this workflow also functions as a test for `docs-deploy-website-latest-release` in
# releases.yml. Any fixes here should probably be duplicated there.
```

ilyagr:
This worked! 🎉 

https://github.com/jj-vcs/jj/commit/d6a54b78994123065fcb672480caecb2c8051ad9

https://github.com/jj-vcs/jj/actions/runs/13560546877/job/37902708179

PR #5823, opened 2025-02-27
Add `--key` argument to `jj sign`

### LINKED ISSUE #5804: Make `jj sign` accept `--key` argument
In the initial implementation of `jj sign` (#4747) we intentionally left out the `--key` argument (see https://github.com/jj-vcs/jj/pull/4747#discussion_r1968720554).

pylbrecht:
Closes #5804

# Checklist

If applicable:
- [ ] ~~I have updated `CHANGELOG.md`~~

### DIFF
--- a/cli/src/commands/sign.rs
+++ b/cli/src/commands/sign.rs
@@ -46,6 +46,10 @@ pub struct SignArgs {
         add = ArgValueCandidates::new(complete::mutable_revisions),
     )]
     revisions: Vec<RevisionArg>,
+
+    /// The key used for signing
+    #[arg(long)]
+    key: Option<String>,
 }
 
 pub fn cmd_sign(ui: &mut Ui, command: &CommandHelper, args: &SignArgs) -> Result<(), CommandError> {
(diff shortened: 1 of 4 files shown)


pylbrecht:
Do I need to update the changelog?
I am asking, because I don't know how fine-grained we want to be here. We already have an entry for `jj sign`, which I feel would already cover this.

cli/src/commands/sign.rs
```diff
@@ -43,6 +43,10 @@ pub struct SignArgs {
         add = ArgValueCandidates::new(complete::mutable_revisions),
     )]
     revisions: Vec<RevisionArg>,
+
+    /// The key used for signing
+    #[arg(long, value_name = "KEY")]
```
yuja, line 48: nit: `value_name` can be omitted

yuja (approved):
Thanks.

ilyagr:
This is very much optional, but it would be nice to have some more docs on how jj signs commits and what kind of key is needed would be nice. For example, we could mention that GPG needs to be installed (assuming that's true; my knowledge mostly ends at the fact that the tests require gpg).

Apparently, it is also possible to sign commits with ssh keys, <https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits>. I'm guessing we don't support that? 

Sorry if the docs exist and I just missed them.

It's also OK to leave this for others to write down later. E.g., I might write something down if I set up signing for myself at some point and figure it out in the process.

ilyagr:
I found https://jj-vcs.github.io/jj/prerelease/config/#commit-signing . I think one of us should remember to add a link there from `jj help (un)sign` (unless it's already there and I missed it too), that should about cover it. Sorry!

pylbrecht:
> I found https://jj-vcs.github.io/jj/prerelease/config/#commit-signing . I think one of us should remember to add a link there from `jj help (un)sign` (unless it's already there and I missed it too), that should about cover it. Sorry I initially missed it.

I will take care of that.

pylbrecht:
> It might not be a priority, but we might also want to allow customizing the signing backend along with the key.

Good point, I created https://github.com/jj-vcs/jj/issues/5835 to track that.

Issue #5748, opened 2025-02-19
Setting a name with `jj config set user.name` with odd characters results in an unfriendly parser error


ephemeralctl:
## Description

Upon following the instructions in my terminal to set a name with `jj config set --user user.name`, I was presented with a very unusual error:

```
$ jj config set --user user.name "Micheal D'Angelo"
error: invalid value 'Micheal D'Angelo' for '<VALUE>': TOML parse error at line 1, column 1
  |
1 | Micheal D'Angelo
  | ^
invalid string
expected `"`, `'`

For more information, try '--help'.
```

## Steps to Reproduce the Problem

Here is the full flow that I, as an end user, experienced to get to this point:

```
$ jj git init
Error: Did not create a jj repo because there is an existing Git repo in this directory.
Hint: To create a repo backed by the existing Git repo, run `jj git init --colocate` instead.

$ jj git init --colocate
Done importing changes from the underlying Git repo.
Setting the revset alias `trunk()` to `prod@origin`
Hint: The following remote bookmarks aren't associated with the existing local bookmarks:
  prod@origin
  some-branch@origin
Hint: Run `jj bookmark track prod@origin some-branch@origin` to keep local bookmarks updated on future pulls.                                                                                         Initialized repo in "."

$ jj bookmark track prod@origin some-branch@origin
Started tracking 2 remote bookmarks.
Warning: Name and email not configured. Until configured, your commits will be created with the empty identity, and can't be pushed to remotes. To configure, run:
  jj config set --user user.name "Some One"
  jj config set --user user.email "someone@example.com"
```

## Specifications

- Version: jj 0.26.0-613742dfbbd89324b25672a75ef8ce9e671ae0d3
- #5233, #5231 seem like similar issues

---

Perhaps for configuration fields of a known type (such as a name which we know is a string), these should be interpreted as a string rather than as a TOML table?

joyously:
That's a bit crazy...
it sounds similar to #5233 Setting a user.name is confusing with `jj config set` if the name contains a comma
...which is closed.

yuja:
The doc is updated by 9cab4d9a0be1b19f66637ec89a77870d0b5301e5.

Maybe we can improve the error message a bit. 

I don't think it's good idea to make `config set` type-aware only for known keys. (It might be good if unknown keys are all rejected, but that's not feasible.)

ilyagr:
I may have brought this up elsewhere, but we could also allow

```
[...]

PR #5827, opened 2025-02-27
Rename the `LocalBackend` to `ToyBackend` and move it the examples

PhilipMetzger:
This implements my proposal from [Discord]. 

# Checklist

If applicable:

[Discord]: https://discord.com/channels/968932220549103686/969291218347524238/1334610927256997991

### DIFF
--- a/cli/tests/test_init_command.rs
+++ b/cli/tests/test_init_command.rs
@@ -15,29 +15,34 @@
 use crate::common::TestEnvironment;
 
 #[test]
-fn test_init_local_disallowed() {
+fn test_init_toy_disallowed() {
     let test_env = TestEnvironment::default();
     let output = test_env.run_jj_in(".", ["init", "repo"]);
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
-    Error: The native backend is disallowed by default.
+    Warning: --ui.allow-init-native is deprecated; use ui.allow-init-toy instead.
+    Error: The toy backend is disallowed by default.
     Hint: Did you mean to call `jj git init`?
-    Set `ui.allow-init-native` to allow initializing a repo with the native backend.
+    Set `ui.allow-init-toy` to allow initializing a repo with the toy backend.
     [EOF]
     [exit status: 1]
     ");
 }
 
 #[test]
-fn test_init_local() {
+fn test_init_toy() {
     let test_env = TestEnvironment::default();
-    test_env.add_config(r#"ui.allow-init-native = true"#);
+    test_env.add_config(r#"ui.allow-init-toy = true"#);
     let output = test_env.run_jj_in(".", ["init", "repo"]);
-    insta::assert_snapshot!(output, @r#"
+    insta::assert_snapshot!(output, @r"
     ------- stderr -------
-    Initialized repo in "repo"
+    Warning: --ui.allow-init-native is deprecated; use ui.allow-init-toy instead.
+    Error: The toy backend is disallowed by default.
+    Hint: Did you mean to call `jj git init`?
+    Set `ui.allow-init-toy` to allow initializing a repo with the toy backend.
     [EOF]
-    "#);
+    [exit status: 1]
+    ");
 
     let workspace_root = test_env.env_root().join("repo");
     let jj_path = workspace_root.join(".jj");
@@ -68,5 +73,20 @@ fn test_init_local() {
     Error: --at-op is not respected
     [EOF]
     [exit status: 2]
+    ");
+}
+
+// TODO: remove in jj 0.33+
+#[test]
+fn test_init_native_deprecation_warning() {
+    let test_env = TestEnvironment::default();
+    test_env.add_config(r#"ui.allow-init-native = true"#);
+    let output = test_env.run_jj_in(".", ["init", "repo"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Config error: Value not found for ui.allow-init-toy
+    For help, see https://jj-vcs.github.io/jj/latest/config/.
+    [EOF]
+    [exit status: 1]
     ");
 }
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -105,6 +105,8 @@ pub mod submodule_store;
 #[cfg(feature = "testing")]
 pub mod test_signing_backend;
 pub mod time_util;
+// Allow this file to be outside the normal crate.
+#[path = "../../cli/examples/toy-backend/toy_backend.rs"]
 pub mod toy_backend;
 pub mod transaction;
 pub mod tree;
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -73,7 +73,6 @@ pub mod graph;
 pub mod hex_util;
 pub mod id_prefix;
 pub mod index;
-pub mod local_backend;
 pub mod local_working_copy;
 pub mod lock;
 pub mod matchers;
@@ -106,6 +105,7 @@ pub mod submodule_store;
 #[cfg(feature = "testing")]
 pub mod test_signing_backend;
 pub mod time_util;
+pub mod toy_backend;
 pub mod transaction;
 pub mod tree;
 pub mod tree_builder;
--- a/lib/src/workspace.rs
+++ b/lib/src/workspace.rs
@@ -30,7 +30,6 @@ use crate::backend::MergedTreeId;
 use crate::commit::Commit;
 use crate::file_util::IoResultExt as _;
 use crate::file_util::PathError;
-use crate::local_backend::LocalBackend;
 use crate::local_working_copy::LocalWorkingCopy;
 use crate::local_working_copy::LocalWorkingCopyFactory;
 use crate::op_heads_store::OpHeadsStoreError;
@@ -53,6 +52,7 @@ use crate::settings::UserSettings;
 use crate::signing::SignInitError;
 use crate::signing::Signer;
 use crate::store::Store;
+use crate::toy_backend::ToyBackend;
 use crate::working_copy::CheckoutError;
 use crate::working_copy::CheckoutOptions;
 use crate::working_copy::CheckoutStats;
@@ -185,7 +185,7 @@ impl Workspace {
         workspace_root: &Path,
     ) -> Result<(Self, Arc<ReadonlyRepo>), WorkspaceInitError> {
         let backend_initializer: &BackendInitializer =
-            &|_settings, store_path| Ok(Box::new(LocalBackend::init(store_path)));
+            &|_settings, store_path| Ok(Box::new(ToyBackend::init(store_path)));
         let signer = Signer::from_settings(user_settings)?;
         Self::init_with_backend(user_settings, workspace_root, backend_initializer, signer)
     }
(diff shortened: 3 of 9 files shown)


cli/tests/test_init_command.rs
```diff
@@ -70,3 +75,18 @@ fn test_init_local() {
...
+    test_env.add_config(r#"ui.allow-init-native = true"#);
+    let output = test_env.run_jj_in(".", ["init", "repo"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Config error: Value not found for ui.allow-init-toy
+    For help, see https://jj-vcs.github.io/jj/latest/config/.
+    [EOF]
+    [exit status: 1]
+    ");
+}
```
PhilipMetzger, line 92: I failed to only get the deprecation message here, so any thoughts on how to achieve it?

yuja:
> We recently had a user try out the "native" backend for some testing and
> they didn't understand that you currently cannot migrate it to Git, so
> make it obvious that it isn't intended for "production" use.

Maybe we can instead rename `jj init` to `jj debug init-native` or something? It will free up `jj init` and user can set alias if they like.

I don't think we need deprecation for that because the native backend shouldn't ever be used.

martinvonz:
I sent #5836 as alternative

PhilipMetzger:
> > We recently had a user try out the "native" backend for some testing and
> > they didn't understand that you currently cannot migrate it to Git, so
> > make it obvious that it isn't intended for "production" use.
> 
> Maybe we can instead rename `jj init` to `jj debug init-native` or something? It will free up `jj init` and user can set alias if they like.

There's also the option of making `jj init` at some point interactive which would also help to partially solve #5748. IMO, it's better if we just hide the command until a truly native backend exists. 

> I don't think we need deprecation for that because the native backend shouldn't ever be used.

My initial approach was the same, so I'll revert back to it.

steveklabnik:
> Maybe we can instead rename jj init to jj debug init-native or something? It will free up jj init and user can set alias if they like.

I will say that I was messing around with the local/toy backend recently, and forgot I had set the config to allow me to make them. I then was trying to help someone in Discord, and typed `jj init` instead of `jj git init` and didn't realize i had accidentally made a backend of the wrong type. So breaking that symmetry a bit might be nice.

PhilipMetzger:
I've dropped the first commit, since Ilya made it obsolete.

yuja:
I personally don't think we need to rename the type to clarify the implementation status. Just adding a few doc comments should be good enough?

PhilipMetzger:
> I personally don't think we need to rename the type to clarify the implementation status.

Ilya made a good point in his PR:  "In my mind toy-backend is a shorter/clearer version of poc-backend or proof-of-concept-backend.", which is something I heavily agree with and _why_ I want to go through with the rename. 

> Just adding a few doc comments should be good enough?

While it improves the situation, I don't think it is a good way to tell everyone to ignore the backend as is.

martinvonz:
I was thinking that the backend would evolve into something that's production-ready. Maybe that's unrealistic. I think we should just delete it in that case.

ilyagr:
TLDR: I feel neutral about this; I'm OK with renaming or with just with adding some doc comments. This comment is very much in the category of "ignore it if it's not helpful".

In principle, I agree with Philip. I'd call this the "toy backend", and if later a fairy comes and turns the toy into a real ~~boy~~ backend, we could rename it.

OTOH, now that the backend is only accessible via `jj debug` (thank you, Yuya, for the suggestion!), I don't think its name really matters.

> Just adding a few doc comments should be good enough?

I did add a doc comment here:

https://github.com/jj-vcs/jj/blob/228385f7f439608969f112f62336e37da2624697/cli/src/commands/debug/init_local.rs#L30-L31

I think the most likely way a user would discover this backend is via our tests, so it might make sense to add something similar to the doc comments of the LocalBackend-related objects/functions that are used in tests.

PhilipMetzger:
> I was thinking that the backend would evolve into something that's production-ready. Maybe that's unrealistic. I think we should just delete it in that case.

I don't particularly think that the LocalBackend will evolve into that, but also think its worthwhile to keep it around in the `examples` folder since it gives people a base for their custom backends.

martinvonz:
I don't like calling it a "toy". How about `PocBackend` then (for proof of concept)? Or `SimpleBackend` to match `SimpleOpStore` and `SimpleOpHeadsStore`?

emilazy:
`DemoBackend`? `ExampleBackend`?

PhilipMetzger:
> I don't like calling it a "toy". How about `PocBackend` then (for proof of concept)? Or `SimpleBackend` to match `SimpleOpStore` and `SimpleOpHeadsStore`?

Renamed to SimpleBackend, is that good enough.

lib/src/workspace.rs
```diff
@@ -185,7 +185,7 @@ impl Workspace {
         workspace_root: &Path,
     ) -> Result<(Self, Arc<ReadonlyRepo>), WorkspaceInitError> {
         let backend_initializer: &BackendInitializer =
-            &|_settings, store_path| Ok(Box::new(LocalBackend::init(store_path)));
+            &|_settings, store_path| Ok(Box::new(SimpleBackend::init(store_path)));
```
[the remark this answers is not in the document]
PhilipMetzger: done

lib/testutils/src/lib.rs
```diff
@@ -195,7 +195,7 @@ impl TestRepoBackend {
     ) -> Result<Box<dyn Backend>, BackendInitError> {
         match self {
             TestRepoBackend::Git => Ok(Box::new(GitBackend::init_internal(settings, store_path)?)),
-            TestRepoBackend::Local => Ok(Box::new(LocalBackend::init(store_path))),
+            TestRepoBackend::Local => Ok(Box::new(SimpleBackend::init(store_path))),
```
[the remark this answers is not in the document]
PhilipMetzger: done

cli/examples/simple-backend/simple_backend.rs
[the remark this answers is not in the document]
PhilipMetzger: > Oh, speaking of that, we may be able to include it only when the `testing` crate feature is enabled.

Gated behind `#[cfg(feature = "testing")]`. I left the file still in the examples though.

lib/src/simple_backend.rs
```diff
@@ -554,7 +555,7 @@ mod tests {
         let temp_dir = new_temp_dir();
         let store_path = temp_dir.path();
 
-        let backend = LocalBackend::init(store_path);
+        let backend = PocBackend::init(store_path);
```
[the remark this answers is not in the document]
PhilipMetzger: done

lib/src/simple_backend.rs
```diff
@@ -352,7 +352,8 @@ impl Backend for LocalBackend {
     }
 }
 
-#[expect(clippy::assigning_clones)]
+#[allow(unknown_lints)] // XXX FIXME (aseipp): nightly bogons; re-test this occasionally
+#[allow(clippy::assigning_clones)]
```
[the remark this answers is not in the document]
PhilipMetzger: done

cli/tests/test_root.rs
```diff
@@ -20,7 +20,7 @@ use testutils::TestWorkspace;
 
 use crate::common::TestEnvironment;
 
-#[test_case(TestRepoBackend::Local ; "local backend")]
+#[test_case(TestRepoBackend::Simple ; "local backend")]
```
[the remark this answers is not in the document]
PhilipMetzger: done

cli/src/commands/debug/init_local.rs
```diff
@@ -34,17 +34,17 @@ use crate::ui::Ui;
 /// does not exist, it will be created. If no directory is given, the current
 /// directory is used.
 #[derive(clap::Args, Clone, Debug)]
-pub(crate) struct DebugInitLocalArgs {
+pub(crate) struct DebugInitSimpleArgs {
```
martinvonz, line 37: nit: update the rustdoc comment too. i'm fine with just deleting the line about why it's called "local"

cli/src/commands/debug/init_local.rs
martinvonz, line 1: rename the file?

cli/src/commands/debug/mod.rs
```diff
@@ -70,7 +70,7 @@ pub enum DebugCommand {
     CopyDetection(CopyDetectionArgs),
     Fileset(DebugFilesetArgs),
     Index(DebugIndexArgs),
-    InitLocal(DebugInitLocalArgs),
+    InitLocal(DebugInitSimpleArgs),
```
martinvonz, line 73: nit: rename the variant too

lib/src/simple_backend.rs
```diff
@@ -352,7 +352,7 @@ impl Backend for LocalBackend {
     }
 }
 
-#[expect(clippy::assigning_clones)]
+#[allow(clippy::assigning_clones)]
```
martinvonz: rename `local_store.proto` too?

cli/examples/simple-backend/simple_backend.rs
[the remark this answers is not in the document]
PhilipMetzger: I removed the `#[cfg(feature = "testing")]` again, as it fails the Git-less build.

lib/src/simple_backend.rs
PhilipMetzger: done

cli/src/commands/debug/mod.rs
PhilipMetzger: done

cli/src/commands/debug/init_local.rs
PhilipMetzger: done

cli/src/commands/debug/init_local.rs
PhilipMetzger: done

cli/examples/simple-backend/simple_backend.rs
martinvonz, line 1: Move to first commit

martinvonz, line 1: Also update `architecture.md`

martinvonz: There's also a reference in `test_backend.rs` (I used the "Find in Files" feature in VS Code for finding these, btw)

PhilipMetzger: done

PhilipMetzger: done

martinvonz (approved):
LGTM. Are you also okay with this name, @yuja?

lib/src/lib.rs
```diff
@@ -96,6 +95,10 @@ pub mod rewrite;
 pub mod secret_backend;
 pub mod settings;
 pub mod signing;
+// Allow this file to be outside the normal crate, as it is only used for
+// testing.
+#[path = "../../cli/examples/simple-backend/simple_backend.rs"]
+pub mod simple_backend;
```
yuja, line 101: This looks ugly (and might break `cargo publish`-ed build?)

Maybe we shouldn't move the underlying file until we can remove all `SimpleBackend` dependencies from jj-lib?

yuja (commented):
> LGTM. Are you also okay with this name, @yuja?

"Simple" sounds good to me.

lib/src/lib.rs
martinvonz: I agree. Thanks for catching that. This was originally part of the second commit and when the second commit was removed, I didn't realize that parts of it had moved into the first commit. Maybe it was just a mistake.

PhilipMetzger: Removed and replaced with a TODO which mentions that we should do that.

PR #5828, opened 2025-02-28
Disable the new bookmark behavior for `jj split`

emesterhazy:
The consensus about this change, if one ever existed, seems to have dissolved into two cohorts, one which is happy with the change, and one which thinks we should move both the bookmark and change id to the child commit instead of leaving them on the parent.

We may decide to add flags to allow users to choose between the two behaviors, but there are also other concerns such as where @ should go (parent or child). Until we agree on a path forward it seems reasonable to delay the breaking change by disabling it via the config option we added. I don't think it's necessary to fully revert the code and new tests since we aren't announcing the option.

#3419

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,13 +33,6 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   of forgetting them, since forgetting a remote bookmark can be unintuitive.
   The old behavior is still available with the new `--include-remotes` flag.
 
-* `jj split` no longer moves bookmarks to the second revision created by the
-  split. Instead, local bookmarks associated with the target revision will move
-  to the first revision created by the split (which inherits the target
-  revision's change id). You can opt out of this change by setting
-  `split.legacy-bookmark-behavior = true`, but this will likely be removed in a
-  future release. [#3419](https://github.com/jj-vcs/jj/issues/3419)
-
 * `jj fix` now always sets the working directory of invoked tools to be the
   workspace root, instead of the working directory of the `jj fix`.
 
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -196,7 +196,7 @@ The remainder will be in the second commit.
         commit_builder.write(tx.repo_mut())?
     };
 
-    let legacy_bookmark_behavior = read_legacy_bookmark_behavior_setting(tx.settings(), ui)?;
+    let legacy_bookmark_behavior = tx.settings().get_bool("split.legacy-bookmark-behavior")?;
     if legacy_bookmark_behavior {
         // Mark the commit being split as rewritten to the second commit. This
         // moves any bookmarks pointing to the target commit to the second
@@ -240,29 +240,4 @@ The remainder will be in the second commit.
     }
     tx.finish(ui, format!("split commit {}", commit.id().hex()))?;
     Ok(())
-}
-
-/// Reads and returns 'split.legacy-bookmark-behavior' from settings and prints
-/// a warning if it is set to true to alert the user of the deprecation.
-fn read_legacy_bookmark_behavior_setting(
-    settings: &jj_lib::settings::UserSettings,
-    ui: &Ui,
-) -> Result<bool, CommandError> {
-    let use_legacy_behavior = settings.get_bool("split.legacy-bookmark-behavior")?;
-    if use_legacy_behavior {
-        writeln!(
-            ui.warning_default(),
-            "`jj split` will leave bookmarks on the first commit in the next release."
-        )?;
-        writeln!(
-            ui.warning_default(),
-            "Run `jj config set --user split.legacy-bookmark-behavior false` to silence this \
-             message and use the new behavior."
-        )?;
-        writeln!(
-            ui.warning_default(),
-            "See https://github.com/jj-vcs/jj/issues/3419"
-        )?;
-    }
-    Ok(use_legacy_behavior)
 }
--- a/cli/src/config/misc.toml
+++ b/cli/src/config/misc.toml
@@ -49,5 +49,7 @@ auto-track = "all()"
 auto-update-stale = false
 
 # TODO: https://github.com/jj-vcs/jj/issues/3419 - Remove when fully deprecated.
+# The behavior when this flag is set to false is experimental and may be changed
+# in the future.
 [split]
-legacy-bookmark-behavior = false
+legacy-bookmark-behavior = true
--- a/cli/tests/test_split_command.rs
+++ b/cli/tests/test_split_command.rs
@@ -881,24 +881,24 @@ fn test_split_with_multiple_workspaces_different_working_copy() {
 
 enum BookmarkBehavior {
     Default,
-    Legacy,
-    Modern,
+    MoveBookmarkToChild,
+    LeaveBookmarkWithTarget,
 }
 
 // TODO: https://github.com/jj-vcs/jj/issues/3419 - Delete params when the config is removed.
 #[test_case(BookmarkBehavior::Default; "default_behavior")]
-#[test_case(BookmarkBehavior::Legacy; "legacy_behavior")]
-#[test_case(BookmarkBehavior::Modern; "modern_behavior")]
+#[test_case(BookmarkBehavior::MoveBookmarkToChild; "move_bookmark_to_child")]
+#[test_case(BookmarkBehavior::LeaveBookmarkWithTarget; "leave_bookmark_with_target")]
 fn test_split_with_bookmarks(bookmark_behavior: BookmarkBehavior) {
     let mut test_env = TestEnvironment::default();
     test_env.run_jj_in(".", ["git", "init", "main"]).success();
     let main_path = test_env.env_root().join("main");
 
     match bookmark_behavior {
-        BookmarkBehavior::Modern => {
+        BookmarkBehavior::LeaveBookmarkWithTarget => {
             test_env.add_config("split.legacy-bookmark-behavior=false");
         }
-        BookmarkBehavior::Legacy => {
+        BookmarkBehavior::MoveBookmarkToChild => {
             test_env.add_config("split.legacy-bookmark-behavior=true");
         }
         BookmarkBehavior::Default => (),
@@ -929,7 +929,7 @@ fn test_split_with_bookmarks(bookmark_behavior: BookmarkBehavior) {
     .unwrap();
     let output = test_env.run_jj_in(&main_path, ["split", "file2"]);
     match bookmark_behavior {
-        BookmarkBehavior::Default | BookmarkBehavior::Modern => {
+        BookmarkBehavior::LeaveBookmarkWithTarget => {
             insta::allow_duplicates! {
             insta::assert_snapshot!(output, @r"
             ------- stderr -------
@@ -949,19 +949,16 @@ fn test_split_with_bookmarks(bookmark_behavior: BookmarkBehavior) {
             ");
             }
         }
-        BookmarkBehavior::Legacy => {
+        BookmarkBehavior::Default | BookmarkBehavior::MoveBookmarkToChild => {
             insta::allow_duplicates! {
-            insta::assert_snapshot!(output, @r"
+            insta::assert_snapshot!(output, @r###"
             ------- stderr -------
-            Warning: `jj split` will leave bookmarks on the first commit in the next release.
-            Warning: Run `jj config set --user split.legacy-bookmark-behavior false` to silence this message and use the new behavior.
-            Warning: See https://github.com/jj-vcs/jj/issues/3419
             First part: qpvuntsm 63d0c5ed first-commit
             Second part: mzvwutvl a9f5665f *le-signet* | second-commit
             Working copy now at: mzvwutvl a9f5665f *le-signet* | second-commit
             Parent commit      : qpvuntsm 63d0c5ed first-commit
             [EOF]
-            ");
+            "###);
             }
             insta::allow_duplicates! {
             insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r"
@@ -985,7 +982,7 @@ fn test_split_with_bookmarks(bookmark_behavior: BookmarkBehavior) {
         .run_jj_in(&main_path, ["split", "file2", "--parallel"])
         .success();
     match bookmark_behavior {
-        BookmarkBehavior::Default | BookmarkBehavior::Modern => {
+        BookmarkBehavior::LeaveBookmarkWithTarget => {
             insta::allow_duplicates! {
             insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r"
             @  vruxwmqvtpmx false second-commit
@@ -996,7 +993,7 @@ fn test_split_with_bookmarks(bookmark_behavior: BookmarkBehavior) {
             ");
             }
         }
-        BookmarkBehavior::Legacy => {
+        BookmarkBehavior::Default | BookmarkBehavior::MoveBookmarkToChild => {
             insta::allow_duplicates! {
             insta::assert_snapshot!(get_log_output(&test_env, &main_path), @r"
             @  vruxwmqvtpmx false *le-signet* second-commit


yuja (approved):
I don't follow the discussion closely, but it should be good to revert the change for the next release. I think the decision should be made within a couple of releases (because it would be weird if `legacy` option persisted forever.)

emesterhazy:
> I don't follow the discussion closely, but it should be good to revert the change for the next release. I think the decision should be made within a couple of releases (because it would be weird if `legacy` option persisted forever.)

Agreed... I'd like for this to be resolved as soon as possible.

PhilipMetzger:
Can I get a summary since I definitely lost track of the conversation, _why_ it should get reverted again?

emesterhazy:
> Can I get a summary since I definitely lost track of the conversation, _why_ it should get reverted again?

I'm not great with Discord, but if I understand correctly (sorry for glossing over the nuances) some people, particularly @emilazy, @ilyagr and @scott2000 now think that split should pull changes out of the target commit and into a new parent instead of pulling them into a new child, which is the behavior at head and the latest stable release. This means that the bookmark and change id would move to the second commit instead of the first. I think the sticking point is that they think this should be the default and maybe only (gasp) behavior.

There's also been discussion about whether we should add `-B, -A, and -d` flags to support both and even more flexible behaviors. It's all sort of a mess and it doesn't seem like theres any real consensus about how split should behave, other than that the current behavior where the bookmark moves off the target change id is bad.

So why don't we just fix the bad bookmark behavior for now? If we agree that split will support a mode where the change id and bookmark stay with the target commit, then I'm not really sure that there's a strong argument for blocking this change other than that it will be inconvenient for people who want the bookmark to move to the second commit, and bookmark movement is a whole can of worms. But if we don't accept that, and think that the change id and bookmark will only move to the second commit in the future, then this is a step in the wrong direction.

emilazy:
Summary of the Discord discussion, necessarily coloured by my POV but I will try to weigh all the different positions expressed:

1. We’ve had a hard time communicating about all of this and a lot of the agreement on the previous PR involved being confused about what the actual state after it would be or not having fully thought things through.

2. Everyone agrees that the behaviour as of the current stable release is bad and we need to fix it. I don’t think anyone likes the status quo.

3. The current state of trunk is that the change ID and bookmarks go to the *first* commit in a split, but `@` moves to the second. @ilyagr, @scott2000, and I all believe that this is a problematic inconsistency: for the same reason that bookmarks should move with the change ID, so should the working copy. My understanding of @emesterhazy’s position is that he doesn’t necessarily agree, but is open to discussion about it. I consider this inconsistency bad enough that it’s a signal we need to work out the actual desired semantics rather than going forward with a breaking change we have conflicting intuitions about and lack a shared coherent model to justify.

4. The behaviour at trunk – ignoring the `@` issue – is well‐suited to a workflow where you want to *evict* some changes from a change; say if you got a review saying that some extraneous change on your CL/PR should be split out for later review. In that case, you want the change ID and bookmarks to go with the commit that lost the split‐out diff hunks.

5. There is another legitimate and common workflow: splitting *preliminary work* out of a change: say, moving a simple preparatory refactor or rename as a new parent for a commit that implements a new feature, for the new feature to be stacked on top of. In this case, you want the change ID and bookmarks to go to the second commit. I believe that @ilyagr, @scott2000, and I all personally end up doing these kinds of splits more than (4). The behaviour of the stable release is closer to this workflow (modulo our unanimous agreement that the change ID should certainly go to the second commit in this case).

6. I understand @emesterhazy usually does (4). @ilyagr, @scott2000, and I all think that for both of these workflows, it makes more sense for `@` to move with the change ID. I believe @emesterhazy is not as convinced. 

7. There are “Git compatibility” workflows that dovetail better with the stable release behaviour of (5): “`jj split` as `git commit -p`”, adding new commits while sliding a pull request bookmark before, etc.; none of the people who consider (5) a legitimate and important workflow do so because of these compatibility concerns. It’s not a matter of constraining Jujutsu’s design based on GitHub; all the people who value (5) do so invariant of such concerns. In fact, Gerrit‐style change‐based code review where change IDs are semantically meaningful and reviews are attached to them helps motivate our position on how the change ID should move for (5). People who use `jj split` for these compatibility properties still must be taken into account when considering compatibility shims and breaking changes, but none of the motivations here are *about* that.

8. One proposal was to make (5) the default behaviour (so `jj split` would behave like the stable release except the change ID would move to the second commit for consistency), and add a flag like `--evict`/`--away`/`--yeet`/… that implements (4). I feel that we were moving in the direction of something vaguely like consensus on this, but the default mode was still a point of contention, and we had a jumble of confusing and conflicting intuitions about how `@` should move and how descendants should be rebased. We were also uncertain about whether you should be selecting changes to keep or changes to split out (i.e., which side goes where).

9. As it seemed like we will not be able to get a consensus on the exact behaviour `jj split` should have by the release, and there are significant concerns about the behaviour in trunk that we did not fully realize when the original PR was merged, I think we all uneasily agreed that it’s sensible to roll back the default behaviour for now to take the time pressure of the upcoming release off. Restoring the default behaviour of the previous stable release is a conservative change that avoids committing to a breaking change that we don’t seem to have full consensus for. It’s not an endorsement of the stable behaviour, though, which again everyone agrees is bad.

10. At that point, I brought up the idea that had come up before of giving `jj split` arguments for `-d`/`-B`/`-A`, a la `new`/`rebase`/`duplicate`. Some subsequent discussion revealed that this seems like a promising avenue to resolve these confusions and conflicts:

    1. It provides a simple model for answering questions about change IDs, bookmarks, `@`, and descendants: the `-r` commit keeps its change ID and bookmarks in all cases, just like those other commands; `@` remains there if it was already there; descendants are rebased according to what destination option you use. With these destination flags, no other behaviour really makes sense.

    2. `-r X -B X` supports the (5) workflow, and is almost identical to the stable behaviour modulo change IDs.

    3. `-r X -A X` supports the (4) workflow. In this case, you select which changes would be split out into the parent, which differs from the behaviour at trunk, but is more consistent with the model implied by these flags (and personally I think this is usually how you would expect to select hunks given this interface). Unlike the `--evict` we were discussing, this also has reasonable behaviour with the squash workflow: `-r @ -A @` leaves the working copy without the split‐out changes, while `-r @- -A 'next()'` (where `next()` is @algmyr’s name for `@-+ ~ @`, the commit `jj next` would move to) achieves the same for a commit you are on top of but not directly editing.

    4. `-r X -d X-` supports a workflow we were calling `--evict-new`, where you want to split out some work from a stack entirely, and don’t want it to be incorporated into the chain sof descendants at all. This also helped resolved some of our confusion about descendant rebasing.

    5. It also unlocks some nice new workflows: `jj split --after 'trunk()' --before @-` can add a new commit as a parent to a mega‐merge, `jj split -r X --before Y` can slot a change in place elsewhere in a stack or into a specific stack below a mega‐merge, and so on.

    6. It seems to extend naturally to the *n*‐way split case.

    7. It helps us explain `--parallel`’s behaviour: `-r X --parallel` is equivalent to `-r X --after X- --before X+` (in the single‐commit case).

    8. `jj split` and `jj squash` would become exact inverses that can undo each other. (Actually, they would become almost the same command: if `--into` was a destination argument like `-d`/`-B`/`-A` are, they could be unified entirely. I think this is fascinating and promising, but also a rabbit hole that we shouldn’t block fixing `jj split` on.)

    9. There is a risk that operations like `-r X -A X` would create conflicts in `@` when selecting overlapping changes, in the same way that `-r X --parallel` can today. This is equivalent to the problem `jj squash --from ‹parent›` has. It is not yet clear how much of a workflow problem this would be, which is why we’d need to experiment with how it feels before committing. @scott2000 points out that this is not avoidable in the general case:

       > I don't think there's any easy way to solve this in general actually, since diff editors can produce arbitrary trees (like they don't even have to select whole lines). It becomes especially obvious when you look at `-d @-` (or `--parallel`). Let's say that `@-` has tree X, `@` has tree Y, and the user selects tree Z in the diff editor. The 2 resulting commits must have trees `(Y-Z+X)` and `Z` respectively. It's clear that `(Y-Z+X)` can have conflicts depending on what the user selected. The only thing that changes depending on whether we have the user select changes to keep or changes to remove is which of these two commits gets the conflict.

       but also proposes a potential solution that could solve this for both `jj split` and `jj squash`:

       > Yeah, I'm thinking adding destination arguments and saying "select what you want to put in the new commit" would work in most cases, and it would put `jj split` in the same spot as `jj squash` is currently. Maybe in the future we could solve the conflict problems with both of them at the same time (e.g. by adding a `--keep-selected` flag to both of them which keeps selected changes in the source commit instead of moving them to the destination commit).

    11. We still do not have consensus on what the default behaviour should be if we add these flags. I believe the conservative option would be for `-r X` to imply `-B X` by default, since it is so close to the stable behaviour and therefore minimally‐breaking. `-r X -B X` is also a case where conflicts will never be created, as with the current `jj split`. (It’s not wholly unbreaking; someone could be depending on the current change ID behaviour. My sense from experience is that when you care about change IDs, you’d prefer the new, more consistent behaviour, and when you don’t, it doesn’t matter. It’s less breaking than the change in trunk, in any case.)

       I also personally believe that `-B X` is *usually* what I (and @ilyagr, and @scott2000) want independent of compatibility concerns, per what I said about the frequency of these relative workflows in (5). However, there is not agreement on this. I know that @emesterhazy thinks that the preference of (5) over (4) is because of the (7) use case of bookmarks for Git compatibility, which I disagree with – I would want (4) more often even if I never used bookmarks and was doing things in a Jujutsu‐native environment with a custom Gerrit‐style forge. We will have to make some decision on what default we should pick for the long term at some point. We could even decide to make a destination argument mandatory if we can’t agree on what workflow to prioritize. It may be that we have to introduce a setting for this, if we can’t hash out the disagreement here.

Although we do not have consensus on what mode/workflow should be preferred by default, or how important the matter of `@` is, given the conversation, I believe that we all ended the discussion at least *relatively* happy with the idea of “restore the default behaviour for the upcoming release, then implement `-d`/`-B`/`-A` options for `jj split` and see how it feels in practice without immediately committing to that path”. I hope that this comment can serve as a reasonable summary of everyone’s positions in that discussion and that we can continue discussing from this. Please do correct me if you think I have misrepresented your position in this comment, especially @emesterhazy.

In terms of this PR, I am fine merging it; given all the above considerations, I think it is better to ship no change to the default behaviour in the next release than it is to commit to the breaking change in trunk. I personally think that the option would ideally be renamed, because although (10) would support the same workflow turning off the “legacy” behaviour supports, it would not behave in exactly the same way, and therefore the “non‐legacy” behaviour here is something that I think we wouldn’t want to have if (10) goes well. However I don’t want to block on that; it is undocumented, after all, so we’re not really offering any stability guarantees about it or enticing users to set it, and this has taken a lot of all of our energy already so I have no desire to pick further nits :)

I also hope that we can resolve this soon in a way that everyone can be happy with, and my hope is that, despite the remaining disagreements, (10) offers a potential path forward that everyone who participated in the discussion seems to consider at least a promising avenue of experimentation.

emesterhazy:
> 3. The current state of trunk is that the change ID and bookmarks go to the first commit in a split, but @ moves to the second. @ilyagr, @scott2000, and I all believe that this is a problematic inconsistency: for the same reason that bookmarks should move with the change ID, so should the working copy. My understanding of @emesterhazy’s position is that he doesn’t necessarily agree

The reason I think it's useful to move @ to the new commit is that I often use split to split out unfinished or "garbage" work that either needs to be edited or abandoned. If @ moves to the new commit, I'm all set to edit it or abandon it without specifying a change id or moving @ manually. Yuja mentioned the same thing [here](https://github.com/jj-vcs/jj/pull/5690#issuecomment-2658104964), so maybe I'm not the only one with this intuition.

I think peoples' preferences here will depend a lot on what the changes in the first and second commit represent to them, finished work, or unfinished work that they want to edit or abandon. If both commits contain finished work then you can make the `-B @` workflow work for you. If the second commit is unfinished or contains garbage, then it's pretty detrimental to move the change id and bookmarks to it. If the first commit is unfinished or contains garbage, then it's pretty detrimental for the tree of the second commit to contain those changes.

Maybe I'm missing something obvious, but I'm having trouble seeing how the `-B @` split works well for unfinished and garbage changes.  I don't want bookmarks to point to them, and I don't want finished work to depend on them.

martinvonz:
> 5\. There is another legitimate and common workflow: splitting _preliminary work_ out of a change: say, moving a simple preparatory refactor or rename as a new parent for a commit that implements a new feature, for the new feature to be stacked on top of. In this case, you want the change ID and bookmarks to go to the second commit.

I also do this frequently, but why does one want the change ID and bookmarks to go on the second commit in this case? Oh, it's more about the working copy, right? That logically follows if we say that the working copy follows the change ID (which I agree with).

> 9\. As it seemed like we will not be able to get a consensus on the exact behaviour `jj split` should have by the release

FWIW, this is something you *could* escalate to the maintainers and have them make a decision. I generally prefer to have the community discuss problems and reach a consensus, though. That reduces the risk of missing an important argument.

> 10\. At that point, I brought up the idea that had come up before of giving `jj split` arguments for `-d`/`-B`/`-A`, a la `new`/`rebase`/`duplicate`

Yes! Rolling back the recent change and starting by adding those options sounds like a great idea to me.

> 8\. Actually, they would become almost the same command: if `--into` was a destination argument like `-d`/`-B`/`-A` are, they could be unified entirely.

For reference, #4708 talks about a generalized "location" argument.

emilazy:
> Maybe I'm missing something obvious, but I'm having trouble seeing how the `-B @` split works well for unfinished and garbage changes. I don't want bookmarks to point to them, and I don't want finished work to depend on them.

My take:

* It works fine for the (5) use case where you have changes already “solidified” in some sense or submitted for review but want to split out preliminary work as a new parent.

* It works fine for the “incrementally making commits out of a heap of changes in `@`” use case because (a) usually that commit will have no meaningful description or bookmarks and (b) insofar as `@` keeps the change ID that’s a good thing because you may, say, want to gradually split out 3 new commits from it into your existing stack, but then also squash some stuff out of it along the way; keeping the change iD gives you a persistent identifier for your scratch work. It also lets you keep juggling various `-B` and `-A` values as appropriate if you want things to go to specific places in the stack, since `@` would remain in the same place.

---

Some additional detail about why the matter of `@` consistency turned into a discussion of workflows: if you use `jj split` to turn multiple changes in `@` into separate commits, then both the stable behaviour and the trunk behaviour work okay. (modulo the change ID of `@` becoming different every time, which I think is suboptimal per the previous part of this post, but it’s a more minor issue). (This task is something I, @ilyagr, @scott2000, and apparently @martinvonz often do – this is some part (5) and some part the “`jj split` as `git commit -p`” of (7), although I do not think it is a problematic case of “Git brain” in the same way that arguments about PR bookmark sliding may be – I wish I had split out those points.)

However, if we made `@` move with the change ID and bookmarks on top of the trunk behaviour, for consistency, then as soon as you split your first changes off, those changes would inherit the change ID, your `@` would end up on that commit, and you couldn’t continue splitting.

This led to different reactions: at first, it made me, @ilyagr, and @scott2000 suggest that we should make `@` move with the change ID and bookmarks but standardize on the (5) behaviour where the change ID and bookmarks move to the second commit. @emesterhazy’s reaction was that `@` should instead always go to the commit that *doesn’t* inherit the change ID. This ended up turning into the fairly intractable discussion of workflows and intuitions around working copies and squash‐vs.‐edit and descendant handling.

I think (10) provides a consistent model for answering the change ID/bookmarks/`@` stuff in such a way that `-r @ -B @` lets you do this incremental splitting and dovetails with the (5) workflow, while still allowing @emesterhazy’s preferred (4) “change eviction” mode as `-r @ -A @`. But I think we’re also not wholly on the same page about that yet, and as I said certainly don’t have an agreement on default or preferred workflows. I personally think the best way forward is to prototype (10) and see how well the interface works for the various workflows, since everyone seems at least curious about it, and talking about `jj split` behaviour in terms of hypotheticals seems to be way more confusing than just trying things out for whatever reason. (I’m not sure we even have a common mental model of how (10) would behave yet, although I think it’s pretty precise in my head at least.)

scott2000:
> Maybe I'm missing something obvious, but I'm having trouble seeing how the `-B @` split works well for unfinished and garbage changes. I don't want bookmarks to point to them, and I don't want finished work to depend on them.

I think for me at least, this isn't really a concern because I generally wouldn't have added a bookmark or pushed the change while it still contained unfinished/garbage changes. Adding a bookmark indicates "this change is finished" to me.

emesterhazy:
> > Maybe I'm missing something obvious, but I'm having trouble seeing how the `-B @` split works well for unfinished and garbage changes. I don't want bookmarks to point to them, and I don't want finished work to depend on them.
> 
> I think for me at least, this isn't really a concern because I generally wouldn't have added a bookmark or pushed the change while it still contained unfinished/garbage changes. Adding a bookmark indicates "this change is finished" to me.

Yeah, good point. That's probably generally true for me as well, but maybe everyone is not so disciplined, and mistakes do happen.

PhilipMetzger:
I still disagree with the decision here, even after taking a two day time-out. I'm currently to tired write an appropriate response of enough length but I assume I will find time for it during the next week. I will continue the conversation in the issue, where most of this should've happened anyway. 

All-in-all I'm utterly disappointed by the process here (it feels like the Google issue from last year regarding the communication). It also feels to me that the current default was chosen out of workflow preference, which is something we avoided until now. 

Thank you @emesterhazy for tackling the issue.

emesterhazy:
I'm planning to add the -A -B and -d flags without changing the default behavior at all. If those work nicely we can discuss what the default should be, or even have no default until the user sets one in their config.

I think this new approach will be more useful and flexible for everyone, so I'm happy we're trying it, even though I was satisfied with the behavior we reverted in this PR.

ilyagr:
I wrote a long reply in https://github.com/jj-vcs/jj/issues/3419#issuecomment-2712573005. I looked through the comments here, but I apologize in advance if I missed some important point, please let me know.

In there, I try to make a clear argument for moving everything to a second commit by default, and I also explain why I think we might want something like `--evict` (likely in addition to `-A`/`-B`, no instead of).

PR #5836, opened 2025-02-28
cli: clarify that the native backend is not production-ready

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/init.rs
+++ b/cli/src/commands/init.rs
@@ -56,9 +56,9 @@ pub(crate) fn cmd_init(
 
     if !command.settings().get_bool("ui.allow-init-native")? {
         return Err(user_error_with_hint(
-            "The native backend is disallowed by default.",
+            "The native backend is not ready for production use.",
             "Did you mean to call `jj git init`?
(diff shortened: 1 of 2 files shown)


ilyagr (commented):
Perhaps it'd be better to rename `jj init` to `jj native init` or `jj toybackend init`?

This PR would also be fine, but I think I'd want to revert it if we did the rename (not that it'd be very important).

(I'm happy to approve this, but I am trying to break a bad habit of approving a PR and starting a long discussion at the same time)

**Update:** I just noticed that Yuya already suggested this in https://github.com/jj-vcs/jj/pull/5827#issuecomment-2689895562.

ilyagr:
A related idea: what if we have a command "jj warnuser", and alias

```toml
aliases.init=["warnuser", """\
Please use `jj git init` instead. You can also configure
`aliases.init=["git", "init"]`. This is not the default since
we hope to have more backends in the future.
"""]
```

ilyagr:
I think I'm about to create a competing PR with the plan described above.

PhilipMetzger:
While this works, it actually doesn't solve the issue I want to address.

PR #5844, opened 2025-03-01
FAQ: detailed example for the "accidentally changed files in the wrong commit" question

ilyagr:
It was requested on Discord recently.

The result is almost a tutorial; we could consider actually moving it into the tutorial and providing a link in the FAQ. That could be done separately.

The `jj` commands outputs are faked in a few, hopefully inconspicuous, ways.
Hopefully, this is not distracting.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -298,14 +298,185 @@ jj config set --user git.private-commits "'''description(glob:'private:*')'''"
 
 ### I accidentally changed files in the wrong commit, how do I move the recent changes into another commit?
 
-Use `jj evolog -p` to see how your working-copy commit has evolved. Find the
-commit you want to restore the contents to. Let's say the current commit (with
-the changes intended for a new commit) are in commit X and the state you wanted
-is in commit Y. Note the commit id (normally in blue at the end of the line in
-the log output) of each of them. Now use `jj new` to create a new working-copy
-commit, then run `jj restore --from Y --into @-` to restore the parent commit
-to the old state, and `jj restore --from X` to restore the new working-copy
-commit to the new state.
+Let's say we are editing a commit for "featureA", and we forgot to run `jj
+new` or `jj commit` before doing some work that belongs in a new commit:
+
+```console
+$ jj log
+@  lnvvtrzo jjfan@example.org 2025-02-28 21:01:10 31a347e0
+│  featureA
+◆  zzzzzzzz root() 00000000
+$ cat file  # Oh no, the work on "feature B" should be in a separate commit!
+Done with feature A
+Working on feature B
+```
+
+#### Step 1: Find the commit id for the "last good version"
+
+<!-- TODO: Reorganize the two related questions, this one and
+  -- the one linked below
+  -->
+You can find [all the past versions of the working copy revision that `jj` has
+saved](#jj-is-said-to-record-the-working-copy-after-jj-log-and-every-other-command-where-can-i-see-these-automatic-saves)
+by running `jj evolog`. The obsolete versions will be marked as "hidden" and
+will have the same change id, but will have different commit ids. This
+represents different [commits] that are all parts of the same [change].
+
+For example, this is what the evolog might look like after you made two edits to
+the same change:
+
+```console
+$ # Note the word "hidden", the commit ids on the right,
+$ # and the unchanging change id on the left.
+$ jj evolog
+@  lnvvtrzo jjfan@example.org 2025-02-28 21:01:10 31a347e0
+│  featureA
+○  lnvvtrzo hidden jjfan@example.org 2025-02-28 21:00:51 b8004ab8
+│  featureA
+○  lnvvtrzo hidden jjfan@example.org 2025-02-28 20:50:05 e4d831d
+   (no description set)
+```
+
+Since commit `b800` is hidden, it is considered obsolete and `jj log` (without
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -298,14 +298,180 @@ jj config set --user git.private-commits "'''description(glob:'private:*')'''"
...
+   --- /dev/null
+   +++ b/file
+   @@ 0,0 +1,1 @@
+   +Working on feature A
+...
+```
+
+The version of the revision when we were actually done with "featureA" is thus
+saved in the commit with id `b800`. The change after that belongs in a new
+revision.
```
joyously, line 380: It is confusing to use "commit", "change", and "revision" this way.
Maybe 
> FeatureA was finished in commit "b800". FeatureB started in commit "31a3".

docs/FAQ.md
```diff
@@ -298,14 +298,180 @@ jj config set --user git.private-commits "'''description(glob:'private:*')'''"
...
+```console
+$ jj new -m "featureB"
+Working copy now at: pvnrkl 47171aa (empty) featureB
+Parent commit      : lnvvtr 31a347e featureA
+$ cat file
+Done with feature A
+Working on feature B
+```
+
+Now, we `jj restore` the revision `lnvvtr` to its state at commit `b80`. We use
```
joyously, line 409: The glossary says that revision and commit are synonyms. 
Should this revision be "change"?

docs/FAQ.md
```diff
@@ -298,14 +298,180 @@ jj config set --user git.private-commits "'''description(glob:'private:*')'''"
...
+   @@ 0,0 +1,1 @@
+   +Working on feature A
+...
+```
+
+The version of the revision when we were actually done with "featureA" is thus
+saved in the commit with id `b800`. The change after that belongs in a new
+revision.
+
+#### Step 2: Restore the existing revision to an older state and create a new revision with the current state
```
joyously, line 382: This sounds backwards, especially when you explain with "**First**, create a new empty child commit".

docs/FAQ.md
```diff
@@ -298,14 +298,180 @@ jj config set --user git.private-commits "'''description(glob:'private:*')'''"
...
+```console
+$ jj log
+@  lnvvtrzo jjfan@example.org 2025-02-28 21:01:10 31a347e0
+│  featureA
+◆  zzzzzzzz root() 00000000
+```
+
+We'd like to restore the "featureA" revision to its state at commit `b80` (see
+step 1 above for how we found that commit id), and to create a new revision
+called "featureB" that has the same contents as commit `31a`.
```
joyously, line 395: like the heading, it sounds backwards

docs/FAQ.md
ilyagr: Good point, I think I was thinking about it backwards.

docs/FAQ.md
ilyagr: I'm not immediately sure what's best to do here. "Change" sounds too generic, but I'm not sure where the consensus on terminology is. I wonder what people think; in particular if others feel that "change" would work OK for now everywhere I used "revision", that'd be helpful to know.

I hope we'll find something that works, even if temporarily, and isn't blocked fully resolving the terminology jungle. From this perspective, "change" seems promising.

joyously: Isn't the `lnvvtr` a change ID?

emilazy: When we had a big discussion about these concepts on Discord, @arxanas broke my faith in the primary of “change” as a concept by pointing out that divergent change IDs break a lot of natural talk of “changes”. In this case, there could be another `lnvvtr` that `jj restore` on `@-` wouldn’t do anything to; it’s actually `31a347e` we’re evolving, not `lnvvtr`. I don’t know what implications that has for this FAQ, but maybe it’s reasonable to say “revision” here because `lnvvtr` is really just acting as a shorthand name for `31a347e` rather than being something we’re addressing directly.

docs/FAQ.md
```diff
@@ -298,14 +298,180 @@ jj config set --user git.private-commits "'''description(glob:'private:*')'''"
...
+Done with feature A
+Working on feature B
+```
+
+Now, we `jj restore` the revision `lnvvtr` to its state at commit `b80`. We use
+the `--restore-descendants` flag so that the *contents* of the "featureB" revision is preserved.
+
+```console
+$ # We refer to `lnvvtr` as `@-` for brevity
+$ jj restore --from b80 --to @- --restore-descendants
```
emilazy, line 414: nit: Can we use `--into`?

docs/FAQ.md
ilyagr: I tried writing it with "change" as the primary concept, and it seems to work OK. As Joy mentioned, it matches our definitions in the glossary.

I think it might be nice to clean up the terminology, but then I guess we should change this text together with the glossary.

docs/FAQ.md
ilyagr: Thanks, I updated it. Hopefully it's clearer.

InCogNiTo124 (approved):
LGTM

ilyagr:
Thanks Marijan and everyone for looking this over!

I think this is better than the state before the PR, even if there's still something slightly off or confusing in here. So, I'm planning to merge this in before the release, unless somebody notices something blocking in the next ~hour.

PR #5845, opened 2025-03-01
cli: rename `jj init` to `jj debug init-local`, make `jj init` and `jj clone` print a hint with the error

ilyagr:
This competes with #5836 and parts of #5827.

We can also take just the first commit and make a dummy `jj init` that prints an error if the approach from the rest of the PR needs work.

**Update:** ~~I'm planning to add `jj clone` alias that acts the same as `jj init`, and add an entry to the changelog. That could be a separate PR, though.~~ (Done)

# Checklist

If applicable:


yuja:
Two comments:

1. Do we have any other use cases of `jj util error`? It looks like an ugly hack, and I think we can achieve the same goal by inspecting `clap::Error`. (We added `jj util exec` because it is useful in practice.)

2. Unless we're going to add more subcommants to `jj toy-backend`, it seems overkill to add namespaced command for testing purpose.

ilyagr:
> Unless we're going to add more subcommants to jj toy-backend, it seems overkill to add namespaced command for testing purpose.

I think it's nice to have a parallel: `jj git init`, `jj piper init`, `jj toy-backend init`. It also makes it extremely clear what is going on. Also, I don't see any cost to having a hidden namespace like this.

But, if you insist, we could rename it to `jj toy-backend-init`.

> Do we have any other use cases of jj util error? It looks like an ugly hack

This is a bit of a quick hack, though it doesn't look overly ugly to me. I'm happy to replace it with something else in the future, but I think it'd be very nice if `jj init` and `jj clone` showed a friendlier message soon for new users. So, I was excited when I realized this would work and is simple enough to potentially get into the upcoming release.

For reference, the current confusing situation is:

```console
$ jj clone something
error: unrecognized subcommand 'clone'

  tip: a similar subcommand exists: 'config'

Usage: jj [OPTIONS] <COMMAND>

```

The new version is:

```console
$ jj clone something
Error: `jj clone` is not defined by default. Perhaps you meant `jj git clone`?
Hint: You can configure `aliases.clone=["git", "clone"]` if you always use the Git backend.
```

>  I think we can achieve the same goal by inspecting clap::Error

I'm not sure I follow your plan. In my mind, the main alternative ideas are:

- Call it `jj debug error` or something instead of `jj util error`
- Allow a TOML syntax like:

```toml
[aliases]
init = { error = "Use jj git init", hint = "You can create an alias" }
```

ilyagr:
Another possible plan is to rename it to `jj toy-backend-init` or `jj toy-backend init` and create a normal dummy command `jj init` (that wouldn't be overridable by aliases). This wouldn't allow us to improve `jj clone`, since if it can't be replaced by aliases, it's a breaking change.

------

BTW, the way this PR was shaped was the following line of thinking:

- I think renaming the current `jj init` is better than changing its error message
- If we rename it, it's better to replace it with something. Together with `clone`, it's one of two commands that a completely new person might try to experiment with `jj`. As I think you pointed out elsewhere, it'd be nice to have `jj init` be overrideable with aliases.
- If we make `jj init` nice, let's do `jj clone` as well.

yuja:
> I think it's nice to have a parallel: `jj git init`, `jj piper init`, `jj toy-backend init`. It also makes it extremely clear what is going on. Also, I don't see any cost to having a hidden namespace like this.

Since `jj toy-backend` will never be user-facing command (as with the name), I don't think we need a nice structure. I would put random internal/testing commands to `jj debug` (because it would be wasteful to find better name for those things.)

> I'm not sure I follow your plan. The main alternative ideas I had were:
> 
> * Call it `jj debug error` or something instead of `jj util error`
> * Allow a TOML syntax like:

I mean the error message can be improved in a way that `jj git clone` is suggested if the command name was `clone`. We don't need to abuse alias machinery to achieve that.

ilyagr:
> I would put random internal/testing commands to jj debug

Should we make it `jj debug toy-backend-init` then?

> I mean the error message can be improved in a way that jj git clone is suggested if the command name was clone. 

Is there an equally trivial way to implement that? My first instinct would be to introduce a separate concept of "commands that should be overridable by aliases" somewhere in our alias resolution code.

Or do you want to try parsing the text of error messages somehow somewhere? Is this possible without reaching into `clap`'s internals too much?

yuja:
I think `clap::Error` stores some context (e.g. command name), but I have no idea if that's enough. Maybe you can `dbg!(error)` to dump what information is available.

`map_clap_cli_error()` does something for `--template` argument.

> That could work, good idea! Should we make it jj debug toy-backend-init then?

or `jj debug init-native-backend` (I don't know if we settled on renaming to the "toy" backend.)

ilyagr:
I like the name "toy-backend". We can also change it later; either way it's invisible to the user.

I'm not sure about looking into `clap` if there is no trivial solution. I might look into it, but at a glance, I'm not sure it's the best use of time or very fun. Perhaps we can go with what I have for now, and maybe improve it later? I find this PR's approach brutally simple and it achieves the exact UI we want, I think.

**Results at a glance:** https://docs.rs/clap/latest/clap/error/struct.Error.html makes it seem that I'd have to parse a `StyledStr`. There's likely a trade-off between brittle but easy-to-write code (parse the error's String representation, worry that a minor clap version might change it) and cleaner but complicated code (create the notion of a command that can be overridden by an alias).

ilyagr:
Oh, I see, there's also the "error-context" `clap` feature. It does not seem to be overly documented, but this must be what you were referring to exploring with `dgb!`.

ilyagr:
Yes, your approach works fine, I think. I haven't followed the entire program execution, but putting this check for "init" or "clone" inside `map_clap_cli_error()` seems to pass the tests.

ilyagr:
OK, it all actually worked beautifully. Thanks, I think it's a definite improvement! But let me know if I overlooked something; the tests pass, but I'm not super-familiar with the alias expansion code.

ilyagr:
One downside of the new approach compared to the aliases approach is that it's now slightly harder to modify these messages for use at (say) Google, where people might want `jj piper init` suggested instead of `jj git init`.

Now, I guess they'd need to add a patch either way. But if we had a concept of "system-wide config", overriding a default alias would be easier.

For other people's reference, the previous approach with aliases is (for now) available at https://github.com/jj-vcs/jj/compare/main...ilyagr:jj:usererror.

cli/tests/test_debug_toy_backend_init_command.rs
```diff
@@ -16,8 +16,10 @@ use crate::common::TestEnvironment;
 
 #[test]
 fn test_init_local_disallowed() {
+    // TODO(ilyagr): it is unclear whether `ui.allow-init-native` is useful now
+    // that `jj init` was renamed to `jj debug toy-backend-init`.
```
martinvonz, line 20: I think we should just delete the config

cli/tests/test_util_command.rs
```diff
@@ -43,7 +43,15 @@ fn test_gc_args() {
     let test_env = TestEnvironment::default();
     // Use the local backend because GitBackend::gc() depends on the git CLI.
```
martinvonz, line 44: We already depend on the git CLI for push and pull tests, so I think it should be okay to update this test case to subprocess to `git` instead

cli/tests/test_global_opts.rs
```diff
@@ -172,7 +172,7 @@ fn test_ignore_working_copy() {
 #[test]
 fn test_repo_arg_with_init() {
     let test_env = TestEnvironment::default();
-    let output = test_env.run_jj_in(".", ["init", "-R=.", "repo"]);
+    let output = test_env.run_jj_in(".", ["debug", "toy-backend-init", "-R=.", "repo"]);
```
martinvonz, line 175: Can we use `jj git init` here?

cli/tests/test_global_opts.rs
```diff
@@ -798,7 +798,7 @@ fn test_invalid_config() {
     let test_env = TestEnvironment::default();
 
     test_env.add_config("[section]key = value-missing-quotes");
-    let output = test_env.run_jj_in(".", ["init", "repo"]);
+    let output = test_env.run_jj_in(".", ["debug", "toy-backend-init", "repo"]);
```
martinvonz, line 801: Same here: can we use `jj git init`?

ilyagr: Indeed we can!

cli/tests/test_util_command.rs
ilyagr: Naively replacing this with "git init" works. I don't understand the implications, but I'll add it in.

ilyagr: Let me know if I can put something more meaningful in the commit description there.

cli/tests/test_debug_toy_backend_init_command.rs
ilyagr: I think so too. I added a TODO because I'm not sure I want to do it in this PR.

cli/src/cli_util.rs
```diff
@@ -3847,6 +3847,18 @@ fn map_clap_cli_error(mut cmd_err: CommandError, ui: &Ui, config: &StackedConfig
...
     };
+    if let Some(ContextValue::String(cmd)) = err.get(ContextKind::InvalidSubcommand) {
+        let cmd = cmd.as_str();
+        if matches!(cmd, "clone" | "init") {
+            // git commands
+            return user_error_with_hint(
+                format!("`jj {cmd}` is not defined by default. Perhaps you meant `jj git {cmd}`?"),
+                format!(
+                    r#"You can configure `aliases.{cmd}=["git", "{cmd}"]` if you always use the Git backend."#
+                ),
```
yuja, line 3858: Just add them as hints? I don't think we should replace the original error.

ilyagr: It then looks as follows, which seems overly verbose to me. I could try to recreate the error with fewer details (getting rid of the useless "tip" and "usage"). Or do you like it like this?

image

I tried fixing it by trying to copy the relevant parts of the error manually, but it did not work. I did:

```rust
            let mut new_err = clap::Error::new(err.kind());
            new_err.insert(
                ContextKind::InvalidSubcommand,
                ContextValue::String(cmd.to_string()),
            );
            // git commands
            return CommandError::from(new_err).hinted(
                formatdoc!(r#"
                   `jj {cmd}` is not defined by default. Perhaps you meant `jj git {cmd}`?
                   You can configure `aliases.{cmd}=["git", "{cmd}"]` if you always use the Git backend."#
                ),
            );
```

This breaks the formatting entirely:

image

WDYT is best?

ilyagr: Another option is to change `SuggestedCommand` property of the error and add a hint so that it says something like:

```
    error: unrecognized subcommand 'clone'

      tip: a similar subcommand exists: 'git clone'

    Usage: jj [OPTIONS] <COMMAND>

    Hint: You can configure `aliases.clone=["git", "clone"]` if
    you want `jj clone` to default to the Git backend.
```

If you like this option, I'll do it tomorrow.

yuja: I'm fine with either the verbose one or the one that modifies `SuggestedComment`. It should be better to reuse the original error object than reconstructing new fake error. The latter can go wrong if our implementation had a bug. (Wrong hint might be okay, but wrong error is really bad.)

PhilipMetzger:
This is currently my preferred variant, and it would be fine to land for me, if you take my renaming commit over.

CHANGELOG.md
```diff
@@ -91,6 +91,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj git fetch` now supports [string pattern syntax](docs/revsets.md#string-patterns)
   on `--remote` option and `git.fetch` configuration.
 
+ * The `jj init` stub that prints an error is now an alias that can be
```
ilyagr, line 94: TODO: it's not an alias anymore.

cli/src/cli_util.rs
ilyagr: The `SuggestedCommand` approach does not seem to work because `clap::Error` is not `Clone`, and I can't extract the original object from a `CommandError`.

I guess I'll go with the verbose version for now. I find it a bit ugly, but it's definitely better than what we have before this PR. It is true that constructing a new error could become an unexpected problem sometime down the line. I might start an FR for `clap` to support the use-case of replacing the suggested command in an error.

ilyagr:
Thanks Philip! If it's OK, I'd like to leave the renaming PR in your hands. Or, you can remind me to look at it later; there's no rush with it.

cli/src/cli_util.rs
ilyagr: I did file https://github.com/clap-rs/clap/issues/5935.

cli/src/cli_util.rs
```diff
@@ -3847,6 +3848,18 @@ fn map_clap_cli_error(mut cmd_err: CommandError, ui: &Ui, config: &StackedConfig
     let Some(err) = cmd_err.error.downcast_ref::<clap::Error>() else {
         return cmd_err;
     };
+    if let Some(ContextValue::String(cmd)) = err.get(ContextKind::InvalidSubcommand) {
+        let cmd = cmd.as_str();
+        if matches!(cmd, "clone" | "init") {
+            // git commands
+            return cmd_err.clone().hinted(formatdoc!(
```
yuja, line 3855: nit: `.clone()` shouldn't be needed. You can `cmd_err.add_hint(..)` and return.

cli/src/cli_util.rs
```diff
@@ -3847,6 +3848,18 @@ fn map_clap_cli_error(mut cmd_err: CommandError, ui: &Ui, config: &StackedConfig
...
+                r#"
+                   Perhaps you meant `jj git {cmd}`?
+                   You can configure `aliases.{cmd}=["git", "{cmd}"]` if
+                   you want `jj {cmd}` to work and always use the Git backend."#
```
yuja, line 3859: nit: I wouldn't insert `\n` after `if`. If you like, these two can be rendered as separate hints.

ilyagr: Oh, I hadn't realized we can do multiple hints. Nice!

cli/src/cli_util.rs
ilyagr: Sure. I hope the resulting `cmd.clone()` doesn't look too confusing, but it is cheaper than `cmd_err.clone()`.

cli/tests/test_debug_toy_backend_init_command.rs
ilyagr: I'll resolve this, assuming that a TODO is OK for now.

cli/tests/test_global_opts.rs
```diff
@@ -172,7 +172,7 @@ fn test_ignore_working_copy() {
 #[test]
 fn test_repo_arg_with_init() {
```
yuja, line 173: nit: rename to `test_repo_arg_with_git_init()` for consistency?

cli/src/cli_util.rs
```diff
@@ -3847,6 +3847,20 @@ fn map_clap_cli_error(mut cmd_err: CommandError, ui: &Ui, config: &StackedConfig
...
+        if matches!(cmd.as_str(), "clone" | "init") {
+            // git commands that a brand-new user might type on their first experiment with
+            // `jj`
+            let cmd = cmd.clone();
+            cmd_err.add_hint(format!(
+                "You probably want `jj git {cmd}`. See also `jj help git`."
+            ));
+            cmd_err.add_hint(format!(
+                    r#"You can configure `aliases.{cmd}=["git", "{cmd}"]` if you want `jj {cmd}` to work and always use the Git backend."#
+                ));
```
yuja, line 3860: nit: bad indent? (`rustfmt` accepts it maybe because of raw string literal, though.)

yuja (commented):
I'm not a fan of "toy-backend" naming, but there are no comments about that, so it should be okay. The last patch is nice. Thanks.

yuja (approved):

cli/tests/test_util_command.rs
```diff
@@ -41,18 +41,7 @@ fn test_util_config_schema() {
...
-            ".",
-            [
-                "debug",
-                "toy-backend-init",
-                "repo",
-                "--config=ui.allow-init-native=true",
-            ],
-        )
-        .success();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
```
martinvonz, line 44: nit: would be better to move this commit first and avoid the churn here

cli/tests/test_workspaces.rs
```diff
@@ -800,7 +800,15 @@ fn test_workspaces_current_op_discarded_by_other(automatic: bool) {
 
     // Use the local backend because GitBackend::gc() depends on the git CLI.
     test_env
-        .run_jj_in(".", ["init", "main", "--config=ui.allow-init-native=true"])
+        .run_jj_in(
```
martinvonz, line 803: switch to git backend here too?

cli/tests/test_global_opts.rs
martinvonz, line 1: > I, the author do not quite understand the
relevance of the removed comment in test_util_command.

I think it was just that we wanted to avoid depending on `git` (and a compatible version of it) being on the path when running tests. These days we require that for `jj git fetch/push` anyway, so there's no harm in requiring it here too.

cli/tests/test_file_track_untrack_commands.rs
martinvonz, line 1: nit: Just switch to the Git backend in this file too. I don't see any reason not to use the Git backend here.

cli/src/cli_util.rs
```diff
@@ -3847,6 +3847,20 @@ fn map_clap_cli_error(mut cmd_err: CommandError, ui: &Ui, config: &StackedConfig
...
+    if let Some(ContextValue::String(cmd)) = err.get(ContextKind::InvalidSubcommand) {
+        if matches!(cmd.as_str(), "clone" | "init") {
+            // git commands that a brand-new user might type on their first experiment with
+            // `jj`
+            let cmd = cmd.clone();
+            cmd_err.add_hint(format!(
+                "You probably want `jj git {cmd}`. See also `jj help git`."
+            ));
+            cmd_err.add_hint(format!(
+                    r#"You can configure `aliases.{cmd}=["git", "{cmd}"]` if you want `jj {cmd}` to work and always use the Git backend."#
```
martinvonz, line 3859: nit: add spaces around `=`? i think toml is usually formatted like that

cli/src/commands/debug/mod.rs
```diff
@@ -74,6 +77,7 @@ pub enum DebugCommand {
     Revset(DebugRevsetArgs),
     Snapshot(DebugSnapshotArgs),
     Template(DebugTemplateArgs),
+    ToyBackendInit(DebugToyBackendInitArgs),
```
martinvonz, line 80: How about calling it `jj debug init-local`? Just being under `debug` seems sufficient to make the user understand that they shouldn't use it. The whole `debug` group is hidden so most users probably don't even know that it exists.

cli/src/cli_util.rs
ilyagr: Suggestions? I experimented with it a bit, and `" \\\n"` didn't seem to work. Then I stopped trying.

cli/tests/test_global_opts.rs
ilyagr: There's also the option of depending on libgit2 for them, I thought these texts were choosing the you backend over that. I guess people prefer a clean break with libgit2.

cli/src/commands/debug/mod.rs
ilyagr: I agree that the name doesn't matter much. I can do `init-local` or whatever people like.

TBH, I don't understand the opposition to `toy-backend`; to me it describes the purpose of the backend well. The only other name that came to me was `demo-backend`.

Whatever we go with, it's easy to rename later if the role of this backend changes.

cli/tests/test_file_track_untrack_commands.rs
ilyagr: I'm considering leaving this for a separate cleanup PR, which is already needed for removing the option. It probably would be quite easy to do here, but I'll feel less rushed once the functional change lands.

cli/src/commands/debug/mod.rs
ilyagr: I'll go with `init-local` unless I hear otherwise.

cli/tests/test_util_command.rs
ilyagr: Done.

cli/tests/test_file_track_untrack_commands.rs
ilyagr: Thanks for https://github.com/jj-vcs/jj/pull/5875!

cli/src/commands/debug/mod.rs
martinvonz: FWIW, I think of it more as proof of concept that might one day evolve into a default native backend. There are many steps required before we get there, though, so it's less important what we call it right now. `init-local` matches the `LocalBackend` name well at least. But I'm also fine with calling it "toy" if that's what people prefer.

ilyagr:
I'll wait a couple of hours before merging. The latest significant change `toy-backend-init` -> `init-local` is discussed in <https://github.com/jj-vcs/jj/pull/5845#discussion_r1979714698>. Note that we are free to change the name later.

cli/src/commands/debug/mod.rs
ilyagr: In my mind `toy-backend` is a shorter/clearer version of `poc-backend` or `proof-of-concept-backend`.

cli/src/cli_util.rs
yuja: FWIW, to fix the wrapping, you could use toml_edit type to format the list. Something like
```
format!("... {cmd}={args} ...", args = ConfigValue::from_iter(["git", &cmd]))
```
(untested)

cli/src/commands/debug/toy_backend_init.rs
yuja, line 1: nit: forget to rename the module?

ilyagr: Indeed, thanks for catching it!

PR #5846, opened 2025-03-01
cli/tests: move remaining tests to gitoxide

bsdinis:
As per #5548, this concludes the port of the current tests to gitoxide.

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_git_clone.rs
+++ b/cli/tests/test_git_clone.rs
@@ -18,36 +18,26 @@ use std::path::PathBuf;
 
 use indoc::formatdoc;
 use test_case::test_case;
+use testutils::git;
 
 use crate::common::to_toml_value;
 use crate::common::CommandOutput;
 use crate::common::TestEnvironment;
 
-fn set_up_non_empty_git_repo(git_repo: &git2::Repository) {
+fn set_up_non_empty_git_repo(git_repo: &gix::Repository) {
     set_up_git_repo_with_file(git_repo, "file");
 }
 
-fn set_up_git_repo_with_file(git_repo: &git2::Repository, filename: &str) {
-    let signature =
-        git2::Signature::new("Some One", "some.one@example.com", &git2::Time::new(0, 0)).unwrap();
-    let mut tree_builder = git_repo.treebuilder(None).unwrap();
-    let file_oid = git_repo.blob(b"content").unwrap();
-    tree_builder
-        .insert(filename, file_oid, git2::FileMode::Blob.into())
-        .unwrap();
-    let tree_oid = tree_builder.write().unwrap();
-    let tree = git_repo.find_tree(tree_oid).unwrap();
-    git_repo
-        .commit(
-            Some("refs/heads/main"),
-            &signature,
-            &signature,
-            "message",
-            &tree,
-            &[],
-        )
-        .unwrap();
-    git_repo.set_head("refs/heads/main").unwrap();
+fn set_up_git_repo_with_file(git_repo: &gix::Repository, filename: &str) {
+    git::add_commit(
+        git_repo,
+        "refs/heads/main",
+        filename,
+        b"content",
+        "message",
+        &[],
+    );
+    git::set_symbolic_reference(git_repo, "HEAD", "refs/heads/main");
 }
 
 #[test_case(false; "use git2 for remote calls")]
@@ -59,7 +49,7 @@ fn test_git_clone(subprocess: bool) {
         test_env.add_config("git.subprocess = false");
     }
     let git_repo_path = test_env.env_root().join("source");
-    let git_repo = git2::Repository::init(git_repo_path).unwrap();
+    let git_repo = git::init(git_repo_path);
 
     // Clone an empty repo
     let output = test_env.run_jj_in(".", ["git", "clone", "source", "empty"]);
@@ -82,8 +72,8 @@ fn test_git_clone(subprocess: bool) {
     Fetching into new repo in "$TEST_ENV/clone"
     bookmark: main@origin [new] tracked
     Setting the revset alias `trunk()` to `main@origin`
-    Working copy now at: uuqppmxq 1f0b881a (empty) (no description set)
-    Parent commit      : mzyxwzks 9f01a0e0 main | message
+    Working copy now at: uuqppmxq f78d2645 (empty) (no description set)
+    Parent commit      : qomsplrm ebeb70d8 main | message
     Added 1 files, modified 0 files, removed 0 files
     [EOF]
     "#);
@@ -202,8 +192,8 @@ fn test_git_clone(subprocess: bool) {
     Fetching into new repo in "$TEST_ENV/nested/path/to/repo"
     bookmark: main@origin [new] tracked
     Setting the revset alias `trunk()` to `main@origin`
-    Working copy now at: uuzqqzqu df8acbac (empty) (no description set)
-    Parent commit      : mzyxwzks 9f01a0e0 main | message
+    Working copy now at: uuzqqzqu cf5d593e (empty) (no description set)
+    Parent commit      : qomsplrm ebeb70d8 main | message
     Added 1 files, modified 0 files, removed 0 files
     [EOF]
     "#);
@@ -253,7 +243,7 @@ fn test_git_clone_colocate(subprocess: bool) {
         test_env.add_config("git.subprocess = false");
     }
     let git_repo_path = test_env.env_root().join("source");
-    let git_repo = git2::Repository::init(git_repo_path).unwrap();
+    let git_repo = git::init(git_repo_path);
 
     // Clone an empty repo
     let output = test_env.run_jj_in(".", ["git", "clone", "source", "empty", "--colocate"]);
@@ -287,8 +277,8 @@ fn test_git_clone_colocate(subprocess: bool) {
     Fetching into new repo in "$TEST_ENV/clone"
     bookmark: main@origin [new] tracked
     Setting the revset alias `trunk()` to `main@origin`
-    Working copy now at: uuqppmxq 1f0b881a (empty) (no description set)
-    Parent commit      : mzyxwzks 9f01a0e0 main | message
+    Working copy now at: uuqppmxq f78d2645 (empty) (no description set)
+    Parent commit      : qomsplrm ebeb70d8 main | message
     Added 1 files, modified 0 files, removed 0 files
     [EOF]
     "#);
@@ -301,41 +291,51 @@ fn test_git_clone_colocate(subprocess: bool) {
         git_repo.head().expect("Repo head should be set").name()
     );
 
-    let jj_git_repo = git2::Repository::open(test_env.env_root().join("clone"))
-        .expect("Could not open clone repo");
+    let jj_git_repo = git::open(test_env.env_root().join("clone"));
     assert_eq!(
         jj_git_repo
-            .head()
+            .head_id()
             .expect("Clone Repo HEAD should be set.")
-            .symbolic_target(),
+            .detach(),
         git_repo
-            .head()
+            .head_id()
             .expect("Repo HEAD should be set.")
-            .symbolic_target()
+            .detach(),
     );
     // ".jj" directory should be ignored at Git side.
-    #[allow(clippy::format_collect)]
-    let git_statuses: String = jj_git_repo
-        .statuses(None)
-        .unwrap()
-        .iter()
-        .map(|entry| format!("{:?} {}\n", entry.status(), entry.path().unwrap()))
-        .collect();
+    let git_statuses = git::status(&jj_git_repo);
     insta::allow_duplicates! {
-    insta::assert_snapshot!(git_statuses, @r###"
-    Status(IGNORED) .jj/.gitignore
-    Status(IGNORED) .jj/repo/
-    Status(IGNORED) .jj/working_copy/
-    "###);
+    insta::assert_debug_snapshot!(git_statuses, @r#"
+    [
+        GitStatus {
+            path: ".jj/.gitignore",
+            status: Worktree(
+                Ignored,
+            ),
+        },
+        GitStatus {
+            path: ".jj/repo",
+            status: Worktree(
+                Ignored,
+            ),
+        },
+        GitStatus {
+            path: ".jj/working_copy",
+            status: Worktree(
+                Ignored,
+            ),
+        },
+    ]
+    "#);
     }
 
     // The old default bookmark "master" shouldn't exist.
     insta::allow_duplicates! {
     insta::assert_snapshot!(
         get_bookmark_output(&test_env, &test_env.env_root().join("clone")), @r"
-    main: mzyxwzks 9f01a0e0 message
-      @git: mzyxwzks 9f01a0e0 message
-      @origin: mzyxwzks 9f01a0e0 message
+    main: qomsplrm ebeb70d8 message
+      @git: qomsplrm ebeb70d8 message
+      @origin: qomsplrm ebeb70d8 message
     [EOF]
     ");
     }
@@ -463,8 +463,8 @@ fn test_git_clone_colocate(subprocess: bool) {
     Fetching into new repo in "$TEST_ENV/nested/path/to/repo"
     bookmark: main@origin [new] tracked
     Setting the revset alias `trunk()` to `main@origin`
-    Working copy now at: vzqnnsmr 9407107f (empty) (no description set)
-    Parent commit      : mzyxwzks 9f01a0e0 main | message
+    Working copy now at: vzqnnsmr 589d0921 (empty) (no description set)
+    Parent commit      : qomsplrm ebeb70d8 main | message
     Added 1 files, modified 0 files, removed 0 files
     [EOF]
     "#);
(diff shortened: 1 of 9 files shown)


lib/testutils/src/git.rs
```diff
@@ -350,3 +350,42 @@ impl<'a> IndexManager<'a> {
             .unwrap();
     }
 }
+
+pub fn add_remote(repo: &gix::Repository, remote_name: &str, mut remote: gix::Remote<'_>) {
```
yuja, line 380: nit: I think it's simpler to pass remote url instead of `gix::Remote`. Most callers don't need the remote object.

lib/testutils/src/git.rs
```diff
@@ -350,3 +350,42 @@ impl<'a> IndexManager<'a> {
...
+    let mut config = repo.config_snapshot().clone();
+    remote.save_as_to(remote_name, &mut config).unwrap();
+    let mut config_file = std::fs::File::create(config.meta().path.as_ref().unwrap()).unwrap();
+    config
+        .write_to_filter(&mut config_file, |section| section.meta() == config.meta())
+        .unwrap();
+}
+
+pub fn rename_remote(repo: &gix::Repository, original: &str, new: &str) {
```
yuja, line 389: It's probably better to rewrite the tests to not rely on `rename_remote()` because we don't have well-tested rename functions in `gix`.

I think most of them use `.remote_rename()` just because it was easier.

cli/tests/test_git_remotes.rs
```diff
@@ -254,15 +256,14 @@ fn test_git_remote_named_git() {
...
-    let git_repo = git2::Repository::init(&repo_path).unwrap();
-    git_repo
-        .remote("git", "http://example.com/repo/repo")
-        .unwrap();
+    let git_repo = git::init(&repo_path);
+    let remote = git_repo.remote_at("http://example.com/repo/repo").unwrap();
+    git::add_remote(&git_repo, "git", remote);
     test_env
-        .run_jj_in(&repo_path, ["git", "init", "--git-repo=."])
+        .run_jj_in(&repo_path, &["git", "init", "--git-repo=."])
```
yuja, line 263: nit: copy paste errors?

cli/tests/test_git_fetch.rs
```diff
@@ -14,46 +14,41 @@
...
-        .unwrap()
-}
-
-fn git2_signature() -> git2::Signature<'static> {
-    git2::Signature::new("Some One", "some.one@example.com", &git2::Time::new(0, 0)).unwrap()
+fn add_commit_to_branch(git_repo: &gix::Repository, branch: &str) -> gix::ObjectId {
+    git::add_commit(
+        git_repo,
+        &format!("refs/heads/{branch}"),
+        branch,
```
yuja, line 26: nit: I would add `branch, // filename` comment here because `branch` is confusing in this context.

cli/tests/test_git_fetch.rs
```diff
@@ -14,46 +14,41 @@
...
-    add_commit_to_branch(&git_repo, remote);
+    let git_repo = git::init(git_repo_path);
+    git::add_commit(
+        &git_repo,
+        &format!("refs/heads/{remote}"),
+        "file",
+        remote.as_bytes(),
+        "message",
+        &[],
+    );
```
yuja, line 45: nit: just use `add_commit_to_branch()`?

cli/tests/test_git_fetch.rs
```diff
@@ -100,25 +95,47 @@ fn clone_git_remote_into(
...
+            )
+            .unwrap()
+            .connect(gix::remote::Direction::Fetch)
+            .unwrap()
+            .prepare_fetch(
+                gix::progress::Discard,
+                gix::remote::ref_map::Options::default(),
+            )
+            .unwrap()
+            .receive(gix::progress::Discard, &gix::interrupt::IS_INTERRUPTED)
```
yuja, line 120: Since this isn't trivial, I would extract `testutils::git::fetch()` function. It might be also better to shell out `git`.

`testutils::git::clone()` might just work, but I haven't checked.

cli/tests/test_git_fetch.rs
```diff
@@ -137,10 +154,10 @@ fn test_git_fetch_with_default_config(subprocess: bool) {
 
     test_env.run_jj_in(&repo_path, ["git", "fetch"]).success();
     insta::allow_duplicates! {
-    insta::assert_snapshot!(get_bookmark_output(&test_env, &repo_path), @r"
-    origin@origin: oputwtnw ffecd2d6 message
-    [EOF]
-    ");
+     insta::assert_snapshot!(get_bookmark_output(&test_env, &repo_path), @r"
```
yuja, line 157: nit: weird indent (there are more instances)

cli/tests/test_git_fetch.rs
```diff
@@ -2026,61 +2039,97 @@ fn test_git_fetch_preserve_commits_across_repos(subprocess: bool) {
...
+    assert!(tree_merge.conflicts.is_empty());
+    let merge_oid = tree_merge.tree.write().unwrap().detach();
+    git::write_commit(
+        &upstream_repo,
+        "refs/heads/upstream",
+        merge_oid,
+        "merge",
+        [merge_base.id(), merge_target.id()],
+    );
+    */
```
yuja, line 2125: nit: bad edits? Can you review diffs before submitting PR?

lib/testutils/src/git.rs
```diff
@@ -158,6 +158,25 @@ pub fn write_commit(
...
+    target: gix::ObjectId,
+) {
+    let signature = signature();
+    let tree_id = gix::ObjectId::empty_tree(repo.object_hash());
+    repo.commit_as(
+        &signature,
+        &signature,
+        reference,
+        "merge",
+        tree_id,
```
yuja, line 174: It looks weird that we're creating merge commit with empty tree. Maybe better to inline if that's the expected behavior?

cli/tests/test_git_clone.rs
```diff
@@ -18,38 +18,12 @@ use std::path::PathBuf;
...
-            Some("refs/heads/main"),
-            &signature,
-            &signature,
-            "message",
-            &tree,
-            &[],
-        )
-        .unwrap();
-    git_repo.set_head("refs/heads/main").unwrap();
-}
```
yuja, line 51: nit: even though it might make sense to inline these functions, I wouldn't do that within this commit. This commit is noisy due to hash changes.

cli/tests/test_git_clone.rs
```diff
@@ -287,55 +283,51 @@ fn test_git_clone_colocate(subprocess: bool) {
...
-        git_repo.head().expect("Repo head should be set").name()
-    );
-
-    let jj_git_repo = git2::Repository::open(test_env.env_root().join("clone"))
-        .expect("Could not open clone repo");
+    let jj_git_repo = git::open(test_env.env_root().join("clone"));
     assert_eq!(
         jj_git_repo
-            .head()
+            .head_id()
```
yuja, line 298: Perhaps, it should be `.head_ref()`?

I'm not sure if that matters here, but the original code compares `.symbolic_target()`s.

cli/tests/test_git_clone.rs
```diff
@@ -287,55 +283,51 @@ fn test_git_clone_colocate(subprocess: bool) {
...
+        git::GitStatus::new(
+            ".jj/repo",
+            git::GitStatusInfo::Worktree(git::WorktreeStatus::Ignored),
+        ),
+        git::GitStatus::new(
+            ".jj/working_copy",
+            git::GitStatusInfo::Worktree(git::WorktreeStatus::Ignored),
+        ),
+    ];
+    assert_eq!(git_statuses, expected);
```
yuja, line 322: nit: maybe we can use `insta::assert_debug_snapshot!()`?

lib/testutils/src/git.rs
bsdinis: is this another case of using `git` directly instead?

cli/tests/test_git_fetch.rs
bsdinis: yeah, it's weird these are not caught by rustfmt, maybe because they're in a macro?

cli/tests/test_git_fetch.rs
bsdinis: apologies, completely missed this

lib/testutils/src/git.rs
bsdinis: I think I found a way here, by merging the trees manually. I think it probably doesn't hurt to have the merging as it is here assuming this function now make more sense, what do you think?

cli/tests/test_git_clone.rs
bsdinis: done

cli/tests/test_git_clone.rs
bsdinis: done

cli/tests/test_git_clone.rs
bsdinis: done

cli/tests/test_git_clone.rs
bsdinis: so i looked into this (`.head_ref` was failing).

`jj git clone` is setting HEAD to the actual commit id, which makes sense to me (unsure how `git2`'s `symbolic_target` was working, but maybe it was resolving it somehow?)

lib/testutils/src/git.rs
yuja: Could be.

As I said, it's also good to simplify tests. I don't think they're explicitly testing "rename" behavior.

lib/testutils/src/git.rs
yuja: Do we need a real merge function? I assumed the caller wouldn't care about the tree content.

cli/tests/test_git_fetch.rs
yuja: yes, `insta` macros can't be formatted. (If they could, this block would be indented with 8 spaces.)

lib/testutils/src/git.rs
bsdinis: It doesn't really, I'm just trying to match the original behaviour as best as possible

bsdinis: And this way it becomes reusable

cli/tests/test_git_clone.rs
yuja: Right. The doc says `.head()` "resolve the reference", so getting `symbolic_target()` there is just wrong.
https://docs.rs/git2/latest/git2/struct.Repository.html#method.head

lib/testutils/src/git.rs
bsdinis: I think all the usecases here are renaming the remote to a value that is not supported by `jj` (e.g., `git` or a remote with slashes).

I think it's ok to use `git` here

lib/testutils/src/git.rs
yuja: reusable but we have to maintain the code on API change, etc. I don't have strong feeling, but I would avoid adding non-trivial helper function unless needed.

lib/testutils/src/git.rs
yuja: Yes, it should be okay to use `git` command. I just mean that we wouldn't have to rename an existing remote to `git`, but we can probably create a remote named `git` (without using `jj`.)

cli/tests/test_git_clone.rs
bsdinis: nice, resolving then

lib/testutils/src/git.rs
bsdinis: sounds good, will inline

bsdinis: will half inline: will move this into the `test_git_fetch` file: this is a long piece of code and the test is big as it is.

lib/testutils/src/git.rs
```diff
@@ -350,3 +350,29 @@ impl<'a> IndexManager<'a> {
...
+
+pub fn rename_remote(repo_dir: impl AsRef<Path>, original: &str, new: &str) {
+    let output = std::process::Command::new("git")
+        .current_dir(repo_dir)
+        .args(["remote", "rename", original, new])
+        .output()
+        .unwrap();
+    assert!(
+        output.status.success(),
+        "git remote rename failed with exit code {}:\n{}\n----- stderr -----\n{}",
```
yuja, line 381: nit: redundant "exit code " (`output.status` would print "exit status N")

cli/tests/test_git_fetch.rs
```diff
@@ -2130,61 +2153,72 @@ fn test_git_fetch_preserve_commits_across_repos(subprocess: bool) {
...
 
     // merge fork/feature into the upstream/upstream
-    let mut fork_remote = upstream_repo
-        .remote("fork", fork_path.to_str().unwrap())
+    let fork_remote = git::add_remote(&upstream_repo, "fork", fork_path.to_str().unwrap());
+    fork_remote
+        .with_refspecs(
+            Some("+refs/heads/*:refs/remotes/fork/*".as_bytes()),
+            gix::remote::Direction::Fetch,
+        )
```
yuja, line 2181: Perhaps, the fetch refspecs should be set by `git::add_remote()` (as `git remote add` would do.)

cli/tests/test_git_clone.rs
```diff
@@ -479,16 +479,18 @@ fn test_git_clone_remote_default_bookmark(subprocess: bool) {
         test_env.add_config("git.subprocess = false");
     }
     let git_repo_path = test_env.env_root().join("source");
-    let git_repo = git2::Repository::init(git_repo_path).unwrap();
+    let git_repo = git::init(git_repo_path.clone());
+
     set_up_non_empty_git_repo(&git_repo);
-    // Create non-default bookmark in remote
```
yuja, line 484: nit: this comment is still valid?

lib/testutils/src/git.rs
```diff
@@ -290,6 +290,13 @@ pub struct GitStatus {
     status: GitStatusInfo,
 }
 
+impl GitStatus {
+    pub fn new(path: impl Into<String>, status: GitStatusInfo) -> Self {
+        let path = path.into();
+        GitStatus { path, status }
+    }
```
yuja, line 297: nit: unused function?

cli/tests/test_git_fetch.rs
bsdinis: agree

cli/tests/test_git_clone.rs
bsdinis: yes, correct

lib/testutils/src/git.rs
bsdinis: The `git::clone` was doing the same thing, I fixed it also in this commit, hope that's ok

jonathantanmy:
Even here, you don't seem to need the "blocking-network-client" feature. I removed it, ran `cargo nextest run --workspace` from the repo toplevel, and everything passes.

bsdinis:
> Even here, you don't seem to need the "blocking-network-client" feature. I removed it, ran `cargo nextest run --workspace` from the repo toplevel, and everything passes.

It fails to compile the tests for me, are you sure you ran the most recent version?
```
error[E0433]: failed to resolve: could not find `ref_map` in `remote`
    --> cli/tests/test_git_fetch.rs:2182:26
     |
2182 |             gix::remote::ref_map::Options::default(),
     |                          ^^^^^^^ could not find `ref_map` in `remote`
     |
note: found an item that was configured out
    --> /home/bsdinis/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gix-0.70.0/src/remote/mod.rs:59:22
     |
59   | pub use connection::{ref_map, AuthenticateFn, Connection};
     |                      ^^^^^^^
note: the item is gated here
    --> /home/bsdinis/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gix-0.70.0/src/remote/mod.rs:58:1
     |
58   | #[cfg(any(feature = "async-network-client", feature = "blocking-network-client"))]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider importing one of these structs
     |
14   + use gix::config::file::includes::Options;
     |
14   + use gix::config::file::init::Options;
     |
14   + use gix::create::Options;
     |
14   + use gix::diff::Options;
```

yuja:
> "blocking-network-client" feature.

Just shell out `git fetch`? Maybe we can also use non-colocated jj repo to set up the remote git repos.

bsdinis:
> Just shell out `git fetch`? Maybe we can also use non-colocated jj repo to set up the remote git repos.

I don't see the problem of adding the test dependency on this gix feature

yuja:
Maybe it's easier for Google to deal with dependencies? I generally prefer fewer dependencies, and fewer test-specific implementations.

cli/tests/test_git_fetch.rs
```diff
@@ -65,6 +53,46 @@ fn add_git_remote(test_env: &TestEnvironment, repo_path: &Path, remote: &str) ->
...
+            base_tree.id(),
+            base_tree.id(),
+            target_tree.id(),
+            labels,
+            options,
+        )
+        .unwrap();
+
+    assert!(!outcome.failed_on_first_unresolved_conflict);
+    let tree_id = outcome.tree.write().unwrap().detach();
```
yuja, line 83: I thought we wouldn't need a real tree merging (and that's the reason to not add `git::merge_commits()` helper.) If the tree content matters, maybe we can just shell out `git merge`?

Another option is to use `jj` to set up the upstream repo.

bsdinis: you're correct, an empty tree is good enough. The only thing that matters is that they are a merge commit. `git::write_commit` with an empty tree is good enough

yuja (approved):
Thanks!

PR #5847, opened 2025-03-01
cli_util: extract shared `compute_commit_location` function

bnjmnt4n:
This will be used to compute the location to place commits from the
`--destination`, `--insert-after`, and `--insert-before` options. This
is used across the `new`, `duplicate`, and `rebase` commands, and can be
used for further commands as well.

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -57,6 +57,7 @@ use jj_lib::backend::CommitId;
 use jj_lib::backend::MergedTreeId;
 use jj_lib::backend::TreeValue;
 use jj_lib::commit::Commit;
+use jj_lib::commit::CommitIteratorExt;
 use jj_lib::config::ConfigGetError;
 use jj_lib::config::ConfigGetResultExt as _;
 use jj_lib::config::ConfigLayer;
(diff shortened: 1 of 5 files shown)


cli/src/cli_util.rs
```diff
@@ -3062,6 +3063,104 @@ impl fmt::Display for RemoteBookmarkNamePattern {
...
+/// The `destination` argument is mutually exclusive to the `insert_after` and
+/// `insert_before` arguments.
+pub fn compute_commit_location(
+    ui: &mut Ui,
+    workspace_command: &mut WorkspaceCommandHelper,
+    destination: &Option<Vec<RevisionArg>>,
+    insert_after: &Option<Vec<RevisionArg>>,
+    insert_before: &Option<Vec<RevisionArg>>,
+    commit_description: &str,
+) -> Result<(Vec<Commit>, Vec<Commit>), CommandError> {
```
martinvonz, line 3077: Is it better to return lists of commit IDs? Many callers don't seem to need the commit objects, so it's better to not load the objects here if not necessary. (Since the `Store` caches commits, it's less of a problem to unnecessarily load a commit twice, if we were to load some commit here and then again in the caller.)

martinvonz (commented):
Thanks for cleaning up this!

cli/src/cli_util.rs
martinvonz: I think that's best. I don't think the extra reading matters much when the commit is cached. Reading extra commits is probably a bigger problem, especially with slow backend. It may make a difference on the server at Google. But let me know if you think it gets to messy.

cli/src/cli_util.rs
```diff
@@ -3062,6 +3063,102 @@ impl fmt::Display for RemoteBookmarkNamePattern {
...
+/// Computes the location (new parents and new children) to place commits.
+///
+/// The `destination` argument is mutually exclusive to the `insert_after` and
+/// `insert_before` arguments.
+pub fn compute_commit_location(
+    ui: &mut Ui,
+    workspace_command: &mut WorkspaceCommandHelper,
+    destination: &Option<Vec<RevisionArg>>,
+    insert_after: &Option<Vec<RevisionArg>>,
+    insert_before: &Option<Vec<RevisionArg>>,
```
yuja, line 3075: nit: `Option<&[..]>` is more common as arguments. It might be better to add named struct.

cli/src/cli_util.rs
bnjmnt4n: It seems like the main reason I returned the commits is that `workspace_command.resolve_some_revsets_default_single` also returned commit objects. Maybe we should create a version of the function that only returns commit IDs instead?

cli/src/cli_util.rs
```diff
@@ -3062,6 +3063,102 @@ impl fmt::Display for RemoteBookmarkNamePattern {
...
+                .iter()
+                .flat_map(|commit| commit.parent_ids().iter().cloned())
+                .unique();
+            let new_parents: Vec<_> = new_parent_ids
+                .map(|commit_id| workspace_command.repo().store().get_commit(&commit_id))
+                .try_collect()?;
+
+            (new_parents, before_commits)
+        }
+        _ => unreachable!(),
```
emesterhazy, line 3122: I think we should return an error rather than panicking here. It seems like there are two scenarios:

1. The caller didn't provide any of the dest, before, or after arguments.
2. The caller provided dest and one or more of the other two.

Can we return an Internal CommandError with a useful message in those cases?

bnjmnt4n: I didn't do this initially because it was expected that the arguments would be passed in directly from the clap command, and clap would have validated these conditions. I'm not sure we need to create a `CommandError` since reaching this case would be an internal error, and probably doesn't have a useful message for external users?

emesterhazy: Users should never see it, but implementers might if they make a mistake (i.e. pass the default destination arg like you called out below).

You can treat it as an optional nit if you want, I just think it's better not to panic if it can be avoided.

yuja: I would use `panic!()` or `unreachable!()` with message in that case.

cli/src/cli_util.rs
```diff
@@ -3062,6 +3063,102 @@ impl fmt::Display for RemoteBookmarkNamePattern {
...
+    Ok((new_parents, new_children))
+}
+
+/// Ensure that there is no possible cycle between the potential children and
+/// parents of the given commits.
+fn ensure_no_commit_loop(
+    repo: &ReadonlyRepo,
+    children_expression: &Rc<ResolvedRevsetExpression>,
+    parents_expression: &Rc<ResolvedRevsetExpression>,
+    commit_type: &str,
```
martinvonz, line 3144: nit: call it `commit_description` here too? or maybe call it `commit_type` above (i think i like that a bit better)

cli/src/commands/duplicate.rs
```diff
@@ -72,7 +69,7 @@ pub(crate) struct DuplicateArgs {
         value_name = "REVSETS",
         add = ArgValueCandidates::new(complete::all_revisions)
     )]
-    destination: Vec<RevisionArg>,
+    destination: Option<Vec<RevisionArg>>,
```
martinvonz, line 72: Why did this have to change? Is there a difference between `None` and `Some(vec![])` now? Oh, looks like it was changed this way for `jj rebase` due to @yuja's comment [here](https://github.com/jj-vcs/jj/pull/4640#discussion_r1800289570). I tried dropping these `Option` wrappings and I found it simpler, fwiw. The closure in `compute_commit_location()` can still return an `Option` for easier and safer pattern matching. I can send a PR once this has been merged to show what it looks like. We can discuss there.

martinvonz (approved):
LGTM. I don't see anything that I think needs to be changed in this PR.

cli/src/commands/duplicate.rs
yuja: My feeling is that `Option<Vec<_>>` encodes the intent better if an empty value is logically different than unspecified (e.g. use default instead of empty list.)

Issue #3831, opened 2024-06-05
FR: Option to sort listed branches based on last commit date


gpanders:
**Is your feature request related to a problem? Please describe.**

Git has the ability to sort the output of `git branch` based on last commit date with `git branch --sort=-committerdate` (or `git config branch.sort -committerdate`). I find this extremely useful since the branches that I am mostly frequently switching between are the ones that have been active recently.

`jj` lists all branches alphabetically by default, but it'd be great to have the ability to list them based on the last commit date as well.

**Describe the solution you'd like**

Simple option would be to add a `-s`/`--sort` flag to `jj branch list`. As a first implementation, this could support only `name` and `committerdate` (I don't have strong opinions on what term to use here, for now just copying from Git). Like Git, prefixing the sort key with `-` sorts in descending order.

**Describe alternatives you've considered**

This could maybe also be done with a template for the entire output of `jj branch list`? I don't know if the templating engine could support sorting though.

**Additional context**
N/A

PR #5849, opened 2025-03-01
cli: Add `--sort` option to the bookmark list command

zummenix:
This allows different sort orders for the bookmarks list, not only alphabetically (by name).

Relates to #3831

The feature is intentionally implemented in somewhat limited capacity. I want to receive feedback first.

For now only two sort keys are supported `name` and `committerdate`. I'm willing to add more in this PR if needed. What other sort keys do you want to see? What about naming?

I also want to note that in current implementation you can't do `--sort` multiple times as with `git branch --sort`, since  I don't have a strong use case for this feature and it complicates the code.

The next stage (next PR (?)) will be to add the sort option in the config. Use case for this I think is well explained in the related issue. Basically, one can set the option `sort=-committerdate` in the config and `jj bookmark list` will show bookmarks that were recently updated on top. Curious to hear your thoughts on ways how to implement this.

# Checklist

If applicable:


zummenix:
Overall, I think the following sort keys could be useful:
- `name`
- `authorname`
- `authoremail`
- `authortimestamp`
- `committername`
- `committeremail`
- `committertimestamp`

Regarding naming, I don't like the long `committertimestamp` and `authortimestamp` vs `committerdate` and `authordate` but it's consistent with that used in the template language.

Perhaps, separate with the dot `.` – `author.name`, `author.timestamp`, etc.?

cli/src/commands/bookmark/list.rs
```diff
@@ -106,6 +115,13 @@ pub struct BookmarkListArgs {
     ///     https://jj-vcs.github.io/jj/latest/templates/
     #[arg(long, short = 'T', add = ArgValueCandidates::new(complete::template_aliases))]
     template: Option<String>,
+
+    /// Sort bookmarks based on the given key. Prefix the key with `-` to sort
+    /// in descending order of the value (e.g. `--sort=-name`).
```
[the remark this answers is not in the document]
zummenix: I like the postfix notation (`name-`) and this works well with `ValueEnum`.

zummenix:
One concern I have regarding sorting as implemented right now is that it happens too early. I'm thinking perhaps better to move the step after the for loop where we print the output and where happens most of the filtering?

yuja:
> One concern I have regarding sorting as implemented right now is that it happens too early. I'm thinking perhaps better to move the step after the for loop where we print the output and where happens most of the filtering?

It seems the current implementation does sorting after filter predicates are applied. I think that's good, and I wouldn't mind if the sorting were made before filtering. The number of bookmarks should be small compared to the number of commits.

cli/src/commands/bookmark/list.rs
```diff
@@ -106,6 +113,11 @@ pub struct BookmarkListArgs {
     ///     https://jj-vcs.github.io/jj/latest/templates/
     #[arg(long, short = 'T', add = ArgValueCandidates::new(complete::template_aliases))]
     template: Option<String>,
+
+    /// Sort bookmarks based on the given key. Suffix the key with `-` to sort
+    /// in descending order of the value (e.g. `--sort name-`).
+    #[arg(long, value_name = "SORT_KEY", value_enum)]
+    sort: Option<SortKey>,
```
martinvonz, line 120: Should this be a list instead? I can imagine `--sort committeremail --sort committerdate-` being useful. Doesn't have to be in this PR. Oh, I see you mentioned this in the PR description already :)

I think it should be a very marginal increase in code complexity. Maybe just replacing this:
```rust
if let Some(sort_key) = &args.sort {
```
by this:
```rust
for sort_key in args.sort.iter().rev() {
```

zummenix: Yeah, I guess this should work :). For some reason I was thinking that it would be more involved...

zummenix:
This is ready for review 

(I have added updates what have changed in the first comment)

cli/tests/test_bookmark_command.rs
yuja, line 1: It's probably better to write jj-lib-like tests, but we don't have a good example in jj-cli. Setting up `Commit` objects would be a bit involved because it depends on the `Store`. I think `TestBackend::with_data()` can be used, but I haven't tried.

cli/src/commands/bookmark/list.rs
```diff
@@ -166,12 +178,47 @@ pub fn cmd_bookmark_list(
...
+                .is_none_or(|bookmark_names| bookmark_names.contains(name))
+                && (!args.conflicted || target.local_target.has_conflict())
+        })
+        .collect_vec();
+
+    // Multi-pass sorting, the first key is most significant.
+    for sort_key in args.sort.iter().rev() {
+        let is_reversed = sort_key.is_reversed();
+        match SortStrategy::from(*sort_key) {
+            SortStrategy::Name => {
```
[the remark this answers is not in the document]
zummenix: This statement is only correct for the first pass of the sorting, right? For example, if user wants `--sort name,authorname`, we first sort by `authorname` and then must sort by `name`. I guess, I can add a check if it is the first pass in ascending order and skip the sort.

cli/src/commands/bookmark/list.rs
```diff
@@ -166,12 +178,47 @@ pub fn cmd_bookmark_list(
...
+                bookmarks_to_list.sort_by(|(name_a, _), (name_b, _)| {
+                    let ordering = name_a.cmp(name_b);
+                    if is_reversed {
+                        ordering.reverse()
+                    } else {
+                        ordering
+                    }
+                });
```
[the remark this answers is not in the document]
zummenix: Could you please clarify the part about externalizing `is_reversed` handling?

The reason I've switched back to `sort_by` is that `sort_by_key` required to wrap the key in `std::cmp::Reverse` when `is_reversed` is true, and use `Either` type to unify return types (std::cmp::Reverse<&str> and &str).

cli/src/commands/bookmark/list.rs
```diff
@@ -244,3 +291,142 @@ pub fn cmd_bookmark_list(
...
+            bookmark
+                .remote_refs
+                .iter()
+                .filter_map(|(_, ref_target)| ref_target.target.as_normal()),
+        )
+        .collect();
+    commit_ids
+        .iter()
+        .filter_map(|commit_id| store.get_commit(commit_id).ok())
+        .collect_vec()
```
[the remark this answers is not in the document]
zummenix: In that case we need to clone `name` and `email` since `commit.author()` and `commit.committer()` return `&Signature`. Though, maybe optimiser is smart enough to know that `clone()` is not needed because `Commit` will be dropped in the same scope. I don't know...

cli/tests/test_bookmark_command.rs
zummenix: Do you mean instead of or additionally?

cli/src/commands/bookmark/list.rs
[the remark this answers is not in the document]
joyously: I would think that `--sort name,authorname` would be one pass on the `name` and only use `authorname` if there is equal values on the `name`.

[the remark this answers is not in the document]
zummenix: @joyously I don't have strong opinions on that but currently we iterate over the keys in reverse order to ensure that the most significant key (the first key in the keys slice) is applied last. Your intuition is correct here – in this example `name` is most significant key. Though, `man git-for-each-ref` mentions the following `You may use the --sort=<key> option multiple times, in which case the last key becomes the primary key.`. Basically, backwards as I understand it.

P.S. I noticed that I forgot to mention this in the doc comment for the `--sort` option. I'll do it later.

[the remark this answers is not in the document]
yuja: > This statement is only correct for the first pass of the sorting, right? 

Correct. It has to be a special case for `sort_keys == [name]`.

cli/src/commands/bookmark/list.rs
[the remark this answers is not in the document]
yuja: > Could you please clarify the part about externalizing `is_reversed` handling?

I just meant
```rust
if is_reversed {
    bookmarks_to_list.sort_by_key(... Reversed(...));
} else {
    bookmarks_to_list.sort_by_key(...);
}
```

It might make sense to add `sort_key.sort(&mut bookmarks_to_list)` to remove typing mess.

cli/src/commands/bookmark/list.rs
```diff
@@ -244,3 +291,142 @@ pub fn cmd_bookmark_list(
...
+        .local_target
+        .as_normal()
+        .into_iter()
+        .chain(
+            bookmark
+                .remote_refs
+                .iter()
+                .filter_map(|(_, ref_target)| ref_target.target.as_normal()),
+        )
+        .collect();
```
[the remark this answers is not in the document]
yuja: Deleted local bookmark doesn't have target commit. Diverged local bookmarks has multiple (negative/positive) target commits (including 0 positive commits.)

> A bookmark can have no local refs, but have some remote refs, right? If that the case, I think we need to include them to get the "proper" sort order

Since `jj bookmark list` primarily shows local bookmarks (and tracked remote bookmarks with indented), I expect the list would be sorted by local bookmark targets.

```
# sort the top-level items ["bar", "foo"] (by local targets)
bar
  # then the second-level items ["origin", "upstream"] ??
  @origin
  @upstream
foo
  @origin
  @upstream
...
```

Oh, but we have untracked remote bookmarks as top-level items. They'll have to be sorted independently. That would require non-trivial refactoring.

cli/src/commands/bookmark/list.rs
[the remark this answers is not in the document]
yuja: Optimizer wouldn't remove extra allocations, but that wouldn't be worse than collecting into `HashSet` and `Vec`. The number of target commits is usually 1.

cli/tests/test_bookmark_command.rs
yuja: Yes, "instead". Edge-case tests in jj-lib-like style + a few e2e tests.

cli/src/commands/bookmark/list.rs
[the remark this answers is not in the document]
yuja: > but we have untracked remote bookmarks as top-level items

https://github.com/jj-vcs/jj/compare/main...yuja:jj:push-rvtmmmyryyyu

cli/tests/test_bookmark_command.rs
zummenix: Ok, I'll look into it

cli/src/commands/bookmark/list.rs
```diff
@@ -244,3 +291,142 @@ pub fn cmd_bookmark_list(
...
+        }
+    }
+}
+
+/// Wrapper type to unify types for Ord.
+#[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
+enum CommitOrdValue<'a> {
+    Str(&'a str),
+    Timestamp(Timestamp),
+}
```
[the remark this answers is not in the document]
zummenix: I don't have much alternatives in my mind to using this unified type. So, do I just inline `bookmarks_to_list.sort_by_key(...)` in each branch of a sort_key?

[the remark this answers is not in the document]
yuja: > just inline `bookmarks_to_list.sort_by_key(...)` in each branch of a sort_key?

Yes. I assumed it would be simpler than erasing types. Maybe we can use declarative macro if the boilerplate code looked too much.

cli/tests/test_bookmark_command.rs
zummenix: @yuja I need guidance on this. Could you please provide an example of a test case (I think even pseudocode would help)?

yuja: I assumed something like `lib/tests/test_*.rs` could be added. These tests set up repository by using internal API and run functions to test. Since sorting functions are private, the tests to be added will have to be implemented in `bookmark/list.rs`.

https://github.com/jj-vcs/jj/blob/main/lib/tests/test_annotate.rs#L104

> I think TestBackend::with_data() can be used,

Oh, and I would probably mean that the test data could be set up in memory.

I'll take a look tomorrow to see if that's practical.

yuja:
```rust
#[cfg(test)]
mod tests {
    use jj_lib::backend::MillisSinceEpoch;
    use jj_lib::backend::Signature;
    use jj_lib::backend::Timestamp;
    use jj_lib::op_store::RefTarget;
    use testutils::TestRepo;

    use super::*;

    #[test]
    fn test_sort_by_name() {
        let test_repo = TestRepo::init();
        let signature = Signature {
            name: "Test User".to_owned(),
            email: "test.user@example.com".to_owned(),
            timestamp: Timestamp {
                timestamp: MillisSinceEpoch(0),
                tz_offset: 0,
            },
        };
        let repo = &test_repo.repo;
        let parents = vec![repo.store().root_commit_id().clone()];
        let tree_id = repo.store().empty_merged_tree_id();
        let commit1 = tx
            .repo_mut()
            .new_commit(parents, tree_id)
            //.set_author(signature.clone())
            //.set_committer(signature.clone())
            //.set_description("commit 1")
            .write()
            .unwrap();
        // XXX not needed if we construct Vec<CommitRef> manually
        // tx.repo_mut()
        //     .set_local_bookmark_target("foo", RefTarget::normal(commit1.id().clone()));
        // tx.repo_mut()
        //     .set_local_bookmark_target("bar", RefTarget::normal(commit2.id().clone()));
        // tx.repo_mut()
        //     .set_local_bookmark_target("baz", RefTarget::normal(commit3.id().clone()));
        let repo = tx.commit("set up").unwrap();

        // XXX extract helper function?
        let mut bookmark_items = vec![
            RefListItem {
                primary: CommitRef::local_only("bar", RefTarget::normal(commit1.id().clone())),
                tracked: vec![],
            },
            RefListItem {
                primary: CommitRef::local_only("baz", RefTarget::normal(commit1.id().clone())),
                tracked: vec![],
            },
            RefListItem {
                primary: CommitRef::local_only("foo", RefTarget::normal(commit1.id().clone())),
                tracked: vec![],
            },
        ];
        // XXX Oh, maybe we can instead pass `|commit_id| -> Commit` function
        // instead of repo.store()? Then, we wouldn't have to set up a real repo at all.
        sort(repo.store(), &mut bookmark_items, &[SortKey::NameDesc]); // function to test
        insta::assert_debug_snapshot!(
            // XXX add helper function that prints enough information?
            bookmark_items.iter().map(|item| item.primary.name()).collect_vec(), @r#"
        [
            "foo",
            "baz",
            "bar",
        ]
        "#);
    }
}
```

* It seemed uneasy to use `TestBackend::with_data()` directly because there are no convenient functions to add commits, etc. It's probably better to use higher-level API.
* Maybe we can instead stub out the function that maps `CommitId` to `Commit` or `backend::Commit`. (The latter is easier to construct.) Then, we don't have to instantiate a real `Repo` object at all. I haven't tried if that makes things simpler.

zummenix: Thank you!

I like the idea of using the function that maps `CommitId` to  `backend::Commit` and not instantiate a real `Repo` object in tests. Though, I see minor complications – `commit::Commit` currently has `pub fn store_commit(&self) -> &backend::Commit`, so we need to clone it to use it in `sort` which seems unacceptable to me. Perhaps add a complementary `pub fn take_store_commit(self) -> Arc<backend::Commit>`? A downside I see is that in tests we need to wrap `backend::Commit` in `Arc` when stubbing out data for a test. With the latter I envision the `sort` would look like this:

```rust
fn sort(
    bookmark_items: &mut Vec<RefListItem>,
    sort_keys: &[SortKey],
    f: impl Fn(CommitId) -> Option<Arc<backend::Commit>>,
) {
    todo!()
}
```

yuja: Perhaps, `store_commit()` can be changed to return `&Arc<backend::Commit>`. There would be no practical reason to hide the underlying type.

zummenix: > Perhaps, store_commit() can be changed to return &Arc<backend::Commit>. There would be no practical reason to hide the underlying type.

I did this change as a separate commit. Please let me know if you would like a separate PR for that.

cli/src/commands/bookmark/list.rs
```diff
@@ -270,3 +291,470 @@ struct RefListItem {
...
+        .cloned()
+        .and_then(f)
+}
+
+fn sort(
+    bookmark_items: &mut [RefListItem],
+    sort_keys: &[SortKey],
+    f: impl Fn(CommitId) -> Option<Arc<backend::Commit>>,
+) {
+    // Multi-pass sorting, the first key is most significant.
```
yuja, line 338: Since we need to propagate error, maybe we can simply build an external table prior to sorting?

```rust
let commits: HashMap<CommitId, Arc<backend::Commit>> = HashMap::new();
if sort_keys.iter().any(...) { // if key depends on commit
    commits = bookmark_items...try_collect()?;
}
let to_commit = |item: &RefListItem| {
    let id = item.primary.target().added_ids().next()?;
    commits.get(id)
};
```

It will also help remove `.clone()` from `.sort_by_key()`.

cli/src/commands/bookmark/list.rs
```diff
@@ -270,3 +291,470 @@ struct RefListItem {
...
+                                    format!(
+                                        "committer.timestamp: {}",
+                                        commit.committer.timestamp.timestamp.0
+                                    )
+                                })
+                                .unwrap_or_default()
+                        }
+                    })
+                    .join(", ")
+            })
```
yuja, line 581: nit: isn't it simpler to just dump all information unconditionally?

cli/src/commands/bookmark/list.rs
```diff
@@ -270,3 +291,470 @@ struct RefListItem {
...
+    f: impl Fn(CommitId) -> Option<Arc<backend::Commit>>,
+) {
+    // Multi-pass sorting, the first key is most significant.
+    for (i, sort_key) in sort_keys.iter().rev().enumerate() {
+        let f = &f;
+        match sort_key {
+            SortKey::Name => {
+                // Bookmarks are already sorted by name.
+                let is_noop = i == 0 && sort_key == &SortKey::Name;
+                if !is_noop {
```
yuja, line 345: nit: I would handle this first (out of the loop)

cli/src/commands/bookmark/list.rs
```diff
@@ -270,3 +291,470 @@ struct RefListItem {
...
+                    bookmark_items.sort_by_key(|item| item.primary.name().to_owned());
+                }
+            }
+            SortKey::NameDesc => {
+                bookmark_items.sort_by_key(|item| cmp::Reverse(item.primary.name().to_owned()));
```
yuja, line 350: nit: we'll need to sort by `(primary.name(), primary.remote_name())` to sort untracked remote bookmarks of the same name.

yuja (commented):
The implementation looks generally good, thanks.

lib/src/commit_builder.rs
```diff
@@ -204,7 +204,8 @@ impl DetachedCommitBuilder {
         predecessor: &Commit,
     ) -> Self {
         let store = repo.store().clone();
-        let mut commit = predecessor.store_commit().clone();
+        // TODO: probably better to make it work without `Arc::unwrap_or_clone`
+        let mut commit = Arc::unwrap_or_clone(predecessor.store_commit().clone());
```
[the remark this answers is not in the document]
zummenix: Ah, I erroneously thought that we can propagate `Arc<backend::Commit>` in `DetachedCommitBuilder`. But of course we can't mutate the commit after that...

cli/src/commands/bookmark/list.rs
```diff
@@ -107,6 +112,14 @@ pub struct BookmarkListArgs {
     ///     https://jj-vcs.github.io/jj/latest/templates/
     #[arg(long, short = 'T', add = ArgValueCandidates::new(complete::template_aliases))]
     template: Option<String>,
+
+    /// Sort bookmarks based on the given key (or multiple keys). Suffix the key
+    /// with `-` to sort in descending order of the value (e.g. `--sort name-`).
+    ///
+    /// Note that when using multiple keys, the first key is the most
+    /// significant.
```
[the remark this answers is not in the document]
zummenix: > period should be omitted in the summary line of command help.

Good to know, thanks! I also made sort keys for descending order hidden, so they don't appear in the command help.

cli/src/commands/bookmark/list.rs
```diff
@@ -219,6 +232,14 @@ pub fn cmd_bookmark_list(
         }
     }
 
+    let store = repo.store();
+    sort(&mut bookmark_list_items, &args.sort, |commit_id| {
+        store
+            .get_commit(&commit_id)
+            .ok()
```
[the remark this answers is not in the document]
zummenix: Ok, I'll take a look later this week. I don't know in what cases `get_commit` can fail and how jj can gracefully handle errors in context of sorting bookmarks.

cli/src/commands/bookmark/list.rs
zummenix: Yes, of course. My concern is that not all fields are relevant for a particular test and just add noise.

cli/src/commands/bookmark/list.rs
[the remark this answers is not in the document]
yuja: Can you check if completion works for hidden keys? (iirc, it doesn't in command-name context, but I don't remember the every detail.)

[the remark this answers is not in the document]
zummenix: No, completion does not work for hidden keys. Do you think it is important to complete also the descending order keys?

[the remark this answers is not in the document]
yuja: > No, completion does not work for hidden keys. Do you think it is important to complete also the descending order keys?

I'm okay with either way, but it might be bad for discoverability. People (including me) don't read help, and just press TAB key.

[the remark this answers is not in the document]
zummenix: > but it might be bad for discoverability

Ok, I reverted that.

cli/src/commands/bookmark/list.rs
```diff
@@ -277,3 +307,492 @@ struct RefListItem {
...
+        let bookmark_names = ["feature", "bug-fix", "chore", "bug-fix"];
+        let remote_names = [None, Some("upstream"), None, Some("origin")];
+        let mut bookmark_items: Vec<RefListItem> = Vec::new();
+        let mut commits: HashMap<CommitId, Arc<backend::Commit>> = HashMap::new();
+        for (((&name, &email), bookmark_name), remote_name) in names
+            .iter()
+            .zip(emails.iter())
+            .zip(bookmark_names.iter())
+            .zip(remote_names.iter())
+        {
```
yuja, line 496: nit: `itertools::izip!(&names, &emails, &bookmark_names, &remote_names)` can be used

cli/src/commands/bookmark/list.rs
```diff
@@ -277,3 +307,492 @@ struct RefListItem {
...
+            SortKey::CommitterDateDesc => bookmark_items.sort_by_key(|item| {
+                cmp::Reverse(to_commit(item).map(|commit| commit.committer.timestamp))
+            }),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
```
yuja, line 428: nit: redundant import

cli/src/commands/bookmark/list.rs
```diff
@@ -277,3 +307,492 @@ struct RefListItem {
...
+
+    use backend::ChangeId;
+    use backend::MergedTreeId;
+    use backend::TreeId;
```
yuja, line 432: nit: `use jj_lib::backend::...` for consistency

cli/src/commands/bookmark/list.rs
```diff
@@ -277,3 +307,492 @@ struct RefListItem {
...
+            if let Some(remote_name) = remote_name {
+                let local_target = RefTarget::normal(commit_id.clone());
+                let remote_ref = RemoteRef {
+                    target: local_target.clone(),
+                    state: RemoteRefState::New,
+                };
+                bookmark_items.push(RefListItem {
+                    primary: CommitRef::remote(b_name, *remote_name, remote_ref, &local_target),
+                    tracked: vec![],
+                });
```
yuja, line 545: nit: should be `CommitRef::remote_only()` (it wouldn't matter in tests, though.)

cli/src/commands/bookmark/list.rs
```diff
@@ -277,3 +307,492 @@ struct RefListItem {
...
+                                )
+                            })
+                            .unwrap_or_default(),
+                    })
+                    .join(", ")
+            })
+            .join("\n")
+    }
+
+    #[test]
```
yuja, line 623: nit: can you add some test sample with deleted bookmarks (= local absent, tracking remote present)?

cli/src/commands/bookmark/list.rs
```diff
@@ -277,3 +307,492 @@ struct RefListItem {
...
+                        SortKey::CommitterDate | SortKey::CommitterDateDesc => to_commit(item)
+                            .map(|commit| {
+                                format!(
+                                    "committer.timestamp: {}",
+                                    commit.committer.timestamp.timestamp.0
+                                )
+                            })
+                            .unwrap_or_default(),
+                    })
+                    .join(", ")
```
yuja, line 618: nit: I think it's simpler to just print all columns. The output is pretty readable if columns are aligned.

cli/tests/test_bookmark_command.rs
```diff
@@ -2212,6 +2212,56 @@ fn test_bookmark_list_conflicted() {
...
+#[test]
+fn test_bookmark_list_sort_multiple_keys() {
+    let test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let repo_path = test_env.env_root().join("repo");
+    for (bookmark, committeremail) in [("c", "bob@g.c"), ("b", "alice@g.c"), ("a", "bob@g.c")] {
+        test_env
+            .run_jj_in(
+                &repo_path,
+                [
```
yuja, line 2241: nit: can you update tests to use `work_dir.run_jj(args)`?

cli/tests/test_bookmark_command.rs
```diff
@@ -2212,6 +2212,56 @@ fn test_bookmark_list_conflicted() {
...
+            )
+            .success();
+        test_env
+            .run_jj_in(&repo_path, ["bookmark", "create", "-r@", bookmark])
+            .success();
+    }
+
+    let template =
+        r#"name ++ ": " ++ if(normal_target, normal_target.committer().email()) ++ "\n""#;
+    insta::assert_snapshot!(test_env.run_jj_in(&repo_path, ["bookmark", "list", "-T", template, "--sort", "committer-email", "--sort", "committer-date-"]), @r"
```
yuja, line 2257: nit: might be better to use `--sort=committer-email,committer-date-` because the comma-separated syntax is explicitly enabled here.

yuja (approved):
Thanks.

PR #5850, opened 2025-03-01
config: format divergent changes similarly to hidden commits

scott2000:
Alternative to #5800.

Since divergent changes don't necessarily always have to be resolved, it would be nice if they weren't displayed in a scary-looking format with red text. This commit changes the formatting of divergent changes to be almost identical to hidden commits, except with the text "divergent" instead of "hidden". This should allow users to make the connection that both of these cases prevent them from using the change ID to refer to the commit, but that both are "normal" states that a commit can be in (i.e. there is no error that must be resolved).

Before:

before

After:

after

# Checklist

If applicable:


ilyagr:
This looks pretty good to me (I haven't followed the relevant discussion or figured out what Martin's PR would look like, though). I like the idea of printing the word "divergent" so that people have a word to describe the situation by.

I don't think these are blockers, but I'm wondering about a couple of things.

I am not sure whether using the same colors as "hidden" commits is helpful or adds to the confusion. As Scott said, there is some similarity, but `jj log -r changeid` gives quite different results for hidden and for divergent ids. I'm hoping the word is enough to distinguish. 

I am wondering whether it makes sense to keep a `?` after the change id, since conflicted change ids are analogous to conflicted bookmarks. OTOH, as a counter-argument, the resolution of the situation is different than resolving conflicted bookmarks.

---------------

I'll wait for a potential discussion to happen and to think a bit more. If nothing happens, please remind me to remember to approve this.

ilyagr:
I also wonder whether there are places in the docs where we describe divergent change ids as red or as having question marks after them.

cli/src/config/templates.toml
```diff
@@ -274,13 +274,19 @@ separate(" ",
...
+  if(commit.divergent(), format_short_change_id_with_label(commit, "divergent")),
+  format_short_change_id(commit.change_id()),
+)
+'''
+
+'format_short_change_id_with_label(commit, label)' = '''
+label(label,
+  separate(" ",
+    format_short_change_id(commit.change_id()),
+    label,
```
ilyagr, line 288: Perhaps we should inline the function? It creates friction to making the color label different from the text we print, and I'm not sure it makes the code much prettier (or does it?).

yuja:
> I am not sure whether using the same colors as "hidden" commits is helpful or adds to the confusion. As Scott said, there is some similarity, but `jj log -r changeid` gives quite different results for hidden and for divergent ids. I'm hoping the word is enough to distinguish.

I would use the default color rather then the hidden color. It's a bit odd that resolvable ids are rendered as hidden. I also like the idea of rendering immutable divergence differently. https://github.com/jj-vcs/jj/pull/5800

ilyagr:
> I would use the default color rather then the hidden color.

Perhaps we could use a third color, e.g. yellow for the prefix? Cyan could be a more muted option. The advantage of using a non-default color is that it makes it clear to the user that it's some fact related to the the *change ids* that the word "divergent" relates to.

(No comment on immutable divergent commits; I don't think I've encountered any immutable ones and I would need to read the discussion to find out how they come up)

scott2000:
> I don't feel strongly, but perhaps we could use a third color, e.g. yellow for the prefix? Cyan could be a more muted option. The advantage of using a non-default color is that it makes it clear to the user that it's some fact related to the the _change ids_ that the word "divergent" relates to.

I had preferred yellow originally (since it seems like a good "warning" color), but I realized that it might look strange having yellow next to the author information, which is already yellow:

<details>
<summary>Screenshots of yellow, cyan, and blue</summary>

yellow

cyan

blue

</details>

Another option might be to keep the hidden color for the change ID, but use the default change ID color for the word "divergent":

<details>
<summary>Screenshot of hidden change ID with magenta "divergent"</summary>

hidden-magenta

</details>

> Another option: if we go with default colors for the change id (for example, if divergent commits are so common for some that using a third color creates too much color soup for many), we should make the word "divergent" purple (the same as the default color for the change id prefix). Perhaps a subtle hint like this is enough.

I actually think this looks really nice! This is probably my favorite now. I think this would also make it unnecessary to render it differently for immutable commits, because the word "divergent" just looks informational already.

magenta

Another option would be to go with red, and just hope that using the word "divergent" is enough to make users realize it's not a conflict. It does seem less scary that the current formatting, ~~but I still think it's a bit too scary-looking~~ (and we would definitely want to style immutable commits differently):

red

joyously:
I think the original is better than these other attempts. Using an uncommon word to explain the color simply draws more attention to it while also making it harder to line up the other fields into the rough "columns" that are there without a long word divergent.
Using the `?` conveys the information without disrupting the column flow. 

Red isn't scary; it's just different than yellow -- to most of us. But 7% of men can't tell them apart. Same for blue and purple.  The better default is to put the most contrasting colors next to each other, or just use two colors that alternate.

scott2000:
> I think the original is better than these other attempts. Using an uncommon word to explain the color simply draws more attention to it while also making it harder to line up the other fields into the rough "columns" that are there without a long word divergent. Using the `?` conveys the information without disrupting the column flow.

Perhaps we could put `divergent` and `hidden` at the end of the line, similar to `conflict`? This would even be an improvement over the current formatting for making the fields line up, since currently the `??` makes it slightly misaligned anyway. For instance:

yellow

magenta

yuja:
> Perhaps we could put `divergent` and `hidden` at the end of the line, similar to `conflict`?

iirc, the reason why `hidden` is there is that it's more related to change id than commit id. I personally don't mind if these labels are moved to right or to the next line, though.

rwjblue:
Just wanted drop my perspective here, I recently had a couple of cases where I had cause divergent commits and was *very* confused about the `??` in the output. It wasn't possible to search the JJ docs site or discord for `??`, and I had to resort to asking for help in discord (thank you @PhilipMetzger!!). Just having the name of the state in the output of either `jj st` or `jj log` would have been enough for me to find the quite good documentation for how to handle this state.

--

A few minor points RE: the change proposed here:

- Shouldn't `jj st` end up with the same visual representation as `jj log`? I'm not 100% sure about the internals to know if this PR changes both or just `jj log` (most of the discussion is about `jj log`)
- The red visual distinction that I have today RE: `??` was actually quite helpful, if it were simply grey or otherwise not "a bad color" I might not have paid much attention until I was further along with other changes (making figuring out which thing to abandon a bit harder)

tschundler:
+1 to the "??" is confusing and hard to search for.  It was for me in https://github.com/jj-vcs/jj/issues/6978

red is a "scary" color - it means to me something is amiss. It it weren't red, I'd probably have ignored the "??". So https://github.com/jj-vcs/jj/pull/5800 to selectively get rid of the red when it doesn't really matter LGTM.

I do feel that it's best for the word "divergent" to be at the end near the commit ID, mostly because that's an extra clue to what divergent means - the commit is what is different.

scott2000:
Now that using a divergent change ID is always an error unless `change_id()` is used, I'm leaning more towards the following format, which leaves the change ID red to show that it is an error, but removes the underline and "??" in favor of the more easily-searchable term "divergent":

after

I'm still not sure whether "divergent" should be next to the change ID (to make it clear that it's a property of the change ID), or at the end (to avoid breaking the alignment of fields in the log).

glehmann:
> I'm still now sure whether "divergent" should be next to the change ID (to make it clear that it's a property of the change ID), or at the end (to avoid breaking the alignment of fields in the log).

FWIW, I like this last version. The `divergent` next to the change-id clearly shows it's linked to the change-id

scott2000:
There is also another alternative to this PR if we end up adding a new revset syntax to disambiguate divergent change IDs: https://github.com/jj-vcs/jj/issues/7028#issuecomment-3096876933

scott2000:
I'm closing this PR since #7194 was merged. We can discuss adding the "divergent" label in #8322 instead.

PR #1913, opened 2023-07-26
cargo: specify MSRV in rust-toolchain.toml

thoughtpolice:
Summary: Split out from from #1851. This specifies the current project-wide MSRV in a `rust-toolchain.toml` file which — if installed — will cause 'rustup' to download and provision the correct version of the rust compiler.

The intent is to ensure most developers stay on the same page and people doing occasional development don't accidentally trip over themselves and submit something based on their own toolchain i.e. newer or older features than our MSRV.

This also migrates the Nix Flake to use `rust-toolchain.toml` and so it creates a single source of truth for both `nix develop` and `rustup`.

Change-Id: I56e48d399e541d7700c7e9652ac88c57

# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #5851, opened 2025-03-01
docs: add explanation of how to check MSRV locally

bryceberger:
There have been a few attempts to add a `rust-toolchain.toml` file to pin developers' rust version. The most recent attempt was [reverted].

The current opinion ([comment]) is that using a newer compiler version is desired, and MSRV violations should be caught by `clippy::incompatible_msrv`. This commit writes that opinion in the `docs/contributing.md`.

[reverted]: https://github.com/jj-vcs/jj/pull/5090

[comment]: https://github.com/jj-vcs/jj/pull/1913#issuecomment-2037636332k

# Checklist

If applicable:


ilyagr:
I'm a little confused by your "Note". What's the story there? Is there no way to enforce that lint in tests by more regular means?

I was thinking of updating our MSRV to 1.84 (IIRC, maybe it was 1.85) for the MSRV-aware cargo resolver, btw (and also in preparation for some of our dependencies switching to 2024 Rust edition and requiring 1.85, which I'm guessing will happen). Enforcing the `incompatible-msrv` lint would seem like a good reason to update the MSRV as well (and you're saying 1.78 is sufficient for that, though my impression is that people are fine with moving straight to 1.84 or something).

docs/contributing.md
ilyagr, line 1: The commit description has a minor but confusing typo in the link.

```
You have:
https://github.com/jj-vcs/jj/pull/1913#issuecomment-2037636332k
The k at the end is wrong, should be
https://github.com/jj-vcs/jj/pull/1913#issuecomment-2037636332
```

bryceberger:
> Is `incompatible-msrv` not enabled by default in newer versions?

It is, but explicitly giving the name of the lint tells what to look out for and will give a warning if they're using a version that doesn't support the lint.

> Is there no way to enforce that lint in tests by more regular means?

Unfortunately, I don't think there is. https://github.com/rust-lang/rust-clippy/issues/12257 argues that tests shouldn't be checked (since they're not required to use as a library). https://github.com/rust-lang/rust-clippy/pull/14279 was just merged, which allows it to run in test code again, but requires a config option.

MSRV-aware resolver is 1.84, edition2024 is 1.85.

In terms of updating MSRV, I know it's been said somewhere that we follow helix, which in turn follows firefox. Currently FF is still on 1.76, but will be on 1.82 in April (https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html).

docs/contributing.md
bryceberger: Editor mishap lol

ilyagr:
I would prefer upgrading our MSRV to documenting the `incompatible-msrv` lint, especially since a relatively small upgrade will do. We might still need to document the issue with tests, though it also seems less important to me.

I personally would also make a larger jump in MSRV. Unlike helix/firefox, we are not packaged in conservative distros yet. Once we are, it might make sense to become as conservative as them, which might also create a headache when our dependencies have a more aggressive MSRV. (BTW, I meant "debian testing" in the above comment, fixed now, pretty significant typo).

PR #5852, opened 2025-03-01
repo: MutableRepo::transform_descendants_with_options: add `new_parents_map`

bnjmnt4n:
The `new_parents_map` will allow the parents of each commit to be
customized before the `transform_descendents` callback is invoked. This
is useful when the order of commits needs to be changed, whereby setting
the new parents in the default callback might lead to repeated rebasing
and cycles if the new parent has not been visited in the reverse
topological order traversal.

# Checklist

If applicable:

### DIFF
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1178,32 +1178,33 @@ impl MutableRepo {
     }
 
     /// Find descendants of `root`, unless they've already been rewritten
-    /// (according to `parent_mapping`), and then return them in
-    /// an order they should be rebased in. The result is in reverse order
-    /// so the next value can be removed from the end.
-    fn find_descendants_to_rebase(&self, roots: Vec<CommitId>) -> BackendResult<Vec<Commit>> {
-        let store = self.store();
-        let to_visit_expression =
-            RevsetExpression::commits(roots)
-                .descendants()
-                .minus(&RevsetExpression::commits(
-                    self.parent_mapping.keys().cloned().collect(),
-                ));
-        let to_visit_revset = to_visit_expression
+    /// (according to `parent_mapping`).
+    pub fn find_descendants_for_rebase(&self, roots: Vec<CommitId>) -> BackendResult<Vec<Commit>> {
+        let to_visit_revset = RevsetExpression::commits(roots)
+            .descendants()
+            .minus(&RevsetExpression::commits(
+                self.parent_mapping.keys().cloned().collect(),
+            ))
             .evaluate(self)
             .map_err(|err| err.expect_backend_error())?;
-        let to_visit: Vec<_> = to_visit_revset
+        let to_visit = to_visit_revset
             .iter()
-            .commits(store)
+            .commits(self.store())
             .try_collect()
             // TODO: Return evaluation error to caller
             .map_err(|err| err.expect_backend_error())?;
-        drop(to_visit_revset);
+        Ok(to_visit)
+    }
+
+    /// Order a set of commits in an order they should be rebased in. The result
+    /// is in reverse order so the next value can be removed from the end.
+    fn order_commits_for_rebase(&self, to_visit: Vec<Commit>) -> BackendResult<Vec<Commit>> {
         let to_visit_set: HashSet<CommitId> =
             to_visit.iter().map(|commit| commit.id().clone()).collect();
         let mut visited = HashSet::new();
         // Calculate an order where we rebase parents first, but if the parents were
         // rewritten, make sure we rebase the rewritten parent first.
+        let store = self.store();
         dag_walk::topo_order_reverse_ok(
             to_visit.into_iter().map(Ok),
             |commit| commit.id().clone(),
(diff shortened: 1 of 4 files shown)


lib/tests/test_rewrite_transform.rs
```diff
@@ -120,3 +122,89 @@ fn test_transform_descendants_sync_linearize_merge() {
...
+                                if parent_id == commit_c.id() {
+                                    commit_b.id().clone()
+                                } else {
+                                    parent_id.clone()
+                                }
+                            })
+                            .collect_vec(),
+                    )
+                }
+            },
```
yuja, line 179: Random thought: Does it make sense to pass `HashMap<CommitId, Vec<CommitId>>` to just reorder the `to_visit` stack? In this example, `{C: [A], B: [C]}` would be passed as parent overrides, and the last `else` block would be handled by the rewrite callback.

My feeling is that the callback API is harder to do things right, but I have no idea if overrides map will help simplify API surface.

bnjmnt4n: Yes, I was considering that too, but didn't try it out yet. I'll try to create a version which uses a map instead.

I do agree that the callback is quite messy, and might be error-prone.

bnjmnt4n: I've updated this PR to accept a `new_parents_map`, which might make it more convenient to construct the lsit of new parents.

lib/src/rewrite.rs
```diff
@@ -190,6 +190,24 @@ impl<'repo> CommitRewriter<'repo> {
         }
     }
 
+    /// Update the intended new parents by replacing any occurrence of
+    /// `old_parent` by `unrewritten_parents`.
+    pub fn replace_rewritten_parent<'a>(
```
yuja, line 194: nit: are we going to add non-test callers of this function?

lib/src/rewrite.rs
```diff
@@ -190,6 +190,24 @@ impl<'repo> CommitRewriter<'repo> {
...
+        &mut self,
+        old_parent: &CommitId,
+        unrewritten_parents: impl IntoIterator<Item = &'a CommitId>,
+    ) {
+        let old_parent = &self.mut_repo.new_parents(&[old_parent.clone()])[0];
```
yuja, line 199: nit: `slice::from_ref(old_parent)`

yuja, line 199: Can we assume that `new_parents()` never returns more than one parent? It might be safer to loop over "new" old parents, but I'm not pretty sure.

lib/src/rewrite.rs
```diff
@@ -190,6 +190,24 @@ impl<'repo> CommitRewriter<'repo> {
...
+        let new_parents = self
+            .mut_repo
+            .new_parents(&unrewritten_parents.into_iter().cloned().collect_vec());
+        if let Some(i) = self.new_parents.iter().position(|p| p == old_parent) {
+            self.new_parents.splice(i..i + 1, new_parents);
+            let mut unique = HashSet::new();
+            self.new_parents.retain(|p| unique.insert(p.clone()));
+        }
```
yuja, line 207: nit: can be forwarded to `self.replace_parent()`?

lib/src/repo.rs
```diff
@@ -1263,20 +1264,25 @@ impl MutableRepo {
...
     ///
+    /// If a commit is in the `new_parents_map` is provided, it will be rebased
+    /// onto the new parents provided in the map instead of its original
+    /// parents.
+    ///
     /// See [`Self::transform_descendants()`] for details.
     pub fn transform_descendants_with_options(
         &mut self,
         roots: Vec<CommitId>,
+        new_parents_map: HashMap<CommitId, Vec<CommitId>>,
```
yuja, line 1280: nit: pass `new_parents_map` by reference

lib/src/repo.rs
```diff
@@ -1194,20 +1194,30 @@ impl MutableRepo {
...
             |commit| commit.id().clone(),
             |commit| -> Vec<BackendResult<Commit>> {
                 visited.insert(commit.id().clone());
                 let mut dependents = vec![];
-                for parent in commit.parents() {
+                let parent_ids = commit.parent_ids();
+                parents_map.insert(
+                    commit.id().clone(),
+                    parent_ids.iter().cloned().collect_vec(),
+                );
```
yuja: Since we now have a static parents mapping, I don't think we need to build `(commit, parents)` pairs here. We can instead re-look up parents by `transform_descendants()`.

lib/src/repo.rs
```diff
@@ -1279,10 +1279,25 @@ impl MutableRepo {
...
-        let mut to_visit = self.order_commits_for_rebase(descendants, new_parents_map)?;
+        self.transform_commits(descendants, new_parents_map, options, callback)
+    }
+
+    /// Rewrite the given commits in reverse topological order.
+    ///
+    /// This function is similar to
+    /// [`Self::transform_descendants_with_options()`], but only rewrites the
+    /// `commits` provided, and does not rewrite their descendants.
+    pub fn transform_commits(
```
yuja, line 1293: This appears to be unused in this PR?

lib/src/rewrite.rs
bnjmnt4n: I've removed this function since it's only used in the tests.

lib/src/repo.rs
bnjmnt4n: Removed the function.

lib/src/rewrite.rs
bnjmnt4n: Removed the function.

lib/src/repo.rs
bnjmnt4n: I think I forgot to update the function to be used. Fixed.

yuja (approved):
I'm not familiar with the rewrite code, but the change looks good to me.

(As a follow up) it would be nice if `move_commits()` were split to smaller functions. It's unclear where the mutation starts, for example.

bnjmnt4n:
Addressed the latest batch of feedback. Will merge this after v0.28.1 is released if there are no other reviews.

PR #5854, opened 2025-03-02
ci: make sure tests fail in CI if `gpg` or `taplo` binaries are not found

### LINKED ISSUE #5696: FR: fail tests if taplo/gnupg are not found in the CI environment
In #5692, we added schema tests that depend on the taplo CLI. These were modeled after the gnupg tests. In both cases, the tests are set up to automatically pass if the required binaries are not available. However, we expect these to always be present in the CI environment. To avoid silently passing these tests in the event that the CI environment breaks, we should make both sets of tests fail if the CI environment is detected but the binaries are not present.

ilyagr:
Fixes #5696

Example failure: https://github.com/jj-vcs/jj/actions/runs/13611152082/job/38048476201?pr=5854#step:7:1543

(This can wait; since the CI passes on the PR, we know the problem this PR aims to detect is not occurring, at least at the commit the PR is synced to).

### DIFF
--- a/lib/testutils/src/lib.rs
+++ b/lib/testutils/src/lib.rs
@@ -135,6 +135,16 @@ pub fn user_settings() -> UserSettings {
     UserSettings::from_config(base_user_config()).unwrap()
 }
 
+/// Panic if `CI` environment variable is set to a non-empty value
+///
+/// Most CI environments set this variable automatically. See e.g.
+/// <https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables>
+#[track_caller]
+pub fn ensure_running_outside_ci(reason: &str) {
+    let running_in_ci = std::env::var("CI").is_ok_and(|value| !value.is_empty());
+    assert!(!running_in_ci, "Running in CI, {reason}.");
+}
+
 #[derive(Debug)]
 pub struct TestEnvironment {
     temp_dir: TempDir,
--- a/cli/tests/test_config_schema.rs
+++ b/cli/tests/test_config_schema.rs
@@ -1,6 +1,8 @@
 use std::process::Command;
 use std::process::Stdio;
 
+use testutils::ensure_running_outside_ci;
+
 fn taplo_check_config(file: &str) {
     if Command::new("taplo")
         .arg("--version")
@@ -8,6 +10,7 @@ fn taplo_check_config(file: &str) {
         .status()
         .is_err()
     {
+        ensure_running_outside_ci("`taplo` must be in the PATH");
         eprintln!("Skipping test because taplo is not installed on the system");
         return;
     }
--- a/lib/tests/test_gpg.rs
+++ b/lib/tests/test_gpg.rs
@@ -12,6 +12,7 @@ use jj_lib::gpg_signing::GpgBackend;
 use jj_lib::signing::SigStatus;
 use jj_lib::signing::SignError;
 use jj_lib::signing::SigningBackend as _;
+use testutils::ensure_running_outside_ci;
 
 static PRIVATE_KEY: &str = r#"-----BEGIN PGP PRIVATE KEY BLOCK-----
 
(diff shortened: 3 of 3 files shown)


lib/testutils/src/lib.rs
```diff
@@ -142,6 +142,12 @@ impl TestEnvironment {
         }
     }
 
+    /// Panic if `CI` environment variable is set to a non-empty value
+    pub fn ensure_running_outside_ci(reason: &str) {
```
yuja, line 146: nit: move to module-level function because this does nothing related to `TestEnvironment`.

lib/testutils/src/lib.rs
```diff
@@ -142,6 +142,12 @@ impl TestEnvironment {
...
+        let running_in_ci = std::env::var("CI").is_ok_and(|value| !value.is_empty());
+        assert!(!running_in_ci, "Running in CI, {reason}.");
```
yuja, line 148: nit: I think it's better to add `JJ_TEST_FORCE_GPG` or something. It might make sense for down-stream packagers to enable this flag.

ilyagr: I didn't follow the situation that you have in mind. My first thought is that if we create such a variable, this could be changed to fail if either CI or JJ_GPG_... is defined.

Or do you want us to use some variable we manually set in our CI workflows?

yuja: It's just about naming. I think explicitly-named variable is better than `CI`.

ilyagr: CI is kind of a standard. I think most CI runners set it by default, I've seen it in GitHub and Sourcehut, I think.


Also, do we have a need of separate variables for `taplo` and `gpg`?

yuja: > CI is kind of a standard. I think most CI runners set it by default

Oh, I didn't know that and assumed we would set `CI=1` somewhere.

> do we have a need of separate variables for taplo and gpg?

Downstream packagers might want to forcibly enable `gpg` tests, but not `taplo`? That wouldn't be a big deal if `CI` is the standard way to handle this kind of things, though.

ilyagr: It's less well-documented than I hoped. The best I found so far is https://github.com/actions/runner/pull/374/files.

There are also crates like https://github.com/zkat/is_ci, https://crates.io/crates/ci_info, <https://docs.rs/built/0.7.7/built/util/fn.detect_ci.html>. Might be a bit of an overkill.

ilyagr: My current plan is to go with CI, but add a comment explaining it. Let me know if you prefer something else.

In my eyes, the main problem with this is that it's not obvious whether, if tests succeeded, they ran in CI mode or not. I'm not sure whether there's a simple solution to this, though.

An alternative idea I like in principle would be to run tests with something like `--feature ci`. However, I'm not familiar enough with Cargo to know whether features that are relevant only for tests are supported or a good idea.

ilyagr: Emily found some docs: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

I think I'll put that link in a comment. Emily also says that merging this would have helped #6058, so perhaps we should merge this.

lib/testutils/src/lib.rs
```diff
@@ -135,6 +135,15 @@ pub fn user_settings() -> UserSettings {
     UserSettings::from_config(base_user_config()).unwrap()
 }
 
+/// Panic if `CI` environment variable is set to a non-empty value
+///
+/// Most CI environments set this variable automatically. See e.g.
+/// <https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables>
+pub fn ensure_running_outside_ci(reason: &str) {
```
yuja, line 142: nit: add `#[track_caller]`?

yuja (approved):

lib/testutils/src/lib.rs
ilyagr: TIL, thanks, nice!

PR #5868, opened 2025-03-04
templates: expose `format_short_signature_oneline` template alias hook

jakobhellermann:
For `builtin_log_compact` you can specify how you want signatures to be displayed via `format_short_signature`, but the same wasn't possible for `builtin_log_compact_oneline` which always chose `email().local()`.

Is this worth a changelog entry? 

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -91,7 +91,7 @@ if(root,
     concat(
       separate(" ",
         format_short_change_id_with_hidden_and_divergent_info(self),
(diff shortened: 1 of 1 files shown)


ilyagr:
I'm leaning towards taking this as-is. Will wait a bit to give others a moment.

It would leave this undocumented (`format_short_signature` is documented in `config.toml`), but perhaps it's not important enough to document/test. 

I don't think it needs a Changelog entry.

ilyagr (approved):

PR #5873, opened 2025-03-04
lib/testutils/Cargo.toml: remove unused gix feature

jonathantanmy:
@bsdinis I think that we are moving towards doing network operations with system git, not with gix, so removing this feature seems safe. Let me know if you (or others) think that we will need this feature eventually and therefore should retain it (although in that case, I would suggest that we use the async client instead of the blocking client).

Commit description:

The gix feature "blocking-network-client" was configured in cd6141693 (cli/tests: add gitoxide helpers, 2025-02-03) most likely because we needed to clone using gix in tests, but 071e724c1 (cli/tests: move test_git_colocated_fetch_deleted_or_moved_bookmark to gitoxide, 2025-02-25) changed the test to clone by subprocessing out to system git (not by using gitoxide, as a cursory read of the commit description may indicate), meaning that we no longer need that feature.

Therefore, remove that feature.

# Checklist

If applicable:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1097,7 +1097,6 @@ dependencies = [
  "gix-command",
  "gix-commitgraph",
  "gix-config",
- "gix-credentials",
  "gix-date",
  "gix-diff",
  "gix-dir",
@@ -1111,13 +1110,11 @@ dependencies = [
  "gix-ignore",
  "gix-index",
  "gix-lock",
- "gix-negotiate",
  "gix-object",
  "gix-odb",
  "gix-pack",
  "gix-path",
  "gix-pathspec",
- "gix-prompt",
  "gix-protocol",
  "gix-ref",
  "gix-refspec",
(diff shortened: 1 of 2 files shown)


bsdinis (approved):
Hi, thank you for catching this.

In fact, the blocking-network-client feature is going to be used, at least in the port of the remaining tests to gitoxide. But it should live in the dev-dependencies of `cli`, not in `testutils`

jonathantanmy:
Thanks for taking a look. I still think that if production code is going to subprocess to git for network operations, then the tests should as well (so no need for any sort of network client feature, whether blocking or not), but we can discuss this when the PR to add those tests is created.

bsdinis:
> Thanks for taking a look. I still think that if production code is going to subprocess to git for network operations, then the tests should as well (so no need for any sort of network client feature, whether blocking or not), but we can discuss this when the PR to add those tests is created.

It is open: https://github.com/jj-vcs/jj/pull/5846

Note that here it's not really network operations, it's fetching on the filesystem.

PR #5875, opened 2025-03-04
tests: avoid using the local backend

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_workspaces.rs
+++ b/cli/tests/test_workspaces.rs
@@ -798,10 +798,7 @@ fn test_workspaces_current_op_discarded_by_other(automatic: bool) {
         test_env.add_config("[snapshot]\nauto-update-stale = true\n");
     }
 
-    // Use the local backend because GitBackend::gc() depends on the git CLI.
-    test_env
-        .run_jj_in(".", ["init", "main", "--config=ui.allow-init-native=true"])
-        .success();
+    test_env.run_jj_in(".", ["git", "init", "main"]).success();
     let main_path = test_env.env_root().join("main");
     let secondary_path = test_env.env_root().join("secondary");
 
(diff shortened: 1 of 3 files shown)


cli/tests/test_workspaces.rs
```diff
@@ -944,11 +941,11 @@ fn test_workspaces_current_op_discarded_by_other(automatic: bool) {
     insta::allow_duplicates! {
         insta::assert_snapshot!(output, @r"
         Working copy changes:
-        A added
+        C {modified => added}
```
ilyagr, line 944: Is this expected?

cli/tests/test_workspaces.rs
```diff
@@ -887,15 +884,15 @@ fn test_workspaces_current_op_discarded_by_other(automatic: bool) {
         let output = test_env.run_jj_in(&secondary_path, ["st"]);
         insta::assert_snapshot!(output, @r"
         Working copy changes:
-        A added
+        C {modified => added}
```
ilyagr, line 887: Is this expected?

cli/tests/test_workspaces.rs
martinvonz: Yes, it's because the Git backend supports detecting copies (the local backend does)

cli/tests/test_workspaces.rs
martinvonz: Same as above

ilyagr:
Thanks!

As long as people think that this is obviously correct to merge now, it sounds great to me.

This PR does introduce some test changes where it's not immediately obvious to me whether they are expected. Most likely, it's just the fact that the toy backend does not track renames, but the diffs do look weird to me.

cli/tests/test_workspaces.rs
ilyagr: Is it normal that `A` was mentioned before and now is not?

ilyagr (approved):
Approving this in expectation that the answer to <https://github.com/jj-vcs/jj/pull/5875/files#r1980307384> is "yes".

cli/tests/test_workspaces.rs
martinvonz: Yes, `C` implies that a file was created/added. We never (yet, at least) indicate a copy to a file that existed before, so no "copy into existing file".

ilyagr:
This didn't touch `test_global_opts`. I'm guessing that wasn't intentional? I think it's in my second commit.

martinvonz:
Sorry I missed that. I searched for `allow-init-native` and that one apparently ran tests that failed too early for that to matter. Thanks for taking care of it in your PR.

PR #5876, opened 2025-03-05
debug init-local: fixups, docs,remove the `ui.allow-init-native` option

ilyagr:
Not sure whether I should merge this before or after the release, feel free to tell me.

---------------

It is not needed to gate `jj debug init-local` on an option now that the command is no longer called `jj init`.

This also includes a separate fixup to #5845 in demos/

# Checklist

If applicable:
- [ ] ~~I have updated `CHANGELOG.md`~~ (I think it's too minor)

### DIFF
--- a/cli/src/commands/debug/init_local.rs
+++ b/cli/src/commands/debug/init_local.rs
@@ -25,10 +25,15 @@ use crate::command_error::user_error_with_message;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
-/// Create a new repo in the given directory
+/// Create a new repo in the given directory using the proof-of-concept native
+/// backend
 ///
-/// If the given directory does not exist, it will be created. If no directory
-/// is given, the current directory is used.
+/// The proof-of-concept backend is called "local" because it does not support
+/// cloning, pushing, or fetching repositories or commits.
+///
+/// This command is otherwise analogous to `jj git init`. If the given directory
+/// does not exist, it will be created. If no directory is given, the current
+/// directory is used.
 #[derive(clap::Args, Clone, Debug)]
 pub(crate) struct DebugInitLocalArgs {
     /// The destination directory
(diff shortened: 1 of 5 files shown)


cli/src/commands/debug/init_local.rs
```diff
@@ -20,15 +20,19 @@ use tracing::instrument;
...
 use crate::ui::Ui;
 
-/// Create a new repo in the given directory
+/// Create a new repo in the given directory using the proof-of-concept native
+/// backend
 ///
-/// If the given directory does not exist, it will be created. If no directory
-/// is given, the current directory is used.
+/// The proof-of-concept backend is called "local" because it does not support
+/// cloning, pushing, or fetching repositories or commits.
```
yuja, line 31: I have no idea if this is true.

yuja (approved):
Just merge before the release? I don't think this change would break something.

cli/src/commands/debug/init_local.rs
ilyagr: The alternative would be to say

> This backend does not currently support
> cloning, pushing, or fetching repositories or commits.

I'm fine with either.

ilyagr: I'll just go with this, we can change it later.

cli/src/commands/debug/init_local.rs
```diff
@@ -25,10 +25,15 @@ use crate::command_error::user_error_with_message;
...
```
martinvonz, line 31: It's called "local" because it stores commits locally (unlike the Google backend), but it doesn't really matter since almost no one will read this anyway :)

ilyagr: That'd make it confusable with the `git` backend, so clearly it's *now* named the "local" backend for another reason. 😛

PR #5885, opened 2025-03-06
contributing.md: describe MSRV guidelines

ilyagr:
This is inspired by <https://github.com/jj-vcs/jj/pull/5851>, and borrows a bit from that PR. Thank you, @bryceberger! Unlike #5851, I do not discuss the issue of `incompatible_msrv` in tests, so this doesn't replace that PR entirely.

See also <https://github.com/jj-vcs/jj/pull/5851#issuecomment-2692408627>, whether @bryceberger provides useful context.

This PR is a question on whether we can agree on my personal favorite MSRV guidelines. If not, that is fine, and I'll drop the PR. I don't think an MSRV policy is urgently needed, I'm just suggesting one in case we already more or less agree.

As I mentioned in the linked discussion, I also think it'd be nice to upgrade our MSRV to 1.84 soon, so that we can make use of the MSRV-aware resolver. Again, this is a "nice-to-have" that we can do without. If we want to following Firefox MSRV, we would still upgrade to 1.82 this release.

PR #5887, opened 2025-03-06
Update MSRV to 1.84 and turn on MSRV-aware Cargo resolver

ilyagr:
See also #5885

Other options would be to update to 1.82 (conservatively, following Firefox, no MSRV-aware resolver), 1.85 (aggressive, not quite yet in nixpkgs or stable in Gentoo, though should be relatively soon), or to do nothing.

I'm guessing (but not sure) that at least one of our dependencies will require 1.85 in the near future for the 2024 Rust edition. If we use the MSRV-aware resolver, that will almost certainly not be a problem for us.

------------

For reference, we last updated the MSRV to 1.76 in https://github.com/jj-vcs/jj/commit/5b517b542ec6c88357adf92369dda7668131da72. According to https://releases.rs/docs/1.76.0/, this was when it barely became stable.

-------

Re some Discord discussion:

I feel strongly that we should *not* upgrade to Rust 2024 at the same time as we upgrade the MSRV. I tried the auto-upgrade, and it's a bit messy. So, to me it matters little that 1.84 does not support 2024 Rust. I also think that if we try auto-conversion to Rust 2024 later, it might be smarter and/or more polished, since others will have tried it and would have reported bugs. (Though I'm not opposed to somebody else doing the 2024 conversion whenever).

Upgrading 1.84 -> 1.85 in a month or so should be nearly trivial, I think. Most of the work here was in changing lints from `allow` to `expect`. If I change the MSRV to 1.85, running `cargo clippy --workspace --all-targets --fix` results in no changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
     - name: Install Rust
       uses: dtolnay/rust-toolchain@a54c7afa936fefeb4456b2dd8068152669aa8203
       with:
-        toolchain: 1.76
+        toolchain: 1.84
         target: ${{ matrix.target }}
     - uses: taiki-e/install-action@970d55e3ce02a46d60ffae7b4fab3dedace6e769
       with:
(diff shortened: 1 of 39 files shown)


martinvonz (approved):

ilyagr:
Thank you, Martin! I decided to merge it quickly, since it's a PR that could be difficult to rebase. Two of the packagers in our Discord community seemed fine with it.

martinvonz:
Sounds good. We can always roll it back if there are objections tomorrow.

PR #5891, opened 2025-03-06
cli: Print @ and @- in status and update_working_copy

### LINKED ISSUE #4700: FR: Show `@` in listed commits after `jj commit`

After doing a `jj commit`, the output shows 2 commits, like this:

```
$ jj commit -i
Working copy now at: mnnxpsqr 4a8c5c20 (no description set)
Parent commit      : szvnmmku 19143488 Test commit
```

I'm still very new to jj, and what I sometimes want to do after the commit is to set a bookmark, so I do:

```
jj bookmark set mybookmark -r @-
```

However, I've found myself to be not 100% sure that `@-` refers to the right commit there.

I think it would help if the output indicated which one is `@` (and maybe even which one is `@-`), e.g. like this:

```
$ jj commit -i
Working copy now at: @  mnnxpsqr 4a8c5c20 (no description set)
Parent commit      : @- szvnmmku 19143488 Test commit
```

n/a

**Additional context**
n/a

robinst:
This helps newcomers remember what `@` and `@-` stand for. Fixes #4700.

`jj st` before:

Screenshot 2025-03-06 at 8 26 02 pm

```
Working copy : srzktwpv 0eeeada1 (empty) (no description set)
Parent commit: usotzrkt 552b98c3 Print @ and @- in status and update_working_copy
```

after:

Screenshot 2025-03-06 at 8 26 08 pm

```
Working copy  @  srzktwpv 0eeeada1 (empty) (no description set)
Parent commit @- usotzrkt 552b98c3 Print @ and @- in status and update_working_copy
```

And the output after e.g. a `jj commit`:

Screenshot 2025-03-06 at 8 17 25 pm

Note that I've removed the `:` because:
- I feel like the `@` symbols and whitespace provide enough separation, also having `:` there makes it a bit noisy
- It doesn't take up as much horizontal space

# Checklist

If applicable:

(Not doing^ yet until the format is agreed.)

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2023,13 +2023,18 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
         if Some(new_commit) != maybe_old_commit {
             if let Some(mut formatter) = ui.status_formatter() {
                 let template = self.commit_summary_template();
-                write!(formatter, "Working copy now at: ")?;
+                write!(formatter, "Working copy now ")?;
+                formatter.with_label("node", |fmt| write!(fmt.labeled("working_copy"), "@"))?;
+                write!(formatter, "  ")?;
                 formatter.with_label("working_copy", |fmt| template.format(new_commit, fmt))?;
                 writeln!(formatter)?;
                 for parent in new_commit.parents() {
                     let parent = parent?;
-                    //                "Working copy now at: "
-                    write!(formatter, "Parent commit      : ")?;
+                    //                "Working copy now @  "
+                    //                "Parent commit    @- "
+                    write!(formatter, "Parent commit    ")?;
+                    write!(formatter.labeled("node"), "@-")?;
+                    write!(formatter, " ")?;
                     template.format(&parent, formatter.as_mut())?;
                     writeln!(formatter)?;
                 }
--- a/cli/src/commands/status.rs
+++ b/cli/src/commands/status.rs
@@ -113,12 +113,16 @@ pub(crate) fn cmd_status(
         }
 
         let template = workspace_command.commit_summary_template();
-        write!(formatter, "Working copy : ")?;
+        write!(formatter, "Working copy  ")?;
+        formatter.with_label("node", |fmt| write!(fmt.labeled("working_copy"), "@"))?;
+        write!(formatter, "  ")?;
         formatter.with_label("working_copy", |fmt| template.format(wc_commit, fmt))?;
         writeln!(formatter)?;
         for parent in wc_commit.parents() {
             let parent = parent?;
-            write!(formatter, "Parent commit: ")?;
+            write!(formatter, "Parent commit ")?;
+            write!(formatter.labeled("node"), "@-")?;
+            write!(formatter, " ")?;
             template.format(&parent, formatter)?;
             writeln!(formatter)?;
         }


cli/src/cli_util.rs
```diff
@@ -2023,13 +2023,18 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
         if Some(new_commit) != maybe_old_commit {
             if let Some(mut formatter) = ui.status_formatter() {
                 let template = self.commit_summary_template();
-                write!(formatter, "Working copy now at: ")?;
+                write!(formatter, "Working copy now ")?;
+                formatter.with_label("node", |fmt| write!(fmt.labeled("working_copy"), "@"))?;
```
robinst, line 2027: I didn't see a way to apply multiple labels except like this. Not sure if worth extending `labeled` to have that?

yuja: Do we need label for `@`? I assumed it would be part of text `"Working copy (@) now at: "` or something.

BTW, `node` sounds odd because it isn't a graph node symbol, but a revset expression.

robinst: If we want the `@` to be in green color, then we currently need both `node` and `working_copy` labels:

https://github.com/jj-vcs/jj/blob/5eae2d92a08a251efc61549fe6ae2e9b480ca052/cli/src/config/colors.toml#L110

(From this comment I thought we'd want to color it consistently with `jj log`: https://github.com/jj-vcs/jj/issues/4700#issuecomment-2434194381)

> BTW, `node` sounds odd because it isn't a graph node symbol, but a revset expression.

Hmm yeah. I guess we could add a new config to `colors.toml`, e.g. `"working_copy symbol"`.

yuja: > (From this comment I thought we'd want to color it consistently with `jj log`: [#4700 (comment)](https://github.com/jj-vcs/jj/issues/4700#issuecomment-2434194381))
> 
> > BTW, `node` sounds odd because it isn't a graph node symbol, but a revset expression.
> 
> Hmm yeah. I guess we could add a new config to `colors.toml`, e.g. `"working_copy symbol"`.

Perhaps, the label mentioned in the comment is `working_copies`. The plural form is weird here, though.
> * `working_copies() -> String`: For multi-workspace repository, indicate
>   working-copy commit as `<workspace name>@`.

martinvonz: As I [said before](https://github.com/jj-vcs/jj/issues/4700#issuecomment-2434195465), I think it's quite confusing if the `@` is expected to be read as "at" and be interpreted as both the regular English word "at" and a reminder of what the symbol for the working copy is. I think "Working copy (@) now at: " is much clearer.

robinst: @martinvonz Changed now (see updated PR description), but I don't particularly like the format. I still don't understand why we're not using these as prefixes:

```
"Working copy:  @  "
"Parent commit: @- "
```

and:

```
"Working copy now at: @  "
"Parent commit:       @- "
```

IMO that looks the nicest:
- No parentheses around the symbol needed
- Only thing that needs to be aligned is the symbol, no `Working copy  (@) now at:`
- Symbol next to change ID
- It wouldn't require any changes to the `commit_summary` template either

yuja: > ```
> "Working copy now at: @  "
> "Parent commit:       @- "
> ```

I think the weird part of this formatting is that static relations is separated by `:`.

```
Working copy now at: @ <commit summary>
^^^^^^^^^^^^         ^ ^^^^^^^^^^^^^^^^
A                    B C
```

`A == B` always holds, but `A == C` dynamically changes.

martinvonz: I agree with Yuya. "Working copy now at: @ " sounds redundant because the working copy is always at "@".

martinvonz: I think I would also prefer to not color the "@". If we do, we should also color the "@" in "@-". Maybe what that implies is that we start defining a syntax highlighting for revset expressions. Maybe we want some color for operators like "-" too in that case, or we can just leave the uncolored. For now, I would not color either of them.

martinvonz (commented):
I think you will need to sync this past PR #6068. To update the tests, you can run `cargo insta test --accept --test-runner nextest`.

cli/src/cli_util.rs
robinst: (I was looking at it from an UX perspective: The `@` or the change ID is what the user will type/copy paste when they want to refer to the commit. So to me it makes a lot of sense to put them next to each other.)

But ok, will change and remove color too.

robinst:
> To update the tests, you can run `cargo insta test --accept --test-runner nextest`.

Getting a lot of tests failing with `Error: Could not find repository at 'origin'` (and updating the snapshots to that). Haven't investigated what the problem could be, and a search in issues/discussion didn't turn up anything.

CHANGELOG.md
```diff
@@ -89,6 +89,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   reversing colors rather than underlining, you can set
   `colors."diff token"={ underline = false, reverse = true }` in your config.
 
+* `jj status` and output showing the new working copy commit now includes `@`
+  and `@-` to help newcomers learn the symbols.
```
martinvonz, line 93: nit: I don't think we need to mention it in the release notes. Users don't need to act on it, and it probably doesn't reduce surprises much either (it should be pretty obvious what the symbols mean to any existing users).

martinvonz (approved):

PR #5900, opened 2025-03-06
docs: Update documentation wrt JJ_CONFIG and --when/--scope

algmyr:
This is hidden gem of a feature, especially in combination.

`JJ_CONFIG` being a directory allows loading multiple TOML config files. `--when` can be used on the top level, which can lend itself to a much cleaner config than `[[--scope]]` tables.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1455,13 +1455,18 @@ services. It is an error for both of these files to exist.
 | Windows  | `{FOLDERID_RoamingAppData}\jj\config.toml`         | `C:\Users\Alice\AppData\Roaming\jj\config.toml`           |
 
 The location of the `jj` config file can also be overridden with the
-`JJ_CONFIG` environment variable. If it is not empty, it should contain the path
-to a TOML file that will be used instead of any configuration file in the
-default locations. For example,
+`JJ_CONFIG` environment variable. If it is not empty, it will be used instead
+of any configuration file in the default locations. If it is a path to a TOML
+file, then that file will be loaded instead. If it is a path to a directory,
+then all the TOML files in that directory will be loaded in lexicographic order
+and merged. This allows you to split your configuration across multiple files
+and can combine well with [Conditional Variables](#conditional-variables).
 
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -1456,13 +1456,17 @@ services. It is an error for both of these files to exist.
 
 The location of the `jj` config file can also be overridden with the
 `JJ_CONFIG` environment variable. If it is not empty, it should contain the path
-to a TOML file that will be used instead of any configuration file in the
-default locations. For example,
+to a TOML file or a directory containing TOML files that will be used instead
+of any configuration file in the default locations. For example,
```
fowles, line 1460: How about:

If it is not empty, it will be used instead of any configuration file in the default locations.  If it is a path to a TOML file, then that file will be loaded instead.  If it is a path to a directory, then all the TOML files in that directory will be loaded in lexicographic order and merged.  This allows you to split your configuration across multiple files and can combine well with [Conditional Variables](#Conditional Variables).

docs/config.md
```diff
@@ -1554,7 +1560,33 @@ paginate = "never"
 pager = "delta"
 ```
 
-Condition keys:
+#### Using multiple files
+`--when` can also be used on the top level of a TOML file, which is convenient
+when splitting your config across multiple files using `JJ_CONFIG`.
+The behavior of conditions are the same as when using `[[--scope]]` tables.
+
```
yuja, line 1567: nit: maybe better to mention the load order again? If `work.toml` were `a.toml`, it would be overridden by `config.toml`.

docs/config.md
algmyr: Thanks, I also felt that the wording became a bit awkward. I still kept the existing example, let me know if it would be useful to extend or remove that.

docs/config.md
algmyr: Added a sentence to point this out.

yuja (approved):
Thanks.

PR #5901, opened 2025-03-06
cli `op diff`: show changes to working copy positions

ilyagr:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/operation/diff.rs
+++ b/cli/src/commands/operation/diff.rs
@@ -29,6 +29,7 @@ use jj_lib::matchers::EverythingMatcher;
 use jj_lib::op_store::RefTarget;
 use jj_lib::op_store::RemoteRef;
 use jj_lib::op_store::RemoteRefState;
+use jj_lib::refs::diff_named_commit_ids;
 use jj_lib::refs::diff_named_ref_targets;
 use jj_lib::refs::diff_named_remote_refs;
 use jj_lib::repo::ReadonlyRepo;
@@ -274,6 +275,40 @@ pub fn show_op_diff(
                     show_change_diff(ui, formatter, diff_renderer, modified_change, width)?;
                 }
             }
+        }
+    }
+
+    let changed_working_copies = diff_named_commit_ids(
+        from_repo.view().wc_commit_ids(),
+        to_repo.view().wc_commit_ids(),
+    )
+    .collect_vec();
+    if !changed_working_copies.is_empty() {
+        writeln!(formatter)?;
+        for (name, (from_commit, to_commit)) in changed_working_copies {
+            with_content_format.write(formatter, |formatter| {
+                // Usually, there is at most one working copy changed per operation, so we put
+                // the working copy name in the heading.
+                write!(formatter, "Changed working copy ")?;
+                write!(formatter.labeled("working_copies"), "{}@", name.as_str())?;
+                writeln!(formatter, ":")?;
+                write_ref_target_summary(
+                    formatter,
+                    current_repo,
+                    commit_summary_template,
+                    &RefTarget::resolved(to_commit.cloned()),
+                    true,
+                    None,
+                )?;
+                write_ref_target_summary(
+                    formatter,
+                    current_repo,
+                    commit_summary_template,
+                    &RefTarget::resolved(from_commit.cloned()),
+                    false,
+                    None,
+                )
+            })?;
         }
     }
 
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1664,6 +1664,7 @@ impl MutableRepo {
     }
 
     fn merge_view(&mut self, base: &View, other: &View) -> BackendResult<()> {
+        // TODO: Use `diff_named_commit_ids` to simplify this.
         // Merge working-copy commits. If there's a conflict, we keep the self side.
         for (workspace_id, base_wc_commit) in base.wc_commit_ids() {
             let self_wc_commit = self.view().get_wc_commit_id(workspace_id);
--- a/lib/src/view.rs
+++ b/lib/src/view.rs
@@ -15,7 +15,6 @@
 #![allow(missing_docs)]
 
 use std::collections::BTreeMap;
-use std::collections::HashMap;
 use std::collections::HashSet;
 
 use itertools::Itertools;
@@ -46,7 +45,7 @@ impl View {
         }
     }
 
-    pub fn wc_commit_ids(&self) -> &HashMap<WorkspaceId, CommitId> {
+    pub fn wc_commit_ids(&self) -> &BTreeMap<WorkspaceId, CommitId> {
         &self.data.wc_commit_ids
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,9 +41,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * The 'how to resolve conflicts' hint that is shown when conflicts appear can
   be hidden by setting `hints.resolving-conflicts = false`.
 
-* `jj op log -d` now has an alias for `jj op log --op-diff`.
+* `jj op diff` and `jj op log --op-diff` now show changes to which commits
+  correspond to working copies.
+
+* `jj op log -d` is now an alias for `jj op log --op-diff`.
 
 ### Fixed bugs
+
 
 ## [0.27.0] - 2025-03-05
 
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2076,9 +2076,7 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
             writeln!(ui.status(), "Rebased {num_rebased} descendant commits")?;
         }
 
-        for (workspace_id, wc_commit_id) in tx.repo().view().wc_commit_ids().clone().iter().sorted()
-        //sorting otherwise non deterministic order (bad for tests)
-        {
+        for (workspace_id, wc_commit_id) in &tx.repo().view().wc_commit_ids().clone() {
             if self
                 .env
                 .find_immutable_commit(tx.repo(), [wc_commit_id])?
--- a/cli/src/commands/workspace/list.rs
+++ b/cli/src/commands/workspace/list.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use itertools::Itertools;
 use jj_lib::repo::Repo;
 use tracing::instrument;
 
@@ -34,7 +33,7 @@ pub fn cmd_workspace_list(
     let repo = workspace_command.repo();
     let mut formatter = ui.stdout_formatter();
     let template = workspace_command.commit_summary_template();
-    for (workspace_id, wc_commit_id) in repo.view().wc_commit_ids().iter().sorted() {
+    for (workspace_id, wc_commit_id) in repo.view().wc_commit_ids() {
         write!(formatter, "{}: ", workspace_id.as_str())?;
         let commit = repo.store().get_commit(wc_commit_id)?;
         template.format(&commit, formatter.as_mut())?;
--- a/cli/src/commit_templater.rs
+++ b/cli/src/commit_templater.rs
@@ -1067,7 +1067,7 @@ fn extract_working_copies(repo: &dyn Repo, commit: &Commit) -> String {
         return "".to_string();
     }
     let mut names = vec![];
-    for (workspace_id, wc_commit_id) in wc_commit_ids.iter().sorted() {
+    for (workspace_id, wc_commit_id) in wc_commit_ids {
         if wc_commit_id == commit.id() {
             names.push(format!("{}@", workspace_id.as_str()));
         }
--- a/cli/tests/test_operations.rs
+++ b/cli/tests/test_operations.rs
@@ -88,11 +88,19 @@ fn test_op_log() {
     │  Changed commits:
     │  ○  + qpvuntsm 19611c99 (empty) description 0
     │     - qpvuntsm hidden 230dd059 (empty) (no description set)
+    │
+    │  Changed working copy default@:
+    │  + qpvuntsm 19611c99 (empty) description 0
+    │  - qpvuntsm hidden 230dd059 (empty) (no description set)
     ○  eac759b9ab75 test-username@host.example.com 2001-02-03 04:05:07.000 +07:00 - 2001-02-03 04:05:07.000 +07:00
     │  add workspace 'default'
     │
     │  Changed commits:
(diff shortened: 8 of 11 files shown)


cli/src/commands/operation/diff.rs
```diff
@@ -486,6 +520,40 @@ fn get_parent_changes(
...
+    let new_working_copies = to_repo.view().wc_commit_ids();
+    itertools::merge_join_by(
+        old_working_copies
+            .iter()
+            .map(|(wid, val)| (wid.as_str(), val)),
+        new_working_copies
+            .iter()
+            .map(|(wid, val)| (wid.as_str(), val)),
+        |(name1, _), (name2, _)| name1.cmp(name2),
+    )
```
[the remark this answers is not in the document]
ilyagr: Good catch, thanks!

cli/src/commands/operation/diff.rs
```diff
@@ -277,6 +278,39 @@ pub fn show_op_diff(
         }
     }
 
+    let changed_working_copies = get_changed_working_copies(from_repo, to_repo);
+    if !changed_working_copies.is_empty() {
+        let as_ref_target = |maybe_commit: Option<CommitId>| {
+            maybe_commit.map_or(RefTarget::absent(), RefTarget::normal)
+        };
```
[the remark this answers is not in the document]
ilyagr: Nice!

It's a bit uglier than I'd like because I need to store `RefTarget`s somewhere to pass `&RefTarget`s to `diff_named_ref_targets`, but I think it's a lot better than what I had.

cli/src/commands/operation/diff.rs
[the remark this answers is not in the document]
ilyagr: Makes sense, done, thanks!

cli/src/commands/operation/diff.rs
```diff
@@ -277,6 +277,42 @@ pub fn show_op_diff(
         }
     }
 
+    let from_wc_ref_targets = get_ordered_wc_ref_targets(from_repo);
+    let to_wc_ref_targets = get_ordered_wc_ref_targets(to_repo);
+    let changed_working_copies: Vec<(&str, (&RefTarget, &RefTarget))> = diff_named_ref_targets(
```
ilyagr, line 282: Can't decide if this is worthwhile, leaning towards no:

```diff
diff --git a/cli/src/commands/operation/diff.rs b/cli/src/commands/operation/diff.rs
index 1958575a1d..044edf00a1 100644
--- a/cli/src/commands/operation/diff.rs
+++ b/cli/src/commands/operation/diff.rs
@@ -279,12 +279,12 @@
 
     let from_wc_ref_targets = get_ordered_wc_ref_targets(from_repo);
     let to_wc_ref_targets = get_ordered_wc_ref_targets(to_repo);
-    let changed_working_copies: Vec<(&str, (&RefTarget, &RefTarget))> = diff_named_ref_targets(
+    let mut changed_working_copies = diff_named_ref_targets(
         from_wc_ref_targets.iter().map(|(k, v)| (k.as_str(), v)),
         to_wc_ref_targets.iter().map(|(k, v)| (k.as_str(), v)),
     )
-    .collect_vec();
-    if !changed_working_copies.is_empty() {
+    .peekable();
+    if changed_working_copies.peek().is_some() {
         writeln!(formatter)?;
         for (name, (from_commit, to_commit)) in changed_working_copies {
             with_content_format.write(formatter, |formatter| {

```

lib/src/view.rs
```diff
@@ -46,7 +45,7 @@ impl View {
         }
     }
 
-    pub fn wc_commit_ids(&self) -> &HashMap<WorkspaceId, CommitId> {
+    pub fn wc_commit_ids(&self) -> &BTreeMap<WorkspaceId, CommitId> {
```
yuja, line 48: nit: we can remove `.sorted()` from callers.

cli/src/commands/operation/diff.rs
```diff
@@ -277,6 +277,42 @@ pub fn show_op_diff(
...
+        from_wc_ref_targets.iter().map(|(k, v)| (k.as_str(), v)),
+        to_wc_ref_targets.iter().map(|(k, v)| (k.as_str(), v)),
```
yuja, line 284: nit: just `.map(|(k, v)| (k, v))`?

cli/src/commands/operation/diff.rs
```diff
@@ -486,6 +522,14 @@ fn get_parent_changes(
     }
 }
 
+fn get_ordered_wc_ref_targets(repo: &ReadonlyRepo) -> Vec<(String, RefTarget)> {
+    repo.view()
+        .wc_commit_ids() // Returns an ordered map
+        .iter()
+        .map(|(k, v)| (k.as_str().to_string(), RefTarget::normal(v.clone())))
```
yuja, line 529: nit: it's probably better to not stringify the `WorkspaceId` here. Typed value is nicer.

cli/src/commands/operation/diff.rs
ilyagr: SG. I wish there was a prettier way to do this.

cli/src/commands/operation/diff.rs
ilyagr: Sure, NP.

lib/src/view.rs
ilyagr: Yes, good idea.

There's more complicated code in `repo.rs` around line 1666 that might be possible to simplify, I can add a TODO there as well.

lib/src/repo.rs
```diff
@@ -1664,6 +1664,7 @@ impl MutableRepo {
     }
 
     fn merge_view(&mut self, base: &View, other: &View) -> BackendResult<()> {
+        // TODO(ilyagr): It might be possible to use a merge join here
```
ilyagr, line 1667: I can remove this TODO. Perhaps this code will be clearer to me tomorrow.

yuja: Yes, I think this can be merge join, and maybe we can use trivial merge function.

Then, it might make sense to add `refs::diff_named_commit_ids(..) -> impl Iterator<Item = (K, (Option<&CommitId>, Option<&CommitId>))>`. There will be two callers.

ilyagr: I think I'll just reword the TODO for now, this cleanup can go into a separate PR (unless I suddenly feel like doing it immediately).

I do like it, if the working copies are returned from views a `<K, &CommitId>`, avoiding the conversion to RefTarget until the very end will simplify things.

ilyagr: OK, I did actually do it, it does turn out quite nicely.

I'm only 90% sure the simplification in `repo.rs` is correct, PTAL. If it needs work, I'll split it out into a separate PR.

lib/src/repo.rs
```diff
@@ -1664,33 +1665,38 @@ impl MutableRepo {
...
-                    self.view_mut()
-                        .set_wc_commit(workspace_id.clone(), other_wc_commit.clone());
-                }
+        let mut workspace_updates = vec![];
+        for (workspace_id, (my_commit, other_commit)) in
+            diff_named_commit_ids(self.view().wc_commit_ids(), other.wc_commit_ids())
+        {
+            // We never have `my_commit == other_commit` as
+            // `diff_named_commit_ids` omits such pairs. This is fine; we want
+            // to keep the `self` side unchanged in such cases.
```
ilyagr, line 1674: I'm thinking that in this code, excluding the different ones makes little sense logically (though it doesn't affect the outcome). I'm thinking of absorbing the following:

```diff
diff --git a/cli/src/commands/operation/diff.rs b/cli/src/commands/operation/diff.rs
index 60b1eb663e..b02f8cb271 100644
--- a/cli/src/commands/operation/diff.rs
+++ b/cli/src/commands/operation/diff.rs
@@ -29,9 +29,9 @@
 use jj_lib::op_store::RefTarget;
 use jj_lib::op_store::RemoteRef;
 use jj_lib::op_store::RemoteRefState;
-use jj_lib::refs::diff_named_commit_ids;
 use jj_lib::refs::diff_named_ref_targets;
 use jj_lib::refs::diff_named_remote_refs;
+use jj_lib::refs::pair_named_commit_ids;
 use jj_lib::repo::ReadonlyRepo;
 use jj_lib::repo::Repo;
 use jj_lib::revset;
@@ -278,10 +278,11 @@
         }
     }
 
-    let changed_working_copies = diff_named_commit_ids(
+    let changed_working_copies = pair_named_commit_ids(
         from_repo.view().wc_commit_ids(),
         to_repo.view().wc_commit_ids(),
     )
+    .filter(|(_, (target1, target2))| target1 != target2)
     .collect_vec();
     if !changed_working_copies.is_empty() {
         writeln!(formatter)?;
diff --git a/lib/src/refs.rs b/lib/src/refs.rs
index 0b5ed3ae35..d6ed25c689 100644
--- a/lib/src/refs.rs
+++ b/lib/src/refs.rs
@@ -74,10 +74,11 @@
     )
 }
 
-/// Compares `commit1` and `commit2` targets, yields entry if they differ.
+/// Matches `commit1` and `commit2` targets with the same names. Includes equal
+/// pairs.
 ///
 /// `commit1` and `commit2` must be sorted by `K`.
-pub fn diff_named_commit_ids<'a, 'b, K: Ord>(
+pub fn pair_named_commit_ids<'a, 'b, K: Ord>(
     ids1: impl IntoIterator<Item = (K, &'a CommitId)>,
     ids2: impl IntoIterator<Item = (K, &'b CommitId)>,
 ) -> impl Iterator<Item = (K, (Option<&'a CommitId>, Option<&'b CommitId>))> {
@@ -87,7 +88,6 @@
         || None,
         || None,
     )
-    .filter(|(_, (target1, target2))| target1 != target2)
 }
 
 fn iter_named_pairs<K: Ord, V1, V2>(
diff --git a/lib/src/repo.rs b/lib/src/repo.rs
index 8c93642ec3..6847d478b3 100644
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -74,11 +74,11 @@
 use crate::op_store::RootOperationData;
 use crate::op_store::WorkspaceId;
 use crate::operation::Operation;
-use crate::refs::diff_named_commit_ids;
 use crate::refs::diff_named_ref_targets;
 use crate::refs::diff_named_remote_refs;
 use crate::refs::merge_ref_targets;
 use crate::refs::merge_remote_refs;
+use crate::refs::pair_named_commit_ids;
 use crate::refs::RemoteRefSymbol;
 use crate::revset;
 use crate::revset::RevsetExpression;
@@ -1667,11 +1667,8 @@
     fn merge_view(&mut self, base: &View, other: &View) -> BackendResult<()> {
         let mut workspace_updates = vec![];
         for (workspace_id, (my_commit, other_commit)) in
-            diff_named_commit_ids(self.view().wc_commit_ids(), other.wc_commit_ids())
+            pair_named_commit_ids(self.view().wc_commit_ids(), other.wc_commit_ids())
         {
-            // We never have `my_commit == other_commit` as
-            // `diff_named_commit_ids` omits such pairs. This is fine; we want
-            // to keep the `self` side unchanged in such cases.
             let base_wc_commit = base.get_wc_commit_id(workspace_id);
             let changed_commit = if my_commit == base_wc_commit {
                 other_commit
@@ -1684,7 +1681,8 @@
                 // side. We want to remove it. TODO(ilyagr): I'm not sure why.
                 None
             } else {
-                // There is a conflict. We keep the `self` side unchanged in this case.
+                // There is a conflict, or both sides changed in the same way. We keep the
+                // `self` side unchanged in this case.
                 continue;
             };
             workspace_updates.push((workspace_id.clone(), changed_commit.cloned()));
```

ilyagr: Another possibility would be to create a version of `iter_named_pairs` that takes two arguments and returns `(K, (Option<V>, Option<V>)`. It could be called `iter_named_pairs_option`, at least nothing better came to my mind yet.

lib/src/repo.rs
```diff
@@ -1664,33 +1665,38 @@ impl MutableRepo {
...
-                // The other side didn't change or both sides changed in the
-                // same way.
-            } else if let Some(other_wc_commit) = other_wc_commit {
-                if self_wc_commit == Some(base_wc_commit) {
-                    self.view_mut()
-                        .set_wc_commit(workspace_id.clone(), other_wc_commit.clone());
-                }
+        let mut workspace_updates = vec![];
+        for (workspace_id, (my_commit, other_commit)) in
+            diff_named_commit_ids(self.view().wc_commit_ids(), other.wc_commit_ids())
```
yuja, line 1670: We compare base and other to get around borrowing issue. I think we can do the same:
```rust
        let changed_wc_commits = diff_named_commit_ids(base.wc_commit_ids(), other.wc_commit_ids());
        for (workspace_id, (base_target, other_target)) in changed_wc_commits {
            let self_target = self.view().get_wc_commit_id(workspace_id);
            if let Some(&new_target) = trivial_merge(&[self_target, base_target, other_target]) {
                if let Some(target) = new_target.cloned() {
                    self.view_mut().set_wc_commit(workspace_id.clone(), target);
                } else {
                    self.view_mut().remove_wc_commit(workspace_id);
                }
            } else // special case for merge conflict
```

ilyagr: I think I'll extract this to a separate PR.

It seems to me this might not have the same (strange IMO; I'm guessing it's a workaround around... something) behavior if `other_target` is `None`.

Even if it does, I want to try writing it out in `if` statements, I think that might look clearer.

lib/src/repo.rs
ilyagr: Moved to https://github.com/jj-vcs/jj/pull/5993.

lib/src/repo.rs
ilyagr: Moved to https://github.com/jj-vcs/jj/pull/5993.

yuja (approved):

lib/src/repo.rs
yuja: I think the intent here is that deletion is always prioritized, no matter if the workspace was deleted at the `self` side or the `other` side.

ilyagr: Yes, but why does this make sense?

ilyagr:
Thank you, Yuya!

lib/src/repo.rs
yuja: I assume the expectation is that `jj workspace forget` is stronger than e.g. `jj new` because the workspace is forgotten anyway. I don't know if that's always true, but it would be more surprising if `jj workspace forget` was cancelled?

ilyagr: Yeah, I'm not sure. Let's continue this in <https://github.com/jj-vcs/jj/pull/5993#discussion_r1990511541>, though my reply there doesn't say much.

PR #5903, opened 2025-03-06
cli `op log`: short `-d` alias for `--op-diff`

ilyagr:
`--op-diff` is often more useful than `-p`/`--patch`.

# Checklist

If applicable:
- ~~[ ] I have added tests to cover my changes~~


yuja (approved):
It's unlikely this would conflict with the other `-d` (such as `--destination`.) So I think this is good.

ilyagr:
Thank you!

PR #5905, opened 2025-03-07
cli: remove `untrack` subcommand

gechr:
`jj untrack` was scheduled to be removed in v0.27 but slipped through the net, so remove support in v0.28 instead.


martinvonz:
> `jj untrack` was scheduled to be removed in v0.27 but slipped through the net

That's perfectly fine IMO. I see the scheduled removal date as just a lower bound.

cli/src/commands/mod.rs
yuja, line 1: nit: `pub mod` can be removed from cli/src/commands/file/mod.rs

yuja (approved):

PR #5907, opened 2025-03-07
cleanup: switch to use<'_> capturing syntax

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/commit.rs
+++ b/lib/src/commit.rs
@@ -93,7 +93,7 @@ impl Commit {
         &self.data.parents
     }
 
-    pub fn parents(&self) -> impl Iterator<Item = BackendResult<Commit>> + '_ {
+    pub fn parents(&self) -> impl Iterator<Item = BackendResult<Commit>> + use<'_> {
         self.data.parents.iter().map(|id| self.store.get_commit(id))
     }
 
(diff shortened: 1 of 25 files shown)


PhilipMetzger (approved):

lib/src/commit.rs
```diff
@@ -93,15 +93,15 @@ impl Commit {
         &self.data.parents
     }
 
-    pub fn parents(&self) -> impl Iterator<Item = BackendResult<Commit>> + '_ {
+    pub fn parents(&self) -> impl Iterator<Item = BackendResult<Commit>> + use<'_> {
```
martinvonz, line 96: IIUC, this will be the default with the 2024 edition, so I suppose we'll drop the `+ use<'_>` then. Is there still a reason to make the change now?

yuja: I thought it would be inconsistent to leave `+ '_` here. My main motivation of this patch is to remove weird `'a: 'b, 'b` constraints.

martinvonz: Sounds good. There's probably some Clippy lint that figures out that we can remove it whether it's `+ '_` or `+ use<'_>`.

martinvonz (approved):

Issue #3694, opened 2024-05-16
`jj move` from parent to child creates conflict


matts1:
## Steps to Reproduce the Problem

Create an empty `jj` repo, then:
```
echo "foo
bar
baz" > file
jj new -m "child"
jj move --from @- --to @ --interactive
```

Then select just the line "bar"

I came across this use case by running `jj split`, then realizing I missed putting something in the child, so I expect this to be a somewhat common use case.

## Expected Behavior
Neither commit should have a conflict in it

## Actual Behavior
The parent commit has a conflict which is then resolved in the child

## Specifications

- Platform: Linux
- Version: `jj google-0.17.1-jj-client_20240515_00_RC00-633966000`

matts1:
Also, just double checked. The same problem also applies to `squash`, which is supposed to replace move.

martinvonz:
> jj google-0.17.1-jj-client_20240515_00_RC00-633966000

That's an internal version. Does the problem happen with a regular upstream build too?

ilyagr:
I think I can reproduce this. The conflict I get in the parent is:

```
<<<<<<< Conflict 1 of 1
+++++++ Contents of side #1
foo
bar
baz
%%%%%%% Changes from base to side #2
-bar
>>>>>>>
```

ilyagr:
If you start with

```
foo
bar
baz
oof
zab
```

and leave only

```
bar
baz
```

during the split, there is no conflict by the way. It feels like this might be about reaching some threshold in the diff algorithm.

bnjmnt4n:
Correct me if I'm wrong, but it seems like there will always be a conflict when removing changes from any commit when doing a squash operation if there are hunks with more than 1 line, with at least 1 and less than all the lines in the hunk being selected. This means that there will always be a conflict when "removing" the selected lines for squashing, since the 3-way merge will result in a conflicting hunk (added hunk, a few lines selected from added hunk, empty hunk) which cannot be resolved trivially.

It seems like this isn't an issue most of the time because squashing is typically done from descendant to ancestor, and the conflict will be resolved after rebasing onto the modified descendant.

matts1:
Suppose you have a commit `A` with the file `f` containing:
```
foo
bar
baz
```

If we try to move the line `bar` to a *parent* commit `P` (the usual case), I believe what happens internally roughly corresponds to (algorithmically speaking - obviously it doesn't actually run these commands):
[...]

PR #5916, opened 2025-03-07
lib rewrite: Add CommitWithSelection::invert() to invert the selection

emesterhazy:
This will be used in `jj split` in a follow up commit and when the new -A, -B, and -d options are added.

# Checklist

If applicable:

### DIFF
--- a/lib/src/rewrite.rs
+++ b/lib/src/rewrite.rs
@@ -1029,6 +1029,7 @@ fn compute_commits_heads(
         .collect_vec()
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct CommitWithSelection {
     pub commit: Commit,
     pub selected_tree: MergedTree,
@@ -1036,6 +1037,24 @@ pub struct CommitWithSelection {
 }
 
 impl CommitWithSelection {
+    /// Returns a new CommitWithSelection where the selection is inverted (i.e.
+    /// it includes the changes that weren't selected previously, and excludes
+    /// those that were).
+    pub fn invert(self) -> BackendResult<CommitWithSelection> {
+        // Merge the original commit tree with its parent using the tree
+        // containing the selected changes as the base for the merge. This
+        // results in a tree with the changes the user didn't select.
+        let inverted_tree = self
+            .commit
+            .tree()?
+            .merge(&self.selected_tree, &self.parent_tree)?;
+        Ok(CommitWithSelection {
+            commit: self.commit,
+            selected_tree: inverted_tree,
+            parent_tree: self.parent_tree,
+        })
+    }
+
     /// Returns true if the selection contains all changes in the commit.
     pub fn is_full_selection(&self) -> bool {
         &self.selected_tree.id() == self.commit.tree_id()
--- a/lib/tests/test_rewrite.rs
+++ b/lib/tests/test_rewrite.rs
@@ -1878,4 +1878,10 @@ fn test_commit_with_selection() {
     };
     assert!(!full_selection.is_empty_selection());
     assert!(full_selection.is_full_selection());
+
+    assert_eq!(empty_selection.clone().invert().unwrap(), full_selection);
+    assert_eq!(
+        empty_selection.clone().invert().unwrap().invert().unwrap(),
+        empty_selection
+    );
 }


bnjmnt4n:
It isn't possible to invert the selection by doing a 3-way merge of `(commit tree, tree with selected changes, parent commit tree)` for all cases without creating conflicts. This is a similar issue faced by https://github.com/jj-vcs/jj/issues/3694.

See https://github.com/jj-vcs/jj/issues/3694#issuecomment-2581806107.

I think we would be able to perform this inversion without creating conflicts when using the native `scm-record` diff editor, but we could run into conflicts when using any external diff editors.

emesterhazy:
Is there an alternative? I think when we discussed adding the -B,-A,-d flags we acknowledged that split may cause conflicts.

The split command has been using this technique for `--parallel` for a while, this PR is just moving the code into a library.

- https://github.com/jj-vcs/jj/blob/29e2ccf4aeae24fa031e561211750d7e6e93f0b0/cli/src/commands/split.rs#L174

emilazy:
> It isn't possible to invert the selection by doing a 3-way merge of `(commit tree, tree with selected changes, parent commit tree)` for all cases without creating conflicts. This is a similar issue faced by #3694.

I think the idea is that you would choose to do it when you know doing it non-inverted would cause a conflict (e.g. squashing from a parent into a child, or `jj split -A`). It was @scott2000's proposal to help smooth the UX there for both commands.

bnjmnt4n:
To clarify, where will this function be used? I didn't see any mentions in https://github.com/jj-vcs/jj/pull/5828 about inverting a selection. Would there be a new `--invert` option for this?

My understanding is that using the above technique creates a conflict whenever a partial selection of any hunk is made. So I'm not sure if it's practical to attempt to perform inversion if any non-trivial selection would immediately lead to a conflict.

> The split command has been using this technique for --parallel for a while, this PR is just moving the code into a library.

I didn't realize this. In that case, using the command would probably not have been a good experience for users who are trying to split a hunk.

> Is there an alternative?

I'm not too sure myself. Perhaps someone else with more experience with merge algorithms might have a better idea.

emesterhazy:
> To clarify, where will this function be used?

The immediate plan is to use it for `jj split --parallel`, where it's identical to what we're currently doing. Here's the draft change:

- https://github.com/jj-vcs/jj/pull/5925

> > It isn't possible to invert the selection by doing a 3-way merge of `(commit tree, tree with selected changes, parent commit tree)` for all cases without creating conflicts. This is a similar issue faced by #3694.
> 
> I think the idea is that you would choose to do it when you know doing it non-inverted would cause a conflict (e.g. squashing from a parent into a child, or `jj split -A`). It was @scott2000's proposal to help smooth the UX there for both commands.
> 
> FWIW I do feel like we should add the destination flags to `jj split` first and see if the issue is common enough that we need to address it ASAP.

I thought when we discussed adding the flags, we decided that the user would always be prompted to select the changes that are being moved out of the target commit and into the new commit. So for `-A @` we need to compute the changes to leave in the target commit by inverting the selection. Is there some other way to do it?

I thought Scott's suggestion was to add a flag like `--keep-selected` that would prompt the user to select the changes to keep in the target instead of prompting them for the changes to move to the child. But if that flag isn't used, I think we need to invert the selection to compute the target commit's new tree.

I agree that the ux isn't ideal... but I'm not sure what we can do if we want to add these new -A, -B, and -d flags.

If we drop the arbitrary destinations we could have --before and --after flags to put the new commit before or after the target, and always request the user select changes for the parent commit, which will either be the new commit or target depending on which of --before or --after is used.

emilazy:
Okay, I misunderstood what you wanted this for. I assumed this was for `--keep-selected`.

> I thought when we discussed adding the flags, we decided that the user would always be prompted to select the changes that are being moved out of the target commit and into the new commit. So for `-A @` we need to compute the changes to leave in the target commit by inverting the selection. Is there some other way to do it?

I believe the simplest algorithm would be to split the selected changes into a parent (`jj split -r X -B X`, similar to the current behaviour) and then to rebase it to the desired destination. There may be more elegant algorithms, but I believe this would always produce correct results and be good enough for an MVP.

scott2000:
Another possible implementation might be something like this:

1. Rewrite target commit tree to be `(target - selected + parent)`
2. Create new commit with tree `selected` on top of `parent`
3. Move the new commit to the appropriate location

In which case it might be useful to have this function. I'm not sure which approach is better though.

I had thought this might run into issues due to the `A - B + A = A` simplification rule, but probably that shouldn't happen in practice (since that would mean that the target commit doesn't have any changes compared to the parent commit).

bnjmnt4n:
> I thought when we discussed adding the flags, we decided that the user would always be prompted to select the changes that are being moved out of the target commit and into the new commit. So for `-A @` we need to compute the changes to leave in the target commit by inverting the selection. Is there some other way to do it?

I think you're right, and this would be a problem if the user splits and rebases manually anyway. So perhaps all my comments should be ignored, because this is also a known problem as mentioned in #3694. I think I need to re-read the comments in #5828 as well.

Sorry for the extra noise.

emesterhazy:
> Okay, I misunderstood what you wanted this for. I assumed this was for `--keep-selected`.
> 
> > I thought when we discussed adding the flags, we decided that the user would always be prompted to select the changes that are being moved out of the target commit and into the new commit. So for `-A @` we need to compute the changes to leave in the target commit by inverting the selection. Is there some other way to do it?
> 
> I believe the simplest algorithm would be to split the selected changes into a parent (`jj split -r X -B X`, similar to the current behaviour) and then to rebase it to the desired destination. There may be more elegant algorithms, but I believe this would always produce correct results and be good enough for an MVP.

Thanks for calling this out. I think this approach is probably better than what I had in mind. I opened a PR to reimplement --parallel this way:

- https://github.com/jj-vcs/jj/pull/5926

Maybe we can revert this invert() PR then.

PR #5926, opened 2025-03-08
cli split: Implement --parallel by rebasing the second commit

emesterhazy:
This implements --parallel by creating the second commit as usual, and then rebasing it onto the parents of the target commit.

I'm not sure if this is more or less efficient. I suppose it might be less efficient since it writes the second commit twice instead of once, but I'm not sure if that matters in practice.

This approach should be simpler if we end up with arbitrary destinations for the split commit via the proposed -A, -B, -d flags since rebase will compute the new tree for us.

If applicable:

### DIFF
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -15,10 +15,12 @@ use std::io::Write;
 
 use clap_complete::ArgValueCandidates;
 use clap_complete::ArgValueCompleter;
+use itertools::Itertools;
 use jj_lib::commit::Commit;
 use jj_lib::matchers::Matcher;
 use jj_lib::object_id::ObjectId;
 use jj_lib::repo::Repo;
+use jj_lib::rewrite::rebase_commit;
 use jj_lib::rewrite::CommitWithSelection;
 use tracing::instrument;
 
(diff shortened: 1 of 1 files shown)


yuja:
There will be two behavior changes:

1. description template wouldn't show the final diff
2. `jj evolog` would include temporary commit as a predecessor

(In addition to that, `A-B+A=A` rule might cause some differences?)

I think 1 can be addressed by editing description after the rebase. I'm not sure if 2 matters.

@bnjmnt4n any comments?

emesterhazy:
Thanks Yuya, I didn't realize that. Maybe computing the correct tree manually is the better approach in that case.

It sounds like to avoid temporaries in the evolog we'll need to place both the rewritten target and new split commit onto the correct parents when they're created. That means that for `jj split -r x -A x` we'll need to rewrite x first, and then create the new commit. And for `jj split -r x -B x` we'll need to create the new commit first, and then rewrite x.

I guess I'll start with adding tests to catch the behavior change since we don't have any currently.

emilazy:
Can we “rebase” at the tree level to get a simple implementation without evolog worries? I guess we’d still need to take parents into account…

bnjmnt4n:
> There will be two behavior changes:
> 
>     1. description template wouldn't show the final diff
> 
> 
> I think 1 can be addressed by editing description after the rebase. I'm not sure if 2 matters.

I think showing the final diff is definitely important, so we should try to do that. This means we need to rewrite the commit again after the editing, but that's probably fine.

>     2. `jj evolog` would include temporary commit as a predecessor

I think avoiding temporary commits in the evolog is probably a low priority issue, and as mentioned can still be fixed by overriding the predecessors manually.

---

Aside:

Is the point of this change just to prepare for adding `split -A/-B`?

I don't think there's an easy way to implement `split -A/-B` without multiple rounds of rebasing without doing something like `transform_descendants`, but I'm not sure if that would work right now either because `transform_descendants` doesn't handle creating new commits in the graph. I had some work based on https://github.com/jj-vcs/jj/pull/5852, but it's also still quite messy.

> (In addition to that, `A-B+A=A` rule might cause some differences?)

I think I've also seen @yuja mention this for the `squash` implementation, but I haven't looked into that, and am not sure if using `transform_descendants` for `split` would affect this either.

PR #5931, opened 2025-03-08
Don't show console on Windows

sheremetyev:
When used inside GUI application, git.subprocess=true option or enabled signing cause console window to appear temporarily.

std::process::Command needs special flags to prevent that. More details: https://stackoverflow.com/a/60958956

# Checklist

If applicable:

### DIFF
--- a/lib/src/git_subprocess.rs
+++ b/lib/src/git_subprocess.rs
@@ -105,6 +105,14 @@ impl<'a> GitSubprocessContext<'a> {
     /// Create the Git command
     fn create_command(&self) -> Command {
         let mut git_cmd = Command::new(self.git_executable_path);
+        // Hide console window on Windows (https://stackoverflow.com/a/60958956)
+        #[cfg(windows)]
+        {
+            use std::os::windows::process::CommandExt;
+            const CREATE_NO_WINDOW: u32 = 0x08000000;
+            git_cmd.creation_flags(CREATE_NO_WINDOW);
+        }
+
         // TODO: here we are passing the full path to the git_dir, which can lead to UNC
         // bugs in Windows. The ideal way to do this is to pass the workspace
         // root to Command::current_dir and then pass a relative path to the git
--- a/lib/src/gpg_signing.rs
+++ b/lib/src/gpg_signing.rs
@@ -156,6 +156,14 @@ impl GpgBackend {
 
     fn create_command(&self) -> Command {
         let mut command = Command::new(&self.program);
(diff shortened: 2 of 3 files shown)


sheremetyev:
Apologies for the noise - apparently mentioning PR in a commit of another repository can close it...

lib/src/git_subprocess.rs
```diff
@@ -80,6 +80,13 @@ impl<'a> GitSubprocessContext<'a> {
     /// Create the Git command
     fn create_command(&self) -> Command {
         let mut git_cmd = Command::new(self.git_executable_path);
+        // Hide console window on Windows
```
yuja, line 83: nit: for future reference, can you also include the stackoverflow link here?

yuja (approved):
Since stdout/err are piped, setting `CREATE_NO_WINDOW` should be okay.

yuja:
nit: can you also update the commit message to include `git: ` prefix?

lib/src/git_subprocess.rs
sheremetyev: added

sheremetyev:
@yuja I've added another commit with similar fix for signing commands. The code is duplicated for now - probably worth extracting a function for creation of `std::process::Command` - happy to do it here, or in a follow-up PR. (There are a few more places where `Command` is created, but they are unlikely to be used in GUI apps on Windows and don't cause issues for now.)

Also see relevant [discussion on Discord](https://discord.com/channels/968932220549103686/1275974903039922259/1347630628824285227) confirming the fix.

yuja:
> I've added another commit with similar fix for signing commands. The code is duplicated for now - probably worth extracting a function for creation of `std::process::Command`

I was also thinking about that. We'll need utility function to configure `&mut Command` for subprocessing. We might have to add subprocess_util, os_util, or io_util?

emilazy:
I imagine this will also affect e.g. the `git gc` calls `GitBackend` can do.

yuja:
> I imagine this will also affect e.g. the `git gc` calls `GitBackend` can do.

Correct. Since `GitBackend::gc()` forwards output to the main process, we'll need to fix that first. iirc, `NO_CREATE_WINDOW` breaks output inheritance.

emilazy:
It might still be worth merging this as‐is to handle the common case, and figure out tricky cases and abstractions later?

sheremetyev:
> It might still be worth merging this as‐is to handle the common case, and figure out tricky cases and abstractions later?

rebased changes to latest main - will merge when tests pass

PR #5932, opened 2025-03-08
cli: add revert command and deprecate backout command

bnjmnt4n:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -42,6 +42,7 @@ mod prev;
 mod rebase;
 mod resolve;
 mod restore;
+mod revert;
 mod root;
 mod run;
 mod show;
@@ -127,6 +128,7 @@ enum Command {
     Rebase(rebase::RebaseArgs),
     Resolve(resolve::ResolveArgs),
     Restore(restore::RestoreArgs),
+    Revert(revert::RevertArgs),
     Root(root::RootArgs),
     #[command(hide = true)]
     // TODO: Flesh out.
@@ -193,6 +195,7 @@ pub fn run_command(ui: &mut Ui, command_helper: &CommandHelper) -> Result<(), Co
         Command::Rebase(args) => rebase::cmd_rebase(ui, command_helper, args),
         Command::Resolve(args) => resolve::cmd_resolve(ui, command_helper, args),
         Command::Restore(args) => restore::cmd_restore(ui, command_helper, args),
+        Command::Revert(args) => revert::cmd_revert(ui, command_helper, args),
         Command::Root(args) => root::cmd_root(ui, command_helper, args),
         Command::Run(args) => run::cmd_run(ui, command_helper, args),
         Command::SimplifyParents(args) => {
--- a/cli/src/commands/mod.rs
+++ b/cli/src/commands/mod.rs
@@ -88,6 +88,7 @@ const STYLES: Styles = Styles::styled()
 enum Command {
     Abandon(abandon::AbandonArgs),
     Absorb(absorb::AbsorbArgs),
+    // TODO: Remove in jj 0.34+
     Backout(backout::BackoutArgs),
     #[cfg(feature = "bench")]
     #[command(subcommand)]
--- a/cli/src/commands/revert.rs
+++ b/cli/src/commands/revert.rs
@@ -1,0 +1,210 @@
+// Copyright 2025 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashSet;
+
+use bstr::ByteVec as _;
+use clap::ArgGroup;
+use clap_complete::ArgValueCandidates;
+use indexmap::IndexSet;
+use itertools::Itertools as _;
+use jj_lib::backend::CommitId;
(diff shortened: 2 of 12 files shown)


cli/src/commands/mod.rs
```diff
@@ -163,12 +164,15 @@ pub fn run_command(ui: &mut Ui, command_helper: &CommandHelper) -> Result<(), Co
...
-        Command::Backout(args) => backout::cmd_backout(ui, command_helper, args),
+        Command::Backout(args) => {
+            let cmd = renamed_cmd("backout", "revert", false, backout::cmd_backout);
+            cmd(ui, command_helper, args)
+        }
         #[cfg(feature = "bench")]
         Command::Bench(args) => bench::cmd_bench(ui, command_helper, args),
         Command::Bookmark(args) => bookmark::cmd_bookmark(ui, command_helper, args),
         Command::Branch(args) => {
-            let cmd = renamed_cmd("branch", "bookmark", bookmark::cmd_bookmark);
```
ilyagr, line 169: Perhaps this should be a new function, e.g. `deprecated_with_replacement(...)`, instead of a boolean argument? In theory, you could also add a more private helper function they both call, thought in  this case it might be simpler to just duplicate the code.

ilyagr (commented):
The second commit LGTM with one nit. I haven't thought about the first commit and the changes to the `revert` command, hopefully others will also look at it. I'm happy with the overall plan, though.

cli/tests/cli-reference@.md.snap
```diff
@@ -2163,6 +2165,25 @@ See `jj diffedit` if you'd like to restore portions of files rather than entire
...
+
+The reverse of each of the given revisions is applied sequentially in reverse topological order at the given location.
+
+The description of the new revisions can be customized with the `templates.revert_description` config variable.
+
+**Usage:** `jj revert [OPTIONS] <--destination <REVSETS>|--insert-after <REVSETS>|--insert-before <REVSETS>>`
+
+###### **Options:**
+
+* `-r`, `--revisions <REVSETS>` — The revision(s) to apply the reverse of
```
ilyagr, line 2159: (No action needed) Can we revert multiple revisions at once now? Yay! 🚀

michaelpj:
I don't know if this PR already implements it, but it would be nice if this solved https://github.com/jj-vcs/jj/issues/5966

cli/src/commands/revert.rs
```diff
@@ -0,0 +1,180 @@
...
+    let original_parent_commit_ids: HashSet<_> = parents.iter().ids().cloned().collect();
+    let mut new_base_tree = merge_commit_trees(tx.repo(), &parents)?;
+
+    for (commit_to_revert, new_commit_description) in
+        &commits_to_revert_with_new_commit_descriptions
+    {
+        let old_base_tree = commit_to_revert.parent_tree(tx.repo())?;
+        let old_tree = commit_to_revert.tree()?;
+        let new_tree = new_base_tree.merge(&old_tree, &old_base_tree)?;
+        let new_parent_ids = parents.iter().map(|commit| commit.id().clone()).collect();
```
yuja, line 145: nit: `.ids().cloned()`

cli/src/commands/revert.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    let mut num_rebased = 0;
+    for child_commit in new_children {
+        let new_parent_ids = child_commit
+            .parent_ids()
+            .iter()
+            .filter(|id| !original_parent_commit_ids.contains(id))
+            .cloned()
+            .chain(parents.iter().ids().cloned())
```
yuja, line 162: nit: maybe we should replace the first `original_parent_commit_ids` to `parents` to preserve the order? I think `duplicate` works that way.

cli/tests/test_revert_command.rs
```diff
@@ -0,0 +1,392 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::path::Path;
+
+use crate::common::CommandOutput;
+use crate::common::TestEnvironment;
+
+fn create_commit(
```
yuja, line 20: nit: use `common::create_commit_with_files()`

cli/src/commands/mod.rs
yuja: Yeah, we ca just insert the warning to `cmd_backout()`.

yuja (approved):
Looks good to me, thanks.

bnjmnt4n:
I tried to address https://github.com/jj-vcs/jj/issues/5966 by printing the following output:

```
    Reverted 3 commits
    Created the following commits:
      wqnwkozp 05f5fa79 Revert "e"
      mouksmqu f5d9e8b2 Revert "c"
      tqvpomtp fb78f44d Revert "b"
```

cli/src/commands/revert.rs
```diff
@@ -0,0 +1,204 @@
...
+            }
+            num_rebased += 1;
+            rewriter.rebase()?.write()?;
+            Ok(())
+        },
+    )?;
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        writeln!(formatter, "Reverted {} commits", reverted_commits.len())?;
+        writeln!(formatter, "Created the following commits:")?;
```
yuja, line 191: nit: maybe these lines can be combined? "Reverted N commits as follows:" or something like that.

cli/src/commands/revert.rs
```diff
@@ -0,0 +1,204 @@
...
+        for commit in &reverted_commits {
+            write!(formatter, "  ")?;
+            tx.write_commit_summary(formatter.as_mut(), commit)?;
```
yuja, line 194: nit: use `tx.commit_summary_template();` to avoid parsing template in a loop

ilyagr:
Just in case this is simply forgotten about: is anything still blocking this? Benjamin, are there changes you're thinking about making before merging?

bnjmnt4n:
I think I just forgot about it. I'll leave this open for a while more in case there's anyone who wants to add more comments.

emilazy:
My comment is: yay!

PR #5935, opened 2025-03-08
cli: git push: rename git.sign-on-push to `git.sign-on-push.enabled` and add `git.sign-on-push.behavior`

bnjmnt4n:
Previously, when `git.sign-on-push` is set to `true`, commits would be
signed upon pushing with Git with the sign behavior of
`SignBehavior::Own`. However, this could not be customized like
`signing.sign-behavior`.

This commit renames `git.sign-on-push` to `git.sign-on-push.enabled` and
adds `git.sign-on-push.behavior` to allow users to customize the sign
behavior when pushing commits.
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -327,20 +327,20 @@ pub fn cmd_git_push(
         return Ok(());
     }
 
-    let sign_behavior = if tx.settings().get_bool("git.sign-on-push")? {
-        Some(SignBehavior::Own)
+    let sign_behavior = if tx.settings().get_bool("git.sign-on-push.enabled")? {
+        Some(tx.settings().get("git.sign-on-push.behavior")?)
     } else {
         None
     };
-    let commits_to_sign =
+    let commits_to_update_signatures =
         validate_commits_ready_to_push(ui, &bookmark_updates, &remote, &tx, args, sign_behavior)?;
-    if !args.dry_run && !commits_to_sign.is_empty() {
+    if !args.dry_run && !commits_to_update_signatures.is_empty() {
         if let Some(sign_behavior) = sign_behavior {
-            let num_updated_signatures = commits_to_sign.len();
+            let num_updated_signatures = commits_to_update_signatures.len();
             let num_rebased_descendants;
-            (num_rebased_descendants, bookmark_updates) = sign_commits_before_push(
+            (num_rebased_descendants, bookmark_updates) = update_commit_signatures_before_push(
                 &mut tx,
-                commits_to_sign,
+                commits_to_update_signatures,
                 sign_behavior,
                 bookmark_updates,
             )?;
@@ -383,7 +383,7 @@ pub fn cmd_git_push(
 /// Validates that the commits that will be pushed are ready (have authorship
 /// information, are not conflicted, etc.).
 ///
-/// Returns the list of commits which need to be signed.
+/// Returns the list of commits which need their signatures updated.
 fn validate_commits_ready_to_push(
     ui: &Ui,
     bookmark_updates: &[(String, BookmarkPushUpdate)],
@@ -421,7 +421,7 @@ fn validate_commits_ready_to_push(
         sign_settings
     });
 
-    let mut commits_to_sign = vec![];
+    let mut commits_to_update_signatures = vec![];
 
     for commit in workspace_helper
         .attach_revset_evaluator(commits_to_push)
@@ -469,25 +469,26 @@ fn validate_commits_ready_to_push(
             return Err(error);
         }
         if let Some(sign_settings) = &sign_settings {
-            if !commit.is_signed() && sign_settings.should_sign(commit.store_commit()) {
-                commits_to_sign.push(commit);
+            if commit.is_signed() != sign_settings.should_sign(commit.store_commit()) {
+                commits_to_update_signatures.push(commit);
             }
         }
     }
-    Ok(commits_to_sign)
+    Ok(commits_to_update_signatures)
 }
 
-/// Signs commits before pushing.
+/// Update commit signatures before pushing.
 ///
 /// Returns the number of commits with rebased descendants and the updated list
 /// of bookmark names and corresponding [`BookmarkPushUpdate`]s.
-fn sign_commits_before_push(
+fn update_commit_signatures_before_push(
     tx: &mut WorkspaceCommandTransaction,
-    commits_to_sign: Vec<Commit>,
+    commits_to_update_signatures: Vec<Commit>,
     sign_behavior: SignBehavior,
     bookmark_updates: Vec<(String, BookmarkPushUpdate)>,
 ) -> Result<(usize, Vec<(String, BookmarkPushUpdate)>), CommandError> {
-    let commit_ids: IndexSet<CommitId> = commits_to_sign.iter().ids().cloned().collect();
+    let commit_ids: IndexSet<CommitId> =
+        commits_to_update_signatures.iter().ids().cloned().collect();
     let mut old_to_new_commits_map: HashMap<CommitId, CommitId> = HashMap::new();
     let mut num_rebased_descendants = 0;
     tx.repo_mut()
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -406,9 +406,21 @@
                     "default": "origin"
                 },
                 "sign-on-push": {
-                    "type": "boolean",
-                    "description": "Whether jj should sign commits before pushing",
-                    "default": "false"
+                    "type": "object",
+                    "description": "Settings for signing commits before pushing",
+                    "properties": {
+                        "enabled": {
+                            "type": "boolean",
+                            "description": "Whether jj should sign commits before pushing",
+                            "default": "false"
+                        },
+                        "behavior": {
+                            "type": "string",
+                            "enum": ["drop", "keep", "own", "force"],
+                            "description": "Which commits to sign before pushing. Values: drop (never sign), keep (preserve existing signatures), own (sign own commits), force (sign all commits)",
+                            "default": "keep"
+                        }
+                    }
                 },
                 "subprocess": {
                     "type": "boolean",
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -26,6 +26,7 @@ use jj_lib::config::ConfigFile;
 use jj_lib::config::ConfigGetError;
 use jj_lib::config::ConfigLayer;
 use jj_lib::config::ConfigLoadError;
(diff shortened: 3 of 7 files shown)


docs/config.md
calebdw, line 1: Would it be good to update the example to `keep`?

https://github.com/jj-vcs/jj/blob/7e5a3dbef42254c2fd5396aa5caf6715391adcf4/docs/config.md?plain=1#L1220

yuja:
Isn't it better to add separate set of "sign-on-push" behavior such as `never|own|always`?

It seems not all combinations are useful. For example, `enabled = true, behavior = keep` would be subtly different from `enabled = false` in a way that the former will re-sign signed commits (with potentially new keys)? I feel this is rather confusing.

bnjmnt4n:
> Isn't it better to add separate set of "sign-on-push" behavior such as `never|own|always`? (EDIT: or sign-on-push revset?)

If defining `sign-on-push` as `never|own|always` or using a revset, does that mean that existing signed commits will never be dropped? Dropping existing signatures is something that this PR allows, but would be ambiguous when using the above ways to configure. If we agree that there is desired then I can update this PR.

> It seems not all combinations are useful. For example, `enabled = true, behavior = keep` would be subtly different from `enabled = false` in a way that the former will re-sign signed commits (with potentially new keys)? I feel this is rather confusing.

`enabled = true, behavior = keep` shouldn't re-sign any existing signed commits in the current implementation.

yuja:
> > Isn't it better to add separate set of "sign-on-push" behavior such as `never|own|always`? (EDIT: or sign-on-push revset?)
> 
> If defining `sign-on-push` as `never|own|always` or using a revset, does that mean that existing signed commits will never be dropped?

`drop` could be added to `never|drop|own|always` (if needed), but yes, if it's a revset, `drop` cannot be supported. I wouldn't expect `sign-on-push` can be configured to drop signature, though.

> > It seems not all combinations are useful. For example, `enabled = true, behavior = keep` would be subtly different from `enabled = false` in a way that the former will re-sign signed commits (with potentially new keys)? I feel this is rather confusing.
> 
> `enabled = true, behavior = keep` shouldn't re-sign any existing signed commits in the current implementation.

Oh, so `enabled = true, behavior = keep` is basically the same as `enabled = false`? (If I understand it, `keep` doesn't sign unsigned commits.)

My feeling is that the `SignBehavior` only makes sense in rebase context, whereas `sign-on-push` is more similar to `jj sign`.

bnjmnt4n:
I think I agree with your sentiments overall. It probably isn't necessary to drop signatures on push anyway. I will update this PR to change `git.sign-on-push` into a revset instead.

docs/config.md
```diff
@@ -1191,24 +1191,38 @@ as follows:
...
+### Signing behavior
+
+The `signing.behavior` config option can be used to customize whether commits
+are signed when creating new commits or creating new commits from of existing
+commits. The valid options are:
+
+- `drop`: Does not sign new commits and drops all signatures when creating new
+  commits from an existing signed commit.
+- `keep` (default): Does not sign new commits, but preserve existing signatures
+  when creating new commits from a signed commit.
```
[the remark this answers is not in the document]
calebdw: Hmmm, how about:
```md
- `keep` (default): Does not sign new commits unless the commit replaces an existing signed one.
```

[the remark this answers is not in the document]
joyously: how about:
`keep existing signatures, except when signed commits are rewritten`

cli/src/commands/git/push.rs
```diff
@@ -468,10 +453,8 @@ fn validate_commits_ready_to_push(
             }
             return Err(error);
         }
-        if let Some(sign_settings) = &sign_settings {
-            if !commit.is_signed() && sign_settings.should_sign(commit.store_commit()) {
-                commits_to_sign.push(commit);
-            }
+        if !commit.is_signed() && should_sign_on_push(commit.id())? {
```
yuja, line 456: (not a blocker) maybe this isn't a new problem, but do we handle the case where signed commit has unsigned ancestors to be signed?

cli/src/config.rs
```diff
@@ -624,6 +625,34 @@ pub fn default_config_migrations() -> Vec<ConfigMigrationRule> {
             "core.watchman.register_snapshot_trigger",
             "core.watchman.register-snapshot-trigger",
         ),
+        // TODO: Delete in jj 0.34+
+        ConfigMigrationRule::custom(
+            |layer| {
+                if let Ok(Some(value)) = layer.look_up_item("git.sign-on-push") {
+                    value.as_bool().is_some()
```
yuja, line 632: nit: `.is_bool()`

cli/src/config.rs
```diff
@@ -624,6 +625,34 @@ pub fn default_config_migrations() -> Vec<ConfigMigrationRule> {
...
+                }
+            },
+            |layer| match layer.look_up_item("git.sign-on-push") {
+                Ok(Some(value)) => {
+                    let old_value = value
+                        .as_bool()
+                        .ok_or("git.sign-on-push expects a boolean")
+                        .map_err(|error| ConfigMigrateLayerError::Type {
+                            name: "git.sign-on-push".to_string(),
+                            error: error.into(),
```
yuja, line 644: nit: `.ok_or_else(|| ..`

Perhaps, `.unwrap()` is also good because the type should have been tested.

yuja (approved):

cli/src/commands/git/push.rs
bnjmnt4n: We go through all mutable commits from `(old_heads | immutable_heads)..new_heads`. So I think all ancestors of signed commits would also be passed through the revset filter of `git.sign-on-push` and this shouldn't be a problem?

Of course, the user could use a revset like `description(...)` instead of something like `all()` then there could definitely be unsigned ancestors, but that would be intentional?

yuja: I meant something like this:
```
B [signed]         B                  B' [sign dropped on rebase?]
|              =>  |              =>  |
A [not signed]     A  A' [signed]     A'
```

The original version of `jj sign` had this problem. I haven't checked if `sign-on-push` has the same problem.

bnjmnt4n: I'm not sure if this is covered in the tests. Let me take a look when I have more time.

nilskch:
@bnjmnt4n thanks for the work! Is there anything left to do to land this PR?

cli/src/config-schema.json
```diff
@@ -470,9 +470,8 @@
                     "default": "origin"
                 },
                 "sign-on-push": {
-                    "type": "boolean",
-                    "description": "Whether jj should sign commits before pushing",
-                    "default": false
+                    "type": "string",
+                    "description": "Revset of commits to sign before pushing"
```
nilskch, line 474:
```suggestion
                    "description": "Revset of commits to sign before pushing",
                    "default": "none()"
```
or `~signed() && mine()`, see other comment

cli/src/commands/git/push.rs
```diff
@@ -591,10 +575,8 @@ fn validate_commits_ready_to_push(
...
```
nilskch, line 578: nit: I would drop `!commit.is_signed()` and set the default of `git.sign-on-push` to `~signed() && mine()`.

```suggestion
        if should_sign_on_push(commit.id())? {
```

cli/src/commands/git/push.rs
nilskch: @yuja, this is exactly what happens. All commits after the commit that you signed lose their signature. The question is whether we want to sign commits from others that lose their signature. I think we should not sign commits from others by default.

nilskch (approved):

cli/src/commands/git/push.rs
bnjmnt4n: Not sure if this would make it more annoying for users to customize `git.sign-on-push`. I think it would be acceptable if it's well-documented though.

cli/src/commands/git/push.rs
bnjmnt4n: Added a new commit which tests this.

bnjmnt4n:
Updated this PR to address all feedback now. @yuja (+ anyone else) would you like to take another look?

cli/tests/test_git_push.rs
```diff
@@ -2085,17 +2093,20 @@ fn test_git_push_sign_on_push() {
...
-    Working copy  (@) now at: kmkuslsw 3d5a9465 (empty) commit which should not be signed 2
-    Parent commit (@-)      : kpqxywon 48ea83e9 (empty) commit which should not be signed 1
+      Move forward bookmark bookmark2 from 38a204733702 to 87d9bf2ca4fb
+    Working copy  (@) now at: wqnwkozp eb60bb86 (empty) commit which is signed initially
+    Parent commit (@-)      : kmkuslsw 50a620ed (empty) commit which should not be signed 2
     [EOF]
     ");
-    // Only commits which are being pushed should be signed
+    // Commits which are being pushed should be signed.
+    // The initially signed commit should be resigned if its ancestors are signed.
```
yuja, line 2121: To test this, we'll need to turn the default signing off: `signing.behavior = "drop"`

That said, the next patch moves it to user's problem. The example configuration might have to be something like `roots(~signed()):: & mine()` (or `roots(mutable() & ~signed()):: & mine()` to not check signature of uninteresting commits?)

bnjmnt4n: > To test this, we'll need to turn the default signing off: `signing.behavior = "drop"`

I might be misunderstanding, but If the default signing behavior is `drop`, then the initially signed commit should not be resigned if its ancestors are signed when pushing.

bnjmnt4n: Thanks, I think I misunderstood your original comment. You're referring to the issue where given:

```
A (not signed) -> B (signed)
```

 and B is to be pushed, the original code before this PR could lead to B not being signed if `signing.behavior = "drop"`.

I was testing a different case where:

```
A (not signed) -> B (not signed) -> C (signed)
```

and B is to be pushed, C should still be resigned after A and B are signed if `signing.behavior = "keep"`.

I think I might rewrite the tests to ensure both cases are tested.

nilskch:
What is missing to get this PR over the line? I am really interested in this!

bnjmnt4n:
> What is missing to get this PR over the line? I am really interested in this!

Sorry, I haven't had time to update the tests based on the comments above. I'm not sure when I'll be able to do it either (hopefully sometime in the next few weeks).

yuja:
iirc, the problem I pointed out exists in the original implementation, so I didn't mean it to be a blocker. We can leave TODO comment about that.

PR #5938, opened 2025-03-08
cargo: bump git2 to 0.20.0

emilazy:
Revival of https://github.com/jj-vcs/jj/pull/5315. We have the necessary version in Nixpkgs now, and it looks like Arch does as well.

cc @demize for Gentoo

# Checklist

If applicable:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1073,9 +1073,9 @@ checksum = "40ecd4077b5ae9fd2e9e169b102c6c330d0605168eb0e8bf79952b256dbefffd"
 
 [[package]]
 name = "git2"
-version = "0.19.0"
+version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b903b73e45dc0c6c596f2d37eccece7c1c8bb6e4407b001096387c63d0d93724"
+checksum = "3fda788993cc341f69012feba8bf45c0ba4f3291fcc08e214b4d5a7332d88aff"
 dependencies = [
  "bitflags 2.6.0",
  "libc",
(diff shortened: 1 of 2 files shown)


emilazy:
I’ve dropped the co‐author credit since I recreated this two‐byte dependency change from scratch and the CLA bot seems unhappy with it. But thank you to @chenrui333 for opening the original PR all the same!

PhilipMetzger (approved):

emilazy:
It’s expected that `--HEAD` breaks, right? (Just checking if it’s meant as actionable feedback or a heads up.)

demize:
> if someone who has more context around the libgit2 changes than me wants to fix it

I don't use Homebrew, so I'm not in a position to submit this, but the fix should be simple:

```patch
diff --git a/jj.rb b/jj.rb
index 344de47..77bfaff 100644
--- a/jj.rb
+++ b/jj.rb
@@ -24,12 +24,6 @@ class Jj < Formula

   uses_from_macos "zlib"

-  # patch to use libgit2 1.9, upstream pr ref, https://github.com/jj-vcs/jj/pull/5315
-  patch do
-    url "https://github.com/jj-vcs/jj/commit/b4f936ac302ee835aa274e4dd186b436781d5d2f.patch?full_index=1"
-    sha256 "7b2f84de2c6bbdce9934384af2f7f2d0b7f7116c4726aeef87581010cdf1564e"
-  end
-
   def install
     ENV["LIBGIT2_NO_VENDOR"] = "1"
     ENV["LIBSSH2_SYS_USE_PKG_CONFIG"] = "1"
```

It looks like Homebrew is applying the old PR for this change as a patch; now that this PR (which does the same thing) has landed, the patch is no longer necessary and can be dropped.

emilazy:
(But it is required for the latest release, so the patch would have to be conditional on `--HEAD` or not before being removed on the next stable release.)

chenrui333:
since it is not released, it should be in the head block, either way, I will take of it.

PR #5943, opened 2025-03-09
cli: bookmark move: allow short aliases for `--to`/`--from`

calebdw:
Hello!

If `--to` is going to become a required argument, it should have a short alias as it will be used quite frequently.

Additionally, I would argue in favor of using `-f` for `--from`, given that is it used for other commands and we should be consistent (if users can use `-f` as an alias for `--from` on other commands then logically they would assume they can use it for this one). Not allowing it because users might mistake it to mean `--force` isn't that big of a concern in my book given that `--force` is not used as an option anywhere else (as far as I am aware) and the option requires an argument so it will fail even if a user does try to use it for such a purpose.

If this is acceptable then I can send a follow up PR to add the alias for `--from` (or edit the current commit).

Thanks!

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj op log -d` now has an alias for `jj op log --op-diff`.
 
+* `jj bookmark move --to/--from` can now be abbreviated to `jj bookmark move -t/-f`
+
 ### Fixed bugs
 
 ## [0.27.0] - 2025-03-05
--- a/cli/src/commands/bookmark/move.rs
+++ b/cli/src/commands/bookmark/move.rs
@@ -43,11 +43,8 @@ use crate::ui::Ui;
 #[command(group(clap::ArgGroup::new("source").multiple(true).required(true)))]
 pub struct BookmarkMoveArgs {
     /// Move bookmarks from the given revisions
-    // We intentionally do not support the short `-f` for `--from` since it
-    // could be confused with a shorthand for `--force`, and people might not
-    // realize they need `-B`/`--allow-backwards` instead.
     #[arg(
-        long,
+        long, short,
         group = "source",
         value_name = "REVSETS",
         add = ArgValueCandidates::new(complete::all_revisions),
@@ -56,11 +53,10 @@ pub struct BookmarkMoveArgs {
 
     // TODO(#5374): Make required in jj 0.32+
     /// Move bookmarks to this revision
-    // We intentionally do not support the short `-t` for `--to` since we don't
-    // support `-f` for `--from`. Currently this defaults to the working copy, but in the near
+    // Currently this defaults to the working copy, but in the near
     // future it will be required to explicitly specify it.
     #[arg(
-        long,
+        long, short,
         value_name = "REVSET",
         add = ArgValueCandidates::new(complete::all_revisions),
     )]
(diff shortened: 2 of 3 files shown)


ilyagr:
I think we should either support neither `-f` nor `-t`, or we should support both. So, you should probably add `-f` to this PR as well.

I'm the one who added those comments, and I'm not sure how strongly I feel about this today.

One issue that comes to mind is that `jj bookmark move -f mybookmark` will be interpreted as `--from mybookmark` even if the user wanted to use `--force` and thought of `mybookmark` as a "names" argument. 

BTW, as an aside, I was hoping that most people would end up typing `--f<TAB>`, which is 4 key presses (compared to `-f` being 2 and `--force` being 6; `--f<TAB>` also removes the need for pressing space afterwards if it's not the last argument). I think I often type out `--from`, though, and I don't mind it too much.

yuja:
Just add both `-f` and `-t`? We already have a couple of places where `-f` means `--from`, and I don't think `jj bookmark move` is special.

`jj bookmark -f name` will be a hard error (if `--to` becomes mandatory.)

calebdw:
I originally had added the `-f` alias, but I removed it because my primary concern was being able to quickly type `-t` for `--to` (yes I'm lazy :upside_down_face:) and I didn't want this PR being rejected for the `-f`. However, as there appears to be concurrence, I've re-added the `-f` alias.

> I think we should either support neither -f nor -t, or we should support both. 

While I agree that we should support `-f`, I would still like to support `-t` even if we decide not to go with `-f`.

> We already have a couple of places where -f means --from, and I don't think jj bookmark move is special.

This is my biggest argument---nothing makes this particular command special and users are going to expect that they can use `-f`/`-t` for `--from`/`--to` as they can for other commands. Not supporting it is only going to cause additional issues and questions (what makes this command special, etc.). Consistency in aliases and argument names across the different commands is key.

ilyagr:
For completeness, I'll post the unfinished draft of a response I was writing. Ultimately, I'm fine with you guys making a decision, both on --from and on whether it makes sense to have -t without -f. It sounds like Yuya is fine with the later, and I'm happy to defer to him on this 

-------

I'm still on the fence. At the moment, I continue to think that allowing `-f` does have the potential to cause minor confusion, but I'm not sure it's significant enough to outweigh the convenience of the shortcut. 

IIRC, I was also concerned that `jj bookmark move` would often be used right before a `jj git push`, so if `jj bookmark move` does the wrong thing without the user noticing, they could make an error that's difficult to recover from.

> `jj bookmark -f name` will be a hard error (if `--to` becomes mandatory.)

I don't think `--to` becoming mandatory would change much. The example would just be[^nice] `jj bookmark move --to @ -f mybookmark` rather than `jj bookmark move -f mybookmark`. 

[^nice]: Though, I guess, it's nice that `jj b m -f --to @ mybookmark`  would be an error.

Some possible confusion this would lead to:

1. If there is another bookmark at the same revision as `mybookmark`, the user might not notice that the command moves both bookmarks. If they thought `-f` meant `--force`, they'd expect only `mybookmark` to move. I *think* we do print a warning when multiple bookmarks are moved, so if they notice that, it's less of a problem.

2. If `mybookmark` is not an ancestor of `@`, the user might not understand why this failed. Hopefully, the error message will help in this case.

ilyagr:
OK, let's go with that then.

calebdw:
Sounds good, thanks folks!

PR #5944, opened 2025-03-09
revset globs: `substr-i` as an alias for `substring-i:`

ilyagr:
My instinct is to try `iglob:` instead of the
correct `glob-i:`.

@yuja, is this correct in every case? I found it difficult to find where the `-i` postfix is parsed.

* [x] Add to CHANGELOG if we keep the second commit

### DIFF
--- a/cli/src/command_error.rs
+++ b/cli/src/command_error.rs
@@ -864,7 +864,7 @@ fn revset_resolution_error_hint(err: &RevsetResolutionError) -> Option<String> {
 fn string_pattern_parse_error_hint(err: &StringPatternParseError) -> Option<String> {
     match err {
         StringPatternParseError::InvalidKind(_) => Some(
-            "Try prefixing with one of `exact:`, `glob:`, `regex:`, `substring:`, or one of these \
+            "Try prefixing with one of `exact:`, `glob:`, `regex:`, `substr:`, or one of these \
              with `-i` suffix added (e.g. `glob-i:`) for case-insensitive matching"
                 .into(),
         ),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj bookmark move --to/--from` can now be abbreviated to `jj bookmark move -t/-f`
 
+* In string patterns for revsets or bookmarks, `substr:` is now a shortcut for
+  `substring:`.
+
 ### Fixed bugs
 
 
--- a/cli/tests/test_bookmark_command.rs
+++ b/cli/tests/test_bookmark_command.rs
@@ -850,7 +850,7 @@ fn test_bookmark_delete_glob() {
     error: invalid value 'whatever:bookmark' for '<NAMES>...': Invalid string pattern kind `whatever:`
 
     For more information, try '--help'.
-    Hint: Try prefixing with one of `exact:`, `glob:`, `regex:`, `substring:`, or one of these with `-i` suffix added (e.g. `glob-i:`) for case-insensitive matching
+    Hint: Try prefixing with one of `exact:`, `glob:`, `regex:`, `substr:`, or one of these with `-i` suffix added (e.g. `glob-i:`) for case-insensitive matching
     [EOF]
     [exit status: 2]
     ");
--- a/cli/tests/test_commit_template.rs
+++ b/cli/tests/test_commit_template.rs
@@ -1094,7 +1094,7 @@ fn test_log_contained_in() {
       |
       = Invalid string pattern
     3: Invalid string pattern kind `x:`
-    Hint: Try prefixing with one of `exact:`, `glob:`, `regex:`, `substring:`, or one of these with `-i` suffix added (e.g. `glob-i:`) for case-insensitive matching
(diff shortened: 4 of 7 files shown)


ilyagr:
Another thought: `substring:` is long to type. WDYT about making `sub:` a synonym for it?

cli/src/command_error.rs
```diff
@@ -876,9 +876,11 @@ fn revset_resolution_error_hint(err: &RevsetResolutionError) -> Option<String> {
 
 fn string_pattern_parse_error_hint(err: &StringPatternParseError) -> Option<String> {
     match err {
-        StringPatternParseError::InvalidKind(_) => {
-            Some("Try prefixing with one of `exact:`, `glob:`, `regex:`, or `substring:`".into())
-        }
+        StringPatternParseError::InvalidKind(_) => Some(
+            "Try prefixing with one of `exact:`, `glob:`, `regex:`, `substring:`, or one of these \
+             with `-i` added (e.g. `glob-i:`) for case-insensitive matching"
```
yuja: nit: "with `-i` suffix" might be better

cli/src/command_error.rs
```diff
@@ -877,8 +877,8 @@ fn revset_resolution_error_hint(err: &RevsetResolutionError) -> Option<String> {
 fn string_pattern_parse_error_hint(err: &StringPatternParseError) -> Option<String> {
     match err {
         StringPatternParseError::InvalidKind(_) => Some(
-            "Try prefixing with one of `exact:`, `glob:`, `regex:`, `substring:`, or one of these \
-             with `-i` added (e.g. `glob-i:`) for case-insensitive matching"
+            "Try prefixing with one of `exact:`, `glob:`, `regex:`, `sub:`, or one of these with \
+             `-i` added (e.g. `glob-i:`) for case-insensitive matching"
```
yuja, line 881: I don't know if `sub:` is common short name for substring. Even if it is, I wouldn't change the error message to use the short form.

ilyagr: I guess `substr` and `substr-i` is another option.

Yet another option would be to allow single-letter versions for all of them: `e:`, `g:`, `e-i:`, `r:`, etc.

My thinking was to eventually softly (or not-so-softly) deprecate the long `substring:` with `sub:` (or `substr:`). But I don't feel strongly about it.

For something like `description()`, shortening `substring-i:` to `i:` would make sense. However, as string patterns default to `exact:` in bookmark names, it would be confusing in that case.

yuja: `substr` sounds better than `sub`. It wouldn't save much typing, though.

iirc, there was an idea to making `glob:` the default in both CLI and revsets, Then, `i:` could be aliased to `glob-i:`?

emilazy: Yes, I still like the `glob:`/`i:` plan. We could maybe even do `regex:` these days but I think `glob:` would be less invasive.

ilyagr: I think `substr` is better than `substring`, so I'll do that. Let me know if you prefer that I get rid of that commit entirely (or move it to a separate PR).

For the `glob:` and "i:" plan, remind me, does `jj bookmark move feat` match all strings containing `feat` or only the string "feat" exactly? (We should document it either way, I could add a commit). In the latter case, I think I like it; it's what I've been assuming naturally before reading the docs. Though, of course, some users who like substring matches will be inconvenienced.

However, if that happens, we have the option of adding an `s:` alias for `substring:`.

yuja: > I think `substr` is better than `substring`, so I'll do that. Let me know if you prefer that I get rid of that commit entirely (or move it to a separate PR).

I don't mind if you think `substr` is strictly better. My feeling is that `substr` isn't so different from `substring`, and I personally would avoid paying extra (documenting) cost for that.

> For the `glob:` and `i:` plan, remind me, does `jj bookmark move feat` match all strings containing `feat` or only the string "feat" exactly?

The latter. There will be a bigger behavior change in revset. You'll need to write `subject('*foo*')` instead of `subject(foo)`. (The parser might have to accept unquoted `*foo*`.)

martinvonz: I prefer `substring` as the canonical form but I don't mind adding `substr` as a hidden alias for `substring` to help people who are used to that form from standard library (e.g. C++ or JavaScript). Actually, if we wanted to match standard libraries, it seems like `contains` is a better match (I think `substr()` usually a range of indexes as input).

emilazy: I like `contains:`. I think I proposed that around when we were talking about defaults. I forget what the reception was.

emilazy: Yeah, I’d personally like to try `glob:` as default across the board and `i:` as short for `glob-i:`, and then see if we still feel the need for shorter names. I intended to push for this last year but didn’t get around to it. IIRC there was some concern about being able to refer to bookmarks with `*` in them or the like, but I think that’s probably not a big deal.

Potential issues with default `glob:` I can foresee:

* Searching for commit messages with a `?` in them. We could/should probably turn off `?` if it’s an option, since it’s pretty niche.
* Searching for commit messages with `[…]` in them, e.g. `[topic] …` instead of `topic: …`. Not sure if there’s anything sensible we could do about that.

ilyagr:
I'll close this for now, it seems that nobody is very keen on it. We can reopen something similar later as needed.

PR #5949, opened 2025-03-09
tests: add workspace test helper, use it in test_global_opts.rs

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/common/test_environment.rs
+++ b/cli/tests/common/test_environment.rs
@@ -76,6 +76,16 @@ impl Default for TestEnvironment {
 }
 
 impl TestEnvironment {
+    /// Returns test helper for the specified directory.
+    ///
+    /// The `root` path usually points to the workspace root, but it may be
+    /// arbitrary path including non-existent directory.
+    #[must_use]
+    pub fn work_dir(&self, root: impl AsRef<Path>) -> TestWorkDir<'_> {
+        let root = self.env_root.join(root);
+        TestWorkDir { env: self, root }
+    }
+
     /// Runs `jj args..` in the `current_dir`, returns the output.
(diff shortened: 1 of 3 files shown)


cli/tests/common/test_environment.rs
```diff
@@ -255,6 +258,74 @@ impl TestEnvironment {
     }
 }
 
+/// Helper to execute `jj` or file operation in sub directory.
+pub struct TestWorkDir<'a> {
+    env: &'a TestEnvironment,
+    root: PathBuf,
+}
+
+impl TestWorkDir<'_> {
```
martinvonz, line 267: Do you imagine that we'll eventually remove `run_jj()` etc on `TestEnvironment` itself require running all commands via a `TestWorkDir`? I suppose the existing uses of `test_env.run_jj("foo");` can now be written as `let root_work_dir = test_env.work_dir(""); root_work_dir.run_jj("foo");`

martinvonz (approved):

cli/tests/common/test_environment.rs
yuja: That could be, but I'm not sure. If there aren't many instances of `.run_jj_in()` (except for the workspace initialization), it might make sense to remove these.

martinvonz: Actually, even if we keep the methods on `TestEnvironment`, maybe we can make them just convenience methods for `.root_work_dir().<method>`?

yuja: That sounds like a good idea. I think we can decide it later.

martinvonz: Yes, don't let this comment block

Issue #3650, opened 2024-05-08
FR: `revsets.push`


cdmistman:
**Is your feature request related to a problem? Please describe.**
basically, #2434. except i have a custom revset i'd like to use instead of whatever the default may end up being

**Describe the solution you'd like**
a new config option akin to `revsets.log`: `revsets.push`

**Describe alternatives you've considered**
since push is backend-specific behavior, it might be better to make the new config key `git.push-revset` instead. (NB would be nice to do some config squashing, ie `git.push.revset`, but that's a larger change)

PR #5950, opened 2025-03-09
cli: git push: add --skip-private flag

### LINKED ISSUE #5936: FR: Warn only when not pushing private commits

I am using the megamerge worklow with some local-only and some public changes on parallel branches. I have a bookmark pointing to the local-only change so I can refer to it easily. I would like to `jj git push` my other bookmark(s) with the public changes, but jj hard-errors on the one bookmark with local changes.

Only emit a warning with the list of bookmarks that weren't pushed, because they contain private commits. Still push the other bookmarks.

Maybe it's possible to configure a default push revset that excludes any bookmarks that contain private commits? I wouldn't know how to do that though.

I might also be holding it wrong™, please tell me if you have a different workflow for working with local-only and public changes in parallel, without running into this problem.

**Additional context**

repro steps:

```sh
# init temp repo with some remote
jj new -m a -r 'root()' ; jj bookmark create a --to @
jj new -m b -r 'root()' ; jj bookmark create b --to @
jj config set --repo "git.private-commits" '"description(\'b\')"'
jj new a b
jj git push
```

senekor:
# Checklist

closes #5936


senekor:
example setup:

```sh
# init temp repo with some remote
jj new -m a -r 'root()' ; jj bookmark create a --to @
jj new -m b -r 'root()' ; jj bookmark create b --to @
jj config set --repo "git.private-commits" '"description(\'b\')"'
jj new a b
jj git push
```

```
$ jj git push
Error: Won't push commit 8e7ff255356c since it is private
Hint: Rejected commit: mvnnt 8e7ff255 b | (empty) b
Hint: Configured git.private-commits: 'description('b')'
Hint: Consider --skip-private to skip pushing bookmarks with private commits

$ jj git push --skip-private
Updated signatures of 1 commits
Rebased 1 descendant commits
Changes to push to origin:
  Add bookmark a to d9d8f996eab7
  Skipped bookmark b with private commits
Working copy now at: mlsmt b65a6c55 (empty) (no description set)
Parent commit      : swkyu d9d8f996 a | (empty) a
Parent commit      : mvnnt 8e7ff255 b | (empty) b
```

yuja:
Isn't it better to make the default revset configurable?

Maybe it's okay that `jj git push --skip-private` excludes private commits, but thigs would be weird if bookmarks explicitly specified by `-b NAME` or created by `-c` were skipped.

A major problem of configurable revset is that we cannot parameterize it by remote name. I don't have a good solution for that, but it might be okay to abuse the alias machinery.

senekor:
> Isn't it better to make the default revset configurable?

I'm not sure I would prefer it to solve this problem. It would mean I have to duplicate my logic for private commits in my config. Unless we add a function `private()` or similar to the revset language to reuse the logic. That would also solve my problem.

If anything, I would still consider just doing this without any flags. _By default_, I probably don't want to push private commits, right? I'm struggling to see the use case for blocking the pushing of perfectly eligible bookmarks, just because some private bookmark also happens to be somewhere in the push revset. I like the idea of private commits, but I would rather they stay out of my way instead of making the normal workflow more tedious.

> Maybe it's okay that `jj git push --skip-private` excludes private commits, but things would be weird if bookmarks explicitly specified by `-b NAME` or created by `-c` were skipped.

Well, that would require the user to run a weird command. `jj git push -b some-bookmark --skip-private` is weird. I think it's ok for jj to behave in a compliant (and thus weird) way if the user makes a weird request.

yuja:
> I'm not sure I would prefer it to solve this problem. It would mean I have to duplicate my logic for private commits in my config.

A common pattern is to add `private()` alias and use it in `git.private-commits`.

> If anything, I would still consider just doing this without any flags. _By default_, I probably don't want to push private commits, right? I'm struggling to see the use case for blocking the pushing of perfectly eligible bookmarks, just because some private bookmark also happens to be somewhere in the push revset.

I'm not against changing the default. I wouldn't want to add new flag unless both use cases are common.

> > Maybe it's okay that `jj git push --skip-private` excludes private commits, but things would be weird if bookmarks explicitly specified by `-b NAME` or created by `-c` were skipped.
> 
> Well, that would require the user to run a weird command. `jj git push -b some-bookmark --skip-private` is weird. I think it's ok for jj to behave in a compliant (and thus weird) way if the user makes a weird request.

It would become weird if the `--skip-private` behavior was the default, or was added to user alias.

senekor:
> A common pattern is to add `private()` alias and use it in `git.private-commits`.

Ah, didn't think of that. I can use this + an alias to solve my problem right now, thanks!

> It would become weird if the `--skip-private` behavior was the default, or was added to user alias.

That's true. I think it would make sense to skip private commits for the default revset and `--all` only. If any of `-r`, `-b`, `-c` are supplied, a hard error makes sense.

@PhilipMetzger do you agree with this, or do you think the hard error should also stay the default behavior for the default revset and `--all`? If so, could you elaborate a little why that would be desired?

PhilipMetzger:
I think I mostly agree with Yuya. I also have a local change which would allow you to customize the `push` revset (#3650), but it currently fails to express the default, since it is parameterized on the chosen remote.

While I still think it is useful, we can shelve this until more people show up with the same problem as yours and then reconsider.

senekor:
I'm going to close this, since it can be solved with a bit of configuration already. A configurable default push revset seems like the cleaner solution.

PR #5955, opened 2025-03-09
git: port Git state clean‐up from `git2`

emilazy:
# Checklist

If applicable:


lib/src/git.rs
```diff
@@ -1148,13 +1148,32 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
...
+            "MERGE_MODE",
+            "MERGE_MSG",
+            "REVERT_HEAD",
+            "CHERRY_PICK_HEAD",
+            "BISECT_LOG",
+        ];
+        const STATE_DIR_NAMES: &[&str] = &["rebase-merge", "rebase-apply", "sequencer"];
+        let handle = |result: std::io::Result<()>| match result {
+            Err(err) if err.kind() == std::io::ErrorKind::NotFound => Ok(()),
+            _ => result.map_err(GitExportError::from_git),
```
yuja, line 1169: nit: better to wrap in `PathError` to provide context

lib/src/git.rs
```diff
@@ -1148,13 +1148,32 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
...
+        };
+        for file_name in STATE_FILE_NAMES {
+            handle(std::fs::remove_file(git_repo.path().join(file_name)))?;
+        }
+        for dir_name in STATE_DIR_NAMES {
+            handle(std::fs::remove_dir_all(git_repo.path().join(dir_name)))?;
```
yuja, line 1175: nit: I would use `.map_err(handle)`.

yuja (approved):

lib/src/git.rs
emilazy: Do you think this should be `GitExportError::from_git(PathError { … })` or a new `GitExportError::PathError`? I’ve done the former (since it is a “Git error” and matches how it’ll be reported once this is upstreamed to `gix`) but I can do the latter if preferred.

lib/src/git.rs
emilazy: It can’t be a `.map_err` because it needs to turn `NotFound` errors into `Ok`. But it can be an `or_else`, which I’ve done.

lib/src/git.rs
yuja: I also mean the former. Thanks.

lib/src/git.rs
yuja: Right. Looks good.

ilyagr:
Oh wow, it looks like GitHub got smart enough not to insert the first commit a second time (it was already merged in another PR).

image

emilazy:
That’s because I targeted that PR’s branch with this PR. GitHub supports rudimentary PR stacking if you do that; when the first PR is merged, the second will retarget to the first PR’s target branch.

ilyagr:
Unfortunately, it does not seem smart enough to start showing this PR as having 2 commits on this webpage. :(

But good to know that GitHub applies some intelligence to the merging.

emilazy:
I think that may be because we use rebase merges, so the commit identities are different. I’m not sure though.

Issue #1153, opened 2023-01-30
`jj` should warn about unrecognized configuration settings


arxanas:
## Description

See https://discord.com/channels/968932220549103686/969829516539228222/1069406648600375347 for an example, where the user set `ui.graph.format` but actually wanted `ui.graph.style`. `jj` should preemptively warn about configuration settings that don't exist, or perhaps at least have a command to lint your configuration files.

## Steps to Reproduce the Problem

1.  Set a value for a configuration setting that doesn't exist.

## Expected Behavior

- `jj` should warn about unused configuration settings. (`cargo` does this, which is helpful.)

## Actual Behavior

- Configuration setting is unused without any diagnostic.

arxanas:
Possibly interesting, although I don't know if we use `serde`: https://github.com/dtolnay/serde-ignored

Ralith:
This seems very difficult/fragile so long as jj is using ad-hoc strings to identify configuration options, especially across the lib/cli boundary. Moving towards a statically defined struct of all options used in the CLI, combined with serde-ignored seems like a good approach.

ilyagr:
In case we stick with strings, there are some imperfect but possibly good enough solutions.

Thinking out loud, https://github.com/martinvonz/jj/blob/main/src/commands/config-schema.json could be helpful here.

Or, we could have a (possibly imperfect) whitelist of acceptable settings and warn when we see a setting outside the whitelist. We could rely on tests and the default config for this whitelist not to become stale.

The whitelist would have to recognize, for example, that `[merge-tools]` can contain a key with any name.

If this whitelist can be expressed as a Rust type, we're back to the statically defined struct idea and should stick with that. A whitelist is more flexible in some ways, though. For example, in tests you could expand the whitelist to include test-only options.

samueltardieu:
The other difficulty in using `serde`, even with `serde-ignored`, is that in the future we might want to be able to write the configuration file (think `jj config --global user.email foo@example.com`). We would need to write back the unknown keys and values because they might be used by external tools used along with `jj`.

dbarnett:
@martinvonz have you done any thinking on how to extend the config with custom company-internal config options, similar to #1069 for custom global flags? Would that involve patching/linking different rust code, or something else?

[...]

PR #5956, opened 2025-03-09
git: make `git2` support optional

emilazy:
This helps us prepare for removing the functionality down the line and makes things easier for people building or packaging their own Jujutsu.

# Checklist

If applicable:

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -2053,7 +2053,7 @@ impl<'a> GitFetchImpl<'a> {
                 GitSubprocessContext::from_git_backend(git_backend, &git_settings.executable_path);
             Ok(GitFetchImpl::Subprocess { git_repo, git_ctx })
         } else {
-            let git_repo = git_backend.open_git_repo()?;
+            let git_repo = git2::Repository::open(git_backend.git_repo_path())?;
             Ok(GitFetchImpl::Git2 { git_repo })
         }
     }
@@ -2351,7 +2351,7 @@ pub fn push_updates(
             callbacks,
         )
     } else {
-        let git_repo = git_backend.open_git_repo()?;
+        let git_repo = git2::Repository::open(git_backend.git_repo_path())?;
         let refspecs: Vec<String> = refspecs.iter().map(RefSpec::to_git_format).collect();
         git2_push_refs(
             repo,
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -18,6 +18,7 @@ use std::borrow::Borrow;
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+#[cfg(feature = "git2")]
 use std::collections::HashSet;
 use std::default::Default;
 use std::fmt;
@@ -41,6 +42,7 @@ use crate::file_util::PathError;
 use crate::git_backend::GitBackend;
 use crate::git_subprocess::GitSubprocessContext;
 use crate::git_subprocess::GitSubprocessError;
+#[cfg(feature = "git2")]
 use crate::index::Index;
 use crate::merged_tree::MergedTree;
 use crate::object_id::ObjectId;
@@ -48,6 +50,7 @@ use crate::op_store::RefTarget;
 use crate::op_store::RefTargetOptionExt;
 use crate::op_store::RemoteRef;
 use crate::op_store::RemoteRefState;
+#[cfg(feature = "git2")]
 use crate::refs;
 use crate::refs::BookmarkPushUpdate;
 use crate::refs::RemoteRefSymbol;
@@ -1411,6 +1414,7 @@ impl GitRemoteManagementError {
     }
 }
 
+#[cfg(feature = "git2")]
 fn is_remote_not_found_err(err: &git2::Error) -> bool {
     matches!(
         (err.class(), err.code()),
@@ -1875,6 +1879,7 @@ pub enum GitFetchError {
     #[error(transparent)]
     RemoteName(#[from] GitRemoteNameError),
     // TODO: I'm sure there are other errors possible, such as transport-level errors.
+    #[cfg(feature = "git2")]
     #[error("Unexpected git error when fetching")]
     InternalGitError(#[from] git2::Error),
     #[error(transparent)]
@@ -1885,12 +1890,14 @@ pub enum GitFetchError {
 // UnexpectedGitBackendError.
 #[derive(Debug, Error)]
 pub enum GitFetchPrepareError {
+    #[cfg(feature = "git2")]
     #[error(transparent)]
     Git2(#[from] git2::Error),
     #[error(transparent)]
     UnexpectedBackend(#[from] UnexpectedGitBackendError),
 }
 
+#[cfg(feature = "git2")]
 fn git2_fetch_options(
     mut callbacks: RemoteCallbacks<'_>,
     depth: Option<NonZeroU32>,
@@ -2035,9 +2042,8 @@ fn expand_fetch_refspecs(
 }
 
 enum GitFetchImpl<'a> {
-    Git2 {
-        git_repo: git2::Repository,
-    },
+    #[cfg(feature = "git2")]
+    Git2 { git_repo: git2::Repository },
     Subprocess {
         git_repo: Box<gix::Repository>,
         git_ctx: GitSubprocessContext<'a>,
@@ -2047,15 +2053,15 @@ enum GitFetchImpl<'a> {
 impl<'a> GitFetchImpl<'a> {
     fn new(store: &Store, git_settings: &'a GitSettings) -> Result<Self, GitFetchPrepareError> {
         let git_backend = get_git_backend(store)?;
-        if git_settings.subprocess {
-            let git_repo = Box::new(git_backend.git_repo());
-            let git_ctx =
-                GitSubprocessContext::from_git_backend(git_backend, &git_settings.executable_path);
-            Ok(GitFetchImpl::Subprocess { git_repo, git_ctx })
-        } else {
+        #[cfg(feature = "git2")]
+        if !git_settings.subprocess {
             let git_repo = git2::Repository::open(git_backend.git_repo_path())?;
-            Ok(GitFetchImpl::Git2 { git_repo })
+            return Ok(GitFetchImpl::Git2 { git_repo });
         }
+        let git_repo = Box::new(git_backend.git_repo());
+        let git_ctx =
+            GitSubprocessContext::from_git_backend(git_backend, &git_settings.executable_path);
+        Ok(GitFetchImpl::Subprocess { git_repo, git_ctx })
     }
 
     fn fetch(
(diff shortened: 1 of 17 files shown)


demize:
I'm not 100% sure on why, but in my testing this out as an ebuild, I'm running into this while building **with** `git2`:

```
>>> Install dev-vcs/jj-9999 into /var/tmp/portage/dev-vcs/jj-9999/image
 * /usr/lib/rust/9999/bin/cargo install --root /var/tmp/portage/dev-vcs/jj-9999/image/usr --path cli
  Installing jj-cli v0.27.0 (/var/tmp/portage/dev-vcs/jj-9999/work/jj-emilazy-push-xvnqslxuupmv/cli)
error: failed to compile `jj-cli v0.27.0 (/var/tmp/portage/dev-vcs/jj-9999/work/jj-emilazy-push-xvnqslxuupmv/cli)`, intermediate artifacts can be found at `/var/tmp/portage/dev-vcs/jj-9999/work/jj-emilazy-push-xvnqslxuupmv/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

My wild speculation as to what's causing this is the addition of "testutils/git2" to the features enabled by "git2". It *builds* fine (though I think I'm running into issues with the tests), but then fails to install.

Builds and installs fine without git2, so that's nice, though I'm also running into issues with the tests there (the same one that triggered the CI failure, and also some others I think may be unrelated regressions since 0.27 that I need to look into more).

emilazy:
Pushed a horrible but apparently working fix for @demize’s issue.

lib/src/git.rs
```diff
@@ -1837,12 +1841,17 @@ pub enum GitFetchError {
 // UnexpectedGitBackendError.
 #[derive(Debug, Error)]
 pub enum GitFetchPrepareError {
+    #[cfg(feature = "git2")]
     #[error(transparent)]
     Git2(#[from] git2::Error),
+    #[cfg(not(feature = "git2"))]
+    #[error("jj was compiled without `git.subprocess = false` support")]
+    Git2Unsupported,
```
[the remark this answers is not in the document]
emilazy: Summary of discussion on Discord:

* When we start the deprecation process, it seems like it would be noisy to warn/error on every single command and not something we have precedent for, rather than just for `jj git fetch`/`jj git push`.
* When we remove support entirely we’ll probably want `jj git fetch`/`jj git push` to be a hard error with migration information when `git.subprocess = false` is set for a while, which would lead to the exact same code here.
* For people who disappear for years and then try to push again after we’ve removed all traces of the support, resolving https://github.com/jj-vcs/jj/issues/1153 would help warn about the unknown configuration setting.

However I could still be convinced otherwise about this.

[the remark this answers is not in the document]
bsdinis: i think this is good

emilazy:
Rebased to make GitHub less confused about how many commits there are here. (And also to get a measurement for how much longer the extra job will make CI.)

emilazy:
Looks like this shouldn’t lengthen CI times appreciably. Hopefully we can cut down on the redundant `push`/`pull_request` checks anyway.

This should be ready for review/merge now.

lib/testutils/src/lib.rs
```diff
@@ -72,23 +71,30 @@ use crate::test_backend::TestBackendFactory;
...
+        CONFIGURE_GIT2.call_once(|| unsafe {
+            git2::opts::set_search_path(git2::ConfigLevel::System, "").unwrap();
+            git2::opts::set_search_path(git2::ConfigLevel::Global, "").unwrap();
+            git2::opts::set_search_path(git2::ConfigLevel::XDG, "").unwrap();
+            git2::opts::set_search_path(git2::ConfigLevel::ProgramData, "").unwrap();
+        });
+    }
 
-    // Prevent GitBackend from loading user and system configurations. For
+    // Prevent `git(1)` from loading user and system configurations. For
```
yuja, line 97: This is mainly for `gix` to not load user configuration.

lib/src/git.rs
```diff
@@ -2199,8 +2216,12 @@ pub enum GitPushError {
     RefUpdateRejected(Vec<String>),
     // TODO: I'm sure there are other errors possible, such as transport-level errors,
     // and errors caused by the remote rejecting the push.
+    #[cfg(feature = "git2")]
     #[error("Unexpected git error when pushing")]
     InternalGitError(#[from] git2::Error),
+    #[cfg(not(feature = "git2"))]
+    #[error("jj was compiled without `git.subprocess = false` support")]
+    Git2Unsupported,
```
yuja, line 2224: Are we going to add test for "git2 unsupported" error?

If not, it would be simpler to just assume `git2` is enable in tests. I think we can still ensure that the `jj` binary can be built without git2.

yuja (approved):

lib/src/git.rs
emilazy: When working on this initially I got a fully‐functional `jj(1)` building, but when I tried to run tests I found the `git2` stuff I had yet to remove. So I wanted to make sure we test a `git2`‐less build in CI so that we don’t regress on that front and to make it easier to drop the dependency.

In terms of the CI burden, it builds in strictly less time than the normal build, so all it takes up is one concurrency slot. After https://github.com/jj-vcs/jj/pull/6005, most pushes to PRs will be using half the concurrency slots they did before, so I hope that one additional build is okay considering it’s only temporary and for one platform.

I’ve added both `jj-lib` and `jj-cli` tests for this now, though, because it does actually raise an unresolved question about the deprecation messaging that I brought up on Discord recently. I presume that when we start the deprecation process for `git.subprocess = false`, we’ll add warnings to `jj git fetch` and `jj git push` when it’s set. But when the deprecation warning period ends and we actually remove the support after that, how should it be handled? Possibilities include:

1. Print a warning about the deprecated setting on every command, like the current configuration migrations do, and otherwise act like `git.subprocess = true`.

2. Print a warning about the deprecated setting only on `jj git fetch`/`jj git push`, and otherwise act like `git.subprocess = true`.

3. Print a hard deprecation error on `jj git fetch`/`jj git push` to let people know the path is no longer supported.

4. Just remove everything that looks at `git.subprocess = false` and have it be silently ignored.

Since half of the point of this PR is to prepare for that deprecation process, I want to make sure the implementation matches what we decide for that. Currently it does (3), but maybe one of the other options would be preferable? (In the long run I assume we’ll remove all code handling it and the end state will be (4), but we might want to still notify users who have set `git.subprocess = false` for a while even once the `git2` dependency is gone. Maybe by the time it becomes (4) we’ll warn about all unknown configuration settings.)

bsdinis: I lean on 1. But happy to do 3 either.

yuja: > When working on this initially I got a fully‐functional `jj(1)` building, but when I tried to run tests I found the `git2` stuff I had yet to remove. So I wanted to make sure we test a `git2`‐less build in CI so that we don’t regress on that front and to make it easier to drop the dependency.
 
I thought it would be enforced by `LIBGIT2_NO_VENDOR=1`, but I don't have strong reason against adding a bit of feature mess to Cargo.toml. Hopefully we can remove it soon.

> 1. Print a warning about the deprecated setting on every command, like the current configuration migrations do, and otherwise act like `git.subprocess = true`.
> 2. Print a warning about the deprecated setting only on `jj git fetch`/`jj git push`, and otherwise act like `git.subprocess = true`.
> 3. Print a hard deprecation error on `jj git fetch`/`jj git push` to let people know the path is no longer supported.
> 4. Just remove everything that looks at `git.subprocess = false` and have it be silently ignored.

I think (4) is good because the user should already know that `git.subprocess = false` was going to be removed soon.

emilazy: > I thought it would be enforced by `LIBGIT2_NO_VENDOR=1`, but I don't have strong reason against adding a bit of feature mess to Cargo.toml. Hopefully we can remove it soon.

Just doing a build would ensure the library and binary code is ready for `git2` removal, but not the test code. Since it took more effort than expected to get the test code ready for that, I wanted to ensure CI continues to check it. We could potentially build the tests for the `git2`‐less variant without actually running them, but it would be more complexity in the GitHub Actions workflow and still require the Cargo suff. Since it should never be the bottleneck, hopefully this is fine for now? I agree that the Cargo stuff is annoying and will be glad to see it go. OTOH, we’d need it if we ever wanted tests to work without the `git` feature. Fixing the `testutils` circular dependency would also make the Cargo stuff less weird.

> I think (4) is good because the user should already know that `git.subprocess = false` was going to be removed soon.

Alright, I’ve adjusted the code for (4). It does mean that `git.subprocess = true` doesn’t warn when the `git2` feature is disabled, but the `git2`‐less build isn’t intended for widespread distribution anyway.

yuja: > Just doing a build would ensure the library and binary code is ready for `git2` removal, but not the test code. Since it took more effort than expected to get the test code ready for that, I wanted to ensure CI continues to check it.

I personally think it's okay if test code require a bit more work for git2 removal, which downstream packages wouldn't have to care much about?, but I don't have strong feeling. The code looks good overall.

> > I think (4) is good because the user should already know that `git.subprocess = false` was going to be removed soon.
> 
> Alright, I’ve adjusted the code for (4). It does mean that `git.subprocess = true` doesn’t warn when the `git2` feature is disabled, but the `git2`‐less build isn’t intended for widespread distribution anyway.

Just to be clear, we'll add a warning (maybe in CLI code) when we decide to deprecate the `git2` support. I think it will be displayed even if `git2` feature is disabled.

emilazy: FWIW, at least the Nix and Gentoo packaging compile and run the tests. So those working correctly without `git2` is necessary for downstream packagers to be able to take advantage of this. (We don’t plan to ship without `git2` in Nixpkgs until it’s removed upstream, though, but I think @demize intends to add a `USE` flag to opt out of it, in the typical Gentoo style.)

> Just to be clear, we'll add a warning (maybe in CLI code) when we decide to deprecate the `git2` support. I think it will be displayed even if `git2` feature is disabled.

Right. I actually have all that implemented in follow‐up commits I’ll open draft PRs for to get ready for whenever we want to start the deprecation process. However, I realized it might be quite annoying to debug a problem with `git.subprocess = false` being ignored in a `git2`‐less build and it’s quite easy to add a warning using the configuration migration system, so I’ve done that and added tests for it. It’ll warn on every command if it’s set, but that seems fine for an opt‐in build flag. (FWIW, it ended up being easier to warn only with `git2` support enabled with my implementation strategy for the deprecation, so this also helps cover that case too.)

If you think that looks fine then I’ll merge this.

demize: > I think @demize intends to add a `USE` flag to opt out of it, in the typical Gentoo style.

That’s the plan—I’ve got most of the work done, and the live ebuild will be updated once this is merged, with the USE flag added to the (unstable) 0.28 ebuild once 0.28’s released :)

yuja: Sure, the code looks good to me overall.

PR #5959, opened 2025-03-10
cli: git fetch: allow short alias for --all-remotes

calebdw:
Allows using `-a` as a short alias for `--all-remotes` in the `git fetch` command.

Thanks!

# Checklist

If applicable:


yuja (commented):
I think `-a`/`--all` in this context is confusing because it could potentially mean "all branches"/"bookmarks" (if we had some config knob for the default fetch spec.)

calebdw:
Even git uses `--all` to fetch from all remotes....I'm not sure that I would think of `fetch` as fetching all bookmarks/branches?

Granted, git uses `-a` for `--append`:

```
    -a, --[no-]append     append to .git/FETCH_HEAD instead of overwriting
```

PR #3447, opened 2024-04-04
cargo: emit less debug info, and compile all deps in optimized mode

thoughtpolice:
See commit messages for details.

PR #5960, opened 2025-03-10
nix: refactor derivation

RossSmyth:
Refactor's the Nix derivation. The main motivation for this is ease of cross-compilation to Windows. But I cleaned up some other things:

1. Added a "ci" profile that eliminates most debug info (not all so that tracebacks are useful), leading to faster build times.
2. Made the build and host relationship more clear in the derivation
3. Use filesets for source filtering
4. Skips tests when applicable due to different situations
    1. There is a test that does not work in the sandbox (I think)
    2. There is a test that fails when the git repo is dirty due to version checking

How to cross-compile:
1. Run `nix-build --arg pkgs '(import <nixpkgs> {}).pkgsCross.mingwW64'`
3. Apply this patch and run `nix build .#windows`
```patch
index fd65686293..8bf3701d01 100644
--- a/flake.nix
+++ b/flake.nix
@@ -49,6 +49,10 @@
       packages = {
         jujutsu = jujutsu;
         default = self.packages.${system}.jujutsu;
+
+        windows = pkgs.pkgsCross.mingwW64.callPackage ./. {
+          gitRev = self.rev or self.dirtyRev or "dirty";
+        };
       };

       checks.jujutsu = self.packages.${system}.jujutsu.overrideAttrs {
```

# Checklist

If applicable:

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,3 +3,7 @@
 # download a raw .exe
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-Ctarget-feature=+crt-static"]
+
+[profile.ci]
+inherits = "dev"
+debug = "line-tables-only"
--- a/default.nix
+++ b/default.nix
@@ -140,14 +140,15 @@ in
 
     postInstall = let
       jj = "$out/bin/jj";
-    in ''
-      ${jj} util install-man-pages man
-      installManPage ./man/man1/*
+    in
+      lib.optionalString (stdenv.buildPlatform.canExecute stdenv.hostPlatform) ''
+        ${jj} util install-man-pages man
+        installManPage ./man/man1/*
 
-      installShellCompletion --cmd jj \
-        --bash <(${jj} util completion bash) \
-        --fish <(${jj} util completion fish) \
-        --zsh <(${jj} util completion zsh)
+        installShellCompletion --cmd jj \
+          --bash <(${jj} util completion bash) \
+          --fish <(${jj} util completion fish) \
+          --zsh <(${jj} util completion zsh)
       '';
 
     meta = {
--- a/default.nix
+++ b/default.nix
@@ -5,6 +5,7 @@
   lib,
   stdenv,
   rustPlatform,
+  stdenvAdapters,
   #
   # nativeBuildInputs
   #
@@ -52,7 +53,13 @@
     (fs.fileFilter (f: f.name == "README.md") ./.)
   ]);
 in
-  rustPlatform.buildRustPackage (finalAttrs: {
+  rustPlatform.buildRustPackage.override {
+    # Mold can run most places, but can only build ELF files.
+    stdenv =
+      if stdenv.hostPlatform.isElf
+      then stdenvAdapters.useMoldLinker stdenv
+      else stdenv;
+  } (finalAttrs: {
     strictDeps = true;
 
     pname = "jujutsu";
--- a/default.nix
+++ b/default.nix
@@ -81,6 +81,17 @@ in
       "-p"
       "jj-cli"
     ];
+    checkFlags =
+      lib.optionals (lib.inPureEvalMode) [
+        # Doesn't work in the sandbox
+        "--skip"
+        "test_git::test_push_bookmarks_deletion::use_git2_for_remote_calls"
+      ]
+      ++ lib.optionals (lib.strings.hasInfix "dirty" gitRev) [
+        # Test fails if is marked dirty.
+        "--skip"
+        "test_global_opts::test_version"
+      ];
 
     src = fs.toSource {
       root = ./.;
--- a/default.nix
+++ b/default.nix
@@ -73,6 +73,7 @@ in
 
     doCheck = false;
     useNextest = true;
+    checkType = finalAttrs.cargoBuildType;
 
     src = fs.toSource {
       root = ./.;
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,124 @@
+{
+  #
+  # Stdenv
+  #
+  lib,
+  stdenv,
+  rustPlatform,
+  #
+  # nativeBuildInputs
+  #
+  gzip,
+  pkg-config,
+  cmake,
+  installShellFiles,
+  #
+  # BuildInputs
+  #
+  openssl,
+  libgit2,
+  libssh2,
+  darwin,
+  libiconv,
+  #
+  # NativeCheckInputs
+  #
+  gnupg,
+  openssh,
+  git,
+  #
+  # User input
+  #
+  gitRev ? "dirty",
+}: let
+  filterSrc = src: regexes:
+    lib.cleanSourceWith {
+      inherit src;
+      filter = path: type: let
+        relPath = lib.removePrefix (toString src + "/") (toString path);
+      in
+        lib.all (re: builtins.match re relPath == null) regexes;
+    };   
+in
+  rustPlatform.buildRustPackage (finalAttrs: {
+    strictDeps = true;
+
+    pname = "jujutsu";
+    version = "unstable-" + gitRev;
+
+    cargoLock.lockFile = ./Cargo.lock;
+
+    cargoBuildFlags = ["--bin" "jj"];
+    buildFeatures = ["packaging"];
+    buildType = "release";
+
+    doCheck = false;
+    useNextest = true;
+
+    src = filterSrc ./. [
+      ".*\\.nix$"
+      "^.jj/"
+      "^flake\\.lock$"
+      "^target/"
+    ];
+
+    nativeBuildInputs = [
+      gzip
+      pkg-config
+      installShellFiles
+
+      # Use nixpkgs zlib-ng once rust-lang/libz-sys#206 merges
+      cmake
+    ];
+
+    buildInputs = let
+      d = darwin.apple_sdk.frameworks;
+    in
+      [
+        libgit2
+        libssh2
+      ]
+      ++ lib.optionals (!stdenv.hostPlatform.isDarwin) [openssl]
+      ++ lib.optionals stdenv.isDarwin [
+        d.Security
+        d.SystemConfiguration
+        libiconv
+      ];
+
+    nativeCheckInputs = [
+      # For signing test
+      gnupg
+      openssh
+
+      # For git subprocess test
+      git
+    ];
+
+    env = {
+      NIX_JJ_GIT_HASH = gitRev;
+      RUST_BACKTRACE = 1;
+      # Use nixpkgs libs rather than the vendored libs
+      LIBGIT2_NO_VENDOR = 1;
+      LIBSSH2_SYS_USE_PKG_CONFIG = 1;
+    };
+
+    postInstall = let
+      jj = "$out/bin/jj";
+    in
+      lib.optionalString ''
+        ${jj} util install-man-pages man
+        installManPage ./man/man1/*
+
+        installShellCompletion --cmd jj \
+          --bash <(${jj} util completion bash) \
+          --fish <(${jj} util completion fish) \
+          --zsh <(${jj} util completion zsh)
+      '';
+
+    meta = {
+      description = "Git-compatible DVCS that is both simple and powerful";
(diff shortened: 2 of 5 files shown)


RossSmyth:
Hm running `nix flake check` is different on my computer compared to in CI.

martinvonz:
I don't speak Nix so I'll leave this for e.g. @thoughtpolice, @emilazy or @necauqua to review.

> * Added a "ci" profile that eliminates most debug info (not all so that tracebacks are useful), leading to faster build times.
> * Made the build and host relationship more clear in the derivation
> * Use filesets for source filtering
> * Skips tests when applicable due to different situations
>   
>   1. There is a test that does not work in the sandbox (I think)
>   2. There is a test that fails when the git repo is dirty due to version checking

This list makes me think that the commit could be split up to make it easier to review. Would that make sense?

necauqua:
First impression this looks turbo opinionated, but I like parts of it

I don't think I'm a big authority on how we do nix, so let's see what others will say

RossSmyth:
It is definitely opinionated, but there is no way getting around that when writing Nix since people will argue about whether you should use flakes or pins or niv and all sorts of things. I can split it up though.

thoughtpolice:
This isn't easily reviewable with all the change, and I would probably make different style choices for one. (I absolutely loathe all the Nix formatting tools above all else, and I could in theory get over it, but more on that in a second.)

Regarding the build time optimization with the cargo config, it's something I care about, but it's not relevant to this PR so please drop that. See https://github.com/jj-vcs/jj/pull/3447 for a related discussion around turning this on globally and in a more widely available way so that everyone can benefit (especially during tight dev loops on the testsuite). Even then, the current primary usage of the default Flake expression is so that Nix users can run `nix flake install github:martinvonz/jj` or whatever every once in a while to get a fresh version of trunk; optimizing that occasional build doesn't matter so much, I think. The `flake check` step is much more sensitive because it's what actually gets run during CI on pull requests so that is important, though.

I am also suspicious of the mark for `test_git::test_push_bookmarks_deletion::use_git2_for_remote_calls` to be skipped. What is the actual failure? We should fix that instead because I don't see why this would be a problem; we just push/pull between local repositories for cases like this so I don't see why the sandbox would fail unless it was some kind of path restriction (maybe touching something outside `/build`?) I also don't even know if skipping the version test is correct either; frankly we just expect everyone uses `jj` to develop so by the time you're running the whole testsuite on your change, you probably have a commit with a clean git index. Well, it's possible but I've never hit it, but there are some other edge cases with this test in particular like merge commits so it's not the first time. There might not be a cleaner solution OTTOMH, though.

The bigger problem though is that I'm quite reluctant to accept these kinds of "rewrite whole file in a way like this" commits like this. One issue that you note (correctly) is that people have all sorts of opinions about how to write Nix. But the real issue would apply to any file in the source tree: in practice, I must assume 99.9% of all people who write patches will not ever return and write any further ones, and so I will continue to maintain this forever, but without knowing all the peculiarities of why things were done. The reason I must assume this is just because, well, I've been doing this for a long time and I've noticed it. :')

From what I can tell, the intent of this PR at a high level is more like:

- Fix some host/build distinctions, so cross compiling can work in a top-level Flake expression.
  - I don't know if we want to deal with supporting Windows cross compilation as an actual CI target in our matrix, especially considering it could be handled by an override on a downstream flake somewhere else, or your own CI system, etc.
  - But I'm also not inherently opposed to adding it either (I can definitely think of some reasons why people would want it).
- Make it possible to run `nix-shell` as an alternative to the current suggested workflow, which is to use direnv, by splitting out a new `default.nix` expression from the main flake.
  - Personally I don't care about this because `direnv`'s UX is far better IMO, but have thought about doing it anyway, and always put it off because nobody ever asked for it. So I guess I'm OK with this.
- Things like filesets make some local patches easier to write because it won't cause spurious rebuilds of the Nix derivation. These are basically just QoL changes that make some things nicer.
  - I probably don't notice them because I use my shell for all my iterative needs and don't try things like `nix build` until much later.

Is that approximately correct? I'm happy to look at each of these changes in isolation and even write some of them myself. But right now I feel like I'd probably just end up reworking your whole PR to be more "Austin

thoughtpolice:
> I'm happy to look at each of these changes in isolation and even write some of them myself. But right now I feel like I'd probably just end up reworking this PR a bit more to my liking ("Austin-amenable") which probably wouldn't happen instantly as I'm quite busy.

Actually, on second thought, I think I'd probably just accept this as-is since rewriting the entire thing is annoying (I can undo anything I dislike later assuming a positive review, of course.) But I would at least like things to be broken up into multiple commits as someone else noted; the 3 bullet points above are probably a good enough starting point on how to organize it. But is that more or less work than redoing it? I'll leave that for you to decide, I suppose.

emilazy:
I agree that this is kinda hard to review in the current state. I’d rather avoid filesets if possible because they’re buggy with current Nix semantics. I approve of making cross work better. I’d rather not encourage `import <nixpkgs> {}` which is likely going to break as we change the versions of libraries we require. If we split out the package into a separate `package.nix` we should use a proper `callPackage` rather than a bunch of `? pkgs.…` defaults; that would also be better for cross. `lib.inPureEvalMode` shouldn’t be used as a proxy for the build sandbox, as those are unrelated things (and I think we couldn’t reproduce that failure you reported to Nixpkgs)?

RossSmyth:
Here it is split up.

RossSmyth:
> (and I think we couldn’t reproduce that failure you reported to Nixpkgs)?

I can reproduce it here upstream though

RossSmyth:
And personally I don't really care about the nix-build usecase, I mainly had it so it would be easier to demonstrate the crosspkgs capability.

RossSmyth:
re the Cargo profile, your PR makes it a default. This only effects what is tested in CI.

RossSmyth:
@emilazy In the commits that I have pushed, if you checkout `6e6fc369` and run `nix flake check` you should get
```
jujutsu-unstable>         FAIL [   0.110s] jj-lib::runner test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls
jujutsu-unstable> ---- STDOUT:             jj-lib::runner test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls
jujutsu-unstable> running 1 test
jujutsu-unstable> test test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls ... FAILED
jujutsu-unstable> failures:
jujutsu-unstable> failures:
jujutsu-unstable>     test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls
jujutsu-unstable> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 439 filtered out; finished in 0.09s
jujutsu-unstable>
jujutsu-unstable> ---- STDERR:             jj-lib::runner test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls
jujutsu-unstable> thread 'test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls' panicked at lib/tests/test_git.rs:3376:5:
jujutsu-unstable> assertion failed: !tx.repo().has_changes()
jujutsu-unstable> stack backtrace:
jujutsu-unstable>    0: rust_begin_unwind
jujutsu-unstable>              at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/std/src/panicking.rs:695:5
jujutsu-unstable>    1: core::panicking::panic_fmt
jujutsu-unstable>              at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/core/src/panicking.rs:75:14
jujutsu-unstable>    2: core::panicking::panic
jujutsu-unstable>              at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/core/src/panicking.rs:145:5
jujutsu-unstable>    3: test_push_bookmarks_mixed_deletion_and_addition
jujutsu-unstable>              at ./tests/test_git.rs:3376:5
jujutsu-unstable>    4: runner::test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls
jujutsu-unstable>              at ./tests/test_git.rs:3307:1
jujutsu-unstable>    5: runner::test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls::{{closure}}
jujutsu-unstable>              at ./tests/test_git.rs:3307:49
jujutsu-unstable>    6: core::ops::function::FnOnce::call_once
jujutsu-unstable>              at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/core/src/ops/function.rs:250:5
jujutsu-unstable>    7: core::ops::function::FnOnce::call_once
jujutsu-unstable>              at /rustc/f9e0239a7bc813b4aceffc7f069f4797cde3175c/library/core/src/ops/function.rs:250:5
jujutsu-unstable> note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
jujutsu-unstable>   Cancelling due to test failure: 18 tests still running, 1 failed, 0 skipped
jujutsu-unstable>         PASS [   0.080s] jj-lib::runner test_git::test_push_updates_unexpectedly_moved_sideways_on_remote::spawn_a_git_subprocess_for_remote_calls
jujutsu-unstable>         PASS [   0.018s] jj-lib::runner test_index::test_index_commits_incremental_empty_transaction
jujutsu-unstable>         PASS [   0.049s] jj-lib::runner test_id_prefix::test_id_prefix_divergentskipped
jujutsu-unstable>         PASS [   0.015s] jj-lib::runner test_index::test_index_store_typem0 skipped
jujutsu-unstable>         PASS [   0.018s] jj-lib::runner test_index::test_index_commits_previous_operationspped
jujutsu-unstable>         PASS [   0.048s] jj-lib::runner test_id_prefix::test_id_prefix_hidden0m skipped
jujutsu-unstable>         PASS [   0.019s] jj-lib::runner test_index::test_index_commits_standard_cases1mskipped
jujutsu-unstable>         PASS [   1.461s] jj-cli::runner test_workspaces::test_workspaces_sparse_patternskipped
jujutsu-unstable>         PASS [   0.017s] jj-lib::runner test_index::test_reindex_corrupt_segment_filesmskipped
jujutsu-unstable>         PASS [   0.057s] jj-lib::runner test_id_prefix::test_id_prefix[1m0 skipped
jujutsu-unstable>         PASS [   0.022s] jj-lib::runner test_index::test_reindex_from_merged_operationskipped
jujutsu-unstable>         PASS [   0.050s] jj-lib::runner test_index::test_index_commits_criss_cross3;1mskipped
jujutsu-unstable>         PASS [   0.074s] jj-lib::runner test_gpg::invalid_signature, 0 skipped
jujutsu-unstable>         PASS [   0.086s] jj-lib::runner test_gpg::unknown_key1;1mfailed, 0 skipped
jujutsu-unstable>         PASS [   0.092s] jj-lib::runner test_gpg::gpg_signing_roundtrip1m0 skipped
jujutsu-unstable>         PASS [   0.094s] jj-lib::runner test_gpg::gpg_signing_roundtrip_explicit_key1mskipped
jujutsu-unstable>         PASS [   0.087s] jj-lib::runner test_index::test_index_commits_incremental_squasheded
jujutsu-unstable>         PASS [   0.141s] jj-lib::runner test_git_backend::test_gcfailed, 0 skipped
jujutsu-unstable> ------------[0m [ 00:00:40]  1512/1783: 0 running, 1511 passed, 1 failed, 0 skipped
jujutsu-unstable>      Summary [  40.208s] 1512/1783 tests run: 1511 passed, 1 failed, 0 skipped
jujutsu-unstable>         FAIL [   0.110s] jj-lib::runner test_git::test_push_bookmarks_mixed_deletion_and_addition::use_git2_for_remote_calls
jujutsu-unstable> warning: 271/1783 tests were not run due to test failure (run with --no-fail-fast to run all tests, or run with --max-fail)
jujutsu-unstable> error: test run failed/1783: 0 running, 1511 passed, 1 failed, 0 skipped
```

RossSmyth:
I guess looking more closely it may be related to gpg instead? Similar to how the gpg tests on Darwin are failing in CI

PR #5971, opened 2025-03-10
cli: config: support multiple user configuration files

calebdw:
~~The `platform_config_path` now defaults to the `jj` directory as opposed to the `config.toml` file. This loads all configuration files in the directory by default, removing the need to set JJ_CONFIG for a multi-file approach.~~

Multiple user configs are now supported and are loaded in the following precedence order:
  - `$HOME/.jjconfig.toml`
  - `$XDG_CONFIG_HOME/jj/config.toml`
  - `$XDG_CONFIG_HOME/jj/conf.d/*.toml`

which removes the need to set `JJ_CONFIG` for a multi-file approach.
Later files override earlier files and the `JJ_CONFIG` environment
variable can be used to override the default paths.

The `JJ_CONFIG` environment variable can now contain multiple paths separated
by a colon (or semicolon on Windows).

Thanks!

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/config/path.rs
+++ b/cli/src/commands/config/path.rs
@@ -22,11 +22,11 @@ use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
-/// Print the path to the config file
+/// Print the paths to the config files
 ///
 /// A config file at that path may or may not exist.
 ///
-/// See `jj config edit` if you'd like to immediately edit the file.
+/// See `jj config edit` if you'd like to immediately edit a file.
 #[derive(clap::Args, Clone, Debug)]
 pub struct ConfigPathArgs {
     #[command(flatten)]
@@ -39,13 +39,14 @@ pub fn cmd_config_path(
     command: &CommandHelper,
     args: &ConfigPathArgs,
 ) -> Result<(), CommandError> {
-    let config_path = args.level.config_path(command.config_env())?;
-    writeln!(
-        ui.stdout(),
-        "{}",
-        config_path
-            .to_str()
-            .ok_or_else(|| user_error("The config path is not valid UTF-8"))?
-    )?;
+    for config_path in args.level.config_paths(command.config_env())? {
+        writeln!(
+            ui.stdout(),
+            "{}",
+            config_path
+                .to_str()
+                .ok_or_else(|| user_error("The config path is not valid UTF-8"))?
+        )?;
+    }
     Ok(())
 }
--- a/cli/src/commands/config/set.rs
+++ b/cli/src/commands/config/set.rs
@@ -30,7 +30,7 @@ use crate::complete;
 use crate::config::parse_value_or_bare_string;
 use crate::ui::Ui;
 
-/// Update config file to set the given option to a given value.
+/// Update a config file to set the given option to a given value.
 #[derive(clap::Args, Clone, Debug)]
 pub struct ConfigSetArgs {
     #[arg(required = true, add = ArgValueCandidates::new(complete::leaf_config_keys))]
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -16,6 +16,7 @@ use std::borrow::Cow;
 use std::collections::BTreeSet;
 use std::collections::HashMap;
 use std::env;
+use std::env::split_paths;
 use std::fmt;
 use std::path::Path;
 use std::path::PathBuf;
@@ -34,7 +35,6 @@ use jj_lib::config::ConfigValue;
 use jj_lib::config::StackedConfig;
 use regex::Captures;
 use regex::Regex;
-use thiserror::Error;
 use tracing::instrument;
 
 use crate::command_error::config_error;
@@ -79,12 +79,6 @@ const fn is_bare_char(b: u8) -> bool {
         // non-ASCII
         b'\x80'..=b'\xff' => true,
     }
-}
-
-#[derive(Error, Debug)]
-pub enum ConfigEnvError {
-    #[error("Both {0} and {1} exist. Please consolidate your configs in one of them.")]
-    AmbiguousSource(PathBuf, PathBuf),
 }
 
 /// Configuration variable with its source information.
@@ -228,31 +222,43 @@ struct UnresolvedConfigEnv {
 }
 
 impl UnresolvedConfigEnv {
-    fn resolve(self) -> Result<Option<ConfigPath>, ConfigEnvError> {
-        if let Some(path) = self.jj_config {
-            // TODO: We should probably support colon-separated (std::env::split_paths)
-            return Ok(Some(ConfigPath::new(PathBuf::from(path))));
+    fn resolve(self) -> Vec<ConfigPath> {
+        if let Some(paths) = self.jj_config {
+            return split_paths(&paths)
+                .filter(|path| !path.as_os_str().is_empty())
+                .map(ConfigPath::new)
+                .collect();
         }
-        // TODO: Should we drop the final `/config.toml` and read all files in the
-        // directory?
-        let platform_config_path = self.config_dir.map(|mut config_dir| {
-            config_dir.push("jj");
-            config_dir.push("config.toml");
-            ConfigPath::new(config_dir)
-        });
+        let mut paths = vec![];
         let home_config_path = self.home_dir.map(|mut home_dir| {
             home_dir.push(".jjconfig.toml");
             ConfigPath::new(home_dir)
         });
+        let platform_config_path = self.config_dir.clone().map(|mut config_dir| {
+            config_dir.push("jj");
+            config_dir.push("config.toml");
+            ConfigPath::new(config_dir)
+        });
+        let platform_config_dir = self.config_dir.map(|mut config_dir| {
+            config_dir.push("jj");
+            config_dir.push("conf.d");
+            ConfigPath::new(config_dir)
+        });
         use ConfigPath::*;
-        match (platform_config_path, home_config_path) {
-            (Some(Existing(platform_config_path)), Some(Existing(home_config_path))) => Err(
-                ConfigEnvError::AmbiguousSource(platform_config_path, home_config_path),
-            ),
-            (Some(Existing(path)), _) | (_, Some(Existing(path))) => Ok(Some(Existing(path))),
-            (Some(New(path)), _) | (_, Some(New(path))) => Ok(Some(New(path))),
-            (None, None) => Ok(None),
+        if let Some(path @ Existing(_)) = home_config_path {
+            paths.push(path);
+        } else if let (Some(path @ New(_)), None) = (home_config_path, &platform_config_path) {
+            paths.push(path);
         }
+        // This should be the default config created if there's
+        // no user config and `jj config edit` is executed.
+        if let Some(path) = platform_config_path {
+            paths.push(path);
+        }
+        if let Some(path @ Existing(_)) = platform_config_dir {
+            paths.push(path);
+        }
+        paths
     }
 }
 
@@ -260,14 +266,14 @@ impl UnresolvedConfigEnv {
 pub struct ConfigEnv {
     home_dir: Option<PathBuf>,
     repo_path: Option<PathBuf>,
-    user_config_path: Option<ConfigPath>,
+    user_config_paths: Vec<ConfigPath>,
     repo_config_path: Option<ConfigPath>,
     command: Option<String>,
 }
 
 impl ConfigEnv {
     /// Initializes configuration loader based on environment variables.
-    pub fn from_environment() -> Result<Self, ConfigEnvError> {
+    pub fn from_environment() -> Self {
         // Canonicalize home as we do canonicalize cwd in CliRunner. $HOME might
         // point to symlink.
         let home_dir = dirs::home_dir().map(|path| dunce::canonicalize(&path).unwrap_or(path));
@@ -276,30 +282,31 @@ impl ConfigEnv {
             home_dir: home_dir.clone(),
             jj_config: env::var("JJ_CONFIG").ok(),
         };
-        Ok(ConfigEnv {
+        ConfigEnv {
             home_dir,
             repo_path: None,
-            user_config_path: env.resolve()?,
+            user_config_paths: env.resolve(),
             repo_config_path: None,
             command: None,
-        })
+        }
     }
 
     pub fn set_command_name(&mut self, command: String) {
         self.command = Some(command);
     }
 
-    /// Returns a path to the user-specific config file or directory.
-    pub fn user_config_path(&self) -> Option<&Path> {
-        self.user_config_path.as_ref().map(ConfigPath::as_path)
+    /// Returns the paths to the user-specific config files or directories.
+    pub fn user_config_paths(&self) -> impl Iterator<Item = &Path> {
+        self.user_config_paths.iter().map(|p| p.as_path())
     }
 
-    /// Returns a path to the existing user-specific config file or directory.
-    fn existing_user_config_path(&self) -> Option<&Path> {
-        match &self.user_config_path {
-            Some(ConfigPath::Existing(path)) => Some(path),
+    /// Returns the paths to the existing user-specific config files or
+    /// directories.
+    pub fn existing_user_config_paths(&self) -> impl Iterator<Item = &Path> {
+        self.user_config_paths.iter().filter_map(|p| match p {
+            ConfigPath::Existing(path) => Some(path.as_path()),
             _ => None,
-        }
+        })
     }
 
     /// Returns user configuration files for modification. Instantiates one if
@@ -316,7 +323,8 @@ impl ConfigEnv {
     }
 
     fn new_user_config_file(&self) -> Result<Option<ConfigFile>, ConfigLoadError> {
-        self.user_config_path()
+        self.user_config_paths()
+            .next()
             .map(|path| {
                 // No need to propagate io::Error here. If the directory
                 // couldn't be created, file.save() would fail later.
@@ -335,7 +343,7 @@ impl ConfigEnv {
     #[instrument]
     pub fn reload_user_config(&self, config: &mut RawConfig) -> Result<(), ConfigLoadError> {
         config.as_mut().remove_layers(ConfigSource::User);
-        if let Some(path) = self.existing_user_config_path() {
+        for path in self.existing_user_config_paths() {
             if path.is_dir() {
                 config.as_mut().load_dir(ConfigSource::User, path)?;
             } else {
@@ -432,7 +440,7 @@ fn config_files_for(
 /// Sources from the lowest precedence:
 /// 1. Default
 /// 2. Base environment variables
-/// 3. [User config](https://jj-vcs.github.io/jj/latest/config/)
+/// 3. [User configs](https://jj-vcs.github.io/jj/latest/config/)
 /// 4. Repo config `.jj/repo/config.toml`
 /// 5. TODO: Workspace config `.jj/config.toml`
 /// 6. Override environment variables
@@ -766,10 +774,10 @@ impl TryFrom<Vec<String>> for NonEmptyCommandArgsVec {
 
 #[cfg(test)]
 mod tests {
+    use std::env::join_paths;
     use std::fmt::Write as _;
 
     use anyhow::anyhow;
-    use assert_matches::assert_matches;
     use indoc::indoc;
     use maplit::hashmap;
 
@@ -1219,67 +1227,85 @@ mod tests {
     }
 
     #[test]
-    fn test_config_path_home_dir_existing() -> anyhow::Result<()> {
+    fn test_config_path_home_existing() -> anyhow::Result<()> {
         TestCase {
-            files: vec!["home/.jjconfig.toml"],
+            files: ["home/.jjconfig.toml"],
             env: UnresolvedConfigEnv {
                 home_dir: Some("home".into()),
                 ..Default::default()
             },
-            want: Want::ExistingAndNew("home/.jjconfig.toml"),
+            wants: [Want::Existing("home/.jjconfig.toml")],
         }
         .run()
     }
 
     #[test]
-    fn test_config_path_home_dir_new() -> anyhow::Result<()> {
+    fn test_config_path_home_new() -> anyhow::Result<()> {
         TestCase {
-            files: vec![],
+            files: [],
             env: UnresolvedConfigEnv {
                 home_dir: Some("home".into()),
                 ..Default::default()
             },
-            want: Want::New("home/.jjconfig.toml"),
+            wants: [Want::New("home/.jjconfig.toml")],
         }
         .run()
     }
 
     #[test]
-    fn test_config_path_config_dir_existing() -> anyhow::Result<()> {
+    fn test_config_path_home_existing_platform_new() -> anyhow::Result<()> {
         TestCase {
-            files: vec!["config/jj/config.toml"],
+            files: ["home/.jjconfig.toml"],
             env: UnresolvedConfigEnv {
-                config_dir: Some("config".into()),
-                ..Default::default()
-            },
-            want: Want::ExistingAndNew("config/jj/config.toml"),
-        }
-        .run()
-    }
-
-    #[test]
-    fn test_config_path_config_dir_new() -> anyhow::Result<()> {
-        TestCase {
-            files: vec![],
-            env: UnresolvedConfigEnv {
-                config_dir: Some("config".into()),
-                ..Default::default()
-            },
-            want: Want::New("config/jj/config.toml"),
-        }
-        .run()
-    }
-
-    #[test]
-    fn test_config_path_new_prefer_config_dir() -> anyhow::Result<()> {
-        TestCase {
-            files: vec![],
-            env: UnresolvedConfigEnv {
-                config_dir: Some("config".into()),
                 home_dir: Some("home".into()),
+                config_dir: Some("config".into()),
                 ..Default::default()
             },
-            want: Want::New("config/jj/config.toml"),
+            wants: [
+                Want::Existing("home/.jjconfig.toml"),
+                Want::New("config/jj/config.toml"),
+            ],
+        }
+        .run()
+    }
+
+    #[test]
+    fn test_config_path_platform_existing() -> anyhow::Result<()> {
+        TestCase {
+            files: ["config/jj/config.toml"],
+            env: UnresolvedConfigEnv {
+                home_dir: Some("home".into()),
+                config_dir: Some("config".into()),
+                ..Default::default()
+            },
+            wants: [Want::Existing("config/jj/config.toml")],
+        }
+        .run()
+    }
+
+    #[test]
+    fn test_config_path_platform_new() -> anyhow::Result<()> {
+        TestCase {
+            files: [],
+            env: UnresolvedConfigEnv {
+                config_dir: Some("config".into()),
+                ..Default::default()
+            },
+            wants: [Want::New("config/jj/config.toml")],
+        }
+        .run()
+    }
+
+    #[test]
+    fn test_config_path_new_prefer_platform() -> anyhow::Result<()> {
+        TestCase {
+            files: [],
+            env: UnresolvedConfigEnv {
+                home_dir: Some("home".into()),
+                config_dir: Some("config".into()),
+                ..Default::default()
+            },
+            wants: [Want::New("config/jj/config.toml")],
         }
         .run()
     }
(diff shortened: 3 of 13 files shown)


cli/src/config.rs
```diff
@@ -234,11 +234,8 @@ impl UnresolvedConfigEnv {
             // TODO: We should probably support colon-separated (std::env::split_paths)
             return Ok(ConfigPath::new(Some(PathBuf::from(path))));
         }
-        // TODO: Should we drop the final `/config.toml` and read all files in the
-        // directory?
         let platform_config_path = ConfigPath::new(self.config_dir.map(|mut config_dir| {
             config_dir.push("jj");
-            config_dir.push("config.toml");
```
yuja, line 241: I tried something like this before, and found that the problem isn't that simple.

1. the existence of an empty `~/.config/jj` directory shouldn't block loading of `~/.jjconfig.toml`
2. `jj config edit` should create new file at `~/.config/jj/config.toml`.

(2) is easy to fix. (1) is unclear. User might put non-config files, so non-empty directory doesn't always mean the directory has at least one config file. It's also unclear if we'll never add `*.toml` file that isn't a config file. 

It might be better to add `~/.config/jj/config` (or `~/Library/../jj/config` on macOS) directory, but then, we'll have to disambiguate `~/.config/jj/config` directory and `~/.config/jj/config.toml` file.

It might be even simpler to remove the ambiguous source error and load config files from both `~/.jjconfig.toml` and `~/.config/jj/*.toml`...

calebdw: You raise some good points:

> (2) is easy to fix. (1) is unclear. User might put non-config files, so non-empty directory doesn't always mean the directory has at least one config file. It's also unclear if we'll never add *.toml file that isn't a config file.
>
> It might be better to add ~/.config/jj/config (or ~/Library/../jj/config on macOS) directory, but then, we'll have to disambiguate ~/.config/jj/config directory and ~/.config/jj/config.toml file.

I would recommend going with a `conf.d` directory as that is pretty standard.

> It might be even simpler to remove the ambiguous source error and load config files from both ~/.jjconfig.toml and ~/.config/jj/*.toml...

I also agree here. It is pretty standard to just load from all locations and have an order of load precedence.

I recommend the following order:
1. `$HOME/.jjconfig.toml`
2. `$XDG_CONFIG_HOME/jj/config.toml`
3. `$XDG_CONFIG_HOME/jj/conf.d/*.toml`

Configuration files loaded later have precedence and replace any setting defined in previous configs. The `JJ_CONFIG` env variable still acts as an override and can point to either a file or a directory (in which case all `*.toml` files in that directory are loaded), this can even be a comma or semicolon separated list to have multiple paths. 

Finally, the following would still work as intended:
> jj config edit should create new file at ~/.config/jj/config.toml.

martinvonz: That sounds good to me. i'm not sure what `jj config edit` should do there are multiple config files. Maybe edit the highest-precedence file? But that's probably not what you want if you have files like these:

```
$XDG_CONFIG_HOME/jj/config.toml
$XDG_CONFIG_HOME/jj/conf.d/personal.toml
$XDG_CONFIG_HOME/jj/conf.d/work.toml
```

It seems more likely that you'll want to edit the first one than the last one in that case. But always editing the lowest-precedence file also doesn't seem ideal because if you have both `$HOME/.jjconfig.toml` and `$XDG_CONFIG_HOME/jj/config.toml`, then your changes to the former might be overridden by the latter. So I'm not sure what's best.

cli/src/commands/config/set.rs
```diff
@@ -30,7 +30,7 @@ use crate::complete;
 use crate::config::parse_value_or_bare_string;
 use crate::ui::Ui;
 
-/// Update config file to set the given option to a given value.
+/// Update a config file to set the given option to a given value.
```
yuja, line 33: nit: I don't think it's uncommon to omit `a`/`the` in the summary line, but I'm not native.

cli/src/config.rs
```diff
@@ -82,10 +83,7 @@ const fn is_bare_char(b: u8) -> bool {
 }
 
 #[derive(Error, Debug)]
-pub enum ConfigEnvError {
-    #[error("Both {0} and {1} exist. Please consolidate your configs in one of them.")]
-    AmbiguousSource(PathBuf, PathBuf),
-}
+pub enum ConfigEnvError {}
```
yuja, line 86: nit: Can you also remove the error type at all? (or leave a TODO comment?)

cli/src/config.rs
```diff
@@ -229,39 +227,54 @@ struct UnresolvedConfigEnv {
...
+        if let Existing(_) | New(_) = platform_config_path {
+            paths.push(platform_config_path);
+        }
+        if let Existing(_) = platform_config_dir {
+            paths.push(platform_config_dir);
+        }
+        if paths.is_empty() {
+            paths.push(Unavailable);
+        }
+        Ok(paths)
```
yuja, line 269: An empty vec should mean there are no config paths available, which overlaps with `ConfigPath::Unavailable`.

I think I have unsent patch that replaces `ConfigPath::Unavailable` with `None`. It might be better to upstream this patch first.

cli/src/config.rs
```diff
@@ -229,39 +227,54 @@ struct UnresolvedConfigEnv {
...
-            (New(path), _) | (_, New(path)) => Ok(New(path)),
-            (Unavailable, Unavailable) => Ok(Unavailable),
+        if matches!(home_config_path, Existing(_))
+            || (matches!(home_config_path, New(_)) && matches!(platform_config_path, Unavailable))
+        {
+            paths.push(home_config_path);
         }
+        // This should be the default config created if there's
+        // no user config and `jj config edit` is executed.
+        if let Existing(_) | New(_) = platform_config_path {
```
yuja, line 260: nit: use `matches!()` consistently

yuja (commented):
I haven't reviewed thoroughly, but the code generally looks good to me.

cli/src/commands/config/set.rs
calebdw: Yes for `the`, but I was trying to differentiate between "a file" (of potentially many) vs. "the file"

cli/src/config.rs
calebdw: I left the type here so it's easier to add errors in the future if needed

yuja: Just delete it unless you have a plan. Reintroducing the error type should be easy.

cli/src/config.rs
calebdw: Yeah, I had the same thought but decided not to remove Unavailable at the time.

However, I've gone ahead and removed it in this PR (so #6012) should no longer be needed---given the number of changes in this file I figured it was easier to implement it here than try to rebase and deal with the conflicts.

cli/src/commands/config/set.rs
calebdw: I can revert if you like?

cli/src/config.rs
yuja: We prefer small incremental changes, so it would be nice if you can rebase this on top (or split out a similar change.) I don't think this patch is big, though.
I think you can just rebase and restore the contents.

cli/src/commands/config/set.rs
yuja: I don't have strong opinion, and I'm not the right person to review documentation changes ;)

cli/src/config.rs
calebdw: So just to confirm, you want me to rebase my commit onto #6012?

cli/src/commands/config/path.rs
yuja, line 1: nit: can you add a test for `jj config path` with multiple config paths?

cli/src/config.rs
```diff
@@ -228,46 +222,58 @@ struct UnresolvedConfigEnv {
...
+        } else if let (Some(New(path)), None) = (home_config_path, &platform_config_path) {
+            paths.push(New(path));
+        }
+        // This should be the default config created if there's
+        // no user config and `jj config edit` is executed.
+        if let Some(path) = platform_config_path {
+            paths.push(path);
         }
+        if let Some(Existing(_)) = platform_config_dir {
+            paths.push(platform_config_dir.unwrap());
```
yuja, line 259: nit: `if let Some(path @ Existing(_))` instead of `.unwrap()`.

cli/src/config.rs
```diff
@@ -276,30 +282,37 @@ impl ConfigEnv {
...
 
     pub fn set_command_name(&mut self, command: String) {
         self.command = Some(command);
     }
 
-    /// Returns a path to the user-specific config file or directory.
-    pub fn user_config_path(&self) -> Option<&Path> {
-        self.user_config_path.as_ref().map(ConfigPath::as_path)
+    /// Returns the paths to the user-specific config files or directories.
+    pub fn user_config_paths(&self) -> Vec<&Path> {
```
yuja, line 299: nit: maybe better to return `impl Iterator<Item = &Path>` because this function is mainly used internally.

cli/src/config.rs
```diff
@@ -276,30 +282,37 @@ impl ConfigEnv {
...
 
-    /// Returns a path to the existing user-specific config file or directory.
-    fn existing_user_config_path(&self) -> Option<&Path> {
-        match &self.user_config_path {
-            Some(ConfigPath::Existing(path)) => Some(path),
-            _ => None,
-        }
+    /// Returns the paths to the existing user-specific config files or
+    /// directories.
+    pub fn existing_user_config_paths(&self) -> Vec<&Path> {
```
yuja, line 308: nit: same here. `impl Iterator<Item = &Path>`

cli/src/config.rs
```diff
@@ -335,13 +350,15 @@ impl ConfigEnv {
...
-        }
+        self.existing_user_config_paths()
+            .iter()
+            .try_for_each(|path| {
+                if path.is_dir() {
+                    config.as_mut().load_dir(ConfigSource::User, path)
+                } else {
+                    config.as_mut().load_file(ConfigSource::User, path)
+                }
+            })?;
```
yuja, line 361: nit: `?; Ok(())` can be omitted. It's also good to use plain `for` loop.

cli/src/config.rs
```diff
@@ -354,7 +371,7 @@ impl ConfigEnv {
 
     /// Returns a path to the repo-specific config file.
     pub fn repo_config_path(&self) -> Option<&Path> {
-        self.repo_config_path.as_ref().map(ConfigPath::as_path)
+        self.repo_config_path.as_ref().map(|p| p.as_path())
```
yuja, line 374: nit: unneeded change

cli/src/config.rs
```diff
@@ -1219,80 +1240,98 @@ mod tests {
...
         TestCase {
-            files: vec!["custom.toml"],
+            files: vec!["custom1.toml"],
             env: UnresolvedConfigEnv {
-                jj_config: Some("custom.toml".into()),
+                jj_config: Some("custom1.toml".into()),
                 ..Default::default()
             },
-            want: Want::ExistingAndNew("custom.toml"),
+            wants: vec![Want::Existing("custom1.toml")],
```
yuja, line 1334: nit: unneeded change (maybe you forgot to revert?)

cli/src/config.rs
```diff
@@ -1305,59 +1344,149 @@ mod tests {
...
+        TestCase {
+            files: vec!["custom1.toml"],
+            env: UnresolvedConfigEnv {
+                jj_config: Some(["custom1.toml", "", "custom2.toml"].join(PATH_SEPARATOR)),
+                ..Default::default()
+            },
+            wants: vec![Want::Existing("custom1.toml"), Want::New("custom2.toml")],
+        }
+        .run()
+    }
```
yuja, line 1392: nit: can you also add test for `jj_config: Some("".to_owned())`? It should effectively disable config loading.

cli/src/config.rs
```diff
@@ -1373,36 +1502,37 @@ mod tests {
     }
 
     enum Want {
-        None,
         New(&'static str),
-        ExistingAndNew(&'static str),
+        Existing(&'static str),
     }
```
yuja, line 1507: nit: Since we longer have `Existing|New` pattern, it's probably better to add `wants_new: Vec<_>` and `wants_existing: Vec<_>`.

cli/src/config.rs
```diff
@@ -1373,36 +1502,37 @@ mod tests {
...
                 home_dir: home_dir.clone(),
-                jj_config: self
-                    .env
-                    .jj_config
-                    .as_ref()
-                    .map(|p| root.join(p).to_str().unwrap().to_string()),
+                jj_config: self.env.jj_config.as_ref().map(|p| {
+                    std::env::split_paths(p)
+                        .filter(|path| !path.as_os_str().is_empty())
+                        .map(|p| root.join(p).to_str().unwrap().to_string())
```
yuja, line 1524: Empty paths shouldn't be filtered out here because we need to test the actual implementation.

cli/src/config.rs
```diff
@@ -1413,35 +1543,52 @@ mod tests {
...
+            let existing: Vec<PathBuf> = env
+                .existing_user_config_paths()
+                .iter()
+                .map(|p| p.to_path_buf())
+                .collect();
+            if existing != expected {
+                return Err(anyhow!(
+                    "existing_config_path mismatch: got {:?}, expected {:?}",
+                    existing,
+                    expected
```
yuja, line 1564: nit: parameters can be inlined as `{existing:?}`, etc.

cli/src/config.rs
```diff
@@ -1413,35 +1543,52 @@ mod tests {
...
+                .user_config_paths()
+                .iter()
+                .map(|p| p.to_path_buf())
+                .filter(|p| !p.try_exists().unwrap_or(false))
+                .collect();
+            if new != expected {
+                return Err(anyhow!(
+                    "new_config_path mismatch: got {:?}, expected {:?}",
+                    new,
+                    expected
```
yuja, line 1590: nit: same here. `{new:?}`, `{expected:?}`

docs/config.md
```diff
@@ -1455,41 +1455,49 @@ Setting this value to zero will disable the limit entirely.
...
-The rest of this section covers the details of where this file can be located.
+On all platforms, the user's global `jj` configurations are loaded in the following
+precedence order (with later configs overriding earlier ones):
 
-On all platforms, the user's global `jj` configuration file is located at either
-`~/.jjconfig.toml` (where `~` represents `$HOME` on Unix-likes, or
-`%USERPROFILE%` on Windows) or in a platform-specific directory. The
-platform-specific location is recommended for better integration with platform
-services. It is an error for both of these files to exist.
+- `$HOME/.jjconfig.toml` (not recommended)
```
yuja, line 1469: nit: remove `(not recommended)`? It sounds a bit stronger than just saying "platform location is recommended for `<reasons>`."

docs/config.md
```diff
@@ -1455,41 +1455,49 @@ Setting this value to zero will disable the limit entirely.
...
+- `$XDG_CONFIG_HOME/jj/config.toml`
+- `$XDG_CONFIG_HOME/jj/conf.d/*.toml`
+
+where `$HOME` represents the user's home directory (`%USERPROFILE%` on Windows),
+and `$XDG_CONFIG_HOME` represents the platform-specific configuration directory.
+The platform-specific location is recommended for better integration with platform
+services.
+
+The `conf.d` directory loads additional files in lexicographic order. This allows
```
yuja, line 1478: nit: maybe you mean "Files in the `conf.d` directory are loaded in lexicographical order."?

cli/src/config.rs
yuja: Oh, but it doesn't make sense to assert filtering in our test code: `.filter(|p| !p.try_exists().unwrap_or(false))`. Maybe we should remove it from `check_new()`.

docs/config.md
calebdw: I wanted to draw attention to it in the list which is what most people will look at---I imagine that some might skip over the recommendation in the text.

Would you be opposed to adding `(preferred)` next to the `config.toml` (or changing `not recommended` to `not preferred`)?

cli/src/config.rs
calebdw: I've switched to the `for` loop, but removing the `Ok(())` would require removing the `Result` return type which would cascade elsewhere---is that okay?

cli/src/commands/config/path.rs
calebdw: Note that `jj show path --user` will only show the directory path (e.g., `~/.config/jj/conf.d`) instead of any files inside the directory (e.g., `~/.config/jj/conf.d/work.toml`)....

If it's desirable to show all the files then they would need to be resolved sooner

cli/src/config.rs
yuja: No, don't remove `Result`. I meant either `for` loop or `.try_for_each()` without `Ok(())`.

docs/config.md
yuja: `(preferred)` sounds good to me.

cli/src/commands/config/path.rs
yuja: Yeah, I noticed that too. I'm not sure if `jj config path` is supposed to show the specified paths or actual file paths. For the latter, the list may be empty if no files exist, which might be undesired.

calebdw: > the list may be empty if no files exist,

This is not the case---the list will always be populated with either `~/.config/jj/config.toml` or `~/.jjconfig.toml` (if the config path doesn't exist`), unless of course it's been overridden by `JJ_CONFIG`.

`jj config edit` will create the config if it doesn't exist

cli/src/config.rs
calebdw: > Since we longer have Existing|New pattern,

I'm not sure what you mean by that---we are still using the Existing|New pattern

> Oh, but it doesn't make sense to assert filtering in our test code: .filter(|p| !p.try_exists().unwrap_or(false)).

The method needs to grab all the paths that don't exist...

cli/tests/test_config_command.rs
```diff
@@ -21,6 +21,11 @@ use crate::common::fake_editor_path;
 use crate::common::to_toml_value;
 use crate::common::TestEnvironment;
 
+#[cfg(unix)]
+const PATH_SEPARATOR: &str = ":";
+#[cfg(windows)]
+const PATH_SEPARATOR: &str = ";";
```
martinvonz, line 27: Can `env::join_paths` be used instead?

calebdw: Ah, that's much cleaner---thanks!

docs/config.md
```diff
@@ -1455,41 +1455,49 @@ Setting this value to zero will disable the limit entirely.
...
+and merged. Multiple paths can be specified by separating them with a
+platform-specific path separator (`:` on Unix-like systems, `;` on Windows).
 
-For example,
-```shell
-env JJ_CONFIG=/dev/null jj log       # Ignores any settings specified in the config file.
+For example, the following could be used to run `jj` without loading any user configs:
+
+```bash
+JJ_CONFIG=/dev/null jj log       # Ignores any settings specified in the config file.
```
martinvonz, line 1499: What do you think about simplifying this to `JJ_CONFIG=`? That seems quite natural now that it's a list.

martinvonz (commented):
This looks good to me, but I'll let Yuya give the final approval since he's already done a few rounds of review. Thank you!

docs/config.md
calebdw: I agree, I'll do that

cli/src/commands/config/path.rs
yuja: > > the list may be empty if no files exist,
> 
> This is not the case---the list will always be populated with either `~/.config/jj/config.toml` or `~/.jjconfig.toml` (if the config dir doesn't exist), unless of course it's been overridden by `JJ_CONFIG`.

That means `jj config list` shows lookup paths, right? I would want to pretty much avoid this inconcistency.

cli/src/config.rs
yuja: > I'm not sure what you mean by that---we are still using the Existing|New pattern

The consumers of `.wants` do either select `Want::Existing(path)` or `Want::New(path)`, so it no longer makes sense to combine them into a single `Vec<Existing|New>`.

That said, the test `check_new()` compares the result filtered by the test function:
```rs
            let new: Vec<PathBuf> = env
                .user_config_paths()
                .map(|p| p.to_path_buf())
                .filter(|p| !p.try_exists().unwrap_or(false))
                .collect();
            if new != expected {
```
so we're basically testing the test function. This doesn't make sense either.

cli/src/commands/config/path.rs
calebdw: No,  `jj config list` shows the resolved configurations and (optionally their source). 

`jj config edit` will list the resolved files if there's more than one. 

`jj config path` shows the configured lookup path or directories

cli/src/config.rs
calebdw: I can split the wants into separate wants_existing and wants_new.

However the user_config_paths() could return existing paths or paths that don't exist but can be created so i still need to filter

cli/src/commands/config/path.rs
yuja: Oops, I meant `jj config path`, not `list`, sorry. My point is that it's weird if `jj config path` _conditionally_ showed unresolved paths. Listing `~/.jj/config/conf.d/*.toml` file paths means that the paths are resolved.

cli/src/config.rs
yuja: > However the user_config_paths() could return existing paths or paths that don't exist but can be created so i still need to filter

Well, I would say that we should test the unfiltered `user_config_paths()` because that's what the production code uses. In other word, maybe we shouldn't rename `ExistingAndNew` to `Existing`?

cli/src/commands/config/path.rs
calebdw: That might be worth doing in a follow up PR though

cli/src/config.rs
calebdw: No because the previous tests didn't make sense---a specific file path can't both New and Existing, it's one or the other

yuja: My understanding is that `Want::ExistingAndNew(path)` means that the path should be included in both the `existing` list and the `new` list. I agree the naming is confusing, though. (I also think the current tests are overly abstracted.)

cli/src/commands/config/path.rs
yuja: Well, if we decide on `jj config path` showing resolved file paths, I would make it not show the default non-existing paths. In any case, this should be addressed separately as you said.

cli/src/config.rs
calebdw: > My understanding is that Want::ExistingAndNew(path) means that the path should be included in both the existing list and the new list. 

The names are confusing, but Existing means that the path exists and New means that the path **does not exist** but can be created---it is impossible for a file to be in both the `existing` and the `new` lists at the same time.

cli/src/commands/config/path.rs
calebdw: I'm not too sure about this...there's a difference between showing the search *paths* and the resolved **files**. Given that the command is called `jj config path` I would tend to expect that it shows me the configured search paths

cli/src/config.rs
yuja: Well, in the original implementation, the test function was:
```rust
        fn check_new(&self, tmp: &tempfile::TempDir) -> anyhow::Result<()> {
            let want = match self.want {
                Want::None => None,
                Want::New(want) => Some(want),
                Want::ExistingAndNew(want) => Some(want),
            }
            .map(|p| tmp.path().join(p));
            let env = self
                .resolve(tmp.path())
                .map_err(|e| anyhow!("new_config_path: {e}"))?;
            let got = env.user_config_path();
            if got != want.as_deref() {
                return Err(anyhow!("new_config_path: got {got:?}, want {want:?}"));
            }
            Ok(())
        }
```

Since `env.user_config_path()` should include both `existing` and `new` paths, `want` is NOT filtered solely by `Want::New`. The test is badly named, but the code makes sense. We should test the production code `user_config_path()`, not the test implementation which filters `user_config_path()` by `path.exists()`.

calebdw: We are testing the production code `user_config_path()`, I just break it up into the existing and the non-existing paths:

```rs
        fn check_existing(&self, tmp: &tempfile::TempDir) -> anyhow::Result<()> {
            let env = self.resolve(tmp.path());
            let expected: Vec<PathBuf> = self
                .wants_existing
                .iter()
                .map(|path| tmp.path().join(path))
                .collect();
            let existing: Vec<PathBuf> = env
                .existing_user_config_paths()
                .map(|p| p.to_path_buf())
                .collect();
            if existing != expected {
                return Err(anyhow!(
                    "existing_config_path mismatch: got {existing:?}, expected {expected:?}",
                ));
            }
            Ok(())
        }

        fn check_new(&self, tmp: &tempfile::TempDir) -> anyhow::Result<()> {
            let env = self.resolve(tmp.path());
            let expected: Vec<PathBuf> = self
                .wants_new
                .iter()
                .map(|path| tmp.path().join(path))
                .collect();
            let new: Vec<PathBuf> = env
                .user_config_paths()
                .map(|p| p.to_path_buf())
                .filter(|p| !p.try_exists().unwrap_or(false))
                .collect();
            if new != expected {
                return Err(anyhow!(
                    "new_config_path mismatch: got {new:?}, expected {expected:?}"
                ));
            }
            Ok(())
        }
```

yuja: And I'm saying that `check_new()` shouldn't do `filter(|p| !p.try_exists().unwrap_or(false))`. What we need to assert is that `.user_config_paths()` works as expected, right?

If the test isn't overly abstracted, it would look something like:
```rust
assert_eq!(
    env.existing_user_config_paths(),
    vec!["foo.toml"]
);
assert_eq!(
    env.user_config_paths(),
    vec!["foo.toml", "bar.toml"]
);
```

calebdw: No, because `user_config_paths()` contains both existing and non-existing paths, and in this test I'm specifically testing that it contains the expected **non-existing** paths. The previous test asserts that it contains the expected **existing** paths and so both sides are tested.

I don't see me filtering out the existent paths in the test any different than the `existing_user_config_paths()` filtering out the non-existing paths:

```rs
    /// Returns the paths to the user-specific config files or directories.
    pub fn user_config_paths(&self) -> impl Iterator<Item = &Path> {
        self.user_config_paths.iter().map(|p| p.as_path())
    }

    /// Returns the paths to the existing user-specific config files or
    /// directories.
    pub fn existing_user_config_paths(&self) -> impl Iterator<Item = &Path> {
        self.user_config_paths.iter().filter_map(|p| match p {
            ConfigPath::Existing(path) => Some(path.as_path()),
            _ => None,
        })
    }
```

I could create a separate `new_user_config_paths()` function, but I didn't want to do that because it would only be referenced in the test and isn't useful

calebdw: Sure I could do something like the following where have two arrays with the same information, but that seems rather duplicative:

```rs
    #[test]
    fn test_config_path_all_existing() -> anyhow::Result<()> {
        TestCase {
            files: [
                "config/jj/conf.d/_",
                "config/jj/config.toml",
                "home/.jjconfig.toml",
            ],
            env: UnresolvedConfigEnv {
                home_dir: Some("home".into()),
                config_dir: Some("config".into()),
                ..Default::default()
            },
            wants_existing: [
                "home/.jjconfig.toml",
                "config/jj/config.toml",
                "config/jj/conf.d",
            ],
            wants: [
                "home/.jjconfig.toml",
                "config/jj/config.toml",
                "config/jj/conf.d",
            ],
        .run()
    }

Comment (calebdw) [reply cli/src/config.rs:1518]: I've removed the filter and the separate `wants_{expected,new}` arrays---hopefully that makes the test more clear

File: cli/tests/test_config_command.rs [hunk 21]
```diff
@@ -984,6 +985,22 @@ fn test_config_path() {
     ");
 }
 
+#[test]
+fn test_config_path_multiple() {
+    let mut test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let config_path = test_env.config_path().join("config.toml");
+    let work_config_path = test_env.config_path().join("conf.d").join("work.toml");
```
yuja, line 993: nit: can you change the `work_config_path` to `conf.d`? I think that's more realistic example.

cli/src/config.rs
yuja: > I could create a separate `new_user_config_paths()` function, but I didn't want to do that because it would only be referenced in the test and isn't useful

That's true, and I believe testing the inlined version of `new_user_config_paths()` wouldn't make sense either.

> I've removed the filter and the separate `wants_{expected,new}` arrays---hopefully that makes the test more clear

Thanks.

yuja (approved):

PR #5985, opened 2025-03-11
git push: add support for `revsets.git-push`

### LINKED ISSUE #2434: Default revset for `jj git push` is imperfect

1. jj branch c a
1. jj new
1. jj branch c b
2. jj git push --all
3. jj branch d b
4. jj branch s a

At this point, the repo should look like:

```
@ a@origin b
|
o b@origin
```

Now, do `jj git push`.

## Expected Behavior

The branch `a` gets deleted on the remote, `b` gets moved.

## Actual Behavior

Nothing happened.

This works as documented, as it pushes `remote_branches(remote="origin")..@`. However, I find this behavior against the spirit of what `jj git push` is supposed to do.

I'm not sure if there's a straightforward solution, though.

## Specifications

- Platform: Debian linux
- Version: recent nightly

### LINKED ISSUE #3650: FR: `revsets.push`

PhilipMetzger:
This let's our users finally configure their own push revset. As the current one may not be perfect for every situation.

Fixes #3650 
Closes #2434

# Checklist

If applicable:


cli/src/config/revsets.toml
```diff
@@ -9,6 +9,7 @@ simplify-parents = "reachable(@, mutable())"
 # suppress symbol resolution error.
 log = "present(@) | ancestors(immutable_heads().., 2) | present(trunk())"
 sign = "reachable(@, mutable())"
+push = "remote_bookmarks(remote=<remote>)..@"
```
PhilipMetzger, line 12: @yuja what alias hack have you mind to solve the problem here?

yuja: Naming aside: `'push(remote)' = 'remote_bookmarks(remote=remote)..@'` (i.e. substitute like an alias function)

PhilipMetzger: done

cli/src/commands/git/push.rs
```diff
@@ -766,15 +769,16 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
-            None,
-        )
-        .range(&RevsetExpression::working_copy(workspace_id.clone()))
-        .intersection(&RevsetExpression::bookmarks(StringPattern::everything()));
+        let revs = workspace_command.settings().get_string("revsets.push")?;
+        // This obviously isn't correct
+        let mut map = TemplateAliasesMap::new();
+        map.insert("remote", remote_name);
+        let template = parse(&revs, &map)?;
+
```
PhilipMetzger, line 777: At this point its no longer clear to me how to proceed, as trying to mutate the existing `aliases_map` in the workspace_command_helper via `borrow_mut()` just doesn't work.

cli/src/commands/git/push.rs
```diff
@@ -766,15 +771,21 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+            "TODO",
+            chrono::Utc::now().fixed_offset().into(),
+            &extensions,
             None,
-        )
-        .range(&RevsetExpression::working_copy(workspace_id.clone()))
-        .intersection(&RevsetExpression::bookmarks(StringPattern::everything()));
+        );
+        let mut diags = RevsetDiagnostics::default();
+        let (expression, _) = revset::parse_with_modifier(&mut diags, &revs, &context)?;
```
yuja, line 787: This will probably help to set local variable `remote`. The first two patches are sent as #6142.

https://github.com/jj-vcs/jj/compare/main...yuja:jj:push-nuxxkpyxzprs

PhilipMetzger:
I think this is ready, it's just missing the section in `config.md` for the new revset.

cli/src/commands/git/push.rs
```diff
@@ -946,24 +952,34 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+        local_variables.insert(remote.as_str(), template_expr);
+        let extensions = RevsetExtensions::default();
+        let context = RevsetParseContext {
+            aliases_map: &aliases_map,
+            local_variables,
+            user_email: "TODO",
+            date_pattern_context: chrono::Utc::now().fixed_offset().into(),
+            extensions: &extensions,
+            workspace: None,
+        };
```
yuja, line 971: use `RevsetParseContext` created by `workspace_command`

cli/src/commands/git/push.rs
```diff
@@ -946,24 +952,34 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
-        let workspace_name = workspace_command.workspace_name();
-        let expression = RevsetExpression::remote_bookmarks(
-            StringPattern::everything(),
-            StringPattern::exact(remote),
-            None,
-        )
-        .range(&RevsetExpression::working_copy(workspace_name.to_owned()))
-        .intersection(&RevsetExpression::bookmarks(StringPattern::everything()));
+        // Get the default revset specified by the user.
+        let push_revset = workspace_command.settings().get_string("revsets.push")?;
```
yuja, line 956: Maybe it should be named as `revsets.git-push` or `git.push-revisions`?

I have no idea which is better overall. This revset expression is a bit special in that we substitute `remote` with the actual remote name. OTOH, it's not so different than `revsets.<command>` for `<command>`'s default.

cli/src/commands/git/push.rs
```diff
@@ -946,24 +952,34 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+        let aliases_map = RevsetAliasesMap::new();
+        let mut local_variables = HashMap::new();
+        let template_expr = parse_program(&push_revset).map_err(|e| {
+            user_error_with_hint(e, "revsets.push requires to take a `remote` by default")
+        })?;
+        local_variables.insert(remote.as_str(), template_expr);
```
yuja, line 962: Maybe you mean `.insert("remote", remote_expr)`? I think `remote_expr` can be constructed programmatically, but it's also okay to parse formatted `remote` symbol.

cli/tests/test_git_push.rs
```diff
@@ -2591,3 +2586,63 @@ fn test_git_push_git2_warning() {
...
+    ");
+    // We should try to push the first two commits but not any containing "wip:".
+    let output = work_dir.run_jj(["git", "push"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Config error: Value not found for revsets.push
+    For help, see https://jj-vcs.github.io/jj/latest/config/ or use `jj help -k config`.
+    [EOF]
+    [exit status: 1]
+    ");
```
yuja, line 2641: The test looks just broken.

cli/src/commands/git/push.rs
PhilipMetzger: > Maybe it should be named as `revsets.git-push` or `git.push-revisions`?

I'm fine with any name, the FR was for `revsets.push` though to be consistent with other commands. 

> OTOH, it's not so different than `revsets.<command>` for `<command>`'s default.

yes, that's also true.

cli/src/commands/git/push.rs
PhilipMetzger: > Maybe you mean `.insert("remote", remote_expr)`?

what do you mean by `remote_expr` here?

> I think `remote_expr` can be constructed programmatically, but it's also okay to parse formatted `remote` symbol.

Yes, that also can be done, I currently think that if we parse the expression before evaluating it we can at least warn if `revsets.push` doesn't take a remote symbol.

yuja: > what do you mean by remote_expr here?

The remote name string, something like `ExpressionNode { ExpressionKind::String("origin") }`. We're going to add substitution rule that replaces `remote` symbol/variable to the actual remote name, right?

PhilipMetzger: Ok, that finally makes it click for me. I've pushed a version which addresses all of your comments.

cli/src/commands/git/push.rs
PhilipMetzger: done

cli/tests/test_git_push.rs
PhilipMetzger: Yes, since my lookup fails (`config.get("revsets.push")` finds nothing) and thus all those tests are currently broken.

cli/tests/test_git_push.rs
```diff
@@ -2195,6 +2193,87 @@ fn test_git_push_rejected_by_remote() {
...
+    [EOF]
+    ");
+    // We should try to push the first two commits but not any containing "wip:".
+    let output = work_dir.run_jj(["git", "push"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Nothing changed.
+    [EOF]
+    ");
+}
```
PhilipMetzger, line 2275: Any idea for a good test for this? thanks

cli/src/commands/git/push.rs
cenviity: +1 for `revsets.git-push` since this config is specifically for pushing to a Git remote.

PhilipMetzger: renamed to `git-push`

cli/tests/cli-reference@.md.snap
```diff
@@ -1269,7 +1269,7 @@ Create a new Git backed repo
 
 Push to a Git remote
 
-By default, pushes tracking bookmarks pointing to `remote_bookmarks(remote=<remote>)..@`. Use `--bookmark` to push specific bookmarks. Use `--all` to push all bookmarks. Use `--change` to generate bookmark names based on the change IDs of specific commits.
+By default, pushes tracking bookmarks defined by the `revsets.'push(remote)'` revset. Use `--bookmark` to push specific bookmarks. Use `--all` to push all bookmarks. Use `--change` to generate bookmark names based on the change IDs of specific commits.
```
pksunkara, line 1272:
```suggestion
By default, pushes tracking bookmarks defined by the `revsets.'git-push(remote)'` revset. Use `--bookmark` to push specific bookmarks. Use `--all` to push all bookmarks. Use `--change` to generate bookmark names based on the change IDs of specific commits.
```

cli/tests/test_git_push.rs
PhilipMetzger: this now shows that we refused to push the "local/" bookmark which is good enough for me

cli/src/config-schema.json
```diff
@@ -585,6 +585,11 @@
                     "type": "string",
                     "description": "Set of revisions to prioritize when rendering the graph for jj log",
                     "default": "present(@)"
+                },
+                "git-push": {
```
martinvonz, line 775: Maybe the reason the test fails is that the config variable is called `git-push(remote)` and here it's just `git-push`

PhilipMetzger: that was it thanks!

cli/src/commands/git/push.rs
```diff
@@ -966,25 +971,29 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
-            StringPattern::everything(),
-            StringPattern::exact(remote),
-            None,
-        )
-        .range(&RevsetExpression::working_copy(workspace_name.to_owned()))
-        .intersection(&RevsetExpression::bookmarks(StringPattern::everything()));
+        // Get the default revset specified by the user.
+        let push_revset = workspace_command
+            .settings()
+            .get_string("revsets.'git-push(remote)'")?;
```
yuja: Shouldn't this be `revsets.git-push`? The parameter name can't be changed by user, and `git-push()` (with dashes) isn't a valid function name.

cli/src/commands/git/push.rs
```diff
@@ -966,25 +971,29 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+        let mut context = workspace_command.env().revset_parse_context();
+        context.local_variables.insert(
+            "remote",
+            ExpressionNode {
+                kind: ExpressionKind::String(remote.as_str().to_string()),
```
yuja: Perhaps, this will have to be `StringPattern { kind: "exact", .. }`. The default is substring match.

cli/src/commands/git/push.rs
```diff
@@ -966,25 +971,29 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+                span: Span::new("<internal>", 0, 10).expect("programmatic span shouldn't fail"),
```
yuja: nit: might be better to use `remote.as_str()`. I think this span can be included in error message if the user expression is wrong.

cli/src/commands/git/push.rs
```diff
@@ -966,25 +971,29 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+            },
+        );
+        let mut diags = RevsetDiagnostics::default();
+        let (expression, _) = revset::parse_with_modifier(&mut diags, &push_revset, &context)?;
```
yuja: `print_parse_diagnostics(..)`

cli/src/commands/git/push.rs
```diff
@@ -966,25 +971,29 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
     if use_default_revset {
-        // remote_bookmarks(remote=<remote>)..@
-        let workspace_name = workspace_command.workspace_name();
-        let expression = RevsetExpression::remote_bookmarks(
-            StringPattern::everything(),
-            StringPattern::exact(remote),
-            None,
-        )
-        .range(&RevsetExpression::working_copy(workspace_name.to_owned()))
-        .intersection(&RevsetExpression::bookmarks(StringPattern::everything()));
```
yuja: `.intersection(&RevsetExpression::bookmarks(StringPattern::everything()));` should be retained. That's probably why the test outputs differ.

CHANGELOG.md
```diff
@@ -122,6 +126,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `colors."diff removed token" = { bg = "ansi-color-52", underline = false }`
   will apply a dark red background on removed words in diffs.
 
+* Log node templates are now specified in toml rather than hardcoded.
```
yuja: nit: bad merge?

cli/src/commands/git/push.rs
PhilipMetzger: How do you pass the `remote` to  `revsets.git-push` if it doesn't take it as a parameter? Should it just try to expand the `remote` in the revset.

yuja: The implementation should be the same. I just mean it seems confusing to use fake function-like config key because the key isn't parsed as a function.

cli/src/commands/git/push.rs
PhilipMetzger: I just did that and it introduces panics since `remote` can be empty.

cli/src/commands/git/push.rs
PhilipMetzger: Ok, done

cli/src/commands/git/push.rs
PhilipMetzger: done

cli/src/commands/git/push.rs
PhilipMetzger: done

CHANGELOG.md
PhilipMetzger: removed

cli/src/commands/git/push.rs
PhilipMetzger: fixed it by not giving it a arbitary range

cli/src/commands/git/push.rs
PhilipMetzger: I added it back but that should definitely be in the config, since we really shouldn't do that to user revsets

CHANGELOG.md
```diff
@@ -124,14 +128,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
 
-* Log node templates are now specified in toml rather than hardcoded.
-
 * Templates now support `json(x)` function to serialize values in JSON format.
 
 * The ANSI 256-color palette can be used when configuring colors. For example,
   `colors."diff removed token" = { bg = "ansi-color-52", underline = false }`
   will apply a dark red background on removed words in diffs.
 
+* Log node templates are now specified in toml rather than hardcoded.
```
yuja: nit: bad edits

cli/src/commands/git/push.rs
yuja: > that should definitely be in the config, since we really shouldn't do that to user revsets

No. It's the requirement of `jj git push`, which pushes bookmarks, not arbitrary commits.

cli/src/commands/git/push.rs
```diff
@@ -966,25 +972,35 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+            .settings()
+            .get_string("revsets.git-push")?;
+        let mut context = workspace_command.env().revset_parse_context();
+        context.local_variables.insert(
+            "remote",
+            ExpressionNode {
+                kind: ExpressionKind::StringPattern {
+                    kind: "exact",
+                    value: remote.as_str().to_string(),
+                },
```
yuja: Can you add test that would fail if this were substring pattern?

yuja (commented):
Thanks.

cli/src/commands/git/push.rs
PhilipMetzger: Ok, then I'll expand the documentation to mention it.

cli/src/commands/git/push.rs
PhilipMetzger: I tried to add one. PTAL

CHANGELOG.md
PhilipMetzger: fixed

cli/tests/test_git_push.rs
```diff
@@ -2244,6 +2244,93 @@ fn test_git_push_rejected_by_remote() {
...
+                "remote",
+                "add",
+                &remote_interpolated,
+                remote_git_dir.to_str().unwrap(),
+            ])
+            .success();
+    };
+    // Add some new remotes which could get matched on substrings.
+    add_new_git_remote("upstreeam");
+    add_new_git_remote("upstream");
```
PhilipMetzger: this currently panics because the remotes contain a trailing slash

CHANGELOG.md
```diff
@@ -208,7 +212,6 @@ Thanks to the people who made this release happen!
 
 * `jj evolog` can show associated operations for commits created by new jj
   versions.
-
```
yuja, line 2150: nit: restore blank line

cli/src/commands/git/push.rs
```diff
@@ -966,25 +972,35 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+                span: Span::new(remote.as_str(), 0, remote.as_str().len())
+                    .expect("programmatic span shouldn't fail"),
+            },
+        );
+        let mut diags = RevsetDiagnostics::default();
+        let (mut expression, _) = revset::parse_with_modifier(&mut diags, &push_revset, &context)?;
+        print_parse_diagnostics(ui, "Failed to expand 'remote' in expression", &diags)?;
```
yuja: nit: the error usually exists in the user expression. Just print "In `revsets.git-push`" or something?

cli/tests/test_git_push.rs
yuja: Maybe you can extract remote repo initialization from `set_up()`? The remote content wouldn't matter.

CHANGELOG.md
PhilipMetzger: done

cli/src/commands/git/push.rs
PhilipMetzger: done

cli/tests/test_git_push.rs
PhilipMetzger: done, this works now

cli/src/commands/git/push.rs
```diff
@@ -968,25 +974,35 @@ fn find_bookmarks_targeted_by_revisions<'a>(
...
+                kind: ExpressionKind::StringPattern {
+                    kind: "exact",
+                    value: remote.as_str().to_string(),
+                },
+                span: Span::new(remote.as_str(), 0, remote.as_str().len())
+                    .expect("programmatic span shouldn't fail"),
+            },
+        );
+        let mut diags = RevsetDiagnostics::default();
+        let (mut expression, _) = revset::parse_with_modifier(&mut diags, &push_revset, &context)?;
```
yuja: nit: can you insert `let (None | Some(RevsetModifier::All)) = modifier;` to ensure that compilation would fail if we add new modifier?

cli/tests/test_git_push.rs
```diff
@@ -2244,6 +2244,88 @@ fn test_git_push_rejected_by_remote() {
...
+fn test_git_push_custom_revset() {
+    let test_env = TestEnvironment::default();
+    set_up(&test_env);
+    let work_dir = test_env.work_dir("local");
+    // add a custom revset which simulates keeping a custom set of local only
+    // bookmarks.
+    test_env.add_config(
+        r#"
+    [revsets]
+    'git-push' = "remote_bookmarks(remote=remote)..@ ~untracked_remote_bookmarks(glob:'local/*')"
```
yuja: You probably mean `bookmarks(glob:'local/*')`. There are no remote bookmarks named `local/*`.

BTW, this doesn't guarantee that the bookmark `local/*` will never be pushed. If there are non-`local/*` bookmark pointing to the same commit, the commit will be selected and all bookmarks pointing to that commit will be pushed.

cli/tests/test_git_push.rs
```diff
@@ -2244,6 +2244,88 @@ fn test_git_push_rejected_by_remote() {
...
+    work_dir.run_jj(["desc", "-m", "Foo"]).success();
+    work_dir.run_jj(["bookmark", "move", "bookmark1"]).success();
+    // This will push to two remotes if we use substring matching.
+    let output = work_dir.run_jj(&["git", "push", "--remote", "stream"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Warning: Refusing to create new remote bookmark bookmark1@stream
+    Hint: Use --allow-new to push new bookmark. Use --remote to specify the remote to push to.
+    Nothing changed.
+    [EOF]
```
yuja: This wouldn't fail with substring match. We'll need to set up two remotes with the same bookmark pointing to different commits, and push it to the other remote, something like:
```
foo@stream foo
|
foo@upstream
```

cli/tests/test_git_push.rs
PhilipMetzger: I realized that this revset also is expressible with `tracked_remote_bookmarks(remote=remote)..@` since the `local/*` is untracked.

cli/src/commands/git/push.rs
PhilipMetzger: done

cli/tests/test_git_push.rs
PhilipMetzger: done

cli/tests/test_git_push.rs
PhilipMetzger: I tried to rewrite the test around this but it currently doesn't look right yet.

cli/tests/test_git_push.rs
```diff
@@ -2729,6 +2729,124 @@ fn test_git_push_unmapped_refs() {
...
+    ");
+    // We should try to push everything except the commit on the "local/" bookmark.
+    let output = work_dir.run_jj(["git", "push"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Warning: Refusing to create new remote bookmark local/stuff@origin
+    Hint: Run `jj bookmark track local/stuff --remote=origin` and try again.
+    Nothing changed.
+    [EOF]
+    ");
```
PhilipMetzger, line 3094: This should only push the bookmark in the revset but it now fails to do that

martinvonz: I'm not sure how to read this comment. Is the PR not ready yet? Mark it as draft in case? Same with the other unresolved comment above.

PhilipMetzger: I was wrong, the output here hasn't changed in the last few versions. 

> Is the PR not ready yet?

It implements the functionality as wanted, except that https://github.com/jj-vcs/jj/pull/5985#discussion_r2269519637 isn't addressed yet.

PR #5992, opened 2025-03-12
github: add `required-checks` job to `ci`

emilazy:
First batch of my attempt to improve the user experience of our CI. See commit messages for details.

I don’t really entirely know what I’m doing, and GitHub Actions is really difficult to test, so it’s entirely likely some of this stuff will break things. Scrutiny is more than welcome. @martinvonz will have to follow the instructions in “github: add `result` job to `ci`” if this is approved.

# Checklist

If applicable:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -285,3 +285,28 @@ jobs:
         with:
           sarif_file: results.sarif
           category: zizmor
+
+  required-checks:
+    name: required checks
+    if: ${{ always() && github.event_name == 'merge_group' }}
+    needs:
+    - test
+    - no-git
+    - build-nix
+    - check-protos
+    - check-rustfmt
+    - check-clippy
+    - check-cargo-deny
+    - check-codespell
+    - check-doctests
+    - check-mkdocs
+    # - check-mkdocs-latest
+    # - check-zizmor
+    runs-on: ubuntu-latest
+    steps:
+    - if: >-
+        ${{
+          contains(needs.*.result, 'failure')
+          || contains(needs.*.result, 'cancelled')
+        }}
+      run: exit 1


martinvonz:
Since this will be tested in prod, i suppose we should try to merge it slowly so it's easier to detect if anything breaks. I'm happy to merge the first two commits if you can move them to a separate PR

.github/workflows/ci.yml
```diff
@@ -280,3 +280,27 @@ jobs:
         with:
           sarif_file: results.sarif
           category: zizmor
+
+  result:
```
martinvonz, line 284: nit: maybe something like `all-checks` or `required-checks` would be easier to understand

martinvonz, line 284: Good idea!

.github/workflows/ci.yml
```diff
@@ -1,15 +1,6 @@
...
 on:
-  push:
-    branches:
-      - '**'
-      # Disable builds on these branches, because they will become a pull
-      # request, and be handled by merge_group below.
-      - '!dependabot/**'
-      # `main` and `gh-readonly-queue` are handled by merge_group specifically.
-      - '!gh-readonly-queue/**'
-      - '!main'
```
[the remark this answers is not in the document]
emilazy: Hmm. Maybe. I haven’t ever seen anyone doing it like that. I think it would be weird with forks; if you look at PRs from forks, they don’t have the (push) checks, because those run in the context of the user’s fork rather than the PR. So it seems worse to me. If we find the `push` trigger useful then we could try and do fancier stuff to mutex it with the `pull_request` one but my guess is that all it accomplishes right now is wasting concurrency slots.

.github/workflows/ci.yml
```diff
@@ -280,3 +280,27 @@ jobs:
         with:
           sarif_file: results.sarif
           category: zizmor
+
+  result:
+    if: ${{ always() && github.event_name == 'merge_group' }}
```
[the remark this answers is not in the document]
emilazy: It’s `true` but, like, [way more true](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions)? It seems like unless you include one of these magical functions in your condition, then failures of `needs` checks will cause your job not to run, even if your `if` is true, and even though `always() && x` is equal to `x` as a boolean.

GitHub Actions is not my favourite continuous integration system.

.github/workflows/ci.yml
emilazy: Sure, done.

emilazy: I stole it from the internet when trying to figure out how to have different requirements for queueing and the final merge :) But moving more things out of the repository settings seems good for all of us as well.

.github/workflows/ci.yml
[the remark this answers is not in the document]
emilazy: BTW I’m not even totally sure if the `&&` works here. There’s hackier more awful stuff we could do if the laws of boolean logic are totally out of the window, but I live in hope.

.github/workflows/ci.yml
martinvonz: nit: update commit message to match

emilazy:
> Since this will be tested in prod, i suppose we should try to merge it slowly so it's easier to detect if anything breaks. I'm happy to merge the first two commits if you can move them to a separate PR

I wrote a reply here but it looks like I forgot to post it. Yes, seems like a good idea; maybe the first three commits to test the dependabot logic? Worst case we have to manually merge them for a bit. I think we would need to approve them after PR checks run until the required checks stuff is merged, but that it should otherwise work fine.

martinvonz:
> Yes, seems like a good idea; maybe the first three commits to test the dependabot logic?

Yes, sounds good. The first two seem safe enough that I'm not worried about bundling them with the third one.

martinvonz:
The last commit also seems pretty independent and could be pulled out to a separate PR that can be fast-tracked IMO

emilazy:
Split off https://github.com/jj-vcs/jj/pull/6004. I personally see the last commit as the scary one, since it’s the one that requires adjusting repository configuration and depends on weird parts of GitHub Actions that I don’t fully understand. Are you sure we shouldn’t merge the remaining three here first instead? I think they have more limited potential for fallout even if they’re wrong.

martinvonz:
> I personally see the last commit as the scary one, since it’s the one that requires adjusting repository configuration and depends on weird parts of GitHub Actions that I don’t fully understand.

Doesn't it just add a check? If we don't mark that check required yet, then it's pretty harmless, I thought.

martinvonz (approved):

emilazy:
Okay, this is now just the `required-checks` commit. I guess we merge this, check that it works, and then handle the repository setting changes from the commit message and https://github.com/jj-vcs/jj/pull/6005.

PR #5993, opened 2025-03-12
repo: simplify `merge_view` using `diff_named_commit_ids`

ilyagr:
Fixes a TODO


ilyagr:
@yuja, at the moment I have three versions of the code in this PR. The second commit is my interpretation of your suggestion in https://github.com/jj-vcs/jj/pull/5901#discussion_r1988551891. The third commit seems to simplify it, and I think I prefer it (the diff of the third commit is probably the most useful thing to review). WDYT?

lib/src/repo.rs
```diff
@@ -1664,34 +1665,27 @@ impl MutableRepo {
...
                     self.view_mut()
-                        .set_wc_commit(workspace_id.clone(), other_wc_commit.clone());
+                        .set_wc_commit(workspace_id.clone(), other_target.clone());
+                } else {
+                    // If there's a conflict, we keep the `self` side unchanged
                 }
             } else {
                 // The other side removed the workspace. We want to remove it even if the self
-                // side changed the working-copy commit.
+                // side changed the working-copy commit. TODO(ilyagr): I'm not sure why
```
ilyagr, line 1684: Continuing the discussion from https://github.com/jj-vcs/jj/pull/5901#discussion_r1990499773, why do you think this makes sense?

I itch to "fix" this (not in this PR), but I'm not sure what the consequences would be.

ilyagr: Quoting @yuja from #5901:

> I assume the expectation is that jj workspace forget is stronger than e.g. jj new because the workspace is forgotten anyway. I don't know if that's always true, but it would be more surprising if jj workspace forget was cancelled?

ilyagr: I guess `jj workspace forget` is the only case a workspace disappears? If so, it makes sense that, whether or not this is a bug, people would encounter it very rarely.

lib/src/repo.rs
```diff
@@ -1665,35 +1666,27 @@ impl MutableRepo {
...
+        // pairs
+        let changed_wc_commits = diff_named_commit_ids(base.wc_commit_ids(), other.wc_commit_ids());
+        for (workspace_id, (base_target, other_target)) in changed_wc_commits {
+            let self_target = self.view().get_wc_commit_id(workspace_id);
+            if let Some(&new_target) = trivial_merge(&[self_target, base_target, other_target]) {
+                if let Some(target) = new_target.cloned() {
+                    self.view_mut().set_wc_commit(workspace_id.clone(), target);
+                } else {
+                    self.view_mut().remove_wc_commit(workspace_id);
+                }
```
yuja: If the underlying type was `RefTarget`, this part will be a single function call: `.set_wc_target(workspace_id, target)`.

(just a comment. no action requested.)

ilyagr: I'm wondering whether we should have some more complex understanding of conflicted workspaces that keeps track of what the workspace commit *should* be (possibly conflicted) and what the working copy on disk looks like.

lib/src/repo.rs
```diff
@@ -1665,35 +1666,27 @@ impl MutableRepo {
...
             } else {
-                self.view_mut().remove_wc_commit(&workspace_id);
+                // If there's a conflict, we keep the `self` side unchanged, with one exception
+                // below.
+                if other_target.is_none() {
+                    // The other side removed the workspace. We want to remove it even if the self
+                    // side changed the working-copy commit. TODO(ilyagr): I'm not sure why
+                    self.view_mut().remove_wc_commit(workspace_id);
+                }
```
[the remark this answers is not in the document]
ilyagr: To me, it seemed to hide the logic inside another function. Also, the function is not really working as a black box: it's not obvious that the code is correct unless you know what `trivial_merge` would return if `base == other` *and* it's output needs to be ignored if `other_target.is_none()`.

I haven't incorporated the "I would write as follows, though" suggestion into my thinking, maybe it simplifies things.

I probably won't get to fixing this further today.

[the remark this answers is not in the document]
yuja: Well, `trivial_merge()` is the function we should try first, right? (We could also use `merge_ref_targets()`, but I don't know if it makes sense to resolve conflicts of working-copy targets based on ancestor relations.)

The logic should be:
1. if merged cleanly (including `A-B+A`), the merge result should be used
2. otherwise, the result is arbitrary, so pick something saner

We don't have 2 for the other refs because conflict can be stored as is.

lib/src/repo.rs
```diff
@@ -1664,34 +1665,38 @@ impl MutableRepo {
...
+            let base_wc_commit = base.get_wc_commit_id(workspace_id);
+            let changed_commit = if my_commit == base_wc_commit {
+                other_commit
+            } else if other_commit == base_wc_commit {
+                // Keep the `self` side unchanged. Returning `my_commit` here
+                // would look more symmetric, but would have the same outcome.
+                continue;
+            } else if other_commit.is_none() {
+                // The other side removed the workspace and it moved on our
+                // side. We want to remove it. TODO(ilyagr): I'm not sure why.
```
martinvonz: Updating a workspace to a different commit doesn't impact the commit graph or bookmarks so I don't think we need to preserve any information about that.

FWIW, what probably happened was that the user ran `jj workspace forget` in one workspace and something like `jj rebase` in another workspace. I don't think the user wants to get the workspace back because they ran that rebase at the same time.

PR #5994, opened 2025-03-12
cli: config {edit,set,unset}: prompt when multiple files exist

calebdw:
Hello!

This allows the user to select a particular file when using multiple configs. In the event that a prompt cannot be displayed, the first file will be automatically selected.

This is part of my work on #5971 but was relatively independent so decided to send it.

Thanks!

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/config/edit.rs
+++ b/cli/src/commands/config/edit.rs
@@ -31,12 +31,12 @@ pub struct ConfigEditArgs {
 
 #[instrument(skip_all)]
 pub fn cmd_config_edit(
-    _ui: &mut Ui,
+    ui: &mut Ui,
     command: &CommandHelper,
     args: &ConfigEditArgs,
 ) -> Result<(), CommandError> {
     let editor = command.text_editor()?;
-    let file = args.level.edit_config_file(command)?;
+    let file = args.level.edit_config_file(ui, command)?;
     if !file.path().exists() {
         file.save()?;
     }
--- a/cli/src/commands/config/mod.rs
+++ b/cli/src/commands/config/mod.rs
@@ -21,7 +21,6 @@ mod unset;
 
 use std::path::Path;
 
-use itertools::Itertools as _;
 use jj_lib::config::ConfigFile;
 use jj_lib::config::ConfigSource;
 use tracing::instrument;
@@ -81,19 +80,25 @@ impl ConfigLevelArgs {
         }
     }
 
-    fn edit_config_file(&self, command: &CommandHelper) -> Result<ConfigFile, CommandError> {
+    fn edit_config_file(
+        &self,
+        ui: &Ui,
+        command: &CommandHelper,
+    ) -> Result<ConfigFile, CommandError> {
         let config_env = command.config_env();
         let config = command.raw_config();
         let pick_one = |mut files: Vec<ConfigFile>, not_found_error: &str| {
             if files.len() > 1 {
-                // TODO: prompt or pick the last?
-                return Err(user_error(format!(
-                    "Cannot determine config file to edit:\n{}",
-                    files
-                        .iter()
-                        .map(|file| format!("  {}", file.path().display()))
-                        .join("\n")
-                )));
+                let mut choices = vec![];
+                let mut formatter = ui.stderr_formatter();
+                for (i, file) in files.iter().enumerate() {
+                    writeln!(formatter, "{}: {}", i + 1, file.path().display())?;
+                    choices.push((i + 1).to_string());
+                }
+                drop(formatter);
+                let choice =
+                    ui.prompt_choice("Choose a config file (default 1)", &choices, Some("1"))?;
+                return Ok(files[choice.parse::<usize>().unwrap() - 1].clone());
             }
             files.pop().ok_or_else(|| user_error(not_found_error))
         };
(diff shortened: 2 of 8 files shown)


cli/src/commands/config/edit.rs
```diff
@@ -36,7 +36,7 @@ pub fn cmd_config_edit(
     args: &ConfigEditArgs,
 ) -> Result<(), CommandError> {
     let editor = command.text_editor()?;
-    let file = args.level.edit_config_file(command)?;
+    let file = args.level.edit_config_file(_ui, command)?;
```
yuja, line 39: nit: rename to `ui`

cli/src/commands/config/mod.rs
```diff
@@ -81,19 +80,33 @@ impl ConfigLevelArgs {
         }
     }
 
-    fn edit_config_file(&self, command: &CommandHelper) -> Result<ConfigFile, CommandError> {
+    fn edit_config_file(
+        &self,
+        ui: &mut Ui,
```
yuja, line 85: nit: `&mut` isn't required

cli/src/commands/config/mod.rs
```diff
@@ -81,19 +80,33 @@ impl ConfigLevelArgs {
...
-                // TODO: prompt or pick the last?
-                return Err(user_error(format!(
-                    "Cannot determine config file to edit:\n{}",
-                    files
-                        .iter()
-                        .map(|file| format!("  {}", file.path().display()))
-                        .join("\n")
-                )));
+                let mut choices = vec![];
+                let mut formatter = ui.stdout_formatter();
```
yuja, line 93: nit: maybe we should use `stderr` because the prompt is now printed to stderr. (I think the existing caller is wrong.)

cli/src/commands/config/mod.rs
```diff
@@ -81,19 +80,33 @@ impl ConfigLevelArgs {
...
+                    choices.push((i + 1).to_string());
+                }
+                drop(formatter);
+                let choice =
+                    ui.prompt_choice("Choose a config file (default 1)", &choices, Some("1"));
+                match choice {
+                    Ok(choice) => {
+                        return Ok(files
+                            .get(choice.parse::<usize>().unwrap() - 1)
+                            .ok_or_else(|| user_error(not_found_error))?
```
yuja, line 105: nit: just `.unwrap()`? `ui.prompt_choice()` should guarantee that the return value is valid.

cli/src/commands/config/unset.rs
```diff
@@ -39,7 +39,7 @@ pub fn cmd_config_unset(
     command: &CommandHelper,
     args: &ConfigUnsetArgs,
 ) -> Result<(), CommandError> {
-    let mut file = args.level.edit_config_file(command)?;
+    let mut file = args.level.edit_config_file(_ui, command)?;
```
yuja, line 42: nit: `ui`

yuja (commented):
Sounds good to me, thanks.

cli/src/commands/config/mod.rs
calebdw: It's `&mut` all the way up, so I'd have to change a lot of other function signatures if you want me to remove it...

yuja: Just change this function. `cmd_*()` should take `&mut Ui` because commands should be able to spawn pager.

FWIW, I've sent https://github.com/jj-vcs/jj/pull/5999 to simplify the prompt handling.

calebdw:
@yuja, is this good or did you want to wait until #5999 is merged and then have this updated?

cli/src/commands/config/mod.rs
```diff
@@ -81,19 +80,33 @@ impl ConfigLevelArgs {
...
-                )));
+                let mut choices = vec![];
+                let mut formatter = ui.stderr_formatter();
+                for (i, file) in files.iter().enumerate() {
+                    writeln!(formatter, "{}: {}", i + 1, file.path().display())?;
+                    choices.push((i + 1).to_string());
+                }
+                drop(formatter);
+                let choice =
+                    ui.prompt_choice("Choose a config file (default 1)", &choices, Some("1"));
```
yuja, line 100: nit: just return error here by `ui.prompt_choice(..)?`. `io::Error` should have a valid message on EOF, and there are other kinds of failures.

cli/src/commands/config/mod.rs
```diff
@@ -81,19 +80,33 @@ impl ConfigLevelArgs {
...
+                match choice {
+                    Ok(choice) => {
+                        return Ok(files
+                            .get(choice.parse::<usize>().unwrap() - 1)
+                            .unwrap()
```
yuja, line 105: nit: `files[choice.parse..]` also works

cli/tests/test_config_command.rs
```diff
@@ -592,11 +592,10 @@ fn test_config_set_for_user_directory() {
...
     ------- stderr -------
-    Error: Cannot determine config file to edit:
-      $TEST_ENV/config/config0001.toml
-      $TEST_ENV/config/config0002.toml
+    1: $TEST_ENV/config/config0001.toml
+    2: $TEST_ENV/config/config0002.toml
+    Choose a config file (default 1): 1
     [EOF]
-    [exit status: 1]
     ");
```
yuja, line 599: nit: Can you insert snapshots of these config files to ensure that the first file is updated?

yuja (approved):
> is this good or did you want to wait until https://github.com/jj-vcs/jj/pull/5999 is merged and then have this updated?

Either way seems good. I can rebase my patch if needed.

cli/tests/test_config_command.rs
calebdw: Done!

PR #5997, opened 2025-03-12
git: use translated remote symbols in export_refs(), refactor import_refs()

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -16,7 +16,6 @@
 
 use std::borrow::Borrow;
 use std::borrow::Cow;
-use std::collections::BTreeMap;
 use std::collections::HashMap;
 #[cfg(feature = "git2")]
 use std::collections::HashSet;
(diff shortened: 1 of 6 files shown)


lib/src/git.rs
```diff
@@ -826,6 +833,13 @@ pub fn export_some_refs(
     mut_repo: &mut MutableRepo,
     git_ref_filter: impl Fn(GitRefKind, RemoteRefSymbol<'_>) -> bool,
 ) -> Result<Vec<FailedRefExport>, GitExportError> {
+    fn get<'a, V>(map: &'a [(RefName, V)], key: &RefName) -> Option<&'a V> {
```
ilyagr: I'd change the argument name here to `sorted_map`. It's obvious from the code, but editor code hints when mousing over invocations of this functions won't show that.

I'm starting to wonder if we should have a `SortedVec` wrapper type that could inidcate this in type names. (I guess creating it would do the `sort_unstable_by`; I haven't fully thought it through, and perhaps there's already a crate available).

lib/src/git.rs
```diff
@@ -802,9 +802,16 @@ pub enum FailedRefExportReason {
 
 #[derive(Debug)]
 struct RefsToExport {
-    branches_to_update: BTreeMap<RefName, (Option<gix::ObjectId>, gix::ObjectId)>,
-    branches_to_delete: BTreeMap<RefName, gix::ObjectId>,
-    failed_branches: HashMap<RefName, FailedRefExportReason>,
+    /// Remote bookmark `(ref_name, (old_oid, new_oid))`s to update, sorted by
+    /// `ref_name`.
+    branches_to_update: Vec<(RefName, (Option<gix::ObjectId>, gix::ObjectId))>,
```
ilyagr: Perhaps we should add _sorted to the names?

ilyagr (commented):
I only have a couple very minor comments so far.

I hadn't realized `RemoteRefSymbolBuf` existed, nice!

This looks fine at a glance, but is slow to review for me. Please ping me if I forget and nobody else gets to it.

lib/src/git.rs
yuja: I wouldn't because this is internal type, and we would have to rename them back and force when the content becomes sorted/unsorted. For external types (`Vec<FailedRefExport>` and `GitImportStats`), the important property is that lists have stable order.

lib/src/git.rs
yuja: I can rename it, but I think the debug assertion is good enough to describe the precondition.

Regarding `Sorted` wrapper, it might make sense to use something like that at public API boundary. I sometimes want to express that this `impl Iterator` is sorted, but that's not easy.

martinvonz:
@ilyagr: I think this is waiting for your re-review. Do you think you'll find time for it again soon? Otherwise I can take it.

ilyagr:
You are welcome to take it! I might get a chance, but I don't want to overpromise.

martinvonz (approved):

ilyagr:
Thanks! Indeed, both comments were meant to be optional, sorry I didn't make it clear.

PR #5999, opened 2025-03-12
cli: print prompt candidates to stderr, simplify choice handling

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/movement_util.rs
+++ b/cli/src/movement_util.rs
@@ -191,17 +191,17 @@ fn get_target_commit(
 }
 
 fn choose_commit<'a>(
-    ui: &mut Ui,
+    ui: &Ui,
     workspace_command: &WorkspaceCommandHelper,
     direction: Direction,
     commits: &'a [Commit],
 ) -> Result<&'a Commit, CommandError> {
     writeln!(
-        ui.stdout(),
+        ui.stderr(),
         "ambiguous {} commit, choose one to target:",
         direction.cmd()
     )?;
-    let mut formatter = ui.stdout_formatter();
+    let mut formatter = ui.stderr_formatter();
     let template = workspace_command.commit_summary_template();
     let mut choices: Vec<String> = Default::default();
     for (i, commit) in commits.iter().enumerate() {
--- a/cli/src/movement_util.rs
+++ b/cli/src/movement_util.rs
@@ -214,16 +214,14 @@ fn choose_commit<'a>(
     choices.push("q".to_string());
     drop(formatter);
 
-    let choice = ui.prompt_choice(
+    let index = ui.prompt_choice(
         "enter the index of the commit you want to target",
         &choices,
         None,
     )?;
-    if choice == "q" {
-        return Err(user_error("ambiguous target commit"));
-    }
-
-    Ok(&commits[choice.parse::<usize>().unwrap() - 1])
+    commits
+        .get(index)
+        .ok_or_else(|| user_error("ambiguous target commit"))
 }
 
 pub(crate) fn move_to_commit(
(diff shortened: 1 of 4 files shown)


cli/src/movement_util.rs
```diff
@@ -214,16 +214,14 @@ fn choose_commit<'a>(
     choices.push("q".to_string());
     drop(formatter);
 
-    let choice = ui.prompt_choice(
+    let index = ui.prompt_choice(
         "enter the index of the commit you want to target",
         &choices,
         None,
     )?;
-    if choice == "q" {
```
calebdw, line 222: I don't think the line above is valid anymore?
```rs
    writeln!(formatter, "q: quit the prompt")?;
```

Perhaps remove or change `q` to `^C`?

cli/src/movement_util.rs
```diff
@@ -214,16 +214,14 @@ fn choose_commit<'a>(
     choices.push("q".to_string());
     drop(formatter);
 
-    let choice = ui.prompt_choice(
+    let index = ui.prompt_choice(
         "enter the index of the commit you want to target",
         &choices,
```
calebdw, line 219: Would it make sense for `prompt_choice` to print the index with the option so we don't have to do that everywhere?

yuja: I think a better option is to add `ui.prompt_choice_with(.., |choice: &str| choice == ...)`. In `jj next`/`prev`, it's nice if the commit can be selected by short change ID, for example.

cli/src/movement_util.rs
yuja: No, `q` still works because `commits[q_index]` doesn't exist. I personally don't think `q` is needed, though. `^D` should always work.

calebdw: I agree with that, no need to add special handling when the terminal behaviour works

cli/src/ui.rs
```diff
@@ -590,29 +590,35 @@ impl Ui {
...
 
     /// Repeat the given prompt until the input is one of the specified choices.
+    /// Returns the index of the choice.
     pub fn prompt_choice(
         &self,
         prompt: &str,
         choices: &[impl AsRef<str>],
-        default: Option<&str>,
-    ) -> io::Result<String> {
+        default_index: Option<usize>,
```
ilyagr, line 598: Perhaps this should just be `usize`? Or do you think the behavior if this is `None` and `Self::can_prompt() == false` is reasonable?

ilyagr (approved):

cli/src/ui.rs
ilyagr: Ah, I realized we do have a use-case.

cli/src/ui.rs
```diff
@@ -626,14 +632,14 @@ impl Ui {
             Some(false) => "(yN)",
             None => "(yn)",
         };
-        let default_choice = default.map(|c| if c { "Y" } else { "N" });
+        let default_index = default.map(|c| if c { 0 } else { 1 });
 
-        let choice = self.prompt_choice(
+        let index = self.prompt_choice(
             &format!("{prompt} {default_str}"),
             &["y", "n", "yes", "no", "Yes", "No", "YES", "NO"],
```
ilyagr, line 639: (No action needed): This code seems confusing to me. Perhaps there should be a map from strings to a generic type (probably an enum) provided by the user. But this seems unrelated to your PR.

cli/src/ui.rs
yuja: `prompt_choice()` should fail in that case.

cli/src/ui.rs
yuja: I don't like it either. Maybe we can add `.prompt_choice_with()` later. It's weird that uppercase `"Y"` isn't accepted, for example.

PR #6008, opened 2025-03-12
git: tolerate unknown `git.fetch` remotes if others are available

JacobHayes:
If `git.fetch` contains remotes that are not available, we currently error even if other remotes are available. For common fork workflows with separate `upstream` and `origin` remotes (for example), this requires a user to either set both remotes in their user config and override single-remote repos or set only one in their user config and override all multi-remote repos to fetch from `upstream` (or both).

This change updates fetching to only *warn* about unknown remotes **if** other remotes are available. If none of the configured remotes are available, an error is still raised as before.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/fetch.rs
+++ b/cli/src/commands/git/fetch.rs
@@ -98,36 +98,26 @@ pub fn cmd_git_fetch(
     let all_remotes = git::get_all_remote_names(workspace_command.repo().store())?;
 
     let mut matching_remotes = HashSet::new();
-    let mut unmatched_patterns = Vec::new();
     for pattern in remote_patterns {
         let remotes = all_remotes
             .iter()
             .filter(|r| pattern.matches(r.as_str()))
             .collect_vec();
         if remotes.is_empty() {
-            unmatched_patterns.push(pattern);
+            writeln!(ui.warning_default(), "No git remotes matching '{pattern}'")?;
         } else {
             matching_remotes.extend(remotes);
         }
     }
 
-    match &unmatched_patterns[..] {
-        [] => {} // Everything matched, all good
-        [pattern] if pattern.is_exact() => {
-            return Err(user_error(format!("No git remote named '{pattern}'")))
-        }
-        patterns => {
-            return Err(user_error(format!(
-                "No matching git remotes for patterns: {}",
-                patterns.iter().join(", ")
-            )))
-        }
+    if matching_remotes.is_empty() {
+        return Err(user_error("No git remotes to push"));
     }
 
-    let remotes = all_remotes
+    let remotes = matching_remotes
         .iter()
-        .filter(|r| matching_remotes.contains(r))
         .map(|r| r.as_ref())
+        .sorted()
         .collect_vec();
 
     #[cfg(feature = "git2")]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Added `ui.streampager.show-ruler` setting to configure whether the ruler should be
   shown when the builtin pager starts up.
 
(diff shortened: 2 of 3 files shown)


cli/src/commands/git/fetch.rs
```diff
@@ -108,16 +108,25 @@ pub fn cmd_git_fetch(
...
-        [] => {} // Everything matched, all good
-        [pattern] if pattern.is_exact() => {
-            return Err(user_error(format!("No git remote named '{pattern}'")))
-        }
-        patterns => {
-            return Err(user_error(format!(
-                "No matching git remotes for patterns: {}",
-                patterns.iter().join(", ")
-            )))
+    match (!matching_remotes.is_empty(), &unmatched_patterns[..]) {
```
JacobHayes, line 111: Still fairly new to rust - open to suggestions on (more) idiomatic patterns here.

cli/src/commands/git/fetch.rs
yuja, line 1: I feel it's a bit weird that unknown remote names are allowed, but I don't have a strong feeling. We already have to allow invalid branch/bookmark names because we can't easily map it to an error.

FWIW, you can also use `git.fetch = ["regex:^(upstream|origin)$"]` to push to matching remotes if any. It won't error out if there are at least one remotes.

cli/src/commands/git/fetch.rs
```diff
@@ -108,16 +108,25 @@ pub fn cmd_git_fetch(
...
+                        "No matching git remotes for patterns: '{}'",
+                        patterns.iter().join("', '")
+                    )
+                }
+            };
+            if has_matches {
+                writeln!(ui.warning_default(), "{msg}; continuing with other remotes")?;
+            } else {
+                return Err(user_error(msg));
+            }
```
yuja, line 129: Since unmatched pattern is no longer a hard error, it's probably better to move the warning to the previous loop, and print error/status message if no match found:
```rs
    for pattern in remote_patterns {
        ...
        if remotes.is_empty() {
            // "No matching remotes" warning
        } else {
            matching_remotes.extend(remotes);
        }
    }

    if matching_remotes.is_empty() {
        // "No remotes to push"
        // I'm not sure if this should be an error because pushing to
        // 0 remotes can be considered a valid operation.
    }

Comment (JacobHayes) [reply cli/src/commands/git/fetch.rs:129]: Yeah, this is probably quite a bit nicer - I'll take a stab at that.

> // I'm not sure if this should be an error because pushing to
> // 0 remotes can be considered a valid operation.

I only considered it because the current behavior is to error in this case but fair point, if you set:
```toml
[git]
fetch = []
```
and run `jj git fetch`, there is no warning about empty remotes, only a branch warning:
```
Warning: No branch matching `*` found on any specified/configured remote
Nothing changed.
```

JacobHayes: FYI: I kept it an error if there are no matching remotes. This avoids the `Warning: No branch matching '*' found on any specified/configured remote`, which would otherwise print after the more relevant `No git remotes to push.` warning.

cli/src/commands/git/fetch.rs
```diff
@@ -98,37 +98,23 @@ pub fn cmd_git_fetch(
...
+    if matching_remotes.is_empty() {
+        return Err(user_error("No git remotes to push"));
     }
 
-    let remotes = all_remotes
-        .iter()
-        .filter(|r| matching_remotes.contains(r))
-        .map(|r| r.as_ref())
-        .collect_vec();
+    let remotes = matching_remotes.iter().map(|r| r.as_ref()).collect_vec();
```
yuja, line 117: `matching_remotes` doesn't maintain stable order. We should either filter `all_remotes` or sort the result.

JacobHayes:
Thanks for the guidance @yuja!

Issue #5978, opened 2025-03-10
`jj git push` thinks the ref moved on remote unexpectedly, but `jj git fetch` does nothing


nachtjasmin:
## Description

It's actually hard to pinpoint the exact problem, as it hasn't occurred yet. If I had to assume it's related to the Git subprocess spawning and the fact that this was once a repository with git-lfs in use.

As a consequence, I'm not sure whether those steps are even the correct ones for reproducibility. (I have stopped touching the repository right now in order to assist with the debugging, even if I know how to workaround it.)

I'll try to create a repro and if wanted, I can compress the local Git state and send it to one of you if it helps with debugging.

**Note:** This was converted to a co-located Git repository once, following the [guide in the documentation](https://jj-vcs.github.io/jj/latest/git-compatibility/#converting-a-repo-into-a-co-located-repo).

## Steps to Reproduce the Problem (wild guess!)

1. Clone a repository from ~~git.sr.ht~~ (edit: it is actually a bare repo on a remote host) that once used Git LFS. (TODO: build repro)
2. Add a new change, change anything you want.
3. Move `main` to the new change, push it. It should now be immutable.
4. Create a new change again and squash it into the previous change `jj squash --ignore-immutable`.
5. Move the bookmark again (should happen automatically) and try to push it.

## Expected Behavior

jj should be able to push it.

## Actual Behavior

`jj git push -b main` writes:

```
remote: error: denying non-fast-forward refs/heads/main (you should pull first)
Error: Refusing to push a bookmark that unexpectedly moved on the remote. Affected refs: refs/heads/main
Hint: Try fetching from the remote, then make the bookmark point to where you want it to be, and push again.
```

However, and here's why this is a bug, the output of `jj git fetch --all-remotes`:

```
Nothing changed.
```

I can repeat this as often as I want, it'll always end up in the same state.

[jj-git-fetch-debug-logs.log](https://github.com/user-attachments/files/19171074/jj-git-fetch-debug-logs.log)
[jj-git-push-debug-logs.log](https://github.com/user-attachments/files/19171075/jj-git-push-debug-logs.log)

## Specifications

- Platform: Fedora Linux 41
- Version: jj 0.27.0-6ce7a77da5a18343f4f3effef49b77428e43bc74, git version 2.48.1

ilyagr:
It would help if you could debug:

- where the ref is on the remote
- what `jj git branch --all main` reports
[...]

PR #4119, opened 2024-07-18
lib conflicts: minor bugfix + tests

ilyagr:
This is the uncontroversial part of https://github.com/martinvonz/jj/pull/4088.

Cc @yuja

PR #6009, opened 2025-03-13
git: properly detect remote refusal on push

bsdinis:
When we push a ref to a git remote, we use `--force-with-lease`
Our understanding was that this could fail iff the expected location of
the ref on the remote was not that of the local tracking ref

However, if the ref is from a protected branch (e.g., main) it will be
rejected by the remote for a different reason.

This commit solves this, by detecting this difference.

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_git_push.rs
+++ b/cli/tests/test_git_push.rs
@@ -2295,6 +2295,64 @@ fn test_git_push_sign_on_push() {
     ");
 }
 
+#[test]
+fn test_git_push_rejected_by_remote() {
+    let (test_env, workspace_root) = set_up();
+    // show repo state
+    insta::assert_snapshot!(get_bookmark_output(&test_env, &workspace_root), @r"
+    bookmark1: xtvrqkyv d13ecdbd (empty) description 1
+      @origin: xtvrqkyv d13ecdbd (empty) description 1
+    bookmark2: rlzusymt 8476341e (empty) description 2
+      @origin: rlzusymt 8476341e (empty) description 2
+    [EOF]
+    ");
+
+    // create a hook on the remote that prevents pushing
+    let hook_path = test_env
+        .env_root()
+        .join("origin")
+        .join(".jj")
+        .join("repo")
+        .join("store")
+        .join("git")
+        .join("hooks")
+        .join("update");
+
+    std::fs::write(&hook_path, "#!/bin/sh\nexit 1").unwrap();
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+
+        std::fs::set_permissions(&hook_path, std::fs::Permissions::from_mode(0o700)).unwrap();
+    }
+
+    // create new commit on top of bookmark1
+    test_env
+        .run_jj_in(&workspace_root, ["new", "bookmark1"])
+        .success();
+    std::fs::write(workspace_root.join("file"), "file").unwrap();
+    test_env
+        .run_jj_in(&workspace_root, ["describe", "-m=update"])
+        .success();
+
+    // update bookmark
+    test_env
+        .run_jj_in(&workspace_root, ["bookmark", "move", "bookmark1"])
+        .success();
+
+    // push bookmark
+    let output = test_env.run_jj_in(&workspace_root, ["git", "push"]);
+    insta::assert_snapshot!(output, @r#"
+    ------- stderr -------
+    Changes to push to origin:
+      Move forward bookmark bookmark1 from d13ecdbda2a2 to dd5c09b30f9f
+    remote: error: hook declined to update refs/heads/bookmark1        
+    Error: Remote rejected the update of some refs (do you have permission to push to ["refs/heads/bookmark1"]?)
+    [EOF]
+    [exit status: 1]
+    "#);
+}
+
 #[must_use]
 fn get_bookmark_output(test_env: &TestEnvironment, repo_path: &Path) -> CommandOutput {
     // --quiet to suppress deleted bookmarks hint
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -2441,15 +2441,14 @@ fn subprocess_push_refs(
         .map(|full_refspec| RefToPush::new(full_refspec, qualified_remote_refs_expected_locations))
         .collect();
 
-    let (failed_ref_matches, successful_pushes) =
-        git_ctx.spawn_push(remote_name, &refs_to_push, &mut callbacks)?;
+    let push_stats = git_ctx.spawn_push(remote_name, &refs_to_push, &mut callbacks)?;
 
-    for remote_ref in successful_pushes {
+    for remote_ref in push_stats.pushed {
         remaining_remote_refs.remove(remote_ref.as_str());
     }
 
-    if !failed_ref_matches.is_empty() {
-        let mut refs_in_unexpected_locations = failed_ref_matches;
+    if !push_stats.rejected.is_empty() {
+        let mut refs_in_unexpected_locations = push_stats.rejected;
         refs_in_unexpected_locations.sort();
         Err(GitPushError::RefInUnexpectedLocation(
             refs_in_unexpected_locations,
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -2404,9 +2404,7 @@ fn git2_push_refs(
         if remaining_remote_refs.is_empty() {
             Ok(())
         } else {
-            // TODO: this is probably dead code right now
-            // The only way this would happen is if a push fails from some
-            // other reason other than a lease failing (which our tests don't cover)
+            // remote rejected refs because of remote config (e.g., no push on main)
             Err(GitPushError::RefUpdateRejected(
                 remaining_remote_refs
                     .iter()
@@ -2431,11 +2429,6 @@ fn subprocess_push_refs(
         return Err(GitPushError::NoSuchRemote(remote_name.to_owned()));
     }
 
-    let mut remaining_remote_refs: HashSet<_> = qualified_remote_refs_expected_locations
-        .keys()
-        .copied()
-        .collect();
-
     let refs_to_push: Vec<RefToPush> = refspecs
         .iter()
         .map(|full_refspec| RefToPush::new(full_refspec, qualified_remote_refs_expected_locations))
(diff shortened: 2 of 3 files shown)


lib/src/git.rs
```diff
@@ -2453,19 +2453,14 @@ fn subprocess_push_refs(
...
-        // The only way this would happen is if a push fails from some
-        // other reason other than a lease failing (which our tests don't cover)
-        Err(GitPushError::RefUpdateRejected(
-            remaining_remote_refs
-                .iter()
-                .sorted()
-                .map(|name| name.to_string())
-                .collect(),
-        ))
+        assert!(remaining_remote_refs.is_empty());
```
bsdinis, line 2462: Not sure if this is the best move.
By all rights, this better be true, but it's implicitely depending on `git` behaving as we expect it.

lib/src/git_subprocess.rs
```diff
@@ -428,8 +427,7 @@ fn parse_ref_pushes(stdout: &[u8]) -> Result<GitPushResult, GitSubprocessError>
         let flag = it.next().ok_or_else(create_error)?;
         let reference = it.next().ok_or_else(create_error)?;
         // we capture the remaining elements to ensure the line is well formed
-        let _summary_or_comment = it.next().ok_or_else(create_error)?;
-        let _comment_opt = it.next();
+        let _summary = it.next().ok_or_else(create_error)?;
```
yuja: nit: we can now use `.collect_tuple()` or `.collect_array()`?

lib/src/git_subprocess.rs
```diff
@@ -384,7 +384,7 @@ fn parse_git_remote_show_default_branch(
 }
 
 // git-push porcelain has the following format (per line)
-// `<flag>\t<from>:<to>\t<summary>\t(<reason>)`
+// `<flag>\t<from>:<to>\t<summary> (<reason>)`
```
yuja, line 390: nit: we have another instance of this comment at line ~420. remove one of them?

lib/src/git.rs
yuja: I wouldn't use assertion because of the reason you described. Since we have to trust the git output anyway, I would remove the `remaining_remote_refs` computation at all.

yuja (commented):
Thanks.

lib/src/git.rs
```diff
@@ -2431,41 +2431,26 @@ fn subprocess_push_refs(
...
         refs_in_unexpected_locations.sort();
         Err(GitPushError::RefInUnexpectedLocation(
             refs_in_unexpected_locations,
         ))
-    } else if remaining_remote_refs.is_empty() {
-        Ok(())
+    } else if !push_result.remote_rejected.is_empty() {
+        let mut rejected_refs = push_result.remote_rejected;
+        rejected_refs.sort();
+        // remote rejected refs because of remote config (e.g., no push on main)
```
ilyagr, line 2450: Could you change the comment similarly in the libgit2 flow?

lib/src/git_subprocess.rs
```diff
@@ -454,7 +456,11 @@ fn parse_ref_pushes(stdout: &[u8]) -> Result<GitPushStats, GitSubprocessError> {
             }
             // ! for a ref that was rejected or failed to push; and
             b"!" => {
-                push_result.rejected.push(reference);
+                if summary.starts_with_str("[remote rejected]") {
+                    push_result.remote_rejected.push(reference);
```
ilyagr, line 460: I wonder if it makes sense to send the message to `tracing::debug` or `tracing::info` in case of failure, including the summary (which you are currently dropping)

cli/tests/test_git_push.rs
```diff
@@ -2295,6 +2295,71 @@ fn test_git_push_sign_on_push() {
     ");
 }
 
+// cannot set executable file in windows
+#[cfg(unix)]
```
ilyagr, line 2299: (Optional) Reading https://stackoverflow.com/questions/18277429/executing-git-hooks-on-windows, I think the hook might work on Windows as long ad you add `#!/bin/sh` to the beginning of it. So, only the setting of the executable bit has to be unix-specific.

ilyagr:
Thank you very much! At a glance, this LGTM. I'll let Yuya approve, or I can look it over again and approve (but probably not today).

lib/src/git.rs
ilyagr, line 1: Perhaps mention https://github.com/jj-vcs/jj/issues/5978 in the commit description?

cli/tests/test_git_push.rs
ilyagr: On second thought, this might be easier to explore in a separate PR. I think it's worth doing in case somebody who uses Windows wants to debug this code, but of course it's less important than having the test on unix.

bsdinis: Mainly my problem here was actually getting an executable for windows at all, wasn't sure how to get that in. I think it can go on this PR

cli/tests/test_git_push.rs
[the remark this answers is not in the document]
bsdinis: I would prefer it if we didn't have a commit on main that purposefully fails, as I believe that is a goal of having the linear history.

I did check and it does fail without the fix

lib/src/git.rs
bsdinis: So, I used to but it seems actually unrelated. See https://github.com/jj-vcs/jj/issues/5978#issuecomment-2722368322

They're getting an error due to a failure to fast forward

cli/tests/test_git_push.rs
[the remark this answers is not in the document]
ilyagr: I meant modifying the test, so that it passes but demonstrates the wrong behavior (with a comment explaining that this behavior needs to be changed). Then, the commit with the fix would change the test to show the correct behavior (and would also pass).

cli/tests/test_git_push.rs
ilyagr: The file wouldn't be executable on Windows, but apparently Git-for-Windows knows the `#!/bin/sh` shebang and executes hooks with that shebang using its own version of bash.

lib/src/git_subprocess.rs
bsdinis: I think this has merit, and have been considering it seperately. 

My view here is that we should actually `debug!` `git`'s output more generally, as it could give interesting feedback on what the errors are.
Because that's a bit more intrusive, I think it should probably go on a separate PR.

cli/tests/test_git_push.rs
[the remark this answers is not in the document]
emilazy: (One trick I like to do is adding an expected‐failing test, and then just removing the expected‐fail annotation in the fix commit.)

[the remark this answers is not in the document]
bsdinis: how do you add an exepected-failing test?

cli/tests/test_git_push.rs
yuja: I think the current state is good, but we'll need a better comment, something mentioning that the hook script is platform dependent.

lib/src/git_subprocess.rs
ilyagr: I think having some rudimentary version of this would make debugging things like #5978 easier. We could change it later to some more general solution.

If this code had the same tracing as the libgit2 code, we'd have known what was going on faster.

> My view here is that we should actually debug! git's output more generally

That might be nice too. I'm not sure whether it might be overly chatty to print git's output in cases of success. Or perhaps we could print one with `warning` priority and the other with `info`/`debug`.

lib/src/git.rs
bsdinis: done

lib/src/git_subprocess.rs
bsdinis: Yeah, I think there is a lot of nuance about what we'd actually want to put there (hence why I think a separate PR would be ideal)

I'm happy to commit to pushing one out in a couple of days

cli/tests/test_git_push.rs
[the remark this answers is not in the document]
ilyagr: I wouldn't do that, actually, unless you really have to. It's much more informative to show the actual behavior before the fix than to show what the behavior is not.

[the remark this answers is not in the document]
ilyagr: But I think `#[should_panic]` is the answer: <https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#testing-panics>.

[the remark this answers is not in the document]
bsdinis: what about the linear history being good at all times? Maybe I'm operating on a wrong assumption and that is a non-goal?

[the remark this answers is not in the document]
emilazy: > I wouldn't do that, actually, unless you really have to. It's much more informative to show the actual behavior before the fix than to show what the behavior is not.

I agree in this case. But sometimes the way I suggested can work out.

> what about the linear history being good at all times? Maybe I'm operating on a wrong assumption and that is a non-goal?

@ilyagr’s suggestion would keep tests passing on every commit. It’s just that the test would show the previous bad behaviour and how it changes into the new one. See e.g. how my remote management PR added tests showing `git2` behaviour and then changed them as I ported it to `gix`.

lib/src/git_subprocess.rs
ilyagr: It's fine to leave this to a separate PR. Whatever you and Yuya have in mind is probably good.

cli/tests/test_git_push.rs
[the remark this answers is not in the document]
ilyagr: Here's an example: https://github.com/jj-vcs/jj/pull/4119/commits/be32bb7ec161d1a1ac9c2f282ed5c8a286b0d0b8 adds the test and the following commit fixes it.

(Again, no need to worry for this PR, but I find this can be helpful in PRs fixing confusing bugs)

cli/tests/test_git_push.rs
bsdinis: managed to get it to run on Windows

cli/tests/test_git_push.rs
yuja, line 1: nit: can you squash the last test commit (or reorder as suggested by ilyagr)? It doesn't make much sense to add a small test by separate patch (unless we want to demonstrate the test failure.)

yuja (approved):

cli/tests/test_git_push.rs
bsdinis: done

cli/tests/test_git_push.rs
```diff
@@ -2295,6 +2295,83 @@ fn test_git_push_sign_on_push() {
...
+
+#[cfg(not(unix))]
+fn create_executable_file(path: impl AsRef<Path>) -> std::fs::File {
+    std::fs::OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(path)
+        .unwrap()
+}
```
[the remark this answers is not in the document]
yuja: Well, you can use `PermissionsExt` there. Grep some callers of `::set_permissions()`.

My point is that we don't have to set up everything at `open()` call.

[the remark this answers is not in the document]
bsdinis: gotcha, done

PR #6011, opened 2025-03-13
cli: config list: show origin of config values

### LINKED ISSUE #1047: Make `config list` output more user-friendly
Now that there's a basic `jj config list` command, we want to improve it to give more user-friendly output:
- [x] Hide implicit defaults and only show explicit config values (with `--include-defaults` option to override)
- [x] Instead of just sorting alphabetically, order output more logically to correspond to the config layers and their contents
- [x] #2827
- [x] #2828
- [x] Colorize output
- [ ] Add a `--show-sources` option to report where each config value came from (similar to git's `--show-origin` and hg's `--debug`)

calebdw:
Hello!

This adds the config origin to the end of the line for each config value in the list.
Options coming from files will show the file path.

image

Thanks!

Closes #1047

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/config/list.rs
+++ b/cli/src/commands/config/list.rs
@@ -44,7 +44,6 @@ pub struct ConfigListArgs {
     pub include_overridden: bool,
     #[command(flatten)]
     pub level: ConfigLevelArgs,
-    // TODO(#1047): Support --show-origin using StackedConfig.
     /// Render each variable using the given template
     ///
     /// The following keywords are available in the template expression:
@@ -52,6 +51,12 @@ pub struct ConfigListArgs {
     /// * `name: String`: Config name.
     /// * `value: ConfigValue`: Value to be formatted in TOML syntax.
     /// * `overridden: Boolean`: True if the value is shadowed by other.
+    /// * `source: String`: Source of the value.
+    /// * `path: String`: Path to the config file.
+    ///
+    /// Can be overridden by the `templates.config_list` setting. To
+    /// see a detailed config list, use the `builtin_config_list_detailed`
+    /// template.
     ///
     /// See [`jj help -k templates`] for more information.
     ///
@@ -73,12 +78,17 @@ pub fn cmd_config_list(
 ) -> Result<(), CommandError> {
     let template = {
         let language = config_template_language(command.settings());
-        let text = match &args.template {
-            Some(value) => value.to_owned(),
+        let template_string = match &args.template {
+            Some(value) => value.to_string(),
             None => command.settings().get_string("templates.config_list")?,
         };
         command
-            .parse_template(ui, &language, &text, GenericTemplateLanguage::wrap_self)?
+            .parse_template(
+                ui,
+                &language,
+                &template_string,
+                GenericTemplateLanguage::wrap_self,
+            )?
             .labeled("config_list")
     };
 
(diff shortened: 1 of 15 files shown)


calebdw:
@yuja, @martinvonz,

It turns out *conditionally* showing the origin is a little tricky with the template. Is there any opposition to just always showing the origin? I intentionally placed it at the end so it's easier to visually parse the key/value pairs

cli/src/commands/config/list.rs
```diff
@@ -129,6 +131,17 @@ fn config_template_language(
...
         Ok(L::wrap_config_value(out_property))
     });
+    language.add_keyword("origin", |self_property| {
+        let out_property = self_property.map(|annotated| {
+            let source = annotated.source;
+            if let Some(path) = &annotated.path {
+                format!("[{source}:{}]", path.to_string_lossy())
+            } else {
+                format!("[{source}]")
+            }
```
yuja, line 141: it's probably better to add separate template keywords (such as `source_type` and `source_path`)

bnjmnt4n:
Instead of adding a new `--show-origin` option, we could provide a different template which shows the origin, and point users to use `jj config list --template X`.

cli/src/config/templates.toml
```diff
@@ -33,8 +33,8 @@ separate(" ",
 
 config_list = '''
 if(overridden,
-  label("overridden", indent("# ", name ++ " = " ++ value)),
-  name ++ " = " ++ value,
+  label("overridden", indent("# ", name ++ " = " ++ value ++ " " ++ origin)),
+  name ++ " = " ++ value ++ " " ++ origin,
 ) ++ "\n"
```
yuja, line 38: The default template output is a valid TOML document. Maybe we can use comment syntax?

It might be also make sense to add `builtin_config_list_detailed` or something. We wouldn't want to see the source type/path by default.

calebdw:
> Instead of adding a new --show-origin option, we could provide a different template which shows the origin, and point users to use jj config list --template X.

That just doesn't feel as nice from a dx perspective...

cli/src/config/templates.toml
calebdw: > The default template output is a valid TOML document. Maybe we can use comment syntax?

Ah good point

cli/src/commands/config/list.rs
calebdw: I thought about it, but I considered the path to really be a "property" of the User and Repo ConfigSource enums---splitting these apart means you have an extra keyword to deal with and need to handle the empty case. I think combining them into a single `origin` keyword is a clean approach so I'd rather go with that unless you insist.

yuja: Let's say user wants to query config value and its file path, the most intuitive way is to use `-T'value ++ "\t" ++ source_path ++ "\n"'` or something. If `source` type/path was formatted in certain form, it would have to be split by string function. That seems a bit annoying.

cli/src/commands/config/list.rs
```diff
@@ -73,9 +75,13 @@ pub fn cmd_config_list(
 ) -> Result<(), CommandError> {
     let template = {
         let language = config_template_language(command.settings());
+        let default_template = match args.show_origin {
+            true => "templates.config_list_detailed",
+            false => "templates.config_list",
+        };
```
yuja, line 81: Naming aside, can't we use `-Tbuiltin_config_list_detailed`? That would require a bit of typing, but we can avoid inconsistency.

cli/src/config/templates.toml
```diff
@@ -31,12 +31,7 @@ separate(" ",
 ) ++ ": " ++ content
 '''
 
-config_list = '''
-if(overridden,
-  label("overridden", indent("# ", name ++ " = " ++ value)),
-  name ++ " = " ++ value,
-) ++ "\n"
-'''
+config.list = 'builtin_config_list'
```
martinvonz, line 34: was this changed from `config_list` to `config.list` on purpose?

calebdw: Yes, should I revert?

calebdw: I suppose it's better to keep the key the same in case folks have overridden it---reverted

martinvonz: We use underscores for `bookmark_list` and `file_annotate` above, so I think we should use that here too for consistency.

martinvonz (approved):
Thanks!

calebdw:
Thank you!

PR #6018, opened 2025-03-13
cli: change debug logging behaviour to whitelist `jj` crates 

bsdinis:
The current behaviour means (transitively) dependent crates' debug logs end up
in our logs when `--debug` is active.

The biggest offender here is `globset`, imported from the `ignore
crate`. These logs are extremely noisy and mostly irrelevant from our usecase.

This commit changes this behaviour to whitelist `jj` related debug logs,
while allowing more debugging to come from env vars

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -260,10 +260,15 @@ impl TracingSubscription {
     pub fn enable_debug_logging(&self) -> Result<(), CommandError> {
         self.reload_log_filter
             .modify(|filter| {
+                // The default is INFO.
+                // jj-lib and jj-cli are whitelisted for DEBUG logging.
+                // This ensures that other crates' logging doesn't show up by default.
                 *filter = tracing_subscriber::EnvFilter::builder()
-                    .with_default_directive(tracing::metadata::LevelFilter::DEBUG.into())
+                    .with_default_directive(tracing::metadata::LevelFilter::INFO.into())
                     .with_env_var(Self::ENV_VAR_NAME)
-                    .from_env_lossy();
+                    .from_env_lossy()
+                    .add_directive("jj_lib=debug".parse().unwrap())
+                    .add_directive("jj_cli=debug".parse().unwrap());
             })
             .map_err(|err| internal_error_with_message("failed to enable debug logging", err))?;
         tracing::info!("debug logging enabled");
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -470,9 +470,9 @@ want to adjust the `site_url` to something like `https://jjfan.github.io/jj`.
 
 You can print internal jj logs using `JJ_LOG`. It acts like the `RUST_LOG`
 environment variable, frequent in Rust codebases, and accepts one or more
-[directives]. You can also use the `--debug` global option that sets
-`debug` log level for all targets by default. `JJ_LOG` is still respected when
-using `--debug`.
+[directives]. You can also run `JJ_LOG=debug jj` to get `debug` level logs
+enabled for all targets. You can also use the `--debug` global option, which
+turns on `debug` log level for `jj-lib` and `jj-cli` only.
 
 [directives]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
 


emilazy:
Would there still be a way to override it with this?

emilazy:
As in, could you still specify that you want `globset` debug logs explicitly in `RUST_LOG`?

bsdinis:
> As in, could you still specify that you want `globset` debug logs explicitly in `RUST_LOG`?

Interesting. Unsure, but can check

martinvonz:
If you can't, it's probably still a good idea to do this? We can add our more targeted debug logging if necessary.

emilazy:
Well, an alternative would be to do `RUST_LOG=jj-lib=debug,jj-cli=debug`. That’s verbose, but we could have shorthand for it (probably in tandem with moving to `JJ_LOG`?). `globset` logs can be useful for debugging things like gitignore problems. I think it’d be weird to break the normal `RUST_LOG` semantics; we’d potentially have to keep naming crates as our dependencies add more debug logs when what we really want is just to show our own debug logs.

cli/src/cli_util.rs
```diff
@@ -265,7 +265,13 @@ impl TracingSubscription {
                 *filter = tracing_subscriber::EnvFilter::builder()
                     .with_default_directive(tracing::metadata::LevelFilter::DEBUG.into())
                     .with_env_var(Self::ENV_VAR_NAME)
-                    .from_env_lossy();
+                    .from_env_lossy()
+                    // deny globset debug traces
+                    .add_directive(
```
ilyagr, line 270: I wonder whether switching the order of this and `.from_env_lossy()` would make it possible to change this directive with an env var.

(Optional) It might also be easier to find a way to disable logging from the `ignore` crate instead of specializing to `globset`.

ilyagr (commented):
Thanks, this would be very nice!

emilazy:
Would a sensible alternative be to just make `jj --debug` only enable debug logging for `jj-lib` and `jj-cli`? Then the environment variable would still have the normal semantics, but we’d just suggest the flag.

ilyagr:
> Would a sensible alternative be to just make `jj --debug` only enable debug logging for `jj-lib` and `jj-cli`? Then the environment variable would still have the normal semantics, but we’d just suggest the flag.

That SGTM to me. We could add other crates on an as-needed basis. I don't remember ever using other crates' logging (that I remember), I'm not sure whether others have.

martinvonz:
I'd be happy if it could also enable it for our customizations at Google somehow but we can figure that out ourselves. I don't even know to enable logging for jj-lib and jj-cli. Maybe it'll be clear once I see how that's done.

cli/src/cli_util.rs
bsdinis: you can't `add_directive` before `from_env_lossy`

for comment no. 2 I think the new behaviour of whitelisting `jj` makes that point moot

emilazy:
> I'd be happy if it could also enable it for our customizations at Google somehow but we can figure that out ourselves. I don't even know to enable logging for jj-lib and jj-cli. Maybe it'll be clear once I see how that's done.

I don’t really know how Google’s extensions work, but maybe the backend trait could have a method for this?

bsdinis:
> I'd be happy if it could also enable it for our customizations at Google somehow but we can figure that out ourselves. I don't even know to enable logging for jj-lib and jj-cli. Maybe it'll be clear once I see how that's done.

The current commit does it.
Basically, you set the default log level to INFO and then whitelist the crates you need to emit `debug`

bsdinis:
Is there some other actionable here?

ilyagr:
Could you also edit the contributing.md section on JJ_LOG? You can just do the minimal edit to make it correct.

I tested this out and it works fine . I might wait a bit before approving so that others can take a look.

For the record, setting `JJ_LOG=debug` restores the previous behavior.

cli/src/cli_util.rs
```diff
@@ -260,10 +260,23 @@ impl TracingSubscription {
     pub fn enable_debug_logging(&self) -> Result<(), CommandError> {
         self.reload_log_filter
             .modify(|filter| {
+                // the default is INFO
+                // jj-lib and jj-cli are whitelisted for DEBUG logging
+                // this ensures that other crates logging doesn't come int by default
```
martinvonz, line 265: nits:
* Start lines 1 and 3 with uppercase. End all lines with period.
* crates => crates'
* int => into play (?), or rephrase it

cli/src/cli_util.rs
```diff
@@ -260,10 +260,23 @@ impl TracingSubscription {
...
                 *filter = tracing_subscriber::EnvFilter::builder()
-                    .with_default_directive(tracing::metadata::LevelFilter::DEBUG.into())
+                    .with_default_directive(tracing::metadata::LevelFilter::INFO.into())
```
emilazy, line 267: Is this the `tracing_subscriber` default? If so, we could drop this line, right?

emilazy: Ah, apparently there is no default.

bsdinis:
> Could you also edit the contributing.md section on JJ_LOG? You can just do the minimal edit to make it correct.
> 
> I tested this out and it works fine . I might wait a bit before approving so that others can take a look if they want.
> 
> For the record, setting `JJ_LOG=debug` restores the previous behavior.

I think I got it, but feel free to comment

docs/contributing.md
```diff
@@ -471,8 +471,8 @@ want to adjust the `site_url` to something like `https://jjfan.github.io/jj`.
 You can print internal jj logs using `JJ_LOG`. It acts like the `RUST_LOG`
 environment variable, frequent in Rust codebases, and accepts one or more
 [directives]. You can also use the `--debug` global option that sets
```
ilyagr, line 473: You could reorder this, e.g.:

> ... directives. For example, `JJ_LOG=debug jj` will run `jj` with `debug` log-level enabled for all targets. You can also use the global option `jj --debug`, which turns on debug logging just for the `jj-lib` and `jj-cli` targets.

cli/src/cli_util.rs
```diff
@@ -260,10 +260,23 @@ impl TracingSubscription {
     pub fn enable_debug_logging(&self) -> Result<(), CommandError> {
         self.reload_log_filter
             .modify(|filter| {
+                // The default is INFO.
+                // jj-lib and jj-cli are whitelisted for DEBUG logging.
+                // This ensures that other crates' logging show up by default.
```
ilyagr, line 265: I'm a bit confused by this last line of comments. I guess you meant INFO-level logging? Perhaps we can just delete it?

cli/src/cli_util.rs
```diff
@@ -260,10 +260,23 @@ impl TracingSubscription {
...
+                    .from_env_lossy()
+                    .add_directive(
+                        "jj_lib=debug"
+                            .parse()
+                            .expect("directive should be well formed"),
+                    )
+                    .add_directive(
+                        "jj_cli=debug"
+                            .parse()
+                            .expect("directive should be well formed"),
```
ilyagr, line 278: BTW, it's probably fine to just use `.unwrap()` here. I don't think this message helps much with either readability or debuggability.

ilyagr (approved):
Thank you!

cli/src/cli_util.rs
bsdinis: sorry, there is a typo, fixed it

yuja:
Does it make sense to bump our `tracing::debug!()` to `tracing::info!()`? Since our `tracing` messages are all for debugging, using `info()` should be okay.

bsdinis:
> Does it make sense to bump our `tracing::debug!()` to `tracing::info!()`? Since our `tracing` messages are all for debugging, using `info()` should be okay.

I think there is still a difference that makes sense. there is a fair bunch of `info!` that make sense. With `JJ_LOG` you still filter for `INFO`, I think

yuja:
Our choice of tracing level looks arbitrary, though? I think `"using credential_helper"` should be the same level as `"remote.download"` for example.

```
% rg '(debug|info)!' --sort=path
cli/src/cli_util.rs:269:        tracing::info!("debug logging enabled");
cli/src/commands/fix.rs:397:    tracing::debug!(?command, ?tool_input.repo_path, "spawning fix tool");
cli/src/commands/fix.rs:412:    tracing::debug!(?command, ?output.status, "fix tool exited:");
cli/src/description_util.rs:87:        tracing::info!(?cmd, "running editor");
cli/src/git_util.rs:157:                tracing::info!(path = ?key_path, "found ssh key");
cli/src/git_util.rs:163:        tracing::info!("no ssh key found");
cli/src/merge_tools/external.rs:254:    tracing::info!(?cmd, "Invoking the external merge tool:");
cli/src/merge_tools/external.rs:392:    tracing::info!(?cmd, "Invoking the external diff editor:");
cli/src/merge_tools/external.rs:443:    tracing::info!(?cmd, "Invoking the external diff generator:");
cli/src/merge_tools/external.rs:457:    tracing::info!(?cmd, ?exit_status, "The external diff generator exited:");
cli/src/ui.rs:76:        tracing::info!(?cmd, "spawning pager");
lib/src/default_index/store.rs:237:        tracing::info!(
lib/src/default_index/store.rs:277:        tracing::info!(
lib/src/fsmonitor.rs:184:            info!("Initializing Watchman filesystem monitor...");
lib/src/fsmonitor.rs:224:            info!("Querying Watchman for changed files...");
lib/src/fsmonitor.rs:268:            info!("Checking for an existing Watchman trigger...");
lib/src/fsmonitor.rs:282:            info!("Registering Watchman trigger...");
lib/src/fsmonitor.rs:305:            info!("Unregistering Watchman trigger...");
lib/src/git.rs:1995:        tracing::debug!("import_refs");
lib/src/git.rs:2141:    tracing::debug!("remote.download");
lib/src/git.rs:2143:    tracing::debug!("remote.prune");
lib/src/git.rs:2145:    tracing::debug!("remote.update_tips");
lib/src/git.rs:2152:    tracing::debug!("remote.disconnect");
lib/src/git.rs:2171:    tracing::debug!("remote.connect");
lib/src/git.rs:2182:    tracing::debug!("remote.default_branch");
lib/src/git.rs:2188:                tracing::debug!(default_branch = branch_name);
lib/src/git.rs:2250:    tracing::debug!(default_branch = default_branch);
lib/src/git.rs:2441:                        tracing::info!(
lib/src/git.rs:2450:                        tracing::info!(
lib/src/git.rs:2464:                        tracing::info!(
lib/src/git.rs:2658:                tracing::info!("using credential_helper");
lib/src/git.rs:2665:                        tracing::info!(username, "trying ssh_key_from_agent");
lib/src/git.rs:2684:                        tracing::info!(username, path = ?path, "trying ssh_key");
lib/src/git.rs:2694:                            tracing::info!(
lib/src/git.rs:2708:                        tracing::info!(username, "using userpass_plaintext");
lib/src/git.rs:2716:            tracing::info!("using default");
lib/src/git_backend.rs:391:        tracing::debug!(
lib/src/git_backend.rs:785:    tracing::info!(
lib/src/git_backend.rs:1181:            tracing::info!("unimported Git commit found");
lib/src/git_subprocess.rs:113:        tracing::debug!(cmd = ?git_cmd, "spawning a git subprocess");
lib/src/gpg_signing.rs:78:    tracing::info!(?command, "running GPG signing command");
lib/src/gpg_signing.rs:82:    tracing::info!(?command, ?output.status, "GPG signing command exited");
lib/src/gpg_signing.rs:95:    tracing::info!(?command, "running GPG signing command");
lib/src/gpg_signing.rs:99:    tracing::info!(?command, ?output.status, "GPG signing command exited");
lib/src/local_working_copy.rs:1063:            tracing::info!("not updating watchman clock because there are untracked files");
lib/src/simple_op_store.rs:297:        tracing::info!(
lib/src/ssh_signing.rs:71:    tracing::info!(?command, "running SSH signing command");
lib/src/ssh_signing.rs:75:    tracing::info!(?command, ?output.status, "SSH signing command exited");
lib/src/ssh_signing.rs:161:        tracing::info!(?command, "running SSH signing command");
lib/src/ssh_signing.rs:164:        tracing::info!(?command, ?output.status, "SSH signing command exited");
```

ilyagr:
Thinking out loud, switching everything to `info` would make this PR simpler (no need for exception for `jj_lib` and `jj_cli`), but at the cost of creating friction to ever using the `debug` loglevel in the future. I'm not sure that's good; even if we do switch to using `info` everywhere, I wouldn't mind this PR living in our Git history so that we know how to go back.

I don't feel strongly, but while the policy of "sure, it'd be `debug` in a library, but we make it `info` in our CLI" makes some sense, it's also a bit weird. It's even weirder for `jj_lib`, which is in theory a library that could be used as part of a larger application in the future.

I think my mild preference would be to just merge this for now, and then maybe change it in another PR.

yuja:
My suspicion is that our use of `debug!()` level might not be aligned with the standard, or at least with globset? I don't know if globset is the outlier, though.

ilyagr:
Yeah, it's definitely a bit of a mess at the moment. If you have an idea for what a good standard or organizing principle would be, it could be good to use it and document it. For me, at least, it's not obvious how to decide between `debug` and `info`.

If I had to choose, I'd put the "Invoking the external merge tool" logs as `info` because they relate to user config, and I'd put everything else as `debug`. However, this is not entirely principled: the path to `git` is user-configurable, but it doesn't seem worthy of `info` to me. Or, perhaps we print it once to `info` and then print logs from `git `invocations to `debug`? I'm not sure.

But, in my mind at least, figuring this out doesn't have to block this PR.

emilazy:
I think it's expected that debug logging for all crates is noisy even if everyone is doing everything right. I feel we're lucky that we have sufficiently few complex dependencies that it's only `globset` being noisy right now. So it seems like a good change anyway to say what we mean: that `--debug` is primarily for Jujutsu's own debug output.

yuja:
I meant jj's debug output might also become too verbose for `--debug` (if our tracing use is aligned), and it might make sense to use `JJ_LOG=debug` to see the full debug output.

emilazy:
I agree with that, but I don’t know what @yuja thinks.

yuja:
I would simply bump the level to `INFO`, but I don't have strong reason against this PR. We can decide that later.

PR #6027, opened 2025-03-14
sign: gpg: automatically use user email as signing key

calebdw:
If a signing key is not configured, the user's email will be used as the signing key. This aligns with `git`'s behavior and allows the users to not have to specify the key in their configs given that they have a key associated with their email.

Thanks!

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj sign` will automatically use the gpg key associated with the author's email
+  in the absence of a `signing.key` configuration.
+
 * Multiple user configs are now supported and are loaded in the following precedence order:
   - `$HOME/.jjconfig.toml`
   - `$XDG_CONFIG_HOME/jj/config.toml`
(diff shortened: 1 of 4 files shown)


yuja:
Maybe the fallback should be handled by `GpgBackend::sign()`? I don't know if it makes sense to ignore the default of `gpg`, though.

cc @necauqua

martinvonz (approved):

calebdw:
> Maybe the fallback should be handled by GpgBackend::sign()?

That's what i tried to do first, but `sign()` has no knowledge of the SignSettings

> I don't know if it makes sense to ignore the default of gpg, though.

Typically,  your signature should be associated with the email address you're using. The gpg default is just the first one, which may not be correct

yuja:
> That's what i tried to do first, but sign() has no knowledge of the SignSettings

The default key can be captured at `GpgBackend::from_settings()`.

calebdw:
@martinvonz, @yuja,


I made changes (I like how things turned out, thanks @yuja), and ci is passing. Are we good to merge?

yuja (approved):
Looks good, thanks!

PR #6031, opened 2025-03-14
cli: annotate: add --domain

bryceberger:
There's been talk about adding this, most recently https://github.com/jj-vcs/jj/pull/5967#discussion_r1991345773.

The main question is what to show in the "diff not contained in domain" case. Some possibilities:

- Root commit
  - (+) Root commit is always empty, no confusion between "blamed because actually changed" and "blamed because oldest ancestor"
  - (-) Not immediately obvious
- Oldest commit in domain
  - (-) Confusion about blame reason as mentioned above
- Other special marker
  - (+) immediately obvious
  - (-) changes `AnnotationLine` type

Currently, the implementation chooses the root commit. I'm partial to either that or the "other special marker" solution --- I do _not_ like the idea of using the farthest ancestor, since it would be both surprising (when the commit does not actually touch the line) and have no way to tell inside the `file annotate` output.

WIP due to missing tests, documentation, etc.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/file/annotate.rs
+++ b/cli/src/commands/file/annotate.rs
@@ -50,6 +50,9 @@ pub(crate) struct FileAnnotateArgs {
         add = ArgValueCandidates::new(complete::all_revisions)
     )]
     revision: Option<RevisionArg>,
+    /// An optional domain to restrict the query to.
+    #[arg(long, value_name = "REVSET", add = ArgValueCandidates::new(complete::all_revisions))]
+    domain: Option<RevisionArg>,
     /// Render each line using the given template
     ///
     /// All 0-argument methods of the [`AnnotationLine` type] are available as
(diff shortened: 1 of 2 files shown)


cli/src/commands/file/annotate.rs
```diff
@@ -76,8 +79,22 @@ pub(crate) fn cmd_file_annotate(
...
     let starting_commit = workspace_command
         .resolve_single_rev(ui, args.revision.as_ref().unwrap_or(&RevisionArg::AT))?;
+
+    if domain
+        .intersection(&RevsetExpression::commit(starting_commit.id().clone()))
+        .evaluate(workspace_command.repo().as_ref())?
+        .is_empty()
+    {
+        return Err(user_error("Starting commit is not contained within domain"));
+    }
```
yuja, line 96: It's probably better to reject non-contiguous range. (See `cmd_diff()` for how we can test that.)

https://github.com/jj-vcs/jj/blob/09d6d4871834cb927d78c4e268bb3508eb673ead/lib/src/annotate.rs#L209

yuja (commented):
> The main question is what to show in the "diff not contained in domain" case. 

I think `git blame` shows the ambiguous result with `^` prefix. I think we can use `-` suffix meaning the line would be modified by ancestor revisions.

bryceberger:
`commit-` feels like it's saying "the parent(s) of" instead of "one of the ancestors of", but maybe `::commit` and `commit::`?

Regardless of specifics, I assume this would be implemented as something like `self.match_kind() -> "exact|ancestor|descendant"` on the `AnnotationLine` template? Though I'm not sure if calculating whether ancestor or descendant would be possible without evaluating commits outside the domain, which potentially nullifies the speed benefits

yuja:
> `commit-` feels like it's saying "the parent(s) of" instead of "one of the ancestors of", but maybe `::commit` and `commit::`?

Technically more correct, but I think single-char prefix/suffix is better in this context.

> Regardless of specifics, I assume this would be implemented as something like `self.match_kind() -> "exact|ancestor|descendant"` on the `AnnotationLine` template? Though I'm not sure if calculating whether ancestor or descendant would be possible without evaluating commits outside the domain, which potentially nullifies the speed benefits

It can be computed within annotation function. I'll send a patch.

PR #6032, opened 2025-03-14
cleanup: always import extension traits anonymously

martinvonz:
# Checklist

If applicable:


ilyagr:
Perhaps we should enable <https://rust-lang.github.io/rust-clippy/master/index.html#unused_trait_names>?

martinvonz:
Seems like a good idea to import those other traits anonymously too (not just extension traits). I can close this if you send yours instead.

ilyagr:
I'm not actually sure about the difference between extension traits and other unnamed traits, but enforcing such a policy seems nice.

martinvonz:
I don't think there's any difference from the Rust language's point of view, but things like `ObjectId` are meant to be implemented by the types that are object ids. The implementations are generally defined in the same module as the implementing type is defined. Extension traits instead add functionality to existing types. The implementations are in the same module as the trait. I'm sure someone can correct me if I'm wrong :)

PR #6033, opened 2025-03-15
cleanup: enable `unused_trait_names` clippy lint and run `clippy --fix`

ilyagr:
This competes with #6032. Cc @martinvonz


martinvonz (approved):
LGTM, but let's give others a chance to object.

ilyagr:
OK, I'll wait 24 hours for objections and comments unless a consensus seems to form sooner.

arxanas:
I'm in favor personally.

emilazy:
It seems sensible if we care about having a canonical form for this. I don’t like https://github.com/jj-vcs/jj/pull/6032 since it feels like a waste of time to apply human judgement to what imports are extension traits. I guess I worry about IDEs generating imports that cause warnings and that causing manual toil?

ilyagr:
> I guess I worry about IDEs generating imports that cause warnings and that causing manual toil?

Yeah, I'm not sure how well `rust-analyzer` supports this config, but I hope for the best, and on the fact that it's actively being improved. At worst, one can always run `cargo clippy --fix --workspace --allow-dirty`.

ilyagr:
Apparently, it's not quite perfectly supported: https://github.com/rust-lang/rust-analyzer/issues/18818. I'd still try the setting out and see if it's a real problem, I'm hoping that usually it'll add at most a few seconds of work.

ilyagr:
Chatted with Emily briefly on Discord, she's OK with merging this.

PR #6035, opened 2025-03-15
git lfs: add support for ignoring Git LFS files

bcspragu:
Copied from #5480:

> This PR adds support for ignoring files handled by git lfs. `jj` previously detected these files as changed as it expected the original file content before the `git lfs` checkout to be present. With this change it's possible to opt in (via the `git.ignore-lfs-files` setting) into just ignoring this files from `jj`. They then need to be manually handled via e.g. git if they are changed.
>
> From a technical point of view this change parses all `.gitattributes` files in the current repository and constructs a ignore list from that. This ignore list is then applied through all subcommands.
> 
> This is still missing tests, as I'm really unsure what to test about and how to write these tests. I likely won't have the capacity to figure that out on my own, so I would appreciate if someone could point to the correct places.
>
> Somewhat addresses #80 

Main changes from that PR:

- Mirror how `.gitignore` files are handled with `GitIgnoreFile` (thanks @kejadlen!)
- Add some (currently not all passing) tests

This is a draft, I'm still actively working on it and I'll squash everything down to a single commit once it looks good.

# Checklist

If applicable:
  - In progress


bcspragu:
Ugh I meant to make this a draft, closing

bnjmnt4n:
FYI, you can convert an existing non-draft PR to draft from the "Reviewers" section in the right: 
CleanShot 2025-03-15 at 16 34 27@2x

Admittedly, it's a bit hard to spot if you don't know what you're looking for.

cli/src/merge_tools/mod.rs
```diff
@@ -431,6 +450,7 @@ mod tests {
                 name,
                 &settings,
                 GitIgnoreFile::empty(),
+                Some(GitAttributesFile::empty()),
```
kejadlen, line 453: I think this (and below) can just be `None` unless you're planning on adding tests around this.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,191 @@
...
+        input: &[u8],
+    ) -> Result<Arc<GitAttributesFile>, GitAttributesError> {
+        let mut source_file = prefix.clone();
+        source_file.push(".gitattributes");
+
+        let mut search = self.search.clone();
+        let mut collection = self.collection.clone();
+        search.add_patterns_buffer(input, source_file, Some(&prefix), &mut collection, true);
+        Ok(Arc::new(GitAttributesFile { search, collection }))
+    }
```
kejadlen, line 60: I wound up doing something similar myself when looking at testing this (https://github.com/kejadlen/jj/commit/5c423f84a5be1e268fb43c33772c3356fab1ac54) but I'm not particularly fond of not testing the path that's actually being exercised in actual usage, so have been mulling on a satisfactory alternative.

For manual testing, I've been using [this](https://github.com/kejadlen/jj-lfs-test/commit/b863adeaec5d549f2fef5dd1af5a0baa78844193) and was wondering if it's worthwhile to package this up and write a more integration-y level test? 

```sh
❯ touch a.png b.jpg foo/c.png foo/d.jpg
❯ jj st
Working copy changes:
A b.jpg
Working copy : wuvuwzpo c3e9dd78 (no description set)
Parent commit: oznmowzq b863adea μ
```

kejadlen: Although that being said, I think I've found a bug in that I've only tested with new files so far, and we probably need to exclude also already tracked files when snapshotting?

kejadlen: Alright, splitting out the LFS conditional in `process_dir_entry` seems to do the trick, although it is a somewhat confusing user experience, with `jj st` showing that the LFS files will be deleted (although not actually doing so in the commit).

PR #6036, opened 2025-03-15
git lfs: add support for ignoring Git LFS files

bcspragu:
Copied from #5480:

> This PR adds support for ignoring files handled by git lfs. `jj` previously detected these files as changed as it expected the original file content before the `git lfs` checkout to be present. With this change it's possible to opt in (via the `git.ignore-lfs-files` setting) into just ignoring this files from `jj`. They then need to be manually handled via e.g. git if they are changed.
>
> From a technical point of view this change parses all `.gitattributes` files in the current repository and constructs a ignore list from that. This ignore list is then applied through all subcommands.
> 
> This is still missing tests, as I'm really unsure what to test about and how to write these tests. I likely won't have the capacity to figure that out on my own, so I would appreciate if someone could point to the correct places.
>
> Somewhat addresses #80 

Main changes from that PR:

- Mirror how `.gitignore` files are handled with `GitIgnoreFile` (thanks @kejadlen!)
- Add some (currently not all passing) tests

This is a draft, I'm still actively working on it and I'll squash everything down to a single commit once it looks good.

# Checklist

If applicable:
  - In progress


lib/src/local_working_copy.rs
```diff
@@ -1267,10 +1289,14 @@ impl FileSnapshotter<'_> {
                 progress(&path);
             }
             if maybe_current_file_state.is_none()
-                && git_ignore.matches(path.as_internal_file_string())
+                && (git_ignore.matches(path.as_internal_file_string())
+                    || git_attributes
+                        .as_ref()
+                        .map(|ga| ga.matches(path.as_internal_file_string()))
+                        .unwrap_or(false))
```
kejadlen, line 1296: I realized that this conditional needs to be separate from `maybe_current_file_state.is_none()` (see https://github.com/jj-vcs/jj/pull/6035/files#r1997468647, which I left on the wrong PR)

cli/src/merge_tools/mod.rs
```diff
@@ -431,6 +450,7 @@ mod tests {
                 name,
                 &settings,
                 GitIgnoreFile::empty(),
+                Some(GitAttributesFile::empty()),
```
kejadlen, line 453: I think this (and below) can just be `None` if we're not going to test LFS files here.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,191 @@
...
+        input: &[u8],
+    ) -> Result<Arc<GitAttributesFile>, GitAttributesError> {
+        let mut source_file = prefix.clone();
+        source_file.push(".gitattributes");
+
+        let mut search = self.search.clone();
+        let mut collection = self.collection.clone();
+        search.add_patterns_buffer(input, source_file, Some(&prefix), &mut collection, true);
+        Ok(Arc::new(GitAttributesFile { search, collection }))
+    }
```
kejadlen, line 60: (Copying over my comment from the closed PR)

I wound up doing something similar myself when looking at testing this (https://github.com/kejadlen/jj/commit/5c423f84a5be1e268fb43c33772c3356fab1ac54) but I'm not particularly fond of not testing the path that's actually being exercised in actual usage, so have been mulling on a satisfactory alternative.

For manual testing, I've been using [this](https://github.com/kejadlen/jj-lfs-test/commit/b863adeaec5d549f2fef5dd1af5a0baa78844193) and was wondering if it's worthwhile to package this up and write a more integration-y level test?

```sh
❯ touch a.png b.jpg foo/c.png foo/d.jpg
❯ jj st
Working copy changes:
A b.jpg
Working copy : wuvuwzpo c3e9dd78 (no description set)
Parent commit: oznmowzq b863adea μ
```

lib/src/local_working_copy.rs
kejadlen: @bcspragu Had to do this also to actually have everything working:

https://github.com/jj-vcs/jj/compare/main...kejadlen:jj:lfs#diff-9654f3ae48bcd40ea4d2cdd4bfdebb9da27adb5373489816740d20642b174170R1435-R1441

kejadlen:
Just a note that I found there's a bug with `[[:space:]]` in the pattern - I think a bug with gitoxide pattern matching itself, since it repros with just `foo[[:space:]]bar` in `.gitignore`.

cli/src/config-schema.json
```diff
@@ -439,6 +439,11 @@
                     "type": "string",
                     "description": "Path to the git executable",
                     "default": "git"
+                },
+                "ignore-lfs-files": {
+                    "type": "boolean",
+                    "description": "Whether jj ignores files handled by git lfs or not",
+                    "default": false
```
Valodim, line 446: I would advocate for defaulting this to `true`. In a repo that has `git lfs` tracked files, it will be an exceedingly rare exception that the user wants `jj` to track these files.

Valodim: Thinking about what possible use cases there could be for this behavior some more, I would go as far as to say that setting this to false is never useful practice: Tracking lfs files will immediately remove their lfs property, so this is effectively the same as running `git lfs untrack .`. This makes a value of false here just a roundabout way of removing `git lfs` from a repo, an action that the user will much more likely want to perform explicitly as lfs untrack && lfs uninstall.

kejadlen:
@bcspragu @Valodim I'm keeping a branch relatively up to date here (includes your suggested changes and some tests as well): https://github.com/jj-vcs/jj/compare/main...kejadlen:jj:lfs?expand=1

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,193 @@
...
+        };
+
+        let mut out = gix_attrs::search::Outcome::default();
+        out.initialize_with_selection(&self.collection, ["filter"]);
+        self.search.pattern_matching_relative_path(
+            path.into(),
+            gix_glob::pattern::Case::Sensitive,
+            Some(is_dir),
+            &mut out,
+        );
```
gusinacio, line 101: I've been trying to use this branch with my LFS setup and it wasn't working.

It turns out that my `.gitattributes` in root was:
```
path/to/data/*.bin filter=lfs diff=lfs merge=lfs -text
```
This code only works fore relative path, so I had to move my .gitattributes to `path/to/data/.gitattributes` and use relative lfs.

kejadlen: I took a quick look and curiously, it seems like relatively pathed files do match the search, but wind up with an assignment with an ["unspecified"](https://docs.rs/gix-attributes/0.24.0/gix_attributes/enum.State.html#variant.Unspecified) filter value (instead of the expected `Value("lfs")`). Not sure if I'm using the API incorrectly or if there's perhaps an upstream bug.

mikelu92:
Hi I'm just wondering if it's possible to expand this pr out to not just lfs files but also different kinds of .gitattributes filters? I use [git-crypt](https://github.com/AGWA/git-crypt) in some of my repos and it has similar lines in the .gitattributes file as git-lfs. From the git-crypt repo:

```
secretfile filter=git-crypt diff=git-crypt
*.key filter=git-crypt diff=git-crypt
secretdir/** filter=git-crypt diff=git-crypt
```

I'm not a rust guy so it's hard for me to follow exactly how this works and modify myself, unfortunately.

kejadlen:
Maybe a `git.ignore_filters` setting that defaults to at least `["lfs"]` and perhaps other common filters? I can take a look at adding this to my branch.

bcspragu:
Hello yes it's me, the delinquent OP!

I've been caught up in other things (so it goes), so anyone should feel free to pick it up from here. I _will_ get it done eventually if nobody else does, but I'm no longer regularly working in LFS (or `git-crypt`) -containing repos, so it has (selfishly) lost its urgency for me to finish it.

kejadlen:
I'm happy to pick it up and I've added the functionality to specify filter types to ignore in [my branch](https://github.com/kejadlen/jj/tree/ignore-filters). Would it be preferable for me to create a new PR? (Which needs some amount of tidying before I can do so.)

mikelu92:
> I'm happy to pick it up and I've added the functionality to specify filter types to ignore in [my branch](https://github.com/kejadlen/jj/tree/ignore-filters). Would it be preferable for me to create a new PR? (Which needs some amount of tidying before I can do so.)

@kejadlen how would you set up your branch for use? add a git config `ignore-files = ['git-crypt']`?

bcspragu:
> Would it be preferable for me to create a new PR? (Which needs some amount of tidying before I can do so.)

Either way works for me, happy to give you write access to my fork if we want to not split the conversation (again), but also feel free to take whatever's useful and run with it!

kejadlen:
@bcspragu Sounds good - I'm going to squash all our previous work into a single commit and continue from there if you don't mind?

adrian-brady:
@kejadlen I have added the config option `ignore-files = ["lfs"]` but lfs files are shown in the status. I built off of your branch "ignore-filters"

If you don't mind I could also do some work on this as well.

adrian-brady:
Would a new git subcommand which manually updates files tracked/matched by gitattributes be something we're interested in? Gitattributes really seem to be incompatible with JJ's workflow and I think they should work as a git-specific feature. This would also allow for a similar feature to be implemented for JJ in the future.

Also, if there is a gitattributes file present, I think there should be _some_ JJ UX to let the user know some files' changes won't be tracked. It would be easy to add an ignore config option and then not realize some repo utilizes gitattributes. Perhaps a warning message when running status or other commands if a gitattributes file is present in the repo?

PhilipMetzger:
Also if you base the work off others please attribute the squashed commits to _all authors_ who contributed with the `Co-authored-by:` trailer whenever you're done.

shanesveller:
> Also if you base the work off others please attribute the squashed commits to _all authors_ who contributed with the `Co-authored-by:` trailer whenever you're done.

> It might be interesting to have a flag to do that automatically in `jj squash`

The `Co-authored-by` trailer is a strong convention (and a very socially healthy practice), but it's not a formal standard that is universal to Git and honored by all tools and forges. It's a useful hack.

I imagine `jj`'s eventual native backend or other non-Git backends would prefer to model that sort of info as something other than a message trailer. With that in mind, as an end-user I'm a little ambivalent towards `jj` taking automatic steps to use or populate `Co-authored-by` because it helps ossify the use of trailers and that one in particular. Opt-in CLI flag seems like the right model if it gets included at all.

cwfitzgerald:
Just want to +1 @kejadlen's up to date fork, it just got me out of a bind and getting this feature landed in jj would allow a major limitation to jj to be reasonably worked around!

oars:
> @kejadlen I have added the config option `ignore-files = ["lfs"]` but lfs files are shown in the status. I built off of your branch "ignore-filters"
> 
> If you don't mind I could also do some work on this as well.

@adrian-brady did you figure it out? I'm seeing the same thing. with:

```
jj --version jj 0.30.0-45e311320fd1bb893bea853bd828346bd1876383
```

kejadlen:
This PR can probably be closed now in favor of #7098

PR #6042, opened 2025-03-16
config docs: touch up platform-specific config location docs

ilyagr:
cc @calebdw, @algmyr if you are interested

Screenshot of `uv run mkdocs serve`, in case it helps (without opening the info block) (**Update:** the screenshot is a bit obsolete):

image

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1463,27 +1463,27 @@ An easy way to find the user config file/directory is:
 jj config path --user
 ```
 
-On all platforms, the user's global `jj` configurations are loaded in the following
-precedence order (with later configs overriding earlier ones):
+On all platforms, the user's global `jj` configurations are by default loaded in
+the following precedence order (with later configs overriding earlier ones):
 
 - `$HOME/.jjconfig.toml`
-- `$XDG_CONFIG_HOME/jj/config.toml` (preferred)
-- `$XDG_CONFIG_HOME/jj/conf.d/*.toml`
+- `<PLATFORM_SPECIFIC>/jj/config.toml` (preferred)
+- `<PLATFORM_SPECIFIC>/jj/conf.d/*.toml`
 
 where `$HOME` represents the user's home directory (`%USERPROFILE%` on Windows),
-and `$XDG_CONFIG_HOME` represents the platform-specific configuration directory.
-The platform-specific location is recommended for better integration with platform
-services.
+and `<PLATFORM_SPECIFIC>` represents the platform-specific configuration
+directory shown in the table below. The platform-specific location is
+recommended for better integration with platform services.
 
 The files in the `conf.d` directory are loaded in lexicographic order. This allows
 configs to be split across multiple files and combines well
 with [Conditional Variables](#conditional-variables).
 
-| Platform | Value                                              | Example                                                   |
-| :------- | :------------------------------------------------- | :-------------------------------------------------------- |
-| Linux    | `$XDG_CONFIG_HOME/jj/config.toml`                  | `/home/alice/.config/jj/config.toml`                      |
-| macOS    | `$HOME/Library/Application Support/jj/config.toml` | `/Users/Alice/Library/Application Support/jj/config.toml` |
-| Windows  | `{FOLDERID_RoamingAppData}\jj\config.toml`         | `C:\Users\Alice\AppData\Roaming\jj\config.toml`           |
+| Platform | Location of `<PLATFORM_SPECIFIC>` dir | Example config file location                              |
+| :------- | :------------------------------------ | :-------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -1467,23 +1467,33 @@ On all platforms, the user's global `jj` configurations are loaded in the follow
 precedence order (with later configs overriding earlier ones):
 
 - `$HOME/.jjconfig.toml`
-- `$XDG_CONFIG_HOME/jj/config.toml` (preferred)
-- `$XDG_CONFIG_HOME/jj/conf.d/*.toml`
+- `$PLATFORM_SPECIFIC/jj/config.toml` (preferred)
+- `$PLATFORM_SPECIFIC/jj/conf.d/*.toml`
```
calebdw, line 1471: These are not real env variables so I would avoid writing them as such, consider something like:

```suggestion
- `<PLATFORM_SPECIFIC>/jj/config.toml` (preferred)
- `<PLATFORM_SPECIFIC>/jj/conf.d/*.toml`
```

docs/config.md
```diff
@@ -1467,23 +1467,34 @@ On all platforms, the user's global `jj` configurations are loaded in the follow
...
+??? info "Using the XDG (Linux) config directory on a Mac"
+
+    If you prefer to use the XDG location on a Mac, you can create a symlink. For
+    example:
+
+    ```shell
+    # In most shells, putting ~ inside quotes does not work, hence the careful quoting
+    mv ~/'Library/Application Support/jj' ~/.config
+    ln -s ~/.config/jj ~/'Library/Application Support/jj'
+    ```
```
calebdw, line 1497: I would consider leaving this out, as a Mac user that wants to use `~/config` would likely already know to create the symlink


If you really want to leave it in then I would suggest making it more general:

```suggestion
??? info "Using the XDG (Linux) config directory on a Mac"

    If you prefer to use the XDG location on a Mac, you can create a symlink. For
    example:

    ```shell
    # In most shells, putting ~ inside quotes does not work, hence the careful quoting
    mv ~/'Library/Application Support' ~/.config
    ln -s ~/.config ~/'Library/Application Support'
    ```
```

Lastly, I think you have a bug in your original script:

```suggestion
??? info "Using the XDG (Linux) config directory on a Mac"

    If you prefer to use the XDG location on a Mac, you can create a symlink. For
    example:

    ```shell
    # In most shells, putting ~ inside quotes does not work, hence the careful quoting
    mv ~/'Library/Application Support/jj' ~/.config/jj
    ln -s ~/.config/jj ~/'Library/Application Support/jj'
    ```
```

calebdw (approved):

docs/config.md
ilyagr: I did have a few bugs there. I fixed them a bit differently than you suggested. Let me know if you still see errors.

I don't think we should recommend linking all of `~/'Library/Application Support'` to `~/.config`. On my machine, `~/'Library/Application Support'` contains a ton of configs of different Mac-specific apps. I'm not sure all of them will respect a symlinked config dir and, even if they did, I'm not sure I want to have `~/.config/com.apple.RemoteManagementAgent/`.

I'm not 100% sure whether we should keep this advice or not. The fact that `jj` follows symlinks when reading configs is not trivial, I was considering making a shorter sentence to that effect instead, but writing down a complete example seemed more helpful.

Aside: For me, at least, the reason I put `jj`'s config in the XDG dir is because I share most of the config between my Mac and Linux machines.

docs/config.md
ilyagr: Good idea, thanks!

docs/config.md
ilyagr: OK, I'll resolve this based on your thumbs up. Thank you again!

ilyagr:
I'll probably merge this in a few hours unless I get more comments.

docs/config.md
ilyagr: Hmm, OTOH maybe it's better to recommend setting `JJ_CONFIG` instead or, at least, in addition to the symlink approach.

docs/config.md
```diff
@@ -1467,23 +1467,34 @@ On all platforms, the user's global `jj` configurations are loaded in the follow
 precedence order (with later configs overriding earlier ones):
```
ilyagr, line 1467: also might want to mention "unless `JJ_CONFIG` is set" somewhere around here.

calebdw: That's already mentioned on line 1499

ilyagr: (I did this edit)

Yes, but not everyone will read all the way down to there. Also, I'd say that strictly speaking, this statement is simply incorrect without a caveat.

docs/config.md
ilyagr: I think I'll take out this part for now and then make another commit. There are a few more changes I'd like to make, but they are less obvious.

PR #6043, opened 2025-03-16
prompt: add a `prompt_choice_with` function

calebdw:
This function takes a list of choices, prints them to the console, and prompts the user to select one via its index. It returns the selected index.

# Checklist

If applicable:


cli/tests/test_config_command.rs
```diff
@@ -1012,9 +1012,9 @@ fn test_config_only_loads_toml_files() {
     std::fs::File::create(test_env.config_path().join("is-not.loaded")).unwrap();
     insta::assert_snapshot!(test_env.run_jj_in(".", ["config", "edit", "--user"]), @r"
     ------- stderr -------
-    1: $TEST_ENV/config/config0001.toml
-    2: $TEST_ENV/config/config0002.toml
-    Choose a config file (default 1): 1
+    0: $TEST_ENV/config/config0001.toml
```
bnjmnt4n, line 1015: Should we preserve the 1-based indexing here?

cli/src/ui.rs
```diff
@@ -625,6 +625,49 @@ impl Ui {
...
+        choices: &[impl AsRef<str>],
+        default_index: Option<usize>,
+    ) -> io::Result<usize> {
+        for (i, choice) in choices.iter().enumerate() {
+            writeln!(self.stderr(), "{i}: {}", choice.as_ref())?;
+        }
+        let default = default_index
+            .map(|i| format!(" (default {i})"))
+            .unwrap_or_default();
+        let default_choice = default_index.and_then(|i| choices.get(i));
```
yuja, line 642: I wouldn't add `ui` function that prints `choices` internally unless we have more callers. Maybe you've noticed, but `jj next/prev` can't use this because they want to render template with color.

It's also not good that the underlying implementation is largely duplicated.

calebdw: I've removed the duplication, but I can't imagine that every caller would want to print the choices themselves (for just a simple list).

If you're still in disagreement then feel free to close this PR

ilyagr: Yeah, I'm not sure about this. It's quite possible that such a function would be needed at some point, but it's often good practice to not implement utility functions until there are more callers. I think there are other refactors that might happen before such a function is needed, for example.

yuja: > can't imagine that every caller would want to print the choices themselves (for just a simple list).

That's true, but we only have one caller who prints a simple list right now, and even in this example, the caller has to set up a stringified list. What the caller really wants is a function that choose `&Path` from `[Path]`?

```rust
let path = ui.prompt_choice(..., &paths, |path| path.display())?;
```

That said, implementing this over generic `T` might not be easy if lifetime `for<'a> (&'a T) -> Display + 'a` is involved. Also, this wouldn't work for templated list. So it might be simpler to just print the list by caller, and pick index by `ui` function:
```rust
for (i, path) in paths.iter().enumerate() {
    writeln!(ui.stderr(), "{}: {}", i + 1, path.display())?;
}
let index = ui.prompt_choice_index("Pick one", 1..=paths.len(), Some(1))? - 1;
```

FWIW, what I meant in the other PR is something like https://github.com/jj-vcs/jj/pull/6052

PR #6045, opened 2025-03-16
cli: use same per-commit labels for node and non-node templates

martinvonz:
Extracted from #5800. Seems like a good idea on its own.

# Checklist

If applicable:


cli/tests/test_evolog_command.rs
```diff
@@ -50,8 +50,8 @@ fn test_evolog_with_or_without_diff() {
     insta::assert_snapshot!(output, @r"
     [1m[38;5;2m@[0m  [1m[38;5;13mr[38;5;8mlvkpnrz[39m [38;5;3mtest.user@example.com[39m [38;5;14m2001-02-03 08:05:10[39m [38;5;12m6[38;5;8m6b42ad3[39m[0m
     │  [1mmy description[0m
-    [1m[38;5;1m×[0m  [1m[39mr[0m[38;5;8mlvkpnrz[39m hidden [38;5;3mtest.user@example.com[39m [38;5;6m2001-02-03 08:05:09[39m [1m[38;5;4m07[0m[38;5;8mb18245[39m [38;5;1mconflict[39m
-    │  my description
+    [1m[38;5;1m×[0m  [1m[39mr[0m[38;5;8mlvkpnrz[38;5;1m hidden [38;5;3mtest.user@example.com[38;5;1m [38;5;6m2001-02-03 08:05:09[38;5;1m [1m[38;5;4m07[0m[38;5;8mb18245[38;5;1m conflict[39m
+    │  [38;5;1mmy description[39m
```
martinvonz, line 54: Ah, this was not intended. Commits with conflicts now get a red description. Maybe we should use different label names for the commit-level conflict indication and the actual conflict text.

martinvonz: Done. I inserted a commit that renames the `node conflict` labels to `node conflicted`, and I updated the second commit to use the `conflicted` label in the commit template.

yuja (approved):
Maybe we can also rename "conflict" and "empty" to "conflict_label" or something, but that can be addressed later. It's confusing that we use "conflict" and "conflicted" for different purposes.

PR #6046, opened 2025-03-16
annotate: report root commit id if no origin found within range

yuja:
Suppose we add "jj file annotate" option to specify the search domain or depth, the root commit within the range can be displayed as the boundary commit. "git blame" for example displays boundary commits with ^ prefix.

# Checklist

If applicable:


martinvonz (approved):
Makes sense

PR #6047, opened 2025-03-16
git: add deprecation warnings for `git.subprocess`

emilazy:
~~Only intended for merge when we decide to start the deprecation process, but~~ should be ready ~~for code review now.~~

~~The first commit could be split out into another PR if we want to merge it now.~~

# Checklist

If applicable:

### DIFF
--- a/cli/src/git_util.rs
+++ b/cli/src/git_util.rs
@@ -30,6 +30,8 @@ use crossterm::terminal::Clear;
 use crossterm::terminal::ClearType;
 use indoc::writedoc;
 use itertools::Itertools as _;
+#[cfg(feature = "git2")]
+use jj_lib::config::ConfigGetResultExt as _;
 use jj_lib::fmt_util::binary_prefix;
 use jj_lib::git;
 use jj_lib::git::FailedRefExportReason;
@@ -41,6 +43,8 @@ use jj_lib::op_store::RemoteRef;
 use jj_lib::ref_name::RemoteRefSymbol;
 use jj_lib::repo::ReadonlyRepo;
 use jj_lib::repo::Repo;
+#[cfg(feature = "git2")]
+use jj_lib::settings::UserSettings;
 use jj_lib::workspace::Workspace;
 use unicode_width::UnicodeWidthStr as _;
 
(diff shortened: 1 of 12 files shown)


emilazy:
Undrafting this since @thoughtpolice expressed interest. This is of course not intended for merge before the upcoming patch release.

I put 0.30 as the version based on the Discord discussion, but that should just be treated as a placeholder for whatever’s decided.

PhilipMetzger:
I think the outlined four releases are good enough for me. Since the last time I expressed stronger opinions on it.

emilazy:
Rebased for conflicts, and added some docs updates. This should be ready now, if we want to do it. https://github.com/jj-vcs/jj/pull/6048 shows the rest of the owl.

cli/src/git_util.rs
```diff
@@ -275,6 +275,24 @@ pub fn with_remote_git_callbacks<T>(ui: &Ui, f: impl FnOnce(git::RemoteCallbacks
...
+            GIT2_WARNING.call_once(|| {
+                _ = writeln!(
+                    ui.warning_default(),
+                    "`git.subprocess = false` will be removed in 0.30; please report any issues \
+                     you have with the default.",
+                );
+            });
+        };
+        callbacks.git2_deprecation_warning = Some(&mut git2_deprecation_warning);
+    }
```
yuja, line 294: I don't think we need callback. Can't we instead add `warn(ui, &git_settings)` function and let `cmd_git_*()` functions call it?

Maybe we can also use `debug.suppress-git2-deprecation-warning` config key instead (because `UserSettings` should be available there.)

emilazy: Yeah, that works, and produces slightly nicer output. Done. (This also means that `with_git_subprocess` could have been a regular method after all, but it doesn’t feel worth the churn to change it.)

cli/src/cli_util.rs
```diff
@@ -2866,6 +2866,27 @@ pub fn print_unmatched_explicit_paths<'a>(
...
+            .unwrap_or(false)
+    {
+        writeln!(
+            ui.warning_default(),
+            "`git.subprocess = false` will be removed in 0.30; please report any issues you have \
+             with the default.",
+        )?;
+    }
+    Ok(())
+}
```
yuja, line 2888: nit: might be simpler to move this function to `git_util`, and handle `cfg(feature = "git2")` internally, but I don't have strong feeling.

yuja (approved):
Thanks.

cli/src/cli_util.rs
emilazy: I somewhat like the `#[cfg]` because it makes it easier to see what’s due for removal. `git_util` doesn’t depend on `cli_util` (which contains `WorkspaceCommandHelper`) currently (but `cli_util` does depend on `git_util`), do you think breaking that property to do the move would be okay?

yuja: I think `settings` can be passed in to this function instead. I tend to avoid inserting `cfg(..)` at random places, but that's just my preference.

emilazy: I moved it and adjusted the signature. I agree about `#[cfg]` in general, but I think in this case it helps make removal a little more mechanical, so I left it for now.

emilazy:
Actually, if we’re doing a point release for the gitoxide issue, should we wait to land this until after that’s done?

PR #6052, opened 2025-03-17
ui: extract prompt_choice_with(..callback)

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/ui.rs
+++ b/cli/src/ui.rs
@@ -597,32 +597,21 @@ impl Ui {
         choices: &[impl AsRef<str>],
         default_index: Option<usize>,
     ) -> io::Result<usize> {
-        if !Self::can_prompt() {
-            if let Some(index) = default_index {
-                // Choose the default automatically without waiting.
-                let choice = choices
+        self.prompt_choice_with(
+            prompt,
+            default_index.map(|index| {
+                choices
                     .get(index)
                     .expect("default_index should be within range")
-                    .as_ref();
-                writeln!(self.stderr(), "{prompt}: {choice}")?;
-                return Ok(index);
-            }
-        }
-
-        loop {
-            let choice = self.prompt(prompt)?;
-            let choice = choice.trim();
-            if choice.is_empty() {
-                if let Some(index) = default_index {
-                    return Ok(index);
-                }
-            }
-            if let Some(index) = choices.iter().position(|c| choice == c.as_ref()) {
-                return Ok(index);
-            }
-
-            writeln!(self.warning_no_heading(), "unrecognized response")?;
-        }
+                    .as_ref()
+            }),
+            |input| {
+                choices
+                    .iter()
+                    .position(|c| input == c.as_ref())
+                    .ok_or("unrecognized response")
+            },
+        )
     }
 
     /// Prompts for a yes-or-no response, with yes = true and no = false.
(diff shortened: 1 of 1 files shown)


cli/src/ui.rs
```diff
@@ -632,14 +621,53 @@ impl Ui {
...
+        prompt: &str,
+        default: Option<(T, &str)>,
+        mut parse: impl FnMut(&str) -> Result<T, E>,
+    ) -> io::Result<T> {
+        if !Self::can_prompt() {
+            if let Some((value, text)) = default {
+                // Choose the default automatically without waiting.
+                writeln!(self.stderr(), "{prompt}: {text}")?;
+                return Ok(value);
+            }
```
ilyagr, line 655: (Optional): Perhaps add a comment in between the `}`s:

```rust
// In this case, self.prompt() below will fail
```

You could also (or instead) add another line to the docstring:

```
/// If the default is not specified, empty input will lead to the function returning an error.
```

cli/src/ui.rs
```diff
@@ -632,14 +621,53 @@ impl Ui {
...
+
+    /// Repeats the given prompt until `parse(input)` returns a value.
+    ///
+    /// If the default `(value, text)` is specified, an empty input will be
+    /// mapped to the default `value`. The default `value` will also be used in
+    /// non-interactive session. The `text` will be printed in that case.
+    pub fn prompt_choice_with<T, E: fmt::Display>(
+        &self,
+        prompt: &str,
+        default: Option<(T, &str)>,
```
ilyagr, line 647: Perhaps it'd be simpler to understand this function if `default` was `Option<&str>`, and the default path evaluated `parse(default.unwrap()`? It would make it very slightly less flexible, but I think also less confusing. It'd also match what would have happened in the interactive case if the user typed in the default value.

cli/src/ui.rs
```diff
@@ -632,14 +621,53 @@ impl Ui {
...
+        if !Self::can_prompt() {
+            if let Some((value, text)) = default {
+                // Choose the default automatically without waiting.
+                writeln!(self.stderr(), "{prompt}: {text}")?;
+                return Ok(value);
+            }
+        }
+
+        loop {
+            let input = self.prompt(prompt)?;
```
ilyagr, line 659: (Optional): I'm not sure, but perhaps adding the default value to the prompt could happen here? There could also be an optional argument that affects how the default value is printed.

This is something that could be changed later as needed, though.

cli/src/ui.rs
yuja: Good point, done.

cli/src/ui.rs
yuja: Updated the doc comment. FWIW, "empty input" should be ignored in that case, and non-interactive prompt will fail.

cli/src/ui.rs
yuja: Yes/No prompt uses `Yn`/`nY`, so we can't simply inline the default text here.

ilyagr (approved):
Thanks!

cli/src/ui.rs
```diff
@@ -632,14 +621,54 @@ impl Ui {
...
+    /// will also be used in non-interactive session. The default `text` must
+    /// be parsable. If no default is given, this function will fail in
+    /// non-interactive session.
+    pub fn prompt_choice_with<T, E: fmt::Debug + fmt::Display>(
+        &self,
+        prompt: &str,
+        default: Option<&str>,
+        mut parse: impl FnMut(&str) -> Result<T, E>,
+    ) -> io::Result<T> {
+        if !Self::can_prompt() {
```
ilyagr, line 651: (Optional) Perhaps add a `debug_assert` that the default is parseable, to catch this error earlier?

yuja: made parsing mandatory. we don't care the parsing cost here.

PR #6055, opened 2025-03-17
tests: split CommandOutput to module, add sub-dir helper

yuja:
# Checklist

If applicable:


cli/tests/common/command_output.rs
martinvonz, line 1: > These types don't depend on TestEnvironment.

I had noticed this, btw, and I appreciate it. I hope we can reuse these in our internal tests at Google (but we don't have a good way of doing snapshot tests yet).

PR #6060, opened 2025-03-17
github: use faster linkers in CI

emilazy:
See commit messages, though I want to figure out if this is actually helping for Windows first.

# Checklist

If applicable:


emilazy:
Okay, the speed‐up here is really minor and I’m not sure it’s not even adding a few seconds for Windows – it’s very hard to tell with how noisy the measurements are, it seems to be faster more often than slower but it’s all lost in the noise – but it does seem to buy about 20 seconds on Linux at least, so I think it’s worth landing.

.cargo/config-ci.toml
```diff
@@ -2,8 +2,15 @@
...
 incremental = false
 
+[target.x86_64-unknown-linux-gnu]
+rustflags = ["-Clink-arg=-fuse-ld=mold"]
+
+[target.aarch64-unknown-linux-gnu]
+rustflags = ["-Clink-arg=-fuse-ld=mold"]
+
+[target.x86_64-pc-windows-msvc]
+linker = "rust-lld.exe"
```
PhilipMetzger, line 12: Nit: can't we just use normal `lld-link` here or is the Rust distribution a big deviation from it?

thoughtpolice (approved):
Reverting the timeout change 💯 🥇

.github/workflows/ci.yml
PhilipMetzger, line 1: Nit: Maybe add the two links here as a comment somewhere, so the next person has no reason to easily revert it.

PhilipMetzger (approved):
LG with minor nits.

.cargo/config-ci.toml
emilazy: `rust-lld` is what Rust is [using by default for `x86_64-unknown-linux-gnu` on Nightly](https://blog.rust-lang.org/2024/05/17/enabling-rust-lld-on-linux.html) and [considering defaulting to on Windows](https://github.com/rust-lang/rust/issues/71520), so this is the better‐paved path; it also matches the majority of the recommendations I saw online. I tested with `lld-link.exe` and it seemed like it was slower rather than faster, but see: incredibly noisy measurements. The Windows GitHub Actions builders have LLVM 14/15 installed, which is considerably older than Rust’s LLVM, though I don’t know if it was using `lld-link.exe` from there or from Rust.

.github/workflows/ci.yml
emilazy: Done.

ilyagr:
I did get the timeout of 15 minutes exceeded on macOS-Intel: https://github.com/jj-vcs/jj/actions/runs/13915898488/attempts/1 .

I have no idea whether this will be common enough to worry about, and on what platforms.

emilazy:
That’s unfortunate. That does seem unusually long compared to the other times I’ve been seeing, but we can roll this back if necessary. I’m hoping to think about how to speed up the x86 macOS CI once Windows isn’t reliably minutes slower than the others, though.

ilyagr:
If we need to, is it possible to increase the timeout just for Intel Mac? (and maybe Windows)

emilazy:
No point; the others never even get close to it anyway.

PR #6061, opened 2025-03-18
installation docs: workarounds for problems with `jj` fish completion

ilyagr:
Follow-up to https://github.com/jj-vcs/jj/pull/6059, see also https://github.com/fish-shell/fish-shell/pull/11046.

Thanks for taking care of the PRs linked above, @krobelus ! I dropped the ball on it a bit. I wasn't sure whether or not it was time to merge https://github.com/fish-shell/fish-shell/pull/11046, but I trust your judgement.

I wonder how common packaging or config problems will be with this. Do instructions like these seem reasonable?

### DIFF
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -272,6 +272,17 @@ jj util completion fish | source
 COMPLETE=fish jj | source
 ```
 
+!!! note
+
+    If dynamic completion work only partially, or if you get only static completions,
+    this might be because your `fish` version is older than 4.1 or because the static
+    completions are loaded by your personal config and/or by your distribution's
+    config. In the latter case, contact whomever made the `jj` package you are using.
+
+    Packagers: it is recommended to configure dynamic completions for `fish` in the
+    `jj` package if the package is likely to be used with older versions of fish, and to
+    not configure fish completions at all if `jj` is likely to be used by `fish` >= 4.1.
+
 ### Nushell
 
 ```nu
--- a/docs/install-and-setup.md
+++ b/docs/install-and-setup.md
@@ -257,19 +257,18 @@ source <(COMPLETE=zsh jj)
 
 ### Fish
 
-!!! note
-
-    No configuration is required with fish >= 4.1 which loads dynamic completions by default.
 
 #### Standard
 
 ```shell
+# Not recommended, use the dynamic completions instead
 jj util completion fish | source
 ```
 
 #### Dynamic
 
 ```shell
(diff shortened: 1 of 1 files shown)


docs/install-and-setup.md
```diff
@@ -257,22 +257,32 @@ source <(COMPLETE=zsh jj)
...
+!!! note
+
+    If dynamic completion work only partially, or if you get only static completions,
+    this might be because your `fish` version is older than 4.1 or because the static
+    completions are loaded by your personal config and/or by your distribution's
+    config. In the latter case, contact whomever made the `jj` package you are using.
+
+    Packagers: it is recommended to configure dynamic completions for `fish` in the
+    `jj` package if the package is likely to be used with older versions of fish, and to
+    not configure fish completions at all if `jj` is likely to be used by `fish` >= 4.1.
```
krobelus, line 284: note that `jj` packagers adding `/usr/share/fish/vendor_completions.d/jj.fish` shouldn't do harm in practice
(it will override the one in `/usr/share/fish/completions/jj.fish`).
Of course in an ideal world shell-specific completion files would not exist..

docs/install-and-setup.md
```diff
@@ -257,22 +257,32 @@ source <(COMPLETE=zsh jj)
...
+# Only needed with fish < 4.1. fish >= 4.1 loads these by default
 COMPLETE=fish jj | source
 ```
 
+!!! note
+
+    If dynamic completion work only partially, or if you get only static completions,
+    this might be because your `fish` version is older than 4.1 or because the static
+    completions are loaded by your personal config and/or by your distribution's
+    config. In the latter case, contact whomever made the `jj` package you are using.
```
krobelus, line 280: may be worth noting that, as a workaround for this, one may drop `COMPLETE=fish jj | source` into `~/.config/fish/completions` -- that will override any `jj.fish` from old fish or a distro package.

docs/install-and-setup.md
emilazy: Is there any shell performance advantage to packagers installing that file?

krobelus: I don't see why it would (sounds like a https://funroll-loops.probablymalware.lol/Gentoo-is-Rice.html thing)

emilazy: Well, startup time can get quite painful if you run a bunch of commands on every shell. But I guess in this case `COMPLETE=fish jj` is only run lazily, so there’s no reason for us to do it?

PR #6062, opened 2025-03-18
repo: suppress warning about unused SimpleBackend on non-dev build

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -396,7 +396,6 @@ impl Default for StoreFactories {
         let mut factories = StoreFactories::empty();
 
         // Backends
-        #[cfg(feature = "testing")]
         factories.add_backend(
             SimpleBackend::name(),
             Box::new(|_settings, store_path| Ok(Box::new(SimpleBackend::load(store_path)))),


ilyagr (approved):

lib/src/repo.rs
```diff
@@ -398,8 +397,12 @@ impl Default for StoreFactories {
         // Backends
         #[cfg(feature = "testing")]
```
martinvonz, line 398: Or maybe we should just remove this? Otherwise you won't be able to load the repo created by `jj util init-simple`. (Unless you build `jj` with the `testing` feature enabled, such as when building the fake binaries.)

emilazy: It seems good or at least harmless for standard binaries to support playing with a non‐Git backend? It’s already locked behind `jj debug`.

martinvonz: Yes, I'm fine with that. I'm just saying that the current state (after #5827) is that you can run `jj debug init-simple` to create a repo backed by `SimpleBackend` but then you can't run any commands in it.

emilazy: I guess the comment in `lib/src/lib.rs` would need adjusting/removing if the feature gate was dropped. (Though I don’t think it would make sense for the CLI to depend on `examples/` code anyway?)

yuja: Removed `#[cfg(..)]`. I assumed `jj debug init-simple` would also be feature-gated, but it isn't.

martinvonz (approved):

lib/src/repo.rs
emilazy: I’m confused about what the plan to execute on this TODO comment is:

https://github.com/jj-vcs/jj/blob/837eee90bea1e23e5aa84c3a4eef87c287ef9ebd/lib/src/lib.rs#L99-L102

Is the idea to have pluggable backends that can be loaded at runtime, or to remove all support for the simple backend from the CLI entirely and make `jj(1)` only capable of working with the Git backend? If we’re not doing plugins and intend to keep `jj debug init-simple` around (which seems reasonable to me) then it seems like the comment can’t be the plan.

PR #6067, opened 2025-03-18
git: forward rejection reason on push

bsdinis:
We've been finding that a lot of bug reports on `jj git push` come from
sub-standard error reporting on the reasons the failure happens.

It can come from a number of places:
 - hook failure
 - remote branch protection
 - git config

This commit forwards the reason as explained by the ouptut of git push
to help users figure out what is happening.
# Checklist

If applicable:

### DIFF
--- a/cli/src/command_error.rs
+++ b/cli/src/command_error.rs
@@ -554,16 +554,6 @@ jj currently does not support partial clones. To use jj with this repository, tr
                     err,
                     "Run `jj git remote rename` to give a different name.",
                 ),
-                GitPushError::RefInUnexpectedLocation(refs) => user_error_with_hint(
-                    format!(
-                        "Refusing to push a bookmark that unexpectedly moved on the remote. \
-                         Affected refs: {}",
-                        refs.join(", ")
-                    ),
-                    "Try fetching from the remote, then make the bookmark point to where you want \
-                     it to be, and push again.",
-                ),
-                GitPushError::RefUpdateRejected(_) => user_error(err),
                 #[cfg(feature = "git2")]
                 GitPushError::InternalGitError(err) => map_git2_error(err),
                 GitPushError::Subprocess(_) => user_error(err),
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -393,9 +393,12 @@ fn process_push_stats(push_stats: &GitPushStats) -> Result<(), CommandError> {
                     formatter,
                     "The following references unexpectedly moved on the remote:"
                 )?;
-                for reference in &push_stats.rejected {
+                for (reference, reason) in &push_stats.rejected {
                     write!(formatter, "  ")?;
                     write!(formatter.labeled("git_ref"), "{reference}")?;
+                    if let Some(r) = reason {
+                        write!(formatter, " (reason: {r})")?;
+                    }
                     writeln!(formatter)?;
                 }
                 Ok(())
@@ -408,9 +411,12 @@ fn process_push_stats(push_stats: &GitPushStats) -> Result<(), CommandError> {
         if !push_stats.remote_rejected.is_empty() {
             error.add_formatted_hint_with(|formatter| {
                 writeln!(formatter, "The remote rejected the following updates:")?;
-                for reference in &push_stats.remote_rejected {
+                for (reference, reason) in &push_stats.remote_rejected {
                     write!(formatter, "  ")?;
                     write!(formatter.labeled("git_ref"), "{reference}")?;
+                    if let Some(r) = reason {
+                        write!(formatter, " (reason: {r})")?;
+                    }
                     writeln!(formatter)?;
                 }
                 Ok(())
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -28,6 +28,7 @@ use jj_lib::commit::CommitIteratorExt as _;
 use jj_lib::config::ConfigGetResultExt as _;
 use jj_lib::git;
 use jj_lib::git::GitBranchPushTargets;
+use jj_lib::git::GitPushStats;
 use jj_lib::object_id::ObjectId as _;
 use jj_lib::op_store::RefTarget;
 use jj_lib::refs::classify_bookmark_push_action;
@@ -375,11 +376,51 @@ pub fn cmd_git_push(
         branch_updates: bookmark_updates,
     };
     let git_settings = tx.settings().git_settings()?;
-    with_remote_git_callbacks(ui, |cb| {
+    let push_stats = with_remote_git_callbacks(ui, |cb| {
         git::push_branches(tx.repo_mut(), &git_settings, remote, &targets, cb)
     })?;
+    process_push_stats(&push_stats)?;
     tx.finish(ui, tx_description)?;
     Ok(())
+}
+
+fn process_push_stats(push_stats: &GitPushStats) -> Result<(), CommandError> {
+    if !push_stats.all_ok() {
+        let mut error = user_error("Failed to push some bookmarks");
+        if !push_stats.rejected.is_empty() {
+            error.add_formatted_hint_with(|formatter| {
+                writeln!(
+                    formatter,
+                    "The following references unexpectedly moved on the remote:"
+                )?;
+                for reference in &push_stats.rejected {
+                    write!(formatter, "  ")?;
+                    write!(formatter.labeled("git_ref"), "{reference}")?;
+                    writeln!(formatter)?;
+                }
+                Ok(())
+            });
+            error.add_hint(
+                "Try fetching from the remote, then make the bookmark point to where you want it \
+                 to be, and push again.",
+            );
+        }
+        if !push_stats.remote_rejected.is_empty() {
+            error.add_formatted_hint_with(|formatter| {
+                writeln!(formatter, "The remote rejected the following updates:")?;
+                for reference in &push_stats.remote_rejected {
+                    write!(formatter, "  ")?;
+                    write!(formatter.labeled("git_ref"), "{reference}")?;
+                    writeln!(formatter)?;
+                }
+                Ok(())
+            });
+            error.add_hint("Try checking if you have permission to push to all the bookmarks.");
+        }
+        Err(error)
+    } else {
+        Ok(())
+    }
 }
 
 /// Validates that the commits that will be pushed are ready (have authorship
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -100,6 +100,23 @@ pub enum GitRefKind {
     Tag,
 }
 
+/// Stats from a git push
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct GitPushStats {
+    /// reference accepted by the remote
+    pub pushed: Vec<String>,
+    /// rejected reference, due to lease failure, with an optional reason
+    pub rejected: Vec<String>,
+    /// reference rejected by the remote, with an optional reason
+    pub remote_rejected: Vec<String>,
+}
+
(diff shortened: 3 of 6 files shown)


bsdinis:
cc @emilazy

lib/src/git.rs
```diff
@@ -2251,10 +2251,11 @@ pub enum GitPushError {
...
     RemoteName(#[from] GitRemoteNameError),
-    #[error("Refs in unexpected location: {0:?}")]
-    RefInUnexpectedLocation(Vec<String>),
-    #[error("Remote rejected the update of some refs (do you have permission to push to {0:?}?)")]
-    RefUpdateRejected(Vec<String>),
+    #[error("Failed to push some refs")]
+    FailedRefs {
+        unexpected_locations: Vec<String>,
+        rejected_updates: Vec<String>,
+    },
```
yuja: It's probably better to add support for partial success by not reporting failed refs as an error. (see TODO comment in `push_branches()`)

I think we can refactor the implementation as follows:

1. make `GitPushStats` public, and add `GitPushError::PartialRefUpdate(GitPushStats)` to fix the immediate problem.
2. remove deprecated git2 code path (or port git2 impl to `GitPushStats`.
3. make `git::push_branches()`/`push_updates()` return `Result<GitPushStats, _>`, update jj's view accordingly.

I think 2 and 3 can be addressed later.

cli/src/command_error.rs
```diff
@@ -554,16 +554,35 @@ jj currently does not support partial clones. To use jj with this repository, tr
...
+                    unexpected_locations,
+                    rejected_updates,
+                } => {
+                    let reference_errors = unexpected_locations
+                        .iter()
+                        .map(|x| format!("{x}: unexpectedly moved on the remote"))
+                        .chain(rejected_updates.iter().map(|x| {
+                            format!("{x}: remote rejected (do you have permission to push?)")
+                        }))
+                        .join("\n");
```
yuja: Maybe we can add a function (similar to `git_util::print_failed_git_export()`) which formats failed refs and reasons to `&mut dyn Formatter`. Then, the formatted output can be embedded as a hint by `.add_formatted_hint_with()`. It isn't actually a hint, but we don't have support for multi-line/labeled error messages.

If we add support for partial push, failed refs might be printed as a warning.

cli/src/command_error.rs
```diff
@@ -554,16 +554,48 @@ jj currently does not support partial clones. To use jj with this repository, tr
...
+                    rejected_updates,
+                } => {
+                    let reference_errors = unexpected_locations
+                        .iter()
+                        .map(|(reference, reason)| {
+                            let reason = reason
+                                .as_ref()
+                                .map(|x| format!(" {x}"))
+                                .unwrap_or("".to_string());
+                            format!("{reference}: unexpectedly moved on the remote{reason}")
```
emilazy, line 568: The ` (…)` is part of [Git’s output format](https://github.com/git/git/blob/683c54c999c301c2cd6f715c411407c413b1d84e/transport.c#L630-L637) rather than from the remote. We should probably strip off `" ("` and `")"` when parsing rather than including it in our return value.

bsdinis:
One problem here is that the reason depends on wheter we are in git2 or subprocessing.
Thoughts on the best way to cleanup the output for that? One idea here would be to just strip everything after " (reason: "

cli/src/commands/git/push.rs
```diff
@@ -375,13 +376,46 @@ pub fn cmd_git_push(
...
+    if !push_stats.all_ok() {
+        let mut error = user_error("Failed to push some bookmarks:");
+        if !push_stats.rejected.is_empty() {
+            error.add_formatted_hint_with(|formatter| {
+                for reference in &push_stats.rejected {
+                    write!(formatter.labeled("reference"), "{reference}")?;
+                    writeln!(formatter, ": unexpectedly moved on the remote")?;
+                }
+                Ok(())
+            });
```
yuja, line 401: Since this can be multi-line output, we'll need to insert header line so the list items will be aligned. Something like:
```
Hint: These refs couldn't be pushed:
  refs/heads/foo: blah blah (reason ...)
  refs/heads/bar...
```

cli/src/commands/git/push.rs
```diff
@@ -375,13 +376,46 @@ pub fn cmd_git_push(
...
     Ok(())
 }
 
+fn process_push_stats(push_stats: &GitPushStats) -> Result<(), CommandError> {
+    if !push_stats.all_ok() {
+        let mut error = user_error("Failed to push some bookmarks:");
+        if !push_stats.rejected.is_empty() {
+            error.add_formatted_hint_with(|formatter| {
+                for reference in &push_stats.rejected {
+                    write!(formatter.labeled("reference"), "{reference}")?;
```
yuja, line 393: nit: `reference` -> `git_ref` (there's no "reference" label defined in colors.toml)

lib/src/git.rs
```diff
@@ -2336,7 +2349,9 @@ pub fn push_branches(
             new_target: update.new_target.clone(),
         })
         .collect_vec();
-    push_updates(mut_repo, git_settings, remote, &ref_updates, callbacks)?;
+
+    let push_stats = push_updates(mut_repo, git_settings, remote, &ref_updates, callbacks)?;
+    tracing::debug!(?push_stats);
```
yuja, line 2354: It might be better to return early if `push_stats` contained rejected refs. The current implementation doesn't handle partially-pushed refs, and we don't have tests for that. It's also wrong to not update `mut_repo` at all, but I think explicit `if failed { return }` makes the problem more obvious.

lib/src/git.rs
```diff
@@ -2508,33 +2526,53 @@ fn git2_push_refs(
...
                     .sorted()
-                    .map(|name| name.to_string())
+                    .map(|x| x.to_string())
                     .collect(),
-            ))
+                ..Default::default()
+            }
+        } else {
+            // remote rejected refs because of remote config (e.g., no push on main)
+            GitPushStats {
```
yuja, line 2571: Does it mean no refs are pushed if `remote_rejected` is not empty? (Just asking because I haven't checked the actual behavior.)

lib/src/git.rs
```diff
@@ -2492,14 +2509,15 @@ fn git2_push_refs(
...
                 Ok(())
             } else {
                 Err(git2::Error::from_str("failed push negotiation"))
             }
         });
         callbacks.push_update_reference(|refname, status| {
-            // The status is Some if the ref update was rejected
+            // The status is Some if the ref update was rejected by the remote
             if status.is_none() {
                 remaining_remote_refs.remove(refname);
```
yuja, line 2522: nit: might be simpler to move `refname` to `pushed_refs` here (instead of expecting that all `qualified_remote_refs_expected_locations` are pushed.)

lib/src/git.rs
```diff
@@ -2555,23 +2593,10 @@ fn subprocess_push_refs(
...
-        let mut rejected_refs = push_stats.remote_rejected;
-        rejected_refs.sort();
-        // remote rejected refs because of remote config (e.g., no push on main)
-        Err(GitPushError::RefUpdateRejected(rejected_refs))
-    } else {
-        Ok(())
-    }
+    let mut push_stats = git_ctx.spawn_push(remote_name, &refs_to_push, &mut callbacks)?;
+    push_stats.rejected.sort();
+    push_stats.remote_rejected.sort();
```
yuja, line 2601: nit: sort `push_stats.pushed` too? (or these lists might be sorted by `git` CLI, but I'm not sure.)

yuja (commented):
> Ended up splitting the output checking

That sounds good to me.

lib/src/git_subprocess.rs
```diff
@@ -489,14 +489,23 @@ fn parse_ref_pushes(stdout: &[u8]) -> Result<GitPushStats, GitSubprocessError> {
...
             }
             // ! for a ref that was rejected or failed to push; and
             b"!" => {
-                if summary.starts_with_str("[remote rejected]") {
-                    push_result.remote_rejected.push(reference);
+                if let Some(reason) = summary.strip_prefix(b"[remote rejected]") {
+                    let reason = reason
+                        .strip_prefix(b" (")
+                        .and_then(|r| r.strip_suffix(b")"))
+                        .map(|x| x.to_str_lossy().to_string());
```
yuja, line 500: nit: `.into_owned()`

lib/src/git.rs
bsdinis: You're right. We still should remove it from the remaining, as the ones left become the remote_rejected

lib/src/git.rs
bsdinis: Right, this is buggy: the only reason the push will fail is if the negotiation fails, not if the remote rejects some refs

lib/src/git.rs
bsdinis: I inverted it (only updating the `mut repo` if the push stats are ok. I still think it makes it pretty obvious

cli/src/commands/git/push.rs
bsdinis: I achieved this somewhat, it looks more like:
```
        Error: Failed to push some bookmarks:
        Hint: The following references unexpectedly moved on the remote:
          refs/heads/bookmark1 (reason: stale info)
        Hint: Try fetching from the remote, then make the bookmark point to where you want it to be, and
```

cli/src/commands/git/push.rs
```diff
@@ -375,13 +376,62 @@ pub fn cmd_git_push(
...
         git::push_branches(tx.repo_mut(), &git_settings, remote, &targets, cb)
     })?;
+    process_push_stats(&push_stats)?;
     tx.finish(ui, tx_description)?;
     Ok(())
 }
 
+fn process_push_stats(push_stats: &GitPushStats) -> Result<(), CommandError> {
+    if !push_stats.all_ok() {
+        let mut error = user_error("Failed to push some bookmarks:");
```
yuja, line 389: nit: remove `:` which no longer makes sense. (and period should be omitted in error message.)

cli/src/commands/git/push.rs
```diff
@@ -375,13 +376,62 @@ pub fn cmd_git_push(
...
+        if !push_stats.rejected.is_empty() {
+            error.add_formatted_hint_with(|formatter| {
+                writeln!(
+                    formatter.labeled("hint"),
```
yuja, line 393: nit: remove `.labeled("hint")`. The "hint" label and "Hint: " prefix is added when printing.

lib/src/git.rs
yuja, line 1: The latest changes look good, but they appear to be squashed into wrong commit. Can you fix that?

cli/src/commands/git/push.rs
bsdinis: done

cli/src/commands/git/push.rs
bsdinis: done

lib/src/git.rs
bsdinis: ah you're right. done

cli/src/commands/git/push.rs
```diff
@@ -375,13 +376,53 @@ pub fn cmd_git_push(
...
+            error.add_formatted_hint_with(|formatter| {
+                writeln!(formatter, "The remote rejected the following updates:")?;
+                for reference in &push_stats.remote_rejected {
+                    write!(formatter, "  ")?;
+                    write!(formatter.labeled("git_ref"), "{reference}")?;
+                    writeln!(formatter)?;
+                }
+                Ok(())
+            });
+            error.add_hint("Try checking if you have permission to push to all the bookmarks");
```
yuja, line 424: nit: add period to hint message.

yuja (approved):
Thanks.

PR #6076, opened 2025-03-18
conflicts: document the "same-change rule"

PhilipMetzger:
I think its time to finally give our Conflict Algebra rule a name and a Glossary term. This is will make our documentation alot easier, and gives it a name for new design docs.

All credit goes to @martinvonz for coming up with the term in #4988.

# Checklist

If applicable:

### DIFF
--- a/docs/technical/conflicts.md
+++ b/docs/technical/conflicts.md
@@ -62,4 +62,19 @@ non-conflict state `C`. We then revert that change. Reverting a change means
 applying its reverse diff `-(E-C)`, so the result is `E+(C-E) =
 (C+(B-A))+(C-(C+(B-A)))`, which we can simplify to just `C` (i.e. no conflict).
 
+## Same-change rule
+
+When all sides of a conflict make the same change,
+[we automatically consider it resolved to that value][resolve]. We call this
+"the same-change rule". This behavior matches what Git and Mercurial do.
+Darcs, on the other hand, considers it a  conflict. The automatic conflict
+resolution we do is lossy in terms of conflict algebra; it means that rebasing
+a commit onto a commit that has the same changes (or a subset thereof) and then
+rebasing it back will lose changes (for a real-life example see [bug #6369]).
+We do it because it is more user-friendly in the vast majority of cases. We
+may at some point experiment with a config option to disable the same-change
+rule.
+
+[bug #6369]: https://github.com/jj-vcs/jj/issues/6369
 [merge-rs]: https://github.com/jj-vcs/jj/blob/main/lib/src/merge.rs
+[resolve]: https://github.com/jj-vcs/jj/blob/53272510bf879086d83bb5eea1406f75ba31f138/lib/src/merge.rs#L85-L99


docs/glossary.md
```diff
@@ -181,6 +181,16 @@ function `root()`. Note that our definition of "root commit" is different from
...
 commits `jj log -r 'root()+'` will show.
 
+## Same Change Rule
+
+The same Change Rule is a pragmatic hack on the patch algebra, which allows
+us to consider a conflict as automatically resolved when two sides are same.
+This is similar to Git and Mercurial (hg) and not like Darcs and its
+descendants which would consider it a conflict. It also has some unfortunate
+consequences like when you reabase a commit to ustream and the upstream rebased
+it  to where it was before, you may lose changes (any that overlapped with
```
neongreen, line 191:
```suggestion
consequences like when you rebase a commit to upstream and the upstream rebased
it to where it was before, you may lose changes (any that overlapped with
```

docs/glossary.md
```diff
@@ -181,6 +181,16 @@ function `root()`. Note that our definition of "root commit" is different from
 Git's; Git's "root commits" are the first commit(s) in the repository, i.e. the
 commits `jj log -r 'root()+'` will show.
 
+## Same Change Rule
```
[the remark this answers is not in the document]
PhilipMetzger: done

docs/glossary.md
```diff
@@ -181,6 +181,16 @@ function `root()`. Note that our definition of "root commit" is different from
...
+
+The same Change Rule is a pragmatic hack on the patch algebra, which allows
```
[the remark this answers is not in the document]
PhilipMetzger: done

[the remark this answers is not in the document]
PhilipMetzger: > Maybe we can call it "conflict algebra" instead?

done

docs/glossary.md
```diff
@@ -181,6 +181,16 @@ function `root()`. Note that our definition of "root commit" is different from
...
+us to consider a conflict as automatically resolved when two sides are same.
```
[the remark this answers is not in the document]
PhilipMetzger: done

docs/glossary.md
PhilipMetzger: done

docs/glossary.md
```diff
@@ -181,6 +181,16 @@ function `root()`. Note that our definition of "root commit" is different from
...
 
+## same-change rule
+
+The same-change rule is a pragmatic hack on the conflict algebra, which allows
+us to consider a conflict as automatically resolved when two sides are the same.
+This is similar to Git and Mercurial (hg) and not like Darcs and its
+descendants which would consider it a conflict. It also has some unfortunate
+consequences like when you rebase a commit to upstream and the upstream rebased
+it  to where it was before, you may lose changes (any that overlapped with
+upstream's changes).
```
martinvonz, line 192: I felt a little hesitant about including this in the glossary before when I saw this PR because it doesn't seem like something users would need to know. After thinking a bit more about it, I feel like docs/technical/conflicts.md is a better place for this. What do you think?

PhilipMetzger: > After thinking a bit more about it, I feel like docs/technical/conflicts.md is a better place for this. What do you think?

Isn't the Glossary the place to make references to though? I think it deserves a outline in the conflicts doc though with a reference to this.

martinvonz: I don't think internal terms belong in the glossary. If we end up with many internal terms, I suppose we can create an internal/technical glossary, but I don't think we're at that point yet.

PhilipMetzger: Ok, I've moved it although it now needs a major rework.

docs/technical/conflicts.md
```diff
@@ -62,4 +62,14 @@ non-conflict state `C`. We then revert that change. Reverting a change means
...
 
+## same-change rule
+
+The same-change rule is a pragmatic hack on the conflict algebra, which allows
+us to consider a conflict as automatically resolved when two sides are the same.
+This is similar to Git and Mercurial (hg) and not like Darcs and its
+descendants which would consider it a conflict. It also has some unfortunate
+consequences like when you rebase a commit to upstream and the upstream rebased
+it  to where it was before, you may lose changes (any that overlapped with
+upstream's changes).
```
martinvonz, line 73: I'm not sure this is a good introduction to what it is. Here's my proposal:

When all sides of conflict make the same change, we [automatically consider it resolved to that value][resolve-same-change]. We call this "the same-change rule". This behavior matches what Git and Mercurial do. Darcs, on the other hand, considers it a conflict. The automatic conflict resolution we do is lossy in terms of conflict algebra; it means that rebasing a commit onto a commit that has the same changes (or a subset thereof) and then rebasing it back again will lose changes. We do it because it is more user-friendly in the vast majority of cases. We may at some point experiment with a config option to disable the same-change rule.

[resolve-same-change]: https://github.com/jj-vcs/jj/blob/53272510bf879086d83bb5eea1406f75ba31f138/lib/src/merge.rs#L85-L99

PhilipMetzger: done, I've also added #6369 as a real-life example

Issue #4386, opened 2024-09-04
SSH login with gpg keycard backed ssh key and pinentry-curses for pin fails


necauqua:
## Description
I know that this is a contrived setup (hence it probably wasn't reported), just making sure this is tracked.

## Steps to Reproduce the Problem
1. Have a yubikey or similar
1. Configure a gpg auth subkey to be used for SSH and export it to the hw key
1. Configure gpg to use pinentry-ncurses for card pin

## Expected Behavior
When you do fetch/push, pinentry should ask for the pin and then everything should work.

## Actual Behavior
Instead you get the classic `Error: authentication required but no callback set; class=Ssh (23); code=Auth (-16)`.

If you ssh into something once for the pin to get cached it starts working (without asking for pin, obviously).
When it was not pinentry-curses but whatever was the default it worked fine, asking for the pin the one time.

And the most interesting thing is that gpg signing works! When you snapshot or whatever for the first time after boot it asks for the pin with pinentry and works just fine. Probably because for gpg we just call the gpg binary and for ssh it's another libssh issue - so this likely may be fixed by the openssh build.

## Specifications

- Platform: NixOS
- Version: 0.20.0

tgross35:
I think this can also fail without a hardware key. I have a local signing key that is password-protected, and running `jj sign` produces the following:

```text
Internal error: Unexpected error from backend
Caused by:
1: Could not write object of type commit
2: Signing error
3: GPG failed with exit status: 2:
gpg: signing failed: Inappropriate ioctl for device
gpg: signing failed: Inappropriate ioctl for device
```

It would probably be ideal if `jj` could open the pinentry prompt if the key isn't unlocked.

As a workaround, I do a dummy signature with `echo hi | gpg --clear-sign`, which provides the password prompt then leaves the key unlocked for a while (24 hours?) during which time `jj sign` completes fine. (I'm assuming GPG has a way to temporarily unlock the key without actually signing "hi", but I have no clue what it is)

necauqua:
This is a different issue, because for me gpg actually _worked_, if you read the last paragraph of Actual Behavior ¯\\\_(ツ)_/¯

PR #6078, opened 2025-03-18
signing: add X.509 backend

### LINKED ISSUE #5856: FR: support signing commits with X.509 certificates
This feature request came up on Discord recently. I am opening this issue to track it.
Right now we only support signing commits using GnuPG or SSH (see [commit signing docs](https://jj-vcs.github.io/jj/latest/config/#commit-signing)).

[Git is shelling out to `gpgsm` for signing commits using X.509 certs.](https://github.com/git/git/blob/2d2a71ce85026edcc40f469678a1035df0dfcf57/gpg-interface.c#L106) We could do something similar by implementing a [`SigningBackend`](https://github.com/jj-vcs/jj/blob/6261d576da39e5f290b064b850012bfb3c08ad5f/lib/src/signing.rs#L96) for X.509.

## Resources
- https://github.com/github/smimesign
- https://linux.die.net/man/1/gpgsm
- [`lib/src/gpg_signing.rs`](https://github.com/jj-vcs/jj/blob/6261d576da39e5f290b064b850012bfb3c08ad5f/lib/src/gpg_signing.rs)
- [`lib/src/ssh_signing.rs`](https://github.com/jj-vcs/jj/blob/6261d576da39e5f290b064b850012bfb3c08ad5f/lib/src/ssh_signing.rs)
- [`lib/src/signing.rs`](https://github.com/jj-vcs/jj/blob/6261d576da39e5f290b064b850012bfb3c08ad5f/lib/src/signing.rs)

## Related
* https://github.com/jj-vcs/jj/issues/58

calebdw:
The adds support for X.509 certificates through the `gpgsm` program.

Closes #5856

@pylbrecht, can you test this out to see if it works for you?

Thanks!

# Checklist

If applicable:


pylbrecht:
> @pylbrecht, can you test this out to see if it works for you?

While I created the issue to track the missing feature, I am not using X.509 certs for signing myself, actually. I already reached out to the person, who was asking for X.509 signing support on Discord.

calebdw:
@pylbrecht,

try adding the following (the key fingerprint) to `~/.gnupg/trustlist.txt` (or wherever the gnupg homedir is):
```
60:C4:65:D6:FA:80:2A:5A:14:DF:B6:DD:2C:D6:96:79:D3:F4:7C:6D S
```


Note that the pinentry failure could be related to #4386

lib/src/x509_signing.rs
yuja, line 1: The implementation looks largely copied from gpg. Can we share the same backend implementation or extract/reuse the underlying functions? I think it's okay to add the backend impl to gpg_signing.rs because they're quite similar.

FWIW, I don't know if the backend should be called "x509" or "gpgsm".

calebdw: > FWIW, I don't know if the backend should be called "x509" or "gpgsm".

I went back and forth on this, I think `x509` makes it a little more clear what type of certificates are being used---I think there could be a potential for confusion if it's just called `gpgsm`. But I can rename everything if desired.

> Can we share the same backend implementation or extract/reuse the underlying functions?

I thought about doing that, but when I first started I wasn't sure how dissimilar they would be. Now that I have a working implementation I can revisit.

emilazy: `gpgsm` sounds good to me because there are other X.509 and S/MIME implementations. For instance, we don’t call the `gpg` backend `openpgp` (though Git does, funnily enough).

yuja: There are other x509 signing tools, but I don't know if we add native support for these (because `git` doesn't, so wrapper programs exist afaik.) So I think `gpgsm` is technically correct as a backend name, but I don't have a strong reason against `x509` either.

pylbrecht:
> @pylbrecht,
> 
> try adding the following (the key fingerprint) to `~/.gnupg/trustlist.txt` (or wherever the gnupg homedir is):
> 
> ```
> 60:C4:65:D6:FA:80:2A:5A:14:DF:B6:DD:2C:D6:96:79:D3:F4:7C:6D S
> ```

The entry was already there:
```shell
% cat ~/.gnupg/trustlist.txt

# O=Internet Widgits Pty Ltd
# ST=Some-State
# C=US
60:C4:65:D6:FA:80:2A:5A:14:DF:B6:DD:2C:D6:96:79:D3:F4:7C:6D S relax
```
Not sure if the "relax" at the end makes a difference.

> I was able to successfully sign the commit with the x.509 cert using:
> 
> ```shell
> gpgsm --gen-key # self-signed cert
> jj sign -r @ --config signing.backend="x509"
> ```

Then it's most likely my local setup. I just tried stitching something together to get some way of verification for a first X.509 backend prototype.

calebdw:
> Then it's most likely my local setup. I just tried stitching something together to get some way of verification for a first X.509 backend prototype.

I imagine it's your pinentry program or something like that---I was even prompted to trust the cert and enter the password when using jj to sign a commit

lib/src/x509_signing.rs
calebdw: If the backend is called `gpgsm` then it makes more sense for the code to live in `gpg_signing.rs`---I can update the name and just make it clear in the docs that this is for X.509 certs

lib/tests/test_gpg.rs
```diff
@@ -79,6 +164,62 @@ impl GpgEnvironment {
...
 }
 
+struct GpgsmEnvironment {
+    homedir: tempfile::TempDir,
+}
+
+impl GpgsmEnvironment {
+    fn new() -> Result<Self, std::process::Output> {
+        let dir = tempfile::Builder::new()
+            .prefix("jj-x509-signing-test-")
```
yuja, line 174: nit: `x509` -> `gpgsm`

lib/tests/test_gpg.rs
```diff
@@ -218,3 +379,124 @@ fn invalid_signature() {
...
+    -----END SIGNED MESSAGE-----
+    ";
+    assert_matches!(
+        backend.verify(b"hello world", signature),
+        Err(SignError::InvalidSignatureFormat)
+    );
+    assert_matches!(
+        backend.verify(b"so bad", signature),
+        Err(SignError::InvalidSignatureFormat)
+    );
```
yuja, line 477: nit: can you leave a comment saying that this is actually no certificates found error or something like that?

As far as I understand it, `gpgsm` doesn't have `NO_PUBKEY` status code or equivalent. We'll probably need to rename the error `InvalidSignatureFormat` because there are other kind of errors mapped to this.

yuja (approved):
Looks good compared to the git implementation. I don't have expertise on signing, though.

docs/config.md
```diff
@@ -1184,6 +1184,36 @@ To consider expired keys as invalid you can set:
 backends.gpg.allow-expired-keys = false
 ```
 
+#### X.509 Certificates
+
+X.509 certificates can be used to sign commits using the `gpgsm` backend.
```
botovq, line 1189: Again, this doesn't seem right.

lib/tests/test_gpg.rs
```diff
@@ -33,6 +34,90 @@ lBqJFFP2PqzT0CRfh94HSIMM
...
 
+static GPGSM_FINGERPRINT: &str = "8F08B1D7DDEEFA91C25F8681F6494FF2E03DEADE";
+
+static GPGSM_PRIVATE_KEY: &str = r#"Issuer ...: /CN=JJ/O=X509 Signing Test
+Serial ...: 4C2B796B46814D39
+Subject ..: /CN=JJ/O=X509 Signing Test
+    aka ..: someone@example.com
+Keygrip ..: E595CEA8D5E74F60DEDF96753E041F64157DFBF6
+
+-----BEGIN PKCS12-----
```
botovq, line 45: I guess it's fine for a test object, but, really, no one should be using `pbewithSHAAnd40BitRC2-CBC` and `pbeWithSHAAnd3-KeyTripleDES-CBC` these days. This suggests tooling with poor defaults was used (such as old openssl). Perhaps this means that someone with the relevant expertise with the tools should provide some guidance how to avoid this.

botovq (commented):
I'm not an expert on PKCS#12 and I'm unfamiliar with gpgsm, but a couple of things didn't look right here at a quick glance.

lib/tests/test_gpg.rs
emilazy: My experience as a distro packager says: tests with bad crypto inevitably break as upstream defaults get tightened. Please do fix this. Consider RFC 9500.

cli/src/config-schema.json
```diff
@@ -651,6 +651,21 @@
...
+                            "type": "object",
+                            "properties": {
+                                "program": {
+                                    "type": "string",
+                                    "description": "Path to the gpgsm program to be called",
+                                    "default": "gpgsm"
+                                },
+                                "allow-expired-keys": {
+                                    "type": "boolean",
+                                    "description": "Whether to consider signatures generated with an expired key as invalid",
```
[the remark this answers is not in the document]
calebdw: I've updated the language to make it more clear

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj sign` can now sign with X.509 certificates through the `gpgsm` backend.
```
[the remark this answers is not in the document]
calebdw: I've updated to refer to PKCS#12 instead of X.509

botovq (commented):
This looks a lot better to me now. Thanks!

I still wonder if it wouldn't be worthwhile to point out somewhere in the docs that it is easy to end up with very weakly protected PKCS#12 files if one isn't careful. This really isn't ideal seeing as the signing is supposed to yield significant proof of authorship.

calebdw:
> I still wonder if it wouldn't be worthwhile to point out somewhere in the docs that it is easy to end up with very weakly protected PKCS#12 files if one isn't careful. This really isn't ideal seeing as the signing is supposed to yield significant proof of authorship.

I'd rather not because that's getting really down into the weeds of key/cert generation that we really don't care about. You could say the same thing for legacy gpg and ssh key types. I imagine most people that might use this are not just generating a pkcs key / self-signed cert for themselves but have rather been issued a certificate by their central authority (e.g., company, corporate office, government office, etc.).

botovq:
Understood. It's a rather sad state of things, but I agree that it's probably not jj's docs' place to try and fix this. After all, it's doing a great job of removing a lot of legacy pain :)

PR #6081, opened 2025-03-18
dev: add mise config

neongreen:
This replaces #5636.

After setting up [Mise](https://mise.jdx.dev), we can use `mise run` to execute tasks. Mise also takes care of installing `uv`.

I didn't add rust/cargo to mise tools, because the mise repo itself doesn't do that. Maybe they have a good reason.

TODO:

- [ ] Try everything on Windows

# Checklist

If applicable:


mise.toml
```diff
@@ -0,0 +1,63 @@
...
+description = "Check GitHub workflows with Zizmor"
+run = "uvx zizmor ."
+
+[tasks."test"]
+description = "Run tests"
+run = [
+  "cargo nextest run --workspace",
+  # TODO: this seems to run the same tests as the previous command,
+  # plus the insta tests. Do we need both?
+  "cargo insta test --workspace --test-runner nextest",
```
neongreen, line 50: I don't understand nextest vs insta

mise.toml
```diff
@@ -0,0 +1,63 @@
...
+run = [
+  # Check for cmake
+  "cmake --version || echo 'cmake is not installed, cargo clippy might fail'",
+  # `cargo +stable` doesn't check for the latest stable, so we have to discover it.
+  # We just extract versions from the output of `rustup check` and take the latest.
+  '''
+  STABLE=$(rustup check | grep '^stable' | grep -Eo '\d+\.\d+\.\d+' | sort --version-sort | tail -1)
+  cargo +$STABLE clippy --all-features --workspace --all-targets -- -D warnings
+  ''',
+]
```
neongreen, line 38: This likely won't work on Windows but I can use `run_windows = ...` to have a different version specifically for Windows

Cargo.toml
```diff
@@ -7,7 +7,7 @@ members = ["cli", "lib", "lib/gen-protos", "lib/proc-macros", "lib/testutils"]
 [workspace.package]
 version = "0.27.0"
 license = "Apache-2.0"
-rust-version = "1.84" # NOTE: remember to update CI, contributing.md, changelog.md, and install-and-setup.md
+rust-version = "1.84" # NOTE: remember to update rust-toolchain.toml, CI, contributing.md, changelog.md, and install-and-setup.md
```
neongreen, line 10: I see rust-toolchain.toml was removed in https://github.com/jj-vcs/jj/pull/5090/files so I'll remove it here as well

neongreen:
fwiw i've been using mise for several months by now (thanks to @arxanas's rec? or someone's rec?) and i like it a lot

PhilipMetzger (commented):
minor stuff, still split on it if we actually want it,

mise.toml
```diff
@@ -0,0 +1,62 @@
...
+uv = "0.5.1"
+
+[tasks."setup"]
+description = "Set up the dev environment (install Rust, etc)"
+run = [
+  "rustup toolchain install 1.84",
+  # We want rustfmt from nightly
+  "rustup toolchain install nightly --profile minimal --component rustfmt",
+  # Tools installed via Cargo
+  "cargo +1.84 install --locked bacon",
```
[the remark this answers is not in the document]
arxanas: You can also use `mise` itself to watch files and run tasks, if that's appropriate here?

mise.toml
```diff
@@ -0,0 +1,62 @@
...
+[tasks."format"]
+description = "Format the code"
+run = "cargo +nightly fmt"
+
+[tasks."lint"]
+description = "Lint the code"
+run = "cargo +1.84 clippy --workspace --all-targets"
+
+[tasks."lint:ci"]
+description = "Lint the code using the same clippy version as CI would"
```
arxanas, line 27: comment (non-blocking): If we wanted, we could also literally use `mise` in CI 👀. You could use https://github.com/jdx/mise-action to cache all of the tool versions as well.

mise.toml
arxanas: comment (non-blocking): Personally, I don't think it's that important to implement Windows task-running up-front, unless we're regressing on the Windows developer platform support somehow.

mise.toml
```diff
@@ -0,0 +1,62 @@
...
+
+[tasks."format"]
+description = "Format the code"
+run = "cargo +nightly fmt"
+
+[tasks."lint"]
+description = "Lint the code"
+run = "cargo +1.84 clippy --workspace --all-targets"
+
+[tasks."lint:ci"]
```
arxanas, line 26: suggestion: The usage of `:` is semantically-meaningful since it separates wildcard boundaries (see [Task grouping](https://mise.jdx.dev/tasks/running-tasks.html#task-grouping)).

If you run `mise run lint:*`, you might expect that to run "all" lints, but it'll actually exclude the most common linting use-case. I would try to organize the groups such that they're hierarchical in some sense to make it a little more intuitive.

suggestion (non-blocking): I've been organizing my tasks roughly by function rather than domain, like `build:*`, `check:*`, `fix:*`, etc.
- Then I also rely on mise's parallelization features (after manually, painstakingly setting the dependencies between tasks).
- In particular, it's useful to separate read-only tasks from read-write tasks, since you don't want writers to clobber each other when running in parallel.

Here's one possible organization, but I don't feel strongly, and I don't mean to suggest that the unimplemented commands below need to be implemented:

- `build:cli-reference`
- `build:docs`
- `build:debug`
- `build:release`
- `check:lint`
- `check:format`
- `check:test`
- `check:zizmor`
- `ci:lint`
- `fix:format`
- `fix:lint`
- `serve:docs`

mise.toml
```diff
@@ -0,0 +1,62 @@
...
+  "cmake --version || echo 'cmake is not installed, cargo clippy might fail'",
+  # `cargo +stable` doesn't check for the latest stable, so we have to discover it.
+  # We just extract versions from the output of `rustup check` and take the latest.
+  '''
+  STABLE=$(rustup check | grep '^stable' | grep -Eo '\d+\.\d+\.\d+' | sort --version-sort | tail -1)
+  cargo +$STABLE clippy --all-features --workspace --all-targets -- -D warnings
+  ''',
+]
+
+[tasks."zizmor"]
```
arxanas, line 39: suggestion: `zizmor` is the specific tool name, while the rest of the task names are pretty abstract. I'd rename this to perhaps `check:workflows` (as per above), or rename other stuff to e.g. `clippy`.

mise.toml
```diff
@@ -0,0 +1,62 @@
...
+[tasks."setup"]
+description = "Set up the dev environment (install Rust, etc)"
+run = [
+  "rustup toolchain install 1.84",
+  # We want rustfmt from nightly
+  "rustup toolchain install nightly --profile minimal --component rustfmt",
+  # Tools installed via Cargo
+  "cargo +1.84 install --locked bacon",
+  "cargo +1.84 install --locked cargo-insta",
+  "cargo +1.84 install --locked cargo-nextest",
```
arxanas, line 15: comment (non-blocking): You might be able to

- use/abuse [`sources`](https://mise.jdx.dev/tasks/task-configuration.html#sources)/[`outputs`](https://mise.jdx.dev/tasks/task-configuration.html#outputs)
- define the `cargo install` command(s) as its own task
- make the test task depend on the installation command(s)

in order to get a one-step `mise run test` install+develop experience. (It's no Nix, but it might work? I haven't specifically tried the above.)

Cargo.toml
```diff
@@ -7,7 +7,7 @@ members = ["cli", "lib", "lib/gen-protos", "lib/proc-macros", "lib/testutils"]
 [workspace.package]
 version = "0.27.0"
 license = "Apache-2.0"
-rust-version = "1.84" # NOTE: remember to update CI, contributing.md, changelog.md, and install-and-setup.md
+rust-version = "1.84" # NOTE: remember to update mise.toml, CI, contributing.md, changelog.md, and install-and-setup.md
```
arxanas, line 10:
```suggestion
rust-version = "1.84" # NOTE: remember to update CI, mise.toml contributing.md, changelog.md, and install-and-setup.md
```

nit: I feel like "CI" should go first since it's most important, in some sense.

arxanas (approved):
Overall, I'm in favor, pending the outstanding comments.

- We could potentially use `mise` to unify CI workflows as well, as I mentioned in a review comment.
- We should of course adopt at most one task-runner system, so we'd have to oust mise if we wanted to use something else later.
- I'm in favor of mise since, in addition to serving as a task runner, it also solves several cross-language devtool problems, unlike e.g. `cargo xtask`.
  - Using `mise` might concretely make the cross-language formatting problem much easier to solve (https://github.com/jj-vcs/jj/pull/3757).

mise.toml
```diff
@@ -0,0 +1,62 @@
+#:schema https://mise.jdx.dev/schema/mise.json
+
+[tools]
+uv = "0.5.1"
+
+[tasks."setup"]
+description = "Set up the dev environment (install Rust, etc)"
```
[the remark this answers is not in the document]
arxanas: Oops, I see you wrote

> I didn't add rust/cargo to mise tools, because the mise repo itself doesn't do that. Maybe they have a good reason.

I didn't see that `mise` pins the version of Rust via `rust-toolchain.toml`, either, so maybe that's the difference?

I'm in favor of pinning the Rust version, as per my [last attempt (reverted due to a GitHub Actions issue)](https://github.com/jj-vcs/jj/pull/4613), but it's fine to not do that, too, if it would be complicated to support the nightly commands we want to run. (But, as per my other comment in this thread, it looks like managing Rust/Cargo via `mise` would also make it easier to manage `cargo-insta` and `cargo-nextest`.)

mise.toml
arxanas: Oh, interestingly, `cargo insta` is itself a `mise` tool: https://github.com/jdx/mise/blob/c67be43fcede947920c8eb17afb059037235a3e0/mise.toml#L12. From trying it out, it looks like `cargo nextest` is as well.

arxanas: This diff seems to work for me in order to register all the tools directly with `mise` and support a one-command workflow:

```diff
diff --git a/mise.toml b/mise.toml
index 3d90390bc..07fd1bc73 100644
--- a/mise.toml
+++ b/mise.toml
@@ -1,27 +1,19 @@
 #:schema https://mise.jdx.dev/schema/mise.json
 
 [tools]
+"cargo:cargo-insta" = "1.42.2"
+"cargo:cargo-nextest" = "0.9.93"
+rust = "1.84"
 uv = "0.5.1"
 
-[tasks."setup"]
-description = "Set up the dev environment (install Rust, etc)"
-run = [
-  "rustup toolchain install 1.84",
-  # We want rustfmt from nightly
-  "rustup toolchain install nightly --profile minimal --component rustfmt",
-  # Tools installed via Cargo
-  "cargo +1.84 install --locked bacon",
-  "cargo +1.84 install --locked cargo-insta",
-  "cargo +1.84 install --locked cargo-nextest",
-]
-
 [tasks."format"]
 description = "Format the code"
-run = "cargo +nightly fmt"
+tools.rust = "nightly"
+run = "cargo fmt"
 
 [tasks."lint"]
 description = "Lint the code"
-run = "cargo +1.84 clippy --workspace --all-targets"
+run = "cargo clippy --workspace --all-targets"
 
 [tasks."lint:ci"]
 description = "Lint the code using the same clippy version as CI would"
@@ -42,16 +34,11 @@ run = "uvx zizmor ."
 
 [tasks."test"]
 description = "Run tests"
-run = [
-  "cargo +1.84 nextest run --workspace",
-  # TODO: this seems to run the same tests as the previous command,
-  # plus the insta tests. Do we need both?
-  "cargo +1.84 insta test --workspace --test-runner nextest",
-]
+run = "cargo insta test --workspace --test-runner nextest"
 
 [tasks."update-cli-help"]
 description = "Update generated CLI help (cli/tests/cli-reference@.md.snap)"
-run = "cargo +1.84 insta test --accept --workspace -- test_generate_md_cli_help"
+run = "cargo insta test --accept --workspace -- test_generate_md_cli_help"
 
 [tasks."docs:build"]
 description = "Build documentation into rendered-docs/ for offline use"
```

Note: It looks like `uv` is installed even though it's not necessary except for the `uv` tasks.
- It might work to declare individual `tools` for the `uv` commands, which presumably would cause them to lazily provision the tools.
- Perhaps the `uv` version could be deduplicated between the two commands via `vars`, rather than inlined.
- (Hmm, if only there were a system which could represent a fine-grained, cross-language build graph...)

mise.toml
PhilipMetzger: > Personally, I don't think it's that important to implement Windows task-running up-front, unless we're regressing on the Windows developer platform support somehow.

Not implementing it would be a regression to the current state for Windows, since at least I wouldn't want a tool only supported by a single platform.

mise.toml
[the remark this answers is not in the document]
ilyagr: I was wondering whether `mise` could support using a different version of Rust for `cargo fmt` than for everything else.

mise.toml
```diff
@@ -0,0 +1,62 @@
...
+run = [
+  "rustup toolchain install 1.84",
```
ilyagr, line 9: (Non-blocking) This might be controversial, but I'd just use stable. Rust is getting pretty good at backward-compatibility, e.g. clippy is aware of which version of Rust is the MSRV and doesn't suggest inappropriate hints. OTOH, refactoring tools and clippy itself get better in newer versions, as does compilation speed, etc.

mise.toml
ilyagr: Thanks! I was wondering about similar questions and how far it can go with `mise`.

mise.toml
ilyagr: Currently, we have the advantage that Windows dev instructions are pretty much the same as all other dev instructions. It'd be nice if we could configure mise so that the same workflow, and a similar CI config, work for Windows and the other platforms.

OTOH, I wonder how doable that is. Looking through mise's docs, it worried me a bit that it seems to have a lot of "but"s when Windows support is discussed.

emilazy: FWIW I’d rather not involve this in CI: I’ve been trying to cut down on the number of bootstrap steps and separate installations to speed up the cycle time (not a huge amount, but every bit helps), and on Windows we have to do some rewiring to get stuff on the fast `D:\` drive already that I’d have to figure out how to replicate with `mise`. Meanwhile the benefit seems limited for the actual builds, where we only bootstrap a Rust toolchain, `cargo-nextest`, and Taplo, and I have a WIP to get rid of the latter. (Of course part of this is just that it’s not my preferred tool so I wouldn’t want it to become load‐bearing, so feel free to add appropriate grains of salt to account for that.)

arxanas: > A possibility is some combination of mise and cargo xtask (Rust instead of shell sctipts).

I'd definitely prefer to only use one or the other, if possible.

> FWIW I’d rather not involve this in CI:

Sounds reasonable. Let's not put `mise` in CI (unless it were to suddenly turn out that putting `mise` in CI makes it much faster somehow).

mise.toml
arxanas: As per https://github.com/jj-vcs/jj/pull/6081/files#r2020168644, we don't want/need to use `mise` in CI directly.

mise.toml
neongreen: - I don’t think mise should be in CI either 
- re/ shell scripts — mise supports anything. if you can shebang it you can make it a mise task (can you shebang rust scripts tho?)
- this said, imo the best quality of a shell scripting language is “can copilot write the entire script for me please” and I don’t know how well rust fares here. I know powershell is atrocious, python and bash are ok. But it’s not my call to make of course.

neongreen:
Please don’t rerereview yet, I finally got around to setting up a windows VM and I want to check if everything works first

mise.toml
neongreen: The "lint like CI" thing will become a separate PR (if at all)

mise.toml
neongreen: I didn't want to set up rust via mise since even the author of mise doesn't do that. But we can try it out and see what happens

I'll try out your patch on mac and windows

neongreen: re/ vars -- they don't work in task tools, see https://github.com/jdx/mise/discussions/4885

neongreen: re/ rust versions -- different versions per tool seem to work just fine unless global `tools.rust` is set

neongreen:
> I’m sorry, I don’t have the energy to work on finishing this in the nearest future. If anyone wants to revisit this later, feel free to use my branch

closing for the same reason

PR #6091, opened 2025-03-19
cli: remove dumb suffix from sideband writer

bsdinis:
The `GitSidebandProgressMessageWriter` was ported from `git`'s implementation:
https://github.com/git/git/blob/43072b4ca132437f21975ac6acc6b72dc22fd398/sideband.c#L138

This implementation includes a `DUMB_SUFFIX` of whitespace, to be used in alternative to a terminal escape code.

However, there's no good reason for this suffix, as far as we can see. This was also messing up with some snapshots.

# Checklist

If applicable:

### DIFF
--- a/cli/src/git_util.rs
+++ b/cli/src/git_util.rs
@@ -178,7 +178,8 @@ impl GitSidebandProgressMessageWriter {
 
         GitSidebandProgressMessageWriter {
             display_prefix: "remote: ".as_bytes(),
-            suffix: if is_terminal { "\x1B[K" } else { "        " }.as_bytes(),
+            // \x1B[K is a terminal code to clear the terminal
+            suffix: if is_terminal { "\x1B[K" } else { "" }.as_bytes(),
             scratch: Vec::new(),
         }
     }
--- a/cli/tests/test_git_push.rs
+++ b/cli/tests/test_git_push.rs
@@ -2346,7 +2346,7 @@ fn test_git_push_rejected_by_remote() {
     ------- stderr -------
     Changes to push to origin:
       Move forward bookmark bookmark1 from d13ecdbda2a2 to dd5c09b30f9f
-    remote: error: hook declined to update refs/heads/bookmark1        
+    remote: error: hook declined to update refs/heads/bookmark1
     Error: Remote rejected the update of some refs (do you have permission to push to ["refs/heads/bookmark1"]?)
     [EOF]
     [exit status: 1]


yuja:
What's the problem of snapshots?

AFAICT, it is a hack to overwrite the previous line that could be slightly longer than the current line. I think it's still valid in jj. We don't provide a handy option to forcibly enable sideband progress output, though.

bsdinis:
> What's the problem of snapshots?
> 
> AFAICT, it is a hack to overwrite the previous line that could be slightly longer than the current line. I think it's still valid in jj. We don't provide a handy option to forcibly enable sideband progress output, though.

But it's only valid up to 8 characters? I think this is needed in git for the progress updates, which we parse separately.

yuja:
> But it's only valid up to 8 characters?

Yes.

> I think this is needed in git for the progress updates, which we parse separately.

Oh, I thought it would be reverted because it seemed inconsistent to intercept part of remote progress messages. Still, I think we forward some remote progress messages ending with `\r`.

bsdinis:
> > But it's only valid up to 8 characters?
> 
> Yes.
> 
> > I think this is needed in git for the progress updates, which we parse separately.
> 
> Oh, I thought it would be reverted because it seemed inconsistent to intercept part of remote progress messages. Still, I think we forward some remote progress messages ending with `\r`.

What messages ending with \r?

yuja:
> What messages ending with \r?

Maybe "Enumerating objects: N", for example? Side-band messages are free-form text, so we can't reliably map them to our `GitProgress`. Git forges can send non-standard progress messages as they want. I think we can merge sideband/progress handlers in a way that free-form progress messages can be supported, but that would be easier to implement after the git2 removal.

bsdinis:
Good point. Maybe we could keep track of the max size of non `\n`-terminated  lines (since the last `\n`) and only pad for those cases?

yuja:
I wouldn't make it complicated further. (If more padding were needed, the message would be visible in `git` anyway.)

What's the problem you're trying to fix?

bsdinis:
> I wouldn't make it complicated further. (If more padding were needed, the message would be visible in `git` anyway.)
> 
> What's the problem you're trying to fix?

We now have 1 snapshot with a remote message, and the padding is very annoying, especially if the editor removes trailing whitespace.

yuja:
I would fix editor settings and leave trailing spaces in snapshots, but we can also strip them by `output.normalize_with()` (or insta filters.) I think that's good enough?

yuja:
> Removing the trailing whitespace is a common enough option (would classify it as a feature, not a bug).

Just FYI, `.editorconfig` is set to not trim whitespace for `*.rs` files. Your editor might have an option to respect it.

PR #6094, opened 2025-03-20
config: move .cargo/config-ci.toml to .config/cargo-ci-config.toml

ilyagr:
The goal is simply to have one fewer top-level dir. I could also move it to .github/something


emilazy:
Cargo [somewhat recommends keeping these in `.cargo`](https://doc.rust-lang.org/1.85.0/cargo/reference/config.html#config-relative-paths) because of how relative paths work. It does suggest `.config` as an alternative, but I don't love that directory as it is inconsistent with how it works under XDG in `$HOME` (where subdirectories are used). I think people will expect to find settings that could affect Cargo behaviour under `.cargo`.

PR #6104, opened 2025-03-21
cli: add short aliases for git fetch/pull

calebdw:
With the frequency that I execute `jj git fetch` and `jj git pull`, I find myself wishing I could just type `jj g f` and `jj g p` instead.

Thanks!

# Checklist

If applicable:


yuja:
We'll need to decide this first: https://github.com/jj-vcs/jj/pull/5870

(fwiw, I personally don't like single-char commands, and it would be nice if we can add support for nested user aliases such as `aliases.'git f' = ['git', 'fetch']`.

cenviity (changes requested):
`s/pull/push/` in commit message

PR #6121, opened 2025-03-23
index: don't preserve commits not referred to by operations/views

yuja:
~I originally considered it would be a bad idea to ignore predecessors that don't exist in the store because doing that could shadow potential data corruption issues. However, we have index built from the op log, which knows all reachable commits. So we can ask index whether the predecessor id is supposed to exist. If a predecessor exists in the index but not in the Git object store, the repository is corrupted.~

~https://github.com/jj-vcs/jj/pull/4953#issuecomment-2495735683~

~In this implementation, we assume that predecessor commits are reachable from at least one of the historical views. Some commands (e.g. squash into descendants) may create transitive commits within an operation. In that case, the predecessor chain is initially preserved, but will be lost after reindexing. We might have to fix these commands to not record transitive predecessors.~

Note that this change should be considered forward-incompatible change. We can't easily revert this patch because the stored commits will have unreachable predecessors once we run "jj op abandon && jj util gc".

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* `jj op abandon` now discards previous versions of a change (or predecessors)
+  if they become unreachable from the operation history. The evolution history
+  is truncated accordingly.
+
+  Once `jj op abandon` and `jj util gc` are run in a repository, old versions of
+  `jj` might get "commit not found" error on `jj evolog`.
+
 ### Deprecations
 
 ### New features
--- a/cli/src/commands/operation/abandon.rs
+++ b/cli/src/commands/operation/abandon.rs
@@ -37,8 +37,10 @@ use crate::ui::Ui;
 /// To discard recent operations, use `jj op restore <operation ID>` followed
 /// by `jj op abandon <operation ID>..@-`.
 ///
-/// The abandoned operations, commits, and other unreachable objects can later
-/// be garbage collected by using `jj util gc` command.
+/// Previous versions of a change (or predecessors) are also discarded if they
+/// become unreachable from the operation history. The abandoned operations,
+/// commits, and other unreachable objects can later be garbage collected by
+/// using `jj util gc` command.
 #[derive(clap::Args, Clone, Debug)]
 pub struct OperationAbandonArgs {
     /// The operation or operation range to abandon
--- a/cli/src/commands/util/gc.rs
+++ b/cli/src/commands/util/gc.rs
@@ -27,9 +27,6 @@ use crate::ui::Ui;
 ///
 /// To garbage-collect old operations and the commits/objects referenced by
 /// them, run `jj op abandon ..<some old operation>` before `jj util gc`.
-///
-/// Previous versions of a change that are reachable via the evolution log are
-/// not garbage-collected.
 #[derive(clap::Args, Clone, Debug)]
 pub struct UtilGcArgs {
     /// Time threshold
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -1709,7 +1709,7 @@ To discard old operation history, use `jj op abandon ..<operation ID>`. It will 
 
 To discard recent operations, use `jj op restore <operation ID>` followed by `jj op abandon <operation ID>..@-`.
 
-The abandoned operations, commits, and other unreachable objects can later be garbage collected by using `jj util gc` command.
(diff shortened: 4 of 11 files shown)


emilazy:
I think I support this, but there’s something I’m not sure I understand: if we go with the “turn `jj evolog` into filtered `jj op log`” idea, is there any need to store predecessors at all? Can’t that information be inferred from the operation log (potentially with a cache)?

PhilipMetzger:
I also like the idea and its potential implications for transferring objects over the wire (https://github.com/jj-vcs/jj/pull/4953#issuecomment-2496065806). 

> if we go with the “turn `jj evolog` into filtered `jj op log`” idea, is there any need to store predecessors at all? Can’t that information be inferred from the operation log (potentially with a cache)?

Aren't the local predecessors already acting as a  cache for cloud-based `jj` implementations such as Google? So they could be retained for correctness if we want to go that way.

bnjmnt4n:
> I think I support this, but there’s something I’m not sure I understand: if we go with the “turn `jj evolog` into filtered `jj op log`” idea, is there any need to store predecessors at all? Can’t that information be inferred from the operation log (potentially with a cache)?

If an operation has an abandoned commit and a new commit which share the same change ID, we could infer that the abandoned commit was the predecessor of the new commit. However, if we use a command that creates a new commit which doesn't have the same change ID as the abandoned commit, we wouldn't be able to trace this. So we would still need a way to store predecessors for this use case. (An alternative might be to store predecessors as part of the operation, but I'm not sure what that would impact.)

emilazy:
Makes sense. It does seem nicer to store it in the operation to me, as an expression of “this operation evolved commit … into …”.

martinvonz:
> However, if we use a command that creates a new commit which doesn't have the same change ID as the abandoned commit, we wouldn't be able to trace this. So we would still need a way to store predecessors for this use case.

Do you mean for things like `jj duplicate`? Maybe it's okay if we don't trace those?

yuja:
I think `squash` is one example that will benefit from predecessors (or that kind of information stored in op log.)

emilazy:
Right. Having this information stored in some form seems sensible, but having it in the actual commits seems sorta weird. (You can imagine arriving at the exact same commit, metadata included, through two different paths – should the hash be different because of the “memory“ of the predecessors?)

yuja:
> Having this information stored in some form seems sensible, but having it in the actual commits seems sorta weird. (You can imagine arriving at the exact same commit, metadata included, through two different paths

You mean weird in op-log-based evolution log?

To be clear, I also think it would probably be better to store predecessors in operation. I hope predecessors in the commit metadata will be phased out in future.

emilazy:
Just conceptually weird. It sounds like we’re on the same page about what the future state should look like, so this PR seems good to me. (I just wasn’t sure if it was an interim workaround or intended to result in a more permanent design.)

ilyagr:
I'd prefer waiting another month, mainly so that `gg` has a chance to catch up. IIUC, the prerequisite is in 0.30, so I'd put this in 0.32 (or 0.33, but that might make rebasing this PR really annoying) if it's not too inconvenient.

(But if it *does* make things a lot simpler to merge this for the next release, I don't think we absolutely *have* to wait longer)

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* `jj op abandon` now discards previous versions of a change (or predecessors)
+  if they become unreachable from the operation history. The evolution history
+  is truncated accordingly.
```
ilyagr, line 17: Perhaps mention that "once this feature is used in a repository, jj 0.29 and earlier will no longer be able to work with that repository". Or is the problem with jj<0.30 going to be subtler than that?

yuja:
> I'd prefer waiting another month, mainly so that `gg` has a chance to catch up. IIUC, the prerequisite is in 0.30, so I'd put this in 0.32 (or 0.33, but that might make rebasing this PR really annoying) if it's not too inconvenient.

I don't think tooling (including `jj` itself) would matter. AFAIK, the main concern was whether we're confident about the new data model. `jj evolog` would get an error if predecessors were truncated, but that's only after the user abandoned old operation data. I don't think that's common enough to wait for another couple of releases.

That said, we should of course delay merging of this PR if we aren't sure about the op-log-based evolution log.

ilyagr:
I see, if older versions can generally still read the repo, `gg` shouldn't be an issue. 

As for how much we trust the model, I haven't been thinking about it, and I'm not sure how long it would take me to get up to date. This also means that I have no reason to mistrust it.

In any case, now I understand what this is a bit better.

CHANGELOG.md
yuja: Updated

PR #6123, opened 2025-03-23
git_backend: don't allow single tree in `jj:trees` header

scott2000:
It is important for this case to be an error, because otherwise it would be possible to construct a non-conflicted commit which appears to have a different tree when viewed using `jj` than when viewed using Git. This could potentially be used to hide malicious code in commits in such a way that on GitHub, the code would appear normal, but it would become malicious when cloned using `jj`.

Prior to f7b14beacc678a9b351ae65224df43a96aa26392, if a commit had a `jj:trees` header with only a single tree, it would result in a panic of "root tree should have been initialized to a legacy id". This commit restores the error behavior by adding an explicit check for this case.

# Checklist

If applicable:

### DIFF
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -92,7 +92,7 @@ const CHANGE_ID_LENGTH: usize = 16;
 const NO_GC_REF_NAMESPACE: &str = "refs/jj/keep/";
 const CONFLICT_SUFFIX: &str = ".jjconflict";
 
-const JJ_TREES_COMMIT_HEADER: &[u8] = b"jj:trees";
+pub const JJ_TREES_COMMIT_HEADER: &[u8] = b"jj:trees";
 
 #[derive(Debug, Error)]
 pub enum GitBackendInitError {
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -496,7 +496,10 @@ fn root_tree_from_header(git_commit: &CommitRef) -> Result<Option<MergedTreeId>,
                 }
                 tree_ids.push(tree_id);
             }
-            if tree_ids.len() % 2 == 0 {
+            // It is invalid to use `jj:trees` with a non-conflicted tree. If this were
+            // allowed, it would be possible to construct a commit which appears to have
+            // different contents depending on whether it is viewed using `jj` or `git`.
+            if tree_ids.len() == 1 || tree_ids.len() % 2 == 0 {
                 return Err(());
             }
             return Ok(Some(MergedTreeId::Merge(Merge::from_vec(tree_ids))));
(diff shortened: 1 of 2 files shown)


martinvonz:
> It is important for this case to be an error, because otherwise it would be possible to construct a non-conflicted commit which appears to have a different tree when viewed using `jj` than when viewed using Git.

FWIW, that's still going to be possible by using `jj:trees X X X` where `X` is the malicious tree.

scott2000:
> FWIW, that's still going to be possible by using `jj:trees X X X` where `X` is the malicious tree.

Somewhat surprisingly, this actually results in a commit that is shown as conflicted, but with an empty list of conflicted files, and `jj new` on the commit makes a new child commit which automatically resolves the conflict as `X`.

martinvonz:
Yes, it's a degenerate state that's not supposed to be recorded. I agree that the `conflict` label in `jj log` etc. helps draw attention to it.

lib/src/git_backend.rs
```diff
@@ -499,7 +499,14 @@ fn root_tree_from_header(git_commit: &CommitRef) -> Result<Option<MergedTreeId>,
             if tree_ids.len() % 2 == 0 {
```
yuja, line 499: nit: just add `!= 1` here?

lib/src/git_backend.rs
```diff
@@ -499,7 +499,14 @@ fn root_tree_from_header(git_commit: &CommitRef) -> Result<Option<MergedTreeId>,
...
                 return Err(());
             }
-            return Ok(Some(MergedTreeId::Merge(Merge::from_vec(tree_ids))));
+            let merged_tree_ids = Merge::from_vec(tree_ids);
+            // It is invalid to use `jj:trees` with a non-conflicted tree. If this were
+            // allowed, it would be possible to construct a commit which appears to have
+            // different contents depending on whether it is viewed using `jj` or Git.
```
martinvonz, line 505: nit: "`jj` or `git`" seems more consistent

lib/tests/test_git_backend.rs
```diff
@@ -288,3 +293,38 @@ fn test_copy_detection() {
...
+    let new_commit_id = git_repo.write_object(&new_commit).unwrap();
+    let new_commit_id = CommitId::from_bytes(new_commit_id.as_bytes());
+
+    // Import new commit into `jj` repo. This should panic, because allowing a
+    // non-conflicted commit to have a different tree in `jj` than in Git could be
+    // used to hide malicious code.
+    let result = catch_unwind(AssertUnwindSafe(|| {
+        git_backend
+            .import_head_commits(&[new_commit_id.clone()])
+            .unwrap();
```
yuja, line 327: assert error content instead of `.unwrap()` + `catch_unwind`

scott2000: Oh, I didn't realize it was an error value instead of a panic.

martinvonz (approved):

Issue #6000, opened 2025-03-12
FR: `jj squash --restore-descendants`


FiloSottile:
**Is your feature request related to a problem? Please describe.**

A common request during code review is to split a PR/CL into multiple ones. In that case the goal is two or more changes *with the same final content*.

`jj split` is the first step in that direction. Sometimes though we need to introduce a bit of conflict to make the intermediate change build or make sense.

The natural flow is to `jj new` on the intermediate change, make the changes, `jj squash`, and then resolve manually the conflicts in the descendant, maybe checking there is no diff from the previous commit.

This is a waste of manual effort though! We know exactly how to resolve the conflicts: by restoring the contents of the original commit.

**Describe the solution you'd like**

`jj squash --restore-descendants` which preserves the descendant contents, instead of their diff.

**Describe alternatives you've considered**

1. `jj diffedit --restore-descendants` on the intermediate change does exactly what we need, but it's a diff editor environment, not a working copy.
2. `jj restore --from @ --to @- --restore-descendants` should do the same operation, but is much less intuitive.
3. Resolving the conflict with `jj restore --from OLDCOMMIT` works but is an extra step.

bnjmnt4n:
Related FR: https://github.com/jj-vcs/jj/issues/4708.

ilyagr:
I'm currently thinking of implementing these two flags for `squash`:

> 
>      `--restore-target-descendants`:
>           Preserve the content (not the diff) when rebasing descendants of the target commit
> 
>           If the target commit is a descendant of the source commit, this will have no effect.
> 
>           [aliases: restore-descendants]
> 
>       `--restore-source-descendants`:
>           Preserve the content (not the diff) when rebasing descendants of the source commit (less commonly used)
> 
>           When using `-r`, or if the target commit is an ancestor of the source commit, this will have no effect.

I think `--restore-target-descendants` is more useful, hence it is aliased to `--restore-descendants`.

I have a toy implementation that I'll share unless there are obvious bugs, but I still need to write a lot of tests.

ilyagr:
Another weird test case is `jj squash --from C --into A --restore-target-descendants` for

    X -> A -> B -> C -> D

It would modify `D` and `A`'s snapshots, but not `B`'s.

[...]

PR #6124, opened 2025-03-23
cli: rename `--restore-descendants` to `--preserve-descendant-contents` AKA `--pdc`

ilyagr:
Fixes https://github.com/jj-vcs/jj/discussions/5951

There have also been discussions on Discord on and off; people seem to agree that `--preserve-descendant-contents` is clear if verbose. I hope that a short alias might address that (though I don't intent to make a habit of having many cryptic short aliases), though we could also decide to skip it for now and see whether it's really needed.

Some other options considered:

- `keep-descendant-contents`. In theory, I like it better, but it'd be confusingly similar to `jj squash --keep-emptied` once `jj squash --keep-descendant-contents` was implemented (#6000), and would also conflict with its tab completion.
- `verbatim-descendants`
- `reparent-descendants`

The last two look a bit cryptic, depending on the person.

Note that `jj rebase` will also need a `--preserve-contents` flag for preserving the contents of the commit actually being rebased. I'm not sure whether that should also be `--pc` or not.


joyously:
It seems like `keep-contents` would not be the same as `reparent-descendants`.
I think it is still unclear what the option does. Are we talking about propagating a change to the commit content to its children, or propagating a change of position in the graph to its children? Or a mixture?
Does `rewrite` sound good?

CHANGELOG.md
```diff
@@ -21,7 +21,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   commits. The list won't show more than 10 commits to not clutter the console.
 
 * `jj unsquash` has been removed in favor of `jj squash` and
-  `jj diffedit --restore-descendants`.
+  `jj diffedit --preserve-descendant-contents`.
```
ilyagr, line 24: TODO: I'm planning to change this to `--preserve-descendant-content` (without the s at the end), unless we go with something completely different. This matches the help message better, and I don't think the `s` helps much.

ilyagr:
Once we get to `jj rebase`, which would need both `--preserve-descendant-content` and `--preserve-content` (for preserving the snapshot of the actual commits being rebased), I wonder whether `--preserve-descendant-snapshots` and `--preserve-snapshots` would be clearer.

joyously:
Your explanation helps a lot, (perhaps it should be in the docs), but I think the part about preserving the diff versus preserving the snapshot is way too much in the implementation details, and reminds me of the porcelain options in Git. 
I don't want to have to know what `jj` is doing behind the scenes. Perhaps the command structure is too liberal or too rigid or something if I need to give it options that specific. 
A tool that is about storing history can have some limitations on what history you can change, can't it?

ilyagr:
> Perhaps the command structure is too liberal or too rigid or something if I need to give it options that specific. A tool that is about storing history can have some limitations on what history you can change, can't it?

Maybe we'll figure out something better eventually, I'm not sure. See also <https://github.com/jj-vcs/jj/issues/6000#issuecomment-2749633772>.

-------------

My current feeling is that people don't find this rename to be enough of an improvement. If this is how people feel, I'm OK with sticking with `--restore-descendants` for now.

Another possibly relevant thought from https://github.com/jj-vcs/jj/issues/6000#issuecomment-2749633772 is that we could have a name like `--restore-children` or `--preserve-children-snapshots`, so that people don't feel like this changes the rules of jj's rebase machinery. I, at least, prefer to think of `jj squash --from X --into Y --restore-descendants` as an operation that affects X, Y, and their children. All other commits are rebased as normal (and it just so happens that their snapshots get preserved, as a consequence of the normal rules).

neongreen:
> My current feeling is that people don't find this rename to be enough of an improvement

Count me in as one of those people. I think restore-descendants is very useful and even if I can make a short alias for myself, I still want other users to not be afraid of it either.

I also like the “snapshots vs diffs” concept and I personally would want to expose it to users rather than hide it. 

I have honestly no idea how other people feel about it though. For my personal use I’m ok with any name if it has a short alias. Doesn’t even have to be a one-letter alias.

(restore-descendants is already too long for me, all names listed in the PR description are too long as well.)

PR #5396, opened 2025-01-18
formatter: add support for italic text

gechr:
Closes #5089

PR #6126, opened 2025-03-24
formatter: add support for reversing colors

mernen:
The main goal here is to support formatting comparable to git's [diff-highlight](https://github.com/git/git/tree/master/contrib/diff-highlight). Using RGB colors is admittedly even better, but sadly not all terminals support them, notably macOS's Terminal.app.

I'd like to thank @gechr, as this is heavily based on his work on #5396.

# Checklist

If applicable:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -129,12 +129,14 @@ change_id = "#ff1525"
 ```
 
 If you use a string value for a color, as in the examples above, it will be used
-for the foreground color. You can also set the background color, or make the
-text bold, italic, or underlined. For that, you need to use a table:
+for the foreground color. You can also set the background color, reverse colors
+(swap foreground and background), or make the text bold, italic, or underlined.
+For that, you need to use a table:
 
 ```toml
 [colors]
-commit_id = { fg = "green", bg = "#ff1525", bold = true, italic = true, underline = true }
+commit_id = { fg = "green", bg = "#ff1525", bold = true, underline = true }
+change_id = { reverse = true, italic = true }
 ```
 
 The key names are called "labels". The above used `commit_id` as label. You can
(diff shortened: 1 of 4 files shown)


docs/config.md
```diff
@@ -129,12 +129,12 @@ change_id = "#ff1525"
 ```
 
 If you use a string value for a color, as in the examples above, it will be used
-for the foreground color. You can also set the background color, or make the
-text bold, italic, or underlined. For that, you need to use a table:
+for the foreground color. You can also set the background color, reverse colors,
```
yuja, line 132: nit: maybe better to say "swap foreground and background colors"? (it's unclear what will be reversed.)

docs/config.md
```diff
@@ -129,12 +129,12 @@ change_id = "#ff1525"
...
+or make the text bold, italic, or underlined. For that, you need to use a table:
 
 ```toml
 [colors]
-commit_id = { fg = "green", bg = "#ff1525", bold = true, italic = true, underline = true }
+commit_id = { fg = "green", bg = "#ff1525", reverse = true, bold = true, italic = true, underline = true }
```
yuja, line 137: nit: might be better to split the example keys. It wouldn't make sense to set fg/bg and reverse them.

yuja (approved):
Looks good to me, thanks.

docs/config.md
mernen: Good idea, I've adjusted the description (while keeping the word "reverse" to associate it with the keyword that's used).

docs/config.md
mernen: I've included a new key. I also moved one other flag, just so neither gets too crowded.

PR #6127, opened 2025-03-24
git: set --intent-to-add on new files

### LINKED ISSUE #6122: Newly created files not added to git
## Description

Hi! I'm using Jujutsu with the Git backend, and whenever I create a new file, it doesn't get "added" to Git immediately, which hinders my workflow in some places (Neovim's find Git files only works when the files are added, Nix flakes only recognize files that are added, etc.). Currently, I've been using `jj new` after creating a new file, and then squashing the two commits together, which works but is quite annoying. Is there any way to fix/get around this?

1. Create a new file  
2. Run `git status`  

## Expected Behavior

All files in the working copy should be tracked by Git.  

## Actual Behavior

New files in the working copy are untracked by Git.  

## Specifications

- Platform: Linux (NixOS)  
- Version: 0.27.0

martinvonz:
Closes #6122

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * A new command `jj git root` is added, which prints the location of the Git
   directory of a repository using the Git backend.
 
+* In colocated repos, any files that jj considers added in the working copy will
+  now show up in `git diff` (as if you had run `git add --intent-to-add` on
+  them).
+
 ### Fixed bugs
 
 * `jj log -p --stat` now shows diff stats as well as the default color-words/git
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -17,17 +17,19 @@
 use std::borrow::Borrow;
 use std::borrow::Cow;
 use std::collections::HashMap;
-#[cfg(feature = "git2")]
 use std::collections::HashSet;
 use std::default::Default;
 use std::fs::File;
 use std::num::NonZeroU32;
 use std::path::PathBuf;
 use std::str;
+use std::sync::Arc;
 
 use bstr::BStr;
 use bstr::BString;
+use futures::StreamExt as _;
 use itertools::Itertools as _;
+use pollster::FutureExt as _;
 use thiserror::Error;
 
 use crate::backend::BackendError;
@@ -42,7 +44,9 @@ use crate::git_subprocess::GitSubprocessContext;
 use crate::git_subprocess::GitSubprocessError;
 #[cfg(feature = "git2")]
 use crate::index::Index;
+use crate::matchers::EverythingMatcher;
 use crate::merged_tree::MergedTree;
+use crate::merged_tree::TreeDiffEntry;
 use crate::object_id::ObjectId as _;
 use crate::op_store::RefTarget;
 use crate::op_store::RefTargetOptionExt as _;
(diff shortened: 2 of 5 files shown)


thoughtpolice (approved):

lib/src/git.rs
```diff
@@ -1424,6 +1430,86 @@ fn build_index_from_merged_tree(
...
+        }
+    }
+
+    if added_paths.is_empty() && removed_paths.is_empty() {
+        return Ok(());
+    }
+
+    for (path, executable) in added_paths {
+        // We have checked that the index doesn't have this entry
+        index.dangerously_push_entry(
```
scott2000, line 1487: Does this properly handle the case where a file `.jj-do-not-resolve-this-conflict` is added in the working copy, and also the file `.jj-do-not-resolve-this-conflict` was already added as a conflict in the index (since there was a conflict that couldn't be represented in the index properly)? It seems like this could result in a duplicate entry.

lib/src/git.rs
```diff
@@ -1424,6 +1430,86 @@ fn build_index_from_merged_tree(
     Ok(index)
 }
 
+pub fn update_intent_to_add(
```
yuja, line 1433: nit: for future reference, I think it's better to include when this function should be called.

CHANGELOG.md
yuja, line 1: nit: include the issue number in commit description

yuja (approved):

lib/src/git.rs
```diff
@@ -1424,6 +1430,86 @@ fn build_index_from_merged_tree(
...
+    old_tree: &MergedTree,
+    new_tree: &MergedTree,
+) -> BackendResult<()> {
+    let mut diff_stream = old_tree.diff_stream(new_tree, &EverythingMatcher);
+    let mut added_paths = vec![];
+    let mut removed_paths = HashSet::new();
+    while let Some(TreeDiffEntry { path, values }) = diff_stream.next().await {
+        let (before, after) = values?;
+        if before.is_absent() {
+            if let Some(TreeValue::File { id: _, executable }) = after.as_normal() {
```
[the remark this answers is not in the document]
martinvonz: Yeah, I wasn't sure either, but I've tested it now and it seems to behave as expected, so I added support for symlinks too.

lib/src/git.rs
```diff
@@ -1424,6 +1430,86 @@ fn build_index_from_merged_tree(
...
+                if index
+                    .entry_index_by_path(BStr::new(path.as_internal_file_string()))
+                    .is_err()
+                {
+                    added_paths.push((
+                        BString::new(path.as_internal_file_string().into()),
```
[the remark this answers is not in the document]
martinvonz: Done

lib/src/git.rs
martinvonz: Do you mean like in [this test case](https://github.com/jj-vcs/jj/blob/e39244828805892d71e770839d4ccc24af3ea12f/cli/tests/test_git_colocated.rs#L1247-L1258)? That passes. Does that mean the scenario you're thinking of is handled correctly?

scott2000: I was thinking of that case, but if a snapshot is taken before `jj new` is run. I'm thinking it could lead to the index having 2 entries for the same file (one conflicted and one unconflicted) until `jj new` is run, and I'm not sure that's considered valid by Git.

martinvonz: A snapshot is always taken before a command is run, so I think that's already covered. Maybe there's some way to cause a crash if you manually add a `.jj-do-not-resolve-this-conflict` file (as that test case does) but I don't really care if people get hurt when they deliberately shoot themselves in the foot. (If there's a way to run into a crash without doing that, I think we should definitely fix it, however.)

Issue #6107, opened 2025-03-21
`jj log` exits with code 3 (broken pipe) if you quit `less` with `q` before consuming the entire pipe


cole-h:
## Description

When I run `jj log` and then exit out of `less` with `q`, I get an exit code of 3. This is just a cosmetic issue, but it had been driving me insane making me think that I had broken something, somehow. It would be nice if that would not happen anymore :P

## Steps to Reproduce the Problem

1. Configure `jj` with a `ui.pager` of `less`
1. Run `jj log` in any repo with enough history that it would need to run the pager
1. Exit with `q`

## Expected Behavior

Exit code 0

## Actual Behavior

Exit code 3 (broken pipe)

## Specifications

- Platform: Linux
- Version: `jj 0.27.0-ca616fade21f559c5d0d3d0d313dd30f80697374`

yuja:
The current behavior is correct because the output stream is closed by `q` (without consuming the whole output.)

> This is just a cosmetic issue, but it had been driving me insane making me think that I had broken something, somehow. 

Perhaps, that's why we have distinct exit code `3` for broken pipe.

cole-h:
Agh, I'm sorry... I confused myself because my shell prompt hides the SIGPIPE return status from `git`, which uses  `128 + signal` (`128 + 13 = 141` in this case), so I thought it was doing some magic to exit 0 when getting SIGPIPE from its pager.

So the actual expected behavior would be "Exit code 141", instead of "Exit code 0" -- I've updated the main post to reflect this.

@yuja Do you think it would be acceptable to change the exit code from 3 to 141? I did some history spelunking, and it seems 3 was chosen "just because" (2 is already used by clap when given bad arguments: https://github.com/jj-vcs/jj/commit/15070bce5fbcbbe8c574d3f7a1ec396515ac531b).

Although, since it's been a part of `jj` for so long already, maybe we'd run into https://xkcd.com/1172/ and break workflows of people who depend on this exit code...

Let me know what you think; I'd be happy to send the one-line patch if it might be accepted, but also happy to just suck it up if there's no appetite for a change as minor as this.

yuja:
> Do you think it would be acceptable to change the exit code from 3 to 141?

iirc, it's (non-interactive) shell who translates signal exit to status code `128 + N`? I personally wouldn't want to fake the shell behavior, but I don't have a strong feeling. Changing exit code to 141 would be harmless.

PR #6135, opened 2025-03-25
RFC: improve SIGPIPE handling

cole-h:
This matches what git does, where it returns exit code 141 (128 + 13) when exiting the pager after running e.g. `git log`.

---

My desire to change this is so that I can continue to have exit statuses printed in my prompt, but ignored for these "not-really-an-error" exit statuses. I use fish, which already ignores  a 141 exit status in its default prompt like `git log` produces (as a result of this function ignoring it: https://github.com/fish-shell/fish-shell/blob/360cfdb7ae7af9a73cc3f357a78bd35b5b12e829/share/functions/__fish_print_pipestatus.fish#L22-L24), but I feel like blanket ignoring exit status 3 the same way would mask other CLI's actually-is-an-error exit statuses.

I marked this as an "RFC" because in https://github.com/jj-vcs/jj/issues/6107#issuecomment-2749797250, @yuja had some reservations (despite it not being "a strong feeling" either way, I still want to point this out) about changing this. That in addition to the fact it's been exit status 3 since 2022 means that this could break people's workflows if they depend on this exit status. I'm hoping that the benefits outweigh the detriments in your eyes, because this would make me really happy (not seeing an exit status of 3 in my prompt every time I use a `jj` command that spawns a pager).

# Checklist

If applicable:

### DIFF
--- a/cli/src/cleanup_guard.rs
+++ b/cli/src/cleanup_guard.rs
@@ -57,6 +57,7 @@ mod platform {
 
     use libc::c_int;
     use libc::SIGINT;
+    use libc::SIGPIPE;
     use libc::SIGTERM;
 
     use super::*;
@@ -89,6 +90,7 @@ mod platform {
         SIGNAL_SEND = send.into_raw_fd();
         libc::signal(SIGINT, handler as libc::sighandler_t);
         libc::signal(SIGTERM, handler as libc::sighandler_t);
+        libc::signal(SIGPIPE, handler as libc::sighandler_t);
         Ok(())
     }
 
--- a/cli/src/cleanup_guard.rs
+++ b/cli/src/cleanup_guard.rs
@@ -57,7 +57,6 @@ mod platform {
 
     use libc::c_int;
     use libc::SIGINT;
-    use libc::SIGPIPE;
     use libc::SIGTERM;
 
     use super::*;
@@ -90,7 +89,6 @@ mod platform {
         SIGNAL_SEND = send.into_raw_fd();
         libc::signal(SIGINT, handler as libc::sighandler_t);
         libc::signal(SIGTERM, handler as libc::sighandler_t);
-        libc::signal(SIGPIPE, handler as libc::sighandler_t);
         Ok(())
     }
 
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3936,8 +3936,25 @@ impl CliRunner {
         let mut ui = Ui::with_config(config.as_ref())
             .expect("default config should be valid, env vars are stringly typed");
         let result = self.run_internal(&mut ui, config);
+
+        let mut is_broken_pipe = false;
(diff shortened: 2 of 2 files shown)


emilazy:
Does Git actually directly exit with 141, or does it let itself be killed by `SIGPIPE` and then that gets translated at another layer to 141? They’re not quite the same thing. I think Rust just ignores `SIGPIPE` by default, but we could change it if we wanted to. I’m not sure if there would be other considerations there (e.g. missing snapshots?).

cole-h:
...That's a great point. It looks like `git` does actually let itself be killed by `SIGPIPE` and exits with signal 13:

```rust
use std::process::{Command, Stdio};

fn main() {
    let g = Command::new("git")
        .stdin(Stdio::inherit())
        .stdout(Stdio::inherit())
        .stderr(Stdio::inherit())
        .arg("-C")
        .arg("/some/repo")
        .arg("log")
        .output();
    let status = g.unwrap().status;
    dbg!(&status);
    dbg!(&status.code());
}

```

```
[src/main.rs:13:5] &status = ExitStatus(
    unix_wait_status(
        13,
    ),
)
[src/main.rs:14:5] &status.code() = None
```

Not sure where that leaves this PR, however... (aside from being unmergeable until we decide what approach to take, if any)

cole-h:
It looks like there is already a cleanup guard signal handler for SIGINT and SIGTERM: https://github.com/jj-vcs/jj/blob/d94aa2062e8f268ae0ebb33fab494daafa7cfb7d/cli/src/cleanup_guard.rs#L90-L91

I wonder if it's as simple as adding a call to `libc::signal(SIGPIPE, handler....)` there. I tested locally and it does indeed do the thing I expect, but I don't know if that's desirable, since it seems to trigger before e.g. `ui.finalize_pager()` can be called: https://github.com/jj-vcs/jj/blob/d94aa2062e8f268ae0ebb33fab494daafa7cfb7d/cli/src/cli_util.rs#L3940

A maybe-too-ugly way around that (if we do actually care about calling `ui.finalize_pager`) would be to set up the signal handler for SIGPIPE (or restore it to `SIG_DFL`?) after that finalize call happens, but before we return the exit code, and then `raise`  SIGPIPE to ourselves.

cole-h:
cf4e20245fce22c9c7d5f3785816736b398e1aa7 adds SIGPIPE to the cleanup guard (and is reverted in 7287d29d446e087903815024cf7d2d956ec46d0c since I'm not sure that's the right solution; but it is one possible solution).

8a1fda12c322b65a9ed2f95685038e68d78be77d restores the default SIGPIPE handler and then raises SIGPIPE after `ui.finalize_pager()` is called.

yuja:
I don't think we need to enable/emulate `SIGPIPE`. The current error handling works reasonably well, and signal handling is complicated topic in general.

I personally don't think we should fake the exit status in a way that the shell would interpret, but I think such change is acceptable if that helps.

PR #6141, opened 2025-03-26
cli `squash`: new `--restore-descendants` option

ilyagr:
See commit description.

I included a [GHC-style note](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/coding-style#2-using-notes), I think these are worth having. I'm also not sure where else to put the information of why the option is designed as it is.

Since #6124 is not very popular, I'm sticking with `--restore-descendants` for now.

# Checklist

If applicable:
- (n/a) I have updated the config schema (cli/src/config-schema.json)


cli/tests/test_squash_command.rs
ilyagr, line 1: I found a bug: if you do `jj new X; <edit X>; jj squash --restore-descendants`, the commit created for the working copy is not empty.

I will demo it shortly. It's not obvious to me how to solve it, would appreciate advice.

cli/tests/test_squash_command.rs
```diff
@@ -749,6 +749,742 @@ fn test_squash_from_to_partial() {
...
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Working copy now at: kpqxywon 3cb3a3a7 (no description set)
+    Parent commit      : qpvuntsm abd9e679 a c | (no description set)
+    Added 0 files, modified 0 files, removed 1 files
+    [EOF]
+    ");
+
+    let output = work_dir.run_jj(["show"]);
+    //  BUG: The current commit should become empty!
```
ilyagr, line 826: I found a bug: if you do jj new X; <edit X>; jj squash --restore-descendants, the commit created for the working copy is not empty.

It's not obvious to me how to solve it, would appreciate advice.

cli/tests/test_squash_command.rs
ilyagr: Added a demo below, https://github.com/jj-vcs/jj/pull/6141/files#r2015460449, closing this thread.

cli/tests/test_squash_command.rs
ilyagr: My current plan is to combine `test_reparent_descendants` in test_mut_repo with the working copy tests in the same file, since I suspect the bug is in `reparent_descendants`.

yuja: Perhaps, this is because we do `reparent` in `squash_commits()` before the destination commit is rewritten. The new wc commit would be created on top of the original `destination` commit, then reparented.

lib/src/rewrite.rs
```diff
@@ -1109,7 +1118,14 @@ pub fn squash_commits<'repo>(
...
     let mut rewritten_destination = destination.clone();
-    if sources.iter().any(|source| {
+    if restore_descendants {
+        repo.reparent_descendants_with_progress(|old_commit, rebased_commit| {
+            if old_commit.id() != destination.id() {
+                return;
+            }
+            rewritten_destination = rebased_commit;
+        })?;
+    } else if sources.iter().any(|source| {
```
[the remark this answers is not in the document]
ilyagr: I haven't followed the details yet, so these are just some first thoughts.

> Maybe we can instead let CLI reject --rebase-descendants if destination is a descendant? 

We could do this for now, but I don't see any reason to prohibit it from a UI perspective, it seems pretty clear what that should do. So, it'd seem to me like a missing feature. OTOH, I'm not sure how often this missing feature would actually be useful.

The same issue would be relevant with `jj rebase -r X --after X+++ --restore-descendants`, which may or may not be more useful.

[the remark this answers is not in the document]
yuja: > I have considered passing a new `&MergeOptions` into the merge code. Then `squash_commits()` could use that to disable the same-change rule and not need the special case. But it's probably a bit more complicated because I suppose we do want to apply the rule when squashing some a change sideways to a destination that already has some of the changes.

FWIW, I wonder if the default could be changed to disable `A-B+A=A`. It's annoying that some changes get lost by rebasing back and forth.

> > Maybe we can instead let CLI reject --rebase-descendants if destination is a descendant?
> 
> We could do this for now, but I don't see any reason to prohibit it from a UI perspective,

Yeah, I think we can unblock that later.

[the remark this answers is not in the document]
yuja: > Maybe we can instead let CLI reject --rebase-descendants if destination is a descendant?

Alternatively, I think we can turn off the (first) rebase if `--rebase-descendants` is specified. Descendant destinations will be reparented anyway, so there wouldn't be `A-B+A=A` issue?

cli/tests/test_squash_command.rs
ilyagr: I couldn't reproduce the other bug I was worrying about in the previous message. Not sure how, but everything seems to be working correctly now  AFAICT :)

cli/src/commands/squash.rs
```diff
@@ -195,6 +327,12 @@ pub(crate) fn cmd_squash(
         };
         commit_builder.set_description(new_description);
         commit_builder.write(tx.repo_mut())?;
+
+        if args.restore_descendants {
+            // See "NOTE: Not implementing `--restore-{target,source}-descendants`"
+            // in squash.rs.
+            tx.repo_mut().reparent_descendants()?;
```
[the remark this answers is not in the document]
ilyagr: Still TODO, together with some comment editing.

cli/tests/test_squash_command.rs
```diff
@@ -749,6 +749,772 @@ fn test_squash_from_to_partial() {
...
+        "--restore-descendants",
+    ]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Working copy now at: kpqxywon a665ec55 f | (no description set)
+    Parent commit      : yostqsxw aa2e7fda e | (no description set)
+    Parent commit      : mzvwutvl d38c68b2 c | (no description set)
+    [EOF]
+    ");
+    // BUG! b should now be empty!
```
ilyagr, line 1334: Found another bug :(

lib/src/rewrite.rs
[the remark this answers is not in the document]
ilyagr: > Alternatively, I think we can turn off the (first) rebase if --rebase-descendants is specified. Descendant destinations will be reparented anyway, so there wouldn't be A-B+A=A issue?

I think your suggestion is equivalent to the current version. I think you're right in that the following also works:

```diff
diff --git a/lib/src/rewrite.rs b/lib/src/rewrite.rs
index 343733f670..6e9d44d0d0 100644
--- a/lib/src/rewrite.rs
+++ b/lib/src/rewrite.rs
@@ -1118,33 +1118,26 @@
     }
 
     let mut rewritten_destination = destination.clone();
-    if sources.iter().any(|source| {
-        repo.index()
-            .is_ancestor(source.commit.id(), destination.id())
-    }) {
+    if !restore_descendants
+        && sources.iter().any(|source| {
+            repo.index()
+                .is_ancestor(source.commit.id(), destination.id())
+        })
+    {
         // If we're moving changes to a descendant, first rebase descendants onto the
         // rewritten sources. Otherwise it will likely already have the content
         // changes we're moving, so applying them will have no effect and the
         // changes will disappear.
-        if restore_descendants {
-            repo.reparent_descendants_with_progress(|old_commit, rebased_commit| {
-                if old_commit.id() != destination.id() {
-                    return;
-                }
-                rewritten_destination = rebased_commit;
-            })?;
-        } else {
             let options = RebaseOptions::default();
             repo.rebase_descendants_with_options(&options, |old_commit, rebased_commit| {
                 if old_commit.id() != destination.id() {
                     return;
                 }
                 rewritten_destination = match rebased_commit {
                     RebasedCommit::Rewritten(commit) => commit,
                     RebasedCommit::Abandoned { .. } => panic!("all commits should be kept"),
                 };
             })?;
-        }
     }
     // Apply the selected changes onto the destination
     let mut destination_tree = rewritten_destination.tree()?;
```

But it doesn't fix https://github.com/jj-vcs/jj/pull/6141#discussion_r2020047068.

> FWIW, I wonder if the default could be changed to disable A-B+A=A. It's annoying that some changes get lost by rebasing back and forth.

I think that exposing `jj` without `A-B+A=A` to users would probably annoy them, even if we provide a separate command to do `A-B+A=A` simplification[^dream].

[^dream]: I'm hoping that, one day, we'll figure out a way to present rebases and merges to the users in such a way that something like this *would* work, but right now it feels more like a dream than anything that could be done in the near term. For example, the rebase could split the common patch `B->A` out from both the rebase source and the target, and then somehow count it once in the result. In my mind, this is related to the pipe dream of having some sort of theory of rebases that makes sense both for the snapshot view of commits and the patch view of commits.

OTOH, we could rewrite `jj`'s internals as though we implemented that. So, every command would rebase without following `A-B+A=A` and then simplify the resulting conflicts of the form `A-B+A`. I *think* this would make the internals more reliable, though I'm not sure whether it will be by enough to make it worth it.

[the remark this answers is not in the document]
yuja: > I think you're right in that the following also works, as I just realized:
> 
> ```diff
> diff --git a/lib/src/rewrite.rs b/lib/src/rewrite.rs
> index 343733f670..6e9d44d0d0 100644
> --- a/lib/src/rewrite.rs
> +++ b/lib/src/rewrite.rs
> @@ -1118,33 +1118,26 @@
>      }
>  
>      let mut rewritten_destination = destination.clone();
> -    if sources.iter().any(|source| {
> -        repo.index()
> -            .is_ancestor(source.commit.id(), destination.id())
> -    }) {
> +    if !restore_descendants
> +        && sources.iter().any(|source| {
> +            repo.index()
> +                .is_ancestor(source.commit.id(), destination.id())

Correct.

> But it doesn't fix [#6141 (comment)](https://github.com/jj-vcs/jj/pull/6141#discussion_r2020047068).

Hmm, I think that's because the source commit was modified prior to reparenting. I'm not sure whether the current behavior is wrong.

If we take the current model:
 1. squash source changes into destination
 2. rebase/reparent descendants of the rewritten revisions

the test snapshot looks correct.

OTOH, if we do squash + rebase/reparent in a single pass (by using `transform_descendants()`), `b` might become empty (dependin on whether reparent is prioritized)? (I didn't think about this thoroughly, so it might be totally wrong.)

> > FWIW, I wonder if the default could be changed to disable A-B+A=A. It's annoying that some changes get lost by rebasing back and forth.
> 
> I think that exposing `jj` without `A-B+A=A` to users would probably annoy them, even if we provide a separate command to do `A-B+A=A` simplification[1](#user-content-fn-dream-efd13bf6c20b3151142fdb29c2fc4f40).

Yeah. It might worth a try, but we'll need an opt-in flag.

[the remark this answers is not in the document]
ilyagr: I haven't looked at this very carefully yet, but I wonder whether we might want to keep the `reparent_descendants` inside the squash, just to keep the count of rebased descendants correct. See the changes in

https://github.com/ilyagr/jj/commit/squash-no-restore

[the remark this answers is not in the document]
ilyagr: At a glance, without `--restore-descendants`, the counts in these tests don't look right to me (too low).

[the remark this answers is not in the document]
yuja: It might make sense to do rebase/reparent internally in `squash_commits()`, but that probably means this function can no longer return `CommitBuilder` for further edits (because the descendants of the destination commit must be rebased within this function.) I think that would be a bigger change.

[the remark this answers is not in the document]
ilyagr: I think I finally wrapped my mind around https://github.com/jj-vcs/jj/pull/6141#discussion_r2020056603 and https://github.com/jj-vcs/jj/pull/6141#discussion_r2020047068.

See the last commit for a very detailed explanation of what I think we should do, and what else we could do.

I'll massage the first two commits a bit more (probably will get rid of the GHC-style note :(. The comment in the new commit could qualify as a note, but I don't know that it's worth it), and send this for review.

cli/tests/test_squash_command.rs
ilyagr: The latest commit should fix it.

ilyagr:
OK, I think this is ready for review again. The issue with partial squashing into ancestors turned out to be quite confusing; I hope I explained things well enough to make it clear what's going on and why I made the choices I made.

The first two commits haven't changed much, except I deleted the long comment I previously had there concerning `--restore-target-descendants` and `--restore-source-descendants`. (I think it was based on the mistaken assumption that those would be trivial to implement)

lib/src/rewrite.rs
```diff
@@ -1161,6 +1208,114 @@ pub fn squash_commits<'repo>(
...
+                // child (that is, we are not in the next case). Behavior I
+                // differs, and is probably less intuitive (?).
+                //
+                // It's probably not worth the complexity to implement IV here.
+                return Err(
+                    SquashCommitsError::RestoreDescendantsMultipleParentSquashUnimplemented {
+                        from: source.commit.commit.id().clone(),
+                        to: destination.id().clone(),
+                    },
+                );
```
ilyagr: For reference, here is what happens if we don't introduce these errors. I think this probably implements everywhere what I called "option I" in the comment, and the base of the diff (the second commit) shows test cases for "option II" (TODO: double-check)

```diff
diff --git a/cli/tests/test_squash_command.rs b/cli/tests/test_squash_command.rs
index 5813323339..d39ca9137e 100644
--- a/cli/tests/test_squash_command.rs
+++ b/cli/tests/test_squash_command.rs
@@ -1519,21 +1519,19 @@
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
     Rebased 5 descendant commits (while preserving their content)
-    Working copy  (@) now at: kpqxywon f25876e1 f | (no description set)
+    Working copy  (@) now at: kpqxywon 452e8669 f | (no description set)
     Parent commit (@-)      : yostqsxw bf3abc1c e | (no description set)
-    Parent commit (@-)      : mzvwutvl 4e3db108 c | (no description set)
+    Parent commit (@-)      : mzvwutvl 6a439c6c c | (no description set)
     [EOF]
     ");
-    // WILL CHANGE NEXT COMMIT: b should now be empty, but isn't.
+    // b is now empty
     insta::assert_snapshot!(run_log(), @r"
-    @    f25876e1c27b f
+    @    452e86696ab5 f
     ├─╮  A b
     │ │  A f
-    │ ○  4e3db108b4ab c
-    │ │  A b
+    │ ○  6a439c6c3e93 c
     │ │  A c
-    │ ○  c8a5db5d033d b
-    │ │  D b
+    │ ○  4a3f4ad6571d b (empty)
     ○ │  bf3abc1c878e e
     │ │  A e
     ○ │  f4921ae685ee d
@@ -1548,6 +1546,7 @@
     let output = work_dir.run_jj(["file", "list", "-r=b"]);
     insta::assert_snapshot!(output, @r"
     a
+    b
     [EOF]
     ");
     let output = work_dir.run_jj(["file", "list", "-r=c"]);
@@ -1580,21 +1579,18 @@
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
     Rebased 5 descendant commits (while preserving their content)
-    Working copy  (@) now at: kpqxywon e68a1aeb f | (no description set)
+    Working copy  (@) now at: kpqxywon 970acd8e f | (no description set)
     Parent commit (@-)      : yostqsxw d748f816 e | (no description set)
-    Parent commit (@-)      : mzvwutvl 4cc7a349 c | (empty) (no description set)
+    Parent commit (@-)      : mzvwutvl d1ee7218 c | (no description set)
     [EOF]
     ");
-    // WILL CHANGE NEXT COMMIT: This is one of several possible behaviors here,
-    // perhaps a more reasonable one, but it is not consistent with what we consider
-    // the "reasonable" in the cases above and below. So, we will forbid
-    // squashing in this case entirely in the next commit, since any choice
-    // would have confusing aspects.
+    // CONFUSING: The --from commit did not become empty
     insta::assert_snapshot!(run_log(), @r"
-    @    e68a1aeb8dc5 f
+    @    970acd8eb65f f
     ├─╮  A c
     │ │  A f
-    │ ○  4cc7a3491d5f c (empty)
+    │ ○  d1ee72180eb7 c
+    │ │  A c
     │ ○  5e52be79326a b
     │ │  A b
     │ │  D c
@@ -1626,31 +1622,25 @@
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
     Rebased 5 descendant commits (while preserving their content)
-    Working copy  (@) now at: kpqxywon b7a0c1f5 f | (no description set)
-    Parent commit (@-)      : yostqsxw b26366b4 e | (no description set)
-    Parent commit (@-)      : mzvwutvl ec212e80 c | (no description set)
+    Working copy  (@) now at: kpqxywon b0086a72 f | (no description set)
+    Parent commit (@-)      : yostqsxw fc5e03fe e | (no description set)
+    Parent commit (@-)      : mzvwutvl 3d6e2c4b c | (no description set)
     [EOF]
     ");
-    // WILL CHANGE NEXT COMMIT: This is one of several possible behaviors here.
-    // The user might expect `b` and `d` to have empty diffs, which is not the
-    // case. We will forbid squashing in this case in the next commit..
+    // CONFUSING: The --from commits did not become empty
     insta::assert_snapshot!(run_log(), @r"
-    @    b7a0c1f57f33 f
+    @    b0086a72b6b7 f
     ├─╮  A b
     │ │  A d
     │ │  A f
-    │ ○  ec212e80dcd7 c
-    │ │  A b
+    │ ○  3d6e2c4b9096 c
     │ │  A c
-    │ ○  6b7ecfcb7195 b
-    │ │  D b
+    │ ○  1bff1cbde9c8 b
     │ │  D d
-    ○ │  b26366b44ef5 e
-    │ │  A d
+    ○ │  fc5e03fe2ccc e
     │ │  A e
-    ○ │  11ce896ba6fa d
+    ○ │  2e43856915f8 d
     ├─╯  D b
-    │    D d
     ○  a92a7b2bca0b a
     │  A a
     │  A b
diff --git a/lib/src/rewrite.rs b/lib/src/rewrite.rs
index 5a66eeee1b..635101c03b 100644
--- a/lib/src/rewrite.rs
+++ b/lib/src/rewrite.rs
@@ -1160,7 +1160,11 @@
         if source.abandon {
             repo.record_abandoned_commit(&source.commit.commit);
             abandoned_commits.push(source.commit.commit.clone());
-        } else {
+        } else if !restore_descendants
+            || !repo
+                .index()
+                .is_ancestor(destination.id(), source.commit.commit.id())
+        {
             let source_tree = source.commit.commit.tree()?;
             // Apply the reverse of the selected changes onto the source
             let new_source_tree = source_tree.merge(
```

cli/src/commands/squash.rs
```diff
@@ -223,6 +250,18 @@ pub(crate) fn cmd_squash(
...
+            // If !args.restore_descendants, the corresponding steps are done inside
+            // tx.finish()
+            let num_reparented = tx.repo_mut().reparent_descendants()?;
+            if let Some(mut formatter) = ui.status_formatter() {
+                writeln!(
+                    formatter,
+                    "Rebased {num_reparented} descendant commits (while preserving their content)",
+                )?;
+            }
+        }
```
yuja, line 264: Can't we simply omit reparenting/rebasing in this case? There should be no A-B+A problem.

(I haven't read the last commit, so feel free to ignore my comment if the reason was described in it.)

yuja (commented):
I'll review this later (maybe this weekend or the next week) as I need to recall what was the problem. It's also nice if someone who uses `--restore-descendants` can review the implemented behavior.

FWIW, my gut feeling is that, if we need detailed implementation comment about the edge cases, the behavior would also be harder for end users to understand.

lib/src/rewrite.rs
```diff
@@ -1173,16 +1182,25 @@ pub fn squash_commits<'repo>(
...
-            };
-        })?;
+        if restore_descendants {
+            repo.reparent_descendants_with_progress(|old_commit, rebased_commit| {
+                if old_commit.id() != destination.id() {
+                    return;
+                }
+                rewritten_destination = rebased_commit;
+            })?;
+        } else {
```
yuja: (EDIT: fixed location to attach the comment)

Can't we simply omit reparenting/rebasing in this case? There should be no A-B+A problem.

(I haven't read the last commit, so feel free to ignore my comment if the reason was described in it.)

ilyagr: I think you are right, there is probably no reason to do anything here. So, we probably don't need `reparent_descenants_with_progress` either. Thank you!

I'm still mildly uncomfortable that there might be some corner case, so I'll keep this open for a bit longer as a reminder to ponder it. But I think it's very likely that `if !restore_descendants` here is all-around better.

(Also, fixing this now would mess up my references to "second commit" and "third commit" around this PR 🙃 )

cli/src/commands/squash.rs
```diff
@@ -169,13 +185,24 @@ pub(crate) fn cmd_squash(
...
     let mut tx = workspace_command.start_transaction();
-    let tx_description = format!("squash commits into {}", destination.id().hex());
+    let tx_description = format!(
+        "squash commits into {}{}",
+        destination.id().hex(),
+        if args.restore_descendants {
+            " while preserving descendant contents"
+        } else {
+            ""
+        }
```
yuja: nit: I don't think we need different transaction message. Other commands just use the same description.

cli/src/commands/squash.rs
```diff
@@ -223,6 +250,18 @@ pub(crate) fn cmd_squash(
...
+
+        if args.restore_descendants {
+            // If !args.restore_descendants, the corresponding steps are done inside
+            // tx.finish()
+            let num_reparented = tx.repo_mut().reparent_descendants()?;
+            if let Some(mut formatter) = ui.status_formatter() {
+                writeln!(
+                    formatter,
+                    "Rebased {num_reparented} descendant commits (while preserving their content)",
+                )?;
```
yuja, line 440: `if num_reparented > 0 { .. }`?

cli/tests/test_squash_command.rs
```diff
@@ -803,6 +803,920 @@ fn test_squash_from_to_partial() {
...
+    // WILL CHANGE NEXT COMMIT: b should now be empty, but isn't.
+    insta::assert_snapshot!(run_log(), @r"
+    @    f25876e1c27b f
+    ├─╮  A b
+    │ │  A f
+    │ ○  4e3db108b4ab c
+    │ │  A b
+    │ │  A c
+    │ ○  c8a5db5d033d b
+    │ │  D b
```
yuja: I agree that it's more natural if `b` gets emptied.

cli/tests/test_squash_command.rs
```diff
@@ -803,6 +803,920 @@ fn test_squash_from_to_partial() {
...
+    // WILL CHANGE NEXT COMMIT: This is one of several possible behaviors here,
+    // perhaps a more reasonable one, but it is not consistent with what we consider
+    // the "reasonable" in the cases above and below. So, we will forbid
+    // squashing in this case entirely in the next commit, since any choice
+    // would have confusing aspects.
+    insta::assert_snapshot!(run_log(), @r"
+    @    e68a1aeb8dc5 f
+    ├─╮  A c
+    │ │  A f
+    │ ○  4cc7a3491d5f c (empty)
```
yuja: I don't think this isn't "reasonable". Since the source commit is the main actor of the `squash` command, and `squash` is a patch-oriented operation, I think it's good to rebase the source.

Maybe can do something like this?
```rust
tx.repo_mut().transform_descendants(roots, |rewriter| {
    if rewritten_sources.contains(rewriter.old_commit().id()) {
        rewriter.rebase()?.write()?;
    } else if args.restore_descendants {
        rewriter.reparent().write()?;
    } else {
        rewriter.rebase()?.write()?;
    }
})
```

Another idea is to not apply `--restore-descendants` to ancestors of `rewritten_sources|rewritten_descendants` (excluding abandoned sources.) I'm not sure which behavior is more correct and useful, but it would be nice if we can find generally-applicable rule and remove edge-case errors.

cli/tests/test_squash_command.rs
```diff
@@ -803,6 +803,920 @@ fn test_squash_from_to_partial() {
...
+    ------- stderr -------
+    Rebased 3 descendant commits (while preserving their content)
+    Working copy  (@) now at: kpqxywon 1d64ccbf f | (no description set)
+    Parent commit (@-)      : yostqsxw cb90d752 e | (no description set)
+    Parent commit (@-)      : mzvwutvl 4e6702ae c | (no description set)
+    [EOF]
+    ");
+    //  `d`` becomes the same as in the above example,
+    // but `c` does not lose file `b` and `e` still does not contain file `b`
+    // regardless of what happened to their parents.
```
martinvonz, line 1107: Interesting case. I had not considered how complicated this feature is. I think I would have expected the `--restore-descendants` flag to only affect descendants of the destination. Did you consider making the effect be that immediate children of the destination commit would be unaffected but everything else uses regular rebasing?

ilyagr: > Did you consider making the effect be that immediate children of the destination commit would be unaffected but everything else uses regular rebasing?

IMO, that only makes things a lot more confusing. I also think it would be non-trivial to implement.

I wrote long comment about that in the older version of this PR, https://github.com/jj-vcs/jj/commit/7b448f573c644db0f5f8073fccfa68b8f6780064#diff-f386bd32c8c9d75fded5b3b43a436edc7d89300d3548fd915999574f95b9ba8fR136, for some details. I later removed the comment, since I wrote it when I thought it'd be easy to implement, and that motivation became out of date. I could resurrect it in some form.

Some TLDRs off the top of my head:

- It's confusing what to do with `--restore-only-target-descendants` when both source and target are ancestors of the same merge commit. Different behaviors seem better depending on whether the ~~source (IIRC)~~ target is an immediate parent of the merge commit or an indirect ancestor.
- More generally, with `--restore-all-descendants` (this PR), situations where commits area squashed fully (without --keep-emptied, -i, or paths; that is, in all the cases where the source commits get abandoned)  all have a clear and unique definition of what should happen. This is not the case for `--restore-only-target-descendants`; the example I mentioned above is confusing even for a full squash. That's why I thought `--restore-all-descendants` is the way to go; I only later realized the lack of a unique clear definition for partial squashes from commits to their ancestors.

To remember details, I'd want to re-read that same comment. There was also [a long discussion about this on Discord](https://discord.com/channels/968932220549103686/1353486339600814191) ~~that I can find~~ (**Update:** found and linked), that implied that that long comment was somewhat controversial, which was another reason I got rid of it.

cli/tests/test_squash_command.rs
ilyagr: I'm not sure whether I follow your suggestions. Do they correspond to the behaviors I wrote about, or are they something different? (I also planned to spend some time analyzing this myself tomorrow or another day, but it might help and expedite things if you just tell me).

> it would be nice if we can find generally-applicable rule

My current understanding (elaborated on in the long comment inside the last commit) is that I came up with four potential generally-applicable rules. All of them agree in every case except for a partial squash into the ancestors[^desc]  (which is what I like to see; in my experience, if several mental models suggest the same behavior, it's more likely to be a behavior that's easy to understand and reason about). Three of them agree in the case I am still allowing in the last commit (partial squash of one commit into its direct parent). In the two cases I forbid, they start disagreeing with each other.

[^desc]: **Update:** Actually, case III differs from the others also in the case of partially squashing into an indirect *descendant*.

I don't think there is a consistent behavior in those cases that wouldn't be confusing to users, but perhaps you thought of something I haven't. (As I said, I don't fully understand your idea at this moment)


---------

**Update:** Also, remembering my previous thoughts, cases I or IV made the most sense if we wanted to do the same thing everywhere. But it did seem they would be confusing to people and non-trivial to implement.

cli/tests/test_squash_command.rs
ilyagr: On second thought, maybe your idea is similar to the idea @lilyball thought was the right thing to do in that long Discord discussion.

Another problematic situation (possibly already discussed in that comment) with your version of `--restore-target-descendants` is what happens if you have

```
A
|
B  Y
| /
X
```

and do `jj squash --from B --from Y --into X --restore-target-descendants`. What happens with A? Can that be consistent with what happens if you do `jj new -m Z --before B --no-edit` and then do the same squash command?

(I'm about to go to sleep, but as of now, I'm going back and forth on whether Y is relevant in this; that is, if there's any world where you'd want A to get the changes from Y here)

cli/tests/test_squash_command.rs
yuja: I think my thought process is different from the ones described in the last commit, but the result might be similar to III or IV. If I understand these, they describe how "squash" can be modeled by using more primitive jj commands/operations. OTOH, I'm wondering which commits should be considered "descendants" in the context of `--restore-descendants`.

In https://github.com/jj-vcs/jj/pull/6141#discussion_r2244368799, Martin suggested using immediate children of the destinations. Mine is somewhat similar but to use descendants of rewritten sources/targets (excluding ancestors of any of them.)

martinvonz:
@ilyagr: In order to get this unstuck, do you think we can make it error out if any of the source commits have children? That seems all that's needed in practice.

PR #6147, opened 2025-03-27
vimdiff merge tool: use jj's diff conflict style and auto-navigate to conflicts

ilyagr:
Cc @scott2000

Now, vimdiff automatically navigates to the first conflict, and looks like this:

image

The main pane now shows jj's diff view, since you can see the snapshot in the top panes

# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md` (probably not needed?)
- ~~[ ] I have added tests to cover my changes~~

### DIFF
--- a/cli/src/config/merge_tools.toml
+++ b/cli/src/config/merge_tools.toml
@@ -52,7 +52,8 @@ program = "vim"
 # `-d` enables diff mode. `-f` makes vim run in foreground even if it starts a GUI.
 # The other options make sure that only the output file can be modified.
 merge-args = ["-f", "-d", "$output", "-M", "$left", "$base", "$right",
-              "-c", "wincmd J", "-c", "set modifiable", "-c", "set write"]
+              "-c", "wincmd J", "-c", "set modifiable", "-c", "set write",
+              "-c", "/<<<<<</+2"]
 merge-tool-edits-conflict-markers = true
 # Using vimdiff as a diff editor is not recommended. For instructions on configuring
 # the DirDiff Vim plugin for a better experience, see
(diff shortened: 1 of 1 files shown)


cli/src/config/merge_tools.toml
```diff
@@ -52,9 +52,10 @@ program = "vim"
 # `-d` enables diff mode. `-f` makes vim run in foreground even if it starts a GUI.
 # The other options make sure that only the output file can be modified.
 merge-args = ["-f", "-d", "$output", "-M", "$left", "$base", "$right",
-              "-c", "wincmd J", "-c", "set modifiable", "-c", "set write"]
+              "-c", "wincmd J", "-c", "set modifiable", "-c", "set write",
+              "-c", "/<<<<<</+2"]
 merge-tool-edits-conflict-markers = true
-conflict-marker-style = "snapshot"
+conflict-marker-style = "diff"
```
scott2000, line 58: Maybe it would be better to just not set `conflict-marker-style` and let it use the user's default instead. I think we only set this because we thought it would show in diffs better, but I'm not sure that's actually the case.

ilyagr: Sounds good!

steadmon (commented):
Could you add some more detail to this commit message about what the old / new behavior is, for those of us not familiar with vimdiff?

ilyagr:
> Could you add some more detail to [3591956's](https://github.com/jj-vcs/jj/pull/6147/commits/359195662600f42341a425233c3ab4737276b3ea) commit message about what the old / new behavior is, for those of us not familiar with vimdiff?

Done, hope it helps.

steadmon (approved):

ilyagr:
Thank you for the review!

Issue #5484, opened 2025-01-27
FR: Prefix empty lines with JJ: in generated draft commit description


schweikert:
**Is your feature request related to a problem? Please describe.**
When editing commit descriptions, particularly when editing multiple commits at once, I find it visually difficult to recognize JJ-ignored blocks from actual content that I should fill in. Here is an example:

```
JJ: Enter or edit commit descriptions after the `JJ: describe` lines.
JJ: Warning:
JJ: - The text you enter will be lost on a syntax error.
JJ: - The syntax of the separator lines may change in the future.

JJ: describe 52fd2a531ca1 -------
Release v5.3

Prepare the release for 5.3.

JJ: This commit contains the following changes:
JJ:     M CHANGELOG.md
JJ:     M configure.ac

JJ: describe 6d2f3c530bb7 -------

JJ: Lines starting with "JJ: " (like this one) will be removed.
```

**Describe the solution you'd like**
It would be great, if the generated draft didn't add newlines without the JJ prefix. Maybe this could be an option, or maybe the templating language could be extended to make this possible:

```
JJ: Enter or edit commit descriptions after the `JJ: describe` lines.
JJ: Warning:
JJ: - The text you enter will be lost on a syntax error.
JJ: - The syntax of the separator lines may change in the future.
JJ:
JJ: describe 52fd2a531ca1 -------
Release v5.3

Prepare the release for 5.3.

JJ: This commit contains the following changes:
JJ:     M CHANGELOG.md
JJ:     M configure.ac
JJ:
JJ: describe 6d2f3c530bb7 -------

```

I also removed the `JJ: Lines starting with "JJ: "` line to remove some more "noise". 

This looks probably a bit like bike shedding, but to me, this would make a significant difference.

**Describe alternatives you've considered**
I tried modifying `draft_commit_description`, but this doesn't seem possible. jj adds these empty lines outside of it

**Additional context**
-

PhilipMetzger:
We recently had a [discussion on Discord][dc] about this, although in relation to `jj split` and the consensus was to either make it configurable or a template. 




[dc]: https://discord.com/channels/968932220549103686/968932220549103689/1331656421997285376

PR #6148, opened 2025-03-27
description_template: Write intro verbatim when present

mmlb:
This way we can make it so split looks more like describe and has an empty line for the first line. Most other caller sets intro as the empty string, so does not need updating.

I'm adding the empty first line to `split` since its use is very similar to `describe`, namely the user is looking at a "blank" screen and expected to write a commit message.

Squash on the other hand is presenting information that the user is going to want to read first so it makes sense to avoid the empty line as the first line.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -146,18 +146,17 @@ pub(crate) fn cmd_split(
 
     // Create the first commit, which includes the changes selected by the user.
     let first_commit = {
+        let mut intro = "".to_string();
         let mut commit_builder = tx.repo_mut().rewrite_commit(&target.commit).detach();
         commit_builder.set_tree_id(target.selected_tree.id());
         if commit_builder.description().is_empty() {
             commit_builder.set_description(tx.settings().get_string("ui.default-description")?);
+            intro += "\n";
         }
+        intro += "JJ: Enter a description for the first commit.";
+
         let temp_commit = commit_builder.write_hidden()?;
-        let template = description_template(
-            ui,
-            &tx,
-            "Enter a description for the first commit.",
-            &temp_commit,
-        )?;
+        let template = description_template(ui, &tx, &intro, &temp_commit)?;
         let description = edit_description(&text_editor, &template)?;
         commit_builder.set_description(description);
         commit_builder.write(tx.repo_mut())?
@@ -193,12 +192,8 @@ pub(crate) fn cmd_split(
             "".to_string()
         } else {
             let temp_commit = commit_builder.write_hidden()?;
-            let template = description_template(
-                ui,
-                &tx,
-                "Enter a description for the second commit.",
-                &temp_commit,
-            )?;
+            let intro = "JJ: Enter a description for the second commit.";
+            let template = description_template(ui, &tx, intro, &temp_commit)?;
             edit_description(&text_editor, &template)?
         };
         commit_builder.set_description(description);
--- a/cli/src/commands/squash.rs
+++ b/cli/src/commands/squash.rs
@@ -183,7 +183,7 @@ pub(crate) fn cmd_squash(
                 if let Some(description) = try_combine_messages(abandoned_commits, &destination) {
                     description
                 } else {
-                    let intro = "Enter a description for the combined commit.";
+                    let intro = "JJ: Enter a description for the combined commit.";
                     let combined = combine_messages_for_editing(abandoned_commits, &destination);
                     // It's weird that commit.description() contains "JJ: " lines, but works.
                     commit_builder.set_description(combined);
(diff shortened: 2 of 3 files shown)


cli/src/commands/split.rs
```diff
@@ -146,18 +146,17 @@ pub(crate) fn cmd_split(
...
     // Create the first commit, which includes the changes selected by the user.
     let first_commit = {
+        let mut intro = "".to_string();
         let mut commit_builder = tx.repo_mut().rewrite_commit(&target.commit).detach();
         commit_builder.set_tree_id(target.selected_tree.id());
         if commit_builder.description().is_empty() {
             commit_builder.set_description(tx.settings().get_string("ui.default-description")?);
+            intro += "\n";
         }
+        intro += "JJ: Enter a description for the first commit.";
```
yuja, line 156: Does it mean the user is supposed to insert new description before the "JJ: ..." line only when the original description was empty? That seems weird.

Can you add a test that shows what this patch tries to achieve?

mmlb: Yep thats correct. Similar to the `describe` template where there's jj provided info but the first line is empty, waiting for user to add the commit message. With my change here it looks a little funky because of the empty line between `JJ: Enter a ...` and the diffstat, but thats been discussed already in https://github.com/jj-vcs/jj/issues/5484 (which I'd love to see get implemented).

I'll try my hand at adding a test.

mmlb: I brought up #5484 because with this patch it looks weird that there's an empty line, some jj info, another empty line then more jj stuff. It would be fantastic if the empty line in the middle was not there, but I was calling it out here just to avoid confusion because it looks a little weird tbh.

I tried tweaking `test_split_by_paths` to get it to fail but it didn't, and I'm not really sure sure why not. I did the following:
```
diff --git a/cli/tests/test_split_command.rs b/cli/tests/test_split_command.rs
index 25c205a893..de3f60c851 100644
--- a/cli/tests/test_split_command.rs
+++ b/cli/tests/test_split_command.rs
@@ -75,6 +75,7 @@
     ");
     insta::assert_snapshot!(
         std::fs::read_to_string(test_env.env_root().join("editor0")).unwrap(), @r#"
+
     JJ: Enter a description for the first commit.

     JJ: This commit contains the following changes:
```
to simulate the empty line I'm inserting at the top. Any idea whats happening? I'm still very new to rust and I'm not sure whats going wrong.

My intention is for the commit template shown for a split with no descriptions to go from to change from (`█` denotes where the cursor is placed when the editor opens up):

```
█J: Enter a description for the first commit.

JJ: This commit contains the following changes:
JJ:     M cli/tests/test_split_command.rs

JJ: Lines starting with "JJ:" (like this one) will be removed.
```

to:
```
█
JJ: Enter a description for the first commit.

JJ: This commit contains the following changes:
JJ:     M cli/tests/test_split_command.rs

JJ: Lines starting with "JJ:" (like this one) will be removed.
```

mmlb: I don't follow the cli param suggestion. Are you suggesting I have jj open $EDITOR with an arg to have it jump to the empty line below the `JJ: Enter a description for the first commit.` line? If so, I'd argue to drop the `JJ: Enter a description for the first commit.` for this specific case completely instead. Just like there's no `JJ: Enter a description for the commit.` line for `jj describe`. Just like there's no "hint" in describe I don't think there's a need for one in this case either.

yuja: > I don't follow the cli param suggestion. Are you suggesting I have jj open $EDITOR with an arg to have it jump to the empty line below the `JJ: Enter a description for the first commit.` line?

Yes, something like `editor = ["vim", "+$line_number"]`.

> If so, I'd argue to drop the `JJ: Enter a description for the first commit.` for this specific case completely instead.

AFAICT, `jj split` inserts the intro line because it can open editor twice. It's not obvious whether the editor is opened for the first or the second commit.

Just FYI, there are some other PRs regarding the editor interaction in `jj split`:
https://github.com/jj-vcs/jj/issues/3060,
https://github.com/jj-vcs/jj/issues/5432

mmlb: Yeah #3060 would be nice to have for me I think. In the mean time how would I go about tweaking the editor invocation only for jj split in thise case?

mmlb: In my ideal scenario, jj would have templates I could use for this and avoid adding non-template lines.

yuja: If you're looking for temporary workaround, maybe you can use conditional variables (`--when.commands = ["split"]`).

https://jj-vcs.github.io/jj/latest/config/#conditional-variables

The implementation of `$line_number` would be basically the same as how we handle merge-tools, I think.

mmlb: Thanks that works quite well for my use case. I'm going to close this then but will open a new PR with the other bits, I think the refactor commits making similar actions/calls look similar is worth bringing in.

PR #6156, opened 2025-03-27
cli: completion: fix renamed paths

senekor:
# Checklist

If applicable:

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -612,14 +612,18 @@ fn modified_files_from_rev_with_jj_cmd(
     let output = cmd.output().map_err(user_error)?;
     let stdout = String::from_utf8_lossy(&output.stdout);
 
-    Ok(stdout
-        .lines()
-        .map(|line| {
-            let (mode, path) = line
-                .split_once(' ')
-                .expect("diff --summary should contain a space between mode and path");
+    let mut candidates = Vec::new();
+    // store renamed paths in a separate vector so we don't have to sort later
+    let mut renamed = Vec::new();
 
-            if let Some(dir_path) = dir_prefix_from(path, current) {
+    'line_loop: for line in stdout.lines() {
+        let (mode, path) = line
+            .split_once(' ')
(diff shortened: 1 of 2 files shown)


steadmon (commented):
In your first commit message, when you say "Correcting this reveals a bug ..." can you give some additional details about what exactly the bug is? If I'm reading correctly, it's suggesting the `{ old => new }` status line, when we actually want to suggest the old and new filenames. Pretty sure that's what's happening, but it would be better to spell it out in the commit message.

cli/src/complete.rs
```diff
@@ -612,14 +612,18 @@ fn modified_files_from_rev_with_jj_cmd(
...
+    let mut candidates = Vec::new();
+    // store renamed paths in a separate vector so we don't have to sort later
+    let mut renamed = Vec::new();
+
+    'line_loop: for line in stdout.lines() {
+        let (mode, path) = line
+            .split_once(' ')
+            .expect("diff --summary should contain a space between mode and path");
+
+        fn path_to_candidate(current: &str, mode: &str, p: impl AsRef<str>) -> CompletionCandidate {
```
steadmon, line 624: Could we move this outside the loop?

steadmon (commented):
Actually, maybe it's better to describe the bug in the second commit message, and then just refer to that in your first commit message.

cli/src/complete.rs
senekor: why?

senekor:
I've added an explanation of the bug to both commit messages.

cli/src/complete.rs
steadmon: To me it feels less readable to interrupt the loop body with a function, but I don't feel strongly about it. I'll leave it up to you whether to keep it or move it.

steadmon (approved):

cli/src/complete.rs
senekor: That makes sense. I find it slightly more readable, because the code is closer to where it's used. I'll keep it without feeling strongly about it either.

PR #6162, opened 2025-03-28
git_backend: derive the change ID from the git `change-id` header

benbrittain:
When read/writing commits from the git-backend, populate the git commit header with a backwards hash of the `change-id`. This should enable preserving change identity across various git remotes assuming a cooperative git server that doesn't strip the git header.

This feature is behind a `git.change-id` configuration flag at least to start.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Release highlights
 
+* Experimental support for transferring the change ID to/from Git remotes behind configuration
+  setting `git.write-change-id-header`. If this is enabled, the change ID will be stored in the Git
+  commit itself (in a commit header called `change-id`), which means it will be transferred by
+  regular `git push` etc. This is an evolving feature that currently defaults to "false". This
+  default will likely change in the future as we gain confidence with forge support and user
+  expectations.
+
 ### Breaking changes
 
 ### Deprecations
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -470,6 +470,11 @@
                     "description": "Whether jj spawns a git subprocess for network operations (push/fetch/clone)",
                     "default": true
                 },
+                "write-change-id-header": {
+                    "type": "boolean",
+                    "description": "Whether the change id should be stored in the Git commit object",
+                    "default": false
+                },
                 "executable-path": {
                     "type": "string",
                     "description": "Path to the git executable",
--- a/docs/git-compatibility.md
+++ b/docs/git-compatibility.md
@@ -217,3 +217,10 @@ directories in your working copy. If you then run e.g. `jj status`, the
 resulting snapshot will contain those directories, making it look like they
 replaced all the other paths in your repo. You will probably want to run
 `jj abandon` to get back to the state with the unresolved conflicts.
+
+Change IDs are stored in git commit headers as reverse hex encodings. These is
+a non-standard header and is not preserved by all `git` tooling. For example,
+the header is preserved by a `git commit --amend`, but is not preserved through
+a rebase operation. GitHub and other major forges seem to preserve them for the
+most part. This functionality is currently behind a `git.write-change-id-header`
+flag.
--- a/lib/src/config/misc.toml
+++ b/lib/src/config/misc.toml
@@ -14,6 +14,7 @@ abandon-unreachable-commits = true
 auto-local-bookmark = false
 subprocess = true
 executable-path = "git"
+write-change-id-header = false
 
 [operation]
 hostname = ""
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -69,8 +69,10 @@ use crate::backend::Timestamp;
 use crate::backend::Tree;
 use crate::backend::TreeId;
 use crate::backend::TreeValue;
+use crate::config::ConfigGetError;
 use crate::file_util::IoResultExt as _;
 use crate::file_util::PathError;
+use crate::hex_util::to_forward_hex;
 use crate::index::Index;
 use crate::lock::FileLock;
 use crate::merge::Merge;
@@ -93,6 +95,7 @@ const NO_GC_REF_NAMESPACE: &str = "refs/jj/keep/";
 const CONFLICT_SUFFIX: &str = ".jjconflict";
 
 pub const JJ_TREES_COMMIT_HEADER: &[u8] = b"jj:trees";
+pub const CHANGE_ID_COMMIT_HEADER: &[u8] = b"change-id";
 
 #[derive(Debug, Error)]
 pub enum GitBackendInitError {
@@ -100,6 +103,8 @@ pub enum GitBackendInitError {
     InitRepository(#[source] gix::init::Error),
     #[error("Failed to open git repository")]
     OpenRepository(#[source] gix::open::Error),
+    #[error(transparent)]
+    Config(ConfigGetError),
     #[error(transparent)]
     Path(PathError),
 }
@@ -159,6 +164,7 @@ pub struct GitBackend {
     empty_tree_id: TreeId,
     extra_metadata_store: TableStore,
     cached_extra_metadata: Mutex<Option<Arc<ReadonlyTable>>>,
+    change_id_setting: bool,
 }
 
 impl GitBackend {
@@ -166,7 +172,11 @@ impl GitBackend {
         "git"
     }
 
-    fn new(base_repo: gix::ThreadSafeRepository, extra_metadata_store: TableStore) -> Self {
+    fn new(
+        base_repo: gix::ThreadSafeRepository,
+        extra_metadata_store: TableStore,
+        change_id_setting: bool,
+    ) -> Self {
         let repo = Mutex::new(base_repo.to_thread_local());
         let root_commit_id = CommitId::from_bytes(&[0; HASH_LENGTH]);
         let root_change_id = ChangeId::from_bytes(&[0; CHANGE_ID_LENGTH]);
@@ -179,6 +189,7 @@ impl GitBackend {
             empty_tree_id,
             extra_metadata_store,
             cached_extra_metadata: Mutex::new(None),
+            change_id_setting,
         }
     }
 
@@ -194,7 +205,12 @@ impl GitBackend {
             gix_open_opts_from_settings(settings),
         )
         .map_err(GitBackendInitError::InitRepository)?;
-        Self::init_with_repo(store_path, git_repo_path, git_repo)
+
+        let change_id_setting = settings
+            .git_settings()
+            .map_err(GitBackendInitError::Config)?
+            .change_id;
+        Self::init_with_repo(store_path, git_repo_path, git_repo, change_id_setting)
     }
 
     /// Initializes backend by creating a new Git repo at the specified
@@ -218,7 +234,11 @@ impl GitBackend {
         )
         .map_err(GitBackendInitError::InitRepository)?;
         let git_repo_path = workspace_root.join(".git");
-        Self::init_with_repo(store_path, &git_repo_path, git_repo)
+        let change_id_setting = settings
+            .git_settings()
+            .map_err(GitBackendInitError::Config)?
+            .change_id;
+        Self::init_with_repo(store_path, &git_repo_path, git_repo, change_id_setting)
     }
 
     /// Initializes backend with an existing Git repo at the specified path.
@@ -238,13 +258,18 @@ impl GitBackend {
             gix_open_opts_from_settings(settings),
         )
         .map_err(GitBackendInitError::OpenRepository)?;
-        Self::init_with_repo(store_path, git_repo_path, git_repo)
+        let change_id_setting = settings
+            .git_settings()
+            .map_err(GitBackendInitError::Config)?
+            .change_id;
+        Self::init_with_repo(store_path, git_repo_path, git_repo, change_id_setting)
     }
 
     fn init_with_repo(
         store_path: &Path,
         git_repo_path: &Path,
         git_repo: gix::ThreadSafeRepository,
+        change_id_setting: bool,
     ) -> Result<Self, Box<GitBackendInitError>> {
         let extra_path = store_path.join("extra");
         fs::create_dir(&extra_path)
@@ -271,7 +296,11 @@ impl GitBackend {
                 .map_err(GitBackendInitError::Path)?;
         };
         let extra_metadata_store = TableStore::init(extra_path, HASH_LENGTH);
-        Ok(GitBackend::new(git_repo, extra_metadata_store))
+        Ok(GitBackend::new(
+            git_repo,
+            extra_metadata_store,
+            change_id_setting,
+        ))
     }
 
     pub fn load(
@@ -294,7 +323,12 @@ impl GitBackend {
         )
         .map_err(GitBackendLoadError::OpenRepository)?;
         let extra_metadata_store = TableStore::load(store_path.join("extra"), HASH_LENGTH);
-        Ok(GitBackend::new(repo, extra_metadata_store))
+        let change_id_setting = settings.git_settings().unwrap_or_default().change_id;
+        Ok(GitBackend::new(
+            repo,
+            extra_metadata_store,
+            change_id_setting,
+        ))
     }
 
     fn lock_git_repo(&self) -> MutexGuard<'_, gix::Repository> {
@@ -516,19 +550,30 @@ fn commit_from_git_without_root_parent(
         .try_to_commit_ref()
         .map_err(|err| to_read_object_err(err, id))?;
 
-    // We reverse the bits of the commit id to create the change id. We don't want
-    // to use the first bytes unmodified because then it would be ambiguous
-    // if a given hash prefix refers to the commit id or the change id. It
-    // would have been enough to pick the last 16 bytes instead of the
-    // leading 16 bytes to address that. We also reverse the bits to make it less
-    // likely that users depend on any relationship between the two ids.
-    let change_id = ChangeId::new(
-        id.as_bytes()[4..HASH_LENGTH]
-            .iter()
-            .rev()
-            .map(|b| b.reverse_bits())
-            .collect(),
-    );
+    // If the git header has a change-id field, we attempt to convert that to a
+    // valid JJ Change Id
+    let change_id = commit
+        .extra_headers()
+        .find("change-id")
+        .and_then(to_forward_hex)
+        .and_then(|change_id_hex| ChangeId::try_from_hex(change_id_hex.as_str()).ok())
+        .filter(|val| val.as_bytes().len() == CHANGE_ID_LENGTH)
+        // Otherwise, we reverse the bits of the commit id to create the change id.
+        // We don't want to use the first bytes unmodified because then it would be
+        // ambiguous if a given hash prefix refers to the commit id or the change id.
+        // It would have been enough to pick the last 16 bytes instead of the
+        // leading 16 bytes to address that. We also reverse the bits to make it
+        // less likely that users depend on any relationship between the two ids.
+        .unwrap_or_else(|| {
+            ChangeId::new(
+                id.as_bytes()[4..HASH_LENGTH]
+                    .iter()
+                    .rev()
+                    .map(|b| b.reverse_bits())
+                    .collect(),
+            )
+        });
+
     // shallow commits don't have parents their parents actually fetched, so we
     // discard them here
     // TODO: This causes issues when a shallow repository is deepened/unshallowed
@@ -664,7 +709,9 @@ fn serialize_extras(commit: &Commit) -> Vec<u8> {
 
 fn deserialize_extras(commit: &mut Commit, bytes: &[u8]) {
     let proto = crate::protos::git_store::Commit::decode(bytes).unwrap();
-    commit.change_id = ChangeId::new(proto.change_id);
+    if !proto.change_id.is_empty() {
+        commit.change_id = ChangeId::new(proto.change_id);
+    }
     if let MergedTreeId::Legacy(legacy_tree_id) = &commit.root_tree {
         if proto.uses_tree_conflict_format {
             if !proto.root_tree.is_empty() {
@@ -1220,6 +1267,13 @@ impl Backend for GitBackend {
                 ));
             }
         }
+        if self.change_id_setting {
+            extra_headers.push((
+                BString::new(CHANGE_ID_COMMIT_HEADER.to_vec()),
+                BString::new(contents.change_id.reverse_hex().into()),
+            ));
+        }
+
         let extras = serialize_extras(&contents);
 
         // If two writers write commits of the same id with different metadata, they
@@ -1511,11 +1565,16 @@ fn bytes_vec_from_json(value: &serde_json::Value) -> Vec<u8> {
 
 #[cfg(test)]
 mod tests {
+    use std::str::FromStr as _;
+
     use assert_matches::assert_matches;
     use hex::ToHex as _;
     use pollster::FutureExt as _;
+    use toml_edit::DocumentMut;
 
     use super::*;
+    use crate::config::ConfigLayer;
+    use crate::config::ConfigSource;
     use crate::config::StackedConfig;
     use crate::content_hash::blake2b_hash;
     use crate::tests::new_temp_dir;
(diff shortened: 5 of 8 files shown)


benbrittain:
There is a test that verifies this through `gitoxide` but it's also nice to see the git output too.

```

**Comment by benbrittain** [comment]
There is a test that verifies this, but it's kinda cool to just look at the git object too:

```
~/w/jj-change-id-test  HEAD (a49f075) > git cat-file commit a49f07530617a6059142a7275aba3fd6b5a6a5b3
tree 20c311cb4b5dcb47cfe45d2ff3474cc2d0fe5b9f
parent a2e49c7cb46255cc8a2aacc049c73379d3979f0e
author Benjamin Brittain <ben@brittain.org> 1743104555 -0400
committer Benjamin Brittain <ben@brittain.org> 1743104555 -0400
change-id mlqnqnkrxpuvuuxzlzoltostwlwyskpx

add some content
```

**Comment by martinvonz** [comment]
> change-id mlqnqnkrxpuvuuxzlzoltostwlwyskpx

That's what we call reverse hex, which is what I think seemed to be the consensus on Discord. But the commit description (and PR description) says that you use forward hex. Just need to update the descriptions?

File: lib/src/git_backend.rs [hunk 1]
```diff
@@ -518,19 +528,31 @@ fn commit_from_git_without_root_parent(
...
-            .collect(),
-    );
+    // If the git header has a change-id field, we attempt to convert that to a
+    // valid JJ Change Id
+    let change_id = commit
+        .extra_headers()
+        .find("change-id")
+        .iter()
+        .filter_map(|reverse_hex| to_forward_hex(reverse_hex.to_string().as_str()))
+        .filter_map(|change_id_hex| ChangeId::try_from_hex(change_id_hex.as_str()).ok())
```
emilazy, line 538: As I mentioned on the Discord, I think that we should treat a `change-id` header that doesn’t match our format as an invalid commit and error out, in the same way that gitoxide would if given sufficiently invalid headers.

Since the only people producing these headers at first will be us, GitButler, and Gerrit, we can decide on a uniform format together, and while nothing stops us loosening our parsing later if we decide we should be more flexible or find that people have been putting unexpected data in these in the wild, failing to reject them upfront now will basically guarantee that someone does it, in a kind of evil mutant Postel’s law consequence.

So I think being strict upfront, as the first consumer of this header, is the best way to keep our options open and reduce the likelihood of ending up with complicated parsing code that sniffs a bunch of subtly‐different formats people end up using without coordination, possibly not even intentionally.

Also: It seems like `ChangeId::try_from_hex` doesn’t enforce any requirements about the length. We should surely change this behaviour, as `change-id z` seems like a rather bad thing to accept. I would suggest that we probably enforce the exact length that we generate for now, per the same reasoning in the rest of this comment, but at least enforcing ≥ 128 bits of reverse hex seems wise.

benbrittain: > ChangeId::try_from_hex doesn’t enforce any requirements about the length. We should surely change this behaviour

I think this is because it's generic across backends. There isn't a minimum change id length at the moment I believe. We could limit it in the git_backend to the `git backend` conformant size though.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 ### Release highlights
+* Support for adding a `change-id` git header behind the configuration setting `git.change-id`. If this is enabled,
```
martinvonz, line 12: nit: git->Git

lib/src/settings.rs
```diff
@@ -88,6 +90,7 @@ impl Default for GitSettings {
             #[cfg(feature = "git2")]
             subprocess: true,
             executable_path: PathBuf::from("git"),
+            change_id: false,
```
emilazy, line 93: I think Martin was up for just doing this by default? Maybe we don’t need a setting, unless we’re concerned about privacy implications (e.g. people not wanting to disclose that they’re using Jujutsu).

martinvonz: Isn't it still useful to have a config so people can opt out if they run into problems for whatever reason? I don't mind enabling it by default, though.

benbrittain: I think this should be behind a setting (at least to start!).  We don't necessarily know how this will interact with everyone's forge at this time. 

Assuming a setting is needed, I'd still lean to defaulting `false` to start with switching to to `true` in the near-mid future.

martinvonz: (I think this comment belongs on `misc.toml`. The line here is only used in tests, I think.)

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+  `jj` will attempt to use that header to indicate what the change id of the commit should be.
```
martinvonz, line 13: I don't think most users know what Git commit headers are. Can you rewrite this in a way that's easier for regular users to understand? Maybe something about making change id transfer on `jj git push/fetch`.

lib/src/settings.rs
emilazy: > Isn't it still useful to have a config so people can opt out if they run into problems for whatever reason? I don't mind enabling it by default, though.

Right. There’s also the difference between maintaining an opt‐out forever and having one temporarily in case there are issues, so I was mostly checking the intent here.

lib/src/git_backend.rs
scott2000: @emilazy If I remember correctly, I think GitButler already used to produce `change-id` headers that don't match our format before they changed to use a custom `gitbutler-change-id` header instead? I'm not sure if it's desirable to error on those commits.

emilazy: I believe those were trailers instead? cc @schacon for confirmation.

benbrittain: I don't think we gain anything by being strict on these `change-id` headers. If the change id works for JJ? Great. If it doesn't? We don't currently lose much by creating a synthetic one. We only gain something if we start enforcing a bunch of other properties on the generic git backend.

benbrittain: https://github.com/gitbutlerapp/gitbutler/blob/e5acd96c6b474d96a3ed1106e9f74ae908177258/crates/gitbutler-commit/src/commit_headers.rs#L15

@emilazy headers

benbrittain: I'm leaving this unresolved for discussion on change-id/postel law shenanigans, but there is length enforcement now since that seems good.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 ### Release highlights
+* Support for adding a `change-id` Git header behind the configuration setting `git.change-id`. If
```
martinvonz, line 12: nit: blank line before for consistency

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+  this is enabled, `jj` will attempt to preserve the Change ID of a commit across Git remotes by
+  storing it in a Git `change-id` header. e.g. If you `jj git push` to a remote, edit it, then
+  `jj git fetch` your remote work, it should preserve the logical Change ID through those
+  modifications. This feature currently defaults to "false".
```
martinvonz: Add "We plan to make this feature unconditionally enabled in a few releases." or something like that

cli/src/config-schema.json
```diff
@@ -438,6 +438,11 @@
                     "description": "Whether jj spawns a git subprocess for network operations (push/fetch/clone)",
                     "default": true
                 },
+                "change-id": {
+                    "type": "boolean",
+                    "description": "Whether jj puts embeds a git change-id header",
```
martinvonz, line 443: "Whether to the change id should be stored in the Git commit object"?

lib/src/git_backend.rs
emilazy: > https://github.com/gitbutlerapp/gitbutler/blob/e5acd96c6b474d96a3ed1106e9f74ae908177258/crates/gitbutler-commit/src/commit_headers.rs#L15
> 
> @emilazy headers

Right, okay. Since there will already be commits like this out in the wild, I agree that enforcement isn’t great. Unfortunate, but as long the ecosystem agrees on ignoring them instead, the existing state of the PR seems fine.

lib/src/git_backend.rs
```diff
@@ -159,18 +161,24 @@ pub struct GitBackend {
...
 impl GitBackend {
     pub fn name() -> &'static str {
         "git"
     }
 
-    fn new(base_repo: gix::ThreadSafeRepository, extra_metadata_store: TableStore) -> Self {
+    fn new(
+        base_repo: gix::ThreadSafeRepository,
+        extra_metadata_store: TableStore,
+        settings: &UserSettings,
```
martinvonz, line 175: nit: I think I would prefer to pass only the boolean value here instead since this is pretty low-level code and it needs only that value.

lib/src/git_backend.rs
```diff
@@ -518,19 +528,32 @@ fn commit_from_git_without_root_parent(
...
-            .collect(),
-    );
+    // If the git header has a change-id field, we attempt to convert that to a
+    // valid JJ Change Id
+    let change_id = commit
+        .extra_headers()
+        .find("change-id")
+        .iter()
+        .filter_map(|reverse_hex| to_forward_hex(reverse_hex.to_string().as_str()))
+        .filter_map(|change_id_hex| ChangeId::try_from_hex(change_id_hex.as_str()).ok())
```
martinvonz, line 538: nit: Instead of `.iter().filter_map().filter_map()`, does `.and_then().and_then()` work?

lib/src/git_backend.rs
```diff
@@ -1218,6 +1241,13 @@ impl Backend for GitBackend {
         }
         let extras = serialize_extras(&contents);
 
+        if self.change_id_setting {
+            extra_headers.push((
+                BString::new(CHANGE_ID_COMMIT_HEADER.to_vec()),
+                BString::new(contents.change_id.reverse_hex().into()),
+            ));
+        }
```
martinvonz, line 1249: nit: move this before `let extras` so it's together with the other `extra_headers` code?

lib/src/git_backend.rs
```diff
@@ -1259,6 +1289,8 @@ impl Backend for GitBackend {
                         source: Box::new(err),
                     })?;
 
+            // NOTE: I don't think this is needed in the world where we write
+            // change-id as a git header since the git id will be different.
```
martinvonz, line 1293: I think we also need to move the predecessors field out of the extras (or remove them completely)

cli/src/config-schema.json
```diff
@@ -438,6 +438,11 @@
                     "description": "Whether jj spawns a git subprocess for network operations (push/fetch/clone)",
                     "default": true
                 },
+                "change-id": {
+                    "type": "boolean",
+                    "description": "Whether to the change id should be stored in the Git commit object",
```
steadmon, line 443: "Whether to the" -> "Whether to" ?

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
 ### Release highlights
 
+* Support for adding a `change-id` Git header behind the configuration setting `git.change-id`. If
+  this is enabled, `jj` will attempt to preserve the Change ID of a commit across Git remotes by
+  storing it in a Git `change-id` header. For example, if you `jj git push` to a remote, edit it,
+  then `jj git fetch` your remote work, it should preserve the logical Change ID through those
+  modifications. This is an evolving feature that currently defaults to "false". This default will
+  likely change in the future as we gain confidence with forge support and user expectations.
+
+
```
cenviity, line 20: nit: remove extra line break

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Release highlights
 
+* Support for adding a `change-id` Git header behind the configuration setting `git.change-id`. If
+  this is enabled, `jj` will attempt to preserve the Change ID of a commit across Git remotes by
```
cenviity, line 14: nit: Looking at the docs, we usually use lowercase ‘change’ in ‘change ID’.

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+  storing it in a Git `change-id` header. For example, if you `jj git push` to a remote, edit it,
+  then `jj git fetch` your remote work, it should preserve the logical Change ID through those
```
cenviity, line 16: nit: Same as above for lowercase ‘change’.

cli/tests/test_git_fetch.rs
```diff
@@ -68,6 +68,12 @@ fn get_log_output(work_dir: &TestWorkDir) -> CommandOutput {
     work_dir.run_jj(["log", "-T", template, "-r", "all()"])
 }
 
+#[must_use]
+fn get_change_log_output(work_dir: &TestWorkDir) -> CommandOutput {
+    let template = r#"change_id.short() ++ " \"" ++ commit_id.short() ++ " \"" ++ description.first_line() ++ "\" " ++ bookmarks"#;
```
cenviity, line 73: nit: Missing a double quote after the short commit ID? Or should the double quote before the short commit ID be removed?

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Release highlights
 
+* Support for adding a `change-id` Git header behind the configuration setting `git.change-id`. If
```
yuja, line 13: nit: If we still consider this as an experimental feature, I wouldn't add it to the release highlights. (If not, enable it by default?)

cli/tests/test_git_fetch.rs
```diff
@@ -1206,6 +1212,125 @@ fn test_git_fetch_some_of_many_bookmarks(subprocess: bool) {
     }
 }
 
+fn create_change_id_repo(work_dir: &TestWorkDir) -> String {
```
yuja, line 1215: nit: just inline this function. It doesn't seem generally useful, and there's only one caller.

cli/tests/test_git_fetch.rs
yuja, line 73: nit: single quote `' "'` also works

cli/src/config/misc.toml
```diff
@@ -17,6 +17,7 @@ private-commits = "none()"
 push-bookmark-prefix = "push-"
 push-new-bookmarks = false
 sign-on-push = false
+change-id = false
```
yuja, line 20: nit: this is redundant (because the default is set by `lib/src/config/misc.toml`)

lib/src/git_backend.rs
```diff
@@ -218,7 +227,8 @@ impl GitBackend {
         )
         .map_err(GitBackendInitError::InitRepository)?;
         let git_repo_path = workspace_root.join(".git");
-        Self::init_with_repo(store_path, &git_repo_path, git_repo)
+        let change_id_setting = settings.git_settings().unwrap_or_default().change_id;
```
yuja, line 230: should propagate config error

lib/src/settings.rs
```diff
@@ -66,6 +66,7 @@ pub struct GitSettings {
     #[cfg(feature = "git2")]
     pub subprocess: bool,
     pub executable_path: PathBuf,
+    pub change_id: bool,
```
yuja, line 69: Maybe this should be called `git.write-change-id-header` or something?

benbrittain: I change it to `change-id-header` to make it more clear. The `write-` prefix seems a bit misleading to me since it's also about loading the change-id from the commit object.

CHANGELOG.md
benbrittain: My thought here was that it was both experimental but also something that some interested individuals would enjoy turning on and this seemed like the right place to communicate that.

lib/src/git_backend.rs
benbrittain: When you say this, do you mean removing them from the commit data structure entirely?

lib/src/settings.rs
yuja: Don't we load the `change-id` header no matter if the write path is enabled or not? (That's why I suggested `write-` instead of `use-`.)

senekor:
There is a [section in the docs](https://jj-vcs.github.io/jj/latest/glossary/#change-id) that says:

> For the Git backend, Change IDs are currently maintained only locally and not exchanged via push/fetch operations.

Maybe that could be adjusted accordingly.

benbrittain:
@senekor I removed the line entirely. Thinking it might be better to keep that info centralized under `git-compatibility.md`

lib/src/git_backend.rs
```diff
@@ -1818,6 +1876,50 @@ mod tests {
         assert_eq!(std::str::from_utf8(&sig.data).unwrap(), commit_str);
     }
 
+    #[test]
+    fn round_trip_change_id_via_git_header() {
```
martinvonz, line 1880: nit: Is the round-trip actually via the header? If I'm reading the code correctly, we update `commit.change_id` in `deserialize_extras()` after we fill it in based on the commit header. So I think the round-trip is actually via the proto storage.

I think we should already make that assignment conditional on the proto having a (non-empty) value for the change id. That should simplify a future transition where we stop writing to the proto storage. Otherwise we'll run into the same thing as we did with the `jj:trees` header in #6186. I.e. otherwise some old jj binaries (but built after this PR) will expect overwrite the change id by an empty change id when newer binaries stop writing to the proto storage.

lib/src/git_backend.rs
```diff
@@ -194,7 +205,11 @@ impl GitBackend {
             gix_open_opts_from_settings(settings),
         )
         .map_err(GitBackendInitError::InitRepository)?;
-        Self::init_with_repo(store_path, git_repo_path, git_repo)
+        let change_id_setting = settings
+            .git_settings()
+            .map_err(GitBackendInitError::MissingGitConfig)?
```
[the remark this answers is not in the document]
benbrittain: unfortunately not, it doesn't implement `Into<Box<GitBackendInitError>>`.

lib/src/git_backend.rs
```diff
@@ -93,13 +95,16 @@ const NO_GC_REF_NAMESPACE: &str = "refs/jj/keep/";
...
+pub const CHANGE_ID_COMMIT_HEADER: &[u8] = b"change-id";
 
 #[derive(Debug, Error)]
 pub enum GitBackendInitError {
     #[error("Failed to initialize git repository")]
     InitRepository(#[source] gix::init::Error),
     #[error("Failed to open git repository")]
     OpenRepository(#[source] gix::open::Error),
+    #[error("Failed to open git backend due to missing config")]
+    MissingGitConfig(#[source] ConfigGetError),
```
yuja, line 107: nit: I would make it `#[error(transparent)] Config(..)` or something. Since the default config should be valid, the error would likely be caused by bad user value.

lib/src/git_backend.rs
```diff
@@ -518,19 +551,30 @@ fn commit_from_git_without_root_parent(
...
-            .rev()
-            .map(|b| b.reverse_bits())
-            .collect(),
-    );
+    // If the git header has a change-id field, we attempt to convert that to a
+    // valid JJ Change Id
+    let change_id = commit
+        .extra_headers()
+        .find("change-id")
+        .and_then(|reverse_hex| to_forward_hex(reverse_hex.to_string().as_str()))
```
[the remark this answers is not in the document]
benbrittain: It's a `BStr`, so I did a tiny refactor in the `hex_util` to avoid allocating a string

lib/src/hex_util.rs
```diff
@@ -30,9 +30,15 @@ fn to_forward_hex_digit(b: u8) -> Option<u8> {
 }
 
 pub fn to_forward_hex(reverse_hex: &str) -> Option<String> {
```
martinvonz, line 32: Can we make the argument a `impl AsRef<[u8]>` and not need two separate functions?

lib/src/git_backend.rs
benbrittain: You are right that the proto shadows the change-id being loaded in this test. Besides not loading empty proto values, I've also made so that we stop populating the change id in the proto storage too.

benbrittain: That does make sense. How about preventing loading it from the proto if the setting is enabled instead? There isn't a great way to test this otherwise because the proto load shadows it.

martinvonz: > How about preventing loading it from the proto if the setting is enabled instead?

That would work in the test but not for existing repos out there, right?

> There isn't a great way to test this otherwise because the proto load shadows it.

Maybe the test can re-init the jj repo using the existing Git repo? If the change-id header is in the Git repo, it should be preserved. Would that work?

benbrittain: That's what the git fetch tests did that we removed earlier basically. I can implement that here instead though.

benbrittain: much simpler to do it here, just a empty store and a populated git backend

lib/src/git_backend.rs
```diff
@@ -516,19 +550,30 @@ fn commit_from_git_without_root_parent(
...
-            .rev()
-            .map(|b| b.reverse_bits())
-            .collect(),
-    );
+    // If the git header has a change-id field, we attempt to convert that to a
+    // valid JJ Change Id
+    let change_id = commit
+        .extra_headers()
+        .find("change-id")
+        .and_then(|bstr| to_forward_hex(bstr.iter()))
```
martinvonz, line 558: nit: `.iter()` is probably not necessary. maybe even `.and_then(to_forward_hex)` works

lib/src/config/misc.toml
```diff
@@ -14,6 +14,7 @@ abandon-unreachable-commits = true
 auto-local-bookmark = false
 subprocess = true
 executable-path = "git"
+write-change-id-header = true
```
martinvonz, line 17: need to set to false or update tests, right?

lib/src/git_backend.rs
```diff
@@ -1815,6 +1876,55 @@ mod tests {
...
+            committer: signature.clone(),
+            encoding: None,
+            message: "git commit message".into(),
+            extra_headers: Vec::from([(
+                BString::new(CHANGE_ID_COMMIT_HEADER.to_vec()),
+                BString::new(original_change_id.reverse_hex().into()),
+            )]),
+        };
+
+        let git_commit_id = git_repo.write_object(&commit).unwrap();
```
martinvonz, line 1913: nit: write this using the `GitBackend` instead of the Gitoxide repo object so we test the write path

lib/src/git_backend.rs
benbrittain: yeah, should have updated this when we switched to `AsRef`

benbrittain:
I don't think it a real problem, but during the `read_commit` store path, we still shadow the change-id from the git commit with the value from the proto. The test really manages to verify the git commit message header load during the second backend initialization, that's when it actually uses the commit header.

martinvonz:
> I don't think it a real problem, but during the `read_commit` store path, we still shadow the change-id from the git commit with the value from the proto.

Yes, if I understand you correctly, that's expected until we've finished the migration.

lib/src/git_backend.rs
```diff
@@ -1815,6 +1874,55 @@ mod tests {
...
+
+        let (initial_commit_id, _init_commit) =
+            backend.write_commit(commit, None).block_on().unwrap();
+        let commit = backend.read_commit(&initial_commit_id).block_on().unwrap();
+        assert_eq!(
+            commit.change_id, original_change_id,
+            "The change-id header did not roundtrip"
+        );
+
+        // Because of how change ids are also persistested in extra proto files,
```
martinvonz, line 1910: nit: typo in "persistested"

martinvonz (approved):

lib/src/git_backend.rs
benbrittain: :man_facepalming:

docs/git-compatibility.md
```diff
@@ -217,3 +217,10 @@ directories in your working copy. If you then run e.g. `jj status`, the
 resulting snapshot will contain those directories, making it look like they
 replaced all the other paths in your repo. You will probably want to run
 `jj abandon` to get back to the state with the unresolved conflicts.
+
+Change IDs are stored in git commit headers as reverse hex encodings. These is
+a non-standard header and is not preserved by all `git` tooling. For example,
+the header is preserved by a `git commit --amend`, but is not preserved through
+a rebase operation. GitHub and other major forges seem to preserve them for the
+most part. This functionality is currently behind a `git.write-change-id-header`
+flag.
```
marc-hb, line 226: Just submitted a doc fixup in #8979

PR #6167, opened 2025-03-28
templates: create new git_format_patch_email_headers template

svmhdvn:
With this template, a `git format-patch` compatible email message can be generated using something like:

`jj show --git --template git_format_patch_email_headers <rev>`

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -402,3 +402,21 @@ coalesce(
   "o",
 )
 '''
+
+# The first line is a fixed magic date string used by
+# programs like file(1) to identify that this is output from
+# 'git format-patch'.
+# See https://git-scm.com/docs/git-format-patch#_description
+git_format_patch_email_headers = '''
+  concat(
+    "From " ++ commit_id ++ " Mon Sep 17 00:00:00 2001\n",
+    "From: " ++ author ++ "\n",
+    "Date: " ++ author.timestamp().format("%a, %-e %b %Y %T %z") ++ "\n",
+    "Subject: [PATCH] " ++ description.first_line() ++ "\n",
+    "\n",
+    description.remove_prefix(description.first_line()).trim_start(),
+    "---\n",
+    indent(" ", diff.stat()),
+    "\n"
+  )
+'''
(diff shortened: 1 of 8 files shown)


cli/src/config/templates.toml
```diff
@@ -402,3 +402,16 @@ coalesce(
   "o",
 )
 '''
+
+git_format_patch_email_headers = '''
+  concat(
+    "From " ++ commit_id ++ " Mon Sep 17 00:00:00 2001\n",
```
yuja, line 408: nit: can you add comment about `Mon Sep 17 00:00:00 2001`? It looks like a bug, but isn't.

cli/src/config/templates.toml
```diff
@@ -402,3 +402,16 @@ coalesce(
...
+    "From: ",
+    coalesce(author.name(), name_placeholder),
+    " <" ++ coalesce(author.email(), email_placeholder) ++ ">\n",
```
yuja, line 411: nit: maybe we can just use the default `author` output. I don't think we need placeholder for patch output.

svmhdvn:
Applied all feedback! Let me know if anything else needs to change.

cli/src/config/templates.toml
yuja, line 1: nit: In commit description:
```
templates: create new git_format_patch_email_headers template
With this template, a 'git format-patch' compatible
email message can be generated using something like:

jj show --git --template git_format_patch_email_headers <rev>
```

Missing blank line after the subject

cli/src/config/templates.toml
```diff
@@ -402,3 +402,19 @@ coalesce(
...
+# See https://git-scm.com/docs/git-format-patch#_description
+git_format_patch_email_headers = '''
+  concat(
+    "From " ++ commit_id ++ " Mon Sep 17 00:00:00 2001\n",
+    "From: " ++ author ++ "\n",
+    "Date: " ++ author.timestamp().format("%a, %-e %b %Y %T %z") ++ "\n",
+    "Subject: [PATCH] " ++ description.first_line() ++ "\n",
+    description.remove_prefix(description.first_line()),
+    "---\n",
+    indent(" ", diff.stat()) ++ "\n"
```
yuja, line 418: nit: `, "\n"` instead of `++ "\n"` to clarify the intent?

yuja (approved):
Looks good, thanks.

cli/src/config/templates.toml
svmhdvn: Done

PR #6169, opened 2025-03-29
RFC: cli `--config` global option: allow spaces around `=`

ilyagr:
I know @yuja originally forbid this intentionally, but I find it confusing that something like

```
--config 'revsets."immutable_heads()" = "none()"'
--config "ui.color = always"
```

was forbidden.

# Checklist

If applicable:


yuja:
Maybe you know, but the reason why whitespace is not stripped is that we allow bare strings. In POSIX shell, `'key= '` can also be expressed as `key=' '`. It seems a bit weird if `--config=foo=' '` didn't set `foo` to ` ` (single space).

ilyagr:
I think it's OK to require `--config="foo=' '"` for that.

ilyagr:
I was originally going to say:

> We do already require quotes if the value is `none()`, as well as for a key like `revsets."immutable_heads()"`, right?

Upon experimenting,

```
jj log --config 'revsets.immutable_heads()=none()'
```

fails but

```
jj log --config 'revsets."immutable_heads()"=none()'
```

works.

To me, this means that stripping the space before the `=` is definitely OK, but maybe there's something important I'm still missing about the space after the ` `? The tests do seem to indicate that `--config a=" "` would work, though.

ilyagr:
Another confusing thing is that, before this PR, we have:

```console
$ jj --config "ui . color =never" status  # Works fine
The working copy has no changes.
Working copy : olqzpw d8e5196 (empty) (no description set)
Parent commit: qqruox 386bb6d parse_bookmark_name plain
[22:14] macaw ~/dev/jj <qqru>→<olqz|∅>
$ jj --config "ui . color = never" status  # Boom!
Config error: Invalid type or value for ui.color
Caused by: unknown variant ` never`, expected one of `always`, `never`, `debug`, `auto`

For help, see https://jj-vcs.github.io/jj/latest/config/ or use `jj help -k config`.
```

After this PR, both would work.

yuja:
> To me, this means that stripping the space before the `=` is definitely OK,

Correct. The key part is parsed strictly, whereas the value parsing is lax.

For consistency reason, I wouldn't want to add extra whitespace handling to key parsing, though.

ilyagr:
> > To me, this means that stripping the space before the = is definitely OK,
>
> Correct. The key part is parsed strictly, whereas the value parsing is lax.

I think you're referring to the example in https://github.com/jj-vcs/jj/pull/6169#issuecomment-2763096694 right before the sentence you quoted.

IIUC, you're implying that example is a victim of issues like https://github.com/jj-vcs/jj/issues/5748, which makes perfect sense.

> For consistency reason, I wouldn't want to add extra whitespace handling to key parsing, though.

The key already seems agnostic to whitespace, I guess you're referring to value parsing? The inconsistency with keys vs values still bothers me, as well as the inconsistency with TOML syntax (which allows spaces around `=`s).

yuja:
> The key already seems agnostic to whitespace, I guess you're referring to value parsing? The inconsistency with keys vs values still bothers me, as well as the inconsistency with TOML syntax (which allows spaces around =s).

Yeah, it would be nice if the key parser were strict about whitespace (for consistency with the unquoted value handling), but I don't think it would worth the effort to roll our own key parsing function.

PR #6181, opened 2025-03-30
cli config edit: Rollback to previous config when invalid TOML is saved

### LINKED ISSUE #3905: `jj config edit` doesn't reject invalid TOML

1. Edit user config `jj config edit --user` 
1. Add some invalid cargo like:
```
[ui]
editor = vim
```
Note the lack of quotes around `vim` 
1. quit your editor

## Expected Behavior

`jj config edit --user` quits with a message like "could not save, invalid toml"

## Actual Behavior

Any `jj` command (including  `jj config edit --user`) fails with: 

```
Config error: newline in string found at line 7 column 14 in ../../../../../Library/Application Support/jj/config.toml
```

Until you fix the problem. 

## Specifications

- Version: 0.18.0

tdaron:
This should fix #3905.

If applicable:

- [x] I have updated CHANGELOG.md.

My implementation is quite simple: if an error is added to the config, jj will prompt the user if he wants to keep editing the config. If the user don't want to, then the bad config is overwritten with the previous working one (Without the need of any temporary file). Otherwise, the user can fix the error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Commit objects in templates now have `trailers() -> List<Trailer>`, the Trailer
   objects have `key() -> String` and `value() -> String`.
 
+* `jj config edit` will now roll back to previous version if a syntax error has been introduced in the new config.
+
+
 ### Fixed bugs
 
 * Fixed crash on change-delete conflict resolution.
--- a/cli/src/commands/config/edit.rs
+++ b/cli/src/commands/config/edit.rs
@@ -12,10 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use jj_lib::config::ConfigLayer;
 use tracing::instrument;
 
 use super::ConfigLevelArgs;
 use crate::cli_util::CommandHelper;
+use crate::command_error::print_error_sources;
 use crate::command_error::CommandError;
 use crate::ui::Ui;
 
@@ -40,6 +42,34 @@ pub fn cmd_config_edit(
     if !file.path().exists() {
         file.save()?;
     }
-    editor.edit_file(file.path())?;
+
+    // Editing again and again until either of these conditions is met
+    // 1. The config is OK
+    // 2. The user restores previous one
+    loop {
+        editor.edit_file(file.path())?;
+
+        // Trying to load back config. If error, prompt to continue editing
+        if let Err(e) = ConfigLayer::load_from_file(file.layer().source, file.path().to_path_buf())
+        {
+            writeln!(
+                ui.warning_default(),
+                "An error has been found inside the config:"
+            )?;
+            print_error_sources(ui, Some(&e))?;
+            let continue_editing = ui.prompt_yes_no(
+                "Do you want to keep editing the file? If not, previous config will be restored.",
+                Some(true),
+            )?;
+            if !continue_editing {
+                // Saving back previous config
+                file.save()?;
+                break;
+            }
+        } else {
+            // config is OK
+            break;
+        }
+    }
     Ok(())
 }
--- a/cli/tests/test_config_command.rs
+++ b/cli/tests/test_config_command.rs
@@ -19,6 +19,7 @@ use indoc::indoc;
 use regex::Regex;
 
 use crate::common::fake_editor_path;
+use crate::common::force_interactive;
 use crate::common::to_toml_value;
 use crate::common::TestEnvironment;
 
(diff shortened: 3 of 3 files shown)


PhilipMetzger:
please adhere to the commit style outlined in our documentation, thanks.

cli/src/commands/config/edit.rs
```diff
@@ -41,5 +42,17 @@ pub fn cmd_config_edit(
         file.save()?;
     }
     editor.edit_file(file.path())?;
+
+    // Trying to load back config. If error, saves the old config and warn the user.
+    if let Err(e) = ConfigLayer::load_from_file(file.layer().source, file.path().to_path_buf()) {
+        writeln!(
+            ui.error_with_heading("Error: "),
+            "An error has been found inside your config. Restoring previous one."
+        )?;
```
yuja, line 51: nit: use `user_error_with_message()` or something to error out. (or use `ui.warning_default()` to print warning and continue.)

cli/src/commands/config/edit.rs
```diff
@@ -41,5 +42,17 @@ pub fn cmd_config_edit(
...
+
+        // Saving back previous config
+        file.save()?;
```
yuja, line 54: We'll probably need to copy the bad file to temporary file. The user wouldn't want to start editing from scratch.

tdaron: it would mean the user would have to manually copy from the temp file to the "true" config file, idk if it's something wanted 🤔

What about saving the old "working" config in a temp file, and if the actual config file don't work then we show a warning and load the last working config from this temp file? So the user could simply do `jj config edit` to edit the non-working config and as soon as it works jujutsu stars re-using it ?

cli/src/commands/config/edit.rs
```diff
@@ -40,6 +41,36 @@ pub fn cmd_config_edit(
...
+        {
+            let command_error: CommandError = e.into();
+            print_error(
+                ui,
+                "Config error: ",
+                &command_error.error,
+                &command_error.hints,
+            )?;
+            edit_config = ui.prompt_yes_no(
+                "Do you want to keep editing the file ? If not, previous config will be restored. ",
```
tdaron, line 63: I had to use print_error here so i can print precise TOML error without exiting jj. i don't know if there is a cleaner way to do this (without having print_error going `pub`)

yuja: You can use `writeln!(ui.warning_default(), ..)` (and `print_error_sources()` if needed.)

cli/src/commands/config/edit.rs
yuja: > Seems like a good solution to simply prompt the user to keep editing, it's simpler

Yes.

cli/src/commands/config/edit.rs
tdaron: oh nice print_error_sources is much simpler. I used it as it allows user to directly see where the error is (even if it's pretty obvious that it's coming from what the user just added to the config)

cli/src/commands/config/edit.rs
```diff
@@ -40,6 +42,34 @@ pub fn cmd_config_edit(
     if !file.path().exists() {
         file.save()?;
     }
-    editor.edit_file(file.path())?;
+
+    // Editing again and again until
+    // 1. The config is OK
+    // 2. The user restore previous one
+    let mut edit_config = true;
+    while edit_config {
```
yuja, line 50: nit: I would simply use `loop {` with `return` or `break`.

cli/src/commands/config/edit.rs
```diff
@@ -40,6 +42,34 @@ pub fn cmd_config_edit(
...
+        // Trying to load back config. If error, prompt to continue editing
+        if let Err(e) = ConfigLayer::load_from_file(file.layer().source, file.path().to_path_buf())
+        {
+            writeln!(
+                ui.warning_default(),
+                "An error has been found inside the config:"
+            )?;
+            print_error_sources(ui, Some(&e))?;
+            edit_config = ui.prompt_yes_no(
+                "Do you want to keep editing the file ? If not, previous config will be restored. ",
```
yuja, line 62: nit: remove space before `?` and after `.`

cli/src/commands/config/edit.rs
yuja, line 1: Can you add a basic test? I think you can copy some bits from `test_next_prev_commands.rs` and `test_describe_command.rs` for prompt and editor interaction respectively.

yuja (commented):
The implementation looks good, thanks.

cli/src/commands/config/edit.rs
tdaron: Just added a test to check if the config error is well detected. However, I don't know if it is possible to change config file content as `jj_run_with` is still running to test if the re-edit works well.

cli/src/commands/config/edit.rs
```diff
@@ -40,6 +42,34 @@ pub fn cmd_config_edit(
...
+    // Editing again and again until
+    // 1. The config is OK
+    // 2. The user restore previous one
+    loop {
+        editor.edit_file(file.path())?;
+
+        // Trying to load back config. If error, prompt to continue editing
+        if let Err(e) = ConfigLayer::load_from_file(file.layer().source, file.path().to_path_buf())
+        {
+            writeln!(
```
yuja, line 55: nit: this could be better expressed as `let Err(e) = ... else { break; };`, but I don't feel strongly.

yuja (approved):

cli/src/commands/config/edit.rs
tdaron: I left the previous explicit if statement as I found it more readable

cli/src/commands/config/edit.rs
tdaron: Nice to know ! The test is updated :)

tdaron:
everything should be all good :)

cli/src/commands/config/edit.rs
```diff
@@ -40,6 +42,34 @@ pub fn cmd_config_edit(
...
+                ui.warning_default(),
+                "An error has been found inside the config:"
+            )?;
+            print_error_sources(ui, Some(&e))?;
+            let contine_editing = ui.prompt_yes_no(
```
cenviity, line 60: nit: continue

cli/src/commands/config/edit.rs
```diff
@@ -40,6 +42,34 @@ pub fn cmd_config_edit(
     if !file.path().exists() {
         file.save()?;
     }
-    editor.edit_file(file.path())?;
+
+    // Editing again and again until
+    // 1. The config is OK
+    // 2. The user restore previous one
```
cenviity, line 48: nit: restores

cli/src/commands/config/edit.rs
```diff
@@ -40,6 +42,34 @@ pub fn cmd_config_edit(
     if !file.path().exists() {
         file.save()?;
     }
-    editor.edit_file(file.path())?;
+
+    // Editing again and again until
```
cenviity, line 46: Perhaps change to ‘[…] until either of these conditions is met:’ for clarity?

cli/tests/test_config_command.rs
```diff
@@ -950,6 +951,48 @@ fn test_config_edit_repo() {
...
+    1: Configuration cannot be parsed as TOML document
+    2: TOML parse error at line 1, column 9
+      |
+    1 | invalid config here
+      |         ^
+    expected `.`, `=`
+
+    Do you want to keep editing the file? If not, previous config will be restored. (Yn): [EOF]
+    ");
+    std::fs::write(&edit_script, "write\n[revset-aliases]").unwrap();
```
yuja, line 985: nit: add test sample for the restore, which was previously implicitly tested by the subsequent `config edit`.

cli/src/commands/config/edit.rs
yuja: If you like explicit control flow, `match` would look nicer.

cli/tests/test_config_command.rs
tdaron: tests should be complete now i think. i removed the "valid config" test as it overlap with others tests that already test that.

cli/tests/test_config_command.rs
```diff
@@ -950,6 +951,89 @@ fn test_config_edit_repo() {
...
+    Do you want to keep editing the file? If not, previous config will be restored. (Yn): [EOF]
+    ");
+
+    let output = work_dir.run_jj_with(|cmd| {
+        force_interactive(cmd)
+            .arg("config")
+            .arg("get")
+            .arg("test")
+            .write_stdin("Y\n")
+    });
```
yuja, line 994: nit: you can use `run_jj(["config", "get", "test"])` here.

you can also use `.args([..])` to save vertical space in interactive case.

tdaron:
should be okay now :dancing_men:

CHANGELOG.md
```diff
@@ -127,6 +127,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj resolve` now accepts new built-in merge tools `:ours` and `:theirs`.
   These merge tools accept side #1 and side #2 of the conflict respectively.
 
+* `jj config edit` will now rollback to previous version if an error has been induced in the new config.
```
yuja, line 130: Just reminder. You'll need to move the changelog entry.

yuja (commented):
@tdaron I thought this would have been merged. If you don't have a permission to push the merge button, please let us know.

cc @martinvonz

martinvonz:
> @tdaron I thought this would have been merged. If you don't have a permission to push the merge button, please let us know.

Looks like I had missed that. I've invited @tdaron now. Thanks for letting me know!

tdaron:
thanks. didn't know if I had to wait for another review ahah

CHANGELOG.md
tdaron: Into fixed bugs ? Already moved it into New features

martinvonz: Into "[Unreleased]"

tdaron: ooh ok my bad. I'll fix and merge asap

PR #6184, opened 2025-03-30
diff: Add diff-do-chdir setting for external tool

### LINKED ISSUE #5801: FR: Run external diff tools from the temporary directory, at least as an option
I've set jj up to use `git diff --no-index` as my diff tool, but the tempdir name makes the header lines excessively long:

```diff
diff --git var/folders/1p/_w9l06b48xj8h8001s7n_mkh0000gn/T/jj-diff-8Mnh6U/left/.config/jj/config.toml var/folders/1p/_w9l06b48xj8h8001s7n_mkh0000gn/T/jj-diff-8Mnh6U/right/.config/jj/config.toml
index 6e3aef2..bb8acf0 100644
--- var/folders/1p/_w9l06b48xj8h8001s7n_mkh0000gn/T/jj-diff-8Mnh6U/left/.config/jj/config.toml
+++ var/folders/1p/_w9l06b48xj8h8001s7n_mkh0000gn/T/jj-diff-8Mnh6U/right/.config/jj/config.toml
```

Since the temporary directory isn't relevant to the diff, it would be nice if the diff tool was run _in_ the temp directory, with `left` and `right` (no prefix) as the directories to diff.

I was able to hack this into my config as a proof of concept:

```diff
diff --git left/.config/jj/config.toml right/.config/jj/config.toml
index 6e3aef2..eaf7e04 100644
--- left/.config/jj/config.toml
+++ right/.config/jj/config.toml
@@ -14,7 +14,7 @@ diff.tool       = "git"

 [merge-tools.git]
-diff-args = ["diff", "--no-index", "--no-prefix", "--color=always", "$left", "$right"]
+diff-args = ["-C", "$left/..", "diff", "--no-index", "--no-prefix", "--color=always", "left", "right"]
 diff-expected-exit-codes = [0, 1]
```

See how much nicer this header looks than the one above? Unfortunately, it relies on `git diff`'s `-C` option, on `$left/..` resolving to the temporary directory, and on the names of the `left` and `right` directories not changing. This feels fragile.

My preferred solution would be to (optionally) change the working directory the external diff tool [runs](https://github.com/jj-vcs/jj/blob/653cc6c1c9ce2be72417d4c51ef77a6fb6c5c168/cli/src/merge_tools/external.rs#L451) in, and use `left` and `right` without the tempdir prefix.

Aside from the hack I'm using (see above), I considered a different solution that adds a `$parent` variable alongside `$left` and `$right` to the `diff-args` configuration. This solution requires that the tool supports something like `git diff`'s `-C` option, and that the prefixless `left` and `right` directory names either never change or are also added as variables.

**Additional context**
The reason I want to use `git diff` is because it can colour moved lines differently, and its patience algorithm sometimes makes diffs more readable.

eyvind:
The temporary folder name can add significant noise to the produced diff when using an external diff tool if the tool displays paths prominently. Add a setting, `diff-do-chdir`, that chdirs to the temporary directory before running the tool.

Closes #5801

Disclaimer: I'm not a real Rust programmer, just cosplaying as one to get my pet change into jj. I'm also not sure about the `diff-do-chdir` name. Comments welcome!

# Checklist

If applicable:

### DIFF
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -509,6 +509,11 @@
                         "description": "Array of exit codes that do not indicate tool failure, i.e. [0, 1] for unix diff.",
                         "default": [0]
                     },
+                    "diff-do-chdir": {
+                      "type": "boolean",
+                      "description": "Invoke the tool in the temporary diff directory. This setting will be removed soon",
+                      "default": true
+                    },
                     "diff-invocation-mode": {
                       "description": "Invoke the tool with directories or individual files",
                       "enum": [
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -1360,18 +1360,18 @@ pub fn show_file_by_file_diff(
             }
             let left_path = create_file(left_path, &left_wc_dir, left_value)?;
             let right_path = create_file(right_path, &right_wc_dir, right_value)?;
+            let patterns = &maplit::hashmap! {
+                "left" => left_path
+                    .strip_prefix(temp_dir.path()).expect("path should be relative to temp_dir")
+                    .to_str().expect("temp_dir should be valid utf-8"),
+                "right" => right_path
+                    .strip_prefix(temp_dir.path()).expect("path should be relative to temp_dir")
+                    .to_str().expect("temp_dir should be valid utf-8"),
+            };
 
             let mut writer = formatter.raw()?;
-            invoke_external_diff(
-                ui,
-                writer.as_mut(),
-                tool,
-                &maplit::hashmap! {
-                    "left" => left_path.to_str().expect("temp_dir should be valid utf-8"),
-                    "right" => right_path.to_str().expect("temp_dir should be valid utf-8"),
-                },
-            )
-            .map_err(DiffRenderError::DiffGenerate)?;
+            invoke_external_diff(ui, writer.as_mut(), tool, temp_dir.path(), patterns)
+                .map_err(DiffRenderError::DiffGenerate)?;
         }
         Ok::<(), DiffRenderError>(())
     }
--- a/cli/src/merge_tools/external.rs
+++ b/cli/src/merge_tools/external.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::io;
 use std::io::Write;
+use std::path::Path;
 use std::process::Command;
 use std::process::ExitStatus;
 use std::process::Stdio;
@@ -56,6 +57,8 @@ pub struct ExternalMergeTool {
     /// Whether to execute the tool with a pair of directories or individual
     /// files.
     pub diff_invocation_mode: DiffToolMode,
+    /// Whether to execute the tool in the temporary diff directory
+    pub diff_do_chdir: bool,
     /// Arguments to pass to the program when editing diffs.
     /// `$left` and `$right` are replaced with the corresponding directories.
     pub edit_args: Vec<String>,
@@ -110,6 +113,7 @@ impl Default for ExternalMergeTool {
             merge_conflict_exit_codes: vec![],
             merge_tool_edits_conflict_markers: false,
             conflict_marker_style: None,
+            diff_do_chdir: true,
             diff_invocation_mode: DiffToolMode::Dir,
         }
     }
@@ -395,7 +399,7 @@ pub fn edit_diff_external(
         &options,
     )?;
 
-    let patterns = diffedit_wc.working_copies.to_command_variables();
+    let patterns = diffedit_wc.working_copies.to_command_variables(false);
     let mut cmd = Command::new(&editor.program);
     cmd.args(interpolate_variables(&editor.edit_args, &patterns));
     tracing::info!(?cmd, "Invoking the external diff editor:");
@@ -436,7 +440,13 @@ pub fn generate_diff(
         .map_err(ExternalToolError::SetUpDir)?;
     set_readonly_recursively(diff_wc.right_working_copy_path())
         .map_err(ExternalToolError::SetUpDir)?;
-    invoke_external_diff(ui, writer, tool, &diff_wc.to_command_variables())
+    invoke_external_diff(
+        ui,
+        writer,
+        tool,
+        diff_wc.temp_dir(),
+        &diff_wc.to_command_variables(true),
+    )
 }
 
 /// Invokes the specified `tool` directing its output into `writer`.
@@ -444,11 +454,32 @@ pub fn invoke_external_diff(
     ui: &Ui,
     writer: &mut dyn Write,
     tool: &ExternalMergeTool,
+    diff_dir: &Path,
     patterns: &HashMap<&str, &str>,
 ) -> Result<(), DiffGenerateError> {
     // TODO: Somehow propagate --color to the external command?
     let mut cmd = Command::new(&tool.program);
-    cmd.args(interpolate_variables(&tool.diff_args, patterns));
+    let mut patterns = patterns.clone();
+    let absolute_left_path = Path::new(diff_dir).join(patterns["left"]);
+    let absolute_right_path = Path::new(diff_dir).join(patterns["right"]);
+    if !tool.diff_do_chdir {
+        patterns.insert(
+            "left",
+            absolute_left_path
+                .to_str()
+                .expect("temp_dir should be valid utf-8"),
+        );
+        patterns.insert(
+            "right",
+            absolute_right_path
+                .to_str()
+                .expect("temp_dir should be valid utf-8"),
+        );
+    } else {
+        cmd.current_dir(diff_dir);
+    }
+    cmd.args(interpolate_variables(&tool.diff_args, &patterns));
+
     tracing::info!(?cmd, "Invoking the external diff generator:");
     let mut child = cmd
         .stdin(Stdio::null())
--- a/cli/tests/test_diff_command.rs
+++ b/cli/tests/test_diff_command.rs
@@ -3051,6 +3051,49 @@ fn test_diff_external_tool() {
 }
 
 #[test]
+fn test_diff_do_chdir() {
+    let mut test_env = TestEnvironment::default();
+    let edit_script = test_env.set_up_fake_diff_editor();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let work_dir = test_env.work_dir("repo");
+    work_dir.write_file("file1", "file1\n");
+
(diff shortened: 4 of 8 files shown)


martinvonz:
Any reason to not do this unconditionally?

I can think of a reason, but I think it's a bad one. Some diff tools may want to read a config file from the working copy. We actually run `jj fix` formatters from the root of the repo's working copy for this reason. I kind of regret that we do.

eyvind:
I can't think of any other reasons, but I'm wary about making a change that might break peoples' workflows and I'm not sure I can think of all the ways it could. _I_ wouldn't mind it being unconditional :)

martinvonz:
I don't mind the config as an escape hatch that we can tell people to use if they run into unforeseen issues. I would prefer that we enable it by default and plan to remote it in a while if we don't hear complaints. It's better to not document the new config because by not documenting, we're more likely to hear if anyone does run into problems.

arxanas:
> I don't mind the config as an escape hatch that we can tell people to use if they run into unforeseen issues. I would prefer that we enable it by default and plan to remote it in a while if we don't hear complaints. It's better to not document the new config because by not documenting, we're more likely to hear if anyone does run into problems.

Rather than being configurable behavior, I think it would be better to expose the original repo path via some environment variable? Then `jj fix` tools can still figure out the path to the config file as necessary (or even `cd` into the original directory). Since it won't be conditional, I think it'll be easier to implement, too?

martinvonz:
I'm suggesting that we remove the configuration in a few releases, so that will be even simpler. If we later get a feature request to provide the path of the source repo (and perhaps the input trees?), then we can add that then, IMO.

eyvind:
Sounds good to me, I'll make a new PR. Just need to make the tests work on Windows.

ilyagr:
> It's better to not document the new config because by not documenting, we're more likely to hear if anyone does run into problems.

I think it'd be best to mention it in the changelog, at least. There, we can tell people that we want to know if they needed to toggle it.

(I'd also be fine with documenting it as a normal config, but I agree that defaulting to the new behavior would probably be best)

cli/tests/test_diff_command.rs
```diff
@@ -2274,6 +2275,59 @@ fn test_diff_external_tool() {
...
+            .lines()
+            .next()
+            .unwrap(),
+        "$TEST_ENV/repo" // I don't know why pwd returns this, but it does
+    );
+    assert!(
+        work_dir
+            .run_jj([
+                "diff",
+                "--tool=pwd",
```
ilyagr, line 2302: To have the command work on Windows, you might need to use something like https://github.com/jj-vcs/jj/blob/main/cli/testing/fake-diff-editor.rs and add a command there.

I suppose you could also use Windows' analogue of `pwd`, but that seems like more of a headache.

cli/tests/test_diff_command.rs
```diff
@@ -2274,6 +2275,59 @@ fn test_diff_external_tool() {
...
+                "--config=merge-tools.pwd.diff-args=[]"
+            ])
+            .to_string()
+            .lines()
+            .next()
+            .unwrap(),
+        "$TEST_ENV/repo" // I don't know why pwd returns this, but it does
```
ilyagr, line 2296: (No action needed): `run_jj` substitutes some strings that change from one test run for another with a constant string. This applies to the directory for the test environment.

cli/tests/test_diff_command.rs
```diff
@@ -2274,6 +2275,59 @@ fn test_diff_external_tool() {
...
+    [EOF]
+    ");
+    let output = work_dir.run_jj([
+        "diff",
+        "--tool=echo",
+        "--config=merge-tools.echo.diff-args=[\"==$left--\", \"==$right--\"]",
+    ]);
+    let regex = Regex::new(r"==.+?/(left|right)--").unwrap();
+    insta::assert_snapshot!(output.normalize_stdout_with(
+        |text| regex.replace_all(&text, "TMPDIR/$1").into_owned()), @r"
```
ilyagr, line 2325: (Optional) You might want to add similar logic for `$TMP_DIR` as `run_jj` already has for `$TEST_ENV`.

cli/src/merge_tools/external.rs
```diff
@@ -427,19 +431,38 @@ pub fn generate_diff(
...
+                "left" => "left",
+                "right" => "right",
+            },
+        ));
+        cmd.current_dir(diff_dir);
+    } else {
+        cmd.args(interpolate_variables(&tool.diff_args, patterns));
+    }
+
     tracing::info!(?cmd, "Invoking the external diff generator:");
```
ilyagr, line 466: I'm curious, if you run `jj diff --debug --tool something`, will this show the directory the program is run from? If not, we should show it manually, so that the user can reproduce the invocation and find the files if they need to debug the diff tool.

cli/src/merge_tools/external.rs
```diff
@@ -427,19 +431,38 @@ pub fn generate_diff(
...
+        ui,
+        writer,
+        tool,
+        diff_wc.temp_dir(),
+        &diff_wc.to_command_variables(),
+    )
 }
 
 /// Invokes the specified `tool` directing its output into `writer`.
 pub fn invoke_external_diff(
```
ilyagr, line 444: (Maybe not for this PR): We should do something similar with diff editors and merge editors.

cli/tests/test_diff_command.rs
ilyagr: Actually, you might not need to add any commands, as long as `fake-diff-editor` can print the file paths and check that it can read the files.

This might also avoid the need for messing with `TMP_DIR`, https://github.com/jj-vcs/jj/pull/6184#discussion_r2023795109

cli/tests/test_diff_command.rs
eyvind: Aha! Thanks, that explains that. But it doesn't appear to do that for Windows `pwd`, which is why that test fails.

eyvind:
I haven't given up on this PR! The `test_diff_external_file_by_file_tool` test fails when diff-do-chdir defaults to true, and I'm trying to figure out why.

cli/src/merge_tools/external.rs
eyvind: It does:

```
2025-04-03T20:10:33.683720Z  INFO run_command:cmd_diff: jj_cli::merge_tools::external: Invoking the external diff generator: cmd=cd "/var/folders/1p/_w9l06b48xj8h8001s7n_mkh0000gn/T/jj-diff-caPuNI" && "something" "left" "right"
```

but the temporary directory is deleted before jj exits.

cli/src/merge_tools/external.rs
```diff
@@ -436,19 +441,38 @@ pub fn generate_diff(
...
     patterns: &HashMap<&str, &str>,
 ) -> Result<(), DiffGenerateError> {
     // TODO: Somehow propagate --color to the external command?
     let mut cmd = Command::new(&tool.program);
-    cmd.args(interpolate_variables(&tool.diff_args, patterns));
+    if tool.diff_do_chdir {
+        cmd.args(interpolate_variables(
+            &tool.diff_args,
+            &maplit::hashmap! {
+                "left" => file_util::relative_path(diff_dir, Path::new(patterns["left"])).to_str().unwrap().to_owned(),
```
eyvind, line 467: This fixes the file-by-file invocation, which works by passing individual files as "left" and "right" mappings.

cli/src/merge_tools/external.rs
```diff
@@ -436,19 +441,38 @@ pub fn generate_diff(
...
+                "right" => file_util::relative_path(diff_dir, Path::new(patterns["right"])).to_str().unwrap().to_owned(),
```
martinvonz, line 468: Can we make "left" and "right" in `patterns` already be relative instead?

cli/src/config-schema.json
```diff
@@ -505,6 +505,11 @@
                         "description": "Array of exit codes that do not indicate tool failure, i.e. [0, 1] for unix diff.",
                         "default": [0]
                     },
+                    "diff-do-chdir": {
+                      "type": "boolean",
+                      "description": "Invoke the tool in the temporary diff directory",
```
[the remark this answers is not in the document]
eyvind: It is the default as of this PR; are you suggesting mentioning that the setting will be removed soon?

cli/src/merge_tools/external.rs
eyvind: I think that's doable.

eyvind: Or... hm. Copying the files to the temp dir uses the absolute path, so I think we'd have to do the moral equivalent of this code in each caller anyway. I'll experiment a bit.

eyvind: The answer turns out to be "sort of". I've implemented using relative paths in `patterns` from the caller as a separate commit on top of the version that munges `patterns` from absolute to relative paths.

The change is labeled as "experimental" because I'm not super happy with it yet, but I wanted to get feedback on if this is the direction I should be going in.

cli/src/diff_util.rs
```diff
@@ -1366,9 +1366,14 @@ pub fn show_file_by_file_diff(
                 ui,
                 writer.as_mut(),
                 tool,
+                temp_dir.path(),
                 &maplit::hashmap! {
-                    "left" => left_path.to_str().expect("temp_dir should be valid utf-8"),
-                    "right" => right_path.to_str().expect("temp_dir should be valid utf-8"),
+                    "left" => left_path
+                        .strip_prefix(temp_dir.path()).expect("path should be relative to temp_dir")
```
martinvonz, line 1372: nit: I would put this right after lines 1361-1362, or maybe even in the same statement. I think being closer to where the path is produced makes it clearer why it's a safe assumption. But I'm also fine with leaving it here if you prefer

cli/src/diff_util.rs
```diff
@@ -1366,9 +1366,14 @@ pub fn show_file_by_file_diff(
...
+                        .to_str().expect("temp_dir should be valid utf-8"),
```
martinvonz, line 1373: Perhaps this part of the code is a good place to use `Utf8PathBuf` from the `camino` crate, but not in this PR.

martinvonz (commented):
I think this looks good, so please squash the two commits, IMO. We can clean it up when we remove support for absolute paths.

cli/src/diff_util.rs
eyvind: Thanks! That makes sense, and made the function look a little nicer, too. I'll squash (and rebase on main, oops).

martinvonz (approved):

PR #6189, opened 2025-03-31
cli: require --deleted to push deleted bookmarks

senekor:
# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -89,7 +89,7 @@ use crate::ui::Ui;
 
 #[derive(clap::Args, Clone, Debug)]
 #[command(group(ArgGroup::new("specific").args(&["bookmark", "change", "revisions", "named"]).multiple(true)))]
-#[command(group(ArgGroup::new("what").args(&["all", "deleted", "tracked"]).conflicts_with("specific")))]
+#[command(group(ArgGroup::new("what").args(&["all", "tracked"]).conflicts_with("specific")))]
 pub struct GitPushArgs {
     /// The remote to push to (only named remotes are supported)
     ///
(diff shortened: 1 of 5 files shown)


senekor:
Tests probably need adjustment, haven't done that yet.

We may also discuss if the flags `--all` and `--tracked` should be renamed to something like `--all-but-deleted` and `--tracked-but-deleted`. My opinion is to leave the names as they are. "all" can reasonably mean "all the ones that exist". Some goes for "tracked".

yuja:
> My opinion is to leave the names as they are. "all" can reasonably mean "all the ones that exist". Some goes for "tracked".

That seems good, but I think deleted bookmarks should be warned in that case. Technically, `--all`/`--tracked` should include deleted bookmarks, but for extra safety, the `--deleted` flag is required.

cli/tests/test_git_push.rs
```diff
@@ -2247,6 +2281,8 @@ fn test_git_push_tracked_vs_all(subprocess: bool) {
     ------- stderr -------
     Warning: Non-tracking remote bookmark bookmark1@origin exists
     Hint: Run `jj bookmark track bookmark1@origin` to import the remote bookmark.
+    Warning: Skipping deleted bookmark: bookmark2
+    Hint: Push deleted bookmarks with --deleted.
```
senekor, line 2285: This warning probably should be there. It's a niche situation. The bookmark was deleted, but it wasn't tracked, so it also get pushed even with `--deleted`. So the suggestion to use `--deleted` is wrong here, it won't actually do anything.

yuja: I think the warning condition can be tested in `classify_bookmark_update()`. It's quite similar to `allow_new`.

cli/src/commands/git/push.rs
```diff
@@ -254,12 +264,18 @@ pub fn cmd_git_push(
             remote = remote.as_symbol()
         );
     } else if args.deleted {
+        let allow_delete = true;
```
yuja, line 267: nit: move closer to the call site for consistency

PR #6196, opened 2025-04-01
cli external merge tools: add tracing when reparsing conflicts

ilyagr:
I was trying to debug `mergiraf` behavior, and having such diagnostic would have been helpful.

### DIFF
--- a/cli/src/merge_tools/external.rs
+++ b/cli/src/merge_tools/external.rs
@@ -258,6 +258,7 @@ fn run_mergetool_external_single_file(
             tool_binary: editor.program.clone(),
             source: e,
         })?;
+    tracing::info!(%exit_status);
 
     // Check whether the exit status implies that there should be conflict markers
     let exit_status_implies_conflict = exit_status
@@ -277,6 +278,12 @@ fn run_mergetool_external_single_file(
     }
 
     let new_file_ids = if editor.merge_tool_edits_conflict_markers || exit_status_implies_conflict {
+        tracing::info!(
+            ?exit_status_implies_conflict,
+            "jj is reparsing output for conflicts, `merge-tool-edits-conflict-markers = {}` in \
+             TOML config;",
+            editor.merge_tool_edits_conflict_markers
+        );
         conflicts::update_from_content(
             file_merge,
             store,


cli/src/merge_tools/external.rs
```diff
@@ -276,6 +276,15 @@ fn run_mergetool_external_single_file(
...
 
+    if exit_status_implies_conflict {
+        tracing::info!(
+            "Parsing conflicts in output file due to tool exiting with code {:?}",
+            // The code will always be Some(...), but we never want to panic here if logic changes
+            exit_status.code().unwrap_or(-9999)
+        );
+    } else if editor.merge_tool_edits_conflict_markers {
+        tracing::info!("Parsing conflicts in output file due to tool config");
+    }
```
yuja, line 287: nit: better to just dump raw values? `tracing::debug!(?exit_status, edits_conflict_marker = ?editor.merge_tool_edits_conflict_marker, ...)`

I think tracing code should be kept simple.

ilyagr: Aside: I'm also wondering about adding a debugging mode where `jj` pauses after invoking the merge tool and gives the option not to delete the temporary dir. I'm running into naming problems, though, since `--debug` is already taking. I was thinking `--config ui.debug-merge-tools=true`, but again, the `debug` here can be confused with `--debug`.

If that happens at some point, we could maybe stop using tracing for user-facing debug messages.

yuja: Hmm, then Display works better? `%exit_status`. I also think it's better to print `%exit_status` unconditionally. When debugging merge-tools behavior, you'll probably want to see it no matter if conflicts are implied or not.

> debugging mode where jj pauses after invoking the merge tool and gives the option not to delete the temporary dir

Adding env var might be okay.

ilyagr: `%exist_status` works, thanks! How about this?

cli/src/merge_tools/external.rs
```diff
@@ -277,6 +278,13 @@ fn run_mergetool_external_single_file(
     }
 
     let new_file_ids = if editor.merge_tool_edits_conflict_markers || exit_status_implies_conflict {
+        tracing::info!(
+            "jj is reparsing output for conflicts{}.",
+            editor
+                .merge_tool_edits_conflict_markers
+                .then_some(" due to `merge-tool-edits-conflict-markers` config")
+                .unwrap_or_default()
+        );
```
yuja, line 287: nit: just print these boolean variables? That should work so long as the variable names are descriptive.

Maybe I don't fully understand the reason why you want to see "due to merge-tool-edits-conflict-markers config" or not. If we want to check typos in config, I would insert `tracing::debug!(?editor)` or something.

cli/src/merge_tools/external.rs
```diff
@@ -269,6 +269,7 @@ fn run_mergetool_external_single_file(
             exit_status,
         }));
     }
+    tracing::info!(%exit_status, "Tool exited successfully{}:", exit_status_implies_conflict.then_some(", informed jj to expect conflicts").unwrap_or_default());
```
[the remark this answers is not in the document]
ilyagr: If the error is returned in the `if` above, the error status will be printed as part of the error, so printing it twice seems redundant.

cli/src/merge_tools/external.rs
ilyagr: I want to print something useful to a user who doesn't understand the internals. For example, they might not know that the underscores in `merge_tool_edits_conflict_markers` correspond to dashes in the option name, or that this even refers to an option. This gives them something to search the docs for.

yuja: Okay, I misunderstood that this would be for people who are struggling with the tool configuration (and wondered why it is printed only when conflict is implied.)

Maybe something like this will work? (I feel it's a bit simpler and still readable.)
```
tracing::info!(
    "reparsing output for conflicts, merge-tool-edits-conflict-markers = {:?}, \
    exit_status_implies_conflict = {:?}",
    ..
)?;
```

cli/src/merge_tools/external.rs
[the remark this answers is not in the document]
yuja: True, but it's debug trace. A bit of redundancy shouldn't matter. (It might help if exit status is printed with function name, for example.)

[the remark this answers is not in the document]
ilyagr: I'm not sure I follow (isn't the function name printed by the previous tracing::info?), but I can move it earlier if you prefer.

cli/src/merge_tools/external.rs
ilyagr: OK. I still want to clarify that `exist_status_implied_conflict` is an internal flag while `merge-tool-edits-conflict-markers` is a config one can set.

ilyagr:
For reference, the debug output now looks like this.

image

cli/src/merge_tools/external.rs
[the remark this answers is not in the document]
yuja: Yes, but technically, caller of this function could print whatever output in error-handling code path.

yuja (approved):
Thanks!

Issue #4549, opened 2024-09-29
FR/RFC: allow edit/diff/merge-tools to have warnings or be explicitly disallowed


ilyagr:
It is currently difficult to deprecate merge tools. I also wonder how many people run `jj split --tool vscode` and get a non-functional interface.

To address this, we could reorganize `merge-tools.toml` ([current state](https://github.com/martinvonz/jj/blob/64e1ae277db390989a6deb8f5a1a4bc62fa0e570/cli/src/config/merge_tools.toml)) and merge tool config a bit; I was thinking about how to do it for a while. One possibility is to change `edit-args` to `edit.args`, and then add a few fields to allow the following:

```toml
[merge-tools.meld]
merge.args = ["$left", "$base", "$right", "-o", "$output", "--auto-merge"]

[merge-tools.vscode]
program = "code"
diff = { error = true }
# error = "message" also allowed, error=true gives a generic message:
# "The tool `vscode` cannot be used for diff editing", etc.
edit = { error = true }
merge = {
  args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
  edits-conflict-markers = true
  warning = "This tool is deprecated and will soon be removed. Use `code` instead"
}

[merge-tools.code]
# ...
```

We could also rename the whole `[merge-tools]` table while we are at it, but I wouldn't do it unless we can come up with a great name. I don't like `[tools]` or `[external-tools]` since there are also a different category of tools for `jj fix`, unless we want those to be `tools.NAME.fix` (which might not make much sense, their config is very different).

yuja:
> It is currently difficult to deprecate merge tools. I also wonder how many people run `jj split --tool vscode` and get a non-functional interface.

I don't know the `vscode` problem, but I think we can freely introduce breaking change if it's already broken.

Regarding the config structure, I tend to prefer somewhat flattened form. To disable "edit", maybe we can explicitly nullify `edit-args = []`? It will be overwritten by the user-configured `edit-args` if any, so it's a bit safer than adding separate `edit.error` field, which can disable user-configured tool.

Related
https://github.com/martinvonz/jj/issues/1285

ilyagr:
> To disable "edit", maybe we can explicitly nullify edit-args = []?

I think I can do this first. IMO, `edit-args = false` would look clearer on first reading if we can implement it (I think I mentioned this somewhere before).

[...]

PR #6198, opened 2025-04-01
cli completion: complete `--tool` for diff tools, diff editors, and merge editors

ilyagr:
# Checklist

If applicable:
- [ ] ~~I have updated `CHANGELOG.md`~~

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -490,9 +490,12 @@ pub fn diff_editors() -> Vec<CompletionCandidate> {
 /// can't tell which these are. So, this not reliable, but probably good enough
 /// for command-line completion.
 pub fn diff_formatters() -> Vec<CompletionCandidate> {
-    // TODO: Add :git and , :stat, etc (next commit)
+    let builtin_format_kinds = crate::diff_util::all_builtin_diff_format_names();
     with_jj(|_, settings| {
-        Ok(configured_merge_tools(settings)
+        Ok(builtin_format_kinds
+            .iter()
+            .map(|s| s.as_str())
+            .chain(configured_merge_tools(settings))
             .map(CompletionCandidate::new)
             .collect())
     })
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -36,6 +36,8 @@ use crate::config::default_config_layers;
 use crate::config::ConfigArgKind;
 use crate::config::ConfigEnv;
 use crate::config::CONFIG_SCHEMA;
+use crate::merge_tools::configured_merge_tools;
+use crate::merge_tools::get_external_tool_config;
 use crate::revset_util::load_revset_aliases;
 use crate::ui::Ui;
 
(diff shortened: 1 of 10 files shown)


cli/src/merge_tools/mod.rs
```diff
@@ -172,6 +172,36 @@ fn get_tool_config(
...
+    std::iter::once(":builtin").chain(settings.table_keys("merge-tools").filter(|name| {
+        settings
+            .get::<Vec<String>>(ConfigNamePathBuf::from_iter([
+                "merge-tools",
+                name,
+                "merge-args",
+            ]))
+            .is_ok()
+    }))
+}
```
yuja, line 203: Maybe better to inline these in complete.rs? If we add `:ours`/`:theirs`, we might want to attach help text to these.

I think we can fully deserialize `ExternalMergeTool` to test `.merge_args`. The extra cost wouldn't matter.

ilyagr: I cleaned it up a bit. I added a new constructor for MergeEditor so that complete.rs does not have to deal with internals, let me know if you have better ideas for naming or the structure. It seemed mildly nicer than having the relevant structures implement Default. It didn't seem worth splitting MergeEditor into two types, but maybe it is?

cli/tests/test_completion.rs
```diff
@@ -741,6 +741,59 @@ fn test_template_alias() {
...
+    diffedit3-ssh
+    difft
+    kdiff3
+    meld
+    meld-3
+    mergiraf
+    smerge
+    vimdiff
+    vscode
+    vscodium
```
ilyagr, line 779: I'm not sure whether I want to merge this before or after the release. If we merge it after the release, we could implement something like https://github.com/jj-vcs/jj/issues/4549 to remove non-functional diff editors like `vscode` and `mergiraf` from the list.

ilyagr: Thanks for pointing me to it! I think that would help with #4549, yes. I don't think it affects this PR much.

cli/src/merge_tools/mod.rs
yuja: Doesn't `get_external_tool_config()` do the job?

ilyagr: `MergeEditor::new` that I'm using does call `get_external_tool_config`, and then also checks that the merge editor is configured.

The naming is confusing.

yuja: > It seems that `MergeEditor::new` that I'm using does call `get_external_tool_config`, and then also checks that the merge editor is configured. I could call it explicitly and check that the merge-args are non-empty, but I thought the point was to not duplicate that logic (we'll probably need cleverer logic for diff editors and diff tools soon).

I think a bit of duplication is good. We don't need `MergeEditor` abstraction in completion code, and it's weird that we have to create a "dummy".

cli/src/complete.rs
```diff
@@ -412,6 +414,46 @@ pub fn workspaces() -> Vec<CompletionCandidate> {
             .collect())
     })
 }
+pub fn merge_tools() -> Vec<CompletionCandidate> {
+    with_jj(|_, settings| {
+        Ok(std::iter::once(":builtin")
```
scott2000, line 419: This should probably also include `:ours` and `:theirs` now that they have been added.

cli/src/merge_tools/mod.rs
ilyagr: OK, I tried to refactor following your suggestion, as best I understood it.

I also added a commit to complete diff tools such as `:diff` that were introduced since I made this PR.

cli/src/complete.rs
```diff
@@ -451,6 +453,50 @@ pub fn workspaces() -> Vec<CompletionCandidate> {
             .collect())
     })
 }
+pub fn merge_tools() -> Vec<CompletionCandidate> {
```
yuja: nit: rename to `merge_editors()` for consistency

cli/src/complete.rs
```diff
@@ -451,6 +453,50 @@ pub fn workspaces() -> Vec<CompletionCandidate> {
...
+    })
+}
+
+/// Approximate list of known diff tools
+///
+/// Diff tools can be used without configuration. Some merge tools that are
+/// configured for 3-way merging may not work for diffing/diff editing, and we
+/// can't tell which these are. So, this not reliable, but probably good enough
+/// for command-line completion.
+pub fn diff_tools() -> Vec<CompletionCandidate> {
```
yuja: nit:  `diff_formatters()` per config name.

cli/src/diff_util.rs
```diff
@@ -211,7 +223,7 @@ impl BuiltinFormatKind {
         }
     }
 
-    fn to_arg_name(self) -> &'static str {
+    pub fn to_arg_name(self) -> &'static str {
```
yuja: nit: maybe better to add a list of names (or `(name, variant)` pairs) instead of `ALL_VARIANTS`?

An "arg" name might be different from `:<name>`. If we add builtin format that doesn't support `--<name>` argument, `to_arg_name()` will return `Option<&'static str>`.

yuja (approved):
Thanks.

cli/src/diff_util.rs
ilyagr: > If we add builtin format that doesn't support --<name> argument, to_arg_name() will return Option<&'static str>.

~~I'm not sure I follow; I imagined all builtin formats would start with `:`, similarly to `:builtin`, `:ours`, etc.~~ **Update:** Ah, I think I see, `to_arg_name` is probably mainly used in the Clap config.

But I don't mind making a list of all names. I think I'll then add an `BuiltinFormatKind::all_names()` function returning a Vec. I just learned that you can actually generate the list automatically in a const context, but it looks a little ugly (though we can make it more constant later if we want).

ilyagr: I guess I'll change the function to `all_builtin_format_kinds`, to avoid making `BuiltinFormatKind` public.

PR #6205, opened 2025-04-01
cli: warn on startup for vimdiff used as a diff editor

ilyagr:
This is more visible than a comment. Once this is supported, I'm considering removing this entirely. I doubt many people intentionally use it, but if so, they can easily remove this warning in their config.

Another option would be to set up a file-by-file diff editing mode, for which vimdiff would be fine.


steadmon (approved):
LGTM

PR #6207, opened 2025-04-01
cli show: add `--no-patch` flag

### LINKED ISSUE #5036: `jj show -T` always generates diff
## Description

When I use `-T` with `jj show`, the diff is always generated and printed after the template output.

In a jj repository with changes to the working copy.

```sh
jj show -T 'self.empty()'
```

## Expected Behavior

Outputs only `true` or `false`.

## Actual Behavior

Outputs `true` or `false` followed by the entire diff of changes in the working copy.

## Specifications

- Platform: Linux x86_64
- Version: jj 0.24.0-32d2a85539254e9d96f9819072fa5c6ac70dd1e4

gechr:
Closes #5036

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj show` patches can now be suppressed with `--no-patch`.
+
 ### Fixed bugs
 
 ## [0.28.1] - 2025-04-04
--- a/cli/src/commands/show.rs
+++ b/cli/src/commands/show.rs
@@ -50,6 +50,9 @@ pub(crate) struct ShowArgs {
     template: Option<String>,
     #[command(flatten)]
     format: DiffFormatArgs,
+    /// Do not show the patch
+    #[arg(long, conflicts_with = "DiffFormatArgs")]
+    no_patch: bool,
 }
 
 #[instrument(skip_all)]
@@ -70,6 +73,8 @@ pub(crate) fn cmd_show(
     let mut formatter = ui.stdout_formatter();
     let formatter = formatter.as_mut();
     template.format(&commit, formatter)?;
-    diff_renderer.show_patch(ui, formatter, &commit, &EverythingMatcher, ui.term_width())?;
+    if !args.no_patch {
+        diff_renderer.show_patch(ui, formatter, &commit, &EverythingMatcher, ui.term_width())?;
+    }
     Ok(())
 }
--- a/cli/tests/test_show_command.rs
+++ b/cli/tests/test_show_command.rs
@@ -256,6 +256,44 @@ fn test_show_with_template() {
 }
 
 #[test]
+fn test_show_with_template_no_patch() {
+    let test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let work_dir = test_env.work_dir("repo");
+    work_dir.run_jj(["new", "-m", "a new commit"]).success();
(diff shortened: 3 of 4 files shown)


yuja:
As I said, `-T` shouldn't be the flag to control the diff output. If we really need this behavior, we should add `--no-patch` or something (or make `jj show` not imply `-p` by default.)

cli/src/commands/show.rs
```diff
@@ -50,6 +50,9 @@ pub(crate) struct ShowArgs {
     template: Option<String>,
     #[command(flatten)]
     format: DiffFormatArgs,
+    /// Do not show the patch.
+    #[arg(long, short = 'P')]
```
yuja, line 54: nit: 

* add `conflicts_with = "DiffFormatArgs"` to reject `--no-patch --git`, etc.
* remove period from summary line of help text
* remove short flag for now? I'm not sure if we need short flag.

cli/src/commands/show.rs
```diff
@@ -65,11 +68,15 @@ pub(crate) fn cmd_show(
         None => workspace_command.settings().get_string("templates.show")?,
     };
     let template = workspace_command.parse_commit_template(ui, &template_string)?;
-    let diff_renderer = workspace_command.diff_renderer_for(&args.format)?;
-    ui.request_pager();
+    if !args.no_patch {
+        ui.request_pager();
+    }
```
yuja, line 73: I don't think it makes sense to suppress pager by `--no-patch`.

cli/src/commands/show.rs
```diff
@@ -65,11 +68,15 @@ pub(crate) fn cmd_show(
         None => workspace_command.settings().get_string("templates.show")?,
     };
     let template = workspace_command.parse_commit_template(ui, &template_string)?;
-    let diff_renderer = workspace_command.diff_renderer_for(&args.format)?;
```
yuja, line 68: nit: better to check configuration error here (before spawning pager)

yuja (commented):
Thanks. I think `show --no-patch` is good, but I'll leave this open for a while to gather comments.

cli/src/commands/show.rs
yuja: Out of curiosity, what's your use case? (I assumed `show --no-patch -T` could be useful in scripting, but wouldn't be used much as interactive command.)

I don't have strong feeling either, but `jj show -P` doesn't sound like an option to suppress patch.

cli/tests/test_show_command.rs
```diff
@@ -255,6 +255,52 @@ fn test_show_with_template() {
...
 }
 
+#[test]
+fn test_show_with_template_no_patch() {
+    let test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let repo_path = test_env.env_root().join("repo");
+
+    test_env
+        .run_jj_in(&repo_path, ["new", "-m", "a new commit"])
```
martinvonz, line 265: nit: use the new `test_env.work_dir()` API

CHANGELOG.md
```diff
@@ -132,6 +132,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj resolve` now accepts new built-in merge tools `:ours` and `:theirs`.
   These merge tools accept side #1 and side #2 of the conflict respectively.
 
+* `jj show` patches can now be suppressed with `--no-patch`.
```
yuja, line 135: reminder: you'll need to relocate the changelog entry.

We'll have a point release today or tomorrow, so please hold this change until then.

yuja (approved):
Looks good, thanks.

PR #6208, opened 2025-04-02
[jj fix] Make fix command more powerful: FixResult allows reporting semantic errors.

drieber:
# Checklist

If applicable:


lib/src/fix.rs~
PhilipMetzger, line 1: nit: this file has conflicts and probably shouldn't be in this commit.

lib/src/fix.rs
```diff
@@ -291,11 +322,29 @@ pub fn fix_files(
...
+                                // TODO: include this information in FixSummary, keyed by commit_id
+                                // and repo_path.
+                                todo!()
+                            }
+                            FixResult::FileTooLarge => {
+                                // TODO: include this information in FixSummary, keyed by commit_id
+                                // and repo_path.
+                                todo!()
+                            }
+                        }
```
PhilipMetzger: nit: I think its better to just add the `FixResult::Fixed` and `FixResult::Unchanged` cases for this PR and then do the rest one by one or in followups

PhilipMetzger (commented):
I think this looks generally good, but there are some discussion points left.

lib/src/fix.rs
drieber: I like your suggestion. I will break it up, unless I get other feedback. I am going to be on vacation starting tomorrow coming back on April 13, so it is possible I won't get to it until I get back.

PhilipMetzger: > I like your suggestion.

Thanks. 

> I am going to be on vacation starting tomorrow coming back on April 13, so it is possible I won't get to it until I get back.

Take your time and enjoy your vacation.

lib/src/fix.rs
```diff
@@ -58,6 +58,35 @@ pub struct FileToFix {
...
+        /// The id of the new file contents.
+        new_file_id: FileId,
+    },
+    /// The file content was already correct.
+    Unchanged,
+    /// The file cannot be fixed due to errors in the current content.
+    InvalidContent(InvalidContentInfo),
+    /// The file cannot be processed because the content is too large.
+    FileTooLarge,
+}
```
hooper, line 88: This doesn't seem right to me. Whether the file content changed is a separate concern from propagating info/warning/error messages. This also seems generally more structured than is useful for the subprocessing model we currently have, although I understand how it would be useful at Google. Keep in mind the outputs for the subprocess model are just stdout, stderr, and exit code (effectively two strings and a bool). The common tools in this space (`clang-format` et al, but also potentially `sort`, `sed`, etc) have no common structured protocol.

Here's the Mercurial implementation for reference:
https://repo.mercurial-scm.org/hg/file/tip/hgext/fix.py#l750

For jj, I'm imagining something more like:

```
struct FixResult {
    new_content: Result<NewContent, ???Error>
    // Could be empty, or one value for accumulated stderr output,
    // or multiple values from a more structured source.
    messages: Vec<Message>,
}

// If the caller always has the old FileId handy, maybe we just return
// the new FileId and let them do the comparison? I'm not sure if
// there's a strong reason for the enum anymore.
enum NewContent {
    Unchanged,
    Changed { new_file_id: FileId },
}

// This could just be a string if we don't want it be in the business of
// designing a generic structure, but then we may need to worry
// about fix_fn implementers trying to serialize things as JSON etc.
// and making those decisions for us.
struct Message {
    // Optional fields like arbitrary text, line numbers, etc. that `fix_fn`
    // can choose to emit, and callers can choose to display or ignore.
    // This is worth thinking about carefully, but at least it's isolated
    // from the success/failure/file content aspects of the FixResult.
    text: Option<String>,
    line_number: Option<usize>,
    ...
}
```

I discussed this a bit with David. We came up with different structs, but I think we generally agree on these ideas.

martinvonz: > This also seems generally more structured than is useful for the subprocessing model we currently have, although I understand how it would be useful at Google.

I had the same feeling. Could the `FileFixer` instance we use on our server instead have some state for that? We can update the interface to make `FileFixer::fix_files()` take a mutable reference. I created #6226 for this idea. Do you think that would work?

yuja: > Could the `FileFixer` instance we use on our server instead have some state for that? We can update the interface to make `FileFixer::fix_files()` take a mutable reference.

Maybe they're separate concerns? We'll still need to fix the subprocess implementation to report stderr messages properly (instead of inheriting stderr), which wouldn't be the job of `ParallelFileFixer`.

lib/src/fix.rs
```diff
@@ -58,6 +58,29 @@ pub struct FileToFix {
...
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub struct FixResult {
+    /// The FileId of the file after applying the fixer. This will be a new
+    /// FileId if the contents changed, or the same as before if nothing
+    /// changed.
+    pub file_id: FileId,
+    /// Could be empty, or one value for accumulated stderr output,
+    /// or multiple values from a more structured source.
+    pub messages: Vec<FixerMessage>,
+}
```
PhilipMetzger, line 71: Nit: I liked the previous variant a bit more, since each state was represented separately. I think it will be worthwhile to move the inner state of the result into a enum in the future. 

For now, I think it'd be best if the `file_id` field was a `Option<FiledId>` so it models the unchanged part correctly and gives the caller a heads up that it did nothing.

lib/src/fix.rs
```diff
@@ -291,11 +307,20 @@ pub fn fix_files(
...
+                                id: fix_result.file_id.clone(),
+                                executable: *executable,
+                            });
+                        } else {
+                            // TODO: consider counting files examined but
+                            // unchanged in FixSummary.
+                        }
+                        if !fix_result.messages.is_empty() {
+                            // TODO: Add messages to FixSummary (but put an
+                            // upper bound on this?).
```
PhilipMetzger, line 322: How do you want to achieve that? Since I think it relates to `clang-format` spewing to many errors, I think @hooper also can chime in.

PhilipMetzger (commented):
I think we're going in the right direction here, still have some nits a question.

martinvonz:
#6226 may work as an alternative to this PR. What do you think, @hooper? (@drieber is on vacation for a week or so)

lib/src/fix.rs
martinvonz: I see. Would the subprocessing fixers return at most one `FixerMessage` then and not include a line number? It's mostly the `FixerMessage` that seems a bit weird to have outside of Google, but maybe that's okay.

yuja: > I see. Would the subprocessing fixers return at most one `FixerMessage` then and not include a line number?

Or `file_messages: Vec<String>, annotations: Vec<{line_number, text}>`?

martinvonz: I think #6226 provides all the flexibility we need for our server at Google. So I think this PR should focus on what's useful for the CLI outside of Google for now. If we think we can get line numbers from formatters, then it makes sense to include that. Otherwise I would wait until we have other callers that can produce line numbers. Does that make sense?

lib/src/fix.rs
drieber: I addressed a couple of the comments, while leaving some unresolved.

In particular, I am not sure how I wound put an upper bound on fix messages.

Please note I like the much simpler approach of https://github.com/jj-vcs/jj/pull/6226. I don't mind withdrawing this PR in favor of that one.

drieber: I addressed a couple of the comments, while leaving some unresolved.

In particular, I am not sure how I wound put an upper bound on fix messages.

Please note I like the much simpler approach of https://github.com/jj-vcs/jj/pull/6226. I don't mind withdrawing this PR in favor of that one.

lib/src/fix.rs
arxanas: [for future consideration] It sounds like we don't necessarily need to move forwards with this approach, but in the future, we could copy/adapt an existing tool's "diagnostics" schema.

- Abstractly: This would tell us whether we expect tools to generally be able to provide line numbers or severities. Many of them may also be able to emit specific formats.
- As a concrete example: VS Code's "problem matchers" (see [Tasks](https://code.visualstudio.com/docs/debugtest/tasks)) define ways to extract diagnostics from semi-structured command output.
  - For certain tools, we may even be able to copy the literal problem matcher definition corresponding to the tool.
- As a concrete example: Here's the [Language Server Protocol's schema (v3.17)](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic). Among other things, it includes ranges (not just line numbers) and severities:

```ts
export interface Diagnostic {
	/**
	 * The range at which the message applies.
	 */
	range: Range;

	/**
	 * The diagnostic's severity. To avoid interpretation mismatches when a
	 * server is used with different clients it is highly recommended that
	 * servers always provide a severity value. If omitted, it’s recommended
	 * for the client to interpret it as an Error severity.
	 */
	severity?: DiagnosticSeverity;

	/**
	 * The diagnostic's code, which might appear in the user interface.
	 */
	code?: integer | string;

	/**
	 * An optional property to describe the error code.
	 *
	 * @since 3.16.0
	 */
	codeDescription?: CodeDescription;

	/**
	 * A human-readable string describing the source of this
	 * diagnostic, e.g. 'typescript' or 'super lint'.
	 */
	source?: string;

	/**
	 * The diagnostic's message.
	 */
	message: string;

	/**
	 * Additional metadata about the diagnostic.
	 *
	 * @since 3.15.0
	 */
	tags?: DiagnosticTag[];

	/**
	 * An array of related diagnostic information, e.g. when symbol-names within
	 * a scope collide all definitions can be marked via this property.
	 */
	relatedInformation?: DiagnosticRelatedInformation[];

	/**
	 * A data entry field that is preserved between a
	 * `textDocument/publishDiagnostics` notification and
	 * `textDocument/codeAction` request.
	 *
	 * @since 3.16.0
	 */
	data?: LSPAny;
}
```

PR #6209, opened 2025-04-02
(the command called) Fix: always rewrite descendants of fixed commits

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/fix.rs
+++ b/lib/src/fix.rs
@@ -282,7 +282,7 @@ pub fn fix_files(
         let repo_paths = commit_paths.get(&old_commit_id).unwrap();
         let old_tree = rewriter.old_commit().tree()?;
         let mut tree_builder = MergedTreeBuilder::new(old_tree.id().clone());
-        let mut changes = 0;
+        let mut has_changes = false;
         for repo_path in repo_paths {
             let old_value = old_tree.path_value(repo_path)?;
             let new_value = old_value.map(|old_term| {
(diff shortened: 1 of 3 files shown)


lib/tests/test_fix.rs
```diff
@@ -326,6 +326,50 @@ fn test_unchanged_file_is_fixed() {
     assert_eq!(*new_commit_b.tree_id(), expected_tree_b.id());
 }
 
+/// If a descendant is already correctly formatted, it should by rebased without
```
ilyagr, line 329: typo: "be rebased". I'd say "its snapshot should be preserved and not inherit any fixes done to the parent".

lib/tests/test_fix.rs
ilyagr, line 1: Add a changelog entry?

lib/tests/test_fix.rs
```diff
@@ -357,17 +357,22 @@ fn test_already_fixed_descendant() {
     )
     .unwrap();
 
-    // TODO: Should have fixed both commits
-    assert_eq!(summary.rewrites.len(), 1);
+    assert_eq!(summary.rewrites.len(), 2);
```
ilyagr, line 360: Optional: this is confusing, I wouldn't expect the second commit to be counted here in this situation. Perhaps a comment would be nice, or just delete some of this and leave the tree test below.

ilyagr (approved):
This makes sense at a glance. I'm approving it a bit rushedly, in case you want to get this in before the release (and if I missed some problem, I don't think it'd be a severe blow to the release).

yuja (approved):

lib/tests/test_fix.rs
martinvonz: Done (slightly rephrased)

lib/tests/test_fix.rs
martinvonz: Done

lib/tests/test_fix.rs
martinvonz: The commit is rewritten, so I think that makes sense. But I agree that it makes sense to not consider the second commit fixed. I've updated the code and test. Feel free to take another look.

lib/src/fix.rs
```diff
@@ -302,12 +302,14 @@ pub fn fix_files(
...
-                changes += 1;
+                has_changes = true;
             }
         }
         summary.num_checked_commits += 1;
-        if changes > 0 {
+        if has_changes {
             summary.num_fixed_commits += 1;
+        }
+        if has_changes || rewriter.parents_changed() {
```
yuja, line 312: If we handle `has_changes` and `.parents_changed()` differently, maybe we can omit `.write_tree()`, etc. at all?

```rust
if has_changes {
    ...
} else if rewriter.parents_changed() {
    rewriter.reparent().write()?;
}
```

I'm not sure if "reparent" should be included in `rewrites`.

martinvonz: Makes sense to omit `.write_tree()`. I've updated that. OTOH, maybe we should optimize `MergedTreeBuilder` so it doesn't do any writes if nothing changed.

We don't use `rewrites` for anything yet, but I think it makes sense to include reparented commits there. I would expect all rewritten commits to be included.

PR #6220, opened 2025-04-02
release: 0.28.0

arxanas:
Let me know if we want to add/remove anything in the the release highlights.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,27 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+### Deprecations
+
+### New features
+
+### Fixed bugs
+
+## [0.28.0] - 2025-04-02
+
+### Release highlights
+
+* jj's configuration can now be split into multiple files more easily.
+
+* `jj resolve` now accepts built-in tools `:ours` and `:theirs`.
+
+* In colocated repos, newly-created files will now appear in `git diff`.
+
+* A long-standing bug relating to empty files in the built-in diff editor was
+  fixed. [#3702](https://github.com/jj-vcs/jj/issues/3702)
+
+### Breaking changes
+
 * The minimum supported Rust version (MSRV) is now 1.84.0.
 
 * The `git.push-branch-prefix` config has been removed in favor of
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,26 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+
+### New features
+
+### Fixed bugs
+
+## [0.28.0] - 2025-04-02
+
+### Release highlights
+
+* All outstanding "pinned" issues in the GitHub repository have been resolved, including a long-standing bug relating to empty files in the built-in diff editor.
```
martinvonz, line 25: nit: wrap at ~80 columns (we should set up a formatter. i don't remember what the latest on that is)

martinvonz (approved):
Thanks!

CHANGELOG.md
ilyagr, line 25: (Optional): I'd either skip these first point, or link to some of the relevant bugs.

In the latter case, it might be sufficient to talk about the one specific bug you called out.

If we want to mention them all, perhaps we could say something like:

> Long-standing bugs relating to git remote compatibility (#XXXX, #YYYY) and empty files in the built-in git editor (#ZZZZ) were resolved.

That said, we could keep what you have if it's easier.

CHANGELOG.md
```diff
@@ -12,6 +12,26 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
...
+
+* Many improvements to configuration management.
```
ilyagr, line 27: (Optional) Perhaps "jj's configuration can now be split into multiple files more easily than before"?

CHANGELOG.md
arxanas: @ilyagr I couldn't remember what all of the historical pinned files were, which is the main reason I didn't list them all 🤣. I changed this to just list that one bug by link and demoted it to a later point in the list.

ilyagr: Setting up a formatter is related to https://github.com/jj-vcs/jj/pull/6081: would `mise` make it easier for end-users to install `dprint` or `prettier`? (IIRC those were the most promising formatters for Markdown; even if we go with `prettier`, it might be easier to run it via `dprint`)

We could then run the formatter with either `mise` (or equivalent) or with `jj fix`.

CHANGELOG.md
arxanas: I've updated to that wording, but dropped "than before" because it seems redundant with "now" to me.

PR #6226, opened 2025-04-03
fix: make `FileFixer::fix_files()` take `&mut self`

martinvonz:
It can be useful for fixers to record some information about the inputs they processed. For example, the `FileFixer` we use in our server at Google get more detailed error information back from its formatter tools. We may want to record these errors in our `FileFixer`'s state so we can then return that to the client.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -143,7 +143,7 @@ pub(crate) fn cmd_fix(
         .to_matcher();
 
     let mut tx = workspace_command.start_transaction();
-    let parallel_fixer = ParallelFileFixer::new(|store, file_to_fix| {
+    let mut parallel_fixer = ParallelFileFixer::new(|store, file_to_fix| {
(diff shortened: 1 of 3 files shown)


martinvonz:
This is meant as an alternative to part of all of #6208. Leaving it as draft until I hear from hooper or drieber.

drieber:
Thank you. I have not tried this approach you end-to-end but I am pretty sure you are right, this would allow me to keep the state in google's server implementation of FileFixer. With this approach I don't think I need any of the changes in #6208.

#6208 makes the API more complex, so I don't mind dropping #6208 in favor of this PR.

arxanas (approved):
LGTM as a maintainer; it sounds like we're also awaiting an approval from one of the aforementioned people.

hooper:
I don't know if I like the statefulness, but I like the idea of not making the kind of decisions discussed in #6208 :)

hooper (approved):

PR #6227, opened 2025-04-03
merge_tools: fix `:ours` and `:theirs` merge tools

scott2000:
These merge tools didn't work properly on conflicted trees with more than two sides since they didn't simplify the conflict before resolving.

https://discord.com/channels/968932220549103686/1357321005642678333

# Checklist

If applicable:

### DIFF
--- a/cli/src/merge_tools/mod.rs
+++ b/cli/src/merge_tools/mod.rs
@@ -316,6 +316,7 @@ struct MergeToolFile {
     repo_path: RepoPathBuf,
     conflict: MergedTreeValue,
     file_merge: Merge<Option<FileId>>,
+    simplified_file_merge: Merge<Option<FileId>>,
     simplified_file_content: Merge<BString>,
 }
 
@@ -347,6 +348,7 @@ impl MergeToolFile {
             repo_path: repo_path.to_owned(),
             conflict,
             file_merge,
+            simplified_file_merge,
             simplified_file_content,
         })
     }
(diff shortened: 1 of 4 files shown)


cli/src/merge_tools/mod.rs
```diff
@@ -455,7 +455,15 @@ fn pick_conflict_side(
...
-            Merge::resolved(merge_tool_file.conflict.get_add(add_index).unwrap().clone()),
+            Merge::resolved(
+                merge_tool_file
+                    .conflict
+                    .clone()
+                    .simplify()
+                    .get_add(add_index)
+                    .unwrap()
+                    .clone(),
+            ),
```
yuja, line 466: Perhaps, we should fix inconsistency in `MergeToolFile` by renaming some variables. It's weird that `content` is simplified, but the other two aren't.

It might also make sense to not use `MergeToolFile` at all for `:theirs`/`:ours`. We can't rely on `MergeToolFile::from_tree_and_path()` rejecting multi-way conflicts. The executable bits may differ in addition to file contents.

scott2000: I updated the naming of the variables in `MergeToolFile` to make it more clear which are simplified and which aren't.

I tried avoiding `MergeToolFile` for `:theirs` and `:ours`, but it made things a lot more complicated, so I'm not sure it's worth the complexity. I think it might also have led to a situation where it could complain about a conflict having too many sides even though `jj resolve --list` shows it as having 2-sides (due to the executable bits differing), which would be difficult to explain to the user.

To make it simpler, I just changed the code to use the simplified file IDs, so now it works more similarly to how external merge tools work. This should hopefully be less surprising behavior for executable conflicts, since it's how it would work for external merge tools anyway. Can you take a look and let me know what you think?

yuja (approved):
Looks good, thanks!

PR #6231, opened 2025-04-03
cargo: don’t enable `testutils/git2` feature in `jj-cli`

emilazy:
Since we depend on `jj-lib/git2`, which sets `testutils/git2`, this should be redundant. Why it makes `cargo install` happy, I don’t know, but it does.

# Checklist

If applicable:


emilazy:
Nope, this doesn’t seem right: `#[cfg(not(feature = "git2"))] fail-to_compile;` in `lib/test` errors out when you try to build the tests.

PR #6235, opened 2025-04-03
cli: Add config setting for bookmark list sort order

### LINKED ISSUE #3831: FR: Option to sort listed branches based on last commit date

zummenix:
This is a conclusion of #5849.
Config setting to list bookmarks in specified order.

Closes #3831

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj show` patches can now be suppressed with `--no-patch`.
 
+* Added `ui.bookmark-list-sort-keys` setting to configure default sort keys for the
+  `jj bookmark list` command.
+
 ### Fixed bugs
 
 ## [0.28.2] - 2025-04-07
--- a/cli/src/commands/bookmark/list.rs
+++ b/cli/src/commands/bookmark/list.rs
@@ -23,6 +23,7 @@ use clap_complete::ArgValueCandidates;
 use itertools::Itertools as _;
 use jj_lib::backend;
 use jj_lib::backend::CommitId;
+use jj_lib::config::ConfigValue;
 use jj_lib::ref_name::RefName;
 use jj_lib::repo::Repo as _;
 use jj_lib::revset::RevsetExpression;
@@ -120,6 +121,8 @@ pub struct BookmarkListArgs {
     /// Suffix the key with `-` to sort in descending order of the value (e.g.
     /// `--sort name-`). Note that when using multiple keys, the first key is
     /// the most significant.
+    ///
+    /// This defaults to the `ui.bookmark-list-sort-keys` setting.
     #[arg(long, value_name = "SORT_KEY", value_enum, value_delimiter = ',')]
     sort: Vec<SortKey>,
 }
@@ -240,9 +243,16 @@ pub fn cmd_bookmark_list(
         }
     }
 
+    let sort_keys = if args.sort.is_empty() {
+        workspace_command
+            .settings()
+            .get_value_with("ui.bookmark-list-sort-keys", parse_sort_keys)?
+    } else {
+        args.sort.clone()
+    };
     let store = repo.store();
     let mut commits: HashMap<CommitId, Arc<backend::Commit>> = HashMap::new();
-    if args.sort.iter().any(|key| key.is_commit_dependant()) {
+    if sort_keys.iter().any(|key| key.is_commit_dependant()) {
         commits = bookmark_list_items
             .iter()
             .filter_map(|item| item.primary.target().added_ids().next())
@@ -254,7 +264,7 @@ pub fn cmd_bookmark_list(
             })
             .try_collect()?;
     }
-    sort(&mut bookmark_list_items, &args.sort, &commits);
+    sort(&mut bookmark_list_items, &sort_keys, &commits);
 
     ui.request_pager();
     let mut formatter = ui.stdout_formatter();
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* Added `ui.bookmark-list-order` configuration setting to configure default sort
```
cenviity, line 19: nit: ‘configuration setting’ can just be ‘configuration’, or ‘setting’ to avoid repetition with ‘configure’ in the same sentence.

docs/config.md
```diff
@@ -197,6 +197,25 @@ fill in things like BUG=, TESTED= etc.
...
+```toml
+[ui]
+bookmark-list-order = ["name"]
+```
+
+The configuration setting works identically to using `--sort` option for the
+`jj bookmark list`. The following sort keys are supported: `name`,
+`author-name`, `author-email`, `author-date`, `committer-name`,
+`committer-email`, `committer-date`. Suffix the key with `-` to sort in
+descending order. The first key is the most significant.
```
cenviity, line 215: We should mention multiple keys can be supplied here, before the note on the first key being the most significant.

cli/src/commands/bookmark/list.rs
```diff
@@ -354,6 +364,28 @@ impl SortKey {
     }
 }
 
+struct SortKeys(Vec<SortKey>);
+
+impl TryFrom<ConfigValue> for SortKeys {
+    type Error = String;
+
+    fn try_from(value: ConfigValue) -> Result<Self, Self::Error> {
```
yuja, line 372: nit: no need to add newtype for parsing. Just pass `parse_sort_keys()` function to `get_value_with()`.

We can also use `serde::Deserialize` to `get()` `Vec<SortKeys>` directly, but it might be tedious to maintain both clap's `name` and `serde(rename = ..)`.

cli/src/commands/bookmark/list.rs
```diff
@@ -354,6 +364,28 @@ impl SortKey {
...
+        if let Some(array) = value.as_array() {
+            let keys: Vec<SortKey> = array
+                .iter()
+                .map(|item| {
+                    item.as_str()
+                        .ok_or("Expected sort key as a string".to_owned())
+                        .and_then(|key| SortKey::from_str(key, true))
```
yuja, line 379: nit: `ignore_case` should be `false`?

cli/src/commands/bookmark/list.rs
```diff
@@ -240,9 +242,17 @@ pub fn cmd_bookmark_list(
         }
     }
 
+    let sort_keys = if args.sort.is_empty() {
+        let SortKeys(sort_keys) = workspace_command
+            .settings()
+            .get_value_with("ui.bookmark-list-order", TryInto::try_into)?;
```
yuja, line 248: nit: better to mention the config key in `--sort` help.

It would be nice if we can use "sort" instead of "order" consistently, but I don't have a good suggestion.

cli/src/commands/bookmark/list.rs
```diff
@@ -240,9 +242,17 @@ pub fn cmd_bookmark_list(
...
+        Cow::Owned(sort_keys)
+    } else {
+        Cow::Borrowed(args.sort.as_slice())
```
yuja, line 251: nit: simply `.clone()`?

cli/src/commands/bookmark/list.rs
zummenix: > nit: better to mention the config key in --sort help.

Good point!

> It would be nice if we can use "sort" instead of "order" consistently, but I don't have a good suggestion.

Yes! Changed it to `bookmark-list-sort-keys`

cli/src/commands/bookmark/list.rs
```diff
@@ -120,6 +121,9 @@ pub struct BookmarkListArgs {
     /// Suffix the key with `-` to sort in descending order of the value (e.g.
     /// `--sort name-`). Note that when using multiple keys, the first key is
     /// the most significant.
+    ///
+    /// Configuration `ui.bookmark-list-sort-keys` is ignored when using this
+    /// option.
```
yuja, line 126: nit: just say it is the default?

"This defaults to the `ui.bookmark-list-sort-keys` setting." (copied from `jj git push`)

cli/tests/test_bookmark_command.rs
```diff
@@ -2259,6 +2259,77 @@ fn test_bookmark_list_sort_multiple_keys() {
...
+            ])
+            .success();
+        work_dir
+            .run_jj(["bookmark", "create", "-r@", bookmark])
+            .success();
+    }
+
+    let template = r#"name ++ ": " ++ if(normal_target, normal_target.author().email()) ++ "\n""#;
+    insta::assert_snapshot!(work_dir.run_jj([
+        "--config=ui.bookmark-list-sort-keys=[\"author-email\", \"author-date-\"]",
```
yuja, line 2284: nit: you can use single quote `'author-email'`

yuja (approved):

PR #6236, opened 2025-04-04
tests: fix subprocess toggle in `git push --named` tests

emilazy:
@ilyagr @bsdinis You might want to review the added conditionals here. Is it expected that it’s different with `git.subprocess = false`?

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_git_push.rs
+++ b/cli/tests/test_git_push.rs
@@ -1101,8 +1101,8 @@ fn test_git_push_changes_with_name(subprocess: bool) {
     let test_env = TestEnvironment::default();
     set_up(&test_env);
     let work_dir = test_env.work_dir("local");
-    if subprocess {
-        test_env.add_config("git.subprocess = true");
+    if !subprocess {
+        test_env.add_config("git.subprocess = false");
     }
     work_dir.run_jj(["describe", "-m", "foo"]).success();
     work_dir.write_file("file", "contents");
@@ -1240,8 +1240,8 @@ fn test_git_push_changes_with_name_deleted_tracked(subprocess: bool) {
     // copy
     test_env.add_config(r#"revset-aliases."immutable_heads()" = "none()""#);
     let work_dir = test_env.work_dir("local");
-    if subprocess {
-        test_env.add_config("git.subprocess = true");
+    if !subprocess {
+        test_env.add_config("git.subprocess = false");
     }
     // Create a second empty remote `another_remote`
     test_env
@@ -1354,8 +1354,8 @@ fn test_git_push_changes_with_name_untracked_or_forgotten(subprocess: bool) {
     let test_env = TestEnvironment::default();
     set_up(&test_env);
     let work_dir = test_env.work_dir("local");
-    if subprocess {
-        test_env.add_config("git.subprocess = true");
+    if !subprocess {
+        test_env.add_config("git.subprocess = false");
     }
     // Unset immutable_heads so that untracking branches does not move the working
     // copy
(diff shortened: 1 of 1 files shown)


ilyagr:
No, it's supposed to be the same. Weird; the libgit2 code seems to be doing something wrong. I'll need to look at what's going on.

Thanks for noticing this! I started the PR when `git.subprocess` was not the default, and it must have become the default even in tests by the time I merged the PR.

emilazy:
Yes. The `git.subprocess` test dance should be abstracted out, which is in fact what I do in my PR to add a warning, which is how I found this. Perhaps I should split that part out. Anyway, maybe we should merge this as‐is to document the current state of things for now, possibly with a comment about it being buggy? Otherwise `cargo insta` gets confused.

cli/tests/test_git_push.rs
```diff
@@ -1440,36 +1440,57 @@ fn test_git_push_changes_with_name_untracked_or_forgotten(subprocess: bool) {
...
+        ------- stderr -------
+        Changes to push to origin:
+          Add bookmark b1 to c9c824c88955
+        [EOF]
+        ");
     }
 
     // The bookmark is still forgotten
     let output = work_dir.run_jj(["bookmark", "list", "--all", "b1"]);
-    insta::allow_duplicates! {
```
ilyagr, line 1459: I think this diff an all the ones below are a consequence of the test not expecting that the bookmark was successfully pushed in the above test case.

cli/tests/test_git_push.rs
```diff
@@ -1440,36 +1440,57 @@ fn test_git_push_changes_with_name_untracked_or_forgotten(subprocess: bool) {
...
     // an error or some sort of warning?
     let output = work_dir.run_jj(["git", "push", "--named", "b1=@"]);
-    insta::allow_duplicates! {
-    insta::assert_snapshot!(output, @r"
-    ------- stderr -------
-    Changes to push to origin:
-      Add bookmark b1 to 10b6b209c4a3
-    [EOF]
-    ");
+    if subprocess {
```
ilyagr, line 1479: I don't think this diff is meaningful, as said in the above comment, but it's possible that fixing it will result in different, meaningful, diff similar to the first diff above.

cli/tests/test_git_push.rs
```diff
@@ -1440,36 +1440,57 @@ fn test_git_push_changes_with_name_untracked_or_forgotten(subprocess: bool) {
     // Make sure push still errors if we try to push a bookmark with the same name
     // to a different location.
     let output = work_dir.run_jj(["git", "push", "--named", "b1=@+"]);
```
ilyagr, line 1442: (Aside) I remember being vaguely surprised when I wrote this test case and it failed to push. That seems to be what we want, but my memory of the libgit2 pushing made me think it would have succeeded.

Unfortunately, the feeling wasn't strong enough to double-check that libgit2 is actually being used.

emilazy: FWIW I don’t think we have to ensure parity here. `git.subprocess = false` is in the process of deprecation, so it’s not too surprising if a new feature like `--named` acts subtly differently. (Which is not to say it’s not worth investigating, but I think it’d also be fine to just leave it like this until it’s removed.)

emilazy: (Whereby “leave it like this” I mean after the PR, since it’s weird to run the same test twice and have it not actually test `git2` in the other branch. I just mean it doesn’t seem critical to solve the divergence.)

ilyagr:
> Anyway, maybe we should merge this as‐is to document the current state of things for now, possibly with a comment about it being buggy?

I think that's OK. Do add a comment that the `libgit2` behavior is buggy, please. It's an annoying problem, but not a severe one; my guess is that most users wouldn't notice.

Thank you very much for finding this!

ilyagr (approved):

cli/tests/test_git_push.rs
ilyagr: You are probably right. I'm just annoyed feeling a trace of an egg on my face.

yuja: Perhaps, this is the difference between our ref merge logic and git-native force-with-lease. We'll need to move the branch sideway (instead of descendant) to make both git2 and subprocess fail.

ilyagr: Yes, I intentionally picked a borderline case. I guess you're implying it's not worth it to make the behavior consistent in the borderline case? 

I'm surprised this didn't cause a difference between libgit2 and subprocess behavior in some of the detailed tests I once wrote for force-with-lease.

yuja: > you're implying it's not worth it to make the behavior consistent in the borderline case?

Yes. The main reason afaik is that it's difficult to hook into `git` command to use our conflict resolution logic. The git2 behavior is more consistent with jj, though.

> I'm surprised the underlying cause didn't cause a difference between libgit2 and subprocess behavior in some of the detailed tests I once wrote for force-with-lease.

We have `if subprocess`, etc. in lib tests. ;)

ilyagr: Sad. But makes sense, I guess. Fixing it would be messy without git libification, we'd probably have to provide `git` with a way to involve jj as a push negotiation script.

emilazy:
Okay, I added some silly comments to make it clear that `git2` isn’t doing what we expected here.

PR #6238, opened 2025-04-04
cargo: bump `gix` to 0.71.0

emilazy:
Fix [GHSA-794x-2rpg-rfgr](https://github.com/jj-vcs/jj/security/advisories/GHSA-794x-2rpg-rfgr).

`gix::Repository::work_dir` was renamed to `workdir` (though strangely not the `gix::ThreadSafeRepository` version), and `lossy_config` is now off by default in all configurations.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Security fixes
+
+* Fixed SHA-1 collision attacks not being detected.
+  ([GHSA-794x-2rpg-rfgr](https://github.com/jj-vcs/jj/security/advisories/GHSA-794x-2rpg-rfgr))
+
 ### Release highlights
 
 ### Breaking changes
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -899,13 +899,13 @@ checksum = "0ce7134b9999ecaf8bcd65542e436736ef32ddca1b3e06094cb6ec5755203b80"
 
 [[package]]
 name = "flate2"
-version = "1.0.33"
+version = "1.1.1"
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Security fixes
+
+* Fixed SHA-1 collision attacks not being detected.
+  ([GHSA-794x-2rpg-rfgr](https://github.com/jj-vcs/jj/security/advisories/GHSA-794x-2rpg-rfgr))
```
yuja, line 14: Wrong link (or the advisory is not public yet)?

yuja (approved):

CHANGELOG.md
emilazy: Yes, not public yet. (But it’s just the “aimed at Jujutsu users/distributors” version of https://github.com/GitoxideLabs/gitoxide/security/advisories/GHSA-2frx-2596-x5r6.)

emilazy: I think @martinvonz can [give the maintainers team permission to manage advisories](https://docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/managing-security-managers-in-your-organization), BTW. Probably a good idea.

emilazy:
Fair warning: this will regress performance to some degree, though it’s mostly cloning where it hurts and we don’t use gitoxide for that anyway. Most things that this would slow down are already bottlenecked on DEFLATE, I think, so hopefully it won’t be noticeable in practice.

CHANGELOG.md
martinvonz: I think I've done that now

PR #6239, opened 2025-04-04
tests: add `TestEnvironment::with_git_subprocess` helper

emilazy:
Seems like this would have helped @ilyagr, so splitting it out of https://github.com/jj-vcs/jj/pull/6047.

# Checklist

If applicable:

### DIFF
--- a/cli/tests/common/test_environment.rs
+++ b/cli/tests/common/test_environment.rs
@@ -198,6 +198,17 @@ impl TestEnvironment {
         self.env_vars.insert(key.into(), val.into());
     }
 
+    // TODO: Remove with the `git.subprocess` setting.
+    pub fn with_git_subprocess(self, subprocess: bool) -> Self {
+        #[cfg(feature = "git2")]
+        if !subprocess {
+            self.add_config("git.subprocess = false");
+            return self;
+        }
+        assert!(subprocess);
+        self
+    }
+
     /// Sets up the fake editor to read an edit script from the returned path
     /// Also sets up the fake editor as a merge tool named "fake-editor"
     pub fn set_up_fake_editor(&mut self) -> PathBuf {
--- a/cli/tests/test_git_clone.rs
+++ b/cli/tests/test_git_clone.rs
@@ -42,12 +42,9 @@ fn set_up_git_repo_with_file(git_repo: &gix::Repository, filename: &str) {
 #[cfg_attr(feature = "git2", test_case(false; "use git2 for remote calls"))]
 #[test_case(true; "spawn a git subprocess for remote calls")]
 fn test_git_clone(subprocess: bool) {
-    let test_env = TestEnvironment::default();
+    let test_env = TestEnvironment::default().with_git_subprocess(subprocess);
     let root_dir = test_env.work_dir("");
     test_env.add_config("git.auto-local-bookmark = true");
-    if !subprocess {
-        test_env.add_config("git.subprocess = false");
-    }
     let git_repo_path = test_env.env_root().join("source");
     let git_repo = git::init(git_repo_path);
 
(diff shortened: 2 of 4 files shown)


cli/tests/test_git_clone.rs
```diff
@@ -1153,8 +1113,8 @@ fn test_git_clone_git2_warning() {
     Fetching into new repo in "$TEST_ENV/clone"
     bookmark: main@origin [new] tracked
     Setting the revset alias `trunk()` to `main@origin`
-    Working copy  (@) now at: sqpuoqvx 2ca1c979 (empty) (no description set)
```
ilyagr, line 1156: I think this needs to be reverted.

ilyagr (approved):
Thanks!

cli/tests/common/test_environment.rs
```diff
@@ -74,6 +74,18 @@ impl Default for TestEnvironment {
...
+    pub fn with_git_subprocess(subprocess: bool) -> Self {
+        #[cfg(feature = "git2")]
+        if !subprocess {
+            let test_env = Self::default();
+            test_env.add_config("git.subprocess = false");
+            return test_env;
+        }
+        assert!(subprocess);
+        Self::default()
+    }
```
yuja, line 87: nit: maybe better to add free function that does `test_env.add_config("git.subprocess = {subprocess}")`? It doesn't have to be set at construction time.

cli/tests/test_git_clone.rs
emilazy: Whoops, sorry. Will fix.

cli/tests/common/test_environment.rs
emilazy: I was considering `TestEnvironment::default().git_subprocess(subprocess)` instead, but it could also be a `()`‐returning method, though part of the idea here was to reduce boilerplate (the rest of it is to help ease the deprecation process).

Or are you thinking it’d not be a method on the `TestEnvironment` type at all? That feels a little odd to me, since it’s akin to `add_config` or `add_env_var`.

yuja: > though part of the idea here was to reduce boilerplate (the rest of it is to help ease the deprecation process that this was split out from, which is why the code structure here is a little odd).

Yeah, I thought it would be easier to delete separate function calls than rewriting back to `TestEnvironment::default()`. I don't have strong feeling because it's temporary, though.

> Or are you thinking it’d not be a method on the `TestEnvironment` type at all?

Yes. My feeling is that it's not so different than `set_up_fake_formatter()` in text_fix_command.rs.

emilazy: After trying this (though as a method, since it’s called from multiple files and is akin to `set_up_fake_editor`), the problem is that this will need to set an environment variable when we add the `git.subprocess = false` warning (to suppress the warning, so that the tests don’t all need to grow conditional snapshots with the warning), which then introduces a lot of `let mut test_env` churn. I guess an alternative would be to filter out the warning, but that makes testing the warning annoying; mostly I’m just writing stuff to be easy to delete later.

I’ve separated it out from `default()` by making it take `self`, though, which avoids the mutability problem and definitely seems nicer. I’ve left the weird code structure since it matches the other `git2` feature gates and (hopefully?) makes the later removal more atomic/reviewable. Hope it looks okay.

(I sort of feel like `add_config` should take `&mut self`, since it’s kinda using the filesystem as “interior mutability”, but that’s another matter.)

emilazy: > We could also use insta filtering with `settings.bind_to_scope()`. In that case, the settings object will have to be allocated by caller.

Right. That would be nicer than an environment variable if it was intended to be more permanent, but for a transition mechanism I prefer doing the less elegant but less invasive thing.

> I'd rather make `add_env_var()` non-mut. It was annoying that `set_up_fake_editor()` in middle of tests breaks borrowing. Every `.run_jj*()` call is technically mutable op in a way that the timestamp seed is updated.

Yeah, that would also be reasonable. I just don’t think that going via operating system APIs or not should be the criterion that determines the signature. (This is a general thing I find awkward about Rust, though.)

PR #6242, opened 2025-04-04
cargo: use `gix/zlib-rs` feature

emilazy:
This uses `zlib-rs`, a native Rust library that is comparable in performance to `zlib-ng`. Since there’s no complicated C build and gitoxide only has one hashing backend now, this lets us drop our `packaging` feature without adding any awkward build requirements.

`zlib-rs` is generally faster at decompression than `zlib-ng`, and faster at compression on levels 6 and 9; see <https://trifectatech.org/blog/zlib-rs-is-faster-than-c/> for details.

I couldn’t get reliable‐looking benchmark results out of my temperamental laptop; `hyperfine` seemed to think that some random `jj` workloads I tested might be slightly slower than with `zlib-ng`, but it wasn’t unambiguously distinguishable from noise, so I’d like to see measurements from others.

It’s certainly a lot faster than the previous default, and I think it’s likely that `zlib-rs` will continue to get faster and that it’s more than worth avoiding the headaches of a native library with a CMake build dependency. (Though on the other hand, if distributions move in the direction of shipping `zlib-ng` by default, maybe there will be more motivation to make `libz-ng-sys` support system libraries.)

---

Not intended for merge before 0.28.1, and semi‐RFC until others do some measurements. I think that the build/packaging benefits here are worth it and that we can expect `zlib-rs` to continue to be competitive with `zlib-ng` in the long run, but if it seems like there might be a slight performance compromise and people consider that unacceptable, then we could maybe make `zlib-rs` the default but keep the `packaging` feature for `zlib-ng`.

# Checklist

If applicable:


emilazy:
I’ve replaced the now‐absent `ISC` with `Zlib` in `deny.toml` to fix the licence‐checking error. They are both in Google’s [`notice` list](https://opensource.google/documentation/reference/thirdparty/licenses#notice), so I think this should be okay (especially since `zlib-ng` is under the `Zlib` licence itself anyway). Let me know if that’s wrong to do.

thoughtpolice (commented):
LGTM. I agree dropping the C code is worthwhile, especially since zlib-ng actually has an unusually complex build (plus, I don't have to go write patches to fix that.) Let's get this in as soon as the point release is out.

emilazy:
It’d be nice if someone could run decompression‐heavy and compression‐heavy `hyperfine` benchmarks against the base commit of this PR with the `packaging` feature enabled. Maybe a revset that looks at diffs and a rebase respectively? But those might spend too much CPU on diff computation, I dunno. It’d just be good to get some data given the ambiguous results I got. (I think we could eat a small performance hit to unify the configurations, avoid the build hassle, and buy a little extra safety, but maybe @yuja has thoughts since he added the `zlib-ng` flag.)

steadmon:
Ahhh, I didn't see the note about enabling `packaging` in time :(. I ran some benchmarks of `jj log` and `jj log -p` with this change rebased onto main. Not sure if it's useful or not. If no one else gets around to it before Monday, I'll try these again with the right features.

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |                         
|:---|---:|---:|---:|---:|                                                       
| `~/tmp/jj-main log -r 'root()..'  --no-pager` | 390.6 ± 25.2 | 343.7 | 429.9 | 1.00 |
| `~/tmp/jj-zlib-rs log -r 'root()..' --no-pager` | 390.8 ± 18.0 | 364.0 | 408.7 | 1.00 ± 0.08 |



| Command | Mean [s] | Min [s] | Max [s] | Relative |                            
|:---|---:|---:|---:|---:|                                                       
| `~/tmp/jj-main log -r 'root()..' -p --no-pager` | 99.950 ± 3.646 | 93.615 | 105.394 | 1.14 ± 0.04 |
| `~/tmp/jj-zlib-rs log -r 'root()..' -p --no-pager` | 87.392 ± 1.183 | 86.218 | 89.820 | 1.00 |

yuja:
In f58145b4, I ran `jj log --ignore-working-copy PATH` query (in non-merge-heavy repo like jj.) It will decompress Git commit and tree objects, but won't compute content diffs. Using deeper `PATH` might be better.

emilazy:
Sounds reasonable. Perhaps something like `jj restore --into ‹old commit› --restore-descendants --ignore-immutable` is a good test for compression, since it has to write a bunch of commit objects but doesn’t do much diff work? (Though I guess snapshotting a large blob would also work, but is probably less relevant for day‐to‐day performance.)

emilazy:
Rebased. Not quite sure how to changelog this one; it’s not quite a highlight from a user perspective, and only a breaking change from a packager perspective (though we could avoid that by keeping around a `packaging` feature that does nothing, and perhaps that would be a good idea if we expect further things that are sufficient hassle that we don’t want a regular build to involve them, though I think ideally we’d avoid that and as a packager I generally prefer to get a notice about obsolete features and dependencies upfront rather than having an unused CMake reference linger indefinitely).

yuja:
> Perhaps something like `jj restore --into ‹old commit› --restore-descendants --ignore-immutable` is a good test for compression, since it has to write a bunch of commit objects but doesn’t do much diff work?

Could be, but I haven't checked whether zlib compression/decompression was measurable in that workload. To stress test the write path, `jj duplicate` might also be useful.

> Not quite sure how to changelog this one; it’s not quite a highlight from a user perspective, and only a breaking change from a packager perspective

I wouldn't add changelog entry for this reason.

emilazy:
I think documenting removed dependencies is important for packagers, and it does break `--features packaging`. So maybe it should just go in the breaking change section?

yuja:
> I think documenting removed dependencies is important for packagers, and it does break `--features packaging`. So maybe it should just go in the breaking change section?

or add "Packaging changes" section? We could list new test dependencies (such as `gpgsm`) there, which end users shouldn't care.

emilazy:
Sounds reasonable to me; done.

arxanas (commented):
suggestion: Should we keep the `packaging` feature, even if it's currently empty? We've often added and removed features to `packaging`. It might be convenient to reduce churn slightly for certain packagers.

arxanas:
@emilazy @steadmon btw I'm curious how many iterations you tested with? (I see stats posted by @steadmon, but not the iteration count).

- I recall that `hyperfine` doesn't adaptively select an iteration count (beyond maybe "total elapsed wall-clock time").
- I've previously wrapped it in a script that would wait for the variance to converge somewhat before stopping.
- I recall the `hyperfine` default might be like n = 10; I've been recommending that my coworkers use like n = 20 or n = 30 by default, as it's been much more stable in my experience.

thoughtpolice:
I missed that before but, yes, the `packaging` feature was actually empty before the usage of the gix zlib-ng path; the goal was to we could just leave it empty and add or remove things at will as time goes on to reduce impact. Of course it only sort of half works that way; if we added a dependency then a maintainer with a package using `--features packaging` would still have to update it to include the new dep anyway.

So we could keep it or remove it (and add a changelog entry), IMO. I don't have super strong feelings either way but there is precedent for just leaving it.

emilazy:
> @emilazy @steadmon btw I'm curious how many iterations you tested with? (I see stats posted by @steadmon, but not the iteration count).
> 
> * I recall that `hyperfine` doesn't adaptively select an iteration count (beyond maybe "total elapsed wall-clock time").
> * I've previously wrapped it in a script that would wait for the variance to converge somewhat before stopping.
> * I recall the `hyperfine` default might be like n = 10; I've been recommending that my coworkers use like n = 20 or n = 30 by default, as it's been much more stable in my experience.

I think I just used the default. My experience is that any benchmark I’m patient enough to wait for gives too‐noisy results on this laptop, usually with a bunch of outliers.

> suggestion: Should we keep the `packaging` feature, even if it's currently empty? We've often added and removed features to `packaging`. It might be convenient to reduce churn slightly for certain packagers.

> I missed that before but, yes, the `packaging` feature was actually empty before the usage of the gix zlib-ng path; the goal was to we could just leave it empty and add or remove things at will as time goes on to reduce impact. Of course it only sort of half works that way; if we added a dependency then a maintainer with a package using `--features packaging` would still have to update it to include the new dep anyway.
> 
> So we could keep it or remove it (and add a changelog entry), IMO. I don't have super strong feelings either way but there is precedent for just leaving it.

My feelings:

* Yes, it would be annoying to have to re‐add the `packaging` feature later, and it’s a breaking change to drop it.

* Dropping the feature lets packagers get a message upfront that lets them know something needs adjusting; as a packager, I’d prefer this to accidentally leaving an unnecessary CMake dependency around indefinitely.

* I think ideally we wouldn’t have the `packaging` feature. It’s good for casual builds to produce the same results as release packages, good for both casual builds and release packages to not require bothersome build environment setup, and having only one main configuration gives us more coverage without needing to test both combinations in CI.

* It’s also easy to forget to set – the Nixpkgs package never set the `packaging` feature and has been using `miniz_oxide` this whole time!

* So in that sense, a little friction around reintroducing it doesn’t seem like such a bad thing – it’s like the dragon, in that it gives us reason to pause and wonder whether we can avoid introducing another build configuration that requires annoying dependencies. Ideally, native dependencies we add in future will have minimal build requirements, and work with both system libraries and vendored source. And of course, all else being equal, Rust dependencies often have other benefits like increased safety.

However, I don’t feel too strongly, so I’m happy to change this if people prefer.

thoughtpolice:
Yeah, sounds good to me then to just drop it. Now that 0.28.2 is out, let's go ahead and try to merge this.

BTW, I do agree that we really do want casual builds to be the same as our release builds; this is a huge headache on Windows for example (because it doesn't use OpenSSL by default and thus does not support Ed25519 hostkeys, among other features; another `git2` issue.) So in general removing these kinds of sharper edges are a good idea anyway.

emilazy:
Right. The `vendored-openssl` feature will be going away with `git2` anyway, so the CI builds would all have the same configuration, which I think is good. However I’ll give this a little bit in case there is a strong differing opinion, and also because it would be really nice to get some benchmark numbers.

Rebased for conflicts, and also I realized that the `zlib` dependency is probably for `libgit2` so I tweaked the flake and changelog warning.

CHANGELOG.md
yuja, line 1: nit: you'll need to drop the first empty commit "changelog: add missing release highlights section"

yuja (approved):
LG, thanks.

thoughtpolice (approved):

PR #6243, opened 2025-04-04
docs: standardize on .method() syntax

strega-nil:
In docs/templates.md, the Commit type and Operation type looked like:

`method() -> ReturnType`

rather than everything else, which looks like:

`.method() -> ReturnType`

This commit changes the Commit and Operation types to look more like the rest of the documentation.

# Checklist

If applicable:

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -108,41 +108,41 @@ No methods are defined. Can be constructed with `false` or `true` literal.
 
 This type cannot be printed. The following methods are defined.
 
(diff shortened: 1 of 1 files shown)


yuja (approved):
Thanks.

> I think both `.<method>` and `<method>` are actually correct, since each language has to have a unique `self` in scope for backcompat.

It's kinda copy paste error when I moved keyword docs to types. I added `()`, but forgot to insert `.`.

PR #6244, opened 2025-04-04
github: ward off future dependency bloat via dragon 

thoughtpolice:
After some discussion on Discord yesterday, Emily floated this idea to have a check that fails if `Cargo.lock` has too many dependencies, where "too many" means "more than a random number I made up and sounds good."

This implements that, as a non-required check, and to do so it invokes the power of an extremely evil and annoying Dragon. We could also ask this Dragon to do other things too I suppose (pending future contract negotiations).

# Checklist

If applicable:

### DIFF
--- a/.github/scripts/dragon-bureaucrat
+++ b/.github/scripts/dragon-bureaucrat
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+# This script invokes the forbidden power of an ancient evil in order to defend
+# the one thing we hold most dear: bureaucratic norms
+
+# Many thanks to Phabricator (and Evan) for the vintage ASCII art (Apache 2.0)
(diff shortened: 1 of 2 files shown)


PhilipMetzger (approved):
Thanks for the dragon.

emilazy:
I genuinely like the idea. I’m sure we will end up just bumping the number, but it adds a little friction to think about it and an incentive to trim stuff off elsewhere, and it seems like a nicer kind of incentive than keeping CI slower by avoiding caching.

martinvonz (approved):
I'm not sure it will help much but we can try

thoughtpolice:
I also think of it as a kind of "budgeting" thing. We have a budget we can spend on things like dependencies, and when do we go from the black to the red? Maybe this can help us keep that in mind, but we'll see.

PR #6248, opened 2025-04-04
release: 0.28.1

arxanas:
I removed the empty sections from `CHANGELOG.md`. Let me know if we want to format patch-release changelogs differently.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,6 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
-### Security fixes
-
-* Fixed SHA-1 collision attacks not being detected.
-  ([GHSA-794x-2rpg-rfgr](https://github.com/jj-vcs/jj/security/advisories/GHSA-794x-2rpg-rfgr))
-
-### Release highlights
-
 ### Breaking changes
 
 ### Deprecations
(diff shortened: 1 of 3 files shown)


docs/releasing.md
```diff
@@ -1,5 +1,13 @@
 # How to do a release
 
+## Check dry-run release
+
+Before updating the crate versions, check the `cargo publish --dry-run` builds:
+
+```shell
+cargo publish --dry-run -p jj-lib-proc-macros && cargo publish --dry-run -p jj-lib && cargo publish --dry-run -p jj-cli
```
martinvonz, line 8: Will the latter two not always fail since the latest `jj-lib-proc-macros` has not been uploaded?

martinvonz (approved):

docs/releasing.md
thoughtpolice: Yeah, I'm not sure what to do about this. We honestly need to smoke test it locally somehow, but I'm not sure OTTOMH what the options are in the Rust ecosystem for that. TBH, yanking and publishing a quickfix release seems fine to me generally speaking...

CHANGELOG.md
```diff
@@ -8,20 +8,30 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
```
scott2000, line 9: I think the `[Unreleased]` section is missing its changelog headers.

docs/releasing.md
arxanas: > Will the latter two not always fail since the latest `jj-lib-proc-macros` has not been uploaded?

@martinvonz No, because the test is run _before_ updating the version numbers, so each `cargo publish` successfully finds the the current versions of their dependencies. This is the error message that it produces on the commit immediately before the version number updates in v0.28.0:

```text
error: all dependencies must have a version specified when publishing.
dependency `testutils` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

That should be enough to catch this kind of configuration error. I don't know if the dry-run `cargo publish` uses the local `path` versions of the crates for actually building, though (which could be a problem).

I'll remove that commit from this PR for now.

PR #6249, opened 2025-04-04
github: show counts and allow offline counting for dependency-counting check

ilyagr:
Follow-up to b1bb5e1

This creates a `.github/scripts/count-cargo-lock-packages` script to count packages with our methodology that one can run outside CI.

Cc @thoughtpolice 

-----

I initially thought of making the whole check a script, but I the dragon would be annoying if you were repeatedly invoking the script locally.


thoughtpolice (approved):
Yeah, makes sense to me. Thanks!

ilyagr:
Sure! I moved a part of the comment as well.

PR #6252, opened 2025-04-04
tests: make `git push --named` tests demo differences in subprocess and libgit2 logic

ilyagr:
Follows up on https://github.com/jj-vcs/jj/pull/5698 and https://github.com/jj-vcs/jj/pull/6236.

The behavior of `--named` in corner cases differs. This wasn't detected
in https://github.com/jj-vcs/jj/pull/5698 because of a bug in the test. This commit better illustrates
the more important case where the behavior is the same, and the less
important case where it differs.


yuja (approved):
Thanks.

Issue #3327, opened 2024-03-19
Cannot override "ui.diff.tool" to use builtin diff in repo config


scott2000:
## Description

I want to use Difftastic for most of my repos, so I have `ui.diff.tool = "difft"` in my user config file, but for a specific repo I want to use the builtin "color-words" diff, since Difftastic doesn't give good diffs for it. However, there seems to be no way to override it to use the builtin word diff on a repo level.

## Steps to Reproduce the Problem

1. `jj config set --user ui.diff.tool "difft"`
1. `jj config set --repo ui.diff.tool ":builtin"`
1. Run `jj diff` with some changes

## Expected Behavior

I expected it to use the builtin diff from `ui.diff.format` (e.g. color-words).

## Actual Behavior

I get the following error since it actually tries to run it as a program:

```
Error: Failed to generate diff
Caused by:
1: Error executing ':builtin' (run with --debug to see the exact invocation)
2: No such file or directory (os error 2)
```

A similar error occurs when set to `"null"` or `""`.

## Specifications

- Platform: macOS 12.7.1
- Version: jj 0.15.1

martinvonz:
Makes sense. An alternative would be to allow `ui.diff.tool = ":color-words"` or something like that. Then we can remove the separate `ui.diff.format` config.

tomafro:
So `ui.diff.tool` could be set to `:summary|:types|:git|:color-words|:stat|<tool>`, builtin formats starting with a `:`. Then deprecate `ui.diff.format`?  

I think that works, though technically an external tool could start with a `:`. If this is the way you want to go I'd be happy to take a crack at this.

martinvonz:
Oh, does this mean that one will also be able to do `jj diff --tool :stat` as an alternative to `jj diff --stat`? I suppose that should ideally work. `--tool` sounds a bit strange in that context. Is `--format` better, or does that instead sound too odd if it's an external tool? Whatever we call the flag, we should probably keep the current flags (`--git`, `-s/--summary`, etc.) as shorthands, or what do people think? We can possibly drop the shorthands from `jj log/obslog`.

yuja:
> one will also be able to do `jj diff --tool :stat` as an alternative to `jj diff --stat`?

Yes, but I think it can be implemented separately. `--stat --summary` is currently rejected at clap layer, but `--tool :stat --summary` wouldn't. We might need to clean up short/long format thingy.

jjt:
[...]

PR #6253, opened 2025-04-05
diff_util: Move diff renderer format length check to constructor

wrzian:
This is a nice simplification that I found in another commit that was fully unrelated to my pull request, so I'm breaking it out into a new one.

In cli_util, I chose to put the `maybe_diff_renderer` function in the command environment instead of the helper because in `log.rs` there is no command helper already existing, and it would be non-trivial to build one first.

### Checklist

(I don't think any of these are applicable)

If applicable:

### DIFF
--- a/cli/src/commands/operation/diff.rs
+++ b/cli/src/commands/operation/diff.rs
@@ -121,13 +121,8 @@ pub fn cmd_op_diff(
     tx.repo_mut().merge_index(&from_repo);
     let merged_repo = tx.repo();
 
-    let diff_renderer = {
-        let formats = diff_formats_for_log(settings, &args.diff_format, args.patch)?;
-        let path_converter = workspace_env.path_converter();
-        let conflict_marker_style = workspace_env.conflict_marker_style();
-        (!formats.is_empty())
-            .then(|| DiffRenderer::new(merged_repo, path_converter, conflict_marker_style, formats))
-    };
+    let formats = diff_formats_for_log(settings, &args.diff_format, args.patch)?;
+    let diff_renderer = workspace_env.maybe_diff_renderer(merged_repo, formats);
     let id_prefix_context = workspace_env.new_id_prefix_context();
     let commit_summary_template = {
         let language = workspace_env.commit_template_language(merged_repo, &id_prefix_context);
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -287,23 +287,28 @@ pub enum DiffRenderError {
 /// Configuration and environment to render textual diff.
 pub struct DiffRenderer<'a> {
     repo: &'a dyn Repo,
+    formats: Vec<DiffFormat>,
     path_converter: &'a RepoPathUiConverter,
     conflict_marker_style: ConflictMarkerStyle,
-    formats: Vec<DiffFormat>,
 }
 
 impl<'a> DiffRenderer<'a> {
-    pub fn new(
+    /// Create a new textual diff renderer if the formats vector is non-empty.
+    pub fn new_if_non_empty(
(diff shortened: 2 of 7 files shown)


cli/src/diff_util.rs
```diff
@@ -287,23 +287,28 @@ pub enum DiffRenderError {
...
-    ) -> Self {
-        DiffRenderer {
-            repo,
-            path_converter,
-            conflict_marker_style,
-            formats,
+    ) -> Option<Self> {
+        if formats.is_empty() {
+            None
+        } else {
```
yuja, line 305: I would move this to `WorkspaceCommandEnvironment::maybe_diff_renderer()` or its callers.

cli/src/commands/operation/diff.rs
```diff
@@ -121,13 +121,8 @@ pub fn cmd_op_diff(
...
 
-    let diff_renderer = {
-        let formats = diff_formats_for_log(settings, &args.diff_format, args.patch)?;
-        let path_converter = workspace_env.path_converter();
-        let conflict_marker_style = workspace_env.conflict_marker_style();
-        (!formats.is_empty())
-            .then(|| DiffRenderer::new(merged_repo, path_converter, conflict_marker_style, formats))
-    };
+    let formats = diff_formats_for_log(settings, &args.diff_format, args.patch)?;
+    let diff_renderer = workspace_env.maybe_diff_renderer(merged_repo, formats);
```
yuja, line 125: Perhaps, it's simpler to handle `.is_empty()` here instead of adding `maybe_*()` and wrapping the result?
```rust
(!formats.is_empty()).then(|| workspace_env.diff_renderer(merged_repo, formats));
```

yuja (commented):
Thanks.

cli/src/diff_util.rs
wrzian: I like that it's part of the type's constructor. I think this should stay with the type, because it really is nonsensical to have a diff renderer that outputs nothing.

cli/src/commands/operation/diff.rs
wrzian: `diff_renderer` only takes a single format, not a vector of formats.
You should try rewriting this if you want to see the pitfalls. I don't think a simpler solution really exists.

cli/src/diff_util.rs
yuja: It's not wrong to construct a diff function that does nothing. I think this could be a contract at cli_util layer, but as I said, I assumed it would be easier to test the condition by caller.

FWIW, we'll need to reorganize the current diff format configuration as "short" and "long" formats pair . (see `DiffFormatArgs` and https://github.com/jj-vcs/jj/issues/3327) `DiffRenderer` wouldn't have to know this distinction, but that might change how we construct the renderer object.

wrzian: I don't care how it changes later. This is a more sensible interface currently. Many types return options in their constructor when they would be constructed and do something unexpected (like print a diff with nothing in it). This change centralizes this code and makes other interfaces more pleasant

cli/src/commands/operation/diff.rs
wrzian: I'm not doing that. This PR is a *simplification*. Your suggestion just makes the code more complex and re-introduces the exact confusing bit of code that I've been trying to remove.

yuja:
Apparently, you have different taste than me regarding simple+correct code structure. Maybe better to request someone else's review?

https://github.com/jj-vcs/jj/pull/6253#discussion_r2029795980
https://github.com/jj-vcs/jj/pull/6253#discussion_r2029795003

cc @martinvonz

PR #6232, opened 2025-04-03
tests: don’t use `git2` in `testutils`

emilazy:
Potential fix for the release issue. ~~If we don’t like the behaviour change here, we could set `JJ_DEBUG_HERMETIC_GIT` or something in `testutils::hermetic_git()` and condition on it.~~

# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #6258, opened 2025-04-05
github: add checks for recent issues

emilazy:
# Checklist

If applicable:


thoughtpolice:
Just to be safe, does this actually blow up if you e.g. revert https://github.com/jj-vcs/jj/pull/6232? Might be worth a test if not (if you've run it yourself then don't mind me.)

emilazy:
After a little iteration, confirmed that:

* Reverting the PR that fixed publication [breaks the publication test](https://github.com/jj-vcs/jj/actions/runs/14280313842/job/40029070548?pr=6258).

* Adding the commmit that would have broken installation [fixes the publication test](https://github.com/jj-vcs/jj/actions/runs/14280360644/job/40029178646) but [breaks the installation test](https://github.com/jj-vcs/jj/actions/runs/14280360644/job/40029178638).

* *But,* making `jj-cli` set a feature on `jj-lib` that isn’t present in the released version [broke the publication test](https://github.com/jj-vcs/jj/actions/runs/14280444977/job/40029365096?pr=6258). So I [fixed that]() by using the nightly `-Z package-workspace` feature, which seems to be [on the path to stabilization]() and which we should probably use for publication in future once it is. (This still leaves the possibility of publication issues with stable Rust that nightly doesn’t catch, but eh.)

* The check still [detects the 0.28.0 problem](https://github.com/jj-vcs/jj/actions/runs/14280738849/job/40029974999) after that change.

* The checks pass on `main`.

* The installation test was recompiling everything because I forgot to use `--profile dev` and `LIBGIT2_NO_VENDOR`, so I fixed that, but it still recompiles `gix-utils` without the `bstr` feature and I have no idea why and I hate debugging Cargo dependency tree configuration things.

thoughtpolice (approved):
SGTM, thanks!

emilazy:
Oh I don't think we should merge this until the install thing is fixed. It adds like a minute+ to CI.

.github/workflows/ci.yml
```diff
@@ -96,6 +96,19 @@ jobs:
         RUST_BACKTRACE: 1
         CARGO_TERM_COLOR: always
         LIBGIT2_NO_VENDOR: ${{ matrix.LIBGIT2_NO_VENDOR || '0' }}
+    # Try to catch packaging issues that will prevent installation.
+    - name: Install
+      run: >-
```
arxanas, line 101: I'm assuming `>-` does the right thing, but I can't look up the documentation easily (thanks YAML).

.github/workflows/ci.yml
```diff
@@ -172,6 +185,28 @@ jobs:
...
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@56f84321dbccf38fb67ce29ab63e4754056677e0
+        with:
+          toolchain: nightly
+          components: cargo
+      - name: Check `cargo publish`
+        run: >-
+          cargo +nightly publish
```
arxanas, line 202: comment (non-blocking): It occurs to me that this might be slow if there wasn't a previous `cargo +nightly build` on the same machine. I did not check the rest of the workflow definition to see if it was already built with `nightly`, and I trust that you will be keenly aware of any CI time regressions 🙂.

arxanas (approved):
Sounds like this is blocked a little, but I will express my support nonetheless.

.github/workflows/ci.yml
PhilipMetzger: see https://stackoverflow.com/a/65808412 and https://yaml-multiline.info/

.github/workflows/ci.yml
emilazy: It doesn’t do the actual build, just the publication checks. But I might be misunderstanding what you mean by “same machine” here?

PR #6259, opened 2025-04-05
cli: fix warning about pushing deleted bookmark

senekor:
# Checklist

If applicable:
- [ ] ~I have updated `CHANGELOG.md`~

### DIFF
--- a/cli/src/commands/bookmark/list.rs
+++ b/cli/src/commands/bookmark/list.rs
@@ -282,8 +282,9 @@ pub fn cmd_bookmark_list(
             Some(true) => {
                 writeln!(
                     ui.hint_default(),
-                    "Bookmarks marked as deleted will be *deleted permanently* on the remote on \
-                     the next `jj git push`. Use `jj bookmark forget` to prevent this."
+                    "Bookmarks marked as deleted can be *deleted permanently* on the remote by \
+                     running `jj git push --deleted`. Use `jj bookmark forget` if you don't want \
+                     that."
                 )?;
             }
             Some(false) => {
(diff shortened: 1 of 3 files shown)


cli/src/commands/bookmark/list.rs
```diff
@@ -268,32 +268,22 @@ pub fn cmd_bookmark_list(
...
                 })
-            })
-            .max();
-        match deleted_tracking {
-            Some(true) => {
-                writeln!(
-                    ui.hint_default(),
-                    "Bookmarks marked as deleted will be *deleted permanently* on the remote on \
-                     the next `jj git push`. Use `jj bookmark forget` to prevent this."
-                )?;
```
yuja, line 287: iirc, this is the hint for how/when scheduled deletion takes effect (and the deleted bookmarks disappear from the list.)

Maybe we'll need to suggest `jj git push --deleted`?

senekor: tracked down the initial commit where this message was added: d01ecc5c46043351ea0a3ab982bb2fd606d81ea3

yuja: > Why would we inform them about how to push deleted bookmarks in the bookmark list? 

because the user might not know why the deleted bookmark is still listed. As far as I understand it, it is the hint for how to "delete" deleted bookmarks from the list.

senekor: Ok that does make sense. I put the hint back, just slightly changed so it makes sense with the current behavior.

yuja (approved):
Thanks.

PR #6262, opened 2025-04-05
formatter: indent multiline strings with heading

senekor:
I think this is an improvement on its own.

But there are a lot of very long single-line strings that should probably be split up. I'm guessing this is the case because multiline strings weren't handled nicely so far, so a long single line was considered the lesser of two evils.

After this change, I'll probably go over some of the other strings and split them on multiple lines.

# Checklist

If applicable:
- [ ] ~I have updated `CHANGELOG.md`~

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2796,11 +2796,11 @@ pub fn print_snapshot_stats(
             ui.hint_default(),
             r"
             This is to prevent large files from being added by accident. You can fix this by:
-              - Adding the file to `.gitignore`
-              - Run `jj config set --repo snapshot.max-new-file-size {size}`
-                This will increase the maximum file size allowed for new files, in this repository only.
-              - Run `jj --config snapshot.max-new-file-size={size} st`
-                This will increase the maximum file size allowed for new files, for this command only.
+            - Adding the file to `.gitignore`
+            - Run `jj config set --repo snapshot.max-new-file-size {size}`
+              This will increase the maximum file size allowed for new files, in this repository only.
+            - Run `jj --config snapshot.max-new-file-size={size} st`
+              This will increase the maximum file size allowed for new files, for this command only.
             "
         )?;
     }
--- a/cli/src/formatter.rs
+++ b/cli/src/formatter.rs
@@ -132,7 +132,17 @@ where
             if let Some(heading) = self.heading.take() {
                 write!(formatter.labeled("heading"), "{heading}")?;
             }
-            formatter.write_fmt(args)
+            let msg = args.to_string();
+            if msg.matches('\n').nth(1).is_some() {
+                // message is multi-line, print it in an indented block
+                writeln!(formatter)?;
+                for line in msg.lines() {
+                    writeln!(formatter, "  {line}")?;
+                }
+                Ok(())
+            } else {
+                formatter.write_fmt(args)
+            }
         })
     }
 }
--- a/cli/src/commands/file/track.rs
+++ b/cli/src/commands/file/track.rs
@@ -109,11 +109,11 @@ pub fn print_track_snapshot_stats(
             ui.hint_default(),
             r"
             This is to prevent large files from being added by accident. You can fix this by:
-              - Adding the file to `.gitignore`
-              - Run `jj config set --repo snapshot.max-new-file-size {size}`
-                This will increase the maximum file size allowed for new files, in this repository only.
-              - Run `jj --config snapshot.max-new-file-size={size} file track {large_files_list}`
-                This will increase the maximum file size allowed for new files, for this command only.
+            - Adding the file to `.gitignore`
+            - Run `jj config set --repo snapshot.max-new-file-size {size}`
+              This will increase the maximum file size allowed for new files, in this repository only.
+            - Run `jj --config snapshot.max-new-file-size={size} file track {large_files_list}`
+              This will increase the maximum file size allowed for new files, for this command only.
             "
         )?;
     }
--- a/cli/tests/test_config_command.rs
+++ b/cli/tests/test_config_command.rs
@@ -1440,9 +1440,10 @@ fn test_config_author_change_warning() {
     let output = work_dir.run_jj(["config", "set", "--repo", "user.email", "'Foo'"]);
     insta::assert_snapshot!(output, @r#"
     ------- stderr -------
-    Warning: This setting will only impact future commits.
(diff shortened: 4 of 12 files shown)


cli/src/cli_util.rs
```diff
@@ -2796,11 +2796,11 @@ pub fn print_snapshot_stats(
...
-              - Adding the file to `.gitignore`
-              - Run `jj config set --repo snapshot.max-new-file-size {size}`
-                This will increase the maximum file size allowed for new files, in this repository only.
-              - Run `jj --config snapshot.max-new-file-size={size} st`
-                This will increase the maximum file size allowed for new files, for this command only.
+            - Adding the file to `.gitignore`
+            - Run `jj config set --repo snapshot.max-new-file-size {size}`
+              This will increase the maximum file size allowed for new files, in this repository only.
+            - Run `jj --config snapshot.max-new-file-size={size} st`
+              This will increase the maximum file size allowed for new files, for this command only.
```
emilazy, line 2803: The extra indentation for the list here is intentional, isn’t it?

senekor: Yeah, I guess that indentation was added to achieve the same thing that now happens for all log messages anyway.

before:
```
Hint: This is to prevent large files
  - Adding the file to `.gitignore`
  - Run `jj config set
```

after, with manual indentation kept:
```
Hint:
  This is to prevent large files
    - Adding the file to `.gitignore`
    - Run `jj config set
```

after, with manual indentation removed:
```
Hint:
  This is to prevent large files
  - Adding the file to `.gitignore`
  - Run `jj config set
```

The latter seems nicer to me but I'll revert if the former is preferred.

emilazy: I guess I'm neutral. In YAML I find unindented lists annoying, but in Markdown they're fine.

senekor:
One possibility for reducing the reduntant vertical space could be to rename the headings so they align nicely:

```
Fail: Something went wrong.
      Really, really wrong.
Warn: Watch out, not everything is jj-undoable
      We can't snapshot the state of github.com, sadly.
Hint: Jujutsus UI is mostly intuitive.
      But here's another friendly tip just because.
```

martinvonz:
> One possibility for reducing the reduntant vertical space could be to rename the headings so they align nicely:

FWIW, I like vertical-space-wasting format better :)

cli/src/formatter.rs
```diff
@@ -132,7 +132,17 @@ where
...
-            formatter.write_fmt(args)
+            let msg = args.to_string();
+            if msg.matches('\n').nth(1).is_some() {
+                // message is multi-line, print it in an indented block
+                writeln!(formatter)?;
+                for line in msg.lines() {
+                    writeln!(formatter, "  {line}")?;
+                }
+                Ok(())
+            } else {
```
yuja, line 143: Since `.write_fmt()` may be called more than once, we'll need to keep track of whether `'\n'` was printed or not. It would be easier to do indentation by caller.

(fwiw, I don't know if `writeln!()` macro is guaranteed to be expanded to a single `.write_fmt()` call, but apparently the current implementation is.)

yuja (commented):
I prefer the original dense format, but I don't have strong feeling.

cli/src/formatter.rs
senekor: Hm. If `write_fmt` is called multiple times for a single log message, and we have to track whether `\n` was already printed or not, then inserting a first line break is not possible.

So this format would _not_ work:
```
Warning:
  foo
  bar
```

but this one would:
```
Warn: foo
      bar
```

If it's better to not do this centrally, we can drop this PR. I can still go through the log messages manually and break some of the long ones onto multiple lines, using the `indoc` macros to ensure the indentation is appropriate.

yuja: Perhaps, the caller can opt in to the multi-line format by `.multiline()` or `.indented()`? I don't know if that's worth the effort, though.

BTW, indented text would look horrible if it's wrapped by terminal. We'll need to keep lines short.

martinvonz (approved):
Sounds like people like the new format.

cli/src/formatter.rs
senekor: For me, many commands use the pager if the lines are too long. But not all of them, and it could be environment dependent, not sure. I agree we need to make the lines short if we're going to add the indentation.

My plan is to go through a bunch of the really long single-line log messages and hard-wrap them to 80 characters or so. I think that would be an improvement for the majority of users.

senekor:
Another idea: What if the hard-wrapping was done at runtime with information from the terminal about its width? Then you always use the space efficiently and there is never weird overflow of indented lines. Bit more code, but maybe worth it.

senekor:
> > What if the hard-wrapping was done at runtime with information from the terminal about its width?
> 
> Perhaps, that would be hard to read in wide terminal?

We could set a maximum width of ~120 for the dynamic hard-wrapping.

> Do we have many examples that need hard-wrapping? I think many warning/hint messages can naturally be split to fit to ~100 columns. For example,
> 
> ```
> Hint:
>   Git doesn't allow a branch name that looks like a parent directory of
>   another (e.g. `foo` and `foo/bar`). Try to rename the bookmarks that failed to
>   export or their "parent" bookmarks."
> ```
> 
> this could be
> 
> ```
> Hint: Git doesn't allow a branch name that looks like a parent directory of another (e.g. `foo` and `foo/bar`).
> Hint: Try to rename the bookmarks that failed to export or their "parent" bookmarks.
> ```

So you'd split it in two separate warnings at the call site? I think this looks slightly confusing, as if the two hints were unrelated.

I'm becoming less sure about an automated solution to this. I kinda tend toward closing it and just doing a manual pass over the log messages, indenting or not as appropriate on a case-by-case basis.

yuja:
> > this could be
> > ```
> > Hint: Git doesn't allow a branch name that looks like a parent directory of another (e.g. `foo` and `foo/bar`).
> > Hint: Try to rename the bookmarks that failed to export or their "parent" bookmarks.
> > ```
> 
> So you'd split it in two separate warnings at the call site? I think this looks slightly confusing, as if the two hints were unrelated.

Yes, or two-line text with one "Hint: " heading. I feel it's easier to follow a sequence of short one-line messages.

> I'm becoming less sure about an automated solution to this. I kinda tend toward closing it and just doing a manual pass over the log messages, indenting or not as appropriate on a case-by-case basis.

+1

Maybe we could add helper function for indented output or hard-wrapping, but I don't think it's good idea to switch the output formats based on `.contains('\n')`.

PR #6263, opened 2025-04-05
commit: add trailer support

glehmann:
trailer support in the commit command

based on #6088 

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,9 +38,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * New `signed` revset function to filter for cryptographically signed commits.
 
-* `jj describe` adds the commit trailers, configured in the `commit_trailers`
-  template, to the commit description. Use cases include DCO Sign Off and
-  Gerrit Change Id.
+* `jj describe` and `jj commit` add the commit trailers, configured in the
+  `commit_trailers` template, to the commit description. Use cases include DCO
+  Sign Off and Gerrit Change Id.
 
 * Added `duplicate_description` template, which allows [customizing the descriptions
   of the commits `jj duplicate` creates](docs/config.md#duplicate-commit-description).
--- a/cli/src/commands/commit.rs
+++ b/cli/src/commands/commit.rs
@@ -22,6 +22,7 @@ use crate::cli_util::CommandHelper;
 use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::complete;
+use crate::description_util::add_trailers;
 use crate::description_util::description_template;
 use crate::description_util::edit_description;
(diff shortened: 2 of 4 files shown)


cli/src/commands/commit.rs
```diff
@@ -131,14 +132,24 @@ new working-copy commit.
     }
 
     let description = if !args.message_paragraphs.is_empty() {
-        join_message_paragraphs(&args.message_paragraphs)
+        let mut description = join_message_paragraphs(&args.message_paragraphs);
+        if !description.is_empty() {
```
yuja, line 136: nit: can you add comment why `!.is_empty()`?

cli/src/commands/commit.rs
```diff
@@ -131,14 +132,24 @@ new working-copy commit.
...
+            commit_builder.set_description(description);
+            let temp_commit = commit_builder.write_hidden()?;
+            description = add_trailers(ui, &tx, &temp_commit)?;
```
yuja, line 139: nit: perhaps, we can apply the same optimization as `jj describe`.

```rust
if let Some(trailer_template) = parse_trailers_template(ui, &tx)? {
    let temp_commit = commit_builder.write_hidden()?;
    ...
}
```

(or do that in `add_trailers(ui, &tx, &mut commit_builder)`?)

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   template, to the commit description. Use cases include DCO Sign Off and
   Gerrit Change Id.
 
+* `jj commit` now uses the `commit_trailers` template.
```
yuja, line 45: nit: combine with the previous entry.

glehmann:
I've tried to keep the trailers PR parallel, but that was becoming complicated to deal with the combined changelog entry.
I hope you don't mind for the review!

cli/tests/test_commit_command.rs
```diff
@@ -446,6 +446,68 @@ fn test_commit_reset_author() {
...
+    let output = work_dir.run_jj(["log", "--no-graph", "-r@-", "-Tdescription"]);
+    insta::assert_snapshot!(output, @r"
+    second
+
+    Reviewed-by: foo@bar.org
+    [EOF]
+    ");
+
+    // the trailer is added in the editor
+    std::fs::write(&edit_script, "dump editor0").unwrap();
```
yuja, line 495: nit: the script can be empty (or snapshot the `editor0` content)

yuja (approved):
> I've tried to keep the trailers PR parallel, but that was becoming complicated to deal with the combined changelog entry.

Feel free to combine them into single PR. I don't think there will be many review rounds (unlike the core implementation PR.)

Thanks.

PR #6264, opened 2025-04-05
new: add trailer support

glehmann:
based on #6088 

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * New `signed` revset function to filter for cryptographically signed commits.
 
-* `jj describe` and `jj commit` add the commit trailers, configured in the
+* `jj describe`, `jj commit` and `jj new` add the commit trailers, configured in the
   `commit_trailers` template, to the commit description. Use cases include DCO
   Sign Off and Gerrit Change Id.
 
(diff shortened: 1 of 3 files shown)


cli/src/commands/new.rs
```diff
@@ -184,11 +185,18 @@ pub(crate) fn cmd_new(
...
         .repo_mut()
         .new_commit(parent_commit_ids, merged_tree.id())
-        .set_description(join_message_paragraphs(&args.message_paragraphs))
-        .write()?;
+        .detach();
+    let mut description = join_message_paragraphs(&args.message_paragraphs);
+    if !description.is_empty() {
+        commit_builder.set_description(description);
+        let temp_commit = commit_builder.write_hidden()?;
+        description = add_trailers(ui, &tx, &temp_commit)?;
```
winterqt, line 196: What's the benefit to using a hidden commit here (setting aside the fact that our current trailers API doesn't support just adding trailers to a string)?

I see you did the same for `jj describe`, and I don't see any comments about it, so figured I should ask.

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   template, to the commit description. Use cases include DCO Sign Off and
   Gerrit Change Id.
 
+* `jj new` now uses the `commit_trailers` template.
```
winterqt, line 45: Might be a good idea to coalesce these "`jj [command]` now supports trailers" lines into one changelog entry as they're merged.

cli/src/commands/new.rs
winterqt: Ah, right, since these are templated... sorry for the noise.

PR #6265, opened 2025-04-05
squash: add trailer support

glehmann:
based on #6088 

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/squash.rs
+++ b/cli/src/commands/squash.rs
@@ -33,6 +33,7 @@ use crate::command_error::user_error;
 use crate::command_error::user_error_with_hint;
 use crate::command_error::CommandError;
 use crate::complete;
+use crate::description_util::add_trailers;
 use crate::description_util::combine_messages_for_editing;
 use crate::description_util::description_template;
 use crate::description_util::edit_description;
@@ -176,15 +177,28 @@ pub(crate) fn cmd_squash(
     )? {
         let mut commit_builder = squashed.commit_builder.detach();
         let new_description = match description {
-            SquashedDescription::Exact(description) => description,
-            SquashedDescription::UseDestination => destination.description().to_owned(),
+            SquashedDescription::Exact(description) => {
+                commit_builder.set_description(description);
+                add_trailers(ui, &tx, &commit_builder)?
+            }
+            SquashedDescription::UseDestination => {
+                commit_builder.set_description(destination.description());
+                add_trailers(ui, &tx, &commit_builder)?
+            }
             SquashedDescription::Combine => {
                 let abandoned_commits = &squashed.abandoned_commits;
                 if let Some(description) = try_combine_messages(abandoned_commits, &destination) {
-                    description
+                    commit_builder.set_description(description);
+                    add_trailers(ui, &tx, &commit_builder)?
                 } else {
                     let intro = "Enter a description for the combined commit.";
-                    let combined = combine_messages_for_editing(abandoned_commits, &destination);
+                    let combined = combine_messages_for_editing(
+                        ui,
+                        &tx,
+                        abandoned_commits,
+                        &destination,
+                        &commit_builder,
+                    )?;
                     // It's weird that commit.description() contains "JJ: " lines, but works.
                     commit_builder.set_description(combined);
                     let temp_commit = commit_builder.write_hidden()?;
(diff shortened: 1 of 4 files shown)


cli/src/commands/squash.rs
```diff
@@ -176,7 +177,10 @@ pub(crate) fn cmd_squash(
     )? {
         let mut commit_builder = squashed.commit_builder.detach();
         let new_description = match description {
-            SquashedDescription::Exact(description) => description,
+            SquashedDescription::Exact(description) => {
+                commit_builder.set_description(description);
+                add_trailers(ui, &tx, &commit_builder)?
+            }
             SquashedDescription::UseDestination => destination.description().to_owned(),
             SquashedDescription::Combine => {
```
yuja, line 185: Did you intentionally make `::Combine` not generate trailers? Can you add comment why?

I can see it's unclear what we should do here, but the behavior seems inconsistent. Depending on the rationale, it might not make sense that `jj split`/`squash` generate new trailers.

glehmann: That was kind of intentional: I didn't see how to integrate the trailers nicely into the combined message.

Now that I rethink about it, we could add a last section in the combined messages with the trailers generated with the `commit_trailers` template that were not found in any other message. That way, the user could decide what to do with them.

Something like

~~~
JJ: Enter a description for the combined commit.
JJ: Description from the destination commit:
message  1

JJ: Description from source commit:
message 2

JJ: Description from source commit:
message 3

JJ: Trailers not found in the squashed commit descriptions:

Signed-off-by: Alice <alice@example.com>

JJ: Lines starting with "JJ:" (like this one) will be removed.
~~~

yuja: > Now that I rethink about it, we could add a last section in the combined messages with the trailers generated with the commit_trailers template that were not found in any other message.

Sounds reasonable. Can you leave a TODO comment (or just implement it if that's easy)?

cli/src/commands/squash.rs
```diff
@@ -176,15 +177,24 @@ pub(crate) fn cmd_squash(
...
                 let abandoned_commits = &squashed.abandoned_commits;
                 if let Some(description) = try_combine_messages(abandoned_commits, &destination) {
                     description
```
yuja, line 188: Perhaps, we should also add trailers to combined `description` (n = 1) if it's not empty?

cli/src/description_util.rs
```diff
@@ -318,15 +319,46 @@ pub fn try_combine_messages(sources: &[Commit], destination: &Commit) -> Option<
...
+            .collect();
+        let commit = commit_builder.write_hidden()?;
+        let mut output = Vec::new();
+        template
+            .format(&commit, &mut PlainTextFormatter::new(&mut output))
+            .expect("write() to vec backed formatter should never fail");
+        let trailer_lines = output
+            .into_string()
+            .map_err(|_| user_error("Trailers should be valid utf-8"))?;
+        let new_trailers: HashSet<_> = parse_trailers(&trailer_lines)?.into_iter().collect();
```
yuja, line 350: Avoid using `HashSet` for `new_trailers` to stabilize order. `.filter()` can be used instead.

cli/src/description_util.rs
```diff
@@ -318,15 +319,46 @@ pub fn try_combine_messages(sources: &[Commit], destination: &Commit) -> Option<
...
+        let trailers = new_trailers
+            .difference(&old_trailers)
+            .map(|trailer| format!("{}: {}", trailer.key, trailer.value))
```
yuja, line 353: nit: `format!("{}: {}\n")` and `.collect()` for consistency?

yuja (approved):
Thanks.

cli/src/commands/squash.rs
glehmann: I didn't do it to have the same behavior as when using `--use-destination-message`, but I'm not sure what the behavior should be.

yuja: Maybe insert trailers to `UseDestination`, too? It seems odd that combined description is updated depending on the number of non-empty texts.

cli/src/commands/squash.rs
```diff
@@ -176,15 +177,28 @@ pub(crate) fn cmd_squash(
...
+            SquashedDescription::UseDestination => {
+                commit_builder.set_description(destination.description());
+                add_trailers(ui, &tx, &commit_builder)?
+            }
             SquashedDescription::Combine => {
                 let abandoned_commits = &squashed.abandoned_commits;
                 if let Some(description) = try_combine_messages(abandoned_commits, &destination) {
-                    description
+                    commit_builder.set_description(description);
+                    add_trailers(ui, &tx, &commit_builder)?
```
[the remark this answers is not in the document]
glehmann: Yes, indeed. Sorry, I merged too fast.
I've created a new PR #6377

PR #6268, opened 2025-04-06
cargo: bump gix and its dependencies recursively

yuja:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* Fixed problem that old commits could be re-imported from Git.
+  https://github.com/GitoxideLabs/gitoxide/issues/1928
+
 ## [0.28.1] - 2025-04-04
 
 ### Security fixes
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.15",
  "once_cell",
  "version_check",
  "zerocopy",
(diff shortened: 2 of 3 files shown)


emilazy:
Maybe we should wait on https://github.com/GitoxideLabs/gitoxide/issues/1934?

(I guess we're going to get a 0.28.2?)

ilyagr:
I don't think we need to wait for the pr.

For a new release, I'm a bit confused. We could update our binaries, but I don't think `cargo publish` notices changes to Cargo.lock.

yuja:
> Optional: might make sense to mention [GitoxideLabs/gitoxide#1928](https://github.com/GitoxideLabs/gitoxide/issues/1928) in the description and/or the change log.

Done.

> Maybe we should wait on [GitoxideLabs/gitoxide#1934](https://github.com/GitoxideLabs/gitoxide/issues/1934)?

I'm not sure if it's new problem which we'll get immediate fix.

emilazy:
> **Update** I think this problem mainly affects `cargo install` without `--locked`. Cargo.lock _does_ exist on crates.io

We do recommend `--locked` in our docs, and I think without `--locked` it would pick up the newer deps automatically.

martinvonz:
> (I guess we're going to get a 0.28.2?)

What do you think, @yuja? How did you run into the bug? Do you think it's likely that other users will?

ilyagr:
The effect is that, in colocated repos, `jj git import` creates phantom bookmark conflicts that reappear on the next `jj git import` no matter what you do to fix them. Calling `git cs` or `jj util gc` should be a workaround.

PhilipMetzger:
I think it's irritating behavior and should be dealt with a point release. See @bnjmnt4n's [confusion in Discord yesterday][dc].

[dc]: https://discord.com/channels/968932220549103686/968932220549103689/1358108166692540468

ilyagr:
But again, I wonder whether we should encourage Byron to create a new patch version of `gix` we could put in our `Cargo.toml` before a patch release.

yuja:
I agree we'll need a point release, either from the current tip or v0.28.1 + new Cargo.lock. The problem could occur if there are bookmark names including slash, and it's difficult for users to diagnose what happened.

PR #6269, opened 2025-04-06
files: add merge() functions that return result as Merge<_>/Option<_>

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/files.rs
+++ b/lib/src/files.rs
@@ -187,18 +187,25 @@ where
     }
 }
 
+/// Merge result in either fully-resolved or conflicts form, akin to
+/// `Result<BString, Vec<Merge<BString>>>`.
 #[derive(PartialEq, Eq, Clone, Debug)]
 pub enum MergeResult {
+    /// Resolved content if inputs can be merged successfully.
     Resolved(BString),
+    /// List of partially-resolved hunks if some of them cannot be merged.
     Conflict(Vec<Merge<BString>>),
 }
 
-pub fn merge<T: AsRef<[u8]>>(slices: &Merge<T>) -> MergeResult {
+/// Splits `inputs` into hunks, resolves trivial merge conflicts for each.
+///
+/// Returns either fully-resolved content or list of partially-resolved hunks.
+pub fn merge<T: AsRef<[u8]>>(inputs: &Merge<T>) -> MergeResult {
     // TODO: Using the first remove as base (first in the inputs) is how it's
     // usually done for 3-way conflicts. Are there better heuristics when there are
     // more than 3 parts?
-    let num_diffs = slices.removes().len();
-    let diff_inputs = slices.removes().chain(slices.adds());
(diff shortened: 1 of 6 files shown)


lib/src/files.rs
```diff
@@ -201,12 +201,57 @@ pub enum MergeResult {
...
 pub fn merge<T: AsRef<[u8]>>(inputs: &Merge<T>) -> MergeResult {
+    merge_inner(inputs)
+}
+
+/// Splits `inputs` into hunks, resolves trivial merge conflicts for each, then
+/// concatenates the outcome back to single `Merge` object.
+///
+/// The returned merge object is either fully resolved or conflict having the
+/// same number of terms as the `inputs`.
+pub fn merge_back<T: AsRef<[u8]>>(inputs: &Merge<T>) -> Merge<BString> {
```
martinvonz, line 212: Maybe this one should be called just `merge()` and the one above should be `merge_partial()`? Or maybe this one should be `merge_fully()`?

martinvonz (approved):

lib/src/files.rs
yuja: That might work, thanks.

"partial" might be a bit misleading because both functions retain partially-resolved hunks, but it's correct that `files::merge()` is "partial" in a way that it doesn't concatenate unresolved hunks.

martinvonz: Yes, I mean in the sense that it's not either/or but may return a result where some hunks have been resolved and some have not. Another option might be `merge_hunks()` because it's the only version that returns hunks.

Btw, when will `merge_back()` (whatever we end up calling it) be used? If the merge fails due to conflicts, it will returned owned versions of the input slices, right? It's not clear to me when it's useful to have that, since the caller already had the borrowed versions of the data (and maybe they even had owned versions already).

yuja: > Another option might be `merge_hunks()` because it's the only version that returns hunks.

Yeah, `merge() -> Merge`, `try_merge() -> Option`, `merge_hunks() -> Result<Option, Merge>` might actually be good.

> Btw, when will `merge_back()` (whatever we end up calling it) be used? If the merge fails due to conflicts, it will returned owned versions of the input slices, right? It's not clear to me when it's useful to have that, since the caller already had the borrowed versions of the data (and maybe they even had owned versions already).

The returned `Merge<BString>` includes partial-resolved hunks. So diff function (and maybe annotate/absorb) won't have to re-run merge based on diff hunks (which might slightly differ depending on diff source/destination pairs?)

Word-level merging will use `try_merge()` or `merge_back()`, and the resulting hunks will be materialized. I haven't decided which will be better overall.

martinvonz: Makes sense.

yuja: Renamed to `merge_hunks()`, `merge()`, and `try_merge()`.

PR #6280, opened 2025-04-07
cargo: general `cargo update`

ilyagr:
I realized what nuggets can hide in minor version updates after out 0.28.2; perhaps we should do this occasionally.

There's also some risk of breakage, so it seems to also good to do it in the beginning of a prerelease cycle.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,12 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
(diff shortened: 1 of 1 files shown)


thoughtpolice (commented):
Yeah, I did this once or twice in the past. Maybe refreshing it along with `flake.lock` regularly is a good idea.

thoughtpolice (approved):

PR #6283, opened 2025-04-08
lib: remove usage of `winreg` crate

thoughtpolice:
This pulls in a whole library just for what is morally 3 function calls and can be done easily in the Windows API. Remove it.

This uses `unsafe`; the usage of `windows-rs` might be more preferrable since we already (and will likely always) depend on it, but it would probably remain unsafe...

# Checklist

If applicable:


thoughtpolice:
I don't know if we actually want this change but I was taking a first crack at thinning up the dependency tree a little and seeing what impacts the features have, etc.

martinvonz:
How much smaller does the binary become? How much faster does the build become?

emilazy:
I would much prefer `windows-rs` over this, but I am not sure it is worth it anyway. `winreg` does not itself have any dependencies that we would be able to remove by dropping it, and outsourcing the Windows domain knowledge and verification of the `unsafe` stuff to them feels worth it. Maybe the `unsafe` stuff would be sufficiently simple with `windows-rs` that I’d feel differently, but I lean towards this not being a fruit worth picking.

If we only use it for `check_symlink_support`, can’t we just do a feature test instead – try to create a symlink and see if it works? Looking at a registry entry that might not correlate perfectly with our ability to do symlinks seems worth avoiding anyway.

thoughtpolice:
(And to be clear I don't think we have to merge this, the alternative suggested by Emily might be just as good. But I think this is a good example of where I differ from a lot of people on this topic; this is the kind of stuff that I think we can and should evaluate and take some responsibility for and be just willing to decide: yes, it's a little extra work, but we don't need that.)

emilazy:
> As far as the compatibility check goes, my understanding is that it's more or less correct; Windows 10 has had symlinks-for-devmode on for at least a couple years at this point so they are basically synonymous. I suppose we could also try just creating them and handling the error too, but I haven't thought to deeply about it.

I think that the toggle determinates whether non‐administrators can create symlinks, though. I think administrators can do it regardless and there’s some [specific permission you can assign](https://portal.perforce.com/s/article/3472) even without developer mode. Whether we’d want to take advantage of running as an administrator in that way, I don’t know, but we’re at least not quite checking what the function name implies.

FWIW, I’d be more inclined to just bite the bullet on the dependency here if not for the `unsafe`, or if the crate was bigger, or pulled in other new dependencies. In this case the crate is less than 2,000 lines of Rust, some of which is `serde` stuff that we’re not pulling in, and the dependencies are `cfg-if` and `windows-sys`, neither of which I see leaving our graph any time soon. So it seems low‐cost as dependencies go to me.

But probably just doing an actual runtime check would be the best of both worlds? The ability to create them might also depend on the filesystem, not just the OS. Like the `samefile` checks we do to detect case‐insensitivity clashes on the specific FS we’re operating on, it seems like we’d be better off just always trying to create symlinks and then falling back if we can’t (with a cache, if it’s performance‐relevant to keep trying somehow).

PR #6288, opened 2025-04-08
cli: resolve: do not use .with_new_file_ids() to construct resolved file value

### LINKED ISSUE #6250: `jj resolve` panics when resolving conflicts involving deleted files
## Description

`jj resolve` panics whenever resolving a conflict involving deleted files using an external merge tool (or `:ours` or `:theirs`).

This was found by [saber on Discord][discord] for `:ours` and `:theirs`, but I tested it and the same issue occurs with external merge tools as well.

[discord]: https://discord.com/channels/968932220549103686/968932220549103689/1357751780133306408

1. Create a modify-delete conflict (`2-sided conflict including 1 deletion`).
1. Run `jj resolve` with an external merge tool (or with `:ours` or `:theirs`).
1. Resolve the conflict using the merge tool.

## Expected Behavior

The resolution of the conflict is saved in the working copy. Due to executable bit handling, it may not be possible to resolve the conflict fully, but at least it shouldn't panic. Alternatively, we could just disallow `jj resolve` when one of the sides is a deletion.

## Actual Behavior

A panic like this when the resolution doesn't delete the file (which is normally the case for external merge tools):

```
thread 'main' panicked at lib/src/merge.rs:626:21:
assertion failed: file_id.is_none()
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: jj_lib::merge::Merge<core::option::Option<T>>::with_new_file_ids::{{closure}}
   4: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::next
   5: <smallvec::SmallVec<A> as core::iter::traits::collect::Extend<<A as smallvec::Array>::Item>>::extend
   6: <smallvec::SmallVec<A> as core::iter::traits::collect::FromIterator<<A as smallvec::Array>::Item>>::from_iter
   7: core::iter::traits::iterator::Iterator::collect
   8: jj_lib::merge::Merge<core::option::Option<T>>::with_new_file_ids
   9: jj_cli::merge_tools::external::run_mergetool_external_single_file
  10: jj_cli::merge_tools::external::run_mergetool_external
  11: jj_cli::merge_tools::MergeEditor::edit_files
  12: jj_cli::commands::resolve::cmd_resolve
```

Otherwise, a panic like this when the resolution deletes the file (which may be the case for `:ours` or `:theirs`):

```
thread 'main' panicked at lib/src/merge.rs:621:46:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: jj_lib::merge::Merge<core::option::Option<T>>::with_new_file_ids::{{closure}}
   5: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::next
   6: <smallvec::SmallVec<A> as core::iter::traits::collect::Extend<<A as smallvec::Array>::Item>>::extend
   7: <smallvec::SmallVec<A> as core::iter::traits::collect::FromIterator<<A as smallvec::Array>::Item>>::from_iter
   8: core::iter::traits::iterator::Iterator::collect
   9: jj_lib::merge::Merge<core::option::Option<T>>::with_new_file_ids
  10: jj_cli::merge_tools::pick_conflict_side
  11: jj_cli::merge_tools::MergeEditor::edit_files
  12: jj_cli::commands::resolve::cmd_resolve
```

## Specifications

- Platform: NixOS
- Version: `jj 0.28.0-29f24ad2d8746e497abb701a99ba00b874f0cea7`

yuja:
Fixes #6250

# Checklist

If applicable:

### DIFF
--- a/cli/src/merge_tools/external.rs
+++ b/cli/src/merge_tools/external.rs
@@ -9,6 +9,7 @@ use std::sync::Arc;
 use bstr::BString;
 use itertools::Itertools as _;
 use jj_lib::backend::MergedTreeId;
+use jj_lib::backend::TreeValue;
 use jj_lib::conflicts;
 use jj_lib::conflicts::choose_materialized_conflict_marker_len;
 use jj_lib::conflicts::materialize_merge_result_to_bytes_with_marker_len;
@@ -308,13 +309,15 @@ fn run_mergetool_external_single_file(
             ExternalToolError::InvalidConflictMarkers { exit_status },
         ));
     }
-    // Update the file ids only, leaving the executable flags unchanged
-    let new_file_ids = if let Some(resolved) = new_file_ids.as_resolved() {
-        Merge::from_vec(vec![resolved.clone(); conflict.as_slice().len()])
-    } else {
-        new_file_ids
+
+    let new_tree_value = match new_file_ids.into_resolved() {
+        Ok(file_id) => {
+            let executable = file.executable.expect("should have been resolved");
+            Merge::resolved(file_id.map(|id| TreeValue::File { id, executable }))
+        }
+        // Update the file ids only, leaving the executable flags unchanged
+        Err(file_ids) => conflict.with_new_file_ids(&file_ids),
     };
-    let new_tree_value = conflict.with_new_file_ids(&new_file_ids);
     tree_builder.set_or_remove(repo_path.to_owned(), new_tree_value);
     Ok(())
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* Fixed crash on change-delete conflict resolution.
+  [#6250](https://github.com/jj-vcs/jj/issues/6250)
+
 ### Packaging changes
 
 * Jujutsu now uses
--- a/cli/src/command_error.rs
+++ b/cli/src/command_error.rs
@@ -439,7 +439,17 @@ impl From<ConflictResolveError> for CommandError {
         match err {
             ConflictResolveError::Backend(err) => err.into(),
             ConflictResolveError::Io(err) => err.into(),
-            _ => user_error_with_message("Failed to resolve conflicts", err),
+            _ => {
+                let hint = match &err {
+                    ConflictResolveError::ExecutableConflict { .. } => {
+                        Some("Use `jj file chmod` to update the executable bit.".to_owned())
+                    }
+                    _ => None,
+                };
+                let mut cmd_err = user_error_with_message("Failed to resolve conflicts", err);
+                cmd_err.extend_hints(hint);
+                cmd_err
+            }
         }
     }
 }
--- a/cli/src/commit_templater.rs
+++ b/cli/src/commit_templater.rs
@@ -29,6 +29,7 @@ use jj_lib::backend::ChangeId;
 use jj_lib::backend::CommitId;
 use jj_lib::backend::TreeValue;
 use jj_lib::commit::Commit;
+use jj_lib::conflicts;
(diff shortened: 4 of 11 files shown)


martinvonz (commented):
I think we should instead make it an error to try to resolve the content-level conflict before resolving the executable-bit-level conflict. I think we should do the same when merging copy records, i.e. we should not pick one side but instead make the user decide which copy id to keep.

martinvonz:
Yes. I think it's fine to use `.with_new_file_ids()` after checking that there's no conflict in the executable bit (or copy id later). Maybe we can update the API somehow to reduce the risk of bugs like this one but I haven't thought about it.

cli/src/merge_tools/external.rs
```diff
@@ -310,13 +311,19 @@ fn run_mergetool_external_single_file(
...
+
+    let new_tree_value = match new_file_ids.into_resolved() {
+        Ok(new_file_id) => Merge::normal(TreeValue::File {
+            id: new_file_id.unwrap(),
+            executable: conflict
+                .to_executable_merge()
+                .as_ref()
+                .and_then(Merge::resolve_trivial)
+                .copied()
+                .unwrap_or_default(),
```
martinvonz: I think instead of using "not executable" as default here, we should make it an error and tell the user to run `jj file chmod` to choose whether it should be executable.

> Since exec bit has only two states, it can always be trivially resolved.

If a file was added as executable in one commit and added as non-executable in another commit and then you rebase on of them onto the other, it would result in a conflict in the executable bit. This is the kind of conflict I think we should inform the user of and make them resolve with `jj file chmod`. Another example is modify/delete conflicts where the modification changed the executable bit. Sorry that I wasn't clear about this before.

Does that make sense?

yuja: > If a file was added as executable in one commit and added as non-executable in another commit and then you rebase on of them onto the other, it would result in a conflict in the executable bit. This is the kind of conflict I think we should inform the user of and make them resolve with `jj file chmod`. Another example is modify/delete conflicts where the modification changed the executable bit. Sorry that I wasn't clear about this before.

Do you mean we should change `.to_executable_merge()` to be tri-state? An absent file is interpreted as non-executable, so `[exec, absent, no-exec]` can be resolved to `exec`. I assumed the rationale here is that setting exec bit is an explicit operation, so `[(add file, set exec), (), (add file)]` can be merged as `(add file, set exec)`.

martinvonz: > Do you mean we should change `.to_executable_merge()` to be tri-state?

It already is:

```
match value.to_executable_merge().map(|merge| merge.resolve_trivial()) {
  None => // some sides are trees or symlinks
  Some(Some(true)) => // executable
  Some(Some(false)) => // not executable
  Some(None) => // conflict in executable bit
}
```

It's for the last case I think we should tell the user to run `jj file chmod` first. Or, even better, have `jj resolve` ask the user whether they want the file to be executable.

martinvonz: No, you're not misunderstanding. I was just blind. I really thought `to_executable_merge()` returned a `Option<Merge<Option<bool>>>` even after looking at it. I feel like it should return that type but I haven't checked what other consequences that would have.

yuja: Ok, I'll take a look then. Perhaps, `jj resolve` will check if executable can be merged first, and error out immediately.

yuja: Inserted patches to detect `[no-exec, exec, absent]` conflicts.

martinvonz (approved):
Thank you!

PR #6290, opened 2025-04-08
Remove deprecated backout command from documentation table

alper:
I just ran into this trying to figure out the equivalent to `git revert`. Turns out that backout is deprecated and revert does not have the same defaults.

### DIFF
--- a/docs/git-command-table.yml
+++ b/docs/git-command-table.yml
@@ -376,7 +376,7 @@
   Git command: >
     `git revert <revision>`
   Jujutsu command: >
-    `jj backout -r <revision>`
+    `jj revert -r <revision> -B @`
   Notes: ''
 
 - Use case: Show what revision and author last modified each line of a file


thoughtpolice:
Note that `backout` was only deprecated three weeks ago. We're probably going to leave it around for at least 3-5 releases before it at this rate, so this PR will be in limbo until then (but you can also close it & refile, leave it, whatever...)

emilazy:
It only replaces it in the docs. (Though I think people probably want `-B @` more than `-d @`.)

thoughtpolice:
Ah, that's what I get for drawing conclusions from only reading the PR title. :)

alper:
Thanks! Sorry for the confusion but I live by this table right now and it being up to date is going to help me at least.

@emilazy Didn't know what the best default would be but I changed it to your suggestion.

PhilipMetzger:
> It only replaces it in the docs. (Though I think people probably want `-B @` more than `-d @`.)

Isn't `-B @`equivalent to `-d @-` though? Also prefix the commit message with `docs:` per our commit style here, https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#commit-guidelines.

emilazy:
`-B @` rebases `@` on top of the revert.

alper:
Fixed up the description and squashed the commits.

bean5:
As a newer user, I was annoyed that documentation used a now-deprecating feature, so I agree with pushing this change. Unless there is a reason to use a deprecated feature that I do not know about.

martinvonz:
Sorry, I think this just fell through the cracks and now the commit conflicts with the main branch. Could you rebase and push it again, @alper?

alper:
@martinvonz Sure thing. Moved the change over.

PR #3466, opened 2024-04-06
config: also use XDG_CONFIG_HOME on MacOS

ckoehler:
This replaces the `dirs` crate with the more lightweight and flexible `etcetera` crate, which also lets us use XDG config on MacOS.

Fixes #3434.

~~I'm not entirely sure how do prevent the line break in the doc table between "Application" and "Support" 😅 ...~~

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [x] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [x] I have added tests to cover my changes

PR #6300, opened 2025-04-09
config: Use XDG-standard config location on macOS

### LINKED ISSUE #3434: FR: look in $XDG_CONFIG_HOME for jj config on OSX

OSX CLI users often have config files in $XDG_CONFIG_HOME, like Linux. `jj` uses the `dirs` crate which is very opinionated about not adding this (see https://github.com/dirs-dev/dirs-rs/pull/40)

In addition to the config location in `Application Support`, I'd like jj to check `$XDG_CONFIG_HOME/jj`, with priority given to the one in `Application Support`. There's [already some code to handle this](https://github.com/martinvonz/jj/blob/main/cli/src/config.rs#L32) under different circumstances, OSX just never gets the option to use another location.

We could use another crate than `dirs` to enable this, but I am not sure what's out there. Would love to hear people's thoughts.

Setting `JJ_CONFIG` works, but OSX CLI users may be surprised by it (I was).

**Additional context**

strega-nil:
Changing where the default config location is on macOS, switching to [etcetera][] from [dirs][].

Some points for discussion:

- I am treating the classic behaviour as a bug, not a feature; I do _not_ believe that it is correct in any way to place configuration in `~/Library/Application Support`, ever, and certainly not for CLI applications
- ~~Currently, if you run `jj config edit --user`, and you already have a config in `~/Library/Application Support`, you'll get a picker for both; this seems less than ideal? Perhaps advice would be good? Or we could ignore the normal file if the legacy file exists?~~
  - ~~This has been changed as of latest patch set – it may be desireable to pull that patch out into its own PR. Let me know.~~
  - ~~This bugfix has been pulled into its own PR: #6315~~
  - This was not a bug. I screwed up my own computer's configuration.

This PR takes a lot of inspiration from @ckoehler's #3466, but has been rewritten since the current codebase is a hell of a lot more amenable to these changes.

Fixes #3434 

[etcetera]: https://crates.io/crates/etcetera
[dirs]: https://crates.io/crates/dirs

# Checklist

If applicable:
- [x] ~~I have updated the config schema (cli/src/config-schema.json)~~
- [x] I have played around with my changes

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -163,26 +163,39 @@ impl AsMut<StackedConfig> for RawConfig {
 }
 
 #[derive(Clone, Debug)]
-enum ConfigPath {
-    /// Existing config file path.
-    Existing(PathBuf),
-    /// Could not find any config file, but a new file can be created at the
-    /// specified location.
-    New(PathBuf),
+enum ConfigPathState {
+    New,
+    Exists,
+}
+
+/// A ConfigPath can be in one of two states:
+///
+/// - exists(): a config file exists at the path
+/// - !exists(): a config file doesn't exist here, but a new file _can_ be
+///   created at this path
+#[derive(Clone, Debug)]
+struct ConfigPath {
+    path: PathBuf,
+    state: ConfigPathState,
 }
 
 impl ConfigPath {
     fn new(path: PathBuf) -> Self {
-        if path.exists() {
-            ConfigPath::Existing(path)
-        } else {
-            ConfigPath::New(path)
+        use ConfigPathState::*;
+        ConfigPath {
+            state: if path.exists() { Exists } else { New },
+            path,
         }
     }
 
     fn as_path(&self) -> &Path {
-        match self {
-            ConfigPath::Existing(path) | ConfigPath::New(path) => path,
+        &self.path
+    }
+
+    fn exists(&self) -> bool {
+        match self.state {
+            ConfigPathState::Exists => true,
+            ConfigPathState::New => false,
         }
     }
 }
@@ -216,6 +229,7 @@ impl UnresolvedConfigEnv {
                 .map(ConfigPath::new)
                 .collect();
         }
+
         let mut paths = vec![];
         let home_config_path = self.home_dir.map(|mut home_dir| {
             home_dir.push(".jjconfig.toml");
@@ -231,19 +245,25 @@ impl UnresolvedConfigEnv {
             config_dir.push("conf.d");
             ConfigPath::new(config_dir)
         });
-        use ConfigPath::*;
-        if let Some(path @ Existing(_)) = home_config_path {
-            paths.push(path);
-        } else if let (Some(path @ New(_)), None) = (home_config_path, &platform_config_path) {
-            paths.push(path);
+
+        if let Some(path) = home_config_path {
+            if path.exists() || platform_config_path.is_none() {
+                paths.push(path);
+            }
         }
+
         // This should be the default config created if there's
         // no user config and `jj config edit` is executed.
         if let Some(path) = platform_config_path {
             paths.push(path);
         }
-        if let Some(path @ Existing(_)) = platform_config_dir {
-            paths.push(path);
+
+        // theoretically this should be an `if let Some(...) = ... && ..., but that
+        // isn't stable
+        if let Some(path) = platform_config_dir {
+            if path.exists() {
+                paths.push(path);
+            }
         }
         paths
     }
@@ -284,16 +304,16 @@ impl ConfigEnv {
 
     /// Returns the paths to the user-specific config files or directories.
     pub fn user_config_paths(&self) -> impl Iterator<Item = &Path> {
-        self.user_config_paths.iter().map(|p| p.as_path())
+        self.user_config_paths.iter().map(ConfigPath::as_path)
     }
 
     /// Returns the paths to the existing user-specific config files or
     /// directories.
     pub fn existing_user_config_paths(&self) -> impl Iterator<Item = &Path> {
-        self.user_config_paths.iter().filter_map(|p| match p {
-            ConfigPath::Existing(path) => Some(path.as_path()),
-            _ => None,
-        })
+        self.user_config_paths
+            .iter()
+            .filter(|p| p.exists())
+            .map(ConfigPath::as_path)
     }
 
     /// Returns user configuration files for modification. Instantiates one if
@@ -349,13 +369,13 @@ impl ConfigEnv {
 
     /// Returns a path to the repo-specific config file.
     pub fn repo_config_path(&self) -> Option<&Path> {
-        self.repo_config_path.as_ref().map(ConfigPath::as_path)
+        self.repo_config_path.as_ref().map(|p| p.as_path())
     }
 
     /// Returns a path to the existing repo-specific config file.
     fn existing_repo_config_path(&self) -> Option<&Path> {
-        match &self.repo_config_path {
-            Some(ConfigPath::Existing(path)) => Some(path),
+        match self.repo_config_path {
+            Some(ref path) if path.exists() => Some(path.as_path()),
             _ => None,
         }
     }
@@ -1256,12 +1276,42 @@ mod tests {
     struct TestCase {
         files: &'static [&'static str],
         env: UnresolvedConfigEnv,
-        wants: &'static [Want],
+        wants: Vec<Want>,
     }
 
-    enum Want {
-        New(&'static str),
-        Existing(&'static str),
+    #[derive(Debug)]
+    enum WantState {
+        New,
+        Existing,
+    }
+    #[derive(Debug)]
+    struct Want {
+        path: &'static str,
+        state: WantState,
+    }
+
+    impl Want {
+        const fn new(path: &'static str) -> Want {
+            Want {
+                path,
+                state: WantState::New,
+            }
+        }
+
+        const fn existing(path: &'static str) -> Want {
+            Want {
+                path,
+                state: WantState::Existing,
+            }
+        }
+
+        fn rooted_path(&self, root: &Path) -> PathBuf {
+            root.join(self.path)
+        }
+
+        fn exists(&self) -> bool {
+            matches!(self.state, WantState::Existing)
+        }
     }
 
     fn config_path_home_existing() -> TestCase {
@@ -1271,7 +1321,7 @@ mod tests {
                 home_dir: Some("home".into()),
                 ..Default::default()
             },
-            wants: &[Want::Existing("home/.jjconfig.toml")],
+            wants: vec![Want::existing("home/.jjconfig.toml")],
         }
     }
 
@@ -1282,7 +1332,7 @@ mod tests {
                 home_dir: Some("home".into()),
                 ..Default::default()
             },
-            wants: &[Want::New("home/.jjconfig.toml")],
+            wants: vec![Want::new("home/.jjconfig.toml")],
         }
     }
 
@@ -1294,9 +1344,9 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[
-                Want::Existing("home/.jjconfig.toml"),
-                Want::New("config/jj/config.toml"),
+            wants: vec![
+                Want::existing("home/.jjconfig.toml"),
+                Want::new("config/jj/config.toml"),
             ],
         }
     }
@@ -1309,7 +1359,7 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[Want::Existing("config/jj/config.toml")],
+            wants: vec![Want::existing("config/jj/config.toml")],
         }
     }
 
@@ -1320,7 +1370,7 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[Want::New("config/jj/config.toml")],
+            wants: vec![Want::new("config/jj/config.toml")],
         }
     }
 
@@ -1332,7 +1382,7 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[Want::New("config/jj/config.toml")],
+            wants: vec![Want::new("config/jj/config.toml")],
         }
     }
 
@@ -1343,7 +1393,7 @@ mod tests {
                 jj_config: Some("custom.toml".into()),
                 ..Default::default()
             },
-            wants: &[Want::Existing("custom.toml")],
+            wants: vec![Want::existing("custom.toml")],
         }
     }
 
@@ -1354,7 +1404,7 @@ mod tests {
                 jj_config: Some("custom.toml".into()),
                 ..Default::default()
             },
-            wants: &[Want::New("custom.toml")],
+            wants: vec![Want::new("custom.toml")],
         }
     }
 
@@ -1370,9 +1420,9 @@ mod tests {
                 ),
                 ..Default::default()
             },
-            wants: &[
-                Want::Existing("custom1.toml"),
-                Want::Existing("custom2.toml"),
+            wants: vec![
+                Want::existing("custom1.toml"),
+                Want::existing("custom2.toml"),
             ],
         }
     }
@@ -1389,7 +1439,7 @@ mod tests {
                 ),
                 ..Default::default()
             },
-            wants: &[Want::Existing("custom1.toml"), Want::New("custom2.toml")],
+            wants: vec![Want::existing("custom1.toml"), Want::new("custom2.toml")],
         }
     }
 
@@ -1405,7 +1455,7 @@ mod tests {
                 ),
                 ..Default::default()
             },
-            wants: &[Want::Existing("custom1.toml"), Want::New("custom2.toml")],
+            wants: vec![Want::existing("custom1.toml"), Want::new("custom2.toml")],
         }
     }
 
@@ -1416,7 +1466,7 @@ mod tests {
                 jj_config: Some("".to_owned()),
                 ..Default::default()
             },
-            wants: &[],
+            wants: vec![],
         }
     }
 
@@ -1428,7 +1478,7 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[Want::Existing("config/jj/config.toml")],
+            wants: vec![Want::existing("config/jj/config.toml")],
         }
     }
 
@@ -1440,9 +1490,9 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[
-                Want::Existing("home/.jjconfig.toml"),
-                Want::New("config/jj/config.toml"),
+            wants: vec![
+                Want::existing("home/.jjconfig.toml"),
+                Want::new("config/jj/config.toml"),
             ],
         }
     }
@@ -1455,9 +1505,9 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[
-                Want::New("config/jj/config.toml"),
-                Want::Existing("config/jj/conf.d"),
+            wants: vec![
+                Want::new("config/jj/config.toml"),
+                Want::existing("config/jj/conf.d"),
             ],
         }
     }
@@ -1470,9 +1520,9 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[
-                Want::Existing("config/jj/config.toml"),
-                Want::Existing("config/jj/conf.d"),
+            wants: vec![
+                Want::existing("config/jj/config.toml"),
+                Want::existing("config/jj/conf.d"),
             ],
         }
     }
@@ -1490,10 +1540,10 @@ mod tests {
                 ..Default::default()
             },
             // Precedence order is important
-            wants: &[
-                Want::Existing("home/.jjconfig.toml"),
-                Want::Existing("config/jj/config.toml"),
-                Want::Existing("config/jj/conf.d"),
+            wants: vec![
+                Want::existing("home/.jjconfig.toml"),
+                Want::existing("config/jj/config.toml"),
+                Want::existing("config/jj/conf.d"),
             ],
         }
     }
@@ -1502,7 +1552,7 @@ mod tests {
         TestCase {
             files: &[],
             env: Default::default(),
-            wants: &[],
+            wants: vec![],
         }
     }
 
@@ -1528,27 +1578,23 @@ mod tests {
         let tmp = setup_config_fs(case.files);
         let env = resolve_config_env(&case.env, tmp.path());
 
-        let expected_existing: Vec<PathBuf> = case
+        let all_expected_paths = case
             .wants
             .iter()
-            .filter_map(|want| match want {
-                Want::Existing(path) => Some(tmp.path().join(path)),
-                _ => None,
-            })
-            .collect();
-        assert_eq!(
-            env.existing_user_config_paths().collect_vec(),
-            expected_existing
-        );
+            .map(|w| w.rooted_path(tmp.path()))
+            .collect_vec();
+        let exists_expected_paths = case
+            .wants
+            .iter()
+            .filter(|w| w.exists())
+            .map(|w| w.rooted_path(tmp.path()))
+            .collect_vec();
 
-        let expected_paths: Vec<PathBuf> = case
-            .wants
-            .iter()
-            .map(|want| match want {
-                Want::New(path) | Want::Existing(path) => tmp.path().join(path),
-            })
-            .collect();
-        assert_eq!(env.user_config_paths().collect_vec(), expected_paths);
+        let all_paths = env.user_config_paths().collect_vec();
+        let exists_paths = env.existing_user_config_paths().collect_vec();
+
+        assert_eq!(all_paths, all_expected_paths);
+        assert_eq!(exists_paths, exists_expected_paths);
     }
 
     fn setup_config_fs(files: &[&str]) -> tempfile::TempDir {
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -218,6 +218,7 @@ fn create_dir_all(path: &Path) -> std::io::Result<()> {
 #[derive(Clone, Default, Debug)]
 struct UnresolvedConfigEnv {
     config_dir: Option<PathBuf>,
+    macos_legacy_config_dir: Option<PathBuf>,
     home_dir: Option<PathBuf>,
     jj_config: Option<String>,
 }
@@ -246,9 +247,22 @@ impl UnresolvedConfigEnv {
             config_dir.push("conf.d");
             ConfigPath::new(config_dir)
         });
+        let legacy_platform_config_path =
+            self.macos_legacy_config_dir.clone().map(|mut config_dir| {
+                config_dir.push("jj");
+                config_dir.push("config.toml");
+                ConfigPath::new(config_dir)
+            });
+        let legacy_platform_config_dir = self.macos_legacy_config_dir.map(|mut config_dir| {
+            config_dir.push("jj");
+            config_dir.push("conf.d");
+            ConfigPath::new(config_dir)
+        });
 
         if let Some(path) = home_config_path {
-            if path.exists() || platform_config_path.is_none() {
+            if path.exists()
+                || (platform_config_path.is_none() && legacy_platform_config_path.is_none())
+            {
                 paths.push(path);
             }
         }
@@ -259,13 +273,25 @@ impl UnresolvedConfigEnv {
             paths.push(path);
         }
 
-        // theoretically this should be an `if let Some(...) = ... && ..., but that
+        // theoretically these should be an `if let Some(...) = ... && ..., but that
         // isn't stable
         if let Some(path) = platform_config_dir {
             if path.exists() {
                 paths.push(path);
             }
         }
+
+        if let Some(path) = legacy_platform_config_path {
+            if path.exists() {
+                paths.push(path);
+            }
+        }
+        if let Some(path) = legacy_platform_config_dir {
+            if path.exists() {
+                paths.push(path);
+            }
+        }
+
         paths
     }
 }
@@ -282,7 +308,13 @@ pub struct ConfigEnv {
 impl ConfigEnv {
     /// Initializes configuration loader based on environment variables.
     pub fn from_environment() -> Self {
-        let config_dir = if cfg!(target_os = "macos") {
+        let config_dir = etcetera::choose_base_strategy()
+            .ok()
+            .map(|s| s.config_dir());
+
+        // older versions of jj used a more "GUI" config option,
+        // which is not designed for user-editable configuration of CLI utilities.
+        let macos_legacy_config_dir = if cfg!(target_os = "macos") {
             etcetera::base_strategy::choose_native_strategy()
                 .ok()
                 .map(|s| {
@@ -291,9 +323,7 @@ impl ConfigEnv {
                     s.data_dir()
                 })
         } else {
-            etcetera::choose_base_strategy()
-                .ok()
-                .map(|s| s.config_dir())
+            None
         };
 
         // Canonicalize home as we do canonicalize cwd in CliRunner. $HOME might
@@ -304,6 +334,7 @@ impl ConfigEnv {
 
         let env = UnresolvedConfigEnv {
             config_dir,
+            macos_legacy_config_dir,
             home_dir: home_dir.clone(),
             jj_config: env::var("JJ_CONFIG").ok(),
         };
@@ -1574,6 +1605,51 @@ mod tests {
         }
     }
 
+    fn config_path_macos_legacy_exists() -> TestCase {
+        TestCase {
+            files: &["macos-legacy/jj/config.toml"],
+            env: UnresolvedConfigEnv {
+                home_dir: Some("home".into()),
+                config_dir: Some("config".into()),
+                macos_legacy_config_dir: Some("macos-legacy".into()),
+                ..Default::default()
+            },
+            wants: vec![
+                Want::new("config/jj/config.toml"),
+                Want::existing("macos-legacy/jj/config.toml"),
+            ],
+        }
+    }
+
+    fn config_path_macos_legacy_both_exist() -> TestCase {
+        TestCase {
+            files: &["macos-legacy/jj/config.toml", "config/jj/config.toml"],
+            env: UnresolvedConfigEnv {
+                home_dir: Some("home".into()),
+                config_dir: Some("config".into()),
+                macos_legacy_config_dir: Some("macos-legacy".into()),
+                ..Default::default()
+            },
+            wants: vec![
+                Want::existing("config/jj/config.toml"),
+                Want::existing("macos-legacy/jj/config.toml"),
+            ],
+        }
+    }
+
+    fn config_path_macos_legacy_new() -> TestCase {
+        TestCase {
+            files: &[],
+            env: UnresolvedConfigEnv {
+                home_dir: Some("home".into()),
+                config_dir: Some("config".into()),
+                macos_legacy_config_dir: Some("macos-legacy".into()),
+                ..Default::default()
+            },
+            wants: vec![Want::new("config/jj/config.toml")],
+        }
+    }
+
     #[test_case(config_path_home_existing())]
     #[test_case(config_path_home_new())]
     #[test_case(config_path_home_existing_platform_new())]
@@ -1592,6 +1668,9 @@ mod tests {
     #[test_case(config_path_platform_existing_conf_dir_existing())]
     #[test_case(config_path_all_existing())]
     #[test_case(config_path_none())]
+    #[test_case(config_path_macos_legacy_exists())]
+    #[test_case(config_path_macos_legacy_both_exist())]
+    #[test_case(config_path_macos_legacy_new())]
     fn test_config_path(case: TestCase) {
         let tmp = setup_config_fs(case.files);
         let env = resolve_config_env(&case.env, tmp.path());
@@ -1631,6 +1710,7 @@ mod tests {
         let home_dir = env.home_dir.as_ref().map(|p| root.join(p));
         let env = UnresolvedConfigEnv {
             config_dir: env.config_dir.as_ref().map(|p| root.join(p)),
+            macos_legacy_config_dir: env.macos_legacy_config_dir.as_ref().map(|p| root.join(p)),
             home_dir: home_dir.clone(),
             jj_config: env.jj_config.as_ref().map(|p| {
                 join_paths(split_paths(p).map(|p| {
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -43,6 +43,7 @@ use crate::command_error::config_error;
 use crate::command_error::config_error_with_message;
 use crate::command_error::CommandError;
 use crate::text_util;
+use crate::ui::Ui;
 
 // TODO(#879): Consider generating entire schema dynamically vs. static file.
 pub const CONFIG_SCHEMA: &str = include_str!("config-schema.json");
@@ -192,7 +193,6 @@ impl ConfigPath {
     fn as_path(&self) -> &Path {
         &self.path
     }
-
     fn exists(&self) -> bool {
         match self.state {
             ConfigPathState::Exists => true,
@@ -218,13 +218,14 @@ fn create_dir_all(path: &Path) -> std::io::Result<()> {
 #[derive(Clone, Default, Debug)]
 struct UnresolvedConfigEnv {
     config_dir: Option<PathBuf>,
+    // TODO: remove after jj 0.35
     macos_legacy_config_dir: Option<PathBuf>,
     home_dir: Option<PathBuf>,
     jj_config: Option<String>,
 }
 
 impl UnresolvedConfigEnv {
-    fn resolve(self) -> Vec<ConfigPath> {
+    fn resolve(self, ui: &Ui) -> Vec<ConfigPath> {
         if let Some(paths) = self.jj_config {
             return split_paths(&paths)
                 .filter(|path| !path.as_os_str().is_empty())
@@ -283,16 +284,40 @@ impl UnresolvedConfigEnv {
 
         if let Some(path) = legacy_platform_config_path {
             if path.exists() {
+                Self::warn_for_deprecated_path(
+                    ui,
+                    path.as_path(),
+                    "~/Library/Application Support",
+                    "~/.config",
+                );
                 paths.push(path);
             }
         }
         if let Some(path) = legacy_platform_config_dir {
             if path.exists() {
+                Self::warn_for_deprecated_path(
+                    ui,
+                    path.as_path(),
+                    "~/Library/Application Support",
+                    "~/.config",
+                );
                 paths.push(path);
             }
         }
 
         paths
+    }
+
+    fn warn_for_deprecated_path(ui: &Ui, path: &Path, old: &str, new: &str) {
+        let _ = indoc::writedoc!(
+            ui.warning_default(),
+            r"
+            Deprecated configuration file `{}`.
+            Configuration files in `{old}` are deprecated, and support will be removed in a future release.
+            Instead, move your configuration files to `{new}`.
+            ",
+            path.display(),
+        );
     }
 }
 
@@ -307,7 +332,7 @@ pub struct ConfigEnv {
 
 impl ConfigEnv {
     /// Initializes configuration loader based on environment variables.
-    pub fn from_environment() -> Self {
+    pub fn from_environment(ui: &Ui) -> Self {
         let config_dir = etcetera::choose_base_strategy()
             .ok()
             .map(|s| s.config_dir());
@@ -341,7 +366,7 @@ impl ConfigEnv {
         ConfigEnv {
             home_dir,
             repo_path: None,
-            user_config_paths: env.resolve(),
+            user_config_paths: env.resolve(ui),
             repo_config_path: None,
             command: None,
         }
@@ -1727,7 +1752,7 @@ mod tests {
         ConfigEnv {
             home_dir,
             repo_path: None,
-            user_config_paths: env.resolve(),
+            user_config_paths: env.resolve(&Ui::null()),
             repo_config_path: None,
             command: None,
         }
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -22,6 +22,7 @@ use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
 
+use etcetera::BaseStrategy as _;
 use itertools::Itertools as _;
 use jj_lib::config::ConfigFile;
 use jj_lib::config::ConfigGetError;
@@ -281,11 +282,28 @@ pub struct ConfigEnv {
 impl ConfigEnv {
     /// Initializes configuration loader based on environment variables.
     pub fn from_environment() -> Self {
+        let config_dir = if cfg!(target_os = "macos") {
+            etcetera::base_strategy::choose_native_strategy()
+                .ok()
+                .map(|s| {
+                    // note that etcetera calls Library/Application Support the "data dir",
+                    // Library/Preferences is supposed to be exclusively plists
+                    s.data_dir()
+                })
+        } else {
+            etcetera::choose_base_strategy()
+                .ok()
+                .map(|s| s.config_dir())
+        };
+
         // Canonicalize home as we do canonicalize cwd in CliRunner. $HOME might
         // point to symlink.
-        let home_dir = dirs::home_dir().map(|path| dunce::canonicalize(&path).unwrap_or(path));
+        let home_dir = etcetera::home_dir()
+            .ok()
+            .map(|d| dunce::canonicalize(&d).unwrap_or(d));
+
         let env = UnresolvedConfigEnv {
-            config_dir: dirs::config_dir(),
+            config_dir,
             home_dir: home_dir.clone(),
             jj_config: env::var("JJ_CONFIG").ok(),
         };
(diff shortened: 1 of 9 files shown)


martinvonz:
> config: Correct macOS config location

nit: Is the old behavior actually incorrect? Say "Follow XDG spec for macOS config location" or something like that if that's the spec we're talking about.

strega-nil:
@martinvonz I personally believe strongly that it is incorrect. User-editable configuration should not be placed in `~/Library/Application Support` on macOS. I don't even really believe non-user-editable configuration should be in `~/Library/Application Support`, but I don't really have strong opinions on what GUI programs do.

cli/src/config.rs
```diff
@@ -229,20 +231,48 @@ impl UnresolvedConfigEnv {
...
-        } else if let (Some(path @ New(_)), None) = (home_config_path, &platform_config_path) {
-            paths.push(path);
+
+        match home_config_path {
+            Some(path @ Existing(_)) => paths.push(path),
+            Some(path @ New(_)) if platform_config_path.is_none() => paths.push(path),
+            _ => (),
+        }
+
+        #[cfg(target_os = "macos")]
```
PhilipMetzger, line 243: nit: missing TODO to remove after jj 0.35+, since it will no longer be supported then.

cli/src/config.rs
```diff
@@ -259,11 +259,31 @@ pub struct ConfigEnv {
...
+        let config_dir = etcetera::base_strategy::choose_native_strategy()
+            .ok()
+            .map(|s| {
+                use etcetera::BaseStrategy as _;
+                // note that etcetera calls Library/Application Support the "data dir",
+                // Library/Preferences is supposed to be exclusively plists
+                s.data_dir()
+            });
+
```
PhilipMetzger: Nit: add a warning that this behavior is deprecated and will be removed in 6 releases and a TODO for the code removal.

PhilipMetzger (commented):
just some minor deprecation stuff but the patch SGTM

cli/src/config.rs
```diff
@@ -202,6 +202,8 @@ fn create_dir_all(path: &Path) -> std::io::Result<()> {
 #[derive(Clone, Default, Debug)]
 struct UnresolvedConfigEnv {
     config_dir: Option<PathBuf>,
+    #[cfg(target_os = "macos")]
+    macos_legacy_config_dir: Option<PathBuf>,
```
yuja, line 206: Make `config_dirs: Vec<PathBuf>` instead? Fewer platform-specific codes is better, and we wouldn't have to care about extra allocation cost.

strega-nil: That was the original idea, but it really is only macOS, and this behaviour should really be deprecated and removed in a future release.

cli/src/config.rs
```diff
@@ -229,20 +231,48 @@ impl UnresolvedConfigEnv {
...
+                paths.push(path);
+            }
+            if let Some(path @ Existing(_)) = legacy_platform_config_dir {
+                paths.push(path);
+            }
         }
+
         // This should be the default config created if there's
         // no user config and `jj config edit` is executed.
+        // NOTE: this should probably only be pushed if it exists, or no other config
```
strega-nil, line 267: I made a note here so I could bring it up - it seems unfortunate that if you already have a `~/.jjconfig.toml`, for example, when you use `jj config edit --user`, it will pop a chooser for either `~/.jjconfig.toml` _or_ `~/.config/jj/config.toml`, even if the latter does not exist.

cli/src/config.rs
yuja: Should we deprecate the current macOS config path? Since people appears to have different opinions, I think it's good to support both Unixy and macOS paths. There might be Windows guys who prefer XDG path, but I don't know.

cli/src/config.rs
yuja: Sounds like a bug. Existing paths should be prioritized over new paths.

cli/src/config.rs
emilazy: FWIW, I think that the `dirs` author is close to the only person who thinks that the current path is more correct for CLI programs (and it's based on a misleading of Apple's guidelines, e.g. the subdirectory is meant to be a bundle ID, which we don't have as we're not an app bundle, and it's for non-user-authored files managed by the app).

yuja: Thanks. In that case, deprecating non-XDG path makes sense.

cli/src/config.rs
calebdw: It needs to be pushed if there's no other configs as `jj config edit --user` will try to automatically create it

strega-nil: Right, but it shouldn't be pushed if there are other configs.

calebdw: I would agree with that

cli/src/config.rs
```diff
@@ -214,35 +255,63 @@ impl UnresolvedConfigEnv {
...
-        } else if let (Some(path @ New(_)), None) = (home_config_path, &platform_config_path) {
-            paths.push(path);
+            push_if_exists(&mut paths, ConfigPath::new(home_dir))
         }
-        // This should be the default config created if there's
-        // no user config and `jj config edit` is executed.
-        if let Some(path) = platform_config_path {
-            paths.push(path);
+
+        #[cfg(target_os = "macos")]
```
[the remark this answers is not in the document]
strega-nil: We could certainly do that; is that something we want?

cli/src/config.rs
strega-nil: Opened a PR: #6315

cli/src/config.rs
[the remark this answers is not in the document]
emilazy: IMO it's better than conditional compilation whenever possible.

strega-nil:
@wjv I think the arguments against using `~/Library/Application Support` are made fairly clear in the rest of the thread; `jj` is _not_ an application, and therefore the standard _does not apply_, and is incorrectly (imo) used here (and in other Rust tools that depend on `dirs`).

XDG, while not an official standard on macOS, has become a de-facto standard among the vast majority of CLI application developers, and as a person who has used macOS for [redacted] decades at this point, it is significantly more correct to use XDG rather than `~/Library/Application Support` – if someone can find an officially released apple CLI tool that uses `~/Library/Application Support`, I might change my tune, but Apple absolutely ships tools that respect XDG (including, just as an example, vim).

strega-nil:
I'll work on this more this week maybe, but easter has been a bit hectic, so no work on this for now :)

cli/src/commands/config/mod.rs
```diff
@@ -67,11 +66,34 @@ impl ConfigLevelArgs {
...
+                            ui.warning_default(),
+                            "configuration files in `{old}` are deprecated, and support will be \
+                             removed in a future release.",
+                        )?;
+                        writeln!(
+                            ui.warning_default(),
+                            "Instead, move your configuration files to `{new}`.",
+                        )?;
+                    }
+                }
```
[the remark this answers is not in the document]
strega-nil: I'm happy to do this if people agree with it. It would certainly make the code change smaller (although I think switching from an enum to a structure is better anyways)

strega-nil:
Alright, it has been worked on!

cli/src/commands/config/mod.rs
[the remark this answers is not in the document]
yuja: Or split deprecation warning to new PR? I'm not sure if people agree on the interpretation of the macOS config paths, and supporting both locations would be harmless?

https://github.com/jj-vcs/jj/pull/6300#issuecomment-2796692963

> switching from an enum to a structure is better anyways

I agree. FWIW, the last time I considered that, I thought the state would have to be `ExistingFile|ExistingDir|NewFile`. `ConfigEnv` doesn't support new directory path.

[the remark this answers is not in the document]
strega-nil: I can certainly pull it into multiple PRs, but I'd appreciate if you and @PhilipMetzger agreed on what is the correct way forward, since I was told to make it a deprecation warning here in https://github.com/jj-vcs/jj/pull/6300#discussion_r2036107786; I'm happy either way (although I personally think it should be deprecated).

It might be nice to deprecate it now and wait for a scream test (see if anyone has a problem with it being deprecated and opens an issue).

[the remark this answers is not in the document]
PhilipMetzger: > It might be nice to deprecate it now and wait for a scream test (see if anyone has a problem with it being deprecated and opens an issue).

The deprecation message is the _scream test_, since they usually give people enough time to react. 

> I'm not sure if people agree on the interpretation of the macOS config paths, and supporting both locations would be harmless?

I think Jujutsu should follow the XDG standard here and interpret macOS as a strange Unix box in this peculiar case and deprecate the current path. In the end I'll leave the decision to the maintainers though.

strega-nil:
And another update; now, it only warns on all `config` commands, so as to avoid annoying people _too_ much:

Screenshot of jj printing a warning on `jj config edit` and `jj config path`, but not `jj st`

<details markdown="1">
<summary>Interdiff</summary>

```diff
diff --git a/cli/src/commands/config/mod.rs b/cli/src/commands/config/mod.rs
index 5d8fefc3af..4f68fffe90 100644
--- a/cli/src/commands/config/mod.rs
+++ b/cli/src/commands/config/mod.rs
@@ -41,7 +41,6 @@
 use crate::command_error::CommandError;
 use crate::config::ConfigEnv;
 use crate::config::ConfigPath;
-use crate::config::DeprecatedLocation;
 use crate::ui::Ui;
 
 #[derive(clap::Args, Clone, Debug)]
@@ -73,32 +72,10 @@
         config_env: &'a ConfigEnv,
     ) -> Result<Vec<&'a ConfigPath>, CommandError> {
         if self.user {
+            config_env.warn_for_deprecated_paths(&ui);
             let paths = config_env.user_config_paths().collect_vec();
             if paths.is_empty() {
                 return Err(user_error("No user config path found"));
-            } else {
-                for p in &paths {
-                    if let Some(DeprecatedLocation {
-                        old_location,
-                        new_location,
-                    }) = p.deprecated_location()
-                    {
-                        writeln!(
-                            ui.warning_default(),
-                            "Deprecated configuration file `{}`:",
-                            p.path().display()
-                        )?;
-                        writeln!(
-                            ui.warning_default(),
-                            "configuration files in `{old_location}` are deprecated, and support \
-                             will be removed in a future release.",
-                        )?;
-                        writeln!(
-                            ui.warning_default(),
-                            "Instead, move your configuration files to `{new_location}`.",
-                        )?;
-                    }
-                }
             }
             Ok(paths)
         } else if self.repo {
@@ -134,6 +111,7 @@
             files.pop().ok_or_else(|| user_error(not_found_error))
         };
         if self.user {
+            config_env.warn_for_deprecated_paths(&ui);
             pick_one(
                 config_env.user_config_files(config)?,
                 "No user config path found to edit",
diff --git a/cli/src/config.rs b/cli/src/config.rs
index 66afd4456d..80b959fbf8 100644
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -40,6 +40,7 @@
 use crate::command_error::config_error;
 use crate::command_error::config_error_with_message;
 use crate::command_error::CommandError;
+use crate::ui::Ui;
 
 // TODO(#879): Consider generating entire schema dynamically vs. static file.
 pub const CONFIG_SCHEMA: &str = include_str!("config-schema.json");
@@ -161,9 +162,9 @@
 }
 
 #[derive(Copy, Clone, Debug)]
-pub(crate) struct DeprecatedLocation {
-    pub(crate) old_location: &'static str,
-    pub(crate) new_location: &'static str,
+struct DeprecatedLocation {
+    old: &'static str,
+    new: &'static str,
 }
 #[derive(Clone, Debug)]
 enum ConfigPathState {
@@ -186,17 +187,10 @@
             path,
         }
     }
-    fn new_deprecated(
-        path: PathBuf,
-        old_location: &'static str,
-        new_location: &'static str,
-    ) -> Self {
+    fn new_deprecated(path: PathBuf, old: &'static str, new: &'static str) -> Self {
         use ConfigPathState::*;
         let state = if path.exists() {
-            Deprecated(DeprecatedLocation {
-                old_location,
-                new_location,
-            })
+            Deprecated(DeprecatedLocation { old, new })
         } else {
             New
         };
@@ -213,7 +207,7 @@
             ConfigPathState::New => false,
         }
     }
-    pub(crate) fn deprecated_location(&self) -> Option<DeprecatedLocation> {
+    fn deprecated_location(&self) -> Option<DeprecatedLocation> {
         match self.state {
             ConfigPathState::Deprecated(d) => Some(d),
             ConfigPathState::Exists | ConfigPathState::New => None,
@@ -371,6 +365,30 @@
         self.command = Some(command);
     }
 
+    pub fn warn_for_deprecated_paths(&self, ui: &Ui) {
+        if cfg!(target_os = "macos") {
+            for p in &self.user_config_paths {
+                let Some(DeprecatedLocation { old, new }) = p.deprecated_location() else {
+                    continue;
+                };
+                let _ = writeln!(
+                    ui.warning_default(),
+                    "Deprecated configuration file `{}`:",
+                    p.path().display()
+                );
+                let _ = writeln!(
+                    ui.warning_default(),
+                    "configuration files in `{old}` are deprecated, and support will be removed \
+                     in a future release.",
+                );
+                let _ = writeln!(
+                    ui.warning_default(),
+                    "Instead, move your configuration files to `{new}`.",
+                );
+            }
+        }
+    }
+
     /// Returns the paths to the user-specific config files or directories.
     pub fn user_config_paths(&self) -> impl Iterator<Item = &ConfigPath> {
         self.user_config_paths.iter()
@@ -379,7 +397,7 @@
     /// Returns the paths to the existing user-specific config files or
     /// directories.
     pub fn existing_user_config_paths(&self) -> impl Iterator<Item = &ConfigPath> {
-        self.user_config_paths.iter().filter(|p| p.exists())
+        self.user_config_paths().filter(|p| p.exists())
     }
 
     /// Returns user configuration files for modification. Instantiates one if
```

</details>

cli/src/commands/config/mod.rs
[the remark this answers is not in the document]
yuja: > I think Jujutsu should follow the XDG standard here and interpret macOS as a strange Unix box in this peculiar case and deprecate the current path.

I personally don't have any opinion (except that I prefer simpler implementation.) If macOS users are happy with that, it's okay. I just pointed out that there are conflicting opinions like https://github.com/jj-vcs/jj/pull/6300#issuecomment-2790937854 and https://github.com/jj-vcs/jj/pull/6300#issuecomment-2796692963, so I'm not sure.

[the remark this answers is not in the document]
PhilipMetzger: I think in the worst case, we'll have to support both but currently the most outspoken voices prefer the XDG Standard one. I'm fine to support both if there are as many complaints about it when the path should be removed or totally deprecated.

cli/src/config.rs
```diff
@@ -259,11 +259,31 @@ pub struct ConfigEnv {
 impl ConfigEnv {
     /// Initializes configuration loader based on environment variables.
     pub fn from_environment() -> Self {
+        #[cfg(not(target_os = "macos"))]
```
martinvonz: nit: I think it's generally preferred to do `if cfg!(target_os = "macos")` so that it's compile-tested on all platforms.

cli/src/config.rs
```diff
@@ -160,27 +161,56 @@ impl AsMut<StackedConfig> for RawConfig {
...
-            ConfigPath::Existing(path)
-        } else {
-            ConfigPath::New(path)
+        use ConfigPathState::*;
+        ConfigPath {
+            state: if path.exists() { Exists } else { New },
+            path,
         }
     }
+    fn new_deprecated(path: PathBuf, old: &'static str, new: &'static str) -> Self {
```
martinvonz, line 190: nit: we generally put a blank line between functions

cli/src/config.rs
```diff
@@ -229,12 +230,36 @@ impl UnresolvedConfigEnv {
...
-        if let Some(path @ Existing(_)) = home_config_path {
-            paths.push(path);
-        } else if let (Some(path @ New(_)), None) = (home_config_path, &platform_config_path) {
-            paths.push(path);
+
+        match home_config_path {
+            Some(path @ Existing(_)) => paths.push(path),
+            Some(path @ New(_)) if platform_config_path.is_none() => paths.push(path),
+            _ => (),
+        }
```
martinvonz: This `match home_config_path` block looks like an unrelated refactoring. Especially since it seems like it gets removed again in the last commit, could you undo the refactoring in the second commit? Sorry if I'm missing something

cli/src/config.rs
```diff
@@ -215,58 +246,62 @@ impl UnresolvedConfigEnv {
...
-                config_dir.push("conf.d");
-                ConfigPath::new(config_dir)
-            });
-
-            if let Some(path @ Existing(_)) = legacy_platform_config_path {
-                paths.push(path);
-            }
-            if let Some(path @ Existing(_)) = legacy_platform_config_dir {
-                paths.push(path);
+            if let Some(mut legacy_platform_config_path) = self.macos_legacy_config_dir {
```
martinvonz, line 268: nit: I suppose we can drop the `if cfg!(target_os = "macos")` condition now since `self. macos_legacy_config_dir` would be `None` on other platforms anyway

cli/src/config.rs
```diff
@@ -215,58 +246,62 @@ impl UnresolvedConfigEnv {
...
         }
 
-        // This should be the default config created if there's
-        // no user config and `jj config edit` is executed.
-        if let Some(path) = platform_config_path {
-            paths.push(path);
-        }
-        if let Some(path @ Existing(_)) = platform_config_dir {
-            paths.push(path);
+        if let Some(mut platform_config_path) = self.config_dir.clone() {
```
martinvonz, line 294: nit: unnecessary clone?

cli/src/commands/config/mod.rs
```diff
@@ -67,8 +66,13 @@ impl ConfigLevelArgs {
...
     }
 
-    fn config_paths<'a>(&self, config_env: &'a ConfigEnv) -> Result<Vec<&'a Path>, CommandError> {
+    fn config_paths<'a>(
+        &self,
+        ui: &Ui,
+        config_env: &'a ConfigEnv,
+    ) -> Result<Vec<&'a ConfigPath>, CommandError> {
         if self.user {
+            config_env.warn_for_deprecated_paths(ui);
```
yuja, line 75: Does this mean the warning won't be printed unless the user tried to edit/set/list/get the config file? That would be bad because the legacy config file would silently become ignored.

As I said, I would make `ConfigEnv::from_environment()` emit the warning because that's simpler and we'll remove it later.

cli/src/config.rs
```diff
@@ -162,27 +163,65 @@ impl AsMut<StackedConfig> for RawConfig {
...
+#[derive(Copy, Clone, Debug)]
+struct DeprecatedLocation {
+    old: &'static str,
+    new: &'static str,
+}
+#[derive(Clone, Debug)]
+enum ConfigPathState {
+    New,
+    Exists,
+    Deprecated(DeprecatedLocation),
```
yuja, line 175: nit: perhaps, `Deprecated` can be moved to `ConfigPath { .., deprecated: Option<DeprecatedLocation> }`? The deprecated state is different axis than file/directory existence.

cli/src/config.rs
```diff
@@ -162,27 +163,65 @@ impl AsMut<StackedConfig> for RawConfig {
...
     fn new(path: PathBuf) -> Self {
-        if path.exists() {
-            ConfigPath::Existing(path)
-        } else {
-            ConfigPath::New(path)
+        use ConfigPathState::*;
+        ConfigPath {
+            state: if path.exists() { Exists } else { New },
+            path,
         }
```
calebdw, line 198: nit: is there a reason these aren't provided in the order they're defined?

```suggestion
        ConfigPath {
            path,
            state: if path.exists() { Exists } else { New },
        }
```

cli/src/config.rs
```diff
@@ -162,27 +163,65 @@ impl AsMut<StackedConfig> for RawConfig {
...
-            ConfigPath::Existing(path) | ConfigPath::New(path) => path,
+    fn new_deprecated(path: PathBuf, old: &'static str, new: &'static str) -> Self {
+        use ConfigPathState::*;
+        let state = if path.exists() {
+            Deprecated(DeprecatedLocation { old, new })
+        } else {
+            New
+        };
+
+        ConfigPath { state, path }
```
calebdw, line 209: nit: same thing here

```suggestion
        ConfigPath { path, state }
```

cli/src/config.rs
```diff
@@ -162,27 +163,65 @@ impl AsMut<StackedConfig> for RawConfig {
...
+    }
+
+    pub fn path(&self) -> &Path {
+        &self.path
+    }
```
calebdw, line 214: nit: it seems like the majority of references use `path.path` instead of using this getter---I would say for consistency that everything should use the getter or this should just be removed

calebdw (approved):
The implementation looks good to me

strega-nil:
I'll come back Monday hopefully and fix people's nits!

cli/src/config.rs
strega-nil: Yes, because the `path` line moves the path that the `state` line refers to.

cli/src/config.rs
strega-nil: It's technically a different axis, but then we'd have:

```rust
#[derive(Clone, Debug)]
pub struct ConfigPath {
    path: PathBuf,
    state: ConfigPathState,
    deprecated: Option<DeprecatedLocation>, // only Some if state == Exists
}
```

which I... don't love. I can do that though.

yuja: I think it's technically valid to have deprecated "New" path. That said, we won't have to record the deprecated state anyway if we address https://github.com/jj-vcs/jj/pull/6300#discussion_r2059420658.

strega-nil:
I hope I've resolved everyone's nits?

<details>
<summary>range-diff</summary>

```
1:  5481bc5a2 ! 1:  e99dbbdaf config: change how config paths are represented
    @@ cli/src/config.rs: impl ConfigEnv {
                      // No need to propagate io::Error here. If the directory
                      // couldn't be created, file.save() would fail later.
     -                if let Some(dir) = path.parent() {
    -+                if let Some(dir) = path.path.parent() {
    ++                if let Some(dir) = path.path().parent() {
                          create_dir_all(dir).ok();
                      }
                      // The path doesn't usually exist, but we shouldn't overwrite it
                      // with an empty config if it did exist.
     -                ConfigFile::load_or_empty(ConfigSource::User, path)
    -+                ConfigFile::load_or_empty(ConfigSource::User, &path.path)
    ++                ConfigFile::load_or_empty(ConfigSource::User, path.path())
                  })
                  .transpose()
          }
    @@ cli/src/config.rs: impl ConfigEnv {
              for path in self.existing_user_config_paths() {
     -            if path.is_dir() {
     -                config.as_mut().load_dir(ConfigSource::User, path)?;
    -+            if path.path.is_dir() {
    -+                config.as_mut().load_dir(ConfigSource::User, &path.path)?;
    ++            if path.path().is_dir() {
    ++                config.as_mut().load_dir(ConfigSource::User, path.path())?;
                  } else {
     -                config.as_mut().load_file(ConfigSource::User, path)?;
    -+                config.as_mut().load_file(ConfigSource::User, &path.path)?;
    ++                config.as_mut().load_file(ConfigSource::User, path.path())?;
                  }
              }
              Ok(())
    @@ cli/src/config.rs: impl ConfigEnv {
     -        match &self.repo_config_path {
     -            Some(ConfigPath::Existing(path)) => Some(path),
     +        match self.repo_config_path {
    -+            Some(ref path) if path.exists() => Some(&path.path),
    ++            Some(ref path) if path.exists() => Some(path.path()),
                  _ => None,
              }
          }
    @@ cli/src/config.rs: impl ConfigEnv {
                  // The path doesn't usually exist, but we shouldn't overwrite it
                  // with an empty config if it did exist.
     -            .map(|path| ConfigFile::load_or_empty(ConfigSource::Repo, path))
    -+            .map(|path| ConfigFile::load_or_empty(ConfigSource::Repo, &path.path))
    ++            .map(|path| ConfigFile::load_or_empty(ConfigSource::Repo, path.path()))
                  .transpose()
          }
      
    @@ cli/src/config.rs: mod tests {
     +        }
     +
     +        fn matches(&self, root: &Path, path: &ConfigPath) -> bool {
    -+            if root.join(self.path) != path.path {
    ++            if root.join(self.path) != path.path() {
     +                return false;
     +            }
     +
2:  16d878a97 = 2:  18d9c6d4b config: switch to etcetera from dirs
3:  40ad94811 = 3:  9f3f1c416 config: default to XDG config files on macOS
4:  8caf64747 ! 4:  9eaf876d5 config: deprecate macOS legacy platform configs
    @@ CHANGELOG.md: to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
      
      * Color-words diff has gained [an option to compare conflict pairs without
     
    - ## cli/src/commands/config/mod.rs ##
    -@@ cli/src/commands/config/mod.rs: impl ConfigLevelArgs {
    - 
    -     fn config_paths<'a>(
    -         &self,
    -+        ui: &Ui,
    -         config_env: &'a ConfigEnv,
    -     ) -> Result<Vec<&'a ConfigPath>, CommandError> {
    -         if self.user {
    -+            config_env.warn_for_deprecated_paths(ui);
    -             let paths = config_env.user_config_paths().collect_vec();
    -             if paths.is_empty() {
    -                 return Err(user_error("No user config path found"));
    -@@ cli/src/commands/config/mod.rs: impl ConfigLevelArgs {
    -             files.pop().ok_or_else(|| user_error(not_found_error))
    -         };
    -         if self.user {
    -+            config_env.warn_for_deprecated_paths(ui);
    -             pick_one(
    -                 config_env.user_config_files(config)?,
    -                 "No user config path found to edit",
    + ## cli/src/cli_util.rs ##
    +@@ cli/src/cli_util.rs: impl CliRunner {
    +                     "Did you update to a commit where the directory doesn't exist?",
    +                 )
    +             })?;
    +-        let mut config_env = ConfigEnv::from_environment();
    ++        let mut config_env = ConfigEnv::from_environment(ui);
    +         let mut last_config_migration_descriptions = Vec::new();
    +         let mut migrate_config = |config: &mut StackedConfig| -> Result<(), CommandError> {
    +             last_config_migration_descriptions =
     
    - ## cli/src/commands/config/path.rs ##
    -@@ cli/src/commands/config/path.rs: pub fn cmd_config_path(
    -     command: &CommandHelper,
    -     args: &ConfigPathArgs,
    - ) -> Result<(), CommandError> {
    --    for config_path in args.level.config_paths(command.config_env())? {
    -+    for config_path in args.level.config_paths(ui, command.config_env())? {
    -         writeln!(
    -             ui.stdout(),
    -             "{}",
    + ## cli/src/complete.rs ##
    +@@ cli/src/complete.rs: fn get_jj_command() -> Result<(JjBuilder, UserSettings), CommandError> {
    +         .and_then(dunce::canonicalize)
    +         .map_err(user_error)?;
    +     // No config migration for completion. Simply ignore deprecated variables.
    +-    let mut config_env = ConfigEnv::from_environment();
    ++    let mut config_env = ConfigEnv::from_environment(&ui);
    +     let maybe_cwd_workspace_loader = DefaultWorkspaceLoaderFactory.create(find_workspace_dir(&cwd));
    +     let _ = config_env.reload_user_config(&mut raw_config);
    +     if let Ok(loader) = &maybe_cwd_workspace_loader {
     
      ## cli/src/config.rs ##
     @@ cli/src/config.rs: use crate::command_error::config_error;
    @@ cli/src/config.rs: impl AsMut<StackedConfig> for RawConfig {
      #[derive(Clone, Debug)]
      enum ConfigPathState {
          New,
    -     Exists,
    -+    Deprecated(DeprecatedLocation),
    - }
    - 
    - /// A ConfigPath can be in one of two states:
     @@ cli/src/config.rs: enum ConfigPathState {
      /// - exists(): a config file exists at the path
      /// - !exists(): a config file doesn't exist here, but a new file _can_ be
      ///   created at this path
     +///
    -+/// Additionally, if it exists, it may be deprecated. In this case, jj will warn
    -+/// when ConfigEnv::warn_for_deprecated_paths is called.
    ++/// Additionally, if it exists, it may be deprecated. In this case, jj will warn the user about the
    ++/// deprecated path.
      #[derive(Clone, Debug)]
      pub struct ConfigPath {
          path: PathBuf,
    +     state: ConfigPathState,
    ++    deprecated: Option<DeprecatedLocation>, // only Some if state == Exists
    + }
    + 
    + impl ConfigPath {
     @@ cli/src/config.rs: impl ConfigPath {
    +         ConfigPath {
    +             state: if path.exists() { Exists } else { New },
    +             path,
    ++            deprecated: None,
              }
          }
      
    -+    fn new_deprecated(path: PathBuf, old: &'static str, new: &'static str) -> Self {
    -+        use ConfigPathState::*;
    -+        let state = if path.exists() {
    -+            Deprecated(DeprecatedLocation { old, new })
    -+        } else {
    -+            New
    -+        };
    -+
    -+        ConfigPath { state, path }
    -+    }
    -+
          pub fn path(&self) -> &Path {
              &self.path
          }
     -
          pub(crate) fn exists(&self) -> bool {
              match self.state {
    --            ConfigPathState::Exists => true,
    -+            ConfigPathState::Exists | ConfigPathState::Deprecated(_) => true,
    -             ConfigPathState::New => false,
    -         }
    -     }
    -+    fn deprecated_location(&self) -> Option<DeprecatedLocation> {
    -+        match self.state {
    -+            ConfigPathState::Deprecated(d) => Some(d),
    -+            ConfigPathState::Exists | ConfigPathState::New => None,
    -+        }
    -+    }
    - }
    - 
    - /// Like std::fs::create_dir_all but creates new directories to be accessible to
    +             ConfigPathState::Exists => true,
     @@ cli/src/config.rs: fn create_dir_all(path: &Path) -> std::io::Result<()> {
      #[derive(Clone, Default, Debug)]
      struct UnresolvedConfigEnv {
    @@ cli/src/config.rs: impl UnresolvedConfigEnv {
                      config_dir.push("jj");
                      config_dir.push("config.toml");
     -                ConfigPath::new(config_dir)
    -+                ConfigPath::new_deprecated(config_dir, "~/Library/Application Support", "~/.config")
    ++                ConfigPath {
    ++                    deprecated: Some(DeprecatedLocation { old: "~/Library/Application Support", new: "~/.config" }),
    ++                    ..ConfigPath::new(config_dir)
    ++                }
                  });
              let legacy_platform_config_dir = self.macos_legacy_config_dir.map(|mut config_dir| {
                  config_dir.push("jj");
                  config_dir.push("conf.d");
     -            ConfigPath::new(config_dir)
    -+            ConfigPath::new_deprecated(config_dir, "~/Library/Application Support", "~/.config")
    ++            ConfigPath {
    ++                deprecated: Some(DeprecatedLocation { old: "~/Library/Application Support", new: "~/.config" }),
    ++                ..ConfigPath::new(config_dir)
    ++            }
              });
      
              if let Some(path) = home_config_path {
    +@@ cli/src/config.rs: pub struct ConfigEnv {
    + 
    + impl ConfigEnv {
    +     /// Initializes configuration loader based on environment variables.
    +-    pub fn from_environment() -> Self {
    ++    pub fn from_environment(ui: &Ui) -> Self {
    +         let config_dir = etcetera::choose_base_strategy().ok().map(|s| {
    +             use etcetera::BaseStrategy as _;
    +             s.config_dir()
    +@@ cli/src/config.rs: impl ConfigEnv {
    +             .ok()
    +             .map(|d| dunce::canonicalize(&d).unwrap_or(d));
    + 
    +-        let env = UnresolvedConfigEnv {
    ++        let user_config_paths = UnresolvedConfigEnv {
    +             config_dir,
    +             macos_legacy_config_dir,
    +             home_dir: home_dir.clone(),
    +             jj_config: env::var("JJ_CONFIG").ok(),
    +-        };
    ++        }.resolve();
    ++        Self::warn_for_deprecated_paths(ui, &user_config_paths);
    ++
    +         ConfigEnv {
    +             home_dir,
    +             repo_path: None,
    +-            user_config_paths: env.resolve(),
    ++            user_config_paths,
    +             repo_config_path: None,
    +             command: None,
    +         }
     @@ cli/src/config.rs: impl ConfigEnv {
              self.command = Some(command);
          }
      
    -+    pub fn warn_for_deprecated_paths(&self, ui: &Ui) {
    ++    fn warn_for_deprecated_paths(ui: &Ui, paths: &[ConfigPath]) {
     +        if cfg!(target_os = "macos") {
    -+            for p in &self.user_config_paths {
    -+                let Some(DeprecatedLocation { old, new }) = p.deprecated_location() else {
    ++            for p in paths {
    ++                let Some(DeprecatedLocation { old, new }) = p.deprecated else {
     +                };
     +                let _ = writeln!(
    @@ cli/src/config.rs: impl ConfigEnv {
     +                );
     +                let _ = writeln!(
    -+                    "configuration files in `{old}` are deprecated, and support will be removed \
    ++                    "Configuration files in `{old}` are deprecated, and support will be removed \
     +                     in a future release.",
     +                );
     +                let _ = writeln!(
    @@ cli/src/config.rs: mod tests {
     +        }
      
              fn matches(&self, root: &Path, path: &ConfigPath) -> bool {
    -             if root.join(self.path) != path.path {
    +             if root.join(self.path) != path.path() {
     @@ cli/src/config.rs: mod tests {
      
                  match self.state {
                      WantState::New => !path.exists(),
     -                WantState::Existing => path.exists(),
    -+                WantState::Existing => path.exists() && path.deprecated_location().is_none(),
    -+                WantState::Deprecated => path.exists() && path.deprecated_location().is_some(),
    ++                WantState::Existing => path.exists() && path.deprecated.is_none(),
    ++                WantState::Deprecated => path.exists() && path.deprecated.is_some(),
                  }
              }
          }
```
</details>

cli/src/commands/config/mod.rs
yuja: > The idea of making this warning not respect how the user set up their stuff feels very wrong to me...

What do you mean by "not respect how the user set up their stuff"?

`ConfigEnv::from_environment(ui)` -> `UnresolvedConfigEnv::resolve(ui)` look up config paths, so it can emit the warning only when a file exists at the deprecated location.

Do you mean it's wrong to not respect `--color`, etc.? I agree, but I think that's acceptable given this warning is temporary, and the user will have to react immediately to suppress the warning.

cli/src/config.rs
```diff
@@ -231,18 +245,26 @@ impl UnresolvedConfigEnv {
...
         }
-        if let Some(path @ Existing(_)) = platform_config_dir {
+
+        // theoretically this should be an `if let Some(...) = ... && ..., but that
+        // isn't stable
+        let get_existing = |opt: Option<ConfigPath>| match opt {
+            Some(path) if path.exists() => Some(path),
+            _ => None,
+        };
+        if let Some(path) = get_existing(platform_config_dir) {
```
martinvonz, line 304: If I'm reading this correctly it can be written more simply like this:

```rust
        if let Some(path) = platform_config_dir {
            if path.exists() {
              paths.push(path);
            }
        }
```

I think this structure will also make it easy to spot as a place where we can use a chained "if let" once stabilized.

cli/src/commands/config/mod.rs
martinvonz, line 1: > config: change how config paths are represented

This doesn't give me much hint to help me review this commit. It's fine since Yuya has already reviewed it now, but I would have appreciated more information if I were to review it. It looks like it's mainly about extracting a `ConfigPathState` from `ConfigPath`. Why is that useful?

It would also be good explain why the signature of `config_paths()` changed. I.e., why is the new signature better (or why will it be better soon)?

strega-nil:
Added some more info, and removed `get_existing`

<details>
<summary>range-diff</summary>

* 1:  bcd366b54 ! 1:  f4a73faf0 config: change how config paths are represented
```diff
@@ Metadata
  ## Commit message ##
     config: change how config paths are represented
 
+    This change, from an enum to a struct, is a more accurate representation
+    of the actual way that a ConfigPath works; additionally, it lets us add
+    different information without modifying every single enumeration field.
+
+    The change to `config_paths()`, to return a `ConfigPath` instead of a
+    `Path`, is less interesting, but in my opinion results in slightly more
+    obvious code. It also makes the test code significantly easier.
+
  ## cli/src/commands/config/mod.rs ##
 @@ cli/src/commands/config/mod.rs: mod path;
  mod set;
```
* 2:  8a31c0b81 ! 2:  f82fac4ea config: switch to etcetera from dirs
```diff
@@ cli/Cargo.toml: clap_complete_nushell = { workspace = true }
  gix = { workspace = true, optional = true }
 
  ## cli/src/config.rs ##
+@@ cli/src/config.rs: impl UnresolvedConfigEnv {
+ 
+         // theoretically this should be an `if let Some(...) = ... && ..., but that
+         // isn't stable
+-        let get_existing = |opt: Option<ConfigPath>| match opt {
+-            Some(path) if path.exists() => Some(path),
+-            _ => None,
+-        };
+-        if let Some(path) = get_existing(platform_config_dir) {
+-            paths.push(path);
++        if let Some(path) = platform_config_dir {
++            if path.exists() {
++                paths.push(path);
++            }
+         }
+         paths
+     }
 @@ cli/src/config.rs: pub struct ConfigEnv {
  impl ConfigEnv {
      /// Initializes configuration loader based on environment variables.
```
* 3:  a4a790ea3 ! 3:  57b4cd795 config: default to XDG config files on macOS
```diff
@@ cli/src/config.rs: impl UnresolvedConfigEnv {
              }
          }
 @@ cli/src/config.rs: impl UnresolvedConfigEnv {
-         if let Some(path) = get_existing(platform_config_dir) {
              paths.push(path);
          }
+ 
+-        // theoretically this should be an `if let Some(...) = ... && ..., but that
++        // theoretically these should be an `if let Some(...) = ... && ..., but that
+         // isn't stable
+         if let Some(path) = platform_config_dir {
+             if path.exists() {
+                 paths.push(path);
+             }
+         }
 +
-+        if let Some(path) = get_existing(legacy_platform_config_path) {
-+            paths.push(path);
++        if let Some(path) = legacy_platform_config_path {
++            if path.exists() {
++                paths.push(path);
++            }
 +        }
-+        if let Some(path) = get_existing(legacy_platform_config_dir) {
-+            paths.push(path);
++        if let Some(path) = legacy_platform_config_dir {
++            if path.exists() {
++                paths.push(path);
++            }
 +        }
 +
          paths
```
* 4:  b6c814370 = 4:  d1110e1b0 config: deprecate macOS legacy platform configs
</details>

cli/src/config.rs
```diff
@@ -163,26 +163,39 @@ impl AsMut<StackedConfig> for RawConfig {
...
+
+/// A ConfigPath can be in one of two states:
+///
+/// - exists(): a config file exists at the path
+/// - !exists(): a config file doesn't exist here, but a new file _can_ be
+///   created at this path
+#[derive(Clone, Debug)]
+pub struct ConfigPath {
+    path: PathBuf,
+    state: ConfigPathState,
```
yuja, line 188: Let's keep it private. I don't think users of `ConfigEnv` should be aware of `ConfigPath`, which is implementation detail.

cli/src/config.rs
```diff
@@ -231,18 +245,26 @@ impl UnresolvedConfigEnv {
...
             paths.push(path);
         }
-        if let Some(path @ Existing(_)) = platform_config_dir {
+
+        // theoretically this should be an `if let Some(...) = ... && ..., but that
+        // isn't stable
+        let get_existing = |opt: Option<ConfigPath>| match opt {
+            Some(path) if path.exists() => Some(path),
+            _ => None,
+        };
```
yuja: nit: maybe this can be moved to `ConfigPath::as_existing()` or something?

cli/src/config.rs
```diff
@@ -283,11 +281,30 @@ pub struct ConfigEnv {
...
+                .ok()
+                .map(|s| {
+                    use etcetera::BaseStrategy as _;
+                    // note that etcetera calls Library/Application Support the "data dir",
+                    // Library/Preferences is supposed to be exclusively plists
+                    s.data_dir()
+                })
+        } else {
+            etcetera::choose_base_strategy().ok().map(|s| {
+                use etcetera::BaseStrategy as _;
```
yuja: nit: just import globally unless we have conflict resolution.

cli/src/config.rs
```diff
@@ -333,16 +356,19 @@ impl ConfigEnv {
...
-        let env = UnresolvedConfigEnv {
+        let user_config_paths = UnresolvedConfigEnv {
             config_dir,
             macos_legacy_config_dir,
             home_dir: home_dir.clone(),
             jj_config: env::var("JJ_CONFIG").ok(),
-        };
+        }
+        .resolve();
+        Self::warn_for_deprecated_paths(ui, &user_config_paths);
```
yuja, line 366: nit: we can pass `ui` to `.resolve()`.

cli/src/config.rs
strega-nil: do I have to? I hate that for traits...

cli/src/config.rs
martinvonz, line 1: nit: the part about `config_paths()` in the commit message is stale now, right?

strega-nil:
I really hope that I've resolved all the nits now 😅

cli/src/config.rs
yuja: nit: this helper function appears to be deleted in the next commit. Just inline it?

cli/src/config.rs
```diff
@@ -1528,27 +1582,15 @@ mod tests {
...
-        let expected_existing: Vec<PathBuf> = case
-            .wants
-            .iter()
-            .filter_map(|want| match want {
-                Want::Existing(path) => Some(tmp.path().join(path)),
-                _ => None,
-            })
-            .collect();
-        assert_eq!(
-            env.existing_user_config_paths().collect_vec(),
```
yuja, line 1540: Test for `existing_user_config_paths()` appears deleted by mistake. Can you restore?

cli/src/config.rs
```diff
@@ -290,10 +312,9 @@ impl ConfigEnv {
     /// Returns the paths to the existing user-specific config files or
     /// directories.
     pub fn existing_user_config_paths(&self) -> impl Iterator<Item = &Path> {
-        self.user_config_paths.iter().filter_map(|p| match p {
-            ConfigPath::Existing(path) => Some(path.as_path()),
-            _ => None,
-        })
+        self.user_config_paths
+            .iter()
+            .filter_map(|p| if p.exists() { Some(p.as_path()) } else { None })
```
yuja, line 394: nit: `.filter(|p| p.exists()).map(|p| p.as_path())`

cli/src/config.rs
```diff
@@ -283,17 +284,46 @@ impl UnresolvedConfigEnv {
...
             }
         }
 
         paths
     }
+
+    fn warn_for_deprecated_path(ui: &Ui, path: &Path, old: &str, new: &str) {
+        let _ = writeln!(
+            ui.warning_default(),
+            "Deprecated configuration file `{}`:",
```
yuja, line 314: nit: maybe `.` instead of `:`?

Each warning line is printed separately with the `Warning: ` header. You can also use `indoc::writedoc!` to format multi-line text cleanly.

yuja (commented):
Looks good, thanks!

cli/src/config.rs
strega-nil: oops, seems like we had a rebase issue

cli/src/config.rs
strega-nil: is there a reason to do that instead of a `FilterMap`?

cli/src/config.rs
strega-nil: It wasn't really a mistake, but yeah I can do that.

strega-nil:
<details>
<summary>range-diff</summary>

* 1:  8d66617c5 ! 1:  ae2b0de24 config: change how config paths are represented
```diff
@@ cli/src/config.rs: impl UnresolvedConfigEnv {
              paths.push(path);
          }
 -        if let Some(path @ Existing(_)) = platform_config_dir {
+-            paths.push(path);
 +
 +        // theoretically this should be an `if let Some(...) = ... && ..., but that
 +        // isn't stable
-+        let get_existing = |opt: Option<ConfigPath>| match opt {
-+            Some(path) if path.exists() => Some(path),
-+            _ => None,
-+        };
-+        if let Some(path) = get_existing(platform_config_dir) {
-             paths.push(path);
++        if let Some(path) = platform_config_dir {
++            if path.exists() {
++                paths.push(path);
++            }
          }
          paths
+     }
 @@ cli/src/config.rs: impl ConfigEnv {
+ 
+     /// Returns the paths to the user-specific config files or directories.
+     pub fn user_config_paths(&self) -> impl Iterator<Item = &Path> {
+-        self.user_config_paths.iter().map(|p| p.as_path())
++        self.user_config_paths.iter().map(ConfigPath::as_path)
+     }
+ 
      /// Returns the paths to the existing user-specific config files or
      /// directories.
      pub fn existing_user_config_paths(&self) -> impl Iterator<Item = &Path> {
@@ cli/src/config.rs: impl ConfigEnv {
 -        })
 +        self.user_config_paths
 +            .iter()
-+            .filter_map(|p| if p.exists() { Some(p.as_path()) } else { None })
++            .filter(|p| p.exists())
++            .map(ConfigPath::as_path)
      }
  
      /// Returns user configuration files for modification. Instantiates one if
@@ cli/src/config.rs: mod tests {
 +            }
 +        }
 +
-+        fn matches(&self, root: &Path, path: &ConfigPath) -> bool {
-+            if root.join(self.path) != path.as_path() {
-+                return false;
-+            }
++        fn rooted_path(&self, root: &Path) -> PathBuf {
++            root.join(self.path)
++        }
 +
-+            match self.state {
-+                WantState::New => !path.exists(),
-+                WantState::Existing => path.exists(),
-+            }
++        fn exists(&self) -> bool {
++            matches!(self.state, WantState::Existing)
 +        }
      }
  
@@ cli/src/config.rs: mod tests {
          let env = resolve_config_env(&case.env, tmp.path());
  
 -        let expected_existing: Vec<PathBuf> = case
--            .wants
--            .iter()
++        let all_expected_paths = case
+             .wants
+             .iter()
 -            .filter_map(|want| match want {
 -                Want::Existing(path) => Some(tmp.path().join(path)),
 -                _ => None,
@@ cli/src/config.rs: mod tests {
 -            env.existing_user_config_paths().collect_vec(),
 -            expected_existing
 -        );
-+        let config_paths = env.user_config_paths.clone();
-+        let correct = case.wants.len() == config_paths.len()
-+            && case
-+                .wants
-+                .iter()
-+                .zip(&config_paths)
-+                .all(|(w, p)| w.matches(tmp.path(), p));
++            .map(|w| w.rooted_path(tmp.path()))
++            .collect_vec();
++        let exists_expected_paths = case
++            .wants
++            .iter()
++            .filter(|w| w.exists())
++            .map(|w| w.rooted_path(tmp.path()))
++            .collect_vec();
  
 -        let expected_paths: Vec<PathBuf> = case
 -            .wants
@@ cli/src/config.rs: mod tests {
 -            })
 -            .collect();
 -        assert_eq!(env.user_config_paths().collect_vec(), expected_paths);
-+        assert!(correct, "{:#?} != {:#?}", case.wants, config_paths);
++        let all_paths = env.user_config_paths().collect_vec();
++        let exists_paths = env.existing_user_config_paths().collect_vec();
++
++        assert_eq!(all_paths, all_expected_paths);
++        assert_eq!(exists_paths, exists_expected_paths);
      }
  
      fn setup_config_fs(files: &[&str]) -> tempfile::TempDir {
```
* 2:  c16bafb82 ! 2:  14091e809 config: switch to etcetera from dirs
  - this is just a move from this commit to the last commit 
* 3:  ac3f5e880 = 3:  ecd6c3c88 config: default to XDG config files on macOS
* 4:  41a110a5a ! 4:  03c3941ba config: deprecate macOS legacy platform configs
```diff
@@ cli/src/config.rs: impl UnresolvedConfigEnv {
      }
 +
 +    fn warn_for_deprecated_path(ui: &Ui, path: &Path, old: &str, new: &str) {
-+        let _ = writeln!(
++        let _ = indoc::writedoc!(
 +            ui.warning_default(),
-+            "Deprecated configuration file `{}`:",
-+            path.display()
-+        );
-+        let _ = writeln!(
-+            ui.warning_default(),
-+            "Configuration files in `{old}` are deprecated, and support will be removed in a \
-+             future release.",
-+        );
-+        let _ = writeln!(
-+            ui.warning_default(),
-+            "Instead, move your configuration files to `{new}`.",
++            r"
++            Deprecated configuration file `{}`.
++            Configuration files in `{old}` are deprecated, and support will be removed in a future release.
++            Instead, move your configuration files to `{new}`.
++            ",
++            path.display(),
 +        );
 +    }
  }
```
</details>

cli/src/config.rs
yuja: I think `.filter(p).map(f)` more directly expresses the intent. If there were a function that returns `Option<&Path>`, I would use `.filter_map()`.

cli/src/config.rs
```diff
@@ -349,13 +369,13 @@ impl ConfigEnv {
...
     pub fn repo_config_path(&self) -> Option<&Path> {
-        self.repo_config_path.as_ref().map(ConfigPath::as_path)
+        self.repo_config_path.as_ref().map(|p| p.as_path())
     }
 
     /// Returns a path to the existing repo-specific config file.
     fn existing_repo_config_path(&self) -> Option<&Path> {
-        match &self.repo_config_path {
-            Some(ConfigPath::Existing(path)) => Some(path),
+        match self.repo_config_path {
```
yuja, line 451: nit: `&self.repo_config_path` instead of `ref path` as before? (I don't have strong preference about that, though.)

cli/src/config.rs
```diff
@@ -1574,6 +1605,49 @@ mod tests {
...
+                macos_legacy_config_dir: Some("macos-legacy".into()),
+                ..Default::default()
+            },
+            wants: vec![
+                Want::new("config/jj/config.toml"),
+                Want::existing("macos-legacy/jj/config.toml"),
+            ],
+        }
+    }
+    fn config_path_macos_legacy_both_exist() -> TestCase {
```
yuja, line 1648: nit: insert blank line between functions

yuja (approved):

cli/src/config.rs
strega-nil: I don't really want to change this.

yuja: Fair. I just pointed out because you've changed the existing style.

Issue #5381, opened 2025-01-16
FR: rebase should detect identical commits


dotdash:
**Is your feature request related to a problem? Please describe.**
I have various old branches for which some commits have been merged to master, or rather, commits with identical changes but with different commit ids (e.g. because they were rebased or the commit message was edited in their PR phase). This is not a problem with git, because it detects identical commits during a rebase and skips them.
```
git rebase master
...
warning: skipped previously applied commit 40ca019a61
warning: skipped previously applied commit 8effd7177c
...
```

With `jj rebase -d master` these commits cause a bunch of conflicts, because they no longer apply cleanly on top of master.

**Describe the solution you'd like**
jj should do the same as git and detect duplicates and drop them (or maybe keep as empty unless `--skip-empty` is given) rather than creating conflicted changes.

**Describe alternatives you've considered**
Search manually for which changes are actually still needed, which is tedious and error prone.

bnjmnt4n:
Related issue: https://github.com/jj-vcs/jj/issues/2979.

scott2000:
*Edit: for reference, the below has been implemented, meaning that divergent commits can be abandoned when rebasing if they are duplicates of commits in the destination with the same change ID.*

---

When #6162 is merged and the change ID is preserved in many cases, maybe we could use the change IDs to skip duplicate commits without needing to compute patch IDs like Git does? I'm not sure if someone has already thought about this, but maybe something like this could work:

When rebasing a commit X onto parents Y, commit X is skipped if `should_skip_rebase(X, Y)` returns true:

```
fn should_skip_rebase(X, Y):
    S := set of visible commits with same change ID as X
    // Only check for duplicates if X is divergent
    if S contains any commits other than X:
        S' := S & (X..Y)
        for each commit Z in S':
            T := tree resulting from rebasing X onto the parents of Z
            if T is equivalent to the tree of Z:
                // X is a duplicate of Z, so it can be skipped
                return true
    return false
```

I think `S` would be efficient to compute due to the change ID index, and most rebased commits wouldn't be divergent so the potentially-expensive computations of the revset `S'` and rebased trees `T` could be skipped in most cases, so it likely would be fast enough even on large repos, but I'm not sure.

Issue #2979, opened 2024-02-07
FR: better detection of commits that were made empty


jyn514:
**Is your feature request related to a problem? Please describe.**
i have a local change that was merged into main as 52f4fb1b27da2874155062c0c3ccfa67cd0fe291. i still have that change locally, and jj does not think it is a parent of `main@origin` (cc https://github.com/martinvonz/jj/issues/2978). if i try to rebase it with `jj rebase -d main@origin unqytsx`, i get a merge conflict:
```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 4884a7383a...0000000000 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,12 +87,18 @@
 
 * `jj workspace root` was aliased to `jj root`, for ease of discoverability
 
+<<<<<<<
++++++++
 * `jj diff` no longer shows the contents of binary files.
 
 * `jj git` now has an `init` command that initializes a git backed repo.
 
 * New template function `surround(prefix, suffix, content)` is added.
 
+%%%%%%%
++* `jj diff` no longer shows the contents of binary files.
++
+>>>>>>>
```
this is very silly; the line it's trying to add is already present in the file, and it was added before the "jj git" line (before in *history*, the fact it's the line above shouldn't be relevant). so jj should notice that it's empty instead of thinking it's a conflict.

**Describe the solution you'd like**
`jj rebase` should notice the change is now empty and i should end up with a commit with the same description as `unqytsx` but an empty diff. ideally that change itself should then be dropped (https://github.com/martinvonz/jj/issues/229), but that's separate from this issue.

**Describe alternatives you've considered**

**Additional context**
i am pretty sure this "just works" with git, although i admittedly haven't tested.

martinvonz:
I think you're right that git does this better. It calculates a "patch id" (basically a hash of a context-less diff output, IIRC) for the commits to rebase and for the upstream commits. Then it skips rebasing the commits that match a patch in upstream. I have been thinking of doing something similar in the planned `jj git sync`. Maybe we should add an option for that to `jj rebase` too.

jyn514:
sounds great :)

> I have been thinking of doing something similar in the planned jj git sync. Maybe we should add an option for that to jj rebase too.

[...]

PR #6302, opened 2025-04-10
rebase: skip duplicate divergent commits by default

scott2000:
This is a partial fix for #5381 and #2979. It doesn't handle the general case (it only detects duplicate changes between divergent commits with the same change ID), but I think it should handle many useful cases, especially for users with the `change-id` header enabled. I'm looking for feedback about whether this is a good general approach and about the implementation.

This serves a similar purpose to Git's patch ID mechanism, however it is slightly different in that it only compares commits which have the same change ID as each other (divergent changes), and it does a full comparison of the commits to see if they would have identical trees if rebased onto the same parents. Since most changes aren't divergent, I believe this should have a negligible performance cost in most cases.

I think skipping these commits by default makes sense for `jj rebase`, since usually this will be a helpful behavior for the user. With this behavior, a safe first step when encountering divergent changes would be to rebase one branch on top of the other, since that will abandon any divergent changes that have identical contents to existing commits, leaving behind any non-trivial divergent changes for the user to resolve manually.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/rebase.rs
+++ b/cli/src/commands/rebase.rs
@@ -24,6 +24,7 @@ use jj_lib::object_id::ObjectId as _;
 use jj_lib::repo::ReadonlyRepo;
 use jj_lib::repo::Repo as _;
 use jj_lib::revset::RevsetExpression;
+use jj_lib::rewrite::find_duplicate_divergent_commits;
 use jj_lib::rewrite::move_commits;
 use jj_lib::rewrite::EmptyBehaviour;
 use jj_lib::rewrite::MoveCommitsLocation;
@@ -324,6 +325,14 @@ pub(crate) struct RebaseArgs {
     /// parents.
     #[arg(long)]
     skip_emptied: bool,
+
+    /// Keep divergent commits while rebasing
+    ///
+    /// Without this flag, divergent commits are abandoned while rebasing if
+    /// another commit with the same change ID is already present in the
+    /// destination with identical changes.
+    #[arg(long)]
+    keep_divergent: bool,
 }
 
 #[derive(clap::Args, Clone, Debug)]
@@ -394,6 +403,20 @@ pub(crate) fn cmd_rebase(
     };
 
     let mut tx = workspace_command.start_transaction();
+    if !args.keep_divergent {
+        let abandoned_divergent =
+            find_duplicate_divergent_commits(tx.repo(), &loc.new_parent_ids, &loc.target)?;
+        for commit in &abandoned_divergent {
+            tx.repo_mut().record_abandoned_commit(commit);
+        }
+        if !abandoned_divergent.is_empty() {
+            writeln!(
+                ui.status(),
+                "Skipped {} divergent commits that were already present in the destination",
+                abandoned_divergent.len(),
+            )?;
+        }
+    };
     let stats = move_commits(tx.repo_mut(), &loc, &rebase_options)?;
     print_move_commits_stats(ui, &stats)?;
     tx.finish(ui, tx_description(&loc.target))?;
--- a/cli/tests/test_rebase_command.rs
+++ b/cli/tests/test_rebase_command.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::common::create_commit;
+use crate::common::create_commit_with_files;
 use crate::common::CommandOutput;
 use crate::common::TestEnvironment;
 use crate::common::TestWorkDir;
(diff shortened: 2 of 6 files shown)


lib/src/rewrite.rs
```diff
@@ -278,10 +284,59 @@ impl<'repo> CommitRewriter<'repo> {
...
+
+        // We only care about commits which previously weren't ancestors of the rebased
+        // commit, but will become ancestors after the rebase.
+        let candidate_duplicates = ResolvedRevsetExpression::commits(candidate_duplicates)
+            .intersection(
+                &ResolvedRevsetExpression::commit(self.old_commit.id().clone())
+                    .range(&ResolvedRevsetExpression::commits(self.new_parents.clone())),
+            )
+            .evaluate(self.mut_repo)
+            .map_err(|err| BackendError::Other(Box::new(err)))?;
```
yuja, line 320: Since we're only interested in ancestors of the new parents (which I think are equivalent to the ancestors of the rebase destination), maybe we can abandon divergent commits before starting the rebase process?

scott2000: I tried to rewrite it with this new approach by finding a list of commits to abandon in `commands/rebase.rs`, and then passing it to `move_commits()` to do the actual abandoning. I'm not sure if there's a better way to do that.

yuja: I think commits can be abandoned separately (before calling `move_commits()`), but I haven't tested. `move_commits()` might have some assumption that `mut_repo.parent_mapping` is empty.

BTW, I think it's better to add this function to jj-lib. We might want to use something similar to abandon old commits and/or generate fake predecessors on `jj git fetch`.

scott2000:
I think this is ready for review now; I updated it to abandon the commits before starting the rebase and added tests.

cli/tests/test_rebase_command.rs
martinvonz, line 1: I'm sorry but could you migrate most of these tests to `jj-lib`? Tests there are much faster to run.

lib/src/rewrite.rs
```diff
@@ -1128,3 +1129,133 @@ pub fn squash_commits<'repo>(
...
+        for targets in change_id_targets.values() {
+            // Checking every pair of commits between these two sets could be expensive if
+            // there are several commits with the same change ID. However, it should be
+            // uncommon to have more than a couple commits with the same change ID being
+            // rebased at the same time, so it should be good enough in practice.
+            for &rebased_commit in &targets.to_rebase {
+                // To maintain the graph structure, we don't abandon any merge commits.
+                if rebased_commit.parent_ids().len() != 1 {
+                    continue;
+                }
```
martinvonz, line 1189: Is this actually necessary? The structure should be maintained even if you abandon a merge. Maybe this is a holdover from a previous version of the code?

lib/src/rewrite.rs
```diff
@@ -1128,3 +1129,133 @@ pub fn squash_commits<'repo>(
...
+        MoveCommitsTarget::Commits(commits) => commits,
+    };
+    let target_commit_root_ids = target_commit_roots
+        .iter()
+        .map(Commit::id)
+        .cloned()
+        .collect_vec();
+
+    let mut abandoned_divergent = Vec::new();
+    {
```
martinvonz, line 1171: nit: drop this scope?

martinvonz (commented):
I'm not sure how this will affect rebase performance when using the Google backends. I'll try to import this into the Google repo quickly once it has been merged, so we can test it and see. We have a custom `jj piper sync` command that already has its own way of skipping "merged" commits, and that won't be affected by this PR. Just letting you know (no action needed).

lib/src/rewrite.rs
scott2000: Yeah, I based it on the skip-emptied code, and I saw that there was an assertion that skipped commits during rebase always have exactly 1 parent, so I thought it was required. But now that the commits are abandoned before rebasing, it's probably not necessary. Or maybe that assertion wasn't even required in the first place.

lib/src/rewrite.rs
scott2000: This scope is required to ensure that the `is_new_ancestor` function is dropped before abandoning the commits, since it borrows the `MutableRepo`. This was the cleanest solution I could think of, but maybe there's a better way? A `drop(is_new_ancestor)` before abandoning the commits also works.

scott2000:
@yuja Your other PR made the code a lot simpler, thanks! I haven't had much time recently, but I'm hoping to address the comments soon.

cli/tests/test_rebase_command.rs
scott2000: I migrated the tests to `jj-lib`, but I left one smaller test in `jj-cli` to test that `--keep-divergent` works properly for `jj rebase`. If this isn't necessary, I can remove that one as well.

lib/src/rewrite.rs
```diff
@@ -1128,3 +1129,133 @@ pub fn squash_commits<'repo>(
...
+            .range(&RevsetExpression::commits(new_parent_ids.to_owned()))
+            .evaluate(mut_repo)
+            .map_err(|err| err.into_backend_error())?
+            .containing_fn();
+
+        for targets in change_id_targets.values() {
+            // Checking every pair of commits between these two sets could be expensive if
+            // there are several commits with the same change ID. However, it should be
+            // uncommon to have more than a couple commits with the same change ID being
+            // rebased at the same time, so it should be good enough in practice.
```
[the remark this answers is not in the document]
scott2000: I'm not sure what you meant by this comment exactly (sorry it took me so long to respond). Are you suggesting to do first do the `is_new_ancestor` check for all of the candidates and build a map with a filtered list, and then afterwards iterate over the map? I tried doing that, but it's a bit difficult because the revset containing fn returns a `Result<bool, _>`, so I can't just use `Vec::retain` or `Iterator::filter`. I'm not sure if there's some better way to do that though.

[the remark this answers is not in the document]
yuja: We currently build `Change: {to_rebase: Vec<Commit>, not_rebased: Vec<Commit>}` mapping, but I think this can be simplified to `Commit: Vec<Commit>` because `Change == [to_rebase]` usually holds, and we'll loop over all `(to_rebase, not_rebased)` combinations anyway.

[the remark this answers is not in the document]
scott2000: Yeah, that does make it a lot simpler, thanks!

cli/src/commands/rebase.rs
```diff
@@ -394,6 +403,20 @@ pub(crate) fn cmd_rebase(
     };
 
     let mut tx = workspace_command.start_transaction();
+    if !args.keep_divergent {
+        let abandoned_divergent =
+            find_duplicate_divergent_commits(tx.repo_mut(), &loc.new_parent_ids, &loc.target)?;
```
yuja, line 408: nit: `tx.repo()` (no `_mut`)

cli/src/commands/rebase.rs
```diff
@@ -394,6 +403,20 @@ pub(crate) fn cmd_rebase(
...
+        for commit in &abandoned_divergent {
+            tx.repo_mut().record_abandoned_commit(commit);
+        }
+        if !abandoned_divergent.is_empty() {
+            writeln!(
+                ui.status(),
+                "Skipped {} divergent commits that were already present in the destination",
+                abandoned_divergent.len(),
+            )?;
```
yuja, line 417: nit: might be better to list abandoned commits as `jj abandon` would do, but we don't do that for `--skip-emptied`, so I don't know which would be better.

lib/src/rewrite.rs
```diff
@@ -1180,3 +1180,82 @@ pub fn squash_commits<'repo>(
...
+            .evaluate(repo)
+            .map_err(|err| err.into_backend_error())?
+            .iter()
+            .commits(repo.store())
+            .try_collect()
+            .map_err(|err| err.into_backend_error())?,
+        MoveCommitsTarget::Commits(commit_ids) => commit_ids
+            .iter()
+            .map(|commit_id| repo.store().get_commit(commit_id))
+            .try_collect()?,
```
yuja, line 1204: nit: swap these to match the definition order of `enum MoveCommitsTarget`?

lib/src/rewrite.rs
```diff
@@ -1180,3 +1180,82 @@ pub fn squash_commits<'repo>(
...
+        .filter(|(_, candidates)| !candidates.is_empty())
+        .collect();
+    if divergent_changes.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    let rebase_root_commit_ids = match target {
+        MoveCommitsTarget::Roots(root_ids) => root_ids,
+        MoveCommitsTarget::Commits(commit_ids) => commit_ids,
+    };
```
yuja, line 1228: nit: same here. `Commits(_)`, `Roots(_)` in that order

lib/src/rewrite.rs
```diff
@@ -1180,3 +1180,82 @@ pub fn squash_commits<'repo>(
...
+        .evaluate(repo)
+        .map_err(|err| err.into_backend_error())?
+        .containing_fn();
+
+    let mut duplicate_divergent = Vec::new();
+    // Checking every pair of commits between these two sets could be expensive if
+    // there are several commits with the same change ID. However, it should be
+    // uncommon to have more than a couple commits with the same change ID being
+    // rebased at the same time, so it should be good enough in practice.
+    for (rebased_commit, ancestor_candidates) in divergent_changes {
```
yuja, line 1243: nit: `rebased_commit` in this context might be confusing because it is the commit "to be" rebased. Maybe we can use `target_commits`, `target_roots`, etc. as we call that way in `move_commits()`?

lib/tests/test_rewrite.rs
```diff
@@ -1889,3 +1893,146 @@ fn test_commit_with_selection() {
...
+        tx.repo_mut()
+            .new_commit(parents, tree_id)
+            .set_change_id(change_id)
+            .write()
+            .unwrap()
+    };
+
+    let commit_a1 = make_commit(
+        tree_a1.id(),
+        ChangeId::from_hex("aaaa"),
```
yuja, line 1957: nit: probably better to generate a `ChangeId` of valid length because it is stored in the test backend.

lib/tests/test_rewrite.rs
```diff
@@ -1889,3 +1893,146 @@ fn test_commit_with_selection() {
...
+        vec![commit_c2.id().clone()],
+    );
+
+    #[track_caller]
+    fn assert_same_ids(commits: &[Commit], ids: &[&CommitId]) {
+        assert_eq!(
+            commits.iter().map(Commit::id).sorted().collect_vec(),
+            ids.iter().sorted().copied().collect_vec(),
+        );
+    }
```
yuja, line 2002: nit: maybe we can just `assert_eq!(&[Commit], &[Commit])`? I think the order of the returned list is deterministic.

yuja (approved):
Looks good, thanks!

cli/src/commands/rebase.rs
scott2000: Yeah, that might be good. I'll make a separate PR for it in case there's more discussion.

scott2000:
Hmm I just tested it after merging, and I found a bug immediately. Apparently `rewrite::move_commits` doesn't work when the root of the set of target commits has already been abandoned. In this case, it just leaves the commits in place. It does work correctly when the abandoned commit isn't the root of the set of target commits. @yuja and @martinvonz do you have any ideas for how to fix this? I can make another PR to revert this if it's not a simple fix.

PR #6303, opened 2025-04-10
cli: duplicate: add setting for templating the new commit descriptions

winterqt:
This allows the customization of the duplicated commit descriptions.

An ideal use case for this is emulating `git cherry-pick -x`, as illustrated in the tests.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/duplicate.rs
+++ b/cli/src/commands/duplicate.rs
@@ -14,8 +14,10 @@
 
 use std::io::Write as _;
 
+use bstr::ByteVec as _;
 use clap_complete::ArgValueCandidates;
 use itertools::Itertools as _;
+use jj_lib::backend::BackendResult;
 use jj_lib::backend::CommitId;
 use jj_lib::repo::Repo as _;
 use jj_lib::rewrite::duplicate_commits;
@@ -30,6 +32,7 @@ use crate::cli_util::RevisionArg;
 use crate::command_error::user_error;
 use crate::command_error::CommandError;
 use crate::complete;
+use crate::formatter::PlainTextFormatter;
 use crate::ui::Ui;
 
 /// Create new changes with the same content as existing ones
@@ -45,6 +48,9 @@ use crate::ui::Ui;
 /// `--insert-after` or `--insert-before` arguments are provided, the new
 /// children indicated by the arguments will be rebased onto the heads of the
 /// specified commits.
+///
+/// By default, the duplicated commits retain the descriptions of the originals.
+/// This can be customized with the `templates.duplicate_description` setting.
 #[derive(clap::Args, Clone, Debug)]
 pub(crate) struct DuplicateArgs {
     /// The revision(s) to duplicate (default: @)
@@ -164,6 +170,27 @@ pub(crate) fn cmd_duplicate(
             }
         }
     }
+
+    let new_descs = {
+        let template = tx
+            .settings()
+            .get_string("templates.duplicate_description")?;
+        let parsed = tx.parse_commit_template(ui, &template)?;
+
+        to_duplicate
+            .iter()
+            .map(|commit_id| -> BackendResult<_> {
+                let mut output = Vec::new();
+                let commit = tx.repo().store().get_commit(commit_id)?;
+                parsed
+                    .format(&commit, &mut PlainTextFormatter::new(&mut output))
+                    .expect("write() to vec backed formatter should never fail");
+
+                Ok((commit_id.clone(), output.into_string_lossy()))
+            })
+            .try_collect()?
+    };
+
     let num_to_duplicate = to_duplicate.len();
     let DuplicateCommitsStats {
         duplicated_commits,
@@ -172,11 +199,12 @@ pub(crate) fn cmd_duplicate(
         duplicate_commits(
             tx.repo_mut(),
             &to_duplicate,
+            &new_descs,
             &parent_commit_ids,
             &children_commit_ids,
         )?
     } else {
-        duplicate_commits_onto_parents(tx.repo_mut(), &to_duplicate)?
+        duplicate_commits_onto_parents(tx.repo_mut(), &to_duplicate, &new_descs)?
     };
 
     if let Some(mut formatter) = ui.status_formatter() {
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -35,6 +35,8 @@ config_list = 'builtin_config_list'
 
 draft_commit_description = 'builtin_draft_commit_description'
 
+duplicate_description = 'description'
+
 commit_trailers = ''
 
 file_list = '''
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -802,6 +802,8 @@ When none of the `--destination`, `--insert-after`, or `--insert-before` argumen
 
 When any of the `--destination`, `--insert-after`, or `--insert-before` arguments are provided, the roots of the specified commits will be duplicated onto the destination indicated by the arguments. Other specified commits will be duplicated onto these newly duplicated commits. If the `--insert-after` or `--insert-before` arguments are provided, the new children indicated by the arguments will be rebased onto the heads of the specified commits.
 
+By default, the duplicated commits retain the descriptions of the originals. This can be customized with the `templates.duplicate_description` setting.
+
 **Usage:** `jj duplicate [OPTIONS] [REVSETS]...`
 
 ###### **Arguments:**
--- a/cli/tests/test_duplicate_command.rs
+++ b/cli/tests/test_duplicate_command.rs
@@ -2427,6 +2427,99 @@ fn test_rebase_duplicates() {
     ");
 }
 
+#[test]
+fn test_duplicate_description_template() {
+    let test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let work_dir = test_env.work_dir("repo");
+
+    create_commit(&work_dir, "a", &[]);
+    create_commit(&work_dir, "b", &["a"]);
+    create_commit(&work_dir, "c", &["b"]);
+
+    // Test the setup
+    insta::assert_snapshot!(get_log_output(&work_dir), @r"
+    @  7e4fbf4f2759   c
+    ○  1394f625cbbd   b
+    ○  2443ea76b0b1   a
+    ◆  000000000000
+    [EOF]
+    ");
+
+    // Test duplicate_commits()
+    test_env.add_config(r#"templates.duplicate_description = "concat(description, '\n(cherry picked from commit ', commit_id, ')')""#);
+    let output = work_dir.run_jj(["duplicate", "a"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Duplicated 2443ea76b0b1 as yostqsxw 20d8e008 a
+    [EOF]
+    ");
+
+    // Test duplicate_commits_onto_parents()
+    let output = work_dir.run_jj(["duplicate", "a", "-B", "b"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Warning: Duplicating commit 2443ea76b0b1 as a descendant of itself
+    Duplicated 2443ea76b0b1 as znkkpsqq 14ab622d (empty) a
+    Rebased 2 commits onto duplicated commits
+    Working copy  (@) now at: royxmykx 7f7870c2 c | c
+    Parent commit (@-)      : zsuskuln c92a59b0 b | b
+    [EOF]
+    ");
+
+    // Test empty template
+    test_env.add_config("templates.duplicate_description = ''");
+    let output = work_dir.run_jj(["duplicate", "b", "-d", "root()"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Duplicated c92a59b0d9e6 as kpqxywon d32164b2 (no description set)
+    [EOF]
+    ");
+
+    // Test `description` as an alias
+    test_env.add_config("templates.duplicate_description = 'description'");
+    let output = work_dir.run_jj([
+        "duplicate",
+        "c",
+        "-d",
+        "root()",
+        // Use an argument here so we can actually see the log in the last test
+        // (We don't have a way to remove a config in TestEnvironment)
+        "--config",
+        "template-aliases.description='\"alias\"'",
+    ]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Duplicated 7f7870c2c58d as kmkuslsw d107a339 alias
+    [EOF]
+    ");
+
+    let template = r#"commit_id.short() ++ "\n" ++ description"#;
+    let output = work_dir.run_jj(["log", "-T", template]);
+    insta::assert_snapshot!(output, @r"
+    @  7f7870c2c58d
+    │  c
+    ○  c92a59b0d9e6
+    │  b
+    ○  14ab622d51b3
+    │  a
+    │
+    │  (cherry picked from commit 2443ea76b0b1c531326908326aab7020abab8e6c)
+    ○  2443ea76b0b1
+    │  a
+    │ ○  d107a3395c1c
+    ├─╯  alias
+    │ ○  d32164b275c2
+    ├─╯
+    │ ○  20d8e008cc0f
+    ├─╯  a
+    │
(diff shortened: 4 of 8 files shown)


winterqt:
Could probably use a mention in the Git command table at the very least (if we decide to keep this as a flag and not a setting a la `templates.draft_commit_description`).

cli/src/commands/duplicate.rs
```diff
@@ -179,6 +199,48 @@ pub(crate) fn cmd_duplicate(
...
+                .map(|c| c.id().clone())
+                .collect(),
+            |rewriter| {
+                let old_dupe_commit_id = rewriter.old_commit().id().clone();
+                let commit_builder = rewriter.reparent();
+                if let Some((desc, old_commit_id)) = new_descs.remove(&old_dupe_commit_id) {
+                    *duplicated_commits
+                        .get_mut(&old_commit_id)
+                        .expect("duplicated_commits should contain the original commit id") =
+                        commit_builder.set_description(desc).write()?;
```
winterqt, line 234: Perhaps we should copy [`jj describe`'s new behavior](https://github.com/jj-vcs/jj/blob/5dd64ae978a20851151cddfe6b6421d929e70687/cli/src/commands/describe.rs#L264-L265) and set the commit timestamp to the old one? Sure, it'll be... a millisecond or two off, but... maybe it makes sense?

winterqt (commented):
Excluding the efficiency of writing 2x the commits with this option enabled, the main downside of this method is that the evolog looks ugly with the extra commit (IMO):

```
    ○  yostqsxw test.user@example.com 2001-02-03 08:05:15 4037fb05
    │  a
    ○  yostqsxw hidden test.user@example.com 2001-02-03 08:05:15 3d341b2f
    │  a
    ○  rlvkpnrz test.user@example.com 2001-02-03 08:05:09 a 2443ea76
    │  a
    ○  rlvkpnrz hidden test.user@example.com 2001-02-03 08:05:08 d8d5f980
       (empty) a
```

However, the op log looks good. :)

```
    @  44a208b90324 test-username@host.example.com 2001-02-03 04:05:15.000 +07:00 - 2001-02-03 04:05:15.000 +07:00
    │  duplicate 1 commit(s)
    │  args: jj duplicate a -T 'concat(description, "\n", "(cherry-picked from ", commit_id, ")")'
    │
    │  Changed commits:
    │  ○  + yostqsxw 4037fb05 a
```

Probably not that big of a deal, though; our working copy snapshots put more "wear" on the repo than this ever could.

cli/tests/test_duplicate_command.rs
```diff
@@ -2427,6 +2427,62 @@ fn test_rebase_duplicates() {
...
+    @    17a00fc21654   c
+    ├─╮
+    │ ○  d370aee184ba   b
+    ○ │  2443ea76b0b1   a
+    ├─╯
+    ◆  000000000000
+    [EOF]
+    ");
+
+    let template = r#"concat(description, "\n", "(cherry-picked from ", commit_id, ")")"#;
```
emilazy, line 2451:
```suggestion
    let template = r#"concat(description, "\n", "(cherry picked from ", commit_id, ")")"#;
```
to match Git.

cli/src/commands/duplicate.rs
yuja, line 234: > the main downside of this method is that the evolog looks ugly with the extra commit

Perhaps, you can fix up `predecessors` to point to the original commit.

A better option is to split `duplicate_commits()` into two parts.

yuja (commented):
The implementation generally looks good to me, thanks.

cli/src/commands/duplicate.rs
winterqt: Yeah, it seems like splitting it may be the best bet. I tried to do some things with callbacks, to no avail, and I was thinking about splitting them, but then I need to think about how exactly to split the functions. I'll figure it out. :)

cli/src/config/templates.toml
```diff
@@ -66,6 +66,8 @@ separate(" ",
 )
 '''
 
+duplicate_commit_description = ''
```
winterqt, line 69: We probably don't want to add a default, but perhaps an alias called `cherry_pick_commit_description` or something would be good; then we can reference it in the Git command table.

winterqt: Hm, should this be named `duplicate_description` instead, like `revert_description`?

lib/src/rewrite.rs
```diff
@@ -776,6 +779,7 @@ pub struct DuplicateCommitsStats {
 pub fn duplicate_commits(
     mut_repo: &mut MutableRepo,
     target_commits: &[CommitId],
+    mut target_descriptions: HashMap<CommitId, String>,
```
yuja, line 782: nit: pass by reference? I don't think the extra `.clone()` cost would matter.

cli/src/commands/duplicate.rs
yuja: Right. Since the caller knows the commits to be duplicated, we don't have to split functions. (I thought there would be something that had to be computed.)

lib/src/rewrite.rs
winterqt: Sure, I suppose.

winterqt:
Strange... it builds locally. I think Cargo is being dumb again. :/

cli/src/commands/duplicate.rs
```diff
@@ -164,6 +170,26 @@ pub(crate) fn cmd_duplicate(
             }
         }
     }
+    let mut new_descs = HashMap::new();
+    if let Some(template) = tx
+        .settings()
+        .get_string("templates.duplicate_description")
+        .ok()
```
yuja, line 177: nit: `.get_string(..)?` to propagate error

cli/src/commands/duplicate.rs
```diff
@@ -164,6 +170,26 @@ pub(crate) fn cmd_duplicate(
...
+        .filter(|t| !String::is_empty(t))
```
yuja, line 178: nit: `!t.is_empty()` or `t != "description"` (see the other comment)

cli/src/config/templates.toml
```diff
@@ -66,6 +66,8 @@ separate(" ",
 )
 '''
 
+duplicate_description = ''
```
yuja, line 69: Perhaps, the default should be `'description'`. The behavior shouldn't change depending on the optimization for the default template.

Can you also move this after `draft_commit_description`? Appears that the templates are sorted lexicographically.

lib/src/rewrite.rs
```diff
@@ -766,8 +766,11 @@ pub struct DuplicateCommitsStats {
 /// The roots of `target_commits` are duplicated on top of the new
 /// `parent_commit_ids`, whilst other commits in `target_commits` are duplicated
 /// on top of the newly duplicated commits in the target set. If
-/// `children_commit_ids` is not empty, the `children_commit_ids` will be
-/// rebased onto the heads of the duplicated target commits.
+/// `target_descriptions` is not empty, it will be consulted to retrieve the new
+/// descriptions of the target commits, falling back to the original if the map
+/// does not contain an entry for a given commit. If `children_commit_ids` is
```
yuja, line 771: nit: I would split this to new paragraph. Here we're discussing about the graph structure.

cli/tests/test_duplicate_command.rs
```diff
@@ -2427,6 +2427,62 @@ fn test_rebase_duplicates() {
...
+    ├─╮
+    │ ○  d370aee184ba   b
+    ○ │  2443ea76b0b1   a
+    ├─╯
+    ◆  000000000000
+    [EOF]
+    ");
+
+    test_env.add_config(r#"templates.duplicate_description = "concat(description, '\n(cherry-picked from commit ', commit_id, ')')""#);
+    let output = work_dir.run_jj(["duplicate", "a"]);
```
yuja, line 2452: Can you add test for `duplicate_commits()`?

yuja (approved):
Thanks.

cli/src/commands/duplicate.rs
winterqt: This was a leftover from when I didn't put it in the config by default, thanks.

cli/src/config/templates.toml
winterqt: > Perhaps, the default should be `'description'`. The behavior shouldn't change depending on the optimization for the default template.

Not sure what you mean -- the exact behavior wouldn't change at all due to the `is_empty` check. We're just changing the optimized case.

Also, to clarify: no matter if the template is `'description'` (optimized out) or `' description '`, the resulting commits will be identical.

cli/tests/test_duplicate_command.rs
winterqt: Let me know if this is sufficient.

lib/src/rewrite.rs
```diff
@@ -776,6 +780,7 @@ pub struct DuplicateCommitsStats {
 pub fn duplicate_commits(
     mut_repo: &mut MutableRepo,
     target_commits: &[CommitId],
+    target_descriptions: &HashMap<CommitId, String>,
```
winterqt, line 783: I chose not to make this `Into<String>` because you then have to use turbofishes for `HashMap::new()` (that is, the empty case).

cli/tests/cli-reference@.md.snap
```diff
@@ -802,6 +802,8 @@ When none of the `--destination`, `--insert-after`, or `--insert-before` argumen
 
 When any of the `--destination`, `--insert-after`, or `--insert-before` arguments are provided, the roots of the specified commits will be duplicated onto the destination indicated by the arguments. Other specified commits will be duplicated onto these newly duplicated commits. If the `--insert-after` or `--insert-before` arguments are provided, the new children indicated by the arguments will be rebased onto the heads of the specified commits.
 
+The descriptions of the duplicated commits can be customized with the `templates.duplicate_description`.
```
winterqt, line 805: Maybe
```suggestion
By default, the duplicated commits retain the description of the originals. This can be customized with the `duplicate_description` template.
```

cli/src/config/templates.toml
yuja: > Not sure what you mean -- the exact behavior wouldn't change at all due to the `is_empty` check.

I mean the behavior shouldn't change if we removed the `if`. It should just exist for optimization purpose. Since an empty string is a valid template expression, unfortunately, it should be evaluated to `""`, not to `description`.

FWIW, if user defined `description` alias, the result would go awry. It might be safer to just always render the template.

cli/tests/test_duplicate_command.rs
yuja: I meant to add CLI test for both `duplicate_commits()` and `duplicate_commits_onto_parents()`, but adding lib tests seems also good.

cli/src/config/templates.toml
winterqt: Good point, thanks!

cli/tests/test_duplicate_command.rs
winterqt: What do you mean add CLI tests for them? `duplicate_commits()` is tested by the case I added (via `jj duplicate`), and I can add one for `duplicate_commits_onto_parents()` if you want.

It seems like you're talking about something different though -- could you please elaborate?

yuja: `jj duplicate` (with no destination arguments) doesn't exercise the `duplicate_commits()` code path, right? For historical reasons, `jj duplicate` has two different behaviors.

winterqt: Ah, right, so you meant that I currently test `duplicate_commits_onto_parents()`, but not `duplicate_parents()`. Yup, can fix. Sorry for the miscommunication, I thought you were talking about something else entirely.

cli/tests/cli-reference@.md.snap
yuja: Looks good, but I wouldn't remove `templates.` prefix from the config name.

winterqt:
Alright, this *should* be good to go now; would appreciate one more review pass. Thanks for the help!

cli/tests/cli-reference@.md.snap
winterqt: https://jj-vcs.github.io/jj/latest/config/#default-description does, which is why I did.

Could do
```suggestion
By default, the duplicated commits retain the description of the originals. This can be customized with `templates.duplicate_description`.
```
but I would like them to be consistent.

cli/src/commands/duplicate.rs
```diff
@@ -164,6 +170,26 @@ pub(crate) fn cmd_duplicate(
...
+        for commit_id in &to_duplicate {
+            let mut output = Vec::new();
+            let commit = tx.repo().store().get_commit(commit_id)?;
+            parsed
+                .format(&commit, &mut PlainTextFormatter::new(&mut output))
+                .expect("write() to vec backed formatter should never fail");
+
+            new_descs.insert(commit_id.clone(), output.into_string_lossy());
+        }
+    }
```
yuja, line 191: nit: (at your option), now we can remove `mut`, but I don't have preference over these.
```rust
let new_descs: HashMap<_, _> = {
    ...
    to_duplicate.iter().map(|commit_id| -> BackendResult<_> { .. }).try_collect()?
};

File: cli/tests/cli-reference@.md.snap (hunk 11 above)
Comment (yuja) [reply cli/tests/cli-reference@.md.snap:805]: In the command help context, it's not obvious where to put the `duplicate_description` config value. That's why I prefer fully-qualified name here.

**Review by yuja — APPROVED** [review]

File: cli/tests/test_duplicate_command.rs [hunk 13]
```diff
@@ -2427,6 +2427,96 @@ fn test_rebase_duplicates() {
...
+        a
+
+        (cherry-picked from commit 2443ea76b0b1c531326908326aab7020abab8e6c)
+
+    A a
+    [EOF]
+    ");
+
+    // Test duplicate_commits_onto_parents()
+    let output = work_dir.run_jj(["duplicate", "7e", "-B", "b8"]);
```
yuja, line 2482: nit: better to use bookmark name? you'll have to update hashes if something added to tests.

cli/src/commands/duplicate.rs
winterqt: I'll think about it -- couldn't we do this before, though?

cli/tests/test_duplicate_command.rs
winterqt: Would it really be useful to use the bookmark names for some of the commits, but not others? (Only the ones from `create_commit` have the bookmarks set, and not the ones created by `jj duplicate`.

(I also think I ran into some issues when using the bookmark names for `create_commit` commits, but I can try again if you think it's worth it.)

cli/src/commands/duplicate.rs
yuja: > couldn't we do this before, though?

You could, but we would need something like `new_descs = if { ... } else { HashMap::new() };` to remove `mut`.

cli/tests/test_duplicate_command.rs
winterqt: Sure, but we would still have to update the `jj show` commands for the newly duplicated commits. Do you think there's a better way to refer to those too?

yuja: Maybe dump all commits at the end (by `jj log -T..`)?

btw, when we have to use hashes, I would suggest using 8-12 characters. The shortest hash is harder to spot, and has higher chance that it points to different commit by accident.

cli/tests/test_duplicate_command.rs
```diff
@@ -2427,6 +2427,118 @@ fn test_rebase_duplicates() {
...
+        "template-aliases.description='\"alias\"'",
+    ]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Duplicated 3fb5c9d2839f as lylxulpl e6877e1e alias
+    [EOF]
+    ");
+
+    let template = r#"commit_id.short() ++ "\n" ++ description"#;
+    let output = work_dir.run_jj(["log", "-T", template]);
```
winterqt, line 2517: @yuja Do you think it's worthwhile to run `jj log` at every other commit if we're doing a big one at the end? Probably would be best to remove the intermediate ones.

yuja: Yeah, I would also remove intermediate `jj log` calls so long as the result is obvious.

winterqt: > so long as the result is obvious

Do you think that's true in this case? :)

PR #6304, opened 2025-04-10
revset: add `signed()` function

### LINKED ISSUE #5786: Don't resign already signed commits in `jj sign`
Right now `jj sign` always resigns commits, even if they are already signed by the user. This is cumbersome for users, which sign via hardware devices.

# References
* https://github.com/jj-vcs/jj/pull/4747#discussion_r1966693979)
* #4747

nilskch:
Related to #5786

This PR adds a new `signed()` revset function that filters for signed commits.

cc @yuja 

# Checklist

If applicable:

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -304,6 +304,8 @@ revsets (expressions) as arguments.
 * `committer_date(pattern)`: Commits with committer dates matching the specified
   [date pattern](#date-patterns).
 
+* `signed()`: Commits that are cryptographically signed.
+
 * `empty()`: Commits modifying no files. This also includes `merges()` without
   user modifications and `root()`.
 
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -181,6 +181,8 @@ pub enum RevsetFilterPredicate {
     },
     /// Commits with conflicts
     HasConflict,
+    /// Commits that are cryptographically signed.
+    Signed,
     /// Custom predicates provided by extensions
     Extension(Rc<dyn RevsetFilterExtension>),
 }
@@ -830,6 +832,11 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
         Ok(RevsetExpression::filter(RevsetFilterPredicate::AuthorDate(
             pattern,
         )))
+    });
+    map.insert("signed", |_diagnostics, function, _context| {
+        function.expect_no_arguments()?;
+        let predicate = RevsetFilterPredicate::Signed;
+        Ok(RevsetExpression::filter(predicate))
     });
     map.insert("mine", |_diagnostics, function, context| {
         function.expect_no_arguments()?;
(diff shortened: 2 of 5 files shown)


cli/src/config/revsets.toml
```diff
@@ -11,7 +11,7 @@ log = "present(@) | ancestors(immutable_heads().., 2) | present(trunk())"
 # Emit the working-copy branch first, which is usually most interesting.
 # This also helps stabilize output order.
 log-graph-prioritize = "present(@)"
-sign = "reachable(@, mutable())"
+sign = "reachable(@, mutable()) & ~signed()"
```
yuja, line 14: I wouldn't change the default because it's not so obvious that signed commits are skipped by default.

fwiw, we'll need to re-sign descendants of commits to be signed, so the query might have to be `(<target> & ~signed())::`.

lib/src/revset.rs
```diff
@@ -831,6 +833,11 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
             pattern,
         )))
     });
+    map.insert("signed", |_diagnostics, function, _context| {
+        function.expect_no_arguments()?;
+        let predicate = RevsetFilterPredicate::Signed;
+        Ok(RevsetExpression::filter(predicate))
+    });
```
yuja, line 840: Can you add a few tests to `lib/tests/test_revset.rs`? I think some setup code can be copied from `test_signing.rs`.

yuja (commented):
Thanks.

cli/src/config/revsets.toml
nilskch: Sounds good, but this PR won't close #5786 in that case. 

I will change the settings locally and play around with the query. Now that you mention it, I'm not sure when you need to sign what commit again (e.g. if you rebase or change an earlier commit). 

Over at Palantir, we have a CI check that ensures that every commit is signed with a known hardware GPG key, so this might be a good test for what setting works best.

I can report back after some time.

docs/revsets.md
```diff
@@ -288,6 +288,8 @@ revsets (expressions) as arguments.
 * `author_date(pattern)`: Commits with author dates matching the specified [date
   pattern](#date-patterns).
 
+* `signed()`: Commits that are cryptographically signed.
```
yuja, line 291: nit: can you move this after `committer_date()` or before `author()`?
`author*()` and `committer*()` functions are quite similar.

lib/tests/test_revset.rs
```diff
@@ -3049,6 +3052,54 @@ fn test_evaluate_expression_mine() {
     );
 }
 
+#[test]
+fn test_evaluate_expression_signed() {
+    let signer = Signer::new(Some(Box::new(TestSigningBackend)), vec![]);
+    let settings = testutils::user_settings();
+    let test_workspace =
+        TestWorkspace::init_with_backend_and_signer(TestRepoBackend::Git, signer, &settings);
```
yuja, line 3060: nit: if we don't have a reason to use the `Git` backend, `TestRepoBackend::Test` is preferred.

yuja (approved):

PR #6306, opened 2025-04-10
cli: abandon: update warning about deleted bookmarks

senekor:


calebdw:
I didn't see your PR, but here's an alternate solution:

https://github.com/jj-vcs/jj/pull/6308

senekor:
Oh, that's nice. I like the downgrade from warning to hint, since the situation is less risky now that `--deleted` is required. Closing because I prefer your version.

Issue #5685, opened 2025-02-13
Improve jj library's reliability by reducing panics


AM5800:
I am creating this bug to discuss this issue, propose some ways to address it and maybe track progress.

We believe that panics can cause significant stability issues in production server environments where jj is used as a library. In these contexts, a panic is not much different from a segfault, and we're looking to minimize this risk.

As demonstrated by #5360, panics can occur in unexpected ways:
While the standard `heads` implementation may be safe, the library's flexibility allows for custom backends that can introduce failure points.
So in our specific use case, our custom `heads` implementation involved an RPC call that could fail, leading to an unavoidable panic.

What I propose:
1. Introduce a coding style guideline that emphasizes the avoidance of new panic sites and propagating errors instead.
2. Add a lint check that flags potential panic sites. Some sources of panic might be benign (e.g. `check - unwrap`, so there should be a mechanism to bypass such lint check.
3. Review existing panic sites

I am happy to join the effort on (3), but I would also like to have some way to prevent new panics from being added.

P.S. While this issue primarily addresses panics within the jj library, it's worth noting that preventing panics in the CLI is also beneficial. Doing so can lead to more informative error messages, enhancing the overall CLI experience.

PR #6307, opened 2025-04-10
docs: add a style guide

martinvonz:
#5685

# Checklist

If applicable:

### DIFF
--- a/docs/style_guide.md
+++ b/docs/style_guide.md
@@ -1,0 +1,12 @@
+# Style guide
+
+## Panics
+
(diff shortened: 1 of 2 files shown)


docs/style_guide.md
```diff
@@ -0,0 +1,16 @@
+# Style guide
+
+## Imports (`use` keyword)
+
+Each symbol should be on a separate line (to simplify merging).
```
emilazy, line 5: I think we don’t need to write things CI enforces here, right? A machine can tell people when it’s wrong and a machine can fix it, so there’s no need for human nitpicking or toil.

PR #3638, opened 2024-05-07
cli: improve `jj version` output

thoughtpolice:
Spruces up the `version` output to be a little more informative. Adds two new flags for more specific information. The new output was somewhat inspired by the way coreutils tools print output, and other similar tools that have some equivalent of things like `--numeric-only`. Best read commit by commit as usual.

Fixes #3629.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #6311, opened 2025-04-10
cli: fix version test when @ has multiple parents

winterqt:
The `jj log` command ran by the build script returns the commit IDs concatenated together, while the test expects only one ID in the version string.

# Checklist

If applicable:

### DIFF
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -61,6 +61,7 @@ fn get_git_hash() -> Option<String> {
             "log",
             "--no-graph",
             "-r=@-",
+            "-n=1",
             "-T=commit_id",
         ])
         .output()


thoughtpolice:
This has hit people before (normally I don't run into it, because when I do a mega-merge I always make my `@` sit on top of the merge rather than be the merge itself) but I wonder if we shouldn't just delete the test instead and make the version string more obvious in the `build.rs` file... I honestly don't know if the version test is worth its weight for stuff like this. /cc @ilyagr

cli/build.rs
```diff
@@ -61,6 +61,7 @@ fn get_git_hash() -> Option<String> {
             "log",
             "--no-graph",
             "-r=@-",
```
ilyagr, line 63: I think we could be slightly more deterministic and do `-r=latest(@-)`.

My initial idea was to include all the commits with a separator, but I guess this is simpler.

There are still many things wrong with it, e.g. it doesn't detect whether `@` is non-empty, I believe, so I think the loss of precision is OK.

------

I wonder if we should implement `jj version --verbose` that would include all the parents and/or, perhaps,

In an ideal world, I think `jj --version` would mention things like:

- Number of commits since last release, similarly to `git describe`.
- Id(s) of the nearest non-empty ancestor of `@` (or `@` itself if it's non-empty)
- Id(s) of `root(@..main)`, if that's different from the above commit, and the number of non-empty commits since then (including `@` if it's non-empty).

ilyagr (approved):
Thanks! This has bothered me a few times, I think it's worth the quick fix.

ilyagr:
Some less actionable and more wishful thoughts.


I wonder if we should implement `jj version --verbose` that would include all the parents and/or, perhaps,

In an ideal world, I think `jj --version` would mention things like:

- Number of commits since last release, similarly to `git describe`.
- Id(s) of the nearest non-empty ancestor of `@` (or `@` itself if it's non-empty)
- Id(s) of `root(@..main)`, if that's different from the above commit, and the number of non-empty commits since then (including `@` if it's non-empty).


It's somewhat tedious to do, though.

PhilipMetzger:
> I wonder if we should implement `jj version --verbose` that would include all the parents and/or, perhaps,
> 
> In an ideal world, I think `jj version --verbose` would mention things like:
> 
> * Number of commits since last release, similarly to `git describe`.
> * Id(s) of the nearest non-empty ancestor of `@` (or `@` itself if it's non-empty)
> * Id(s) of `root(@..main)`, if that's different from the above commit, and the number of non-empty commits since then (including `@` if it's non-empty).
> 
> It's somewhat tedious to do, though.

There is @thoughtpolice PR which brings some of these features here, https://github.com/jj-vcs/jj/pull/3638 imo longterm its still the correct fix, even though nothing has happened with it for half a year.

PR #6312, opened 2025-04-10
config-schema: add templates

winterqt:
# Checklist

If applicable:

### DIFF
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -782,6 +782,67 @@
             "additionalProperties": {
                 "type": "boolean"
             }
+        },
+        "templates": {
+            "type": "object",
+            "description": "Definitions for the templates that various jj commands use",
+            "properties": {
+                "backout_description": {
+                    "type": "string",
+                    "description": "The description of commits backed out by `jj backout` (deprecated; use `revert_description` instead)"
+                },
+                "bookmark_list": {
+                    "type": "string",
+                    "description": "`jj bookmark list`'s output"
+                },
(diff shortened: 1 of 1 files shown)


cli/src/config-schema.json
```diff
@@ -782,6 +782,64 @@
...
             }
+        },
+        "templates": {
+            "type": "object",
+            "description": "Definitions for the templates that various jj commands use",
+            "properties": {
+                "backout_description": {
+                    "type": "string",
+                    "description": "(DEPRECATED) The description of commits backed out by `jj backout`"
+                },
```
martinvonz, line 793: Maybe we don't need to include this since we want people to use `revert_description` instead, or what do you think? For auto-complete purposes it's better to not include it. I don't know if there's value in having it for syntax highlighting or something of a value the user already has in their config.

martinvonz (approved):
Thanks!

cli/src/config-schema.json
thoughtpolice: I think it's probably fine? Some users may not have upgraded yet to where backout is deprecated, but might be using the `prerelease` schema anyway (it's annoying to update every time you update.) It will always be approximate by that metric I guess, but leaving it in doesn't hurt, and removing it later should be easy when we grep around for it.

winterqt: It would only be useful for autocomplete. I'm going to keep it in, given what Austin said above. We still have to remove it anyways.

I changed the wording/formatting of the deprecation notice in the description to be a bit clearer.

PR #6315, opened 2025-04-11
config: only add new when no other config exists

strega-nil:
In current jj, when you have something like:

```
~/
  .jjconfig.toml
  .config/
```

even though `.jjconfig.toml` exists, and `~/.config/jj/config.toml` does not, `jj config edit --user` will ask you if you would like to either edit `.jjconfig.toml`, or create `~/.config/jj/config.toml` newly.

After this patch, `jj config edit --user` in this situation will just edit the config.

# Checklist

If applicable:

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -235,14 +235,21 @@ impl UnresolvedConfigEnv {
         } else if let (Some(path @ New(_)), None) = (home_config_path, &platform_config_path) {
             paths.push(path);
         }
-        // This should be the default config created if there's
-        // no user config and `jj config edit` is executed.
-        if let Some(path) = platform_config_path {
-            paths.push(path);
-        }
+
         if let Some(path @ Existing(_)) = platform_config_dir {
             paths.push(path);
         }
+
+        // This should be the default config created if there's
+        // no user config and `jj config edit` is executed.
+        if let Some(path) = platform_config_path {
+            if paths.is_empty() {
+                paths.push(path);
+            } else if let ConfigPath::Existing(_) = path {
+                paths.push(path);
+            }
+        }
+
         paths
     }
 }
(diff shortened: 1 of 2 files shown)


cenviity:
@strega-nil I edited the OP to fix up the `jj config edit --user` commands.

cli/src/config.rs
```diff
@@ -1457,8 +1455,8 @@ mod tests {
                 ..Default::default()
             },
             wants: &[
-                Want::Existing("config/jj/config.toml"),
                 Want::Existing("config/jj/conf.d"),
+                Want::Existing("config/jj/config.toml"),
```
calebdw, line 1459: Order matters here so this is a regression---the `config.toml` should come before the `conf.d` directory

cli/src/config.rs
```diff
@@ -1478,8 +1476,8 @@ mod tests {
             // Precedence order is important
             wants: &[
                 Want::Existing("home/.jjconfig.toml"),
-                Want::Existing("config/jj/config.toml"),
                 Want::Existing("config/jj/conf.d"),
+                Want::Existing("config/jj/config.toml"),
```
calebdw, line 1480: This is a regression

cli/src/config.rs
```diff
@@ -1441,10 +1442,7 @@ mod tests {
                 config_dir: Some("config".into()),
                 ..Default::default()
             },
-            wants: &[
-                Want::New("config/jj/config.toml"),
-                Want::Existing("config/jj/conf.d"),
-            ],
+            wants: &[Want::Existing("config/jj/conf.d")],
```
calebdw, line 1445: I think we still want the new `config.toml` in this case

strega-nil:
While attempting to reproduce this, I had the unfortunate realization that this was me being a bit of a dumbass. It turns out, I had put an empty config file in `~/.config/jj/config.toml`, which was found _only_ after I made the changes in #6300. Closing as "not a bug".

PR #6318, opened 2025-04-11
git_backend: load "git gc" executable name from settings

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -81,6 +81,7 @@ use crate::object_id::ObjectId;
 use crate::repo_path::RepoPath;
 use crate::repo_path::RepoPathBuf;
 use crate::repo_path::RepoPathComponentBuf;
+use crate::settings::GitSettings;
 use crate::settings::UserSettings;
 use crate::stacked_table::MutableTable;
 use crate::stacked_table::ReadonlyTable;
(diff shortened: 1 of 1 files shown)


thoughtpolice (approved):

emilazy:
Could we use this to drop the `hermetic_git` environment unsafety after `git2` is gone, by setting `git.executable-path` to a wrapper that isolates configuration in the tests?

yuja:
> Could we use this to drop the `hermetic_git` environment unsafety after `git2` is gone, by setting `git.executable-path` to a wrapper that isolates configuration in the tests?

Maybe no? It might work in CLI tests, but we'll still need to set env vars for in-process `gix` operations.

emilazy:
Isn’t that what we [pass `gix::open::Options::isolated()`](https://github.com/jj-vcs/jj/blob/1305912fcf3deb359730c54844bd19b3d259e871/lib/src/git_backend.rs#L1587-L1591) for? Or are there code paths we test that don’t let us pass that in?

yuja:
> Isn’t that what we [pass `gix::open::Options::isolated()`](https://github.com/jj-vcs/jj/blob/1305912fcf3deb359730c54844bd19b3d259e871/lib/src/git_backend.rs#L1587-L1591) for? Or are there code paths we test that don’t let us pass that in?

We have `GitBackend` which should respect the user configuration. Maybe we could inject some parameters to `StoreFactories` to not load user configuration, but I don't know if it would be worth the complexity to be added.

emilazy:
Right. I’d like to eliminate the environment variable setting (because even though [it should be safe with `cargo nextest`](https://nexte.st/docs/configuration/env-vars/#altering-the-environment-within-tests), that doesn’t necessarily transfer to `cargo test`, and the function itself definitely isn’t unconditionally safe, so we’re kind of breaking the language rules by making it and the helpers that call it not `unsafe`), so I might look into it once `git2` is gone, but makes sense that it’s not quite that simple.

PR #6321, opened 2025-04-11
revset: don't crash if revset engine returns non-BackendError

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -1186,13 +1186,12 @@ impl MutableRepo {
                 self.parent_mapping.keys().cloned().collect(),
             ))
             .evaluate(self)
-            .map_err(|err| err.expect_backend_error())?;
+            .map_err(|err| err.into_backend_error())?;
         let to_visit = to_visit_revset
             .iter()
             .commits(self.store())
             .try_collect()
-            // TODO: Return evaluation error to caller
-            .map_err(|err| err.expect_backend_error())?;
+            .map_err(|err| err.into_backend_error())?;
         Ok(to_visit)
     }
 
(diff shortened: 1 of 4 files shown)


lib/src/repo.rs
```diff
@@ -1186,13 +1186,13 @@ impl MutableRepo {
                 self.parent_mapping.keys().cloned().collect(),
             ))
             .evaluate(self)
-            .map_err(|err| err.expect_backend_error())?;
+            .map_err(|err| err.into_backend_error())?;
         let to_visit = to_visit_revset
             .iter()
             .commits(self.store())
             .try_collect()
             // TODO: Return evaluation error to caller
```
yuja, line 1194: nit: remove TODO

lib/src/revset.rs
```diff
@@ -103,10 +103,10 @@ pub enum RevsetEvaluationError {
 }
 
 impl RevsetEvaluationError {
-    pub fn expect_backend_error(self) -> BackendError {
+    pub fn into_backend_error(self) -> BackendError {
         match self {
             Self::StoreError(err) => err,
-            Self::Other(err) => panic!("Unexpected revset error: {err}"),
+            Self::Other(err) => BackendError::Other(err),
```
yuja, line 109: > We may want to create some higher-level `RepoError` to return here,

Can you leave a TODO comment? I rejected this implementation once because `RevsetEvaluationError::Other` shouldn't be a `BackendError` (or `CommitStoreError`.)

yuja (approved):

lib/src/repo.rs
martinvonz: Done

lib/src/revset.rs
martinvonz: Done

PR #6323, opened 2025-04-11
merge_tools: don't return error from builtin editor if the user cancelled

winterqt:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* The `:builtin` diff and merge editors no longer throw an error if the operation is
+  cancelled.
+
 ### Packaging changes
 
 * Jujutsu now uses
--- a/cli/src/merge_tools/builtin.rs
+++ b/cli/src/merge_tools/builtin.rs
@@ -481,8 +481,14 @@ pub fn edit_diff_builtin(
         },
         &mut input,
     );
-    let result = recorder.run().map_err(BuiltinToolError::Record)?;
-    let tree_id = apply_diff_builtin(&store, left_tree, right_tree, changed_files, &result.files)
+
+    let files = match recorder.run() {
+        Ok(res) => res.files,
+        Err(scm_record::RecordError::Cancelled) => return Ok(left_tree.id()),
+        Err(e) => return Err(e.into()),
+    };
+
+    let tree_id = apply_diff_builtin(&store, left_tree, right_tree, changed_files, &files)
         .map_err(BuiltinToolError::BackendError)?;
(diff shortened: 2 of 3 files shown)


winterqt:
Nit: should the first commit message have a `cli: ` prefix? Doesn't look like many people do it for merge_tools specifically, but I'd rather err on the side of perfection. 😉

CHANGELOG.md
```diff
@@ -38,6 +38,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* The `:builtin` diff and merge editors no longer throw an error if the operation is
+  cancelled.
```
winterqt, line 42:
```suggestion
* The `:builtin` diff and merge editors no longer throw an error when the operation is
  cancelled.
```
probably?

cli/src/merge_tools/builtin.rs
```diff
@@ -481,8 +481,14 @@ pub fn edit_diff_builtin(
...
         &mut input,
     );
-    let result = recorder.run().map_err(BuiltinToolError::Record)?;
-    let tree_id = apply_diff_builtin(&store, left_tree, right_tree, changed_files, &result.files)
+
+    let files = match recorder.run() {
+        Ok(res) => res.files,
+        Err(scm_record::RecordError::Cancelled) => return Ok(left_tree.id()),
+        Err(e) => return Err(e.into()),
+    };
```
yuja, line 489: Does this mean the rewriting process is no longer interrupted? That seemed not what user expects for cancelled edits.

FWIW, if we need such behavior, it's probably better to change the return type to `Result<Option<MergedTreeId>>` so the caller should be aware of that.

winterqt: Ah, yeah, that was my mistake... I only tested a subset of commands where it is interrupted in the case that the tree ID does not change, which is what led me to do this instead of switching it to `Optional`.

yuja: btw, if we change the interpretation of the "cancelled" error, we'll also need to update the external tool code path consistently. That might mean `None` isn't enough to propagate the canceling token.

winterqt: > That might mean `None` isn't enough to propagate the canceling token.

I'm not sure what you mean here. Assuming we successfully signal the cancellation from every external merge tool (...somehow...), why isn't it enough?

(I guess an easy way to check this would be compare the tree IDs within the external tool code path, actually, then return `None` if they're identical.)

yuja: iirc, cancellation in external tools is either by exit code (-1 or something) or by no content changes. It's important to tell what caused the command to be cancelled.

PR #6328, opened 2025-04-12
nix: fix tests in darwin sandbox

winterqt:
I imagine it may be a good idea to poke the Taplo folks about this, but this is a good stopgap in the meantime for the two people who use Jujutsu via Nix on macOS with the sandbox enabled. 😉

# Checklist

If applicable:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -116,6 +116,12 @@
             "^target/"
           ];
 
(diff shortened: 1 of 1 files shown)


winterqt:
Will also cherrypick to the Nixpkgs package once this is merged.

thoughtpolice (approved):

PR #6332, opened 2025-04-13
cli: git push: show hint when there are bookmarks pending deletion

winterqt:
This comes from a personal experience where the following sequence of events happened:

1. `jj git push -c @-`
2. Oops, that was the wrong one, so I'll `jj bookmark delete` and `jj git push`
3. Hm... why didn't that do anything? Oh, right, `--deleted`

As described in the diff, this hint only fires when `jj git push` is used with no arguments, as any user that passes `--all`, `--tracked`, or `--revisions` (targeting a deleted bookmark) will already get an error instructing them to use `--deleted`.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Added `duplicate_description` template, which allows [customizing the descriptions
   of the commits `jj duplicate` creates](docs/config.md#duplicate-commit-description).
 
+* `jj git push` now shows a hint to use `--deleted` if there are bookmarks
+  pending deletion.
+
 ### Fixed bugs
 
 * Fixed crash on change-delete conflict resolution.
--- a/cli/src/commands/git/push.rs
+++ b/cli/src/commands/git/push.rs
@@ -393,6 +393,27 @@ pub fn cmd_git_push(
             ),
             remote = remote.as_symbol()
         );
+
+        // We only show this hint when pushing with no arguments because if the user is
+        // explicitly selecting a tracked bookmark with `--revisions`, then they'll get
+        // an error instructing them to use `--deleted`.
+        if use_default_revset {
+            let deleted_bookmarks = tx
+                .repo()
+                .view()
+                .local_remote_bookmarks(remote)
+                .filter(|(_, targets)| {
+                    targets.local_target.is_absent() && targets.remote_ref.is_tracked()
+                })
+                .count();
+
+            if deleted_bookmarks > 0 {
+                writeln!(
+                    ui.hint_default(),
+                    "You can push {deleted_bookmarks} deleted bookmark(s) with `--deleted`"
+                )?;
+            }
+        }
     }
     if bookmark_updates.is_empty() {
         writeln!(ui.status(), "Nothing changed.")?;
--- a/cli/tests/test_git_push.rs
+++ b/cli/tests/test_git_push.rs
@@ -1961,6 +1961,14 @@ fn test_git_push_deleted(subprocess: bool) {
     work_dir
         .run_jj(["bookmark", "delete", "bookmark1"])
         .success();
+    let output = work_dir.run_jj(["git", "push"]);
+    insta::assert_snapshot!(output, @r"
+    ------- stderr -------
+    Warning: No bookmarks found in the default push revset: remote_bookmarks(remote=origin)..@
+    Hint: You can push 1 deleted bookmark(s) with `--deleted`
+    Nothing changed.
+    [EOF]
+    ");
     let output = work_dir.run_jj(["git", "push", "--deleted"]);
     insta::allow_duplicates! {
     insta::assert_snapshot!(output, @r"


cli/src/commands/git/push.rs
```diff
@@ -395,6 +395,25 @@ pub fn cmd_git_push(
...
+                    && matches!(classify_bookmark_push_action(*targets),
+                        BookmarkPushAction::Update(update) if update.new_target.is_none())
+            })
+            .count();
+
+        if deleted_bookmarks > 0 {
+            writeln!(
+                ui.hint_default(),
+                "You can push {deleted_bookmarks} deleted bookmark{s} with `--deleted`",
+                s = if deleted_bookmarks > 1 { "s" } else { "" }
```
winterqt, line 413: Trailing comma?

cli/tests/test_git_push.rs
```diff
@@ -1961,6 +1961,14 @@ fn test_git_push_deleted(subprocess: bool) {
     work_dir
         .run_jj(["bookmark", "delete", "bookmark1"])
         .success();
+    let output = work_dir.run_jj(["git", "push"]);
```
winterqt, line 1964: Should I add a comment here explaining what exactly this is testing for, like elsewhere in this file?

cli/src/commands/git/push.rs
```diff
@@ -393,6 +393,27 @@ pub fn cmd_git_push(
...
+                        && matches!(classify_bookmark_push_action(*targets),
+                        BookmarkPushAction::Update(update) if update.new_target.is_none())
+                })
+                .count();
+
+            if deleted_bookmarks > 0 {
+                writeln!(
+                    ui.hint_default(),
+                    "You can push {deleted_bookmarks} deleted bookmark{s} with `--deleted`",
+                    s = if deleted_bookmarks > 1 { "s" } else { "" }
```
winterqt, line 413: Should I add a trailing comma here?

cli/src/commands/git/push.rs
```diff
@@ -393,6 +393,27 @@ pub fn cmd_git_push(
             ),
             remote = remote.as_symbol()
         );
+
+        if bookmark_updates.is_empty() {
```
yuja, line 397: Can you include in the commit message why you want this hint?

If your intuition was that `jj git push` (with no arguments) would push deleted bookmarks, it doesn't make sense to mention deleted bookmarks only if there are no other updates.

cli/src/commands/git/push.rs
```diff
@@ -393,6 +393,27 @@ pub fn cmd_git_push(
...
+            let deleted_bookmarks = tx
+                .repo()
+                .view()
+                .local_remote_bookmarks(remote)
+                .filter(|(_, targets)| {
+                    !targets.local_target.is_present()
```
yuja, line 403: nit: `targets.local_target.is_absent() && targets.remote_ref.is_tracked()`?

cli/src/commands/git/push.rs
```diff
@@ -393,6 +393,27 @@ pub fn cmd_git_push(
...
+                        && matches!(classify_bookmark_push_action(*targets),
+                        BookmarkPushAction::Update(update) if update.new_target.is_none())
+                })
+                .count();
+
+            if deleted_bookmarks > 0 {
+                writeln!(
+                    ui.hint_default(),
+                    "You can push {deleted_bookmarks} deleted bookmark{s} with `--deleted`",
```
yuja, line 412: nit: we avoid ad-hoc singular/plural handling. just use `N bookmarks` or `N bookmark(s)`.

CHANGELOG.md
```diff
@@ -38,6 +38,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * New `signed` revset function to filter for cryptographically signed commits.
 
+* `jj git push` now shows a hint to use `--deleted` if the only changed bookmarks
+  are deleted ones.
+
```
yuja, line 43: nit: I don't think changelog entry is needed for stuff like this. The user doesn't have to know that a better hint will be displayed.

cli/src/commands/git/push.rs
winterqt: What happened was that I created a local bookmark that pointed to `@-`, but then wanted to delete it. So I did that, then `jj git push`ed, assuming it would push since it was one of `@`'s ancestors, but then forgot that I needed to pass `--delete`.

I had that case in mind, which is why I did that. You're right that it would probably be better to do it unconditionally.

CHANGELOG.md
winterqt: I assumed it was worth mentioning because of prior art, e.g. https://github.com/jj-vcs/jj/blob/ee0e86fa63d8fa8165927f0f727cbb4d19568f3c/CHANGELOG.md?plain=1#L338-L339.

cli/src/commands/git/push.rs
winterqt: How's that for the commit message, does it explain why I think this is useful? I can add more detail, it's not the easiest to articulate :)

cli/src/commands/git/push.rs
winterqt: `rustfmt` doesn't, so on second thought, I won't; it's inconsistent with how all of the other macros are formatted throughout the codebase. (Ideally it'd be able to do it same as function calls, but I understand why it can't/doesn't.)

CHANGELOG.md
yuja: It might be because the hint could be misleading. I personally wouldn't add changelog for `jj undo`, though.

cli/src/commands/git/push.rs
yuja: > What happened was that I created a local bookmark that pointed to `@-`, but then wanted to delete it. So I did that, then `jj git push`ed,

In that case, maybe we can look for remote bookmarks to be deleted within the default push revset range (plus the remote bookmarks themselves.) It won't help if the deleted local bookmarks had been rewritten/rebased, but can greatly reduce the amout of false positives.

cli/src/commands/git/push.rs
yuja: Oh, I misread the comment, sorry. I meant period should be added to hint/warning messages. (Our style is a bit random about that, though.)

CHANGELOG.md
winterqt: Misleading how?

cli/src/commands/git/push.rs
winterqt: > In that case, maybe we can look for remote bookmarks to be deleted within the default push revset range (plus the remote bookmarks themselves.)

Won't we have to look at a previous operation for that, since the default push revset doesn't match in this case?

winterqt: > greatly reduce the amout of false positives

How would this lead to false positives? I'm unaware.

yuja: > Won't we have to look at a previous operation for that,

I meant something like `ancestors(remote_bookmarks(..)..@, 2) & <remote bookmarks to be deleted>`.

> How would this lead to false positives? I'm unaware.

There's no clue to assume that the user intended to push deleted bookmarks, right? It seems just wrong to list random deleted bookmarks.

CHANGELOG.md
yuja: because reverting undone operation is valid? Anyway, I don't think we needed to add it to the changelog.

cli/src/commands/git/push.rs
winterqt: > I meant something like `ancestors(remote_bookmarks(..)..@, 2) & <remote bookmarks to be deleted>`.

Hmm... yeah, maybe that might work well. Perhaps the message should be something like `Hint: You can delete X nearby remote bookmark(s) (and Y additional ones) with jj git push --deleted` or something.

> There's no clue to assume that the user intended to push deleted bookmarks, right? It seems just wrong to list random deleted bookmarks.

Fair. Well, they are tracked, so I'd imagine there's a good chance they want to? (And if they don't, then they'll be deleted as soon as they pass `--deleted`.

winterqt: Though I am now questioning the usefulness of this 😅 I'll defer to you on that end.

yuja: Since the default isn't `jj git push --all`, I don't think it's great idea to list "all" deleted bookmarks.

> Though I am now questioning the usefulness of this 😅 I'll defer to you on that end.

I personally don't think this hint will help. It's not easy to guess which deleted bookmarks would belong to the current push target (if they weren't deleted), and it's kidna obvious that `jj git push` doesn't push deleted bookmarks unless specified.

winterqt:
I think this has run its course, we can always revisit.

Issue #6298, opened 2025-04-09
Deprecate (and eventually delete) `ui.default-description`


avamsi:
With `templates.draft_commit_description` and now `templates.commit_trailers` (#6088), I don't think there's many use cases for `ui.default-description` (https://github.com/jj-vcs/jj/blob/main/docs/config.md#default-description).

PR #6335, opened 2025-04-14
cli: deprecate ui.default-description

winterqt:
Closes https://github.com/jj-vcs/jj/issues/6298.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   Please report any remaining issues you have with the Git
   subprocessing path.
 
+* `ui.default-description` has been deprecated, and will be removed
+  in a future release. Please migrate to [`templates.draft_commit_description`](docs/config.md#default-description)
+  and/or [`templates.commit_trailers`](docs/config.md#commit-trailers).
+
 ### New features
 
 * Color-words diff has gained [an option to compare conflict pairs without
--- a/cli/src/description_util.rs
+++ b/cli/src/description_util.rs
@@ -415,12 +415,6 @@ pub fn description_template(
     intro: &str,
     commit: &Commit,
 ) -> Result<String, CommandError> {
-    // TODO: Should "ui.default-description" be deprecated?
-    // We might want default description templates per command instead. For
-    // example, "backout_description" template will be rendered against the
-    // commit to be backed out, and the generated description could be set
-    // without spawning editor.
-
     // Named as "draft" because the output can contain "JJ:" comment lines.
     let template_key = "templates.draft_commit_description";
     let template_text = tx.settings().get_string(template_key)?;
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   Please report any remaining issues you have with the Git
   subprocessing path.
 
+* `ui.default-description` has been deprecated, and will be removed
+  in a future release. Please migrate to [`templates.draft_commit_description`](docs/config.md#default-description)
+  and/or [`templates.commit_trailers`](docs/config.md#commit-trailers) instead.
```
winterqt, line 31: Should we explicitly specify a release that this will be dropped in?

cli/src/description_util.rs
```diff
@@ -435,6 +429,21 @@ pub fn description_template(
     Ok(output.into_string_lossy())
 }
 
+pub fn print_default_description_deprecation_warning(
```
yuja, line 432: You can inject warning via `default_config_migrations()`. See how `git.subprocess` is tested.

CHANGELOG.md
yuja: I don't think we need the release number in changelog, but adding `TODO: Delete in jj x.xx` comment will help.

cli/src/description_util.rs
winterqt: Why would we rather use a `ConfigMigrationRule` for this? `git.subprocess` only uses one for when `git.subprocess` = `true` (and it doesn't look like that case is *tested* (e.g. in `tests/`) anywhere, but maybe I'm just misunderstanding), but uses a function like I did here when checking if the deprecation warning should be fired.

CHANGELOG.md
winterqt: When should we deprecate this? I read six releases somewhere, but unsure if that's universal or just on a case-by-case basis.

cli/src/description_util.rs
emilazy: I did it that way because someone setting `git.subprocess = false` might not immediately be able to move to the default (because of issues we haven’t yet fixed), so warning on every command would be spammy. If there’s a new configuration option users can immediately migrate to, just using the migration rule system is simpler.

winterqt: There's not one single option that can be moved to, though -- two options take the place of the one that people may have been using for either or both of the cases that the new options cover.

winterqt: Oh, you mean just warn, not automatically migrate. Gotcha.

CHANGELOG.md
yuja: The 6-month rule seems good for this. There are alternative options the user can switch to.

yuja (approved):
Thanks.

PR #6341, opened 2025-04-15
repo_path: remove assertion from constructors

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/repo_path.rs
+++ b/lib/src/repo_path.rs
@@ -668,6 +668,10 @@ mod tests {
         RepoPath::from_internal_string(value)
     }
 
+    fn repo_path_component(value: &str) -> &RepoPathComponent {
+        RepoPathComponent::new(value)
+    }
+
     #[test]
     fn test_is_root() {
         assert!(RepoPath::root().is_root());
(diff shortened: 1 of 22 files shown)


lib/src/repo_path.rs
```diff
@@ -821,28 +825,22 @@ mod tests {
...
-            vec![RepoPathComponent::new("dir")]
+            vec![repo_path_component("dir")]
         );
         assert_eq!(
             repo_path("dir/subdir").components().rev().collect_vec(),
-            vec![
-                RepoPathComponent::new("subdir"),
-                RepoPathComponent::new("dir"),
-            ]
+            vec![repo_path_component("subdir"), repo_path_component("dir"),]
```
yuja, line 867: nit: remove trailing `,`

lib/src/repo_path.rs
```diff
@@ -217,6 +227,15 @@ impl Debug for RepoPathBuf {
...
 }
 
+/// The `value` is not a valid repo path because it contains empty path
+/// component. For example, `"/"`, `"/foo"`, `"foo/"`, `"foo//bar"` are all
+/// invalid.
+#[derive(Clone, Debug, Eq, Error, PartialEq)]
+#[error(r#"Invalid repo path input "{value}""#)]
+pub struct InvalidNewRepoPathError {
+    value: String,
+}
```
yuja, line 237: nit: It seems inconsistent that we add new error type for `RepoPath::from_internal_string()`, but reuse the non-syntactic error type for `RepoPathComponent::new()`.

I don't have strong feeling whether we should add/reuse the error types, but it might be easier to use this `InvalidNewRepoPathError` type for malformed inputs of both `RepoPath` and `RepoPathComponent`.

yuja (approved):

lib/src/repo_path.rs
martinvonz: Makes sense. Switched to `InvalidNewRepoPathError`

PR #6363, opened 2025-04-18
cli: add config option for built-in pager's ruler

Samasaur1:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj absorb` can now squash a deleted file if it was added by one of the
   destination revisions.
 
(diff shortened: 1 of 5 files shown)


yuja (approved):
Thanks.

rwjblue:
Please excuse my ignorance here, what does the ruler do? It's not immediately obvious upon reading the docs here (though maybe if I knew the underlying tools better it would be clearer).

Samasaur1:
> Please excuse my ignorance here, what does the ruler do? It's not immediately obvious upon reading the docs here (though maybe if I knew the underlying tools better it would be clearer).

The ruler is what streampager calls the one-line-tall gray bar at the bottom of the screen that shows your progress through the file and a couple other things (whether the pager is waiting for more data, I think?). You can always toggle it with control-R, but you couldn't configure it to start disabled even though streampager itself has an option to do so

rwjblue:
Thank you!

PR #6376, opened 2025-04-19
templates: extract a template alias as hook for default commit descri…

martinvonz:
…ption

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -91,7 +91,7 @@ label(if(overridden, "overridden"),
 
 builtin_draft_commit_description = '''
 concat(
-  coalesce(description, "\n"),
+  coalesce(description, default_commit_description, "\n"),
   surround(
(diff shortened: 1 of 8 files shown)


cli/src/config/templates.toml
```diff
@@ -91,7 +91,7 @@ label(if(overridden, "overridden"),
 
 builtin_draft_commit_description = '''
 concat(
-  coalesce(description, "\n"),
+  coalesce(description, default_commit_description),
```
yuja, line 94: nit: I would leave `"\n"` here as `coalesce(description, default_commit_description, "\n")` because it's not so obvious why the default is `"\n"`.

yuja (approved):

cli/src/config/templates.toml
martinvonz: Done. I also made the default description an empty string.

PR #6378, opened 2025-04-19
docs: convert Git command table from HTML to YAML

cenviity:
The YAML format is much easier to maintain and allows using Markdown instead of having to escape symbols such as < and >.

# Checklist

If applicable:


martinvonz (approved):
I don't know much about MkDocs but this looks great to me. Thanks!

PR #6380, opened 2025-04-19
merge: add and use helpers for joining merge of futures

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/merge.rs
+++ b/lib/src/merge.rs
@@ -20,11 +20,13 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use std::fmt::Formatter;
 use std::fmt::Write as _;
+use std::future::Future;
 use std::hash::Hash;
 use std::iter::zip;
 use std::slice;
 use std::sync::Arc;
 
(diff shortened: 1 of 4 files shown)


lib/src/merge.rs
```diff
@@ -639,6 +641,20 @@ where
...
+    F: Future<Output = Result<T, E>>,
+{
+    /// Evaluates the futures in the input merge concurrently. Returns an error
+    /// if any of the futures returned an error.
+    pub async fn try_join_all(self) -> Result<Merge<T>, E> {
+        let values = try_join_all(self.values).await?;
+        Ok(Merge {
+            values: values.into(),
+        })
+    }
```
yuja, line 655: nit: this means we have to create a temporary `Merge { values: Vec<F> }`. The cost wouldn't matter, but it seems wasteful. Maybe we can instead add `.try_map_async()`?

martinvonz: Good idea. Done.

yuja (approved):

PR #6382, opened 2025-04-20
update_from_content: write sides to backend concurrently

martinvonz:
# Checklist

If applicable:


martinvonz:
>  async closures are unstable

I'll close this for now then. (It was just a minor improvement I noticed. I don't think it matters for the Google backend. And it certainly doesn't matter when using the Git backend.)

yuja:
We could use `|(content, file_id)| async { .. }`, but the lifetime would be a bit messier.

PR #6384, opened 2025-04-20
config: migrate ui.default-description to template alias

yuja:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   Please report any remaining issues you have with the Git
   subprocessing path.
 
-* `ui.default-description` has been deprecated, and will be removed
-  in a future release. Please migrate to `template-aliases.default_commit_description`,
+* `ui.default-description` has been deprecated, and will be migrated to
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -28,8 +28,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   Please report any remaining issues you have with the Git
   subprocessing path.
 
-* `ui.default-description` has been deprecated, and will be removed
-  in a future release. Please migrate to `template-aliases.default_commit_description`,
+* `ui.default-description` has been deprecated, and will be migrated to
+  `template-aliases.default_commit_description`. Please also consider using
```
martinvonz, line 32: nit: insert "automatically" or similar, to try to clarify that users don't need to do anything manually

martinvonz (approved):
Thanks!

CHANGELOG.md
yuja: The process isn't fully automated. The old value is rewritten in memory, but the underlying config file isn't updated.

martinvonz: Ah, I misremembered how it works. Sounds good as it is then. Thanks again.

PR #6386, opened 2025-04-20
docs: releasing: use `gh api --paginate` to construct contributor list

bnjmnt4n:
# Checklist

If applicable:

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -14,9 +14,8 @@ to run something like this:
 
 ```shell
 root=$(jj log --no-graph -r 'heads(tags())' -T 'commit_id')
-for i in $(seq 5); do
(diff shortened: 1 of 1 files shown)


martinvonz (approved):

docs/releasing.md
```diff
@@ -14,9 +14,8 @@ to run something like this:
 
 ```shell
 root=$(jj log --no-graph -r 'heads(tags())' -T 'commit_id')
```
arxanas, line 16: comment (off-topic): btw this command is probably wrong now that we have tags without common descendants 🙃 (i.e. `root` might resolve to more than one revision?).

bnjmnt4n: I guess we could use `latest(heads(tags))` which should usually return the desired tag.

bnjmnt4n: I will defer that for another PR if we decide that's good enoough.

PR #6387, opened 2025-04-20
Make TreeBuilder write independent trees concurrently

martinvonz:
# Checklist

If applicable:


martinvonz:
Actually, it should be better to use `FuturesUnordered` here instead of requiring a whole set of trees to be written before the next set can be started. I'll look into reworking it.

Issue #3772, opened 2024-05-27
FR: Add `--insert-before` and `--insert-after` to `split`, `new`, `duplicate`, etc.


thoughtpolice:
The `--insert-before` (AKA `--before`) and `--insert-after` (AKA `--after`) arguments are recent, useful additions to the `jj rebase` command that allow you to move revisions to any given part of the graph, with the desired edges.

However, these flags could generally apply to almost any command where we move or insert a commit in the graph. For example:

- `jj split`: given `X -> Y -> Z`, splitting `Y` results in the commits `B -> A` taking its place. Instead, we could actually insert `B` anywhere in the graph; this is useful for example when using [Austin's Mega Merge Strategy](https://ofcr.se/jujutsu-merge-workflow) in order to manipulate many series at once, and you want to make a small change and put it within the resulting graph.
  - Notably, the argument would probably apply to the *first* resulting commit `B`, not the second one `A`. This is I feel a better default, because in the case `A = @`, you probably want to remain where you are.
  - Currently, `jj split` supports a `--siblings` argument in order to make the resulting commit a sibling of the working copy; so it will split the commit `Y` above in such a way that we have edges `X -> B` and `X -> A`, which is just a specific instance of this functionality; something like `jj split -r Y --before Y+ --after Y-`, which can work for any `Y` at all.
  - This behavior is certainly incompatible with `--split` being provided, but as it's more general, I think that "special case" is OK.
- `jj duplicate`: this would allow you to insert a duplicate anywhere. This could be used for cherry picking, for example, with something like `jj duplicate -r BUGFIX --before 'empty()' --after X` where `X` is some destination, like a stable branch.
- `jj new`: this would allow you to insert a new working copy anywhere, not just as a leaf, but with arbitrary parents and children. Now, this one is a bit whacky, but I don't necessarily see a reason to *forbid* it, off the top of my head. I can think of some (small, easy) use cases for this, perhaps.

These are the main ones I can think of that don't have this functionality, roughly in order of actual real-world importance.

## Justification

While we can currently achieve all these with an extra `rebase` command, I think this is a much nicer and more consistent UX for users, especially ones who *already* use `rebase` with the `--before` and `--after` flags. These are very common wants people tend to have; "do a thing, then move it", and I think when you do this enough, you tend to want to do it often. So I think some reasonable shorthands for compound actions like this are valuable and reduce friction for new and experienced users alike.

[...]

PR #6389, opened 2025-04-20
split: add the --message option

glehmann:
to set the message on the first commit, where the selected changes are going. The commit with the remaining changes keeps the original description.
With this option it becomes possible to use split in a fully non-interactive way, in combination with the filset feature. It also makes jj split more consistent with other commands like squash, commit or new.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* `jj split` has gained a `--message` option to set the description of the
+  commit with the selected changes.
+
 ### Fixed bugs
 
 ### Packaging changes
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -32,6 +32,7 @@ use crate::complete;
 use crate::description_util::add_trailers;
 use crate::description_util::description_template;
 use crate::description_util::edit_description;
+use crate::description_util::join_message_paragraphs;
 use crate::ui::Ui;
 
 /// Split a revision in two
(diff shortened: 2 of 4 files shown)


PhilipMetzger:
I think this is the 3rd iteration of this, and its still unclear on how to do it. See https://github.com/jj-vcs/jj/pull/4469 and https://github.com/jj-vcs/jj/pull/5620 and their discussions.

martinvonz:
With the recent discussion about adding `-d/-A/-B` options to `jj split`, I think it's become more clear that the description should apply to the changes that were split out. It may be better to wait for someone to add those flags first.

glehmann:
> With the recent discussion about adding `-d/-A/-B` options to `jj split`, I think it's become more clear that the description should apply to the changes that were split out. It may be better to wait for someone to add those flags first.

I was also thinking about that. Could you point me to that discussion?

PhilipMetzger:
> I was also thinking about that. Could you point me to that discussion?

It's split between https://github.com/jj-vcs/jj/issues/3419 and Discord threads.

glehmann:
Thanks. I've also found https://github.com/jj-vcs/jj/issues/3772 interesting, including [that comment](https://github.com/jj-vcs/jj/issues/3772#issuecomment-2135791482) where @martinvonz proposes to rename `--destination` to `--onto`, which is also something I'd like to suggest.
I'll read all that before bothering people with changes that don't match what was already discussed!

glehmann:
BTW, I also wanted to suggest that the commit that gets the changes the user selects should have the new change id, and the commit that gets the rest of the changes should keep the change id of the original commit. I'm kind of expecting to find that idea in the discussions :-)

glehmann:
I finally found some time to implement `jj split -A/-B/-d` — see #6458

martinvonz (approved):
Thanks!

PR #6393, opened 2025-04-21
git_backend: do not include renamed trees in copy records

### LINKED ISSUE #6390: diff: panic when file and directory swap names
## Description

The following panic occurs if a file and a directory have their names swapped:

    thread 'main' panicked at cli/src/diff_util.rs:1240:13:
    Unexpected tree in diff at path "b"

The same thing happens if the two entities are a directory and a symlink.

Tested within a blank repo with jj 0.28.2:

    $ echo 'first file' > a
    $ mkdir b
    $ echo 'second file' > b/file
    $ jj new
    Working copy  (@) now at: toulwzts a0a5d2cd (empty) (no description set)
    Parent commit (@-)      : tlqysmuz 52817723 (no description set)
    $ mv a a~
    $ mv b a
    $ mv a~ b
    $ jj diff

    thread 'main' panicked at cli/src/diff_util.rs:1240:13:
    Unexpected tree in diff at path "b"

For convenience, I've also included the test repo here: https://github.com/mernen/jj-diff-test

## Expected Behavior

Should be presented as two renames. Git output, for comparison:

    $ git --git-dir=$(jj git root) show $(jj log --no-graph -T commit_id -r @)

    commit 05f6679114b0403c812ad7b709f2c666812f1bc6
    Author: Daniel Luz <dev.gsz@mernen.com>
    Date:   Sun Apr 20 16:03:31 2025 -0300

    diff --git a/b/file b/a/file
    similarity index 100%
    rename from b/file
    rename to a/file
    diff --git a/a b/b
    similarity index 100%
    rename from a
    rename to b

## Actual Behavior

Panic in diff_util.rs:

https://github.com/jj-vcs/jj/blob/b9ebe2f03c976515d2a155a411a368ae773c5493/cli/src/diff_util.rs#L925-L927

## Specifications

- Platform: macOS
- Version: 0.28.2

yuja:
Fixes #6390

# Checklist

If applicable:

### DIFF
--- a/lib/src/git_backend.rs
+++ b/lib/src/git_backend.rs
@@ -1352,13 +1352,20 @@ impl Backend for GitBackend {
             |change: gix::object::tree::diff::Change| -> BackendResult<Option<CopyRecord>> {
                 let gix::object::tree::diff::Change::Rewrite {
                     source_location,
+                    source_entry_mode,
                     source_id,
+                    entry_mode: dest_entry_mode,
                     location: dest_location,
                     ..
                 } = change
                 else {
                     return Ok(None);
                 };
+                // TODO: Renamed symlinks cannot be returned because CopyRecord
+                // expects `source_file: FileId`.
+                if !source_entry_mode.is_blob() || !dest_entry_mode.is_blob() {
+                    return Ok(None);
+                }
 
                 let source = str::from_utf8(source_location)
                     .map_err(|err| to_invalid_utf8_err(err, root_id))?;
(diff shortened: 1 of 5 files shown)


lib/src/git_backend.rs
```diff
@@ -1352,13 +1352,20 @@ impl Backend for GitBackend {
...
                     ..
                 } = change
                 else {
                     return Ok(None);
                 };
+                // TODO: Renamed symlinks cannot be returned because CopyRecord
+                // expects `source_file: FileId`.
+                if !source_entry_mode.is_blob() || !dest_entry_mode.is_blob() {
+                    return Ok(None);
+                }
```
martinvonz, line 1368: I'm a bit surprised that we get tree entries from Gitoxide. I wonder if that's a bug or intentional. I can see it easily being a bug that happens due to comparing tree entries exactly as a first step.

martinvonz (approved):
Thanks!

lib/src/git_backend.rs
yuja: `gix` appears to emit directory rename entry even if new file is added to the directory. So I think it's intentional?

PR #6395, opened 2025-04-21
cli: rebase: consolidate call sites of move_commits()

yuja:
cc @scott2000 

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/rebase.rs
+++ b/cli/src/commands/rebase.rs
@@ -560,16 +560,9 @@ fn rebase_descendants_transaction(
     }
 
     let mut tx = workspace_command.start_transaction();
-    let tx_description = if target_roots.len() == 1 {
-        format!(
-            "rebase commit {} and descendants",
-            target_roots.first().unwrap().id().hex()
-        )
-    } else {
-        format!(
-            "rebase {} commits and their descendants",
-            target_roots.len()
-        )
+    let tx_description = match &*target_roots {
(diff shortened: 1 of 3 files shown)


cli/src/commands/rebase.rs
```diff
@@ -554,11 +554,6 @@ fn rebase_descendants_transaction(
     target_roots: Vec<Commit>,
     rebase_options: &RebaseOptions,
 ) -> Result<(), CommandError> {
-    if target_roots.is_empty() {
-        writeln!(ui.status(), "Nothing changed.")?;
-        return Ok(());
-    }
-
```
steadmon, line 561: IIUC, we'll be creating (line 562) and finishing (line 580) a no-op transaction which previously wouldn't exist, but it looks like the logic in WorkspaceCommandHelper::finish_transaction avoids actually committing a no-op transaction, so I think this is OK. (Same situation for rebase_revisions_transaction() below.

steadmon (approved):
LGTM

cli/src/commands/rebase.rs
yuja: Correct. We need the manual "no-op" operation check only when that's easier and/or to print a better status message.

bnjmnt4n (approved):
Thanks for cleaning this up!

martinvonz (approved):

PR #6402, opened 2025-04-24
docs: document code to avoid duplicating gerrit change-id

mateusauler:
Adapted from:
- https://gist.github.com/thoughtpolice/8f2fd36ae17cd11b8e7bd93a70e31ad6#file-jjconfig-toml-L247
- https://github.com/avamsi/dotfiles/blob/28a0e40439de56fe179a1b0e705727c85e075464/.jjconfig.toml#L150

I'm not sure if I should add tests. I can add them if needed.

# Checklist

If applicable:

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -537,13 +537,23 @@ commits associated with it.
 
 ### How do I integrate Jujutsu with Gerrit?
 
-At the moment you'll need a script, which adds the required fields for Gerrit
-like the `Change-Id` footer. Then `jj` can invoke it via an `$EDITOR` override
-in an aliased command. Here's an [example][gerrit-integration] from a
-contributor (look for the `jj signoff` alias).
+Add this to your configuration to automatically add Change-Id trailers to commit messages:
+```toml
+[templates]
+commit_trailers = '''
+if(
+  !trailers.contains_key("Change-Id"),
+  format_gerrit_change_id_trailer(self)
+)
+'''
+```
+Note: If you don't check for the presence of the "Change-Id" trailer, you might
+occasionally get duplicate trailers.
+This happens when Jujutsu's change-id isn't in sync with the "Change-Id" trailer.
+Eg. after `jj split`, the "Change-Id" trailer generated for the new change would
+be different from the original one, it wouldn't be deduplicated.
 
-After you have attached the `Change-Id:` footer to the commit series, you'll
-have to manually invoke `git push` of `HEAD` on the underlying git repository
+You'll have to manually invoke `git push` of `HEAD` on the underlying git repository
 into the remote Gerrit bookmark `refs/for/$BRANCH`, where `$BRANCH` is the base
 bookmark you want your changes to go to (e.g., `git push origin
 HEAD:refs/for/main`). Using a [co-located][co-located] repo


cli/src/config/templates.toml
```diff
@@ -430,4 +430,4 @@ git_format_patch_email_headers = '''
 
 # Gerrit change id is 40 chars, jj change id is 32, so we need some padding.
 # `6a6a6964` is the hexadecimal of `jjid` in ASCII, just to not pad with zeros.
-"format_gerrit_change_id_trailer(commit)" = '"Change-Id: I6a6a6964" ++ commit.change_id().normal_hex() ++ "\n"'
+"format_gerrit_change_id_trailer(commit)" = 'if(!commit.description().contains("Change-Id: I"), "Change-Id: I6a6a6964" ++ commit.change_id().normal_hex() ++ "\n")'
```
yuja, line 433: nit: better to test this by caller? it's odd that `format_*()` function behaves differently depending on the old description.

yuja (commented):
@glehmann do you have any idea on this kind of requirements?

glehmann (commented):
I was hoping we wouldn't have to do this kind of check, because we can't easily know when searching a string in the description if it's actually a trailer.
We could implement the check in the rust code, with a configurable list of trailer to keep unique?

glehmann:
@mateusauler could you describe in what situation you already have a `Change-Id` trailer. Was it added by someone else?
Just trying to understand how often it might happen.

joyously:
> could you describe in what situation you already have a `Change-Id` trailer.

Which commands reprocess the description? (`rebase`, `split`, `new x y`, `squash` are some)
Does a signing of the commit affect it?

mateusauler:
> could you describe in what situation you already have a `Change-Id` trailer. Was it added by someone else? Just trying to understand how often it might happen.

This happens when the change's gerrit change-id does not match the one generated by jj.
Eg. When a new change is pulled, jj assigns a new change-id. Which, in turn, makes jj generate a different gerrit change-id.
Maybe we could generate jj's change-id based on gerrit's? I don't know if this is viable...

> Which commands reprocess the description? (`rebase`, `split`, `new x y`, `squash` are some) 

describe, commit, squash, split, maybe more

> Does a signing of the commit affect it?

If by signing you mean `jj sign`, then doesn't seem like it.

cli/src/config/templates.toml
mateusauler: By the caller, do you mean the user-configured template or the internal rust function? Or something else?

I get what you're saying, but since this trailer is based on "random" text (not something consistent like the user's name), it seems like this is a relatively unique case when checking for repetition.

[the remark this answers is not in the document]
mateusauler: For now, I added the beginning newline check. Is there a way to do a regex check?

yuja: > By the caller, do you mean the user-configured template or the internal rust function? 

The user-configured template (and the documentation example.)

I think glehmann's idea is better, btw.
https://github.com/jj-vcs/jj/pull/6402#pullrequestreview-2791080674
> We could implement the check in the rust code, with a configurable list of trailer to keep unique?

glehmann:
> This happens when the change's gerrit change-id does not match the one generated by jj. Eg. When a new change is pulled, jj assigns a new change-id. Which, in turn, makes jj generate a different gerrit change-id. Maybe we could generate jj's change-id based on gerrit's? I don't know if this is viable...

I don't think that's possible: the id generated by `format_gerrit_change_id_trailer` in padded in order to match the gerrit id usual length. If we were to generate the jj change id from the gerrit change id, we would loose some data, and wouldn't be able to regenerate the gerrit change id. With the current trailer mechanism, the `Change-id` trailer would be added again, because it wouldn't exactly match the one already there.

> > we can't easily know when searching a string in the description if it's actually a trailer.
> 
> Maybe we could add a .trailers() to the commit type?

yes, that's an option @yuja suggested in the trailer PR
 
> > We could implement the check in the rust code, with a configurable list of trailer to keep unique?
> 
> This could work, but how would that list be defined? In this case, a regex pattern or similar would be necessary, since the id is essentially random.

We would only match the trailer key, and keep the value already there.

@yuja do you have a preference between the `.trailers()` method in the template or the config for the trailer key unicity? I think I prefer the first one, but I don't have a strong opinion. I could even give a try at implementing one or the other, once I'm done with `jj split -A/-B/-d/-m` — unless someone feels like implementing it already :-)

mateusauler:
> I don't think that's possible: the id generated by format_gerrit_change_id_trailer in padded in order to match the gerrit id usual length. If we were to generate the jj change id from the gerrit change id, we would loose some data, and wouldn't be able to regenerate the gerrit change id. With the current trailer mechanism, the Change-id trailer would be added again, because it wouldn't exactly match the one already there.

We could just ignore those first 8 characters in the gerrit change-id. Both when checking for the trailer and generating jj's change-id after a pull. Since jj's template always generates the same ones, from our perspective, they're irrelevant even if they're not 6a6a6964.

> We would only match the trailer key, and keep the value already there.

Makes sense.

> unless someone feels like implementing it already :-)

I might try hacking on it a bit over the weekend. No promises though...

yuja:
> We would only match the trailer key, and keep the value already there.
> 
> @yuja do you have a preference between the `.trailers()` method in the template or the config for the trailer key unicity? I think I prefer the first one, but I don't have a strong opinion.

I think config is better if that's sufficient and common requirement. If it's rarely needed, `if(!trailers.contains("Change-Id"), format_change_id(..))` seems good.

`.trailers()` will be useful as its own (to render trailers differently in `jj show` for example), but it's inconsistent that we deduplicate new trailers by `(key, value)` match, whereas the user has to do that manually for `key` match.

mateusauler:
I have opened #6419 that implements `.trailers()`.

docs/FAQ.md
```diff
@@ -537,13 +537,18 @@ commits associated with it.
...
-At the moment you'll need a script, which adds the required fields for Gerrit
-like the `Change-Id` footer. Then `jj` can invoke it via an `$EDITOR` override
-in an aliased command. Here's an [example][gerrit-integration] from a
-contributor (look for the `jj signoff` alias).
+Add this to your configuration:
+```toml
+[templates]
+commit_trailers = '''
+if(
+  !self.trailers().filter(|t| t.key() == "Change-Id"),
```
prattmic, line 545: Now that #6425 is in, this can use the exact same form as https://jj-vcs.github.io/jj/latest/config/#commit-trailers, minus the Signed-off-by portion.

docs/config.md
```diff
@@ -255,6 +255,9 @@ Some ready-to-use trailer templates are available for frequently used trailers:
   using the committer info;
 * `format_gerrit_change_id_trailer(commit)` creates a "Change-Id" trailer
   suitable to be used with Gerrit. It is based Jujutsu's change id.
+  Note: If you simply set your trailers to this template,
+  you might get duplicate "Change-Id" trailers sometimes.
```
prattmic, line 259: I think it would help to explain why this occurs and what you should do in those cases. (Probably in the Gerrit FAQ entry, not here.)

The most common case that comes to mind for me is splitting a change. One change id will remain the same, and the other will change. In the second change, you'll get two Change-Ids without `contains_key` deduplication. Even with deduplication, the second change will end up with the same Change-Id as the first, which is not what you actually want. It would be nice to have some guidance about needing to decide which Change-Id you want to keep and which you want to regenerate.

I'm also happy to send a PR for this.

docs/FAQ.md
```diff
@@ -537,13 +537,23 @@ commits associated with it.
 
 ### How do I integrate Jujutsu with Gerrit?
 
-At the moment you'll need a script, which adds the required fields for Gerrit
-like the `Change-Id` footer. Then `jj` can invoke it via an `$EDITOR` override
-in an aliased command. Here's an [example][gerrit-integration] from a
-contributor (look for the `jj signoff` alias).
+Add this to your configuration:
```
prattmic, line 540: nit: I think this should be explicit about what this is for. Something like "Add this to your configuration to automatically add Change-Id footers to commit messages:"

prattmic (approved):
LGTM

yuja (approved):

glehmann (changes requested):

docs/FAQ.md
```diff
@@ -537,13 +537,23 @@ commits associated with it.
...
+```toml
+[templates]
+commit_trailers = '''
+if(
+  !trailers.contains_key("Change-Id"),
+  format_gerrit_change_id_trailer(self)
+)
+'''
+```
+Note: If don't check for the presence of the "Change-Id" trailer, you might
```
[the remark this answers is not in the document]
mateusauler: oops...

glehmann (approved):

PR #6403, opened 2025-04-24
cli: put editor-*.jjdescription file in /tmp instead of .jj/repo/

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj git push -c`/`--change` no longer moves existing local bookmarks.
 
+* The `editor-*.jjdescription` files passed to your editor by e.g. `jj describe`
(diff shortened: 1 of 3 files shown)


yuja:
> I'm not sure I've ever used one of these files. I have certainly never used a file that's not from the most recent edit. I have, however, cleaned up old such files.

Do you have any other problems? I don't have strong feeling, but there might be tooling that depends on the description file path to look up the repository root. We can of course use environment variable or cwd for that purpose, though.

martinvonz:
> Do you have any other problems?

No, that's all. I've seen these files in my own repos and other people's repos. Sometimes there are tens or hundreds of them. Seemingly for no good reason.

> I don't have strong feeling, but there might be tooling that depends on the description file path to look up the repository root. We can of course use environment variable or cwd for that purpose, though.

Yes, that's probably better.

yuja (approved):

sins921 (approved):

PR #6407, opened 2025-04-24
completion: fix bash and zsh shell completion when completing aliases of multiple arguments

### LINKED ISSUE #5377: Argument completion in bash doesn't work for aliases with arguments
## Description

For aliases, the bash completion is broken. It only completes one element and it's always completed to the first argument in the alias.

1. Create an alias  `gp = ["git", "push"]`
1. `source <(COMPLETE=bash jj)`
1. Type `jj gp ` and then hit `<Tab>`

## Expected Behavior
Completion behaves the same as if I had typed `jj git push`, i.e. it would add `--` to my prompt.

## Actual Behavior
The completion adds `push` to my prompt. It even replaces partial input with `push`, i.e. `jj gp --he` + `<Tab>` becomes `jj gp push`.

## Specifications

- Platform: Linux Debian Sid
- Version: jj 0.25.0-041c4fecb77434dd6720e7d7f1ce48d9575ac5f7

jgreitemann:
Fixes #5377.

The fix ended up as exactly like @yuja mentioned in the issue and what the `TODO` comment in the code was already proposing. Still, it took a fair bit of fiddling around with the different shells to understand where they are differing in behavior exactly. 

I've gotten a wee bit carried away refactoring the existing tests to cover multiple shells (where sensible) without duplicating too much of the test code. Even so, insta needs a `match` over the different shell flavors to snapshot their output individually. For a while, I've entertained the idea of adding a `.normalize_shell_completions(shell)` function on the test output to basically strip away the comment (because bash doesn't have that) such that it would produce the same (normalized) output for each shell and we could do away with the `match`es. Obviously, that would lose the comment snapshots, so I decided against it for now. Let me know if that's something you want me to revisit.

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3551,29 +3551,55 @@ fn handle_shell_completion(
     config: &StackedConfig,
     cwd: &Path,
 ) -> Result<(), CommandError> {
+    let mut orig_args = env::args_os();
+
     let mut args = vec![];
     // Take the first two arguments as is, they must be passed to clap_complete
     // without any changes. They are usually "jj --".
-    args.extend(env::args_os().take(2));
+    args.extend(orig_args.by_ref().take(2));
 
     // Make sure aliases are expanded before passing them to clap_complete. We
     // skip the first two args ("jj" and "--") for alias resolution, then we
     // stitch the args back together, like clap_complete expects them.
-    let orig_args = env::args_os().skip(2);
     if orig_args.len() > 0 {
-        let arg_index: Option<usize> = env::var("_CLAP_COMPLETE_INDEX")
+        let complete_index: Option<usize> = env::var("_CLAP_COMPLETE_INDEX")
             .ok()
             .and_then(|s| s.parse().ok());
-        let resolved_aliases = if let Some(index) = arg_index {
+        let resolved_aliases = if let Some(index) = complete_index {
             // As of clap_complete 4.5.38, zsh completion script doesn't pad an
             // empty arg at the complete position. If the args doesn't include a
             // command name, the default command would be expanded at that
             // position. Therefore, no other command names would be suggested.
-            // TODO: Maybe we should instead expand args[..index] + [""], adjust
-            // the index accordingly, strip the last "", and append remainder?
             let pad_len = usize::saturating_sub(index + 1, orig_args.len());
-            let padded_args = orig_args.chain(std::iter::repeat_n(OsString::new(), pad_len));
-            expand_args(ui, app, padded_args, config)?
+            let padded_args = orig_args
+                .by_ref()
+                .chain(std::iter::repeat_n(OsString::new(), pad_len));
+
+            // Expand aliases left of the completion index.
+            let mut expanded_args = expand_args(ui, app, padded_args.take(index + 1), config)?;
+
+            // Adjust env var to compensate for shift of the completion point in the
+            // expanded command line.
+            // SAFETY: Program is running single-threaded at this point.
+            unsafe {
+                env::set_var(
+                    "_CLAP_COMPLETE_INDEX",
+                    (expanded_args.len() - 1).to_string(),
+                );
+            }
+
+            // Remove extra padding again to align with clap_complete's expectations for
+            // zsh.
+            let split_off_padding = expanded_args.split_off(expanded_args.len() - pad_len);
+            assert!(
+                split_off_padding.iter().all(|s| s.is_empty()),
+                "split-off padding should only consist of empty strings but was \
+                 {split_off_padding:?}",
+            );
+
+            // Append the remaining arguments to the right of the completion point.
+            expanded_args.extend(to_string_args(orig_args)?);
+            expanded_args
         } else {
             expand_args(ui, app, orig_args, config)?
         };
@@ -3598,17 +3624,22 @@ pub fn expand_args(
     args_os: impl IntoIterator<Item = OsString>,
     config: &StackedConfig,
 ) -> Result<Vec<String>, CommandError> {
-    let mut string_args: Vec<String> = vec![];
-    for arg_os in args_os {
-        if let Some(string_arg) = arg_os.to_str() {
-            string_args.push(string_arg.to_owned());
-        } else {
-            return Err(cli_error("Non-utf8 argument"));
-        }
-    }
-
+    let string_args = to_string_args(args_os)?;
     let string_args = resolve_default_command(ui, config, app, string_args)?;
     resolve_aliases(ui, config, app, string_args)
+}
+
+fn to_string_args(
+    args_os: impl IntoIterator<Item = OsString>,
+) -> Result<Vec<String>, CommandError> {
+    args_os
+        .into_iter()
+        .map(|arg_os| {
+            arg_os
+                .into_string()
+                .map_err(|_| cli_error("Non-UTF-8 argument"))
+        })
+        .collect()
 }
 
 fn parse_args(app: &Command, string_args: &[String]) -> Result<(ArgMatches, Args), clap::Error> {
--- a/cli/tests/test_completion.rs
+++ b/cli/tests/test_completion.rs
@@ -321,6 +321,7 @@ fn test_aliases_are_resolved(shell: Shell) {
 
     // user config alias
     test_env.add_config(r#"aliases.b = ["bookmark"]"#);
+    test_env.add_config(r#"aliases.rlog = ["log", "--reversed"]"#);
     // repo config alias
     work_dir
         .run_jj(["config", "set", "--repo", "aliases.b2", "['bookmark']"])
@@ -359,6 +360,30 @@ fn test_aliases_are_resolved(shell: Shell) {
         }
         _ => unimplemented!("unexpected shell '{shell}'"),
     }
+
+    let output = work_dir.complete_at(shell, 2, ["rlog", "--rev"]);
+    match shell {
+        Shell::Bash => {
+            insta::assert_snapshot!(output, @r"
+            --revisions
+            --reversed[EOF]
+            ");
+        }
+        Shell::Zsh => {
+            insta::assert_snapshot!(output, @r"
+            --revisions:Which revisions to show
+            --reversed:Show revisions in the opposite order (older revisions first)[EOF]
+            ");
+        }
+        Shell::Fish => {
+            insta::assert_snapshot!(output, @r"
+            --revisions	Which revisions to show
+            --reversed	Show revisions in the opposite order (older revisions first)
+            [EOF]
+            ");
+        }
+        _ => unimplemented!("unexpected shell '{shell}'"),
+    }
 }
 
 #[test]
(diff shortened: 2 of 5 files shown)


steadmon (approved):
Generally LGTM, but I'm not familiar with completion details so it might be worth having an expert review.

cli/src/cli_util.rs
```diff
@@ -3553,29 +3553,52 @@ fn handle_shell_completion(
...
+
+            // Expand aliases left of the completion index.
+            let mut expanded_args = expand_args(ui, app, padded_args.take(index + 1), config)?;
+
+            // Adjust env var to compensate for shift of the completion point in the
+            // expanded command line.
+            env::set_var(
+                "_CLAP_COMPLETE_INDEX",
+                (expanded_args.len() - 1).to_string(),
+            );
```
yuja, line 3588: nit: needs `unsafe { .. }` (in edition 2024)

(It would be nice if clap provided saner API.)

cli/tests/common/test_environment.rs
```diff
@@ -286,6 +299,66 @@ impl TestWorkDir<'_> {
         self.run_jj_with(|cmd| cmd.args(args))
     }
 
+    /// Runs `jj` as if shell completion had been triggered for the argument of
+    /// the given index.
+    #[must_use = "either snapshot the output or assert the exit status with .success()"]
+    pub fn complete_at<I>(&self, shell: Shell, index: usize, args: I) -> CommandOutput
```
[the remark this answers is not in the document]
jgreitemann: I can do that. Just to clarify: are you ok with it still being defined in an `impl TestEnvironment`/`impl TestWorkDir<'_>` to be able to call it as a member function? I think that should work across modules as long as it's the same crate? Or would you prefer an extension trait or a free function? I originally went with the free function but it was less appealing to read.

cli/tests/test_completion.rs
```diff
@@ -244,6 +245,30 @@ fn test_aliases_are_resolved(shell: Shell) {
             }
         }
     }
+
+    let output = work_dir.complete_at(shell, 2, &["jj", "rlog", "--rev"]);
```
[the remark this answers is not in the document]
jgreitemann: You are right that there is in fact a difference in behavior between bash/zsh and fish here. This is completely independent of alias expansion and I think it is an inherent flaw in how `clap_complete` chooses to drive the completer.

Take your example, but with the alias "manually" expanded (I use `|` to denote the caret; not a shell pipe):
```sh
> jj file show -r | CHANGELOG.md
```
This shows all revisions in fish, but in bash and zsh it will attempt to complete `CHANGELOG.md` as a revision and thus comes back empty-handed. You can see that a little more clearly when using a token like `m` which could be completed as a file (`mkdocs.yml`) or as a revision (`main@origin`):
```sh
> jj file show -r | m
```
inserts `m` when pressing tab once (so, `jj file show -r m| m`) and will then list revisions starting in "m" when pressing tab a second time. That's true for both bash and zsh and it holds true with or without the changes in this PR. The effects of this PR are only observed when attempting to complete
```sh
> jj cat -r | m
```
which used to complete to `jj cat -r -r| m` but now behaves exactly like the manually expanded version, completing to `jj file show -r m| m`. That's not great, but at least it's consistent.

The problem is that the completion script that is sourced by the shells will omit any intermediate empty tokens. Bash preserves a trailing empty token, zsh will also remove the trailing empty token. We can insert the `""` only in the trailing case, i.e., when completion takes place at the end of the command line (which we know because ` index == args.len()`). But for intermediate completion positions, it is not possible to distinguish between `jj cat -r | m` and `jj cat -r m|` in bash or zsh because both will result in `["jj", "cat", "-r", "m"], index=3`; meanwhile in fish they map to `["jj", "cat", "-r", ""]` and `["jj", "cat", "-r", "m"]`, respectively. I have listed a few more cases in the comments of the `complete_at()` function.

As I said, it is a shortcoming in how `clap_complete` drives the completer. In bash, they only forward `COMP_WORDS` (the arguments) and `COMP_CWORD` (the index) but there are [more variables](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-COMP_005fCWORD) which could be used to differentiate the two cases. I assume the situation is similar for zsh.

cli/src/cli_util.rs
jgreitemann: Yikes. Is it ok to just slap `unsafe` on there? What is the policy regarding `unsafe` code? Only `jj-lib` has `#![forbid(unsafe_code)]`, but I've seen very little `unsafe` in the CLI, too. Is it considered good practice to preemptively aim for the new edition? 

It is a fair bet that the program is still single-threaded at this point but none of `handle_shell_completion()`, `run_internal()`, and `run()` can actually guarantee that, only `main()` can. So technically, `main` would need to have the `unsafe {...}` block and the three functions down the stack would all themselves be `unsafe`. But I guess that would be a little overkill for a niche feature like shell completion.

I see the following escape hatch: after expanding the arguments, `jj` could shell out to itself, provide itself with the expanded command line and adjusted `_CLAP_COMPLETE_INDEX` env var _and_ an additional `_JJ_NO_EXPAND_ALIASES` env var which then bypasses argument expansion. We could even do that if and only if the index has changed, so we'd only pay the additional overhead of spawning yet another process when needed. 

> It would be nice if clap provided saner API.

:100: It's totally weird that they allow passing the arguments explicitly but then still read the rest from the environment.

cli/tests/common/test_environment.rs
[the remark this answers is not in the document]
yuja: I prefer a free function, but I'm not strongly against the method syntax. It would look a bit weird, though.

cli/src/cli_util.rs
yuja: I think `unsafe` block with comment is okay here. AFAIK, `set_env()` can go wrong only if the environment is accessed by threaded, non-Rust code, or something like that.

I hope this is a temporary hack.

cli/tests/test_completion.rs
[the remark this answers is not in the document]
yuja: Thanks for the detailed explanation.

> As I said, it is a shortcoming in how `clap_complete` drives the completer. In bash, they only forward `COMP_WORDS` (the arguments) and `COMP_CWORD` (the index) but there are [more variables](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-COMP_005fCWORD) which could be used to differentiate the two cases. I assume the situation is similar for zsh.

I'm afraid that `clap` will fix the problem in a way that we need more messy workarounds.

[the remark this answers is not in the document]
jgreitemann: I've created https://github.com/clap-rs/clap/issues/5979 for this.

To be fair, the dynamic completion feature is still unstable. I would hope that it eventually supports alias resolution out of the box.

If we're concerned the upstream changes of the "protocol" between glue script and completer might change, perhaps it would be a good idea to also snapshot the completion script for all shells? Currently, the test `test_completions_are_generated` runs only for fish because I thought the point was primarily to test that whatever glue code is correctly printed to stdout when `COMPLETE` is specified without additional arguments.

cli/tests/common/test_environment.rs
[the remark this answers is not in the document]
jgreitemann: I took the liberty to go with the method syntax, then. The free function would have to be "overloaded" on both `TestEnvironment` and `TestWorkDir` which made naming it quite awkward.

cli/tests/test_completion.rs
```diff
@@ -12,9 +12,91 @@
...
+impl TestWorkDir<'_> {
+    /// Runs `jj` as if shell completion had been triggered for the argument of
+    /// the given index.
+    #[must_use = "either snapshot the output or assert the exit status with .success()"]
+    fn complete_at<I>(&self, shell: Shell, index: usize, args: I) -> CommandOutput
+    where
+        I: IntoIterator,
+        I::Item: AsRef<OsStr>,
+    {
+        self.run_jj_with(|cmd| {
```
yuja, line 47: nit: insert `assert!(index + 1 < args.len())` for sanity check?

cli/tests/test_completion.rs
```diff
@@ -78,15 +160,12 @@ fn test_bookmark_names() {
...
-    let mut test_env = test_env;
     // Every shell hook is a little different, e.g. the zsh hooks add some
     // additional environment variables. But this is irrelevant for the purpose
     // of testing our own logic, so it's fine to test a single shell only.
-    test_env.add_env_var("COMPLETE", "fish");
-    let test_env = test_env;
     let work_dir = test_env.work_dir("repo");
 
-    let output = work_dir.run_jj(["--", "jj", "bookmark", "rename", ""]);
+    let output = work_dir.complete_at(Shell::Fish, 3, &["jj", "bookmark", "rename", ""]);
```
yuja, line 168: nits:
* `&` can be omitted
* might be good to omit `"jj", ` too.
* add `.complete(shell, args)` (without `index`) or `.complete_fish(args)` for convenience?

cli/tests/test_completion.rs
```diff
@@ -227,22 +301,28 @@ fn test_aliases_are_resolved() {
...
-    let output = work_dir.run_jj(["--", "jj", "b2", "rename", "a"]);
-    insta::assert_snapshot!(output, @r"
-    aaa	(no description set)
-    [EOF]
-    ");
+    insta::allow_duplicates! {
+        for output in [
+            work_dir.complete_at(shell, 3, &["jj", "b", "rename", "a"]),
+            work_dir.complete_at(shell, 3, &["jj", "b2", "rename", "a"]),
+        ] {
```
yuja, line 309: nit: just flatten loop instead of using `allow_duplicates`? It's harder to debug if one of these samples failed.

cli/src/cli_util.rs
```diff
@@ -3551,29 +3551,55 @@ fn handle_shell_completion(
...
+                env::set_var(
+                    "_CLAP_COMPLETE_INDEX",
+                    (expanded_args.len() - 1).to_string(),
+                );
+            }
+
+            // Remove extra padding again to align with clap_complete's expectations for
+            // zsh.
+            let split_off_padding = expanded_args.split_off(expanded_args.len() - pad_len);
+            debug_assert!(
```
yuja, line 3594: nit: just `assert!()`? The padding should be short.

yuja (commented):
The implementation looks good, thanks.

cli/tests/test_completion.rs
jgreitemann: `&` was actually intentional to avoid monomorphizing the generic for each length, but doing so would throw the design as a generic API as a whole into question, so I've changed it now to pass the arrays by value. Might be interesting to measure compile times...

cli/tests/test_completion.rs
```diff
@@ -12,9 +12,122 @@
...
+        I: IntoIterator,
+        I::IntoIter: ExactSizeIterator,
+        I::Item: AsRef<OsStr>,
+    {
+        self.work_dir("").complete_at(shell, index, args)
+    }
+
+    /// Run `jj` as if fish shell completion had been triggered at the last item
+    /// in `args`.
+    fn complete_fish<I>(&self, args: I) -> CommandOutput
```
yuja, line 40: nit: add `#[must_use = ..` here

cli/tests/test_completion.rs
```diff
@@ -12,9 +12,122 @@
...
+                Shell::Fish => cmd.args(args.take(index)),
+
+                _ => todo!("{shell} completion behavior not implemented yet"),
+            }
+        })
+    }
+
+    /// Run `jj` as if fish shell completion had been triggered at the last item
+    /// in `args`.
+    fn complete_fish<I>(&self, args: I) -> CommandOutput
```
yuja, line 120: nit: add `#[must_use = ..`

cli/tests/test_completion.rs
```diff
@@ -370,42 +549,40 @@ fn test_aliases_are_completed() {
...
-    let output = work_dir.run_jj(["--", "jj", "user-al"]);
-    insta::assert_snapshot!(output, @r"
-    user-alias
-    [EOF]
-    ");
+    let output = work_dir.complete_at(shell, 1, ["user-al"]);
+    match shell {
+        Shell::Bash | Shell::Zsh => {
+            insta::assert_snapshot!(output, @"user-alias[EOF]");
+        }
```
yuja, line 672: either wrap in `insta::allow_duplicates! { .. }` or split match arms.

(you'll need to run tests without using `nextest` to detect this kind of problems.)

cli/tests/test_completion.rs
```diff
@@ -404,6 +404,99 @@ fn test_completions_are_generated() {
...
+        .take_stdout_n_lines(2);
+    match shell {
+        Shell::Bash => {
+            insta::assert_snapshot!(output, @r"
+            --revisions
+            --limit
+            [EOF]
+            ");
+        }
+        Shell::Zsh => {
```
yuja, line 424: Maybe these tests were snapshotted after the fix? To demonstrate the bug, you'll need to run tests at this revision. You can also squash the added tests to the last patch.

jgreitemann: I created the snapshots originally by "manually" resolving the aliases. The tests were intentionally failing in those two revisions. Personally, I don't mind that as it is then easier to prove that the tests are actually sensitive to the bug that's fixed in a later commit. But I also understand the desire to have each revision pass CI individually. I've squashed them, as you've suggested.

cli/tests/test_completion.rs
```diff
@@ -227,22 +326,39 @@ fn test_aliases_are_resolved() {
...
+            insta::assert_snapshot!(output, @"aaa[EOF]");
+        }
+        Shell::Zsh => {
+            insta::assert_snapshot!(output, @"aaa:(no description set)[EOF]");
+        }
+        Shell::Fish => {
+            insta::assert_snapshot!(output, @r"
+                    aaa	(no description set)
+                    [EOF]
+                    ");
```
yuja, line 342: nit: rerun `insta test --force-update-snapshots` or something to fix weird indent

yuja (approved):
Thanks!

cli/tests/test_completion.rs
jgreitemann: Done. Thanks!

cli/tests/test_completion.rs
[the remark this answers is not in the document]
mernen: FYI, clap-rs/clap#5979 is fixed in `clap_complete` v4.5.49, and @yuja's scenario works as expected. No changes will be necessary on jj's end.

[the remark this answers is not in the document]
jgreitemann: Thanks a bunch for addressing this so quickly, @mernen.

I've given v4.5.50 a quick try and it all works perfectly in bash now. However, in zsh the situation has not changed which is no surprise given that your https://github.com/clap-rs/clap/pull/5985 only fixed bash AFAICT. Do you think we should re-open or clone the clap issue to track the zsh case?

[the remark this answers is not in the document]
mernen: My bad, I completely missed the "and zsh" part; I haven't looked into zsh completion yet, so I can't write a patch at the moment. IMO the issue deserves to be reopened.

PR #6415, opened 2025-04-25
docs conflicts.md: fix and clarify the revert example

ilyagr:


docs/technical/conflicts.md
```diff
@@ -58,8 +58,9 @@ conflicts.
 
 As another example, let's go through what happens when you back out a conflicted
 commit. Let's say we have the usual `C+(B-A)` conflict on top of non-conflict
-state C. We then revert that change. Reverting a change means applying its
-reverse diff, so the result is `(C+(B-A))+(C-(C+(B-A)))`, which we can simplify
-to just `C` (i.e. no conflict).
+state C. We then revert that change. The diff of the conflicted commit is then
+`(C+(B-A))-C = B-A`. Reverting a change means applying its reverse diff, `A-B`
+in this case. The result is `(C+(B-A))+(A-B)`, which we can simplify to just `C`
+(i.e. no conflict).
```
martinvonz, line 64: This is not as close to how it's implemented because we don't simplify the diff before we apply it. I like the old form better, but I'm okay with the new form too if you prefer.

martinvonz (approved):

docs/technical/conflicts.md
ilyagr: This is an interesting point. I think it's especially confusing because this distinction wouldn't matter if the rules described in this doc were the only rules.

I added a footnote trying to explain this in a new commit. WDYT? #6076 is also related.

We can always only merge the first commit in this PR and discuss anything further later.

yuja: Maybe we can instead explain non-conflict example first (because it's identical to the conflict case in symbolic notation), then explain the implementation detail?

```
X  <- revert X
|
C

X+(C-X)=C

when X=C+(B-A):
(C+(B-A))+(C-(C+(B-A))  # flattened and each term pair is simplified
```

ilyagr: I think it's useful to explain the theoretical simplification of `C - (C-(B-A))`, since it points out that you're actually still dealing with the same `(B-A)` diff you originally rebased onto C, at least in theory.

So, if the goal is to remove that part, I'm mildly against it, though I'd be fine with putting what actually happens in the main text and the theoretical simplification in a footnote/parentheses, or restructuring things in some other way.

I think adding a non-conflicted example might make sense, but given the above, I'm not sure whether it's necessary. It could be a separate PR.

ilyagr: As an aside, I have no problem leaving people unsatisfied that `jj` doesn't perfectly follow its own theory, because that is how I feel. Our theory is good, especially in practice, but definitely flawed; perhaps if people are unsatisfied enough, somebody will come up with a better one. :)

joyously: I have a hard time reading this notation. It could be because it mixes arithmetic and graph symbols. In the example graph of C is the parent of X and you want to revert X, the X is the name of the result, not the thing that is "added" or "subtracted". My first impulse seeing`X+(C-X)=C` is to ask "Why start with X?", but looking at it some more, I can see that is the node to affect. But my brain expected something like `C+diff=X`, so to revert X, you simply subtract diff.

yuja: As a theoretical explanation, isn't `X+(C-X)=X` easier to follow? It's not important whether the `X` is conflicted tree or not. And we don't have to think about diff-level simplification, which differs from the implementation anyway.

yuja: Something like this:
```
As another example, let's go through what happens when you back out a conflicted
commit. Let's say we have the usual `E = C+(B-A)` conflict on top of
non-conflict state `C`. We then revert that change. Reverting a change means
applying its reverse diff `-(E-C)`, so the result is `E+(C-E) =
(C+(B-A))+(C-(C+(B-A)))`, which we can simplify to just `C` (i.e. no conflict).
```

I use `E` because the tree `D` is already defined.

ilyagr: Thanks! I think this is an uncontroversial addition to the smallest bugfix. It doesn't explain some of the things I wanted to explain, but they can be separated to a separate draft PR, as I planned before.

So, I'll make this PR consist of my commit and Yuya's, which both seem like straightforward improvements.

ilyagr: I'll wait for 15 hours or so before merging, just in case.

For the other things I wanted to do, we can discuss them in #6416.

martinvonz (approved):

PR #6419, opened 2025-04-26
templates: implement `Commit.trailers()`

mateusauler:
A couple of things I'm unsure of:
- ~Where should I create tests for this?~ (nvm, found it)
- Should `key()` and `value()` be the `Trailer` method names?

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj git fetch` now warns instead of erroring for unknown `git.fetch` remotes
   if other remotes are available.
 
+* Commit objects in templates now have `trailers() -> List<Trailer>`, the Trailer
+  objects have `key() -> String` and `value() -> String`.
+
 ### Fixed bugs
 
 * Fixed crash on change-delete conflict resolution.
--- a/cli/src/commit_templater.rs
+++ b/cli/src/commit_templater.rs
@@ -64,6 +64,8 @@ use jj_lib::signing::SignError;
 use jj_lib::signing::SignResult;
 use jj_lib::signing::Verification;
 use jj_lib::store::Store;
+use jj_lib::trailer;
+use jj_lib::trailer::Trailer;
 use once_cell::unsync::OnceCell;
 use pollster::FutureExt as _;
 
@@ -332,6 +334,23 @@ impl<'repo> TemplateLanguage<'repo> for CommitTemplateLanguage<'repo> {
                 let build = template_parser::lookup_method(type_name, table, function)?;
                 build(self, diagnostics, build_ctx, property, function)
             }
+            CommitTemplatePropertyKind::Trailer(property) => {
+                let table = &self.build_fn_table.trailer_methods;
+                let build = template_parser::lookup_method(type_name, table, function)?;
+                build(self, diagnostics, build_ctx, property, function)
+            }
+            CommitTemplatePropertyKind::TrailerList(property) => {
+                // TODO: migrate to table?
+                template_builder::build_formattable_list_method(
+                    self,
+                    diagnostics,
+                    build_ctx,
+                    property,
+                    function,
+                    Self::wrap_trailer,
+                    Self::wrap_trailer_list,
+                )
+            }
         }
     }
 }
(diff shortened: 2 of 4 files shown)


cli/src/commit_templater.rs
```diff
@@ -479,6 +510,8 @@ pub enum CommitTemplatePropertyKind<'repo> {
         Box<dyn TemplateProperty<Output = Option<CryptographicSignature>> + 'repo>,
     ),
     AnnotationLine(Box<dyn TemplateProperty<Output = AnnotationLine> + 'repo>),
+    Trailer(Box<dyn TemplateProperty<Output = Trailer> + 'repo>),
+    Trailers(Box<dyn TemplateProperty<Output = Vec<Trailer>> + 'repo>),
```
yuja, line 514: nit: `TrailerList` (and `trailer_list`) per convention

cli/src/commit_templater.rs
```diff
@@ -543,6 +578,8 @@ impl<'repo> IntoTemplateProperty<'repo> for CommitTemplatePropertyKind<'repo> {
                 Some(Box::new(property.map(|sig| sig.is_some())))
             }
             CommitTemplatePropertyKind::AnnotationLine(_) => None,
+            CommitTemplatePropertyKind::Trailer(_) => None,
+            CommitTemplatePropertyKind::Trailers(_) => None,
```
yuja, line 582: List types support implicit conversion to bool

cli/src/commit_templater.rs
```diff
@@ -1344,6 +1401,18 @@ impl Template for Vec<Rc<CommitRef>> {
...
+        write!(formatter, "{}: {}", self.key, self.value)
+    }
+}
+
+impl Template for Vec<Trailer> {
+    fn format(&self, formatter: &mut TemplateFormatter) -> io::Result<()> {
+        templater::format_joined(formatter, self, "\n")
+    }
+}
+
```
yuja, line 1415: nit: can you move these closer to `builtin_trailer_methods()`?

cli/src/commit_templater.rs
```diff
@@ -2264,3 +2333,25 @@ pub fn builtin_annotation_line_methods<'repo>(
     );
     map
 }
+
+pub fn builtin_trailer_methods<'repo>() -> CommitTemplateBuildMethodFnMap<'repo, Trailer> {
```
yuja, line 2337: nit: remove `pub` (appears that some existing functions unintentionally have `pub`)

yuja (commented):
The code looks good to me overall, thanks.

Can you squash test and doc changes?
https://jj-vcs.github.io/jj/latest/contributing/#commit-guidelines

cli/src/commit_templater.rs
mateusauler: should I remove the unnecessary `pub`s from the other functions?

cli/src/commit_templater.rs
mateusauler: should I also change the exposed method name?

CHANGELOG.md
yuja, line 1: > Can you squash test and doc changes?
> https://jj-vcs.github.io/jj/latest/contributing/#commit-guidelines

Oops, I meant squash test/doc changes into the implementation commit.

"Include tests and documentation in the same commit as the code they test and document."

cli/src/commit_templater.rs
yuja: https://github.com/jj-vcs/jj/pull/6420

cli/src/commit_templater.rs
yuja: > should I also change the exposed method name?

You mean `commit.trailers()` to `.trailer_list()`? No. It's not a function/enum that processes an object of certain template type.

yuja (approved):

glehmann:
I'm not sure we have everything necessary to check that a trailer key is defined in the trailers, as required for #6402.

We could have a `contains()` method in `List`:  `trailers.map(|t| b.key()).contains("Change-id")`
Or something more direct: `trailers.contains_key("Change-id")`, if it's possible to have a method specifically for the `List<Trailer>` type.
Or any better idea :-)

CHANGELOG.md
mateusauler: I read that exact phrase and ignored the second half :/ sorry about that!

mateusauler:
> We could have a `contains()` method in `List`: `trailers.map(|t| b.key()).contains("Change-id")` Or something more direct: `trailers.contains_key("Change-id")`, if it's possible to have a method specifically for the `List<Trailer>` type. Or any better idea :-)

`!self.trailers().filter(|t| t.key() == "Change-Id")` works but is a bit janky.
I agree that `List` could have more methods. But at first glance, it requires some deeper modifications, mainly because `ListTemplate` doesn't implement `TemplateProperty`.

glehmann:
It works. For better readability and simpler syntax, I would prefer something like `commit.contains_trailer(key)`.

PR #6422, opened 2025-04-28
completion: revset expression completer

mernen:
This is a work-in-progress, but the basic foundation is laid out. I'd love some feedback if the approach I took is acceptable — I'm not particularly fond of how I implemented `find_trailing_partial_name()`, but since the Pest grammar can only do lexing and parsing at once and we want to deal with incomplete (i.e. potentially invalid/unbalanced) expressions, it seemed like the least hacky solution.

The work is split into multiple commits for ease of review in this draft.

The main goal here is to allow revset completion in more complex expressions — for example, being able to type `jj log -r first-bookmark..secon`&lt;TAB> and getting `second-bookmark` filled in. It will also work for significantly more complex expressions, with a few important caveats:

- `foo::bar` and `all:foo` won't work in Bash by default, due to `$COMP_WORDBREAKS` including `:`. The recommended solution is apparently to consider this a mistake and remove it globally in your `.bashrc`:

      COMP_WORDBREAKS=${COMP_WORDBREAKS//:}

  (or you can just run it in your current shell to try it.) This can, however, be worked around at the completion script, and I'll see if the `clap` folks are interested in building the workaround into `clap_complete`. It works out of the box in fish and zsh, though.
- Expressions prefixed with quotes work in bash, but break in fish and don't complete at all in zsh. I'm not familiar enough with these shells to know what can be done.

# Checklist

If applicable:


yuja:
> I'm not particularly fond of how I implemented `find_trailing_partial_name()`,

Since we don't handle quoted strings, etc., maybe we can simply split the last part of an input string matching certain character ranges (e.g. ascii alphanumeric, `-`, `>= 0x80`, ..)? I don't think we need to use `pest` to do lax tokenization.

mernen:
> Since we don't handle quoted strings, etc., maybe we can simply split the last part of an input string matching certain character ranges (e.g. ascii alphanumeric, `-`, `>= 0x80`, ..)? I don't think we need to use `pest` to do lax tokenization.

My big concern with that was the use of `XID_CONTINUE`, as I couldn't find an easy way to replicate it without invoking Pest. Turns out `regex` supports `\p{XID_CONTINUE}`, so we don't have to compromise on Unicode support! I'll update the patch later today.

mernen:
Other than that, any comments? Only this I consider missing: a few more tests and applying `revset_expression_multi_*` to more commands. The list for `revset_expression_default_single_*` should be complete now.

yuja:
> My big concern with that was the use of `XID_CONTINUE`, as I couldn't find an easy way to replicate it without invoking Pest. Turns out `regex` supports `\p{XID_CONTINUE}`, so we don't have to compromise on Unicode support!

I think it's okay for completer to accept any non-ASCII chars, but it's also good to use regex.

> Other than that, any comments?

I haven't reviewed the PR thoroughly. I have no idea why we need new `_multi` functions, but other than that, the change looks straightforward.

mernen:
> I have no idea why we need new `_multi` functions, but other than that, the change looks straightforward.

I made a distinction between arguments that return a single revision and multiple revisions, but perhaps that's just overcomplicating things. We could reduce everything to `revset_expression_{all,mutable}`.

cli/src/complete.rs
```diff
@@ -329,35 +333,116 @@ fn revisions(revisions: Option<&str>) -> Vec<CompletionCandidate> {
...
+            .collect()
+    }
+}
+
+fn strip_shell_quote(current: &str) -> &str {
+    current
+        .strip_prefix('\'')
+        .or_else(|| current.strip_prefix('"'))
+        .unwrap_or(current)
+}
```
mernen, line 403: Looking more into this, I think this is a mistake on `clap_complete`. I'll submit a patch there so that `current` is always an unquoted string, similar to what shell completion functions receive.

yuja:
> I made a distinction between arguments that return a single revision and multiple revisions, but perhaps that's just overcomplicating things. We could reduce everything to `revset_expression_{all,mutable}`.

Yeah. Even if the computed result has to be a single revision, the user might want to feed multiple symbols into it (e.g `latest(foo|bar|_`).

mernen:
After a few more changes, I think this is ready. The implementation is simpler now, only touching the completion-related code, and treating all revset expressions equally.

cli/src/complete.rs
```diff
@@ -241,6 +241,7 @@ fn revisions(revisions: Option<&str>) -> Vec<CompletionCandidate> {
         const REMOTE_BOOKMARK: usize = 5;
         const REVSET_ALIAS: usize = 6;
 
+        let match_pattern = (!match_prefix.is_empty()).then(|| format!("glob:{match_prefix}*"));
```
yuja, line 244: maybe better to do `glob::Pattern::escape`?

(I guess glob meta characters wouldn't match the regex pattern, but relying on that would be a source of future bugs.)

cli/src/complete.rs
```diff
@@ -329,35 +332,88 @@ fn revisions(revisions: Option<&str>) -> Vec<CompletionCandidate> {
...
+/// If the expression ends with an name or a partial name, returns a tuple that
+/// splits the string at the point the name starts.
+/// If the expression is empty or ends with a prefix or infix operator that
+/// could plausibly be followed by a name, returns a tuple where the first
+/// item is the entire input string, and the second item is empty.
+/// Otherwise, returns `None`.
+///
+/// The input expression may be incomplete (e.g. missing closing parentheses),
+/// and the ability to reject invalid expressions is limited.
+pub fn split_revset_trailing_name(incomplete_revset_str: &str) -> Option<(&str, &str)> {
```
yuja, line 407: nit: remove `pub`?

cli/src/complete.rs
```diff
@@ -329,35 +332,88 @@ fn revisions(revisions: Option<&str>) -> Vec<CompletionCandidate> {
...
+    if incomplete_revset_str.trim_ascii().is_empty() {
+        return Some((incomplete_revset_str, ""));
+    }
+    let re = regex::Regex::new(
+        r"(?:^|[:~\|&\(,]|\.\.)\s*((?:[\p{XID_CONTINUE}_/]+[@.+-])*[\p{XID_CONTINUE}_/]*)$",
+    )
+    .unwrap();
```
yuja, line 414: This pattern is hard to read. Can you switch to the verbose `(?x)` syntax?

It might also be better to first do `.rsplit_once()` by a known set of infix/prefix operators.

cli/src/complete.rs
```diff
@@ -557,13 +613,8 @@ pub fn branch_name_equals_any_revision(current: &std::ffi::OsStr) -> Vec<Complet
         // Don't complete branch names since we want to create a new branch
         return Vec::new();
     };
-    all_revisions()
+    revisions(None, revision)
```
yuja, line 616: Perhaps, we can use `revset_expression()` here. The right-hand side is a revset expression.

yuja (commented):
Looks generally good to me, thanks.

cli/src/complete.rs
mernen: `revset_expression()` takes an `&OsStr` as expression, and we already converted to `&str` to perform the split, so I'm not sure how we could do this.

(By the way, another thing is bothering me now: I just noticed that `revisions()` and `revset_expression()` take the same arguments, apart from string type, but the arguments are in different orders!)

cli/src/complete.rs
mernen: I had to do two splits because `..` couldn't fit with the others, but the regex got dramatically simpler in the process, so I suppose the need for `(?x)` is reduced.

mernen:
Just noticed that `example@o` wasn't working, because `jj bookmark list` only matches patterns against the local name of the bookmark (i.e. `jj b l glob:main@*` returns no results). I'm not sure changing `bookmark list` would be desirable, as it could lead to surprising consequences, so I've changed `revisions()` instead to apply a `b l --remote` glob. This is a nontrivial change, so you might want to check that.

yuja:
> Just noticed that `example@o` wasn't working, because `jj bookmark list` only matches patterns against the local name of the bookmark (i.e. `jj b l glob:main@*` returns no results). I'm not sure changing `bookmark list` would be desirable, as it could lead to surprising consequences, so I've changed `revisions()` instead to apply a `b l --remote` glob. This is a nontrivial change, so you might want to check that.

Maybe we can filter the output manually? The output of `jj bookmark list` shouldn't usually be large.

cli/src/complete.rs
```diff
@@ -241,6 +241,8 @@ fn revisions(revisions: Option<&str>) -> Vec<CompletionCandidate> {
         const REMOTE_BOOKMARK: usize = 5;
         const REVSET_ALIAS: usize = 6;
 
+        let match_pattern = (!match_prefix.is_empty())
+            .then(|| format!("glob:{}*", glob::Pattern::escape(match_prefix)));
```
yuja, line 245: nit: update `tag list` to do `.filter(..)` consistently, and remove `match_pattern`

cli/src/complete.rs
```diff
@@ -329,35 +336,92 @@ fn revisions(revisions: Option<&str>) -> Vec<CompletionCandidate> {
...
+fn split_revset_trailing_name(incomplete_revset_str: &str) -> Option<(&str, &str)> {
+    let final_part = incomplete_revset_str
+        .rsplit_once([':', '~', '|', '&', '(', ','])
+        .map(|(_, rest)| rest)
+        .unwrap_or(incomplete_revset_str);
+    let final_part = final_part
+        .rsplit_once("..")
+        .map(|(_, rest)| rest)
+        .unwrap_or(final_part)
+        .trim_start();
```
yuja, line 420: nit: `.trim_ascii_start()`

cli/src/complete.rs
```diff
@@ -557,13 +621,8 @@ pub fn branch_name_equals_any_revision(current: &std::ffi::OsStr) -> Vec<Complet
         // Don't complete branch names since we want to create a new branch
         return Vec::new();
     };
-    all_revisions()
+    revisions(revision, None)
```
yuja, line 624: nit: `revset_expression(revision.as_ref(), None)` ?

cli/src/complete.rs
```diff
@@ -230,7 +230,7 @@ pub fn aliases() -> Vec<CompletionCandidate> {
     })
 }
 
-fn revisions(revisions: Option<&str>) -> Vec<CompletionCandidate> {
+fn revisions(match_prefix: &str, filter: Option<&str>) -> Vec<CompletionCandidate> {
```
yuja, line 233: nit: `filter` is ambiguous in this context. Maybe rename to `revset` or `revset_filter`?

yuja (approved):

cli/src/complete.rs
mernen: I wasn't aware of `AsRef<OsStr>`, fixed now

PR #6423, opened 2025-04-28
CliRunner: replace start hook by dispatch hook

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3656,11 +3656,14 @@ pub struct CliRunner {
     commit_template_extensions: Vec<Arc<dyn CommitTemplateLanguageExtension>>,
     operation_template_extensions: Vec<Arc<dyn OperationTemplateLanguageExtension>>,
     dispatch_fn: CliDispatchFn,
-    start_hook_fns: Vec<CliDispatchFn>,
+    dispatch_hook_fns: Vec<CliDispatchHookFn>,
     process_global_args_fns: Vec<ProcessGlobalArgsFn>,
 }
 
-type CliDispatchFn = Box<dyn FnOnce(&mut Ui, &CommandHelper) -> Result<(), CommandError>>;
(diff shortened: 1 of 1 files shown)


cli/src/cli_util.rs
```diff
@@ -3779,8 +3780,15 @@ impl CliRunner {
         self
     }
 
-    pub fn add_start_hook(mut self, start_hook_fn: CliDispatchFn) -> Self {
-        self.start_hook_fns.push(start_hook_fn);
+    /// Add a hook that gets called when it's time to run the command. It is
+    /// the hook's responsibility to call the given inner dispatch function to
+    /// run the command.
+    pub fn add_dispatch_hook(mut self, dispatch_hook_fn: CliDispatchHookFn) -> Self {
```
yuja, line 3786: nit: `dispatch_hook_fn` doesn't have to be `Box<dyn ..>`.

And I think `.add_subcommand()` can be forwarded to `.add_dispatch_hook()` as an example. (This means the order of `.add_dispatch_hook()` and `.add_subcommand()` is important?)

cli/src/cli_util.rs
```diff
@@ -3940,8 +3949,12 @@ impl CliRunner {
...
             data: Rc::new(command_helper_data),
         };
-        for start_hook_fn in self.start_hook_fns {
-            start_hook_fn(ui, &command_helper)?;
+        for dispatch_hook_fn in self.dispatch_hook_fn {
+            let old_dispatch_fn = self.dispatch_fn;
+            let new_dispatch_fn = move |ui: &mut Ui, command_helper: &CommandHelper| {
+                dispatch_hook_fn(ui, command_helper, old_dispatch_fn)
+            };
+            self.dispatch_fn = Box::new(new_dispatch_fn);
```
yuja, line 3957: nit: I think it's simpler to move `self.dispatch_fn` to local variable. We can also use `.into_iter().fold()`.

yuja (approved):

cli/src/cli_util.rs
martinvonz: Done

PR #6425, opened 2025-04-28
templates: add commit.contains_trailer(key)

glehmann:
as a simpler and more readable alternative to

    self.trailers().filter(|t| t.key() == "Change-Id")

if that's ok to add a method to `Commit`.

# Checklist

If applicable:


yuja:
I think it's better to add `.trailers().contains(key)` or `.contains_key(key)`. We might also need `.trailers().get(key) -> Option<String>` or something to render a certain trailer value, and it would be a mess if we had to add another `commit.<verb>_trailer()` function.

Maybe `Vec<Trailers>` can be generalized as `Map` or `MultiMap` template type if needed.

mateusauler:
How about `trailers().contains_key(key)`, `trailers().contains(key, value)` and maybe `trailers().contains_value(value)`?
Although I agree with changing it to a `Map`/`MultiMap`.

joyously:
If the key and the value are always String, is there a need to reinvent the String functions?

glehmann:
I agree this doesn't scale well with other trailer-related features we may add.

@joyously not sure why you are talking about reinventing the String functions. We are rather introducing some Map functions.
The trailers should be best modeled as an ordered multi map. Whatever solution is chosen, adding it in the template engine looks like quite a lot more work than what was necessary for this PR, though.

glehmann:
> Whatever solution is chosen, adding it in the template engine looks like quite a lot more work than what was necessary for this PR, though.

I was mistaken; moving the function to `List<Trailer>` proved to be quite simple.

docs/templates.md
```diff
@@ -254,6 +254,13 @@ The following methods are defined. See also the `List` type.
 
 * `.join(separator: Template) -> Template`
 
+### List<Trailer> type
```
yuja, line 257: nit: move before `ListTemplate` because `List<Trailer>` is a `List<_>` type, and this section only lists additional methods.

docs/templates.md
```diff
@@ -254,6 +254,13 @@ The following methods are defined. See also the `List` type.
...
+
+The following methods are defined. See also the `List` type.
+
+* `.contains_key(key: String) -> Boolean`: True if the commit description
```
yuja, line 261: nit: `key: Template` (`expect_plain_text_expression()` accepts any printable expression.)

yuja (approved):
Looks good to me, thanks.

PR #6431, opened 2025-05-01
restore: do not output "Created ..." message

hotsphink:

### DIFF
--- a/cli/src/commands/diffedit.rs
+++ b/cli/src/commands/diffedit.rs
@@ -136,8 +136,7 @@ don't make any changes, then the operation will be aborted.",
     if tree_id == *target_commit.tree_id() {
         writeln!(ui.status(), "Nothing changed.")?;
     } else {
-        let new_commit = tx
-            .repo_mut()
+        tx.repo_mut()
             .rewrite_commit(&target_commit)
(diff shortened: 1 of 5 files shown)


martinvonz:
Thanks! I noticed that `diffedit` seems to print the same message. Feel like removing that too? It seems like it's equally useless there.

hotsphink:
Note that I changed the commit message header from "restore:" to "restore, diffedit:". Not sure if there's a better way.

hotsphink:
Oh right, tests are a thing. Hey, this `cargo insta review` workflow is nice! Though it seems to do a few failures at a time... maybe `--all`?

hotsphink:
Nope, it was `cargo test --no-fail-fast`. I think? I only had 1 failure left by the time I ran that.

Plus two failures that still remain that seem unrelated: test_gpg::gpgsm_signing_roundtrip and test_gpg::gpgsm_signing_roundtrip_explicit_key. I'll re-push and see what happens.

hotsphink:
...and I finally read the contributor guidelines, which would have told me.

martinvonz (approved):
Thanks!

thoughtpolice (approved):

PR #6434, opened 2025-05-02
docs cli reference: update and pin clap-markdown, include aliases in reference

ilyagr:
clap-markdown doesn't follow semver, so I pinned the version exactly.

If we don't like the changes, we could pin the older versions instead. The main goal is so to prevent this from messing up dependabot PRs.


martinvonz (approved):
Makes sense. Thanks!

PR #6442, opened 2025-05-03
Executable config PR 1 - Small fixes

wrzian:
As requested, I'm splitting up the commits in the executable bit PR into smaller stacked PRs so they can be reviewed separately.

See https://github.com/jj-vcs/jj/pull/4478#pullrequestreview-2746646793 for context.

---

This PR is focused on commits that fix some small issues I encountered which should be trivial to review.

### DIFF
--- a/cli/src/commands/debug/local_working_copy.rs
+++ b/cli/src/commands/debug/local_working_copy.rs
@@ -37,7 +37,7 @@ pub fn cmd_debug_local_working_copy(
     let wc = check_local_disk_wc(workspace_command.working_copy().as_any())?;
     writeln!(ui.stdout(), "Current operation: {:?}", wc.operation_id())?;
     writeln!(ui.stdout(), "Current tree: {:?}", wc.tree_id()?)?;
-    for (file, state) in wc.file_states()? {
+    for (file, state) in wc.file_states()?.iter() {
         writeln!(
             ui.stdout(),
             "{:?} {:13?} {:10?} {:?} {:?}",
--- a/cli/src/description_util.rs
+++ b/cli/src/description_util.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::fmt::Write as _;
 use std::fs;
 use std::io;
 use std::io::Write as _;
@@ -349,11 +350,13 @@ pub fn combine_messages_for_editing(
             .into_string()
             .map_err(|_| user_error("Trailers should be valid utf-8"))?;
         let new_trailers = parse_trailers(&trailer_lines)?;
-        let trailers: String = new_trailers
+        let trailers = new_trailers
             .iter()
             .filter(|trailer| !old_trailers.contains(trailer))
-            .map(|trailer| format!("{}: {}\n", trailer.key, trailer.value))
-            .collect();
+            .fold(String::new(), |mut output, trailer| {
+                writeln!(output, "{}: {}", trailer.key, trailer.value).unwrap();
+                output
+            });
         if !trailers.is_empty() {
             combined.push_str("\nJJ: Trailers not found in the squashed commits:\n");
             combined.push_str(&trailers);
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -616,7 +616,7 @@ fn remove_old_file(disk_path: &Path) -> Result<bool, CheckoutError> {
     match fs::remove_file(disk_path) {
         Ok(()) => Ok(true),
         Err(err) if err.kind() == io::ErrorKind::NotFound => Ok(false),
-        // TODO: Use io::ErrorKind::IsADirectory if it gets stabilized
+        Err(err) if err.kind() == io::ErrorKind::IsADirectory => Ok(false),
         Err(_) if disk_path.symlink_metadata().is_ok_and(|m| m.is_dir()) => Ok(false),
         Err(err) => Err(CheckoutError::Other {
             message: format!("Failed to remove file {}", disk_path.display()),
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -418,23 +418,6 @@ impl<'a> FileStates<'a> {
     }
 }
 
-pub struct TreeState {
-    store: Arc<Store>,
-    working_copy_path: PathBuf,
-    state_path: PathBuf,
-    tree_id: MergedTreeId,
-    file_states: FileStatesMap,
-    // Currently only path prefixes
-    sparse_patterns: Vec<RepoPathBuf>,
-    own_mtime: MillisSinceEpoch,
-    symlink_support: bool,
-
-    /// The most recent clock value returned by Watchman. Will only be set if
-    /// the repo is configured to use the Watchman filesystem monitor and
-    /// Watchman has been queried at least once.
-    watchman_clock: Option<crate::protos::working_copy::WatchmanClock>,
-}
-
 fn file_state_from_proto(proto: &crate::protos::working_copy::FileState) -> FileState {
     let file_type = match proto.file_type() {
         crate::protos::working_copy::FileType::Normal => FileType::Normal {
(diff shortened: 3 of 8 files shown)


lib/src/local_working_copy.rs
```diff
@@ -598,7 +565,7 @@ fn remove_old_file(disk_path: &Path) -> Result<bool, CheckoutError> {
     match fs::remove_file(disk_path) {
         Ok(()) => Ok(true),
         Err(err) if err.kind() == io::ErrorKind::NotFound => Ok(false),
-        // TODO: Use io::ErrorKind::IsADirectory if it gets stabilized
+        Err(err) if err.kind() == io::ErrorKind::IsADirectory => Ok(false),
         Err(_) if disk_path.symlink_metadata().is_ok_and(|m| m.is_dir()) => Ok(false),
```
wrzian, line 569: Removing the second line causes test failures on Windows and Mac, I don't have the energy to really look into those, but the force push was me reverting that change.

ofek:
Thanks for breaking these up! Is the plan to have multiple PRs with the final being the original or would that also be in a new PR?

wrzian:
> Thanks for breaking these up! Is the plan to have multiple PRs with the final being the original or would that also be in a new PR?

Yes. I'm still finalizing a third split PR, but the original will contain the final set of changes and be based off the others.

thoughtpolice:
Can we split the first 3 clippy lint fixes out to another PR, if you don't mind? Those seem like genuine independent fixes and would help anyone developing on Windows, like you. (Actually, I wonder if adding another CI matrix entry to run clippy on all platforms would be worth it; it's fast compared to the whole testsuite so I doubt it would be a bottleneck. We can handle that separately from this, though.)

cli/src/description_util.rs
```diff
@@ -349,11 +350,13 @@ pub fn combine_messages_for_editing(
...
             .map_err(|_| user_error("Trailers should be valid utf-8"))?;
         let new_trailers = parse_trailers(&trailer_lines)?;
-        let trailers: String = new_trailers
+        let trailers = new_trailers
             .iter()
             .filter(|trailer| !old_trailers.contains(trailer))
-            .map(|trailer| format!("{}: {}\n", trailer.key, trailer.value))
-            .collect();
+            .fold(String::new(), |mut output, trailer| {
+                let _ = writeln!(output, "{}: {}", trailer.key, trailer.value);
```
yuja, line 357: nit: `.unwrap()` instead of ignoring error

btw, I think a better option is to filter out redundant `trailers`, and `write!()` into `combined`.

lib/src/local_working_copy.rs
yuja: Can you include the failing outputs (here or as inline comment)?

It's understandable that Windows has weird behavior in general, but I'm not sure why the tests are failing on macOS.

cli/src/commands/debug/local_working_copy.rs
```diff
@@ -37,7 +37,7 @@ pub fn cmd_debug_local_working_copy(
     let wc = check_local_disk_wc(workspace_command.working_copy().as_any())?;
     writeln!(ui.stdout(), "Current operation: {:?}", wc.operation_id())?;
     writeln!(ui.stdout(), "Current tree: {:?}", wc.tree_id()?)?;
-    for (file, state) in wc.file_states()? {
+    for (file, state) in wc.file_states()?.iter() {
```
yuja, line 40: nit: I wouldn't include this change in this PR because I assume this won't be needed if we refactor the executable config change in a way that the underlying data doesn't depend on the config knob.

yuja (commented):
Thanks for splitting PRs!

martinvonz:
Heads up that #6465 may conflict with this PR (I assume without even looking at the code). @yuja, perhaps your PR was even written to make this PR simpler?

yuja:
> Heads up that #6465 may conflict with this PR (I assume without even looking at the code). @yuja, perhaps your PR was even written to make this PR simpler?

It won't conflict with this PR itself, but will conflict with later PRs. I think we'll need something similar to #6465 in order to avoid future bugs, so I sent it out. (My origination motivation was to avoid CI failure because of Windows code.)

cli/src/commands/debug/local_working_copy.rs
wrzian: None of the other changes affect this.

wrzian:
Dude. What the fuck?

lib/src/local_working_copy.rs
wrzian: You can find them here: https://github.com/jj-vcs/jj/actions/runs/14812847212/job/41589698612

Here's one of the outputs from macos:
```
        FAIL [   0.179s] jj-lib::runner test_local_working_copy::test_git_submodule::ignore_all
──── STDOUT:             jj-lib::runner test_local_working_copy::test_git_submodule::ignore_all

running 1 test
test test_local_working_copy::test_git_submodule::ignore_all ... FAILED

failures:

failures:
    test_local_working_copy::test_git_submodule::ignore_all

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 455 filtered out; finished in 0.15s

──── STDERR:             jj-lib::runner test_local_working_copy::test_git_submodule::ignore_all
ignoring git submodule at "sub/module"
ignoring git submodule at "sub/module"
thread 'test_local_working_copy::test_git_submodule::ignore_all' panicked at lib/tests/test_local_working_copy.rs:1440:10:
called `Result::unwrap()` on an `Err` value: Other { message: "Failed to remove file /var/folders/hk/wjxbksl168x1c7_ycgqjys980000gn/T/jj-test-e8PMlC/repo/sub/module", err: Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" } }
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: runner::test_local_working_copy::test_git_submodule
   4: runner::test_local_working_copy::test_git_submodule::ignore_all
   5: runner::test_local_working_copy::test_git_submodule::ignore_all::{{closure}}
   6: core::ops::function::FnOnce::call_once
```

wrzian:
@yuja The changes you made are literally a subset of the changes I was already making but in a slightly different way. This causes a major amount of conflicts to the main executable bit changes that I am now going to have to rewrite.

Did you not think that I might deserve to be pinged on that PR since it directly conflicts with what I'm doing? Maybe request a review from me?

I feel incredibly disrespected.

cli/src/description_util.rs
wrzian: Something like this?

```rust
let new_trailers = parse_trailers(&trailer_lines)?;
let mut once = true;
for trailer in new_trailers {
    if !old_trailers.contains(&trailer) {
        if once {
            combined.push_str("\nJJ: Trailers not found in the squashed commits:\n");
            once = false;
        }
        writeln!(combined, "{}: {}", trailer.key, trailer.value).unwrap();
    }
}
```

wrzian:
I rebased off main, fixed a partial nit in the first commit, and added one more commit at the end with another small change.

> Can we split the first 3 commits with the clippy lint fixes out to another PR, if you don't mind? Those seem like genuine independent fixes and would help anyone developing on Windows, like you.

There's already some review here, and I don't expect any of these commits to take time to review/merge. also is already broken out from another pull request, and I'm running low on energy to complete this.

cli/src/commands/debug/local_working_copy.rs
yuja: Well, I meant we won't need https://github.com/jj-vcs/jj/pull/6442/commits/20d5d1b2d68c394c13fa8e78ebff8793cb3d0fae if we addressed the old review comment in https://github.com/jj-vcs/jj/pull/4478#discussion_r1772775928 and https://github.com/jj-vcs/jj/pull/4478#discussion_r2031155547. `FileType::Normal { ... }` will store `bool` unconditionally, and it's up to caller (of `FileStates`) whether or not ignore the stored `executable: bool` value.

martinvonz:
> @yuja The changes you made are literally a subset of the changes I was already making but in a slightly different way. This causes a major amount of conflicts to the main executable bit changes that I am now going to have to rewrite.

@yuja, should we roll back your PR if it conflicts a lot with @wrzian's changes in a way that his PR will effectively roll back your changes anyway? I feel like we can replace `FileExecutableFlag` by a simple `bool`.

martinvonz:
> > I feel like we can replace `FileExecutableFlag` by a simple `bool`.
> 
> We'll still need a wrapper type `FileExecutableFlag(bool)` to ensure that the `FileState` value is interpreted where necessary. Using `bool` would be a source of bugs because it's easy to copy the value back and forth to `TreeValue`.

I see. Makes sense to me to use a wrapper type to reduce risk of bugs.

@wrzian: Have you tried the approach of having an executable bit on all platforms and just ignoring its value when writing a tree if we're on Windows or if the config has been set to ignore it?

wrzian:
I'm closing this because I'm moving to a new account, @isuffix. I'll restart this PR from that account with more atomic PRs.

PR #6448, opened 2025-05-04
builtin merge-tool: fix edge cases found by property-based testing

### LINKED ISSUE #5189: Panic in built-in diff editor when replacing directory with file
I'm not sure if I should make a new issue, but I got a consistent unreachable panic corner case type of thing that seems related to this (there is a possibility this is precisely what you guys are trying to fix :sweat_smile:)

Basically if you replace a folder with a file (in my case it was a symlink) and then try to split you get a panic.

minimal repro:
```bash
λ jj git init test
Initialized repo in "test"
λ cd test
λ mkdir folder
λ touch folder/.keep
λ jj ci -m 'added folder'
Working copy now at: (x) c76fc22 (empty) (no description set)
Parent commit      : (u) 9092cba added folder
λ rm -r folder
λ touch folder
λ jj split
Hint: Using default editor ':builtin'; run `jj config set --user ui.diff-editor :builtin` to disabl
e this message.
thread 'main' panicked at cli/src/merge_tools/builtin.rs:205:13:
internal error: entered unreachable code: list of changed files included a tree: TreeId("29a422c192
51aeaeb907175e9b3219a9bed6c616")
```

_Originally posted by @necauqua in https://github.com/jj-vcs/jj/issues/3702#issuecomment-2561200255_

On this topic, I've successfully used property-based testing for incremental file changes in a similar context. It would be good to try it here to fuzz for panics, at least.

jgreitemann:
Fixes #5189. As discussed in the issue, I went ahead with property-based testing and followed the state machine approach mentioned by @arxanas. Thanks for the pointers you gave in https://github.com/jj-vcs/jj/issues/5189#issuecomment-2836104306; they were quite helpful to get my head wrapped around it.

You mentioned that it would be nice to create working copy states for other tests. AFAIU the split been the `ReferenceStateMachine` and `StateMachineTest` traits in `proptest-state-machine` is intended to facilitate this kind of reuse of the reference state machine while keeping the test's state separate, so I've created a dedicated `proptest` module in `testutils` containing the data structure implementing `ReferenceStateMachine` the and associated strategies. The `StateMachineTest` implementation in `merge_tools::builtin` then applies the same transitions to its domain by manipulating the "right" `MergedTree`; another implementation could pick a different architectual flight level.

Regarding the state machine modeling:

- The transitions `CreateFile`/`DeleteFile` are implemented in a way that they don't require the path to be vacant/extant. `CreateFile` will overwrite existing files or even directories (as in the case of the issue) and `DeleteFile` is a no-op if the file does not exist. Of course, the strategies will generate the transitions such that `CreateFile` references a random file in an existing directory and `DeleteFile` references an existing file, but collisions might happen and shrinking can make existing files disappear (or even new ones appear, if a prior `DeleteFile` transition has been shrunk away). It is possible to impose preconditions for transitions and I did so at first but I've found that this really impedes the exploration of the state space and leads into dead ends when shrinking, presumably because the initial state can no longer be shrunk independently from the transitions and vice versa. This is a little hard to explain without going too deep; feel free to hit me up on Discord.
  
- I've not found a great way to test for other properties than what is done in the existing manual tests: selecting none of the changes produces the left tree, selecting them all produces the right tree. As mentioned in the discussion of the issue, it might make sense to apply only some changes, produce an intermediate tree, then apply the remaining changes to go all the way to the right tree. Usually, this sort of thing should be a perfect fit for proptesting but I've struggled to reconcile this with the state machine approach.
  
- The present incarnation is pretty limited to the issue at hand. There are several ways to expand on that (some of which you already mentioned). One thing that I _did_ include is file mode (well, at least the executable flag) and this promptly found another issue that is also fixed in this PR (in the last commit). (File mode was not covered by the manual tests so far because `testutils::create_tree` didn't account for it.)
  

The fix for the original issue is in the penultimate `tree_builder: ignore tombstone overrides on nested trees`. I've also added a regular regression test for this particular scenario. (Proptest docs recommend adding regular tests for known regressions as changes in the strategies effectively invalidate the checked-in seeds).

I'm not super happy with the fixes for either of the bugs. In both cases, it is the best thing I could do given the information that scm-record produces. I've highlighted the pain points in the ~commit messages~ change descriptions; perhaps you can judge for yourself whether to take any of that upstream.

# Checklist

If applicable:

### DIFF
--- a/cli/src/merge_tools/builtin.rs
+++ b/cli/src/merge_tools/builtin.rs
@@ -420,6 +420,15 @@ fn apply_changes(
     );
     // TODO: Write files concurrently
     for (path, file) in changed_files.into_iter().zip(files) {
+        let file_mode_change_selected = file
+            .sections
+            .iter()
+            .find_map(|sec| match sec {
+                scm_record::Section::FileMode { is_checked, .. } => Some(*is_checked),
+                _ => None,
+            })
+            .unwrap_or(false);
+
         let (
             scm_record::SelectedChanges {
                 contents,
@@ -428,10 +437,16 @@ fn apply_changes(
             _unselected,
         ) = file.get_selected_contents();
 
-        // If a file was not present in the selected changes (i.e. split out of a
-        // change, deleted, etc.) remove it from the tree.
         if file_mode == scm_record::FileMode::Absent {
-            tree_builder.set_or_remove(path, Merge::absent());
+            // The file is not present in the selected changes.
+            // Either a file mode change was selected to delete an existing file, so we
+            // should remove it from the tree,
+            if file_mode_change_selected {
+                tree_builder.set_or_remove(path, Merge::absent());
+            }
+            // or the file's creation has been split out of the change, in which case we
+            // don't need to change the tree.
+            // In either case, we're done with this file afterwards.
             continue;
         }
 
@@ -639,7 +654,9 @@ mod tests {
     use jj_lib::matchers::EverythingMatcher;
     use jj_lib::merge::MergedTreeValue;
     use jj_lib::repo::Repo as _;
+    use testutils::dump_tree;
     use testutils::repo_path;
+    use testutils::repo_path_component;
     use testutils::TestRepo;
 
     use super::*;
@@ -1250,6 +1267,85 @@ mod tests {
             all_changes_tree.id(),
             right_tree.id(),
             "all-changes tree was different",
+        );
+    }
+
+    #[test]
+    fn test_edit_diff_builtin_replace_directory_with_file() {
+        let test_repo = TestRepo::init();
+        let store = test_repo.repo.store();
+
+        let folder_path = repo_path("folder");
+        let file_in_folder_path = folder_path.join(repo_path_component("file_in_folder"));
+        let left_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder.file(&file_in_folder_path, vec![]);
+        });
+        let right_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder.file(folder_path, vec![]);
+        });
+
+        let (changed_files, files) = make_diff(store, &left_tree, &right_tree);
+        insta::assert_debug_snapshot!(changed_files, @r#"
+        [
+            "folder/file_in_folder",
+            "folder",
+        ]
+        "#);
+        insta::with_settings!({filters => vec![(r"\\\\", "/")]}, {
+            insta::assert_debug_snapshot!(files, @r#"
+            [
+                File {
+                    old_path: None,
+                    path: "folder/file_in_folder",
+                    file_mode: Unix(
+                        33188,
+                    ),
+                    sections: [
+                        FileMode {
+                            is_checked: false,
+                            mode: Absent,
+                        },
+                    ],
+                },
+                File {
+                    old_path: None,
+                    path: "folder",
+                    file_mode: Absent,
+                    sections: [
+                        FileMode {
+                            is_checked: false,
+                            mode: Unix(
+                                33188,
+                            ),
+                        },
+                    ],
+                },
+            ]
+            "#);
+        });
+        let no_changes_tree_id = apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
+        let no_changes_tree = store.get_root_tree(&no_changes_tree_id).unwrap();
+        assert_eq!(
+            left_tree.id(),
+            no_changes_tree.id(),
+            "no-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
+            dump_tree(store, &left_tree.id()),
+            dump_tree(store, &no_changes_tree.id()),
+        );
+
+        let mut files = files;
+        for file in &mut files {
+            file.toggle_all();
+        }
+        let all_changes_tree_id =
+            apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
+        let all_changes_tree = store.get_root_tree(&all_changes_tree_id).unwrap();
+        assert_eq!(
+            right_tree.id(),
+            all_changes_tree.id(),
+            "all-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
+            dump_tree(store, &right_tree.id()),
+            dump_tree(store, &all_changes_tree.id()),
         );
     }
 
--- a/cli/src/merge_tools/builtin.rs
+++ b/cli/src/merge_tools/builtin.rs
@@ -863,17 +863,11 @@ mod tests {
 
         let added_executable_file_path = repo_path("executable_file");
         let left_tree = testutils::create_tree(&test_repo.repo, &[]);
-        let right_tree = {
-            // let store = test_repo.repo.store();
-            let mut tree_builder = store.tree_builder(store.empty_tree_id().clone());
-            testutils::write_executable_file(
-                &mut tree_builder,
-                added_executable_file_path,
-                "executable",
-            );
-            let id = tree_builder.write_tree().unwrap();
-            MergedTree::resolved(store.get_tree(RepoPathBuf::root(), &id).unwrap())
-        };
+        let right_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder
+                .file(added_executable_file_path, "executable")
+                .executable(true);
+        });
 
         let (changed_files, files) = make_diff(store, &left_tree, &right_tree);
         insta::assert_debug_snapshot!(changed_files, @r#"
--- a/cli/src/merge_tools/builtin.rs
+++ b/cli/src/merge_tools/builtin.rs
@@ -341,6 +341,18 @@ async fn make_diff_files(
                 sections.extend(make_diff_sections(&old_contents, &new_contents)?);
             }
 
+            (
+                FileContents::Binary {
+                    hash: Some(left_hash),
+                    ..
+                },
+                FileContents::Binary {
+                    hash: Some(right_hash),
+                    ..
+                },
+            ) if left_hash == right_hash => {
+                // Binary file contents have not changed.
+            }
             (left, right @ FileContents::Binary { .. })
             | (left @ FileContents::Binary { .. }, right) => {
                 sections.push(scm_record::Section::Binary {
--- a/cli/src/merge_tools/builtin.rs
+++ b/cli/src/merge_tools/builtin.rs
@@ -388,6 +388,7 @@ fn apply_diff_builtin(
         &mut tree_builder,
         changed_files,
         files,
+        |path| left_tree.path_value(path),
         |path| right_tree.path_value(path),
         |path, contents, executable| {
             let old_value = left_tree.path_value(path)?;
@@ -422,6 +423,7 @@ fn apply_changes(
     tree_builder: &mut MergedTreeBuilder,
     changed_files: Vec<RepoPathBuf>,
     files: &[scm_record::File],
+    select_left: impl Fn(&RepoPath) -> BackendResult<MergedTreeValue>,
     select_right: impl Fn(&RepoPath) -> BackendResult<MergedTreeValue>,
     write_file: impl Fn(&RepoPath, &[u8], bool) -> BackendResult<MergedTreeValue>,
 ) -> BackendResult<()> {
@@ -449,7 +451,7 @@ fn apply_changes(
             _unselected,
         ) = file.get_selected_contents();
 
-        if file_mode == scm_record::FileMode::Absent {
+        if file_mode == mode::ABSENT {
             // The file is not present in the selected changes.
             // Either a file mode change was selected to delete an existing file, so we
             // should remove it from the tree,
@@ -462,9 +464,16 @@ fn apply_changes(
             continue;
         }
 
+        let executable = file_mode == mode::EXECUTABLE;
         match contents {
             scm_record::SelectedContents::Unchanged => {
-                // Do nothing.
+                if file_mode_change_selected {
+                    // File contents haven't changed, but file mode needs to be updated on the tree.
+                    let value = override_file_executable_bit(select_left(&path)?, executable);
+                    tree_builder.set_or_remove(path, value);
+                } else {
+                    // Neither file mode, nor contents changed => Do nothing.
+                }
             }
             scm_record::SelectedContents::Binary {
                 old_description: _,
@@ -474,13 +483,25 @@ fn apply_changes(
                 tree_builder.set_or_remove(path, value);
             }
             scm_record::SelectedContents::Text { contents } => {
-                let executable = file_mode == mode::EXECUTABLE;
                 let value = write_file(&path, contents.as_bytes(), executable)?;
                 tree_builder.set_or_remove(path, value);
             }
         }
     }
     Ok(())
+}
+
+fn override_file_executable_bit(
+    mut merged_tree_value: MergedTreeValue,
+    new_executable_bit: bool,
+) -> MergedTreeValue {
+    for tree_value in merged_tree_value.iter_mut().flatten() {
+        let TreeValue::File { executable, .. } = tree_value else {
+            panic!("incompatible update: expected a TreeValue::File, got {tree_value:?}");
+        };
+        *executable = new_executable_bit;
+    }
+    merged_tree_value
 }
 
 pub fn edit_diff_builtin(
@@ -648,8 +669,11 @@ pub fn edit_merge_builtin(
             .map(|file| file.repo_path.clone())
             .collect_vec(),
         &state.files,
-        // TODO: It doesn't make sense to select new value from the source tree.
-        // Perhaps, "their" tree value should be extracted from a conflict?
+        |path| tree.path_value(path),
+        // FIXME: It doesn't make sense to select a new value from the source tree.
+        // Presently, `select_right` is never actually called, since it is used to select binary
+        // sections, but `make_merge_file` does not produce `Binary` sections for conflicted files.
+        // This needs to be revisited when the UI becomes capable of representing binary conflicts.
         |path| tree.path_value(path),
         |path, contents, executable| {
             let id = store.write_file(path, &mut &contents[..]).block_on()?;
@@ -949,6 +973,207 @@ mod tests {
             all_changes_tree.id(),
             right_tree.id(),
             "all-changes tree was different",
+        );
+    }
+
+    #[test]
+    fn test_edit_diff_builtin_empty_file_mode_change() {
+        let test_repo = TestRepo::init();
+        let store = test_repo.repo.store();
+
+        let empty_file_path = repo_path("empty_file");
+        let left_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder.file(empty_file_path, vec![]).executable(false);
+        });
+        let right_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder.file(empty_file_path, vec![]).executable(true);
+        });
+
+        let (changed_files, files) = make_diff(store, &left_tree, &right_tree);
+        insta::assert_debug_snapshot!(changed_files, @r#"
+        [
+            "empty_file",
+        ]
+        "#);
+        insta::assert_debug_snapshot!(files, @r#"
+        [
+            File {
+                old_path: None,
+                path: "empty_file",
+                file_mode: Unix(
+                    33188,
+                ),
+                sections: [
+                    FileMode {
+                        is_checked: false,
+                        mode: Unix(
+                            33261,
+                        ),
+                    },
+                ],
+            },
+        ]
+        "#);
+        let no_changes_tree_id = apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
+        let no_changes_tree = store.get_root_tree(&no_changes_tree_id).unwrap();
+        assert_eq!(
+            left_tree.id(),
+            no_changes_tree.id(),
+            "no-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
+            dump_tree(store, &left_tree.id()),
+            dump_tree(store, &no_changes_tree.id()),
+        );
+
+        let mut files = files;
+        for file in &mut files {
+            file.toggle_all();
+        }
+        let all_changes_tree_id =
+            apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
+        let all_changes_tree = store.get_root_tree(&all_changes_tree_id).unwrap();
+        assert_eq!(
+            right_tree.id(),
+            all_changes_tree.id(),
+            "all-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
+            dump_tree(store, &right_tree.id()),
+            dump_tree(store, &all_changes_tree.id()),
+        );
+    }
+
+    #[test]
+    fn test_edit_diff_builtin_text_file_mode_change() {
+        let test_repo = TestRepo::init();
+        let store = test_repo.repo.store();
+
+        let text_file_path = repo_path("text_file");
+        let left_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder.file(text_file_path, "text").executable(false);
+        });
+        let right_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder.file(text_file_path, "text").executable(true);
+        });
+
+        let (changed_files, files) = make_diff(store, &left_tree, &right_tree);
+        insta::assert_debug_snapshot!(changed_files, @r#"
+        [
+            "text_file",
+        ]
+        "#);
+        insta::assert_debug_snapshot!(files, @r#"
+        [
+            File {
+                old_path: None,
+                path: "text_file",
+                file_mode: Unix(
+                    33188,
+                ),
+                sections: [
+                    FileMode {
+                        is_checked: false,
+                        mode: Unix(
+                            33261,
+                        ),
+                    },
+                    Unchanged {
+                        lines: [
+                            "text",
+                        ],
+                    },
+                ],
+            },
+        ]
+        "#);
+        let no_changes_tree_id = apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
+        let no_changes_tree = store.get_root_tree(&no_changes_tree_id).unwrap();
+        assert_eq!(
+            left_tree.id(),
+            no_changes_tree.id(),
+            "no-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
+            dump_tree(store, &left_tree.id()),
+            dump_tree(store, &no_changes_tree.id()),
+        );
+
+        let mut files = files;
+        for file in &mut files {
+            file.toggle_all();
+        }
+        let all_changes_tree_id =
+            apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
+        let all_changes_tree = store.get_root_tree(&all_changes_tree_id).unwrap();
+        assert_eq!(
+            right_tree.id(),
+            all_changes_tree.id(),
+            "all-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
+            dump_tree(store, &right_tree.id()),
+            dump_tree(store, &all_changes_tree.id()),
+        );
+    }
+
+    #[test]
+    fn test_edit_diff_builtin_binary_file_mode_change() {
+        let test_repo = TestRepo::init();
+        let store = test_repo.repo.store();
+
+        let binary_file_path = repo_path("binary_file");
+        let left_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder
+                .file(binary_file_path, vec![0xff, 0x00])
+                .executable(false);
+        });
+        let right_tree = testutils::create_tree_with(&test_repo.repo, |builder| {
+            builder
+                .file(binary_file_path, vec![0xff, 0x00])
+                .executable(true);
+        });
+
+        let (changed_files, files) = make_diff(store, &left_tree, &right_tree);
+        insta::assert_debug_snapshot!(changed_files, @r#"
+        [
+            "binary_file",
+        ]
+        "#);
+        insta::assert_debug_snapshot!(files, @r#"
+        [
+            File {
+                old_path: None,
+                path: "binary_file",
+                file_mode: Unix(
+                    33188,
+                ),
+                sections: [
+                    FileMode {
+                        is_checked: false,
+                        mode: Unix(
+                            33261,
+                        ),
+                    },
+                ],
+            },
+        ]
+        "#);
+        let no_changes_tree_id = apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
+        let no_changes_tree = store.get_root_tree(&no_changes_tree_id).unwrap();
+        assert_eq!(
+            left_tree.id(),
+            no_changes_tree.id(),
+            "no-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
+            dump_tree(store, &left_tree.id()),
+            dump_tree(store, &no_changes_tree.id()),
+        );
+
+        let mut files = files;
+        for file in &mut files {
+            file.toggle_all();
+        }
+        let all_changes_tree_id =
+            apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
+        let all_changes_tree = store.get_root_tree(&all_changes_tree_id).unwrap();
+        assert_eq!(
+            right_tree.id(),
+            all_changes_tree.id(),
+            "all-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
+            dump_tree(store, &right_tree.id()),
+            dump_tree(store, &all_changes_tree.id()),
         );
     }
 
--- a/cli/src/merge_tools/builtin.rs
+++ b/cli/src/merge_tools/builtin.rs
@@ -477,9 +477,17 @@ fn apply_changes(
             }
             scm_record::SelectedContents::Binary {
                 old_description: _,
-                new_description: _,
+                new_description: Some(_),
             } => {
-                let value = select_right(&path)?;
+                let value = override_file_executable_bit(select_right(&path)?, executable);
+                tree_builder.set_or_remove(path, value);
+            }
+            scm_record::SelectedContents::Binary {
+                old_description: _,
+                new_description: None,
+            } => {
+                // File contents emptied out, but file mode is not absent => write empty file.
+                let value = write_file(&path, &[], executable)?;
                 tree_builder.set_or_remove(path, value);
             }
             scm_record::SelectedContents::Text { contents } => {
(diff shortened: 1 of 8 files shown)


lib/src/tree_builder.rs
```diff
@@ -89,9 +89,13 @@ impl TreeBuilder {
...
-                Override::Tombstone => {
-                    tree.remove(basename);
-                }
+                Override::Tombstone => match tree.value(basename) {
+                    Some(TreeValue::Tree(_)) => {}
+                    Some(_) => {
+                        tree.remove(basename);
+                    }
+                    None => {}
+                },
```
yuja, line 98: Can you move this to caller? As you said, I think this is the problem of scm_record. It's better to insert a workaround in the integration layer, not in the lower-level API.

BTW, since this patch fixes the problem, there should be a test change within this patch.

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,322 @@
...
+                        Node::File(_) => None,
+                        Node::Directory(tree) => Some((prefix.join(name), tree)),
+                    })
+                    .flat_map(|(path, tree)| tree.files(&path)),
+            )
+            .collect()
+    }
+}
+
+fn arb_tree() -> impl Strategy<Value = Tree> {
```
arxanas, line 121: suggestion: By default, I recommend implementing `Arbitrary` where possible and relying on type-directed strategy lookup, rather than writing explicit strategies (see also <https://proptest-rs.github.io/proptest/proptest-derive/modifiers.html>). I find it tends to be more ergonomic for writing new property-based tests, since you can just reach for some of the types that you already have on hand.

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,322 @@
...
+    btree_map(arb_path_component(), arb_node(), 1..8).prop_map(|nodes| Tree { nodes })
+}
+
+fn arb_path_component() -> impl Strategy<Value = RepoPathComponentBuf> {
+    // biased towards naming collisions (alpha-delta) but with the option to
+    // generate arbitrary UTF-8
+    "(alpha|beta|gamma|delta|[\\PC&&[^/]]+)".prop_map(|s| RepoPathComponentBuf::new(s).unwrap())
```
arxanas, line 128: nit: I was surprised when I saw `alpha`/`beta`/`gamma`/`delta` in the output, despite having read this code beforehand. I would recommend that you keep it simple and just use `a`/`b`/`c`.

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,322 @@
...
+}
+
+fn arb_node() -> impl Strategy<Value = Node> {
```
arxanas, line 131: suggestion: Unfortunately, I don't think I communicated the design I was proposing well. In this case, I think you should need only one of

- recursive generation
- the state machine or a hand-rolled operation log approach

The search space for the recursive generation should already be handled by the state machine approach (assuming that your transitions are rich enough). Furthermore, it'll be difficult to get the recursive to "look back" and construct values with inter-references between unrelated trees. Therefore I recommend dropping the recursive generation and relying only on the state machine to explore the state space.

I've opened a PR against this one showing one way it could be done: <https://github.com/jj-vcs/jj/pull/6502>.

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,322 @@
...
+            )
+            .collect()
+    }
+}
+
+fn arb_tree() -> impl Strategy<Value = Tree> {
+    btree_map(arb_path_component(), arb_node(), 1..8).prop_map(|nodes| Tree { nodes })
+}
+
+fn arb_path_component() -> impl Strategy<Value = RepoPathComponentBuf> {
```
arxanas, line 125: suggestion (non-blocking): I'd recommend just generating `String`s or `PathBuf`s, and only converting them to `RepoPathComponentBuf`s right before feeding the values into the jj API, just as a matter of convenience, given that `RepoPathComponentBuf` isn't something we're specifically testing.

arxanas (changes requested):
Exciting! Thanks for doing this.

For the benefit of others, I don't think any of the commit messages/comments include an example of what the output looks like, so here's what the failing test looks like (before the fix):

```text
thread 'merge_tools::builtin::tests::test_edit_diff_builtin_proptest' panicked at cli/src/merge_tools/builtin.rs:1425:5:
Test failed: assertion `left == right` failed: all-changes tree was different
  left: Merge(Resolved(TreeId("cce6abbe6e5c05234075")))
 right: Merge(Resolved(TreeId("1131d42426fbb8e2ad12"))).
minimal failing input: (initial_state, transitions, seen_counter) = (
    RepoRefState {
        root: Directory {
            gamma: RegularFile {
                contents: "",
                executable: false,
            },
        },
    },
    [
        CreateFile {
            path: "gamma",
            contents: "",
            executable: true,
        },
    ],
    None,
)
```

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,322 @@
...
+                .boxed(),
+            (_, []) => select(files).boxed(),
+            (_, _) => prop_oneof![
+                select(files),
+                select(subdirs)
+                    .prop_flat_map(|(subdir, subtree)| arb_extant_file_recursive(&subdir, subtree)),
+            ]
+            .boxed(),
+        }
+    }
```
arxanas, line 196: suggestion: To make this easier, I recommend you store a flat list/map of files and manipulate that (which is the approach I used in <https://github.com/jj-vcs/jj/pull/6502>). Besides implementation difficulty, this approach will make it difficult to avoid biasing the selection, since it'll be strongly affected by directory structure.

arxanas (commented):
> It is possible to impose preconditions for transitions and I did so at first but I've found that this really impedes the exploration of the state space and leads into dead ends when shrinking, presumably because the initial state can no longer be shrunk independently from the transitions and vice versa

I've also noticed this, but the shrinking hasn't been too bad for me so far. (I suspect that's partially because I always start from an empty state, instead of a populated tree.)

I think a principled approach might involve allocating a fixed number of optional slots up-front, and then address elements by their index, ignoring operations that would address unoccupied slots. Then you could e.g. prevent a file from being generated without necessarily invalidating all the operations that happen afterwards. I haven't thought too much about it, though.

> I've not found a great way to test for other properties than what is done in the existing manual tests: selecting none of the changes produces the left tree, selecting them all produces the right tree. As mentioned in the discussion of the issue, it might make sense to apply only some changes, produce an intermediate tree, then apply the remaining changes to go all the way to the right tree. Usually, this sort of thing should be a perfect fit for proptesting but I've struggled to reconcile this with the state machine approach.

The state machine approach indeed seems a little limiting. I think you'd have to represent the desire to check only some of the checkboxes as a state transition, which is kind of clunky because it represents an entirely different system.

We could also drop the state machine approach and just switch to generating an operation (transition) list as the test case. Then we would have a helper function which applies the operation to a tree/repo. For the left vs right sides, the left side would start generating from the empty operation list, and the right side would start generating from the left operation list. Then we should also be able to use both sides as input to another strategy which selects some number of checkboxes in the resulting diff editor.

cli/src/merge_tools/builtin.rs
```diff
@@ -437,7 +437,13 @@ fn apply_changes(
 
         match contents {
             scm_record::SelectedContents::Unchanged => {
-                // Do nothing.
+                // scm_record currently only reports `Unchanged` if the file is empty.
```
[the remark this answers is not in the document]
jgreitemann: It is accurate insofar as that `Unchanged` is never reported for non-empty text files, irrespective of any unchecked `Changed` sections. However, binary files will also report `Unchanged` selected changes if the `Binary` section is not checked, so my fix here would potentially erase unchanged binary files.

I agree that this behavior is surprising and given that it was not intended, I've created https://github.com/arxanas/scm-record/issues/104 to track this upstream. However, this code right here currently depends on this quirky behavior to be able to apply file mode changes at least for unchanged text files.

cli/src/merge_tools/builtin.rs
```diff
@@ -437,7 +437,13 @@ fn apply_changes(
...
+                // TODO: It would be sufficient to only update the executable flag on the tree
+                // value, but there's no way to retrieve the FileId or only flip the executable
+                // flag. We resort to re-writing the file, knowing that it's empty.
+                let executable = file_mode == mode::EXECUTABLE;
+                let value = write_file(&path, &[], executable)?;
```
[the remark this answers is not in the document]
jgreitemann: > question: Shouldn't the Binary case below have the same issue? If the file has binary contents, and we change the executable bit, then shouldn't it also fail to update?

This is indeed the case and kind of exacerbates the issue since changing the executable flag of binary files is not at all uncommon. I guess that is further incentive to extend the proptests to non-UTF-8 file contents, like you mentioned.

lib/testutils/src/proptest.rs
jgreitemann: Agreed, the code in https://github.com/jj-vcs/jj/pull/6502 is much more concise.

lib/testutils/src/proptest.rs
jgreitemann: My motivation for using these over single letters was to have them appear less accidental. If someone who's not aware of how the path components are generated were to encounter `"a"`, they might assume that the string could not have been shrunk down any further and that the failure cannot be reproduced using uppercase letters, numbers, etc., which have a lower ASCII value. A recognizable word might raise more of an eyebrow and cause them to dig into the strategy.

At least, this was my thinking here... I'm not at all attached to the idea. Since path components are not properly shrunk in most cases anyway, I'm fine with sticking to single letters to reduce surprise.

lib/testutils/src/proptest.rs
jgreitemann: > Unfortunately, I don't think I communicated the design I was proposing well.

Your original comments were most helpful and I understood your arguments in favor of operation-based over recursive generation as it pertained to going from the left tree to the right tree.

I originally used the same operation-based generation to produce the `init_state` which I think would be roughly equivalent to the approach you have taken with a dedicated `Commit` transition. However, my transitions originally had pretty bad shrinking behavior due to the overly restrictive `preconditions()`. Recursive generation shrunk the initial state much better so my transitions only sucked while going from left to right tree. 😅 Now that the transition "deadlock" is less severe, I agree that it is much simpler to use the same generation scheme throughout.

The idea of using a dedicated transition to mark the "left" (or previous) tree occurred to me as well but I hesitated to implement it because it seemed to me as something that would be specific to the SUT and might limit the reusability of the reference state machine. By naming it `Commit`, you've convinced me that it is actually a natural operation to perform on a working copy state which is probably desirable in other situations as well.

> I've opened a PR against this one showing one way it could be done: https://github.com/jj-vcs/jj/pull/6502.

Thanks a lot for that and sorry for taking some time to get back to it. I'm still playing around with it and might leave some comments on it. In parallel, I'm investigating ways to fix the bugs in a way that doesn't require invasive action it the tree (builder) internals (as @yuja asked) by getting a little more info from scm-record. Eventually, I would squash your changes into mine and update this PR if that's alright with you?

lib/src/tree_builder.rs
jgreitemann: I think I found a better way to fix the issue that's not intruding into lower levels. As I mentioned, deleting a file vs. not creating a new file is not distinguishable from the _output_ of `scm_record::File::get_selected_contents()` alone, but when also taken its _input_ into account (i.e., whether the file mode change has been selected), it becomes decidable.

I've also reorganized the changes such that the individual bug fixes are committed along with the corresponding regression tests come first and only then added the property-based test. This makes it a little hard to verify that any of the tests are actually failing without the corresponding fix, or that the property-based test is capable of uncovering these edge cases, but it ensures that each patch is individually in a clean state with all tests passing.

cli/proptest-regressions/merge_tools/builtin.txt
```diff
@@ -0,0 +1,8 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc 50055483105a0234d9fb9d2eaf623033f583b31856f280a75f4d3877ea68046f # shrinks to (initial_state, transitions, seen_counter) = (RepoRefState { root: Directory { gamma: RegularFile { contents: "", executable: false } } }, [CreateFile { path: "gamma", contents: "", executable: true }], None)
+cc 7f8b4d5e36149675c133884be076c13317673a2fe0484986a2faf667924af673 # shrinks to (initial_state, transitions, seen_counter) = (RepoRefState { root: Directory { gamma: Directory { alpha: RegularFile { contents: "", executable: false } } } }, [CreateFile { path: "gamma", contents: "", executable: false }], None)
```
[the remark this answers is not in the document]
jgreitemann: I originally checked these in as per [guidance in the proptest book](https://proptest-rs.github.io/proptest/proptest/failure-persistence.html). But I've also found it quite tiring to keep them up-to-date as I extended the property-based test. I've now removed this failure persistence file.

All of the bugs that I've found through proptesting are now also covered by manual regression tests. The refactoring of `create_tree()` in terms of `DirEntry` was particularly useful to allow testing the various edge cases in that regard, something I struggled with before.

I've also experimented with calling the proptest SUT functions from within regular tests with fixed inputs. This works and could potentially get rid of a lot of boilerplate in the manual tests. I ended up not going this way because it also reduces the readability of individual tests and demands an understanding of the state machine model from the reader trying to debug a test failure.

cli/src/merge_tools/builtin.rs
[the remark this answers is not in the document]
jgreitemann: My new fix for this problem (see comment below) now handles file mode changes with `Unchanged` contents correctly. That is, with the present version of scm-record, the bug is fixed for empty and binary files. If arxanas/scm-record#104 is fixed at a later time, it would also correctly cover text files producing `Unchanged` contents. There would not be any observable effects to jj users, but the unchanged text file would no longer be rewritten to the store.

cli/src/merge_tools/builtin.rs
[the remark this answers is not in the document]
jgreitemann: I've created a new API `Merge::with_new_file_executable_bit()` which allows updating an existing tree value without having to rewrite the underlying file. This also fixes the issue of "lost" file mode changes on otherwise unchanged file without making any assumptions as to the contents of the existing file and thus workings on empty files and binary files alike.

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,322 @@
...
+
+#[derive(Debug, Clone)]
+pub enum Transition {
+    /// Create a file with the given contents at `path`.
+    ///
+    /// Parent directories are created as necessary, existing files or
+    /// directories are replaced.
+    CreateFile {
+        path: RepoPathBuf,
+        contents: String,
```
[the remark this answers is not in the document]
jgreitemann: By changing the regex to `(\n|[a-z]|.)*`, you were already able to produce files that would be treated as binary files due to the inclusion of control characters, specifically `\0`, as seen above. But it is still valid UTF-8 at this point, as enforced by the `String` type. I now took it up a notch and changed `contents`' type to `Vec<u8>` and explicitly including unconstrained binary data in the `arb_file_contents()` strategy.

Doing so admittedly did not uncover any additional bugs. However, having the option to  create `DirEntry` from `Vec<u8>` increased its flexibility in tests across other modules.

lib/testutils/src/proptest.rs
jgreitemann: I ended up not using proptest-derive after all. In #6502, you derived `Arbitrary` for `DirEntry` and `Transition`. You didn't actually use derived implementation on `Transition` anywhere because the actual generation strategy has to be informed by the current state. I could have kept the derived implementation for `DirEntry`, but I moved the type definition to the parent `testutils` module and since I wanted to keep the `proptest` stuff confined to the `testutils::proptest` module, I manually implemented `Arbitrary` on `DirEntry`.

lib/testutils/src/proptest.rs
jgreitemann: I adopted the flat map of files and even went a little further by not explicitly storing `DirEntry`s for directories. Instead, I only store files (and in future possibly symlinks or submodules); the ancestors of the files' paths are implicitly treated as directories.

This avoids including `Directory` as a variant in `DirEntry` which makes it more suited to be reused by `create_tree()`. As I understand it, empty directories are not tracked by jj, so that fits the working copy data model better.

jgreitemann:
I've updated the PR now with a revised implementation that incorporates many of the suggestions in #6502.

The `DirEntry` type has been moved up into `testutils` and is now used by `testutils::create_tree()` to allow the creation of `MergedTree`s which contain executable files, binary files, symlinks, and even submodules. This made some of the other helpers for writing these files individually obsolete and allows `create_tree()` to be used in more places. To keep the change minimally invasive to existing uses of `create_tree()` across all tests, `DirEntry` is convertible from `&str` to represent text files and `create_tree()` has a more generic signature now. This limited the change in most cases to the removal of an ampersand, i.e., passing an array rather than a slice.

---

> The state machine approach indeed seems a little limiting. I think you'd have to represent the desire to check only some of the checkboxes as a state transition, which is kind of clunky because it represents an entirely different system.

I managed to adapt the state machine testing in a way that covers partial selection of changes without having to reimplement the "runner" for the state machine test. Technically, selection of changes from the diff depends on the diff, but that would have been tricky to incorporate. I opted for an approach where the initial state includes a selection mask, which is then successively applied to the changes in the diff to decide whether to check them, each step of the way.

This approach is not perfect because diffs might change significantly in a way that breaks the "continuity" of the selections, which might impede the ability to shrink the transitions list. However, in practice, it worked surprisingly well and gave sufficiently easy-to-interpret failures. Sure enough, this test uncovered another bug that is fixed separately.

This sort of "roundtrip" property — first applying parts of a diff to produce an intermediate tree, then applying the complementary changes to it to go all the way to the "right" tree — does not apply to arbitrary diffs. There are certain combinations of selected vs. unselected changes where the selected changes necessary imply parts of the unselected changes, or where the "complementary" changes would end up undoing parts of the selected changes. Examples are given in the SUT's doc comment. The selection of changes is tweaked to avoid these conflicts. (Sidebar: One could argue that scm-record should only allow selections that are not conflicted in this way. It kind of does already, e.g., it will automatically select the file mode change when selecting any of the line additions in a newly created text file.)

As such, the test verifying this "roundtrip" property covers a search space that is not a superset of the original proptest verifying only that the left or right tree is reproduced when selecting none or all of the changes. For this reason, I've created a separate SUT for the "roundtrip" test, so now there are two independent state machine proptests using the same underlying state machine.

lib/src/repo_path.rs
```diff
@@ -503,6 +503,17 @@ impl PartialOrd for RepoPathBuf {
     }
 }
 
+impl<P: AsRef<RepoPathComponent>> Extend<P> for RepoPathBuf {
+    fn extend<T: IntoIterator<Item = P>>(&mut self, iter: T) {
+        for pc in iter {
```
martinvonz, line 527: nit: We generally avoid non-standard abbreviations and acronyms. `component` seems good to me

(sure, this is a very standard acronym but it doesn't mean "personal computer" here :))

lib/src/repo_path.rs
```diff
@@ -814,6 +825,17 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_extend() {
+        let mut path = RepoPath::root().to_owned();
+        path.extend([repo_path_component("dir")]);
+        assert_eq!(path.as_ref(), repo_path("dir"));
+        path.extend(std::iter::repeat_n(repo_path_component("subdir"), 3));
+        assert_eq!(path.as_ref(), repo_path("dir/subdir/subdir/subdir"));
+        path.extend(std::iter::empty::<RepoPathComponentBuf>());
```
martinvonz, line 854: nit: Also test this on when the path is empty. I could imagine a bug causing it to make the path change from "" to "/" (it would happen if you moved the condition to be before the loop).

lib/testutils/src/lib.rs
```diff
@@ -340,6 +339,88 @@ impl TestWorkspace {
...
+        }
+    }
+
+    pub fn is_executable(&self) -> bool {
+        match self {
+            DirEntry::File { executable, .. } => *executable,
+            DirEntry::Symlink { .. } => panic!("symlinks cannot be executable"),
+            DirEntry::GitSubmodule { .. } => panic!("submodules cannot be executable"),
+        }
+    }
```
martinvonz, line 399: nit: This doesn't seem used, so remove it and maybe it later if we need it? It would be nice if we don't even need this function with the panics. Maybe any callers can use `match` instead?

lib/testutils/src/lib.rs
```diff
@@ -340,6 +339,88 @@ impl TestWorkspace {
...
+    pub fn executable(self, executable: bool) -> Self {
+        match self {
+            Self::File { contents, .. } => Self::File {
+                contents,
+                executable,
+            },
+            DirEntry::Symlink { .. } => panic!("symlinks cannot be executable"),
+            DirEntry::GitSubmodule { .. } => panic!("submodules cannot be executable"),
+        }
+    }
```
martinvonz, line 391: To avoid these panics, and to simplify the API overall, I would make this a "static" function taking the content as input. We can rename `text_file()` to `normal_file()` then. Hopefully we only need one `executable(contents: &str) -> Self`, but adding a `binary_executable(contents: Vec<u8>)` if necessary also seems fine.

lib/testutils/src/lib.rs
```diff
@@ -340,6 +339,88 @@ impl TestWorkspace {
...
+
+    pub fn is_executable(&self) -> bool {
+        match self {
+            DirEntry::File { executable, .. } => *executable,
+            DirEntry::Symlink { .. } => panic!("symlinks cannot be executable"),
+            DirEntry::GitSubmodule { .. } => panic!("submodules cannot be executable"),
+        }
+    }
+
+    pub fn text_contents(&self) -> Option<&str> {
```
martinvonz, line 401: Same here: can the callers use `match` instead?

cli/src/merge_tools/builtin.rs
```diff
@@ -341,6 +341,16 @@ async fn make_diff_files(
                 sections.extend(make_diff_sections(&old_contents, &new_contents)?);
             }
 
+            (
+                FileContents::Binary {
+                    hash: left_hash, ..
+                },
+                FileContents::Binary {
+                    hash: right_hash, ..
+                },
```
yuja, line 350: nit: require `hash: Some(..)` for a bit of safety?

The hash field is used only for displaying purpose in other places.

lib/src/merge.rs
```diff
@@ -645,6 +645,24 @@ where
         Merge { values }
     }
 
+    /// Creates a new merge by setting the executable bit from the given merge.
+    /// In other words, the file ids from `self` will be preserved.
+    pub fn with_new_file_executable_bit(&self, executable: bool) -> Merge<Option<TreeValue>> {
```
yuja, line 650: nit: Since this is very specific to the current builtin editor, I would move this function to merge_tools/builtin.rs. If the builtin editor had a proper support for conflicts, the resolution might be handled differently.

BTW, `jj file chmod` can be a potential user of this function, but the implementation is relatively simple to just inline.

yuja (commented):
Since this PR is getting bigger, it might be better to split the property-based testing patches to another PR. We'll need different expertise to review these changes.

lib/src/repo_path.rs
jgreitemann: `RepoPathComponent[Buf]` enforce that a path component cannot be empty. I _could_ use `RepoPathComponent::new_unchecked("")` in the test but I see a point in testing behavior for unsound input.

That being said, shouldn't `new_unchecked()` or `from_internal_string_unchecked()` be `unsafe`?

lib/testutils/src/lib.rs
jgreitemann: `is_executable()` and `text_contents()` were indeed unused. I added them when originally sketching out the API for `DirEntry`. I agree it's better to only add them later when the need arises and that `match` is the better option when it comes to the `executable` bit.

lib/testutils/src/lib.rs
jgreitemann: Not really because `contents: Vec<u8>` does not match nicely against strings. But this function was not used and perhaps it won't ever be needed because `DirEntry` is so far only consumed by `create_tree()`. I've removed the function.

lib/src/repo_path.rs
```diff
@@ -427,6 +442,10 @@ impl RepoPath {
         RepoPathComponentsIter { value: &self.value }
     }
 
+    pub fn ancestors(&self) -> RepoPathAncestorsIter<'_> {
+        RepoPathAncestorsIter { path: Some(self) }
+    }
```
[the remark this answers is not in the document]
jgreitemann: Awesome! `successors()` is so damn powerful, yet I somehow always forget about it.

lib/tests/test_git.rs
```diff
@@ -2559,30 +2555,30 @@ fn test_reset_head_with_index_no_conflict() {
...
-            repo_path("some/dir/symlink"),
-            "./normal-file",
-        );
-        tree_builder.set(
-            repo_path_buf("some/dir/commit"),
-            TreeValue::GitSubmodule(testutils::write_random_commit(mut_repo).id().clone()),
-        );
-        MergedTreeId::resolved(tree_builder.write_tree().unwrap())
-    };
+    let tree_id = testutils::create_tree(
```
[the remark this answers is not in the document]
jgreitemann: Currently, I don't see a benefit in creating a builder on the tree level. `create_tree()` can now accept for example a `BTreeMap<&RepoPath, DirEntry>` and the API for `BTreeMap` (or `HashMap`) is well-known, so I would suggest to treat the `BTreeMap` as the builder. For the actual `DirEntry`, a ["builder lite"](https://matklad.github.io/2022/05/29/builder-lite.html) is already what I had in mind with `executable()`. I've changed it a bit now as per [Martin's comment](https://github.com/jj-vcs/jj/pull/6448#discussion_r2105394644).

lib/src/merge.rs
jgreitemann: I've moved the helper to `merge_tools/builtin.rs`. I found it quite natural to want such a function right next to the complementary `with_new_file_ids()`.

> If the builtin editor had a proper support for conflicts, the resolution might be handled differently.

Not quite sure how that might look like. I thought this was a pretty reasonable operation to have. Like you said, it's more or less `file chmod`.

jgreitemann:
Following yuja's advise, I've split off the changes which actually add the property-based tests into #6623.

This PR now only contains the actual bug fixes and their hand-crafted regression tests.

lib/testutils/src/lib.rs
jgreitemann: The new builder API I've added works with a dedicated type for file entries, so it's not possible to call `executable()` on symlink or submodule entries. This serves to avoid these panics while still keeping file contents and executable bit separate.

jgreitemann:
I've updated the PR, now including a builder API for creating trees in tests.

lib/tests/test_local_working_copy.rs
```diff
@@ -565,7 +565,13 @@ fn test_tree_builder_file_directory_transition() {
...
     // Add file at parent_path
     let mut tree_builder = store.tree_builder(store.empty_tree_id().clone());
-    testutils::write_normal_file(&mut tree_builder, parent_path, "");
+    tree_builder.set(
+        parent_path.to_owned(),
+        TreeValue::File {
+            id: testutils::write_file(store, parent_path, ""),
+            executable: false,
+        },
+    );
```
yuja, line 574: nit: maybe we shouldn't remove `write_normal_file()`, etc.? They seems to be useful.

lib/testutils/src/lib.rs
```diff
@@ -383,61 +383,136 @@ pub fn write_file(store: &Store, path: &RepoPath, contents: &str) -> FileId {
...
+            tree_builder,
+        }
+    }
+
+    pub fn entry(&mut self, path: &RepoPath) -> TestTreeEntryBuilder<'_> {
+        TestTreeEntryBuilder {
+            tree_builder: &mut self.tree_builder,
+            path: path.to_owned(),
+        }
+    }
```
yuja, line 405: nit: I think we can simply add `binary_file(path, content)`, `symlink(path, target)`, etc. here.

Sorry, I just said `.entry(path)` without considering how the API would look like.

lib/testutils/src/lib.rs
```diff
@@ -383,61 +383,136 @@ pub fn write_file(store: &Store, path: &RepoPath, contents: &str) -> FileId {
...
+                id,
+                executable: self.executable,
+            },
+        );
     }
-    let id = tree_builder.write_tree().unwrap();
-    store.get_tree(RepoPathBuf::root(), &id).unwrap()
 }
 
-pub fn create_tree(repo: &Arc<ReadonlyRepo>, path_contents: &[(&RepoPath, &str)]) -> MergedTree {
```
yuja, line 432: nit: just leave the current `create_tree()` for convenience, and add `create_tree_with()` or something for advanced use case?

cli/src/merge_tools/builtin.rs
```diff
@@ -648,6 +669,7 @@ pub fn edit_merge_builtin(
             .map(|file| file.repo_path.clone())
             .collect_vec(),
         &state.files,
+        |path| tree.path_value(path),
```
yuja, line 680: Can you check the behavior of "merge" (or `jj resolve`) editor?

If returning `tree.path_value(path)` is correct, the callback function shouldn't be named as `select_left()`. It's `select_original` or something. If this function should return the left tree, the same TODO comment as `select_right` would apply.

yuja (commented):
Thanks.

lib/tests/test_local_working_copy.rs
jgreitemann: These three uses in `test_local_working_copy.rs` were in fact the only ones. I would argue that the goal here is to test the `TreeBuilder` itself, so it is better to "inline" this helper so the calls to the tested `TreeBuilder` APIs become apparent.

Additionally, the helpers `write_normal_file()`, etc., operated on another different level compared to `write_file()`. The latter really only writes the object to the store, whereas `write_normal_file()` does that too _and_ sets the corresponding `TreeValue` on the `TreeBuilder`. That's not really apparent from the naming.

lib/testutils/src/lib.rs
jgreitemann: I changed it such that the extra `entry()` call is not necessary. However, the `*_file()` functions still just return a builder type for the entry such that the executable bit can be changed.

lib/testutils/src/lib.rs
jgreitemann: I've reinstated `create_tree()` and `create_single_tree()`. These now delegate to `create_tree_with()` and `create_single_tree_with()`, respectively. I've reverted all of the existing uses of these helpers to their original form unless there is some tangible benefit from using the builder API.

cli/src/merge_tools/builtin.rs
jgreitemann: I don't think this code path can be exercised at the moment. `select_left()` is only called when the file mode change is selected. `jj resolve` refuses to launch the merge editor when an executable file is conflicted.

If I'm not mistaken, `select_right()` also does not seem to be called in the merge editor, but for different reasons: conflicted binary files are currently not rendered as `Binary` sections. If the files happen to be valid UTF-8 (but containing the NUL control character), the merge editor will actually render them to the terminal:
```
[File] [Edit] [Select] [View]
[ ] bin.dat                                [-]
  ( ) Section 1/1                          (-)
    [ ] - ␀␁
    [ ] + ␀␁␂
```
I think it might the terminal that ultimately replaces the control characters by their Unicode replacement symbols, e.g., replacing NUL (U+0000) with ␀ (U+2400), but I'm not sure. The same change is rendered like this in the diff editor:
```
[File] [Edit] [Select] [View]
( ) bin.dat                                                                                                             (-)
  [ ] (binary contents: bdc955b7b2e610ad5a72302b139a2e6cb325519a (2B) -> 8352675d67aed6625ece79af41c27fdb4ee2e867 (3B))
```
If one of the files is not valid UTF-8, then the merge editor errors out:
```
Error: Failed to resolve conflicts
Caused by:
1: Failed to decode UTF-8 text for item conflicting hunk (this should not happen)
2: invalid utf-8 sequence of 1 bytes from index 0
```
This is most definitely not expected behavior but it kind of makes sense that it's not possible to describe the result of selecting all the changes shown by the `Recorder` because that does not correspond to any of the snapshotted files.

I think conceptually, you're right that `select_left` this is more of a `select_original`: In the diff editor, the original would be the left tree, but in a three-way merge, this would correspond to the middle element in the `MergedTree`, the base, not "ours". `select_right` should then perhaps be renamed `select_changed` or something? I've not renaming anything, yet, waiting on your thoughts on this.

Ultimately, the limitations of the single-column `Recorder` as a merge editor make all of this a little awkward. If it were capable of displaying the three sides of a three-way merge, like Meld does, then it wouldn't need to operate on a materialized form of the conflict at all...

lib/testutils/src/lib.rs
```diff
@@ -383,61 +383,142 @@ pub fn write_file(store: &Store, path: &RepoPath, contents: &str) -> FileId {
...
-    path: &RepoPath,
-    contents: &str,
-) -> FileId {
-    let id = write_file(tree_builder.store(), path, contents);
-    tree_builder.set(
-        path.to_owned(),
-        TreeValue::File {
-            id: id.clone(),
+pub struct TestTreeBuilder {
+    store: Arc<Store>,
```
yuja, line 387: nit: delete `store` or use it consistently. It seems weird that `TestTreeFileEntryBuilder` uses `tree_builder.store()`, but the main `TestTreeBuilder` doesn't.

lib/testutils/src/lib.rs
```diff
@@ -383,61 +383,142 @@ pub fn write_file(store: &Store, path: &RepoPath, contents: &str) -> FileId {
...
+            path: path.to_owned(),
+            contents,
             executable: false,
-        },
-    );
-    id
+        }
+    }
+
+    pub fn text_file(&mut self, path: &RepoPath, contents: &str) -> TestTreeFileEntryBuilder<'_> {
```
yuja: nit: `contents` can be `AsRef<[u8]>` or `Into<Vec<u8>>`. Then, we can merge `binary`/`text_file()` functions, which should be identical. `empty_file()` can also be removed, but I don't have strong feeling about that.

cli/src/merge_tools/builtin.rs
yuja: > If I'm not mistaken, `select_right()` also does not seem to be called in the merge editor, but for different reasons: conflicted binary files are currently not rendered as `Binary` sections.

Right. I think I added TODO comment about that. It seemed wrong to split hunks without checking if the content has null character. And I have no idea how left/right binary contents would be selected with the current UI.

Can you insert a comment to `|path| tree.path_value(path)` of `edit_merge_builtin()`? With the current state, it looks `select_right` is wrong whereas `select_left` is correct.

yuja (approved):
The interpretation of `scm_record` output is getting messy, but the implementation looks good to me.

cli/src/merge_tools/builtin.rs
jgreitemann: I've turned it into a `FIXME`, given that we're pretty sure that the `|path| tree.path_value(path)` implementation is wrong for `select_right`. I've considered replacing it with `unreachable!()`, but that might be a little too hot.

> And I have no idea how left/right binary contents would be selected with the current UI.

I think it might be possible to use the current scm_record UI to select between the "ours" or "theirs" sides to the conflict if both are represented by separate `Binary` sections. It would then be possible to obtain sensible descriptions for either one from the `Merge`, too. Obviously, it wouldn't make sense to select both sections, so scm_record would need to treat multiple `Binary` sections as mutable exclusive.

lib/testutils/src/lib.rs
jgreitemann: `Store::get_tree()` (called from `TestTreeBuilder::write_single_tree()`) receives `self: &Arc<Store>` but `TreeBuilder::store()` only provides access to `&Store`. That requires either keeping `Arc<Store>` as a member, or changing the public API of `TreeBuilder` to expose `&Arc<Store>` instead.

I've changed to now such that `TestTreeBuilder` uses `self.store` consistently, but I left `TestTreeFileEntryBuilder` alone because there no need to add a `store` member to it.

PR #6458, opened 2025-05-05
split: implement -A/-B/-d flags

glehmann:
See the comments in the commit descriptions.

The last two commits implement `jj --parallel -r x` as `jj split -r x -A x- -B x+` as discussed in #3419, but I'm not sure to see the point, except to document `--parallel` as equivalent to `-A x- -B x+` (or `-d x-` if x has no descendant). I'm including them here to continue the discussion.

Still some tests to implement, but some early feedback would be appreciated!

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -11,16 +11,26 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+use std::collections::HashMap;
 use std::io::Write as _;
 
 use clap_complete::ArgValueCompleter;
+use jj_lib::backend::CommitId;
 use jj_lib::commit::Commit;
 use jj_lib::matchers::Matcher;
 use jj_lib::object_id::ObjectId as _;
 use jj_lib::repo::Repo as _;
+use jj_lib::rewrite::move_commits;
 use jj_lib::rewrite::CommitWithSelection;
+use jj_lib::rewrite::EmptyBehaviour;
+use jj_lib::rewrite::MoveCommitsLocation;
+use jj_lib::rewrite::MoveCommitsTarget;
+use jj_lib::rewrite::RebaseOptions;
+use jj_lib::rewrite::RebasedCommit;
+use jj_lib::rewrite::RewriteRefsOptions;
 use tracing::instrument;
 
+use crate::cli_util::compute_commit_location;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::DiffSelector;
 use crate::cli_util::RevisionArg;
@@ -70,6 +80,40 @@ pub(crate) struct SplitArgs {
         add = ArgValueCompleter::new(complete::revset_expression_mutable),
     )]
     revision: RevisionArg,
+    /// The revision(s) to rebase onto (can be repeated to create a merge
+    /// commit)
+    #[arg(
+        long,
+        short,
+        conflicts_with = "parallel",
+        value_name = "REVSETS",
+        add = ArgValueCompleter::new(complete::revset_expression_all),
+    )]
+    destination: Option<Vec<RevisionArg>>,
+    /// The revision(s) to insert after (can be repeated to create a merge
+    /// commit)
+    #[arg(
+        long,
+        short = 'A',
+        visible_alias = "after",
+        conflicts_with = "destination",
+        conflicts_with = "parallel",
+        value_name = "REVSETS",
+        add = ArgValueCompleter::new(complete::revset_expression_all),
+    )]
+    insert_after: Option<Vec<RevisionArg>>,
+    /// The revision(s) to insert before (can be repeated to create a merge
+    /// commit)
+    #[arg(
+        long,
+        short = 'B',
+        visible_alias = "before",
+        conflicts_with = "destination",
+        conflicts_with = "parallel",
+        value_name = "REVSETS",
+        add = ArgValueCompleter::new(complete::revset_expression_mutable),
+    )]
+    insert_before: Option<Vec<RevisionArg>>,
     /// The change description to use (don't open editor)
     ///
     /// The description is used for the commit with the selected changes. The
@@ -116,11 +160,29 @@ impl SplitArgs {
             self.tool.as_deref(),
             self.interactive || self.paths.is_empty(),
         )?;
+        let use_move_flags = self.destination.is_some()
+            || self.insert_after.is_some()
+            || self.insert_before.is_some();
+        let (new_parent_ids, new_child_ids) = if use_move_flags {
+            compute_commit_location(
+                ui,
+                workspace_command,
+                self.destination.as_deref(),
+                self.insert_after.as_deref(),
+                self.insert_before.as_deref(),
+                "split-out commit",
+            )?
+        } else {
+            Default::default()
+        };
         Ok(ResolvedSplitArgs {
             target_commit,
             matcher,
             diff_selector,
             parallel: self.parallel,
+            use_move_flags,
+            new_parent_ids,
+            new_child_ids,
         })
     }
 }
@@ -130,6 +192,9 @@ struct ResolvedSplitArgs {
     matcher: Box<dyn Matcher>,
     diff_selector: DiffSelector,
     parallel: bool,
+    use_move_flags: bool,
+    new_parent_ids: Vec<CommitId>,
+    new_child_ids: Vec<CommitId>,
 }
 
 #[instrument(skip_all)]
@@ -144,6 +209,9 @@ pub(crate) fn cmd_split(
         matcher,
         diff_selector,
         parallel,
+        use_move_flags,
+        new_parent_ids,
+        new_child_ids,
     } = args.resolve(ui, &workspace_command)?;
     let text_editor = workspace_command.text_editor()?;
     let mut tx = workspace_command.start_transaction();
@@ -155,6 +223,12 @@ pub(crate) fn cmd_split(
     let first_commit = {
         let mut commit_builder = tx.repo_mut().rewrite_commit(&target.commit).detach();
         commit_builder.set_tree_id(target.selected_tree.id());
+        if use_move_flags {
+            commit_builder
+                // Generate a new change id so that the commit being split doesn't
+                // become divergent.
+                .generate_new_change_id();
+        }
         let description = if !args.message_paragraphs.is_empty() {
             let description = join_message_paragraphs(&args.message_paragraphs);
             if !description.is_empty() {
@@ -183,7 +257,7 @@ pub(crate) fn cmd_split(
     // select.
     let second_commit = {
         let target_tree = target.commit.tree()?;
-        let new_tree = if args.parallel {
+        let new_tree = if parallel {
             // Merge the original commit tree with its parent using the tree
             // containing the user selected changes as the base for the merge.
             // This results in a tree with the changes the user didn't select.
@@ -199,10 +273,13 @@ pub(crate) fn cmd_split(
         let mut commit_builder = tx.repo_mut().rewrite_commit(&target.commit).detach();
         commit_builder
             .set_parents(parents)
-            .set_tree_id(new_tree.id())
-            // Generate a new change id so that the commit being split doesn't
-            // become divergent.
-            .generate_new_change_id();
+            .set_tree_id(new_tree.id());
+        if !use_move_flags {
+            commit_builder
+                // Generate a new change id so that the commit being split doesn't
+                // become divergent.
+                .generate_new_change_id();
+        }
         let description = if target.commit.description().is_empty() {
             // If there was no description before, don't ask for one for the
             // second commit.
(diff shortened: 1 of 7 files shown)


cli/tests/test_split_command.rs
```diff
@@ -1125,49 +1123,30 @@ fn test_split_with_message() {
...
-#[test_case(BookmarkBehavior::Default; "default_behavior")]
-#[test_case(BookmarkBehavior::MoveBookmarkToChild; "move_bookmark_to_child")]
-#[test_case(BookmarkBehavior::LeaveBookmarkWithTarget; "leave_bookmark_with_target")]
-fn test_split_with_bookmarks(bookmark_behavior: BookmarkBehavior) {
+#[test]
+fn test_split_with_bookmarks() {
     let mut test_env = TestEnvironment::default();
     let edit_script = test_env.set_up_fake_editor();
     test_env.run_jj_in(".", ["git", "init", "main"]).success();
     let main_dir = test_env.work_dir("main");
```
PhilipMetzger, line 1151: Can you restore this test? I don't think this addresses the discussion of choosing the correct default for the issue.

cli/src/commands/split.rs
```diff
@@ -70,6 +71,12 @@ pub(crate) struct SplitArgs {
         add = ArgValueCandidates::new(complete::mutable_revisions)
     )]
     revision: RevisionArg,
+    /// The change description to use (don't open editor)
+    ///
+    /// The description is used for the split commit. The source
```
martinvonz, line 120: "the split commit" is not clear to me. It's pretty clear once you're read the next sentence, but I like the phrasing you used in the commit description and in the changelog (i.e. "the commit with the selected changes")

cli/tests/test_split_command.rs
PhilipMetzger: > Do you mean restoring the two possible bookmark behaviors?

Yes, since it deletes the test required for a currently hidden but supported flag and leaves it untested. 

> The split-out commit may be moved anywhere; I wouldn't expect the bookmark to be attached to that commit. If I have a commit tree like:

yes, that's also the thing I want but I would rather have a separate test for now until the maintainers come to a conclusion for #3419 (I at least posted my final opinion on it after seeing this PR).

cli/src/commands/split.rs
martinvonz, line 1: > split: the second commit keeps the change id

I thought this was still controversial. Wasn't the idea to only add the `-A/-B/-d` flags for now and let people test them for a while? @emilazy

glehmann: I've moved too fast on the implementation that made the most sense to me (but not to everyone, obviously).
I'll move that change to another PR, with a configuration to switch the behavior, and restore the configuration for the bookmark behavior as pointed by @PhilipMetzger

cli/src/commands/split.rs
```diff
@@ -228,16 +298,15 @@ pub(crate) fn cmd_split(
 
     let legacy_bookmark_behavior = tx.settings().get_bool("split.legacy-bookmark-behavior")?;
     if legacy_bookmark_behavior {
-        // Mark the commit being split as rewritten to the second commit. This
-        // moves any bookmarks pointing to the target commit to the second
-        // commit.
```
martinvonz, line 213: Why did you remove this comment? Is it no longer valid?

cli/src/commands/split.rs
```diff
@@ -160,8 +160,23 @@ impl SplitArgs {
             self.tool.as_deref(),
             self.interactive || self.paths.is_empty(),
         )?;
-        let (new_parent_ids, new_child_ids) = if self.parallel {
-            Default::default()
+        let (destination, insert_after, insert_before) = if self.parallel {
+            let mut destination = None;
+            let mut insert_after = Some(vec![RevisionArg::from(format!("({})-", self.revision))]);
```
martinvonz, line 165: We shouldn't have to parse a revset expression here. This should work:

```rust
        let (new_parent_ids, new_child_ids) = if self.parallel {
            let new_child_ids: Vec<_> = RevsetExpression::commit(target_commit.id().clone())
                .children()
                .evaluate(workspace_command.repo().as_ref())?
                .iter()
                .try_collect()?;
            (target_commit.parent_ids().to_vec(), new_child_ids)
        } else {
            let insert_before = if self.destination.is_none()
                && self.insert_after.is_none()
                && self.insert_before.is_none()
            {
                Some(vec![self.revision.clone()])
            } else {
                self.insert_before.clone()
            };
            compute_commit_location(
                ui,
                workspace_command,
                self.destination.as_deref(),
                self.insert_after.as_deref(),
                insert_before.as_deref(),
                "split-out commit",
            )?
        };
```

cli/src/commands/split.rs
glehmann: At first I completely removed the legacy bookmark config and used this call to change the parent in the rewrite, so yes it was no longer valid. (I'm not saying that necessarily the best way to do it :))

Since I reintroduced the legacy bookmark config and had to set the parent explicitly in the rewrite, I think that I can completely remove the call associated with that commit.

glehmann: I confused the situation with legacy change id / legacy bookmark.
We still need the call, and the comment that goes with it

cenviity:
_(non-blocking)_

Since the 'first commit / part' (with the selected changes to be split out) can now go anywhere in the commit tree, including **after** the 'second commit / part'. While historically the names made sense because of the parent-child relationship, it may make sense to refer to these two commits with more meaningful names in the CLI output, docs and elsewhere.

Quoting https://github.com/jj-vcs/jj/pull/6458#discussion_r2074211895 by @martinvonz:

> "the split commit" is not clear to me. It's pretty clear once you're read the next sentence, but I like the phrasing you used in the commit description and in the changelog (i.e. "the commit with the selected changes")

My suggestion is to replace 'first commit / part' and 'second commit / part' with '**commit with selected changes**' and '**commit with remaining changes**' consistently (or 'revision with…'?). In the CLI, it would be:

```
Selected changes:  qpvuntsm c002c01e fix in file1
Remaining changes: kkmpptxz 0841f981 my feature
```

An alternative is to rename 'second commit' to 'source commit', as in the 'source' of the split operation. But then renaming 'first commit' to 'destination commit' would be confusing, since this collides with the commit referenced by the `-d`/`--destination` flag if used (which would actually become the parent of the 'destination commit').

cli/tests/test_split_command.rs
```diff
@@ -1086,6 +1088,296 @@ fn test_split_with_non_empty_description_and_trailers() {
...
+    let output = work_dir.run_jj([
+        "split",
+        "-m",
+        "file1",
+        "-r",
+        "qpvuntsmwlqt",
+        "--insert-before",
+        "qpvuntsmwlqt",
+        "file1",
+    ]);
```
cenviity, line 1185: This is more relevant to #6466 but is more easily demonstrated in the changes here.

Trying to reconcile the command that was run here with the log output below makes it quite obvious to me that the revision with the remaining changes (i.e. the 'second commit') should keep the change ID, as in #6466.

The command indicates we are inserting a revision **before** `qpvuntsmwlqt`. But because the revision being inserted (i.e. the 'first commit') keeps that change ID, comparing the log outputs before and after appears to show a new revision `vruxwmqvtpmx` has been added **after** `qpvuntsmwlqt`, when in fact `vruxwmqvtpmx` is the original revision and `qpvuntsmwlqt` is the newly inserted revision.

The same goes for all of the other examples in this test.

cli/src/commands/split.rs
glehmann: I've removed those commits from this PR. If someone really wants them, please ask :-)

cli/src/commands/split.rs
```diff
@@ -152,7 +222,7 @@ pub(crate) fn cmd_split(
     let target = select_diff(ui, &tx, &target_commit, &matcher, &diff_selector)?;
 
     // Create the first commit, which includes the changes selected by the user.
-    let first_commit = {
+    let mut first_commit = {
```
yuja, line 225: Perhaps, new change id should be assigned to the `first_commit` if `-A`/`-B`/`-d` are specified. These options only makes sense if the split-out commit is considered created. It will also help review the test output.

I also think it's better to split the functions. Legacy bookmark/working-copy handling shouldn't exist in the new "split" behavior.

```rs
    let first_commit = {
        if new_behavior {
            .generate_new_change-id()
    };
    let second_commit = ...;
    if new_behavior {
        new() // no need to adjust rewritten commits, move working-copy commit, etc.
    } else {
        legacy()  // handle --parallel, legacy-bookmark-behavior, etc.
    }
```

cli/src/commands/split.rs
```diff
@@ -259,8 +378,20 @@ pub(crate) fn cmd_split(
...
     if let Some(mut formatter) = ui.status_formatter() {
-        if num_rebased > 0 {
-            writeln!(formatter, "Rebased {num_rebased} descendant commits")?;
+        let mut num_rebased_descendants = rewritten_commits.len() + rebased_commits.len();
+        // don't count the commit generated by the split in the rebased commits
+        num_rebased_descendants -= num_new_rebased;
+        for (_, ref rewritten) in rewritten_commits {
+            if rebased_commits.contains_key(rewritten) {
+                // only count once a commit that may have been rewritten twice in the process
+                num_rebased_descendants -= 1;
```
yuja, line 387: nit: rebasing twice would leave intermediate state in evolution history. Maybe this can be addressed later.

emilazy:
Just want to bring up a consideration that @ilyagr raised about these flags a while back that I think hasn’t been resurfaced since then: it’s a little unclear how the UI should work with these flags when using external diff editors. With the built‐in editor, it’s very obvious: you’re selecting hunks that get placed in a new commit on top of the selected destination. But with other diff editors, it’s a little weirder, since you’re constructing a state that immediately gets rebased – and indeed that rebase can result in conflicts (even with the built‐in editor). Arguably you would want the base to be the destination commit and you move hunks selectively from the source commit – that way you’re constructing the actual resulting tree directly and can avoid creating conflicts – but I think it’s not clear how we would achieve that without making the UI with the built‐in editor weird.

This is not intended as a blocking comment of any kind but just as a note that we should try this out with external diff editors and see what seems to make sense. Possibly we could have a per‐editor flag for how it should act, even.

cli/src/commands/split.rs
glehmann: > Perhaps, new change id should be assigned to the `first_commit` if `-A`/`-B`/`-d` are specified. These options only makes sense if the split-out commit is considered created. It will also help review the test output.

I'm all in favor of assigning the new change id to the first commit, but wouldn't that already be implementing part of #6466, which was moved there because it is controversial?
That would also introduce an inconsistency in the behavior of split depending on the usage of `-A/-B/-d` that I would prefer to avoid.

> I also think it's better to split the functions. Legacy bookmark/working-copy handling shouldn't exist in the new "split" behavior.
> 
> ```rust
>     let first_commit = {
>         if new_behavior {
>             .generate_new_change-id()
>     };
>     let second_commit = ...;
>     if new_behavior {
>         new() // no need to adjust rewritten commits, move working-copy commit, etc.
>     } else {
>         legacy()  // handle --parallel, legacy-bookmark-behavior, etc.
>     }
> ```

There is also the parallel handling to do in the second commit block, and the original commit descendants still need to be rewritten in the new behavior, so I'm not sure it would clarify much the code. I'll try, and we'll see :-)

Removing the legacy bookmark config would help much more IMO, but that's not for this PR!

cli/src/commands/split.rs
glehmann: I tried to call `move_commit` and then rewrite the descendant commit that may not have been touched by `move_commit`, but I wasn't always getting a consistent result.
That may be easier to do now that we have a set of tests to make sure it works properly.
If you don't mind, I'll try that again in another PR

cli/src/commands/split.rs
glehmann: > I'll try, and we'll see

@yuja I've pushed the split implementation. I don't know if I like it more, but I don't like it less :-)
Please let me know if you prefer that version.

yuja: > I'm all in favor of assigning the new change id to the first commit, but wouldn't that already be implementing part of #6466, which was moved there because it is controversial? That would also introduce an inconsistency in the behavior of split depending on the usage of `-A/-B/-d` that I would prefer to avoid.

My understanding is that we aren't settled on the default. If one of `-A/-B/-d` is specified, the user is opt in to the new behavior, so I think it's okay to implement whatever makes sense with these flags. (iirc, the current/legacy default can be emulated by using `-A/-B/-d`, but I don't remember the detail.)

cli/src/commands/split.rs
yuja: Yeah, in another PR. Feel free to add some TODO comment if that makes sense.

I think this can be addressed by removing predecessors records from the commit object, but we don't agree on that yet.

cli/src/commands/split.rs
glehmann: I've updated the code to assign the change-id to the second commit when any of `-A/-B/-d` is used.

glehmann:
I've created #6509 to rename first/second commit in selected/remaining changes

cli/src/commands/split.rs
```diff
@@ -70,6 +80,40 @@ pub(crate) struct SplitArgs {
...
     )]
     revision: RevisionArg,
+    /// The revision(s) to rebase onto (can be repeated to create a merge
+    /// commit)
+    #[arg(
+        long,
+        short,
+        conflicts_with = "parallel",
+        value_name = "REVSETS",
+        add = ArgValueCompleter::new(complete::revset_expression_mutable),
```
yuja, line 90: nit: `revset_expression_all`

cli/src/commands/split.rs
```diff
@@ -70,6 +80,40 @@ pub(crate) struct SplitArgs {
...
+    /// The revision(s) to insert after (can be repeated to create a merge
+    /// commit)
+    #[arg(
+        long,
+        short = 'A',
+        visible_alias = "after",
+        conflicts_with = "destination",
+        conflicts_with = "parallel",
+        value_name = "REVSETS",
+        add = ArgValueCompleter::new(complete::revset_expression_mutable),
```
yuja, line 102: nit: `revset_expression_all`

cli/src/commands/split.rs
```diff
@@ -226,6 +303,111 @@ pub(crate) fn cmd_split(
...
+            second_commit,
+            new_parent_ids,
+            new_child_ids,
+        )
+    } else {
+        legacy(ui, tx, target, first_commit, second_commit, parallel)
+    }
+}
+
+fn new(
```
yuja, line 321: nit: rename to `rebase_new()` or something saner?

cli/src/commands/split.rs
```diff
@@ -226,6 +303,111 @@ pub(crate) fn cmd_split(
...
+    mut tx: WorkspaceCommandTransaction,
+    target: CommitWithSelection,
+    mut first_commit: Commit,
+    mut second_commit: Commit,
+    new_parent_ids: Vec<CommitId>,
+    new_child_ids: Vec<CommitId>,
+) -> Result<(), CommandError> {
+    let mut rewritten_commits: HashMap<CommitId, CommitId> = HashMap::new();
+    tx.repo_mut()
+        .set_rewritten_commit(target.commit.id().clone(), second_commit.id().clone());
```
yuja, line 332: nit: this shouldn't be needed since the second commit inherits the original change id.

cli/src/commands/split.rs
```diff
@@ -226,6 +303,111 @@ pub(crate) fn cmd_split(
...
+        &RebaseOptions {
+            empty: EmptyBehaviour::Keep,
+            rewrite_refs: RewriteRefsOptions {
+                delete_abandoned_bookmarks: false,
+            },
+            simplify_ancestor_merge: false,
+        },
+    )?;
+
+    let mut num_new_rebased = 1;
```
yuja, line 368: nit: can you add comment why it is `1`?

I think it would decrement the fake entry for `rewritten_commits[target]`, but I'm not sure.

cli/src/commands/split.rs
```diff
@@ -226,6 +303,111 @@ pub(crate) fn cmd_split(
...
+    }
+    if let Some(RebasedCommit::Rewritten(commit)) = stats.rebased_commits.get(second_commit.id()) {
+        second_commit = commit.clone();
+    }
+
+    if let Some(mut formatter) = ui.status_formatter() {
+        let mut num_rebased_descendants = rewritten_commits.len() + stats.rebased_commits.len();
+        // don't count the commit generated by the split in the rebased commits
+        num_rebased_descendants -= num_new_rebased;
+        for (_, ref rewritten) in rewritten_commits {
```
yuja, line 381: nit: `rewritten_commits.values()`

It's also good to exclude duplicated entries upfront by `.iter().filter(..).count()`.

cli/tests/test_split_command.rs
```diff
@@ -1142,6 +1142,240 @@ fn test_split_with_message() {
...
+    ○  mzvwutvlkqwt false file5
+    │ ○  kkmpptxzrspx false file4
+    │ ○  rlvkpnrzqnoo false file3
+    │ ○  qpvuntsmwlqt false file2
+    │ ○  vruxwmqvtpmx false file1
+    ├─╯
+    ◆  zzzzzzzzzzzz true
+    [EOF]
+    ");
+
```
yuja, line 1203: I think it's better to assert content of the split commits. Maybe we can include real file names in the log output?

cli/src/commands/split.rs
```diff
@@ -116,11 +160,29 @@ impl SplitArgs {
...
+            || self.insert_before.is_some();
+        let (new_parent_ids, new_child_ids) = if use_move_flags {
+            compute_commit_location(
+                ui,
+                workspace_command,
+                self.destination.as_deref(),
+                self.insert_after.as_deref(),
+                self.insert_before.as_deref(),
+                "split-out commit",
+            )?
```
yuja, line 174: nit: maybe better to add `split -B` to test_immutable_commits.rs

yuja (approved):
Looks good to me overall.

cli/src/commands/split.rs
glehmann: Yes that's it

cli/src/commands/split.rs
glehmann: done in #6661!

PR #6465, opened 2025-05-06
working_copy: minimize use of #[cfg(windows)] around executable bit handling

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -116,10 +116,34 @@ use crate::working_copy::WorkingCopy;
 use crate::working_copy::WorkingCopyFactory;
 use crate::working_copy::WorkingCopyStateError;
 
+/// On-disk state of file executable bit.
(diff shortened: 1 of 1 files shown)


lib/src/local_working_copy.rs
```diff
@@ -116,10 +116,34 @@ use crate::working_copy::WorkingCopy;
 use crate::working_copy::WorkingCopyFactory;
 use crate::working_copy::WorkingCopyStateError;
 
+/// On-disk state of file executable bit.
+// TODO: maybe better to preserve the executable bit on all platforms, and
+// ignore conditionally? #3949
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct FileExecutableFlag(#[cfg(unix)] bool);
```
thoughtpolice, line 123: I didn't realize you could `#[cfg]` inside tuple structs, but it makes sense. I'm going to keep that trick in my back pocket.

thoughtpolice (approved):

PR #6466, opened 2025-05-06
split: the revision with the remaining changes keeps the change id

glehmann:
based on: #6458

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj gerrit upload` now supports the `-o` (`--option`) flag, which works like
   `git push -o` (`--push-option`).
 
+* `jj split` now consistently assigns the source revision's change ID to the
+  revision containing the non-selected changes—and ensures any associated
+  bookmark references that same change ID—regardless of the options used.
+  You can opt out of this change by setting `split.legacy-bookmark-behavior = true`,
+  but this will likely be removed in a future release.
+
 ### Fixed bugs
 
 * `jj` now creates a new working-copy revision during snapshotting if the
--- a/cli/src/commands/commit.rs
+++ b/cli/src/commands/commit.rs
@@ -45,9 +45,10 @@ use crate::ui::Ui;
 /// * `jj commit` doesn't have a `-r` option. It always acts on the working-copy
 ///   commit (@).
 ///
-/// * `jj split` (without `-o`/`-A`/`-B`) will move bookmarks forward from the
-///   old change to the child change. `jj commit` doesn't move bookmarks
-///   forward.
+/// * `jj commit` creates the new commit as a child of the working-copy, and
+///   sets the new commit as the working-copy. `jj split` creates the new commit
+///   as a parent of the source commit and leaves the working copy on the child
+///   commit.
 ///
 /// * `jj split` allows you to move the selected changes to a different
 ///   destination with `-o`/`-A`/`-B`.
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -292,11 +292,14 @@ pub(crate) async fn cmd_split(
     // Prompt the user to select the changes they want for the first commit.
     let target = select_diff(ui, &tx, &target_commit, &matcher, &diff_selector).await?;
 
+    let source_id_to_first_commit =
+        !use_move_flags && tx.settings().get_bool("split.legacy-bookmark-behavior")?;
+
     // Create the first commit, which includes the changes selected by the user.
     let first_commit = {
         let mut commit_builder = tx.repo_mut().rewrite_commit(&target.commit).detach();
         commit_builder.set_tree(target.selected_tree.clone());
-        if use_move_flags {
+        if !source_id_to_first_commit {
             commit_builder.clear_rewrite_source();
             // Generate a new change id so that the commit being split doesn't
             // become divergent.
@@ -365,7 +368,7 @@ pub(crate) async fn cmd_split(
         let mut commit_builder = tx.repo_mut().rewrite_commit(&target.commit).detach();
         commit_builder.set_parents(parents).set_tree(new_tree);
         let mut show_editor = args.editor;
-        if !use_move_flags {
+        if source_id_to_first_commit {
             commit_builder.clear_rewrite_source();
             // Generate a new change id so that the commit being split doesn't
             // become divergent.
--- a/cli/tests/test_split_command.rs
+++ b/cli/tests/test_split_command.rs
@@ -83,10 +83,10 @@ fn test_split_by_paths() -> TestResult {
     ]);
     insta::assert_snapshot!(output, @"
     ------- stderr -------
-    Selected changes : qpvuntsm 6dbc7747 (no description set)
-    Remaining changes: zsuskuln 42cbbc02 (no description set)
-    Working copy  (@) now at: zsuskuln 42cbbc02 (no description set)
-    Parent commit (@-)      : qpvuntsm 6dbc7747 (no description set)
+    Selected changes : zsuskuln 8a73f71d (no description set)
+    Remaining changes: qpvuntsm c4d8ebac (no description set)
+    Working copy  (@) now at: qpvuntsm c4d8ebac (no description set)
+    Parent commit (@-)      : zsuskuln 8a73f71d (no description set)
     [EOF]
     ");
     // Trailers should be added to the editor template
@@ -97,7 +97,7 @@ fn test_split_by_paths() -> TestResult {
 
     Trailer: value
 
-    JJ: Change ID: qpvuntsm
+    JJ: Change ID: zsuskuln
     JJ: This commit contains the following changes:
     JJ:     A file2
     JJ:
@@ -106,8 +106,8 @@ fn test_split_by_paths() -> TestResult {
     assert!(!test_env.env_root().join("editor1").exists());
 
     insta::assert_snapshot!(get_log_output(&work_dir), @"
-    @  zsuskulnrvyr false
-    ○  qpvuntsmwlqt false
+    @  qpvuntsmwlqt false
+    ○  zsuskulnrvyr false
     ◆  zzzzzzzzzzzz true
     [EOF]
     ");
@@ -142,17 +142,17 @@ fn test_split_by_paths() -> TestResult {
     ------- stderr -------
     Warning: All changes have been selected, so the original revision will become empty
     Rebased 1 descendant commits
-    Selected changes : qpvuntsm 9fd1c9e1 (no description set)
-    Remaining changes: znkkpsqq 41e0da21 (empty) (no description set)
-    Working copy  (@) now at: zsuskuln a06e40b8 (no description set)
-    Parent commit (@-)      : znkkpsqq 41e0da21 (empty) (no description set)
+    Selected changes : znkkpsqq d6e65134 (no description set)
+    Remaining changes: zsuskuln aa27eaa3 (empty) (no description set)
+    Working copy  (@) now at: qpvuntsm e94cab21 (no description set)
+    Parent commit (@-)      : zsuskuln aa27eaa3 (empty) (no description set)
     [EOF]
     ");
 
     insta::assert_snapshot!(get_log_output(&work_dir), @"
-    @  zsuskulnrvyr false
-    ○  znkkpsqqskkl true
-    ○  qpvuntsmwlqt false
+    @  qpvuntsmwlqt false
+    ○  zsuskulnrvyr true
+    ○  znkkpsqqskkl false
     ◆  zzzzzzzzzzzz true
     [EOF]
     ");
@@ -174,17 +174,17 @@ fn test_split_by_paths() -> TestResult {
     Warning: No matching entries for paths: nonexistent
     Warning: No changes have been selected, so the new revision will be empty
     Rebased 1 descendant commits
-    Selected changes : qpvuntsm 49416632 (empty) (no description set)
-    Remaining changes: lylxulpl 718afbf5 (no description set)
-    Working copy  (@) now at: zsuskuln 0ed53ee6 (no description set)
-    Parent commit (@-)      : lylxulpl 718afbf5 (no description set)
+    Selected changes : lylxulpl 3d639d71 (empty) (no description set)
+    Remaining changes: znkkpsqq 706a0e77 (no description set)
+    Working copy  (@) now at: qpvuntsm 502cf440 (no description set)
+    Parent commit (@-)      : znkkpsqq 706a0e77 (no description set)
     [EOF]
     ");
 
     insta::assert_snapshot!(get_log_output(&work_dir), @"
-    @  zsuskulnrvyr false
-    ○  lylxulplsnyw false
-    ○  qpvuntsmwlqt true
+    @  qpvuntsmwlqt false
+    ○  znkkpsqqskkl false
+    ○  lylxulplsnyw true
     ◆  zzzzzzzzzzzz true
     [EOF]
     ");
@@ -202,10 +202,10 @@ fn test_split_by_paths() -> TestResult {
     insta::assert_snapshot!(output, @"
     ------- stderr -------
     Warning: All changes have been selected, so the original revision will become empty
-    Selected changes : uyznsvlq 971ccc0b (no description set)
-    Remaining changes: xznxytkn a267cd96 (empty) (no description set)
-    Working copy  (@) now at: smwtzssm 6715dc2c (empty) (no description set)
-    Parent commit (@-)      : uyznsvlq 971ccc0b (no description set)
+    Selected changes : xznxytkn e69850ba (no description set)
+    Remaining changes: uyznsvlq df0b606a (empty) (no description set)
+    Working copy  (@) now at: smwtzssm 6c647048 (empty) (no description set)
+    Parent commit (@-)      : xznxytkn e69850ba (no description set)
     [EOF]
     ");
     Ok(())
@@ -236,10 +236,10 @@ fn test_split_with_non_empty_description() -> TestResult {
     let output = work_dir.run_jj(["split", "file1"]);
     insta::assert_snapshot!(output, @"
     ------- stderr -------
-    Selected changes : qpvuntsm c7f7b14b part 1
-    Remaining changes: kkmpptxz ac33a5a9 part 2
-    Working copy  (@) now at: kkmpptxz ac33a5a9 part 2
-    Parent commit (@-)      : qpvuntsm c7f7b14b part 1
+    Selected changes : kkmpptxz 530f78ed part 1
+    Remaining changes: qpvuntsm 88189e08 part 2
+    Working copy  (@) now at: qpvuntsm 88189e08 part 2
+    Parent commit (@-)      : kkmpptxz 530f78ed part 1
     [EOF]
     ");
 
@@ -248,7 +248,7 @@ fn test_split_with_non_empty_description() -> TestResult {
     JJ: Enter a description for the selected changes.
     test
 
-    JJ: Change ID: qpvuntsm
+    JJ: Change ID: kkmpptxz
     JJ: This commit contains the following changes:
     JJ:     A file1
     JJ:
@@ -259,15 +259,15 @@ fn test_split_with_non_empty_description() -> TestResult {
     JJ: Enter a description for the remaining changes.
     test
 
-    JJ: Change ID: kkmpptxz
+    JJ: Change ID: qpvuntsm
     JJ: This commit contains the following changes:
     JJ:     A file2
     JJ:
     JJ: Lines starting with "JJ:" (like this one) will be removed.
     "#);
     insta::assert_snapshot!(get_log_output(&work_dir), @"
-    @  kkmpptxzrspx false part 2
-    ○  qpvuntsmwlqt false part 1
+    @  qpvuntsmwlqt false part 2
+    ○  kkmpptxzrspx false part 1
     ◆  zzzzzzzzzzzz true
     [EOF]
     ");
@@ -292,10 +292,10 @@ fn test_split_with_default_description() -> TestResult {
     let output = work_dir.run_jj(["split", "file1"]);
     insta::assert_snapshot!(output, @"
     ------- stderr -------
-    Selected changes : qpvuntsm ff633dcc TESTED=TODO
-    Remaining changes: rlvkpnrz b1d20b7e (no description set)
-    Working copy  (@) now at: rlvkpnrz b1d20b7e (no description set)
-    Parent commit (@-)      : qpvuntsm ff633dcc TESTED=TODO
+    Selected changes : rlvkpnrz 16dc7e13 TESTED=TODO
+    Remaining changes: qpvuntsm f40d53f2 (no description set)
+    Working copy  (@) now at: qpvuntsm f40d53f2 (no description set)
+    Parent commit (@-)      : rlvkpnrz 16dc7e13 TESTED=TODO
     [EOF]
     ");
 
@@ -310,7 +310,7 @@ fn test_split_with_default_description() -> TestResult {
 
     TESTED=TODO
 
-    JJ: Change ID: qpvuntsm
+    JJ: Change ID: rlvkpnrz
     JJ: This commit contains the following changes:
     JJ:     A file1
     JJ:
@@ -318,8 +318,8 @@ fn test_split_with_default_description() -> TestResult {
     "#);
     assert!(!test_env.env_root().join("editor2").exists());
     insta::assert_snapshot!(get_log_output(&work_dir), @"
-    @  rlvkpnrzqnoo false
-    ○  qpvuntsmwlqt false TESTED=TODO
+    @  qpvuntsmwlqt false
+    ○  rlvkpnrzqnoo false TESTED=TODO
     ◆  zzzzzzzzzzzz true
     [EOF]
     ");
@@ -370,17 +370,17 @@ fn test_split_with_descendants() -> TestResult {
     insta::assert_snapshot!(output, @"
     ------- stderr -------
     Rebased 2 descendant commits
-    Selected changes : qpvuntsm 74306e35 Add file1
-    Remaining changes: royxmykx 0a37745e Add file2
-    Working copy  (@) now at: kkmpptxz 7ee84812 Add file4
-    Parent commit (@-)      : rlvkpnrz d335bd94 Add file3
+    Selected changes : royxmykx e13e94b9 Add file1
+    Remaining changes: qpvuntsm cf8ebbab Add file2
+    Working copy  (@) now at: kkmpptxz 73a16519 Add file4
+    Parent commit (@-)      : rlvkpnrz ec4d3a14 Add file3
     [EOF]
     ");
     insta::assert_snapshot!(get_log_output(&work_dir), @"
     @  kkmpptxzrspx false Add file4
     ○  rlvkpnrzqnoo false Add file3
-    ○  royxmykxtrkr false Add file2
-    ○  qpvuntsmwlqt false Add file1
+    ○  qpvuntsmwlqt false Add file2
+    ○  royxmykxtrkr false Add file1
     ◆  zzzzzzzzzzzz true
     [EOF]
     ");
@@ -392,7 +392,7 @@ fn test_split_with_descendants() -> TestResult {
     JJ: Enter a description for the selected changes.
     Add file1 & file2
 
-    JJ: Change ID: qpvuntsm
+    JJ: Change ID: royxmykx
     JJ: This commit contains the following changes:
     JJ:     A file1
     JJ:
@@ -403,7 +403,7 @@ fn test_split_with_descendants() -> TestResult {
     JJ: Enter a description for the remaining changes.
     Add file1 & file2
 
-    JJ: Change ID: royxmykx
+    JJ: Change ID: qpvuntsm
     JJ: This commit contains the following changes:
     JJ:     A file2
     JJ:
@@ -414,12 +414,12 @@ fn test_split_with_descendants() -> TestResult {
     // - The initial empty commit.
     // - The rewritten commit from the snapshot after the files were added.
     // - The rewritten commit once the description is added during `jj commit`.
-    // - The rewritten commit after the split.
-    let evolog_1 = work_dir.run_jj(["evolog", "-r", "qpvun"]);
+    // - The rewritten commit after the split with a new change ID.
+    let evolog_1 = work_dir.run_jj(["evolog", "-r", "royxm"]);
     insta::assert_snapshot!(evolog_1, @"
-    ○  qpvuntsm test.user@example.com 2001-02-03 08:05:12 74306e35
+    ○  royxmykx test.user@example.com 2001-02-03 08:05:12 e13e94b9
     │  Add file1
-    │  -- operation ec53c39a72a8 split commit 1d2499e72cefc8a2b87ebb47569140857b96189f
+    │  -- operation a46f740694f0 split commit 1d2499e72cefc8a2b87ebb47569140857b96189f
     ○  qpvuntsm/1 test.user@example.com 2001-02-03 08:05:08 1d2499e7 (hidden)
     │  Add file1 & file2
     │  -- operation 8185a834cefe commit f5700f8ef89e290e4e90ae6adc0908707e0d8c85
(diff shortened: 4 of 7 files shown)


cli/tests/test_split_command.rs
```diff
@@ -383,9 +381,9 @@ fn test_split_with_descendants() {
     // - The rewritten commit from the snapshot after the files were added.
     // - The rewritten commit once the description is added during `jj commit`.
     // - The rewritten commit after the split.
```
cenviity, line 383: Suggestion: Add ‘[…] with a new change ID’.

cli/tests/test_split_command.rs
```diff
@@ -537,9 +535,9 @@ fn test_split_parallel_no_descendants() {
     // - The initial empty commit.
     // - The rewritten commit from the snapshot after the files were added.
     // - The rewritten commit after the split.
```
cenviity, line 537: Same here. Suggestion: Add ‘[…] with a new change ID’.

cli/src/commands/split.rs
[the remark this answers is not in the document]
glehmann: > It also still removes the test which should be kept for the time being.

That wasn't supposed to happen. Thanks for noticing!

cli/src/commands/split.rs
```diff
@@ -300,6 +309,9 @@ pub(crate) fn cmd_split(
     if legacy_bookmark_behavior {
         tx.repo_mut()
             .set_rewritten_commit(target.commit.id().clone(), second_commit.id().clone());
+    } else {
+        tx.repo_mut()
+            .set_rewritten_commit(target.commit.id().clone(), first_commit.id().clone());
```
yuja, line 314: I don't closely follow the `-A/-B/-d` discussion, but it seems odd that we need to add this workaround to non-"legacy" code path. Suppose our goal is to remove special case like this, I think we should either make the first commit inherit the original change id (so descendants will be rebased onto it), or make the second commit inherit both change id and descendants.

cli/src/commands/split.rs
```diff
@@ -221,10 +221,17 @@ pub(crate) fn cmd_split(
     // Prompt the user to select the changes they want for the first commit.
     let target = select_diff(ui, &tx, &target_commit, &matcher, &diff_selector)?;
 
+    let legacy_change_id_behavior = tx.settings().get_bool("split.legacy-change-id-behavior")?;
```
yuja, line 224: nit: remove the other `let legacy_change_id_behavior`

cli/src/commands/split.rs
```diff
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
```
yuja, line 1: nit: bad merge?

cli/src/commands/split.rs
```diff
@@ -94,7 +138,7 @@ impl SplitArgs {
     /// command. Returns an error if the command cannot proceed.
     fn resolve(
         &self,
-        ui: &Ui,
+        ui: &mut Ui,
```
yuja, line 141: nit: unneeded change

cli/src/commands/split.rs
glehmann: that's a weird place — not sure what happened!

cli/src/commands/split.rs
glehmann: I don't see any other `let legacy_change_id_behavior`. Did you confuse it with `let legacy_bookmark_behavior`? I did it when writing the code, and had some unexpected test failures :)

cli/src/commands/split.rs
glehmann: I think it's more the other way around: we don't need it for the “legacy” code.
At first I completely removed the legacy bookmark config and used this call to change the parent in the rewrite. Then I reintroduced the legacy bookmark config and had to set the parent explicitly in the rewrite, making this call useless.

But I wonder if I understood the intent of the legacy bookmark config properly. Was it to

1. keep the bookmark on the same change id;
2. or to have the bookmark on the first commit?

I assumed 2. If it's 1., then this PR already does that, and `legacy-bookmark-behavior=false` should only have an effect when used with `legacy-change-id=true`.

glehmann: from #5618, which introduced this behavior

> Since the first revision inherits the change id of the target revision, moving the bookmarks to the first revision is less surprising (i.e. the bookmarks stay with the change id). This no-implicit-move behavior also aligns with how jj abandon drops bookmarks instead of moving them to the parent revision.

so the intent was clearly option 1.

@PhilipMetzger would that work for you if the only way to have the bookmark on the first commit is to set both `legacy-bookmark-behavior=false` and `legacy-change-id-behavior=true`?

cli/src/commands/split.rs
yuja: > I don't see any other `let legacy_change_id_behavior`. Did you confuse it with `let legacy_bookmark_behavior`?

Oh, yes. Do we need two separate "legacy" options? I think they serve the same role. We have `legacy-bookmark-behavior` because in legacy mode, bookmarks (and the working copy) move forward to the second commit whereas the change id stays in the first. In non-legacy behavior, bookmarks should move/stay along with the change id.

cli/src/commands/split.rs
PhilipMetzger: > @PhilipMetzger would that work for you if the only way to have the bookmark on the first commit is to set both `legacy-bookmark-behavior=false` and `legacy-change-id-behavior=true`?

In my opinion, this change shouldn't introduce a flag since it is controversial by itself and we ideally only should support one variant of `jj split -r X -B X` or `jj split -r X -A X` as a default, this PR makes `jj split -r X -B` the default. See Martins comment in your other PR https://github.com/jj-vcs/jj/pull/6458#discussion_r2074254264. I'd also hold off on landing this until we have a consensus on #3419.

cli/src/commands/split.rs
glehmann: In my opinion, we only need this PR, without any config ;-)

Obviously not everyone agrees, so I'll probably step back and wait for a consensus on #3419

cli/src/commands/split.rs
glehmann: I don't think it's about `jj split -r X -B X` vs `jj split -r X -A X`, but rather which commit keeps the change id: the one with the selected changes or the other one?

With `jj split -A/-B/-d` and the selected changes that may be moved anywhere, the commit with the non-selected changes is the only one that makes sense (to me) to keep the change id.
That's why I originally included it in the `jj split -A/-B/-d` PR.

We could also discuss if the commit with the selected changes should go before or after the other commit _by default_, but that's not the point of this PR.

Anyway, you seem quite opposed to that change, and I don't remember why. I'll see if I find that info in #3419, and maybe continue the discussion there.

cli/tests/test_split_command.rs
```diff
@@ -1210,3 +1501,38 @@ fn test_split_with_bookmarks(bookmark_behavior: BookmarkBehavior) {
...
+fn test_split_legacy_change_id_behavior() {
+    let test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let work_dir = test_env.work_dir("repo");
+
+    work_dir.write_file("file1", "foo\n");
+    work_dir.write_file("file2", "bar\n");
+    work_dir.run_jj(["describe", "-m", "my feature"]).success();
+
+    let output = work_dir.run_jj([
```
cenviity, line 1515: We need to snapshot the output and log output before running this command to demonstrate the legacy behaviour of the revision with the selected changes (i.e. the 'first commit') keeping the change ID.

cli/tests/test_split_command.rs
```diff
@@ -1086,6 +1086,296 @@ fn test_split_with_non_empty_description_and_trailers() {
...
+        "split",
+        "-m",
+        "file1",
+        "-r",
+        "qpvuntsmwlqt",
+        "--insert-before",
+        "qpvuntsmwlqt",
+        "file1",
+    ]);
```
cenviity, line 1183: _(not actionable)_

See my comment at <https://github.com/jj-vcs/jj/pull/6458#discussion_r2078704162> comparing the same set of tests here with differing log outputs for why I think the revision with remaining changes (i.e. the 'second commit') keeping the change ID makes more sense.

Here, the command indicates we are inserting a revision **before** `qpvuntsmwlqt`, and that is what the log output below shows – `vruxwmqvtpmx` is the newly inserted revision with the changes that have been split out.

emilazy: I haven’t been following the latest discussions, so just to reiterate my quite strong view at this point: `jj split -r X -d/-B/-A …` is creating a new commit from extracted parts of `X`. The commit that remains behind, containing the parts that were *not* split out, should be the one to keep the change ID and bookmarks. So in `jj split -r X -B X`, the second commit should inherit the change ID; in `jj split -r X -A X`, the first commit should. (I’d rather not speak of “the first/second commit” at all, for this reason, but rather the original commit and the split‐out commit, or similar.)

I think we have pretty strong consensus around this at this point, when operating within the `-d/-B/-A` framework (but if anyone disagrees they should say so). I believe the remaining disagreement is only about what the default action should be.

I haven’t read the PR or played with the implementation to check if it agrees with my intuitions on how the flags should work, so I don’t know if this comment is useful.

cli/src/commands/split.rs
yuja: > With `jj split -A/-B/-d` and the selected changes that may be moved anywhere, the commit with the non-selected changes is the only one that makes sense (to me) to keep the change id.

That makes sense. Specifying `-A/-B/-d` effectively means that the "legacy" option is turned off. I think it's similar to how we implemented `jj duplicate -A/-B/-D`.

cenviity:
See my other comment at <https://github.com/jj-vcs/jj/pull/6458#issuecomment-2860861884> for a suggestion on cleaning up the ‘first commit’ / ‘second commit’ terminology.

cli/tests/test_split_command.rs
cenviity: > I haven’t been following the latest discussions, so just to reiterate my quite strong view at this point: `jj split -r X -d/-B/-A …` is creating a new commit from extracted parts of `X`. The commit that remains behind, containing the parts that were _not_ split out, should be the one to keep the change ID and bookmarks.

In case it weren't apparent, I agree and am in favour of this PR.

> So in `jj split -r X -B X`, the second commit should inherit the change ID; in `jj split -r X -A X`, the first commit should.

For the benefit of others following this thread, IIUC I think you meant to say 'the second commit' in both cases, in the context of this PR and the previous paragraph where 'the second commit' is the revision with the parts that weren't split out (the remaining changes). Maybe here you meant 'the second commit' as in the **child** commit after `jj split -r X -B X` and 'the first commit' as in the **parent** commit after `jj split -r X -A X`.

> (I’d rather not speak of “the first/second commit” at all, for this reason, but rather the original commit and the split‐out commit, or similar.)

Agreed. See my comment at <https://github.com/jj-vcs/jj/pull/6458#issuecomment-2860861884> for cleaning up the terminology now that `jj split -A`/`-B`/`-d` is possible and the 'order' of the 'first' and 'second' commits is less clear.

emilazy: > Maybe here you meant 'the second commit' as in the **child** commit after `jj split -r X -B X` and 'the first commit' as in the **parent** commit after `jj split -r X -A X`.

Right. (A good example of why the terminology becomes inadequate with the new flags.)

cli/src/commands/split.rs
PhilipMetzger: > With `jj split -A/-B/-d` and the selected changes that may be moved anywhere, the commit with the non-selected changes is the only one that makes sense (to me) to keep the change id.
> That's why I originally included it in the `jj split -A/-B/-d` PR.

I still agree with that. 

> We could also discuss if the commit with the selected changes should go before or after the other commit _by default_, but that's not the point of this PR.
> 
> Anyway, you seem quite opposed to that change, and I don't remember why. I'll see if I find that info in #3419, and maybe continue the discussion there.

I currently don't think that `rebase -r X -B X` should be default which happens to be the current implementation, although this opinion is a minority but that's talk for #3419.

ilyagr (commented):
Thanks, @glehmann for working on this and on #6458! Even though I missed most of the discussion, I really appreciate it! 

Upon a quick test, unless I missed something, this PR's behavior at the moment is exactly what I thought the default behavior should be. So, it SGTM, but it sounds like we need to decide between this and Philip's suggestion of having no default behavior (see my other comment; perhaps there are also other suggestions I am less aware of), and there's no rush on that.

Whatever we decide in the end, having the option to play with all the behaviors we were discussing is really nice!

cli/src/commands/split.rs
PhilipMetzger: > If we decide to not go with this PR, Philip's suggestion (as I understood it) to have no default and to rely fully on `-A`/`-B`/`-d` seems reasonable to me. IMO (as a bit of a minor aside), that would be blocked on significantly improving[1](#user-content-fn-docs-d0b305c19073cea8256498cbad2545d1) the docs for `-A`/`-B`/`-d`, which would be good regardless, but would become urgent and crucial if we remove the default.

That's correct, but as I mentioned in the issue it should be a temporary stopgap before the users (people who are less involved in the decision process) decide on which behavior they want, this can take up to at least 6 months but that should be no issue (it is extremely likely that `-r X -B X` will win, since `-r X -A X` is a minority position).

cenviity (commented):
> With the -A/-B/-d options, the split-out commit may be moved anywhere.
> With commit tree
> 
>     @  pmpklomk (empty) (no description set)
>     ○  pzskstlk *letters* fileB
>     ○  kmvzrqtu fileA
>     │ ○  lyzvpymy *numbers* file3&C
>     │ ○  kpmxxkwk file2
>     │ ○  xzvtlrrt file1
>     ├─╯
>     ◆  zzzzzzzz root()
> 
> when splitting `fileC` out of `lyzvpymy` to the other branch with
> 
>     jj split -r lyzvpymy -d pzskstlk
> 
> the change id and the bookmark stay in the branch with the source commit.
> 
>     @  pmpklomk (empty) (no description set)
>     ○  vowztxqo fileC
>     ○  pzskstlk *letters* fileB
>     ○  kmvzrqtu fileA
>     │ ○  lyzvpymy *numbers* file3
>     │ ○  kpmxxkwk file2
>     │ ○  xzvtlrrt file1
>     ├─╯
>     ◆  zzzzzzzz root()

I noticed that in the current commit message the command says `-d pzskstlk` but the resulting tree looks like `-A pzskstlk` was applied instead (the new change `vowztxqo` is inserted between `pzskstlk` and its child. So, either the command or the resulting tree needs to be updated.

Another nit – perhaps use diff summaries rather than descriptions to show the before and after states? The description for `lyzvpymy` should remain the same after the `split`.

ilyagr:
Is this issue with `-d` related to this PR (beyond its commit description) or should it be a separate issue?

Perhaps we should remove the discussion of `-d/-A/-B` from the commit description, assuming that this PR does not change what `-d/-A/-B` do?

cli/src/commands/split.rs
cenviity: Currently `jj split -r X` (i.e. the default) and `jj split -r X -B X` both insert a new commit before `X`, but the change ID behaviour is different. So it would make sense to me either to a) remove the default and warn if none of `-A`/`-B`/`-d` are supplied or b) keep the default and merge this PR to make the change ID behaviour consistent.

cenviity:
> Is this issue with `-d` related to this PR (beyond its commit description) as it is now, or should it be a separate issue?
> 
> Perhaps we should remove the discussion of `-d/-A/-B` from the commit description, assuming that this PR does not change what `-d/-A/-B` do?

I think the example is useful to demonstrate that the change ID `lyzvpymy` stays in place rather than get taken by the newly created commit, which is what this PR changes. It's just that the argument used in the command there doesn't match the output tree.

yuja:
Is this PR still relevant for review? or are we just discussing a potential default to be introduced later? iirc, the idea was to try out the new `-A/-B/-d` flags for a couple of weeks, and pick one as the default.

glehmann:
iirc, the idea was to try out the new `-A/-B/-d` flags for a couple of weeks, and pick one as the default.

yes, no hurry to review this PR

glehmann:
> I noticed that in the current commit message the command says `-d pzskstlk` but the resulting tree looks like `-A pzskstlk` was applied instead (the new change `vowztxqo` is inserted between `pzskstlk` and its child. So, either the command or the resulting tree needs to be updated.

Good catch!
On the other hand, I think that @ilyagr is right: this PR is not about the behavior with `-A/-B/-d`. This behavior is already this one when using these flags.
I'll use another example that doesn't use the `-A/-B/d` flags

cenviity:
> On the other hand, I think that @ilyagr is right: this PR is not about the behavior with `-A/-B/-d`. This behavior is already this one when using these flags.

Ah yes, I'd forgotten that the new flags already exhibit the behaviour in this PR.

ilyagr:
Thanks for updating this; I use it and it makes it easier to use. :)

glehmann:
> Thanks for updating this; I use it and it makes it easier to use. :)

You're welcome!

I'm using the `-A/-B/-d` flags even more than I thought I would, but when I'm not using them, I'm indeed very pleased to have `jj` compiled with this change on my computer :)
I hope it will make it to the main branch eventually.

In the meantime, I'll continue to refresh this PR (hopefully with all the tests passing!).

glehmann:
Some time has passed since merging the `-d/-A/-B` flags, and hopefully, we have had a chance to experiment with the new split behavior.
@jj-vcs/maintainers Do you have any opinion on merging this PR?

martinvonz:
I have personally found the new flags very useful for extracting parts of a commit to an independent commit (`-d main`). I haven't tried to use them in place of the default behavior, however. My guess is that I would generally want the change id and the bookmark to stay on the parent commit, however. Maybe others have tried `-A/-B @`?

dnnr:
I wondered today what I could do to get this exact behavior as my default and I'm glad to see that this PR exists. For what it's worth, I'd appreciate seeing this merged! :pray:

emilazy:
FWIW, I still regularly feel pain from the lack of this and would very much like to either see it land, or see the default `split` behaviour dropped entirely. I think if we have a default it must correspond to one of the destination flags, and the `-B` this change implements is both the closest to the current behaviour, the only option that never results in conflicts, and the option that is consistent with the order we prompt for descriptions in.

Dropping the default and requiring an explicit specification would be an acceptable alternative, but IMO at a high cost to convenience. I won't repeat all the pages of earlier discussion from elsewhere on Discord/GitHub, but I agree with @ilyagr's comment from a while ago that `-r X -B X` is the "one true split" if you are going to privilege anything at all - as evidenced by the fact that the other destination flag options are built on top of it and correspond to rebasing the resulting new commit away.

Would be great if maintainers could make a decision here.

martinvonz:
At Google, we do stacked diffs and use a bookmark to identify each review in the stack. In that workflow, it seems clear from internal feedback that most people expect the parent to keep the change id and the bookmark. I'm pretty sure they still expect descendants to be rebased onto the child, though.

martinvonz:
I just noticed that we have 31 upvotes on the internal tracking bug for making the bookmark end up on the first commit. That's our 3rd most upvoted bug (after "support copies and renames" and "make `jj fix` format only changed lines").

glehmann:
I use stacked diffs too—probably not at the google scale though!—with the behavior from this PR.

IMO what is important is to have to bookmark that stays on the same revision, which is not currently the case, but is the case either with `split.legacy-bookmark-behavior` or with this PR.
Am I missing something?

martinvonz:
I haven't had time to re-review this PR so I'm not sure, but the users at Google seem to want the bookmark to stay on the parent commit.

matts1:
Speaking for gerrit users, with my new behavior added in `jj gerrit upload`, it directly maps the `jj` change ID to a gerrit change.

This means that which of the two commits keeps the change ID matters. And not only does it matter, but sometimes I want the parent commit to be associated with a change, and other times I want the child commit.

Our old internal system at google would seem to match this train of thought. I forget the precise workflow since I haven't used fig in ages, but I think when you ran `hg split`, it asked "which of the two commits do you want to associate with the exsiting CL uploaded for code review".

My personal thoughts is that I don't really care whether we choose the first or the second commit to associate with the old change by default. However, I strongly advocate that this be a command-line flag because the same user could want different choices at different times.

Also, I strongly believe that whichever commit retains the Change ID should retain the bookmarks.

martinvonz:
I'm still not sure which commit I think should keep the change id when doing a regular `jj split -r` (i.e. not `-o/-A/-B`), but this PR actually seems like an improvement regardless because I think everyone(?) agrees that the bookmark should follow the change id. Any objections? I haven't re-read all the discussion above.

glehmann:
0.36.0 has just been released. What would you think of merging this and collecting feedback during the next weeks before the next release?

glehmann:
It's probably too late for this month, with the holidays approaching. Can we agree to do it next month?

cenviity:
@glehmann Would it be a good idea to update the PR title to align with the current version of the commit message, so it is clearer we are referring to the commit with the remaining changes? With the availability of the `-o`/`-A`/`-B` flags it is less clear nowadays what the ‘first’ commit refers to.

zx8:
I haven't followed along with all the discussions behind this change so apologies if I'm treading old ground, but given it's been 8 months and maintainers obviously aren't 100% comfortable to move forward with merging this, is there any harm in just adding a (permanent) config knob that allows users to specify their desired behavior, to keep people in both camps happy?

PhilipMetzger:
> I haven't followed along with all the discussions behind this change so apologies if I'm treading old ground, but given it's been 8 months and maintainers obviously aren't 100% comfortable to move forward with merging this, is there any harm in just adding a (permanent) config knob that allows users to specify their desired behavior, to keep people in both camps happy?

It has been over 11 months with the previous change which got reverted because some people preferred this model. Also each permanent config knob makes the behavior of `jj` less intuitive to new users, so I'd rather have a hard stance from the maintainers on this.

matts1:
No-one seemed to take issue with my earlier proposal that we have an option on the split command to select which one keeps the change ID.

Changing the default and adding a config setting to set the default are potential options, but are very contentious. I'm honestly a little skeptical this discussion will go anywhere because there isn't a right answer.

Maybe we should just start by adding the option to split and see where things go from there? We might find that, for example, the people who are advocating for one specific behavior just want to be *able* to do that, rather than needing it to be the default. And even if they still don't like it, at the very least, this way, they can do what I did with `rebase --skip-emptied` and just do something like alias `sp` to `["split", "--remaining-changes-keeps-change-id-flag"]`.

martinvonz:
I'll try to summarize the arguments I can remember for each side:

For leaving change & bookmarks on parent commit:
* Usually what you want when doing per-commit reviews (at least in my experience)
* Makes `jj split && jj squash` preserve the change
 
For leaving the change & bookmarks on the child commit:
* Usually what you want  in PR-based reviews (IMO, we should not optimize for this)
* Consistent with where the descendants end up (they're presumably parented on the child commit regardless of choice)

Anything else I'm missing?

I also wonder if the `-o/-A/-B` options are too different from the regular kind of parent/child split so we should move them to a new `jj extract/evict` command.

joyously:
> I don't really care whether we choose the first or the second commit to associate with the old change by default. However, I strongly advocate that this be a command-line flag because the same user could want different choices at different times.
> 
> Also, I strongly believe that whichever commit retains the Change ID should retain the bookmarks.

I think this comment is accurate, and it got 5 +1 emojis.
The whole concept of `split` is to make two out of one, and every situation could be different. To me, that means either a default, a prompt, or a flag...or a mixture.

> For leaving the change & bookmarks on the child commit:
> 
>     * Usually what you want  in PR-based reviews (IMO, we should not optimize for this)

I may be misinterpreting this comment, but it seems that the `jj` philosophy is a "change" is a unit of work, which would be equivalent to a PR. Regardless of optimizing for the child keeping the ID, there are times when splitting off a small part should be the new ID, and this might more often coincide with the `-o,-A,-B` flags.

salty-horse:
> I also wonder if the `-o/-A/-B` options are too different from the regular kind of parent/child split so we should move them to a new `jj extract/evict` command.

I was just "hit" by misusing `split` (not for the first time), expecting the default behavior to work like the flags do, with the "selected changes" being the ones I want to split.
For a command called `split`, I would have expected the selected lines to be "split off", rather than "kept back".

IIRC, in other commands, the `-o`/`-A`/`-B` just control where something goes, without fundamentally altering how the command works.

I don't mind the functionality being moved/replicated in a new command, assuming too many people are used to the current default behavior.

martinvonz:
Sorry that we let this wait for so very long. We have finally discussed it among the maintainers, and although not all maintainers have chimed in, I think it's safe to say that we agree to go ahead with this PR. It will mean that bookmarks, working copies, and descendants follow the change id, making `jj split` consistent with other commands. It's also nice that `jj split` becomes effectively the same thing as `jj split -B @` (rather than the special case it is today). So I'll try to get around to reviewing this PR as soon as I can so we can get this merged.

CHANGELOG.md
```diff
@@ -115,6 +115,11 @@ None
   revisions by default (defined by `revsets.op-diff-changes-in`). A new flag,
   `--show-changes-in`, can be used to override this. [#6083](https://github.com/jj-vcs/jj/issues/6083)
 
+* `jj split` assigns the change id and the bookmarks of the source revision
+  to the revision with the non-selected changes.
+  You can opt out of this change by setting `split.legacy-bookmark-behavior = true`,
+  but this will likely be removed in a future release.
+
```
martinvonz: Don't forgot to move to the unreleased section when you rebase

CHANGELOG.md
```diff
@@ -115,6 +115,11 @@ None
   revisions by default (defined by `revsets.op-diff-changes-in`). A new flag,
   `--show-changes-in`, can be used to override this. [#6083](https://github.com/jj-vcs/jj/issues/6083)
 
+* `jj split` assigns the change id and the bookmarks of the source revision
```
martinvonz: nit: I think we try to use uppercase "ID" in user-facing docs and messages

CHANGELOG.md
```diff
@@ -115,6 +115,11 @@ None
...
+  to the revision with the non-selected changes.
```
martinvonz: nit: Can drop "and the bookmarks" since that doesn't change AFAICT.

Also, since we're only changing the behavior of the `jj split [-r]` flavor, where the split results in a parent and child commit, let's say "assigns the change ID of the source revision to the child commit". That seems clearer to me. (Let me know if you disagree, of course.)

cli/src/commands/commit.rs
martinvonz, line 1: From commit description:

> split: the commit with the remaining changes keeps the change id

Since this only affects the `jj split [-r]` case, "split: the child commit keeps the change id" seems clearer. Similarly, the next sentence could say "The parent commit gets a new change id.". Similar in the sentence after that.

cli/src/commands/commit.rs
```diff
@@ -47,9 +47,9 @@ use crate::ui::Ui;
 /// * `jj commit` doesn't have a `-r` option. It always acts on the working-copy
 ///   commit (@).
 ///
-/// * `jj split` (without `-o`/`-A`/`-B`) will move bookmarks forward from the
-///   old change to the child change. `jj commit` doesn't move bookmarks
-///   forward.
+/// * `jj split` creates the new commit as a parent of the source commit, and
+///   doesn't move the working-copy. `jj commit` creates the new commit as a
+///   child of the working-copy, and sets the new commit as the working-copy.
```
martinvonz: I find the "and doesn't move the working-copy" bit confusing. I understood what you meant only after reading the sentence about `jj commit`. Perhaps move that sentence first? Something like this:

```
/// * `jj commit` creates the new commit as a child of the working-copy, and
/// sets the new commit as the working-copy. `jj split` creates the new commit
/// as a parent of the source commit and leaves the working copy on the child commit.
```

cli/src/commands/split.rs
martinvonz: Why did this change? It seems to me like it should be mentioned in the description, and maybe it should be moved to a separate commit.

cli/src/commands/split.rs
```diff
@@ -292,11 +292,14 @@ pub(crate) async fn cmd_split(
...
 
+    let legacy_bookmark_behavior =
+        !use_move_flags && tx.settings().get_bool("split.legacy-bookmark-behavior")?;
+
     // Create the first commit, which includes the changes selected by the user.
     let first_commit = {
         let mut commit_builder = tx.repo_mut().rewrite_commit(&target.commit).detach();
         commit_builder.set_tree(target.selected_tree.clone());
-        if use_move_flags {
+        if !legacy_bookmark_behavior {
```
martinvonz: This feels like an independent fix that should be its own commit. I think it was never intended that `split.legacy-bookmark-behavior` should affect `-o/-A-/-B`. Should we also pass `legacy_bookmark_behavior` into `rewrite_descendants()` for the same reason (to make it not apply when `-o/-A/-B` is used)?

matts1:
Are we able to have a flag for which one keeps the change ID? I mentioned this earlier and people seemed happy with the idea. It's particularly relevant for gerrit and internal at google, where the change ID is associated with the code review, so sometimes you want to have the first commit associated with the code review, and sometimes the second.

martinvonz:
> Are we able to have a flag for which one keeps the change ID?

I suppose you mean that the option would make the first commit keep the change id and bookmarks but not working copies and descendants? I think that sounds reasonable. That can come after this PR, IMO.

yuja:
> I suppose you mean that the option would make the first commit keep the change id and bookmarks but not working copies and descendants? I think that sounds reasonable. That can come after this PR, IMO.

Implementation-wise, it seems weird that bookmarks and working copies go different locations. They are moving pointers to commits.

EDIT: oh, but it wouldn't hurt to add a special case similar to the one we have right now in commands/split.rs.

CHANGELOG.md
glehmann: Your proposition may not be accurate depending on the options used—for example with `jj split -A@`.
Maybe rather focus on the consistent behavior independently of the flags, and the non-moving bookmark behavior?

```markdown
* `jj split` now consistently assigns the source revision's change ID to the
  revision containing the non-selected changes—and ensures any associated
  bookmark references that same change ID—regardless of the options used.
```

glehmann:
I've been overwhelmed with GitHub notifications, and managed to miss the ones from this PR.
Nice to see this PR moving forward! :)

cli/src/commands/split.rs
glehmann: I think I did this change to match the expected behavior. It's a bit old, I have to rethink about it.

cli/src/commands/split.rs
glehmann: I also think `split.legacy-bookmark-behavior` shouldn't affect `-A/-B/-o`, that's why `legacy_bookmark_behavior` has `!use_move_flags` in it, line 296.

glehmann:
About `split.legacy-bookmark-behavior`:

At first I implemted both `split.legacy-bookmark-behavior` and `split.legacy-changeid-behavior`, but someone suggested we didn't needed both, which I think is true, so I kept only `split.legacy-bookmark-behavior`:
* `true` means the current jj behavior, with or without this PR. The change-id goes to the revision with the selected changes, and bookmarks go to the revision with the remaining changes.
* `false` doesn't produce the same result with or without this PR:
  * currently: the change-id and the bookmarks go to the revision with the selected changes
  * with this PR: the change-id and the bookmarks go to the revisions with the remaining changes.

starting from (`#fileC#` is a bookmark)

```
@  zwpmvsnu #fileC# fileC
│  A fileB
│  A fileC
│
○  yvlyqonw fileA
♦  zzzzzzzz root()
```

The command `jj split -m fileB fileB` gives these results:

<table>
<tr>
 <th>
 <th>`legacy-bookmark-behavior=true`
 <th>`legacy-bookmark-behavior=false`
<tr>
 <td>jj (main)
 <td>
<pre><code>
@  nqtsxmyo #fileC# fileC
│  A fileC
│
○  zwpmvsnu fileB
○  yvlyqonw fileA
♦  zzzzzzzz root()
</pre></code>
 <td>
<pre><code>
@  mrumvkxw fileC
│  A fileC
│
○  zwpmvsnu #fileC# fileB
○  yvlyqonw fileA
♦  zzzzzzzz root()
</pre></code>
<tr>
 <td>jj (#6466)
 <td>
<pre><code>
@  qspkkvtt #fileC# fileC
│  A fileC
│
○  zwpmvsnu fileB
○  yvlyqonw fileA
♦  zzzzzzzz root()
</pre></code>
 <td>
<pre><code>
@  zwpmvsnu #fileC# fileC
│  A fileC
│
○  qttlqokm fileB
○  yvlyqonw fileA
♦  zzzzzzzz root()
</pre></code>
</table>

What really changed is the change-id behavior, but as the intent from #5618 was to keep the bookmark referencing the same change-id, I thought it was OK to keep the same config name.

Should we rather name the config `split.legacy-changeid-behavior`?

Issue #6440, opened 2025-05-03
`jj git clone --colocate` hangs forever on macOS if the built-in git macOS `fsmonitor` is enabled in the global or user .gitconfig


dljsjr:
## Description

This isn't necessarily an issue with JJ itself, but rather, an edge case in `git` that JJ is hitting because of the operations that it uses to create a colocated repo. I don't actually understand *why*, though.

You can see this by performing the git clone operations by hand:

```console
$ git config set --global core.fsmonitor true

$ git clone --no-checkout git@github.com:tim-janik/jj-fzf.git
Cloning into 'jj-fzf'...
remote: Enumerating objects: 1525, done.
remote: Counting objects: 100% (276/276), done.
remote: Compressing objects: 100% (74/74), done.
remote: Total 1525 (delta 264), reused 207 (delta 202), pack-reused 1249 (from 2)
Receiving objects: 100% (1525/1525), 307.26 KiB | 2.77 MiB/s, done.
Resolving deltas: 100% (922/922), done.

# here is where we replicate the JJ fetch being done as part of the colocated clone.
# i've set GIT_TRACE to show where the hang happens relevant to the fsmonitor

$ LC_ALL=C GIT_TRACE=1 /usr/local/bin/git --git-dir /Users/dougstephen/git/jj-fzf/.git fetch --verbose --prune --no-write-fetch-head --progress -- origin '+refs/heads/*:refs/remotes/origin/*'
11:08:59.850406 git.c:476               trace: built-in: git fetch --verbose --prune --no-write-fetch-head --progress -- origin '+refs/heads/*:refs/remotes/origin/*'
11:08:59.851538 run-command.c:667       trace: run_command: git fsmonitor--daemon start
11:08:59.851564 run-command.c:759       trace: start_command: /usr/local/opt/git/libexec/git-core/git fsmonitor--daemon start
11:08:59.861526 git.c:476               trace: built-in: git fsmonitor--daemon start
11:08:59.862760 run-command.c:667       trace: run_command: git fsmonitor--daemon run --detach --ipc-threads=8
11:08:59.862785 run-command.c:759       trace: start_command: /usr/local/opt/git/libexec/git-core/git fsmonitor--daemon run --detach --ipc-threads=8

# Ctrl-C

$ rm -rf jj-fzf

# disabling fsmonitor fixes this, though

$ git config set --global core.fsmonitor false

$ git clone --no-checkout git@github.com:tim-janik/jj-fzf.git
[...]

PR #6468, opened 2025-05-06
Workaround `git` bug effecting `jj git clone --colocate` on Mac

dljsjr:
This workaround addressed #6440.

The issue itself that this addresses is likely an upstream issue in `git` itself, as I can reproduce it using raw `git` commands (this is demonstrated on the bug ticket).

The cliff's notes for the issue are as follows:

If the macOS-specific `fsmonitor` is enabled in the Global or System `.gitconfig`, The `fetch` operation that `jj` uses during `jj git clone --colocate` will hang forever if it is invoked *outside* of the actual working copy.

~~The workaround here, then, is pretty brain-dead: Change the working directory before calling `fetch` via the subprocess.~~

~~To avoid forcing everything to suddenly change directories all over the place, I added a `with_working_dir` helper that can be used to wrap a closure with the relevant directory changes.~~

~~This helper is then used during the `fetch` operation that occurs during `jj git clone --colocate`.~~

~~I've attached a video showing the failure mode with the release version of JJ, and the behavior of the same operation under this patch:~~

~~https://github.com/user-attachments/assets/bc715496-0bdf-4e50-a195-a9d8631ecadb~~

I've updated this PR to use the approach proposed by @yuja, which makes a lot of sense.

The new approach is to explicitly set `core.fsmonitor-false` via git's `-c` flag if `target_os = "macos"` for any subprocess operation.

# Checklist

If applicable:
- (N/A) I have updated the documentation (README.md, docs/, demos/)
- (N/A) I have updated the config schema (cli/src/config-schema.json)
- (N/A) I have added tests to cover my changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed bugs
 
+* Work around a git issue that could cause subprocess operations to hang if the
+  `core.fsmonitor` gitconfig is set in the global or system gitconfigs.
+  [#6440](https://github.com/jj-vcs/jj/issues/6440)
+
 ### Packaging changes
 
 * Due to the removal of the `libgit2` code path, packagers should
--- a/lib/src/git_subprocess.rs
+++ b/lib/src/git_subprocess.rs
@@ -111,6 +111,17 @@ impl<'a> GitSubprocessContext<'a> {
         // root to Command::current_dir and then pass a relative path to the git
         // dir
         git_cmd
+            // The gitconfig-controlled automated spawning of the macOS `fsmonitor--daemon`
+            // can cause strange behavior with certain subprocess operations.
+            // For example: https://github.com/jj-vcs/jj/issues/6440.
+            //
+            // Nothing we're doing in `jj` interacts with this daemon, so we force the
+            // config to be false for subprocess operations in order to avoid these
+            // interactions.
+            //
+            // In a colocated repo, the daemon will still get started the first time a `git`
+            // command is run manually if the gitconfigs are set up that way.
+            .args(["-c", "core.fsmonitor=false"])
             .arg("--git-dir")
             .arg(&self.git_dir)
             // Disable translation and other locale-dependent behavior so we can


yuja:
I don't think `chdir()`ing the current process is right approach. iirc, we've tried something similar in git_subprocess.rs (to work around Windows UNC paths), and got weird test failure.

Can we disable Git's fsmonitor by passing `-c`? It wouldn't be useful in `jj` anyway.

dljsjr:
> Can we disable Git's fsmonitor by passing `-c`? 

Might work to temporarily disable it at fetch time with `-c`, I'll give that a try.

> It wouldn't be useful in `jj` anyway.

Not necessarily true since this bug only shows up when performing a colocated clone. You could make the argument that it's still useful in this situation for anything that relies on `git status` or `git status --porcelain` like external tools.

yuja:
> > It wouldn't be useful in `jj` anyway.
> 
> Not necessarily true since this bug only shows up when performing a colocated clone. You could make the argument that it's still useful in this situation for anything that relies on `git status` or `git status --porcelain` like external tools.

In that case, external `git status` users will spawn the daemon process, right? Since `jj` doesn't use Git's worktree API internally, I don't think it would make sense for `jj` to start Git's fsmonitor thingy.

dljsjr:
> > > It wouldn't be useful in `jj` anyway.
> > 
> > 
> > Not necessarily true since this bug only shows up when performing a colocated clone. You could make the argument that it's still useful in this situation for anything that relies on `git status` or `git status --porcelain` like external tools.
> 
> In that case, external `git status` users will spawn the daemon process, right? Since `jj` doesn't use Git's worktree API internally, I don't think it would make sense for `jj` to start Git's fsmonitor thingy.

I think we agree, I was just clarifying that the daemon wasn't *completely* without a use case.

dljsjr:
> Can we disable Git's fsmonitor by passing `-c`?

@yuja I've updated the PR to use this approach, and it *does* also provide a valid workaround to the bug.

CHANGELOG.md
```diff
@@ -90,6 +90,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Fixed bash and zsh shell completion when completing aliases of multiple arguments.
   [#5377](https://github.com/jj-vcs/jj/issues/5377)
 
+* Work around a git issue on macOS that could cause `fetch` operations to hang if the
+  macOS `fsmonitor` feature is enabled.
+  [#6440](https://github.com/jj-vcs/jj/issues/6440)
+
```
yuja, line 96: nit: you'll need to update the changelog after the release today.

lib/src/git_subprocess.rs
```diff
@@ -106,6 +106,20 @@ impl<'a> GitSubprocessContext<'a> {
...
+        // can cause strange behavior with certain subprocess operations.
+        // For example: https://github.com/jj-vcs/jj/issues/6440.
+        //
+        // Nothing we're doing in `jj` interacts with this daemon, so we force the
+        // config to be false for subprocess operations in order to avoid these
+        // interactions.
+        //
+        // In a colocated repo, the daemon will still get started the first time a `git`
+        // command is run manually if the gitconfigs are set up that way.
+        #[cfg(target_os = "macos")]
```
yuja, line 119: Remove `cfg()`? As we discussed, it doesn't make sense to spawn fsmonitor by `jj git fetch`/`clone`.

(fwiw, `if cfg!(..)` is preferred as long as the code compiles on all platforms.)

yuja (commented):
Thanks.

lib/src/git_subprocess.rs
dljsjr: My thought process was to limit the change in behavior to macOS since it's only an issue on macOS, but to your point, it's probably both fine and sensical to avoid the complexity of platform-based branch.

CHANGELOG.md
dljsjr: No problem, I'll do so once the release happens!

lib/src/git_subprocess.rs
```diff
@@ -106,6 +106,18 @@ impl<'a> GitSubprocessContext<'a> {
...
+        // can cause strange behavior with certain subprocess operations.
+        // For example: https://github.com/jj-vcs/jj/issues/6440.
+        //
+        // Nothing we're doing in `jj` interacts with this daemon, so we force the
+        // config to be false for subprocess operations in order to avoid these
+        // interactions.
+        //
+        // In a colocated repo, the daemon will still get started the first time a `git`
+        // command is run manually if the gitconfigs are set up that way.
+        git_cmd.args(["-c", "core.fsmonitor=false"]);
```
yuja, line 119: nit: can you inline this in `git_cmd.arg(..)....env()` method chain?

yuja (approved):

dljsjr:
@yuja CHANGELOG rebased, args have been inlined in to the method chain 🎉

yuja:
Maybe you didn't upload the latest version?

(It's also good to merge the change if your PR got approval, and the last-minute change is trivial.)

dljsjr:
> Maybe you didn't upload the latest version?

:facepalm: I had unsquashed changes
 
> (It's also good to merge the change if your PR got approval, and the last-minute change is trivial.)

Bad habit from `$dayjob` where force pushes reset reviews.

PR #6470, opened 2025-05-07
release: 0.29.0

thoughtpolice:
# Checklist

If applicable:


docs/releasing.md
```diff
@@ -13,7 +13,7 @@ Producing the list of contributors is a bit annoying. The current suggestion is
 to run something like this:
 
 ```shell
-root=$(jj log --no-graph -r 'heads(tags())' -T 'commit_id')
+root=$(jj log --no-graph -r 'heads(tags() & ::trunk())' -T 'commit_id')
```
yuja, line 16: nit: depending on your intent, `tags(glob:"v*.*.*")` might be better.

yuja (approved):

docs/releasing.md
thoughtpolice: That's probably a good idea to include anyway. In my case I sometimes have tags like `v0.99.0-aseipp-test` which I push to my fork to make binaries, and I think that would match that revset if it was more recent than the last upstream tag. Then you get multiple results from the `log` call, which screws up the `gh` call. So I think we probably want this in addition to the `trunk()` filter.

martinvonz (approved):
Thanks!

PR #6475, opened 2025-05-07
git: write change-id header by default

martinvonz:
# Checklist

If applicable:


cli/tests/common/test_environment.rs
```diff
@@ -66,6 +66,8 @@ impl Default for TestEnvironment {
             r#"
 [template-aliases]
 'format_time_range(time_range)' = 'time_range.start() ++ " - " ++ time_range.end()'
+[git]
+write-change-id-header = false
```
emilazy, line 70: Is this tests stuff to avoid changing all the hashes? I think we should drop it and regenerate them all in another commit/PR, to avoid the test environment drifting from the actual defaults. (After all, maybe the change ID logic will cause bugs.)

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * The old `libgit2` code path for fetches and pushes has been removed,
   and the `git.subprocess` setting along with it.
 
+* Change IDs are now stored in the Git commit by default. Disable by
+  setting `git.write-change-id-header` to `false`. Please let us know if you do.
```
emilazy, line 19: Does this count as a breaking change? Maybe it should be a release highlight instead (possibly with mention of the consequence of being able to sync change IDs, and caveats around forge rebases)?

emilazy (commented):
Yay.

cli/tests/common/test_environment.rs
martinvonz: Yes, agreed. As I said in the commit message, I did it this way for now to avoid churn if we need to roll it back. Maybe some time after next release it's become clear enough that we're not going to roll it back, so then we can drop the config and update the tests. I don't feel strongly at all about exactly when we do it. I'm even fine with doing it in this PR if people don't mind the risk of updating it back and forth.

CHANGELOG.md
martinvonz: I think you're right. I've moved it to "Release highlights". PTAL at the wording.

cli/tests/common/test_environment.rs
emilazy: Whoops, sorry for missing the commit message.

I think switching the tests over in a separate commit is certainly good, for reviewability. I would personally lean towards doing it in this PR and considering running the tests with the option on part of the testing we’d want to do over the next release period. Reverting should “just” be a matter of going over a tedious but automated snapshot diff, right? (IIRC the tests used to depend on exact commit hashes in places but I think that was reduced or removed?)

(My guess is that we won’t feel the need to revert this, since we’re already being quite conservative in a world where GitButler has been putting this stuff in headers for a while now.)

CHANGELOG.md
emilazy: LGTM. I would be tempted to explicitly mention that some forges will clobber it, but maybe that kind of thing is better kept out of release notes since it duplicates things that need to be in regular docs anyway. (I just anticipate “I upgraded and it doesn’t work with GitLab” type issues.)

lib/src/git_backend.rs
```diff
@@ -2359,15 +2359,17 @@ mod tests {
     // UserSettings type. testutils returns jj_lib (2)'s UserSettings, whereas
     // our UserSettings type comes from jj_lib (1).
     fn user_settings() -> UserSettings {
```
yuja, line 2361: nit: Can you leave a comment why `git.write-change-id-header` is turned off by default in tests? It seems odd that `user_settings()` with no suffix does more things than `user_settings_with_change_id()`.

yuja (approved):

lib/src/git_backend.rs
martinvonz: I've turned it on instead :)

cli/tests/common/test_environment.rs
martinvonz: I'm also pretty confident it will work. I went ahead and enabled the config in tests too.

CHANGELOG.md
martinvonz: Done. I mentioned GitLab and `git rebase` as examples of where change ids are not preserved.

martinvonz:
There are still several tests I need to fix. I'll let you know when it's done.

lib/tests/test_revset.rs
```diff
@@ -284,7 +284,7 @@ fn test_resolve_symbol_change_id(readonly: bool) {
     .map(ChangeId::from_hex);
     let mut commits = vec![];
     let mut tx = repo.start_transaction();
-    for (i, change_id) in iter::zip([133, 664, 840, 5085], change_ids) {
+    for (i, change_id) in iter::zip([0, 1, 2, 5085], change_ids) {
```
emilazy: I sort of wish we had a tests helper to automatically ~~mine Bitcoin~~ brute‐force commits with certain hash properties (unambiguous short prefix, colliding prefixes, etc.). When I’ve worked on changes that cause treewide hash differences (e.g. UTC timestamps), it’s a real pain to manually fiddle with magic numbers until the right thing happens.

On the other hand, it would complicate tests like these that observe them in terms of snapshots. But maybe moving them off snapshots – or at least to more precise snapshots – would alleviate that? In this case, the snapshots are purely to test that the brute‐force works, which could be dropped if we had a tested helper API for the purpose.

Of course I don’t want to make you do this now, but if it annoys you as much as me then maybe you’ll be amenable to the idea. I imagine an API like `make_commits_with_hash_prefixes(["", "", "", "040"], |nonce| …set_description(format!("test {nonce}"))…)`, with the contract that (a) every commit’s hash has its corresponding entry as a prefix and (b) commit hashes are unambiguous after their prefixes; i.e., in this case, none of the other commits would start with `040` or with a character that any of the others do, `make_commits_with_hash_prefixes(["04", "040"], …)` would never create a first commit that starts `040`, and `make_commits_with_hash_prefixes(["1", "1"], …)` would provide two commits that start with `1` but immediately diverge after that. (Perhaps we might need something more elaborate to allow accessing the previously‐generated commits if we want the commits to relate to each other in terms of ancestry; I’m not sure which cases like this we have. And of course in this case we would be doing unnecessary computational work to avoid collisions in the other commits, but it buys us API and implementation simplicity it’s probably very quick to brute‐force.)

emilazy (commented):
nit: Can this still be organized like before, with an initial commit that migrates the user settings but not the tests? That way the change with review‐worthy things like the changelog entry is separated out from the huge mechanical tree‐wide change.

It’s not a huge deal though, since even the mechanical commit will necessarily have logic changes for the tests – but those should at least themselves be mechanical reverts of the prophylactic changes to the test settings. GitHub makes it really unpleasant to view and add comments on such a large commit.

martinvonz:
> nit: Can this still be organized like before, with an initial commit that migrates the user settings but not the tests? That way the change with review‐worthy things like the changelog entry is separated out from the huge mechanical tree‐wide change.

I didn't bother because the second commit would then undo most of the changes from the first commit. There are very few non-test changes in the commit. FYI, they are in `CHANGELOG.md` and `lib/src/settings.rs` (there's a change in `lib/src/settings.rs`, but that's only for tests).

thoughtpolice (approved):
Yeah, I'd rather just keep things in a single commit tbh since this seems rare enough, and the code changes are tiny enough anyway that I doubt I'll need to take a second look.

Regarding review, in this case it's kind of easy enough to just filter out the tests in the file finder UX, since they exist in subdirectories, but it's unfortunate that GitHub's file finder only can take literal fuzzy matches and not ~~regular expressions~~ filesets, otherwise you could filter those out way more easily...

PR #6479, opened 2025-05-08
docs: clarify commit_trailers deduplication

prattmic:
I spent quite a while confused by the presence of
`if(!trailers.contains_key("Change-Id")` in the `commit_trailers` example. My thinking was: if `commit_trailers` already performs deduplication, then isn't this logic useless?

Eventually I realized that the `commit_trailers` deduplication works on the entire line, not just the key as I'd assumed. (And that makes sense for trailers like Signed-off-by that can appear more than once.)

I've reworded the `commit_trailers` documentation to try to make this more clear.

To be honest, I'm not thrilled with the way the wording turned out, so suggestions are welcome. In particular, I've tried to define each trailer as a "line in the template", but since `format_signed_off_by_trailer` and `format_gerrit_change_id_trailer` (helpfully!) include a newline, that definition isn't super obvious in the example.

cc @glehmann

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -240,9 +240,16 @@ is ignored.
 
 ### Commit trailers
 
-You can configure automatic addition of trailers to commit descriptions using
-the `commit_trailers` template. Trailers defined in this template will only be
-added if they are not already present in the description.
+You can configure automatic addition of one or more trailers to commit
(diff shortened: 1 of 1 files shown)


yuja (approved):
Thanks.

PR #6486, opened 2025-05-09
cli: warn if `git.write-change-id-header` is explicitly enabled

winterqt:
# Checklist

If applicable:

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -726,6 +726,16 @@ pub fn default_config_migrations() -> Vec<ConfigMigrationRule> {
                 Ok(format!(r#""{escaped}""#).into())
             },
         ),
+        ConfigMigrationRule::custom(
+            |layer| {
+                let Ok(Some(write_change_id)) = layer.look_up_item("git.write-change-id-header")
+                else {
+                    return false;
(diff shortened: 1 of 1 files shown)


cli/src/config.rs
```diff
@@ -726,6 +726,16 @@ pub fn default_config_migrations() -> Vec<ConfigMigrationRule> {
                 Ok(format!(r#""{escaped}""#).into())
             },
         ),
+        ConfigMigrationRule::custom(
+            |layer| {
+                let Ok(Some(write_change_id)) = layer.look_up_item("git.write-change-id-header")
+                else {
```
winterqt, line 732: No idea why `rustfmt` formatted it like that.

yuja:
Do we have any target release where `git.write-change-id-header` will be deleted? Without it, a warning would be rather noisy. Setting the default value should be valid. The user might turn it off globally, and re-enable it for a certain repositories, for example.

PhilipMetzger:
yes, this will only be needed _when_ we'll start to deprecate the config, so doing it now is a bit early.

winterqt:
Oops, yup.

PR #6491, opened 2025-05-09
github: try building windows-on-arm binaries

thoughtpolice:
# Checklist

If applicable:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: ${{ github.event_name == 'merge_group' }}
       matrix:
-        build: [linux-x86_64-gnu, linux-aarch64-gnu, macos-x86_64, macos-aarch64, windows-x86_64]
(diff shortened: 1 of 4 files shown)


thoughtpolice:
This should be working now, and the binaries seem to work. The ARM build is 25% faster or so than x86 too 🥲 (test time is the same, though)

.github/workflows/ci.yml
```diff
@@ -55,7 +58,7 @@ jobs:
       with:
         persist-credentials: false
     - name: Set up Windows Builders
-      if: startswith(matrix.os, 'windows')
+      if: startswith(matrix.os, 'windows-x86_64') # FIXME: aarch64 doesn't have D:\ yet
```
emilazy, line 61: Maybe the AArch64 VMs are different enough that it just doesn’t matter, but one option would be [setting up a Dev Drive](https://github.com/samypr100/setup-dev-drive). It seemed slower than the native `D:\` when I was messing with it for x86, but maybe it’d be a win on AArch64.

arxanas (approved):

strega-nil (approved):

PR #6495, opened 2025-05-09
Aec3 wc settings

wrzian:
As requested, I'm splitting up the commits in the executable bit PR into smaller stacked PRs so they can be reviewed separately.

See https://github.com/jj-vcs/jj/pull/4478#pullrequestreview-2746646793 for context.

---

This PR is focused on updating the way settings are passed to the working copy. We split out configurations from `SnapshotOptions` and `CheckoutOptions` that can be loaded solely from `UserSettings` into a new `WorkingCopySettings` struct, which is in to `start_working_copy_mutation`.

### DIFF
--- a/lib/src/local_working_copy.rs
+++ b/lib/src/local_working_copy.rs
@@ -112,6 +112,7 @@ use crate::working_copy::SnapshotStats;
 use crate::working_copy::UntrackedReason;
 use crate::working_copy::WorkingCopy;
 use crate::working_copy::WorkingCopyFactory;
+use crate::working_copy::WorkingCopySettings;
 use crate::working_copy::WorkingCopyStateError;
 
 /// On-disk state of file executable bit.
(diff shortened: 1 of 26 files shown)


lib/src/local_working_copy.rs
```diff
@@ -616,7 +616,7 @@ fn remove_old_file(disk_path: &Path) -> Result<bool, CheckoutError> {
     match fs::remove_file(disk_path) {
         Ok(()) => Ok(true),
         Err(err) if err.kind() == io::ErrorKind::NotFound => Ok(false),
-        // TODO: Use io::ErrorKind::IsADirectory if it gets stabilized
```
martinvonz, line 619: should the next line be deleted?

lib/src/local_working_copy.rs
```diff
@@ -408,7 +406,7 @@ impl<'a> FileStates<'a> {
     }
 
     /// Iterates file state entries sorted by path.
-    pub fn iter(&self) -> FileStatesIter<'a> {
+    pub fn iter(&self) -> impl Iterator<Item = (&'_ RepoPath, FileState)> {
```
martinvonz, line 409: nit: can this still use the `'a` lifetime?

lib/src/local_working_copy.rs
wrzian: This is part of PR 1, see discussion here: https://github.com/jj-vcs/jj/pull/6442#discussion_r2072432152

lib/src/local_working_copy.rs
wrzian: Yes. Updating

wrzian:
I'm closing this because I'm moving to a new account, @isuffix. I'll restart this PR from that account with more atomic PRs.

PR #6509, opened 2025-05-11
split: use "selected/remaining changes" instead of "first/second commit"

glehmann:
based on: https://github.com/jj-vcs/jj/pull/6458

The usage of "revision" and "commit" may not be fully consistent, as I'm not totally sure when to use one or the other.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -48,7 +48,7 @@ use crate::ui::Ui;
 /// Split a revision in two
 ///
 /// Starts a [diff editor] on the changes in the revision. Edit the right side
-/// of the diff until it has the content you want in the first revision. Once
+/// of the diff until it has the content you want in the new revision. Once
 /// you close the editor, your edited content will replace the previous
 /// revision. The remaining changes will be put in a new revision on top.
 ///
@@ -57,8 +57,8 @@ use crate::ui::Ui;
 ///
 /// If the change you split had a description, you will be asked to enter a
 /// change description for each commit. If the change did not have a
-/// description, the second part will not get a description, and you will be
-/// asked for a description only for the first part.
+/// description, the remaining changes will not get a description, and you will
+/// be asked for a description only for the selected changes.
 ///
 /// Splitting an empty commit is not supported because the same effect can be
 /// achieved with `jj new`.
@@ -124,7 +124,7 @@ pub(crate) struct SplitArgs {
     /// child
     #[arg(long, short)]
     parallel: bool,
-    /// Files matching any of these filesets are put in the first commit
+    /// Files matching any of these filesets are put in the selected changes
     #[arg(
         value_name = "FILESETS",
         value_hint = clap::ValueHint::AnyPath,
@@ -244,7 +244,7 @@ pub(crate) fn cmd_split(
             let template = description_template(
                 ui,
                 &tx,
-                "Enter a description for the first commit.",
+                "Enter a description for the selected changes.",
                 &temp_commit,
             )?;
             edit_description(&text_editor, &template)?
@@ -294,7 +294,7 @@ pub(crate) fn cmd_split(
             let template = description_template(
                 ui,
                 &tx,
-                "Enter a description for the second commit.",
+                "Enter a description for the remaining changes.",
                 &temp_commit,
             )?;
             edit_description(&text_editor, &template)?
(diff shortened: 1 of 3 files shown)


martinvonz (commented):
> the jj version I'm currently using

I suppose this commit was not supposed to be part of the PR

glehmann:
> > the jj version I'm currently using
> 
> I suppose this commit was not supposed to be part of the PR

Indeed, fixed!

PhilipMetzger:
can you move the relevant commit off your stack, so it can land immediately as both #6458 and #6466 are still under their own review?

glehmann:
> can you move the relevant commit off your stack, so it can land immediately as both #6458 and #6466 are still under their own review?

Unfortunately, no, because the wording I chose depends on the previous changes.

I'm not sure what the best workflow is here. If it's inconvenient to review due to the other PR it's based on, please just ignore it for now, and I'll bring it up (and possibly adjust it) once the others are accepted or rejected.

PhilipMetzger:
> > can you move the relevant commit off your stack, so it can land immediately as both #6458 and #6466 are still under their own review?
> 
> Unfortunately, no, because the wording I chose depends on the previous changes.

imo, it's just enough if you drop the #6466 commit from this PR since the rewording is revelant _after_ adding the destination args to _split_.

glehmann:
The command description actually describes the behavior after #6466. Sure I could update the PR to avoid that, but then there would be the conflicts with #6466…
Do you think it would be more relevant to rebase #6466 on this PR?

PhilipMetzger:
> The command description actually describes the behavior after #6466. Sure I could update the PR to avoid that, but then there would be the conflicts with #6466…

That isn't true since https://github.com/jj-vcs/jj/pull/6458#issuecomment-2860861884 was on #6458 not on #6466 which the project will hold off for a while. I don't know how you managed the PRs, but this should just be a simple rebase of #6466 on top of this commit.

glehmann:
> That isn't true since [#6458 (comment)](https://github.com/jj-vcs/jj/pull/6458#issuecomment-2860861884) was on #6458 not on #6466 which the project will hold off for a while.

I was talking about the command description as modified in this PR…

>  I don't know how you managed the PRs, but this should just be a simple rebase of #6466 on top of this commit.

That would be true if there were no tests.

PhilipMetzger:
> > That isn't true since [#6458 (comment)](https://github.com/jj-vcs/jj/pull/6458#issuecomment-2860861884) was on #6458 not on #6466 which the project will hold off for a while.
> 
> I was talking about the command description as modified in this PR…

I don't know but that seems like a smaller problem to me, since it still can be adjusted w/o having the [9a8c1c0](https://github.com/jj-vcs/jj/pull/6509/commits/9a8c1c0242322d0675a18fb430f627c9c0c3d500) commit. 

> > I don't know how you managed the PRs, but this should just be a simple rebase of #6466 on top of this commit.
> 
> That would be true if there were no tests.

true, but that also is solvable right? it may be a bit tedious but that's the case when rewording messages.

cenviity (commented):
Thanks for this. Just a few minor suggestions for now.

cenviity:
> > That isn't true since [#6458 (comment)](https://github.com/jj-vcs/jj/pull/6458#issuecomment-2860861884) was on #6458 not on #6466 which the project will hold off for a while.
> 
> I was talking about the command description as modified in this PR…
> 
> > I don't know how you managed the PRs, but this should just be a simple rebase of #6466 on top of this commit.
> 
> That would be true if there were no tests.

FWIW I think the tests (and commit description, etc.) in #6466 would benefit from rebasing #6466 on top of this PR and using the new terminology introduced here, to avoid confusing situations like in [this thread](https://github.com/jj-vcs/jj/pull/6466#discussion_r2078800092).

cenviity (commented):
nit: 'in place of "first/second commit"' (or 'instead of […]' as in this PR's title) should be part of the commit summary rather than the start of the description. It would make it clearer what we are renaming in a single-line log.

glehmann:
> nit: 'in place of "first/second commit"' (or 'instead of […]' as in this PR's title) should be part of the commit summary rather than the start of the description. It would make it clearer what we are renaming in a single-line log.

I'm aiming to follow the 50/72 maximum line length convention that many projects use for their commit messages... but jj doesn't seem to follow that convention :+1:

martinvonz (approved):

PR #6511, opened 2025-05-11
rename `--destination` to `--onto`

glehmann:
The flag `--onto` more accurately conveys the target location onto which
the commit will be rebased. Furthermore, it ensures greater consistency
with the other location flags, `--after` and `--before`, both of which
function as prepositions.

There is still a lot of documentation to update and testing to do. I've mostly modified the commands with a `--destination`/`-d` to ensure there wouldn't be any collision with an existing short flag. Moreover I'm not sure that setting `--destination`/`-d` as an alias is the right way to go.

I'm not willing to push another controversial change, so if anyone thinks that's not a good change, please tell me before I spend more time on it!

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/duplicate.rs
+++ b/cli/src/commands/duplicate.rs
@@ -70,18 +70,20 @@ pub(crate) struct DuplicateArgs {
     /// commit)
     #[arg(
         long,
-        short,
+        short = 't',
+        visible_alias = "destination",
+        visible_short_alias = 'd',
         value_name = "REVSETS",
-        add = ArgValueCompleter::new(complete::revset_expression_all),
+        add = ArgValueCompleter::new(complete::revset_expression_mutable),
     )]
-    destination: Option<Vec<RevisionArg>>,
+    onto: Option<Vec<RevisionArg>>,
     /// The revision(s) to insert after (can be repeated to create a merge
     /// commit)
     #[arg(
         long,
         short = 'A',
         visible_alias = "after",
-        conflicts_with = "destination",
+        conflicts_with = "onto",
         value_name = "REVSETS",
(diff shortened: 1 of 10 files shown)


cli/src/commands/duplicate.rs
PhilipMetzger, line 1: nit: should be `commands: rename --destination to --onto` or `cli commands:` whatever you prefer

cli/src/commands/duplicate.rs
```diff
@@ -70,18 +70,20 @@ pub(crate) struct DuplicateArgs {
     /// commit)
     #[arg(
         long,
-        short,
+        short = 't',
+        visible_alias = "destination",
+        visible_short_alias = 'd',
         value_name = "REVSETS",
-        add = ArgValueCompleter::new(complete::revset_expression_all),
+        add = ArgValueCompleter::new(complete::revset_expression_mutable),
```
PhilipMetzger, line 77: nit: why did you change this to all mutable revisions? I think all revisions is a better default, since you should be able to rebase your revisions anywhere in the repos history. This is for all such changes in this commit.

cli/src/commands/duplicate.rs
```diff
@@ -70,18 +70,20 @@ pub(crate) struct DuplicateArgs {
     /// commit)
     #[arg(
         long,
-        short,
+        short = 't',
```
PhilipMetzger, line 73: Nit: I'm not really sure about `-t` for `onto`, there's `-o` which will make it like a compiler invocation but I don't have a better suggestion.

PhilipMetzger (commented):
I have left some relatively minor stuff and some points which lead to longer discussions for now.

glehmann:
see #6512 instead

Issue #2554, opened 2023-11-08
FR: Fix inconsistencies in how subcommands handle revision and path args


necauqua:
My proposal is that every command that works with a revset and an optional list of paths should have the following form:

jj subcommand revset* (-p path+ | -r revset+ | other-args)*

Here the first revset is an optional repeating positional arg, followed by (potentially) interleaved -p/--path(s) and -r/--revisions arguments, I described it with a pseudo-regex here.

This way, any of the following is valid:
- `jj log`
- `jj log rev`
- `jj log rev1 rev2`
- `jj log -r rev`
- `jj log rev -p path1 path2`
- `jj log -r rev -p path1 path2`
- `jj log -p path1 path2 -r rev`
- `jj log -p path1 -r rev -p path2 path3`
- `jj log rev1 rev2 -p path1 path2`
- `jj log rev -r rev2 rev3`
- `jj log -p path1 -r rev -p path2 -r rev2`
- etc

Another way to describe it is that we consume "positional" args to be either a revset or a path depending on the state, which is initially "accept as revset", which can be switched to "accept as path" by `-p` and back to "accept as revset" by `-r` (but also `-p` or `-r` are not allowed at the end).

---

If command doesn't work with paths, we just dont have them, and now it's similar to how we have `unused_revision: bool` flag in some places (except it's not allowed at the end now too).
This should cover most commands, actually, all of them that work with commits I guess.

Multiple revsets also don't make sense for all commands - for those we can just check and error out - or we can join them by `|` and then check if they _resolve_ to a single commit.
Note that something like `jj new a b` should work, but if `a` diverges we must pass `all:` to signify that we're sure we want to merge a divergent branch or whatever that is and also add `b` into the mix - basically we don't want unexpected (where it matters, e.g. log is ok) revision resolutions without `all:`.

The main thing everything should be consistent :)

spectral54:
I'm not generally a huge fan of "argument that changes the behavior of the following arguments", unless I can convince myself that it's a mode selector for the entire command. So something like `tar -x` seems fine to me, because I treat that as equivalent to an "untar" command (or maybe because I just run it so frequently). Similarly, "subcommands" seem fine. But things like `find` and `ps` just have inscrutable behavior to me, so I end up memorizing the few usages I commonly need, and need to look up anything more than that.

[...]

PR #6512, opened 2025-05-11
cli: rename `--destination` to `--onto`

martinvonz:
Here's my version of #6511 from this morning :)

# Checklist

If applicable:


cli/src/commands/duplicate.rs
```diff
@@ -37,13 +37,13 @@ use crate::ui::Ui;
...
-/// When none of the `--destination`, `--insert-after`, or `--insert-before`
+/// When none of the `--onto`, `--insert-after`, or `--insert-before`
 /// arguments are provided, commits will be duplicated onto their existing
 /// parents or onto other newly duplicated commits.
 ///
-/// When any of the `--destination`, `--insert-after`, or `--insert-before`
+/// When any of the `--onto`, `--insert-after`, or `--insert-before`
 /// arguments are provided, the roots of the specified commits will be
-/// duplicated onto the destination indicated by the arguments. Other specified
+/// duplicated onto the onto indicated by the arguments. Other specified
```
cenviity, line 46: Bad replace?

cli/src/commands/new.rs
```diff
@@ -52,10 +52,10 @@ pub(crate) struct NewArgs {
         add = ArgValueCompleter::new(complete::revset_expression_all),
     )]
     revisions: Option<Vec<RevisionArg>>,
-    /// Ignored (but lets you pass `-d`/`-r` for consistency with other
+    /// Ignored (but lets you pass `-o/-d`/`-r` for consistency with other
```
cenviity, line 55:
```suggestion
    /// Ignored (but lets you pass `-o`/`-d`/`-r` for consistency with other
```

cli/src/commands/duplicate.rs
martinvonz: Yeah, probably. Fixed.

martinvonz:
@glehmann: I think my version looks more complete than yours but I'm fine with you copying over changes from my PR to yours if you prefer. I guess we were working on it independently without knowing.

cli/src/commands/rebase.rs
```diff
@@ -334,17 +334,19 @@ pub struct RebaseDestinationArgs {
     #[arg(
         long,
         short,
```
yuja, line 336: nit: use `-t` instead of `-o`? We have `--into`/`-t` in a couple of commands, and `--onto` sounds similar.

martinvonz: I like `-o` better, maybe especially because `--into` has different semantics. One could imagine `split` and `squash` getting superseded by some more generic `diffmove` command with both `--into` and `--onto` arguments.

yuja: Hmm, in `diffmove`, I feel it's confusing to use `--into` and `--onto` to differentiate operations. Both operations could be explained as "move $stuff to destination." Maybe better to find more descriptive name?

emilazy: I think that’s fine: with `--onto X` the destination is “a new commit on top of `X`” and with `--into X` the destination is “the patch of `X`”. (Treating those uniformly as destinations would be the whole point of the exercise – every command that permits `--onto` could permit `--into`.)

(I’m not saying `--onto` is the best possible name for this, but I don’t know if anyone has come up with anything better. I liked `--atop` but I might be the only one.)

joyously: I think 'source' and 'destination' are clearer than 'into' or 'onto'.

glehmann:
> @glehmann: I think my version looks more complete than yours but I'm fine with you copying over changes from my PR to yours if you prefer. I guess we were working on it independently without knowing.

 The conversations and modifications regarding the `-A/-B/-d` flags are likely sparking similar thoughts for several people :-)

Yours is indeed much more comprehensive. I'll close #6511.

cli/src/commands/rebase.rs
glehmann: I like `--onto` because the command can almost be read in English: `jj rebase --revision x --onto y`.

It's also the word used in jj's help message:

```
  -d, --destination <REVSETS>
          The revision(s) to rebase onto (can be repeated to create a merge commit)
 ```

For the short version, I proposed `-t` in my version of this PR, because of the similarity with `--into`/`-t` in squash. When I first started using jj, I found it difficult to remember to use `-d` for rebase and often instinctively tried `-t` instead. That being said, the semantics are different, and some commands may use both, so `-o` seems like a good option.

cenviity: > I like `-o` better, maybe especially because `--into` has different semantics. One could imagine `split` and `squash` getting superseded by some more generic `diffmove` command with both `--into` and `--onto` arguments.

With the recently landed `jj split -d`/`-A`/`-B` flags I have also been thinking about unifying `jj split` and `jj squash`, where it would make sense having both `--into` and `--onto` with different semantics. So I vote for `-o` over `-i`.

YPares: Maybe `--above` (or `--atop`) / `-a` would lead to less clashes?

But in case it's too late for that, my vote goes to `-o`.

AngelEzquerra: I don't like `-o` too much because it makes me think of "output". Also, it'd be odd to use `-o` as a shortcut for `--onto` but not use `-i` as a shortcut for `--into` (which I don't think is possible since `-i` is used for `--interactive`, I think?)
I like `-t` because I suspect that the distinction between `--into` and `--onto` is a bit subtle for non native English speakers. Being able to use `-t` to indicate a "target" would be nice.
My only concern is whether some commands might get both `--into` and `--onto` arguments, as suggested above. However, I would claim that we should above that since I think that'd be quite confusing (specially for non native English speakers).

yuja: Can we add some verb to differentiate `--into` and `--onto` (something like `--squash-into`)? I don't think the semantics carried by `split --into`/`--onto` is obvious enough for non-native speakers.

glehmann: The difference between 'into' and 'onto' is clear to me, despite being a non-native English speaker. I’m not claiming to represent the majority, but this may not be a valid argument for all non-native English speakers.

emilazy: I think the dictionary definitions make it clear: “To or towards the inside of.” vs. “Arriving upon or on top of (speaking of a physical or metaphorical movement).” – not to say that we should necessarily expect people to have them memorized, but in this case they’re common words and IMO unambiguous, so I think it’s okay to expect people to look them up if they’re confused.

Maybe the issue is just that they only differ by one letter? I would be okay with `--above` or `--atop` as @YPares suggested too, though. I proposed `--atop` on Discord but people didn’t seem to like it. I worry `--above` would be confusable with `--after` more than `--into` is with `--onto`. We could always do `--upon`!

joyously: > I think that’s fine: with `--onto X` the destination is “a new commit on top of `X`” and with `--into X` the destination is “the patch of `X`”.

I didn't follow this, because it seems like it would be different for `squash` and for `split`.
I don't see any reason to rename `--destination`, especially if it can be a tag or bookmark or revset or revision (I don't know if this parameter can...). 
We already have before and after, which are a little confusing because you can show the tree in reverse. Using more words with positional meanings (above, atop, onto, upon) is just more confusing. If the point is to distinguish between a new commit and an old patch, maybe use `--new` or `--patch`, but I never interpreted the existing options that way (maybe that's what I'm misunderstanding).

martinvonz: `--atop` with a short form of `-a` may also be easy to confuse with the existing `-A`

cenviity: > > I think that’s fine: with `--onto X` the destination is “a new commit on top of `X`” and with `--into X` the destination is “the patch of `X`”.
> 
> 
> 
> I didn't follow this, because it seems like it would be different for `squash` and for `split`.

`squash` could have `--after`/`--before`/`--destination` (or whatever the latter gets renamed to) to make it easier to squash changes into a new commit that didn’t exist prior to the operation, e.g. when squashing together diffs from multiple commits.

If we had `split --into`, it would be conceptually similar to `squash -i --into`. So `squash` and `split` have enough similarities to warrant a command that generalises over both.

yuja: might be better to add place expression than introducing various destination arguments that work either graph or patch content.

`jj squash -i -d 'after(rev)'` to insert split commit after `rev`, etc.

https://github.com/jj-vcs/jj/issues/4708

cenviity: > might be better to add place expression than introducing various destination arguments that apply to either graph or patch content.
> 
> 
> 
> `jj squash -i -d 'after(rev)'` to insert split commit after `rev`, etc.
> 
> 
> 
> https://github.com/jj-vcs/jj/issues/4708

How would that work with shell completions?

hotsphink: Other options:
* --on : better distinguishes from --into, no other advantage I see
* --upon / -u : distinct from --into, allows -u as short form
* --parent / -p : describes the purpose of the parameter, has a good short form, does not flow all that well when used in a sentence "rebase X onto its new parent(s) Y"
* --target / -t : current docs say "`--destination/-d` to rebase the revisions onto the specified targets"
* --adopter : leans into the analogy with ancestors/descendants/parents/children, because -r X --adopter Y means Y is the new parent of X. Cons: weird clunky word, -a conflicts with -A, does not work at all in a sentence

Modest proposal: --onto-parent / --onto / -P. `--insert-after` => `-A` helps allow `--onto-parent` => `-P`, and it's fine if nobody ever uses `--onto-parent` but only `--onto`; it can just exist as the full option name in order to justify `-P`.

Or if `-p` is preferred over `-P`, that's fine too. `-P` *would* have the side benefit of making all of the "target" options be uppercase, fwiw: `-A` / `-B` / `-P`.

cenviity: `-p` has been discussed in the past as the short form of `--path` (see #2554).

`-P`/`--onto-parent` might be nice though.

PhilipMetzger: I still prefer `--onto` if the project ever wants to go down the `rebase` => `move` rename, which IMO is a really good outocome to newbies to VCS.

glehmann:
@martinvonz You may want to close this one now that https://github.com/jj-vcs/jj/pull/7978 is merged

martinvonz:
Thanks. I keep forgetting that "Closes #NNNN" no longer works.

martinvonz:
Wait, this was not the FR but an older PR from me? I had completely forgotten I wrote it :P

Issue #6513, opened 2025-05-12
FR: By default, hide `.jj` in windows


Natural-selection1:
**Is your feature request related to a problem? Please describe.**
On macOS and Linux, directories starting with a dot (.) are hidden by default.
However, on Windows, we need to set additional attributes to achieve the same behavior.
To maintain consistency across platforms, we now hide the .jj directory on Windows as well.

**Describe the solution you'd like**
Just call Windows API

**Describe alternatives you've considered**
After creating the '.jj 'directory, manually hide it in a Command

1298345095:
Is there any working solution? I noticed your PR is still a draft.

Natural-selection1:
https://github.com/jj-vcs/jj/pull/6514#issuecomment-2875923163  

According to `yuja`, it might be better to have this task handled by gitoxide.

PR #6514, opened 2025-05-12
feat: hide .jj directory on Windows

Natural-selection1:
Closes: #6513
# Checklist

If applicable:

This is my first attempt to contribute to the jj project. I have read https://jj-vcs.github.io/jj/prerelease/contributing/, but it's still unclear, which caused me to encounter many issues (and I'm not sure how to resolve them) while trying to use jj to manage this contribution. If there's anything I've overlooked, any suggestions are welcome.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,7 +117,7 @@ version_check = "0.9.5"
 watchman_client = { version = "0.9.0" }
 whoami = "1.6.0"
 winreg = "0.52"
-
+winapi= {version = "0.3.9", features = ["winuser"] }
 # put all inter-workspace libraries, i.e. those that use 'path = ...' here in
 # their own (alphabetically sorted) block
 
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -16,7 +16,7 @@
 
 #![warn(missing_docs)]
 #![deny(unused_must_use)]
-#![forbid(unsafe_code)]
+#![deny(unsafe_code)]
 
 // Needed so that proc macros can be used inside jj_lib and by external crates
 // that depend on it.
--- a/lib/src/workspace.rs
+++ b/lib/src/workspace.rs
@@ -15,6 +15,8 @@
 #![allow(missing_docs)]
 
 use std::collections::HashMap;
+#[cfg(windows)]
+use std::ffi::CString;
 use std::fs;
 use std::fs::File;
 use std::io;
@@ -24,6 +26,10 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use thiserror::Error;
+#[cfg(windows)]
+use winapi::um::fileapi::SetFileAttributesA;
+#[cfg(windows)]
+use winapi::um::winnt::FILE_ATTRIBUTE_HIDDEN;
 
 use crate::backend::BackendInitError;
 use crate::backend::MergedTreeId;
@@ -119,7 +125,17 @@ pub struct Workspace {
 fn create_jj_dir(workspace_root: &Path) -> Result<PathBuf, WorkspaceInitError> {
     let jj_dir = workspace_root.join(".jj");
     match std::fs::create_dir(&jj_dir).context(&jj_dir) {
-        Ok(()) => Ok(jj_dir),
+        Ok(()) => {
+            #[cfg(windows)]
+            #[allow(unsafe_code)]
+            {
+                let jj_dir = CString::new(jj_dir.to_str().unwrap()).unwrap();
+                unsafe {
+                    SetFileAttributesA(jj_dir.as_ptr(), FILE_ATTRIBUTE_HIDDEN);
+                };
+            }
+            Ok(jj_dir)
+        }
         Err(ref e) if e.error.kind() == io::ErrorKind::AlreadyExists => {
             Err(WorkspaceInitError::DestinationExists(jj_dir))
         }
--- a/lib/tests/test_init.rs
+++ b/lib/tests/test_init.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[cfg(windows)]
+use std::ffi::CString;
 use std::path::Path;
 use std::path::PathBuf;
 
@@ -26,6 +28,10 @@ use testutils::git;
 use testutils::write_random_commit;
 use testutils::TestRepoBackend;
 use testutils::TestWorkspace;
+#[cfg(windows)]
+use winapi::um::fileapi::GetFileAttributesA;
+#[cfg(windows)]
+use winapi::um::winnt::FILE_ATTRIBUTE_HIDDEN;
 
 fn canonicalize(input: &Path) -> (PathBuf, PathBuf) {
     let uncanonical = input.join("..").join(input.file_name().unwrap());
@@ -49,6 +55,20 @@ fn test_init_local() {
     // Just test that we can write a commit to the store
     let mut tx = repo.start_transaction();
     write_random_commit(tx.repo_mut());
+}
+
+#[test]
+#[cfg(windows)]
+fn test_init_hidden() {
+    let settings = testutils::user_settings();
+    let temp_dir = testutils::new_temp_dir();
+    let (canonical, uncanonical) = canonicalize(temp_dir.path());
+    Workspace::init_internal_git(&settings, &uncanonical).unwrap();
+
+    let dot_jj_path = CString::new(canonical.join(".jj").to_str().unwrap()).unwrap();
+    let dot_jj_folder_attributes = unsafe { GetFileAttributesA(dot_jj_path.as_ptr()) };
+
+    assert_ne!(dot_jj_folder_attributes & FILE_ATTRIBUTE_HIDDEN, 0);
 }
 
 #[test]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2531,6 +2531,7 @@ dependencies = [
  "tracing",
  "version_check",
  "watchman_client",
+ "winapi",
  "winreg",
 ]
 
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -79,6 +79,7 @@ rustix = { workspace = true }
 
 [target.'cfg(windows)'.dependencies]
 winreg = { workspace = true }
+winapi = { workspace = true }
 
 [dev-dependencies]
 assert_matches = { workspace = true }


martinvonz:
Sounds reasonable, but do you know if Git, Mercurial, or other tools do this? Otherwise I'd be worried that this behavior will surprise users.

Natural-selection1:
> Sounds reasonable, but do you know if Git, Mercurial, or other tools do this? Otherwise I'd be worried that this behavior will surprise users.

I used to be a Windows git user(as for other tools, I have not used). For the '.git' directory in Windows, it is hidden, this is the reason why I discovered this problem

image

martinvonz:
Looks like it's controlled by [this Git config](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehideDotFiles). (I don't think we need a config for it.)

Cargo.toml
```diff
@@ -117,7 +117,7 @@ version_check = "0.9.5"
 watchman_client = { version = "0.9.0" }
 whoami = "1.6.0"
 winreg = "0.52"
-
+winapi= {version = "0.3.9", features = ["winuser"] }
```
martinvonz, line 120: Do we need the `winuser` feature? I don't even know what it's for; I'm just checking to make sure because I didn't see any mention of "user" in the code.

martinvonz, line 120: nit: keep the blank line (i don't care strongly about the blank line per se, but someone probably put it here on purpose)

lib/tests/test_init.rs
martinvonz, line 1: nit: please squash this into the parent commit

lib/src/workspace.rs
```diff
@@ -15,6 +15,8 @@
 #![allow(missing_docs)]
 
 use std::collections::HashMap;
+#[cfg(windows)]
+use std::ffi::CString;
```
martinvonz, line 19: nit: I think we usually move conditional imports inside the conditional block where they're needed. So move these imports to line 132 and drop the attribute on it.

Natural-selection1:
Git seems to have many 'quite peculiar' settings. For me, apart from:
* CRLF/LF
* username, email
* SSH
I don't think I've really encountered any others, haha.
Regarding the issue with the current PR, I think it would be best to maintain consistency with Git's default behavior.

lib/src/workspace.rs
```diff
@@ -119,7 +125,17 @@ pub struct Workspace {
 fn create_jj_dir(workspace_root: &Path) -> Result<PathBuf, WorkspaceInitError> {
     let jj_dir = workspace_root.join(".jj");
     match std::fs::create_dir(&jj_dir).context(&jj_dir) {
-        Ok(()) => Ok(jj_dir),
+        Ok(()) => {
+            #[cfg(windows)]
+            #[allow(unsafe_code)]
+            {
+                let jj_dir = CString::new(jj_dir.to_str().unwrap()).unwrap();
```
martinvonz, line 132: `jj_dir.to_str().unwrap()` will panic if `jj_dir` is not valid utf-8, which I think is not guaranteed here. I think `CString::new(jj_dir.as_os_str().as_encoded_bytes())` should be better.

(The second `.unwrap()` will panic if the string has internal NUL bytes, but I think that is a safe assumption.)

lib/src/workspace.rs
```diff
@@ -119,7 +125,17 @@ pub struct Workspace {
...
+                unsafe {
+                    SetFileAttributesA(jj_dir.as_ptr(), FILE_ATTRIBUTE_HIDDEN);
+                };
```
martinvonz, line 135: nit: Add a comment explaining why this is sound. I can only guess that it's something like this:

```rust
// Safety: `jj_dir` is a valid pointer to a null-terminated string.
```

lib/src/workspace.rs
```diff
@@ -119,7 +125,17 @@ pub struct Workspace {
...
     let jj_dir = workspace_root.join(".jj");
     match std::fs::create_dir(&jj_dir).context(&jj_dir) {
-        Ok(()) => Ok(jj_dir),
+        Ok(()) => {
+            #[cfg(windows)]
+            #[allow(unsafe_code)]
+            {
+                let jj_dir = CString::new(jj_dir.to_str().unwrap()).unwrap();
+                unsafe {
+                    SetFileAttributesA(jj_dir.as_ptr(), FILE_ATTRIBUTE_HIDDEN);
```
yuja, line 134: Perhaps, we should use the "W" API by using the underlying `OsString` data. I don't know if there's a safer/convenient wrapper, though. We'll also need to check the error code somehow.

yuja (commented):
Just for reference, Mercurial doesn't do, and there were some push back against the feature.
https://lists.mercurial-scm.org/pipermail/mercurial-devel/2014-March/205421.html

yuja:
Can you also check if the `.git` directory created by `jj git init`/`clone --colocate` is hidden on Windows? I couldn't find any relevant code in `gitoxide`. It would be odd if only `.jj` directory were hidden.

lib/src/lib.rs
PhilipMetzger, line 1: Nit: per contributing guidelines you can squash the third commit into this and probably should use `workspace: <title>` as the topic, since `feat:` just means feature which isn't something used by the project.

Natural-selection1:
> Can you also check if the `.git` directory created by `jj git init`/`clone --colocate` is hidden on Windows? I couldn't find any relevant code in `gitoxide`. It would be odd if only `.jj` directory were hidden.

Yes, your concern is valid - there is indeed an issue where the `.git` directory doesn't get hidden. Now I'm thinking of moving the timing of hiding the folder from when the folder is created to near the end of the command execution. 

This would be implemented around lines 108-109 in this code:
https://github.com/jj-vcs/jj/blob/ee9917e102a131ee9a6d27b2b30a4969d5b0e68f/cli/src/commands/git/init.rs#L100-L115

And also around line 147 in this code:
https://github.com/jj-vcs/jj/blob/9d1f799a8a554c15d3c986741f5bb0a718971bea/cli/src/commands/git/clone.rs#L126-L149

I'm not very familiar with jj yet - are there any other possible scenarios/cases I should be aware of?

yuja:
I think ".git" directory should be (ideally) handled by gitoxide, and would have to respect the Git configuration.

An easier (but adhoc) option might be to copy the hidden attribute from colocated `.git` directory by abusing `std::fs::set_permissions()`, which calls `SetFileAttributesW()` under the hood. It means `.jj` directory would be hidden only when `.git` exists and was hidden.

https://doc.rust-lang.org/stable/std/fs/fn.set_permissions.html

Natural-selection1:
Here's the translation of the GitHub PR discussion:

---

> I think the ".git" directory should (ideally) be handled by gitoxide and should respect the Git configuration.  
>  
> An easier (but ad-hoc) option might be to copy the hidden attribute from a colocated `.git` directory by abusing `std::fs::set_permissions()`, which calls `SetFileAttributesW()` under the hood. This means the `.jj` directory would only be hidden if `.git` exists and is hidden.  
>  
> [doc.rust-lang.org/stable/std/fs/fn.set_permissions.html](https://doc.rust-lang.org/stable/std/fs/fn.set_permissions.html)  

---

1.  
I don’t know much about gitoxide, but based on my search, it's a Rust implementation of Git. If possible, could you provide more details? :)  

2.  
https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehideDotFiles  
On Windows, Git always hides `.git`. I’m not sure how gitoxide handles the visibility of `.git`, but for Git users, `.git` is always hidden. Therefore, I think directly setting `.git` as hidden wouldn’t conflict with Git users' expectations.  

3.  
Based on the current discussion, I’ve extracted the logic for hiding `.jj` and `.git` into the following function:  
```rust  
#[cfg(windows)]  
pub fn hide_dotjj_and_dotgit(workspace_root: &Path) {  
    use std::ffi::CString;  
    use winapi::um::fileapi::SetFileAttributesA;  
    use winapi::um::winnt::FILE_ATTRIBUTE_HIDDEN;  
    let jj_dir = workspace_root.join(".jj");  
    let git_dir = workspace_root.join(".git");  
    let c_jj_dir = CString::new(jj_dir.as_os_str().as_encoded_bytes()).unwrap();  
    let c_git_dir = CString::new(git_dir.as_os_str().as_encoded_bytes()).unwrap();  
    #[allow(unsafe_code)]  
    unsafe {  
        if jj_dir.exists() {  
            let hide_jj_dir = SetFileAttributesA(c_jj_dir.as_ptr(), FILE_ATTRIBUTE_HIDDEN);  
            if hide_jj_dir == 0 {  
                println!("Failed to hide .jj directory");  
            }  
        }  
        if git_dir.exists() {  
            let hide_git_dir = SetFileAttributesA(c_git_dir.as_ptr(), FILE_ATTRIBUTE_HIDDEN);  
            if hide_git_dir == 0 {  
                println!("Failed to hide .git directory");  
            }  
        }  
    }  
}  
```  

It is applied at **line 121** in the following function to cover `jj init`:  
https://github.com/jj-vcs/jj/blob/cb877a94c16a21e20fcf81934142b63950e9d42e/lib/src/workspace.rs#L118-L127  

And at **lines 244-245** in the following function to cover `jj git init --colocate` and `jj git clone --colocate`:  
https://github.com/jj-vcs/jj/blob/cb877a94c16a21e20fcf81934142b63950e9d42e/lib/src/git_backend.rs#L222-L245  

Any further thoughts?

yuja:
> I don’t know much about gitoxide, but based on my search, it's a Rust implementation of Git. If possible, could you provide more details? :)

https://github.com/search?q=repo%3AGitoxideLabs%2Fgitoxide%20hidedotfiles&type=code

Appears that they have a plan.

> https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehideDotFiles On Windows, Git always hides `.git`. I’m not sure how gitoxide handles the visibility of `.git`, but for Git users, `.git` is always hidden. Therefore, I think directly setting `.git` as hidden wouldn’t conflict with Git users' expectations.

It's unclear from the doc, but Git supports `hideDotFiles=false|true|dotGitOnly`. If `false`, `.git` wouldn't be hidden. (I don't have Windows machine, so the actual behavior isn't tested.)

martinvonz:
@Natural-selection1: Could you sign the Google CLA? We won't be able to accept your PR without it.

Natural-selection1:
> @Natural-selection1: Could you sign the Google CLA? We won't be able to accept your PR without it.  

I signed it before, but forgot to rescan as required. Sorry about that. : )

Natural-selection1:
> I think ".git" directory should be (ideally) handled by gitoxide, and would have to respect the Git configuration.


Respecting Git settings seems unlikely, as many tools bundle their own Git implementations. Here's a demonstration of cargo's built-in Git behavior:  
image

---  
However, after configuration, `cargo new` doesn't appear to comply:  

image

---  

Perhaps only respecting pre-existing `.git` directories would be practical. The basic logic would be:  
1. Check if the time difference between the `.git` directory's creation and the current moment is within 1 second  
2. If no, then `.jj` should match its state  
3. If yes (meaning we just created it), then both should be hidden  

I wonder if this approach would be convincing?

yuja:
> Respecting Git settings seems unlikely, as many tools bundle their own Git.

Well, I assume `gitoxide` will respect Git settings when they implement the feature. And it's nice if we don't have to "fix" the `.git` directory attribute created by `gitoxide` by ourselves.

I don't know whether the `.jj` directory should be made hidden/visible consistently with the colocated `.git` directory.

1. The easiest option is to do nothing about `.jj` directory (= the current behavior.)
2. Another easy (but a bit weird depending on user's expectation) option I think is to just copy file attribute from the colocated `.git` directory (which would require no unsafe code, but I haven't tested.)
3. Lastly, we can make `.jj` directory hidden unconditionally (= your implementation, I think.)

> a demonstration of cargo's built-in Git behavior:

Maybe `cargo` doesn't because they still use `git2`/`libgit2`?

Natural-selection1:
> 1. The easiest option is to do nothing about the `.jj` directory (= the current behavior).

Remember the title of our PR? It's to ensure `.xx` remains hidden in Windows just like it does in Mac or Linux.

---
> 2. Another easy (but a bit weird depending on user's expectation) option I think is to just copy file attribute from the colocated `.git` directory (which would require no unsafe code, but I haven't tested.)
> 3. Lastly, we can make `.jj` directory hidden unconditionally (= your implementation, I think.)

Perhaps you could take a look at my current implementation (please ignore the rough eprintln for now). You can start reading from around `match dotgit_path.exists()`:

```rust
#[cfg(windows)]
pub fn set_dotgit_and_dotjj_visibility(workspace_root: &Path) {
    use std::ffi::CString;
    use std::time::Duration;
    use std::time::SystemTime;
    use winapi::um::fileapi::GetFileAttributesA;
    use winapi::um::fileapi::SetFileAttributesA;
    use winapi::um::winnt::FILE_ATTRIBUTE_HIDDEN;

    let dotjj_path = workspace_root.join(".jj");
    let dotgit_path = workspace_root.join(".git");
    let c_dotgit_path = CString::new(dotgit_path.as_os_str().as_encoded_bytes()).unwrap();
    fn hide_folder(path: &Path) {
        let dir_name = Path::new(path).file_name().unwrap().to_str().unwrap();
        let c_path = CString::new(path.as_os_str().as_encoded_bytes()).unwrap();
        #[allow(unsafe_code)]
        unsafe {
            if SetFileAttributesA(c_path.as_ptr(), FILE_ATTRIBUTE_HIDDEN) == 0 {
                eprintln!("Failed to hide {dir_name}");
            }
        }
    }
    fn get_folder_age(path: &Path) -> Result<Duration, io::Error> {
        SystemTime::now()
            .duration_since(fs::metadata(path)?.created()?)
            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))
    }

    match dotgit_path.exists() {
        // cover `jj git init`
        // because there is no `.git` created by us just now or exists before,
        // so we hide `.jj`, just like dot_folder on Mac or Linux
        false => {
            hide_folder(&dotjj_path);
        }
        // cover `jj git init --colocate` and `jj git clone [Source] --colocate`
        true => {
            // check if it's created by us just now or is old
            let is_dotgit_exist_before = match get_folder_age(&dotgit_path) {
                Ok(duration) => duration > Duration::from_secs(1),
                Err(_) => true,
            };

            if is_dotgit_exist_before {
                // if `.git` exists before, keep `.jj` the same visibility as `.git`
                #[allow(unsafe_code)]
                unsafe {
                    if GetFileAttributesA(c_dotgit_path.as_ptr()) & FILE_ATTRIBUTE_HIDDEN != 0 {
                        hide_folder(&dotjj_path);
                    }
                }
            } else {
                // if `.git` is newly created, hide `.jj` and `.git` just like dot_folder on Mac or Linux
                hide_folder(&dotjj_path);
                hide_folder(&dotgit_path);
            }
        }
    }
}
```

I think the above approach might be the best solution without adding user configuration options.

---
> 2. Another easy (but a bit weird depending on user's expectation) option I think is to just copy file attribute from the colocated `.git` directory (which would require no unsafe code, but I haven't tested.)

I don't think we need to be overly cautious about unsafe here - we're just calling FFI, not playing with raw pointers.

---
> Maybe `cargo` doesn't because they still use `git2`/`libgit2`?

I haven't found relevant doc, but judging from `cargo help new`, it seems they don't even provide visibility options.

yuja:
> > 1. The easiest option is to do nothing about the `.jj` directory (= the current behavior).
> 
> Remember the title of our PR? It's to ensure `.xx` remains hidden in Windows just like it does in Mac or Linux.

I know, but `git` doesn't apply this rule to e.g. `.gitignore` despite any dot files are "hidden" on Unix, so there may be conflicting requirements.

(Just for reference, Mercurial doesn't make `.hg` directory hidden. Git does for some reasons, though.)

> > 2. Another easy (but a bit weird depending on user's expectation) option I think is to just copy file attribute from the colocated `.git` directory (which would require no unsafe code, but I haven't tested.)
> > 3. Lastly, we can make `.jj` directory hidden unconditionally (= your implementation, I think.)
> 
> Perhaps you could take a look at my current implementation (please ignore the rough eprintln for now). You can start reading from around `match dotgit_path.exists()`:

I believe we shouldn't do timestamp-based heuristics. (I suggested (2) just because it's easier to implement.)

To make both `.git` and `.jj` directories hidden when the `.git` directory was also created by `jj`, insert a relevant code to `GitBackend::init_colocated()` and its caller. (To be clear, I don't think we should do this for `.git` directory, and instead send PR/FR to `gitoxide`.)

BTW, you should use `SetFileAttributeW` instead of `A`. A Rust `String` is UTF-8, and `OsString` is platform-dependent bytes (iirc, it's less-strict UTF-8 on Windows), whereas the `A` (or ANSI) Windows API expects a string of legacy encoding (such as ISO-8859-1) depending on the configured locale. I don't know if there are better/safer Windows API wrapper.

yuja:
> you should use `SetFileAttributeW` instead of `A`.

Appears that Microsoft's `windows-rs` implements automatic conversion to the native 16-bit string type. The API is still `unsafe`, though.

https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Storage/FileSystem/fn.SetFileAttributesW.html

@thoughtpolice might have an opinion about the dependency.

Natural-selection1:
I've pushed a new version. Please ignore the suboptimal git commit messages for now (I'm currently having some difficulties using jj to manage the GitHub repository). If we agree this PR is mergeable, I'll reorganize the commits properly in Git  : )

Natural-selection1:
> Appears that Microsoft's `windows-rs` implements automatic conversion to the native 16-bit string type. The API is still `unsafe`, though.

Oh, I noticed https://crates.io/crates/windows when I was looking up information about SetFileAttributeW. `winapi` hasn't been updated in six years! I'll try to update the dependency.

PR #6516, opened 2025-05-12
workspace: remove `load_working_copy()` from `WorkspaceLoader`

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/workspace.rs
+++ b/lib/src/workspace.rs
@@ -53,7 +53,6 @@ use crate::settings::UserSettings;
 use crate::signing::SignInitError;
 use crate::signing::Signer;
 use crate::simple_backend::SimpleBackend;
-use crate::store::Store;
 use crate::transaction::TransactionCommitError;
 use crate::working_copy::CheckoutError;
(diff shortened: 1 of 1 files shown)


lib/src/workspace.rs
```diff
@@ -580,6 +573,18 @@ impl DefaultWorkspaceLoader {
...
+    fn load_working_copy(
+        &self,
+        store: &Arc<Store>,
+        working_copy_factory: &dyn WorkingCopyFactory,
+    ) -> Result<Box<dyn WorkingCopy>, WorkspaceLoadError> {
+        Ok(working_copy_factory.load_working_copy(
+            store.clone(),
+            self.workspace_root.clone(),
+            self.working_copy_state_path.clone(),
+        )?)
```
yuja, line 586: nit: or just inline it because this is merely a forwarding function?

yuja (approved):

lib/src/workspace.rs
martinvonz: True, that's even better. Done.

PR #6517, opened 2025-05-12
signing: add support for bpb backend

dragonmaus:
[Boat's Personal Barricade](https://github.com/withoutboats/bpb) (bpb) is an alternative commit signing mechanism meant to act as a drop-in replacement for gpg. The way that Jujutsu handles signing is slightly different than Git, so some changes were necessary to allow bpb to work properly.

# Checklist

If applicable:


dragonmaus:
Not quite sure how to add the `bpb` command to the CI environments. The tests will fail until that is accomplished.

yuja:
> [Boat's Personal Barricade](https://github.com/withoutboats/bpb) (bpb) is an alternative commit signing mechanism meant to act as a drop-in replacement for gpg. The way that Jujutsu handles signing is slightly different than Git, so some changes were necessary to allow bpb to work properly.

Can we somehow adjust our `GpgBackend` to be compatible with bpb? or improve bpb's compatibility? It's unfortunate that we have to implement new backend for tools supposed to be a drop-in replacement.

emilazy:
I don’t think it’s worth supporting `bpb` when it hasn’t been updated in seven years and doesn’t build with Rust ≥ 1.76.

If we want a GnuPG‐less OpenPGP signing backend, it would make more sense to integrate the pure‐Rust [Sequoia](https://sequoia-pgp.org) as an alternative to shelling out to GnuPG (or a replacement? it can speak the agent protocol, so it would not mean sacrificing GnuPG compatibility).

dragonmaus:
> Can we somehow adjust our `GpgBackend` to be compatible with bpb? or improve bpb's compatibility?

I poked around at that path for a bit, but my impression is that doing so would either slightly regress the GPG support or needlessly complicate BPB's code and interface.

----

> I don’t think it’s worth supporting `bpb` when it hasn’t been updated in seven years and doesn’t build with Rust ≥ 1.76.

The lack of official maintenance is unfortunate. There are several (also old, but not as old) forks that modernize it and/or tweak it slightly, but other than that I view it as a more-or-less "complete" piece of software. That is, I have it installed on my machines, where it has worked perfectly for many years now. My guess is that the original developer is the same.

> If we want a GnuPG‐less OpenPGP signing backend, it would make more sense to integrate the pure‐Rust [Sequoia](https://sequoia-pgp.org) as an alternative to shelling out to GnuPG (or a replacement? it can speak the agent protocol, so it would not mean sacrificing GnuPG compatibility).

I'll look into this. I'm by no means any kind of crypto expert, but I really hate having the mess that is PGP/GPG on my system. This might be a viable path forward, both for `jj` and `bpb`.

PR #6524, opened 2025-05-13
Add a `fswatch`-based `jj log` mechanism to FAQ

chendo:
I was asked to contribute my `fswatch`-based to the FAQ as an alternative to using `watch`, which will run `jj` even though nothing has changed.

It does require `fswatch` to be installed, which is not standard on most machines, but will not run `jj log` unless necessary.


chendo:
Do documentation changes require a CLA? I am happy for my changes to be considered public domain.

martinvonz:
> Do documentation changes require a CLA?

Yes, all contributions to this repo do.

PR #6528, opened 2025-05-13
cli git init: show `jj bookmark track` command on new line

cenviity:
I often copy and run this command after running `jj git init --colocate`. Having it on a separate line makes it stand out more and easier to copy, especially if there are many bookmarks listed.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/init.rs
+++ b/cli/src/commands/git/init.rs
@@ -19,6 +19,7 @@ use std::path::PathBuf;
 use std::str;
 use std::sync::Arc;
 
+use indoc::writedoc;
 use itertools::Itertools as _;
 use jj_lib::file_util;
 use jj_lib::git;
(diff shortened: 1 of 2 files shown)


yuja (commented):
The output looks good to me.

cli/src/commands/git/init.rs
```diff
@@ -270,20 +270,20 @@ fn print_trackable_remote_bookmarks(ui: &Ui, view: &View) -> io::Result<()> {
         return Ok(());
     }
 
+    writeln!(
+        ui.hint_default(),
+        "The following remote bookmarks aren't associated with the existing local bookmarks:"
+    )?;
```
martinvonz, line 277: Since this move outside the `if` block, I think it means that it will no longer respect `--quiet`. Is that intentional?

yuja: `ui.hint_default()` checks `--quit` internally, so this should technically work. That said, I like the original form because it's more obvious that the header and its body are printed under the same condition.

cenviity: Makes sense. I've dropped the first commit so we're back to the original form.

martinvonz (approved):

cenviity:
Thanks!

PR #6534, opened 2025-05-14
cli: git clone: add config to not track default remote bookmark

yuja:
# Checklist

If applicable:


CHANGELOG.md
```diff
@@ -40,6 +40,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   anywhere in the revision tree with the `--insert-before`, `--insert-after` and
   `--destination` command line flags.
 
+* Added `git.track-default-bookmark-on-clone` setting to not create tracking
+  local bookmark on `jj git clone`.
```
cenviity, line 44: It sounds like setting this to `true` would _not_ create the tracking local bookmark when it's actually the opposite. Perhaps something along the lines of '[…] setting to configure whether to create […]'?

yuja: Good point, updated.

martinvonz (approved):
Makes sense. The only reason I want the local bookmark for the default branch to exist and be tracked is so I can do `jj new main` and `jj rebase -d main`. Do you instead use `'trunk()'` or maybe a revset alias for it?

PR #6539, opened 2025-05-14
templater: add arithmetic operators

### LINKED ISSUE #5982: FR: Integer arithmetic

I might be missing something, but is there no way to do integer arithmetic in the templating language? It seems that there is an integer negation operator (`-x`), but I couldn't find a way to add two integers together.

So, for example, `jj log -r@ -T'-("hello".len())'` is `-5`, but `jj log -r@ -T'("hello".len()) + 1'` crashes with
```
Error: Failed to parse template: Syntax error
Caused by:  --> 1:17
  |
1 | ("hello".len()) + 1
  |                 ^---
  |
  = expected <EOI>, `++`, `||`, `&&`, `==`, `!=`, `>=`, `>`, `<=`, or `<`
```

Implement Integer arithmetic.

You might be able to do simple arithmetic by creating/manipulating strings of target length, but that sounds like a horribly hacky way to do it.

tjjfvi:
Adds arithmetic operators on integers. Fixes #5982.

I ran into this when trying to do some alignment-type things inside a template. I'm currently using
```
'len(template)' = 'stringify(template).len()'
'repeat(str, int)' = 'stringify(pad_end(int, "", str))'
'spaces(int)' = 'repeat(" ", int)'

'add(a, b)' = '''
len(concat(spaces(a), spaces(b)))
'''

'sub(a, b)' = '''
len(spaces(a).substr(0, -b))
'''
```

~~I have not yet added tests, as I am unsure of the best place to put them. `test_templater.rs` seems the most applicable, although it seems currently to mostly test higher-level functionality and not specific templating operations.~~

# Checklist

If applicable:

### DIFF
--- a/cli/src/template_builder.rs
+++ b/cli/src/template_builder.rs
@@ -813,6 +813,36 @@ fn build_binary_operation<'a, L: TemplateLanguage<'a> + ?Sized>(
             };
             Ok(L::Property::wrap_property(out))
         }
+        BinaryOp::Add | BinaryOp::Sub | BinaryOp::Mul | BinaryOp::Div | BinaryOp::Rem => {
+            let lhs = expect_integer_expression(language, diagnostics, build_ctx, lhs_node)?;
+            let rhs = expect_integer_expression(language, diagnostics, build_ctx, rhs_node)?;
+            let build = |op: fn(i64, i64) -> Option<i64>, msg: fn(i64) -> &'static str| {
+                (lhs, rhs).and_then(move |(l, r)| {
+                    op(l, r).ok_or_else(|| TemplatePropertyError(msg(r).into()))
+                })
+            };
+            let out = match op {
+                BinaryOp::Add => build(i64::checked_add, |_| "Attempt to add with overflow"),
+                BinaryOp::Sub => build(i64::checked_sub, |_| "Attempt to subtract with overflow"),
+                BinaryOp::Mul => build(i64::checked_mul, |_| "Attempt to multiply with overflow"),
+                BinaryOp::Div => build(i64::checked_div, |r| {
+                    if r == 0 {
+                        "Attempt to divide by zero"
+                    } else {
+                        "Attempt to divide with overflow"
+                    }
+                }),
+                BinaryOp::Rem => build(i64::checked_rem, |r| {
+                    if r == 0 {
+                        "Attempt to divide by zero"
+                    } else {
+                        "Attempt to divide with overflow"
+                    }
+                }),
+                _ => unreachable!(),
+            };
+            Ok(out.into_dyn_wrapped())
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


yuja:
I think the tests should be added to `template_builder.rs`. There's `test_arithmetic_operation()`

tjjfvi:
Ahh yes thank you. I don't often write tests inline so I was just looking in the `tests` dir haha.

cli/src/template_builder.rs
```diff
@@ -813,6 +813,40 @@ fn build_binary_operation<'a, L: TemplateLanguage<'a> + ?Sized>(
             };
             Ok(L::Property::wrap_property(out))
         }
+        BinaryOp::Add | BinaryOp::Sub | BinaryOp::Mul | BinaryOp::Div | BinaryOp::Rem => {
+            let lhs = expect_integer_expression(language, diagnostics, build_ctx, lhs_node)?;
+            let rhs = expect_integer_expression(language, diagnostics, build_ctx, rhs_node)?;
+            let out = lhs.and_then(move |l| {
```
yuja, line 819: `(lhs, rhs).and_then(|(l, r)|` (no short-circuiting is needed here)

Maybe we can also move `match op` out of the property builder chain. It's a little bit nicer that bad `unreachable!()` condition would be detected at the parsing phase.

yuja (commented):
Looks generally good, thanks.

cli/src/template_builder.rs
tjjfvi: I applied both changes, although pulling the `match op` out does make it significantly more verbose.

cli/src/template_builder.rs
```diff
@@ -813,6 +813,40 @@ fn build_binary_operation<'a, L: TemplateLanguage<'a> + ?Sized>(
...
+                    BinaryOp::Add => l.checked_add(r).ok_or_else(|| {
+                        TemplatePropertyError("Attempt to add with overflow".into())
+                    }),
+                    BinaryOp::Sub => l.checked_sub(r).ok_or_else(|| {
+                        TemplatePropertyError("Attempt to subtract with overflow".into())
+                    }),
+                    BinaryOp::Mul => l.checked_mul(r).ok_or_else(|| {
+                        TemplatePropertyError("Attempt to multiply with overflow".into())
+                    }),
+                    BinaryOp::Div => l.checked_div(r).ok_or_else(|| {
```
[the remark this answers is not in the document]
tjjfvi: Ah, that's true. I suspect most usage would just be dividing by positive integers; it could error on negative divisors for now so as to not decide now.

cli/src/template_builder.rs
yuja: Maybe you can add helper function of `(lhs, rhs, f) -> BoxedTemplateProperty<'a, i64>` type? Then, each operation will be like `apply(lhs, rhs, i64::checked_add, "Attempt to add with overflow")`.

cli/src/template_builder.rs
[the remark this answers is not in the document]
yuja: > I suspect most usage would just be dividing by positive integers

Yeah, I think we can decide/change that later.

yuja (approved):

PR #6546, opened 2025-05-15
git: export git ref to parent branch when possible

jakobhellermann:
This is helpful for
- editors with git integration that show a warning when on an detached head (intellij etc.)
- using git and jj together on a repository

It is only done if there is a single parent with a single bookmark, otherwise it wouldn't be obvious which one to pick.

# Checklist

### DIFF
--- a/lib/src/commit.rs
+++ b/lib/src/commit.rs
@@ -93,6 +93,14 @@ impl Commit {
         &self.data.parents
     }
 
+    pub fn single_parent(&self) -> Option<&CommitId> {
+        if self.data.parents.len() == 1 {
+            Some(&self.data.parents[0])
+        } else {
+            None
+        }
+    }
+
     pub fn parents(&self) -> impl Iterator<Item = BackendResult<Commit>> + use<'_> {
         self.data.parents.iter().map(|id| self.store.get_commit(id))
     }
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -950,7 +950,7 @@ pub fn export_some_refs(
                 update_git_head(
                     &git_repo,
                     gix::refs::transaction::PreviousValue::MustExistAndMatch(old_target),
-                    current_oid,
+                    current_oid.map(gix::refs::Target::Object),
                 )
                 .map_err(GitExportError::from_git)?;
             }
@@ -1191,11 +1191,11 @@ fn update_git_ref(
 fn update_git_head(
     git_repo: &gix::Repository,
     expected_ref: gix::refs::transaction::PreviousValue,
-    new_oid: Option<gix::ObjectId>,
+    new_target: Option<gix::refs::Target>,
 ) -> Result<(), gix::reference::edit::Error> {
     let mut ref_edits = Vec::new();
-    let new_target = if let Some(oid) = new_oid {
-        gix::refs::Target::Object(oid)
+    let new_target = if let Some(target) = new_target {
+        target
     } else {
         // Can't detach HEAD without a commit. Use placeholder ref to nullify
         // the HEAD. The placeholder ref isn't a normal branch ref. Git CLI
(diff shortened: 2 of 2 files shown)


lib/src/git.rs
```diff
@@ -1276,9 +1276,38 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
...
-            .map(|id| gix::ObjectId::from_bytes_or_panic(id.as_bytes()));
+
+        let new_oid = new_head_target.as_normal().map(|id| {
+            let single_parent_bookmark = wc_commit
+                .single_parent()
+                .and_then(|_| {
+                    mut_repo
+                        .view()
+                        .bookmarks()
+                        .filter(|(_, bookmark)| bookmark.local_target == &new_head_target)
```
jakobhellermann, line 1287: This is a scan over all bookmarks. The commit template uses an `CommitRefsIndex` to accelerate this, but as far as I can tell there's nothing to improve this available here.

lib/src/git.rs
```diff
@@ -1259,7 +1259,7 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
 
     // If the first parent of the working copy has changed, reset the Git HEAD.
     let old_head_target = mut_repo.git_head();
-    if old_head_target != new_head_target {
+    if old_head_target != new_head_target || true {
```
seh, line 1262: Did you intend to leave the `|| true` there? If so, what is accomplishing?

jakobhellermann: Oh, that was just for testing. I'll remove it

lib/src/git.rs
yuja: Maybe you can use `local_bookmarks_for_commit()`? It's not optimized in any sense, though.

lib/src/git.rs
```diff
@@ -1276,9 +1276,38 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(),
...
+                })
+                .and_then(|(name, _)| {
+                    to_git_ref_name(
+                        GitRefKind::Bookmark,
+                        name.to_remote_symbol(REMOTE_NAME_FOR_LOCAL_GIT_REPO),
+                    )
+                    .and_then(|name| {
+                        gix::refs::FullName::try_from(BString::from(name.into_string())).ok()
+                    })
+                    .map(gix::refs::Target::Symbolic)
```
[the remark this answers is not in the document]
yuja: Perhaps, we can check if `<name>` bookmark is in sync with `<name>@git`. If they point to the same commit, the bookmark would exist in Git (ignoring TOCTOU issue.) A stricter option might be to issue `RefEdit` to update the `HEAD` ref, plus noop `RefEdit` to update the target bookmark (to the same value.0

yuja (commented):
We'll need a config knob to opt in to this behavior. It would be surprising if HEAD becomes undetached by checking out an arbitrary commit.

lib/src/commit.rs
```diff
@@ -93,6 +93,14 @@ impl Commit {
         &self.data.parents
     }
 
+    pub fn single_parent(&self) -> Option<&CommitId> {
+        if self.data.parents.len() == 1 {
+            Some(&self.data.parents[0])
+        } else {
+            None
+        }
```
cenviity, line 101: nit: `.as_slice()` then pattern match?

martinvonz:
Another option is to always create/update a jj-specific branch. We could call it something like `jj-<workspace id>` (since there can be multiple workspaces).

AngelEzquerra:
> Another option is to always create/update a jj-specific branch. We could call it something like `jj-<workspace id>` (since there can be multiple workspaces).

I think this wojks be nice. It has the advantage that it can work even if you don’t use bookmarks.

AngelEzquerra:
> Another option is to always create/update a jj-specific branch. We could call it something like `jj-<workspace id>` (since there can be multiple workspaces).

FWIW I’ve been trying this approach manually ha today (by doing `git branch --force jj-default` followed by `git switch jj-default` after every `jj new` operation) and it works really well.

PR #6547, opened 2025-05-15
docs: Don't include unnecessary shebangs

icorbrey:
It isn't necessary to include shebangs in scripts passed into `util exec` since you're already explicitly calling out the shell you'd like to use. This change removes shebangs from examples for `util exec` and adds a note below calling this out.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/util/exec.rs
+++ b/cli/src/commands/util/exec.rs
@@ -50,7 +50,6 @@ use crate::ui::Ui;
 /// ```toml
 /// [aliases]
 /// my-inline-script = ["util", "exec", "--", "bash", "-c", """
-/// #!/usr/bin/env bash
 /// set -euo pipefail
 /// echo "Look Ma, everything in one file!"
 /// echo "args: $@"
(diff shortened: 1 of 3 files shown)


icorbrey:
Whoops, rustfmt

cli/src/commands/util/exec.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 The Jujutsu Authors
+// Copyright 2024-2025 The Jujutsu Authors
```
PhilipMetzger, line 1: nit: copyright years don't need to be updated

PhilipMetzger (approved):
LG, with copyright nit

cli/src/commands/util/exec.rs
```diff
@@ -59,6 +58,9 @@ use crate::ui::Ui;
 /// # This last empty string will become "$0" in bash, so your actual arguments
 /// # are all included in "$@" and start at "$1" as expected.
 /// ```
+///
+/// > Note: Shebangs (e.g. `#!/usr/bin/env`) aren't necessary since you're
+/// > already explicitly passing your script into the right shell.
```
martinvonz, line 63: I would not include this. I think most users are just going to copy and paste and not wonder about it. If they do wonder about it, then they're probably able to figure it out themselves. But feel free to keep it if you think it'll help

martinvonz (approved):

cli/src/commands/util/exec.rs
icorbrey: I think it's not harmful to include and might head off some questions before they hit issues or the Discord, so I'd like to keep it in

PR #6552, opened 2025-05-16
cli: accept --tool=:<format> argument as builtin diff format

yuja:
# Checklist

If applicable:


martinvonz (approved):
> I don't think this is practically useful, but the
:<format> syntax should work in both config files and command arguments.

I think it can be useful by letting us add support for diff formats to commands without taking so many flag names. Especially the short flag `-s` is unfortunate to occupy for some commands.

It's also nice that we can give the diff format flag a clearer name in contexts where e.g. just `--summary` is unclear. For example, it might be clearer with `jj op show --diff-tool :summary`.

PR #6553, opened 2025-05-16
test: use `--at-op` and templates instead of manually parsing op ids from op log output

cenviity:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_concurrent_operations.rs
+++ b/cli/tests/test_concurrent_operations.rs
@@ -213,7 +213,7 @@ fn test_concurrent_snapshot_wc_reloadable() {
     let previous_op_id = op_log_lines[6].split_once("  ").unwrap().1;
 
     // Another process started from the "initial" operation, but snapshots after
-    // the "child1" checkout has been completed.
+    // the "child1" checkout have been completed.
     std::fs::rename(
         op_heads_dir.join(current_op_id),
         op_heads_dir.join(previous_op_id),
--- a/cli/tests/test_concurrent_operations.rs
+++ b/cli/tests/test_concurrent_operations.rs
@@ -96,9 +96,9 @@ fn test_concurrent_operations_auto_rebase() {
     // We should be informed about the concurrent modification
     let output = get_log_output(&work_dir);
     insta::assert_snapshot!(output, @r"
-    ○  b748b7189c0b301be205ebaf74b941c4ec209e69 new child
-    @  084d39fe0ab1cd01a6b7fd21fdea3a8dcc9a48fc rewritten
-    ◆  0000000000000000000000000000000000000000
+    ○  new child
+    @  rewritten
+    ◆
     [EOF]
     ------- stderr -------
     Concurrent modification detected, resolving automatically.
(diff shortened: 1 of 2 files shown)


cli/tests/test_concurrent_operations.rs
```diff
@@ -187,33 +167,36 @@ fn test_concurrent_snapshot_wc_reloadable() {
...
     [EOF]
     ");
+    let template = r#"id ++ "\n""#;
+    let output = work_dir.run_jj(["op", "log", "--no-graph", "-T", template]);
     let op_log_lines = output.stdout.raw().lines().collect_vec();
-    let current_op_id = op_log_lines[0].split_once("  ").unwrap().1;
-    let previous_op_id = op_log_lines[6].split_once("  ").unwrap().1;
+    let current_op_id = op_log_lines[0];
+    // Get the op id before the latest snapshot.
+    let previous_op_id = op_log_lines[2];
```
yuja, line 196: nit: can you insert `insta::assert_snapshot!(current_op_id)`, etc. for sanity checking?

yuja (approved):

cli/tests/test_concurrent_operations.rs
cenviity: Done.

cli/tests/test_concurrent_operations.rs
```diff
@@ -187,27 +187,29 @@ fn test_concurrent_snapshot_wc_reloadable() {
...
-    ○  2affa702525487ca490c4bc8a9a365adf75f972efb5888dd58716de7603e822ba1ed1ed0a50132ee44572bb9d819f37589d0ceb790b397ddcc88c976fde2bf02
+    ○  2affa7025254
     │  add workspace 'default'
-    ○  00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
+    ○  000000000000
 
     [EOF]
     ");
+    let template = r#"id ++ "\n" ++ description ++ "\n" ++ tags"#;
+    let output = work_dir.run_jj(["op", "log", "-T", template]);
```
yuja: nit: I wouldn't include this here because it makes unclear what `op_log_lines[i]` would return, and snapshotting the new `output` would go against the intent of this patch.

cli/tests/test_concurrent_operations.rs
```diff
@@ -209,16 +209,14 @@ fn test_concurrent_snapshot_wc_reloadable() {
...
     // Another process started from the "initial" operation, but snapshots after
     // the "child1" checkout have been completed.
-    std::fs::rename(
-        op_heads_dir.join(current_op_id),
-        op_heads_dir.join(previous_op_id),
-    )
-    .unwrap();
+    let move_op_head =
+        |from: &str, to: &str| std::fs::rename(op_heads_dir.join(from), op_heads_dir.join(to));
+    move_op_head(op_id_after_snapshot, op_id_before_snapshot).unwrap();
```
yuja, line 202: nit: I personally don't think this helps understand the code. It's obvious that we're moving/renaming the op heads, and the "why" would have to be explained in the comment anyway.

yuja (approved):

cli/tests/test_concurrent_operations.rs
cenviity: Upon re-reading I agree that the closure doesn't help explain anything, so I've inlined it back now. Reworded the comment instead to hopefully make it clearer what is being simulated.

PR #6570, opened 2025-05-19
lib: return correct op set subexpression when reporting multiple operations found

cenviity:
# Checklist

If applicable:

### DIFF
--- a/lib/src/op_walk.rs
+++ b/lib/src/op_walk.rs
@@ -148,17 +148,27 @@ fn resolve_single_op(
         "@" => get_current_op(),
         s => resolve_single_op_from_store(op_store, s),
     }?;
-    for c in op_postfix.chars() {
+    for (i, c) in op_postfix.chars().enumerate() {
         let mut neighbor_ops = match c {
             '-' => operation.parents().try_collect()?,
             '+' => find_child_ops(head_ops.as_ref().unwrap(), operation.id())?,
             _ => unreachable!(),
         };
         operation = match neighbor_ops.len() {
+            // Since there is no hint provided for `EmptyOperations` in
+            // `opset_resolution_error_hint()` (there would be no useful hint for the
+            // user to take action on anyway), we don't have to worry about op ids being
+            // incoherent with the op set expression shown to the user, unlike for the
+            // `MultipleOperations` variant.
+            //
+            // The full op set expression is guaranteed to be empty in this case,
+            // because ancestors/descendants of an empty operation are empty.
             0 => Err(OpsetResolutionError::EmptyOperations(op_str.to_owned()))?,
             1 => neighbor_ops.pop().unwrap(),
+            // Returns the exact subexpression that resolves to multiple operations,
+            // rather than the full expression provided by the user.
             _ => Err(OpsetResolutionError::MultipleOperations {
-                expr: op_str.to_owned(),
+                expr: op_str[..=op_symbol.len() + i].to_owned(),
                 candidates: neighbor_ops.iter().map(|op| op.id().clone()).collect(),
             })?,
         };
--- a/lib/tests/test_operations.rs
+++ b/lib/tests/test_operations.rs
@@ -592,12 +592,29 @@ fn test_resolve_op_parents_children() {
     let parent_op_ids = repo.operation().parent_ids();
 
     let op5_id_hex = repo.operation().id().hex();
-    let op_str = format!("{op5_id_hex}-");
-    let error = op_walk::resolve_op_with_repo(&repo, &op_str).unwrap_err();
+    let parents_op_str = format!("{op5_id_hex}-");
+    let error = op_walk::resolve_op_with_repo(&repo, &parents_op_str).unwrap_err();
     assert_eq!(
         extract_multiple_operations_error(&error).unwrap(),
-        (&op_str, parent_op_ids)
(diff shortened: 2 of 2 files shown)


lib/src/op_walk.rs
```diff
@@ -58,7 +58,7 @@ pub enum OpsetEvaluationError {
 
 /// Error that may occur during parsing and resolution of operation set
 /// expression.
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq, Eq)]
```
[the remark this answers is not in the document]
cenviity: Dropped the additional `derive`s and updated to use `assert_matches!` now.

lib/tests/test_operations.rs
```diff
@@ -588,20 +588,36 @@ fn test_resolve_op_parents_children() {
...
+            assert_matches!(
+                error,
+                OpsetResolutionError::MultipleOperations {
+                    expr,
+                    candidates,
+                } if expr == *op_str && candidates == parent_op_ids
+            );
+        }
+        _ => unreachable!("should resolve to divergent operations"),
+    }
```
yuja, line 619: nit: why not inline the `match` as `OpsetEvaluationError::OpsetResolution(OpsetResolutionError::MultipleOperations { expr, candidates })) ...`?

If you think it's too verbose, maybe we can extract a helper function that unwraps the inner values. Another option is to use `insta::assert_debug_snapshot!()`.

lib/src/op_walk.rs
```diff
@@ -158,7 +158,7 @@ fn resolve_single_op(
             0 => Err(OpsetResolutionError::EmptyOperations(op_str.to_owned()))?,
             1 => neighbor_ops.pop().unwrap(),
             _ => Err(OpsetResolutionError::MultipleOperations {
-                expr: op_str.to_owned(),
+                expr: format!("{op_symbol}{}", &op_postfix[..=i]),
                 candidates: neighbor_ops.iter().map(|op| op.id().clone()).collect(),
```
yuja, line 162: nit: maybe this change should also apply to `EmptyOperations` (though it's guaranteed that ancestors/descendants of an empty operation are empty)?

lib/src/op_walk.rs
```diff
@@ -158,7 +158,7 @@ fn resolve_single_op(
             0 => Err(OpsetResolutionError::EmptyOperations(op_str.to_owned()))?,
             1 => neighbor_ops.pop().unwrap(),
             _ => Err(OpsetResolutionError::MultipleOperations {
-                expr: op_str.to_owned(),
+                expr: format!("{op_symbol}{}", &op_postfix[..=i]),
```
yuja, line 161: nit: `op_str[..=op_symbol.len()+i]`?

CHANGELOG.md
```diff
@@ -59,6 +59,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   `core.fsmonitor` gitconfig is set in the global or system gitconfigs.
   [#6440](https://github.com/jj-vcs/jj/issues/6440)
 
+* When running `jj op abandon`/`diff`/`restore`/`show`/`undo` and multiple operations
+  are found, the correct op set subexpression is shown, matching the op ids suggested
+  in the accompanying hint.
```
yuja, line 64: nit: I wouldn't add changelog entry for minor improvements of error/hint messages.

lib/src/op_walk.rs
cenviity: I did consider applying the same change here, but since there is no hint provided for `EmptyOperations` in `opset_resolution_error_hint()` (and there is no useful hint for the user to take action on anyway), we don't have to worry about op ids being incoherent with the expression shown to the user. So I've left it unchanged. As you said, the full op set expression is guaranteed to be empty in this case.

yuja: Can you leave a comment then? The code looks inconsistent without reasoning.

lib/tests/test_operations.rs
cenviity: I tried `insta::assert_debug_snapshot!()` but didn't like it as we would lose the explicitly named relationships among grandparents, parents and children expressions.

I've instead inlined the pattern and extracted a helper function to reduce the noise a bit.

lib/src/op_walk.rs
cenviity: Done.

CHANGELOG.md
cenviity: Makes sense. Removed.

lib/src/op_walk.rs
cenviity: Done.

lib/tests/test_operations.rs
```diff
@@ -687,3 +678,17 @@ fn test_gc() {
     assert_eq!(expected_op_entries.len(), 1);
     assert_eq!(expected_view_entries.len(), 1);
 }
+
+fn assert_multiple_operations_error(
```
yuja, line 691: nit: add `#[track_caller]`

It's also good to move the assertion to caller by making this function return `(expr, candidates)`.

yuja (approved):
Thanks!

lib/tests/test_operations.rs
cenviity: Thanks, I've rewritten the helper function to do that.

PR #6572, opened 2025-05-19
backend: make read_file() return a AsyncRead

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,7 +102,7 @@ test-case = "3.3.1"
 textwrap = "0.16.2"
 thiserror = "2.0.12"
 timeago = { version = "0.4.2", default-features = false }
-tokio = { version = "1.44.2" }
+tokio = { version = "1.44.2", features = ["io-util"] }
 toml_edit = { version = "0.22.26", features = ["serde"] }
 tracing = "0.1.41"
 tracing-chrome = "0.7.2"
--- a/lib/src/conflicts.rs
+++ b/lib/src/conflicts.rs
@@ -15,9 +15,9 @@
 #![allow(missing_docs)]
 
 use std::io;
-use std::io::Read;
 use std::io::Write;
 use std::iter::zip;
+use std::pin::Pin;
 
 use bstr::BString;
 use bstr::ByteSlice as _;
@@ -27,6 +27,8 @@ use futures::Stream;
 use futures::StreamExt as _;
 use itertools::Itertools as _;
 use pollster::FutureExt as _;
+use tokio::io::AsyncRead;
+use tokio::io::AsyncReadExt as _;
 
 use crate::backend::BackendError;
 use crate::backend::BackendResult;
@@ -97,11 +99,11 @@ async fn get_file_contents(
 ) -> BackendResult<BString> {
     match term {
         Some(id) => {
+            let mut reader = store.read_file(path, id).await?;
             let mut content = vec![];
-            store
-                .read_file(path, id)
-                .await?
+            reader
                 .read_to_end(&mut content)
+                .await
                 .map_err(|err| BackendError::ReadFile {
                     path: path.to_owned(),
                     id: id.clone(),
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -102,7 +103,7 @@ test-case = "3.3.1"
 textwrap = "0.16.2"
 thiserror = "2.0.12"
 timeago = { version = "0.4.2", default-features = false }
-tokio = { version = "1.44.2" }
+tokio = { version = "1.44.2", features = ["fs", "io-util"] }
```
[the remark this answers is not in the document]
martinvonz: Correct. Fixed.

lib/src/file_util.rs
```diff
@@ -162,6 +168,68 @@ pub fn persist_content_addressed_temp_file<P: AsRef<Path>>(
...
+                }
+            }
+        }
+    }
+}
+
+/// Reads from an async source and writes to a sync destination. Does not spawn
+/// a task, so writes will block.
+pub async fn copy_async_to_sync<R: AsyncRead, W: Write>(reader: R, writer: W) -> io::Result<u64> {
+    AsyncToSyncCopier {
```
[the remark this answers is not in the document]
martinvonz: Heh, yes. I think I did it this way because `AsyncRead` only has `poll_read()` and I didn't think to check if there was a `.read()` in `AsyncReadExt` (there is, of course). I've rewritten it to use that. Thanks!

lib/src/git_backend.rs
```diff
@@ -425,17 +425,6 @@ impl GitBackend {
...
-    fn read_file_sync(&self, id: &FileId) -> BackendResult<Box<dyn Read>> {
-        let git_blob_id = validate_git_object_id(id)?;
-        let locked_repo = self.lock_git_repo();
-        let mut blob = locked_repo
-            .find_object(git_blob_id)
-            .map_err(|err| map_not_found_err(err, id))?
-            .try_into_blob()
-            .map_err(|err| to_read_object_err(err, id))?;
-        Ok(Box::new(Cursor::new(blob.take_data())))
-    }
```
yuja, line 437: Perhaps, a better option is to make this return `Vec<u8>`? Then, `read_conflict()` can just do `serde_json::from_slice(&data)`. (sorry I didn't notice before.)

lib/src/file_util.rs
```diff
@@ -162,6 +165,26 @@ pub fn persist_content_addressed_temp_file<P: AsRef<Path>>(
     }
 }
 
+/// Reads from an async source and writes to a sync destination. Does not spawn
+/// a task, so writes will block.
+pub async fn copy_async_to_sync<R: AsyncRead, W: Write>(
+    mut reader: R,
+    mut writer: W,
+) -> io::Result<usize> {
```
yuja, line 173: A similar function in tokio takes `&mut AsyncRead + Unpin` instead of pinning internally. I'm not sure which is better, and I don't think that would matter much (because traits are implemented for `&mut T`.)

https://docs.rs/tokio/1.45.0/tokio/io/fn.copy.html

For the writer side, taking `&mut W` might make more sense because the writer can be reused.

lib/src/file_util.rs
```diff
@@ -256,4 +281,14 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_copy_async_to_sync() {
+        let input = b"hello";
+        let mut output = vec![];
+
+        let result = copy_async_to_sync(Cursor::new(&input), &mut output).block_on();
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), 5);
```
yuja, line 292: nit: can you add test for data bigger than the buffer size?

lib/src/file_util.rs
martinvonz: I don't know which is better either. I updated the writer to be `&mut W`.

lib/src/file_util.rs
martinvonz: Makes sense. Done.

lib/src/conflicts.rs
```diff
@@ -162,6 +164,7 @@ impl MaterializedFileValue {
         let mut buf = Vec::new();
         self.reader
             .read_to_end(&mut buf)
+            .block_on()
```
yuja, line 167: nit: this can be changed to `async fn` (can be addressed later)

yuja (approved):

lib/src/conflicts.rs
martinvonz: True. I'll do that later.

PR #6573, opened 2025-05-19
cli undo: detect "double undo" from operation metadata instead

cenviity:
Based on refactoring in #6589.

To identify an `undo` operation, checking that the views of the "bad" op and its grandparent are equal is necessary but not sufficient. For example, creating a bookmark and then immediately deleting it also "resets" the view. This commit fixes this by also checking the "description" in the operation metadata to identify "double undo" properly.

With this commit, the "double undo" warning is now shown whether users undid the latest undo implicitly (with `jj (op) undo`) or explicitly (with `jj (op) undo <op set>` or `jj (op) undo @`).

If undoing an older undo (prior to `@`), the warning is not shown.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/operation/undo.rs
+++ b/cli/src/commands/operation/undo.rs
@@ -63,18 +63,16 @@ pub fn cmd_op_undo(
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
     let bad_op = workspace_command.resolve_single_op(&args.operation)?;
-    let mut parent_ops = bad_op.parents();
-    let Some(parent_op) = parent_ops.next().transpose()? else {
-        return Err(user_error("Cannot undo root operation"));
+    let parent_of_bad_op = match bad_op.parents().at_most_one() {
+        Ok(Some(parent_of_bad_op)) => parent_of_bad_op?,
+        Ok(None) => return Err(user_error("Cannot undo root operation")),
+        Err(_) => return Err(user_error("Cannot undo a merge operation")),
     };
-    if parent_ops.next().is_some() {
-        return Err(user_error("Cannot undo a merge operation"));
-    }
 
     let mut tx = workspace_command.start_transaction();
     let repo_loader = tx.base_repo().loader();
     let bad_repo = repo_loader.load_at(&bad_op)?;
-    let parent_repo = repo_loader.load_at(&parent_op)?;
+    let parent_repo = repo_loader.load_at(&parent_of_bad_op)?;
     tx.repo_mut().merge(&bad_repo, &parent_repo)?;
     let new_view = view_with_desired_portions_restored(
         tx.repo().view().store_view(),
@@ -90,7 +88,7 @@ pub fn cmd_op_undo(
     }
     tx.finish(ui, format!("undo operation {}", bad_op.id().hex()))?;
 
-    if args.operation == "@" && is_undo(&bad_op, &parent_op)? {
+    if args.operation == "@" && is_undo(&bad_op, &parent_of_bad_op)? {
         writeln!(
             ui.hint_default(),
             "This action reverted an 'undo' operation. The repository is now in the same state as \
--- a/cli/src/commands/operation/undo.rs
+++ b/cli/src/commands/operation/undo.rs
@@ -47,13 +47,17 @@ pub struct OperationUndoArgs {
     what: Vec<UndoWhatToRestore>,
 }
 
-fn is_undo(op: &Operation, parent_op: &Operation) -> Result<bool, OpStoreError> {
-    let grand_parents: Vec<_> = parent_op.parents().try_collect()?;
-    if let [grand_parent_op] = &grand_parents[..] {
-        Ok(op.view_id() == grand_parent_op.view_id())
-    } else {
-        Ok(false)
-    }
+// Checks whether `op` resets the view of `parent_op` to the view of the
+// grandparent op.
+//
+// This is a necessary condition for `op` to be an undo of `parent_op` but is
+// not sufficient. For example, deleting a bookmark also resets the view
+// similarly but is not a literal `undo` operation.
+fn resets_view_of(op: &Operation, parent_op: &Operation) -> Result<bool, OpStoreError> {
+    let Ok(grandparent_op) = parent_op.parents().exactly_one() else {
+        return Ok(false);
+    };
+    Ok(op.view_id() == grandparent_op?.view_id())
 }
 
 pub fn cmd_op_undo(
@@ -88,7 +92,7 @@ pub fn cmd_op_undo(
     }
     tx.finish(ui, format!("undo operation {}", bad_op.id().hex()))?;
 
-    if args.operation == "@" && is_undo(&bad_op, &parent_of_bad_op)? {
+    if args.operation == "@" && resets_view_of(&bad_op, &parent_of_bad_op)? {
         writeln!(
             ui.warning_default(),
             "The second-last `jj undo` was reverted by the latest `jj undo`. The repo is now in \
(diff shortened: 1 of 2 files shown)


cli/src/commands/operation/undo.rs
```diff
@@ -93,13 +93,13 @@ pub fn cmd_op_undo(
     if args.operation == "@" && is_undo(&bad_op, &parent_op)? {
         writeln!(
             ui.warning_default(),
-            "This action reverted an 'undo' operation. The repository is now in the same state as \
-             it was before the original 'undo'."
+            "A bare 'undo' operation was reverted. The repo is now in the same state as it was \
```
yuja, line 91: nit: I'm not sure what "a bare 'undo'" means. What "undo" operation is categorized as bare (or non-bare)?

cli/src/commands/operation/undo.rs
```diff
@@ -47,13 +45,10 @@ pub struct OperationUndoArgs {
...
-fn is_undo(op: &Operation, parent_op: &Operation) -> Result<bool, OpStoreError> {
-    let grand_parents: Vec<_> = parent_op.parents().try_collect()?;
-    if let [grand_parent_op] = &grand_parents[..] {
-        Ok(op.view_id() == grand_parent_op.view_id())
-    } else {
-        Ok(false)
-    }
+fn is_bare_undo(op: &Operation) -> bool {
+    let tags = &op.metadata().tags;
+    let args = tags.get("args").expect("`args` key should exist");
```
yuja, line 50: `tags` can be empty.

If we have to rely on the metadata, it's probably better to compare `description`. Command arguments would have more variants than description formatted programatically.

cli/src/commands/operation/undo.rs
cenviity: I inherited the "'bare' undo" term from the existing `test_shows_a_warning_when_undoing_an_undo_operation_as_bare_jj_undo()` test and an [earlier implementation](https://github.com/jj-vcs/jj/pull/5580#discussion_r1941264483) of `is_undo()` to indicate an invocation of `jj (op) undo` without any arguments.

From #5580 (and #3700 before that), what we want to alert users to is the fact that after running `jj (op) undo` twice in a row the repo is back to the state before the first `undo`. Users may have expected to have "gone back in time" by running `undo` multiple times, when instead they are essentially back in the same state as before.

cenviity: I agree that this warning could be worded better without alluding to the vague "'bare' undo" term though. I'll put up a new version.

cli/src/commands/operation/undo.rs
cenviity: Originally my intention was to warn users when they run two `undo` commands with no arguments consecutively. That was why the implementation leveraged tags (with the user-supplied command input) rather than the description (which would always include the op id, even if the command input didn't specify it). In that context, running `jj undo` followed by `jj undo @` or `jj undo <op id of the previous undo>` would not trigger a warning.

After a bit of rethinking, I wonder if it would make sense to warn unconditionally on a double `undo`, whether users did specify an op set argument or not for either `undo` command. In this context, running `jj undo` then `jj undo @` or `jj undo <op id of the previous undo>` **would** trigger a warning.

Under both approaches, since we are no longer comparing the views of the `bad_op` and its grandparent op, the spurious warning after undoing a bookmark deletion would still no longer appear.

cli/src/commands/operation/undo.rs
yuja: Ah, okay, in that context, the idea was `jj undo` issues a warning, whereas `jj op undo` doesn't. This idea was rejected because it would be confusing.

cenviity: Oh, looks like I misinterpreted what the "bare" meant. I thought it refers to the lack of arguments, but IIUC it refers to `jj undo` as opposed to the fuller form `jj op undo` then.

cli/src/commands/operation/undo.rs
yuja: I don't have strong feeling, but to reduce false positives, maybe we can check `description` in addition to `op.view_id() == grand_parent_op.view_id()`?

cli/src/commands/operation/undo.rs
cenviity: I've updated the warning to remove any mention of "'bare' undo" now.

cli/src/commands/operation/undo.rs
cenviity: I've ended up rewriting most of this patch to check for the `description` in the operation metadata in addition to the existing `view` check.

cli/tests/test_undo.rs
```diff
@@ -520,6 +520,8 @@ fn test_undo_latest_undo_explicitly() {
     Undid operation: 94337eb25498 (2001-02-03 08:05:09) undo operation c7b028ea7b47461d4328dde306e13337ea9000b6abfde4cb751902fae3124d578f2e0082cbd1e1d32b5e64afc001a933be5acc81a015a7f15d62d90750eaa9ca
     Working copy  (@) now at: rlvkpnrz 43444d88 (empty) (no description set)
     Parent commit (@-)      : qpvuntsm e8849ae1 (empty) (no description set)
+    Warning: The second-last `jj undo` was reverted by the latest `jj undo`. The repo is now in the same state as it was before the second-last `jj undo`.
+    Hint: To undo multiple operations, use `jj op log` to see past states and `jj op restore` to restore one of these states.
```
yuja, line 524: Perhaps, this doesn't need a warning since the user explicitly specified the operation to be reverted.

cli/src/commands/operation/undo.rs
```diff
@@ -60,23 +62,31 @@ fn resets_view_of(op: &Operation, parent_op: &Operation) -> Result<bool, OpStore
     Ok(op.view_id() == grandparent_op?.view_id())
 }
 
+// Checks if `op` reverts the `maybe_bad_op` operation.
+fn is_undo_of(op: &Operation, maybe_bad_op: &Operation) -> bool {
+    op.metadata()
+        .description
+        .strip_prefix(UNDO_OPERATION_DESCRIPTION_PREFIX)
+        .is_some_and(|op_id| op_id == maybe_bad_op.id().hex())
+}
```
yuja, line 71: nit: I think checking the description prefix is good enough, but if we require exact match, it's simpler to extract `tx_description(op) -> String` function? The extra formatting cost wouldn't matter.

cli/src/commands/operation/undo.rs
```diff
@@ -63,13 +63,11 @@ pub fn cmd_op_undo(
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
     let bad_op = workspace_command.resolve_single_op(&args.operation)?;
-    let mut parent_ops = bad_op.parents();
-    let Some(parent_op) = parent_ops.next().transpose()? else {
-        return Err(user_error("Cannot undo repo initialization"));
+    let parent_op = match bad_op.parents().at_most_one() {
+        Ok(Some(parent_op)) => parent_op?,
+        Ok(None) => return Err(user_error("Cannot undo repo initialization")),
```
yuja, line 82: nit: "Cannot undo root operation"

This should have been updated when we introduce the virtual root operation.

yuja (approved):

cli/tests/test_undo.rs
cenviity: True. I suppose we can assume if the user had gone to the trouble of looking up the op id, then they did mean to undo the undo.

cli/src/commands/operation/undo.rs
cenviity: Done.

cli/tests/test_undo.rs
cenviity: Reverted back to only checking if the current `undo`'s argument is exactly `@`, so that if an op set was supplied we skip the warning.

cenviity:
@yuja Thanks for the thorough review!

PR #6581, opened 2025-05-20
rebase: fix rebase with abandoned divergent commit as root

scott2000:
# Checklist

If applicable:


yuja:
Perhaps, we need to abandon duplicated commits _after_ they get rebased (or record as such) so that the descendants are automatically rebased to the destination, not to the source. Sorry I didn't notice that.

Naming aside, maybe we can reorganize the API as follows?
```
compute_move_commits(..)
  .find_duplicate_divergent_commits(..) // collect into to_abandon: HashSet<CommitId>
  .apply(..) // call rewriter.abandon()
```

scott2000:
> Naming aside, maybe we can reorganize the API as follows?
> 
> ```
> compute_move_commits(..)
>   .find_duplicate_divergent_commits(..) // collect into to_abandon: HashSet<CommitId>
>   .apply(..) // call rewriter.abandon()
> ```

That looks good to me.

scott2000:
I implemented it like this for now since it seemed simpler to me (I'm also unsure about the naming):

```rust
compute_move_commits(..)
  .abandoning(commits_to_abandon)
  .apply(..)
```

What do you think?

yuja:
Seems good. Maybe the function can be named as `record_to_abandon(&mut self, commit_ids)`?

scott2000:
> Seems good. Maybe the function can be named as `record_to_abandon(&mut self, commit_ids)`?

Yeah, that seems a bit clearer. I updated it.

cli/tests/test_rebase_command.rs
```diff
@@ -2978,54 +2978,74 @@ fn test_rebase_skip_duplicate_divergent() {
...
+    @  d  znkkpsqq  ecbe1d2f:  b1
     ○  b1  zsuskuln  48bf33ab:  a
     ○  a  rlvkpnrz  08789390
     ◆    zzzzzzzz  00000000
     [EOF]
     ");
 
+    // Rebasing should work even if the root of the target set is abandoned
+    work_dir.run_jj(["undo"]).success();
+    insta::assert_snapshot!(work_dir.run_jj(["rebase", "-s", "b1", "-d", "b2"]).success(), @r"
```
yuja, line 3018: nit: `.success()` isn't needed unless you want to make it fail immediately even in `cargo insta test`.

lib/src/rewrite.rs
```diff
@@ -433,8 +434,21 @@ impl ComputedMoveCommits {
             target_commit_ids: IndexSet::new(),
             descendants: vec![],
             commit_new_parents_map: HashMap::new(),
+            to_abandon: HashSet::new(),
         }
     }
+
+    pub fn record_to_abandon(&mut self, commit_ids: impl IntoIterator<Item = CommitId>) {
```
yuja, line 441: nit: can you add doc comment describing difference from doing `mut_repo.record_abandoned_commit()`?

yuja (approved):

lib/src/rewrite.rs
scott2000: I added a doc comment, but I'm not sure if I explained it well. Could you take a look and let me know if there's a better way to explain it?

yuja: LG, thanks.

PR #6582, opened 2025-05-20
cli: fix: emit an informative warning when a tool exits with a nonzero status

winterqt:
# Checklist

If applicable:


cli/src/commands/fix.rs
```diff
@@ -194,20 +193,41 @@ fn fix_one_file(
...
+                    Err(Some(e)) => {
+                        let stderr_text = if e.stderr.is_empty() {
+                            "(no stderr)".to_string()
+                        } else {
+                            format!("'{}'", String::from_utf8_lossy(&e.stderr).trim())
+                        };
+
+                        writeln!(
+                            ui.warning_default(),
+                            "tool '{}' exited with status {} when fixing '{}': {}",
```
winterqt, line 215: I'm not 100% sure how we want to fit the commit in question here, it seems a bit too info-dense already.

martinvonz: We don't even know the commit here, right? The same version of the file may appear in multiple commits. Seems fine to not say which commit. We can improve it later if we find that it's unclear.

winterqt: Good point — I was just thinking of how to format it rather than the actual implementation. I’d be fine with just leaving the warning as is and go from there.

Though I guess that begs the question: even if the same version of a file can appear in multiple commits, it may be useful to somehow have a way to show the copy that failed? e.g. if one commit implemented a parsing error that caused a formatter to fail that a subsequent commit fixed, it may be useful to get some sort of pointer to what the broken one was?

cli/src/commands/fix.rs
```diff
@@ -195,20 +194,41 @@ async fn fix_one_file(
...
```
martinvonz, line 216: nit: We generally start sentences with uppercase in error messages

cli/src/commands/fix.rs
```diff
@@ -195,20 +194,41 @@ async fn fix_one_file(
...
+                            tool_config.name,
+                            e.status.code().unwrap_or_default(),
+                            // We use `RepoPath::as_internal_file_string` in this warning
+                            // as it's what is (optionally) passed as an argument to the
+                            // tool.
```
martinvonz, line 221: Hmm, that seems like a minor bug. @hooper, do you remember if there's a reason we don't format the path using using backslashes on Windows (as `path.to_fs_path()` would do)?

cli/src/commands/fix.rs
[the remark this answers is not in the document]
winterqt: Ah, good point, even the tool that made me write this uses multiple lines.

cli/src/commands/fix.rs
```diff
@@ -226,16 +246,15 @@ async fn fix_one_file(
...
+/// `Err(Some(ToolExecError))` if the tool exited with a nonzero status.
 fn run_tool(
     workspace_root: &Path,
     tool_command: &CommandNameAndArgs,
     file_to_fix: &FileToFix,
     old_content: &[u8],
-) -> Result<Vec<u8>, ()> {
-    // TODO: Pipe stderr so we can tell the user which commit, file, and tool it is
-    // associated with.
+) -> Result<Vec<u8>, Option<ToolExecError>> {
```
yuja, line 257: nit: `enum Wait(io::Error)|{ status, stderr }` instead of `Option`?

PR #6596, opened 2025-05-21
lib, cli: migrate from `chrono` to `jiff` for timestamp handling

winterqt:
Jiff [0] is a datetime library which "encourages you to jump into the pit of success" by "providing high level datetime primitives that are difficult to misuse and have reasonable performance."

What a perfect match for us. :)

See [1] for a comparison between Jiff and Chrono, but the main advantage for us codebase-wise is that `Zoned`, Jiff's main datetime type, integrates the timezone without a generic. This lets us get rid of `DatePatternContext` entirely, replacing it with just a `Zoned` to represent the current time. It's also generally faster than Chrono, which is an obvious plus.

[0]: https://github.com/BurntSushi/jiff
[1]: https://docs.rs/jiff/latest/jiff/_documentation/comparison/index.html#chrono-v0438

# Checklist

If applicable:


cli/src/template_builder.rs
```diff
@@ -1223,10 +1224,19 @@ fn builtin_timestamp_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
...
-                    time_util::FormattingItems::parse(format)
-                        .ok_or_else(|| TemplateParseError::expression("Invalid time format", span))
+                    // We format a dummy timestamp in order to validate the format string.
+                    match strtime::format(
+                        format,
+                        &jiff::civil::date(2020, 12, 20)
+                            .at(6, 37, 0, 0)
+                            .to_zoned(jiff::tz::TimeZone::UTC)
+                            .unwrap(),
+                    ) {
```
PhilipMetzger, line 1234: this would be the perfect place to add the date of the first commit in the private Mercurial repo as an easter egg cc @martinvonz

cli/tests/test_revset_output.rs
```diff
@@ -745,23 +742,23 @@ fn test_revset_committer_date_with_time_zone() {
 
     work_dir
         .run_jj([
-            "--config=debug.commit-timestamp=2023-01-25T11:30:00-05:00",
+            "--config=debug.commit-timestamp=\"2023-01-25T11:30:00[-05:00]\"",
```
PhilipMetzger, line 745: nit: convert to a raw string to not avoid the needless escaping

cli/tests/test_revset_output.rs
```diff
@@ -745,23 +742,23 @@ fn test_revset_committer_date_with_time_zone() {
...
             "describe",
             "-m",
             "first",
         ])
         .success();
     work_dir
         .run_jj([
-            "--config=debug.commit-timestamp=2023-01-25T12:30:00-05:00",
+            "--config=debug.commit-timestamp=\"2023-01-25T12:30:00[-05:00]\"",
```
PhilipMetzger, line 753: ditto for all similar changes

docs/templates.md
```diff
@@ -399,7 +399,7 @@ The following methods are defined.
 
 * `.ago() -> String`: Format as relative timestamp.
 * `.format(format: String) -> String`: Format with [the specified strftime-like
-  format string](https://docs.rs/chrono/latest/chrono/format/strftime/).
+  format string](https://docs.rs/jiff/latest/jiff/fmt/strtime/index.html#conversion-specifications).
```
PhilipMetzger, line 402: Nit: since this is user-visible it should have a deprecation period before changing or at least a warning for the deprecated format

PhilipMetzger (commented):
Thanks, here are some nits and a deprecation discussion

cli/src/template_builder.rs
winterqt: This date wasn't quite so random either... 😉

docs/templates.md
winterqt: The formats are *basically* identical, so I'd be surprised if this actually affected anyone. (Also, see the changelog entry.)

cli/tests/test_revset_output.rs
winterqt: Fair enough.

cli/tests/test_revset_output.rs
```diff
@@ -770,7 +767,7 @@ fn test_revset_committer_date_with_time_zone() {
 
     let mut log_commits_before_and_after = |committer_date: &str, now: &str, tz: &str| {
         test_env.add_env_var("TZ", tz);
-        let config = format!("debug.commit-timestamp={now}");
+        let config = format!(r#"debug.commit-timestamp="{now}""#);
```
winterqt, line 770: tbh I think reverting just this one may be fine but meh they're both ugly given the interpolation
```suggestion
        let config = format!("debug.commit-timestamp=\"{now}\"");
```

lib/src/revset.rs
```diff
@@ -2580,7 +2580,7 @@ pub struct RevsetParseContext<'a> {
     pub aliases_map: &'a RevsetAliasesMap,
     pub local_variables: HashMap<&'a str, ExpressionNode<'a>>,
     pub user_email: &'a str,
-    pub date_pattern_context: DatePatternContext,
+    pub date_pattern_now: Zoned,
```
winterqt, line 2583: Open to a better name.

lib/src/revset.rs
```diff
@@ -984,7 +984,7 @@ pub fn expect_string_pattern(
 pub fn expect_date_pattern(
     diagnostics: &mut RevsetDiagnostics,
     node: &ExpressionNode,
-    context: &DatePatternContext,
+    zoned: &Zoned,
```
winterqt, line 987: Might be best to just take a `Zoned` here and clone at the call sites? 🤷‍♀️

cli/src/template_builder.rs
PhilipMetzger: I'm pretty sure its the initial Git import :-).

winterqt: Indeed: 25caaa0081223508c45b1dfe9feeac60c925eb45 ;-)

winterqt: Guess it might make sense to do 6b1427cb46a68315d66fa93ea1ebf54b0ecc4b0a if not the hg repo though.

docs/templates.md
PhilipMetzger: The problem is that _almost identical_ interfaces will depend on [Hyrum's Law], so any migration is not that easy. OK. then it should be fine with minor issues.  

[Hyrum's Law]: https://www.hyrumslaw.com/

cli/src/cli_util.rs
```diff
@@ -828,17 +829,18 @@ impl WorkspaceCommandEnvironment {
             workspace_name: &self.workspace_name,
         };
         let now = if let Some(timestamp) = self.settings.commit_timestamp() {
-            chrono::Local
-                .timestamp_millis_opt(timestamp.timestamp.0)
-                .unwrap()
+            jiff::Zoned::new(
+                Timestamp::from_millisecond(timestamp.timestamp.0).unwrap(),
+                TimeZone::system(),
+            )
```
BurntSushi, line 835: There is also [`Timestamp::to_zoned`](https://docs.rs/jiff/latest/jiff/struct.Timestamp.html#method.to_zoned), which might be a bit more succinct here.

lib/src/backend.rs
```diff
@@ -73,15 +74,13 @@ pub struct Timestamp {
...
 
-    pub fn from_datetime<Tz: chrono::TimeZone<Offset = chrono::offset::FixedOffset>>(
-        datetime: chrono::DateTime<Tz>,
-    ) -> Self {
+    pub fn from_zoned(zoned: Zoned) -> Self {
         Self {
-            timestamp: MillisSinceEpoch(datetime.timestamp_millis()),
-            tz_offset: datetime.offset().local_minus_utc() / 60,
+            timestamp: MillisSinceEpoch(zoned.timestamp().as_millisecond()),
+            tz_offset: zoned.time_zone().to_offset(zoned.timestamp()).seconds() / 60,
```
BurntSushi, line 83: A `Zoned` has an offset inside of it for its timestamp, so you shouldn't need to re-compute it here. [`Zoned::offset()`](https://docs.rs/jiff/latest/jiff/struct.Zoned.html#method.offset) should do the trick.

cli/src/cli_util.rs
```diff
@@ -50,6 +49,8 @@ use indexmap::IndexMap;
 use indexmap::IndexSet;
 use indoc::writedoc;
 use itertools::Itertools as _;
+use jiff::tz::TimeZone;
+use jiff::Timestamp;
```
yuja, line 53: nit: use qualified `jiff::Timestamp` to avoid future name conflicts with our `Timestamp` type?

cli/src/template_builder.rs
yuja: Just construct zeroed `backend::Timestamp` and convert it to `Zoned`?

cli/src/time_util.rs
```diff
@@ -1,97 +1,37 @@
...
 
-/// Parsed formatting items which should never contain an error.
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub struct FormattingItems<'a> {
-    items: Vec<chrono::format::Item<'a>>,
+fn datetime_from_timestamp(context: &Timestamp) -> Result<Zoned, jiff::Error> {
+    Ok(
+        jiff::Timestamp::from_millisecond(context.timestamp.0)?.to_zoned(TimeZone::fixed(
+            Offset::constant((context.tz_offset / 60).try_into().unwrap_or_default()),
+        )),
```
yuja, line 11: Use `Offset::from_hours()` and propagate error? The doc says `Offset::constant()` may panic.

BTW, it's probably better to add `timestamp.to_zoned()`.

cli/tests/test_config_command.rs
```diff
@@ -310,9 +310,9 @@ fn test_config_list_origin() {
     test-layered-key = "test-layered-val" # repo $TEST_ENV/repo/.jj/repo/config.toml
     user.name = "Test User" # env
     user.email = "test.user@example.com" # env
-    debug.commit-timestamp = "2001-02-03T04:05:11+07:00" # env
+    debug.commit-timestamp = "2001-02-03T04:05:11+07:00[+07:00]" # env
```
yuja, line 313: This looks a bit weird (though it might be technically correct to show `+07:00` twice.) Maybe we can set TOML-compatible timestamp string instead of using jiff's default formatting?

cli/tests/test_revset_output.rs
```diff
@@ -796,7 +793,7 @@ fn test_revset_committer_date_with_time_zone() {
     };
 
     let (before_log, after_log) =
-        log_commits_before_and_after("2023-01-25 12:00", "2023-02-01T00:00:00-05:00", NEW_YORK);
+        log_commits_before_and_after("2023-01-25 12:00", "2023-02-01T00:00:00[-05:00]", NEW_YORK);
```
yuja, line 796: unrelated change?

lib/src/revset.rs
yuja: I would call it `current_time` or something.

cli/src/time_util.rs
winterqt: What timezone would that returned `Zoned` have, though? Do we want to make it take a TZ parameter, or just default to UTC, or system?

yuja: > What timezone would that returned `Zoned` have, though?

The fixed offset as we do in this function.

winterqt: Oh, wait, sorry, you mean `jj_lib::Timezone::to_zoned` would take the tz as a parameter, right? Not sure why I was thinking otherwise.

yuja: No, I just meant move `datetime_from_timestamp()` to `Timestamp::to_zoned() -> Result<Zoned, ..>`. The `Timestamp` knows the fixed timezone.

cli/src/time_util.rs
```diff
@@ -1,97 +1,37 @@
...
-use thiserror::Error;
 
-/// Parsed formatting items which should never contain an error.
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub struct FormattingItems<'a> {
-    items: Vec<chrono::format::Item<'a>>,
+fn datetime_from_timestamp(context: &Timestamp) -> Result<Zoned, jiff::Error> {
+    Ok(
+        jiff::Timestamp::from_millisecond(context.timestamp.0)?.to_zoned(TimeZone::fixed(
+            Offset::constant((context.tz_offset / 60).try_into().unwrap_or_default()),
```
yuja, line 10: nit: instead of falling back to `0`, maybe we can clamp by `i8::MIN`/`MAX`. Then jiff will report overflow error.

cli/src/cli_util.rs
winterqt: Not sure why I used this elsewhere but not here. Thanks!

winterqt:
Rebased + some fixes to the above review comments (probably missed marking some as resolved, will do a pass later).

Not 100% sure what's wrong with this implementation so that the `test_revset_committer_date_with_time_zone` failure occurs, but I'll look at it later -- figured I'd push what I have, though.

CHANGELOG.md
```diff
@@ -30,6 +30,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * The previously predefined `amend` alias has been removed. You can restore it
   by setting the config `aliases.amend = ["squash"]`.
 
+* The templating language's [`Timestamp` type](docs/templates.md#timestamp-type)
+  now uses [Jiff](https://github.com/BurntSushi/jiff)'s [`strftime`-like formatting
+  specifiers](https://docs.rs/jiff/latest/jiff/fmt/strtime/index.html#conversion-specifications),
+  which may differ from Chrono's (the previous implementation).
```
winterqt, line 36: Also needs a note about `json(...)`'s format changing, unless I can stop that somehow; need to peek at jiff's strftime again, but its RFC3339 impl obviously differs from chrono's.

BurntSushi: The strftime implementation should be largely compatible, but there are undoubtedly some minor differences (that I cannot remember off the top of my head).

The RFC 3339 formatting should be the same though?

winterqt:
Looking at the interdiff between the two iterations of this change, I'm really not sure what could be causing this! 🙃

The test didn't change between May and now, nor did the underlying implementation of the commit timestamp revset stuff...

---

Thought this might fix it:
```diff
diff --git a/lib/src/backend.rs b/lib/src/backend.rs
index c35f8f71ac..46cafb3a1f 100644
--- a/lib/src/backend.rs
+++ b/lib/src/backend.rs
@@ -114,7 +114,9 @@
         Ok(jiff::Timestamp::from_millisecond(self.timestamp.0)
             .map_err(TimestampOutOfRange)?
             .to_zoned(TimeZone::fixed(
-                Offset::from_seconds(self.tz_offset * 60).map_err(TimestampOutOfRange)?,
+                // Offset::from_seconds(self.tz_offset * 60).map_err(TimestampOutOfRange)?,
+                Offset::from_hours((self.tz_offset / 60).try_into().unwrap())
+                    .map_err(TimestampOutOfRange)?,
             )))
     }
 }
```
but nope, expectedly.

CHANGELOG.md
winterqt: > The RFC 3339 formatting should be the same though?

Sorry, yes, you're right; was mixing up two different issues.

winterqt:
Narrowing it down to the most minimal failing test case:

```rust
let z = Timestamp::from_millisecond(86400000)
    .unwrap()
    .to_zoned(TimeZone::fixed(Offset::from_seconds(-3600).unwrap()));

dbg!(z.to_string()); // => 1970-01-02T00:00:00Z
```

Why is it an hour off than what we expect (`1970-01-01T23:00:00-01:00`)? And why didn't this happen with the previous iteration of this change?

winterqt:
Ah, but running it on the Rust playground produces what we want: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5e92d482b4ef0fbb0e500b3a22b0edf7

winterqt:
Okay yeah so it's `Timestamp`'s RFC 3339 formatting versus `Zoned`'s default format... hm.

BurntSushi:
> Okay yeah so it's `Timestamp`'s RFC 3339 formatting versus `Zoned`'s default format... hm.

`Zoned` uses RFC 9557, which is a superset of RFC 3339. It enables lossless (de)serialization.

winterqt:
So:

- `Timestamp`'s `Display` impl gives us `1970-01-01T00:00:00Z` for our `Timestamp(0, 0)`
	- ...but then gives us an "incorrectly"-formatted (that is, the "wrong" time) for the 86400 case (as above)
- We can try something like `Timestamp::display_with_offset`, but that doesn't give us the Zulu time format that we'd expect from chrono

I wonder what the best way to handle this is, outside of just making this really backwards-incompatible with the `chrono` implementation.

BurntSushi:
If you know the offset from UTC, then you shouldn't use Zulu. Either way you slice it, formatting a `Timestamp` will always give you an RFC 3339 compatible timestamp string.

PR #6599, opened 2025-05-21
cli undo: identify "double undo" by comparing operation descriptions

cenviity:
Fixup to 96e73bda.

Instead of checking if the parent op id is contained in the bad op's description metadata, we construct the "undo operation X" description for the bad op on the fly (using the parent op id X) and compare it with the existing bad op description.

# Checklist

If applicable:


cenviity:
@yuja After re-reading your comment at <https://github.com/jj-vcs/jj/pull/6573#discussion_r2099644630>, I think this is what you meant.

yuja (approved):

PR #6602, opened 2025-05-22
cli: use more precise wording in conflict resolution hint

cenviity:
We already use "conflicted commit" at the very end of the hint, so I've reused that term here.

# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2301,14 +2301,20 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
             .try_collect()?;
 
         if !root_conflict_commits.is_empty() {
-            let instruction = if only_one_conflicted_commit {
-                "To resolve the conflicts, start by updating to it"
+            let suggested_change = if only_one_conflicted_commit {
+                "the conflicted commit"
             } else if root_conflict_commits.len() == 1 {
-                "To resolve the conflicts, start by updating to the first one"
+                "the first conflicted commit"
             } else {
-                "To resolve the conflicts, start by updating to one of the first ones"
+                "one of the first conflicted commits"
             };
-            writeln!(fmt.labeled("hint").with_heading("Hint: "), "{instruction}:")?;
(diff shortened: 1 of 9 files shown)


cli/tests/test_absorb_command.rs
```diff
@@ -182,13 +182,13 @@ fn test_absorb_replace_single_line_hunk() {
     Parent commit (@-)      : kkmpptxz 6068e8fc 2
     New conflicts appeared in 1 commits:
       qpvuntsm 5bdb5ca1 (conflict) 1
-    Hint: To resolve the conflicts, start by updating to it:
+    Hint: To resolve the conflicts, start by creating a commit on top of the conflicted commit:
```
martinvonz, line 185: nit: This line now stands out a bit because it's longer than the others. Wrap after "of"?

martinvonz (approved):

cli/tests/test_absorb_command.rs
cenviity: Done. Also updated the examples in the docs while I was at it.

cli/src/cli_util.rs
```diff
@@ -2301,14 +2301,20 @@ See https://jj-vcs.github.io/jj/latest/working-copy/#stale-working-copy \
...
             } else {
-                "To resolve the conflicts, start by updating to one of the first ones"
+                "one of the first conflicted commits"
             };
-            writeln!(fmt.labeled("hint").with_heading("Hint: "), "{instruction}:")?;
+            writedoc!(
+                fmt.labeled("hint").with_heading("Hint: "),
+                "
+                To resolve the conflicts, start by creating a commit on top of
+                {suggested_change}:
```
yuja, line 2315: nit: we generally avoid constructing English sentence like this because it's potentially bad for i18n.

cenviity: Ah, so it'd be better to revert to a separate string for each case, e.g. "To resolve the conflicts, start by creating a commit on top of the conflicted commit"?

PR #6610, opened 2025-05-22
test: replace `cfg!` macro with `cfg_attr` attribute

cenviity:
Using the `cfg` attribute means the test is not compiled at all in non-macOS environments.

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_config_command.rs
+++ b/cli/tests/test_config_command.rs
@@ -946,12 +946,9 @@ fn test_config_edit_user_new_file() {
 }
 
 // TODO: remove after deprecation period in 0.35
-// NOTE: macOS-only test
+#[cfg_attr(not(target_os = "macos"), ignore)]
 #[test]
 fn test_config_edit_user_deprecated_file() {
-    if !cfg!(target_os = "macos") {
-        return;
-    }
     let test_env = TestEnvironment::default();
     let user_config_path = test_env
         .home_dir()


cli/tests/test_config_command.rs
```diff
@@ -946,12 +946,9 @@ fn test_config_edit_user_new_file() {
 }
 
 // TODO: remove after deprecation period in 0.35
-// NOTE: macOS-only test
+#[cfg(target_os = "macos")]
```
martinvonz, line 949: I think we generally prefer `cfg!` because we want to make sure that macOS devs at least notice compilation errors in this code when they change some API. It is a bit weird that the test runs and passes on macOS, however.

emilazy: Yes, I think this is worse. `#[cfg]` should only be used when necessary (i.e. the code would not even compile on another architecture). Otherwise compiler checking, IDE support, etc. suffer.

PhilipMetzger: > It is a bit weird that the test runs and passes on macOS, however.

It passes because Rust libtest doesn't have something like `GTEST_SKIP()`, so the early return just makes the test succeed on every other platform.

martinvonz: `#[cfg_attr(target_os = "macos", ignore)]` (or whatever the syntax is) seems best.

cenviity: This test is for user config paths in macOS, so perhaps we want `#[cfg_attr(not(target_os = "macos"), ignore)]`? Then the test would still be compiled, enabling compiler and LSP checks, etc., but would not run on non-macOS architectures.

martinvonz: Yes, sorry, that's what I meant :)

martinvonz (approved):

PR #6619, opened 2025-05-23
diff: Add color config for external diff tools

eyvind:
Add config settings to enable and disable color in external diff tools, and pass them on the command line when the tool is invoked. This PR was inspired by the TODO comment in `external.rs`.

(Using the US spelling of colour is _killing me_)

# Checklist

If applicable:

### DIFF
--- a/cli/src/merge_tools/external.rs
+++ b/cli/src/merge_tools/external.rs
@@ -59,6 +59,9 @@ pub struct ExternalMergeTool {
     pub diff_invocation_mode: DiffToolMode,
     /// Whether to execute the tool in the temporary diff directory
     pub diff_do_chdir: bool,
+    /// Arguments to pass to enable and disable color
+    pub enable_color_args: Vec<String>,
+    pub disable_color_args: Vec<String>,
     /// Arguments to pass to the program when editing diffs.
     /// `$left` and `$right` are replaced with the corresponding directories.
     pub edit_args: Vec<String>,
@@ -114,6 +117,8 @@ impl Default for ExternalMergeTool {
             merge_tool_edits_conflict_markers: false,
             conflict_marker_style: None,
             diff_do_chdir: true,
+            enable_color_args: vec![],
+            disable_color_args: vec![],
             diff_invocation_mode: DiffToolMode::Dir,
         }
     }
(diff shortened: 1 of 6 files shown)


cli/src/merge_tools/external.rs
```diff
@@ -114,6 +117,8 @@ impl Default for ExternalMergeTool {
             merge_tool_edits_conflict_markers: false,
             conflict_marker_style: None,
             diff_do_chdir: true,
+            enable_color_args: vec![],
+            disable_color_args: vec![],
```
yuja, line 121: Maybe better to add `diff_color_args` which defaults to `diff_args`? Appending/prepending arguments doesn't always work.

yuja: > I think I see the problem, which is that putting flags after the `left` and `right` paths might not work with some tools (please correct me if I'm wrong),

Correct.

> Is it to have two copies of `diff_args`, one with colour enabled and one disabled?

Yes.

> While that would allow the argument to be put anywhere on the command line, it wouldn't scale to more arguments

That's true, but I don't have any better idea than introducing template-like language.

> I'm working on a PR to pass ignore-space arguments to the external diff tool as well

I don't think external tools should support the same set of command flags as the builtin diffs. It wouldn't scale, and the external tool may have various options which `jj diff` doesn't provide flags. Maybe better to register them with different names.

cli/src/merge_tools/external.rs
```diff
@@ -478,7 +482,20 @@ pub fn invoke_external_diff(
...
+        ),
+        None => &(tool.diff_args.as_slice(), &[] as &[String]),
+    };
+    cmd.args(interpolate_variables(before_args, &patterns));
+    if ui.color() {
+        cmd.args(&tool.enable_color_args);
+    } else {
+        cmd.args(&tool.disable_color_args);
+    }
+    cmd.args(interpolate_variables(after_args, &patterns));
```
eyvind, line 498: @yuja Something like this? I need to add documentation (and tests), but it injects the colour argument at an `$args` marker, or at the end of the command if there's no marker.

This assumes that all extra arguments can be injected in one spot, but that seems like a reasonable assumption.

yuja: I would say no. It works, but looks like a hack. I would simply add separate tools, and make `diff --tool=<name> --context=N`, etc. fail or show warning.

Alternatively, maybe we can pass parameters by environment variables (e.g. `JJ_DIFF_IGNORE_ALL_SPACE=1`) which can be interpreted by shell script wrapper, but I don't think it's good idea either.

eyvind: Ah well, if this feature is too niche I'll maintain it in a private fork. Thanks for the review!

PR #6620, opened 2025-05-24
repo: store commit predecessors in Operation object

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/op_store.rs
+++ b/lib/src/op_store.rs
@@ -369,6 +369,20 @@ pub struct Operation {
     pub view_id: ViewId,
     pub parents: Vec<OperationId>,
     pub metadata: OperationMetadata,
+    /// Mapping from new commit to its predecessors, or `None` if predecessors
+    /// weren't recorded when the operation was written.
+    ///
+    /// * `commit_id: []` if the commit was newly created.
+    /// * `commit_id: [predecessor_id, ..]` if the commit was rewritten.
+    ///
+    /// This mapping may contain transitive predecessors if a commit was
+    /// rewritten multiple times within the same transaction.
+    ///
+    /// Existing commits (including commits imported from Git) aren't tracked
+    /// even if they became visible at this operation.
+    // BTreeMap for ease of deterministic serialization. If the deserialization
+    // cost matters, maybe this can be changed to sorted Vec.
+    pub commit_predecessors: Option<BTreeMap<CommitId, Vec<CommitId>>>,
 }
 
 impl Operation {
@@ -390,6 +404,10 @@ impl Operation {
             view_id: root_view_id,
             parents: vec![],
             metadata,
+            // The root operation is guaranteed to have no new commits.
+            // TODO: Alternatively, the root commit could be "created" at the
+            // root operation?
+            commit_predecessors: Some(BTreeMap::new()),
         }
     }
 }
--- a/lib/src/operation.rs
+++ b/lib/src/operation.rs
@@ -22,6 +22,7 @@ use std::hash::Hash;
 use std::hash::Hasher;
 use std::sync::Arc;
 
+use crate::backend::CommitId;
 use crate::op_store;
 use crate::op_store::OpStore;
 use crate::op_store::OpStoreResult;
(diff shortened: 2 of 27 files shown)


PhilipMetzger:
wrt. Emily's question in the Discord, I think that we should support the now "old" format for at least a year, like other binary changes which have happened previously.

lib/src/op_store.rs
```diff
@@ -369,6 +369,20 @@ pub struct Operation {
...
     pub parents: Vec<OperationId>,
     pub metadata: OperationMetadata,
+    /// Mapping from new commit to its predecessors, or `None` if predecessors
+    /// weren't recorded when the operation was written.
+    ///
+    /// * `commit_id: []` if the commit was newly created.
+    /// * `commit_id: [predecessor_id, ..]` if the commit was rewritten.
+    ///
+    /// This mapping may contain transitive predecessors if a commit was
+    /// rewritten multiple times within the same transaction.
```
martinvonz, line 379: Could you clarify here *how* it contains such transitive predecessors? It's unclear to me if this is saying that if X was rewritten Y and then to Z, this mapping will contain X->Y and Y->Z, or if it will contain only X->Z, or maybe all of them.

lib/src/op_store.rs
```diff
@@ -390,6 +404,10 @@ impl Operation {
             view_id: root_view_id,
             parents: vec![],
             metadata,
+            // The root operation is guaranteed to have no new commits.
+            // TODO: Alternatively, the root commit could be "created" at the
+            // root operation?
```
martinvonz, line 409: That seems inconsistent with other changes not being in the operation that creates them (they're only in the operation that rewrites them).

lib/src/operation.rs
```diff
@@ -117,6 +118,19 @@ impl Operation {
         &self.data.metadata
     }
 
+    /// Returns true if predecessors are recorded in this operation.
+    ///
+    /// This returns false only if the operation was written by old jj.
```
martinvonz, line 123: nit: since we know the current version, we might as well say "written by jj < 0.30"

lib/src/operation.rs
```diff
@@ -117,6 +118,19 @@ impl Operation {
...
+    pub fn stores_commit_predecessors(&self) -> bool {
+        self.data.commit_predecessors.is_some()
+    }
+
+    /// Returns predecessors of the specified commit if recorded.
+    pub fn predecessors_by_commit(&self, commit_id: &CommitId) -> Option<&[CommitId]> {
```
martinvonz, line 129: nit: The name `predecessors_by_commit()` makes me think this returns a mapping of predecessors by commit. I would expect this to be called `predecessors_for_commit()`

martinvonz (approved):

lib/src/op_store.rs
yuja: The former. Does this looks good?
```
    /// This mapping preserves all transitive predecessors if a commit was
    /// rewritten multiple times within the same transaction. For example, if
    /// `X` was rewritten as `Y`, then rebased as `Z`, these modifications are
    /// recorded as `{Y: [X], Z: [Y]}`.
```

lib/src/op_store.rs
yuja: Creation is recorded as `{new_commit_id: []}`, which helps terminate traversal of evolution history. What I'm not certain is whether the root commit is considered existent before the root operation, or newly created at the root operation.

yuja:
> Will older jj versions still be able to parse the repo, at least for a bit?

Yes, they should be able to deserialize the operation written by new jj, and predecessors are duplicated to both commit objects and operation objects for the moment.

lib/src/op_store.rs
yuja: Thinking this a bit more. Since commits created/rewritten in abandoned ancestor operations aren't considered born in the root operation, it'll probably make sense that the root commit existed before the root operation as well.

lib/src/op_store.rs
martinvonz: Yes, sounds good. Thanks!

PR #6623, opened 2025-05-25
builtin merge-tool: add property-based tests

jgreitemann:
This PR was split from #6448 to review and land the bug fixes separately from the property-based testing.

# Checklist

If applicable:

### DIFF
--- a/cli/src/merge_tools/builtin.rs
+++ b/cli/src/merge_tools/builtin.rs
@@ -713,7 +713,11 @@ mod tests {
     use jj_lib::matchers::EverythingMatcher;
     use jj_lib::merge::MergedTreeValue;
     use jj_lib::repo::Repo as _;
+    use proptest_state_machine::prop_state_machine;
+    use proptest_state_machine::StateMachineTest;
     use testutils::assert_tree_eq;
+    use testutils::proptest::Transition;
+    use testutils::proptest::WorkingCopyReferenceStateMachine;
     use testutils::repo_path;
     use testutils::repo_path_component;
     use testutils::TestRepo;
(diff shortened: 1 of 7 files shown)


jgreitemann:
Continuing the discussion in #6448 (@arxanas):

> The state machine approach indeed seems a little limiting. I think you'd have to represent the desire to check only some of the checkboxes as a state transition, which is kind of clunky because it represents an entirely different system.

I managed to adapt the state machine testing in a way that covers partial selection of changes without having to reimplement the "runner" for the state machine test. Technically, selection of changes from the diff depends on the diff, but that would have been tricky to incorporate. I opted for an approach where the initial state includes a selection mask, which is then successively applied to the changes in the diff to decide whether to check them, each step of the way.

This approach is not perfect because diffs might change significantly in a way that breaks the "continuity" of the selections, which might impede the ability to shrink the transitions list. However, in practice, it worked surprisingly well and gave sufficiently easy-to-interpret failures. Sure enough, this test uncovered another bug that is fixed separately.

This sort of "roundtrip" property — first applying parts of a diff to produce an intermediate tree, then applying the complementary changes to it to go all the way to the "right" tree — does not apply to arbitrary diffs. There are certain combinations of selected vs. unselected changes where the selected changes necessary imply parts of the unselected changes, or where the "complementary" changes would end up undoing parts of the selected changes. Examples are given in the SUT's doc comment. The selection of changes is tweaked to avoid these conflicts. (Sidebar: One could argue that scm-record should only allow selections that are not conflicted in this way. It kind of does already, e.g., it will automatically select the file mode change when selecting any of the line additions in a newly created text file.)

As such, the test verifying this "roundtrip" property covers a search space that is not a superset of the original proptest verifying only that the left or right tree is reproduced when selecting none or all of the changes. For this reason, I've created a separate SUT for the "roundtrip" test, so now there are two independent state machine proptests using the same underlying state machine.

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    fn apply(mut state: Self::State, transition: &Self::Transition) -> Self::State {
+        match transition {
+            Transition::Commit => {
+                // Do nothing; this is handled by the system under test.
+            }
+
+            Transition::SetDirEntry { path, dir_entry } => {
+                assert_ne!(path.as_ref(), RepoPath::root());
+                let Self { entries } = &mut state;
```
arxanas: nit: I would just write `let entries = &mut state.entries;` here, or otherwise `let Self { mut entries }` and return a new `State` from the entire `match`.

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,242 @@
...
+                .map(|path| path.to_owned())
+                .unique()
+                .collect_vec(),
+        )
+    }
+
+    fn arb_transition_create(&self) -> impl Strategy<Value = Transition> {
+        (
+            self.arb_extant_directory(),
+            proptest::collection::vec(arb_path_component(), 1..3),
```
arxanas, line 160: question (non-blocking): Does creating the extra path components (`1..3`) seem to trigger any bugs/new code paths? It should be reachable from a sequence of create transitions already. If not, I would weakly prefer to just create one new path component at a time.

lib/testutils/src/proptest.rs
```diff
@@ -0,0 +1,242 @@
...
+    Symlink { target: String },
+    GitSubmodule { commit: CommitId },
+}
+
+impl Arbitrary for DirEntry {
+    type Parameters = ();
+    type Strategy = BoxedStrategy<DirEntry>;
+
+    fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+        // TODO: Currently only creates `File`s; extend test to other variants.
```
arxanas: nit: Can you move/put this comment on each of `Symlink`/`GitSubmodule`? It should be realize at tell at a glance that those code paths aren't tested.

cli/src/merge_tools/builtin.rs
```diff
@@ -844,9 +852,11 @@ mod tests {
         let no_changes_tree_id = apply_diff(store, &left_tree, &right_tree, &changed_files, &files);
         let no_changes_tree = store.get_root_tree(&no_changes_tree_id).unwrap();
         assert_eq!(
-            no_changes_tree.id(),
             left_tree.id(),
-            "no-changes tree was different",
+            no_changes_tree.id(),
+            "no-changes tree was different:\nexpected tree:\n{}\nactual tree:\n{}",
```
arxanas: comment (non-blocking): Might be worth extracting this assertion into a helper function/macro? But don't need to do this now.

arxanas (approved):
Thanks!

> (Sidebar: One could argue that scm-record should only allow selections that are not conflicted in this way. It kind of does already, e.g., it will automatically select the file mode change when selecting any of the line additions in a newly created text file.)

Alternatively, it should return the selected and unselected contents in a way that indicates to you what implied changes were made.

lib/testutils/src/proptest.rs
jgreitemann: Done. After I went through with the `TestTreeBuilder` refactor in the preceding PR, there was no reason anymore for `DirEntry` to be defined in the `testutils` module, so I had already moved it back into the `testutils::proptest` module. Given this, I now decided to use `proptest-derive` after all to derive the implementation of `Arbitrary` for `DirEntry`. The omission of `Symlink`/`GitSubmodule` from the strategy for `DirEntry` thus translates into adding `#[proptest(skip)]` to the variants in the enum which is also a natural place to put those TODO comments.

cli/src/merge_tools/builtin.rs
jgreitemann: I added a new commit for this.

lib/testutils/src/proptest.rs
jgreitemann: Since I've removed `Directory` from `DirEntry` and only store leaf `entries` (with the directories implicitly given by the leaves' ancestors), appending `foo/bar` to an extant directory path effectively creates a new directory `foo` containing the file/symlink/submodule `bar`. Only appending a single new path component at a time would result in a perfectly flat working tree.

jgreitemann:
I've updated the PR to address your comments. Since there're some non-trivial changes, you might wanna take another look? Otherwise, feel free to put it in the merge queue.

PR #6641, opened 2025-05-27
Add copy id to data model

martinvonz:
# Checklist

If applicable:


lib/src/backend.rs
```diff
@@ -194,6 +195,24 @@ pub struct CopyRecord {
...
+    /// IDs of the files that became the current incarnation of this file.
+    ///
+    /// A newly created file has no parents. A regular copy or rename has one
+    /// parent. A merge of multiple files has multiple parents.
+    pub parents: Vec<CopyId>,
+    /// An optional piece of data to give the Copy object a different ID. May be
+    /// randomly generated. This allows a commit to say that a file was replaced
+    /// by a new incarnation of it, indicating a logically distinct file
+    /// taking the place of the previous file at the path.
+    pub salt: Vec<u8>,
```
yuja, line 222: Maybe this can be `enum salt|parents`, but `salt, parents` might be handy as low-level representation.

lib/src/backend.rs
```diff
@@ -438,6 +459,19 @@ pub trait Backend: Send + Sync + Debug {
...
+
+    /// Write the `CopyHistory` object and return its ID.
+    fn write_copy(&self, copy: &CopyHistory) -> BackendResult<CopyId>;
+
+    /// Find all copy histories that are related to the specified one. This is
+    /// defined as those that are ancestors of the given specified one, plus
+    /// their descendants. It is valid (but wasteful) to include other copy
+    /// histories, such as siblings, or even completely unrelated copy
+    /// histories.
+    fn get_related_copies(&self, copy_id: &CopyId) -> BackendResult<Vec<CopyHistory>>;
```
yuja, line 486: Do we have any implication about the order?

lib/src/conflicts.rs
```diff
@@ -188,6 +190,7 @@ pub struct MaterializedFileConflictValue {
     /// Merged executable bit. `None` if there are changes in both executable
     /// bit and file absence.
     pub executable: Option<bool>,
+    pub copy_id: CopyId,
```
yuja, line 193: Shouldn't this be `Option<CopyId>`? iirc, the idea was that `jj resolve` errors out if `executable` or `copy_id` can't be resolved.

lib/src/git_backend.rs
```diff
@@ -1022,6 +1024,18 @@ impl Backend for GitBackend {
         Ok(SymlinkId::new(oid.as_bytes().to_vec()))
     }
 
+    fn read_copy(&self, _id: &CopyId) -> BackendResult<CopyHistory> {
+        unimplemented!();
```
yuja, line 1028: nit: maybe better to return `BackendError::Unsupported` if we aren't going to implement these functions anytime soon.

lib/src/local_working_copy.rs
```diff
@@ -1513,14 +1514,36 @@ impl FileSnapshotter<'_> {
...
+            let copy_id = {
+                if let Some(TreeValue::File {
+                    id: _,
+                    executable: _,
+                    copy_id,
+                }) = current_tree_value
+                {
+                    copy_id.clone()
+                } else {
+                    CopyId::placeholder()
```
yuja, line 1538: I'm not sure if this `placeholder` is a temporary stub or a marker to generate new `salt`-ed `CopyHistory`. Maybe leave a TODO comment?

lib/src/merge.rs
```diff
@@ -569,25 +570,66 @@ where
             })
     }
 
+    /// If this merge contains only files or absent entries, and they all have
+    /// the same copy ID, returns a merge of the `FileId`s`. The executable
+    /// bits will be ignored. Use `Merge::with_new_file_ids()` to produce a
+    /// new merge with the original executable bits preserved.
+    pub fn to_file_merge(&self) -> Option<(Merge<Option<FileId>>, CopyId)> {
```
yuja, line 577: Shouldn't we ignore copy-id conflict consistently, and use `to_copy_merge()`?

lib/src/repo_path.rs
```diff
@@ -199,7 +199,7 @@ impl DoubleEndedIterator for RepoPathComponentsIter<'_> {
 impl FusedIterator for RepoPathComponentsIter<'_> {}
 
 /// Owned repository path.
-#[derive(Clone, Eq, Hash, PartialEq)]
+#[derive(ContentHash, Clone, Eq, Hash, PartialEq)]
```
yuja, line 202: nit: add `ContentHash` to `RepoPath` for consistency?

lib/testutils/src/test_backend.rs
```diff
@@ -249,6 +252,31 @@ impl Backend for TestBackend {
...
+        self.locked_data()
+            .copies
+            .insert(id.clone(), contents.clone());
+        Ok(id)
+    }
+
+    fn get_related_copies(&self, _copy_id: &CopyId) -> BackendResult<Vec<CopyHistory>> {
+        // Return all copy histories to test that the caller correctly ignores histories
+        // that are not relevant to the trees they're working with.
+        Ok(self.locked_data().copies.values().cloned().collect_vec())
```
yuja, line 277: nit: maybe better to ensure that `copy_id` exists?

yuja (commented):
Looks generally good, but @ilyagr might want to do more thorough review.

lib/src/local_working_copy.rs
martinvonz: True, this is one of few places where we will create a new copy id.

lib/src/backend.rs
```diff
@@ -60,6 +61,15 @@ impl ChangeId {
     }
 }
 
+impl CopyId {
+    ///  Returns a placeholder copy id to be used when we don't have a real copy
```
yuja, line 65: nit: remove extra leading space

lib/src/local_working_copy.rs
```diff
@@ -1513,14 +1514,43 @@ impl FileSnapshotter<'_> {
...
+                copy_id,
+            }))
         } else if let Some(old_file_ids) = current_tree_values.to_file_merge() {
+            // Safe to unwrap because the copy id exists exactly on the file variant
+            let copy_id_merge = current_tree_values.to_copy_id_merge().unwrap();
+            let copy_id = copy_id_merge
+                .resolve_trivial()
+                .cloned()
+                .flatten()
+                .unwrap_or_else(CopyId::placeholder);
```
yuja, line 1553: Just a comment. We'll need to handle unresolved case and resolved to `None` differently. There may be change-delete content conflict `A - B + 0` with unchanged copy history `X - X + 0`. Perhaps, `resolve_file_executable()` can be generic over `Merge<Option<T>>`.

This can be addressed later.

lib/testutils/src/test_backend.rs
```diff
@@ -249,6 +254,51 @@ impl Backend for TestBackend {
...
+        let mut emitted = HashSet::new();
+        while let Some((id, copy)) = work.pop_front() {
+            if copy.parents.iter().all(|id| emitted.contains(id)) {
+                histories.push(copy.clone());
+                emitted.insert(id);
+            } else {
+                work.push_back((id, copy));
+            }
+        }
+        histories.reverse();
```
yuja, line 298: nit: maybe better to use `topo_order_reverse()` to sort ids (or id/object pairs) first? Quadratic sorting wouldn't matter in tests, though.

lib/testutils/src/test_backend.rs
```diff
@@ -356,3 +406,46 @@ impl Backend for TestBackend {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+
+    use pollster::FutureExt as _;
```
yuja, line 413: nit: squash `as _` change in previous commit

lib/src/merge.rs
```diff
@@ -569,25 +570,51 @@ where
             })
     }
 
-    /// If this merge contains only files or absent entries, returns a merge of
-    /// the `FileId`s`. The executable bits will be ignored. Use
-    /// `Merge::with_new_file_ids()` to produce a new merge with the original
-    /// executable bits preserved.
+    /// If this merge contains only files or absent entries, and they all have
+    /// the same copy ID, returns a merge of the `FileId`s`. The executable
+    /// bits will be ignored. Use `Merge::with_new_file_ids()` to produce a
```
yuja, line 575: nit: stale doc. copy ids are ignored now.

yuja (approved):

lib/src/backend.rs
```diff
@@ -51,6 +51,7 @@ id_type!(pub TreeId { hex() });
 id_type!(pub FileId { hex() });
 id_type!(pub SymlinkId { hex() });
 id_type!(pub ConflictId { hex() });
+id_type!(pub CopyId { hex() });
```
ilyagr, line 54: Perhaps we should call this `CopyHistoryId`, to make the relationship with `CopyHistory` clear?

lib/src/backend.rs
```diff
@@ -438,6 +459,20 @@ pub trait Backend: Send + Sync + Debug {
 
     async fn write_symlink(&self, path: &RepoPath, target: &str) -> BackendResult<SymlinkId>;
 
+    /// Read the specified `CopyHistory` object.
+    async fn read_copy(&self, id: &CopyId) -> BackendResult<CopyHistory>;
```
ilyagr, line 481: We could mention that always returning `Err(Unsupported)` is OK.

lib/src/backend.rs
```diff
@@ -438,6 +459,20 @@ pub trait Backend: Send + Sync + Debug {
...
+
+    /// Write the `CopyHistory` object and return its ID.
+    async fn write_copy(&self, copy: &CopyHistory) -> BackendResult<CopyId>;
+
+    /// Find all copy histories that are related to the specified one. This is
+    /// defined as those that are ancestors of the given specified one, plus
+    /// their descendants. Children must be returned before parents.
```
ilyagr, line 491: We could explicitly add "Descendants of the specified copyid don't have to be included". Strictly speaking, the current text does not imply that is the specified copyid has an ancestor.

What about (other) ancestors of the descendants of ancestors? If they don't have to be included (I think they probably don't?), we could explicitly mention that as well, so that people understand that this is different than the undirected connected component.

lib/src/backend.rs
```diff
@@ -438,6 +459,20 @@ pub trait Backend: Send + Sync + Debug {
...
+    ///
+    /// It is valid (but wasteful) to include other copy histories, such as
+    /// siblings, or even completely unrelated copy histories.
+    async fn get_related_copies(&self, copy_id: &CopyId) -> BackendResult<Vec<CopyHistory>>;
```
ilyagr, line 498: (Doesn't really matter): Should this perhaps return CopyId-s? Perhaps this is a premature optimization.

lib/testutils/src/test_backend.rs
```diff
@@ -356,3 +406,46 @@ impl Backend for TestBackend {
...
+
+    fn copy_history(path: &str, parents: &[CopyId]) -> CopyHistory {
+        CopyHistory {
+            current_path: repo_path_buf(path),
+            parents: parents.to_vec(),
+            salt: vec![],
+        }
+    }
+
+    #[test]
```
ilyagr, line 421: (No action needed) I might be nice to later add some property tests that can be applied to any backend.

lib/testutils/src/test_backend.rs
```diff
@@ -249,6 +254,51 @@ impl Backend for TestBackend {
...
+                hash: copy_id.hex(),
+                source: "".into(),
+            });
+        }
+        // Return all copy histories to test that the caller correctly ignores histories
+        // that are not relevant to the trees they're working with.
+        let mut work: VecDeque<_> = copies.iter().collect();
+        let mut histories = vec![];
+        let mut emitted = HashSet::new();
+        while let Some((id, copy)) = work.pop_front() {
```
ilyagr: (Optional) Perhaps check for an infinite loop if there's a cycle in some test? E.g., check that the length of `emitted` increased after going through `copies.len()` items?

lib/src/backend.rs
```diff
@@ -438,6 +459,20 @@ pub trait Backend: Send + Sync + Debug {
...
+
+    /// Write the `CopyHistory` object and return its ID.
+    async fn write_copy(&self, copy: &CopyHistory) -> BackendResult<CopyId>;
+
+    /// Find all copy histories that are related to the specified one. This is
+    /// defined as those that are ancestors of the given specified one, plus
+    /// their descendants. Children must be returned before parents.
+    ///
+    /// It is valid (but wasteful) to include other copy histories, such as
+    /// siblings, or even completely unrelated copy histories.
```
ilyagr, line 494: (Optional) It's a little confusing to me what you have in mind here. It might be nice to remove the "(but wasteful)", mention that the caller will know to ignore unrelated copy histories (as opposed to creating overly complicated conflicts), and say how efficient or inefficient you expect that to be.

Allowing this doesn't really fit well in my understanding of how this will work, but that's probably OK. I was imagining that, at some point, the copy ids would be tracked by an index, similar to the commitid/changeid index, maybe with similar generation numbers. I'm not sure whether that's supposed to be part of the backend, or built by the caller of this function, or if you have something entirely different in mind.

But I'm sure you have something reasonable in mind, and that it can be changed later, so I don't think I have to understand this now.

lib/src/backend.rs
```diff
@@ -288,7 +297,11 @@ pub type BackendResult<T> = Result<T, BackendError>;
 pub enum TreeValue {
     // TODO: When there's a CopyId here, the copy object's path must match
     // the path identified by the tree.
-    File { id: FileId, executable: bool },
+    File {
+        id: FileId,
+        executable: bool,
+        copy_id: CopyId,
```
ilyagr, line 305: (Could be postponed) Should `Symlink` and `GitSubmodule` get a CopyId too? (Trees shouldn't, though).

Or we could maybe change this to:

```rust
pub enum TreeValue {
   Tree(TreeId),
   Leaf(LeafTreeValue, CopyId),
   Conflict(IAmObsoleteAndIDontKnowWhereToGo),
}
```

-----------

I also was wondering whether `CopyId` should be part of `TreeValue` at all. I was wondering whether some backends might want more flexibility than having a static CopyId for each tree entry will allow. I guess the plan is to treat the CopyId for the Git backend similarly to the change id: in the local `jj` repo, we treat it as though it was part of the commit id, and we generate it on fetch?

I was and maybe am a bit worried about this, but after some thought, I don't see any obvious blockers.

The other option would be to for a Tree to have a field for the copy ids of its entries that's ignored for the purpose of computing the TreeId. Then, you could generate a CopyId tree dynamically for the relevant commits for each operation. I guess you think we can make the Git backend work without that? I guess it'd be a last resort.

---------

For a different kind of Git backend, if invisibility to Git users was not a concern and `jj` repos only needed to be readable by Git, we could store the CopyHistory database for a commit and its ancestors inside each commit. When we have a rebase or merge operation involving commits A, B, and C, the database for each of these commits and their ancestors should be sufficient to do the rebase/merge.

But, at that point, maybe we'd be storing this data outside Git commits entirely.

lib/src/backend.rs
```diff
@@ -438,6 +459,20 @@ pub trait Backend: Send + Sync + Debug {
...
+    async fn write_copy(&self, copy: &CopyHistory) -> BackendResult<CopyId>;
+
+    /// Find all copy histories that are related to the specified one. This is
+    /// defined as those that are ancestors of the given specified one, plus
+    /// their descendants. Children must be returned before parents.
+    ///
+    /// It is valid (but wasteful) to include other copy histories, such as
+    /// siblings, or even completely unrelated copy histories.
+    async fn get_related_copies(&self, copy_id: &CopyId) -> BackendResult<Vec<CopyHistory>>;
+
```
ilyagr, line 499: (No action needed): As I mention in a later comment, I feel like it might be a useful optimization to pass `get_related_copies` a list of commits and allow it to restrict its output to the related copies that are relevant when rebasing between or merging these commits. So, no need to return the ancestors of the CopyId that are only relevant in commits that are older than the common ancestor of the given commit, for example.

But this is probably premature right now.

lib/src/backend.rs
martinvonz: Done

lib/src/backend.rs
ilyagr: This is not blocking, by the way. 

Thinking about it further, perhaps we could have NamingHistory and NamingId? If we use the word naming for this concept specifically, the relationship will be clearer.

If we stick to common words, I think I prefer RenameHistory/RenameId to CopyHistory. This graph tracks renames more naturally, copies are inferred from the same file being renamed to two different things. 

We can also keep thinking. I do think I'd prefer to rename this to *something*, though, if we can.

lib/src/backend.rs
martinvonz: Let's say we're rebasing a single commit to the new tip of the main branch. If the commit we're rebasing touched only 5 files, then we will ask for all copies related to those copy ids. We will then check which of those copy ids exist at the expected path (the path recorded in the `CopyHistory` object) in the target commit. The algorithm will still work even if the backend returns every copy id it's ever seen, it will just be slower.

lib/src/backend.rs
martinvonz: I didn't call it `CopyHistory` because I felt that it made it too long, especially the variable names (presumably `copy_history_id`). My preference is still to not rename it. Let me know if you think we should.

Regarding renaming it to something completely different, I like `Incarnation`. I think that name makes sense also for the first incarnation of a file (when it's not copied or renamed), and when creating a new incarnation of an existing file (similar to what `git diff` calls [`--break-rewrites`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-code--break-rewritesltngtltmgtcode)).

lib/src/backend.rs
martinvonz: We often need the copy objects so we can find the path at least.

lib/src/backend.rs
martinvonz: Yes, I agree that we may want to restrict the search. I also agree that we can figure it out later.

lib/testutils/src/test_backend.rs
martinvonz: Since it's content addressed here in the test backend, that's extremely unlikely to happen. And if it ever happens, it's only going to make tests hang, without impacting any real commands.

lib/testutils/src/test_backend.rs
martinvonz: It's simpler anyway, so I've switched to that function.

lib/testutils/src/test_backend.rs
martinvonz: Makes sense.

lib/src/backend.rs
martinvonz: The definition I wrote here is correct as far as I know at least. I think we want to include descendants of the specified copy id. That would happen if modified a file locally and it was renamed upstream and you then rebase your local commit onto the updated upstream.

lib/src/backend.rs
martinvonz: > (Could be postponed) Should `Symlink` and `GitSubmodule` get a CopyId too? (Trees shouldn't, though).

Yes, this is an open question in the design doc. I'm leaning against including a copy id for symlinks (and more strongly against it for the other types). We can probably add it later if necessary.

ilyagr: I'm a bit confused about this, why not add it for every leaf in the tree? Why does it matter what kind of a node it is? But it's not super-important.

For example, a symlink could be replaced with a copy of the file. I'd represent this by keeping the CopyId the same and changing the LeafTreeValue.

lib/src/backend.rs
ilyagr: My first feeling is that I'd prefer something like `Naming` or `Label` or `Rename` to "Incarnation". OTOH, I'd prefer Incarnation to Copy. Copy feels like it's a bit too common of a word in the VCS context to be used for this specialized thing.

In any case, I'm fine with not renaming it here, in the hope of picking something completely different in the future.

yuja: I don't think `CopyHistoryId` is right because I think the "history" is represented by a graph of copy objects. A single `CopyHistory` node represents a location/place at a certain point in time? I'd rather rename `struct CopyHistory` to something else, but `Copy` seemed too short and generic.

martinvonz: Yes, and `Copy` conflicts with `std::marker::Copy`.

lib/src/backend.rs
martinvonz: I'm not sure I agree with the example, but I don't mind adding a copy ID for non-File entries. I think Tree entries should probably not have copy IDs, however, but that's just because the consequences are unclear to me. Maybe it makes sense after I think more about it. In particular, it seems complicated what it should mean if `foo/` is copied to `bar/` but all the files remain where they were, and other such examples.

ilyagr: > I think Tree entries should probably not have copy IDs, however, but that's just because the consequences are unclear to me.

I agree, as long as `jj` is like Git in that it does not track empty directories. So, moving a directory is the same as moving all the files in it.

Hence the distinction between leaf nodes and non-leaf nodes that I made.

lib/src/backend.rs
ilyagr: That makes some sense, I guess it's also necessary for the definition of "related to" to be symmetric.

We should at some point add the motivation (that these correspond to tracking all the copies of the file) to the comment.

This is not immediately actionable, but I am still wondering about the fact that the definition is not transitive: if A is related to B, and B is related to C, this does not imply that C is related to A. (E.g. if C and A are parents of B).

Perhaps a better way to think of it is as a measure of closeness. Let's say that A and Z have 1 degree of separation if there is a directed path from A to Z. They have 2 degrees of separate if there is a B such that there is a directed path from A to B and from B to Z. For this function, you are picking out the vertices with at most 2 degrees of separation. I'm guessing that at some point, tracking the 1-degree-of-separation vertices and maybe 3-degrees-of-separation vertices will become relevant.

Or, at any point, we can bail out and just look at the undirected connected components ("finite degree of separation" in the lingo of the above paragraph), perhaps abandoning the graph altogether.

lib/src/backend.rs
ilyagr: > I don't think CopyHistoryId is right because I think the "history" is represented by a graph of copy objects. 

It could be `CopyNode`, from this perspective. But I think something entirely different might be better.

(I un-resolved this and a bunch of other conversations that I don't want to forget about. None of them were blocking, though).

PR #6659, opened 2025-06-01
revset_engine: optimize heads of filtered revset

### LINKED ISSUE #6656: Pathological performance for `heads(::@ & ~empty())`
## Description

When I run this in a local checkout of:

```
time jj log --revisions "latest(heads(::@ & ~empty()))"
```

… it takes 1.65 seconds. Since I run this calculation in my shell prompt under the assumption that it's super fast, it's causing my shell prompt to hang for about 5 seconds.

Unfortunately, I don't have a repro. It doesn't happen in a fresh checkout (where it runs in about 0.02 seconds there).

I can't figure out anything in my checkout that would cause this. This is not an especially large repo, and the calculation should not is not substantially different from the checkout where it takes 0.02 seconds. This is what my tree looks like in `gg.app`:

Image

Note that *all commits are non-empty*, including all four that are missing a description.

## Expected Behavior

It should run in about 0.02 seconds.

## Actual Behavior

Takes 1.65 seconds.

## Specifications

- Platform: macOS 15.5
- Version: 0.29.0

scott2000:
Resolves #6656. On my machine, this reduces execution time from ~115ms to ~25ms for revsets like `latest(heads(~empty()))` in the `jj` repo.

# Checklist

If applicable:


lib/src/default_index/composite.rs
scott2000, line 1: I'm not sure whether I need to add any tests or run any benchmarks for this type of change.

lib/src/default_index/composite.rs
```diff
@@ -381,44 +379,40 @@ impl CompositeIndex {
...
-                dedup_pop(&mut work).unwrap();
-                continue;
-            }
-            for parent_entry in parent_entries {
-                assert!(parent_entry.position() < cur_pos);
-                work.push(IndexPositionByGeneration::from(&parent_entry));
+    /// Returns a function which can be used to find the heads of a set of index
+    /// positions. The index positions must be iterated in reverse sorted order.
+    pub fn heads_pos_predicate<'a>(&'a self) -> impl FnMut(&IndexPosition) -> bool + use<'a> {
+        let mut parents = BinaryHeap::new();
```
scott2000, line 407: Instead of having the priority queue be based on generation number, I changed it to be based on index position (since using index position means it's not necessary to know the full set of candidate commits ahead of time). I'm not sure if there's any efficiency issues with using the index position though.

martinvonz: There's a `jj bench revset` command (required the `bench` crate feature) and a `bench-revsets-git.txt` file you can pass to it.

lib/src/default_index/composite.rs
yuja: I think we can at least test the tree transformation if we move the rewriting rule to the frontend function `revset::optimize()`. The rule might be something like `heads(ancestors(x) & y) -> heads_ancestors(x, y)`.

lib/src/default_index/composite.rs
yuja: I think we can just cut off uninteresting parents based on generation number. I have unsent patch for that.

lib/src/default_index/revset_engine.rs
```diff
@@ -934,9 +933,11 @@ impl EvaluationContext<'_> {
...
                 let candidate_set = self.evaluate(candidates)?;
-                let head_positions: BTreeSet<_> =
-                    index.heads_pos(candidate_set.positions().attach(index).try_collect()?);
-                let positions = head_positions.into_iter().rev().collect();
+                let positions = candidate_set
+                    .positions()
+                    .attach(index)
+                    .filter_ok(index.heads_pos_predicate())
+                    .try_collect()?;
                 Ok(Box::new(EagerRevset { positions }))
```
yuja, line 961: If the goal is to make it evaluate lazily, we can use `RevWalkRevset` or `FilterRevset`.

lib/src/default_index/composite.rs
scott2000: Ok, that makes sense. I was actually attempting to do a slightly different optimization though, more like `heads(x & filter) -> heads_with_filter(x, filter)`. Instead of evaluating the filter on every candidate commit, it only would evaluate the filter on commits which aren't already known to not be heads (since they've already been found to be an ancestor of a previous filtered head). But I think the same suggestion would apply for this as well.

yuja: > more like `heads(x & filter) -> heads_with_filter(x, filter)`. Instead of evaluating the filter on every candidate commit, it only would evaluate the filter on commits which aren't already known to not be heads

Doesn't it mean we have to traverse `x` to the end? `filter` could return `true|false|no_more`, but we don't have such machinery right now.

lib/src/default_index/composite.rs
yuja: > cut off uninteresting parents based on generation number.

https://github.com/jj-vcs/jj/pull/6667

lib/src/default_index/composite.rs
scott2000: Yeah, we do have to traverse `x` to the end, but we don't actually have to call `filter` on every commit in `x`. For instance, given this graph for `x`:

```
A <- B <- C
  \- D <- E
```

If the filter matches A, C, and D, then we can call the filter only 3 times instead of 5 times (we only need to call the filter on C, E, and D).

Assuming the reverse index position order is `C E D B A`, then the steps would look like this:

1. `parents=[] heads=[] candidate=C` - C is a head, so we need to call filter on it. Since the filter returns true, we save it as a head.
1. `parents=[B] heads=[C] candidate=E` - E is a head, so we need to call filter on it. Since the filter returns false, we skip it.
1. `parents=[B] heads=[C] candidate=D` - D is a head, so we need to call filter on it. Since the filter returns true, we save it as a head.
1. `parents=[B, A] heads=[C, D] candidate=B` - B is not a head, since it is an ancestor of an already-found head. We don't need to call the filter.
1. `parents=[A] heads=[C, D] candidate=A` - A is not a head, since it is an ancestor of an already-found head. We don't need to call the filter.
1. `heads=[C, D]`

So we still need to traverse every commit in `x`, but we only need to call the filter on `heads(x & filter)::`, which is potentially much smaller. Since the filter predicate is often significantly slower than scanning using the index, this actually saves a huge amount of time if `x` is large it seems like.

Edit: I only figured this out with iterating over index positions rather than generation numbers, but I think it would also be possible with the generation numbers with a bit of refactoring. I'll update the PR.

yuja: > we don't actually have to call `filter` on every commit in `x`.

True, but the input set given to `heads()` is usually small. If the set is large enough to skip expensive functions, it's likely to have some form of `ancestors(x)`?

> I only figured this out with iterating over index positions rather than generation numbers, but maybe it would also be possible with the generation numbers?

If the input set has arbitrary graph shape, I don't think it would make sense to try to omit uninteresting ancestors without fully consuming the input.

scott2000: Yeah, that makes sense. I think the only reason we can't do lazy evaluation for `heads()` right now is because it needs to find the minimum generation number. So if we wanted to make it be lazy, it would definitely make sense to restrict it to `ancestors(x)`, since then we would already know the minimum generation number is 0 ahead of time, since `ancestors(x)` would go back to `root()` anyways.

But even with eager evaluation, I was able to get the `heads(author(peff))` benchmark to run 20x faster, so I'm thinking lazy evaluation probably isn't worth the effort to implement. In that case, I don't think it's necessary to restrict the filtering optimization to just `ancestors(x)`, since skipping unnecessary predicate evaluations should save time regardless of graph shape.

lib/src/default_index/revset_engine.rs
scott2000: I think I'm giving up on trying to make it evaluate lazily, because it seems unlikely that it would be useful in most cases. The main speedup is coming from skipping unnecessary filtering.

lib/src/default_index/composite.rs
scott2000: @yuja I kept most of the logic I had the same as before (it still iterates by index position), but I added the minimum generation number cutoff (similar to your PR).

Could you take a look and see if this approach makes sense? I think I'm seeing a ~10% improvement in performance on `heads(tags())`, and ~45% improvement in performance on `heads(::v2.40.0)` compared to your PR.

lib/src/default_index/composite.rs
yuja: > I think the only reason we can't do lazy evaluation for `heads()` right now is because it needs to find the minimum generation number.

I think `heads(x)` can be lazy by determining `min_generation` lazily, but I don't know if it would be worth the complexity. Lazy version might be slow if the heads set is small.

```
while next = select_greater(candidates, ancestors) {
    if next is candidate {
        min_generation = min(next, min_generation);
        ancestors.extend(next.parents);
        ...
    } else if candidates.is_empty() && next <= min_generation {
        continue;
    }
    ...
}
```

> I was able to get the `heads(author(peff))` benchmark to run 20x faster,

Maybe you know, but `heads(author(peff))` is evaluated as `heads(::visible_heads() & author(peff))` internally, so it can be caught by the subtitution rule `heads(::x & y) -> heads_ancestors(x, y)`.

scott2000: Yeah, I had figured it was something like that; I'm just not sure what the benefit of restricting it to `ancestors(x)` would be. It seems like it might be useful to have `heads(immutable_heads().. ~ empty())` for instance, and it would be nice to avoid the unnecessary predicate evaluations there as well (since I believe they could both take advantage of the same algorithm).

yuja: > I'm just not sure what the benefit of restricting it to `ancestors(x)` would be.

My feeling is that it's silly to traverse ~100k ancestor entries to answer queries like `heads(::x ...)`. It's relatively fast, though.

> `heads(immutable_heads().. ~ empty())` for instance,

```
heads(immutable_heads().. ~ empty())
= heads(immutable_heads()..visible_heads() ~ empty())
= heads(::visible_heads() ~ ::immutable_heads() ~ empty())
= heads_ancestors(visible_heads(), ~ ::immutable_heads() ~ empty())
```

I'm not saying all possible large `heads(x)` can be caught, but I assume many of them can be transformed in that way.

scott2000: Hmm yeah, I see. Let me take a shot at implementing something like that and see if I can figure it out.

scott2000:
Closing in favor of #6679.

PR #6661, opened 2025-06-01
split: write the commits only once

glehmann:
# Checklist

If applicable:


yuja (commented):
I haven't closely reviewed the implementation yet, but my gut feeling is that the added functions need extensive tests since they are no longer wrapper functions of `move_commits()`. I think they can be moved to jj-lib and tested there. BTW, it'll be nice if we don't have to add many variants of `move_commits()`.

Can you also add tests that should be fixed by this change? Maybe this is a workaround for the `A-B+A=A` merge problem?

https://github.com/jj-vcs/jj/blob/92cd17c095d6c6725604f27b9b9ac30f3cb3f28f/lib/src/rewrite.rs#L1163-L1182

glehmann:
> BTW, it'll be nice if we don't have to add many variants of `move_commits()`.

I sure can understand that. I think that's a compromise between complicating `move_commits()`, which is not simple already, and avoiding duplicating code.

You mentioned altering the commit predecessors in #6458 as a possible way to solve the problem, but I didn't find how to do that without rewriting the commits, and @martinvonz was more in favor of avoiding the multiple writes of the commits if possible — that's possible, but that comes with a significant complexity.

We could also consider that the multiple entries in the evolog for some of the commits modified by `jj split` when using the `-A/-B/-d` options are fine, and just keep it that way.

> Can you also add tests that should be fixed by this change? Maybe this is a workaround for the `A-B+A=A` merge problem?

Sure, I should have done that in the first place.

I don't think that's the same problem here: I'm just trying to avoid multiple entries in the evolog of the commits for a single `jj split` command, and maybe getting a better performance when rebasing many descendants? (I haven't done any timing)

glehmann:
I've added a test showing the multiple entries in the evolog before the changes in this PR

yuja:
> You mentioned altering the commit predecessors in #6458 as a possible way to solve the problem, but I didn't find how to do that without rewriting the commits,

I just meant that can be done *with* rewriting commits twice.

> and @martinvonz was more in favor of avoiding the multiple writes of the commits if possible

I agree in general, but as you said, it comes with a complexity, and I feel it's huge.

glehmann:
As someone who believes in making decisions by experimenting and discarding what doesn't meet expectations, I have no issue doing so with this proposed change. Solving that problem was quite interesting, and I learned a lot about JJ's internals.
It's not for me to make that decision, though, so I would appreciate some guidance on what to do next about the multiple evolog entries.

yuja:
> some guidance on what to do next about the multiple evolog entries.

@martinvonz any thoughts?

My current opinions are:

* The added complexity is too high compared to the functionality brought by these functions. We'll also need more tests.
* It's nice if we can somehow refactor `move_commits()` in a way that `jj split`/`squash`/etc. can rewrite/rebase commits without creating transitive predecessors.
* Transitive entries in evolution log aren't great, but seem okay. If needed, maybe we can flatten them when displaying evolog?

Thanks for the PR, btw.

martinvonz:
Sorry, I missed the question here for a while. I've been very busy with internal work lately. I think I'll need some time to understand what this PR does to form an opinion. I'm just going on vacation now. We'll see if I find more time to look at it during my vacation (I'm not very optimistic).

yuja:
> maybe we can flatten them when displaying evolog?

Another idea is to add flag to omit rebase entry from `evolog`. Computing emptiness of evolog diff would be expensive, but if we had an `is_rebase=true` flag in predecessors metadata, it should be easy.

martinvonz:
> > some guidance on what to do next about the multiple evolog entries.
> 
> @martinvonz any thoughts?
> 
> My current opinions are:
> 
> * The added complexity is too high compared to the functionality brought by these functions. We'll also need more tests.

I think I agree with that.

> * It's nice if we can somehow refactor `move_commits()` in a way that `jj split`/`squash`/etc. can rewrite/rebase commits without creating transitive predecessors.

Makes sense, but I haven't even thought about what that would look like.

> * Transitive entries in evolution log aren't great, but seem okay. If needed, maybe we can flatten them when displaying evolog?

For the entries written to the operation object, we can simply intermediate commits, right? So if X was rewritten to Y and then again to Z, we'd record just that X was rewritten to Z. There's still the issue that we write commit objects unnecessarily but that doesn't seem so bad. We probably write many more tree objects unnecessarily anyway.

> 
> Thanks for the PR, btw.

+1

yuja:
> For the entries written to the operation object, we can simply intermediate commits, right? So if X was rewritten to Y and then again to Z, we'd record just that X was rewritten to Z. There's still the issue that we write commit objects unnecessarily but that doesn't seem so bad. We probably write many more tree objects unnecessarily anyway.

Yes. Simplification can be done either before writing a transaction or when displaying evolution log. I didn't implement the former because of the (minor) issue you pointed out.

glehmann:
I keep this PR open just to complete the discussion.
I'll close it once I better understand the way to go :-)

> > For the entries written to the operation object, we can simply intermediate commits, right? So if X was rewritten to Y and then again to Z, we'd record just that X was rewritten to Z. There's still the issue that we write commit objects unnecessarily but that doesn't seem so bad. We probably write many more tree objects unnecessarily anyway.
> 
> Yes. Simplification can be done either before writing a transaction or when displaying evolution log. I didn't implement the former because of the (minor) issue you pointed out.

For the former, are you thinking of an automatic simplification that would be done when writing the transaction, or of something more explicit, as done in #6725?

yuja:
> > Yes. Simplification can be done either before writing a transaction or when displaying evolution log. I didn't implement the former because of the (minor) issue you pointed out.
> 
> For the former, are you thinking of an automatic simplification that would be done when writing the transaction, or of something more explicit, as done in #6725?

An automatic simplification. Basically, we'll insert `resolve_transitive_edges()` somewhere in `Transaction::write()`.

https://github.com/jj-vcs/jj/pull/6757/commits/52d0bc7110c191dbbf62f0cd5ec8130efb4fcbb8#diff-501dd7bfb50e99e8ae26a10fe9c8a7799a30526f3766f2edad2f8c7c9a3032fc

glehmann:
We'll have a better option based on #6757

Issue #2957, opened 2024-02-05
FR: Better Identity Management


steveklabnik:
Before we get into the details, I want to emphasize that I see this FR as more of a "consider and investigate" rather than any expectation that things will change. Identity is a difficult and complex topic, and the current implementation is serviceable. But, I also think, in the same way that `jj` improves upon previous DVCSes in many ways, maybe it could for identity as well.

**Is your feature request related to a problem? Please describe.**

In Git, as with many VCS systems, commits have some sort of identity attached to them. In Git specifically, commit objects contain author information, a name and an email. Because this information is part of the commit itself, changing this information will change the SHA of the commit, which means you need a lot of rebasing to update the tree.

In the real world, identity is not so simple. See articles like [Falsehoods Programmers Believe about Names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) for some examples. But even beyond many of these (which to be clear, git does not fall afoul of all of them), sometimes things happen where [people would like to revoke their authorship](https://github.com/rust-lang/team/pull/1071). Note that while this got merged, it doesn't do what the author would like it to do:

> I would also like to request that the Rust team strip all authorship information from my commits to the projects in the rust-lang organization.

Due to the way that git's model works, this is effectively an impossible request. (Which I find personally unfortunate but I am trying to stay neutral in the description of this issue.)

I link to this issue because I think it's a good proxy for thinking about this problem in general, revocation is one of the more difficult aspects of identity management.

**Describe the solution you'd like**

At this time, I am not asking for a specific solution, but would like to have this FR as a place to discuss possibilities and collect information about how to possibly support these sorts of operations in the future.

**Describe alternatives you've considered**

I think the first step towards evolution here is to decide what properties a given identity solution would require, and then go from there. The two that I can think of are rotation (changing an identity) and revocation (abandoning an identity), but I am sure there are more. Weighing these various factors will point towards a specific technical solution.

There is also the additional complexity of how this affects various backends, as you won't necessarily be able to represent the full solution when sending things to an upstream git repository, for example.

[...]

PR #6662, opened 2025-06-01
squash: use the authors of the squashed commit as co-authors

glehmann:
# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * Unquoted `*` is now allowed in revsets. `bookmarks(glob:foo*)` no longer
   needs quoting.
 
+* `jj squash` stores the co-authors of the squashed commits in the destination
+  commit when using the `--co-authors` command line option or the
+  `squash.co-authors = true` configuration.
+
 ### Fixed bugs
 
 * `jj fix` now prints a warning if a tool failed to run on a file.
--- a/cli/src/commands/commit.rs
+++ b/cli/src/commands/commit.rs
@@ -25,7 +25,7 @@ use crate::cli_util::CommandHelper;
 use crate::command_error::CommandError;
 use crate::command_error::user_error;
 use crate::complete;
-use crate::description_util::add_trailers;
+use crate::description_util::add_config_trailers;
 use crate::description_util::description_template;
 use crate::description_util::edit_description;
 use crate::description_util::join_message_paragraphs;
@@ -182,11 +182,11 @@ new working-copy commit.
             // can be discarded as soon as it's no longer the working copy. Adding a
             // trailer to an empty description would break that logic.
             commit_builder.set_description(description);
-            description = add_trailers(ui, &tx, &commit_builder)?;
+            description = add_config_trailers(ui, &tx, &commit_builder)?;
         }
         description
     } else {
-        add_trailers(ui, &tx, &commit_builder)?
+        add_config_trailers(ui, &tx, &commit_builder)?
     };
     let description = if args.message_paragraphs.is_empty() || args.editor {
         commit_builder.set_description(description);
--- a/cli/src/commands/describe.rs
+++ b/cli/src/commands/describe.rs
@@ -228,7 +228,8 @@ pub(crate) fn cmd_describe(
             // trailer to an empty description would break that logic.
             if use_editor || !commit_builder.description().is_empty() {
                 let temp_commit = commit_builder.write_hidden()?;
-                let new_description = add_trailers_with_template(&trailer_template, &temp_commit)?;
+                let new_description =
+                    add_trailers_with_template(&trailer_template, &temp_commit, &[])?;
                 commit_builder.set_description(new_description);
             }
         }
--- a/cli/src/commands/new.rs
+++ b/cli/src/commands/new.rs
@@ -29,7 +29,7 @@ use crate::cli_util::RevisionArg;
 use crate::cli_util::compute_commit_location;
 use crate::command_error::CommandError;
 use crate::complete;
-use crate::description_util::add_trailers;
+use crate::description_util::add_config_trailers;
 use crate::description_util::join_message_paragraphs;
 use crate::ui::Ui;
 
@@ -199,7 +199,7 @@ pub(crate) fn cmd_new(
         // can be discarded as soon as it's no longer the working copy. Adding a
         // trailer to an empty description would break that logic.
         commit_builder.set_description(description);
-        description = add_trailers(ui, &tx, &commit_builder)?;
+        description = add_config_trailers(ui, &tx, &commit_builder)?;
     }
     commit_builder.set_description(&description);
     let new_commit = commit_builder.write(tx.repo_mut())?;
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -42,7 +42,7 @@ use crate::cli_util::print_unmatched_explicit_paths;
 use crate::command_error::CommandError;
 use crate::command_error::user_error_with_hint;
 use crate::complete;
-use crate::description_util::add_trailers;
+use crate::description_util::add_config_trailers;
 use crate::description_util::description_template;
 use crate::description_util::edit_description;
 use crate::description_util::join_message_paragraphs;
@@ -262,7 +262,7 @@ pub(crate) fn cmd_split(
         };
         let description = if !description.is_empty() || args.editor {
             commit_builder.set_description(description);
-            add_trailers(ui, &tx, &commit_builder)?
+            add_config_trailers(ui, &tx, &commit_builder)?
         } else {
             description
         };
@@ -317,7 +317,7 @@ pub(crate) fn cmd_split(
             commit_builder.description().to_owned()
         };
         let description = if show_editor {
-            let new_description = add_trailers(ui, &tx, &commit_builder)?;
+            let new_description = add_config_trailers(ui, &tx, &commit_builder)?;
             commit_builder.set_description(new_description);
             let temp_commit = commit_builder.write_hidden()?;
             let intro = "Enter a description for the remaining changes.";
--- a/cli/src/commands/squash.rs
+++ b/cli/src/commands/squash.rs
@@ -28,6 +28,7 @@ use jj_lib::repo::Repo as _;
 use jj_lib::rewrite;
 use jj_lib::rewrite::CommitWithSelection;
 use jj_lib::rewrite::merge_commit_trees;
+use jj_lib::trailer::Trailer;
 use pollster::FutureExt as _;
 use tracing::instrument;
 
@@ -41,7 +42,7 @@ use crate::command_error::CommandError;
 use crate::command_error::user_error;
 use crate::command_error::user_error_with_hint;
 use crate::complete;
-use crate::description_util::add_trailers;
+use crate::description_util::add_config_trailers_with_extras;
 use crate::description_util::combine_messages_for_editing;
 use crate::description_util::description_template;
 use crate::description_util::edit_description;
@@ -176,6 +177,16 @@ pub(crate) struct SquashArgs {
         add = ArgValueCandidates::new(complete::diff_editors),
     )]
     tool: Option<String>,
+
+    /// Set the authors of the squashed commits as co-authors of the
+    /// destination commit
+    #[arg(long, value_enum, group = "co-authors")]
+    co_authors: bool,
+
+    /// Don't set the authors of the squashed commits as co-authors of the
+    /// destination commit
+    #[arg(long, value_enum, group = "co-authors")]
+    no_co_authors: bool,
 
     /// Move only changes to these paths (instead of all paths)
     #[arg(
@@ -332,12 +343,36 @@ pub(crate) fn cmd_squash(
         source_commits.iter().map(|commit| &commit.selected_tree),
     )?;
 
+    let use_co_authors_trailers = if args.co_authors {
+        true
+    } else if args.no_co_authors {
+        false
+    } else {
+        tx.settings().get_bool("squash.co-authors")?
+    };
     if let Some(squashed) = rewrite::squash_commits(
         tx.repo_mut(),
         &source_commits,
         &destination,
         args.keep_emptied,
     )? {
+        let co_author_trailers: Vec<_> = if use_co_authors_trailers {
+            squashed
+                .abandoned_commits
+                .iter()
+                .filter(|commit| {
+                    commit.author().name != destination.author().name
+                        || commit.author().email != destination.author().email
+                })
+                .map(|commit| Trailer {
+                    key: "Co-authored-by".to_string(),
+                    value: format!("{} <{}>", commit.author().name, commit.author().email),
+                })
+                .unique()
+                .collect()
+        } else {
+            vec![]
+        };
         let mut commit_builder = squashed.commit_builder.detach();
         let single_description = match squashed_description {
             SquashedDescription::Exact(description) => Some(description),
@@ -352,7 +387,8 @@ pub(crate) fn cmd_squash(
                 description
             } else {
                 commit_builder.set_description(&description);
-                let description_with_trailers = add_trailers(ui, &tx, &commit_builder)?;
+                let description_with_trailers =
+                    add_config_trailers_with_extras(ui, &tx, &commit_builder, &co_author_trailers)?;
                 if args.editor {
                     commit_builder.set_description(&description_with_trailers);
                     let temp_commit = commit_builder.write_hidden()?;
@@ -372,6 +408,7 @@ pub(crate) fn cmd_squash(
                 abandoned_commits,
                 (!insert_destination_commit).then_some(&destination),
                 &commit_builder,
+                &co_author_trailers,
             )?;
             // It's weird that commit.description() contains "JJ: " lines, but works.
             commit_builder.set_description(combined);
--- a/cli/src/config-schema.json
+++ b/cli/src/config-schema.json
@@ -836,6 +836,17 @@
                 }
             }
         },
+        "squash": {
+            "type": "object",
+            "description": "Settings for jj squash",
+            "properties": {
+                "co-authors": {
+                    "type": "boolean",
+                    "description": "If true, the co-authors will be collected from the source commits.",
+                    "default": false
+                }
+            }
+        },
         "hints": {
             "type": "object",
             "description": "Various hints in jj's UI that can be disabled",
--- a/cli/src/config/misc.toml
+++ b/cli/src/config/misc.toml
@@ -57,3 +57,6 @@ auto-update-stale = false
 # in the future.
 [split]
 legacy-bookmark-behavior = true
+
+[squash]
+co-authors = false
--- a/cli/src/description_util.rs
+++ b/cli/src/description_util.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::fs;
 use std::io;
 use std::io::Write as _;
@@ -18,6 +19,7 @@ use jj_lib::config::ConfigGetError;
 use jj_lib::file_util::IoResultExt as _;
 use jj_lib::file_util::PathError;
 use jj_lib::settings::UserSettings;
+use jj_lib::trailer::Trailer;
 use jj_lib::trailer::parse_description_trailers;
 use jj_lib::trailer::parse_trailers;
 use thiserror::Error;
@@ -325,6 +327,7 @@ pub fn combine_messages_for_editing(
     sources: &[Commit],
     destination: Option<&Commit>,
     commit_builder: &DetachedCommitBuilder,
+    co_author_trailers: &[Trailer],
 ) -> Result<String, CommandError> {
     let mut combined = String::new();
     if let Some(destination) = destination {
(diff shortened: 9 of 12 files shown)


glehmann:
@shanesveller  I'm continuing the discussion from https://github.com/jj-vcs/jj/pull/6036#issuecomment-2919688593 here, to not pollute the other discussion.

> > Also if you base the work off others please attribute the squashed commits to _all authors_ who contributed with the `Co-authored-by:` trailer whenever you're done.
> 
> > It might be interesting to have a flag to do that automatically in `jj squash`
> 
> The `Co-authored-by` trailer is a strong convention (and a very socially healthy practice), but it's not a formal standard that is universal to Git and honored by all tools and forges. It's a useful hack.

That's indeed a useful hack. More generally, trailers look like a hack for the lack of a better place to add metadata.
But `Co-Authored-By` is [documented in git](https://git-scm.com/docs/SubmittingPatches), and supported by most forges.

> I imagine `jj`'s eventual native backend or other non-Git backends would prefer to model that sort of info as something other than a message trailer. With that in mind, as an end-user I'm a little ambivalent towards `jj` taking automatic steps to use or populate `Co-authored-by` because it helps ossify the use of trailers and that one in particular.

Nothing prevents us to populate the co-authors in an other way with a backend that supports a better option than a trailer.

>  Opt-in CLI flag seems like the right model if it gets included at all.

My preference would go to an opt-out cli flag:
* it would help properly credit the authors
* dropping the co-authors would be much easier after the squash, if we wanted to ignore them but forgot to use the flag, than to recollect the data and add it to the destination commit.

yuja:
Can't we achieve this by user templates? (We can add `squash_description` or `squash_trailers` template if needed.)

Since trailers are just convention of Git-based projects, I don't think these templates should be enabled by default.

> This is using the trailers to store the co-author data, but could be
> replaced by another mechanism for a backend that supports multiple
> authors for example.

Good point. If our internal model supported multiple authors, `jj squash` might copy them from the source commits. I assume that would be processed differently than trailers, though.

shanesveller:
> But `Co-Authored-By` is [documented in git](https://git-scm.com/docs/SubmittingPatches), and supported by most forges.

At the top of that page is the text:
> This information is specific to the Git project

Which is to say, they've mentioned their internal use of the trailer **for contributing to Git itself**, but it's not documented broadly (AFAIK it's in no manpages for example, and that's probably the only spot on git-scm.com) nor does this page suffice to make it a standard of Git practices for Git end-users. 

> Nothing prevents us to populate the co-authors in an other way with a backend that supports a better option than a trailer.

Well, presumably the common data model would not be changed to distinguish concerns that should be stored in one data format for some backends, and mutually-exclusively as another format for other backends, because that would interfere with portability between backends. So most likely a principled implementation of that kind of sophisticated support would lead to the information being duplicated. Once as-is in the message trailer and once again in whatever KV metadata store a backend has implemented, probably just because it would be faster to lookup from a proper index than parsing it back out of a lossy string kludge.

I don't have anything else to add other than to reaffirm that I think opt-in is more procedurally correct than opt-out given the above hair-splitting. I'll let the PR proceed without further interjections on my part.

martinvonz:
> If our internal model supported multiple authors

That was my thinking as well. We may want to think a bit more about #2957 (user identity management) if we're going to change the model anyway.

glehmann:
> Can't we achieve this by user templates? (We can add `squash_description` or `squash_trailers` template if needed.)
> 
> Since trailers are just convention of Git-based projects, I don't think these templates should be enabled by default.

My original thought process was inverted: from a semantic standpoint, we have the option to preserve or omit co-author data, which would then be stored via the backend's available mechanisms.
This is achieved with the trailers today because that's git's mechanism, but could be done with any kind of metadata supported by the backend.

glehmann:
This proves more useful to me than expected. I've been using this PR for 2 weeks, and this is the third time I've seen a `Co-authored-by` header being created when I wouldn't have thought to create it myself.
I hope we can make that feature happen with one form or another.

emilazy:
FWIW I think this is a good idea and while Jujutsu should natively support multi-author commits, trailers are clearly the right way to represent them in the Git backend and have wide tooling support.

yuja:
It might be okay for the Git backend to derive multiple authors from the commit description when reading/importing (just like we generate change id from the commit id header), but I don't think the backend should implicitly insert trailers when writing commit object. The description content should be fully controlled by user. (Maybe the idea was to append trailers by frontend/CLI if the backend doesn't support multiple authorship, but just to be clear.)

I'm not against adding this feature in some form, but I think it should be opt-in (for backends which don't have native support for multiple authors.)

glehmann:
> It might be okay for the Git backend to derive multiple authors from the commit description when reading/importing (just like we generate change id from the commit id header), but I don't think the backend should implicitly insert trailers when writing commit object. The description content should be fully controlled by user. (Maybe the idea was to append trailers by frontend/CLI if the backend doesn't support multiple authorship, but just to be clear.)

When you say "The description content should be fully controlled by user." do you mean with a template?

The implementation in this PR takes care to put the co-authors in the description before we open the editor, so the user has full control over the description content. There is also the command-line flag to ignore the co-authors.

> I'm not against adding this feature in some form, but I think it should be opt-in (for backends which don't have native support for multiple authors.)

I was hoping that we could check if the backend supports multiple authors in `squash.rs`, and completely bypass the trailer part if it does.
Is it currently possible to have multiple authors in some backends and to check that capability in the code?

PhilipMetzger:
> Is it currently possible to have multiple authors in some backends and to check that capability in the code?

AFAIK, there's no such backend yet and with it no possibility to add a simple check for it. 

> The implementation in this PR takes care to put the co-authors in the description before we open the editor, so the user has full control over the description content. There is also the command-line flag to ignore the co-authors.

I think Yuya means that the flag should be opt-in instead of opt-out since that allows people to configure the templates themselves.

yuja:
> When you say "The description content should be fully controlled by user." do you mean with a template?

No. I just mean the description stored in Git commit shouldn't be mechanically transformed as structured data. The backend shouldn't apply preprocessing rule like `extract_authors(stored_description) -> (authors, description_without_trailers)` when reading a commit, therefore, the inverse `append_authors_trailers(edited_description, authors) -> description_to_store` seems also bad.

> Is it currently possible to have multiple authors in some backends and to check that capability in the code?

No. So I prefer template-based solution (or "workaround") with the current data model.

glehmann:
For an implementation using a template, it would probably be better to have access to the co-authors or the squashed commits in the `commit_trailers` template. That way, we could define the template as something like:

```toml
[templates]
commit_trailers = """
format_signed_off_by_trailer(self) ++
concat(
  co_authors
    .filter(|co_author| !trailers.contains('Co-authored-by', co_author)
    .map(|co_author| 'Co-authored-by: ' ++ co_author ++ '\n')
)
"""
```

That seems quite usable from a user's point of view.
I don't know how much sense it makes for the implementation, as there is no notion of co-author or multi-author on a commit.
Any better idea for a template interface is of course welcome.

PhilipMetzger:
I think an idea could be providing a `description_context` type to the `commit_trailers` template, which contains the information about the authors of the squashed commits and whether it should be used. It makes the implementation contextual and complex though. 

e.g 
```toml
[templates]
'co_authored_by(author)' = "'Co-authored-by: ' ++ author ++ '\n'"
commit_trailers = """
format_signed_off_by_trailer(self) ++
if (description_context.in_squash,
concat(
  description_context.authors
    .filter(|author| !trailers.contains(co_authored_by(author))
    .map(|author| co_auhored_by(author))
   
) ++ ' # from' ++ description_context.commits.count().as_string() ++ ' squashed commits', ++ '\n')
"""

```

martinvonz:
> I just mean the description stored in Git commit shouldn't be mechanically transformed as structured data.

Why is that bad? Is it because it's not enforced by Git, so there might be a typo on `Co-authored-by`? (Maybe I even spelled it wrong now.)

yuja:
> > I just mean the description stored in Git commit shouldn't be mechanically transformed as structured data.
> 
> Why is that bad? Is it because it's not enforced by Git, so there might be a typo on `Co-authored-by`?

Because I (as a user) wouldn't expect `description` is processed differently than Git. It would be weird if Gerrit's `Change-Id` trailer were stripped/added automatically by backend, for example.

> For an implementation using a template

I thought `squash_trailers` template could be of `List<Commit>` type, but we'll have to somehow remove the "primary" author from trailers, so `List<Commit>` might not go well. It might be easier to add special case or config knob for `Co-authored-by` trailers.

glehmann:
> I think an idea could be providing a `description_context` type to the `commit_trailers` template, which contains the information about the authors of the squashed commits and whether it should be used. It makes the implementation contextual and complex though.

Looks good to me. Do we already have an example of such an approach? I can't find something similar in the code.

yuja:
> > I think an idea could be providing a `description_context` type to the `commit_trailers` template, which contains the information about the authors of the squashed commits and whether it should be used. It makes the implementation contextual and complex though.
> 
> Looks good to me. Do we already have an example of such an approach? I can't find something similar in the code.

No, and I'm not sure if this approach will scale (when we have to add other contexts for other commands.)

I think it's better to add per-command templates (like `revert_description`.) The inputs to "squash" template will be `Vec<Commit>`, but as I said, we need to somehow remove `commits[0].author()` from the list of `Co-authored-by` trailers.

PhilipMetzger:
> > I think an idea could be providing a `description_context` type to the `commit_trailers` template, which contains the information about the authors of the squashed commits and whether it should be used. It makes the implementation contextual and complex though.
> 
> Looks good to me. Do we already have an example of such an approach? I can't find something similar in the code.

No, it's not a thing which exists and also is very complex to implement, leaving the contextual parts aside. 

> No, and I'm not sure if this approach will scale (when we have to add other contexts for other commands.)

👍 

> I think it's better to add per-command templates (like `revert_description`.) The inputs to "squash" template will be `Vec<Commit>`, but as I said, we need to somehow remove `commits[0].author()` from the list of `Co-authored-by` trailers.

so the template will be `squash_description(commits)` where the type of commits is `List[Commit]`

yuja:
> so the template will be `squash_description(commits)` where the type of commits is `List[Commit]`

Yes, that's the idea. (The config key won't have parameters list, btw.)

We could add `CommitsToSquash { destination: Commit, sources: Vec<Commit> }` template type, but doing that would be tedious as you mentioned.

PhilipMetzger:
> > so the template will be `squash_description(commits)` where the type of commits is `List[Commit]`
> 
> Yes, that's the idea. (The config key won't have parameters list, btw.)

Care to elaborate what you mean with the config key, since magically injecting this information seems not a thing we've done before and adding a parameter aligns with `revsets.git-push(remote)`. 

> We could add `CommitsToSquash { destination: Commit, sources: Vec<Commit> }` template type, but doing that would be tedious as you mentioned.

Yeah, adding contextual information with a new type seems to be really tedious.

yuja:
> Care to elaborate what you mean with the config key, since magically injecting this information seems not a thing

`templates.<name>` (and `revsets.<name>`) config keys don't include parameters list because they have to be looked up by `<name>`. See cli/src/config/templates.toml.

PhilipMetzger:
ah, OK got it

glehmann:
It appears that introducing a new type to the templating system would be quite a complex endeavor. Additionally, the user interface implications—specifically, needing a separate template and not being enabled by default—are not what I'd ideally want for users. So I'm not continuing with the template-based implementation of this feature.

The non-template-based implementation is ready enough for review if a maintainer is interested in that version. I'm perfectly fine if it's rejected; Just let me know so I can close this PR.

PhilipMetzger:
> It appears that introducing a new type to the templating system would be quite a complex endeavor. Additionally, the user interface implications—specifically, needing a separate template and not being enabled by default—are not what I'd ideally want for users. So I'm not continuing with the template-based implementation of this feature.

I assume the current state is good enough to land, although the approach suggested by Yuya is definitely better (also keeping it opt-in is important because we don't make choices for users).

emilazy:
> also keeping it opt-in is important because we don't make choices for users

I don't mean to imply a strong opinion one way or another on this feature and its implementation, but could you explain this principle and how you believe Jujutsu embodies it? For example, how does it disapprove of making this functionality opt-out, but approve of making writing the change ID header opt-out, as was done recently?

PhilipMetzger:
> I don't mean to imply a strong opinion one way or another on this feature and its implementation, but could you explain this principle and how you believe Jujutsu embodies it?

In the years I've been active in the project, neither a specific workflow nor any custom template was forced upon users just because it was a "good idea" and this shows because we have a large config page allowing you to customize almost everything (from the graph style and prioritization and how your graph symbols should look). At every such decision the project has not chosen to make the "most community accepted" thing the default, not because some member preferred it.  

> For example, how does it disapprove of making this functionality opt-out, but approve of making writing the change ID header opt-out, as was done recently?

The change-id header was planned as _only opt-out_ (as the semantics don't really change, https://github.com/jj-vcs/jj/issues/6689#issuecomment-2946761017)  and it was done in collaboration with other projects which also want to use it, you were there during the discussions.

ilyagr:
(Non-actionable) I haven't followed the discussion or looked carefully at the PR yet, but I'm glad people are thinking about this. I was also wondering about something like `jj describe --add-coauthor "JJFan <jjfan@examp[le.com>"` or `jj describe --add-coauthors-from CHANGEID`.

emilazy:
> In the years I've been active in the project, neither a specific workflow nor any custom template was forced upon users just because it was a "good idea"

We do have default templates and the like, though. So that corresponds to an opt-out situation. And Jujutsu clearly has preferences in workflow about what it makes easy and difficult - workflows involving lots of unignored but untracked files were impossible for years, workflows involving very regularly bumping a bookmark are still somewhat second class, and so on.

That's not to say those can't or don't get addressed but clearly Jujutsu is a tool with opinions about version control - otherwise it would accommodate people who want it to be exactly like Git. That's inherently making decisions for users; programming is the act of turning universal machines into ones that can only do a specific subset of things. Design and making trade-offs are unavoidable, and the decision to use a tool necessarily entails accepting all the decisions that have been made for you as part of that.

> At every such decision the project has not chosen to make the "most community accepted" thing the default, not because some member preferred it.

I'm not sure this is true? I think that things that have a strong level of consensus among the most informed and experienced and engaged users of Jujutsu are much more likely to be the default than things that aren't. Maybe I misunderstand what you mean by this. If you mean that defaults should be decided based on their individual merits rather than popular vote, then I of course agree.

> The change-id header was planned as _only opt-out_ (as the semantics don't really change, [#6689 (comment)](https://github.com/jj-vcs/jj/issues/6689#issuecomment-2946761017)) and it was done in collaboration with other projects which also want to use it, you were there during the discussions.

Right, I agree with all of that. But "change ID header writing that is opt out and then mandatory" and "coauthor writing that is opt out" are both equally making decisions for users, right? They both might work well with some workflows and badly for others.

I'm not saying that the decision needs to be the same in both cases, but I don't understand how "we don't make decisions for users" can be used as a rubric to accept one and reject the other.

There are many other reasons you could take that position - e.g., considering that the CLI should never alter the user-supplied description by default (though it already does in a very minor way but ensuring a trailing newline on the subject line), or taking the position that the bases where the change ID header breaks things are essentially bugs in other software but that wanting to squash while discarding cosuthorship is a common enough workflow to not work against it by default, or being opposed to any integration of Git trailers beyond templates and revsets even in the Git backend and keeping additional metadata like this exclusive to a future native backend.

But I think it's better for other contributors to go into detail about those reasons, since it's hard to respond to this kind of generality, especially when it doesn't seem like a rubric that accurately predicts the project's decisions as stated (unless I am misinterpreting it from what you have said, but this is my best attempt to infer the principle from your comment, so if it's inaccurate then that's a good indication that being more specific is likely to help others understand what you mean as well).

PhilipMetzger:
> > At every such decision the project has not chosen to make the "most community accepted" thing the default, not because some member preferred it.
> 
> I'm not sure this is true? I think that things that have a strong level of consensus among the most informed and experienced and engaged users of Jujutsu are much more likely to be the default than things that aren't.

Yes, you understood correctly since the _not_ in the sentence is wrong. 

> > The change-id header was planned as _only opt-out_ (as the semantics don't really change, [#6689 (comment)](https://github.com/jj-vcs/jj/issues/6689#issuecomment-2946761017)) and it was done in collaboration with other projects which also want to use it, you were there during the discussions.
> 
> Right, I agree with all of that. But "change ID header writing that is opt out and then mandatory" and "coauthor writing that is opt out" are both equally making decisions for users, right? They both might work well with some workflows and badly for others.
> 
> I'm not saying that the decision needs to be the same in both cases, but I don't understand how "we don't make decisions for users" can be used as a rubric to accept one and reject the other.
> 
> There are many other reasons you could take that position - e.g., considering that the CLI should never alter the user-supplied description by default (though it already does in a very minor way but ensuring a trailing newline on the subject line), or taking the position that the bases where the change ID header breaks things are essentially bugs in other software but that wanting to squash while discarding cosuthorship is a common enough workflow to not work against it by default, or being opposed to any integration of Git trailers beyond templates and revsets even in the Git backend and keeping additional metadata like this exclusive to a future native backend.

Yes, the change-id header only got standardized as a opt-out as an experiment as the project debated on doing something like it for three years (and then finally did it hesitantly). 

Authorship is another problem because it's not something we can represent or support in other backends yet, which makes it just not a good default for "opt-out", because all in all while Jujutsu contains a nicer Git frontend, it definitely isn't that at its core.

It's very clear to me that it is a good default for _Git clients_ to respect authorship while rewriting commits, but as that hasn't happened yet and only GH does something like it with suggestions feature it very clearly should stay opt-in. (You're more than free to configure `jj` to add the squashed co-authors by default, it just isn't something the upstream should do).

yuja:
FWIW, if `Change-Id` were added to trailers (i.e. commit description), I would say it should be opt-in. IMO, anything in trailers is just convention, and different projects follow different conventions. I believe some Git users don't use trailers at all, and they might have to delete `Co-authored-by` if added automatically.

martinvonz:
Maybe we could support having multiple authors in the data model (`jj_lib::backend::Commit`) and have a config to make the Git backend store any authors after the first one in the commit trailer then? If disabled, it would just discard additional users.

glehmann:
> Maybe we could support having multiple authors in the data model (`jj_lib::backend::Commit`) and have a config to make the Git backend store any authors after the first one in the commit trailer then? If disabled, it would just discard additional users.

Support for multiple authors in `Commit` would make possible to configure the `Co-authored-by` trailers in the `commit_trailers` template.

```toml
[templates]
commit_trailers = "concat(authors.filter(|a| a != author).map(|a| 'Co-authored-by: ' ++ a ++ '\n'))"
```

That way we wouldn't have to modify the description in the backend.
The co-authors should still be read from the description by the backend, though.

yuja:
> > Maybe we could support having multiple authors in the data model (`jj_lib::backend::Commit`) and have a config to make the Git backend store any authors after the first one in the commit trailer then? If disabled, it would just discard additional users.
> 
> Support for multiple authors in `Commit` would make possible to configure the `Co-authored-by` trailers in the `commit_trailers` template.

Since trailer/description template will use a commit object temporarily written to the store, the backend will have to serialize authors somewhere. It might work (without serialization) if `GitBackend::write_commit()` didn't drop authors from the in-memory commit data, but that would be a source of bugs.

Co-authors could be stored in extras, but then we cannot deprecate the extras table. That's unfortunate.

FWIW, if the backend adds `Co-authored-by` trailers when writing commit, I think these trailers should be stripped when reading commit.

emilazy:
> Yes, the change-id header only got standardized as a opt-out as an experiment as the project debated on doing something like it for three years (and then finally did it hesitantly).
> 
> Authorship is another problem because it's not something we can represent or support in other backends yet, which makes it just not a good default for "opt-out", because all in all while Jujutsu contains a nicer Git frontend, it definitely isn't that at its core.
> 
> It's very clear to me that it is a good default for _Git clients_ to respect authorship while rewriting commits, but as that hasn't happened yet and only GH does something like it with suggestions feature it very clearly should stay opt-in. (You're more than free to configure `jj` to add the squashed co-authors by default, it just isn't something the upstream should do).

Thanks, this reasoning makes more sense to me. I don’t fully agree that it’s obvious what the default should be here, but I think we agree now that the reasoning is different from just “we don’t make decisions from users” so it resolves my confusion.

> Maybe we could support having multiple authors in the data model (`jj_lib::backend::Commit`) and have a config to make the Git backend store any authors after the first one in the commit trailer then? If disabled, it would just discard additional users.

This is what makes sense to me as the cleanest solution. In this case I think it would have to be unconditional when writing commits with multiple authors for the reasons @yuja said, but you could set a mode to always interpret Git commits as having one author and refuse to create commits with multiple authors, I guess? (In other words, it’s not really an interoperability problem; since Git doesn’t fully natively support commits with multiple authors you can always just not create them.)

I think @yuja was opposed to such processing of commit descriptions in https://github.com/jj-vcs/jj/pull/6662#issuecomment-2984352193, but personally I think it makes sense. Trailers are kind of intended for machine interpretation (with Git shipping native tooling for it), and GitHub already treats `Co-authored-by:` lines as first‐class authors unconditionally. It’s also normal for things like `Signed-off-by:` and `Change-Id:` to get processed automatically.

I think that whether to strip the processed lines from the Jujutsu idea of the `description` is tricky though, and it’s problematic when it reduces user control of the final description though. e.g., you might want to use a certain ordering of trailer lines on a commit with multiple authors, so stripping it out would be weird. This is of course downstream of Git’s unfortunate decision to make trailers both machine‐processed metadata and an exposed part of the user‐controlled description field. I would personally lean to a solution that lets the user produce the trailer ordering they want, but I’m not sure how that could be reliably achieved; it’s not clear to me how you write out to Git a commit with authors {A, B} but `Co-authored-by: C` in the description, since it would have to change along the way.

PR #6672, opened 2025-06-03
index: heads_pos(): push parents to queue while finding heads

scott2000:
I implemented this in #6659, but I think it can be a separate PR.

Instead of immediately populating the queue with the parents of all candidate commits, we can push the parents as we iterate over candidate commits, which seems to improve performance slightly.

```
revsets/heads(tags())           28.238% improvement
revsets/heads(::v2.40.0)        50.456% improvement
```

# Checklist

If applicable:

### DIFF
--- a/lib/src/default_index/composite.rs
+++ b/lib/src/default_index/composite.rs
@@ -15,7 +15,6 @@
 #![allow(missing_docs)]
 
 use std::cmp::max;
-use std::cmp::min;
 use std::cmp::Ordering;
 use std::collections::binary_heap;
 use std::collections::BTreeSet;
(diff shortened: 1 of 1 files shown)


lib/src/default_index/composite.rs
```diff
@@ -380,44 +379,47 @@ impl CompositeIndex {
...
-                continue;
-            }
-            for parent_pos in parent_positions {
-                assert!(parent_pos < cur_pos);
-                work.push(parent_pos);
+        let min_generation = candidate_positions
+            .iter()
+            .map(|&pos| self.entry_by_pos(pos).generation_number())
+            .min()
+            .unwrap_or(u32::MAX);
```
yuja: nit: simply return early if the candidates set is empty?

lib/src/default_index/composite.rs
```diff
@@ -380,44 +379,47 @@ impl CompositeIndex {
...
+        // already-found head, then it can be removed.
+        let mut parents = BinaryHeap::new();
+        let mut heads = BTreeSet::new();
+        'outer: for candidate in candidate_positions.into_iter().rev() {
+            while let Some(&parent) = parents.peek() {
+                if parent < candidate {
+                    break;
+                }
+                let entry = self.entry_by_pos(parent);
+                if entry.generation_number() < min_generation {
```
yuja: nit: `entry.generation_number() <= min_generation`, but we'll need to test `parent == candidate`.

It might be better to extract a helper function (similar to `common_ancestors_pos()`) to clean up the control flow.

lib/src/default_index/composite.rs
```diff
@@ -380,44 +379,47 @@ impl CompositeIndex {
...
+                    dedup_pop(&mut parents).unwrap();
+                    continue;
+                }
+                let mut new_parents = entry.parent_positions().into_iter();
+                if let Some(first_parent) = new_parents.next() {
+                    dedup_replace(&mut parents, first_parent).unwrap();
+                    parents.extend(new_parents);
```
yuja: Can you restore assertions, which prevents infinite loop?

yuja (commented):
Nice.

lib/src/default_index/composite.rs
scott2000: I just extracted a common function that should work for both `heads_pos()` and `common_ancestors_pos()` now.

lib/src/default_index/composite.rs
```diff
@@ -380,44 +360,43 @@ impl CompositeIndex {
...
+        // Iterate though the candidates by reverse index position, keeping track of the
+        // ancestors of already-found heads. If a candidate is an ancestor of an
+        // already-found head, then it can be removed.
+        let mut parents = BinaryHeap::new();
+        let mut heads = BTreeSet::new();
+        'outer: for candidate in candidate_positions.into_iter().rev() {
+            while let Some(&parent) = parents.peek() {
+                if parent < candidate {
+                    break;
+                }
```
yuja: nit: (at your option,) this could be `.peek().filter(..)`

yuja (approved):

PR #6679, opened 2025-06-03
revset: optimize heads of a range with filters

### LINKED ISSUE #6656: Pathological performance for `heads(::@ & ~empty())`

scott2000:
Resolves #6656. Suggested by @yuja.

Currently, evaluating an expression like `heads(::@ ~ empty())` requires iterating over all commits in `::@` and checking the predicate against each commit, which can be very slow. This PR implements an optimization to convert expressions to the form `heads(x..y & filter)` when it is possible to do so efficiently. Expressions of this form can be evaluated using a specialized algorithm which terminates iteration early without checking all commits.

```
revsets/heads(author(peff))     97.248% improvement
revsets/heads(::v2.40.0)        98.467% improvement
```

# Checklist

If applicable:

### DIFF
--- a/lib/src/default_index/composite.rs
+++ b/lib/src/default_index/composite.rs
@@ -404,6 +404,38 @@ impl CompositeIndex {
         heads
     }
 
+    /// Find the heads of a range of positions `roots..heads`, applying a filter
+    /// to the commits in the range. The heads are sorted in descending order.
+    /// The filter will also be called in descending index position order.
+    pub fn heads_from_range_and_filter<E>(
+        &self,
+        roots: Vec<IndexPosition>,
+        heads: Vec<IndexPosition>,
+        mut filter: impl FnMut(IndexPosition) -> Result<bool, E>,
+    ) -> Result<Vec<IndexPosition>, E> {
+        if heads.is_empty() {
+            return Ok(heads);
+        }
+        let mut wanted_queue = BinaryHeap::from(heads);
+        let mut unwanted_queue = BinaryHeap::from(roots);
+        let mut found_heads = Vec::new();
+        while let Some(&pos) = wanted_queue.peek() {
+            if shift_to_parents_until(&mut unwanted_queue, self, pos) {
+                dedup_pop(&mut wanted_queue);
+                continue;
+            }
+            let entry = self.entry_by_pos(pos);
+            if filter(pos)? {
+                dedup_pop(&mut wanted_queue);
+                unwanted_queue.extend(entry.parent_positions());
+                found_heads.push(pos);
+            } else {
+                shift_to_parents(&mut wanted_queue, &entry);
+            }
+        }
+        Ok(found_heads)
+    }
+
     pub(super) fn evaluate_revset(
         &self,
         expression: &ResolvedExpression,
@@ -579,6 +611,22 @@ pub struct IndexStats {
     pub num_heads: u32,
     pub num_changes: u32,
     pub levels: Vec<IndexLevelStats>,
+}
+
+/// Repeatedly `shift_to_parents` until reaching a target position. Returns true
+/// if the target position matched a position in the queue.
+fn shift_to_parents_until(
+    queue: &mut BinaryHeap<IndexPosition>,
+    index: &CompositeIndex,
+    target_pos: IndexPosition,
+) -> bool {
+    while let Some(&pos) = queue.peek().filter(|&&pos| pos >= target_pos) {
+        shift_to_parents(queue, &index.entry_by_pos(pos));
+        if pos == target_pos {
+            return true;
+        }
+    }
+    false
 }
 
 /// Removes an entry from the queue and replace it with its parents.
--- a/lib/src/default_index/mod.rs
+++ b/lib/src/default_index/mod.rs
@@ -46,6 +46,8 @@ pub use self::store::DefaultIndexStoreInitError;
 
 #[cfg(test)]
 mod tests {
+    use std::cmp::Reverse;
+    use std::convert::Infallible;
     use std::sync::Arc;
 
     use itertools::Itertools as _;
@@ -1225,6 +1227,103 @@ mod tests {
         assert_eq!(
             index.all_heads_for_gc().unwrap().collect_vec(),
             vec![id_3.clone(), id_5.clone()]
+        );
+    }
+
+    #[test]
+    fn test_heads_range_with_filter() {
+        let mut new_change_id = change_id_generator();
+        let mut index = DefaultMutableIndex::full(3, 16);
+        // 5
+        // |\
+        // 4 | 3
+        // | |/
+        // 1 2
+        // |/
+        // 0
+        let id_0 = CommitId::from_hex("000000");
+        let id_1 = CommitId::from_hex("111111");
+        let id_2 = CommitId::from_hex("222222");
+        let id_3 = CommitId::from_hex("333333");
+        let id_4 = CommitId::from_hex("444444");
+        let id_5 = CommitId::from_hex("555555");
+        index.add_commit_data(id_0.clone(), new_change_id(), &[]);
+        index.add_commit_data(id_1.clone(), new_change_id(), &[id_0.clone()]);
+        index.add_commit_data(id_2.clone(), new_change_id(), &[id_0.clone()]);
+        index.add_commit_data(id_3.clone(), new_change_id(), &[id_2.clone()]);
+        index.add_commit_data(id_4.clone(), new_change_id(), &[id_1.clone()]);
+        index.add_commit_data(id_5.clone(), new_change_id(), &[id_4.clone(), id_2.clone()]);
+
+        // Helper function to convert commit IDs to/from index positions and call
+        // `heads_from_range_and_filter`.
+        let heads_range = |roots: &[&CommitId],
+                           heads: &[&CommitId],
+                           filter: &dyn Fn(CommitId) -> bool|
+         -> Vec<CommitId> {
+            let roots = roots
+                .iter()
+                .map(|id| index.as_composite().commit_id_to_pos(id).unwrap())
+                .sorted_by_key(|&pos| Reverse(pos))
+                .collect_vec();
+            let heads = heads
+                .iter()
+                .map(|id| index.as_composite().commit_id_to_pos(id).unwrap())
+                .sorted_by_key(|&pos| Reverse(pos))
+                .collect_vec();
+            index
+                .as_composite()
+                .heads_from_range_and_filter::<Infallible>(roots, heads, |pos| {
+                    Ok(filter(index.as_composite().entry_by_pos(pos).commit_id()))
+                })
+                .unwrap()
+                .into_iter()
+                .map(|pos| index.as_composite().entry_by_pos(pos).commit_id())
+                .collect_vec()
+        };
+
+        // heads(::none())
+        assert!(heads_range(&[], &[], &|_| true).is_empty());
+        // heads(all())
+        assert_eq!(
+            heads_range(&[], &[&id_5, &id_3], &|_| true),
+            vec![id_5.clone(), id_3.clone()]
+        );
+        // heads(~5)
+        assert_eq!(
+            heads_range(&[], &[&id_5, &id_3], &|id| id != id_5),
+            vec![id_4.clone(), id_3.clone()]
+        );
+        // heads(5..)
+        assert_eq!(
+            heads_range(&[&id_5], &[&id_5, &id_3], &|_| true),
+            vec![id_3.clone()]
+        );
+        // heads(5.. ~ 3)
+        assert!(heads_range(&[&id_5], &[&id_5, &id_3], &|id| id != id_3).is_empty());
+        // heads(2..4)
+        assert_eq!(
+            heads_range(&[&id_2], &[&id_4], &|_| true),
+            vec![id_4.clone()]
+        );
+        // heads(2..4 ~ 4)
+        assert_eq!(
+            heads_range(&[&id_2], &[&id_4], &|id| id != id_4),
+            vec![id_1.clone()]
+        );
+        // heads((3 | 1).. ~ 5)
+        assert_eq!(
+            heads_range(&[&id_3, &id_1], &[&id_5, &id_3], &|id| id != id_5),
+            vec![id_4.clone()]
+        );
+        // heads(::(5 | 4) ~ 5)
+        assert_eq!(
+            heads_range(&[], &[&id_5, &id_4], &|id| id != id_5),
+            vec![id_4.clone(), id_2.clone()]
+        );
+        // heads(::(5 | 5))
+        assert_eq!(
+            heads_range(&[], &[&id_5, &id_5], &|_| true),
+            vec![id_5.clone()]
         );
     }
 }
--- a/lib/src/default_index/revset_engine.rs
+++ b/lib/src/default_index/revset_engine.rs
@@ -19,6 +19,7 @@ use std::cmp::Ordering;
 use std::cmp::Reverse;
 use std::collections::BinaryHeap;
 use std::collections::HashSet;
+use std::convert::Infallible;
 use std::fmt;
 use std::iter;
 use std::ops::Range;
@@ -935,6 +936,39 @@ impl EvaluationContext<'_> {
                 let candidate_set = self.evaluate(candidates)?;
                 let positions =
                     index.heads_pos(candidate_set.positions().attach(index).try_collect()?);
+                Ok(Box::new(EagerRevset { positions }))
+            }
+            ResolvedExpression::HeadsRange {
+                roots,
+                heads,
+                filter,
+            } => {
+                let root_set = self.evaluate(roots)?;
+                let root_positions: Vec<_> = root_set.positions().attach(index).try_collect()?;
+                // Pre-filter heads so queries like 'immutable_heads()..' can
+                // terminate early. immutable_heads() usually includes some
+                // visible heads, which can be trivially rejected.
+                let head_set = self.evaluate(heads)?;
+                let head_positions = difference_by(
+                    head_set.positions(),
+                    EagerRevWalk::new(root_positions.iter().copied().map(Ok)),
+                    |pos1, pos2| pos1.cmp(pos2).reverse(),
+                )
+                .attach(index)
+                .try_collect()?;
+                let positions = if let Some(filter) = filter {
+                    let mut filter = self.evaluate_predicate(filter)?.to_predicate_fn();
+                    index.heads_from_range_and_filter(root_positions, head_positions, |pos| {
+                        filter(index, pos)
+                    })?
+                } else {
+                    let Ok(positions) = index.heads_from_range_and_filter::<Infallible>(
+                        root_positions,
+                        head_positions,
+                        |_| Ok(true),
+                    );
+                    positions
+                };
                 Ok(Box::new(EagerRevset { positions }))
             }
             ResolvedExpression::Roots(candidates) => {
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -260,6 +260,13 @@ pub enum RevsetExpression<St: ExpressionState> {
         domain: Rc<Self>,
     },
     Heads(Rc<Self>),
+    /// Heads of the set of commits which are ancestors of `heads` but are not
+    /// ancestors of `roots`, and which also are contained in `filter`.
+    HeadsRange {
+        roots: Rc<Self>,
+        heads: Rc<Self>,
+        filter: Rc<Self>,
+    },
     Roots(Rc<Self>),
     ForkPoint(Rc<Self>),
     Latest {
@@ -620,6 +627,13 @@ pub enum ResolvedExpression {
         domain: Box<Self>,
     },
     Heads(Box<Self>),
+    /// Heads of the set of commits which are ancestors of `heads` but are not
+    /// ancestors of `roots`, and which also are contained in `filter`.
+    HeadsRange {
+        roots: Box<Self>,
+        heads: Box<Self>,
+        filter: Option<ResolvedPredicateExpression>,
+    },
     Roots(Box<Self>),
     ForkPoint(Box<Self>),
     Latest {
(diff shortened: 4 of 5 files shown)


lib/src/revset.rs
```diff
@@ -2315,7 +2363,16 @@ impl VisibilityResolutionContext<'_> {
                 domain: self.resolve(domain).into(),
             },
             RevsetExpression::Heads(candidates) => {
-                ResolvedExpression::Heads(self.resolve(candidates).into())
+                let candidates = self.resolve(candidates);
+                if let Some((roots, heads, predicates)) = candidates.to_range_and_predicates() {
+                    ResolvedExpression::HeadsRange {
```
scott2000: I added this here instead of in `optimize()` because it was easier to construct based on a `ResolvedExpression` than based on a `RevsetExpression`, and it looks like `FilterWithin` does something similar here already. If this isn't the most appropriate place to add it, I can move it somewhere else though.

lib/src/default_index/composite.rs
```diff
@@ -420,6 +422,41 @@ impl CompositeIndex {
...
+        if heads.is_empty() {
+            return Ok(BTreeSet::new());
+        }
+        let min_pos = roots
+            .first()
+            .map_or(IndexPosition::MIN, |&IndexPosition(pos)| {
+                // The minimum root index position should be excluded from the range.
+                IndexPosition(pos + 1)
+            });
+        let mut wanted_queue = RevWalkQueue::with_min_pos(min_pos);
```
scott2000: I'm not sure if this is the right way to do this or if I should be using a plain `BinaryHeap<IndexPosition>` instead. `RevWalkQueue` seems convenient, but I had to make `flush_queue_until` visible in order to use it easily.

lib/src/revset.rs
yuja: `optimize()` might be easier to test, but I don't have strong feeling about that. Another possible reason to do that in `optimize()` is that we can ignore `Range { .. }`.

lib/src/default_index/composite.rs
yuja: I think `roots` can be moved to intersection of negated ancestors (when rewriting the expression.) Suppose `heads()` will terminate early (compared to `ancestors()`), there should be no reason to evaluate `~ancestors(roots)` within the same loop.

scott2000: Yeah, I had considered that. The reason I was thinking `roots` would be good to include here is that if the filter returns false for all commits, then `heads()` won't terminate early. For instance, someone might use a revset like `heads(immutable_heads()..@ ~ empty())` to find the latest non-empty mutable commit, but then if there are no non-empty commits in `immutable_heads()..@`, it would iterate over all of `::@` unnecessarily unless we include `roots` here. So since we already need both queues here anyway, I was thinking that it would be better to just include `roots`.

lib/src/revset.rs
scott2000: > Another possible reason to do that in `optimize()` is that we can ignore `Range { .. }`.

I'm not sure I understand this. Do you mean that we shouldn't handle `heads(x..y)`?

lib/src/default_index/composite.rs
yuja: Hm, true.

Feel free to move helper function to e.g. `rev_walk_queue`, add `heads_from_range_and_filter` to `rev_walk`, or use std `BinaryHeap`. I think the std `BinaryHeap` would perform better, but I don't know if that would matter.

lib/src/revset.rs
yuja: `x..y` is decomposed to `::x & ~::y` during `optimize()` passes.

I think it's easier to do tree transformation as `RevsetExpression` since there are helper functions for that.

scott2000:
I think this should be ready for review now. I moved the optimization to the `optimize()` function, and I added tests.

lib/src/default_index/mod.rs
```diff
@@ -1227,4 +1229,91 @@ mod tests {
...
+        assert_eq!(
+            heads_range(&[&id_2], &[&id_4], &|id| id != id_4),
+            vec![id_1.clone()]
+        );
+        // heads((id_3 | id_1).. ~ id_5)
+        assert_eq!(
+            heads_range(&[&id_3, &id_1], &[&id_5, &id_3], &|id| id != id_5),
+            vec![id_4.clone()]
+        );
+    }
```
yuja, line 1328: nit: can you add some tests having duplicates or duplicates filtered out by predicate? something like: `heads(::(5|4) ~ 5)`, `heads(::(5|5))`.

lib/src/default_index/revset_engine.rs
```diff
@@ -937,6 +937,42 @@ impl EvaluationContext<'_> {
...
+                    .positions()
+                    .attach(index)
+                    .try_collect()?;
+                let head_positions: Vec<_> = difference_by(
+                    self.evaluate(heads)?.positions(),
+                    EagerRevWalk::new(root_positions.iter().copied().map(Ok)),
+                    |pos1, pos2| pos1.cmp(pos2).reverse(),
+                )
+                .attach(index)
+                .try_collect()?;
```
yuja: nit: is this for optimization? can you copy the comment from `Range` handling?

lib/src/revset.rs
```diff
@@ -260,6 +260,13 @@ pub enum RevsetExpression<St: ExpressionState> {
         domain: Rc<Self>,
     },
     Heads(Rc<Self>),
+    /// Heads of the set of commits which are ancestors of `heads` and also
+    /// satisfy all of `predicates`, but are not ancestors of `roots`.
+    HeadsRange {
+        roots: Rc<Self>,
+        heads: Rc<Self>,
+        predicates: Vec<Rc<Self>>,
```
yuja: Just `predicate: Rc<Self>`?

lib/src/revset.rs
```diff
@@ -632,6 +639,13 @@ pub enum ResolvedExpression {
         domain: Box<Self>,
     },
     Heads(Box<Self>),
+    /// Heads of the set of commits which are ancestors of `heads` and also
+    /// satisfy all of `predicates`, but are not ancestors of `roots`.
+    HeadsRange {
+        roots: Box<Self>,
+        heads: Box<Self>,
+        predicates: Vec<ResolvedPredicateExpression>,
```
yuja: Same here, `Box<_>`. It might also be better to use `Box<Self>` so the backend can evaluate it as a set if that makes sense.

lib/src/revset.rs
```diff
@@ -1775,6 +1848,98 @@ fn fold_generation<St: ExpressionState>(
...
+                if matches!(
+                    expression2.as_ref(),
+                    RevsetExpression::Filter(_) | RevsetExpression::AsFilter(_)
+                ) =>
+            {
+                to_filtered_range(expression1).map(|mut filtered_range| {
+                    filtered_range.predicates.push(expression2.clone());
+                    filtered_range
+                })
+            }
```
yuja: I think we can move `ancestors(_)` and `not(ancestors(_))` expressions up (or left) recursively in a similar way to `internalize_filter()`. Then we can peek the leftmost expression at `heads(_)`?

BTW, silly expressions like `heads(all())`, `heads(::x)` can also be transformed.

lib/src/revset.rs
```diff
@@ -1785,7 +1950,8 @@ pub fn optimize<St: ExpressionState>(
     let expression = fold_generation(&expression).unwrap_or(expression);
     let expression = internalize_filter(&expression).unwrap_or(expression);
     let expression = fold_difference(&expression).unwrap_or(expression);
```
yuja, line 2058: it's probably better to insert `fold_heads_range()` between `internalize_filter()` and `fold_difference()`. Then, we wouldn't have to deal with composite expression nodes.

lib/src/revset.rs
```diff
@@ -4292,6 +4471,131 @@ mod tests {
...
+
+        // Heads of basic range operators can be folded
+        insta::assert_debug_snapshot!(optimize(parse("heads(::)").unwrap()), @r"
+        HeadsRange {
+            roots: None,
+            heads: VisibleHeads,
+            predicates: [],
+        }
+        ");
+        insta::assert_debug_snapshot!(optimize(parse("heads(::foo)").unwrap()), @r#"
```
yuja, line 4838: nit: can you add test for `heads(::foo-)`? `::foo-` will be transformed to `ancestors(foo, depth=1..)`.

yuja (commented):
Generally looks good to me, thanks.

lib/src/revset.rs
scott2000: > I think we can move `ancestors(_)` and `not(ancestors(_))` expressions up (or left) recursively in a similar way to `internalize_filter()`. Then we can peek the leftmost expression at `heads(_)`?

Ohh I see, yeah that seems simpler, thanks! Would it also be acceptable to transform expressions like `::foo & ::bar` to something like this?

```rust
HeadsRange {
  roots: None,
  heads: CommitRef("foo"),
  predicate: Ancestors(CommitRef("bar")),
}
```

Currently, I was only transforming intersections where the RHS was a filter or a negation, but it might be simpler to just transform everything.

yuja: > to transform expressions like `::foo & ::bar` to something like this?
> 
> ```rust
> HeadsRange {
>   roots: None,
>   heads: CommitRef("foo"),
>   predicate: Ancestors(CommitRef("bar")),
> }
> ```

Yeah, that sounds like a good idea.

lib/src/revset.rs
scott2000: Would it be okay to add `ResolvedPredicateExpression::Intersection` for this? There's already `Union` it looks like, so it seems consistent. Because otherwise, I think `f1 & f2` would need to be represented as `Set(FilterWithin(FilterWithin(all(), f1), f2))` or `Set(Intersection(FilterWithin(all(), f1), FilterWithin(all(), f2)))`, both of which would be unnecessarily inefficient.

lib/src/revset.rs
scott2000: How should I represent no predicate then? Maybe I can use `RevsetExpression::All` here, and then make the backend expression have predicate be `Option<ResolvedPredicateExpression>` (with `All` mapping to `None`)?

Alternatively, I guess I could add a `ResolvedPredicateExpression::True`, but that would be difficult to handle in `revset_engine` currently I think?

lib/src/revset.rs
yuja: Can't we simply use `ResolvedExpression::Intersection`? Filters should be pushed down, so `set & f1 & f2` would be represented as `filter_within(filter_within(set, f1), f2)`, which is equivalent to `p(x) = (set(x) & f1(x)) & f2(x)`.

scott2000: How would `heads(::x & f1 & f2)` be represented? It seems inefficient to duplicate `::x` when it will never affect the filter. Like I think it would look like this:

```rust
HeadsRange {
    roots: None,
    heads: x,
    filter: FilterWithin {
        candidates: FilterWithin {
            candidates: Ancestors(x),
            predicate: f1,
        },
        predicate: f2,
    },
}
```

Which would require evaluating `::x` twice.

lib/src/revset.rs
yuja: Simply use `RevsetExpression::All` and transform it to `resolve_all()`?

BTW, if `predicate == all()`, I don't think we should emit `HeadsRange`. `heads(::y)` is equivalent to `heads(y)`, and `heads(x..y)` can be represented as `heads_range(none(), y, ~::x)`?

scott2000: I think `heads_range(none(), y, ~::x)` might be less efficient than `heads_range(x, y, all())` actually, because it couldn't stop iterating early (e.g. if `x..y` is empty). It would also make the logic more complicated to add a special case for this I think.

At that point, it would probably be better to just do `heads_ancestors(heads, filter)` and accept the performance hit when no commits match the filter and we need to iterate over the entire `ancestors(heads)`.

lib/src/revset.rs
yuja: > `heads(::x & f1 & f2)`

```rust
HeadsRange {
    roots: None,
    heads: x,
    filter: FilterWithin {
        candidates: FilterWithin {
            candidates: All,
            predicate: f1,
        },
        predicate: f2,
    },
}
```
?

Oh, btw, I think there are some TODO around optimizing the filter version of `All` out.

lib/src/revset.rs
yuja: I don't have strong feeling about that. `heads_ancestors(heads, filter)` seems also good. However, the complexity to support `heads_range(roots, heads, filter)` wouldn't be that high, I think?

lib/src/revset.rs
scott2000: Yeah, I suppose that would work then, and the TODO could be addressed later. Thanks!

lib/src/revset.rs
scott2000: > I think we can move `ancestors(_)` and `not(ancestors(_))` expressions up (or left) recursively in a similar way to `internalize_filter()`. Then we can peek the leftmost expression at `heads(_)`?

I attempted to do this, but I'm not sure this is what you had in mind. I split it into 3 passes:
1. Flatten intersections to be left-recursive to make other optimizations easier.
2. Push `ancestors(_)` and `~ancestors(_)` down/left in intersections.
3. Optimize `heads()` to `heads_range()` when the leftmost expression in the intersection matches a certain pattern.

Changes 1 and 2 also required some additional modifications to `fold_difference` to preserve its optimizations, since they changed where some negated ancestors are and how they're grouped.

lib/src/revset.rs
```diff
@@ -1536,6 +1536,52 @@ fn flatten_intersections<St: ExpressionState>(
...
+    }
+
+    fn should_swap<St: ExpressionState>(
+        expression1: &Rc<RevsetExpression<St>>,
+        expression2: &Rc<RevsetExpression<St>>,
+    ) -> bool {
+        !is_ancestors(expression1) && is_ancestors(expression2)
+    }
+
+    fn intersect_ancestors_down<St: ExpressionState>(
```
scott2000: If this approach is okay, maybe I could extract this as a helper function `sort_intersection(expression1, expression2, should_swap)` and use it both for `push_ancestors_down()` and `internalize_filter()`?

lib/src/revset.rs
```diff
@@ -3760,32 +3807,35 @@ mod tests {
         )
         "#);
 
-        // Negated ancestors are combined into differences when possible.
+        // TODO: We shouldn't need a range with `visible_heads()`, since it was
+        // originally a difference.
         insta::assert_debug_snapshot!(optimize(parse("roots(a) ~ ::c ~ ::d").unwrap()), @r#"
```
yuja: nit: this problem can be triggered by `~::c & ~::d & roots(a)`, and can be fixed separately.

It might also make sense to add pass to transform `~::c & ~::d` to `~::(c|d)` (or `~::c & ~::d -> ~(::c | ::d) -> ~(::(c|d))`), then the earlier bad example `a..b & c..d` could be translated to `(a|c)..b & ::d` or something. The latter would be as good as the original expression, but a fewer adhoc rule, the better.

lib/src/revset.rs
yuja: I didn't have a concrete idea, but your implementation looks generally good to me, thanks.

lib/src/revset.rs
yuja: > Would it be okay to add `ResolvedPredicateExpression::Intersection` for this?

While it's true that we can use `Box<Self>` (i.e. `ResolvedExpression::Intersection`) here, you're probably right. Since we have rule to combine `f1() & f2() | f3()` into a single filter node, it makes sense to translate it to a single backend filter expression. Going back to a set intersection is just silly.

scott2000: Sounds good, thanks!

lib/src/revset.rs
```diff
@@ -1583,6 +1583,41 @@ fn sort_intersection<St: ExpressionState>(
...
+                )
+            }
+            _ => false,
+        }
+    }
+
+    transform_expression_bottom_up(expression, |expression| match expression.as_ref() {
+        RevsetExpression::Intersection(expression1, expression2) => {
+            sort_intersection(expression1, expression2, |a, b| {
+                !is_ancestors(a) && is_ancestors(b)
```
scott2000: I was also considering moving all negative ancestors to the start of the intersection, so that they can be combined by `fold_negations_and_ancestors`. For instance:

```
    ~::a & b..c
==> ~::a & ::c & ~::b
==> ~::a & ~::b & ::c
==> ~::(a | b) & ::c
==> (a | b)..c
```

This could also make the code for `fold_heads_range` simpler, since it would only need to set the roots based on the first expression in the intersection (currently it starts with `none()` and then adds roots as it encounters them).

Do you think this would be a good idea, or is it better to keep `push_ancestors_down` simple?

---

Also, currently this would lead to `::visible_heads()` being included unnecessarily in some cases (but #6721 would fix this):

```
    a..b & c..
==> ::b & ~::a & ::visible_heads() & ~::c
==> ~::a & ~::c & ::b & ::visible_heads()
==> ~::(a | c) & ::b & ::visible_heads()
==> (a | c)..b & ::visible_heads()
```

lib/src/revset.rs
scott2000: I changed it to use `filter: ResolvedPredicateExpression` in `ResolvedExpression::HeadsRange`, but I left it as `filter: Rc<Self>` in `RevsetExpression::HeadsRange`. I think once #6720 gets merged, this would result in pretty optimal evaluation for intersections of filters.

Let me know if you think I should change it back to `Box<Self>`.

scott2000:
@yuja I think this should be ready to review again, but it's a pretty big PR now. Do you think I should split it into smaller PRs, or do you think it's fine as-is?

yuja:
I think the first three patches can be extracted to a separate PR, but I don't feel strongly. (I don't think this PR is too big to review. I just mean some of them can be merged even if the last patch required some churn.)

lib/src/default_index/revset_engine.rs
```diff
@@ -937,6 +937,39 @@ impl EvaluationContext<'_> {
...
+                    .transpose()?
+                    .map(|filter| filter.to_predicate_fn());
+                let positions =
+                    index.heads_from_range_and_filter(root_positions, head_positions, |pos| {
+                        if let Some(filter) = &mut filter {
+                            filter(index, pos)
+                        } else {
+                            Ok(true)
+                        }
+                    })?;
```
yuja: `filter` can be tested out of the loop.

```rust
if let Some(filter) = filter {
    let mut filter = self.evaluate_predicate(filter)?...;
    index.heads_from_range_and_filter(...)?
} else {
    let Ok(positions) = index.heads_from_range_and_filter::<Infallible>(...)?;
    positions
};
```

lib/src/revset.rs
```diff
@@ -1744,6 +1790,101 @@ fn fold_ancestors_union<St: ExpressionState>(
...
+        fn add(mut self, expression: &Rc<RevsetExpression<St>>) -> Self {
+            if self.heads.is_none() {
+                // x.. & ::y -> x..y
+                if let Ok(heads) = ancestors_to_heads(expression) {
+                    self.heads = Some(heads);
+                    return self;
+                }
+            }
+
+            if let RevsetExpression::All = self.filter.as_ref() {
```
yuja, line 1829: nit: I slightly prefer splitting this function here, and call `.add_filter()` where `expression` is obviously not `::heads`.

lib/src/revset.rs
```diff
@@ -1744,6 +1790,101 @@ fn fold_ancestors_union<St: ExpressionState>(
...
+                    // If the first expression is a non-ancestors negation, we still want to use
+                    // `HeadsRange` since `~x` is equivalent to `::visible_heads() ~ x`.
+                    Some(FilteredRange::new(RevsetExpression::none()).add(expression))
+                }
+            }
+            // Otherwise, we only want to optimize `heads()` if the first expression in the
+            // intersection is `all()`, `ancestors(x)`, or a filter.
+            RevsetExpression::All
+            | RevsetExpression::Ancestors {
+                generation: Range { end: u64::MAX, .. },
```
yuja: Can you add test for `end != u64::MAX`?

lib/src/revset.rs
```diff
@@ -4663,6 +4816,151 @@ mod tests {
...
+            heads: CommitRef(Symbol("foo")),
+            filter: All,
+        }
+        "#);
+        insta::assert_debug_snapshot!(optimize(parse("heads(..)").unwrap()), @r"
+        HeadsRange {
+            roots: None,
+            heads: VisibleHeads,
+            filter: NotIn(Root),
+        }
```
yuja, line 4852: nit: can you add comment why not `roots: Root`?

(We could add special case for `Root`, but I don't have strong feeling about that.)

lib/src/default_index/composite.rs
yuja, line 1: nit: can you add the issue number to the commit description?

yuja (approved):
Thanks for working on this!

scott2000:
@yuja Thanks for all of your help with this, I learned a lot!

lgarron:
I'm able to confirm that a local build of `main` now restores the performance issue at https://github.com/jj-vcs/jj/issues/6656#event-18150179100 from 1.65 seconds down to 0.02 seconds in my checkout! Thanks so much for implementing. 🤩

PR #6684, opened 2025-06-04
commit_builder: error out if newly-created commit was already known to repo

yuja:
# Checklist

If applicable:


lib/src/commit_builder.rs
```diff
@@ -351,6 +352,14 @@ impl DetachedCommitBuilder {
     pub fn write(self, mut_repo: &mut MutableRepo) -> BackendResult<Commit> {
         let predecessors = self.commit.predecessors.clone();
         let commit = write_to_store(&self.store, self.commit, &self.sign_settings)?;
+        if mut_repo.index().has_id(commit.id()) {
```
martinvonz: `has_id()` in the Google backend currently always returns true, so we will either have to implement it correctly and efficiently, or we will have to patch it out from our internal build.

Is it an option to make this a debug assertion?

yuja: I don't think `debug_assert` is right because this can be technically triggered by user. Maybe insert the same hack as `merge_view()`?

martinvonz: That would work.

> This can happen if we remove predecessors from the commit object, and if
committer timestamp wasn't updated (like in some jj-lib tests.) It's not wrong
for the store to "overwrite" existing objects, but the repo expects
new/rewritten commits have unique ids. If they didn't, cycle would be created in
predecessor/parent mappings.

I don't understand this. If we remove the predecessor from the commit object, then there can be no cycles in the predecessor mappings. Oh, you mean in `MutableRepo`'s `commit_predecessors`/`parent_mapping`. Could you clarify that in the description?

yuja: Updated commit description, and inserted type check.

PR #6692, opened 2025-06-04
merged_tree: add async version of `path_value()`, use in annotate

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -194,26 +194,30 @@ impl MergedTree {
     /// Gets the `MergeTree` in a subdirectory of the current tree. If the path
     /// doesn't correspond to a tree in any of the inputs to the merge, then
     /// that entry will be replace by an empty tree in the result.
-    pub fn sub_tree(&self, name: &RepoPathComponent) -> BackendResult<Option<MergedTree>> {
+    pub async fn sub_tree(&self, name: &RepoPathComponent) -> BackendResult<Option<MergedTree>> {
         match self.value(name).into_resolved() {
             Ok(Some(TreeValue::Tree(sub_tree_id))) => {
                 let subdir = self.dir().join(name);
                 Ok(Some(MergedTree::resolved(
-                    self.store().get_tree(subdir, sub_tree_id)?,
+                    self.store().get_tree_async(subdir, sub_tree_id).await?,
                 )))
             }
             Ok(_) => Ok(None),
             Err(merge) => {
-                let trees = merge.try_map(|value| match value {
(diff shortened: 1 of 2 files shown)


lib/src/merged_tree.rs
```diff
@@ -223,9 +225,14 @@ impl MergedTree {
     /// `self` is a `Conflict`, which happens if the value at the path can be
     /// trivially merged.
     pub fn path_value(&self, path: &RepoPath) -> BackendResult<MergedTreeValue> {
+        self.path_value_async(path).block_on()
+    }
+
+    /// Async version of `path_value()`.
+    pub async fn path_value_async(&self, path: &RepoPath) -> BackendResult<MergedTreeValue> {
```
PhilipMetzger, line 234: nit: can we reverse the names here, so `async fn path_value` and `fn path_value_sync` since it advertises that `async` is preferred by the project which is true longterm. It's also inconsistent here since `sub_tree` has no sync implementation.

PhilipMetzger (approved):
LG

lib/src/merged_tree.rs
martinvonz: We already have a e.g. `Store::get_commit_async()` and `Merge::try_map_async()`, so the naming here is consistent with those. I agree that we should try to get rid of the sync versions and drop the `_async` suffix at some point.

PhilipMetzger: > We already have a e.g. `Store::get_commit_async()` and `Merge::try_map_async()`, so the naming here is consistent with those.

Then `sub_tree_async` is also more appropriate, if the file already contains so many `*_async` named functions. 

> I agree that we should try to get rid of the sync versions and drop the `_async` suffix at some point.

Yes, and this would be a starting point to make all future PRs shaped like this to follow that. Anyway, it's your decision and I wanted to bring up the conversation we had with Yuya the last time around it.

martinvonz: There's only one `sub_tree()` but there's both `try_map()` and `try_map_async()`.

PhilipMetzger: ok, fair enough

lib/src/merged_tree.rs
```diff
@@ -204,16 +204,18 @@ impl MergedTree {
             }
```
yuja, line 204: `.get_tree_async()` here

yuja (approved):

lib/src/merged_tree.rs
martinvonz: Thanks. Fixed

PR #6704, opened 2025-06-06
CHANGELOG fixup: move diff-do-chdir entry to 0.30, add check in `releasing.md`

ilyagr:
I'm also planning to update the release notes for 0.30 at https://github.com/jj-vcs/jj/releases/tag/v0.30.0 after I merge this, let me know if you think we should handle it differently.

I can't think of a better (simple) technical solution than just looking at the diff before every release. I briefly considered a presubmit check for this, but it doesn't seem fun or worth it to write or maintain (though people are welcome to if they feel otherwise). I think that, as long as it's part of a checklist, it shouldn't be onerous.

I noticed this because of https://github.com/keanemind/jjk/issues/127#issuecomment-2949223664.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   argument and no longer produces Bash completions by default. The deprecated
   optional arguments for different shells have been removed.
 
+* External diff tools are now run in the temporary directory containing
+  the before (`left`) and after (`right`) directories, making diffs appear
+  more pleasing for tools that display file paths prominently. Users can
+  opt out of this by setting `merge-tools.<tool>.diff-do-chdir = false`,
+  but this will likely be removed in a future release. Please report any
+  issues you run into.
+
 ### Deprecations
 
 * The `ui.diff.format` and `ui.diff.tool` config options have been merged as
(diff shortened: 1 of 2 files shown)


thoughtpolice (approved):

cenviity:
Could we also make use of GitHub [milestones](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) to track which upcoming version each PR is targeting?

We would still have to make sure changelog entries are placed under the correct position when merging each PR, but at least the milestone is a good indicator/reminder of what that position should be.

I haven’t thought in depth about the logistics. One thing to do is, after each release, to close the associated milestone and move all unmerged PRs to a new milestone for the next version. This could be done via the GitHub web UI or `gh` CLI.

PhilipMetzger:
> Could we also make use of GitHub [milestones](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) to track which upcoming version each PR is targeting?
> 
> We would still have to make sure changelog entries are placed under the correct position when merging each PR, but at least the milestone is a good indicator/reminder of what that position should be.

@arxanas brought this up in the Discord a while ago, see [this message](https://discord.com/channels/968932220549103686/969291218347524238/1334279519472844840) and the surrounding discussion which certainly was inconclusive.

ilyagr:
Thanks Austin!

Re milestones, I think they'd mainly increase the complexity of doing releases. Syncing a changelog to a list of issues seems annoying and error-prone. Also, I like the reliability of keeping this information in a file.

It wouldn't be hard to add a milestone to a pull request while merging it, but I'm not sure how much benefit that would create on its own.

Another interesting way to compile a changelog is done by Restic, see item 7 in https://github.com/restic/restic/blob/master/CONTRIBUTING.md#providing-patches and <https://github.com/restic/restic/tree/master/changelog>. It relies on the fact that when renaming a dir from "unreleased" to "v1.XX", git tracks this as moving each file, and open PRs will still add files to "unreleased". However, IMO that might also be more trouble than it's worth.

cenviity:
> Re milestones, I think they'd mainly increase the complexity of doing releases. Syncing a changelog to a list of issues seems annoying and error-prone. Also, I like the reliability of keeping this information in a file.

To make it clear, I wasn’t proposing generating the changelog from the merged PRs in a milestone upon releasing a new version. Currently we only update the changelog on a case-by-case basis for changes that are genuinely worth mentioning, and I think it’s fine to keep it that way.

PR #6705, opened 2025-06-06
lib: fix a clippy lint caught by nightly

thoughtpolice:
<https://rust-lang.github.io/rust-clippy/master/index.html#cloned_ref_to_slice_refs>

# Checklist

If applicable:

### DIFF
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -2966,6 +2966,8 @@ pub fn format_remote_symbol(name: &str, remote: &str) -> String {
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::cloned_ref_to_slice_refs)]
+
     use std::path::PathBuf;
 
     use assert_matches::assert_matches;
--- a/lib/src/default_index/mod.rs
+++ b/lib/src/default_index/mod.rs
@@ -45,6 +45,7 @@ pub use self::store::DefaultIndexStoreError;
 pub use self::store::DefaultIndexStoreInitError;
 
 #[cfg(test)]
+#[allow(clippy::cloned_ref_to_slice_refs)]
 mod tests {
     use std::cmp::Reverse;
     use std::convert::Infallible;
--- a/lib/src/default_index/rev_walk.rs
+++ b/lib/src/default_index/rev_walk.rs
@@ -706,6 +706,7 @@ impl AncestorsBitSet {
 }
 
 #[cfg(test)]
+#[allow(clippy::cloned_ref_to_slice_refs)]
 mod tests {
     use itertools::Itertools as _;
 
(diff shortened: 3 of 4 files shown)


ilyagr:
IIRC, there are many more instances of this lint getting triggered in tests. 

For many of those, I'm not sure that fixing it is a good idea, they become less readable. 

So, I think we need to decide between: 

- fixing the lint nonetheless (but everywhere) 
- ignoring this lint globally
- ignoring it for tests only


WDYT?

ilyagr (changes requested):
See above. My personal decision was to add the lint to my `Cranky.toml` and to wait to see if it sticks. Perhaps we should find an issue on clippy forums, I didn't put time into that.

ilyagr:
IIRC, there are many more instances of this lint in tests. 

For many of those, I'm not sure that fixing it is a good idea, they become less readable. 

So, I think we need to decide between: 

- fixing the lint nonetheless (but everywhere) 
- ignoring this lint globally
- ignoring it for tests only


WDYT?

ilyagr:
OTOH, maybe I overreacted. if you prefer this PR to just fix the lint in a few spots where it's a clear improvement (like the spot in the current diff, I think?), that would be fine. We don't *have* to resolve the larger issue right now.

thoughtpolice:
I think it was mainly just because I ran `cargo clippy` without any extra flags so I didn't see the test failures; I guess you need `bacon` or `--all-targets`. TBH, that's good enough for me just to merge this.

But I'm also happy to turn the lint off for tests, frankly. I think that would be the best way forward.

ilyagr:
I forgot about the issue with older versions not knowing this lint. IIRC, there are a couple of crates that can add an `allow(unkown_lints)` that are conditional on the Rustc version. It's a bit awkward, and AFAICT requires changes to `build.rs`. There might be some better way to deal with it.

I, personally, use <https://github.com/ericseppanen/cargo-cranky> here to suppress the lint for now, but that's an unsatisfying and temporary solution.

We can still merge the easy fixes part of this PR, though I suspect that the lint will start bugging you again whenever you work on tests.

lib/src/revset.rs
```diff
@@ -2661,6 +2661,8 @@ pub fn format_remote_symbol(name: &str, remote: &str) -> String {
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::cloned_ref_to_slice_refs)]
```
martinvonz, line 2969: is this the same thing as putting the attribute without the `!` on the line above?

ilyagr:
BTW, `clap` seems to use the `rustversion` crate for something similar.

https://github.com/clap-rs/clap/blob/b69d0ee9ea6d7f09f56aa220cac5c2955ae74a2a/tests/derive_ui.rs#L11



I wonder whether something like `#[rustversion::attr(stable(1.89), expect(clippy::thislint)]` or `#[rustversion::attr(not(stable(1.89)), allow(unknown_lints)]` will work.

ilyagr:
`rustversion` seems to work, I'll make a PR in a bit.

**Update:** https://github.com/jj-vcs/jj/pull/6818

PR #6707, opened 2025-06-07
clippy: fix clippy error

06393993:
* ~Add `#[allow(clippy::result_large_err)]` for functions that return `Result<T, ConfigLoadError>`.~
* Change `std::os::windows::process::CommandExt` to underscore imports.
* ~Change `&[target_commit.clone()]` to `std::slice::from_ref(target_commit)`.~

I just silence the clippy error for the large error type `ConfigLoadError` instead of fixing it, because fixing can introduce breaking changes in the API: `ConfigLoadError` is a public type used in public methods, e.g., `ConfigFile::load_or_empty`. I am Ok to further update this PR to fix that clippy error, but I need more instructions on:
* Can I just create a new `#[derive(Error)] struct ConfigLoadError(#[error(transparent)] Box<ConfigLoadErrorImpl>)` type to fix the clippy error or it would be better if I dig in the enum variants and change fields to `Box` as needed?
* That's a breaking change. What's the complete list of actions I need to take for a breaking change? e.g., should I bump the version? should I mention that in the CHANGELOG.md?

# Checklist

If applicable:

- [x] I have updated `CHANGELOG.md`: N/A. clippy fix, no actual changes.
- [x] I have updated the documentation (`README.md`, `docs/`, `demos/`). clippy fix, no need for documentation update.
- [x] I have updated the config schema (`cli/src/config-schema.json`): N/A. No change to the config schema.
- [x] I have added/updated tests to cover my changes: N/A. No new logic added for testing.

### DIFF
--- a/lib/src/config.rs
+++ b/lib/src/config.rs
@@ -64,7 +64,7 @@ pub enum ConfigLoadError {
     Parse {
         /// Source error.
         #[source]
-        error: toml_edit::TomlError,
+        error: Box<toml_edit::TomlError>,
         /// Source file path.
         source_path: Option<PathBuf>,
     },
@@ -337,7 +337,7 @@ impl ConfigLayer {
     /// Parses TOML document `text` into new layer.
     pub fn parse(source: ConfigSource, text: &str) -> Result<Self, ConfigLoadError> {
         let data = ImDocument::parse(text).map_err(|error| ConfigLoadError::Parse {
-            error,
+            error: Box::new(error),
             source_path: None,
         })?;
         Ok(Self::with_data(source, data.into_mut()))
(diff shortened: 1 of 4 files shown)


cenviity:
> * Change `&[target_commit.clone()]` to `std::slice::from_ref(target_commit)`.

This fix is being discussed in #6705.

cli/src/config.rs
```diff
@@ -400,13 +400,15 @@ impl ConfigEnv {
     /// The parent directory for the new file may be created by this function.
     /// If the user configuration path is unknown, this function returns an
     /// empty `Vec`.
+    #[allow(clippy::result_large_err)]
```
yuja: Just for reminder: box `TomlError` instead

06393993: Done.

06393993:
> nit: I would split the two sets of fixes into separate commits as they are not related to each other.

Done.

ilyagr:
Thanks! We don't have as many people taking care of Windows-specific code, so maintaining it is especially nice. (I don't know if there's a way to trigger those lints on Linux/Mac)

Perhaps we should add a clippy-on-Windows CI check (perhaps an optional one).

This LGTM, but I'll wait for Yuya and Vincent to get a chance to look at it.

cenviity (approved):

yuja (approved):

PR #6715, opened 2025-06-07
status: Warn when no files match the given fileset

### LINKED ISSUE #6556: FR: `jj status` should throw an error when the fileset does not exist

When I run `jj diff stringthatmatchesnofiles`, I get

```
$ jj diff stringthatmatchesnofiles
Warning: No matching entries for paths: stringthatmatchesnofiles
```

This is helpful! It suggests that I may have made a mistake (often, specifying a change id without `-r`).

When I do the same for `jj status`, I get

```
$ jj st stringthatmatchesnofiles
Working copy changes:
Working copy  (@) : xxxxxxxx bbbbbbbb (no description set)
Parent commit (@-): yyyyyyyy aaaaaaaa (no description set)
```

It identifies that no strings match, and then prints an empty changelog.

Reproducer:
```
jj git init
jj diff x
jj st x
```

It would be more helpful to show a message that identifies the issue an suggests an alternative. `jj show` seems best: my initial reaction to `jj st` when I began was that if no arguments gave me the current change, maybe a different change ID would give me another change.

```
$ jj st x
Warning: No matching entries for paths: x
Maybe you want 'jj show x'?
```

There could also be a flag on `jj st` to cause this behavior, but I believe that's not helpful because anyone who knows of the flag is also likely to know how to properly use it to begin with. Then the error is not helpful as a discovery tool.

killpack:
Closes #6556. (It *was* a good first issue! 😄 )

If there's more than one unmatched path, the suggestion will only show the first, since I don't think it makes sense to suggest `jj show x y z`.

I have signed the CLA.

# Checklist

If applicable:


cli/src/commands/status.rs
```diff
@@ -162,6 +163,15 @@ pub(crate) fn cmd_status(
                 }
             }
         }
+        let unmatched_paths =
+            print_unmatched_explicit_paths(ui, &workspace_command, &fileset_expression, [&tree])?;
+        if let Some(first_path) = unmatched_paths.first() {
+            writeln!(
+                ui.warning_no_heading(),
```
cenviity: nit: I would make this `ui.hint_default()` as this is more of a suggestion to the user.

cli/src/commands/status.rs
```diff
@@ -162,6 +163,15 @@ pub(crate) fn cmd_status(
...
+                "Maybe you meant `jj show {path}`?",
```
yuja: Why `jj show {path}`?

If the intent is to help user who thinks `path` was a revset, `jj diff --summary -r {first_arg}` might be better. I also think it's better to copy the condition from `jj log`, in which we test that no multiple positional arguments are specified, and the only path argument looks like a revset.

killpack: Only because it was the command specified in the [feature request](https://github.com/jj-vcs/jj/issues/6556)-
>  jj show seems best: my initial reaction to jj st when I began was that if no arguments gave me the current change, maybe a different change ID would give me another change.

I don't have any particular strong feelings about `jj show`, though- happy to do whatever makes sense.

cli/src/commands/status.rs
killpack: Done.

cli/src/commands/status.rs
killpack: Went ahead and implemented showing `jj diff --summary -r {revision}`, and checking that there's only one arg and that it could be a revset.

cli/src/cli_util.rs
```diff
@@ -2824,19 +2824,19 @@ Discard the conflicting changes with `jj restore --from {}`.",
...
-pub fn print_unmatched_explicit_paths<'a>(
+/// entries, returning the unmatched paths.
+pub fn print_unmatched_explicit_paths<'a, 'expr>(
     ui: &Ui,
     workspace_command: &WorkspaceCommandHelper,
-    expression: &FilesetExpression,
+    expression: &'expr FilesetExpression,
     trees: impl IntoIterator<Item = &'a MergedTree>,
-) -> io::Result<()> {
+) -> io::Result<Vec<&'expr RepoPath>> {
```
yuja: Can you revert this change? It's odd that `print()` function returns a list of paths to be printed, and we can easily test the same condition by caller now.

cli/src/commands/status.rs
```diff
@@ -162,6 +164,21 @@ pub(crate) fn cmd_status(
...
+
+        if let [only_path] = args.paths.as_slice() {
+            if !unmatched_paths.is_empty()
+                && workspace_command
+                    .parse_revset(ui, &RevisionArg::from(only_path.to_owned()))
+                    .is_ok()
+            {
+                writeln!(
+                    ui.hint_default(),
+                    "Maybe you meant `jj diff --summary -r {only_path:?}`?",
```
yuja: nit: "To show changes in a certain revision, use `jj diff ...`" or something? It's not so obvious that the user expects `jj status` could show changes of arbitrary revision.

yuja (commented):
Can you also include the issue number in the commit message?

cli/src/cli_util.rs
killpack: Good point- not necessary anymore.

cli/src/commands/status.rs
```diff
@@ -162,6 +164,33 @@ pub(crate) fn cmd_status(
                 }
             }
         }
+
+        print_unmatched_explicit_paths(ui, &workspace_command, &fileset_expression, [&tree])?;
```
yuja: `[&parent_tree, &tree]` because `jj status` shows diffs between these trees.

cli/src/commands/status.rs
```diff
@@ -162,6 +164,33 @@ pub(crate) fn cmd_status(
...
+
+        if let [only_path] = args.paths.as_slice() {
+            let is_unmatched_path = fileset_expression
+                .explicit_paths()
+                .next()
+                .map(|repo_path| {
```
yuja: nit: `.is_some_and()`

cli/src/commands/status.rs
```diff
@@ -162,6 +164,33 @@ pub(crate) fn cmd_status(
...
+                    .parse_revset(ui, &RevisionArg::from(only_path.to_owned()))
+                    .is_ok()
+            {
+                writeln!(
+                    ui.hint_default(),
+                    "To show changes in a given revision, use `jj diff --summary -r \
+                     {only_path:?}`.",
+                )?;
+            };
+        };
```
yuja: nit: remove `;`s

yuja (approved):
Thanks.

thoughtpolice (approved):

cli/src/commands/status.rs
```diff
@@ -162,6 +164,37 @@ pub(crate) fn cmd_status(
...
+
+        if let [only_path] = args.paths.as_slice() {
+            let is_unmatched_path =
+                fileset_expression
+                    .explicit_paths()
+                    .next()
+                    .is_some_and(|repo_path| {
+                        tree.path_value(repo_path)
+                            .ok()
+                            .is_some_and(|p| p.is_absent())
```
yuja: nit: `parent_tree.path_value(repo_path).is_ok_and(|p| p.is_absent()) && ..` for consistency.

martinvonz:
I've invited @killpack to join the contributors team now

cli/src/commands/status.rs
killpack: While we're at it, I noticed that `print_unmatched_explicit_paths` [iterates over all the trees](https://github.com/jj-vcs/jj/blob/e72f1611893d2eecd34bf11898b25c2f86f3322b/cli/src/cli_util.rs#L2826-L2839)- we should probably do that here, too, right?

yuja: If you prefer `[&parent_tree, &tree].iter().all(..)`. feel free to do that.

whereistejas:
ping @killpack: do you plan on merging this PR? If not, I would love to give it a try :)

killpack:
@whereistejas ah, apologies, had some major life things happen and forgot about this. I'll try to get this merged ASAP.

killpack:
Actually @whereistejas - I am not sure when I'm going to be able to get to this- if you want to take over rebasing this and getting it shipped, that'd be amazing :)

whereistejas:
> Actually @whereistejas - I am not sure when I'm going to be able to get to this- if you want to take over rebasing this and getting it shipped, that'd be amazing :)

Sounds good! I can take over :) thanks for letting me know

Issue #6690, opened 2025-06-04
Update partial git clone docs


gmorenz:
The [git-compatibility docs](https://github.com/jj-vcs/jj/blob/main/docs/git-compatibility.md?plain=1#L62) still say

> * **Partial clones: No.** We use the [libgit2](https://libgit2.org/) library, which [doesn't have support for partial clones](https://github.com/libgit2/libgit2/issues/5564).

This is presumably no longer correct since libgit2 is no longer a dependency. I see https://github.com/jj-vcs/jj/pull/6451 so I presume partial clones aren't fully supported, but I'm hesitant to edit this myself since I'm not sure if they might partially work (and the other comments in there have some nuance).

@anguslees any chance you can tell us if they work at all on mainline jj?

anguslees:
They don't work on mainline jj.  Gitoxide currently lacks support for promisor packs, and so won't download missing objects.

My dodgy patch just calls out to git to do that at all the right points. This part works, but it can create many .pack files suddenly (just like git itself can when using partial clones), which explodes a part of gitoxide that uses an immutable slotmap data structure and assumes number of pack files stays much smaller.

I use my dodgy jj PR every day, for everything I do, fwiw.  After I do a fresh `jj git fetch`, I switch to the newly updated master (`jj new master`) which triggers a bunch of objects to get downloaded and often aborts with the above "no more slots" error - leaving random modified files from the colocated git repo.  Then I do a `git reset --hard; git clean -d -f` to reset the colocated git clone back to jj's idea of git_head() (currently master), and then I'm good to switch to some other jj change and do what I actually wanted.  It's a small hassle and probably exposing a bug in jj's error handling, but much better than dealing with a full git clone of our monster repo.

Oh, and the many pack files also seems to slow down jj/gitoxide.  I'm guessing it also lacks one of the newer git indices introduced for this problem.  So run `git gc` occasionally.

PR #6717, opened 2025-06-08
Don't blame libgit2 for lack of partial clone support

gmorenz:
`jj` no longer uses libgit2, so the lack of support can't be blamed on it. We [still don't support partial clones](https://github.com/jj-vcs/jj/issues/6690#issuecomment-2946785320) though.

There's no real tracking issue for this. [This PR](https://github.com/jj-vcs/jj/pull/6451/) is as close as it gets, but it doesn't feel appropriate to link in the docs.

PS. Thanks for the detailed reply to my question @anguslees

### DIFF
--- a/docs/git-compatibility.md
+++ b/docs/git-compatibility.md
@@ -59,8 +59,7 @@ a comparison with Git, including how workflows are different, see the
   create a repo backed by a bare Git repo.
 * **Submodules: No.** They will not show up in the working copy, but they will
   not be lost either.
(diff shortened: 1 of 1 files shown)


ilyagr (approved):
Thank you!

gmorenz:
I feel pretty safe in saying that the CI failure - `test_concurrent_read_write_commit`  - is the result of a flaky test and not editing the documentation.

~~I assume this blocks merging... do I need to do something to trigger the tests to run again?~~

PR #6719, opened 2025-06-08
revset: do not reinterpret `set&filter` intersection as filter

### LINKED ISSUE #6713: `jj file annotate` doesn't work on hidden commits
## Description

If the `--revision` points to a hidden commit, all lines originate at `root()`.

Given for example `jj`'s repository:

1.  `jj file annotate README.md | head -n3`

        plulkmrl aseipp   2023-10-07 18:35:06    1: <div class="title-block" style="text-align: center;" align="center">
        plulkmrl aseipp   2023-10-07 18:35:06    2: 
        plulkmrl aseipp   2023-10-07 18:35:06    3: # Jujutsu—a version control system

2.  `touch random-file    # any command that hides the previous @`
3.  `jj file annotate -r 'at_operation(@-, @)' README.md | head -n3`

        zzzzzzzz          1969-12-31 21:00:00    1: <div class="title-block" style="text-align: center;" align="center">
        zzzzzzzz          1969-12-31 21:00:00    2: 
        zzzzzzzz          1969-12-31 21:00:00    3: # Jujutsu—a version control system

Note: this particular example could of course be worked around with `--at-op=@-`, but my original use case was analyzing a revision identified by `evolog`.

## Expected Behavior

A certain commit being hidden shouldn't prevent traversing its parents, so it should produce the exact same output in this case

## Actual Behavior

If `-r` is given a hidden commit, all lines map to `zzzzzzz`.

## Specifications

- Platform: macOS
- Version: 0.30.0

yuja:
Fixes #6713

# Checklist

If applicable:

cc @scott2000 as this might have conflicts with your patch.

### DIFF
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -1525,7 +1525,13 @@ fn internalize_filter<St: ExpressionState>(
     }
 
     fn is_filter_tree<St: ExpressionState>(expression: &RevsetExpression<St>) -> bool {
-        is_filter(expression) || as_filter_intersection(expression).is_some()
+        if let RevsetExpression::Intersection(expression1, _) = expression {
+            // 'f1 & f2' can't be evaluated by itself, but 's1 & f2' can be
+            // filtered within 's1'. 'f1 & s2' should have been reordered.
+            is_filter(expression1)
+        } else {
+            is_filter(expression)
+        }
     }
 
     // Extracts 'c & f' from intersect_down()-ed node.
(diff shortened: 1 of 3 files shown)


lib/src/revset.rs
```diff
@@ -1527,7 +1527,13 @@ fn internalize_filter<St: ExpressionState>(
     }
 
     fn is_filter_tree<St: ExpressionState>(expression: &RevsetExpression<St>) -> bool {
-        is_filter(expression) || as_filter_intersection(expression).is_some()
+        if let RevsetExpression::Intersection(expression1, _) = expression {
```
scott2000, line 1528: After #6720, maybe we could transform filter-only intersections like `f1 & f2 & f3` to `AsFilter(f1 & f2 & f3)`, and then we could just use `is_filter` instead of `is_filter_tree` to avoid checking the entire intersection every time?

yuja: Could be, but I'm not sure if that's better. Pseudo nodes like `AsFilter` could make later transformation a bit complicated. It might be better to try to remove `AsFilter` at all, by sorting union in a similar way to #6720, etc. I have no idea if that's feasible, though.

thoughtpolice (approved):

PR #6721, opened 2025-06-08
revset: parse `x..` as `~ancestors(x)`

scott2000:
Currently, this is parsed as `x..visible_heads()`, which prevents optimizing `x.. & ::y` to `x..y`.

An alternative would be to remove `::visible_heads()` from intersections, similar to `all()`, but it might be better to avoid removing `visible_heads()` if it was explictly requested by a user.

# Checklist

If applicable:


lib/src/revset.rs
```diff
@@ -1063,17 +1063,15 @@ pub fn lower_expression(
             ))
         }
         ExpressionKind::DagRangeAll => Ok(RevsetExpression::all()),
-        ExpressionKind::RangeAll => {
-            Ok(RevsetExpression::root().range(&RevsetExpression::visible_heads()))
-        }
+        ExpressionKind::RangeAll => Ok(RevsetExpression::root().ancestors().negated()),
```
yuja: nit: simply map to `root().negated()`?

yuja (approved):

lib/src/revset.rs
```diff
@@ -1063,17 +1063,15 @@ pub fn lower_expression(
...
         ExpressionKind::Unary(op, arg_node) => {
             let arg = lower_expression(diagnostics, arg_node, context)?;
             match op {
                 UnaryOp::Negate => Ok(arg.negated()),
                 UnaryOp::DagRangePre => Ok(arg.ancestors()),
                 UnaryOp::DagRangePost => Ok(arg.descendants()),
                 UnaryOp::RangePre => Ok(RevsetExpression::root().range(&arg)),
-                UnaryOp::RangePost => Ok(arg.range(&RevsetExpression::visible_heads())),
+                UnaryOp::RangePost => Ok(arg.ancestors().negated()),
```
cenviity, line 1074: Also update the `x..` entry in `revsets.md`.

lib/src/revset.rs
cenviity: Also update the `..` entry in `revsets.md`.

scott2000: This entry already mentions that it is equivalent to `~root()`, so I don't think it would be necessary to update it. This is more of an implementation detail; it's still generally true that it's a shorthand for `root()..visible_heads()`.

lib/src/revset.rs
scott2000: I added a note that it is equivalent to `~::x`. I left the explanation of it being a shorthand for `x..visible_heads()` since that is still generally true, and it seems more understandable to me to explain it in terms of the `x..y` operator since they are visually similar. Do you think this is good?

PR #6722, opened 2025-06-08
docs and uv: update mkdocs-material and other docs dependencies

ilyagr:
mkdocs-material seems to have been in bugfix mode for a bit, so let's try out the latest version. If something breaks in the prerelease docs, we can revert.


thoughtpolice (approved):

thoughtpolice:
> If something breaks in the prerelease docs, we can revert.

Ideally, more bugs are fixed and we don't have to :)

PR #6725, opened 2025-06-08
split: keep a single entry in the evolog

glehmann:
This is an alternative implementation of #6661 that continues to rewrite the commit multiple times, but manipulates the commit predecessors in order to keep a single entry for the split in the evolog.
Plus it's much simpler.

# Checklist

If applicable:

### DIFF
--- a/lib/src/rewrite.rs
+++ b/lib/src/rewrite.rs
@@ -313,7 +313,19 @@ pub fn rebase_commit_with_options(
         _ => None,
     };
     let new_parents_len = rewriter.new_parents.len();
-    if let Some(builder) = rewriter.rebase_with_empty_behavior(options.empty)? {
+    let predecessors = if options.preserve_predecessors {
+        rewriter
+            .mut_repo
+            .predecessors()
+            .get(rewriter.old_commit.id())
+            .cloned()
+    } else {
+        None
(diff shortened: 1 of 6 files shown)


lib/src/rewrite.rs
```diff
@@ -318,7 +318,11 @@ pub fn rebase_commit_with_options(
         _ => None,
     };
     let new_parents_len = rewriter.new_parents.len();
-    if let Some(builder) = rewriter.rebase_with_empty_behavior(options.empty)? {
+    let predecessors = rewriter.old_commit().store_commit().predecessors.clone();
```
yuja: Perhaps, we should copy predecessors from `mut_repo.commit_predecessors[old_commit]`. It wouldn't make sense to preserve predecessors across operations.

glehmann: you're right I focused on the double entry, but the rewritten commits should have an entry for the split.

glehmann: in fact, I can't think of a situation where we would want the multiple evolog entries when using `move_commits`, so I'm not even sure we need the `preserve_predecessors` option.

glehmann:
We'll have a better option based on #6757

PR #6730, opened 2025-06-09
Add Send bounds to the AsyncRead future returned by read_file

benbrittain:
This is closer aligned to the type signature for `write_file` as well, which is currently:

```
    async fn write_file(
        &self,
        path: &RepoPath,
        contents: &mut (dyn AsyncRead + Send + Unpin),
    ) -> BackendResult<FileId>;
```

This change enables the jj-lib api to be used in multi-threaded executor contexts.

# Checklist

If applicable:

### DIFF
--- a/cli/examples/custom-backend/main.rs
+++ b/cli/examples/custom-backend/main.rs
@@ -153,7 +153,7 @@ impl Backend for JitBackend {
         &self,
         path: &RepoPath,
         id: &FileId,
-    ) -> BackendResult<Pin<Box<dyn AsyncRead>>> {
+    ) -> BackendResult<Pin<Box<dyn AsyncRead + Send>>> {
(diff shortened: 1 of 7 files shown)


thoughtpolice:
We had a chat about this and it seems sensible to me (in short: the read path for any backend must be thread safe). I can't approve this, though, and other people probably have thoughts, too.

ilyagr:
We might want to wait for somebody from Google to confirm this works with their backend.

PhilipMetzger:
Can you prefix the commit with `store:` to conform to our guidelines and the point Martin mentioned in the Discord.

ilyagr (approved):
Approving based on the Discord discussion. Thanks for fixing up the commit message as well.

PR #6735, opened 2025-06-10
revset: add `All` filter predicate

scott2000:
Useful for optimizing the filter in #6679.

This allows `all()` to be efficiently represented as a filter predicate, and I think having a way to represent a predicate that always returns true makes the language more consistent.

# Checklist

If applicable:

### DIFF
--- a/lib/src/default_index/revset_engine.rs
+++ b/lib/src/default_index/revset_engine.rs
@@ -1144,6 +1144,7 @@ fn build_predicate_fn(
     predicate: &RevsetFilterPredicate,
 ) -> Box<dyn ToPredicateFn> {
     match predicate {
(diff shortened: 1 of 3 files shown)


yuja:
Maybe we can use `predicate: Option<_>` to represent no filtering? If we're strict about naming, there will be `ResolvedExpression::HeadsRange { roots, heads }` and `ResolvedExpression::FilterWithinHeadsRange { roots, heads, predicate }`. If that seemed too verbose, using `Option<_>` would be good.

Since filter predicates aren't primary expressions, I think it's better to optimize `all()`/`none()` out. Oh, and I think the comment in `fold_redundant_expression()` is obsolete. User symbols are resolved earlier, so removing `x & none()` is probably okay.

PR #6741, opened 2025-06-11
revset: optimizations for intersections and ranges

scott2000:
Split off from #6679.

# Checklist

If applicable:

### DIFF
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -1506,6 +1506,66 @@ where
     Ok(expression)
 }
 
+/// Flatten all intersections to be left-recursive. For instance, transforms
+/// `(a & b) & (c & d)` into `((a & b) & c) & d`.
+fn flatten_intersections<St: ExpressionState>(
+    expression: &Rc<RevsetExpression<St>>,
+) -> TransformedExpression<St> {
+    fn flatten<St: ExpressionState>(
+        expression1: &Rc<RevsetExpression<St>>,
+        expression2: &Rc<RevsetExpression<St>>,
+    ) -> TransformedExpression<St> {
+        let recurse = |a, b| flatten(a, b).unwrap_or_else(|| a.intersection(b));
+
+        match expression2.as_ref() {
+            // flatten(a & (b & c)) -> flatten(a & b) & c
+            RevsetExpression::Intersection(inner1, inner2) => {
+                Some(recurse(expression1, inner1).intersection(inner2))
+            }
+            _ => None,
+        }
+    }
+
+    transform_expression_bottom_up(expression, |expression| match expression.as_ref() {
+        RevsetExpression::Intersection(expression1, expression2) => {
+            flatten(expression1, expression2)
+        }
+        _ => None,
+    })
+}
+
+/// Intersects `expression` with `base`, maintaining sorted order using the
+/// provided key. If `base` is an intersection, it must be left-recursive, and
+/// it must already be in sorted order.
+fn sort_intersection_by_key<St: ExpressionState, T: Ord>(
+    base: &Rc<RevsetExpression<St>>,
+    expression: &Rc<RevsetExpression<St>>,
+    mut get_key: impl FnMut(&RevsetExpression<St>) -> T,
+) -> TransformedExpression<St> {
+    // We only want to compute the key for `expression` once instead of computing it
+    // on every iteration.
+    fn sort_intersection_helper<St: ExpressionState, T: Ord>(
+        base: &Rc<RevsetExpression<St>>,
+        expression: &Rc<RevsetExpression<St>>,
+        expression_key: T,
+        mut get_key: impl FnMut(&RevsetExpression<St>) -> T,
+    ) -> TransformedExpression<St> {
+        if let RevsetExpression::Intersection(inner1, inner2) = base.as_ref() {
+            // sort_intersection(a & b, c) -> sort_intersection(a, c) & b
+            (expression_key < get_key(inner2)).then(|| {
+                sort_intersection_helper(inner1, expression, expression_key, get_key)
+                    .unwrap_or_else(|| inner1.intersection(expression))
+                    .intersection(inner2)
+            })
+        } else {
+            // a & b -> b & a
+            (expression_key < get_key(base)).then(|| expression.intersection(base))
+        }
+    }
+
+    sort_intersection_helper(base, expression, get_key(expression), get_key)
+}
+
 /// Transforms filter expressions, by applying the following rules.
 ///
 /// a. Moves as many sets to left of filter intersection as possible, to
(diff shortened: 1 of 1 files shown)


lib/src/revset.rs
```diff
@@ -1506,6 +1506,68 @@ where
...
+) -> TransformedExpression<St> {
+    fn flatten<St: ExpressionState>(
+        expression1: &Rc<RevsetExpression<St>>,
+        expression2: &Rc<RevsetExpression<St>>,
+    ) -> TransformedExpression<St> {
+        match expression2.as_ref() {
+            // flatten(a & (b & c)) -> flatten(a & b) & c
+            RevsetExpression::Intersection(inner1, inner2) => Some(
+                flatten(expression1, inner1)
+                    .unwrap_or_else(|| expression1.intersection(inner1))
```
yuja: nit: I slightly prefer extracting this `f(e1, e2).unwrap_or_else(|| e1.g(e2))` part to closure because `.unwrap_or_else(||` looked a bit noisy.

lib/src/revset.rs
```diff
@@ -1506,6 +1506,68 @@ where
...
+    }
+}
+
+/// Assuming `expression1` is already sorted properly (and is left-recursive),
+/// intersects `expression2` while maintaining sorted order. Sort order is
+/// specified by a function returning true if two elements need to be swapped.
+fn sort_intersection<St: ExpressionState>(
+    expression1: &Rc<RevsetExpression<St>>,
+    expression2: &Rc<RevsetExpression<St>>,
+    mut should_swap: impl FnMut(&RevsetExpression<St>, &RevsetExpression<St>) -> bool,
```
yuja: nit: maybe we can instead add `sort_intersection_by_key(e1, e2, f(x) -> T), T: Ord`? It seems a bit odd that the compare callback doesn't follow `|a, b| a < b`.

lib/src/revset.rs
```diff
@@ -1677,17 +1711,19 @@ fn fold_difference<St: ExpressionState>(
...
         RevsetExpression::Intersection(expression1, expression2) => {
-            match (expression1.as_ref(), expression2.as_ref()) {
-                // For '~x & f', don't move filter node 'f' left
-                (_, RevsetExpression::Filter(_) | RevsetExpression::AsFilter(_)) => None,
-                (_, RevsetExpression::NotIn(complement)) => {
-                    Some(to_difference(expression1, complement))
-                }
-                (RevsetExpression::NotIn(complement), _) => {
-                    Some(to_difference(expression2, complement))
+            transform_intersection_last(expression1, |expression1| {
```
yuja: I don't fully understand why the last node is picked, (or the original order is preferred.)

I think this is a workaround for `a..b & c..d`, but it will be transformed differently at the last patch. The added test `~::a & (::b & ::c) ~ ::d` can be caught by using `transform_intersection_last()`, but `~::a & ~::b & ::c & ::d` wouldn't.

lib/src/revset.rs
```diff
@@ -1697,6 +1710,64 @@ fn to_difference_range<St: ExpressionState>(
...
+                        generation: g1,
+                    },
+                    RevsetExpression::Ancestors {
+                        heads: h2,
+                        generation: g2,
+                    },
+                ) if g1 == g2 => Some(
+                    union_ancestors(h1, h2)
+                        .unwrap_or_else(|| h1.union(h2))
+                        .ancestors_range(g1.clone()),
```
yuja: Just question. Does this process have to be recursive because `fold_generation()` isn't applied yet?

lib/src/revset.rs
```diff
@@ -1697,6 +1710,64 @@ fn to_difference_range<St: ExpressionState>(
...
+    expression: &Rc<RevsetExpression<St>>,
+) -> TransformedExpression<St> {
+    fn union_ancestors<St: ExpressionState>(
+        expression1: &Rc<RevsetExpression<St>>,
+        expression2: &Rc<RevsetExpression<St>>,
+    ) -> TransformedExpression<St> {
+        // If `expression1` is a union created from an intersection of negations, it
+        // will be left-recursive, so we will want to try to combine with the last
+        // expression in the union.
+        transform_union_last(expression1, |expression1| {
```
yuja: For the same reason as `transform_intersection_last()`, I'm not pretty sure why we choose the last node.

(My understanding was that we sort the expressions so that we can transform expressions from the bottom.)

lib/src/revset.rs
scott2000: Yeah, in an earlier version of the PR I was intending it to handle both `::a & ~::b` => `b..a` and `(a & ::b) & ~::c` => `a & c..b`, but now it's unnecessary I think.

lib/src/revset.rs
scott2000: Yeah, that was the idea, so that `~::a- & ~::b-` => `~(::a- | ::b-)` => `~::(a- | b-)` => `~::(a | b)-`.

lib/src/revset.rs
scott2000: Yeah, I think this one should also be unnecessary now since all `~::x` are moved to the start.

lib/src/revset.rs
yuja: It might be better to do `fold_generation()` first, then split `ancestors(h1, g1..) | ancestors(h2, g2..)` at `min(g1, g2)`. It will be a bit more involved, but should be less order-dependent.

scott2000: Hmm yeah that might be good. I think also `fold_heads_range()` required moving `fold_generation()` later, but maybe I could extract a helper function for splitting ancestors and then use it there as well?

scott2000: I extracted a function for splitting ancestors, but I only implemented it for ancestors with a generation like `x..u64::MAX` because I think this should be sufficient, and it was simpler than making it fully general. Since `to_difference_range()` had the same logic, I was also able to convert it to use this function, as well as `fold_heads_range()` in my other PR.

I also changed this pass so that it no longer folds `~x & ~y` into `~(x | y)` in general, because that was rewriting `x.. ~ y` into `~(::x | y)`, which seems undesirable since it would scan all of `::visible_heads()` instead of just `x..visible_heads()`. So now it only folds ancestors expressions and nothing else, which also makes it simpler I think.

I handled the `~x ~ y & z` case by sorting non-ancestors negations to the end of the intersection, so that this can be optimized as `z ~ x ~ y` (whereas in the previous version of this PR, it would've been optimized as `z ~ (x | y)`, but now this would no longer work due to the other change I made).

Let me know what you think.

lib/src/revset.rs
```diff
@@ -1654,41 +1654,77 @@ fn fold_redundant_expression<St: ExpressionState>(
     })
 }
 
+/// Extracts `heads` from a revset expression `ancestors(heads)`. Unfolds
+/// generations as necessary, so `ancestors(heads, 2..)` would return
+/// `ancestors(heads, 2..3)`, which is equivalent to `heads--`.
+fn ancestors_to_heads<St: ExpressionState>(
+    expression: &RevsetExpression<St>,
+) -> Option<Rc<RevsetExpression<St>>> {
```
yuja: nit: Did you intentionally avoid using `TransformedExpression<St>` here? That's understandable because this function doesn't return an equivalent expression, but it seems also a bit weird since the caller just propagate `None` by `?` operator.

To encode the semantics, maybe we can return `Result<_, ()>`, and let caller use `.ok()?`?

lib/src/revset.rs
```diff
@@ -1654,41 +1654,77 @@ fn fold_redundant_expression<St: ExpressionState>(
...
+
+    transform_expression_bottom_up(expression, |expression| match expression.as_ref() {
+        RevsetExpression::Union(expression1, expression2) => {
+            // ::x | ::y -> ::(x | y)
+            union_ancestors(expression1, expression2)
+        }
+        RevsetExpression::Intersection(expression1, expression2) => {
+            match (expression1.as_ref(), expression2.as_ref()) {
+                // ~::x & ~::y -> ~(::x | ::y) -> ~::(x | y)
+                (RevsetExpression::NotIn(candidates1), RevsetExpression::NotIn(candidates2)) => {
```
yuja: nit: we typically call negated set as `complement`, not `candidates`.

lib/src/revset.rs
```diff
@@ -1566,6 +1566,42 @@ fn sort_intersection_by_key<St: ExpressionState, T: Ord>(
...
+/// All `~ancestors(x)` will be moved before `ancestors(x)`, since negated
+/// ancestors can be converted to ranges. All other negations are moved to the
+/// right, since these negations can usually be evaluated better as differences.
+fn sort_negations_and_ancestors<St: ExpressionState>(
+    expression: &Rc<RevsetExpression<St>>,
+) -> TransformedExpression<St> {
+    transform_expression_bottom_up(expression, |expression| match expression.as_ref() {
+        RevsetExpression::Intersection(expression1, expression2) => {
+            #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+            enum AncestorsOrder {
```
yuja: nit: let's move type definition to the (outer) function scope, so the match arms look concise.

yuja (approved):
Thanks!

lib/src/revset.rs
```diff
@@ -1534,44 +1594,6 @@ fn internalize_filter<St: ExpressionState>(
...
-                // ((c1 & f1) & g1) & e2 -> ((c1 & f1) & e2) & g1 -> ((c1 & e2) & f1) & g1
-                (Some((c1, f1)), _) => Some(recurse(c1, expression2).intersection(f1)),
-                (None, None) => None,
-            },
-        }
-    }
-
     // Bottom-up pass pulls up-right filter node from leaf '(c & f) & e' ->
     // '(c & e) & f', so that an intersection of filter node can be found as
     // a direct child of another intersection node. However, the rewritten
```
yuja: nit: can you update this comment? Perhaps, the second sentence can be removed, and the comment can be moved to `RevsetExpression::Intersection(_, _)`.

lib/src/revset.rs
scott2000: I actually hadn't considered using `TransformedExpression<St>` even. But yeah I agree, I think `Result` would be better.

lib/src/revset.rs
```diff
@@ -1586,8 +1603,11 @@ fn internalize_filter<St: ExpressionState>(
         }
         RevsetExpression::Union(e1, e2) => (is_filter_tree(e1) || is_filter_tree(e2))
             .then(|| Rc::new(RevsetExpression::AsFilter(expression.clone()))),
+        // Bottom-up pass pulls up-right filter node from leaf '(c & f) & e' ->
+        // '(c & e) & f', so that an intersection of filter node can be found as
+        // a direct child of another intersection node.
         RevsetExpression::Intersection(expression1, expression2) => {
-            intersect_down(expression1, expression2)
+            sort_intersection_by_key(expression1, expression2, is_filter)
```
yuja, line 1646: (Just a comment. No action requested.)

I might have to reintroduce `intersect_down()` to fix handling of nested filter intersections. If we add `f1 & f2 -> as_filter(f1 & f2)`, it would probably make sense to rewrite `s1 & f2 & f3` to `s1 & as_filter(f2 & f3)` rather than `(s1 & f2) & f3`, because `f2 & f3 & s1` would be transformed to `s1 & as_filter(f2 & f3)`.

scott2000: Yeah, I had been thinking about that actually, that sounds good.

Issue #6738, opened 2025-06-10
Avoid calling `Path::to_str`


imp:
While learning the `jj` codebase I noticed that there are some places where `Path::to_str()` is called to convert `Path` to string. Unfortunately it returns `Option<&str>` to protect against non-UTF8 paths, so it needs to be either `.unwrap()`-ed or converted to error in order to proceed. In majority of the cases, however, calling `Path::display()` would be good enough, as the purpose is to show these to use user or write it to the file where it takes `&[u8]` rather than strict UTF8 string anyway.

Would it be ok to propose a PR that addresses that? I mean convert calls to `Path::to_str` to non-fallible alternatives?

If so I am willing to do this work - I think it'll be a great practical exercise for a new contributor.

thoughtpolice:
I think we'd need to see the patch in question, but it sounds good! Go ahead :) Removing unwraps and extra error values seems like a nice touch up if we can do it, IMO.

PhilipMetzger:
There's also the discussion of using Camino if we want to strictly enforce `UFT-8`.

imp:
`camino` seems to be a good choice, however it feels like a larger scale effort. I wouldn't mind first spending some on de-`to_str()`-ification, just to see how that is going. Unless of course there is a strong reason not to do it.

PhilipMetzger:
I think your idea still is worthwhile because I currently think strictly enforcing `UTF-8` won't do us any favors. Maybe it'll come up on your PR though.

yuja:
I think some of `path.to_str()` (e.g. those in `RepoPathUiConverter`) are wrong, and should be switched to `.to_string()`. Some (e.g. in `GitBackend::init_with_repo()`) should be an error (or should use bytes API on Unix.) In tests, just use `.to_str().unwrap()` until we migrate to `camino`?

In any case, I wouldn't suggest making a big patch that replaces all of them in bulk. Small patches are nice!

imp:
Yeah, completely agree about usage in tests - no reason to touch those. I'll try to group the patches according to change nature.

PhilipMetzger:
I'm closing this as #6746 has landed

PR #6746, opened 2025-06-12
6738 path display

imp:
There are also quite a few instances of calling `Path::to_str()` in the test, where they can be just dropped (where the expected type of argument is `impl AsRef<Path>`, but as discussed in the ticket (#6738) I didn't touch those. If necessary this can be done as additional PR.

I used clippy lint `disallowed-methods = ["std::path::Path::to_str"]` to spot the usage of the `to_str` method. And I think there a couple of places where I left them untouched (besides tests) - I didn't feel that the change there would be appropriate or at least it won't be non-controversial.

Anyway, feel free to review and comment.

# Checklist

If applicable:

### DIFF
--- a/lib/src/repo_path.rs
+++ b/lib/src/repo_path.rs
@@ -602,9 +602,8 @@ impl RepoPathUiConverter {
         match self {
             RepoPathUiConverter::Fs { cwd, base } => {
                 file_util::relative_path(cwd, &file.to_fs_path_unchecked(base))
-                    .to_str()
-                    .unwrap()
-                    .to_owned()
+                    .display()
+                    .to_string()
             }
         }
     }
@@ -649,7 +648,7 @@ impl RepoPathUiConverter {
                     .min(target_components.len().saturating_sub(1));
 
                 fn format_components(c: &[std::path::Component]) -> String {
-                    c.iter().collect::<PathBuf>().to_str().unwrap().to_owned()
+                    c.iter().collect::<PathBuf>().display().to_string()
                 }
 
                 if prefix_count > 0 {


cli/src/commands/root.rs
```diff
@@ -31,11 +30,7 @@ pub(crate) fn cmd_root(
     command: &CommandHelper,
     RootArgs {}: &RootArgs,
 ) -> Result<(), CommandError> {
-    let root = command
-        .workspace_loader()?
-        .workspace_root()
-        .to_str()
-        .ok_or_else(|| user_error("The workspace root is not valid UTF-8"))?;
+    let root = command.workspace_loader()?.workspace_root().display();
```
yuja: These ones are a bit unclear. Since `jj root` output can be passed to other commands as input, lossy data can be worse than error. A better option is to write raw bytes on Unix.
https://doc.rust-lang.org/stable/std/os/unix/ffi/trait.OsStrExt.html#tymethod.as_bytes

cli/src/commands/sparse/edit.rs
```diff
@@ -55,13 +54,8 @@ fn edit_sparse(
...
         let workspace_relative_sparse_path = sparse_path.to_fs_path_unchecked(Path::new(""));
-        let path_string = workspace_relative_sparse_path.to_str().ok_or_else(|| {
-            internal_error(format!(
-                "Stored sparse path is not valid utf-8: {}",
-                workspace_relative_sparse_path.display()
-            ))
-        })?;
-        writeln!(&mut content, "{path_string}").unwrap();
+        let path = workspace_relative_sparse_path.display();
+        writeln!(&mut content, "{path}").unwrap();
```
yuja: Data loss here is unwanted. Just error out.

cli/src/progress.rs
```diff
@@ -52,9 +52,11 @@ pub fn snapshot_progress(ui: &Ui) -> Option<impl Fn(&RepoPath) + use<>> {
...
         let line_width = state.output.term_width().map(usize::from).unwrap_or(80);
         let max_path_width = line_width.saturating_sub(13); // Account for "Snapshotting "
-        let fs_path = path.to_fs_path_unchecked(Path::new(""));
-        let (display_path, _) =
-            text_util::elide_start(fs_path.to_str().unwrap(), "...", max_path_width);
+        let path = path
+            .to_fs_path_unchecked(Path::new(""))
+            .display()
+            .to_string();
+        let (display_path, _) = text_util::elide_start(&path, "...", max_path_width);
```
yuja: Since `RepoPath` is a string wrapper, and `Path::new("")` should never contain non-UTF-8 bytes, `.to_str()` shouldn't fail. Just `.unwrap()` here.

cli/src/diff_util.rs
```diff
@@ -1432,11 +1432,15 @@ pub async fn show_file_by_file_diff(
         let right_path = create_file(right_path, &right_wc_dir, right_value)?;
         let patterns = &maplit::hashmap! {
             "left" => left_path
-                .strip_prefix(temp_dir.path()).expect("path should be relative to temp_dir")
-                .to_str().expect("temp_dir should be valid utf-8"),
+                .strip_prefix(temp_dir.path())
+                .expect("path should be relative to temp_dir")
```
yuja: As commented, this should never fail. Just do `.expect()`.

lib/src/git_backend.rs
```diff
@@ -290,14 +291,14 @@ impl GitBackend {
...
             let git_repo_path_string = git_repo_path
-                .components()
-                .map(|component| component.as_os_str().to_str().unwrap().to_owned())
-                .join("/");
+                .display()
+                .to_string()
+                .replace(path::MAIN_SEPARATOR, "/");
             fs::write(&target_path, git_repo_path_string.as_bytes())
                 .context(&target_path)
                 .map_err(GitBackendInitError::Path)?;
```
yuja: No. Lossy conversion here is wrong.

lib/src/local_working_copy.rs
```diff
@@ -1623,13 +1623,8 @@ impl FileSnapshotter<'_> {
...
                 err: err.into(),
             })?;
-            let str_target =
-                target
-                    .to_str()
-                    .ok_or_else(|| SnapshotError::InvalidUtf8SymlinkTarget {
-                        path: disk_path.to_path_buf(),
-                    })?;
-            Ok(self.store().write_symlink(path, str_target).await?)
+            let contents = target.display().to_string();
```
yuja: Same here. Use lossy conversion only for displaying purpose.

lib/src/repo_path.rs
```diff
@@ -602,9 +602,8 @@ impl RepoPathUiConverter {
         match self {
             RepoPathUiConverter::Fs { cwd, base } => {
                 file_util::relative_path(cwd, &file.to_fs_path_unchecked(base))
-                    .to_str()
-                    .unwrap()
-                    .to_owned()
+                    .display()
+                    .to_string()
```
yuja, line 606: This one looks good.

lib/src/workspace.rs
```diff
@@ -358,15 +356,7 @@ impl Workspace {
...
-        let mut repo_file = File::create(&repo_file_path).context(&repo_file_path)?;
-        repo_file
-            .write_all(
-                repo_dir
-                    .to_str()
-                    .ok_or(WorkspaceInitError::NonUnicodePath)?
-                    .as_bytes(),
-            )
-            .context(&repo_file_path)?;
+        fs::write(&repo_file_path, repo_dir.display().to_string()).context(&repo_file_path)?;
```
yuja: Don't. Just error out.

We can also use `.as_bytes()` if we want to support non-UTF-8 paths.
https://doc.rust-lang.org/stable/std/os/unix/ffi/trait.OsStrExt.html#tymethod.as_bytes

cli/src/commands/root.rs
imp: As a matter of fact I've considered `.as_bytes()`, but then realized that this may not work on Windows at all. (I mean `use std::os::unix::ffi::OsStrExt`).

Windows counterpart (`std::os::windows::ffi::OsStrExt` doesn't have `fn as_bytes()` at all.

I don't think there is a clean and platform-independent way of printing `PathBuf` as raw bytes. What do you think?

cli/src/commands/sparse/edit.rs
imp: ok, I'll revert.

cli/src/progress.rs
imp: So basically revert, right? Ok.

cli/src/diff_util.rs
imp: Essentially, drop this commit entirely, right? Ok.

lib/src/git_backend.rs
imp: ok, I'll revert this.

lib/src/local_working_copy.rs
imp: alright

lib/src/workspace.rs
imp: Same as earlier comment - `OsStrExt::as_bytes` is unix-only. If potentially lossy conversion is not appropriate here, I'd just revert. `fs::write` optimization probably doesn't worth it by itself.

cli/src/commands/root.rs
yuja: Yes, that's why I said "on Unix." On Windows, paths are usually Unicode, so non-UTF-8 path handling wouldn't be needed. So, if we want to add extra support for Unix paths, we can use `.as_bytes()`. If not, the current implementation looks better than suppressing encoding errors.

emilazy: Unfortunately Windows paths can contain unpaired surrogates.

FWIW I feel trying to support non‐UTF‐8 paths is somewhat hopeless due to the “Makefile problem” and so on. Maybe we should consider using `camino` to enforce it at the boundary. I don’t know if our Git tree handling would need adjusting.

yuja: > Unfortunately Windows paths can contain unpaired surrogates.

Yeah, so _usually_.

> FWIW I feel trying to support non‐UTF‐8 paths is somewhat hopeless due to the “Makefile problem” and so on. Maybe we should consider using `camino` to enforce it at the boundary.

`jj` doesn't support non-UTF-8 paths in repository content, so the Makefile problem isn't relevant here. I don't have strong feeling whether we should support non-UTF-8 workspace directory paths.

imp: I see, reverting then.

imp:
I've updated the PR according to review comments. Not sure how much value it (this PR) has in its distilled version. Nonetheless it was a great codebase learning experience for me. Perhaps migrating to `camino` makes better sense than chasing non-UTF8 paths mishandling here and there all the time.

cli/src/commands/config/path.rs
```diff
@@ -40,13 +39,7 @@ pub fn cmd_config_path(
...
 ) -> Result<(), CommandError> {
     for config_path in args.level.config_paths(command.config_env())? {
-        writeln!(
-            ui.stdout(),
-            "{}",
-            config_path
-                .to_str()
-                .ok_or_else(|| user_error("The config path is not valid UTF-8"))?
-        )?;
+        writeln!(ui.stdout(), "{}", config_path.display())?;
```
yuja: revert (for the same reason as `workspace root`)

cli/src/commands/git/root.rs
```diff
@@ -35,10 +34,7 @@ pub fn cmd_git_root(
     let workspace_command = command.workspace_helper(ui)?;
     let store = workspace_command.repo().store();
     let git_backend = jj_lib::git::get_git_backend(store)?;
-    let root = git_backend
-        .git_repo_path()
-        .to_str()
-        .ok_or_else(|| user_error("The workspace root is not valid UTF-8"))?;
+    let root = git_backend.git_repo_path().display();
```
yuja: revert

cli/src/commands/workspace/root.rs
```diff
@@ -31,11 +30,7 @@ pub fn cmd_workspace_root(
     command: &CommandHelper,
     _args: &WorkspaceRootArgs,
 ) -> Result<(), CommandError> {
-    let root = command
-        .workspace_loader()?
-        .workspace_root()
-        .to_str()
-        .ok_or_else(|| user_error("The workspace root is not valid UTF-8"))?;
+    let root = command.workspace_loader()?.workspace_root().display();
```
yuja: revert

yuja (commented):
Can you remove empty commits?

cli/src/commands/config/path.rs
imp: Done

cli/src/commands/workspace/root.rs
imp: Done

cli/src/commands/git/root.rs
imp: Done

imp:
> Can you remove empty commits?

Removed

yuja (approved):
Thanks.

PR #6747, opened 2025-06-12
lib: Retry locking on ESTALE

algmyr:
This caused some issues with the Google internal NFS filesystem. Treating `ESTALE` (stale file handle) as a file deletion should be the right thing to do.

# Checklist

If applicable:

### DIFF
--- a/lib/src/lock/unix.rs
+++ b/lib/src/lock/unix.rs
@@ -45,17 +45,31 @@ impl FileLock {
                 }
             })?;
 
-            let stat = rustix::fs::fstat(&file).map_err(|errno| FileLockError {
-                message: "failed to stat lock file",
-                path: path.clone(),
-                err: errno.into(),
-            })?;
-            if stat.st_nlink == 0 {
-                // Lockfile was deleted, probably by the previous holder's `Drop` impl; create a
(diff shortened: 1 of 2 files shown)


algmyr:
FWIW, I don't think this code is tested currently, and I have no clue how one would test this.

yuja (approved):
Looks good to me, but can you take another look, @Ralith ?

thoughtpolice (approved):
Yes, seems fine to me. Frankly there's no way in hell we can test this reliably anytime soon, but maybe one day...

lib/src/lock/unix.rs
```diff
@@ -45,17 +45,29 @@ impl FileLock {
...
+                        // create a new one so our ownership is visible,
+                        // rather than hidden in an unlinked file. Not
+                        // always necessary, since the previous holder might
+                        // have exited abruptly.
+                        continue;
+                    }
+                }
+                Err(rustix::io::Errno::STALE) => {
+                    // The file handle is stale. This can happen on NFS filesystems
+                    // and should be treated as if the lockfile was deleted.
```
Ralith: Maybe note that this probably indicates the file was deleted and recreated remotely?

Ralith (approved):

PR #6748, opened 2025-06-12
website: Add the CHANGELOG

PhilipMetzger:
This was requested by @lilyball on Discord. I still think we should create a blog longterm so we can announce stuff and run surveys, but this small step won't block it.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -802,7 +802,7 @@ Thanks to the people who made this release happen!
 * The Jujutsu documentation site now publishes a schema for the official
   configuration file, which can be integrated into your editor for autocomplete,
   inline errors, and more.
-  Please [see the documentation](/docs/config.md#json-schema-support) for more
+  Please [see the documentation](docs/config.md#json-schema-support) for more
   on this.
 
 ### Fixed bugs
@@ -894,7 +894,7 @@ changes. Two select improvements:
 * `jj config list` now prints inline tables `{ key = value, .. }` literally.
   Inner items of inline tables are no longer merged across configuration files.
   See [the table syntax
-  documentation](docs/config.md#dotted-style-headings-and-inline-tables) for
+  documentation](docs/config.md#dotted-style-and-headings) for
   details.
 
 * `jj config edit --user` now opens a file even if `$JJ_CONFIG` points to a
@@ -2426,7 +2426,7 @@ Thanks to the people who made this release happen!
   If the deduced tracking flags are wrong, use `jj branch track`/`untrack`
   commands to fix them up.
 
-  See [automatic local branch creation](docs/config.md#automatic-local-branch-creation)
+  See [automatic local branch creation](docs/config.md#automatic-local-bookmark-creation)
   for details.
 
 * Non-tracking remote branches aren't listed by default. Use `jj branch list
@@ -2504,7 +2504,7 @@ Thanks to the people who made this release happen!
 * A default revset-alias function `trunk()` now exists. If you previously
   defined
   your own `trunk()` alias it will continue to overwrite the built-in one.
-  Check [revsets.toml](cli/src/config/revsets.toml)
+  Check [revsets.toml](docs/revsets.toml)
   and [revsets.md](docs/revsets.md)
   to understand how the function can be adapted.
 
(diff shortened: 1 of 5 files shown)


PhilipMetzger:
@ilyagr Do you have any suggestions how I can preserve the current CHANGELOG.md without needing to adjust all current links, since it fails the link-strict check of `mkdocs`?

ilyagr:
I'm not sure whether it will work or is a good idea, but you could try using <https://github.com/mondeja/mkdocs-include-markdown-plugin> (which we are already using for the CLI reference) with `rewrite_relative_urls: true`.

ilyagr:
I double-checked, and the `docs/...` links seem to magically work when copy-pasted into the release notes, e.g. <https://github.com/jj-vcs/jj/releases/tag/v0.29.0>. So, we probably don't want to change the links inside `CHANGELOG.md` itself. (Context: I briefly considered whether moving it to `docs/` would be reasonable, but it seems not).

PhilipMetzger:
> So, we probably don't want to change the links inside `CHANGELOG.md` itself. (Context: I briefly considered whether moving it to `docs/` would be reasonable, but it seems not).

Yes, I already had thrown the option of moving the file out of the window 😄  since it's important to preserve the top-level CHANGELOG as is. 

> I'm not sure whether it will work or is a good idea, but you could try using https://github.com/mondeja/mkdocs-include-markdown-plugin (which we are already using for the CLI reference) with `rewrite_relative_urls: true`.

Thanks for this suggestion, it'll hopefully work just like the CLI reference.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+<!--changelog-start-->
```
ilyagr: Do you really need these tags? I think the plugin would import the entire file by default.

PhilipMetzger: In a previous version the import would fail without these annotations. include-markdown "../CHANGELOG.md" just fails

PhilipMetzger: also I need to modify three links to make it pass the check, since the document contains stale links

PhilipMetzger:
I've adjusted all the links now, @ilyagr care to review?

docs/index.md
```diff
@@ -28,3 +28,4 @@ You may want to jump to:
 - [Tutorial and bird's eye view](tutorial.md)
 - [Working with GitHub](github.md)
 - [Development roadmap](roadmap.md)
+- [CHANGELOG](changelog.md)
```
ilyagr: Let's use normal capitalization ("Changelog") here as well.

mkdocs.yml
```diff
@@ -171,6 +171,7 @@ nav: # This lists all the files that become part of the documentation
       - Copy tracking and tracing: 'design/copy-tracking.md'
 
   - Development roadmap: 'roadmap.md'
+  - CHANGELOG: 'changelog.md'
```
ilyagr: Let's use normal capitalization: "Changelog".

CHANGELOG.md
ilyagr, line 1: Minor: `mkdocs serve` still complains about one thing:

```
INFO    -  Doc file 'changelog.md' contains an absolute link '/docs/config.md#json-schema-support', it was left as is.
```

It doesn't break the build, but you can see it at https://github.com/jj-vcs/jj/actions/runs/16355903335/job/46214093815?pr=6748#step:4:26.

Would you mind fixing that as well?

ilyagr (approved):
Seems to work well and look good, thank you very much!

One downside is that every released version will likely have a useless skeleton section for `[Unreleased]`, as in https://github.com/jj-vcs/jj/commit/312c610c592b41ec81f0661e124fd660cb510ab0. But this is not fatal, and we could fix it in another PR if we come up with an idea.

docs/index.md
ilyagr: (Optional) I'm not quite sure, but maybe "Change Log" is even better?

PhilipMetzger: done, used the first one

mkdocs.yml
PhilipMetzger: done

CHANGELOG.md
PhilipMetzger: done

PR #6750, opened 2025-06-13
Make `jj bookmark create` tab complete on existing bookmarks

hintron:
This makes it really convenient to create frequent "checkpoint" bookmarks that follow a pattern. E.g., myfeature-v1, myfeature-v2, etc.

# Checklist

If applicable:


yuja:
@senekor any thoughts on this? I think it's generally good for creation commands to suggest existing items.

senekor:
I don't like this. The first completion will fail or complete something wrong and the second completion may complete _too much_. Only once there are at least two "checkpoint" bookmarks for some feature can we expect completions that aren't actively harmful.

It also seems to over-optimize for an opinionated specific workflow. I don't know what these checkpoint bookmarks are supposed to accomplish.

I sometimes create named bookmarks, in which case I'd like them to be prefixed the same way as auto-generated ones, so this is my custom fish completion:
```fish
# complete `bookmark create` with push prefix
complete --command jj \
    --condition "__fish_seen_subcommand_from b bookmark; and __fish_seen_subcommand_from c create" \
    --arguments "(jj config get git.push-bookmark-prefix)"
```

I would be confused and annoyed if the completions tried to do more than that.

hintron:
Thanks for the feedback! Can I ask some follow up q's?

> The first completion will fail or complete something wrong and the second completion may complete _too much_. Only once there are at least two "checkpoint" bookmarks for some feature can we expect completions that aren't actively harmful.

Can you explain what you mean by "actively harmful"? Adding bookmark tab completion is pretty harmless, the way I see it. If you tab complete to an existing branch and try to create it, you just get a helpful error message saying that the branch already exists. And since there was no tab completion occurring beforehand, it's not changing anyone's workflow. The people that didn't tab complete before or don't want to tab complete can continue to do so with no change.

Git also tab completes bookmarks when creating branches. So it's not some wildly unexpected behavior.

> It also seems to over-optimize for an opinionated specific workflow. I don't know what these checkpoint bookmarks are supposed to accomplish.

I make checkpoint bookmarks as I work on a PR so I have a history of how my PR progressed. If I try something out that doesn't work, or if I remove something and later realize that I actually want it, or if I squashed things that should have remained separate, I can go back and restore my work from different checkpoints.

This was my workflow in Git, too, although you could argue that perhaps I should have been using tags. But in JJ, we can't create tags, so bookmarks it is. 

But regardless of what my workflow is, all this PR is doing is enabling potential workflows, not disrupting existing ones.

> I sometimes create named bookmarks, in which case I'd like them to be prefixed the same way as auto-generated ones, so this is my custom fish completion:
> 
> ```fish
> # complete `bookmark create` with push prefix
> complete --command jj \
>     --condition "__fish_seen_subcommand_from b bookmark; and __fish_seen_subcommand_from c create" \
>     --arguments "(jj config get git.push-bookmark-prefix)"
> ```
> 
> I would be confused and annoyed if the completions tried to do more than that.

If you type 1-3 letters more with this PR, you get similar behavior without needing a custom completion :) I think that could help others who are not interested in writing shell completion scripts.

However, if you insist on that functionality, I'm working on a proof of concept for this case as well - My idea is if you set `push-bookmark-prefix-complete = "true"`, bookmark tab completions will discard all bookmark suggestions except the bookmarks that start with the push prefix.

Thanks,
-Michael

senekor:
> Thanks for the feedback! Can I ask some follow up q's?
> 
> > The first completion will fail or complete something wrong and the second completion may complete _too much_. Only once there are at least two "checkpoint" bookmarks for some feature can we expect completions that aren't actively harmful.
> 
> Can you explain what you mean by "actively harmful"? Adding bookmark tab completion is pretty harmless, the way I see it. If you tab complete to an existing branch and try to create it, you just get a helpful error message saying that the branch already exists. And since there was no tab completion occurring beforehand, it's not changing anyone's workflow. The people that didn't tab complete before or don't want to tab complete can continue to do so with no change.

But there was tab completion beforehand. My custom one. Currently, if I type `jj b c s<tab>`, my shell will complete that to `jj b c senekor-`, which is exactly what I want. With this PR, there are two possibilities of that workflow being made worse:

1. Assume there already exists another bookmark called "super-cool-feature". The completion would prompt me for two options, "senekor-" or "super-cool-feature".

2. Assume there is exactly one other bookmark starting with "senekor-", e.g. "senekor-feature-1". Now I want to start working on "senekor-other-feature". But the full existing bookmark name is completed, forcing me to hit backspace a bunch of times.

This is what I mean by "actively harmful".

> Git also tab completes bookmarks when creating branches. So it's not some wildly unexpected behavior.

That depends on where you're getting your completions from. Upstream git doesn't have any fish completions and the [zsh completions](https://github.com/git/git/blob/master/contrib/completion/git-completion.zsh) are very barebones. Your git completions may depend on your distribution package, your shell, maybe you use oh-my-zshell etc.

Don't get me wrong, I see the utility and why some people would want it. But I don't think it should be included upstream, by default.

> > It also seems to over-optimize for an opinionated specific workflow. I don't know what these checkpoint bookmarks are supposed to accomplish.
> 
> I make checkpoint bookmarks as I work on a PR so I have a history of how my PR progressed. If I try something out that doesn't work, or if I remove something and later realize that I actually want it, or if I squashed things that should have remained separate, I can go back and restore my work from different checkpoints.
> 
> This was my workflow in Git, too, although you could argue that perhaps I should have been using tags. But in JJ, we can't create tags, so bookmarks it is.
>
> But regardless of what my workflow is, all this PR is doing is enabling potential workflows, not disrupting existing ones.

Yeah let's discuss that a bit, because your workflow seems very strange to me. It's definitely relevant, because this PR could be an [XY problem](https://en.wikipedia.org/wiki/XY_problem).

On your PRs, do you put commits on top of each other without editing old ones? If so, why do you need the bookmarks? The commits of the old state are always there. Or do you amend existing commits (which is more common with jujutsu)? If you amend the commits, how do the bookmarks help keep track of previous states? Because jujutsu will rebase bookmarks automatically, making the "checkpoint" bookmarks point at the new commits, instead of the old ones.

If I understand you correctly, you would have to continuously create copies of your commits with `jj duplicate` to be able to amend existing commits while your checkpoint bookmarks still point at the old state.

Do you know about the evolog? jj keeps track of previous states of commits for you in the background, there is no need to do that manually with bookmarks.

> If you type 1-3 letters more with this PR, you get similar behavior without needing a custom completion :) I think that could help others who are not interested in writing shell completion scripts.
> 
> However, if you insist on that functionality, I'm working on a proof of concept for this case as well - My idea is if you set `push-bookmark-prefix-complete = "true"`, bookmark tab completions will discard all bookmark suggestions except the bookmarks that start with the push prefix.

I don't think it's a good idea to introduce lots of complicated configuration options to allow users to tweak completions exactly how they want. Such configuration needs to be looked up in documentation. At that point, we could just put the custom completion snippets in the documentation / wiki that people can copy. That's essentially the same user experience without any additional complexity in jj itself.

I think we should strive for upstream completions with a very low false-positive rate, which this PR doesn't achieve. More opinionated completions should be left up to user customization via the shell. Scripts can be shared on the wiki so others benefit as well.

hintron:
> On your PRs, do you put commits on top of each other without editing old ones? If so, why do you need the bookmarks? The commits of the old state are always there. Or do you amend existing commits (which is more common with jujutsu)? If you amend the commits, how do the bookmarks help keep track of previous states? Because jujutsu will rebase bookmarks automatically, making the "checkpoint" bookmarks point at the new commits, instead of the old ones.
> 
> If I understand you correctly, you would have to continuously create copies of your commits with `jj duplicate` to be able to amend existing commits while your checkpoint bookmarks still point at the old state.
> 
> Do you know about the evolog? jj keeps track of previous states of commits for you in the background, there is no need to do that manually with bookmarks.

I guess there are two issues with relying solely on evolog:

1) I want my history to be preserved on the remote and not "garbage collected". I want to be able to clone from my fork on different computers and not find out that the evolog is different. Perhaps that is just a Git problem, and JJ properly saves those hidden commits in my upstream repo. But even if that were the case,

2) I want to be able to see those branches on my remote, and I want to easily see my "blessed" checkpoints instead of looking through the evolog to try to figure out what state to return to. Instead of sifting through 100s of evolog entries, I can sift through a handful of checkpoint bookmarks. And I can see that history on Github via the branches tab. Can't use evolog on Github or with non-JJ git tools.

Happy to hear of any ideas to make this type of workflow more efficient.

hintron:
And yeah, the commits in my checkpoint bookmark are immutable, but then I start a new branch, rebase on the latest main, and squash any fixup commits, so now my new branch is cleaned up. Then, I add fixup and/or feature commits to that, and the process repeats.

Git explicitly teaches that "branches are cheap" so I am abusing them :)

hintron:
I guess if we're talking about XY problems, then what I really want is if jj could have some kind of "named snapshot" mode (`jj snapshot <name> --remote <remote>`) that automatically creates a remote-only bookmark of the name `<name>` (creates tracking bookmark and pushes to remote, then forgets local bookmark). `jj git push --named` is the closest thing to this, but you need a `jj forget` afterwards and there is no tab completion out of the box.

senekor:
Yeah, unfortunately jj doesn't sync your evolog (yet).

> the commits in my checkpoint bookmark are immutable, but then I start a new branch, rebase on the latest main, and squash any fixup commits

Am I correct in assuming you do this with the git CLI and not jj? I feel like if you used jj, the story would sound different. If you use jj to rebase, the checkpoint-bookmark would be dragged along. You'd need `jj duplicate` to avoid that.

I recommend using jj exclusively to create / edit commits. Is there a specific reason you do it with git?

Out of curiosity, do you delete these checkpoint-branches after a PR is merged? Or do you preserve them essentially forever?

If history-syncing is that important to you, then I'd say your approach is basically the right one, given the limitations.

> I really want is if jj could have some kind of "named snapshot" mode (`jj snapshot <name> --remote <remote>`) that automatically creates a remote-only bookmark of the name `<name>` (creates tracking bookmark and pushes to remote, then forgets local bookmark). `jj git push --named` is the closest thing to this, but you need a `jj forget` afterwards and there is no tab completion out of the box.

I agree, you really need a script that automates your entire workflow, not just tab completion for new bookmarks. This can be done. Very rough pseudo-code I'm thinking of:

```bash
current_bookmark=$( jj log -r "heads(::@- & bookmarks())" )
jj duplicate "trunk()..$current_bookmark"
tip=TODO # parse tip of checkpoint branch from `jj duplicate` output (stderr)
nr=TODO # list remote bookmarks, detect how many checkpoint branches there are already, pick the next number

ckp_bookmark="${current_bookmark}-ckp-${nr}"
jj git push --named $ckp_bookmark=$tip
jj bookmark forget $ckp_bookmark
```

You can create an alias that let's you invoke this script with `jj snapshot` or whatever you like.

hintron:
> Am I correct in assuming you do this with the git CLI and not jj? I feel like if you used jj, the story would sound different. If you use jj to rebase, the checkpoint-bookmark would be dragged along. You'd need `jj duplicate` to avoid that.
> 
> I recommend using jj exclusively to create / edit commits. Is there a specific reason you do it with git?

I don't use git. I have been doing this exclusively in JJ over the last two weeks as I evaluate it and try to see if it helps my workflow. 

You are right that rebase will move my checkpoint branches, so I make sure to `jj bookmark forget` them beforehand so they stay put on the remote.

> Out of curiosity, do you delete these checkpoint-branches after a PR is merged? Or do you preserve them essentially forever?

No, I don't delete them, so they end up forming a history of what I work on, which I like.

> If history-syncing is that important to you, then I'd say your approach is basically the right one, given the limitations.
> 
> > I really want is if jj could have some kind of "named snapshot" mode (`jj snapshot <name> --remote <remote>`) that automatically creates a remote-only bookmark of the name `<name>` (creates tracking bookmark and pushes to remote, then forgets local bookmark). `jj git push --named` is the closest thing to this, but you need a `jj forget` afterwards and there is no tab completion out of the box.
> 
> I agree, you really need a script that automates your entire workflow, not just tab completion for new bookmarks. This can be done. Very rough pseudo-code I'm thinking of:
> 
> ```shell
> current_bookmark=$( jj log -r "heads(::@- & bookmarks())" )
> jj duplicate "trunk()..$current_bookmark"
> tip=TODO # parse tip of checkpoint branch from `jj duplicate` output (stderr)
> nr=TODO # list remote bookmarks, detect how many checkpoint branches there are already, pick the next number
> 
> ckp_bookmark="${current_bookmark}-ckp-${nr}"
> jj git push --named $ckp_bookmark=$tip
> jj bookmark forget $ckp_bookmark
> ```
> 
> You can create an alias that let's you invoke this script with `jj snapshot` or whatever you like.

Ok, thanks for that! I'll have to give that a try.

Is there any appetite in JJ for supporting this use case natively? It not, I guess we can close this PR.

senekor:
> You are right that rebase will move my checkpoint branches, so I make sure to `jj bookmark forget` them beforehand so they stay put on the remote.

I see, didn't think of that. Don't you run into problems with divergent change-ids then? The old versions of commits will have the same change-id as the new ones, right? `jj duplicate` would avoid that by giving the new commits new change-ids.

> No, I don't delete them, so they end up forming a history of what I work on, which I like.

I'm not gonna tell you what to do in your own repos, but that seems very ugly to me. You'll accumulate a large number of bookmarks over time and the only way to get rid of them is to delete that history. You also have to watch out to actually record the name of the bookmark into the commit tree when you merge a PR, e.g. with an explicit "Merge branch XYZ into main" commit, otherwise it'll be hard later to figure out to what part of the final commit tree these old checkpoint branches belong. I wouldn't tolerate that in repos where I have some say. Allow me to suggest a different approach. If you want the history to be preserved indefinitely, you should just record it in the commit graph, that's what it's for. How about something like this:

```
* merge
|\
| * v2
| |\
| | ------* v1
| |       |
| * C'    * A
| |       |
| * B'    * B
| |       |
| * A'    * C
|/       /
*--------
```

Every version is marked with a commit, which is a merge commit of (1) the commits that belong to that version and (2) the version commit of the previous version. That way all versions are clearly separate but preserved in the tree without any bookmarks.

hintron:
> > You are right that rebase will move my checkpoint branches, so I make sure to `jj bookmark forget` them beforehand so they stay put on the remote.
> 
> I see, didn't think of that. Don't you run into problems with divergent change-ids then? The old versions of commits will have the same change-id as the new ones, right? `jj duplicate` would avoid that by giving the new commits new change-ids.

Yeah, I do end up with divergent commits. I didn't realize that I could do `jj duplicate <base>..@ -d <base>` to create a copy of a branch! That is awesome. I will use that from now before creating my checkpoint branches to avoid divergent change ids. 

> I'm not gonna tell you what to do in your own repos, but that seems very ugly to me. You'll accumulate a large number of bookmarks over time and the only way to get rid of them is to delete that history. You also have to watch out to actually record the name of the bookmark into the commit tree when you merge a PR, e.g. with an explicit "Merge branch XYZ into main" commit, otherwise it'll be hard later to figure out to what part of the final commit tree these old checkpoint branches belong. I wouldn't tolerate that in repos where I have some say.

This is only for my personal forks. If I accumulate too many checkpoint bookmarks, I can always go back and delete them, but I prefer the more explicit history, in case I want to look at any of my rejected implementations. 

I do think that perhaps I could change my checkpoint branch naming strategy to have some common "chk-" prefix, so I could potentially delete all of them in bulk if needed.

> Allow me to suggest a different approach. If you want the history to be preserved indefinitely, you should just record it in the commit graph, that's what it's for. How about something like this:
> ...
> Every version is marked with a commit, which is a merge commit of (1) the commits that belong to that version and (2) the version commit of the previous version. That way all versions are clearly separate but preserved in the tree without any bookmarks.

This won't work for me because most of my repos require a clean linear history, so I need to heavily squash and rebase my own branches to be linear with no merge commits.

I'll go ahead and close out this PR and try refining this workflow a bit. Thanks!

PR #6751, opened 2025-06-13
revset: fix `is_filter_tree()` for nested intersections

scott2000:
@yuja I noticed this while resolving the conflicts after #6719 was merged.

Currently, `is_filter_tree()` only works for intersections with 2 expressions. If there are 3 expressions, it always returns false.

# Checklist

If applicable:

### DIFF
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -1528,7 +1528,7 @@ fn internalize_filter<St: ExpressionState>(
         if let RevsetExpression::Intersection(expression1, _) = expression {
             // 'f1 & f2' can't be evaluated by itself, but 's1 & f2' can be
             // filtered within 's1'. 'f1 & s2' should have been reordered.
-            is_filter(expression1)
+            is_filter_tree(expression1)
         } else {
             is_filter(expression)
         }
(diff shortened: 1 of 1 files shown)


yuja:
Oh, maybe missed `AsFilter` insertion somewhere. It exists for avoiding recursion at `is_*()` function.

scott2000:
> Oh, maybe missed `AsFilter` insertion somewhere. It exists for avoiding recursion at `is_*()` function.

Yeah, I don't think `f1 & f2 -> AsFilter(f1 & f2)` is implemented anywhere currently. I'm not sure that the recursion would necessarily be an issue in this case, because `is_filter_tree()` isn't called at each intersection node; it's only called at the non-intersection node that's a parent of all of the nested intersections.

yuja:
> I don't think `f1 & f2 -> AsFilter(f1 & f2)` is implemented anywhere currently.

and `is_filter_tree()` would be just `is_filter()` if we had one. Please feel free to ignore the problem for now if you don't have immediate problem with your PR. Maybe we can fix that later after your PR.

> I'm not sure that the recursion would necessarily be an issue in this case, because `is_filter_tree()` isn't called at each intersection node; it's only called at the non-intersection node that's a parent of all of the nested intersections.

But intersection is left-recursive, so we'll have to check whether the deepest leaf is set. That seems as bad as inspecting union descendants whether one of them is a filter (without relying on `AsFilter` cache node.)

scott2000:
> and `is_filter_tree()` would be just `is_filter()` if we had one.

Yeah, that would be nice.

> Please feel free to ignore the problem for now if you don't have immediate problem with your PR. Maybe we can fix that later after your PR.

Sounds good. I don't think this would be an issue in practice anyway.

> But intersection is left-recursive, so we'll have to check whether the deepest leaf is set. That seems as bad as inspecting union descendants whether one of them is a filter (without relying on `AsFilter` cache node.)

I think it's actually significantly worse for unions, because the check has to happen several times since `is_filter_tree()` is evaluated for each union expression recursively (e.g. for `((f1 | f2) | f3) | f4`, it would have to check if `f1` is a filter 3 times: once while transforming `f1 | f2`, once while transforming `(f1 | f2) | f3`, and then again while transforming `((f1 | f2) | f3) | f4`).

But transforming an intersection doesn't require checking `is_filter_tree()`, so for `((f1 & f2) & f3) & f4 | f5`, it would only have to check `f1` once: while transforming the entire union expression `((f1 & f2) & f3) & f4 | f5`. Like it doesn't get called multiple times redundantly anyways, because `is_filter_tree()` only iterates over intersection nodes, and transforming an intersection node doesn't require checking `is_filter_tree()`. But I could be wrong.

yuja:
> I think it's actually significantly worse for unions, because the check has to happen several times

I think this could be said the other way around. If `is_filter()` were a recursive process, we wouldn't have to insert `AsFilter` at union. "filter"-ness only matters at intersection (and at the root.)

scott2000:
> > I think it's actually significantly worse for unions, because the check has to happen several times
> 
> I think this could be said the other way around. If `is_filter()` were a recursive process, we wouldn't have to insert `AsFilter` at union. "filter"-ness only matters at intersection (and at the root.)

Oh yeah that's true, I hadn't thought of that. Thanks!

PR #6761, opened 2025-06-15
Prepend `backout` description with '(deprecated; use `revert`) '

giladwo:
Hi!

This builds on #6290 - I probably should have just tried `jj backout` before reading through many `--help` pages trying to understand when I'd rather use `backout` over `revert` and vice versa, but I think it'd be an improvement to simply put this at the top and in the tab completion help.

WDYT?

# Checklist

If applicable:


thoughtpolice (approved):
Seems reasonable. Thanks!

PR #6763, opened 2025-06-15
formatter: add support for ANSI 256-color palette

### LINKED ISSUE #6691: FR: Support color codes 16-255
jj supports setting UI colors to terminal color codes 0-15 via names such as "green" and "bright green."

However, most terminals support modifying color codes 0-255. This is useful to to create terminal themes that have more than just 16 colors. I personally use color codes 16-20 in my themes for UI colors that aren't safe to map the standard 16 range to, such as "diff added background" and "muted text foreground."

Supporting all color codes 0-255 is pretty standard for TUIs, and I would love it if jj did so too. I propose something very simple like

```toml
[colors]
"rest" = 17
# or
"rest" = "17"
```

This would also help support the rare terminal environment that only has 256 color capability.

mernen:
Fixes #6691, using @ilyagr's proposed syntax. My first instinct would have been a shorter `ansi<N>`, but I see that would be ambiguous with other escape codes.

Since the 256-color palette isn't well-known and hard to reason about without a lookup table, I included a link to [Wikipedia's article](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit). It goes around a bit with technical details, but the compact layout of the table itself is very nice for lookup. The best contender would be [Ditig's 256 colors cheat sheet](https://www.ditig.com/256-colors-cheat-sheet), which is the top result in most search engines, but AFAIK not a well-known website otherwise, and I find its vertical layout not very practical.

I'm also including fixes for typos I spotted in the changelog.

# Checklist

If applicable:

### DIFF
--- a/cli/src/formatter.rs
+++ b/cli/src/formatter.rs
@@ -465,8 +465,8 @@ fn deserialize_color<'de, D>(deserializer: D) -> Result<Color, D::Error>
 where
     D: serde::Deserializer<'de>,
 {
-    let name_or_hex = String::deserialize(deserializer)?;
-    color_for_name_or_hex(&name_or_hex).map_err(D::Error::custom)
+    let color_str = String::deserialize(deserializer)?;
+    color_for_string(&color_str).map_err(D::Error::custom)
 }
 
 fn deserialize_color_opt<'de, D>(deserializer: D) -> Result<Option<Color>, D::Error>
@@ -476,8 +476,8 @@ where
     deserialize_color(deserializer).map(Some)
 }
 
-fn color_for_name_or_hex(name_or_hex: &str) -> Result<Color, String> {
-    match name_or_hex {
+fn color_for_string(color_str: &str) -> Result<Color, String> {
+    match color_str {
         "default" => Ok(Color::Reset),
         "black" => Ok(Color::Black),
         "red" => Ok(Color::DarkRed),
@@ -495,8 +495,18 @@ fn color_for_name_or_hex(name_or_hex: &str) -> Result<Color, String> {
         "bright magenta" => Ok(Color::Magenta),
         "bright cyan" => Ok(Color::Cyan),
         "bright white" => Ok(Color::White),
-        _ => color_for_hex(name_or_hex).ok_or_else(|| format!("Invalid color: {name_or_hex}")),
+        _ => color_for_ansi256_index(color_str)
+            .or_else(|| color_for_hex(color_str))
+            .ok_or_else(|| format!("Invalid color: {color_str}")),
     }
+}
+
+fn color_for_ansi256_index(color: &str) -> Option<Color> {
+    color
+        .strip_prefix("ansi-color-")
+        .filter(|s| *s == "0" || !s.starts_with('0'))
+        .and_then(|n| n.parse::<u8>().ok())
+        .map(Color::AnsiValue)
 }
 
 fn color_for_hex(color: &str) -> Option<Color> {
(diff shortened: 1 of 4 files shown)


cli/src/formatter.rs
```diff
@@ -465,8 +465,8 @@ fn deserialize_color<'de, D>(deserializer: D) -> Result<Color, D::Error>
 where
     D: serde::Deserializer<'de>,
 {
-    let name_or_hex = String::deserialize(deserializer)?;
-    color_for_name_or_hex(&name_or_hex).map_err(D::Error::custom)
+    let name_or_index_or_hex = String::deserialize(deserializer)?;
```
mernen, line 468: This is getting a bit ridiculous — perhaps it'd be better to drop the fully descriptive name now that there are 3 options?

cli/src/formatter.rs
```diff
@@ -495,10 +495,20 @@ fn color_for_name_or_hex(name_or_hex: &str) -> Result<Color, String> {
...
+        _ => color_for_ansi256_index(name_or_index_or_hex)
+            .or_else(|| color_for_hex(name_or_index_or_hex))
+            .ok_or_else(|| format!("Invalid color: {name_or_index_or_hex}")),
     }
 }
 
+fn color_for_ansi256_index(color: &str) -> Option<Color> {
+    color
+        .strip_prefix("ansicolor")
+        .filter(|s| s.len() <= 3)
```
ilyagr: Is this `filter` needed? I guess it does no harm, but still.

docs/config.md
```diff
@@ -128,6 +128,16 @@ You can also use a 6-digit hex code for more control over the exact color used:
 change_id = "#ff1525"
 ```
 
+Not every terminal supports "true color" palettes, though. If that doesn't work,
```
ilyagr: I would remove the beginning of this, and say instead

> `jj` also supports colors from the \[ANSI 256-color palette\]...

I'd also mention that 

> `ansicolor0` through `ansicolor7` are the same as the named colors, while `ansicolor8` through `ansicolor15` are the same as their bright versions. For example, `ansicolor8` is the same as `bright black`.

**Update:** Though, on second thought, I'm not 100% sure this is true on all terminals, perhaps some can be configured to show them differently. If not, we could say "Most terminals treat `ansicolor0` through `ansicolor7` the same...". But they *are* the same in every terminal I tried on my Mac (Ghostty, Kitty, Apple Terminal.app), regardless of the terminal theme.

**Update 2:** Maybe we can just say "... are usually the same ...".

-----------

I'd also link to <https://www.ditig.com/256-colors-cheat-sheet>, for example you could call link "another reference". Since the Wikipedia link is a bit long, footnote-style Markdown links might be a bit better.

ilyagr (commented):
This looks great to me!

I'll wait a bit before approving, in case others have comments.

`ansicolor55` seems perfectly workable for me, though it might not be beautiful. I think I do prefer it to `ansi55`.

cli/src/formatter.rs
mernen: It's just to keep things minimally normalized, or else `ansicolor0000000000000000123` would be accepted. Supporting zero-padding to keep all entries three-digit long seems reasonable to me, though, so I'm intentionally letting e.g. `001` through.

docs/config.md
ilyagr: Or maybe https://askubuntu.com/a/821163/277701 is more useful. However, that's Linux-only.

cli/src/formatter.rs
ilyagr: I see. Perhaps add a test for these two cases? Say `000` and `0000`?

It's a clever test case.

docs/config.md
ilyagr: I have also not found a perfect reference immediately :(. I don't feel strongly about including a second reference or not. I'm a bit sad that the script in the Wikipedia article doesn't do what I thought it does.

https://gist.githubusercontent.com/HaleTom/89ffe32783f89f403bba96bd7bcd1263/raw/e50a28ec54188d2413518788de6c6367ffcea4f7/print256colours.sh seems like a nice script (from Stackoverflow link, requires bash).

cli/src/formatter.rs
```diff
@@ -835,6 +845,33 @@ mod tests {
         ");
     }
 
+    #[test]
+    fn test_color_formatter_ansi256() {
+        // Test the color code for each color.
+        let config = config_from_string(indoc! {"
```
ilyagr: Do ANSI colors work as background colors? It'd be nice to have a test for that. If it doesn't work, perhaps we need some logic to prohibit it.

E.g. try a color label like

```toml
bgpurple = { fg = 'white', bg = 'ansicolor93' }
```

**Update:** They should work fine. Still, I think let's make something like this one of the test cases.

mernen: Indeed, good idea.

cli/src/formatter.rs
ilyagr: Or maybe `009`, to make sure it's not octal. Doesn't matter too much.

yuja: If `ansicolor000` is accepted, I would expect `ansicolor0000` also works.

yuja:
> `ansicolor55` seems perfectly workable for me, though it might not be beautiful. I think I do prefer it to `ansi55`.

Our convention in config is kebab-case, so `ansi-color-55` seems better. If we want to omit dashes, I would use `ansi55`.

ilyagr:
Another thought that came to mind is `ansi8bit-55` or `8bit-55`. I think I prefer that to `ansi-55`.

`ansicolor-55` sounds better to me than `ansi-color-55`, but I guess you'd have mentioned it if it were acceptable to you. Ultimately any of the options we discussed are OK with me (except for plain numbers).

yuja:
> Another thought that came to mind is `ansi8bit-55` or `8bit-55`. I think I prefer that to `ansi-55`.
> 
> `ansicolor-55` sounds better to me than `ansi-color-55`, but I guess you'd have mentioned it if it were acceptable to you. Ultimately any of the options we discussed are OK with me (except for plain numbers).

We use dashed words in config (e.g. `private-commits`), so `ansicolor` should have to be `ansi-color`, I think. I'm not sure if we should insert separator between name and integer, but `ansi-color55` looks a bit unusual to me. OTOH, `ansi55` looks okay.

mernen:
I changed the prefix to `ansi-color-N`, and now *N* must not have leading zeros. `ansi-colorN` seems odd to me as well, and `ansi-N` seems ambiguous/misleading (could be interpreted as Select Graphic Rendition *N*, i.e. to enable arbitrary effects that we don't currently support like blinking). `8bit-N` is another good candidate IMHO.

docs/config.md
mernen: The documentation text has also been tweaked since this conversation was started, but it still only includes a reference to Wikipedia. I'm not sure Ditig brings much to be worth an additional link, and I'm wary of encouraging people to run arbitrary scripts from sources outside our control to test the colors.

cli/src/formatter.rs
```diff
@@ -476,8 +476,8 @@ where
     deserialize_color(deserializer).map(Some)
 }
 
-fn color_for_name_or_hex(name_or_hex: &str) -> Result<Color, String> {
-    match name_or_hex {
+fn color_for_string(name_or_index_or_hex: &str) -> Result<Color, String> {
```
yuja: nit: I would call it `name`, `color`, or `color_str`.

net:
Wonderful, many thanks for this!

ilyagr:
Thank you!

PR #6764, opened 2025-06-15
cli: add `json()` template functions

thoughtpolice:
These template functions just emit the given commit objects as JSON, allowing you to do something like this:

    jj --no-graph --ignore-working-copy -T 'json()' | jq

The `json()` function has two specific things I optimized for usability:

- It always outputs a newline after an object
- `json() = json(self)`

The reasons for these are pretty simple; it is overwhelmingly the case that JSON is taken and redirected elsewhere, and requiring users to always remember a `++ "\n"` at the end is fairly unfriendly. In contrast, just including it changes nothing about parsing, and even helps the users terminal when they only write out a single commit. And the other thing is that the 99% of uses will be for `jj log`, so it makes sense to optimize for the common case when the commit object is `self`.

In practice this means the output is actually 'ndjson' but requiring the user to remember that or care is also pointless, IMO.

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -433,3 +433,33 @@ git_format_patch_email_headers = '''
 # Gerrit change id is 40 chars, jj change id is 32, so we need some padding.
 # `6a6a6964` is the hexadecimal of `jjid` in ASCII, just to not pad with zeros.
 "format_gerrit_change_id_trailer(commit)" = '"Change-Id: I6a6a6964" ++ commit.change_id().normal_hex() ++ "\n"'
+
+# JSON formatting function(s)
+'json()' = 'json(self)'
+'json(commit)' = '''concat("{",
+  separate(",",
+    '"commit_id":"' ++ stringify(commit.commit_id()) ++ '"',
+    '"change_id":"' ++ stringify(commit.change_id()) ++ '"',
+    '"parents":' ++ "[" ++ commit.parents().map(|p| '"' ++ stringify(p.commit_id()) ++ '"').join(",") ++ "]",
+    '"author":{' ++ separate(",",
+      '"name":"' ++ coalesce(commit.author().name(), "") ++ '"',
+      '"email":"' ++ coalesce(commit.author().email(), "") ++ '"',
+      '"timestamp":"' ++ commit.author().timestamp().format("%Y-%m-%dT%H:%M:%S%:z") ++ '"'
+    ) ++ "}",
+    '"committer":{' ++ separate(",",
+      '"name":"' ++ coalesce(commit.committer().name(), "") ++ '"',
+      '"email":"' ++ coalesce(commit.committer().email(), "") ++ '"',
+      '"timestamp":"' ++ commit.committer().timestamp().format("%Y-%m-%dT%H:%M:%S%:z") ++ '"'
+    ) ++ "}",
+    '"description":' ++ commit.description().escape_json(),
+    '"empty":' ++ if(commit.empty(), "true", "false"),
+    '"conflict":' ++ if(commit.conflict(), "true", "false"),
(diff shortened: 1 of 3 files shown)


cli/src/config/templates.toml
```diff
@@ -433,3 +433,33 @@ git_format_patch_email_headers = '''
 # Gerrit change id is 40 chars, jj change id is 32, so we need some padding.
 # `6a6a6964` is the hexadecimal of `jjid` in ASCII, just to not pad with zeros.
 "format_gerrit_change_id_trailer(commit)" = '"Change-Id: I6a6a6964" ++ commit.change_id().normal_hex() ++ "\n"'
+
+# JSON formatting function(s)
+'json()' = 'json(self)'
+'json(commit)' = '''concat("{",
+  separate(",",
+    '"commit_id":"' ++ stringify(commit.commit_id()) ++ '"',
```
thoughtpolice, line 441: Most of these can be cleaned up (Claude seems to like coalesce when it generated this in my personal config), my initial draft here is just to get something passing with the fields in place.

cli/tests/test_templater.rs
```diff
@@ -530,6 +530,33 @@ fn test_templater_config_function() {
...
 
+#[test]
+fn test_templater_json_function() {
+    let test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let work_dir = test_env.work_dir("repo");
+    let render = |template| get_template_output(&work_dir, "@", template);
+
+    insta::assert_snapshot!(get_template_output(&work_dir, "@", r#"json()"#), @r#"
+    {"commit_id":"e8849ae12c709f2321908879bc724fdb2ab8a781","change_id":"qpvuntsmwlqtpsluzzsnyyzlmlwvmlnu","parents":["0000000000000000000000000000000000000000"],"author":{"name":"Test User","email":"test.user@example.com","timestamp":"2001-02-03T04:05:07+07:00"},"committer":{"name":"Test User","email":"test.user@example.com","timestamp":"2001-02-03T04:05:07+07:00"},"description":"","empty":true,"conflict":false,"divergent":false,"hidden":false,"current_working_copy":true,"immutable":false,"bookmarks":[],"tags":[],"remote_bookmarks":[]}
```
thoughtpolice, line 540: TBH, I don't really know of a good set of tests here beyond "the json looks like this", and some escaping testing. I do not want to rely on jq, and adding a bunch of stuff for serde or whatever seems like overkill. 🤷🏻‍♂️

yuja:
I don't think it's good idea to implement serialization functions (for various types) in template. We can't do `json(tags)` for example, and it would get messy if we start adding `json_<type>()` template for each type.

Suppose we don't have to guarantee backward compatibility of serialized fields, I think we can add a real `json(serializable)` template function that processes objects backed by serde?

https://crates.io/crates/erased-serde

thoughtpolice:
> I don't think it's good idea to implement serialization functions (for various types) in template. We can't do `json(tags)` for example, and it would get messy if we start adding `json_<type>()` template for each type.
> 
> Suppose we don't have to guarantee backward compatibility of serialized fields, I think we can add a real `json(serializable)` template function that processes objects backed by serde?
> 
> https://crates.io/crates/erased-serde

Yeah, I thought about doing it this way at first but mostly wanted to get the idea out there for starters. FWIW, I don't think we really need to care about backwards compatibility any moreso than we do about object method compatibility, which I see as "Mostly stable but still free to change". Personally I think a lot of the most obvious details will probably remain forever (commit id, change id, parents, etc)

yuja:
https://github.com/jj-vcs/jj/pull/6777

thoughtpolice:
You're a hero dude, thanks!

PR #6782, opened 2025-06-18
revset: expand all() set to include all referenced commits

### LINKED ISSUE #5871: Incorrect revsets involving hidden commits and negation
## Description

Hidden revisions are not always included in a revset when they are explicitly mentioned by commit id.

Here are some commands that demonstrate the issue
```
# The hidden commit is logged as normal
jj log --no-graph --template builtin_log_oneline --revisions 'e59f122f'
zksmnskl hidden dta 2 hours ago e59f122f (empty) <redacted description>
# The hidden commit is not logged when unioned a revset containing no revisions
jj log --no-graph --template builtin_log_oneline --revisions '(none() ~ empty()) | e59f122f'
# The hidden commit is still not logged when the negation does not match it
jj log --no-graph --template builtin_log_oneline --revisions '(none() ~ description(exact:"does not match actual description")) | e59f122f'
# Not all revsets cause this bug in conjunction with negation
jj log --no-graph --template builtin_log_oneline --revisions '(none() ~ description(exact:"does not match actual description")) | e59f122f'
jj log --no-graph --template builtin_log_oneline --revisions '(none() ~ all()) | e59f122f'
zksmnskl hidden dta 2 hours ago e59f122f (empty) <redacted description>
```

## Expected Behavior

Hidden revisions are always included in a revset when they are unioned with another revset

## Actual Behavior

This is not always the case

## Specifications

- Platform: Mac OSX
- Version: 0.26.0

yuja:
Fixes #5871

# Checklist

If applicable:

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -12,9 +12,16 @@ commits) to such commands.
 
 The words "revisions" and "commits" are used interchangeably in this document.
 
+## Hidden revisions
+
 Most revsets search only the [visible commits](glossary.md#visible-commits).
 Other commits are only included if you explicitly mention them (e.g. by commit
-ID or a Git ref pointing to them).
+ID, `<name>@<remote>` symbol, or `at_operation()` function).
+
+If hidden commits are specified, their ancestors also become available to the
+search space. They are included in `all()`, `x..`, `~x`, etc., but not in
+`..visible_heads()`, etc. For example, `hidden_id | all()` is equivalent to
+`hidden_id | ::(hidden_id | visible_heads())`.
 
 ## Symbols
 
@@ -52,10 +59,10 @@ only symbols.
 
 * `x-`: Parents of `x`, can be empty.
 * `x+`: Children of `x`, can be empty.
-* `x::`: Descendants of `x`, including the commits in `x` itself. Shorthand for
-  `x::visible_heads()`.
-* `x..`: Revisions that are not ancestors of `x`. Shorthand for
-  `x..visible_heads()`. Equivalent to `~::x`.
+* `x::`: Descendants of `x`, including the commits in `x` itself. Equivalent to
+  `x::visible_heads()` if no hidden revisions are mentioned.
+* `x..`: Revisions that are not ancestors of `x`. Equivalent to `~::x`, and
+  `x..visible_heads()` if no hidden revisions are mentioned.
 * `::x`: Ancestors of `x`, including the commits in `x` itself. Shorthand for
   `root()::x`.
 * `..x`: Ancestors of `x`, including the commits in `x` itself, but excluding
@@ -64,10 +71,11 @@ only symbols.
    to `x:: & ::y`. This is what `git log` calls `--ancestry-path x..y`.
 * `x..y`: Ancestors of `y` that are not also ancestors of `x`. Equivalent to
   `::y ~ ::x`. This is what `git log` calls `x..y` (i.e. the same as we call it).
-* `::`: All visible commits in the repo. Shorthand for
-  `root()::visible_heads()`. Equivalent to `all()`.
+* `::`: All visible commits in the repo. Equivalent to `all()`, and
+  `root()::visible_heads()` if no hidden revisions are mentioned.
 * `..`: All visible commits in the repo, but excluding the root commit.
-  Shorthand for `root()..visible_heads()`. Equivalent to `~root()`.
+  Equivalent to `~root()`, and `root()..visible_heads()` if no hidden revisions
+  are mentioned.
 * `~x`: Revisions that are not in `x`.
 * `x & y`: Revisions that are in both `x` and `y`.
 * `x ~ y`: Revisions that are in `x` but not in `y`.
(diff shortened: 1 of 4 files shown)


lib/src/revset.rs
```diff
@@ -1908,7 +1915,7 @@ fn fold_heads_range<St: ExpressionState>(
                     roots: filtered_range.roots,
                     heads: filtered_range
                         .heads
-                        .unwrap_or_else(RevsetExpression::visible_heads),
+                        .unwrap_or_else(|| Rc::new(RevsetExpression::ReferencedHeads)),
```
martinvonz: nit: add a `referenced_heads()` (or `all_heads()` if you agree with my other comment) constructor in the same vein as `all()` etc and use that here and below?

lib/src/revset.rs
```diff
@@ -231,6 +231,9 @@ pub type ResolvedRevsetExpression = RevsetExpression<ResolvedExpressionState>;
 pub enum RevsetExpression<St: ExpressionState> {
     None,
     All,
+    /// Visible heads and all referenced commits within the current expression
+    /// scope. Used as the default of `Range`/`DagRange` heads.
+    ReferencedHeads,
```
martinvonz: I think `AllHeads` seems like a better name

docs/revsets.md
```diff
@@ -52,10 +59,10 @@ only symbols.
 
 * `x-`: Parents of `x`, can be empty.
 * `x+`: Children of `x`, can be empty.
-* `x::`: Descendants of `x`, including the commits in `x` itself. Shorthand for
-  `x::visible_heads()`.
-* `x..`: Revisions that are not ancestors of `x`. Shorthand for
-  `x..visible_heads()`. Equivalent to `~::x`.
+* `x::`: Descendants of `x`, including the commits in `x` itself. Equivalent to
+  `x::visible_heads()` if no hidden revisions are mentioned.
```
martinvonz, line 63: optional: If we make an `all_heads()` function available in the revset language, this can again be simplified to say that `x::` is a shorthand for `x::all_heads()`.

lib/src/revset.rs
```diff
@@ -1523,6 +1545,18 @@ where
             operation,
             candidates,
         } => folder.fold_at_operation(operation, candidates)?,
+        RevsetExpression::WithinReference {
+            candidates,
+            commits,
+        } => {
+            let candidates = folder.fold_expression(candidates)?;
+            let commits = commits.clone();
```
martinvonz, line 1559: nit: for consistency, inline here or un-inline on line 1339

lib/src/revset.rs
yuja: I avoided `AllHeads` because it sounds like all historical heads. How about `ReferencedOrVisibleHeads`?

lib/src/revset.rs
yuja: I copied the style of the surrounding code. Maybe I choose un-inlined version here because it looked better in binary expression nodes.

lib/src/revset.rs
yuja: I could add a private function. I don't think there will be external callers.

docs/revsets.md
scott2000: Just a thought: if we define `all()` as `::all_heads()`, it might also be nice to also have `visible()` as `::visible_heads()` in the future (since now there's better support for hidden commits).

lib/src/revset.rs
martinvonz: I would not make the connection to historical heads, so I still prefer `AllHeads`. It matches `All` (which I also had not thought might include historical commits). If we add expressions for all historical heads, then `HistoricalHeads` sounds like a good name to me.

lib/src/revset.rs
martinvonz: I would slightly prefer that for consistency.

lib/src/revset.rs
martinvonz: Leaving as is makes sense then. I only noticed the context lines shown by default.

docs/revsets.md
yuja: Yeah, `visible()` could be added. I'm not sure if we should add `all_heads()`, though, because I don't have any use cases other than documenting purpose, and we'll need to calculate `Heads(visible + referenced)` to remove non-head commits.

lib/src/revset.rs
yuja: In context of https://github.com/jj-vcs/jj/pull/6782#discussion_r2155998760, if we're going to add `all_heads()`, this `ReferencedHeads` should have to be named differently since `AllHeads = Heads(ReferencedHeads)`.

martinvonz: I see. Since this is not just the heads of the combined set, I think `VisibleHeadsOrReferenced` is better (i.e. reverse from your suggestion), because the `Heads` made me think that this only refers to heads.

What would it take to make this expression refer to only the heads?

yuja: > What would it take to make this expression refer to only the heads?

Basically we'll need to evaluate `index.heads(referred + visible)`.

yuja: Renamed to `VisibleHeadsOrReferenced`, and added constructor fn.

PR #6783, opened 2025-06-18
github: bump `taiki-e/install-action` to 2.53.2

thoughtpolice:
Apparently there is a bug in the 2.53.0 version of `install-action` that means it will get a 403 from crates.io when it tries to download nextest. Ideally this should fix that, because we need it to unblock the CI pipeline.

# Checklist

If applicable:


ilyagr (approved):
Thanks!

PR #6790, opened 2025-06-19
file: add --separator option to `file show`

apoelstra:
Currently if you pass multiple files to `jj file show`, it is impossible to tell when one file ends and the next begins. This commit provides a CLI option to insert a separator string between the files.

This isn't perfect, of course -- no matter what separator string the user provides, there is a possibility that that string appears in the file. But in cases where the user has a string they are sure does not appear (as an example, they could use the git commit ID of the revision the file lives in; it is cryptographically impossible for this ID to occur somewhere in the file), this gives us some of the functionality of git cat-file --batch with very little code change or API surface increase.

Given multiple files, this inserts the string between them. Maybe instead we want to always prefix or suffix the string? I have no strong feelings.

### DIFF
--- a/cli/src/commands/file/show.rs
+++ b/cli/src/commands/file/show.rs
@@ -51,6 +51,8 @@ pub(crate) struct FileShowArgs {
         add = ArgValueCompleter::new(complete::revset_expression_all),
     )]
     revision: RevisionArg,
+    #[arg(long, short, default_value = "")]
+    separator: String,
     /// Paths to print
     #[arg(
         required = true,
@@ -83,7 +85,7 @@ pub(crate) fn cmd_file_show(
         }
         if !value.is_tree() {
             ui.request_pager();
-            write_tree_entries(ui, &workspace_command, [(path, Ok(value))])?;
+            write_tree_entries(ui, &workspace_command, &args.separator, [(path, Ok(value))])?;
             return Ok(());
         }
     }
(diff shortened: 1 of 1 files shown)


apoelstra:
I have not updated the CHANGELOG or insta snapshoti (which is why CI is failing) or added tests -- though I did test this myself, using a variant of this [sha512-computing function](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/main/github-merge.py#L206).

That function uses `git cat-file --batch` and takes about 100ms on my system. Using `jj file show` on individual files it takes 30-60 seconds (even with `--ignore-working-copy`, without which there is another factor 3 or so). Using `jj file show` with this patch and the commit ID as separator (which is guaranteed not to appear inline in any files) it takes about 400ms.

Will sign the CLA and fix CI if I can get a concept ACK on this.

PhilipMetzger:
I think providing a template (so making the command template-able) seems to be a better idea.

apoelstra:
Agreed. I am going to close this since I don't think I have the time to do such a big change anytime soon, and it doesn't make sense to add extra API surface like this for something that will/should be superceded by the template system.

PhilipMetzger:
imo, you should keep it open so a maintainer can give their opinion and _then_ decide to close it.

apoelstra:
Oh, sure. I thought you were a maintainer since you're so active here :).

yuja:
> I think providing a template (so making the command template-able) seems to be a better idea.

Yes, `jj show file` can have a template option of `TreeEntry` type. I'm not sure if the file content should be a part of the template output, but the other `show` commands don't include contents (or diffs) in template, so maybe not.

apoelstra:
Covered by https://github.com/jj-vcs/jj/pull/7198

PR #6805, opened 2025-06-22
id_prefix: check conflicting bookmarks/tags when calculating shortest length

### LINKED ISSUE #2416: Conflicting {change, commit}.shortest() and branch name
## Description

I have a change for which jj reports the shortest change ID is `l`, but turns out there's an (old) branch in the repo named `l` (which jj resolves to, when asked for `l`), but I would probably expect for jj to resolve to both (out of caution) or maybe consider branches (and other references) when reporting shortest, so this is not an issue in the first place.

Pick {change, commit}.shortest() from a random commit and create a new branch with the same name on another commit.

yuja:
Fixes #2416

# Checklist

If applicable:

### DIFF
--- a/lib/src/id_prefix.rs
+++ b/lib/src/id_prefix.rs
@@ -36,6 +36,7 @@ use crate::revset::RevsetResolutionError;
 use crate::revset::SymbolResolver;
 use crate::revset::SymbolResolverExtension;
 use crate::revset::UserRevsetExpression;
+use crate::view::View;
 
 #[derive(Debug, Error)]
 pub enum IdPrefixIndexLoadError {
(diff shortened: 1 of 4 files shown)


lib/src/id_prefix.rs
```diff
@@ -244,6 +255,21 @@ impl IdPrefixIndex<'_> {
...
+    (min_len..id_sym.len())
+        .find(|&n| {
+            // Tags, bookmarks, and Git refs have higher priority, but Git refs
+            // should include "/" char. Extension symbols have lower priority.
+            let prefix = &id_sym[..n];
+            view.get_tag(prefix.as_ref()).is_absent()
+                && view.get_local_bookmark(prefix.as_ref()).is_absent()
+        })
+        // No need to test conflicts with the full ID. We have to return some
+        // valid length anyway.
```
martinvonz, line 269: Perhaps we should check when resolving a bookmark or tag if it also matches a commit id or change id. If it does, we could print a warning. But that's something we can decide later.

yuja: We can also prioritize full change/commit ID resolution, but the situation where all prefixes are shadowed by tags/bookmarks would be pretty rare.

martinvonz: I was thinking of the case where someone has accidentally created a bookmark with a full commit id or change id as name because they misused some command. Perhaps something like this:

```
COMMIT_ID=$(...)
jj bookmark create my-bookmark $COMMIT_ID
```

I've seen people do that with hg.

yuja: I mean there would usually be shorter prefix `$COMMIT_ID[..len - 1]` to return.

A warning might be nice, btw.

PR #6813, opened 2025-06-23
Converting a repo into a co-located repo: .git dir handling

pmorch:
Handle any pre-existing .git directory to avoid the .git dir ending up in .git/git.

The docs do currently say:

> If you try this, we would appreciate feedback and bug reports.

I had tried (with gemini and search and others) to do this myself and had ended up with a (bad) .git directory, so this failed at first. So I suggest adding a guard for the problems I encountered.

Feel free to summarily decline this extra check if you think this is just too weird a corner case.

# Checklist

If applicable:


docs/git-compatibility.md
```diff
@@ -177,6 +177,10 @@ co-located repo like so:
 # Ignore the .jj directory in Git
 echo '/*' > .jj/.gitignore
 # Move the Git repo
+if [ -d .git ] ; then
```
ilyagr: I think it might be better to use `-e` here, since `.git` could be a file. You could also change the "Warning" to say ".git dir (or file)" to match.

docs/git-compatibility.md
```diff
@@ -177,6 +177,10 @@ co-located repo like so:
...
+    echo '*Warning*: Unexpected .git dir moved to .git.old'
+    mv .git .git.old
```
ilyagr: Perhaps this should be just `exit 1`, so that the user can figure out what to do? You'd also need to change the message above to match.

One situation when this is better is if `.git.old` already exists.

The one downside is that, if the user is pasting the commands into a shell, this will exit the shell. Still, that might be better than doing something unexpected; hopefully they'll read the script more carefully if this happens.

ilyagr:
Thanks! This is a good experience report.

I think it probably makes sense to warn people if they are about to mess up their repo. I'm not 100% sure whether this is the best approach: should we move the file? exit the shell? something else?. If nobody comes up with a better idea, I think exiting the shell might be best.

docs/git-compatibility.md
yuja: Just `set -e` and test first? I don't think the documentation example should have verbose error handling.

ilyagr: I'm worried about users that do this in their regular shell, but I'm not sure how to make that work well. 

I guess we could do as you suggest and `set +e` at the end? Or should we maybe wrap this in a block and a subprocess?

yuja: Hm, yeah. `exit` has the same problem, though. Just document that the user should check if `.git` already exists?

pmorch:
I think we can do better than the `exit` approach. If somebody copies the snippet to the terminal and executes it, the result will be to close the terminal immediately, leaving the user with no idea what just happened. How about this instead (`jj diff --git -w`):

```diff
diff --git a/docs/git-compatibility.md b/docs/git-compatibility.md
index 32a480c085..65abd749ec 100644
--- a/docs/git-compatibility.md
+++ b/docs/git-compatibility.md
@@ -174,6 +174,9 @@
 co-located repo like so:
 
 ```bash
+if [ -e .git ] ; then
+    echo '*Error*: .git exists - please remove it first'
+else
   # Ignore the .jj directory in Git
   echo '/*' > .jj/.gitignore
   # Move the Git repo
@@ -184,6 +187,7 @@
   git config --unset core.bare
   # Convince jj to update .git/HEAD to point to the working-copy commit's parent
   jj new && jj undo
+fi
```

If this is acceptable, I'll update my PR. I'm assuming you prefer me to modify the single commit and then force-push my branch. Let me know if you prefer me adding a second commit modifying the first one instead.

yuja:
As I said, I like simpler examples in docs. The user will have to review the content before executing it. It's also a bit harder to copy-paste if the script had control flow. If we really need error handling, the whole script can be executed in sub shell by wrapping in `(  )`, but that's not my preference.

So, just document the precondition?

pmorch:
Point taken. This instead?:

```diff
diff --git a/docs/git-compatibility.md b/docs/git-compatibility.md
index 32a480c085..3ec16d14a8 100644
--- a/docs/git-compatibility.md
+++ b/docs/git-compatibility.md
@@ -171,7 +171,11 @@
 
 A Jujutsu repo backed by a Git repo has a full Git repo inside, so it is
 technically possible (though not officially supported) to convert it into a
-co-located repo like so:
+co-located repo.
+
+First ensure that there is no pre-existing `.git` file or directory.
+
+Then:
 
 ```bash
 # Ignore the .jj directory in Git
```

pmorch:
I have now force-pushed a new commit: c1a3a98 - I hope I didn't mess anything up.  My first force-push using `jj` 🙂.

PhilipMetzger:
mind prefixing the commit message with `docs:` to align with our commit guidelines here: https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#commit-guidelines, otherwise LGTM

pmorch:
> mind prefixing the commit message with docs: to align with our commit guidelines...

Done.

PR #6814, opened 2025-06-23
revset: add change_id/commit_id(prefix) predicates

yuja:
# Checklist

If applicable:

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -211,6 +211,12 @@ revsets (expressions) as arguments.
 * `none()`: No commits. This function is rarely useful; it is provided for
   completeness.
 
+* `change_id(prefix)`: Commits with the given change ID prefix. It is an error
+  to use a non-unique prefix. Unmatched prefix isn't an error.
+
+* `commit_id(prefix)`: Commits with the given commit ID prefix. It is an error
+  to use a non-unique prefix. Unmatched prefix isn't an error.
+
 * `bookmarks([pattern])`: All local bookmark targets. If `pattern` is specified,
   this selects the bookmarks whose name match the given [string
   pattern](#string-patterns). For example, `bookmarks(push)` would match the
--- a/lib/src/hex_util.rs
+++ b/lib/src/hex_util.rs
@@ -12,29 +12,53 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![allow(missing_docs)]
+//! Hex string helpers.
 
 const REVERSE_HEX_CHARS: &[u8; 16] = b"zyxwvutsrqponmlk";
 
-fn to_forward_hex_digit(b: u8) -> Option<u8> {
-    let value = match b {
-        b'k'..=b'z' => b'z' - b,
-        b'K'..=b'Z' => b'Z' - b,
-        _ => return None,
-    };
-    if value < 10 {
-        Some(b'0' + value)
-    } else {
-        Some(b'a' + value - 10)
+fn reverse_hex_value(b: u8) -> Option<u8> {
+    match b {
+        b'k'..=b'z' => Some(b'z' - b),
+        b'K'..=b'Z' => Some(b'Z' - b),
+        _ => None,
     }
 }
 
-pub fn to_forward_hex(reverse_hex: impl AsRef<[u8]>) -> Option<String> {
-    reverse_hex
-        .as_ref()
-        .iter()
-        .map(|b| to_forward_hex_digit(*b).map(char::from))
-        .collect()
+/// Decodes `reverse_hex` as hex string using `z-k` "digits".
+pub fn decode_reverse_hex(reverse_hex: impl AsRef<[u8]>) -> Option<Vec<u8>> {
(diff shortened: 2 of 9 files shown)


yuja:
I'll rebase this PR on top of https://github.com/jj-vcs/jj/pull/6824.

lib/src/hex_util.rs
```diff
@@ -12,29 +12,79 @@
...
+    /// Bytes decoded from even-length hex string.
+    Ok(Vec<u8>),
+    /// Bytes decoded from odd-length hex string, with the padded last byte.
+    Odd(Vec<u8>),
+    /// Hex string had invalid character.
+    InvalidChar,
+}
+
+impl DecodeHexResult {
+    fn from_decoded(decoded: Option<(Vec<u8>, bool)>) -> Self {
```
martinvonz: Why not make `decode_hex_inner()` return a `DecodeHexResult` directly instead of having this conversion step? Will there be future callers of `decode_hex_inner()` that don't want this type of output?

Similarly, if there are no other callers planned, it would be simpler to inline `decode_hex_inner()`.

lib/src/hex_util.rs
```diff
@@ -12,29 +12,79 @@
...
 
-pub fn to_forward_hex(reverse_hex: impl AsRef<[u8]>) -> Option<String> {
-    reverse_hex
-        .as_ref()
-        .iter()
-        .map(|b| to_forward_hex_digit(*b).map(char::from))
-        .collect()
+/// Decoded hex or error, which may include odd-length byte.
+#[derive(Clone, Debug)]
+pub enum DecodeHexResult {
```
martinvonz: This type seems very similar to `HexPrefix` (well `Result<HexPrefix, ...>`). Do we need both?

lib/src/hex_util.rs
yuja: Just because `?` can be used if the return type is `Option` or `Result`.

> Similarly, if there are no other callers planned, it would be simpler to inline decode_hex_inner().

I'm thinking of porting (forward) `hex::decode()` and remove `hex` crate. It's not important (and our implementation would be a bit slower), though.

lib/src/hex_util.rs
yuja: They're similar, but it depends on caller whether odd-length bytes is acceptable or not, so I think the intermediate `Result`-ish type helps clarify that.

lib/src/hex_util.rs
```diff
@@ -12,29 +12,79 @@
...
+    fn from_decoded(decoded: Option<(Vec<u8>, bool)>) -> Self {
+        match decoded {
+            Some((bytes, false)) => DecodeHexResult::Ok(bytes),
+            Some((bytes, true)) => DecodeHexResult::Odd(bytes),
+            None => DecodeHexResult::InvalidChar,
+        }
+    }
+
+    /// Extracts bytes decoded from even-length hex string.
+    pub fn ok(self) -> Option<Vec<u8>> {
```
martinvonz: Since some callers consider odd-length inputs okay, should we call this `.even()` or `.even_ok()` or similar?

We would perhaps need to rename `.ok_or_odd()` then (since we're no longer saying that "even == okay"), but that's also related to my other comment.

yuja: I consider that odd-length hex is usually an error, hence `Ok|Odd`.

Should we add `decode_reverse_hex() -> Option<_>` and `decode_reverse_hex_maybe_odd() -> Option<(_, bool)>`?

martinvonz: Yes, I think I would like that better, thanks. Then we won't need `DecodeHexResult` at all, right?

yuja: Added `decode_reverse_hex()` and `decode_reverse_hex_prefix()`.

docs/revsets.md
```diff
@@ -211,6 +211,12 @@ revsets (expressions) as arguments.
 * `none()`: No commits. This function is rarely useful; it is provided for
   completeness.
 
+* `change_id(prefix)`: Commits with the given change ID prefix. It is an error
+  to use a non-unique prefix. Unmatched prefix isn't an error.
```
martinvonz, line 215: Do you think users will mostly likely want unmatched prefixes to not be an error? I feel like I would expect an error. We can of course change it later if we make the wrong guess about what users want, so I'm fine with trying either.

yuja: Returning error might be useful, but I don't want to introduce inconsistency. If we make `change_id(unknown)` error out, I would also want to change `bookmarks(unknown)` to be an error. Suppose we'll make diverged symbol resolution fail, `change_id()` and `bookmarks()` will be used as non-failing alternative to the symbol expression. These functions should have similar behavior.

martinvonz: I suppose another option is to have `change_id(xyz, allow_divergent=true)` and make it error out otherwise?

Either way, I'm fine with this PR as is, and we can change it later if we get feedback about it.

yuja: or add `exactly_one(change_id(xyz))`.

I'll merge this PR with the current form.

PR #6818, opened 2025-06-23
lib: add `rustversion`, turn off a clippy lint in nightly in tests

ilyagr:
This turns off the `clippy::cloned_ref_to_slice_refs` lint in some tests and fixes it in others, for Rust 1.89+. This seems to make `cargo clippy --workspace --all-targets --all-features` work in stable, beta, and nightly (1.89).

This depends on the `rustversion` crate. Other than that, it's based on Austin's https://github.com/jj-vcs/jj/pull/6705.

Co-authored-by: aseipp@pobox.com


ilyagr:
Currently, the lint exception is at the level of `mod test {...}`. It would probably be better to pull it down to the individual test level. If we are happy with this approach, I might do it here or in a separate PR.

On second though, this might be better to do once our clippy CI check tests this lint (so, in a couple of months).

thoughtpolice (approved):
Nice, seems good to me. Thanks!

ilyagr:
Ok, I'll remove the todos then.

PR #6819, opened 2025-06-24
squash: allow --interactive and fileset to be used together

glehmann:
everything required seems to be in place, but this was explicitly forbidden in the command line parser configuration.

# Checklist

If applicable:


cli/src/commands/squash.rs
```diff
@@ -103,7 +103,7 @@ pub(crate) struct SquashArgs {
     tool: Option<String>,
     /// Move only changes to these paths (instead of all paths)
     #[arg(
-        conflicts_with_all = ["interactive", "tool"],
+        conflicts_with_all = ["tool"],
```
yuja: nit: can be `conflict_with = "tool"`

yuja: Ah, no, just delete `conflict_with_all`? `--tool` == `--interactive`.

Issue #2837, opened 2024-01-16
FR: Add a `jj track` command for tracking otherwise ignored files


martinvonz:
**Is your feature request related to a problem? Please describe.**

If you have ignored files that you want to track, you currently have to temporary un-ignore them, which can be quite annoying, especially if they're in a directory that's ignored (unless you want to track everything in that directory).

**Describe the solution you'd like**

`jj track <pattern>`  seems logical (we already have `jj untrack <pattern>`).

benbrittain:
`track` is a pretty easy command to implement, but it has some interesting overlap with the forget mechanism. If there is a `track`, should `untrack` automatically add a file to be forgotten?

martinvonz:
`untrack` doesn't currently automatically add files to the ignore patterns, mostly because it's not clear if the user wants that particular file or a more general pattern to be ignored. @hooper, did you say you're also looking into this?

hooper:
Not really.

I'm trying to implement `track` by refactoring `GitIgnoreFile` into a trait to allow it to interact with an impl that uses a `Matcher`, so I can just construct that and plug it into `snapshot()`.

It would be a more invasive change to act further on the decisions to ignore paths during snapshot by persisting them. I agree with Martin that there are issues with defining a good UX for that.

Such a feature could become unpopular for spamming `.gitignore` files with confusing entries, even if we do some basic optimizations. Then it might turn into a tempting target for scope creep to "fix" that. Also, it's not obvious to me that every `<pattern>` we might want to accept would map reasonably to `.gitignore` patterns. For examples of how that could get harder or more ambiguous in the future, see https://github.com/martinvonz/jj/issues/2915.

I've occasionally imagined something like `jj [un]ignore <pattern>` that does this more directly, but it would have at least as many issues as doing it during `[un]track`.

Another detail we would want to consider is the UX for how we determine *which* `.gitignore` file we manipulate under which circumstances. I don't think something along the lines of `jj config set --user|--repo` would be sufficient.

So, I don't mean to shoot the idea down entirely, but I don't think it's low hanging fruit.

Nemo157:
[...]

PR #6820, opened 2025-06-24
status: omit untracked files with --ignore-untracked

71:
I have a repo with a lot of untracked files, so `jj st` displays many untracked paths. This hides these paths. There were a lot of alternatives I considered, but I went for the simplest solution:

1. What should the option be named? `--ignore-untracked`? `--hide-untracked`? `--no-untracked`? `--only-tracked`? `--tracked-only`?
1. Should the default change to ignore untracked files instead? In which case this would be `--keep-untracked`, `--include-untracked`, ...
1. Instead of a boolean, we could have `--untracked-paths-limit <n>`, so users have an overview of untracked files without it being overwhelming.
1. Instead of only hiding files in the output of `jj status`, we could instead modify the implementation of `workspace_helper_with_stats()` to completely ignore untracked paths, which I assume could speed up the command if there are many untracked files. After a quick look at `workspace_helper_with_stats()`, this seems much harder to implement.
1. Instead of displaying "4 untracked paths ignored", we could display nothing, or "Untracked paths: 4", or "Ignored 4 untracked paths".

Feel free to comment on any of the above points (or any other point), I'm absolutely fine with making further changes.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Log node templates are now specified in toml rather than hardcoded.
 
+* `jj status` now accepts `--ignore-untracked` to hide the "Untracked paths"
+  section.
+
 ### Fixed bugs
 
 * `jj file annotate` can now process files at a hidden revision.
--- a/cli/src/commands/status.rs
+++ b/cli/src/commands/status.rs
@@ -46,6 +46,9 @@ pub(crate) struct StatusArgs {
     /// Restrict the status display to these paths
     #[arg(value_name = "FILESETS", value_hint = clap::ValueHint::AnyPath)]
     paths: Vec<String>,
+    /// Ignore untracked files
+    #[arg(long)]
+    ignore_untracked: bool,
 }
 
 #[instrument(skip_all)]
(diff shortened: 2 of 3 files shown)


joyously:
> 2\. Should the default change to ignore untracked files instead? 

I would vote for this, since in my view, untracked and ignored should be the same thing. 
Other VCSs use a `-u` option to show untracked.
Options starting with `i` can be mistaken for `--interactive`.

yuja:
> 4. Instead of only hiding files in the output of `jj status`, we could instead modify the implementation of `workspace_helper_with_stats()` to completely ignore untracked paths, which I assume could speed up the command if there are many untracked files.

A better option might be to ignore files by your local `.git/info/exclude`, and do `jj file track` as needed. We'll need a better support for tracking ignored files, though. See also https://github.com/jj-vcs/jj/issues/2837.

71:
> A better option might be to ignore files by your local `.git/info/exclude`, and do `jj file track` as needed.

Thanks for the suggestion. That's what I currently do, but I wish I didn't have to take additional actions as I already set `snapshot.auto-track = "none()"`. As an example, when working in a C++ project `build` is in `.gitignore`, but I often create other builds with specific C++ flags (e.g. `build-asan`) which are not in `.gitignore`. They are build directories so they have a lot of files, and `snapshot.auto-track` does exactly what I want and doesn't track these files.
- I could add `build-*` to `.gitignore`, but that doesn't work for other similar scenarios.
- I could add `build-asan` to `.git/info/exclude`, but it (slightly) defeats the purpose of working with files quickly with little interaction needed. For instance if I want to dump the output of a command somewhere, I can't "just" do `command > tmp-out`, I need to follow that with `echo /tmp-out > .git/info/exclude`.
- I could add `snapshot.*` to the global `.gitignore`, but now my short-to-type commands with short file names are longer.

All in all this is definitely not necessary for me to use `jj`, but it would make it slightly more convenient.

**Edit**:

> Options starting with `i` can be mistaken for `--interactive`.

I intentionally didn't add a short flag for this argument as no such flag would be obvious IMO. In my case the intention is to have an alias that invokes `status --ignore-untracked`, so I don't mind the longer option name.

yuja:
> > A better option might be to ignore files by your local `.git/info/exclude`, and do `jj file track` as needed.
> 
> Thanks for the suggestion. That's what I currently do, but I wish I didn't have to take additional actions as I already set `snapshot.auto-track = "none()"`.

The `snapshot.auto-track` option primarily exists for avoiding accidental snapshotting. There should usually be zero or a few untracked files because as you mentioned, scanning untracked files and directories aren't cheap. From the design standpoint, your use case should be achieved by ignore patterns.

71:
That's fair. I'll close this PR and try to get more used to `.git/info/exclude` then.

PR #6822, opened 2025-06-24
better handling of non-UTF-8 workspace/repository paths

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -2543,7 +2543,7 @@ jj git init --colocate",
         ) => internal_error_with_message("The repository appears broken or inaccessible", err),
         WorkspaceLoadError::StoreLoadError(StoreLoadError::Signing(err)) => user_error(err),
         WorkspaceLoadError::WorkingCopyState(err) => internal_error(err),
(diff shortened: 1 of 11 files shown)


martinvonz (approved):

magneticflux-:
The newly added test `test_init::test_init_load_non_utf8_path` fails on ZFS with `utf8only=on` set. The error is as expected; the file cannot be created:
```
thread 'test_init::test_init_load_non_utf8_path' panicked at /build/source/lib/testutils/src/git.rs:48:6:
    called `Result::unwrap()` on an `Err` value: Init(CreateDirectory { source: Os { code: 84, kind: Uncategorized, message: "Invalid or incomplete multibyte or wide character" }, path: "/build/jj-test-1KkuKP/git\xE0/.git" })
```
Could we instead run the tests regardless of filesystem and just try to create an empty file/directory before `git::init(&git_repo_path);`, skipping the rest of the test if it fails?

yuja:
Hmm, yeah, we'll need runtime check for filesystem (similar to test_local_working_copy.rs.)

PR #6824, opened 2025-06-25
revset: refactor alias expansion stack and error handling

yuja:
# Checklist

If applicable:


lib/src/revset_parser.rs
```diff
@@ -824,25 +824,43 @@ pub fn expect_literal<T: FromStr>(
 }
 
 /// Applies the give function to the innermost `node` by unwrapping alias
```
martinvonz: nit: could fix the typo in "give function" (should be "given function") here while at it

yuja: Fixed.

PR #6826, opened 2025-06-25
git: add jj git sync command

### LINKED ISSUE #1039: Implement `jj sync` command

0xdeafbeef:
# Checklist

If applicable:

Closes #1039

### DIFF
--- a/lib/src/git_sync.rs
+++ b/lib/src/git_sync.rs
@@ -1,0 +1,693 @@
+// Copyright 2026 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Rebases selected local commits after Git remote bookmarks move forward.
+//!
+//! This module provides the rebase step for a `git pull --rebase`-like
+//! workflow. The caller selects local commits before fetching, imports the
+//! updated remote bookmarks, and then asks this module to rebase the selected
+//! commits onto the new remote targets.
+//!
+//! Multiple independent remote bookmark updates can be applied in one rewrite.
+//! Bookmarks and working copies follow the rewritten commits. Heads of
+//! anonymous branches are updated as well, so those branches remain visible.
+//!
+//! The operation is intentionally conservative. Unrelated remote updates are
+//! ignored. It rejects deletions, conflicts, force-pushes, incomplete
+//! selections, and overlapping updates that cannot be applied without losing
+//! commits or parent relationships.
+//!
+//! See [`sync_imported_refs()`] for the phase and transaction-safety contract.
+
+use std::collections::HashMap;
+use std::collections::VecDeque;
+
+use futures::future::try_join_all;
+use indexmap::IndexMap;
+use indexmap::IndexSet;
+use itertools::Itertools as _;
+use thiserror::Error;
+
+use crate::backend::BackendError;
+use crate::backend::CommitId;
+use crate::commit::Commit;
+use crate::git::GitImportRefUpdate;
+use crate::git::GitImportStats;
+use crate::index::IndexError;
+use crate::op_store::RefTarget;
+use crate::ref_name::RefNameBuf;
+use crate::ref_name::RemoteRefSymbolBuf;
+use crate::repo::MutableRepo;
+use crate::repo::Repo as _;
+use crate::rewrite::EmptyBehavior;
+use crate::rewrite::RebaseOptions;
+use crate::rewrite::RebasedCommit;
+use crate::rewrite::rebase_commit_with_options;
+
+/// Rebases selected commits onto updated tracked Git remote bookmarks.
+///
+/// For each fast-forward update, this function rebases selected commits with a
+/// parent between the old and new remote targets onto the new target. Their
+/// selected descendants are rewritten as part of the same operation. Multiple
+/// independent updates can be applied in one rewrite.
+///
+/// # Rebase policy
+///
+/// The caller's empty-commit and merge-simplification policies apply to rewrite
+/// roots. Other rewritten commits keep empty commits and explicit merge
+/// parents.
+///
+/// Reference rewriting for the whole operation remains controlled by
+/// `rebase_options.rewrite_refs`.
+///
+/// # Returns
+///
+/// The returned map contains every commit rewritten or abandoned by this
+/// operation, keyed by its original commit ID.
+///
+/// Unaffected selected commits and commits outside the selection are omitted.
+///
+/// # Errors
+///
+/// Returns an error if:
+///
+/// - the repository has pending rewrite mappings;
+/// - an imported target is absent or conflicted, or did not move by
+///   fast-forward;
+/// - imported updates disagree or overlap in a way that could lose a selected
+///   commit or parent edge;
+/// - the selection omits an intermediate commit required by the rebase; or
+/// - an imported destination would itself be rewritten.
+///
+/// Backend and index failures are propagated.
+///
+/// # Preconditions
+///
+/// - The commits to synchronize must be selected before fetching, and their IDs
+///   passed in `selected_commit_ids`.
+/// - `import_stats` must come from the immediately following Git import on the
+///   same `mut_repo`.
+/// - The function must be called before processing rewrite mappings left by the
+///   import or otherwise rewriting the selected commits.
+///
+/// The function checks that the repository has no pending rewrite mappings,
+/// but cannot verify when the commits were selected or where `import_stats`
+/// came from.
+///
+/// If the import leaves rewrite mappings, the pre-fetch commit IDs can no
+/// longer be used safely. The caller must abort or use another fallback.
+///
+/// # Transaction safety
+///
+/// The caller must discard the transaction if this function returns an error
+/// or if its future is cancelled before completion. The function cannot roll
+/// back partial in-memory mutations.
+pub async fn sync_imported_refs(
+    mut_repo: &mut MutableRepo,
+    import_stats: &GitImportStats,
+    selected_commit_ids: Vec<CommitId>,
+    rebase_options: &RebaseOptions,
+) -> Result<HashMap<CommitId, RebasedCommit>, GitSyncError> {
+    if mut_repo.has_rewrites() {
+        return Err(GitSyncError::PendingRewrites);
+    }
+
+    let selected = SelectedGraph::load(mut_repo, selected_commit_ids).await?;
+    let moves = derive_boundary_moves(mut_repo, &selected, import_stats).await?;
+    let plan = GitSyncPlanBuilder::new(mut_repo, selected, moves)
+        .build()
+        .await?;
+    plan.apply(mut_repo, rebase_options).await
+}
+
+/// Why an imported remote bookmark update cannot be synchronized.
+#[derive(Clone, Debug, Eq, Error, PartialEq)]
+pub enum GitSyncUnsupportedReason {
+    /// The old remote target does not resolve to a single commit.
+    #[error("the old remote target {target:?} is conflicted")]
+    OldTargetConflicted {
+        /// The conflicted target recorded before import.
+        target: RefTarget,
+    },
+    /// The imported remote bookmark no longer points to a commit.
+    #[error("the imported target is absent")]
+    NewTargetAbsent,
+    /// The imported remote bookmark points to a conflicted target.
+    #[error("the imported target {target:?} is conflicted")]
+    NewTargetConflicted {
+        /// The conflicted target recorded after import.
+        target: RefTarget,
+    },
+    /// The imported target is not a descendant of the previous target.
+    #[error("moving from {old} to the imported target {new} is not a fast-forward")]
+    NonFastForward {
+        /// The previous remote commit.
+        old: CommitId,
+        /// The imported remote commit.
+        new: CommitId,
+    },
+}
+
+/// Errors from synchronizing selected commits with imported Git updates.
+#[derive(Debug, Error)]
+pub enum GitSyncError {
+    /// The transaction already contains rewrite mappings.
+    #[error("the transaction has pending rewrites")]
+    PendingRewrites,
+    /// An imported remote bookmark update cannot be synchronized.
+    #[error("unsupported remote bookmark change {symbol}: {reason}")]
+    UnsupportedRemoteChange {
+        /// The imported remote reference that changed.
+        symbol: RemoteRefSymbolBuf,
+        /// The reason the imported change cannot be synchronized.
+        reason: GitSyncUnsupportedReason,
+    },
+    /// Tracked remote bookmarks with the same name were imported with
+    /// different destinations.
+    #[error(
+        "tracked remote bookmarks named {name} have different destinations: {destinations:?}",
+        name = .name.as_str()
+    )]
+    AmbiguousBookmark {
+        /// The tracked bookmark name shared by the updates.
+        name: RefNameBuf,
+        /// The distinct destinations observed for the bookmark.
+        destinations: Vec<CommitId>,
+    },
+    /// The same old target would move to different destinations.
+    #[error("upstream boundary {old} has different destinations: {destinations:?}")]
+    AmbiguousBoundary {
+        /// The old upstream commit shared by the updates.
+        old: CommitId,
+        /// The distinct replacement destinations observed for that commit.
+        destinations: Vec<CommitId>,
+    },
+    /// The same parent of a selected commit would be replaced with different
+    /// destinations.
+    #[error(
+        "selected commit {commit_id} would replace parent {parent_id} with different \
+         destinations: {destinations:?}"
+    )]
+    ConflictingParentReplacement {
+        /// The selected commit whose parent list would conflict.
+        commit_id: CommitId,
+        /// The parent commit targeted for replacement.
+        parent_id: CommitId,
+        /// The distinct replacement destinations requested for that parent.
+        destinations: Vec<CommitId>,
+    },
+    /// A commit would be both a rewrite root and a descendant of another
+    /// rewrite root.
+    #[error(
+        "selected commits {commit_ids:?} would be rewrite roots for one remote update and \
+         structural descendants of another"
+    )]
+    ConflictingRewriteRoles {
+        /// The selected commits that would receive both rewrite roles.
+        commit_ids: Vec<CommitId>,
+    },
+    /// A selected parent would be replaced even though it is already being
+    /// rewritten.
+    #[error(
+        "selected commit {commit_id} would replace selected parent {parent_id} with imported \
+         target {destination}"
+    )]
+    OverlappingRewrite {
+        /// The selected commit whose parent would be replaced.
+        commit_id: CommitId,
+        /// The selected parent that is already part of the rewrite.
+        parent_id: CommitId,
+        /// The imported target requested as the replacement parent.
+        destination: CommitId,
+    },
+    /// The selection omits an intermediate commit needed to follow an update.
+    #[error(
+        "selected commits {commit_ids:?} cannot follow remote update {old} -> {new} without \
+         rewriting unselected intermediate commits"
+    )]
+    SelectionBoundary {
+        /// The old upstream commit.
+        old: CommitId,
+        /// The imported upstream commit.
+        new: CommitId,
+        /// Selected commits that cannot be reached from the rewrite roots.
+        commit_ids: Vec<CommitId>,
+    },
+    /// An imported destination would itself be rewritten by the plan.
+    #[error("imported destinations would be rewritten: {commit_ids:?}")]
+    MovingDestination {
+        /// Imported destinations that are also part of the rewrite plan.
+        commit_ids: Vec<CommitId>,
+    },
+    /// A backend operation failed while loading or rewriting commits.
+    #[error(transparent)]
+    Backend(#[from] BackendError),
+    /// The commit index could not answer a graph query.
+    #[error(transparent)]
+    Index(#[from] IndexError),
+}
+
+/// The validated set of commits and parent overrides to rewrite.
+#[derive(Debug)]
+struct GitSyncPlan {
+    commits_to_rewrite: Vec<Commit>,
+    root_parent_overrides: HashMap<CommitId, Vec<CommitId>>,
+}
+
+impl GitSyncPlan {
+    async fn apply(
+        self,
+        mut_repo: &mut MutableRepo,
+        rebase_options: &RebaseOptions,
+    ) -> Result<HashMap<CommitId, RebasedCommit>, GitSyncError> {
+        let Self {
+            commits_to_rewrite,
+            root_parent_overrides,
+        } = self;
+        if commits_to_rewrite.is_empty() {
+            return Ok(HashMap::new());
+        }
+
+        // Rewrite roots honor the caller's empty-commit and merge-simplification
+        // policies. Structural descendants keep empty commits and retain explicit
+        // ancestor merge parents. Reference rewriting remains caller-controlled.
+        let descendant_options = RebaseOptions {
+            empty: EmptyBehavior::Keep,
+            rewrite_refs: rebase_options.rewrite_refs.clone(),
+            simplify_ancestor_merge: false,
+        };
+        let mut rebased_commits = HashMap::new();
+        mut_repo
+            .transform_commits_exact(
+                commits_to_rewrite,
+                &root_parent_overrides,
+                &rebase_options.rewrite_refs,
+                async |rewriter| {
+                    let old_id = rewriter.old_commit().id().clone();
+                    let options = if root_parent_overrides.contains_key(&old_id) {
+                        rebase_options
+                    } else {
+                        &descendant_options
+                    };
+                    let rebased = rebase_commit_with_options(rewriter, options).await?;
+                    rebased_commits.insert(old_id, rebased);
+                    Ok(())
+                },
+            )
+            .await?;
+        Ok(rebased_commits)
+    }
+}
+
+/// The old and new targets of a remote bookmark fast-forward.
+#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
+struct BoundaryMove {
+    old: CommitId,
+    new: CommitId,
+}
+
+/// A remote bookmark fast-forward and the selected commits it affects.
+struct BoundarySelection {
+    boundary: BoundaryMove,
+    affected_commits: IndexSet<CommitId>,
+    rewrite_roots: IndexSet<CommitId>,
+}
+
+/// Builds and validates a sync plan.
+struct GitSyncPlanBuilder<'repo> {
+    repo: &'repo MutableRepo,
+    selected: SelectedGraph,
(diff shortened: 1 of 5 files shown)


0xdeafbeef:
Wrote it according https://github.com/jj-vcs/jj/issues/1039#issuecomment-2465742400.

cli/tests/test_git_sync.rs
```diff
@@ -0,0 +1,403 @@
+// Copyright 2020-2023 The Jujutsu Authors
```
PhilipMetzger: nit: 2025 since its a new file

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unreleased]
 
 ### Release highlights
+* git: Introduce `jj git sync` for a `git pull --rebase`-like workflow across multiple bookmarks.
```
PhilipMetzger: nit: newline before the entry

cli/src/commands/git/sync.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// [string pattern]:
+    ///     https://jj-vcs.github.io/jj/latest/revsets#string-patterns
+    #[arg(
+        long = "remote",
+        short = 'r',
+        value_name = "REMOTE",
+        value_parser = StringPattern::parse,
+        add = ArgValueCandidates::new(complete::git_remotes),
+    )]
+    remotes: Vec<StringPattern>,
```
PhilipMetzger: Nit: this is missing an argument with the bookmarks to create merge commits for, since that workflow also exists. I noted the same thing in the @essiene's design doc.

PhilipMetzger (commented):
🎉 finally, here are some minor nits.

cli/src/commands/git/sync.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    let mut tx = workspace_command.start_transaction();
+
+    // Capture the pre-fetch state of remote tracking bookmarks
+    let mut pre_fetch_heads: HashMap<RemoteRefSymbolBuf, CommitId> = HashMap::new();
+
+    for remote in &remotes {
+        for (name, remote_ref) in tx.repo().view().remote_bookmarks(remote) {
+            // Only capture non-conflicted targets
+            if let Some(commit_id) = remote_ref.target.as_normal() {
```
yuja: I think we need to use the local bookmark tracking the remote. Remote bookmarks often point to hidden commits.

cli/src/commands/git/sync.rs
```diff
@@ -0,0 +1,261 @@
...
+                let symbol = name.to_remote_symbol(remote).to_owned();
+                pre_fetch_heads.insert(symbol, commit_id.clone());
+            }
+        }
+    }
+
+    // Perform the fetch (fetch all branches to properly handle merged/deleted
+    // branches)
+    let fetch_branches = vec![StringPattern::everything()];
+    do_git_fetch(ui, &mut tx, &remotes, &fetch_branches)?;
```
yuja: It's probably better to use `GitFetch` API. iirc, the idea was to add something like `git_fetch.rebase_descendants()`. We might have to refactor `import_refs()` to get enough information to rebase descendants of old bookmarks.

0xdeafbeef:
Status report - i'll try to finish it on this weekend. Daily driving it, found a nasty bug :)

0xdeafbeef:
@yuja 
> I think we need to use the local bookmark tracking the remote. Remote bookmarks often point to hidden commits.

It didn't work well, so i've a hybrid solution.

The problem which i debuged for so long was when 2 bookmarks pointed to the same comit and second bookmark moved in remote. Sync will move my local bookmark to the same position.

Now daily drive it for a week, no issues so far.

cli/tests/test_git_sync.rs
0xdeafbeef: done

yuja:
> > I think we need to use the local bookmark tracking the remote. Remote bookmarks often point to hidden commits.
> 
> It didn't work well, so i've a hybrid solution.
> 
> The problem which i debuged for so long was when 2 bookmarks pointed to the same commit and second bookmark moved in remote. Sync will move my local bookmark to the same position.

Hmm, I'm not sure how it's related to whether to use local/remote bookmarks, but that sounds good. afaiui, `sync` should simulate the movement that would occur when remote changes were made locally. If the same commit were rewritten locally, these two bookmarks would be moved to the new commit.

For tracked remote bookmarks, the latest local states are represented by the tracking local bookmarks, so I think it makes sense to use the corresponding local bookmarks. For untracked remote bookmarks, we'll need to use remote bookmarks.

EDIT: That said, if the same change were rewritten concurrently, the result should be divergence. So I'm getting less sure about that. It might be safer to leave descendants unchanged if the unsynced tracked remote bookmarks were rewritten at the remote.

jgilchrist:
@0xdeafbeef I've tried this out and unless I'm misunderstanding, it doesn't seem to be working as expected for me. Prior to running `jj git sync` things look like:

image

i.e. I've got a few changes, some with bookmarks, all based on `main`. `main@upstream` has a single new commit, so I'm expecting that after the sync, all of these changes are rebased. However, after `jj git sync`, I see:

image

i.e. all of the WIP changes are hidden, with the working copy now on top of the new `main` (synced with main@upstream) and with all of the bookmarks now pointing to that commit. The output of the sync shows:

image

It looks like we do attempt to rebase the changes ('Rebasing 7 commits') but then they are somehow not rebased successfully ('7 already merged').

Any idea what might be going wrong here?

zx8:
Please don't let this PR become stale! I'm super excited to see it merged - is there anything you need in terms of testing/data to get this over the line @0xdeafbeef? 🙏

0xdeafbeef:
> Please don't let this PR become stale! I'm super excited to see it merged - is there anything you need in terms of testing/data to get this over the line @0xdeafbeef? 🙏

Example of @jgilchrist showed that my impl is unsound and works only in basic cases. I'll try to fix it in the next 2 days and push updated version. Sync won't have merge support cause this rabbit hole appears to be deep enough :)

0xdeafbeef:
> @0xdeafbeef I've tried this out and unless I'm misunderstanding, it doesn't seem to be working as expected for me. Prior to running `jj git sync` things look like:
> image
> 
> i.e. I've got a few changes, some with bookmarks, all based on `main`. `main@upstream` has a single new commit, so I'm expecting that after the sync, all of these changes are rebased. However, after `jj git sync`, I see:
> image
> 
> i.e. all of the WIP changes are hidden, with the working copy now on top of the new `main` (synced with main@upstream) and with all of the bookmarks now pointing to that commit. The output of the sync shows:
> image
> 
> It looks like we do attempt to rebase the changes ('Rebasing 7 commits') but then they are somehow not rebased successfully ('7 already merged').
> 
> Any idea what might be going wrong here?

strangely that i've not caught this during my daily-driving. 

Problem was that instead of

```

  Before sync:                After sync:

      C---D---E (local)                  C'--D'--E' (local rebased)
     /                                  /
    A---B (origin)             A---B---F (origin with new commit F)
```
    I was doing this:
    
```
     Before sync:                After sync:

      C---D---E (local)          [commits C,D,E abandoned]
     /
    A---B (origin)             A---B---F (origin, all bookmarks collapsed here)
```
    
    Because of incorrect api usage. It worked fine with a single  feature branch, but moved branches to the same commit in case of several branches, which caused it to think that these commits are duplicate for F and should be abandoned.

jgilchrist:
Thanks @0xdeafbeef! Tried this again with a similar setup and my changes get rebased as I expected 👍 

However some other changes also got rebased too. In my case it looks like everything in `descendants(master@upstream)` got rebased, but this included immutable changes that were on untracked remote feature branches, i.e.:

### Before:

```mermaid
gitGraph
    commit id: "main"
    branch local
    checkout local
    commit id: "a"
    commit id: "b"
    checkout main
    branch "another@upstream"
    commit id: "c"
```

### After (expected)

```mermaid
gitGraph
    commit id: "oldmain"
    branch "another@upstream"
    commit id: "c"
    checkout main
    commit id: "main"
    branch local
    checkout local
    commit id: "a'"
    commit id: "b'"
```

### After (actual)

```mermaid
gitGraph
    commit id: "oldmain"
    checkout main
    commit id: "main"
    branch local
    checkout local
    commit id: "a'"
    commit id: "b'"
    checkout main
    branch "another@upstream"
    commit id: "c'"
```

0xdeafbeef:
> Thanks @0xdeafbeef! Tried this again with a similar setup and my changes get rebased as I expected 👍
> 
> However some other changes also got rebased too. In my case it looks like everything in `descendants(master@upstream)` got rebased, but this included immutable changes that were on untracked remote feature branches, i.e.:
> ### Before:
> 
> ```mermaid
> gitGraph
>     commit id: "main"
>     branch local
>     checkout local
>     commit id: "a"
>     commit id: "b"
>     checkout main
>     branch "another@upstream"
>     commit id: "c"
> ```
> 
> ### After (expected)
> 
> ```mermaid
> gitGraph
>     commit id: "oldmain"
>     branch "another@upstream"
>     commit id: "c"
>     checkout main
>     commit id: "main"
>     branch local
>     checkout local
>     commit id: "a'"
>     commit id: "b'"
> ```
> 
> ### After (actual)
> 
> ```mermaid
> gitGraph
>     commit id: "oldmain"
>     checkout main
>     commit id: "main"
>     branch local
>     checkout local
>     commit id: "a'"
>     commit id: "b'"
>     checkout main
>     branch "another@upstream"
>     commit id: "c'"
> ```

yep:( . already working on it. When we rebase master on top of master@remote we automatically rebase all feature branches branched from master, which causes big mess

0xdeafbeef:
@jgilchrist i've updated it

0xdeafbeef:
@yuja @PhilipMetzger I've changed how sync works because the previous design had serious flaws - when syncing
  main it would rebase ALL its descendants.

  Now we use fork-point rebasing for each changed bookmark, equivalent to  `jj rebase --branch bookmark --destination
  bookmark@remote`
  This only rebases commits that actually diverged from the remote, not all descendants.
  
  Do you have any comments for it?
  
  I'll refactor and rebase onto main after you approve/disapprove design changes

cli/src/commands/git/sync.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        writeln!(
+            ui.status(),
+            "Rebasing bookmark '{}' onto {} (from {})",
+            symbol.name.as_str(),
+            &new_head_id.hex()[..12],
+            symbol
+        )?;
+
+        let root_commit_ids = crate::commands::rebase::find_branch_fork_point_roots(
```
PhilipMetzger: nit: import the symbol so its only `find_branch_fork_points_roots(...)`

cli/src/commands/git/sync.rs
```diff
@@ -0,0 +1,339 @@
...
+        );
+
+        let move_location = MoveCommitsLocation {
+            new_parent_ids: vec![new_head_id.clone()],
+            new_child_ids: vec![],
+            target: MoveCommitsTarget::Roots(root_commit_ids),
+        };
+
+        let rebase_options = RebaseOptions {
+            empty: jj_lib::rewrite::EmptyBehaviour::AbandonAllEmpty,
```
PhilipMetzger: nit: Ditto with importing the name

cli/tests/test_git_sync.rs
PhilipMetzger, line 1: nit: copyright only should be 2025 since it is new, the same counts for `cli/src/commands/git/sync.rs` and for future readers it'd be beneficial if all of these tests contained some `jj log` outputs similar to `jj parallelize`.

cli/src/commands/git/sync.rs
```diff
@@ -0,0 +1,339 @@
...
+        };
+
+        let computed_move = compute_move_commits(tx.repo(), &move_location)?;
+        let stats = computed_move.apply(tx.repo_mut(), &rebase_options)?;
+
+        let rebased_count = stats.num_rebased_targets + stats.num_rebased_descendants;
+        if rebased_count > 0 || stats.num_abandoned_empty > 0 {
+            writeln!(
+                ui.status(),
+                "  Rebased {} commits for bookmark '{}' ({} abandoned as already merged)",
```
PhilipMetzger: nit: I prefer to have the message like `({} abandoned as they're already in trunk())` since that avoids giving a specific name to the thing called merging/landing/submitted in different VCS.

PhilipMetzger (commented):
SGTM but Yuya will do the final reviews

yuja:
> Now we use fork-point rebasing for each changed bookmark, equivalent to `jj rebase --branch bookmark --destination bookmark@remote`

I haven't reviewed the implementation thoroughly, but does it rebase anonymous branches?
```
                      B'
B                     |
|               =>    A' foo@origin
A foo@origin       A
```

The condition to rebase might be something like `old_base..head & ::untracked_remote_bookmarks() == none()`.

BTW, I expect the core logic and tests will be moved to jj-lib.

zx8:
> @0xdeafbeef I'll refactor and rebase onto main after you approve/disapprove design changes

Is there anything you need from the `jj` maintainers to get this merged, or is it simply a matter of finding the time to work on it? 🙏

0xdeafbeef:
> > @0xdeafbeef I'll refactor and rebase onto main after you approve/disapprove design changes
> 
> Is there anything you need from the `jj` maintainers to get this merged, or is it simply a matter of finding the time to work on it? 🙏

problem is the lack of time and that it's not so trivial as i thought :) I'll rebase on main today and try to resolve all comments

cli/src/commands/git/sync.rs
```diff
@@ -0,0 +1,339 @@
...
+    ///
+    /// By default, the specified remote names match exactly. Use a [string
+    /// pattern], e.g. `--remote 'glob:*'`, to select remotes using
+    /// patterns.
+    ///
+    /// [string pattern]:
+    ///     https://jj-vcs.github.io/jj/latest/revsets#string-patterns
+    #[arg(
+        long = "remote",
+        short = 'r',
```
matts1: I don't think we should use `-r`. In every other context in jj, `-r` refers to revsets, and in this case, I could see a user thinking that they could type `jj sync -r <revision>` to sync a particular revision / revset to head (which could be something we'd choose to add support for later).

cli/src/commands/git/sync.rs
matts1: I don't think `trunk()` is technically the right terminology. `trunk()` refers to specifically the main branch, but some of these commits may have merged into feature branches.

We'd need to say instead that they were in the bookmark name.

zx8:
I think it's clear at this point that @0xdeafbeef doesn't have the time to finish this PR up. This isn't a complaint, I understand how life works, just an observation. Especially considering that I don't have the skillset to help carry it over the line myself!

That said, it's been open for a while now and it's one of the few remaining gaps that gives me pause before recommending `jj` to people. Is there any way I or others could help move it forward, whether that's testing, picking up specific pieces, or someone else stepping in to finish it up?

0xdeafbeef:
> I think it's clear at this point that @0xdeafbeef doesn't have the time to finish this PR up. This isn't a complaint, I understand how life works, just an observation. Especially considering that I don't have the skillset to help carry it over the line myself!
> 
> That said, it's been open for a while now and it's one of the few remaining gaps that gives me pause before recommending `jj` to people. Is there any way I or others could help move it forward, whether that's testing, picking up specific pieces, or someone else stepping in to finish it up?

I've used jj sync for a while and found it kinda useless :( For big repos sync takes really LONG time. Eg for linux it will try to rebase every tag/branch onto upstream and it's wasteful. And also I'm not sure that the rebase of all braches is desired and expected behavior. Personally i just do `jj git fetch && jj rebase --into branch@remote`. I have time now and i can recheck the status(I guess current should be reimpled from scratch). But code itself is much simpler than ergonomics side. Mb we should discuss it in Discord(or here) 1 more time. And yeah, sorry for going radio silent :)

martinvonz:
IMO, `jj sync` should default to rebasing `reachable(@, trunk()..)` and have a `--all` switch to rebase all revisions.

matts1:
> For big repos sync takes really LONG time

If you're rebasing tags, that sounds like you've either got it rebasing immutable commits on sync, which it definitely shouldn't do, or for some reason your tags are mutable (which seems like misconfiguration to me). I have jj configured to treat only my local branches as mutable, so thus rebasing "everything" is just rebasing a small set of mutable commits.

The way I've been syncing is to use `jj git fetch && jj rebase -s 'roots(mutable())' -d 'trunk()'` (or equivalently, `-b 'mutable()'). I've been doing this on chromium, which is also a massive repository, and the rebase is an extremely quick operation.

See https://source.chromium.org/chromium/chromium/src/+/main:tools/jj/sync.py;drc=dbb5f40c731c7171b295d62b96463d56c6e6bbc1

0xdeafbeef:
@zx8 finally i've found the time. Rewrote sync from scratch, for now only lib part is done.

It's much much trickier than it seams. Will write a bigger comment a bit latter :)

UPD:

It should cover normal use cases and fail fast on strange graph shapes. Played with Z3 (thanks, we have LLMs) to check that the graph isn't broken after the rebase. Found a few bugs using it, and this version is much more robust than the previous one.

What it does and does not do:

We select all commits from the revsets describing the local branch. Then we find the upstream point from which each child branch started and rebase it onto the new upstream position.

We skip cases where:
- upstream was deleted, conflicted or force-pushed;
- different upstream updates would move the same branch in different directions;
- there are unselected commits in the middle of the branch;
- several updates overlap in a way that could lose commits or graph edges.

Some tricky things that should work:
- anonymous branches without bookmarks;
- branches containing merges;
- several independent upstream branches moving at once;
- bookmarks and working copies following the rewritten commits;
- unrelated branches staying untouched.

For now only the lib part and tests are done. CLI will come later. And it should be pretty it easy to add.

I'll make it draft for now, need a bit more time  to sort everything out  before asking for a review

0xdeafbeef:
@yuja @martinvonz

i've found another design issue (i've lost count of them) :sob: :sob: :sob: :sob: :sob:

Right now the planner consumes the output of git import after fetch. From that output, it knows that a remote bookmark moved from A to B during the fetch.

This makes it possible to find local branches based on A and rebase them onto B. It also works for anonymous branches, which is pretty important for a normal jj workflow.

But it makes the CLI a bit weird:

```sh
jj git sync
```

can fetch and rebase the branch, while:

```sh
jj git fetch
jj git sync
```

makes the second command a no-op, because the A -> B movement was only available during the previous import.

I can see three possible options:

1. Keep the current design and accept that sync only works together with its own fetch. A separate fetch does not preserve the information needed by a later sync.

2. Try to infer the upstream only from the current commit graph. I don't think this is safe, and I tried it in the first iteration:

```text
              C  (@, anonymous branch)
             /
    A --- P --- R --- D  (release@origin)
```

The graph says that rebasing C onto D is valid:

```text
    A --- P --- R --- D --- C'
```

But it does not say whether C was actually created from `release@origin` or is an independent branch that should stay on P. Both interpretations have exactly the same graph, so ancestry alone is not enough.

3. Persist a small, non-consumable record of imported movement for each tracked remote bookmark.

For example, before fetch there may be two anonymous branches based on A:

```text
A (main@origin) ──┬──> C (@)
                  └──> E (another anonymous head)
```

Fetch imports A -> B:

```text
A ──┬──> B (main@origin)
    ├──> C (@)
    └──> E (another anonymous head)

stored sync frontier: main@origin: A -> B
```

A later, separate sync can rebase only the current branch:

```text
A ──┬──> B (main@origin) ──> C' (@)
    └──> E (another anonymous head)

stored sync frontier: main@origin: A -> B
```

The frontier is not consumed, so a later `--all` can still rebase E:

```text
A ──> B (main@origin) ──┬──> C' (@)
                        └──> E'

stored sync frontier: main@origin: A -> B
```

Sync would only consider stored movements matching the remote and bookmark scope selected by the command.

Commits already based on the current target will naturally be skipped by the next sync.

There are some annoying details here. If fetch deletes or force-pushes the bookmark, a later sync still needs to know about it and fail. But push, track/untrack, or a manual remote bookmark change should probably forget the saved movement, since it no longer describes a fetch.

This does not require assigning an upstream to every anonymous branch. The revset still selects the commits, the persisted frontier only tells the planner which remote bookmarks moved, and the existing graph checks decide what can be safely rebased.

tl;dr: `View` already stores the current remote target B. I think it also needs to remember that the imported movement started at A, or the first unsafe update if the bookmark was deleted or force-pushed. Sync uses that state but does not consume it.

This needs some merge/serialization semantics, but it seems less surprising than making sync depend on whether fetch happened in the same command.

Does this sound reasonable, or is there some existing state/API that would make this simpler?

It’s the deepest rabbit hole I’ve ever mistaken for an easy weekend pr :)

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+    let selected = SelectedGraph::load(mut_repo, selected_commit_ids).await?;
+    let moves = derive_boundary_moves(mut_repo, &selected, import_stats).await?;
+    let plan = GitSyncPlanBuilder::new(mut_repo, selected, moves)
+        .build()
+        .await?;
+    plan.apply(mut_repo, rebase_options).await
+}
+
+/// Why an imported remote bookmark update cannot be synchronized.
+#[expect(missing_docs)]
```
PhilipMetzger: nit: write the documentation for new modules. You also had a previous version which already dealt with this, with a minor nit that the doc comments should be above all `#[derive(...)]` or `#[error(...)` annotations.

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+    #[error("the imported target is absent")]
+    NewTargetAbsent,
+    #[error("the imported target {target:?} is conflicted")]
+    NewTargetConflicted { target: RefTarget },
+    #[error("moving from {old} to the imported target {new} is not a fast-forward")]
+    NonFastForward { old: CommitId, new: CommitId },
+}
+
+/// Errors from synchronizing selected commits with imported Git updates.
+#[expect(missing_docs)]
```
PhilipMetzger: ditto

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+impl<'repo> GitSyncPlanBuilder<'repo> {
+    fn new(repo: &'repo MutableRepo, selected: SelectedGraph, moves: Vec<BoundaryMove>) -> Self {
+        let boundaries = moves
+            .into_iter()
+            .map(|boundary| BoundarySelection {
+                boundary,
+                affected_commits: IndexSet::new(),
+                rewrite_roots: IndexSet::new(),
+            })
+            .collect();
```
PhilipMetzger: nit: use `collect_vec` to adhere to our style guide.

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+    fn validate_selected_coverage(&self) -> Result<(), GitSyncError> {
+        // Check each move separately. Otherwise another move can make a path with an
+        // unselected intermediate commit appear covered.
+        for selection in &self.boundaries {
+            let reachable = self.selected.descendant_closure(&selection.rewrite_roots);
+            let uncovered: Vec<_> = selection
+                .affected_commits
+                .difference(&reachable)
+                .cloned()
+                .collect();
```
PhilipMetzger: ditto with `collect_vec()`

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+            .collect();
+        let planned_ids = self.selected.descendant_closure(&rewrite_root_ids);
+        // A destination cannot be rewritten because another move may use it as a
+        // replacement parent.
+        let mut moving_destinations: Vec<_> = self
+            .boundaries
+            .into_iter()
+            .map(|selection| selection.boundary.new)
+            .filter(|id| planned_ids.contains(id))
+            .collect();
```
PhilipMetzger: ditto with `collect_vec()`

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+        selected_ids.sort_unstable();
+        selected_ids.dedup();
+
+        let store = repo.store();
+        let commits = try_join_all(selected_ids.iter().map(|id| store.get_commit_async(id)))
+            .await?
+            .into_iter()
+            .map(|commit| (commit.id().clone(), commit))
+            .collect::<IndexMap<_, _>>();
+        let mut children_by_parent = IndexMap::<CommitId, Vec<CommitId>>::new();
```
PhilipMetzger, line 542: optional: I'm not sure but I think the type can be inferred here so it just can be `IndexMap::new`

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+        .chunk_by(|update| update.symbol.name.clone())
+    {
+        let tracked_updates: Vec<_> = updates
+            .filter(|update| {
+                update.old_remote_ref.is_tracked()
+                    || repo
+                        .get_remote_bookmark(update.symbol.as_ref())
+                        .is_tracked()
+            })
+            .collect();
```
PhilipMetzger: ditto with `collect_vec` and dropping the annotation.

martinvonz:
> Keep the current design and accept that sync only works together with its own fetch. A separate fetch does not preserve the information needed by a later sync.

This sounds good to me.

lib/src/repo.rs
```diff
@@ -1442,6 +1442,83 @@ impl MutableRepo {
...
+    /// Panics if the mutable repository already contains rewrite mappings,
+    /// `commits` contains duplicate commit IDs, or the callback leaves mappings
+    /// for a different set of commits.
+    ///
+    /// # Transaction safety
+    ///
+    /// The caller must discard the transaction if this method returns an error
+    /// or if its future is cancelled before completion. On failure, the
+    /// repository may contain partial in-memory mutations.
+    #[cfg(feature = "git")]
```
PhilipMetzger: nit: I don't think this needs to be Git only

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+    rewrite_roots: IndexSet<CommitId>,
+}
+
+struct GitSyncPlanBuilder<'repo> {
+    repo: &'repo MutableRepo,
+    selected: SelectedGraph,
+    boundaries: Vec<BoundarySelection>,
+    parent_replacements: IndexMap<CommitId, IndexMap<CommitId, CommitId>>,
+}
+
```
PhilipMetzger, line 334: nit: I guess these types should have some internal documentation

lib/src/git_sync.rs
```diff
@@ -0,0 +1,628 @@
...
+    async fn build(mut self) -> Result<GitSyncPlan, GitSyncError> {
+        self.classify_affected_commits_and_rewrite_roots().await?;
+        self.validate_selected_coverage()?;
+        self.validate_non_overlapping_rewrite_roles()?;
+        self.finish()
+    }
+
+    async fn classify_affected_commits_and_rewrite_roots(&mut self) -> Result<(), GitSyncError> {
+        let index = self.repo.index();
```
PhilipMetzger, line 368: nit: I'd be nice if the algorithm here had a explanation.

PhilipMetzger (commented):
minor comments. Also will you add the final implementation commit for `jj git sync` here as a parent or will that be a separate PR?

0xdeafbeef:
> minor comments. Also will you add the final implementation commit for `jj git sync` here as a parent or will that be a separate PR?

I've already started implementing the cli as a second commit in this PR, and that's how I found the fetch + sync issue described above. I don't think it makes sense to split them into separate PRs, cause cli shapes lib after all

lib/src/git_sync.rs
0xdeafbeef: I thought the documenting error variants which have literal text is a bit dumb, but added docs

PR #6828, opened 2025-06-25
FAQ: Answer on how to revert merge commits

PhilipMetzger:
This is idiosyncratic to new users coming from Git, since `jj revert <merge>` does not  usually do what users want. Maybe a followup could be adding a warning in `jj revert`.

The explanation is from a Discord message from Austin.

# Checklist

If applicable:

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -515,6 +515,27 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
 indicate which commits are empty. It's used in the `files()` revset function
 (and by `jj log <path>`) to find commits that modify a certain path. And so on.
 
+## How do I revert a merge commit? `jj revert <merge>` does nothing
+
+Instead use `jj restore --from <first parent>` to revert the changes merged
+from the second parent.
+
+Example:
+```text
+@
+|
+C
+| \
+B D
+|/
+A
+```
+To revert the merge in `C`, create a new commit with `jj new` and
+then `jj restore --from B` and then describe the message
+with something like `jj desc -m "Revert the merge of B into D`. Now, commit `@`
+undoes the merge of `D` into  C`. If necessary, you can now rebase it
+elsewhere, e.g. `jj rebase -r @ -d main`.
+
 ### How do I deal with divergent changes ('??' after the [change ID])?
 
 A [divergent change][glossary_divergent_change] represents a change that has two


docs/FAQ.md
```diff
@@ -515,6 +515,29 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
 indicate which commits are empty. It's used in the `files()` revset function
 (and by `jj log <path>`) to find commits that modify a certain path. And so on.
 
+## How do I revert a merge commit? `jj revert <merge>` does nothing
+
+Since a merge commit is typically empty, because assuming two branches merge
+without conflict, there is no resolution that needs to be recorded in the
+merge commit itself, so its "empty". So `jj revert M` will not undo anything,
+because the merge itself was (presumpably) empty. Instead use
```
joyously: Can you condense this repetition, so it makes more sense?
_(and this `its` should be `it's`)_

docs/FAQ.md
```diff
@@ -515,6 +515,29 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
...
+@
+|
+C
+| \
+B D
+|/
+A
+```
+To revert the merge in `C`, `jj restore --from B` and then describe the message
+with something like `jj desc -m "Revert merge <Cs commit-id>`.
```
joyously: Either show what the graph would be after the command, or explain why the command reverts the merge.
Perhaps explain what `revert` actually means, if it's different from Git.

docs/FAQ.md
PhilipMetzger: I removed the section, since it's explained in the answer above

docs/FAQ.md
PhilipMetzger: I added some more words on what actually happens

docs/FAQ.md
```diff
@@ -515,6 +515,28 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
 indicate which commits are empty. It's used in the `files()` revset function
 (and by `jj log <path>`) to find commits that modify a certain path. And so on.
 
+## How do I revert a merge commit? `jj revert <merge>` does nothing
+
+Instead use `jj restore --from <first parent>` and add the respective message with
+`jj desc` or `jj commit`, since removing the parent edge to the merged commits
+removes the added trees.
```
yuja: Maybe better to say "To revert changes merged from the second parent," or something?

> removing the parent edge to the merged commits removes the added trees

This part was difficult for me to understand, maybe because `revert` doesn't actually remove the parent edges.

PhilipMetzger: rewritten as that

docs/FAQ.md
```diff
@@ -515,6 +515,27 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
...
+C
+| \
+B D
+|/
+A
+```
+To revert the merge in `C`, `jj restore --from B` and then describe the message
+with something like `jj desc -m "Revert merge <C's commit-id>`. As this removes
+the parents edge to `D`, so the commits tree is reset to a state without its
+changes.
```
yuja: Same here. "removes the parents edge to `D`" wasn't easy to understand.

docs/FAQ.md
```diff
@@ -515,6 +515,27 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
...
+
+
```
yuja, line 538: nit: remove extra blank line

yuja (approved):

docs/FAQ.md
```diff
@@ -515,6 +515,27 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
...
+@
+|
+C
+| \
+B D
+|/
+A
+```
+To revert the merge in `C`, `jj restore --from B` and then describe the message
+with something like `jj desc -m "Revert merge <C's commit-id>`. As this removes
```
ilyagr: I'd add "with" after "Revert merge". **Update:** I was actually thinking "Revert merge with <D's commit-id>".

docs/FAQ.md
```diff
@@ -515,6 +515,27 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
 indicate which commits are empty. It's used in the `files()` revset function
 (and by `jj log <path>`) to find commits that modify a certain path. And so on.
 
+## How do I revert a merge commit? `jj revert <merge>` does nothing
+
+Instead use `jj restore --from <first parent>` to revert the changes merged
```
ilyagr, line 520: I think you should add `jj new C` or something before `jj restore`.

docs/FAQ.md
ilyagr: It's a little arguable whether "Revert merge with D" is clear, but I think "Revert merge C" is definitely *not* clear. So, if we don't have a good idea for a description here that makes things clearer, we could just delete the suggestion "... then describe the message with something like ...".

docs/FAQ.md
```diff
@@ -515,6 +515,27 @@ by `jj rebase` to rebase the changes in a commit. It's used in `jj log` to
...
+```text
+@
+|
+C
+| \
+B D
+|/
+A
+```
+To revert the merge in `C`, `jj restore --from B` and then describe the message
```
ilyagr: I think you should add `jj new C` before `jj restore`.

I previously wrote this above in an about to be deleted comment. I'm more sure about needing it here than above.

docs/FAQ.md
ilyagr: A possible replacement:

> Now, commit `@` undoes the merge of `D` into `B`. If necessary, you can now rebase it elsewhere, e.g. `jj rebase -r @ -d main`.

docs/FAQ.md
ilyagr: Perhaps "Revert the merge of B into D"?

docs/FAQ.md
PhilipMetzger: used Ilya's suggestion

docs/FAQ.md
PhilipMetzger: removed

docs/FAQ.md
PhilipMetzger: I took your last suggestion.

docs/FAQ.md
PhilipMetzger: done

PhilipMetzger:
I'm going to wait for Ilya's approval before landing it.

Issue #3262, opened 2024-03-08
Bikeshedding how we use/offer templates


khionu:
Templating (like scripting) is quite a deep topic, and how it's relevant to this project has had a fair amount of conversation to date, and will likely continue to need more for a while to go. This issue is meant to serve as a place to aggregate and hash out desires, concerns, and anything else that might inform the total story of templating in Jujutsu.

A gentle reminder that push back does not mean that desires/needs are invalid, we're all here to try and make Jujutsu as good of a tool as it can be.

It would be appreciated if collaborators could help keep the below sections updated

## Stance summaries

-

## Comments of note

-

khionu:
My thoughts, in no particular order, and assuming no particular prior knowledge of the reader:
- The flexibility that templates can provide is often very high. Sometimes that flexibility is needed, sometimes it isn't. Too much can be painful, to both maintainers of jj and the templates, as can too little can be stifling.
- More popular Helm charts can be virtually untouchable by anyone who doesn't work on the charts themselves regularly
- It might be possible to create loose measures of template potential to estimate how much flexibility we're comfortable with. May be worth a working group of some kind?
- I'm thinking that some good bounds may be:
  - No user functions
  - No variables
  - No arithmetic
  - No ability to access or measure time beyond what's already in eg commits
  - Can overwrite built-in templates
  - Values for a given template can be `T | List<T>`, but builtin functions can return nested maps/objects as well. T being one concrete type for the context (eg string or rev)
  - User expressions can return new values, but can't mutate existing values
  - Focus on user expressions being filters, joins, queries. Additive or selective operations.
  - Filesystem access restricted to repo directory, -x -w, no following links
    - If there are some files we want them to be able to save to, maybe a persistent kv that they access with functions, then any modules can work form the same store.

PhilipMetzger:
Summarizing my thougths on the templating system: 

 * It has been clear for a while that our users want to use it, wherever possible
 * A minimal subset of the language should be available for configuration
 * We definitely need to avoid the language in a language problem
 * It shouldn't be a bigger burden than the current log templating language

dairyisscary:
From #3219:

I'd love to see some common encoding functions built into the templating language to facilitate integrations with editors, shell prompts, and really anything that would to prefer to parse structured data, rather than human-readable plain text.

Something like:
```console
jj log --template 'json({ id: commit_id, author: author, desc: description })'
jj log --template 'csv([commit_id, author, description ])'
[...]

Issue #3219, opened 2024-03-05
FR: Add `jj api` command to establish structured interprocess communication for seamless IDE integration


ya-goodfella:
**Is your feature request related to a problem? Please describe.**

There's no integrations available for any mainstream IDE on the market, and also that described as one of the most drawbacks by article authors in [media](https://v5.chriskrycho.com/essays/jj-init/#is-it-ready), ultimately damaging adoption rate.

**Describe the solution you'd like**

I propose adding `jj api` command which provides a way for IDE and library to communicate through sending structured data. Having written API specification `jj api` might start serving (gRPC?) requests for it over `stdin`/`stdout` with the calling process, including sending events back to IDE itself. Another benefit of doing it that way is that we can have a single requirement for IDE extension authors - having `jj` in the `PATH`, so no need to package native libraries or dynamically linking with the unknown version of library that need to be installed somehow.

**Describe alternatives you've considered**

Surfacing of integration with CLI itself involves painful construction of arguments and parsing outputs which can change from version to version. Integration with the native library is even harder because most of Electron-based IDEs (like VSCode) cannot use native node modules in extensions (see [microsoft/vscode#658](https://github.com/microsoft/vscode/issues/658)).

Using native library in the third-party is harsh because then it would work with storage directly and it would make changes to storage format much harder because of constant commitment for backward-compatibility in the heart of library.

**Additional context**

joyously:
See issue #3034

khionu:
Using jj as a library would mean bindings as a blocker to creating integrations. I think that that should definitely be an option, but being able to simply using the CLI would be more accessible.

An alternative would be to print the output of each command as a serialized format, eg JSON. It would likely be the same internal overhead as a dedicated command, but by the difference being a single flag, it would be much more straight forward for implementers to get precisely the behaviour they want. If there's information that integrations need/want that wouldn't be natural to include in any command's `--json` output, it could be a good point to figure out if we should be including that as part of the human-readable output.

dairyisscary:
> An alternative would be to print the output of each command as a serialized format, eg JSON. It would likely be the same internal overhead as a dedicated command, but by the difference being a single flag, it would be much more straight forward for implementers to get precisely the behaviour they want. If there's information that integrations need/want that wouldn't be natural to include in any command's `--json` output, it could be a good point to figure out if we should be including that as part of the human-readable output.

[...]

PR #6838, opened 2025-06-27
cli: add built-in `ndjson()` template alias

thoughtpolice:
This makes the most common case much more intuitive and easy to handle without requiring `self` to be passed, because the overwhelming majority of use cases are formulated as piping newline-delimited JSON objects to other tools.

# Checklist

If applicable:

### DIFF
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -451,3 +451,6 @@ git_format_patch_email_headers = '''
 # Gerrit change id is 40 chars, jj change id is 32, so we need some padding.
 # `6a6a6964` is the hexadecimal of `jjid` in ASCII, just to not pad with zeros.
 "format_gerrit_change_id_trailer(commit)" = '"Change-Id: I6a6a6964" ++ commit.change_id().normal_hex() ++ "\n"'
+
+"ndjson(obj)" = 'json(obj) ++ "\n"'
+"ndjson()" = 'ndjson(self)'


cli/src/config/templates.toml
```diff
@@ -451,3 +451,6 @@ git_format_patch_email_headers = '''
 # Gerrit change id is 40 chars, jj change id is 32, so we need some padding.
 # `6a6a6964` is the hexadecimal of `jjid` in ASCII, just to not pad with zeros.
 "format_gerrit_change_id_trailer(commit)" = '"Change-Id: I6a6a6964" ++ commit.change_id().normal_hex() ++ "\n"'
+
+"ndjson(obj)" = 'json(obj) ++ "\n"'
+"ndjson()" = 'ndjson(self)'
```
ilyagr, line 456: I'm not sure about the name. Perhaps `jsonline`, since it's a single line? (Possibly with a dash or underscore to match other templates)

I also wondered about `jsonln` by analogy with `println` or `jsonlf`, but I realized this might not always be `json()` followed by a newline if `json()` gains an option for outputting human-facing json with more than one line per object.

ilyagr:
I'm also wondering how to document it, and whether it should be a function or not.

For example, it could be a built-in template, so that you could do `jj log -T builtin_json_line`.

yuja:
Yes, the 0-ary version could be called `builtin_<something>`. I personally wouldn't add alias for `json(x) ++ "\n"` because it wouldn't provide much convenience.

cli/src/config/templates.toml
thoughtpolice: `ndjson` is already a well-known name and format; I don't think deviating from the name is really that meaningful, but `jsonline` wouldn't be bad I guess.

thoughtpolice:
> Yes, the 0-ary version could be called `builtin_<something>`. I personally wouldn't add alias for `json(x) ++ "\n"` because it wouldn't provide much convenience.

I disagree; the short syntax, both with the `self` and `++ "\n"` shortcut, it makes it *vastly* more convenient to use in ~all situations, which is especially useful when piping things around the terminal in an ad hoc way. People also might want to actually print out subsets of the object or combine it with other function templates e.g. `ndjson(f(self))`, which is simply much easier to handle when it's a function.

(In fact, I already find it annoying to write `commit_id ++ "\n"` in templates sometimes mostly because I always default to double-quoting things! I might even actually like a template function `nl(commit_id)` that does it for me.)

But even if we gave it an actual non-function name though, the other thing is that I strongly believe it's much better this way because it *does what you mean*. It is more correct. If I write `json(x)`, there is pretty much no case in which I want the output all on a single line if there are multiple entries. Tools like `jq` can handle this cleanly, but there's no guarantee most tools will be that robust; the amount of code that does something like `xs = input.split('\n'); jsonLines = map(xs, parseJsonObject)` or something like that is vast, and I've written plenty of it. It's a very obvious solution to the problem of "parse ndjson" and the one 99% of people will use.

Just as an example, `gron` does not handle this as expected, while `jq` does:

```
austin@GANON:~/src/jj$ jj
◉  zkwuymqm <aseipp@pobox.com> [8 hours ago] 75ceef9a
◆  ysoormqp <yuya@tcha.org> [10 hours ago] main* 981d0b7d git_head()
│  revset: inline expect_<construct>_with() parsing helpers
~
austin@GANON:~/src/jj$ jj --no-graph --ignore-working-copy -T 'json(self)' | jq
{
  "commit_id": "75ceef9aab4f3abac214b588c5491b98f5960e68",
  "parents": [
    "981d0b7de4a3593a48288f36b83ceddb6386477d"
  ],
  "change_id": "zkwuymqmpquwlwpqylwltkxourwwzxtx",
  "description": "",
  "author": {
    "name": "Austin Seipp",
    "email": "aseipp@pobox.com",
    "timestamp": "2025-06-26T22:39:40-05:00"
  },
  "committer": {
    "name": "Austin Seipp",
    "email": "aseipp@pobox.com",
    "timestamp": "2025-06-26T22:39:40-05:00"
  }
}
{
  "commit_id": "981d0b7de4a3593a48288f36b83ceddb6386477d",
  "parents": [
    "1f91541355e970413df11aa247c6135d177a8e08"
  ],
  "change_id": "ysoormqpwzmvyolvtpyyystpzonlopvp",
  "description": "revset: inline expect_<construct>_with() parsing helpers\n\nThese functions existed because we had to deal with AliasExpanded nodes, which I\nthink is better handled by catch_aliases() + expect_<construct>() without\ncallback.\n\nSome redundant error nodes are flattened.\n",
  "author": {
    "name": "Yuya Nishihara",
    "email": "yuya@tcha.org",
    "timestamp": "2025-06-25T18:17:55+09:00"
  },
  "committer": {
    "name": "Yuya Nishihara",
    "email": "yuya@tcha.org",
    "timestamp": "2025-06-27T01:29:11Z"
  }
}
```

vs

```
austin@GANON:~/src/jj$ jj --no-graph --ignore-working-copy -T 'json(self)' | gron -s
json = {};
json.author = {};
json.author.email = "aseipp@pobox.com";
json.author.name = "Austin Seipp";
json.author.timestamp = "2025-06-26T22:39:40-05:00";
json.change_id = "zkwuymqmpquwlwpqylwltkxourwwzxtx";
json.commit_id = "75ceef9aab4f3abac214b588c5491b98f5960e68";
json.committer = {};
json.committer.email = "aseipp@pobox.com";
json.committer.name = "Austin Seipp";
json.committer.timestamp = "2025-06-26T22:39:40-05:00";
json.description = "";
json.parents = [];
json.parents[0] = "981d0b7de4a3593a48288f36b83ceddb6386477d";
```

If anything, I'd rather we rename the `json()` function to `builtin_json` and `ndjson` to `json`.

I do think this should be documented, but the reason I didn't do that yet is because I wanted us to hash out these sorts of concerns. :)

yuja:
Short aliases can be added by user, right? I don't think we should add many builtin short aliases in general because they might conflict with user's definitions.

And the newline issue isn't specific to `json()`, so I wouldn't make `json()` do `++ "\n"` automatically, which is inconsistent. It's annoying that `jj config list -Tname` doesn't insert `"\n"` for example, but the current behavior makes the most sense. Maybe we could add some syntax or config knob to insert `"\n"` automatically (e.g. `-Tnl:<expr>`), but I don't have a good idea.

thoughtpolice:
I believe I agree we don't want to add too many short aliases, but I think that has to be weighed on a case-by-case basis in most instances. There is a balance to be struck. Plus, if this definition is *not* useful for a user, or conflicts with what they have, I would be interested in learning why; we might be able to generalize these things appropriately for everyone's benefit. I also somewhat doubt there is anyone out there with this particular function in their `config.toml` that's going to do it better or much differently than what you've implemented, but who knows...

I feel similarly about the `++ "\n"` thing. Yes, almost ~all other templates don't do it, because it obviously doesn't make any sense for most of them. But I don't think it should therefore be taken as a *fait accompli* that we can never do this. Indeed, it's baked in the name `ndjson()` that there will be newlines! And on reflection I think `ndjson()` really is the right name because of that (so I retract my previous thing about `builtin_json()`)

More broadly I think of it as a "soft" user expectation thing. Let me put it this way: why would I ever *not* want newlines? Like, there is no situation in which a newline is *bad*, and almost a million situations in which it's *useful*. So why am I forced to add it? Why am I having to write out an obvious thing on my own? Similarly, it was originally talked about in Discord that this `ndjson()` could have been a template like my original PR, that we stuck on a wiki somewhere and pointed people at, and then it's "not our problem." But I find that rather user unfriendly. Why make someone click a link to go copy and paste a thing so they can then script a thing, when we can just... do that for them? It is just that extra 1% of polish that makes life nice, and those things often are cases where "exceptions make the rule". But they also make users happy, often with only a little cost.

PhilipMetzger:
> More broadly I think of it as a "soft" user expectation thing. Let me put it this way: why would I ever _not_ want newlines? Like, there is no situation in which a newline is _bad_, and almost a million situations in which it's _useful_. So why am I forced to add it? Why am I having to write out an obvious thing on my own? Similarly, it was originally talked about in Discord that this `ndjson()` could have been a template like my original PR, that we stuck on a wiki somewhere and pointed people at, and then it's "not our problem." But I find that rather user unfriendly. Why make someone click a link to go copy and paste a thing so they can then script a thing, when we can just... do that for them? It is just that extra 1% of polish that makes life nice, and those things often are cases where "exceptions make the rule". But they also make users happy, often with only a little cost.

I don't think that's worth the exception in the longterm, as we _the CLI_ only should provide _building blocks_ and really shouldn't make decisions on the users behalf, even if it hits ourselves sometimes. Since these expectations will become the norm which is definitely wrong, it also prevents building a natural community-built ecosystem for templates around Jujutsu. See all the discussion around #3219. 

I also disagree that this is sorta frozen like `.method()` syntax since that actually would've broke the world if it wasn't done in a backwards compatible way.

yuja:
> I feel similarly about the `++ "\n"` thing. Yes, almost ~all other templates don't do it, because it obviously doesn't make any sense for most of them.

It does make sense if the template expression is used stand-alone, and if there are multiple outputs. That's the same for `json()`, I believe. If I write `-Tjson(foo)` or `-Tfoo`, I would want the output separated by newline or something. If I write `-T '{"foo":' ++ json(foo) ++ ..'`, I don't want newlines in between. If I write `-T '... ++ "\0"'`, I don't want newlines at all.

emilazy:
> I don't think that's worth the exception in the longterm, as we _the CLI_ only should provide _building blocks_ and really shouldn't make decisions on the users behalf, even if it hits ourselves sometimes.

This essentially implies the `jj(1)` CLI is plumbing that needs someone else to create porcelain on top of, right? I don’t think most things about the current UX fit with that model (and I suspect it would be bad for Jujutsu adoption; there’s a reason Git folded in the porcelain pretty quickly, and CLIs don’t make good plumbing to begin with).

PhilipMetzger:
> > I don't think that's worth the exception in the longterm, as we _the CLI_ only should provide _building blocks_ and really shouldn't make decisions on the users behalf, even if it hits ourselves sometimes.
> 
> This essentially implies the `jj(1)` CLI is plumbing that needs someone else to create porcelain on top of, right?

Sorry I was imprecise there, the template language should provide the building blocks, the upstream shouldn't really care if the individual user runs `snapshot.track = "none()"` or what they do with their private templates. Also #3219 has plenty discussion around that problem already (there's also older discussion in #3262). 

It's your interpretation that this is a plumbing and porcelain problem (I disagree that this model is a good framing, see our previous discussions) and not something I consciously implied.

emilazy:
Alright; I think I just don’t fully understand the distinction being drawn, then. From my perspective, the template language is part of the CLI UX, especially default aliases added by the CLI.

thoughtpolice:
> > I feel similarly about the `++ "\n"` thing. Yes, almost ~all other templates don't do it, because it obviously doesn't make any sense for most of them.
> 
> It does make sense if the template expression is used stand-alone, and if there are multiple outputs. That's the same for `json()`, I believe. If I write `-Tjson(foo)` or `-Tfoo`, I would want the output separated by newline or something. If I write `-T '{"foo":' ++ json(foo) ++ ..'`, I don't want newlines in between. If I write `-T '... ++ "\0"'`, I don't want newlines at all.

I mean that by default, there's no reason to expect most of the template aliases (e.g. timestamp formatting) to have `++ "\n"`, so from that perspective it's an obvious outlier. But the case you mention is exactly handled by `json()` already, this doesn't change that; `ndjson()` is nothing more than a shorthand for something that's common that is a very simple and common case. If I called it `newline_delimited_json()` then I don't think it would surprise anyone that a newline is attached; and the solution if you don't want that is simple, just use `json()` instead. I feel that's fairly straightforward?

thoughtpolice:
> I don't think that's worth the exception in the longterm, as we _the CLI_ only should provide _building blocks_ and really shouldn't make decisions on the users behalf, even if it hits ourselves sometimes. Since these expectations will become the norm which is definitely wrong, it also prevents building a natural community-built ecosystem for templates around Jujutsu. 

I've said this before, but I just legitimately don't understand your point here at all, I'm sorry. There is no reason to believe adding simplistic JSON render functions is going to inhibit any "natural community" from forming, nor will it inhibit peoples usage of templates to their own ends, which they already do in spades. People already rely on templates extensively and any change to data models will need to account for that *regardless of the output*. For instance, Yuya's tests for making things Serializable will immediately tell us if things break.

I feel like this argument is something like "If we ever do this, even once, then we'll do it 1,000,000 times again and everything will be awful", but that is *not* a given! That is why I am saying we evaluate these things on a case-by-case basis. There is no reason to think this is any kind of slope or any actual inhibitor to users. If it was, how is the `json()` functionality in the first place not also guilty?

> See all the discussion around #3219.

OK, but nothing about this change precludes usage of `jj api` or its development at all. In fact, it is preferrable to use `json()` for many cases for things like scripts or prompts, for example. It is completely fine for there to be multiple ways to get information from jj.

cli/src/config/templates.toml
ilyagr: I am just one data point, but I haven't heard of `ndjson` before, and it took me a second to figure out. I would call this format `JSONL` or `jsonlines`. If you said "newline-delimented json", that'd be clear to me.

At the moment, I really like the name `builtin_json_line` (for singular), and `json_line`. For a template name, `builtin_json` would also make sense to me. Still, if you insist on `ndjson`, I'm OK with it, especially if it turns out that my unfamiliarity with it is unusual (but either way really).

ilyagr:
I didn't follow the whole discussion, but my first impression was also that the `json(x) ++ "\n"` alias would probably be less useful to me than the other one.

I think it might make sense to add the other one, see how it's used, and then come back to the question of whether we miss the `++ "\n"` alias for `json(x)`, and whether that should be solved specially or somehow addressed for other templates as well.

However, I don't feel very strongly about this, and I'm not sure I'd be ready to personally subscribe to any general policy here (though, again, I haven't followed everyone's arguments carefully at this point).

PhilipMetzger:
> > I don't think that's worth the exception in the longterm, as we _the CLI_ only should provide _building blocks_ and really shouldn't make decisions on the users behalf, even if it hits ourselves sometimes. Since these expectations will become the norm which is definitely wrong, it also prevents building a natural community-built ecosystem for templates around Jujutsu.
> 
> I've said this before, but I just legitimately don't understand your point here at all, I'm sorry. There is no reason to believe adding simplistic JSON render functions is going to inhibit any "natural community" from forming, nor will it inhibit peoples usage of templates to their own ends, which they already do in spades. People already rely on templates extensively and any change to data models will need to account for that _regardless of the output_. For instance, Yuya's tests for making things Serializable will immediately tell us if things break.
 
Until this point we've left out many things to the community and never chosen to "bless" a implementation or workflow, which is a thing you're currently pushing to _me_. Why should we provide such a functionality besides that its liked by the community and that a certain maintainer wants it . 
I also only _accepted_ the `json()` and never signaled any change in my opinion that it is a _good idea_.

> I feel like this argument is something like "If we ever do this, even once, then we'll do it 1,000,000 times again and everything will be awful", but that is _not_ a given! That is why I am saying we evaluate these things on a case-by-case basis. There is no reason to think this is any kind of slope or any actual inhibitor to users. If it was, how is the `json()` functionality in the first place not also guilty?

 To _me_ feature conservatism building on-top of a _good_ technical base (there's a reason why we have Design Docs in the project's processes) make the project currently as good and coherent as it is. It is a slippery slope though, since each similar feature request building on top of this _will be_ using this as a precedent, even if you say that you'll ensure that you'll scrutinize it equally to this.

Anyway, I still agree with Yuya's and Ilya's stances here, even if I'm just a lowly voice in the project.

yuja:
> I mean that by default, there's no reason to expect most of the template aliases (e.g. timestamp formatting) to have `++ "\n"`,

I see. My point was that the inconvenience caused by lack of `++ "\n"` is general problem, so ideally, it should be solved in a general consistent way. Since user can easily define `ndjson()` or whatever to work around the problem, I don't see much benefit to provide it by default.

PR #6839, opened 2025-06-27
trailers: make separators configurable

### LINKED ISSUE #6837: FR: Support for trailer separator customization
In git, I can customize my trailer separator. This is the character that delimits the key from the value. By default it is a colon, but git allows users to customize it in their config.

I would like jj to also allow customization of this separator.

To match git's functionality here seems like the most obvious solution, so I haven't considered alternatives.

**Additional context**
For me, I have a trailer in the format of key-value instead of key:value. I need to extract it for display in my one-line log messages as it's a jira task id.

glehmann:
fix #6837

# Checklist

If applicable:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -285,6 +285,9 @@ The recognizable trailer separators can be customized with the
 separators = ":%="
 ```
 
+The `format_trailer(key, value)` template uses the first `trailer.separators` character as
+separator as in [git](https://git-scm.com/docs/git-interpret-trailers#Documentation/git-interpret-trailers.txt-trailerseparators).
+
 ### Diff colors and styles
 
 In color-words and git diffs, word-level hunks are rendered with underline. You
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Trailers separators are now specified in the toml configuration.
 
+* The `commit_trailers` template now generates trailers without modification, and
+  their separators are defined in the TOML configuration.
+
 ### Fixed bugs
 
 * Broken symlink on Windows. [#6934](https://github.com/jj-vcs/jj/issues/6934).
--- a/cli/src/config/templates.toml
+++ b/cli/src/config/templates.toml
@@ -618,12 +618,20 @@ concat(
 )
 '''
 
+"format_trailer(key, value)" = '''
+  key
+  ++ config("trailer.separators").as_string().substr(0, 1)
+  ++ " "
+  ++ value
+  ++ if(!stringify(value).ends_with("\n"), "\n")
+'''
+
 "format_signed_off_by_trailer(commit)" = '''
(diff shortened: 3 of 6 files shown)


dolangish:
I'm a C++ dev and not a rust guy, but to my eyes this looks like it addresses my FR #6837 and is in line with how the git documentation describes their trailer separators. Thanks for the speedy turnaround on this!

cli/src/config-schema.json
```diff
@@ -792,6 +792,17 @@
                 }
             }
         },
+        "trailer": {
+            "type": "object",
+            "description": "Trailer configuration",
+            "properties": {
+                "separators": {
+                    "type": "string",
+                    "description": "A set of characters recognized as trailer separators",
```
ilyagr, line 854: Is it still true that the first of these is used by `jj` by default, like [git](https://git-scm.com/docs/git-interpret-trailers#Documentation/git-interpret-trailers.txt-trailerseparators)? If so, I'd mention it here.

docs/config.md
```diff
@@ -276,6 +276,14 @@ Some ready-to-use trailer templates are available for frequently used trailers:
 
 Existing trailers are also accessible via `commit.trailers()`.
 
+The recognizable trailer separators can be customized with the
+`trailer.separators` configuration. Its default value is `:`.
```
ilyagr, line 281: Same as above comment. You might also want to link to <https://git-scm.com/docs/git-interpret-trailers#Documentation/git-interpret-trailers.txt-trailerseparators>

ilyagr (commented):
Thanks for working on this! In terms of implementation, this looks good at a glance.

I'm not very familiar with trailers, but my first reaction is uncertaintly about whether customizing the trailer separator does more harm or good. https://github.com/jj-vcs/jj/issues/6837#issuecomment-3025620498

I'd be very happy if others looked at this as well.

jennings:
Some people in Discord [were surprised](https://discord.com/channels/968932220549103686/968932220549103689/1387695325770416199) that the commit trailer separator was configurable in Git, including myself.

That seems to me like the kind of thing that is better left unconfigurable for better compatibility. I wonder if it's a good idea to wait for more people to ask for this before making it an official feature?

The person in Discord asked for this because [their team had the habit of adding Jira ticket numbers to the end of their commit messages](https://discord.com/channels/968932220549103686/968932220549103689/1387822060553637899):

```
feat: Do something cool

A longer explanation of the feature.

CORE-1234
```

They were using `-` as the trailer separator so they could get the ticket number. But I think that could also be done in other ways.

cli/src/config-schema.json
glehmann: We copy the trailers as generated by the `commit_trailers` template configured by the user, so we don't use the first as the default value.
I suppose it could be done in `format_change_id_trailer()`, but for now `:` is used.

docs/config.md
glehmann: I've added the `format_trailer` template, which uses the first char of `trailer.separator` to format the trailer, updated the existing trailer templates to use it, and added that in the doc.

PR #6840, opened 2025-06-27
docs: show which template types support implicit conversions

yuja:
# Checklist

If applicable:

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -96,6 +96,8 @@ The following functions are defined.
 
 ### `AnnotationLine` type
 
+_Conversion: `Boolean`: no, `Serialize`: no, `Template`: no_
+
 The following methods are defined.
 
 * `.commit() -> Commit`: Commit responsible for changing the relevant line.
@@ -106,9 +108,13 @@ The following methods are defined.
 
 ### `Boolean` type
 
+_Conversion: `Boolean`: yes, `Serialize`: yes, `Template`: yes_
+
 No methods are defined. Can be constructed with `false` or `true` literal.
 
 ### `Commit` type
+
+_Conversion: `Boolean`: no, `Serialize`: yes, `Template`: no_
 
 This type cannot be printed. The following methods are defined.
 
@@ -151,6 +157,8 @@ This type cannot be printed. The following methods are defined.
 
 ### `ChangeId` type
 
+_Conversion: `Boolean`: no, `Serialize`: yes, `Template`: yes_
+
 The following methods are defined.
 
 * `.normal_hex() -> String`: Normal hex representation (0-9a-f) instead of the
@@ -160,12 +168,16 @@ The following methods are defined.
 
 ### `CommitId` type
 
+_Conversion: `Boolean`: no, `Serialize`: yes, `Template`: yes_
+
 The following methods are defined.
 
 * `.short([len: Integer]) -> String`
 * `.shortest([min_len: Integer]) -> ShortestIdPrefix`: Shortest unique prefix.
 
 ### `CommitRef` type
+
+_Conversion: `Boolean`: no, `Serialize`: yes, `Template`: yes_
 
 The following methods are defined.
 
@@ -190,6 +202,8 @@ The following methods are defined.
 
 ### `ConfigValue` type
 
+_Conversion: `Boolean`: no, `Serialize`: no, `Template`: yes_
+
 This type can be printed in TOML syntax. The following methods are defined.
 
 * `.as_boolean() -> Boolean`: Extract boolean.
@@ -199,6 +213,8 @@ This type can be printed in TOML syntax. The following methods are defined.
 * `.as_string_list() -> List<String>`: Extract list of strings.
 
 ### `CryptographicSignature` type
+
+_Conversion: `Boolean`: no, `Serialize`: no, `Template`: no_
 
 The following methods are defined.
 
@@ -223,6 +239,8 @@ The following methods are defined.
 
 ### `DiffStats` type
 
+_Conversion: `Boolean`: no, `Serialize`: no, `Template`: yes_
+
 This type can be printed as a histogram of the changes. The following methods
 are defined.
 
@@ -230,6 +248,8 @@ are defined.
 * `.total_removed() -> Integer`: Total number of deletions.
 
 ### `Email` type
+
+_Conversion: `Boolean`: yes, `Serialize`: yes, `Template`: yes_
 
 The email field of a signature may or may not look like an email address. It may
 be empty, may not contain the symbol `@`, and could in principle contain
@@ -244,9 +264,13 @@ The following methods are defined.
 
 ### `Integer` type
 
+_Conversion: `Boolean`: no, `Serialize`: yes, `Template`: yes_
+
 No methods are defined.
 
 ### `List` type
+
+_Conversion: `Boolean`: yes, `Serialize`: maybe, `Template`: maybe_
 
 A list can be implicitly converted to `Boolean`. The following methods are
 defined.
@@ -268,11 +292,15 @@ The following methods are defined. See also the `List` type.
 
 ### `ListTemplate` type
 
+_Conversion: `Boolean`: no, `Serialize`: no, `Template`: yes_
+
 The following methods are defined. See also the `List` type.
 
 * `.join(separator: Template) -> Template`
 
 ### `Operation` type
+
+_Conversion: `Boolean`: no, `Serialize`: yes, `Template`: no_
 
 This type cannot be printed. The following methods are defined.
 
@@ -287,11 +315,15 @@ This type cannot be printed. The following methods are defined.
 
 ### `OperationId` type
 
+_Conversion: `Boolean`: no, `Serialize`: yes, `Template`: yes_
+
 The following methods are defined.
 
 * `.short([len: Integer]) -> String`
 
 ### `Option` type
+
+_Conversion: `Boolean`: yes, `Serialize`: maybe, `Template`: maybe_
 
 An option can be implicitly converted to `Boolean` denoting whether the
 contained value is set. If set, all methods of the contained value can be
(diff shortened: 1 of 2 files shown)


thoughtpolice (approved):
I somewhat wish there was a way to generate this from the source code, but I think that's way out of reach for now. LGTM.

docs/templates.md
```diff
@@ -70,14 +70,14 @@ The following functions are defined.
...
-* `label(label: Template, content: Template) -> Template`: Apply label to
+* `label(label: PlainText, content: Template) -> Template`: Apply label to
   the content. The `label` is evaluated as a space-separated string.
 * `raw_escape_sequence(content: Template) -> Template`: Preserves any escape
   sequences in `content` (i.e., bypasses sanitization) and strips labels.
   Note: This function is intended for escape sequences and as such, its output
   is expected to be invisible / of no display width. Outputting content with
   nonzero display width may break wrapping, indentation etc.
-* `stringify(content: Template) -> String`: Format `content` to string. This
+* `stringify(content: PlainText) -> String`: Format `content` to string. This
```
ilyagr, line 80: I'm a bit confused by this. Isn't the point of `stringify` is that it converts any template to a plain string?

I also didn't follow why we need `PlainText` in addition to `Template` and `String`. Off the top of my head, I'd expect `label` and `contains_key` to both take a String. I kind of assumed that them taking a `Template` was am implementation detail, I don't know what happens if you pass a template with labels as a name for a new label.

docs/templates.md
```diff
@@ -300,6 +300,10 @@ invoked. If not set, an error will be reported inline on method call.
 On comparison between two optional values or optional and non-optional values,
 unset value is not an error. Unset value is considered less than any set values.
 
+### `PlainText` type
+
+A `String`, or any expression that can be converted to `Template`.
```
ilyagr, line 337: It might be nice to write down something about the purpose of `PlainText` here, e.g. the answers to my questions above. I may or may not be able to help once I understand what's going on.

I'm guessing this has something to do with the following commit and the note about `Serialize` below, but the connection is not obvious from reading the docs.

docs/templates.md
yuja: I added `PlainText` because I had to document `Serialize` somehow, and `PlainText` belongs to the same category. They're both protocol/trait-ish stuff. I think it's also a bit nicer to say `PlainText` is expected in error message than `Template` where string-ish type is expected.

`stringify()` uses `.try_into_plain_text()` under the hood, so the input type should be `PlainText`. Maybe it wouldn't look so bad if the type were documented as `content: Template`, but that's technically wrong.

ilyagr: It makes sense how "Serialize" is trait-ish, and the docs you added explain what it is quite well. I still don't follow what `PlainText` is about. Do we already have some user-facing docs for what `.try_into_plain_text` does and how the user should think of it? Why does the user need to know whether that function is being used?

Unless I missed something, it currently looks out of place in the docs. It leads the user to have more questions (why do some functions take PlainText if the docs say it's the same as Template?) without answering any (why is a label not a String?).

I can, of course, look at the code to try to figure it out, and will hopefully get to doing it eventually with the hope of then explaining it in the docs.

ilyagr: Another way to ask this: by the structure of this PR, I thought the docs for `Serialize` would mention or otherwise depend on `PlainText`, but I couldn't find anything like that.

ilyagr: Or is the distinction between a Template and things that can be converted to Template? I thought pretty much all of the template types can be converted to Template, but I guess there's `TreeDiff` and such.

If this is the issue, maybe `IntoTemplate` would be a better name than `PlainText`?

(I also assumed that anything that can be converted into `Template` can be passed to a function that takes `Template`. From this perspective, it's still not clear *why* we need to distinguish between `IntoTemplate` and `Template`, but I guess I must have missed something. Or maybe the answer to my first question in this comment is "no", and then `IntoTemplate` is not relevant)

yuja: > It makes sense how "Serialize" is trait-ish, and the docs you added explain what it is quite well. I still don't follow what `PlainText` is about. Do we already have some user-facing docs for what `.try_into_plain_text` does and how the user should think of it? Why does the user need to know whether that function is being used?

I don't think the user should be aware of every implementation detail. I just added it because it's _technically_ the same as `Serialize`, and it's odd that one of them were documented, and the others weren't.

And I also think `label(label: PlainText, content: Template)` is better than `label(label: Template, content: Template)`, especially in error message.

> Unless I missed something, it currently looks out of place in the docs. It leads the user to have more questions (why do some functions take PlainText if the docs say it's the same as Template?)

Maybe we can extend the documentation of `PlainText`, but I don't have an immediate plan for that.

yuja: > Or is the distinction between a Template and things that can be converted to Template?

No. `PlainText` is a kind of `String`-ish types. Any printable (or `Template`) expression can be converted to `String` through `Template` if there are no errors. Maybe it can be considered that there exists implicit `stringify()`.

> maybe IntoTemplate would be a better name than PlainText?

In that regards, `IntoString` (or `Stringify`) might be a better name.

ilyagr: > In that regards, IntoString might be a better name.

I was actually just considering mentioning that `IntoString` would make more sense from user/UI perspective. If I pass something weird as the first argument of `label`, I'd expect it to be converted into a string. For `strip_prefix`, it's a bit less clear. I could imagine matching color labels, or converting the "prefix" argument to String and then matching without regard to labels. If the "prefix" argument was "IntoString", it'd suggest the latter.

I hesitated to actually say that, because the docs you wrote for `PlainText` defined it as " any expression that can be converted to `Template`", which seems to mean the opposite of `IntoString`.

So, for example, if it would be correct to change the docs to say that `PlainText` means that this template is about to be converted into a string, I think that might be an improvement. Renaming it to `IntoString` would make sense to me as well.

yuja: > > In that regards, IntoString might be a better name.
> 
> I was actually just considering mentioning that `IntoString` would make more sense from user/UI perspective.

Are you also okay with `Stringify`? It aligns with `Serialize`, and we can rename internal functions accordingly. (FWIW, our `Template` and `PlainText` types roughly correspond to `Display` and `ToString` in Rust.)

BTW, thinking it a bit more, `Boolean` in argument position is actually `IntoBoolean`. So if we're strict, `PlainText` should be documented as `String`, and the other `String` arguments should be renamed to `StringLiteral` or something. I personally don't want to go this direction because `String` conversion is lax, so many types that don't seem like a "string" can be accepted as `String`.

> I hesitated to actually say that, because the docs you wrote for `PlainText` defined it as " any expression that can be converted to `Template`", which seems to mean the opposite of `IntoString`.
> 
> So, for example, if it would be correct to change the docs to say that `PlainText` means that this template is about to be converted into a string,

Hmm, good point. The current doc says what can be a `PlainText`, but it doesn't explain what the `PlainText` is.

ilyagr: > BTW, thinking it a bit more, Boolean in argument position is actually IntoBoolean.

I think things are fine as is. We could interpret "Boolean" as an adjective, if we really wanted.

>  I personally don't want to go this direction

I agree. If all we did was to change "Stringify" into "String" and "String" into "StringLiteral", that'd make the docs more confusing. There might be some useful point to communicate here, but I don't think it's urgent. We'd probably want to figure out what it is and why it's good for people to know more explicitly before putting it into the docs.

> many types that don't seem like a "string" can be accepted as String.

This remark makes me a bit confused about the difference between what you called "StringLiteral" and the current "Stringify". But, again, I don't think we need to fully figure this out now. I think what you have in #6860 is good.

ilyagr: Oh, and "Stringify/Serialize" sounds nice. We could also go with IntoString and IntoJSON, maybe, but that would only have noticeable benefits if we have more Into... types that we can't think of a verb for.

hotsphink: Warning: I don't know what you're talking about, I only know I've been confused about why I couldn't use a Template where a String was expected in a template.

But I have two underinformed comments anyway: (1) "Into*" is Rust jargon and if it shows up in the docs then there would need to at least be something that says "IntoX is short for 'can be converted to X'" or perhaps "IntoX means the value will first be converted to X". And (2) does Stringify here mean Stringifiable, and if so, would that be a better term for the docs? (It's not a great pattern for other types; "Booleanable" and "Integerable" are yuck.) It reminds me of "iterables" in JS, Java, and Python. I'd almost prefer custom syntax like "~String", except `~` is pretty firmly set negation/subtraction in jj-land. "≈String"? "->String"? Actually, "ToString" might be the most clear of these, despite its similarity to "IntoString".

Anyway, ignore me if I'm completely missing the point. And it sounds like there's now a Serialize precedent for using verbs as names, so maybe it's all fine. I don't understand the `label` example, so I'm probably missing a lot here. Probably something about richer-than-strings being stripped down to strings. Though I'll note that `label` currently explicitly talks about its parameter: "The `label` is evaluated as a space-separated string."

ilyagr: PTAL at whether things are any clearer after #6860, which includes an explanation of what `Stringify` means.

ilyagr: @hotsphink, I think that `ToString` is a good suggestion, maybe better than `IntoString`. I don't have a settled opinion on how good `Stringify` is; it made sense to me, but I'm not sure about everyone else. I don't like "Strigifiable" very much, it sounds a bit ugly.

> I've been confused about why I couldn't use a Template where a String was expected in a template.

I'm curious what you were doing. The solution probably would have been to use `template.to_string()` (if I recall the syntax correctly).

----

The `label` example is pretty confusing, probably picking another example would have been better. I don't think it's super-important to understand, I'm much more curious whether the docs in their post-#6860 seem generally understandable. To reply to your points, it is indeed about " richer-than-strings being stripped down to strings", and it's still true that "The label is evaluated as a space-separated string.".

hotsphink: Oh, you're right. I couldn't find any way to do it back then, but now I see that `stringify(t)` exists and would work, eg `jj --config 'template-aliases."terminate(s, t)"=if(stringify(s), s ++ t, "")'`

I can't remember what exactly it was that I was trying that didn't work; this example is a little different, in that I'm passing a Template to something wanting a Boolean.

The specific scenario where I ran into this was where I ended up using the template `diff.files().map(|f| surround("", "\n", separate("\t", f.status(), f.source().path(), f.target().path()))).join("")` -- the details don't matter, but I basically wanted to print every file touched by any change in a revset, terminated with newlines, and the straightforward `... ++ "\n"` would produce a blank line for a revision with no (relevant) changes. I couldn't figure out how to do it without using `surround` and `.join("")`. I needed `stringify` to implement a `terminate` but didn't see it. Oh! That's probably because @yuja added it just 4 months ago in 0d2becbac36c.

PR #6845, opened 2025-06-27
repo: migrate to Rust 2024 edition

thoughtpolice:
# Checklist

If applicable:


thoughtpolice:
Note: I'm not sure if we should merge this before the next release. It's less than a week away and can go in afterwords.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   hidden revision and its ancestors](docs/revsets.md#hidden-revisions) as well
   as all visible revisions.
 
+* The minimum supported Rust Version (MSRV) is now 1.88.0.
```
ilyagr: If we're going up to 1.88, I'd definitely prefer for this to go in after the upcoming release. If it goes into this release, I'd only go to 1.85. (I'm not aware of anything super-useful in 1.86 or 1.87)

1.88 does add let chains, which might be nice to be able to use. I *think* that 1.88 for the August release should be OK, it's slightly faster than I'd prefer, but might still be worth it. If somebody whom this would affect negatively sees this comment, please reply!

(My main concern is that it might negatively affect people who do not use rustup and want to use our prerelease, but there are probably not many and it hopefully won't be long until 1.88 is packaged in whatever they are using. At the moment, Debian testing is not a useful reference; they are stuck on 1.85 because trixie is being prepared for release)

flake.lock
ilyagr, line 1: (Optional) Would it make sense to split this into a separate PR?

lib/testutils/src/proptest.rs
```diff
@@ -127,7 +127,7 @@ impl WorkingCopyReferenceStateMachine {
 }
 
 impl WorkingCopyReferenceStateMachine {
-    fn arb_extant_directory(&self) -> impl Strategy<Value = RepoPathBuf> {
+    fn arb_transition_create(&self) -> BoxedStrategy<Transition> {
```
yuja: Does `+ use<>` fix the problem?

CHANGELOG.md
yuja: I would also choose 1.85 or whatever released a month back or earlier.

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * The minimum supported Rust Version (MSRV) is now 1.88.0.
 
+* The Jujutsu crates `jj_lib`/`jj_cli` are now compatible with, and compiled
+  under the 2024 Edition of Rust.
```
ilyagr: We can mention this, but we could also skip it. I think this change is purely jj-developer-facing; in particular 2020-edition Rust crates can still depend on jj-cli, and 2024-edition Rust crates could depend on it before.

martinvonz: Yes, I think we should skip it.

lib/testutils/src/proptest.rs
ilyagr: I didn't think much about what's better style, but `cargo fix --edition` would likely add `+ use<>` here (I haven't double-checked, but I'm pretty sure). That's likely the smallest change that's guaranteed to work.

In fact, if we think the refactor you did here is worth doing either way, it might still be good to use `+ use<>` first and then do the refactor in a separate commit.

(Aside: I think it's fine that you didn't use `cargo fix`. It would do a bunch things I like don't like, such as replacing all the `if let`-s with `match` to guarantee no change in behavior at the cost of style. I hope that the change to `if let` scoping rules is small enough and our tests are exhaustive enough that replacing them all with `match` is unnecessary).

CHANGELOG.md
steveklabnik: A lot of people are very excited about let chains, I expect 1.88 is going to be a very common MSRV.

ilyagr: I think we'll get to 1.88 eventually, the question is when.

Personally, I'd be OK with going to 1.88 now, though I'd mildly prefer waiting until it's packaged in nixpkgs (if it's not already, @emilazy would know), in the unlikely case it takes them longer than usual. My attempt at a more general policy, #5885, is based on a guess on how long it'd take for most packagers to catch up.

ilyagr: I think nixpkgs-unstable just successfully moved up to 1.88.

flake.lock
martinvonz: Looks like it can even be dropped now that #7023 is in

CHANGELOG.md
ilyagr: @yuja, are you OK with 1.88 now, or would you prefer to wait longer?

At this point, depending on when Austin gets to working on this, I'd be OK with waiting another couple of weeks until the release, but I'd also be happy with 1.88 now.

1.88 will be a month old in one week.

Also, Austin, if you're busy, I could try to take this over.

martinvonz: Another option is to upgrade to 1.85 now. We could also take a few of the commits in this PR without switching to the 2024 edition.

martinvonz: I'll send a PR for just that to start with.

thoughtpolice: Sure, upgrading to 1.85 sounds good to me. I can rebase this on top of it once it lands; sorry, just been busy at work and didn't allocate a lot of upstream time this past week...

yuja: > @yuja, are you OK with 1.88 now, or would you prefer to wait longer?

I would stick to the rule described in contributing.md, which is "the current `rustc` stable version, minus one." (My personal preference is more conservative, though.)

ilyagr: SGTM. We probably can merge the vast majority of this PR now that we're on 1.85 (after #7030) anyway. Let chains are nice, but I don't think they are urgent in any way.

**Update:** For reference, according to https://releases.rs, 1.89 is scheduled for 7 August, 2025. Of course, there's no reason for us to upgrade at that moment, we could wait longer.

thoughtpolice: OK, I have some work to bring this to Rust 1.85. I'll push it soon.

CHANGELOG.md
thoughtpolice: Done

lib/testutils/src/proptest.rs
thoughtpolice: Yeah, the use-bound fixed it. I've made that change and dropped this commit (but left it in my local repo in case I want to revive it.)

thoughtpolice:
This should now be ready to land if we think it's good.

lib/src/default_index/rev_walk.rs
```diff
@@ -14,8 +14,8 @@
 
 #![allow(missing_docs)]
 
-use std::cmp::max;
 use std::cmp::Reverse;
+use std::cmp::max;
```
yuja, line 18: nit: perhaps, this should be squashed into the later patch?

cli/src/cleanup_guard.rs
```diff
@@ -55,41 +55,43 @@ mod platform {
     use std::sync::atomic::Ordering;
     use std::thread;
 
-    use libc::c_int;
     use libc::SIGINT;
     use libc::SIGTERM;
+    use libc::c_int;
```
yuja, line 60: nit: same here. squash into "repo: move to Rust 2024 language edition"?

Cargo.toml
yuja, line 1: In commit message:
> repo: move to Rust 2024 language edition

"repo:" usually means lib/src/repo.rs. "cleanup:", "cli, lib:", etc. seems better.

yuja (approved):

lib/src/default_index/rev_walk.rs
thoughtpolice: Not sure how these slipped in. :) Thanks

Cargo.toml
PhilipMetzger: I still think Yuya's suggestions are better than `chore:` since that is essentially useless.

martinvonz:
Sorry, I just queued #7071 without thinking about this PR. I suppose you'll have to rebase it again since this PR touches basically every file. I'll avoid queuing anything else until this is in.

thoughtpolice:
Yeah it's no big deal, not really a problematic one. I'll go ahead and just auto queue this one.

PR #6847, opened 2025-06-28
cli: git-push: make "change" bookmark names templated

yuja:
# Checklist

If applicable:


cli/src/complete.rs
```diff
@@ -57,6 +58,21 @@ fn split_help_text(line: &str) -> (&str, Option<StyledStr>) {
     }
 }
 
+// TODO: It's weird that the completion candidates are prioritized based on the
+// prefix found in the template. Delete this feature or add explicit config?
+fn extract_git_push_bookmark_prefix(settings: &UserSettings) -> Option<String> {
+    let text = settings.get_string("templates.git_push_bookmark").ok()?;
+    let node = template_parser::parse_template(&text).ok()?;
```
PhilipMetzger: cc @senekor for a explanation?

senekor: The idea is that bookmarks with a matching prefix are more likely created by the user themself. This pattern is often used in the jj repo where people use their name or github username as push prefix. I do it in all my other repos as well. In places where the prefix is prioritized, it's based on the assumption that users want to interact with their own bookmarks more often than the ones of other people.

I don't know the different use cases people want to use this new template for, so I can't say what could be a better prioritization. This function seems to match the old behavior well.

PhilipMetzger: > I don't know the different use cases people want to use this new template for, so I can't say what could be a better prioritization. This function seems to match the old behavior well.

@urschrei's use-case was Jira-Task prefixes in a work environment, so making it making it configurable for that also makes sense.

PhilipMetzger: sorry for the misinformation here it wasn't Urschrei's request but another user, see the request [here](https://discord.com/channels/968932220549103686/968932220549103689/1387398018453737654).

senekor: Got it. I guess this function shouldn't interfere with that, since such a template wouldn't consist of a string and then a concatenation.

yuja: Just curious. Do you have many local bookmarks that don't match the prefix (or local bookmarks you aren't going to update)?

It would be nice if we could remove this sorting feature because extracting prefix from the template is ugly. My feeling is that if we really need this, we should add `completion.prioritized-bookmarks` or something.

senekor: I think it's fine to remove this. It made sense to me at the time, but doesn't add a ton of value.

yuja: Removed the prefix handling from complete.rs, thanks.

ilyagr (approved):
Thanks Yuya and Philip!

yuja:
I'm gonna merge this as we have ~one more day before the release.

cli/src/complete.rs
joneshf: > sorry for the misinformation here it wasn't Urschrei's request but another user, see the request [here](https://discord.com/channels/968932220549103686/968932220549103689/1387398018453737654).

👋 Hi! I'm that other user.

@yuja Thanks a bunch for implementing this. I'm excited to give this a whirl next chance I get!

PR #6848, opened 2025-06-28
revset: add `parents(x, depth)` and `children(x, depth)`

### LINKED ISSUE #3337: FR: Revset expression for nth parent/child
In Git, I'm used to doing commands like `git diff HEAD~3` to see the changes from the most recent 3 commits. However, there seems to be no easy way to do this with jj. Either I have to look at the log to find the change ID prefix like `jj diff --from v`, or I have to use multiple minus signs like `jj diff --from @---`. The same sort of issue might happen with `+`, but just less commonly.

I'm not sure about the syntax, but something like `@-3` to mean `@---` would be nice. However it seems like it was already discussed in #46, and it sounds like there were some concerns about ambiguity with branch names containing hyphens. Perhaps it could use braces for disambiguation similar to Git's `x@{N}` syntax, so `x-{N}` would mean the Nth level of parents of `x`, but without braces `x-N` would be a regular branch name.

If it isn't possible to have it be an operator, it might be possible to add it to the `children` and `parents` revset functions. For instance, `children(x[, count])` and `parents(x[, count])`. This would have the advantage of matching with `ancestors(x[, depth])`, but it is a bit clunky to write.

scott2000:
Resolves #3337.

# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_revset_output.rs
+++ b/cli/tests/test_revset_output.rs
@@ -104,27 +104,27 @@ fn test_bad_function_call() {
     let output = work_dir.run_jj(["log", "-r", "parents()"]);
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
-    Error: Failed to parse revset: Function `parents`: Expected 1 arguments
+    Error: Failed to parse revset: Function `parents`: Expected 1 to 2 arguments
     Caused by:  --> 1:9
       |
     1 | parents()
       |         ^
       |
-      = Function `parents`: Expected 1 arguments
+      = Function `parents`: Expected 1 to 2 arguments
     [EOF]
     [exit status: 1]
     ");
 
-    let output = work_dir.run_jj(["log", "-r", "parents(foo, bar)"]);
+    let output = work_dir.run_jj(["log", "-r", "parents(foo, bar, baz)"]);
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
-    Error: Failed to parse revset: Function `parents`: Expected 1 arguments
+    Error: Failed to parse revset: Function `parents`: Expected 1 to 2 arguments
     Caused by:  --> 1:9
       |
-    1 | parents(foo, bar)
-      |         ^------^
+    1 | parents(foo, bar, baz)
+      |         ^-----------^
       |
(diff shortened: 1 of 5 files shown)


cli/tests/test_revset_output.rs
```diff
@@ -104,27 +104,27 @@ fn test_bad_function_call() {
...
     1 | parents()
       |         ^
       |
-      = Function `parents`: Expected 1 arguments
+      = Function `parents`: Expected 1 to 2 arguments
     [EOF]
     [exit status: 1]
     ");
 
     let output = work_dir.run_jj(["log", "-r", "parents(foo, bar)"]);
```
yuja: nit: can you add one more parameter? I think this is the test for lack/excessive arguments.

lib/src/revset.rs
```diff
@@ -683,14 +683,24 @@ static BUILTIN_FUNCTION_MAP: Lazy<HashMap<&'static str, RevsetFunction>> = Lazy:
     // code completion inside macro is quite restricted.
     let mut map: HashMap<&'static str, RevsetFunction> = HashMap::new();
     map.insert("parents", |diagnostics, function, context| {
-        let [arg] = function.expect_exact_arguments()?;
+        let ([arg], [depth_opt_arg]) = function.expect_arguments()?;
         let expression = lower_expression(diagnostics, arg, context)?;
-        Ok(expression.parents())
+        if let Some(depth_arg) = depth_opt_arg {
+            let depth = expect_literal("integer", depth_arg)?;
+            Ok(expression.ancestors_at(depth))
```
yuja, line 690: nit: this can panic by `u64::MAX`. Since it's unlikely, and we clap overflow values in optimization pass, I think it's okay to clap overflow at `.ancestors`/`descendants_at()`. Returning error is also good, but might be a bit more verbose.

lib/src/revset.rs
```diff
@@ -3629,9 +3639,13 @@ mod tests {
         "#);
         insta::assert_debug_snapshot!(
             parse("parents(foo,foo)").unwrap_err().kind(), @r#"
```
yuja: nit: same here, add one more parameter.

yuja (approved):

PR #6855, opened 2025-06-28
squash: allow to squash into the sources

glehmann:
squashing into the same change as the source becomes a no-op.

related to #5301

# Checklist

If applicable:


cli/src/commands/squash.rs
```diff
@@ -133,8 +133,11 @@ pub(crate) fn cmd_squash(
...
         destination = workspace_command
             .resolve_single_rev(ui, args.into.as_ref().unwrap_or(&RevisionArg::AT))?;
-        if sources.iter().any(|source| source.id() == destination.id()) {
-            return Err(user_error("Source and destination cannot be the same"));
+        // remove the destination from the sources
+        sources.retain(|source| source.id() != destination.id());
+        if sources.is_empty() {
+            writeln!(ui.status(), "Nothing changed.")?;
+            return Ok(());
         }
```
yuja: This shouldn't be needed since it's equivalent to `--from=none()`.

yuja (approved):
Sounds good since moving patch to the same commit should be noop.

PR #6863, opened 2025-06-29
config: move fsmonitor settings to working-copy

06393993:
According to the suggestion in https://github.com/jj-vcs/jj/issues/6762#issuecomment-2974778137.

# Checklist

If applicable:

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -741,6 +741,10 @@ pub fn default_config_migrations() -> Vec<ConfigMigrationRule> {
                 Ok(format!(":{value}").into())
             },
         ),
+        // TODO: Delete in jj 0.37.0+
+        ConfigMigrationRule::rename_value("core.fsmonitor", "working-copy.fsmonitor"),
+        // TODO: Delete in jj 0.37.0+
+        ConfigMigrationRule::rename_value("core.watchman", "working-copy.watchman"),
     ]
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Deprecations
 
+* Rename the `core.fsmonitor` and `core.watchman` settings to
+  `working-copy.fsmonitor` and `working-copy.watchman` settings respectively.
+
 ### New features
 
 * `jj diff` now accepts `-T`/`--template` option to customize summary output.
--- a/cli/src/commands/debug/watchman.rs
+++ b/cli/src/commands/debug/watchman.rs
@@ -59,10 +59,13 @@ pub fn cmd_debug_watchman(
             // TODO(ilyagr): It would be nice to add colors here
             let config = match workspace_command.settings().fsmonitor_settings()? {
                 FsmonitorSettings::Watchman(config) => {
-                    writeln!(ui.stdout(), "Watchman is enabled via `core.fsmonitor`.")?;
                     writeln!(
                         ui.stdout(),
-                        r"Background snapshotting is {}. Use `core.watchman.register-snapshot-trigger` to control it.",
+                        "Watchman is enabled via `working-copy.fsmonitor`."
+                    )?;
+                    writeln!(
+                        ui.stdout(),
+                        r"Background snapshotting is {}. Use `working-copy.watchman.register-snapshot-trigger` to control it.",
                         if config.register_trigger {
                             "enabled"
                         } else {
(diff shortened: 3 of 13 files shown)


martinvonz (commented):
Thanks! This LGTM but I'll let someone else approve it since I suggested the change.

cli/src/config.rs
```diff
@@ -741,6 +741,10 @@ pub fn default_config_migrations() -> Vec<ConfigMigrationRule> {
                 Ok(format!(":{value}").into())
             },
         ),
+        // TODO: Delete in jj 0.37.0+
+        ConfigMigrationRule::rename_value("core.fsmonitor", "working-copy.fsmonitor"),
+        // TODO: Delete in jj 0.37.0+
+        ConfigMigrationRule::rename_value("core.watchman", "working-copy.watchman"),
```
PhilipMetzger: nit: it's still missing a warning here, that the old value is deprecated and to be removed in the future and that it was migrated to the new value

PhilipMetzger (approved):
LG, @yuja should approve though since he wrote these APIs

cli/src/config.rs
06393993: I thought this change generates the expected CLI warnings:

```
PS C:\src\jj> cargo run -- config set core.fsmonitor none --repo
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.45s
     Running `target\debug\jj.exe config set core.fsmonitor none --repo`
Warning: Deprecated repo-level config: core.fsmonitor is renamed to working-copy.fsmonitor
PS C:\src\jj> cargo run -- config get --config=core.fsmonitor=none core.fsmonitor
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s
     Running `target\debug\jj.exe config get --config=core.fsmonitor=none core.fsmonitor`
Warning: Deprecated repo-level config: core.fsmonitor is renamed to working-copy.fsmonitor
Warning: Deprecated CLI-provided config: core.fsmonitor is renamed to working-copy.fsmonitor
Config error: Value not found for core.fsmonitor
For help, see https://jj-vcs.github.io/jj/latest/config/ or use `jj help -k config`.
error: process didn't exit successfully: `target\debug\jj.exe config get --config=core.fsmonitor=none core.fsmonitor` (exit code: 1)
```

Can you be specific on what I am missing? Thanks.

PhilipMetzger: Sorry, it seems the last few depreactions/renames no longer mention the exact removal date, so I should move with the times.

yuja:
Looks good, but do we want this in tomorrow's release? I think the rename only makes sense when we add eol configuration.

Another concern is that we can integrate fsmonitor into Git import. Since Git import could be considered a part of snapshotting process, `snapshot.fsmonitor` might work.

06393993:
> we can integrate fsmonitor into Git import

What about another setting to control the git import behavior separately and land this PR as is?

martinvonz:
> Looks good, but do we want this in tomorrow's release?

I think it can wait. No rush at all.

> I think the rename only makes sense when we add eol configuration.

My reason for suggesting `[working-copy]` instead of `[core]` was that we don't really have a concept of "core". I thought this feature was specific to the working copy, but good point that we might want to add watchman triggers for Git imports. Perhaps that means that `core.fsmonitor` makes sense to move under `[working-copy]` but `core.watchman.register-snapshot-trigger` should move under `[watchman]` or something?

Either way, I think the rename can easily wait for the eol configuration. That might be good as a way of making it clearer to users why we renamed the config.

yuja:
Or just `fsmonitor`?

```
[fsmonitor]
backend = "watchman"
#working-copy = true
#git = true
#snapshot-trigger = false
```

I don't think users would want to enable fsmonitor integration per underlying function. They just want faster `jj log`, etc.

martinvonz:
> Or just `fsmonitor`?

True, that would also work. I don't feel strongly for one or the other.

06393993:
@yuja , given that the EOL conversion PR https://github.com/jj-vcs/jj/pull/6728 is merged. What's our opinion on the naming of the settings? Can we make a final decision? Thanks!

yuja:
I think `[fsmonitor]` is the least bad choice. I don't have a strong feeling about `[fsmonitor]` vs `[snapshot.fsmonitor]`, but I think `[working-copy.fsmonitor]` will be bad for future extension.

06393993:
In that case, let me start with the flat `[fsmonitor]` choice. Thanks.

martinvonz:
The commit description and PR description seem stale

06393993:
> The commit description and PR description seem stale

Done.

PhilipMetzger (approved):

bryceberger:
Just ran into an issue with this. Given the following config:

```toml
[core]
fsmonitor = "watchman"
watchman.register-snapshot-trigger = true
```

Results in the following hard error on every `jj` command (including `jj config edit --user`!!):
```
Config error: Migration failed
Caused by: Would delete entire table core.watchman
Hint: Check the config file: /home/bryce/.config/jj/config.toml
For help, see https://jj-vcs.github.io/jj/latest/config/ or use `jj help -k config`.
```

Given that I believe this is a fairly common configuration, it's a bit annoying that it works out like this. Unsure of the best way to address.

yuja:
> Caused by: Would delete entire table core.watchman

https://github.com/jj-vcs/jj/pull/7054

PR #6866, opened 2025-06-30
make diff --stat show byte counts for binary files

evmar:
This makes `jj diff --stat` show text like `(binary) +123 bytes` for binary files, rather than the current behavior of counting the newlines within the binary files.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj git remote add` and `jj git clone` now support `--fetch-tags` to control
   when tags are fetched
 
+* `jj diff --stat` now shows the change in size to binary files.
+
 ### Fixed bugs
 
 ### Packaging changes
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -1984,11 +1984,7 @@ impl DiffStats {
                 let (left, right) = values?;
                 let left_content = diff_content(path.source(), left, conflict_marker_style)?;
                 let right_content = diff_content(path.target(), right, conflict_marker_style)?;
-                let stat = get_diff_stat_entry(
-                    path,
-                    [&left_content.contents, &right_content.contents].map(BStr::new),
-                    options,
-                );
+                let stat = get_diff_stat_entry(path, [&left_content, &right_content], options);
                 BackendResult::Ok(stat)
             })
             .try_collect()
@@ -2001,49 +1997,64 @@ impl DiffStats {
         &self.entries
     }
 
-    /// Total number of insertions.
+    /// Total number of inserted lines.
     pub fn count_total_added(&self) -> usize {
-        self.entries.iter().map(|stat| stat.added).sum()
+        self.entries
+            .iter()
+            .filter_map(|stat| stat.added_removed.map(|(added, _)| added))
+            .sum()
     }
 
-    /// Total number of deletions.
+    /// Total number of deleted lines.
     pub fn count_total_removed(&self) -> usize {
-        self.entries.iter().map(|stat| stat.removed).sum()
+        self.entries
+            .iter()
+            .filter_map(|stat| stat.added_removed.map(|(_, removed)| removed))
+            .sum()
     }
 }
 
 #[derive(Clone, Debug)]
 pub struct DiffStatEntry {
     pub path: CopiesTreeDiffEntryPath,
-    pub added: usize,
-    pub removed: usize,
+    /// Lines added and removed; None for binary files.
+    pub added_removed: Option<(usize, usize)>,
+    /// Change in file size in bytes.
+    pub bytes_delta: isize,
 }
 
 fn get_diff_stat_entry(
     path: CopiesTreeDiffEntryPath,
-    contents: [&BStr; 2],
+    contents: [&FileContent<BString>; 2],
     options: &DiffStatOptions,
 ) -> DiffStatEntry {
-    // TODO: this matches git's behavior, which is to count the number of newlines
-    // in the file. but that behavior seems unhelpful; no one really cares how
-    // many `0x0a` characters are in an image.
-    let diff = diff_by_line(contents, &options.line_diff);
-    let mut added = 0;
-    let mut removed = 0;
-    for hunk in diff.hunks() {
-        match hunk.kind {
-            DiffHunkKind::Matching => {}
-            DiffHunkKind::Different => {
-                let [left, right] = hunk.contents[..].try_into().unwrap();
-                removed += left.split_inclusive(|b| *b == b'\n').count();
-                added += right.split_inclusive(|b| *b == b'\n').count();
+    let [left_content, right_content] = contents;
+    let added_removed = if left_content.is_binary || right_content.is_binary {
+        None
+    } else {
+        let diff = diff_by_line(
+            contents.map(|content| &content.contents),
+            &options.line_diff,
+        );
+        let mut added = 0;
+        let mut removed = 0;
+        for hunk in diff.hunks() {
+            match hunk.kind {
+                DiffHunkKind::Matching => {}
+                DiffHunkKind::Different => {
+                    let [left, right] = hunk.contents[..].try_into().unwrap();
+                    removed += left.split_inclusive(|b| *b == b'\n').count();
+                    added += right.split_inclusive(|b| *b == b'\n').count();
+                }
             }
         }
-    }
+        Some((added, removed))
+    };
+
     DiffStatEntry {
         path,
-        added,
-        removed,
+        added_removed,
+        bytes_delta: right_content.contents.len() as isize - left_content.contents.len() as isize,
     }
 }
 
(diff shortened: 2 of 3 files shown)


cli/src/diff_util.rs
```diff
@@ -1962,50 +1958,76 @@ impl DiffStats {
...
-    pub added: usize,
-    pub removed: usize,
+    pub stats: DiffStatStats,
+}
+
+#[derive(Clone, Debug)]
+pub enum DiffStatStats {
+    Lines { added: usize, removed: usize },
+    Bytes { before: usize, after: usize },
 }
```
yuja: I think it's better to store before/after bytes unconditionally. It could be provided as `diff.stat().entries().map(|e| e.before_size())` in template, if we add `.entries()` function.

```rs
added_removed: Option<(usize, usize)>,
before_bytes: usize,
after_bytes: usize,
```

cli/tests/test_diff_command.rs
```diff
@@ -3722,12 +3722,12 @@ fn test_diff_binary() {
...
-    binary_modified.png         | 6 +++---
-    binary_modified_to_text.png | 5 ++---
-    binary_removed.png          | 3 ---
-    binary_valid_utf8.png       | 1 +
-    5 files changed, 9 insertions(+), 9 deletions(-)
+    binary_added.png            | 0 -> 12 bytes
+    binary_modified.png         | 16 -> 9 bytes
+    binary_modified_to_text.png | 16 -> 8 bytes
+    binary_removed.png          | 16 -> 0 bytes
+    binary_valid_utf8.png       | 0 -> 3 bytes
```
yuja: Git appears to print sizes as `| Bin xxx -> yyy bytes`

cli/src/diff_util.rs
```diff
@@ -2029,7 +2051,10 @@ pub fn show_diff_stats(
     let max_diffs = stats
         .entries()
         .iter()
-        .map(|stat| stat.added + stat.removed)
+        .filter_map(|stat| match &stat.stats {
+            DiffStatStats::Lines { added, removed } => Some(added + removed),
+            DiffStatStats::Bytes { .. } => None,
+        })
```
yuja, line 2103: Perhaps, we'll need to calculate maximum length of ` | Bin xxx -> yyy bytes` to allocate space for that.

cli/tests/test_diff_command.rs
evmar: I saw that, but I wasn't sure if we want to match Git's output exactly.  If not, I didn't think the word "Bin" was really good UI for it.  What would you prefer?

cli/src/diff_util.rs
evmar: I pushed a new change with this structure in place.

cli/src/diff_util.rs
evmar: Is there a test that exercises these narrow screen widths?  I looked a bit but I am not sure where to look.

cli/tests/test_diff_command.rs
martinvonz: Maybe `binary_modified.png         | 16 -> 9 bytes (binary)` or `binary_modified.png (binary)   | 16 -> 9 bytes` but I don't feel strongly.

evmar: I guess between those, the former is better than the latter, because the ` | ` part is the left edge of the graph shown for regular diffs and we want to give it the most space.

yuja: If we don't have to copy the Git behavior, just printing "(binary)" might also be good. I don't think the exact size in bytes is useful.

cli/src/diff_util.rs
yuja: search `COLUMNS` in test_diff_command.rs

cli/tests/test_diff_command.rs
joyously: I thought the byte count was to fulfill the `--stat` option.

evmar: Regarding usefulness: the reason I discovered this bug in the first place is that I had run the command to see the difference in byte size!

If you think it is truly not useful, I guess I could rely on `git diff --stat` when I want to see this information, though, hrm.

yuja: I personally don't think the change in file size is interesting, but I'm not against adding it if there is a use case. Maybe we can also use `fmt_utils::binary_prefix()` to make these numbers readable, but I don't know if that's what you want.

ilyagr: Two use-cases come to my mind: 1) when I was updating the png images for jj's website, I cared about the change to the filesize and 2) some problems can be caught from noticing an unexpected change in file size (e.g. 500KB -> 5B, or vice-versa).

@yuja, what does `binary_prefix()` do? I couldn't immediately find it.

I can think of a few possible vague categories of formats:

0) Number of bytes -> Number of bytes
1) Same, but with `_` separators between thousands, millions, etc. We could later add an option to the templater to make the `_`s grey.
2) Humanized and rounded sizes, but this would require some cleverness. It could be "2.0 MB -> 1.5 MB" or it might have to be "2.5 MB (-15 bytes)".
3) Final size (percentage change), e.g. "2.5MB (-0.1%)". I'm not sure how to round this if 1 byte turns into 2.5 MB, though; perhaps `+250000000%` is OK.
4) Something in between 2 and 3: "2.0 MB -> 1.5 MB" or, if the rounded number are the same, go with "2.5 MB (-0.1%)".

My first thought would be to do option 1 (that, in my mind, is a compromise option that's both easy to implement and useful, at the cost of prettiness) or 3, but I'm not sure. Option 3 would cover all of my use-cases, but I'm not sure about everyone else. I'm OK with just going with option 0 for a first pass, and then we can think about it later.

-----

I do like Martin's idea of putting `(binary)` somewhere.

ilyagr: > what does binary_prefix() do? I couldn't immediately find it.

Found it. It's in our codebase, not stdlib. It converts to KiB, MiB, GiB, etc.

https://github.com/jj-vcs/jj/blob/main/lib%2Fsrc%2Ffmt_util.rs#L17-L23

evmar: Re use cases, I think the overall use of `diff --stat` is to get a big-picture view of "what kind of change is this?"  For code, lines added and removed show whether a module is growing or shrinking and by how much, so you can kind of eyeball "oh this change removes code from X and adds code to Y".  For binary files, bytes are showing roughly the same thing, so it's the same kinds of questions.  So e.g. if you're archiving zip files or something, you can see whether they are growing or shrinking in size.  For my specific use case I was also trying to see whether a file was growing, like the png example above.

I think the purpose of `--stat` overall isn't trying to give you the detail of what changed, so I like the ideas around showing percentages.  The specific byte counts are more information than you really need and can make the output harder to skim.  Here's a snippet of the output for a real change I was working on:

```
│ ○  knyyrsok 3 months ago wip: drop retrowin32_syscall
├─╯  win32/derive/src/dll.rs           |  2 +-
│    win32/dll/advapi32.dll            | 2560 -> 2048 bytes
│    win32/dll/bass.dll                | 2560 -> 2048 bytes
│    win32/dll/comctl32.dll            | 2560 -> 2048 bytes
│    win32/dll/ddraw.dll               | 13824 -> 13824 bytes
│    win32/dll/dinput.dll              | 3584 -> 3584 bytes
│    win32/dll/dsound.dll              | 3584 -> 3584 bytes
│    win32/dll/gdi32.dll               | 3584 -> 3072 bytes
│    win32/dll/kernel32.dll            | 10240 -> 8704 bytes
│    win32/dll/ntdll.dll               | 2560 -> 2048 bytes
```

If you read the numbers carefully you can see some files sizes went down (which was the purpose of the change) and some stayed the same, but it's not great.  It's certainly better than what jj does now though, which is showing the counts of newline bytes in these files!


The number of ideas brought up here make me wonder whether this kind of choice belongs in the templating language, though I'm not sure how/whether that works in general (this is my first jj change).  From that perspective maybe I ought to just match what Git does for now?

evmar: Thinking about this more: for text files, the --stat output doesn't show the original size of the files, so you cannot at a glance tell whether a file with a bunch of lines removed is a complete deletion vs just partially removed.  As a user of stat this has been fine in my experience.

We could do the same for binary: only show the delta, in either bytes, human-friendly bytes, or percentages.  I think it addresses all the use cases we've discussed so far with simpler output.

ilyagr: I like this idea. Percentages might be an orthogonal issue; I can imagine an option where percentages are shown both for binary and for text files (indicating the change to the number of lines, I guess).

cli/src/diff_util.rs
evmar: I added this, thanks for catching it!
You can now see the text getting elided in the test: `...fied_to_text.png | (binary) -8 bytes`

cli/src/diff_util.rs
```diff
@@ -2052,56 +2063,89 @@ pub fn show_diff_stats(
...
+        } else {
+            max_bytes = max_bytes.max(stat.bytes_delta.abs());
+        }
+    }
+
+    let max_binary_output_width = if max_bytes > 0 {
+        "(binary) +".len() + max_bytes.to_string().len() + " bytes".len()
+    } else {
+        0
+    };
```
yuja, line 2111: Since binary diff can be the same size, `max_bytes == 0` doesn't mean there are no binary entries. I think it's simpler to use separate `.iter().filter_map().max()` loop to handle the empty case.

Can you also add a test for binary of 0-size delta?

cli/src/diff_util.rs
```diff
@@ -2052,56 +2063,89 @@ pub fn show_diff_stats(
...
+
+    // Choose how many columns to use for the path.  The bar will use the rest.
+    // Always assume at least a tiny bit of room:
+    let available_width = max(display_width.saturating_sub(" | ".len()), 5);
+    let max_path_width = max_path_width
+        .min(available_width.saturating_sub(max_binary_output_width))
```
yuja: nit: maybe we should subtract `diff_number_width + margin` as well for consistency?

cli/src/diff_util.rs
```diff
@@ -2052,56 +2063,89 @@ pub fn show_diff_stats(
...
+            write!(
+                formatter,
+                "{:>diff_number_width$}{}",
+                added + removed,
+                if bar_added + bar_removed > 0 { " " } else { "" },
+            )?;
+            write!(formatter.labeled("added"), "{}", "+".repeat(bar_added))?;
+            writeln!(formatter.labeled("removed"), "{}", "-".repeat(bar_removed))?;
+        } else {
+            write!(formatter, "(binary)")?;
```
yuja: nit: `.labeled("binary")`

cli/src/diff_util.rs
```diff
@@ -2052,56 +2063,89 @@ pub fn show_diff_stats(
...
+            if stat.bytes_delta != 0 {
+                if stat.bytes_delta < 0 {
+                    write!(formatter.labeled("removed"), " {}", stat.bytes_delta)?;
+                } else if stat.bytes_delta > 0 {
+                    write!(formatter.labeled("added"), " +{}", stat.bytes_delta)?;
```
yuja: nit: we could use `{:+}` instead, but I don't feel strongly.

yuja (commented):
The output looks good to me, thanks.

cli/src/diff_util.rs
evmar: This is subtle in a kind of annoying way, so I will type it here rather than in a comment in the code and ask your thoughts.

The last `.min(0.7 * available)` means, we will always give at least 30% of the available space to the right side.  So the other `.min()`s here only will have an effect if the right side demands >30% of the available space.

The `diff_number_width + margin` computation is the minimum size we allocate to the text diff lines, where the previous code's margin was 1 and so the total value would typically be something small like 6.  This means adding would only matter when 6 > 0.3*available, which is when the available screen width is 20 or less.  At that display size I think we don't really expect this to render well.

The case where we only have 0-byte-sized binary diffs, where the right side output is the string `"(binary)"`, is similar, which is why I hadn't already handled the issue you spotted in your previous comment.

For both of these I am fine with writing the code to try to measure them.  (In a previous iteration on this patch, which I never sent for review, I just ran the printing logic twice -- once into a throwaway buffer to measure the width, then second time to actually render it.)

I am only worried about whether I am making this code extra complex for these cases where it doesn't end up mattering too much.

So, just to say again what I'm trying to say here:
1. I am happy to fix both of these in the way you suggest;
2. I hadn't done it up front because I was worried I was making this code extra complex for something that didn't matter much.

Please let me know what you would prefer and I will do it!  I also wasn't sure whether/how to write the above reasoning in a comment in the code or not...


(By the way, the reason we have this 30% threshold is because we want to reserve some space for the `++++----` lines.  If a diff has only binary changes, we wouldn't render those lines and wouldn't need the 30% threshold part.  I could also address this if you wanted me to try to make every line the maximal possible width.)

cli/src/diff_util.rs
evmar: I think your idea is better, in particular because I can use it in both branches for symmetry.

cli/src/diff_util.rs
evmar: After writing this comment, I think I convinced myself I should just write the code, since it would be shorter than this comment anyway.  I will update this patch with the full fix.

evmar:
Note to self, a nice way to eyeball this on a lot of code is to run:

`./target/debug/jj diff -f zzzzzzzz  --stat`

which looks like this on the jj project with this patch

image

cli/src/diff_util.rs
yuja: > This is subtle in a kind of annoying way, so I will type it here rather than in a comment in the code and ask your thoughts.
> 
> The last `.min(0.7 * available)` means, we will always give at least 30% of the available space to the right side. So the other `.min()`s here only will have an effect if the right side demands >30% of the available space.
> 
> The `diff_number_width + margin` computation is the minimum size we allocate to the text diff lines, where the previous code's margin was 1 and so the total value would typically be something small like 6. This means adding would only matter when 6 > 0.3*available, which is when the available screen width is 20 or less. At that display size I think we don't really expect this to render well.
> 
> The case where we only have 0-byte-sized binary diffs, where the right side output is the string `"(binary)"`, is similar, which is why I hadn't already handled the issue you spotted in your previous comment.

These statements are true, but aren't obvious in the code. We could add inline comments describing why we can omit these computations, but I think it's better to make the code do the right thing. The extra calculation cost wouldn't matter. I think it's easier to follow if `available_width - max_binary_output_width` were `available_width - right_side_width_to_allocate`.

cli/src/diff_util.rs
evmar: I have rewritten this for clarity, and added a test that includes 0-size delta, see `test_diff_stat_binary_and_text` at the bottom of the new diff.  In particular:
- when there is only 0-size delta then the right side width is just "(binary)"
- when there is additionally a text diff then the right side width is larger, for space for the graph
- when there is a binary delta size change then the right side width is even larger

In the code, this is handled by separately computing the maximum right side based on text diffs (including the case where there are none) from the the maximum right side based on binary diffs (including the case where there are none).

cli/tests/test_diff_command.rs
```diff
@@ -3726,17 +3726,18 @@ fn test_diff_stat_rounding() {
     let work_dir = test_env.work_dir("repo");
     let output = work_dir.run_jj(["diff", "--stat"]);
     insta::assert_snapshot!(output, @r"
-    ... | 101 +-
-    ... | 301 +-
-    ... |  10 +
+    .. | 101 +-
```
evmar, line 3729: This changed due to small changes in the math, but it's testing we don't crash here so I don't think the difference matters.

CHANGELOG.md
```diff
@@ -186,6 +186,8 @@ Thanks to the people who made this release happen!
   `colors."diff removed token" = { bg = "ansi-color-52", underline = false }`
   will apply a dark red background on removed words in diffs.
 
+* `jj diff --stat` now shows the change in size to binary files.
```
yuja: nit: move to "Unreleased"

cli/src/diff_util.rs
```diff
@@ -2067,60 +2078,111 @@ pub fn show_diff_stats(
...
-            (len, bar_length - len)
+    // Measure widest right side for binary diffs and reduce max_path_width if
+    // needed.
+    let max_bytes = stats
+        .entries
+        .iter()
+        .filter_map(|stat| match stat.added_removed {
+            Some(_) => None,
+            None => Some(stat.bytes_delta.abs()),
+        })
```
yuja: nit: `.filter(|stat| stat.added_removed.is_none()).map(..)` also works

cli/src/diff_util.rs
```diff
@@ -2067,60 +2078,111 @@ pub fn show_diff_stats(
...
+                if bar_added + bar_removed > 0 { " " } else { "" },
+            )?;
+            write!(formatter.labeled("added"), "{}", "+".repeat(bar_added))?;
+            writeln!(formatter.labeled("removed"), "{}", "-".repeat(bar_removed))?;
+        } else {
+            write!(formatter.labeled("binary"), "(binary)")?;
+            if stat.bytes_delta != 0 {
+                if stat.bytes_delta < 0 {
+                    write!(formatter.labeled("removed"), " {:+}", stat.bytes_delta)?;
+                } else if stat.bytes_delta > 0 {
```
yuja: nit: just `} else {` (because we'll print " bytes" unconditionally)?

It might be also good to do `let label = if ...` and `write!(formatter.labeled(label), ..)`.

yuja (approved):
Looks good, thanks!

PR #6477, opened 2025-05-08
revset: add strict parameter to diff_contains()

winterqt:
# Checklist

If applicable:
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [ ] I have added tests to cover my changes

PR #6867, opened 2025-06-30
docs: Add git log -S to git command table

raylu:
# Checklist

If applicable:


docs/git-command-table.yml
```diff
@@ -141,6 +141,13 @@
     `jj log`
   Notes: ''
 
+- Use case: Show log of commits adding/removing the string "stuff"
+  Git command: >
+    `git log -S stuff`
+  Jujutsu command: >
+    `jj log 'diff_contains(stuff)'`
```
martinvonz: needs a `-r`

docs/git-command-table.yml
```diff
@@ -141,6 +141,13 @@
     `jj log`
   Notes: ''
 
+- Use case: Show log of commits adding/removing the string "stuff"
+  Git command: >
+    `git log -S stuff`
```
yuja: Maybe this should be `git log -G`? `diff_contains()` doesn't compare occurrences.

raylu: maybe I messed up, but i added `frotz` to a file, committed, then moved it to a different place in the file and committed again. `diff_contains(frotz)` only found the first commit
although #6477 implies I'm wrong...

yuja: `diff_contains()` has optimization to compare only matching lines, so if there are no context lines matching `frotz` in between, the change isn't detected. It usually works more like `git log -G` than `-S`.

ilyagr:
Now that you changed the text to use `-G`, could you also change the commit description and the PR title?

martinvonz (approved):

PR #6869, opened 2025-06-30
templater: serialize map literal by `json({..})`

yuja:
I don't have strong feeling whether we should add this syntax, but it would be
tedious to concatenate multiple serializable values (such as commit, tags, and
bookmarks) manually as JSON object.

# Checklist

If applicable:

### DIFF
--- a/cli/src/template_parser.rs
+++ b/cli/src/template_parser.rs
@@ -97,6 +97,9 @@ impl Rule {
             Self::function_arguments => None,
             Self::lambda => None,
             Self::formal_parameters => None,
+            Self::map => None,
+            Self::map_entries => None,
+            Self::map_entry => None,
             Self::pattern_identifier => None,
             Self::pattern => None,
             Self::pattern_value_expression => None,
@@ -300,6 +303,7 @@ pub enum ExpressionKind<'i> {
     FunctionCall(Box<FunctionCallNode<'i>>),
     MethodCall(Box<MethodCallNode<'i>>),
     Lambda(Box<LambdaNode<'i>>),
+    Map(Vec<MapNodeEntry<'i>>),
     /// Identity node to preserve the span in the source template text.
     AliasExpanded(AliasId<'i>, Box<ExpressionNode<'i>>),
 }
@@ -341,6 +345,17 @@ impl<'i> FoldableExpression<'i> for ExpressionKind<'i> {
                     body: folder.fold_expression(lambda.body)?,
                 });
                 Ok(Self::Lambda(lambda))
+            }
+            Self::Map(entries) => {
+                let entries = entries
+                    .into_iter()
+                    .map(|(key, value)| {
+                        let key = folder.fold_expression(key)?;
+                        let value = folder.fold_expression(value)?;
+                        Ok((key, value))
+                    })
+                    .try_collect()?;
+                Ok(ExpressionKind::Map(entries))
             }
             Self::AliasExpanded(id, subst) => {
                 let subst = Box::new(folder.fold_expression(*subst)?);
@@ -423,6 +438,8 @@ pub struct LambdaNode<'i> {
     pub body: ExpressionNode<'i>,
 }
 
+pub type MapNodeEntry<'i> = (ExpressionNode<'i>, ExpressionNode<'i>);
+
 fn parse_identifier_or_literal(pair: Pair<Rule>) -> ExpressionKind {
     assert!(matches!(
         pair.as_rule(),
@@ -474,6 +491,19 @@ fn parse_lambda_node(pair: Pair<Rule>) -> TemplateParseResult<LambdaNode> {
         params_span,
         body,
     })
+}
+
+fn parse_map_node(entries_pair: Pair<Rule>) -> TemplateParseResult<Vec<MapNodeEntry>> {
+    assert_eq!(entries_pair.as_rule(), Rule::map);
+    entries_pair
+        .into_inner()
+        .map(|pair| {
+            let [key_pair, value_pair] = pair.into_inner().collect_array().unwrap();
+            let key = parse_template_node(key_pair)?;
+            let value = parse_template_node(value_pair)?;
+            Ok((key, value))
+        })
+        .try_collect()
 }
 
 fn parse_string_literal(pair: Pair<Rule>) -> String {
@@ -531,6 +561,7 @@ fn parse_term_node(pair: Pair<Rule>) -> TemplateParseResult<ExpressionNode> {
             let lambda = Box::new(parse_lambda_node(expr)?);
             ExpressionKind::Lambda(lambda)
         }
+        Rule::map => ExpressionKind::Map(parse_map_node(expr)?),
         // Ignore inner span to preserve parenthesized expression as such.
         Rule::template => parse_template_node(expr)?.kind,
         other => panic!("unexpected term: {other:?}"),
@@ -938,6 +969,13 @@ mod tests {
                 });
                 ExpressionKind::Lambda(lambda)
             }
+            ExpressionKind::Map(entries) => {
+                let entries = entries
+                    .into_iter()
+                    .map(|(key, value)| (normalize_tree(key), normalize_tree(value)))
+                    .collect();
+                ExpressionKind::Map(entries)
+            }
             ExpressionKind::AliasExpanded(_, subst) => normalize_tree(*subst).kind,
         };
         ExpressionNode {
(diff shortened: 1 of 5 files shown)


ilyagr:
We could maybe have a helper

```
json_object("commitid", commit_id, "author", author)
```

that calls `json()` on every second value?

I might be missing some corner case, but I think it might be possible to define it in our template language, if it allows variadic arguments.

This is assuming you have no other plans for the mapping type.

-----

I guess this is assuming we don't care about creating nested objects. I don't think I do.

Alternatively, we could also let the caller call `json()` or `json_object()`, which is a little dangerous.

Or we can call `json()` if the argument is not the output of `json()` or `json_object`, similarly to how templating libraries deal with html that no longer needs escaping. Then, instead of a mapping type, we'd have yet another string type, e.g. JSONString, that serializes to itself.

We could also not support nested objects for now, and add JSONString later if needed.

ilyagr:
If you think having a mapping type is worth it, I'm ok with that as well.

yuja:
> We could maybe have a helper
> 
> ```
> json_object("commitid", commit_id, "author", author)
> ```

It's basically `make_map(key=value, ...)` described in the commit message.

> Similar to the lambda construct, the map literal cannot be used as a first-class
> template expression. That's the main reason why I decided to add a literal
> syntax instead of make_map(key=value, ..) function. A template function has to
> return a property of a certain type, but the templater doesn't support generics
> or heterogeneous containers. OTOH, a literal expression can be interpreted
> whatever the surrounding function expects. A downside is that this makes the
> language bigger.

Instead of a generic `make_map()`, we can add `serialize_map(key=value, ...) -> Serialize` function, which can be used as `json(serialize_map(...))`. `json_object()` could be a thin wrapper for that. We can also add `serialize(x).json()` and `serialize_map(key=value, ..).json()` (and deprecate `json(x)`) if method syntax looks better.

Implementation complexity wouldn't be much different.

ilyagr:
I'm not sure I follow your reasoning about `make_map`, though I also don't think this is blocking for this PR.

In my mind, `json_object` would not return a mapping type, but rather the same thing as the `json()` function returns: either `String` or a new `JSONString`. I'm not sure whether this makes a difference to your argument, but I'm guessing `make_map` would have returned a mapping type?

While `json_object` seems superficially simpler to me, I wouldn't be surprised if there were some subtler downsides or complexities to it. So, if your intuition prefers the mapping type, I'm happy to go with that.

We probably should wait until after the release for this. I'd be happy if somebody else reviewed this, but I'd hopefully get to it eventually myself as well.

yuja:
> In my mind, `json_object` would not return a mapping type, but rather the same thing as the `json()` function returns: either `String` or a new `JSONString`.

I think your `JSONString` is `Serialize`. (Well, `JSONString` would be more restricted, but there should be no reason to add `JSONString` type to disable re-serialization of JSON-encoded string because we already have `Serialize` type.)

> I'm not sure whether this makes a difference to your argument, but I'm guessing `make_map` would have returned a mapping type?

Yes. If the function is named `make_map`, it should be able to handle generic key-value pairs.

> While `json_object` seems superficially simpler to me,

A potential problem is that we have to add `<serialization-format>_<object-type>()` functions for each serialization format. I don't think we'll add many serialization functions, but we might have to add a few more because JSON is really bad at handling arbitrary bytes.

martinvonz:
@ilyagr: Are you going to finish the review of this PR or do you need me to take it over?

ilyagr:
Feel free to take over. Thanks!

martinvonz:
> Feel free to take over. Thanks!

To clarify, it's not that I *want to* take it over :) I just meant to ask if you plan on getting back to it. If you do, I prefer to let you take it.

ilyagr:
I would appreciate it if you or somebody took this over.

Also, I realized I never acknowledged Yuya's reply. Sorry about that!

> A potential problem is that we have to add <serialization-format>_<object-type>() functions for each serialization format. I don't think we'll add many serialization functions, but we might have to add a few more because JSON is really bad at handling arbitrary bytes.

Yes, this PR makes more sense in this light. I'm a bit curious about the details of what you have in mind (do we want to put file contents in JSON? what format would be better?), but chances are it's a fine plan and I might not be the right person to evaluate it. Serialization is a subject with much know-how I don't have.

PhilipMetzger:
> > A potential problem is that we have to add _() functions for each serialization format. I don't think we'll add many serialization functions, but we might have to add a few more because JSON is really bad at handling arbitrary bytes.
> 
> Yes, this PR makes more sense in this light. I'm a bit curious about the details of what you have in mind (do we want to put file contents in JSON? what format would be better?), but chances are it's a fine plan and I might not be the right person to evaluate it. Serialization is a subject with much know-how I don't have.

It's probably the conversation from Austin about supporting two or three formats beside the ubiquitous JSON which [started here][dc] in the Discord.

Also in my opinion this approach is fine to land as is, it gives some flexibility to users which @bryceberger proposed. 

[dc]:  https://discord.com/channels/968932220549103686/968932220549103689/1378474266995195935

yuja:
> do we want to put file contents in JSON? what format would be better?

I don't have any format in particular, but yes, templater supports diffs, so the file content may be embedded.

Just to be clear, I don't have strong feeling about the syntax, `serialize_map(key=value, ..).json()` vs `json({'key' => value, ..})`. The former is a bit restricted and verbose, but it doesn't make the language bigger.

davidbarsky:
As requested, I want this feature, to the point that I opened a separate pull request: https://github.com/jj-vcs/jj/pull/8507.

yuja:
Rebased the patches. I'll add changelog entry (after the release) if we agree on the syntax. I don't feel strongly whether we should add `json({"k" => v})` or something like `serialize_map(k=v).json()`.

davidbarsky:
> I don't feel strongly whether we should add json({"k" => v}) or something like serialize_map(k=v).json().

I don't feel strongly either (I prefer the literal...), but I'd also like to have arrays as well.

To centralize discussion here:
> BTW, if we add literal syntax, it should be usable for other purposes than constructing JSON object/array.

Three notes:
1. I thought of map/list literals to be immediately useful non-JSON contexts.
2. I didn't realize that templates had a different syntax than revsets and that templates didn't define a `"key": object`!
3. I'd like to make use of pest's spans for error reporting, as I did in: https://github.com/jj-vcs/jj/pull/8507. Should I rebase my changes atop of this change, or...?

yuja:
> 1. I thought of map/list literals to be immediately useful non-JSON contexts.

Yes, that's why a map literal should construct `Map<K, V>` or something than a type-erased object (if the literal has to converted to a template value.) This isn't possible with the current templater, though.

> 2. I didn't realize that templates had a different syntax than revsets and that templates didn't define a `"key": object`!

I don't think I follow. Can you elaborate? It's true that revsets and filesets are quite similar, but templates are different in that expressions are typed, etc.

> 3. I'd like to make use of pest's spans for error reporting, as I did in: [templater: add object and array literal syntax for `json()` #8507](https://github.com/jj-vcs/jj/pull/8507). Should I rebase my changes atop of this change, or...?

Which kind of errors? I think parsing errors should already have spans. If we decide to go with map literals, and if we think we also need lists, I think your PR can be rebased on top.

cli/src/template_parser.rs
```diff
@@ -346,7 +350,18 @@ impl<'i> FoldableExpression<'i> for ExpressionKind<'i> {
...
+                    .into_iter()
+                    .map(|(key, value)| {
+                        let key = folder.fold_expression(key)?;
+                        let value = folder.fold_expression(value)?;
+                        Ok((key, value))
+                    })
+                    .try_collect()?;
+                Ok(ExpressionKind::Map(entries))
+            }
+            ExpressionKind::AliasExpanded(id, subst) => {
```
higgsd: Mixture of Self:: and ExpressionKind:: here

yuja: Fixed.

abrenneke:
I'm very much in favor of a literal syntax as implemented in #8507! It feels very intuitive to me to build objects using `{ "key": value }` and arrays using `[]`. If these objects were usable outside of `json()` then even better! Are there any downsides to the literal syntax over `json_object` etc, besides the extra complexity of the additional syntax parsing? Any weird edge cases?

yuja:
> I'm very much in favor of a literal syntax as implemented in #8507! It feels very intuitive to me to build objects using `{ "key": value }` and arrays using `[]`.

As I mentioned in the commit message, `key: value` syntax conflicts with `kind:value` pattern, so we have to use different separator. This PR uses `key => value`.

PR #6876, opened 2025-07-01
hex_util: roll our own decode/encode_hex() functions

yuja:
# Checklist

If applicable:


thoughtpolice (approved):

lib/src/hex_util.rs
```diff
@@ -24,6 +34,17 @@ fn reverse_hex_value(b: u8) -> Option<u8> {
     }
 }
 
+/// Decodes `hex` as normal hex string.
```
ilyagr, line 37: (Optional): I might add a comment here similar to what you said in the commit id: "There might be better-optimized implementations in crates like `faster-hex`, but we avoid the dependency since we do not use these functions where performance matters".

emilazy:
FWIW, `gix` depends on `faster-hex` (and presumably always will depend on at least some hex implementation), so we could achieve high performance here while still dropping the `hex` dependency.

lib/src/hex_util.rs
yuja: I don't think the caller of this function should care about that. It's more like project-wide decision whether we need optimized hex function overall, and my conclusion is no.

If needed, our `decode`/`encode_hex()` can be forwarded to faster-hex.

ilyagr: I didn't mean a doc comment, I meant an implementation comment with two /s. It could be inside the function body. 

If there is a clear answer to Emily's question, you could put it here as well.

But this is still optional. It's mostly about whether somebody else changing these functions to use an external crate would be a problem or not. If that wouldn't be a big deal, there's less need for a comment.

ilyagr: To clarify, I previously assumed that it would be a big deal to you, and that assumption was probably wrong. In my eyes, it wouldn't be a big deal, and if somebody asked me, I'd probably send them to ask you instead.

yuja: Ah, yes. Implementation comment could be added, but it's not a big deal as you guessed. It's obvious that our decode/encode functions aren't super optimized, but they aren't dumb either.

Issue #4579, opened 2024-10-04
FR: revset for nth_parent(revision, n)


adampetrovic:
**Is your feature request related to a problem? Please describe.**
This feature request came about when trying to list all the changes that a merge commit introduced to the trunk of a repo I am working on.

**Describe the solution you'd like**
Essentially what I would like is, given a merge commit (i.e. two distinct parents), show me the list of changes that were introduced as a result of that merge (`jj log -r 'parent1..parent2'`).

Currently you can get the parents of a merge commit with the revset `parents(commit)`. This spits out 2 commits.

What I would like is to be able to specify which parent I want, something along the lines of `nth_parent(commit, 1)` for the first parent. Which mean the above could be simplified with `jj log -r 'nth_parent(commit, 1)..nth_parent(commit,2)'`

**Describe alternatives you've considered**
Alternative would be a bash alias that runs multiple `jj` commands to achieve the same thing. 

**Additional context**

jennings:
I know I hypothesized `nth_parents(rev, n)` in Discord, but I don't think I put enough thought into it.

Maybe it should be an option to `parents()` instead.

- `parents(rev, first-parent=true)`
- `parents(rev, n=1)`, though this implies you could do `n=2`, which should probably still give you the first parents of non-merge commits.

A common desire might be to see the first parent lineage, like `git log --first-parent main`. I can't think of a way to do this with `parents()`. So maybe there should be a first-parent option for `ancestors()` too/instead? For example: `ancestors(rev, first-parent=true)` might do the same thing as `git log --first-parent`.

yuja:
Yes, optional parameter is probably better, though it's a bit confusing whether `n` means `parents[n]` or `parents.parents....` (n times.)

btw, if we had `parents(rev, nth=N)`, I would expect `parents(rev, nth=2)` to be empty for non-merge commits.

adampetrovic:
What if revsets reported arbitrary indexing? 

For example:
`parents(commit)[0]`
`(..trunk())[1]`
etc

yuja:
Revset doesn't have a defined order, and `expr[n]` wouldn't work for `ancestors(..)` of first-parents.

scott2000:
I was planning to work on this now that `first_ancestors(x)` is implemented, since the implementation is mostly the same.

[...]

PR #6877, opened 2025-07-01
revset: add `first_ancestors()` function

### LINKED ISSUE #6606: FR: Add an equivalent to `--first-parent`
A project I work on uses (mostly) merge commits to merge branches into `main`, where the shape of the commits in the individual branches is largely up to the developer working on that branch.  When I create a new release, I want to get a list of merge commits and other commits that live directly on `main`.

Currently, in a collocated repo, I can do this with `git log <prev tag>..main --first-parent`.  This has the semantics: "When finding commits to include, follow only the first parent commit upon seeing a merge commit."  (from the [git docs](https://git-scm.com/docs/git-log#Documentation/git-log.txt---first-parent)).

As I understand it, there is currently no equivalent in Jujutsu.

I suggest a function like `first_parents(srcs, domain)`, which returns all commits in `domain` that can be reached by a commit in `srcs`, following first-parent connections only.

For example, given this history:
```
o F
|
o E
|\
| o D
| |
| o C
| |
o | B
|/
o A
|
o root()
```

then:

* `first_parents(F, A::)` would return `F | E | B | A`
* `first_parents(F, A..)` would return `F | E | B`
* `first_parents(D, A::)` would return `D | C | A` (because even though C is A's second child, A is C's first parent)
* `first_parents(F | D, A::)` would return `F | E | D | C | B | A`  (all commits are either on the first-commit branch or the branch that ends in D)
* `first_parents(E, A::)` would return `E | B | A` (even though `A::` includes F, F is not a parent of E)

* A function like `first_parent(X)` that returns only the first parent of `X`.  This might also be useful, but I don't think it's sufficient to get the same results as `--first-parent`, which explicitly *follows* those first-parent relationships recursively.
* This could be an extension of `reachable(srcs, domain)`, i.e. `reachable(srcs, domain, 1)` would be roughly the equivalent of the above.  However, `reachable` currently traverses all parent _and child_ edges, which may be a bit unintuitive when trying to figure out which commits are reachable via first-parent relationships only.  And `reachable(..., 1)` looks weird: what does the `1` refer to here?
* To make this slightly more generic, we could instead have a function like `nth_parents(srcs, domain[, N])`, which would follow the `N`th parent rather than the first parent.  The behaviour is difficult to specify for commits with a single parent though, and I suspect the only two interesting cases are "has a first-parent relationship" and "does not have a first-parent relationship", which are covered by `first_parents(...)` and the usual revset operations.
* This could be a CLI flag like in Git, but that feels like the wrong approach given the revset language exists.

scott2000:
Fixes #6606. Not sure if this is the best approach, but I wanted to try it out.

# Checklist

If applicable:

### DIFF
--- a/lib/src/default_index/rev_walk.rs
+++ b/lib/src/default_index/rev_walk.rs
@@ -30,6 +30,7 @@ use super::entry::GlobalCommitPosition;
 use super::entry::SmallGlobalCommitPositionsVec;
 use super::rev_walk_queue::RevWalkQueue;
 use super::rev_walk_queue::RevWalkWorkItem;
+use crate::revset::PARENTS_RANGE_FULL;
 
 /// Like `Iterator`, but doesn't borrow the `index` internally.
 pub(super) trait RevWalk<I: ?Sized> {
@@ -245,7 +246,7 @@ impl<I, W: RevWalk<I>> FusedIterator for RevWalkOwnedIndexIter<I, W> {}
 
 pub(super) trait RevWalkIndex {
     type Position: Copy + Ord;
-    type AdjacentPositions: IntoIterator<Item = Self::Position>;
+    type AdjacentPositions: IntoIterator<Item = Self::Position> + AsRef<[Self::Position]>;
 
     fn adjacent_positions(&self, pos: Self::Position) -> Self::AdjacentPositions;
 }
@@ -306,6 +307,7 @@ pub(super) struct RevWalkBuilder<'a> {
     index: &'a CompositeIndex,
     wanted: Vec<GlobalCommitPosition>,
     unwanted: Vec<GlobalCommitPosition>,
+    wanted_parents_range: Range<u32>,
 }
 
 impl<'a> RevWalkBuilder<'a> {
@@ -314,12 +316,20 @@ impl<'a> RevWalkBuilder<'a> {
             index,
             wanted: Vec::new(),
             unwanted: Vec::new(),
+            wanted_parents_range: PARENTS_RANGE_FULL,
         }
     }
 
     /// Sets head positions to be included.
     pub fn wanted_heads(mut self, positions: Vec<GlobalCommitPosition>) -> Self {
         self.wanted = positions;
+        self
+    }
+
+    /// Sets range of parents to iterate over for wanted heads.
+    pub fn wanted_parents_range(mut self, wanted_parents_range: Range<u32>) -> Self {
+        assert!(wanted_parents_range.start <= wanted_parents_range.end);
+        self.wanted_parents_range = wanted_parents_range;
         self
     }
 
@@ -336,6 +346,7 @@ impl<'a> RevWalkBuilder<'a> {
 
     fn ancestors_with_min_pos(self, min_pos: GlobalCommitPosition) -> RevWalkAncestors<'a> {
         let index = self.index;
+        let wanted_parents_range = self.wanted_parents_range;
         let mut wanted_queue = RevWalkQueue::with_min_pos(min_pos);
         let mut unwanted_queue = RevWalkQueue::with_min_pos(min_pos);
         wanted_queue.extend(self.wanted, ());
@@ -345,6 +356,7 @@ impl<'a> RevWalkBuilder<'a> {
             walk: RevWalkImpl {
                 wanted_queue,
                 unwanted_queue,
+                wanted_parents_range,
             },
         }
     }
@@ -357,6 +369,7 @@ impl<'a> RevWalkBuilder<'a> {
         generation_range: Range<u32>,
     ) -> RevWalkAncestorsGenerationRange<'a> {
         let index = self.index;
+        let wanted_parents_range = self.wanted_parents_range;
         let mut wanted_queue = RevWalkQueue::with_min_pos(GlobalCommitPosition::MIN);
         let mut unwanted_queue = RevWalkQueue::with_min_pos(GlobalCommitPosition::MIN);
         let item_range = RevWalkItemGenerationRange::from_filter_range(generation_range.clone());
@@ -367,6 +380,7 @@ impl<'a> RevWalkBuilder<'a> {
             walk: RevWalkGenerationRangeImpl {
                 wanted_queue,
                 unwanted_queue,
+                wanted_parents_range,
                 generation_end: generation_range.end,
             },
         }
@@ -443,6 +457,9 @@ impl<'a> RevWalkBuilder<'a> {
             walk: RevWalkGenerationRangeImpl {
                 wanted_queue,
                 unwanted_queue,
+                // Since we're using the descendants index, this actually is a range for the
+                // children instead of the parents.
+                wanted_parents_range: PARENTS_RANGE_FULL,
                 generation_end: generation_range.end,
             },
         }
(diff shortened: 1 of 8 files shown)


lib/src/default_index/rev_walk.rs
```diff
@@ -247,6 +248,17 @@ pub(super) trait RevWalkIndex {
     type AdjacentPositions: IntoIterator<Item = Self::Position>;
 
     fn adjacent_positions(&self, pos: Self::Position) -> Self::AdjacentPositions;
+
+    fn adjacent_positions_in_range(
```
scott2000: If we enforce that `RevsetExpression::Range` uses the same parents range for both the roots and the heads, then we could maybe use a custom `RevWalkIndex` implementation which overrides `adjacent_positions` to only return the parents in a certain range instead.

But I think `first_parents(x) & y..` would be more common than `first_parents(x) ~ first_parents(y)`, so I thought it might be better to leave the unwanted roots using the full set of parents so that this could be optimized to a `RevsetExpression::Range`.

yuja: It might be good to add separate `heads_parents_range` and `roots_parents_range`, but I agree we shouldn't use the same range for both heads and roots. The goal of `RevsetExpression::Range` is to omit unwanted ancestors efficiently, and the `roots` and `heads` are usually independent expressions.

thoughtpolice:
nit: I wonder if there's a better name than `first_parents`. I know this sounds (a little) ridiculous, but I somewhat detest compound words for "first-party" revsets, and I could also say the same of e.g. `fork_point`. It shouldn't block this but if anyone has any clever ideas, please post them!

yuja:
There are previous discussions in
https://github.com/jj-vcs/jj/issues/4579 and https://github.com/jj-vcs/jj/issues/6606. If anyone has naming suggestion, it's probably better to continue in either of these issues. FWIW, Mercurial calls it a `p1(x)`.

emilazy:
I think this actually implements `first_ancestors()`. The `first_parents()` we discussed is different. Doesn't solve the compound thing of course.

yuja:
> I think this actually implements `first_ancestors()`. The `first_parents()` we discussed is different. Doesn't solve the compound thing of course.

Good catch. `first_parents(heads, domain)` described originally in #6606 is equivalent to `ancestors(heads, follow_parents_range=0..1) & domain`, and this `first_parents(heads)` implements `ancestors(heads, follow_parents_range=0..1)`? I don't know if we'll need [this version](https://github.com/jj-vcs/jj/issues/6606#issuecomment-2901277705) because it can be expressed by combining more primitive expressions. For the same reason, I don't think we need the original `first_parents(heads, domain)`.

BTW, I misunderstood that this `first_parents(heads)` would compute `ancestors(heads, depth=1..2, follow_parents_range=0..1)` (i.e. `c.parents()[0] for c in heads`.)

scott2000:
I updated it to `first_ancestors(heads)` for now, since it seems like people find `first_parents(heads)` to be a confusing name.

scott2000:
I think this should be ready for review now. I left it as `first_ancestors()` since that seems to be the most preferred option currently.

lib/src/default_index/rev_walk.rs
```diff
@@ -248,6 +249,17 @@ pub(super) trait RevWalkIndex {
...
```
yuja: nit: can you extract this to free function? I don't think trait method is needed.

BTW, I think it's simpler to use slice than `.skip()` and `.take()`.

lib/src/default_index/rev_walk.rs
```diff
@@ -248,6 +249,17 @@ pub(super) trait RevWalkIndex {
...
+        &self,
+        pos: Self::Position,
+        parents_range: &Range<u32>,
+    ) -> impl IntoIterator<Item = Self::Position> {
+        self.adjacent_positions(pos)
+            .into_iter()
+            .skip(parents_range.start as usize)
+            .take(parents_range.end.saturating_sub(parents_range.start) as usize)
```
yuja: nit: maybe better to normalize the range out of the loop.

lib/src/default_index/rev_walk.rs
```diff
@@ -399,6 +423,10 @@ impl<'a> RevWalkBuilder<'a> {
         self,
         root_positions: HashSet<GlobalCommitPosition>,
     ) -> RevWalkDescendants<'a> {
+        assert_eq!(
+            self.wanted_parents_range, PARENTS_RANGE_FULL,
+            "cannot use parents range with descendants"
+        );
```
yuja: nit: I think the output would be correct in that descendants reachable by following the filtered parents range are returned, but I agree it's useless.

lib/src/revset.rs
```diff
@@ -2213,9 +2252,11 @@ fn fold_generation<St: ExpressionState>(
                 RevsetExpression::Ancestors {
                     heads,
                     generation: generation2,
-                } => Some(Rc::new(RevsetExpression::Ancestors {
+                    parents_range: parents2,
+                } if parents2 == parents1 => Some(Rc::new(RevsetExpression::Ancestors {
```
yuja, line 2300: Can you update optimization test to include edge cases like this?

yuja (approved):
LGTM, thanks.

lib/src/default_index/rev_walk.rs
scott2000: After switching to a slice, I don't think this is applicable anymore.

lib/src/default_index/rev_walk.rs
scott2000: Yeah, that's true. I removed the assertion since it doesn't seem like an actual error, just something that isn't useful.

lib/src/default_index/rev_walk.rs
```diff
@@ -245,7 +247,7 @@ impl<I, W: RevWalk<I>> FusedIterator for RevWalkOwnedIndexIter<I, W> {}
 
 pub(super) trait RevWalkIndex {
     type Position: Copy + Ord;
-    type AdjacentPositions: IntoIterator<Item = Self::Position>;
+    type AdjacentPositions: IntoIterator<Item = Self::Position> + Deref<Target = [Self::Position]>;
```
scott2000: @yuja do you think it's fine to add this `Deref` bound? It makes it so we can use slice operations without having to do `positions.into_iter().collect()` unnecessarily.

yuja: I didn't notice the return type was abstract. As a trait bound, `AsRef<[Self::Position]>` should be better.

lib/src/default_index/rev_walk.rs
```diff
@@ -643,6 +668,16 @@ impl RevWalk<CompositeIndex> for RevWalkDescendantsImpl {
     }
 }
 
+/// Filter a slice by a range, clamping the range to the length of the slice.
+fn filter_slice_by_range<'a, T: Copy>(
+    slice: &'a [T],
+    range: &Range<u32>,
+) -> impl Iterator<Item = T> + use<'a, T> {
```
yuja: nit: simply return `&'a [T]`?

lib/src/default_index/rev_walk.rs
```diff
@@ -643,6 +668,16 @@ impl RevWalk<CompositeIndex> for RevWalkDescendantsImpl {
...
+    let start = (range.start as usize).min(slice.len());
+    let end = (range.end as usize).min(slice.len());
+    slice[start..end].iter().copied()
```
yuja: nit: we'll still need to handle `start > end` somewhere, maybe in `RevWalkBuilder`.

PR #6879, opened 2025-07-02
releasing docs: fixup command, add dry-run publishing command

ilyagr:


docs/releasing.md
```diff
@@ -21,7 +21,7 @@ Producing the list of contributors is a bit annoying. The current suggestion is
 to run something like this:
 
 ```shell
-root=$(jj log --no-graph -r 'heads(tags(glob:"v*.*.*") & ::trunk())' -T commit_id)
+root=$(jj log --no-graph --color=never -r 'heads(tags(glob:"v*.*.*") & ::trunk())' -T commit_id)
```
martinvonz, line 24: Do you have `ui.color = "always"`? Why? I'm just curious.

(no objection to this change)

ilyagr: I do have `ui.color = "always"`. Initially, this was about running `watch jj log` or `hwatch jj log`. Also, it could be nice for testing out different pagers.

Now, it's less useful, but reminds me to file a bug report to every tool invoking `jj` (`jjk`, `jjui`, etc) if they forget to add `--color=never` to their invocations.

This also reminds me that we need something like Mercurial's `--plain` flag (or is it just `HGPLAIN` env var), but it's a pain. Perhaps it could be a `--tooling` flag that means `--color=never --no-pager`, but perhaps people will sometimes want tooling to get the colored output. So, should we allow `--tooling --color=always`? Should it be an environment variable instead? Is it even worth it for just two flags?

I can also imagine a world where `--tooling` implies `--no-commit-transaction`, and tools would explicitly `jj snapshot`. But would all tools want this? We'd then also probably want a `--commit-transaction` flag.

I can see how Mercurial ended up with `HGPLAINEXCEPT` environment variable. I keep hoping for something better, but maybe that's just the best we can have.

PR #6882, opened 2025-07-02
Make diff --stat fill screen width

evmar:
(Two commits, can be reviewed separately; pre-refactoring before I make the larger change of #6866)

Using jj 0.30.0 and a narrow terminal, see how "diffstat" is slightly wider (filling my terminal).
After this change jj matches diffstat.

image

# Checklist

If applicable:

(Not sure if this belongs in CHANGELOG.md, let me know)

### DIFF
--- a/cli/src/diff_util.rs
+++ b/cli/src/diff_util.rs
@@ -2025,6 +2025,13 @@ pub fn show_diff_stats(
             }
         })
         .collect_vec();
+
+    // Entries format like:
+    //   path/to/file | 123 ++--
+    //
+    // Depending on display widths, we can elide part of the path,
+    // and the the ++-- bar will adjust its scale to fill the rest.
+
     let max_path_width = ui_paths.iter().map(|s| s.width()).max().unwrap_or(0);
     let max_diffs = stats
         .entries()
@@ -2033,12 +2040,13 @@ pub fn show_diff_stats(
         .max()
         .unwrap_or(0);
 
-    let number_padding = max_diffs.to_string().len();
-    // 4 characters padding for the graph
-    let available_width = display_width.saturating_sub(4 + " | ".len() + number_padding);
-    // Always give at least a tiny bit of room
+    let number_width = max_diffs.to_string().len();
+    // Available width to distribute between path and bar:
+    let available_width = display_width.saturating_sub(" | ".len() + number_width + " ".len());
+    // Always use at least a tiny bit of room
     let available_width = max(available_width, 5);
-    let max_path_width = max_path_width.clamp(3, (0.7 * available_width as f64) as usize);
+
+    let max_path_width = max_path_width.min((0.7 * available_width as f64) as usize);
     let max_bar_length = available_width.saturating_sub(max_path_width);
     let factor = if max_diffs < max_bar_length {
         1.0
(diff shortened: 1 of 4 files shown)


cli/src/diff_util.rs
```diff
@@ -2033,12 +2040,13 @@ pub fn show_diff_stats(
         .max()
         .unwrap_or(0);
 
-    let number_padding = max_diffs.to_string().len();
-    // 4 characters padding for the graph
-    let available_width = display_width.saturating_sub(4 + " | ".len() + number_padding);
```
yuja, line 2038: I think the intent of `4` is to allocate at least 4 characters for the graph drawings.

BTW, the original PR had 1-space indent as `git` would insert, so I wouldn't be surprised if there was off-by-one error.

cli/src/diff_util.rs
```diff
@@ -2047,14 +2055,14 @@ pub fn show_diff_stats(
     };
 
     for (stat, ui_path) in iter::zip(stats.entries(), &ui_paths) {
-        let bar_added = (stat.added as f64 * factor).ceil() as usize;
-        let bar_removed = (stat.removed as f64 * factor).ceil() as usize;
+        let bar_added = (stat.added as f64 * factor).round() as usize;
+        let bar_removed = (stat.removed as f64 * factor).round() as usize;
```
yuja: Maybe we should use `.floor()`? If both added/removed numbers are rounded up, I think the bar width exceeds the allocated space.

cli/src/diff_util.rs
evmar: I stared at this 4 for a while and couldn't quite figure it out, because from the code the `available_width` number is distributed between the path and the graph.  This is why I rewrote the comment in this PR.

The ratio in the code is 70% for path and 30% for graph.  When the terminal size gets very narrow the behavior can get a bit weird (where the path gets totally elided, or the graph is only one character wide), but in those cases the terminal is too narrow to even display a file name, so I think the only behavior we really care about is that jj doesn't crash.

cli/src/diff_util.rs
evmar: That is a really good catch!  Unfortunately using floor mean whenever the values aren't integers, the bar will be end up one character too short.  I tinkered a bit and I think another reason to use .ceil() is so that you always get at least a single `+` or `-` if there are any adds or removes.

Example output (which also fills the terminal width):
```
$ git diff --stat
 lines | 400 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-
 1 file changed, 399 insertions(+), 1 deletion(-)
```
note the final `-` character even though there are 399 adds.

So in this last update, I fixed this, and also added a test for this behavior.

(This may seem like too much of a detail, but I noticed it pretty quickly in tinkering with this change, because otherwise the diffstat makes it look like the file has only adds or removes.)

evmar:
(I think this is ready for another review, no rush)

cli/src/diff_util.rs
```diff
@@ -2047,14 +2055,22 @@ pub fn show_diff_stats(
     };
 
     for (stat, ui_path) in iter::zip(stats.entries(), &ui_paths) {
-        let bar_added = (stat.added as f64 * factor).ceil() as usize;
-        let bar_removed = (stat.removed as f64 * factor).ceil() as usize;
+        let bar_length = ((stat.added + stat.removed) as f64 * factor) as usize;
+        // Ensure that any fractional space after scaling is given to whichever
+        // of adds/removes is smaller.
```
yuja: nit: can you add comment why smaller side is weighed? I think that's in order to print at least one character.

evmar: Yes, done.

evmar:
Friendly ping, not sure what I need to do from here!

yuja:
Oh, appears that you don't have a permission to merge the PR. Sorry.

cc @martinvonz

martinvonz:
Sorry, I thought I had already invited you. You should have an invitation now.

PR #6883, opened 2025-07-02
Make `MergedTree::diff_stream()` sort `foo` before `foo/bar`

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -204,6 +204,9 @@ impl MergedTree {
             }
             Ok(_) => Ok(None),
             Err(merge) => {
+                if !merge.is_tree() {
+                    return Ok(None);
+                }
                 let trees = merge
                     .try_map_async(|value| async move {
                         match value {
@@ -211,7 +214,8 @@ impl MergedTree {
                                 let subdir = self.dir().join(name);
                                 self.store().get_tree_async(subdir, sub_tree_id).await
                             }
-                            _ => {
+                            Some(_) => unreachable!(),
+                            None => {
                                 let subdir = self.dir().join(name);
                                 Ok(Tree::empty(self.store().clone(), subdir))
                             }
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -1039,10 +1039,10 @@ impl<'matcher> TreeDiffStreamImpl<'matcher> {
                 let (before, after) = self.items.remove(&key).unwrap().unwrap();
                 // If this was a transition from file to tree or vice versa, add back an item
                 // for just the removal/addition of the file.
-                if before.is_present() && !before.is_tree() {
+                if before.is_file_like() {
                     self.items
                         .insert(key.clone(), Ok((before, Merge::absent())));
-                } else if after.is_present() && !after.is_tree() {
+                } else if after.is_file_like() {
                     self.items.insert(
                         DiffStreamKey::file_after_dir(key.path.clone()),
                         Ok((Merge::absent(), after)),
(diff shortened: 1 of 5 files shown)


lib/src/merged_tree.rs
```diff
@@ -204,6 +204,9 @@ impl MergedTree {
             }
             Ok(_) => Ok(None),
             Err(merge) => {
+                if !merge.is_tree() {
+                    return Ok(None);
+                }
                 let trees = merge
                     .try_map_async(|value| async move {
                         match value {
```
yuja, line 213: nit: insert `Some(_) => unreachable!()` to this match to clarify that the value should be either `Some(Tree(_))` or `None`?

lib/src/merged_tree.rs
```diff
@@ -692,24 +692,19 @@ impl Iterator for ConflictIterator {
...
-    Dir(TreeDiffDirItem),
-    // This is used for making sure that when a directory gets replaced by a file, we
-    // yield the value for the addition of the file after we yield the values
-    // for removing files in the directory.
-    File(RepoPathBuf, MergedTreeValue, MergedTreeValue),
-}
-
 impl<'matcher> TreeDiffIterator<'matcher> {
     /// Creates a iterator over the differences between two trees. Generally
     /// prefer `MergedTree::diff()` of calling this directly.
```
yuja: nit: remove or update doc about `MergedTree::diff()` since `TreeDiffIterator::new()` is no longer identical to `MergedTree::diff_stream()`.

lib/src/merged_tree.rs
```diff
@@ -1074,41 +980,105 @@ impl Stream for TreeDiffStreamImpl<'_> {
...
+
+impl<'a> Stream for DiffStreamForFileSystem<'a> {
+    type Item = TreeDiffEntry;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        loop {
+            if let Some(next) = self.next_item.take() {
+                // If there's a held file "foo" and the next item to emit is not "foo/...", then
+                // we must be done with the "foo/" directory and it's time to emit "foo" as a
+                // removed file.
```
yuja: I think it's easier to follow if `poll_next()` is attempted here (than temporarily queuing to `self.next_item`.)

```rust
while let Some(next) = match self.next_item.take() {
    Some(next) => Some(next),
    None => ready!(self.inner.as_mut().poll_next(cx)),
} {
    ...
}
Poll::Ready(self.held_file.take())
```

lib/src/merged_tree.rs
```diff
@@ -828,27 +807,10 @@ impl Iterator for TreeDiffIterator<'_> {
...
-                }
-            } else if tree_before && !tree_after {
-                if after.is_present() {
-                    self.stack.insert(
-                        post_subdir,
-                        TreeDiffItem::File(path, Merge::absent(), after),
-                    );
-                }
-            } else if !tree_before && !tree_after {
+            if before.is_file_like() || after.is_file_like() {
```
yuja, line 835: (No action requested)

It might be better to let caller filter tree->tree diff entries. We might want to add `jj diff --name-only --include-directory` or something for completion of path prefixes.

yuja (approved):

lib/src/merged_tree.rs
martinvonz: Good idea. I also replaced the `_` match by `None`.

lib/src/merged_tree.rs
martinvonz: True

lib/src/merged_tree.rs
martinvonz: Good idea. Done.

lib/src/merged_tree.rs
martinvonz: Done (removed).

PR #6888, opened 2025-07-02
ui: make help text color configurable

gechr:
So I started working on this as exercise to test out Agentic LLM workflows and their capabilities (1) in a language I've very little practical experience with and (2) on a codebase that I'm entirely unfamiliar with.

I originally had no intention of actually submitting this, and am happy to simply close it if the `jj` maintainers don't want to accept it in its current form, but it seems to work well in my local testing, and resolves https://github.com/jj-vcs/jj/discussions/5917, so I figured I'd raise it and let the maintainers decide.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### New features
 
+* Help text colors are now configurable. You can customize the colors used in
+  `jj help` output by setting `colors."help heading"`, `colors."help usage"`,
+  `colors."help literal"`, and `colors."help placeholder"` in your config.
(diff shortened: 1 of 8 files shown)


martinvonz:
I haven't looked at the code in detail but I suspect that if we want to add this feature, we should do it by taking over the rendering of the help text from clap. I think we have previously decided that we'll need to do that for other features anyway (e.g. #2866).

gechr:
Aha, that seems more sensible, though it sounds like a bigger piece of work than I had intended. In fact, I didn't intend for this PR to end up so long when I started fiddling around in the first place.

I'm happy to close this until that prerequisite work is complete by someone more familiar with the internals, and then take a stab at it again at a later date.

Let me know!

martinvonz:
Yes, I would prefer to close this for now. Thanks, though.

PR #6897, opened 2025-07-04
docs: document Commit.signature()

jakalx:
# Checklist

If applicable:

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -125,7 +125,8 @@ This type cannot be printed. The following methods are defined.
 * `.parents() -> List<Commit>`
 * `.author() -> Signature`
 * `.committer() -> Signature`
-* `.signature() -> Option<CryptographicSignature>`
+* `.signature() -> Option<CryptographicSignature>`: Cryptographic signature if the
+  commit was signed.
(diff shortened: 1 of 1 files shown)


docs/templates.md
```diff
@@ -214,7 +214,7 @@ This type can be printed in TOML syntax. The following methods are defined.
 
 ### `CryptographicSignature` type
 
-_Conversion: `Boolean`: no, `Serialize`: no, `Template`: no_
+_Conversion: `Boolean`: yes, `Serialize`: no, `Template`: no_
```
yuja: This type itself isn't convertible to bool. Maybe we can improve the documentation of `Commit.signature() -> Option<CryptographicSignature>` method instead, which is currently empty?

jakalx: Thanks for clarifying this. I'll think about a better way to document this.

Turns out it is an error to call any of those functions if not checking the presence first - which makes perfect sense.

jakalx: @yuja I've updated the PR and documented just the signature function. What do you think?

yuja (approved):
Thanks.

Issue #6893, opened 2025-07-03
FR: Regex/parsing support in string templates


lf-:
**Is your feature request related to a problem? Please describe.**
At work I use commit messages that start with some identifier like `DUX-1234:`. I want to be able to extract these from the commit messages for use in a branch name template like shown here: https://github.com/jj-vcs/jj/discussions/6891

Currently it seems that there's not any string functions that would allow for checking if a string matches some pattern.

**Describe the solution you'd like**
If regex is acceptable, I would like a function to the effect of `matches(message, '^([A-Z]+-[0-9]+):', |match| match.groups().at(0))`

**Describe alternatives you've considered**
I guess I could somehow call an external process from a template, but that's not ideal and is sort of overkill.

**Additional context**

yuja:
Maybe we can add support for string patterns (e.g. `s.matches(regex:"...")`, `s.replace("foo", "bar")`, etc.), though some of them wouldn't be useful.
https://jj-vcs.github.io/jj/latest/revsets/#string-patterns

lf-:
That would be really cute and would make the two languages more similar!

lf-:
I'm going to try to implement this and see how far I get.

joneshf:
It looks like progress has slowed here a bit–which is totally fine, nobody is required to do any of this work, least of all on a deadline! Is there anything someone like myself can do to help move things along, and hopefully get more string support in the next release (0.33.0, I assume)? In particular, I'm looking for a [`replacen` function](https://github.com/jj-vcs/jj/pull/6998#issuecomment-3132589645) (or something equivalent).

lf-:
I'm on vacation and prior to being on vacation, I was working on preparing for vacation. This doesn't require significant work and I think I've landed the especially conflicty parts of the stack so I just need to do a rebase now.

lf-:
since https://github.com/jj-vcs/jj/pull/7467 landed, i think this is officially done! proud of our work on getting this corner of the template language improved together.

PR #6899, opened 2025-07-05
templates: support string patterns in template language

lf-:
This is a basic implementation of the same string pattern system as in
the revset language. It's currently only used for `string.matches`, so
you can now do:

```
"foo".matches(regex:'[a-f]o+')
```

In the future this could be added to more string functions (and e.g.
the ability to parse things out of strings could be added).

CC: https://github.com/jj-vcs/jj/issues/6893

# Checklist

If applicable:

### DIFF
--- a/cli/src/template_builder.rs
+++ b/cli/src/template_builder.rs
@@ -910,6 +910,25 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>()
         },
     );
     map.insert(
+        "match",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            let [needle_node] = function.expect_exact_arguments()?;
+            let needle = template_parser::expect_string_pattern(needle_node)?;
+            let regex = needle.to_regex();
+
+            let out_property = self_property.and_then(move |haystack| {
+                if let Some(m) = regex.find(haystack.as_bytes()) {
+                    Ok(std::str::from_utf8(m.as_bytes())?.to_owned())
+                } else {
+                    // We don't have optional strings, so empty string is the
+                    // right null value.
+                    Ok(String::new())
+                }
+            });
+            Ok(out_property.into_dyn_wrapped())
+        },
+    );
+    map.insert(
         "starts_with",
         |language, diagnostics, build_ctx, self_property, function| {
             let [needle_node] = function.expect_exact_arguments()?;
@@ -2989,6 +3008,32 @@ mod tests {
             env.render_ok(r#""foo".contains("f" ++ bad_string) ++ "bar""#), @"<Error: Bad>bar");
         insta::assert_snapshot!(
             env.render_ok(r#""foo".contains(separate("o", "f", bad_string))"#), @"<Error: Bad>");
+
+        insta::assert_snapshot!(env.render_ok(r#""fooo".match(regex:'[a-f]o+')"#), @"fooo");
+        insta::assert_snapshot!(env.render_ok(r#""fa".match(regex:'[a-f]o+')"#), @"");
+        insta::assert_snapshot!(env.render_ok(r#""hello".match(regex:"h(ell)o")"#), @"hello");
+        insta::assert_snapshot!(env.render_ok(r#""HEllo".match(regex-i:"h(ell)o")"#), @"HEllo");
+        insta::assert_snapshot!(env.render_ok(r#""hEllo".match(glob:"h*o")"#), @"hEllo");
+        insta::assert_snapshot!(env.render_ok(r#""Hello".match(glob:"h*o")"#), @"");
+        insta::assert_snapshot!(env.render_ok(r#""HEllo".match(glob-i:"h*o")"#), @"HEllo");
+        insta::assert_snapshot!(env.render_ok(r#""hello".match("he")"#), @"he");
+        insta::assert_snapshot!(env.render_ok(r#""hello".match(substring:"he")"#), @"he");
+        insta::assert_snapshot!(env.render_ok(r#""hello".match(exact:"he")"#), @"");
+
+        // Evil regexes can cause invalid UTF-8 output, which nothing can
+        // really be done about given we're matching against non-UTF-8 stuff a
+        // lot as well.
+        insta::assert_snapshot!(env.render_ok(r#""🥺".match(regex:'(?-u)^(?:.)')"#), @"<Error: incomplete utf-8 byte sequence from index 0>");
+
+        insta::assert_snapshot!(env.parse_err(r#""🥺".match(not-a-pattern:"abc")"#), @r#"
+         --> 1:11
+          |
+        1 | "🥺".match(not-a-pattern:"abc")
+          |           ^-----------------^
+          |
+          = Bad string pattern
+        Invalid string pattern kind `not-a-pattern:`
+        "#);
 
         insta::assert_snapshot!(env.render_ok(r#""".first_line()"#), @"");
         insta::assert_snapshot!(env.render_ok(r#""foo\nbar".first_line()"#), @"foo");
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -426,6 +426,10 @@ defined.
 * `.len() -> Integer`: Length in UTF-8 bytes.
 * `.contains(needle: Stringify) -> Boolean`: Whether the string contains the
   provided stringifiable value as a substring.
+* `.match(needle: StringPattern) -> String`: Extracts
+  the first matching part of the string for the given pattern.
+
+  An empty string is returned if there is no match.
 * `.first_line() -> String`
 * `.lines() -> List<String>`: Split into lines excluding newline characters.
 * `.upper() -> String`
--- a/lib/src/str_util.rs
+++ b/lib/src/str_util.rs
@@ -60,6 +60,11 @@ impl GlobPattern {
     pub fn as_str(&self) -> &str {
         self.glob.glob()
     }
+
+    /// Converts this glob pattern to a bytes regex.
+    pub fn to_regex(&self) -> regex::bytes::Regex {
+        self.regex.clone()
+    }
 }
 
 impl Debug for GlobPattern {
@@ -289,6 +294,35 @@ impl StringPattern {
         }
     }
 
+    /// Converts the pattern into a bytes regex.
+    pub fn to_regex(&self) -> regex::bytes::Regex {
+        match self {
+            Self::Exact(literal) => {
+                regex::bytes::RegexBuilder::new(&format!("^{}$", regex::escape(literal)))
+                    .build()
+                    .expect("impossible to fail to compile regex of literal")
+            }
+            Self::ExactI(literal) => {
+                regex::bytes::RegexBuilder::new(&format!("^{}$", regex::escape(literal)))
+                    .case_insensitive(true)
+                    .build()
+                    .expect("impossible to fail to compile regex of literal")
+            }
+            Self::Substring(literal) => regex::bytes::RegexBuilder::new(&regex::escape(literal))
+                .build()
+                .expect("impossible to fail to compile regex of literal"),
+            Self::SubstringI(literal) => regex::bytes::RegexBuilder::new(&regex::escape(literal))
+                .case_insensitive(true)
+                .build()
+                .expect("impossible to fail to compile regex of literal"),
+            Self::Glob(glob_pattern) => glob_pattern.to_regex(),
+            // The regex generated represents the case insensitivity itself
+            Self::GlobI(glob_pattern) => glob_pattern.to_regex(),
+            Self::Regex(regex) => regex.clone(),
+            Self::RegexI(regex) => regex.clone(),
+        }
+    }
+
     /// Iterates entries of the given `map` whose string keys match this
     /// pattern.
     pub fn filter_btree_map<'a, 'b, K: Borrow<str> + Ord, V>(
@@ -486,5 +520,26 @@ mod tests {
                 .unwrap()
                 .is_match("\u{c0}")
         );
+    }
+
+    #[test]
+    fn test_string_pattern_to_regex() {
+        let check = |pattern: StringPattern, match_to: &str| {
+            let regex = pattern.to_regex();
+            regex.is_match(match_to.as_bytes())
+        };
+        assert!(check(StringPattern::exact("$a"), "$a"));
+        assert!(!check(StringPattern::exact("$a"), "$A"));
+        assert!(!check(StringPattern::exact("a"), "aa"));
+        assert!(!check(StringPattern::exact("a"), "aa"));
+        assert!(check(StringPattern::exact_i("a"), "A"));
+        assert!(check(StringPattern::substring("$a"), "$abc"));
+        assert!(!check(StringPattern::substring("$a"), "$Abc"));
+        assert!(check(StringPattern::substring_i("$a"), "$Abc"));
+        assert!(!check(StringPattern::glob("a").unwrap(), "A"));
+        assert!(check(StringPattern::glob_i("a").unwrap(), "A"));
+        assert!(check(StringPattern::regex("^a{1,3}").unwrap(), "abcde"));
+        assert!(!check(StringPattern::regex("^a{1,3}").unwrap(), "Abcde"));
+        assert!(check(StringPattern::regex_i("^a{1,3}").unwrap(), "Abcde"));
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   output using template expressions, similar to `jj op log`. Also added
   `--no-op-diff` flag to suppress the operation diff.
 
+* A nearly identical string pattern system as revsets is now supported in the
+  template language, and is exposed as `string.match(pattern)`.
+
 ### Fixed bugs
 
 * `jj git clone` now correctly fetches all tags, unless `--fetch-tags` is


lf-:
cc @yuja

cli/src/generic_templater.rs
```diff
@@ -198,6 +198,15 @@ impl<'a, C> CoreTemplatePropertyVar<'a> for GenericTemplatePropertyKind<'a, C> {
         }
     }
 
+    fn try_into_string_pattern(
```
lf-: i am sus of the default impl for this being used as it already caused one bug and i wonder if it will be buggy for the other languages as well. the cause of the bug is that it doesn't consistently forward to an inner impl if it's one of these ::Core types

yuja: Appears that you have bad squash in the first commit like this.

cli/src/commit_templater.rs
```diff
@@ -497,6 +498,17 @@ impl<'repo> CoreTemplatePropertyVar<'repo> for CommitTemplatePropertyKind<'repo>
         }
     }
 
+    fn try_into_string_pattern(
+        self,
+    ) -> Option<BoxedTemplateProperty<'repo, jj_lib::str_util::StringPattern>> {
```
yuja: I don't think we'll need support for dynamic construction of string patterns. See callers of `template_parser::expect_string_literal()` for parsing of static values.

cli/src/template_builder.rs
```diff
@@ -888,10 +938,10 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
...
             let [needle_node] = function.expect_exact_arguments()?;
             // TODO: or .try_into_string() to disable implicit type cast?
-            let needle_property =
-                expect_stringify_expression(language, diagnostics, build_ctx, needle_node)?;
-            let out_property = (self_property, needle_property)
-                .map(|(haystack, needle)| haystack.contains(&needle));
+            let needle_pattern =
+                expect_string_pattern(language, diagnostics, build_ctx, needle_node)?;
+            let out_property =
+                (self_property, needle_pattern).map(|(haystack, needle)| needle.matches(&haystack));
```
yuja: Since `contains()` implies substring matching, the current `StringMatcher` API wouldn't be useful here.

Maybe we can add `.matches(pattern)`, etc. instead? Since we don't have to support dynamic strings, the implementation will become simpler.

cli/src/template_parser.rs
```diff
@@ -145,6 +148,9 @@ pub enum TemplateParseErrorKind {
     RedefinedFunctionParameter,
     #[error("{0}")]
     Expression(String),
+    // XXX: converted to String so that it becomes Eq
+    #[error("Invalid string pattern: {0}")]
+    StringPattern(String),
```
yuja: We often attach the original error as the source.
```
TemplateParseError::expression(message, span).with_source(err)
```

cli/src/template_builder.rs
lf-: This is actually symmetric with `diff_contains` in the revset language which does the same thing. Granted that is kind of weird, but yeah.

cli/src/commit_templater.rs
lf-: I don't fully understand what you mean here. I think I understand that while doing the AST-walking evaluation, we could write something like `expect_string_literal`, but that's not going to work in the existing `contains` function which tries to stringify; we would have to be able to try a static expression and then try a dynamic one if we don't want observable behaviour changes. This is possible but it feels a bit complicated; though so is wiring a new type through the whole expr system so idk.

It doesn't feel entirely inconceivable that someone might swap out the pattern depending on e.g. some setting's value; perhaps it's setting which ticket format to use or something like that.

cli/src/generic_templater.rs
lf-: oops. I will have to fix that.

cli/src/template_parser.rs
lf-: ah, that makes sense. Should I use the `Expression` error type in that case and not have a special error type?

cli/src/template_builder.rs
yuja: `diff_contains()` is a function to test if the diff contains lines matching the pattern, so that seems okay. OTOH, `string.contains(needle)` doesn't mean `string == needle`.

cli/src/template_parser.rs
yuja: Yeah, just use `Expression` unless you need special handling of this error type.

cli/src/commit_templater.rs
yuja: > we could write something like `expect_string_literal`, but that's not going to work in the existing `contains` function which tries to stringify;

That's another reason to not implement `contains(pattern)`.

BTW, in order to support arbitrary `kind:(expr ++ ..)` pattern, we can parse the `kind:` part statically, and evaluate `(expr ++ ..)` as stringifiable expression. I personally wouldn't try this route.

PhilipMetzger:
just a reminder to adhere to commit guidelines here: https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#commit-guidelines since. a `feat:` isn't really a topic. My suggestion would be `templater:`.

lf-:
The full completion of this feature to be able to pull out the matching text in question probably involves introducing a new type that represents a match result (and match groups in regex), but doing so requires some rework and semantics design of the string pattern system itself.

I think that would probably be `.match`. Which probably means this boolean one should be called `is_match` actually as `matches` might mean "what is the list of matches in this string". Either that or *this* one should be called `match` and then we extend the output type to be a more substantive type later and just assume that everyone's using it as truthiness now. Thoughts?

yuja:
> The full completion of this feature to be able to pull out the matching text in question probably involves introducing a new type that represents a match result (and match groups in regex), but doing so requires some rework and semantics design of the string pattern system itself.

For this use case, I think we can add `.match`/`es` and `.replace`. The user might have to write the same regex twice, but that seems okay?

```
# .match(pat) -> String and/or .matches(path) -> List<String>
# .replace_all(pat, subst) -> String
.match(regex:'#\d+\b").replace_all(regex:'^#(.*)', 'issue-$2')
```
(`match`/`matches` might be a bad name, but I'm not sure.)

Implementation-wise, we'll need `StringPattern::to_regex()` or something, and use string-based substitution provided by `Regex`. We could add lambda syntax for `subst` expression, but that would be more complex, and I don't know if we'll need expressivity provided by arbitrary templates.

https://docs.rs/regex/latest/regex/struct.Regex.html#method.replace

lf-:
The issue I have is that it seems nontrivial to make a to_regex translator for globs, and that IME it is most ideal to have the ability to match groups with a regex. But that's probably going to require creating a language type to be able to extract groups and matches; a regex feature that only does full matches is quite a bit harder to use.

So I'm partially trying to figure out how to incrementally deliver something useful without simply . inventing good regex support in my first code PR to the project heh. I think the path forward on that is calling the function introduced in this PR `is_match`.

yuja:
> The issue I have is that it seems nontrivial to make a to_regex translator for globs,

I think we'll need to switch to the globset crate at some point. The current `StringPattern` API doesn't have a separate `.to_matcher()` step, so adapting to `GlobSet` might not be straightforward. Still I think we can leverage `globset` to implement `.to_regex()`.

https://docs.rs/globset/0.4.16/globset/struct.Glob.html

lf-:
I've rebased and rewritten this based on https://github.com/jj-vcs/jj/pull/6967. It should be ready to go from my end minus one code question I couldn't figure out and need help with :)

cli/src/template_builder.rs
```diff
@@ -895,6 +908,59 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
...
+                                ))
+                                .ok_or_else(|| {
+                                    TemplateParseError::expression(
+                                        format!(
+                                            "Nonexistent capture group: {group} in regex {regex:?}"
+                                        ),
+                                        todo!(
+                                            "how do I fix this? I can't keep the span here due to \
+                                             lifetimes. this error doesn't appear statically"
+                                        ),
```
lf-: I need help with how to write this diagnostic in a way that actually works

yuja: I wouldn't add `capture_group` argument to `.match()`. I think we can instead add `.replace(pattern, subst)`, which can be used as `description.match(regex:'h(ell)o').replace(regex:'h(ell)o', '$1')`.

lib/src/str_util.rs
```diff
@@ -250,21 +288,48 @@ impl StringPattern {
...
+            StringPattern::GlobI(pattern) => pattern.is_match(haystack.as_bytes()),
             StringPattern::Regex(pattern) => pattern.is_match(haystack),
             StringPattern::RegexI(pattern) => pattern.is_match(haystack),
         }
     }
 
+    /// Converts the pattern into a string regex.
+    ///
+    /// XXX: this is kind of problematic for byte regexes as would be created
+    /// by globs. I dunno what to do about that.
```
yuja: I'm going to change the underlying `regex` type to `bytes::Regex` to fix bytes handling in `diff_contains()` revset. Suppose this change is accepted, I think there are two options:

1. in templater, use `bytes::Regex` and convert the result back to `String`. If the matched slice wasn't at unicode boundary, `.match()` returns an error.
2. make this `to_regex()` fail if bytes regex cannot be re-interpreted as a string regex.

I think (1) is simpler.

cli/src/commit_templater.rs
```diff
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Template environment for `jj log`, `jj evolog` and similar.
+//!
+//! See: <https://jj-vcs.github.io/jj/latest/templates/#commit-keywords>.
```
yuja: nit: I would remove the doc link because the commit templater can have various context types than the `Commit` type.

cli/src/operation_templater.rs
```diff
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Templates for `jj op log`.
```
yuja: nit: "Template environment" sounds more correct.

cli/src/operation_templater.rs
```diff
@@ -51,6 +55,7 @@ pub trait OperationTemplateLanguageExtension {
     fn build_cache_extensions(&self, extensions: &mut ExtensionsMap);
 }
 
+/// Templates for `jj op log`.
```
yuja: nit: Same here. This isn't a "Template", but "Template environment".

cli/src/templater.rs
```diff
@@ -313,9 +320,9 @@ where
 #[derive(Debug)]
 pub struct TemplatePropertyError(pub Box<dyn error::Error + Send + Sync>);
 
-// Implements conversion from any error type to support `expr?` in function
-// binding. This type doesn't implement `std::error::Error` instead.
-// https://github.com/dtolnay/anyhow/issues/25#issuecomment-544140480
+/// Implements conversion from any error type to support `expr?` in function
+/// binding. This type doesn't implement `std::error::Error` instead.
+/// <https://github.com/dtolnay/anyhow/issues/25#issuecomment-544140480>
```
yuja: nit: This is an implementation comment. Revert the change.

cli/src/templater.rs
```diff
@@ -594,6 +603,9 @@ where
     }
 }
 
+/// Template which, upon being forced (`extract()`ed), will evaluate its
+/// condition and select between a template for the true case and a template
+/// for the false case (which will yield nothing if it is [`None`]).
```
yuja: This sounds more like a textual explanation of the code. Can you simplify? I think it'll be something like "Template which selects output based on a boolean condition."

cli/src/template.pest
```diff
@@ -88,11 +91,18 @@ formal_parameters = {
   identifier ~ (whitespace* ~ "," ~ whitespace* ~ identifier)* ~ (whitespace* ~ ",")?
   | ""
 }
+string_pattern_identifier = @{ (ASCII_ALPHA | "_") ~ (ASCII_ALPHANUMERIC | "_" | "-")* }
```
yuja, line 97: Just a comment. It's nice that you noticed that `-` is allowed as a pattern prefix whereas `-` can also be infix/prefix operators. The grammar isn't ambiguous yet, but it's a bit odd. I don't have a good idea to address this potential problem.

cli/src/template_builder.rs
```diff
@@ -161,6 +167,12 @@ where
     Self: WrapTemplateProperty<'a, Timestamp>,
     Self: WrapTemplateProperty<'a, TimestampRange>,
 {
+    // Don't write default implementations on these (and certainly not ones
+    // using another function on self): wrapper types that contain a sum type
+    // of a CoreTemplatePropertyKind or a custom type will hit the default case
+    // every time rather than the implementation on CoreTemplatePropertyKind if
+    // a custom implementation is omitted.
```
yuja: nit: I wouldn't add this comment because I don't think it is any special here. We shouldn't add default implementation unless that makes sense.

cli/src/template_parser.rs
```diff
@@ -482,6 +497,28 @@ fn parse_term_node(pair: Pair<Rule>) -> TemplateParseResult<ExpressionNode> {
...
+            let kind = kind.as_str();
+            let text = match literal.as_rule() {
+                Rule::string_literal => STRING_LITERAL_PARSER.parse(literal.into_inner()),
+                Rule::raw_string_literal => parse_raw_string_literal(literal),
+                other => {
+                    return Err(TemplateParseError::expression(
+                        format!("Unexpected literal in string pattern: {other:?}"),
+                        span,
+                    ))
+                }
```
yuja: nit: should be `panic!()`?

cli/src/template_parser.rs
```diff
@@ -482,6 +497,28 @@ fn parse_term_node(pair: Pair<Rule>) -> TemplateParseResult<ExpressionNode> {
...
+            };
+            // XXX(jade): this should probably make the StringPattern
+            // itself here, but because StringPattern is not PartialEq (and
+            // cannot be due to regexes), we defer that to later and Deal With
+            // the AST being slightly less clean (and slightly less validated)
+            // than is ideal.
+            ExpressionNode::new(ExpressionKind::StringPattern { kind, value: text }, span)
```
yuja: I think the current implementation is correct. We aren't doing any revset/fileset/string pattern parsing in the parser. And we might add other kind of patterns (e.g. date pattern) that share the same syntax.

cli/src/templater.rs
```diff
@@ -91,6 +92,12 @@ impl Template for &BStr {
     }
 }
 
+impl Template for StringPattern {
+    fn format(&self, formatter: &mut TemplateFormatter) -> io::Result<()> {
+        write!(formatter, "{self}")
+    }
+}
+
```
yuja: nit: shouldn't be needed anymore?

cli/src/template_parser.rs
lf-: yep it should as this should be statically impossible based on the grammar

cli/src/template_parser.rs
lf-: done, added a comment that it is deferred to later

cli/src/templater.rs
lf-: that's right!

lf-:
Review comments fixed, added a test or two, got rid of capture group support, switched to bytes regex within this code (so that the rest of the migration is easier), PTAL.

cli/src/operation_templater.rs
```diff
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Template environment for `jj op log`.
+//!
+//! See <https://jj-vcs.github.io/jj/latest/templates/#operation-keywords>.
```
yuja: nit: remove the link consistently?

lib/src/str_util.rs
```diff
@@ -439,27 +439,47 @@ mod tests {
...
+        assert!(!check(StringPattern::exact("a"), "aa"));
+        assert!(check(StringPattern::exact_i("a"), "A"));
+        assert!(check(StringPattern::substring("a"), "abc"));
+        assert!(!check(StringPattern::substring("a"), "Abc"));
+        assert!(check(StringPattern::substring_i("a"), "Abc"));
+        assert!(!check(StringPattern::glob("a").unwrap(), "A"));
+        assert!(check(StringPattern::glob_i("a").unwrap(), "A"));
+        assert!(check(StringPattern::regex("^a{1,3}").unwrap(), "abcde"));
+        assert!(!check(StringPattern::regex("^a{1,3}").unwrap(), "Abcde"));
+        assert!(check(StringPattern::regex_i("^a{1,3}").unwrap(), "Abcde"));
```
yuja: nit: bad squash. should belong to the last patch.

cli/src/templater.rs
```diff
@@ -594,6 +603,10 @@ where
     }
 }
 
+/// Template which selects an output based on a boolean condition.
+///
+/// When `None` is specified for the false template and the condition is false,
+/// this returns an empty string.
```
yuja: nit: "returns an empty string" isn't technically correct. Maybe "the output is empty", "writes nothing", etc.?

cli/src/template_parser.rs
```diff
@@ -665,6 +695,24 @@ pub fn expect_string_literal<'a>(node: &'a ExpressionNode<'_>) -> TemplateParseR
...
 }
 
+/// Unwraps inner value if the given `node` is a string pattern
+///
+/// This forces it to be static so that it need not be part of the type system.
+pub fn expect_string_pattern(node: &'_ ExpressionNode<'_>) -> TemplateParseResult<StringPattern> {
+    catch_aliases_no_diagnostics(node, |node| match &node.kind {
+        ExpressionKind::StringPattern { kind, value } => StringPattern::from_str_kind(value, kind)
+            .map_err(|err| {
+                TemplateParseError::expression(format!("Bad string pattern: {err}"), node.span)
```
yuja: nit: remove `: {err}`. the source error should be printed separately.

cli/src/template_parser.rs
```diff
@@ -1131,6 +1180,45 @@ mod tests {
...
+            }),
+        );
+        assert_eq!(
+            parse_into_kind(r#"bad:"meow""#),
+            Ok(ExpressionKind::StringPattern {
+                kind: "bad",
+                value: "meow".to_owned()
+            }),
+            "we currently accept bad kinds and reject them after parsing due to PartialEq \
+             limitations"
```
yuja: nit: remove this assertion message. The current implementation is correct, and I think this test can be moved to template_builder.rs where `kind:"pattern"` expression is interpreted as a string pattern.

cli/src/template_builder.rs
```diff
@@ -902,6 +902,28 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
...
+                let regex = needle.to_regex();
+                let match_ = regex.find(haystack.as_bytes());
+
+                if let Some(m) = match_ {
+                    Ok(String::from_utf8(m.as_bytes().to_owned())?)
+                } else {
+                    // FIXME(jade): maybe suboptimal to return string
+                    // unconditionally but we don't have an Option<String> type
+                    // yet.
+                    Ok(String::new())
```
yuja: We can return evaluation error. (see `"format"` in `builtin_timestamp_methods()` for example.)

To propagate utf-8 error, you might want to use `str::from_utf8(..)?.to_owned()` instead, because `String::from_utf8` embeds the source `Vec<u8>` data in the error struct and we'll need to strip it off.

lib/src/str_util.rs
```diff
@@ -59,6 +59,14 @@ impl GlobPattern {
     pub fn as_str(&self) -> &str {
         self.glob.glob()
     }
+
+    /// Converts to a bytes regex.
+    pub fn to_regex(&self) -> regex::bytes::Regex {
+        regex::bytes::RegexBuilder::new(self.glob.regex())
+            .dot_matches_new_line(true)
+            .build()
+            .expect("glob regex should be valid")
```
yuja: `self.regex.clone()`?

lib/src/str_util.rs
```diff
@@ -283,6 +291,44 @@ impl StringPattern {
...
     }
 
+    /// Converts the pattern into a bytes regex.
+    pub fn to_regex(&self) -> regex::bytes::Regex {
+        match self {
+            StringPattern::Exact(literal) => {
+                regex::bytes::RegexBuilder::new(&format!("^{}$", regex::escape(literal)))
+                    .build()
+                    .expect("impossible to fail to compile regex of literal")
+            }
```
yuja, line 304: See https://github.com/jj-vcs/jj/pull/7013, which is blocked by the other PR, but I think it will be accepted.

cli/src/template_builder.rs
yuja: Appears that I misunderstood the code. In order to represent a "no match" result, an empty string is the most correct value. We can't add `Option<String>` because `String` itself implements an implicit conversion to bool, so it's unclear whether `if(opt_string, ..)` means `if(None` or `if(Some("")`.

lib/src/str_util.rs
lf-: Filed a bug since the tools jj tells me to use for reviewing `jj absorb` are not readable for me, so I did this mistake: https://github.com/jj-vcs/jj/issues/7039

lib/src/str_util.rs
lf-: what's a little more rebasing between friends. rebased onto that cuz it is indeed helpful to this.

lf-:
Should be sufficiently rebased and feedback addressed.

cli/src/template_builder.rs
yuja, line 1: nit: I wouldn't call it a bug because the test doesn't care much about the source errors. Of course, it's good to include non-parsing source errors in snapshots, though.

cli/src/template_builder.rs
```diff
@@ -2137,8 +2137,17 @@ mod tests {
         }
 
         fn parse_err(&self, template: &str) -> String {
-            let err = self.parse(template).err().unwrap();
-            iter::successors(Some(&err), |e| e.origin()).join("\n")
+            let Err(orig_err) = self.parse(template) else {
```
yuja: nit: use `self.parse(template).err().unwrap()` as before.

cli/src/template_builder.rs
```diff
@@ -2137,8 +2137,17 @@ mod tests {
...
+                panic!("Got unexpected successful template rendering");
+            };
+
+            let mut err: Option<&dyn std::error::Error> = Some(&orig_err);
+            let mut out = String::new();
+            while let Some(inner) = err {
+                writeln!(&mut out, "{inner}").unwrap();
+                err = std::error::Error::source(inner);
+            }
+            out
```
yuja: nit: this can be `iter::successors(Some(&err as &dyn std::error::Error), |e| e.source()).join("\n")`. we don't care about the extra allocation cost.

cli/src/template_builder.rs
```diff
@@ -902,6 +902,28 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
...
+            let out_property = self_property.and_then(move |haystack| {
+                let regex = needle.to_regex();
+                let match_ = regex.find(haystack.as_bytes());
+
+                if let Some(m) = match_ {
+                    Ok(std::str::from_utf8(m.as_bytes())?.to_owned())
+                } else {
+                    // FIXME(jade): maybe suboptimal to return string
+                    // unconditionally but we don't have an Option<String> type
+                    // yet.
```
yuja: nit: remove this TODO as it's correct to return an empty String.

cli/src/template_builder.rs
PhilipMetzger, line 1: nit: drop the conventional commit prefix to adhere to our style

PhilipMetzger (commented):
last minor thing from me.

lf-:
rebased, I'm going to fix up the feedback now

edit: done!

cli/src/template.pest
```diff
@@ -89,10 +92,21 @@ formal_parameters = {
   | ""
 }
 
+// NOTE: string pattern identifiers additionally allow "-" in them, which
+// results in some oddness with the `-` operator, though does not yet cause
+// ambiguity. This may prove annoying at some future point.
+string_pattern_identifier = @{ (ASCII_ALPHA | "_") ~ (ASCII_ALPHANUMERIC | "_" | "-")* }
+string_pattern = {
```
yuja: Since I've merged https://github.com/jj-vcs/jj/pull/7380, we'll probably need `${ .. }`. Can you also add a test for `prefixed-pattern : "with space"`, which should be a parse error?

cli/src/template_builder.rs
```diff
@@ -909,6 +909,27 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>()
...
         },
     );
+    map.insert(
+        "match",
+        |_language, _diagnostics, _build_ctx, self_property, function| {
+            let [needle_node] = function.expect_exact_arguments()?;
+            let needle = template_parser::expect_string_pattern(needle_node)?;
+
+            let out_property = self_property.and_then(move |haystack| {
+                let regex = needle.to_regex();
```
yuja: nit: `to_regex()` can be moved out of the closure.

cli/src/template_builder.rs
```diff
@@ -909,6 +909,27 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>()
...
+                let match_ = regex.find(haystack.as_bytes());
```
yuja: nit: inline this as the `if` condition is short?

docs/templates.md
```diff
@@ -426,6 +426,10 @@ defined.
 * `.len() -> Integer`: Length in UTF-8 bytes.
 * `.contains(needle: Stringify) -> Boolean`: Whether the string contains the
   provided stringifiable value as a substring.
+* `.match(needle: StringPattern) -> Boolean`: Extracts
```
yuja: `-> String`

yuja (commented):
LG, thanks.

docs/templates.md
```diff
@@ -474,6 +475,21 @@ An expression that can be converted to a `String`.
...
+These are the exact same as the [String pattern type] in revsets, except that
+quotes are mandatory.
+
+Literal strings may be used, which are interpreted as case-sensitive substring
+matching.
+
+Currently `StringPattern` values cannot be passed around as values and may
+only occur directly in the call site they are used in.
+
+[String pattern type]: ./revsets.md#string-patterns
```
PhilipMetzger: nit: we use relative links throughout our docs, so drop the `./`

PhilipMetzger (commented):
LG

lf-:
okay, i *think* that's good to go now

cli/src/template.pest
```diff
@@ -88,10 +91,21 @@ formal_parameters = {
...
```
yuja: `${ .. }`

cli/src/template_parser.rs
```diff
@@ -69,8 +70,11 @@ impl Rule {
             Self::string_literal => None,
             Self::raw_string_content => None,
             Self::raw_string_literal => None,
+            Self::any_string_literal => None,
+            Self::string_pattern => None,
             Self::integer_literal => None,
             Self::identifier => None,
+            Self::string_pattern_identifier => None,
```
yuja: nit: insert these in definition order? `formal_parameters, string_pattern_identifier, string_pattern`

cli/src/template_parser.rs
```diff
@@ -663,6 +695,23 @@ pub fn expect_string_literal<'a>(node: &'a ExpressionNode<'_>) -> TemplateParseR
     })
 }
 
+/// Unwraps inner value if the given `node` is a string pattern
+///
+/// This forces it to be static so that it need not be part of the type system.
+pub fn expect_string_pattern(node: &'_ ExpressionNode<'_>) -> TemplateParseResult<StringPattern> {
```
yuja: nit: `&ExpressionNode<'_>` (remove redundant `&'_`)

docs/templates.md
```diff
@@ -426,6 +426,10 @@ defined.
 * `.len() -> Integer`: Length in UTF-8 bytes.
 * `.contains(needle: Stringify) -> Boolean`: Whether the string contains the
   provided stringifiable value as a substring.
+* `.match(needle: StringPattern) -> String`: Extracts
+  the matching part of the string for the given pattern.
```
yuja: nit: maybe better to say "the first matching part" for clarity.

lib/src/str_util.rs
```diff
@@ -487,4 +521,24 @@ mod tests {
...
+    #[test]
+    fn test_string_pattern_to_regex() {
+        let check = |pattern: StringPattern, match_to: &str| {
+            let regex = pattern.to_regex();
+            regex.is_match(match_to.as_bytes())
+        };
+        assert!(check(StringPattern::exact("a"), "a"));
+        assert!(!check(StringPattern::exact("a"), "A"));
+        assert!(!check(StringPattern::exact("a"), "aa"));
+        assert!(check(StringPattern::exact_i("a"), "A"));
```
yuja, line 535: nit: can you add some test samples that would fail without `regex::escape()`?

lf-:
Feedback should be fully addressed, PTAL.

yuja (approved):

PR #6905, opened 2025-07-05
status: collapse entirely untracked directories

afishhh:
I wanted to try out `jj` since it seemed cool, was happy that `snapshot.auto-track` was added since I keep many untracked files in my repos, however once I ran `jj st` it spat out a line for every single file in my untracked directories into a pager which is not optimal. This PR collapses directories which consist entirely of untracked files (i.e. without any files that are tracked in the working copy inside).

Note that this is purely a `jj status` post processing step and does not affect how untracked files are collected. It may also be nice to do something like this in other places. ~~I think git also does this for stuff like added files too so if you add a directory it just shows the directory, that just seems a bit more involved and is not a problem for me right now~~. git does not do this.

Not sure where in the changelog this would go (New features or bug fixes?) or whether any docs need to be updated.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/status.rs
+++ b/cli/src/commands/status.rs
@@ -12,13 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::io;
-
 use itertools::Itertools as _;
 use jj_lib::copies::CopyRecords;
+use jj_lib::merged_tree::MergedTree;
 use jj_lib::repo::Repo as _;
+use jj_lib::repo_path::RepoPath;
+use jj_lib::repo_path::RepoPathBuf;
 use jj_lib::revset::RevsetExpression;
 use jj_lib::revset::RevsetFilterPredicate;
+use pollster::FutureExt as _;
 use tracing::instrument;
 
 use crate::cli_util::print_conflicted_paths;
@@ -104,11 +106,24 @@ pub(crate) fn cmd_status(
             if wc_has_untracked {
                 writeln!(formatter, "Untracked paths:")?;
                 formatter.with_label("diff", |formatter| {
-                    for path in snapshot_stats.untracked_paths.keys() {
-                        let ui_path = workspace_command.path_converter().format_file_path(path);
-                        writeln!(formatter.labeled("untracked"), "? {ui_path}")?;
-                    }
-                    io::Result::Ok(())
+                    visit_collapsed_untracked_files(
+                        snapshot_stats.untracked_paths.keys(),
+                        tree,
+                        |path, is_dir| {
+                            let ui_path = workspace_command.path_converter().format_file_path(path);
+                            writeln!(
+                                formatter.labeled("untracked"),
+                                "? {ui_path}{}",
+                                if is_dir {
+                                    std::path::MAIN_SEPARATOR_STR
+                                } else {
+                                    ""
+                                }
+                            )?;
+                            Ok(())
+                        },
+                    )
+                    .block_on()
                 })?;
             }
         }
@@ -211,4 +226,136 @@ pub(crate) fn cmd_status(
     }
 
     Ok(())
+}
+
+async fn visit_collapsed_untracked_files(
+    untracked_paths: impl IntoIterator<Item = impl AsRef<RepoPath>>,
+    tree: MergedTree,
+    mut on_path: impl FnMut(&RepoPath, bool) -> Result<(), CommandError>,
+) -> Result<(), CommandError> {
+    let mut stack = vec![tree];
+
+    // TODO: This loop can be improved with BTreeMap cursors once that's stable,
+    // would remove the need for the whole `skip_prefixed_by` thing and turn it
+    // into a B-tree lookup.
+    let mut skip_prefixed_by_dir: Option<RepoPathBuf> = None;
+    'untracked: for path in untracked_paths {
+        let path = path.as_ref();
+        if skip_prefixed_by_dir
+            .as_ref()
+            .is_some_and(|p| path.starts_with(p))
+        {
+            continue;
+        } else {
+            skip_prefixed_by_dir = None;
+        }
+
+        let mut it = path.components().dropping_back(1);
+        let first_mismatch = it.by_ref().enumerate().find(|(i, component)| {
+            stack.get(i + 1).is_none_or(|tree| {
+                tree.dir()
+                    .components()
+                    .next_back()
+                    .expect("should always have at least one element (the root)")
+                    != *component
+            })
+        });
+
+        if let Some((i, component)) = first_mismatch {
+            stack.truncate(i + 1);
+            for component in std::iter::once(component).chain(it) {
+                let parent = stack
+                    .last()
+                    .expect("should always have at least one element (the root)");
+
+                if let Some(subtree) = parent.sub_tree(component).await? {
+                    stack.push(subtree);
+                } else {
+                    let dir = parent.dir().join(component);
+
+                    on_path(&dir, true)?;
+                    skip_prefixed_by_dir = Some(dir);
+
+                    continue 'untracked;
+                }
+            }
+        }
+
+        on_path(path, false)?;
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod test {
+    use testutils::repo_path;
+    use testutils::TestRepo;
+    use testutils::TestTreeBuilder;
+
+    use super::*;
+
+    fn collect_collapsed_untracked_files_string(
+        untracked_paths: &[&RepoPath],
+        tree: MergedTree,
+    ) -> String {
+        let mut result = String::new();
+        visit_collapsed_untracked_files(untracked_paths, tree, |path, is_dir| {
+            result.push_str("? ");
+            if is_dir {
+                result.push_str(&path.to_internal_dir_string());
+            } else {
+                result.push_str(path.as_internal_file_string());
+            }
+            result.push('\n');
+            Ok(())
+        })
+        .block_on()
+        .unwrap();
+        result
+    }
+
+    #[test]
+    fn test_collapsed_untracked_files() {
+        let repo = TestRepo::init();
+
+        let tracked = {
+            let mut builder = TestTreeBuilder::new(repo.repo.store().clone());
+
+            builder.file(repo_path("top_level_file"), "");
+            // ? "untracked_top_level_file"
+            // ? "dir"
+            // ? "dir2/c"
+            builder.file(repo_path("dir2/d"), "");
+            // ? "dir3/partially_tracked/e"
+            builder.file(repo_path("dir3/partially_tracked/f"), "");
+            // ? "dir3/fully_untracked/"
+            builder.file(repo_path("dir3/j"), "");
+            // ? "dir3/k"
+
+            builder.write_merged_tree()
+        };
+        let untracked = &[
+            repo_path("untracked_top_level_file"),
+            repo_path("dir/a"),
+            repo_path("dir/b"),
+            repo_path("dir2/c"),
+            repo_path("dir3/partially_tracked/e"),
+            repo_path("dir3/fully_untracked/g"),
+            repo_path("dir3/fully_untracked/h"),
+            repo_path("dir3/k"),
+        ];
+
+        insta::assert_snapshot!(
+            collect_collapsed_untracked_files_string(untracked, tracked),
+            @r"
+        ? untracked_top_level_file
+        ? dir/
+        ? dir2/c
+        ? dir3/partially_tracked/e
+        ? dir3/fully_untracked/
+        ? dir3/k
+        "
+        );
+    }
 }
--- a/cli/tests/test_status_command.rs
+++ b/cli/tests/test_status_command.rs
@@ -392,8 +392,7 @@ fn test_status_untracked_files() {
     Untracked paths:
     ? always-untracked-file
     ? initially-untracked-file
-    ? sub/always-untracked
-    ? sub/initially-untracked
+    ? sub/
     Working copy  (@) : qpvuntsm e8849ae1 (empty) (no description set)
     Parent commit (@-): zzzzzzzz 00000000 (empty) (no description set)
     [EOF]
@@ -449,8 +448,7 @@ fn test_status_untracked_files() {
     Untracked paths:
     ? always-untracked-file
     ? initially-untracked-file
-    ? sub/always-untracked
-    ? sub/initially-untracked
+    ? sub/
     Working copy  (@) : mzvwutvl 240f261a (no description set)
     Parent commit (@-): qpvuntsm b8c1286d (no description set)
     [EOF]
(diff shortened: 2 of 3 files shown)


yuja:
I'm not against adding collapsed output, but there should usually be zero or a few untracked files. `snapshot.auto-track` is primarily designed for temporary exclusion. If there are files that should be permanently excluded, these files should be added to `.gitignore` or `.git/info/exclude` to get better runtime performance.

afishhh:
> I'm not against adding collapsed output, but there should usually be zero or a few untracked files. `snapshot.auto-track` is primarily designed for temporary exclusion. If there are files that should be permanently excluded, these files should be added to `.gitignore` or `.git/info/exclude` to get better runtime performance.

I know that's the "jj way" but to be honest I just disagree and find untracked files cleaner than `.git/info/exclude` (you won't convince me otherwise). I'm fine with having a slight slow down since it's not a problem on my scale. Performance could be improved later anyway by actually short-circuiting when per-file data is unnecessary and then jj would be no worse than git (I hope such a change wouldn't be rejected on "just don't have untracked files" grounds).

Anyway I think it's not worth discussing the merit of untracked files here since that's been done extensively in the untracked files issue.

edit: I suppose a better way to word what I want to say would be that "temporary files" often consist of many files, for example a build directory. This *is* what `snapshot.auto-track` was designed for but `jj st` printing the whole build directory is not friendly to this use-case.

joyously:
Regardless of what _should_ be, it seems like there could be an option for `--verbose` that shows all the individual files, but since it's a status command, perhaps that's the default, and the option should be `--quiet` or something more suitable.

afishhh:
> Regardless of what _should_ be, it seems like there could be an option for `--verbose` that shows all the individual files, but since it's a status command, perhaps that's the default, and the option should be `--quiet` or something more suitable.

git has an option like this:
```
       -u[<mode>], --untracked-files[=<mode>]
           Show untracked files.

           •   no - Show no untracked files.
           •   normal - Shows untracked files and directories.
           •   all - Also shows individual files in untracked directories.
```
(from the `git status` man page, extremely simplified)
The default behavior being `normal` and if a bare `-u` flag is specified `all`.

I think that's a reasonable design, but if we decide `jj` should default to `all` then I think this should be made configurable.

joyously:
This sort of brings up the distinction between "ignored" and "untracked" again. 
IMO there should be one source of truth for files, and it would be binary (tracked/ignored), but that is not the way it is written.
Should the `status` command show the status of files that are explicitly untracked, which means they are ignored? Or should it show only the untracked files not specified in the various places(.ignore, auto-snapshot, exclude list, untrack command) that list files?

hotsphink:
Another benefit of collapsing is that it is more likely what you would want to add to an ignore file: if an entire directory's contents are untracked and you're ok with ignoring them, then you'd probably want to ignore that directory not the individual files.

Going further, I could imagine reporting untracked files in `.gitignore` syntax, so they could be cut & paste into an ignore file.

(Note that I'm not arguing for or against leaving things untracked.)

afishhh:
I think my first comment came off a bit rough, I was under residual annoyance caused by people arguing against untracked files in the original issue, really sorry for that. I agree with how untracked files currently work with `jj`, just think they could use some more work.

> This sort of brings up the distinction between "ignored" and "untracked" again.
IMO there should be one source of truth for files, and it would be binary (tracked/ignored), but that is not the way it is written.
Should the status command show the status of files that are explicitly untracked, which means they are ignored? Or should it show only the untracked files not specified in the various places(.ignore, auto-snapshot, exclude list, untrack command) that list files?

In my opinion the current handling, which seems to be the second alternative, is fine. Untracked files exist so you can temporarily (for whatever definition of "temporarily", which may vary) clutter up the repository without cluttering up the index. `jj status` listing such files is useful so that you don't forget they're still there, however listing files which have been explicitly excluded would not be useful in the same way.
Also I'm not sure I fully understand because the untrack command as I understand it does not store anything and files untracked by it will get re-tracked if auto-snapshot doesn't exclude them?

> Going further, I could imagine reporting untracked files in .gitignore syntax, so they could be cut & paste into an ignore file.

The current `? path` lines can be easily turned into `.gitignore` syntax by replacing `? ` with `/`. Not sure whether it's worth it to sacrifice the `?` sigil for that, it'd make viewing the status without color harder at a glance. I'm fine either way though.

thoughtpolice:
I think this is a good change, and I don't see any real reason to special case it or make it configurable at the moment TBQH; it can just be always enabled as the default output setting. It's familiar enough from Git anyway, and I've actually never even heard of or even used that `-u` flag for any reason. I think it's fine to just leave it out and add it later.

Note that Git also just shows `./` when you are inside a directory that is fully untracked. I also think that's ok, but there doesn't seem to be a test for it, but I haven't looked closely. If you can add one or just double-check, then I'll do a proper review (and likely approve it.)

ilyagr:
The performance consideration Yuya mentioned is interesting. I wonder whether we should have a separate feature to warn the user if they have so many untracked files that it affects performance. In fact, it could be combined with the feature from this PR: we could report a count of files under the directory.

But, in my mind, this goes beyond the scope of this PR.

For the larger question, it bothers me that we don't have a jj-native answer to `.git/repo/exlcude`, but that's tricky and also seems beyond the scope of this PR. (Perhaps it'd be easier if the jj-native way was disabled for co-located repos?)

It's unclear to me how many (or few) use-cases there are that are not addressed by `.gitignore` and `.git/repo/exclude` or by `jj file untrack`-ing only a few files.

martinvonz:
I agree that collapsing the output seems useful already even if we don't optimize the directory walk.

> In fact, it could be combined with the feature from this PR: we could report a count of files under the directory.

That would make a later optimiztion harder, though (because we would need o find that count somehow without walking the directory), so I'd rather not do that.

cli/src/commands/status.rs
martinvonz, line 1: This change deserves an entry in CHANGELOG.md

cli/src/commands/status.rs
```diff
@@ -212,3 +218,60 @@ pub(crate) fn cmd_status(
...
+    formatter: &mut dyn Formatter,
+    untracked_paths: &BTreeMap<RepoPathBuf, UntrackedReason>,
+    tree: &MergedTree,
+    path_converter: &RepoPathUiConverter,
+) -> io::Result<()> {
+    formatter.with_label("diff", |formatter| {
+        let tracked = tree
+            .entries()
+            .map(|entry| entry.0.into_internal_string())
+            .collect::<BTreeSet<String>>();
```
yuja: It can be super expensive to collect all tree entries recursively. Maybe we can instead visit trees from the root by using `MergedTree::sub_tree(name)` or `Tree::value(name)`, and cache visited entries.

cli/src/commands/status.rs
```diff
@@ -212,3 +218,60 @@ pub(crate) fn cmd_status(
...
+
+        // TODO: This loop can be improved with BTreeSet cursors once that's stable,
+        // would remove the need for the whole `skip_prefixed_by` thing and turn it
+        // into a BTree lookup.
+        let mut skip_prefixed_by_dir = None;
+        for original_path in untracked_paths.keys() {
+            if skip_prefixed_by_dir
+                .as_ref()
+                .is_some_and(|p| original_path.as_internal_file_string().starts_with(p))
```
yuja: It's probably better to compare paths as `&RepoPath`, not as `&str`.

afishhh: Can't really be done because `RepoPath` can't represent paths with a trailing `/`. This comparison only works because `skip_prefixed_by_dir` contains a trailing `/`, otherwise it would cause false positives if there is a tree like this:
```
directory/
directory_2/file
```
then printing `directory_2/file` would be skipped because it starts with `directory`, this doesn't happen if we check whether it starts with `directory/`.
At least I think this would happen unless I'm missing something.

cli/src/commands/status.rs
afishhh: I'll try doing that, it should be possible by keeping a stack of `MergedTree` and pushing and popping them incrementally while iterating. I assume doing something like `tree.entries().next().is_none()` on each tree would be fine though right, or is there a better way to check for empty `MergedTree`s?

afishhh: Did this and it seems to work. It's slightly more complex and makes me think it may be conceptually simpler if actually done while snapshotting, so that's another reason to do it properly in the future.
(also I changed emptiness to do `tree.names().next()` since that seems like it would be cheaper)

ilyagr:
> > In fact, it could be combined with the feature from this PR: we could report a count of files under the directory.
> 
> That would make a later optimiztion harder, though (because we would need o find that count somehow without walking the directory), so I'd rather not do that.

There is also the usual option of ceasing to count once there are too many. If the main goal is to tell the user the number of untracked files they have is affecting performance, this seems like it would be fine for the UI.

It would add complexity, though. With elided revisions, we talked about something similar, but it never got implemented because it was tricky IIRC.

cli/src/commands/status.rs
yuja: I'm not sure what the underlying problem is, but `RepoPath("directory_2/file").starts_with(RepoPath("directory"))` should be false (because they don't share the common prefix.)

afishhh: Oh you're right I didn't notice it does actually check for the `/`. That should be a bit cleaner then, thanks.

cli/src/commands/status.rs
```diff
@@ -212,3 +217,70 @@ pub(crate) fn cmd_status(
 
     Ok(())
 }
+
+fn print_collapsed_untracked_files(
+    formatter: &mut dyn Formatter,
+    untracked_paths: &BTreeMap<RepoPathBuf, UntrackedReason>,
+    tree: MergedTree,
+    path_converter: &RepoPathUiConverter,
+) -> Result<(), CommandError> {
```
yuja: nit: instead of printing paths in this function, I think we can collect result to `Vec<&RepoPath>` or `Vec<RepoPathBuf>`. That will help write unit test for this function.

cli/src/commands/status.rs
```diff
@@ -212,3 +217,70 @@ pub(crate) fn cmd_status(
...
+
+            let mut it = path.components().dropping_back(1).enumerate();
+            while let Some((i, component)) = it.next() {
+                if stack.get(i + 1).is_some_and(|tree| {
+                    // This shouldn't panic because all trees in the stack after the first one
+                    // should be descendants of the root and therefore have at least one component.
+                    tree.dir().components().next_back().unwrap() == component
+                }) {
+                    continue;
+                }
```
yuja: nit: I think the first part can be split to a separate loop.
```rust
it.by_ref().enumerate().find(|(i, component)| ..);
...
for component in iter::once(component).chain(it) {
...
```

cli/src/commands/status.rs
```diff
@@ -212,3 +217,70 @@ pub(crate) fn cmd_status(
...
+
+                stack.drain(i + 1..);
+                for component in std::iter::once(component).chain(it.map(|(_, comp)| comp)) {
+                    // `stack` should always have at least one element (the root)
+                    // so this shouldn't panic
+                    let parent = stack.last().unwrap();
```
yuja: nit: `.expect("should always have at least one element (the root)")` instead of `.unwrap()` + comment.

cli/src/commands/status.rs
```diff
@@ -212,3 +217,70 @@ pub(crate) fn cmd_status(
...
+            while let Some((i, component)) = it.next() {
+                if stack.get(i + 1).is_some_and(|tree| {
+                    // This shouldn't panic because all trees in the stack after the first one
+                    // should be descendants of the root and therefore have at least one component.
+                    tree.dir().components().next_back().unwrap() == component
+                }) {
+                    continue;
+                }
+
+                stack.drain(i + 1..);
```
yuja: nit: `stack.truncate(i + 1)`

cli/tests/test_status_command.rs
```diff
@@ -463,8 +461,18 @@ fn test_status_untracked_files() {
...
+    Parent commit (@-): mzvwutvl 240f261a (no description set)
+    [EOF]
+    ");
+
+    let output = work_dir.dir("sub").run_jj(["status"]);
+    insta::assert_snapshot!(output.normalize_backslash(), @r"
+    Untracked paths:
+    ? ../always-untracked-file
+    ? ../initially-untracked-file
+    ? ./
```
yuja, line 475: Can you add test for nested directories?

Since CLI e2e tests are slow, it'll be nice if you can add unit test for the collapse function. To set up tree, I think you can use `testutils::TestTreeBuilder`.

cli/src/commands/status.rs
```diff
@@ -212,3 +217,70 @@ pub(crate) fn cmd_status(
...
+                for component in std::iter::once(component).chain(it.map(|(_, comp)| comp)) {
+                    // `stack` should always have at least one element (the root)
+                    // so this shouldn't panic
+                    let parent = stack.last().unwrap();
+                    let subtree = parent
+                        .sub_tree(component)
+                        .block_on()?
+                        .filter(|subtree| subtree.names().next().is_some());
```
yuja: I think this `.filter()` is redundant. `MergedTree::sub_tree()` appears to return `None` if the entry isn't a tree (or missing.)

afishhh: It seems to return an empty tree if that's what a multiple-input merge would result in, which the doc comment kind of states. I don't know if this is applicable here so I chose the safe option, I can change it if we're certain this works though.

cli/src/commands/status.rs
afishhh: Right, that is a bit simpler, thanks!

cli/src/commands/status.rs
martinvonz: An individual side of the merge can be an empty tree, but the overall merged tree should never be empty.

cli/src/commands/status.rs
afishhh: Added a callback parameter instead, hope that's fine.

cli/src/commands/status.rs
afishhh: I've added an entry under `New features`.

cli/src/commands/status.rs
afishhh: Ahhhh so the doc comment means that if that tree does not exist in one of the inputs it will be treated as an empty tree during the merge? That makes sense then, sorry I got really confused by that doc comment.

martinvonz: Yes, that comment is confusing. Feel free to send another PR to clarify it :)

cli/src/commands/status.rs
```diff
@@ -212,3 +226,132 @@ pub(crate) fn cmd_status(
...
+        });
+
+        if let Some((i, component)) = first_mismatch {
+            stack.truncate(i + 1);
+            for component in std::iter::once(component).chain(it) {
+                let parent = stack
+                    .last()
+                    .expect("should always have at least one element (the root)");
+
+                if let Some(subtree) = parent.sub_tree(component).block_on()? {
```
yuja: nit: make this `async fn` and move `.block_on()` to caller

cli/src/commands/status.rs
```diff
@@ -212,3 +226,132 @@ pub(crate) fn cmd_status(
...
+        .unwrap();
+        result
+    }
+
+    #[test]
+    fn test_collapsed_untracked_files() {
+        let repo = testutils::TestRepo::init();
+
+        let tracked = {
+            let mut builder = testutils::TestTreeBuilder::new(repo.repo.store().clone());
```
yuja: nit: we usually do `use testutils::<type>` per module

cli/src/commands/status.rs
```diff
@@ -212,3 +226,132 @@ pub(crate) fn cmd_status(
...
+                } else {
+                    let dir = parent.dir().join(component);
+
+                    on_path(&dir, true)?;
+                    skip_prefixed_by_dir = Some(dir);
+
+                    continue 'untracked;
+                }
+            }
+        };
```
yuja: nit: redundant semicolon

martinvonz:
@afishhh: Since this has been approved, feel free to merge it (but no rush).

afishhh:
Alright did that, thanks guys. Thank you for the very nice piece of software that `jj` is.

Issue #6809, opened 2025-06-22
FR: Include source line number in `FileAnnotation`


mernen:
`gitk` has a feature that I consider immensely useful, even though it's seldom available in other tools: you can right-click any diff line and pick "Show origin of this line". It'll jump straight to the diff of the commit that produced said line.

I'd like to implement this using `jj`, but in order to accurately show the line being introduced, I need to know its number in the source commit. AIUI this information is already computed, but it's discarded right here (the `_`):

https://github.com/jj-vcs/jj/blob/8afd0dd8ee4d6bdc59ccf2f3334e2d0642ece64a/lib/src/annotate.rs#L353-L355

Perhaps `OriginalLineMap` could be changed to store not just `CommitId`, but also line number? Something like:

```rust
struct LineSource {
    commit_id: CommitId,
    line_number: usize,
}
```

This struct would also allow future expansion to include the source file, in case a backend supports tracking renames, or the annotation algorithm is changed to automatically detect renames like `git blame` does.

PhilipMetzger:
@bryceberger made the output format template-able which should work for you. Otherwise we should extend it.

mernen:
`AnnotationLine` does have a `line_number()` method, but it returns the resulting file's number, not the line's original number. For example: if I run `jj file annotate README.md` on this repo, line 350 says:

    plulkmrl aseipp   2023-10-07 18:35:06  350: ## License

On revision `plulkmrl`, however, this matches to line 224:

https://github.com/jj-vcs/jj/blob/b66adb4bb78c0199fe0683c2474b1fc271bf7a75/README.md?plain=1#L224

AFAIK, right now there's no way to obtain this number 224, even though the annotation engine did calculate it at some point.

PhilipMetzger:
So the fix would be to provide a `.original_line()` on `AnnotationLine`? I assume this field would be something like an optional since for some files the `original_line()` may be missing.

mernen:
That would be an option. I also considered a new `AnnotationLine.source()` that returns a `LineSource` object, which then has `commit()` and `line_number()` methods (`commit()` being synonymous with the current `AnnotationLine.commit()`).

Regarding being optional: other than when the source is outside the search domain, I can't think of any scenario where it could fail (and in this case `annotate` already returns `root()` as commit rather than none, so I don't think it'd be surprising if it returned something like the same as `AnnotationLine.line_number()`).

PhilipMetzger:
[...]

PR #6907, opened 2025-07-06
annotate: provide original line number for each line

### LINKED ISSUE #6103: FR: `jj file annotate` should be able to report the line numbers in the original commit to enable nice TUI functionality

One useful functionality for a `jj file annotate` TUI would be the ability to jump from a line in commit A to the corresponding line in the commit responsible for changing that line.  **Update:** This should also include original file path, in case the line was copied from a different file.

One solution would be for the [`AnnotationLine` type](https://jj-vcs.github.io/jj/latest/templates/#annotationline-type) to have an `original_line` method or field.

It would probably not be printed in the default template, though.

Not sure.

Git does not print this info by default, but it is shown by `git blame -p` and `git blame -n`.

**Additional context**

The guts of `jj file annotate` functionality already have this functionality. Additionally, I suspect `jj absorb` also uses this kind of functionality.

Another thing to consider is that the user might want to jump to the "previous version" of the line as it was before the commit that last changed it. Also, we'd probably want to be able to jump from the `annotate` view to the `diff` view of the relevant commit, and vice-versa.

I'm currently basing my ideas about a "nice `jj file annotate` UI" on `tig blame`. It is a bit unpolished ([the useful key bindings are a bit unintuitive](https://github.com/jonas/tig/issues/1315) and [there are some bugs](https://github.com/jonas/tig/pull/1370)), but is very much worth playing with and using.

I'm not sure when somebody will want to create a nice `jj annotate` TUI; until then this may or may not be relevant.

mernen:
Fixes #6103 and #6809.

# Checklist

If applicable:


lib/src/annotate.rs
```diff
@@ -243,9 +265,28 @@ impl Source {
...
-/// List of commit IDs that originated lines, indexed by line numbers in the
-/// original file.
-type OriginalLineMap = Vec<Result<CommitId, CommitId>>;
+/// List of sources that originated lines, indexed by line numbers in the
+/// starting file.
+type OriginalLineMap = Vec<Result<LineSource, CommitId>>;
+
+/// Information about the origin of an annotated line.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct LineSource {
```
yuja: nit: `LineOrigin` might be better. "source" also has a different meaning.

lib/src/annotate.rs
```diff
@@ -243,9 +265,28 @@ impl Source {
...
+    commit_id: CommitId,
+    line_number: usize,
+}
```
yuja: nit: just make these fields public because this type is purely a data struct.

lib/src/annotate.rs
```diff
@@ -243,9 +265,28 @@ impl Source {
     }
 }
 
-/// List of commit IDs that originated lines, indexed by line numbers in the
-/// original file.
-type OriginalLineMap = Vec<Result<CommitId, CommitId>>;
+/// List of sources that originated lines, indexed by line numbers in the
+/// starting file.
+type OriginalLineMap = Vec<Result<LineSource, CommitId>>;
```
yuja: nit: simply make it store `Result<LineSource, LineSource>`? It's a bit odd that the intermediate value lacks line number whereas the final value knows one.

yuja (commented):
Thanks.

lib/src/annotate.rs
mernen: That was my original plan, but it turns out that this information wasn't accurate enough in my tests. In this case:

https://github.com/jj-vcs/jj/blob/2747dafeb3390bc7a939571f759f3c5184f5a147/lib/src/annotate.rs#L380-L388

The current line number here (first value in the tuple, `_`) doesn't always match the number I expected it to be, and I couldn't figure out another way to obtain it.

mernen: So, the issue isn't so much intermediate values, but the final state when the source possibly falls outside of the domain.

yuja: Ah, good point. I remember I had a similar problem when refactoring the state object in a way that `original_line_map` is constructed later. It might be better to use `Err((_, parent_commit_id))`, but I'm not sure.

PhilipMetzger:
Thanks for tacklling this, it'd be nice if the second and third commit contained some motivation for their changes in the description, otherwise this seems great.

lib/src/annotate.rs
mernen: I've pushed this change experimentally here, for comparison:

- Changing to `Err(parent_commit_id)` in isolation: https://github.com/mernen/jj/commit/3d1d83f75e67d89344eb5b1d5fa3bef579077844
- This PR on top of it: https://github.com/jj-vcs/jj/compare/main...mernen:jj:push-rqzwpuzqrpzm

Observations:

- Line numbers are indeed correctly tracked when we do this, so this fork uses `Result<LineOrigin, LineOrigin>`
- This appears to be closer to what `git blame` does
- This isn't currently represented in the tests, but I found the results somewhat more accurate when there are gaps in the domain

Personally, I'm in favor of changing the behavior for the error case.

mernen:
I've included an example use case in the second commit. Not sure what I could say in the third one, though, since it's the same thing just exposed to the CLI.

lib/src/annotate.rs
yuja: Thanks, let's take this approach then. The data model becomes simpler, and it's nice that the behavior is closer to `git`. Can you also update the doc comments?

yuja:
Just reminder. Maybe you forgot to merge this? We expect contributors to merge PRs when ready.

mernen:
I thought PRs needed explicit approval from someone else? GitHub is still blocking me.

To be clear, this is finished! Just waiting final review/approval. All recent pushes were simply resolving conflicts.

cli/src/commands/file/annotate.rs
```diff
@@ -120,18 +120,22 @@ fn render_file_annotation(
...
+    for (line_number, (line_origin, content)) in annotation.line_origins().enumerate() {
         /* At least in cases where the repository was jj-initialized shallowly,
         then unshallow'd with git, some changes will not have a commit id
         because jj does not import the unshallow'd commits. So we default
         to the root commit id for now. */
-        let commit_id = commit_id.unwrap_or(default_id);
+        let (commit_id, original_line_number) = match line_origin {
+            Ok(origin) => (&origin.commit_id, origin.line_number),
+            Err(_) => (default_id, 0),
+        };
```
yuja: nit: we could use `line_origin.unwrap_or(&default_line_origin)` or something.

mernen: Indeed, now that `LineOrigin` is a plain struct, this is an option. That being said, I'm not sure I really prefer it, or maybe I'm misinterpreting what you had in mind? I've pushed the proposed change here: https://github.com/mernen/jj/commit/9b8fa260e1143ac20adf67807db96189dfc1ed9a

yuja: Yes, I meant that. It is a nasty hack to hide data corruption of unshallow-ed repository, and we'll need to remove `default_line_origin` at some point.

mernen: Hopefully the final question before merging: since I'm repositioning this comment anyway, may I reformat it as `//` lines? A `/*` block seems very unusual for the repo.

martinvonz: Yes, sounds good to me.

PR #6912, opened 2025-07-06
revset: don't resolve symbol expression to multiple revisions

### LINKED ISSUE #5632: FR: Make it an error to use a divergent change id?

When there is a divergent change, one usually doesn't want to use the change id to refer to all of the commits. A recent Discord discussion had an example where one side of the divergence was not shown in the log output. In such cases, it's easy to think that e.g. `jj abandon <change id>` would just abandon the visible commit.

If users rarely want to refer to all the sides of a divergent change, would could just make it an error. We can provide a revset function for when the user actually does want all sides of the divergence.

Keep the current behavior.

yuja:
Closes #5632

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* In revsets, symbol expressions (such as change ID prefix) no longer resolve to
+  multiple revisions, and error out if resolved to more than one revisions. Use
+  `change_id(prefix)` or `bookmarks(exact:name)` to query divergent changes or
+  conflicted bookmarks.
+
 * `jj op abandon` now discards previous versions of a change (or predecessors)
   if they become unreachable from the operation history. The evolution history
   is truncated accordingly.
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -32,15 +32,16 @@ workspace. Use `<name>@<remote>` to refer to a remote-tracking bookmark.
 A full commit ID refers to a single commit. A unique prefix of the full commit
 ID can also be used. It is an error to use a non-unique prefix.
 
-A full change ID refers to all visible commits with that change ID (there is
-typically only one visible commit with a given change ID). A unique prefix of
-the full change ID can also be used. It is an error to use a non-unique prefix.
+A full change ID refers to a visible commit with that change ID. A unique prefix
+of the full change ID can also be used. It is an error to use a non-unique
+prefix or [a divergent change ID][divergent-change].
 
 Use [single or double quotes][string-literals] to prevent a symbol from being
 interpreted as an expression. For example, `"x-"` is the symbol `x-`, not the
 parents of symbol `x`. Taking shell quoting into account, you may need to use
 something like `jj log -r '"x-"'`.
 
+[divergent-change]: glossary.md#divergent-change
 [string-literals]: templates.md#string-literals
 
 ### Priority
@@ -211,8 +212,9 @@ revsets (expressions) as arguments.
 * `none()`: No commits. This function is rarely useful; it is provided for
   completeness.
 
-* `change_id(prefix)`: Commits with the given change ID prefix. It is an error
-  to use a non-unique prefix. Unmatched prefix isn't an error.
+* `change_id(prefix)`: Commits with the given change ID prefix. If the specified
+  change is divergent, this resolves to multiple commits. It is an error to use a
+  non-unique prefix. Unmatched prefix isn't an error.
 
 * `commit_id(prefix)`: Commits with the given commit ID prefix. It is an error
   to use a non-unique prefix. Unmatched prefix isn't an error.
--- a/cli/examples/custom-commit-templater/main.rs
+++ b/cli/examples/custom-commit-templater/main.rs
@@ -15,7 +15,6 @@
 use std::any::Any;
 use std::rc::Rc;
 
-use itertools::Itertools as _;
 use jj_cli::cli_util::CliRunner;
 use jj_cli::commit_templater::CommitTemplateBuildFnTable;
 use jj_cli::commit_templater::CommitTemplateLanguageExtension;
@@ -92,20 +91,17 @@ impl PartialSymbolResolver for TheDigitestResolver {
         &self,
         repo: &dyn Repo,
         symbol: &str,
-    ) -> Result<Option<Vec<CommitId>>, RevsetResolutionError> {
+    ) -> Result<Option<CommitId>, RevsetResolutionError> {
         if symbol != "thedigitest" {
             return Ok(None);
         }
 
-        Ok(Some(
-            RevsetExpression::all()
-                .evaluate(repo)
-                .map_err(|err| RevsetResolutionError::Other(err.into()))?
-                .iter()
-                .map(Result::unwrap)
-                .filter(|id| num_digits_in_id(id) == self.cache.count(repo))
-                .collect_vec(),
-        ))
+        Ok(RevsetExpression::all()
+            .evaluate(repo)
+            .map_err(|err| RevsetResolutionError::Other(err.into()))?
+            .iter()
+            .map(Result::unwrap)
+            .find(|id| num_digits_in_id(id) == self.cache.count(repo)))
     }
 }
 
--- a/cli/src/command_error.rs
+++ b/cli/src/command_error.rs
@@ -23,6 +23,7 @@ use std::sync::Arc;
 use itertools::Itertools as _;
 use jj_lib::absorb::AbsorbError;
 use jj_lib::backend::BackendError;
+use jj_lib::backend::CommitId;
 use jj_lib::config::ConfigFileSaveError;
 use jj_lib::config::ConfigGetError;
 use jj_lib::config::ConfigLoadError;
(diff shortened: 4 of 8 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Breaking changes
 
+* In revset, symbol expressions (such as change ID prefix) no longer resolve to
+  multiple revisions. Use `change_id(prefix)` or `bookmarks(exact:name)` to
+  query divergent changes or conflicted bookmarks.
```
martinvonz: nit: It seems useful to say what the new behavior is (i.e. error if the symbol would have resolved to more than one commit).

I also think "In revset" should be "In revsets". Hopefully a native English speaker can let us know either way.

docs/revsets.md
```diff
@@ -211,8 +212,9 @@ revsets (expressions) as arguments.
 * `none()`: No commits. This function is rarely useful; it is provided for
   completeness.
 
-* `change_id(prefix)`: Commits with the given change ID prefix. It is an error
-  to use a non-unique prefix. Unmatched prefix isn't an error.
+* `change_id(prefix)`: Commits with the given change ID prefix. If the specified
+  change is diverged, this resolves to multiple commits. It is an error to use a
```
martinvonz: nit: I think "is diverged" (here and elsewhere) should be either "is divergent" or "has diverged". I think the former is better since it matches the template function. We also seem to use that form much more frequently in our docs.

martinvonz (approved):
Thanks!

CHANGELOG.md
yuja: Updated.

docs/revsets.md
yuja: Updated to "divergent"

ilyagr:
Thanks for working on this!

This might be a comment for #6814, I'm unsure, but why did you decide to make unmatched prefixes not be an error? If non-unique prefixes are an error, shouldn't unmatched ones be too? Could you also do `present(change_id(abracadabra))`?

This is not a big issue in my mind, but it does trigger my sense of consistency/inconsistency and thus a desire to change it if possible. In my mind "more than one when one is expected" and "none when one is expected" should be treated similarly when possible. (I think it's a good idea to make non-unique prefixes an error, BTW.)

Are you also considering something like `non_unique_change_id(ab)` or `change_id(ab, allow_non_unique=true)`? There, it'd make sense to never error. I've mildly wanted this feature whenever I saw "this prefix is ambiguous" error. (There's also the separate syntax issue similar to the discussion following https://github.com/jj-vcs/jj/issues/6606#issuecomment-3034108515, so it makes sense not to rush with it)

**Update:** To motivate the `non_unique_` functions more clearly, one use-case I have in mind is `non_unique_commit_id(abc) & today()`. Most use-cases would be resolving a commit in a custom set, possibly manually (`jj log -r non_unique_commit_id(abc)`, and then look at the commits to find the one you actually care about and maybe look up its longer prefix).

ilyagr:
The other minor question is whether it's even worth it to have `commit_id(prefix)` while it's an error to use a non-unique prefix. If we do, we can document either that it's the same as just using "prefix", or that it's the same as long as there is at least one commit with the prefix (unless I missed something, and depending on which behavior we prefer).

Or are we considering allowing multiple commits with the same id for the some backend in some (probably exceptional) situations?

martinvonz:
> This might be a comment for #6814, I'm unsure, but why did you decide to make unmatched prefixes not be an error? If non-unique prefixes are an error, shouldn't unmatched ones be too? Could you also do `present(change_id(abracadabra))`?

I feel the same way. See [my comment on that PR](https://github.com/jj-vcs/jj/pull/6814#discussion_r2174939119) for context.

yuja:
> If non-unique prefixes are an error, shouldn't unmatched ones be too?

See also 8db40c7fa278d243cce8bd7b73ba8f4a67489261. Non-unique prefixes are error because we can't return a valid result (= all commits matching the prefix), and doing that wouldn't be useful.

ilyagr:
We might be miscommunicating, I'm not sure.

I agree that non-unique prefixes should be an error, by default. (*Less importantly*: In my mind, this also has to do with UI showing the difference between divergent change ids and non-unique prefixes)

My sense of consistency says that if non-unique prefixes are an error, prefixes that don't match to anything should also be an error, unless there's some important use-case or reason to act differently (which might be explicitly known, or you might feel intuitively this way is better for some reason we haven't pinned down exactly yet).

Separately, I was thinking that it'd be nice to have functions that expand non-unique prefixes as well, for completion reasons for example, or for the user to make an ad-hoc version of resolution within a small set of revisions. I was wondering whether you were thinking of this too, but it sounds like you think that's technically difficult? (**Update:** I added a motivating example to <https://github.com/jj-vcs/jj/pull/6912#issuecomment-3050070834>).

Finally, I'm not sure about the role and docs for `commit_id` as per https://github.com/jj-vcs/jj/pull/6912#issuecomment-3050449618

yuja:
My reasoning is that `bookmarks()`, `tags()`, `change_id()`, and `commit_id()` belong to the same category in which the inner symbol is resolved within a certain namespace. Since `bookmarks(unknown)` doesn't error out, `change_id(unknown)` shouldn't either. However, we can't produce a meaningful result for `change_id(ambiguous)`, so error is unavoidable. We can make all of these functions error out on unknown symbol, but that's a separate issue.

> Finally, I'm not sure about the role and docs for commit_id as per https://github.com/jj-vcs/jj/pull/6912#issuecomment-3050449618

`commit_id(0123)` resolves to commit with an id `0123` even if `0123` were tag/bookmark.

ilyagr:
> However, we can't produce a meaningful result for change_id(ambiguous), so error is unavoidable. 

From my perspective, this is a UI issue. It's pretty clear what `change_id(ambiguous)` would mean, and this could even be useful in some context. We don't *want* to allow this by default because this is *usually* not what the user wants.

In  other words, the operation

- "Give me all commits whose change ids start with this prefix, but only if exactly one change id matches this prefix. Otherwise, report an error!"

is the more useful one, but

- "Give me all commits whose change ids start with this prefix."

still makes sense. (Whether it's feasible to implement or worth implementing is a separate question; I discussed the "worth it?" question a little in the above comments)

And, my point is that, if all else were equal, the operation of "give me all commits whose change ids start with this prefix, but only if exactly one change id matches this prefix" would make slightly more sense to me than "give me all commits whose change ids start with this prefix, but only if at most one change id matches this prefix"

> commit_id(0123) resolves to commit with an id 0123 even if 0123 were tag/bookmark.

That makes sense, thanks! TODO: One of us should remember to add this to the docs.

yuja:
> And, my point is that "give me all commits whose change ids start with this prefix, but only if exactly one change id matches this prefix" makes slightly more sense to me than "give me all commits whose change ids start with this prefix, but only if at most one change id matches this prefix"

It's the same for `bookmarks(exact:name)`, right? If I write `bookmarks(exact:main)`, I want the commit pointed to by the bookmark `main`. (It's a bit weird that the pattern matches substring by default, but there's an idea to make it by default parsed as a glob.)

> > commit_id(0123) resolves to commit with an id 0123 even if 0123 were tag/bookmark.
> 
> That makes sense, thanks! TODO: One of us should remember to add this to the docs.

It's documented as "commits with the given commit ID prefix." Maybe you get confused because we're discussing these functions in context of divergence.

ilyagr:
> It's the same for bookmarks(exact:name), right? If I write bookmarks(exact:main), I want the commit pointed to by the bookmark main.

In my mind, a better comparison would be with a theoretical `bookmarks(prefix:push)`. That shows the different desired behavior: results for multiple bookmark names for `bookmarks(prefix:push)` vs an error for `change_id(xyz)` if multiple changes have this prefix. Then, the rest of https://github.com/jj-vcs/jj/pull/6912#issuecomment-3050831997 applies.

I'm guessing that, perhaps, in your mind `xyz` is expanded to the full change id, or even to the revision it points to (that is, an object of a theoretical type `Revision`), before being passed to the `change_id` function? (That would explain why the analogy with `bookmarks(exact:...)` seems natural to you). I don't think most people would find it natural to think this way, at least if they don't know the details of how the revset language works. To me, this point of view would also make it more confusing that, if a bookmark `abc` exists, `commit_id(abc)` does not expand `abc` to its value *before* the result is passed to the function.

> It's documented as "commits with the given commit ID prefix." Maybe you get confused because we're discussing these functions in context of divergence.

Indeed! The docs are correct. I just think that I'm not the only person that might get confused about that, and an example would help. I just sent https://github.com/jj-vcs/jj/pull/6948.

yuja:
> I'm guessing that, perhaps, in your mind `xyz` is expanded to the full change id,

Yes, an unambiguous commit/change ID prefix identifies a thing, so it's closer to exact bookmark/tag name. Even if it didn't, `bookmarks(pattern)` nor `change_id(prefix)` doesn't error out when no matches found.

apoelstra:
This is a pretty big usability hit for divergent commits. Previously to see all the versions of a divergent commit you could just use `jj log -r <prefix>` where the prefix could usually be just a couple of letters. Now you get a multi-line error telling you to type out `change_id(<prefix>)` instead, which is 5 times as many characters and also requires shell quoting.

martinvonz:
The thinking was that divergent commits are not encountered frequently so it's okay to make it a bit harder to work with them.

PR #6921, opened 2025-07-07
index: pack length parameters into struct

yuja:
# Checklist

If applicable:


yuja:
> Btw, I think it would be nice to some day remove `default_index`'s requirement that the ids have fixed length. Not very important at all, though.

Hmm, by using maximum possible length or something? I don't think we'll need to support ids of arbitrary lengths.

martinvonz:
> Hmm, by using maximum possible length or something? I don't think we'll need to support ids of arbitrary lengths.

I mean by making the index allow variable-length keys. I think many key-value stores allow that. But changing that would be a large change.

PR #6931, opened 2025-07-08
cli `diff`: warn if combining diffs of non-linear set of revisions

### LINKED ISSUE #6723: `jj diff -r 'all()'` works without warning

1. Make sure `ui.always-allow-large-revsets` is set to the default value of `false`
1. Run `jj diff -r ::`

## Expected Behavior

This should fail. `jj diff -r 'all:all()'` should continue to work (it's a bit slow 😸  ).

## Actual Behavior

Both succeed.

This is especially confusing and problematic if you do `jj diff -r conflicted_change_id`. I also find it inconsistent: while `diff` is not mutating, providing confusing information without an explanation, so I think it should be near the top of the list of commands that require `all:` (while we haven't figured out some replacement for it, which is an orthogonal issue).

## Alternatives

I also wonder if people think there are other ways to improve the situation. For example, if `jj diff -r 'all()'` explained which commits it's showing the diff for, things would be less bad. Should we add such a line to the output when diffing complicated expressions? Always?

cc #5632 #6016.

ilyagr:
Fixes #6723

----------

I have run into confusing output for `jj diff -r @-` a few times since I last complained.

Other options would be to disallow non-linear revsets unless an option is passed to `jj diff`, require `all:`, or require `all:` but only if the revset is non-linear.

# Checklist

If applicable:

- [ ] (TODO) I have updated `CHANGELOG.md`
- n/a I have updated the documentation (`README.md`, `docs/`, `demos/`)


cli/src/commands/diff.rs
```diff
@@ -207,6 +212,14 @@ pub(crate) fn cmd_diff(
     }
 
     ui.request_pager();
+    if show_nonlinear_revset_warning {
+        // The warning must be passed to the pager; otherwise it is very easy to miss
+        // when the pager is used.
+        writeln!(
+            ui.warning_default(),
+            "Showing combined diff of multiple revisions that do not form a linear sequence"
```
yuja: Maybe better to say there are multiple heads (or roots), and they got merged? It seems not easy to understand what "form a linear sequence" means. It might be also good to print commit summary of multiple heads/roots, but I'm not sure.

martinvonz:
> cc @martinvonz

Thanks. I had missed the discussion in the linked issue. This warning seems reasonable to me.

cli/src/commands/diff.rs
ilyagr: Indeed, the message is imperfect, I'm having some trouble coming up with a better one.

I don't think the user really cares about the multiple heads/roots. What matters more is that when there is a single head & root, the diff can be reasoned about as the diff between two snapshots. When there are multiples, you'd need to rebase some revisions before you can reason about the diff that way.

We could still mention heads/tails to help the user reason about the situation, but I think we'd need to add other words to explain why the user should care.

Another wording that comes to mind is "merging unrelated diffs", but I'm not sure that "unrelated" is clearer than "does not form a linear sequence".

I'm still thinking. Perhaps someone will have a fresh idea, or perhaps we'll have to go with the best of bad options.

ilyagr: Perhaps we can add some actionable advice to whatever explanation we come up with. Something like "If you intended to see the diff of a single revision or a diff between two revisions, check your revset with `jj log -r`". But that makes the warning long...

martinvonz: I think the current message is good enough. I don't think I would find a message about heads and roots clearer. I don't think the average user would either. Something like "Showing combined diff of a set of revisions with multiple heads and/or roots" is more verbose and not as immediately clear to me.

Of course it's also not technically correct to say that it has to be linear (we allow internal merges) but I don't think we need to explain that part.

> Perhaps we can add some actionable advice to whatever explanation we come up with.

It does make it a bit long. I would skip it at least to start with.

yuja: How about this?

"Warning: Showing combined diff between multiple head and root revisions."

The help text mentions "heads" and "roots", but we don't explain what "a linear sequence" means.

martinvonz: Ah, good point about the help text. That does mention "linear chain" (not "linear sequence"), so that should also be fine. It's just used as an example in the help text, though. Since we already mention roots and heads in the help text, I'm also fine with using your proposal even though I think it's less obvious to the user.

yuja: The help says "if you have a linear chain of revisions A..D, then `jj diff -r B::D` equals `jj diff --from A --to D`. Multiple heads and/or roots are supported", so I assumed the multiple heads/roots case can be considered "a linear chain" between multiple heads and roots (if there aren't no other merges/forks in between.)

FWIW, `--from`/`--to` doesn't support more than one revisions, but I think this restriction can be lifted if needed.

martinvonz: Oh, I see. That was not what I meant by that text. The first sentence there was meant as an example. The second sentence was meant to apply generally, not implying that we use the term "linear chain" for a range with multiple heads and/or root. Do you think it helps if I send a PR that simply separates those two sentences into two paragraph? Or is there a better way to clarify it?

yuja: Oops, I didn't mean the help text is unclear. I just mean it could be read as my interpretation, and that's not so wrong in the context of this help.

That said, I think it's also good to separate paragraphs and add more details about multiple heads/roots.

ilyagr: My preferred solution to this would be to define "linear sequence of revisions" in the docs. I pushed a new version of this PR that reworks the `jj diff -r` docs in a second commit.

> That said, I think it's also good to separate paragraphs and add more details about multiple heads/roots.

+1, good idea!

> "Warning: Showing combined diff between multiple head and root revisions."

I agree with Martin that, IMO, most users would find this confusing, for reasons I described in https://github.com/jj-vcs/jj/pull/6931/files#r2193084272.

Yuya's wording also makes it seem that there are both multiple heads *and* multiple roots. Finally, if there were in fact both multiple heads and multiple roots, I'm not sure describing what's shown as a diff between these is correct.

Martin's original version of "Showing combined diff of a set of revisions with multiple heads and/or roots" solves some of these problems, but is also a bit awkward.

-----

All of that said, however, I'm willing to remove the second commit and/or go with any text Yuya considers acceptable. Some warning, even if confusing, will be less confusing in this situation than no warning.

I could then put the remaining changes into a separate PR (e.g., I think we all agree that something along the lines of the second commit would be worth doing regardless of what the warning message says, though my version may or may not be a good start depending on what else we decide), or I could abandon them.

cli/src/commands/diff.rs
```diff
@@ -52,15 +52,24 @@ use crate::ui::Ui;
 pub(crate) struct DiffArgs {
     /// Show changes in these revisions
     ///
+    /// If a single merge commit is given, this shows changes *from* the
+    /// automatic merge of the contents of all of its parents *to* the contents
+    /// of the revision itself.
+    ///
     /// If there are multiple revisions, then the total diff for all of them
     /// will be shown. For example, if you have a linear chain of revisions
```
yuja, line 60: It seems confusing to define "linear chain" and "linear sequence". Perhaps, "linear chain" is stricter than "linear sequence", but I'm not sure.

cli/src/commands/diff.rs
yuja: I'm not sure why "linear sequence" is easier to understand than multiple heads/roots, but maybe I'm a minority. Since Martin and you agree, it's probably better to use "linear sequence" or something like that.

cli/src/commands/diff.rs
ilyagr: That's fair. I was hoping it's not too terrible like this, but I should look it over again. Perhaps I'll try rewriting it once more, or perhaps you or Martin will have an idea.

The way its used here, "linear chain" is stricter as you said.

martinvonz: I'd say let's avoid using two such similar terms. I'd use "linear chain" since we already have that, but I'm fine if you want to replace the existing instance(s) of "linear chain" by "linear sequence" (or "linear range" or perhaps something else).

ilyagr: OK, changing everything to "linear chain" sounds reasonable. How about this?

I think I like "linear range" even better, but I'm not sure "linear range of revisions A..D" is as clear for e.g. the FAQ. We could change it to "linear range" later.

CHANGELOG.md
```diff
@@ -54,6 +54,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   [`Operation`](docs/templates.md#operation-type) type in the templating
   language.
 
+* `jj diff -r multiple_revisions` will warn if `multiple_revisions` do not form
+  a line in the ancestry graph. Such calls are often accidental, e.g. `jj diff
+  -r @-` where the user doesn't realize they are on a merge commit.
```
martinvonz, line 65: I'm fine with having this entry but I'm not sure whom it will help. Users who accidentally run `jj diff -r @-` when on a merge commit would notice the feature anyway. Maybe there's a small chance that some users who have done that in the past will read the release notes and decide to upgrade sooner and thereby benefit from the hint. As I said, I'm fine with having it here if you think it's useful.

cli/src/commands/diff.rs
```diff
@@ -52,15 +52,25 @@ use crate::ui::Ui;
...
-    /// A..D, then `jj diff -r B::D` equals `jj diff --from A --to D`. Multiple
-    /// heads and/or roots are supported, but gaps in the revset are not
-    /// supported (e.g. `jj diff -r 'A|C'` in a linear chain A..C).
+    /// `root -> A -> B -> C -> D`, then `jj diff -r B::D` equals `jj diff
```
martinvonz, line 61: nit: perhaps use "main" instead of "root" to make the example better match a common setup?

martinvonz, line 61: nit: should we switch from `B::D` to `A..D` to make it a closer match to `--from A --to D`? Could say it like this:

> then `jj diff -r A..D` (or `jj diff -r B::D`) equals `jj diff --from A --to D`

cli/src/commands/diff.rs
```diff
@@ -52,15 +52,25 @@ use crate::ui::Ui;
...
+    /// --from A --to D`.
     ///
-    /// If a revision is a merge commit, this shows changes *from* the
-    /// automatic merge of the contents of all of its parents *to* the contents
-    /// of the revision itself.
+    /// Any revset with a single head and a single root will behave similarly;
+    /// the diff of such a revset will be the diff between the head and the
+    /// root. For the purpose of this command, we call any such revsets "linear
+    /// chains of revisions".
```
martinvonz, line 67: "single head and single root" would also match this case:

```
D
|
C
|\
B |
| |
A |
| |
| Y
|/
X
```

Here, `jj diff -r B::D` will currently include the changes in Y. That seems like a bug. I'll send a PR to fix that.

ilyagr: Indeed, to me this does indeed seem like a clear bug and that it should be equivalent to `jj diff --from A --to D`. 

Nice catch!

(**Update:** As per following comments, I now think you meant the opposite. It's still a nice catch, but I'm much less sure what to do about it now)

CHANGELOG.md
ilyagr: Makes sense, I'll remove it.

cli/src/commands/diff.rs
ilyagr: (**Update:** This comment is mostly obsolete in light of the following ones)

Ah, I see, the issue is that `diff -r C` wouldn't show the same thing as `diff --from B --to C`, but rather something like ~~a difference `(diff --from B --to C) - (diff --from A --to Y)`~~ (!, unless I mixed something up; this is reminiscent of the "rotated graph" idea you once mentioned). (**Update:** No, my equation seems wrong; I think you had the right idea above, the diff of `Y`, or more generally `diff --from fork_point(B|Y) --to Y`, that seems to be added. The point in the next paragraph still stands.)

So, adding up `diff -r C`, `diff -r B`, and `diff -r D` wouldn't give us `diff --from A --to D`.

I still think that `diff --from A --to D` is what the user would want for `diff -r B::D`; adding up diffs "along the line" seems to make the most sense. (**Update:** Actually, after thinking about it a bit more in https://github.com/jj-vcs/jj/pull/6931#discussion_r2218088465, I see problems whatever we do) 

I'm now more confused about what `diff -r B::D|Y` means. It should probably be the sum of `diff --from A --to D` and `jj diff -r Y`, but is this actually consistent with other ways to compute it?

I also wonder whether there are other weird cases. Is there actually a consistent way of defining `jj diff -r revset` for an arbitrary revset if we want it to behave sanely for revsets with one head/root? I'd guess so, but I'm not sure.

ilyagr: Actually, I think we already have it function with `jj diff -r B::D` being the same as `jj diff --from A --to D`.

```console
$ jj log --summary -r @::owx -T "change_id.shortest(6) ++ ' ' ++ description"
○    owxopr add x
├─╮  A x
│ ○  qukvlr add c
│ │  A c
│ ○  vlsmyk add b
│ │  A b
○ │  uonxzo add a
├─╯  A a
@  kkxutv
│
~
$ jj diff --from quk- --to owx -s
A a
A c
A x
$ jj diff -r quk -r owx -s
A a
A c
A x
```

So, it's already not equivalent to taking `diff -r` of each commit:

```console
$ jj diff -r quk -s
A c
$ jj diff -r owx -s
A x
```

ilyagr: I'm now realizing you might be thinking of making `jj diff -r quk -r owx` show just "c" and "x". I think this is counter-intuitive for people who do `jj diff -r quk::owx`. If we do that, we maybe want to show a warning? Or find some way to reason about it?

Granted, the opposite decision could be confusing to people who write `jj diff -r quk -r owx`.

If we do define `jj diff -r revset` as the sum of individual `diff -r`-s, I think the statement about `jj diff -r revset` being the same as a diff between two commits depends on whether `revset|revset-` is linear. If it is, it's probably equivalent to `jj diff --from roots(revset-) --to head(revset)`.

I'm not sure how to explain this to a user than gets an unexpected result because `revset|revset-` has complicated topology unexpectedly. My first instinct would be to error in all even mildly confusing cases.

martinvonz: > I'm now realizing you might be thinking of making `jj diff -r quk -r owx` show just "c" and "x".

Yes, I think of `jj diff -r` as showing the combined diff for all specified commits.

> I'm not sure how to explain this to a user than gets an unexpected result because `revset|revset-` has complicated topology unexpectedly. My first instinct would be to error in all even mildly confusing cases.

This PR adds a warning in some cases. Maybe we can extend that to any other cases we consider possibly confusing? It seems ugly to warn about mildly confusing cases and error out in what we consider more confusing cases. But I'm not sure if that's what you meant.

ilyagr: > Yes, I think of jj diff -r as showing the combined diff for all specified commits.

That makes sense, especially for somebody that does `jj diff -r B|C|D` without thinking of the topology of these revisions. However, I think this behavior would also be quite confusing who somebody that does `jj diff -r B::D` without remembering that `C` was a merge commit.

So, I still find that the current behavior makes more sense in simple cases of how I imagine people actually use `-r`. But you are also right. Both in theory and in complicated cases, the current behavior seems to make a lot less sense than the behavior you have in mind.

At this point, I wonder whether we'd have been better off implementing `jj diff --branch-from main --to abc` instead of suggesting `jj diff -r main..abc`. (Though, we'd still have to figure out what to do if the topology of `main..abc` is non-trivial. I'm hoping this is doable, even if there's not a unique commit in `::main` that's closest to `abc`)

> It seems ugly to warn about mildly confusing cases and error out in what we consider more confusing cases. But I'm not sure if that's what you meant.

I'm not sure what's best. In that particular comment, I was thinking we'd make it an error in all even somewhat confusing cases, including upgrading the warning from this PR to an error.

But I'm not sure how to address it fully, now that we have your example. Making it an error might help, but doesn't solve the problem. I had plenty of difficulty finding a way to explain the much simpler case from this PR, implementing a good warning (or error) for your example seems hard.

One problem is that `jj diff -r B::D` and `jj diff -r B -r C -r D` indicate a user with different things in mind. If we did allow ourselves to distinguish between these, we could say that `jj diff -r B::D` means the same as `jj diff --from A --to D` while `jj diff -r B -r C -r D` means the other thing. Then we'd probably want to error if any single `-r` argument doesn't have a revset with a single head and tail, similarly to how `all:` used to work (but without an override maybe?). This would have the problem that `jj diff -r B|C|D` would mean something different than `jj diff -r B -r C -r D`, but we could hope people won't use the former.

Maybe some better solution will come to mind another day.

ilyagr: Actually, there might be an elegant way to think of this after all.

In my example https://github.com/jj-vcs/jj/pull/6931#discussion_r2218081075, `jj diff --from quk- --to owx` is equivalent to `jj diff -r quk-..owx` (with the definition of -r you suggested), and the same might be true more generally. (**Update** with some assumptions, though)

Reusing that example, we have:

```console
$ # The only reason I am calling the "from" revision `quk-` instead
$ # of `vls` is for consistency with the previous comment.
$ jj diff --from quk- --to owx --summary
A a
A c
A x
$ jj log -r quk-..owx   -T "change_id.shortest(6) ++ ' ' ++ description ++ diff.summary()"
○    owxopr add x
├─╮  A x
│ ○  qukvlr add c
│ │  A c
│ ~
│
○  uonxzo add a
│  A a
~
```

The diffs in the log in the second command add up to the diff in the former command.

We could explain this; at the very least, the statement that `jj diff -r revset` behaves predictably when `revset` is of the form `A..B` is something we could put in help messages (after verifying it) if/once we change the behavior as you suggested.

I'm not sure, but perhaps there is some way to check whether `revset` is of the form `A..B` for some A and B. (We have a fun algorithmic problem now! I'm not sure whether this is possible in any performant way; it might also be smart to allow A and B to consist of multiple revisions, or perhaps any revset could be decomposed into `A..B|C..D|...`). If we could do this, we could warn whenever `revset` is not of that form.

**Update:** There should be a way to check whether `revset` is of the form A..B for some A and B. I think A would have to be in `roots(revset)-`, so as long as that's not massive, things are probably OK. (**Update**: Let's assume A is an ancestor of B; not sure if that's sufficient) Actually, it's likely that `A=heads(~((A..B)..))` and `B=heads(A..B)`. OTOH, it's getting late, so I could have easily missed something; I also don't yet have a proof that `jj diff -r A..B` is always the same as the diff from A to B.

ilyagr: Here's a shape of a potential medium-term plan I can see. I'm still not 100% sure all of the assumptions that go into it are correct, but I think they likely are.

1. Change the behavior so that the merge commit are treated as Martin suggested.

2. Make sure `jj diff x..y` can always be computed, e.g. that it doesn't hit the check for "a revset with gaps" (Actually, maybe this is true by definition of `..`?). I think it can and should have the property that, at least if `x` and `y` are single commits, `jj diff x..y` is the same as `jj diff --from fork_point(x|y) --to y`, which reduces to `jj diff --from x --to y` if `x` is an ancestor of `y`.

    I'm not actually 100% sure what the correct behavior is with regard to the base of the final diff `jj diff` shows for `jj diff -r arbitrary_revset`, but I'm hoping there's some answer (at least for some subset of all revsets that includes all the `x..y`s). If there is an answer, I'm hoping it can have the above property for `jj diff -r x..y`.

4. Warn the user of `jj diff -r revset` if `revset` expands to multiple commits and doesn't textually contain `..`. The warning can be slightly different if the revset doesn't contain `..` and contains `::`, explaining (either in the warning or in some help text the warning refers to) that while, in simple cases, `jj diff -r x::y` is the same as `jj diff -r x-..y` and as `jj diff --from x- --to y`, the first of these will be wrong if any commit between `x` and `y` is a merge commit.

The goal would be to educate the users that they should switch from `jj diff -r x+::y` to `jj diff -r x..y` and, more generally, that if they are doing `jj diff -r` with something other than `x..y`, they are on their own :).

This textual check is a bit ugly, but I think it'd ultimately be both easier and more helpful than checking whether `revset` can be expressed as `x..y` for some `x` and `y`. We could look into that too, but one problem is that, I think, `jj diff -r x..y` might make sense even if `x` or `y` expands to multiple commits, and I don't want figuring out which exact cases to check and how they can be presented to the user to be a blocker.

We could also outright forbid `jj diff -r` with something other than `x..y`, or replace `jj diff -r multiple_revisions` with something like `jj diff --from-branch x --to y`, but I feel this might be too breaking at this point (?), and perhaps too restrictive.

----------

Steps 1 and 2 could be difficult. I'm still not sure what the plan should be until those are done. Perhaps we can go with something like this PR but make it clear that the focus on "linear revsets" is something that will change soon, since they are the wrong concept in the presence of merge commits.

ilyagr: I tried to convert these ideas into a PR: #7080. It includes both a short-term fix for #6723 and my thoughts of how we could proceed.

ilyagr: My hope is that, with #7080, changing the behavior of `diff -r` will not be a blocker for #6723.

PR #6932, opened 2025-07-08
Buffer stderr from fix tools

jennings:
The stderr from fix tools is currently all mixed together. I think the problem is worse on Windows: I regularly see line noise as multiple stderr streams are interleaved character-by-character. On Windows, this is _extremely_ slow and can make `jj fix` very annoying, if not unusable.

This change buffers the stderr output line-by-line, so at least there's some hope of reading it. I'm also prefixing the file name.

Ways this could be improved:

- Maybe prefix the commit ID and/or tool name as well as the file name: `path/to/file:clippy:a8b9c3d:This is some output`
- Instead of releasing the mutex holding the main process's Stderr after emitting each line, it would be nice to emit all available lines from a process before releasing the mutex. However, I didn't see a way to `read_line_if_ready()` on a BufReader. If there's an obvious way to do this, I'd love to improve it.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -129,6 +129,7 @@ pub(crate) fn cmd_fix(
 ) -> Result<(), CommandError> {
     let mut workspace_command = command.workspace_helper(ui)?;
     let workspace_root = workspace_command.workspace_root().to_owned();
+    let path_converter = workspace_command.path_converter().to_owned();
     let tools_config = get_tools_config(ui, workspace_command.settings())?;
     let root_commits: Vec<CommitId> = if args.source.is_empty() {
         let revs = workspace_command.settings().get_string("revsets.fix")?;
@@ -145,7 +146,15 @@ pub(crate) fn cmd_fix(
 
     let mut tx = workspace_command.start_transaction();
     let mut parallel_fixer = ParallelFileFixer::new(|store, file_to_fix| {
-        fix_one_file(ui, &workspace_root, &tools_config, store, file_to_fix).block_on()
+        fix_one_file(
+            ui,
+            &workspace_root,
+            &path_converter,
+            &tools_config,
+            store,
+            file_to_fix,
+        )
+        .block_on()
     });
     let summary = fix_files(
         root_commits,
@@ -179,6 +188,7 @@ pub(crate) fn cmd_fix(
 async fn fix_one_file(
     ui: &Ui,
     workspace_root: &Path,
+    path_converter: &RepoPathUiConverter,
     tools_config: &ToolsConfig,
     store: &Store,
     file_to_fix: &FileToFix,
@@ -201,6 +211,7 @@ async fn fix_one_file(
             match run_tool(
                 ui,
                 workspace_root,
+                path_converter,
                 &tool_config.command,
                 file_to_fix,
                 &prev_content,
@@ -234,6 +245,7 @@ async fn fix_one_file(
 fn run_tool(
     ui: &Ui,
     workspace_root: &Path,
+    path_converter: &RepoPathUiConverter,
     tool_command: &CommandNameAndArgs,
     file_to_fix: &FileToFix,
     old_content: &[u8],
@@ -262,6 +274,12 @@ fn run_tool(
     tracing::debug!(?command, ?output.status, "fix tool exited:");
     if !output.stderr.is_empty() {
         let mut stderr = ui.stderr();
+        writeln!(
+            stderr,
+            "{}:",
+            path_converter.format_file_path(&file_to_fix.repo_path)
+        )
+        .ok();
         stderr.write_all(&output.stderr).ok();
         writeln!(stderr).ok();
     }
--- a/cli/src/commands/fix.rs
+++ b/cli/src/commands/fix.rs
@@ -145,7 +145,7 @@ pub(crate) fn cmd_fix(
 
     let mut tx = workspace_command.start_transaction();
     let mut parallel_fixer = ParallelFileFixer::new(|store, file_to_fix| {
-        fix_one_file(&workspace_root, &tools_config, store, file_to_fix).block_on()
+        fix_one_file(ui, &workspace_root, &tools_config, store, file_to_fix).block_on()
     });
     let summary = fix_files(
         root_commits,
@@ -177,6 +177,7 @@ pub(crate) fn cmd_fix(
 /// TODO: Better error handling so we can tell the user what went wrong with
 /// each failed input.
 async fn fix_one_file(
+    ui: &Ui,
     workspace_root: &Path,
     tools_config: &ToolsConfig,
     store: &Store,
@@ -198,6 +199,7 @@ async fn fix_one_file(
         read.read_to_end(&mut old_content).await?;
         let new_content = matching_tools.fold(old_content.clone(), |prev_content, tool_config| {
             match run_tool(
+                ui,
                 workspace_root,
                 &tool_config.command,
                 file_to_fix,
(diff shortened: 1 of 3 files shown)


martinvonz:
@hooper: Could you review this?

cli/src/commands/fix.rs
```diff
@@ -245,13 +259,27 @@ fn run_tool(
...
+            for line in reader.lines() {
+                let Ok(line) = line else {
+                    continue;
+                };
+                if let Ok(stderr) = &mut stderr_mutex.lock() {
+                    (*stderr).write_all(line.as_bytes()).ok();
+                    (*stderr).write_all(b"\n").ok();
+                }
+            }
+        });
```
yuja: Maybe better to collect stderr outputs to `Vec<u8>`? I don't think the latency would matter. If we don't want to buffer everything, I think we can send per-file stderr output over the channel, and print them by the main thread.

yuja (commented):
There's a previous attempt: https://github.com/jj-vcs/jj/pull/6582

martinvonz:
> There's a previous attempt: #6582

Thanks. I had already forgotten that. @winterqt, FYI.

cli/src/commands/fix.rs
jennings: I thought about that but didn’t know if it would be okay to buffer the entire stderr steam of a fix tool invocation. If that’s not a concern, then I like that change because it avoids interleaving output entirely.

jennings: I updated it to collect stderr from each process into a Vec<u8>, then acquire the mutex and write it all at once.

cli/src/commands/fix.rs
```diff
@@ -144,8 +147,16 @@ pub(crate) fn cmd_fix(
         .to_matcher();
 
     let mut tx = workspace_command.start_transaction();
+    let stderr_mutex = Mutex::new(std::io::stderr());
```
yuja: We can `ui.stderr()` in place of `stderr_mutex.lock()`. The stderr stream should be locked there. The pager stream doesn't provide the same guarantee, but I think that's okay.

A better option is to send the stderr output back to the main function, and print it there.

cli/src/commands/fix.rs
```diff
@@ -245,13 +259,24 @@ fn run_tool(
...
         });
+        s.spawn(move || {
+            let mut buf: Vec<u8> = Vec::new();
+            stderr.read_to_end(&mut buf).ok();
+            if let Ok(stderr) = &mut stderr_mutex.lock() {
+                for line in buf.lines_with_terminator() {
+                    (*stderr).write_all(line).ok();
+                }
+            }
+        });
```
yuja: Simply rely on `child.wait_with_output()` instead of manually capturing stderr.

cli/src/commands/fix.rs
```diff
@@ -271,8 +271,19 @@ fn run_tool(
...
             stderr.read_to_end(&mut buf).ok();
+            let prefix = {
+                let path_converter = RepoPathUiConverter::Fs {
+                    cwd: "".into(),
+                    base: "".into(),
+                };
+                format!(
+                    "{}:",
+                    path_converter.format_file_path(&file_to_fix.repo_path)
+                )
```
yuja: I don't think we need to add file name to each line. Just print file name once, then stderr lines?

cli/src/commands/fix.rs
```diff
@@ -271,8 +271,19 @@ fn run_tool(
         s.spawn(move || {
             let mut buf: Vec<u8> = Vec::new();
             stderr.read_to_end(&mut buf).ok();
+            let prefix = {
+                let path_converter = RepoPathUiConverter::Fs {
+                    cwd: "".into(),
+                    base: "".into(),
+                };
```
yuja: For UI message, I think we should use `workspace_command.path_converter()` or `.format_file_path()`.

cli/src/commands/fix.rs
jennings: Could you give me a brief description of what "send the stderr back to the main function" would look like? My lack of Rust experience might be keeping an obvious/idiomatic solution from being apparent.

- Do you mean send all _n_ Vec\<u8\> instances back to the `cmd_fix` function?
  - If so, would `fix_files` return them?
  - Is there any concern about having to have potentially dozens or hundreds of stderr outputs buffered simultaneously?

I originally thought I would pass `workspace_command` as an argument to `fix_one_file`, but it's not Sync: "`std::cell::OnceCell<IdPrefixContext>` cannot be shared between threads safely", contained inside the ReadonlyRepo inside WorkspaceCommandHelper.

yuja: I thought `mpsc::channel()` could be used in order to report status back, but apparently that's not easy. `fix::fix_files()` carries objects that cannot be moved across thread boundary. We can of course use the channel as `Vec<Vec<u8>>` (as we do in `TreeState::snapshot()`), though.

https://doc.rust-lang.org/stable/std/sync/mpsc/fn.channel.html

EDIT: we can instead move the writer side to thread, but I'm not sure if you like it:
```rust
let status = thread::scope(|s| {
    let (sender, receiver) = mpsc::channel();
    s.spawn(move || {
        for output in receiver {
            print(output);
        }
    });
    ...
    fix_files(...).block_on()
})?;
```

> * If so, would `fix_files` return them?

I think `fix_files()` and `FileFixer` should have a functionality to report non-fatal errors, but I don't expect it will be implemented by this PR. That's a separate issue.

> * Is there any concern about having to have potentially dozens or hundreds of stderr outputs buffered simultaneously?

I don't have a concern about the memory usage, but the latency could matter.

So, pass `&Ui` and `&RepoPathUiConverter` for now?

cli/src/commands/fix.rs
jennings: We're now calling `workspace_command.path_converter()` in `fn cmd_fix` and passing that in.

cli/src/commands/fix.rs
jennings: Done.

cli/src/commands/fix.rs
jennings: I'm now passing in `&Ui` and `&RepoPathUiConverter` and using those to lock stderr and format file names. That's much cleaner, thanks.

cli/src/commands/fix.rs
```diff
@@ -256,6 +272,15 @@ fn run_tool(
     })
     .unwrap()?;
     tracing::debug!(?command, ?output.status, "fix tool exited:");
+    if !output.stderr.is_empty() {
+        let prefix = format!(
+            "{}:\n",
+            path_converter.format_file_path(&file_to_fix.repo_path)
+        );
+        let mut stderr = ui.stderr();
+        stderr.write_all(prefix.as_bytes()).ok();
```
yuja: nit: `writeln!(stderr, ...)`

cli/src/commands/fix.rs
```diff
@@ -256,6 +272,15 @@ fn run_tool(
...
+        stderr.write_all(&output.stderr).ok();
+    }
```
yuja, line 285: maybe better to do `writeln!(stderr)` if stderr isn't terminated by newline?

jennings: I can use `writeln!` for the file name, but it's not guaranteed stderr is UTF-8, and I think using `writeln!` needs the format args to be Display or Debug, unless I'm missing something.

What do you think about just emitting a `\n` after we write to stderr? Like this:

```rust
writeln!(stderr, "{}:", path_converter.format_file_path(&file_to_fix.repo_path)).ok();
stderr.write_all(&output.stderr).ok();
stderr.write_all(b"\n").ok(); // or writeln!(stderr, ""), if that's more idiomatic, but it looks odd to me
```

cli/src/commands/fix.rs
jennings: Ah, thanks. Will fix that.

cli/src/commands/fix.rs
yuja: `format_file_path()` returns a `String`, so it can be passed to `writeln!(..)`.

> stderr.write_all(b"\n").ok(); // or writeln!(stderr, ""), if that's more idiomatic, but it looks odd to me

`writeln!(stderr)` is often used.

PR #6936, opened 2025-07-08
Make `jj file annotate` faster with slow backends

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/annotate.rs
+++ b/lib/src/annotate.rs
@@ -276,9 +276,8 @@ fn process_commits(
         .evaluate(repo)?;
 
     state.num_unresolved_roots = 0;
-    for node in revset.iter_graph() {
+    let node_iter = revset.iter_graph().map(|node| {
         let (commit_id, edge_list) = node?;
-
         let mut parents = vec![];
         for parent_edge in edge_list {
             let commit = repo.store().get_commit(&parent_edge.target)?;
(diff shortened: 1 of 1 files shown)


lib/src/annotate.rs
```diff
@@ -270,7 +278,22 @@ fn process_commits(
...
+            let tree = commit.tree()?;
+            let file_value = tree.path_value(file_name)?;
+            let parent = FileVersion {
+                commit_id: parent_edge.target,
+                file_name: file_name.to_owned(),
+                file_value,
+                is_missing: parent_edge.edge_type == GraphEdgeType::Missing,
+            };
+            parents.push(parent);
+        }
```
yuja: This means parent trees (and `MaterializedTreeValue`s) can be fetched more than once if there are multiple branches. (Before, redundant fetches were rejected by `commit_source_map.entry(parent_id)`.) That seems not good for merge-heavy Git repositories.

I think we can instead add a separate prefetch buffer which is backed by a list of incomplete `CommitId -> MaterializedTreeValue` futures and a list/map of completed-yet-popped `MaterializedTreeValue`. I guess `FuturesUnordered` can be used for this purpose, but I'm not sure how complicated the implementation would be.

martinvonz: Makes sense. This PR was just my spare time project, so I have less time to work on it. (It's not a priority for Google because we can point users to our web-based blame viewer.)

jmr: I have tried this in https://github.com/jj-vcs/jj/pull/9674.

PR #6937, opened 2025-07-08
cli: allow "large" revsets by default

yuja:
# Checklist

If applicable:

### DIFF
--- a/cli/tests/test_rebase_command.rs
+++ b/cli/tests/test_rebase_command.rs
@@ -777,7 +777,6 @@ fn test_rebase_multiple_destinations() {
     Hint: The revset `b|c` resolved to these revisions:
       royxmykx c12952d9 c | c
       zsuskuln d18ca3e8 b | b
-    Hint: Prefix the expression with `all:` to allow any number of revisions (i.e. `all:b|c`).
     [EOF]
     [exit status: 1]
     ");
@@ -793,6 +792,13 @@ fn test_rebase_multiple_destinations() {
     ]);
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
+    Warning: In revset expression
+     --> 1:1
+      |
+    1 | all:b|c
+      | ^-^
+      |
+      = Multiple revisions are allowed by default; `all:` is planned for removal
     Rebased 1 commits to destination
     [EOF]
     ");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Deprecations
 
+* The `all:` revset modifier and `ui.always-allow-large-revsets` setting is
+  planned to be removed in a future release.
+  [#6016](https://github.com/jj-vcs/jj/issues/6016)
+
 ### New features
 
 * `jj workspace list` now accepts `-T`/`--template` option to customize its output via templates.
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -1563,13 +1563,9 @@ to the current parents may contain changes from multiple commits.
         revision_arg: &RevisionArg,
     ) -> Result<Commit, CommandError> {
         let expression = self.parse_revset(ui, revision_arg)?;
-        let should_hint_about_all_prefix = false;
-        revset_util::evaluate_revset_to_single_commit(
-            revision_arg.as_ref(),
-            &expression,
-            || self.commit_summary_template(),
-            should_hint_about_all_prefix,
-        )
+        revset_util::evaluate_revset_to_single_commit(revision_arg.as_ref(), &expression, || {
+            self.commit_summary_template()
+        })
     }
 
     /// Evaluates revset expressions to non-empty set of commit IDs. The
@@ -1591,12 +1587,10 @@ to the current parents may contain changes from multiple commits.
                     all_commits.insert(commit_id?);
                 }
             } else {
-                let should_hint_about_all_prefix = true;
                 let commit = revset_util::evaluate_revset_to_single_commit(
                     revision_arg.as_ref(),
                     &expression,
                     || self.commit_summary_template(),
-                    should_hint_about_all_prefix,
                 )?;
                 if !all_commits.insert(commit.id().clone()) {
                     let commit_hash = short_commit_hash(commit.id());
--- a/cli/src/revset_util.rs
+++ b/cli/src/revset_util.rs
@@ -262,7 +262,6 @@ pub(super) fn evaluate_revset_to_single_commit<'a>(
     revision_str: &str,
     expression: &RevsetExpressionEvaluator<'_>,
     commit_summary_template: impl FnOnce() -> TemplateRenderer<'a, Commit>,
-    should_hint_about_all_prefix: bool,
 ) -> Result<Commit, CommandError> {
     let mut iter = expression.evaluate_to_commits()?.fuse();
     match (iter.next(), iter.next()) {
@@ -279,7 +278,6 @@ pub(super) fn evaluate_revset_to_single_commit<'a>(
                 &commits,
                 elided,
                 &commit_summary_template(),
-                should_hint_about_all_prefix,
             ))
         }
     }
@@ -290,7 +288,6 @@ fn format_multiple_revisions_error(
     commits: &[Commit],
     elided: bool,
     template: &TemplateRenderer<'_, Commit>,
-    should_hint_about_all_prefix: bool,
 ) -> CommandError {
     assert!(commits.len() >= 2);
     let mut cmd_err = user_error(format!(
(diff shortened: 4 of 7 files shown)


martinvonz:
@ilyagr, FYI, since you've show interest in this before, though it sounded like you're okay with the direction last I heard.

ilyagr:
My personal instinct is that I'd prefer to change things a bit slower, e.g. by just making `allow_large_revisions` true by default but keeping the old behavior when it's `false` for now (without numerous warnings), for a release at least.

Would that block any work Yuya or somebody else is planning?

I think this together with #6912 and https://github.com/jj-vcs/jj/pull/6814 is a big change to our UI that I'd like people (including myself) to try out and see what kind of confusing situations might come up.

OTOH, I don't have any specific concerns at the moment, I think this is a good direction to go as long as we can address any potential confusion it causes in specific cases, and it seems to me that many people find the kind of confusing situations that I'm thinking of (see #6931 for an example) less problematic than I do. From my limited experiments so far, it all seem to work well apart from #6931, which we are in the process of fixing. So, if people feel confident enough in this approach to commit to it now, I won't stand in the way.

----------

I'll add a little manifesto to remind myself and others why I find the UI for divergent revisions and revsets unexpectedly expanding to multiple revisions important.

The reason is that the times when this UI is important heavily overlaps with the times when a user is confused about what `jj` is doing, and is worried about making a mistake. So, while these moments hopefully aren't very frequent, I think that these are the moments that are likely to shape a user's trust in the tool, and the feeling of safety or unsafety while using it.

In other words, if the UI is confusing in situations where user might end up in accidentally and without a good understanding of what's happening, I get worried, even if those situations are not very frequent in absolute terms.

So, my first instinct is to give ourselves some time to hunt for such situations and making sure we can address each one, as long as we don't feel that this slows down things too much.

yuja:
> My personal instinct is that I'd prefer to change things a bit slower, e.g. by just making `allow_large_revisions` true by default but keeping the old behavior when it's `false` for now (without numerous warnings), for a release at least.

This PR implements that. You're right that a new warning is added, but it's displayed only when the `all:` prefix is used, which wouldn't be overly verbose, I think.

cli/tests/test_rebase_command.rs
```diff
@@ -774,22 +763,42 @@ fn test_rebase_multiple_destinations() {
...
+    ]);
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
+    Warning: In revset expression
+     --> 1:1
+      |
+    1 | all:b|c
+      | ^-^
+      |
+      = Modifier `all:` is deprecated; multiple revisions are allowed by default
```
ilyagr: Could we not warn when `always-allow-large-revsets=false`?

We could later start warning that `always-allow-large-revsets=false` is deprecated. All else being equal, I'd prefer not to do that now, this is one part of what my other comment is about.

cli/tests/test_rebase_command.rs
```diff
@@ -774,22 +763,42 @@ fn test_rebase_multiple_destinations() {
...
+        "-d",
+        "b|c",
+    ]);
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
     Error: Revset `b|c` resolved to more than one revision
     Hint: The revset `b|c` resolved to these revisions:
       royxmykx c12952d9 c | c
       zsuskuln d18ca3e8 b | b
-    Hint: Prefix the expression with `all:` to allow any number of revisions (i.e. `all:b|c`).
```
ilyagr, line 784: Would it be a problem to keep hinting when `ui.always-allow-large-revsets=false`?

Here, it might make sense to also mention that the default is likely to switch to `ui.always-allow-large-revsets=true`

cli/tests/test_rebase_command.rs
martinvonz: I think you're suggesting that we take the first commit but wait with the second one? I'm fine with that. (I'm also fine with taking both commits.)

ilyagr:
> This PR implements that. You're right that a new warning is added, but it's displayed only when the `all:` prefix is used, which wouldn't be overly verbose, I think.

I pulled out a couple of places where we could do a softer deprecation. The other thing would be to change the tone of the docs changes from "(DEPRECATED)" to something like "(planned to be phased out in jj 0.33+)".

As I said in the previous comment, we could decide to deprecate things more decisively now. I'm leaning against that, but we can decide either way. The main purpose of these latest comments is to illustrate what I'm leaning towards as opposed to insisting on it more strongly.

cli/tests/test_rebase_command.rs
ilyagr: I think so, in terms of functionality. If we agree on this, we might also want to rewrite the docs a little more and change some warnings (instead of removing them). That could be a separate PR though.

yuja: I don't want to make the warning conditional because the user should be prepared for the future removal of `always-allow-large-revsets`.

If we aren't sure whether we'll deprecate the config, I would just remove the second patch.

ilyagr: It sounds like the three of us are all happy about the first commit, and we might all be happy about waiting with the second commit?

I'm not sure whether we're all on the same page about what warnings should exist while the first commit is merged and the second one isn't, but that seems less important.

yuja: My feeling is that we're likely to deprecate `always-allow-large-revsets` in jj 0.33 or so. If that's the plan, we should enable warnings early so the user can complain about that if they need `always-allow-large-revsets` and `all:`.

ilyagr: I agree that having warnings to prepare for likely deprecation is probably a good idea. I think we might disagree about exactly what warnings are best, but I'm getting tired, so both your position and mine are becoming less clear in my mind at the moment. But I don't feel that they are urgent to have.

ilyagr: I might be repeating myself, but one point became clearer for me. In a world where the first commit is in but we want users who have `always-allow-large-revsets=false` to have a good experience, I imagine that people with `always-allow-large-revsets=true` could get a nudge when they use `all:`.

If the warning was a person, it would go: "Hey! I see you're using `all:`. You don't have to do that anymore🎉! But if you would miss this safety net, set `always-allow-large-revsets=false` and *please* let us know your usecase, since we're thinking of deprecating it".

martinvonz: > In a world where the first commit is in but we want users who have `always-allow-large-revsets=false` to have a good experience

Since we're planning to deprecate `all:`, I don't think we care much about this world.

> I imagine that people with `always-allow-large-revsets=true` could get a nudge when they use `all:`.

Yes, I suppose we could get a modified version of the warning in that says that `all:` is redundant without saying that it's deprecated.

A warning if `always-allow-large-revsets=false` is set could also be reasonable.

Are you all okay with this plan? I think we could still take the first commit from this PR unchanged then, right? The other two changes, if you both are fine with then, can come a bit later (whenever someone feels like implementing them).

ilyagr: > The other two changes, if you both are fine with then, can come a bit later (whenever someone feels like implementing them).

My sense is that the amount of difference my suggestions would make in the best-case scenario might not be worth the length of the discussion we've been having. Yuya's plan is OK. Having the warning setup he already implemented might be better than having no warning at all. I can't promise to work on my version of the warnings; I certainly don't feel like doing it knowing that this goes directly against Yuya's wishes (which is the sense that I'm getting).

Is it OK if I chicken out, and let you finish reviewing the PR?

> > In a world where the first commit is in but we want users who have always-allow-large-revsets=false to have a good experience
>
> Since we're planning to deprecate all:, I don't think we care much about this world.

Just for the sake of being on the same page (sorry if I'm repeating myself and you already considered this), my concern was that there might be workflows we haven't thought about that aren't well-served by the new UI, in which case some people would need `always-allow-large-revsets=false`. E.g., we could discover more issues like https://github.com/jj-vcs/jj/pull/6931. In my eyes, at least, that might be a reason to delay full deprecation of `all:` or even backtrack (I tried to explain this in the little manifesto in https://github.com/jj-vcs/jj/pull/6937#issuecomment-3050456345).

It's not a certainty and, hopefully, something like this would only be a temporary annoyance even if it did come to pass.

yuja: I've slightly adjusted the warning message. Does it look good?

martinvonz: > Is it OK if I chicken out, and let you finish reviewing the PR?

Sure, no problem.

> Just for the sake of being on the same page (sorry if I'm repeating myself and you already considered this), my concern was that there might be workflows we haven't thought about that aren't well-served by the new UI, in which case some people would need `always-allow-large-revsets=false`.

Yeah, I understand. I was thinking that if we hear about such cases, then we simply need to change our plans, either to just not deprecate the config, or maybe even to roll it back to on-by-default.

> I've slightly adjusted the warning message. Does it look good?

Yes, looks good to me, thanks.

martinvonz (approved):

PR #6938, opened 2025-07-08
index: hide CompositeIndex, IndexEntry, and IndexPosition types

yuja:
# Checklist

If applicable:


yuja:
FWIW, changed-files index will have to be stored in separate files because indexing all commits upfront would be too slow. So we'll need a lot more refactoring.

martinvonz:
> indexing all commits upfront would be too slow

This seems possibly related to what we decide to do about indexing copy information. One option is to do it in the background after cloning. I don't know what the best solution to either indexing problem is.

yuja:
> One option is to do it in the background after cloning.

Yeah. I'll leave it for later. The initial version will add debug command to index N commits from the head of the commits index.

PR #6943, opened 2025-07-09
complete: don't complete local + remove bookmark when on same commit

jakobhellermann:
When you `jj new my-bran<TAB>`, you have tracked `my-branch` and both `my-branch` and `my-branch@origin` point to the same commit, the completion for the `@origin` branch is mostly noise.

Especially if you want to type more arguments like `jj new branchA<TAB> branchB<TAB>` the duplicate branches only get in the way.

# Checklist

If applicable:

  - Do minor features like this go in there?
- [x] I have updated the documentation (`README.md`, `docs/`, `demos/`) n/a
- [x] I have updated the config schema (`cli/src/config-schema.json`) n/a
  - is there a good place to test this? the `jj bookmark` tests don't exercise completion, and the `complete` tests don't set up the JJ environment

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -20,6 +20,7 @@ use clap::FromArgMatches as _;
 use clap::builder::StyledStr;
 use clap_complete::CompletionCandidate;
 use indoc::indoc;
+use indexmap::IndexMap;
 use itertools::Itertools as _;
 use jj_lib::config::ConfigNamePathBuf;
 use jj_lib::file_util::normalize_path;
@@ -300,8 +301,6 @@ fn revisions(match_prefix: &str, revset_filter: Option<&str>) -> Vec<CompletionC
         const REMOTE_BOOKMARK: usize = 3;
         const REVSET_ALIAS: usize = 4;
 
-        let mut candidates = Vec::new();
-
         // bookmarks
 
         let mut cmd = jj.build();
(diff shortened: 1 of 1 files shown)


ilyagr:
There's a typo in the commit description and the PR name (remove -> remote).

ilyagr:
Also, this needs a test. You should probably test the case when `branch1` and `branch2@remote` are at the same commit. IMO, both should show up in the completion.

Another thing I wonder about is this: let's say `branch` and `branch@remote` are at the same commit. It makes sense to me that `br<TAB>` doesn't need to show `branch@remote`. However, I think `branch@<TAB>` should.

I haven't followed the code carefully to check what it does in these cases, so these are just first thoughts and test case suggestions.

ilyagr:
> is there a good place to test this?

I *think* there should be similar tests in `cli/tests/test_completion.rs`, or did your comment about the JJ enviornment apply to them too? I think there are bookmark tests.

https://github.com/jj-vcs/jj/blob/362ab91972eb3ebeba869399b18f9cb9f63122fb/cli/tests/test_completion.rs#L134

(Sorry, I didn't see this part of your message immediately, should've replied slower)

cli/src/complete.rs
martinvonz, line 1: From commit description?

> complete: don't complete local + remove bookmark when on same commit

"remove" -> "remote"

PR #6948, opened 2025-07-09
revset docs: add an example for `commit_id`, change plural to singular

ilyagr:
I think an example illustrates the purpose of `commit_id` nicely.

Thanks to Yuya for explaining it to me.

I also don't think it can ever return more than one commit, since
non-unique prefixes are an error.

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -216,8 +216,10 @@ revsets (expressions) as arguments.
   change is divergent, this resolves to multiple commits. It is an error to use a
   non-unique prefix. Unmatched prefix isn't an error.
 
-* `commit_id(prefix)`: Commits with the given commit ID prefix. It is an error
-  to use a non-unique prefix. Unmatched prefix isn't an error.
+* `commit_id(prefix)`: Commit with the given commit ID prefix. It is an error
+  to use a non-unique prefix. Unmatched prefix isn't an error. For example,
+  `commit_id(abc)` will refer to a commit with prefix `abc` even if there also
+  exists a bookmark named `abc`.
 
 * `bookmarks([pattern])`: All local bookmark targets. If `pattern` is specified,
   this selects the bookmarks whose name match the given [string


docs/revsets.md
```diff
@@ -216,8 +216,10 @@ revsets (expressions) as arguments.
   change is divergent, this resolves to multiple commits. It is an error to use a
   non-unique prefix. Unmatched prefix isn't an error.
 
-* `commit_id(prefix)`: Commits with the given commit ID prefix. It is an error
-  to use a non-unique prefix. Unmatched prefix isn't an error.
+* `commit_id(prefix)`: Commit with the given commit ID prefix. It is an error
```
PhilipMetzger, line 219: nit: I don't think changing the Commits to Commit is correct here, since there may be multiple with the same prefix. That's taken from https://github.com/jj-vcs/jj/pull/6912#issuecomment-3050911051

PhilipMetzger (approved):
SGTM, you should hash this out with Yuya though

docs/revsets.md
ilyagr: As it says right below, "It is an error to use a non-unique prefix". So, I don't think this function can ever return more than one commit.

**Update:** I added this point to the commit description.

Separately, I think we potentially might also want something like `non_unique_commit_id` that doesn't have that restriction (with a better name), but it'd be used much less frequently. In my mind, it'd be used mainly to resolve a commit id among a custom set of commits, e.g. (using imaginary functions) `non_unique_commit_id(abc) & today()`. If so, `non_unique_commit_id` could of course return multiple commits.

docs/revsets.md
```diff
@@ -216,8 +216,10 @@ revsets (expressions) as arguments.
...
+  to use a non-unique prefix. Unmatched prefix isn't an error. For example,
+  `commit_id(abc)` will refer to a commit with prefix `abc` even if there also
+  exists a bookmark named `abc`.
```
yuja, line 222: This also applies to `change_id(xyz)`, `bookmarks(name)`, etc. My feeling is that it would be too verbose to add this note to `<namespace>(symbol)` functions. Instead, maybe we can add a note to the symbol priority section?

ilyagr: My intention was to clarify the purpose of `commit_id`, which I find uniquely confusing. Except for the issue with a bookmark of the same name, and the behavior if there are no commits with the prefix, using `commit_id(abc)` seems identical in every way to just using `abc` on its own. So, IMO, people will wonder why it exists.

Adding a note to the symbol priority section might also be useful, but it wouldn't address this intention of mine.

yuja: IMHO, `commit_id()` exists primarily for completeness. It's odd if there were no "search commit id" function despite there are "search change id"/"tag"/"bookmark" functions. In practice, `commit_id()` wouldn't be that useful.

ilyagr: You yourself cited the bookmark issue in https://github.com/jj-vcs/jj/pull/6912#issuecomment-3050787524, in response to me doubting whether `commit_id` should exist. (**Update:** And, in case there was any doubt, I appreciate that response very much! The whole idea of this PR was to share this nugget of wisdom with others)

In my mind, at least, completeness alone would not be a sufficient reason for `commit_id` to exist. You might disagree, and I might even be wrong about whether that's a sufficient reason, but the point is that the bookmark issue resolves the matter for me and I think would resolve the matter for a reader thinking along these lines.

yuja: Yes, it's a potential use case of `commit_id()`. In practice, I would use longer commit id prefix, though.

Real users of `commit_id()`, `change_id()`, etc. might be scripting/GUI. Since GUI frontends have to support arbitrary repositories, it wouldn't be good enough to use `full_commit_id_hex` to resolve the symbol as a commit id. I think this use case can be better explained in the symbol priority section.

> In my mind, at least, completeness alone would not be a sufficient reason for commit_id to exist. You might disagree, 

Yeah, I disagree. Completeness and consistency are important for me.

martinvonz: I agree with Ilya that it's not obvious what `commit_id()` is useful for so it's helpful if we explain that. I think the disambiguation use case is important, especially for script authors, so I think it's good to mention it. Mentioning that in the "Priority" section makes sense to me.

I think `bookmark()` is a bit different because it provides more functionality with its pattern-matching support.

As I said earlier, I also find it surprising that `commit_id()` doesn't fail for non-matching ids. But as long as it does, I suppose the function can be used/abused as an alternative way of writing `present(abc)`. Since I'm not sure we should keep the current behavior, I don't really think we should document that use case, however.

martinvonz: I was reminded of the `change_id()` function when reviewing some Google-internal docs. I sent #7000 to clarify how it's useful for overriding the default resolution priority. I hope you don't mind, Ilya.

ilyagr: Yes, #7000 helps a lot. I'll close this. Ideally, my mild preference would be to do both, but I'm much more comfortable not merging something like this PR if we do something like #7000.

Aside: I don't think I understood this idea very well when Yuya first suggested it in https://github.com/jj-vcs/jj/pull/6948#discussion_r2196250997.

> > In my mind, at least, completeness alone would not be a sufficient reason for commit_id to exist. You might disagree,
> 
> Yeah, I disagree. Completeness and consistency are important for me.

The implication (intended or not) that I know nothing of completeness or consistency is frustrating to me. I was wondering whether to reply, and at what length. I decided to state one thesis in two different ways:

- Actionably: it can be extremely helpful to point out that an idea A is consistent with the idea X to people who might not realize it. Once people understand the consistency, this stops being helpful; if they still disagree, that likely means they feel that something else might be more important for A to be consistent with.

- Philosophically: all minds love consistency and completeness. Consistent patterns are the only reason any mind has any chance to make sense of the immensely complicated world it experiences; we love finding patterns and creating them. In a world where perfect consistency or completeness are impossible, the question is not who loves consistency more, but how we choose whether to make A consistent with X or Y when we can't do both.

yuja: Sorry, I just meant I tend to weigh more on consistent UI than usefulness. In this context, the existence of `change_id()` is enough reason for me to add `commit_id()` to provide consistent UI, even if they weren't so useful.

PR #6950, opened 2025-07-10
commit_builder: allow recreating commit, don't create cycle in parent…

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/commit_builder.rs
+++ b/lib/src/commit_builder.rs
@@ -19,7 +19,6 @@ use std::sync::Arc;
 use pollster::FutureExt as _;
 
 use crate::backend;
-use crate::backend::BackendError;
 use crate::backend::BackendResult;
 use crate::backend::ChangeId;
 use crate::backend::CommitId;
@@ -352,19 +351,14 @@ impl DetachedCommitBuilder {
     pub fn write(self, mut_repo: &mut MutableRepo) -> BackendResult<Commit> {
         let predecessors = self.commit.predecessors.clone();
         let commit = write_to_store(&self.store, self.commit, &self.sign_settings)?;
-        // FIXME: Google's index.has_id() always returns true.
-        if mut_repo.is_backed_by_default_index() && mut_repo.index().has_id(commit.id()) {
-            // Recording existing commit as new would create cycle in
-            // predecessors/parent mappings within the current transaction, and
-            // in predecessors graph globally.
-            return Err(BackendError::Other(
-                format!("Newly-created commit {id} already exists", id = commit.id()).into(),
-            ));
-        }
         mut_repo.add_head(&commit)?;
-        mut_repo.set_predecessors(commit.id().clone(), predecessors);
+        if !predecessors.contains(commit.id()) {
+            mut_repo.set_predecessors(commit.id().clone(), predecessors);
+        }
         if let Some(rewrite_source) = self.rewrite_source {
-            if rewrite_source.change_id() == commit.change_id() {
+            if commit.change_id() == rewrite_source.change_id()
+                && commit.id() != rewrite_source.id()
+            {
                 mut_repo.set_rewritten_commit(rewrite_source.id().clone(), commit.id().clone());
             }
         }
--- a/lib/src/repo.rs
+++ b/lib/src/repo.rs
@@ -951,6 +951,7 @@ impl MutableRepo {
     /// docstring for `record_rewritten_commit` for details.
     pub fn set_rewritten_commit(&mut self, old_id: CommitId, new_id: CommitId) {
         assert_ne!(old_id, *self.store().root_commit_id());
+        assert_ne!(new_id, old_id);
         self.parent_mapping
             .insert(old_id, Rewrite::Rewritten(new_id));
     }
@@ -968,10 +969,10 @@ impl MutableRepo {
         new_ids: impl IntoIterator<Item = CommitId>,
     ) {
         assert_ne!(old_id, *self.store().root_commit_id());
-        self.parent_mapping.insert(
-            old_id.clone(),
-            Rewrite::Divergent(new_ids.into_iter().collect()),
-        );
+        let new_ids = new_ids.into_iter().collect_vec();
+        assert!(!new_ids.contains(&old_id));
+        self.parent_mapping
+            .insert(old_id.clone(), Rewrite::Divergent(new_ids));
     }
 
     /// Record a commit as having been abandoned in this transaction.
--- a/lib/tests/test_commit_builder.rs
+++ b/lib/tests/test_commit_builder.rs
@@ -16,7 +16,6 @@ use assert_matches::assert_matches;
 use futures::StreamExt as _;
 use indoc::indoc;
 use itertools::Itertools as _;
-use jj_lib::backend::BackendError;
 use jj_lib::backend::ChangeId;
 use jj_lib::backend::MillisSinceEpoch;
 use jj_lib::backend::Signature;
@@ -386,6 +385,11 @@ fn test_rewrite_to_identical_commit(backend: TestRepoBackend) {
         )
         .write()
         .unwrap();
+    let _child_commit = tx
+        .repo_mut()
+        .new_commit(vec![commit1.id().clone()], store.empty_merged_tree_id())
+        .write()
+        .unwrap();
     let repo = tx.commit("test").unwrap();
 
     // Create commit identical to the original
@@ -394,12 +398,13 @@ fn test_rewrite_to_identical_commit(backend: TestRepoBackend) {
     builder.set_predecessors(vec![]);
     // Writing to the store should work
     let commit2 = builder.write_hidden().unwrap();
-    assert_eq!(commit1, commit2);
-    // Writing to the repo shouldn't work, which would create cycle in
-    // predecessors/parent mappings
-    let result = builder.write(tx.repo_mut());
-    assert_matches!(result, Err(BackendError::Other(_)));
-    tx.repo_mut().rebase_descendants().unwrap();
+    assert_eq!(commit2, commit1);
+    // Writing to the repo should be allowed but should not result in any changes
+    // (to parent_mapping)
+    builder.write(tx.repo_mut()).unwrap();
+    assert!(!tx.repo().has_rewrites());
+    // _child_commit should, consequently, not be rebased
+    assert_eq!(tx.repo_mut().rebase_descendants().unwrap(), 0);
     tx.commit("test").unwrap();
 
     // Create two rewritten commits of the same content and metadata
@@ -409,12 +414,11 @@ fn test_rewrite_to_identical_commit(backend: TestRepoBackend) {
         .set_description("rewritten")
         .write()
         .unwrap();
-    let result = tx
-        .repo_mut()
+    tx.repo_mut()
         .rewrite_commit(&commit1)
         .set_description("rewritten")
-        .write();
-    assert_matches!(result, Err(BackendError::Other(_)));
+        .write()
+        .unwrap();
     tx.repo_mut().rebase_descendants().unwrap();
     tx.commit("test").unwrap();
 }


lib/src/commit_builder.rs
```diff
@@ -352,19 +351,14 @@ impl DetachedCommitBuilder {
...
         let commit = write_to_store(&self.store, self.commit, &self.sign_settings)?;
-        // FIXME: Google's index.has_id() always returns true.
-        if mut_repo.is_backed_by_default_index() && mut_repo.index().has_id(commit.id()) {
-            // Recording existing commit as new would create cycle in
-            // predecessors/parent mappings within the current transaction, and
-            // in predecessors graph globally.
-            return Err(BackendError::Other(
-                format!("Newly-created commit {id} already exists", id = commit.id()).into(),
-            ));
-        }
```
yuja, line 363: Can you leave this error and add special case for `commit.id() == rewrite_source.id()`?

We don't handle indirect cycles within/across operations. `evolog` won't detect cycles across operations, but `op diff` can notice cycles in accumulated predecessors graph.

lib/src/commit_builder.rs
```diff
@@ -352,19 +351,14 @@ impl DetachedCommitBuilder {
...
         mut_repo.add_head(&commit)?;
-        mut_repo.set_predecessors(commit.id().clone(), predecessors);
+        if !predecessors.contains(commit.id()) {
```
yuja, line 355: nit: I would use `commit.id() != rewrite_source.id()` for consistency. If the caller set weird predecessors, I think panicking is good.

lib/src/commit_builder.rs
martinvonz: So instead of filtering out the call to `mut_repo.set_rewritten_commit()` we make it an error here?

The case we ran into at Google was code like this:

```rust
tx.repo_mut().rewrite_commit(&commit).set_description(description).write()?;
tx.commit("set description")?;
```

If the description is unchanged and this code is run within the same second as the commit was created, then we end up with the new commit matching the old commit here. I kind of feel like it should be allowed to write such code without checking for the previous description, but I also see value in erroring out so we learn about these cases (I've since updated the code at Google to avoid creating any commits or operations). So I think I'm saying that I think I'm okay with the error for now as I think you're suggesting, but I also wonder if we should allow such cases in the long term.

yuja: IMHO, creating an existing commit should be an error in general, because we don't know how such commit should be recorded as rewritten. I also think it's okay to accept no-op rewrite (as a special case) because that's common and relatively easy to work around.

martinvonz: I disagree. I think it should be allowed to create an existing commit because it's a strange requirement to put on callers to first check if the commit already exists, and it's going to be a racy check too. So I think we will need to allow callers to create an existing commit. I think even requiring that it's not a visible commit is an unnecessary restriction. What is the caller supposed to do? As I said, I think we should even allow code like the snippet above (i.e. predecessor cycles), but I don't mind if we're more restrictive to start with. In my mind, checking that the description is different in that code is mostly an optimization; the caller should be allowed to have the naive code if they prefer.

I'm okay with either allowing predecessor cycles or rejecting them for now. I don't feel strongly either way, but I do feel strongly that we should allow creating existing commits. Does that make sense?

yuja: I don't think the caller needs to check the precondition. In practice, `commit.id() == rewrite_source.id()` would be the only situation where the existing commit can reappear. The user could set fake timestamp to trigger the error, but I think it's okay to just return an error in that case.

> I'm okay with either allowing predecessor cycles or rejecting them for now. I don't feel strongly either way, but I do feel strongly that we should allow creating existing commits. Does that make sense?

Since we don't have a good way to handle indirect predecessors/rewrite cycles, I would just unblock the common case `commit.id() == rewrite_source.id()`.

martinvonz: > The user could set fake timestamp to trigger the error, but I think it's okay to just return an error in that case.

Given Git's second-level granularity, I can see it quite easily happening in other cases too once we stop writing predecessors. We saw it happening on a server at Google where we use the Git hashing scheme. I don't know the details in that scenario but somehow the server received a request to update the description to what it already was. That request arrived within a second of whatever previous request had created the commit.

> Since we don't have a good way to handle indirect predecessors/rewrite cycles, I would just unblock the common case `commit.id() == rewrite_source.id()`.

I.e. what I already have in this PR? Or do you mean that you prefer it I make it an error? I'm fine with either.

yuja: > I.e. what I already have in this PR? Or do you mean that you prefer it I make it an error? I'm fine with either.

I meant something like (because we have no idea how to resolve cycles):
```
if commit.id() != rewrite_source.id() && mut_repo.is_backed_by_default_index() && mut_repo.index().has_id(commit.id()) {
    return Err(...);
```

> Given Git's second-level granularity, I can see it quite easily happening in other cases too once we stop writing predecessors.

You mean, in Google-scale? That's possible.

martinvonz: > You mean, in Google-scale? That's possible.

I meant in general. Something as simple as this currently fails:

```
$ JJ_TIMESTAMP=2025-01-01T00:00:00+00:00 jj desc -m foo
$ jj undo
$ JJ_TIMESTAMP=2025-01-01T00:00:00+00:00 jj desc -m foo
```

Setting JJ_TIMESTAMP e.g. when importing patches from somewhere else seems like a legitimate use case, so I think we should make that work. It also happen when running some script without setting JJ_TIMESTAMP.

ilyagr:
You likely already considered this, and I mainly looked at the commit description, but perhaps it's worth mentioning, and in any case I'm curious: does our current trick of subtracting a second from the committer time-stamp to de-duplicate commit ids not help here? 

One worry is that, in the worst case, you could keep subtracting for a while.


**Update** Ah, I guess the main problem is that you might need to write the commits concurrently on different machines, and there isn't a good way to reserve the commit id.

If this is the problem, mentioning concurrency in the commit description might help.

martinvonz:
> You likely already considered this, and I mainly looked at the commit description, but perhaps it's worth mentioning, and in any case I'm curious: does our current trick of subtracting a second from the committer time-stamp to de-duplicate commit ids not help here?

We (both Yuya and I, I think) hope to get rid of that hack once we no longer store any commit-linked metadata outside of the Git commit object.

@yuja : Perhaps what I'm missing is what the problem with cycles is. Is it the kind of cycle you would get if you rewrote commit A to commit B in operation X and then rewrote commit B to commit A in operation Y? If that's it, what problem does that kind of cycle cause?

ilyagr:
I see, the old problem of duplicating a commit and then updating both copies in the same second (to make their committer dates the same) will no longer result in commits with the same hash if the change id is stored inside the commit.

Though, if somebody (say on a remote) rebased both commits with some tool that strips the change id header, they might have some trouble.

(This is of course out of scope for this pr)

lib/src/commit_builder.rs
yuja: I thought it would be okay to error out if the timestamp was faked, but yeah, importing patches might be a valid use case. I don't know if `git` preserves the committer timestamp in that case, though.

FWIW, for non-patch imports (akin to `hg unbundle`), I assume we'll use `backend.write_commit()` directly.

yuja:
> We (both Yuya and I, I think) hope to get rid of that hack once we no longer store any commit-linked metadata outside of the Git commit object.

+1

> Perhaps what I'm missing is what the problem with cycles is. Is it the kind of cycle you would get if you rewrote commit A to commit B in operation X and then rewrote commit B to commit A in operation Y? If that's it, what problem does that kind of cycle cause?

An immediate problem is that `MutableRepo` would panic if cycle existed within the same operation. I'm not pretty sure if inter-operation cycles should be banned, but the existence of cycles will mean that we can't simply apply DAG algorithms. For example, I don't know if Sapling's graph renderer can deal with cyclic graph.

lib/src/commit_builder.rs
martinvonz: > I thought it would be okay to error out if the timestamp was faked

I was thinking about cases like where someone wants to import history from some other VCS.

> importing patches might be a valid use case. I don't know if `git` preserves the committer timestamp in that case, though.

They don't preserve it on `git format-patch`+`git am`. But *maybe* they will do that some day. I happened to [talk Junio about that](https://lore.kernel.org/git/xmqqecus6uno.fsf@gitster.g/) just a week ago :)

yuja: > I was thinking about cases like where someone wants to import history from some other VCS.

My assumption is that we won't use `CommitBuilder` when importing history losslessly. It wouldn't make much sense to generate predecessors/rewrite history for the import.

martinvonz: I meant if someone writes a script for importing history. 

I suspect we're going to have to remove the error at some point, but if you feel strongly about it, I can leave the error for now. What do you prefer?

yuja: > I suspect we're going to have to remove the error at some point, but if you feel strongly about it, I can leave the error for now. What do you prefer?

I would leave the error because the current implementation can't handle cycles (and will either panic or error out.) If we find a better way to deal with that, I think it's good to remove the restriction.

joyously: > I meant if someone writes a script for importing history.

Is the script the thing that interfaces with [git-fast-import](https://git-scm.com/docs/git-fast-import) or the fast import backend itself, which has options for date formats?
Also, just a reminder that `git commit` has a flag to supply a date, which I expected `jj` to have as well. Breezy has one also.

martinvonz: I was just talking about a hypothetical script. I've written such scripts for importing from ClearCase to Git in the past, for example. Anyway, it was just an example. I think we need to handle recreating commits regardless of why someone happens to create them.

martinvonz:
> An immediate problem is that `MutableRepo` would panic if cycle existed within the same operation. I'm not pretty sure if inter-operation cycles should be banned, but the existence of cycles will mean that we can't simply apply DAG algorithms. For example, I don't know if Sapling's graph renderer can deal with cyclic graph.

That seems to behave reasonably at least in this simple case (created by a hacked jj binary):

Screenshot 2025-07-16 at 07 54 12

It looks funny with the two working-copy nodes but I think that's the right thing to do. Other than `evolog`, I don't think we have anything that walks the predecessors, right?

Regarding intra-operation cycles, I can add a stricter check in addition to filtering out length-1 cycles as this PR currently does. Would that make you comfortable with removing the error for recreating commits?

yuja:
> > An immediate problem is that `MutableRepo` would panic if cycle existed within the same operation. I'm not pretty sure if inter-operation cycles should be banned, but the existence of cycles will mean that we can't simply apply DAG algorithms. For example, I don't know if Sapling's graph renderer can deal with cyclic graph.
> 
> That seems to behave reasonably at least in this simple case (created by a hacked jj binary):

Yeah, I guessed that might work so long as the graph renderer doesn't keep track of emitted nodes to detect data inconsistency.

> It looks funny with the two working-copy nodes but I think that's the right thing to do. Other than `evolog`, I don't think we have anything that walks the predecessors, right?

`jj op diff` will do, and transitive predecessors will be resolved there unlike `evolog`. Another potential problem is that we might add some revset-like filtering to `evolog`, and cycles could break topological sorting of the filtered graph.

> Regarding intra-operation cycles, I can add a stricter check in addition to filtering out length-1 cycles as this PR currently does. Would that make you comfortable with removing the error for recreating commits?

No. Just curious, why do you want to remove the error checking? I know you have a real problem caused by the length-1 cycles, but I don't see why we also need to unblock the other unusual cases without addressing the underlying problems.

martinvonz:
> Just curious, why do you want to remove the error checking?

Only because I think it's the right thing to do. Since we've fixed our server at Google to not create operations that do no-op updates of commit descriptions, it doesn't matter to us anymore AFAIK.

>I know you have a real problem caused by the length-1 cycles, but I don't see why we also need to unblock the other unusual cases without addressing the underlying problems.

I do want to address the underlying problems. Preventing intra-operation cycles is just the only problem I'm aware of, so I thought that if we prevent that, then we can allow recreating existing commits.

What would make you comfortable with removing the error? To be clear, there's no rush from my side.

yuja:
> I do want to address the underlying problems. Preventing intra-operation cycles is just the only problem I'm aware of, so I thought that if we prevent that, then we can allow recreating existing commits.
> 
> What would make you comfortable with removing the error? To be clear, there's no rush from my side.

I see. As I said, inter-operation cycles can also cause problems atm, so I don't want to remove the error. That's the point.

I don't know if it's practical to accept cycles in the predecessors graph. It might be just simpler to store predecessors (or random bytes, incremented timestamp, etc.) in the commit object to prevent this kind of problems.

martinvonz:
> I see. As I said, inter-operation cycles can also cause problems atm, so I don't want to remove the error. That's the point.

Ah, you're talking about the cycle-detection in `jj op diff` you mentioned earlier, right? Sorry, I missed that until now. I think we'll need to try to fix that somehow. I agree that this PR should wait until we've fixed that. Will convert this to a draft until then.

yuja:
> Ah, you're talking about the cycle-detection in `jj op diff` you mentioned earlier, right?

Yes, and some other potential problems on graph filtering/grouping. iirc, `TopoGroupedGraphIterator` doesn't work if multiple items of the same identity were queued.

PR #6956, opened 2025-07-10
cli: add --{no-,}tags option to `jj git fetch`

### LINKED ISSUE #6954: Duplicate change-id from CI tagging on push
As reported [on discord](https://discord.com/channels/968932220549103686/1392799949753946182), by @avsaase.

Background: their CI appears to create a tag on every push to github, resulting in tags `branch.0`, `branch.1`, etc. Presumably this is mean to be used as a pseudo-evolog / aid for interdiff / range-diff.

This causes problems when using jj. Example workflow:

```bash
jj git push
# amend, rebase, etc
jj git fetch
# generated tag is fetched, making old commits visible, causing duplicated change ids
```

It's possible to work around this by abandoning the commits that are visible due to the generated tags, with something like `jj abandon --ignore-immutable '@..(visible_heads() & tags() ~ remote_bookmarks())'` (revset might need tweaking for exact situation, unsure). However, this is annoying to do on _every_ fetch, and possibly dangerous depending on the exact `abandon` invocation.

(Original suggestion on discord was to the effect of `git tag | xargs git tag -d`. I think hiding the commits by abandoning them is the nicer solution than deleting all tags.)

Request: a way to _not_ fetch tags when calling `jj git fetch`. Possibly `jj git fetch --no-tags`, or a config option `git.fetch-tags = false`.

bryceberger:
closes #6954 

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * Added `templates.workspace_list` template to customize the output of `jj workspace list`.
 
+* `jj git fetch` now accepts `--tags` and `--no-tags` options, to enable or
+  disable fetching of tags. The default is to fetch tags, controllable with the
+  `git.fetch-tags` config option.
+
 ### Fixed bugs
 
 ### Packaging changes
--- a/cli/src/commands/git/clone.rs
+++ b/cli/src/commands/git/clone.rs
@@ -220,7 +220,7 @@ fn fetch_new_remote(
     let mut tx = workspace_command.start_transaction();
     let mut git_fetch = GitFetch::new(tx.repo_mut(), &git_settings)?;
     with_remote_git_callbacks(ui, |cb| {
-        git_fetch.fetch(remote_name, &[StringPattern::everything()], cb, depth)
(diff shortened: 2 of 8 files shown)


bryceberger:
(need to add tests, that might help my following confusion)

It's unclear, from `git fetch --help`, whether this changes behavior. Specifically:

> -n, --no-tags
>     By default, tags that point at objects that are downloaded from the remote
>     repository are fetched and stored locally. This option disables this automatic tag
>     following.

Does this mean that, by default, tags that are inside of `old_remote_bookmarks()..new_remote_bookmarks()` are not fetched? If that's the case, this changes jj's behavior in that case.

yuja:
iirc, `git` by default fetches tags pointing to the commits to be fetched. `--no-tags` will turn it off, and `--tags` will fetch other tags (i.e. fetch unrelated commits.)

To address the problem described in https://github.com/jj-vcs/jj/issues/6954, I think it's better to add a setting to filter out uninteresting tags. I have no idea if we can translate such setting to refspec/option which `git fetch` can understand, though.

ipetkov:
Sorry for noticing that this PR existed before I opened https://github.com/jj-vcs/jj/pull/7175. Just wanted to mention we now have `jj git fetch --fetch-tags ...` which can be used to explicitly fetch tags

PR #6966, opened 2025-07-11
governance: specify that >1/3 maintainers from a single company is di…

martinvonz:
…sallowed

We've discussed this a few times among the maintainers. Let's write it down before it happens. The Mercurial steering committee has a similar rule, FWIW.

# Checklist

If applicable:

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -142,3 +142,14 @@ Maintainer in question). This can be due to lack of participation or conduct
 violations, among other things. Note that Maintainers are subject to a higher
 set of behavioral and communicative standards than average contributor or
 participant.
+
+### Single-Company Influence
+
+At most 1/3 of the maintainers may be paid for their contributions by a single
+company. This is to reduce the risk of a single company controlling the
+project's direction. If the 1/3 limit gets exceeded because an existing
+maintainer gets hired by the same company as some other existing maintainer(s),
+then the maintainers will have to decide how to resolve the situation. The
+maintainer in question gets to vote, as long as this doesn't mean the company
+in question has half the votes (usually meaning there are at least 5 maintainers
+total).


nasamuffin:
It might be nice to give a little bit more flexibility to the maintainer group than this PR describes. For example, if from the existing group one person gets hired by a company and will over-represent, a different remedy could be to add another maintainer to balance the group out. Or, let's say Maintainer 1 hires Maintainer 2 for their company, but because Maintainer 1 is now a hiring manager, Maintainer 1 would rather step down than Maintainer 2 - that might be more appropriate.

Could it be better to state that the maintainer group can try to decide what to do, but if no decision is reached, the fallback is to remove the person who recently joined the company that would now have undue influence?

martinvonz:
> Could it be better to state that the maintainer group can try to decide what to do, but if no decision is reached, the fallback is to remove the person who recently joined the company that would now have undue influence?

True, they can make a decision even if we include the maintainer in question in the vote. Will update.

GOVERNANCE.md
```diff
@@ -142,3 +142,13 @@ Maintainer in question). This can be due to lack of participation or conduct
...
 set of behavioral and communicative standards than average contributor or
 participant.
+
+### Single-Company Influence
+
+At most 1/3 of the maintainers may be paid for their contributions by a single
+company. This is to reduce the risk of a single company controlling the
+project's direction. If the 1/3 limit gets exceeded because an existing
+maintainer gets hired by the same company as some other existing maintainer(s),
+then the maintainers will have to decide how to resolve the situation. The
```
ilyagr, line 152: I think violations of the 1/3 rule should be temporary, but I wonder if we could have a defined process for votes while it's violated.

Perhaps we could say that each company gets at most 1/3 of the votes. So, if 3/7 maintainers are suddenly hired by Skynet, Inc, they'd all remain maintainers for commit review purposes, etc. However, if there's a vote, they'd get only 2 votes. They'd have to decide among themselves how to distribute them (I'm guessing they'd either vote among themselves, or decide which person doesn't vote while the 1/3 rule is violated).

If we didn't have this rule, something similar would happen: the 3 maintainers would probably have to decide among themselves which one of them steps down. I think that's less good.

In other words, when we the rule is violated, we could either:

- Have 3 maintainers from the same company go into a black box, and wait for 2 maintainers to come out.
- Every vote, have 3 maintainers from the same company go into a black box, and wait for them all to come out with 2 votes.

This does make ties in votes more likely, but this is the same problem that'd exist if one of the 7 maintainers stepped down.

ilyagr: (Hopefully entertaining aside; I jumped into a rabbit hole and didn't quite navigate it fully yet)

As another example, if there are 999 maintainers from Skynet[^cyberdyne], 9999 maintainers from Weyland-Yutani, and 1 unaffiliated maintainer, each group should have 1 vote (1/3 of the total number of votes, which is 3).

[^cyberdyne]: Aside/update: I remembered that the company was called Cyberdyne, but apparently there's now [a real company called Cyberdyne](https://en.wikipedia.org/wiki/Cyberdyne_Inc.). They make a product called HAL. AFAIK, Weyland-Yutani is not yet a real company.

I think you can always compute the correct number of votes per group by an algorithm:

  0. Start with each group getting as many votes as it has people.
  1. If each group has <= 1/3 votes, stop.
  2. If one group has the largest fraction of votes, remove a vote from it. This also decreases the total number of votes.
  3. If two groups have (the same) largest fraction of votes, remove a vote from both.
  4. GOTO 1

There's probably an easier way to express the answer.

emilazy: I feel it's probably a good thing to make these CoI situations enough of a problem that there's pressure to fix them quickly. Though I guess you can imagine situations where there are truly urgent decisions that need making and it's intractable for someone to just stand down to let that happen.

thoughtpolice:
FWIW, I do think this is a good rule and (barring Ilya's suggestions which I find interesting) a correct thing to do now that I'm working full time on jj-related stuff. So, if nothing else I think we can go with this. I'll leave Waleed and Yuya to also leave comments if they have any first, though.

(Note that this comment is 100% in my personal capacity as a member of the project, not as an employee of ERSC.)

PhilipMetzger:
This may be out of scope for this PR but I'd (out of pure interest to the community) really like to have some kind of communication from the maintainers how the conflict was solved and what remediation was chosen.

martinvonz:
> This may be out of scope for this PR but I'd (out of pure interest to the community) really like to have some kind of communication from the maintainers how the conflict was solved and what remediation was chosen.

Yes, I think that's out of scope, but I agree that the maintainers need to tell the community what happened. I think this is more about transparency in general. Feel free to ask the maintainers how they made some decision in general. They're of course also free to not share that information (sometimes it's sensitive for some reason).

PhilipMetzger:
> Feel free to ask the maintainers how they made some decision in general. 

If you require the community to ask the questions by themselves then it's very possible that it won't happen since the people asking these questions may get overwhelmed by their own life which then could result in minimal transparency. So I'd prefer if the current maintainers could establish a transparency culture in the maintainer group as a baseline so new maintainers will have it ingrained in them from the start. 

> They're of course also free to not share that information (sometimes it's sensitive for some reason).

Of course, I think that was somewhat implied since the community cannot judge the scope which is required for that. 

Note: I only raised this concern since I may leave at some point.

martinvonz:
> If you require the community to ask the questions by themselves

That's not what I meant. Of course the maintainers should strive for transparency. I just meant that you should feel free to ask if there's a case where you feel like they/we are fail.

nasamuffin:
I think I agree with @PhilipMetzger's general point, though, that if there are expectations of how the maintainer group should behave (e.g. proactively sharing the outcome of disagreements or discussions), we should write it down. That makes sure that the entire community understands what standard we're holding our maintainers to.

(Yes, I'm coming to the thread very late, but - ) I also agree with @emilazy's point that the conflicting state should have friction to force it to be as short as possible. Maintainer influence extends beyond voting; maintainers also theoretically talk to each other and convince each other of positions during informal discussions, and perform review of topics which might not rise to needing a vote. If we have 15 Google maintainers and 3 hobbyist maintainers, yes, Google will lose the vote, but Google will win in the public tide of opinion among the maintainers. I think it's much simpler to just even out the maintainer group ASAP. (and again - that might mean "we should onboard another handful of hobbyist maintainers to counteract", if we really are at the scale where having 18+ maintainers is valuable for the project.)

martinvonz:
I'm happy to review a PR that adds a section somewhere about how maintainers should strive for transparency. I think most of you are better at articulating that, so I'll avoid sending the PR myself.

GOVERNANCE.md
```diff
@@ -142,3 +142,13 @@ Maintainer in question). This can be due to lack of participation or conduct
...
+maintainer in question gets to vote (the company will still not have half of
+the votes).
```
ilyagr: Strictly speaking, the parenthetical is only true if there are at least 5 maintainers (or did I miss something?)

Apologies for the pedantry, but it seems at least somewhat plausible this could matter one day.

ilyagr: I guess we could say:

>The maintainer in question gets to vote, as long as this doesn't mean the company in question has half the votes (usually meaning there are at least 5 maintainers total)

?

This feels mildly better to me than just removing the parenthetical.

We could also instead say "over half the votes" which would mean at least 4 maintainers total for the newly corporate person to vote.

The rule of <=1/3 seems to imply we can never have <3 maintainers :) (I'm not sure that this is bad, we could later change the rules to allow it if it becomes relevant)

GOVERNANCE.md
ilyagr: To be clear, I'm happy with adopting the policy without this amendment.

I personally still like it and haven't figured out a way it would go wrong that scares me, but clearly the feeling is not universal.

> If we have 15 Google maintainers and 3 hobbyist maintainers, yes, Google will lose the vote, but Google will win in the public tide of opinion among the maintainers.

@nasamuffin, I'll explain why I disagree, mainly for future reference. This should be tempered by the fact that I'm not familiar with as many open source communities as probably either of the Emilys who disliked the idea. 

If the issue is controversial enough, I think the 3 hobbyists could become quite vocal. Spamming public opinion with a large number of people could be done without them being maintainers. The important thing, in my mind, is that two independent maintainers would need to be swayed to make a decision; no one independent maintainer feels like they have the unilateral power to unblock something for the company (or the unilateral responsibility for deciding not to).

PhilipMetzger:
> I'm happy to review a PR that adds a section somewhere about how maintainers should strive for transparency. I think most of you are better at articulating that, so I'll avoid sending the PR myself.

Done in #7006.

GOVERNANCE.md
martinvonz: Makes sense. Done.

martinvonz:
Is this good enough now? I'll merge as soon as I have approval from anyone. Someone can always send another PR to tweak it later.

PR #6967, opened 2025-07-12
fileset, str_util: switch to globset

### LINKED ISSUE #6919: Switch to globset
From:

https://github.com/jj-vcs/jj/pull/6899#issuecomment-3041182119

> > The issue I have is that it seems nontrivial to make a to_regex translator for globs,
>
> I think we'll need to switch to the globset crate at some point. The current StringPattern API doesn't have a separate .to_matcher() step, so adapting to GlobSet might not be straightforward. Still I think we can leverage globset to implement .to_regex().
> 
> https://docs.rs/globset/0.4.16/globset/struct.Glob.html

This is worth doing because it allows us to uniformly translate all StringTemplates to regexes which then allows us to expose a `.match` instead of just a `.contains`.

Required by: https://github.com/jj-vcs/jj/issues/6893

yuja:
Closes #6919

# Checklist

If applicable:


lib/src/matchers.rs
```diff
@@ -243,38 +245,37 @@ fn prefix_tree_to_visit_sets(tree: &RepoPathTree<PrefixNodeKind>) -> Visit {
...
-        dir_patterns: impl IntoIterator<Item = (D, glob::Pattern)>,
+    pub fn new<D: AsRef<RepoPath>, P: Borrow<Glob>>(
+        dir_patterns: impl IntoIterator<Item = (D, P)>,
     ) -> Self {
-        let mut tree: RepoPathTree<Vec<glob::Pattern>> = Default::default();
+        let mut tree: RepoPathTree<Vec<regex::bytes::Regex>> = Default::default();
         for (dir, pattern) in dir_patterns {
-            tree.add(dir.as_ref()).value.push(pattern);
+            // Based on new_regex() in globset. We don't use GlobMatcher because
+            // RepoPath separator should be "/" on all platforms.
```
lf-, line 259: is this consistent with the test above in `fileset.rs` being disabled on windows?

lib/src/matchers.rs
```diff
@@ -243,38 +245,37 @@ fn prefix_tree_to_visit_sets(tree: &RepoPathTree<PrefixNodeKind>) -> Visit {
...
         }
         FileGlobsMatcher { tree }
     }
 }
 
 impl Matcher for FileGlobsMatcher {
     fn matches(&self, file: &RepoPath) -> bool {
-        // TODO: glob::Pattern relies on path::is_separator() internally, but
-        // RepoPath separator should be '/'. One way to address this problem is
-        // to switch to globset::Glob, and use the underlying regex pattern.
```
lf-, line 265: this might mean that windows-only users experience a perceived regression if they were using `\` in a file glob, right?

lib/src/str_util.rs
```diff
@@ -54,10 +67,22 @@ impl Debug for GlobPattern {
...
-    glob::Pattern::new(src)
-        .map(GlobPattern)
-        .map_err(StringPatternParseError::GlobPattern)
+fn parse_glob(src: &str, icase: bool) -> Result<GlobPattern, StringPatternParseError> {
+    let glob = GlobBuilder::new(src)
+        .case_insensitive(icase)
+        // Don't use platform-dependent default. This pattern isn't meant for
+        // testing file-system paths. If backslash escape were disabled, "\" in
+        // pattern would be normalized to "/" on Windows.
+        .backslash_escape(true)
```
lf-, line 76: as above with re to the test.

lib/src/str_util.rs
```diff
@@ -54,10 +67,22 @@ impl Debug for GlobPattern {
...
+        .build()
+        .map_err(StringPatternParseError::GlobPattern)?;
+    // Based on new_regex() in globset. We don't use GlobMatcher::is_match(path)
+    // because the input string shouldn't be normalized as path.
+    let regex = regex::bytes::RegexBuilder::new(glob.regex())
+        .dot_matches_new_line(true)
+        .build()
+        .expect("glob regex should be valid");
```
lf-, line 84: possibly this could be extracted as a function shared with `matchers.rs` to make sure future callers don't have inconsistent regex options.

CHANGELOG.md
```diff
@@ -42,6 +42,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj fix` now buffers lines from subprocesses' stderr streams and emits them a
   complete line at a time. Each line is prepended with the file name.
 
+* Glob patterns now support `{foo,bar}` syntax. There may be subtle behavior
+  changes as we use the [globset](https://crates.io/crates/globset) library now.
```
lf-, line 73:
```suggestion
* Glob patterns now support `{foo,bar}` syntax. Backslash is now consistently treated as an escape character in globs including on Windows.

  There may be other subtle behavior changes as we use the [globset](https://crates.io/crates/globset) library now.
```

lf-:
also FYI https://github.com/jj-vcs/jj/blob/008f770ef2ad01b52217a94127e3c586008e66f4/lib/src/str_util.rs#L260-L265 this comment becomes outdated I think!

lib/src/matchers.rs
yuja: I think so. The test for backslash-escape is disabled on Windows because backslash is a directory separator on Windows, and should be transformed to "/".

lib/src/matchers.rs
yuja: This was a bug in the old implementation. I have no idea if the problem could be triggered on Windows, but that would be rare.

lib/src/str_util.rs
yuja: If StringPattern had `.to_matcher()` step, it might make sense to add a common `Glob` wrapper or helper function. For now, the common part is just `.dot_matches_new_line(true)`, so I don't have strong motivation to extract a helper module/function. It would be odd if `matcher` dependend on `str_util`.

CHANGELOG.md
yuja: I avoided documenting backslash escape because it only applies to string pattern on Windows.

lib/src/matchers.rs
lf-: Hm, but it probably isn't transformed in practice because of `.backslach_escape(true)`, right? Or am I misunderstanding how the globs are used in `StringPattern` vs elsewhere?

We might want to have a test that verifies that backslash works as intended on Windows in places where they *are* expected to be path separators.

yuja: Backslash-escape isn't enabled for fileset (or path) patterns on Windows, and we have some test coverage in matcher.rs or fileset.rs.

yuja:
> also FYI
> 
> https://github.com/jj-vcs/jj/blob/008f770ef2ad01b52217a94127e3c586008e66f4/lib/src/str_util.rs#L260-L265
> this comment becomes outdated I think!

You mean `glob` -> `globset`? I think the comment is still valid. We'll need to decide whether we accept divergence of case-folding rules across pattern types.

Thanks for the review, btw.

lib/src/matchers.rs
lf-: ah awesome then we're good. is there a reason the fileset.rs test is disabled on windows? is it broken? should we just re-enable it?

yuja: I don't remember which one we're talking about, but that's probably because the result differs on Windows, and I don't have a strong reason to test that behavior on Windows. (It's tedious to update Windows-based test output if you don't have a Windows VM.)

martinvonz:
What's left on this PR? Does it just need a review from a maintainer? I have not started reviewing it because there are unresolved comments. Are those unresolved because we're waiting for @lf- to approve of them?

lf-:
Not sure, I think this can just be merged.

Issue #5894, opened 2025-03-06
FR: diff command should default to `--git` when there is no TTY (e.g. writing a file)


afranchuk:
**Is your feature request related to a problem? Please describe.**
I've accidentally written patch files with the wrong format a few times by writing `jj diff > file.patch` rather than `jj diff --git > file.patch`.

**Describe the solution you'd like**
I think it would be reasonable to default to `--git` when there is no TTY, as it's more than likely that the diff will be shared with other tools (which overwhelmingly support the unified diff format).

**Describe alternatives you've considered**
This behavior could be configurable. It could also be default off if there's some concern around the change in default behavior?

yuja:
I don't like the idea of changing the output format drastically depending on tty or not. It would be confusing if external pager was involved, for example.

I think user can add alias `gen-patch = ['diff', '--git']` or something?

ilyagr:
I'm on the fence, but it might make sense to change the format when color is disabled.

Though, it might be better to make a native diff format that's easy to see without color, e.g. skip lines and put pluses or minuses in between.

avamsi:
Was going to file a similar issue -- this is a problem with coding agents as well, IME (I have to instruct them to use --git). I see we have max-inline-alternation now (https://github.com/jj-vcs/jj/blob/main/docs/config.md#color-words-diff-options), perhaps we use 0 when color is disabled and 3 otherwise?

avamsi:
@yuja I'm not sure I understand your position (re: https://github.com/jj-vcs/jj/issues/5894#issuecomment-2705719838) -- my thinking is that if jj decides to render colors, it's okay to inline but if it decides to not render colors (whatever that scenario may be), we should simply not inline because that leads to a bad user (and agent) experience.

yuja:
> [@yuja](https://github.com/yuja) I'm not sure I understand your position (re: [#5894 (comment)](https://github.com/jj-vcs/jj/issues/5894#issuecomment-2705719838)) -- my thinking is that if jj decides to render colors, it's okay to inline but if it decides to not render colors (whatever that scenario may be), we should simply not inline

Correct. I don't like the idea of selecting diff formats based on tty or not, but I think it's okay to slightly adjust the parameters so that the output can be parsable without colors.

PR #6972, opened 2025-07-12
cli: support using a separate diff format for human consumption

zhaofengli:
A human user may prefer a diff format that is distinct from one intended for machine consumption.

This change adds a new configuration option, `ui.human-diff-formatter`, that specifies the default diff format to be used when the output is directed at a human user. Currently, we assume the output to be human-directed if stdout is a terminal or a pager is used.

```console
$ jj config list
[snip]
ui.diff-formatter = ":git"
ui.human-diff-formatter = ":color-words"
$ jj show                # uses :color-words
$ jj show >working.patch # uses :git
```

If `ui.human-diff-formatter` is unset, the value of `ui.diff-formatter` is used instead.

Unlike human users who typically support the `:color-words` format, most existing tools find it to be non-legible. In the future, we may want to adopt the following defaults:

```toml
[ui]
diff-formatter = ":git"
human-diff-formatter = ":color-words"
```

# Checklist

I will update the changelog and documentation if we agree on the approach.

If applicable:


yuja:
Is it tedious to specify `--git`? I personally don't like changing output significantly based on tty or not.

zhaofengli:
> Is it tedious to specify `--git`? I personally don't like changing output significantly based on tty or not.

I think it makes sense from a UX perspective. I often run `jj diff`, skim through the `difft` diff, press up in the shell, then add `>wip.patch` at the end which doesn't result in a valid patch. And I just came across #5894 so it appears that I'm not the only one who expects that piping `jj {diff,show}` yields a usable patch.

Moreover, this PR doesn't change the existing behavior and merely makes it possible for people to opt into such a setup themselves.

emilazy:
I think the default `jj diff … | cat` is pretty useless, since you can’t extract an actual diff from it at all (even visually).

yuja:
> I think the default `jj diff … | cat` is pretty useless, since you can’t extract an actual diff from it at all (even visually).

I agree about this point. We could fix that by disabling inlining if the output doesn't support colors, though.

ilyagr:
In theory, we could one day also offer an inlining format that doesn't need color. Git can do something like this with  `git diff --word-diff --color=never`. It may or may not be worth it.

Here's an example. I think there's a bit of a learning curve, though some things are easy to pick up.

```diff
diff --git a/cli/src/config/merge_tools.toml b/cli/src/config/merge_tools.toml
index e77b05bac..6adca3604 100644
--- a/cli/src/config/merge_tools.toml
+++ b/cli/src/config/merge_tools.toml
@@ -7,16 +7,17 @@ program="diffedit3"
# We use a random port as a fallback if all 5 of the preferred port numbers are
# busy.
edit-args = ["$left", "$right", "$output", "--port", "17376-17380", "--port", "0"]
{+# Not the best diff viewer, but functional. It's possible to hide the 3rd pane in the UI.+}
{+diff-args = ["$left", "$right", "--port", "17376-17380", "--port", "0"]+}

[merge-tools.diffedit3-ssh]
program="diffedit3"
# [-17376 is a verified random number, as in https://xkcd.com/221/ :). I am trying-]
[-# to avoid 8000 or 8080 in case those, more commonly used, port numbers are used-]
[-#-]{+See comments+} for [-something else.-]{+`diffedit3` above.+}
#
# [-We do-]{+Unlike `diffedit3`, `diffedit3-ssh` does+} NOT use a random port as a fallback
{+#+} since we recommend that the user[-#-] configure SSH to forward these 5 ports
edit-args = ["$left", "$right", "$output", "--port", "17376-17380", "--no-browser"]
{+diff-args = ["$left", "$right", "--port", "17376-17380", "--no-browser"]+}

[merge-tools.difft]
diff-args = ["--color=always", "$left", "$right"]
```

xzfc:
> I personally don't like changing output significantly based on tty or not.

By default, `git diff --word-diff` uses markers (`{+` `+}` `[-` `-]`) with colors for tty, and markers without colors for non-tty. (tho, `--word-diff=color` don't use markers just like jj) [\[doc\]](https://github.com/git/git/blob/v2.53.0/Documentation/diff-options.adoc?plain=1#L433-L453)

If we add these markers to non-tty jj, should we also add them for colored tty for consistency?

PR #6974, opened 2025-07-13
Fix bug I recently introduces in stream diff

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -1007,36 +1007,8 @@ impl Stream for TreeDiffStreamImpl<'_> {
 
         // Now emit the first file, or the first tree that completed with an error
         if let Some(entry) = self.items.first_entry() {
-            match entry.get() {
-                Err(_) => {
-                    // File or tree with error
-                    let (path, result) = entry.remove_entry();
-                    Poll::Ready(Some(match result {
-                        Err(err) => TreeDiffEntry {
-                            path,
-                            values: Err(err),
-                        },
-                        Ok((before, after)) => TreeDiffEntry {
-                            path,
-                            values: Ok((before, after)),
-                        },
-                    }))
-                }
-                Ok(_) => {
-                    // A diff with a file on at least one side
-                    let (path, result) = entry.remove_entry();
-                    Poll::Ready(Some(match result {
-                        Err(err) => TreeDiffEntry {
-                            path,
-                            values: Err(err),
-                        },
-                        Ok((before, after)) => TreeDiffEntry {
-                            path,
-                            values: Ok((before, after)),
-                        },
-                    }))
-                }
-            }
+            let (path, values) = entry.remove_entry();
+            Poll::Ready(Some(TreeDiffEntry { path, values }))
         } else if self.pending_trees.is_empty() {
             Poll::Ready(None)
         } else {
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -860,7 +860,8 @@ pub struct TreeDiffStreamImpl<'matcher> {
     matcher: &'matcher dyn Matcher,
     /// Pairs of tree values that may or may not be ready to emit, sorted in the
     /// order we want to emit them. If either side is a tree, there will be
-    /// a corresponding entry in `pending_trees`.
+    /// a corresponding entry in `pending_trees`. The item is ready to emit
+    /// unless there's a smaller or equal path in `pending_trees`.
     items: BTreeMap<RepoPathBuf, BackendResult<(MergedTreeValue, MergedTreeValue)>>,
     // TODO: Is it better to combine this and `items` into a single map?
     #[expect(clippy::type_complexity)]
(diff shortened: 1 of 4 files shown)


lib/testutils/src/test_backend.rs
```diff
@@ -194,134 +216,172 @@ impl Backend for TestBackend {
...
+                    hash: id.hex(),
+                    source: format!("at path {path:?}").into(),
+                }),
+                Some(contents) => {
+                    let reader: Pin<Box<dyn AsyncRead + Send>> = Box::pin(Cursor::new(contents));
+                    Ok(reader)
+                }
+            }
+        })
+        .await
```
yuja, line 233: Maybe we can instead add a `Backend` wrapper that inserts random delay and forwards requests to `TestBackend`? I think that's easier to follow than directly wrapping the TestBackend function body.

lib/src/merged_tree.rs
```diff
@@ -862,7 +871,7 @@ pub struct TreeDiffStreamImpl<'matcher> {
     /// Pairs of tree values that may or may not be ready to emit, sorted in the
     /// order we want to emit them. If either side is a tree, there will be
     /// a corresponding entry in `pending_trees`.
-    items: BTreeMap<RepoPathBuf, BackendResult<(MergedTreeValue, MergedTreeValue)>>,
+    items: BTreeMap<RepoPathBuf, DiffStreamItem>,
     // TODO: Is it better to combine this and `items` into a single map?
     #[expect(clippy::type_complexity)]
     pending_trees: VecDeque<(
```
yuja: Another idea is to keep `pending_trees` sorted, and compare first `items` and `pending_trees`. Since we need to emit result in that order, it might be actually good to poll pending trees in that order.

martinvonz: Done. That turned out to be much simpler. Thanks!

lib/testutils/src/test_backend.rs
martinvonz: Yeah, I considered doing it that way but I'm not sure it would be worth it. It would certainly be simpler to understand, but it means we need another backend. I actually meant to remove the sleep because it makes the property-based tests too slow (done now). I think refactoring it into a Backend adapter makes more sense if we want to add back the sleep and have two separate (fast and slow) backends.

lib/src/merged_tree.rs
```diff
@@ -975,14 +973,17 @@ impl<'matcher> TreeDiffStreamImpl<'matcher> {
...
-        let mut pending_index = 0;
-        while pending_index < self.pending_trees.len()
-            && (pending_index < self.max_concurrent_reads
-                || self.items.len() < self.max_queued_items)
+        let mut current_dir: Bound<RepoPathBuf> = Bound::Unbounded;
+        let mut num_polled = 0;
+        while let Some((dir, future)) = self
+            .pending_trees
+            .range_mut((current_dir, Bound::Unbounded))
+            .next()
```
yuja: nit: since this has to look up `current_dir` inside a loop, I think it's simpler to use plain `for _ in &mut ` loop and delete completed futures by later pass.

lib/src/merged_tree.rs
```diff
@@ -1011,8 +1012,16 @@ impl Stream for TreeDiffStreamImpl<'_> {
         self.poll_tree_futures(cx);
 
         // Now emit the first file, or the first tree that completed with an error
-        if let Some(entry) = self.items.first_entry() {
-            let (path, values) = entry.remove_entry();
+        if let Some((path, _)) = self.items.first_key_value() {
+            // Check if there are any pending trees earlier that we need to finish polling
+            // before we can emit this item.
+            if let Some((dir, _)) = self.pending_trees.first_key_value() {
+                if dir <= path {
```
yuja: nit: maybe `dir < path`? If `dir == path`, child entries would be `> path`.

lib/testutils/src/test_backend.rs
yuja: I think the backend factory can always create a wrapped entry for now? I don't have a strong opinion about this, but it might be nice if we can easily remove random/async-ness when debugging tests.

yuja (approved):

lib/src/merged_tree.rs
martinvonz: Makes sense. I had to refactor it a bit more to make sure that `poll_next()` would never return `Poll::Pending` if none of the pending trees returned `Poll::Pending`. Otherwise we could get stuck with the stream never getting polled again.

lib/testutils/src/test_backend.rs
martinvonz: Right, but it would still mean that we need to create that Backend adapter, which I think would be more work than what I have here. I suspect we may want to do that at some point but I prefer to leave it for later.

lib/src/merged_tree.rs
martinvonz: True, even if this is a file->dir or dir->file transition, we should be able to emit the item without waiting. Done.

martinvonz:
I suspect the change to poll futures in order by path is buggy because some pending future now may not have a `Waker` from a previous call to `poll_next()`. From https://doc.rust-lang.org/src/core/future/future.rs.html#65-67:

```
    /// Note that on multiple calls to `poll`, only the [`Waker`] from the
    /// [`Context`] passed to the most recent call should be scheduled to
    /// receive a wakeup.
```

I'll try to fix this soon (not sure how I'm going to fix it).

yuja:
Good catch. Maybe we can split `pending_trees` into scheduled futures and yet-scheduled trees? For the former, I think we can use `FuturesUnordered`.

martinvonz:
Ah, that's a good point about using `FuturesUnordered` to simplify the polling a bit.

PR #6987, opened 2025-07-14
cli: move `desc` and `st` aliases to config

martinvonz:
I'm sorry if we have already discussed this and rejected the idea for some reason.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/describe.rs
+++ b/cli/src/commands/describe.rs
@@ -39,12 +39,11 @@ use crate::description_util::ParsedBulkEditMessage;
 use crate::text_util::parse_author;
 use crate::ui::Ui;
 
-/// Update the change description or other metadata
+/// Update the change description or other metadata [default alias: desc]
 ///
 /// Starts an editor to let you edit the description of changes. The editor
 /// will be $EDITOR, or `pico` if that's not defined (`Notepad` on Windows).
 #[derive(clap::Args, Clone, Debug)]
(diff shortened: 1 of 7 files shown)


yuja (approved):
One potential problem is/was completion, but I think it works these days.

@ilyagr might remember something else.

martinvonz:
> One potential problem is/was completion, but I think it works these days.

Oh, good point. It doesn't seem to work if I test with the existing `jj amend` alias in Fish.

yuja:
Right. Completion of aliases works only when dynamic completion is enabled.

martinvonz:
Ah, I thought I had that set up but turns out I had not. This PR seems fine then but I'll leave it up for a while to give others a chance to comment.

ilyagr:
I think this is good, but we can and probably should also add `[default alias]` to the help text, like we did for `jj bookmark`.

ilyagr:
Another related issue is the "don't complete short aliases" issue that I don't know what to do about. In my eyes, that's not blocking.

martinvonz:
> I think this is good, but we can and probably should also add `[default alias]` to the help text, like we did for `jj bookmark`.

Done.

martinvonz:
> This PR seems fine then but I'll leave it up for a while to give others a chance to comment.

Sorry, I forgot about this. We can roll it back if anyone has objections.

PR #6988, opened 2025-07-14
diffedit: add support for filesets

### LINKED ISSUE #6973: FR: Accept fileset argument to `jj diffedit`
Sometimes I want to edit a particular set of files when running `jj diffedit`, but it doesn't support an optional fileset like `jj diff`:

```
jj diff [OPTIONS] [FILESETS]...
```

Would it be possible to add support for `[FILESETS]...` to `jj diffedit` too?

gechr:
Closes #6973

### DIFF
--- a/cli/src/commands/diffedit.rs
+++ b/cli/src/commands/diffedit.rs
@@ -17,11 +17,11 @@ use std::io::Write as _;
 use clap_complete::ArgValueCandidates;
 use clap_complete::ArgValueCompleter;
 use itertools::Itertools as _;
-use jj_lib::matchers::EverythingMatcher;
 use jj_lib::object_id::ObjectId as _;
 use jj_lib::rewrite::merge_commit_trees;
 use tracing::instrument;
 
+use crate::cli_util::print_unmatched_explicit_paths;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::RevisionArg;
 use crate::command_error::CommandError;
@@ -77,6 +77,13 @@ pub(crate) struct DiffeditArgs {
         add = ArgValueCompleter::new(complete::revset_expression_mutable),
     )]
     to: Option<RevisionArg>,
+    /// Restrict the diff to these paths
+    #[arg(
+        value_name = "FILESETS",
+        value_hint = clap::ValueHint::AnyPath,
+        add = ArgValueCompleter::new(complete::modified_revision_or_range_files),
+    )]
+    paths: Vec<String>,
     /// Specify diff editor to be used
     #[arg(
         long,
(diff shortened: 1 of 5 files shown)


cli/src/commands/diffedit.rs
```diff
@@ -136,7 +145,7 @@ don't make any changes, then the operation will be aborted.",
     };
     let base_tree = merge_commit_trees(tx.repo(), base_commits.as_slice())?;
     let tree = target_commit.tree()?;
-    let tree_id = diff_editor.edit(&base_tree, &tree, &EverythingMatcher, format_instructions)?;
+    let tree_id = diff_editor.edit(&base_tree, &tree, &matcher, format_instructions)?;
```
martinvonz, line 148: [This FIXME](https://github.com/jj-vcs/jj/blob/4201af2107a8003e04ba81b68697eb6c577a5542/cli/src/merge_tools/mod.rs#L286C8-L292) makes me think that this doesn't do what you (and I) would want when using the internal diff editor. Can you check?

gechr: Would something like this suffice as a workaround for now?

```diff
+
+    // First, create trees that only differ in files matching the matcher.
+    // For unmatched files, we use the left tree's version.
+    let right_tree_filtered_id = restore_tree(right_tree, left_tree, matcher).block_on()?;
+    let right_tree_filtered = store.get_root_tree(&right_tree_filtered_id)?;
+
     let diffedit_wc = DiffEditWorkingCopies::check_out(
         store,
         left_tree,
-        right_tree,
-        matcher,
+        &right_tree_filtered,
+        &EverythingMatcher,
```

If not, can you (or @yuja, who added the original comment) provide some further guidance on the desired end result here, since the comment isn't too clear, i.e. _"**If** we adopt the interpretation of the builtin tool"_

martinvonz: I think we should fix it by making the internal tool match external tools, i.e. by making `edit_diff_builtin()` keep the right tree for unmatched paths. That could perhaps be done by making `apply_diff_builtin()` create the `tree_builder` from `right_tree` and adjusting `apply_changes()` accordingly. I'll let you figure out the details.

gechr: Thanks! I'll circle back to this when I have some time.

I don't like leaving PRs open indefinitely, so I'll close it for now. If anyone else wants to pick it up in the meantime, feel free.

gechr: Think I found a quick win which didn't take too much time: https://github.com/jj-vcs/jj/pull/6995

(I can't re-open this PR because the branch was deleted)

yuja: https://github.com/jj-vcs/jj/issues/5390 can also be related. If the default state was all selected, it might make sense to keep files excluded by filesets as selected. (iirc, the external tools behave differently, though.)

PR #3670, opened 2024-05-12
Add workspace config

HybridEidolon:
This adds support for per-workspace configuration as was described as a TODO in the LayeredConfig struct. Like repo configuration in `.jj/repo/config.toml`, the workspace config goes in `.jj/config.toml` in the workspace root, and its settings take precedence over repo, user and default configs.

The config subcommands that take `--user` and `--repo` are additionally given `--workspace` which behaves as one would expect.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [x] I have updated the documentation (README.md, docs/, demos/)
- [x] I have added tests to cover my changes

Issue #1595, opened 2023-05-10
In-repo configs are a potential security risk


chooglen:
## Description

Git has longstanding security issues with .git/config being inside the repo. The config settings that specify which executable to run are extremely prone to abuse. Notably, core.fsmonitor is invoked on git status, which is run if the user has configured some sort of Git prompt and `cd`-s into a repo, but even more innocuous-looking settings like `core.pager` are quite prone to this sort of abuse. This is pretty scary because to perform an attack, the attacker only needs to convince the victim to unzip a repo with a poisoned config file and wait for them to run either `cd` or `git log`.

jj already has similar settings (`ui.merge-editor`, `ui.pager`, etc), so we're also prone to such attacks. if possible, i'd like to find a safer way around in-repo config before it becomes unsolvable. the problem of config in the repo has become pretty much unsolvable for Git already. recent security fixes in this area have focused on which repos to trust (e.g. `safe.directory` says that repos owned by another user are untrustworthy). see another discussion from a prominent Git contributor about limiting which executables Git can run  https://lore.kernel.org/git/6af83767-576b-75c4-c778-0284344a8fe7@github.com/ (See "Example Security Boundary Question: Unprotected Config and Executing Code").

Other proposals to fix this issue:

- Safe-listing which config that we will read from the repo config, and everything else will be ignored. This is probably doable at our current scale of config options, but it precludes some desirable features, e.g. configuring a different merge tool for a specialized repo.
- Keeping per-repo config outside of the repo. This could be made `mv`-resistant by associating a UUID with a repo (thanks @Ralith for the idea). This probably requires implementing an import/export workflow for copying repos, but this lets us sanitize possibly unsafe config, so this is a lot safer than relying on `zip`.

necauqua:
It's probably possible to do both

Safe and _very_ common options like `user.name`/`user.email` can be in-repo, and the few unsafe options, the ones that deal with executables, can be kept outside of the repo (if even needed) and ignored for in-repo config.

dbarnett:
Would this warn about any config options it's ignoring?

I also wonder if we'd have use cases for semi-automating any of those, like having a one-line way for your configs _outside_ the repo to load e.g. a set of merge tool configurations from inside the repo.

yuja:
There's a new proposal to address this problem:
https://github.com/jj-vcs/jj/pull/8046

PR #6990, opened 2025-07-15
config: Allow both user-managed and repo-managed configs per repo

### LINKED ISSUE #3684: FR: Per-repo jj configuration
I believe that there are 4 different types of configuration that may be useful for jj:
* Configuration useful for everybody using jj (default configs)
* Configuration useful for just me, but for all the repos on my system (`~/.config/jj/config.toml`)
* Configuration useful for all users of a particular repo
* Configuration useful for just me on a particular repo

Sometimes it's quite handy to be able to have per-repo configuration. For example:
* One repo might be the exception to the rule, and thus I need to come up with a custom revset (eg. `immutable_heads()`)
* Potential configuration for `jj fix` and `jj lint` in the future
* Pre-upload hook configuration (#3577)

`jj` already supports layering of configuration via the `--config-toml` argument.

This is very similar to [bazel](https://bazel.build/run/bazelrc#bazelrc-file-locations)'s `.bazelrc` files, which support workspace, user, and user-specified rc files (and most workspaces have `try-import %workspace%/user.bazelrc` at the bottom of their files, so for all intents and purposes, that's supported too).

If it weren't for security concerns, I'd think we should adopt that model in an instant, but since bazel isn't worried about arbitrary code execution (that's literally it's job), we unfortunately have to go further. This is very similar to git hooks and their security model, so I propose the following changes.

* Read an allowlisted subset of configuration from `<repo>/.jj/repo/repo_config.toml`
* Read a full set of configuration from `<repo>/.jj/repo/user_config.toml` (path open for discussion)

However, to deal with security concerns, we print a warning and ignore the file if either file is tracked by jj. This will prevent someone upstream sending you a file and allowing arbitrary code execution.

The next change we make is very simple. When `jj` starts, we read `<repo>/jj/config.toml` (path open for discussion), and, if the file doesn't match `<repo>/.jj/repo/repo_config.toml`, we print a warning that there is a more up to date configuration, and instruct them to run a command `jj install-config` or similar.

`jj install-config` would print a diff, warn them about the security concerns, and ask the user if they want to install the config to the repo (copying the file to `<repo>/.jj/repo/repo_config.toml`).

We could potentially instead put repo-specific conifguration as its own data format, but I think this is much more flexible and easy to use

**Additional context**

matts1:
Fixes #3684

This adds the following workflow to jj to support repo-level workflows:
1) Repo updates config.toml in the VCS
2) User retrieves changes
3) User runs jj command, gets a warning saying "please run `jj config update-repo`"
4) `jj config update-repo` adds a TUI requesting the user to approve the diff
5) The approved changes are copied to a permanent config file.

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,13 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `jj util exec` sets the environment variable `JJ_WORKSPACE_ROOT`
 
+* Introduced an additional config layer `repo-managed` between `user` and
+  `repo`. This layer is stored in version control, and if it ever changes,
+  we recommend to the user to review the changes (for security reasons) by
+  running the new command `jj config review-managed`. This can be used to
+  create config shared between all users who use a repository (eg.
+  [fix commands](https://chromium-review.googlesource.com/c/chromium/src/+/6703030))
+
 ### Fixed bugs
 
 * Fetching repositories that have submodules no longer errors even if
--- a/cli/src/commands/config/review_managed.rs
+++ b/cli/src/commands/config/review_managed.rs
@@ -1,0 +1,125 @@
+// Copyright 2025 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::path::Path;
+
+use tracing::instrument;
+
+use crate::cli_util::CommandHelper;
+use crate::command_error::CommandError;
+use crate::command_error::internal_error_with_message;
+use crate::command_error::user_error;
+use crate::command_error::user_error_with_message;
+use crate::merge_tools::make_diff_sections;
+use crate::repo_managed_config::MANAGED_PATH;
+use crate::repo_managed_config::RepoManagedConfig;
+use crate::ui::Ui;
+
+/// Reviews and updates configuration stored in version control.
+/// You should never need to run this command unless jj tells you to.
+/// This command needs to be run when the config checked in to the repo is
+/// changed, and allows you to approve or reject said changes on a line-by-line
+/// basis.
+#[derive(clap::Args, Clone, Debug)]
+pub struct ConfigReviewManagedArgs {
+    /// Trust the repository's config and skip review of it.
+    /// Use this when you absolutely trust the repo config (eg. you're the only
+    /// contributor).
+    #[arg(long)]
+    trust: bool,
+}
+
+#[instrument(skip_all)]
+pub fn cmd_review_managed(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &ConfigReviewManagedArgs,
+) -> Result<(), CommandError> {
+    if let Some(repo_config) = command.config_env().repo_managed_config()? {
+        let workspace_command = command.workspace_helper(ui)?;
+        let workspace_root = workspace_command.workspace_root();
+        let vcs = match repo_config.get_vcs_config(workspace_root)? {
+            Some(vcs) => vcs,
+            None => {
+                // We don't need the user to review this since it's not a security issue.
+                writeln!(ui.status(), "Your config doesn't need review")?;
+                return Ok(());
+            }
+        };
+
+        let new_config = if args.trust {
+            vcs.clone()
+        } else {
+            let old_config = repo_config
+                .last_approved()
+                .ok()
+                .flatten()
+                .and_then(|last_approved| std::fs::read_to_string(last_approved).ok())
+                .unwrap_or_default();
+            let sections = make_diff_sections(
+                &old_config,
+                &String::from_utf8(vcs.clone()).map_err(|e| {
+                    user_error_with_message("Config stored in VCS was not utf-8", e)
+                })?,
+            )
+            .map_err(|e| internal_error_with_message("Failed to create diff sections", e))?;
+            // Ideally we'd use the user's chosen diff selector, but that
+            // heavily relies on jj's objects such as Tree and Store.
+            let recorded = scm_record::Recorder::new(
+                scm_record::RecordState {
+                    is_read_only: false,
+                    commits: vec![],
+                    files: vec![scm_record::File {
+                        old_path: None,
+                        path: std::borrow::Cow::Borrowed(Path::new(MANAGED_PATH)),
+                        // This doesn't do anything.
+                        file_mode: scm_record::FileMode::Unix(0o777),
+                        sections,
+                    }],
+                },
+                &mut scm_record::helpers::CrosstermInput,
+            )
+            .run()
+            .map_err(|_| user_error("Failed to select changes"))?;
+
+            // There's always precisely one file.
+            reconstruct(&recorded.files[0].sections).into_bytes()
+        };
+        repo_config.approve_content(&RepoManagedConfig::digest(&vcs), &new_config)?;
+        writeln!(ui.status(), "Updated repo config file")?;
+        Ok(())
+    } else {
+        Err(user_error(
+            "Unable to detect location of config files. Are you in a repo?",
+        ))
+    }
+}
+
+fn reconstruct(sections: &[scm_record::Section]) -> String {
+    let mut out: Vec<&str> = Default::default();
+    for section in sections {
+        match section {
+            scm_record::Section::Unchanged { lines } => out.extend(lines.iter().map(AsRef::as_ref)),
+            scm_record::Section::Changed { lines } => {
+                for line in lines {
+                    if line.is_checked == (line.change_type == scm_record::ChangeType::Added) {
+                        out.push(&line.line);
+                    }
+                }
+            }
+            _ => {}
+        }
+    }
+    out.join("")
+}
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -44,6 +44,8 @@ use tracing::instrument;
 use crate::command_error::CommandError;
 use crate::command_error::config_error;
 use crate::command_error::config_error_with_message;
+use crate::command_error::internal_error_with_message;
+use crate::repo_managed_config::RepoManagedConfig;
 use crate::text_util;
 use crate::ui::Ui;
 
@@ -372,7 +374,9 @@ impl UnresolvedConfigEnv {
 #[derive(Clone, Debug)]
 pub struct ConfigEnv {
     home_dir: Option<PathBuf>,
+    state_dir: Option<PathBuf>,
     repo_path: Option<PathBuf>,
+    workspace_root: Option<PathBuf>,
     user_config_paths: Vec<ConfigPath>,
     repo_config_path: Option<ConfigPath>,
     command: Option<String>,
@@ -381,9 +385,8 @@ pub struct ConfigEnv {
 impl ConfigEnv {
     /// Initializes configuration loader based on environment variables.
     pub fn from_environment(ui: &Ui) -> Self {
-        let config_dir = etcetera::choose_base_strategy()
-            .ok()
-            .map(|s| s.config_dir());
+        let strategy = etcetera::choose_base_strategy().ok();
+        let config_dir = strategy.as_ref().map(|s| s.config_dir());
 
         // older versions of jj used a more "GUI" config option,
         // which is not designed for user-editable configuration of CLI utilities.
@@ -409,6 +412,11 @@ impl ConfigEnv {
             .ok()
             .map(|d| dunce::canonicalize(&d).unwrap_or(d));
 
+        let state_dir = strategy
+            .and_then(|s| s.state_dir())
+            .map(|d| dunce::canonicalize(&d).unwrap_or(d))
+            .or_else(|| home_dir.as_deref().map(|d| d.join(".local/state/jj")));
+
         let env = UnresolvedConfigEnv {
             config_dir,
             macos_legacy_config_dir,
@@ -417,7 +425,9 @@ impl ConfigEnv {
         };
         Self {
             home_dir,
+            state_dir,
             repo_path: None,
+            workspace_root: None,
             user_config_paths: env.resolve(ui),
             repo_config_path: None,
             command: None,
@@ -488,9 +498,33 @@ impl ConfigEnv {
 
     /// Sets the directory where repo-specific config file is stored. The path
     /// is usually `.jj/repo`.
-    pub fn reset_repo_path(&mut self, path: &Path) {
-        self.repo_path = Some(path.to_owned());
-        self.repo_config_path = Some(ConfigPath::new(path.join("config.toml")));
+    pub fn reset_repo_path(&mut self, repo_path: &Path, workspace_root: &Path) {
+        self.repo_path = Some(repo_path.to_owned());
+        self.workspace_root = Some(workspace_root.to_owned());
+        self.repo_config_path = Some(ConfigPath::new(repo_path.join("config.toml")));
+    }
+
+    /// Returns the repo-managed configuration.
+    pub fn repo_managed_config(&self) -> Result<Option<RepoManagedConfig>, CommandError> {
+        let workspace_root = self.workspace_root.as_deref().map(|d| {
+            dunce::canonicalize(d)
+                .ok()
+                .unwrap_or_else(|| d.to_path_buf())
+        });
+        Ok(
+            match (
+                self.state_dir.as_deref(),
+                self.repo_path.as_deref(),
+                workspace_root.as_deref(),
+            ) {
+                (Some(state_dir), Some(repo_path), Some(workspace_root)) => Some(
+                    RepoManagedConfig::new(state_dir, repo_path, workspace_root).map_err(|e| {
+                        internal_error_with_message("Unable to convert path to bytes", e)
+                    })?,
+                ),
+                _ => None,
+            },
+        )
     }
 
     /// Returns a path to the repo-specific config file.
@@ -527,6 +561,27 @@ impl ConfigEnv {
             .transpose()
     }
 
+    /// Loads repo-managed config file for the user into the given `config`.
+    /// Returns true if a layer was loaded, and false if no layer was loaded.
+    #[instrument(skip(ui))]
+    fn try_reload_repo_managed_config(
+        &self,
+        ui: &Ui,
+        config: &mut RawConfig,
+    ) -> Result<bool, CommandError> {
+        if let Some(repo_config) = self.repo_managed_config()? {
+            let workspace_root = self
+                .workspace_root
+                .as_deref()
+                .expect("workspace root must be set");
+            if let Some(layer) = repo_config.create_layer(ui, workspace_root)? {
+                config.as_mut().add_layer(layer);
+                return Ok(true);
+            }
+        }
+        Ok(false)
+    }
+
     /// Loads repo-specific config file into the given `config`. The old
     /// repo-config layer will be replaced if any.
     #[instrument]
@@ -538,15 +593,40 @@ impl ConfigEnv {
         Ok(())
     }
 
-    /// Resolves conditional scopes within the current environment. Returns new
-    /// resolved config.
-    pub fn resolve_config(&self, config: &RawConfig) -> Result<StackedConfig, ConfigGetError> {
+    pub fn resolve_config_without_repo_managed(
+        &self,
+        config: &RawConfig,
+    ) -> Result<StackedConfig, ConfigGetError> {
         let context = ConfigResolutionContext {
             home_dir: self.home_dir.as_deref(),
             repo_path: self.repo_path.as_deref(),
             command: self.command.as_deref(),
         };
         jj_lib::config::resolve(config.as_ref(), &context)
+    }
+
+    /// Resolves conditional scopes within the current environment. Returns new
+    /// resolved config.
+    pub fn resolve_config(
+        &self,
+        ui: &Ui,
+        config: &mut RawConfig,
+    ) -> Result<StackedConfig, CommandError> {
+        let mut resolved = self.resolve_config_without_repo_managed(config)?;
+        // In case you call this function twice, we remove the layers so we don't
+        // duplicate.
+        resolved.remove_layers(ConfigSource::RepoManaged);
+        // The repo-managed config is enabled.
+        Ok(
+            if resolved.get("repo-managed-config.enabled")?
+                && self.try_reload_repo_managed_config(ui, config)?
+            {
+                // Now that we've updated config, we need to re-resolve it.
+                self.resolve_config_without_repo_managed(config)?
+            } else {
+                resolved
+            },
+        )
     }
 }
 
@@ -1811,7 +1891,9 @@ mod tests {
         };
         ConfigEnv {
             home_dir,
+            state_dir: None,
             repo_path: None,
+            workspace_root: None,
             user_config_paths: env.resolve(&Ui::null()),
             repo_config_path: None,
             command: None,
--- a/cli/src/repo_managed_config.rs
+++ b/cli/src/repo_managed_config.rs
@@ -1,0 +1,220 @@
+// Copyright 2025 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::io;
+use std::io::Write as _;
+use std::path::Path;
+use std::path::PathBuf;
+use std::sync::atomic::AtomicBool;
+
+use jj_lib::config::ConfigLayer;
+use jj_lib::config::ConfigSource;
+use jj_lib::content_hash::blake2b_hash;
+use jj_lib::file_util;
+use jj_lib::file_util::BadPathEncoding;
+use jj_lib::file_util::IoResultExt as _;
+use jj_lib::file_util::PathError;
+use jj_lib::file_util::persist_temp_file;
+use jj_lib::hex_util;
+use jj_lib::repo_path::RepoPathBuf;
+use tempfile::NamedTempFile;
+
+use crate::command_error::CommandError;
+use crate::ui::Ui;
+
+pub const MANAGED_PATH: &str = ".config/jj/config.toml";
+pub const LAST_APPROVED: &str = "last_approved";
+
+/// Manages the repo-managed config. The file structure for the repo-managed
+/// config for a repo contained at /path/to/repo is:
+/// ~/.local/state/jj/repos/hash(/path/to/repo)/last_approved:
+///   hash(config1)
+/// ~/.local/state/jj/repos/hash(/path/to/repo)/last_approved_hash(workspace):
+///   hash(config1)
+/// ~/.local/state/jj/repos/hash(/path/to/repo)/hash(config1):
+///   approved content for config1
+/// ~/.local/state/jj/repos/hash(/path/to/repo)/hash(config2):
+///   approved content for config2
+pub struct RepoManagedConfig {
+    config_dir: PathBuf,
+    repo_path: PathBuf,
+    last_approved_name: String,
+}
+
+static PRINTED_WARNING: AtomicBool = AtomicBool::new(false);
+
+impl RepoManagedConfig {
+    pub fn new(
+        state_dir: &Path,
+        repo_path: &Path,
+        workspace_path: &Path,
+    ) -> Result<Self, BadPathEncoding> {
+        Ok(Self {
+            config_dir: state_dir
+                .join("repos")
+                .join(Self::digest(file_util::path_to_bytes(repo_path)?)),
+            repo_path: repo_path.to_path_buf(),
+            last_approved_name: format!(
+                "{LAST_APPROVED}_{}",
+                Self::digest(file_util::path_to_bytes(workspace_path)?)
+            ),
+        })
+    }
+
+    fn write(path: &Path, content: &[u8]) -> Result<(), PathError> {
+        let mut temp_file = NamedTempFile::new_in(path.parent().unwrap()).context(path)?;
+        temp_file.as_file_mut().write_all(content).context(path)?;
+        persist_temp_file(temp_file, path).context(path)?;
+        Ok(())
+    }
+
+    fn approved_config(&self, digest: &str) -> PathBuf {
+        self.config_dir.join(format!("{digest}.toml"))
+    }
+
+    pub fn last_approved(&self) -> Result<Option<PathBuf>, PathError> {
+        let read = |path: &Path| match std::fs::read_to_string(path) {
+            Ok(content) => Ok(Some(self.approved_config(&content))),
+            Err(e) if e.kind() == io::ErrorKind::NotFound => Ok(None),
+            Err(e) => Err(e).context(path),
+        };
+
+        Ok(
+            if let Some(last_approved) = read(&self.config_dir.join(&self.last_approved_name))? {
+                Some(last_approved)
+            } else {
+                read(&self.config_dir.join(LAST_APPROVED))?
+            },
+        )
+    }
+
+    /// Updates last-approved to the current digest.
+    fn approve(&self, digest: &str) -> Result<(), PathError> {
+        // Do not use a symlink since they don't play nice with windows.
+        // Apparently it's requires privilege escalation on windows.
+        Self::write(
+            &self.config_dir.join(&self.last_approved_name),
+            digest.as_bytes(),
+        )?;
+        Self::write(&self.config_dir.join(LAST_APPROVED), digest.as_bytes())
+    }
+
+    /// Sets the approved config to content for the given digest.
+    pub fn approve_content(&self, digest: &str, content: &[u8]) -> Result<(), PathError> {
+        // This never removes old ones. This is by design. In the future, we
+        // might want to do some kind of TTL or some way to cleanup old ones.
+        // It doesn't matter that much though, since configs are likely small
+        // and infrequently changed.
+        std::fs::create_dir_all(&self.config_dir).context(&self.config_dir)?;
+        // This isn't used right now, but it may be used in the future to clean
+        // up deleted repos.
+        Self::write(
+            &self.config_dir.join("repo_path"),
+            self.repo_path.as_os_str().as_encoded_bytes(),
+        )?;
+        Self::write(&self.approved_config(digest), content)?;
+        self.approve(digest)
+    }
+
+    pub fn get_vcs_config(&self, workspace_root: &Path) -> Result<Option<Vec<u8>>, CommandError> {
+        let managed = RepoPathBuf::from_internal_string(MANAGED_PATH)
+            .unwrap()
+            .to_fs_path_unchecked(workspace_root);
+        Ok(match std::fs::read(&managed) {
+            Ok(val) => {
+                if !val.is_empty() {
+                    Some(val)
+                } else {
+                    None
+                }
+            }
+            Err(e) if e.kind() == io::ErrorKind::NotFound => None,
+            Err(e) => Err(e).context(&managed)?,
+        })
+    }
+
+    pub fn digest(content: &[u8]) -> String {
+        let hash: &[u8] = &blake2b_hash(content);
+        // There are two digests in a path (digest of repo path and digest of config
+        // file). Each digest is 512 bits, and is encoded as hex, so takes up
+        // 128 characters. This means that the pathname has 256 characters of
+        // hashes. On windows, the maximum path length is 260 characters, so we
+        // use xor to compress 512 bits into 256 so that it fits within the
+        // maximum path length.
+        #[cfg(windows)]
+        let hash = &hash
+            .as_chunks::<2>()
+            .0
+            .iter()
+            .map(|[lhs, rhs]| lhs ^ rhs)
+            .collect::<Vec<_>>();
+        hex_util::encode_hex(hash)
+    }
+
+    /// If the digest of MANAGED_PATH exists in the cache, returns
+    /// (config_file, true). Otherwise, returns
+    /// (the most recent content passed to `approve_content`, false)
+    pub fn get_config_file(
+        &self,
+        workspace_root: &Path,
+    ) -> Result<(Option<PathBuf>, bool), CommandError> {
+        let config = self.get_vcs_config(workspace_root)?;
+        if let Some(config) = config
+            && !config.is_empty()
+        {
+            let digest = Self::digest(&config);
+            let out = self.approved_config(&digest);
+            let last_approved = self.last_approved()?;
+            if out.try_exists()? {
+                // This particular config was previously approved, but we need to reset
+                // last_approved.
+                if Some(&out) != last_approved.as_ref() {
+                    self.approve(&digest)?;
+                }
+                return Ok((Some(out), true));
+            } else {
+                return Ok((last_approved, false));
+            }
+        }
+        Ok((None, true))
+    }
+
+    /// Creates the configuration layer for the repo managed config.
+    /// This layer is best-effort based on the approved repo config.
+    /// If we don't get a perfect match, it will print a warning that your
+    /// config is out of date.
+    pub fn create_layer(
+        &self,
+        ui: &Ui,
+        workspace_root: &Path,
+    ) -> Result<Option<ConfigLayer>, CommandError> {
+        let (last_approved, up_to_date) = self.get_config_file(workspace_root)?;
+        if !up_to_date && !PRINTED_WARNING.swap(true, std::sync::atomic::Ordering::AcqRel) {
+            writeln!(
+                ui.warning_default(),
+                "Your repo-managed config is out of date"
+            )?;
+            writeln!(ui.hint_default(), "Run `jj config review-managed`")?;
+        }
+
+        let last_approved = match last_approved {
+            Some(x) => x,
+            None => return Ok(None),
+        };
+        match ConfigLayer::load_from_file(ConfigSource::RepoManaged, last_approved) {
+            Ok(layer) => Ok(Some(layer)),
+            Err(e) => Err(e.into()),
+        }
+    }
+}
--- a/cli/tests/test_repo_managed_config.rs
+++ b/cli/tests/test_repo_managed_config.rs
@@ -1,0 +1,271 @@
+// Copyright 2025 The Jujutsu Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::common::TestEnvironment;
+
+#[test]
+fn test_repo_managed_config() {
+    let test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    let work_dir = test_env.work_dir("repo");
+    test_env.add_config(r#"ui.pager = "user pager""#);
+
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    "###);
+
+    work_dir.write_file(
+        ".config/jj/config.toml",
+        r#"repo-managed-config.enabled = false"#,
+    );
+
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+
+    // We have to use the --trust flag here because we can't interact with the TUI.
+    let output = work_dir.run_jj(["config", "review-managed", "--trust"]);
+    insta::assert_snapshot!(output, @r###"
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    Updated repo config file
+    [EOF]
+    "###);
+
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    "###);
+
+    test_env.add_config(r"repo-managed-config.enabled = false");
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    "###);
+
+    test_env.add_config(
+        r###"
+    [[--scope]]
+    --when.commands = ["config get"]
+    [--scope.repo-managed-config]
+    enabled = true
+    "###,
+    );
+    work_dir.write_file(
+        ".config/jj/config.toml",
+        r#"ui.pager = "repo-managed pager""#,
+    );
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+
+    let other_dir = test_env.work_dir(".");
+    let output = other_dir.run_jj([
+        "-R",
+        work_dir.root().to_str().unwrap(),
+        "config",
+        "get",
+        "ui.pager",
+    ]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+
+    // If the repo says to enable repo-managed config, but the user disallows it,
+    // it should definitely not be enabled.
+    work_dir.write_file(
+        ".jj/repo/config.toml",
+        r#"repo-managed-config.enabled = false"#,
+    );
+    work_dir.write_file(
+        ".config/jj/config.toml",
+        r#"repo-managed-config.enabled = true"#,
+    );
+    other_dir.write_file(
+        ".config/jj/config.toml",
+        r#"repo-managed-config.enabled = true"#,
+    );
+    let output = other_dir.run_jj([
+        "-R",
+        work_dir.root().to_str().unwrap(),
+        "config",
+        "get",
+        "ui.pager",
+    ]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    "###);
+
+    work_dir.write_file(".jj/repo/config.toml", "");
+
+    // We have to use the --trust flag here because we can't interact with the TUI.
+    let output = work_dir.run_jj(["config", "review-managed", "--trust"]);
+    insta::assert_snapshot!(output, @r###"
+    ------- stderr -------
+    Updated repo config file
+    [EOF]
+    "###);
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    "###);
+}
+
+#[test]
+fn test_multi_workspace_config() {
+    let test_env = TestEnvironment::default();
+    test_env.run_jj_in(".", ["git", "init", "repo"]).success();
+    test_env
+        .run_jj_in("repo", ["workspace", "add", "../second"])
+        .success();
+    let work_dir = test_env.work_dir("repo");
+    let second_dir = test_env.work_dir("second");
+    test_env.add_config(r#"ui.pager = "user pager""#);
+
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    "###);
+
+    work_dir.write_file(".config/jj/config.toml", r#"ui.pager = "repo pager""#);
+
+    // The config should be out of date in the main repo
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+
+    // But the second repo should still rely on its own local config
+    let output = second_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    "###);
+
+    second_dir.write_file(".config/jj/config.toml", r#"ui.pager = "repo pager""#);
+
+    // Now the second dir is out of date
+    let output = second_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+
+    work_dir
+        .run_jj(["config", "review-managed", "--trust"])
+        .success();
+
+    // We should now be using the new config.
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    repo pager
+    [EOF]
+    "###);
+
+    // Approving the same content for one workspace should approve it for the other.
+    let output = second_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    repo pager
+    [EOF]
+    "###);
+
+    // Now we update them divergently
+    work_dir.write_file(".config/jj/config.toml", r#"ui.pager = "repo pager v2""#);
+    second_dir.write_file(".config/jj/config.toml", r#"ui.pager = "second pager v2""#);
+
+    // Both directories should use the last approved config.
+    // This is despite the fact that the the second workspace doesn't have a
+    // last approved config, so it needs to use the global last approved config
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    repo pager
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+    let output = second_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    repo pager
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+
+    // This time we update them divergently, but both have a last approved config.
+    work_dir
+        .run_jj(["config", "review-managed", "--trust"])
+        .success();
+    second_dir
+        .run_jj(["config", "review-managed", "--trust"])
+        .success();
+    work_dir.write_file(".config/jj/config.toml", r#"ui.pager = "repo pager v3""#);
+    second_dir.write_file(".config/jj/config.toml", r#"ui.pager = "second pager v3""#);
+
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    repo pager v2
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+    let output = second_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    second pager v2
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
+    Hint: Run `jj config review-managed`
+    [EOF]
+    "###);
+}
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -359,9 +359,11 @@ impl CommandHelper {
         let mut config_env = self.data.config_env.clone();
         let mut raw_config = self.data.raw_config.clone();
         let repo_path = workspace_root.join(".jj").join("repo");
-        config_env.reset_repo_path(&repo_path);
+        config_env.reset_repo_path(&repo_path, workspace_root);
+        // Since it's a new workspace, we don't need to bother loading the
+        // repo-managed config.
         config_env.reload_repo_config(&mut raw_config)?;
-        let mut config = config_env.resolve_config(&raw_config)?;
+        let mut config = config_env.resolve_config_without_repo_managed(&raw_config)?;
         // No migration messages here, which would usually be emitted before.
         jj_lib::config::migrate(&mut config, &self.data.config_migrations)?;
         Ok(self.data.settings.with_new_config(config)?)
@@ -3820,10 +3822,10 @@ impl<'a> CliRunner<'a> {
             .map_err(|err| map_workspace_load_error(err, Some(".")));
         config_env.reload_user_config(&mut raw_config)?;
         if let Ok(loader) = &maybe_cwd_workspace_loader {
-            config_env.reset_repo_path(loader.repo_path());
+            config_env.reset_repo_path(loader.repo_path(), loader.workspace_root());
             config_env.reload_repo_config(&mut raw_config)?;
         }
-        let mut config = config_env.resolve_config(&raw_config)?;
+        let mut config = config_env.resolve_config(ui, &mut raw_config)?;
         migrate_config(&mut config)?;
         ui.reset(&config)?;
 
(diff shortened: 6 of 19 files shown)


PhilipMetzger:
Please format the commits to adhere to our commit guidelines here: https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#commit-guidelines and add a motivation to each commit. This is also missing tests and a changelog entry since the second commit even introduces a new command.

matts1:
I know, that's why it's a draft commit. Just want to make sure the concept is approved before I add finishing touches.

yuja:
cc @arxanas, @ilyagr

The idea sounds generally good to me. We might also want a similar mechanism for the zip file problem. Since the `.jj` directory may be untrusted, it might be better to store trusted config contents or hashes in e.g. `~/.local` directory.

https://github.com/jj-vcs/jj/issues/1595

I have no idea about TUI. I personally would want to just see the diff and accept/reject.

matts1:
> The idea sounds generally good to me. We might also want a similar mechanism for the zip file problem.

I'm not familiar with "the zip file problem". Could you link to the issue?

> Since the `.jj` directory may be untrusted, it might be better to store trusted config contents or hashes in e.g. `~/.local` directory.

I'm not opposed to that. I chose this method because it was the same directory that the repo config was stored. I would request, however, that we submit this first and worry about that later, since this is explicitly not making the security any worse than it already is.

> I have no idea about TUI. I personally would want to just see the diff and accept/reject.

I do agree, as a user, that that's what *I* would want, but that's also relatively easy to achieve with the TUI. I think, however, that from a security perspective, a binary accept / reject choice is not great, because it might feel like you have to make a choice between new features and security. With a binary choice, if there's one particular line that you are concerned about and so you reject it, that means you have to reject *every* future change to the repo config.

martinvonz:
> I'm not familiar with "the zip file problem". Could you link to the issue?

I don't think we have a link, but it's a familiar problem from other VCSs (familiar to VCS developers). The problem is what to do about repo-level configs in `.jj`/`.git`/`.hg`. For example, an attacker can send you a zip file of a repo where the repo-level config has `rm -rf /` as diff tool. They might ask you to help them troubleshoot some problem with their repo, for example, and you might not expect that simply running `jj/git/hg diff` in the repo would be harmful.

matts1:
Makes sense. So I assume the solution would be to store it out of the repo, and require users to run `jj config edit --repo` instead of something like what I currently do where I type `vim .jj/repo/config.toml`?

martinvonz:
I think Yuya's proposal was to record trusted config contents somewhere in `~/.local`. When you unpack a zip file and run a command in that repo, you'd get a warning saying that it has untrusted configs. Then you run some command to say which of the configs you approve.

I'm not sure if it's safe to trust configs at the key-value level or only at coarser granularity. Could there be some case where you say that you trust `key1="value1"` and `key2="value2"` but somehow the combination is still harmful? I hope not, and it feels like it should be safe.

matts1:
> I'm not sure if it's safe to trust configs at the key-value level or only at coarser granularity. Could there be some case where you say that you trust `key1="value1"` and `key2="value2"` but somehow the combination is still harmful? I hope not, and it feels like it should be safe.

AFAICT, the attack surfaces are, in order of danger:
* Mess with the repo
* Send private code
* Remote code execution

This is my analysis of the danger:
* Revsets: Minor  nuisances
  * Can mess with the repo
  * Can send code, but generally only to existing remotes
  * Can't escape the repo
* New remotes:
  * Should be obvious to spot, so easy to mitigate
  * RCE possible via pulling code
  * Possible to send private code by accident
* Arbitrary command execution. Only two sources of this are (to the best of my knowledge):
  * jj fix tools
  * Command aliases to `jj util exec`

I think that it's possible to make one harmful thing and one alias to it, but I don't think it's possible to make two things that are independently safe but together harmful (it's technically possible with two arbitrary commands that execute files in the repo, but if they're doing that then they can already achieve this).

.config/jj/config.toml
PhilipMetzger, line 1: nit: unrelated to this PR so split it into another PR for discussion

docs/contributing.md
PhilipMetzger, line 1: Nit: also belongs into the other PR for discussion

cli/src/commands/config/review_managed.rs
```diff
@@ -0,0 +1,121 @@
...
+use crate::merge_tools::make_diff_sections;
+use crate::ui::Ui;
+
+#[derive(clap::Args, Clone, Debug)]
+pub struct ConfigReviewManagedArgs {
+    /// Trust the repository config. Copy the file over without reviewing
+    /// changes.
+    #[arg(long)]
+    trust: bool,
+}
```
PhilipMetzger, line 41: Nit: this is missing a user-friendly doc comment which explains the commands use-case. I know what you intend to do with it but others may not.

cli/src/commands/config/review_managed.rs
PhilipMetzger, line 1: In my opinion this shouldn't belong in the VCS.

CHANGELOG.md
PhilipMetzger, line 1: nit: This commit is also missing its motivation, since it's use-case partially originated  in the Chrom{e, ium] project which you're involved in.

PhilipMetzger (commented):
Here's a bunch of actual review comments.

cli/src/commands/config/review_managed.rs
matts1: I quite strongly disagree with this. If you leave it up to the user to review the config and copy it, it's far less secure due to psychology:
* The user can run `ln -s` to create a symlink instead of copying the config, thus never even reviewing changes
* The user is far more likely to review the diff if you actually force a diff open for them than if you tell them "please review and copy if you trust it" - most users will not be bothered to open a diff

What makes you say this shouldn't be in the VCS?

PhilipMetzger: > I quite strongly disagree with this. If you leave it up to the user to review the config and copy it, it's far less secure due to psychology:
> 
> * The user can run `ln -s` to create a symlink instead of copying the config, thus never even reviewing changes
> * The user is far more likely to review the diff if you actually force a diff open for them than if you tell them "please review and copy if you trust it" - most users will not be bothered to open a diff

I don't disagree that the config should be reviewed by the users it just inherently is to late at the point when you run an updating command (`jj git sync`, `jj git fetch`) since you trusted the repo's owner the moment you put it on your disk. It'll also be the case that people will just "skip" (and even be encouraged to by repo owners) this process since they blindly trust the organization behind it. 

All in all this is to me a technical solution to a social problem.

> What makes you say this shouldn't be in the VCS?

Because building a meta-configuration system doesn't belong into a VCS's responsibility and you essentially trust the organization/person or repo when you download or unzip it. Ideally this is done via a configuration management system outside the VCS which is individually reviewed and approved by the respective organization.

CHANGELOG.md
PhilipMetzger: while it's definitely better now, it still fails to mention that this change adds a new command which is a approving mechanism for the per-repo configuration.

cli/src/commands/config/review_managed.rs
emilazy: It's not true that download some files is consent to execute arbitrary code on your machine and to the extent this is true it's a deficiency of systems that is being increasingly moved away from. Editors ask you whether you want to enable risky LSPs in new workspaces or accept configuration from them, tools like `direnv` won't load new files without a trust prompt, unsandboxed build scripts in things like Cargo are acknowledged as unfortunate design mistakes (but builds are easier to sandbox than VCS config).

I haven't reviewed the design of this PR and don't have strong opinions on whether we should support dangerous config settings in repos at all - I just want to speak against the risk of falling into a security nihilism where the only solution is for everyone to exercise unrealistic caution and tools shouldn't try to be safe by default. If we do let repos set dangerous settings then we should certainly have a trust model for it. Git just had a CVE very recently that allowed escalating from repository contents to writes/code execution as a user. If we deliberately expose that kind of functionality it should be behind user consent.

cli/tests/test_repo_managed_config.rs
```diff
@@ -0,0 +1,75 @@
...
+    test_env.add_config(r#"ui.pager = "user pager""#);
+
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    user pager
+    [EOF]
+    "###);
+
+    work_dir.write_file(
+        ".config/jj/config.toml",
```
ilyagr, line 31: If we get close to merging this, this path should be documented in `config.md`. This probably deserves a whole section there, but it also makes sense to postpone writing it until we agree.

cli/tests/test_repo_managed_config.rs
```diff
@@ -0,0 +1,75 @@
...
+        r#"ui.pager = "repo-managed pager v2""#,
+    );
+    let output = work_dir.run_jj(["config", "get", "ui.pager"]);
+    insta::assert_snapshot!(output, @r###"
+    repo pager
+    [EOF]
+    ------- stderr -------
+    Warning: Your repo-managed config is out of date
```
ilyagr, line 102: If the user decides that they don't like some of the repo-managed settings, will they keep getting this warning?

I think they shouldn't. Perhaps there should be a `jj config review-managed --stop-warning` command and a `jj config review-managed --diff` command (to see the differences after you suppressed the warning).

This does add some complexity, since now we have to figure out when it makes sense to show the warning again after that.

One possible trick (that wouldn't solve the problem fully) would be to store the hash of the repo-managed config somewhere (e.g. in the user config), and stop warning while the repo-managed config matches the hash. This also *could* allow remembering decisions of what part of config to accept as you jump from commit to commit (this may be a bad idea).

This will also give a way for a user who can't run a TUI to manually copy the config they want over.

(This is also polish, it might make sense to wait until we're more or less confident in the approach. I'm also basing this mainly on a quick glance at the test, so I may have misinterpreted how this works)

cli/tests/test_repo_managed_config.rs
ilyagr, line 1: (Very much optional): We could think of a few configs to put in the repo-managed config for this repo. This would help with both experimenting with this, and also with knowing more about the use-cases.

ilyagr (commented):
I think this is probably OK. My worries are:

- Is this too inflexible to be useful? For example, as the user jumps from commit to commit, the repo-managed config may change, and I don't see a way to decide automatically whether they want their config to change accordingly (even if we trusted the config; see my hash idea in the other comment for how we could).

- How easy or hard will it be for the users to miss something important in the diff interface?

Figuring this out might require real-world testing. I'd probably merge some version of this, but disable it by default and call this feature "experimental" for quite a bit after. This is to give us time to polish it, but also to see whether it's worth the complexity/security tradeoff in practice.

Most of my other comments are various ideas about how to polish this further, which only make sense once we have a consensus that we're going forward with this (It's not clear to me how close we are at the moment).

cli/tests/test_repo_managed_config.rs
ilyagr: Complete docs would be less critical if we disable the feature by default as I suggest in https://github.com/jj-vcs/jj/pull/6990#pullrequestreview-3035022120, though we would still want to mention *something* about this in config.md probably.

cli/tests/test_repo_managed_config.rs
PhilipMetzger: That was done initially but I requested a separate review for that since the discussion should be around the feature not its direct use-case here. https://github.com/jj-vcs/jj/pull/6990#discussion_r2213519929

ilyagr: I managed to recover that config suggestion using <https://reviewstack.dev/jj-vcs/jj/pull/6990>:

```toml
# Contains the common configuration for working on this repo.

[revset-aliases]
'trunk()' = 'main@origin'

[fix.tools.rustfmt]
command = ["rustfmt", "+nightly"]
patterns = ["glob:**/*.rs"]
```

Aside: I think I previously could do similar things on GitHub, but lately I either couldn't remember how, or maybe the functionality is gone.

matts1: I agree with @PhilipMetzger that we should probably do them in seperate PRs

cli/tests/test_repo_managed_config.rs
matts1: > If the user decides that they don't like some of the repo-managed settings, will they keep getting this warning?

No, they won't. The current mechanism is:
1) The repo updates the config
1) It warns you that the config is out of date because `last_reviewed.toml != <repo>/.config/jj/config.toml`
1) You run `repo config review-managed`
1) You don't like one line, so you approve every line of it except for that one
1) jj writes every line except that one to the actual config file in use, and the whole file including lines you didn't approve to `last_reviewed.toml`
1) You run commands again. It won't complain that it's out of date because `last_reviewed.toml == <repo>/.config/jj/config.toml`

matts1:
> * Is this too inflexible to be useful? For example, as the user jumps from commit to commit, the repo-managed config may change, and I don't see a way to decide automatically whether they want their config to change accordingly (even if we trusted the config; see my hash idea in the other comment for how we could).

My thoughts were, when designing this:
* The simplest approach is to simply not have the config to change at all
* The config will need to be approved, so it should not change accordingly IMO
* The only other option I can see is to have it, instead of reading the config from disk, automatically track `trunk()`.
  * This presents potential problems in the sense that a change could be breaking (eg. jj fix referencing a file `format.py` that appeared in the same commit as the config change).
  * We could do something akin to dynamic approval where we have a map from hash of config file on disk to approved config. This seems way too complex for the initial solution though.
* I figured that it was unlikely someone would jump from commit to commit across different configs, and even if they did they'd only get a warning
  * Maybe this is wrong. It might just be because my workflow involves every commit being rebased on top of `trunk()` whenever I sync. Maybe people who use long-term feature branches that eventually get merged into main might not agree that it's unlikely?

> * How easy or hard will it be for the users to miss something important in the diff interface?

I think that if the user cares about security, it's pretty hard to miss. It is also relatively easy to just blindly approve, though.

> Figuring this out might require real-world testing. I'd probably merge some version of this, but disable it by default and call this feature "experimental" for quite a bit after. This is to give us time to polish it, but also to see whether it's worth the complexity/security tradeoff in practice.

I wouldn't be opposed to putting this behind a flag. Since this is inherently pre config-load, do you know if this would be able to be in our config schema, or would it have to be something like an environment variable flag?

CHANGELOG.md
matts1: I've now mentioned that it's a new command.

cli/tests/test_repo_managed_config.rs
matts1: I've now locked it behind the feature flag `repo-managed-config.enabled` and documented it in `config.md`

matts1:
I've added a config setting to toggle this feature, as requested by @ilyagr.

Does anyone have opinions on what the config file should be? My candidates were:
* `.jjconfig.toml`
* `jj/config.toml`
* `.config/jj.toml`
* `.config/jj/config.toml`

I ended up settling on the last one for a few reasons:
* I saw someone opining somewhere (can't find it now though) that workspace root tends to get cluttered already, so it was probably better to put it under a subdirectory.
  * My two cents is that for ownership purposes, it's nice to be a directory rather than a file.
* `.jj` is taken by the repo itself, and the file should be hidden, so `jj/config.toml` is ruled out.
* `.config` seems like a good place to put it
* `.config/jj.toml` doesn't give much flexibility if we need to add more jj-related files later (eg. `.jjignore`, `.jjattributes`)
* Adding the directory `.config/jj` allows users to store both the jj config and any commands they reference in the same directory
  * Eg. `alias.sync = ["jj", "util", "exec", ".config/jj/sync.py"]`
  * Then you can have `.config/jj/sync.py` in the same directory, which makes things much easier to review from a security standpoint.
  * This also allows `.config/jj/OWNERS` to exist

That being said, I'd welcome other opinions.

CHANGELOG.md
PhilipMetzger: the thread was about the commit message, which hasn't changed.

cli/src/commands/config/review_managed.rs
PhilipMetzger: To be clear I don't really have an issue with supporting per repo configs, I just see unfortunate social interactions with people adding a `jj util exec` alias which blasts everything away. And since they already executed `jj` and established such a trust with their upstream project, then they'll come back to blame us. 

It seems both of you misunderstood it that it was about this (which is entirely my opinion on the whole thing, so not blocking at all). Anyway I see the same social issues upcoming as this lobste.rs thread talks about: https://lobste.rs/s/nm5lho/stop_putting_nix_setup_your_checked_envrc. (Also the discussion in https://github.com/jj-vcs/jj/pull/4759#pullrequestreview-2416330036 somewhat relevant).

cli/src/commands/config/review_managed.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::command_error::user_error;
+use crate::command_error::CommandError;
+use crate::config::maybe_read_to_string;
+use crate::merge_tools::make_diff_sections;
+use crate::ui::Ui;
+
+/// Reviews and updates configuration stored in version control.
+/// You should never need to run this command unless jj tells you to.
+#[derive(clap::Args, Clone, Debug)]
+pub struct ConfigReviewManagedArgs {
```
PhilipMetzger, line 35: nit: this still can be expanded to explain _when_ you're prompted and what you're able to do with it.

PhilipMetzger (commented):
@matts1 How do you imagine this working with multiple "correct" upstreams, like Eidolon's which inspired #3670? It isn't clear to me what should happen then, since there are multiple options which available beside erroring out.

matts1:
> @matts1 How do you imagine this working with multiple "correct" upstreams, like Eidolon's which inspired #3670? It isn't clear to me what should happen then, since there are multiple options which available beside erroring out.

I think in general, my answer is:
* They *probably* have the same config, in which case there's no problem
* If they don't have the same config, you probably spend a lot more time on one branch than the other. Just use the config from that branch.
* Otherwise, they can disable this feature if they like (and potentially manually specify the config file on the command-line)

There is a potential solution to this, but it's definitely not within the scope of the MVP. We can create a mapping from `hash of checked-in config -> approved config`. Then when you switch between revisions, it would just use the previously approved config instead of warning you that it's out of date.

joyously:
> with people adding a `jj util exec` alias which blasts everything away.

What if repo configs don't allow aliases?

matts1:
> > with people adding a `jj util exec` alias which blasts everything away.
> 
> What if repo configs don't allow aliases?

That wouldn't really solve the problem. You could achieve the same just as easily with a malicious `jj fix` command, which definitely *should* be in the repo config, since each repo has their own rules for formatting.

Also, I think that would somewhat defeat the point of the repo config. In chromium, for example:
* The process to sync is:
  1) Run a git fetch
  1) `jj rebase` onto `trunk()`
  1) Run `gclient sync`, which updates the git submodules
* The process to upload (with git) is to run `git cl upload`, which:
  1) Runs a formatter
  1) Runs pre-upload checks
  1) Uploads to gerrit

While the former might be (somewhat) unique to chromium, I think that an alias `jj pre-upload` would be pretty reasonable and standard, and I've been running python scripts in lieu of hooks to run the equivalent of `jj fix && jj pre-upload && jj git push`.

So TLDR, I think they have too much value to disable. I think one approach that could be taken though is some stricter security measures. For example:
* We could execute `jj fix` in a read-only sandbox so it can't write (or delete) files
* We could, by default, make `jj util exec` run in a read-only sandbox using overlayfs or similar, with flags such as:
  * `--no-sandbox`: completely disable the sandbox
  * `--allow-repo-writes`: allows writes to all files in the repo excluding `.jj`
  * `--allow-global-writes`: allows writes to any file in the filesystem
* We could sandbox file reads to, by default, only allow:
  * entries in `PATH`
  * system directories
  * files in repo
  * `$HOME/.*/**` (eg. `~/.config/...`, `~/.bazelrc`

In general, sandboxing reads is a lot more painful, and I'm not sure it's a good idea, but sandboxing writes does sound like a very good idea.

PhilipMetzger:
> > @matts1 How do you imagine this working with multiple "correct" upstreams, like Eidolon's which inspired #3670? It isn't clear to me what should happen then, since there are multiple options which available beside erroring out.
> 
> I think in general, my answer is:
> 
> * They _probably_ have the same config, in which case there's no problem
> * If they don't have the same config, you probably spend a lot more time on one branch than the other. Just use the config from that branch.
> * Otherwise, they can disable this feature if they like (and potentially manually specify the config file on the command-line)
> 
> There is a potential solution to this, but it's definitely not within the scope of the MVP.

I pretty much agree with your assessment and that it's out of the scope of the MVP. So I'm fine with dealing with the aforementioned issue when it comes up. 

> > with people adding a `jj util exec` alias which blasts everything away.
> 
> What if repo configs don't allow aliases?

I don't think that's possible since it nerfs the usability of this feature by a large amount (making it unusable for most of the use-cases Matt wants to use it for).  

> So TLDR, I think they have too much value to disable. I think one approach that could be taken though is some stricter security measures. For example:
> 
> * We could execute `jj fix` in a read-only sandbox so it can't write (or delete) files
> * We could, by default, make `jj util exec` run in a read-only sandbox using overlayfs or similar, with flags such as:
>   
>   * `--no-sandbox`: completely disable the sandbox
>   * `--allow-repo-writes`: allows writes to all files in the repo excluding `.jj`
>   * `--allow-global-writes`: allows writes to any file in the filesystem
> * We could sandbox file reads to, by default, only allow:
>   
>   * entries in `PATH`
>   * system directories
>   * files in repo
>   * `$HOME/.*/**` (eg. `~/.config/...`, `~/.bazelrc`

While I like the idea of sandboxing it to me isn't a cross-platform solution. I'd still prefer to have a scripting language at some point in which we control the entire environment. 

At this point this LGTM.

martinvonz:
> Since the `.jj` directory may be untrusted, it might be better to store trusted config contents or hashes in e.g. `~/.local` directory.

This seems like the right direction to me. @matts1: Are you willing to work on the feature in that form?

martinvonz:
Btw, the PR title is confusing to me. We already support repo-level configs. What's the "user+repo level" about? Would one of the following work?

> Allow repo-managed configs

> Allow in-tree configs

jennings:
I like this feature in principle because it's annoying to copy `[fix.tool]` configs everywhere, and putting them in my user config with `enabled = false` isn't as convenient as I had hoped when I built it.

These are my thoughts on the design of this feature. I've read the commit message and the comments here, but not reviewed the implementation.

### What should be the precedence of the various configs?

I think the precedence should be:

- `--config` flags
- `.jj/repo/config.toml`
- Activated repo-managed config
- User-level config (`~/.config/jj/*`, etc.)

I haven't thought about how `$JJ_CONFIG` fits in here.

### How are repo-managed configs transmitted between repositories?

Stored in the repo under the well-known path `.config/jj/config.toml`. I think [Matt's argument for this is a good one](https://github.com/jj-vcs/jj/pull/6990#issuecomment-3095097907).

### What happens when checking out a revision with different repo-level config?

Once config has been "approved", I think that should be remembered between checking out different revisions. For example, suppose these commits change the repo-managed config:

```
D  Repo config v2
|
C  Repo config v1
|
B  Repo config v1
|
A  No repo-managed config
```

Once the user does `jj new B; jj config approve` (or whatever the procedure is), then the config in version B is used even after editing different versions:

```shell
$ jj new B
# No repo-mangaged config is active
# Approval required message

$ jj config approve # or whatever
# config v1 is active

$ jj new A
# No repo-managed config is active

$ jj new B
# config v1 is active

$ jj new C
# config v1 is active

$ jj new D
# No repo-managed config is active
# Approval required message

$ jj config approve
# config v2 is active
```

### Where is approval information stored?

- The user should have to approve in some way beyond just running `jj new <some_version_with_config>`
- The config probably shouldn't just be copied into `.jj/repo/config.toml` because that could override some setting the user set explicitly.
- Approvals can't be stored in in the `.jj` directory because of the zip file problem.
- So, something like `~/.local/jj/*` seems appropriate.

### What exactly is stored when config is approved?

- Rather than copying the whole repo-managed config file here, we could probably just store the SHA-256 of every approved version?
  - And probably store the _current_ active version here, copied from `$REPO/.config/jj/config.toml`, so we can just read and trust it here rather than validate the in-repo version every time.
- Do we want some mapping between a repo-managed config and the exact repo it was approved for?
  - For example, if the same `.config/jj/config.toml` exists in `~/A` and `~/B`, we probably don't want an approval in repo A to approve it for repo B.
  - What if the user does `mv ~/A ~/C`? Is the approval forgotten?

### Miscellaneous

- Different platforms probably need different tool config (_sigh_, Windows). I would want this to support `--scope` blocks so I could, eventually, have `--when.platform = ["windows"]` to create per-platform config.

matts1:
> > Since the `.jj` directory may be untrusted, it might be better to store trusted config contents or hashes in e.g. `~/.local` directory.
> 
> This seems like the right direction to me. @matts1: Are you willing to work on the feature in that form?

That feels like it's better as a follow-up PR, especially since I've disabled it with a feature flag by default. I think that the local directory is probably better in the long term, but in the short term I think there's nontrivial decisions that need to be made.
* A trusted config hash for one repo *should not* be trusted for another, since paths are relative to the repo
  * Doing so would introduce an attack surface where you copy a config file with a relative path, and add your own malicious file at that path. For example, copy chromium's [config.toml](https://chromium-review.googlesource.com/c/chromium/src/+/6711999) and have your own malicious `tools/jj/sync.py`
* Is it based on the local path on disk?
  * What happens if you delete a repo and put another repo at the same location on disk? That new repo inherits the trust of the old repo
  * Maybe this is fine?
* Is it based on the remotes? That could have the nice property that if I have multiple checkouts of the same repo I only need to approve them once, but how do you check that two repos are the same? Probably based on remotes or something?
  * What happens with multiple remotes?

> Btw, the PR title is confusing to me. We already support repo-level configs. What's the "user+repo level" about?

Sorry, my original PR renamed repo configs to user-repo configs, and called this "repo configs", but I didn't like that so I changed it. Updated the title to make it more clear.



> ### What should be the precedence of the various configs?

Currently repo-managed configs are between the global user config and the repo's `config.toml`

> ### What happens when checking out a revision with different repo-level config?
> Once config has been "approved", I think that should be remembered between checking out different revisions. For example, suppose these commits change the repo-managed config:

This is the current behaviour, with the hope that in future versions we'll store a map from checked in config -> approved config instead of just remembering the most recently approved version.

> ### Where is approval information stored?
> * Approvals can't be stored in in the `.jj` directory because of the zip file problem.

As stated above, I think that for now we should store it in the `.jj` directory, and move it to `.local` once we answer some of the more difficult questions.

> ### What exactly is stored when config is approved?
> * Rather than copying the whole repo-managed config file here, we could probably just store the SHA-256 of every approved version?

I think we would want to store:
* filename is the hash of the content stored in version control
* contents is the actual content that was approved (since you can approve or reject it line by line)

>   * What if the user does `mv ~/A ~/C`? Is the approval forgotten?

I don't think we should worry about this use case. It's not frequent enough to be a concern.

> 
> ### Miscellaneous
> * Different platforms probably need different tool config (_sigh_, Windows). I would want this to support `--scope` blocks so I could, eventually, have `--when.platform = ["windows"]` to create per-platform config.

Yeah, I was looking at this for chromium's fix tools. We have `clang-format` and `clang-format.exe` respectively. I ended up calling `clang-format.py` which chose the correct `clang-format` based on the platform.

martinvonz:
> A trusted config hash for one repo _should not_ be trusted for another, since paths are relative to the repo
> * Doing so would introduce an attack surface where you copy a config file with a relative path, and add your own malicious file at that path. For example, copy chromium's [config.toml](https://chromium-review.googlesource.com/c/chromium/src/+/6711999) and have your own malicious `tools/jj/sync.py`

That's a very good point! I don't think we had considered that the config would contain repo-relative paths. I guess we'll have to think more about this.

Sorry, that's all I have time to reply about right now.

emilazy:
Just wanted to share some thoughts; I talked about some of this on the Discord before but wanted to get it posted here and expanded upon.

> That feels like it's better as a follow-up PR, especially since I've disabled it with a feature flag by default. I think that the local directory is probably better in the long term, but in the short term I think there's nontrivial decisions that need to be made.

I agree that since `.jj/repo/config.toml` is already trusted, it’s fine to store trust status in `.jj` for now – the threat model does not change.

I do also think that we should make `.jj` not trusted in this way; that unpacking a repository in a `.zip` file and running a Jujutsu command inside it should not be able to achieve arbitrary code execution. So I want to explicitly disclaim that none of this should be a blocker for the current PR, but since I think the current behaviour (before your PR) is a security issue and it’s very much worth thinking about as part of the design for this stuff, I hope you don’t mind me talking about the points raised a little more :)

> I think we would want to store:
> 
> * filename is the hash of the content stored in version control
> * contents is the actual content that was approved (since you can approve or reject it line by line)

> * A trusted config hash for one repo _should not_ be trusted for another, since paths are relative to the repo
>   
>   * Doing so would introduce an attack surface where you copy a config file with a relative path, and add your own malicious file at that path. For example, copy chromium's [config.toml](https://chromium-review.googlesource.com/c/chromium/src/+/6711999) and have your own malicious `tools/jj/sync.py`

(The following treats configuration files in the working copy as monolithic things to accept or reject in their entirety, since I’m not sure the diff‐selecting interface makes sense if we move from just copying stuff into `.jj`. But it can be adapted as appropriate for partial approval.)

I agree that it makes sense to key trust on the repository, because configuration that you are okay with for one repository may not be okay for another. (For instance, perhaps the `origin` remote is a trusted internal repository in one, but in another, `origin` is the FOSS upstream and `fork` is the internal repository. You are happy agreeing to push things to `origin` in the former, but the latter would result in leaks. Or consider private commit configuration, and so on.)

I am not sure to what extent it makes sense to include the other contents of the working copy in the threat model, however:

1. If it’s something like `rustfmt.toml` being passed as a configuration file, then the trust we’re expressing is “it’s okay to run `rustfmt`”, and it’s `rustfmt`’s responsibility to ensure that a malicious `rustfmt.toml` can’t do anything evil. If you’re worried about that and don’t trust `rustfmt`, then you should not approve the configuration that invokes it. In other words, we’re the first line of defence against executing things that shouldn’t be executed at all, but beyond that we have to pass the responsibility for safety on to the things we invoke.

2. The reason to have approval for in‐working‐copy configuration is that we may switch to malicious commits. Therefore, it makes no sense to want to approve each configuration change individually in case it comes from switching to a malicious commit for review, but not worry about the fact that the malicious commit could simply put the malicious code in `tools/jj/sync.py`. This kind of threat applies within a single repository, therefore, not just when transferring configuration across repositories.

   In other words: if you’re trying to mitigate “developers getting owned” risks, then you don’t want to directly invoke arbitrary code as part of the workflow, and if the user approves a configuration that does that, they’re accepting the risk that involves. In an ideal world, we could prompt the user for re‐approval when `tools/jj/sync.py` changes. But we do not have a reliable, practical method to pin hashes of every single file or bit of state that could affect an arbitrary command inside the configuration. <del>(Unless we make people use Nix.)</del>

   One way to mitigate this is to have an explicit pinned revision for tooling, i.e. doing something like `python3 <(jj file show -r abcdef tools/jj/sync.py)` (or `-r main@upstream`) in the alias configuration. Then the user will be prompted if the tooling will potentially‐untrusted source. Of course, the tooling must then be careful not to trust the working copy. Nixpkgs does something like this, where an explicit revision of the repository is pinned inside Nixpkgs itself to obtain CI tooling from, though mostly for unrelated reasons.

Therefore, I believe that our principle should be that changes to the working copy other than to the configuration file are out of scope for the threat model, and that users approving a configuration accept the risk of trusting the invoked tooling to handle it safely, or trusting arbitrary future states of the working copy itself if the approved configuration implicitly hands it privileges. (The latter does not necessarily make the feature useless, in the presence of a trusted remote enforcing ACLs or similar.)

> * Is it based on the local path on disk?
>   
>   * What happens if you delete a repo and put another repo at the same location on disk? That new repo inherits the trust of the old repo
>   * Maybe this is fine?
> * Is it based on the remotes? That could have the nice property that if I have multiple checkouts of the same repo I only need to approve them once, but how do you check that two repos are the same? Probably based on remotes or something?
>   
>   * What happens with multiple remotes?

FWIW, the [direnv](https://direnv.net/) tool has been in the business of executing arbitrary code to set up development environments when you `cd` into a directory for a long time now. The model it uses is that trust is conditioned on the combination of the path to the `.envrc` and the hash of its contents. The same `.envrc` at two different paths isn’t necessarily trusted, because it can be (and often is) as simple as “run this arbitrary code from another file the same directory to get environment variables”, and an `.envrc` at a trusted path that changes needs re‐approval. Explicit denials (to stop it notifying about an untrusted `.envrc`) block the path itself regardless of contents.

I think this is a broadly reasonable solution, especially if approvals for paths that no longer exist are automatically pruned. However, for Jujutsu, I wonder if we can’t do something that is both simpler and more effective, given our position as the VCS: why don’t we key approvals on, say, the hash of the `@` operation? (Or maybe the commit hash, but I feel less confident about that, especially in the presence of remote configuration and Git subtrees and so on.)

That should close over the hashes of basically everything we could care about that contributes to the state of the configuration. We might want to throw in the workspace path too just in case you do something wonky with `../`, but in general I feel good about reusing an existing mechanism here rather than trying to manually throw a bunch of stuff into the hash.

I think it’s best to err on the side of requiring re‐approval, because it’s only a mild annoyance unless you are cloning entirely new copies of a repository constantly, but being too lax is a CVE.

> As stated above, I think that for now we should store it in the `.jj` directory, and move it to `.local` once we answer some of the more difficult questions.

FWIW, my proposal for `.jj/repo/config.toml` when we move approvals to `$XDG_DATA_HOME` (or `$XDG_STATE_HOME`?) would be to extend the same trust mechanism to the configuration in there, and have `jj git init`, `jj git clone`, and `jj config set --repo` automatically update the approved revision. That would cover the “unzipping a malicious repository” risk, at the cost of having to approve configuration changes that are made from outside of Jujutsu, while still allowing repositories to be reasonably portable.

(We could also of course have a list of configuration options that are guaranteed to be safe to read from untrusted sources to reduce the prompting here, but it’s risky business if you haven’t architected your whole program for paranoia from the ground up; seemingly benign values can end up load‐bearing, especially in the presence of user configuration that looks at things like remotes.)

joyously:
Just an opinion: it's oh so much easier to avoid the risk when the tool never loads config from the repo it is managing. And that applies to all VCS, not just `jj`.
If you implement this, wouldn't it affect commands that extract historical commits with a different config? And some of those commands you want to run unattended?
If you don't implement this, can't a team document in their contributing.md how they want code formatted or whatever? Does the config of the VCS really belong in the repo? (I think it doesn't belong there.)

Edit: I forgot to mention, I was reading an [article](https://thiswillhold.substack.com/p/ex-cia-whistleblower-the-nsa-audited/#:~:text=unlocked) alleging the US 2024 election was manipulated, which tied back to a small approved change concerning the config file...that its hash was in the dynamic list instead of the static list to check for. This seemed related.

matts1:
IMO, the config of a VCS doesn't need to load config from the repo it's managing. However, by choosing to implement several features, I believe that `jj` has chosen a route where it should.

I think that if a repo wants to tell a user how to format their code, that's fine. If a VCS wants to do formatting of code, that's also fine. But IMO you can't do both at the same time without allowing a repo to tell jj how to format their code. The same goes, in my opinion, for `jj util exec` aliases, which constitute workflows.

I strongly believe that adding this feature increases security, not decreases it. While you're right that this feature introduces potential security concerns, I think that by doing this you are ensuring that everyone does things in a relatively secure way. I think that if you do not add this feature, you will get things such as:
* Repos that instruct users to run `ln -s <repo config> .jj/repo/config.toml` (or `cp`), thus bypassing any review (this is what we currently recommend for chromium)
* Repos that instruct users to run a script to set up your jj config

To use a metaphor, a door with a lock on it might not be perfectly secure, but if you don't make a door, people are gonna smash a hole through the wall, and that's definitely not secure.

joyously:
> I think that by doing this you are ensuring that everyone does things in a relatively secure way.

You didn't address what happens with multiple versions of the config file that get surfaced when you switch to that version, or when there are both a user and a repo config in the repo, or the CI hitting this approval thingy (command that isn't supposed to be interactive suddenly is).
My own opinion is that a VCS shouldn't format code, only version it. It doesn't really need to have `util exec` either, but aliases are really handy (and are in several VCSs).

matts1:
I've addressed these in earlier comments, but I'll reiterate:

> What happens with multiple versions of the config file that get surfaced when you switch to that version

When you switch to a version, you simply get a warning that you're using a different version of the config file. You can keep using your current version (as mentioned in previous commits, this isn't ideal, but is a good MVP, and I mentioned in an [earlier comment](https://github.com/jj-vcs/jj/pull/6990#issuecomment-3094941766) a strategy we could use to improve on this by using hashes).

> or when there are both a user and a repo config in the repo

The order of application is (with the last overriding):
1) Builtins
2) User config (`~/.config/jj/config.toml`)
3) Managed repo config (this PR)
4) User's repo config (`<repo>/.repo/config.toml`)
5) Command-line

> or the CI hitting this approval thingy (command that isn't supposed to be interactive suddenly is).

Nothing becomes interactive here. It simply outputs a warning that says "The config is out of date. We suggest updating the config by running `jj config review-managed`".

> My own opinion is that a VCS shouldn't format code, only version it. It doesn't really need to have `util exec` either, but aliases are really handy (and are in several VCSs).

That's a fair opinion, and I think that if jj had chosen to go down that path, you'd probably be right that we don't need this, but I think that given the decisions that jj has made, this path is the right one.

jennings:
> > **What happens when checking out a revision with different repo-level config?**
> >
> > Once config has been "approved", I think that should be remembered between checking out different revisions. For example, suppose these commits change the repo-managed config:
>
> This is the current behaviour, with the hope that in future versions we'll store a map from checked in config -> approved config instead of just remembering the most recently approved version.

That's actually the _opposite_ of what I thought, but I now see how what I said was ambiguous :)

But as I think about it, I think only remembering "the most recently approved version" is probably a perfectly fine way to start. Repo-managed config shouldn't be changing often, so it isn't much of a problem in practice.

*(I previously had here a suggestion about using the diff editor to review config changes. Then I realized there was prior art for that: The implementation of the PR I'm commenting on. Whoops.)*

matts1:
I was hoping we could get this reviewed and submitted soon.

I think there are some open questions, but they're a non-issue for this particular PR IMO:
* How do you manage switching between configs
  * While it's somewhat important, it doesn't affect the MVP
* Can we avoid the zip file problem
  * It's definitely possible, but there's not a single solution. There's many different options we could take. I don't want to block this CL on a discussion on that, since this CL does not degrade security (it uses an insecure method, but it's the same method we already use for existing config files).

joyously:
Has anyone looked up the discussions on other VCS using config from the repo it is managing, or is the only one?

cli/src/commands/config/review_managed.rs
```diff
@@ -0,0 +1,127 @@
...
+
+/// Reviews and updates configuration stored in version control.
+/// You should never need to run this command unless jj tells you to.
+/// This command needs to be run when the config checked in to the repo is
+/// changed, and allows you to approve or reject said changes on a line-by-line
+/// basis.
+#[derive(clap::Args, Clone, Debug)]
+pub struct ConfigReviewManagedArgs {
+    /// Trust the repository's config and skip review of it.
+    /// Use this when you absolutely trust the repo config (eg. you're the only
```
martinvonz, line 37: nit: "eg." -> "e.g."

cli/src/config.rs
```diff
@@ -326,11 +329,31 @@ impl UnresolvedConfigEnv {
...
     repo_path: Option<PathBuf>,
+    workspace_root: Option<PathBuf>,
     user_config_paths: Vec<ConfigPath>,
     repo_config_path: Option<ConfigPath>,
     command: Option<String>,
 }
 
+pub struct RepoManagedConfigPaths {
+    /// The path to the file managed by the repo.
+    pub managed: PathBuf,
```
martinvonz: I feel like this should be a `RepoPathBuf` and that the contents should be read from the working-copy commit. That's generally the same as reading it from the working copy; the main difference is that it works even if the config file is sparsed away. I'm guess the feature should ideally work even if the user has configured the sparse patterns to match only `lib/`, for example. Even if you agree, I don't think you need to implement it right now. Feel free to leave a TODO IMO.

cli/src/config.rs
```diff
@@ -441,9 +465,22 @@ impl ConfigEnv {
...
+        self.repo_config_path = Some(ConfigPath::new(repo_path.join("config.toml")));
+    }
+
+    /// Returns all paths associated with the repo-managed configuration.
+    pub fn repo_managed_config_paths(&self) -> Option<RepoManagedConfigPaths> {
+        match (&self.repo_path, &self.workspace_root) {
+            (Some(repo_path), Some(workspace_root)) => Some(RepoManagedConfigPaths {
+                config: repo_path.join("generated_repo_config.toml"),
+                last_reviewed: repo_path.join("last_reviewed_repo_config.toml"),
+                managed: workspace_root.join(".config/jj/config.toml"),
```
martinvonz: nit: match order in struct declaration

cli/src/commands/config/review_managed.rs
```diff
@@ -0,0 +1,127 @@
...
+#[instrument(skip_all)]
+pub fn cmd_review_managed(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &ConfigReviewManagedArgs,
+) -> Result<(), CommandError> {
+    if let Some(paths) = command.config_env().repo_managed_config_paths() {
+        // Treat an empty file the same as a nonexistent one.
+        let vcs = maybe_read_to_string(&paths.managed)?.unwrap_or_else(String::new);
+        let config = maybe_read_to_string(&paths.config)?.unwrap_or_else(String::new);
```
martinvonz: nit: `unwrap_or_default()`?

cli/src/commands/config/review_managed.rs
```diff
@@ -0,0 +1,127 @@
...
+        Ok(())
+    } else {
+        Err(user_error(
+            "Unable to detect location of config files. Are you in a repo?",
+        ))
+    }
+}
+
+fn reconstruct<'a>(sections: &[scm_record::Section<'a>]) -> String {
+    let mut out: Vec<&Cow<'a, str>> = Default::default();
```
martinvonz: nit: It seems cleaner to make this a `Vec<&'a str>`  to start with instead of converting it at the end of the function. It does make the `out.extend(lines)` more complicated (e.g. `out.extend(lines.iter().map(AsRef::as_ref))`) but I think that's fine.

I think you can also remove the explicit lifetimes.

cli/src/config.rs
martinvonz: Oh, but the config is checked on every command, right? Then perhaps we should avoid looking up this file in the repo. I'm not sure. TODO seems good.

matts1: I've added a TODO when reading on every invocation to check if it's out of date, but I did it for review-managed.

cli/src/config.rs
```diff
@@ -528,6 +566,71 @@ impl ConfigEnv {
             .transpose()
     }
 
+    /// Loads repo-managed config file for the user into the given `config`.
+    /// The old repo-managed config layer will be replaced if any.
+    fn reload_repo_managed_config(
+        &self,
+        ui: &Ui,
+        config: &mut RawConfig,
+    ) -> Result<(), ConfigLoadError> {
```
yuja: Make it return `CommandError`? It's odd that console I/O fails with "failed to read configuration file", and `ConfigLoadError` has `::Get`.

cli/src/config.rs
```diff
@@ -528,6 +566,71 @@ impl ConfigEnv {
...
+            writeln!(ui.hint_default(), "Run `jj config review-managed`")
+                .context(PathBuf::from("stderr"))
+                .map_err(ConfigLoadError::Read)?;
+        }
+        Ok(())
+    }
+
+    // When we pass UI it can't be instrumented (since it doesn't implement
+    // std::fmt::Debug), so we split this up into two functions.
+    #[instrument]
```
yuja: `instrument(skip(..))`?

cli/src/config.rs
```diff
@@ -528,6 +566,71 @@ impl ConfigEnv {
...
+
+            // TODO: Currently, this relies on the file being materialized on
+            // disk. Technically, we should do something like we do in
+            // review_managed.rs where we load from the repo.
+            // On the other hand, martinvonz says:
+            // > Oh, but the config is checked on every command, right? Then
+            // > perhaps we should avoid looking up this file in the repo.
+            // > I'm not sure.
+            let vcs_content = match &self.workspace_root {
+                None => "".to_string(),
```
yuja: nit: `.expect(..)`? I think `workspace_root` should be set if `repo_managed_config_paths()` is not `None`.

cli/src/config.rs
```diff
@@ -528,6 +566,71 @@ impl ConfigEnv {
...
+                Some(workspace_root) => {
+                    let managed_path = workspace_root.join(paths.managed.as_internal_file_string());
```
yuja: `paths.managed.to_fs_path(..)`

cli/src/config.rs
```diff
@@ -528,6 +566,71 @@ impl ConfigEnv {
...
+                    maybe_read_to_string(&managed_path)
+                        .context(&managed_path)
+                        .map_err(ConfigLoadError::Read)?
+                        .unwrap_or_default()
+                }
+            };
+            let last_reviewed_content = maybe_read_to_string(&paths.last_reviewed)
```
yuja: nit: simply read to `Vec<u8>`?

cli/src/config.rs
```diff
@@ -528,6 +566,71 @@ impl ConfigEnv {
...
+                .context(paths.last_reviewed)
+                .map_err(ConfigLoadError::Read)?
+                .unwrap_or_default();
+            vcs_content == last_reviewed_content
+        } else {
+            true
+        })
```
yuja: nit: could be `let ... else { return Ok(true); };`.

cli/src/commands/config/review_managed.rs
```diff
@@ -0,0 +1,151 @@
...
+    trust: bool,
+}
+
+#[instrument(skip_all)]
+pub fn cmd_review_managed(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &ConfigReviewManagedArgs,
+) -> Result<(), CommandError> {
+    if let Some(paths) = command.config_env().repo_managed_config_paths() {
```
yuja: Maybe we should check `repo-managed-config.enabled` first? or make `repo_managed_config_paths()` return `None` if disabled?

cli/src/commands/config/review_managed.rs
```diff
@@ -0,0 +1,151 @@
...
+        let workspace_command = command.workspace_helper(ui)?;
+        let path_converter = workspace_command.path_converter();
+        let store = workspace_command.repo().store();
+        let commit = workspace_command.resolve_single_rev(ui, &crate::cli_util::RevisionArg::AT)?;
+        // Use into_resolved to prevent attempting to set the config from a
+        // merge-conflicted file.
+        let vcs = match commit
+            .tree()?
+            .path_value(&paths.managed)?
```
yuja: Since `reload_repo_managed_config_internal()` relies on the checked-out content, it might be better to load the config file from disk.

cli/tests/test_repo_managed_config.rs
```diff
@@ -0,0 +1,83 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::common::TestEnvironment;
+
+#[test]
+fn test_repo_managed_config() {
```
yuja, line 18: Can you add test for `repo-managed-config.enabled = true` set in `.jj/repo/config.toml` and `.config/jj/config.toml`?

Maybe it's interesting to see what happens if the managed config is enabled by the cwd repo, but disabled in the repo specified by `-R`.

yuja (commented):
> * Can we avoid the zip file problem
>   * It's definitely possible, but there's not a single solution. There's many different options we could take. I don't want to block this CL on a discussion on that, since this CL does not degrade security (it uses an insecure method, but it's the same method we already use for existing config files).

I prefer to address this security concern because repo-managed config is new feature, but I'm not strongly against adding it as the current form if people agree.

cli/src/commands/config/review_managed.rs
```diff
@@ -0,0 +1,151 @@
...
+pub fn cmd_review_managed(
+    ui: &mut Ui,
+    command: &CommandHelper,
+    args: &ConfigReviewManagedArgs,
+) -> Result<(), CommandError> {
+    if let Some(paths) = command.config_env().repo_managed_config_paths() {
+        let workspace_command = command.workspace_helper(ui)?;
+        let path_converter = workspace_command.path_converter();
+        let store = workspace_command.repo().store();
+        let commit = workspace_command.resolve_single_rev(ui, &crate::cli_util::RevisionArg::AT)?;
```
PhilipMetzger: nit: import `crate::cli_util` so this only is a  `RevisionArg::AT`

cli/src/commands/config/review_managed.rs
matts1: I think this actually helps sometimes when it comes to sparse checkouts. The file may not yet exist on disk, but the backend can probably reasonably be expected to put it on disk when you attempt to read it, which makes this work for sparse checkouts when it otherwise wouldn't.

yuja: But the user would get warned because the checked-out content differs from the last snapshot, right? That seems bad UX.

cli/tests/test_repo_managed_config.rs
yuja: `--scope`/`--when` will also be an interesting test vector. I don't know whether we should respect `repo-managed-config.enabled` set in repo configs, though. Since it is the config knob to load config files, there would be weird edge cases.

cli/src/commands/config/review_managed.rs
matts1: If that's indeed how it worked, I do agree with you. My understanding of how it worked would be that the backend should cache the file on disk when review-managed attempts to load it and so the user wouldn't get warned. Is that incorrect?

cli/tests/test_repo_managed_config.rs
matts1: `repo-managed-config.enabled` set in repo configs is ignored (unless there's some mechanism through which we reload the config files multiple times in one jj invocation?).

The current situation if we enable it in a config file is that we:
1) Check whether repo configs are enabled
2) Enable them only if they're enabled
3) Print a warning if they're now disabled saying that the attempt to disable it was ignored.

`--scope/--when` should be fine as long as the scope can be resolved before the command starts running.

cli/src/commands/config/review_managed.rs
yuja: Do you mean `store.read_file()` updates the working-copy content? Then, the answer is no. In any case, adding tests for sparse checkout will help.

cli/tests/test_repo_managed_config.rs
yuja: > (unless there's some mechanism through which we reload the config files multiple times in one jj invocation?).

Just for reference, `run_internal()` loads config files from cwd to resolve command aliases, then switch to the workspace specified by `-R`. I don't know whether we should respect configuration loaded from cwd-relative workspace, but testing the current behavior will help understand the code.

matts1: You were right, it was very confusing, and in fact broke with scopes. It turns out the only solution is to resolve the config, see if `repo-managed-config.enabled` was set, then if it was, potentially re-resolve it.

I've fixed it and added a test for scopes.

cli/src/commands/config/review_managed.rs
matts1: I've just set it to use the disk for now and added a TODO, like martin originally suggested.

martinvonz:
> I prefer to address this security concern because repo-managed config is new feature, but I'm not strongly against adding it as the current form if people agree.

I agree with this, but it sounded like we couldn't come up with a solution that works for both cases because of the problem with approving a config that points a `jj fix` tool path that's inside the repo (and similar problems with paths into the repo). I unfortunately don't think I'll find time to think much about this in the near future and I don't want to block this feature on that.

cli/tests/test_repo_managed_config.rs
yuja: or ignore scoped/repo configs if that's easy? I don't think `repo-managed-config.enabled` has to be looked up by `COnfigEnv`. It could be set by e.g. `CliRunner` after loading user settings.

matts1: We definitely need some way to toggle this on a per-repo basis. I'd be open to the idea of it only being toggleable via repo configs and not via scopes. However, we'd need to be able to warn the user if they try and enable it through a scope and we didn't support that.

Since I've already gone to the effort of supporting scopes, I don't really see  a reason to ignore them now.

yuja: > We definitely need some way to toggle this on a per-repo basis.

Is it bad to enable repo-managed configs globally, and reject them as needed by `jj config review`? I thought this configuration was a temporary hack until we'll implement proper support for repo-managed configs.

matts1: Hmm, I was thinking that maybe some users explicitly want to disable the feature and don't want to constantly have to reject it every time. Maybe it's fine though.

Even if it is temporary though, I think that as I've said, I've already gone to the effort of supporting scopes, so no reason to ignore them now.

matts1:
I did a complete rework of it and fixed the zip file problem by storing it in the `.local/state` directory.

It now does the stuff that I mentioned where we would store hashes. This allows it to work very well in a multi-workspace repo, or when you're frequently changing between commits with different hashes.

matts1:
I'm currently running into some issues on windows, but I don't have a windows machine to test on. I've just got a windows machine from work, but I've got to go through some red tape in order to install cargo on it.

I'd appreciate it if people would review it ignoring the error on windows, and I'll come back to it once the approval process is complete.

cli/src/repo_managed_config.rs
```diff
@@ -0,0 +1,199 @@
...
+pub const MANAGED_PATH: &str = ".config/jj/config.toml";
+const LAST_APPROVED: &str = "last_approved";
+
+fn jj_state_dir() -> Option<PathBuf> {
+    if let Ok(x) = std::env::var("XDG_STATE_HOME")
+        && !x.is_empty()
+    {
+        Some(Path::new(&x).join("jj"))
+    } else {
+        std::env::var("HOME")
```
jennings: `$HOME` isn't a standard environment variable on Windows.

The platform-specific place to put this would probably be in either `$AppData/jj` or `$LocalAppData/jj`, though that's a bit like the `Library/Application Support` directory on macOS, so maybe `$USERPROFILE/.local/jj` is more desirable.

matts1: Oh, that's good to know. I've never developed anything more complex than a few python scripts on windows, so I'm not familiar with this. Does anyone have any opinions on which of those options is better?

cli/src/repo_managed_config.rs
```diff
@@ -0,0 +1,199 @@
...
+            for component in repo_path.components() {
+                match component {
+                    std::path::Component::Prefix(prefix_component) => {
+                        // Sanitize windows path prefixes.
+                        let component = prefix_component
+                            .as_os_str()
+                            .to_string_lossy()
+                            .replace(":", "")
+                            .replace("/", "")
+                            .replace("\\", "");
```
jennings: The tests are barfing on the path `$TEST_ENV/home/.local/state/jj/repos/?C/Users/runneradmin/AppData/Local/Temp/jj-test-uJbU0e/repo/.jj/repo`.

This looks a lot like a DOS device path (starts with `\\?\C:\Users\...`) with the slashes stripped, which is how the question mark is getting there and creating an invalid path.

There's a canonicalize function in the dunce crate, but not all paths can be changed to "normal" paths. Nearly all real-world paths, but not all.

I also kind of worry about hitting MAX_PATH with this directory scheme. I don't know if that's an obsolete concern or if it is still an issue in modern Windows. I haven't hit it in a long time, anyway.

matts1: It seems that windows makes this far more difficult than I'd originally thought. I've taken a more simplified approach which should solve it. Now we hash it and just store the original path in another file, so we can work out what repo the config is for without having access to the repo (for cleanup, if required in the future).

cli/src/repo_managed_config.rs
matts1: I'm just gonna assume `$USERPROFILE/.local/jj` for now. We can change it later if need be.

cli/src/repo_managed_config.rs
jennings: Makes sense to me, thanks.

cli/src/config.rs
```diff
@@ -527,6 +543,39 @@ impl ConfigEnv {
...
+    #[instrument(skip(ui))]
+    fn try_reload_repo_managed_config(
+        &self,
+        ui: &Ui,
+        config: &mut RawConfig,
+    ) -> Result<bool, CommandError> {
+        if let Some(repo_config) = self.repo_managed_config()?
+            && let Some(layer) = self
+                .repo_managed_layer
+                .get_or_init(|| {
```
yuja: I don't think we need to cache the repo-managed config, but if needed, we can use `.get_or_try_init()` to report the error here.

If the purpose of caching is to suppress repeated warnings, I think we can instead add a separate function that loads repo-managed config when everything gets ready.

cli/src/config.rs
```diff
@@ -538,16 +587,40 @@ impl ConfigEnv {
...
+        ui: &Ui,
+        config: &mut RawConfig,
+    ) -> Result<StackedConfig, CommandError> {
+        let mut resolved = self.resolve_config_without_repo_managed(config)?;
+        // In case you call this function twice, we remove the layers so we don't
+        // duplicate.
+        resolved.remove_layers(ConfigSource::RepoManaged);
+        // The repo-managed config is enabled.
+        Ok(
+            if resolved.get("repo-managed-config.enabled")?
```
yuja, line 621: Since we now have a state directory for the repo-managed settings, I think we can track this `enabled` property there. I think that's better than introducing config knob to load config.

I personally think it's okay to enable repo-managed settings unconditionally, and add escape hatch later when needed.

cli/src/repo_managed_config.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    if let Ok(x) = std::env::var("XDG_STATE_HOME")
+        && !x.is_empty()
+    {
+        Some(Path::new(&x).join("jj"))
+    } else if let Ok(x) = std::env::var("HOME") {
+        Some(Path::new(&x).join(".local/state/jj"))
+    } else {
+        None
+    }
```
yuja: We can instead use `etcetera` to resolve platform-native state path. See `ConfigEnv::from_environment()`.

cli/src/repo_managed_config.rs
```diff
@@ -0,0 +1,192 @@
...
+///   approved content for config2
+pub struct RepoManagedConfig {
+    config_dir: PathBuf,
+    // Canonicalized repo path. It must be canonicalized because we hash this.
+    repo_path: PathBuf,
+}
+
+impl RepoManagedConfig {
+    pub fn new(repo_path: &Path) -> Result<Self, CommandError> {
+        let repo_path = dunce::canonicalize(repo_path)?;
```
yuja: nit: I think `repo_path` would be canonicalized by caller. Better to not do path normalization at various places.

cli/src/repo_managed_config.rs
```diff
@@ -0,0 +1,192 @@
...
+        let state = jj_state_dir().ok_or_else(|| {
+            user_error("Unable to calculate jj state dir. Is the home directory set?")
+        })?;
+        Ok(Self {
+            config_dir: state
+                .join("repos")
+                .join(Self::digest(repo_path.as_os_str().as_encoded_bytes())),
```
yuja: The encoding is unspecified and not guaranteed to be stable. We can use `file_utils::path_to_bytes()` instead.

https://doc.rust-lang.org/stable/std/ffi/struct.OsStr.html#method.as_encoded_bytes

cli/src/repo_managed_config.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/// Manages the repo-managed config. The file structure for the repo-managed
+/// config for a repo contained at /path/to/repo is:
+/// ~/.local/state/jj/repos/hash(/path/to/repo)/last_approved:
+///   hash(config1)
+/// ~/.local/state/jj/repos/hash(/path/to/repo)/hash(config1):
+///   approved content for config1
+/// ~/.local/state/jj/repos/hash(/path/to/repo)/hash(config2):
+///   approved content for config2
```
yuja, line 48: Suppose we need to support multiple workspaces with different revisions (= configs), maybe ~approved paths~ (EDIT: last approved path?) would have to be keyed by the workspace root, not by the repo path?

cli/src/repo_managed_config.rs
```diff
@@ -0,0 +1,192 @@
...
+        // and infrequently changed.
+        std::fs::create_dir_all(&self.config_dir).context(&self.config_dir)?;
+        // This isn't used right now, but it may be used in the future to clean
+        // up deleted repos.
+        let p = self.config_dir.join("repo_path");
+        std::fs::write(&p, self.repo_path.as_os_str().as_encoded_bytes()).context(p)?;
+        let p = self.approved_config(digest);
+        std::fs::write(&p, content).context(p)?;
+        self.approve(digest)
+    }
```
yuja, line 128: It's probably better to use `tempfile` and `persist_temp_file()` to not leave partially-written content.

cli/src/repo_managed_config.rs
```diff
@@ -0,0 +1,192 @@
...
+        workspace_root: &Path,
+    ) -> Result<(Option<PathBuf>, bool), CommandError> {
+        let config = self.get_vcs_config(workspace_root)?;
+        if let Some(config) = config
+            && !config.is_empty()
+        {
+            let digest = Self::digest(&config);
+            let out = self.approved_config(&digest);
+            let last_approved = self.last_approved()?;
+            if out.is_file() {
```
yuja: nit: safer to use `.try_exist()` here?

yuja (commented):
Thanks.

cli/src/config.rs
matts1: IMO we should leave the escape hatch in. Automated tooling that invokes `jj`, for example, may want to disable this, since they can't necessarily trust the repo. Since I've already gone to the effort of making it, I don't think I should remove it just because it's a rarer use case.

Also, I'm ok to enable them unconditionally soon, but it generally feels safer to submit a feature with it disabled by default, then send a follow-up PR to enable it.

cli/src/config.rs
matts1: While I agree that it's probably an unnecessary optimization, it is still an optimization. The purpose of the caching was both to suppress repeated warnings, and I figured while I was at it we could save a few file reads, config file parsing, and writes.

I don't think we should actively try to remove such an optimization, even if it is only a small one.

yuja: If it is an optimization, I think the state of `repo_managed_config()` can also be cached. I tend to remove unneeded caching because the caller should be aware of the cache layer and invalidate it when needed.

cli/src/config.rs
yuja: > Automated tooling that invokes jj, for example, may want to disable this, since they can't necessarily trust the repo.

Isn't the repo-managed config untrusted by default? I think automated tooling can ignore the warning.

> it generally feels safer to submit a feature with it disabled by default, then send a follow-up PR to enable it.

If the config knob is supposed to be temporary, I agree it's good approach. My concern is that this config knob can be a maintenance burden because its interpretation is order/layer-sensitive. The current implementation gets a bit messy because of that, I think.

matts1: > Isn't the repo-managed config untrusted by default? I think automated tooling can ignore the warning.

Correct, but if this is part of our CI, for example, I wouldn't want it to be spitting out the warnings to the logs that users have to debug.

I guess I'll enable it by default then. I've just done significant simplification and now it's no longer order dependent, we just do what we do in every other place and remove the layers before adding them back. So the code is much simpler now.

matts1:
@yuja @martinvonz This should be ready for re-review

mmhat:
> I was hoping we could get this reviewed and submitted soon.
> 
> I think there are some open questions, but they're a non-issue for this particular PR IMO:
> 
> * How do you manage switching between configs
>   
>   * While it's somewhat important, it doesn't affect the MVP
> * Can we avoid the zip file problem
>   
>   * It's definitely possible, but there's not a single solution. There's many different options we could take. I don't want to block this CL on a discussion on that, since this CL does not degrade security (it uses an insecure method, but it's the same method we already use for existing config files).

Well, here's another idea... (This is not intended as a change request; It's rather a rough idea I had while reading this thread here.)

When approving a repo-managed config:

1. Generate a cryptographic key in `$XDG_CONFIG_HOME/jj/key` (or `$XDG_STATE_HOME/jj/key`) unless there exists one.
2. Sign the repo-managed config with that key.
3. Store the signature somewhere in the `.jj` directory; Either in a log or somehow "attach" it to the revision that contains the approved repo-managed config.

Then, we can check for a certain revision if its config was approved by looking at the most recent signatures found in its ancestors, and if one of those is a valid certificate for the config we are fine.

It's just a rough idea, but that would allow us to keep evidence of the approval close to the repo-managed config without relying on some "global state" in `~/.local/...`.

yuja:
Hmm, true. We can use cryptographic hashing/signing with per-user secret key. That makes the zip file problem a totally separate issue from the repo-managed settings. Thoughts?

I'll review this PR later in this week. I didn't have time last week, sorry.

matts1:
That idea gave me a few other ideas. I didn't really like the idea of storing approvals inside the `$XDG_STATE_HOME` directory, for a few reasons:
* Approvals are based on the repo path, not the repo itself.
  * Moving the repo directory breaks it
  * Copying the repo directory breaks it
  * If you put a different repo inside the same directory, you'll have preapprovals
* Cleanup is difficult

So what if instead we kept both user and repo-user config inside the `.jj` directory, but added that cryptographic key you suggested to allow us to sign things. Then instead of ensuring that we don't load config from zip files, we create a signed data structure that cannot be tampered with.

We would then create the following data structure stored in the `.jj` directory:
```
struct Security {
  // This is randomly generated
  // This ensures that if you upload your own zip file, someone can't just send you a zip file with the same secure_repo_id
  salt: Vec<u8>,

  // This is sign(repo_path + salt)
  // We include the repo path to ensure that if you upload your own zip file, someone can't just send you a zip file with the same salt and signature.
  // Note that this specifically does not work if you uploaded a repo rooted at /path/to/repo, and then you download the repo to the same directory. However, this also wouldn't be solved by storing the config elsewhere, as then they could just use your preapproved configs.
  // If this does not match the signature we generate, we disable user configs.
  secure_repo_id: Signature,
  
  // Key: sign(secure_repo_id + repo config)
  // Value LHS: sign(secure_repo_id + repo_config + value RHS)
  // Value RHS: approved config for the given config
  // To look up an entry for a given config, we can resign it to recalculate the key.
  // We then need to also sign it including the value to ensure that the value was not tampered with.
  approved_repo_configs: map<Signature, (Vec<u8>, Signature)>
}
```

Note that the addition of this feature would mark existing user configs as insecure (as would any method that stops the zip file problem). We could work around this issue by automatically adding salts and marking repos as trusted for the first few months, then eventually stopping this practice.

joyously:
Does that still allow for workspaces and local clones?
(sure seems more simple to just not read repo configs)

Does this impact CI testing of a repo with a repo config? (specifically the `jj` repo, if it had a repo config)

matts1:
> Does that still allow for workspaces and local clones?

Yes, that's unaffected

> Does this impact CI testing of a repo with a repo config? (specifically the `jj` repo, if it had a repo config)

No. A ci bot would not be able to approve repo Configs, so would ignore them

PR #6991, opened 2025-07-15
config: add --when.config to match previously-set config values

Gingeh:
I wanted to be able to set my user.name and user.email based on a single per-repo value:
```toml
[[--scope]]
--when.config.org = 'work'
user.email = "YOUR_WORK_EMAIL@example.org"

[[--scope]]
--when.config.org = 'oss'
user.email = "YOUR_OSS_EMAIL@example.org"
```

# Checklist

If applicable:

- [ ] I have updated the config schema (`cli/src/config-schema.json`) Not Applicable (?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   It still fully traverses them while snapshotting but they won't clutter up
   the output with all of their contents.
 
+* Added `--when.config` scope condition to match on previously-set config values.
+
 ### Fixed bugs
 
 ### Packaging changes
--- a/docs/config.md
+++ b/docs/config.md
@@ -1734,6 +1734,16 @@ paginate = "never"
 --when.commands = ["diff", "show"]
 [--scope.ui]
 pager = "delta"
+
+# override user.email based on a custom config value,
+# ..which can be set per-repository
+[[--scope]]
+--when.config.org = 'work'
+user.email = "YOUR_WORK_EMAIL@example.org"
+
+[[--scope]]
+--when.config.org = 'oss'
+user.email = "YOUR_OSS_EMAIL@example.org"
 ```
 
 #### Using multiple files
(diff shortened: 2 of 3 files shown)


yuja:
This was discussed in https://github.com/jj-vcs/jj/pull/5538, but I personally don't want to introduce conditional that depends on config values because doing that would make the resolution order dependent.

> I wanted to be able to set my user.name and user.email based on a single per-repo value:

Maybe you know, but a common workaround for this scenario is to use separate base directories (e.g. `~/work`, `~/oss`.)

Gingeh:
Ah, I hadn't seen that discussion. I personally don't feel that this makes resolution too complicated, but that's ultimately for you (and other maintainers) to decide. @ilyagr's idea of a dedicated `[config-toggles]` table ([here](https://github.com/jj-vcs/jj/pull/5538#issuecomment-2629969079)) would also solve this problem, but would require some way to delay scope resolution until all `[config-toggles]` are resolved.

> Maybe you know, but a common workaround for this scenario is to use separate base directories

Yeah I'm aware of that workaround but would very much prefer my configuration to not be location-dependent.

yuja:
> I personally don't feel that this makes resolution too complicated

A potential problem is that the previously-evaluated condition can change if a new `--when` block is enabled. We could make that an error or order-dependent, but that seems odd.

Gingeh:
Yeah that scenario is included in the tests I added. Would it be better to make the behaviour more explicit in the documentation?

Gingeh:
also note that scope resolution already depends on the evaluation order:
```toml
[[--scope]]
foo = 1

[[--scope]]
foo = 2
```

yuja:
> also note that scope resolution already depends on the evaluation order:

Yes, but the result should be expected. The last value always wins.

If a new `--when` block is enabled and if it contains values referred to by previous conditionals, we'll need to re-evaluate the previous blocks until the resolution converses. That's the complexity I mentioned in the other PR.

PhilipMetzger:
@Gingeh it'd be nice if you could provide a bit more background info on your use-case with it because conditional evaluating config is something icky a la C++ SFINAE (.e post-monophorism errors) which isn't that liked in Rust circles.

Gingeh:
re: @yuja 
> Yes, but the result should be expected. The last value always wins.

Same with this PR, scopes only see previously defined scopes.

> If a new `--when` block is enabled and if it contains values referred to by previous conditionals, we'll need to re-evaluate the previous blocks until the resolution converses. That's the complexity I mentioned in the other PR.

That's not what my PR does. Each layer only needs to be evaluated once because they only depend on values defined in previous layers.

yuja:
> > Yes, but the result should be expected. The last value always wins.
> 
> Same with this PR, scopes only see previously defined scopes.

I meant this PR introduces new rule. Scopes only see the previously-resolved (intermediate) values, but the resolved configs have the final values.

@ilyagr had an idea to introduce a section dedicated for `--when.config` to work around the problem. I personally don't like it, but I think it's less bad than allowing arbitrary overrides.

https://github.com/jj-vcs/jj/pull/5538#issuecomment-2629969079

PR #6995, opened 2025-07-16
diffedit: add support for filesets

### LINKED ISSUE #6973: FR: Accept fileset argument to `jj diffedit`

gechr:
Closes #6973

Follow-up to https://github.com/jj-vcs/jj/pull/6988 (which I can't re-open because the branch was deleted)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   diff formatters. Built-in tools that do not function well as diff editing
   tools or as diff formatters will now report an error when used as such.
 
+* `jj diffedit` now accepts filesets to edit only the specified paths.
+
 ### Fixed bugs
 
 ### Packaging changes
--- a/cli/src/commands/diffedit.rs
+++ b/cli/src/commands/diffedit.rs
@@ -17,11 +17,11 @@ use std::io::Write as _;
 use clap_complete::ArgValueCandidates;
 use clap_complete::ArgValueCompleter;
 use itertools::Itertools as _;
-use jj_lib::matchers::EverythingMatcher;
 use jj_lib::object_id::ObjectId as _;
 use jj_lib::rewrite::merge_commit_trees;
 use tracing::instrument;
 
+use crate::cli_util::print_unmatched_explicit_paths;
 use crate::cli_util::CommandHelper;
 use crate::cli_util::RevisionArg;
 use crate::command_error::CommandError;
@@ -77,6 +77,13 @@ pub(crate) struct DiffeditArgs {
         add = ArgValueCompleter::new(complete::revset_expression_mutable),
     )]
     to: Option<RevisionArg>,
+    /// Edit only these paths (unmatched paths will remain unchanged)
+    #[arg(
+        value_name = "FILESETS",
+        value_hint = clap::ValueHint::AnyPath,
+        add = ArgValueCompleter::new(complete::modified_revision_or_range_files),
+    )]
+    paths: Vec<String>,
     /// Specify diff editor to be used
     #[arg(
         long,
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   It still fully traverses them while snapshotting but they won't clutter up
   the output with all of their contents.
 
+* `jj diffedit` now accepts filesets to restrict the diff to specific paths.
```
martinvonz: nit: Here and in the argument description, I would say "to edit only the specified paths". I think "restrict the diff" can be interpreted to mean that the unmatched paths will be removed from the diff.

cli/src/merge_tools/builtin.rs
```diff
@@ -1758,6 +1783,59 @@ mod tests {
...
+        let matcher = FilesMatcher::new([matched_path]);
+
+        let (changed_files, files) =
+            make_diff_with_matcher(store, &left_tree, &right_tree, &matcher);
+
+        insta::assert_debug_snapshot!(changed_files, @r#"
+        [
+            "matched",
+        ]
+        "#);
```
martinvonz: nit: this is such a simple type that i would not use snapshot testing here. `assert_eq!(changed_files, vec![matched_path.to_owned()]);` should work, right?

cli/src/merge_tools/builtin.rs
```diff
@@ -1758,6 +1783,59 @@ mod tests {
...
+            changed_files,
+            &files,
+            ConflictMarkerStyle::Diff,
+        )
+        .unwrap();
+        let result_tree = store.get_root_tree(&result_tree_id).unwrap();
+
+        let matched_id = to_file_id(&result_tree, matched_path);
+        let matched_content = testutils::read_file(store, matched_path, &matched_id);
+        assert_eq!(matched_content, b"left matched\n");
```
martinvonz: nit: I think it's simpler to compare the `MergedTreeValue`s instead of the contents:

```rust
assert_eq!(result_tree.path_value(matched_path).unwrap(), left_tree.path_value(matched_path).unwrap());
```

cli/src/commands/diffedit.rs
```diff
@@ -77,6 +77,13 @@ pub(crate) struct DiffeditArgs {
         add = ArgValueCompleter::new(complete::revset_expression_mutable),
     )]
     to: Option<RevisionArg>,
+    /// Restrict the diff to these paths
```
martinvonz: Even if replace "Restrict" by "Edit only" here, it can't hurt to explicitly say that unmatched paths will be unchanged.

martinvonz (changes requested):
Could you split this up in two commits? At least it seems to me that addressing the FIXME could be a separate commit before adding support for filesets to `diffedit`. Then you also get a chance to explain what that first commit is about.

gechr:
@martinvonz This should be good to go now.

cli/src/commands/diffedit.rs
martinvonz: This doesn't seem to be addressed

gechr: I originally changed it to "Edit only these paths" - I assume you were referring to the "unmatched paths will be unchanged" bit not being addressed, so I've just pushed another commit to add that as well.

https://github.com/jj-vcs/jj/compare/7664e1a1338adc8b5939dd5ff3e4a2aa08069ba5..8b9f593e29fc94941eb58b600d8ec0e072e0fc2e

martinvonz (approved):
Thanks!

ilyagr:
Thank you!

ilyagr:
I wonder whether we could make `jj diffedit xyz` print the same warning as `jj diff xyz`, "Warning: No matching entries for paths: xyz". This would be useful if the user meant `jj diffedit -r xyz`.

Unfortunately, with `jj diffedit`, the warning would be less visible; I'm not sure what we can do about that.

PhilipMetzger:
This is the everlasting `-p` discussion, see #3809. But to be consistent it should be done.

Issue #5744, opened 2025-02-18
FR: Make individual revset (etc.) functions linkable in docs


arxanas:
**Is your feature request related to a problem? Please describe.**

Currently, you can get the link to a section in the docs via the paragraph symbol/icon for the headers. For example: https://jj-vcs.github.io/jj/v0.26.0/revsets/#functions

However, I don't see a way to link to an individual functions (since they're just entries in an unordered list?).

**Describe the solution you'd like**

Add buttons/links to the docs for individual functions.
- If we add them as separate headers, then they'd also appear in the sidebar, which is good? But if we add them as separate headers, the docs might get a little clunky/unwieldy.
- It might also work to add inline anchors (and somehow surface them to users).

**Describe alternatives you've considered**

In the longer term, it might be good to generate formal documentation for the revset (etc.) functions via a normal API reference system.

**Additional context**

N/A

shanesveller:
There are preferable solutions in revising how the docs are generated, but linking to text fragments that don't have individual anchor tags appears to be [supported in all major browsers](https://web.dev/articles/text-fragments) now.

Example:
https://jj-vcs.github.io/jj/v0.26.0/revsets/#functions:~:text=empty()

lgarron:
> There are preferable solutions in revising how the docs are generated, but linking to text fragments that don't have individual anchor tags appears to be [supported in all major browsers](https://web.dev/articles/text-fragments#browser_compatibility) now.
>
> Example:
> https://jj-vcs.github.io/jj/v0.26.0/revsets/#functions:~:text=empty()

Note that this will link the first occurrence, which may be an earlier reference. For example, https://jj-vcs.github.io/jj/v0.26.0/revsets/#:~:text=visible_heads() will show nowhere near the `visible_heads()` definition. (The `#functions` part of the example above also doesn't do anything to help.) There are obviously workarounds, but adding actual anchors to the page would be the most stable.

PR #6996, opened 2025-07-16
Add cross-links to type references in the template language reference page.

lgarron:
Note that this does not include references to functions, such as the reference to `self.commit_id()` in the "Commit keywords" section. This is because functions themselves do not have link anchors and I consider it out of scope to add them here. (It would be nice to add such cross-reference links to functions in the revset language reference page, but that's also out of scope here and would additionally also require adding link anchors to all the functions. Maybe in a follow-up)

Since there are a limited number of types (and all but one are single-token), it's relatively straightforward to perform this update programmatically using naïve text replacement. Note that:

- Unfortunately, the only semantic way to include links insicde code in Markdown is to convert the relevant inline code snippets to `<code>` tags.
- To keep both the code and the diff as simple as possible, I did not try to rewrap the Markdown text after transformation. This could be done in an additional formatting commit if desired, but it would certainly produce a lot of diff noise (although adding the commit to `.git-blame-ignore-revs` would help).

I used the following code:

```ts
#!/usr/bin/env bun

// Run from: `./docs/templates.transform.ts`

// Uses: https://bun.sh/
// The code is also trivial to rewrite for plain `node`, but I wanted to use a few type annotation as guardrails.

import { file } from "bun";

const contents = (
	await file(new URL("./templates.md", import.meta.url)).text()
).replaceAll("List<Trailer>", "ListTrailer");

const typeKeywords = [
	"AnnotationLine",
	"Boolean",
	"Commit",
	"ChangeId",
	"CommitId",
	"CommitRef",
	"ConfigValue",
	"CryptographicSignature",
	"DiffStats",
	"Email",
	"Integer",
	"List",
	"ListTemplate",
	"Operation",
	"OperationId",
	"Option",
	"RefSymbol",
	"RepoPath",
	"Serialize",
	"ShortestIdPrefix",
	"Signature",
	"SizeHint",
	"String",
	"Stringify",
	"Template",
	"Timestamp",
	"TimestampRange",
	"Trailer",
	"TreeDiff",
	"TreeDiffEntry",
	"TreeEntry",
]
	.reverse() // Reverse to avoid prefix collisions.
	.concat(["ListTrailer"]); // Place `ListTrailer` last to avoid being pre-empted by `List` or `Trailer`.

function typeKeywordToLink(keyword: string): string {
	// We don't perform escaping because we know the set of type keywords is safe.
	return `#${keyword.toLowerCase().replaceAll(/[^a-z]/g, "")}-type`;
}

function typeKeywordToTextContent(keyword: string): string {
	if (keyword === "ListTrailer") {
		return "List&lt;Trailer&gt;";
	}
	return keyword;
}

const lines = contents.split("\n");

const newContents = lines
	.map((line) => {
		if (line.startsWith("#")) {
			return line;
		}

		return line.replaceAll(/`([^`\n]+)`/g, (_, group) => {
			let out = group;
			// Iterating over all keywords is not clever, but it's more than fast enough.
			for (const typeKeyword of typeKeywords) {
				// We don't perform escaping because we know the set of type keywords is safe.
				out = out.replaceAll(
					new RegExp(`\\b${typeKeyword}\\b`, "g"),
					`<a href="${typeKeywordToLink(typeKeyword)}">${typeKeywordToTextContent(typeKeyword)}</a>`,
				);
			}
			return out === group ? `\`${out}\`` : `<code>${out}</code>`;
		});
	})
	.join("\n")
	.replaceAll("ListTrailer", "List<Trailer>");

await file(new URL("./templates.md", import.meta.url)).write(newContents);
```

# Checklist

If applicable:


lgarron:
This PR isn't associated with an issue, because I figured I might as well test it out first. You can see it working at: https://github.com/lgarron/jj/blob/template-reference-cross-links/docs/templates.md

(The generated anchor links on the website match those generated by the GitHub UI, including [`#listtrailer-type`](https://github.com/lgarron/jj/blob/template-reference-cross-links/docs/templates.md#listtrailer-type).)

PhilipMetzger:
this should contain the title `docs:` per our commit guidelines outlined here: https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#commit-guidelines and can be squashed down to a single commit. I like the idea though

lgarron:
> this should contain the title `docs:` per our commit guidelines outlined here: https://github.com/jj-vcs/jj/blob/main/docs/contributing.md#commit-guidelines and can be squashed down to a single commit. I like the idea though

Happy to squash if you think the contents are appropriate as-is. Is that the case? :D

PhilipMetzger:
I'm happy with the contents so you can squash and adjust the _commit message_ not the PR title. cc @arxanas who wanted that in  #5744

lgarron:
> I'm happy with the contents so you can squash and adjust the _commit message_ not the PR title. cc @arxanas who wanted that in #5744

Done!

PhilipMetzger:
the commit description still hasn't changed to something to the suggested topic though. An obvious and simple fix is to change "Add cross-links to type references in the template language reference page." => "docs: Add cross-links to type references in the template language reference page.

lgarron:
> the commit description still hasn't changed to something to the suggested topic though. An obvious and simple fix is to change "Add cross-links to type references in the template language reference page." => "docs: Add cross-links to type references in the template language reference page.

Ah, sorry, fixed.

ilyagr:
It took me a moment to figure out what this does, because the first few changes are to links that were already links before this PR.

Perhaps add an introductory sentence like "This adds a link to the relevant type for every argument of every function in the template language reference" to the commit and PR descriptions?

-------------

Thanks, it's a nice proof of concept!

I'm not entirely sure whether this is worth making the Markdown uglier and harder to edit. It might be. Or perhaps this could be better as some sort of mkdocs plugin or `mkdocs-material` plugin that linkifies every mention of `Commit` that's inside a code block in this file. (This is where I'm leaning right now, but that'd take some work and research which you may or may not be up to)

------------

If we did decide we want to go with the approach you took, I think we would want to re-wrap the Markdown, perhaps as a separate commit. We'd also want to do it to the revset language page.

lgarron:
> It took me a moment to figure out what this does, because the first few changes are to links that were already links before this PR.

Yeah, this is why I didn't squash the commits until asked. (The first commit had made that clear. I also have a sentence about it in the PR description, but it's more buried.)

> This adds a link to the relevant type for every argument of every function in the template language reference

I mean, that's pretty much of what the title says? I wrote "type references" instead of "arguments" because arguments are only some of the references. Certainly open to something else more accurate if you think it's warranted.

> I'm not entirely sure whether this is worth making the Markdown uglier and harder to edit. It might be. 

Yeah, I'm also conflicted over this. There isn't really an alternative in Markdown syntax, though.

For me personally, having these links would have made some of my work these last few days a *lot* easier, though. If this doesn't land, I'll probably use my local fork over the website. 🤷

> Or perhaps this could be better as some sort of mkdocs plugin or `mkdocs-material` plugin that linkifies every mention of `Commit` that's inside a code block in this file. (This is where I'm leaning right now, but that'd take some work and research which you may or may not be up to)
> 
> If we did decide we want to go with the approach you took, I think we would want to re-wrap the Markdown, perhaps as a separate commit. We'd also want to do it to the revset language page.

I mean, I'd be up for a bit of that, but it would be a very different scope and I have a lot of other open source contributions to work on. Note that there are some things that would probably require some tweaks to work with any tool, like `List<Trailer>`. So the only "real" solution would be to have something that understands the grammar and the multi-token references, at which point it would probably be better to generate the docs rather than maintain them in Markdown itself.

But it would definitely be nice to have function anchors as well, especially for the revset reference.

PR #6998, opened 2025-07-17
template: Add the replace method to string

matts1:
# Checklist

If applicable:

### DIFF
--- a/cli/src/template_builder.rs
+++ b/cli/src/template_builder.rs
@@ -1023,6 +1023,16 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
         },
     );
     map.insert(
+        "replace",
+        |language, diagnostics, build_ctx, self_property, function| {
+            let [from, to] = function.expect_exact_arguments()?;
+            let from = expect_stringify_expression(language, diagnostics, build_ctx, from)?;
+            let to = expect_stringify_expression(language, diagnostics, build_ctx, to)?;
+            let out_property = (self_property, from, to).map(|(s, from, to)| s.replace(&from, &to));
+            Ok(out_property.into_dyn_wrapped())
+        },
+    );
+    map.insert(
         "escape_json",
         |_language, _diagnostics, _build_ctx, self_property, function| {
             function.expect_no_arguments()?;
@@ -2814,6 +2824,10 @@ mod tests {
         insta::assert_snapshot!(
             env.render_ok(r#""bar@other.example.com".remove_suffix("@other.example.com")"#),
             @"bar");
+
+        insta::assert_snapshot!(
+            env.render_ok(r#""abc abc def".replace("bc", "ghi")"#),
+            @"aghi aghi def");
 
         insta::assert_snapshot!(env.render_ok(r#"" \n \r    \t \r ".trim()"#), @"");
         insta::assert_snapshot!(env.render_ok(r#"" \n \r foo  bar \t \r ".trim()"#), @"foo  bar");
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -408,6 +408,8 @@ defined.
 * `.lines() -> List<String>`: Split into lines excluding newline characters.
 * `.upper() -> String`
 * `.lower() -> String`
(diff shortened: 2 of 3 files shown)


matts1:
My use case here is that I want a template that outputs the canonical form of a list of bugs associated with a commit. Our bugs can be:
* `b:12345`
* `b/12345`
* `12345`

Additionally, they can be seperated by spaces, commas, or ", ". Replace is required to canonicalize this.

This is required to be done in the templating language because I'd like to be able to pre-fill a commit description with the bugs from all of its parents.

yuja:
cc @lf-

We have ongoing PRs that add `regex:<pattern>`. I think it's better to add `.replace(pattern, subst)` after these PRs get merged.

https://github.com/jj-vcs/jj/issues/6893

lf-:
yeah here: https://github.com/jj-vcs/jj/pull/6899

i think that we might want to do some way to grab capture groups while doing the api design for replacement

PhilipMetzger:
Can you put your motivation into the commits description, thanks.

joneshf:
This is great!

Could this be given an optional `count: Integer` argument, so we can choose the maximum number of replacements to make (like [`String#replacen` in Rust](https://doc.rust-lang.org/std/string/struct.String.html#method.replacen))? So it'd be something like:

> `.replace(from: Stringify, to: Stringify[, count: Integer]) -> String`: Replaces up to `count` instances
  of `from` with `to`. With no `count`, replaces all instance of `from` with `to`.

Or however makes sense to word it.

lf-:
https://github.com/jj-vcs/jj/pull/6899 is merged, feel free to rebase and integrate with the new state of the world now that you can take string templates as parameters :)

cli/src/template_builder.rs
```diff
@@ -1022,6 +1022,16 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
             Ok(out_property.into_dyn_wrapped())
         },
     );
+    map.insert(
+        "replace",
+        |language, diagnostics, build_ctx, self_property, function| {
+            let [from, to] = function.expect_exact_arguments()?;
+            let from = expect_stringify_expression(language, diagnostics, build_ctx, from)?;
```
lf-, line 1029: This should take a string template like `match` does.

cli/src/template_builder.rs
```diff
@@ -1022,6 +1022,16 @@ fn builtin_string_methods<'a, L: TemplateLanguage<'a> + ?Sized>(
...
+            let to = expect_stringify_expression(language, diagnostics, build_ctx, to)?;
```
lf-, line 1030: I wonder if this should be stringify at all, if it should be a non-coercing string, or if it should be forced to be a literal string. The reason to not allow it to be stringify is that dealing with dynamic template strings is a known problem spot in basically every language (c.f. printf vulnerabilities in C, etc) since you have to deal with invalid ones, and it would probably be surprising behaviour to the user to allow them to shoot themselves in the foot with this not being interpreted as a literal string.

Specifically, in a world where this supports `$1`, `$2`, etc regex capture group references, allowing this to vary could prove somewhat annoying.

docs/templates.md
```diff
@@ -408,6 +408,8 @@ defined.
 * `.lines() -> List<String>`: Split into lines excluding newline characters.
 * `.upper() -> String`
 * `.lower() -> String`
+* `.replace(from: Stringify, to: Stringify) -> String`: Replaces all instances
```
lf-, line 411: as mentioned, should support a number of replacements to make as optional arg :)

matts1:
I don't think I really have time to finish this off. My other PRs are much higher priority, and have been a pretty large time sink. If anyone else wants to finish this off, they're welcome to.

gechr:
@matts1 I hadn't seen your PR and started working on this separately myself. My local version supports capture groups and already addresses some of the review comments. I was planning to raise a separate PR for it, unless you've had a change of heart and want to finish this off?

matts1:
Please do, that sounds great. In that case I'll close this PR

gechr:
@matts1 Thanks! Opened https://github.com/jj-vcs/jj/pull/7467

PR #6999, opened 2025-07-17
template: commit trailers respect JJ: prefixes when not using an editor.

matts1:
This will allow users to write instructions in their commit trailers such as:

```
JJ: Feel free to modify this if you did any local testing
Test: CI/CQ
```

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * `jj fix` now buffers lines from subprocesses' stderr streams and emits them a
   complete line at a time. Each line is prepended with the file name.
 
+* commit trailers now respect `JJ:` prefixes and ignore them when not using an
+  editor.
+
 ### Fixed bugs
 
 ### Packaging changes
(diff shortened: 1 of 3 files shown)


yuja:
This seems incorrect because the trailer template output is parsed as trailers, not as description lines with comments. I haven't tried, but multiple `"JJ: foo\nJJ: foo\n"` lines would be deduplicated, for example.

@glehmann, thoughts?

glehmann:
I don't think `commit_trailer` is the right place for that.
I'm not sure to understand the use case. Can't it be done with `default_commit_description` template?

matts1:
> I don't think `commit_trailer` is the right place for that. I'm not sure to understand the use case. Can't it be done with `default_commit_description` template?

The problem I had with that is that default_commit_description is a variable, not a function. I wanted to implement a feature where commits would inherit their bugs from their parents.

Here, if a parent had `Bug: 123,456` in their commit trailers, a child commit would inherit the `Bug: 123,456`

```
[template-aliases]
'bugs(change)' = '''stringify(change.trailers(
).filter(|trailer| trailer.key() == "Bug" && !trailer.value().lower().contains("none")
).map(|trailer| trailer.value()
).join(","))
'''
'parent_bugs(change)' = 'stringify(change.parents().filter(|c| !c.immutable()).map(|c| bugs(c)).join(","))'

[templates]
commit_trailers = '''
    if(!trailers.contains_key("Bug"), if(parent_bugs(self),
        "JJ: This bug was inherited from the parent change. Please ensure it's correct.\nBug: " ++
        parent_bugs(self),
    )),
'''
```

PR #7000, opened 2025-07-17
docs: explain how e.g. the commit_id() revset can override a bookmark

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -53,6 +53,10 @@ Jujutsu attempts to resolve a symbol in the following order:
 3. Git ref
 4. Commit ID or change ID
 
+To override the priority, use the appropriate [revset function](#functions). For
+example, to resolve `abc` as a commit ID even if there happens to be a bookmark
+by the same name, use `commit_id(abc)`. This is particularly useful in scripts.
+
 ## Operators
 
 The following operators are supported. `x` and `y` below can be any revset, not


docs/revsets.md
```diff
@@ -53,6 +53,10 @@ Jujutsu attempts to resolve a symbol in the following order:
 3. Git ref
 4. Commit ID or change ID
 
+To override the priority, use the appropriate [revset function](#functions). For
+example, to resolve `xyz` as a change ID even if there happens to be a bookmark
+by the same name, use `change_id(xyz)`. This is particularly useful in scripts.
```
ilyagr: (Non-blocking) If we really want to substitute for #6948, I'd prefer mentioning the `commit_id` function explicitly, so that people can figure out the purpose of `commit_id` by searching for it.

For example, we could change the first sentence:

> To override the priority, use a [revset function] such as `tag`, `change_id`, `commit_id`, etc. For example, ...

Another option would be to keep the first sentence but use a commit id as the example. I'm OK with that, but I'm guessing you decided against it.

If neither of these options look OK to both you and Yuya (and a third option doesn't come to mind), we can skip this.

ilyagr:
LGTM, but I'll let Yuya review. (Of course, others are welcome to take a look as well)

docs/revsets.md
martinvonz: > Another option would be to keep the first sentence but use a commit id as the example. I'm OK with that, but I'm guessing you decided against it.

No, I just picked `change_id()` quite arbitrarily :) Well, I guess I picked it because it was what made me write this PR. But I didn't consider which was better between that and `commit_id()` anyway. Changed to `commit_id()` now.

ilyagr: Thanks!

Nit: now the commit/PR description doesn't match the content :)

martinvonz: Oops, fixed

yuja (approved):
LG, thanks1

Issue #6979, opened 2025-07-14
FR: `.parents()` method for the `Operation` template language construct


ilyagr:
**Is your feature request related to a problem? Please describe.**
`jjui` can show the operation log, and it'd be nice if the "jump to parent" command worked in the operation log. However, currently, I'm not aware of a way to implement this.

https://github.com/idursun/jjui/issues/176#issuecomment-3067418703

**Describe the solution you'd like**

The [`Operation` object](https://jj-vcs.github.io/jj/latest/templates/#operation-type) could have a `.parents()` method of type `List<Operation>`, similar to the [`Commit` object](https://jj-vcs.github.io/jj/latest/templates/#commit-type)/

**Describe alternatives you've considered**
`jj show abc- -T id` could work even if `abc` has multiple parents.

This would require `jj show` supporting `-T` and working with multiple operations.

Alternatively, `jj op log --op abc- -T id` could work. This would require `jj op log --op`.

**Additional context**

The best workaround I can imagine at the moment would be to run `jj op show abc-`, and then extract the id either from the output or from the hint that goes with the error message.

ilyagr:
Implemented in #7005. Thank you!

PR #7005, opened 2025-07-17
Add parents() method for Operation template type

steadmon:
Add a parents method for Operation, as requested in #6979 

# Checklist

If applicable:

### DIFF
--- a/cli/src/operation_templater.rs
+++ b/cli/src/operation_templater.rs
@@ -126,6 +126,11 @@ impl TemplateLanguage<'static> for OperationTemplateLanguage {
                 let build = template_parser::lookup_method(type_name, table, function)?;
                 build(self, diagnostics, build_ctx, property, function)
             }
+            OperationTemplatePropertyKind::OperationList(property) => {
+                let table = &self.build_fn_table.operation_list_methods;
+                let build = template_parser::lookup_method(type_name, table, function)?;
+                build(self, diagnostics, build_ctx, property, function)
+            }
         }
     }
 }
@@ -139,6 +144,7 @@ impl OperationTemplateLanguage {
 pub enum OperationTemplatePropertyKind {
     Core(CoreTemplatePropertyKind<'static>),
     Operation(BoxedTemplateProperty<'static, Operation>),
+    OperationList(BoxedTemplateProperty<'static, Vec<Operation>>),
     OperationId(BoxedTemplateProperty<'static, OperationId>),
 }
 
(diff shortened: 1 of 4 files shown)


docs/templates.md
```diff
@@ -307,11 +307,11 @@ This type cannot be printed. The following methods are defined.
 * `.current_operation() -> Boolean`
 * `.description() -> String`
 * `.id() -> OperationId`
+* `.root() -> Boolean`: True if the operation is the root operation.
+* `.snapshot() -> Boolean`: True if the operation is a snapshot operation.
 * `.tags() -> String`
 * `.time() -> TimestampRange`
 * `.user() -> String`
-* `.snapshot() -> Boolean`: True if the operation is a snapshot operation.
-* `.root() -> Boolean`: True if the operation is the root operation.
```
yuja, line 314: nit: I don't have a strong feeling, but other functions/methods lists aren't sorted, but roughly grouped by similarity.

cli/src/operation_templater.rs
```diff
@@ -140,12 +145,14 @@ pub enum OperationTemplatePropertyKind {
     Core(CoreTemplatePropertyKind<'static>),
     Operation(BoxedTemplateProperty<'static, Operation>),
     OperationId(BoxedTemplateProperty<'static, OperationId>),
+    OperationList(BoxedTemplateProperty<'static, Vec<Operation>>),
```
yuja, line 147: nit: can you move `OperationList` next to `Operation`? In commit templater, generic types are defined in that order.

cli/tests/test_operations.rs
```diff
@@ -2710,6 +2710,96 @@ fn test_op_show_patch() {
...
+    │  P: 1 000000000000
+    ○  000000000000
+       P: 0
+    [EOF]
+    ------- stderr -------
+    Concurrent modification detected, resolving automatically.
+    [EOF]
+    "###);
+
+    // List<Operation> can be filtered
```
yuja: I don't think we need to test every list method here. Perhaps, `test_log_parents()` in test_commit_template.rs has these tests because it was the first implementation of `List` type.

cli/src/operation_templater.rs
steadmon: Done

cli/tests/test_operations.rs
steadmon: Ack, fixed.

docs/templates.md
steadmon: I'm open to argument, but I think it's more useful as a reference to have things sorted.

martinvonz: I don't know which I prefer, but if we're going to change it, I think we should do so consistently across all types, and also including the global functions. Try that see what it looks like? Might be better as a separate PR so it doesn't hold your second commit hostage, but you can decide that.

steadmon: Ack, will drop it from this PR.

PR #7007, opened 2025-07-17
Improve discoverability and documentation of `jj file track`

salty-horse:
1. `snapshot.auto-track` is described in several places, with its ability to disable auto-tracking of files. Next to it, `jj file untrack` is mentioned.

   That sounded cool to me, but I couldn't figure out how to track files. One could make the connection between `untrack` and `track`, but I think it should be more explicit.

    (BTW, the FAQ item where the "don't track anything" option is mentioned is discussing scratch files, but I think this option is enticing enough to people familiar with other VCS tools to deserve its own entry)

2. The documentation of `jj file track` starts with:
    > New files in the working copy can be automatically tracked.

    That "can" feels wrong to me - new files *are* automatically tracked by default. I changed the phrasing a bit, to move the default behavior to the top. Now the description says "default" twice, which isn't great, so please suggest improvements (or insist on keeping the old description).

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1562,6 +1562,9 @@ get automatically tracked when they're added to the working copy. See the
 [fileset documentation](filesets.md) for the syntax. Files with paths matching
 [ignore files](working-copy.md#ignored-files) are never tracked automatically.
 
+If you set `snapshot.auto-track` to a non-default value, untracked files can be
+tracked with `jj file track`.
+
 You can use `jj file untrack` to untrack a file while keeping it in the working
 copy. However, first [ignore](working-copy.md#ignored-files) them or remove them
 from the `snapshot.auto-track` patterns; otherwise they will be immediately
--- a/docs/working-copy.md
+++ b/docs/working-copy.md
@@ -22,6 +22,9 @@ tracked when they're added to the working copy. See the
 [fileset documentation](filesets.md) for the syntax. Files with paths matching
 [ignore files](#ignored-files) are never tracked automatically.
 
+If you set `snapshot.auto-track` to a non-default value, untracked files can be
+tracked with `jj file track`.
+
 You can use `jj file untrack` to untrack a file while keeping it in the working
 copy. However, first [ignore](#ignored-files) them or remove them from the
 `snapshot.auto-track` patterns; otherwise they will be immediately tracked again.
--- a/cli/src/commands/file/track.rs
+++ b/cli/src/commands/file/track.rs
@@ -31,11 +31,12 @@ use crate::ui::Ui;
 ///
 /// Without arguments, all paths that are not ignored will be tracked.
 ///
-/// New files in the working copy can be automatically tracked.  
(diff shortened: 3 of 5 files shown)


docs/config.md
```diff
@@ -1562,6 +1562,9 @@ get automatically tracked when they're added to the working copy. See the
 [fileset documentation](filesets.md) for the syntax. Files with paths matching
 [ignore files](working-copy.md#ignored-files) are never tracked automatically.
 
+If you set `snapshot.auto-track` to a non-default value, untracked files can be
```
ilyagr, line 1565: (Optional) I think the amount of duplication of information is a bit unfortunate here. One way I can think of to fix it would be:

- Move the information about the ignore files and the link to the fileset docs to `jj help file track`.
- Don't add anything here, but replace the last two senteces of the above paragraph with:

> ... when they're added to the working copy. See \[`jj help file track`\] for more details.

(With a link to the CLI reference).

----------

If that's simpler, I think it would also be fine to submit this as is and maybe discuss the change I suggested here in a separate PR (that either of us could make). If it were code, it often makes sense to separate adding a feature from refactoring.

Yet another option would be to add a TODO in an HTML-style comment in this PR.

docs/working-copy.md
```diff
@@ -22,6 +22,9 @@ tracked when they're added to the working copy. See the
 [fileset documentation](filesets.md) for the syntax. Files with paths matching
 [ignore files](#ignored-files) are never tracked automatically.
 
+If you set `snapshot.auto-track` to a non-default value, untracked files can be
```
ilyagr, line 25: Same as my other comment.

ilyagr (commented):
This looks very helpful to me. Thanks!

I'd like @martinvonz or @yuja to review this for correctness. It looks correct to me, but I haven't experimented with `jj file track` much.

(Or, if they are busy with other things, I'll probably approve this in a few days myself; please remind me if I forget)

docs/config.md
salty-horse: If I understand correctly, you want ``jj help file track` to document how ignore files work? Feels too specific for me, as it's not related to how that command *works*, and more to when it's *useful*.

If there's one place everything should point to, I think it should be the Configuration entry for `snapshot.auto-track`. So it should describe how to use `file track` and `file untrack`.

I'm OK with committing the FAQ change separately from the others (or not at all), and discuss about how to refactor the rest of the stuff in a separate issue.

yuja (approved):

docs/config.md
ilyagr: Yeah, we can do that separately. I think it's fine to merge this as is, and then rearrange things later.

ilyagr (approved):

docs/config.md
ilyagr: And you might be right about another place being better as the main source of the information. I tried to suggest the quickest fix, but a different one might be ultimately better.

PR #7008, opened 2025-07-17
absorb: complete only changed files

senekor:

### DIFF
--- a/cli/src/commands/absorb.rs
+++ b/cli/src/commands/absorb.rs
@@ -60,7 +60,11 @@ pub(crate) struct AbsorbArgs {
     )]
     into: Vec<RevisionArg>,
     /// Move only changes to these paths (instead of all paths)
-    #[arg(value_name = "FILESETS", value_hint = clap::ValueHint::AnyPath)]
+    #[arg(
+        value_name = "FILESETS",
+        value_hint = clap::ValueHint::AnyPath,
+        add = ArgValueCompleter::new(complete::modified_from_files),
+    )]
     paths: Vec<String>,
 }
 
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -821,6 +821,12 @@ pub fn modified_range_files(current: &std::ffi::OsStr) -> Vec<CompletionCandidat
     }
 }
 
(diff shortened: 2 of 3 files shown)


cli/src/complete.rs
ilyagr, line 1: This seems like a good idea, and I think it's probably correct, but could you add a test in `test_completions`?

Say, you could set up

```
B
|
| @
|/
A
```

and test completion for `jj absorb` and for `jj absorb --from B`.

cli/src/commands/absorb.rs
ilyagr, line 1: Minor: let's change the commit/PR description to something like "absorb: complete only files changed in the `--from` revision"?

Or, if this feels too long for the first line, you could mention it in the body of the description.

I think it's useful to clarify that these are *not* the files in the diff between (say) the `--from` revision and `@`.

senekor: I like the suggested commit message, but shortened it to `absorb: complete files changed in --from revision` to stay under 50 characters.

I added a comment clarifying the behavior to the completion function.

cli/src/complete.rs
senekor: done 👍

cli/src/commands/absorb.rs
```diff
@@ -60,7 +60,7 @@ pub(crate) struct AbsorbArgs {
     )]
     into: Vec<RevisionArg>,
     /// Move only changes to these paths (instead of all paths)
-    #[arg(value_name = "FILESETS", value_hint = clap::ValueHint::AnyPath)]
+    #[arg(value_name = "FILESETS", add = ArgValueCompleter::new(complete::modified_from_files))]
```
yuja: Can you leave `value_hint` for non-dynamic completion?

senekor: good catch 👍 I had falsely assumed they would clash somehow.

cli/src/commands/absorb.rs
PhilipMetzger: > I like the suggested commit message, but shortened it to `absorb: complete files changed in --from revision` to stay under 50 characters.

The project doesn't follow the Git rules for descriptions, so no need to shorten it.

yuja (approved):

PR #7009, opened 2025-07-18
merge tools: allow specifying that a tool does not support diff editing or viewing

ilyagr:
If `diff-args` or `edit-args` are unset, default values are used.

This PR makes it so that explicitly setting them to `[]` causes them
to error. The last commit includes many examples where this is useful.

https://github.com/jj-vcs/jj/issues/4549

The first few commits are related touch-ups that could be independently
useful.

# Checklist

If applicable:

- n/a (?) I have updated the config schema (`cli/src/config-schema.json`)

### DIFF
--- a/cli/src/config/merge_tools.toml
+++ b/cli/src/config/merge_tools.toml
@@ -7,16 +7,17 @@ program="diffedit3"
 # We use a random port as a fallback if all 5 of the preferred port numbers are
 # busy.
 edit-args = ["$left", "$right", "$output", "--port", "17376-17380", "--port", "0"]
+# Not the best diff viewer, but functional. It's possible to hide the 3rd pane in the UI.
+diff-args = ["$left", "$right", "--port", "17376-17380", "--port", "0"]
 
 [merge-tools.diffedit3-ssh]
 program="diffedit3"
-# 17376 is a verified random number, as in https://xkcd.com/221/ :). I am trying
-# to avoid 8000 or 8080 in case those, more commonly used, port numbers are used
-# for something else.
+# See comments for `diffedit3` above.
 #
-# We do NOT use a random port as a fallback since we recommend that the user
-# configure SSH to forward these 5 ports
+# Unlike `diffedit3`, `diffedit3-ssh` does NOT use a random port as a fallback
+# since we recommend that the user configure SSH to forward these 5 ports
 edit-args = ["$left", "$right", "$output", "--port", "17376-17380", "--no-browser"]
+diff-args = ["$left", "$right", "--port", "17376-17380", "--no-browser"]
 
 [merge-tools.difft]
 diff-args = ["--color=always", "$left", "$right"]
@@ -70,10 +71,19 @@ merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
 merge-tool-edits-conflict-markers = true
 # VS Code prefers Git-style conflict markers
 conflict-marker-style = "git"
+# Not recommended, but functional. Without `--wait`, the behavior may be more
+# convenient, but is also more confusing and brittle (all files get deleted).
+# TODO: We could consider a "file-by-file-keep-files" invocation mode to make it
+# work better.
+diff-args=["--diff", "$left", "$right", "--wait"]
+diff-invocation-mode="file-by-file"
 
 # free/libre distribution of vscode, functionally more or less the same
 [merge-tools.vscodium]
 program = "codium"
 merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
+# See notes for vscode above
 merge-tool-edits-conflict-markers = true
 conflict-marker-style = "git"
+diff-args=["--diff", "$left", "$right", "--wait"]
+diff-invocation-mode="file-by-file"
--- a/cli/src/config/merge_tools.toml
+++ b/cli/src/config/merge_tools.toml
@@ -21,6 +21,7 @@ diff-args = ["$left", "$right", "--port", "17376-17380", "--no-browser"]
 
 [merge-tools.difft]
 diff-args = ["--color=always", "$left", "$right"]
+edit-args = []  # Non-interactive diff formatter
 
 [merge-tools.kdiff3]
 # --merge to open output pane, CreateBakFiles=0 to not include backup files in commit
@@ -33,6 +34,7 @@ merge-args = ["$left", "$base", "$right", "-o", "$output", "--auto-merge"]
 
 [merge-tools.meld-3]
 program="meld"
+diff-args = ["$left", "$right"]  # Same as the default value
 # If using this as a template, note that `$output` is repeated twice below
 edit-args = ["$left", "$output", "$right", "-o", "$output"]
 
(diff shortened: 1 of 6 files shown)


cli/src/merge_tools/mod.rs
```diff
@@ -179,10 +181,18 @@ impl DiffEditTool {
         settings: &UserSettings,
         name: &str,
     ) -> Result<Option<Self>, MergeToolConfigError> {
-        match name {
-            BUILTIN_EDITOR_NAME => Ok(Some(DiffEditTool::Builtin)),
-            _ => Ok(get_external_tool_config(settings, name)?.map(DiffEditTool::external)),
+        if let BUILTIN_EDITOR_NAME = name {
```
yuja: nit: `if name == BUILTIN_EDITOR_NAME`

cli/src/merge_tools/mod.rs
```diff
@@ -179,10 +181,18 @@ impl DiffEditTool {
...
+            return Ok(Some(DiffEditTool::Builtin));
+        }
+        let maybe_tool = get_external_tool_config(settings, name)?;
+        if let Some(tool) = &maybe_tool {
+            if tool.edit_args.is_empty() {
+                return Err(MergeToolConfigError::EditArgsNotConfigured {
+                    tool_name: name.to_string(),
+                });
+            }
         }
```
yuja: I think it's better to check bad configuration by `DiffEditor::new_inner()` as we do for `MergeEditor`.

docs/config.md
```diff
@@ -929,6 +929,8 @@ edit-args = ["--newtab", "$left", "$right"]
 
 - If no `edit-args` are specified, `["$left", "$right"]` are set by default.
 
+- If `edit-args = []`, `jj` will refuse to use this tool for diff editing. This is a way to explicitly state that a certain tool (e.g. `mergiraf`) does not work for diff editing.
```
yuja, line 936: nit: wrap this line

cli/src/diff_util.rs
```diff
@@ -336,6 +336,11 @@ fn diff_formats_from_args(
         } else {
             let tool = merge_tools::get_external_tool_config(settings, name)?
                 .unwrap_or_else(|| ExternalMergeTool::with_program(name));
+            if tool.diff_args.is_empty() {
+                return Err(cli_error(format!(
+                    "The tool `{name}` cannot be used for diff formatting"
+                )));
+            }
```
yuja: We have one more caller of `get_external_tool_config()` in diff_util.rs.

cli/src/config/merge_tools.toml
```diff
@@ -71,10 +82,16 @@ merge-args = ["--wait", "--merge", "$left", "$right", "$base", "$output"]
 merge-tool-edits-conflict-markers = true
 # VS Code prefers Git-style conflict markers
 conflict-marker-style = "git"
+# `code --diff` does not support dir-diff
```
yuja, line 91: Just question. Does `vscode` work with `diff-invocation-mode = file-by-file`? (It would be practically bad to spawn GUI for each file, though.)

yuja (approved):
Thanks.

cli/src/diff_util.rs
ilyagr: Thanks for catching this!

I'll probably create a helper function, unless you tell me you prefer to duplicate the check.

cli/src/merge_tools/mod.rs
ilyagr: Ah, I didn't look carefully enough, and didn't previously figure out the `&args` as `impl ToString` trick for passing the name to `new_inner`.

My instinct would be to move both checks to the constructors of MergeTool and DiffEditTool, to avoid this, as well as the somewhat confusing wrapping-and-then-unwrapping-again. But I don't mind switching it to the `impl ToString` trick if you prefer that.

cli/src/config/merge_tools.toml
ilyagr: Good catch! It's not super-convenient, but it does work.

There's a weird trade-off: with the "--wait" argument, it's inconvenient. Without it, it's more convenient, but VS Code shows all files as deleted. I'll do with `--wait`, as that's the more predictable behavior.

cli/src/merge_tools/mod.rs
ilyagr: In particular, AFAICT there is currently no case when `&args` is passed to `MergeEditor::new_inner` as `name`, and the `name` is actually used. So, the logic is currently more confusing than necessary.

OTOH, if you foresee this changing and there being other uses for this `name`, it might be worth keeping.

cli/src/diff_util.rs
ilyagr: The helper function is done. I can still inline it if you like.

cli/src/diff_util.rs
```diff
@@ -346,19 +361,19 @@ fn diff_formats_from_args(
...
                 name: "ui.diff-formatter".to_owned(),
                 error: err.into(),
                 source_path: None,
             })?
-            .to_format(settings, args)
+            .to_format(settings, args)?)
     } else {
         let tool = if let Some(name) = tool_args.as_str() {
-            merge_tools::get_external_tool_config(settings, name)?
+            diff_formatter_tool(settings, name)?
```
ilyagr, line 376: I could see whether it makes sense for the new error from `diff_formatter_tool` to be a `ConfigGetError`....

ilyagr: No, it wouldn't, we'd have to wrap it in another type.

cli/src/merge_tools/mod.rs
yuja: Maybe we could move handling of `::with_program(name)` to `::get_tool_config()`, but I have no idea if it's better to add separate check for `ui.merge-editor = ["command-name-without-args"]`. Beware that merge editors don't have a predefined `merge_args`.

`DiffEditor` doesn't have this problem, but I would just copy the structure of `MergeEditor`.

ilyagr: I don't quite follow, I wasn't thinking about moving handling of `::with_program`. But I don't mind moving the check to `new_inner`. Perhaps it'll make the code easier to refactor later.

ilyagr:
I'll probably merge this in an hour or so, unless someone tells me to wait.

cli/src/merge_tools/mod.rs
yuja: I mean `ExternalMergeTool` is constructed in various ways, and some/all of them should be checked if `args` is empty or not. `::get_tool_config()` is just one example.

ilyagr: First of all, are you happy with the current state of the PR?

------

I think I see what you're saying, what I said in https://github.com/jj-vcs/jj/pull/7009#discussion_r2217034130 is not right in case the user has `ui.merge-tool=[]`. We could maybe address it by adding a check for empty arguments to `MergeTool::external` and removing the `unwrap_or_else` from `MergeEditor::with_name`, since I think that codepath never results in a valid tool.

One of us could try to refactor it later, or not.

yuja: The one I last checked looked good to me, thanks.

> in case the user has `ui.merge-tool=[]`.

An empty list isn't allowed as `CommandNameAndArgs`, so `[]` should be rejected earlier.

> We could maybe address it by adding a check for empty arguments to MergeTool::external and removing the unwrap_or_else from MergeEditor::with_name, since I think that codepath never results in a valid tool.

That could be an option. I personally wouldn't refactor in that way, though, because I don't want to move the knowledge about `ExternalMergeTool::default()` to caller. The current implementation should work if we add/remove default `*_args`.

ilyagr: OK, I won't think about it too hard then.

PR #7013, opened 2025-07-18
str_util: add matches_bytes(), use it in diff_contains() revset

yuja:
# Checklist

If applicable:

### DIFF
--- a/lib/src/str_util.rs
+++ b/lib/src/str_util.rs
@@ -102,9 +102,9 @@ pub enum StringPattern {
     /// Matches with a case‐insensitive Unix‐style shell wildcard pattern.
     GlobI(Box<GlobPattern>),
     /// Matches substrings with a regular expression.
-    Regex(regex::Regex),
+    Regex(regex::bytes::Regex),
     /// Matches substrings with a case‐insensitive regular expression.
-    RegexI(regex::Regex),
+    RegexI(regex::bytes::Regex),
 }
 
 impl StringPattern {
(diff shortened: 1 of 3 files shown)


lib/tests/test_revset.rs
```diff
@@ -4162,6 +4163,30 @@ fn test_evaluate_expression_diff_contains() {
     );
 }
 
+#[test]
+fn test_evaluate_expression_diff_contains_non_utf8() {
```
ilyagr, line 4167: (Optional) I got a bit confused here. Perhaps it'd help to replace the note that "haystack is usually a string" with something like the following to the commit description (if it's correct):

> We don't have a way (a good way?) to search for non-UTF-8 strings with `diff_contains`, but this allows the user to search for Unicode patterns
inside not-necessarily-Unicode files.

I think that if we want to allow searching for non-Unicode haystacks, we need to start supporting some sort of `\x81` syntax to express them in Unicode.

I guess another way is to rely on TOML's support for such syntax, but that would mean that our config values are no longer String-s (or are they not already)? On Windows, I'm guessing the command line is UTF-16 or something.

ilyagr (approved):

lib/tests/test_revset.rs
ilyagr: Something similar could also be added to the test as a comment.

yuja: Updated the commit description. This function merely test that a non-utf-8 line isn't silently ignored, so I didn't add detailed explanation about unicode/non-unicode patterns.

You can use bytes regex to match `\x80`, etc. btw.

lib/src/str_util.rs
```diff
@@ -244,11 +245,16 @@ impl StringPattern {
...
+    /// Returns true if this pattern matches the `haystack` string.
     ///
     /// When matching against a case‐insensitive pattern, only ASCII case
     /// differences are currently folded. This may change in the future.
     pub fn matches(&self, haystack: &str) -> bool {
+        self.matches_bytes(haystack.as_bytes())
+    }
+
+    /// Returns true if this pattern matches the `haystack` bytes.
+    pub fn matches_bytes(&self, haystack: &[u8]) -> bool {
```
lf-: both should be renamed to `is_match` or similar at some point soon since "matches" is either "the matches" or "it matches". I do that in https://github.com/jj-vcs/jj/pull/6899/commits/2c99222a0699726d10425cf80a3fba44954da2ce, but I can split it out and land early to make it easier for this one? Or we can land this and I can fix it there as well. Either way is fine.

yuja: Yeah, if you split it to new PR, I'll rebase on top.

I don't feel strongly about the rename, btw. Both `matches() -> bool` and `is_match() -> bool` sounds good.

lf-: https://github.com/jj-vcs/jj/pull/7049 done, let's just land that one in main and save ourselves some conflicts

PR #7016, opened 2025-07-18
backend: assert that tree entries are sorted also in release build

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/backend.rs
+++ b/lib/src/backend.rs
@@ -407,7 +407,7 @@ pub struct Tree {
 
 impl Tree {
     pub fn from_sorted_entries(entries: Vec<(RepoPathComponentBuf, TreeValue)>) -> Self {
-        debug_assert!(entries.is_sorted_by(|(a, _), (b, _)| a < b));
+        assert!(entries.is_sorted_by(|(a, _), (b, _)| a < b));
         Tree { entries }
     }
 


yuja:
I don't have a strong feeling, but we tend to use `debug_assert!()` for non-trivial assertions, and I think this assertion belongs to that.

Maybe we can instead add `from_entries()` that does sorting internally?

FWIW, gix has a similar assertion when writing a Tree object. Since it is a debug assertion, we could write a bad Git tree object if we didn't sort the entries in Git order.

martinvonz:
> I don't have a strong feeling, but we tend to use `debug_assert!()` for non-trivial assertions, and I think this assertion belongs to that.

Since all entries had already been processed by the backend in order to produce them, it's already linear in the size, so I think this check is no worse than having an `assert!(path > previous_path);` inside a loop.

> Maybe we can instead add `from_entries()` that does sorting internally?

That means that callers have to make sure they use the right method. That may seem trivial, but I missed it in the Google code because we have two backends (one for our monorepo and one for our Git repos) and they share a Tree protobuf type that uses different sort order for the two backends. So I would prefer the non-debug assertion. I doubt it will have measurable performance impact but I can time it if you think might be significant.

yuja:
> > I don't have a strong feeling, but we tend to use `debug_assert!()` for non-trivial assertions, and I think this assertion belongs to that.
> 
> Since all entries had already been processed by the backend in order to produce them, it's already linear in the size, so I think this check is no worse than having an `assert!(path > previous_path);` inside a loop.

Yes, and we tend to use `debug_assert` for liner case afaik.

> > Maybe we can instead add `from_entries()` that does sorting internally?
> 
> That means that callers have to make sure they use the right method. That may seem trivial, but I missed it in the Google code because we have two backends (one for our monorepo and one for our Git repos) and they share a Tree protobuf type that uses different sort order for the two backends. So I would prefer the non-debug assertion. I doubt it will have measurable performance impact but I can time it if you think might be significant.

iirc, `entries.is_sorted_by_key(|(name, _)| name)` in `GitBackend::read_tree()` had some measurable (but not significant) cost when indexing changed paths.

Adding non-debug assertion means that the minimal cost of `from_sorted_entries()` and `from_entries()` are roughly the same, so I think we can remove `from_sorted_entries()`.

martinvonz:
> Yes, and we tend to use `debug_assert` for liner case afaik.

My argument was that the linear time is effectively a constant time if we already do other linear processing. We do have constant-time non-debug assertions already, including in loops, I think, but it's possible that we don't have them in as hot loops as these (where the Git backend is basically just hashing the contents and writing it to disk).

> iirc, `entries.is_sorted_by_key(|(name, _)| name)` in `GitBackend::read_tree()` had some measurable (but not significant) cost when indexing changed paths.

I wouldn't be surprised if `entries.sort_by_key()` takes measurable time, but I'm surprised that even checking if they are ordered takes measurable time.

Anyway, I've fixed the bug at Google, and it's probably going to take a while before we see other backends that might have similar bugs. It sounds like you care somewhat strongly so I'll drop this PR.

yuja:
> > iirc, `entries.is_sorted_by_key(|(name, _)| name)` in `GitBackend::read_tree()` had some measurable (but not significant) cost when indexing changed paths.
> 
> I wouldn't be surprised if `entries.sort_by_key()` takes measurable time, but I'm surprised that even checking if they are ordered takes measurable time.

Checking had measurable cost iirc, but was less significant than unconditional `entries.sort_by_key()`.

Anyway, I would add `::from_entries()` and remove `::from_sorted_entries()` because `::from_sorted_entries()` + non-debug assertion wouldn't be any faster than `::from_entries()`.

PR #3015, opened 2024-02-10
add support for `jj help alias`

jyn514:
fixes https://github.com/martinvonz/jj/issues/2866. built on https://github.com/martinvonz/jj/pull/3006 for convenience, but i can separate it out if desired.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [x] I have updated the documentation (README.md, docs/, demos/)
- [x] I have updated the config schema (cli/src/config-schema.json)
- [x] I have added tests to cover my changes

PR #7018, opened 2025-07-18
cli help: document the `ci` alias

ilyagr:
Follows up on 59de83a.

Now, all of the aliases from `cli/src/config/misc.toml` will be documented in the same way. I think I was previously unsure whether we'll keep them, but they seem to be doing no harm.

### DIFF
--- a/cli/src/config/misc.toml
+++ b/cli/src/config/misc.toml
@@ -1,6 +1,8 @@
 # The code assumes that this table exists, so don't delete it even if you remove
 # all aliases from here.
 [aliases]
+# The aliases here should be kept in sync with the `[default alias: blah]` notes
+# in the first line of the relevant commands' help text.
 b = ["bookmark"]
 ci = ["commit"]
 desc = ["describe"]
--- a/cli/src/commands/commit.rs
+++ b/cli/src/commands/commit.rs
@@ -31,7 +31,7 @@ use crate::description_util::join_message_paragraphs;
 use crate::text_util::parse_author;
 use crate::ui::Ui;
 
-/// Update the description and create a new change on top.
+/// Update the description and create a new change on top [default alias: ci]
 ///
 /// This command is very similar to `jj split`. Differences include:
 ///
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -127,7 +127,7 @@ To get started, see the tutorial [`jj help -k tutorial`].
 * `abandon` — Abandon a revision
 * `absorb` — Move changes from a revision into the stack of mutable revisions
 * `bookmark` — Manage bookmarks [default alias: b]
-* `commit` — Update the description and create a new change on top
+* `commit` — Update the description and create a new change on top [default alias: ci]
 * `config` — Manage config options
 * `describe` — Update the change description or other metadata [default alias: desc]
 * `diff` — Compare file contents between two revisions
@@ -520,7 +520,7 @@ If you want to forget a local bookmark while also untracking the corresponding r
 
 ## `jj commit`
 
-Update the description and create a new change on top.
+Update the description and create a new change on top [default alias: ci]
 
 This command is very similar to `jj split`. Differences include:
 


PhilipMetzger:
ideally we should generate this somehow, since I expect the remaining aliases to move the config soonish.

cli/src/commands/squash.rs
```diff
@@ -41,7 +41,7 @@ use crate::description_util::join_message_paragraphs;
 use crate::description_util::try_combine_messages;
 use crate::ui::Ui;
 
-/// Move changes from a revision into another revision
+/// Move changes from a revision into another revision [default alias: amend]
```
martinvonz: I'm less sure about this one since the alias exists just to ease the transition from `hg amend` (I don't think there's a `git amend`). Thoughts?

ilyagr:
Good point! For now, I'll add poor man's automation, namely a comment :)

https://old.reddit.com/r/pics/comments/hr5zp/im_a_sign_not_a_cop/

>  the remaining aliases to move the config soonish

What's remaining? I haven't looked carefully.

PhilipMetzger:
There are multiple subcommands which have a single letter aliases builtin e.g `bookmark create (b c)` which should be moved into the config when something like https://github.com/jj-vcs/jj/pull/6104#issuecomment-2742415565 is implemented. It'll also require #3015 in the future.

cli/src/commands/squash.rs
ilyagr: I wouldn't mind removing it. We could put the workaround in the changelog.

I have no way to judge how popular it is without removing it. I stopped using it a long time ago.

Another option, if we want, would be to treat it similarly to `jj clone`:

```console
$  JJ_CONFIG= jj clone
error: unrecognized subcommand 'clone'

Hint: You probably want `jj git clone`. See also `jj help git`.
Hint: You can configure `aliases.clone = ["git", "clone"]` if you want `jj clone` to work and always use the Git backend.
```

https://github.com/jj-vcs/jj/pull/5845/commits/1ef18ffbae5e08c4767bfd429bc76c22ef2f1d67

ilyagr:
These currently get shown by `clap` in `jj help bookmark`, they are official clap aliases.

```console
$ jj help bookmark
Manage bookmarks [default alias: b]

See [`jj help -k bookmarks`] for more information.

[`jj help -k bookmarks`]: https://jj-vcs.github.io/jj/latest/bookmarks

Usage: jj-darwin-aarch64 bookmark [OPTIONS] <COMMAND>

Commands:
  create   Create a new bookmark [aliases: c]
  delete   Delete an existing bookmark and propagate the deletion to remotes on the next push [aliases: d]
  forget   Forget a bookmark without marking it as a deletion to be pushed [aliases: f]
  list     List bookmarks and their targets [aliases: l]
  move     Move existing bookmarks to target revision [aliases: m]
  rename   Rename `old` bookmark name to `new` bookmark name [aliases: r]
  set      Create or update a bookmark to point to a certain commit [aliases: s]
  track    Start tracking given remote bookmarks [aliases: t]
```

The reason `b` is not an official clap alias is to allow people to override its meaning. It seems less important to allow people to override the meaning of `c` in `jj b c`.

PhilipMetzger:
> These currently get shown by `clap` in `jj help bookmark`, they are official clap aliases.

Yes, that's why I mentioned them as builtin.

Anyway I think it should be a goal to have the subcommand aliases in the config since that would align with supporting configuration profiles (https://github.com/jj-vcs/jj/pull/5538#issuecomment-2629969079). But this is just good have, and as you said there are now no remaining core aliases 🥳 .

cli/src/commands/squash.rs
ilyagr: It would look like this: https://github.com/jj-vcs/jj/pull/7019.

ilyagr: It's probably easiest to keep this PR open until we make a decision on #7019. I'll mark it as draft.

ilyagr: It's gone now.

cli/src/config/misc.toml
```diff
@@ -1,3 +1,6 @@
+# The aliases here should be kept in sync with the `[default alias: blah]` notes
+# in the first line of the relevant commands' help text.
```
martinvonz: nit: i might put this inside the table to reduce the risk of the reader missing it (it might look like just a file-level comment describing the contents of this file)

ilyagr: Yeah, I was wondering about this too. I'll put the new part of the comment inside; maybe then it'll be clearer that there are two different issues the comments are about.

PR #7019, opened 2025-07-18
cli: remove `amend` alias

ilyagr:
This also adds a helpful message to the error if the user does `jj amend`. I don't think this is absolutely necessary, but I also don't see any harm in it.

------

We do use the term "the commit gets amended" in the docs a few times, but there's no mention of `jj amend` except in old changelog entries, I think.

I'm not sure this needs a test. It looks like this:

```console
$ cargo run -- amend --help
error: unrecognized subcommand 'amend'

Hint: You probably want `jj squash`. You can configure `aliases.amend = ["squash"]` if you want `jj amend` to work.
```

# Checklist

- n/a? I have updated the documentation (`README.md`, `docs/`, `demos/`)
- [?] I have added/updated tests to cover my changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 * `commit.working_copies()` template method now returns `List<WorkspaceRef>`
 
+* The previously predefined `amend` alias has been removed. You can restore it
+  by setting the config `aliases.amend = ["squash"]`.
+
 ### Deprecations
 
 * The `all:` revset modifier and `ui.always-allow-large-revsets` setting is
--- a/cli/src/cli_util.rs
+++ b/cli/src/cli_util.rs
@@ -3990,23 +3990,31 @@ impl<'a> CliRunner<'a> {
 
 fn map_clap_cli_error(err: clap::Error, ui: &Ui, config: &StackedConfig) -> CommandError {
     if let Some(ContextValue::String(cmd)) = err.get(ContextKind::InvalidSubcommand) {
+        let remove_useless_error_context = |mut err: clap::Error| {
+            // Clap suggests unhelpful subcommands, e.g. `config` for `clone`.
+            // We don't want suggestions when we know this isn't a misspelling.
+            err.remove(ContextKind::SuggestedSubcommand);
+            err.remove(ContextKind::Suggested); // Remove an empty line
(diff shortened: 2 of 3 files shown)


martinvonz (approved):
Thanks. Let's give others a chance to comment since it's breaking change and possibly controversial.

3geek14:
I'm an amend user (probably because it's what I got used to with Google's Mercurial before I leaned about jj). I tried switching to squash yesterday, and I was surprised to notice that `sq` felt awkward to type on my keyboard (qwerty).

I think it makes sense to bring the docs and the built-in aliases into alignment, either by removing amend or by adding amend to the docs. I recently asked a question on Discord where I said I used `jj amend`, and this caused friction when someone reading the question didn't know about that alias. Reducing that friction feels more valuable to me than catering to my muscle memory. I don't have much preference for which alignment strategy is followed. If my left hand really doesn't want to write about gourds, I can always make an alias, and I appreciate the help text you added.

apoelstra:
As another user voice, personally I dislike all the aliases that attempt to mimick other VCSes. In part because they're an extra thing that new users might encounter but old users might not know (although this is also true of e.g. `git restore`/`git switch` which I think were create UX choices, so by itself this isn't reason to drop stuff).

I've never used `amend` (and didn't know about it) but I have used `commit` which serves a similar purpose, so I can talk about that.

The big thing, for me, is that these aliases are (almost by definition) leaky abstractions, because the jj model is different from the git model or the hg model. So as a user you feel a need to know what's "really" happening, i.e. what the alias maps to, and then you feel that you'd be better served by a line in the "git vs jj commands" table than an alias, which would provide more context about how the models are different. `commit` in particular was quite disorienting, because by running `jj describe && jj new` it operates on two commits in one shot; coming from git it feels like the commands are in the wrong order; and if I were to attempt to transliterate it back to git I'd get something like `git commit --amend && git commit -a -m ""`. (I know this is not correct, but this is how I recall my early-user experience around this alias, and once I understood jj well enough to get it, I just never used it again.)

For some projects (e.g. my nixos config) where I use jj almost exclusively as history log with no branches and no rebasing, I find myself doing `jj describe && jj new` a fair bit, but I'm still hesitant to use `jj commit` because commit feels like "a git verb" and the jj meaning is just different enough to give me pause. Although perhaps if I did use it, I'd quickly get used to it and there'd be no problem.

PR #7021, opened 2025-07-19
cli completion: remove non-functional diff editors/formatters

ilyagr:
Follows up on  https://github.com/jj-vcs/jj/pull/7009

# Checklist

If applicable:

- n/a I have updated `CHANGELOG.md`
- n/a I have updated the documentation (`README.md`, `docs/`, `demos/`)

### DIFF
--- a/cli/src/complete.rs
+++ b/cli/src/complete.rs
@@ -38,6 +38,7 @@ use crate::config::ConfigEnv;
 use crate::config::CONFIG_SCHEMA;
 use crate::merge_tools::configured_merge_tools;
 use crate::merge_tools::get_external_tool_config;
+use crate::merge_tools::ExternalMergeTool;
 use crate::revset_util::load_revset_aliases;
 use crate::ui::Ui;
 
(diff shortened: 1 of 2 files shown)


cli/src/complete.rs
```diff
@@ -469,33 +469,39 @@ pub fn merge_editors() -> Vec<CompletionCandidate> {
...
-            .chain(configured_merge_tools(settings))
+            .chain(configured_merge_tools(settings).filter(|name| {
+                let Ok(Some(tool)) = get_external_tool_config(settings, name) else {
+                    return false;
+                };
+                // The args are empty only if `edit-args` are explicitly set to
+                // `[]` in TOML. If they are not specified, the default
+                // `["$left", "$right"]` value would be used.
+                !tool.edit_args.is_empty()
+            }))
```
yuja: nit: maybe extract helper function as we have three callers?

ilyagr: You mean a function that takes `Settings` and `Fn(MergeTool) -> bool`, and returns a bunch of names? That would probably work.

yuja: Yes, something like `merge_tools_filtered_by(settings, |tool| !tool.edit_args.is_empty())`

ilyagr: Yes, that works well, thanks for the suggestion (and for the review)!

PR #7022, opened 2025-07-19
revset_engine: evaluate `X` as predicate in `reachable(X, Y)`

scott2000:
From Discord discussion: https://discord.com/channels/968932220549103686/968932220549103689/1396106445153374249

# Checklist

If applicable:

### DIFF
--- a/lib/src/default_index/revset_engine.rs
+++ b/lib/src/default_index/revset_engine.rs
@@ -919,6 +919,9 @@ impl EvaluationContext<'_> {
                         }
                     }
                 }
+                // `UnionFind::find` is somewhat slow, so it's faster to only do this once and
+                // then cache the result.
+                let domain_reps = domain_vec.iter().map(|&pos| sets.find(pos)).collect_vec();
 
                 // Identify disjoint sets reachable from sources. Using a predicate here can be
                 // significantly faster for cases like `reachable(filter, X)`, since the filter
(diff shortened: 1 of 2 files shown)


lib/src/default_index/revset_engine.rs
```diff
@@ -916,15 +916,21 @@ impl EvaluationContext<'_> {
...
-                .attach(index)
-                .map_ok(|pos| sets.find(pos))
-                .try_collect()?;
+                // Identify disjoint sets reachable from sources. Using a predicate here can be
+                // significantly faster for cases like `reachable(filter, X)`, since the filter
+                // can be checked for only commits in `X` instead of for all visible commits,
+                // and the difference is usually negligible for non-filter revsets.
+                let sources_revset = self.evaluate(sources)?;
+                let mut sources_predicate = sources_revset.to_predicate_fn();
+                let set_reps: HashSet<_> = domain_vec
```
martinvonz: It seems like it should be possible to iterate over the sets and then include the whole set as soon as we find a single matching in the set, right? But it looks like UnionFind doesn't have an API for that yet. What you have looks like an improvement to me either way. I'll let Yuya approve since he knows the revset engine better.

scott2000: I think it's a bit difficult to iterate over the sets actually, because we have to evaluate the predicate in order by reverse index position (since I believe that is a requirement for some predicates), and we also have to return the final positions in reverse index order as well.

Instead, I think we can just iterate over `domain_vec` still, but skip evaluating the predicate whenever we already know that that set is included. I added another commit to this PR to try implementing that. @yuja do you think this works, or do you have any better ideas for how to implement this?

lib/src/default_index/revset_engine.rs
```diff
@@ -927,15 +930,21 @@ impl EvaluationContext<'_> {
...
+                        set_reps.insert(rep);
                     }
                 }
 
                 let positions = domain_vec
                     .into_iter()
-                    .filter(|pos| set_reps.contains(&sets.find(*pos)))
+                    .zip(domain_reps)
+                    .filter(|(_, rep)| set_reps.contains(rep))
+                    .map(|(domain, _)| domain)
```
martinvonz: nit: maybe `position` is a better variable name so it doesn't sound like it's the whole domain

martinvonz (commented):
This looks good to me, thanks, but let's wait for Yuya's review, so I'll let him approve.

lib/src/default_index/revset_engine.rs
yuja: > we have to evaluate the predicate in order by reverse index position

Can you add a test that would otherwise break? I think some ancestry query should do the job.

PR #7037, opened 2025-07-21
cli: describe how `commit` is different from `split`

martinvonz:
# Checklist

If applicable:


cli/src/commands/commit.rs
```diff
@@ -32,6 +32,18 @@ use crate::text_util::parse_author;
 use crate::ui::Ui;
 
 /// Update the description and create a new change on top.
+///
+/// This command is very similar to `jj split`. Differences include:
+///
+/// * `jj commit` always acts on the working-copy commit (@).
+///
+/// * `jj commit` is not interactive by default (it selects all changes).
```
ilyagr, line 38: (Optional) I'd put this first. That is, I'd swap it with the "working-copy commit" item.

cli/src/commands/commit.rs
```diff
@@ -32,6 +32,18 @@ use crate::text_util::parse_author;
 use crate::ui::Ui;
 
 /// Update the description and create a new change on top.
+///
+/// This command is very similar to `jj split`. Differences include:
+///
+/// * `jj commit` always acts on the working-copy commit (@).
```
ilyagr: (Optional): I'd reword this as something like:

> `jj commit` does not have a `-r` option. It always acts on the working copy commit (@).

Otherwise, some people might not realize that we're talking about `split -r`.

cli/src/commands/commit.rs
martinvonz: I went back and forth on that. Since you prefer the other order, I've swapped them.

cli/src/commands/commit.rs
martinvonz: Makes sense. Done.

PR #7040, opened 2025-07-22
docs: Move language references to a Reference section

jennings:
This PR contains just the first commit from !5375, which moves the docs for revset, fileset, and template languages to a section called "Reference".

I neglected that PR for so long that I worry I might accidentally lose documentation changes that have been made in the past six months if I rebase it. So, I'd rather PR each commit individually so I can pay more attention to each one.

# Checklist

If applicable:


yuja (commented):
Looks good, but I'll wait for approval from @ilyagr.

mkdocs.yml
```diff
@@ -135,9 +135,6 @@ nav: # This lists all the files that become part of the documentation
 
   - Configuration:
```
thoughtpolice: Maybe it would be better to just outright rename this section `Reference` instead. After all, `config.md` is the reference page for the configuration settings, and the words `Configuration` and `Settings` are otherwise the same, so why nest one under the other?

I basically approve of this change though because the word `Reference` is, I think, much clearer.

jennings: No, see, if you really think about it, they're actually subtly different and you can't just... Yeah, no, that makes way more sense.

ilyagr:
Sorry, I didn't notice you were waiting on me for a bit until much too late. This looks good to me; Reference is a good name.

PR #7050, opened 2025-07-22
template_builder: stop eating non-TemplateError error sources in tests

lf-:
I was under the impression that the error sources were not possible to actually *see* so I was disinclined to use with_source as I couldn't confirm it worked in my tests, but it turns out that was because of an unexpectedly incomplete printer in the testing code! (.origin() does not include third party error origins)

Pulled out of https://github.com/jj-vcs/jj/pull/6899

# Checklist

If applicable:


yuja (approved):
Thanks!

PR #7055, opened 2025-07-23
index: use bit set to deduplicate/remove transitive edges

yuja:
# Checklist

If applicable:


lib/src/default_index/bit_set.rs
```diff
@@ -17,6 +17,72 @@ use std::cmp::max;
...
+    ///
+    /// Panics if the `pos` exceeds the capacity.
+    #[cfg_attr(not(test), expect(dead_code))] // TODO
+    pub fn get(&self, pos: GlobalCommitPosition) -> bool {
+        self.get_bit(self.to_bitset_pos(pos))
+    }
+
+    fn get_bit(&self, (bitset_pos, bit_pos): (u32, u32)) -> bool {
+        let bit = 1_u64 << bit_pos;
+        self.data[usize::try_from(bitset_pos).unwrap()] & bit != 0
```
martinvonz, line 58: Can this be written as `self.get_set_bit((bitset_pos, bit_pos));` without performance impact? I would hope that the compiler would inline and simplify it but I don't know much about compilers. It's clearly not very complicated as is anyway, so I'm mostly just curious. If it does get optimized well, then we can also simplify the public interface to just have `get_set()`. Not a big deal either way if you don't feel like testing the performance impact, or if you prefer not to rely on what gets optimized out.

martinvonz (approved):
Nice improvement!

lib/src/default_index/bit_set.rs
yuja: No, `.get_set_bit()` needs `&mut self`. Maybe we can instead merge `.get_set*()` and `.set*()` (or delete `.get_set*()` at all), but I didn't feel strongly and just duplicated the code.

martinvonz: Ah, of course. Yeah, I don't feel strongly either.

Issue #175, opened 2022-03-27
[Ergonomy] Support for git-like rerere


LifeIsStrange:
https://delicious-insights.com/en/posts/git-rerere/
Git rerere is probably the best git feature for productivity and I wonder if JJ already support such feature, or otherwise wether it's on the roadmap.
Also given that you have interop with git repositories I wonder if you could leverage previous git records to apply after the repository has migrated from git to jj

arxanas:
I suspect rerere wouldn't be necessary when using jj change-ids. You wouldn't have to resolve conflicts again because the logical change has been tracked and you can skip applying it again.

I suspect rerere might be less useful in a primarily rebase- rather than merge-oriented workflow? I never have to use rerere, I think because I always rebase my changes instead of merging them.

LifeIsStrange:
> I suspect rerere might be less useful in a primarily rebase- rather than merge-oriented workflow?
I don't know for rebase but from experience it is very useful for merge oriented worflows.

Look I haven't used JJ yet so about
> I suspect rerere wouldn't be necessary when using jj change-ids. You wouldn't have to resolve conflicts again because the logical change has been tracked and you can skip applying it again.
I have to trust you on that one, the day I'll decide to experiment with JJ I could share wether it would benefit or not from rerere.

martinvonz:
> I don't know for rebase but from experience it is very useful for merge oriented worflows. I did had many redundant conflicts in the past using rebase too but it's possible I was using git rebase in a bad way since the ergonomy was so awful it made me slighly question my usage.

I have also never really used rerere, but I know reasonably well how it works. There are a few cases where it can be useful. I think jj provides good solutions in the cases I can think of at least. It's the first-class support for conflicts that helps.

## Use case: "test merge"

Let's start with the scenario from `git help rerere`. They have this history:

```
           o---*---o topic
          /
 o---o---o---*---o---o master
```

The use case they describe there is to merge `master` into `topic` as a "test merge", to see what conflicts there are. The idea is that you'll continue to work on `topic` afterwards. So they do that test merge and resolve the conflicts, and then undo the merge (using `git reset --hard HEAD^1`, presumably), because they want to keep history free from "unnecessary" merge commits, i.e. they want to avoid two merge commits like this:

```
           o---*---o---+---o---o topic
          /           /         \
 o---o---o---*---o---o---o---o---+ master
```

[...]

PR #7057, opened 2025-07-23
Add support for rerere

Mic92:
Probably the best way to approach this PR is by discussing the high-level design first (see markdown file, [rendered](https://github.com/jj-vcs/jj/blob/e2a4f020b856fa70ac9ad3b0bac33d23d381a527/docs/design/rerere.md)) and than I can update the code accordingly

# Checklist

If applicable:


PhilipMetzger:
I don't think this will happen, see https://github.com/jj-vcs/jj/issues/175.

martinvonz:
I appreciate the PR but I would like to see that there's enough demand for this feature to justify the added code. I understand that it can be useful. I don't think I've heard many requests at all for the feature. Maybe people who have wanted it can let us know here.

Mic92:
In git for me it's mainly a time saver when I mess up a rebase, because I can replay stuff and git will propose me the same resolutions, I already did. But no worries, I can also just stay with git. I am missing other features anyhow i.e. fetching/pushing pull requests and a robust alternative to lazygit (especially patch editing is pretty good).

martinvonz:
> In git for me it's mainly a time saver when I mess up a rebase, because I can replay stuff and git will propose me the same resolutions, I already did.

Can you elaborate on your workflow here? Do you mean that you rebase and resolve some conflicts and then decide to `jj op restore` back to before you resolved the conflicts?

Mic92:
> > In git for me it's mainly a time saver when I mess up a rebase, because I can replay stuff and git will propose me the same resolutions, I already did.
> 
> Can you elaborate on your workflow here? Do you mean that you rebase and resolve some conflicts and then decide to `jj op restore` back to before you resolved the conflicts?

Yes something like that. In git it's `git rebase --abort`. It's just one of these features that are nice because they don't require any mental overhead and just work in the background.

martinvonz:
> Yes something like that. In git it's `git rebase --abort`.

In that case, we have alternative solutions to consider, depending on what your use case is. 

For me, IIRC (it's been a while), I typically ran `git rebase --abort` or `hg rebase --abort` because I didn't feel like resolving all those conflicts at that point in time. In such cases, when using jj, I would simply leave the conflict later in the series for later.

From your design doc:

> When the same logical conflict appears in different contexts (e.g., cherry-picking changes or merging through different paths), users must resolve it manually each time.

For this use case, you should be able cherry-pick the conflict resolution as well. It may not be obvious but that actually works. For example, let's say you have a diamond-shaped graph like this:

```
  D
 / \
B   C
 \ /
  A
```

Let's say there's a conflict between B and C, which is resolved in D. You can now cherry-pick D with `jj duplicate -r D -d B -d C` and it will include the conflict resolution. The parents don't have to be exactly the same as they were in this simple example.

There are also the case of "test merges" and rebasing internal merges. I explained those in #175 (which Philip linked to above).

Mic92:
> Let's say there's a conflict between B and C, which is resolved in D. You can now cherry-pick D with `jj duplicate -r D -d B -d C` and it will include the conflict resolution. The parents don't have to be exactly the same as they were in this simple example.

So this requires me having to manually lookup those commits, right? Some automation would be nice nevertheless also for this alternative.

martinvonz:
Yes, that is correct.

Mic92:
Bye

zoriya:
> I don't think I've heard many requests at all for the feature. Maybe people who have wanted it can let us know here.

Hey, I don't know if this use case was forgotten or if i'm missing something obvious but one use case of rerere I have that JJ doesn't seem to cover is:

```
 B
 |
 A
 |
master
```

You finish work on branch A and merge it (via github's ui in my case). You then rebase branch B on master and need to fix again the conflicts you already fixed when you rebase branch A on master before merging it.

martinvonz:
I don't really follow the scenario. Are you saying your resolved conflicts in GitHub's UI when merging it? But that doesn't populate your (local) rerere cache, so that's probably not what you meant. I don't have another theory.

PR #7058, opened 2025-07-23
git: track pseudo refs as bookmarks.

orthros:
Updates version of gix to 0.73.0

When the refs are diffed for import, we now check the pseudo refs reported by `gix`, ignoring the `HEAD` pseduo ref, and add them to the diff.

This will allow pseudo refs like `FETCH_HEAD` and `JIRI_HEAD` to be tracked as a bookmark in `jj`

# Checklist

If applicable:

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -250,6 +250,14 @@ pub fn parse_git_ref(full_name: &GitRefName) -> Option<(GitRefKind, RemoteRefSym
         let name = RefName::new(name);
         let remote = REMOTE_NAME_FOR_LOCAL_GIT_REPO;
         Some((GitRefKind::Tag, RemoteRefSymbol { name, remote }))
+    } else if full_name.as_str().ends_with("HEAD") {
+        // HEAD we skip
+        if full_name == "HEAD" {
+            return None;
+        }
+        let name = RefName::new(full_name.as_str());
+        let remote = REMOTE_NAME_FOR_LOCAL_GIT_REPO;
+        Some((GitRefKind::Bookmark, RemoteRefSymbol { name, remote }))
     } else {
         None
     }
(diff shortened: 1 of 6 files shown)


lib/src/git.rs
```diff
@@ -250,6 +250,14 @@ pub fn parse_git_ref(full_name: &GitRefName) -> Option<(GitRefKind, RemoteRefSym
...
         let remote = REMOTE_NAME_FOR_LOCAL_GIT_REPO;
         Some((GitRefKind::Tag, RemoteRefSymbol { name, remote }))
+    } else if full_name.as_str().ends_with("HEAD") {
+        // HEAD we skip
+        if full_name == "HEAD" {
+            return None;
+        }
+        let name = RefName::new(full_name.as_str());
+        let remote = REMOTE_NAME_FOR_LOCAL_GIT_REPO;
+        Some((GitRefKind::Bookmark, RemoteRefSymbol { name, remote }))
```
yuja, line 260: Pseudo refs shouldn't be mapped to bookmarks because they aren't local branches in Git. Moved `FETCH_HEAD` doesn't mean the old `FETCH_HEAD` branch has been rewritten to the new `FETCH_HEAD`, for example.

Maybe it's okay to import them as anonymous heads (just like we do for the `HEAD`), but I'm not sure. We might need a config knob. Another option is to add `jj git import <GIT-NAME>` option or a similar command to explicitly import unnamed commits.

orthros: Personally, having them be automatic (e.g. anonymous heads like `HEAD`) would be far more conducive to my (and my team)'s workflows. Its not a lot to run `jj git import <GIT-NAME>`, but its an extra command we'd run every time we run another command in our workflow (which we run multiple times a day).

yuja: My concern about automatically importing random heads like `FETCH_HEAD` is that the ref can be quite old. The user or IDE might ran `git fetch`, and he might not remember about that when `jj` imported the `FETCH_HEAD`. I'm not strongly against adding automatic imports, but I think it's better to add an opt-in config like `git.auto-import-refs = ["glob:*HEAD"]`.

Issue #4388, opened 2024-09-04
FR: Pushing and fetching arbitrary refs


ilyagr:
**Is your feature request related to a problem? Please describe.**
Currently, there is no way to use `jj` to push or fetch an arbitrary ref from the remote. This causes problems for e.g. Gerrit as [discussed on Discord](https://discord.com/channels/968932220549103686/1281020382249357423/1281020382249357423).

On one hand, this reinforces the desire for https://github.com/martinvonz/jj/issues/4387, but separately it would be nice to support something for the general case.

**Describe the solution you'd like**
I can think of 2 possible solutions:

1. Re-design our branches (soon to be called bookmarks) to support tracking an arbitrary ref on a remote. This probably would imply having to support tracking different refs on different remotes, tracking remote branches of a different name, and perhaps tracking multiple refs per remote.
2. Set up simple commands to push and fetch arbitrary refs unsafely, either permanently or until we can implement 1 or think of a better solution.

Here's a possible interface for option 2:

- `jj git fetch-ref [--remote REMOTE] REF_NAME [branch_name]` would fetch an arbitrary ref and optionally create a branch at the newly fetched commit for easier reference later. This new local branch wouldn't track anything. It'd just be a bookmark :).
- `jj git push-ref <--force|--expected-at COMMIT_ID> [--remote REMOTE] REVISION REF_NAME`

Both 1 and 2 have disadvantages. 1 adds complexity to the UI and the model. I'm not sure whether or not 2. is better than telling people to co-locate the repo and use `git fetch/git push`.

In any case, I'm curious what people think should be done here.

yuja:
Some previous discussions:

> 1. Re-design our branches (soon to be called bookmarks) to support tracking an arbitrary ref on a remote.

https://github.com/martinvonz/jj/issues/3235

> 2. Set up simple commands to push and fetch arbitrary refs

https://github.com/martinvonz/jj/pull/2098#issuecomment-1682522251

For 2, we might want to introduce command namespace for plumbing commands so `jj git fet[TAB]` wouldn't stop at `jj git fetch(-refs)?`.

emilazy:
This is also nice for GitHub (refs like `pull/123/head`).

ilyagr:
Thanks, I forgot about those. 

For a subcommand, `jj git ref fetch` could work.

Minion3665:
Heya! I was the one who originally prompted the discussion on Discord - I have a few thoughts to put here too :)

[...]

PR #7061, opened 2025-07-24
docs: Add guide for handling divergent changes

jennings:
Divergent changes are so frequently asked that I think it deserves more explanation than a FAQ entry can give it.

I think making a new "Guides" section and putting this there makes us better able to explain what's going on.

# Checklist

If applicable:


docs/guides/divergence.md
```diff
@@ -0,0 +1,89 @@
+# Handling divergent changes
+
+## What are divergent changes?
+
+A [divergent change][glossary_divergent_change] occurs when multiple visible
+commits have the same change ID.
```
zx8: Add a link to https://jj-vcs.github.io/jj/prerelease/glossary/#visible-commits?

When I was new to `jj` (coming from `git`), I found it tricky to navigate/understand the docs because there was a bunch of new terminology that seemed to be assumed knowledge, so it'd be handy to linkify these sorts of things.

docs/guides/divergence.md
```diff
@@ -0,0 +1,89 @@
...
+becomes hidden and the new commit (the "successor") is visible. Thus, only one
+commit with a given change ID is visible at a time.
+
+But, a hidden commit can become visible again. This can happen if:
+
+- A visible descendant is added. For example, `jj new REV` will make `REV`
+  visible even if it was hidden before.
+
+- It is made the working copy. `jj edit REV` will make `REV` visible if it
+  wasn't already.
```
apoelstra: In 97a57fe7027480f3898fec4e65ebee74bac71bc7:

We should add the case where a new ref is imported from git, e.g. by the user running `git fetch` or otherwise directly manipulating the git repo. For me this is the most common source of divergent change IDs, and it's especially confusing because "I didn't do anything, I just `git fetch`ed!", which in pure git is one of the least destructive things you can do.

apoelstra: (See https://github.com/jj-vcs/jj/issues/4388 in case you are wondering why somebody would use `git fetch`. This is a very common workflow with Github and Gitlab, which provide non-branch refs for every PR, saving you the work of adding every contributor's fork as a remote.)

docs/guides/divergence.md
```diff
@@ -0,0 +1,89 @@
...
+commits have the same change ID.
+
+Normally, when commits are rewritten, the original version (the "predecessor")
+becomes hidden and the new commit (the "successor") is visible. Thus, only one
+commit with a given change ID is visible at a time.
+
+But, a hidden commit can become visible again. This can happen if:
+
+- A visible descendant is added. For example, `jj new REV` will make `REV`
+  visible even if it was hidden before.
```
PhilipMetzger: nit: if you want to be complete here, the `bookmark` commands also should be mentioned

docs/guides/divergence.md
```diff
@@ -0,0 +1,89 @@
...
+
+- It is made the working copy. `jj edit REV` will make `REV` visible if it
+  wasn't already.
+
+Divergent changes also occur if two different users or processes amend the same
+change, creating two visible successors. This can happen when:
+
+- Multiple people edit the same change simultaneously in different repositories.
```
PhilipMetzger: nit: expand this to mention to mention a coworker working on the same branch and then re-pushing, since many don't understand that as a rewrite

PhilipMetzger (approved):
lgtm, mostly left some minor nits with discussion points

docs/FAQ.md
```diff
@@ -540,15 +540,6 @@ with something like `jj desc -m "Revert the merge of D into B`. Now, commit `@`
 undoes the merge of `D` into  `B`. If necessary, you can now rebase it
 elsewhere, e.g. `jj rebase -r @ -d main`.
 
-### How do I deal with divergent changes ('??' after the [change ID])?
```
joyously, line 543: I think you should leave the FAQ entry, and link to the new doc for more details.

jennings: I originally did that, but then realized we keep old versions of the documentation published.

Removing this question will only affect `/latest/` and `/prerelease/`.

If someone needs a permanent link to documentation, I think they really should be linking to a specific version.

It also seemed untenable to keep old FAQ questions around forever for this purpose.

What do others think?

joyously: > Removing this question will only affect `/latest/` and `/prerelease/`.

No, the point is that someone reading the FAQ will find the divergent answer, and then follow the link to the new doc to get all the details. Two ways to discover the answer are better than one. Just to note, I'm not concerned about versions of the docs, just that the info is discoverable.

PhilipMetzger: I think @ilyagr's opinion on that should matter. I have no preference here.

docs/guides/divergence.md
jennings: How does this sound?

> - A visible descendant is added locally. For example, `jj new REV` will make `REV` visible even if it was hidden before.
> 
> - A visible descendant is fetched from a remote. If the hidden commit was pushed to a remote, others may base new commits off of them. When their new commits are fetched, their visibility makes the hidden commit visible again.

docs/guides/divergence.md
jennings: Are you referring to a coworker appending a new commit to your branch (as opposed to amending a commit)? Do you think [this suggestion would address that](https://github.com/jj-vcs/jj/pull/7061#discussion_r2229505692) too?

docs/FAQ.md
ilyagr: I think people often link to the `/latest` version of the docs. for example, most links inside `jj`'s source code are to `/latest`.

Joy's point also seems reasonable to me; this would make the article more discoverable, and I don't see much harm in the FAQ question remaining.

So, I'd keep the question, even if it becomes a stub.

docs/guides/divergence.md
PhilipMetzger: I mean this case here: https://github.com/jj-vcs/jj/discussions/7004

docs/guides/divergence.md
apoelstra: This sounds great to me! I didn't realize that this was the specific mechanism by which my fetches were causing me grief.

I kinda think that `jj` itself should help with this. by noticing when previously-hidden commits become visible again. (And distinguishing between commits that were hidden manually with `abandon` and those that were hidden implicitly by just updating a change). Though this is out of scope for a docs change, of course.

docs/guides/divergence.md
jennings: How does this sound if I replace all the bullet points with this?

> - Another author modifies commits in a branch that you have also modified
>   locally.
> 
> - You perform operations on the same change from different workspaces of the
>   same repository.
> 
> - Two programs modify the repository at the same time. For example, you run
>   `jj describe` and, while writing your commit description, an IDE integration
>   fetches and rebases the branch you're working on.

docs/guides/divergence.md
```diff
@@ -0,0 +1,89 @@
...
+
+If one of the divergent commits is clearly obsolete or incorrect, simply abandon
+it:
+
+```shell
+# First, identify the divergent commits
+jj log -r 'divergent()'
+
+# Abandon the unwanted commit using its commit ID
+jj abandon <unwanted-commit-id>
```
apoelstra, line 60: In 97a57fe7027480f3898fec4e65ebee74bac71bc7:

I think this instance of `jj abandon` (the one right after `jj log -r 'divertgent()'`, since Github display doesn't have enough context) should illustrate that you can provide multiple commit IDs. If you abandon them one at a time, descendants (which you intend to abandon anyway) get rebased and chenge their commit IDs from under you, and get conflicts.

apoelstra: The one below (in "duplicate and abandon") I think should also do this: it should probably illustrate duplicating commits one by one, for simplicity's sake, but then abandon them all at once.

docs/guides/divergence.md
PhilipMetzger: that should be good enough

docs/guides/divergence.md
apoelstra: BTW, I don't feel strongly about this -- if you're unsure at all you can just ignore me. This PR is super valuable and I don't mean to hang it up or waste mental energy on minutae.

docs/guides/divergence.md
jennings: added!

docs/guides/divergence.md
jennings: Updated!

docs/FAQ.md
jennings: I put back the question as a stub that links to the guide.

docs/guides/divergence.md
jennings: Added a couple examples: `xyz wxy vwx` and `xyz::`.

docs/guides/divergence.md
```diff
@@ -0,0 +1,104 @@
...
+commit will be abandoned.
+
+### Strategy 4: Ignore the divergence
+
+Divergence isn't an error. If the divergence doesn't cause immediate problems,
+you can leave it as-is. If both commits are part of immutable history, this may
+be your only option.
+
+However, it can be inconvenient since you cannot refer to divergent changes
+unambiguously using their change ID.
```
offbyone, line 104: It might be worth this:

```suggestion
However, it can be inconvenient since you cannot refer to divergent changes
unambiguously using their change ID. It can also be annoying in your log display,
as those divergent changes will be flagged in a way that looks like an error.
```

jennings: I think I'd rather leave that out, because we're actually in discussions about changing that.

Rather than document "this is annoying", I'd prefer to make it not-annoying :)

PR #7071, opened 2025-07-24
cli: print warning about `git clean -fdx` when creating colocated repo

martinvonz:
We hear every now and then from users who have run `git clean -fdx` and accidentally delete their repo. Let's at least warn users about it when the create a colocated repo.

We could perhaps allow the `.jj/` directory to be inside a `.git/` directory like Sapling does, but that seems a little ugly and it's a larger change.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/git/clone.rs
+++ b/cli/src/commands/git/clone.rs
@@ -162,6 +162,14 @@ pub fn cmd_git_clone(
             tx.finish(ui, "check out git remote's default branch")?;
         }
     }
+
+    if args.colocate {
+        writeln!(
+            ui.hint_default(),
+            r"Running `git clean -xdf` will remove `.jj/`!",
+        )?;
+    }
+
     Ok(())
 }
 
--- a/cli/src/commands/git/init.rs
+++ b/cli/src/commands/git/init.rs
@@ -112,6 +112,12 @@ pub fn cmd_git_init(
         r#"Initialized repo in "{}""#,
         relative_wc_path.display()
     )?;
+    if args.colocate {
+        writeln!(
+            ui.hint_default(),
+            r"Running `git clean -xdf` will remove `.jj/`!",
+        )?;
+    }
 
     Ok(())
 }
--- a/cli/tests/test_git_clone.rs
+++ b/cli/tests/test_git_clone.rs
@@ -197,6 +197,7 @@ fn test_git_clone_colocate() {
     ------- stderr -------
     Fetching into new repo in "$TEST_ENV/empty"
     Nothing changed.
+    Hint: Running `git clean -xdf` will remove `.jj/`!
     [EOF]
     "#);
 
@@ -220,6 +221,7 @@ fn test_git_clone_colocate() {
     Working copy  (@) now at: uuqppmxq 3711b3b5 (empty) (no description set)
     Parent commit (@-)      : qomsplrm ebeb70d8 main | message
     Added 1 files, modified 0 files, removed 0 files
+    Hint: Running `git clean -xdf` will remove `.jj/`!
     [EOF]
     "#);
     let clone_dir = test_env.work_dir("clone");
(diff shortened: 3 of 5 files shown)


cli/src/commands/git/clone.rs
```diff
@@ -144,6 +144,12 @@ pub fn cmd_git_clone(
             .ok();
         }
     }
+    if args.colocate {
+        writeln!(
+            ui.warning_default(),
+            r#"Running `git clean -xdf` will remove `.jj/`!"#,
+        )?;
+    }
```
yuja: nit: print at the end of the command because the warning isn't useful if `clone` failed.

cli/src/commands/git/clone.rs
```diff
@@ -144,6 +144,12 @@ pub fn cmd_git_clone(
             .ok();
         }
     }
+    if args.colocate {
+        writeln!(
+            ui.warning_default(),
```
yuja: nit: maybe `ui.hint_default()` is better? This only matters for people who casually run `git clean -f`, and we might want to add `hints.git-clean-colocated=false` to suppress it.

cli/src/commands/git/clone.rs
```diff
@@ -144,6 +144,12 @@ pub fn cmd_git_clone(
...
+            r#"Running `git clean -xdf` will remove `.jj/`!"#,
```
yuja: nit: `r#` can be removed

yuja (approved):

cli/src/commands/git/clone.rs
martinvonz: This is already after the code that removes the directory if the clone failed

cli/src/commands/git/clone.rs
martinvonz: Done

cli/src/commands/git/clone.rs
martinvonz: Hint makes sense to make it less annoying for people who clone lots of repos. I think changing it to warning and opt-out could also make sense. I suggest we go with the hint for now.

zx8: I am a sample size of 1, but I frequently clone repos and will find this annoying for sure. Would it be possible to implement the opt-out configuration?

martinvonz: Even it's just a hint (cyan)? Do you find that much more annoying than the other ~10 lines we print?

zx8: I'll live with it - I just thought the whole idea of hints was to be able to opt-out of them if you so wish.

cli/src/commands/git/clone.rs
yuja: Beware that `clone_result` isn't unwrapped yet. I also think it's better to print the hint after the checkout information because it is a hint for future user action.

cli/src/commands/git/clone.rs
martinvonz: I don't think we have any hints that (or at least very few) that you can opt out of so far. I can leave this PR up for a while so we can hear from others if they want a hint or a warning you can opt out from.

cli/src/commands/git/clone.rs
martinvonz: I put it before the working-copy messages on purpose but I don't feel strongly so I've moved it last.

cli/src/commands/git/clone.rs
PhilipMetzger: I think @zx8 should request the flag on #5766, since that is the issue for disabling certain hints.

martinvonz: I'll merge this now and then we have about a week to test it before next release. If people find it annoying, we can roll it back or add a config for it (and I would prefer to make it a warning if we have an opt-out anyway).

PR #7073, opened 2025-07-25
More async stuff (mostly merge-/rebase-related)

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/rewrite.rs
+++ b/lib/src/rewrite.rs
@@ -57,6 +57,7 @@ pub fn merge_commit_trees(repo: &dyn Repo, commits: &[Commit]) -> BackendResult<
         merge_commit_trees_no_resolve_without_repo(repo.store(), repo.index(), commits)
             .block_on()?
             .resolve()
+            .block_on()
     }
 }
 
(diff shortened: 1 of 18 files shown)


lib/src/rewrite.rs
```diff
@@ -252,9 +254,11 @@ impl<'repo> CommitRewriter<'repo> {
...
             )
         } else {
-            let old_base_tree = merge_commit_trees(self.mut_repo, &old_parents).await?;
-            let new_base_tree = merge_commit_trees(self.mut_repo, &new_parents).await?;
-            let old_tree = self.old_commit.tree_async().await?;
+            let old_base_tree_fut = merge_commit_trees(self.mut_repo, &old_parents);
+            let new_base_tree_fut = merge_commit_trees(self.mut_repo, &new_parents);
+            let old_tree_fut = self.old_commit.tree_async();
+            let (old_base_tree, new_base_tree, old_tree) =
+                try_join3(old_base_tree_fut, new_base_tree_fut, old_tree_fut).await?;
```
yuja: nit: we can also use `try_join!()` macro, but I don't know which is preferred.

yuja (approved):

lib/src/rewrite.rs
martinvonz: I don't know either, but it's nice that it works for a variable number of arguments. Done.

PR #7079, opened 2025-07-25
docs: Clarify "order of binding strengths" is ascending

dnnr:
# Checklist

If applicable:

### DIFF
--- a/docs/filesets.md
+++ b/docs/filesets.md
@@ -53,7 +53,7 @@ expressions.
 * `x ~ y`: Matches `x` but not `y`.
 * `x | y`: Matches either `x` or `y` (or both).
 
-(listed in order of binding strengths)
+(listed in ascending order of binding strengths)
 
 You can use parentheses to control evaluation order, such as `(x & y) | z` or
 `x & (y | z)`.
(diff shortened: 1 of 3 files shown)


InCogNiTo124 (approved):

docs/filesets.md
```diff
@@ -53,7 +53,7 @@ expressions.
 * `x ~ y`: Matches `x` but not `y`.
 * `x | y`: Matches either `x` or `y` (or both).
 
-(listed in order of binding strengths)
+(listed in ascending order of binding strengths)
```
scott2000, line 56: I think this would be descending order actually since it goes from highest binding strength (`~x`) to lowest binding strength (`x | y`)?

Also I think maybe `x ~ y` should be above `x & y` because `a & b ~ c & d` is parsed as `((a & b) ~ c) & d` I believe, which is also equivalent to `a & (b ~ c) & d`.

yuja: Correct, it's from highest (strongest) to lowest (weakest). I don't know if "descending binding strengths" is intuitive, though.

> Also I think maybe `x ~ y` should be above `x & y`

They have the equal strength. We'll need a nested list or table to describe these precisely.

PR #7080, opened 2025-07-25
cli `diff`: warn if `-r` arg does not contain `::` or `..` and expands to multiple revisions (plus tests, another fixup)

ilyagr:
I will probably extract the added tests (first commit) into a separate PR.

Adding warning if there are 0 revisions is a bit of an afterthought, but seems nice.

This is meant to replace #6931.

# Checklist

If applicable:

- (n/a) I have updated `CHANGELOG.md`
- (n/a) I have updated the documentation (`README.md`, `docs/`, `demos/`)
- (n/a) I have updated the config schema (`cli/src/config-schema.json`)


cli/src/commands/diff.rs
```diff
@@ -207,6 +218,28 @@ pub(crate) fn cmd_diff(
     }
 
     ui.request_pager();
+    // The warnings are printed here since they must be passed to the pager.
+    // Otherwise, warnings would be very easy to miss when the pager is used.
+    if let Some([single_revset]) = args.revisions.as_deref() {
+        if multiple_revision_revset
+            && !single_revset.as_ref().contains("::")
+            && !single_revset.as_ref().contains("..")
+        {
```
yuja: We'll need to test the parsed expression since the user might have `stack()` alias, for example. There will be other revsets that make sense in this context, so I don't think it's good idea to show a warning based on the form of the expression. I had something like `jj diff -r'@|@-'` in my shell history.

ilyagr: I don't agree. I often found this warning very useful in cases where analyzing the set of revisions would not have helped. For example, I once abandoned a commit and got a graph like this:

```
@    vswtmr ilyagr@users 2 seconds ago e797260
├─╮  (empty) (no description set)
│ │ ×  szmnmu ilyagr@users 2 seconds ago e30f975 conflict
├───╯  (no description set)
○ │  xkywux ilyagr@users 2 seconds ago git_head() fed2624
├─╯  Test
○  xpoylt ilyagr@users 8 seconds ago b3254bd
│  pass through generated files ....
◆  uuprvm yacintmimi@g 1 month ago master 0332da0
│  remove `misc` section

```

Then, I did `jj diff -r @-` and got a helpful warning. OTOH, if I did ~~`jj diff uupr..@`~~ (**Update:* `jj diff -r uupr..xkyw`), it'd be the same set of revisions, but I wouldn't want a warning.

> I had something like jj diff -r'@|@-' in my shell history.

I think showing this warning as a false positive sometimes is not the end of the world, especially if we have some way to suppress it. (Currently, it would be adding `-r none()` to the command, we could also have a flag or remember the `all:` syntax).

I did change the warning to say "potentially unrelated revisions", though.



---------------------

All that said, unless the above is convincing enough to people that we can finish this up quickly, I think I might abandon this PR. I can compile it into my personal version of `jj`, and this discussion over several issues/prs has exhausted me. It's still unclear how many people would appreciate this warning, even though it continues to seem important to me (since I have a severe allergy to feeling that a tool might confuse people instead of teaching them something useful when something they haven't learned about happens).

yuja: Just to be clear, I didn't mean the warning is useless. I just meant it seemed bad idea to rely on the input unparsed/unexpanded query.

> Then, I did jj diff -r `@-` and got a helpful warning. OTOH, if I did `jj diff uupr..@`, it'd be the same set of revisions, but I wouldn't want a warning.

Maybe you mean `uupr..@-`? I think it's good to show the warning in that case, though.

```
@- = xkywux|xpoylt
uupr..@ = vswtmr|xkywux|xpoylt
```

ilyagr: > I just meant it seemed bad idea to rely on the input unparsed/unexpanded query.

What would you prefer?

> Maybe you mean uupr..@-?

I think I meant `uupr..xkyw`. Sorry, my mind is a bit fuzzy right now.

yuja: > What would you prefer?

I think the original PR that checks if the range has multiple root/head ends is better overall.

ilyagr: In my mind, there are several serious problems with that:

1. `jj diff -r @-` would quietly do the wrong thing in the example from my above comments.
2. This is wrong in the presence of merge commits as per https://github.com/jj-vcs/jj/pull/6931#discussion_r2217975635
3. Relatedly to the above, but perhaps more importantly, this suggests a false narrative/intuition about what `jj diff -r multiple_commits` means. It's complicated enough as it is without us adding to the confusion with our warnings[^explicit]. As I explained in other threads, I think always doing `jj diff -r a..b` suggests a more correct narrative as to what the meaning of this operation is. I think checking if a revset is of the form `a..b` is possible, but it's unpleasant to explain (includes a proof I never quite get around to finishing), and wouldn't solve problem 1.

[^explicit]: The confusion we'd be adding would be the suggestion that heads and roots of a revset is what's important to understand what the diff means. Heads are, roots aren't, except to the degree they may allow you to express your revset as `a..b` with `a` being the branch point.

To emphasize, teaching a false intuition is more important to me than the frequency with which 2. is actually encountered.

This is the reason it seems that we disagree and the way I feel this discussion got stuck (I said all of the above before, not sure if my previous statements were less clear, or perhaps more clear). If you agree with another maintainer about the single root/head heuristic, I guess I won't stop you, but I would not feel good about it.

martinvonz: Perhaps we should always warn if the set has more than 1 commit instead? Maybe this feature is used infrequently enough that users won't mind. Or perhaps we should deprecate the feature and add it under `jj diff --range <revset>` or something instead.

I agree with Yuya that basing the warning on the form of the input expression feels ugly. I haven't thought much about it, so maybe "ugly" is all it is and it's actually fine in practice.

yuja: I personally don't think the `diff -r@-` example above is eligible for warning. As the range is linear, it shows diffs between two existing states. To me, auto-merge of arbitrary revisions is the confusing part of `jj diff -r<multiple-revisions>`.

> Perhaps we should always warn if the set has more than 1 commit instead?

or print from/to revisions? We'll need `--no-commit-summary` or something to suppress that, though.

ilyagr: > or print from/to revisions?

I think this will work for the revsets we now support under `-r`. I don't know whether the notion of "from/to revisions" would even makes sense if we tried to support all revsets, let alone whether it would be possible to figure out what those are in every case.

Possibly related: Expanding `-r` to support all revsets would also encounter a separate problem. It's difficult to figure out what the base should be for the diff shown in some cases. I don't know whether or not there is a good choice in every case.

ilyagr: Most sections of this comment can begin with "To be clear...".

-----------

I'm generally OK with the "print from/to revisions" approach, if we're all OK with the downside I described above (I am).

Another headache I foresee is deciding when to print the revisions to `stdout` and when to print them to `stderr`. E.g., what do we do for `diff --git`? Your ideas (or my future ideas) might resolve this headache, but currently I don't find it fun to think about these decisions.

------

I'm OK with both proposals from Martin's https://github.com/jj-vcs/jj/pull/7080#discussion_r2292599400. They each have downsides; I'm OK with that.

The proposal of always printing a warning would also probably need an option to suppress that warning. While that is in some ways very similar to my original proposal from long ago of using `all:` here, I think using an option is better by being boring in a good way. It will not look out of place the way `all:` did. Also, times have changed, Yuya and others have done some work, and we don't seem to need a general mechanism for suppressing such warnings anymore.

In fact, Martin's two proposals are very similar to each other. Choosing between an option to suppress the warning and the flag `--range` that is just like `-r` but accepts multiple revisions depends, in my mind, mainly on whether people find `--range` long to type. The name `--range` is great, I can't think of a better one, but it has a downside that we can't have a short option for it easily. Something like `--allow-multi-revset-diff` could easily be `-M`.

------

>  warning on the form of the input expression feels ugly.

Yeah, I know, it does. Every option we're considering is a compromise with downsides.

In my mind, this inelegance is one of two problems with this PR's current approach, the other being Yuya's observation that "the user might have stack() alias, for example".

OTOH, this PR has been very helpful to me. I can live with those downsides. I think the second of the two problems could be managed later.

I also think this is slightly nicer than the "always print a warning when there are many revisions" by omitting the common cases where the user definitely expects multiple commits (but I recognize there are also downsides to this added bit of intelligence, and I'm OK with not having it).

BTW, for better or worse, with this PR as is, one way to suppress the warning would be to have an alias `diffmany = ["diff", "-rnone()"]`. Then, you can `jj diffmany -r whatever`. We could of course provide another option.

------

OTOH, I think that going with the single heads/roots idea would be a mistake, as explained in previous comments.

There's also the revset deconstructing idea I repeated a few times (e.g. in item 3 of https://github.com/jj-vcs/jj/pull/7080#discussion_r2292591807), but I don't like that one either. While in theory it is both correct and satisfies most (or all?) of Yuya's requirements, it is confusing and complicated, would only help me half the time, and I don't think Yuya ever liked it either.

PR #7087, opened 2025-07-26
Make use of `Self` consistent

adamnemecek:


martinvonz:
Thanks. I think I'm for this change, but let's hear what others think. If we do this, we should add it to `[workspace.lints.clippy]` so it's enforced. Also please squash all or most of your commits. See https://jj-vcs.github.io/jj/latest/contributing/

adamnemecek:
I think that the Self functionality is still in nursery so you can’t put it in the clippy file yet. Also squashed the commits.

PhilipMetzger:
can you prefix the commit description with `everything:` or `cli lib:`to adhere to our commit guidelines Martin already linked to, thanks.

martinvonz:
> I think that the Self functionality is still in nursery so you can’t put it in the clippy file yet.

How can you tell? I just added `use_self = "warn"` and ran `cargo +stable clippy` and got the warnings.

adamnemecek:
Last time I checked was a while back so things may have changed.

martinvonz:
Could you add it to the config since it does seem to be stabilized now?

adamnemecek:
Done and done.

martinvonz:
Looking at the test failures, you probably want to pass `--all-targets` to the `cargo clippy --fix` command.

adamnemecek:
Done.

thoughtpolice (approved):
I think this is a nice change for consistency (most Rust projects seem to use this style.) So I'm fine with merging it.

minor nit: you could also slightly tweak the commit message with a newline; normally we just have a one sentence description but without Title case or period; e.g. 

    cli lib: make use of `Self` consistent

and then the rest in the body. It's not a huge deal either way if you don't.

adamnemecek:
Ok fixed the title.

martinvonz:
As I said, I'm also for it. I don't think I've seen objections to previous changes that have replaced a type name by `Self` from others either, so I think you can go ahead and merge this, @adamnemecek.

adamnemecek:
Ok in the merge queue. Thanks to all!

PR #7092, opened 2025-07-27
cli lib: use the `..` syntax instead of destructuring

adamnemecek:


martinvonz:
Many (most? all?) of these use destructuring so the compiler will help us and we don't forget to update some usage sites if we add more fields to the struct. So I don't think we want this change.

adamnemecek:
Ok.

PR #7093, opened 2025-07-27
git: write empty blob when setting intent to add

### LINKED ISSUE #7062: `jj util gc` fails with internal error after a file is ignored in the colocated repo
## Description

After ignoring a previously tracked file in the colocated repo, running `jj util gc` fails. This problem isn't present in a non-colocated repo. Adding `jj new` prior to the `jj util gc` line fixes the issue (GC run finishes succesfully).

```bash
jj git init --colocate test_repo
cd test_repo
echo "hello" > hello
jj st
echo "hello" > .gitignore
jj file untrack hello
# jj new  # uncommenting this line fixes the issue
jj util gc
```

## Expected Behavior
GC run is finished succesfully (or if this isn't possible due to the internal mechanics then maybe a more descriptive error message could be shown).

## Actual Behavior
```
Enumerating objects: 10, done.
Counting objects: 100% (10/10), done.
Delta compression using up to 12 threads
Compressing objects: 100% (5/5), done.
fatal: unable to read e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
fatal: failed to run repack
Internal error: Unexpected error from backend
Caused by: git gc command exited with an error: exit status: 128
```

## Specifications

- Platform: Linux
- Version: 0.31.0

scott2000:
Fixes #7062.

# Checklist

If applicable:

### DIFF
--- a/lib/src/git.rs
+++ b/lib/src/git.rs
@@ -1345,8 +1345,7 @@ pub fn reset_head(mut_repo: &mut MutableRepo, wc_commit: &Commit) -> Result<(), 
     };
 
     let wc_tree = wc_commit.tree()?;
-    update_intent_to_add_impl(&mut index, &parent_tree, &wc_tree, git_repo.object_hash())
-        .block_on()?;
+    update_intent_to_add_impl(&git_repo, &mut index, &parent_tree, &wc_tree).block_on()?;
 
     // Match entries in the new index with entries in the old index, and copy stat
     // information if the entry didn't change.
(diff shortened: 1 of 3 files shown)


lib/src/git.rs
scott2000, line 1: Do you think it would be a good idea to include a changelog entry and/or a test for this fix?

thoughtpolice: I think both a test and a changelog entry would be nice, imo.

thoughtpolice (approved):

lib/src/git.rs
```diff
@@ -1550,19 +1549,26 @@ async fn update_intent_to_add_impl(
...
-                gix::index::entry::Mode::FILE_EXECUTABLE
-            } else {
-                gix::index::entry::Mode::FILE
-            },
-            path.as_ref(),
-        );
+    if !added_paths.is_empty() {
+        // We need to write the empty blob, otherwise `jj util gc` will report an error.
+        let empty_blob = git_repo
+            .write_blob([])
```
yuja: nit: I would use `b""`

yuja (approved):

PR #7098, opened 2025-07-27
Git Lfs

### LINKED ISSUE #80: Add support for Git LFS

gusinacio:
Fixes #80 

Thanks to @bcspragu with #6036, @weiznich #5480, and @kejadlen (where I forked the current branch).
# Checklist


gusinacio:
@weiznich is missing CLA 😔

CHANGELOG.md
PhilipMetzger, line 1: nit: squash into the first commit which implements this

lib/src/gitattributes.rs
PhilipMetzger, line 1: nit: can you add some motivation to the commit message

lib/src/lib.rs
PhilipMetzger, line 1: Optional: It may be useful to just drop this commit for the moment, since a design doc is currently being written for the correct integration of `.gitattributes` in `jj`

PhilipMetzger (commented):
some minor nits, I am also cc @06393993 since he's actively working on the Gitattributes design.

06393993:
While none of the following questions should block the merge, they are probably worth answering:

1. It seems that this change only handles the snapshot path. I am wondering if I am missing anything, or we haven't considered the update path(store -> disk) yet? If so, how are merge conflicts handled?
2. IIUC, on snapshot, all the `.gitattributes` files in the repo are parsed. It is possible to come up with a desgin that avoids reading and parsing `.gitattributes` files only relevant to the unchanged files. For example, if the current change happens only under the folder `a`, it's possible to not read and parse the `.gitattributes` files under the folder `b`. But this is probably less of an issue, because even [Firefox has only 8 `.gitignore` files](https://github.com/search?q=repo%3Amozilla-firefox%2Ffirefox++path%3A**%2F.gitattributes&type=code). However, we should document that and be clear that we are reading and parsing irrelevant `gitattributes` files.
3. [GIt doesn't follow symbolic links for `gitattributes` files](https://git-scm.com/docs/gitattributes#_notes). In this PR, IIUC, it seems that we always follows the symbolic links. We should either leave a TODO comment or an explanation if we decide to behave differently from git.
4. How do we expect the future integration with the git smudge/clean filter feature, and the EOL conversion feature?

lib/src/gitattributes.rs
gusinacio: this is a fix from the previous work. They usually calculated the folder where the .gitattributes was found but looking deeper in gix::attr, I found that it's already calculated by the lib.

lib/src/lib.rs
gusinacio: CI was breaking without this because of feature flag git. Do you want me to try to move to inside gitattributes.rs?

gusinacio:
> While none of the following questions should block the merge, they are probably worth answering:
> 
>     1. It seems that this change only handles the snapshot path. I am wondering if I am missing anything, or we haven't considered the update path(store -> disk) yet? If so, how are merge conflicts handled?

I haven't considered it. My initial goal is to ignore all git-lfs files so we can handle them properly using normal git lfs commands. Like in gitignore, if you want to start tracking using gitlfs, you need first to `jj file untrack <path>`, so hopefully no merge conflicts happen after it's untracked by jj. Although,  there's a known issue where if you move between changes with and without gitattributes, you need to do `git lfs checkout` manually.

I'm open/available to do any requested change.


>     2. IIUC, on snapshot, all the `.gitattributes` files in the repo are parsed. It is possible to come up with a desgin that avoids reading and parsing `.gitattributes` files only relevant to the unchanged files. For example, if the current change happens only under the folder `a`, it's possible to not read and parse the `.gitattributes` files under the folder `b`. But this is probably less of an issue, because even [Firefox has only 8 `.gitignore` files](https://github.com/search?q=repo%3Amozilla-firefox%2Ffirefox++path%3A**%2F.gitattributes&type=code). However, we should document that and be clear that we are reading and parsing irrelevant `gitattributes` files.

That is a great question. If we don't parse .gitiattributes, the snapshot engine will interpret that gitattributes filtered files are part of the change. It's already how we are doing for gitignore. If we can answer this question for gitattributes we can also proabably do the same for gitignore.

>     3. [GIt doesn't follow symbolic links for `gitattributes` files](https://git-scm.com/docs/gitattributes#_notes). In this PR, IIUC, it seems that we always follows the symbolic links. We should either leave a TODO comment or an explanation if we decide to behave differently from git.

Thanks for pointing this out. That part of the code was taken from `gitignore.rs` which looks like follows symlinks as well since it uses same method `path.is_file()` which follows symlinks even though [Git says it shouldn't](https://git-scm.com/docs/gitignore#_notes). Is there any issue about this? Should I update .gitattributes to not follow symlinks and let .gitignore follow it?

>     4. How do we expect the future integration with the git smudge/clean filter feature, and the EOL conversion feature?

Probably for next steps, is to add commands under `jj git lfs` that are gonna update .gitattributes directly. I also expect that we're gonna reuse `GitAttributesFile` for the diff engine, if not to ignore, at least to have a different path when a file is flagged with LFS. One thing that will change for sure is the config file, since we currently have a list of "lfs filters" to ignore, that won't be necessary. 

In any case, I haven't planned for that and I'm open to suggestions or at least make the code more prepared for future work.

lib/src/gitattributes.rs
gusinacio: I added a description, is it better to squash with the previous commit?

PhilipMetzger: no, that should be good.

06393993:
I am asking in details just because this CL is very different from [my design](https://github.com/06393993/jj/blob/gitattr-design/docs/design/gitattributes.md), that I plan to propose early this August, and am nervous about the unfortunate conflict: either my design will be rejected, or we need to revert this PR to adopt my design. If you are not hurry with this PR, I am wondering if you could read my design if you haven't already?

> My initial goal is to ignore all git-lfs files

Sorry for not being clear, but I was asking about the behavior of the `git.ignore-filters` settings on checkout, not about how one should checkout git lfs files. If we are ignoring files, IMO, we should also consider ignoring those files on the checkout path(store to disk). IIUC, your current PR only ignore those files when snapshotting(disk to store). 

For example, if one runs `jj git clone <SOURCE>` with the proper `git.ignore-filters` settings, those git LFS files will still be checked out in their original shape instead of being ignored. We probably do that on purpose, so that the following `git lfs pull` just work, but we probably need to document that somewhere.

And when it comes to implement the checkout path, merge conflicts on `gitattributes` files become an issue - and this is what I mean by merge conflicts.

Applogize again that I didn't make myself clear what I mean on my original question. I personally don't think this potentially inconsistency should block this PR, but I generally am not sure about the standard of the jj project, so just raise the question so that the core maintainers won't overlook them.

>  If we don't parse .gitiattributes, the snapshot engine will interpret that gitattributes filtered files are part of the change.

That's not necessarily the case. We can still ignore those files in `TreeState::snapshot` where we receive the changes from `deleted_files_rx` and `file_states_rx`. We can read and parse `gitattributes` files there to avoid handling irrelevant `gitattributes` files. 

> If we can answer this question for gitattributes we can also proabably do the same for gitignore.

I don't think we should do the same to `gitignore` file, because the performance cost can be expensive, while for `gitattributes` files it's a pure performance win. `gitattributes` files are different from `gitignore` files: `gitattributes` won't match folders, while `gitignore` does:

> patterns that match a directory do not recursively match paths inside that directory (so using the trailing-slash path/ syntax is pointless in an attributes file; use path/** instead)

From [the gitattributes documentation](https://git-scm.com/docs/gitattributes).

While it is also possible to read and parse `gitignore` files after we generate the list of potentially changed files and folders to save some parsing, but it is likely that if we can parse the `gitignore` file early and skip some folders when we visit the file system, we can save much more. On the contrast, `gitattributes` files can't be used to skip directories, so parsing them proactively won't save us anything(theoretically you could, via the `path/**` pattern, but ultimately `gitattributes` files are not used for ignoring files, and if we want to use it to skip folders, we need special treatments for those patterns).

For example, in a Rust project, people usually put `target` in `.gitignore`. Parsing `gitignore` files after we generate the list of potentially changed files and folders means that we need to visit every single file under the `target` folder because we don't know the `target` folder should be ignored at visit time. And that will just be very slow. This scenario is much less likely for `gitattributes` files: not only because `gitattributes` files don't accept folder patterns, it's also because `gitattributes` files are not meant to be used to ignore files.

> Is there any issue about this? Should I update .gitattributes to not follow symlinks and let .gitignore follow it?

Nice catch on the inconsistent gitignore behavior with git.

As of the next step, I am not sure. However, if I were you I would:

* File a github issue on the potential problematic jj gitignore file handling on symbolic link and check what the behavior the core maintainer wants to follow. Just in case there is a reason for the current gitignore implementation.
* Ask other core developers to make a decision on how to handle symbolic links for `gitattributes` files. (probably ask martinvonz who makes many desgin decisions when I have a question).

But, changes to gitignore behaviors will be in another commit if not another PR.

> Probably for next steps, is to add commands under jj git lfs that are gonna update .gitattributes directly.

Sorry I don't understand why would we need to introduce a command to update `gitattributes` files. Can you elaborate that a bit? Thanks.

> I'm open to suggestions

My simplisitic thought is that, the current interface doesn't make it easy to integrate with future tasks. For both the filter feature(which should solve the git lfs support once and for all) and the EOL conversion feature, we need to know the exact state of an attribute. And the EOL conversion feature needs to query the state of the `eol` attribute and the `text` attribute. Neither is supported by today's interface.

> I also expect that we're gonna reuse `GitAttributesFile` for the diff engine

I would argue that my design is more ergonomic, because with my design the diff tool code doesn't need to build up the `gitattributes` tree on its own. The current design requires the caller of `GitAttributesFile` to build up the `gitattributes` tree on its own by calling `chain_with_file` on every directory entry.

gusinacio:
> I am asking in details just because this CL is very different from [my design](https://github.com/06393993/jj/blob/gitattr-design/docs/design/gitattributes.md), that I plan to propose early this August, and am nervous about the unfortunate conflict: either my design will be rejected, or we need to revert this PR to adopt my design. If you are not hurry with this PR, I am wondering if you could read my design if you haven't already?

Thanks for sending me the design proposal. I was out of the loop but now I'm gonna take a look at it.

I'm not in a hurry with this PR, it's just that I've been using outdated versions of jj and alternatives PR/branches for months now and I'd like to have some support soon. In any case, I'm gonna review this proposal and try to make this PR compatible with it or give feedback about it (if I find something that doesn't fit).

06393993:
Thanks.

The plan for my design is to go through [the formal design process](https://jj-vcs.github.io/jj/latest/design_docs/#process) to avoid surprises on code review. I was somewhat actively discussing the design in the dicord channel mentioned in the readme file:

https://github.com/jj-vcs/jj/blob/main/README.md?plain=1#L11

yuja:
Regarding `.gitignore`, I don't think `jj` should follow symlinks in the working copy in general. Suppose `git` doesn't, I would say the current behavior is a bug. Please feel free to file a bug report.

zx8:
I built this PR locally with `cargo build -p jj-cli --release` and am attempting to use it with my dotfiles repository.

I added the following to my `~/.config/jj/config.toml`:

```toml
[git]
ignore-filters = ["git-crypt"]
```

And I have the following in my `.gitattributes`:

```
ssh/.ssh/* filter=git-crypt diff=git-crypt
```

Yet when I run `jj status`, I still see:

```
M ssh/.ssh/id_rsa
```

What am I doing wrong?

gusinacio:
> I built this PR locally with `cargo build -p jj-cli --release` and am attempting to use it with my dotfiles repository.
> 
> I added the following to my `~/.config/jj/config.toml`:
> 
> ```toml
> [git]
> ignore-filters = ["git-crypt"]
> ```
> 
> And I have the following in my `.gitattributes`:
> 
> ```
> ssh/.ssh/* filter=git-crypt diff=git-crypt
> ```
> 
> Yet when I run `jj status`, I still see:
> 
> ```
> M ssh/.ssh/id_rsa
> ```
> 
> What am I doing wrong?

You probably need to do `jj file untrack ssh/.ssh/id_rsa` since the file was probably already in the commit.

udaya2899:
@gusinacio @06393993  do you folks think this PR is ready? We'd really like to make jj become usable in our org as our repo has a lot of LFS files.

I've been trying this off of this branch, works fairly well for our use case. But it's harder to share with the rest of the org as we'd prefer a pre-installed binary in the VSCode Devcontainer. Any further progress in this PR would be helpful for us. Thanks in advance

06393993:
Re udaya2899,

I am not a maintainer, so I can't make the decision. But I personally am not against this PR, because we can always introduce fixes afterwards. However, this PR does have some problems that I don't think we have a good answers for:

* It always reads and parses all the `.gitattributes` files. While this may not introduce performance issues in practice, we probably need to document that somewhere.
* `git.ignore-filters` only takes effect on snapshot, but not on checkout. While this may be a deliberate design, it's not intuitive.
* This question is for the maintainers: this PR is clearly not compatible with [the proposed design](https://github.com/jj-vcs/jj/pull/7164). What's our plan if we merge this PR? Are we Ok with accepting the alternative design and refactor this PR(mostly revert) accodingly in the future?

udaya2899:
Thanks 06393993 for clarifying. I request @yuja or @martinvonz or any other maintainer to pitch-in and help resolve the questions and merge this. 

This PR makes `jj` usable in repos with LFS (#80) and we've been using this branch for a long while and would like to use from the main branch.

zx8:
I've also been manually compiling this branch locally in order to make `jj` usable. Given how long this feature has been outstanding for (including all the previous attempts), I'd argue that at this point just getting _something_ merged in to get things moving is worthwhile, and iterate on it later.

It could even be flagged as an experimental feature, as has been done with other configuration options (like auto-advance branches), just so users can start playing with it without having to rebase and compile custom branches, and most importantly allow the average `jj` user to provide feedback.

martinvonz:
Is this still missing a CLA? I'm not even allowed to look at it until the CLA has been signed. I've also been very busy, so the lack of CLA is not the only reason I haven't looked at it. But it would be good if you can sign the CLA so it is at least ready for review.

PhilipMetzger:
> Is this still missing a CLA? I'm not even allowed to look at it until the CLA has been signed. I've also been very busy, so the lack of CLA is not the only reason I haven't looked at it. But it would be good if you can sign the CLA so it is at least ready for review.

I assume @weiznich's public email is not the one with the CLA, which is the reason it fails but not for his individual PR (see https://github.com/jj-vcs/jj/pull/7098#issuecomment-3124631293). Otherwise this is extremely similar to @bcspragu PR with minor changes.

PhilipMetzger:
So an option would be to remove him from the co-authors list and see if it passes the CLA check.

udaya2899:
@gusinacio can you please edit the commit message to remove @weiznich from co-author?

Or is it a possibility to change the email of the author with a rebase with @weiznich's permission to use their personal email that has CLA signed?

weiznich:
As already expressed to multiple parties: I'm aware of this PR and the problems with the CLA and I'm willing to spend some time figuring out how to resolve them **if** this is the last thing that prevents this PR from getting merged. So far I have not seen any such indication from any of the jj contributors here or in any older PR. For me personally it does seem  like there is a great interest from the contributor team to merge such a solution. To be really clear here: That's totally fine as it's their project and they ultimately need to maintain this code. For me that means I rather would like to spend my time on other things than spending it on figuring out how to resolve problems with other peoples CLA setup.

If that blocks this PR and you end you end up removing my co-authorship from the commit I kindly ask that you also remove all my contributions from this commit. As far as I can see they shouldn't be required for the final change, so this shouldn't make things harder.

Finally I would like to ask that you all stop pinging me (or even send private mails to me). That's rather spammy and I'm honestly a bit disappointed that the jj team even allows that in their repository.

kejadlen:
Although I used `@weiznich`'s original PR as a starting point, at this point there's no code from his approach left between switching to a) using `gix`, b) not using a `Matcher`, and c) using `ignore-files` instead of `ignore-lfs-files` - I had left `@weiznich` as a co-author on the commit largely since I thought he still deserved credit for kicking off this whole thing, but probably someone else should double-check me that there is indeed no more of his code in the commit.

gusinacio:
@kejadlen is absolutely correct.

I double checked and all the initial ideas were now replaced by gix-attributes and a totally different approach to config, using a boolean to ignore lfs, while now it's ignoring any filtered file in .gitattribute based on config.

I just rebased and removed the previous Co-author.

lib/src/lib.rs
miallo: For the reference: The design document in question is worked on in #7164

udaya2899:
Replying to https://github.com/jj-vcs/jj/pull/7098#issuecomment-3225075874 (@martinvonz), the CLA issues are now fixed and the author says the PR has been ready since more than a month.  

There's been more and more interest for some solution to this in https://github.com/jj-vcs/jj/issues/80. It'd benefit the jj users community a lot if we can merge this PR.

cli/src/cli_util.rs
```diff
@@ -1434,6 +1437,19 @@ to the current parents may contain changes from multiple commits.
...
+    }
+
+    #[cfg(feature = "git")]
+    #[instrument(skip_all)]
+    pub fn base_attributes(&self) -> Result<Arc<GitAttributesFile>, ConfigGetError> {
+        Ok(Arc::new(GitAttributesFile::new(
+            &self.settings().git_settings()?.ignore_filters,
+        )))
+    }
+
```
martinvonz: nit: use `cfg!` instead:
```rust
    #[instrument(skip_all)]
    pub fn base_attributes(&self) -> Result<Arc<GitAttributesFile>, ConfigGetError> {
        if cfg!(feature = "git") {
            Ok(Arc::new(GitAttributesFile::new(
                &self.settings().git_settings()?.ignore_filters,
            )))
        } else {
            Ok(Arc::new(GitAttributesFile::default()))
        }
    }
```

That way we always compile-test both branches and IDEs can better understand both branches too.

martinvonz:
> Thanks for sending me the design proposal. I was out of the loop but now I'm gonna take a look at it.

@gusinacio: Did you get a chance to do that? Sorry if I missed it.

kejadlen:
> What's our plan if we merge this PR? Are we Ok with accepting the alternative design and refactor this PR(mostly revert) accodingly in the future?

FWIW, as the original author of this particular implementation, that seems ideal to me, as it would be much better to have real smudge/clean support for LFS rather than just ignoring the files in question.

gusinacio:
> > Thanks for sending me the design proposal. I was out of the loop but now I'm gonna take a look at it.
> 
> @gusinacio: Did you get a chance to do that? Sorry if I missed it.

I'm implementing it today, I got time to take a look here. When I looked the first time, it wasn't complete and I decided to wait the full proposal.

> I plan to propose early this August


So far the design is looking good, there are just a few async functions that are not async in the design.

I'm gonna still just ignore files with a given config filter like I was doing but I'll switch to the new design for GitAttributes.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,589 @@
...
+                .map_err(|err| GitAttributesError {
+                    message: "Could not retrieve the value from path".to_string(),
+                    source: err.into(),
+                })?;
+        let Some(file_to_merge) = merged_tree_value.to_file_merge() else {
+            // this means the file doesn't exist in the tree
+            return Ok(None);
+        };
+        // try to resolve the file
+        let id = match &file_to_merge.resolve_trivial(SameChange::Accept) {
```
gusinacio: Should this be `SameChange::Accept`?

cli/src/merge_tools/diff_working_copies.rs
```diff
@@ -153,6 +153,7 @@ pub(crate) fn check_out_trees(
             conflict_marker_style,
             eol_conversion_mode: EolConversionMode::None,
             fsmonitor_settings: FsmonitorSettings::None,
+            ignore_filters: Vec::new(),
```
gusinacio, line 155: Not sure how to load from user settings here, or if it's needed.

CHANGELOG.md
PhilipMetzger, line 1: nit: squash this into the first commit which completes `git-lfs`

gusinacio: Done!

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![expect(missing_docs)]
+#![allow(unused)]
+
+use std::collections::HashMap;
+use std::fs::File;
+use std::fs::{self};
```
06393993: Just use `std::fs`.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+        let merged_tree_value =
+            self.tree
+                .path_value_async(path)
+                .await
+                .map_err(|err| GitAttributesError {
+                    message: "Could not retrieve the value from path".to_string(),
+                    source: err.into(),
+                })?;
+        let Some(file_to_merge) = merged_tree_value.to_file_merge() else {
+            // this means the file doesn't exist in the tree
```
06393993: This assumption is wrong: if the merge has multiple terms, and *one of* the terms is not a file(e.g. one term is a folder, another term is a file; a more practical case will be one term is a symlink, another term is a file[^1]), `None` will also be returned. A better way to handle this case is to fallback to check if one of the terms is file, and use that. Check https://github.com/jj-vcs/jj/compare/main...06393993:jj:gitattr-example-impl#diff-09e46fa0436a80b91b08a5861cd4df4ee1a87865c4853a055d78eb37a334202aR74-R92 on how to implement such fallback.

Having said so, I don't think we have to implement such fallback, we can do it in a follow-up PR with tests.

[^1]: While gitattributes shouldn't follow symlink per https://git-scm.com/docs/gitattributes#_notes. This can happen when the user thinks a symlink can work, and then realize it doesn't work, so they change it to a file in another "branch", make modification, and merge.

lib/src/gitattributes.rs
06393993: I am not very familiar with the purpose of adding the `same_change` parameter, but most old call sites just use `SameChange::Accept`, so I assume `SameChange::Accept` is correct.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+            .map_err(|err| GitAttributesError {
+                message: "Could not convert path into fs path".to_string(),
+                source: err.into(),
+            })?;
+        let file = match File::open(&path) {
+            Ok(file) => file,
+            Err(error) if error.kind() == ErrorKind::NotFound => return Ok(None),
+            Err(err) => {
+                return Err(GitAttributesError {
+                    message: "There was an io error".to_string(),
```
06393993: Can we record the path in the error message, e.g. `format!("Failed to open the file at {}", path.display())`?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+                    source: err.into(),
+                });
+            }
+        };
+        // we use std::fs::symlink_metadata to not follow symlinks
+        let metadata = fs::symlink_metadata(&path).map_err(|err| GitAttributesError {
+            message: "There was an io error".to_string(),
```
06393993: Can we record the path in the error message, e.g. `format!("Failed to obtain the file metadata of {}", path.display())`?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+            source: err.into(),
+        })?;
+        if !metadata.is_file() {
+            return Ok(None);
+        }
+        Ok(Some(Box::new(BlockingAsyncReader::new(file))))
+    }
+}
+
+type CacheEntry = Arc<(Search, MetadataCollection)>;
```
06393993: This name can be confusing, because the actual cache entry type is `GitAttributesNode`, but I also don't have a good naming idea.

This is not a blocking comment.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+                    git_attributes_path
+                        .to_fs_path(&PathBuf::new())
+                        .map_err(|err| GitAttributesError {
+                            message: "Could not convert gitattributes path into PathBuf"
+                                .to_string(),
+                            source: err.into(),
+                        })?,
+                    Some(&PathBuf::new()),
+                    &mut collection,
+                    self.parent.is_none(), /* allow macros */  // check only for root
```
06393993: Can we add a comment with a link to https://git-scm.com/docs/gitattributes#_defining_macro_attributes, so that it's clear that this is a git behavior?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+        }
+    }
+
+    pub(crate) async fn search(
+        &self,
+        path: &RepoPath,
+        attribute_names: impl AsRef<[&str]>,
+        priority: SearchPriority,
+    ) -> Result<HashMap<String, State>, GitAttributesError> {
+        let attributes = self.get_git_attributes_node(path);
```
06393993, line 334: I think, we should just call `get_git_attributes_node` with the parent of `path`, so that when we search the gitattributes file for `t1/t2/abc`, we don't look into the not existing `t1/t2/abc/.gitattributes` file, and we don't populate a cache entry for all files instead of just folders.

EDIT: I see that you change to use the parent path in the "git-attributes: use .gitattributes file for current folder" commit. That's also fine if we squash that commit into this.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+/// Errors for GitAttributes
+#[derive(Debug, thiserror::Error)]
+#[error("{message}")]
+pub struct GitAttributesError {
+    message: String,
+    #[source]
+    source: Box<dyn std::error::Error + Send + Sync>,
+}
+
+type TestStore = HashMap<RepoPathBuf, Result<String, String>>;
```
06393993: Move `type TestStore` and `impl FileLoader for TestStore` to `mod tests`. Rust should allow such structure. And I assume that we only need such store in unit tests.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+                    RepoPathBuf::from_internal_string(key.to_string()).unwrap(),
+                    Ok(value.to_string()),
+                )
+            })
+            .collect();
+
+        GitAttributes::new(HashMap::new(), data)
+    }
+
+    // macros so tests fail on the right line
```
06393993: Can we just use a function instead of macros?

If we use functions, and the test fails, the entire stack will be printed, so we don't lose the precise line where the assertion fails. If we use macros, we only see the line where the macro is used, we won't able to identify the exact line in macro where the panic happens.

In addition, functions have better readability than macro.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+    }
+
+    #[test]
+    fn test_search_unset() {
+        assert_attribute_output!("foo -text", "foo", "text", State::Unset);
+    }
+
+    #[test]
+    fn test_search_unspecified() {
+        assert_attribute_output!("!foo -text", "foo", "text", State::Unspecified);
```
06393993: Should the gitattributes contents be `"foo !text"`?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+    fn path_and_pattern_matching() {
+        // Using https://git-scm.com/docs/gitattributes#_examples as example
+        // It's not 1:1 because we don't support $GIT_DIR/info/.gitattributes
+        // that would override in-tree settings.
+        let git_attributes = create_git_attributes(&[
+            // ("$GIT_DIR/info/.gitattributes", "a*	foo !bar -baz"),
+            (".gitattributes", "abc	foo bar baz"),
+            (
+                "t/.gitattributes",
+                "ab*	merge=filfre
```
06393993: Can we use `indoc!` for multiple line string for better readability? `jj` almost always use `indoc` for that.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+            abc	-foo -bar
+            *.c	frotz",
+            ),
+        ]);
+
+        // Based on the example, this should be "Set"
```
06393993: Should we change the comments so that they are not confusing? e.g. "In the example, foo should be set, but we don't have the local $GIT_DIR/info/.gitattributes file, so in our case, it should be unset".

I am also Ok with just removing those comments.

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+    }
+
+    #[test]
+    fn test_search_unspecified_no_pattern() {
+        assert_attribute_output!("foo elo=lf", "bar", "text", State::Unspecified);
+        assert_attribute_output!("foo elo=lf", "bar", "elo", State::Unspecified);
+    }
+
+    #[test]
+    fn path_and_pattern_matching() {
```
06393993: According to my sampling from `jj`, almost all the tests are prefixed with `test_`. Can we rename the test function with `test_` prefixed?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+            &git_attributes,
+            "t/abc",
+            "merge",
+            State::Value(Value::from("filfre"))
+        );
+        assert_search_output!(&git_attributes, "t/abc", "frotz", State::Unspecified);
+    }
+
+    #[test]
+    fn glob_matching() {
```
06393993: According to my sampling from `jj`, almost all the tests are prefixed with `test_`. Can we rename the test function with `test_` prefixed?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+        assert_attribute_output!("*.txt text", "bar.txt", "text", State::Set);
+        assert_attribute_output!("foo/ text", "foo/bar.txt", "text", State::Unspecified);
+        assert_attribute_output!("**/bar.rs text", "baz/bar.rs", "text", State::Set);
+    }
+
+    #[test]
+    fn test_case_sensitive_attr() {
+        assert_attribute_output!(
+            "foo text
```
06393993: Can we use `indoc` for multi-line string?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+    fn test_case_sensitive_attr() {
+        assert_attribute_output!(
+            "foo text 
+            FOO -text",
+            "foo",
+            "text",
+            State::Set
+        );
+        assert_attribute_output!(
+            "foo text
```
06393993: Can we use indoc for multi-line string?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+        let git_attributes = create_git_attributes(&[("foo/.gitattributes", "bar.txt text")]);
+        assert_search_output!(&git_attributes, "foo/bar.txt", "text", State::Set);
+    }
+
+    #[test]
+    fn test_macro_definition() {
+        let git_attributes = create_git_attributes(&[
+            (
+                ".gitattributes",
+                "[attr]base_macro a
```
06393993: Can we use indoc for multi-line string?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+                [attr]override_macro b",
+            ),
+            (
+                "foo1/.gitattributes",
+                " # this macro definition shouldn't take effect.
```
06393993: Can we use indoc for multi-line string?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+                    [attr]override_macro c
+                    # this macro definition shouldn't take effect.
+                    [attr]new_macro e
+                    bar base_macro override_macro new_macro ",
+            ),
+            (
+                "foo2/.gitattributes",
+                "# this macro definition shouldn't take effect.
```
06393993: Can we use indoc for multi-line string?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+        assert_search_output!(&git_attributes, "foo2/bar", "new_macro", State::Set);
+    }
+
+    #[test]
+    fn search_priority_and_fallback() {
+        let disk: TestStore = [("foo/.gitattributes", "*.txt text")]
+            .iter()
+            .map(|(key, value)| {
+                (
+                    RepoPathBuf::from_internal_string(key.to_string()).unwrap(),
```
06393993: Not sure if the following is better. I don't feel strong on this.

```rust
let disk = TestStore::from([
    (
        repo_path("foo/.gitattributes").to_owned(),
        Ok("*.txt text\n".to_string()),
    ),
]);
```

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+            "bar/bar.txt",
+            "text",
+            State::Set,
+            SearchPriority::Disk
+        );
+    }
+
+    #[test]
+    fn file_loader_io_error() {
+        let store: TestStore = [(".gitattributes", "There was an IO error")]
```
06393993: Not sure if the following is better. I don't feel strong on this.

```rust
let disk = TestStore::from([
    (
        repo_path(".gitattributes").to_owned(),
        Err("There was an IO error".to_string()),
    ),
]);
```

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+        assert_search_output!(&git_attributes, "foo2/bar", "a", State::Set);
+        assert_search_output!(&git_attributes, "foo2/bar", "b", State::Set);
+        assert_search_output!(&git_attributes, "foo2/bar", "c", State::Unspecified);
+        assert_search_output!(&git_attributes, "foo2/bar", "d", State::Unspecified);
+        assert_search_output!(&git_attributes, "foo2/bar", "e", State::Unspecified);
+        assert_search_output!(&git_attributes, "foo2/bar", "new_macro", State::Set);
+    }
+
+    #[test]
+    fn search_priority_and_fallback() {
```
06393993: According to my sampling from `jj`, almost all the tests are prefixed with `test_`. Can we rename the test function with `test_` prefixed?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,705 @@
...
+            &git_attributes,
+            "bar/bar.txt",
+            "text",
+            State::Set,
+            SearchPriority::Disk
+        );
+    }
+
+    #[test]
+    fn file_loader_io_error() {
```
06393993: According to my sampling from `jj`, almost all the tests are prefixed with `test_`. Can we rename the test function with `test_` prefixed?

lib/src/gitattributes.rs
```diff
@@ -305,11 +308,13 @@ impl GitAttributes {
     pub fn new(
         store_file_loader: impl FileLoader + 'static,
         disk_file_loader: impl FileLoader + 'static,
+        ignore_filters: &[String],
```
06393993: Can we just move this paramter directly to `filter_matches`?

`GitAttributes` is somewhat self-contained and a common type used by all functionality related to git attributes. I don't think we should directly add interfaces to the type itself to implement those funcitons. This makes tests more difficult, as you can tell, you need to add an irrelevant `&[]` to tests.

I know such requests result that you need to hold and pass `ignore_filters` in various different layers on the caller side(i.e. `TreeState` and `FileSnapshotter`), but I believe that's for the better.

I am not sure if other reviewers ask the other way around. If so, just ignore my comment then. They may argue that, we should put git attributes logic in one file/type, so that everything is more tightly cohesive, and I think that's reasonable. I don't consistantly maintain the project, so my words weigh less.

An alternative design is that, you can create a `GitAttributesExt` trait for `GitAttributes`, and implement `fn filter_matches` separately. This allows a better isolation between the `filter_matches` logic and the core `GitAttributes` logic. Furthermore, you can introduce a `GitAttributes` trait with only the `fn search(&self, ...)` function. In this case, you can mock `GitAttributes` in tests, resulting a even better decoupling in testing. But I don't feel strong on either adding a `GitAttributesExt` trait or changing `GitAttributes` to a trait.

lib/src/gitattributes.rs
```diff
@@ -381,6 +386,23 @@ impl GitAttributes {
         map.insert(path.to_owned(), new_node.clone());
         new_node
     }
+    /// Matches if the given `path` contains the `filter` value
+    /// in .gitattributes and if the value is contained in
+    /// the provided list of ignored filters
+    pub fn filter_matches(&self, path: &RepoPath) -> bool {
+        let Ok(result) = self
+            .search(path, ["filter"], SearchPriority::Disk)
+            .block_on()
```
06393993: While I don't very much like the idea to call `block_on` here, I am Ok with it, because it doesn't have practical influence: `GitBackend` implements `async fn read_file(...)` in a purely sync way.

Accidental use of `block_on` in async environment can result in hang, so my personal principal is to move the `block_on` to as higher in the stack as possible. If all `block_on` call happens only in `local_working_copy.rs`, it's easier to audit the code when we want to fix a hang, and we suspect that `block_on` is the cause.

06393993 (changes requested):
1. Can we squash the "git-attributes: use .gitattributes file for current folder" commit into the "git-attributes: add git attributes file support" commit?
2. We are missing some tests but that can be another PR from me if you don't want to do the job:
    1. Tests for the 2 `FileLoader` implementations. `TreeFileLoader` is especially complicated and worth testing.
    2. Stress test on the depth of the folder to ensure that the recursion implementation won't result in stack overflow under reasonable workload.
    3. Simple integration tests to make sure that everything is glue-ed correctly.
3. I may change the design to specify the `FileLoader` priority at `GitAttributes` creation time, and we allow aribtrary number of `FileLoader` to initialize `GitAttributes`. This could simplify the implementation. But that's another issue to solve, and will be a different PR.

lib/src/gitattributes.rs
gusinacio: Thanks!

lib/src/gitattributes.rs
gusinacio: Done!

lib/src/gitattributes.rs
gusinacio: I moved this to filter_matches. I created an extension trait called `GitAttributesFilter`. Leaving this as not resolved to wait for further reviews.

lib/src/gitattributes.rs
gusinacio: Ok, using the same code as the example.

lib/src/gitattributes.rs
gusinacio: I moved out because I was using in some other place where I couldn't reach the root nor the merge tree, but it's not being used anymore, moving back to mod tests.

lib/src/gitattributes.rs
06393993: It seems that you left 👍 , but doesn't change this place. Unresolved so that we don't miss this place

lib/src/gitattributes.rs
06393993: Unresolved, because indoc is still not used here.

lib/src/gitattributes.rs
06393993: Unresolved, because indoc is still not used here.

lib/src/gitattributes.rs
```diff
@@ -394,6 +393,32 @@ impl GitAttributes {
...
+    fn filter_matches(
+        &self,
+        path: &RepoPath,
+        ignore_filters: &[String],
+    ) -> impl Future<Output = bool>;
+}
+
+impl GitAttributesFilter for GitAttributes {
+    async fn filter_matches(&self, path: &RepoPath, ignore_filters: &[String]) -> bool {
+        let Ok(result) = self.search(path, ["filter"], SearchPriority::Disk).await else {
```
06393993: Should the caller decide the priority, i.e. when updating(tree -> disk) tree(`Store`) is the priority, when snapshotting, disk is the priority, just in case the users move between revisions where `gutattribues` files are changed?

lib/src/settings.rs
```diff
@@ -73,6 +74,10 @@ impl GitSettings {
             executable_path: settings.get("git.executable-path")?,
             write_change_id_header: settings.get("git.write-change-id-header")?,
             colocate: settings.get("git.colocate")?,
+            ignore_filters: settings
```
06393993, line 77: Should we also change `cli/src/config-schema.json` and relevant tests?

lib/src/gitattributes.rs
```diff
@@ -435,6 +434,7 @@ mod tests {
 
     use gix_attributes::state::Value;
     use indoc::indoc;
+    use pollster::FutureExt as _;
```
06393993, line 447: Can we squash changes in this file into the previous commit? Thanks

lib/src/gitattributes.rs
06393993: LGTM

lib/src/local_working_copy.rs
```diff
@@ -918,6 +937,8 @@ impl TreeState {
             conflict_marker_style,
             fsmonitor_settings: fsmonitor_settings.clone(),
             target_eol_strategy: TargetEolStrategy::new(eol_conversion_mode),
+            git_attributes,
```
06393993, line 938: It's not complete to initialize `GitAttributes` only once. We should update `git_attributes` everytime `TreeState::tree_id` is updated. With the current implementation, we end up with `GitAttributes` not actually reading from the tree. While I am personally Ok with such out-of-sync, I am wondering if we could at least add a TODO comment for this obvious culprit?

lib/src/local_working_copy.rs
```diff
@@ -902,9 +914,16 @@ impl TreeState {
             conflict_marker_style,
             eol_conversion_mode,
             ref fsmonitor_settings,
+            ref ignore_filters,
         }: &TreeStateSettings,
     ) -> Self {
         let tree_id = store.empty_merged_tree_id();
+
+        let tree = store.get_root_tree(&tree_id).unwrap();
```
06393993: I am not sure if we want to avoid `unwrap()` here or at least comment why we won't ever panic here @PhilipMetzger ?

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,712 @@
...
+        path: &RepoPath,
+    ) -> Result<Option<Box<dyn AsyncRead + Send + Unpin>>, GitAttributesError> {
+        let path = path
+            .to_fs_path(&self.repo_root)
+            .map_err(|err| GitAttributesError {
+                message: "Could not convert path into fs path".to_string(),
+                source: err.into(),
+            })?;
+
+        // we use std::fs::symlink_metadata to not follow symlinks
```
PhilipMetzger: nit: can you finish the sentence here with "to follow Git's behavior."

lib/src/gitattributes.rs
PhilipMetzger: I'd also consider making it a named struct, so it's obvious which field you access

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,712 @@
...
+                            }
+                        })?);
+                let mut async_reader = match self
+                    .primary_loader(priority)
+                    .load(&git_attributes_path)
+                    .await?
+                {
+                    Some(reader) => reader,
+                    None => {
+                        // bail to the secondary loader
```
PhilipMetzger: nit: s/bail/fallback

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,712 @@
...
+                    git_attributes_path
+                        .to_fs_path(&PathBuf::new())
+                        .map_err(|err| GitAttributesError {
+                            message: "Could not convert gitattributes path into PathBuf"
+                                .to_string(),
+                            source: err.into(),
+                        })?,
+                    Some(&PathBuf::new()),
+                    &mut collection,
+                    // Macros can only be defined in top-level gitattirbutes:
```
PhilipMetzger: nit: typo "gitattirbutes" => "gitattributes"

lib/src/local_working_copy.rs
PhilipMetzger: I'm pretty sure we should avoid new panics even though this implementation is only used in the Git backend.

lib/src/local_working_copy.rs
PhilipMetzger: Yes, please leave a TODO because otherwise we'll have the same wonky behavior as with the EOL handling

lib/src/gitattributes.rs
gusinacio: thanks, updated

lib/src/gitattributes.rs
gusinacio: done.

lib/src/gitattributes.rs
gusinacio: done

lib/src/gitattributes.rs
gusinacio: done

lib/src/local_working_copy.rs
gusinacio: After the latest rebase, I removed it since I can access directly using `store.empty_merged_tree()`.

lib/src/local_working_copy.rs
gusinacio: I left a TODO

lib/src/gitattributes.rs
gusinacio: I added the SearchPriority parameter.

lib/src/settings.rs
gusinacio: I added to config-schema.json. Any examples on which tests to add for it?

lib/src/gitattributes.rs
gusinacio: oh, missed that, thanks

lib/src/gitattributes.rs
gusinacio: thanks!

lib/src/gitattributes.rs
gusinacio: Missed it, thanks!

lib/src/gitattributes.rs
gusinacio: I updated to names struct `SearchAndCollection` and I'm using it directly as `Arc<SearchAndCollection>` where it used to be `CacheEntry`

lib/src/settings.rs
06393993: > I added to config-schema.json. Any examples on which tests to add for it?

Check the test failure and you can tell what you missed. Currently, you fail the `test_config_get_yields_values_consistent_with_schema_defaults` test. But it's possible that you fail some other tests once you fix this one. So just rerun the entire test suites locally to know what you miss.

06393993 (changes requested):
In case you miss the message from discord channel:

Can you also take a look at https://github.com/jj-vcs/jj/pull/7098#pullrequestreview-3408862309 and reply to it? The current behavior is understandably weird in that we don't handle update(store -> disk), but only handle snapshot(disk -> store). I think we need to document in details why we need such weird behavior to avoid future maintainers from confusing or even incorrectly "fixing" this weird behavior. Besides, we should make sure maintainers know that this design is deliberate, and they accept such design.

romen:
> I think this would look good if we didn't choose another approach in the Design doc.

@PhilipMetzger which design doc are you referring to? I am looking at https://github.com/jj-vcs/jj/tree/1e521f564dcec8f93ba3b0a494259d38ef0d8341/docs/design and cannot see any mention of LFS.

But maybe what you are referring to is some indirect effect of another design doc on attribute handling?

It would be nice to understand if it would be doable to go with this "ignore lfs" approach and the work done by @gusinacio in this PR for the time being until the time when jujutsu will have implemented the "better-LFS" plans!

06393993:
> > I think this would look good if we didn't choose another approach in the Design doc.
> 
> @PhilipMetzger which design doc are you referring to? I am looking at https://github.com/jj-vcs/jj/tree/1e521f564dcec8f93ba3b0a494259d38ef0d8341/docs/design and cannot see any mention of LFS.
> 
> But maybe what you are referring to is some indirect effect of another design doc on attribute handling?
> 
> It would be nice to understand if it would be doable to go with this "ignore lfs" approach and the work done by @gusinacio in this PR for the time being until the time when jujutsu will have implemented the "better-LFS" plans!

Please refer to https://github.com/jj-vcs/jj/pull/8144 for the WIP filter gitattributes design. Thanks.

romen:
> Please refer to #8144 for the WIP filter gitattributes design. Thanks.

Thanks @06393993 for the additional context!

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,710 @@
...
+    disk_file_loader: Arc<dyn FileLoader>,
+    store_file_loader: Arc<dyn FileLoader>,
+    node_cache: Mutex<HashMap<RepoPathBuf, Arc<GitAttributesNode>>>,
+}
+
+impl std::fmt::Debug for GitAttributes {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("GitAttributes").finish()
+    }
+}
```
martinvonz, line 59: nit: Not needed?

cli/src/merge_tools/diff_working_copies.rs
martinvonz: Should not be needed. The user shouldn't be editing LFS files with jj, so they shouldn't need to edit diffs containing changes to LFS files. And even if they did, it would be fine to let them edit the tracked file (the one with a hash in it).

lib/src/local_working_copy.rs
```diff
@@ -808,6 +813,9 @@ pub struct TreeStateSettings {
     pub eol_conversion_mode: EolConversionMode,
     /// The fsmonitor (e.g. Watchman) to use, if any.
     pub fsmonitor_settings: FsmonitorSettings,
+
+    /// Ignore filters in .gitattributes
```
martinvonz, line 817: Can you add a bit more detail? What are the entries here? Here's a possibly better comment (not sure if correct):

```rust
/// Names of .gitattributes filters whose matching files should be ignored in the working copy
```

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,710 @@
...
+        let path = path
+            .to_fs_path(&self.repo_root)
+            .map_err(|err| GitAttributesError {
+                message: "Could not convert path into fs path".to_string(),
+                source: err.into(),
+            })?;
+
+        // we use std::fs::symlink_metadata to not follow symlinks to follow Git's
+        // behavior.
+        let metadata = match std::fs::symlink_metadata(&path) {
```
martinvonz, line 155: nit: We usually write this as `path.symlink_metadata()`

lib/src/gitattributes.rs
```diff
@@ -0,0 +1,710 @@
...
+            SearchPriority::Store => &self.disk_file_loader,
+            SearchPriority::Disk => &self.store_file_loader,
+        }
+    }
+
+    async fn primary_then_secondary(
+        &self,
+        priority: SearchPriority,
+    ) -> Result<Arc<SearchAndCollection>, GitAttributesError> {
+        // we use pin because this is a recursive call
```
martinvonz, line 230: nit: move this a few lines down so it's just before the `Box::pin()` call?

lib/src/gitattributes.rs
```diff
@@ -397,6 +397,38 @@ impl GitAttributes {
...
+    ) -> bool {
+        let Ok(result) = self.search(path, ["filter"], priority).await else {
+            return false;
+        };
+
+        let Some(State::Value(value)) = result.get("filter") else {
+            return false;
+        };
+        let value = value.as_ref().as_bstr();
+        ignore_filters.iter().any(|state| value == state)
```
martinvonz, line 428: nit: Should we use a `HashSet<BString>` instead so we don't have to scan the list? Or is the list more or less guaranteed to be very short?

lib/src/gitattributes.rs
```diff
@@ -397,6 +397,38 @@ impl GitAttributes {
...
+pub(crate) trait GitAttributesFilter {
+    /// Matches if the given `path` contains the `filter` value
+    /// in .gitattributes and if the value is contained in
+    /// the provided list of ignored filters
+    fn filter_matches(
+        &self,
+        path: &RepoPath,
+        ignore_filters: &[String],
+        priority: SearchPriority,
+    ) -> impl Future<Output = bool>;
```
martinvonz, line 410: nit: Make the function async instead of (explicitly) returning a future

martinvonz (commented):
Thanks! Sorry about the very long delay in reviewing this. I hope you haven't given up on this PR.

gusinacio:
I'm closing this, there's too much uncertainty on how to handle lfs.

kejadlen:
If we're serious about getting this merged, I don't mind picking it back up to get it over the line.

martinvonz:
@kejadlen: Thanks!

gusinacio:
@kejadlen go ahead!

kejadlen:
New PR here: https://github.com/jj-vcs/jj/pull/9068

PR #7102, opened 2025-07-27
docs: display optional arguments more simply

scott2000:
docs: display optional arguments more simply

I was initially planning to try adding an `nth` keyword argument to `parents()` instead of adding `first_parent()`, and I noticed that based on the current style, it would have been shown as one of the following in the revset documentation:

* `parents(x[, depth[, [nth=]index]])`
* `parents(x[, depth][, [nth=]index])`

These both seem difficult to read to me, so I'm proposing changing the style to something more like this:

* `parents(x, [depth], [[nth=]index])`

I also added a brief section explaining the syntax in case it isn't immediately obvious what the square brackets mean.

# Checklist

If applicable:

### DIFF
--- a/docs/revsets.md
+++ b/docs/revsets.md
@@ -192,28 +192,43 @@ You can use parentheses to control evaluation order, such as `(x & y) | z` or
 You can also specify revisions by using functions. Some functions take other
 revsets (expressions) as arguments.
 
-* `parents(x[, depth])`: `parents(x)` is the same as `x-`.
+??? note "Function argument syntax"
+
+    In this documentation, optional arguments are indicated with square
+    brackets like `[arg]`. Some arguments also have an optional label which can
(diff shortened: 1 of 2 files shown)


yuja (commented):
I don't have a strong feeling, and both old/new states look good to me. The old notation wouldn't be correct anyway since `remote_bookmarks(remote_pattern)` is invalid for example.

> `parents(x[, depth[, [nth=]index]])`

I would use `parents(x[, depth][, [nth=]index])`, btw.

scott2000:
I'll leave this PR open to see if anyone else has thoughts about it.

bryceberger:
This seems to resolve the confusion brought up in discord a while ago, https://discord.com/channels/968932220549103686/968932220549103689/1387577477622403073. :+1:

scott2000:
Since there seem to be no objections, I think I'll merge this now.

PR #3663, opened 2024-05-10
Refuse to split an empty commit with `jj split`.

aspotashev:
Rationale: The user may be confused by the empty diff in the diff editor tool if they accidentally run `jj split` on a wrong (empty) commit.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- [ ] I have updated the documentation (README.md, docs/, demos/)
- [ ] I have updated the config schema (cli/src/config-schema.json)
- [x] I have added tests to cover my changes

PR #7109, opened 2025-07-28
cli: make `split` behave more like `commit`

thoughtpolice:
(This was a private stack I had and am publishing for some discussion.)

There was some talk on the Discord about how `jj commit` is actually just `jj split` in disguise and the differences are mostly artificial. This removes two of them:

1. It re-allows splitting an empty commit; this was disallowed in https://github.com/jj-vcs/jj/pull/3663 and thus is mostly a revert.
2. It does not warn when one side of the split contains all files; `jj commit` is roughly equivalent to `jj split 'all()'`, but split warns in this case.

But I'm leaving this as draft; I'm not sure if we actually want to do this. I personally think it's nice but there are probably more arguments to be had.

# Checklist

If applicable:

### DIFF
--- a/cli/src/commands/split.rs
+++ b/cli/src/commands/split.rs
@@ -38,7 +38,6 @@ use crate::cli_util::DiffSelector;
 use crate::cli_util::RevisionArg;
 use crate::cli_util::WorkspaceCommandHelper;
 use crate::cli_util::WorkspaceCommandTransaction;
-use crate::command_error::user_error_with_hint;
 use crate::command_error::CommandError;
 use crate::complete;
 use crate::description_util::add_trailers;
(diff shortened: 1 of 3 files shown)


thoughtpolice:
The original motivation for this discussion was https://github.com/jj-vcs/jj/commit/a00b5b8327494f100907b31319133da5ae0134ea which lead to us talking about the radical-but-somewhat-joking proposal of removing `jj commit` entirely in favor of `split`, much like we did for `new` and `merge/checkout`. I don't think we'd ever be able to pull that off without people rioting, but it led us down the path of talking about split's differences.

Now that I read the backlog, I realize we probably should keep the warning for "one side contains all changes" but only when the user is invoking things interactively. I can fix that later.

cli/src/commands/split.rs
martinvonz, line 1: > cli: don't error on splitting an empty commit

I think this needs more motivation, especially since it rolls back something someone deliberately added. I think it would make sense to error out only in the interactive case. It probably makes sense to also have `commit -i` error out if no changes were selected.

CC @aspotashev

thoughtpolice: Yeah, I'm normally reluctant to just revert things like this, hence the draft :) I think the interactive case is probably a good compromise, fwiw.

PR #7116, opened 2025-07-29
revset: optimize `heads(first_ancestors(x) & ...) `

scott2000:
This optimization is already done for `heads(ancestors(x) & ...)`, so I think it makes sense to extend it to `first_ancestors()` as well. This is especially important if the expression involves a filter that requires reading commits. For instance, in the nixpkgs repo, evaluating `heads(first_ancestors(@) & description(jujutsu))` takes 2.6 seconds on my machine before this optimization, and only 0.2 seconds after.

# Checklist

If applicable:


lib/src/default_index/composite.rs
```diff
@@ -41,6 +42,7 @@ use super::readonly::ReadonlyCommitIndexSegment;
 use super::revset_engine;
 use crate::backend::ChangeId;
 use crate::backend::CommitId;
+use crate::default_index::rev_walk::filter_slice_by_range;
```
yuja: nit: `super::rev_walk`

lib/src/revset.rs
```diff
@@ -2010,20 +2016,29 @@ fn fold_redundant_expression<St: ExpressionState>(
...
         RevsetExpression::Ancestors {
             heads,
             generation: Range {
                 start,
                 end: u64::MAX,
             },
-            parents_range: PARENTS_RANGE_FULL,
-        } => Ok(heads.ancestors_at(*start)),
+            parents_range,
+        } => Ok((heads.ancestors_at(*start), parents_range.clone())),
```
yuja: Perhaps, `parents_range` should also be propagated to the `heads`?

`ancestors(heads, generation=start, parents_range)`

scott2000: Oh that's a good catch, thanks!

yuja (approved):

PR #7117, opened 2025-07-29
Ig/rustfmt2024

ilyagr:


martinvonz (commented):
nit: Would CI pass after just the first commit? If not, I would squash the two commits.

ilyagr:
The CI actually fails if we include the second commit. We need to find some way to exclude `lib/src/protos` from `jj fix`.

**Update:** We can keep discussing this in https://github.com/jj-vcs/jj/pull/7118.

ilyagr:
The CI passes now.

ilyagr:
> Ah, I guess formatting of if-let chains and such would fail without this

The most noticeable difference is that 2024 edition orders imports differently. Before this PR, `cargo fmt` ordered imports one way, whole `jj fix` ordered them another way

PR #7118, opened 2025-07-29
contributing.md: exclude autogenerated files from `jj fix`

ilyagr:


ilyagr:
Let me test out `format_generated_files = false` rustfmt option first.

ilyagr:
This is superseded by ~~the latest version of #7117~~ #7119

ilyagr:
Actually, #7119 doesn't seem to work, see https://github.com/jj-vcs/jj/pull/7119#issuecomment-3130865615. :( So, maybe this is better.

I'm not entirely sure, but I'll mark this PR as ready. It's mainly a question of what approach we want to take, and of the two approaches that I thought of, this is the one that seems to work.

ilyagr:
A third approach that I haven't explored would be to change the process of generating these files so that we reformat the files with the correct settings and with *nightly* rustfmt after generating them with `protoc`.

docs/contributing.md
```diff
@@ -200,7 +200,7 @@ These are listed roughly in order of decreasing importance.
 Run this in the jj repo:
 
 ```shell
-jj config set --repo fix.tools.rustfmt '{ command = ["rustfmt", "+nightly"], patterns = ["glob:**/*.rs"] }'
+jj config set --repo fix.tools.rustfmt '{ command = ["rustfmt", "+nightly"], patterns = ["glob:**/*.rs ~ glob:lib/src/protos/**/*.rs"] }'
```
yuja, line 203: nit: negative pattern doesn't need glob

```
patterns = ["glob:**/*.rs ~ lib/src/protos"]
```

ilyagr: Good point. I'd probably then write it as `glob:**/*.rs ~ root:lib/src/protos` to be a bit clearer.

I'm still not sure whether it's best to suggest this or to change how those files are generated. I intend to maybe look into the latter at some point, but if you think this would be helpful sooner, let me know. It only affects running `jj fix` with `--include-unchanged-files`, which makes it less urgent.

emilazy: Formatting them on generation also makes people with global `nixfmt` fix rules more likely to get correct results.

ilyagr: Turns out `prost_build` uses <https://docs.rs/prettyplease/> to format the code, not `rustfmt`.

Adding an `ignore` config in `rustfmt.toml` does not affect `jj fix` ~~for some reason. This might be a bug and might be related to https://github.com/jj-vcs/jj/pull/7119 not working~~. This is because there seems to be no way to pass a filename to `rustfmt` when it's reading the code from stdin, see https://github.com/rust-lang/rustfmt/issues/5792 https://github.com/rust-lang/rustfmt/issues/5137? (~~Shouldn't affect #7119 though~~ See the last comment in #7119 for why that doesn't work.)

Perhaps the proto builder could try invoking `rustfmt +nightly` on each `.rs` file in the relevant dir after generating the files. This will only work with `rustup`-installed `rustfmt`, though.

PR #7119, opened 2025-07-29
rustfmt.toml: don't format generated files

ilyagr:
This mostly affects the files in `lib/src/protos`, and makes it so that `rustfmt` invoked via `jj fix --include-unchanged-files` does not format those.

This avoids errors like
https://github.com/jj-vcs/jj/actions/runs/16586832790/job/46913734860?pr=7117

An interesting fact is that this commit does *not* affect `cargo fmt`. `cargo fmt` skips those files either way because they are not proper modules, but are rather `include!`-d in their `mod.rs`.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -5,3 +5,4 @@ format_strings = true
 error_on_line_overflow = true
 group_imports = "StdExternalCrate"
 imports_granularity = "Item"
+format_generated_files = false


ilyagr:
Actually, upon more careful testing, this doesn't actually work.

With this PR, `jj fix --include-unchanged-files` still reformats e.g. `lib/src/protos/op_store.rs`.

It does make some difference: `rustfmt +nightly lib/src/protos/op_store.rs` does nothing with this PR. However, `jj new this_pr; jj fix --include-unchanged-files -s @` does reformat that same file.

So, perhaps #7118 is a better approach.

ilyagr:
The reason this doesn't work is https://github.com/rust-lang/rustfmt/issues/5080#issuecomment-1500232131.

I also filed https://github.com/rust-lang/rustfmt/issues/6620 for tracking it more specifically.

PR #7126, opened 2025-07-29
revset: don't reorder `~first_ancestors(x)` to start

scott2000:
Negated `first_ancestors()` can't be converted to a range currently, so it doesn't make sense to move it to the start when optimizing the order of an intersection.

# Checklist

If applicable:

### DIFF
--- a/lib/src/revset.rs
+++ b/lib/src/revset.rs
@@ -1883,13 +1883,17 @@ fn sort_negations_and_ancestors<St: ExpressionState>(
         RevsetExpression::Intersection(expression1, expression2) => {
             sort_intersection_by_key(expression1, expression2, |expression| match expression {
                 RevsetExpression::Ancestors {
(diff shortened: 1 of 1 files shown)


lib/src/revset.rs
```diff
@@ -1889,6 +1889,9 @@ fn sort_negations_and_ancestors<St: ExpressionState>(
                 RevsetExpression::NotIn(complement) => match complement.as_ref() {
                     RevsetExpression::Ancestors {
                         generation: Range { end: u64::MAX, .. },
+                        // We only want to move negated ancestors with a full parents range, since
+                        // these are the only negated ancestors which can be converted to a range.
+                        parents_range: PARENTS_RANGE_FULL,
                         ..
```
yuja: nit: might be better to omit unused fields explicitly so compilation would fail if we add one.

Issue #7085, opened 2025-07-26
FR: Add `--yes-pager` option to re-enable paging for individual commands


strangeglyph:
**Is your feature request related to a problem? Please describe.**
I've disabled pagination because I like being able to reference the output of commands like `jj status` and `jj log` for the next commands. However when I use `jj diff`, I do want pagination.

**Describe the solution you'd like**
A `--yes-pager` that works like `--no-pager` in reverse to re-enable the pager for commands when I need it

**Describe alternatives you've considered**
A more finegrained pagination control in the config would probably also work. Sometihng like `ui.paginate = diff`

martinvonz:
> I've disabled pagination because I like being able to reference the output of commands like `jj status` and `jj log` for the next commands.

Do you use `less`? Consider setting `LESS=FRX` in your environment (I think it's the `X` that's relevant here).

yuja:
You can also use `--when.commands` to enable pager conditionally.
https://jj-vcs.github.io/jj/latest/config/#conditional-variables

FWIW, I would add `--pager=<mode>` instead of `--yes/no-pager`.

strangeglyph:
> > I've disabled pagination because I like being able to reference the output of commands like `jj status` and `jj log` for the next commands.
> 
> Do you use `less`? Consider setting `LESS=FRX` in your environment (I think it's the `X` that's relevant here).

I'm not sure - if that is the default, then yes.



> You can also use `--when.commands` to enable pager conditionally. https://jj-vcs.github.io/jj/latest/config/#conditional-variables
> 
> FWIW, I would add `--pager=<mode>` instead of `--yes/no-pager`.

Thanks, I wasn't aware of those settings! Look like they'll solve my usecase perfectly.

Dane-Halverson:
#7127

Dane-Halverson:
#7127

PhilipMetzger:
I think this bug is also related https://github.com/jj-vcs/jj/issues/2926

PR #7127, opened 2025-07-29
CLI: added a cli option "--use-pager" that sets pager to auto

Dane-Halverson:
The option will set the pagination to auto. This is useful if you have pagination disabled in the config.toml and would like to use a pager. Changed tests to recognise description of option.


martinvonz:
> The option will set the pagination to auto. This is useful if you have pagination disabled in the config.toml and would like to use a pager. Changed tests to recognise description of option.

Please move this text to the commit description and squash the two commits. See https://jj-vcs.github.io/jj/latest/contributing/

I don't think we've had other requests for this feature. When would you use it? Is it that you normally turn off the pager but sometimes you want to enable it when you expect long output?

Why do you normally turn off the pager? Specifically, do you know about the `-F` and `-X` options to `less`?

Dane-Halverson:
I like to have pagination disabled by default as almost always I like to reference the output in other commands and see previous ones. When I know something is going to have a long output, and I'm not going to want to see it later, then I like the pager. I know if the -F and -X options, but prefer just a simple cli option. Sorry, I'm new to this.

martinvonz:
> I like to have pagination disabled by default as almost always I like to reference the output in other commands and see previous ones. When I know something is going to have a long output, and I'm not going to want to see it later, then I like the pager.

I'm the same but `less -FRX` gives me that behavior. What don't you like about it?

To be clear, I'm not strongly against adding a new flag. I just want to make sure that we do it for a good reason.

Dane-Halverson:
> > I like to have pagination disabled by default as almost always I like to reference the output in other commands and see previous ones. When I know something is going to have a long output, and I'm not going to want to see it later, then I like the pager.
> 
> I'm the same but `less -FRX` gives me that behavior. What don't you like about it?
> 
> To be clear, I'm not strongly against adding a new flag. I just want to make sure that we do it for a good reason.

I mostly just find it more friendly and batteries included. I also use PowerShell when on Windows, and while I'm sure there's a way to do it in PowerShell, I'm not sure what that is as PowerShell doesn't have a less command. Git also has a similar option I like to use that's built in, "-p", that does functionally the same thing.

martinvonz:
Did you see @yuja's [comment about making it `--pager=<mode>` instead](https://github.com/jj-vcs/jj/issues/7085#issuecomment-3124142880)?

Dane-Halverson:
If that's more desirable I could change it to that. I did --use-pager because that's closer to what Git uses. Just --pager would probably be better than --use-pager too.

martinvonz:
Mercurial has a `--pager` argument that accepts "true"/"always", "false"/"never", or "auto". I'm not sure we're ever going to want to support the "true"/"always" case. Maybe @yuja (who was also a Mercurial contributor) remembers if there was a good use case for that? Maybe it was just useful before more commands supported pagination? If that's the case, i.e. that we don't foresee supporting more modes than "never" and "auto", then just `--no-pager` and `--pager` seem best to me.

yuja:
I personally don't think `--pager=auto` was useful in Mercurial. My concern of `--use-pager` is that `ui.paginate=auto` doesn't mean the pager will be forcibly enabled.

BTW, last time we discussed about the naming, someone pointed out `--pager=<x>` sounds like an argument to specify the pager executable.

PR #7132, opened 2025-07-30
eol: convert the EOL for conflict on update

06393993:
The old code misses the handling for conflicts on the update path.

# Checklist

N/A:


lib/tests/test_eol.rs
```diff
@@ -373,6 +373,176 @@ fn test_eol_conversion_none_snapshot_conflicts() {
...
+            &root_commit,
+            &CheckoutOptions::empty_for_test(),
+        )
+        .unwrap();
+    testutils::write_working_copy_file(
+        test_workspace.workspace.workspace_root(),
+        file_repo_path,
+        parent2_contents,
+    );
+    let tree = test_workspace.snapshot().unwrap();
```
yuja: These `check_out()` and `snapshot()` look redundant. Maybe you can just write trees to repo, and check out the merge commit at the end.

yuja (approved):
Looks good, thanks.

lib/tests/test_eol.rs
06393993: Get rid of intermmediate `checkout` and `snapshot`, and directly creates the tree in the repo.

PR #7136, opened 2025-07-30
Allow changing the author date from describe

OvidiusCicero:
This is unfinished as it needs tests, probably breaks tests and has room for improvement but I first want to evaluate whether this change has change to land before polishing. Also the CLI is up for debate (should this be its own command or is describe the right place?) (Also should similar nobs be introduced for committer information)

Proof it works
~~~
❯ jj show wpltwvzr
Commit ID: 10c39be28a519cfb067bc0fd616d06d091eb43c09d30dcb5fe138c2182fc6bfa2a99c0c105778ec317acb3007c3a9bfba6fa51fc6f4743acb08f20bdbe95f8e8
Change ID: wpltwvzrrymrzszypnnlywkyulqwtoyq
Author   : …  (1985-12-20 02:39:57)
Committer: … (2025-07-30 18:29:04)

    tests 2

Modified regular file test.txt:
   1    1: ugl<dugflguflzg
        2: lufgl<uaszgduoz
        3: iuglugl
        4: 
        5: öyfiuvhöydiuvg
        6: yfuvöiufgv
        7: iuguig
        8: 
        9: dsuvpuidgv

~/trasch/testj on ☁️  🥋 wplt | tests 2 
❯ ~/trasch/cargo/debug/jj desc -r wpltwvzr --author-date 1995-12-19T16:39:57-08:00 --no-edit
Working copy  (@) now at: wpltwvzr 2b8d46f1 tests 2
Parent commit (@-)      : rppnmuxk b7004367 test 1

~/trasch/testj on ☁️  🥋 wplt | tests 2 
❯ jj show wpltwvzr
Commit ID: 2b8d46f110a6fb71fc9788b8e166a3e38c2047fa696f60be9ddae6b1955449bbf023809281b9af81e7cc913209a2a9c349d5ebb211d6ab053c2fa4c9c41bc011
Change ID: wpltwvzrrymrzszypnnlywkyulqwtoyq
Author: … (1995-12-20 02:39:57)
Committer: … (2025-07-30 18:29:55)

    tests 2

Modified regular file test.txt:
   1    1: ugl<dugflguflzg
        2: lufgl<uaszgduoz
        3: iuglugl
        4: 
        5: öyfiuvhöydiuvg
        6: yfuvöiufgv
        7: iuguig
        8: 
        9: dsuvpuidgv
~~~

# Checklist

If applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,7 +91,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
   precedence over `origin` if both exist.
 
 * Add the `jj touch` command, which modifies a revision's metadata. This can be
-  used to generate a new change-id, which may help resolve some divergences.
+  used to generate a new change-id, which may help resolve some divergences. It also has options to modify author name, email and time stamp.
 
 * Filesets now support case-insensitive glob patterns with the `glob-i:`,
   `cwd-glob-i:`, and `root-glob-i:` pattern kinds. For example, `glob-i:"*.rs"`
--- a/cli/src/commands/touch.rs
+++ b/cli/src/commands/touch.rs
@@ -16,8 +16,10 @@ use std::collections::HashSet;
 
 use clap_complete::ArgValueCompleter;
 use itertools::Itertools as _;
+use jj_lib::backend::Timestamp;
 use jj_lib::commit::Commit;
 use jj_lib::object_id::ObjectId as _;
+use jj_lib::time_util::parse_datetime;
 use tracing::instrument;
 
 use crate::cli_util::CommandHelper;
@@ -54,7 +56,7 @@ pub(crate) struct TouchArgs {
 
     /// Update the author timestamp
     ///
-    /// This update the author date to now, without modifying the author.
+    /// This updates the author date to now, without modifying the author.
     #[arg(long)]
     update_author_timestamp: bool,
 
@@ -81,6 +83,16 @@ pub(crate) struct TouchArgs {
         value_parser = parse_author
     )]
     author: Option<(String, String)>,
+
+    /// Set the author date to the given date either human
+    /// readable, eg Sun, 23 Jan 2000 01:23:45 JST) or as a time stamp, eg
+    /// 2000-01-23T01:23:45+09:00)
+    #[arg(
+        long,
+        conflicts_with = "update_author_timestamp",
+        value_parser = parse_datetime
+    )]
+    author_timestamp: Option<Timestamp>,
 }
 
 #[instrument(skip_all)]
@@ -141,6 +153,9 @@ pub(crate) fn cmd_touch(
                 }
                 if args.update_author_timestamp {
                     new_author.timestamp = commit_builder.committer().timestamp;
+                }
+                if let Some(author_date) = args.author_timestamp {
+                    new_author.timestamp = author_date;
                 }
                 commit_builder = commit_builder.set_author(new_author);
 
--- a/cli/tests/test_touch_command.rs
+++ b/cli/tests/test_touch_command.rs
@@ -240,6 +240,59 @@ fn test_touch() {
 
     [EOF]
     ");
+
+    // new author date
+    work_dir.run_jj(["op", "restore", &setup_opid]).success();
+    work_dir
+        .run_jj(["touch", "--author-timestamp", "1995-12-19T16:39:57-08:00"])
+        .success();
+    insta::assert_snapshot!(get_log(&work_dir), @r"
+    @  Commit ID: a527219f85839d58ddb6115fbc4f0f8bc6649266
+    │  Change ID: mzvwutvlkqwtuzoztpszkqxkqmqyqyxo
+    │  Bookmarks: c
+    │  Author   : Test User <test.user@example.com> (1995-12-19 16:39:57.000 -08:00)
+    │  Committer: Test User <test.user@example.com> (2001-02-03 04:05:26.000 +07:00)
+    │
+    │      (no description set)
+    │
+    ○  Commit ID: 75591b1896b4990e7695701fd7cdbb32dba3ff50
+    │  Change ID: kkmpptxzrspxrzommnulwmwkkqwworpl
+    │  Bookmarks: b
+    │  Author   : Test User <test.user@example.com> (2001-02-03 04:05:11.000 +07:00)
+    │  Committer: Test User <test.user@example.com> (2001-02-03 04:05:11.000 +07:00)
+    │
+    │      (no description set)
+    │
+    ○  Commit ID: e6086990958c236d72030f0a2651806aa629f5dd
+    │  Change ID: qpvuntsmwlqtpsluzzsnyyzlmlwvmlnu
+    │  Bookmarks: a
+    │  Author   : Test User <test.user@example.com> (2001-02-03 04:05:09.000 +07:00)
+    │  Committer: Test User <test.user@example.com> (2001-02-03 04:05:09.000 +07:00)
+    │
+    │      (no description set)
+    │
+    ◆  Commit ID: 0000000000000000000000000000000000000000
+       Change ID: zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz
+       Author   : (no name set) <(no email set)> (1970-01-01 00:00:00.000 +00:00)
+       Committer: (no name set) <(no email set)> (1970-01-01 00:00:00.000 +00:00)
+
+           (no description set)
+
+    [EOF]
+    ");
+
+    // invalid date gives an error
+    work_dir.run_jj(["op", "restore", &setup_opid]).success();
+    let command_result = work_dir.run_jj(["touch", "--author-timestamp", "aaaaaa"]);
+    assert!(!command_result.status.success());
+    insta::assert_snapshot!(command_result, @r"
+    ------- stderr -------
+    error: invalid value 'aaaaaa' for '--author-timestamp <AUTHOR_TIMESTAMP>': input contains invalid characters
+
+    For more information, try '--help'.
+    [EOF]
+    [exit status: 2]
+    ");
 }
 
 #[test]
--- a/lib/src/time_util.rs
+++ b/lib/src/time_util.rs
@@ -122,6 +122,16 @@ impl DatePattern {
     }
 }
 
+// @TODO ideally we would have this unified with the other parsing code. However
+// we use the interim crate which does not handle explicitly given time zone
+// information
+/// Parse a string with time zone information into a `Timestamp`
+pub fn parse_datetime(s: &str) -> chrono::ParseResult<Timestamp> {
+    Ok(Timestamp::from_datetime(
+        DateTime::parse_from_rfc2822(s).or_else(|_| DateTime::parse_from_rfc3339(s))?,
+    ))
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -188,5 +198,11 @@ mod tests {
         test_equal(now, "yesterday 5pm", "2024-01-01T01:00:00Z");
         test_equal(now, "yesterday 10am", "2023-12-31T18:00:00Z");
         test_equal(now, "yesterday 10:30", "2023-12-31T18:30:00Z");
+    }
+
+    #[test]
+    fn test_parse_datetime_non_sense_yields_error() {
+        let parse_error = parse_datetime("aaaaa").err().unwrap();
+        assert_eq!(parse_error.kind(), chrono::format::ParseErrorKind::Invalid);
     }
 }
--- a/cli/tests/cli-reference@.md.snap
+++ b/cli/tests/cli-reference@.md.snap
@@ -2687,7 +2687,7 @@ Modify the metadata of a revision without changing its content
    This generates a new change-id for the revision.
 * `--update-author-timestamp` — Update the author timestamp
 
-   This update the author date to now, without modifying the author.
+   This updates the author date to now, without modifying the author.
 * `--update-author` — Update the author to the configured user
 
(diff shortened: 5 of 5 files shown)


martinvonz:
What's the use case?

You can actually already do using this undocumented method:
```
JJ_TIMESTAMP=2000-01-01T00:00:00+00:00 jj describe --reset-author
```

I'm not sure if we should document that method or if we should add an argument like this PR does.

PhilipMetzger:
@AngelEzquerra also wanted something similar sometime ago now in Discord.

OvidiusCicero:
“undocumented”; maybe there was a misunderstanding but I did ask about it in the Discord before and the answer was JJ only supported this for name and E-Mail.

I have a slight preference for using flags instead of environment variables because then discovery is easier but I don't really care. It should be documented though, I need this functionality quite often tbh.

martinvonz:
We already have `--author`, so it seems consistent to add `--committer`, `--author-date`, and `--committer-date` too. What do you think?

> I need this functionality quite often tbh.

What do you use it for? Do you write tooling that imports commits from some other VCS (that's what I've used the feature for in Git)?

joyously:
> I have a slight preference for using flags instead of environment variables

I can't tell which date will be set with `JJ_TIMESTAMP`. I also think a flag could be more forgiving about date format, and could use a function to get the date from some other commit.

> What do you use it for?

I've asked about it before because I have generated a repo from zip files (work from before I heard about VCS) and from someone else's project that wasn't in a VCS.

martinvonz:
Another reason to prefer flags is that it more obviously applies only to the specified commits. A big problem with `JJ_TIMESTAMP` (and `JJ_USER` and `JJ_EMAIL`) is that they apply to all rewritten commits.

OvidiusCicero:
> We already have --author, so it seems consistent to add --committer, --author-date, and --committer-date too. What do you think?

yes, I'd add it

> What do you use it for? Do you write tooling that imports commits from some other VCS (that's what I've used the feature for in Git)?

Oh just something I edit via phone where I don't have access to command line and make commits (weeks) later

AngelEzquerra:
As @PhilipMetzger said, I'd like to have a way to change the commit and author dates after the fact and to keep them in sync. As an ex mercurial user I find the author / committee distinction unnecessary and a bit confusing. I also think that it's a little privacy unfriendly. Why should people know or care when I first committed the first draft of a change? It's not even that useful since when you first committed something and when you first started working on it may be two completely different dates... Perhaps in jujutsu it's a little more meaningful but I'd still prefer to not expose those details to the world.
Regarding how to expose this, I very much would prefer to have command line flags to do this. Having to use environment variables for this is super annoying in git (and is inconsistent between author and committer dates).
If possible, I'd like to have a config flag that told jujutsu to always keep both in sync (i.e. to update them both at the same time all the time).

icorbrey:
> Why should people know or care when I first committed the first draft of a change? It's not even that useful since when you first committed something and when you first started working on it may be two completely different dates... 

I'd definitely like to echo this sentiment, I run into this all the time and Azure DevOps doesn't know how to handle it wrt PRs etc.

image

joyously:
> It's not even that useful since when you first committed something and when you first started working on it may be two completely different dates

It can be useful for some people. I see it like the file system, which has creation date and modified date, but most people only look at modified date. (I forget, is there access date too?)

PhilipMetzger:
> > Why should people know or care when I first committed the first draft of a change? It's not even that useful since when you first committed something and when you first started working on it may be two completely different dates...
> 
> I'd definitely like to echo this sentiment, I run into this all the time and Azure DevOps doesn't know how to handle it wrt PRs etc.

FYI, the author date is set on `jj new` while the commiter date gets set whenever the commit is rewritten (the reason for this is https://github.com/jj-vcs/jj/issues/2000). I still like the older model though.

AngelEzquerra:
> > > Why should people know or care when I first committed the first draft of a change? It's not even that useful since when you first committed something and when you first started working on it may be two completely different dates...
> > 
> > 
> > I'd definitely like to echo this sentiment, I run into this all the time and Azure DevOps doesn't know how to handle it wrt PRs etc.
> 
> FYI, the author date is set on `jj new` while the commiter date gets set whenever the commit is rewritten (the reason for this is #2000). I still like the older model though.

Thanks for the pointer to #2000. If I understand that discussion jj currently sets the author date when a commit becomes non discardable, right? I guess that's sensible _if you want to share the start of the work as well as the end of the work_ (which some people in the thread find useful).

IMHO, the author / committer dichotomy is yet another example of git overcomplicating things unnecessarily _and_ exposing that complexity to the user by default. Pretty much every other VCS I know doesn't have that concept and they get by fine without it.

I think it'd be better to hide the author / committer dichotomy by default. In the same way that jj has simplified git's workflow in many ways, it could do the same here are remove yet another layer of git's complexity (or at least hide it by default).

I realize that this discussion goes a bit beyond the scope of this PR (which I think goes in the right direction).

OvidiusCicero:
Added tests, documented the commands and added the same functionality with respect to committer. There is one TODO I would need your oppinion on it's reviewable from my stand

(sorry I was distracted the last weeks)

bnjmnt4n:
https://github.com/jj-vcs/jj/pull/7184 was actually just merged which adds a command for updating the revision metadata. https://github.com/jj-vcs/jj/pull/7184#issuecomment-3189378788 suggests that the `jj describe --reset-author/--author` should be removed following this new command, so perhaps we should add `jj touch --author-date` instead?

OvidiusCicero:
moved the flags to `jj touch` and added a changelog entry

CHANGELOG.md
PhilipMetzger, line 1: nit: you can squash this into the implementing commit.

cli/src/commands/touch.rs
```diff
@@ -81,6 +85,32 @@ pub(crate) struct TouchArgs {
...
     author: Option<(String, String)>,
+    /// Set the author date to the given date either in RFC 2822 (human
+    /// readable, eg Sun, 23 Jan 2000 01:23:45 JST) or RFC 3339 (time stamp, eg
+    /// 2000-01-23T01:23:45+09:00)
+    #[arg(
+        long,
+        conflicts_with = "update_author_timestamp",
+        value_parser = parse_date
+    )]
+    author_date: Option<DateTime<FixedOffset>>,
```
glehmann: It should use `timestamp` as in the other options, or the other options should be renamed?

cli/src/commands/touch.rs
```diff
@@ -81,6 +85,32 @@ pub(crate) struct TouchArgs {
...
+    )]
+    committer: Option<(String, String)>,
+    /// Set the committer date to the given date either in RFC 2822 (human
+    /// readable, eg Sun, 23 Jan 2000 01:23:45 JST) or RFC 3339 (time stamp, eg
+    /// 2000-01-23T01:23:45+09:00)
+    #[arg(
+        long,
+        value_parser = parse_date
+    )]
+    committer_date: Option<DateTime<FixedOffset>>,
```
glehmann: I think you may have the same objections I had when I tried to add a flag to modify the committer timestamp.
See https://github.com/jj-vcs/jj/pull/7184#discussion_r2265235061

cli/src/text_util.rs
```diff
@@ -547,6 +549,10 @@ pub fn parse_author(author: &str) -> Result<(String, String), &'static str> {
     Ok((captures["name"].to_string(), captures["email"].to_string()))
 }
 
+pub fn parse_date(date: &str) -> chrono::ParseResult<DateTime<FixedOffset>> {
+    DateTime::parse_from_rfc2822(date).or_else(|_| DateTime::parse_from_rfc3339(date))
+}
```
yuja: We have similar code in lib/time_util.rs. Maybe we can extract a common parsing function?

cli/src/commands/touch.rs
OvidiusCicero: removed it

cli/src/commands/touch.rs
OvidiusCicero: renamed to time stamp

cli/src/text_util.rs
OvidiusCicero: I am honestly not entirely sure what you mean. There is date parsing code there yes  but this date parsing code is a thin wrapper around the interim crate adapting for ranges of dates. So I did use `interim::parse_date_string` instead

OvidiusCicero:
uhm can someone explain to me why the test fails in CI but not locally for me?

glehmann:
I successfully ran the test suite with your branch on my computer. Not sure what is happening…

cli/src/text_util.rs
yuja: I think it's still better to add a thin jj-lib wrapper that returns our `Timestamp` type. It will help us use the same parsing function.

cli/src/commands/touch.rs
yuja: Just a comment. We have `author_date()` revset function, but in template (and internal API), it's called `author().timestamp()`. Should we rename the revset function as well?

OvidiusCicero:
it might have something to do with the date parsing now being dependent on `now()` and us both @glehmann sitting in the same time zone (central europe). Will have a look in the next days. If someone can enlighten me I'd still be grateful

yuja:
Appears that `interim` converts timezone internally to the `now` timezone, so there's no way to know the timezone specified by the user?

https://github.com/conradludgate/interim?tab=readme-ov-file#api

OvidiusCicero:
and log output does not show time zones, so you only see it in a different commit ID even though the test explicitly sets a time zone.

@yuja what follows from this? Revert back to the “home grown” combination of two std functions and move this to `lib/time_utils.rs`? Still use this but fix the test, knowing this breaks tests locally and yields wrong results even when inputting a time zone explicitly? (not liking the latter one)

Fixing interim is probably out of reach for me

yuja:
> and log output does not show time zones, so you only see it in a different commit ID even though the test explicitly sets a time zone.

Yes. I think we should include non-local timestamps in those tests.

> Revert back to the “home grown” combination of two std functions and move this to `lib/time_utils.rs`?

Yeah, we'll have to parse timezone-aware strings by ourselves. I don't know how many zone-aware forms `interim` supports, so I'm not sure if it's feasible to intercept all of them and fall back to `interim`. Feel free to leave TODO comments about that.

OvidiusCicero:
did just that, moved the parsing function back to the self written one but moved that to the lib, left a comment and added a test which includes the time zone in output

cli/src/commands/touch.rs
```diff
@@ -81,6 +83,24 @@ pub(crate) struct TouchArgs {
         value_parser = parse_author
     )]
     author: Option<(String, String)>,
+    /// Set the author date to the given date either human
```
yuja, line 87: nit: can you insert blank line between options? Looks like it is the style we use here.

cli/src/commands/touch.rs
```diff
@@ -81,6 +83,24 @@ pub(crate) struct TouchArgs {
...
+    author_timestamp: Option<Timestamp>,
+    /// Set committer to the provided string
+    ///
+    /// This changes committer name and email while retaining committer
+    /// timestamp for non-discardable commits.
+    #[arg(
+        long,
+        value_parser = parse_author
+    )]
+    committer: Option<(String, String)>,
```
yuja: Do you need the `--committer` flag? It seems a bit odd that we have `--committer`, but no `--committer-timestamp` nor `--update-committer`.

Can you also update the commit description to match the current implementation?

cli/tests/test_touch_command.rs
```diff
@@ -240,32 +240,142 @@ fn test_touch() {
...
+       Committer: (no name set) <(no email set)> (1970-01-01 11:00:00)
+
+           (no description set)
+
+    [EOF]
+    ");
+
+    // explicitly test full time stamp to check for time zone
+    insta::assert_snapshot!(work_dir.run_jj(["log", "-r", "@", "-T", "author.timestamp()"]).success(), @r"
+    @  1995-12-19 16:39:57.000 -08:00
```
yuja: I think we can instead make `get_log()` print time zone. Something like `--config='template-aliases.format_timestamp(t)'='t'` should work.

Can you insert a commit to update the test output?

lib/src/time_util.rs
```diff
@@ -122,6 +122,16 @@ impl DatePattern {
     }
 }
 
+// @TODO ideally we would have this unified with the other parsing code. However
+// we use the interim crate which does not handle explicitly given time zone
+// information
+/// Parse a string with time zone information into a `Timestamp`
+pub fn parse_datetime(s: &str) -> chrono::ParseResult<Timestamp> {
+    Ok(Timestamp::from_datetime(
+        DateTime::parse_from_rfc2822(s).or_else(|_| DateTime::parse_from_rfc3339(s))?,
```
yuja, line 131: Can you add a test that captures the error message?

cli/src/commands/touch.rs
OvidiusCicero: we ended up having it because martin found it odd to only have an author flag…

Given the argumentations here to not make it possible to change the dates around committer I don't find the asymmetry odd tbh

lib/src/time_util.rs
OvidiusCicero: added a unit test and a system test

cli/tests/test_touch_command.rs
OvidiusCicero: done

cli/src/commands/touch.rs
OvidiusCicero: done

cli/src/commands/touch.rs
OvidiusCicero: actually lets leave it out

PR #7140, opened 2025-07-31
Preparations for making `merge_trees()` concurrent and non-recursive

martinvonz:
# Checklist

If applicable:

### DIFF
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -501,8 +501,8 @@ async fn merge_trees(merge: Merge<Tree>) -> BackendResult<Merge<Tree>> {
     }
     if conflicts.is_empty() {
         let data = backend::Tree::from_sorted_entries(new_tree_entries);
-        let new_tree_id = store.write_tree(dir, data).await?;
-        Ok(Merge::resolved(new_tree_id))
+        let new_tree = store.write_tree(dir, data).await?;
+        Ok(Merge::resolved(new_tree))
     } else {
         // For each side of the conflict, overwrite the entries in `new_tree` with the
         // values from  `conflicts`. Entries that are not in `conflicts` will remain
--- a/lib/src/merged_tree.rs
+++ b/lib/src/merged_tree.rs
@@ -31,6 +31,7 @@ use futures::Stream;
 use futures::StreamExt as _;
 use futures::future::BoxFuture;
 use futures::future::try_join;
+use futures::future::try_join_all;
 use futures::stream::BoxStream;
 use itertools::EitherOrBoth;
 use itertools::Itertools as _;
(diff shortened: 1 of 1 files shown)


lib/src/merged_tree.rs
```diff
@@ -518,12 +516,19 @@ async fn merge_trees(merge: Merge<Tree>) -> BackendResult<Merge<Tree>> {
...
+            backend_trees.push(backend::Tree::from_sorted_entries(
+                new_tree_entries.into_iter().collect(),
+            ));
         }
-        Ok(Merge::from_vec(new_trees))
+        Merge::from_vec(backend_trees)
+    };
+
+    let mut new_trees = Vec::with_capacity(backend_trees.iter().count());
+    for backend_tree in backend_trees.into_iter() {
```
yuja: nit: `.into_iter()` can be omitted (this code will be removed, though.)

lib/src/merge.rs
```diff
@@ -477,7 +471,15 @@ impl<T> Merge<Option<T>> {
 impl<T: Clone> Merge<Option<&T>> {
     /// Creates a new merge by cloning inner `Option<&T>`s.
     pub fn cloned(&self) -> Merge<Option<T>> {
-        self.map(|value| value.cloned())
+        self.as_ref().map(|value| value.cloned())
+    }
+}
+
+impl<T> Merge<T> {
```
yuja: nit: merge with the previous `impl<T> Merge<T>` block?

lib/src/merge.rs
```diff
@@ -477,7 +471,15 @@ impl<T> Merge<Option<T>> {
...
+    /// Creates a new merge where each value is a reference to the original
+    /// merge's value.
+    pub fn as_ref(&self) -> Merge<&T> {
+        Merge::from_vec(self.iter().collect_vec())
```
yuja: nit: rename to `to_ref()`? Since this involves allocation of new vec, it shouldn't be called `as_<thing>()`.

FWIW, I don't think the compiler is clever enough to optimize stuff across allocation points.

lib/src/merged_tree.rs
```diff
@@ -524,12 +524,9 @@ async fn merge_trees(merge: Merge<Tree>) -> BackendResult<Merge<Tree>> {
...
 
-    let mut new_trees = Vec::with_capacity(backend_trees.iter().count());
-    for backend_tree in backend_trees.into_iter() {
-        let new_tree = store.write_tree(dir, backend_tree).await?;
-        new_trees.push(new_tree);
-    }
-    Ok(Merge::from_vec(new_trees))
+    backend_trees
+        .try_map_async(|backend_tree| store.write_tree(dir, backend_tree))
+        .await
```
yuja: nit: another options is to pass `Arc<backend::Tree>` to `.write_tree()` so we can cheaply clone it.

lib/src/merged_tree.rs
martinvonz: Done.

lib/src/merge.rs
martinvonz: Ah, yes, renamed to `to_ref()`.

lib/src/merge.rs
martinvonz: Done.

lib/src/merged_tree.rs
martinvonz: Good point. I wonder if that's a good idea in general, actually. It's possible that it would allow some backends to do some less cloning too, in case they want to keep their own cache or something. We already wrap trees and commits in `Arc` in `Store`.

martinvonz: I removed that commit and left a TODO for now.

PR #7145, opened 2025-07-31
revset: insert `HeadsRange` in `Ancestors` for nested range/filter 

scott2000:
This allows evaluating ancestors/ranges involving filters significantly faster. For instance, naively evaluating `::mine()` requires reading every commit in the repo to check `mine()` on each commit, then finding the ancestors of that set. This optimization rewrites `::mine()` to `::heads(mine())`, since `heads(mine())` can be evaluated more efficiently by only reading commits until the first successful match.

If someone is unaware of how revsets are implemented, this case can come up pretty easily, such as by including `~mine()` in `immutable_heads()` to make other people's commits immutable. In my local `jj` repo, this optimization reduces the runtime of `jj log` with `~mine()` in `immutable_heads()` from about 800ms down to about 50ms.

```
::(v1.0.0..v2.40.0)     55.5% faster
author(peff)..          96.8% faster
```

# Checklist

If applicable:

### DIFF
--- a/cli/testing/bench-revsets-git.txt
+++ b/cli/testing/bench-revsets-git.txt
@@ -42,6 +42,9 @@ heads(author(peff))
 # Roots and heads of range
 roots(::v2.40.0)
 heads(::v2.40.0)
+# Ancestors and ranges of large subsets
+::(v1.0.0..v2.40.0)
+author(peff)..
 # Parents and ancestors of old commit
 v1.0.0-
 v1.0.0---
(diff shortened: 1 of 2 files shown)


martinvonz (commented):
Thanks. This looks good to me. Yuya knows the revset engine best, so I'll leave approval for him.

cli/src/commands/diff.rs
```diff
@@ -158,7 +158,7 @@ pub(crate) fn cmd_diff(
             .attach_revset_evaluator(
                 target_expression
                     .roots()
-                    .range(&target_expression.heads())
+                    .range(target_expression)
```
yuja: nit: I wouldn't change this because we use `.roots()` for the other end.

yuja (approved):
LGTM. Can you add some examples to cli/testing/bench-revsets-git.txt?

PR #1298, opened 2023-02-20
cli: New `--changes-in` argument to `jj restore`

ilyagr:
This argument can restore a merge commit to the merge of its parents. This allows one to apply the default behavior of `jj restore` to another commit. In other words, `jj restore` without arguments is equivalent to `jj restore -c @`. 

### Original description (before switching `-r` => `--rp`)

The title was "cmd: Add `-r` argument to `jj restore

The interface is now similar to `jj diffedit`, `jj diff`, etc.

# Checklist

If applicable:
- [x] I have updated `CHANGELOG.md`
- (not planned) I have updated the documentation (README.md, docs/, demos/)
- (n/a) I have updated the config schema (src/config-schema.json)
- [x] I have added tests to cover my changes

PR #7146, opened 2025-08-01
restore: adjust error message for `--revision`

mernen:
Commands like `diff`, `squash` and `restore` are directional, taking `--from` and `--to`/`--into` parameters. The former two also support `-r`, as a shortcut when one wishes to operate on a revision in relation to its parent(s).

Users who know that might deduce then that `restore` would also support `-r`, but it doesn’t — at some point it was decided that `-r` would be ambiguous, and people could attempt `restore -r <revision>` when they actually mean `-f`. Searching for “restore r” on Discord, I found a few old messages where people seemingly made this exact mistake, so I presume it’s a fair concern, particularly for a command that could be so critical to newcomers.

The command does know people may try to use `-r`, though, and rathern than showing a generic “unexpected argument” error, it shows a specific message:

```
Error: `jj restore` does not have a `--revision`/`-r` option. If you'd like to modify
the *current* revision, use `--from`. If you'd like to modify a *different* revision,
use `--into` or `--changes-in`.
```

Looking at my shell history, over the last 5 months I’ve typed `jj restore -r` at least 12 times, and *not once* did I read the error message, or I would have realized there’s `--changes-in`/`-c`. Personal embarrassments aside, I suspect the reason I always skipped the message without even skimming is that it feels like a wall of text.

This PR changes the message to:

```
Error: `jj restore` does not have a `--revision`/`-r` option.
Hint: If you'd like to modify the current revision, use `--from`.
Hint: To undo changes in a revision compared to its parents, use `--changes-in`.
```

- The “Error” part is significantly shorter and more objective.
- Each line is self-contained, and the “Hint:” headings (in cyan, not shown here) should help grab the user’s attention if they were looking for one of the alternatives.
- The mention of `--into` was removed, as it doesn’t seem like anyone confuses `-r` with that.
- The longest line is exactly 80 characters with the “Hint:” header, so it shouldn’t wrap in most terminals — worst case it fills the line and the user gets a blank line after.

# Checklist

If applicable:

---

I also considered highlighting `--from` and `--changes-in`, but I don’t think any of the defined labels would fit — we’d have to define a new label, e.g.:

```toml
[colors]
"highlight" = { bold = true }
```

### DIFF
--- a/cli/src/commands/restore.rs
+++ b/cli/src/commands/restore.rs
@@ -117,11 +117,13 @@ pub(crate) fn cmd_restore(
     let mut workspace_command = command.workspace_helper(ui)?;
     let (from_commits, from_tree, to_commit);
     if args.revision.is_some() {
-        return Err(user_error(
-            "`jj restore` does not have a `--revision`/`-r` option. If you'd like to modify\nthe \
-             *current* revision, use `--from`. If you'd like to modify a *different* \
-             revision,\nuse `--into` or `--changes-in`.",
-        ));
+        return Err(
+            user_error("`jj restore` does not have a `--revision`/`-r` option.")
+                .hinted("To modify the current revision, use `--from`.")
+                .hinted(
+                    "To undo changes in a revision compared to its parents, use `--changes-in`.",
+                ),
+        );
     }
     if args.from.is_some() || args.into.is_some() {
         to_commit = workspace_command
--- a/cli/tests/test_restore_command.rs
+++ b/cli/tests/test_restore_command.rs
@@ -35,9 +35,9 @@ fn test_restore() {
     let output = work_dir.run_jj(["restore", "-r=@-"]);
     insta::assert_snapshot!(output, @r"
     ------- stderr -------
-    Error: `jj restore` does not have a `--revision`/`-r` option. If you'd like to modify
-    the *current* revision, use `--from`. If you'd like to modify a *different* revision,
-    use `--into` or `--changes-in`.
+    Error: `jj restore` does not have a `--revision`/`-r` option.
+    Hint: To modify the current revision, use `--from`.
+    Hint: To undo changes in a revision compared to its parents, use `--changes-in`.
     [EOF]
     [exit status: 1]
     ");


ilyagr:
I think this is an improvement. Thank you!

I am not sure `-r` cannot be confused with `--into`, but I suppose we could see if anybody complains about that. I agree that `--from` and `--changes-in` are more important to mention.

I'll wait for other comments or a better idea of what to do with `--into` before formally approving.

> I also considered highlighting --from and --changes-in, but I don’t think any of the defined labels would fit — we’d have to define a new label

We could consider using the same styling that the help text uses for arguments. I remember there was some discussion of configuring the styles `clap` uses to print help with `jj config`, and there was also some discussion of rendering the help ourselves entirely. I don't remember what the latest status of those is.

I think that can be done separately, though. If we wanted to do that, we'd probably want to modify many messages at once to use those styles.

cli/src/commands/restore.rs
```diff
@@ -117,11 +117,13 @@ pub(crate) fn cmd_restore(
...
-            "`jj restore` does not have a `--revision`/`-r` option. If you'd like to modify\nthe \
-             *current* revision, use `--from`. If you'd like to modify a *different* \
-             revision,\nuse `--into` or `--changes-in`.",
-        ));
+        return Err(
+            user_error("`jj restore` does not have a `--revision`/`-r` option.")
+                .hinted("If you'd like to modify the current revision, use `--from`.")
+                .hinted(
+                    "To undo changes in a revision compared to its parents, use `--changes-in`.",
+                ),
```
yuja: nit: "To modify the current revision" for consistency?

yuja (commented):
Looks also good to me. Can you update the commit description to include a short version of the PR description? We don't care the PR description, btw.

cli/src/commands/restore.rs
mernen: I'm not sure what you mean here — isn't that exactly what's written already? If you mean changing something on the second hint, I didn't understand.

yuja: I mean it seems inconsistent that the first hint says "If you'd like to", but the second one is just "To".

joyously: Two things:
1) "undo changes" uses both a different command verb and the established name for a piece of work (labeled by change ID), which is kind of confusing when talking about `restore` and revision.
2) I don't understand what "changes in a revision compared to its parents" means.

joyously:
It seems like a general rule of "an option error should show the option section of the help" would solve a bunch of these at once instead of one at a time Hints that are incomplete.

mernen:
> It seems like a general rule of "an option error should show the option section of the help" would solve a bunch of these at once instead of one at a time Hints that are incomplete.

The goal of the existing message as well as this proposed change is to redirect the user to the specific options they might want. I don't think dumping 7+ flags would help with that.

cli/src/commands/restore.rs
mernen: 1. I see what you mean, but the "undo changes" was taken from `--changes-in`. Despite conflating things with `jj undo`, it doesn't sound too bad to me.
2. Yeah, I agree it could be better; I restricted myself to an 80-character limit. Not sure if others think it's worth going a bit over that to choose a different phrasing?

joyously:
> The goal of the existing message as well as this proposed change is to redirect the user to the specific options they might want.

The key word here is "might". Since they used an option that doesn't exist, showing them the ones that do should satisfy every case, instead of just the one you guessed at. It saves them from having to type the help command, and gives them the complete explanation for each, whereas the hint is sparse and not as helpful, as well as only addressing one case out of the whole alphabet of option choices for every command in the tool.

cli/src/commands/restore.rs
joyously: 1. If you have a hint, it should word it differently than the help (to give a different perspective), but if you repeat the option help, show the whole thing. The person needing the hint won't be thinking the same way as a developer of the tool, hence why I said it was confusing.
2. "could be better" means you aren't done yet. Your hint inspired confusion. Is that the goal?

ilyagr:
I would hope that people would be likely to type `jj restore --help` after getting this hint if they would like more details. Do you think they wouldn't know to do this? Would saying explicitly "See `jj restore --help` for more information"[^clap] or "See `jj restore --help` about these flags and the `--into` flag that may also be relevant" help?

I think `jj restore --help` is too much information to dump every time, while dumping `jj restore -h` output wouldn't add much to these hints (and it's also pretty long). The length of the docs in `-h` output is similar to the length of the hints, so it's only useful for discovering other options. Apart from `--from` and `--changes-in`, I think `--into` is the only other functionality of `jj restore` the user may want when they use `-r`. So, I'd probably prefer adding another hint about `--into` over dumping the whole `-h` output every time (though I'm torn about adding a `--into` hint, because the more the hints look like a wall of text, the less likely they are to serve their purpose well).

[^clap]: This also reminds me of how `clap` prints "For more information, try '--help'" after every error (including the error that would exist for `jj restore -r` if we didn't customize anything, BTW). I find this entirely useless, except maybe the first few times a user uses a program, and learned to ignore it.

    So, if we do say something like "See `jj restore --help` for more information", I wonder how to make it clear to the user that this message is not generic, but rather we are really suggesting it because we know there's something useful in there. This is why I suggested the second version which, while a little awkward, does make this clearer. Perhaps other people find automated behaviors like `clap`'s more helpful than I do.

cli/src/commands/restore.rs
ilyagr: > Two things: ...

I share these concerns, but similarly to @mernen, I think that perfect shouldn't be the enemy of good (this thesis is the TLDR for this message). We could later separately discuss both improving the help text and improving the hint further. To me, it doesn't seem trivial to make a hint that's better in every way without being overly verbose.

> If you have a hint, it should word it differently than the help (to give a different perspective), but if you repeat the option help, show the whole thing. 

In this case, I think the goal is to quickly communicate what the likely alternatives are. IMO, people would then naturally run `jj restore --help` if they want more details. In that case, they'll see more details about the `--changes-in` option. Again, the help is not perfect, but we can improve it later. (This is also more relevant to the discussion starting with https://github.com/jj-vcs/jj/pull/7146#issuecomment-3144963275, perhaps)

> "could be better" means you aren't done yet. Your hint inspired confusion. Is that the goal?

I think that this PR makes the hint *strictly less confusing*[^exception] than it was before the PR because, for example, it's now marked properly as a hint, and is separated from the error message. I also think that removing the hints entirely or having a generic `clap` error would make things more confusing.

[^exception]: with the potential exception of the `--into` concern that I have, see my other comments; I don't think it's very relevant to this discussion.

The PR also reminds us that our hints and help here are not perfect, which is also helpful, but shouldn't be a blocker to the PR IMO. Incremental improvement is a good thing. Perhaps, in the future, we'll replace these hints entirely. Perhaps we'll continue to find ways to improve them incrementally. Either way, it's not a reason to avoid improving things now.

joyously:
> I would hope that people would be likely to type `jj restore --help` after getting this hint

You miss the point. I'm advocating for saving them the typing, answering all queries instead of just a guess at which one they might want, and using it for all commands that have an option error. And not the entire help, just the option help (a condensed list of options for that command).

mernen:
> The key word here is "might".

It’s “might” in the sense there are two possible interpretations, not seven. We’re not guessing wildly here. Earlier discussion on why not have `-r` in the first place as well as why have this specialized guiding message rather leaving it to the default happened in #1298. If we don’t point this out, they’re more likely to assume there’s no equivalent to `squash -r` and do it manually with `-f <revision>- -t <revision>`, as we’ve seen in chat logs happening at least twice, and possibly ask for this feature even though it already exists with a(n intentionally) different name.

And now we have the history to confirm both interpretations occur in practice, plus at least two data points that the current message could be improved in a way that simply dumping all options is unlikely to help.

cli/src/commands/restore.rs
joyously: I'm all for improving the output. But I think both of you are actually writing code for this project, and so can't accurately assess what is confusing to someone who needs the hint. I'm slightly beyond novice at `jj`, and I say it's confusing, and would be better served seeing the list of options when I make a mistake on an option.

mernen: I think that's fair, which is why I mentioned I'm not perfectly happy with it either. But so far we're yet to see a better alternative — IMO this is better than the current message, better than removing the `-r` special case entirely, and also better than printing several unrelated flags which are definitely not what the user intended.

If anyone has a better short description for `--changes-in` we could change it in this PR or even in the future, but ultimately this hint is more focused on bringing awareness to the option that can be easily missed.

mernen: Proposal for the second hint:

> ~~``Hint: To undo changes in a revision compared to its parents, use `--changes-in`.``~~  
> ``Hint: To undo changes in a specific revision, use `--changes-in`.``

The parents part would be implied by the analogy to `diff -r`/`squash -r`, if this is how you arrived at this option. There's still “undo”, which is unfortunate, but the only other word I can think of is “revert”, which is already defined to mean creating a new commit rather than modifying the existing one.

ilyagr: Yes, I originally wrote the "undo" text, and it was very much a compromise. I am still not very happy about it, and I still don't have a much better idea.

> Hint: To undo changes in a specific revision, use `--changes-in`.

I think that makes `--changes-in` harder to distinguish from `--into`, depending on where the target revision is. We could try something like this text, but then we'd probably want to explicitly mention `--into` and how to think of that.

Another approach I was thinking of was to say something like "to undo the changes you'd see with `jj diff -r revision`, use `--changes-in`".

--------

Overall, unless we come up with something we're happy with quickly, I'd postpone that to another PR.

joyously: Are you saying that `jj restore --changes-in revision`  is being described as 
> undo changes in revision

?
Those sound opposite to me.

joyously: Does it really mean "restore the working copy to a specific revision"? (I don't understand the `--changes-in` thing)

ilyagr (approved):
I'll approve this since I still think it improves the previous version (https://github.com/jj-vcs/jj/pull/7146#discussion_r2248890992). We can improve it further in separate PRs and discussions.

ilyagr:
>  I'm advocating for saving them the typing, answering all queries instead of just a guess at which one they might want

I think we could keep discussing these sorts of ideas elsewhere (and we can keep discussing it here after it's merged). I think I don't agree with this specific idea for the reasons I tried to explain in the second half of an the footnote of https://github.com/jj-vcs/jj/pull/7146#issuecomment-3145787815, but I also may have misunderstood what exactly you meant.

Unfortunately, I personally may or may not have the time and energy to be involved in thinking about this too heavily, but perhaps I will, or perhaps you'll find other people to throw around some ideas with, or perhaps the discussion might be helpful even if it goes pretty slowly.

cli/src/commands/restore.rs
ilyagr: > Does it really mean "restore the working copy to a specific revision"?

This is what `jj restore --from specific_revision` would do.

OTOH, `jj restore --changes-in some_revision` would make `some_revision` empty. If you specify some paths, then the command will not make that revision *completely* empty, but it will make it "emptier": any changes `some_revision` made to those paths will disappear.

As an example, if `some_revision` had a single parent, `jj restore --changes-in some_revision` would be the same as `jj restore --from some_revision- --into some_revision`. This is conceptually similar to how, in the single-parent case, `jj diff -r some_revision` is equivalent to `jj diff --from some_revision- --to some_revision`.

cli/src/commands/restore.rs
ilyagr, line 1: (Non-blocking) Minor comments re the commit description:

You could have made it longer and put even more information from the PR comment into there. Anything that can answer the question "why are we doing it, and why we are doing it this way" can go there. I don't think you have to, though.

I wouldn't say "Remove mention of `--into` as it doesn't seem like a common
interpretation of `-r`", I still have doubts about that. (Or do you have some reason to think that? In that case, you could mention it). I'd say something like "Remove mention of `--into` for brevity, hoping that it's not a common interpretation of `-r`". But it doesn't matter much.

mernen: I've pushed an updated description that hopefully provides context for the change. Regarding `--into`, the reason I considered it an unlikely interpretation is that I couldn't find anyone arguing for it or even using it by mistake, either on GitHub discussions or on Discord, whereas there were occasions where people mentioned `restore -r` but from context I could see they actually meant `--from`, and not `--changes-in`.

PR #7151, opened 2025-08-01
Write `jj status` instead of `jj st` in docs

pingiun:
I think it's better to have full commands instead of abbreviated commands in the docs

# Checklist

N/A (only docs change)

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -35,10 +35,10 @@ Added 1 files, modified 0 files, removed 0 files
 $ cd Hello-World
 ```
 
-Running `jj st` (short for `jj status`) now yields something like this:
+Running `jj status` now yields something like this:
 
 ```shell
-$ jj st
+$ jj status
 The working copy has no changes.
 Working copy  (@) : kntqzsqt d7439b06 (empty) (no description set)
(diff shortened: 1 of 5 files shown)


PhilipMetzger:
can you prefix the commit with message with `docs: <Your Title>` to adhere to our commit guidelines, thanks.

docs/working-copy.md
martinvonz, line 1: Searching for `jj st` (with the backticks) in `docs/` yields several more matches. Can you fix the others too (where it makes sense, if not everywhere)?

pingiun: I've checked all the matches, and changed them

docs/tutorial.md
```diff
@@ -35,10 +35,10 @@ Added 1 files, modified 0 files, removed 0 files
 $ cd Hello-World
 ```
 
-Running `jj st` (short for `jj status`) now yields something like this:
+Running `jj status` (short for `jj status`) now yields something like this:
```
bnjmnt4n:
```suggestion
Running `jj status` now yields something like this:
```

docs/tutorial.md
PhilipMetzger, line 1: nit: squash this into the parent

